PR #794, opened 2019-09-14
stacks: RBAC policy rules should not have any version information

jbw976:
This PR ensures that no version information goes into the RBAC policy rules that the Stack Manager creates.  Those rules should not have any version information and will not work if versions are included.

This was tested end to end with the Wordpress stack (with updates from https://github.com/crossplaneio/sample-stack-wordpress/pull/12).

### Checklist

I have:

### DIFF
--- a/pkg/stacks/unpack.go
+++ b/pkg/stacks/unpack.go
@@ -286,9 +286,12 @@ func (sp *StackPackage) applyAnnotations() {
 	}
 }
 
-func generateRBAC(apikind, apigroupversion string) rbacv1.PolicyRule {
+// generateRBAC generates a RBAC policy rule for the given kind and group.
+// Note that apiGroup should not contain a version, only the group, e.g., database.crossplane.io
+// RBAC policy rules are intended to be versionless.
+func generateRBAC(apikind, apiGroup string) rbacv1.PolicyRule {
 	return rbacv1.PolicyRule{
-		APIGroups:     []string{apigroupversion},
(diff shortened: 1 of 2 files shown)


pkg/stacks/unpack.go
```diff
@@ -311,21 +314,25 @@ func (sp *StackPackage) applyRules() error {
...
+			// versions are not allowed in RBAC PolicyRules, remove any trailing version denoted by a "/"
+			// e.g., kind.group.com/v1alpha1 -> kind.group.com
+			if i := strings.Index(crd, "/"); i != -1 {
+				crd = crd[:i]
+			}
+
+			gk := schema.ParseGroupKind(crd)
 			if gk.Group == "" || gk.Kind == "" {
-				return errors.New(fmt.Sprintf("cannot parse CustomResourceDefinition %q as Kind and GroupVersion", dependency.CustomResourceDefinition))
+				return errors.New(fmt.Sprintf("cannot parse CustomResourceDefinition %q as Kind and Group", crd))
```
suskin: Nitpick: in an error condition, it may be helpful for the human to have the full original input too, before it was trimmed. Did you already consider including the original `dependency.CustomResourceDefinition` here as well?

jbw976: i thought about that, but I wanted to keep the error message focused exactly what was passed (and failed) to the ParseGroupKind function, so as to not introduce any ambiguity.

PR #802, opened 2019-09-16
GCP Services Guide

muvaf:

This PR adds GCP Services Guide by installing Wordpress using an existing cluster but provisioning database via Crossplane.
Tested end-to-end. Inspired by https://github.com/crossplaneio/crossplane/pull/800

Related to https://github.com/crossplaneio/crossplane/issues/778

### Checklist

I have:

### DIFF
--- a/docs/services/gcp-services-guide.md
+++ b/docs/services/gcp-services-guide.md
@@ -1,0 +1,620 @@
+---
+title: "GCP Services Guide"
+toc: true
+weight: 420
+indent: true
+---
+
+# GCP Services Guide
+
+## Table of Contents
+
+1. [Introduction](#introduction)
+2. [Pre-requisites](#pre-requisites)
+3. [Preparation](#preparation)
+     1. [GKE Cluster Steps](#gke-cluster-steps)
+     2. [Crossplane Steps](#crossplane-steps)
+          1. [Installation](#installation)
+          2. [GCP Provider](#gcp-provider)
+          3. [Cloud Specific Resource Classes](#cloud-specific-resource-classes)
+          4. [Namespaces](#namespaces)
+          5. [Portable Resource Classes](#portable-resource-classes)
+          6. [Connecting MySQL Instance and GKE Cluster](#connecting-mysql-instance-and-gke-cluster)
+4. [Install Wordpress](#install-wordpress)
+5. [Clean Up](#clean-up)
+     1. [Wordpress](#wordpress)
+     2. [Crossplane](#crossplane)
+
+## Introduction
+
+This user guide will walk you through Wordpress application deployment using
+your existing Kubernetes cluster and Crossplane managed resources. We will:
+* Install Crossplane to your cluster.
+* Create necessary resource classes for MySQL database instance.
+* Create network resources to get GKE cluster to connect to MySQL instance.
+* Deploy Wordpress.
+
+## Pre-requisites
+
+* [gcloud CLI](https://cloud.google.com/sdk/docs/quickstarts)
+    * Make sure to configure the authentication after installing gcloud cli.
+* [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/)
+* A GKE cluster.
+
+## Preparation
+
+This guide assumes that you have setup the gcloud CLI and are logged in to your
+desired account.
+
+*Note: environment variables are used throughout this guide. You should use
+your own values.*
+
+Run the following:
+```bash
+export PROJECT_ID=crossplane-playground # the project that all resources reside.
+export NETWORK_NAME=default # the network that your GKE cluster lives in.
+export SUBNETWORK_NAME=default # the subnetwork that your GKE cluster lives in.
+``` 
+
+### GKE Cluster Steps
+
+We are assuming you've got a GKE cluster up and running created either via 
+GCP Console or gcloud CLI. If you don't, the basic command to create one is
+the following:
+
+1. Create a GKE cluster:
+```bash
+gcloud beta container --project "${PROJECT_ID}" clusters create "my-existing-cluster" \
+  --zone "us-central1-a" \
+  --cluster-version "1.13.7-gke.8" \
+  --machine-type "n1-standard-1" \
+  --image-type "COS" \
+  --disk-type "pd-standard" \
+  --disk-size "100" \
+  --metadata disable-legacy-endpoints=true \
+  --scopes "https://www.googleapis.com/auth/devstorage.read_only","https://www.googleapis.com/auth/logging.write","https://www.googleapis.com/auth/monitoring","https://www.googleapis.com/auth/servicecontrol","https://www.googleapis.com/auth/service.management.readonly","https://www.googleapis.com/auth/trace.append" \
+  --num-nodes "1" \
+  --enable-cloud-logging --enable-cloud-monitoring --enable-ip-alias --enable-autoupgrade --enable-autorepair --no-enable-basic-auth \
+  --network "projects/${PROJECT_ID}/global/networks/${NETWORK_NAME}" \
+  --subnetwork "projects/${PROJECT_ID}/regions/us-central1/subnetworks/${SUBNETWORK_NAME}" \
+  --default-max-pods-per-node "110" \
+  --addons HorizontalPodAutoscaling,HttpLoadBalancing
+```
+
+This may take a while to finish.
+
+2. Connect to your GKE cluster
+```bash
+gcloud container clusters get-credentials "my-existing-cluster" --zone us-central1-a --project "${PROJECT_ID}"
+```
+
+3. Make sure `kubectl` is able to communicate with your GKE cluster
+```bash
+kubectl cluster-info
+```
+4. Make sure Helm is installed with permissions to work on `crossplane-system`
+namespace. The easiest way to achieve that is to run the following command
+that makes Helm's server side component `tiller` cluster admin:
+```bash
+kubectl -n kube-system create serviceaccount tiller
+kubectl create clusterrolebinding tiller --clusterrole cluster-admin --serviceaccount=kube-system:tiller
+helm init --service-account=tiller
+```
+### Crossplane Steps
+
+#### Installation
+
+* Install Crossplane from alpha channel using the [Crossplane Installation
+  Guide](../install-crossplane.md#alpha)
+* Install the GCP stack into Crossplane using the [GCP stack
+  section](../install-crossplane.md#gcp-stack) of the install guide.
+
+We will be using `gcp-infra-dev` namespace to store our cloud-specific resource
+classes, cloud resources and providers. But you can use any namespace name for
+that matter.
+```bash
+kubectl create namespace gcp-infra-dev
+export INFRA_NAMESPACE=gcp-infra-dev
+```
+
+#### GCP Provider
+
+It is essential to make sure that a GCP service account to be used by 
+Crossplane is configured with all the necessary permissions outlined in the 
+[provider guide](../cloud-providers/gcp/gcp-provider.md).
+
+Using the service account json `crossplane-gcp-provider-key.json` that you
+acquired from GCP:
+
+* Generate Base64 encoded value to store in a `Secret`:
+```bash
+export BASE64ENCODED_GCP_PROVIDER_CREDS=$(base64 crossplane-gcp-provider-key.json | tr -d "\n")
+```
+
+* Define a GCP `Provider` and `Secret`:
+```bash
+cat > gcp-provider.yaml <<EOF
+---
+apiVersion: v1
+data:
+  credentials.json: $BASE64ENCODED_GCP_PROVIDER_CREDS
+kind: Secret
+metadata:
+  name: gcp-provider-creds
+type: Opaque
+---
+apiVersion: gcp.crossplane.io/v1alpha2
+kind: Provider
+metadata:
+  name: gcp-provider
+spec:
+  credentialsSecretRef:
+    name: gcp-provider-creds
+    key: credentials.json
+  projectID: $PROJECT_ID
+EOF
+
+kubectl -n $INFRA_NAMESPACE apply -f gcp-provider.yaml
+unset BASE64ENCODED_GCP_PROVIDER_CREDS # we don't need this anymore.
+```
+* Verify GCP provider was successfully registered by the crossplane
+```bash
+kubectl -n $INFRA_NAMESPACE get providers.gcp.crossplane.io
+kubectl -n $INFRA_NAMESPACE get secrets
+```
+
+#### Cloud-Specific Resource Classes
+
+Cloud-specific resource classes are used to define a reusable configuration for
+a specific managed service. Wordpress requires a MySQL database, which can be
+satisfied by a [Google Cloud SQL Instance](https://cloud.google.com/sql/docs/mysql/).
+
+* Define a GCP CloudSQL class `CloudsqlInstanceClass`:
+```bash
+cat > gcp-mysql-standard.yaml <<EOF
+---
+apiVersion: database.gcp.crossplane.io/v1alpha2
+kind: CloudsqlInstanceClass
+metadata:
+  name: standard-cloudsql
+specTemplate:
+  databaseVersion: MYSQL_5_7
+  tier: db-n1-standard-1
+  region: us-central1
+  storageType: PD_SSD
+  storageGB: 10
+  # Note from GCP Docs: Your Cloud SQL instances are not created in your VPC network.
+  # They are created in the service producer network (a VPC network internal to Google) that is then connected (peered) to your VPC network.
+  privateNetwork: projects/$PROJECT_ID/global/networks/$NETWORK_NAME
+  providerRef:
+    name: gcp-provider
+    namespace: $INFRA_NAMESPACE
+  reclaimPolicy: Delete
+EOF
+
+kubectl -n $INFRA_NAMESPACE apply -f gcp-mysql-standard.yaml
+```
+* You can verify creation with the following command and output:
+
+*Command*
+```bash
+kubectl get cloudsqlinstanceclass -n ${INFRA_NAMESPACE}
+```
+*Output*
+```bash
+NAME                   PROVIDER-REF   RECLAIM-POLICY   AGE
+standard-cloudsql      gcp-provider   Delete           11s
+```
+
+You are free to create more GCP `CloudsqlInstanceClass` instances to define more
+potential configurations. For instance, you may create `large-gcp-mysql` with
+field `storageGB: 100`.
+
+#### Namespaces
+
+Kubernetes namespaces allow for separation of environments within your cluster.
+You may choose to use namespaces to group resources by team, application, or any
+other logical distinction. For this demo, we will create a namespace called
+`app-project1-dev`, which we will use to group our Wordpress resources.
+
+```bash
+kubectl create namespace app-project1-dev
+```
+
+#### Portable Resource Classes
+
+Portable resource classes are used to define a class of service in a single
+namespace for an abstract service type. We want to define our GCP
+`CloudsqlInstanceClass` as the standard MySQL class of service in the namespace 
+that our Wordpress resources will live in.
+
+* Define a `MySQLInstanceClass` in `mysql-standard.yaml` for namespace
+  `app-project1-dev`:
+```bash
+cat > mysql-standard.yaml <<EOF
+---
+apiVersion: database.crossplane.io/v1alpha1
+kind: MySQLInstanceClass
+metadata:
+  name: mysql-standard
+classRef:
+  kind: CloudsqlInstanceClass
+  apiVersion: database.gcp.crossplane.io/v1alpha2
+  name: standard-cloudsql
+  namespace: $INFRA_NAMESPACE
+EOF
+
+kubectl -n app-project1-dev apply -f mysql-standard.yaml
+```
+
+* You can verify creation with the following command and output:
+
+*Command*
+```bash
+kubectl -n app-project1-dev get mysqlinstanceclasses
+```
+*Output*
+```bash
+NAME             AGE
+mysql-standard   27s
+```
+
+Once again, you are free to create more `MySQLInstanceClass` instances in this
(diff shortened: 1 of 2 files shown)


docs/services/gcp-services-guide.md
```diff
@@ -0,0 +1,536 @@
...
+
+5. Make sure `kubectl` is able to communicate with your GKE cluster
+```bash
+kubectl cluster-info
+```
+
+### Crossplane
+Using your GKE cluster:
+
+* Install Crossplane from alpha channel using the [Crossplane Installation
```
prasek: use alpha link instead?
https://github.com/crossplaneio/crossplane/blob/master/docs/install-crossplane.md#alpha

docs/services/gcp-services-guide.md
```diff
@@ -0,0 +1,536 @@
...
+  --subnetwork "projects/${PROJECT_ID}/regions/us-central1/subnetworks/${SUBNETWORK_NAME}" \
+  --default-max-pods-per-node "110" \
+  --addons HorizontalPodAutoscaling,HttpLoadBalancing
+```
+
+This may take a while to finish.
+
+4. Connect to your GKE cluster
+```bash
+gcloud container clusters get-credentials "my-existing-cluster" --zone us-central1-a --project "${PROJECT_ID}"
```
suskin: How does this set up a connection? Does it automatically populate our `kubectl` config for us and switch to it?

docs/services/gcp-services-guide.md
```diff
@@ -0,0 +1,536 @@
...
+5. Make sure `kubectl` is able to communicate with your GKE cluster
+```bash
+kubectl cluster-info
+```
+
+### Crossplane
+Using your GKE cluster:
+
+* Install Crossplane from alpha channel using the [Crossplane Installation
+  Guide](../install-crossplane.md#master)
```
suskin: Could we reproduce the steps here so the user doesn't need to open a bunch of other documents to get through the guide? The way we have been handling this in the stack guide is to give the commands and to tell the user to [refer to the install guide for more detail](https://github.com/crossplaneio/crossplane/pull/803/files#diff-378fbcada1ba53bb6c4e2afb09c18c3aR100)

docs/services/gcp-services-guide.md
```diff
@@ -0,0 +1,536 @@
...
+* Install the GCP stack into Crossplane using the [GCP stack
```
suskin: I feel it would be a better experience if the user was able to copy and paste the command from this document and open another document for more detail, instead of having to go to another document to get the thing to copy and paste. Have we already considered this?

docs/services/gcp-services-guide.md
```diff
@@ -0,0 +1,536 @@
...
+  section](../install-crossplane.md#gcp-stack) of the install guide.
+
+We will be using `gcp-infra-dev` namespace to store our cloud-specific resource
```
suskin: Why are we using this namespace specifically? As a user, Is there a document I can be linked to that explains the concepts and the best practices behind this pattern?

docs/services/gcp-services-guide.md
```diff
@@ -0,0 +1,536 @@
...
+kubectl create namespace gcp-infra-dev
+export INFRA_NAMESPACE=gcp-infra-dev
+```
+
+#### GCP Provider
+It is essential to make sure that a GCP service account to be used by 
+Crossplane is configured with all the necessary permissions outlined in the 
+[provider guide](../cloud-providers/gcp/gcp-provider.md).
+
+Using the service account json `crossplane-gcp-provider-key.json`:
```
suskin: Where does this come from?

docs/services/gcp-services-guide.md
```diff
@@ -0,0 +1,536 @@
...
+* Generate Base64 encoded value to store in a `Secret`:
+```bash
+export BASE64ENCODED_GCP_PROVIDER_CREDS=$(base64 crossplane-gcp-provider-key.json | tr -d "\n")
```
suskin: I would recommend unsetting this when it's no longer neede, as a security best practice, so that it's harder to accidentally expose the credentials by, say, pasting environment variables somewhere when asking for help.

docs/services/gcp-services-guide.md
```diff
@@ -0,0 +1,536 @@
...
+* Verify GCP provider was successfully registered by the crossplane
+```bash
+kubectl get providers.gcp.crossplane.io -n $INFRA_NAMESPACE
+kubectl get secrets -n $INFRA_NAMESPACE
+```
+
+#### Cloud-Specific Resource Classes
+Cloud-specific resource classes are used to define a reusable configuration for
+a specific managed service. Wordpress requires a MySQL database, which can be
+satisfied by a [Google Cloud SQL Instance](https://cloud.google.com/sql/docs/mysql/) instance.
```
suskin: `Google Cloud SQL Instance instance` should be more like `Google Cloud SQL Instance`

docs/services/gcp-services-guide.md
```diff
@@ -0,0 +1,536 @@
...
+  # Note from GCP Docs: Your Cloud SQL instances are not created in your VPC network.
+  # They are created in the service producer network (a VPC network internal to Google) that is then connected (peered) to your VPC network.
+  privateNetwork: projects/$PROJECT_ID/global/networks/$NETWORK_NAME
+  providerRef:
+    name: gcp-provider
+    namespace: $INFRA_NAMESPACE
+  reclaimPolicy: Delete
+EOF
+
+kubectl apply -f gcp-mysql-standard.yaml
```
suskin: Are these yamls also in the repo in some examples folder somewhere? It may be worth telling people something like "for more details, see the full example at . . ."

docs/services/gcp-services-guide.md
```diff
@@ -0,0 +1,536 @@
...
+```
+* You can verify creation with the following command and output:
```
suskin: I like that we're including verification steps

docs/services/gcp-services-guide.md
```diff
@@ -0,0 +1,536 @@
...
+
+*Command*
```
suskin: The pattern I usually like to use is:

```
$ kubectl get cloudsqlinstanceclass -n ${INFRA_NAMESPACE}
NAME                   PROVIDER-REF   RECLAIM-POLICY   AGE
standard-cloudsql      gcp-provider   Delete           11s
```

Because that's what it looks like when people are in the terminal. Sometimes I'll put `> ` at the beginning of the stuff that's printed out if there's only one line of output, or if there's another prompt.

docs/services/gcp-services-guide.md
```diff
@@ -0,0 +1,536 @@
...
+potential configurations. For instance, you may create `large-gcp-mysql` with
+field `storageGB: 100`.
+
+#### Namespaces
+Kubernetes namespaces allow for separation of environments within your cluster.
+You may choose to use namespaces to group resources by team, application, or any
+other logical distinction. For this demo, we will create a namespace called
+`app-project1-dev`, which we will use to group our Wordpress resources.
+
+* Create a `Namespace`:
```
suskin: This line probably isn't needed; we could just do something more like this:

```
. . . which we will use to group our Wordpress resources:

\```
kubectl create . . . 
\```
```

docs/services/gcp-services-guide.md
```diff
@@ -0,0 +1,536 @@
...
+* Define a `MySQLInstanceClass` in `mysql-standard.yaml` for namespace
+  `app-project1-dev`:
+```bash
+cat > mysql-standard.yaml <<EOF
+---
+apiVersion: database.crossplane.io/v1alpha1
+kind: MySQLInstanceClass
+metadata:
+  name: mysql-standard
+  namespace: app-project1-dev
```
suskin: This is a bit of a nitpick, but I would lean toward no namespace here and putting the namespace on the `kubectl` command instead, so that people can reuse the same yaml file for a different namespace.

docs/services/gcp-services-guide.md
```diff
@@ -0,0 +1,536 @@
...
+NAME             AGE
+mysql-standard   27s
+```
+
+Once again, you are free to create more `MySQLInstanceClass` instances in this
+namespace to define more classes of service. For instance, if you created
+`mysql-gcp-large` above, you may want to create a `MySQLInstanceClass` named
+`mysql-large` that references it. You may also choose to create MySQL resource
+classes for other non-GCP providers, and reference them for a class of service
+in the `app-project1-dev` namespace.
```
suskin: As a user, is there a document I can read which will give me more details about these concepts? If so, I would like a link to it from this document.

docs/services/gcp-services-guide.md
```diff
@@ -0,0 +1,536 @@
...
+
+You may specify *one* instance of a portable class kind as *default* in each
+namespace. This means that the portable resource class instance will be applied
+to claims that do not directly reference a portable class. If we wanted to make
```
suskin: I am not sure I would understand this as a new user. Maybe it'd make sense to tell the user to run the `label` command so that we can provision our stuff, and then link them to a document which explains the concepts in more dept?

docs/services/gcp-services-guide.md
```diff
@@ -0,0 +1,536 @@
...
+
+```bash
+kubectl -n app-project1-dev label mysqlinstanceclass mysql-standard default="true"
+```
+
+#### Resource Claims
+Resource claims are used to create external resources by referencing a class of
+service in the claim's namespace. When a claim is created, Crossplane uses the
+referenced portable class to find a cloud-specific resource class to use as the
+configuration for the external resource. We need a to create a claim to
```
suskin: `We need a to` seems like it should be ` We need to`

docs/services/gcp-services-guide.md
```diff
@@ -0,0 +1,536 @@
...
+```
+
+*Output*
+```bash
+NAME          STATUS   CLASS            VERSION   AGE
+mysql-claim   Bound    mysql-standard   5.6       11m
+```
+
+If the `STATUS` is blank, we are still waiting for the claim to become bound.
+You can observe resource creation progression using the following:
```
suskin: Can we can also use `--wait` to wait for the status we want? If so, it may make sense to show the user how to do that.

docs/services/gcp-services-guide.md
```diff
@@ -0,0 +1,536 @@
...
+    Status:                True
+    Type:                  Synced
+Events:                    <none>
+```
+
+*Note: You must wait until the claim becomes bound before continuing with this
+guide. It could take a few minutes for GCP to complete CloudSQL creation.*
+
+#### Connecting MySQL Instance and GKE Cluster
+Before we install Wordpress, we need establish connectivity between our MySQL
```
suskin: `we need establish` should be `we need to establish`

docs/services/gcp-services-guide.md
```diff
@@ -0,0 +1,536 @@
...
+database and our GKE cluster. We can do this by creating a [Private Service 
+Connection](https://cloud.google.com/vpc/docs/configure-private-services-access).
+
+You can create it following the link above or use Crossplane to do it:
```
suskin: `You can create it by following the instructions at the link above, or you could use Crossplane to do it:` would be more grammatically correct

docs/services/gcp-services-guide.md
```diff
@@ -0,0 +1,536 @@
...
+EOF
+
+kubectl apply -f network.yaml
+```
+
+* You can verify creation with the following command and output:
+
+*Command*
+```bash
+kubectl -n $INFRA_NAMESPACE get connection example-connection -o custom-columns='NAME:.metadata.name,FIRST_CONDITION:.status.conditions[0].status,SECOND_CONDITION:.status.conditions[1].status'
```
suskin: What are these conditions? Are there names we can use to explain them, instead of first and second?

docs/services/gcp-services-guide.md
```diff
@@ -0,0 +1,536 @@
...
+```
+*Output*
+```
+NAME                 FIRST_CONDITION   SECOND_CONDITION
+example-connection   True              True
+```
+Wait for both conditions to be true to continue.
+
+## Install Wordpress
```
suskin: Is this part common to all providers? Are we planning on having a single high-level service guide which links out to the provider-specific documents for the provider-specific parts? The benefit would be that we wouldn't need to maintain separate copies of the same thing, and so that the user would have a single high-level guide for the instructions

docs/services/gcp-services-guide.md
```diff
@@ -0,0 +1,536 @@
...
+Installing Wordpress requires creating a Kubernetes `Deployment` and load
+balancer `Service`. We will point the deployment to the `wordpressmysql` secret
+that we specified in our claim above for the Wordpress container environment
+variables. It should have been populated with our MySQL connection details after
+the claim became `Bound`.
```
suskin: As a user, is there a document I can read to learn more about what this is talking about? Could I get a link to it from this document?

docs/services/gcp-services-guide.md
```diff
@@ -0,0 +1,536 @@
...
+Type:  Opaque
+
+Data
+====
+endpoint:  75 bytes
+password:  27 bytes
+username:  58 bytes
+```
+
+* Define the `Deployment` and `Service` in `wordpress-app.yaml`:
```
suskin: This line talks about `wordpress-app.yaml`, but the example code uses `wordpress.yaml`.

docs/services/gcp-services-guide.md
```diff
@@ -0,0 +1,536 @@
...
+deployment.apps/wordpress   1/1     1            1           11m
+
+NAME                TYPE           CLUSTER-IP    EXTERNAL-IP   PORT(S)        AGE
+service/wordpress   LoadBalancer   10.0.128.30   52.168.69.6   80:32587/TCP   11m
+```
+
+If the `EXTERNAL-IP` field of the `LoadBalancer` is `<pending>`, wait until it
+becomes available, then navigate to the address. You should see the following:
+
+![alt azure wordpress](azure-wordpress.png)
```
suskin: Is this right? I was not expecting a link to an `azure*` file from the GCP guide.

docs/services/gcp-services-guide.md
```diff
@@ -0,0 +1,536 @@
+# Deploying Wordpress to GKE
```
suskin: Structural comments:

* I would suggest that we add a table of contents to this doc, as it does have some length on it.
* I would suggest that we add a Conclusion section and Next Steps section. Conclusion would summarize what we did, and Next Steps would suggest other documents to read, or other exercises to try. See the [stack guide](https://github.com/crossplaneio/crossplane/pull/803/files#diff-378fbcada1ba53bb6c4e2afb09c18c3aR228) for some examples

docs/services/gcp-services-guide.md
```diff
@@ -0,0 +1,536 @@
...
+  --scopes "https://www.googleapis.com/auth/devstorage.read_only","https://www.googleapis.com/auth/logging.write","https://www.googleapis.com/auth/monitoring","https://www.googleapis.com/auth/servicecontrol","https://www.googleapis.com/auth/service.management.readonly","https://www.googleapis.com/auth/trace.append" \
+  --num-nodes "1" \
+  --enable-cloud-logging --enable-cloud-monitoring --enable-ip-alias --enable-autoupgrade --enable-autorepair --no-enable-basic-auth \
+  --network "projects/${PROJECT_ID}/global/networks/${NETWORK_NAME}" \
+  --subnetwork "projects/${PROJECT_ID}/regions/us-central1/subnetworks/${SUBNETWORK_NAME}" \
+  --default-max-pods-per-node "110" \
+  --addons HorizontalPodAutoscaling,HttpLoadBalancing
+```
+
+This may take a while to finish.
```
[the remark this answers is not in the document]
muvaf: gcloud cli shows the progress and finishes the execution when creation is done.

docs/services/gcp-services-guide.md
muvaf: > Does it automatically populate our kubectl config for us and switch to it?

That's exactly what it does.

docs/services/gcp-services-guide.md
hasheddan: This link will actually point to just a picture of the Wordpress starter page. We can use it for each of the services guide because it will look the same, but you are right that we should change the name. @muvaf I'll change image to `wordpress-start.png`.

docs/services/gcp-services-guide.md
muvaf: I'm not sure. Having Crossplane installation command in every guide doesn't seem like ideal to me. Maintenance of that in each guide is also a burden IMO. Quickstart could be a good candidate for this but services guide is a little bit more in-depth guide.

docs/services/gcp-services-guide.md
muvaf: See https://github.com/crossplaneio/crossplane/pull/802#discussion_r325178739

docs/services/gcp-services-guide.md
muvaf: It's about showing how one can logically group their resources via namespaces. This specific name was inspired by @prasek 's work with GitLab.

docs/services/gcp-services-guide.md
muvaf: It's from the provider guide, just added more context in the next commit.

docs/services/gcp-services-guide.md
muvaf: No, these are extracted from Wordpress workload example YAMLs. We'd need to consider whether it's worth creating specific folder for this guide.

docs/services/gcp-services-guide.md
muvaf: I actually don't like that pattern because it doesn't allow you to just select the whole line and copy&paste into the terminal, you'd have to be careful to select only the command. I think k8s docs are also prepared this way https://kubernetes.io/docs/tutorials/stateless-application/guestbook/#creating-the-redis-master-deployment

docs/services/gcp-services-guide.md
muvaf: Added a link and explanation about portable & cloud-specific resource classes.

docs/services/gcp-services-guide.md
muvaf: They have names but it's an array and no one knows the order of the conditions at the time of kubectl command execution.

docs/services/gcp-services-guide.md
muvaf: It's not specific to GCP. @prasek will we have a main doc where we refer to cloud-specific guides and link back to the main one to keep going with Wordpress or is it the case that each cloud services guide is standalone guide?

docs/services/gcp-services-guide.md
muvaf: I am not sure whether there is a doc that talks about the binding status except for @hasheddan 's podcast. I'll add more context there.

docs/services/gcp-services-guide.md
muvaf: It's a generic Wordpress welcome page image that I borrowed from https://github.com/crossplaneio/crossplane/pull/800 , not specific to azure. It should work when that PR is merged.

docs/services/gcp-services-guide.md
```diff
@@ -0,0 +1,536 @@
...
+
+This user guide will walk you through Wordpress application deployment using
+your existing Kubernetes cluster and Crossplane managed resources. We will:
+* Install Crossplane to your cluster.
+* Create necessary resource classes for MySQL database instance.
+* Create network resources to get GKE cluster to connect to MySQL instance.
+* Deploy Wordpress.
+
+
+## Pre-requisites
```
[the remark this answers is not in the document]
muvaf: I think this section is just so short already to need an explanation.

muvaf:
@suskin Thank you for your thorough review! I think it's ready to be reviewed again.

docs/services/gcp-services-guide.md
suskin: Dang

docs/services/gcp-services-guide.md
suskin: I agree with your point. Being able to copy and paste more easily is the way to go; we should separate input and output everywhere!

jbw976: +1 on this, I think it's a much better experience when you can copy the entire block to paste and run it in your terminal, and not have to be careful about which parts you're selecting.

docs/services/gcp-services-guide.md
```diff
@@ -0,0 +1,613 @@
+# Deploying Wordpress to GKE
```
prasek: @muvaf can you add this to be consistent with the crossplane.io side nav being updated for #816 

---
title: "Using GCP Services"
toc: true
weight: 420
indent: true
---

similar to https://raw.githubusercontent.com/crossplaneio/crossplane/7ecc2df467b5c0683fe257a7cc1d316b76a9445a/docs/services/azure-services-guide.md

docs/services/gcp-services-guide.md
```diff
@@ -0,0 +1,620 @@
+---
+title: "Using GCP Services"
```
negz: Can you change this to "GCP Service Guide" per @prasek.

docs/services/gcp-services-guide.md
```diff
@@ -0,0 +1,620 @@
...
+toc: true
+weight: 420
+indent: true
+---
+
+# Deploying Wordpress to GKE
```
negz: I believe this should also be "GCP Service Guide" for consistency across services guides.

docs/services/gcp-services-guide.md
muvaf: His comment suggested that actually https://github.com/crossplaneio/crossplane/pull/802/#pullrequestreview-289772546

negz: I'm pretty sure the thinking changed while you were asleep. See https://crossplane.slack.com/files/UJDEZ82R1/FN3B9Q8M8/image.png.

negz:
Disclaimer: I'm merging these changes without thoroughly reviewing them myself. They've gone through extensive review up until this point and we're taking a "merge it now, tweak it later" approach to documentation in order to ensure we have something in time for the 0.3 release.

Issue #779, opened 2019-09-12
Add Stacks Guide to getting-started.md on crossplane.io docs


prasek:
### What problem are you facing?
Would like to get started deploying a portable App Stack (e.g. Wordpress) into multiple clouds from kubectl using one of the Stacks for GCP, AWS, or Azure.

part of #625

### How could Crossplane help solve your problem?
Add a [Getting Started](https://crossplane.io/docs/v0.2/getting-started.html) section for:
 *  S3: Stacks Guide: Deploy portable Wordpress App Stack into GCP, AWS, or Azure

See the [complete v0.3 docs plan](https://docs.google.com/document/d/11SQHf1v_1iisgVqbGwdRnEOmpoAMpbiEglb_H664To4/edit?usp=sharing) for more detail

### Todo:
 - [ ] Finalize [DRAFT](https://docs.google.com/document/d/1jJncy5hZcarZU6w9YpqHeO_z1f1wx0pw9tDXS0tyUGE/edit)
 - [ ] Add an general overview page for S3: Stacks Guide: Deploy portable Wordpress App Stack into GCP, AWS, or Azure
   - [ ] AWS Stacks Guide 
   - [ ] GCP Stacks Guide
   - [ ] Azure Stacks Guide #650

suskin:
There has been a lot of work happening here, and we should be good for now. It could benefit from a bit more testing; the AWS stack guide has had the least coverage. But worst case, if we find anything critical, we can do a patch release. I am going to close this

PR #803, opened 2019-09-16
Docs: Stack guide with GCP

suskin:

Related to #779 

As part of the 0.3 release, we want to document how to use stacks for the crossplane website, and for anyone looking at the docs in the repository. This changeset adds a first pass at an overall guide document, with a first pass at the GCP-specific provider section. We'll also need Azure-specific and AWS-specific instructions, but this should at least allow us to start testing the documentation.

I am thinking that we'll review and merge this documentation, and then polish it and add the other provider guides in future changesets.

### Testing done

I have clicked around on a bunch of the links, and have been viewing the documents locally as described in [this changeset](https://github.com/crossplaneio/crossplaneio.github.io/pull/25).

More testing is needed to make sure everything works smoothly!


cluster/examples/workloads/kubernetes/wordpress/gcp/provider.yaml
```diff
@@ -4,49 +4,17 @@ data:
...
-  namespace: crossplane-system
+  namespace: gcp
   name: gcp-provider-creds
 type: Opaque
 ---
 apiVersion: gcp.crossplane.io/v1alpha2
 kind: Provider
 metadata:
-  namespace: crossplane-system
+  namespace: gcp
```
hasheddan: I believe that we are wanting to move this to being `gcp-infra-dev`

docs/stacks-guide-gcp.md
```diff
@@ -0,0 +1,424 @@
...
+Since this is an infrastructure stack, we need to specify that it's
+cluster-scoped via `--cluster` flag.
+```
+kubectl create namespace gcp
+kubectl crossplane stack generate-install --cluster 'crossplane/stack-gcp:master' stack-gcp | kubectl apply --namespace gcp -f -
+```
+
+The namespace that we install the stack to is also the one where our
+managed GCP resources will reside. When a developer requests a resource
+by creating a **resource claim** in a namespace `mynamespace`, the
```
hasheddan: I believe we will want `mynamespace` to be something like `app-project1-dev`

docs/stacks-guide-gcp.md
```diff
@@ -0,0 +1,424 @@
...
+```
+
+Now we need to tell Crossplane which resource classes should be used to
+satisfy our claims in that app namespace. We will create portable
+classes that have have reference to non-portable ones that we created
+earlier. In our claims, we can refer to those portable classes directly
+or label one as the default portable class to be used in claims that do
+not have class reference.
+
+> Portable classes are a way of referring to non-portable resource classes in other namespaces.
```
hasheddan: May need to update if we use terminology other than "non-portable" in user-facing docs.

docs/stacks-guide-gcp.md
```diff
@@ -0,0 +1,424 @@
...
+
+```
+kubectl -n gcp get connection.servicenetworking.gcp.crossplane.io/example-connection -o custom-columns='NAME:.metadata.name,FIRST_CONDITION:.status.conditions[0].status,SECOND_CONDITION:.status.conditions[1].status'
+```
+
+## Configure Provider Resources
+
+Once we have the network set up, we also need to tell Crossplane how to
+satisfy WordPress's claims for a database and a Kubernetes cluster.
+The resource classes serve as template for the new claimswe make.
```
hasheddan: "claimswe" -> "claims we"

docs/stacks-guide-gcp.md
```diff
@@ -0,0 +1,424 @@
...
+```
+kubectl -n gcp get connection.servicenetworking.gcp.crossplane.io/example-connection -o custom-columns='NAME:.metadata.name,FIRST_CONDITION:.status.conditions[0].status,SECOND_CONDITION:.status.conditions[1].status'
+```
+
+## Configure Provider Resources
+
+Once we have the network set up, we also need to tell Crossplane how to
+satisfy WordPress's claims for a database and a Kubernetes cluster.
+The resource classes serve as template for the new claimswe make. 
+The following resource classes allow the GKECluster and CloudSQL claims
```
hasheddan: I think this should be "allow the KubernetesCluster and MySQLInstance claims"

docs/stacks-guide.md
```diff
@@ -0,0 +1,301 @@
...
+Now that we've got Crossplane set up and configured to use a cloud
+provider, we're ready to add support for creating WordPresses! We'll do
+this using a Crossplane Stack. For more information about stacks, see
+the [full Stack documentation][stack-docs].
+
+We can use the [Crossplane CLI][crossplane-cli] to install our stack which adds support for
+Wordpress:
+
+```
+kubectl crossplane stack install 'crossplane/sample-stack-wordpress:latest' 'sample-stack-wordpress'
```
hasheddan: We will need to make sure that the wordpress stack is installed in the same `namespace` as the portable classes created in each individual stack guide. We could either standardize on the namespace or just mention that it needs to be the same here. I used this to install wordpress stack into the namespace of my portable classes:

```
kubectl crossplane stack generate-install  'crossplane/sample-stack-wordpress:latest' 'sample-stack-wordpress' | kubectl apply --namespace app-project1-dev -f -
```

docs/stacks-guide.md
```diff
@@ -0,0 +1,301 @@
...
+Now that Crossplane supports Wordpress creation, we can ask Crossplane
+to spin up a Wordpress for us. We can do this by creating a Kubernetes
+resource that our Wordpress stack will recognize:
+
+```
+cat > my-wordpress.yaml <<EOF
+apiVersion: wordpress.samples.stacks.crossplane.io/v1alpha1
+kind: WordpressInstance
+metadata:
+  name: my-wordpressinstance
```
hasheddan: The `WordpressInstance` will also need to be in the correct namespace. I used this:

```
cat > my-wordpress.yaml <<EOF
apiVersion: wordpress.samples.stacks.crossplane.io/v1alpha1
kind: WordpressInstance
metadata:
  name: my-wordpressinstance
  namespace: app-project1-dev
EOF

kubectl apply -f my-wordpress.yaml
```

docs/stacks-guide.md
```diff
@@ -0,0 +1,301 @@
...
+### Wait
+
+The Wordpress can take a while to spin up, because behind the scenes
+Crossplane is creating all of its dependendencies, which is a database
+and Kubernetes cluster. To check the status, we can look at the
+resources that Crossplane is creating for us:
+
+```
+# The claim for the database
+kubectl get mysqlinstance
```
hasheddan: Might be useful to include namespaces on these as well if we standardize.

muvaf: @suskin we need to change the namespace `mynamespace` to `app-project1-dev` in the gcp guide as well if we want to standardize on that.

docs/stacks-guide.md
```diff
@@ -0,0 +1,301 @@
...
+### Use
+
+Once everything has been created, the ip address for the Wordpress
+instance will show up in the [Crossplane
+KubernetesApplicationResource][kubernetesapplicationresource-docs]
+which represents the workload's service. Here's a way to watch for the
+ip:
+
+```
+kubectl get kubernetesapplicationresource -n wordpresses -o custom-columns='NAME:.metadata.name,NAMESPACE:.spec.template.metadata.namespace,KIND:.spec.template.kind,SERVICE-EXTERNAL-IP:.status.remote.loadBalancer.ingress[0].ip' --watch
```
muvaf: `-n wordpresses` won't work

docs/stacks-guide.md
muvaf: @hasheddan @suskin we can just omit the namespace for portable classes as well actually. Use whatever is set on kubectl for wordpress stack, wordpress instance and portable classes

hasheddan: @muvaf definitely true. However, I think we are trying to standardize on these namespace patterns, and it is more explicit IMO. I could be persuaded though... cc @prasek

docs/stacks-guide.md
suskin: Should be fixed

docs/stacks-guide.md
suskin: This should be fixed. I went with the explicit approach so readers can see an example of how to specify a namespace they want.

docs/stacks-guide.md
suskin: Should be fixed; I ended up using a slightly inconsistent way of specifying it, but we can always fix that in a future changeset.

docs/stacks-guide.md
suskin: This should be fixed now

docs/stacks-guide-gcp.md
suskin: Should be fixed; I went even more general and didn't us the kind names

docs/stacks-guide-gcp.md
suskin: Should be fixed

cluster/examples/workloads/kubernetes/wordpress/gcp/provider.yaml
suskin: Let's keep this note and roll it into the next changeset

docs/stacks-guide-gcp.md
suskin: Should be fixed

docs/stacks-guide-gcp.md
suskin: Good point. Let's keep track of this and consider it for a future changeset

PR #804, opened 2019-09-17
stacks: allow icon files to be specified per resource

jbw976:
This PR adds support for a stack to specify an icon file for all CRDs in the stack.

The stacks design doc is also updated to be in alignment with the implementation.

An example of per resource metadata can be found in https://github.com/crossplaneio/stack-gcp/pull/28

This PR was tested with the full GCP stack and the additional per resource metadata shown in https://github.com/crossplaneio/stack-gcp/pull/28.  All CRDs were successfully installed and the GCP stack controller started up OK.  A runtime example of the installed and fully annotated GKE Cluster CRD can be found in this gist: https://gist.github.com/jbw976/6c8197dddcb14a0d172a4ad52aaad176

### Checklist

I have:


pkg/stacks/unpack.go
```diff
@@ -44,14 +44,19 @@ const (
...
-	crdFileNamePattern      = "*crd.yaml"
-	uiSpecFileNamePattern   = "*ui-schema.yaml"
-	uiSpecFileName          = "ui-schema.yaml" // global ui schema filename
-	yamlSeparator           = "\n---\n"
+
+	// iconFileNamePattern is the pattern used when walking the stack package and looking for icon files.
+	// Icon files that are for a single resource can be prefixed with the kind of the resource, e.g.,
+	// mytype.icon.svg
+	// Multiple icon types and therefore file extensions are supported: svg, png, jpg, gif
+	iconFileNamePattern = "*icon.*"
```
suskin: This seems potentially a little **too** fuzzy in the long run . . . I'm not thinking of a *good* example, but what if I have a `decepticon.crd.yaml`?

design/design-doc-stacks.md
```diff
@@ -238,10 +238,11 @@ In this example, the directory names "databases.foocompany.io", "mysql", "v1alph
...
+    * Multiple `*resource.yaml` files may exist alongside multiple CRD files in the same directory.  The `resource.yaml` files should include an `id:` referencing the `Kind` of their matching CRD.
+    * Multiple `*icon.svg` files may be included in the same directory as the CRD yaml files they modify. In this case the filename should be prefixed to match the `Kind` of the associated CRD (`mytype.icon.svg`).
   * `*ui-schema.yaml`: UI metadata that will be transformed and annotated according to the [Stack UI Metadata One Pager](one-pager-stack-ui-metadata.md)
-  Multiple `ui-schema.yaml` files may be included in the same directory as the CRD yaml files they modify. In this case the filename should be prefixed to match the `Kind` of the associated CRD (`mytype.ui-schame.yaml`).
+    * Multiple `*ui-schema.yaml` files may be included in the same directory as the CRD yaml files they modify. In this case the filename should be prefixed to match the `Kind` of the associated CRD (`mytype.ui-schema.yaml`).
```
suskin: Nice fix! It would have been a `schame` to not have caught this before releasing!

pkg/stacks/unpack.go
```diff
@@ -576,5 +595,30 @@ func (sp *StackPackage) applyUISpecAnnotations(crdPath string, crd *apiextension
 			}
 		}
 	}
+}
+
+// isMetadataApplicableToCRD determines if the given metadata file path is applicable to the given CRD.
+func isMetadataApplicableToCRD(crdPath, metadataPath string, globalFileNames []string, crdKind string) bool {
+	// compare the directory of the given metadata file path to the CRDs path
+	metadataDir := filepath.Dir(metadataPath)
```
suskin: Nitpick: I am thinking we may want to make a note for future maintainers that this implementation assumes that there is at least one level of directories above the path, or that the paths are all absolute (because `filepath.Dir` [can return](https://golang.org/pkg/path/filepath/#Dir) `"."` otherwise)

pkg/stacks/unpack.go
jbw976: Good point, thank you for bringing up that edge case.  Sounds like in the long term we'd want to use more powerful regular expressions instead of the more simple `filepath.Match` we are using right now.

jbw976:
I updated the PR description to include the testing I performed @suskin, thank you for the review!

pkg/stacks/unpack.go
jbw976: I think this comment actually applies at a higher level and not just this function.  I think it applies to the entire unpack processing.  The way the Stack Manager invokes this unpacking will always result in a directory structure with a root, so we shouldn't run into any problems in practical usage, but it's probably worth adding a top level comment.

jbw976: Pushed comments in all applicable places that these paths should be absolute (including the applicable main.go argument).

Issue #650, opened 2019-08-05
Azure: Update examples and README files containing Wordpress examples


hasheddan:
Update `cluster/examples` and `docs/` with portable Wordpress on Azure.

Part of #617
Depends on #572, #647, #648, #649 
### What seems to be the problem?

Examples and documentation do not exist for deploying the portable Wordpress application stack [#572] on Azure.

### What could Crossplane do to help?

Examples and documentation should be updated to contain necessary `YAML` files and directions for deploying the portable Wordpress application stack on Azure.

### What does it look like when we're done?
A user can:

* Find all resource classes, network resources, and defaulting policies needed to allow the portable Wordpress stack to be provisioned on Azure.
* Find documentation with instructions on how to setup Azure and install the Wordpress app stack.

### Todo
- [ ] Azure portable Wordpress examples created.
- [ ] Azure portable Wordpress docs created.

displague:
I had problems with my Azure account and left this PR behind.  It may be a starting point for this work.
https://github.com/crossplaneio/crossplane/pull/574

hasheddan:
Awesome, thanks @displague!

displague:
The issue would be a parent to https://github.com/crossplaneio/crossplane/issues/538 and would close https://github.com/crossplaneio/crossplane/issues/519 (so long as the GCP and AWS examples and docs haven't drifted away).

jbw976:
There may be some work in #574 that we could reuse for the Azure docs update, but potentially not if our docs overhaul for v0.3 is fairly drastic.

jbw976:
We have documentation for installing Wordpress in Azure via the services guide and the stacks guide:
* https://crossplane.io/docs/v0.3/services/azure-services-guide.html
* https://crossplane.io/docs/v0.3/stacks-guide-azure.html

Users should be able to copy and paste from those guides to end up with WordPress running in Azure.

So I think what's still missing from this issue is the `cluster/examples` yaml files, which we have for GCP and AWS, but not for Azure:
https://github.com/crossplaneio/crossplane/tree/master/cluster/examples/workloads/kubernetes/wordpress

hasheddan:
@jbw976 I should be able to get this done and close out the issue today!

PR #805, opened 2019-09-17
azure stacks guide

hasheddan:

Azure component of stacks guide.

Partial fix for #650 

### Checklist

I have:

### DIFF
--- a/docs/stacks-guide-azure.md
+++ b/docs/stacks-guide-azure.md
@@ -1,5 +1,8 @@
 ---
-title: "Crossplane Stacks Guide: Azure Setup" toc: true weight: 332 indent: true
+title: "Crossplane Stacks Guide: Azure Setup"
+toc: true
+weight: 431
+indent: true
 ---
 
 # Crossplane Stacks Guide: Azure Setup
--- a/docs/stacks-guide-azure.md
+++ b/docs/stacks-guide-azure.md
@@ -1,0 +1,429 @@
+---
+title: "Crossplane Stacks Guide: Azure Setup" toc: true weight: 332 indent: true
+---
+
+# Crossplane Stacks Guide: Azure Setup
+
+## Table of Contents
+
+1. [Introduction](#introduction)
+2. [Install the Azure Stack](#install-the-azure-stack)
+3. [Configure Azure Account](#configure-azure-account)
+4. [Configure Crossplane Azure Provider](#configure-crossplane-azure-provider)
+5. [Set Up Network Resources](#set-up-network-resources)
+6. [Configure Provider Resources](#configure-provider-resources)
+7. [Recap](#recap)
+8. [Next Steps](#next-steps)
+
+## Introduction
+
+In this guide, we will set up a Azure provider in Crossplane so that we can
+install and use the WordPress sample stack, which depends on MySQL and
+Kubernetes!
+
+Before you begin, you will need:
+
+* Everything from the [Crossplane Stacks Guide][stacks-guide] before the cloud
+  provider setup
+  - A `kubectl` pointing to a Crossplane control cluster
+  - The [Crossplane CLI][crossplane-cli] installed
+* An account on [Azure][azure]
+
+At the end, we will have:
+
+* A Crossplane control cluster configured to use Azure
+* The boilerplate of a Azure-based project spun up
+* Support in the control cluster for managing MySQL and Kubernetes cluster
+  dependencies
+* A slightly better understanding of:
+  - The way cloud providers are configured in Crossplane
+  - The way dependencies for cloud-portable workloads are configured in
+    Crossplane
+
+We will **not** be teaching first principles in depth. Check out the [concepts
+document][crossplane-concepts] for that.
+
+## Install the Azure Stack
+
+After Crossplane has been installed, it can be extended with more functionality
+by installing a Crossplane Stack! Let's install the stack for Microsoft Azure to
+add support for that cloud provider. We can use the Crossplane CLI for this
+operation:
+
+```
+kubectl crossplane stack install 'crossplane/stack-azure:master' stack-azure
+```
+To install to a particular namespace, you can use the `generate-install` command
+and pipe it to `kubectl apply` instead, which gives you more control over how
+the stack's installation is handled. Everything is Kubernetes object!
+
+Since this is an infrastructure stack, we need to specify that it's
+cluster-scoped via `--cluster` flag.
+```
+kubectl crossplane stack generate-install --cluster 'crossplane/stack-azure:master' stack-azure | kubectl apply --namespace crossplane-system -f -
+```
+
+We have installed the Azure stack into the `crossplane-system` namespace, but we
+want to group our Azure-specific resources in their own environment namespaces.
+For the purpose of this guide, we will create all Azure-specific resources in
+the `azure-infra-dev` namespace, and all application-specific resources in the
+`app-project1-dev` namespace. Let's create these namespaces before we get
+started:
+
+```
+kubectl create namespace azure-infra-dev
+kubectl create namespace app-project1-dev
+```
+
+## Configure Azure Account
+
+We will make use of the following services on Azure:
+
+*   Resource Group
+*   AKS
+*   Azure Database for MySQL
+*   Virtual Network
+*   Subnetwork
+*   Virtual Network Rule
+
+In order to utilize each of these services, you will need to follow the Adding
+Microsoft Azure to Crossplane [guide](cloud-providers/azure/azure-provider.md)
+to obtain appropriate credentials in a JSON file referred to as
+`crossplane-azure-provider-key.json`.
+
+## Configure Crossplane Azure Provider 
+
+Before creating any resources, we need to create and configure a cloud provider
+in Crossplane. This helps Crossplane know how to connect to the cloud provider.
+All the requests from Crossplane to Azure can use that resource as their
+credentials. The following command assumes that you have a
+`crossplane-azure-provider-key.json` file that belongs to the account you’d like
+Crossplane to use.
+
+```
+export BASE64ENCODED_AZURE_PROVIDER_CREDS=$(base64 crossplane-azure-provider-key.json | tr -d "\n")
+```
+
+Now we’ll create our `Secret` that contains the credential and `Provider`
+resource that refers to that secret:
+
+```
+cat > provider.yaml <<EOF
+---
+# Azure Admin service account secret - used by Azure Provider
+apiVersion: v1
+kind: Secret
+metadata:
+  name: demo-provider-azure-dev
+  namespace: azure-infra-dev
+type: Opaque
+data:
+  credentials: $BASE64ENCODED_AZURE_PROVIDER_CREDS
+---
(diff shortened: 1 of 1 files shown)


docs/stacks-guide-azure.md
```diff
@@ -0,0 +1,429 @@
...
+*   AKS
+*   Azure Database for MySQL
+*   Virtual Network
+*   Subnetwork
+*   Virtual Network Rule
+
+In order to utilize each of these services, you will need to follow the Adding
+Microsoft Azure to Crossplane [guide](cloud-providers/azure/azure-provider.md)
+to obtain appropriate credentials in a JSON file referred to as
+`crossplane-azure-provider-key.json`.
```
suskin: Hmm, it'd be great if we could include the instructions here so the user doesn't have to open up another document! They seem like they might be able to fit well.

docs/stacks-guide-azure.md
```diff
@@ -0,0 +1,429 @@
...
+The name of the `Provider` resource in the file above is `demo-azure`; we'll use
+the name `demo-azure` to refer to this provider when we configure and set up
+other Crossplane resources.
+
+We also will need to use our Azure subscription id to provision some of the
+resources we will need. You can set an environment variable so that you will
+have access when creating resources that require it:
+
+```bash
+export SUBSCRIPTION_ID=$(cat crossplane-azure-provider-key.json | jq -j '.subscriptionId')
```
suskin: I believe this is the first time `jq` is mentioned; I am thinking we should add a prerequisites section that calls it out as a dependency, or try to do this in a way that doesn't use `jq`.

docs/stacks-guide-azure.md
```diff
@@ -0,0 +1,429 @@
...
+  reclaimPolicy: Delete
+EOF
+
+kubectl apply -f network.yaml
+```
+
+For more details about networking and what happens when you run this command,
+see [this document with more details][crossplane-azure-networking-docs].
+
+It should not take too long for these resources to provision. You can monitor
```
suskin: As a user, how do I know when they're provisioned?

docs/stacks-guide-azure.md
```diff
@@ -0,0 +1,429 @@
...
+```
+
+For more details about what is happening behind the scenes, read more about
+[portable claims in Crossplane][portable-claims].
+
+## Configure Network Connection
+
+After the Wordpress stack is installed, we will need the AKS Cluster it
+provisions to be able to communicate with the MySQL database it provisions. In
+Azure, we can do so using a [Virtual Network Rule][vnet-rule]. However, the rule
```
suskin: Nitpick: I have been using a pattern of prefixing provider link reference names with their provider, so I would use something like `azure-vnet-rule` here. The benefit is it makes it easy to tell which ones are provider links, even if you don't know much about their concepts.

docs/stacks-guide-azure.md
```diff
@@ -0,0 +1,429 @@
...
+  resourceGroupName: demo-rg
+  properties:
+    virtualNetworkSubnetId: /subscriptions/${SUBSCRIPTION_ID}/resourceGroups/demo-rg/providers/Microsoft.Network/virtualNetworks/demo-vnet/subnets/demo-subnet
+  providerRef:
+    name: demo-azure
+    namespace: azure-infra-dev
+  reclaimPolicy: Delete
+EOF
+
+cat <<EOF > vnetwatch.sh
```
suskin: Nitpick: we have been putting the output redirect before the input redirect, so I would do it in that order here too for consistency. I prefer putting the output redirect before the input redirect in general because I find it to be more readable to not have things after the `EOF`.

docs/stacks-guide-azure.md
```diff
@@ -0,0 +1,429 @@
...
+#!/bin/bash
+
+set -e
+
+echo -n "waiting..."
+while kubectl -n azure-infra-dev get mysqlservers -o yaml | grep -q  'items: \[\]'; do
```
suskin: If we have a `while` loop in the script, I am thinking we may want to add a `trap 'exit 1' SIGINT` at the top to make sure the loop is interrupted reliably.

docs/stacks-guide-azure.md
```diff
@@ -0,0 +1,429 @@
...
+  properties:
+    virtualNetworkSubnetId: /subscriptions/${SUBSCRIPTION_ID}/resourceGroups/demo-rg/providers/Microsoft.Network/virtualNetworks/demo-vnet/subnets/demo-subnet
+  providerRef:
+    name: demo-azure
+    namespace: azure-infra-dev
+  reclaimPolicy: Delete
+EOF
+
+cat <<EOF > vnetwatch.sh
+#!/bin/bash
```
suskin: Bash style: It's a nitpick in this case, but I would suggest that we use `#!/usr/bin/env bash` instead, just on general principle, so that we are making fewer assumptions about the user's environment. The `/usr/bin/env` form runs `bash` from wherever it lives in the environment, instead of assuming that it lives in a specific path.

docs/stacks-guide-azure.md
```diff
@@ -0,0 +1,429 @@
...
+
+set -e
+
+echo -n "waiting..."
```
suskin: Bash style: any messages other than the final result should be redirected to stderr: 

```
echo -n "waiting..." >&2
```

The reason is that a user may want to pipe output to another command. Everything on stdout goes through the pipe, and stderr does not. In this case, this message does not seem like something that someone would want to pipe to another command. It also makes progress messages easier to squelch if they go to stderr.

docs/stacks-guide-azure.md
```diff
@@ -0,0 +1,429 @@
...
+while kubectl -n azure-infra-dev get mysqlservers -o yaml | grep -q  'items: \[\]'; do
+  echo -n "."
```
suskin: This log message should also be sent to stderr

docs/stacks-guide-azure.md
```diff
@@ -0,0 +1,429 @@
...
+  sleep 5
+done
+
+export MYSQL_NAME=$(kubectl -n azure-infra-dev get mysqlservers -o json | jq -j '.items[0].status.providerID')
+
+sed "s/MYSQL_NAME/$MYSQL_NAME/g" vnet-rule.yaml | kubectl create -f -
+
+EOF
+
+chmod +x vnetwatch.sh && ./vnetwatch.sh
```
suskin: Yessss

docs/stacks-guide-azure.md
```diff
@@ -0,0 +1,429 @@
...
+
+set -e
+
+echo -n "waiting..."
+while kubectl -n azure-infra-dev get mysqlservers -o yaml | grep -q  'items: \[\]'; do
+  echo -n "."
+  sleep 5
+done
+
+export MYSQL_NAME=$(kubectl -n azure-infra-dev get mysqlservers -o json | jq -j '.items[0].status.providerID')
```
suskin: Same comment about `jq` here. In this case, we could probably use some custom column options in the `kubectl` output, combined with `awk` to select the field.

docs/stacks-guide-azure.md
suskin: I am thinking it would be useful if this log message also indicates what it is waiting for, so that it is obvious to any readers and users

docs/stacks-guide-azure.md
```diff
@@ -0,0 +1,429 @@
...
+  and KubernetesClusterClass that refer to our AKSClusterClass and
+  SQLServerClass.
+
+## Next Steps
+
+Next we'll set up a Crossplane App Stack and use it! Head [back over to the
+Stacks Guide document][stacks-guide-continue] so we can pick up where we left
+off.
+
+## TODO
```
suskin: I am thinking we can safely remove the todo items below and this section header

docs/stacks-guide-azure.md
```diff
@@ -0,0 +1,429 @@
...
+This should not go in the final document, but is here for tracking.
+
+* Add references
+* Add next steps, with link to WordPress-specific stuff
+
+<!-- Links -->
```
suskin: It looks like the line breaks for some of the links got weird; each link should be on its own line.

docs/stacks-guide-azure.md
```diff
@@ -0,0 +1,429 @@
...
+[crossplane-cli]: https://github.com/crossplaneio/crossplane-cli
+[crossplane-azure-networking-docs]: TODO [stacks-guide]: stacks-guide.html
+
+[crossplane-concepts]: TODO [portable-claims]: TODO
+
+[azure]: https://azure.microsoft.com [vnet-rule]:
+https://docs.microsoft.com/en-us/azure/mysql/concepts-data-access-and-security-vnet
+
+[stacks-guide-continue]:
```
suskin: I am not sure what happens if the link is split across two lines, so I had been leaving these as one long line. Have you tried this one out yet?

docs/stacks-guide-azure.md
hasheddan: This is a somewhat lengthy process that I believe may add clutter this guide if included.

docs/stacks-guide-azure.md
hasheddan: This actually has to be this way for writing the bash script or else it will just execute the commands directly.

docs/stacks-guide-azure.md
hasheddan: Yep! Accidentally used the line wrap vs code extension on them :/

docs/stacks-guide-azure.md
hasheddan: I think they work but I prefer one long line as you mentioned

docs/stacks-guide-azure.md
hasheddan: I am going to be testing this guide today and will see if I can implement a workaround for this and the later use of `jq`

docs/stacks-guide-azure.md
hasheddan: jk fixed this

docs/stacks-guide-azure.md
```diff
@@ -0,0 +1,426 @@
+---
+title: "Crossplane Stacks Guide: Azure Setup" toc: true weight: 332 indent: true
```
soorena776: The `Rewrap` seems to have joined these lines

hasheddan: Good catch! Thanks :)

docs/stacks-guide-azure.md
```diff
@@ -0,0 +1,429 @@
...
+
+We will **not** be teaching first principles in depth. Check out the [concepts
+document][crossplane-concepts] for that.
+
+## Install the Azure Stack
+
+After Crossplane has been installed, it can be extended with more functionality
+by installing a Crossplane Stack! Let's install the stack for Microsoft Azure to
+add support for that cloud provider. We can use the Crossplane CLI for this
+operation:
```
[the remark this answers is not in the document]
suskin: @hasheddan these changes have been made, and are merged into master in the gcp stacks guide: https://github.com/crossplaneio/crossplane/blob/master/docs/stacks-guide-gcp.md

PR #806, opened 2019-09-17
Add API reference documentation

### LINKED ISSUE #763: Generate Crossplane (Custom Resource) API documentation

Currently Crossplane relies on a directory of hand curated example YAML to document what custom resources exist and what their fields are. This is suboptimal for a number of reasons:

* It's prone to becoming stale and inaccurate as the codebase changes.
* It's typically not possible to write a working example that uses all resource fields, meaning many of our supported managed resource and resource class parameters are undocumented.
* Examples typically do not explain what each field actually configures, i.e. examples will show you can configure `.spec.numCores`, but does not explain what `numCores` means.
* It's a lot of work for Crossplane contributors to manually update all examples when making wide ranging changes to the codebase such as introducing strongly typed resource classes.

It makes sense for Crossplane to publish some examples to show how our various concepts fit together, but once a potential Crossplane user moves past examples and wants to write their own classes and claims they're forced to go read the Crossplane source code.

https://github.com/ahmetb/gen-crd-api-reference-docs/

The above tool is used by a small handful of Kubernetes ecosystem projects (Knative, Kubeflow, etc) to automatically generate API documentation from their Go API definitions. Note in this context "API" means the custom resources they add to the Kubernetes API server, not their developer facing Go APIs. I recommend Crossplane:

* Use this tool (or one like it) to generate API documentation for each release.
* Scale back our use of hand-maintained examples to important concepts.

negz:

Fixes #763

This PR adds Crossplane API reference documentation, as well as an updated index of what Crossplane API groups exist and their status (hint: everything is alpha). The API reference (everything under `docs/api`) is generated from our GoDoc with https://github.com/negz/crossdocs. Crossdocs is under my personal org because it's a giant hack. We should be generating this documentation as part of our build process, and should not be keeping it all here under the core Crossplane repo. We don't have time to build that automation in time for v0.3.

When reviewing __please focus on whether the API documentation renders correctly, and whether the links work__. This PR is not the place to give feedback on the actual content - it's generated from our GoDoc comments. __If you spot something that really should not make it into the v0.3 docs please raise a PR to fix the GoDoc and I'll regenerate this documentation__. `api.md` and `README.md` are the exceptions to this rule - they're hand-written.

### Checklist

I have:


negz:
It's worth noting that because this is not automated we'll need to remember to re-run crossdocs immediately before we release v0.3 to make sure we don't miss any last minute changes to our API types.

negz:
See https://github.com/crossplaneio/crossplane/issues/763#issuecomment-532077791 for examples of this documentation rendered on https://crossplane.io.

PR #812, opened 2019-09-18
stacks: fixes for unpacking stacks

jbw976:

This commit ensures that the unpacking process starts with a base
dir that includes the .registry folder.  This matters now since
we look for top level icons only in the base dir, so knowing the
right base dir is important.

The troubleshooting ability of the stack manager is also increased
by adding the path to error messages when we fail to parse a
resource.

Finally, we also use filepath.Clean() to ensure that the path being
passed in as the base dir is always clean and normalized.

### Checklist

I have:

### DIFF
--- a/cmd/crossplane/main.go
+++ b/cmd/crossplane/main.go
@@ -109,7 +109,7 @@ func main() {
 
 		// TODO(displague) afero.NewBasePathFs could avoid the need to track Base
 		fs := afero.NewOsFs()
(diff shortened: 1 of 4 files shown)


jbw976:
I have also tested this manually in minikube with the AWS, GCP, and Azure stacks.  They all install OK, all CRDs are installed, their controllers are running OK, and the Stack records have their icons (except for AWS since we don't have an icon in that stack yet)

PR #814, opened 2019-09-18
concept.md updates

prasek:
Updated concepts.md 
 - updated concepts
 - added feature areas
 - added glossary

### Checklist
I have:

[skip ci]

### DIFF
--- a/docs/concepts.md
+++ b/docs/concepts.md
@@ -3,17 +3,19 @@ title: Concepts
 toc: true
 weight: 1510
 ---
+# Table of Contents
+1. [Concepts](#concepts)
+2. [Feature Areas](#feature-areas)
+3. [Glossary](#glossary)
+
 # Concepts
-
 ## Control Plane
-
 Crossplane is an open source multicloud control plane that consists of smart controllers that can work across clouds to enable workload portability, provisioning and full-lifecycle management of infrastructure across a wide range of providers, vendors, regions, and offerings.
 The control plane presents a declarative management style API that covers a wide range of portable abstractions that facilitate these goals across disparate environments, clusters, regions, and clouds.
 Crossplane can be thought of as a higher-order orchestrator across cloud providers.
 For convenience, Crossplane can run directly on-top of an existing Kubernetes cluster without requiring any changes, even though Crossplane does not necessarily schedule or run any containers on the host cluster.
 
 ## Resources and Workloads
-
 In Crossplane, a *resource* represents an external piece of infrastructure ranging from low level services like clusters and servers, to higher level infrastructure like databases, message queues, buckets, and more.
 Resources are represented as persistent object within the crossplane, and they typically manage one or more pieces of external infrastructure within a cloud provider or cloud offering.
 Resources can also represent local or in-cluster services.
@@ -26,9 +28,7 @@ For example, a container workload can include a set of objects that will be depl
 A serverless workload could include a function that will run on a serverless managed service.
 Workloads can contain requirements for where and how the workload can run, including regions, providers, affinity, cost, and others that the scheduler can use when assigning the workload.
 
-
 ## Resource Claims and Resource Classes
-
 To support workload portability we expose the concept of a resource claim and a resource class.
 A resource claim is a persistent object that captures the desired configuration of a resource from the perspective of a workload or application.
 Its configuration is cloud-provider and cloud-offering independent and it’s free of implementation and/or environmental details.
@@ -37,7 +37,6 @@ A resource claim can be thought of as a request for an actual resource and is ty
 A resource class is configuration that contains implementation details specific to a certain environment or deployment, and policies related to a kind of resource.
 A ResourceClass acts as a template with implementation details and policy for resources that will be dynamically provisioned by the workload at deployment time.
 A resource class is typically created by an admin or infrastructure owner.
-
 
 ## Dynamic and Static Provisioning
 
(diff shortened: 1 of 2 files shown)


docs/concepts.md
```diff
@@ -50,9 +53,325 @@ Dynamic provisioning is when an resource claim does not find a matching resource
...
+
+## Kubernetes
+Crossplane is built on the Kubernetes API machinery as a platform for declarative management,
+and common terminology used from the [Kubernetes Glossary][kubernetes-glossary] where possible,
+and we don't seek to reproduce that glossary here.
+
+[kubernetes-glossary]: https://kubernetes.io/docs/reference/glossary/?all=true
+However we'll summarize some commonly used concepts for convenience.
+
+### CRD
```
negz: Should we use CRD or Custom Resource Definition as the heading? The former is probably more recognisable while the latter is probably more correct.

docs/concepts.md
```diff
@@ -50,9 +53,325 @@ Dynamic provisioning is when an resource claim does not find a matching resource
...
+A standard Kubernetes Custom Controller, providing active control loops that own or more CRDs. 
+Can be implemented in different ways, such as 
+golang code (controller-runtime), templates, functions/hooks, templates, a new DSL, etc. 
+The implementation itself is versioned using semantic versioning (e.g., v1.0.4)
+
+### Namespace
+Allows logical grouping of resources in Kubernetes that can be secured with RBAC rules.
+
+## Crossplane
+### External CRD
```
negz: I haven't seen this term used before. Is it a Stacks thing?

docs/concepts.md
```diff
@@ -50,9 +53,325 @@ Dynamic provisioning is when an resource claim does not find a matching resource
...
+Usually represent an external resource, and cannot be broken down any further (leaves)
+
+aka Atomic CRD
+
+### Composite CRD
+A resource that captures parent/child relationships. They have a selector that can help query/find children resources.
+
+### Claim CRD
+An abstract resource that binds to managed resources, like a MySQLInstance..
```
negz: This sounds like the same thing as a resource claim. I suggest we just say "Resource Claim".

docs/concepts.md
```diff
@@ -50,9 +53,325 @@ Dynamic provisioning is when an resource claim does not find a matching resource
...
+
+### App CRD
+An abstract resource that manages an application instance, like a WordpressInstance.
```
negz: Here and in other cases we seem to be using "CRD" as an _instance of the resource_ rather than as an _instance of the definition of the resource_. Anything of `kind: MySQLInstance` is a (custom) resource, not a CRD.

docs/concepts.md
```diff
@@ -50,9 +53,325 @@ Dynamic provisioning is when an resource claim does not find a matching resource
...
+
+### Stack Package Format
+The package format for Stacks that contains the Stack definition, metadata, icons, CRDs, and other Stack specific files.  
+
+### Stack Manager
+The component that is responsible for installing a Stack’s custom controllers and resources in Crossplane. It can download packages, resolve dependencies, install resources and execute controllers.  This component is also responsible for managing the complete life-cycle of Stacks, including upgrading them as new versions become available.
+
+### External Resource
+An actual resource that exists outside Kubernetes, typically in the cloud. 
+AWS RDS and GCP Cloud Memorystore instances are external resources.
```
negz: I've been trying to say something like "in an external system, often a cloud provider" in an attempt to suggest that external resources could be on prem, while also providing some slightly less abstract/hypothetical context.

docs/concepts.md
```diff
@@ -51,8 +58,325 @@ The newly provisioned resource is automatically bound to the resource claim.
...
+### Networking Services
+Support for Networks, Subnets, and Firewall Rules
+
+### Load Balancing Services
+Native Kubernetes load balancing is supported. 
+
+External load balancers may be supported in the future.
+
+### Cloud DNS Services
+Kubernetes external-dns is supported.
```
negz: You can (and probably should) run external-dns to automatically create DNS entries for your Kubernetes Services, but that's tangential/complimentary to Crossplane rather than something we empower in any way.

The spirit of Crossplane support for External DNS services would be more like allowing you to declare zones and records as managed resources, which is something we probably want to do some day.

docs/concepts.md
```diff
@@ -51,8 +58,325 @@ The newly provisioned resource is automatically bound to the resource claim.
...
+### Claimant
+The Kubernetes representation of a process wishing
+to connect to a managed resource, typically a `Pod` or some abstraction
+thereupon such as a `Deployment` or `KubernetesApplication`.
+
+### Consumer
+See claimant.
+
+### Claim Kind
+The specific kind of claim, for example a MySQLInstance claim.
```
negz: Perhaps we should just include the concept of a Kind instead, or assume the reader is familiar with Kubernetes terminology? It's not not useful, but it's also relevant to managed resource kinds, resource class kinds, etc, etc.

docs/concepts.md
```diff
@@ -51,8 +58,325 @@ The newly provisioned resource is automatically bound to the resource claim.
...
+
+### Kubernetes Application
+* aka Complex Workload
```
negz: Technically Kubernetes Application is a kind of complex workload. We named it such to allow space for e.g. `kubernetesapplication.workload.crossplane.io`, `vmapplication.workload.crossplane.io`, `function.workload.crossplane.io`, etc.

docs/concepts.md
```diff
@@ -51,8 +58,325 @@ The newly provisioned resource is automatically bound to the resource claim.
...
+
+### In-Tree
```
negz: FWIW I notice myself and others saying "core Crossplane" or "Crossplane core" for this.

docs/concepts.md
```diff
@@ -51,8 +58,325 @@ The newly provisioned resource is automatically bound to the resource claim.
...
+application developer needing to know any of the details, which promotes portability and reusability.
+
+### Storage Services
+Currently supported including S3, Buckets, and Azure Blob storage. 
+
+### Networking Services
+Support for Networks, Subnets, and Firewall Rules
+
+### Load Balancing Services
+Native Kubernetes load balancing is supported.
```
negz: This is a similar grey area to Cloud DNS Services below. IMO if we don't have a managed resource for it we don't "support" it in the Crossplane Services sense so much as "exist in an ecosystem that supports it".

docs/concepts.md
prasek: I just want to #crd

prasek: but happy to swap them around if y'all want to #custom-resource-definition

docs/concepts.md
prasek: gone, was from stacks design doc, can add it later if needed

docs/concepts.md
prasek: also gone

docs/concepts.md
prasek: removed

docs/concepts.md
prasek: I'll remove this for now

docs/concepts.md
prasek: fixed

docs/concepts.md
prasek: fixed

docs/concepts.md
prasek: fixed

docs/concepts.md
```diff
@@ -51,8 +50,324 @@ The newly provisioned resource is automatically bound to the resource claim.
 To enable dynamic provisioning the administrator needs to create one or more resource class objects.
 
 ## Connection Secrets
-
```
hasheddan: I believe we have blank lines following headers in other docs. Not sure it even renders differently on the site though.

docs/concepts.md
```diff
@@ -51,8 +50,324 @@ The newly provisioned resource is automatically bound to the resource claim.
...
+[kubernetes-glossary]: https://kubernetes.io/docs/reference/glossary/?all=true
+However we'll summarize some commonly used concepts for convenience.
+
+### CRD
+A standard Kubernetes Custom Resource Definition (CRD), which defines a new type of resource that can be managed declaratively. 
+This serves as the unit of management in Crossplane.
+The CRD is composed of spec and status sections and supports API level versioning (e.g., v1alpha1)
+
+### Controller
+A standard Kubernetes Custom Controller, providing active control loops that own or more CRDs.
```
hasheddan: Should be "that own one or more CRDs" and I think I would actually prefer "that watch one ore more CRDs" instead

docs/concepts.md
```diff
@@ -51,8 +50,324 @@ The newly provisioned resource is automatically bound to the resource claim.
...
+
+### Stack Registry
+A registry where Stacks can be published, downloaded, explored, and categorized. 
+The registry understands a Stack’s custom controller and its CRDs and indexes by both -- you could lookup a custom controller by the CRD name and vice versa.
+
+### Stack Package Format
+The package format for Stacks that contains the Stack definition, metadata, icons, CRDs, and other Stack specific files.  
+
+### Stack Manager
+The component that is responsible for installing a Stack’s custom controllers and resources in Crossplane. It can download packages, resolve dependencies, install resources and execute controllers.  This component is also responsible for managing the complete life-cycle of Stacks, including upgrading them as new versions become available.
```
hasheddan: Looks like this line (and a few others) did not get wrapped,

docs/concepts.md
```diff
@@ -51,8 +50,324 @@ The newly provisioned resource is automatically bound to the resource claim.
...
+specify how they should be satisfied by a managed resource.
+
+### Cloud-Specific Resource Class
+Cloud-specific Resource Classes capture reusable, best-practice configurations for a specific managed service.
+
+For example, Wordpress requires a MySQL database which can be satisfied by CloudSQL, RDS, or Azure DB, so
+cloud-specific resource classes would be created for CloudSQL, RDS, and Azure DB.
+
+### Non-Portable Resource Class
+See [cloud-specific resource class](#cloud-specific-resource-class).
```
hasheddan: Could we say "Another term for `cloud-specific resource class`"?

docs/concepts.md
negz: Doesn't render any differently, but does make it a little nicer to read the markdown. Needn't block getting this merged.

docs/concepts.md
negz: I don't feel strongly.

docs/concepts.md
negz:
```suggestion
A standard Kubernetes Custom Controller, providing active control loops that own one or more CRDs. 
```

PR #826, opened 2019-09-18
Update troubleshooting for 0.3

negz:
### Checklist

I have:

### DIFF
--- a/docs/troubleshoot.md
+++ b/docs/troubleshoot.md
@@ -13,8 +13,10 @@ indent: true
 
 ## Crossplane Logs
 
-The first place to look to get more information or investigate a failure would be in the Crossplane pod logs, which should be running in the `crossplane-system` namespace.
-To get the current Crossplane logs, run the following:
+The first place to look to get more information or investigate a failure would
+be in the Crossplane pod logs, which should be running in the
+`crossplane-system` namespace. To get the current Crossplane logs, run the
+following:
 
 ```console
 kubectl -n crossplane-system logs -lapp=crossplane
@@ -22,14 +24,17 @@ kubectl -n crossplane-system logs -lapp=crossplane
 
 ## Resource Status and Conditions
 
-All of the objects that represent managed resources such as databases, clusters, etc. have a `status` section that can give good insight into the current state of that particular object.
-In general, simply getting the `yaml` output of a Crossplane object will give insightful information about its condition:
+All of the objects that represent managed resources such as databases, clusters,
+etc. have a `status` section that can give good insight into the current state
+of that particular object. In general, simply getting the `yaml` output of a
+Crossplane object will give insightful information about its condition:
 
 ```console
 kubectl get <resource-type> -o yaml
 ```
 
-For example, to get complete information about an Azure AKS cluster object, the following command will generate the below sample (truncated) output:
+For example, to get complete information about an Azure AKS cluster object, the
+following command will generate the below sample (truncated) output:
 
 ```console
 > kubectl -n crossplane-system get akscluster -o yaml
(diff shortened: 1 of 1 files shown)


docs/troubleshoot.md
```diff
@@ -6,92 +6,93 @@ indent: true
...
 
-## Crossplane Logs
+## Resource Status and Conditions
 
-The first place to look to get more information or investigate a failure would be in the Crossplane pod logs, which should be running in the `crossplane-system` namespace.
-To get the current Crossplane logs, run the following:
+Most Crossplane resources have a `status` section that can represent the current
+state of that particular resource. Running `kubectl describe` against a
+Crossplane resource will frequently give insightful information about its
+condition.For example, to determine the status of a MySQLInstance resource
```
soorena776: Nitpick: `condition. For`

docs/troubleshoot.md
```diff
@@ -6,92 +6,93 @@ indent: true
...
+    Status:                False
+    Type:                  Ready
+    Last Transition Time:  2019-09-16T13:46:42Z
+    Reason:                Successfully reconciled managed resource
+    Status:                True
+    Type:                  Synced
 ```
 
-For example, to get complete information about an Azure AKS cluster object, the following command will generate the below sample (truncated) output:
+Most Crossplane resources set exactly two condition types; `Ready` and `Synced`.
```
soorena776: should `exactly` be rephrased as `at least`?

docs/troubleshoot.md
```diff
@@ -6,92 +6,93 @@ indent: true
...
-      Type: Ready
-    bindingPhase: Bound
-    endpoint: crossplane-aks-14af6e93.hcp.centralus.azmk8s.io
-    state: Succeeded
-```
+## Crossplane Logs
+
+The next place to look to get more information or investigate a failure would
+be in the Crossplane pod logs, which should be running in the
+`crossplane-system` namespace. To get the current Crossplane logs, run the
```
soorena776: given that we have introduced an `infra` namespace in stack and services guide, we should also probably mention that here instead of `crossplane-system`

docs/troubleshoot.md
negz:
```suggestion
condition. For example, to determine the status of a MySQLInstance resource
```

docs/troubleshoot.md
```diff
@@ -6,92 +6,93 @@ indent: true
...
-Conditions that have `Status: "True"` are currently active, while conditions with `Status: "False"` happened in the past, but are no longer happening currently.
+```shell
+kubectl -n crossplane-system logs -lapp=crossplane
+```
 
 ## Pausing Crossplane
 
-Sometimes, it can be useful to pause Crossplane if you want to stop it from actively attempting to manage your resources, for instance if you have encountered a bug.
-To pause Crossplane without deleting all of its resources, run the following command to simply scale down its deployment:
+Sometimes, it can be useful to pause Crossplane if you want to stop it from
```
soorena776: nitpick: 
could be rephrased as:
Sometimes when you are encountering a bug, it can be useful to pause Crossplane if you want to stop it from actively attempting to manage your resources.

docs/troubleshoot.md
negz: This is discussing pod logs for the actual crossplane process, which I we still run in `crossplane-system`. We should probably expand this to mention that you might want to look at stack logs too though.

docs/troubleshoot.md
```diff
@@ -6,92 +6,93 @@ indent: true
...
 kubectl -n crossplane-system scale --replicas=1 deployment/crossplane
 ```
 
 ## Deleting a Resource Hangs
 
-The resources that Crossplane manages will automatically be cleaned up so as not to leave anything running behind.
-This is accomplished by using finalizers, but in certain scenarios, the finalizer can prevent the Kubernetes object from getting deleted.
+The resources that Crossplane manages will automatically be cleaned up so as not
+to leave anything running behind. This is accomplished by using finalizers, but
+in certain scenarios, the finalizer can prevent the Kubernetes object from
```
soorena776: nitpick: remove comma `in certain scenarios, the finalizer`

docs/troubleshoot.md
negz: I considered that, but I think it is actually exactly in all cases. At least would leave us space to add more in the future, but exactly is more accurate today.

PR #833, opened 2019-09-19
getting started: change connect your provider page

muvaf:

Our [_Connect Your Cloud Provider_](https://crossplane.io/docs/v0.3/cloud-providers.html) page assumes that cloud-specific pages create `Provider` resource which only AWS one does.

In the current state, all guides start with assuming that _credentials file_(json for azure and gke) is ready for `Provider` resource creation. Also, `Provider` resource has to live in a namespace that the user decides but here in this page the context is not there yet.

Suggestion:
* Change _Connect Your Cloud Provider_ page to _Configure Your Cloud Provider Account_, both title and the content (this PR does that)
* Change AWS to account creation with needed permissions that can work with all guides we have, just like other cloud-specific pages. 

We can also extend these pages to cover `Provider` creation and remove that part from other guides but having the user make the namespace decision early and without much context concerns me.

### Checklist

I have:

### DIFF
--- a/docs/cloud-providers.md
+++ b/docs/cloud-providers.md
@@ -1,43 +1,18 @@
 ---
-title: Connect Your Cloud Providers
+title: Configure Your Cloud Provider Account
 toc: true
 weight: 230
 indent: true
 ---
 
-# Connect Your Cloud Provider
+# Configure Your Cloud Provider Account
 
-In order for Crossplane to be able to manage resources across all your clouds, you will need to add your cloud provider account credentials to Crossplane.
-Use the links below for specific instructions to add each of the following cloud providers:
+In order for Crossplane to be able to manage resources in a specific cloud
+provider, you will need to create an account for Crossplane to use.
+Use the links below for cloud-specific instructions to create an account
+that can be used throughout the guides:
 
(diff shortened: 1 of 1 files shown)


docs/cloud-providers.md
```diff
@@ -1,43 +1,18 @@
...
 weight: 230
 indent: true
 ---
 
-# Connect Your Cloud Provider
+# Configure Your Cloud Provider Account
 
-In order for Crossplane to be able to manage resources across all your clouds, you will need to add your cloud provider account credentials to Crossplane.
-Use the links below for specific instructions to add each of the following cloud providers:
+In order for Crossplane to be able to manage resources across all your clouds,
```
soorena776: should we rephrase `across all your clouds` to `in a specific cloud provider`

hasheddan: How about:
> "In order for Crossplane to be able to manage resources on a cloud provider,
you will need to add your account credentials for that provider to Crossplane.
Use the links below for specific instructions to configure accounts that
can be used throughout the guides:"

muvaf: I don't want to mention _adding to Crossplane_ since the links do not include that at all. They just _create_ the accounts, a process that happens completely on provider UI. Then later in the guides we _add_ those credentials.

PR #838, opened 2019-09-22
Add Instance claim kind

hasheddan:

Adds a VM claim kind and a corresponding default class controller such that infrastructure stacks can add VM resources that can be dynamically provisioned. Because VM configuration can vary greatly across cloud providers (or even within a cloud provider), this initial claim implementation does not surface any configuration details. It is possible we would want `OS` or some other field surfaced in the future.

### Checklist

I have:


apis/compute/v1alpha1/types.go
```diff
@@ -151,3 +151,128 @@ func (kc *KubernetesClusterClassList) GetPortableClassItems() []resource.Portabl
...
+
+var _ resource.Claim = &VirtualMachine{}
+
+// +kubebuilder:object:root=true
+
+// A VirtualMachine is a portable resource claim that may be satisfied by
+// binding to a Virtual Machine managed resource such as an AWS EC2 instance
+// or an Azure VM.
+// +kubebuilder:printcolumn:name="STATUS",type="string",JSONPath=".status.bindingPhase"
+// +kubebuilder:printcolumn:name="CLUSTER-CLASS",type="string",JSONPath=".spec.classRef.name"
```
muvaf: `CLUSTER-CLASS` and `CLUSTER-REF` are forgotten here I guess

hasheddan: Good catch! Fixed now.

hasheddan:
@muvaf I agree that those are general enough that we should consider supporting. Would be happy to add if all are in agreement. cc @negz @jbw976

apis/compute/v1alpha1/types.go
```diff
@@ -151,3 +151,127 @@ func (kc *KubernetesClusterClassList) GetPortableClassItems() []resource.Portabl
...
+	runtimev1alpha1.ResourceClaimSpec `json:",inline"`
+}
+
+var _ resource.Claim = &VirtualMachine{}
+
+// +kubebuilder:object:root=true
+
+// A VirtualMachine is a portable resource claim that may be satisfied by
+// binding to a Virtual Machine managed resource such as an AWS EC2 instance
+// or an Azure VM.
```
negz: I recall that the initial driver for this change was adding support for Packet, who market as bare metal. If we want to bind this kind to bare metal machines (either Packet or eventually on prem) perhaps `Machine` would be appropriate? This would fit both virtual and bare metal machines.

We might also consider `Instance`, which would have symmetry with the existing `CloudSQLInstance`, `RDSInstance`, etc types which tend to mean "a server running software X", but would probably make less sense for a bare metal machine.

hasheddan: Agreed, it seems wrong to call it a VM if binding to bare metal. I would personally be in support of `Instance` because of the generality of the term. Packer refers to their bare metal machines as `Devices`.

soorena776: Just `Instance` seems very generic to me, especially when we have resulted derivatives like `InstanceController`. Maybe something more like `ComputeInstance`, `MachineInstance` or `CPUInstance`?

hasheddan: @soorena776 looking back at this I would tend to agree with you. @prasek @bassam any thoughts on these options?

negz: I'd prefer `ComputeInstance` over `Instance`. I like `Machine` too, and would probably pick that in a vacuum, but `ComputeInstance` matches our other `FooInstance` claim names.

negz:
@hasheddan Do you have an idea of how far off any Stacks (e.g. Packet) are from using this claim? I'm a little hesitant to ship a new claim kind until at least one Stack actually makes use of it.

hasheddan:
@negz official work on Packet stack starts in earnest tomorrow. It might be nice to have a claim type to be able to test with. Still think we need @prasek and @bassam thoughts here.

apis/compute/v1alpha1/types.go
prasek: MachineInstance  👍 

You’re getting a machine (VM or bare-metal). Compute is too general IMO.

hasheddan:
Closed in favor of #942

Issue #728, opened 2019-08-29
Separation of levels between Cloud Spec and Crossplane Spec in CR


muvaf:
### What problem are you facing?

Right now, an example CR looks like this:
```
apiVersion: compute.gcp.crossplane.io/v1alpha1
kind: Network
metadata:
  creationTimestamp: "2019-08-28T17:39:24Z"
  finalizers:
  - finalizer.managedresource.crossplane.io
  generation: 2
  name: my-legacy-network
  namespace: crossplane-system
  resourceVersion: "29218"
  selfLink: /apis/compute.gcp.crossplane.io/v1alpha1/namespaces/crossplane-system/networks/my-legacy-network
  uid: 55de6569-0085-43d2-8931-4a0da76cb682
spec:
  name: monus-legacy-network-res
  providerRef:
    name: monus-gcp
    namespace: crossplane-system
  reclaimPolicy: Delete
  routingConfig:
    routingMode: REGIONAL
  writeConnectionSecretToRef: {}
status:
  IPv4Range: 10.240.0.0/16
  conditions:
  - lastTransitionTime: "2019-08-29T05:55:16Z"
    reason: Successfully reconciled managed resource
    status: "True"
    type: Synced
  creationTimestamp: "2019-08-28T10:45:27.445-07:00"
  gatewayIPv4: 10.240.0.1
  id: 75666386684937048
  kind: compute#network
  name: monus-legacy-network-res
  routingConfig:
    routingMode: REGIONAL
  selfLink: https://www.googleapis.com/compute/v1/projects/crossplane-playground/global/networks/monus-legacy-network-res
```
As you see the cloud provider specifications and Crossplane fields are on the same level both in `Spec` and `Status`. As you can see, the fields that are retrieved from cloud provider have collision possibility since they are not that special, like `name` or `creationTimestamp` or more in the future. I'd like to know that what fields are cloud provider info what fields belong to Crossplane itself.

### How could Crossplane help solve your problem?
I believe that we should have a field named `specForProvider` in `Spec` and `statusAtProvider` in `Status`. It'd look like the following:
```
apiVersion: compute.gcp.crossplane.io/v1alpha1
kind: Network
metadata:
  creationTimestamp: "2019-08-28T17:39:24Z"
  finalizers:
  - finalizer.managedresource.crossplane.io
  generation: 2
  name: my-legacy-network
  namespace: crossplane-system
  resourceVersion: "29218"
[...]

PR #840, opened 2019-09-23
One-pager for Managed Resources API Patterns

### LINKED ISSUE #831: Converge on a consistent Custom Resource design for managed resources

Currently, as a developer, I am having a hard time implementing a managed resource kind for a specific cloud provider's service due to the decisions around what fields should go where and what should guide me in those decisions.

Have a one-pager that list guiding principles and examples for managed resource CR design that all implementations can start from and deviate when required so that entry barrier gets lower and we have a consistent API throughout all managed resources.

muvaf:

This one-pager serves as ground for the discussions that will hopefully get us closer to having a consistent managed resource API.

I think that we should aim to make implementation of managed resources as mechanical as possible. This guide will serve as starting point on how we want things to look like even if we do/don't have code-generation sometime in the future.

Fixes #831, #728
### Checklist

I have:

### DIFF
--- a/design/one-pager-managed-resource-api-design.md
+++ b/design/one-pager-managed-resource-api-design.md
@@ -1,0 +1,402 @@
+# Managed Resources API Patterns
+* Owner: Muvaffak Onus (@muvaf)
+* Reviewers: Crossplane Maintainers
+* Status: Draft
+
+## Terminology
+
+* _External resource_. An actual resource that exists outside Kubernetes, typically in the cloud. AWS RDS or GCP Cloud
+  Memorystore instances are external resources.
+* _Provider_. The entity that owns the physical layer of the external resource. AWS, GCP and Azure are examples of this.
+* _Managed resource_. The Crossplane representation of an external resource. The `RDSInstance` and `CloudMemorystoreInstance`
+  kinds are managed resources.
+* _Claim_. The Crossplane representation of a request for the allocation of a managed resource. Resource claims
+  typically represent the need for a managed resource that implements a particular protocol. MySQLInstance and
+  RedisCluster are examples of resource claims.
+* _Provider_. Cloud provider such as GCP, AWS, Azure offering IaaS, cloud networking, and managed services.
+* _Spec_. The sub-resource of Kubernetes resources that represents the desired state of the user.
+* _Status_ The sub-resource of Kubernetes resources that represents the most up-to-date information about the resource.
+
+For other terms, see [glossary].
+
+## Background
+Crossplane extends Kubernetes heavily and the maintainers always try to adopt the patterns that are already seen in 
+Kubernetes resources, notably declarative style API. However, it's not always intuitive for everyone to see what goes to
+where in a way that is consistent and just as expected by the consumers of those resources.
+
+This document tries to capture how the bridge between Crossplane and provider is shaped. So, the features of managed
+resources that are generic to all of them are not included here, for example reclaim policies. This is more about the
+things that a developer should keep in mind when they try to implement a new managed resource.
+
+A few questions that this document needs to help answering:
+- What goes into `Spec` and what goes into `Status` of a managed resource?
+- Should _all_ information from the provider be represented in the Custom Resource?
+- What parts of the Custom Resource should be allowed to be consumed by the consumers?
+- Do we represent _desired_ state, _observed_ state or both?
+- How should we differentiate the fields that are used by provider from Crossplane's fields?
+
+While this document is heavily opinionated, each provider can wildly vary between each other. So, treat this document as
+a starting point, try to stick to it for consistency but deviate if required and open an issue about why you had to
+deviate so that we can reassess and improve it.
+
+## Guiding Principles
+
+Our users are the stakeholders on how we make decisions. There are two types of users for managed resources:
+* Input Users: Users who configure/create/manage the resource. Crossplane Claim Controller, Cluster Admin are a few
+  examples of this type. We want our `Spec` to be sole place for them to configure everything about the resource.
+* Output Users: Users who refer to/query/use/extract information from the managed resource. We want them to be able to
+  perform all read-only operations on the given resource as if they have the cloud provider's API available. This can
+  include configurations and metadata information about the resource. We also want them to access only `Status`
+  sub-resource of the managed resource because this is where the most up-to-date information about the resource is
+  stored and we are able to limit RBAC of these users to access only `Status`.
+  
+### Naming Conventions
+
+There are two naming decisions here to make:
+* Managed resource CR name
+* Identifier string of the external resource in the cloud provider.
+
+_Name_ in this context means the identifier string for the resource in its environment. CR name is the identifier
+for resource in Kubernetes environment. But for external resources, this can be either `id`, `uid` or `name` properties
+of the resource. The one we want to name is the one that is used when _referring to_ that resource in the same
+environment. For example, in AWS, VPCs do not have a name but `VpcID` properties that is used for identification.
+
+#### Custom Resource Instance Name
+
+The name of the managed resource CR, which lives in our control cluster. The claim controller who creates the managed
+resource gives its name or some other entity might create the managed resource, like user themselves. So, do not assume
+that all the managed resources will have the same naming format anywhere.
+One enforcing condition is that it has to be unique amongst other instances of the CRD. The best way to guarantee that
+is to include Kubernetes UID in the name. The following format is valid as long as the length of the name is not longer
+than 253 characters. Given that we know UID is 36 characters long, name for the service you choose should be
+differentiable enough with 217 characters.
+
+For example, the following would be a good name for a CloudSQL managed resource CR:
+```
+cloudsql-7cb46d78-4556-4904-93d2-b34f6d5ccadf
+```
+
+#### External Resource Name
+
+Related to https://github.com/crossplaneio/crossplane/issues/624
+
+The decision for an external resource name is made by the controller of that managed resource in `Create` phase. The
+following is the possible cases:
+* The provider doesn't allow us to specify a name. Fetch it after its client's `POST` call.
+* The provider allows to specify a name.
+  1. Check whether CR name is suitable as resource's external name. If so, use it. If not;
+  2. Use `servicename-k8s UID of CR`, i.e. `cloudsql-7cb46d78-4556-4904-93d2-b34f6d5ccadf`. If that doesn't comply;
+  3. Generate your own name in the following format in a compliant way:
+```
+servicename-<random-string>
+<random-string>: lower case letters, upper case letters, integers and dash (-). Dash (-) should be neither in the
+beginning nor in the end. Until 36 characters(k8s UID length), the longer random string the better.
+```
+  4. If 3 doesn't work for you, come up with your own format that includes the name of the service if possible.
+
+Note that the managed resource might be created manually by the user, so, the CR name can be anything and you need to
+have a validation step for the CR name before you use it.
+
+In all cases, write the name to annotation with key `crossplane.io/external-name` if a value does not already exist. Use 
+the value of that annotation as external resource name in _all_ queries.
+
+#### Field Names
+
+The thinking process of naming decision could respect provider's decision as the starting point and deviate from that
+if there is a very good reason to. Naming parity makes inter-resource references easier and lowers the entry barrier for
+both developers and users. This is especially handy in cases where Crossplane does not yet support the _referred_ resource
+or user may not want some of their resources in their Crossplane environment.
+
+Ideally, for the resources that do have Crossplane representations, you can append `Ref` to the field name,
+i.e. `network (string)` -> `networkRef (Network)`
+
+In some cases, we might have collisions or cases where provider field name is too similar to another field in the
+managed resource's Custom Resource fields. The section [Owner-based Struct Tree](#owner-based-struct-tree) tries to
+tackle that.
+
+#### Example
+
+```yaml
+
+---
+# example-network will be the VPC that all cloud instances we'll create will use.
+apiVersion: compute.gcp.crossplane.io/v1alpha2
+kind: Network
+metadata:
+  name: network-7cb46d78-4556-4904-93d2-b34f6d5ccadf
+  namespace: gcp
+  annotations:
+    crossplane.io/external-name: "network-7cb46d78-4556-4904-93d2-b34f6d5ccadf"
+spec:
+  ...
+```
+
+```yaml
+
+---
+# example-network will be the VPC that all cloud instances we'll create will use.
+apiVersion: compute.gcp.crossplane.io/v1alpha2
+kind: Network
+metadata:
+  name: <too-long-string>
+  namespace: gcp
+  annotations:
+    crossplane.io/external-name: "network-123-my-random-gen-id"
+spec:
+  ...
+```
+
+### High Fidelity
+
+Related to https://github.com/crossplaneio/crossplane/issues/621 and https://github.com/crossplaneio/crossplane/issues/530
+
+Crossplane managed resources should expose everything that provider exposes to its users as much as possible. A few 
+benefits of that high fidelity:
+
+* The ability to read and set every 'knob' the external resource supports. This would make it easy to support wide 
+variety of customizations without having to think of them in development time.
+* An obvious mapping of Crossplane's fields to cloud provider API fields, exact name match as much as possible. This 
+would make it easier to work with Crossplane for users who are familiar with the given provider in many ways 
+including easier troubleshooting and cooperation with other existing infrastructure tools.
+
+Ideally, what we want is to expose all possible configurations in `Spec` for input users and whatever might be needed
+ by the output users of this resource in the `Status`. Generally, it's better to just copy all fields from the
+ provider's API as starting point and eliminate them one by one if needed.
+
+What goes into `Spec`:
+* Does the provider allow configuration of this parameter at _any_ point in the life cycle of the resource? Include if yes.
+  This includes the fields that are late-initialized, meaning some fields could be populated by the provider when you
+  actually create the resource but it also gives you the ability to change them. An example for this could be
+  auto-generated resource tags or resource-specific defaults. If the provider tags the resource without us telling them to
+  do so, controller should update `Spec` and input user should make changes on that current value of the field.
+  Related to https://github.com/crossplaneio/crossplane-runtime/issues/10
+* Is this field represented as standalone managed resource? Do not include if the answer is yes.
+  We should not manage an external resource in the CR other than its original external resource. For example, Azure
+  VirtualNetwork object allows you to create Subnets by providing an array of Subnet objects as one of its fields. However,
+  Subnet is already another managed resource supported by Crossplane. In that case, we should not allow configuration of
+  Subnets through VirtualNetwork CR but require people to do it via Subnet CR. Though you might refer to it for
+  configuration purposes, having two controllers managing one resource (VirtualNetwork CR and Subnet CR controllers) would
+  not work well. So, do not include it in the fields and be careful during your provider calls not to affect the other
+  resource.
+  
+  What if the sub-resource is not yet supported as managed resource in Crossplane? In that case, you should first
+  consider implementing that managed resource, if not suitable, only then include it in the CR.
+
+Note that you should make updates to `Spec` fields that are empty. We do not override user's desired state and if they
+have no control over it in any case, do not include it in `Spec`.
+
+What goes into `Status`:
+* Does the provider give the information of that field on the first level of the query, `GET`? Include if yes.
+  For the subresource that are represented as standalone managed resource by Crossplane, include only identification
+  information instead of full object representation. For example, if received Azure VirtualNetwork object has an array
+  of Subnet objects (which is represented as standalone), represent each one by their name in a string array.
+
+### Owner-based Struct Tree
+
+Related to https://github.com/crossplaneio/crossplane/issues/728
+
+The provider API can be assumed as the owner of its configuration and output fields since the decisions around these
+fields are left to them. It makes sense to separate those fields from Crossplane or Kubernetes-related fields for a few
+reasons:
+* Prevent name collisions (or too much similarity) as the decision-maker of provider API's naming is different than
+  other fields' decision maker.
+* A better representation of what goes to/comes from provider and what's Crossplane or Kubernetes specific.
+* Let the user know what configuration is being sent and received as much as possible in the provider's API language.
+
+A good separation method can be having provider fields as a sub-struct of `Spec` and `Status` instead of directly
+embedding them.
+
+Note that, in all cases, we should be able to guarantee that `spec.forProvider` is the last representation of the things
+that we'll include in our CRUD calls after all override/default/dependency operations are done. Also `status.atProvider`
+represents the most up-to-date state in the provider as raw as possible. In other words, when user sees an error, they
+should be able to tell what exact configuration was used that caused that error by looking at `spec.forProvider` _and_
+they should be able to tell whatever operation is done `status.atProvider` is the current state.
+
+An actual example:
+
+Embedding:
+
+```yaml
+apiVersion: compute.gcp.crossplane.io/v1alpha1
+kind: Network
+metadata:\
+  name: my-legacy-network
+  namespace: crossplane-system
+  selfLink: /apis/compute.gcp.crossplane.io/v1alpha1/namespaces/crossplane-system/networks/my-legacy-network
+spec:
+  name: monus-legacy-network-res
+  providerRef:
+    name: monus-gcp
+    namespace: crossplane-system
+  reclaimPolicy: Delete
+  routingConfig:
+    routingMode: REGIONAL
+status:
+  IPv4Range: 10.240.0.0/16
+  conditions:
+  - lastTransitionTime: "2019-08-29T05:55:16Z"
+    reason: Successfully reconciled managed resource
+    status: "True"
+    type: Synced
+  creationTimestamp: "2019-08-28T10:45:27.445-07:00"
+  gatewayIPv4: 10.240.0.1
+  id: 75666386684937048
+  kind: compute#network
+  name: monus-legacy-network-res
+  routingConfig:
+    routingMode: REGIONAL
+  selfLink: https://www.googleapis.com/compute/v1/projects/crossplane-playground/global/networks/monus-legacy-network-res
+```
+
+Sub-struct:
+
+```yaml
+apiVersion: compute.gcp.crossplane.io/v1alpha1
+kind: Network
+metadata:
+  name: my-legacy-network
+  namespace: crossplane-system
+  selfLink: /apis/compute.gcp.crossplane.io/v1alpha1/namespaces/crossplane-system/networks/my-legacy-network
+spec:
+  providerRef:
+    name: monus-gcp
+    namespace: crossplane-system
+  reclaimPolicy: Delete
+  writeConnectionSecretToRef: {}
+  forProvider:
+    name: monus-legacy-network-res
+    routingConfig:
+      routingMode: REGIONAL
+status:
+  conditions:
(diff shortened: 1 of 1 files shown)


design/one-pager-managed-resource-api-design.md
```diff
@@ -0,0 +1,265 @@
+# Managed Resources API Patterns
+* Owner: Muvaffak Onus (@muvaf)
+* Reviewers: Crossplane Maintainers
+* Status: Draft
+
+## Terminology
+
+* _External resource_. An actual resource that exists outside Kubernetes, typically in the cloud. AWS RDS or GCP Cloud
```
soorena776: nitpick: Examples: AWS RDS, or GCP CouldSQL instances.

design/one-pager-managed-resource-api-design.md
```diff
@@ -0,0 +1,265 @@
...
+  Memorystore instances are external resources.
+* _Provider_. The entity that owns the physical layer of the external resource. AWS, GCP and Azure are examples of this.
```
soorena776: I think @negz have been thinking of a more specific definition for a Provider.

design/one-pager-managed-resource-api-design.md
```diff
@@ -0,0 +1,265 @@
+# Managed Resources API Patterns
+* Owner: Muvaffak Onus (@muvaf)
+* Reviewers: Crossplane Maintainers
+* Status: Draft
+
+## Terminology
```
soorena776: Should we link to the existing glossary?
https://github.com/crossplaneio/crossplane/blob/master/docs/concepts.md#glossary

muvaf: I think terminology section provides information about just the terms that you need to know for that document, so, it's more contained and easy to digest than full glossary. But adding the glossary link in the end of that section makes sense probably.

design/one-pager-managed-resource-api-design.md
```diff
@@ -0,0 +1,269 @@
...
+update `Spec` and input user should make changes on that current value of the field.
+
+What goes into `Status`:
+Does the provider give the information of that field on the first level of the query, `GET`?
+For example, `network` could be a top level information in `GET` request of a database resource but the details of that
+network should not be included via making extra calls, we want only the information that is present on the top level.
+If there is a reference to another provider API object (the object could be a tree of other objects), only include the
+provider's identifier of that object to make it easier to query it when needed.
+
+### Owner-based Struct Tree
```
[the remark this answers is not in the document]
muvaf: `spec.external` or `status.external` seems more formal but I believe the name should imply that this is what we send to/receive from the provider and it's close to provider's representation. So, if we can come up with another name pair that implies that, I'd fine with it.

If we keep the current ones, `spec.forProvider` and `status.atProvider` make sense.

design/one-pager-managed-resource-api-design.md
```diff
@@ -0,0 +1,269 @@
...
+or user may not want some of their resources in their Crossplane environment.
+
+Ideally, for the resources that do have Crossplane representations, we'll have `ObjectReference` type for
+these fields. In that case, you can append `Ref` to the field name, i.e. `network (string)` -> `networkRef (ObjectReference)`
+
+In some cases, we might have collisions or cases where provider field name is too similar to another field in the
+managed resource's Custom Resource fields. This is partly solved in [Consumer-based Struct Levels](#consumer-based-struct-levels)
+section.
+
+### High Fidelity
```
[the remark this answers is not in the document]
muvaf: This part tries to tackle that:
```
For example, `network` could be a top level information in `GET` request of a database resource but the details of that
network should not be included via making extra calls, we want only the information that is present on the top level.
If there is a reference to another provider API object (the object could be a tree of other objects like in Azure), only
include the provider's identifier of that object to make it easier to query it when needed. Trying to include the whole
tree can even go to the state where all your setup that is connected to each other is represented under one resource's
status, which is a very bad UX.
```

muvaf:
@negz Thanks for pointing out that document! I've made additions but I didn't want to touch all of the topics as I want this document to be more about the steps where the developer tries to develop the parts of the managed resource that is related to the relation between provider's API and our managed reconciler. For example, naming or high fidelity depend on the resource/API you call but reclaim policy is generic for all managed resources.

The parts that appear in the Google doc but is not included in this document:
* Importing existing resources. The naming section touches on how to make this easier but, in general, I don't go into details of that.
* Dependency references. As we talked, this will have its own one-pager.
* Unclaimable managed resources. This is more about general managed resource logic that we have instead of provider<>managed resource/its controller relation.
* Reclaim policy. This is also a generic one that doesn't depend on the specific resource that the dev is trying to implement.
* Claim & Class references in Spec and claim<>class<>nonportable class<>managed resource relations. This one is also generic.

design/one-pager-managed-resource-api-design.md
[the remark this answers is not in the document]
hasheddan: Yes, @negz you are correct. There is not currently an issue open specifically for this Azure case. One thing that could be difficult with your suggestion here @muvaf is that it is not always readily apparent which information is "top level" for the apis, and it can require some digging and testing to determine if a nested object can actually be applied at creation time.

design/one-pager-managed-resource-api-design.md
```diff
@@ -0,0 +1,294 @@
...
+it. The `Create` call uses whatever it sees in the `Spec`
+
+There are cases where naming by the API caller is not allowed, like VPC in AWS, we do not make any decision and do not
+include a field for that in the `Spec` since it's not configurable by the user. We include it in the `Status` when we
+receive it and other entities access `Status` to fetch the name.
+
+If we are able to provide a name, the decision flow is as following:
+
+1. Use CR name if the provider allows 253 characters for resource name. If not:
+2. Write your own name generation format with a random string generator that you can be sure the generated name
```
displague: What if:
* the provider auto-generates unique names
* the provider lets you supply a name
* the character max is less than 253

I think this may be a common case worth included in this ruleset.

design/one-pager-managed-resource-api-design.md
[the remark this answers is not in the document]
negz: I think my main gripe about "provider" is I always read it as short for "cloud provider". If you interpret it this way it doesn't make sense in the context of Rook backed managed resources, for example.

If I instead think of a "Crossplane provider" (as in the thing configured by `kind: Provider`) as being a superset of "cloud providers", then I agree `spec.forProvider` and `status.atProvider` make sense.

design/one-pager-managed-resource-api-design.md
```diff
@@ -0,0 +1,294 @@
...
+update `Spec` and input user should make changes on that current value of the field. Related to https://github.com/crossplaneio/crossplane-runtime/issues/10
+
+What goes into `Status`:
+Does the provider give the information of that field on the first level of the query, `GET`?
+For example, `network` could be a top level information in `GET` request of a database resource but the details of that
+network should not be included via making extra calls, we want only the information that is present on the top level.
+If there is a reference to another provider API object (the object could be a tree of other objects like in Azure), only
+include the provider's identifier of that object to make it easier to query it when needed. Trying to include the whole
+tree can even go to the state where all your setup that is connected to each other is represented under one resource's
+status, which is a very bad UX.
```
negz: My understanding is that we may need to make a recommendation around this Azure-ish case for spec inputs as well as status outputs. Two examples of this are:

* Creating or updating an [Azure virtual network](https://docs.microsoft.com/en-us/rest/api/virtualnetwork/virtualnetworks/createorupdate). The Azure API allows you to specify a set of subnets and peerings that also exist as distinct Azure API objects and thus _could_ be modeled as distinct managed resources.
* Creating or updating a [GKE cluster](https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1/projects.locations.clusters/create). The API allows you to provide a set of node pools, that also exist as distinct GKE API objects and thus _could_ be modeled as distinct managed resources.

Perhaps Crossplane should take a stance something like "if this external resource could be CRUD as either a subfield of a parent resource _or_ as a distinct resource, Crossplane supports only the latter". So basically we'd omit the `nodePools` of a GKE cluster managed resource and instead require you to create distinct node pool managed resources instead.

design/one-pager-managed-resource-api-design.md
```diff
@@ -0,0 +1,294 @@
...
+
+What goes into `Spec`:
+Does the provider allow configuration of this parameter at _any_ point in the life cycle of the resource? Include if yes.
+This includes the fields that are late-initialized, meaning some fields could be populated by the provider when you
+actually create the resource but it also gives you the ability to change them. An example for this could be
+auto-generated resource tags. If the provider tags the resource without us telling them to do so, controller should
+update `Spec` and input user should make changes on that current value of the field. Related to https://github.com/crossplaneio/crossplane-runtime/issues/10
+
+What goes into `Status`:
+Does the provider give the information of that field on the first level of the query, `GET`?
```
negz: Are you still thinking this would be a superset of the spec? i.e. That we'd include the current observed state of all spec fields in case they diverged from the spec? I think that's implied here, but I think it may be worth making it explicit given that it's an uncommon pattern.

(FWIW I'm still super on the fence about whether we should do this.)

negz: Ah - I see this is covered below in immutable properties but I still think it's worth pointing out out here.

design/one-pager-managed-resource-api-design.md
```diff
@@ -0,0 +1,294 @@
...
+
+Related to https://github.com/crossplaneio/crossplane/issues/727
+
+Some of the fields that include in `Spec` can only be configured in creation call of the resource, later you cannot
+update them. However, Kubernetes Custom Resources validation mechanisms do not yet support that behavior, see https://github.com/kubernetes/enhancements/pull/1099
+for details.
+
+There are some solutions like admission webhooks to enforce immutability of some of the fields, however, current behavior
+is that Crossplane shows the user what the error is received when the call that includes a change in the immutable
+property is made and the most up-to-date status under `Status`. In this scenario, we have to make sure that all immutable
```
negz: > Crossplane shows the user what the error is received when the call that includes a change in the immutable
property is made

I wish this were true, but I'm pretty sure it's worse in many cases. Some cloud provider APIs have distinct endpoints (and thus Go structs) for update operations, so it's possible that when the user edits a field the provider considers immutable we'll just never ask that it be updated. ReplicationGroup is an example of this. If someone edited a field that wasn't covered in `NewModifyReplicationGroupInput` we'd just happily accept the change to our custom resource and do nothing:

https://github.com/crossplaneio/stack-aws/blob/b64415a/pkg/controller/cache/managed.go#L217
https://github.com/crossplaneio/stack-aws/blob/b64415a/pkg/clients/elasticache/elasticache.go#L142

I think a good first step toward preventing this would be to require managed resource authors include the `//+immutable` comment marker that is proposed by https://github.com/kubernetes/enhancements/pull/1099. While the KEP is up for debate we can at least teach our API documentation generator to document that such fields are immutable, and if we're lucky we'll get CRD based immutability for free in some future version of Kubernetes. 😬

design/one-pager-managed-resource-api-design.md
```diff
@@ -0,0 +1,294 @@
...
+abides the rules of the provider. Following format is preferred if possible:
+```
+servicename-<random-string>
+<random-string>: lower case letters, upper case letters, integers and dash (-). Dash (-) should be neither in the
+beginning nor in the end.
+```
+3. If 2 doesn't work for you, come up with your own format that includes the name of the service if possible.
+
+In _all_ cases, we need to show the name in the `Status` so that other resources can know whatever is decided and
+successfully applied in a guaranteed way.
```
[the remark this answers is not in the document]
muvaf: This is a very good point. I agree that we should be able to reconstruct `Status` when it's deleted. According to the section that defines what goes to `Spec`, this isn't covered.
> Does the provider allow configuration of this parameter at any point in the life cycle of the resource? Include if yes. This includes the fields that are late-initialized, meaning some fields could be populated by the provider when you actually create the resource but it also gives you the ability to change them.

So, external UID that is generated by the provider _has to_ live somewhere outside of `Status`. I want to defer the discussion of _where_ to your other comment's thread https://github.com/crossplaneio/crossplane/pull/840#discussion_r329294308

design/one-pager-managed-resource-api-design.md
```diff
@@ -0,0 +1,294 @@
...
+
+In some cases, we might have collisions or cases where provider field name is too similar to another field in the
+managed resource's Custom Resource fields. The section [Owner-based Struct Tree](#owner-based-struct-tree) tries to
+tackle that.
+
+### High Fidelity
+
+Related to https://github.com/crossplaneio/crossplane/issues/621 and https://github.com/crossplaneio/crossplane/issues/530
+
+In most of the cases, an application setup requires the cloud resources to be connected/related in a custom way for
```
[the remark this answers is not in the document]
muvaf: Agreed. Changed that section to include your additions.

design/one-pager-managed-resource-api-design.md
```diff
@@ -0,0 +1,294 @@
...
+  perform all read-only operations on the given resource as if they have the cloud provider's API available. This can
+  include configurations and metadata information about the resource. We also want them to access only `Status`
+  sub-resource of the managed resource because this is where the most up-to-date information about the resource is
+  stored and we are able to limit RBAC of these users to access only `Status`.
+  
+### Naming Conventions
+
+There are two naming decisions here to make:
+* Managed resource CR name
+* Identifier string of the external resource in the cloud provider.
```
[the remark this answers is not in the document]
muvaf: I see your point. Only in `Status` doesn't make sense either since we want to be able to reconstruct `Status` when it's lost. Having it in `Spec` doesn't make sense since it'd mean that a field is made available in _desired state_ but user can never actually desire it, that's already covered in _what goes to spec_ section

Annotation makes total sense in this case, also it's not an unfamiliar concept to indicate external config as annotation https://kubernetes.io/docs/concepts/services-networking/service/#ssl-support-on-aws

I couldn't see where you mention the case where the name should be chosen by us (claim controller case) and metadata name format doesn't comply with the provider's style, I guess you're onboard with having the developer write a name generator for that specific resource. This would also go to  `crossplane.io/external-name-override`.

In case where user creates the resource and it doesn't comply, I don't think we should take action other than showing the error.

I'll change the section as you suggested.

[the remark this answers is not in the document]
muvaf: How about we have the annotation in _all_ cases? In happy case, `Create` function of the given resource can assign the name in metadata to the annotation field's value so that we can make it easier for clients that wants to access the external name by pointing to only an annotation key instead of having them consider this alternation logic. Though this logic is set once per each resource type, so, it's not too much to ask them to be careful per resource type.

design/one-pager-managed-resource-api-design.md
muvaf: Yeah, I didn't take that case into account. I guess the best we can do for now to have `//+immutable` and hope one day we either have admission webhooks generated using this marker or native k8s support to actually incorporate it.

> If someone edited a field that wasn't covered in NewModifyReplicationGroupInput we'd just happily accept the change to our custom resource and do nothing

This upset me a little.

design/one-pager-managed-resource-api-design.md
muvaf: As long as we can come up with something else that covers the following cases, I'd be more than happy to change that.

* It should be sufficient to allow access to `Status` sub-resource(via RBAC) for all-purpose information extraction by other entities. Meaning, I would like to be able to extract the current settings, identification information, metadata etc. only by looking at `Status`.
* We need to represent the most up-to-date state in happy cases as well as error cases. If user changed the desired state and if we either in the middle of applying it/couldn't apply it/never plan to apply it, the user should be able to know get the current state even if no error shows up. If we have an update-able config only in `Spec`, we're not able to do that.

However, we can also discuss the need for two points mentioned here. For instance, if we decide the first one is not worth covering, we can remove the creation-time and metadata fields from `Status` and keep only update-able fields there.

design/one-pager-managed-resource-api-design.md
muvaf: > if this external resource could be CRUD as either a subfield of a parent resource or as a distinct resource, Crossplane supports only the latter

I thought about this a lot and I think we might _have to_ do this way even though we want to support referring resources that are not managed by the given Crossplane instance or Crossplane has no support for that resource type yet.

Let's say we created a struct that has identification information of []*SubnetReference and included this as a field in Azure VirtualNetwork to support adding non-Crossplane-managed Subnet resource.

```
type VirtualNetwork struct {
  subnets []*SubnetReference
}
```
I created a VirtualNetwork object with this array having a subnet named `A`. Azure will create the VirtualNetwork and the Subnet `A`. Then I went ahead and created an actual Subnet CR with subnet name `B` that refers to my VirtualNetwork. In that case my desired state about how VirtualNetwork's subnets should be constructed is divided; virtual network controller will delete `B` since it sees the desired state of VirtualNetwork having only `A` subnet, on the other hand, subnet controller will say `B` who refers to my VirtualNetwork should exist. So, possibly, one will delete and the other one will create.

In the current implementation, VirtualNetwork object doesn't mention any Subnet at all. So, the desired state of Subnet part of the VirtualNetwork is managed only by our Subnet resource and that is consistent. So, I guess we'll either expand VirtualNetwork resource to make it the only way of creating Subnets or we will eliminate other resource references in it.

We can expand VirtualNetwork and remove Subnet, but we don't know where this might end. It can be a really long tail of resources provisioned in only one CR. So, we might have to take that stance where we say you have to create a CR for that sub-resource if we have support for it. I think having external resource policy read-only mode(https://github.com/crossplaneio/crossplane-runtime/issues/22#issuecomment-526459345) will help users to add resource only for reference purposes even though they don't want to manage them with Crossplane or with _that_ Crossplane instance.

I think this analogy applies to GKE node pools as well.

design/one-pager-managed-resource-api-design.md
muvaf: How about this; as long as `resourceType+UID` is less than what's allowed by the provider, use it. If it's longer that allowed, do not cut the tail, implement your own random generator for UID part whose length is suitable.

design/one-pager-managed-resource-api-design.md
hasheddan: One thing that came up when syncing with @muvaf on this topic was that, as mentioned, we are currently ignoring the embedded `Subnet` slice on the Azure `VirtualNetwork`, so adding a subnet via a `Subnet` custom resource does not cause the `VirtualNetwork` controller to update the `VirtualNetwork`. However, I am concerned that if some other part of the `VirtualNetwork` changes that causes the controller to update, that it may delete any subnets that were created. The subnets would then be recreated by the `Subnet` controller, but it could have some pretty severe negative effects on other resources. It is possible that the Azure API will not delete subnets if update is called on a virtual network, but I am not certain. I am going to test this out today.

design/one-pager-managed-resource-api-design.md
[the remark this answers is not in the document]
hasheddan: I think the implementation of the Kubernetes `Provider` type as proposed in #841 would expand the notion of a "provider" and thus make `spec.forProvider` and `status.atProvider` suitable.

muvaf:
> I do think we'll need to somehow tie the topics touched on in this document together with those covered in others (e.g. dependencies, reclaim policy) to make sure everything makes sense as a consistent whole.

Makes sense. What we could do is maybe have a design doc that has summary of _every_ feature managed resources have in one paragraph and then link to the related one-pager/design doc/issue for details about that specific feature.

> I believe @bassam suggested it would be a good idea to write up example YAMLs representing our ideal of how various Crossplane resources would be shaped. That's probably something we should do separately to this document.

I believe if we refactor one or two managed resources to adopt this pattern, then we can easily point to them and also use their YAMLs in the documents. After merging this, I'll open issues for that refactoring effort. I tried to add example snippets to this document anyway.

@negz Thanks a lot for your thorough review! I have made the necessary changes and answered your comments.

design/one-pager-managed-resource-api-design.md
hasheddan: @muvaf @negz I confirmed this behavior and have opened an issue to track crossplaneio/stack-azure#23

muvaf:
@negz That's how it'd look like in an example resource https://github.com/crossplaneio/stack-gcp/pull/41
Though that doesn't cover all the cases in the one-pager, it's simple enough to cover the basic scenario and applicability of the things mentioned

Here is creation YAML:
```
apiVersion: compute.gcp.crossplane.io/v1alpha2
kind: Network
metadata:
  name: example-network
  namespace: gcp
spec:
  providerRef:
    name: gcp-provider
    namespace: gcp
  reclaimPolicy: Delete
```

After the reconciliation:
```
apiVersion: compute.gcp.crossplane.io/v1alpha2
kind: Network
metadata:
  annotations:
    crossplane.io/external-name: example-network
  creationTimestamp: "2019-10-02T16:39:32Z"
  finalizers:
  generation: 8
  name: example-network
  namespace: gcp
  resourceVersion: "6485"
  selfLink: /apis/compute.gcp.crossplane.io/v1alpha2/namespaces/gcp/networks/example-network
  uid: 05e0ec72-cc7d-4bcc-99fc-e11d0cb2bf6e
spec:
  forProvider:
    IPv4Range: 10.240.0.0/16
    autoCreateSubnetworks: false
    description: ""
    routingConfig:
      routingMode: GLOBAL
  providerRef:
    name: gcp-provider
    namespace: gcp
  reclaimPolicy: Delete
  writeConnectionSecretToRef: {}
status:
  atProvider:
    IPv4Range: 10.240.0.0/16
    autoCreateSubnetworks: false
    creationTimestamp: "2019-10-02T09:48:57.869-07:00"
    description: ""
    gatewayIPv4: 10.240.0.1
    id: 4711805157157464598
    routingConfig:
      routingMode: GLOBAL
    selfLink: https://www.googleapis.com/compute/v1/projects/crossplane-playground/global/networks/example-network
  conditions:
  - lastTransitionTime: "2019-10-02T16:50:51Z"
    reason: Managed resource is available for use
    status: "True"
    type: Ready
  - lastTransitionTime: "2019-10-02T17:19:36Z"
    reason: Successfully reconciled managed resource
    status: "True"
    type: Synced
```

Note how `Spec` has been populated with the provider's defaults instead of Golang defaults that may not be what provider thinks should be default when a setting is not given.

design/one-pager-managed-resource-api-design.md
negz: Give there's a solid technical reason to do this, but that it's an uncommon pattern for Kubernetes, I suggest we avoid doing it for the time being but make sure there's an issue tracking the idea. We can then implement this if the main problem we foresee (hidden drift between the desired and actual state) becomes a real issue based on user feedback.

muvaf:
@negz updated with the alignment that we landed on with you and bassam.

design/one-pager-managed-resource-api-design.md
```diff
@@ -0,0 +1,398 @@
...
+## Guiding Principles
+
+Our users are the stakeholders on how we make decisions. There are two types of users for managed resources:
+* Input Users: Users who configure/create/manage the resource. Crossplane Claim Controller, Cluster Admin are a few
+  examples of this type. We want our `Spec` to be sole place for them to configure everything about the resource.
+* Output Users: Users who refer to/query/use/extract information from the managed resource. We want them to be able to
+  perform all read-only operations on the given resource as if they have the cloud provider's API available. This can
+  include configurations and metadata information about the resource. We also want them to access only `Status`
+  sub-resource of the managed resource because this is where the most up-to-date information about the resource is
+  stored and we are able to limit RBAC of these users to access only `Status`.
```
negz: > We also want them to access only Status sub-resource of the managed resource because this is where the most up-to-date information about the resource is stored and we are able to limit RBAC of these users to access only Status.

Does this need updating? It's true that we can use RBAC to limit users to viewing the status, but we can also use RBAC to limit users to viewing (and not updating) the entire object.

design/one-pager-managed-resource-api-design.md
```diff
@@ -0,0 +1,398 @@
...
+However, the resources in the cloud provider are in the same environment, either per project (GCP) or same account (AWS).
+In order to provide a unique name that incorporates name and namespace, claim controller should make use of `GenerateName`
+feature of resource creation. The claim controller should assign `namespace-name` to `GenerateName` field of the `ObjectMeta`
+of the managed resource and Kubernetes will automatically append `-<5 characters random string>` to that when the creation
+is completed. For example, the following will be created by the claim controller:
+```yaml
+apiVersion: database.gcp.crossplane.io/v1alpha2
+kind: CloudsqlInstance
+metadata:
+  generateName: mydatabase-myappnamespace
```
negz: This looks like `name-namespace`, not `namespace-name` as you suggest earlier. I prefer `namespace-name`, because it leads with the least specific part of the identifier.

design/one-pager-managed-resource-api-design.md
```diff
@@ -0,0 +1,398 @@
...
+  namespace: gcp
+```
+After Kubernetes API server completes the creation, we'll see something like the following:
+```yaml
+apiVersion: database.gcp.crossplane.io/v1alpha2
+kind: CloudsqlInstance
+metadata:
+  name: mydatabase-myappnamespace-5sc8a
```
negz: Ditto here - should this be `myappnamespace-mydatabase-5sc8a`?

design/one-pager-managed-resource-api-design.md
```diff
@@ -0,0 +1,398 @@
...
+  namespace: gcp
+```
+`mydatabase` is the name of the `Claim` and `mynamespace` is the namespace that `Claim` lives in. That way we get a
```
negz: Should `mynamespace` be `myappnamespace` to match the above examples?

design/one-pager-managed-resource-api-design.md
```diff
@@ -0,0 +1,398 @@
...
+
+What goes into `Spec`:
+* Does the provider allow configuration of this parameter at _any_ point in the life cycle of the resource? Include if yes.
+  This includes the fields that are late-initialized, meaning some fields could be populated by the provider when you
+  actually create the resource but it also gives you the ability to change them. An example for this could be
+  auto-generated resource tags or resource-specific defaults. If the provider tags the resource without us telling them to
+  do so, controller should update `Spec` and input user should make changes on that current value of the field.
+  Related to https://github.com/crossplaneio/crossplane-runtime/issues/10
+
+Note that you should make updates to `Spec` fields that are empty. We do not override user's desired state and if they
```
negz: > Note that you should make updates to `Spec` fields that are empty.

Does this mean that controllers should only make updates to `Spec` fields that are empty, i.e. late initialization? It's a little ambiguous who "you" is.

design/one-pager-managed-resource-api-design.md
```diff
@@ -0,0 +1,402 @@
...
+- How should we differentiate the fields that are used by provider from Crossplane's fields?
+
+While this document is heavily opinionated, each provider can wildly vary between each other. So, treat this document as
+a starting point, try to stick to it for consistency but deviate if required and open an issue about why you had to
+deviate so that we can reassess and improve it.
+
+## Guiding Principles
+
+Our users are the stakeholders on how we make decisions. There are two types of users for managed resources:
+* Input Users: Users who configure/create/manage the resource. Crossplane Claim Controller, Cluster Admin are a few
```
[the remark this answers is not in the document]
negz: It could be that "configure" in this case means "write the YAML" (for humans) or "configure the managed resource" (for resource claim based dynamic provisioning). If that were the case, this would already be correctly ordered.

design/one-pager-managed-resource-api-design.md
muvaf: Good catch, removed that part.

muvaf:
@negz I think this is ready to merge!

design/one-pager-managed-resource-api-design.md
muvaf: Opened https://github.com/crossplaneio/crossplane/issues/925

Issue #497, opened 2019-05-28
Support "BYO bare metal" clusters


negz:
**Is this a bug report or feature request?** Feature Request

**What should the feature do:** Crossplane users with existing bare metal clusters that are not managed by Crossplane would like to be able to deploy workloads to said clusters. It may be possible to support this by manually creating a `KubernetesCluster` resource and its associated connection secret, representing the bare metal cluster. We may need to ensure our controllers don't attempt to act on these resources - perhaps by adding a `metal` "provisioner" (as opposed to EKSCluster, GKECluster, AKSCluster, etc).

**What is use case behind this feature:** One or two folks asked for this at Kubecon EU 2019. One use case I recall was using Crossplane in a lab environment to coordinate testing containerised distributed systems across many clusters.

jbw976:
I think this scenario of "BYO / bring your own" existing Kubernetes cluster may also have value beyond just bare metal.  Existing cloud managed clusters (GKE) may also be useful to import into Crossplane so that workloads can be scheduled to them.

negz:
Relates to (possibly a dupe of?) #409.

jbw976:
ah, i thought we had an issue covering this already, i couldn't find it though! I'm going to clean this up so we are tracking in only one place.

jbw976:
Closing as a dupe of #409

hasheddan:
Reopening this issue because I believed the BYO Kubernetes cluster scenario is somewhat isolated from the general importing of resources if we follow @negz design outline in [this comment](https://github.com/crossplaneio/crossplane/pull/841#discussion_r329227306).

Adding a Kubernetes `Provider` type would allow for usage of any Kubernetes cluster that a client could connect to.

negz:
Crosslinking to https://github.com/crossplaneio/crossplane/pull/1101.

negz:
I believe we now support this thanks to the `KubernetesTarget` work. :)

PR #841, opened 2019-09-23
One-Pager: K8s-Native Infrastructure Providers

### LINKED ISSUE #842: Investigate Rook-based managed resources
DoD: The Crossplane maintainers have approved a one pager scoping and suggesting the first steps toward Rook-based Crossplane managed resources.

https://rook.io/ is a Kubernetes operator that manages storage systems. It currently supports several, all of which (I believe) provide blob storage and/or a SQL database API. Rook has a close relationship with Crossplane as the projects share some founders. Crossplane may be able to make its first headway toward on premises external resources by targeting Rook. We'd like to produce a one pager investigating the feasibility of wrapping Rook as a way to spin up some or all of the storage/database services Rook supports.

hasheddan:

First pass at one-pager for k8s-native infrastructure providers motivated by the desire to add support for [Rook](https://github.com/rook/rook) in v0.4.

Fixes #842 

### Checklist

I have:


design/one-pager-k8s-native-providers.md
```diff
@@ -0,0 +1,217 @@
...
+## Technical Design
+
+### Claim Reconciler
+
+The claim controllers could use the same shared [claim reconciler] that other
+existing infrastructure stacks use currently, providing their own
+`ManagedConfigurators` to configure the managed resource instance using the
+referenced class.
+
+### Scheduler Reconciler
```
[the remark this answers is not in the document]
hasheddan: The purpose of the scheduler is to decide which remote cluster to use for creation of the "external resources". Much like `KubernetesApplication`, this will usually involve just getting the single cluster referenced using the `clusterSelector` field, but could involve round-robin scheduling if multiple clusters match selectors. I don't think we necessarily *need* this functionality, so I would be open to removing.

design/one-pager-k8s-native-providers.md
```diff
@@ -0,0 +1,217 @@
...
+
+Implementation of the shared managed reconciler for the managed CockroachDB
+`Cluster` controller (`cockroachdb.rook.crossplane.io/v1alpha1`) if a
+`nativeExternal` type was able to be utilized could look as follows:
+
+```go
+func (c *ClusterController) SetupWithManager(mgr ctrl.Manager) error {
+  r := resource.NewManagedReconciler(mgr,
+    resource.ManagedKind(v1alpha1.ClusterGroupVersionKind),
+    resource.WithExternalConnecter(&resource.NativeConnecter{client: mgr.GetClient(), kind: v1alpha1.Cluster}))
```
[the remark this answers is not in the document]
hasheddan: Yeah, I could see many of these just providing their own implementation, which they are still open to do of course. However, in the cases where it is possible this might be nice. We could certainly just push this kind of functionality down the road and see if it is useful / desired after multiple k8s-native infrastructure providers have done it manually.

design/one-pager-k8s-native-providers.md
```diff
@@ -0,0 +1,217 @@
...
+Though this design is motivated by the desire to support Rook, the pattern can
+be generalized to be applicable to other Kubernetes-native infrastructure
+providers. While most existing infrastructure provider stacks must create a
+client wrapper library to use the CRUD functions of the provider API, every
+Kubernetes-native infrastructure stack will use the [client-go] SDK for CRUD
+operations. Therefore, the pluggable `Observe`, `Create`, `Update`, `Delete`
+functions in the [crossplane-runtime] can likely have a common implementation
+for Kubernetes-native infrastructure stacks.
+
+This can be done via generic `NativeConnecter` and `nativeExternal` types in
```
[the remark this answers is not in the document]
hasheddan: I think the connector will be almost identical for all of these k8s-native providers, but I agree that the external functions will likely be much harder to generalize. I wonder if we could add a `resource.WithNativeExternalConnector()` function and just pass it an `external` object that satisfies the external object but just automatically gets the k8s client injected into it from a shared connector implementation. All of these are sort of "nice-to-have" features, rather than imperative for functionality.

design/one-pager-k8s-native-providers.md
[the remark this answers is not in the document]
muvaf: But we have only 1 remote cluster that we get the kubeconfig of during the connector phase. I am not sure we want to support more than one Rook cluster that we can provision resources on at this phase. The current support for that kind of setup is simply having another `Provider` resource that resource classes refer to statically.

design/one-pager-k8s-native-providers.md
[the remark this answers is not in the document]
muvaf: I think I don't know how many k8s-native providers we can have currently or plan to have. That's why I'm worried that we add a generic functionality to crossplane-runtime that is used by one provider. If we were to use GCP's config connector and AWS service operator, I'd definitely go for it though.

design/one-pager-k8s-native-providers.md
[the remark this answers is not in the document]
hasheddan: Important note here is that the managed reconciler would not accept a resource that had not already been scheduled, so by the time the resource gets to the managed reconciler, there will only be one kubeconfig that it needs to obtain in the connector phase.

design/one-pager-k8s-native-providers.md
```diff
@@ -0,0 +1,217 @@
...
+`crossplane-runtime`. Similar to the `clusterConnecter` in the
+`KubernetesApplicationResource` controller, it would get the cluster connection
+information of the Kubernetes cluster that the managed resource was scheduled to
+by the scheduler controller, then return a `nativeExternal` type that implements
+the `ExternalClient` interface. Because external client functions will
+essentially be making sure the "external resource" in the target cluster matches
+the "managed resource" (i.e. they should be nearly identical Kubernetes
+resources), it should be possible to have generic implementations of the
+`ExternalClient` functions. However, if it proves difficult to do so, or it just
```
negz: All current implementations of `resource.ExternalClient` type assert the `resource.Managed` they're supplied with to a concrete type. This is primarily because each type we deal with has different `Spec` and `Status` structs that we must translate to cloud provider API types.

Do you imagine that (for example) `cockroachdb.rook.crossplane.io/v1alpha1` would be a strict superset of `cockroachdb.rook.io/v1alpha1` to the extent we could have the `nativeExternal` `resource.ExternalClient` treat the "external resources" as `*unstructured.Unstructured` or similar? If so, do you have an idea of how we'd assess the readiness of such external resources so that we could update our `BindingPhase` and `Ready` condition? `KubernetesApplicationResource` basically punts on this by calling the opaque external resource `Submitted` rather than `Ready`.

design/one-pager-k8s-native-providers.md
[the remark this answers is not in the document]
negz: I'm a little ambivalent around whether we need a scheduler here. It's useful for all the reasons being able to schedule a workload is useful, but I lean slightly toward the simpler and faster path that omits it to begin with.

design/one-pager-k8s-native-providers.md
```diff
@@ -0,0 +1,217 @@
...
+## Workflow Design
+
+The general design from a user experience perspective should differ very little
+from utilizing any other infrastructure provider stack. Developers should be
+able to request the creation of an abstract resource type via a claim, and
+cluster operators should be able to define the classes of service available to
+satisfy that claim kind.
+
+As an example, Rook provides the option to run [CockroachDB] in a Kubernetes
+cluster. Assuming you have installed Rook and the corresponding CockroachDB
```
jbw976: what do you think about scenarios where you have a target cluster running, but you don't actually have the operator running there yet?  I'm wondering if that would be something we can support or have knowledge of.  If we want to deploy a resource to a target cluster, but it's operator isn't yet running, can we automatically install the operator first so that it is then ready to handle any CRDs that we propagate to the target cluster it's running in?

design/one-pager-k8s-native-providers.md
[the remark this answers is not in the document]
jbw976: if there's no scheduling capability, how would the reconciler know what target cluster to propagate the resource to?  Say there are 2 `KubernetesCluster` resources known to Crossplane...it has to pick one somehow, does it?  Perhaps the naive `clusterSelector` that @hasheddan mentioned is a cheap initial solution to this.

design/one-pager-k8s-native-providers.md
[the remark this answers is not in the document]
jbw976: I think the primary scenario here is operators, not cloud provider declarative k8s APIs.  There are TONS of operators out there now and being able to integrate with the general scenario of the CRD API that is backed by operators is pretty compelling.  It completely opens up on-premises scenarios for being managed by Crossplane.

Likely, we'd end up with a generalized solution in the first pass here because Rook itself is 7+ different operators.  If we want to do S3 and Postgres types then that covers CockroachDB, Yugabyte, Ceph, EdgeFS, and Minio, or essentially 5 different operators.  Sounds like there could already be a case for generalizing this in the first pass :)

design/one-pager-k8s-native-providers.md
```diff
@@ -0,0 +1,217 @@
...
+  writeConnectionSecretToRef:
+    name: postgresqlconn
+  engineVersion: "9.6"
+```
+
+On creation of the `PostgreSQLInstance` claim, the claim controller would
+configure and create a `Cluster` (`cockroachdb.rook.crossplane.io/v1alpha1`)
+instance. Then, the managed controller would use the Crossplane `Cluster`
+instance to create a `Cluster` (`cockroachdb.rook.io/v1alpha1`) in the target
+Kubernetes cluster where Rook and the CockroachDB operator are installed. This
```
jbw976: I'm thinking that for on-premises scenarios, having "import" or "bring your own" functionality would be really useful, i.e. an existing cluster already running on-premises being imported into Crossplane so that it can be used as a target cluster for scheduling.

Being able to deploy resources that are managed by operators to target k8s clusters is still useful in managed k8s clusters that we first spin up in a cloud provider using Crossplane, but it may be more useful for true on-premises scenarios instead.

This is tracked in https://github.com/crossplaneio/crossplane-runtime/issues/22.

design/one-pager-k8s-native-providers.md
```diff
@@ -0,0 +1,217 @@
...
+workflow is similar to dynamic provisioning in a typical infrastructure
+provider:
+
+![K8s Native Providers](./images/k8s-native-providers.png)
```
jbw976: this example for CockroachDB is a bit hard to follow along with because of its type name `Cluster`. I'm confusing that with target k8s cluster a bit, but there may not be much we can do about that.

design/one-pager-k8s-native-providers.md
```diff
@@ -0,0 +1,217 @@
...
+
+## Workflow Design
+
+The general design from a user experience perspective should differ very little
+from utilizing any other infrastructure provider stack. Developers should be
+able to request the creation of an abstract resource type via a claim, and
+cluster operators should be able to define the classes of service available to
+satisfy that claim kind.
+
+As an example, Rook provides the option to run [CockroachDB] in a Kubernetes
```
jbw976: I think longer term the REALLY killer scenario is going to be deploying globally distributed solutions across multiple target clusters.  For example, deploying 1 CockroachDB node to `west-on-prem`, 1 to `east-on-prem`, and maybe a 3rd one to `AWS us-east-1`.  Having a controller at a high level that understands CockroachDB or Yugabyte failure zones, topology, data locality, etc. and can deploy a globally distributed cluster across multiple sites/regions with a single command from the user would be EPIC 🙇 

Out of scope for this 1st pass, but it's something I've been wanting for a long time and would take Crossplane to the next level! 🚀

design/one-pager-k8s-native-providers.md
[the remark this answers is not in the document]
prasek: simpler, faster path 👍

design/one-pager-k8s-native-providers.md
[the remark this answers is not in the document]
prasek: > All of these are sort of "nice-to-have" features, rather than imperative for functionality.
Could we skip the "nice-to-have" features for now and circle back if the need arises?

design/one-pager-k8s-native-providers.md
hasheddan: In the general case, we could just check that the resource exists in the remote cluster. However, if the external client was manually implemented then we could do more intelligent checking of any `Status` fields that the "provider" surfaces on their resources.

design/one-pager-k8s-native-providers.md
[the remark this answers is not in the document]
hasheddan: @jbw976 I guess the alternative would be to have an object reference to a `KubernetesCluster` that we set directly, rather than labels that the scheduler uses to find and set a reference to a cluster.

design/one-pager-k8s-native-providers.md
hasheddan: We certainly could! It would just need to be built into the `Create()` method of the managed reconciler for the resource I believe.

design/one-pager-k8s-native-providers.md
hasheddan: Agreed. I will use a different Rook example for clarity!

design/one-pager-k8s-native-providers.md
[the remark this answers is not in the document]
negz: I don't know that we should take it for granted that we want to target `KubernetesCluster` here. Remember that every other managed resource uses a `Provider` for this case, and that we have no BYO `KubernetesCluster` or "use the `KubernetesCluster` you're running on" story.

design/one-pager-k8s-native-providers.md
negz: I like this idea, but would suggesting starting by targeting a pre-installed operator then add this in later. I think if we keep this narrowly scoped to begin with there's a good chance this functionality (1-2 Rook based managed resources) could ship super soon.

design/one-pager-k8s-native-providers.md
```diff
@@ -0,0 +1,271 @@
...
+
+### Managed Reconciler
+
+Kubernetes-native infrastructure providers can make use of the shared managed
+reconciler in `crossplane-runtime` in the same manner as traditional
+infrastructure providers. However, instead of creating a new client for a cloud
+provider API, Kubernetes-native infrastructure providers will return a
+Kubernetes client that is configured to talk to the target cluster using the
+`targetClusterRef`. This is similar to the client configuration for the
+`KubernetesApplicationResource` [managed reconciler].
```
negz: Does this mean these providers use a `KubernetesCluster` claim rather than a `Provider`? How will we solve the cases mentioned in https://github.com/crossplaneio/crossplane/pull/841#pullrequestreview-293432522?

hasheddan: Yes, the current implementation here suggests using a `KubernetesCluster` ref and deferring that effort to #497. However, implementing a new `Provider` would fix these issues for Rook and be more in-line with how existing infrastructure stacks operate. Do you think that is the appropriate direction here?

design/one-pager-k8s-native-providers.md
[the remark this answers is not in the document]
muvaf: I think if we don't have a good reason for a scheduler between `Provider` objects in other infra-stacks, we don't have one for Rook, too.

> Remember that every other managed resource uses a Provider for this case

@negz As far as I see, the only difference for Rook should be about the content of that `Provider` object. It will be a kubeconfig instead of username/password or a json. If the user wants to use `KubernetesCluster` as a Rook cluster, they can just put its kubeconfig in a `Rook Provider` object and it'd work. I agree that we don't need to know whether it's `KubernetesCluster` or some cluster we have no idea of or even control cluster itself.

> if there's no scheduling capability, how would the reconciler know what target cluster to propagate the resource to?

@jbw976 Rook Provider object that is referred as `providerRef` in the resource would have the right kubeconfig to provision that resource.

muvaf:
IMO, using a `provider.rook.crossplane.io` or a `provider.kubernetes.crossplane.io` as the reference(`providerRef`) of the cluster that we're trying to provision resources in covers many scenarios for free compared to `targetClusterRef *KubernetesCluster` idea.

About making k8s API interaction generic, I think I agree with @negz; let's first implement one or two resources just like any other external client and then come back and assess how easy and valuable it'd be to have a generic one. I believe the time that you'd need for implementation of generic k8s client will also decrease after you already implemented a specific client for one resource.

design/one-pager-k8s-native-providers.md
```diff
@@ -0,0 +1,287 @@
...
+Kubernetes credentials using a `Provider` object rather than a
+`KubernetesClusterObject`. This `Provider` type
+(`provider.kubernetes.crossplane.io`) should likely exist outside of the Rook
+stack (i.e. in core Crossplane) due to its generic applicability. 
+
+## Future Considerations
+
+### More Sophisticated Scheduling
+
+This initial proposal embeds a `proivderRef` field in the non-portable classes
```
negz:
```suggestion
This initial proposal embeds a `providerRef` field in the non-portable classes
```

design/one-pager-k8s-native-providers.md
```diff
@@ -0,0 +1,287 @@
...
+  namespace: rook-infra-dev
+type: Opaque
+data:
+  endpoint: MY_K8S_ENDPOINT
+  username: MY_K8S_USERNAME
+  password: MY_K8S_PASSWORD
+  clusterCA: MY_K8S_CLUSTER_CA
+  clientCert: MY_K8S_CLIENT_CERT
+  clientKey: MY_K8S_CLIENT_KEY
+  token: MY_K8S_TOKEN
```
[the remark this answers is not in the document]
negz: > This pattern almost makes me wonder whether Kubernetes workloads (KubernetesApplication and friends) should be targeting a Kubernetes Provider rather than a KubernetesCluster directly. Doing so would effectively solve the "BYO cluster" scenario for Kubernetes workloads without having to support "importing" a Kubernetes cluster as a literal KubernetesCluster resource claim.

This isn't strictly relevant to this design, but I'd like to capture it somewhere. Not sure if it belongs in the future considerations of this document or as a separate issue. 🤔

design/one-pager-k8s-native-providers.md
```diff
@@ -0,0 +1,287 @@
...
+---
+apiVersion: kubernetes.crossplane.io/v1alpha2
+kind: Provider
+metadata:
+  name: demo-kubernetes
+  namespace: azure-infra-dev
+spec:
+  credentialsSecretRef:
+    name: demo-provider-rook
+```
```
negz: I wonder whether we should support a parameter like `thisCluster: true` on the Kubernetes Provider. This flag would instruct any connection logic to use [in cluster config](https://github.com/kubernetes/client-go/tree/95884bf844a94d573a28c0136282f4f8d0d4b8ae/examples/in-cluster-client-configuration), bypassing the need to read in a Secret.

Another (better?) approach could be to add a `credentialsServiceAccountRef` field to the spec, which would configure a (Kubernetes) service account whose bearer token should be used to provision managed resources in the local cluster, allowing you to use a different service account for each Kubernetes provider from the one that Crossplane is running as.

Another (maybe even better?) approach could be to just use `credentialsSecretRef` as we do today, but ensure our connection logic handles the case in which the supplied credentials secret is a Kubernetes [service account secret](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#service-account-tokens), not a Crossplane `KubernetesCluster` secret, i.e.:

```yaml
data:
  ca.crt: (APISERVER'S CA BASE64 ENCODED)
  namespace: ZGVmYXVsdA==
  token: (BEARER TOKEN BASE64 ENCODED)
```

This enables the following user flows.

Managing Rook resources in a Crossplane provisioned external cluster:
1. Provision a `KubernetesCluster` resource claim.
1. Create a Kubernetes `Provider` whose `credentialsSecretRef` references the `KubernetesCluster` `writeConnectionSecretToRef`.
1. Provision your Rook resource using the aforementioned `Provider`.

Managing Rook resources in an existing external cluster:
1. Create a Kubernetes `Provider` whose `credentialsSecretRef` references a hand curated `Secret` using the same format as that of a `KubernetesCluster` claim.
1.  Provision your Rook resource using the aforementioned `Provider`

Managing Rook resources in the Crossplane control cluster:
1. Create a Kubernetes `ServiceAccount` and grant it the RBAC permissions required to do Rook things.
1. Create a Kubernetes `Provider` whose `credentialsSecretRef` references the aforementioned `ServiceAccount` token.
1. Provision your Rook resource using the aforementioned `Provider`.

negz: @hasheddan If you're on board with the above (my preference is making the connection logic that parses a `credentialsSecretRef` also handle Service Account secrets) it would be good to capture that in this design.

design/one-pager-k8s-native-providers.md
```diff
@@ -0,0 +1,287 @@
...
+cluster operators should be able to define the classes of service available to
+satisfy that claim kind.
+
+As an example, Rook provides the option to provision [Minio] object storage in a
+Kubernetes cluster. Assuming you have installed Rook and the corresponding Minio
+operator into a cluster, you should be able to dynamically provision a Minio
+object storage cluster into the Kubernetes cluster from your Crossplane control
+cluster. The experience would look as follows:
+
+1. Create a Kubernetes `Provider` for target cluster:
```
jbw976: i wonder if there may be a nice crossplane CLI command that could make this easy, using the locally available kubeconfig contexts.  e.g.:

```
kubectl crossplane create provider --namespace=rook-infra-dev --kubectx=on-prem-1
```

design/one-pager-k8s-native-providers.md
```diff
@@ -0,0 +1,287 @@
...
+  name: bucket-standard
+  namespace: app-project1-dev
+classRef:
+  kind: ObjectStoreClass
+  apiVersion: minio.rook.crossplane.io/v1alpha1
+  name: rook-minio
+  namespace: rook-infra-dev
+```
+
+4. Create a `Bucket` claim
```
jbw976: so an important distinction to make here is that the `ObjectStoreClass` that this `Bucket` claim will use via `Bucket` -> `BucketClass` -> `ObjectStoreClass` is actually the entire underlying object storage solution, not just a single bucket.  The analogy here for AWS is the entire S3 service.  Furthermore, when you create a Minio object store, it doesn't start with any buckets at all.  As written, this claim -> class relationship would result in the user requesting a bucket but instead getting a blank object store with 0 buckets.

We need to think through a bit more on how the user requesting a Bucket can result in a bucket getting created for them, and how to deal with the cases where the Minio object store (or any of the other object storage providers in Rook) does not exist yet and already exists.

Steps to create a Minio object store:
1. Deploy the Minio operator
1. Create an instance of an objectstore.minio.rook.io
1. Create a bucket (there is currently no CRD for this) in that object store

negz: Thanks for catching this @jbw976 - it seems like it could expand the scope of this work, at least a little. A bunch of questions come to mind

Am I correct in my understanding that this concern only applies to Rook's object storage types, and not the SQL "instance" types? It seems like deploying CockroachDB or Yugabyte with Rook gets you similar functionality to deploying a CloudSQL instance; i.e. a Postgres-compliant API you could use to create and manage databases. So it seems to make sense that a deployment of CockroachDB, for example, would satisfy a Crossplane `PostgreSQLInstance` claim.

Is there some way we could reuse or extend the `S3Bucket.storage.aws.crossplane.io` managed resource for use with S3 compatible object stores? Perhaps it would be better to duplicate it as `S3Bucket.storage.rook.crossplane.io` alongside the existing, especially given folks may not want to install `stack-aws` just to be able to use buckets with `stack-rook`?

How would deployment of an object store API (e.g. Minio) fit into Crossplane's model? Given Rook provides us with multiple ways to deploy an S3-compatible object store would it make sense to add an `ObjectStore` resource claim type that could eventually be satisfied by Minio, EdgeFS, or Ceph managed resources?

What would the user experience be, assuming we were had already deployed Rook (but no storage providers) and configured a Kubernetes provider appropriately? Something like:

1. Create a `MinioClass`, point an `ObjectStoreClass` at it.
1. Create a `ObjectStore` claim, point it at the `ObjectStoreClass` from step 1.

Then, for each bucket we wanted to create:
1. Create a `S3BucketClass` non-portable class, point a `BucketClass` at it.
1. Create a `Bucket` claim, point it at the `BucketClass` from step 1.

Would we need to configure some kind of `provider.s3.rook.crossplane.io` to actually connect to and use the API exposed by the Rook object storage providers? Would we need to explicitly create a user in order to get credentials to add that provider?

design/one-pager-k8s-native-providers.md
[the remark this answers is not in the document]
hasheddan: I think it should be called out in the future considerations, and also tracked by #497. I will add a comment to this effect on the issue.

design/one-pager-k8s-native-providers.md
hasheddan: @jbw976 thanks for catching this, I was much too hasty in translating the doc from CockroachDB to Minio, so I appreciate your watchful eye! I broadly agree with @negz thoughts above, although I am hesitant to add a generic claim type that is only used by one provider. It seems that it may be an opportunity to introduce a "provider-specific" claim for `ObjectStore`. However, maybe the generic `ObjectStore` claim does make sense because Rook is basically just an API for many other in-cluster storage providers.

design/one-pager-k8s-native-providers.md
hasheddan: Really big fan of this design :) Will make sure to update the doc to reflect it.

design/one-pager-k8s-native-providers.md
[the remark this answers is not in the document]
muvaf: > I'm a little rusty on how our existing KubernetesCluster connection secret format looks

I second this. Today, if I have a k8s cluster that I want to import or use as `Provider`,  I have to be _really_ careful about what goes to where. By _really_, I mean I couldn't do it. I believe using the kubeconfig file directly as value of 1 field in the secret makes manual interaction easier. Our go k8s clients have support for that format and it's easy to use the client to do the parsing when fine-grained field extraction is needed.

> This pattern almost makes me wonder whether Kubernetes workloads (KubernetesApplication and friends) should be targeting a Kubernetes Provider rather than a KubernetesCluster directly.

I see that this makes things easier but it sounds to me that we'd be bending the definition of `Provider` too much to adopt this usage. One can manually refer to the provider's secret during k8s cluster import if they wanted to do that.

design/one-pager-k8s-native-providers.md
```diff
@@ -0,0 +1,287 @@
...
+data:
+  endpoint: MY_K8S_ENDPOINT
+  username: MY_K8S_USERNAME
+  password: MY_K8S_PASSWORD
+  clusterCA: MY_K8S_CLUSTER_CA
+  clientCert: MY_K8S_CLIENT_CERT
+  clientKey: MY_K8S_CLIENT_KEY
+  token: MY_K8S_TOKEN
+---
+apiVersion: kubernetes.crossplane.io/v1alpha2
```
muvaf: I like that we're using generic kubernetes provider type instead of something rook-specific.

hasheddan: Implemented due to your and @negz insightful feedback :)

design/one-pager-k8s-native-providers.md
```diff
@@ -0,0 +1,692 @@
...
+  name: demo-k8s-provider
+  namespace: my-infra-dev
+spec:
+  credentialsSecretRef:
+    # Get secret name from service account
+    name: my-service-account-secret
+```
+3. Provision your Kubernetes resource using the aforementioned `Provider`.
+
+## Proposed Workflow for `KubernetesApplication` Resources
```
negz: Detailed discussion of changes to `KubernetesApplication` seems a little out of scope given the goals of this design. I wonder whether we should capture this separately - perhaps cut and paste the `KubernetesApplication` bits into a Github issue for us to follow up at some point after this design has been ratified and implemented?

hasheddan: Done! #859

design/one-pager-k8s-native-providers.md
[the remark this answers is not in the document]
negz: Discussion of scheduling has moved to future considerations, so I'm going to resolve this thread.

design/one-pager-k8s-native-providers.md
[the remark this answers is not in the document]
negz: I believe the current doc is aligned on tackling Rook specifically to begin with, so I'm going to resolve this thread.

design/one-pager-k8s-native-providers.md
```diff
@@ -0,0 +1,217 @@
...
+* Reviewers: Crossplane Maintainers
+* Status: Draft
+
+## Introduction
+
+Currently, all infrastructure provider stacks manage resources by creating an
+external resource and keeping it in sync with an instance of a CRD (managed
+resource) in the Crossplane control cluster. CRUD operations are executed on
+external resources using the API provided by the infrastructure provider.
+However, there are a number of infrastructure providers, with more being created
```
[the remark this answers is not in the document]
negz: > As opposed to provisioning resources externally on
provider-owned infrastructure, a Kubernetes-native provider provisions resources
in-cluster, meaning that they can run on-premises or anywhere else you may be
running a Kubernetes cluster.

I think this new wording addresses this comment, so I'm going to resolve the thread. Please reopen if you feel it needs more work!

design/one-pager-k8s-native-providers.md
negz: This idea is now captured under future considerations, so I'm going to resolve this thread.

design/one-pager-k8s-native-providers.md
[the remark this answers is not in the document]
negz: Given that under this proposal we'll be expecting some users to write Kubernetes `Provider` secrets by hand, I agree with @muvaf that "create a secret whose key encodes your kubeconfig file" feels simpler than "create a secret whose keys match this Crossplane-specific format".

That said, our current format is a little closer to the Kubernetes service account format, and I'm hesitant to predicate this work on changing how KubernetesCluster connection secrets are formed. I'm also unsure whether it would be a good idea to align on kubeconfig based secrets given that moves us further from compatibility with the format used by service accounts.

Perhaps we should:

1. Solve this with documentation to begin with - i.e. make sure we thoroughly document the expected Kubernetes provider secret format.
1. Raise an issue to track potential improvements?

What do you think @hasheddan?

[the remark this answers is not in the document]
hasheddan: Sounds like a good plan! I will open an issue to track now, and I will make sure that this is reflected in documentation when Kubernetes `Provider` is implemented.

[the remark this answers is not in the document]
hasheddan: #861 created to track potential improvements

design/one-pager-k8s-native-providers.md
negz: @jbw976 Do you see Rook being involved in such a direction, or this being more the role of Crossplane-native CockroachDB and Yugabyte stacks?

design/one-pager-k8s-native-providers.md
```diff
@@ -0,0 +1,562 @@
...
+![K8s Native Providers](./images/k8s-native-providers.png)
+
+### Non-Portable Resources
+
+Some resources exposed by Kubernetes-native infrastructure providers may not be
+initially supported by dynamic provisioning. This is true for existing
+infrastructure stacks as well, as types like `VirtualNetwork` in Azure can be
+provisioned statically, but not in the dynamic `claim <-> class` pattern. These
+types of resources may eventually be supported by portable claim types in core
+Crossplane.
```
negz: I think there's something more subtle going on here than support for dynamic provisioning. Specifically:

* You can't claim these managed resources.
* You can't dynamically provision these managed resources, because dynamic provisioning is triggered by a claim.

This raises the question "what is the purpose of a claim?" Is it just a request to dynamically provision a managed resource, or is it a request to "claim" a managed resource in the sense of claiming an exclusive right to its use, indicating that it's no longer available to satisfy other claims.

In the Kubernetes PV/PVC pattern we're modeled on you can't (easily) use a PersistentVolume without claiming it. So whether a volume is provisioned dynamically or statically, you must always claim it. I think we must follow this pattern for resource kinds that _can_ be claimed (e.g. `CloudSQLInstance`). If we were to start using them without ever claiming them there would be no way for a `MySQLInstance` claim to distinguish whether an unclaimed `CloudSQLInstance` was or was not actually in use. The lack of a claim binding is taken to be a signal that the `CloudSQLInstance` is available for use.

We're kind of punting on what to do about these managed resource kinds that can't be claimed for the time being. I'm not sure there's an actual action to be taken in the context of this design doc, but I wanted to communicate that I think of these managed resources as more in some kind of "unclaimable danger zone" rather than simply "not dynamically provisionable".

design/one-pager-k8s-native-providers.md
```diff
@@ -0,0 +1,562 @@
...
+dynamic provisioning for this resource type in the initial implementation.
+
+## Technical Design
+
+### The Kubernetes Provider Kind
+
+The Kubernetes `Provider` kind will be implemented in core Crossplane in a new
+API group: `kubernetes.crossplane.io`. This will allow it to be used by core
+Crossplane custom resources as well as Kubernetes-native infrastructure stacks
+like Rook.
```
negz: I think this makes sense for now, but I wonder if one day the Kubernetes provider and workloads move into `stack-kubernetes`. 🤔

design/one-pager-k8s-native-providers.md
```diff
@@ -0,0 +1,562 @@
...
+Kubernetes client that is configured to talk to the target cluster using the
+`providerRef` to a Crossplane Kubernetes `Provider` object. 
+
+## Future Considerations
+
+### Provisioning Buckets in Rook `ObjectStores`
+
+The ability to provision buckets in the `ObjectStore` options provided by Rook
+will need to be supported in the future. Three main options to implement this
+functionality include waiting until the functionality natively added to Rook,
```
negz:
```suggestion
functionality include waiting until the functionality is natively added to Rook,
```

design/one-pager-k8s-native-providers.md
```diff
@@ -0,0 +1,562 @@
...
+   dynamically using a `Bucket` claim type or the Minio `Bucket` kind directly.
+
+Though the Minio stack strategy is somewhat cumbersome in that multiple
+Crossplane stacks must be installed in order to provision consumable object
+storage using Rook, it is more in-line with the structure of the Crossplane
+stack isolation model. In contrast, using controller logic in the Rook stack to
+provision buckets would be a divergence from our desire to model provider APIs
+in a high-fidelity manner. **For this reason, either waiting for support to be
+natively provided by Rook or creating a separate Minio stack appears to be the
+sensible direction forward.**
```
negz: 👍 I feel most on board with the minio-stack direction at this stage.

design/one-pager-k8s-native-providers.md
```diff
@@ -0,0 +1,562 @@
...
+
+### Automatic Operator Installation
+
+For any Kubernetes-native provider, there must be an operator installed in the
+target cluster for the resources to be provisioned. However, it would be nice to
+be able to provision resources in a cluster where the operator is not currently
+installed. This could be implemented by checking if the operator exists in the
+target cluster in the managed reconciler's `Create` method, and installing it if
+not. This may not be part of initial implementation, but should definitely be
+tracked and supported.
```
negz: This almost makes me wonder whether some kind of distinct, optional, run-once `Bootstrap()` method would make sense in `resource.ExternalClient`, but I think that's something to ponder another day.

design/one-pager-k8s-native-providers.md
hasheddan: I actually would be in favor of keeping it in core Crossplane as I see it as the appropriate place for resources that may be utilized my multiple stacks.

design/one-pager-k8s-native-providers.md
hasheddan: I like this pattern, updated crossplaneio/crossplane-runtime#34 with reference

design/one-pager-k8s-native-providers.md
hasheddan: Yep! Fixed :)

design/one-pager-k8s-native-providers.md
hasheddan: I definitely understand the concern here, but one thing to keep in mind is that most of these resources (e.g. virtual networks, subnets, object stores, etc.) are resources that could be "used" by many different apps / services / etc. For instance, I may want to use Crossplane to manage multiple MySQL database instances, each for a different application, that all exist in the same network. If I can't provision resources in a network unless I can claim an unbound `VirtualNetwork`, then I can only use that `VirtualNetwork` for a single application. This differs from something like a `CloudSQLInstance` which is likely only used by a single application. One way we could potentially allow for claiming these currently "unclaimable" resources would be to allow multiple claims to bind to them. However, this would also be divorced from the PV/PVC pattern, so I am not sure this is an improvement.

negz: Perhaps "exclusive right to use" was not the right wording for what a claim represents. In the case of persistent volume claims it's more a sign that "it is being used". There is a strict 1:1 PV:PVC relationship, but I don't think there is a strict 1:1 PVC:Pod relationship.

I'm not really making the case that we need something like a `NetworkClaim` so much as pointing out that we should probably form a stance on these "unclaimable" resources before they proliferate further, and that agreeing on what exactly the point of a claim is (is it simply a request for dynamic provisioning?) could help do that.

jbw976: Just FYI, the way that PVCs express their intended consumption is with an `accessModes` field on their spec, e.g.:
```
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: mypvc
spec:
  accessModes:
  - ReadWriteOnce
```

If multiple consumers are allowed/expected, then the accessModes value would be `ReadWriteMany`.

PR #844, opened 2019-09-24
Move crds directory into templates

### LINKED ISSUE #839: Support Helm 3

Helm 3 brought some breaking changes on how Helm works and what flags its CLI accepts.

When you fix your CLI flags to the new ones a few problems arise:
* Our helm chart does not have `crossplane-system` namespace, hence we're not able to do `helm3 install crossplane -n crossplane-system` because that behavior is identical to how kubectl namespace works now, i.e. you get `namespaca is not found`.
* Just raw `helm3 install crossplane crossplane-alpha/crossplane --version 0.3.0` doesn't work, it returns `Error: customresourcedefinitions.apiextensions.k8s.io "kubernetesapplicationresources.workload.crossplane.io" already exists` on a fresh cluster. The actual resource changes randomly.

I haven't invested so much time into this, these are my initial observations. Helm3 is still in beta phase, so, not sure when we want to support it. Wanted to track this here.

muvaf:

Moves `crds` folder of helm chart into templates folder for compatibility with helm 3 and adds note to documentation about using helm 3.

Currently, we have `crds` folder that Helm2 does not recognize, however, `templates/crds.yaml` renders that into `templates` folder so we are able to install them. In Helm 3, though, `crds` folder is reserved. So, when we do `helm3 install`, we try to crd-install `crds` folder first(See https://github.com/helm/helm/issues/5871 and the merged PR https://github.com/helm/helm/pull/6243) and then normal install the crds that are rendered into `templates` folder. That's why helm3 install results in `x crd already exists` error.

We can just use `crds` folder and delete `templates/crds.yaml` but that would result in breaking Helm2 support. The solution that keeps both working is moving `crds` folder into templates. 

This doesn't make use of the new crd-install mechanism in Helm 3, but, the main problem that they try to solve with that solution is that when you have CRs in `templates` folder, you have to have an ordered installation that installs CRDs first. So, we don't need that part since we don't install any CR instance. The caveat there is that we keep treating all CRDs just like normal resources that will be deleted when release is deleted, which is a bit dangerous since deleting a CRD results in deleting all CR instances in all namespaces.

This PR does not upgrade our build system or minikube scripts, it only allows people to use helm 3 to install crossplane. We will likely update everything after Helm 3 has got a stable version.

See https://github.com/helm/helm/issues/5892 for details about namespaces.

Fixes #839 
### Checklist

I have:


muvaf:
This PR adds that actually by removing the crds folder before generation.

docs/install-crossplane.md
```diff
@@ -14,13 +14,16 @@ The Helm chart contains all the custom resources and controllers needed to deplo
...
 * [Helm](https://docs.helm.sh/using_helm/), minimum version `v2.9.1+`.
-  * Make sure Helm is initialized with sufficient permissions to work on `crossplane-system` namespace.
+  * For Helm 2, make sure Tiller is initialized with sufficient permissions to work on `crossplane-system` namespace.
 
 ## Installation
 
 Helm charts for Crossplane are currently published to the `alpha` and `master` channels.
 In the future, `beta` and `stable` will also be available.
 
+> If you're using Helm 3, make sure you create `crossplane-system` namespace (or the namespace you'd like to install
```
negz:
```suggestion
> If you're using Helm 3, make sure you create the `crossplane-system` namespace (or the namespace in which you'd like to install
```

cluster/charts/crossplane/templates/crds/workload.crossplane.io_kubernetesapplicationresources.yaml
```diff
@@ -22,7 +22,9 @@ spec:
   group: workload.crossplane.io
   names:
     kind: KubernetesApplicationResource
+    listKind: KubernetesApplicationResourceList
```
jbw976: that's odd that these workload CRDs didn't seem to get these listKind/singular additions that all the other CRDs got in v0.3.  Looks like we may have had the wrong directory name in the controller-gen calls in our Makefile?

Makefile
```diff
@@ -87,13 +87,14 @@ go.test.unit: $(KUBEBUILDER)
...
-	go run vendor/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go crd:trivialVersions=true paths=./apis/compute/... output:dir=cluster/charts/crossplane/crds
-	go run vendor/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go crd:trivialVersions=true paths=./apis/core/... output:dir=cluster/charts/crossplane/crds
-	go run vendor/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go crd:trivialVersions=true paths=./apis/database/... output:dir=cluster/charts/crossplane/crds
-	go run vendor/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go crd:trivialVersions=true paths=./apis/storage/... output:dir=cluster/charts/crossplane/crds
-	go run vendor/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go crd:trivialVersions=true paths=./apis/workloads/... output:dir=cluster/charts/crossplane/crds
-	go run vendor/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go crd:maxDescLen=0,trivialVersions=true paths=./apis/stacks/... output:dir=cluster/charts/crossplane/crds
+	rm -rf cluster/charts/crossplane/templates/crds
```
jbw976: do you want to preface this line with `@` so that it will not be output during `make`?

Makefile
```diff
@@ -87,13 +87,14 @@ go.test.unit: $(KUBEBUILDER)
...
+	go run vendor/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go crd:trivialVersions=true paths=./apis/cache/... output:dir=cluster/charts/crossplane/templates/crds
```
jbw976: I looked in the helm migration issue https://github.com/helm/helm/issues/5892 quickly to see where it talked about this requirement to move the CRDs to under the templates dir, but I couldn't find it quickly.  Can you help point me to it?

cluster/charts/crossplane/templates/crds/workload.crossplane.io_kubernetesapplicationresources.yaml
muvaf: Yes, that is the case unfortunately. Indeed, we kept registering `Policy` types as well since the generation did not remove anything but only add/replace.

Makefile
muvaf: I dug more and added more context about this to the description of the PR.

One can possibly add an if/else statement in `crds.yaml` to decide whether to copy crds folder. However, I couldn't find a good way to get helm version in the yaml. The closest is `{{if not Capabilities.TillerVersion }}` but that results in error in Helm3 saying `Error: parse error at (crossplane/templates/crds.yaml:1): function "Capabilities" not defined`

jbw976: Cool, from the PR description you added, it sounds like a good solution currently that works for both helm 2 and helm 3 is to just use the templates folder.  Thanks for clarifying all that!

turkenh:
> This doesn't make use of the new crd-install mechanism in Helm 3, but, the main problem that they try to solve with that solution is that when you have CRs in templates folder, you have to have an ordered installation that installs CRDs first. So, we don't need that part since we don't install any CR instance.

@muvaf with https://github.com/crossplaneio/crossplane/pull/932 this is no longer the case. Could you share some info on the new crd-install mechanism in Helm 3? Can we keep both helm 2 & helm 3 working also with CR instances?

muvaf:
@turkenh The new mechanism is that you have a folder called `crds` alongside `templates` folder in your helm chart, not inside the `templates` folder. You can read the detailed explanation here https://github.com/helm/helm/issues/5871

In short, moving CRD YAMLs to `crds` folder does not work with Helm2 because it doesn't scan that folder. If we render it via a template(like before this PR), Helm 3 will first install `crds` folder and then try installing CRDs again since they are rendered into the `templates` folder, which results in `already exists` error. Putting `pre-install` annotation will make it work with Helm2 but not with Helm3 since these annotations are not used at all. It would've been good if we could switch rendering depending on Helm version but I couldn't find a good way to get Helm's version in the template functions.

So, I'm not sure a good, native solution exists. Maybe we can have a field in `values.yaml` for Helm version that user can supply, then use that to switch rendering on/off.

PR #855, opened 2019-09-30
Remove low value API tests

negz:

It seems that https://github.com/crossplaneio/crossplane/pull/844 somehow broke our API tests in a fashion that wasn't caught by Jenkins. The straightforward fix is to update [the testing CRD path](https://github.com/crossplaneio/crossplane/blob/master/pkg/test/path.go), but I want to propose a more controversial fix: deleting the affected tests.

This PR removes two classes of test that I consider to be of low value:

* Schema registration tests. This class of test tests that the AddToScheme functions that register API types with the Kubernetes scheme is working. I consider this test to be low value for two reasons. Firstly, it's testing a very small, generated scaffold around runtime.SchemeBuilder. Crossplane should not be testing the logic of the runtime package; that's the runtime package's job. Secondly, the intent of the tests seems to be to ensure that we remember to register all of the API types that we intended to register; i.e. that we do not add a new API type and forget to register it with the scheme. Unfortunately for the test to catch this case we'd have to remember to update the test itself, but not the AddToSchemes slice.

* "Storage" tests. These appear to be variants on old tests scaffolded by kubebuilder. Each test starts up a Kubernetes API server then ensures it can CRUD the API type under test against that API server. I consider this test to be low value because as far as I can tell this is just testing that our API types are in fact valid Kubernetes types, i.e. that they have type and object metadata.

### Checklist

I have:


negz:
> Aren't CRD registration errors covered by integration test anyway since it installs crossplane via its helm chart?

I believe so!

negz:
We discussed this after standup this morning and there are no dissenters, so I will now merge. We do agree that the "storage" tests this PR removes could have some value in the context of a larger and more thorough integration test framework, e.g. one that tested end-to-end dynamic provisioning.

PR #857, opened 2019-09-30
docs: add definition of done

suskin:

In the retrospective for the 0.3 release (and in the sig-stacks sprint
retrospective recently), the topic came up of documenting the definition
of done, so that we can all be agreed about it. The difference between
the definition we had previously been using and the definition in this
changeset is that documentation and error reporting are now included.

[skip ci]

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -18,14 +18,12 @@ can uncomment the below line to indicate which issue your PR fixes, for example
 Fixes #
 -->
 
-### Testing done
+### How has this code been tested?
 <!--
-How has this code been tested?
-
-Before reviewers can be confident in the correctness of a changeset,
+Before reviewers can be confident in the correctness of a pull request,
 it needs to be tested. In this section, briefly describe the testing
 that has already done or which is planned. The testing should be enough
-to help people feel comfortable in the changeset's correctness.
+to help people feel comfortable in the pull request's correctness.
 -->
 
 ### Checklist
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -21,9 +21,8 @@ Fixes #
 ### How has this code been tested?
 <!--
 Before reviewers can be confident in the correctness of a pull request,
-it needs to be tested. In this section, briefly describe the testing
-that has already done or which is planned. The testing should be enough
-to help people feel comfortable in the pull request's correctness.
+it needs to tested and shown to be correct. In this section, briefly
+describe the testing that has already been done or which is planned.
 -->
 
 ### Checklist
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -18,6 +18,16 @@ can uncomment the below line to indicate which issue your PR fixes, for example
 Fixes #
 -->
 
+### Testing done
+<!--
+How has this code been tested?
+
+Before reviewers can be confident in the correctness of a changeset,
+it needs to be tested. In this section, briefly describe the testing
+that has already done or which is planned. The testing should be enough
+to help people feel comfortable in the changeset's correctness.
+-->
+
 ### Checklist
 <!--
 Please run through the below readiness checklist. The first two items are
@@ -26,10 +36,13 @@ relevant to every Crossplane pull request.
 I have:
 - [ ] Run `make reviewable` to ensure this PR is ready for review.
 - [ ] Ensured this PR contains a neat, self documenting set of commits.
-- [ ] Updated any relevant [documentation], [examples], or [release notes].
-- [ ] Updated the RBAC permissions in [`clusterrole.yaml`] to include any new types.
+- [ ] Updated any relevant [documentation] and [examples].
+- [ ] Reported all new error conditions into the log or as an event, as
+  appropriate.
+
+For more about what we believe makes a pull request complete, see our
+[definition of done].
 
 [documentation]: https://github.com/crossplaneio/crossplane/tree/master/docs
 [examples]: https://github.com/crossplaneio/crossplane/tree/master/cluster/examples
-[release notes]: https://github.com/crossplaneio/crossplane/tree/master/PendingReleaseNotes.md
-[`clusterrole.yaml`]: https://github.com/crossplaneio/crossplane/blob/master/cluster/charts/crossplane/templates/clusterrole.yaml
+[definition of done]: https://github.com/crossplaneio/crossplane/tree/master/design/one-pager-definition-of-done.md
--- a/design/one-pager-definition-of-done.md
+++ b/design/one-pager-definition-of-done.md
@@ -20,15 +20,21 @@ Work means a contribution to the project, typically code.
 
 ## Definition of done
 
-A chunk of work is considered done when the following criteria have been met:
+The following criteria **MUST** be met for work to be considered done:
 
-* All code is written
 * Code is reviewed and approved
-* New behavior is documented in the documentation that will go onto the
-  crossplane website
+* Code passes linting (use `make reviewable` to lint)
 * All error conditions are reported into the log or as an event, as
   appropriate
-* Code is tested, ideally in an integration test
+* Code is tested
+
+Additionally, the following criteria **SHOULD** be met for the work to
+be considered done, as they are best practices:
+
+* Code is tested with an integration test
+* New behavior is documented in the documentation that will go onto the
+  crossplane website
+* Examples have been updated to reflect any changes
 
 ## Applying definition of done
 
--- a/design/one-pager-definition-of-done.md
+++ b/design/one-pager-definition-of-done.md
@@ -1,0 +1,40 @@
+# Definition of done
+
+* Owner: Daniel Suskin (@suskin)
+* Reviewers: Crossplane Maintainers
+* Status: Draft
+
+## Introduction
+
+*Why does this document exist?*
+
+The goal of this document is to answer the questions:
+
+* As a developer, when should I consider a chunk of work to be complete?
+* As a reviewer, what should I be looking for to verify that a chunk of work is
+  complete?
+
+## Definition of work
+
+Work means a contribution to the project, typically code.
+
+## Definition of done
+
+A chunk of work is considered done when the following criteria have been met:
+
+* All code is written
+* Code is reviewed and approved
+* New behavior is documented in the documentation that will go onto the
+  crossplane website
+* All error conditions are reported into the log or as an event, as
+  appropriate
+* Code is tested, ideally in an integration test
+
+## Applying definition of done
+
+The definition of done can be applied in the following ways:
+
+* Pull requests should not be merged until the work contained in them is
+  done.
+* Issues tracking work should not be closed until the work being tracked
+  is done.
--- a/design/one-pager-definition-of-done.md
+++ b/design/one-pager-definition-of-done.md
@@ -22,25 +22,33 @@ Work means a contribution to the project, typically code.
 
 The following criteria **MUST** be met for work to be considered done:
 
-* Code is reviewed and approved
-* Code passes linting (use `make reviewable` to lint)
-* All error conditions are reported into the log or as an event, as
-  appropriate
-* Code is tested
+* Code is reviewed and approved.
+* Code passes linting (use `make reviewable` to lint).
+* Error conditions are reported as and where appropriate. For more about
+  how to think about error reporting, see our [philosophical
+  guide][error-reporting-philosophy].
+* Code is tested to an extent such that: we are confident in its
+  correctness; and future maintainers will be able to tell if they have
+  introduced a regression.
 
 Additionally, the following criteria **SHOULD** be met for the work to
 be considered done, as they are best practices:
 
-* Code is tested with an integration test
+* Code is exercised with an integration test (existing or new).
 * New behavior is documented in the documentation that will go onto the
-  crossplane website
-* Examples have been updated to reflect any changes
+  crossplane website.
+* Examples have been updated to reflect any changes.
 
 ## Applying definition of done
 
 The definition of done can be applied in the following ways:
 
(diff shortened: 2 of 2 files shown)


suskin:
I have added a few people specifically as reviewers, but theoretically anyone with the context from the retrospective should be able to review this.

design/one-pager-definition-of-done.md
```diff
@@ -0,0 +1,40 @@
...
+
+Work means a contribution to the project, typically code.
+
+## Definition of done
+
+A chunk of work is considered done when the following criteria have been met:
+
+* All code is written
+* Code is reviewed and approved
+* New behavior is documented in the documentation that will go onto the
```
hasheddan: Do we want to call out any required updates to `examples/` as well? They are not specifically used in the Crossplane website, but I think we should desire for them to always be up to date.

design/one-pager-definition-of-done.md
```diff
@@ -0,0 +1,40 @@
...
+  crossplane website
+* All error conditions are reported into the log or as an event, as
+  appropriate
+* Code is tested, ideally in an integration test
```
negz: This reads like we prefer integration tests to unit tests, which I don't recall us agreeing on. Is it trying to suggest an integration test in addition to unit tests?

design/one-pager-definition-of-done.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Applying definition of done
+
+The definition of done can be applied in the following ways:
+
+* Pull requests should not be merged until the work contained in them is
+  done.
```
negz: Is there a way we could integrate this document with the existing checklist that is exposed as part of the [Github PR template](https://github.com/crossplaneio/crossplane/blob/master/.github/PULL_REQUEST_TEMPLATE.md)?

design/one-pager-definition-of-done.md
negz: Either way - is this actually something we expect today? We've only barely started establishing a pattern for integration testing in Crossplane. I would not personally know where to start if I were opening a new PR against, for example, one of the infrastructure stacks.

In general, I caution that [broken windows theory](https://en.wikipedia.org/wiki/Broken_windows_theory) (or a variant thereof) is [definitely a thing](https://github.com/crossplaneio/crossplane/pull/535#issuecomment-501458787) when it comes to checklists. If there's 1-2 things on the checklist that are skipped much of the time, then it becomes increasingly easy to ignore more and more of the checklist.  Therefore I would suggest:

* Keeping this list as short as possible.
* Breaking it up into _absolutely definitely required for every code change_ and _desired for every code change_.


We may consider looking for automation opportunities at some future time. One example - lint for PRs that touch code but that do not touch `docs/`.

design/one-pager-definition-of-done.md
suskin: This is a good point . . . let me think about how to include this as well

design/one-pager-definition-of-done.md
suskin: I agree with the point about broken windows here, and that it would be helpful to have a MUST and SHOULD section. I'll take a look into reworking the list to take that into account.

> This reads like we prefer integration tests to unit tests, which I don't recall us agreeing on. Is it trying to suggest an integration test in addition to unit tests?

This is something we should discuss. The reasoning behind this wording is that we don't know whether the changes really worked end-to-end until we write and execute an integration test. Having unit tests does increase our confidence, but doesn't necessarily guarantee that the behavior is correct from the perspective of a user scenario. What are our thoughts on whether we want to consider a chunk of work complete if we don't know whether the user scenario works?

A related discussion would be what we want to do to distinguish between the testing requirements for merging a changeset versus the testing requirements for promoting a changeset into a release. And how that relates to our definition of when a work item is done: if there **is** a distinction, is the work done when it is merged, or when it is released?

My personal preference is to consider work done when it is ready to be released.

design/one-pager-definition-of-done.md
suskin: Ooo, nice! Let me take a closer look at this.

suskin: I have done some integration now

design/one-pager-definition-of-done.md
suskin: I have made some changes based on these ideas, but I believe the discussion about the place of integration tests and the difference between the merge gate and the release gate are discussions we will want to have before we settle on a definition of done.

design/one-pager-definition-of-done.md
suskin: I have added a line about this in the new section of things which **should** be done

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -18,6 +18,16 @@ can uncomment the below line to indicate which issue your PR fixes, for example
 Fixes #
 -->
 
+### Testing done
+<!--
+How has this code been tested?
```
negz: Perhaps this should be the title of this section, rather than "How has this code been tested?"

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -18,6 +18,16 @@ can uncomment the below line to indicate which issue your PR fixes, for example
...
+
+Before reviewers can be confident in the correctness of a changeset,
```
negz: "Changeset" is an unusual term for Github interactions. I suggest "pull request" would be more idiomatic.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -18,6 +18,16 @@ can uncomment the below line to indicate which issue your PR fixes, for example
...
+it needs to be tested. In this section, briefly describe the testing
+that has already done or which is planned. The testing should be enough
```
negz: "been done"?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -18,6 +18,16 @@ can uncomment the below line to indicate which issue your PR fixes, for example
...
+to help people feel comfortable in the changeset's correctness.
```
negz: Is there value in both opening by stating that testing helps reviewers feel comfortable with a change's correctness, and closing by stating that the tests should help reviewers feel comfortable with a change's correctness?

design/one-pager-definition-of-done.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Definition of done
+
+The following criteria **MUST** be met for work to be considered done:
+
+* Code is reviewed and approved
+* Code passes linting (use `make reviewable` to lint)
+* All error conditions are reported into the log or as an event, as
+  appropriate
+* Code is tested
```
negz: Is this purposefully vague about how a code has been tested? If so, should we suggest a unit testing goal or attempt to briefly summarise our philosophy toward unit tests in the below section?

design/one-pager-definition-of-done.md
```diff
@@ -0,0 +1,46 @@
...
+Work means a contribution to the project, typically code.
+
+## Definition of done
+
+The following criteria **MUST** be met for work to be considered done:
+
+* Code is reviewed and approved
+* Code passes linting (use `make reviewable` to lint)
+* All error conditions are reported into the log or as an event, as
+  appropriate
```
negz: This feels overly prescriptive to me. I would prefer this sentence without the "all", and the details i.e. "Error conditions are reported where appropriate". Given how errors work in Go, I think it would be easy to interpret this as "every time you find `err != nil` you must log it", which is not a position I agree with.

https://www.youtube.com/watch?v=PAAkCSZUG1c&t=16m13s
https://blog.golang.org/errors-are-values

I'd prefer to keep this vague for now, then come back and update it when we ratify your proposal around error handling.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -26,10 +36,13 @@ relevant to every Crossplane pull request.
 I have:
 - [ ] Run `make reviewable` to ensure this PR is ready for review.
 - [ ] Ensured this PR contains a neat, self documenting set of commits.
-- [ ] Updated any relevant [documentation], [examples], or [release notes].
-- [ ] Updated the RBAC permissions in [`clusterrole.yaml`] to include any new types.
```
negz: This file still exists, and is frequently forgotten when adding new types. Can you provide some context around why you think it should not be in this checklist?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -26,10 +36,13 @@ relevant to every Crossplane pull request.
...
+- [ ] Updated any relevant [documentation] and [examples].
```
negz: I notice this removes 'release notes' from the checklist. I think I'm on board with this because I don't think I've ever seen a Crossplane PR actually update the release notes.

design/one-pager-definition-of-done.md
```diff
@@ -0,0 +1,46 @@
...
+be considered done, as they are best practices:
+
+* Code is tested with an integration test
+* New behavior is documented in the documentation that will go onto the
+  crossplane website
+* Examples have been updated to reflect any changes
+
+## Applying definition of done
+
+The definition of done can be applied in the following ways:
```
negz: > As a reviewer, what should I be looking for to verify that a chunk of work is complete?

Given the above, do we prefer all work to be tracked in a PR? Put otherwise, is it acceptable to add code and documentation in separate PRs? Is it acceptable to add code and tests in separate PRs? My preference would be to encourage that all work - code, tests, docs, etc - be grouped (potentially as separate commits) in one PR.

design/one-pager-definition-of-done.md
```diff
@@ -0,0 +1,46 @@
...
+* Code is reviewed and approved
+* Code passes linting (use `make reviewable` to lint)
+* All error conditions are reported into the log or as an event, as
+  appropriate
+* Code is tested
+
+Additionally, the following criteria **SHOULD** be met for the work to
+be considered done, as they are best practices:
+
+* Code is tested with an integration test
```
negz: Perhaps "code is exercised by an integration test"? I think this makes it a little clearer that the nature of integration tests is such that you don't always need to write a net new integration test (or even update one) for your code to be covered by integration tests.

.github/PULL_REQUEST_TEMPLATE.md
displague: I've been using a Markdown Linter that tends to disagree.
https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md#md026---trailing-punctuation-in-header
I've been blindly abiding, but have created an issue there to suggest they offer some rationale for this ruling.

design/one-pager-definition-of-done.md
displague: The previous section ends ", typically code."

Bullets 1,2, and 4 assert that this applies to "code" change, but bullet 3 does not -- it is both too specific (exact use of errors) and not specific enough (only for code?).

We may need different "definitions of done" sections for: code, design, documentation, examples, artwork, build artifacts (and Stacks). And what of meta-work like issue and epic creation and preening?

We could refer to existing guides when possible.

On the other hand, if we can keep this generic, we may not need to spell out DoD for each type of work.

design/one-pager-definition-of-done.md
displague: https://github.com/crossplaneio/crossplane/pull/931/files#diff-f31a09f3f47ef35a6f5d5b83aa497071R83-R90 specifies that all work (docs, examples, testing, code) related to a change or set of changes in Crossplane must be included in a single PR.

jbw976:
@suskin were you able to take a look at the feedback from @negz? It would be good to get this PR merged.

.github/PULL_REQUEST_TEMPLATE.md
jbw976: yes, this check is still valid to me

.github/PULL_REQUEST_TEMPLATE.md
jbw976: i actually dislike that, it makes the burden of authoring the release notes greater when it comes time for release.  I would actually like to see the pending release notes get utilized and be a check item that is respected in this PR template.  We've had success with this in Rook, e.g. the release notes basically take shape **during** the milestone: https://github.com/rook/rook/blob/master/PendingReleaseNotes.md

design/one-pager-definition-of-done.md
```diff
@@ -0,0 +1,46 @@
...
+* New behavior is documented in the documentation that will go onto the
```
jbw976: i'm leaning towards documentation updates being in the **MUST** category honestly

design/one-pager-definition-of-done.md
suskin: I have added some more detail to try to make things clearer

.github/PULL_REQUEST_TEMPLATE.md
jbw976: actually, i think these basic permissions would be exercised by the integration tests.  when the pod starts up and creates watches for all the types its controllers care about, the pod should crash immediately if we have't added any new necessary permissions.  I think our integration automation would catch this.

.github/PULL_REQUEST_TEMPLATE.md
jbw976: i don't think it's a particularly strong argument to say we should not include the pending release notes in the definition of done just because they aren't currently being used.  i think this is an opportunity to define what our process is going forward, we can establish and make changes to the process right now.  That being said, I won't hold this PR up for it.

jbw976:
@suskin the unit tests failed with:
```
stat /home/upbound/go/src/github.com/crossplaneio/crossplane/cluster/charts/crossplane/crds: no such file or directory
```
does this need to be rebased to pick up some changes? it's been open for a while, maybe it's gone stale/incompatible with the build pipelines?

.github/PULL_REQUEST_TEMPLATE.md
suskin: Awesome :squid:

suskin:
> does this need to be rebased

@jbw976 I've rebased it now, so it should be good

Issue #859, opened 2019-10-01
Use KubernetesTarget kind for scheduling KubernetesApplications


hasheddan:
### What problem are you facing?

The introduction of the Kubernetes `Provider` type (as specified in #841) allows for enhanced ability to schedule resources to *any* Kubernetes cluster. This functionality should be utilized by `KubernetesApplication` types instead of the current pattern of referencing a Crossplane provisioned `KubernetesCluster` by label.

### How could Crossplane help solve your problem?

The `KubernetesApplication` resource type will need to be modified to include a
`providerRef` field so that in can be referenced by the new Kubernetes `Provider` kind.
However, this should be an optional field that is only required if there is no
`clusterRef` `matchLabels` fields defined. If there is only a `clusterRef`
`matchLabels` field and no `providerRef`, the `KubernetesApplication` [scheduler
controller] should be modified such that it finds Kubernetes `Provider` objects
with the labels (instead of the current implementation that gets
`KubernetesCluster` objects with matching labels), and schedules the
`KubernetesApplication` to them in a round-robin fashion.

If a user desires to use the `providerRef` field to provision the
`KubernetesApplication`, the workflow would look as follows: 

1. Create a Kubernetes `Provider` for target cluster (using one of scenarios
   outlined above):

```yaml
---
apiVersion: kubernetes.crossplane.io/v1alpha1
kind: Provider
metadata:
  name: demo-k8s-provider
  namespace: my-infra-dev
spec:
  credentialsSecretRef:
    name: demo-cluster-creds
```

2. Create a `KubernetesApplication` that references the `Provider` in its
   `providerRef`:

```yaml
apiVersion: workload.crossplane.io/v1alpha1
kind: KubernetesApplication
metadata:
  name: wordpress-demo
  namespace: mynamespace
  labels:
    app: wordpress-demo
spec:
  resourceSelector:
    matchLabels:
      app: wordpress-demo
  providerRef:
    name: demo-k8s-provider
    namespace: my-infra-dev
  resourceTemplates:
  - metadata:
      name: wordpress-demo-namespace
      labels:
        app: wordpress-demo
    spec:
      template:
        apiVersion: v1
        kind: Namespace
        metadata:
          name: wordpress
          labels:
            app: wordpress
  - metadata:
      name: wordpress-demo-deployment
      labels:
        app: wordpress-demo
    spec:
      secrets:
        # This must match the writeConnectionSecretToRef field
        # on the database claim; it is the name of the secret to
[...]

PR #873, opened 2019-10-02
k8s provider: implement kubernetes provider

### LINKED ISSUE #872: Implement Kubernetes Provider
Part of #862 

#841 introduced the design of a Kubernetes `Provider` kind. It is required for the desired functionality of `stack-rook`and design is outlined in [K8s-Native Provider one-paged](https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-k8s-native-providers.md).

Implementation (at least initially) should be in the core Crossplane repo so that it can be used across stacks. There is also a possibility of using the Kubernetes `Provider` with `KubernetesApplication` as tracked by #859.

hasheddan:

Fixes #872 

### Checklist

I have:


apis/kubernetes/v1alpha1/types.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// A ProviderSpec defines the desired state of a Provider.
+type ProviderSpec struct {
+	// Kubernetes credentials Secret reference
+	// A Secret containing connection credentials for a Kubernetes cluster
+	// client that will be used to authenticate to this Kubernetes Provider.
```
negz: This comment will be used to generate user facing API documentation, so I think there would be value in explaining what we expect this secret to look like. We probably don't need to go as far as outlining all the expected fields, but it would probably be useful to say something like "this will typically be the connection secret of a KubernetesCluster claim, or the secret created by a Kubernetes service account".

hasheddan: Description fixed and docs generated using crossdocs!

hasheddan:
@negz would you say that this fixes #497 or should we wait for #859 to be implemented?

negz:
> @negz would you say that this fixes #497 or should we wait for #859 to be implemented?

I don't think #497 is really fixed until we've implemented, tested, and documented #859. I could see a pretty strong case for those two issues being deduplicated though, because #859 is pretty much proposing a specific technical solution to the functionality requested by #497.

Issue #922, opened 2019-10-09
Manage the Portable Resource Classes at a higher level in a single namespace


prasek:
### What problem are you facing?
From: Dylan Griffith
https://gitlab.com/groups/gitlab-org/-/epics/1866#note_223600767

> @prasek: 
>
> @tkuah and I were discussing this just in time stuff the other day and I'm curious to know if it's necessary to create these "Crossplane policy yamls" in each namespace.
>
> I was curious to know if Crossplane has some way to manage these at a higher level in a single namespace and still allow all namespaces to create the crossplane resource claims. It would be  preferable if we didn't need to update the GitLab backend code to add all of these resources to every namespace we create.
>
> I would also think that the downside of installing all of these policies into every namespace is that it's tricky to keep them updated.

### Background
The bulk of the managed service config (resource class templates, provider config, etc.) is typically stored in a single infra namespace (per environment e.g. aws-infra-dev) and used by multiple app project namespaces.

The policy yamls (portable resource classes) are very simple mappings that make classes-of-service available for use in an app project namespace, and to configure default classes-of-service.

```yaml
apiVersion: database.crossplane.io/v1alpha1
kind: MySQLInstanceClass
metadata:
  name: mysql-standard
  namespace: app-project1-dev
  labels:
    default: true
classRef:
  kind: RDSInstanceClass
  apiVersion: database.aws.crossplane.io/v1alpha1
  name: rds-mysql-standard
  namespace: aws-infra-dev
```

In an AWS environment offering multiple classes of service, the following Kubernetes objects would result:
```text
namespaces
└── aws-infra-dev
      └── provider                 # AWS provider configuration
      └── provider-creds           # AWS provider account credentials
      └── rds-mysql-standard       # RDS-specific class, non-portable config
      └── rds-mysql-replicated     # RDS-specific class, non-portable config
      └── rds-postgres-standard    # RDS-specific class, non-portable config
      └── rds-postgres-replicated  # RDS-specific class, non-portable config
└── app-project1-dev
      └── mysql-standard           # portable MySQL class of service
      └── mysql-replicated         # portable MySQL class of service
      └── postgres-standard        # portable PostgreSQL class of service
      └── postgres-ha              # portable PostgreSQL class of service
      └── mysql-claim              # portable MySQL claim for mysql-standard class of service
      └── mysql-claim-secret       # generated secret to access database
      └── wordpress-deployment     # standard Kubernetes deployment
[...]

Issue #703, opened 2019-08-23
Claim portability improvements


eonpatapon:
## Current situation

Currently claims are not really portable if they are not using the default class configured with ressource class policy.

To target a specific class which is not the default a user needs to know the specific class kind and api as well as its name and namespace using the `classRef` attribute:

```yaml
apiVersion: database.crossplane.io/v1alpha1
kind: MySQLInstance
metadata:
  name: mysql-instance
spec:
  classRef:
    kind: RDSInstanceClass
    apiVersion: database.aws.crossplane.io/v1alpha1
    name: mysql-standard
    namespace: crossplane-system
```

## Proposal

Instead of targeting a specific class the user could define a set of wanted properties for the claim. At creation crossplane would automatically choose the ressource class that matches the user claim.

For example in the case of a database instance the user could define the minimum size of the DB:

```yaml
apiVersion: database.crossplane.io/v1alpha1
kind: MySQLInstance
metadata:
  name: mysql-instance
spec:
  classRequirements:
    size: 40
```
In this case crossplane will try to find a concrete class that has a `size` at least of `40`. If not a proper error should be returned to the user.

To be able to find the available classes for an abstract type the relation between policy and classes could be done the other way.

Currenly with a `MySQLInstancePolicy` resource it is possible to define the default resource class to use:

```yaml
apiVersion: storage.crossplane.io/v1alpha1
kind: MySQLInstancePolicy
metadata:
  name: mysql-policy
  namespace: demo
defaultClassRef:
  kind: RDSInstanceClass
  apiVersion: database.aws.crossplane.io/v1alpha1
  name: standard-mysql
  namespace: crossplane-system
```

Instead of defining a ref `MySQLInstancePolicy - <defaultClassRef> -> RDSInstanceClass` it is maybe possible to reverse it to reference multiple concrete class in the policy: `(RDSInstanceClass|CloudsqlInstanceClass|...) - <policyRef (default: true|false)> -> MySQLInstancePolicy`:

```yaml
apiVersion: database.aws.crossplane.io/v1alpha1
kind: RDSInstanceClass
metadata:
  name: standard-mysql
  namespace: crossplane-system
specTemplate:
  size: 20
  providerRef: ...
  policyRef: 
    kind: MySQLInstancePolicy
    apiVersion: storage.crossplane.io/v1alpha1
    default: true
---
apiVersion: database.gcp.crossplane.io/v1alpha1
kind: CloudsqlInstanceClass
metadata:
[...]

Issue #113, opened 2018-11-15
Abstract type values translation and validation


jbw976:
This is a general problem that we need to solve better. When an abstract type exposes a common field, we want the user to be able to think about the values for that field in a general way, and to not worry about provider specifics.  An example is the MySQL version that is exposed on the abstract MySQLInstance type.  We want the user to be able to input something like `5.6` or `5.7` and under the covers during provisioning those values get validated and translated to a provider specific value, such as `MYSQL_5_6` for CloudSQL, or `5.6.39` for RDS.

We want a solution that has the following properties:
* we want the app author to be able to get the version that they need for their app
* we don’t want to have to rev any software when new versions are added on the provider/server side (e.g., hard coded "valid value" lists are not good)
* we want a consistent experience across all providers when using the abstract type, the app author shouldn't be thinking about provider specifics (but the admin that is defining concrete resource classes sure can)
* we don’t want the user to be inputting invalid values, but that’s less of a priority since the create  call could just fail downstream later with hopefully a good error message

negz:
Relates to #870.

negz:
We have a sketch of how we might do this (transform functions) in https://github.com/crossplaneio/crossplane/pull/1163.

negz:
We moved forward with #1163, which I believe addresses all of the requirements here!

Issue #91, opened 2018-11-05
Consider matching the best `ResourceClass` for a given abstract resource


bassam:
Currently `classReference` is required for dynamic provisioning, and if it's omitted we are thinking of supporting a default class.

It would be more useful if instead of just a single default class, we could select the best available class that meets the requirements of the abstract resource. Abstract resources would need to specify more granular requirements for that to be useful, for example, for `MySqlInstance` they might need to specify the requirements on IOPS.

The end result here is that the abstract resource would not need to know about classes at all, and instead define more granular requirements that get matched heuristically.

negz:
Relates closely to #870.

negz:
Cross linking relevant discussion from https://github.com/crossplaneio/stack-gcp/pull/46#discussion_r334800076.

negz:
I _think_ this is no longer relevant given that we're deprecating resource classes. I suppose we could do something like this for selecting a composition based on the composite resource's configuration, though I'm not sure whether we still see that as a direction we want to go. I'm going to close this - let me know if you want it to stay open and we can instead reframe it around composition.

PR #926, opened 2019-10-10
Simple Resource Class Selection one pager

### LINKED ISSUE #870: Match claims to classes using label selectors
## What problem are you facing?

In Crossplane 0.3 we introduced support for "provider specific" (aka "non-portable", aka "[strongly typed](https://github.com/crossplaneio/crossplane/blob/69e69c/design/one-pager-strongly-typed-class.md)") resource classes. During the development cycle [we identified](https://github.com/crossplaneio/crossplane/issues/723) that if a claim must explicitly reference a provider specific resource class in order to enable dynamic provisioning, then the claim itself is now provider specific and no longer portable across providers. We solved this issue by introducing [portable resource classes](https://github.com/crossplaneio/crossplane/blob/69e69c/design/one-pager-default-resource-class.md).

A portable resource class is effectively an indirection to a non-portable resource class. Portable resource classes can be set as the default for a particular namespace, and will be used by any resource claims of their corresponding kind that do not specify a resource class for dynamic provisioning, or a managed resource for static provisioning. This can be thought of as "publishing" non-portable resource classes (which may exist in a distinct namespace used to logically group infrastructure) to other namespaces, where applications may use them to satisfy claims for their infrastructure requirements.

This pattern is flexible and powerful, but Crossplane maintainers and [community members](https://github.com/crossplaneio/crossplane/issues/703#issuecomment-536958545) have observed that it's verbose; three distinct Kubernetes resources (a claim, portable class, and non-portable class) must exist for dynamic provisioning to occur.

## How could Crossplane help solve your problem?

Crossplane could remove the need for portable resource classes by using [label selectors](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors) at the resource claim level, rather than the concept of a default class.

### User experience

Resource claim authors who have a specific, non-portable resource class in mind just set the claim's `classRef` to that resource class (including its type metadata) just like they did before https://github.com/crossplaneio/crossplane/issues/723. Resource claim authors who have a specific, managed resource in mind just set the claim's `resourceRef` as they do today for static provisioning. Resource claim authors who don't know exactly what resource class they want omit the `classRef`, omit the `resourceRef`, and supply zero or more `classSelector` labels to match.

For example:

```yaml
---
# Imagine this resource claim is created by a hypothetical GitLab stack
# controller to satisfy a request for a new GitLab installation. It does not
# specifically name a resource class, but declares it wants an experimental
# grade PostgreSQLInstance that is compatible with the Gitlab stack, in
# region us-east-1.
kind: PostgreSQLInstance
metadata:
  name: gitlab-database
  namespace: gitlab-experiments
spec:
  classSelector:
    matchLabels:
      stack: gitlab
      grade: experimental
      region: us-east-1
  writeConnectionSecretToRef:
    name: app-postgresql-connection
```

Note that under this proposal the concept of a "portable class" is removed, so any reference to "class" or "resource class" means "non-portable resource class", for example `CloudSQLInstanceClass`.

From the claim author's perspective Crossplane would handle dynamic provisioning of the above `PostgreSQLInstance` claim by:

1. "Finding" all classes that match its `.spec.classSelector.matchLabels`.
1. "Picking one at random".
1. Seting the `PostgreSQLInstance` `.spec.classRef` to the randomly selected resource class.
1. Dynamically provisioning and binding as we do today.

This proposal has the following properties:

* It supports portable, strongly typed, dynamic provisioning using only two concepts; resource claims and resource classes.
* It works the same for a hypothetical non-portable `CloudSQLInstanceClaim` as it does for a portable `PostgreSQLInstanceClaim`.
* It approximates support for "default resource classes". Label a resource class `{default: "true"}` to make it the global default for compatible resource claims (that select that label). Label a resource class `{default: "true", namespace: app-project1-dev}`, to make it the "default for a namespace" (actually, the default for any claim that selects those labels).
* It makes resource class selection non-deterministic. If multiple resource classes match the labels, one is chosen at random.
* It removes the concept of "publishing" a class to a namespace. This means a cluster operator can set a class as the "default" (kind of - see above) for app namespaces that don't exist yet, but also makes it harder for app operators to determine what resource classes they could choose from, presuming the resource classes lived in a distinct infrastructure namespace.
* It enables a rudimentary form of "matching" (dare I say scheduling?) of resource claims by decoupling the language used to match claims to classes from the (strongly typed) language used to describe classes.

Elaborating on that last bullet, it's hard to match the requirements of a resource claim to the capabilities of a resource class on `spec` alone due to the "lowest common denominator" problem of multi cloud. The `RedisCluster` claim, for example, exposes a field named `engineVersion`, which specifies its desired version of Redis (e.g. `"3.2"`). Ideally we'd use this snippet of information to intelligently match a `RedisCluster` claim to a resource class for dynamic provisioning, but that is hard(TM). The following resource classes could satisfy a `RedisCluster` claim today:

* GCP `CloudMemorystoreInstanceClass` has an equivalent field to `engineVersion` - it's named `redisVersion` and expects the version to be written as `"REDIS_3_2"`, not `"3.2"`.
* AWS `ReplicationGroupClass` has an `engineVersion` field that expects the patch version to be included, i.e. `"3.2.0"`.
* Azure `RedisClass` has no field for engine version. It's just always version 3.2.

We could try to solve this by having resource claims expose the intersection (or the union?) of every possible resource class's set of fields in some standardised fashion, then teach each resource claim controller how to match the standardised resource claim fields to the provider specific resource class fields, or we could just use labels.

For example this resource claim specifies that it needs Redis 3.2:

```yaml
---
apiVersion: cache.crossplane.io/v1alpha1
kind: RedisCluster
metadata:
  name: gitlab-cache
  namespace: gitlab-experiments
spec:
  classSelector:
    matchLabels:
      stack: gitlab
      engineVersion: "3.2"
  writeConnectionSecretToRef:
    name: gitlab-cache-connection
```

While these resource classes specify that they can satisfy claims that need Redis 3.2, while also using documented, high fidelity, validated, schemas when dynamically provisioning managed resources:

```yaml
---
apiVersion: cache.gcp.crossplane.io/v1alpha2
kind: CloudMemorystoreInstanceClass
metadata:
  name: gitlab-development-default
  namespace: gcp-infra-dev
  labels:
    engineVersion: "3.2"
specTemplate:
  tier: STANDARD_HA
  region: us-west2
  memorySizeGb: 1
  redisVersion: REDIS_3_2
  providerRef:
    name: example
    namespace: gcp-infra-dev
  reclaimPolicy: Delete
---
apiVersion: cache.azure.crossplane.io/v1alpha2
kind: RedisClass
metadata:
  name: azure-redis-standard
  namespace: azure-infra-dev
  labels:
    engineVersion: "3.2"
specTemplate:
  resourceGroupName: group-westus-1
  location: West US
  sku:
    name: Basic
    family: C
    capacity: 0
  enableNonSslPort: true
  providerRef:
    name: example
    namespace: azure-infra-dev
  reclaimPolicy: Delete
```

### Implementation Details

There are a lot of quotation marks in the above steps because what would actually be happening under the hood would be more like a race to satisfy the claim. In Crossplane there is not one controller for each resource claim kind, but rather one controller for each possible (resource claim, managed resource) tuple. This allows an infrastructure stack that adds support for a managed resource that could satisfy a resource claim to implement the dynamic provisioning and claim binding logic for said claim without having to touch Crossplane core. Put otherwise, `crossplaneio/stack-gcp` can add support for `CloudSQLInstance` resources to satisfy `MySQLInstance` claims in `crossplaneio/stack-gcp`, instead of teaching `crossplaneio/crossplane` how to dynamically provision a `CloudSQLInstance`. Hence the race. In reality the following would happen for the example above:

* All resource claim controllers are updated to only reconcile resources with a `resourceRef` or a `classRef`.
* A new "class selector" controller is introduced for each (resource claim, resource class) tuple. Its job is to set the `classRef` for resource classes that omit it, by using their `classSelector`.

Upon the creation of a `PostgreSQLInstance` without a `classRef` or `resourceRef`:

1. Every class selector controller watching for `PostgreSQLInstance` claims has a reconcile queued. I'll use the `(PostgreSQLInstance, CloudSQLInstanceClass)` controller for this example, but `(PostgreSQLInstance, RDSInstanceClass)` and `(PostgreSQLInstance, MysqlServerClass)` controllers would run through the process in parallel.
1. The controller would list all `CloudSQLInstanceClass` resources (in any namespace) that matched the `PostgresSQLInstance` claim' label selectors.
1. If no `CloudSQLInstanceClass` matched the labels, the reconcile is done. Otherwise, one of the matching `CloudSQLInstanceClass` resources is selected at random.
1. The controller sleeps for a small, jittered amount of time then sets the `classRef` of the `PostgreSQLInstance` to the selected `CloudSQLInstanceClass`. The jitter reduces the chance of two controllers trying to set the `classRef` at the same time. If two controllers _do_ try to set the `classRef` at the same time one will fail to commit the change due to the `PostgreSQLInstance` claim's resource version having changed since it was read.
1. The reconcile is done. With the `classRef` set the `PostgresSQLInstance` now passes the watch predicates of the `(PostgreSQLInstance, CloudSQLInstance)` resource claim reconciler, which dynamically provisions and binds it.

negz:

This one pager cleans up the proposal that we have been discussing in #870. In short, it proposes removing portable resource classes and matching resource claims directly to (non-portable) resource classes using label selectors. This maintains portability of resource classes, while removing a layer of complexity that has been a sticking point for potential Crossplane users (#922, https://github.com/crossplaneio/crossplane/issues/703#issuecomment-536958545). As a nice side effect, it enables rudimentary scheduling of resource claims to resource classes.

Fixes #870 (and possibly #922, #91, #113, #703).

### Checklist

I have:

### DIFF
--- a/design/one-pager-simple-class-selection.md
+++ b/design/one-pager-simple-class-selection.md
@@ -1,0 +1,424 @@
+# Simple Resource Class Selection
+* Owner: Nic Cope (@negz)
+* Reviewers: Crossplane Maintainers
+* Status: DRAFT
+
+## Background
+
+In Crossplane 0.3 we introduced support for "provider specific" (aka
+"non-portable", aka "[strongly typed]") resource classes. During the development
+cycle [we identified] that if a claim must explicitly reference a provider
+specific resource class in order to enable dynamic provisioning, then the claim
+itself is now provider specific and no longer portable across providers. We
+solved this issue by introducing [portable resource classes].
+
+A portable resource class is effectively an indirection to a non-portable
+resource class. Portable resource classes can be set as the default for a
+particular namespace, and will be used by any resource claims of their
+corresponding kind that do not specify a resource class for dynamic
+provisioning, or a managed resource for static provisioning. This can be thought
+of as "publishing" non-portable resource classes (which may exist in a distinct
+namespace used to logically group infrastructure) to other namespaces, where
+applications may use them to satisfy claims for their infrastructure
+requirements.
+
+This pattern is flexible and powerful, but Crossplane maintainers and [community
+members] have observed that it's verbose; three distinct Kubernetes resources (a
+claim, portable class, and non-portable class) must exist for dynamic
+provisioning to occur. Community members have also [provided feedback] that
+having to create portable resource classes for every new application namespace
+can be onerous.
+
+## Terminology
+
+Most terminology is covered by the [Crossplane glossary]. Key terms used in this
+document include:
+
+* Crossplane uses a _provider_ to satisfy infrastructure needs. Google Cloud
+  Platform (GCP), Elastic Cloud, and Kubernetes are all providers.
+* A resource is said to be _portable_ if it is not tightly coupled to any one
+  provider. Managed resources like `RDSInstance` are coupled to a provider
+  because they expose provider specific configuration parameters, whereas
+  resource claims like `PostgreSQLInstance` are portable because they do not.
+* _Dynamic provisioning_  is the act of satisfying a _portable_ request for
+  infrastructure (i.e. a resource claim) by automatically provisioning a
+  non-portable managed resource.
+* _Resource classes_ enable dynamic provisioning. A resource class determines
+  what class of managed resource should be created to satisfy a resource claim.
+* _Separation of concern_ happens when an opinionated party can publish classes of
+  infrastructure to be consumed by parties unconcerned with such details.
+  Imagine an organisation in which an infrastructure team supports many product
+  teams who own the lifecycle of their infrastructure. Separation of concerns
+  allows the infrastructure team to define a 'production database' class of
+  service that happens to be a high availability, SSD backed GCP CloudSQL
+  instance. Product teams can then self-service by provisioning a production
+  database and trust that it will be configured appropriately.
+
+Note that two forms (classes?) of resource class currently exist; see [portable
+resource classes] for details.
+
+## Goals
+
+The goal of this proposal is to __make it easier for potential Crossplane users
+to learn and take advantage of the portability and separation of concerns__ that
+are enabled by dynamic provisioning. A _secondary_ goal is to maintain the
+flexibility and power that Crossplane offers complex organisations with many
+application and infrastructure namespaces.
+
+## Proposal
+
+This document proposes Crossplane retire the concept of portable resource
+classes. Resource claims would instead be matched directly to non-portable
+resource classes (referred to henceforth as simply 'resource classes') using
+[label selectors]. Providers, managed resources, and resource classes would be
+cluster scoped rather than namespaced.
+
+```yaml
+apiVersion: database.crossplane.io/v1alpha1
+kind: PostgreSQLInstance
+metadata:
+  namespace: acme-team
+  name: gitlab-database
+spec:
+  classSelector:
+    matchLabels:
+      stack: gitlab
+      grade: experimental
+      region: us-east-1
+```
+
+Consider the hypothetical resource claim above, created by an equally
+hypothetical GitLab [stack]. It declares a need for an experimental grade
+PostgreSQL instance in region us-east-1 that is suitable for use with the GitLab
+stack. Any compatible resource class (whether it be an `RDSInstanceClass` or a
+`CloudSQLInstanceClass`) matching the resource claim's labels would be eligible
+to satisfy it. If more than one such class existed one would be selected at
+random.
+
+This pattern is likely familiar to Kubernetes and to Crossplane users - it's
+used to select which nodes a pod may run on, which pods are endpoints of a
+service, and which Kubernetes cluster a Crossplane workload may run on.
+
+### Infrastructure Scope
+
+Crossplane 0.3 introduced the concept of an 'infrastructure namespace' - a
+logical grouping of infrastructure resources such as providers, resource
+classes, and managed resources. These namespaces might group infrastructure by
+its geographical region, or its environment (e.g. "production"), distinct from
+application namespaces. This potentially many to many application to
+infrastructure namespace pattern - currently established by portable resource
+classes - makes it difficult to mentally model the relationship between
+applications and the infrastructure that supports them.
+
+It is an uncommon pattern in Kubernetes to namespace infrastructure resources.
+Existing infrastructure kinds such as nodes, persistent volumes, and storage
+classes are cluster scoped concepts. This document proposes Crossplane Services
+resources (with the exception of resource claims) become cluster scoped. This
+provides a single, easy to find scope in which resource claim authors can
+discover the resource classes available to them, and the managed resources
+produced by their claim.
+
+Under this proposal the logical groupings provided by infrastructure namespaces
+that we frequently think of as “production”, “development”, or “east-coast” can
+be modelled using labels. One thing we would lose is the ability to bind RBAC
+policies to _groups_ of infrastructure; this [cannot be done using label
+selectors]. Instead RBAC policy would only be applicable at kind (e.g. all
+`RDSInstance` resources) or resource (e.g. the `RDSInstance` named `example`)
+granularity.
+
+### Exact Matching and Static Provisioning
+
+Resource claims would retain their existing `.spec.classRef` field under this
+proposal. In fact the `.spec.classSelector` would be used to set the `classRef`,
+just like Crossplane's [workload scheduling]. Resource claim authors who desired
+dynamic provisioning but had a specific resource class in mind could bypass the
+label selection process by explicitly referencing a resource class, for example:
+
+```yaml
+apiVersion: database.crossplane.io/v1alpha1
+kind: PostgreSQLInstance
+metadata:
+  namespace: acme-team
+  name: gitlab-database
+spec:
+  classRef:
+    apiVersion: database.gcp.crossplane.io/v1alpha2
+    kind: CloudSQLInstance
+    namespace: us-east-1
+    name: gitlab-experimental
+  writeConnectionSecretToRef:
+    name: app-postgresql-connection
+```
+
+Static provisioning is unaffected by this proposal; setting the
+`.spec.resourceRef` field of a resource claim would continue to bypass resource
+classes entirely. The resource claim would attempt to bind to the explicitly
+referenced managed resource.
+
+### Rudimentary Scheduling
+
+While not a goal of this proposal, it's worth noting that the label selector
+pattern enables rudimentary scheduling of resource claims to resource classes.
+Class and claim authors _could_ define a labelling schema that allowed claim
+authors to describe the desired properties of their claim using labels. The
+below example illustrates a labelling schema that could be used to describe a
+`RedisCluster` claim that needs a replicated Redis 3.2 instance running on the
+west coast of the USA:
+
+```yaml
+apiVersion: cache.crossplane.io/v1alpha1
+kind: RedisCluster
+metadata:
+  name: my-cache
+spec:
+  classSelector:
+    matchLabels:
+      engineVersion: "3.2"
+      region: us-west
+      replicated: "true"
+```
+
+These resource classes specify that they can satisfy claims for replicated Redis
+3.2 instances on the west coast of the USA via their labels:
+
+```yaml
+apiVersion: cache.gcp.crossplane.io/v1alpha2
+kind: CloudMemorystoreInstanceClass
+metadata:
+  name: gcp-dev
+  labels:
+    engineVersion: "3.2"
+    region: us-west
+    replicated: "true"
+specTemplate:
+  tier: STANDARD_HA
+  region: us-west2
+  memorySizeGb: 1
+  redisVersion: REDIS_3_2
+---
+apiVersion: cache.azure.crossplane.io/v1alpha2
+kind: RedisClass
+metadata:
+  name: azure-dev
+  labels:
+    engineVersion: "3.2"
+    region: us-west
+    replicated: "true"
+specTemplate:
+  resourceGroupName: group-westus-1
+  location: West US
+  sku:
+    name: Basic
+    family: C
+    capacity: 0
+```
+
+Crossplane could (at some future time) solve the specific example above by
+[teaching each resource claim controller] how to match standardised resource
+claim fields to their provider specific resource class fields, but this is true
+only for a very small class of fields that are both highly relevant to claim
+authors and highly likely to translate to every conceivable managed resource
+that could satisfy a claim. Geographic region, node count, and database size are
+examples of such fields.
+
+Label selectors approximate this functionality while it does not yet exist and
+will compliment it when it does. There will likely be characteristics of
+resource classes that claim authors want to match on that cannot be first class
+configurable fields of the resource claim `spec` because they do translate to
+all managed resources.
+
+### Unopinionated Resource Claims
+
+In the absence of spec based scheduling a resource claim that omits the
+`.spec.classSelector` field, as well as the `.spec.classRef` and
+`.spec.resourceRef` fields that allow them to explicitly specify a class or
+existing managed resource respectively will be deemed unopinionated about what
+resource class it needs.
+
+Kubernetes [states that]:
+
+> The semantics of empty or non-specified selectors are dependent on the
+> context, and API types that use selectors should document the validity and
+> meaning of them.
+
+This document proposes that Crossplane fall back to a default resource class at
+the cluster scope in order to satisfy an unopinionated resource claim. The
+`resourceclass.crossplane.io/is-default-class: "true"` annotation (not label)
+would indicate the default resource class. If there is no default resource class
+the resource claim will not be satisfied and will remain unbound. If there are
+multiple default resource classes one will be chosen at random.
+
+## Technical Implementation
+
+Under the hood, resource class selection for any resource class that specified a
+`classSelector` (and did not specify a `classRef` or `resourceRef`) would be
+implemented as a race between a series of "class scheduler" controllers.
+
+In Crossplane there is not one controller for each resource claim kind, but
+rather one controller for each possible (resource claim, managed resource)
+tuple. This allows an infrastructure stack that adds support for a managed
+resource to implement the dynamic provisioning and claim binding logic for said
+managed resource without having to touch Crossplane core. Put otherwise, the GCP
+stack can enable `CloudSQLInstance` resources to satisfy `MySQLInstance` claims
+without teaching `crossplaneio/crossplane` how to dynamically provision a
+`CloudSQLInstance`. This same pattern would be applied to class scheduler
+controllers - one controller would be responsible for each possible (resource
+claim, resource class) tuple.
+
+In more detail:
+
+* All resource claim controllers are updated to only reconcile resources with
+  either a `resourceRef` or a `classRef` set.
+* A class scheduler controller is introduced for each (resource claim, resource
+  class) tuple. Its job is to set the `classRef` for resource classes that omit
+  it (and omit `resourceRef`), either by using their `classSelector` or the
+  default resource class.
+
+Upon the creation of a `PostgreSQLInstance` without a `classRef` or
+`resourceRef`:
+
+1. Every class scheduler watching for `PostgreSQLInstance` claims has a
+   reconcile queued. This example discusses the `(PostgreSQLInstance,
+   CloudSQLInstanceClass)` scheduler , but `(PostgreSQLInstance,
+   RDSInstanceClass)` and `(PostgreSQLInstance, MysqlServerClass)` schedulers
+   would run through the process in parallel.
+1. The scheduler lists all `CloudSQLInstanceClass` resources that matched the
+   `PostgresSQLInstance` claim's label selectors.
+1. If no `CloudSQLInstanceClass` matched the labels, the reconcile is done.
+   Otherwise, one of the matching `CloudSQLInstanceClass` resources is selected
+   at random.
+1. The scheduler sleeps for a small, randomly jittered amount of time. This
+   increases the randomness of the potential race between controllers to set the
+   claim's `classRef`. Without this jitter it's more likely that one controller
+   consistently wins the race, for example because it must list and "randomly
+   choose" from only one matching resource class while other controllers must
+   choose from many.
+1. The scheduler sets the `classRef` of the `PostgreSQLInstance` to the selected
+   `CloudSQLInstanceClass`. If two controllers try to set the `classRef` at the
+   same time one will fail to commit the change due to the `PostgreSQLInstance`
+   claim's resource version having changed since it was read.
+1. The reconcile is done. With the `classRef` set the `PostgresSQLInstance` now
+   passes the watch predicates of the `(PostgreSQLInstance, CloudSQLInstance)`
+   resource claim reconciler, which dynamically provisions and binds it.
+
+A similar process would be enacted when an unopinionated resource claim - one
+that also omitted its `classSelector` - was encountered. This process would
+implicitly select on the `resourceclass.crossplane.io/is-default-class: "true"`
+annotation rather than labels. This is why Crossplane would not enforce that
+only one default class existed and fall back to selecting one at random when
+multiple existed; because any other behaviour would require coordination between
+resource class controllers.
+
+All portable resource class definitions and defaulting controllers would be
+removed from Crossplane core.
+
+## Future Considerations
+
+The following are explicitly out of scope for this initial proposal, but may be
+considered in future.
+
+### Enforcing a Single Default Class
+
+Under the proposal put forward by this document Crossplane would allow multiple
+classes to be annotated as the default, and pick one at random if there were
+more than one. This is a strategy informed by technical limitations rather than
+desired user experience - it requires no coordination between claim scheduler
+controllers. It this behaviour proved problematic in practice it should be
+possible to enforce that only a single default resource class exist with little
+coordination by employing a series of admission webhooks. When a resource class
+of a particular kind was annotated as the default its admission webhook would
+take the lease, alerting other admission webhooks for other resource class kinds
+that they should not allow a default class to be set.
+
+### Weighted Classes
+
+It may be possible to improve upon the random scheduling of resource claims to
+resource classes by supporting weighted random scheduling. Under such a proposal
+resource classes would have a spec field or annotation specifying their weight,
+with the highest weight winning. This could be used by infrastructure
+administrators to ensure conservative (or affordable) behaviour in cases where a
+resource claim matches multiple resource classes.
+
+Consider the following resource claim:
+
+```yaml
+apiVersion: cache.crossplane.io/v1alpha1
+kind: RedisCluster
+metadata:
+  namespace: acme-team
+  name: gitlab-cache
+spec:
+  classSelector:
+    matchLabels:
+      stack: gitlab
+      region: us-west
+```
+
+Matched against the below resource classes:
+
+```yaml
+apiVersion: cache.gcp.crossplane.io/v1alpha2
+kind: CloudMemorystoreInstanceClass
+metadata:
+  name: really-quite-a-lot-of-memory
+  labels:
+    stack: gitlab
+    region: us-west
+    grade: production
+specTemplate:
+  tier: STANDARD_HA
+  region: us-west2
+  memorySizeGb: 99999999999999999999999
+  redisVersion: REDIS_3_2
+---
+apiVersion: cache.gcp.crossplane.io/v1alpha2
+kind: CloudMemorystoreInstanceClass
+metadata:
+  name: cheap-redis
+  labels:
+    stack: gitlab
+    region: us-west
+    grade: experimental
+specTemplate:
+  tier: STANDARD_HA
+  region: us-west2
+  memorySizeGb: 1
+  redisVersion: REDIS_3_2
+```
+
+Both classes match the resource claim's desired properties, but one is much more
+expensive than the other. Given the resource claim did not specifically request
+a production database, it should probably be scheduled to use the cheaper class.
+Weighting could be used to approximate the "cost" of resource classes - higher
+weights are cheaper (or safer) resource classes.
+
+Note that anything other than random scheduling would require the various class
+scheduler controllers to collaborate, increasing the complexity of the system.
+In the case of weighted scheduling each controller would need to determine the
+highest weighted matching resource class, then determine whether its highest
+weight was higher than that chosen by any other scheduler controller.
+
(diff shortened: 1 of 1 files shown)


design/one-pager-class-selection.md
```diff
@@ -0,0 +1,371 @@
...
+   would run through the process in parallel.
+1. The scheduler lists all `CloudSQLInstanceClass` resources that matched the
+   `PostgresSQLInstance` claim's label selectors.
+1. If no `CloudSQLInstanceClass` matched the labels, the reconcile is done.
+   Otherwise, one of the matching `CloudSQLInstanceClass` resources is selected
+   at random.
+1. The scheduler sets the `classRef` of the `PostgreSQLInstance` to the selected
+   `CloudSQLInstanceClass`. If two controllers try to set the `classRef` at the
+   same time one will fail to commit the change due to the `PostgreSQLInstance`
+   claim's resource version having changed since it was read.
```
negz: I initially thought we might need to artificially jitter this process to avoid conflicts, but @suskin pointed out that's not necessary. In retrospect, it may be worth jittering the process simply to increase randomness. It's likely that in cases where one scheduler controller needs to list and randomly select from 100 resource classes while another only has to list and select from 1 resource class the controller doing the least work would consistently win the race.

design/one-pager-class-selection.md
```diff
@@ -0,0 +1,371 @@
...
+metadata:
+  name: acme-team
+  annotations:
+    resource-class.crossplane.io/select-namespaces: "acme-team, experimental-infra"
+```
+
+The above `acme-team` application namespace declares that any resource claims
+created within it should be matched against resource classes from namespaces
+`acme-team` or `experimental-infra`. Note that when the `select-namespaces`
+annotation exists it overrides the default behaviour of selecting only within
```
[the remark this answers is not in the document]
hasheddan: After finishing my read the doc, it appears that without some sort of weighting implemented, then there is really no such thing as a traditional `default` anymore. This is in-line with this comment https://github.com/crossplaneio/crossplane/issues/870#issuecomment-537396943

> Strongly suggesting that by convention resource claims with no opinion use a label selector like `{"default": "true"}`

So I guess my question above boils down to how do we handle the scenario where there are multiple classes of the same `kind` that match labels but are in different namespaces that have been specified in the `select-namespaces` annotation. The `Technical Implementation` section below covers the case that two separate controllers both find a satisfactory class (race condition), but not where a single controller finds multiple satisfactory classes in different namespaces.

design/one-pager-class-selection.md
```diff
@@ -0,0 +1,371 @@
...
+```
+
+Consider the hypothetical resource claim above, created by an equally
+hypothetical GitLab [stack]. It declares a need for an experimental grade
+PostgreSQL instance in region us-east-1 that is suitable for use with the GitLab
+stack. Any compatible resource class (whether it be an `RDSInstanceClass` or a
+`CloudSQLInstanceClass`) matching the resource claim's labels would be eligible
+to satisfy it. If more than one such class existed one would be selected at
+random. A resource claim that omitted `.spec.classSelector`, `.spec.classRef`,
+and `.spec.resourceRef` would be deemed to be completely unopinionated and would
```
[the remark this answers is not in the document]
negz: That's a good question. I could definitely see the case that if there's an existing, unbound, unblemished managed resource sitting around it could be worth skipping dynamic provisioning and using that managed resource. Perhaps dynamically provisioned managed resources should inherit the labels of their resource classes to enable this at some point?

I agree it's probably best to keep it out of scope for the moment. It would be a forcing function to codify our feelings around https://github.com/crossplaneio/crossplane/issues/539.

design/one-pager-class-selection.md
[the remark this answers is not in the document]
negz: > How do we handle the scenario where there are multiple classes of the same kind that match labels but are in different namespaces that have been specified in the select-namespaces annotation.

We pick one at random. The selection process within the global set of matching resource classes is effectively:

1. One of each kind of eligible resource class is picked at random.
1. One of the kinds randomly selected in class one is picked at random.

I'd argue that this means that it's still _possible_ to support the "defaulting" pattern that we do today - but doing so requires a convention to be established when writing resource classes and claims. Resource claim authors that wanted the default would have to select on some label (i.e. `default: "true"), and resource class authors would have to ensure only one class was labelled the default within each matching scope. The latter constraint exists today - you may only label one portable resource class as the default. The issue is what happens when multiple classes are labelled as the default - under this proposal you get a randomly selected default whereas today you get an error. I do see that this could be a [footgun](https://en.wiktionary.org/wiki/footgun), but I don't think it's any more a footgun than existing instances of label selection and scheduling within Kubernetes.

I do wonder how many folks will still choose to use simple defaulting when they have a slightly more expressive language to describe things like "I want the default for gitlab", or similar.

design/one-pager-class-selection.md
```diff
@@ -0,0 +1,371 @@
...
+pattern would be enabled by creating an association between application and
+infrastructure namespaces using namespace level annotations. For example:
+
+```yaml
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: acme-team
+  annotations:
+    resource-class.crossplane.io/select-namespaces: "acme-team, experimental-infra"
```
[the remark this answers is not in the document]
negz: No strong opinion there.

design/one-pager-class-selection.md
```diff
@@ -0,0 +1,371 @@
...
+a production database, it should probably be scheduled to use the cheaper class.
+Weighting could be used to approximate the "cost" of resource classes - higher
+weights are cheaper (or safer) resource classes.
+
+Note that anything other than random scheduling would require the various class
+scheduler controllers to collaborate. In the case of weighted scheduling each
+controller would need to determine the highest weighted matching resource class,
+then determine whether its highest weight was higher than that chosen by any
+other scheduler controller.
+
```
negz: Another slightly pie-in-the-sky future consideration. If resource claims are matched to resource classes using labels, and don't expose any actual configuration fields for the resource claim author to set, is there a point in resource claims being of distinct kinds? i.e. Could we have simply `ResourceClaim` rather than `PostgreSQLInstance`?

This would look something like:

```yaml
apiVersion: crossplane.io/v1alpha1
kind: ResourceClaim
metadata:
  namespace: acme-team
  name: gitlab-cache
spec:
  classSelector:
    matchLabels:
      protocol: postgresql
      stack: gitlab
      region: us-west
```

This direction would remove the need to add new resource claim kinds to Crossplane core in order to support new kinds of managed resources, and would remove any distinction between "portable" and "non-portable" resource claims - you could dynamically provision a hypothetical `DynamoDB` using `ResourceClaim` just as easily is you could provision a `PostgreSQLInstance`.

design/one-pager-class-selection.md
```diff
@@ -0,0 +1,371 @@
...
+    apiVersion: database.gcp.crossplane.io/v1alpha2
+    kind: CloudSQLInstance
+    namespace: us-east-1
+    name: gitlab-experimental
+  writeConnectionSecretToRef:
+    name: app-postgresql-connection
+```
+
+Static provisioning is unaffected by this proposal; setting the
+`.spec.resourceRef` field of a resource claim would continue to bypass resource
```
soorena776: Is `.spec.resourceRef` an alias `.spec.classRef` in the example? If so, a clarification could help the reader.

design/one-pager-class-selection.md
```diff
@@ -0,0 +1,371 @@
...
+classes entirely. The resource claim would attempt to bind to the explicitly
+referenced managed resource.
+
+### Rudimentary Scheduling
+
+While not a goal of this proposal, it's worth noting that it begins to enable
+scheduling of resource claims to resource classes by decoupling the language
+used to match claims to classes from the (non-portable) language used to
+describe resource class parameters.
```
soorena776: nit: consider rephrasing, I had to read multiple times.

design/one-pager-class-selection.md
```diff
@@ -0,0 +1,371 @@
...
+
+It is challenging to match the requirements of a resource claim to the
+capabilities of a resource class on `spec` alone due to the "lowest common
+denominator" problem of multi cloud. The `RedisCluster` claim, for example,
```
soorena776: nit: could we add a link
https://thenewstack.io/avoiding-least-common-denominator-approach-hybrid-clouds/

design/one-pager-class-selection.md
```diff
@@ -0,0 +1,371 @@
...
+removed from Crossplane core.
+
+## Future Considerations
+
+It may be possible to improve upon the random scheduling of resource claims to
+resource classes by supporting weighted random scheduling. Under such a proposal
+resource classes would have a spec field or annotation specifying their weight,
+with the highest weight winning. This could be used by infrastructure
+administrators to ensure conservative (or affordable) behaviour in cases where a
+resource claim matches multiple resource classes.
```
soorena776: I like this approach over randomly choosing one resource class.

design/one-pager-class-selection.md
```diff
@@ -0,0 +1,371 @@
...
+1. The reconcile is done. With the `classRef` set the `PostgresSQLInstance` now
+   passes the watch predicates of the `(PostgreSQLInstance, CloudSQLInstance)`
+   resource claim reconciler, which dynamically provisions and binds it.
+
+All portable resource class definitions and defaulting controllers would be
+removed from Crossplane core.
+
+## Future Considerations
+
+It may be possible to improve upon the random scheduling of resource claims to
```
soorena776: since you only have one subject for future consideration, maybe you could  add more specific headline, or make a section like `Weighted Scheduling`

design/one-pager-class-selection.md
soorena776: The generic notion of this idea is appealing, but I feel like this is just moving strongly typed fields and assumptions from the type implementation  to, `classSelector` as "weak" fields, which brings out analogous challenges we faced in generic resource classes.

negz:
@soorena776 (Quotes edited a little for brevity)

> I would also expect a workload to run and operate the same on different cloud providers, without any modifications.  I am a bit concerned that this will be a beginning to workload modification depending where it is being deployed, and infrastructure details leaks into the workload metadata.

Can you elaborate on how using label selectors might leak more details into workload metadata than we do today? Keep in mind:

1. Contemporary resource classes already allow the author to specify fields like `engineVersion`, `bucketACL`, etc.
1. We're not opinionated about how the label schema must be used. i.e. You don't have to match on labels like `us-east-1`. You could match on `{workload: gitlab, default: "true"}` for similar behaviour to today's "just use the default class for this workload" behaviour. 

> Could we at least don't retire the existing concept of namespace level non-portable resource classes which enable defining completely portable claims?

A big part of this design is to _simplify_ things. I think if this design results in us:

1. Introducing new concepts.
2. Removing no concepts.

Then it has failed at that goal, and in fact done the opposite. For that reason I don't think we should keep portable resource classes around if we move forward with label selector derived claim->class relationships.

design/one-pager-class-selection.md
hasheddan: @negz one drawback to this is that stack developers basically have to coordinate with each other to say what the protocol value is that they are going standardize on. Since there is no centralized `PostrgreSQLInstance` in core crossplane, developers on stack-aws and stack-gcp need to make sure they both accept `protocol: postgresql` for `RDSInstance` and `CloudSQLInstance` respectively, and not one accepts `protocol: postgres` or something. I do like the idea in theory that we do not have to add new claim types though (plus less CRDs is always a plus IMO). There would be incentive for stack developers to standardize with this pattern, but I am not sure we want to put the burden on them to look around and see what everyone else is doing rather than just knowing what core crossplane claim types are available.

negz: I think I'm going to close this thread out to prevent it distracting too much from the rest of the proposal. Definitely agree with the feedback.

design/one-pager-class-selection.md
negz: We should! I wanted to but couldn't immediately find an article that describes the concern as well as this one does. Thanks!

soorena776:
@negz The scenario that I am having in mind is when you have the WordPress [workload] (https://github.com/crossplaneio/crossplane/blob/master/cluster/examples/workloads/kubernetes/wordpress/gcp/workload.yaml), and you want to apply it in different cloud providers. I would like to be use this file without any modification (think of it as part of a packaged solution) to different environments that I have. The current labeling proposal would require me to modify this workload and add labels to the [MySql Claim](https://github.com/crossplaneio/crossplane/blob/master/cluster/examples/workloads/kubernetes/wordpress/gcp/workload.yaml#L13)

As for achieving the simplification goal, one way to think about it would be to say we now simplify how one could tie a claim to a resource class, but also have more advanced ways of doing it, if you really wanted to keep your workloads truly portable across providers.

negz:
> The current labeling proposal would require me to modify this workload and add labels to the MySql Claim class selector.

It sounds like by workload we're talking about the things an application stack might create. e.g. A controller installed by a Wordpress app creates a `KubernetesApplication` and a `MySQLInstance` for every `Wordpress` a user creates. Currently that `MySQLInstance` can either:


a. Specify a portable class, by name:
```yaml
kind: MySQLInstance
metadata:
  namespace: acme-team
  name: wordpress-database-239df
spec:
  classRef:
    wordpress-database-class  # A portable class in the acme-team namespace
```

b. Rely on a default resource class to exist:

```yaml
kind: MySQLInstance
metadata:
  namespace: acme-team
  name: wordpress-database-239df
spec:
  # No classRef or resourceRef. A default portable class is selected if exactly one exists
```

Under this proposal, it could instead write the below, which seems no less portable than option a.:

```yaml
kind: MySQLInstance
metadata:
  namespace: acme-team
  name: wordpress-database-239df
spec:
  classSelector:
    wordpress-database-default: "true"
```

It's not obvious to me how this is less portable than the status quo. Perhaps it's not clear that I'm not proposing any specific labelling schema be required? Any labels I've included in the design doc are just examples of patterns an organisation or stack author could use.

soorena776:
@negz I think I have a better understanding of your vision now. I was under the impression that the labels are first given to the resource classes, and then related claims'  `classSelector` are populated. However this could be also done as not touching the clais, but labeling the resources classes (non-portable layer) accordingly. Thanks for the explanation!

design/one-pager-class-selection.md
```diff
@@ -0,0 +1,371 @@
...
+document include:
+
+* Crossplane uses a _provider_ to satisfy infrastructure needs. Google Cloud
+  Platform (GCP), Elastic Cloud, and Kubernetes are all providers.
+* A resource is said to be _portable_ if it is not tightly coupled to any one
+  provider. Managed resources like `RDSInstance` are coupled to a provider
+  because they expose provider specific configuration parameters. Resource
+  claims like `PostgreSQLInstance` are portable because they do not.
+* _Dynamic provisioning_  is the act of satisfying a _portable_ request for
+  infrastructure (i.e. a resource claim) by automatically provisioning a
```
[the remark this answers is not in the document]
negz: I don't think it's intuitive to the uninitiated what "dynamic" means in the context of provisioning. Given this sentence is attempting to _define_ what dynamic means in this context, I think automatic is appropriate.

design/one-pager-class-selection.md
```diff
@@ -0,0 +1,371 @@
...
+  infrastructure to be consumed by parties unconcerned with such details.
+  Imagine an organisation in which an infrastructure team supports many product
+  teams who own the lifecycle of their infrastructure. Separation of concerns
+  allows the infrastructure team to define a 'production database' class of
+  service that happens to be a high availability, SSD backed GCP CloudSQL
+  instance. Product teams can then self-service by provisioning a production
+  database and trust that it will be configured appropriately.
+* Crossplane recommends using Kubernetes namespaces to logically group
+  infrastructure into an _infrastructure namespace_, for example
+  `production-infra` or `us-east-coast`. This infrastructure could be claimed
```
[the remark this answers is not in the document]
negz: This will be less relevant in the impending update to the design so I'm just going to resolve this for now. :P

design/one-pager-class-selection.md
```diff
@@ -0,0 +1,371 @@
...
+provisioning to occur. Community members have also [provided feedback] that
+having to create portable resource classes for every new application namespace
+can be onerous.
+
+## Terminology
+
+Most terminology is covered by the [Crossplane glossary]. Key terms used in this
+document include:
+
+* Crossplane uses a _provider_ to satisfy infrastructure needs. Google Cloud
```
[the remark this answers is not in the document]
negz: In this case I'm talking about "provider" in the general sense, rather than as a the CR kind named `Provider`. I think it's useful to define here because we talk about portability across providers a lot in this design, which is my general take on the glossary - surface some terms that are important to the specific design in question and defer to the glossary for the rest.

design/one-pager-class-selection.md
negz: Likewise, but I don't think we need it for the first pass.

design/one-pager-class-selection.md
negz: Added some context at the first mention of these fields.

design/one-pager-class-selection.md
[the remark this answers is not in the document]
negz: No longer relevant in the latest iteration - though for the default class annotation I went with `resourceclass` to more closely match the default `storageclass` annotation.

design/one-pager-simple-class-selection.md
```diff
@@ -0,0 +1,411 @@
...
+  infrastructure to be consumed by parties unconcerned with such details.
+  Imagine an organisation in which an infrastructure team supports many product
+  teams who own the lifecycle of their infrastructure. Separation of concerns
+  allows the infrastructure team to define a 'production database' class of
+  service that happens to be a high availability, SSD backed GCP CloudSQL
+  instance. Product teams can then self-service by provisioning a production
+  database and trust that it will be configured appropriately.
+* Crossplane recommends using Kubernetes namespaces to logically group
+  infrastructure into an _infrastructure namespace_, for example
+  `production-infra` or `us-east-coast`. This infrastructure could be claimed
```
prasek: should we remove this bullet since there will only be cluster-scoped infra going forward, or qualify it with Crossplane v0.3 recommends using infrastructure namespaces?

design/one-pager-simple-class-selection.md
```diff
@@ -0,0 +1,411 @@
...
+```yaml
+apiVersion: database.crossplane.io/v1alpha1
+kind: PostgreSQLInstance
+metadata:
+  namespace: acme-team
+  name: gitlab-database
+spec:
+  classSelector:
+    matchLabels:
+      stack: gitlab
```
prasek: does the `stack:gitlab` label  example recommend a pattern for stack-specific, cluster-scoped resource classes be created for each type of stack in the cluster?

design/one-pager-simple-class-selection.md
```diff
@@ -0,0 +1,411 @@
...
+
+In Crossplane 0.3 we introduced support for "provider specific" (aka
+"non-portable", aka "[strongly typed]") resource classes. During the development
+cycle [we identified] that if a claim must explicitly reference a provider
+specific resource class in order to enable dynamic provisioning, then the claim
+itself is now provider specific and no longer portable across providers. We
+solved this issue by introducing [portable resource classes].
+
+A portable resource class is effectively an indirection to a non-portable
+resource class. Portable resource classes can be set as the default for a
```
prasek: does this proposal also eliminate per namespace defaults, such that only cluster-scoped defaults will be available going forward?

design/one-pager-simple-class-selection.md
```diff
@@ -0,0 +1,411 @@
...
+applications and the infrastructure that supports them.
+
+It is an uncommon pattern in Kubernetes to namespace infrastructure resources.
+Existing infrastructure kinds such as nodes, persistent volumes, and storage
+classes are cluster scoped concepts. This document proposes Crossplane Services
+resources (with the exception of resource claims) become cluster scoped. This
+provides a single, easy to find scope in which resource claim authors can
+discover the resource classes available to them, and the managed resources
+produced by their claim. The logical groupings provided by infrastructure
+namespaces that we frequently think of as “production”, “development”, or
```
prasek: one consideration here is that infra namespaces support simpler RBAC for a prod-infra vs. dev-infra namespaces, reducing the likelihood of a shared cluster / control plane environments where one group can define the dev/test resource classes to experiment with new settings, and a more limited group has access to prod-infra.

design/one-pager-simple-class-selection.md
```diff
@@ -0,0 +1,411 @@
...
+
+```yaml
+apiVersion: cache.crossplane.io/v1alpha1
+kind: RedisCluster
+metadata:
+  name: my-cache
+spec:
+  classSelector:
+    matchLabels:
+      engineVersion: "3.2"
```
prasek: this is a very compelling aspect of labels/selectors and offers users a lot of flexibility in modeling different label schemas

design/one-pager-simple-class-selection.md
```diff
@@ -0,0 +1,411 @@
...
+
+Crossplane could (at some future time) solve the specific example above by
+[teaching each resource claim controller] how to match standardised resource
+claim fields to their provider specific resource class fields, but this is true
+only for a very small class of fields that are both highly relevant to claim
+authors and highly likely to translate to every conceivable managed resource
+that could satisfy a claim. Geographic region, node count, and database size are
+examples of such fields.
+
+Label selectors approximate this functionality while it does not yet exist and
```
prasek: agree, this is great

design/one-pager-simple-class-selection.md
```diff
@@ -0,0 +1,411 @@
...
+existing managed resource respectively will be deemed unopinionated about what
+resource class it needs.
+
+Kubernetes [states that]:
+
+> The semantics of empty or non-specified selectors are dependent on the
+> context, and API types that use selectors should document the validity and
+> meaning of them.
+
+This document proposes that Crossplane fall back to a default resource class - a
```
prasek: suggest pointing out that class defaults are not cluster-scoped defaults, not per app project namespace as they were with portable classes.

design/one-pager-simple-class-selection.md
negz: Good catch - I actually thought I already deleted this. My thinking was that I'd remove this and rely on the brief mention of infrastructure namespaces at line ~20 and under "Infrastructure Scope" to provide this context.

design/one-pager-simple-class-selection.md
negz: That's what I was thinking, though I'd frame it more as an example of a potential pattern or maybe a _speculative_ recommendation.

I could see this either working as:

* The Gitlab stack installs a beginner resource class (or two) that the resource claims  produced by new `GitLab` resources match by default.
* The Gitlab stack documents that new `GitLab` instances will create resource claims that match on something like these labels, and it's up to the user to make sure there are classes that would match these labels?

Not sure how prescriptive we should be about this, but it would at least open the possibility to stack authors.

design/one-pager-class-selection.md
muvaf: I feel like it's not random if it depends on how much it takes to execute a block of code. But, given the controller-tuples structure, I see that it may not be technically feasible to build a shared list of classes and choose from that list. On the other hand, fallback to random doesn't seem to be a very frequent case to me, but, I think it's worth opening an issue to track a fix for that bias.

design/one-pager-simple-class-selection.md
```diff
@@ -0,0 +1,411 @@
...
+unbound. If there are multiple default resource classes one will be chosen at
+random.
+
+## Technical Implementation
+
+Under the hood, resource class selection for any resource class that specified a
+`classSelector` (and did not specify a `classRef` or `resourceRef`) would be
+implemented as a race between a series of "class scheduler" controllers.
+
+In Crossplane there is not one controller for each resource claim kind, but
```
muvaf: How about we allow managed resource CRDs to have label that they can tell what resource claim they can be used for? `CloudSQLInstance` CRD can have a label `crossplane.io/resource-claim: MySqlInstance`. One generic `MySqlInstance` controller could get all CRDs with that label, lists all instance of these CRDs and pick one randomly. Pretty much like kinds would register themselves to a portable resource claim kind as candidate.

design/one-pager-simple-class-selection.md
negz: This is true - updated the doc to capture this.

design/one-pager-simple-class-selection.md
negz: Done.

design/one-pager-simple-class-selection.md
negz: This is an intriguing idea. I like that it would allow us to assess all possible classes that could satisfy a claim in one place (instead of assessing each kind of class separately). If we did want to have classes effectively register support for a kind of claim, I think the CRD is a good place to do so.

Two potential issues I could see, though:

* If we did want to expand to making claim scheduling decisions by comparing the spec of the claim to the spec of the class we'd probably need whatever was doing that comparison to have access to the class's Go struct. It would probably be _possible_ to do this using `*unstructured.Unstructured`, but we'd be working with raw JSON (informed by the CRD schema) to make scheduling decisions.
* If/when we introduced non-portable claims like a `DynamoDb` claim, the scheduling/defaulting controllers would like in the stack where that claim was defined, while portable claims like `MySQLInstance` would have their scheduling/defaulting controllers defined in Crossplane runtime. It's probably a stretch to call this an "issue" - more of an asymmetry.

negz:
> If I'm not mistaken, this proposal doesn't prevent me from using any resource class I want. 

That's true. It's worth noting it was also true in Crossplane 0.2, but is a regression in access control compared to 0.3 similar to the regression @prasek pointed out regarding RBAC access. I think we're trading a little flexibility for a little simplicity here.

> Though it's technically more challenging, a mutating webhook could be considered.

Good point - I think a controller will work fine to begin with and is the path of least resistance, but I agree a webhook would probably make more sense. I'll add this to future considerations.

negz:
Going to merge this but leave it in draft until we've worked through the technical implementation a little more.

design/one-pager-simple-class-selection.md
muvaf: > If we did want to expand to making claim scheduling decisions by comparing the spec of the claim to the spec of the class

I don't understand why we'd need to do that comparison. In my imagination, this controller's job would be just to choose a class and put its reference to `classRef` of `MySqlInstance` claim that it reconciles. The type of the class would be unknown but only thing needed for referencing is metadata anyway.

> It's probably a stretch to call this an "issue" - more of an asymmetry.

I agree. On the positive side, stack developers can implement their own unique scheduling mechanism for non-portable claims if they really wanted to.

PR #927, opened 2019-10-10
Support Simple Resource Class Selection

negz:

This enables label based class matching per https://github.com/crossplaneio/crossplane/pull/926. Changes will also be required to crossplane-runtime, and all stacks.

### Checklist

I have:


apis/cache/v1alpha1/rediscluster_types.go
```diff
@@ -39,14 +30,14 @@ type RedisClusterSpec struct {
...
-// +kubebuilder:printcolumn:name="VERSION",type="string",JSONPath=".spec.engineVersion"
+// +kubebuilder:printcolumn:name="RESOURCE",type="string",JSONPath=".spec.resourceRef.name"
 // +kubebuilder:printcolumn:name="AGE",type="date",JSONPath=".metadata.creationTimestamp"
 // +kubebuilder:subresource:status
 type RedisCluster struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
-	Spec   RedisClusterSpec                    `json:"spec,omitempty"`
+	Spec   runtimev1alpha1.ResourceClaimSpec   `json:"spec,omitempty"`
```
hasheddan: This will need to be updated in `crossplane-runtime` [here](https://github.com/crossplaneio/crossplane-runtime/blob/15eb580eecd8b4a91fe0a72cb255b6ec6a0f55cf/apis/core/v1alpha1/resource.go#L63) so that it includes a `corev1.ObjectReference` instead of `corev1.LocalObjectReference`. We should probably change name back to `ClassReference` instead of `PortableClassReference` as well.

hasheddan: Also, the `Claim` interface will need to be updated [here](https://github.com/crossplaneio/crossplane-runtime/blob/15eb580eecd8b4a91fe0a72cb255b6ec6a0f55cf/pkg/resource/interfaces.go#L90) to be a `ClassReferencer` that can get and set a `corev1.ObjectReference` and angryjet will need to be updated [here](https://github.com/crossplaneio/crossplane-tools/blob/2e3995c538e39c12342474024cc0e18d047e6667/internal/method/method.go#L160) to generate the corresponding methods in `zz_generated.claim.go`.

apis/kubernetes/v1alpha1/types.go
```diff
@@ -17,26 +17,27 @@ limitations under the License.
...
 // A ProviderSpec defines the desired state of a Provider.
 type ProviderSpec struct {
-	// Kubernetes credentials Secret reference
 	// A Secret containing connection credentials for a Kubernetes cluster
 	// client that will be used to authenticate to this Kubernetes Provider.
 	// This will typically be the connection secret of a KubernetesCluster claim,
 	// or the secret created by a Kubernetes service account, but could also be
 	// manually configured to connect to a preexisting cluster.
-	Secret corev1.LocalObjectReference `json:"credentialsSecretRef"`
+	Secret v1alpha1.SecretKeySelector `json:"credentialsSecretRef"`
```
hasheddan:
```suggestion
	Secret v1alpha1.SecretReference `json:"credentialsSecretRef"`
```

hasheddan: This secret will contain k8s client config data so the key portion of the `SecretKeySelector` will always go unused (i.e. we are interested in multiple keys). ref: https://github.com/crossplaneio/stack-rook/blob/b6df7a99300a77a949ab7b4f18dcd839f0b3b802/pkg/clients/client.go#L33

Issue #877, opened 2019-10-04
Produce a one-pager design for a Template Stack Controller


displague:
### What problem are you facing?

Early thinking on Stacks included the idea of simplifying the overhead involved in life-cycle management for complex workloads, especially as this relates to consuming managed services. However, to contribute a new Stack in the Stack eco-system today, it is necessary to create a new Controller, typically in Go, typically using Kubebuilder.

We believe that a more robust Stack ecosystem would emerge if the requirement to write new controllers was removed.

Clearly, there must be a controller or actor at some level to operate on state, within a cluster or outside of a cluster.

### How could Crossplane help solve your problem?

A template controller would be some process that could examine template containing records (Custom Resources) within a cluster. It would then expand those records into new records rendered by some templating engine. The templates would be processed using the template itself and some set of user supplied data and some set of cluster state.

Some questions to consider in this design:

* Can templates be reused?
* How should variables be represented in a template?
* How should the value for variables be supplied? Where are template values stored?
* What sets of variables are available and should be offered? How are their values validated?
* How should a set of variables be grouped together? What is the hierarchy of variables available for use within templates?
* What process creates and manages the objects rendered from the template? Is this a new or sub-controller?

Part of https://github.com/crossplaneio/crossplane/issues/853

Issue #879, opened 2019-10-04
Package versioning and upgrading


displague:
### What problem are you facing?

* Stacks need to be versioned.
* Stacks need to be able to coexist with other versions of Stacks.
* Stacks need to be upgradeable.

Consider a GitLab Stack is released for version 12.3 and Gitlab 12.4 is released.  How does a namespaced Stack package get updated and how does the stack-manager (or other service) update the components of that stack?  

Consider a Google infrastructure Stack installed that needs to be updated.  How does a user manipulate StackInstall records to declare this need? How are the resources of that Stack affected? Can multiple versions of a Cluster scoped Stack co-exist?

Can a Stack be downgraded?

How does a version change affect the components of the stack, including:
* Stack resources (metadata and icons, whose annotations reside in CRD)
* CRD files
* Deployments or Jobs

What happens when StackA and StackB depend on a CRD provided by StackX but an update to StackA warrants an update to StackX version A, which would be incompatible with StackB?

Updates may also be mindful of application resources that must be updated, unless this can be deferred to the application itself:

* Database schema migrations

Is there a particular versioning scheme that must be used? What is the significance of the components of the version and what are the expectations for developers and users.

Do upgrades necessitate design thinking around dependency management? (What happens when an upgrade requires upgraded dependencies?)

### How could Crossplane help solve your problem?

It should be possible to upgrade a Stacks so there are few if any situations that would require a user to delete and re-install a Stack.

### Related Issues

* #435 Design how Stacks should define versions and apply upgrades
  * #900 Design how Stacks format is versioned and updated
* TBD Design how different versions of Stacks can coexist in a single cluster
* #434 Design dependency management and subscription based upgrades
* TBD Implement Stack versioning and upgrading
* TBD Document Stack versioning and upgrading
* TBD Provide examples or blog posts illustrating the usage pattern

prasek:
@hasheddan @negz with the v2 package manager now supporting upgrade are we good to close this out?

hasheddan:
@prasek yep!

PR #928, opened 2019-10-10
One pager for Template Stacks

displague:

This PR introduces the concept and benefits of Template Stacks as well as exploring a proposed Template Stack Manager which may facilitate this experience.  Implementation challenges for a Template Stack Manager and approaches to overcome them are provided.

Throughout the design, Go templates are used to illustrate potential usage, but this is an implementation detail that will be the subject of additional design documents.

Addresses #877 

`[skip ci]` 🚫🤖

### Checklist

I have:

### DIFF
--- a/design/one-pager-template-stacks.md
+++ b/design/one-pager-template-stacks.md
@@ -1,0 +1,1110 @@
+# Templates Stacks
+
+* Owner: Marques Johansson (@displague)
+* Reviewers: Crossplane Maintainers
+* Status: Draft, revision 1.0
+
+## Outline
+
+* [Background](#Background)
+* [Terms and Acronyms](#Terms-and-Acronyms)
+* [Benefits of Template Stacks](#Benefits-of-Template-Stacks)
+  * [Simple Variable Substitution](#Simple-Variable-Substitution)
+  * [No Custom Controller Needed](#No-Custom-Controller-Needed)
+  * [Composition](#Composition)
+* [Template Stack Manager](#Template-Stack-Manager)
+  * [Deployment](#Deployment)
+  * [Owner References](#Owner-References)
+  * [RBAC](#RBAC)
+  * [Life Cycle and Updates](#Life-Cycle-and-Updates)
+  * [Unknown Types](#Unknown-Types)
+* [Examples](#Examples)
+* [Related Issues](#Related-Issues)
+
+## Background
+
+Early thinking on Stacks included the idea of simplifying the overhead involved
+in life-cycle management for complex workloads, especially as this relates to
+consuming Crossplane managed services. However, to contribute a new Stack in the
+Stack eco-system today, it is necessary to create a new Controller, typically in
+Go using Kubebuilder.
+
+We believe that a more robust Stack ecosystem would emerge if the requirement to
+write new controllers was removed. The experience for users would remain the
+same, while developers would have a simpler alternative for simpler use-cases.
+
+Clearly, there must be a controller or actor at some level to produce the
+desired state, operating within a cluster or outside of a cluster.
+
+A template controller would be some process that could examine template
+containing records (Custom Resources) within a cluster. It would then expand
+those records into new records rendered by some template engine. The templates
+would be processed using the template itself and some set of user supplied data
+and some set of cluster state.
+
+## Terms and Acronyms
+
+* **Stacks**
+  : Crossplane Stacks provide a means of deploying isolated and appropriately
+  role restricted `Deployment` resources to manage a set of CRDs.
+  
+  UNDER THIS PROPOSAL, additional features would be added.
+* **Crossplane Control Cluster**
+  : Any Kubernetes cluster where a Crossplane Deployment is installed. This does
+  not include Crossplane Managed Clusters
+  
+* **Crossplane Managed Cluster**
+  : A Kubernetes cluster being managed by Crossplane, which may or may not have
+  Crossplane installed within that cluster
+* **SM (Stack Manager)** : The cluster priviliged deployment that manages
+  `Stack`, `ClusterStackInstall` and `StackInstall` resources. One per
+  Crossplane cluster. The Stack Manager creates `Deployment` resources as
+  defined by a `Stack` to handle resources of the CRD types that `Stack`
+  manages. The Stack Manager creates `Stack` resources based on `StackInstall`
+  and `ClusterStackInstall` resources.
+* **CRD (Custom Resource Definition)** : A standard Kubernetes Custom Resource
+  Definition
+* **CR (Custom Resource)** : An instance of a Kubernetes type that was defined
+  using a CRD
+* **GVK (Group Version Kind)** : The API Group, Version, and Kind for a type of
+  Kubernetes resource (including CRDs)
+* **TS (Template Stack)** : UNDER THIS PROPOSAL, a Stack that declares dependent
+  resources and includes managed CRDs, but includes templates to be rendered for
+  each CR rather than a `Deployment` controller
+* **TSM (Template Stack Manager)** : UNDER THIS PROPOSAL, a restricted
+  `Deployment` that manages resources for a Template Stack's CRDs
+
+## Introduction
+
+### In Scope
+
+This design introduces the concept and benefits of Template Stacks as well as
+exploring a proposed Template Stack Manager which may facilitate this
+experience. Implementation details and challenges for the existing Stack Manager
+and the proposal for a Template Stack Manager are in scope. The life cycle and
+security boundaries of these applications are in scope.
+
+### Out of Scope
+
+Throughout the design, Go templates are used to illustrate potential usage, but
+this is an implementation detail that will be the subject of additional design
+documents. In fact, Template Stacks should eventually benefit from having more
+than one template engine option.
+
+This design is also not heavily concerned with the format of the `Stack` object
+or the layout of the Stack filesystem. Those topics will be discussed in another
+one-pager or subsequent updates to this design. It is assumed that the reader
+has some familiarity with [`Stack`
+resources](https://github.com/crossplaneio/crossplane/blob/master/design/design-doc-stacks.md#stack-crd)
+and the way [they are automatically created from StackInstall resources](https://github.com/crossplaneio/crossplane/blob/master/design/design-doc-stacks.md#installation-flow).
+
+Stack resources that are managed as Template Stacks should require Namespaced
+scoping. Cluster scoped Template Stacks may require additional thinking.
+
+The ultimate Stack format will need to be open for use by multiple Template
+Stack Engines, as the format will certainly undergo changes from this proposal through the
+[Template Stacks UX design](https://github.com/crossplaneio/crossplane/issues/915).
+
+Additional thinking to support multiple template engines is not in scope for
+this proposal. Can multiple engines be used from a single Template Stack? Can
+these engines share resources and resource values? How does engine choice affect
+the Stack fields (if at all) and the validation of resources at build and
+install time. This proposal will not investigate those questions.
+
+Finally, the specific resources that a Template Stack may render is not in scope
+for this design. There is some thinking on this included in [Concerning the use
+of Core Types](#Concerning-the-use-of-Core-Types).
+
+## Benefits of Template Stacks
+
+The current design of Stacks allows for a new CRD to be managed by an
+independent controller. A CR will be handled by the Stack controller in any way
+that controller sees fit. This includes issuing remote API calls for additional
+inputs or creating external outputs, accessing local resources, or performing
+complex computations.
+
+Stacks may also opt to take advantage of the Crossplane Runtime to simplify the
+life cycle management of resources.
+
+In contrast, this proposal introduces Template Stacks with the limited ability
+to use the CR as input to create, update, or delete dependent Kubernetes
+resources. Once the dependent resources have been created they can also be used
+as input, their outputs can be recycled as inputs.
+
+Application owners and managers can take advantage of Template Stacks to package
+complex deployments, while users benefit from a simple installation and usage
+interface. Developers can use Template Stacks as application owners would, to
+create Kubernetes resource machinery using template engine logic, circumventing
+the need for more complex controllers.
+
+The Template Stack Manager will permit the use of template variables to create
+and manage a set of resources provided as a set of strings included in the
+definition of the Stack.
+
+The benefits of Template Stacks are as follows:
+
+* [Simple Variable Substitution](#Simple-Variable-Substitution)
+  Stack Template resources can take input from a CR spec or from sibling
+  resources.
+* [No Custom Controller Needed](#No-Custom-Controller-Needed)
+  A single all-purpose controller manages Template Stacks. The existing Stack
+  Manager will wire up the Template Stack Manager to manage the resources defined in a Template Stack.
+* [Composition](#Composition)
+  Stacks can be nested, and may reside in Crossplane managed clusters or control clusters.
+  
+  Common charts and custom YAML files used today reside on local disks.
+  With Template Stacks these templates and can reside in the cluster.
+
+  Template Stacks introduce template variables that can take advantage of the
+  field values from other template rendered resources. Template variable
+  substitution is performed in each Kubernetes reconciliation pass, affecting
+  all template resources for a given Stack CRD.
+
+### Simple Variable Substitution
+
+If an application owner wanted to parameterize a single resource within a
+more complex set of YAML resources, say the creamer type for a Coffee resource,
+we could imagine placing a `creamer` template variable in that `Coffee`.
+
+Supposing we used an inline handlebar template syntax, the `Coffee` would contain:
+
+```yaml
+apiVersion: template.stacks.example.org/v1alpha1
+kind: Coffee
+metadata:
+  name: delicious
+spec:
+  creamer: {{.coffeeCreamer}}
+```
+
+Today, Kustomize overlays can be used to provide this level of functionality, as
+can Helm chart values. However, the capabilities provided by these tools are
+limited in two ways:
+
+* Original template values are not stored in the cluster
+* Replacement values can not be taken from other resources
+
+Template Stacks will provide both of these advantages. _([Cross resource
+references](https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-cross-resource-referencing.md)
+also provide this feature for Stacks with custom controllers.)_
+
+Common template value files today live on developer or operator laptops where
+the whole file or past versions can be lost. They do not benefit from the same
+security and retention policies put in place to protect the cluster. Template
+Stacks values are stored in the CR, which make them easy to locate and examine.
+
+Even with backups, git, or online storage techniques, the template values are
+not available within the cluster, which is where processing happens. Common
+template values today can not be dynamically updated based on the conditions of
+the environment in which they operate. Custom controllers or external scripts
+are needed to fetch values from the Kubernetes cluster. Template Stacks provide
+these capabilities, within the cluster, where the resources and operations are
+regulated as they are for all Kubernetes resources.
+
+*To apply the example above to this Template Stacks proposal,
+`{{.coffeeCreamer}}` would actually be `{{.spec.coffeeCreamer}}`. A `Breakfast`
+resource would include a `Coffee` resource template. The `.spec.coffeeCreamer`
+property of a `Breakfast` resource would be applied to the template and rendered
+as new, updated, or removed Kubernetes `Coffee` resource with the appropriate
+creamer.*
+
+#### Default values
+
+Default values for Template Stacks should be supplied by the CRD author.
+
+The following example CRD shows a `Breakfast` resource definition with a
+`coffeeCreamer` `spec` property
+[defaulted](https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/20190426-crd-defaulting.md)
+to `none`.
+
+```yaml
+apiVersion: "apiextensions.k8s.io/v1beta1"
+kind: "CustomResourceDefinition"
+metadata:
+  name: "breakfast.example.crossplane.io"
+spec:
+  group: "example.crossplane.io"
+  version: "v1alpha1"
+  scope: "Namespaced"
+  names:
+    plural: "breakfasts"
+    singular: "breakfast"
+    kind: "Breakfast"
+  validation:
+    openAPIV3Schema:
+      required: ["spec"]
+      properties:
+        spec:
+          required: ["coffeeCreamer"]
+          properties:
+            coffeeCreamer:
+              type: "string"
+              default: "none"
+```
+
+This specification does not propose any additional means to supply default field
+values. The omission of an alternative mechanism for setting default values
+should encourage resource authors to take advantage of the latest [Kubernetes
+CRD features](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#defaulting).
+
+The Template Stack Manager should be capable of replacing controllers in
+a range of use-cases. To that end CRDs should not need to be tailored for use by
+the TSM. The pros and cons of this viewpoint are left for future discussion.
+
+Offering a means to provide default values would fit Template Stacks to more CRD
+adoption scenarios, especially in cases where Template Stacks were not
+considered at the inception of the CRD.
+
+### No Custom Controller Needed
+
+This is where the value of Template Stacks starts to materialize. A stack author
+doesn't need to write their own controller, because a common controller is
+provided with the Template Stacks Manager.
+
+Existing Stacks must specify a `Deployment`. These are typically written in Go
+and generally require understanding some combination of client-go,
+controller-runtime, crossplane-runtime, and Kubebuilder.
+
+For some applications, the intended work of the controller could be simple:
+
+* For every new resource, create a set of Kubernetes resources
+* Use the spec fields of that resource to affect the Kubernetes resources
+* Upon changes to these spec fields, update the Kubernetes resources
+
+These are responsibilities the Template Stack Manager will handle.
+
+### Composition
+
+#### Vertical Composition
+
+Crossplane Stacks manage multiple CRDs. Template Stacks should be no different.
+
+Template Stack `template` definitions will need to express in text the resources
+to render for each CRD that is managed. The GVK can be used as a map key to
+denote the CRD template being described.
+
+```yaml
+kind: Stack
+metadata:
+  name: RedisStack
+spec:
+  customresourcedefinitions:
+  - kind: Redis
+    apiVersion: redis.example.org/v1
+  templates:
+    redis.example.org/v1:
+      deployment: |
+          …
+          containers:
+          - name: redis-controller
+            image: example/redis-controller:{{.spec.redisVersion}}
+```
+
+*In this example, templates are found within a map keyed by the GVK and a
+template grouping name. The reasoning for this map format will be explained in
+a future design.*
+
+With a more complex example, we see that each template needed to render a GVK
+should be provided with a name that represents the intent of that template.
+This distinguishing name can be reused to reference the template and its single
+resource (see [Horizontal Composition](#Horizontal-Composition)).
+
+##### Nesting Template Stacks
+
+In the example below, distinct templates are defined for the Redis CRD and the
+CachingWebService. CachingWebService includes two templates, one of which relies
+on the other.
+
+```yaml
+apiVersion: stacks.crossplane.io/v1alpha1
+kind: Stack
+metadata:
+  name: CachingWebServiceStack
+spec:
+  customresourcedefinitions:
+  - kind: Redis
+    apiVersion: redis.example.org/v1
+  - kind: CachingWebService
+    apiVersion: cachingwebservice.example.org/v1
+  templates:
+    redis.example.org/v1:
+      deployment: |
+        kind: Deployment
+        …
+          containers:
+          - name: redis-controller
+            image: example/redis-controller:{{.spec.redisVersion}}
+    cachingwebservice.example.org/v1:
+      cache: |
+        apiVersion: example.org/v1
+        kind: Redis
+        …
+        spec:
+          redisVersion: "5"
+      web: |
+        kind: Deployment
+        …
+          containers:
+          - name: nginx-controller
+            image: example/nginx-controller:{{.spec.nginxVersion}}
+--
+apiVersion: example.org/v1
+kind: CachingWebService
+metadata:
+  name: cacheme
+spec:
+  nginxVersion: "1.17.4"
+```
+
+*When the `CachingWebService` instance is reconciled by the TSM, Kubernetes
+resources represented by the `cache` and `web` templates will be produced. These
+templates are effectively concatenated by the TSM, as if to form a single
+`cachingwebservice.yaml`. The TSM then applies the template variables and
+installs these resources as though `kubectl apply -f cachingwebservice.yaml` was
+invoked.*
+
+The `nginxVersion` is defined in the `spec` of the `CachingWebService` instance,
+while `redisVersion` has been declared in the `cache` template for
+`cachingwebservice.example.org/v1`. In each case, a template variable has been
+passed into a dependent template. One was user supplied, the other was template
+supplied.
+
+An additional form of composition and reuse is available by referencing Template
+Stack kinds within new Template Stacks. For example, `MegaWebsite` could include
+`CachingWebServer` in the `dependsOn` field of that Stack.
+
+#### Using Reconciled Values as Template Variables
+
+Kubernetes resources typically set their state within a `status` field, which is
+often available as a Status Sub-Resource. In fact, CRD resources specifically
+support scale and status sub-resources. Access to this sub-resource can be
+granted independently from the resource itself.
+
+The template capabilities available to a Template Stack should not be restricted
+for use within the `spec` or other resource creation features.
+
+Template Stack managed resources should be able to manage their status, as any
+controller managed resource would. Arguably, the most useful values to expose
+through this status would be derived from resources created with templates.
+
+While maintaining that the `spec.templates` field of a `Stack` can present the
+entire template body of CRD resources, let `templateStatus` present a template
+for a resource's `status` body. As these `templateStatus` templates will define
+the `status` of a single CR, we will only need a GVK key to identify which
+`templateStatus` is being described.
+
+```yaml
+…
+kind: Stack
+metadata:
+  name: Foo
+spec:
+  customresourcedefinitions:
+  - kind: Foo
+    apiVersion: foo.example.org/v1
+  template:
+    foo.example.org/v1:
+      templateA: |
+        kind: DependentThing
+        spec:
+        …
+  templateStatus:
+    foo.example.org/v1: |
+        statusField: {{.some.value}}
+```
+
+*If the CRD for `example.example.org/v1` includes the `status` sub-resource, the
+Template Stack Manager can potentially make `status` sub-resource setting
+optimizations. Otherwise, the TSM can set the status by updating the while
+`Foo` resource.*
+
+Crossplane Stacks, like all Kubernetes resources, benefit from an active state
+reconciliation loop. Status values may not be immediately available. The
+template syntax should allow for templates to overcome these situations, for
+example (with Go templates):
+
+```yaml
+  templateStatus:
+    example.example.org/v1: |
+      {{if pipeline}}
+      statusField: {{.some.value}}
+      {{end}}
+```
+
+##### Restrictions on Use
+
+In order for templates to be resolved when their corresponding Kubernetes
+resources are not available, some restrictions must be made on where templates
+can be used. At install or build time, each resource template should be
+inspected to verify that apiVersion, kind, and name exist within the template
+and can be parsed from the template without the need for dependent or mutable variables.
+
+A single unique name declared within a template creates a limitation of its own.
+With that limitation, a Template Stack could only have one resource defined per
+namespace. Fortunately, the UID of the resource is immutable and would be
+available for use within templates.
+
+##### Labels and Annotations
+
+Labels and Annotations provide users with countless extensibility and
+categorization options. Template Stacks should permit these field values to be
+retained between resource updates. Likewise, the Template Stack author should
+have the ability to reset the labels and annotations. Through the use of list
+merging functions, any Template Stack engine should give the Template Stack
+author the means to define the desired behavior for their Stack.
+
+```yaml
+  templates:
+    example.example.org/v1:
+      fooresource: |
+      ...
+      metadata:
+        annotations:
+          {{- range $key, $value := (mergeFn .fooresource.metadata.annotations ["foo"]) }}
+          - {{$key}}: {{$value}}
+          {{- end }}
+```
+
+#### Horizontal Composition
+
+We've now demonstrated several properties of Template Stacks:
+
+* a means to concatenate Kubernetes Resource YAML
+* nesting supported CRD types or the CRDs of other Stacks
+* a means to set the status of a resource using source or template generated
+  properties
+
+Another property that is possible in a home-grown controller, is the ability to
+use the status of any one resource that controller manages to affect other
+resources. Template Stacks should offer this same capability.
+
+```yaml
+kind: Stack
+metadata:
+  name: Foo
+spec:
+  templateStatus:
+    foo.example.org/v1: |
+        statusField: {{.some.value}}
+  template:
+    foo.example.org/v1:
+      controller: |
+        kind: Thing
+        spec:
+        …
+      other: |
+        kind: DependentThing
+        spec:
+          someInput: {{.controller.status.statusField}}
+```
+
+*The `DependentThing` included in the `other` template relies on the resolved
+`statusField` from the `Thing` included in the `controller` template.*
+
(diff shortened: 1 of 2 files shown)


design/one-pager-template-stacks.md
```diff
@@ -0,0 +1,344 @@
...
+* [Template Stack Manager](#Template-Stack-Manager)
+  * [Deployment](#Deployment)
+  * [Owner References](#Owner-References)
+  * [RBAC](#RBAC)
+  * [Life Cycle and Updates](#Life-Cycle-and-Updates)
+  * [Unknown Types](#Unknown-Types)
+
+## Background
+
+Early thinking on Stacks included the idea of simplifying the overhead involved in life-cycle management for complex workloads, especially as this relates to consuming Crossplane managed services. However, to contribute a new Stack in the Stack eco-system today, it is necessary to create a new Controller, typically in Go, typically using Kubebuilder.
```
soorena776: nit: "typically in Go, typically using Kubebuilder." -> "typically in Go and using Kubebuilder."

design/one-pager-template-stacks.md
```diff
@@ -0,0 +1,344 @@
...
+        args: ["--stack", "stackname"]
+```
+
+A proposed `stack` argument to the TSM would allow the controller to find a matching Stack. The TSM and Stack should reside within the same namespace. The Stack Manager will abide.
+
+### Owner References
+
+The [Controller Reference](https://godoc.org/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil#SetControllerReference) will be set on the TSM `Deployment` so the TSM is removed when the Stack is removed. The TSM can identify the Stack to work on through this controller reference, but this approach may make it difficult for admins to identify which deployments are operating on behalf of Stacks (and their CRDs).
+
+Stacks could also take on a reference to the `Deployment` and `ServiceAccount` assuring that the correct resource will be updated when the Stack is updated. What of the CRDs? Revised CRDs in an updated Stack will be updated because they are unique, but how will Stack CRDs that have been removed be deleted? This is a trickier problem to be considered in future designs.
```
turkenh: > Stacks could also take on a reference to the `Deployment` and `ServiceAccount` assuring that the correct resource will be updated when the Stack is updated.

Deployment only has the name of the stack, so updating would mean just restarting the pod  as stack name is not expected to change?

design/one-pager-template-stacks.md
```diff
@@ -0,0 +1,344 @@
...
+
+By creating a TSM deployment for each Template Stack that is installed, we are able to scope the permissions given to each TSM instance to the permissions requested by the Stack (i.e. the CRDs it owns and depends on). This is analogous to how the Stack Manager currently creates a `Deployment` to host each controller-runtime based Stack’s controller which is also scoped down to the needed permissions.
+
+#### Concerning the use of Core Types
+
+In several examples throughout this design, Template Stacks are demonstrated generating `Deployment` resources by including them in template text. There are compelling justifications for both permitting and denying Template Stacks from declaring a dependency on core and extension types, such as `deployment.apps/v1`.
```
turkenh: Before reading this line, I was just trying to figure out this by checking examples again and again. 
Good to see that you put this explanation here otherwise I would start thinking that I misunderstand some basics 😊

design/one-pager-template-stacks.md
```diff
@@ -0,0 +1,344 @@
...
+
+The ability to use Template Stacks to render arbitrary Kubernetes resources is a powerful feature. Such a privilege would change the boundaries of a Stack, permitting it to function outside of Crossplane managed types. Template Stacks would compete with `KubernetesApplication` on this capability.
+
+If Template Stacks are prohibited from using core and extension types, the complement of `KubernetesApplication` and a local `KubernetesCluster` can be used to overcome the limitation while providing additional security boundaries. These are features and challenges to investigate in a future design.
+
+### Life Cycle and Updates
```
turkenh: Below questions are not specific to template stacks and valid for current Stacks as well, right?

design/one-pager-template-stacks.md
```diff
@@ -0,0 +1,344 @@
...
+Hand-crafted controllers have the advantage of knowing the types and structure of the resources they manage. In Go, this allows for the advantages of a strongly typed environment including compile time error checking and code completion during development.
+
+A Template Stack would not include a controller. An external controller will have to handle reconciliation for types that the controller could not have been aware of at compile time.
+
+How can an external controller manage arbitrary types declared at runtime?
+
+```go
+var obj unstructured.Unstructured
+obj.SetAPIVersion("foo/v1")
+obj.SetKind("Bar")
```
turkenh: `foo/v1` and `Bar` has no effect but need not to be empty, right?

design/one-pager-template-stacks.md
```diff
@@ -0,0 +1,344 @@
...
+  name: CachingWebServiceStack
+spec:
+  templates:
+    redis.example.org/v1:
+      deployment: |
+        kind: Deployment
+        …
+          containers:
+          - name: redis-controller
+            image: example/redis-controller:{{.spec.redisVersion}}
```
turkenh: circular dependencies need to handled/disallowed.

design/one-pager-template-stacks.md
```diff
@@ -0,0 +1,344 @@
...
+    cachingwebservice.example.org/v1:
+      cache: |
+        apiVersion: example.org/v1
+        kind: Redis
+        …
+        spec:
+          redisVersion: "5"
```
turkenh: What happens if one changes redisVersion here after a stack instance is created? TSM will reconcile, how? Or otherside, what happens created instance edited?

design/one-pager-template-stacks.md
```diff
@@ -0,0 +1,399 @@
...
+
+Existing Stacks must specify a `Deployment`. These are typically written in Go and generally require understanding some combination of client-go, controller-runtime, crossplane-runtime, and Kubebuilder.
+
+For some applications, the intended work of the controller could be simple:
+
+* For every new resource, create a set of Kubernetes resources
+* Use the spec fields of that resource to affect the Kubernetes resources
+* Upon changes to these spec fields, update the Kubernetes resources
+
+This is where the value of Template Stacks starts to materialize.
```
suskin: I am thinking it would probably make sense to repeat the benefit at this point. Maybe something like:

```
This is where the value of Template Stacks starts to materialize. A stack author doesn't need to write their own controller, because a common one is provided by the Template Stacks Manager.
```

It might make sense to put the restatement of the benefit at the end of the first paragraph though.

I am thinking it makes sense to restate it so that the section is a bit more self-contained. It also helps me as a reader understand what I am looking at when I get to the example code snippet.

design/one-pager-template-stacks.md
```diff
@@ -0,0 +1,399 @@
...
+
+### Composition
+
+#### Vertical Composition
+
+Crossplane Stacks manage multiple CRDs. Template Stacks should be no different.
+
+Template definitions will need to express the template to render for each CRD that is managed. The GVR can be used as a map key to denote the CRD template being described.
+
+Each template needed to render a GVR should be provided (demonstrated below as a map), additionally a name that represents that intent of that template should distinguish each template (also demonstrated as a map). This name can then be reused to reference a template (see [Horizontal Composition](#Horizontal-Composition)).
```
suskin: Where did the templates come from?

design/one-pager-template-stacks.md
```diff
@@ -0,0 +1,399 @@
...
+            image: example/nginx-controller:{{.spec.nginxVersion}}
+--
+kind: CachingWebService
+spec:
+  nginxVersion: "1.17.4"
+```
+
+In the example above, distinct templates are defined for the Redis CRD and the CachingWebService. CachingWebService includes two templates, one of which relies on the other.
+
+The `nginxVersion` is defined in the `spec` of the `CachingWebService` instance, while `redisVersion` has been declared in the `cache` template for `cachingwebservice.example.org/v1`. The template variable has been passed into a descendent Stack.
```
suskin: What is a `descendent Stack`?

design/one-pager-template-stacks.md
```diff
@@ -0,0 +1,399 @@
...
+Supposing we used an inline handlebar template syntax, the `deployment` field could contain:
+
+```yaml
+containers:
+- name: redis-controller
+  image: redis/redis-crossplane-controller:{{.spec.version}}
+  imagePullPolicy: Always
+```
+
+_In practice, `version` would most likely be static, pinned to a specific version of a Stack_.
```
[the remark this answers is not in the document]
displague: There is no difference, I should pick one format though.

design/one-pager-template-stacks.md
```diff
@@ -0,0 +1,344 @@
...
+
+In this example, the `.spec.version` value would be taken from the body of each CR this Stack manages. That value would be applied to the template and rendered as a new or updated `Deployment` resource.
```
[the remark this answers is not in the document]
displague: This section was very confusing.  I've simplified the Benefits sections to not  introduce `Stack` deployment strategies.  I moved most of this text to the Template Stack Manager introduction (where much of it was already being repeated).

Thanks for pointing this out!

design/one-pager-template-stacks.md
```diff
@@ -0,0 +1,399 @@
...
+
+Should the `controller.deployment` value be treated as a template? This could have unintended consequences for existing Stack deployments, especially if their definitions includes pieces that match the template format.
```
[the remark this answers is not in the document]
displague: https://github.com/crossplaneio/crossplane/pull/928/files#r334236256

design/one-pager-template-stacks.md
displague: I should point out in the design that the controller doesn't need to be concerned with this (although the template developer does).

The rendering process of the Template Stack Manager will use a single (cached) pass at applying templates.

* All of the current resources are fetched
* The resources are combined to form a variables to pass to the template
* Template substitutions are made for each resource
* All of the templated resources are applied to the API
  Including a pass at deleting resources that whose templates have become empty

I do not believe the template rendering process can get caught in a circular dependency loop here. However, the Template Stack creator could improperly craft their templates in a way that they would never resolve.  This could be left as a `crossplane-cli` concern.

Similarly, if a template contained invalid resource yaml, or contained fields and values that were not valid for the resource being generated, the Template Stack Manager will need to a defined means of surfacing that error condition (perhaps as a template variable or function, but also through logging).

I intended to include some of this thinking in the design - thanks for calling out the omissions!

displague: This will be found in the `Life Cycle and Updates` section which was woefully bare.

design/one-pager-template-stacks.md
```diff
@@ -0,0 +1,344 @@
...
+        …
+        spec:
+          redisVersion: "5"
+      web: |
+        kind: Deployment
+        …
+          containers:
+          - name: nginx-controller
+            image: example/nginx-controller:{{.spec.nginxVersion}}
+--
```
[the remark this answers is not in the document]
displague: I've been trying to be consistent about the use of `…`.

If I can avoid writing more stock YAML between a first level and a nest value, I'll put the ellipsis in between.

Since the order of resource values is not important, an ellipsis on the top or the bottom of a resource carries equal value, but I will tend to start with the ellipsis so it is more obvious that I am skipping over some values.

This may have been more of a convenience for me than the reader.  I was trying to draw focus to the important pieces without the distraction of other elements.

Would it be better if I included valid YAML blocks in all examples?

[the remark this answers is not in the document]
displague: I think I see what you mean here.. I didn't include the '...' on `CachingWebService`.. Github inline messages sometime leave me wonder what line the comment was concerned with.  I'll add the '...'.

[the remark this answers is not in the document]
displague: I went with `apiVersion` instead here.. Good call.

design/one-pager-template-stacks.md
displague: Beats me..

I'm replacing that last sentence with:

> In each case, a template variable has been passed into a dependent template. One was user supplied, the other was template supplied.

Thanks for pointing this out.

Related to this, I have started using "dependent resource" to describe resources the templated sub-resources of a Template Stack.  Look for this in further updates and let me know how it fits?

design/one-pager-template-stacks.md
```diff
@@ -0,0 +1,344 @@
...
+spec:
+  template:
+    example.example.org/v1:
+      controller: |
+        kind: Deployment
+        spec:
+        …
+  templateStatus:
+    example.example.org/v1:
+      controller: |
```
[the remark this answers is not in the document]
displague: `controller` is an arbitrary map key in this case.  It is confusing. I'll use a word with less baggage.

design/one-pager-template-stacks.md
displague: Covered in Life Cycle now..

design/one-pager-template-stacks.md
displague: Correct, the type will be taken at runtime by examining the `Stack` resource that this TSM instance is operating against.

design/one-pager-template-stacks.md
displague: I've added more text to this section and tried to split the thinking up into `Resource life cycle` and `Stack life cycle`.  I believe I may have omitted some useful details from the `Stack life cycle` section that may or may not have been mentioned elsewhere in the design.

Let me know if you find more holes.

design/one-pager-template-stacks.md
```diff
@@ -0,0 +1,399 @@
...
+* **TS (Template Stack)**
+  : UNDER THIS PROPOSAL, a Stack that declares dependent resources and includes managed CRDs, but includes templates to be rendered for each CR rather than a `Deployment` controller
+* **TSM (Template Stack Manager)**
+  : UNDER THIS PROPOSAL, a restricted `Deployment` that manages resources for a Template Stack's CRDs
+
+## Introduction
+
+### In Scope
+
+This design introduces the concept and benefits of Template Stacks as well as exploring a proposed Template Stack Manager which may facilitate this experience. Implementation challenges for a the existing Stack Manager and the proposed Template Stack Manager (and approaches to overcome them) are provided.
```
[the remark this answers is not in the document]
displague: Will do.. The following VS Code tips and extensions helped:
https://jmarcher.io/vs-code-markdown-and-word-wrap/
https://marketplace.visualstudio.com/items?itemName=stkb.rewrap

design/one-pager-template-stacks.md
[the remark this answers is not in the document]
turkenh: I see you changed `controller` as `templateA`. Thanks for that, however my point was a bit different. 

When I check the value for this arbitrary key in `template.example.example.org/v1.templateA`, I see a k8s `Deployment` as value and feel like its status would be updated due to `templateStatus.example.example.org/v1.templateA`

What I mean is, putting statusField under a key seems not necessary and confusing since we are updating status of, `example.example.org/v1` and not `Deployment`.

For example, if we are two templates as below, would it make any difference to write `templateA` or `templateB` in `templateStatus` ? 
```yaml
kind: Stack
metadata:
  name: StatusExampleStack
spec:
  template:
    example.example.org/v1:
      templateA: |
        kind: Deployment
        spec:
        …
      templateB: |
        kind: Service
        spec:
        …
  templateStatus:
    example.example.org/v1:
      templateA: |
        statusField: {{.some.value}}
```

...or, do you mean, `{{.some.value}}` coming from status of object defined as `templateA` ?

design/one-pager-template-stacks.md
displague: updated in eff1e90a

design/one-pager-template-stacks.md
```diff
@@ -0,0 +1,679 @@
...
+
+The rendered template will replace changes to dependent resources that are found
+in the Kubernetes API. More importantly, since the values are obtained directly
+from the API on each pass, if a resource is updated by a user or some other
+process, those values will be used when rendering.
+
+If a templates rendering contains invalid resource YAML, or contains fields and
+values are not valid for the resource being generated, the Template Stack
+Manager will need a defined means of surfacing these error conditions. Just as
+sibling resources should be made available through templates, the error messages
```
jbw976: I'm not sure I'm following what this means to make error messages available through templates.  Is an example helpful here, or maybe another sentence or two describing this functionality?

Is it in scope for this intial effort?

design/one-pager-template-stacks.md
```diff
@@ -0,0 +1,679 @@
...
+`ServiceAccount` and `Deployment` are deleted.
+
+What should happen if the Stack `Deployment` is manually deleted or modified?
+Should the Stack Manager recreate or modify it?
+
+What happens when the `dependsOn` or `customresourcedefinitions` field of a
+`Stack` is updated? The new CRDs should be added, and removed CRDS should be
+deleted.  When and how should the RBAC roles be updated?
+
+This are questions left for future designs affecting all types of Stacks, not
```
jbw976: Link to #879 here?

design/one-pager-template-stacks.md
```diff
@@ -0,0 +1,679 @@
...
+
+*The `DependentThing` included in the `other` template relies on the resolved
+`statusField` from the `Thing` included in the `controller` template.*
+
+We will not focus on how the names of adjacent templates are exposed in a
+templating syntax, just that they should be made available.
+
+Depending on reconciled fields is bound to inject latency into a full
+resolution. Instead of relying on a static set of inputs, our inputs can now be
+dynamic. The inputs are driven by an independent, yet necessarily sibling,
```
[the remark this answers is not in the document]
displague: You should be to reference any "sibling" template, and also the current template itself from within a template. The ability to feed a variable from status back into spec is technically possible and _free_, but I can't think of a good use-case immediately, nor any reason to restrict this.

You could also use a different Stack in your template, but you will have to treat it like any other Kubernetes resource.  The values of that Stack would have to be accessed through the template which includes it.

```
StackA:
  crdOne:
    templateA: |
      kind: Apple
    templateB: |
      kind: Banana
    templateC: |
      kind: Cherry
    templateStatus: |
      apples: {{ .templateA.status.appleStatus }}
      bananas: {{ .templateB.status.bananaStatus }}
      cherries: {{ .templateC.status.cherryStatus }}    
```
```
StackB:
  crdTwo:
    templateD: |
       kind: StackA 
       spec:
          # Use the current template's fields
         fromSelf: {{ .templateD.someVar}}

         # or those of sibling templates.
         fromSiblings:
           - {{ .templateE.someVar}}
           - {{ .templateF.someVar}}

         # It is not possible to reference templateA-templateC here.
         # but you can access them as templateD so long as StackA
         # exposed their properties in some way.
         # In this example, StackA, used templateStatus to expose
         # select members of templateA-templateC.
         fromStackA:
         - {{ .templateD.status.apples }}
         - {{ .templateD.status.bananas }}
         - {{ .templateD.status.cherries }}
    templateE: |
    templateF: |
  crdThree: ...
```

I don't think an initial version of template stacks should allow you to fetch arbitrary values from external resources, resources that were not generated by the Stack's own templates.

design/one-pager-template-stacks.md
```diff
@@ -0,0 +1,679 @@
...
+        kind: Deployment
+        spec:
+        …
+  templateStatus:
+    example.example.org/v1:
+      templateA: |
+        statusField: {{.some.value}}
+```
+
+*The `status` could be included in the previously described template body, but
```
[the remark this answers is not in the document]
displague: Your senses are correct about something being off here.

The `templateStatus` will not need to include the various template names.  There is a single `templateStatus` per CRD.

```
  templateStatus:
    example.example.org/v1:
      templateA: |
        statusField: {{.some.value}}
```

should be:

```
  templateStatus:
    example.example.org/v1: |
        statusField: {{.some.value}}
```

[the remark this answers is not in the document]
displague: > So if the author defines their CRD to use or not use a separate status subresource, we have to honor that. The TSM can't make it's own independent decision.

That's a good point.  The TSM can optimize to use `UpdateStatus` if a Status sub-resource is included in the CRD (which can be detected by the TSM), otherwise it will just update status with the rest of the resource.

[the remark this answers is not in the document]
displague: This less abstract example may help:

```yaml
apiVersion:  stacks.crossplane.io/v1alpha1
kind: Stack
metadata:
  name: Foo
spec:
   templates:
     foo.group/version:
       templateA:
         kind: athing
         spec:
           foovar: {{ .spec.foo }}
   templateStatus:
     foo.group/version:
       statusthing: {{ .templateA.status.bar }}
```

Given:

```yaml     
apiVersion:  stacks.crossplane.io/v1alpha1
kind: Foo
spec:
  foo: "foo"
```

The TSM will generate the following:

```yaml
kind: athing
spec:
  foovar: "foo"
```

As `athing` type is reconciled by an outside controller, a `status` emerges:

```yaml
kind: athing
spec:
  foovar: "foo"
status:
  bar: "bar"
```

On the next TSM pass, the Foo resource will be updated:

```yaml     
apiVersion:  stacks.crossplane.io/v1alpha1
kind: Foo
spec:
  foo: "foo"
status:
  statusthing: "bar"
```

design/one-pager-template-stacks.md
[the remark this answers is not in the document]
displague: @suskin points out that a `HelloWorld` and `PlusOne` Stack example, illustrating the simplest Template Stack (no `templates`, only `templateStatus`) would be helpful.

design/one-pager-template-stacks.md
```diff
@@ -0,0 +1,679 @@
...
+
+The benefits of Template Stacks are as follows:
+
+* [Simple Variable Substitution](#Simple-Variable-Substitution)
+
+  Stack Template resources can take input from a CR spec or from sibling
+  resources.
+* [No Custom Controller Needed](#No-Custom-Controller-Needed)
+
+  There is a single all-purpose controller that is designed to managed Template
```
soorena776: `is designed to manage Template Stacks`

design/one-pager-template-stacks.md
```diff
@@ -0,0 +1,679 @@
...
+  image: redis/redis-crossplane-controller:{{.someVariable}}
+  imagePullPolicy: Always
+```
+
+Today, Kustomize overlays can be used to provide this level of functionality, as
+can Helm chart values. However, the capabilities provided by these tools are
+limited in two ways:
+
+* Original template values are not stored in the cluster
+* Replacement values can not be taken from other resources
```
soorena776: https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-cross-resource-referencing.md 
is doing exactly this: taking values from other managed resources.

design/one-pager-template-stacks.md
```diff
@@ -0,0 +1,679 @@
...
+- name: redis-controller
+  image: redis/redis-crossplane-controller:{{.someVariable}}
+  imagePullPolicy: Always
+```
+
+Today, Kustomize overlays can be used to provide this level of functionality, as
+can Helm chart values. However, the capabilities provided by these tools are
+limited in two ways:
+
+* Original template values are not stored in the cluster
```
soorena776: Is there a specific value in storing all the value options in the cluster as an object?

design/one-pager-template-stacks.md
```diff
@@ -0,0 +1,679 @@
...
+        args: ["--stack", "stackname"]
+```
+
+A `--stack` argument to the TSM would allow the controller to find a matching
+`Stack` resource. The TSM and Stack should reside within the same namespace. The
+Stack Manager will abide.
+
+### Owner References
+
+The [Controller
```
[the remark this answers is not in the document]
displague: A ControllerReference is an OwnerReference with the "controller" flag set.

The one permitted controlling owner knows that it is responsible for a given resource.  If that controller has gone away, and other owners are still present, then the orphaned resource could potentially be adopted by a new controller (owner).

https://github.com/kubernetes/community/blob/master/contributors/design-proposals/api-machinery/controller-ref.md#the-three-laws-of-controllers

My thinking with having `Wordpress` be the controllerreference was that it was about grabbing the authority of the singular controller owner. It's not clear to me (from k8s docs) if this makes sense.  Should an object (that is not a 'process' (deployment/pod)) be the controller for another object.

---

I'll revisit the owner reference examples you cited.  

I'll also ping @negz for any insight he may have on the topic. (ping)

design/one-pager-template-stacks.md
```diff
@@ -0,0 +1,679 @@
...
+
+#### Concerning the use of Core Types
+
+In several examples throughout this design, Template Stacks are demonstrated
+generating `Deployment` resources by including them in template text. There are
+compelling justifications for both permitting and denying Template Stacks from
+declaring a dependency on core and extension types, such as
+`deployment.apps/v1`.
+
+The ability to use Template Stacks to render arbitrary Kubernetes resources is a
```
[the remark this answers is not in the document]
displague: I won't resolve this thread until there is an issue to track the use of core types.

I think we are going to have a much easier time creating template stacks if we don't need to create a KubernetesApplication for each independent resource.

Take 
https://gitlab.com/gitlab-org/charts/gitlab/blob/master/charts/gitlab/charts/operator/templates/operator_deployment.yaml
and the neighboring templates for example.  We could make a set of Template Stacks, each representing one of the Gitlab CRDs, or a single Template Stack representing all of the CRDs (depending on how much control (roles) you want to give the TSM(s)).

Either way, there will be lots of core resources needed.  Does each Gitlab Template Stack put all of these core resources in a single KubernetesApplication resource within the Stack?

Does each Template Stack resource (or template) then require a KubernetesCluster reference?

I think we'll need a real world example to determine if Crossplane managed types *CAN* be sufficient to translate existing chart projects to Template Stacks.

I'm reluctant to assert that Template Stacks can only handle Crossplane managed types until we can prove that this offers sufficient capability in practice.

I think this section of the design already exposes both ideologies, deferring to future design work.  To your point, I could clarify by concluding the section stating core types are not expressly blocked by this design and the current Stacks implementation, but this may not always be the case.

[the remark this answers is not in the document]
displague: Looking at those Gitlab charts has made me wonder if we may also need a third field:
* `templates` K8s resource template text keyed by both GVK, then a unique template name
* `templateStatus` sub-resource template text keyed by GVK
* `includes` (not really this name...) arbitrary templates that are not necessarily k8s resources or sub-resources, which could be reused by `templates` or `templateStatus`.
   I don't know if these should be shared between CRDs (how could they if they include CRD specific template variables).
   These could perhaps live under `templates`, just using arbitrary paths / names instead of CRD names.

design/one-pager-template-stacks.md
suskin: In general I am not a huge fan of the term `dependent` because it is so similar to `dependency`; I like to use some version of `consume` or `consumer` when possible so that the terms are More Different.

I'll keep an eye out for the terminology changes though!

design/one-pager-template-stacks.md
```diff
@@ -0,0 +1,679 @@
...
+* Get the current resource
+* Get the `templates` from the `Stack` resource that owns this CRD
+* All of the dependent resources (generated from templates) are fetched.
+* The current resource and dependent resources are combined to form variables to
+  pass to the template.
+* Template substitutions are made for each resource.
+* All of the dependent, templated, resources are applied to the API, including a
+  pass at deleting resources whose templates result is empty.
+
+Because this process operates in a single read then write pass, it can not get
```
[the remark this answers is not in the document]
displague: The TSM as the controllerReference/ownerRef will update these resources.  The TSM will [watch](https://kubernetes.io/docs/reference/using-api/api-concepts/#efficient-detection-of-changes  ) the resources belonging to its set of CRDs.  

I was not expecting to "detect" changes on the dependent resources, in this way. 
 Rather, I expect a [`Result{RequeueAfter}`](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/reconcile/reconcile.go#L30-L32) would be sufficient for an initial implementation. There should be advantages to [watching each dependent resource](https://book-v1.book.kubebuilder.io/beyond_basics/controller_watches.html#watching-arbitrary-resources), creating a more event driven "loop" (requeueing), but a timer based approach should be sufficient for now.

design/one-pager-template-stacks.md
displague: The set of variables (and/or functions) to present to a template should include:
* the Stack managed object (the `Foo`)
* the set of dependent resource objects (the `DependentThing`)
* the set of errors encountered rendering each dependent resource
  This will allow `Foo` to reflect errors rendering `DependentThing` as a conditioned status using `templateStatus`.
* helper functions for string manipulation, list handling, and simple numeric operations, such as https://github.com/leekchan/gtf#index or https://github.com/hashicorp/consul-template/

Additionally, we may want to expose
* the errors encountered rendering the templateStatus
* the raw templates themselves.
   This may be useful for storing partial templates in the Stack.  The key name would not match a GVK in this case, which means it could be an arbitrary identifier.  The key could resemble a filepath, which could help in the migration of Helm chart "includes" if an "include" function is provided in the template engine.
   ```yaml
   kind: Stack
   spec: { name: "Foo" }
   templates:
     this/name/can/be/anything:
       so/can/this/name: |
         Reusable partial template {{.someVar}}.
         The keys do not have to match GVKs owned by this Stack,
         permitting this use case.
     foo.group/version:
       resourceTemplate:
          kind: DependentThing
          spec:
             text: {{ include "this/name/can/be/anything" "so/can/this/name" }}
   ```

design/one-pager-template-stacks.md
```diff
@@ -0,0 +1,679 @@
...
+controller. A CR will be handled by the Stack controller in any way that
+controller sees fit. This includes issuing remote API calls for additional
+inputs or creating external outputs, accessing local resources, or performing
+complex computations.
+
+Stacks may also opt to take advantage of the Crossplane Runtime to simplify the
+life cycle management of resources.
+
+In contrast, this proposal introduces Template Stacks with the limited ability
+to use their CR and child resources as input to create, update, or delete child
```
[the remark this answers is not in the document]
displague: Replacing "child" with "dependent".

Rewording this to:

> In contrast, this proposal introduces Template Stacks with the limited ability
to use the CR as input to create, update, or delete dependent Kubernetes
resources. Once the dependent resources have been created they can also be used
as input, their outputs can be recycled as inputs.
>
> Application owners and managers can take advantage of Template Stacks to package
complex deployments, while users benefit from a simple install and usage
interface. Developers can use Template Stacks as application owners would, to
create Kubernetes resource machinery using template engine logic, circumventing
the need more complex controllers.

design/one-pager-template-stacks.md
```diff
@@ -0,0 +1,679 @@
...
+  Stack Template resources can take input from a CR spec or from sibling
+  resources.
+* [No Custom Controller Needed](#No-Custom-Controller-Needed)
+
+  There is a single all-purpose controller that is designed to managed Template
+  Stacks. The existing Stack Manager will wire the Template Stack Manager for
+  managing the resources defined in a Template Stack.
+* [Composition](#Composition)
+
+  Stacks can be embedded in different Cluster environments, including other
```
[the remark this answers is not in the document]
displague: updated to:
>   Stacks can be nested, and may reside in Crossplane managed clusters or control clusters.
>
>  Common charts and custom YAML files used today reside on local disks.
  With Template Stacks these templates and can reside in the cluster.

design/one-pager-template-stacks.md
```diff
@@ -0,0 +1,679 @@
...
+  a Template Stack.
+
+  Template Stacks introduce template variables that can take advantage of the
+  field values from other template rendered resources. Template variable
+  substitution is performed in each Kubernetes reconciliation pass, affected
+  all template resources for a given Stack CRD.
+
+### Simple Variable Substitution
+
+If an application owner wanted to parameterize a single `Deployment` within a
```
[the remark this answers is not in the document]
displague: I was trying to start with a resource that may be more generally understood.  I could replace `Deployment` with any arbitrary kind, or argue that the missing `apiVersion` on `Deployment` means that this could be a `deployment.crossplane.example.org/v1alpha1` :smile_cat: 

I'll replace it with a silly example type that has less pretext. 

```
kind: Coffee
spec:
  creamer: {{ .creamer }}
```
[additives?](https://arstechnica.com/civis/viewtopic.php?t=82197)

design/one-pager-template-stacks.md
displague: I'll mention it here too.. Cross-resource referencing only works if the resource subscribes to the Crossplane development ideology, and only for predefined well known object types.

In Template Stacks, you can reuse arbitrary values from Crossplane and non-Crossplane resources.

I worry about expanding the scope of the benefits section to differentiate between all possible competitors, but I'll work this in here for now.

design/one-pager-template-stacks.md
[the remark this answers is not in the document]
displague: Perhaps this would be better?

```yaml
  templates:
     status:
     kindA.group/version:
       nameA: |
       nameB: |
     kindB.group/version:
       nameA: |
       nameB: |
````

File: design/one-pager-template-stacks.md (hunk 26 above)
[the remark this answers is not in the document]
Comment (displague) [reply design/one-pager-template-stacks.md]: This text has been included in the design.

File: design/one-pager-template-stacks.md (hunk 17 above)
Comment (displague) [reply design/one-pager-template-stacks.md]: This text has been included in the design.

File: design/one-pager-template-stacks.md [hunk 30]
```diff
@@ -0,0 +1,436 @@
...
+In contrast, this proposal introduces Template Stacks with the limited ability to use their CR and child resources as input to create, update, or delete child Kubernetes resources. Application owners and managers can take advantage of Template Stacks to package complex deployments, while users benefit from a simple install and usage interface. Developers can use Template Stacks as application owners would, to create Kubernetes resource machinery using template engine logic, circumventing the need more complex controllers.
+
+The Template Stack Manager will permit the use of template variables to create and manage a set of resources provided as a set of strings included in the definition of the Stack.
+
+The benefits of Template Stacks are as follows:
+
+* [Simple Variable Substitution](#Simple-Variable-Substitution)
+
+  Stack Template resources can take input from a CR spec or from sibling resources.
+* [No Custom Controller Needed](#No-Custom-Controller-Needed)
```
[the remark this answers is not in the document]
displague: I am using a single list here.  Each bullet has additional text underneath it.  What appears to be a blank line is the extra newline needed to prevent the additional text from each bullet from wrapping on the first line of the bullet.

I've included an example here, but it seems that Github Markdown doesn't need the extra newline.  My IDE's markdown preview did need it.  The rendering doesn't suffer from wrapping on the first line. I'll remove it.

```markdown
* **bulletA**

   a second line of text about bulletA
* **bulletB**
   a line of text I expected would wrap into the first line of bulletB
```

---

* **bulletA**

   a second line of text about bulletA
* **bulletB** 
   a line of text I expected would wrap into the first line of bulletB

---

design/one-pager-template-stacks.md
```diff
@@ -0,0 +1,925 @@
...
+We will not focus on how the names of adjacent templates are exposed in a
+template syntax, just that they should be made available.
+
+The set of variables to present to a template should include:
+
+* The Stack managed object (the `Foo`)
+* The set of dependent resource objects (the `DependentThing`)
+* The set of errors encountered rendering each dependent resource
+  This will allow `Foo` to reflect errors rendering `DependentThing` as a conditioned status using `templateStatus`.
+* Helper functions for string manipulation, list handling, and simple numeric operations, such as <https://github.com/leekchan/gtf#index>
```
displague: @turkenh points out https://github.com/Masterminds/sprig as well :+1:

design/one-pager-template-stacks.md
```diff
@@ -0,0 +1,679 @@
...
+        statusField: {{.some.value}}
+        {{end}}
+```
+
+#### Horizontal Composition
+
+We've now demonstrated several properties of Template Stacks:
+
+* a means to concatenate Kubernetes Resource YAML
+* nesting supported CRD types or the CRDs of other Stacks
```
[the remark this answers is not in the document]
displague: I updated the previous section to better explain this:
https://github.com/crossplaneio/crossplane/pull/928/files#diff-bc9a72d48c0cb5e0a8178de9d945293dR246

design/one-pager-template-stacks.md
```diff
@@ -0,0 +1,679 @@
...
+          - name: redis-controller
+            image: example/redis-controller:{{.spec.redisVersion}}
+```
+
+*In this example, templates are nested within a map keyed by the GVK and a
+template grouping name. The reasoning for this map format will be explained in
+the [One Pager for Template Stack Fields](one-pager-template-stack-fields).*
+
+With a more complex example, we see that each template needed to render a GVK
+should be provided with a name that represents that intent of that template.
```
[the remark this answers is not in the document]
displague: I don't go into the unpack logic here.  I've left the filesystem of a stack up for UX.  

The current resource files are `kind.resource.yaml`.  With the template map structure defined here, per CRD, we could use `kind.template.yaml` and include a set of named templates in that file.

```yaml
# foo.template.yaml
fooA:
  apiVersion: group/version
  kind: foo
  ...
bar:
  apiVersion: group/version
  kind: bar
  ...
```

Or if we switch to the template format that you've proposed offline, we may prefer:

```yaml
# kind.arbitraryname.tpl
apiVersion: group/version
kind: Foo
spec:
  field: {{.var}}
```

```yaml
# kind.templates.yaml
resources:
- arbitraryname
- othername
status:
- statustemplatename
```

design/one-pager-template-stacks.md
```diff
@@ -0,0 +1,679 @@
...
+
+* For every new resource, create a set of Kubernetes resources
+* Use the spec fields of that resource to affect the Kubernetes resources
+* Upon changes to these spec fields, update the Kubernetes resources
+
+These are responsibilities the Template Stack Manager will handle.
+
+### Composition
+
+#### Vertical Composition
```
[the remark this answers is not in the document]
displague: This is the idea that stacks and templates can be applied in layers.  I should probably say that forthright.

design/one-pager-template-stacks.md
```diff
@@ -0,0 +1,679 @@
...
+would actually be `{{.spec.version}}`. If this were variable was included in the
+template body of a Template Stack for `Example` resources, the `.spec.version`
+of each `Example` resource would be applied to the template and rendered as new,
+updated, or removed Kubernetes resources.
+
+### No Custom Controller Needed
+
+This is where the value of Template Stacks starts to materialize. A stack author
+doesn't need to write their own controller, because a common controller is
+provided with the Template Stacks Manager.
```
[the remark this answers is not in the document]
displague: That was suggested here:
https://github.com/crossplaneio/crossplane/pull/928/files#diff-bc9a72d48c0cb5e0a8178de9d945293dR103-R115 
Perhaps the wording could be moved around or be made clearer about what template stacks would be missing from the traditional controller experience i.e. "can not make arbitrary storage, compute, and API calls".

design/one-pager-template-stacks.md
```diff
@@ -0,0 +1,1009 @@
...
+kind: Stack
+metadata:
+  name: Foo
+spec:
+  customresourcedefinitions:
+  - kind: Foo
+    apiVersion: group/version
+  templates:
+    # the GVK for the CRD the Stack consumer will create
+    foo.group/version:
```
displague: This should be updated to include `sources`, `behaviors`, `resources`, `defaults`, and `status` (per #956).

https://github.com/crossplaneio/crossplane/pull/956/files#diff-e81d9b5b6cbb8b4bdfc4d7a5ef0f2f25R261-R293

design/one-pager-template-stacks.md
displague: Mentioned in 66983657

design/one-pager-template-stacks.md
```diff
@@ -0,0 +1,679 @@
...
+resources. Template Stacks should offer this same capability.
+
+```yaml
+kind: Stack
+metadata:
+  name: StatusExampleStack
+spec:
+  templateStatus:
+    foo.example.org/v1:
+      controller: |
```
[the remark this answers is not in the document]
displague: This was an error. `templateStatus` should only refer to the CRD, not a specific dependent resource.

The templateStatus is applied to the `status` of the "foo", not the "subfoo".  "subfoo" template values would be available without having to use them in a "controller" scope.

Something like this should be possible:

```
  templateStatus:
    foo.example.org/v1: |
        statusField: {{.controller.value}}
```

design/one-pager-template-stacks.md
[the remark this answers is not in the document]
displague: I'm leaving it up to https://github.com/crossplaneio/crossplane/pull/956 and related design updates to sort out how templates, status, and resources are defined.

design/one-pager-template-stacks.md
displague: Deferring these updates to #956.

design/one-pager-template-stacks.md
```diff
@@ -0,0 +1,1110 @@
...
+
+Rather, a
+[`Result{RequeueAfter}`](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/reconcile/reconcile.go#L30-L32)
+on `Foo` would be sufficient for an initial implementation of Template Stacks.
+
+On each reconciliation, the dependent resources will be fetched (for use as
+template variables), and then the template will be rendered, and the dependent
+resources will be updated (or created or deleted).
+
+There should be advantages to [watching each dependent
```
jbw976: we'll definitely want to improve on a timer based approach not too long after initial implementation both for improved user experience and decreased API server load.

PR #930, opened 2019-10-11
One pager on Stacks Status spec and related changes/additions

rathpc:

This PR defines the first draft of the one pager that designs the spec around adding keys to drive status information in a GUI.

Addresses: https://github.com/crossplaneio/crossplane/issues/929

Related: https://github.com/crossplaneio/sample-stack/pull/4

I have:

### DIFF
--- a/design/one-pager-stack-status-spec.md
+++ b/design/one-pager-stack-status-spec.md
@@ -1,0 +1,82 @@
+# Stack UI Schema Status Spec
+
+- Owner: Steven Rathbauer (@rathpc)
+- Reviewers: Crossplane Maintainers
+- Status: Draft
+
+## Proposal
+
+[crossplaneio/crossplane#929](https://github.com/crossplaneio/crossplane/issues/929)
+
+## Problem
+
+As of right now there is no defined spec to be able to drive status information for a given stack or resource from a GUI perspective. It would be useful to add keys to the existing `ui-schema.yaml` file that can be used for this purpose.
+
+We can leverage the types from the array of **conditions** within the **conditionedStatus** key defined in **status**.
+
+This 1-pager will identify the keys that should be added and what that structure looks like.
+
+## Design
+
+#### Add two new keys to the `ui-schema.yaml` file:
+
+- `resourceConditionedStatusTypes` _(this is an array of condition types as strings)_
+- `stackConditionedStatusTypes` _(this is an array of condition types as strings)_
+
+#### The arrays could potentially look like this inside of the `ui-schema.yaml` file:
+
+```yaml
+version: 0.4
+configSections: 
+- title: Title
+  description: Description
+resourceConditionedStatusTypes:
+- Ready
+stackConditionedStatusTypes:
+- Ready
+```
+
+- You can use that key to get the various values you want for a GUI related to that type.
+
+- For example if you wanted a status title, status value and additional details for the 'Ready' status type you could build an object like this for example (_Using JS for example purposes_):
+
+```js
+import { JSONPath } from 'jsonpath-plus';
+
+// ...
+
+const statusTypesFromUISchema = resourceConditionedStatusTypes;
+
+const statusInfo = statusTypesFromUISchema.map(statusType => ({
+    title: JSONPath({
+        path: `$.status.conditionedStatus.conditions[?(@.type=='${statusType}')].type`,
+        json: crdReturnedFromK8SAsJSON,
+    }),
+    value: JSONPath({
+        path: `$.status.conditionedStatus.conditions[?(@.type=='${statusType}')].status`,
+        json: crdReturnedFromK8SAsJSON,
+    }),
+    details: JSONPath({
+        path: `$.status.conditionedStatus.conditions[?(@.type=='${statusType}')].reason`,
+        json: crdReturnedFromK8SAsJSON,
+    }),
+}));
+
+// ...
+```
+
+- You can use a javascript implementation of JSONPath ([jsonpath-plus](https://github.com/s3u/JSONPath)) as shown above to leverage these keys and "_look up_" the status values and additional info if present.
+
+-----
+
+## Fallback Option
+
+If a package author does not include these additional keys in the `ui-schema.yaml` file **OR** if there is no `ui-schema.yaml` file included at all, you can try to leverage the **additionalPrinterColumns** values if they exist within the spec.
+
+```yaml
+spec:
+  additionalPrinterColumns:
+    - JSONPath: .status.conditionedStatus.conditions[?(@.type=='Ready')].status
+      name: READY
+      type: string
+```
--- a/design/one-pager-stack-status-spec.md
+++ b/design/one-pager-stack-status-spec.md
@@ -10,18 +10,39 @@
 
 ## Problem
 
-As of right now there is no defined spec to be able to drive status information for a given stack or resource from a GUI perspective. It would be useful to add keys to the existing `ui-schema.yaml` file that can be used for this purpose.
+For a given CRD I want to be able to expose various statuses to a frontend GUI. Initially it makes
+sense to at least have the Ready state available. In addition to the value of the Ready state it
+would also be helpful to have some additional details about that state if applicable.
 
-We can leverage the types from the array of **conditions** within the **conditionedStatus** key defined in **status**.
+This also assumes that statuses are following a particular condition convention, similar to that of a
+[podcondition](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.16/#podcondition-v1-core)
 
-This 1-pager will identify the keys that should be added and what that structure looks like.
+The conditions should contain at least the following keys:
+- type
+- status
+- reason
+- message
+
+#### For example:
+
+If a CRD has the condition type 'Ready' defined but has a status of that type of
+_"False"_, I could show some visual indication to a user through a GUI that the CRD is **NOT Ready**.
+
+When that status changes to _"True"_ then I can adjust that visual cue to now reflect that the CRD is
+in fact **Ready**.
+
+As of right now there is no defined spec to be able to drive status information for a given resource from a GUI perspective. It would be useful to add a key to the existing `ui-schema.yaml`
+file that can be used for this purpose.
+
+This key would contain an array of JSONPath's that point directly to conditions following the convention described above.
+
+This 1-pager will identify the key that should be added and what that structure looks like.
 
 ## Design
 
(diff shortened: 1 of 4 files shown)


pkg/stacks/unpack_test.go
```diff
@@ -531,15 +555,23 @@ metadata:
...
       ---
-      version: 0.3
       configSections:
       - title: sibling Title
         description: sibling Description
+      resourceConditionedStatusTypes:
+      - Ready
+      stackConditionedStatusTypes:
+      - Ready
+      version: 0.4
```
displague: The diffs would be cleaner if the `version: 0.4` lines replaced the `version: 0.3` lines.  Any reason why the version is being tacked on to the bottom of the examples instead of the top?

rathpc: Really the only reason was just to keep it all alphabetical, I can move them back up if needed though

rathpc: just updated

design/one-pager-stack-status-spec.md
```diff
@@ -0,0 +1,82 @@
...
+- Reviewers: Crossplane Maintainers
+- Status: Draft
+
+## Proposal
+
+[crossplaneio/crossplane#929](https://github.com/crossplaneio/crossplane/issues/929)
+
+## Problem
+
+As of right now there is no defined spec to be able to drive status information for a given stack or resource from a GUI perspective. It would be useful to add keys to the existing `ui-schema.yaml` file that can be used for this purpose.
```
suskin: What would you think of wrapping the long lines earlier? It makes it easier to comment on the lines in github since the lines end up shorter, and they show up the same in the rendered markdown : )

design/one-pager-stack-status-spec.md
```diff
@@ -0,0 +1,82 @@
...
+
+We can leverage the types from the array of **conditions** within the **conditionedStatus** key defined in **status**.
```
suskin: Are conditions a common pattern in Kubernetes? (My initial research indicates that it [probably is](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-conditions)). If so, we may be able to just define a default behavior that doesn't require any configuration. And then we could offer configurable behavior for people who want to override the default behavior.

design/one-pager-stack-status-spec.md
```diff
@@ -0,0 +1,82 @@
...
+}));
+
+// ...
+```
+
+- You can use a javascript implementation of JSONPath ([jsonpath-plus](https://github.com/s3u/JSONPath)) as shown above to leverage these keys and "_look up_" the status values and additional info if present.
+
+-----
+
+## Fallback Option
```
suskin: Nice - I like that there is a fallback option described in the one-pager.

pkg/stacks/unpack_test.go
```diff
@@ -246,15 +246,23 @@ metadata:
     stacks.crossplane.io/icon-data-uri: data:image/svg+xml;base64,bW9jay1pY29uLWRhdGEtc3Zn
     stacks.crossplane.io/stack-title: Sample Crossplane Stack
     stacks.crossplane.io/ui-schema: |-
-      version: 0.3
+      version: 0.4
```
suskin: Do we normally change the design files and the go files in the same pull request? Not opposed to it, but I'm asking because I would usually expect them to happen in separate pull requests, and I want to see if I'm just not aware of a pattern we have.

design/one-pager-stack-status-spec.md
jbw976: kubernetes has a general concept of conditions for status, but the implementation around `conditionedStatus` and query paths like what's found below are a crossplane specific concept. 

```
status.conditionedStatus.conditions[?(@.type=='${statusType}')]
```

This assumption and approach that everything has a `conditionedStatus` field would not hold true for arbitrary types defined by arbitrary stacks.  Have we considered a more general approach here?  Could the `ui-schema.yaml` allow the author to define a full query path instead of just a key name that gets injected into a hard-coded query path that assumes everything is using `conditionedStatus`?

design/one-pager-stack-status-spec.md
jbw976: This problem statement could use more background information.  It doesn't provide necessary context for a new reader to this subject to understand what is being discussed and why it is needed.  Can you please add more background and/or examples about what the design wants to accomplish?

pkg/stacks/unpack_test.go
rathpc: That is a good question and I am not sure I have an answer to it, it just felt appropriate to update this file. @jbw976 any thoughts here?

design/one-pager-stack-status-spec.md
rathpc: Those are both good points.

I was originally trying to limit the amount of configuration necessary while still having a valuable result. It seems that unless there is a concrete standard we may not be able to accommodate a "simple" config.

I have some thoughts on a solution but I need to sort them out. Let me do some thinking and get back to this.

design/one-pager-stack-status-spec.md
rathpc: Will add more info

rathpc: I can do that!

design/one-pager-stack-status-spec.md
rathpc: @suskin and @jbw976 

What do both of you think about this?

I think we can be flexible with this array and accept either a full path **OR** just a condition type string. So if an author is abiding by the crossplane specific concept, they have less work and can just pass the types as strings in the array, however it can also accept an array of full paths.

The only caveat here would be to strictly enforce if a path is sent, that the object it ultimately resolves to, follows the structure of a **[podcondition](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.16/#podcondition-v1-core)**. This will ensure that all of the properties we expect within that object to be able to leverage from a GUI perspective can still be accessible.

Does that make sense? I would be happy to elaborate if needed.

design/one-pager-stack-status-spec.md
```diff
@@ -0,0 +1,114 @@
...
+
+[crossplaneio/crossplane#929](https://github.com/crossplaneio/crossplane/issues/929)
+
+## Problem
+
+For a given CRD I want to be able to expose various statuses to a frontend GUI. Initially it makes
+sense to at least have the Ready state available. In addition to the value of the Ready state it
+would also be helpful to have some additional details about that state if applicable.
+
+This also assumes that statuses are following the podCondition object convention as described here:
```
jbw976: I think something I'm struggling with here is that not everything is a pod nor will they have the same semantics that pod conditions have.  In the core k8s types, there are different condition types for different API types.  Here are a few examples:

* `PersistentVolumeClaimCondition`: https://github.com/kubernetes/api/blob/master/core/v1/types.go#L487
* `NodeCondition`: https://github.com/kubernetes/api/blob/master/core/v1/types.go#L4511
* `ComponentCondition`: https://github.com/kubernetes/api/blob/master/core/v1/types.go#L5570

And each one of them has their own set of available condition types, e.g. [`ComponentConditionType`](https://github.com/kubernetes/api/blob/master/core/v1/types.go#L5566) has a totally separate set of types than [`NodeConditionType`](https://github.com/kubernetes/api/blob/master/core/v1/types.go#L4497-L4508)

I think it's reasonable that for surfacing status we can standardize on conditions, and that those conditions can have `type`, `status`, `reason`, and `message` fields.  But I'm not sure how we can best standardize on the **semantics** of those conditions, i.e. what each one actually **means**.

One CRD may say it's ready with a `Ready` condition, while another CRD may say the same thing with `Available`.

What do you think about how we can standardize on **semantics** or **meaning** of conditions in a world of arbitrary CRDs?

design/one-pager-stack-status-spec.md
```diff
@@ -0,0 +1,114 @@
...
+the full JSONPath as a string within that array as long as it follows the podcondition convention.
+
+This 1-pager will identify the keys that should be added and what that structure looks like.
+
+## Design
+
+#### Add two new keys to the `ui-schema.yaml` file:
+
+- `resourceStatusTypes` _(this is an array of condition types **OR** JSONPaths)_
+- `stackStatusTypes` _(this is an array of condition types **OR** JSONPaths)_
```
jbw976: can you explain what each of these are and how they are expected to be used?  i.e. what do we mean by stack status? is that the overall status of the entire stacks health? is that expected to be exposed on just one of the stack's CRDs?

design/one-pager-stack-status-spec.md
```diff
@@ -0,0 +1,114 @@
...
+
+#### The arrays could potentially look like this inside of the `ui-schema.yaml` file:
+
+```yaml
+version: 0.4
+configSections: 
+- title: Title
+  description: Description
+resourceStatusTypes:
```
jbw976: should we call these `*ConditionTypes` instead, to be more inline with the upstream k8s API terminology?  https://github.com/kubernetes/api/blob/master/core/v1/types.go#L2433-L2447

design/one-pager-stack-status-spec.md
```diff
@@ -0,0 +1,114 @@
...
+- Ready
+stackStatusTypes:
+- .status.conditionedStatus.conditions[?(@.type=='Ready')]
```
jbw976: what do you think about the usability of overloading these new `*Types` keys with both type names and query paths?  Do you think it would be more intuitive and understandable to have 4 new keys instead?

* `resourceConditionTypes`
* `resourceConditionPaths`
* `stackConditionTypes`
* `stackConditionPaths`

design/one-pager-stack-status-spec.md
jbw976: Or, do we not even need to standardize on meaning of conditions? maybe that doesn't have to be a goal?  A GUI could blindly display the conditions that are surfaced without any understanding of its meaning, if we wanted to go that way.

rathpc: I will be updating this to reference the following of a convention rather than a specific type

design/one-pager-stack-status-spec.md
rathpc: I will be updating this namin g

design/one-pager-stack-status-spec.md
rathpc: I am going to be revising what these arrays will contain to eliminate the confusion here

design/one-pager-stack-status-spec.md
rathpc: I am removing stack related status types

PR #931, opened 2019-10-13
design: release process and engineering

### LINKED ISSUE #909: Release automation design doc
part of #864 

In order to improve our release automation and ability to release frequently for #864, we should understand what it is that we are going to build and change to accomplish that goal.

A 1-pager should be authored that describes the scope of changes for release engineering and includes details about implementation that will be performed.

jbw976:

This PR contains a design doc that outlines improvements we can make to our release process to increase the frequency of releases and the level of automation in the process.

Fixes #909 

### Checklist
I have:

[skip ci]

### DIFF
--- a/design/one-pager-release-process.md
+++ b/design/one-pager-release-process.md
@@ -1,0 +1,331 @@
+# Release Process and Engineering
+
+## Background
+
+The current lifecycle process for a new version of the Crossplane software has gotten the project
+through the `v0.3` release, but also has a number of shortcomings. A summary of the current process
+that is very succinct, and skips many of the details, would look like the following:
+
+1. The maintainer team creates a [milestone](https://github.com/crossplaneio/crossplane/milestones)
+   to represent the current/active release
+1. Issues are added to the milestone and taken through their lifecycle on the [project
+   board](https://github.com/crossplaneio/crossplane/projects)
+1. The community continues to discuss the scope of the milestone and what issues should be included,
+   through up front planning sessions, backlog grooming, and daily stand-ups
+1. When all issues for the milestone have been completed and merged into `master`, the official
+   release process begins
+    1. The HEAD commit from `master` is tagged with the release version (e.g.,
+       [`v0.3.0`](https://github.com/crossplaneio/crossplane/releases/tag/v0.3.0)) and the release
+       branch is created (e.g.,
+       [`release-0.3`](https://github.com/crossplaneio/crossplane/tree/release-0.3))
+    1. Build all release artifacts of the software (e.g., binaries, container images, Helm charts,
+       documentation, stack packages) for all supported architectures (`amd64`, `arm64`)
+    1. Publish all artifacts to their hosted locations (GitHub, Docker Hub, crossplane.io, S3, etc.)
+    1. Promote release to relevant channel (e.g. `alpha`)
+
+### Shortcomings of Current Process
+
+The process outlined above has a number of deficiencies that we would like to improve on:
+
+1. Each release cycle is very long.  We did not release v0.3 for more than 5 months.
+1. New features and improvements are not getting into the hands of the community quickly, preventing
+   problems from being solved and feedback from being collected in a timely fashion.
+1. Each long release culminates with a big deadline for a large number of features. The team feels
+   immense stress and pressure because missing the current release means months until their work can
+   be included in the next one.
+1. The release process itself has a lot of manual steps that are time consuming, error prone, and
+   insecure. The cost of performing a release is high and inhibits faster releases from happening.
+
+## Proposed Improvements
+
+In this section, we will cover the set of proposed changes to improve our release engineering and
+process.
+
+## Summary
+
+For convenience, a summary of the proposal is included below:
+
+* Increase the frequency of releases to a consistent **monthly** schedule
+* Follow [semantic versioning](https://semver.org) so that each monthly release is a minor release
+  (e.g. `v0.4.0`)
+* Run patch releases (e.g., `v0.4.1`) as needed to address issues with quality and stability in the
+  previous release
+* Add pipelines to Jenkins that leverage the [build submodule](https://github.com/upbound/build/) to
+  automate release tagging/branching, building, publishing, and promoting
+
+## Release Schedule
+
+Increasing our release cadence will have multiple benefits for both users and contributors of the
+project. Users will be able to get access to new features and improvements at a faster pace with
+less wait time. Contributors will have more frequent opportunities to include their work into a
+release, which will decrease the stress and pressure we currently feel leading up to each large and
+long awaited release.
+
+We propose that a desirable frequency of release will be **monthly**, as that will provide a
+predictable and consistent schedule that gets new features out to the community at a reasonable
+pace.  We currently execute on work items in a series of sprints, each lasting 2 weeks.  To
+accommodate a monthly release schedule, this sprint schedule will need to be updated.
+
+As part of the timeline that we use to release software, there are many other events along the way
+besides the release itself that facilitate a healthy software development lifecycle. A brief
+description of these events and their associated goals and expectations can be found below:
+
+* **Milestone planning**: Before a milestone begins, we need to discuss the high level priorities
+  and themes, and identify the epics that will deliver on these priorities.
+* **Epic discovery**: After epics are identified, we need to do some discovery and investigation
+  work to appropriately scope the epics, identify the risks, and agree upon the epic's definition of
+  done.
+* **Weekly planning**: On a weekly basis, the team can get together for a more focused status check
+  with potential course correcting.  We can identify if any priorities have changed and if any
+  resources or efforts need to be reorganized. At the end of this meeting, everyone should have a
+  clear understanding of what their work will focus on for the week. This is similar to the goal of
+  sprint planning previously.
+* **Stand-up scrums**: This is a daily event where we quickly meet to talk about our progress, if we
+  are blocked on anything, and if we need to have any break-out sessions to help work through
+  issues.
+* **Backlog grooming and triage**: The backlog of issues and work items needs to be addressed and
+  explored on a regular basis, especially in response to user demand.  In this session, we can
+  triage new issues to understand their priority, we can revisit items in the backlog to see if any
+  priorities have changed, and we can work on the epic discovery mentioned above if that also needs
+  attention.
+* **Demo party**: This is a chance to celebrate and show off the fruits of our recent labors to the
+  rest of the team.  Each developer can show a demo of new things they have been working on to
+  collect feedback and disseminate information about our progress and limitations to the rest of the
+  team.
+* **Retrospective**: After a milestone is completed, a retrospective look back over it can be very
+  helpful at identifying where we were successful and what we need to improve on.  This retro should
+  result in a series of action items that will help us improve our processes.  These action items
+  should be captured as issues in GitHub as appropriate so they can be formally tracked and included
+  in planning.
+
+In an attempt to include all of these events in a **monthly** release timeline, we propose the below
+regularly recurring schedule for each month. Given our current trajectories, the actual release
+itself will occur mid-month, e.g. around the **15th of the month**.
+
+Note that every month will have different alignment for weekdays and weekends, so this schedule is a
+guide to the overall structure, not a set of specific dates that will be used every month. The team
+should define the specific days/events of each monthly schedule **before** its cycle starts and
+those dates should be published to a new `release-schedule.md` that is linked from the main readme
+and roadmap.
+
+| Day of release cycle | Event | Expectations |
+| ------------ | ----- | ------------ |
+|**Week 1**|||
+|1 Mon|Milestone starts|(no actual event or meeting held for this)|
+|1 Mon|Retrospective from previous release|A public formal retrospective is held for the community to participate in. The events of the previous milestone will be discussed to identify improvements. These improvements should be taken into the first planning of the milestone.|
+|2 Tues|Weekly planning|Group planning session to identify priorities and focus for the week.|
+|3 Wed|Backlog grooming & epic discovery|(optional) Session to triage new issues and help investigate and flesh out epics.  Early in the milestone, the emphasis is more on epic discovery since we are still working through understanding the epics.|
+|**Week 2**|||
+|8 Mon|Weekly planning|Group planning session to identify priorities and focus for the week.|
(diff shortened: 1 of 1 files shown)


design/one-pager-release-process.md
```diff
@@ -0,0 +1,172 @@
...
+
+Currently, our tagging logic performs the tag on the HEAD commit of `master` and then creates the release branch.
+This means that the very first build out of the release branch will be the official minor release, since you cannot re-tag in `git` without essentially rewriting history.
+If we find an issue within final testing of the release build, we'd essentially need to either rewrite the tag history or run a patch build to release instead.
+
+This issue can also happen when we update the [example yaml files](https://github.com/crossplaneio/crossplane/blob/release-0.3/cluster/examples/stacks/stack-aws.yaml#L12) to embed the current release version in them.
+We would not want to make this update in `master` before tagging and branching, since that change is not relevant in `master`,
+but we also don't want to do it in the release branch after we've already tagged.
+One proposal would be to separate the tagging and branching functionality so that we branch from `master` first,
+make any updates/commits in the new release branch, and then finally tag the final commit within the release branch (not `master`).
```
hasheddan: I like this strategy. Have we thought at all about potentially cutting the release branch some period of time (it seems like a week might be appropriate given our cadence) before the release date and instituting a "code freeze" (i.e. "feature complete") phase?

design/one-pager-release-process.md
```diff
@@ -0,0 +1,172 @@
...
+Since these stack repositories currently also take advantage of the build submodule, they will also follow a very similar process as Crossplane and receive a similar set of improvements.
+
+Recall that a Stack is simply just an [OCI image](https://github.com/crossplaneio/crossplane/blob/master/design/design-doc-stacks.md#stack-package-format).
+Therefore, the build and publishing process can follow essentially the same process that Crossplane uses.
+Other functionality in the build submodule, such as creating a Helm chart, is not necessary for Stacks, but relevant functionality should definitely be reused.
+For example, building and publishing the stack package, tagging and creating a release branch, promoting the release, etc.
+
+The GCP, AWS, and Azure stacks will all receive new `tag` and `promote` pipelines similar to Crossplane, as well as receive a fix for [#330](https://github.com/crossplaneio/crossplane/issues/330).
+
+Note that each Stack can be released independently and on their own schedule, but for now they will follow the same monthly release schedule as Crossplane so the community can receive frequent updates.
```
hasheddan: Will each stack receive patch releases with Crossplane? Can we do patch releases of individual stacks out of cadence with Crossplane and other stacks?

design/one-pager-release-process.md
```diff
@@ -0,0 +1,172 @@
+# Release Process and Engineering
+
+## Background
+
+The current lifecycle process for a new version of the Crossplane software has gotten the project through the `v0.3` release, but also has a number of shortcomings.
+A summary of the current process that is very succinct, and skips many of the details, would look like the following:
+
+1. The team creates a [milestone](https://github.com/crossplaneio/crossplane/milestones) to represent the current/active release
```
displague: For this OSS project, it might help to be more specific about the team members involved here.

"maintainer team"? "maintainers"?

design/one-pager-release-process.md
```diff
@@ -0,0 +1,172 @@
...
+1. Issues are added to the milestone and taken through their lifecycle on the [project board](https://github.com/crossplaneio/crossplane/projects)
+1. The team continues to discuss the scope of the milestone and what issues should be included, through up front planning sessions, backlog grooming, and daily stand-ups
+1. When all issues for the milestone have been completed and merged into `master`, the official release process begins
+    1. The HEAD commit from `master` is tagged with the release version (e.g., [`v0.3.0`](https://github.com/crossplaneio/crossplane/releases/tag/v0.3.0))
+    and the release branch is created (e.g., [`release-0.3`](https://github.com/crossplaneio/crossplane/tree/release-0.3))
```
displague: I find release branches can be redundant to release tags. In the past we I have used a release branch until it was released (and tagged) and then scrapped the release branch.

The release branch could also act as a branch that holds the latest v0.3.x tag.

What's your take?

displague: I grep'd for `release-` and missed this section which explains the release branch.

https://github.com/crossplaneio/crossplane/pull/931/files#diff-f31a09f3f47ef35a6f5d5b83aa497071R78-R80 :+1:

design/one-pager-release-process.md
```diff
@@ -0,0 +1,172 @@
...
+
+In the meantime, every monthly release will be a **minor** release, meaning we will be incrementing the minor version number.
+For example, the next 3 monthly releases of Crossplane will be `v0.4.0`, `v0.5.0`, and `v0.6.0`.
+
+### Patch Releases
+
+In addition to monthly minor releases, we will also be able to perform patch releases in the event that a severe issue is identified in the last release.
+Patch releases will increment the `PATCH` version field, for example a patch for the `v0.4.0` release will have a version number of `v0.4.1`.
+
+The exact criteria for what justifies performing a patch release is difficult to identify since every issue is unique.
```
displague: How many versions will we commit to supporting.  Will patches be released for `v0.4` once `v0.5.0` is released? What of `v0.3`?

design/one-pager-release-process.md
```diff
@@ -0,0 +1,172 @@
...
+### Upgrade
+
+While Crossplane and Stacks do not have an officially supported upgrade path, this proposal for more frequent updates may cause some pain within the community.
+This isn't ideal but can be acceptable while we are still in alpha.
+We should prioritize working out our upgrade story to remove this obstacle and allow the community to take a hard dependency on running Crossplane across versions.
+
+### Commits after tagging
+
+Currently, our tagging logic performs the tag on the HEAD commit of `master` and then creates the release branch.
+This means that the very first build out of the release branch will be the official minor release, since you cannot re-tag in `git` without essentially rewriting history.
```
suskin: Nitpick: I would describe the reason to not re-tag as more because it goes against what people expect out of the behavior of a tag, rather than because it is rewriting history. The standard expectation of a tag is that it is immutable, and git makes it hard to change a tag because of this, **not** because history is being rewritten. It used to be the case that `git fetch --tags` was sufficient to update changed tags, but it [looks like](https://git-scm.com/docs/git-fetch#Documentation/git-fetch.txt---force) the command is now more like `git fetch --tags --force`. Anyway, there's some [additional discussion about expectations in the manpage for git tag](https://git-scm.com/docs/git-tag#_on_re_tagging).

negz:
I haven't made it all the way through this PR yet so apologies if this is already covered, but I wanted to drop some breadcrumbs for https://github.com/crossplaneio/crossplane-runtime/issues/11.

I liked @hasheddan's suggestion of copying the controller-runtime versioning patterns for crossplane-runtime. Specifically I like that they encourage contributors to tag their PRs so that maintainers can easily determine what PRs are breaking changes, new features, or bug fixes. Unfortunately I am a very grumpy old man and found their use of emojis in the PR titles hard to swallow, so I went down a rabbit role trying to figure out if we could use Github labels for this somehow, then got distracted and gave up.

design/one-pager-release-process.md
```diff
@@ -0,0 +1,172 @@
...
+It is believed that [#330](https://github.com/crossplaneio/crossplane/issues/330) may be the root cause for this and we should fix this issue to get automatic publishing of release builds.
+
+* **tag**: This new pipeline would be run on the `master` branch when we have identified the release candidate commit and are ready to tag it and create the release branch.
+This pipeline would run the following build command:
+
+    ```bash
+    ./build/run make tag VERSION=${params.version}
+    ```
+
+* **promote**: This new pipeline would perform the promotion of a release build to a particular channel.
```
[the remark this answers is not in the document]
jbw976: All automated tests should be run as part of the build pipeline as they currently are.  We need to make more of an investment in those tests, but that's the ideal place for them to run.  `promote` is actually just a very simple publishing metadata change only, artifacts are not regenerated at all, so they will have already been run through the testing process by the time we get to the `promote` step.

design/one-pager-release-process.md
jbw976: Yes, we can do patch releases for stacks completely on their own schedule as well.  They can also even NOT have a patch release while other stacks or crossplane does get one.  I'll add a note for that.

design/one-pager-release-process.md
jbw976: sounds good, i'll change the wording here to tags being immutable as opposed to re-writing history.

design/one-pager-release-process.md
jbw976: Sounds good guys, I'll update the process so that we are tagging after we have branched and vetted the release candidate.  Yes @hasheddan, a code freeze period is something I want to incorporate into the concrete schedule I'll be adding soon.  A code freeze period would actually work well in the following flow:

1. `master` has all the features we're expecting to include in the release
1. create the release branch, but do not tag the release yet
1. "code freeze" period starts but is only applicable to the release branch.  `master` can continue taking changes.
1. only fixes to stabilize the release are accepted into the release branch during code freeze
1. testing and vetting occurs, every thing looks good.
1. tag the final commit in the release branch, e.g. `v0.4.0.`.
1. publish the release
1. profit 💰

jbw976:
>  I wanted to drop some breadcrumbs for crossplaneio/crossplane-runtime#11.

@negz I think it would be reasonable to include in this proposal a section on how we version "package" repos that do not have release artifacts and process on their own (people directly import the repo source code).  I don't think I'll get into any details about how PRs are labeled, but the general versioning approach seems in scope enough for this doc.

jbw976:
All feedback has been responded to and/or incorporated into the design.  The biggest addition to the document since its original version was for the specific release schedule, including all meetings and events that are proposed as part of monthly release lifecycle.  Please pay attention to that new schedule (captured in an easy to read table) and provide feedback.

^ @prasek @hasheddan @displague @suskin

design/one-pager-release-process.md
```diff
@@ -0,0 +1,321 @@
...
+  triage new issues to understand their priority, we can revisit items in the backlog to see if any
+  priorities have changed, and we can work on the epic discovery mentioned above if that also needs
+  attention.
+* **Demo party**: This is a chance to celebrate and show off the fruits of our recent labors to the
+  rest of the team.  Each developer can show a demo of new things they have been working on to
+  collect feedback and disseminate information about our progress and limitations to the rest of the
+  team.
+* **Retrospective**: After a milestone is completed, a retrospective look back over it can be very
+  helpful at identifying where we were successful and what we need to improve on.  This retro should
+  result in a series of action items that will help us improve our processes.
```
suskin: What is the recommendation for tracking retrospective action items and their progress?

design/one-pager-release-process.md
```diff
@@ -0,0 +1,321 @@
...
+|**Week 2**|||
+|8 Mon|Weekly planning|Group planning session to identify priorities and focus for the week.|
+|10 Wed|Backlog grooming|(optional) Session to triage new issues and prioritize the backlog.|
+|12 Fri|Halftime Demo party|Show off our progress in demos for the team and take in critical feedback and discuss potential course corrections to set effective strategy.|
+|**Week 3**|||
+|15 Mon|Weekly planning|Group planning session to identify priorities and focus for the week.|
+|17 Wed|Backlog grooming|(optional) Session to triage new issues and prioritize the backlog.|
+|**Week 4**|||
+|22 Mon|Weekly planning|Group planning session to identify priorities and focus for the week.|
+|25 Wed|Backlog grooming & Milestone planning|Session to triage new issues as we get closer to feature freeze.  Milestone planning begins for the **next** milestone as we identify themes, priorities, and high level epics that we want to address in the **next** release.|
```
suskin: Based on how some previous retrospectives have gone, I could see it being beneficial to hold the retrospective before planning, so that feedback from the retrospective can be incorporated into planning earlier. Otherwise, changes would need to wait until the following planning cycle, which is three weeks after the retrospective, and it may be hard to remember everything we wanted to do. How do we imagine that the process as written would incorporate feedback from the retrospectives?

That said, I am assuming we'll be adjusting the process over time, so I'm not sure this comment needs to block merging the document.

My proposal would be that if we did the retrospective on day 1, and also started with milestone planning on day 1 (with the intent that we'd finish it by the end of week 1), then it may be easier to incorporate the ideas from the retrospective into the planning. There could also be an additional benefit of allowing people to focus on the planning instead of closing the release. One drawback would be that there would be a reset period in between actual work on the milestones. What do other people think?

design/one-pager-release-process.md
```diff
@@ -0,0 +1,321 @@
...
+   merge commit)
+1. The cherry-pick PR is merged into the release branch
+1. The patch release pipeline automation is run with an incremented `PATCH` field from the last
+   release, e.g. `v0.4.1`.
+
+## Golden Master
+
+To facilitate the quality of releases and the ease of performing a release, we will be adopting a
+"golden master" philosophy for the `master` branch. Essentially, all changes that are accepted and
+merged into `master` should be **complete** and of high quality. In addition to the functional code
```
displague: To be able to move forward fast, it may be helpful to permit merges for partial features which have no breaking changes, no side-effects, and can only be accessed deliberately.  This could leave the new feature without documentation unless the onus is passed to the PR that exposes the feature.

For example, @suskin @turkenh and I were considering introducing the Template Stacks harnessing in piece-meal PRs rather than one larger PR later.

How would you characterize that endeavour as it relates to the golden master philosophy?

design/one-pager-release-process.md
jbw976: I can definitely understand the the benefits of having the retrospective so that informs the planning.  However, I'm struggling with how we'd be able to fit in the retrospective (either per SIG or joint one for the entire community) and run planning sessions (per SIG) all within the same day and also accommodate all time zones.

Maybe we could do the retrospective on day 1 and then day 2 has planning?  This schedule will be flexible and open to changes over time as we experience it and learn from it, so it's not entirely set in stone.

design/one-pager-release-process.md
jbw976: I think we should do a better job of opening issues for tracking action items.  I'll add a note for that.

design/one-pager-release-process.md
jbw976: Yes that's a good point @displague to be able to include features that must be deliberately accessed.  We don't have a feature gate functionality right now, but you could consider behavior requiring deliberate commands from the user (as opposed to behavior on mainline paths) to be similar to a feature gate.  I'll include a note on this.

jbw976:
I have addressed and incorporated all feedback on this final round, proceeding with merging this design doc now.

Issue #897, opened 2019-10-07
Stack Manager fails install if ClusterStackInstall created too quickly


hasheddan:
### What happened?

When creating integration tests for stacks, I wrote a script that installs the stack immediately after the `crossplane` and `stack-manager` pods become ready. This routinely caused the following failure with the stack installation:

Stack Manager Logs:
```
2019-10-04T18:13:11.942Z	DEBUG	crossplane.stackinstall.stacks.crossplane.io	reconciling	{"kind": "stacks.crossplane.io/v1alpha1, Kind=ClusterStackInstall", "request": "azure/stack-azure"}
2019-10-04T18:13:11.942Z	DEBUG	crossplane.util	getting pod	{"name": "crossplane-system/crossplane-stack-manager-7fd8d6b9fc-bf9w2"}
2019-10-04T18:13:12.532Z	DEBUG	crossplane.stackinstall.stacks.crossplane.io	created install job	{"jobRef": "&ObjectReference{Kind:,Namespace:azure,Name:stack-azure,UID:,APIVersion:,ResourceVersion:,FieldPath:,}", "jobOwnerRefs": [{"apiVersion":"stacks.crossplane.io/v1alpha1","kind":"Stack","name":"stack-azure","uid":"b693d18b-0389-4f19-b968-c9eeb2b0a493"}]}
2019-10-04T18:13:13.032Z	DEBUG	crossplane.stackinstall.stacks.crossplane.io	reconciling	{"kind": "stacks.crossplane.io/v1alpha1, Kind=ClusterStackInstall", "request": "azure/stack-azure"}
2019-10-04T18:13:13.032Z	DEBUG	crossplane.util	getting pod	{"name": "crossplane-system/crossplane-stack-manager-7fd8d6b9fc-bf9w2"}
2019-10-04T18:13:13.433Z	DEBUG	crossplane.stackinstall.stacks.crossplane.io	failed stack install	{"i": "stack-azure", "error": "Job.batch \"stack-azure\" not found"}
``` 

ClusterStackInstall:

```
Name:         stack-azure
Namespace:    azure
Labels:       <none>
Annotations:  kubectl.kubernetes.io/last-applied-configuration:
                {"apiVersion":"stacks.crossplane.io/v1alpha1","kind":"ClusterStackInstall","metadata":{"annotations":{},"name":"stack-azure","namespace":"...
API Version:  stacks.crossplane.io/v1alpha1
Kind:         ClusterStackInstall
Metadata:
[...]

PR #932, opened 2019-10-14
Add clusterStacks option to Crossplane helm chart

### LINKED ISSUE #867: Crossplane Helm chart accepts optional stack=gcp|aws|azure values file attribute
Crossplane work to enable [GitLab Auto DevOps integration phase 1](https://gitlab.com/groups/gitlab-org/-/epics/1866#note_216080986)

Part of the [San Diego Milestone](tbd)

### What seems to be the problem?
To support managed service provisioning in GitLab pipelines (DBaaS, cache, etc.), would like to add Crossplane as a GitLab managed app with a selected infrastructure Stack to install: GCP, AWS, or Azure that can be done from the GitLab managed apps GUI.

image

GitLab managed apps are Helm charts that GitLab can install to a connected Kubernetes cluster. The GitLab UI for installing managed apps will be updated to install the Crossplane Helm chart and pass the selected infrastructure stack in a helm values file, so the Crossplane Helm chart can install it along with core Crossplane.

### What could Crossplane do to help?
Update the Crossplane Helm chart to accept an optional stack attribute to install one of the cloud provider stacks (AWS, GCP, Azure)

### What does it look like when we're done?
 - Crossplane Helm chart accepts an optional values file `stack` attribute that if specified will install the associated infrastructure stack along with core Crossplane.
 - Crossplane v0.4.0 helm chart needs to continue supporting Helm 2 installs, since [GitLab 12.5 will continue to require Helm 2](https://gitlab.com/groups/gitlab-org/-/epics/1866#note_227663961).

### How could this be demonstrated?
This can be demoed stand alone by installing the Crossplane Helm chart (using Helm 2) with `--set stack=aws|gcp|azure` or as part of [GitLab #1866 phase 1](https://gitlab.com/groups/gitlab-org/-/epics/1866#note_216080986):
```
helm repo add crossplane-alpha https://charts.crossplane.io/alpha
helm install --set stack=aws|gcp|azure --name crossplane --namespace crossplane-system crossplane-alpha/crossplane
```

Verify the infrastructure stack is installed along with core Crossplane.

### Todo
 * [ ] Update Crossplane v0.4.0 Helm chart that will be published to the alpha channel to accept an optional `stack` attribute for installing one of the infrastructure stacks: AWS, GCP, Azure

turkenh:

This PR introduces `clusterStacks` option to Crossplane helm chart so that additional infrastructure stacks can be installed via --set command line option of helm.

Example:
```bash
helm install ... --set clusterStacks.gcp.deploy=true --set clusterStacks.gcp.version=v0.1.1
```
Above command will create `ClusterStackInstall` resource named `stack-gcp` with package`crossplane/stack-gcp:v0.1.1`. 

Default version for all stacks is set as latest released version, which is `v0.1.0` for all.

Closes #867 
### Checklist

I have:

### DIFF
--- a/cluster/charts/crossplane/templates/clusterstackinstalls.yaml
+++ b/cluster/charts/crossplane/templates/clusterstackinstalls.yaml
@@ -4,6 +4,8 @@ apiVersion: stacks.crossplane.io/v1alpha1
 kind: ClusterStackInstall
 metadata:
   name: "stack-{{ $key }}"
+  annotations:
+    "helm.sh/hook": post-install,post-upgrade
 spec:
   package: "crossplane/stack-{{ $key }}:{{ $val.version }}"
 ---
--- a/cluster/charts/crossplane/templates/clusterstackinstalls.yaml
+++ b/cluster/charts/crossplane/templates/clusterstackinstalls.yaml
@@ -1,9 +1,11 @@
-{{- range $arg := .Values.clusterStacks }}
+{{- range $key, $val := .Values.clusterStacks }}
+{{- if $val.deploy }}
 apiVersion: stacks.crossplane.io/v1alpha1
 kind: ClusterStackInstall
 metadata:
-  name: "stack-{{ $arg }}"
+  name: "stack-{{ $key }}"
 spec:
-  package: "crossplane/stack-{{ $arg }}:master"
+  package: "crossplane/stack-{{ $key }}:{{ $val.version }}"
 ---
+{{- end }}
 {{- end }}
--- a/cluster/charts/crossplane/templates/clusterstackinstalls.yaml
+++ b/cluster/charts/crossplane/templates/clusterstackinstalls.yaml
@@ -1,0 +1,9 @@
+{{- range $arg := .Values.clusterStacks }}
+apiVersion: stacks.crossplane.io/v1alpha1
+kind: ClusterStackInstall
+metadata:
+  name: "stack-{{ $arg }}"
+spec:
+  package: "crossplane/stack-{{ $arg }}:master"
+---
+{{- end }}
--- a/cluster/charts/crossplane/values.yaml.tmpl
+++ b/cluster/charts/crossplane/values.yaml.tmpl
@@ -13,7 +13,13 @@ args:
 imagePullSecrets:
 - dockerhub
 
-clusterStacks: []
-#- aws
-#- gcp
-#- azure
+clusterStacks:
+  aws:
+    version: v0.1.0
+    deploy: false
+  gcp:
+    version: v0.1.0
+    deploy: false
+  azure:
+    version: v0.1.0
+    deploy: false
--- a/cluster/charts/crossplane/values.yaml.tmpl
+++ b/cluster/charts/crossplane/values.yaml.tmpl
@@ -12,3 +12,8 @@ args:
 
 imagePullSecrets:
 - dockerhub
+
+clusterStacks: []
+#- aws
+#- gcp
+#- azure


cluster/charts/crossplane/templates/clusterstackinstalls.yaml
```diff
@@ -0,0 +1,9 @@
+{{- range $arg := .Values.clusterStacks }}
+apiVersion: stacks.crossplane.io/v1alpha1
+kind: ClusterStackInstall
+metadata:
+  name: "stack-{{ $arg }}"
+spec:
+  package: "crossplane/stack-{{ $arg }}:master"
```
prasek: can we make this the stable version of a stack that was available at the time the main crossplane released was made? This will be consumed by GitLab ADO users from stable releases.

turkenh: I think we don't have `stable` tags and all infrastructure stacks have version `v0.1.0`. I'll set tags as `v0.1.0`.

turkenh:
Now using stable versions (v0.1.0 as default) and versions are configurable. 

However, I just realized that our crd definitions do not have helm `crd-install` hook which results failing with helm 2. I think, crds were somehow installed on my cluster in my previous tests and didn't catch this issue. 
I'll try to figure out adding this annotation to crds in an appropriate way to make this work with helm 2.

jbw976:
@turkenh can you expand a little more on what the failure is and why it happens with helm 2? I hadn't seen any CRD installs being missed before and I'd like to better understand what is different for this scenario that is causing that to happen.  Maybe we can find an easy workaround.  Thanks!

turkenh:
@jbw976 sure. 

The failure happens when installation of infra stacks are enabled which is introduced with this PR (e.g. `--set clusterStacks.aws.deploy=true`). This flags deploys a `ClusterStackInstall` resource instance with appropriate spec. But since CRD of `ClusterStackInstall` is also part of the same chart, helm install fails as follows:

```bash
Error: validation failed: unable to recognize "": no matches for kind "ClusterStackInstall" in version "stacks.crossplane.io/v1alpha1"
```

AFAIK, helm 3 has some improvements with this (like dedicated CRD directory) but with helm 2, it is suggested to use either a separate chart for CRDs or add `crd-install` hook to the CRD definition: https://helm.sh/docs/chart_best_practices/#custom-resource-definitions

Since it looks like kubebuilder does not support adding such an annotation via kubebuilder annotations, we can consider adding an additional makefile step which adds this annotation after generating CRDs.

One other workaround, just coming to my mind could be adding helm's `post-install` hook to `ClusterStackInstall` instances that we want to create. I need to make sure that this does not break something with upgrades or reinstalls.

turkenh:
> One other workaround, just coming to my mind could be adding helm's post-install hook to ClusterStackInstall instances that we want to create. I need to make sure that this does not break something with upgrades or reinstalls.

This workaround is working, so pushed a new commit with this.

Currently, it is possible to install infra stacks with helm 2 but not possible with helm 3 :/

```
helm3 install crossplane-test --namespace crossplane-system -f values.yaml . --set clusterStacks.aws.deploy=true
Error: apiVersion "stacks.crossplane.io/v1alpha1" in crossplane/templates/clusterstackinstalls.yaml is not available
```

We need to figure out a way to support CRD installation together with CR instance with both version.

hr1sh1kesh:
@turkenh i think with the helm `post-install` hook there still is a small window where the `ClusterStackInstall` CRD might run before the stackmanager is ready unless we use the `--wait` flag. I am just catching up on xp and if i understand correctly the Stackmanager reconciles the state, but just wanted to bring this to notice.  Ref: https://github.com/helm/helm/blob/master/docs/charts_hooks.md#hooks-and-the-release-lifecycle

turkenh:
@hr1sh1kesh thanks for the notice.
The reason for `post-install` hook actually is not to wait until stackmanager ready, rather to wait until CRDs are deployed and registered to api-server. If we don't have it, helm validation simply fails with some error indicating that the api-server does not recognize this type:
```
... no matches for kind "ClusterStackInstall" in version "stacks.crossplane.io/v1alpha1"
```
I understand that you mean the problem described in #897. I believe the root cause of that issue has been resolved on master and expect not to see that problem again.

turkenh:
@prasek @jbw976 I suggest to merge this PR now. 

What we have now:
- Both `helm 2` and `helm 3` works as before if clusterStacks not enabled (no regression for current usage)
- When cluster stacks enabled, it works with `helm 2` but fails with `helm 3`

I'll go ahead and create an issue to track `helm 3 + cluster stacks enabled` case.

cluster/charts/crossplane/values.yaml.tmpl
```diff
@@ -12,3 +12,14 @@ args:
...
+
+clusterStacks:
+  aws:
+    version: v0.1.0
+    deploy: false
+  gcp:
+    version: v0.1.0
+    deploy: false
+  azure:
+    version: v0.1.0
```
prasek: @jbw976 are these versions going to bump to v0.2.0 in conjunction with Crossplane v0.4.0?

jbw976: yes, these should get bumped to v0.2.0 along with Crossplane v0.4.0.

jbw976: note however that these are just the default values, and that whoever is invoking Helm can put whatever values they want in.  We could leave this as is and GitLab when they install the Crossplane helm chart can specify v0.2.0 instead of this default.

I think we'll change the default to v0.2.0 when those ship anyways, but just wanted to be clear these are only the defaults.

jbw976:
Anything else needed for you to change your review to approved @prasek?

cluster/charts/crossplane/values.yaml.tmpl
prasek: that's super helpful, thanks @jbw976 ... also think that will unblock @turkenh since they can start passing `master` in the value instead of relying on the default version.

Issue #435, opened 2019-05-07
Package versioning and upgrading design


jbw976:
Following the [initial design for Stacks](https://github.com/crossplaneio/crossplane/issues/397), more thought is needed around supporting the updating of Stacks as well as the upgrading of Stacks.

This issue is focused on the Stacks ecosystem and Stacks usability.  Specific Crossplane claim types (aside from StackInstall) are not the focus of this issue.  The interoperability of all Crossplane claim types as they change versions is a concern for this design.

The format and changes in format for Stacks and StackInstall resources are also left to later discussion, as this is a broader Crossplane concerns.

Some discussion: https://github.com/crossplaneio/crossplane/pull/397#discussion_r276371864

Some thoughts and issues to consider in this design:
* #937 Do we want to support the concept of subscriptions and channels that can be subscribed to?
  * this is starting to make a lot of sense in terms of upgrades for new versions of the Stack that are released
* How do we deal with upgrades? Manually started by user? automatic in the case of a subscription? Dependency resolution across new versions of Stacks that are released at different times?
* #1051 Changes in version could affect multiple instances of the same StackInstall / Stack in different namespaces
* Versioning of CRDs, is all the necessary support in upstream Kubernetes to add new versions of CRDs and run them all side-by-side? We may need to research/prototype/confirm that our scenarios are supported upstream.
  * #1052 Reject instances of a StackInstall that conflict with installed StackInstalls
  * #1014 What problems existing with multiple versions of the same namespaced stack?
  * #1013 Merging or using the best CRDs (inclusive of the most versions) to fulfill StackInstall resources
  * #1152 conversion web-hooks
* #1042 Watch child artifacts for updates and re-sync / heal resources (e.g. deployments)

Part of #879

negz:
I believe this is fixed (i.e. designed, not implemented) by #1616.

PR #937, opened 2019-10-16
Stacks version design

lukeweber:

Stacks Installer will need to be version aware in the future. This doc is a first step at exploring the use of semantic versions and potential use cases for release channels and how we might attempt to define them.

Part of #435 

### Checklist

I have:

### DIFF
--- a/design/one-pager-stacks-versioning.md
+++ b/design/one-pager-stacks-versioning.md
@@ -1,0 +1,243 @@
+# Stack Versioning
+* Owner: Luke Weber (@lukeweber)
+* Reviewers: Crossplane Maintainers
+* Status: Draft, revision 1.0
+
+## Problem
+
+Stacks installer currently relies on installing a particular version of stack from a docker tag. The stack installer in
+ the future will be able to support stack update policy, maintenance windows, warnings on major updates, and more.
+ To build more complex features around stack version management we will need a common way to express versions that is
+ widely understood and standards based, so we will unify on best practices and
+ [semantic version 2.0](https://semver.org/spec/v2.0.0.html) as the Crossplane versioning standard.
+
+#### Didn't docker tags already solve this?
+
+ * I don't want my versions underlying artifacts to change:
+    * Tags are a symbolic reference to an actual sha of an image and they can be updated or deleted
+ * I want to have context from a version:
+    * A sha is hard to compare when debugging
+      * Which is newer? - cb44dd0 or d2fb617?
+      * Is this alpha, beta?
+    * A tag like latest is opaque to the end user and so not incredibly useful:
+      * Which version am I running? you need to run special commands to figure out which sha is loaded
+      * on a node and then lookup that sha to see which version it's tagged against, if it's even still tagged.
+      * If the latest tag on this node the same as the latest tag on that node? When did it get restarted?
+
+As a note, if a user wants to continue to install from an arbitrary tag, stack-installer will work, but behavior relying
+ on detecting the tags underlying sha changing will not be as the Crossplane stack installer by convention will assume
+ the immutability of tags.
+
+#### Why standardize on semantic versions?
+
+Semantic versions are a well adopted standard that give you context just from looking at the version. Based on semantic
+ versions we could implement a variety of useful features, that we couldn't reasonably or reliably do if everyone is
+ using a unique versioning system.
+
+* Auto-upgrade a major, minor, patch version based on policy.
+* Detect if an upgrade is available based only on the tag and the latest version in a channel, no introspecting of a
+  docker package.
+* Warning if the next latest version is a major upgrade.
+* Compare versions running easily based on status of stack install.
+* Tie human readable versions to change logs.
+* I can remember a version, it's very hard to remember a sha.
+* Semantic versions used by helm and widely adopted in kubernetes community.
+
+## Design
+
+We will rely on the established standard of [semantic version 2.0](https://semver.org/spec/v2.0.0.html) embedding these
+ versions inside of the stack app.yaml metadata, and inside of the tag in a docker image.
+
+The stack app.yaml version should always be a simple stable semantic version with no pre-release metadata, and tags
+ must exist that match the semantic version in the release channel.
+
+Creation dates will be tracked in the release channel versions for auditing purposes and will be iso-8601 format.
+ This is the same format as kubernetes and is a well established standard. We will also include a sha id of the
+ underlying stack container, which is immutable.
+
+The stack install CRD will get a new optional field for channel in the spec, that will include the url to the channel.
+
+There will be two static lists for channel information, namely the latest available in a channel, as well as the list of
+ all available versions for the stack installer to manage more complicated install scenarios based on upgrade policies
+ that will be defined in the future.
+
+### Considerations
+
+Our underlying stack storage is a docker registry, which can tag versions of stacks in any arbitrary way, but tags must
+ point to a single sha of an actual docker image, so there is no historical context, i.e. previous stable versions.
+ Further, if you were to use a :beta tag as a stand in for a beta channel, you would have little recourse to know how to
+ downgrade to the previous beta version in the event of a problem, nor would you even know with absolute certainty which
+ version you were running by looking at a spec or status in a kubernetes resource.
+
+### Versioning
+
+Semantic versions are widely respected as a standard to version software and are well
+ understood in the software community. Helm charts rely on semantic versions.
+
+```
+Given a version number MAJOR.MINOR.PATCH, increment the:
+
+MAJOR version when you make incompatible API changes,
+MINOR version when you add functionality in a backwards compatible manner, and
+PATCH version when you make backwards compatible bug fixes.
+Additional labels for pre-release and build metadata are available as extensions to the MAJOR.MINOR.PATCH format.
+```
+
+Additional Info: [Semver Reading](https://semver.org/)
+
+Semantic versions for a stack will be managed directly in the channel metadata, and will also be represented in docker
+ tags that point to the same underlying images. To avoid confusion, by convention we should never replace a semantic
+ version tag of software with a different underlying artifact. We should roll forward with a PATCH version. In the event
+ of a security patch we should un-publish the previous stack version when necessary, ideally after publishing a new
+ higher version.
+
+### Major Versions
+
+We can assume that major versions like MySQL 5 vs 8 series, might be better served as separate stacks, but the stack
+ system will not place arbitrary restrictions on pushing a major update and assume the author of the stack understands
+ the implications of a major update and can handle it in the stack upgrade logic.
+
+Having update policies should help mitigate major version upgrade problems in the future.
+
+### Enforcement of Version Upgrades
+
+Versions that need to be replaced will be replaced on a roll-forward basis with a new version. We will allow
+ arbitrary version updates that are less than latest to support patch versions to different minor versions of a stack
+ in the event that there are multiple supported minor versions by the software vendor.
+
+Rules for valid version:
+* MAJOR.MINOR.PATCH version in the app.yaml matches the MAJOR.MINOR.PATCH part from the tag ignoring build info in tag.
+* MAJOR.MINOR.PATCH tag must not already be a released stable version else promoting this version makes no sense and
+ would conflict.
+* By convention we can not modify an existing semantic version tag
+
+### Modes for install (Implicit)
+
+Pinned Install: The user will include a fully qualified package name and a specific tag to install. In this mode we will
+ assume an external versioning system is in use and not attempt to check for updates.
+
+Channel Install: If there is a channel set and there is no tag on the package, we will install the latest from the
+ channel.
+
+### Hosted artifacts
+
+It is assumed that we will have two hosted artifacts at a channel url, i.e.
+
+Channel url:
+https://s3.amazon.com/mybucket/mychannel
+
+Which would contain:
+https://s3.amazon.com/mybucket/mychannel/latest # Latest info
+https://s3.amazon.com/mybucket/mychannel/all # All version info
+
+The stack installer would initially only use https://s3.amazon.com/mybucket/mychannel/latest, but in the future if
+ the user set an upgrade policy it would likely read from all if necessary depending on rules.
+
+### Installing from stable
+
+The stack installer will be version aware, namely it will understand either how to upgrade to the tip of a release
+ channel, or how to follow a pre-defined upgrade policy by reading the entire list of available versions.
+
+Consider the following scenario where the installer wishes to auto-upgrade to the lastest version. If the current
+ version is 1.2.3 this would be reflected in the stack install status and the package would be set to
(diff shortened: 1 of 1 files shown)


design/one-pager-stacks-versioning.md
```diff
@@ -0,0 +1,248 @@
+# Stack Versioning
+* Owner: Luke Weber (@lukeweber)
+* Reviewers: Crossplane Maintainers
+* Status: Draft, revision 1.0
+
+## Problem
+
+Stacks installer currently relies on installing a particular version of stack from a docker tag. The stack installer in
```
suskin: > a particular version of a stack

Is this referring to the version of the stack which is being installed, or is this referring to the version of the Crossplane Stack Schema which the stack adheres to?

design/one-pager-stacks-versioning.md
```diff
@@ -0,0 +1,248 @@
...
+#### Didn't docker tags already solve this?
+
+ * I don't want my versions underlying artifacts to change:
+    * Tags are a symbolic reference to an actual sha of an image and they can be updated or deleted
+ * I want to have context from a version:
+    * A sha is hard to compare when debugging
+      * Which is newer? - cb44dd0 or d2fb617?
+      * Is this alpha, beta?
+    * A tag like latest is opaque to the end user and so not incredibly useful:
+      * Which version am I running? you need to run special commands to figure out which sha is loaded
```
suskin: Word to this

design/one-pager-stacks-versioning.md
```diff
@@ -0,0 +1,248 @@
...
+      * on a node and then lookup that sha to see which version it's tagged against, if it's even still tagged.
+      * If the latest tag on this node the same as the latest tag on that node? When did it get restarted?
+
+As a note, if a user wants to continue to install from an arbitrary tag, stack-installer will work as expected, but
+ behavior relying on re-tagging latest and the stack installer detecting the tags underlying container being updated
+ will not trigger a re-install as the Crossplane stack installer by convention will assume that docker tags are
```
suskin: > the Crossplane stack installer by convention will assume

Is this a recommendation, or is it the common convention when it comes to docker tags? Or is it an implementation detail that we can control?

design/one-pager-stacks-versioning.md
```diff
@@ -0,0 +1,248 @@
...
+ immutable and represent a single version. Any system that relies on tagging all builds with a :latest tag for example
+ is not recommended and in practice won't work as expected.
```
suskin: > is not recommended and in practice won't work

Is this recommendation coming from a document or article thing that I can read if I want to know more about the background?

design/one-pager-stacks-versioning.md
```diff
@@ -0,0 +1,248 @@
...
+
+#### Why standardize on semantic versions?
```
suskin: I like that there are FAQs in this document

design/one-pager-stacks-versioning.md
```diff
@@ -0,0 +1,248 @@
...
+
+ * I don't want my versions underlying artifacts to change:
+    * Tags are a symbolic reference to an actual sha of an image and they can be updated or deleted
+ * I want to have context from a version:
+    * A sha is hard to compare when debugging
+      * Which is newer? - cb44dd0 or d2fb617?
+      * Is this alpha, beta?
+    * A tag like latest is opaque to the end user and so not incredibly useful:
+      * Which version am I running? you need to run special commands to figure out which sha is loaded
+      * on a node and then lookup that sha to see which version it's tagged against, if it's even still tagged.
```
suskin: Is this bullet point intended to be part of the previous bullet point?

design/one-pager-stacks-versioning.md
```diff
@@ -0,0 +1,248 @@
...
+
+#### Didn't docker tags already solve this?
+
+ * I don't want my versions underlying artifacts to change:
+    * Tags are a symbolic reference to an actual sha of an image and they can be updated or deleted
+ * I want to have context from a version:
+    * A sha is hard to compare when debugging
+      * Which is newer? - cb44dd0 or d2fb617?
+      * Is this alpha, beta?
+    * A tag like latest is opaque to the end user and so not incredibly useful:
```
suskin: Since stacks are in general built from a repo, what are our thoughts on where the gitref of the source comes into play? It seems like it may be useful for whenever someone (probably a developer) is trying to debug a problem by pulling up the source code.

design/one-pager-stacks-versioning.md
```diff
@@ -0,0 +1,248 @@
...
+* Tie human readable versions to change logs.
+* I can remember a version, it's very hard to remember a sha.
+* Semantic versions used by helm and widely adopted in kubernetes community.
+
+## Design
+
+We will rely on the established standard of [semantic version 2.0](https://semver.org/spec/v2.0.0.html) embedding these
+ versions inside of the stack app.yaml metadata, and inside of the tag in a docker image.
+
+The stack app.yaml version should always be a simple stable semantic version with no pre-release metadata, and tags
```
suskin: > pre-release metadata

What is "pre-release metadata" in this context?

design/one-pager-stacks-versioning.md
```diff
@@ -0,0 +1,248 @@
...
+ must exist that match the semantic version in the release channel.
+
+Creation dates will be tracked in the release channel versions for auditing purposes and will be iso-8601 format.
```
suskin: Is there a document about release channels that we could link interested readers to, in case they want to read more about the release channel design?

design/one-pager-stacks-versioning.md
```diff
@@ -0,0 +1,248 @@
...
+ downgrade to the previous beta version in the event of a problem, nor would you even know with absolute certainty which
+ version you were running by looking at a spec or status in a kubernetes resource.
+
+### Versioning
+
+Semantic versions are widely respected as a standard to version software and are well
+ understood in the software community. Helm charts rely on semantic versions, and many others. Please see examples of
+ versioning systems linked at the bottom in background research on versioning.
+
+```
```
suskin: Nitpick: this seems more like a [block quote](https://daringfireball.net/projects/markdown/syntax#blockquote) than a code block

design/one-pager-stacks-versioning.md
```diff
@@ -0,0 +1,248 @@
...
+
+The stack installer would initially only use https://s3.amazon.com/mybucket/mychannel/latest, but in the future if
+ the user set an upgrade policy it would likely read from all if necessary depending on rules.
+
+### Installing from stable
+
+The stack installer will be version aware, namely it will understand either how to upgrade to the tip of a release
+ channel, or how to follow a pre-defined upgrade policy by reading the entire list of available versions.
+
+Consider the following scenario where the installer wishes to auto-upgrade to the lastest version. If the current
```
suskin: It seems like `lastest` should be `latest`

design/one-pager-stacks-versioning.md
```diff
@@ -0,0 +1,248 @@
...
+
+The stack app.yaml version should always be a simple stable semantic version with no pre-release metadata, and tags
+ must exist that match the semantic version in the release channel.
+
+Creation dates will be tracked in the release channel versions for auditing purposes and will be iso-8601 format.
+ This is the same format as kubernetes and is a well established standard. We will also include a sha id of the
+ underlying stack container, which is immutable, and will be used by the installer to fetch the exact artifact
+ referenced in the release channel.
+
+The stack install CRD will get a new optional field for channel in the spec, that will include the url to the channel.
```
suskin: What is a release channel? What's the mechanism the design proposes for implementing a Thing Which Is A Release Channel? I am noticing some references to release channel URLs and I see the description of an s3 bucket later in the document, but I am imagining it may make sense to create a `### Release Channels` section which explains the structure of a channel and its API contract, to help put these other descriptions in context.

design/one-pager-stacks-versioning.md
```diff
@@ -0,0 +1,248 @@
...
+
+### Hosted artifacts
+
+It is assumed that we will have two hosted artifacts at a channel url, i.e.
+
+Channel url:
+https://s3.amazon.com/mybucket/mychannel
+
+Which would contain:
+https://s3.amazon.com/mybucket/mychannel/latest # Latest info
```
suskin: It feels like these lines are intended to be in a code block, since they are using `#` for comments

design/one-pager-stacks-versioning.md
```diff
@@ -0,0 +1,248 @@
...
+  {
+    "version": "1.2.3",
+    "id": "461324714c7d",
+    "createTime": "2019-09-12T17:39:04Z"
+  }
+}
+```
+
+To enact an upgrade in this scenario, you would update the payload to be a higher semantic version, i.e
+ `{"latest": {"version": 1.2.4", ...} ... }`. The installer would then install a new stack image from
```
suskin: There is a pattern in Kubernetes of writing annotations on objects when they're changed; I'm wondering whether it would make sense for the installer to record information about the upgrade in an annotation on the stack when it does one (things like from version, to version, channel, timestamp).

design/one-pager-stacks-versioning.md
```diff
@@ -0,0 +1,248 @@
...
+ `crossplane/aws-stack:shaofversion` by looking up the actual underlying artifact for that version in the channel.
+ The stack installer status should reflect that is installing 1.2.4, with the particular sha.
+
+If you update your release channel to something of a lesser value the stack installer would return an error that release
+ channel is of a lesser version and not upgrade. Further, if we pushed a version that contained pre-release info, this
```
suskin: > the stack installer would . . . not upgrade

What is the reason we want this behavior, instead of an auto-downgrade?

design/one-pager-stacks-versioning.md
```diff
@@ -0,0 +1,248 @@
...
+ would not match as a stable version and would also fail to update. Ideally we would disallow pushing invalid versions
```
suskin: > and would also fail to update

Semi nitpick: What would be the recommended graceful degradation or workaround in this case? It would be nice if someone could install something for the first time even if `stable` appeared to be misconfigured.

design/one-pager-stacks-versioning.md
```diff
@@ -0,0 +1,248 @@
...
+ to channels as part of our internal tooling.
+
+### Arbitrary channel support
+
+Stable is a reserved channel name to indicate clean semantic versions with no pre-release metadata. Any arbitrary
```
suskin: Can anyone spin up their own release channel on an arbitrary url?

design/one-pager-stacks-versioning.md
```diff
@@ -0,0 +1,248 @@
...
+
+## Problem
+
+Stacks installer currently relies on installing a particular version of stack from a docker tag. The stack installer in
+ the future will be able to support stack update policy, maintenance windows, warnings on major updates, and more.
+ To build more complex features around stack version management we will need a common way to express versions that is
+ widely understood and standards based, so we will unify on best practices and
+ [semantic version 2.0](https://semver.org/spec/v2.0.0.html) as the Crossplane versioning standard.
+
+#### Didn't docker tags already solve this?
```
suskin: Sorry if this was already discussed somewhere, but have we already considered a Docker registry-based implementation of the described release channel mechanism? I am imagining conventions and patterns which we would implement on top of the existing tagging functionality. For example:

* A registry's `/tags/list` [is very similar](https://docs.docker.com/registry/spec/api/#listing-image-tags) to the functionality of the `/all` url on a release channel
* Getting an individual tag from the registry (such as `stable`) is very similar to the `/latest` url on a release channel
* For historical lookups, the individual image shas which become the "head" of a release channel could be tagged with some convention for historical purposes, such as `beta-2019-10-10-0.0.1` and `beta-2019-10-17-0.0.2`

design/one-pager-stacks-versioning.md
```diff
@@ -0,0 +1,248 @@
...
+   }
+ }
+ ```
+
+### Support for complex upgrades with a list of versions + update policy
+
+Future work could further build on the latest concept in our release channels and instead define update policy.
+ Instead of forcing the stack to only update to the next version, we will support maintenance and upgrade policies.
+
+An example: Imaging the following upgrade policy - Apply patch versions immediately, minor versions in a pre-defined
```
suskin: What are the considerations for pull-driven release channel upgrades versus push-driven release channel upgrades? Considerations such as scaling request volume, and servicing control clusters which are not accessible to be pushed to. Are we assuming pull-driven here?

design/one-pager-stacks-versioning.md
displague:
```console
kubectl get pod -l app=crossplane -n crossplane-system \
  -o jsonpath='{.items[0].status.containerStatuses[0].imageID}'
```

design/one-pager-stacks-versioning.md
displague: The bullet there seems to be indent intended.

> then lookup that sha to see which version it's tagged against, if it's even still tagged.

For those following along at home.. 
```console
$ IMAGEID=$(kubectl get pod -l app=crossplane -n crossplane-system \
  -o jsonpath='{.items[0].status.containerStatuses[0].imageID}')
$ docker inspect $IMAGEID -f '{{.RepoTags}}'
[build-ad66fa5c/crossplane-amd64:latest crossplane/crossplane:master crossplane/crossplane:v0.3.0-233.g98e8520e]
```

design/one-pager-stacks-versioning.md
displague: I think there is an expectation that Stacks would use a `pullPolicy` like other image references in Kubernetes.  We haven't supported this in the past and I like that we are declaring that we are not going to.

This makes sense for Stacks because Stacks do not get scaled, destroyed and recreated, as Pods do.  The Stack Manager expects to install a particular version precisely one time and if that version of the Stack image does not change, the SM will do nothing else.

design/one-pager-stacks-versioning.md
```diff
@@ -0,0 +1,248 @@
...
+Semantic versions are a well adopted standard that give you context just from looking at the version. Based on semantic
+ versions we could implement a variety of useful features, that we couldn't reasonably or reliably do if everyone is
+ using a unique versioning system.
+
+* Auto-upgrade a major, minor, patch version based on policy.
+* Detect if an upgrade is available based only on the tag and the latest version in a channel, no introspecting of a
+  docker package.
+* Warning if the next latest version is a major upgrade.
+* Compare versions running easily based on status of stack install.
+* Tie human readable versions to change logs.
```
displague: I was just wondering if changelogs would be in scope for this design.  Should changelogs be required and should their format and representation as Stack metadata be explored here? Probably not.

design/one-pager-stacks-versioning.md
displague: I think this is stipulates that `app.yaml:version` should never include `-alpha`, `-beta`, `-rc` suffixes.

This implies, to me, that either images tagged in the registry are also not allowed to contain these suffixes, or that the the _match_ check between registry images and app.yaml must be softened to ignore omitted suffixes in app.yaml.

PR #956, opened 2019-10-21
design: add design doc for template stacks experience

suskin:
## Overview

As part of creating template stacks, we want to explore what the
experience looks like at the beginning, and then work backwards from
there. This helps us focus on the user experience early, and helps us
iron out **some** kinks before we even begin. This design doc and the
code which it links to are an attempt at that.

This document is intended to be consumed alongside these code examples: https://github.com/suskin/template-stack-experience

[skip ci]

### DIFF
--- a/design/design-doc-template-stacks-experience.md
+++ b/design/design-doc-template-stacks-experience.md
@@ -185,16 +185,21 @@ At a high level, the steps would be as follows:
 2. In the project directory, `kubectl crossplane stack init --template
    myorg/mystack` to create the boilerplate of the stack layout.
 3. Relative to the project directory, put the yamls in
-   `config/stack/manifests/resources/templates` .
+   `config/stack/manifests/resources/templates`, or a different
+   directory if desired.
 4. Add CRD definitions; this could be done by putting CRD yaml
    definitions in `config/stack/manifests/resources` or in
-   `config/crd/bases`.
+   `config/crd/bases`. We plan to make this simpler for the user by
+   adding a crossplane-cli command for it.
 5. Add configuration to `stack.yaml` to
    specify how configurations are rendered.
 6. Edit `config/stack/manifests/resources/app.yaml` to specify that the
    stack will be working with all of the kinds that it will be working
    with (using the `dependsOn` field).
 7. Edit `config/stack/manifests/resources/app.yaml` as appropriate.
+
+In the future, we expect that we will combine `app.yaml` and
+`stack.yaml`.
 
 #### Adding a CRD
 
@@ -331,7 +336,7 @@ behaviors:
   crds:
     # This is a particular CRD which is being configured. When the
     # controller sees an object of this type, it will do something.
-    wordpressinstance.wordpress.samples.stacks.crossplane.io/v1alpha1:
+    wordpressinstance.wordpress.samples.stacks.crossplane.io:
       # These are the templates which should be rendered when an object
       # of the type above is seen.
       #
@@ -339,15 +344,18 @@ behaviors:
       # for the CRD fields, using the standard Kubernetes mechanism for
       # specifying CRD field default values.
       resources:
-        # These are individual files which will be rendered.
-        - templates/kubernetescluster.yaml
-        - templates/mysqlinstance.yaml
-        - tempaltes/kubernetesapplication.yaml
+        # Each list item is a folder of files which will be rendered.
+        - wordpress
 EOF
 ```
 
 The `stack.yaml` should be in the build root of the repository. For most
 single-stack repositories, this means the root of the repository.
+
+For some realistic samples, see the [helm engine
+example][helm-engine-example] or the [quick start
+example][quick-start-example] in the template stack experience
+repository.
 
 ### Specifying default values
 
@@ -424,9 +432,10 @@ implementation](https://github.com/crossplaneio/crossplane/blob/master/design/on
 See the [Template Stack
 Experience](https://github.com/suskin/template-stack-experience) code
 examples for some realistic examples of how different use-cases work
-with a realistic workload.
+with a realistic workload. For a coherent user scenario, see the [quick
+start example](https://github.com/suskin/template-stack-experience/tree/master/wordpress-workload/quick-start).
 
-Examples covered by the repository all currently use go templates as the
+Examples covered by the repository mostly use go templates as the
 templating engine. The following cases are covered by the examples:
 
 * No variables at all in a set of configurations.
@@ -437,11 +446,13 @@ templating engine. The following cases are covered by the examples:
 * Multiple CRDs defined in a single Stack.
 * Multiple rendered variants of a given configuration.
 * Composing configuration using multiple Stacks.
+* A user scenario of setting up cloud provider resources and deploying
+  an application bundled as a stack.
 
 ### Helm charts
 
-To author a template stack from a simple helm chart, these steps could
-be followed:
+To author a [template stack from a simple helm
+chart][helm-engine-example], these steps could be followed:
 
 1. Create and initialize a template stack project, with a helm flag:
    `kubectl crossplane stack init --template`.
@@ -459,11 +470,15 @@ When consuming the stack, the default template values can be overridden
 by specifying fields on the render request object.
 
 This use-case will probably need additional thought if we want to
-support all permutations of helm chart.
+support all permutations of helm chart. For more realistic and complete
+examples, see the [helm engine example][helm-engine-example] and the
+[quick start example][quick-start-example] in the template stack
+experience repository.
 
 ## Further reading
 
 * [Template Stack Experience](https://github.com/suskin/template-stack-experience) examples
+* [Template Stack Experience - Quick Start Example](https://github.com/suskin/template-stack-experience/tree/master/wordpress-workload/quick-start)
 * [Declarative Application Management in Kubernetes](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/architecture/declarative-application-management.md)
 * [Template Stacks internals design doc](https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-template-stacks.md)
 * [Stacks CLI design](https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-stack-cli.md)
@@ -471,3 +486,5 @@ support all permutations of helm chart.
 
 <!-- Reference-style links -->
 [kubernetes-default-values]: https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#defaulting
+[helm-engine-example]: https://github.com/suskin/template-stack-experience/tree/master/wordpress-workload/go-templates/helm-engine
+[quick-start-example]: https://github.com/suskin/template-stack-experience/tree/master/wordpress-workload/quick-start
--- a/design/design-doc-template-stacks-experience.md
+++ b/design/design-doc-template-stacks-experience.md
@@ -350,9 +350,8 @@ behaviors:
 The `stack.yaml` should be in the build root of the repository. For most
 single-stack repositories, this means the root of the repository.
 
-For some realistic samples, see the [helm engine
-example][helm-engine-example] or the [quick start
-example][quick-start-example] in the template stack experience
+For a realistic sample for a complete user scenario, see the [quick
+start example][quick-start-example] in the template stack experience
 repository.
 
 ### Specifying default values
@@ -398,28 +397,67 @@ For the full example, see the sample in the
 
 The imagined implementation for how objects sent to the stack for
 rendering become resources is that the object's fields become the input
-for the template rendered by the controller. For more details, see the
-[detailed examples](https://github.com/suskin/template-stack-experience)
-or the [design document about the
+for the template rendered by the controller.
+
+For example, given this object as input:
+
+```yaml
+apiVersion: wordpress.samples.stacks.crossplane.io/v1alpha1
+kind: WordpressInstance
+metadata:
+  name: "my-wordpress-app-from-helm2"
+  namespace: dev
+spec:
+  engineVersion: "8.0"
+```
+
+The `engineVersion` field would be available to the templating engine
+which was being used, and would have a value of `8.0`. This means that
+if the engine being used were helm, the behavior would be equivalent to
+if there were a `values.yaml` passed in which looked like this:
+
+```yaml
+engineVersion: "8.0"
+```
+
+For the complete example, see the [quick start
+example][quick-start-example] in the template stack experience
+repository. The snippet was taken from the part of the example where the
+app stack is used.
+
+For more details, see the [design document about the
 internals](https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-template-stacks.md).
 
 ### Templating/configuration engine
 
 Template stacks will not be opinionated about which templating engine is
 used. We plan to support multiple configuration engines. The engine will
-be configurable by setting values in the `stack.yaml`.
+be configurable by setting values in the `stack.yaml`. Here's an example
+of a snippet from a `stack.yaml` which specifies a particular engine:
 
-For some realistic examples of what a stack would look like when it uses
-different engines, see the [quick start example][quick-start-example].
-There are also some more details in the [stack yaml
-section](#the-stackyaml), and in the [helm charts
-section](#helm-charts).
+```yaml
+behaviors:
+  engine:
+    type: kustomize
+```
+
+In some cases, the engine configuration line may be optional; the system
+may be able to infer the engine based on the structure of the stack. For
+example, if no engine is specified, and a `kustomization.yaml` is found,
+the engine could be inferred to be kustomize.
+
+For a more complete example of a user scenario, including usage of
+multiple different engines, see the [quick start
+example][quick-start-example]. There are also some more details in the
+[stack yaml section of this document](#the-stackyaml), and in the [helm
+charts section](#helm-charts).
 
 ### Lifecycle hooks
 
 We expect to eventually support lifecycle hooks. See the [speculative
 design in the template stacks experience repo](https://github.com/suskin/template-stack-experience/tree/master/wordpress-workload/go-templates/lifecycle-hook)
-for more details about what that could look like.
+for more details about what that could look like. Lifecycle hooks are
+out of the scope of this document, and will be revisited in the future.
 
 ### Internal representation of templates
 
@@ -433,25 +471,12 @@ implementation](https://github.com/crossplaneio/crossplane/blob/master/design/on
 
 ## Example use-cases
 
-See the [Template Stack
-Experience](https://github.com/suskin/template-stack-experience) code
-examples for some realistic examples of how different use-cases work
-with a realistic workload. For a coherent user scenario, see the [quick
-start example](https://github.com/suskin/template-stack-experience/tree/master/wordpress-workload/quick-start).
+For a coherent user scenario, see the [quick start
+example][quick-start-example].
 
-Examples covered by the repository mostly use go templates as the
-templating engine. The following cases are covered by the examples:
-
-* No variables at all in a set of configurations.
-* Variables in configurations with default variable values.
-* Variables in configurations, with values specified by users when a
-  render is requested.
-* Resource packs.
-* Multiple CRDs defined in a single Stack.
-* Multiple rendered variants of a given configuration.
-* Composing configuration using multiple Stacks.
-* A user scenario of setting up cloud provider resources and deploying
-  an application bundled as a stack.
+The scenario shows what it might look like for a user to set up an
+application and its infrastructure from scratch using Crossplane and
+Template Stacks. Multiple configuration engines are shown.
 
 ### Helm charts
 
@@ -461,7 +486,7 @@ chart][helm-engine-example], these steps could be followed:
 1. Create and initialize a template stack project, with a helm flag:
    `kubectl crossplane stack init --template`.
 2. Create a CRD: `kubectl crossplane stack crd init WordpressInstance
-   wordpressinstances wordpress.samples.stacks.crossplane.io`
+   wordpress.samples.stacks.crossplane.io`
 3. Put the chart's contents into the templates folder.
 4. Create a `stack.yaml` in the root of the repostory, and configure it
    to use the templates for the created CRD.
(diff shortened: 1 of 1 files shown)


suskin:
This has been updated based on some design discussion that happened outside of this PR, so it's ready for another review. Also, I'm removing Hasan as a reviewer as he is working on some other projects now.

design/design-doc-template-stacks-experience.md
```diff
@@ -0,0 +1,473 @@
...
+
+1. Create a project directory.
+2. In the project directory, `kubectl crossplane stack init --template
+   myorg/mystack` to create the boilerplate of the stack layout.
+3. Relative to the project directory, put the yamls in
+   `config/stack/manifests/resources/templates` .
+4. Add CRD definitions; this could be done by putting CRD yaml
+   definitions in `config/stack/manifests/resources` or in
+   `config/crd/bases`.
+5. Add configuration to `stack.yaml` to
```
displague: I'm curious about the path to stack.yaml. If the omitted path indicates that this is a topmost file, I think "app.yaml" deserves that position (and the name stack.yaml).

The definition of CRDs -> Templates is a lesser concern than the definition of the Stack itself (which app.yaml provides).

Would stack.yaml be more "at home" alongside the CRDs (with other Stack extending files) or perhaps alongside the template files?

Should stack.yaml be a single file for the entire Stack or could it be split up by CRD as other stack metadata can?

design/design-doc-template-stacks-experience.md
```diff
@@ -0,0 +1,473 @@
...
+      # of the type above is seen.
+      #
+      # Defaults for the variables can be defined in the default values
+      # for the CRD fields, using the standard Kubernetes mechanism for
+      # specifying CRD field default values.
+      resources:
+        # These are individual files which will be rendered.
+        - templates/kubernetescluster.yaml
+        - templates/mysqlinstance.yaml
+        - tempaltes/kubernetesapplication.yaml
```
displague: `templates` typo

Are you concerned that by not using a name mapping to template files that we will lose the ability to map template derived resource values within templates `{{ .templateName.resourceValue }}`?

I realize there are alternatives:
* `{{ template("template/path").resourceValue }}`
* `{{ resource("kind", "name").resourceValue }}` 
   _note: this would not create a GET API request, this would be rendered from an pre-rendering fetch of all the template's known resources_
* others?

I haven't personally pinned one down and I appreciate your investigation into this part of the design.

design/design-doc-template-stacks-experience.md
```diff
@@ -0,0 +1,473 @@
...
+EOF
+```
+
+The `stack.yaml` should be in the build root of the repository. For most
```
displague: This duplicates https://github.com/crossplaneio/crossplane/pull/956/files#diff-e81d9b5b6cbb8b4bdfc4d7a5ef0f2f25R313-R317 which was only a few lines earlier.

design/design-doc-template-stacks-experience.md
```diff
@@ -0,0 +1,473 @@
...
+7. Edit `config/stack/manifests/resources/app.yaml` as appropriate.
+
+#### Adding a CRD
+
+The tooling will make it simpler to add a CRD from scratch. For example,
+the following would create a basic CRD in the appropriate folder, and
+would update the `dependsOn` field of the stack's `app.yaml`:
+
+```
+kubectl crossplane stack crd init WordpressInstance wordpressinstances wordpress.samples.stacks.crossplane.io
```
displague: I believe this would create an `owns` relationship.  A dependsOn relationship does not result in the inclusion of a new physical CRD file in the Stack, merely a line item in the app.yaml (which results in RBAC rules permitting the Stack access to the resources it depends on).

design/design-doc-template-stacks-experience.md
```diff
@@ -0,0 +1,473 @@
...
+metadata:
+  name: thinginstance-sample
+spec:
+  myfield: myvalue
+EOF
+
+kubectl delete -f my-instance.yaml
+```
+
+The instance is also deleted if the stack is uninstalled.
```
displague: nice addition here.

design/design-doc-template-stacks-experience.md
```diff
@@ -0,0 +1,473 @@
...
+```
+kubectl crossplane stack uninstall mystack
+```
+
+#### Upgrade and update
+
+Upgrading a stack version, and updating an instance of an object managed
+by the stack, are out of scope of this document. However, one could
+imagine that the process of upgrading a stack version would change a
+version number on object instances which go with the stack, and that the
```
displague: Says they are out of scope and then sets up some interesting half-thoughts :horse_racing: 

> change a version number on object instances which go with the stack

Would this be spec, label, or annotation on the Foo resource? Or are you suggesting that the RenderedBar resources would carry the Stack version?

This may be helpful for Template Stacks to make version related changes to RenderedBar resources.

design/design-doc-template-stacks-experience.md
```diff
@@ -0,0 +1,473 @@
...
+version number change would underneath cause the template to be rendered
+again and applied to the cluster.
+
+### The stack.yaml
+
+Configuration for the template stack will be in a `stack.yaml` file at
```
displague: My views on this are here https://github.com/crossplaneio/crossplane/pull/956/files#r343679051.

design/design-doc-template-stacks-experience.md
```diff
@@ -0,0 +1,473 @@
...
+   to use the templates for the created CRD.
+5. To set up default values:
+    1. Put the contents of the chart's `values.yaml` into the CRD's
+       [default field value definitions][kubernetes-default-values].
+6. Bundle and publish the stack.
+
+When consuming the stack, the default template values can be overridden
+by specifying fields on the render request object.
+
+This use-case will probably need additional thought if we want to
```
displague: It might be worth mentioning some helm chart differences one may encounter when converting a helm  chart to fit this initial template stack engine (or we can add this after the implementation).

* includes may not initially be supported but are planned, file paths may need to change.
* template variables may need to be altered
* template functions may not align with helm

It may also be worth pointing out that a future template stacks engine could potentially consume helm charts with less manipulation.

design/design-doc-template-stacks-experience.md
```diff
@@ -0,0 +1,473 @@
...
+The imagined implementation for how objects sent to the stack for
+rendering become resources is that the object's fields become the input
+for the template rendered by the controller. For more details, see the
+[detailed examples](https://github.com/suskin/template-stack-experience)
+or the [design document about the
+internals](https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-template-stacks.md).
+
+### Templating/configuration engine
+
+Template stacks will not be opinionated about which templating engine is
```
jbw976: Can we go into a lot more detail here about the experience around using existing engines? If that is the general direction we are heading in, I'd love to see what it looks like for a stack author to basically reuse their existing helm chart inside the new stack they are building.  How do they specify what engine to use at render time? How do they include their existing assets/artifacts into the stack repo/package?

design/design-doc-template-stacks-experience.md
```diff
@@ -0,0 +1,473 @@
...
+* No variables at all in a set of configurations.
+* Variables in configurations with default variable values.
+* Variables in configurations, with values specified by users when a
+  render is requested.
+* Resource packs.
+* Multiple CRDs defined in a single Stack.
+* Multiple rendered variants of a given configuration.
+* Composing configuration using multiple Stacks.
+
+### Helm charts
```
jbw976: this is perhaps where we can get into more depth around how helm charts could be reused and what the experience looks like.  it would be **great** to see that in the https://github.com/suskin/template-stack-experience repo too, with basically a concrete example of what we expect a fully building repo that reuses a helm chart in a stack to look like.

design/design-doc-template-stacks-experience.md
jbw976: why would there be a conversion necessary of a helm chart to fit into a stack? if we are reusing the helm rendering engine under the covers, similar to how we'd reuse any existing engine for existing formats we want to support, why would there be any differences that a stack author would have to learn and deal with?

displague: > converting a helm chart to fit this initial template stack engine 

I have been expecting that the initial engine would not use helm directly under the covers, but would provide helm-like functionality through the use of Go templates and functions similar when not identical to those offered by helm.

If the first engine or any subsequent engine does employ Helm directly, I still expect there to be differences as helm runs locally against a local environment and filesystem and maintains some local state, while the Template Stacks experience of this would be fundamentally different.

Whether there are limitations (on customizations or functions available, versioning, channel upgrades and rollbacks), new features, caveats, conversion steps, or work-arounds for Helm users coming to Template Stacks, I would expect a brief explanation of the expected differences.  For example, fields outside of those defined in Helm's values.yaml can be updated, similar to how Kustomize can overlay values.  Will the Helm Template Stack Engine offer that capability?  Would an arbitrary yaml field be the interface? Is Tiller used and/or required? If not how does this change the Helm user experience?

jbw976: I'm definitely thinking that reusing existing engines (such as helm but there are others too) instead of writing our own implementation would be the fastest way to bootstrap this effort and provide the most functionality at the same time. Agreed that there could be differences that come up in reality and it would be good to understand them.

suskin: > why would there be a conversion necessary

It is very difficult to be confident _a priori_ that something that works in one situation will work in a slightly different situation where a different code path is being executed. In this case, we would be going from using `helm install` to using `helm template`, which is not the same code path and already has some known other differences:

* `helm template` does not use tiller
* From the helm template documentation, `any values that would normally be
looked up or retrieved in-cluster will be faked locally.`

So, while we may not know of any obviously necessary conversion steps yet, that doesn't mean there won't be any.

> Kustomize. Will the Helm Template Stack Engine offer that capability?

Based on my understanding, I believe using kustomize after helm still would require a kustomization file, but the short answer is that this should be possible. Since we're planning on mostly decoupling the processing engine from the stack itself, "helm to kustomize" could just be another engine that is supported.

design/design-doc-template-stacks-experience.md
suskin: I've added the links to other sections, and I've also linked out to the recently created [quick start example](https://github.com/suskin/template-stack-experience/tree/master/wordpress-workload/quick-start) in the experience sample repository to give readers somewhere to go to get more details about this.

design/design-doc-template-stacks-experience.md
suskin: I added a few links to this section to some more detailed samples, so give readers a place to see more details.

design/design-doc-template-stacks-experience.md
suskin: This is changed

jbw976:
@suskin is there ready to take another look at?  I see that the quick start in the template stack experience repo has been updated to remove the generic config sections from the StackInstall, so that's good: https://github.com/suskin/template-stack-experience/commit/ed1a4431dcd5081103b7f7305f2b272e7a10d49e

suskin:
> @suskin is there ready to take another look at? I see that the quick start in the template stack experience repo has been updated to remove the generic config sections from the StackInstall, so that's good: [suskin/template-stack-experience@ed1a443](https://github.com/suskin/template-stack-experience/commit/ed1a4431dcd5081103b7f7305f2b272e7a10d49e)

@jbw976 yep! It's ready. I took another look just to double check, and fixed some syntax that was stale, so it should be good to go.

jbw976:
Assigning to @suskin to respond to comments/feedback from @negz in https://github.com/crossplaneio/crossplane/pull/956#pullrequestreview-323263969

design/design-doc-template-stacks-experience.md
```diff
@@ -0,0 +1,505 @@
...
+   `config/crd/bases`. We plan to make this simpler for the user by
+   adding a crossplane-cli command for it.
+5. Add configuration to `stack.yaml` to
+   specify how configurations are rendered.
+6. Edit `config/stack/manifests/resources/app.yaml` to specify that the
+   stack will be working with all of the kinds that it will be working
+   with (using the `dependsOn` field).
+7. Edit `config/stack/manifests/resources/app.yaml` as appropriate.
+
+In the future, we expect that we will combine `app.yaml` and
```
jbw976: `stack.yaml` doesn't exist yet, and I actually thought `stack.yaml` was simply just `app.yaml` renamed and moved to the root of the repo or stack package filesystem.  Any reason not to just combine them now?

design/design-doc-template-stacks-experience.md
```diff
@@ -0,0 +1,505 @@
...
+
+There will also be a convenience flag in the `stack init` command so
+that people starting from scratch can initialize the stack and the first
+CRD with a single command:
+
+```
+$ kubectl crossplane stack init --template mygroup/mystackname --init-crd
+> CRD name: WordpressInstance
+> CRD plural: wordpressinstances
+> CRD subdomain: wordpress.samples.stacks.crossplane.io
```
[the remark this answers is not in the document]
jbw976: should version be specified here too?

design/design-doc-template-stacks-experience.md
```diff
@@ -0,0 +1,505 @@
...
+the build root of the stack in the repository. For most cases, this will
+mean the `stack.yaml` lives at the root of the repository.
+
+#### Specifying how to process render requests
+
+A Stack author must create a `stack.yaml` in order to configure how
+templates are rendered in response to a render request. Here's an
+example, with comments explaining the directives:
+
+```
```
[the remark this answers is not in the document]
jbw976: agreed, `cat <<EOF` isn't as useful here as using a fenced code block a la:
```
 ```yaml
```

design/design-doc-template-stacks-experience.md
```diff
@@ -0,0 +1,505 @@
...
+        postCreate:
+          # These are the templates which should be rendered when an object
+          # of the type above is seen.
+          #
+          # Defaults for the variables can be defined in the default values
+          # for the CRD fields, using the standard Kubernetes mechanism for
+          # specifying CRD field default values.
+          # Note that this is a list of objects, so multiple can be
+          # specified.
+          - directory: wordpress
```
jbw976: what other types of values can be used here besides `directory`?

design/design-doc-template-stacks-experience.md
```diff
@@ -0,0 +1,505 @@
...
+behaviors:
+  crds:
+    # This is a particular CRD which is being configured. When the
+    # controller sees an object of this type, it will do something.
+    wordpressinstance.wordpress.samples.stacks.crossplane.io:
+      # This is a top-level object to namespace hook configurations.
+      # There can be hooks for multiple different types of events.
+      hooks:
+        # Post create is triggered after an instance is created
+        postCreate:
```
jbw976: `post*` implies that this is the hook run after the `create` hook has been run.  I wonder if it will simplify the semantics of this hook to just simply call it `create`?  same with `postUpdate`.

design/design-doc-template-stacks-experience.md
jbw976: is specifying the "engine" needed, or can that be inferred by the template controller by looking at the given directory and examining it's structure? (i.e. this dir has a kustomization.yaml so I should render with kustomize)

design/design-doc-template-stacks-experience.md
```diff
@@ -0,0 +1,505 @@
...
+          # of the type above is seen.
+          #
+          # Defaults for the variables can be defined in the default values
+          # for the CRD fields, using the standard Kubernetes mechanism for
+          # specifying CRD field default values.
+          # Note that this is a list of objects, so multiple can be
+          # specified.
+          - directory: wordpress
+        # Post create is triggered after an instance is changed
+        postUpdate:
```
jbw976: for the simple case (and probably popular case) where you simply want the same set of template yamls to get rendered any time the CR is created or updated, do you need to still specify both the create and update hooks? is there a more simple way to express "every time something happens to the CR, just render this one directory for me"?

design/design-doc-template-stacks-experience.md
```diff
@@ -0,0 +1,505 @@
...
+          - directory: wordpress
+EOF
+```
+
+The `stack.yaml` should be in the build root of the repository. For most
+single-stack repositories, this means the root of the repository.
+
+For some realistic samples, see the [helm engine
+example][helm-engine-example] or the [quick start
+example][quick-start-example] in the template stack experience
```
jbw976: having this quick start example in another repo is sort of making reviewing this PR challenging, because you can't leave any comments on that content within the context of this PR.  Should the quickstart maybe be folded into this design doc for colocation purposes?

design/design-doc-template-stacks-experience.md
```diff
@@ -0,0 +1,505 @@
...
+
+For the full example, see the sample in the
+[template stack experience repository](https://github.com/suskin/template-stack-experience/tree/master/wordpress-workload/go-templates/default-variables).
+
+### How processing objects works under the covers
+
+The imagined implementation for how objects sent to the stack for
+rendering become resources is that the object's fields become the input
+for the template rendered by the controller. For more details, see the
+[detailed examples](https://github.com/suskin/template-stack-experience)
```
jbw976: this is another example where it hurts readability and cognitive processing of these concepts by having the examples in another repo and just linking to the root of it, putting the burden on the reader to navigate through that other repo and discover on their own the concepts being mentioned here.

i think this design would be easier to parse and understand if 1 example was put inline here that demonstrates a template consuming input from the CR.

design/design-doc-template-stacks-experience.md
```diff
@@ -0,0 +1,505 @@
...
+
+1. Create and initialize a template stack project, with a helm flag:
+   `kubectl crossplane stack init --template`.
+2. Create a CRD: `kubectl crossplane stack crd init WordpressInstance
+   wordpressinstances wordpress.samples.stacks.crossplane.io`
+3. Put the chart's contents into the templates folder.
+4. Create a `stack.yaml` in the root of the repostory, and configure it
+   to use the templates for the created CRD.
+5. To set up default values:
+    1. Put the contents of the chart's `values.yaml` into the CRD's
```
jbw976: echoing @negz feedback from earlier, I wonder what tooling and experience improvements we can make around setting defaults in the CRDs, especially when coming from an existing helm chart, in addition to what @negz mentioned of tooling to define the CRD's OpenAPI schema.

I wonder if there should be any investment in Helm chart migration/import automation.  e.g. `kubectl crossplane stack init --template --from-helm-chart=some-chart-dir`?

suskin: Yes, there should absolutely be an investment in improving the authoring and importing/converting stories. However, I believe we will develop a better idea of how we want to improve those as we build a working system and gain some experience with it. It seems out of scope for now.

If we are brainstorming, I imagine that a CRD, the start of a schema, some of the hook configuration, and defaults could all be generated from a helm chart.

design/design-doc-template-stacks-experience.md
suskin: Thanks for pointing this out! This was originally intended to be a bit hand-wavey; since we are talking about supporting multiple engines, it was unclear whether getting into specifics of how things are transformed made sense yet. Maybe it makes sense now though, since we have a "complete" example in the repo.

Hmm, this link should probably just link to the quick start now.

I'll look into whether there's a nice concise example that can be pasted in here or linked to directly. I'm thinking the helm version of the app stack in the quick start could work well.

design/design-doc-template-stacks-experience.md
```diff
@@ -0,0 +1,505 @@
...
+  this time.
+
+Also out of scope is the internal representation of configuration, and
+the implementation of the stack manager. See the [complementary
+internals-oriented Template Stack design doc](https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-template-stacks.md)
+for those details.
+
+## Terms
+
+* A **render request** is an instance of a CRD recognized by the
```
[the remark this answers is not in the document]
suskin: I tried rewording this to talk about Kinds instead of CRDs specifically, but it feels about the same to me

design/design-doc-template-stacks-experience.md
```diff
@@ -0,0 +1,505 @@
...
+particular directory. The rbac requirements for the stack must also be
+configured, though some of that is done by the stack manager at runtime.
+At a high level, the steps would be as follows:
+
+1. Create a project directory.
+2. In the project directory, `kubectl crossplane stack init --template
+   myorg/mystack` to create the boilerplate of the stack layout.
+3. Relative to the project directory, put the yamls in
+   `config/stack/manifests/resources/templates`, or a different
+   directory if desired.
```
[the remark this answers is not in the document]
suskin: I've added a few more words to attempt to indicate that the directory doesn't really matter so long as the files show up in the right place in the artifact.

design/design-doc-template-stacks-experience.md
suskin: I've reworded the steps to only mention `stack.yaml`, and I've changed this sentence to say that the numbered section was written as though the merge/rename/move has already happened.

design/design-doc-template-stacks-experience.md
```diff
@@ -0,0 +1,505 @@
...
+`stack.yaml`.
+
+#### Adding a CRD
+
+The tooling will make it simpler to add a CRD from scratch. For example,
+the following would create a basic CRD in the appropriate folder, so
+that it becomes part of the stack:
+
+```
+kubectl crossplane stack crd init WordpressInstance wordpressinstances wordpress.samples.stacks.crossplane.io
```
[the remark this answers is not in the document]
suskin: I made some changes to try this out

design/design-doc-template-stacks-experience.md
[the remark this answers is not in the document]
suskin: I've made some changes to use better terminology and to explain that we should be able to provide reasonable generated names and versions

design/design-doc-template-stacks-experience.md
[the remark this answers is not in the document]
suskin: I changed several of these to improve the readability

design/design-doc-template-stacks-experience.md
suskin: After thinking about it more, I am thinking perhaps running `postUpdate` if no `postCreate` exists could work well.

design/design-doc-template-stacks-experience.md
suskin: I've added some words to clarify this, so I'm going to resolve this thread.

suskin:
Now that some revisions have been added, assigning to @jbw976 for review and discretionary merging

jbw976:
Thank you for driving this design to completion @suskin! can't wait to see it get implemented an into the hands of the community 🙇

Issue #758, opened 2019-09-09
Scoped verb access to CRDs a package depends on


jbw976:
This issue started with discussion in https://github.com/crossplaneio/crossplane/pull/672#discussion_r319581410

### What problem are you facing?
As per the stacks security and isolation [design doc](https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-stacks-security-isolation.md#allowed-resource-access), a stack will be given access to all verbs on CRDs that the stack depends on:

> The allowed verbs for the CRDs the stack owns and depends on will be the same. Stacks will likely need full CRUD access, regardless of if the stack owns or depends on the CRD.

This is because most stacks will likely need to create instances of CRDs they depend on (e.g. WordPress stack will create instances of `MySqlInstance` and `KubernetesCluster`) and the stack will also need to update and delete those instances.  Essentially full CRUD access.

However, some stacks may want more scoped down access though, such as read only verbs (GET, LIST, WATCH).

This isn't blocking v0.3, as this isn't a critical part of the design and the user experience design isn't clear yet so it will take a bit of thinking through.

### How could Crossplane help solve your problem?

When a stack declares the CRDs that it depends on, perhaps it could also declare the type of access (verbs) that it wants for those CRDs as well.  Example:

```yaml
dependsOn:
- kind: Foo
  apiversion: mygroup.io/v1alpha1
  verbs: "get,create,update,delete"
- kind: Bar
  apiversion: yourgroup.io/v1alpha2
  verbs: "get,watch"
```

Another example that instead of explicitly specifying verbs, specifies usage intent instead:

```yaml
dependsOn:
- kind: Foo
  apiversion: mygroup.io/v1alpha1
  readOnly: true
```

Since the stack author has declared `readOnly: true`, the Stack Manager would grant read only verbs, such as GET, LIST, and WATCH.

These examples are not completely in agreement with how dependencies are being specified (the author should be able to express a stack package they depend on instead of just CRDs), but it is meant to convey the spirit of the experience that we may want to allow.

negz:
We no longer support `dependsOn` as a way to request additional permissions, though our package manager design calls for support for something similar per https://github.com/crossplane/crossplane/issues/1767.

PR #959, opened 2019-10-22
Cluster scoped cluster stacks

### LINKED ISSUE #958: Cluster scoped stacks should be able to install cluster scoped CRDs
With the new simple resource class selection experience described in the [design doc](https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-simple-class-selection.md#proposal):

> Providers, managed resources, and resource classes would be cluster scoped rather than namespaced.

The current infrastructure stacks have a valid reason to install cluster scoped CRDs now, which is currently blocked.

Initially we chose to block cluster scoped CRDs as a first attempt of the security model to be very restrictive. Details can be found in the [design doc](https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-stacks-security-isolation.md#allowed-resource-access), but the relevant excerpts are here:

> All CRDs the stack owns and depends on are expected to be namespace scoped CRDs. Access to cluster scoped CRDs will not be granted, even for Cluster Stacks.

> … the allowed set of permissions and scope could be expanded, but we are intentionally starting now with a restrictive scope.
> We will allow feedback and real use cases from the community to inform any decisions to potentially expand this security model.

We now have a valid use case that should drive the slight expansion the security model to allow cluster scoped CRDs to be installed by a cluster scoped stack.  Namespaced stacks should still not be allowed to install cluster scoped CRDs.

An initial (non compiling) fix for this can be found in this gist: https://gist.github.com/jbw976/7487710e00e93904f17ca2b86957f4e6

There's a few snags to work out in that fix, but I wanted to provide it as an optional starting point:

*  this func doesn’t return an error in its signature that is part of an interface, and the calling function doesn’t have context to know what type of stack install it’s working with. so some refactoring would be needed.
* there’s not a const for “Namespaced” or “Cluster” stack permissions scope either, which is a bit undesirable.

displague:

Cluster scoped CRDs are now supported since the [Simple Resource Class
Selection](https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-simple-class-selection.md#proposal) design was adopted.

> Providers, managed resources, and resource classes would be cluster scoped rather than namespaced.

The original Stack Isolation design has also been updated, reversing
this previous stance:

> All CRDs the stack owns and depends on are expected to be namespace scoped CRDs. Access to cluster scoped CRDs will not be granted, even for Cluster Stacks.

Closes #958

### Checklist

I have:

### DIFF
--- a/design/one-pager-stacks-security-isolation.md
+++ b/design/one-pager-stacks-security-isolation.md
@@ -2,7 +2,14 @@
 
 * Owner: Jared Watts (@jbw976)
 * Reviewers: Crossplane Maintainers
-* Status: Draft, revision 1.0
+* Status: Draft, revision 1.1
+
+## Revisions
+
+* 1.1
+  * Cluster stacks may own Cluster scoped CRDs
+    [#958](https://github.com/crossplaneio/crossplane/pull/958)
+  * Namespaced stacks may depend on Cluster scoped CRDs [#958](https://github.com/crossplaneio/crossplane/pull/958)
 
 ## Background
 
@@ -69,10 +76,12 @@ This is described in greater details in the sections below.
 
 A few notes on the scope of access to CRDs:
 
-* All CRDs the stack owns and depends on are expected to be namespace scoped CRDs. Access to cluster scoped CRDs will not be granted, even for Cluster Stacks.
+* All CRDs a Namespaced stack owns must be namespace scoped CRDs. Cluster scoped Stacks may own cluster scoped CRDs.
+* Namespaced and Cluster scoped Stacks may depend on a combination of namespaced or cluster scoped CRDs.
 * The allowed verbs for the CRDs the stack owns and depends on will be the same. Stacks will likely need full CRUD access, regardless of if the stack owns or depends on the CRD.
 
-In the future, other security modes could be enabled such as a more explicit model, and the allowed set of permissions and scope could be expanded, but we are intentionally starting now with a restrictive scope.
+In the future, other security modes could be enabled such as a more explicit model, and the allowed set of permissions and scope could be expanded.  We are intentionally starting
+with a restrictive scope compatible with Crossplane Managed Resource Infrastructure Stacks and [Resource Class Selection](https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-simple-class-selection.md#proposal).
 We will allow feedback and real use cases from the community to inform any decisions to potentially expand this security model.
 Any future expansions to the model should be accompanied by a thoughtful design.
 
--- a/apis/stacks/v1alpha1/types.go
+++ b/apis/stacks/v1alpha1/types.go
@@ -23,6 +23,7 @@ import (
 	batch "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbac "k8s.io/api/rbac/v1"
+	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	runtime "k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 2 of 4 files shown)


design/one-pager-stacks-security-isolation.md
```diff
@@ -69,10 +69,12 @@ This is described in greater details in the sections below.
 
 A few notes on the scope of access to CRDs:
 
-* All CRDs the stack owns and depends on are expected to be namespace scoped CRDs. Access to cluster scoped CRDs will not be granted, even for Cluster Stacks.
+* All CRDs a Namespaced stack owns must be namespace scoped CRDs. Cluster scoped Stacks may own cluster scoped CRDs.
```
hasheddan: Thoughts on bumping revision version of this doc? Relatively significant change here so might be worth increment.

design/one-pager-stacks-security-isolation.md
```diff
@@ -69,10 +69,12 @@ This is described in greater details in the sections below.
 
 A few notes on the scope of access to CRDs:
 
-* All CRDs the stack owns and depends on are expected to be namespace scoped CRDs. Access to cluster scoped CRDs will not be granted, even for Cluster Stacks.
```
displague: @jbw976 We never actually enforced namespaced vs cluster scoped dependencies.

https://github.com/crossplaneio/crossplane/blob/master/pkg/stacks/unpack.go#L359-L376

The ServiceAccount is still namespaced, which may have an effect on cluster scoped resource interactions, but I'm not entirely sure what.  Could a namespaced or clustered stack now delete cluster scoped stack?

https://github.com/crossplaneio/crossplane/blob/master/pkg/controller/stacks/stack/stack.go#L176-L187

design/one-pager-stacks-security-isolation.md
displague: Good thinking.  Updated with a revision history.

design/one-pager-stacks-security-isolation.md
jbw976: > Could a namespaced or clustered stack now delete cluster scoped stack?

I'm not sure I understand the question here @displague, why would one stack delete another stack?

negz:
Thanks @displague! This is a huge help.

jbw976:
@displague were you able to test this out today?  or @negz, did you either use this in your simple resource class selection work, or did you apply a suitable workaround to get unblocked?

negz:
> did you either use this in your simple resource class selection work, or did you apply a suitable workaround to get unblocked?

I'm not blocked just yet - I tend to just `go run` my stack binaries rather than actually installing them - but I likely will be tomorrow morning when we want to start merging this work. Per https://github.com/crossplaneio/stack-azure/pull/35#issuecomment-544797791 our infra stack integration tests are currently broken by the simple resource class selection work. Right now they're running Crossplane 0.3 so fixing master won't fix them, but @hasheddan is working on getting them running a build of Crossplane from master.

displague:
Verified that a build of the GCP stack from https://github.com/crossplaneio/stack-gcp/pull/56 installs using this branch.

design/one-pager-stacks-security-isolation.md
jbw976: Ah got it, thank you for that detailed explanation @displague.  Providing the appropriate set of verbs for resources that a stack depends on definitely relevant and important here, which is tracked in #758.

displague: A little more thinking on this..

The controller is created in the namespace the Stack was installed into.  The pods of the controller run using a serviceaccount that resides in the same namespace (all pods must). 

The Stack Manager does not include a rule to do anything on `stacks.stacks.crossplane.io` resources in the generated role for the SA, although a dependency included in a stack today could request this rule be added to the role.  Such a dependency would allow this rule to affect (add, change, or delete) namespaced stacks or cluster stacks (since either may reside in the same namespace as the stack).

To prevent a namespaced stack from being able to affect cluster scoped stacks, we would want to Stacks that manage cluster scoped CRDs to themselves be installed at the cluster scope.

displague: > The Stack Manager does not include a rule to do anything on stacks.stacks.crossplane.io resources in the generated role for the SA, although a dependency included in a stack today could request this rule be added to the role.

https://github.com/crossplaneio/crossplane/issues/899 would have the types s stack can depend on to claim and class types, not "stack" and "stackinstall" types.

> Such a dependency would allow this rule to affect (add, change, or delete) namespaced stacks or cluster stacks (since either may reside in the same namespace as the stack).

If such a role were granted (as has been considered in the design of template stacks, potentially giving the template stack controller access to the stack resource) the rule to do so should be restricted to the named resource.  Likewise, if the default SA rules were extended to include Stack, ClusterStackInstall, or StackInstall access, this could be restricted by name. 

> To prevent a namespaced stack from being able to affect cluster scoped stacks (which would affect (potentially delete) their cluster scoped CRDs and CRs), we would want Stacks that manage cluster scoped CRDs to themselves be installed at the cluster scope.

I don't think my earlier argument holds continued merit based on the revised thinking and alternatives reflected here.  Isolating Stacks from ClusterStack is not justified here.

Issue #752, opened 2019-09-06
Establish pattern for browsing generated resources and their parent child relationships


jbw976:
### What problem are you facing?
Controllers in Crossplane, for example the controller installed by a stack, are likely to create multiple resources as part of their normal operations.  It would be a useful experience for the user to be able to view and browse all of these resources. But how can they find them?

Take for example the [WordPress stack](https://github.com/crossplaneio/sample-stack-wordpress/).  In response to the user creating a `WordPressInstance`, the wordpress controller will create multiple resources such as `MySqlInstance` and `KubernetesCluster` claims, `KubernetesApplication` and associated `KubernetesApplicationResources`, etc.

How can the user easily find, browse, and view all of these resources that the stack's controller generated, as well as the hierarchical parent child relationships between them?

### How could Crossplane help solve your problem?
One approach to provide an experience around this would be for the user or the stack to specify a selector that can be used to query all the API objects the stack has generated.  The stack's controller would add these lables/annotations to each resource it generates.  For example:
```
 # a selector for all the API objects generated by this stack
 selector:
   matchLabels:
    core.crossplane.io/name: "wordpress"
```

How can these easily be queried, considering there are multiple different types in play, and the Kubernetes API is completely type centric?  Would a query need to be sent off for every known type in the API? That sounds suboptimal.

We need to do some thinking to establish the pattern here, then document it or build  functionality in a library that each stack can reuse to easily implement this.

jbw976:
/cc @rathpc @thephred 

@rathpc were you already doing some thinking and prototyping on this topic? Is there any write-up of your learnings that you can share here?

turkenh:
I was also trying to figure out this in the context of trace cli #744.

> Take for example the WordPress stack. In response to the user creating a WordPressInstance, the wordpress controller will create multiple resources such as MySqlInstance and KubernetesCluster claims, KubernetesApplication and associated KubernetesApplicationResources, etc.
> How can the user easily find, browse, and view all of these resources that the stack's controller generated, as well as the hierarchical parent child relationships between them?

Currently, all resources generated as a result of creating a WordPressInstance has owner reference set  back to the WordPressInstance. Thanks to jsonpath support of kubectl, these resources can be found by the following _ugly_ command (putting here as a workaround until we have a better solution):

```
[...]

Issue #845, opened 2019-09-24
Kinds: add all Crossplane-managed objects to the `all` category


suskin:
### What problem are you facing?

As a user, when I run `kubetl get all -A`, I expect all of the resources on the cluster to show up, but no Crossplane-managed resources do.

### How could Crossplane help solve your problem?

`all` is a [category](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#categories); we could add all Crossplane-created resources to the `all` category.

https://book.kubebuilder.io/reference/markers/crd.html

suskin:
I believe there was some discussion about this elsewhere, but I also want to note that this ties into a larger discussion about using categories in other ways. For example, a `crossplaneresources` category for listing only Crossplane-managed resources, or a `crossplaneproviders` category for showing all of the configured providers.

I am thinking we can still add categories that seem nice to have (such as `all`) before we have come up with a comprehensive categorization approach, so that we can start using them earlier.

negz:
It's not clear to me after a brief Googling whether this categories concept exists only at the `kubectl` level, or is supported by the API server. If the latter were true it could be a game changer for a lot of our implementations; i.e. being able to grab a list of all resource classes, regardless of kind.

muvaf:
There is support for `category` in `CustomResourceDefinition`s and we use a set of defaults everywhere like `managed` for managed resources but it's up to the author of the kinds. I'm not sure if we'd like to manipulate the CRDs installed by package manager though. Feel free to re-open if you think we should do something additional here.

PR #961, opened 2019-10-22
add initial one pager for relationship labels

rathpc:
This PR adds a one pager to address #752 

[skip ci]

### DIFF
--- a/design/one-pager-stack-relationship-labels.md
+++ b/design/one-pager-stack-relationship-labels.md
@@ -4,9 +4,18 @@
 - Reviewers: Crossplane Maintainers
 - Status: Draft
 
+**_NOTE: The focus for this design is long term and may not be implemented immediately_**
+
 ## Proposal
 
 [crossplaneio/crossplane#752](https://github.com/crossplaneio/crossplane/issues/752)
+
+#### These are the labels we are proposing to add:
+
+- **`core.crossplane.io/parent-kind`**
+- **`core.crossplane.io/parent-name`**
+- **`core.crossplane.io/parent-namespace`**
+- **`core.crossplane.io/parent-uid`**
 
 ## Problem
 
@@ -82,6 +91,7 @@ metadata:
     core.crossplane.io/parent-kind: "stack.stacks.crossplane.io"
     core.crossplane.io/parent-name: "sample-stack-wordpress"
     core.crossplane.io/parent-namespace: "app-project1-dev"
+    core.crossplane.io/parent-uid: "ec52c8c2-379f-45ec-9458-e40f070f8d2e"
     app.kubernetes.io/managed-by: stack-manager
 ...
 ```
@@ -92,17 +102,6 @@ metadata:
 apiVersion: wordpress.samples.stacks.crossplane.io/v1alpha1
 kind: WordpressInstance
 metadata:
-  annotations:
-    kubectl.kubernetes.io/last-applied-configuration: |
-      {
-        "apiVersion":"wordpress.samples.stacks.crossplane.io/v1alpha1",
-        "kind":"WordpressInstance",
-        "metadata":{
-          "annotations":{},
-          "name":"my-wordpressinstance",
-          "namespace":"app-project1-dev"
-        }
-      }
   creationTimestamp: "2019-10-23T23:47:40Z"
   generation: 1
   name: my-wordpressinstance
@@ -111,9 +110,10 @@ metadata:
   selfLink: /apis/wordpress.samples.stacks.crossplane.io/v1alpha1/namespaces/app-project1-dev/wordpressinstances/my-wordpressinstance
   uid: f2d13a15-1f9b-40a7-a173-a40abefa61bf
   labels:
-    core.crossplane.io/parent-kind: "customresourcedefinition.apiextensions.k8s.io"
+    core.crossplane.io/parent-kind: "stack.stacks.crossplane.io"
     core.crossplane.io/parent-name: "sample-stack-wordpress"
     core.crossplane.io/parent-namespace: "app-project1-dev"
+    core.crossplane.io/parent-uid: "ec52c8c2-379f-45ec-9458-e40f070f8d2e"
     app.kubernetes.io/managed-by: stack-manager
 ...
 ```
@@ -130,6 +130,7 @@ metadata:
     core.crossplane.io/parent-kind: "wordpressinstances.wordpress.samples.stacks.crossplane.io"
     core.crossplane.io/parent-name: "my-wordpressinstance"
     core.crossplane.io/parent-namespace: "app-project1-dev"
+    core.crossplane.io/parent-uid: "f2d13a15-1f9b-40a7-a173-a40abefa61bf"
     app.kubernetes.io/managed-by: stack-manager
 ...
 ```
@@ -146,6 +147,7 @@ metadata:
     core.crossplane.io/parent-kind: "wordpressinstances.wordpress.samples.stacks.crossplane.io"
     core.crossplane.io/parent-name: "my-wordpressinstance"
     core.crossplane.io/parent-namespace: "app-project1-dev"
+    core.crossplane.io/parent-uid: "f2d13a15-1f9b-40a7-a173-a40abefa61bf"
     app.kubernetes.io/managed-by: stack-manager
 ...
 ```
@@ -162,6 +164,7 @@ metadata:
     core.crossplane.io/parent-kind: "wordpressinstances.wordpress.samples.stacks.crossplane.io"
     core.crossplane.io/parent-name: "my-wordpressinstance"
     core.crossplane.io/parent-namespace: "app-project1-dev"
+    core.crossplane.io/parent-uid: "f2d13a15-1f9b-40a7-a173-a40abefa61bf"
     app.kubernetes.io/managed-by: stack-manager
 ...
 ```
--- a/design/one-pager-stack-relationship-labels.md
+++ b/design/one-pager-stack-relationship-labels.md
@@ -1,0 +1,159 @@
+# Stack Parent/Child Relationship Label Spec
+
+- Owner: Steven Rathbauer (@rathpc)
+- Reviewers: Crossplane Maintainers
+- Status: Draft
+
+## Proposal
+
+[crossplaneio/crossplane#752](https://github.com/crossplaneio/crossplane/issues/752)
+
+## Problem
+
+There is currently no easy way to get a list of all resources that are owned by a parent resource.
+We can leverage CRD labels to make this easier. As long as we have a standard format for doing so.
+
+Per the proposal, it seems that the closest example of what I am referring to is shown on 
+[this page](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels) 
+within the `app.kubernetes.io/part-of` label. However since we are not exactly using these labels 
+for the intentions defined within that page, we should use something following the patterns instead.
+
+As long as we define this label on all children CRD's, we should then be able to query all children
+of a given parent object using those labels.
+
+## Design
+
+> _**Important Note**: These files are post processed yaml outputs, **NOT** static package files. The purpose of showing these is only to better outline and explain the relationships with respect to this specific example._
+
+Wordpress example file tree structure of the example files below:
+
+```text
+stack.yaml
+|
+`-- crd.yaml
+    |
+    `-- resource.yaml
+        |-- kubernetesapplication.yaml
+        |-- kubernetescluster.yaml
+        |-- mysqlinstance.yaml
+        |-- ...
+```
+
+### Example wordpress `stack.yaml`
+
+```yaml
+apiVersion: stacks.crossplane.io/v1alpha1
+kind: Stack
+metadata:
+  creationTimestamp: "2019-10-23T23:47:38Z"
+  generation: 1
+  name: sample-stack-wordpress
+  namespace: app-project1-dev
+  ownerReferences:
+  - apiVersion: stacks.crossplane.io/v1alpha1
+    kind: StackInstall
+    name: sample-stack-wordpress
+    uid: 86c89c94-cfc0-474e-b3b3-beffc09e7793
+  resourceVersion: "6431"
+  selfLink: /apis/stacks.crossplane.io/v1alpha1/namespaces/app-project1-dev/stacks/sample-stack-wordpress
+  uid: ec52c8c2-379f-45ec-9458-e40f070f8d2e
+  labels:
+    app.kubernetes.io/managed-by: stack-manager
+...
+```
+
+### Example wordpress `crd.yaml`
+
+```yaml
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  creationTimestamp: null
+  name: wordpressinstances.wordpress.samples.stacks.crossplane.io
+  labels:
+    core.crossplane.io/parent-kind: "stack.stacks.crossplane.io"
+    core.crossplane.io/parent-name: "sample-stack-wordpress"
+    core.crossplane.io/parent-namespace: "app-project1-dev"
+    app.kubernetes.io/managed-by: stack-manager
+...
+```
+
+### Example wordpress `resource.yaml`
+
+```yaml
+apiVersion: wordpress.samples.stacks.crossplane.io/v1alpha1
+kind: WordpressInstance
+metadata:
+  annotations:
+    kubectl.kubernetes.io/last-applied-configuration: |
+      {
+        "apiVersion":"wordpress.samples.stacks.crossplane.io/v1alpha1",
+        "kind":"WordpressInstance",
+        "metadata":{
+          "annotations":{},
+          "name":"my-wordpressinstance",
+          "namespace":"app-project1-dev"
+        }
+      }
+  creationTimestamp: "2019-10-23T23:47:40Z"
+  generation: 1
+  name: my-wordpressinstance
+  namespace: app-project1-dev
+  resourceVersion: "6445"
+  selfLink: /apis/wordpress.samples.stacks.crossplane.io/v1alpha1/namespaces/app-project1-dev/wordpressinstances/my-wordpressinstance
+  uid: f2d13a15-1f9b-40a7-a173-a40abefa61bf
+  labels:
+    core.crossplane.io/parent-kind: "customresourcedefinition.apiextensions.k8s.io"
+    core.crossplane.io/parent-name: "sample-stack-wordpress"
+    core.crossplane.io/parent-namespace: "app-project1-dev"
+    app.kubernetes.io/managed-by: stack-manager
+...
+```
+
+### Example wordpress `kubernetesapplication.yaml`
+
+```yaml
+apiVersion: workload.crossplane.io/v1alpha1
+kind: KubernetesApplication
+metadata:
+  name: wordpress-app-wordpress
+  labels:
+    stack: sample-stack-wordpress
+    core.crossplane.io/parent-kind: "wordpressinstances.wordpress.samples.stacks.crossplane.io"
+    core.crossplane.io/parent-name: "my-wordpressinstance"
+    core.crossplane.io/parent-namespace: "app-project1-dev"
+    app.kubernetes.io/managed-by: stack-manager
+...
+```
+
+### Example wordpress `kubernetescluster.yaml`
+
+```yaml
+apiVersion: compute.crossplane.io/v1alpha1
+kind: KubernetesCluster
+metadata:
+  name: wordpress-cluster-wordpress
+  labels:
+    stack: sample-stack-wordpress
+    core.crossplane.io/parent-kind: "wordpressinstances.wordpress.samples.stacks.crossplane.io"
+    core.crossplane.io/parent-name: "my-wordpressinstance"
+    core.crossplane.io/parent-namespace: "app-project1-dev"
+    app.kubernetes.io/managed-by: stack-manager
+...
+```
+
+### Example wordpress `mysqlinstance.yaml`
+
+```yaml
+apiVersion: database.crossplane.io/v1alpha1
+kind: MySQLInstance
+metadata:
+  name: wordpress-mysql-wordpress
+  labels:
+    stack: sample-stack-wordpress
+    core.crossplane.io/parent-kind: "wordpressinstances.wordpress.samples.stacks.crossplane.io"
+    core.crossplane.io/parent-name: "my-wordpressinstance"
+    core.crossplane.io/parent-namespace: "app-project1-dev"
+    app.kubernetes.io/managed-by: stack-manager
+...
+```
--- a/design/one-pager-stack-relationship-labels.md
+++ b/design/one-pager-stack-relationship-labels.md
@@ -1,6 +1,6 @@
 # Stack Parent/Child Relationship Label Spec
 
-- Owner: Steven Rathbauer (@rathpc)
+- Owner: Steven Rathbauer ([@rathpc](https://github.com/rathpc))
 - Reviewers: Crossplane Maintainers
 - Status: Draft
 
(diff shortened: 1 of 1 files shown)


design/one-pager-stack-relationship-labels.md
```diff
@@ -0,0 +1,49 @@
...
+- Reviewers: Crossplane Maintainers
+- Status: Draft
+
+## Proposal
+
+[crossplaneio/crossplane#934](https://github.com/crossplaneio/crossplane/issues/934)
+
+## Problem
+
+There is currently no easy way to get a list of all resources that are owned by a parent resource. We can leverage CRD labels to make this easier. As long as we have a standard format for doing so.
```
suskin: Would you be able to wrap long lines at a shorter length? It makes them easier to comment on in GitHub, since each line is shorter : )

design/one-pager-stack-relationship-labels.md
```diff
@@ -0,0 +1,49 @@
...
+
+Per the proposal, it seems that the clearest example of what I am referring to is shown on this page https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels within the `app.kubernetes.io/part-of` label.
```
suskin: Awesome.

design/one-pager-stack-relationship-labels.md
```diff
@@ -0,0 +1,49 @@
...
+kind: CustomResourceDefinition
+metadata:
+  labels:
+    controller-tools.k8s.io: "1.0"
+    core.crossplane.io/part-of: "example-app-name"
+  name: mytypes.samples.crossplane.io
+...
+```
+
+As per the above two file example snippets, as long as the `core.crossplane.io/part-of` label is defined on children CRD's to the value of the stacks name, then we can use that label to look up all children of a given parent.
```
suskin: If I'm understanding the [recommended labels document that was linked earlier](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels), it would make the most sense to use the labels as described in that document. So, `app.kubernetes.io/part-of` instead of `core.crossplane.io/part-of`. The reason to do this would be so that other tools can use the same labels.

design/one-pager-stack-relationship-labels.md
```diff
@@ -0,0 +1,49 @@
...
+
+[crossplaneio/crossplane#934](https://github.com/crossplaneio/crossplane/issues/934)
+
+## Problem
+
+There is currently no easy way to get a list of all resources that are owned by a parent resource. We can leverage CRD labels to make this easier. As long as we have a standard format for doing so.
+
+Per the proposal, it seems that the clearest example of what I am referring to is shown on this page https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels within the `app.kubernetes.io/part-of` label.
+
+As long as we define this label on all children CRD's, we should then be able to query all children of a given parent name using those labels.
```
suskin: What are your thoughts on how to record multiple levels of parent/child relationships?

design/one-pager-stack-relationship-labels.md
```diff
@@ -0,0 +1,49 @@
...
+
+### Example `crd.yaml`
+
+```yaml
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  labels:
+    controller-tools.k8s.io: "1.0"
+    core.crossplane.io/part-of: "example-app-name"
```
suskin: If I'm reading the document correctly, it'd make sense to include all of the "recommended labels", so that a query could select on multiple of them. What are your thoughts?

design/one-pager-stack-relationship-labels.md
turkenh: >  We can leverage CRD labels to make this easier.

Is CRD labels automatically propagated to instances (CRs) ?

design/one-pager-stack-relationship-labels.md
turkenh: What about differentiating resources of different instances of a particular CRD? `app.kubernetes.io/instance` looks like a good candidate on the page you linked.

design/one-pager-stack-relationship-labels.md
rathpc: That works for me!

design/one-pager-stack-relationship-labels.md
rathpc: I am ok with doing this as long as it doesn't feel too "heavy-handed".

rathpc: I think it seems fair to assume the following if we include more of the recommended labels:

```yaml
app.kubernetes.io/name: mysql
```
`name` would be the current crd

```yaml
app.kubernetes.io/instance: wordpress-abcxzy
```
`instance` would be the current instance of the current part-of stack

```yaml
app.kubernetes.io/version: "5.7.21"
```
`version` would be the current crd's semantic version

```yaml
app.kubernetes.io/component: database
```
`component` would be the current crd's category

```yaml
app.kubernetes.io/part-of: wordpress
```
`part-of` would be the current crd's parent stack

design/one-pager-stack-relationship-labels.md
rathpc: @turkenh yes thank you!

@suskin that is a good question. As of right now I was constraining this relationship pattern to only retrieving direct children of a parent.

It seems that down the line we can explore retrieving all children of a parent and children of those children. Additionally maybe even the other way around and getting parents from a given child. Although all of this obviously introduces far more complexity.

Is it fair to limit the scope of these relationships at this stage to just single levels?

suskin: Hmm well, as I'm thinking more about this, I am realizing I may not have been thinking about the same use-cases as what the document is describing. @rathpc would you be able to also explain the specific use-cases we **may** want to satisfy (e.g. "get me all the children of my Wordpress object", "get me all the children of the Wordpress stack"), and also call out which ones we're trying to satisfy with this proposal?

rathpc: You hit the nail on the head there, the use case I am trying to satisfy with this proposal is "Get me all children of some parent object." Whether that is a stack or not shouldn't matter since the larger goal of this is to just be able to "dive" down through resources.

The most analogous example would be `node_modules` in npm. Some package in npm has dependencies and those dependencies may have their own dependencies and so on.

The same should be illustrated by this proposal where some stack has children crds and those crds can have other children crds and so on, while limiting the diving to one level at a time.

Does that help explain my thought process at all? @suskin

design/one-pager-stack-relationship-labels.md
thephred: Do we also need to include a `namespace` so a reference can cross namespace boundaries?

rathpc: per zoom call, this is the more preferred pattern to go with for this specific use case. We will reserve the _recommended labels_ for when we want to explore that particular use case
```yaml
core.crossplane.io/parent-kind: 'kind'
core.crossplane.io/parent-name: 'name'
core.crossplane.io/parent-namespace: 'namespace'
```

Using this new pattern for a more real example:
```yaml
core.crossplane.io/parent-kind: 'wordpress.samples.crossplane.io'
core.crossplane.io/parent-name: 'wordpress-blue'
core.crossplane.io/parent-namespace: 'xyz'
```

design/one-pager-stack-relationship-labels.md
thephred: I don't believe so.

turkenh:
@rathpc I've been working on [adding a tracing functionality](https://github.com/crossplaneio/crossplane-cli/pull/15) to Crossplane cli which will allow walking all children/related objects for a given crossplane resource (namely KubernetesApplication, KubernetesApplicationResource, Claims, Managed Resources, Resource Classes and Provider). Based on this work I thought I can provide some input to this PR. But not sure if this kind of relations are in scope of this documentation.
It would be great if we can make this more explicit, so that I can review accordingly.

design/one-pager-stack-relationship-labels.md
displague: CRs are managed by their own controller.  The Stack Manager does not intervene in the creation or updates of those resources.  It would be possible to setup a webhook from the StackManager to add these annotations if we want to pursue this later.  That would be very helpful in identifying Stack parentage. Nice, @turkenh.

design/one-pager-stack-relationship-labels.md
```diff
@@ -0,0 +1,57 @@
...
+  name: mytypes.samples.crossplane.io
+...
+```
+
+As per the above two file example snippets, as long as the labels are defined on children CRD's,
+then we can use that label to look up all children of a given parent.
+
+Using `kubectl` for example:
+```
+kubectl get pods -l core.crossplane.io/parent-name=wordpress-blue
```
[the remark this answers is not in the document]
displague: This should probably `get crds`

design/one-pager-stack-relationship-labels.md
```diff
@@ -0,0 +1,57 @@
...
+- Reviewers: Crossplane Maintainers
+- Status: Draft
+
+## Proposal
+
+[crossplaneio/crossplane#934](https://github.com/crossplaneio/crossplane/issues/934)
+
+## Problem
+
+There is currently no easy way to get a list of all resources that are owned by a parent resource.
```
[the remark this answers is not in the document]
displague: I had some thoughts [here](https://github.com/crossplaneio/crossplane/pull/961#discussion_r338846328) which could allow us to explore setting the labels on grandchild resources of the stack in the future, but I agree that we should keep the scope of this design on what we are prepared to implement (and mentally carry) now.

design/one-pager-stack-relationship-labels.md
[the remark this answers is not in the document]
rathpc: correcting this, thanks

design/one-pager-stack-relationship-labels.md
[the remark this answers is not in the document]
rathpc: I am updating the examples included to make it clearer what this is trying to accomplish, thanks

design/one-pager-stack-relationship-labels.md
displague: In the latest update, https://github.com/crossplaneio/crossplane/pull/961/files#diff-4daf64a45270834c13f7f119f2d6aa98R14, make it more clear that we may not need to label/annotate these 'grandchild' resources at all.

By knowing that a Stack owns a particular Kind/CRD, we can assert that **all** resources of that Kind are owned by the CRD owning Stack.

design/one-pager-stack-relationship-labels.md
```diff
@@ -0,0 +1,165 @@
...
+We can leverage CRD labels to make this easier. As long as we have a standard format for doing so.
+
+Per the proposal, it seems that the closest example of what I am referring to is shown on this page
+https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels within the
```
suskin: Nitpick: I would normally expect this to be a fancy link, such as:

```
shown on [this page](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels), within the . . . 
```

That way it takes up less space for readers who're looking at the rendered markdown.

design/one-pager-stack-relationship-labels.md
```diff
@@ -0,0 +1,165 @@
...
+`app.kubernetes.io/part-of` label. However since we are not exactly using these labels for the
+intentions defined within that page, we should use something following the patterns instead.
+
+As long as we define this label on all children CRD's, we should then be able to query all children
+of a given parent name using those labels.
```
suskin: Nitpick: `parent name` isn't entirely accurate, since we need the kind as well. I would probably use `parent object` instead.

design/one-pager-stack-relationship-labels.md
```diff
@@ -0,0 +1,165 @@
...
+  ownerReferences:
+  - apiVersion: stacks.crossplane.io/v1alpha1
+    kind: StackInstall
+    name: sample-stack-wordpress
+    uid: 86c89c94-cfc0-474e-b3b3-beffc09e7793
+  resourceVersion: "6431"
+  selfLink: /apis/stacks.crossplane.io/v1alpha1/namespaces/app-project1-dev/stacks/sample-stack-wordpress
+  uid: ec52c8c2-379f-45ec-9458-e40f070f8d2e
+  labels:
+    app.kubernetes.io/managed-by: stack-manager
```
suskin: Nitpick: theoretically, the `Stack` **could** be considered a child of the `StackInstall` or `ClusterStackInstall` object, so we may include that in the labels in the future.

design/design-doc-stacks.md
```diff
@@ -495,14 +495,19 @@ spec:
...
 
+Labels and annotations provided in the source CRD are preserved, such as
+`controller-tools.k8s.io` in the example above.
+
 The labels added by Crossplane tools serve the following purpose:
 
 Label | Value | Purpose
 ---|--- | ---
 `app.kubernetes.io/managed-by` | `stack-manager` | This [recommended label](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/) identifies the tool being used to manage the operation of the application.
+`core.crossplane.io/parent-kind` | `{kind}.{apiGroup}` | The kind of the Stack ("stack.stacks.crossplane.io")
```
suskin: I am not sure whether it makes more sense to use `parent-kind` or `parent.kind`, but I believe it's worth considering. I think this is probably a question for someone with more prior experience in Kubernetes land, such as @jbw976

design/one-pager-stack-relationship-labels.md
```diff
@@ -0,0 +1,165 @@
...
+  labels:
+    stack: sample-stack-wordpress
+    core.crossplane.io/parent-kind: "wordpressinstances.wordpress.samples.stacks.crossplane.io"
+    core.crossplane.io/parent-name: "my-wordpressinstance"
+    core.crossplane.io/parent-namespace: "app-project1-dev"
+    app.kubernetes.io/managed-by: stack-manager
+...
+```
+
+As per the above file example snippets, as long as the labels are defined on children CRD's,
```
suskin: I would probably use `objects` instead of `CRDs` here, since we're talking about finding any object type, not just CRDs. CRD typically refers to the definition itself, so in this context, that objects like the one which came from `crd.yaml` in this document.

design/one-pager-stack-relationship-labels.md
```diff
@@ -0,0 +1,165 @@
...
+then we can use that label to look up all children of a given parent.
+
+Using `kubectl` for example:
+```
+kubectl get crds -l core.crossplane.io/parent-name=my-wordpressinstance
```
suskin: This wouldn't match anything in the examples provided above, so I would recommend changing it to be one of the following instead:

```
kubectl get kubernetesapplication -l core.crossplane.io/parent-name=my-wordpressinstance
```

or

```
kubectl get crds -l core.crossplane.io/parent-name=sample-stack-wordpress
```

I would probably go with the first one since most users who are using this for debugging will probably not be looking at the crds, since they don't change very often. I guess you could always put both examples though!

Some related discussion that would influence how these queries are done can be found in #845

design/one-pager-stack-relationship-labels.md
rathpc: I agree. Although, for the purposes of showing an example I think this is sufficient so I will leave that out for now.

design/design-doc-stacks.md
rathpc: I think going with hyphenated labels follows the common theme taken by k8s in many of their recommended and common labels so I think it works but I will defer to @jbw976 for the call.

rathpc:
@suskin I just pushed changes per your PR feedback. Thank you!

design/one-pager-stack-relationship-labels.md
```diff
@@ -0,0 +1,166 @@
+# Stack Parent/Child Relationship Label Spec
+
+- Owner: Steven Rathbauer (@rathpc)
+- Reviewers: Crossplane Maintainers
+- Status: Draft
+
+## Proposal
+
+[crossplaneio/crossplane#934](https://github.com/crossplaneio/crossplane/issues/934)
```
jbw976: This issue is closed as a dupe of https://github.com/crossplaneio/crossplane/issues/752, which is the tracking issue we are using.  Can you update this link to point to the open issue?

design/one-pager-stack-relationship-labels.md
```diff
@@ -0,0 +1,166 @@
...
+
+## Problem
+
+There is currently no easy way to get a list of all resources that are owned by a parent resource.
+We can leverage CRD labels to make this easier. As long as we have a standard format for doing so.
+
+Per the proposal, it seems that the closest example of what I am referring to is shown on 
+[this page](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels) 
+within the `app.kubernetes.io/part-of` label. However since we are not exactly using these labels
```
jbw976: can you expand on how our goals differ from what's described on that page? I liked how that page was saying:

>  A common set of labels allows tools to work interoperably, describing objects in a common manner that all tools can understand.

I'd love to understand specifically what we're doing that is divergent from the goals on that page, can you add details here?

design/one-pager-stack-relationship-labels.md
```diff
@@ -0,0 +1,166 @@
...
+[this page](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels) 
+within the `app.kubernetes.io/part-of` label. However since we are not exactly using these labels 
+for the intentions defined within that page, we should use something following the patterns instead.
+
+As long as we define this label on all children CRD's, we should then be able to query all children
+of a given parent object using those labels.
+
+## Design
+
+Wordpress example file tree structure of the example files below:
```
jbw976: We are presenting this as information that is provided and known statically, at build/package time.  When a stack author is building their stack package, they won't include all this metadata in statically within their package, right?  This is information that will need to be applied dynamically at runtime by a controller.  Do you think it would be less confusing to not mention the "file tree structure" or ".yaml files" here to avoid confusion?  It may be easier to convey this as a runtime resource graph or list, instead of a build time stack package filesystem layout.  What do you think?

design/one-pager-stack-relationship-labels.md
```diff
@@ -0,0 +1,166 @@
...
+...
+```
+
+As per the above file example snippets, as long as the labels are defined on children objects,
+then we can use that label to look up all children of a given parent.
+
+Using `kubectl` for example:
+
+```bash
+kubectl get kubernetesapplication -l core.crossplane.io/parent-name=my-wordpressinstance
```
jbw976: I think this is glossing over a particular detail about children types.  This query is using information that appears to be already known that they are looking for a child of type `kubernetesapplication`.  Is there a good way to do this generically, when you don't know what types the children will be?  How would an entity query for all children of a given parent, when they don't know what their types may be?

another thought, queries may probably need to use more information from the parent's labels, e.g. specify the parent's kind and namespace too.  `parent-name` isn't specific enough by itself since many resources in k8s can all have the same name.

design/one-pager-stack-relationship-labels.md
jbw976: By the way, I don't think you need to get into the details of how a controller specifies these labels at runtime, that's probably out of scope for your efforts in this 1-pager that is simply trying to capture the experience and details of the metadata we'll use.  But I do think it's worth mentioning that a controller will be applying the labels at run time as opposed to the stack author at build time.

design/one-pager-stack-relationship-labels.md
rathpc: Will do!

design/one-pager-stack-relationship-labels.md
rathpc: I think @suskin might be able to convey the reasoning we decided to sway away from using those predefined labels better than I can.

design/one-pager-stack-relationship-labels.md
rathpc: @jbw976 So the initial reason I had added all of these file examples was to provide a clear example of what these would look like after. I found that within some of the previous design docs, there was a divergence in continuity between one example file to the next so there wasn't a way to construct a full picture of the example.

I can certainly add some clarification that these are post processed yaml outputs and not static package files.

design/one-pager-stack-relationship-labels.md
rathpc: I was kind of on the fence including this line at all since it really isn't a complete accurate call and was really just a very abstract thought about how you *might* leverage these labels.

I think I would prefer to remove it all together

displague: > Is there a good way to do this generically, when you don't know what types the children will be? 

I don't believe so.

For example, `kubectl get all -lapp=crossplane -A`, would only list the types in the "all" category, which does not include KubernetesApplication.

If Crossplane types included a categories (`//+kubebuilder:categories=crossplane`), then `kubectl get crossplane,all -lapp=crossplane -A` would be possible.  Multiple categories can be provided in this way (`//+kubebuilder:categories=all,crossplane`).

design/one-pager-stack-relationship-labels.md
jbw976: Sounds good, I think that clarification would help here.

design/one-pager-stack-relationship-labels.md
jbw976: @suskin do you remember any of the details about why we wouldn't want to use the common upstream labels here?

design/one-pager-stack-relationship-labels.md
rathpc: removing this chunk from the one-pager

design/one-pager-stack-relationship-labels.md
rathpc: adding now

design/one-pager-stack-relationship-labels.md
suskin: @jbw976 @rathpc the reason is summarized in this passage, which can be found in the introductory section:

> The metadata is organized around the concept of an application.

Essentially, because the labels are organized around the concept of an application, they do not really fit our concept of tracing multiple levels of parent/child relationships. Consider [the label descriptions](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels). They are missing the concept of a hierarchy like what we're trying to build. The label which comes the closest is `part-of`, which reads `The name of a higher level application this one is part of` (example: `wordpress`). None of the other labels are for describing a hierarchical relationship.

Let's make things more concrete with an example. We **could use** `part-of` and/or `instance` to point to parents if we wanted to. But this is not ideal. Consider the following example hierarchy, where `A -> B` means A is a parent of B:

```
WordpressInstance -> KubernetesCluster -> KubernetesApplication -> KubernetesApplicationResource
```

Semantically, the `app.kubernetes.io` labels "are organized around the concept of an application". Since the resources in the example hierarchy are all part of the same application (Wordpress), they should all have the same `instance` and `part-of` labels:

```
WordpressInstance labels
app.kubernetes.io/name: wordpress
app.kubernetes.io/instance: wordpress-instance-efafe342
app.kubernetes.io/component: application-claim
app.kubernetes.io/part-of: wordpress

KubernetesCluster labels
app.kubernetes.io/name: kubernetes
app.kubernetes.io/instance: wordpress-instance-efafe342
app.kubernetes.io/component: cluster
app.kubernetes.io/part-of: wordpress

KubernetesApplication labels
app.kubernetes.io/name: kubernetes-application
app.kubernetes.io/instance: wordpress-instance-efafe342
app.kubernetes.io/component: workload
app.kubernetes.io/part-of: wordpress

KubernetesApplicationResource labels
app.kubernetes.io/name: kubernetes-resource
app.kubernetes.io/instance: wordpress-instance-efafe342
app.kubernetes.io/component: workload-resource
app.kubernetes.io/part-of: wordpress
```

However, what we want is for each one to point back one step in the hierarchy. And that would contradict organizing the labels around an application. The `app.kubernetes.io` labels are still useful, just not for describing a multi-level hierarchy. What do you think? Did that make sense? If this question came up, maybe it would make sense to try to answer it in a little more detail in the document.

design/design-doc-stacks.md
jbw976: Yes, hyphens is what is fairly broadly used in the ecosystem, that sounds good to me

design/one-pager-stack-relationship-labels.md
rathpc: Thank you @suskin! I have added most of this into the document to memorialize these findings and decisions

negz:
https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#owners-and-dependents

I've only skimmed this design, but on the surface it seems to have a lot of overlap with the existing Kubernetes owner reference metadata that we set on most of our references. Would we be reinventing the wheel by also trying to model these relationships with a label?

design/one-pager-stack-relationship-labels.md
```diff
@@ -0,0 +1,229 @@
+# Stack Parent/Child Relationship Label Spec
+
+- Owner: Steven Rathbauer ([@rathpc](https://github.com/rathpc))
+- Reviewers: Crossplane Maintainers
+- Status: Draft
```
jbw976: @rathpc can you add a note for clarity that this may not be implemented right away and it's potentially focused on longer term?

design/one-pager-stack-relationship-labels.md
```diff
@@ -0,0 +1,229 @@
...
+
+## Proposal
```
jbw976: can we list the proposed labels right here in this section? they are currently only listed inline in the examples and it would be nice to see them in a more prominent highlighted location, like here in this section

design/one-pager-stack-relationship-labels.md
```diff
@@ -0,0 +1,229 @@
...
+
+```yaml
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  creationTimestamp: null
+  name: wordpressinstances.wordpress.samples.stacks.crossplane.io
+  labels:
+    core.crossplane.io/parent-kind: "stack.stacks.crossplane.io"
+    core.crossplane.io/parent-name: "sample-stack-wordpress"
```
jbw976: seems useful to also have a UID like @displague suggested in this comment:
https://github.com/crossplaneio/crossplane/pull/1091#discussion_r350961773

and @negz reinforced in https://github.com/crossplaneio/crossplane/pull/1091#discussion_r349346189

maybe `core.crossplane.io/parent-uid`?

rathpc: consider it done!

design/one-pager-stack-relationship-labels.md
```diff
@@ -0,0 +1,229 @@
...
+      }
+  creationTimestamp: "2019-10-23T23:47:40Z"
+  generation: 1
+  name: my-wordpressinstance
+  namespace: app-project1-dev
+  resourceVersion: "6445"
+  selfLink: /apis/wordpress.samples.stacks.crossplane.io/v1alpha1/namespaces/app-project1-dev/wordpressinstances/my-wordpressinstance
+  uid: f2d13a15-1f9b-40a7-a173-a40abefa61bf
+  labels:
+    core.crossplane.io/parent-kind: "customresourcedefinition.apiextensions.k8s.io"
```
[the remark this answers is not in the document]
rathpc: that must have been a typo on my part. updating now

PR #962, opened 2019-10-22
Pin code generation tools

negz:

This successfully invokes the code generation tools that are pinned using go modules when used in combination with https://github.com/upbound/build/pull/78. This will allow us to pin `angryjet` (rather than running it from `master`), and to use `go.mod` `replace` statements to easily work with different tooling builds.

### Checklist

I have:


Makefile
```diff
@@ -86,6 +86,12 @@ fallthrough: submodules
 	@echo Initial setup complete. Running make again . . .
 	@make
 
+# TODO(negz): Merge tags support back into https://github.com/upbound/build
+generate:
+	@$(INFO) go generate -tags generate $(PLATFORM)
```
soorena776: one of the principles that @bassam likes to adhere to, is to download go modules inside `$(WORK_DIR)`. For this to happen, I exported `GOPATH` to a subfolder there. 
https://github.com/upbound/build/pull/74/files#diff-a01ea346de0fbd4a8a8cfccb3b0d7793R143

My suspicion is that the above command will download the required packages in the default GOPATH (which is `$(HOME)/go/mod`), breaking that principle.

apis/generate.go
```diff
@@ -0,0 +1,33 @@
+// +build generate
```
soorena776: Do you think `./tools.go` would be a more generic and appropriate name for this file?

negz: I thought about that, but I went with `generate.go` because this file contains our code generation tags as well as the tool imports required to power those tags. Most of the examples that use `tools.go` seem to be basing the file name on the fact that it _only_ handles the imports for tools that are used elsewhere.

Makefile
negz: I raised https://github.com/upbound/build/pull/78 to make the `go generate` changes upstream, and have confirmed tooling is downloaded to `$WORK_DIR`.

negz:
Adding @turkenh, as I've changed CRD manifest generation to use the version of `controller-gen` pinned by `go.mod`, which required a little refactoring of the recent StackInstall CRD annotation fix.

PR #963, opened 2019-10-23
THE GREAT EXAMPLE CLEANUP OF 2019

negz:

This PR:

* Removes all portable classes.
* Removes namespaces from providers, classes, and managed resources.
* Removes all namespaces. Namespaced objects will default to 'default'.
* Removes all providers except those generated by scripts and those under `providers/`.
* Removes all superfluous resource claims and classes.
* Uses the `crossplane-system` namespace for all infra stacks and secrets.

I've tested this change by applying the latest CRDs from my simple resource class selection branches (freshly rebased on master) to a Kubernetes API server and verifying that all YAML under this directory applies successfully, with the exception of templated files and files with commented-out values that the user must supply.

We'll need to figure out some pattern to associate resource claims with classes in the examples. As part of this PR I removed all `classRef` fields (they all pointed at portable resource classes, except the few that we missed updating for 0.3). We could either:

* Use `.spec.classSelector`
* Use `.spec.classRef`
* Annotate resource classes with `resourceclass.crossplane.io/is-default-class: "true`

```yaml
spec:
  classSelector:
    matchLabels:
      example: "true"
```

At the moment I've standardised on having claims use the above `classSelector` and labelling all resource classes as `example: "true"`. This way whether you're trying out a bucket claim or a kubernetes cluster claim you just apply the example class(es) for your cloud(s) of choice then create the claim. This strategy may be broken for some examples, but will be _less_ broken than what is in master now.

### Checklist

I have:


cluster/examples/gitlab/aws/secrets.sh
```diff
@@ -1,25 +1,25 @@
 #!/usr/bin/env bash
```
negz: Hrm, I guess my editor decided to run https://github.com/mvdan/sh against all the shell scripts under `examples/` too.

cluster/examples/cache/rediscluster/resource-claim.yaml
```diff
@@ -3,10 +3,7 @@ apiVersion: cache.crossplane.io/v1alpha1
 kind: RedisCluster
 metadata:
   name: app-redis
-  namespace: app-project1-dev
```
hasheddan: Noticed the `namespace` is removed on all of the claims. Is this just for simplicity?

negz: Exactly. I figured given they were examples it was fine for everything to be created in `default` or whatever namespace was passed to `kubectl -n`.

cluster/examples/wordpress/aws/wordpress.yaml
```diff
@@ -21,8 +21,7 @@ spec:
   engine: mysql
   masterUsername: masteruser
   providerRef:
-    name: demo-aws-provider
-    namespace: crossplane-system
+    name: example
   writeConnectionSecretToRef:
     name: demo-rds
```
turkenh: I think we need to add `namespace` here.

negz: Good catch! Thank you. I wonder how I was able to submit this example successfully without that. 🤔

negz: Ah - it has commented out security groups so I must have assumed this example did not apply because of that.

PR #973, opened 2019-10-24
Update developer guide to reflect simple resource class selection

negz:

What it says on the tin.

### Checklist

I have:

### DIFF
--- a/docs/services-developer-guide.md
+++ b/docs/services-developer-guide.md
@@ -23,35 +23,32 @@ one or more controllers. A controller is an endless loop that:
 1. Tries to make the actual state match the desired state.
 
 A typical Crossplane managed service consists of five configuration resources
-and four controllers. The GCP Stack's support for Google Cloud Memorystore
+and five controllers. The GCP Stack's support for Google Cloud Memorystore
 illustrates this. First, the configuration resources:
 
-1. A [managed resource]. Managed resources are high-fidelity representations of
-   a resource in an external system such as a cloud provider's API. Managed
-   resources are _non-portable_ across external systems (i.e. cloud providers);
-   they're tightly coupled to the implementation details of the external
-   resource they represent. Managed resources are defined by a Stack. The GCP
-   Stack's [`CloudMemorystoreInstance`] resource is an example of a managed
-   resource.
-1. A [resource claim]. Resource claims are abstract declarations of a need for a
-   service. Resource claims are frequently portable across external systems.
-   Crossplane defines a series of common resource claim kinds, including
-   [`RedisCluster`]. A resource claim is satisfied by _binding_ to a managed
-   resource.
-1. A non-portable [resource class]. Non-portable resource classes represent a
-   class of a specific kind of managed resource. They are the template used to
-   create a new managed resource in order to satisfy a resource claim during
-   [dynamic provisioning]. Non-portable resource classes are tightly coupled to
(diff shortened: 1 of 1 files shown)


docs/services-developer-guide.md
```diff
@@ -62,6 +59,16 @@ These resources are powered by:
    controller watches for changes to `CloudMemorystoreInstance` resources and
    calls Google's Cloud Memorystore API to create, update, or delete an instance
    as necessary.
+1. The resource claim scheduling controller. A claim scheduling controller
+   exists for each kind of resource class that could satisfy a resource claim.
+   This controller is unaware of any external system - it simply schedules
+   resource claims to resource classes that match their class selector labels,
+   so that they may be handled by the resource claim controller.
```
hasheddan: Would it be appropriate to add here that if multiple classes exist that match labels then the controller will select at random?

docs/services-developer-guide.md
```diff
@@ -62,6 +59,16 @@ These resources are powered by:
...
+1. The resource claim defaulting controller. A claim defaulting controller
+   exists for each kind of resource class that could satisfy a resource claim.
+   This controller is unaware of any external system - it allocates resource
+   claims that do not specify a class selector to a resource class annotated as
```
hasheddan: Maybe add does not specify a class selector / class reference / managed reference? Not sure if that would just make it more confusing? Potentially solved by the example implementation below which specifies the predicates.

docs/services-developer-guide.md
```diff
@@ -161,10 +163,9 @@ They can be added by hand, but new services are encouraged to use [`angryjet`]
 to generate them automatically using `go generate` per the `angryjet`
 documentation.
 
-Note that in many cases a suitable provider, resource claim, and portable
-resource class will already exist. Frequently adding support for a new managed
-service requires only the definition of a new managed resource and non-portable
-resource class.
+Note that in many cases a suitable provider and resource claim will already
+exist. Frequently adding support for a new managed service requires only the
+definition of a new managed resource and resource class.
```
hasheddan: Probably fine because we say "Frequently", but I don't know if it would be useful to point out network resources and other types that do not have a class.

docs/services-developer-guide.md
hasheddan: I think these comments below https://github.com/crossplaneio/crossplane/pull/973/files#diff-06c88f31ffd6d41fcdf2a50c38c6c9afR803 are sufficient.

docs/services-developer-guide.md
negz: I thought about that, but I don't think going into detail on how these controllers work from a user's perspective is best explained in the developer guide. I do think the developers should be familiar with these concepts, but it's probably better to clarify things like this in (e.g) the services guide and consider knowing how to use services a prerequisite to building new services.

docs/services-developer-guide.md
negz: > I don't think going into detail on how these controllers work from a user's perspective is best explained in the developer guide.

This is another area where I went with the above philosophy.

docs/services-developer-guide.md
negz: True, though that could change if https://github.com/crossplaneio/crossplane/issues/968 becomes a reality. Since this is arguably good enough I'm going to take the path of least resistance and leave it as is for now.

PR #983, opened 2019-10-25
optionally install stack-rook during Crossplane helm chart install

### LINKED ISSUE #982: Crossplane install should optionally install Rook stack also
With #867, we added the optional install of AWS/GCP/Azure stacks during the installation of the Crossplane Helm chart.  Now that we have a Rook stack, we should add an option for installing that stack as well.

jbw976:

This PR adds `stack-rook` to the set of optional stacks that can be installed through the Crossplane helm chart.  Very similar to the existing options for the AWS/GCP/Azure stacks, it will be installed at Helm chart install time with the following value set:

```
--set clusterStacks.rook.deploy=true
```

Note there wasn't documentation added for the original 3 cloud provider stacks, so it would be useful to add docs for how to use this general feature to install stacks along with the helm chart, but that should not block this PR since this docs gap already exists.

Fixes #982 

### Checklist
I have:

### DIFF
--- a/cluster/charts/crossplane/values.yaml.tmpl
+++ b/cluster/charts/crossplane/values.yaml.tmpl
@@ -23,3 +23,6 @@ clusterStacks:
   azure:
     version: v0.1.0
     deploy: false
+  rook:
+    version: master
+    deploy: false


cluster/charts/crossplane/values.yaml.tmpl
```diff
@@ -22,4 +22,7 @@ clusterStacks:
     deploy: false
   azure:
     version: v0.1.0
-    deploy: false
\ No newline at end of file
+    deploy: false
+  rook:
+    version: master
+    deploy: false
```
displague: With the 0.4 corssplane release, can we give stack-rook an initial 0.0.1 tag and use that as a starting point (at least for the tagged/released crossplane helm chart).  This will keep our Rook install docs in sync.

jbw976: i was planning on doing that as part of the release process, but perhaps i should do that now since it's the default.

jbw976: actually, i'll need to update all the other ones to v0.2.0 after we release, so i'll just follow up on all of those at the same time after release.

turkenh:
> Note there wasn't documentation added for the original 3 cloud provider stacks, so it would be useful to add docs for how to use this general feature to install stacks along with the helm chart, but that should not block this PR since this docs gap already exists.

I'll take care of adding this to documentation.

PR #984, opened 2019-10-26
Update the getting started guide

### LINKED ISSUE #907: Introduce static provisioning in the quick start guide

The https://crossplane.io/docs/v0.3/quick-start-gcp.html guide is our most simple example of using Crossplane, but it does not focus on the most simple scenario - static provisioning. This guide will be of higher importance if and when the services guides become gitops-focused.

Crossplane could start https://crossplane.io/docs/v0.3/quick-start-gcp.html by introducing the static provisioning scenario, then build on that to cover the dynamic provisioning scenario

negz:

Fixes #907

This commit updates the getting started guide to:

* Start simple by introducing static provisioning before dynamic provisioning.
* Discuss dynamic provisioning using simple resource class selection.
* Restructures the guide to bypass the (fairly intimidating) guides to installing Crossplane and setting up your cloud provider account, in favor of giving the cliff notes and linking to the detailed guides. There's a good chance much of our target audience already knows how to use Helm and GCP.

It also updates the concepts doc for simple resource class selection and makes some style and formatting changes in separate commits. It may be easiest to review at the commit level.

### Checklist

I have:


docs/quick-start.md
```diff
@@ -3,22 +3,374 @@ title: Getting Started
...
 ---
-# Quick Start Guide
+# Getting Started
 
 This quick start will demonstrate using Crossplane to deploy a portable MySQL
-database on a single cloud provider, Google Cloud Platform. It serves as an initial introduction to
-Crossplane, but only displays a small portion of its functionality.
+database on a single cloud provider, Google Cloud Platform (GCP). It serves as
+an initial introduction to Crossplane, but only displays a small portion of its
+functionality.
```
soorena776: nit: small set of its features.

docs/quick-start.md
```diff
@@ -3,22 +3,374 @@ title: Getting Started
...
-1. [Provision a MySQL database](quick-start-gcp.md) on GCP. 
+1. [Install Crossplane and the GCP stack](#install-crossplane).
+1. [Configure Crossplane to connect to GCP](#configure-crossplane-to-connect-to-gcp).
+1. [Provision a MySQL instance](#provision-a-mysql-instance).
+1. [Enable Dynamic Provisioning of MySQL databases](#enable-dynamic-provisioning).
+
+## Install Crossplane
+
+We'll start by installing Crossplane using [Helm]. You'll need a working
+Kubernetes cluster ([minikube] will do just fine). Crossplane is currently in
```
soorena776: fyi, `kind` and `k8s` are super popular these days.

docs/quick-start.md
```diff
@@ -3,22 +3,374 @@ title: Getting Started
...
+
+```bash
+# Crossplane lives in the crossplane-system namespace by convention.
+kubectl create namespace crossplane-system
+
+helm repo add crossplane-alpha https://charts.crossplane.io/alpha
+helm install --name crossplane --namespace crossplane-system crossplane-alpha/crossplane
+```
+
+Once Crossplane is installed we'll need to install the GCP [stack]. Installing
```
soorena776: it might be helpful to mention that the user could install other cloud providers stacks if they wanted to, to make it clear that GCP stack is not required

docs/quick-start.md
```diff
@@ -3,22 +3,374 @@ title: Getting Started
...
+apiVersion: stacks.crossplane.io/v1alpha1
+kind: ClusterStackInstall
+metadata:
+  name: stack-gcp
+  namespace: crossplane-system
+spec:
+  package: "crossplane/stack-gcp:master"
+```
+
+Save the above as `stack.yaml`, and apply it by running:
```
soorena776: I like this non-scripted way of explaining. 👍

docs/quick-start.md
```diff
@@ -3,22 +3,374 @@ title: Getting Started
...
+
+```bash
+kubectl apply -f stack.yaml
+```
+
+We've now installed Crossplane with GCP support! Take a look at the guide to
+[installing Crossplane] for more installation options, and to learn how to
```
soorena776: nit: Take a look at the [Crossplane installation][installing Crossplane] for ...

docs/quick-start.md
```diff
@@ -3,22 +3,374 @@ title: Getting Started
...
+Save the above `Provider` as `provider.yaml`, then run:
+
+```bash
+kubectl -n crossplane-system create secret example-gcp-credentials --from-file=credentials.json
+kubectl apply -f provider.yaml
+```
+
+Crossplane can now manage your GCP project! Your service account will need the
+CloudSQL Admin role for this guide. Check out GCP's [Getting Started With
+Authentication] guide if you need help creating a service account and
```
soorena776: This authentication setup might better be moved to before Provider creaetion

docs/quick-start.md
```diff
@@ -3,22 +3,374 @@ title: Getting Started
...
+downloading its `credentials.json` file, and Crossplane's [GCP provider
+documentation] for detailed instructions on setting up your project and service
+account permissions.
+
+## Provision a MySQL Instance
+
+GCP provides MySQL databases using [CloudSQL] instances. Crossplane uses a
```
soorena776: nit: GCP provisions

docs/quick-start.md
```diff
@@ -3,22 +3,373 @@ title: Getting Started
...
+resource and claim pattern to provision and manage cloud resources like CloudSQL
```
soorena776: resource class and claim pattern

docs/quick-start.md
```diff
@@ -3,22 +3,373 @@ title: Getting Started
...
+    settings:
+      tier: db-n1-standard-1
+      dataDiskType: PD_SSD
+      dataDiskSizeGb: 10
+      ipConfiguration:
+        ipv4Enabled: true
+```
+
+First we create a CloudSQL instance. Save the above as `cloudsql.yaml`, then
+apply it using `kubectl`:
```
soorena776: since we have used `kubectl` already, mentioning it here might be confusing

docs/quick-start.md
```diff
@@ -3,22 +3,373 @@ title: Getting Started
...
+```bash
+kubectl --namespace default apply -f mysql.yaml
+```
+
+In Crossplane cloud provider specific resources like the `CloudsqlInstance` we
+created above are called _managed resources_. They're considered infrastructure,
+like a Kubernetes `Node` or `PersistentVolume`. Managed resources exist at the
+cluster scope (they're not namespaced) and let you specify nitty-gritty provider
+specific configuration details. Managed resources that have reached `v1beta1`
+support at least 90% of their underlying cloud provider resource's configuration
```
soorena776: is providing exact percentage necessary?

docs/quick-start.md
```diff
@@ -3,22 +3,373 @@ title: Getting Started
...
+publicIP:                             13 bytes
+serverCACertificateCertSerialNumber:  1 bytes
+serverCACertificateCreateTime:        24 bytes
+serverCACertificateExpirationTime:    24 bytes
+serverCACertificateSha1Fingerprint:   40 bytes
+endpoint:                             13 bytes
+serverCACertificateCert:              1272 bytes
+```
+
+That's all there is to static provisioning with Crossplane! We've create a
```
soorena776: We've created

docs/quick-start.md
```diff
@@ -3,22 +3,373 @@ title: Getting Started
...
+    Type:                  ReferencesResolved
+    Last Transition Time:  2019-10-25T08:09:16Z
+    Reason:                Managed resource is available for use
+    Status:                True
+    Type:                  Ready
+```
+
+## Enable Dynamic Provisioning
+
+Now that we've learned how to statically provision and claim managed resource
```
soorena776: claim managed resources, 
or
claim a managed resource,

docs/quick-start.md
```diff
@@ -3,22 +3,374 @@ title: Getting Started
...
+like; simply submit more `MySQLInstance` resource claims to create more CloudSQL
+instances.
+
+You may have noticed that your resource claim included a `classSelector`. The
+class selector lets you select which resource class to use by [matching its
+labels]. Resource claims like `MySQLInstance` can match different kinds of
+resource class using label selectors, so you could just as easily use the
+exact same `MySQLInstance` to create an Amazon Relational Database Service (RDS)
+instance by creating an `RDSInstanceClass` labelled as `guide: getting-started`.
+When multiple resource classes match the class selector a matching class is
```
soorena776: class selector, a matching c

docs/quick-start.md
```diff
@@ -3,22 +3,378 @@ title: Getting Started
...
-The general steps for this example are as follows:
+In this guide we will:
 
-1. [Install Crossplane and the GCP stack](install-crossplane.md) into your Kubernetes cluster. 
-1. [Configure GCP credentials](cloud-providers/gcp/gcp-provider.md) in Crossplane.
-1. [Provision a MySQL database](quick-start-gcp.md) on GCP. 
+1. [Install Crossplane and the GCP stack](#install-crossplane).
+1. [Configure Crossplane to connect to GCP](#configure-crossplane-to-connect-to-gcp).
+1. [Provision a MySQL instance](#provision-a-mysql-instance).
+1. [Enable Dynamic Provisioning of MySQL databases](#enable-dynamic-provisioning).
```
prasek: Until I understood what you were saying, it was confusing having "**Enable** dynamic provisioning" _after_ "Provision a MySQL instance"

suggest ToC:

1. Install Crossplane
1. Connect to GCP
1. Provision a CloudSQL Instance
1. Dynamic Provisioning of MySQL Claims

docs/quick-start.md
```diff
@@ -3,22 +3,378 @@ title: Getting Started
...
+
+```bash
+kubectl apply -f cloudsql.yaml
+```
+
+Crossplane is now creating the `CloudsqlInstance`! Before we can use it, we need
+to claim it.
+
+```yaml
+apiVersion: database.crossplane.io/v1alpha1
```
prasek: these examples are a little less magic (in a good way) than the `cat` examples; esp. nice for readability when just scrolling through

docs/quick-start.md
prasek: maybe "a large portion" or "majority" instead?

docs/quick-start.md
negz: Are we sure Crossplane works well with `kind` and/or `k3s`? I see we have a microk8s script under `cluster/local/` but I've never tried anything beside minikube. I imagine it would work, but I'd like to know for sure before saying so in the guide. 😬

docs/quick-start.md
negz: Reworded to at least imply there are other cloud provider stacks. We do mention there are others, but not until the end of the guide.

docs/quick-start.md
negz: Thanks - I feel like the simplicity outweighs the potential automation of scripted examples.

docs/quick-start.md
negz: I think I had this worded slightly awkwardly because at some point it made the line endings align more pleasingly. :|

docs/quick-start.md
negz: We don't actually have an exact percentage requirement, though we may as part of https://github.com/crossplaneio/crossplane/issues/933.

What I want to say here is that v1beta1 resources support _all_ of their underlying cloud provider resource's configuration. The catch is that there are a few edge cases such that we support 98/100 configuration options, so I can't truthfully say that. A "large portion" or "majority" feels too vague to me - I could read that as supporting 2/3rds of what CloudSQL could do when in reality we support almost all of what CloudSQL can do.

docs/quick-start.md
negz: Funnily enough it was intuition to write the sentence that way originally, but I tend to overuse commas and decided to avoid one in this case. I just checked your suggestion using a few grammar apps and the agree with you - fixed.

docs/quick-start.md
negz: I ended up just using the terminology we use - "high fidelity". I feel like this is a common enough term that it will communicate what I intended.

docs/quick-start.md
negz: I meant to write provides. Provisions means more like "GCP creates MySQL databases", while I meant literally "GCP provides you with MySQL databases via CloudSQL", which is also true.

docs/quick-start.md
negz: My intention was to communicate that we use a "managed resource and claim" pattern here, in that we let you create managed resources but then you must claim them. I think that's more appropriate here in the static provisioning section.

docs/quick-start.md
negz: I went back and forth on this, but ended up deciding to take a gamble that a large enough portion of our target audience will know what `credentials.json` is (or will just be skimming the guide and will get the gist from the name), so I figured I'd lead with the example then provide context and links immediately afterward in case folks weren't familiar with the concepts.

docs/quick-start.md
negz: I see your point. I didn't go with your suggestion, but I played around with the wording of the headings and this "table of contents" (which is supposed to be more like a list of tasks that happens to link to titles, than a true ToC) to try to come up with something that is:

1. Succinct
1. Reasonably faithful to our terminology and concepts
1. Not so heavy on out terminology that it's meaningless to folks who don't already know it

PR #986, opened 2019-10-26
Small docs landing and getting started updates

negz:
### Checklist

I have:

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,7 +1,7 @@
 # Welcome to Crossplane!
 
 Crossplane is an open source multicloud control plane to manage your
-cloud-native infrastructure and applications across environments, clusters,
+cloud-native applications and infrastructure across environments, clusters,
 regions and clouds. It enables provisioning and full-lifecycle management of
 applications and managed services from your choice of cloud using `kubectl`.
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -75,7 +75,7 @@ to further isolate the blast radius of applications.
 Infra Stacks are pre-built and published to the [Stacks
 registry][stack-registry] where they can be installed by a Cluster Admin using a
 [`ClusterStackInstall`][stack-install-docs] Kubernetes API kind or with the
-[`stack install`][crossplane-cli-usage] command that installs a Stacks with
+[`stack install`][crossplane-cli-usage] command that installs a Stack with
 permissions to watch resources at a cluster level, across all namespaces.
 
 Infra Stacks are covered in both the [Crossplane Services
--- a/docs/README.md
+++ b/docs/README.md
@@ -2,8 +2,8 @@
 
 Crossplane is an open source multicloud control plane to manage your
 cloud-native applications and infrastructure across environments, clusters,
-regions and clouds. It enables provisioning and full-lifecycle management
- of applications and managed services from your choice of cloud using `kubectl`.
+regions and clouds. It enables provisioning and full-lifecycle management of
+applications and managed services from your choice of cloud using `kubectl`.
 
 Crossplane can be installed into an existing Kubernetes cluster to add managed
 service provisioning or deployed as a dedicated control plane for multi-cluster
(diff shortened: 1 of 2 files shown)


negz:
> Crossplane is an open source multicloud control plane to manage your cloud-native applications and infrastructure

I went back and forth on whether this should say "infrastructure and applications" rather than "applications and infrastructure". I feel like it could make sense to lead with infrastructure since I feel Services is currently quite a bit more mature than Workload - we lead with Crossplane Services when we discuss our feature areas. I left it as "applications and infrastructure" because it reads better (and because it's possible I'm just biased about Crossplane Services).

docs/README.md
```diff
@@ -2,137 +2,139 @@
...
 
 Crossplane can be installed into an existing Kubernetes cluster to add managed
 service provisioning or deployed as a dedicated control plane for multi-cluster
 management and workload scheduling.
 
 Crossplane enables the community to build and publish Stacks to add more clouds
 and cloud services to Crossplane with support for out-of-tree extensibility and
-independent release schedules. Crossplane includes Stacks for [GCP][stack-gcp], 
+independent release schedules. Crossplane includes Stacks for [GCP][stack-gcp],
 [AWS][stack-aws], and [Azure][stack-azure] today.
```
hasheddan: Should we add `stack-rook` here? Might also be nice to add something like: "There are also community driven stacks that are owned by other organizations, such as [`stack-packet`](https://github.com/packethost/stack-packet)."

docs/README.md
```diff
@@ -2,137 +2,139 @@
...
 
-The [Crossplane Services Guide][services-user-guide] explores related concepts and
-shows how to use Crossplane Services to deploy a Wordpress instance that
-securely consumes a managed `MySQLInstance` from GCP, AWS, or Azure from
-`kubectl`.
+Take a look at the [getting started guide][getting-started] for an introduction
+to Crossplane Services. The [Crossplane Services Guide][services-user-guide]
+also explores related concepts and shows how to use Crossplane Services to
+deploy a Wordpress instance that securely consumes a managed `MySQLInstance`
+from GCP, AWS, or Azure using `kubectl`.
```
hasheddan: I still feel like our distinction between the stacks and services guides are somewhat hazy. To me, the difference is that in the services guide you bring your k8s cluster that you will run the application in and in stacks you provision it (i.e. make use of `KubernetesApplication`). Not sure if we can make that clearer here.

docs/README.md
```diff
@@ -2,137 +2,139 @@
...
-registry][stack-registry] where they can
-be installed by a Cluster Admin using a
-[`ClusterStackInstall`][stack-install-docs]
-Kubernetes API kind or with the [`stack
-install`][crossplane-cli-usage] command that
-installs a Stacks with permissions to watch resources at a cluster level, across
-all namespaces.
+registry][stack-registry] where they can be installed by a Cluster Admin using a
+[`ClusterStackInstall`][stack-install-docs] Kubernetes API kind or with the
+[`stack install`][crossplane-cli-usage] command that installs a Stacks with
```
hasheddan:
```suggestion
[`stack install`][crossplane-cli-usage] command that installs a Stack with
```

docs/README.md
```diff
@@ -2,137 +2,139 @@
...
-Kubernetes clusters from a single control plane with consistent multi-cluster
-best-practice configuration and secure connectivity between target Kubernetes
-clusters and the managed services provisioned for applications. Managed
-Kubernetes clusters can be dynamically provisioned with a `KubernetesCluster`
+## Crossplane Workloads
+
+Crossplane includes an extensible workload scheduler that observes application
+policies to select a suitable target cluster from a pool of available clusters.
+The workload scheduler can be customized to consider a number of criteria
+including capabilities, availability, reliability, cost, regions, and
```
hasheddan: For the first time I actually feel like this is `TRUE`. All of these features can be enabled by label-based scheduling. I do think this sentence makes it sound like this process is a little bit more intelligent than it is at the moment, but I feel better about it than I have in the past.

docs/README.md
negz: I think so.

In this area I was also thinking it feels a little weird to call out that we support "out-of-tree extensibility". Seems like something that anyone who was around when we _didn't_ support that might be interested in, but that is kind of a given in context now.

PR #1002, opened 2019-10-28
docs: update crossplane cli to v0.2.0

suskin:

Now that we have a new version of the crossplane cli, and we're
referencing the `trace` command that it introduced, we should also tell
readers to install that version.

[skip ci]

### DIFF
--- a/docs/stacks-guide-gcp.md
+++ b/docs/stacks-guide-gcp.md
@@ -399,7 +399,7 @@ where we left off.
 
 <!-- Links -->
 [crossplane-concepts]: concepts.md
-[crossplane-cli]: https://github.com/crossplaneio/crossplane-cli/tree/release-0.1
+[crossplane-cli]: https://github.com/crossplaneio/crossplane-cli/tree/release-0.2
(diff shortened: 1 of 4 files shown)


docs/stacks-guide-gcp.md
```diff
@@ -421,4 +421,4 @@ where we left off.
 [gcp-ip-address]: https://cloud.google.com/compute/docs/ip-addresses/
 [gcp-connection]: https://cloud.google.com/vpc/docs/configure-private-services-access
 [resource class selection]: https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-simple-class-selection.md
-[gcp-provider-guide]: cloud-providers/gcp/gcp-provider.md
\ No newline at end of file
+[gcp-provider-guide]: cloud-providers/gcp/gcp-provider.md
```
hasheddan: Any ideas why this is showing a diff?

suskin: No o: . Must have something to do with whitespace as this change disappears when I hide whitespace changes, but I'm not sure what exactly.

PR #1008, opened 2019-10-30
Stack managed ClusterRoles to support self-service

lukeweber:
Stack manager and by extension Crossplane currently assumes that RBAC will be configured by an admin as an experience largely left to the reader, but each stack-install changes the available CRDs that a particular persona need to access, thus creating work and friction to our extension story for stacks. For a truly self-service model, a person who installs a stack, should then be able to consume the resources that were installed as part of the machinery of Crossplane.

Kubernetes provides the concept of built-in roles, like cluster-admin, admin, edit, view, that provides a clear role to bind to for expected behavior and I think Crossplane should provide the same to make secure integration with crossplane even easier.

Cluster / Namespace scoped stacks + stack-manager provides a very clean way to implement a concise consumption and permissioning model around all crossplane installed resources using automatic role aggregation provided by Kubernetes.

Additional updates were made to reflect that we no longer have multiple environments in different namespaces, as infrastructure scoped resources have moved to the Cluster scope. In the same respect, Cluster stacks and their installation should no longer be namespaced, instead living in the cluster scope as well as part of the infrastructure.

### Checklist

I have:


design/one-pager-stacks-security-isolation.md
```diff
@@ -241,13 +235,134 @@ permissionScope: Cluster
...
-A simple way to approach this would be to create different installation request CRDs for each and then use RBAC to restrict access to them by different roles.
+Since cluster stacks and namespace stacks have differing scopes of execution in the control plane, it will be wise to
+ allow differing access to request their installation. A simple way to approach this would be to create different
+ installation request CRDs for each and then use RBAC to restrict access to them by different roles.
+
 Currently, we use a single `StackRequest` type, but going forward we should differentiate by stack type:
 
-* `ClusterStackInstall`: requests the installation of a cluster stack
-* `StackInstall`: requests the installation of a namespace stack
+* `ClusterStackInstall`: requests the installation of a cluster stack who's resources are all cluster scoped
```
prasek: > who's resource are all clusters scoped

So this translates into the Stacks Manager only needing to create / update `ClusterRole` definitions for the environment admin, editor, viewer roles, as part of processing a `ClusterStackInstall`?

design/one-pager-stacks-security-isolation.md
```diff
@@ -241,13 +235,134 @@ permissionScope: Cluster
...
+* `StackInstall`: requests the installation of a namespace stack who's resources are all namespace scoped
```
prasek: > who's resource are all namespace scoped

So the Stacks Manager would only need to create / update `Role` definitions for namespace (workspace) admin, editor, viewer roles - that could be used by a client to assign permissions to
 * workspace admin role: install / upgrade app stacks via StackInstall within a namespace (workspace) + below
 * workspace editor role: manage CRs for installed app stacks -- separate from the `Role` granting access to create resource claims in the namespace (workspace)
 * workspace viewer role: view CRs for installed app stacks -- separate from the `Role` granting access to view resource claims, and the `ClusterRole` granting access to view the cluster-scoped managed resource instance

design/one-pager-stacks-security-isolation.md
```diff
@@ -241,13 +235,134 @@ permissionScope: Cluster
...
+
+With these two separate types, an administrator can allow access to installing cluster stacks to only certain restricted
+ roles, then perhaps be more permissive with access to installing namespaced stacks.
+
+If the install scope type does not match what the stack author has declared within their stack metadata, an error status
```
prasek: What in the stack metadata will identify it as a cluster-scoped or namespace-scoped stack? Inferred from stack metadata that says owned stack CRDs should be added to a `Role` or `ClusterRole` for each of the supported roles (admin, editor, viewer)?

design/one-pager-stacks-security-isolation.md
```diff
@@ -241,13 +235,134 @@ permissionScope: Cluster
...
+ should be set on the install resource and installation of the stack should not proceed.
+
+### RBAC
+
+We will rely on RBAC to further define a clear separation of concerns for infrastructure resources, and workspace
+ resources in a namespace. In our model we assume that any Cluster Stack resources as they are global will be for the
```
prasek: perhaps it's simpler to say that Crossplane will provide `Role` and `ClusterRole` definitions for environment and workspace roles (admin, editor, viewer) to simplify RBAC configuration with user-defined `User`, `Group`, `RoleBinding` and `ClusterRoleBinding` resources.

design/one-pager-stacks-security-isolation.md
```diff
@@ -241,13 +235,134 @@ permissionScope: Cluster
...
+ infrastructure persona, whereas any namespace stack resources will be for the namespace that that stack is installed
+ into, thus the application persona.
+
+At install time:
+ - Cluster Stack: generate cluster roles for infra persona that auto-aggregate to built-in roles for
```
prasek:
```suggestion
 - Cluster Stack: generate cluster roles that auto-aggregate to well-known `ClusterRole` definitions for environment admin, editor, and viewer roles
```

design/one-pager-stacks-security-isolation.md
```diff
@@ -241,13 +235,134 @@ permissionScope: Cluster
...
+ infrastructure owners.
+ - Namespaced Stack: generate cluster roles for the workspace persona different permission
```
prasek:
```suggestion
 - Namespaced Stack: generate `Role` definitions for workspace admins, editor, and viewer roles
```
Unless you're suggesting that everyone becomes cluster-scoped, including namespace stacks? Looks like that is the case, and would definitely let us use the aggregated ClusterRoles, since aggregated Roles don't seem to be supported AFAIK.

design/one-pager-stacks-security-isolation.md
```diff
@@ -241,13 +235,134 @@ permissionScope: Cluster
...
+The following built in roles will be installed and packaged with Crossplane, and updated with role aggregation as we
+ install additional stacks.
+
+The verbs supported in these roles(admin,edit,view) will model based on existing logic around
+[built-in kubernetes roles](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#default-roles-and-role-bindings).
+
+|ClusterRole Name | Permissions |
+|:-------|:----------|
+| crossplane:admin | Admin: environment, workspaces(namespaces), ClusterRole, Role, ClusterRoleBinding, RoleBinding |
+| crossplane:infra-admin | Admin: environment, manage cluster stacks |
```
prasek:
```suggestion
| crossplane:env-admin | Admin: environment, manage cluster stacks |
```

to keep the # of terms down

design/one-pager-stacks-security-isolation.md
```diff
@@ -241,13 +235,134 @@ permissionScope: Cluster
...
+kind: RoleBinding
+metadata:
+  name: admin-workspace
+  namespace: foo
+subjects:
+- kind: User
+  name: jane # Name is case sensitive
+  apiGroup: rbac.authorization.k8s.io
+roleRef:
+  kind: ClusterRole
```
prasek: interesting, so if the Stacks Manager created `ClusterRole` definitions for each Namespace Stack, and they were aggregated up to the well-known / build-in workspace-admin `ClusterRole`, then they a single `RoleBinding` could be used to grant access to all Namespace Stack CRDs in the cluster. Is that the behavior we're after? Sounds like we'd need a way to filter the CRDs that have been installed into a given namespace to help ensure there's an app stack controller listening in that namespace.

design/one-pager-stacks-security-isolation.md
```diff
@@ -241,13 +235,134 @@ permissionScope: Cluster
...
+  name: crossplane:foo:workspace-admin
+  apiGroup: rbac.authorization.k8s.io
+```
 
-If the install scope type does not match what the stack author has declared within their stack metadata, an error status should be set on the install resource and installation of the stack should not proceed.
+#### Managing CRD Permissions of Dynamically (Aggregation)
+
+As we install new stacks and new CRDs we will need to grant our built in roles access to delete, edit, create, and view
+ the given newly installed resources, and we can use the built in
+ [aggregated cluster roles](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#aggregated-clusterroles)
```
prasek: @lukeweber it looks like aggregated roles only work for cluster roles, so for workspace roles we'd need to create additional role binding for each stacks, or it looks like you're proposing `ClusterRole` definitions would be created for Namespace Stacks as well?

design/one-pager-stacks-security-isolation.md
```diff
@@ -241,13 +235,134 @@ permissionScope: Cluster
...
+```
+
+### Discoverability of available Crossplane CRDs for a given workspace or environment
+
+Because CRDs are cluster scoped, but permissions of workspaces and their stack-installs are workspace scoped we will
+ need to label them to indicate which ones are appropriate for a given context. As the stack-installer processes a
+ cluster or namespace stack install it needs to add labels indicating that it's crossplane environment scoped or
+ workspace scoped.
+
+A crossplane-infra-admin could then fetch environment crd types
```
prasek:
```suggestion
A crossplane-env-admin could then fetch environment crd types
```

design/one-pager-stacks-security-isolation.md
lukeweber: Cluster role aggregation doesn’t exist for just roles. It may be operationally simpler to model them all as cluster roles, especially since binding a clusterRole with a rolebinding creates the same effect as if it was in a role.

If we do roles we would have to manage them without aggregation tooling from kubernetes.

design/one-pager-stacks-security-isolation.md
lukeweber: All roles become cluster scoped, but since we get aggregation(main reason), and permissions for a workspace are a function of admin and stack installer this seems ok. Workspace clusterroles would then be bound to their respective namespace with a roleBinding. I could kind of go either way on this. I’m tipped slightly to use kubernetes tooling for aggregation of clusterroles for its simplicity, but happy to be overruled if you prefer roles inside the workspace.

I think all admin level roles should roll up to crossplane-admin for example, regardless of if they’re namespace or cluster scoped.

design/one-pager-stacks-security-isolation.md
lukeweber: Yes the admin concept is to make sure they can see everything. Due to our model it might make sense that crossplane admin only gets viewer rights  on crossplane namespaced stacks to avoid writing stuff in the wrong place.

lukeweber: Yes the admin concept is to make sure they can see everything. Due to our model it might make sense that crossplane admin only gets viewer rights  on crossplane namespaced stacks to avoid writing stuff in the wrong place.

design/one-pager-stacks-security-isolation.md
lukeweber: ClusterRoles for every stack was this proposal. Still considering the implications of which you’ve surfaced a few in this doc. Maybe it’s not the right approach, but was admittedly stretching the previous aggregation idea to its limits to see what it looked like.

design/one-pager-stacks-security-isolation.md
```diff
@@ -241,13 +246,212 @@ permissionScope: Cluster
...
+```
+
+On install of crossplane-aws stack, the stack-manager would create a ClusterRoles for each built in role. In the case of
+ crossplane-admin it would contain full access to all CRDs in the stack and look like the following.
+
+```yaml
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: cossplane:crossplane:stack-aws:1.0 # Unique by namespace:stack-name:version
```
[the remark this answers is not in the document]
lukeweber: services, pods, namespaces, and many kubernetes resources are dns resolvable by their given name as part of kubernetes machinery - https://github.com/kubernetes/community/blob/master/contributors/design-proposals/architecture/identifiers.md

However internal objects in kubernetes seem to be more relaxed about naming. I can't find documentation on what the rules around this are. 

As a note, because ":" are allowed in a ClusterRole, but not in namespaces, by using a ":" you can be assured by convention that a malicious actor couldn't collide with your naming scheme.

As a note, looking at kubernetes roles the user facing ones don't include ":" like view, edit, admin, cluster-admin. 

We might consider following this convention for our aggregated, but user facing roles:
```
crossplane-admin
crossplane-env-admin
crossplane-env-edit
crossplane-env-view
crossplane-ns-foo-admin
...
```

``` 
> kubectl -n kube-system create clusterrole crossplane:foo:baz --verb=get --resource=pod
clusterrole "crossplane:foo:baz" created
> kubectl -n kube-system get clusterrole crossplane:foo:baz -o yaml
kind: ClusterRole
metadata:
  creationTimestamp: 2019-10-31T18:15:42Z
  name: crossplane:foo:baz
  resourceVersion: "361976"
  selfLink: /apis/rbac.authorization.k8s.io/v1/clusterroles/crossplane%3Afoo%3Abaz
  uid: a16a0ef9-6dab-4905-984b-73eb5149940b
rules:
- apiGroups:
  - ""
  resources:
  - pods
  verbs:
  - get
```

design/one-pager-stacks-security-isolation.md
displague: I think even for a namespaced StackInstall, environment admins, editors, and viewers should have access to interact with the namespaced resources where ever they are found.  Clusterroles for the stack installed types should aggregate to the roles used by these environment personas.

design/one-pager-stacks-security-isolation.md
displague: Within the `app.yaml` there is a permissionScope field. This is either `Cluster` or `Namespaced` and the StackManager verifies that the appropriate installation type was used (StackInstall, ClusterStackInstall).

design/one-pager-stacks-security-isolation.md
```diff
@@ -56,6 +56,33 @@ From the [Kubernetes documentation](https://kubernetes.io/docs/tasks/access-kube
...
 
+#### Discoverability of Scope
+
+Because CRDs are cluster scoped, but our permission model for stacks is both namespaced and global we will
+ need to label CRDs to indicate which are available in a given context. As the stack-installer processes a
+ cluster or namespace stack install it needs to add labels indicating that it's crossplane environment scoped or
+ namespace scoped to a given namespace.
+
+Labels on a CRD:
+    crossplane.{scope=namespace|environment}=true
```
displague: > In our model we assume that any Cluster Stack resources are global will be for the
 environment or infrastructure scope, whereas any namespace stack resources will be for the namespace that that stack is installed into.

Why do we need this label? I believe we've ruled out any scenario where a CRD scope does not match it's intended Crossplane scope.

Namespaced scope CRD are for namespaces
Cluster scoped CRDs are for clusters (environment)

(Contingent on https://github.com/crossplaneio/crossplane/pull/1023)

design/one-pager-stacks-security-isolation.md
lukeweber: I think it would be good to clear this up. When I hear separation of concerns, I hear two different groups of people without overlap, thus separation. I think if cluster-admin wants to make someone a crossplane-admin they would have access to admin everything, but an environment-admin would only be able to admin the environment or infrastructure related resources and config.

We haven't specified if cluster-admin will get aggregated roles from the namespace stack installs, but since cluster-admin will be cluster scoped, this would potentially allow a namespaced stack to grant the cluster-admin to write the newly installed namespace scoped resource into any namespace, which would be unideal.

design/one-pager-stacks-security-isolation.md
displague: fwiw, I wanted to verify that a clusterrole, restricted to a namespaced rolebinding, did restrict access to that kind from outside namespaces:
 
```
$ kubectl create clusterrole see-pods --verb=list --resource=pods
$ kubectl create rolebinding podseersight --clusterrole=see-pods --group=pod-seers
$ kubectl get pods -n crossplane-system --as-group=pod-seers --as=cluster-admin
Error from server (Forbidden): pods is forbidden: User "cluster-admin" cannot list resource "pods" in API group "" in the namespace "crossplane-system"
```

Looks good :+1:

design/one-pager-stacks-security-isolation.md
```diff
@@ -241,13 +277,209 @@ permissionScope: Cluster
...
+ * `{namespace-name}: "true"`
+
+Match Roles:
+ * `rbac.crossplane.io/aggregate-to-{install-scope}-{role}: "true"`
+ * `{namespace-name}: "true" // to match specific namespace.`
+
+#### ClusterRoles for a given (Cluster)Stack Install
+
+For a stack given stack version it will have a unique set of roles in the format
+ {repo-name}:{stack-name}:{version}:{role=system|admin|edit|view}. System role is for internal use only and will be
```
displague: Where does `repo-name` come from?  A stack is currently aware of the app.yaml `Company`, and `Source` (registry.hub.docker.com, registry.crossplane.io)?  What should be used when source was omitted (deferring to local images and the default registry)?

displague: Special precaution will need to ensure that this is not set to `system`.

Would `stack` suffice as a prefix?

displague: Two stacks with the same name from different sources would collide on their cluster role if that were the case. hmm

design/one-pager-stacks-security-isolation.md
```diff
@@ -241,13 +277,209 @@ permissionScope: Cluster
...
+ bound by the controller for the stack, and will include the set of dependent resources for a given stack.
+
+
+Consider the following example for a wordpress stack of version 1.1 from wordpressinc, you would have the following set.
+ * wordpressinc:wordpress:1.1:view:system // Used by wordpress operator - stack resources + dependent resources grants
```
displague: Did you unintentionally insert "view:" into all of these roles?

design/one-pager-stacks-security-isolation.md
displague: If the name is based on the mutable `Source`, we'll have to be careful about updating the roles to avoid orphaning them.

design/one-pager-stacks-security-isolation.md
lukeweber: scope is not a label that can be searched on, i.e

```
kubectl get crds --selector=crossplane.scope=environment
```

design/one-pager-stacks-security-isolation.md
displague: > this would potentially allow a namespaced stack to grant the cluster-admin to write the newly installed namespace scoped resource into any namespace, which would be unideal

Can't the cluster-admin do that anyway? If they want to create a resource in a namespace where no controller will act on them - can't they do that, if they want?

design/one-pager-stacks-security-isolation.md
```diff
@@ -241,13 +277,209 @@ permissionScope: Cluster
...
+```
+
+On install of crossplane-aws stack, the stack-manager would create a ClusterRoles for each built in role. In the case of
+ crossplane-admin it would contain full access to all CRDs in the stack and look like the following.
+
+```yaml
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: cossplane:crossplane:stack-aws:1.0:admin # Unique by crossplane:namespace:stack-name:version:role
```
displague: "crossplane" although cos'play'n is an interesting spin.

displague: Is this format different than the format defined earlier? 
`{repo}:{stack name}:{version}:{role}`?

I think we may want to specify that stack name is the name of the stack (stack-wordpress), not the name of the namespaced stack install / stack resource (my-stack-wordpress).

I don't believe this name is exposed by the stack resource, so unless we expose that somehow we are going to be using the local stack name and won't be consolidating ClusterRoles.

You included version number in the ClusterRole name, perhaps this is a hint that you were already thinking that the ClusterRoles consolidation may not be desirable.  If the namespaced stack's ClusterRole is based on the name of the StackInstall, then perhaps version is not needed.  We already have a unique ClusterRole name without version.

design/one-pager-stacks-security-isolation.md
lukeweber: Good idea on the prefix. Definitely needed. Any recommendation on what you would do with Source?

If a CRD that collides is installed from a different stack, is that a failure? 

Do we handle repo-name/stack-name as needing to be unique per crossplane, regardless of source, do we need to prefix everything with its source? Really open to ideas here.

design/one-pager-stacks-security-isolation.md
```diff
@@ -56,6 +56,33 @@ From the [Kubernetes documentation](https://kubernetes.io/docs/tasks/access-kube
 Therefore, any time a CRD is created, it will also be at the cluster scope, even if instances of the CRD are scoped to a namespace.
 This has caused confusion in the past, so it's important to clear that up here.
 
+#### Discoverability of Scope
+
+Because CRDs are cluster scoped, but our permission model for stacks is both namespaced and global we will
+ need to label CRDs to indicate which are available in a given context. As the stack-installer processes a
+ cluster or namespace stack install it needs to add labels indicating that it's crossplane environment scoped or
+ namespace scoped to a given namespace.
```
negz:
```
$ kubectl get -o json crd clusterstackinstalls.stacks.crossplane.io|jq .spec.scope
"Namespaced"
```

CRDs already encode this data as `.spec.scope`. Is the idea that we want to duplicate it as a label because we can't list filtered by `.spec.scope` (it's not a supported field selector)? That may be worthwhile, especially if we expect folks to want to list CRDs by scope using `kubectl`, but if we expect this list to be aggregated by some other UX it may be worth simply having the frontend list all CRDs and filter them manually.

design/one-pager-stacks-security-isolation.md
```diff
@@ -56,6 +56,33 @@ From the [Kubernetes documentation](https://kubernetes.io/docs/tasks/access-kube
...
+
+Labels on a CRD:
+  * crossplane.{scope=namespace|environment}=true
```
negz: Nitpick on the labelling, `crossplane.io/scope=namespace` and `crossplane.io/scope=environment` would be more consistent with our current label and annotation conventions, which include `crossplane.io/external-name=foo` and `resourceclass.crossplane.io/is-default-class="true"`.

design/one-pager-stacks-security-isolation.md
```diff
@@ -95,11 +122,14 @@ Let's look at a specific example of the [GCP cluster stack](https://github.com/c
...
+ `crossplane.environment=true` to denote that they are from a cluster scoped install and for the environment.
+* A `ClusterRole` is created for for system that includes the RBAC rules for the [allowed core Kubernetes kinds](#allowed-resource-access), the stack's owned CRDs, and the CRDs that it depends on.
+* `ClusterRoles` for admin, edit, and view will be created in the cluster scope for the stacks owned CRDs.
 * A `ServiceAccount` is created in the namespace that the stack's controller will be running.  This will serve as the runtime identity of the controller.
-* A `ClusterRoleBinding` is created at the cluster level to bind the `ClusterRole` and the namespaced `ServiceAccount`
-* A `Deployment` is created in the same namespace as the stack and it is expected to launch and manage the pod that hosts the GCP controllers.
+* A `ClusterRoleBinding` is created at the cluster level to bind the system `ClusterRole` and the namespaced `ServiceAccount`
+* A `Deployment` is created in the crossplane-system namespace and it is expected to launch and manage the pod that hosts the GCP controllers.
```
negz: Is there a reason we need to specifically call out `crossplane-system` here? Currently that namespace is by convention the place we install Crossplane and its stacks, rather than something we hard-code or require anywhere.

design/one-pager-stacks-security-isolation.md
```diff
@@ -109,57 +139,56 @@ The diagram below summarizes the artifacts in the control plane after the GCP st
...
 
-### Environments
+### Environment
 
-While cluster stacks are installed at the cluster level (there is only one instance of each cluster stack), they can still support the concept of multiple "environments", each represented by a namespace.
-For example, an admin can create a namespace to model their production "environment" and another namespace to model their development "environment".
-All resources within each of those environments can share a common set of provider credentials.
-This namespaced "environment" approach provides the administrator with a way to collect or aggregate all the infrastructure and services associated with a given environment.
+Cluster stacks are installed at the cluster level (there is only one instance of each cluster stack), and there will
+ be a single environment for infrastructure which will be represented by the crossplane-system namespace as well as
```
negz: > will be represented by the crossplane-system namespace as well as cluster scoped resources

There's nothing special about `crossplane-system` as far as any Crossplane logic is concerned; I don't think we can assume it's a special infrastructure namespace. Today:

* Providers, resource classes, and managed resources are cluster scoped.
* Provider and managed resource secrets are written _by convention_ to `crossplane-system`.
* The use of `crossplane-system` as a place to read and write secrets is configurable by the infrastructure admin.

```yaml
apiVersion: gcp.crossplane.io/v1alpha3
kind: Provider
metadata:
  name: example
spec:
  credentialsSecretRef:
    # If you change this the provider will read its secret from the namespace
    # you specify. This is a required field; there is no default.
    namespace: crossplane-system
    name: example-provider-gcp
    key: credentials.json
  projectID: PROJECT_ID
```

```yaml
kind: CloudSQLInstanceClass
metadata:
  name: standard-mysql
specTemplate:
  # If you change this any managed resource dynamically provisioned
  # using this class will write its connection secrets to the namespace
  # you specify. This is a required field; there is no default.
  forProvider:
    databaseVersion: MYSQL_5_7
    region: us-central1
    settings:
      tier: db-n1-standard-1
      dataDiskType: PD_SSD
      dataDiskSizeGb: 10
      ipConfiguration:
        privateNetworkRef:
          name: sample-network
  providerRef:
    name: example
```

```yaml
kind: CloudSQLInstance
metadata:
  name: standard-mysql-negz
spec:
  writeConnectionSecretToRef:
    # If you change this the managed resource will write its connection
    # secret to the namespace you specify. This is a required field; there
    # is no default.
    namespace: crossplane-system
    name: standard-mysql-negz
  forProvider:
    databaseVersion: MYSQL_5_7
    region: us-central1
    settings:
      tier: db-n1-standard-1
      dataDiskType: PD_SSD
      dataDiskSizeGb: 10
      ipConfiguration:
        privateNetworkRef:
          name: sample-network
  providerRef:
    name: example
```

design/one-pager-stacks-security-isolation.md
```diff
@@ -241,13 +277,255 @@ permissionScope: Cluster
...
+Example running a command as crossplane-admin
+```
+kubectl {any crossplane-admin operation} --as-group=crossplane:master
+```
+
+For discoverability of the roles for the environment we will add the following labels:
+ * `rbac.crossplane.io/role-environment: "true"`
+
+Match labels on these roles for auto-aggregation:
+ * `rbac.crossplane.io/aggregate-to-crossplane-admin: "truse"`
```
negz:
```suggestion
 * `rbac.crossplane.io/aggregate-to-crossplane-admin: "true"`
```

design/one-pager-stacks-security-isolation.md
```diff
@@ -241,13 +277,255 @@ permissionScope: Cluster
...
+ * `rbac.crossplane.io/role-namespace: "true"`
+ * `{namespace-name}: "true"`
+
+Match Roles:
+ * `rbac.crossplane.io/aggregate-to-{install-scope}-{role}: "true"`
+ * `{namespace-name}: "true" // to match specific namespace.`
+
+#### ClusterRoles for a given (Cluster)Stack Install
+
+For a stack given stack version the stack-manager will create and or label a unique set of roles in the format
```
negz: This confused me at first because I thought the suggested format was how these roles would be labelled, but I think it's intended to be how they're named. It could be worth rewording to clarify.

design/one-pager-stacks-security-isolation.md
```diff
@@ -241,13 +277,255 @@ permissionScope: Cluster
...
+ * stack:wordpressinc:wordpress:1.1:edit
+ * stack:wordpressinc:wordpress:1.1:view
+
+If these were a from a cluster scoped stack, they would be aggregated to the built in environment roles, and if they
+ were from a namespaced stack install they would be aggregated to the corresponding [ClusterRoles for a given app
+ namespace](#clusterroles-for-a-given-app-namespace).
+
+Labels for these roles will be for aggregation purposes:
+ * rbac.crossplane.io/aggregate-to-{install-scope}-{role}: "true"
+ * {namespace-name}: "true" // Only included on namespaced stacks install
```
negz: This scheme for the namespace label is a little unintuitive; it would be nice if the reader could tell that the key was a namespace name. It seems we need the namespace in the key (not value) because we want to explicitly whitelist (using labels) the names of multiple namespaces that a `ClusterRole` will be allowed to aggregate up to?

lukeweber: Namespaces have a limit of 63 characters, as do labels. I'm not sure about internals of labels, but kubernetes seems to so weird stuff like aggregate-to-view: true, as opposed to aggregate-to: view. I assume maybe the key is indexed and the value is compared, but that's just a guess and what I was following when I did this. `namespace: foo` seems clearer, but not sure if it takes a hit on efficiency, especially considering how many we might have.

design/one-pager-stacks-security-isolation.md
lukeweber: This would require maybe making a clusterStackInstall have it's controller location be configurable instead of implied from the namespace, which has gone away now that it's going ClusterScoped.

design/one-pager-stacks-security-isolation.md
lukeweber: Maybe we should have a quick sync about final wording on all of this stuff. I'm not particularly concerned about choosing one convention over the other, and would agree consistency is probably the best.

design/one-pager-stacks-security-isolation.md
lukeweber: > A general theme; I'm not sure we should leak the Upbound internal concept of an "environment" into this Crossplane design document. Crossplane doesn't currently (as far as I know) have the concept of an environment.

This was pre-existing in this doc in master so I think it leaked into UbC :) If we want to remove, I guess I'll defer to you guys.

lukeweber:
> A general theme; I'm not sure we should leak the Upbound internal concept of an "environment" into this Crossplane design document. Crossplane doesn't currently (as far as I know) have the concept of an environment.

This was pre-existing in this doc in master so I think it leaked into UbC :) If we want to remove, I guess I'll defer to you guys.

lukeweber:
@negz - Many thanks for the great feedback.  I've made an attempt to address feedback, but I'll summarize all the labels.

Namespace identifier - Because as you noted, this is used in aggregation and we can have multiple namespaces on a stack ClusterRole for aggregation purposes, we need to encode the key/value of namespace in the key - `namespace.crossplane.io/{namespace-name}: "true"`

Scope to identify ClusterRoles and CRDs as being part of environment or namespace scope - `crossplane.io/scope: "{environment|namespace}"` - If namespace, a namespace label will be included. 

Aggregate-to labels used by ClusterRole aggregation built in of format: `rbac.crossplane.io/aggregate-to-{scope}-{role}: "true"` - If scope=namespace, a namespace label will be included.

design/one-pager-stacks-security-isolation.md
lukeweber: I've updated this to reflect that it will be installed as defined in the Spec of the ClusterStackInstall. I'll leave that up to an implementation detail as to which pattern is chosen.

design/one-pager-stacks-security-isolation.md
lukeweber: reworded.

For a stack given stack version the stack-manager will create a unique set of roles in the format stack:{repo-name}:{stack-name}:{version}:{role=system|admin|edit|view}.

design/one-pager-stacks-security-isolation.md
displague: > This would require maybe making a clusterStackInstall have it's controller location be configurable instead of implied from the namespace

https://github.com/crossplaneio/crossplane/pull/1023/files#diff-df082ec602125906f2f5c204029cf3f2R128-R135

design/one-pager-stacks-security-isolation.md
jbw976: I think an important feature here @negz is the ability to know **which** namespaces a CRD is "applicable" for, or in other words, which namespaces a stack has been installed into and therefore which namespaces a CRD has truly been intended to be used in.  e.g., the following query that gets all CRDs installed from stacks that are applicable to the `foo` namespace is not something you can do with the built in `.spec.scope` that CRDs have:

```
kubectl get crds --selector crossplane.io/scope=namespace,namespace.crossplane.io/foo=true
```

displague: I raised a similar concern here, https://github.com/crossplaneio/crossplane/pull/1008#discussion_r342233345

Yes, we can't filter the API fetched CRDs by `scope` (`kubectl` or otherwise) without some secondary filter.  Labels permit indexing and filtering for the Crossplane/Stack CRDs.

design/one-pager-stacks-security-isolation.md
```diff
@@ -109,57 +141,57 @@ The diagram below summarizes the artifacts in the control plane after the GCP st
 
 ![Diagram of a Cluster Stack Installed](./images/cluster-stack-installed.png)
```
jbw976: the 2 images of cluster stacks and namespaced stacks are out of date now. do you have new architecture diagrams that demonstrate the proposed way of organizing roles/controllers/CRDs/etc.?

design/one-pager-stacks-security-isolation.md
```diff
@@ -241,13 +281,258 @@ permissionScope: Cluster
...
+The following built in roles will be installed and packaged with Crossplane, and updated with role aggregation as we
+ install additional stacks.
+
+The verbs supported in these roles(admin,edit,view) will model behavior based on existing logic around
+ [built-in kubernetes roles](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#default-roles-and-role-bindings).
+
+|ClusterRole Name | Permissions |
+|:---------------------|:----------|
+| crossplane-admin | Admin: ClusterStacks, Crossplane CRDs, Namespaces, (Cluster)RoleBindings, (Cluster)Roles |
+| crossplane-env-admin | Admin: ClusterStacks, globally scoped Crossplane CRDs |
```
jbw976: "globally scoped Crossplane CRDs" == "globally scoped CRDs installed from cluster stacks"?

design/one-pager-stacks-security-isolation.md
```diff
@@ -241,13 +281,258 @@ permissionScope: Cluster
...
+| crossplane-env-edit | Edit: ClusterStacks, globally scoped Crossplane CRDs |
+| crossplane-env-view | View: Globally scoped Crossplane CRDs, Not permitted: secrets |
+
+We will provide a convenience group - crossplane:master - that binds to crossplane-admin with a
+ ClusterRoleBinding. This mirrors Kubernetes functionality for cluster-admin and system:master. At install time you will
+ then be able to impersonate a group that is bound to crossplane-admin.
+
+Example running a command as crossplane-admin
+```
+kubectl {any crossplane-admin operation} --as-group=crossplane:master
```
jbw976: i'm not familiar with how k8s enforces that the invoker of `kubectl` is allowed to impersonate as this group, but I trust that system has been well thought through upstream :)

design/one-pager-stacks-security-isolation.md
```diff
@@ -241,13 +281,258 @@ permissionScope: Cluster
...
+
+For discoverability of the roles for the environment we will add the following labels:
+ * `crossplane.io/scope: "environment"`
+
+Match labels on these roles for auto-aggregation:
+ * `rbac.crossplane.io/aggregate-to-crossplane-{install-scope}-{role}: "true"`
+
+#### ClusterRoles for a given app namespace.
+
+For any namespace that is annotated with `rbac.crossplane.io/managed-roles: true`, the stack-manager will install and manage
```
jbw976: who is expected to set and use this annotation? this is the first time it is introduced in this doc, and it's not clear who has the responsibility of setting it.

design/one-pager-stacks-security-isolation.md
```diff
@@ -241,13 +281,258 @@ permissionScope: Cluster
...
+ * `crossplane:ns:{ns-name}:admin`
+ * `crossplane:ns:{ns-name}:edit`
+ * `crossplane:ns:{ns-name}:view`
+
+For discoverability of the roles for a given namespace we will add the following labels:
+ * `crossplane.io/scope: "namespace"`
+ * `namespace.crossplane.io/{namespace-name}: true`
+
+Match Roles:
+ * `rbac.crossplane.io/aggregate-to-{install-scope}-{role}: "true"`
```
jbw976: same here, since this is talking about a namespaced role, this match role statement example could be filled in with the install-scope right here in this doc, right? it's not variable, is it?

should it just always be "namespace", i.e. the value of `crossplane.io/scope: "namespace"` just above?

design/one-pager-stacks-security-isolation.md
```diff
@@ -241,13 +281,258 @@ permissionScope: Cluster
...
+Example running a command as crossplane-admin
+```
+kubectl {any crossplane-admin operation} --as-group=crossplane:master
+```
+
+For discoverability of the roles for the environment we will add the following labels:
+ * `crossplane.io/scope: "environment"`
+
+Match labels on these roles for auto-aggregation:
+ * `rbac.crossplane.io/aggregate-to-crossplane-{install-scope}-{role}: "true"`
```
jbw976: since this is talking about a cluster/environment role, this match role statement example could be filled in with the `install-scope` right here in this doc, right? it's not variable, is it?

should it just always be "environment", i.e. the value of crossplane.io/scope: "environment" just above?

design/one-pager-stacks-security-isolation.md
```diff
@@ -241,13 +281,258 @@ permissionScope: Cluster
...
+ * `namespace.crossplane.io/{namespace-name}: true`
+
+Match Roles:
+ * `rbac.crossplane.io/aggregate-to-{install-scope}-{role}: "true"`
+ * `namespace.crossplane.io/{namespace-name}: true // to match specific namespace.`
+
+#### ClusterRoles for a given (Cluster)Stack Install
+
+For a stack given stack version the stack-manager will create a unique set of roles in the format
+ stack:{repo-name}:{stack-name}:{version}:{role=system|admin|edit|view}. All of these roles are for internal use only.
```
jbw976: could you use code blocks (back ticks) here for readability?

design/one-pager-stacks-security-isolation.md
```diff
@@ -241,13 +281,258 @@ permissionScope: Cluster
...
+ * `crossplane.io/scope: "namespace"`
+ * `namespace.crossplane.io/{namespace-name}: true`
+
+Match Roles:
+ * `rbac.crossplane.io/aggregate-to-{install-scope}-{role}: "true"`
+ * `namespace.crossplane.io/{namespace-name}: true // to match specific namespace.`
+
+#### ClusterRoles for a given (Cluster)Stack Install
+
+For a stack given stack version the stack-manager will create a unique set of roles in the format
```
jbw976: > stack given stack version

just "given stack version" here?

negz:
> Aggregate-to labels used by ClusterRole aggregation built in of format: rbac.crossplane.io/aggregate-to-{scope}-{role}: "true" - If scope=namespace, a namespace label will be included.

Probably bike-shedding this a little, but would `rbac.crossplane.io/aggregate-to-{role}: {scope}` work?

lukeweber:
Then you couldn’t aggregate a role to different scopes, so unfortunately
no. This is a straight copy of kubernetes pattern though.

design/one-pager-stacks-security-isolation.md
displague: https://kubernetes.io/docs/reference/access-authn-authz/rbac/

> A RoleBinding may also reference a ClusterRole to grant the permissions to namespaced resources defined in the ClusterRole within the RoleBinding’s namespace. This allows administrators to define a set of common roles for the entire cluster, then reuse them within multiple namespaces.

In support of the notion that a stack's CRDs could be installed once (upon first StackInstall) and then multiple instances of the "Stack" (or StackInstall) could make that stack's controller available within new namespaces, this ClusterRole could be reusable.

The ClusterRole could be reused when creating the `:system` rolebinding needed for the controller.  The `system` roles could change between Stack versions, so we wouldn't want to reuse clusterroles between versions - this aligns with this design.

When creating RoleBindings for workspace admins, we should still be able to benefit from ClusterRole aggregation, allowing a workspace admin to administer any (and only) the Stacks installed in the namespace where their (namespaced) RoleBinding (of a workspace clusterrole) was bound.

This sounds right to me.

design/one-pager-stacks-security-isolation.md
displague: Luke pointed out in a later discussion and subsequent updates, that versioned stack ClusterRoles  would not be recommended for user rolebinding.  These clusterroles are only used for aggregation and for the the `system` role to provide the functionality needed for the controller.

design/one-pager-stacks-security-isolation.md
```diff
@@ -87,19 +114,24 @@ Any future expansions to the model should be accompanied by a thoughtful design.
...
+* All CRDs related to GCP infrastructure and services will be installed to the cluster, such as `CloudSQLInstance`,
+ `CloudSQLInstanceClass`, `GKECluster`, `GKEClusterClass`, `Network`, `Subnetwork`, etc. All CRDs will be labeled
+ `crossplane.io/scope: environment` to denote that they are from a cluster scoped install and for the environment.
+* A `ClusterRole` is created for for system that includes the RBAC rules for the [allowed core Kubernetes kinds](#allowed-resource-access), the stack's owned CRDs, and the CRDs that it depends on.
+* `ClusterRoles` for admin, edit, and view will be created in the cluster scope for the stacks owned CRDs.
 * A `ServiceAccount` is created in the namespace that the stack's controller will be running.  This will serve as the runtime identity of the controller.
-* A `ClusterRoleBinding` is created at the cluster level to bind the `ClusterRole` and the namespaced `ServiceAccount`
-* A `Deployment` is created in the same namespace as the stack and it is expected to launch and manage the pod that hosts the GCP controllers.
+* A `ClusterRoleBinding` is created at the cluster level to bind the system `ClusterRole` and the namespaced `ServiceAccount`
```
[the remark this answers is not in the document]
lukeweber: Explained above.

design/one-pager-stacks-security-isolation.md
lukeweber: updated

design/one-pager-stacks-security-isolation.md
lukeweber: done.

lukeweber:
I've updated images and think I've addressed all feedback at this point. 

Thanks @displague @jbw976 @negz @prasek for all the comments and work on this.

Let me know if there's anything else pending on this one.

design/one-pager-stacks-security-isolation.md
displague: I was also curious about this and found the following helpful:
https://kubernetes.io/docs/reference/access-authn-authz/authentication/#user-impersonation

> Impersonation requests first authenticate as the requesting user, then switch to the impersonated user info.
...
> API server ensures the authenticated users have impersonation privileges.
...
> To impersonate a user, group, or set extra fields, the impersonating user must have the ability to perform the “impersonate” verb on the kind of attribute being impersonated (“user”, “group”, etc.). For clusters that enable the RBAC authorization plugin, the following ClusterRole encompasses the rules needed to set user and group impersonation headers:

PR #1020, opened 2019-11-04
Helm chart: fix crd issues when stacks enabled

### LINKED ISSUE #1026: Installation of stacks fails for helm upgrade
### What happened?
(taken from #1020)

Currently helm chart installation fails when stacks deployment enabled and installed with `helm upgrade --install`

`helm install --name crossplane crossplane-alpha/crossplane --namespace crossplane-system  --version 0.4.0 --set 'clusterStacks.gcp.deploy=true'` ==> Works

`helm upgrade --install crossplane crossplane-alpha/crossplane --namespace crossplane-system  --version 0.4.0 --set 'clusterStacks.gcp.deploy=true'` ==> Fails

The root cause is `post-install` and `post-update` hooks do not work reliably (i.e. work differently for helm install vs helm upgrade —install even it is first time install both). 

Crossplane version:  `v0.4.0`

turkenh:
Fixes #1026 

Currently helm chart installation fails when stacks deployment enabled and installed with `helm upgrade --install`

`helm install --name crossplane crossplane-alpha/crossplane --namespace crossplane-system  --version 0.4.0 --set 'clusterStacks.gcp.deploy=true'` ==> Works

`helm upgrade --install crossplane crossplane-alpha/crossplane --namespace crossplane-system  --version 0.4.0 --set 'clusterStacks.gcp.deploy=true'` ==> Fails

The root cause is `post-install` and `post-update` hooks do not work reliably (i.e. work differently for helm install vs helm upgrade —install even it is first time install both). 

This PR removes `post-install` and `post-update` hooks from `ClusterStackInstall` resources and adds `crd-install` hooks to CRD.

Our CRDs are auto generated by kubebuilder tools, thus this PR also updates makefile so that generated CRDs contain the annotations. Since kubebuilder does not provide a mechanism to add annotations to CRDs, I used `kustomize` in Makefile as suggested [in this comment.](https://github.com/kubernetes-sigs/kubebuilder/issues/387#issuecomment-428380728)

### Checklist

I have:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,6 +93,11 @@ go.test.unit: $(KUBEBUILDER)
 manifests: vendor kubebuilder.manifests
 	@$(INFO) Generating CRD manifests
 	$(CONTROLLERGEN) crd:maxDescLen=0,trivialVersions=true paths=./apis/stacks/... output:dir=$(CRD_DIR)
+# Add "helm.sh/hook: crd-install" and "helm.sh/hook-delete-policy: before-hook-creation" annotations for clusterstackinstalls and stackinstalls CRDs
+	$(eval TMPDIR := $(shell mktemp -d))
+	kustomize build cluster/charts -o $(TMPDIR)
+	mv $(TMPDIR)/apiextensions.k8s.io_v1beta1_customresourcedefinition_clusterstackinstalls.stacks.crossplane.io.yaml $(CRD_DIR)/stacks.crossplane.io_clusterstackinstalls.yaml
+	mv $(TMPDIR)/apiextensions.k8s.io_v1beta1_customresourcedefinition_stackinstalls.stacks.crossplane.io.yaml $(CRD_DIR)/stacks.crossplane.io_stackinstalls.yaml
 	@$(OK) Generating CRD manifests
 
 # Generate a coverage report for cobertura applying exclusions on
--- a/Makefile
+++ b/Makefile
@@ -90,12 +90,20 @@ go.test.unit: $(KUBEBUILDER)
 
 # Generate manifests e.g. CRD, RBAC etc. locally for Stacks API types
 # as it needs the custom "maxDescLen=0" option
-manifests: vendor kubebuilder.manifests
+manifests: vendor kubebuilder.manifests $(KUSTOMIZE)
 	@$(INFO) Generating CRD manifests
 	$(CONTROLLERGEN) crd:maxDescLen=0,trivialVersions=true paths=./apis/stacks/... output:dir=$(CRD_DIR)
-# Add "helm.sh/hook: crd-install" and "helm.sh/hook-delete-policy: before-hook-creation" annotations for clusterstackinstalls and stackinstalls CRDs
+# Add "helm.sh/hook: crd-install" and "helm.sh/hook-delete-policy: before-hook-creation" annotations for
+# clusterstackinstalls and stackinstalls CRDs.
+# Since Crossplane helm chart contains both CRD and ClusterStackInstall CRs, helm fails to install both together.
+# One option was to use `post-install,post-update` hooks in CR to deploy it after CRDs are installed, but this didn't
+# work reliably with "helm upgrade --install" command. Using "crd-install" hook is already suggested in helm best
+# practices doc: https://helm.sh/docs/chart_best_practices/#method-2-crd-install-hooks and we verified that it works
+# reliably for all use cases. The other hook for deletion policy is necessary to be able to redeploy helm chart after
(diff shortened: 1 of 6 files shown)


cluster/charts/crossplane/values.yaml.tmpl
```diff
@@ -15,14 +15,14 @@ imagePullSecrets:
 
 clusterStacks:
   aws:
-    version: v0.1.0
+    version: v0.2.0
```
hasheddan: Do you think it would be a good idea to have these pointed at `master` versions on `master`, then point to latest on `release-0.4`?

turkenh: For master:
Makes sense to me but this was actually suggested by @prasek https://github.com/crossplaneio/crossplane/pull/932#pullrequestreview-301925590

For `release-0.4`
I didn't know `latest`, are they maintained by our release process and point to the versions that I updated here? If so, this would definitely be better.

hasheddan: It looks like @prasek just meant that Crossplane releases should be pinned to stable stacks, which I definitely agree with! I think any time we cut a release we should pin to most recent stable version of each stack, but in master, we should point to stack master versions (the `master` tag is equivalent to `latest` for us). Therefore, in this PR I think it would be useful to pin to `master` versions and then after cherry-picking onto `release-0.4` we should update the versions to stable.

turkenh: Agreed, done!
My only concern is, we should not forget setting to stable versions before making a release. But this is already valid and looks like it is already forgotten on `release-0.4` :)

turkenh: No, I am wrong, `release-0.4` already points to correct thanks to @jbw976

Makefile
```diff
@@ -93,6 +93,11 @@ go.test.unit: $(KUBEBUILDER)
 manifests: vendor kubebuilder.manifests
 	@$(INFO) Generating CRD manifests
 	$(CONTROLLERGEN) crd:maxDescLen=0,trivialVersions=true paths=./apis/stacks/... output:dir=$(CRD_DIR)
+# Add "helm.sh/hook: crd-install" and "helm.sh/hook-delete-policy: before-hook-creation" annotations for clusterstackinstalls and stackinstalls CRDs
+	$(eval TMPDIR := $(shell mktemp -d))
+	kustomize build cluster/charts -o $(TMPDIR)
```
jbw976: all tools in our build flow should be known versions, this appears to use whatever `kustomize` happens to be on the developers machine.  also, it may fail completely if the dev doesn't already have kustomize installed.

For example, you can see how the build submodule installs a specific version of `kubebuilder` into the common tools cache, and will reference that specific version throughout the build, so we are always using a known good version across everyone's machines.  The goal here is consistent and repeatable builds.
https://github.com/upbound/build/blob/master/makelib/kubebuilder.mk#L83

cluster/charts/kustomization.yaml
```diff
@@ -0,0 +1,7 @@
+commonAnnotations:
+  helm.sh/hook: crd-install
+  helm.sh/hook-delete-policy: before-hook-creation
+
+resources:
```
jbw976: this pattern of using `kustomize` as part of the build process may be useful for adding other types of metadata/information to our generated CRDs that the controller-tools and kubebuilder markers don't support.  Also for dynamically filling in version information in a stack's app.yaml, install.yaml, etc. at build time when the version is known.

Makefile
```diff
@@ -93,6 +93,11 @@ go.test.unit: $(KUBEBUILDER)
 manifests: vendor kubebuilder.manifests
 	@$(INFO) Generating CRD manifests
 	$(CONTROLLERGEN) crd:maxDescLen=0,trivialVersions=true paths=./apis/stacks/... output:dir=$(CRD_DIR)
+# Add "helm.sh/hook: crd-install" and "helm.sh/hook-delete-policy: before-hook-creation" annotations for clusterstackinstalls and stackinstalls CRDs
```
jbw976: do you think another quick comment about why `crd-install` is the right hook, and what specifically it does for us (why is it important) would be useful here?

Makefile
turkenh: Yes, this makes sense, thanks for the information.

Here is the PR adding kustomize: https://github.com/upbound/build/pull/76

Makefile
turkenh: Done

turkenh:
Depends on https://github.com/upbound/build/pull/76

turkenh:
@jbw976 done!

PR #1023, opened 2019-11-04
[WIP] Cluster scoped clusterstackinstall

### LINKED ISSUE #981: ClusterPackageInstall should be Cluster Scoped

@lukeweber points out that following https://github.com/crossplaneio/crossplane/pull/959 and the introduction of the Simple Class Selection design / implementation, `ClusterStackInstall` records should not be namespaced.

I explored this a bit here: https://github.com/crossplaneio/crossplane/pull/959#discussion_r338174480 (TODO, bring those thoughts into this issue).

Notably, it is not preferable for a "namespace" admin to remove a namespaced resource with the side effect of removing cluster wide resources (ignoring CRDs).

Luke also points out that a more explicit `// +kubebuilder:resource:scope=Namespaced` would create a better visibility of the CRD scope for Crossplane managed kinds (`StackInstall`).

* Does StackManager need to listen for StackInstall and ClusterStackInstall in different manager contexts to accomplish this?
* Where does StackManager deploy cluster scoped stack controllers?  This is currently inferred from the namespace of the ClusterStackInstall.
  * Do we need a `spec.deploymentNamespace` parameter to compensate? Default to `crossplane-system`?
* Likewise, where are Stack records created for ClusterStackInstalls?
  * Should we investigate ClusterStack resources instead?
  * Do we need a `spec.stackNamespace` argument for this? Default to `crossplane-system`?
  * Does `stackNamespace` solve the problem of where the deployment goes? It follows the namespaced Stack record.
* documentation about installing ClusterStacks will need to be updated
* crossplane-cli `stack install --cluster` will need to be updated

displague:

Following #959, Cluster Stacks now installed at the Cluster scope. 
    
This PR:
* makes ClusterStackInstall objects Cluster scoped
* makes scoping of namespaced CRDs (StackInstall, Stack) explicit
* adds a "controllerNamespace" spec property to ClusterStackInstall
* for ClusterStackInstalls:
  * creates install Jobs in the controllerNamespace
  * creates Stack records, Deployments, and SA in controllerNamespace
    
Closes #981

~_Update: This PR is going to be rebased on #1059.  Making ClusterStackInstalls cluster scoped will require changes on top of #1059.  Additionally, cross-permission-scope ownership triggers garbage collection, so this PR must pick up and extend the label based ownership enhancements brought in #1059._~ _done_

### Checklist

I have:


apis/storage/v1alpha1/zz_generated.claim.go
```diff
@@ -60,27 +49,22 @@ func (cm *Bucket) SetBindingPhase(p runtimev1alpha1.BindingPhase) {
...
+
 // SetResourceReference of this Bucket.
 func (cm *Bucket) SetResourceReference(r *corev1.ObjectReference) {
 	cm.Spec.ResourceReference = r
 }
 
 // SetWriteConnectionSecretToReference of this Bucket.
-func (cm *Bucket) SetWriteConnectionSecretToReference(r *runtimev1alpha1.LocalSecretReference) {
+func (cm *Bucket) SetWriteConnectionSecretToReference(r corev1.LocalObjectReference) {
 	cm.Spec.WriteConnectionSecretToReference = r
```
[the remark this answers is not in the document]
displague: @negz I only added this diff because `make reviewable` pulled my arm.  Can #962 include an update to crossplane-runtime v0.2.0 and include an updated `make manifests` (I also found a `go mod tidy` was needed)?

apis/stacks/v1alpha1/types.go
```diff
@@ -109,6 +119,21 @@ func (si *StackInstall) PermissionScope() string { return string(apiextensions.N
...
+// Controller, ServiceAccount, and namespaced RBAC roles will be created.
+func (si *StackInstall) ControllerNamespace() string { return si.Namespace }
+
+// ControllerNamespace is the namespace where a Stack's ClusterStackInstall Job,
+// Controller, and ServiceAccount will be created.
+func (si *ClusterStackInstall) ControllerNamespace() string {
+	if si.Spec.ControllerNamespace == "" {
+		// TODO(displague) make this a stack-manager CLI and ENV param or return
+		// namespace of the stack-manager (preventing namespace DNE problems)
+		return "crossplane-system"
```
negz: Could we not just make `.spec.controllerNamespace` a required field and avoid setting a default?

displague: I was trying to keep the "ClusterStackInstall" `spec` simple and compatible with existing documentation.

Making the variable explicit now doesn't prevent us from defaulting it later.

Are there other compelling reasons to not use a default, or specifically "crossplane-system"?
Would the implementation of the `TODO` comment help?

negz: I typically prefer to avoid defaulting in favor of convention. This is what we do for Crossplane Services types; I don't believe `crossplane-system` is hardcoded anywhere in Crossplane's codebase but we use it by convention in all examples and documentation for any type that can't function without a namespace specified (`writeConnectionSecretToRef` etc).

The implementation of the TODO would be better than hard coding this default value, but I feel it would still be a worse user experience than making the field mandatory. When the field is mandatory the mental model for setting and getting the value is always the same: read or write the `controllerNamespace` of your manifest. When it's optional with a default (set somewhere else within Crossplane) the mental model becomes more complicated. It's what you set in your YAML, unless it's not, in which case you (as a user who may not be able to see the stack manager pod) need to go somehow figure out what env var or flag Crossplane is running with. Or worse, just submit your YAML and see what happens.

displague: What happens to existing ClusterStackInstalls upon updating Crossplane to a version that now requires this field?  Will users even be permitted to update the CRD in this case if existing CRs are not considered valid against it?

Is there an opportunity for the StackManager to 'heal' existing ClusterStackInstalls by setting an empty controllerNamespace using the existing controller reference?

Are these non-concerns for our current stability level?

negz:
@displague I know there's a lot in flight regarding the Stack API/resource shapes  at the moment. Do you think this PR will still be relevant when the dust settles, or is it likely to be defunct?

displague:
@negz the implementation here does what it says on the tin, but I agree that with the current flux around Stack vs TemplateStack vs ClusterStack, their install types, their dependency and versioning, and possible merging of types, that this work can be set aside and easily resurrected following some  emergence of need and refined design work.

PR #1032, opened 2019-11-06
Integration testing framework design doc

hasheddan:

Adds a design doc for establishing a framework that allows for integration tests to be implemented across projects in the Crossplane ecosystem.

### Checklist

I have:

### DIFF
--- a/design/design-doc-integration-tests.md
+++ b/design/design-doc-integration-tests.md
@@ -110,24 +110,24 @@ following broadly defined steps:
 The Kubernetes [controller-runtime] project provides an [envtest] package for
 creating local control planes for testing. 
 
-2. Registering CRD's
+2. Registering CRDs
 
-For any controller's that watch CRD's to be successfully started, the CRD's must
+For any controller's that watch CRDs to be successfully started, the CRDs must
 have been successfully created in the cluster. If they are unable to be created,
 the controller's will not be started and the `Job` will immediately return an
-error. This serves as a validation test for CRD's.
+error. This serves as a validation test for CRDs.
 
 3. Registering Controllers
 
-Controller's can be registered directly with the manager by calling their
+Controllers can be registered directly with the manager by calling their
 `SetUpWithManager` functions directly. Because all claim controllers
 (`scheduling`, `defaulting`, `claim`) are now part of the individual stacks, it
 is not necessary for the Crossplane controller to be running to test the stacks.
-However, it is necessary for the Crossplane CRD's to be present.
+However, it is necessary for the Crossplane CRDs to be present.
 
 4. Starting the Manager
 
-After all controllers and CRD's are registered, the manager can be started. 
+After all controllers and CRDs are registered, the manager can be started. 
 
 5. Executing Tests
 
@@ -148,7 +148,7 @@ type jobCleanFunc func(client.Client) error
 `Clean` is a special function that must be present in all `Jobs` and always runs
 before a job exits. It specifies how to clean up any resources that are
 left-over from tests executed as part of the `Job`. While it is sufficient to
-simply destroy the control plane to delete Kubernetes built-in API types, CRD's
+simply destroy the control plane to delete Kubernetes built-in API types, CRDs
 that create external resources must be cleaned up directly to ensure the
 deletion of their external infrastructure.
 
--- a/design/design-doc-integration-tests.md
+++ b/design/design-doc-integration-tests.md
@@ -59,7 +59,7 @@ testable.
 ## Proposal
 
 The following sections describe a testing framework that utilizes the Go
-[testing] package to execute tests against a local Kubernetes control plane. The
+[testing] package to execute tests against any Kubernetes control plane. The
 framework is designed such that it can be minimally implemented in the
 short-term, deferring most of the execution and configuration to the tests that
 are written for each implementation. It is intended to evolve over time to
@@ -70,10 +70,9 @@ allowing flexibility to test any scenario.
 
 The framework should be made up of two core components:
 
-- `Job`: a testing environment. Every job runs the full setup and tear-down of a
-  local Kubernetes control plane. A job is made up of configuration, which
-  informs the framework of how to setup the control plane, and `Tests` which
-  test actual execution in the control plane.
+- `Job`: a testing environment. A job is made up of configuration, which informs
+  the framework of how to setup the control plane, and `Tests` which test actual
+  execution against the control plane.
 - `Test`: a test is an action or set of actions to execute against the
   Kubernetes control plane. It could involve one or more steps, each of which
   may cause the test to fail. Each test should be a logical operation, meaning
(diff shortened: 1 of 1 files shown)


design/design-doc-integration-tests.md
```diff
@@ -0,0 +1,333 @@
...
+
+#### Jobs
+
+A `Job` contains the following components. `*Config` will likely evolve over
+time, but will start with just defining the `SyncPeriod` for the controller
+`manager`.
+
+```go
+// A Job is a set of tests that run in a custom configured local Kubernetes environment.
+type Job struct {
```
hasheddan: We will need to pass in a CRD directory path here as well in order to be able to actually create them in the local environment.

design/design-doc-integration-tests.md
```diff
@@ -0,0 +1,333 @@
...
+[bash scripts][current-integration-tests] that are triggered in a Jenkins
+pipeline. While this is effective for determining if the stack can successfully
+be installed into a Crossplane Kubernetes cluster, it does not go so far as to
+test any of the controllers beyond that they start.
+
+All other testing is performed in a manual ad-hoc manner for individual PR's or
+leading up to a release. This leads to uncertainty around how last-minute
+changes landing at the same time will affect the functionality of stacks.
+
+## Goals
```
negz: Are you thinking that these integration tests would hit real cloud provider APIs, or use fake ones mocked out with something like https://golang.org/pkg/net/http/httptest/ or https://github.com/dnaeon/go-vcr? My suspicion is there would be value in supporting both.

I think we'd get value on top of our existing unit tests by testing a controller as a whole even if we mocked out the cloud provider APIs, because we'd be exercising how the controller as a whole functions against a real API server when you manipulate resources. You could write, for example, a test that submits a `CloudSQLInstanceClass`, submits a `MySQLInstance`, and ensures everything works as expected with a fake CloudSQL API.

Using the _real_ CloudSQL API would further increase confidence, but costs more money, is more fragile, and would need us to support injecting cloud provider credentials in a secure fashion.

hasheddan: Hmm yes it was definitely my intention to run against the real provider APIs. I agree that the mocks would be a slight improvement over the unit tests, but anecdotally it wouldn't really increase my confidence of the controllers.

> Using the real CloudSQL API would further increase confidence, but costs more money, is more fragile, and would need us to support injecting cloud provider credentials in a secure fashion.

* "costs more money": this is definitely true when compared to testing against mock APIs. However, I would argue that if we are running a bunch of manual tests on a regular basis, automating those and running them periodically (maybe a few times per day), could actually be more cost-efficient or at least only a marginal increase.

* "is more fragile": Crossplane is run against the real APIs when it is being used. I think running these tests in a controlled environment is not only beneficial, but necessary.

* "injecting cloud provider credentials": because cloud provider credentials are injected into Crossplane in a consistent manner (i.e. as a k8s secret), I believe this will be easier than anticipated. A hook to do this is definitely a common area that could be moved into the framework itself (maybe there is a `Setup` component in addition to the `Clean`).

negz: My suspicion is that even using mocks would be more than a slight improvement on the cloud provider APIs, especially for integration testing the full dynamic and static provisioning scenarios rather than just managed resource controllers.

My preference would be to do both. I’d like to be able to gate every PR on these tests, and that’s not realistic if we’re using real cloud provider APIs. I’m okay starting with just one (real APIs) or the other (mocks) to begin with, though.

design/design-doc-integration-tests.md
```diff
@@ -0,0 +1,333 @@
...
+    tests            []Test
+}
+```
+
+A job's primary method is `func (j *Job) Run() error` which executes the
+following broadly defined steps:
+
+1. Starting the API Server
+
+The Kubernetes [controller-runtime] project provides an [envtest] package for
```
hasheddan: @negz the more I work with this, the more I don't really see a lot of value in wrapping the `envtest` package. I think we will hardly ever be loading the binaries from env variables directly and will likely be using a remote cluster or KIND cluster. I think removing this dependency and just injecting a `client.Config` into the framework at runtime will be better option, and we will have more control over what the API looks like as this framework evolves. Does that sound like the right direction to you?

hasheddan: One area I already have been able to improve since removing `envtest` is the support for `v1` CRDs. We are still currently using only `v1beta1` CRDs across the Crossplane ecosystem, but the ability to enhance the framework at our own desired paces is a big win IMO.

hasheddan: Circling back on this, I have somewhat changed my tune in that all of the controllers are using `controller-runtime` so it makes sense to use it here as well. I have made significant updates to the API though and am feeling better about the architecture: https://github.com/hasheddan/athodyd

design/design-doc-integration-tests.md
```diff
@@ -84,21 +83,43 @@ The framework should be made up of two core components:
...
+}
+```
+
+An example of a `JobOption` function:
+
+```go
+// WithCRDDirectoryPaths sets custom CRD locations for a Job
+func WithCRDDirectoryPaths(crds []string) JobOption {
+    return func(j *Job) {
+        j.cfg.CRDDirectoryPaths = crds
```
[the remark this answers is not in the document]
hasheddan: I think that would make sense here. I chose to operate on `*Job` because it is more flexible for future implementation. However, if we did go with operating on `*JobConfig` for now any future change to operating on `*Job` would not actually require code changes from an end-user perspective, so I would be inclined to go with your recommendation here. Thanks!

negz:
> I think we could include the SetupWithManager and AddToScheme calls to the athodyd.New() using With... functions, or do you think it is important to not obscure that?

My thinking was more that passing them as options couples us to their function signatures without seeming to give us much benefit in return. We just exchange two fairly simple function calls with two options to `New`.

> For example, let's say you want to create a Provider, then create a CloudSQLInstanceClass, then create a MySQLInstance claim and wait for its managed resource to be provisioned, then delete the MySQLInstance claim and wait for the CloudSQLInstance to be deleted. This could be done all in a single test, but that would get rather long.

This _does_ sound like a single test to me. Even if you break it up into multiple subtests (`athodyn.Test`) within in the same test (`athodyn.Job`) it's still tightly coupled because the tests share and depend on each other's state in the underlying API server. Assuming you create the Provider in Test A, create the MySQLInstance in B, and delete it in C then test C is dependent on test B, which is dependent on test A.

I think part of what I'm struggling with is that this is a novel use of the table driven test pattern. Typically in Go a table driven unit test executes the same logic with different inputs for each table entry, whereas with `athodyn` each table entry is an arbitrary set of client calls against an API server, and the success of those client calls could depend on what was done in previous table entries. I think the main advantage to this is that it allows you to avoid repeating yourself; if you want to test three different operations that all need a `Provider` to exist you can setup the `Provider` once, and test operations B and C even if operation A fails.

> "this is where your clean up for your action goes"

Part of using `defer` in my example was because I feel that `defer` is an existing and fairly well understood Go paradigm for doing this kind of thing, and that I'm wary of reinventing that wheel.

> I would defer to your experience here and will move forward as you suggest if you feel strongly.

I'm not sure I'd say I feel strongly about this, but I _am_ a little concerned that we're trying to build a framework for writing integration tests in Go without first writing some integration tests in Go. It's often better to solve the problem without abstraction first, then come back and build the abstraction once you're more familiar with the problem.

I'm happy for you to decide which direction you would prefer to go. If you do stick with the current scaffold based design I suggest that perhaps something like `Test` and `Step` might be a better representation of what is happening than `Job` and `Test`.

design/design-doc-integration-tests.md
```diff
@@ -0,0 +1,499 @@
...
+A `Job` contains the following components:
+
+```go
+// A Job is a set of tests that are executed sequentially in the same cluster environment
+type Job struct {
+    Name        string
+    Description string
+    Tests       []Test
+    cfg         *JobConfig
+    runner      *testing.T
```
negz: Could this be an argument of `Run` rather than a field of `Job`? I believe it's only used in `Run`.

hasheddan:
> I'm not sure I'd say I feel strongly about this, but I am a little concerned that we're trying to build a framework for writing integration tests in Go without first writing some integration tests in Go.

@negz I think this is the key point here :) Let's go with your proposal and and build on it later if we need to. I should have this doc updated by EOD. Thanks for your insight and feedback!

design/design-doc-integration-tests.md
```diff
@@ -0,0 +1,298 @@
...
+    }
+}()
+```
+
+To override the default `Cleaner` function, pass in your own to `athodyd.New()`:
+
+```go
+a, err := athodyd.New(cfg,
+    athodyd.WithCRDDirectoryPaths([]string{"../crds"}),
+    athodyd.WithCleaner(func(*envtest.Environment, client.Client){ return nil }))
```
negz: What do you think about:

1. Either making this `WithCleaners(cleaner ...OperationFn)` or introducing an `OperationChain` type that chains operations.
1.  Renaming [`athodyd.NewCleaner`](https://github.com/hasheddan/athodyd/blob/5821b7d750d33c9cdd0a38581f97ab2adbba5ca4/athodyd.go#L58) to `athodyd.NewCRDCleaner`

I think most folks are going to want to clean their CRDs after running a test, so it would be nice to be able to write something like `athodyd.New(cfg, athodyd.Cleaners(myCustomCleaner, athodyd.NewCRDCleaner()))`.

negz:
>  I'm excited to see how our first integration tests shake out.

@hasheddan Speaking of which, I suggest we use the [`integration` build tag](https://filipnikolovski.com/separating-tests-in-go/) to gate whether integration tests are run alongside our unit tests.

PR #1041, opened 2019-11-10
Refresh README.md

negz:

This update attempts to make `README.md` as succinct, targeted, and useful to potential Crossplane community members as possible. My goals were to:

* Explain what Crossplane does - what problems it solves.
* Touch on Crossplane's architecture to the extent that it could help potential
  users explain how and why they might use Crossplane.
* Funnel readers into getting started and getting involved.
* Remove anything that doesn't _really_ need to be on the landing page.

I suggest reviewing the proposed `README.md` as a whole rather than a diff on the existing document.

### Checklist

I have:

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,117 +1,84 @@
-<p align="center"><img src="docs/media/banner.png" alt="Crossplane"></p>
+[![Build Status](https://jenkinsci.upbound.io/buildStatus/icon?job=crossplaneio/crossplane/build/master)](https://jenkinsci.upbound.io/blue/organizations/jenkins/crossplaneio%2Fcrossplane%2Fbuild/activity) [![GitHub release](https://img.shields.io/github/release/crossplaneio/crossplane/all.svg?style=flat-square)](https://github.com/crossplaneio/crossplane/releases) [![Docker Pulls](https://img.shields.io/docker/pulls/crossplane/crossplane.svg)](https://img.shields.io/docker/pulls/crossplane/crossplane.svg) [![Go Report Card](https://goreportcard.com/badge/github.com/crossplaneio/crossplane)](https://goreportcard.com/report/github.com/crossplaneio/crossplane) [![Slack](https://slack.crossplane.io/badge.svg)](https://slack.crossplane.io) [![Twitter Follow](https://img.shields.io/twitter/follow/crossplane_io.svg?style=social&label=Follow)](https://twitter.com/intent/follow?screen_name=crossplane_io&user_id=788180534543339520)
 
-[![Build Status](https://jenkinsci.upbound.io/buildStatus/icon?job=crossplane/build/master)](https://jenkinsci.upbound.io/blue/organizations/jenkins/crossplane%2Fbuild/activity)
-[![GitHub release](https://img.shields.io/github/release/crossplaneio/crossplane/all.svg?style=flat-square)](https://github.com/crossplaneio/crossplane/releases)
-[![Docker Pulls](https://img.shields.io/docker/pulls/crossplane/crossplane.svg)](https://img.shields.io/docker/pulls/crossplane/crossplane.svg)
-[![Go Report Card](https://goreportcard.com/badge/github.com/crossplaneio/crossplane)](https://goreportcard.com/report/github.com/crossplaneio/crossplane)
-[![FOSSA Status](https://app.fossa.io/api/projects/git%2Bgithub.com%2Fcrossplaneio%2Fcrossplane.svg?type=shield)](https://app.fossa.io/projects/git%2Bgithub.com%2Fcrossplaneio%2Fcrossplane?ref=badge_shield)
-[![Slack](https://slack.crossplane.io/badge.svg)](https://slack.crossplane.io)
-[![Twitter Follow](https://img.shields.io/twitter/follow/crossplane_io.svg?style=social&label=Follow)](https://twitter.com/intent/follow?screen_name=crossplane_io&user_id=788180534543339520)
+![Crossplane](docs/media/banner.png)
 
-[![Quality Gate Status](https://sonarcloud.io/api/project_badges/measure?project=crossplaneio_crossplane&metric=alert_status)](https://sonarcloud.io/dashboard?id=crossplaneio_crossplane)
-[![Coverage](https://sonarcloud.io/api/project_badges/measure?project=crossplaneio_crossplane&metric=coverage)](https://sonarcloud.io/dashboard?id=crossplaneio_crossplane)
-[![Maintainability Rating](https://sonarcloud.io/api/project_badges/measure?project=crossplaneio_crossplane&metric=sqale_rating)](https://sonarcloud.io/dashboard?id=crossplaneio_crossplane)
-[![Reliability Rating](https://sonarcloud.io/api/project_badges/measure?project=crossplaneio_crossplane&metric=reliability_rating)](https://sonarcloud.io/dashboard?id=crossplaneio_crossplane)
-[![Security Rating](https://sonarcloud.io/api/project_badges/measure?project=crossplaneio_crossplane&metric=security_rating)](https://sonarcloud.io/dashboard?id=crossplaneio_crossplane)
+[Crossplane] is the open source multicloud control plane. With Crossplane you
+can manage your applications and infrastructure across clouds, regions, and
+clusters. Crossplane is composable and Kubernetes native. Deploy it to an
+existing Kubernetes cluster to manage cloud infrastructure like databases,
+buckets, and caches using `kubectl`. Deploy it standalone to manage Kubernetes
+clusters, the workloads running on them, and the cloud infrastructure those
+workloads depend upon.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,117 +1,84 @@
...
+   databases, caches, and Kubernetes clusters.
+1. Define **portable application workloads**, including their cloud
+   infrastructure dependencies, and schedule them to the most appropriate
+   infrastructure across any cloud.
+1. Package their workloads as an **application stack** that others may easily
+   deploy.
 
-<h4 align="center"><img src="docs/media/crossplane-overview.png" alt="Crossplane"></h4>
+The full vision and architecture of the Crossplane project is described in our
+[architecture document].
```
hasheddan: Took a read through this again, and while it has some valuable information about the motivation for Crossplane, it contains a lot of outdated information about the mechanics of the system. How would you feel about omitting this until we have that document updated? Or, alternatively, we could update that document before merging this (happy to help out with that if needed!).

README.md
```diff
@@ -1,117 +1,84 @@
...
-1. [Crossplane Workloads](docs/README.md#crossplane-workloads) - define complete applications and schedule across
-   clusters, regions, and clouds.
-1. [Crossplane Clusters](docs/README.md#crossplane-clusters) - manage multiple Kubernetes clusters from a single
-   control plane.
+## Get Started
 
-## Architecture and Vision
+Take a look at the [getting started] guide to learn how to install Crossplane,
+install a stack, and provision cloud infrastructure using `kubectl`. Refer to
+the Crossplane [documentation] for comprehensive guides to using Crossplane,
```
hasheddan:
```suggestion
the Crossplane [documentation] for comprehensive guides for using Crossplane,
```

README.md
```diff
@@ -1,117 +1,84 @@
...
+including deploying an application packaged as a stack.
 
-The full architecture and vision of the Crossplane project is described in depth in the [architecture document](https://docs.google.com/document/d/1whncqdUeU2cATGEJhHvzXWC9xdK29Er45NJeoemxebo/edit?usp=sharing). It is the best place to learn more about how Crossplane fits into the Kubernetes ecosystem, the intended use cases, and comparisons to existing projects.
+## Get Involved
 
-## Getting Started and Documentation
+* Discuss Crossplane on [Slack] or our [developer mailing list].
```
hasheddan: I think we would still be responsive if someone were to utilize our mailing list, but it has not had a post since 2018. Do we still want to direct people there? If so, it might be worth cross-posting announcements and releases on the mailing list as well. I know there is still a large group of people that prefer mailing lists to github / slack communication.

README.md
negz: I think there may be value in adding a note to the architecture document reflecting that there have been some changes since we first released Crossplane and directing folks to our design documents. I would prefer not to block this PR on doing so, given `README.md` already links to the architecture document.

README.md
negz: I believe "guides to using" is more correct in this context.

README.md
negz: I'm happy either way.

README.md
hasheddan: Fair enough! :)

README.md
hasheddan: Agreed, would not block on this!

README.md
```diff
@@ -1,117 +1,84 @@
...
+The Crossplane community meeting takes place every other [Tuesday at 9:00am
+Pacific Time]. Anyone who wants to discuss the direction of the project, design
+and implementation reviews, or raise general questions with the broader
+community is encouraged to join.
 
 * Meeting link: https://zoom.us/j/425148449
-* [Current agenda and past meeting notes](https://docs.google.com/document/d/1q_sp2jLQsDEOX7Yug6TPOv7Fwrys6EwcF5Itxjkno7Y/edit?usp=sharing)
-* [Past meeting recordings](https://www.youtube.com/playlist?list=PL510POnNVaaYYYDSICFSNWFqNbx1EMr-M)
-
-## Project Status
```
jbw976: there is no mention of project status anymore that I see, do you think we should mention at least some summary of the status? and/or link to the API status table?

README.md
```diff
@@ -1,117 +1,84 @@
...
+extend Crossplane with new functionality. A stack can extend Crossplane with the
+ability to manage the infrastructure of a new cloud provider, or to deploy and
+manage a new cloud-native application. This enables Crossplane users to:
 
-Crossplane enables the community to build and publish Stacks to add more clouds
-and cloud services to Crossplane with support for out-of-tree extensibility and
-independent release schedules. Crossplane includes Stacks for [GCP][stack-gcp], 
-[AWS][stack-aws], and [Azure][stack-azure] today.
+1. Install the **infrastructure stacks** for their desired clouds to enable
+   on-demand, **cloud-portable provisioning of infrastructure** services like
```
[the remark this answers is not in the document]
negz: Could you expand on what we'd be trying to communicate by doing so? Keep in mind we have not touched on anything to do with separation of concerns at this point in the README. I'm wondering what "self-service" would mean to a first time user unfamiliar with Crossplane but familiar with Kubernetes? Isn't everything in Kubernetes "self service"?

To be fair this same argument could be made for "on-demand", so I'm not opposed to including it if you feel it adds value/context.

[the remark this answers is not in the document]
jbw976: i don't feel strongly about it, you can keep it as is

README.md
negz: I don't, to be honest.

My thinking is:

* This was useful to call out when we were a super early tech demo, but now that we're a little more mature (but not yet stable) I'm not sure it adds much value. I'd say we're approaching a similar level of maturity to most other projects in our space.
* The status of Crossplane as a project is increasingly split across repos; i.e. it's probably better to reflect the maturity of stack-aws over at stack-aws.
* All of our releases are 0.x.x and tagged as "pre-release" in Github, which I feel represents the status of Crossplane itself reasonably well.
* Anyone who decides to try out Crossplane will see that it's alpha when they get to the install steps.
* I surveyed a bunch of similar OSS project READMEs to inform this update, and found that it was rare for such details to be so front and center.
* The detailed status of each API group is still there in the docs

README.md
```diff
@@ -1,117 +1,84 @@
...
-[![Coverage](https://sonarcloud.io/api/project_badges/measure?project=crossplaneio_crossplane&metric=coverage)](https://sonarcloud.io/dashboard?id=crossplaneio_crossplane)
-[![Maintainability Rating](https://sonarcloud.io/api/project_badges/measure?project=crossplaneio_crossplane&metric=sqale_rating)](https://sonarcloud.io/dashboard?id=crossplaneio_crossplane)
-[![Reliability Rating](https://sonarcloud.io/api/project_badges/measure?project=crossplaneio_crossplane&metric=reliability_rating)](https://sonarcloud.io/dashboard?id=crossplaneio_crossplane)
-[![Security Rating](https://sonarcloud.io/api/project_badges/measure?project=crossplaneio_crossplane&metric=security_rating)](https://sonarcloud.io/dashboard?id=crossplaneio_crossplane)
+[Crossplane] is the open source multicloud control plane. With Crossplane you
+can manage your applications and infrastructure across clouds, regions, and
+clusters. Crossplane is composable and Kubernetes native. Deploy it to an
+existing Kubernetes cluster to manage cloud infrastructure like databases,
+buckets, and caches using `kubectl`. Deploy it standalone to manage Kubernetes
```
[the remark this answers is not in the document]
negz: > it sounds like it's making the point that the only way to manage k8s clusters and workloads is if you deploy it standalone.

Hrm, my intent here is to highlight the two most common ways you might deploy and use Crossplane. I'm personally having trouble reading it back and taking away that it's saying you can _only_ manage Kubernetes clusters and workloads if you deploy Crossplane standalone.

> You could just as easily deploy it to an existing k8s cluster and still manage other k8s clusters and schedule workloads to them, right?

You _could_, but do you think this is a use case anyone has? I already run my own Kubernetes cluster(s), and I'm installing Crossplane there to manage other Kubernetes clusters and the workloads running on them? My feeling is that if this use case exists at all, it's a pretty rare one.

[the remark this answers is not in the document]
negz: My broader thinking here is:

* Crossplane is useful as a tool dropped into an existing Kubernetes cluster. My intuition is that this will be the most common on-ramp into Crossplane; folks who have a cluster and want to manage their infrastructure from inside Kubernetes.
* Crossplane's Kubernetes cluster and workload management functionality doesn't, in my opinion, make much sense in the context of a service installed into an existing Kubernetes cluster.  I think wanting to install a tool into your existing cluster to manage other clusters is a niche use case at best.
* The cluster and workload management functionality makes much more sense if you start to look at Crossplane as your multicloud control plane above the cluster level, but I'm not sure folks will make that "mental leap" without a little prompting. I certainly would not have.

So the goal here is to lead with "Hey, are you using Kubernetes today? If you drop Crossplane into your existing setup you'll get this value" but to also set context as to where Crossplane is going with all this cluster and workload management functionality.

Issue #1054, opened 2019-11-12
ClusterStackInstalls should create stack+version specific clusterroles for personas


displague:
### What problem are you facing?

Part of #1036

The [Stacks Security and Security Isolation One-Pager](https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-stacks-security-isolation.md) was updated and expanded requiring new Stack behaviors.

https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-stacks-security-isolation.md#clusterroles-for-a-given-clusterstack-install:

> For a given stack version the stack-manager will create a unique set of roles in the format `stack:{repo-name}:{stack-name}:{version}:{role=system|admin|edit|view}`. All of these roles are for internal use only. The system role is bound by the controller for the stack, and will include the set of dependent resources for a given stack. The other (admin,edit,view) will be labeled for aggregation to their appropriate namespaced or environment scoped crossplane managed roles.

### How could Crossplane help solve your problem?

ClusterStackInstalls should create stack+version specific clusterroles including CRD types.  These roles should include the ability to work with the including CRD types.

They should be aggregated to the built in environment roles.

Issue #1053, opened 2019-11-12
Crossplane should install base ClusterRoles for cluster and namespace admin, edit, and view personas


displague:
### What problem are you facing?

Part of #1036

The [Stacks Security and Security Isolation One-Pager](https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-stacks-security-isolation.md) was updated and expanded requiring new Stack behaviors.

https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-stacks-security-isolation.md#default-clusterroles-for-aggregation:

> Defaults that stack-manager installs that would aggregate to the crossplane-env-{role} ClusterRole objects, for example, permission on ClusterStackInstall objects.
> 
> * `crossplane:stack-manager:env:default:admin`
> * `crossplane:stack-manager:env:default:edit`
> * `crossplane:stack-manager:env:default:view`
> 

> Defaults that stack-manager installs that would aggregate to the crossplane:ns:{ns-name}:{role} ClusterRole objects, for example, permissions on StackInstall objects.
> 
> `crossplane:stack-manager:ns:default:admin`
> `crossplane:stack-manager:ns:default:edit`
> `crossplane:stack-manager:ns:default:view`

### How could Crossplane help solve your problem?

Crossplane should install with base ClusterRoles for namespace and environment admin, edit, and view roles.  

These roles do not receive aggregated roles from new stacks.  These roles are the base roles that `crossplane:stack-manager:{scope}:{namespace}:{role}` roles will receive, permitting access to the default Crossplane types (Stacks, StackInstalls, ClusterStackInstalls, claims) needed by those environment wide or namespace specific personas.

displague:
In #1053 I am implementing this through an additional Helm chart template.

I don't believe it is necessary to "manage" the clusterroles, including `crossplane-admin`, through the `stack-manager`.  These ClusterRoles do not change.  They inherit rules through clusterrole aggregation.

I think it is sufficient to install these clusterroles with crossplane (or opt to exclude them).  The stack-manager will ensure that stack cluster roles include labels that will aggregate to these predefined persona cluster roles so long as they exist.

displague:
Extending this issue to cover the default namespace clusterroles:

https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-stacks-security-isolation.md#default-clusterroles-for-aggregation

[...]

Issue #1050, opened 2019-11-12
Create a reusable ClusterRole for Stack admins, editors, and viewers


displague:
### What problem are you facing?

Part of #1036

The [Stacks Security and Security Isolation One-Pager](https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-stacks-security-isolation.md) was updated and expanded requiring new Stack behaviors.

https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-stacks-security-isolation.md#installation-flow: 

> ClusterRoles for admin, edit, and view will be created in the cluster scope for the stacks owned CRDs.

https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-stacks-security-isolation.md#default-clusterroles-for-aggregation

> Default ClusterRoles for aggregation
> We've defined the roles that get created that will aggregate the permissions of installed stacks and are meant to be bound to RBAC subjects, like a group or User, but these roles themselves need to be seeded with initial permissions so that a user who has a namespace or environment level role could for example install the first stack.
> 
> Defaults that stack-manager installs that would aggregate to the crossplane-env-{role} ClusterRole objects, for example, permission on ClusterStackInstall objects.
> 
> crossplane:stack-manager:env:default:admin
> crossplane:stack-manager:env:default:edit
> crossplane:stack-manager:env:default:view
> 

### How could Crossplane help solve your problem?

Create a reusable ClusterRole for Stack admins, editors, and viewers.

Issue #1048, opened 2019-11-12
CRDs should be labeled with their 'scope' (environment/workspace)


displague:
### What problem are you facing?

Part of #1036

The [Stacks Security and Security Isolation One-Pager](https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-stacks-security-isolation.md) was updated and expanded requiring new Stack behaviors.

https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-stacks-security-isolation.md#installation-flow: 

> All CRDs defined by the WordPress stack will be installed into the cluster (remember that CRDs themselves are always cluster scoped). In the case of WordPress, a single CRD is installed that represents a WordPress application instance, WordPressInstance. All CRDs will be labeled crossplane.io/scope: namespace and namespace.crossplane.io/{namespace-name}: "true" where namespace-name is the actual namespace the stack was installed into.

### How could Crossplane help solve your problem?

CRDs should be labeled with their 'scope' (environment/workspace).

PR #1059, opened 2019-11-13
Stack ClusterRoles and Parent + RBAC Labelling

### LINKED ISSUE #1049: Create a reusable ClusterRole for the Stack deployments

Part of #1036

https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-stacks-security-isolation.md#cluster-stacks: 

> * A `ClusterRole` for the controller that we will denote as the system `ClusterRole` that includes the RBAC rules for the [allowed core Kubernetes kinds](#allowed-resource-access), the stack's owned CRDs, and the CRDs that it depends  on. This system role is in contrast to the admin, view, and edit roles, which only provide access to the installed CRDs of a stack.
> * A `ServiceAccount` is created in the namespace that the stack's controller will be running.  This will serve as the runtime identity of the controller.	

Create a reusable ClusterRole for the Stack `deployment`.

displague:

This PR implements many, not all, of the changes recommended in @lukeweber's [cluster roles to support self-service](https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-stacks-security-isolation.md) updates, PR #1008, which extends the Stack Security & Isolation design.

The majority of this work can be viewed as managing Stack, Environment (Cluster-wide), and Namespace specific ClusterRoles along with labels to take advantage of AggregationRules.

---

During the implementation of this PR, we became aware of problems settings ownership across permission scopes on Kubernetes resources.

The design for Parent/Child resource labelling, having recently been merged, became the underpinning for our [approach](#1091) to remedy the cross-scope ownership problem on CRD resources.

As this PR introduces several more child resources to Stacks, it became necessary to follow the same labelling practice and fix additional owner references to clusterroles and clusterrolebindings which this PR was touching.  Stacks now  delete their `ClusterRoles` and `ClusterRoleBindings`, detected using parent labels, through a finalizer, since `ownerReferences` (and garbage collection) can not cross permission scopes.

Similarly, `StacksInstalls` are now made to delete their `Stack` resources (through parent label matching) before removing the `CustomResourceDefinitions`.  This is to avoid the condition where the `Stack` controller loses the ability to watch its owned resources, which results in the Stack being unable to clear its finalizer.

---

This PR closes the following issues:

* #1049 clusterrole for namespaced stack deployments
* #1054 ClusterStackInstalls should create stack+version specific clusterroles for personas
* #1099 create stack version specific clusterroles for aggregation
* #1015 CRDs should be labeled with the namespaces of their stacks for namespaced stacks
* #1048 CRDs should be labeled with stack scope and stack names
* #1053 install base ClusterRoles for cluster and namespace admin, edit, and view personas
* #1050 create default clusterroles for environment, ns, and admin
* #1113 create admin/edit/view clusterroles for stack namespaces
* #1111 create a crossplane:master group on install
* #752 - closed by assigning parent labels on remaining stack created resources
* #1121 - closed by not setting ownerref on cluster resources from namespaced resources

Additional details are provided in most commit messages and the issues linked above.

#### Testing

The following commands are useful for seeing this PR in action

##### examining the effect of environment persona clusterroles

```
$ kubectl create clusterrolebinding marques-has-crossplane-env-view --clusterrole crossplane-env-view --user marques@example.com
clusterrolebinding.rbac.authorization.k8s.io/marques-has-crossplane-env-view created
$ kubectl auth can-i --list --as=marques@example.com
```

##### examining the effect of environment persona clusterroles after installing a cluster scoped stack

```
$ kubectl crossplane stack generate-install --cluster crossplane/stack-gcp:v0.3.1 stack-gcp | kubectl -n crossplane-system apply -f - 
$ kubectl auth can-i --list --as=marques@example.com

$ kubectl delete clusterrole marques-has-crossplane-env-view
```

##### examining the effect of environment persona clusterroles after installing a namespaced scoped stack

```
$ kubectl create namespace precipice
$ kubectl crossplane stack generate-install crossplane/stack-wordpress:master stack-wordpress | kubectl -n precipice apply -f - 
$ kubectl auth can-i --list --as=marques@example.com

$ kubectl create rolebinding marques-has-crossplane-ns-precipice-view --clusterrole crossplane:ns:precipice:view --user marques@example.com --namespace precipice
$ kubectl auth can-i --list --as=marques@example.com --namespace default
$ kubectl auth can-i --list --as=marques@example.com --namespace precipice

$ kubectl delete rolebinding marques-has-crossplane-ns-precipice-view 
```

##### examining the effect of binding the crossplane-admin role

```
$ kubectl create clusterrolebinding marques-has-crossplane-admin --clusterrole crossplane-admin  --user marques@example.com
clusterrolebinding.rbac.authorization.k8s.io/marques-has-crossplane-admin created
$ kubectl auth can-i --list --as=marques@example.com 
# notice Crossplane core, Wordpress, GCP, StackInstall, ClusterStackInstall, Stacks, Secrets, and ConfigMaps types
$ kubectl auth can-i --as=marques@example.com create secret
$ kubectl auth can-i --as=marques@example.com delete wordpressinstance --namespace precipice

$ kubectl delete clusterrole marques-has-crossplane-admin
```

##### examining the capabilities of the preinstalled group binding crossplane:masters

```
$ kubectl auth can-i --list --as-group=crossplane:masters --as-group=system:authenticated --as=nobodyspecial
```

##### examining the labels applied for RBAC roles and resource relationship mapping

```
$ kubectl get -A  -lcore.crossplane.io/parent-name=stack-gcp  clusterrole,stack,stackinstall,clusterstackinstall,deployment,clusterrolebinding,rolebinding,role,job,pod
```
##### examining the labels available for assigning personas
```
$ kubectl get clusterrole -l core.crossplane.io/scope=environment
$ kubectl get clusterrole -l core.crossplane.io/scope=namespace
```
##### demonstrating admins granting roles to other admins

```
$ kubectl create rolebinding demo-crossplane-onboarded-ns-admin --clusterrole crossplane:ns:precipice:admin --user new-guy@example.com --namespace precipice --as=precipice-admin@example.com
rolebinding.rbac.authorization.k8s.io/demo-crossplane-onboarded-ns-admin created

$ kubectl create clusterrolebinding demo-crossplane-onboarded-admin --clusterrole crossplane-env-admin --user new-guy@example.com --as=env-admin@example.com
clusterrolebinding.rbac.authorization.k8s.io/demo-crossplane-onboarded-admin created
```

### Checklist

I have:


cluster/charts/crossplane/templates/crossplane-admin-clusterrole.yaml
```diff
@@ -0,0 +1,30 @@
+{{- if .Values.personas.deploy }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ template "name" . }}-admin
+  labels:
+    app: {{ template "name" . }}
```
suskin: Would it make sense to use the `app.kubernetes.io/name` label?

https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

cluster/charts/crossplane/templates/crossplane-admin-clusterrole.yaml
```diff
@@ -0,0 +1,30 @@
...
+    chart: {{ template "chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+    crossplane.io/scope: "environment"
+aggregationRule:
```
suskin: Yess

cluster/charts/crossplane/templates/environment-personas-clusterroles.yaml
```diff
@@ -0,0 +1,156 @@
...
+  - database.crossplane.io
+  - kubernetes.crossplane.io
+  - core.crossplane.io
+  - stacks.crossplane.io
+  - storage.crossplane.io
+  - workload.crossplane.io
+  resources:
+  - "*"
+  verbs:
+  - create
```
suskin: What about verbs like `watch`, or `list`?

cluster/charts/crossplane/templates/environment-personas-clusterroles.yaml
```diff
@@ -0,0 +1,156 @@
...
+  - deletecollection
+  - patch
+  - update
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ template "name" . }}-env-view
+  labels:
+    rbac.crossplane.io/aggregate-to-environment-edit: "true"
```
suskin: > What about verbs like watch, or list?

It looks like this answers my question, but if I'm reading this correctly, there are many levels of aggregation, and we've removed the overlapping parts of the permissions from the higher levels of the hierarchy. Did we already consider duplicating the permissions in the definitions of the higher levels? Or having fewer levels of rollup? It seems like having many levels (as in, more than two) could make it hard for a human to reason about what the resolved set of permissions is for a given role.

pkg/controller/stacks/install/installjob.go
```diff
@@ -302,3 +377,13 @@ func isStackObject(obj *unstructured.Unstructured) bool {
 	return gvk.Group == v1alpha1.Group && gvk.Version == v1alpha1.Version &&
 		strings.EqualFold(gvk.Kind, v1alpha1.StackKind)
 }
+
+func isCRDObject(obj runtime.Object) bool {
```
suskin: Should this be a shared utility function somewhere? Seems like it might be useful in other places

cluster/charts/crossplane/templates/crossplane-admin-clusterrole.yaml
displague: We've already got `-lapp=crossplane` in use in this repo.

After a quick bit of surveying, I'm not sure why `app` is as prevalent as it is.  There are alternatives such as `application` and `k8s-app`.  With such diversity, I'm more inclined to adopt the recommended label.  In my local `kind` `core-dns` deployment, `k8s-app` is used.  Tiller and [this nginx example in the Kubernetes docs](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set) use `app`.

Since that change would have to affect existing docs and tools, I would like to defer on adopting the recommended label.

cluster/charts/crossplane/templates/environment-personas-clusterroles.yaml
displague: This is an interesting question.. 

The `crossplane-env-edit` role would pick up the `get`, `watch`, and `list` verbs because of https://github.com/crossplaneio/crossplane/pull/1059/files/9e5ec6336d2854c06bd962e9ad8918e6edc6ec5a#diff-de6ef6770bec0bdeca8902bda7136f60R107

In the design doc, we state that some of the cluster roles, ``stack:{repo-name}:{stack-name}:{version}:{role=system|admin|edit|view}`, are for internal use only.

If `crossplane:stack-manager:{scope}:default:{role}` cluster roles are for internal use only, then we don't need to worry about the verbs they provide, so long as `crossplane-{scope}-{role}` picks up all the needed verbs through aggregation.  However, if we expect users to consume the `crossplane:stack-manager:{scope}:default:{role}` clusterroles directly, we should include the read-only verbs.

We've been using the Kubernetes model as a guide. 
```
kubectl get -o yaml clusterrole view,system:aggregate-to-view
kubectl get -o yaml clusterrole  edit,system:aggregate-to-edit
kubectl get -o yaml clusterrole  admin,system:aggregate-to-admin
```

cluster/charts/crossplane/templates/environment-personas-clusterroles.yaml
displague: Right, let's continue this here: https://github.com/crossplaneio/crossplane/pull/1059/files/9e5ec6336d2854c06bd962e9ad8918e6edc6ec5a#r354889405

pkg/controller/stacks/install/installjob.go
displague: Perhaps, once we have more uses.  We've been moving away from generic `utils` packages: https://github.com/crossplaneio/crossplane-runtime/issues/1

pkg/controller/stacks/stack/stack.go
```diff
@@ -160,29 +175,157 @@ func (h *stackHandler) update(ctx context.Context) (reconcile.Result, error) {
...
+	} {
+		rules := []rbacv1.PolicyRule{}
+		for _, crd := range h.ext.Spec.CRDs {
+			rules = append(rules, rbacv1.PolicyRule{
+				APIGroups: []string{crd.GroupVersionKind().Group},
+				Resources: []string{crd.Kind},
+				Verbs:     verbs,
+			})
+		}
+		name := fmt.Sprintf(clusterRoleNameFmt, h.ext.GetNamespace(), h.ext.GetName(), h.ext.Spec.Version, persona)
```
lukeweber: Because this string so many arguments, I might just make the templating of clusterRoleNameFmt a helper function to avoid any accidental mixups on ordering of fields.

cluster/charts/crossplane/templates/environment-personas-clusterroles.yaml
```diff
@@ -0,0 +1,156 @@
...
+rules:
+- apiGroups:
+  - cache.crossplane.io
+  - compute.crossplane.io
+  - database.crossplane.io
+  - kubernetes.crossplane.io
+  - core.crossplane.io
+  - stacks.crossplane.io
+  - storage.crossplane.io
+  - workload.crossplane.io
```
lukeweber: I think we need to consider that we need access to secrets + configmaps. Per k8s standards view has no access to secrets, but edit and admin roles would. Configmaps I assume standard rules apply and I would add them to our permissions so they stand on their own as being useful.

pkg/controller/stacks/stack/stack.go
```diff
@@ -160,29 +175,157 @@ func (h *stackHandler) update(ctx context.Context) (reconcile.Result, error) {
...
+
+		namespaceToCrossplaneScopes := map[bool]string{
```
lukeweber: I might just change this to an because I think it's clearer for me

```
if h.isNamespaced() {
    scope = namespaced
else {
    scope = environment
}

File: cluster/charts/crossplane/templates/environment-personas-clusterroles.yaml (hunk 7 above)
Comment (displague) [reply cluster/charts/crossplane/templates/environment-personas-clusterroles.yaml]: The `...default:edit` (namespaced and environment) `ClusterRoles` now get read and write verbs for `secrets` and `configmaps`.

File: cluster/charts/crossplane/templates/namespace-personas-clusterroles.yaml [hunk 9]
```diff
@@ -0,0 +1,139 @@
...
+    rbac.crossplane.io/aggregate-to-namespace-default-view: "true"
+    app: {{ template "name" . }}
+    chart: {{ template "chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+    crossplane.io/scope: "namespace"
+aggregationRule:
+  clusterRoleSelectors:
+  - matchLabels:
+      rbac.crossplane.io/aggregate-to-namespace-view: "true"
```
lukeweber: I think this may cause problems as any namespace specific install will match here. Also if we have match labels we wouldn't define any rules statically.

```suggestion

```

cluster/charts/crossplane/templates/namespace-personas-clusterroles.yaml
```diff
@@ -0,0 +1,139 @@
...
+    rbac.crossplane.io/aggregate-to-namespace-default-admin: "true"
+    app: {{ template "name" . }}
+    chart: {{ template "chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+    crossplane.io/scope: "namespace"
+aggregationRule:
+  clusterRoleSelectors:
+  - matchLabels:
+      rbac.crossplane.io/aggregate-to-namespace-admin: "true"
```
lukeweber: Remove match labels here, these are the defaults and should be the leaf in RBAC model, they should aggregate into every generated role

cluster/charts/crossplane/templates/namespace-personas-clusterroles.yaml
```diff
@@ -0,0 +1,139 @@
+{{- if .Values.personas.deploy }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ template "name" . }}:stack-manager:ns:default:admin
+  labels:
+    rbac.crossplane.io/aggregate-to-crossplane-admin: "true"
+    rbac.crossplane.io/aggregate-to-namespace-default-admin: "true"
```
lukeweber: I don't think this label is needed.
```suggestion
```

cluster/charts/crossplane/templates/namespace-personas-clusterroles.yaml
```diff
@@ -0,0 +1,139 @@
...
+    rbac.crossplane.io/aggregate-to-namespace-admin: "true"
+    rbac.crossplane.io/aggregate-to-namespace-default-edit: "true"
+    app: {{ template "name" . }}
+    chart: {{ template "chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+    crossplane.io/scope: "namespace"
+aggregationRule:
+  clusterRoleSelectors:
+  - matchLabels:
```
lukeweber: remove match labels, these are defaults.

pkg/controller/stacks/stack/stack.go
```diff
@@ -160,29 +188,169 @@ func (h *stackHandler) update(ctx context.Context) (reconcile.Result, error) {
 	return reconcile.Result{}, nil
 }
 
-func (h *stackHandler) createNamespacedRoleBinding(ctx context.Context, owner metav1.OwnerReference) error {
-	cr := &rbac.Role{
+// createPersonaClusterRoles creates admin, edit, and view clusterroles that are
+// namespace+stack+version specific
+func (h *stackHandler) createPersonaClusterRoles(ctx context.Context, labels map[string]string) error {
```
jbw976: this function is a pretty clean approach for creating the personal cluster roles that include all the CRDs for the stack and the correct aggregation labels etc., but it could use a few comments sprinkled through to explain the steps being taken and the context around why they are taken. that will definitely be helpful for future maintainers.

Same with the `generateNamespaceClusterRoles` function too.

pkg/controller/stacks/stack/stack.go
```diff
@@ -160,29 +188,169 @@ func (h *stackHandler) update(ctx context.Context) (reconcile.Result, error) {
...
+
+	return roles
+}
+
+func (h *stackHandler) createNamespaceClusterRoles(ctx context.Context) error {
+	ns := &corev1.Namespace{}
+	nsName := h.ext.GetNamespace()
+
+	if err := h.kube.Get(ctx, types.NamespacedName{Name: nsName}, ns); err != nil {
+		return errors.Wrapf(err, "failed to get namespace %q for stackinstall %q", nsName, h.ext.GetName())
```
jbw976: should this error message should say `stack` instead of `stackinstall`? (same for the error message below if creating a clusterrole fails)

pkg/controller/stacks/stack/stack.go
```diff
@@ -160,29 +188,169 @@ func (h *stackHandler) update(ctx context.Context) (reconcile.Result, error) {
...
+	}
+	roles := generateNamespaceClusterRoles(h.ext)
+
+	for _, role := range roles {
+		role.SetOwnerReferences([]metav1.OwnerReference{
+			{
+				APIVersion: "v1",
+				Kind:       "Namespace",
```
jbw976: is setting the ownerRef to the namespace explicitly required? I thought it was implicit that when a namespace is deleted that all resources within the namespace are deleted too.

pkg/controller/stacks/stack/stack.go
```diff
@@ -160,29 +188,169 @@ func (h *stackHandler) update(ctx context.Context) (reconcile.Result, error) {
...
+			return errors.Wrap(err, "failed to create persona cluster roles")
+		}
+	}
+	return nil
+}
+
+func generateNamespaceClusterRoles(stack *v1alpha1.Stack) (roles []*rbacv1.ClusterRole) {
+	personas := []string{"admin", "edit", "view"}
+
+	namespaced := (stack.Spec.PermissionScope == string(apiextensions.NamespaceScoped))
```
jbw976: there's a new function in this PR that does this same check, right? can we use that? `isNamespaced`

cluster/charts/crossplane/templates/crossplane-admin-clusterrole.yaml
```diff
@@ -0,0 +1,30 @@
...
+metadata:
+  name: {{ template "name" . }}-admin
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: {{ template "name" . }}-admin
+subjects:
+- apiGroup: rbac.authorization.k8s.io
+  kind: Group
+  name: {{ template "name" . }}:masters
```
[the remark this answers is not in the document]
displague: I had similar concerns that I expressed with @lukeweber.  My concern was that an LDAP/AD admin would be able to create a "crossplane:masters" group and add themselves to it.  In doing so they could perhaps grant themselves into Kubernetes powers that their LDAP/AD superiors did not intend.

The justification, and K8s symmetry here, can be seen with: `kubectl get clusterrolebinding cluster-admin -o yaml`

The intent to add this was expressed in https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-stacks-security-isolation.md#crossplane-clusterroles--rbac, and has been revised in this PR to correct `master` to `masters`.

> We will provide a convenience group - crossplane:master - that binds to crossplane-admin with a ClusterRoleBinding. This mirrors Kubernetes functionality for cluster-admin and system:master. At install time you will then be able to impersonate a group that is bound to crossplane-admin.
> 
> Example running a command as crossplane-admin
>
> `kubectl {any crossplane-admin operation} --as-group=crossplane:master`

pkg/controller/stacks/install/installjob.go
```diff
@@ -40,20 +42,11 @@ import (
...
-	labelParentNamespace = "core.crossplane.io/parent-namespace"
-	labelParentName      = "core.crossplane.io/parent-name"
-	labelParentUID       = "core.crossplane.io/parent-uid"
-)
-
 var (
 	jobBackoff                = int32(0)
 	registryDirName           = ".registry"
 	packageContentsVolumeName = "package-contents"
+	labelNamespaceFmt         = "namespace.crossplane.io/%s"
```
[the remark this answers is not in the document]
displague: I noticed this in my most recent squashing and rebasing.
This var will join the other constants at `stacks.Label*`

pkg/controller/stacks/install/stackinstall.go
```diff
@@ -255,24 +255,25 @@ func (h *stackInstallHandler) update(ctx context.Context) (reconcile.Result, err
...
-	gvk := h.ext.GroupVersionKind()
-	labels := map[string]string{
-		labelParentGroup:     gvk.Group,
-		labelParentVersion:   gvk.Version,
-		labelParentKind:      gvk.Kind,
-		labelParentNamespace: h.ext.GetNamespace(),
-		labelParentName:      h.ext.GetName(),
-		labelParentUID:       string(h.ext.GetUID()),
+	labels := stacks.ParentLabels(h.ext)
+	if err := h.kube.DeleteAllOf(ctx, &v1alpha1.Stack{}, client.InNamespace(h.ext.GetNamespace()), client.MatchingLabels(labels)); runtimeresource.IgnoreNotFound(err) != nil {
```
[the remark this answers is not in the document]
displague: Did you notice how this sets us up to have a single StackInstall resolving its requests/dependencies through multiple Stacks ;-) .. Ofcourse, the `status` of a `StackInstall` doesn't support multiple Stacks (yet).

It is definitely worth commenting that this could potentially remove multiple resources, and that one day that may be the case.

pkg/controller/stacks/stack/stack.go
```diff
@@ -18,29 +18,38 @@ package stack
...
 	"github.com/crossplaneio/crossplane-runtime/pkg/meta"
+	runtimeresource "github.com/crossplaneio/crossplane-runtime/pkg/resource"
 	"github.com/crossplaneio/crossplane/apis/stacks/v1alpha1"
+	"github.com/crossplaneio/crossplane/pkg/stacks"
 )
 
 const (
-	controllerName = "stack.stacks.crossplane.io"
+	controllerName     = "stack.stacks.crossplane.io"
+	clusterRoleNameFmt = "stack:%s:%s:%s:%s"
```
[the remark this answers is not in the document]
displague: @lukeweber had similar concerns, which is why the constant is only used in a function that applies the fields in the correct order given a Stack object and persona as an argument.

I could move the constant into that function to avoid anyone from using it directly, or add a comment to the constant incase we end up needing to use this string outside of that helper function.

pkg/controller/stacks/stack/stack.go
```diff
@@ -50,6 +59,12 @@ var (
 	log              = logging.Logger.WithName(controllerName)
 	resultRequeue    = reconcile.Result{Requeue: true}
 	requeueOnSuccess = reconcile.Result{RequeueAfter: requeueAfterOnSuccess}
+
+	roleVerbs = map[string][]string{
+		"admin": {"get", "list", "watch", "edit", "create", "delete"},
```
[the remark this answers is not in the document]
displague: `kubectl get clusterrole admin edit view  -o yaml` are fair comparisons.

Based on that, `edit` should pick 
```
  - create
  - delete
  - deletecollection
  - patch
  - update
```

and `admin` should include:

```
  - create
  - delete
  - deletecollection
  - get
  - list
  - patch
  - update
  - watch
```
```


I'll also remove the made-up verb, `edit` :man_facepalming:

File: pkg/controller/stacks/stack/stack.go [hunk 22]
```diff
@@ -136,6 +156,14 @@ func (h *stackHandler) sync(ctx context.Context) (reconcile.Result, error) {
 func (h *stackHandler) create(ctx context.Context) (reconcile.Result, error) {
 	h.ext.Status.SetConditions(runtimev1alpha1.Creating())
 
+	// Add the finalizer before the RBAC and Deployments. If the Deployment
+	// irreconcilably fails, the finalizer must be in place to delete the Roles
+	patchCopy := h.ext.DeepCopy()
```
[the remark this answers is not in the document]
displague: By watching the logs and running into errors, I found that the `Update` was failing because the `Stack` had already been changed (by `Status.SetConditions`?).  By using `Patch` this problem was avoided.

pkg/controller/stacks/stack/stack.go
```diff
@@ -160,29 +188,169 @@ func (h *stackHandler) update(ctx context.Context) (reconcile.Result, error) {
 	return reconcile.Result{}, nil
 }
 
-func (h *stackHandler) createNamespacedRoleBinding(ctx context.Context, owner metav1.OwnerReference) error {
-	cr := &rbac.Role{
+// createPersonaClusterRoles creates admin, edit, and view clusterroles that are
+// namespace+stack+version specific
+func (h *stackHandler) createPersonaClusterRoles(ctx context.Context, labels map[string]string) error {
+	// TODO(displague) refine 'editor' persona roles
```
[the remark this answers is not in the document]
displague: Absolutely.  This was a flashing neon light for reviewers.  

I think https://github.com/crossplaneio/crossplane/pull/1059#discussion_r359944099 will settle this `TODO`, along with any feedback there.

pkg/controller/stacks/stack/stack.go
```diff
@@ -160,29 +188,169 @@ func (h *stackHandler) update(ctx context.Context) (reconcile.Result, error) {
...
+			})
+		}
+		name := fmtPersonaRole(h.ext, persona)
+
+		var crossplaneScope string
+
+		if h.isNamespaced() {
+			crossplaneScope = "namespace"
+		} else {
+			crossplaneScope = "environment"
```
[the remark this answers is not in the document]
displague: They are similar to `apiextensions.NamespaceScoped` and `apiextensions.ClusterScoped`, but neither is identical.

As `apiextensions` considered these worthy of making public constants, I can do the same as `stacks.NamespaceScoped` and `stacks.EnvironmentScoped`.  We'll have to keep in mind not to confuse or compare the two sets.

pkg/controller/stacks/stack/stack.go
displague: We are creating a Cluster scoped `ClusterRole` with a Cluster scoped `Namespace` as the owner.

This is not the same as setting a namespaced `Role` within a `Namespace`, where the behavior you suggested would apply.

If the namespace is deleted, the clusterroles that are bound to that namespace (by name only) should go away, which is why OwnerReferences are a convenient approach.

When we take on https://github.com/crossplaneio/crossplane/issues/1124 we could consider letting the stack-manager clean up these clusterroles if it detects that the namespace was deleted, but I think we will still find OwnerReferences to be more reliable (especially if stack-manager is uninstalled).

pkg/controller/stacks/stack/stack.go
displague: Indeed there is.. :point_left: :nerd_face: :airplane:

displague:
The changes since my review from @jbw976 can be seen here https://gist.github.com/displague/5b955fb471b6fcd9a2918212594d82df

negz:
Should we be using the term `env` to describe "cluster wide" roles? It's not a concept that exists in Kubernetes, or in the open source Crossplane project. I'm wary about introducing it if we could instead use a more Kubernetes native term.

lukeweber:
Regards to env role name: we have crossplane-admin, env-admin, and ns-admin levels. If we wanted to rename env admin which just means the globally scoped resources to maybe cluster-admin I think that would cause more naming confusion as cluster in kubernetes means global + namespaced. Since our infra owner view is really just the global scope but not “cluster” in the kubernetes RBAC sense I think this is probably the best choice as it’s been built, but open to better ideas that get around this confusion.

negz:
> cluster in kubernetes means global + namespaced

Would an explicit name like `crossplane-cluster-scoped-admin` help with this concern?

lukeweber:
Also I should note to make things just a little more weird, “env” scope also includes at least one namespace to store secrets for objects in the global space. As well crds are cluster scoped but this role will not manage crds. Personally my vote is to describe env as a new concept, because it’s concise and mirrors upbound internal naming and there isn’t great terminology available to describe this view in kubernetes as it’s an RBAC curated slice of global scope. A workspace is a namespace and in this case made sense to drop workspace. I’m not feeling it as much in this case.

pkg/controller/stacks/stack/stack.go
```diff
@@ -160,29 +184,233 @@ func (h *stackHandler) update(ctx context.Context) (reconcile.Result, error) {
...
+	}
+	return labelsCopy
+}
+
+func (h *stackHandler) crdListsDiffer(crds []apiextensions.CustomResourceDefinition) bool {
+	return len(crds) != len(h.ext.Spec.CRDs)
+}
+
+// crdsFromStack fetches the CRDs of the Stack using the shared parent labels
+// TODO(displague) change this to use GET on each, the CRDs may not be ready
```
displague: This TODO is arguable now.  

I think the List approach and length comparison is infallible outside of admin intervention (by setting false parent labels).

Once the expected count of CRDs are accounted for, the stack can't have a discrepancy in the matched CRDs and listed CRDs because there is no inplace upgrading of Stacks (and their CRDs).

When we revisit the scenario of Foo Stack having instances residing in multiple namespaces, parent labels won't be sufficient.  We'll have to change this behavior or the approach to StackInstall/Stacks at that time.

If an expected CRD somehow survived the previous instance of the stack, the labels won't match, so the Stack won't sync (expected) until the CRD or alternate copy of the Stack is removed.

The TODO can stand and be removed when we revisit this behavior and solve the multiple namespace stack installation problem.

pkg/controller/stacks/stack/stack.go
```diff
@@ -50,6 +55,12 @@ var (
 	log              = logging.Logger.WithName(controllerName)
 	resultRequeue    = reconcile.Result{Requeue: true}
 	requeueOnSuccess = reconcile.Result{RequeueAfter: requeueAfterOnSuccess}
+
+	roleVerbs = map[string][]string{
+		"admin": {"get", "list", "watch", "create", "delete", "deletecollection", "patch", "update"},
+		"edit":  {"get", "list", "watch", "create", "delete", "deletecollection", "patch", "update"},
```
jbw976: are admin and edit supposed to be the same here?

displague: We originally had an invalid `edit` role prescribed in the design doc. `update` and `patch` seem reasonable, but the Kubernetes `edit` role adopts create, delete, and delete collection as well. 

`admin` gets the advantage of working with secrets.

@lukeweber thoughts?

lukeweber: The goal was to model based on upstream kubernetes verb behavior for their built in roles. 

https://kubernetes.io/docs/reference/access-authn-authz/rbac/

admin | None | Allows admin access, intended to be granted within a namespace using a RoleBinding. If used in a RoleBinding, allows read/write access to most resources in a namespace, including the ability to create roles and rolebindings within the namespace. It does not allow write access to resource quota or to the namespace itself.
-- | -- | --
edit | None | Allows read/write access to most objects in a namespace. It does not allow viewing or modifying roles or rolebindings.
view | None | Allows read-only access to see most objects in a namespace. It does not allow viewing roles or rolebindings. It does not allow viewing secrets, since those are escalating.

From this, admin is manage role bindings and all behavior of edit, edit can manage most objects except the bindings, and view can view most objects except secrets.

displague: Good point, @lukeweber.

I've added rolebinding and clusterrolebinding rules in 83e7fbf3

cluster/charts/crossplane/templates/environment-personas-clusterroles.yaml
```diff
@@ -0,0 +1,180 @@
...
+  resources:
+  - "*"
+  verbs:
+  - create
```
jbw976: should this set of verbs match what's listed in https://github.com/displague/crossplane/blob/gh1049-stack-controllers-use-clusterrole/pkg/controller/stacks/stack/stack.go#L61?

cluster/charts/crossplane/templates/namespace-personas-clusterroles.yaml
```diff
@@ -0,0 +1,136 @@
...
+  - cache.crossplane.io
+  - compute.crossplane.io
+  - database.crossplane.io
+  - kubernetes.crossplane.io
+  - core.crossplane.io
+  - storage.crossplane.io
+  - workload.crossplane.io
+  resources:
+  - "*"
+  verbs:
```
jbw976: same for these verbs, should they match the edit roleVerbs in https://github.com/displague/crossplane/blob/gh1049-stack-controllers-use-clusterrole/pkg/controller/stacks/stack/stack.go#L61?

cluster/charts/crossplane/templates/environment-personas-clusterroles.yaml
displague: The shorter list of verbs are provided because these `default` roles are only intended for internal user.  The `-edit` and `-admin` clusterroles pick up verbs from both the default `view` and `edit` clusterroles.

The verbs are cumulative.

cluster/charts/crossplane/templates/namespace-personas-clusterroles.yaml
displague: https://github.com/crossplaneio/crossplane/pull/1059#discussion_r360626059

PR #1060, opened 2019-11-14
Don't use muvaffak's personal resource group for SQL instances

negz:
### Checklist

I have:

### DIFF
--- a/cluster/examples/database/mysqlinstance/azure/resource-class.yaml
+++ b/cluster/examples/database/mysqlinstance/azure/resource-class.yaml
@@ -10,7 +10,7 @@ metadata:
 specTemplate:
   forProvider:
     administratorLogin: myadmin
-    resourceGroupName: muvaf-group-westus-1
+    resourceGroupName: example-group
     location: West US
     sslEnforcement: Disabled
     version: "5.6"


cluster/examples/database/mysqlinstance/azure/resource-class.yaml
```diff
@@ -10,7 +10,7 @@ metadata:
 specTemplate:
   forProvider:
     administratorLogin: myadmin
-    resourceGroupName: muvaf-group-westus-1
+    resourceGroupName: example-group
```
jbw976: looks like this group name used to be `group-westus-1`, do you want to restore it back to that? (or does it not matter)

https://github.com/crossplaneio/crossplane/blob/b3e96eeae7fd9330a55e352ea7bee4fafec96eb4/cluster/examples/database/mysqlinstance/azure/resource-class.yaml#L13

jbw976:
side note: this makes me wonder how many of these examples still work and are correct with up to date paradigms 🤔

cluster/examples/database/mysqlinstance/azure/resource-class.yaml
hasheddan: @jbw976 whatever group is listed here will have to have been previously created, so whatever we put here isn’t valid by default. In general, many resources are dependent on the existence (or absence) of other resources, so unless a resource is part of a full guide, we are only basically saying “this will not be rejected by the API server”

negz:
> this makes me wonder how many of these examples still work and are correct with up to date paradigms

We've been updating and checking them as we go (for example this group name was introduced when Azure was bumped to v1beta1). That said, I think trying to maintain this big bundle of examples all here in crossplaneio/crossplane is a losing game - we need a better strategy.

PR #1071, opened 2019-11-15
CD Workflow: GitLab guide

prasek:
Adds a CD Workflow guide for GitLab

### Checklist

I have:

[skip ci]


docs/workflow/gitlab.md
```diff
@@ -0,0 +1,43 @@
...
+weight: 592
+indent: true
+---
+
+# Continuous Delivery with GitLab
+
+## GitLab
+[GitLab][GitLab Kubernetes docs] supports deploying Kubernetes applications
+using GitLab pipelines or the GitLab Auto Deploy App today, but cloud service
+dependencies had to be separately provisioned, connected, and secured.
```
hasheddan:
```suggestion
dependencies have to be separately provisioned, connected, and secured. 
```

prasek: Good catch, thx!

docs/workflow/gitlab.md
```diff
@@ -0,0 +1,43 @@
...
+weight: 592
+indent: true
+---
+
+# Continuous Delivery with GitLab
+
+## GitLab
+[GitLab][GitLab Kubernetes docs] supports deploying Kubernetes applications
+using GitLab pipelines today, but cloud service dependencies must be
+separately provisioned, connected, and secured.
```
negz: I think we could provide a little more context here. This is succinct to the extent that it comes across to me more like a criticism of GitLab than (what I assume should be) an introduction to the value Crossplane can provide atop GitLab.

docs/workflow/gitlab.md
```diff
@@ -0,0 +1,43 @@
...
+
+# Continuous Delivery with GitLab
+
+## GitLab
+[GitLab][GitLab Kubernetes docs] supports deploying Kubernetes applications
+using GitLab pipelines today, but cloud service dependencies must be
+separately provisioned, connected, and secured. 
+
+## Provision cloud services for Kubernetes apps with Crossplane
+With GitLab 12.5, Crossplane is now available as a GitLab-managed app that can
```
negz: Should we provide context on what a GitLab-managed app is? Perhaps a link?

docs/workflow/gitlab.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Securely connect cloud services to Kubernetes
+Crossplane is designed to integrate with existing cloud networks, projects, and
+resource groups, so managed services like RDS, CloudSQL, and Azure DB can be
+provisioned and securely consumed by Kubernetes applications. This includes
+[secure connectivity] between the Kubernetes cluster and the managed cloud
+services.
+
+## Next Steps
+Get started with [GitLab and Crossplane][GitLab Crossplane docs] today!
```
negz: Is the purpose of this document to direct readers to the GitLab-hosted documentation about this integration? This reads to me more like a sales pitch than documentation as it stands. Perhaps this should be a little more front and center?

docs/workflow/gitlab.md
prasek: this is the same intro GitLab is including in their 12.5 release notes:
https://gitlab.com/gitlab-com/www-gitlab-com/merge_requests/34721/diffs

docs/workflow/gitlab.md
prasek: good catch, link added

docs/workflow/gitlab.md
prasek: yes, providing a short overview and a link to the GitLab docs is the main purpose, so it's discoverable from the crossplane.io/docs, without recreating the guide in the GitLab docs.

re: front and center: yes added a link at the top of the doc as well

Issue #1064, opened 2019-11-14
helm install is not working.


bshetti:
### What happened?
Using helm 3 and ran 
helm install crossplane crossplane-alpha/crossplane --namespace crossplane-system

we get
NAMESPACE           NAME                                                        READY     STATUS             RESTARTS   AGE
crossplane-system   crossplane-cd67688d6-2flrh                                  1/1       Running            0          17s
crossplane-system   crossplane-stack-manager-564bcb979-rj9mg                    0/1       CrashLoopBackOff   1          17s

Here is the LOG info for the pod that is crossplane-stack-manager:

```bash
ubuntu@ip-172-31-9-213:~$ kubectl logs crossplane-stack-manager-564bcb979-rj9mg -n crossplane-system
2019-11-14T08:07:50.718Z	INFO	crossplane	Sync period	{"duration": "1h0m0s"}
2019-11-14T08:07:51.264Z	INFO	controller-runtime.metrics	metrics server is starting to listen	{"addr": ":8080"}
2019-11-14T08:07:51.265Z	INFO	crossplane	Adding schemes
2019-11-14T08:07:51.266Z	INFO	crossplane	Adding controllers
2019-11-14T08:07:51.269Z	INFO	controller-runtime.controller	Starting EventSource	{"controller": "stackinstall.stacks.crossplane.io", "source": "kind source: /, Kind="}
2019-11-14T08:07:51.269Z	ERROR	controller-runtime.source	if kind is a CRD, it should be installed before calling Start	{"kind": "StackInstall.stacks.crossplane.io", "error": "no matches for kind \"StackInstall\" in version \"stacks.crossplane.io/v1alpha1\""}
github.com/go-logr/zapr.(*zapLogger).Error
	/home/upbound/go/pkg/mod/github.com/go-logr/zapr@v0.1.1/zapr.go:128
sigs.k8s.io/controller-runtime/pkg/source.(*Kind).Start
	/home/upbound/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.2.0/pkg/source/source.go:88
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Watch
	/home/upbound/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.2.0/pkg/internal/controller/controller.go:122
sigs.k8s.io/controller-runtime/pkg/builder.(*Builder).doWatch
[...]

Issue #1062, opened 2019-11-14
Crossplane helm chart: Crossplane installation using helm3 fails because `crd-install` hook is deprecated


soorena776:
### What happened?

Installing Crossplane using helm3:

```bash
helm3 install crossplane  --namespace crossplane-system crossplane-alpha/crossplane 
```

Produces the output:

```bash
manifest_sorter.go:175: info: skipping unknown hook: "crd-install"
manifest_sorter.go:175: info: skipping unknown hook: "crd-install"
NAME: crossplane
LAST DEPLOYED: Wed Nov 13 20:39:46 2019
NAMESPACE: crossplane-system
STATUS: deployed
REVISION: 1
TEST SUITE: None
NOTES:
Release: crossplane

Chart Name: crossplane
Chart Description: Crossplane - Managed Cloud Resources Operator
Chart Version: 0.4.1
Chart Application Version: 0.4.1
```

However, running `kubectl get pods -A` results:

```bash
NAMESPACE           NAME                                         READY   STATUS             RESTARTS   AGE
crossplane-system   crossplane-6cf4c86df9-dcfh9                  1/1     Running            0          10s
crossplane-system   crossplane-stack-manager-59bb94f5d-nvfzt     0/1     CrashLoopBackOff   1          10s
```

After some digging, it turned out this is because two of the CRDs (`StackInstall` and `ClusterStackInstall`) that have [`helm.sh/hook: crd-install`](https://github.com/crossplaneio/crossplane/blob/master/cluster/charts/crossplane/templates/crds/stacks.crossplane.io_stackinstalls.yaml#L5) annotation, are skipped from being created (because this hook is deprecated in [v3](https://helm.sh/docs/topics/charts_hooks/#the-available-hooks)), hence causing the corresponding stack-manager controller to fail.

### Potential fix

Since helm v3 is now stable, and supports `crds/` folder, we can increase the minimum required version of helm to v3 in our [doc](https://github.com/crossplaneio/crossplane/blob/master/docs/install-crossplane.md#installation), and then remove this hook.

### What environment did it happen in?
Crossplane version: v0.4.1
helm version: v3.0.0

### Related
#839 
#1027

displague:
Noting what the current experience looks like from the master channel:

```
$ kind create cluster --name foo
$ kubectl config set-context kind-foo
$ helm3 repo add crossplane-master  https://charts.crossplane.io/master/
$ helm3 repo update
$ kubectl create namespace crossplane-system
[...]

PR #1077, opened 2019-11-15
Workaround for helm3 compatibility

### LINKED ISSUE #943: Helm deployment fails with helm3 when "deploying clusterStacks with helm" enabled
### What happened?

The "cluster stack installation with Crossplane helm chart" feature introduced with #932 is not working with helm 3.

### How can we reproduce it?

Run the following to deploy Crossplane helm chart with helm3 enabling at least one clusterStack:
```bash
$ helm3 install crossplane-test --namespace crossplane-system -f values.yaml . --set clusterStacks.aws.deploy=true
Error: apiVersion "stacks.crossplane.io/v1alpha1" in crossplane/templates/clusterstackinstalls.yaml is not available
```

Crossplane version: master including #932 
Helm version: v3.0.0-beta.4

turkenh:
Helm 3 introduced some breaking changes one of which is [deprecating `crd-install` hook](https://helm.sh/docs/topics/charts_hooks/#the-available-hooks) and skipping CRDs with this annotation which causes some of the [existing helm charts to fail](https://github.com/helm/charts/issues/9597). We need these hooks for `stackinstall` and `clusterstackinstall` CRDs to be able to [install infrastructure stacks with helm](https://github.com/crossplaneio/crossplane/pull/932). 

Ideally we should drop helm2 support and specify helm3 as minimum version as suggested by @soorena776 [here](https://github.com/crossplaneio/crossplane/issues/1062) as potential fix. However, there are still users (like [gitlab](https://gitlab.com/groups/gitlab-org/-/epics/1866#note_227663961)) who depend on helm2 and we need to support it at least for some more time.

One workaround could be using some sort of built-in variable which lets us to understand it is helm 3 and we do not put annotation if so, however [there seems to be no such variable yet](https://github.com/helm/helm/issues/6689).

As a workaround, this PR copies the two crds, without `crd-install` hook annotation, under `<chart>/crds` directory which is introduced with helm3 but ignored in helm2. 

So, with this change, we will have `clusterstackinstalls` and `stackinstalls` CRD yamls:

- with hook under `<crossplane-chart>/templates/crds`: for `helm2` skipped in `helm3`
- without hook under `<crossplane-chart>/crds`: for `helm3` ignored in `helm2`

Fixes #943, #1062, #1064 
### Checklist

I have:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,7 +83,6 @@ manifests:
 
 generate: $(KUSTOMIZE) go.generate manifests.annotate
 
-
 # Add "helm.sh/hook: crd-install" and "helm.sh/hook-delete-policy:
 # before-hook-creation" annotations for clusterstackinstalls and stackinstalls
 # CRDs. Since Crossplane helm chart contains both CRD and ClusterStackInstall
@@ -96,8 +95,20 @@ generate: $(KUSTOMIZE) go.generate manifests.annotate
 # policy is necessary to be able to redeploy helm chart after it is deleted
 # since CRDs with "crd-install" hooks will not be deleted with "helm delete" and
 # cause next "helm install" to fail.
-CRD_DIR = cluster/charts/crossplane/templates/crds
+# "helm.sh/hook: crd-install" was deprecated in helm3 and CRDs with this annotation are
+# skipped. This results in the StackInstall and ClusterStackInstall CRDs not being
+# installed in helm3 when they have that annotation.
+# As a workaround, we first copy those CRDs under <chart>/crds directory which
+# was introduced with helm3 and ignored in helm2, then afterwards apply the annotation to
+# those CRDs under <chart>/templates/crds for helm2.
+CROSSPLANE_CHART_DIR = cluster/charts/crossplane
+CRD_DIR = $(CROSSPLANE_CHART_DIR)/templates/crds
+HELM3_CRD_DIR = $(CROSSPLANE_CHART_DIR)/crds
 manifests.annotate:
+	@$(INFO) Copying StackInstall CRD manifests for helm3 compatibility
+	cp $(CRD_DIR)/stacks.crossplane.io_stackinstalls.yaml $(HELM3_CRD_DIR)/stacks.crossplane.io_stackinstalls.yaml
+	cp $(CRD_DIR)/stacks.crossplane.io_clusterstackinstalls.yaml $(HELM3_CRD_DIR)/stacks.crossplane.io_clusterstackinstalls.yaml
+	@$(OK) Copied StackInstall CRD manifests for helm3 compatibility
 	@$(INFO) Annotating generated StackInstall CRD manifests
 	$(eval TMPDIR := $(shell mktemp -d))
 	$(KUSTOMIZE) build cluster/charts -o $(TMPDIR)
--- a/cluster/charts/crossplane/crds/stacks.crossplane.io_clusterstackinstalls.yaml
+++ b/cluster/charts/crossplane/crds/stacks.crossplane.io_clusterstackinstalls.yaml
@@ -1,0 +1,124 @@
+
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  creationTimestamp: null
+  name: clusterstackinstalls.stacks.crossplane.io
+spec:
+  additionalPrinterColumns:
+  - JSONPath: .status.conditionedStatus.conditions[?(@.type=='Ready')].status
+    name: READY
+    type: string
+  - JSONPath: .spec.source
+    name: SOURCE
+    type: string
+  - JSONPath: .spec.package
+    name: PACKAGE
+    type: string
+  - JSONPath: .spec.crd
+    name: CRD
+    type: string
+  - JSONPath: .metadata.creationTimestamp
+    name: AGE
+    type: date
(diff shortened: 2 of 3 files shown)


soorena776:
@turkenh Thanks for the quick fix! As for backward compatibility, my understanding is that Gitlab is using an already tagged version of Crossplane (e.g. v0.4), so not supporting helm2 going forward shouldn't break their existing flow. Isn't that the case @prasek and @jbw976?

Makefile
```diff
@@ -83,7 +83,18 @@ manifests:
 
 generate: $(KUSTOMIZE) go.generate manifests.annotate
 
-
+CROSSPLANE_CHART_DIR=cluster/charts/crossplane
+CRD_DIR = $(CROSSPLANE_CHART_DIR)/templates/crds
+HELM3_CRD_DIR = $(CROSSPLANE_CHART_DIR)/crds
+manifests.annotate:
```
displague: nitpic about spacing and flow..

The comment block that was above manifests.annotate should stay above the `manifests.annotate:` variables.  Your new comments should be incorporated into that comment block.

https://github.com/crossplaneio/crossplane/pull/1077/files#diff-b67911656ef5d18c4ae36cb6741b7965R86 should match the spacing of the other two variables.

jbw976:
@soorena776 I thought GitLab managed apps was using whatever the latest is from the alpha charts repo, but you might be right that they are actually always using v0.4.1 from the alpha repo.  I'm confirming with @mahendrabagul in this note on the MR: https://gitlab.com/gitlab-org/gitlab/merge_requests/18797/diffs#note_245343086

Makefile
turkenh: Thanks, both should be done now.

hasheddan:
@turkenh any update on this PR? Does is need further review?

turkenh:
@hasheddan PR is ready from my POV.

But from the discussions above it looks like we can also drop helm 2 support and adapt the chart only for helm 3, so we wouldn't need this workaround. @jbw976, I see that Gitlab using a hardcoded version of  Crossplane Helm chart, does this mean we can drop Helm 2 support? Wouldn't they need to update Crossplane version to get new feature and bug fixes? 

In any case, my suggestion would be to merge this workaround ASAP and follow dropping Helm 2 support with [another issue](https://github.com/crossplaneio/crossplane/issues/1100).

jbw976:
>  @jbw976, I see that Gitlab using a hardcoded version of Crossplane Helm chart, does this mean we can drop Helm 2 support? Wouldn't they need to update Crossplane version to get new feature and bug fixes?
Right, in GitLab I imagine they will update to v0.5 and v0.6 over time, which could very well be before GitLab moves to Helm 3.  I think we need to maintain compatibility with Helm 2 for an undetermined amount of time going forward still.

Makefile
```diff
@@ -96,8 +95,18 @@ generate: $(KUSTOMIZE) go.generate manifests.annotate
 # policy is necessary to be able to redeploy helm chart after it is deleted
 # since CRDs with "crd-install" hooks will not be deleted with "helm delete" and
 # cause next "helm install" to fail.
-CRD_DIR = cluster/charts/crossplane/templates/crds
+# "helm.sh/hook: crd-install" deprecated in helm3 and crds with this annotation
```
jbw976: Could this comment be cleaned up and explained just a bit more?  e.g.,

```make
# "helm.sh/hook: crd-install" was deprecated in helm3 and CRDs with this annotation are
# skipped. This results in the StackInstall and ClusterStackInstall CRDs not being
# installed in helm3 when they have that annotation.
# As a workaround, we first copy those CRDs under <chart>/crds directory which
# was introduced with helm3 and ignored in helm2, then afterwards apply the annotation to
# those CRDs under <chart>/templates/crds for helm2.
```

turkenh: Done! Thanks for the suggested text.

turkenh:
> I assume you tested this for both helm2 and helm3?

Yes, tested with both versions (v2.15.1 & v3.0.0) with/without cluster stacks.

> On helm3, is the <chart>/templates/crds directory ignored? How does helm3 know not to just render everything under <chart>/templates?

No, helm3 already renders everything under `<chart>/templates/crds` but skips the two with `crd-install` hook. Those two comes from `<chart>/crds` as helm3 recognizes this directory.

infinitecompute:
When this is merged, what will the end state be regarding installation from alpha when using either Helm 2 or 3?  Will the docs need a new section describing how to install with Helm 3 changes (e.g. generating the namespace prior to running helm, and how to get pull down the workaround for <chart>/templates/crds)?  When I do `brew install helm`, I am now on helm 3 by default, so I have to decide whether to pin myself to the older version.

turkenh:
> When this is merged, what will the end state be regarding installation from alpha when using either Helm 2 or 3? Will the docs need a new section describing how to install with Helm 3 changes (e.g. generating the namespace prior to running helm?

There should be no difference regarding installing with helm 2 or helm 3, except:
- With helm3, namespace to be created prior to install as you pointed out, but I see it is already documented [here](https://github.com/crossplaneio/crossplane/blob/master/docs/install-crossplane.md#installation).
- With helm3, Following two lines of log which would be totally harmless:
```
manifest_sorter.go:175: info: skipping unknown hook: "crd-install"
manifest_sorter.go:175: info: skipping unknown hook: "crd-install"
```

PR #1087, opened 2019-11-20
docs: update quickstart to use cloud memorystore

### LINKED ISSUE #1083: Change docs Quickstart to use CloudMemorystoreInstance

The current [quickstart](https://crossplane.io/docs/v0.5/quick-start.html) walks through statically provisioning a `CloudSQLInstance`. However, this can fail due to the fact that GCP does not allow reuse of instance name [for a week following deletion](https://cloud.google.com/sql/docs/mysql/delete-instance).

The quickstart should be updated to use a resource that does not have this same constraint. A simple solution would be to provision a GCP `CloudMemorystoreInstance`.

hasheddan:

Fixes #1083 

### Checklist

I have:

### DIFF
--- a/docs/quick-start.md
+++ b/docs/quick-start.md
@@ -5,9 +5,9 @@ weight: 210
 ---
 # Getting Started
 
-This guide will demonstrate using Crossplane to deploy a portable MySQL database
-on the Google Cloud Platform (GCP). It serves as an initial introduction to
-Crossplane, but only displays a small set of its features.
+This guide will demonstrate using Crossplane to deploy a portable Cloud
+Memorystore on the Google Cloud Platform (GCP). It serves as an initial
+introduction to Crossplane, but only displays a small set of its features.
 
 In this guide we will:
 
@@ -174,9 +174,9 @@ NAME                  STATUS   CLASS-KIND   CLASS-NAME   RESOURCE-KIND          
 example-redis-claim   Bound                              CloudMemorystoreInstance   example-cloudmemorystore-instance   8m39s
 ```
 
-You'll find all the details you need to connect to your new MySQL instance saved
-in the Kubernetes `Secret` you specified via `writeConnectionSecretToRef`, ready
-to [use with your Kubernetes pods].
+You'll find all the details you need to connect to your new Redis cluster
+instance saved in the Kubernetes `Secret` you specified via
+`writeConnectionSecretToRef`, ready to [use with your Kubernetes pods].
 
 ```bash
 $ kubectl --namespace default describe secret example-redis-connection-details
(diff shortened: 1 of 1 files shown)


docs/quick-start.md
```diff
@@ -5,16 +5,18 @@ weight: 210
 ---
 # Getting Started
 
-This guide will demonstrate using Crossplane to deploy a portable MySQL database
-on the Google Cloud Platform (GCP). It serves as an initial introduction to
-Crossplane, but only displays a small set of its features.
+This guide will demonstrate using Crossplane to deploy a portable Cloud
+Memorystore on the Google Cloud Platform (GCP). It serves as an initial
```
negz: Portable Redis cluster would be more in line with the previous documentation.

docs/quick-start.md
```diff
@@ -80,278 +82,274 @@ Save the above `Provider` as `provider.yaml`, save your Google Application
...
-      Ip Configuration:
-        ipv4Enabled:  true
-      Location Preference:
-        Zone:               us-west2-b
-      Pricing Plan:         PER_USE
-      Replication Type:     SYNCHRONOUS
-      Storage Auto Resize:  true
-      Tier:                 db-n1-standard-1
+    Alternative Location Id:  us-west2-b
+    Authorized Network:       projects/crossplane-playground/global/networks/default
```
negz: We should avoid leaking our playground project's name in our docs and examples. I suggest finding and replacing `crossplane-playground` with some other value.

docs/quick-start.md
```diff
@@ -80,278 +82,274 @@ Save the above `Provider` as `provider.yaml`, save your Google Application
...
-labels]. Resource claims like `MySQLInstance` can match different kinds of
-resource class using label selectors, so you could just as easily use the
-exact same `MySQLInstance` to create an Amazon Relational Database Service (RDS)
-instance by creating an `RDSInstanceClass` labelled as `guide: getting-started`.
-When multiple resource classes match the class selector, a matching class is
-chosen at random. Claims can be matched to classes by either:
+labels]. Resource claims like `RedisCluster` can match different kinds of
+resource class using label selectors, so you could just as easily use the exact
+same `RedisCluster` to create an Amazon Replication Group instance by creating a
+`ReplicationGroupsClass` labelled as `guide: getting-started`. When multiple
```
negz:
```suggestion
`ReplicationGroupClass` labelled as `guide: getting-started`. When multiple
```

hasheddan:
Thanks for the review @negz! All comments addressed :)

hasheddan:
@negz @jbw976 do you think this is worth backporting to 0.5?

negz:
I think the problem this is solving is that if someone attempts to follow the static provisioning section of the CloudSQL quick start guide _multiple times_ they'll encounter an error on the subsequent attempts due to CloudSQL resource names not being reusable for a few days after the old resource is deleted. The error is easily worked around by using a different resource name, but it's not obvious that you need to do that.

Given it's straightforward to backport this commit to the 0.5 docs branch we may as well do so, but it doesn't feel like a priority to me.

PR #1091, opened 2019-11-21
Cleanup CRDs by label, not garbage collection

negz:

Fixes https://github.com/crossplaneio/crossplane/issues/895

(Cluster)StackInstall is namespaced, and CRDs are cluster scoped. Namespaced resources cannot own cluster scoped resources, so we can't rely on garbage collection to cleanup CRDs that were installed by a StackInstall. This uses labels instead.

I have removed the draft qualifier on this PR now so it is ready for full review.  The approach here uses the labels that are defined in the [stack relationship labels design doc](https://github.com/negz/crossplane/blob/refrace/design/one-pager-stack-relationship-labels.md), which has been updated by this PR also to be in alignment with the current implementation.

I think that this is a good approach for a `v0.5.1` patch release, but @displague is thinking about the longer term approach and how it aligns with ownership, garbage collection, multiple stack installs and versions, parent / child relationships, etc. 

Below is an example of a labeled CRD after stack-gcp has been installed:

```yaml
> kubectl get crd gkeclusters.compute.gcp.crossplane.io -o yaml
...
labels:
    app.kubernetes.io/managed-by: stack-manager
    core.crossplane.io/parent-group: stacks.crossplane.io
    core.crossplane.io/parent-kind: ClusterStackInstall
    core.crossplane.io/parent-name: stack-gcp
    core.crossplane.io/parent-namespace: gcp
    core.crossplane.io/parent-uid: 858ab465-ff60-49d8-a3e7-624ce841f339
    core.crossplane.io/parent-version: v1alpha1
name: gkeclusters.compute.gcp.crossplane.io
...
```

### Checklist
I have:

### DIFF
--- a/pkg/controller/stacks/install/installjob.go
+++ b/pkg/controller/stacks/install/installjob.go
@@ -42,6 +42,8 @@ import (
 
 // Labels used to track ownership across namespaces and scopes.
 const (
+	labelParentGroup     = "core.crossplane.io/parent-group"
+	labelParentVersion   = "core.crossplane.io/parent-version"
 	labelParentKind      = "core.crossplane.io/parent-kind"
 	labelParentNamespace = "core.crossplane.io/parent-namespace"
 	labelParentName      = "core.crossplane.io/parent-name"
@@ -239,7 +241,8 @@ func (jc *stackInstallJobCompleter) createJobOutputObject(ctx context.Context, o
 	}
 
 	// when the current object is a Stack object, make sure the name and namespace are
-	// set to match the current StackInstall (if they haven't already been set)
+	// set to match the current StackInstall (if they haven't already been set). Also,
+	// set the owner reference of the Stack to be the StackInstall.
 	if isStackObject(obj) {
 		if obj.GetName() == "" {
 			obj.SetName(i.GetName())
@@ -257,13 +260,16 @@ func (jc *stackInstallJobCompleter) createJobOutputObject(ctx context.Context, o
 	// on garbage collection because a namespaced object (StackInstall) can't
 	// own a cluster scoped object (CustomResourceDefinition), so we use labels
 	// instead.
-	// TODO(displague) is it sufficient to provide the "Kind" and not full type?
-	meta.AddLabels(obj, map[string]string{
-		labelParentKind:      i.GroupVersionKind().Kind,
+	gvk := i.GroupVersionKind()
+	labels := map[string]string{
+		labelParentGroup:     gvk.Group,
+		labelParentVersion:   gvk.Version,
+		labelParentKind:      gvk.Kind,
 		labelParentNamespace: i.GetNamespace(),
 		labelParentName:      i.GetName(),
 		labelParentUID:       string(i.GetUID()),
-	})
+	}
+	meta.AddLabels(obj, labels)
 
 	// TODO(displague) pass/inject a controller specific logger
 	log.V(logging.Debug).Info(
@@ -273,10 +279,12 @@ func (jc *stackInstallJobCompleter) createJobOutputObject(ctx context.Context, o
 		"namespace", obj.GetNamespace(),
 		"apiVersion", obj.GetAPIVersion(),
 		"kind", obj.GetKind(),
-		"parentKind", i.GroupVersionKind().Kind,
-		"parentName", i.GetName(),
-		"parentNamespace", i.GetNamespace(),
-		"parentUID", string(i.GetUID()),
+		"parentGroup", labels[labelParentGroup],
+		"parentVersion", labels[labelParentVersion],
+		"parentKind", labels[labelParentKind],
+		"parentName", labels[labelParentName],
+		"parentNamespace", labels[labelParentNamespace],
+		"parentUID", labels[labelParentUID],
 	)
 	if err := jc.client.Create(ctx, obj); err != nil && !kerrors.IsAlreadyExists(err) {
 		return errors.Wrapf(err, "failed to create object %s from job output %s", obj.GetName(), job.Name)
--- a/pkg/controller/stacks/install/installjob.go
+++ b/pkg/controller/stacks/install/installjob.go
@@ -40,6 +40,13 @@ import (
 	"github.com/crossplaneio/crossplane/pkg/stacks"
 )
 
+// Labels used to track ownership across namespaces and scopes.
+const (
+	labelInstallNamespace = "stacks.crossplane.io/install-namespace"
+	labelInstallName      = "stacks.crossplane.io/install-name"
+	labelInstallUID       = "stacks.crossplane.io/install-uid"
+)
+
 var (
 	jobBackoff                = int32(0)
 	registryDirName           = ".registry"
@@ -236,11 +243,21 @@ func (jc *stackInstallJobCompleter) createJobOutputObject(ctx context.Context, o
 		if obj.GetNamespace() == "" {
 			obj.SetNamespace(i.GetNamespace())
 		}
+
+		obj.SetOwnerReferences([]metav1.OwnerReference{
+			meta.AsOwner(meta.ReferenceTo(i, i.GroupVersionKind())),
+		})
 	}
 
-	// set an owner reference on the object
-	obj.SetOwnerReferences([]metav1.OwnerReference{
-		meta.AsOwner(meta.ReferenceTo(i, i.GroupVersionKind())),
+	// We want to clean up any installed CRDS when we're deleted. We can't rely
+	// on garbage collection because a namespaced object (StackInstall) can't
+	// own a cluster scoped object (CustomResourceDefinition), so we use labels
+	// instead. TODO(negz): Can obj be anything other than a Stack or a
+	// CustomResourceDefinition here?
+	meta.AddLabels(obj, map[string]string{
+		labelInstallNamespace: i.GetNamespace(),
+		labelInstallName:      i.GetName(),
+		labelInstallUID:       string(i.GetUID()),
 	})
 
 	// TODO(displague) pass/inject a controller specific logger
@@ -250,8 +267,7 @@ func (jc *stackInstallJobCompleter) createJobOutputObject(ctx context.Context, o
 		"name", obj.GetName(),
 		"namespace", obj.GetNamespace(),
 		"apiVersion", obj.GetAPIVersion(),
-		"kind", obj.GetKind(),
-		"ownerRefs", obj.GetOwnerReferences())
+		"kind", obj.GetKind())
 	if err := jc.client.Create(ctx, obj); err != nil && !kerrors.IsAlreadyExists(err) {
 		return errors.Wrapf(err, "failed to create object %s from job output %s", obj.GetName(), job.Name)
 	}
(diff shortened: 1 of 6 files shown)


pkg/controller/stacks/install/stackinstall.go
```diff
@@ -237,6 +250,31 @@ func (h *stackInstallHandler) update(ctx context.Context) (reconcile.Result, err
...
+		labelInstallName:      h.ext.GetName(),
+		labelInstallUID:       string(h.ext.GetUID()),
+	}
+	list := &apiextensionsv1beta1.CustomResourceDefinitionList{}
+	if err := h.kube.List(ctx, list, client.MatchingLabels(labels)); err != nil {
+		return fail(ctx, h.kube, h.ext, err)
+	}
+
+	for i := range list.Items {
+		if err := h.kube.Delete(ctx, &list.Items[i]); err != nil {
```
turkenh: we should ignore "err not found" in case it was already deleted.

pkg/controller/stacks/install/installjob.go
```diff
@@ -40,6 +40,13 @@ import (
 	"github.com/crossplaneio/crossplane/pkg/stacks"
 )
 
+// Labels used to track ownership across namespaces and scopes.
+const (
+	labelInstallNamespace = "stacks.crossplane.io/install-namespace"
+	labelInstallName      = "stacks.crossplane.io/install-name"
```
turkenh: These looks quite similar what is trying to be introduced here: https://github.com/crossplaneio/crossplane/pull/961/files#diff-4daf64a45270834c13f7f119f2d6aa98R82-R84
Would it make sense to use same labels?

pkg/controller/stacks/install/installjob.go
```diff
@@ -40,6 +40,13 @@ import (
...
+	labelInstallUID       = "stacks.crossplane.io/install-uid"
```
turkenh: Just wondering if we really need UIDs? I know owner references uses them but isn't name and namespace enough?

pkg/controller/stacks/install/installjob.go
```diff
@@ -236,11 +243,21 @@ func (jc *stackInstallJobCompleter) createJobOutputObject(ctx context.Context, o
...
 	}
 
-	// set an owner reference on the object
-	obj.SetOwnerReferences([]metav1.OwnerReference{
-		meta.AsOwner(meta.ReferenceTo(i, i.GroupVersionKind())),
+	// We want to clean up any installed CRDS when we're deleted. We can't rely
+	// on garbage collection because a namespaced object (StackInstall) can't
+	// own a cluster scoped object (CustomResourceDefinition), so we use labels
+	// instead. TODO(negz): Can obj be anything other than a Stack or a
+	// CustomResourceDefinition here?
```
turkenh: RBACs also created for stacks, where a clusterrole could be owned by a namespaced stack instance which is the same issue. But, AFAIU, it is in stack controller not here.

negz:
> Is this preferable to moving forward with #1023?

I don't know.

I'm not planning to do further work on this, because I don't feel I have a good enough understanding of the current state and goals for stacks to fix it in a reasonable way. I'll leave the PR open in draft for now, since it sparked some discussion.

pkg/controller/stacks/install/installjob.go
negz: Unsure - seems likely though.

pkg/controller/stacks/install/installjob.go
negz: Namespace and name are only unique at a particular point in time; UID is necessary to identify the owning object over time. In this case the UID is an extra layer of protection ensuring that the delete logic only cleans up CRDs that were really installed by this `StackInstall`, not a previous `StackInstall` with the same namespace and name that didn't clean up after itself.

It also allows us to use the same label set for `StackInstall` and `ClusterStackInstall`, given one of each could exist with the same namespace and name at the same time. We could also address this by adding a type metadata label.

hasheddan:
@jbw976 @lukeweber @displague wondering if anyone has thoughts here, Would be good to fix this in one way or another soon.

pkg/controller/stacks/install/installjob.go
```diff
@@ -236,11 +243,21 @@ func (jc *stackInstallJobCompleter) createJobOutputObject(ctx context.Context, o
...
+		})
 	}
 
-	// set an owner reference on the object
-	obj.SetOwnerReferences([]metav1.OwnerReference{
-		meta.AsOwner(meta.ReferenceTo(i, i.GroupVersionKind())),
+	// We want to clean up any installed CRDS when we're deleted. We can't rely
+	// on garbage collection because a namespaced object (StackInstall) can't
+	// own a cluster scoped object (CustomResourceDefinition), so we use labels
+	// instead. TODO(negz): Can obj be anything other than a Stack or a
```
jbw976: > TODO(negz): Can obj be anything other than a Stack or a CustomResourceDefinition here?

Right now, I don't think any other types will be flowing through here.  but in the future that is definitely a likely possibility.

pkg/controller/stacks/install/installjob.go
```diff
@@ -250,8 +267,7 @@ func (jc *stackInstallJobCompleter) createJobOutputObject(ctx context.Context, o
 		"name", obj.GetName(),
 		"namespace", obj.GetNamespace(),
 		"apiVersion", obj.GetAPIVersion(),
-		"kind", obj.GetKind(),
-		"ownerRefs", obj.GetOwnerReferences())
```
jbw976: want to log the labels now, instead of the owner refs? this is debug level, so adding lots of info is good :)

negz:
> Were you already able to run this scenario manually to see if it worked?

@jbw976 I have not. Would it make sense for someone from the stacks squad to take ownership of this issue?

jbw976:
Sure @negz, I'd be happy to drive this PR to completion.  I'll take it for a test spin and fix any potential issues I find.  Thanks for getting this so far already! 🙇

pkg/controller/stacks/install/installjob.go
displague: I like the very specific labels that call out that the 'stackinstall' parent is the named resource, but I think that the abstract labels provided in #961 work just as well and can be used for general tracing processes.

#961 does not include a label for the UID, so I wonder if that should be a requirement.

jbw976: that all makes a lot of sense to me.  it would be nice for this effort to go ahead and use the labels that we intend to use elsewhere for parent/child relationships in #961.  I'll continue driving this effort here to completion.

pkg/controller/stacks/install/installjob.go
displague: I've added this in 49a8ce58

pkg/controller/stacks/install/stackinstall.go
displague: Added in 66f7584a

pkg/controller/stacks/install/stackinstall.go
```diff
@@ -252,17 +252,18 @@ func (h *stackInstallHandler) update(ctx context.Context) (reconcile.Result, err
...
+		labelParentUID:       string(h.ext.GetUID()),
 	}
 	list := &apiextensionsv1beta1.CustomResourceDefinitionList{}
 	if err := h.kube.List(ctx, list, client.MatchingLabels(labels)); err != nil {
 		return fail(ctx, h.kube, h.ext, err)
 	}
 
 	for i := range list.Items {
-		if err := h.kube.Delete(ctx, &list.Items[i]); err != nil {
+		if err := h.kube.Delete(ctx, &list.Items[i]); err != nil && !kerrors.IsNotFound(err) {
```
negz: You could also use `if err := resource.IgnoreNotFound(h.kube.Delete(ctx, &list.Items[i])); err != nil` here, though neither is significantly more succinct or readable.

muvaf: I go for `if err := h.kube.Delete(ctx, &list.Items[i]); resource.IgnoreNotFound(err) != nil` for readability.

jbw976:
I have pushed a series of commits that take this PR to the finish line in terms of implementation and testing.  PTAL @displague and/or @negz. 

PR description has been updated https://github.com/crossplaneio/crossplane/pull/1091#issue-343861577

jbw976:
@rathpc please be advised of the updates in this PR to your design doc on parent child relationship labels.  The big change there is that group, version, and kind are all tracked by these labels now to be completely unambiguous about the parent of an object.  The latest version of the design doc can be seen at https://github.com/negz/crossplane/blob/refrace/design/one-pager-stack-relationship-labels.md

jbw976:
SonarCloud has politely reminded me that the `delete` scenario is new code that is uncovered by unit tests, I will add tests to cover that too. Thank you SonarCloud 💖

jbw976:
@displague @negz SonarCloud is happy again, this PR is ready for another look.

pkg/controller/stacks/install/stackinstall.go
```diff
@@ -237,6 +251,38 @@ func (h *stackInstallHandler) update(ctx context.Context) (reconcile.Result, err
...
+		return fail(ctx, h.kube, h.ext, err)
+	}
+
+	for i := range list.Items {
+		if err := h.kube.Delete(ctx, &list.Items[i]); runtimeresource.IgnoreNotFound(err) != nil {
+			return fail(ctx, h.kube, h.ext, err)
+		}
+	}
+
+	meta.RemoveFinalizer(h.ext, installFinalizer)
```
displague: Do we have to, or should we filter, on `installFinalizer` before performing all of these delete specific actions? 

I'm not familiar enough with the delete and finalizer flow.  While I don't doubt that this code would work, do we end up doing double lookup/delete work if this `delete` method is called multiple times, perhaps because multiple finalizers are in place?

jbw976: good question.  filtering on finalizer isn't strictly necessary because the main purpose of the finalizer is to block deletion of the object from the API server until this controller has finished its deletion work and clear the finalizer.  If the object has a deletion timestamp, it's fair game for running the deletion handling logic.

I think we used to filter by finalizer earlier in this project's history, but the contemporary patterns do not, and it's not strictly necessary.

All deletion operations should be idempotent and can be called multiple times.

jbw976:
Thanks for building and testing this out @displague and for sharing your results! 💪 

That's a good observation and point about the ClusterRoles ownerRef to the namespaced Stack records.  Can you please take a note to address that in #1023, or another relevant PR that you are working on for 0.6?

I'm OK with keeping that known deficiency for now in v0.5.1 patch release and shipping with what has been tested and vetted in this PR.  One of the biggest reasons in my mind for being OK with that is that the severity/impact is limited to just the stack controller itself.  If it were to have its ClusterRole garbage collected, it could still be manually fixed and the controller could run again with no data loss. 

That is a much less severe scope (it still isn't great though) than this particular issue being solved in this PR where the CRDs get garbage collected and all instances of them are deleted as well, resulting in true data loss.

Let's merge this PR and backport it to the release-0.5 branch then run a v0.5.1 patch release! 🙇

PR #1101, opened 2019-11-27
design: consuming k8s clusters

hasheddan:

Design portion of https://github.com/crossplaneio/crossplane/issues/859 and related discussions. Most notable feature is the implementation of a new custom resource `WorkloadTarget`, which encompasses the functionality of "publishing" a k8s cluster for consumption in a namespace.

[skip ci]

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/design/one-pager-consuming-k8s-clusters.md
+++ b/design/one-pager-consuming-k8s-clusters.md
@@ -266,7 +266,7 @@ publish *a* cluster (i.e. create `KubernetesTarget` resources) being able to
 publish *any* cluster would be to automatically create a `KubernetesTarget` as a
 side effect for creating a `KubernetesCluster` claim. This could be implemented
 as an "opt-in feature" by adding a `publishTarget` field on the
-`KubernetesCluster claim. In scenario 3 [above](#user-workflows), this would
+`KubernetesCluster` claim. In scenario 3 [above](#user-workflows), this would
 mean that application owners would only need permissions to create
 `KubernetesCluster` claims, and thus only be able to consume their dynamically
 provisioned clusters because a `KubernetesTarget` is created by the controller
--- a/design/one-pager-consuming-k8s-clusters.md
+++ b/design/one-pager-consuming-k8s-clusters.md
@@ -5,9 +5,6 @@
 
 ## Terminology
 
-* Kubernetes `Provider`: a cluster-scoped custom resource that points to any
-  Kubernetes `Secret` that contains connection information for a Kubernetes
-  cluster.
 * Kubernetes cluster **managed resource**: a cluster-scoped custom resource that
   represents the existence and configuration of a specific cloud provider's
   managed Kubernetes service (e.g. GKE, AKS, EKS).
@@ -63,29 +60,33 @@ referred to as a "Bring Your Own Cluster" scenario.
 
 ## Proposal
 
-This proposal can be separated into two main categories:
+This proposal can be separated into three main categories:
 1. The implementation of a new `KubernetesTarget` custom resource and
-   corresponding controller.
+   corresponding controllers.
 1. The modification of the scheduling behavior for `KubernetesApplication`.
+1. The implementation of a controller that automatically creates a
+   `KubernetesTarget` resource when a `KubernetesClaim` is bound to a Kubernetes
+   cluster managed resource. 
 
 ### KubernetesTarget
 
 The purpose of the `KubernetesTarget` resource is to effectively "publish"
 Kubernetes clusters for usage in a namespace. It is namespace-scoped and can
-only reference Kubernetes cluster managed resources or the Kubernetes `Provider`
-type, both of which are cluster-scoped. The creation of a `KubernetesTarget`
-resource in a namespace will be watched by a single new secret propagation
-controller, which will get the `Secret` referenced by either the Kubernetes
-cluster managed resource or Kubernetes `Provider` and propagate it to the
+only reference a cluster-scoped Kubernetes cluster managed resource or a
+namespace-scoped `Secret` that lives in the same namespace as the
+`KubernetesTarget`. The creation of a `KubernetesTarget` resource that
+references a Kubernetes cluster managed resource will be watched by controllers
+in stacks that provide the referenced resource type (i.e. a specific controller
+must be implemented in each stack that wishes to provide a Kubernetes cluster
+managed resource that `KubernetesApplication` resources can be scheduled to).
+These controllers will check to see if the `KubernetesTarget` references their
+cluster type and, if so, will propagate the cluster's connection `Secret` to the
 namespace of the `KubernetesTarget`. A local object reference to the `Secret`
-will be set on the `KubernetesTarget` object. The reasoning for referencing a
-`Provider` or Kubernetes cluster managed resource instead of a `Secret` object
-directly is to eliminate any cross-namespace references and to be explicit about
-the purpose of a `KubernetesTarget`. `Secret` objects are used for a variety of
-purposes in Kubernetes, and we are only interested in `Secret` objects that
-include Kubernetes cluster connection information in this case. An example of a
-`KubernetesTarget` in namespace `my-app-team` that is referencing a GKE cluster
-managed resource could look as follows:
+will then be set on the `KubernetesTarget` object. If a `KubernetesTarget` is
+created with a reference to a local `Secret`, it will be ignored by these
+controllers as the required `Secret` must already be present in the namespace.
+An example of a `KubernetesTarget` in namespace `my-app-team` that is
+referencing a GKE cluster managed resource could look as follows:
 
 ```yaml
 apiVersion: workload.crossplane.io/v1alpha1
@@ -95,11 +96,16 @@ metadata:
   name: enable-dev-k8s
   labels:
     env: dev
-clusterRef:
-  kind: GKECluster
-  apiVersion: compute.gcp.crossplane.io/v1alpha3
-  name: dev-k8s-cluster
+spec:
+  clusterRef:
+    kind: GKECluster
+    apiVersion: compute.gcp.crossplane.io/v1alpha3
+    name: dev-k8s-cluster
 ```
+
+Upon creation, the connection `Secret` associated with `dev-k8s-cluster` will be
+propagated to the `my-app-team` namespace and a reference to it will be set in
+the `connectionSecretRef` field of the `KubernetesTarget`.
 
 Note that `kind` and `apiVersion` are required for a `KubernetesTarget` that
 utilizes the `clusterRef` field because Crossplane cannot be knowledgeable of
@@ -107,7 +113,7 @@ every Kubernetes cluster managed resource type that may have been installed into
 the cluster.
 
 An example of a `KubernetesTarget` in namespace `my-app-team` that is
-referencing a Kubernetes `Provider` could look as follows:
+referencing a local `Secret` directly could look as follows:
 
 ```yaml
 apiVersion: workload.crossplane.io/v1alpha1
@@ -117,15 +123,17 @@ metadata:
   name: enable-dev-k8s
   labels:
     env: dev
-providerRef:
-  name: onprem-dev-k8s-cluster
+spec:
+  connectionSecretRef:
+    name: my-k8s-connection
 ```
 
-Because `KubernetesTarget` resources can point to the Kubernetes `Provider`
-resource type, any existing Kubernetes cluster that was not provisioned using
-Crossplane (including on-prem clusters) can now be targeted for provisioning
-workloads. The name is the only information needed for `providerRef` due to the
-fact that Crossplane is knowledgeable of the `Provider` kind.
+Because `KubernetesTarget` resources can point directly to a local `Secret`, any
+existing Kubernetes cluster that was not provisioned using Crossplane (including
+on-prem clusters) can now be targeted for provisioning workloads. To enable this
+functionality, a user would manually create a `Secret` containing the
+`kubeconfig` information for their cluster in the `my-app-team` namespace, then
+create a `KubernetesTarget` that references it.
 
 ### KubernetesApplication Scheduling
 
@@ -133,19 +141,47 @@ As mentioned above, `KubernetesApplication` resources are currently scheduled to
 `KubernetesCluster` claims in their namespace. With the introduction of the
 `KubernetesTarget`, `KubernetesApplication` resources can still be scheduled via
 label selectors, but will now be scheduled to a `KubernetesTarget` in their
-namespace, then use its `secretRef` to get the `Secret` that was propagated to
-the namespace from the Kubernetes cluster managed resource or Kubernetes
-`Provider` object reference. Importantly, for a `KubernetesApplication` to be
-able to be scheduled to a Kubernetes cluster, a `KubernetesTarget` that points
-to that cluster managed resource or `Provider` *must* be present in the
-namespace of the `KubernetesApplication`.
+namespace, then use its `connectionSecretRef` to get the `Secret` that was
+propagated to the namespace from the Kubernetes cluster managed resource or
+already existed due to manual creation. Importantly, for a
+`KubernetesApplication` to be able to be scheduled to a Kubernetes cluster, a
+`KubernetesTarget` that points to that cluster managed resource or local
+`Secret` *must* be present in the namespace of the `KubernetesApplication`.
 
 This change isolates the actions of *provisioning* (`KubernetesCluster` or
 cloud-specific managed resource), *publishing* (`KubernetesTarget`), and
 *consuming* (`KubernetesApplication`) Kubernetes clusters, actions which may be
 executed by [almost disjoint] sets of users. Having these actions represented by
 separate custom resources allows for more granular [RBAC] permissions around
-Kubernetes cluster usage in Crossplane. 
+Kubernetes cluster usage in Crossplane.
+
+### Automatic KubernetesTarget Creation Controller
+
+Because the ability to *publish* a Kubernetes cluster provides immense power in
+a Crossplane environment, it is desirable for Kubernetes cluster managed
+resources to be able to be consumed in the namespace that they are claimed
+without a user requiring the ability to create a `KubernetesTarget`. For
+instance, if an infrastructure owner wants to allow for an application team to
+provision some clusters dynamically by creating a `KubernetesCluster` claim in
+their namespace that references a `GKEClusterClass`, that application team
+should be able to then schedule `KubernetesApplication` resources to that
+cluster without requiring the permissions to create a `KubernetesTarget` (which
+would allow them to enable to consumption of any Kubernetes cluster managed
+resource) or asking the infrastructure owner to cerate a `KubernetesTarget` for
+them (which would break the self-service model).
+
+To enable this functionality, a single controller should be implemented in core
+Crossplane that watches for the creation of `KubernetesCluster` claim resources
+and automatically creates a `KubernetesTarget` with the `connectionSecretRef`
+set to the connection `Secret` that was propagated to the namespace.
+Importantly, this controller should also clean up the `KubernetesTarget` when
+the `KubernetesCluster` (and corresponding) `Secret` are deleted. This can be
+accomplished by setting an `ownerReference` to the `KubernetesCluster` claim on
+the `KubernetesTarget`. In addition, because this `KubernetesTarget` resource
+may be scheduled to by a `KubernetesApplication` using labels, any labels that
+are present on the `KubernetesCluster` claim should be propagated to the
+`KubernetesTarget` resource and the name of the resource should match that of
+the claim.
 
 ### User Workflows
 
@@ -161,8 +197,9 @@ a variety of use-cases.
 1. An application owner asks for an environment to deploy their application.
 1. Infrastructure owner creates a new namespace and a `KubernetesTarget` that
    points to the `dev` cluster.
-1. Application owner provisions a `KubernetesApplication` with labels that
-   select the `KubernetesTarget` referencing the `dev` cluster for provisioning.
+1. Application owner provisions a `KubernetesApplication` in that namespace with
+   labels that select the `KubernetesTarget` referencing the `dev` cluster for
+   provisioning.
 1. When ready for testing, the application owner requests access to the `stage`
    cluster, which the infrastructure owner then "publishes" by creating another
    `KubernetesTarget` in the namespace that points to the `stage` cluster.
@@ -182,9 +219,10 @@ owner and application owner.
 1. Infrastructure owner creates a new namespace.
 1. Application owner creates a `KubernetesCluster` claim in the namespace that
    references the `dev` `GKEClusterClass` and provisions a new Kubernetes
-   cluster.
-1. Application owner creates a `KubernetesTarget` to point at the newly
-   provisioned `GKECluster`.
+   cluster. This causes the creation of a `KubernetesTarget` in the namespace
+   when the connection `KubernetesCluster` claim becomes `Bound`.
+1. Application owner provisions a `KubernetesApplication` in that namespace with
+   labels that select the automatically created `KubernetesTarget`.
 1. When ready for testing, the application owner requests access to the `stage`
    cluster, which the infrastructure owner then "publishes" by creating a
    `KubernetesTarget` in the namespace that points to the `stage` cluster.
@@ -205,43 +243,33 @@ infrastructure owner (i.e. "clusters as cattle").
    Application owners are given permissions to create clusters using these
    classes as they see fit.
 1. There is one particular set of workloads that is required to be run in an
-   existing on-premises cluster (for some important reason). The infrastructure
-   owner creates a Kubernetes `Secret` with connection information to that
-   cluster in an infrastructure namespace (secrets are namespace-scoped). They
-   then create a cluster-scoped Kubernetes `Provider` that references the
-   `Secret`.
+   existing on-premises cluster (for some important reason).
 1. Application owner says that they need to run a sensitive workload on that
    cluster. Infrastructure owner creates a new namespace called
-   `super-secret-workloads` and creates a `KubernetesTarget` in that namespace
-   that references the previously created Kubernetes `Provider`.
+   `super-secret-workloads` and creates a `Secret` with the on-prem cluster's
+   `kubeconfig` information. They also create a `KubernetesTarget` in that
+   namespace that references the `Secret`.
 1. Application owner provisions a `KubernetesApplication` with labels that
    select that `KubernetesTarget` for provisioning.
 
-This example shows the power of this model, but also one of the notable
-shortcomings: if a user can publish clusters to a namespace with a
-`KubernetesTarget`, then they can consume any cluster. For the application
-owners that are doing self-service provisioning in this scenario, they will
-likely need to be able to create `KubernetesTarget` resources for their
-dynamically provisioned Kubernetes cluster managed resources. However, if they
-can do so, they could just as easily create a `KubernetesTarget` for the
-Kubernetes `Provider` that is only to be used for sensitive workloads. While it
-would be nice to restrict the clusters that could be published by users with the
-ability to create `KubernetesTarget` resources, there is not an immediately
-clear and elegant way to do so. It could be argued that only trusted users
-should have this permission level anyway, so it may be less of an issue than
-expressed here. Nevertheless, a potential solution is outlined
-[below](#future-considerations).
+This example shows the power of this model, as users are now able to consume
+clusters that were provisioned with Crossplane as well as those that may be
+pre-existing or cannot be managed by Crossplane.
 
 ## Technical Implementation
 
 The implementation of this proposal is relatively lightweight and requires
 minimal churn on any existing exposed APIs. The first step would be implementing
-the `KubernetesTarget` type in [core Crossplane]. In addition to the creation of
-the custom resource, a new controller would also be required to watch
-`KubernetesTarget` object creation and propagate `Secret` objects to its
-namespace as described above. This controller should use [predicates] to only
-reconcile `KubernetesTarget` objects that have *one* of either a `clusterRef`
-and `providerRef`. 
+the `KubernetesTarget` type in [core Crossplane]. As mentioned above, in
+addition to the creation of the custom resource, a new controller would also be
+required in each stack to watch `KubernetesTarget` object creation and propagate
+`Secret` objects to its namespace if its `clusterRef` references a Kubernetes
+cluster managed resource type supplied by that stack. These controllers should
+use [predicates] to only reconcile `KubernetesTarget` objects that reference
+their resource type and do not already have a `connectionSecretRef` set, and
+should refuse to propagate a `Secret` if the `KubernetesTarget` references a
+managed resource that is not `Bound`. The majority of this logic can be shared
+across controllers by adding a new reconciler to [crossplane-runtime].
 
 The next step would be to modify the `KubernetesApplication` [scheduling
 controller] to list `KubernetesTarget` resources in the namespace that match
@@ -254,33 +282,29 @@ referencing a `KubernetesTarget` resource with a reference to a local `Secret`
 with connection information. Because a `KubernetesApplication` is now being
 scheduled to something other than a `KubernetesCluster` claim (and may be
 scheduled to other resource types in the future), updating the `clusterRef`
-field to be named `targetRef` would be more clear. This is the only API change
-proposed by this design and will result in the increment of the
-`KubernetesApplication` `apiVersion` from `v1alpha1` to `v1alpha2`.
+field to be named `targetRef` and `clusterSelector` to `targetSelector` would be
+more clear. This is the only API change proposed by this design and will result
+in the increment of the `KubernetesApplication` `apiVersion` from `v1alpha1` to
+`v1alpha2`.
+
+Lastly, a single controller must be added to core Crossplane that watches for
+the creation of `KubernetesCluster` claims and automatically creates a
+`KubernetesTarget` resource in the namespace when the `KubernetesCluster` claim
+becomes `Bound`. As mentioned above, the created `KubernetesTarget` should have
+an `ownerReference` to the `KubernetesCluster` claim to ensure it is cleaned up
+by [garbage collection] when the claim is deleted.
 
 ## Future Considerations
 
-One feature that would eliminate a step for self-service dynamic cluster
-provisioning and also fix the issue described above of users with permissions to
-publish *a* cluster (i.e. create `KubernetesTarget` resources) being able to
-publish *any* cluster would be to automatically create a `KubernetesTarget` as a
-side effect for creating a `KubernetesCluster` claim. This could be implemented
-as an "opt-in feature" by adding a `publishTarget` field on the
-`KubernetesCluster` claim. In scenario 3 [above](#user-workflows), this would
-mean that application owners would only need permissions to create
-`KubernetesCluster` claims, and thus only be able to consume their dynamically
-provisioned clusters because a `KubernetesTarget` is created by the controller
-automatically. For clusters created by the infrastructure owner, the claims
-could just be created in a infrastructure-specific namespace as to not
-accidentally publish a cluster to an application team namespace. Importantly,
-all Kubernetes cluster managed resources created by an infrastructure owner in
-this scenario, whether provisioned statically or dynamically, should be claimed
-immediately such that application owners with permissions to create
-`KubernetesCluster` claims are not able to use a cluster that they are not
-intended to. Implementation of this functionality would increase the scope of
-this design to include implementing a new claim controller exclusively for
-`KubernetesCluster` claims, so it has been deferred to future implementation for
-the time being.
+Because the scheduling of `KubernetesApplication` resources is now isolated to
+target the `KubernetesTarget` resource, more intelligent scheduling can be
+enabled without touching other parts of the Crossplane ecosystem. Previously, a
+`KubernetesCluster` claim was used for claiming, consuming, and dynamically
+provisioning Kubernetes cluster resources so changes to the the API type related
+to scheduling (i.e. consuming) could unintentionally affect those other
+capabilities as well. Potential future scheduling improvements could involve
+price, latency, and geographic optimization by surfacing additional fields or
+labels on the `KubernetesTarget` type.
 
 <!-- Named Links -->
 [almost disjoint]: https://en.wikipedia.org/wiki/Almost_disjoint_sets
@@ -289,3 +313,5 @@ the time being.
 [scheduling controller]: https://github.com/crossplaneio/crossplane/blob/2cab9826a1d5088a87b5d24693336d228279a3a7/pkg/controller/workload/kubernetes/scheduler/scheduler.go#L57
 [remote cluster connection information]: https://github.com/crossplaneio/crossplane/blob/2cab9826a1d5088a87b5d24693336d228279a3a7/pkg/controller/workload/kubernetes/resource/resource.go#L401
 [predicates]: https://godoc.org/sigs.k8s.io/controller-runtime/pkg/predicate
+[crossplane-runtime]: https://github.com/crossplaneio/crossplane-runtime
+[garbage collection]: https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/
--- a/design/one-pager-consuming-k8s-clusters.md
+++ b/design/one-pager-consuming-k8s-clusters.md
@@ -1,0 +1,270 @@
+# Consuming Kubernetes Clusters
+* Owner: Dan Mangum (@hasheddan)
+* Reviewers: Crossplane Maintainers
+* Status: Draft
+
+## Terminology
+
+* Kubernetes `Provider`: a cluster-scoped custom resource that points to any
+  Kubernetes `Secret` that contains connection information for a Kubernetes
+  cluster.
+* Kubernetes cluster **managed resource**: a cluster-scoped custom resource that
+  represents the existence and configuration of a specific cloud provider's
+  managed Kubernetes service (e.g. GKE, AKS, EKS).
+* `KubernetesCluster` claim: a namespace-scoped custom resource that can be used
+  to claim a Kubernetes cluster managed resource, or dynamically provision one.
+* **Workload**: a collection of Kubernetes resources that are intended to be
+  deployed into a target Kubernetes cluster. A workload is represented by a
+  `KubernetesApplication` custom resource.
+
+## Background
+
+Currently, Crossplane models portable cloud provider managed services as
+*managed resources* that can be either statically or dynamically provisioned
+using resource *classes* and *claims*. One such resource type is a Kubernetes
+cluster, which uses a `KubernetesCluster` claim and cloud provider-specific
+managed resource types (e.g. `GKECluster`, `AKSCluster`, `EKSCluster`, etc.). It
+is appropriate for Kubernetes clusters to be modeled in the same manner as other
+managed services due to the fact that they are presented by the provider in the
+same fashion. However, in Crossplane, Kubernetes clusters are unique to other
+managed service offerings in that they can also be used as a target for running
+workloads.
+
+Workloads are provisioned using the `KubernetesApplication` type. It is
+namespace-scoped and is scheduled to a target cluster using labels to match a
+`KubernetesCluster` claim in its namespace. Because managed resources may only
+be bound to a single claim, the claim is namespaced, and `KubernetesApplication`
+resources can only be scheduled to `KubernetesCluster` claims in their
+namespace, any Kubernetes managed resource offering may only be consumed (i.e.
+deployed to) from a single namespace.
+
+In order to fully conceptualize the usage of Kubernetes clusters in Crossplane,
+it is helpful to separate the notions of a **Kubernetes cluster as a managed
+service** and a **Kubernetes cluster as a target for deploying workloads**. With
+this distinction made, we can assert that this proposal in no way modifies the
+managed service notion of a Kubernetes cluster, but instead formalizes this
+concept of a Kubernetes cluster as a deployment target.
+
+## Goals
+
+Because it is unlikely that Crossplane users will always want a 1-to-1
+namespace-to-cluster consumption model, it is desirable to support deploying
+workloads in any namespace to any Kubernetes cluster. However, it is also
+desirable to continue to treat managed Kubernetes services as a managed resource
+that is supplied by cloud providers. For this reason, any solution should be
+additive to the existing model.
+
+As a secondary goal for this design, it should be possible to use existing
+Kubernetes clusters (wherever they may be running) that were not provisioned by
+Crossplane as targets for deploying workloads. This concept has frequently been
+referred to as a "Bring Your Own Cluster" scenario.
+
+## Proposal
+
+This proposal can be separated into two main categories:
+1. The implementation of a new `WorkloadTarget` custom resource.
+1. The modification of the scheduling behavior for `KubernetesApplication`.
+
+### WorkloadTarget
+
+The purpose of the `WorkloadTarget` resource is to effectively "publish"
+Kubernetes clusters for usage in a namespace. It is namespace-scoped and can
+only reference Kubernetes cluster managed resources or the Kubernetes `Provider`
+type, both of which are cluster-scoped. An example of a `WorkloadTarget` in
+namespace `my-app-team` that is referencing a GKE cluster managed resource could
+look as follows:
+
+```yaml
+apiVersion: workload.crossplane.io/v1alpha1
+kind: WorkloadTarget
+metadata:
+  namespace: my-app-team
+  name: enable-dev-k8s
+  labels:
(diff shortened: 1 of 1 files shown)


design/one-pager-consuming-k8s-clusters.md
```diff
@@ -0,0 +1,270 @@
...
+service** and a **Kubernetes cluster as a target for deploying workloads**. With
+this distinction made, we can assert that this proposal in no way modifies the
+managed service notion of a Kubernetes cluster, but instead formalizes this
+concept of a Kubernetes cluster as a deployment target.
+
+## Goals
+
+Because it is unlikely that Crossplane users will always want a 1-to-1
+namespace-to-cluster consumption model, it is desirable to support deploying
+workloads in any namespace to any Kubernetes cluster. However, it is also
```
negz: I know what you're getting at here because we've discussed it out of band, but I don't know if it would be obvious to uninitiated readers. I wonder if we could expand on the use case(s) a little more to make this clearer. I'm going to try do that in this comment:

We're assuming Crossplane is deployed as a standalone** "control cluster" tasked with:

* Managing cloud infrastructure such as databases, buckets, and caches.
* Managing Kubernetes clusters - a type of cloud infrastructure.
* Managing the applications deployed to said Kubernetes clusters.

Each Crossplane control cluster is divided up into namespaces. _Possibly_ each namespace belongs to some application team; for example the payments team will create all the payments related resource claims and workloads in the `payments` namespace. We expect that the infrastructure operator configuring Crossplane for these teams to use may want to:

* Require all teams to share one Kubernetes cluster that is managed by the infrastructure operator.
* Require all teams to share a small set of Kubernetes clusters that are managed by the infrastructure operator.
* Allow each team to provision one or more Kubernetes clusters to deploy their workloads to.

Currently we support only the last strategy, but we think the first two strategies are likely to be more popular.

** All of this is possible when Crossplane is deployed into an existing Kubernetes cluster too, but it seems like an uncommon use case to install Crossplane onto an existing Kubernetes cluster, then use Crossplane to create _other_ Kubernetes clusters, and to deploy workloads to them.

design/one-pager-consuming-k8s-clusters.md
```diff
@@ -0,0 +1,270 @@
...
+
+As a secondary goal for this design, it should be possible to use existing
+Kubernetes clusters (wherever they may be running) that were not provisioned by
+Crossplane as targets for deploying workloads. This concept has frequently been
+referred to as a "Bring Your Own Cluster" scenario.
+
+## Proposal
+
+This proposal can be separated into two main categories:
+1. The implementation of a new `WorkloadTarget` custom resource.
```
negz: Given that `KubernetesApplication` is just one kind of workload, I think `KubernetesApplicationTarget` would fit better here, though I don't love either name. `KubernetesCluster` feels like the best name for this, but it's (kind of) taken by the `KubernetesCluster` resource claim. We could have `kubernetescluster.workload.crossplane.io` as the scheduling target
and `kubernetescluster.compute.crossplane.io` as the claim, but that would probably be pretty confusing.

design/one-pager-consuming-k8s-clusters.md
```diff
@@ -0,0 +1,270 @@
...
+```yaml
+apiVersion: workload.crossplane.io/v1alpha1
+kind: WorkloadTarget
+metadata:
+  namespace: my-app-team
+  name: enable-dev-k8s
+  labels:
+    env: dev
+clusterRef:
+  kind: Provider
```
negz: Given that a `Provider` is not shaped like a managed resource, would it make sense to have distinct and mutually exclusive `providerRef` and `clusterRef` fields? I could go either way on this.

negz: It might also be worth explaining why, given we only really need a `Secret` to tell us how to connect to a cluster, we target `Provider`, `GKECluster` etc rather than targeting the secret directly. i.e. Because it feels cleaner (to me) for a namespaced object (`WorkloadTarget`) to reference a cluster scoped object (`GKECluster`) which in turn references a namespaced object in another namespace (`Secret`) than it does to allow `WorkloadTarget` to directly reference across namespaces to a `Secret`.

design/one-pager-consuming-k8s-clusters.md
negz: Aha, I see you went into this in detail further down into the doc. I'll leave my comment for posterity, but I don't think we need to take any action.

design/one-pager-consuming-k8s-clusters.md
```diff
@@ -0,0 +1,270 @@
...
+step would be implementing the `WorkloadTarget` type in [core Crossplane]. It
+would not require any additional controllers.
+
+The next step would be to modify the `KubernetesApplication` [scheduling
+controller] to list `WorkloadTarget` resources in the namespace that match
+labels, instead of the current implementation that lists `KubernetesCluster`
+claims. Currently, this controller sets a reference to a `KubernetesCluster`
+claim, which is then later propagated to the child
+`KubernetesApplicationResource` objects, which use it to get a `Secret` with
+[remote cluster connection information].
```
negz: It may make sense to rename the `KubernetesApplication` `clusterRef` field if we're scheduling to something other than `KubernetesCluster`. Maybe `targetRef`? I don't think renaming that field would be too much of an issue given it's relatively rarely explicitly set in current examples, and is a `v1alpha1` API.

design/one-pager-consuming-k8s-clusters.md
```diff
@@ -0,0 +1,270 @@
...
+connection `Secret`, it would likely be cleanest to go ahead and get that
+`Secret` and set a reference to it on the `KubernetesApplication`, then
+propagate to the `KubernetesApplicationResource` objects. The alternative option
+would be to set a reference to either the managed resource or the `Provider`,
+then follow the references to the corresponding `Secret` in the remote cluster
+connection step in the `KubernetesApplicationResource` object controller. While
+setting the reference to the `Secret` is potentially a cross-namespace
+reference, it is one that only exists in the status of the resource, and a valid
+`WorkloadTarget` would have to exist in the `KubernetesApplication` namespace
+for that cross-namespace reference to be set by the controller.
```
negz: I see your point that folks down in a namespace can see that the referenced secret exists in some other namespace by inspecting the `Provider` (or `GKECluster` etc) `.spec.writeConnectionSecretToRef`, but it still feels like a violation of namespace boundaries to allow one namespaced object to directly reference an object in another namespace, to me.

design/one-pager-consuming-k8s-clusters.md
```diff
@@ -0,0 +1,270 @@
...
+
+## Future Considerations
+
+One feature that would eliminate a step for self-service dynamic cluster
+provisioning and also fix the issue described above of users with permissions to
+publish *a* cluster (i.e. create `WorkloadTarget` resources) being able to
+publish *any* cluster would be to automatically create a `WorkloadTarget` as a
```
negz: I like this approach - perhaps the `KubernetesCluster` claim would have a `publishTarget` field to opt in, or we'd just do it inherently.

design/one-pager-consuming-k8s-clusters.md
zoharm: Based on my current understanding, a draft of naming that makes sense in my mind:

`KubernetesClusterClaim` - For claiming a (workload) target cluster (currently `KubernetesCluster`) assuming here that kubernetes clusters can only be claimed to run workloads on, which is portable between cloud managed and custom/on-prem clusters.
`KubernetesCluster` - A workload target kubernetes cluster refering to managed or on-prem cluster (`WorkloadTarget` in this doc, since it only targets kubernetes)
`ManagedKubernetesCluster` - What is currently `KubernetesCluster` (which i believe only handles managed clusters, which I assume includes things such as scaling and upgrading their nodes, etc...)

I believe that a custom/on-prem workload target cluster should either be a separately managed `KubernetesCluster` (as per above naming,) or something like a `CustomKubernetesCluster` (where both `CustomKubernetesCluster` and `ManagedKubernetesCluster` reference or own the published `KubernetesCluster` workload target)

Will add some more notes about my thoughts on `Provider(s)` in a later comment.

Please let me know if I'm getting anything mixed up or any other clarficiations for my understanding would be appreciated!

negz: > `KubernetesClusterClaim` 

I definitely think it would make sense for our claim kinds to end with the word `Claim`, as our `Class` kinds end with the word `Class`. Though there is an interesting twist in Crossplane in that a `FooClaim` claims a `Bar`, not a `Foo`. 🤔

negz: > ManagedKubernetesCluster - What is currently KubernetesClusterResource? (which i believe only handles managed clusters, which I assume includes things such as scaling and upgrading their nodes, etc...)

No such kind exists today; are you thinking of our managed resources? They're "strongly typed" to their underlying cloud provider concept, so they're `kind: GKECluster`, `kind: EKSCluster` etc.

negz:
> I think that Provider(s) should do the "Managing" of a managed kubernetes cluster (and other service offerings.) So, whether its a public cloud, or some custom on-prem API, that would make sense for me to be a Provider.

https://github.com/crossplaneio/crossplane/blob/09bc80/design/one-pager-k8s-native-providers.md

@zoharm the above design document provides a little context on the "Kubernetes provider". In short, a Crossplane provider represents an external API account that can be used to "provide" (provision) managed resources. i.e. A GCP Provider is basically a GCP service account that can be used to access a GCP project, while an AWS Provider is an AWS IAM user that can be used to access an AWS account.

The Kubernetes provider is a bit of an odd one, in that it's not used to _provide Kubernetes clusters_, but rather _uses Kubernetes to provide other managed resources_. An example of this using the Kubernetes provider to provision a `CockroachDBCluster` managed resource.

design/one-pager-consuming-k8s-clusters.md
zoharm: > > ManagedKubernetesCluster - What is currently KubernetesClusterResource? (which i believe only handles managed clusters, which I assume includes things such as scaling and upgrading their nodes, etc...)
> 
> No such kind exists today; are you thinking of our managed resources? They're "strongly typed" to their underlying cloud provider concept, so they're `kind: GKECluster`, `kind: EKSCluster` etc.

Ok I see. Then I can just look at this `ManagedKubernetesCluster` as a theoretical abstraction of these cloud provider kinds. So if the cloud provider kinds are explicitly decoupled (and don't really have) a high level abstraction ("Managed" kind) by design, these names do make perfect sense to me.

zoharm:
> The Kubernetes provider is a bit of an odd one, in that it's not used to _provide Kubernetes clusters_, but rather _uses Kubernetes to provide other managed resources_. An example of this using the Kubernetes provider to provision a `CockroachDBCluster` managed resource.

@negz Ok, thank you for this clarification! So, a statically provisioned (custom/on-prem) cluster could be a Kubernetes-Native-Provider provisioning a `CockroachDBCluster` managed resource as mentioned.

However, is the Kubernetes-Native-Provider the mechanism by which a cluster is consumed by a workload (such as an app's kubernetes resources, like a deployment and service)? It seems to me this path is different than going through a Provider (instead something among the lines of KubernetesApplication > KubernetesCluster(Resource) - where KubernetesApplication is not something that actually consumes a Kubernetes-native-provider?), please correct me if I'm wrong. Thank you!

design/one-pager-consuming-k8s-clusters.md
```diff
@@ -0,0 +1,270 @@
...
+propagate to the `KubernetesApplicationResource` objects. The alternative option
+would be to set a reference to either the managed resource or the `Provider`,
+then follow the references to the corresponding `Secret` in the remote cluster
+connection step in the `KubernetesApplicationResource` object controller. While
+setting the reference to the `Secret` is potentially a cross-namespace
+reference, it is one that only exists in the status of the resource, and a valid
+`WorkloadTarget` would have to exist in the `KubernetesApplication` namespace
+for that cross-namespace reference to be set by the controller.
+
+## Future Considerations
```
muvaf: I think having a namespace-pointing `WorkloadTarget` could be a future consideration. As infra-owner, when I give access to two different teams, I trust them to not mess with each other's resources. But if I can limit their access to 1 namespace in the target cluster, it'd be much safer for all parties.

muvaf:
> It seems to me this path is different than going through a Provider (instead something among the lines of KubernetesApplication > KubernetesCluster(Resource) - where KubernetesApplication is not something that actually consumes a Kubernetes-native-provider?), please correct me if I'm wrong. 

@zoharm you're right. I also think that the usage of `Provider` resource has changed in Crossplane and naming could be improved https://github.com/crossplaneio/crossplane/pull/1101#discussion_r351647247

design/one-pager-consuming-k8s-clusters.md
```diff
@@ -0,0 +1,270 @@
...
+  Kubernetes `Secret` that contains connection information for a Kubernetes
+  cluster.
+* Kubernetes cluster **managed resource**: a cluster-scoped custom resource that
+  represents the existence and configuration of a specific cloud provider's
+  managed Kubernetes service (e.g. GKE, AKS, EKS).
+* `KubernetesCluster` claim: a namespace-scoped custom resource that can be used
+  to claim a Kubernetes cluster managed resource, or dynamically provision one.
+* **Workload**: a collection of Kubernetes resources that are intended to be
+  deployed into a target Kubernetes cluster. A workload is represented by a
+  `KubernetesApplication` custom resource.
```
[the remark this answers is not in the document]
hasheddan: Thanks for clarification! I'll add here the a workload is something that can be scheduled to a k8s cluster, and currently the only resource that exists in this category is `KubernetesApplication`.

design/one-pager-consuming-k8s-clusters.md
hasheddan: I would back either `KubernetesConnection` or `KubernetesTarget`. I think this makes sense as it seems like it may be true that we could schedule something other than a `KubernetesApplication` to a `KubernetesTarget/Connection`.

design/one-pager-consuming-k8s-clusters.md
hasheddan: @muvaf I really really like that idea! I was feeling the most uncomfortable with this part of the design and this solution is much better :)

design/one-pager-consuming-k8s-clusters.md
hasheddan: @negz that sounds good to me! @muvaf I definitely can understand that desired experience, but one of the major values of the design here is that we are isolating the *publishing* of k8s clusters with a `WorkloadTarget`. If a cluster could be referenced directly, then any cluster in the Crossplane cluster could be used. To improve this user experience when using self-service cluster provisioning, I have proposed a solution in the `Future Considerations` section.

design/one-pager-consuming-k8s-clusters.md
hasheddan: It sounds like maybe this should be moved from a future consideration to being part of this design, what do you think?

design/one-pager-consuming-k8s-clusters.md
```diff
@@ -0,0 +1,270 @@
...
+kind: WorkloadTarget
+metadata:
+  namespace: my-app-team
+  name: enable-dev-k8s
+  labels:
+    env: dev
+clusterRef:
+  kind: GKECluster
+  apiVersion: compute.gcp.crossplane.io/v1alpha3
+  name: dev-k8s-cluster
```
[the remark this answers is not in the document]
hasheddan: Hmm, I am not sure that this would ever happen. Names of Kubernetes objects are immutable so the cluster object would have to be deleted and then recreated, in which case I think a new `WorkloadTarget` (naming pending update haha) should have to be published again.

design/one-pager-consuming-k8s-clusters.md
```diff
@@ -0,0 +1,270 @@
...
+
+```yaml
+apiVersion: workload.crossplane.io/v1alpha1
+kind: WorkloadTarget
+metadata:
+  namespace: my-app-team
+  name: enable-dev-k8s
+  labels:
+    env: dev
+clusterRef:
```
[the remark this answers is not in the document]
hasheddan: That's an interesting thought! I am not sure how much this would be utilized as the purpose of the `WorkloadTarget` is to make a *specific* cluster available to be scheduled to in an application namespace. Generally, I would think that an infrastructure owner would be creating a `WorkloadTarget`, knowing exactly what cluster they wanted to "publish". If an application owner was creating a `WorkloadTarget` I think they would likely of just dynamically provisioned a cluster and know exactly what they want to target (or the `WorkloadTarget` was created for them as a side-effect of the `KubernetesCluster` claim; see `Future Considerations`).

So it seems like the most useful case for using a `clusterSelector` would be if an infrastructure team just spins up a bunch of "dev" k8s clusters and want to make any one of those available in an application namespace.

design/one-pager-consuming-k8s-clusters.md
```diff
@@ -0,0 +1,270 @@
...
+  kind: Provider
+  apiVersion: kubernetes.crossplane.io/v1alpha1
+  name: onprem-dev-k8s-cluster
+```
+
+Because `WorkloadTarget` resources can point to the Kubernetes `Provider`
```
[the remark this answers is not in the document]
hasheddan: I would definitely be opening to renaming the Kubernetes `Provider`, although it does behave like a true `Provider` with things like `stack-rook`.

design/one-pager-consuming-k8s-clusters.md
```diff
@@ -0,0 +1,270 @@
...
+Crossplane (including on-prem clusters) can now be targeted for provisioning
+workloads.
+
+### KubernetesApplication Scheduling
+
+As mentioned above, `KubernetesApplication` resources are currently scheduled to
+`KubernetesCluster` claims in their namespace. With the introduction of the
+`WorkloadTarget`, `KubernetesApplication` resources can still be scheduled via
+label selectors, but will now be scheduled to a `WorkloadTarget` in their
+namespace, then use its `clusterRef` to get the `Secret` referenced by the
```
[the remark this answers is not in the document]
hasheddan: Responded to your similar comment above, but I really like this pattern :)

design/one-pager-consuming-k8s-clusters.md
```diff
@@ -0,0 +1,270 @@
...
+
+#### Scenario 3: Infrastructure owner enables self-service, except for one on-prem cluster with sensitive data
+
+1. Infrastructure owner creates a variety of Kubernetes cluster classes with
+   configuration for different environments and different cloud providers.
+   Application owners are given permissions to create clusters using these
+   classes as they see fit.
+1. There is one particular set of workloads that is required to be run in an
+   existing on-premises cluster (for some important reason). The infrastructure
+   owner creates a Kubernetes `Secret` with connection information to that
```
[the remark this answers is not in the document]
hasheddan: I think this could be greatly helped by the crossplane-cli as well :)

design/one-pager-consuming-k8s-clusters.md
hasheddan: That is an interesting idea for sure! I think it might be cool to even create a `Role` / `RoleBinding` / `ServiceAccount` to establish RBAC permissions for deploying as well. I definitely think that would be a future consideration because I am not certain how much it would actually be used in practice in a true "cluster as cattle" scenario where the cluster is the unit of isolation.

design/one-pager-consuming-k8s-clusters.md
hasheddan: I like the idea of having both a `providerRef` and `clusterRef` field because it means that a user only needs to provide the `name` of the resource if it is a `Provider` type.

design/one-pager-consuming-k8s-clusters.md
```diff
@@ -0,0 +1,291 @@
...
+## Technical Implementation
+
+The implementation of this proposal is relatively lightweight and requires
+minimal churn on any existing exposed APIs. The first step would be implementing
+the `KubernetesTarget` type in [core Crossplane]. In addition to the creation of
+the custom resource, a new controller would also be required to watch
+`KubernetesTarget` object creation and propagate `Secret` objects to its
+namespace as described above. This controller should use [predicates] to only
+reconcile `KubernetesTarget` objects that have *one* of either a `clusterRef`
+and `providerRef`.
```
hasheddan: I think that we should enforce one of either `clusterRef` or `providerRef` due to the fact that having one just take precedence over the other could lead to confusion about where a particular workload is being scheduled.

design/one-pager-consuming-k8s-clusters.md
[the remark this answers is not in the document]
muvaf: But we can't guarantee that the old `WorkloadTarget` will be deleted, too. Someone can delete `Provider`/`KubernetesCluster` and after a day or two, create another one with the same name. Old `WorkloadTarget` will suddenly be working again with a different cluster whose intent may or may not be to be a target for that specific `WorkloadTarget`.

design/one-pager-consuming-k8s-clusters.md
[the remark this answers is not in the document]
muvaf: I meant all `Provider` types, not only Kubernetes one. I think what `Credentials` covers is a super-set of what `Provider` does and IMO we need that stretch.

design/one-pager-consuming-k8s-clusters.md
muvaf: > If a cluster could be referenced directly, then any cluster in the Crossplane cluster could be used.

Not any but the ones in your app namespace. I don't know, as a user, having to create another resource while I have `KubernetesCluster` in my namespace feels odd to me.

Even though we do the side-effect `WorkloadTarget` creation, I'd still want to be able to refer to what I created directly. Thinking of as a stack developer or helm chart author of my app, this would lower the burden of having to realize those relations in the provisioning logic.

design/one-pager-consuming-k8s-clusters.md
```diff
@@ -62,21 +64,32 @@ referred to as a "Bring Your Own Cluster" scenario.
...
+The purpose of the `KubernetesTarget` resource is to effectively "publish"
 Kubernetes clusters for usage in a namespace. It is namespace-scoped and can
 only reference Kubernetes cluster managed resources or the Kubernetes `Provider`
-type, both of which are cluster-scoped. An example of a `WorkloadTarget` in
-namespace `my-app-team` that is referencing a GKE cluster managed resource could
-look as follows:
+type, both of which are cluster-scoped. The creation of a `KubernetesTarget`
+resource in a namespace will be watched by a single new secret propagation
+controller, which will get the `Secret` referenced by either the Kubernetes
+cluster managed resource or Kubernetes `Provider` and propagate it to the
```
muvaf: Is there a way to make this reference exclusive to `Kubernetes Provider` type? Like if a user who is able to create a `KubernetesTarget` shouldn't be able to pull in credentials of a `AWS Provider` or other cloud provider secrets.

design/one-pager-consuming-k8s-clusters.md
```diff
@@ -193,78 +212,80 @@ infrastructure owner (i.e. "clusters as cattle").
...
-they could just as easily create a `WorkloadTarget` for the Kubernetes
-`Provider` that is only to be used for sensitive workloads. While it would be
-nice to restrict the clusters that could be published by users with the ability
-to create `WorkloadTarget` resources, there is not an immediately clear and
-elegant way to do so. It could be argued that only trusted users should have
-this permission level anyway, so it may be less of an issue than expressed here.
-Nevertheless, a potential solution is outlined [below](#future-considerations).
+`KubernetesTarget`, then they can consume any cluster. For the application
+owners that are doing self-service provisioning in this scenario, they will
+likely need to be able to create `KubernetesTarget` resources for their
```
muvaf: I guess if we keep the ability to provision `KubernetesApplication` to `KubernetesCluster`, those self-serving users wouldn't need to have the role that is needed to create `KubernetesTarget`.

design/one-pager-consuming-k8s-clusters.md
muvaf: I remember a similar discussion in the cross-resource referencing topic. It'd have been very helpful if OpenAPI v3 schema validation had _only one of these can be filled with a value_ type of validation.

design/one-pager-consuming-k8s-clusters.md
hasheddan: @muvaf yep! That way they could consume clusters that they provisioned themselves, but would not be able to enable the consumption of any arbitrary cluster.

design/one-pager-consuming-k8s-clusters.md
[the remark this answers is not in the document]
muvaf: > So it seems like the most useful case for using a clusterSelector would be if an infrastructure team just spins up a bunch of "dev" k8s clusters and want to make any one of those available in an application namespace.

My concern was more about having to manually enter the `kind` and `apiVersion` but that also seems like a valid case though I am not sure about whether we should worry about that in the first implementation. If we go with `uid` per https://github.com/crossplaneio/crossplane/pull/1101/files#r352281520 , this would be even more non-user friendly. But this could be pretty subjective, so, I'm not pushing too hard for this.

design/one-pager-consuming-k8s-clusters.md
negz: I like that `KubernetesConnection` has symmetry with our existing "connection secret" concept.

design/one-pager-consuming-k8s-clusters.md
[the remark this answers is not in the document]
negz: I've had similar thoughts around this. I believe for us `Provider` means "a thing that can be used to provide infrastructure". This is subtly different from "a cloud provider", but I think the name "provider" will cause most people to think "this represents GCP", when in fact it more specifically represents "a service account that can be used to provision GCP infrastructure".

I think we should at least raise an issue floating renaming the `Provider` concept to something like `Credentials`, `ProviderCredentials`, or `ProviderAccount`.

design/one-pager-consuming-k8s-clusters.md
```diff
@@ -0,0 +1,291 @@
...
+should have this permission level anyway, so it may be less of an issue than
+expressed here. Nevertheless, a potential solution is outlined
+[below](#future-considerations).
+
+## Technical Implementation
+
+The implementation of this proposal is relatively lightweight and requires
+minimal churn on any existing exposed APIs. The first step would be implementing
+the `KubernetesTarget` type in [core Crossplane]. In addition to the creation of
+the custom resource, a new controller would also be required to watch
```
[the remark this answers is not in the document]
hasheddan: I was thinking that it would be a single controller and because the various Kubernetes cluster managed resources support the `Managed` interface which includes [`ConnectionSecretWriterTo`](https://github.com/crossplaneio/crossplane-runtime/blob/bd43e78aa7637acd75c371104bac2cc4778f6e30/pkg/resource/interfaces.go#L74) then the controller could work similar to how the managed resource reconciler interacts with managed resources.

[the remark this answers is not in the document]
negz: It's not clear to me how that would work.

If there was a single controller it would not be able to know that the resource kind it was dealing with was (for example) `*v1alpha3.GKECluster` and thus satisfied the `ConnectionSecretWriterTo` interface, unless the controller imported every possible type of Kubernetes cluster struct from their respective stacks. I believe the controller would instead have to treat the type as `*unstructured.Unstructured`, which does not satisfy the `ConnectionSecretWriterTo` interface.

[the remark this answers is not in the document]
hasheddan: Ah yes this is a good point. It seems like the most straightforward solution would be to implement a lightweight shared reconciler in crossplane-runtime that could be implemented with k8s cluster managed resources in each stack.

design/one-pager-consuming-k8s-clusters.md
```diff
@@ -0,0 +1,291 @@
...
+### KubernetesTarget
+
+The purpose of the `KubernetesTarget` resource is to effectively "publish"
+Kubernetes clusters for usage in a namespace. It is namespace-scoped and can
+only reference Kubernetes cluster managed resources or the Kubernetes `Provider`
+type, both of which are cluster-scoped. The creation of a `KubernetesTarget`
+resource in a namespace will be watched by a single new secret propagation
+controller, which will get the `Secret` referenced by either the Kubernetes
+cluster managed resource or Kubernetes `Provider` and propagate it to the
+namespace of the `KubernetesTarget`. A local object reference to the `Secret`
```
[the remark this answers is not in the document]
hasheddan: I see your point here. However, couldn't a similar situation occur even if the cluster was bound? Is it just that the act of claiming the resource feels stronger than just having a `KubernetesTarget` resource that point to it?

design/one-pager-consuming-k8s-clusters.md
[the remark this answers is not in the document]
muvaf: I think it'd work since `kind` and `apiVersion` is known in `clusterRef`, the controller that watches `WorkloadTarget` could create a `runtime.Object` via https://github.com/crossplaneio/crossplane-runtime/blob/ca4b6b46c9a43f0bf46990ea05f1b23480f8a94f/pkg/resource/resource.go#L75 with that GVK information and then cast that object into `Managed`, extract the secret ref.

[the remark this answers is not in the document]
hasheddan: @muvaf I think you are correct. I was thinking that this should be possible, but was too quick to abandon my original design when confronted with the possibility that it may not be 😅 thanks for pointing this out!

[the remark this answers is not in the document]
muvaf: Ah, sorry to disappoint 😞  but I realized @negz is right. `MustCreateObject`'s `runtime.ObjectCreater` arguments needs to be aware of the specific struct, which is done usually via stack's `AddToScheme` calls.

I think a controller in `crossplane` repo that uses `*unstructured.Unstructured` makes sense as @negz suggested. We know that `writeConnectionSecretToRef` exists for all managed k8s cluster resources though it exists in all resources that implements `Managed`. So, maybe after getting the secret reference we can validate it's indeed a k8s cluster by looking up a specific key, probably `kubeconfig` and only then propagate the secret.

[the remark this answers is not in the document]
hasheddan: Ah yes, the reason we are able to use `MustCreateObject()` in stacks is due to the fact that we add all the Crossplane types to the scheme. I would push back on the `*unstructured.Unstructured` controller in favor of a lightweight controller for each managed k8s type in stacks. My main reasoning for this is that it feels more like a stack creator intentionally specifying that `KubernetesApplication` resources can be scheduled to a certain resource, which feels in-line with saying that a given resource can be dynamically provisioned by a given claim type.

design/one-pager-consuming-k8s-clusters.md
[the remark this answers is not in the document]
negz: My understanding is that the thinking (at the time I commented) was that `KubernetesTarget` would cause a secret to be propagated from the (e.g.) `GKECluster` to its namespace. This secret would need to be propagated for a `KubernetesTarget` to actually be used to deploy a `KubernetesApplication`. So if the secret was only propagated if and when the `GKECluster` was bound, then no I don't think a similar situation could occur.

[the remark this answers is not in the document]
hasheddan: So in the scenario above, let’s say that App team B claims the managed resource before any KubernetsTarget is created. It isn’t now bound. However, this provides no assurance to App team B that another team (i.e. App team A) will not also start to utilize the cluster via a KubernetesConnection, which can now be used with the cluster because it is bound. It does provide assurance that when App team B first claimed this resource that no other team was currently using it. Is this what you are communicating here? I may still not be fully grasping but your intent.

negz:
> I still see great value in isolating that "publishing" step with a different API object.

I agree. Consider what we might want from a scheduling target**:

* A way to declare that a Kubernetes cluster exists as a scheduling target.
* Details on how to connect to the Kubernetes cluster (the secret).
* Metadata that we could use to schedule on (currently labels, eventually other things like price, latency, and geography).

While I see it solves some problem, I don't love the UX of scheduling a `KubernetesApplication` across both `KubernetesCluster` and `KubernetesTarget`. I imagine `KubernetesApplication` would need both `clusterRef` and `targetRef` under this proposal, and maybe even `clusterSelector` and `targetSelector`, with the cluster and target fields being mutually exclusive?

How about:

* A `KubernetesApplication` can only be scheduled to a `KubernetesTarget` in its namespace.
* A `KubernetesApplication` is scheduled to a `KubernetesTarget` by setting its `targetRef`. Its optional `targetSelector` limits which `KubernetesTarget` the `KubernetesApplication` may be scheduled to.
* A `KubernetesTarget` contains a `connectionSecretRef` and a `clusterRef`. The former references the connection secret that will be used to connect to the cluster. The latter points to a GKE, EKS, etc cluster whose connection secret will be propagated to the namespace of the `KubernetesTarget`.

For example:

```yaml
apiVersion: crossplane.io/v1alpha1
kind: KubernetesTarget
metadata:
  namespace: default
  name: coolcluster
spec:
   # If this reference is set _and_ connectionSecretRef is unset its secret
   # is propagated to the namespace of the KubernetesTarget, whose
   # connectionSecretRef is updated to point to the propagated Secret.
   # The propagation only happens if and when the cluster is bound to a
   # claim.
   clusterRef:
     apiVersion: compute.gcp.crossplane.io/v1alpha3
     kind: GKECluster
     name: coolcluster
  # If this reference is set the secret is assumed to exist and clusterRef is
  # ignored.
  connectionSecretRef:
    coolsecret
```

The UX for BYO is then, for each namespace in which you want to use that cluster.

1. Create a connection secret for it.
1. Point a `KubernetesTarget` at it.

The UX for clusters the app team claims as their own is then:

1. Create a `KubernetesCluster` claim.
1. Point a `KubernetesTarget` at the claim's connection secret.

The UX for infra operators wishing to share a cluster with many app teams is:

1. Create a `KubernetesCluster` claim in some system namespace, like `crossplane-system`.
1. For each app team that you want to use that cluster, create a `KubernetesTarget` in their namespace and set its `clusterRef` to the managed resource it is bound to.

The big danger zone with this proposal is that anyone who can create a `KubernetesTarget` can allow themselves to schedule to _and read the connection secret of_ any bound Kubernetes cluster. I think we can simply state that RBAC will prevent app owners from creating their own `KubernetesTarget` resources for the first pass of this design. I like the idea of adding a controller that automatically creates a `KubernetesTarget` for every `KubernetesCluster` claim as a follow on.

** Given that I keep calling this a "scheduling target", and that part of its job is eventually going to be describing any properties of the Kubernetes cluster it represents that we might want to schedule on, I'm now leaning more toward the `KubernetesTarget` name than `KubernetesConnection`. :( The name `KubernetesEndpoint` (like the `Endpoint` resources that back a `Service`) also came to mind while writing this.

design/one-pager-consuming-k8s-clusters.md
```diff
@@ -0,0 +1,317 @@
...
+and automatically creates a `KubernetesTarget` with the `connectionSecretRef`
+set to the connection `Secret` that was propagated to the namespace.
+Importantly, this controller should also clean up the `KubernetesTarget` when
+the `KubernetesCluster` (and corresponding) `Secret` are deleted. This can be
+accomplished by setting an `ownerReference` to the `KubernetesCluster` claim on
+the `KubernetesTarget`. In addition, because this `KubernetesTarget` resource
+may be scheduled to by a `KubernetesApplication` using labels, any labels that
+are present on the `KubernetesCluster` claim should be propagated to the
+`KubernetesTarget` resource and the name of the resource should match that of
+the claim.
```
hasheddan: How do we feel about this label propagation and naming strategy for automatically created `KubernetesTarget` resources?

design/one-pager-consuming-k8s-clusters.md
```diff
@@ -0,0 +1,317 @@
...
+would allow them to enable to consumption of any Kubernetes cluster managed
+resource) or asking the infrastructure owner to cerate a `KubernetesTarget` for
+them (which would break the self-service model).
+
+To enable this functionality, a single controller should be implemented in core
+Crossplane that watches for the creation of `KubernetesCluster` claim resources
+and automatically creates a `KubernetesTarget` with the `connectionSecretRef`
+set to the connection `Secret` that was propagated to the namespace.
+Importantly, this controller should also clean up the `KubernetesTarget` when
+the `KubernetesCluster` (and corresponding) `Secret` are deleted. This can be
```
hasheddan: How do we feel about the this automatic clean-up? I am accomplishing this by setting an `ownerReference` (see technical implementation section). My thought process here is that we want the life-cycle of automatically created `KubernetesTarget` resources strictly tied to that of the `KubernetesCluster` claim (i.e. if Crossplane creates something it should also clean it up). Currently, I believe all of our managed k8s cluster resource controllers use the [default secret propagation](https://github.com/crossplaneio/crossplane-runtime/blob/0f37bea5496e8dd4a47dd731a83bc42b97a43878/pkg/resource/publisher.go#L62) mechanism, so the `Secret` that the `KubernetesTarget` referenced would be cleaned up with the `KubernetesCluster` claim, rendering the `KubernetesTarget` unusable anyway.

design/one-pager-consuming-k8s-clusters.md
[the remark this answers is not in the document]
hasheddan: The technical implementation section now includes the requirement that a managed resource referenced by a `KubernetesTarget` must be `Bound` for the connection `Secret` to be propagated :)

design/one-pager-consuming-k8s-clusters.md
negz: I was thinking the same thing - this seems reasonable to me.

design/one-pager-consuming-k8s-clusters.md
negz: This seems reasonable to me.

It's possible someone could delete the `KubernetesCluster` with `--cascade=false` and thus the `KubernetesTarget` becomes orphaned, but I don't think that's a huge issue given that its technically the person doing the deleting asked for, even if it doesn't make sense.

design/one-pager-consuming-k8s-clusters.md
```diff
@@ -117,35 +123,65 @@ metadata:
...
+a Crossplane environment, it is desirable for Kubernetes cluster managed
+resources to be able to be consumed in the namespace that they are claimed
+without a user requiring the ability to create a `KubernetesTarget`. For
+instance, if an infrastructure owner wants to allow for an application team to
+provision some clusters dynamically by creating a `KubernetesCluster` claim in
+their namespace that references a `GKEClusterClass`, that application team
+should be able to then schedule `KubernetesApplication` resources to that
+cluster without requiring the permissions to create a `KubernetesTarget` (which
+would allow them to enable to consumption of any Kubernetes cluster managed
+resource) or asking the infrastructure owner to cerate a `KubernetesTarget` for
```
negz: Looks like a typo in "create".

hasheddan:
@muvaf would you mind approving this? I believe your requested changes will block merging

design/one-pager-consuming-k8s-clusters.md
```diff
@@ -0,0 +1,321 @@
...
+provide Kubernetes cluster managed resources (i.e. a specific controller must be
+implemented in each stack that wishes to provide a Kubernetes cluster managed
+resource that `KubernetesApplication` resources can be scheduled to). These
+controllers will check to see if the `KubernetesTarget` references their cluster
+type n its `clusterRef` and, if so, will propagate the cluster's connection
+`Secret` to the namespace of the `KubernetesTarget`. A local object reference to
+the `Secret` will then be set on the `KubernetesTarget` object in the
+`connectionSecretRef` field. If a `KubernetesTarget` is created with a reference
+to a local `Secret`, it will be ignored by these controllers as the required
+`Secret` must already be present in the namespace. An example of a
```
muvaf: I think it should keep refreshing that secret as credentials may change, for example EKS clusters. Maybe a logic like `if connectionSecretRef empty, fill with a name. If it's filled go ahead and populate that secret` could work instead of checking the reference's existence.

design/one-pager-consuming-k8s-clusters.md
```diff
@@ -0,0 +1,321 @@
...
+spec:
+  connectionSecretRef:
+    name: my-k8s-connection
+```
+
+Because `KubernetesTarget` resources can point directly to a local `Secret`, any
+existing Kubernetes cluster that was not provisioned using Crossplane (including
+on-prem clusters) can now be targeted for provisioning workloads. To enable this
+functionality, a user would manually create a `Secret` containing the
+`kubeconfig` information for their cluster in the `my-app-team` namespace, then
```
muvaf: Maybe saying `kubeconfig` **key** in the secret could be more descriptive.

design/one-pager-consuming-k8s-clusters.md
```diff
@@ -0,0 +1,321 @@
...
+The implementation of this proposal is relatively lightweight and requires
+minimal churn on any existing exposed APIs. The first step would be implementing
+the `KubernetesTarget` type in [core Crossplane]. As mentioned above, in
+addition to the creation of the custom resource, a new controller would also be
+required in each stack to watch `KubernetesTarget` object creation and propagate
+`Secret` objects to its namespace if its `clusterRef` references a Kubernetes
+cluster managed resource type supplied by that stack. These controllers should
+use [predicates] to only reconcile `KubernetesTarget` objects that reference
+their resource type and do not already have a `connectionSecretRef` set, and
+should refuse to propagate a `Secret` if the `KubernetesTarget` references a
```
muvaf: Check my comment above about continuously propagating.

design/one-pager-consuming-k8s-clusters.md
hasheddan: This is a good point, and as you mentioned with EKS we will actually *have* to do this. This gets a little bit complicated as the current proposal says that we don't do anything if `connectionSecretRef` is populated, but what we actually want to know is if the `connectionSecretRef` is populated by a manually created `Secret`. I think the easiest fix for this problem would be:

1. Controllers reconcile any `KubernetesTarget` that references their managed resource type in the `clusterRef`.
1. If the `connectionSecretRef` is unset then the `Secret` is propagated to the namespace and the reference is set.
1. If the `connectionSecretRef` is set, then it is updated to match the managed resource's connection `Secret`.

The primary areas of concern here to me are:
1. If you create a `KubernetesTarget` that with a `connectionSecretRef` to a manually created "BYO Cluster" `Secret` *and* a `clusterRef`, that manually created `Secret` will be overwritten by the `clusterRef` connection `Secret`. This seems fine to me. Instead of saying that we defer to the `connectionSecretRef` when it is set, we are saying that if a `clusterRef` is set then we will always use it.
1. Because these controllers are separate from the managed resource controllers (where continuous secret propagation currently happens), there is a longer lag time from when connection information changes to the `Secret` being updated. I also, don't see this as a huge issue when framed as a max of a 2 minute lag (managed reconciler long wait + KubernetesTarget reconciler long wait) compared to a max of a 1 minute lag (managed reconciler long wait) in the current system where the managed reconciler continuously published connection secret updates.

@muvaf I think all of this is in-line with your thoughts, but let me know if I have deviated. Thanks for bringing this up :)

@negz could you indicate if this update sounds right to you?

hasheddan: @negz pointed out in stand-up today that the existing [secret reconciler](https://github.com/crossplaneio/crossplane-runtime/blob/master/pkg/resource/secret_reconciler.go) can be used to accomplish this. If I understand correctly, with EKS (where this reconciler is already [used](https://github.com/crossplaneio/stack-aws/blob/master/pkg/controller/compute/secret.go)), the controller could just be modified to also watch secrets controlled by a `KubernetesTarget`.

```go
func (c *EKSClusterSecretController) SetupWithManager(mgr ctrl.Manager) error {
	p := resource.NewPredicates(resource.AnyOf(
		resource.AllOf(resource.IsControlledByKind(v1alpha1.KubernetesClusterGroupVersionKind), resource.IsPropagated()),
		resource.AllOf(resource.IsControlledByKind(v1alpha1.KubernetesTargetGroupVersionKind), resource.IsPropagated()),
		resource.AllOf(resource.IsControlledByKind(v1alpha3.EKSClusterGroupVersionKind), resource.IsPropagator()),
	))

		Named(strings.ToLower(fmt.Sprintf("connectionsecret.%s.%s", v1alpha3.EKSClusterKind, v1alpha3.Group))).
		Watches(&source.Kind{Type: &corev1.Secret{}}, &resource.EnqueueRequestForPropagator{}).
		For(&corev1.Secret{}).
		WithEventFilter(p).
		Complete(resource.NewSecretPropagatingReconciler(mgr))
}
```
And we would only need to set the [required annotations](https://github.com/crossplaneio/crossplane-runtime/blob/bd43e78aa7637acd75c371104bac2cc4778f6e30/pkg/resource/secret_reconciler.go#L41) when the `Secret` is initially propagated. This would allow the `KubernetesTarget` controller to function as currently proposed in this design and refuse to operate on `KubernetesTarget` resources where the `connectionSecretRef` is already set.

negz: One thing we'd have to consider with the above controller is that a source secret can currently only be propagated to one destination, due to how the annotations work. Currently we only instantiate the secret propagation controller for EKS, but we set the propagation annotations on connection secrets for all managed resources at claim binding time. We'll need to update the annotations to support propagating a secret to multiple destinations.

Another thing to keep in mind is that the secret propagation reconciler will not, by design, propagate _to_ a secret that does consent to the process via a set of annotations. This means we won't be able to (and IMO should not try to) let `KubernetesTarget` specify an existing secret to propagate to.

So I believe the behaviour should be, if the user specifies:

* Only `connectionSecretRef` no propagation happens, we just try to use the referenced secret.
* Only `clusterRef` we'll create and start propagating to a secret.
* Both, and the referenced secret does not exist, we'll create it and start propagating to it.
* Both, and the referenced secret exists without the consenting propagation annotations we'll refuse to propagate to the existing secret.

design/one-pager-consuming-k8s-clusters.md
```diff
@@ -0,0 +1,321 @@
...
+1. The implementation of a new `KubernetesTarget` custom resource and
+   corresponding controllers.
+1. The modification of the scheduling behavior for `KubernetesApplication`.
+1. The implementation of a controller that automatically creates a
+   `KubernetesTarget` resource when a `KubernetesClaim` is bound to a Kubernetes
+   cluster managed resource. 
+
+### KubernetesTarget
+
+The purpose of the `KubernetesTarget` resource is to effectively "publish"
```
prasek: With namespaced claims now exclusively matching to cluster-scoped resource classes, the publishing of `KubernetesTarget` to a namespace seems more aligned with the previous portable resource classes approach where `KubernetesApplication` is matched to a namespaced `KubernetesTarget` much like claims were matched to namespaced portable resource classes.

Would it make sense for `KubernetesTarget` be cluster-scoped to be more consistent with the claims/classes model we have today? That would impact the ability for app teams to create their own clusters as cattle, but if app teams are creating their own Kubernetes clusters you could argue they should have their own Crossplane instance as well - esp. given they have joint access to all the cluster-scoped resource classes which references the provider secret.

design/one-pager-consuming-k8s-clusters.md
```diff
@@ -0,0 +1,321 @@
...
+
+1. Infrastructure owner statically provisions three `GKECluster` managed
+   resources: `dev`, `stage`, and `prod`
+1. An application owner asks for an environment to deploy their application.
+1. Infrastructure owner creates a new namespace and a `KubernetesTarget` that
+   points to the `dev` cluster.
+1. Application owner provisions a `KubernetesApplication` in that namespace with
+   labels that select the `KubernetesTarget` referencing the `dev` cluster for
+   provisioning.
+1. When ready for testing, the application owner requests access to the `stage`
```
prasek: The simplified UX of cluster-scoped classes push things towards a crossplane instance per environment, and while you can definitely model classes with environment labels (dev, stage, prod), there is no means of securing access via RBAC via labels for resource class usage AFAIK, so wonder if a cluster-scoped `KubernetesTarget` would be a simpler way to go.

design/one-pager-consuming-k8s-clusters.md
negz: If we made `KubernetesTarget` cluster scoped we would no longer support (or would at least make quite awkward) the concept of a namespace owner provisioning their own clusters to deploy their workloads to. I think that would be a shame - I'd prefer to leave that avenue open while also allowing infrastructure operators to supply a set of clusters for folks to use. We'd also be preventing infrastructure operators from having any control over who gets to use which cluster. If a target exists, it's fair game for any namespace to deploy applications to.

If we do decide `KubernetesTarget` should be cluster scoped, then `KubernetesTarget` should not  exist. A cluster scoped `KubernetesTarget` exists, and is named `Provider`.

negz: Some carry-on effects of making `KubernetesTarget` cluster scoped would be:

* Namespace owners cannot BYO cluster; only infrastructure owners can do this.
* There's close to no reason for the `KubernetesCluster` claim to exist anymore. There would be little reason for namespace owners to claim Kubernetes clusters, because they could not use them as deployment targets.

negz:
Thanks for persisting with this @hasheddan.

Issue #1095, opened 2019-11-25
One pager for Easy Stack controller architecture


negz:
TODO: Flesh this out.

We'd like to enable Crossplane infrastructure operators to install a stack that contained some best practice/sane default resource classes. So there would be an "easy AWS stack" you can use with AWS, etc. This issue tracks the _how_ the easy stacks would be installed.

Related Designs:
 * [UI one pager](/crossplaneio/crossplane/blob/master/design/one-pager-stack-ui-metadata.md)
 * [Stacks one pager](/crossplaneio/crossplane/blob/master/design/design-doc-stacks.md)

FAQ:
 * Q: does the UI spec support field validation, including regex with human readable validation error messages?
   * A: yes and yes

PR #1103, opened 2019-11-27
Easy stacks parameter tiers

muvaf:

This PR is a placeholder for discussions around _what_ configuration sets we want to have as easy stacks and **not intended to be merged**. Folder structure and namings are subject to change per one-pager coming for #1095 .

The folder `easystacks_temp` consists of configuration sets for 3 cloud providers. I imagine we'll likely have _small_, _medium_ and _large_ options for each of them, in total 9 different configuration set. ~~Right now, all tiers are copies of each other. Via reviews on this PR, we can arrive on the best suited parameters for each.~~ We start with the minimal configuration set for each provider.

Initial configuration values are taken from Wordpress example, a few things are updated like Azure k8s version from 1.12.8 to 1.14.8.

Likely, we'll have a generic controller as base per #1095 and each set will fork from that, change the values/files and publish as stack. It could be the case that stack CRD to expose a spec field like `tier` and user can use which tier they'd like to use without needing to install a different stack for different tiers. So, there would be 3 stacks, one for each provider but one stack might include more than 1 configuration set. All those implementation details to be discussed on #1095 

The files look like this right now: (for easy browsing)
```
├── aws
│   ├── environment
│   └── resources
├── azure
│   ├── environment
│   └── resources
└── gcp
    ├── environment
    └── resources
```

[skip ci]

### Checklist

I have:
  appropriate.

[definition of done].


suskin:
Great! :squid:

Only two thoughts so far:

* Did we intentionally use `resources` as the folder name in the AWS packs, but `services` in the other ones? It seems like they're fulfilling the same purpose, so I'm not sure why we have two different names instead of a single one.
* I would lean more toward creating a single "tier" for now, especially if we only know what one of the tiers would look like. We can add more in the future. I also suspect that tiers will be roughly aligned with use-cases. For example, the first tier that **we** provide will probably be intended to provide sensible defaults so that someone can play around with examples easily, so I'm not sure yet whether a name like "small", "medium", or "large" even makes sense for that example tier.

negz:
> I imagine we'll likely have small, medium and large options for each of them, in total 9 different configuration set.

Is size the only variant we expect to need? I'm wondering whether we might also want, for example, resource classes with different access strategies. I can't imagine everyone will want to create a whole new VPC just to deploy a database, for example. Do we need an alternative that allows them to deploy into an existing VPC, or allow traffic from a particular IP range rather than a private VPC?

Perhaps it would help to identify some common user stories for folks who would be installing an Easy Stack? Are we targeting folks coming to Crossplane with a fresh new cloud provider account and no opinions about how infrastructure works?

muvaf:
> I would lean more toward creating a single "tier" for now, especially if we only know what one of the tiers would look like. We can add more in the future.

Yeah, I guess you're right. I'll scope down this PR into 1 configuration set for each provider.

> Did we intentionally use resources as the folder name in the AWS packs, but services in the other ones? It seems like they're fulfilling the same purpose, so I'm not sure why we have two different names instead of a single one.

Not intentionally, `resources` is the one for now. Though both the folder structure and naming of the resources are subject to change. I wanted this PR only to discuss what the actual parameters should be. I believe it's better to discuss the folder structure in #1095 since it'll be somehow tied to how the controller will work.

> Are we targeting folks coming to Crossplane with a fresh new cloud provider account and no opinions about how infrastructure works?

I think this is the initial intent. I am updating this PR to create a minimal version of each resource class. I will try to make sure publishing a new configuration set is as easy as dumping your YAML files and build the stack.

suskin:
Nice. This is a bit of a nitpick, but I believe it still makes sense to have the resource pack name in the folder path, so that it's easier to add more of them later. Maybe something like changing `cluster/easystacks_temp/aws` to `cluster/easystacks_temp/quickstart-aws` would work well. I'm not sure whether `quickstart-aws` or `aws-quickstart` would be better; guess it depends on how people are searching for them, and whether they want multiple providers or not.

muvaf:
> Maybe something like changing cluster/easystacks_temp/aws to cluster/easystacks_temp/quickstart-aws would work well.

This PR is not intended to be merged though. It's just to show what configurations will go into each easy stack. So, the folder name in `cluster/easystacks_temp/<foldername>` doesn't really matter.

muvaf:
All easy stacks implementations are completed and this PR is updated with the validated configurations.

Here is the repositories:
https://github.com/muvaf/minimal-gcp
https://github.com/muvaf/minimal-aws
https://github.com/muvaf/minimal-azure

This is the generic reconciler used by all of them:
https://github.com/muvaf/crossplane-resourcepacks

When template stacks are ready, easy stacks will be ported to template stacks and their repositories will live under `crossplaneio` org.

@prasek @jbw976 actually I think we can move them right away since eventually they will be moved, template stacks porting work could be a PR instead of a new repo. What do you think?

jbw976:
> @prasek @jbw976 actually I think we can move them right away since eventually they will be moved, template stacks porting work could be a PR instead of a new repo. What do you think?

I would love to move them sooner than later, I am actually a fan of always using the crossplaneio org even for prototyping work.

What exactly needs to be moved right now to the crossplaneio org? The repos listed below? Can you open an issue with all the information so that we can follow up on that soon? Thanks @muvaf!

* https://github.com/muvaf/minimal-gcp
* https://github.com/muvaf/minimal-aws
* https://github.com/muvaf/minimal-azure
* https://github.com/muvaf/crossplane-resourcepacks

muvaf:
> I am actually a fan of always using the crossplaneio org even for prototyping work.

@jbw976 Honestly, I didn't know how to do it properly with the specific licensing stuff, upbound build and all and you were away, so I just didn't want to be blocked.

Here is the issue https://github.com/crossplaneio/crossplane/issues/1136

PR #1109, opened 2019-11-28
helm: improve quality of Helm chart metadata

displague:

Following https://github.com/crossplaneio/crossplane/pull/1077, this PR adds a NOTES.txt file, README.md, and provides more details about the
project in Chart.yaml

* https://helm.sh/docs/topics/charts/#the-chart-file-structure
* https://helm.sh/docs/topics/chart_template_guide/notes_files/
* https://helm.sh/docs/topics/charts/#the-chart-yaml-file

The README.md is a copy of docs/install-crossplane.md.  This is a
great starting point.  I didn't want to duplicate the file but I didn't
want to introduce a symlink.  The install-crossplane.md frontmatter may
not work well in chart rendering tools, so it was removed in this copy.

In the future we may want to tailor the messaging in the README.md with
more description of what Crossplane is.  The file should act as a
product introduction and helm installation/configuration guide.

All of these changes benefit chart trackers and hubs, such as https://hub.helm.sh/charts/crossplane/crossplane.  Compare this, in its current state, to that of another chart which provides more metadata: https://hub.helm.sh/charts/mattermost/mattermost-enterprise-edition

Helm lint comes up clean, but this is not a superb test.  Seeing NOTES.txt upon install, and seeing README.md rendered on remote sites (like hub.sh) can guide future changes.

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/cluster/charts/crossplane/README.md
+++ b/cluster/charts/crossplane/README.md
@@ -1,0 +1,246 @@
+# Install Crossplane
+
+Crossplane can be easily installed into any existing Kubernetes cluster using the regularly published Helm chart.
+The Helm chart contains all the custom resources and controllers needed to deploy and configure Crossplane.
+
+## Pre-requisites
+
+* [Kubernetes cluster](https://kubernetes.io/docs/setup/)
+  * For example [Minikube](https://kubernetes.io/docs/tasks/tools/install-minikube/), minimum version `v0.28+`
+* [Helm](https://docs.helm.sh/using_helm/), minimum version `v2.12.0+`.
+  * For Helm 2, make sure Tiller is initialized with sufficient permissions to work on `crossplane-system` namespace.
+
+## Installation
+
+Helm charts for Crossplane are currently published to the `alpha` and `master` channels.
+In the future, `beta` and `stable` will also be available.
+
+### Alpha
+
+The alpha channel is the most recent release of Crossplane that is considered ready for testing by the community.
+
+*Note: if installing Crossplane into a Kubernetes cluster that is running a version older than 1.15, Helm 2 installation is recommended. If using Helm 2, [Tiller](https://v2.helm.sh/docs/install/#installing-tiller) must be installed.* 
+
+Install with Helm 2:
+
+```console
+helm repo add crossplane-alpha https://charts.crossplane.io/alpha
+helm install --name crossplane --namespace crossplane-system crossplane-alpha/crossplane
+```
+
+Install with Helm 3:
+
+```console
+kubectl create namespace crossplane-system
+helm repo add crossplane-alpha https://charts.crossplane.io/alpha
+helm install crossplane --namespace crossplane-system crossplane-alpha/crossplane
+```
+
+### Master
+
+The `master` channel contains the latest commits, with all automated tests passing.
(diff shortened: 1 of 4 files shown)


cluster/charts/crossplane/README.md
```diff
@@ -0,0 +1,270 @@
+# Install Crossplane
```
negz: Would it make sense for this document to link to our existing documentation on these topics, rather than repeating it? Doing so may reduce our documentation maintenance burden.

hasheddan: @displague any thoughts here? Looks like we can get this merged after resolving this comment :)

displague: @negz @hasheddan Sorry for missing this.

https://hub.helm.sh/charts/jetstack/cert-manager is an example of what this README.md looks like rendered.  If think we are addressing the wrong audiences at the wrong places if we defer to our documentation in Helm hosted locations.

As an alternative, our documentation could link to Helm Hub as the best source for detailed Helm usage instructions.  Helm Hub's content is rendered from this chart published file (which is lightly modified from docs/install-crossplane.md to remove the front-matter).

```patch
diff docs/install-crossplane.md cluster/charts/crossplane/README.md
1,6d0
< ---
< title: Install Crossplane
< toc: true
< weight: 220
< indent: true
< ---
```

hasheddan: > I think we are addressing the wrong audiences at the wrong places if we defer to our documentation in Helm hosted locations.

@displague are you saying you prefer not to link to the crossplane docs in the chart README? If so, it looks like this is the same as our install page on the docs, so maybe we just setup automation to sync it here during build? As is, these instructions are already out of date from our docs :/

displague: > setup automation to sync it here during build
That would be a good solution.  I considered using a symlink, but symlinks in git repos can be troublesome in certain settings and the contents of the files are slightly different.

displague: Would you be ok with this file being sync'd via the Makefile's `manifests.annotate`?
```
sed '1 { /^---/ { :a N; /\n---/! ba; d} }' $(SOURCE_DOCS_DIR)/install-crossplane.md > $(CROSSPLAN E_CHART_DIR)/README.md
```

As we do now for chart files, we would have to commit changes to this file as needed.

hasheddan: That seems reasonable to me. I would want a check in place in CI to make sure these two files were in sync, seems like we could just run your command as part of [`check-diff`](https://github.com/crossplaneio/crossplane/blob/1b8c665bb2511f47d089444234d8762b2661f6d6/Makefile#L133).

displague: `check-diff` would detect any drift as-is, so long as a make `manifests` is run during CI (part of `reviewable`, I think).

hasheddan: Sounds great!

displague: The file is generated in the latest push, but the sed command isn't working on OSX.

displague: `sed '/^---/,/^---/d'` does the trick with both BSD and GNU sed.

Issue #1096, opened 2019-11-25
Design for Easy Stack resources


negz:
Easy stacks will install various resource classes and managed resources when an instance of their CRD is created, e.g. installing an "EasyGCP" stack might install a GKECluster resource class and the static network managed resources.

We need to figure out what exactly will be created by each cloud's stack, how the EasyFoo CR will allow the user to configure those things, etc.

suskin:
There is slightly more to this as well: there is also the UI experience to consider.

Anyway, here are some random thoughts and links.

### Existing thinking

Here is an example of what we imagine [the experience for using a wordpress stack looking like](https://github.com/suskin/template-stack-experience/tree/master/wordpress-workload/quick-start) when we have template stacks, including installing a resource pack. Here's a link to the [imagined structure of a resource pack](https://github.com/suskin/template-stack-experience/tree/master/wordpress-workload/quick-start/resource-packs/dev/gcp). If possible, it'd be ideal if we could structure the source files of our resource packs in the same way, so that we can test our ideas and have something which is easy to convert to a true template stack.

I am also thinking that the controller side of this work will be a good place to experiment with logic that we may use in the controller side of template stacks.

### Open questions

There are some parts which are ambiguous, which we'll want to clarify as part of this process. It probably makes sense to try to clarify them early, because the clarifications will affect the user experience.

* How do users configure providers, especially the credentials?
* What would the flow of the UI look like, from installing the stack to applying the resources?

muvaf:
> Here is an example of what we imagine the experience for using a wordpress stack looking like when we have template stacks, including installing a resource pack. Here's a link to the imagined structure of a resource pack.

I'd personally prefer a separation between resource classes and managed resources in the context of easy stacks. The reasoning is that managed resources may or may not refer to each other but the resource classes definitely refer to managed resources and in the context of _easy stacks_ resource classes are the top level resources that we expose to the user. It's a good peace of mind for someone who wants to fork the easy stack and make changes to know that he only needs to play with the managed resources folder.

In terms of controller side of the things, I imagine a top level folder will exist to host all possible YAML directories and controller will basically scan that top level folder recursively. So, it's just a preference in the end. I don't think folder structure will affect the execution logic.

> How do users configure providers, especially the credentials?

[...]

PR #1110, opened 2019-11-30
Easy Stacks one-pager

muvaf:

Introducing the concept of _easy stacks_ which allow a stack to be built with a set of default managed resources and resource classes and then deployed to the cluster easily in order to provide a ready-to-use environment to the user easily.

The one-pager lays down the general principles to adhere when developing such stacks. #1096 tracks what the sets of YAMLs will be in the initial implementations of easy stacks. Depending on timeline, a generic controller will be implemented following the implementation of the initial stacks.

Fixes https://github.com/crossplaneio/crossplane/issues/1095

Manual tests around whether the statements would actually make sense when implemented.
### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/design/one-pager-configuration-stacks.md
+++ b/design/one-pager-configuration-stacks.md
@@ -65,13 +65,14 @@ areas that we can improve during this process:
 
 ## Goals
 
-The main goal of this design is to make it easier for people to write sets of
-Crossplane resources and be able to manage those resource sets on a higher level.
+The main goal of this design is to make it easier for infrastructure operators
+who provide environments for others to build sets of environment resources and
+provide them to their users.
 
 It is important that the design puts forward:
 
-* A base boilerplate tooling where creating a new configuration set that has a
-  controller is as easy as changing the YAML files.
+* A base tooling for infrastructure operators who would like to package a set
+  of resources as a stack with minimal Go code and provide that stack to their users.
 * An easy way for users to deploy a set of pre-defined resources that has
   a controller which reconciles all resources deployed.
 
--- a/design/one-pager-configuration-stacks.md
+++ b/design/one-pager-configuration-stacks.md
@@ -122,19 +122,20 @@ rules to be adhered by initial configuration stacks.
   * See details about [defaulting mechanism here].
 * All the labels in the configuration stack CR is propagated down to all resources
   that it deploys.
+
 ## User Experience
 
-1. Install AWS stack and create a secret that contains AWS credentials.
-2. Install a AWS configuration stack via:
+1. Install GCP stack and create a secret that contains GCP credentials.
+2. Install a GCP configuration stack via:
 ```yaml
 # exact apiVersion is TBD.
 apiVersion: stacks.crossplane.io/v1alpha1
 kind: ClusterStackInstall
 metadata:
-  name: minimal-aws
+  name: minimal-gcp
   namespace: crossplane-system
 spec:
-  package: "crossplane/minimal-aws:latest"
+  package: "crossplane/minimal-gcp:latest"
 ```
 
 Now, I want an environment where all my database instances and kubernetes clusters
@@ -142,29 +143,39 @@ are connected to each other in a private VPC, which what this specific configura
 stack does. Create the following:
 
 ```yaml
-apiVersion: aws.configurationstacks.crossplane.io/v1alpha1
-kind: MinimalAWS
+apiVersion: gcp.configurationstacks.crossplane.io/v1alpha1
+kind: MinimalGCP
 metadata:
   name: small-infra
+  labels:
+      "foo-key": bar-value
 spec:
+  region: us-west2
+  projectID: foo-project
+  keepDefaultingAnnotations: true
   credentialsSecretRef:
-    name: aws-credentials
+    name: gcp-credentials
     namespace: crossplane-system
+    key: credentials
 ```
 
 Then I wait for `Synced` condition to become `true`. After it's done, all resources
 are deployed.
 ```yaml
-apiVersion: aws.configurationstacks.crossplane.io/v1alpha1
-kind: MinimalAWS
+apiVersion: gcp.configurationstacks.crossplane.io/v1alpha1
+kind: MinimalGCP
 metadata:
   name: small-infra
   labels:
     "foo-key": bar-value
 spec:
+  region: us-west2
+  projectID: foo-project
+  keepDefaultingAnnotations: true
   credentialsSecretRef:
-    name: aws-credentials
+    name: gcp-credentials
     namespace: crossplane-system
+    key: credentials
 status:
   conditions:
   - lastTransitionTime: "2019-12-03T23:16:58Z"
@@ -175,63 +186,73 @@ status:
 
 An example deployed resource would be:
 ```yaml
-apiVersion: database.aws.crossplane.io/v1beta1
-kind: RDSInstanceClass
+apiVersion: database.gcp.crossplane.io/v1beta1
+kind: CloudSQLInstanceClass
 metadata:
-  name: small-infra-mysql
-  ownerReferences:
-  - apiVersion: aws.configurationstacks.crossplane.io/v1alpha1
-    kind: MinimalAWS
-    name: small-infra
-    uid: 1ca16960-973b-4d58-a6fa-5696638ec631
   labels:
-    "aws.configurationstacks.crossplane.io/name": small-infra
-    "aws.configurationstacks.crossplane.io/uid": 1ca16960-973b-4d58-a6fa-5696638ec631
+    # All labels on MinimalGCP are propagated down to all resources.
     "foo-key": bar-value
+    # Default labels for all deployed resources.
+    gcp.resourcepacks.crossplane.io/name: minimal-setup
+    gcp.resourcepacks.crossplane.io/uid: 34646233-f58e-4c99-b0a8-0d766533b12c
+  annotations:
+    # The defaulting annotation is kept since keepDefaultingAnnotations was true.
+    # Otherwise, it'd have been removed.
+    resourceclass.crossplane.io/is-default-class: "true"
+  name: minimal-setup-cloudsqlinstance-mysql
+  ownerReferences:
+  # Once the referred MinimalGCP instance is deleted, this resource will be
+  # deleted by Kubernetes api-server.
+  - apiVersion: gcp.resourcepacks.crossplane.io/v1alpha1
+    kind: MinimalGCP
+    name: minimal-setup
+    uid: 34646233-f58e-4c99-b0a8-0d766533b12c
 specTemplate:
-  writeConnectionSecretsToNamespace: crossplane-system
   forProvider:
-    dbInstanceClass: db.t2.small
-    masterUsername: root
-    vpcSecurityGroupIDRefs:
-      - name: small-infra-rds-security-group
-    dbSubnetGroupNameRef:
-      name: small-infra-dbsubnetgroup
-    allocatedStorage: 20
-    engine: mysql
-    skipFinalSnapshotBeforeDeletion: true
+    databaseVersion: MYSQL_5_7
+    # Propagated from MinimalGCP instance.
+    region: us-west2
+    settings:
+      dataDiskSizeGb: 10
+      dataDiskType: PD_SSD
+      ipConfiguration:
+        ipv4Enabled: false
+        privateNetworkRef:
+          # Hard-coded value was "network" but since Network with name "network"
+          # has a new name, the ref here is also updated.
+          name: minimal-setup-network
+      tier: db-n1-standard-1
   providerRef:
-    name: small-infra-aws-provider
+    name: minimal-setup-gcp-provider
   reclaimPolicy: Delete
+  writeConnectionSecretsToNamespace: crossplane-system
+
 ```
 
-There could be several instances of `MinimalAWS` custom resource and each would 
+There could be several instances of `MinimalGCP` custom resource and each would 
 have their own similar environment with resources that have different names.
 
 ## Technical Implementation
 
 Reconciliation will mainly consist of the following steps:
-
-1. Edit `kustomization.yaml` to look like the following:
+1. A `Kustomization` overlay object will be generated that looks like the following:
 
 ```yaml
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
-resources:
-  - gcp
-
-namePrefix: CRNAME-
+namePrefix: <CRNAME>-
 commonLabels:
-  "gcp.configurationstacks.crossplane.io/name": CRNAME
-  "gcp.configurationstacks.crossplane.io/uid": CRUID
+  "gcp.configurationstacks.crossplane.io/name": <CRNAME>
+  "gcp.configurationstacks.crossplane.io/uid": <CRUID>
 ```
 
-2. Call `kustomize` to generate resource YAMLs.
-3. Read the stream of YAMLs from kustomize output and apply the following logics:
-   * Add `ownerReference` that refers to the CR that is being reconciled to all objects.
-   * Change the secret reference in the `Provider` resource.
-   * Other custom things you'd like to do...
-4. _Apply_ all resources. Set reconciliation condition to success if no error is
+2. Call custom patch functions that consumer of the reconciler provided in order
+   to make changes on `Kustomization` object.
+3. Call Kustomize and generate the resources.
+4. Read the stream of YAMLs from kustomize output.
+5. Call custom patch functions that consumer of the reconciler provided in order
+   to make changes on the generated resources before deployment.
+6. _Apply_ all resources. Set reconciliation condition to success if no error is
    present.
 
 Note that if the CR has deletion timestamp, we do not reconcile at all, letting
@@ -255,7 +276,7 @@ like the following:
 │   │   │   ├── network.yaml
 │   │   │   └── subnetwork.yaml
 │   │   ├── database
-│   │   │   ├── cloudsqlinstancemysql.yaml
+│   │   │   ├── cloudsqlinstanceclass.yaml
 │   │   │   ├── kustomization.yaml
 │   │   │   └── kustomizeconfig.yaml
 │   │   ├── kustomization.yaml
@@ -277,9 +298,123 @@ sense for developer to have a folder for each tier under `resources` folder. It'
 basically up to you to try various structures as long as kustomize is able to
 work through your structure.
 
+Note that nothing in `resources` folder is changed. The reconciler generates a
+new overlay with its own `kustomization.yaml` in a temporary directory and refers
+to the `resources` folder.
+
 [UI annotations] should be present to make it easy for frontend software to process
 the stack.
- 
+
+### Custom Patchers
+
+The configuration stack reconciler that has two types of patcher functions where
+developer can intercept the reconciler flow and provide their own logic:
+* `KustomizationPatcher`: Its signature includes the generic `ParentResource` object
+  that represents the stack CR and `Kustomization` object that represents the
+  overlay `kustomization.yaml` file.
+  * Developers who want to make additions to the default `kustomization.yaml` file
+    with data from runtime can provide their own patchers to the pipeline.
+* `ChildResourcePatcher`: Its signature includes the generic `ParentResource` object
+  as well as the list of the generated `ChildResource`s that will be deployed.
+  * Developers who'd like to make changes to the resources generated via `kustomize`
+    will provide their own functions.
+
+Configuration stack reconciler will have default patchers for the functionality
+that is expected to be common for all configuration stacks such as label propagation
+from stack CR to deployed resources.
+
+### Referencing
+
+We will use Kustomize [custom transformer configurations] to achieve the referencing
+behaviors.
+
+It's developer's responsibility to declare the reference dependencies between
+the resources. Kustomize supports the following referencers as of writing:
+* Name Referencers: If a resource references to another one, you need to declare
+  this dependency as a kustomize config so that when a different name is generated
+  for the referred resource, related references under the referrer are also
+  changed.
+* Variant Referencers: If a resource needs a value from another resource, you can
+  use `$(VALUE)` and then in `Kustomization` object, you can declare where to fetch
+  that `VALUE`. However, Kustomize requires you to explicitly declare which fields
+  of the CRD you expect to have a variant like `$(VALUE)`. So, developer needs to
+  declare this in the kustomize config file.
+  
+An example kustomize config file looks like following:
+
+```yaml
+nameReference:
+  - kind: Provider
+    fieldSpecs:
+      - path: specTemplate/providerRef/name
+        kind: CloudMemoryInstanceClass
+varReference:
+  - path: specTemplate/forProvider/region
+    kind: CloudMemorystoreInstanceClass
+```
+
+What the `nameReference` in the snippet above says is that the kind
+`CloudMemoryInstanceClass`'s field path `specTemplate/providerRef/name` refers
+to the name of the kind `Provider`. So, during transformations, if the `Provider`
+resource with name in `specTemplate.providerRef.name` of the resources with kind
+`CloudMemoryInstanceClass` ends up with a different name, go ahead and update
+the value in `specTemplate.providerRef.name`.
+
+What the `varReference` declares is that during variant calculations, the path
+`specTemplate/forProvider/region` of resources of kind `CloudMemorystoreInstanceClass`
+should be taken into consideration. If the value is bare string, nothing will be
+done. But if it's like `$(REGION)` and you did add a `Var` with name `REGION` to
+the list `Vars` of the `Kustomization` object, then the calculated value will be
+written to `specTemplate.forProvider.region` of the said resource. A `Vars` array
+looks like the following:
+```yaml
+vars:
+- name: REGION
+  objref:
+    kind: MinimalGCP
+    apiVersion: gcp.configurationstacks.crossplane.io/v1alpha1
+    name: <CR NAME>
+  fieldref:
+    fieldpath: spec.region
+```
+
+There is a tricky part when you'd like to refer a value in your CR since the CR
+instance doesn't exist in the `resources` folder. Because of this reason,
+the generic reconciler will dump the CR instance YAML file to the kustomization
+folder so that Kustomize takes it into the calculation but it will remove it
+from the resource list that Kustomize returned after the generation is completed.
+
+Here is an example flow of adding a new field, say `region`, to the stack's CR
+and use it in the child resources:
+* In the `resources` folder, go to resources that you'd like the change its `region`
+  property, put `$(REGION)` string.
+* In `kustomizeconfig.yaml`, declare the field like (create that file if it doesn't
+  exist):
+```yaml
+varReference:
+  - path: specTemplate/forProvider/region
+    kind: CloudMemorystoreInstanceClass
+```
+* In `kustomization.yaml` file of the same folder, make sure the `kustomizeconfig.yaml`
+  is declared as Kustomize configuration like:
+```yaml
+configurations:
+  - kustomizeconfig.yaml
+```
+
+* In your controller code, add a `KustomizationPatcher` that adds the necessary
+  `Var` object to the `Vars` array of `Kustomization` file that refers to your
+  CR instance and the field that you want `region` value to be taken. An example
+  `Var` object would look like:
+```go
+{
+  Name:   "REGION",
+  ObjRef: ref,
+  FieldRef: types.FieldSelector{
+    FieldPath: "spec.region",
+  },
+},
+```
 
 ## Alternatives Considered
 
@@ -295,3 +430,4 @@ the stack.
 [UI annotations]: https://github.com/crossplaneio/crossplane/blob/5758662818fc1e840adbfbf1a9fb37b87c3d5a5c/design/one-pager-stack-ui-metadata.md
 [class and claim]: https://static.sched.com/hosted_files/kccncna19/2d/kcconna19-eric-tune.pdf
 [defaulting mechanism here]: https://github.com/crossplaneio/crossplane/blob/c38561d/design/one-pager-simple-class-selection.md#unopinionated-resource-claims
+[custom transformer configurations]: https://github.com/kubernetes-sigs/kustomize/blob/master/examples/transformerconfigs/crd/README.md
--- a/design/one-pager-configuration-stacks.md
+++ b/design/one-pager-configuration-stacks.md
@@ -110,7 +110,18 @@ rules to be adhered by initial configuration stacks.
   owner.
 * Stack type will be `ClusterStack` as most of the resources it deploys are
   cluster-scoped.
-
+* CRD of the configuration stack should have a boolean spec field called
+  `keepDefaultingLabels` which indicates what happens to the resource classes that
+  are marked default.
+  * By default, the zero-value of the boolean is _false_. Meaning, if user is
+    indifferent we remove that special default label from all resource classes
+    to avoid conflict and unexpected randomization between resource classes.
+  * Users who deploy the CR will need to mark this field as _true_ if they'd like
+    to keep the defaults, which is expected to appear on only one configuration
+    stack CR in the cluster.
+  * See details about [defaulting mechanism here].
+* All the labels in the configuration stack CR is propagated down to all resources
+  that it deploys.
 ## User Experience
 
 1. Install AWS stack and create a secret that contains AWS credentials.
@@ -148,6 +159,8 @@ apiVersion: aws.configurationstacks.crossplane.io/v1alpha1
 kind: MinimalAWS
 metadata:
   name: small-infra
+  labels:
+    "foo-key": bar-value
 spec:
   credentialsSecretRef:
     name: aws-credentials
@@ -174,6 +187,7 @@ metadata:
   labels:
     "aws.configurationstacks.crossplane.io/name": small-infra
     "aws.configurationstacks.crossplane.io/uid": 1ca16960-973b-4d58-a6fa-5696638ec631
+    "foo-key": bar-value
 specTemplate:
   writeConnectionSecretsToNamespace: crossplane-system
   forProvider:
@@ -280,3 +294,4 @@ the stack.
 [Dependency list]: https://github.com/crossplaneio/crossplane/blob/98e8520e2a2285cd6944fcd67fbef427299891e8/design/design-doc-stacks.md#stack-crd
 [UI annotations]: https://github.com/crossplaneio/crossplane/blob/5758662818fc1e840adbfbf1a9fb37b87c3d5a5c/design/one-pager-stack-ui-metadata.md
 [class and claim]: https://static.sched.com/hosted_files/kccncna19/2d/kcconna19-eric-tune.pdf
+[defaulting mechanism here]: https://github.com/crossplaneio/crossplane/blob/c38561d/design/one-pager-simple-class-selection.md#unopinionated-resource-claims
--- a/design/one-pager-configuration-stacks.md
+++ b/design/one-pager-configuration-stacks.md
@@ -1,4 +1,4 @@
-# Configuration Stacks
+# Resource Packs
 
 * Owner: Muvaffak Onus (@muvaf)
 * Reviewers: Crossplane Maintainers
@@ -55,7 +55,7 @@ areas that we can improve during this process:
   may not seem like a big hurdle, changing some high level parameters of that
   environment is cumbersome _after_ it's deployed.
 * A user who is familiar with AWS may not be familiar with Azure or GCP. So,
-  ready-made configuration stacks that promise a very similar setup would be
+  ready-made resource packs that promise a very similar setup would be
   useful for them to see how they can achieve similar setups and run various
   benchmarks according to their business needs.
 * After you run `kubectl apply -f <directory of YAMLs>`, a dependency graph of
@@ -66,14 +66,14 @@ areas that we can improve during this process:
 ## Goals
 
 The main goal of this design is to make it easier for people to write sets of
-configurations and be able to manage those configuration sets on a higher level.
+Crossplane resources and be able to manage those resource sets on a higher level.
 
 It is important that the design puts forward:
 
 * A base boilerplate tooling where creating a new configuration set that has a
   controller is as easy as changing the YAML files.
-* An easy way for users to deploy a set of pre-defined configurations that has
-  a controller which reconciles the set of the resources deployed.
+* An easy way for users to deploy a set of pre-defined resources that has
+  a controller which reconciles all resources deployed.
 
 ## Proposal
 
@@ -88,9 +88,9 @@ and updates the YAMLs that are packaged in the stack image according to user's
 input on the CR instance of that stack that user creates.
 
 As sky is the limit when it comes to stacks, we need to put down some general
-rules to be adhered by initial configuration stacks.
+rules to be adhered by initial resource packs.
 
-* It should be possible to have different sets of configurations in one stack
+* It should be possible to have different sets of resources in one stack
   and user should be able to choose from them via the CR that is an instance
   of stack's CRD.
   * Note that it's highly preferred to have only one custom resource definition that
@@ -105,12 +105,12 @@ rules to be adhered by initial configuration stacks.
 * Controller should update the resources continuously and treat the given CR as
   the source of truth even though user manually changes the resources that are
   deployed by the stack.
-* CRD of the configuration stack should be cluster-scoped since the author of
-  the CR instances of the configuration stack is assumed to be the infrastructure
+* CRD of the resource pack should be cluster-scoped since the author of
+  the CR instances of the resource pack is assumed to be the infrastructure
   owner.
 * Stack type will be `ClusterStack` as most of the resources it deploys are
   cluster-scoped.
-* CRD of the configuration stack should have a boolean spec field called
+* CRD of the resource pack should have a boolean spec field called
   `keepDefaultingLabels` which indicates what happens to the resource classes that
   are marked default.
   * By default, the zero-value of the boolean is _false_. Meaning, if user is
@@ -120,13 +120,13 @@ rules to be adhered by initial configuration stacks.
     to keep the defaults, which is expected to appear on only one configuration
     stack CR in the cluster.
   * See details about [defaulting mechanism here].
-* All the labels in the configuration stack CR is propagated down to all resources
+* All the labels in the resource pack CR is propagated down to all resources
   that it deploys.
 
 ## User Experience
 
 1. Install GCP stack and create a secret that contains GCP credentials.
-2. Install a GCP configuration stack via:
+2. Install a GCP resource pack via:
 ```yaml
 # exact apiVersion is TBD.
 apiVersion: stacks.crossplane.io/v1alpha1
@@ -143,7 +143,7 @@ are connected to each other in a private VPC, which what this specific configura
 stack does. Create the following:
 
 ```yaml
-apiVersion: gcp.configurationstacks.crossplane.io/v1alpha1
+apiVersion: gcp.resourcepacks.crossplane.io/v1alpha1
 kind: MinimalGCP
 metadata:
   name: small-infra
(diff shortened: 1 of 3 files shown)


design/one-pager-easy-stacks.md
```diff
@@ -0,0 +1,294 @@
...
+are connected to each other in a private VPC, which what this specific configuration
+stack does. Create the following:
+
+```yaml
+# exact apiVersion is TBD.
+apiVersion: aws.configurationstacks.crossplane.io/v1alpha1
+kind: EasyAWS
+metadata:
+  name: project-future
+  namespace: dev
```
prasek: per comment above -- an environment (dev/test, stage, prod) generally maps to crossplane instance, so could this be cluster-scoped? If not perhaps we should recommend it go into the crossplane-system namespace [like we do for provider secret today](https://github.com/crossplaneio/crossplane/blob/master/docs/cloud-providers/aws/aws-provider.md#optional-setup-aws-provider-manually).

design/one-pager-easy-stacks.md
```diff
@@ -0,0 +1,294 @@
...
+
+To achieve the mentioned goals, we can publish the YAML files in a stack with a
+controller that has its own CRD with high level parameters. Controller deploys
+and updates the YAMLs that are packaged in the stack image according to user's
+input on the CR instance of that stack that user creates.
+
+As sky is the limit when it comes to stacks, we need to put down some general
+rules to be adhered by initial configuration stacks.
+
+* CRDs of the resources that will be deployed by the stack are declared as
```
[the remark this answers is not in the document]
muvaf: Yes, like this example https://github.com/crossplaneio/crossplane/blob/master/design/design-doc-stacks.md#example-appyaml

design/one-pager-easy-stacks.md
```diff
@@ -0,0 +1,294 @@
...
+  dependency so that user knows what CR instances will be deployed by the given
+  configuration stack.
+* It should be possible to have different sets of configurations in one stack
+  and user should be able to choose from them via the CR that is an instance
+  of stack's CRD.
+  * Note that it's highly preferred to have 1 custom resource definition that
+    the stack uses and expose the high level variations of the parameters through
+    `spec` instead of having different custom resource definitions for each class
+    of configuration.
+* A minimal high level set of configurations are exposed to the user through
```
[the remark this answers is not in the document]
muvaf: > if this will include passing the provider secret content via environment config CR

Not sure about passing down the secret _content_ via CR.

I think taking a `Secret` name and namespace in a field like `spec.credentialsSecretRef` is the easiest option. This way the client that will put the credentials doesn't have to know how to construct a specific `Provider` CR. Client will;
1. Create a secret.
2. Create a `EasyAWS` CR with `spec.credentialsSecretRef` that points to the created secret.

> requiredResources:
> resource:

I couldn't find a reference to these properties of `app.yaml`. Could you point me to a resource that explains how they work?

> Note that maxAllowed would limit the number of environment config CRs created, at least for now as a simplifying assumption, unless it would make sense to allow multiple env. config CRs for a given config stack in the first release.

Limiting the number of CRs to 1 would indeed simplify the initial implementation. But I'm not aware of a way to tell api-server not to accept creation call of specific CR after some point.

design/one-pager-easy-stacks.md
```diff
@@ -0,0 +1,294 @@
...
+  the stack's CR, such as `Provider` reference to refer in the deployed resource
+  or `region`.
+* Controller should signal readiness of all resources in CR's status.
+* All resources should be labelled referring to the CR instance of the stack.
+* In YAMLs, resources should refer to each other using cross-resource references
+  wherever possible.
+* **TBD** Controller should update the resources continuously and treat the given CR as
+  the source of truth even though user manually changes the resources that are
+  deployed by the stack.
+* **TBD** CRD of the configuration stack should be namespaced so that different
```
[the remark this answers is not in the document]
muvaf: How I imagined is like the following; the users and UI clients can create as many config CRs as they want. To prevent the name collisions of deployed cluster-scoped resources, we'd prefix the names with `namespace-name-` of the specific config CR. This way people can have many configuration environments independently in one Crossplane instance.

This thinking treats the creator of the config CR as app user rather than infra owner role that  who has cluster-wide access. Technically, implementing the CRDs namespaced or cluster-scoped requires same amount of effort. So, if the creator is assumed to be the infra owner and we don't expect app developers to create config CRs, we can go with cluster-scoped.

design/one-pager-easy-stacks.md
```diff
@@ -0,0 +1,294 @@
...
+  * For resource classes; if the resource exists in the api-server it's considered
+    to be ready.
+  * For managed resources; if the `Ready` condition status is `true`, then it's
+    considered to be ready.
+  * For other possible resources, the developer chooses what to look for as
+    readiness.
+* Apply; continuously reproduce the resources from YAMLs and apply it.
+* Deletion: should be done via label selector when the CR is deleted.
+
+In the `Apply` phase, there will be some processing phasse before calling `Create`
```
hasheddan:
```suggestion
In the `Apply` phase, there will be some processing phase before calling `Create`
```

design/one-pager-easy-stacks.md
```diff
@@ -0,0 +1,294 @@
...
+private network to connect to each other and expose only your application's IP
+to the world. There are usually a few resources that you need to create to achieve
+this like VPC, sub-network, internet gateway etc. After creating these resources,
+you need to refer to them in your resource classes that you author for database
+and cluster claims to use.
+
+Today, it's possible to do all this by manually configuring each resource YAML
+files and run `kubectl apply -f <directory of YAMLs>`. However, there are two areas
+that we can improve during this process:
+* The preparation phase of these YAMLs requires to be very careful and changing
```
hasheddan:
```suggestion
* The preparation phase of these YAMLs requires a user to be very careful and changing
```

design/one-pager-easy-stacks.md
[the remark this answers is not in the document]
hasheddan: > But I'm not aware of a way to tell api-server not to accept creation call of specific CR after some point.

As of Kubernetes 1.15 you can use a `ResourceQuota` object to [restrict the count](https://kubernetes.io/docs/concepts/policy/resource-quotas/#object-count-quota) of custom resource instances in a namespace. Maybe we could create `ResourceQuota` objects on stack installation based on the `maxAllowed` fields!

design/one-pager-easy-stacks.md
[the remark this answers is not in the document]
hasheddan: +1 for cluster-scoped. I think that these stacks could be used for base-level infrastructure across multiple teams, so it seems likely they will likely be deployed by infra-owners in most cases.

design/one-pager-easy-stacks.md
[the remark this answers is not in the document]
displague: The deployed resources of a Stack are not contained in the `dependsOn` block, they do not exist in the `app.yaml` file.  The StackManager walks the CRD files in the Stack OCI, installs them all, and creates a Stack record with a `customresourcedefinitions` list (as in `stack/stack-gcp`):

```yaml
 customresourcedefinitions:
  - apiVersion: cache.gcp.crossplane.io/v1beta1
    kind: CloudMemorystoreInstanceClass
  - apiVersion: cache.gcp.crossplane.io/v1beta1
    kind: CloudMemorystoreInstance
  - apiVersion: compute.gcp.crossplane.io/v1alpha3
    kind: GKEClusterClass
  - apiVersion: compute.gcp.crossplane.io/v1alpha3
    kind: GKECluster
  - apiVersion: compute.gcp.crossplane.io/v1alpha3
    kind: GlobalAddress
  - apiVersion: compute.gcp.crossplane.io/v1alpha3
    kind: Network
  - apiVersion: compute.gcp.crossplane.io/v1alpha3
    kind: Subnetwork
  - apiVersion: database.gcp.crossplane.io/v1beta1
    kind: CloudSQLInstanceClass
  - apiVersion: database.gcp.crossplane.io/v1beta1
    kind: CloudSQLInstance
  - apiVersion: gcp.crossplane.io/v1alpha3
    kind: Provider
  - apiVersion: servicenetworking.gcp.crossplane.io/v1alpha3
    kind: Connection
  - apiVersion: storage.gcp.crossplane.io/v1alpha3
    kind: BucketClass
  - apiVersion: storage.gcp.crossplane.io/v1alpha3
    kind: Bucket
  dependsOn:
  - crd: '*.cache.crossplane.io/v1alpha1'
  - crd: '*.compute.crossplane.io/v1alpha1'
  - crd: '*.database.crossplane.io/v1alpha1'
  - crd: '*.core.crossplane.io/v1alpha1'
  - crd: '*.storage.crossplane.io/v1alpha1'
  - crd: '*.workload.crossplane.io/v1alpha1'
```

design/one-pager-easy-stacks.md
```diff
@@ -0,0 +1,294 @@
...
+  useful for them to see how they can achieve similar setups and run various
+  benchmarks according to their business needs.
+* After you run `kubectl apply -f <directory of YAMLs>`, a dependency graph of
+  resources starts to resolve. However, it's not always clear for user to know _when_
+  everything is actually ready to use unless they know readiness of which resources
+  signal the readiness of the whole environment.
+
+## Goals
+
+The main goal of this design is to make it easier for people to write sets of
```
suskin: > main goal of this design

I would also try to mention the goal for the end user, which is to make it easier to set up a working environment.

design/one-pager-easy-stacks.md
```diff
@@ -0,0 +1,294 @@
...
+configurations and be able to manage those configuration sets on a higher level.
+
+It is important that the design puts forward:
+
+* A base boilerplate package where creating a new configuration set that has a
+  controller is as easy as changing the YAML files.
+* A way to declare pre-requisites that needs to be satisfied by the cluster
```
suskin: What is the benefit of including this in the design? I would not expect a formal mechanism to be needed at this point, since people can just install things in the right order according to some documentation. Is this referring to listing the crds it needs to be able to CRUD?

design/one-pager-easy-stacks.md
```diff
@@ -0,0 +1,294 @@
...
+  for the configuration set to work.
+* A way to change a high level configuration by changing 1 value that is then
```
suskin: Grammar/style nitpick: normally the numbers one through ten would be written out, so `changing one value` in this case

design/one-pager-easy-stacks.md
```diff
@@ -0,0 +1,294 @@
...
+As sky is the limit when it comes to stacks, we need to put down some general
+rules to be adhered by initial configuration stacks.
+
+* CRDs of the resources that will be deployed by the stack are declared as
+  dependency so that user knows what CR instances will be deployed by the given
+  configuration stack.
+* It should be possible to have different sets of configurations in one stack
+  and user should be able to choose from them via the CR that is an instance
+  of stack's CRD.
+  * Note that it's highly preferred to have 1 custom resource definition that
```
suskin: > Note that it's highly preferred to have 1 custom resource definition

What is the reason for this preference?

design/one-pager-easy-stacks.md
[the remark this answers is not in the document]
suskin: > I think taking a Secret name and namespace in a field like spec.credentialsSecretRef is the easiest option.

I agree with this. Creating a secret separately and then putting a reference in the spec field of the single CR seems like it would be more manageable than requiring multiple resources (and coordinating that).

design/one-pager-easy-stacks.md
```diff
@@ -0,0 +1,294 @@
...
+    the stack uses and expose the high level variations of the parameters through
+    `spec` instead of having different custom resource definitions for each class
+    of configuration.
+* A minimal high level set of configurations are exposed to the user through
+  the stack's CR, such as `Provider` reference to refer in the deployed resource
+  or `region`.
+* Controller should signal readiness of all resources in CR's status.
```
suskin: What is `readiness`? How does the controller know they are ready?

design/one-pager-easy-stacks.md
```diff
@@ -0,0 +1,294 @@
...
+  wherever possible.
+* **TBD** Controller should update the resources continuously and treat the given CR as
+  the source of truth even though user manually changes the resources that are
+  deployed by the stack.
+* **TBD** CRD of the configuration stack should be namespaced so that different
+  teams/users can create their own configured environment.
+* Stack type will be `ClusterStack` as most of the resources it deploys are
+  cluster-scoped.
+
+Name of the concept is **TBD**; _EasyStack_, _ConfigurationStack_,
```
suskin: "Resource Pack" is another candidate, and my personal preference : ). I am also liking the "Configuration Stack" option though (or "Configuration Pack").

design/one-pager-easy-stacks.md
```diff
@@ -0,0 +1,294 @@
...
+
+Other people in different namespaces can create their own instances of `EasyAWS`
+custom resource and have their own similar environment with different names.
+
+## Technical Implementation
+
+### Metadata of The Stack
+
+[Dependency list] has to be filled with each kind of the resource that the stack
+deploys, i.e. f the stack only deploys `RDSInstanceClass` resource, it should
```
suskin: `i.e. f` seems like it should be `i.e. if`.

design/one-pager-easy-stacks.md
```diff
@@ -0,0 +1,294 @@
...
+
+All directories containing base YAML files should be in a top-level directory
+of the repository named `resources` so that it's easier for others to fork & change.
+
+Controller will scan this folder recursively to get all YAML files. So, the folder
+structure inside that `resources` directory won't matter to the controller but
+it's good to separate the folders cleanly so that it's easier to fork & change.
+
+There are mainly 2 types of resources that configuration stacks will usually
+need to deploy; resource classes and managed resources. Under `resources` folder
```
suskin: Grammar nitpick: I would normally use a colon when giving an example or a list of examples, so `deploy: `. Semicolons (`;`) are for [connecting related clauses](https://en.wikipedia.org/wiki/Semicolon#Usage); often a semicolon can be used where a period would also work.

design/one-pager-easy-stacks.md
```diff
@@ -0,0 +1,294 @@
...
+we can have 2 separate folders for each of those resource types; `classes` and
+`services`. Under these folders, resources should be in directories recursively
+named after their group and kind. An example folder structure:
```
suskin: Maybe I am misunderstanding, but it doesn't seem like the example folder structure matches "`directories recursively named after their group and kind`". It seems more like the folder is the group, and the filename is `{{kind}}.yaml`. Am I reading the example correctly?

design/one-pager-easy-stacks.md
```diff
@@ -0,0 +1,294 @@
...
+│           ├── internetgateway.yaml
+│           ├── routetable.yaml
+│           ├── securitygroup.yaml
+│           ├── subnet.yaml
+│           └── vpc.yaml
+```
+
+Example above is a suggestion for stacks with one tier of configuration.
+There could also be cases where one configuration stack has the YAML files for
+different tiers and a `spec` allows to deploy one of them, in that case it'd make
```
suskin: > a `spec` allows to deploy one of them

Hmm, I'm wondering how that would be implemented. But, I'm not sure we need to figure that out before merging this design.

design/one-pager-easy-stacks.md
```diff
@@ -0,0 +1,294 @@
...
+sense for developer to have a folder for each tier under `resources` folder.
+
+### Reconciliation
+
+There are mainly 3 responsibility of the controller:
+* When all deployed resources are ready, `status.state` should be `ready`.
```
suskin: I like this

design/one-pager-easy-stacks.md
```diff
@@ -0,0 +1,294 @@
...
+  * For resource classes; if the resource exists in the api-server it's considered
+    to be ready.
+  * For managed resources; if the `Ready` condition status is `true`, then it's
+    considered to be ready.
+  * For other possible resources, the developer chooses what to look for as
+    readiness.
+* Apply; continuously reproduce the resources from YAMLs and apply it.
+* Deletion: should be done via label selector when the CR is deleted.
```
suskin: Are we able to use garbage collection? I like the label idea, but I want to check.

design/one-pager-easy-stacks.md
```diff
@@ -0,0 +1,294 @@
...
+
+In the `Apply` phase, there will be some processing phasse before calling `Create`
+for the resources:
+
+1. Names of the resources should be converted to the following format
+  `<CR namespace>-<CR name>-<Name given in YAML>` to prevent name collisions
```
suskin: Is there a name length limit, and could we run into problems? Another option is to include the CR UID in the name, though it's a little less human-friendly.

design/one-pager-easy-stacks.md
```diff
@@ -0,0 +1,294 @@
...
+  since most of the resources stack deploys are cluster-scoped.
+  So, the name that appears on YAML files are valid for referencing each other
+  while building the stack but in actual operation, do not expect exact names
+  to appear.
+  * This will require cross-resource references to be scanned and replaced with
```
suskin: Hmm, this feels similar to having a templated name; what are your thoughts on using explicit variables to represent the values that will be injected into the yamls? That way, the yamls are closer to what will be applied, which people will like for predictability.

design/one-pager-easy-stacks.md
suskin: Also, I am thinking that we may want to require that it is a single CRD at this time, because it seems like in order to support multiple we would need a lot of the configuration that we are designing as part of the template stacks work.

design/one-pager-easy-stacks.md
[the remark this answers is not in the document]
negz: > I think taking a Secret name and namespace in a field like spec.credentialsSecretRef is the easiest option.

Throwing my support behind this too - Secrets are the right place for credentials. Let's not include sensitive credentials inline in our custom resources; it shouldn't be too onerous to create a (provider credentials) `Secret` and an `EasyAWS` that references it in one fell swoop.

> Limiting the number of CRs to 1 would indeed simplify the initial implementation.

I can't recall ever coming across an example of a singleton resource in Kubernetes (e.g. only one instance of `kind: Foo` is allowed per namespace). I suspect that it would not be significantly more work to cleanly handle multiple instances of each easy stack custom resource being created in one namespace, and also suspect that the user experience would be fragile/surprising/weird if we _did not_ support this.

design/one-pager-easy-stacks.md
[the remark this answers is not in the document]
negz: Dropping a note here that if these easy stack custom resources are namespaced we won't be able to rely on garbage collection to cleanup the cluster scoped managed resources and classes they create, thus introducing a touch more complexity to handle cleanup. I'm thinking we should assume installing an easy stack is the domain of the infrastructure operator (and thus they're cluster scoped) unless we think it's really compelling to allow app owners to install them.

design/one-pager-easy-stacks.md
negz: https://www.slideshare.net/SarahZelechoski/how-up-running-culture-is-devaluing-engineering-work

I watched the above presentation recently and it resonated with me. It cautions against branding tools that partially abstract _really complicated stuff_ (like deploying and managing cloud infrastructure) as making those things "easy" - see the deck for more context. Therefore I'd prefer a name that either puts no particular value on what the thing does like `ResourcePack` (this is a pack of resources), or something that emphasises a little more specifically the value they bring. I think they intend to quickly provide infrastructure operators a basic configuration foundation on which to build, so something like `QuickConfig`, or `BaseConfig` could make sense.

negz: Are there any emergent patterns around stack names and the custom resources they install?

There's a few _classes of stack_, e.g. "infrastructure stack", "application stack", "template stack". Then there are specific instances of each class of stack, like `stack-gcp`, `stack-azure`, and `stack-wordpress`. We don't have data yet but I suspect many app stacks will be named `stack-foo` and define exactly one custom resource of `kind: Foo`.

So, perhaps:

* This new class of stack is an "infrastructure config stack".
* Each instances of an infrastructure config stack is named like `stack-aws-config-upbound`, or `stack-aws-config-acme` ("this the Upbound variant of the AWS infrastructure configuration stack").
* They each define one custom resource something like `kind: AWSConfigUpbound`.

🤷‍♂

design/one-pager-easy-stacks.md
negz: Do we get much value from modelling the fact that the entire configuration set is ready?

Resource classes have no concept of readiness - if they exist they're ready - so I imagine this would be communicating that the various supporting managed resources (VPCs, etc) were ready to use. Will anything block on this readiness (e.g. you can't create a `KubernetesApplication` until the `EasyAWS` is ready), or would it just be a signal to the infrastructure operator that everything came up okay?

design/one-pager-easy-stacks.md
```diff
@@ -0,0 +1,294 @@
...
+apiVersion: aws.configurationstacks.crossplane.io/v1alpha1
+kind: EasyAWS
+metadata:
+  name: project-future
+  namespace: dev
+spec:
+  providerRef:
+    name: aws-provider
+status:
+  state: ready
```
negz: Most Crossplane resources today use a conditioned status to represent readiness. Should we do so here?

design/one-pager-easy-stacks.md
```diff
@@ -0,0 +1,294 @@
...
+
+There are mainly 3 responsibility of the controller:
+* When all deployed resources are ready, `status.state` should be `ready`.
+  * For resource classes; if the resource exists in the api-server it's considered
+    to be ready.
+  * For managed resources; if the `Ready` condition status is `true`, then it's
+    considered to be ready.
+  * For other possible resources, the developer chooses what to look for as
+    readiness.
+* Apply; continuously reproduce the resources from YAMLs and apply it.
```
negz: Do you expect there will be one generic controller loading this YAML as `*unstructured.Unstructured`, or a different controller for each stack variant that knows the actual kinds it supports at the code level?

It doesn't seem like the controller will really care much about the shape of the resources with two exceptions - it will want to add metadata (e.g. set labels) when creating resources, and will need to know how each resource reflects it readiness.

design/one-pager-easy-stacks.md
negz: I'd like to understand a little better how the controller would scan and replace these cross resource references. If the controller is operating on raw YAML (e.g. `*unstructured.Unstructured`) I suspect it would be pretty hard to identify them and determine what they should be referencing.

design/one-pager-easy-stacks.md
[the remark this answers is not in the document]
muvaf: From comments and our chat yesterday, the direction we're taking is expose a field to accept a secret ref and support multiple CR instances.

design/one-pager-easy-stacks.md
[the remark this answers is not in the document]
muvaf: Cluster-scoped sounds good.

design/one-pager-easy-stacks.md
```diff
@@ -0,0 +1,294 @@
...
+  benchmarks according to their business needs.
+* After you run `kubectl apply -f <directory of YAMLs>`, a dependency graph of
+  resources starts to resolve. However, it's not always clear for user to know _when_
+  everything is actually ready to use unless they know readiness of which resources
+  signal the readiness of the whole environment.
+
+## Goals
+
+The main goal of this design is to make it easier for people to write sets of
+configurations and be able to manage those configuration sets on a higher level.
```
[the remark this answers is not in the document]
muvaf: You deploy a set of configurations, which is a bunch of YAML files. Then manage them through the CR. _Managing_ in this context usually means controlling their lifecycle via create/update/delete methods. Update could cover a surface as minimal as changing the credential secret or as big as changing the region on each and every resource.

design/one-pager-easy-stacks.md
```diff
@@ -0,0 +1,294 @@
...
+this like VPC, sub-network, internet gateway etc. After creating these resources,
+you need to refer to them in your resource classes that you author for database
+and cluster claims to use.
+
+Today, it's possible to do all this by manually configuring each resource YAML
+files and run `kubectl apply -f <directory of YAMLs>`. However, there are two areas
+that we can improve during this process:
+* The preparation phase of these YAMLs requires to be very careful and changing
+  something usually means touching a few different files. While this may not seem
+  like a big hurdle, changing some high level parameters of that environment is
```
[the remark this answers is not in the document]
muvaf: Usually, it's a parameter that affects more than one resource. For example, region parameter that exist in many resources could be controlled on CR, controller would go ahead and make the changes on each resource. But it doesn't have to be that wide-affecting, it's basically up to the developer to decide what is high level/low level parameter.

design/one-pager-easy-stacks.md
muvaf: I was under the misconception that the stack `app.yaml` actually provided that ability. But I see that's not the case https://github.com/crossplaneio/crossplane/pull/1110#discussion_r352728755

So, I'll just remove that and expect people to have installed all the CRDs needed.

design/one-pager-easy-stacks.md
muvaf: This has to do with how we name the deployed resources. Basically, when you have only one CRD, you can just prefix the resources with the CR instance name. The api-server would reject creation of CR instances with the same name, so, name as prefix is enough. When you have multiple CRDs, name is not unique anymore. So, you either use `uid` or some other method for making deployed resources exclusive to one CR that intended to deploy them.

However, it's also possible that you can have multiple CRDs that deploy completely different kind sets of resources, you'd not need to worry about the name collision in that case. So, it's not that you should never do it but it's just highly recommended to go with one custom resource definition and build another stack for that extra CRD that you want.

design/one-pager-easy-stacks.md
muvaf: > What is readiness? How does the controller know they are ready?

Its details are in `Reconciliation` section.

> Do we get much value from modelling the fact that the entire configuration set is ready?

Thinking about it again, not a lot really. It'd help the infra-owner to know everything is ready but that'd be just informational. I'll remove that part.

muvaf: Oh, actually I think if we remove that there will be no `status`, meaning no way for anyone to know the stuff is actually working. So, I think we'd want to have at least the `conditions` part of the status with reconciliation status.

design/one-pager-easy-stacks.md
muvaf: @suskin I think the word _stack_ should be in it somewhere. I am leaning towards `Configuration Stacks`.

@negz I feel like `Config` and `Upbound` make the stack names too long. How about? `kind: AWSMinimal group: aws.configurationstacks.upbound.io`

design/one-pager-easy-stacks.md
muvaf: It can filter the resources with labels while reading the YAML files.

design/one-pager-easy-stacks.md
muvaf: It's not possible if the configuration stack CRD is namespace-scoped. But the direction is to go with making it cluster-scoped, so I will switch that to owner references, which is easier to implement and maintain.

design/one-pager-easy-stacks.md
muvaf: I believe 63 is the character limit. Since it's now cluster-scoped, we'll only use the CR's name and I think `uid` is longer than most of the CR names. So, it should be safer to go with name.  But, in general, yes if we hapen to have very long CR name and resource name, we'll run into a problem.

design/one-pager-easy-stacks.md
muvaf: Yes, that makes sense. Per discussion on https://github.com/crossplaneio/crossplane/pull/1110#discussion_r353461481 I think representing a condition for reconciliation status could be enough.

design/one-pager-easy-stacks.md
muvaf: I am not 100% sure about this one actually. Going as deep as having a folder for each `kind` looked too verbose to me. What do you think about having `resources/classes/<groupname>/<resource name>.yaml` ? For resources of different kinds with same names, the author probably would want to make the file name something like `<kind>-<object name>.yaml`

design/one-pager-easy-stacks.md
muvaf: Changed that section to be a bit more generic.

design/one-pager-configuration-stacks.md
```diff
@@ -70,12 +70,10 @@ configurations and be able to manage those configuration sets on a higher level.
...
 It is important that the design puts forward:
 
-* A base boilerplate package where creating a new configuration set that has a
-  controller is as easy as changing the YAML files.
-* A way to declare pre-requisites that needs to be satisfied by the cluster
-  for the configuration set to work.
-* A way to change a high level configuration by changing 1 value that is then
-  translated to the YAMLs that are already deployed.
+* A way for infrastructure owners to write their configuration sets with
+  reconciliation easier than writing a controller from scratch.
```
negz: It feels like we're really close to having something that allows infrastructure owners to write their configuration sets without writing _any_ controller. Should that be a goal?

design/one-pager-configuration-stacks.md
```diff
@@ -254,30 +244,35 @@ sense for developer to have a folder for each tier under `resources` folder.
...
+
+In the `Apply` phase, there will be some processing phase before calling `Create`
 for the resources:
 
-1. Names of the resources should be converted to the following format
-  `<CR namespace>-<CR name>-<Name given in YAML>` to prevent name collisions
-  since most of the resources stack deploys are cluster-scoped.
-  So, the name that appears on YAML files are valid for referencing each other
+1. Names of the resources should be converted to the processing phase following
+  format `<CR name>-<Name given in YAML>` to prevent name collisions.
```
negz: Is there something we can do to make it more obvious that this transformation will happen? I worry that it could be surprising.

It seems like the stack author will write something like:

```yaml
providerRef:
  name: cool-provider
```

But the resource we actually create will be modified to:

```yaml
providerRef:
  name: mycoolcustomresource-cool-provider
```

design/one-pager-configuration-stacks.md
```diff
@@ -288,6 +283,10 @@ for the resources:
     environment by just creating a CR instance instead of having everything in
     local.
   * Stacks do a better job for declaring CRD dependencies and UI annotation metadata.
+* Using kustomize or Go templates could save us from adding a Go function for each
+  kind we add even though this'd limit the future addition abilities it'd could
+  be preferrable.
```
negz: I have an inkling that using kustomize could be elegant here. Something like:

* User provides the base YAML for a bunch of resources.
* Our controller generates an overlay that sets resource names,  labels, and cross-resource reference.
* Our controller applies the generated overlay to the base layer and applies it to the API server.

This could allow folks to fork and define their own minimal/easy stacks simply by writing YAML (and stack scaffolding metadata). It seems like there's a bunch of potential overlap with template stacks here, but I don't think that needs to block this effort. It could be fine to write a purpose-built controller that uses Kustomize to do this specific task, then assess replacing it with template stacks if and when they're ready.

muvaf: My initial intent was to just use Kustomize for all transformations and let the controller be generic. However, I had gotten stuck while trying to do it, especially the part where you declare a reference to be filled by X kind that exists in the pack with that name.

I went through a few examples, dug deeper and finally found that there is a support for CRDs to declare custom reference fields. https://github.com/kubernetes-sigs/kustomize/blob/master/examples/transformerconfigs/crd/README.md

I have converted the minimal GCP YAML files into that structure and it works very well. We move the burden of the reference handling from the controller to the `kustomizeconfig.yaml` file. Note that we still need to add some lines to that file for each _kind_ that we add. But I guess it's better than adding a Go function.

In fact, we can have a giant `kustomizeconfig.yaml` file in `stack-gcp` that declares the dependencies for each type and fetch it, so, author will not worry about the reference bindings at all.

The controller will use the `kustomization.yaml` file that has `apiVersion: kustomize.config.k8s.io/v1beta1 kind: Kustomization` metadata. This introduces a dependency on kustomize but as far as I see, it's already pretty much the standard in the ecosystem for configuration management and templating. It's even in `kubectl`. So, it should be fine to depend on it IMO.

@prasek what do you think? The controller will be generic but the developer experience will change from dumping all yamls to dumping a kustomize directory and we'll depend on kustomize. Keep in mind that this is pretty much an implementation detail, nothing will change for clients that create the CR, i.e. Upbound backend.

muvaf: @negz There is one caveat: the custom transformer config identifies resources with only `name` and `kind`, it doesn't include `apiVersion` during its search. So, the resources that share the same `kind` but has different `apiVersion`, i.e. multicloud conf stacks, will be treated in the same way.

This would cause problem only if say the stack has one AWS Provider and GCP Provider that share the same name. For other resources, most of the time `kind` is different, like VPC is named `VPC` in AWS but `Network` in GCP. So, it should be fine except for `Provider` kinds which share the same kind across infra stacks.

Opened https://github.com/kubernetes-sigs/kustomize/issues/1885 to track this.

prasek: @muvaf using a kustomize directory sounds good and is aligned with the template stacks direction towards pluggable templating/config engines including kustomize support, so this direction makes sense.

design/one-pager-configuration-stacks.md
```diff
@@ -0,0 +1,282 @@
...
+```
+
+Then I wait for `Synced` condition to become `true`. After it's done, all resources
+are deployed.
+```yaml
+apiVersion: aws.configurationstacks.crossplane.io/v1alpha1
+kind: MinimalAWS
+metadata:
+  name: small-infra
+spec:
```
prasek: We'll also want to include spec attributes like region, projectId, and likely allow overriding defaults for things like CIDR / IP ranges for nodes, pods, services, etc. If we're using kustomize then this would likely generate a patch.yaml similar to what @hasheddan put together [here](https://github.com/crossplaneio/crosscd/blob/master/classpacks/workflow/argo/infra/gcp/patch.yaml).

design/one-pager-configuration-stacks.md
```diff
@@ -0,0 +1,282 @@
...
+Now, I want an environment where all my database instances and kubernetes clusters
+are connected to each other in a private VPC, which what this specific configuration
+stack does. Create the following:
+
+```yaml
+apiVersion: aws.configurationstacks.crossplane.io/v1alpha1
+kind: MinimalAWS
+metadata:
+  name: small-infra
+spec:
```
prasek: We'll want to support creating resource classes marked as default, but if more than one MinimalAWS CRs are created, there should be a way to disable the default label being applied so there is only 1 set of default classes.

design/one-pager-configuration-stacks.md
```diff
@@ -0,0 +1,282 @@
...
+apiVersion: database.aws.crossplane.io/v1beta1
+kind: RDSInstanceClass
+metadata:
+  name: small-infra-mysql
+  ownerReferences:
+  - apiVersion: aws.configurationstacks.crossplane.io/v1alpha1
+    kind: MinimalAWS
+    name: small-infra
+    uid: 1ca16960-973b-4d58-a6fa-5696638ec631
+  labels:
```
prasek: Will the `MinimalAWS` CR labels be copied down to the templated resources? Suggest adding an example of this.

design/one-pager-easy-stacks.md
suskin: I agree with the point @negz is making about it being helpful if we are more specific about which end-user persona we're targeting. I'm making a note of this because it appears that the goal is not yet written in that way.

design/one-pager-easy-stacks.md
suskin: It does seem valuable to have a way to check the status of everything in a single place. However, I am not sure we need to make it a requirement for resource packs. Since the main goal is the focus on applying a set of configuration best practices without writing your own, that goal seems somewhat orthogonal to the goal of seeing the statuses in one place.

Put another way, I think it is an improvement even without the status aggregation, so we don't necessarily need to include the status aggregation in the first version. For example, it is possible that we add status/readiness aggregation in some other way, with a UI or with some `trace`-like functionality.

design/one-pager-easy-stacks.md
suskin: > It cautions against branding . . . as making those things "easy"

I agree with @negz wholeheartedly on this. I have a strong dislike of the name "Easy", and the presentation Nic linked explains the reasons better than I was able to articulate them.

> or something that emphasises a little more specifically the value they bring

I see the general name (e.g. Resource Pack, Configuration Stack) as the name for the style of stack, and I see something like the proposed "Base Config" as a more specific term to describe the value that a specific pack/stack is providing. I see them as complementary, and not exclusive. I see it as a corollary to how a `GKECluster` resource is a GKECluster, and is a managed resource, and is a resource.

> I am leaning towards `Configuration Stacks`

I slightly prefer "pack" over "stack" because "stack" just happens to be the format that we're using to bundle it, whereas "pack" describes what it is conceptually, without tying it to a particular format. That makes it more accessible for people who are not yet familiar with our idioms and terms. My **favorite** is "resource pack", but I also like "configuration pack". I slightly prefer "resource" because it is more literal about what is in the pack (a bunch of resources), whereas "configuration" could be interpreted as referring to other types of configuration.

design/one-pager-easy-stacks.md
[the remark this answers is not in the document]
muvaf: >  they do not exist in the app.yaml file

@displague I guess the design doc is not updated with the different implementation, right?

design/one-pager-easy-stacks.md
muvaf: @suskin I have added `ConditionedStatus` to status of the initial gcp implementation. We'll have `Synced` type condition to be set by the generic controller.

muvaf:
@negz https://github.com/muvaf/configuration-stacks/pull/1 is the initial implementation of the generic reconciler and https://github.com/muvaf/minimal-gcp/pull/1 is the first integration. It's unpolished, but I'd love your general feedback about the direction in generic reconciler.

@prasek I have added your comments as TODOs in the implementation PRs. I'll update the one-pager as soon as I can.

@suskin @prasek @negz I guess I agree with @suskin on the name `Resource Packs`. What do you guys think? Note that this will be visible to users both in the implementation as well as documentation.

design/one-pager-easy-stacks.md
suskin: I don't think we need to enforce any particular structure. I would provide some examples of what we feel are best practices for logical groupings, but keep the tooling unopinionated and leave it up to the author to decide what works well for them. If it's supported by the tool they're using underneath (in this case, kustomize), then we should support it. The grouping that we're using in the example structure looks pretty reasonable to me.

design/one-pager-configuration-stacks.md
muvaf: I have added a section about this. I suggest we have a field named `keepDefaultingLabels`(name suggestions are welcome) and if user intends to keep the defaulting label in resource classes, they have to make this field `true`. Otherwise, the controller will remove the defaulting labels whenever it sees one. So, by default, if several users deploy empty CRs they won't have any conflict. If they want the default, they'd need to explicitly say so.

design/one-pager-configuration-stacks.md
muvaf: Yes, I think that's a good UX improvement for users who use selectors in claims.

design/one-pager-configuration-stacks.md
muvaf: The `patch.yaml` approach requires the types to be known by the controller and iteration over each and every one of them, which is not ideal IMO since the developer will have to implement more complex functions.

I've experimented with [transformer configs of Kustomize for variant referencing](https://github.com/kubernetes-sigs/kustomize/blob/master/examples/transformerconfigs/crd/README.md#adding-a-custom-resource) and I think it works for us. An example implementation is here https://github.com/muvaf/minimal-gcp/pull/2 . As you'll see, we define the region as `$(REGION)` in the YAML files and then in the controller code, only [define the path in the CR](https://github.com/muvaf/minimal-gcp/pull/2/files#diff-1fa7bf35a53cd97938369f8733221602R70) that the value should be taken.

design/one-pager-easy-stacks.md
muvaf: > Do you expect there will be one generic controller loading this YAML as *unstructured.Unstructured, or a different controller for each stack variant that knows the actual kinds it supports at the code level?

Became unrelated as we started to use `kustomize`. But in patchers, we do make the operations on `*unstructured.Unstructured` type.

> will need to know how each resource reflects it readiness.

No readiness anymore. If the reconciliation is completed with no errors, it means the resources are deployed and we tell that to user via `Synced` condition.

design/one-pager-easy-stacks.md
```diff
@@ -0,0 +1,294 @@
...
+
+## Goals
+
+The main goal of this design is to make it easier for people to write sets of
+configurations and be able to manage those configuration sets on a higher level.
+
+It is important that the design puts forward:
+
+* A base boilerplate package where creating a new configuration set that has a
+  controller is as easy as changing the YAML files.
```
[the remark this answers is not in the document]
muvaf: Not exactly _forking_. The developer has to build the stack from ground up since they will implement their own `CRD`. I mean they can still fork the example stacks but then they'd have to use the same `CRD`.

Currently, we provide a generic reconciler that they can fetch with custom modification functions that applies their custom logics in different steps. How does the following statement sound?

```
A base tooling for infrastructure operators who would like to package a set of resources as a stack and provide that to their users with minimal Go code.
```

muvaf:
I have changed the name to `resource packs`. The PR https://github.com/muvaf/minimal-gcp/pull/2 is our first resource pack for minimal GCP. It has region and reference to cred secret as spec. Tested and it works as expected.

I think the major new thing is how we bind a field from the CR to the child resources that affects the developer experience. Other than that I will put comments in the code and write tests for the generic reconciler.

@prasek @negz @suskin IMO, both generic reconciler and minimal GCP looks like complete feature-wise for an initial implementation. Please let me know if you have suggestions that affect the user experience directly, otherwise I'd like to deliver that implementation as initial step and iterate as we go on.

design/one-pager-configuration-stacks.md
```diff
@@ -105,53 +106,77 @@ rules to be adhered by initial configuration stacks.
...
   deployed by the stack.
-* CRD of the configuration stack should be cluster-scoped since the author of
-  the CR instances of the configuration stack is assumed to be the infrastructure
+* CRD of the resource pack should be cluster-scoped since the author of
+  the CR instances of the resource pack is assumed to be the infrastructure
   owner.
 * Stack type will be `ClusterStack` as most of the resources it deploys are
   cluster-scoped.
+* CRD of the resource pack should have a boolean spec field called
+  `keepDefaultingLabels` which indicates what happens to the resource classes that
```
negz: Should this be annotations?

design/one-pager-configuration-stacks.md
```diff
@@ -256,16 +292,130 @@ like the following:
...
+new overlay with its own `kustomization.yaml` in a temporary directory and refers
+to the `resources` folder.
+
 [UI annotations] should be present to make it easy for frontend software to process
 the stack.
- 
+
+### Custom Patchers
+
+The resource pack reconciler that has two types of patcher functions where
```
negz: I think the "that" on this line is redundant.

design/one-pager-configuration-stacks.md
muvaf: Yes, updated it.

muvaf:
@negz I have updated this one-pager with the latest thinking. I think we're ready to merge this. The implementations are completed feature-wise, so I'd appreciate your review on:
* https://github.com/muvaf/crossplane-resourcepacks/pull/1 Generic reconciler
* https://github.com/muvaf/minimal-gcp/pull/2 Minimal GCP stack

After the merge I think we'll need to move the repositories to `crossplaneio` org.

As we synced with @suskin and @prasek, at some point, we're going to provide easy stacks via template stacks when that work lands. I added the code generation approach under the future considerations because it seems we'll not go in that direction. Though IMO, it'd have been really cool to have the `resources` folder and a YAML with descriptions about CRD and patchers and then just do `crossplane stack build . -t minimal-gcp` && `crossplane push minimal-gcp` with no Go code seen by the user.

design/one-pager-resource-packs.md
```diff
@@ -0,0 +1,508 @@
...
+  * See details about [defaulting mechanism here].
+* All the labels in the resource pack CR is propagated down to all resources
+  that it deploys.
+
+## User Experience
+
+1. Install GCP stack and create a secret that contains GCP credentials.
+2. Install a GCP resource pack via:
+```yaml
+# exact apiVersion is TBD.
```
negz: Is this actually still TBD?

PR #1117, opened 2019-12-06
managed resource api design: add exceptions and edge cases section

hasheddan:

A few updates based on work porting resources to `v1beta1` quality.

n/a

[skip ci]

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/design/one-pager-managed-resource-api-design.md
+++ b/design/one-pager-managed-resource-api-design.md
@@ -1,7 +1,13 @@
 # Managed Resources API Patterns
 * Owner: Muvaffak Onus (@muvaf)
 * Reviewers: Crossplane Maintainers
-* Status: Accepted, Revision 1.0
+* Status: Accepted, Revision 1.1
+
+## Revisions
+
+* 1.1
+  * Added [Exceptions & Edge Cases](#exceptions-&-edge-cases) section
+  * Updated examples with cluster-scoped managed resource and `Provider` objects
 
 ## Terminology
 
@@ -77,7 +83,6 @@ apiVersion: database.gcp.crossplane.io/v1alpha2
 kind: CloudSQLInstance
 metadata:
   generateName: myappnamespace-mydatabase-
-  namespace: gcp
 ```
 After Kubernetes API server completes the creation, we'll see something like the following:
 ```yaml
@@ -85,7 +90,6 @@ apiVersion: database.gcp.crossplane.io/v1alpha2
 kind: CloudSQLInstance
 metadata:
   name: myappnamespace-mydatabase-5sc8a
-  namespace: gcp
 ```
 `mydatabase` is the name of the `Claim` and `myappnamespace` is the namespace that `Claim` lives in. That way we get a
 managed resource that user can relate to what they actually created.
@@ -199,13 +203,11 @@ apiVersion: compute.gcp.crossplane.io/v1alpha1
 kind: Network
 metadata:\
   name: my-legacy-network
-  namespace: crossplane-system
-  selfLink: /apis/compute.gcp.crossplane.io/v1alpha1/namespaces/crossplane-system/networks/my-legacy-network
+  selfLink: /apis/compute.gcp.crossplane.io/v1alpha1/networks/my-legacy-network
 spec:
   name: monus-legacy-network-res
   providerRef:
     name: monus-gcp
-    namespace: crossplane-system
   reclaimPolicy: Delete
   routingConfig:
     routingMode: REGIONAL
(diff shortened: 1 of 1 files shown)


design/one-pager-managed-resource-api-design.md
```diff
@@ -379,6 +379,149 @@ type SubnetworkParameters struct {
 }
 ```
 
+### Exceptions and Edge Cases
```
muvaf: I think people reading this document are unlikely to go check whether there is a section about exceptions. If we have the two sections here in their corresponding sections as an addition, they'll more likely be picked up by the readers.

design/one-pager-managed-resource-api-design.md
```diff
@@ -379,6 +379,149 @@ type SubnetworkParameters struct {
...
+
+Because managed resources could represent API types for any provider, there is
+potential to encounter scenarios that fall outside the rules expressed in this
+design document. Some common exceptions and how to address them are expressed
+below.
+
+#### Embedded Structs with Mixed Fields
+
+Some provider APIs include an embedded struct that may contain some fields that
```
muvaf: I think the section `High Fidelity` is the most relevant one, so, maybe we could move this up there as an example case.

design/one-pager-managed-resource-api-design.md
```diff
@@ -379,6 +379,149 @@ type SubnetworkParameters struct {
...
+// Only the non-configurable fields in EmbeddedStruct
+type EmbeddedStructStatus struct {
+  // This field is not configurable so it should be in status.atProvider
+  SomeNonConfigurableField string `json:"someNonConfigurableField,omitempty"`
+}
+```
+
+#### Optional Embedded Structs with One Field
+
+Some provider APIs include an embedded struct that is optional and contains only
```
muvaf: This feels like it could belong to section `Pointer Types and Markers`

design/one-pager-managed-resource-api-design.md
```diff
@@ -169,6 +173,79 @@ For both `Status` and `Spec`:
...
+  // This field is configurable so it should be in spec.forProvider
+  AnotherConfigurableField *string `json:"anotherConfigurableField,omitempty"`
+
+  // This field is not configurable so it should be in status.atProvider
+  SomeNonConfigurableField string `json:"someNonConfigurableField,omitempty"`
+}
+```
+
+In this case, the solution is to divide the embedded struct into
+`EmbeddedStructSpec` and `EmbeddedStructStatus`.
```
negz: Would `<T>` and `<T>Observation` be more in line with our existing patterns? For example https://github.com/crossplaneio/stack-gcp/blob/d45813c/apis/database/v1beta1/cloudsql_instance_types.go#L92

design/one-pager-managed-resource-api-design.md
```diff
@@ -330,6 +403,69 @@ type GlobalAddressParameters struct {
...
+spec:
+  providerRef:
+    name: provider-gcp
+  reclaimPolicy: Delete
+  writeConnectionSecretToRef: {}
+  forProvider:
+    embeddedStructOne: # valid if we do not require embedded field
+    fieldOne: my-cool-field
+```
+
```
negz: Do we need to consider the case in which additional optional fields are added over time to the cloud provider struct?

I could imagine a case in which:

1. We mark field A, which is optional to the cloud provider, required.
1. The cloud provider adds field B, which is also optional.
1. Our users now cannot specify field B without also specifying field A.

My inclination would be to avoid overriding the cloud provider here, even if it does mean certain nonsense configurations such as the above become valid.

design/one-pager-managed-resource-api-design.md
hasheddan: I could go either way. @muvaf suggested the `___Spec` and `___Status` pattern here, but I had also thought of the `Observed___` pattern in PR body [here](https://github.com/crossplaneio/stack-gcp/pull/111). I agree that `___Observation` more clearly matches the encompassing struct.

hasheddan: I am actually realizing the `Spec` / `Status` pattern has already been implemented. It could be updated without modifying the API though.

https://github.com/crossplaneio/stack-gcp/blob/f7522e08bccbf856cb1cd3ad713efe597f45785b/apis/database/v1beta1/cloudsql_instance_types.go#L168

Thoughts?

design/one-pager-managed-resource-api-design.md
hasheddan: Hmm, in that case our API would have to be updated to include that additional field, and at that point we would no longer have the situation described here. In addition, I think we have already implemented in this manner as `omitempty` is removed in our API where it is present in GCP: https://github.com/crossplaneio/stack-gcp/blob/f7522e08bccbf856cb1cd3ad713efe597f45785b/apis/database/v1beta1/cloudsql_instance_types.go#L569

Changing this *would* require an update to the API.

design/one-pager-managed-resource-api-design.md
negz: I'm happy with spec and status as long as it doesn't go against the status quo.

design/one-pager-managed-resource-api-design.md
negz: I think this update would require us to make a required field optional, which should be a backward compatible change. I think I was getting mixed up and thinking we'd have to make an optional field required and thus break API compatibility. This seems fine to me, though TBH I'm not super concerned about the problem it's solving.

PR #1118, opened 2019-12-09
ResourceClaimDefinition design sketch

### LINKED ISSUE #1105: Allow a claim to be satisfied by multiple resources

One managed resource is frequently not useful enough to truly satisfy a resource claim.

Crossplane uses a class and claim model to provision and manage resources in an external system, such as a cloud provider. _External resources_ in the provider's API are modelled as _managed resources_ in the Kubernetes API server. Managed resources are considered the domain of _infrastructure operators_; they're cluster scoped infrastructure like a `Node` or `PersistentVolume`. _Application operators_ may claim a managed resource for a particular purpose by creating a namespaced _resource claim_. Managed resources may be provisioned explicitly before claim time (static provisioning), or automatically at claim time (dynamic provisioning). The initial configuration of dynamically provisioned managed resources is specified by a _resource class_.

A managed resource is a _high-fidelity_ representation of its corresponding external resource. High-fidelity in this context means two things:

* A managed resource maps to exactly one external resource - one API object.
* A managed resource is as close to a direct translation of its corresponding external API object as is possible without violating Kubernetes API conventions.

These properties make managed resources - Crossplane's lowest level infrastructure primitive - flexible and self documenting. Managed resources in and of themselves hold few opinions about _how_ they should be used, and are easily related back to the APIs they represent. This provides a solid foundation upon which to build Crossplane's multicloud capability.

_Application operators_ are typically prevented by RBAC from creating and modifying managed resources directly; they are instead expected to dynamically provision the managed resources they require by submitting a resource claim. Crossplane provides claim kinds for common, widely supported resource variants like `MySQLInstance` and `KubernetesCluster`. There is a one-to-one relationship between claims and the managed resources they bind to; a `KubernetesCluster` claim binds to exactly one `GKECluster` managed resource. However, a solitary resource is often not particularly useful without supporting infrastructure, for example:

* An RDS instance may be inaccessible without a security group.
* An Azure SQL instance may be inaccessible without a virtual network rule.
* A GKE, EKS, or AKS cluster (control plane) may not be able to run pods without a node group.

Crossplane stacks frequently model this supporting infrastructure (there is a`SecurityGroup` managed resource, for example) but it cannot be dynamically provisioned or bound to a resource claim. Instead a cluster operator must statically provision any supporting managed resources ahead of time, then author resource classes that reference them. This can be limiting:

* Often supporting resources must reference the managed resource they support, for example an Azure `MySQLServerVirtualNetworkRule` must reference the `MySQLServer` it applies to. Dynamically provisioned managed resources such as a `MySQLServer` have non-deterministic names, making it impossible to create a `MySQLServerVirtualNetworkRule` until the `MySQLServer` it must reference has been provisioned.
* When a resource class references a statically provisioned managed resource every managed resource that is dynamically provisioned using that class will reference that specific managed resource. For example if a `GKEClusterClass` references a `Subnetwork` then every `GKECluster` dynamically provisioned using said class will attempt to share said `Subnetwork`, despite it often being desirable to create a unique `Subnetwork` for each dynamically provisioned `GKECluster`.

The one-to-one relationship between resource claims and resource classes thus weakens portability, separation of concerns, and support for [GitOps]. An infrastructure operator can publish a resource class representing a single managed resource that an application operator may dynamically provision, but in the likely event that managed resource requires supporting managed resources to function usefully the application operator must ask an infrastructure operator to provision them.

Crossplane could provide some mechanism such that one resource claim could be satisfied by multiple managed resources.

This has some further possible implications under the current class and claim model, i.e.:

* A resource class must be able to specify that, and how, multiple managed resources should be dynamically provisioned to satisfy a resource claim.
* A resource claim must be satisfiable by binding to a group of statically provisioned managed resources.

### LINKED ISSUE #1106: Allow infrastructure operators to define claim kinds

In short, Crossplane's built in set of portable resource claims attempt to be applicable to all possible cloud providers, and this greatly reduces their potential to express configuration options.

Crossplane defines a handful of _resource claim_ kinds whose CRDs are installed by Helm along with the "core" Crossplane controllers. Infrastructure stacks define _managed resources_ that may satisfy these claims by binding to them. Most extant resource claims are _portable_ in that they can be satisfied by the managed resources of different providers. For example a `MySQLInstance` claim could be satisfied by a GCP `CloudSQLInstance` or an AWS `RDSInstance`. Claims may trigger the dynamic provisioning of the managed resource that will satisfy them. The managed resource is configured using the claim and a _resource class_ as inputs.

Portable resource claims are subject to the "lowest common denominator" problem of multi-cloud; when a claim may provide configuration inputs that may be used to either match or dynamically provision many kinds of managed resource it may support only the settings that apply to all managed resources. This is in part why Crossplane currently supports relatively few resource claim kinds. Meanwhile, it's possible that a particular infrastructure operator deploys Crossplane to provide an opinionated abstraction for the application operators in their organisation and that their organisation _only uses AWS_. This organisation values Crossplane's _separation of concern_ but does not need its portability. Why should this organisation's application operators be limited to submitting resource claims that could be satisfied by any cloud provider when they could expose more settings, knowing that their claims would always be satisfied by an RDS instance? Perhaps the organisation uses AWS and Azure, but not GCP. Why should their resource claims be constrained to features supported by GCP?

Crossplane could defer the task of defining resource claims to infrastructure operators; each organisation could define (and/or install via a stack) the set of resource claims most useful to them. This would include defining the schema of their new resource claims, as well as which managed resources those resource claims may bind to, and how they may be used to dynamically provision those managed resources.

### LINKED ISSUE #1107: Surface claim to resource compatibility in Crossplane's API

Crossplane supports a set of resource claims that may be satisfied by binding to a managed resource. Resource claims may also trigger the dynamic provisioning of a managed resource, which is configured using the resource claim and a resource class as inputs. There is a many-to-many compatibility relationship between resource claims and managed resources, e.g.:

|                    | CloudSQLInstance | RDSInstance | GKECluster |
|--------------------|------------------|-------------|------------|
| **MySQLInstance**      | Compatible       | Compatible  | -          |
| **PostgreSQLInstance** | Compatible       | Compatible  | -          |
| **KubernetesCluster**  | -                | -           | Compatible |

However, this compatibility relationship is defined _only in code_. Each infrastructure stack instantiates a series of claim controllers (for binding, scheduling, and defaulting), each handling the combination of a particular managed resource and resource claim.

This means that there is no obvious way to discover (or as this issue relates to https://github.com/crossplaneio/crossplane/issues/1105, _define_) which managed classes of managed resource may satisfy a particular kind of claim.

At a minimum, Crossplane should make it easy for application owners to discover which classes of managed resource might satisfy a particular kind of claim. This issue is somewhat coupled to https://github.com/crossplaneio/crossplane/issues/1105, in which it would also be useful for infrastructure operators to _define_ which claims a managed resource can satisfy.

negz:

Fixes #1105 
Fixes #1106 
Fixes #1107 

This is a rough sketch for the concept of a `ResourceClaimDefinition`, which would allow infrastructure operators to define new resource claims, their schemas, and their relationships to the managed resources that could satisfy them, all at the API level without writing code.

I'm hoping the API design and notes speak for themselves enough for now, but I'd like to flesh this out into a proper design doc once I've recovered from the flu.

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/design/design-doc-resource-claim-definition.md
+++ b/design/design-doc-resource-claim-definition.md
@@ -1,0 +1,174 @@
+# Resource Claim and Binding Definition
+* Owner: Nic Cope (@negz)
+* Reviewers: Crossplane Maintainers
+* Status: Draft
+
+## Background
+
+See the following issues this design doc intends to address:
+
+* https://github.com/crossplaneio/crossplane/issues/1105
+* https://github.com/crossplaneio/crossplane/issues/1106
+* https://github.com/crossplaneio/crossplane/issues/1107
+
+TODO(negz): Write this up.
+
+## Goals
+
+* Allow infrastructure operators to define claim kinds.
+* Allow a claim to be satisfied by multiple resources.
+* Surface claim to resource compatibility in Crossplane's API.
+
+TODO(negz): Write this up.
+
+## Proposal
+
+This document proposes the introduction of two new, cluster scoped, resources
+named `ResourceClaimDefinition` and `ResourceBindingDefinition`.
+
+```yaml
+apiVersion: core.crossplane.io/v1alpha1
+kind: ResourceClaimDefinition
+metadata:
+  name: mysqlinstance
+spec:
+  group: database.crossplane.io
+  names:
+    kind: MySQLInstance
+    listKind: MySQLInstanceList
+    plural: mysqlinstances
+    singular: mysqlinstance
+  version: v1alpha1
+  validation:
+    openAPIV3Schema:
+      description: A MySQLInstance is a portable resource claim that may be satisfied
+        by binding to a MySQL managed resource such as an AWS RDS instance or a GCP
+        CloudSQL instance.
+      properties:
+        spec:
+          properties:
+            engineVersion:
+              description: EngineVersion specifies the desired MySQL engine version,
+                e.g. 5.7.
+              enum:
+              - "5.6"
+              - "5.7"
+              type: string
+```
+
+```yaml
+---
+apiVersion: core.crossplane.io/v1alpha1
+kind: ResourceBindingDefinition
+metadata:
+  name: mysqlinstance-cloudsqlinstance
+spec:
+  claim:
+    apiVersion: database.crossplane.io/v1alpha1
+    kind: MySQLInstance
+  resources:
+  - resource:
+      apiVersion: database.gcp.crossplane.io/v1beta1
+      kind: CloudSQLInstance
+    class:
+      apiVersion: database.gcp.crossplane.io/v1beta1
+      kind: CloudSQLInstanceClass
+      strategies:
+        provisioning:
+        - overlay
+        - translate
+        overlay:
+        - fromClassField: ".specTemplate"
+          toResourceField: ".spec"
+        translate:
+        - fromClaimField: ".spec.engineVersion"
+          toResourceField: ".spec.forProvider.databaseVersion"
+          transforms:
+            toupper: true
+            replace:
+              string: "."
+              with: "_"
+---
+apiVersion: core.crossplane.io/v1alpha1
+kind: ResourceBindingDefinition
+metadata:
+  name: mysqlinstance-cloudsqlinstance
+spec:
+  claim:
+    apiVersion: database.crossplane.io/v1alpha1
+    kind: MySQLInstance
+  resources:
+  - resource:
+      apiVersion: database.azure.crossplane.io/v1beta1
+      kind: MySQLServer
+    class:
+      apiVersion: database.azure.crossplane.io/v1beta1
+      kind: SQLServerClass
+      strategies:
+        selection:
+        - match
+        match:
+        - fromClaimField: ".spec.engineVersion"
+          toClassField: ".specTemplate.forProvider.version"
+          transforms:
+            toupper: true
+  - resource:
+      apiVersion: database.azure.crossplane.io/v1alpha3
+      kind: MySQLServerVirtualNetworkRule
+    class:
+      apiVersion: database.azure.crossplane.io/v1alpha3
+      kind: MySQLServerVirtualNetworkRuleClass
+```
+
+`ResourceClaimDefinition` is an abstraction that provides guard rails and
+opinions around the process of authoring a `CustomResourceDefinition` that is
+intended for use as a resource claim. Crossplane resource claims are of many
+different kinds, but all have certain things in common. These include:
+
+* Declaring use of the status subresource.
+* Supporting standard Kubernetes object and type metadata.
+* Supporting a binding phase and conditioned status in their `.status`.
+* Supporting a class selector, class reference, resource reference, and
+  connection secret reference in their `.spec`.
+* Supporting a standard set of kubectl printer columns.
+
+A `ResourceClaimDefinition` allows an infrastructure operator to define a
+resource claim without having to specify all of these things correctly. They can
+instead focus on what they're concerned with - defining the purpose specific
+`.spec` fields of their claim. A `ResourceClaimDefinition` controller will watch
+for `ResourceClaimDefinition` instances, filling in the blanks to create a
+corresponding `CustomResourceDefinition`.
+
+`ResourceBindingDefinition` maps a kind of resource claim to the kinds of
+managed resource (and class) that may satisfy it. In addition to declaring the
+relationship between a claim kind and one or more managed resource kinds, a
+`ResourceBindingDefinition` defines the _strategies_ that should be used when
+dynamically provisioning a set of managed resources to satisfy a claim. There
+are two types of strategy:
+
+* _Selection strategies_ constrain which resource classes can satisfy a
+  resource claim, for example by requiring that a particular claim field match
+  a particular resource class field, perhaps after a transform is applied.
+* _Provisioning strategies_ configure how dynamic provisioning is handled, i.e.
+  how to form a managed resource from a claim and a class. One example might be
+  using the resource class's `.specTemplate` as an overlay on the managed
+  resource's `.spec`.
+
+Random notes:
+
+* It's possible to use `ResourceBindingDefinition` without using
+  `ResourceClaimDefinition`.
+* Creating a `ResourceBindingDefinition` spins up claim scheduling, defaulting,
+  binding, etc controllers for the resource claim kind it is concerned with. I'm
+  about 80% sure it's possible to build controllers using this RCD that operate
+  on unstructured.Unstructured.
+* This proposal implies a claim can bind to more than one managed resource, and
+  can thus reference more than one class for dynamic provisioning. We know how
+  many resources must be bound for the claim to be ready thanks to the RCD.
+* Dynamic provisioning would occur in the order the class kinds were specified
+  in the schema. Classes may use affinity / anti-affinity labels so that e.g. a
+  compatible vnet rule class will be used for the chosen MySQLServer class.
+* If a resource claim needs three resources of a particular kind to be satisfied
+  you must repeat that resource (and its class) three times in its resource
+  group. We could instead add a 'count' field but that would require three
+  identically configured resources are dynamically provisioned.


design/design-doc-resource-claim-definition.md
```diff
@@ -0,0 +1,158 @@
...
+  # from different providers (e.g. a GKE cluster and an S3 bucket) if that was
+  # somehow useful to you.
+  satisfaction:
+  # GCP is a resource group that can satisfy this claim with a single resource.
+  - name: GCP
+    description: A MySQLInstance can be satisfied by a CloudSQLInstance.
+    resources:
+    - resource:
+        apiVersion: database.gcp.crossplane.io/v1beta1
+        kind: CloudSQLInstance
```
negz: Note that the CRD for these resources and classes are still published by their infra stacks as they are today, so this is a reference to an existing type, not a schema for one.

Also note there's no concept of "composite" managed resources in this design. I think the line between an infrastructure operator defined resource claim and an infrastructure operator defined managed resource becomes too blurry, so instead I prefer to allow a claim to reference more than one class and bind to more than one managed resource. We know how many and what specific kind of classes/resources are needed thanks to the RCD schema.

suskin:
We've been experimenting with assigning the people we're expecting to take action next, so I've marked all of the reviewers as assignees. Using the reviewers field isn't quite expressive enough, because it's hard to model (for example) when the author should respond to some feedback, or when a second round of reviews from the reviewers is requested. After I review, I'll remove myself as an assignee, and I'll add Nic as an assignee.

design/design-doc-resource-claim-definition.md
```diff
@@ -0,0 +1,158 @@
...
+provision those resources all without needing to write any code.
+
+```yaml
+apiVersion: core.crossplane.io/v1alpha1
+kind: ResourceClaimDefinition
+metadata:
+  name: mysqlinstance
+spec:
+  # A controller watching for ResourceClaimDefinition uses the group, names,
+  # version, and validation fields to create a CustomResourceDefinition.
```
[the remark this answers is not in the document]
negz: @bassam had similar feedback. I was using a similar approach at first in which this resource would reference a claim CRD rather than define it. I ended up merging them because I feel (quite strongly) that we need a layer of abstraction atop "write a CRD freehand" for claim authors, so I put that in here.

However I could see it being more flexible if we took something like this and broke it up into two kinds, e.g.:

* `ResourceClaimDefinition` - Makes it easy to write compliant and sensible resource claim CRDs. Basically a subset of `CustomResourceDefinition`, each produces a CRD.
* `ResourceClaimSatsifaction` - Basically everything covered by a "resource group" in this sketch.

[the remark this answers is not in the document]
negz: I've pushed a change that breaks this into two kinds.

design/design-doc-resource-claim-definition.md
```diff
@@ -0,0 +1,158 @@
...
+  - name: GCP
+    description: A MySQLInstance can be satisfied by a CloudSQLInstance.
+    resources:
+    - resource:
+        apiVersion: database.gcp.crossplane.io/v1beta1
+        kind: CloudSQLInstance
+      class:
+        apiVersion: database.gcp.crossplane.io/v1beta1
+        kind: CloudSQLInstanceClass
+        strategies:
```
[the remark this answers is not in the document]
negz: > Have you thought at all yet about how to accommodate more strategies? 

I imagine a strategy to be two things:

1. Fields on a `ResourceClaimSatisfaction` CR that configure a particular strategy.
1. Code in the `ResourceClaimSatisfaction` controller(s) that enact the strategy.

So therefore accommodating new strategies would involve adding fields, and releasing new controller code.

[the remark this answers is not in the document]
negz: > My guess would be that someone at some point will probably want a complex strategy, and may want to use a familiar resource engine (e.g. helm, kustomize, kapitan).

I'm not convinced that such tools are a good fit (or bring much value) for the dynamic provisioning use case, so my thinking is we should focus on supporting a set of targeted, useful strategies rather than attempting to deliver an open ended template-stacks-like experience in which you can bring whatever templating engine you're familiar with.

While dynamic provisioning is template-ish (we take a claims and classes as inputs, and produce managed resources as outputs) there are also a lot of constraints on the process. We require claims, classes, and managed resources to have certain shapes, act certain ways, and support certain patterns. We probably only want to use the class `.specTemplate` and a subset of the claim `.spec.` as inputs, and only produce a subset of the managed resource's `.spec` as an output, leaving the rest of the object's values (name, external name annotation, binding phase, references, etc) up to the opinions of the dynamic provisioning controller.

design/design-doc-resource-claim-definition.md
```diff
@@ -0,0 +1,158 @@
...
+            toClassField: ".specTemplate.forProvider.version"
+            transforms:
+            - toupper: {}
+
+    - resource:
+        apiVersion: database.azure.crossplane.io/v1alpha3
+        kind: MySQLServerVirtualNetworkRule
+      class:
+        apiVersion: database.azure.crossplane.io/v1alpha3
+        kind: MySQLServerVirtualNetworkRuleClass
```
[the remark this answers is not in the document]
negz: My thinking was that if no parameters were supplied we'd use some sane default, such as an `overlay` of the class's `.specTemplate` field onto the managed resource's `.spec` field.

design/design-doc-resource-claim-definition.md
```diff
@@ -0,0 +1,174 @@
...
+
+`ResourceClaimDefinition` is an abstraction that provides guard rails and
+opinions around the process of authoring a `CustomResourceDefinition` that is
+intended for use as a resource claim. Crossplane resource claims are of many
+different kinds, but all have certain things in common. These include:
+
+* Declaring use of the status subresource.
+* Supporting standard Kubernetes object and type metadata.
+* Supporting a binding phase and conditioned status in their `.status`.
+* Supporting a class selector, class reference, resource reference, and
```
muvaf: All these reference interfaces would need to change to accomodate more than one references. It could be helpful to demonstrate how you'd tackle this.

design/design-doc-resource-claim-definition.md
```diff
@@ -0,0 +1,174 @@
...
+* Supporting a standard set of kubectl printer columns.
+
+A `ResourceClaimDefinition` allows an infrastructure operator to define a
+resource claim without having to specify all of these things correctly. They can
+instead focus on what they're concerned with - defining the purpose specific
+`.spec` fields of their claim. A `ResourceClaimDefinition` controller will watch
+for `ResourceClaimDefinition` instances, filling in the blanks to create a
+corresponding `CustomResourceDefinition`.
+
+`ResourceBindingDefinition` maps a kind of resource claim to the kinds of
```
muvaf: When an `XClaim` is created, the only selection we normally do is deciding which `XClass` to refer. For generic `ResourceClaim`, we need to have a mechanism to select which `ResourceBindingDefinition` to use. It'd be helpful to mention how you approach that problem.

design/design-doc-resource-claim-definition.md
```diff
@@ -0,0 +1,174 @@
...
+              enum:
+              - "5.6"
+              - "5.7"
+              type: string
+```
+
+```yaml
+---
+apiVersion: core.crossplane.io/v1alpha1
+kind: ResourceBindingDefinition
```
muvaf: Could you elaborate the benefits of not embedding `ResourceBindingDefinition`s into the `ResourceClaimDefinition`? I feel like having only `ResourceClaimDefinition` type that also defines the bindings is more concise and gives the user the comfort of checking only one place to see what the possible outcomes if they were to create a `ResourceClaim`.

design/design-doc-resource-claim-definition.md
```diff
@@ -0,0 +1,174 @@
...
+  resource claim, for example by requiring that a particular claim field match
+  a particular resource class field, perhaps after a transform is applied.
+* _Provisioning strategies_ configure how dynamic provisioning is handled, i.e.
+  how to form a managed resource from a claim and a class. One example might be
+  using the resource class's `.specTemplate` as an overlay on the managed
+  resource's `.spec`.
+
+Random notes:
+
+* It's possible to use `ResourceBindingDefinition` without using
```
muvaf: I'm not sure whether this is a good or bad thing. Ability to disturb what `KubernetesCluster` claim does via another resource seems like an open door for surprises for the users of the claims.

design/design-doc-resource-claim-definition.md
```diff
@@ -0,0 +1,174 @@
...
+apiVersion: core.crossplane.io/v1alpha1
+kind: ResourceBindingDefinition
+metadata:
+  name: mysqlinstance-cloudsqlinstance
+spec:
+  claim:
+    apiVersion: database.crossplane.io/v1alpha1
+    kind: MySQLInstance
+  resources:
+  - resource:
```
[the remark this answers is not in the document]
negz: Yeah, I had a similar thought. We do need to know what kind to support binding to during static provisioning though. I thought we could infer it from the class kind, but in some cases (e.g. Azure which has `SQLServerClass` and `MySQLServer`) that's not true. It seemed safer to be explicit about the claim -> resource mapping as well as the claim -> class mapping.

design/design-doc-resource-claim-definition.md
negz: Agreed - I'm holding off on putting too much detail here until @bassam has had a chance to weigh in.

design/design-doc-resource-claim-definition.md
negz: I actually had them embedded at first as you suggest, but @bassam and @suskin pointed out that it's possible the claim schema author won't be the same person as the binding author. For example we might have a stack that publishes a few common claim schemas, but doesn't enable binding to them. Then later an infrastructure operator might want to add binding definitions, but only for the claims that are interesting to them.

Will definitely add this context to the document if we go down this path.

design/design-doc-resource-claim-definition.md
negz: I would imagine that would depend on whether the user's mental model was "the behaviour of this claim is defined by its RCD" or "the behaviour of this claim is defined by its RCD and RBDs".

design/design-doc-resource-claim-definition.md
negz: I don't think we'd select one as such. I imagine it would be more using the set of RBDs as constraints when selecting a class. The RBDs constrain which resource classes are eligible for static and dynamic provisioning.

design/design-doc-resource-claim-definition.md
muvaf: I've given some more thought to this and I think I agree with you guys.

When an application stack developer makes decision of what infrastructure to use, they think of the claim _type_ as the medium between the app and the infrastructure. Claim is basically the abstraction layer. In this context, as long as the resource usage pattern (basically connection secret's shape) stays the same, infra operator should be able to change what you _actually_ get when you create that claim, which is way easier with separate `ResourceBindingDefinition`s.

negz:
I'm going to close this in favor of #1163, though the `ResourceClaimDefinition` part of the design bares exploring in a future PR.

PR #1129, opened 2019-12-19
api patterns: sensitive input field handling and external resource labelling

muvaf:

Currently, Crossplane does not support supplying sensitive data over the custom resource. This PR suggests a plan although it's useful only in static provisioning currently since _any_ input for dynamic provisioning has to be implemented in claim level.

It's easy to identify what external resources correspond to what managed resource by looking at your managed resource but the other way around is not really trivial, in some cases not possible. For example, AWS VPCs have random generated name and when you look at it in the console you can't even differentiate whether it was provisioned by Crossplane or not. This PR suggests that we label (if possible) the resource with the manage resource CR name.

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/design/one-pager-managed-resource-api-design.md
+++ b/design/one-pager-managed-resource-api-design.md
@@ -1,13 +1,17 @@
 # Managed Resources API Patterns
 * Owner: Muvaffak Onus (@muvaf)
 * Reviewers: Crossplane Maintainers
-* Status: Accepted, Revision 1.1
+* Status: Accepted, Revision 1.2
 
 ## Revisions
 
-* 1.1
+* 1.1 - Daniel Mangum (@hasheddan)
   * Added [Embedded Structs with Mixed Fields](#embedded-structs-with-mixed-fields) and [Optional Embedded Structs with One Field](#optional-embedded-structs-with-one-field) sections
   * Updated examples with cluster-scoped managed resource and `Provider` objects
+* 1.2 - Muvaffak Onus (@muvaf)
+  * Added [how to handle sensitive input fields](#sensitive-input-fields).
+  * Added [external resource labelling](#external-resource-labeling).
+  * Added [cross-resource reference edge case](#pointer-types-and-markers) for types.
 
 ## Terminology
 
@@ -115,12 +119,25 @@ Related to https://github.com/crossplaneio/crossplane/issues/624
 The decision for an external resource name is made by the controller of that managed resource in `Create` phase.
 Possible cases are as following:
 * The provider doesn't allow us to specify a name. Fetch it after its client's `POST` call. Override `crossplane.io/external-name`
-annotation's value with what you get as name.
+  annotation's value with what you get as name.
 * The provider allows to specify a name.
   * If `crossplane.io/external-name` annotation has a value use it. If it's empty;
   * Use managed resource's name and write that into `crossplane.io/external-name` annotation.
 
 Use the value of that annotation as external resource name in _all_ queries.
+
+#### External Resource Labeling
+
+If the external resource supports labelling, we should label it with the managed
+resource name. While this is helpful for services with non-deterministic naming,
+it'd benefit the services that allow you to name the external resource, too,
+since the labels could be used for searching and batch operations on the cloud level.
+The key to use in those labels is `crossplane.io/resource-name` and the value is
+the name of the managed resource. An example would look like:
+```
+A tag for a VPC in AWS:
+  "crossplane.io/resource-name": "myappnamespace-mynetwork-5sc8a"
+```
 
 #### Field Names
 
@@ -246,6 +263,23 @@ type EmbeddedStructStatus struct {
 }
 ```
 
+#### Sensitive Input Fields
+
+Some cloud services can take sensitive input such as passwords, certificates or
+tokens. However, exposing those fields on the CR is not the best way to handle it.
+For such fields, the field name should be `<field-name>SecretRef` and the type
+of that field should be `SecretKeySelector`([from crossplane-runtime]). The controller
+should fetch the value from the given secret and populate the field in the provider's
+corresponding SDK object.
+
+In case the secret reference is not provided **and** the field is a required one,
+such as password, the controller should generate it randomly. Then it has to make
+sure that:
+* GoDoc comment explicitly states that if `<field-name>SecretRef` is not populated,
+  a random value will be generated for it.
+* The generated value ends up in the connection details secret that is published
+  by the controller so that no information is lost or not exposed to the user.
+
 ### Owner-based Struct Tree
 
 Related to https://github.com/crossplaneio/crossplane/issues/728
@@ -354,6 +388,13 @@ Here is the flow to decide how to make the decision for `Spec` fields:
 1. Start with assuming all fields are optional.
 2. Convert the fields to required if any of CRUD calls you make requires them, i.e. it should be possible to call all
 CRUD operations with all those required fields filled.
+
+There is an edge case where `FieldA` is required but there is also `FieldARef`,
+which means in the runtime that reference will be resolved and `FieldA` will be
+populated. If there is a guarantee that no call will be made to the provider until
+`FieldA` is populated (at the time of this writing, there is), you need to add
+`omitempty` as struct tag so that Kubernetes allows the creation of the resource
+with only `FieldARef` populated.
 
 The decision flow for `Status` fields are different. The values for those fields are provided by the provider and
 overridden by the latest observation no matter what and we know that we'll get a full object body. However;
@@ -529,3 +570,4 @@ Generic managed reconciler's `ExternalObservation` struct could be extended by a
 and reconciler can mark the sync status in one of the `Condition`s we already have or add a new one.
 
 [glossary]: https://github.com/crossplaneio/crossplane/blob/master/docs/concepts.md#glossary
+[from crossplane-runtime]: https://github.com/crossplaneio/crossplane-runtime/blob/ca4b6b4/apis/core/v1alpha1/resource.go#L77
--- a/design/one-pager-managed-resource-api-design.md
+++ b/design/one-pager-managed-resource-api-design.md
@@ -130,15 +130,22 @@ Use the value of that annotation as external resource name in _all_ queries.
 #### External Resource Labeling
 
 If the external resource supports labelling, we should label it with the managed
-resource name. While this is helpful for services with non-deterministic naming,
-it'd benefit the services that allow you to name the external resource, too,
-since the labels could be used for searching and batch operations on the cloud level.
-The key to use in those labels is `crossplane.io/resource-name` and the value is
-the name of the managed resource. An example would look like:
+resource name and kind. This is helpful in variety of scenarios like:
+* Identify services with non-deterministic naming, for example AWS VPC.
+* Provider level operations that are done via label filtering, for example search
+  and batch operations.
+* Find a resource in the provider UI and construct an easy `kubectl` query on
+  Crossplane cluster to find the associated resource, like `kubectl get <kind> <name>`
+  
+The keys to use in labels are like the following:
 ```
-A tag for a VPC in AWS:
-  "crossplane.io/resource-name": "myappnamespace-mynetwork-5sc8a"
+A tag set for a VPC in AWS:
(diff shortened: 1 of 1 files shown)


design/one-pager-managed-resource-api-design.md
```diff
@@ -115,13 +118,24 @@ Related to https://github.com/crossplaneio/crossplane/issues/624
...
-annotation's value with what you get as name.
+  annotation's value with what you get as name.
 * The provider allows to specify a name.
   * If `crossplane.io/external-name` annotation has a value use it. If it's empty;
   * Use managed resource's name and write that into `crossplane.io/external-name` annotation.
 
 Use the value of that annotation as external resource name in _all_ queries.
 
+If the external resource supports labelling, we should label it with the managed
+resource name. While this is helpful for services with non-deterministic naming,
```
hasheddan: Are you thinking that we add this label automatically? So if a user supplies labels we append this one to the set in the controller?

design/one-pager-managed-resource-api-design.md
```diff
@@ -246,6 +260,20 @@ type EmbeddedStructStatus struct {
 }
 ```
 
+#### Sensitive Input Fields
+
+Some cloud services can take sensitive input such as passwords, certificates or
+tokens. However, exposing those fields on the CR is not the best way to handle it.
+For such fields, the field name should be `<field-name>SecretRef` and the type
+of that field should be `SecretKeySelector`([from crossplane-runtime]). The controller
```
hasheddan: One thing that immediately come to mind here is do we setup a watch on this secret such that if a user changes it then it triggers an update in the managed resource controller?

design/one-pager-managed-resource-api-design.md
muvaf: > Are you thinking that we add this label automatically?

The infra stack developer would have to make sure they add the label in `ExternalClient` functions because the field to add the label differs.

> So if a user supplies labels we append this one to the set in the controller?

Yes

design/one-pager-managed-resource-api-design.md
muvaf: That's an implementation detail but I don't think we need a watch for those secrets. In general, you don't really update those. For example, you can't really update the root certificate that you supplied. Not a strong opinion, but I feel like 1 minute long-wait interval we have would be sufficient.

hasheddan: Gotcha, so setting these would just be part of the reference resolution chain?

muvaf: > would just be part of the reference resolution chain

No, not really. Reference resolution populates the field on the CR but we don't want that for sensitive fields. So, in `Create` and/or `Update`, the controller has to fetch the secret data and include it in the final SDK object.

hasheddan: Hmm I wonder if we could somehow be more generic about how sensitive data is fetched for CR's such that a user could potentially specify a variety of ways to fetch secrets other than just from a Kubernetes `Secret`.

muvaf: In the implementation, we don't impose any mechanism like we do for reference resolving. So, they are actually free to do so in controller code. But in general, I believe as long as you're on Kubernetes level, it's safe to assume that the sensitive data is in a `Secret`.

I just looked into how Vault is used in Kubernetes but apparently, it just injects the secret values into the pods without using Kubernetes Secrets. For our case, this is simply not possible since the controller pod works for multiple user entities. So, I feel like our best bet for now is to rely on Kubernetes secrets. @negz what do you think?

muvaf:
@hasheddan I added the edge case of referenced fields having `omitempty` even if they are required, inspired by your decision in NodePools.

design/one-pager-managed-resource-api-design.md
hasheddan: I think this seems like a solid direction for now, and we can iterate on it as community requests additional functionality :)

design/one-pager-managed-resource-api-design.md
```diff
@@ -246,6 +263,20 @@ type EmbeddedStructStatus struct {
...
+should fetch the value from the given secret and populate the field in the provider's
+corresponding SDK object.
+
+In case the secret reference is not provided **and** the field is a required one,
+such as password, the controller should generate it randomly. Then it has to make
+sure the generated value ends up in the connection details secret that is published
+by the controller so that no information is lost or not exposed to the user.
```
negz: Should we include a requirement in this document that such fields must document this behaviour in their GoDoc comment?

design/one-pager-managed-resource-api-design.md
negz: I agree that secrets are the best place to start for now. If we wanted to future proof this API for other credential stores we could use the "union-with-discriminator" pattern described in https://github.com/kubernetes/enhancements/pull/926, e.g.

```yaml
fooCredentials:
  credentialStore: Secret
  secretRef:
    name: coolsecret
```

This would allow us to add `credentialStore: Vault`, etc in future. This all said, I probably would not worry about this for now given that none of our other credentials (providers, connection secrets) support anything other than secrets.

muvaf:
@negz added your suggestion. This should be ready to merge.

design/one-pager-managed-resource-api-design.md
```diff
@@ -115,13 +119,26 @@ Related to https://github.com/crossplaneio/crossplane/issues/624
...
+
+If the external resource supports labelling, we should label it with the managed
+resource name. While this is helpful for services with non-deterministic naming,
+it'd benefit the services that allow you to name the external resource, too,
+since the labels could be used for searching and batch operations on the cloud level.
+The key to use in those labels is `crossplane.io/resource-name` and the value is
+the name of the managed resource. An example would look like:
+```
+A tag for a VPC in AWS:
+  "crossplane.io/resource-name": "myappnamespace-mynetwork-5sc8a"
```
muvaf: @negz @hasheddan I am actually not sure about the `resource-name` part here. Should it be the kind of the managed resource? Like `crossplane.io/cloudsqlinstance` or `crossplane.io/cloudsqlinstance-name`?

hasheddan: I think either way would be fine, but I agree that using the actual resource type in the tag would make it easier to identify.

negz: If we wanted to encode the managed resource kind (as Crossplane sees it), perhaps we should do that with a separate label? I imagine the relationship would be pretty obvious though (i.e. a `cloudsqlinstance` is never not going to be a CloudSQL instance in GCP and vice versa).

muvaf: > I imagine the relationship would be pretty obvious though (i.e. a cloudsqlinstance is never not going to be a CloudSQL instance in GCP and vice versa).

I agree. But I'm thinking more like how a user sees an external resource and be able to find it in the cluster easily and exact kind string could help. User knows that the service is Cloud SQL or Azure Database but the kind name is a bit different in some cases, for example https://github.com/crossplaneio/stack-azure/issues/65

How about the following pair?
```
"crossplane.io/resource-name": "myappnamespace-mynetwork-5sc8a"
"crossplane.io/resource-kind": "cloudsqlinstance"
```
or
```
"crossplane.io/name": "myappnamespace-mynetwork-5sc8a"
"crossplane.io/kind": "cloudsqlinstance"
```
Maybe `resource` is unnecessary?

muvaf:
I just added a condition for `status` to be reproducible. @negz once told me about it and it's present in the API conventions. I just ran into such a case in an aws routetable controller and thought it'd be useful if we pin that down.

@negz @hasheddan you might want to check it before merging.

design/one-pager-managed-resource-api-design.md
muvaf: @hasheddan @negz ping

hasheddan: @muvaf apologies for letting this slip by, I would advocate for the first of the two options you propose here :)

design/one-pager-managed-resource-api-design.md
```diff
@@ -173,6 +191,14 @@ For both `Status` and `Spec`:
   What if the sub-resource is not yet supported as managed resource in Crossplane? In that case, you should first
   consider implementing that managed resource, if not suitable, only then include it in the CR.
 
+* Can the value of this field be reproduced when the whole `Status` is deleted? Do not include if the answer is no.
+  `Status` sub-resource is the _representation_ of the current state, so, the
+  controller should be able to reproduce it as long as the state that's represented
+  is still there. In practice, this means the controller should not rely on any
+  field that is present in the `Status`.
```
hasheddan: Solid improvement here :+1:

design/one-pager-managed-resource-api-design.md
negz: > I'm thinking more like how a user sees an external resource and be able to find it in the cluster easily and exact kind string could help.

Ah - it took me a few minutes just now to grasp this concept but it makes sense to me now. I see the value in adding the kind as a label/tag and think we should do it.

> I would advocate for the first of the two options you propose here.

What's your thinking here @hasheddan? My suspicion is that we're going to run into a lot of weird and wonderful limitations across different providers around what characters and lengths are valid for tags and labels, so we should probably pick the shortest, simplest keys possible.

design/one-pager-managed-resource-api-design.md
negz: Should this section move up under "What goes into status"? It's currently under "For both status and Spec".

> In practice, this means the controller should not rely on any field that is present in the `Status`.

Should we elaborate on this a little? It's a bit vague here. I think it's safe for controller A to rely on the status of a resource X controlled by controller B, presuming that controller B can reproduce the status of resource X if and when it is lost. I agree that controller B should be able to function (perhaps with degraded behaviour) when the status of resource X is lost.

Or perhaps we don't need to discuss this at all? We already link to the relevant section of the API conventions - perhaps that's enough?

design/one-pager-managed-resource-api-design.md
hasheddan: @negz my primary thinking was that `resource-*` was more explicit, but I could go either way. On this topic, what is our course of action if whatever structure we choose is not acceptable to the provider?

negz: Got it! I'd prefer to leave off the `resource-` prefix. Ideally it's implied (I don't expect us to be needing to label anything in the provider that did not correspond to a Crossplane managed resource).

> What is our course of action if whatever structure we choose is not acceptable to the provider?

I think we're going to have to be pretty flexible around how this is implemented in crossplane-runtime. For example I imagine in some tagging systems we're going to have to add single tags like `crossplane.io/name=foo`, in other systems we're going to need to live without the `./` characters, in some we're going to have to live within character limits (`c8e.io/name`?). Given these labels are more for humans than automation I think it's okay to be a little inconsistent if we need to, as long as we can come up with a format that communicates "this is a Crossplane thing" in each case.

hasheddan: @negz what would you think about making a `managed.WithTagger` parameter for the managed reconciler that used the pattern @muvaf outlines above by default but also allows for overriding? Could potentially allow for a stack to define a single tagging strategy that could be used across all of their resource types, which would help enforce consistency across a given provider.

muvaf: @hasheddan I think what we'll need from managed resource developers is a function that takes at least `key string, value string` parameters because Crossplane doesn't know which field of the resource is used for tagging, this part could be left to the developer. But I feel like we'll discover the edge cases when we do start to implement this. As you mentioned, they should be able to override the default, too. Maybe we'd just leave it completely to the developer.

design/one-pager-managed-resource-api-design.md
muvaf: > Should this section move up under "What goes into status"?

Yes, that makes more sense.

> We already link to the relevant section of the API conventions - perhaps that's enough?

I think explaining what goes to where in a logical expression fashion helps with making the decisions faster. Reader could go and read that section but it's easier to answer the questions and come up with a conclusion for many fields repeatedly.

muvaf:
@negz I've made the updates. Feel free to merge it if you're onboard with the changes.

negz:
🎉 Thanks for persisting with this @muvaf!

PR #1132, opened 2019-12-21
Design sketch: Crossplane as a Kubernetes native OAM runtime

negz:
This PR introduces a design that proposes how Crossplane might become an [Open Application Model](https://oam.dev/) runtime - a project capable of enacting OAM configuration. The design is currently an early sketch, and is the result of discussions with @bassam and @prasek. It proposes Crossplane support a variant of the OAM specification that maintains OAM's core concepts while aligning them more closely with Crossplane's "Kubernetes (API) native" approach to application and infrastructure orchestration.

### DIFF
--- a/design/design-sketch-oam.md
+++ b/design/design-sketch-oam.md
@@ -1,0 +1,334 @@
+# Kubernetes Native OAM and Crossplane
+
+* Owners: Nic Cope (@negz), Bassam Tabbara (@bassam), Phil Prasek (@prasek)
+* Reviewers: [Crossplane maintainers], [OAM maintainers]
+* Status: Draft
+
+## Background
+
+The [Open Application Model] (OAM) is a specification for building cloud native
+applications. OAM aims to enable the separation of concerns between application
+developers, application operators, and infrastructure operators. Crossplane is
+an open source multicloud control plane that shares many of OAM's goals around
+modelling cloud native applications and separation of concerns.
+
+The OAM specification defines four core configuration documents:
+
+* A `ComponentSchematic` describes a component. It is a schema - something that
+  may be composed to run an application, but does not represent a running
+  application in and of itself. A component is of a particular workload type.
+  The OAM defines core types, all of which are shapes of containerized workload.
+  OAM runtimes may define "extended workloads", which define their own schema.
+* An `ApplicationScope` groups components, for example by placing them in a
+  particular VPC network or applying resource usage constraints on them.
+* A `Trait` augments a `ComponentSchematic` with additional functionality, for
+  example by adding load balancing or rate limiting to the component.
+* An `ApplicationConfiguration` instantiates a group of components by specifying
+  and providing inputs to their schematics, applying any traits associated with
+  each schematic, and grouping the component instances into one or more scopes.
+
+Crossplane models an application as a _workload_, of which the only extant kind
+is `KubernetesApplication`. A `KubernetesApplication` is an array of templates
+for Kubernetes resources of any kind that are co-scheduled and submitted to a
+Kubernetes control plane (distinct from the Kubernetes control plane upon which
+Crossplane runs). Crossplane _may_ support other workload kinds, for example
+`ServerlessApplication`, in future.
+
+Crossplane also models and manages the cloud infrastructure that an application
+may depend upon in the form of _managed resources_, and _resource claims_. A
+resource claim models the need for a service from an application's perspective,
+i.e. "a MySQL instance" (`kind: MySQLInstance`). These needs are resolved by
+binding the claim to a managed resource, which represents a concrete, specific,
+instance of cloud infrastructure (for example `kind: MySQLServer` in Azure, or
+`kind: RDSInstance` in AWS).
+
+The Crossplane and OAM projects have similar goals. Both projects recognise the
+need to model cloud native applications, and to enable the separation of
+concerns between the personas who typically build and run such applications.
+Both projects draw inspiration from Kubernetes, but each have embraced it to a
+different extent. The configuration documents outlined by the OAM specification
+are a subset of the Kubernetes [API conventions], and Rudr (the only extant open
+source OAM runtime) is built on Kubernetes, but the OAM specification does not
+require that runtimes take a dependency on Kubernetes. Crossplane's ambitions
+extend far beyond orchestrating Kubernetes clusters - it intends to model
+workloads and infrastructure that have nothing to do with Kubernetes - but it
+_does_ take a dependency and build on the Kubernetes control plane. All
+Crossplane configuration documents _must_ be defined as Kubernetes [Custom
+Resource Definitions] (CRDs), and _must_ be reconciled by Kubernetes
+[controllers].
+
+## Goals
+
+The goal of this proposal is to align OAM and Crossplane configuration concepts,
+enabling Crossplane to act as an OAM runtime - a project capable of enacting
+applications defined via OAM.
+
+## Proposal
+
+This document proposes that OAM take a dependency on the Kubernetes API server.
+OAM configuration schema would be defined as a series of CRDs. We do not believe
+this to be incompatible with OAM’s goal of supporting multiple orchestrators and
+runtimes. Our experience with the Crossplane project has demonstrated that the
+API server and controller combination provides a strong framework upon which to
+build orchestration software, even when Kubernetes is not the desired platform
+upon which to _run_ the orchestrated software and infrastructure. We believe
+taking this dependency will simplify the OAM specification and its runtime
+implementations. Taking this dependency would result in a stronger alignment
+with the Cloud-Native community, enabling powerful scenarios around multicloud
+applications. Embracing Kubernetes API conventions would also serve to make the
+OAM specification itself more succinct by removing the need to re-specify common
+requirements such as type and object metadata.
+
+Specifically this document proposes:
+
+* Each type of OAM workload - be it core or extended - be represented by a
+  distinct kind of custom resource. Such resources would represent an _instance_
+  of configuration to be reconciled by a runtime (i.e. controller), not a
+  _schema_. The schema would be defined by a CRD. For example the core workload
+  types may be recast as `kind: ContainerizedWorkload`. The existing, unmodified
+  Crossplane resource claim kinds (for example `kind: RedisCluster`) could
+  effectively function as an extended workload type.
+* Similarly, each type of application scope and trait would be a distinct kind
+  of Kubernetes resource, defined by its own CRD, for example
+  `kind: NetworkScope`, or `kind: ManualScalerTrait`.
+* Each `kind: ApplicationConfiguration` resource would instantiate application
+  configurations by specifying the configuration for the application's
+  components and traits inline. The creation of an `ApplicationConfiguration`
+  would result in the creation of these 'templated' resources, each of which
+  could be reconciled by its own Kubernetes controller. This is similar to how
+  Crossplane's `KubernetesApplication` functions. Scopes, which may apply to
+  many applications, would be configured in advance and referenced by an
+  `ApplicationConfiguration`.
+
+The following is an example of the proposed `ApplicationConfiguration` (drawing
+from the examples included in the contemporary OAM specification) that includes
+a Crossplane `RedisCluster` resource claim in the place of an extended workload
+type, and uses a `SchedulingScope` to determine where Crossplane should deploy
+the application.
+
+```yaml
+---
+apiVersion: oam.crossplane.io/v1alpha1
+kind: SchedulingScope
+metadata:
+  namespace: default
+  name: production-europe
+  annotations:
+    description: >
+      Ensures applications are scheduled to production infrastructure in Europe.
+spec:
+  # This label selector is used to limit where this application's components
+  # may run. What exactly it selects (a Kubernetes cluster, an instance
+  # group, a serverless platform, etc) is up to the runtime.
+  selector:
+    matchLabels:
+      environment: production
+      region: europe
+---
+kind: ApplicationConfiguration
+metadata:
+  namespace: default
+  name: coolest-app
+  annotations:
+    description: A very cool cloud-native application.
+spec:
+  scopes:
+  - scopeRef:
+      apiVersion: core.oam.dev/v1alpha1
+      kind: SchedulingScope
+      name: production-europe
+  components:
+  - component:
+      apiVersion: core.oam.dev/v1alpha1
+      kind: ContainerizedWorkload
+      metadata:
+        annotations:
+          version: v1.0.0
+      spec:
+        # Under this proposal the workloadType would be limited to the core OAM
+        # types, i.e. it would define the function of the resulting workload -
+        # Server, Singleton, Worker, etc. Extended workloads are a distinct kind
+        # and thus kind: ContainerizedWorkload has no workloadSettings field.
+        workloadType: core.oam.dev/v1alpha1.Server
+        osType: linux
+        containers:
+        - name: my-twitter-bot-frontend
+          image:
+            name: example/my-twitter-bot-frontend:v1.0.0
+          resources:
+            cpu:
+              required: 1.0
+            memory:
+              required: 100MB
+          ports:
+          - name: http
+            value: 8080
+          env:
+          - name: USERNAME
+            value: cooladmin
+          - name: PASSWORD
+            valueFrom:
+              secretKeyRef:
+                name: cool-frontend
+                key: password
+          - name: BACKEND_ADDRESS
+            value: https://example.org
+          livenessProbe:
+            httpGet:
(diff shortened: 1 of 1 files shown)


design/design-sketch-oam.md
```diff
@@ -0,0 +1,334 @@
...
+        port: 8080
+        path: /healthz
+---
+apiVersion: core.oam.dev/v1alpha1
+kind: ManualScalerTrait
+metadata:
+  namespace: default
+  name: coolest-app-n3v3y
+spec:
+  replicaCount: 3
```
negz: > Traits should be applied into the runtime at installation/upgrade time. Traits should not be checked "lazily", but should be checked when the trait-holding components are created.

> Traits are applied in order so that should some set of traits express dependency (e.g. ingress must be set up before SSL), this can be resolved by setting an explicit ordering.

> A deployment SHOULD NOT be marked complete until all trait configurations are completed. For example, if a web server component is deployed with an autoscaler trait, the web server should not be considered "running" until (a) the web server itself is running, as determined by health checks, and (b) the autoscaler trait is running, as determined by the underlying platform.

Each `ContainerizedWorkload` resource would need to maintain references to the traits that should apply to it in order for the controller that processes it to satisfy the above requirements of the contemporary OAM spec. Perhaps this would take the form of a `.spec.traitRefs` array, or perhaps we could model this using owner references (from trait to the component they apply to).

https://github.com/oam-dev/spec/blob/master/5.traits.md#traits-system-rules-and-characteristics

design/design-sketch-oam.md
```diff
@@ -0,0 +1,334 @@
...
+apiVersion: cache.crossplane.io/v1alpha1
+kind: RedisCluster
+metadata:
+  namespace: default
+  name: coolest-app-jnkf9
+spec:
+  classSelector:
+    matchLabels:
+      environment: production
+      region: europe
```
negz: Ideally this `classSelector` would relate to/be derived from the hypothetical `SchedulingScope`. Doing so would require either:

* The referenced `SchedulingScope` be used to modify the templated `RedisCluster` (which would then differ from its template).
* The `RedisCluster` resource maintain references to any referenced `Scope` resources, and the `RedisCluster` controllers be aware of those scope resources.

design/design-sketch-oam.md
```diff
@@ -0,0 +1,334 @@
...
+
+Submitting the above `ApplicationController` would cause an instance of each of
+the scopes, components, and traits it templates to be created verbatim (modulo
+some generated object metadata) similar to how the `template` of a Kubernetes
+`ReplicaSet` becomes a set of `Pod` resources. For example:
+
+```yaml
+---
+apiVersion: core.oam.dev/v1alpha1
+kind: ContainerizedWorkload
```
hongchaodeng: Actually this would be kind `ComponentSchematic`:

```yaml
kind: Component
metadata:
  name: containerized-workload
spec:
  parameters:
    ... the schema of the spec of ContainerizedWorkload ...
```

A deployment would be done in ApplicationConfiguration components section:

```yaml
kind: AppConfig
metadata: ...
spec:
  components:
  - componentName: ContainerizedWorkload
    instanceName: coolest-app-f92dm
    parameterValues: ...
```

Note that we are trying to migrate all of the schema to WorkloadType. So ideally ContainerizedWorkload is a Workload type, as the name suggested itself :)

design/design-sketch-oam.md
hongchaodeng: Well, the deployment of traits only happens in AppConfig:

```yaml
kind: AppConfig
metadata: ...
spec:
  components:
  - componentName: ContainerizedWorkload
    instanceName: coolest-app-f92dm
    parameterValues: ...
    traits:
    - type: ManualScalerTrait
      properties:
        replicaCount: 3
```

design/design-sketch-oam.md
```diff
@@ -0,0 +1,334 @@
...
+apiVersion: core.oam.dev/v1alpha1
+kind: ManualScalerTrait
+metadata:
+  namespace: default
+  name: coolest-app-n3v3y
+spec:
+  replicaCount: 3
+---
+apiVersion: cache.crossplane.io/v1alpha1
+kind: RedisCluster
```
hongchaodeng: This should be modelled as an OAM Workload.

Proposal https://github.com/oam-dev/spec/issues/281 will map it more intuitively.

design/design-sketch-oam.md
hongchaodeng: > Ideally this classSelector would relate to/be derived from the hypothetical SchedulingScope.

Exactly!
In OAM this relationship is declared via AppConfig:

```
kind: AppConfig
metadata: ...
spec:
  components:
  - componentName: ContainerizedWorkload
    instanceName: coolest-app-f92dm
    scopes:
    - name: scheduling-scope-eu-prod
    parameterValues: ...
    traits:
    - type: ManualScalerTrait
      properties:
        replicaCount: 3
```

design/design-sketch-oam.md
negz: Similar to my comment above regarding application scopes, we're proposing a change to how OAM components are defined here. Specifically we're exploring the idea the one `ComponentSchematic` kind be broken down into many kinds of component, e.g. `ContainerizedWorkload`, `RedisCluster`, etc. The schema for each of these kinds would be defined by their custom resource definitions - not a workload type. In general we feel that Kubernetes CRDs should define configuration schema, and CRs should instantiate configuration. What do you think about this idea?

design/design-sketch-oam.md
negz: We're suggesting that an `ApplicationConfiguration` effectively be a template for instances of components and their traits - a little like how `KubernetesApplication` is a template for arbitrary Kubernetes resources (`Deployment`, etc). So under this pattern the OAM runtime (i.e. a Kubernetes controller) watching for `ApplicationConfiguration` would not actually deploy anything - instead it would create instances of each of the templated custom resource (in this case a `ContainerizedWorkload`, a `ManualScalerTrait`, and a `RedisCluster`. This would allow separate controllers to reconcile each of these custom resources. One advantage of this pattern is that it allows Crossplane's current resource claim controllers to work with OAM without a layer of abstraction (e.g. a `RedisCluster` workload type). What do you think?

design/design-sketch-oam.md
negz: Similar to my comments above, we're proposing that custom resource definitions (not component schematics) be used to define the available kinds of workload and their configuration schema.

design/design-sketch-oam.md
```diff
@@ -0,0 +1,334 @@
...
+The following is an example of the proposed `ApplicationConfiguration` (drawing
+from the examples included in the contemporary OAM specification) that includes
+a Crossplane `RedisCluster` resource claim in the place of an extended workload
+type, and uses a `SchedulingScope` to determine where Crossplane should deploy
+the application.
+
+```yaml
+---
+apiVersion: oam.crossplane.io/v1alpha1
+kind: SchedulingScope
```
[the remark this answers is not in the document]
resouer: It's great, we saw the core idea of this proposal. This is mainly about whether we could use CRD as the generic definition of schema. Essentially, OAM has a "schematic object" to carry the schema of Component/Trait/Scope. In non-k8s runtimes, they could interpret this "schematic object" as OAM object schemas.

If we change to use CRD to model these schemas, how would you imagine, for example, if we want to define workload on FaaS (or other non-K8s etc)? I can think of: 1. always install a kube-apiserver; 2. implement a OAM apiserver.

design/design-sketch-oam.md
```diff
@@ -0,0 +1,334 @@
...
+    traits:
+    # Contemporary traits model the workload kinds to which they may apply. This
+    # could be modelled as an annotation on the CRD of each Trait kind.
+    - apiVersion: core.oam.dev/v1alpha1
+      kind: ManualScalerTrait
+      spec:
+        replicaCount: 3
+  - component:
+      apiVersion: cache.crossplane.io/v1alpha1
+      kind: RedisCluster
```
resouer: Nits: seems this `RedisCluster` need to be changed to an extended workload spec.

design/design-sketch-oam.md
```diff
@@ -0,0 +1,334 @@
...
+  Similar to the above concerns, if an `ApplicationConfiguration` templates and
+  produces resource claims verbatim (as opposed to producing a resource claim
+  modified based on the supplied scopes), the controllers for said claims would
+  need to be aware of scopes in order to consider them when acting on the
+  resource claim.
+* What is the best approach to validate the array of components and traits under
+  the `ApplicationConfiguration` `components` field? One option may be to use a
+  validating webhook that finds the CRD that defines the templated kind and uses
+  its OpenAPI schema to validate it.
+* How could we maintain the existing user experience around easily discovering
```
wonderflow: CRD categories will list all resources belong to that category. In fact we just want to list Trait CRDs, so we may need to use label.

```
kubectl get crds -l  oam.dev/part-of=trait
```

design/design-sketch-oam.md
```diff
@@ -0,0 +1,334 @@
...
+* Should Crossplane resource claims natively support the concept of OAM scopes?
+  Similar to the above concerns, if an `ApplicationConfiguration` templates and
+  produces resource claims verbatim (as opposed to producing a resource claim
+  modified based on the supplied scopes), the controllers for said claims would
+  need to be aware of scopes in order to consider them when acting on the
+  resource claim.
+* What is the best approach to validate the array of components and traits under
+  the `ApplicationConfiguration` `components` field? One option may be to use a
+  validating webhook that finds the CRD that defines the templated kind and uses
+  its OpenAPI schema to validate it.
```
wonderflow: Why don't we just use `WorkloadType` and `Trait` to define the schema? We can also use `validating webhook` to check the schema. In fact, as we use `ApplicationConfiguration`, we already have a layer between user application and K8s CRD. So `WorkloadType` and `Trait` didn't bring any burden, they can help unify different CRDs.

design/design-sketch-oam.md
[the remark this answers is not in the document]
negz: > If we change to use CRD to model these schemas, how would you imagine, for example, if we want to define workload on FaaS (or other non-K8s etc)? I can think of: 1. always install a kube-apiserve; 2. implement a OAM apiserver.

We're strong proponents of approach one - it's exactly the approach we've taken with Crossplane and where we're betting the Kubernetes ecosystem will head. I think the fact that the OAM spec is inspired by Kubernetes API documents, and that the first OSS OAM implementation (Rudr) builds on Kubernetes is telling. It's a great general infrastructure control plane, even if you're not doing anything with containers.

We've always imagined Crossplane would grow to include a `ServerlessApplication`, `VMApplication`, etc alongside `KubernetesApplication`. These resources could orchestrate applications running outside of Kubernetes, but we'd still use the Kubernetes API server and controllers to orchestrate them.

design/design-sketch-oam.md
negz: > CRD categories will list all resources belong to that category.

I see your point that listing the CRDs (the schema) would be closer to today's `kubectl get trait`, given that today traits are schema. It could also be useful to be able to list all instances of traits (even if they were different CR kinds), though.

wonderflow: But I think we could still use `WorkloadType` and `Trait` to do register/discovery thing.

vturecek: Traits are meant to be behavioral characteristics on instances of components. It wouldn't make sense to list _instances_ of traits.

design/design-sketch-oam.md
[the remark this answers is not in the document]
resouer: Got it. I agree we could assume K8s as the generic underlying control plane. This aligns with our current practice internally in Alibaba for VM workloads.

vturecek:
Thanks for pulling this together @negz . So the overall idea, if I'm understanding it correctly, is to define OAM in terms of the Kubernetes API, rather than just using Kubernetes conventions, and in doing so, do it in a way that's more natural to the Kubernetes API, i.e., CRDs for each _type_ within the primary constructs (traits, scopes, workloads).

A few questions that came up for me:
 - How does this affect the separation of concerns between an application developer, an application operator, and an infrastructure operator? 
 - How does it affect portability or consistency across implementations of the spec? 
 - Does it still make sense to think of OAM as a platform-neutral specification? 
 - How does this change tooling and user experience around discovery of capabilities for a given implementation of the spec?

design/design-sketch-oam.md
negz: This is actually by design - one thing we were going for in this proposal is that existing Kubernetes kinds like `RedisCluster` (a Crossplane's resource claim) could effectively be "extended workloads". For a runtime like Crossplane this means we can tightly integrate our first class concepts (like resource claim) with OAM, instead of needing to define a new layer of abstraction (an extended workload kind) that would produce a resource claim behind the scenes.

design/design-sketch-oam.md
negz: > It wouldn't make sense to list instances of traits.

Would it make sense if these characteristics were modelled as distinct API objects when instantiated? If I follow correctly you're saying a trait, applied by the app config, might somehow mutate the component instance produced by a schematic. Our thinking was more along the lines that a component with traits (at the app config level) would produce a component instance and several trait instances - there would be relationships (references) between these documents such that the runtime (controller) reconciling the component instance could factor in the trait instances.

negz:
> Thanks for pulling this together @negz .

My pleasure!

> So the overall idea, if I'm understanding it correctly, is to define OAM in terms of the Kubernetes API, rather than just using Kubernetes conventions, and in doing so, do it in a way that's more natural to the Kubernetes API, i.e., CRDs for each type within the primary constructs (traits, scopes, workloads).

Exactly. Our feeling is that if we can assume the use of the Kubernetes API server, and thus the use of the actual Kubernetes API the OAM spec could be both simpler and a more natural fit for projects like Crossplane.

> How does this affect the separation of concerns between an application developer, an application operator, and an infrastructure operator?

I don't believe there's a dramatic change here - it's more a reshaping of existing concepts. An application developer can still publish the shape/schema of their components by publishing a component CRD. The same is true for infrastructure operators with trait/scope CRDs. Application operators would tie these all together by publishing an `ApplicationConfiguration` that references these things.

> How does it affect portability or consistency across implementations of the spec?

I don't believe this would be affected in any meaningful way. We could require all OAM runtimes to support a certain set of resource kinds, just as we currently require all OAM runtimes to support a certain set of workload types and scopes.

> Does it still make sense to think of OAM as a platform-neutral specification?

Is it important that the _specification_ to be platform-neutral, or that the runtimes that implement the specification be platform neutral? We're proposing that the specification assume it is implemented atop a specific API and technology (the Kubernetes API server), but this is distinct from proposing that the specification only manage Kubernetes applications or only be useful to folks who use Kubernetes. I'd argue that Crossplane is a platform neutral runtime, for example. We built Crossplane atop the API server, but we expect folks who do not use Kubernetes clusters at all will use Crossplane to manage their applications and infrastructure.

> How does this change tooling and user experience around discovery of capabilities for a given implementation of the spec?

It seems that the main impact under this proposal would be that a concept ("a component", or "a scope") would have more than one resource kind. Our thinking is that labels on CRDs (e.g. `kubectl get crd -l component=true`) (for schema discovery) and [CRD categories](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#categories) (for instance discovery) could help with this.

negz:
Hi folks, thanks for the great discussion around this proposal yesterday. It seems like we have a lot of alignment around the need for separation of concerns between application and infrastructure persona, and the problems we're trying to solve. Do you think we can align on OAM runtimes assuming the actual Kubernetes API (e.g. the use of the API server)? I think that once we do that we can workshop any potential improvements to the spec it might enable.

vturecek:
@negz the Alibaba Cloud team has a second draft of this proposal. It retains the main point of using unique resource kinds for each trait, scope, and workload. I posted it up to our mailing list so we can get more eyes on it (with a link back to here for the source material).

https://groups.google.com/forum/#!forum/oam-dev

Thanks again for putting in the time on the first draft, it's exciting to see this take shape.

design/design-sketch-oam.md
resouer: I see. 

Can you accept:

```
- component:
    instance:
      apiVersion: cache.crossplane.io/v1alpha1
      kind: RedisCluster
      spec:
        classSelector:
          matchLabels:
            environment: production
            region: europe
            ...
```
The only tiny compromise is a `instance`  field.

negz:
Hi all,

The latest iteration of this design has moved to [a Google Document](https://docs.google.com/document/d/1XBYj5Fg7A0gO6HYpvGqe2xPbm3iCulQfk9FMY6KCYV0/edit?ts=5e3219fa#). I'm going to close this pull request - please comment on the document!

Thanks,
Nic

PR #1137, opened 2019-12-26
workload: use kubeconfig key of KubernetesCluster connection secret f…

muvaf:

`kubeconfig` key is common to all `KuberentesCluster` connection secrets and it saves us from doing the parsing and client construction.

Fixes https://github.com/crossplaneio/crossplane/issues/1108

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/pkg/controller/workload/kubernetes/resource/resource.go
+++ b/pkg/controller/workload/kubernetes/resource/resource.go
@@ -31,6 +31,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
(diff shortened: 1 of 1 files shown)


muvaf:
Manually validated this PR with Wordpress sample stack & GCP.

muvaf:
@hasheddan that makes sense, just updated the PR.

negz:
I'm going to echo @hasheddan's sentiment in that it would be good to confirm that we've tested this and it works, but given how small the change is I'm going to optimistically merge it in the meantime.

muvaf:
> Manually validated this PR with Wordpress sample stack & GCP.

@negz I've tested the case where it uses only `kubeconfig` key without a fallback. In its final form, it has a fallback, which is the old logic, in case that key doesn't have any value. So, IMO it should be fine but I'll do some tests for other things today and I'd use the latest master to test this as well, just in case.

PR #1138, opened 2019-12-27
One-pager for Host Aware Stack Manager 

turkenh:
This PR introduces one-pager design document for host aware stack manager proposed in issue https://github.com/crossplaneio/crossplane/issues/1038

[skip ci]

PR #1139, opened 2019-12-27
Host aware stack manager

turkenh:

This PR resolves the issue https://github.com/crossplaneio/crossplane/issues/1038 by implementing the design proposed with one-pager in https://github.com/crossplaneio/crossplane/pull/1138.

By deploying `host cluster - tenant kubernetes` setup presented in [one-pager](design/one-pager-host-aware-stack-manager.md).

### Checklist
- [x] Update Makefile for preparing helm charts (using existing yamls)
- [x] Unit Tests
  - [x] Adapt existing unit tests
  - [x] Add new unit tests
- [x] Add comments to code
- [x] Test and make sure regular installation works as before
  appropriate.

[definition of done].


cluster/charts/crossplane-controllers/.gitignore
```diff
@@ -0,0 +1,2 @@
+values.yaml
+values.yaml-e
```
displague: This would appear to be a typo'd filename.  Is it?

design/one-pager-host-aware-stack-manager.md
```diff
@@ -0,0 +1,211 @@
+# Host Aware Stack Manager
```
displague: It looks like you've included #1138 in this PR.  I don't think we need to include it here since it doesn't provide any code changes and design changes can be tackled separately.

turkenh: Agreed, done!

cluster/charts/crossplane-controllers/.gitignore
turkenh: No, this file is generated after a `make build`
Same `.gitignore` already exist for current helm chart: https://github.com/crossplaneio/crossplane/blob/master/cluster/charts/crossplane/.gitignore

pkg/controller/stacks/host/config.go
```diff
@@ -0,0 +1,85 @@
...
+
+// NewHostedConfig returns a new HostAwareConfig based on the available environment variables.
+func NewHostedConfig() (*HostedConfig, error) {
+	tenantKubeconfig := os.Getenv(EnvTenantKubeconfig)
+	if tenantKubeconfig == "" {
+		// Hosted Mode is off, regular installation.
+		return nil, nil
+	}
+	// Hosted Mode is on, we need all the following environment variables to be set.
+	ns := os.Getenv(envControllerNamespace)
```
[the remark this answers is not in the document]
turkenh: Yes, it make sense and sounds better. 
The reason that I did like this is not to flow this parameter set down to this function from main.go. For example, I will need to change signature of [`setupWithManagerFunc`](https://github.com/crossplaneio/crossplane/pull/1139/files#diff-66525c8a65d872f540928ed763ef9d54R92) to accept extra config. 
Do you think it worths doing so? Or do you see a cleaner way?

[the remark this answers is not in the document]
negz: I agree that the plumbing could be a bit awkward, but typically I think it is best to have all config loaded in one place (in or close to `main.go`) and plumbed down to where it's used. This makes it easier for code readers to figure out where the config comes from, and can make it easier to override config in tests. So I do think it's worth doing. 

I would not shy away from solutions that involve refactoring our current `main.go` implementation (to whatever extent is necessary, including getting rid of the pluggable `setupWithManager` pattern). I think we arrived at the current implementation somewhat organically, rather than because anyone really loves it.

One pattern I like to use in this situation is to add [functional options](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis) to my controller setup function, then build those options from Kingpin flags in `main.go`.  You can see examples of this in [crossplane-tools](https://github.com/crossplaneio/crossplane-tools/blob/9033bd/cmd/angryjet/main.go) and [draino](https://github.com/planetlabs/draino/blob/a877e7f0852/cmd/draino/draino.go).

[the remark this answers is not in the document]
turkenh: Thanks a lot for the links.
Now, we are getting the config options via flags as you suggested.

cmd/crossplane/main.go
```diff
@@ -59,7 +65,11 @@ func main() {
...
 
 		// stack manage - adds the stack manager controllers and starts their reconcile loops
-		extManageCmd = extCmd.Command("manage", "Manage stacks (run stack manager controllers)")
+		extManageCmd                     = extCmd.Command("manage", "Manage stacks (run stack manager controllers)")
+		extManageTenantKubeconfig        = extManageCmd.Flag("tenant-kubeconfig", "The absolute path of the kubeconfig file to Tenant Kubernetes instance. Setting this will activate host aware mode of Stack Manager").String()
+		extManageHostControllerNamespace = extManageCmd.Flag("host-controller-namespace", "The namespace on Host Cluster where install and controller jobs/deployments will be created (required for host aware mode, ignored otherwise)").String()
+		extManageTenantAPIServerHost     = extManageCmd.Flag("tenant-apiserver-host", "The Kubernetes apiserver Host for Tenant Kubernetes instance (required for host aware mode, ignored otherwise)").String()
+		extManageTenantAPIServerPort     = extManageCmd.Flag("tenant-apiserver-port", "The Kubernetes apiserver Port for Tenant Kubernetes instance (required for host aware mode, ignored otherwise)").String()
```
negz: Do we need API server and port if we're using a kubeconfig file? Can't those settings be in the kubeconfig?

cmd/crossplane/main.go
```diff
@@ -59,7 +65,11 @@ func main() {
 		extCmd = app.Command("stack", "Perform operations on stacks")
 
 		// stack manage - adds the stack manager controllers and starts their reconcile loops
-		extManageCmd = extCmd.Command("manage", "Manage stacks (run stack manager controllers)")
+		extManageCmd                     = extCmd.Command("manage", "Manage stacks (run stack manager controllers)")
+		extManageTenantKubeconfig        = extManageCmd.Flag("tenant-kubeconfig", "The absolute path of the kubeconfig file to Tenant Kubernetes instance. Setting this will activate host aware mode of Stack Manager").String()
+		extManageHostControllerNamespace = extManageCmd.Flag("host-controller-namespace", "The namespace on Host Cluster where install and controller jobs/deployments will be created (required for host aware mode, ignored otherwise)").String()
```
negz: Should we set a default namespace? I believe adding `.Default('crossplane-system')` would do that.

cmd/crossplane/main.go
```diff
@@ -59,7 +65,11 @@ func main() {
 		extCmd = app.Command("stack", "Perform operations on stacks")
 
 		// stack manage - adds the stack manager controllers and starts their reconcile loops
-		extManageCmd = extCmd.Command("manage", "Manage stacks (run stack manager controllers)")
+		extManageCmd                     = extCmd.Command("manage", "Manage stacks (run stack manager controllers)")
+		extManageTenantKubeconfig        = extManageCmd.Flag("tenant-kubeconfig", "The absolute path of the kubeconfig file to Tenant Kubernetes instance. Setting this will activate host aware mode of Stack Manager").String()
```
negz: I think if you use `.ExistingFile()` instead of `.String()` Kingpin will ensure the referred path (if any) is an existing file.

cmd/crossplane/main.go
```diff
@@ -84,7 +94,7 @@ func main() {
 		runtimelog.SetLogger(zl)
 	}
 
-	var setupWithManagerFunc func(manager.Manager) error
+	var setupWithManagerFunc func(manager.Manager, appConfig) error
```
negz: This doesn't need to block this PR, but I feel like this `setupWithManagerFunc` pattern is getting a little awkward, especially now that we're adding an `appconfig` struct that is only used by one of the two variants.

cmd/crossplane/main.go
turkenh: Yes, good point! I'll try that.
Originally I've implemented `Option 2` [here](https://github.com/crossplaneio/crossplane/blob/3a2671c46a98a8a46fbe90ff6c1f8c7b5b0a688f/design/one-pager-host-aware-stack-manager.md#host-cluster-access),  and missed this while switching to `Option 1`.

cmd/crossplane/main.go
turkenh: Host cluster is typically expected to be shared across tenants, which means host controller namespace should have some unique identifier (e.g. `tenant-123-controllers`). So, I don't think it make sense to have a default here.

cmd/crossplane/main.go
turkenh: I refactored a bit which seems cleaner.

Makefile
```diff
@@ -81,7 +81,34 @@ fallthrough: submodules
 manifests:
 	@$(WARN) Deprecated. Please run make generate instead.
 
-generate: $(KUSTOMIZE) go.generate manifests.annotate
+generate: $(KUSTOMIZE) go.generate manifests.prepare manifests.annotate
```
turkenh: @jbw976 as discussed:
We'll need `make generate` to be run for helm chart being prepared appropriately. Could you check if this is already called during our release process?

turkenh: May be we should move this process to `make build` just like `values.yaml.tmpl -> values.yaml`.

cmd/crossplane/main.go
turkenh: Just realized another reason to get API server and port seperately. 
We need to set these parameters for crossplane deployment as well to configure it against tenant apiserver and we don't pass kubeconfig file to it at all: https://github.com/turkenh/crossplane/blob/host-aware-sm-onepager/design/one-pager-host-aware-stack-manager.md#configuring-a-pod-against-tenant-instance-api-server

I believe getting these two params separately would be better instead of passing kubeconfig to crossplane just for this purpose.

jbw976:
@turkenh is this implementation of host aware stack manager rebased with the latest in master and all merge conflicts resolved with template stack initial implementation? thanks!

turkenh:
> @turkenh is this implementation of host aware stack manager rebased with the latest in master and all merge conflicts resolved with template stack initial implementation? thanks!

@jbw976 yes, correct. 
And I could just able to pass quality gate with enough coverage :)

Makefile
turkenh: @jbw976 
AFAIU, `make generate` is not called per build. So, we should either:

a. move this process to make build like `values.yaml.tmpl` -> `values.yaml`
or
b. also commit generated files like we do for crd's already, i.e. despite CRD's are generated files, they are also committed.

I couldn't find the proper way to achieve option a, if there is, I would vote for it. Otherwise, option b would also be fine, except we will have duplicate files which will be overridden with `make generate`

Makefile
```diff
@@ -83,7 +83,34 @@ fallthrough: submodules
...
+
+CROSSPLANE_TYPES_CHART_DIR = cluster/charts/crossplane-types
+CROSSPLANE_CONTROLLERS_CHART_DIR = cluster/charts/crossplane-controllers
+CROSSPLANE_CHART_DIR = cluster/charts/crossplane
+
+CRD_DIR = $(CROSSPLANE_TYPES_CHART_DIR)/crds
+CROSSPLANE_CHART_HELM2_CRD_DIR = $(CROSSPLANE_CHART_DIR)/templates/crds
+CROSSPLANE_CHART_HELM3_CRD_DIR = $(CROSSPLANE_CHART_DIR)/crds
+
+TYPE_MANIFESTS = clusterrole.yaml clusterrolebinding.yaml crossplane-admin-clusterrole.yaml \
```
jbw976: would a `*.yaml` pattern work here instead of having to explicitly list each file name? i can see this being tricky to maintain or figure out in the future first that our helm chart is missing a new file then figuring out why.

apis/generate.go
```diff
@@ -20,13 +20,13 @@ limitations under the License.
 // https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module
 
 // Remove existing CRDs
-//go:generate rm -rf ../cluster/charts/crossplane/templates/crds
+//go:generate rm -rf ../cluster/charts/crossplane-types/crds
```
jbw976: should this clean up step not delete from both locations?

cluster/charts/crossplane-types/crds/workload.crossplane.io_kubernetestargets.yaml
```diff
@@ -0,0 +1,145 @@
+
```
jbw976: why is this one showing up as an add but the other ones are showing up as renames?

the other "renamed" CRDs will be committed in both places instead of being renamed after https://github.com/crossplaneio/crossplane/pull/1139/files#r371825341 is resolved, right?

pkg/controller/stacks/stack/stack.go
```diff
@@ -492,6 +555,136 @@ func (h *stackHandler) isNamespaced() bool {
...
+	saSecretOnHost := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:            toSecretRef.Name,
+			Namespace:       toSecretRef.Namespace,
+			OwnerReferences: []metav1.OwnerReference{owner},
+		},
+		Data: saSecret.Data,
+	}
+
+	err = toKube.Create(ctx, saSecretOnHost)
```
jbw976: any reason we'd want to do an update here if it already exists?

pkg/controller/stacks/install/stackinstall.go
```diff
@@ -273,6 +315,18 @@ func (h *stackInstallHandler) delete(ctx context.Context) (reconcile.Result, err
...
 	}
 
+	stackControllerNamespace := h.ext.GetNamespace()
+	if h.hostAwareConfig != nil {
+		stackControllerNamespace = h.hostAwareConfig.HostControllerNamespace
+	}
+
+	// Once the Stacks are gone, we can remove install job associated with the StackInstall using hostKube since jobs
+	// were deployed into host Kubernetes cluster.
+	if err := h.hostKube.DeleteAllOf(ctx, &batchv1.Job{}, client.MatchingLabels(labels),
```
jbw976: do we need to delete the deployments of the stack controllers as well? (or remind me on how they get cleaned up again)

EDIT: nevermind, found it below in stack.go's delete func :)

jbw976:
also note the minor conflicts coming from @displague recent PRs to remove some permissions from the *admin roles in #1165

jbw976:
hi @turkenh, just wanted to make sure you weren't waiting for me on anything right now in this PR. There were a few questions left to answer from my [last review](https://github.com/crossplaneio/crossplane/pull/1139#pullrequestreview-350554773), and I just wanted to make sure I wasn't blocking you in any way. Thanks for your ongoing effort on this!

pkg/controller/stacks/stack/stack.go
turkenh: Service account token secrets [does not change](https://github.com/kubernetes/kubernetes/issues/20165) once created (as long as corresponding SA is not deleted and recreated). So, I don't see a reason we need to update this. 
If Kubernetes somehow supports SA token secret rotation in the future, we need more changes to support this  because this code path (i.e. stack install) is only called once per stack during create.

cluster/charts/crossplane-types/crds/workload.crossplane.io_kubernetestargets.yaml
turkenh: This file was not available when I started this PR and introduced recently. I believe it is related to this as well as we were gitignoring main CRD dir with this PR (which will be changed with  https://github.com/crossplaneio/crossplane/pull/1139/files#r371825341).

> the other "renamed" CRDs will be committed in both places instead of being renamed after https://github.com/crossplaneio/crossplane/pull/1139/files#r371825341 is resolved, right?

Yes

Makefile
turkenh: Agreed, done.

apis/generate.go
turkenh: `crossplane-types/crds` is the main place where kubebuilder generates the CRD. So it deletes that directory before dumping CRDs. 
Copying to other locations happen as part of `make` so it make sense to delete them there. Updated makefile to cleanup those directories before copying which also resolves your concern about zombie files in comment: https://github.com/crossplaneio/crossplane/pull/1139/files#r372736521

Makefile
turkenh: Ok, I'll go with option b then.

Regarding zombie files, this should not be an issue, please see: https://github.com/crossplaneio/crossplane/pull/1139/files#r373966967

turkenh:
> hi @turkenh, just wanted to make sure you weren't waiting for me on anything right now in this PR. There were a few questions left to answer from my [last review](https://github.com/crossplaneio/crossplane/pull/1139#pullrequestreview-350554773), and I just wanted to make sure I wasn't blocking you in any way. Thanks for your ongoing effort on this!

@jbw976, thanks, yes, I was aware and just could not find enough time to work on this.

Now all your comments should be resolved and this PR should be ready for merge.

Makefile
turkenh: I also added a simple readme file as a note for working on helm charts.

PR #1144, opened 2019-12-30
1-pager for Resource High Level Stop-Light Status

rathpc:
This PR adds the initial 1-pager for Resource High Level Stop-Light Status.

As a stack author there is currently no way to author rules that combine arbitrary conditions of a given state to represent high level _"stop-light"_ conditions. This PR aims to resolve that with a new design.

### DIFF
--- a/design/one-pager-resource-status.md
+++ b/design/one-pager-resource-status.md
@@ -9,7 +9,7 @@
 As a stack author there is currently no way to author rules that combine arbitrary conditions of a given state to
 represent high level _"stop-light"_ conditions.
 
-The high level conditions would be:
+The high level conditions should be:
 - Online
   > _When a resource instance is fully available for use with no errors or failures_
 - Offline
@@ -98,14 +98,16 @@ bare minimum.
 
 The `states` key should follow these rules:
 
-- There are 3 nested keys which you can define:
+- There are 3 nested keys which you should define:
   > `Online`
 
   > `Offline`
 
   > `Warning`
 
-  - There is technically a 4th state (`Unknown`) but that is defined automatically as a default status catch.
+  - There is technically another state (`Unknown`) but that is defined automatically as a default status catch.
+
+  _You can define other states if you wish, but the 3 mentioned above are the **standard recommended states**_
 
 - Each key should contain an array of objects. Each object contains a collection of **AND** comparisons, and each object
 within the array is treated as **OR** comparisons to adjacent objects.
--- a/design/one-pager-resource-status.md
+++ b/design/one-pager-resource-status.md
@@ -58,18 +58,18 @@ resourceStatus:
       lastTransitionTime: .status.someRandomStatuses.SyncFailure.updatedTime
   states:
     Online:
-    - Ready: 'True'
-      Synchronized: 'True'
-      SyncFailure: 'False'
-    - Ready: 'True'
-      Synchronized: 'True'
+    - Ready: True
+      Synchronized: True
+      SyncFailure: False
+    - Ready: True
+      Synchronized: True
       SyncFailure: not 'True'
     Offline:
-    - Ready: 'False'
-      Synchronized: 'False'
-    - Ready: 'True'
+    - Ready: False
+      Synchronized: False
+    - Ready: True
       Synchronized: not 'True'
-      SyncFailure: 'True'
+      SyncFailure: True
     Warning:
     - Ready: 'False'
       Synchronized: 'True'
--- a/design/one-pager-resource-status.md
+++ b/design/one-pager-resource-status.md
@@ -11,12 +11,14 @@ represent high level _"stop-light"_ conditions.
 
 The high level conditions would be:
 - Online
+  > _When a resource instance is fully available for use with no errors or failures_
 - Offline
+  > _When a resource instance is **not** available for use, potentially with many errors and failures_
 - Warning
+  > _When a resource instance is **Online** but may have some errors or failures that do not affect its "Readiness"_
 - Unknown
-
-The **Unknown** condition is interesting because this would be returned if the conditions were annotated but we cannot
-verify a known state based on the given rules. _(Essentially a default case in a switch block)_.
+  > _This would be returned if the conditions were annotated but we cannot verify a known state based on the given rules._
+  > _(Essentially a default case in a switch block)_
 
 ## Design
 
--- a/design/one-pager-resource-status.md
+++ b/design/one-pager-resource-status.md
@@ -1,0 +1,133 @@
+# Resource High Level Stop-Light Status
+
+- Owner: Steven Rathbauer ([@rathpc](https://github.com/rathpc))
+- Reviewers: Crossplane Maintainers
+- Status: Draft
+
+## Problem
+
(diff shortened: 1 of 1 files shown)


design/one-pager-resource-status.md
```diff
@@ -0,0 +1,133 @@
...
+      )
+    ) {
+      // Do something for Online state
+    }
+    ```
+
+- The keys in the objects should relate to the condition types found in the `paths` key and the values should be
+whatever is required to meet that particular criteria.
+
+  - You can also define a logical not by typing `not` before the value you are comparing against so that you can
```
displague: I like that unset conditions (if `SyncFailure` was not defined), are not a problem because the ui-schema is under the same domain as the controller so all known states should be accounted for.  The ability to create custom `paths` using JSONPath allows for additional flexibility and conditional value remapping.

It may be worthwhile to include examples showing that `field: True` is still possible, without the single-quotes.  Yaml parsers will ignore/remove these unless `not ` is leading the text.

design/one-pager-resource-status.md
```diff
@@ -0,0 +1,133 @@
...
+      Synchronized: 'True'
+      SyncFailure: 'True'
+```
+
+<a name="example-2"></a>
+
+```yaml
+# Example 2
+resourceStatus:
+  paths:
```
displague: I am concerned about the parsing complexities of having `paths` be either a string or a map, mirrored here https://github.com/crossplaneio/crossplane/pull/1145#discussion_r362497437.

Perhaps `generatePaths` or `customPaths` could be used as a name for the map structure.

design/one-pager-resource-status.md
```diff
@@ -0,0 +1,133 @@
...
+    - Ready: 'True'
+      Synchronized: 'True'
+    Offline:
+    - Ready: 'False'
+      Synchronized: 'False'
+    Warning:
+    - Ready: 'False'
+      Synchronized: 'True'
+      SyncFailure: 'True'
+```
```
negz: What's the thinking behind defining separate matching logic for the offline and online states? Are there cases in which offline can't be inferred by online being false?

Also, do we want to align the terms used here with those that Crossplane is using? Most (all?) of our resources currently use "available and unavailable" rather than "online and offline" - https://godoc.org/github.com/crossplaneio/crossplane-runtime/apis/core/v1alpha1#Condition

rathpc: The idea is that the logic behind each case is entirely arbitrary and up to the stack author to decide. This is also why there are technically 4 states (Online, Offline, Warning and Unknown).

So technically there can be a case where offline is not specifically the direct negation of online and warning or unknown can also play a part in anything in between.

design/one-pager-resource-status.md
rathpc: As per a discussion we had, I think the best short term solution to this problem is to establish that the ui-schema file is explicitly formatted per OpenAPI specs which allows for "OR" conditions. This will also allow us to lint (to a degree) this file to prevent issues after publishing due to incorrectly formatted files.

design/one-pager-resource-status.md
rathpc: I can change some of the examples to not use quoted values to help illustrate that!

rathpc: Updated 👍

design/one-pager-resource-status.md
```diff
@@ -0,0 +1,133 @@
...
+As a stack author there is currently no way to author rules that combine arbitrary conditions of a given state to
+represent high level _"stop-light"_ conditions.
+
+The high level conditions would be:
+- Online
+- Offline
+- Warning
+- Unknown
+
+The **Unknown** condition is interesting because this would be returned if the conditions were annotated but we cannot
```
displague: Additional context on what it means to be Online, Offline, or in a Warning state would be helpful, especially because this context has been provided for Unknown.

While Online/Offline may appear obvious, there may be nuance in the description.  (This might even compel a rewording to Available/Unavailable).

rathpc: Just updated 👍

design/one-pager-resource-status.md
```diff
@@ -0,0 +1,135 @@
...
+
+There are two main keys within `resourceStatus`. The first is `paths` which can be a string or map of key/value pairs. Additionally the key/value pairs can be set as <string>: <string> **OR** <string>: <object>.
+
+To expand on how to set the `paths` key a bit:
+
+- If the **states** are using status types that can all be found within a single conditions array that follow the object
+structure of a standard [PodCondition](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#podcondition-v1-core),
+then you can just set the `paths` key to the JSONPath of that conditions array. (_[Example 1](#example-1)_)
+
+- If the states are using status types from various or arbitrary locations, then you can individually set each path as a
```
jbw976: the marshalling/unmarshalling for this might be a little odd since we're saying that the `paths` field can either be an array or it could be a map/dictionary.  golang doesn't handle this super smoothly that I know of (at least the code consuming this data will need to be aware of the various types it could be and do some checks to know how to proceed), but maybe it works better in javascript/typescript?

design/one-pager-resource-status.md
```diff
@@ -0,0 +1,135 @@
...
+- There are 3 nested keys which you can define:
+  > `Online`
+
+  > `Offline`
+
+  > `Warning`
+
+  - There is technically a 4th state (`Unknown`) but that is defined automatically as a default status catch.
+
+- Each key should contain an array of objects. Each object contains a collection of **AND** comparisons, and each object
```
jbw976: do we want to consider an experience that provides a little more flexibility in the logic? e.g. allow the user to express the comparisons themselves (and vs. or).

i think that this is probably the right trade-off between simplification of experience and expressiveness of logic, but just making sure with you again.

design/one-pager-resource-status.md
```diff
@@ -0,0 +1,135 @@
...
+
+- Lastly, if you want to be very explicit about the individual values of a particular status type, you can set JSONPaths
+for each individual condition key. (_[Example 2a](#example-2)_)
+
+- As a rule of thumb, you will need to define the `status`, `message` and `lastTransitionTime` for a path key, at the
+bare minimum.
+
+The `states` key should follow these rules:
+
+- There are 3 nested keys which you can define:
```
jbw976: since this is a general UI schema, do you think we should allow arbitrary states to be defined here too? if a UI doesn't know what a state means, it could always ignore it.

design/one-pager-resource-status.md
rathpc: Referring to resolution above: https://github.com/crossplaneio/crossplane/pull/1144#discussion_r363055081

We also hashed this out in person but we came to the consensus that this is very manageable in js/ts so we should be ok for now.

design/one-pager-resource-status.md
rathpc: we decided that this should be sufficient for now and an acceptable trade-off

design/one-pager-resource-status.md
rathpc: I will be updating the language to address this

PR #1145, opened 2019-12-30
Update 1-pager for Stack Status Spec

rathpc:
This updates resourceConditionPaths --> printerColumns and related changes

Related PRs to make followup changes to stack repos:
https://github.com/crossplaneio/sample-stack/pull/5
https://github.com/crossplaneio/stack-aws/pull/93
https://github.com/crossplaneio/stack-azure/pull/86
https://github.com/crossplaneio/stack-gcp/pull/124
https://github.com/crossplaneio/stack-rook/pull/34

### DIFF
--- a/design/one-pager-stack-status-spec.md
+++ b/design/one-pager-stack-status-spec.md
@@ -1,8 +1,8 @@
 # Stack UI Schema Status Spec
 
-- Owner: Steven Rathbauer (@rathpc)
+- Owner: Steven Rathbauer ([@rathpc](https://github.com/rathpc))
 - Reviewers: Crossplane Maintainers
-- Status: Draft
+- Status: Draft, revision 1.0
 
 ## Proposal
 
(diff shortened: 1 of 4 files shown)


rathpc:
> Really love this improvement, thanks @rathpc! One thing I am wondering is if we can actually formalize the conditions as code and generate these `ui-schema` files as part of code generation. This would allow stack builders to annotate the ui-schema printer columns in code similar to how we do with k8s printer columns. It makes me somewhat uncomfortable to be writing these by hand, and I recently experienced some confusion as to how the files should be formatted, etc. when adding a new type to `stack-gcp`. Don't think we need to block on this, but wanted to get your thoughts :)

Thanks Dan! I think I follow but let's chat later this week about it 👍

design/one-pager-stack-status-spec.md
```diff
@@ -1,96 +1,85 @@
...
 
 This 1-pager will identify the key that should be added and what that structure looks like.
 
 ## Design
 
-#### Add two new keys to the `ui-schema.yaml` file:
+#### New key to be added to the `ui-schema.yaml` file:
 
-- `resourceConditionPaths` _(this is an array of JSONPaths)_
+- `printerColumns` _(this is a string or map of key/value pairs)_
```
displague: For code generation or usage in typed languages, there could be a benefit to keeping this variable defined as a map and perhaps using a difference variable for the string interpretation.

That said, I don't know of any doctrine that recommends avoiding variable types.  Was there an existing pattern you were pairing this with? It feels (Kubernetes) familiar, but I can't place it.

rathpc: This is adjacent to this PR which uses a similar pattern https://github.com/crossplaneio/crossplane/pull/1144

displague: Discussed offline and settled with https://github.com/crossplaneio/crossplane/pull/1144/files#r362861161

PR #1169, opened 2020-01-14
docs: include kubernetesapplication description in the stacks guide

displague:
# Description of your changes

Adds more context to the Stacks Guide about the KubernetesApplication resource being created by the Wordpress Stack.

The additional text here helps to explain the mention of `KubernetesApplication` that immediately follows these changes:

> `# Get the name of the Kubernetes Application resource which is the top level workload definiton for our Wordpress instance`

Related to #1168 

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/docs/stacks-guide.md
+++ b/docs/stacks-guide.md
@@ -199,11 +199,26 @@ If the control cluster doesn't recognize the Wordpress instance type, it
 could be because the stack is still being installed. Wait a few seconds,
 and try creating the Wordpress instance again.
 
-### Wait
+### Wait for Claims
 
-The Wordpress can take a while to spin up, because behind the scenes
-Crossplane is creating all of its dependendencies, which is a database
-and Kubernetes cluster. To check the status, you can use [the trace 
+The Wordpress instance can take a while to spin up, because behind the scenes
+the Crossplane Stack is creating all of its dependendencies using Crossplane resources.
+
+The resources we are waiting for are:
+
+* `MySQLInstance` - a claim for a MySQL database, that could be provisioned by CloudSQL, RDS, or Azure Database, for example
+* `KubernetesCluster` - a claim for a Kubernetes cluster that could be provided by GKE, AKS, EKS, or others
+* `KubernetesApplication` - a workload to deploy on a Kubernetes cluster
+
+When the database and cluster claims have become available, their `Ready` condition will become `Available`
+and they will provide a reference to a secret with credentials for the provisioned service.
+
+The powerful `KubernetesApplication` Crossplane resource bundles native Kubernetes resource types and manages
+them on a target cluster. The `KubernetesApplication` deployed by this Wordpress Stack includes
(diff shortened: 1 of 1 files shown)


docs/stacks-guide.md
```diff
@@ -199,11 +199,26 @@ If the control cluster doesn't recognize the Wordpress instance type, it
...
 The Wordpress can take a while to spin up, because behind the scenes
-Crossplane is creating all of its dependendencies, which is a database
-and Kubernetes cluster. To check the status, you can use [the trace 
+the Crossplane Stack is creating all of its dependendencies using claim resources.
+
+The claims we are waiting for are:
+
+* `MySQLInstance` - a MySQL database, that could be provisioned by CloudSQL, RDS, or Azure Database, for example
+* `KubernetesCluster` - a claim for a Kubernetes cluster that could be provided by GKE, AKS, EKS, or others
+* `KubernetesApplication` - a claim for a Kubernetes application workload to deploy on a Kubernetes cluster
```
hasheddan: This is a bit of a nitpick, but do we usually refer to a `KubernetesApplication` as a claim type? I don't necessarily think it is wrong, and I like how you refer to it as a "powerful claim" below, but I just don't usually think of it as a claim type. This could totally just be my own mental model getting in the way here so please feel free to push back :)

docs/stacks-guide.md
```diff
@@ -199,11 +199,26 @@ If the control cluster doesn't recognize the Wordpress instance type, it
 could be because the stack is still being installed. Wait a few seconds,
 and try creating the Wordpress instance again.
 
-### Wait
+### Wait for Claims
 
 The Wordpress can take a while to spin up, because behind the scenes
-Crossplane is creating all of its dependendencies, which is a database
-and Kubernetes cluster. To check the status, you can use [the trace 
+the Crossplane Stack is creating all of its dependendencies using claim resources.
```
displague: "claim resources" may be better as "resource claims"

docs/stacks-guide.md
negz: I think I'd prefer to avoid referring to KA as a claim - it feels like it risks overloading the term.

negz:
General feedback for documenting `KubernetesApplication`

* We consider it to be a "kind of workload" (it's in the workload API group). We imagine there will be other kinds of workload eventually (`ServerlessApplication`, maybe `VMApplication`).
* I see `KubernetesApplication` more as a low level bundle of resources that would be produced by some higher level resource (a CR installed by a Stack?) rather than as something users will be interacting with or authoring directly, at least eventually.

docs/stacks-guide.md
displague: Revised here..
https://github.com/crossplaneio/crossplane/compare/0f9555d026b17907b2ab4959c88333c03afbfbd0..c498c764f3b9c33f5fa2e3772e16984f5fd1e50d

PR #1177, opened 2020-01-17
api docs: generate template stacks api reference

hasheddan:

Generates template stacks API docs using https://github.com/negz/crossdocs.

N/A

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/docs/api/crossplaneio/crossplane/stacks-crossplane-io-v1alpha1.md
+++ b/docs/api/crossplaneio/crossplane/stacks-crossplane-io-v1alpha1.md
@@ -6,6 +6,7 @@ This API group contains the following Crossplane resources:
 
 * [ClusterStackInstall](#ClusterStackInstall)
 * [Stack](#Stack)
+* [StackConfiguration](#StackConfiguration)
 * [StackInstall](#StackInstall)
 
 ## ClusterStackInstall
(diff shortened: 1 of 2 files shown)


docs/api/crossplaneio/crossplane/stacks-crossplane-io-v1alpha1.md
```diff
@@ -178,6 +220,95 @@ Name | Type | Description
...
+StackConfigurationBehaviors specifies behaviors for the stack
+
+Appears in:
+
+* [StackConfigurationSpec](#StackConfigurationSpec)
+
+
+Name | Type | Description
+-----|------|------------
+`crds` | [map[github.com/crossplaneio/crossplane/apis/stacks/v1alpha1.GVK]github.com/crossplaneio/crossplane/apis/stacks/v1alpha1.StackConfigurationBehavior](#StackConfigurationBehavior) |
```
suskin: Is this right? It seems like it should be links to the `GVK` alias above and to `StackConfigurationBehavior`. I'm noticing that the GVK alias also doesn't have any `Appears in:`, and I'm noticing that the map in `StackConfigurationBehavior.hooks` also looks funny.

hasheddan: Looks like `crossdocs` doesn't know how to handle a map of other structs :disappointed: options here are:
1. Just go with this current state and we can assume that experimental docs are okay for experimental features.
2. Exclude these docs for the time being (issue here is that when we incorporate automatically generating the docs there will be a diff until this is fixed, but we could also just cross that bridge when we come to it).
3. Fix crossdocs, which would maybe delay this release (not recommended).

jbw976: my vote would be #1, keep the docs as is so that the new type is at least included (all alpha types should be included), but the docs have a small content bug in them.  I'm OK with that for now.

PR #1192, opened 2020-01-23
Existing one-pager updates and new implementation reference inclusion

rathpc:
This PR updates some information within the stack ui metadata one pager as well as introduces a new design doc authored by [Upbound](https://github.com/upbound) which outlines one specific implementation of the `ui-schema.yaml` file that is meant to serve as a helpful reference to other potential implementors in the future

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/design/design-doc-upbound-ui-spec.md
+++ b/design/design-doc-upbound-ui-spec.md
@@ -1,0 +1,276 @@
+# Upbound `ui-schema.yaml` Specification
+
+* Owner: Steven Rathbauer ([@rathpc](https://github.com/rathpc))
+* Reviewers: Upbound Maintainers, Crossplane Maintainers
+* Status: Accepted
+
+## Terms
+
+* **Stack**: A Stack is a Crossplane stack-manager managed package. These packages may bundle one or many Crossplane
+applications or infrastructure provider controllers.
+* **Stacks**: Stacks could refer to more than one Crossplane stack-manager packages or the Stack-Manager system as a whole.
+
+See the [Stacks Design Doc](design-doc-stacks.md) for more details about Crossplane Stacks.
+
+## Objective
+
+This document outlines one specific implementation of the `ui-schema.yaml` file that is meant to serve as a helpful
+reference to other potential implementors in the future. This specification is driven by what is currently supported via
+the Upbound GUI and meant to be used as the ground work for further UI development using this file.
+
+## Proposed File format
+
+OpenAPI v3
+
+## Supported Fields
+
+### Required Fields
+
+Field | Description
+--- | ---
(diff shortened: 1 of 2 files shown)


rathpc:
@displague, I will do my best to address your concerns/questions:

> I continue to question whether we will ultimately run into problems with dual types (boolean or object in this case)

I think this is a valid question and will likely be answered as soon as we define a validator/validation system for this file or more specifically implementations of this file since the base formal specification is unbiased and only required to be under a certain size and valid YAML.

> type` seems HTML inspired. If comparing this to HTML5, visibility (hidden) and readonly properties would not prevent the use of data type properties (hidden read-only integer, for example).

It is definitely HTML inspired but I am not sure I follow your statement exactly. Are you asking if we can support the usage of multiple types?

If so, currently no but that I can see as being valuable, especially combining `readonly` with another input type.

> Do you see the eventual need for more HTML5 input types - email, date, etc?

Sure, I don't see why we couldn't or shouldn't allow those in the future. Currently they are not supported by the UI however.

> Defaults must be provided as a string and "," is used for multiple defaults. Why not allow this to be an array in that case? (which goes against my mixed type concerns but matches the practices of the design)

That is a good point. Currently the UI logic is the limitation for this, but that code can be updated to support this as an array instead in the near future.

> As in other specs, I think the common parameters could be extracted into a "common input properties" section of the design.

I was contemplating that, however most of the language even around the "common" properties is written to be specific to that control type. Additionally it felt easier to see a control types properties as a whole rather than having to refer to multiple sections of the doc to get the whole picture.

displague:
> Are you asking if we can support the usage of multiple types?

The definition of "types" is what I am getting at. Are "readonly" and "hidden" types that should compete with "int", "string", "email", or "tele", or are they independent features that should use their own properties. `<input type="tele" readonly style="display:none">`

rathpc:
> The definition of "types" is what I am getting at. Are "readonly" and "hidden" _types_ that should compete with "int", "string", "email", or "tele", or are they independent features that should use their own properties. `<input type="tele" readonly style="display:none">`

Well, hidden is a valid input type so I am not sure that really competes and while readonly is an attribute, it can be argued that if something is readonly, does it really matter if the type is set to text or something more specific. I think at this time, the necessity to separate HTML input types from input attributes is not dire but can certainly be a consideration for the future.

displague:
We may want to revisit this section of the UI Schema format to account for this one example of a specification.

 https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-stack-ui-metadata.md#formal-specification

rathpc:
> We may want to revisit this section of the UI Schema format to account for this one example of a specification.
> 
> https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-stack-ui-metadata.md#formal-specification

This PR includes changes to that as well, feel free to suggest any changes you see fit there too. 👍

Issue #1174, opened 2020-01-17
definition of done: experimental features


suskin:
### What problem are you facing?

In #1131, we worked on creating a first version of the shared template stack controller, in an experimental state. As we reviewed and discussed the changes, we realized our [existing frameworks for determining whether a changeset is ready for merging](https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-definition-of-done.md) didn't quite cover the situation we were in, because we were intentionally merging something which is experimental and which must be explicitly enabled by someone who knows they want to use it.

### How could Crossplane help solve your problem?

We should think more about how we want to handle new, unproven, experimental features or systems that we're developing. Some considerations would include:

* What's the definition of done?
* How do we get them merged without them potentially affecting existing functionality?
* How do we explain experimental things to users, to set expectations properly?
* What sort of test coverage do we want in this type of situation?

PR #1217, opened 2020-01-30
Unpack creates StackDefinition resources from Template Stacks behavior.yaml files

### LINKED ISSUE #1210: template stacks: Unpacking should create a TemplateStack object instead of Stack if it's a template stack

Currently, stack manager unpacking process creates all CRDs and `Stack` object. What we do in template stacks is that we reconcile a `TemplateStack` object and generate a `Stack` object accordingly.

We need unpacking process to see that the stack source is a template stack and only create the TemplateStack CR. How it works today is that `config/stack/app.yaml` and `config/stack/install.yaml` is merged to create a `Stack` object with additional information from CRDs.

What we need is that we can have a file that defines the `Behaviors` of `TemplateStack` and instead of `Stack`, we should generate a `TemplateStack` object. As an initial step, we can have a file called `behaviors.yaml` to define `TemplateStack.spec.behaviors` object and treat the stack as template stack if that file exists and do the necessary operations.

The structure of `TemplateStack` is pretty similar to `Stack` [with a few changes](https://github.com/crossplaneio/crossplane/issues/1209) but I believe it should be possible to use some of the existing mechanisms that creates `Stack` objects using those files in the bundle.

There is also an [issue](https://github.com/crossplaneio/crossplane/issues/1178) merging `app.yaml` and `install.yaml` into one `stack.yaml` that lives in the root of the directory. However, how that `stack.yaml` will look like is not certain yet. So, I'd go with what we have today to implement this issue.

displague:

The Stack Manager is taught how to handle `behavior.yaml` files adjacent to `app.yaml`.

In the presence of this file, the Stack-Manager Unpack process will emit a `StackDefinition` resource rather than a `Stack` resource.  Since these template stacks do not (and should not) include an `install.yaml`, the controller in the StackDefinition resource will be configured as `crossplane/templating-controller` with parameters understood by that controller. This image is configurable as a Helm value.

* [templating-controller](https://github.com/crossplaneio/templating-controller/) should be tagged `0.1.0` and pushed to DockerHub to match this PR

The StackDefinition controller has been modified to include an additional RBAC permission in the Stack that it emits.  This added rule enables the templating-controller to access the StackDefinition resource since the behavior metadata is stored in that resource.

Fixes #1210 

https://github.com/muvaf/stack-minimal-gcp/tree/template-stacks is successfully deployed using:

```
kubectl crossplane stack install --cluster --namespace crossplane-system crossplane/stack-minimal-gcp:0.0.2 stack-minimal-gcp
```
### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/apis/stacks/v1alpha1/stackdefinition.go
+++ b/apis/stacks/v1alpha1/stackdefinition.go
@@ -19,6 +19,8 @@ package v1alpha1
 import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	"github.com/crossplaneio/crossplane-runtime/pkg/meta"
 )
 
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
@@ -120,12 +122,7 @@ func (sd *StackDefinition) DeepCopyIntoStack(s *Stack) {
 	sd.Spec.CRDs.DeepCopyInto(&s.Spec.CRDs)
 	sd.Spec.Controller.DeepCopyInto(&s.Spec.Controller)
 	sd.Spec.Permissions.DeepCopyInto(&s.Spec.Permissions)
-	if s.Labels == nil {
-		s.Labels = map[string]string{}
-	}
-	for key, val := range sd.Labels {
-		s.Labels[key] = val
-	}
+	meta.AddLabels(s, sd.GetLabels())
 }
 
 // DeepCopyIntoStackDefinition copies a Stack to a StackDefinition
@@ -135,5 +132,5 @@ func (s *Stack) DeepCopyIntoStackDefinition(sd *StackDefinition) {
 	s.Spec.CRDs.DeepCopyInto(&sd.Spec.CRDs)
 	s.Spec.Controller.DeepCopyInto(&sd.Spec.Controller)
 	s.Spec.Permissions.DeepCopyInto(&sd.Spec.Permissions)
-	sd.ObjectMeta.SetLabels(s.ObjectMeta.GetLabels())
+	meta.AddLabels(sd, s.GetLabels())
 }
--- a/apis/stacks/v1alpha1/stackdefinition.go
+++ b/apis/stacks/v1alpha1/stackdefinition.go
@@ -113,3 +113,27 @@ type StackDefinitionList struct {
 	metav1.ListMeta `json:"metadata,omitempty"`
 	Items           []StackDefinition `json:"items"`
 }
+
+// DeepCopyIntoStack copies a StackDefinition to a Stack
+func (sd *StackDefinition) DeepCopyIntoStack(s *Stack) {
+	sd.Spec.AppMetadataSpec.DeepCopyInto(&s.Spec.AppMetadataSpec)
+	sd.Spec.CRDs.DeepCopyInto(&s.Spec.CRDs)
+	sd.Spec.Controller.DeepCopyInto(&s.Spec.Controller)
+	sd.Spec.Permissions.DeepCopyInto(&s.Spec.Permissions)
+	if s.Labels == nil {
+		s.Labels = map[string]string{}
+	}
+	for key, val := range sd.Labels {
+		s.Labels[key] = val
+	}
+}
+
+// DeepCopyIntoStackDefinition copies a Stack to a StackDefinition
+// TODO(displague) should this reside in types.go with the Stack type
+func (s *Stack) DeepCopyIntoStackDefinition(sd *StackDefinition) {
+	s.Spec.AppMetadataSpec.DeepCopyInto(&sd.Spec.AppMetadataSpec)
+	s.Spec.CRDs.DeepCopyInto(&sd.Spec.CRDs)
+	s.Spec.Controller.DeepCopyInto(&sd.Spec.Controller)
+	s.Spec.Permissions.DeepCopyInto(&sd.Spec.Permissions)
+	sd.ObjectMeta.SetLabels(s.ObjectMeta.GetLabels())
+}
--- a/cluster/charts/crossplane/README.md
+++ b/cluster/charts/crossplane/README.md
@@ -211,6 +211,7 @@ The following tables lists the configurable parameters of the Crossplane chart a
 | `clusterStacks.rook.version`     | Rook stack version to deploy                                    | `<latest released version>`   
 | `personas.deploy`                | Install roles and bindings for Crossplane user personas         | `true`     
 | `templateStacks.enabled`         | Enable experimental template stacks support                     | `false`     
+| `templateStacks.controllerImage` | Template Stack controller image                                 | `crossplane/templating-controller:0.1.0`
  
 ### Command Line
 
--- a/cluster/charts/crossplane/templates/stack-manager-deployment.yaml
+++ b/cluster/charts/crossplane/templates/stack-manager-deployment.yaml
@@ -35,6 +35,10 @@ spec:
         {{- end }}
         {{- if .Values.templateStacks.enabled }}
         - --templates
+        {{- if .Values.templateStacks.controllerImage }}
+        - --templating-controller-image
+        - {{ .Values.templateStacks.controllerImage | quote }}
+        {{- end }}
         {{- end }}
         {{- range $arg := .Values.args }}
         - {{ $arg }}
--- a/cluster/charts/crossplane/values.yaml.tmpl
+++ b/cluster/charts/crossplane/values.yaml.tmpl
@@ -35,3 +35,4 @@ hostedConfig:
 
 templateStacks:
   enabled: false
+  controllerImage: crossplane/templating-controller:0.1.0
--- a/cmd/crossplane/main.go
+++ b/cmd/crossplane/main.go
@@ -52,21 +52,27 @@ func main() {
 		// permissions, the SM itself must have cluster-admin permissions. We
 		// isolate these elevated permissions as much as possible by running the
 		// Crossplane stack manager in its own isolated deployment.
-		extManageCmd                     = extCmd.Command("manage", "Start Crosplane Stack Manager controllers")
-		extManageSupportTemplates        = extManageCmd.Flag("templates", "Enable support for template stacks").Bool()
+		extManageCmd                 = extCmd.Command("manage", "Start Crosplane Stack Manager controllers")
+		extManageTemplates           = extManageCmd.Flag("templates", "Enable support for template stacks").Bool()
+		extManageTemplatesController = extManageCmd.Flag("templating-controller-image", "The image of the Template Stacks controller (implies --templates)").Default("").String()
+
 		extManageHostControllerNamespace = extManageCmd.Flag("host-controller-namespace", "The namespace on Host Cluster where install and controller jobs/deployments will be created. Setting this will activate host aware mode of Stack Manager").String()
 		extManageTenantKubeconfig        = extManageCmd.Flag("tenant-kubeconfig", "The absolute path of the kubeconfig file to Tenant Kubernetes instance (required for host aware mode, ignored otherwise).").ExistingFile()
 
 		// Unpack the given stack package content. This command is expected to
 		// parse the content and generate manifests for stack related artifacts
 		// to stdout so that the SM can read the output and use the Kubernetes
-		// API to create the artifacts. Users are not expected to run this
-		// command themselves, the stack manager itself should execute this
-		// command.
-		extUnpackCmd             = extCmd.Command("unpack", "Unpack a Stack").Alias("unstack")
-		extUnpackDir             = extUnpackCmd.Flag("content-dir", "The absolute path of the directory that contains the stack contents").Required().String()
-		extUnpackOutfile         = extUnpackCmd.Flag("outfile", "The file where the YAML Stack record and CRD artifacts will be written").String()
-		extUnpackPermissionScope = extUnpackCmd.Flag("permission-scope", "The permission-scope that the stack must request (Namespaced, Cluster)").Default("Namespaced").String()
+		// API to create the artifacts.
+		//
+		// Users are not expected to run this command themselves, the stack
+		// manager itself should execute this command.
+		//
+		// Unpack does not interact with the Kubernetes API.
+		extUnpackCmd                 = extCmd.Command("unpack", "Unpack a Stack").Alias("unstack")
+		extUnpackDir                 = extUnpackCmd.Flag("content-dir", "The absolute path of the directory that contains the stack contents").Required().String()
+		extUnpackOutfile             = extUnpackCmd.Flag("outfile", "The file where the YAML Stack record and CRD artifacts will be written").String()
+		extUnpackPermissionScope     = extUnpackCmd.Flag("permission-scope", "The permission-scope that the stack must request (Namespaced, Cluster)").Default("Namespaced").String()
+		extUnpackTemplatesController = extUnpackCmd.Flag("templating-controller-image", "The image of the Template Stacks controller").Default("").String()
 	)
 	cmd := kingpin.MustParse(app.Parse(os.Args[1:]))
 
@@ -108,9 +114,17 @@ func main() {
 
 		kingpin.FatalIfError(apis.AddToScheme(mgr.GetScheme()), "Cannot add core Crossplane APIs to scheme")
 		kingpin.FatalIfError(apiextensionsv1beta1.AddToScheme(mgr.GetScheme()), "Cannot add API extensions to scheme")
-		kingpin.FatalIfError(stacks.Setup(mgr, log, *extManageHostControllerNamespace), "Cannot add stacks controllers to manager")
+		kingpin.FatalIfError(stacks.Setup(mgr, log, *extManageHostControllerNamespace, *extManageTemplatesController), "Cannot add stacks controllers to manager")
 
-		if *extManageSupportTemplates {
+		if *extManageTemplatesController != "" {
+			*extManageTemplates = true
+		}
+
+		if *extManageTemplates {
+			if *extManageTemplatesController == "" {
+				kingpin.Fatalf("--templating-controller-image is required with --templates")
+			}
+
 			kingpin.FatalIfError(templates.SetupStackConfigurations(mgr, log), "Cannot add stack configuration controller to manager")
 			kingpin.FatalIfError(templates.SetupStackDefinitions(mgr, log), "Cannot add stack definition controller to manager")
 		}
@@ -132,7 +146,7 @@ func main() {
 		// TODO(displague) afero.NewBasePathFs could avoid the need to track Base
 		fs := afero.NewOsFs()
 		rd := &walker.ResourceDir{Base: filepath.Clean(*extUnpackDir), Walker: afero.Afero{Fs: fs}}
-		kingpin.FatalIfError(stack.Unpack(rd, outFile, rd.Base, *extUnpackPermissionScope), "failed to unpack stacks")
+		kingpin.FatalIfError(stack.Unpack(rd, outFile, rd.Base, *extUnpackPermissionScope, *extUnpackTemplatesController), "failed to unpack stacks")
 
 	default:
 		kingpin.FatalUsage("unknown command %s", cmd)
--- a/docs/install-crossplane.md
+++ b/docs/install-crossplane.md
@@ -269,6 +269,7 @@ The following tables lists the configurable parameters of the Crossplane chart a
 | `clusterStacks.rook.version`     | Rook stack version to deploy                                    | `<latest released version>`   
 | `personas.deploy`                | Install roles and bindings for Crossplane user personas         | `true`     
 | `templateStacks.enabled`         | Enable experimental template stacks support                     | `false`     
+| `templateStacks.controllerImage` | Template Stack controller image                                 | `crossplane/templating-controller:0.1.0`
  
 ### Command Line
 
--- a/pkg/controller/stacks/install/installjob.go
+++ b/pkg/controller/stacks/install/installjob.go
@@ -32,7 +32,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/yaml"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
@@ -51,7 +50,7 @@ var (
 
 // JobCompleter is an interface for handling job completion
 type jobCompleter interface {
-	handleJobCompletion(ctx context.Context, i v1alpha1.StackInstaller, job *batchv1.Job) error
+	handleJobCompletion(ctx context.Context, i stacks.KindlyIdentifier, job *batchv1.Job) error
 }
 
 // StackInstallJobCompleter is a concrete implementation of the jobCompleter interface
@@ -62,7 +61,7 @@ type stackInstallJobCompleter struct {
 	log          logging.Logger
 }
 
-func createInstallJob(i v1alpha1.StackInstaller, executorInfo *stacks.ExecutorInfo, hCfg *hosted.Config) *batchv1.Job {
+func createInstallJob(i v1alpha1.StackInstaller, executorInfo *stacks.ExecutorInfo, hCfg *hosted.Config, tscImage string) *batchv1.Job {
 	name := i.GetName()
 	namespace := i.GetNamespace()
 
@@ -86,7 +85,7 @@ func createInstallJob(i v1alpha1.StackInstaller, executorInfo *stacks.ExecutorIn
 					RestartPolicy: corev1.RestartPolicyNever,
 					InitContainers: []corev1.Container{
 						{
-							Name:    "stack-package",
+							Name:    "stack-copy-to-volume",
 							Image:   i.Image(),
 							Command: []string{"cp", "-R", registryDirName, "/ext-pkg/"},
 							VolumeMounts: []corev1.VolumeMount{
@@ -99,7 +98,7 @@ func createInstallJob(i v1alpha1.StackInstaller, executorInfo *stacks.ExecutorIn
 					},
 					Containers: []corev1.Container{
 						{
-							Name:  "stack-executor",
+							Name:  "stack-unpack-and-output",
 							Image: executorInfo.Image,
 							// "--debug" can be added to this list of Args to get debug output from the job,
 							// but note that will be included in the stdout from the pod, which makes it
@@ -109,6 +108,7 @@ func createInstallJob(i v1alpha1.StackInstaller, executorInfo *stacks.ExecutorIn
 								"unpack",
 								fmt.Sprintf("--content-dir=%s", filepath.Join("/ext-pkg", registryDirName)),
 								"--permission-scope=" + i.PermissionScope(),
+								"--templating-controller-image=" + tscImage,
 							},
 							VolumeMounts: []corev1.VolumeMount{
 								{
@@ -132,9 +132,7 @@ func createInstallJob(i v1alpha1.StackInstaller, executorInfo *stacks.ExecutorIn
 	}
 }
 
-func (jc *stackInstallJobCompleter) handleJobCompletion(ctx context.Context, i v1alpha1.StackInstaller, job *batchv1.Job) error {
-	var stackRecord *v1alpha1.Stack
-
+func (jc *stackInstallJobCompleter) handleJobCompletion(ctx context.Context, i stacks.KindlyIdentifier, job *batchv1.Job) error {
 	// find the pod associated with the given job
 	podName, err := jc.findPodNameForJob(ctx, job)
 	if err != nil {
@@ -163,29 +161,7 @@ func (jc *stackInstallJobCompleter) handleJobCompletion(ctx context.Context, i v
 		if err := jc.createJobOutputObject(ctx, obj, i, job); err != nil {
 			return err
 		}
-
-		if isStackObject(obj) {
-			// we just created the stack record, try to fetch it now so that it can be returned
-			stackRecord = &v1alpha1.Stack{}
-			n := types.NamespacedName{Name: obj.GetName(), Namespace: obj.GetNamespace()}
-			if err := jc.client.Get(ctx, n, stackRecord); err != nil {
-				return errors.Wrapf(err, "failed to retrieve created stack record %s/%s from job %s", obj.GetNamespace(), obj.GetName(), job.Name)
-			}
-		}
 	}
-
-	if stackRecord == nil {
-		return errors.Errorf("failed to find a stack record from job %s", job.Name)
-	}
-
-	// save a reference to the stack record in the status of the stack install
-	i.SetStackRecord(&corev1.ObjectReference{
-		APIVersion: stackRecord.APIVersion,
-		Kind:       stackRecord.Kind,
-		Name:       stackRecord.Name,
-		Namespace:  stackRecord.Namespace,
-		UID:        stackRecord.ObjectMeta.UID,
-	})
 
 	return nil
 }
@@ -233,9 +209,9 @@ func (jc *stackInstallJobCompleter) readPodLogs(namespace, name string) (*bytes.
 	return b, nil
 }
 
-// createJobOutputObject names, labels, sets ownership, and creates resources
-// resulting from a StackInstall or ClusterStackInstall. These expected
-// resources are currently CRD and Stack objects
+// createJobOutputObject names, labels, and creates resources in the API
+// Expected resources are CRD, Stack, StackDefinition, & StackConfiguration
+// nolint:gocyclo
 func (jc *stackInstallJobCompleter) createJobOutputObject(ctx context.Context, obj *unstructured.Unstructured,
 	i stacks.KindlyIdentifier, job *batchv1.Job) error {
 
@@ -247,12 +223,20 @@ func (jc *stackInstallJobCompleter) createJobOutputObject(ctx context.Context, o
 	// when the current object is a Stack object, make sure the name and namespace are
 	// set to match the current StackInstall (if they haven't already been set). Also,
 	// set the owner reference of the Stack to be the StackInstall.
-	if isStackObject(obj) || isStackConfigurationObject(obj) {
+	if isStackObject(obj) || isStackDefinitionObject(obj) || isStackConfigurationObject(obj) {
 		if obj.GetName() == "" {
 			obj.SetName(i.GetName())
 		}
 		if obj.GetNamespace() == "" {
 			obj.SetNamespace(i.GetNamespace())
+		}
+	}
+
+	// StackDefinition controllers need the name of the StackDefinition
+	// which, by design, matches the StackInstall
+	if isStackDefinitionObject(obj) {
+		if err := setStackDefinitionControllerEnv(obj, i.GetNamespace(), i.GetName()); err != nil {
+			return err
 		}
 	}
 
@@ -301,6 +285,17 @@ func isStackObject(obj *unstructured.Unstructured) bool {
 		strings.EqualFold(gvk.Kind, v1alpha1.StackKind)
 }
 
+func isStackDefinitionObject(obj *unstructured.Unstructured) bool {
+	if obj == nil {
+		return false
+	}
+
+	gvk := obj.GroupVersionKind()
+
+	return gvk.Group == v1alpha1.Group && gvk.Version == v1alpha1.Version &&
+		strings.EqualFold(gvk.Kind, v1alpha1.StackDefinitionKind)
+}
+
 func isStackConfigurationObject(obj *unstructured.Unstructured) bool {
 	if obj == nil {
 		return false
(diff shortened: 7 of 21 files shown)


pkg/stacks/unpack.go
```diff
@@ -219,12 +211,24 @@ func (sp *StackPackage) IsNamespaced() bool {
...
-	sp.Behavior = behavior
+// SetStack sets the Stack's configuration
+// This is primarily for defining Template Stack behaviors, but is intended
+// to replace SetApp in a merger of TemplateStack and Stack
+func (sp *StackPackage) SetStack(tss v1alpha1.TemplateStackSpec) {
+	tss.AppMetadataSpec.DeepCopyInto(&sp.Stack.Spec.AppMetadataSpec)
+
+	sp.TemplateStack.Spec.StackConfigurationSpec = tss.StackConfigurationSpec
+
+	deployment := appsv1.Deployment{}
```
suskin: For generating this deployment, we'll need to inspect the template stack / stack definition to either:

* Pass the namespaced name of the stack definition into the controller, OR
* Pass the GVK of the CRD and the namespaced name of the stack definition into the controller

Also, just making a note that we were imagining setting an owner reference for the stack which points back to the stack definition (though I've got some code which does that in #1214)

displague:
@suskin @muvaf see https://gist.github.com/displague/ee4621d2d25ac03af3af10a12ccf018b for the current state of this branch.

There are some TODO's in the code where I'm not sure what values to use.  I'm awaiting changes in #1214 which will require a rebase in this PR and variable renaming / refactoring.

pkg/stacks/unpack.go
displague: Closing this as resolved since this code has changed. Please resurface this concern if it is still applies.

pkg/controller/stacks/install/installjob.go
```diff
@@ -235,7 +235,7 @@ func (jc *stackInstallJobCompleter) createJobOutputObject(ctx context.Context, o
 	// when the current object is a Stack object, make sure the name and namespace are
 	// set to match the current StackInstall (if they haven't already been set). Also,
 	// set the owner reference of the Stack to be the StackInstall.
-	if isStackObject(obj) || isStackConfigurationObject(obj) {
+	if isStackObject(obj) || isStackDefinitionObject(obj) || isStackConfigurationObject(obj) {
```
muvaf: `handleJobCompletion` also needs some care to accommodate `stackdefinition` output.

pkg/stacks/unpack.go
```diff
@@ -218,13 +238,97 @@ func (sp *StackPackage) IsNamespaced() bool {
...
+// This is primarily for defining Template Stack behaviors
+func (sp *StackPackage) SetBehavior(sd v1alpha1.StackDefinitionBehavior) {
+	sd.DeepCopyInto(&sp.StackDefinition.Spec.Behavior)
+
+	// Set the Stack deployment, which will be copied to StackDefinition in
+	// Yaml()
+	// TODO(displague) what controller are we running?
+	spec := sp.createBehaviorController(sd)
+	sp.Stack.Spec.Controller.Deployment = &v1alpha1.ControllerDeployment{
+		Spec: spec,
```
muvaf: I'm getting `failed to create deployment: Deployment.apps "" is invalid: [metadata.name: Required value: name or generateName is required` I think you need to fill `Name`, too. At least that's what `(sp *StackPackage) SetInstall(install unstructured.Unstructured)` function does.

pkg/stacks/unpack.go
```diff
@@ -218,13 +238,97 @@ func (sp *StackPackage) IsNamespaced() bool {
...
+	behaviorContentsVolumeName := "behaviors"
+	behaviorMountPoint := "/behaviors"
+	spec := appsv1.DeploymentSpec{
+		Selector: &metav1.LabelSelector{MatchLabels: map[string]string{"app": controllerDeploymentName}},
+		Template: corev1.PodTemplateSpec{
+			ObjectMeta: metav1.ObjectMeta{
+				Labels: map[string]string{"app": controllerDeploymentName},
+			},
+			Spec: corev1.PodSpec{
+				RestartPolicy: corev1.RestartPolicyNever,
```
muvaf: I'm getting an error saying `Never` is not supported. I believe it's because that's a `Deployment`, not a `Job`, hence it requires user to put a container image that's _restartable_

pkg/stacks/unpack.go
```diff
@@ -218,13 +238,97 @@ func (sp *StackPackage) IsNamespaced() bool {
...
+				InitContainers: []corev1.Container{
+					{
+						Name:    "stack-package",
+						Image:   sd.Source.Image,
+						Command: []string{"cp", "-xR", behaviorDirName, behaviorMountPoint},
```
muvaf: In alpine, I get `cp: unrecognized option: x`. Is there any difference if you use `cp -a`?

pkg/stacks/unpack.go
```diff
@@ -218,13 +238,97 @@ func (sp *StackPackage) IsNamespaced() bool {
...
+	// TODO(displague) where does the name come from? behavior? cmd line arg?
+	// sp.Stack.GetName() is not available when this runs
+	controllerDeploymentName := "tc-" + "something"
+	controllerContainerName := controllerDeploymentName
+	controllerImageName := "bymafmaf/templating-controller:0.0.1"
+
+	// TODO(displague) Where do these files reside within the Stack?
+	// are sp.baseDir or sd.Source.Path useful here?
+	// are we always mounting a remote source image?
+	behaviorDirName := "/"
```
muvaf: Answering the question in comment, yes, `sd.Source.Path` exists for this.

pkg/stacks/unpack.go
```diff
@@ -218,13 +238,97 @@ func (sp *StackPackage) IsNamespaced() bool {
...
+					},
+				},
+				Containers: []corev1.Container{
+					{
+						Name:  controllerContainerName,
+						Image: controllerImageName,
+						Args: []string{
+							"manager",
+							"--resources-dir", behaviorMountPoint,
+							"--stack-definition-namespace", sp.Stack.GetNamespace(),
```
muvaf: These flags are not filled. Looking at your comment above, it seems because `sp.Stack` is empty at this step.

pkg/controller/stacks/install/installjob.go
displague: I've moved the special StackInstall updating from handleJobCompletion to the StackInstall sync function.

pkg/stacks/unpack.go
displague: Because I had a root path of `/` I was trying to avoid copying `/proc`, `/sys`, `/dev`, `/tmp`, etc into the mounted volume.  `-x` keeps copy in a single filesystem.

It is a weakness of the current stack packaging system that Stacks must contain a shell at all, and here we see that `cp` within that environment must be a compatible with our expectations.

We can avoid this problem by denying `/` as an image path.

pkg/stacks/unpack.go
displague: `unpack` does not know the name of the StackInstall / Stack / etc.  This is why the StackManager currently applies the name.  This is only becoming a problem in my PR because the name has to be injected in a slice of command line arguments instead of in a well known location (deployment name, template name, stack name, etc).

`unpack.go:NewStackPackage(baseDir string)` could take a name parameter, which it could acquire through `--stack-name` and `--stack-namespace` arguments to `unpack` (set via environment, or we could use a configmap).

pkg/stacks/unpack.go
muvaf: Yeah, I think it makes sense just to go with `cp -a`, as far as I see `-a` is compatible with all variants of `cp`. A check for `/` could be helpful before the struct definition.

On another note, I agree that relying on a shell is not the best state to be in but I don't have much solution to overcome this, a `cp` operation needs to be made. Maybe when [CSI drivers ](https://kubernetes-csi.github.io/docs/ephemeral-local-volumes.html)are widely supported, we'd just use the image as a volume for the generic controller and don't need to copy anything.

pkg/controller/stacks/install/stackinstall.go
```diff
@@ -167,22 +168,52 @@ func (f *handlerFactory) newHandler(ctx context.Context, ext v1alpha1.StackInsta
...
+			h.ext.SetStackRecord(&corev1.ObjectReference{
+				APIVersion: s.APIVersion,
+				Kind:       s.Kind,
+				Name:       s.Name,
+				Namespace:  s.Namespace,
+				UID:        s.ObjectMeta.UID,
+			})
+			h.ext.SetConditions(runtimev1alpha1.Available(), runtimev1alpha1.ReconcileSuccess())
+
+			return requeueOnSuccess, h.kube.Status().Update(ctx, h.ext)
```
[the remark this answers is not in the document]
displague: Unless there is a fatal configuration, we always want the Stacks and StackInstalls controllers to be requeued so that our `Reconcile` and `sync()` functions get another opportunity to act on the resource since in both cases they are dependent on outside resources that are not event or watch driven.

https://github.com/crossplaneio/crossplane/blob/18ebbb0b800e6b70886809997dc4378cc9379ded/pkg/controller/stacks/stack/stack.go#L75-L76

_`requeueAfterOnSuccess` is a duration of 10s_

Reconcile Result is defined upstream: https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/reconcile/reconcile.go#L25-L33

From a simpler review premise, this is what were already doing here, https://github.com/crossplaneio/crossplane/pull/1217/files#diff-a2cd537f6531321fbef9a92d96b2c425L272 where the StackInstall asserted that a completed InstallJob was a complete StackInstall.  Now that the StackInstall has to update the StackRef itself, this call moved around.

I suppose https://github.com/crossplaneio/crossplane/pull/1217/files#diff-a2cd537f6531321fbef9a92d96b2c425R306 could be requeued without a timer (`resultRequeue`), since a complete installjob is no longer a complete stackinstall.  This could save up to 10s.

I would want to confirm this and see other examples of Crossplane returning reconcile results like this in a partial success condition before making that change.

pkg/stacks/steps.go
```diff
@@ -78,16 +78,29 @@ func appStep(sp StackPackager) walker.Step {
...
+		sp.SetStackConfig(config)
+		return nil
+	}
+}
+
+// behaviorStep unmarshals stack.yaml bytes
+func behaviorStep(sp StackPackager) walker.Step {
+	return func(path string, b []byte) error {
+		behavior := v1alpha1.StackDefinitionBehavior{}
+		if err := yaml.Unmarshal(b, &behavior); err != nil {
```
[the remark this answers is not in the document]
displague: If the file is missing, this step will not be called. The path is only for context information (used by other steps where the filename adds details about what CRD is being described).

If the file were empty (or providing alternate yaml documents), the behavior object would be default/empty, but it will still be known.  We could add validation in `behaviorStep` to prevent illogical configurations.

If the contents were not yaml conformant and error would be returned.

pkg/stacks/unpack.go
displague: > Then the name of the Stack/StackDefinition doesn't matter, and it can be generated by the same code which is generating the StackDefinition

A drawback of this is that the StackInstall and Stack names would not match, but in considering that StackInstalls could in the future lead to multiple Stack records, the names wouldn't have the luxury of such simple naming for ever. I defer to https://github.com/crossplaneio/crossplane/issues/1089#issuecomment-571814987 for thoughts how to approach that problem.

pkg/stacks/unpack.go
displague: I don't think `cp -a` would work. `-a` tries to set the ownership and mode on the target filesystem.  In Docker build environments I've frequently been snagged by images trying to `chown` or `chmod` volume mounted (or in container) files that the runtime user (non-root) is not permitted to affect or the target volume does not support.

I'll remove the `-x` in anycase, returning the `cp` to its original form and I'll add proper image path usage and validation.

pkg/stacks/unpack.go
```diff
@@ -218,13 +248,89 @@ func (sp *StackPackage) IsNamespaced() bool {
...
+	sp.StackConfig = config
+}
+
+func (sp *StackPackage) createBehaviorController(sd v1alpha1.StackDefinitionBehavior) appsv1.DeploymentSpec {
+	controllerContainerName := "stack-behavior-manager"
+
+	// TODO(displague) Where do these files reside within the Stack?
+	// are sp.baseDir or sd.Source.Path useful here?
+	// are we always mounting a remote source image?
+	behaviorDirName := "/"
```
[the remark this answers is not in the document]
suskin: > I think /.registry is good for metadata stuff but not much for actual resources

I agree - I am not the biggest fan of it as the root folder for real resources. **But** - we have been treating it as the root folder for resources from the perspective of the stacks specification, so I think it makes sense to stick with it for now until we change the spec.

For example, consider a stack which has a git repo as a source. The root of the repo is the root of the stack, and `stack.yaml`/`app.yaml` would live in the root of the repo in the simplest hierarchy. And the resource files would also live in the root of the repo. If you make the image filesystem equivalent to the git repo "filesystem", then everything should be inside the root of the stack filesystem, which would be `/.registry` with today's implementation.

pkg/controller/stacks/install/installjob.go
```diff
@@ -82,11 +84,15 @@ func createInstallJob(i v1alpha1.StackInstaller, executorInfo *stacks.ExecutorIn
 			BackoffLimit: &jobBackoff,
 			Template: corev1.PodTemplateSpec{
 				Spec: corev1.PodSpec{
-					RestartPolicy: corev1.RestartPolicyNever,
+					AutomountServiceAccountToken: &passServiceAccount,
```
displague: Chatted with @suskin and @muvaf about this, we agree that the lockdown of the `initContainers` and unpack should introduced in a separate PR.

docs/install-crossplane.md
```diff
@@ -269,6 +269,7 @@ The following tables lists the configurable parameters of the Crossplane chart a
 | `clusterStacks.rook.version`     | Rook stack version to deploy                                    | `<latest released version>`   
 | `personas.deploy`                | Install roles and bindings for Crossplane user personas         | `true`     
 | `templateStacks.enabled`         | Enable experimental template stacks support                     | `false`     
+| `templateStacks.controllerImage` | Template Stack controller image                                 | `bymafmaf/templating-controller:0.0.1`
```
suskin: Would the intended default be `crossplane/resourcepacks:master` or something at the time that this is merged? Making a note in case that's intended, so it's easier to remember

pkg/controller/stacks/install/installjob.go
suskin: Yes - we aren't opposed to doing it, but moving it into a separate PR helps us reduce the reviewing and testing overhead, and the potential blast radius, by making more smaller PRs. Which is a benefit that we like about smaller PRs.

cmd/crossplane/main.go
```diff
@@ -113,14 +119,22 @@ func main() {
...
 
-		log.Info("Adding controllers")
-		kingpin.FatalIfError(stacksControllerSetupWithManager(mgr, *extManageHostControllerNamespace), "Cannot add controllers to manager")
+		if *extManageTemplatesController != "" {
+			*extManageTemplates = true
+		}
+
+		if *extManageTemplates {
+			if *extManageTemplatesController == "" {
+				kingpin.Fatalf("--templates-controller is required with --templates")
```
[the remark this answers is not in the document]
displague: The helm chart will ship a default. The version of that controller will likely need frequent updating so I would rather see it configured in a chart (or other yaml) rather than code.

If we had a good image location that wasn't from a personal repo, I would be more inclined to consider hard coding it (in .go)
.

cluster/charts/crossplane/README.md
```diff
@@ -209,7 +209,9 @@ The following tables lists the configurable parameters of the Crossplane chart a
 | `clusterStacks.azure.version`    | Azure stack version to deploy                                   | `<latest released version>`   
 | `clusterStacks.rook.deploy`      | Deploy Rook stack                                               | `false`    
 | `clusterStacks.rook.version`     | Rook stack version to deploy                                    | `<latest released version>`   
-| `personas.deploy`                | Install roles and bindings for Crossplane user personas        | `true`     
+| `personas.deploy`                | Install roles and bindings for Crossplane user personas         | `true`     
+| `templateStacks.enabled`         | Enable experimental template stacks support                     | `false`     
+| `templateStacks.controllerImage` | Template Stack controller image                                 | `bymafmaf/templating-controller:0.0.1`
```
displague:
```suggestion
| `templateStacks.controllerImage` | Template Stack controller image                                 | `crossplane/templating-controller:0.1.0`
```

(I'm experimenting with the Github batching of suggestions)

cluster/charts/crossplane/templates/stack-manager-deployment.yaml
```diff
@@ -36,6 +36,10 @@ spec:
         {{- if .Values.templateStacks.enabled }}
         - --templates
         {{- end }}
+        {{- if .Values.templateStacks.controllerImage }}
```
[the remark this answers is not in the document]
displague: `make generate`
produces: `cluster/charts/crossplane/templates/stack-manager-deployment.yaml`
from: `cluster/charts/crossplane-controllers/templates/stack-manager-deployment.yaml`

cluster/charts/crossplane/values.yaml.tmpl
```diff
@@ -35,3 +35,5 @@ hostedConfig:
 
 templateStacks:
   enabled: false
+  # TODO(displague) dont merge until this can point at crossplane/resourcepacks:master
```
displague:
```suggest
  controllerImage: crossplane/templating-controller:0.1.0
```

cluster/charts/crossplane/values.yaml.tmpl
```diff
@@ -35,3 +35,5 @@ hostedConfig:
...
+  controllerImage: bymafmaf/templating-controller:0.0.1
```
displague:
```suggestion
  controllerImage: crossplane/templating-controller:0.1.0
```

cluster/charts/crossplane/templates/stack-manager-deployment.yaml
```diff
@@ -35,6 +35,10 @@ spec:
         {{- end }}
         {{- if .Values.templateStacks.enabled }}
         - --templates
+        {{- if .Values.templateStacks.controllerImage }}
+        - --templates-controller
```
displague:
```suggestion
        - --templating-controller-image
```

pkg/controller/stacks/install/installjob.go
```diff
@@ -328,3 +331,49 @@ func isCRDObject(obj runtime.Object) bool {
...
+	if u, err := convertToUnstructured(sd); err == nil {
+		u.DeepCopyInto(obj)
+	}
+
+	return err
+}
+
+// convertToUnstructured takes a Kubernetes object and converts it into
+// *unstructured.Unstructured that can be used as KubernetesApplication template.
+func convertToUnstructured(o *v1alpha1.StackDefinition) (*unstructured.Unstructured, error) {
```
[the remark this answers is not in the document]
displague: Maybe. I'm beginning to think my middle name is `cyclomatic complexity`, that's what Jenkins and `golangci-lint` call me when they are mad.

pkg/controller/stacks/install/installjob.go
```diff
@@ -328,3 +331,49 @@ func isCRDObject(obj runtime.Object) bool {
...
+
+	return &unstructured.Unstructured{Object: u}, nil
+}
+
+// convertToStackDefinition takes a Kubernetes object and converts it into
+// *v1alpha1.StackDefinition
+func convertToStackDefinition(o *unstructured.Unstructured) (*v1alpha1.StackDefinition, error) {
+	sd := &v1alpha1.StackDefinition{}
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(o.UnstructuredContent(), sd); err != nil {
+		return &v1alpha1.StackDefinition{}, err
```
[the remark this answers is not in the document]
displague: I agree. I think I borrowed these functions as-is from resourcepacks or stack-minimal-gcp, before applying your awesome `scheme.Convert`'less approach.

I'll change these to nil.

pkg/controller/stacks/install/stackinstall.go
```diff
@@ -199,35 +204,66 @@ func (f *handlerFactory) newHandler(ctx context.Context, ext v1alpha1.StackInsta
...
+		templatesControllerImage: templatesControllerImage,
 	}
 }
 
 // ************************************************************************************************
 // Syncing/Creating functions
 // ************************************************************************************************
 func (h *stackInstallHandler) sync(ctx context.Context) (reconcile.Result, error) {
-	if h.ext.StackRecord() == nil {
+	stack := h.ext.StackRecord()
```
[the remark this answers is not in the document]
displague: I don't see `stack` defined, but I do see `stacks`.  I'll rename this to `sr` to avoid potential confusion since `stack` is imported elsewhere.

pkg/controller/stacks/install/stackinstall.go
```diff
@@ -297,8 +337,22 @@ func (h *stackInstallHandler) update(ctx context.Context) (reconcile.Result, err
 // This function ensures that all the resources (e.g., CRDs) that this StackInstall owns
 // are also cleaned up.
 func (h *stackInstallHandler) delete(ctx context.Context) (reconcile.Result, error) {
-	// Delete all Stacks created by this StackInstall or ClusterStackInstall
 	labels := stacks.ParentLabels(h.ext)
+
+	// Delete all StackDefintions created by this StackInstall or
+	// ClusterStackInstall
+	if err := h.kube.DeleteAllOf(ctx, &v1alpha1.StackDefinition{}, client.InNamespace(h.ext.GetNamespace()), client.MatchingLabels(labels)); runtimeresource.IgnoreNotFound(err) != nil {
```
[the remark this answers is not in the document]
displague: I was worried about some template stacks behavior leaking into users that don't have `--templates` enabled.  I think you might be right about this causing problems.

The templates enabled flag may need to be plumbed deeply into the controllers to gate statements like this. :slightly_frowning_face: 

I'll try it out.

pkg/controller/stacks/install/stackinstall_test.go
```diff
@@ -82,6 +83,9 @@ func withInstallJob(jobRef *corev1.ObjectReference) resourceModifier {
 	return func(r v1alpha1.StackInstaller) { r.SetInstallJob(jobRef) }
 }
 
+// TODO(displague) this should be used in a test that asserts stacksinstalls
```
displague:
```suggestion
// TODO(displague) this should be used in a test that asserts stackinstalls
```

pkg/controller/stacks/templates/stackdefinition_controller.go
```diff
@@ -35,8 +36,9 @@ import (
 // StackDefinitions. At that time, the Stack controller logic will likely just be pointed
 // over at StackDefinitions.
 type StackDefinitionReconciler struct {
-	Client client.Client
-	Log    logr.Logger
+	Client          client.Client
+	Log             logr.Logger
+	ControllerImage string
```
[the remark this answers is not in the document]
displague: I had started plumbing `ControllerImage` into the StackDefinition controller before realizing that it was, oddly enough, not involved in generating the controller image and deployment (in this current design, `unpack`, stackinstall, and stack share that responsibility).  The StackDefinition controller does add an RBAC rule, but that's the extent of its deployment manipulation.

As you have pointed out elsewhere, this creates a dependency on the `StackInstall` controller that preventing users from deploying  standalone `StackDefinition` resources.

(TODO: turn this observation into an issue)

pkg/stacks/unpack.go
```diff
@@ -159,52 +173,52 @@ type StackPackage struct {
...
+
+	// tmplCtrlImage is the Template Controller image to handle template stacks
+	tmplCtrlImage string
 }
 
 // Yaml returns a multiple document YAML representation of the Stack Package
 // This YAML includes the Stack itself and and all CRDs managed by that Stack.
 func (sp *StackPackage) Yaml() (string, error) {
-	var builder strings.Builder
+	var err error
```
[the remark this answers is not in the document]
displague: It doesn't appear to be, I think this was a remnant of attempting to reuse some other variable before I extracted `writeYaml` from `Yaml`.  I'll remove this and change all the `err` assignments in this function to `:=`.

pkg/controller/stacks/install/installjob.go
```diff
@@ -246,7 +229,7 @@ func (jc *stackInstallJobCompleter) createJobOutputObject(ctx context.Context, o
 	// when the current object is a Stack object, make sure the name and namespace are
 	// set to match the current StackInstall (if they haven't already been set). Also,
 	// set the owner reference of the Stack to be the StackInstall.
-	if isStackObject(obj) || isStackConfigurationObject(obj) {
```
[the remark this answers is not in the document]
displague: I lost code coverage when I switched to the group based match. I'll leave this block alone for now. We intend to remove the `StackConfiguration` type soon, so this condition will simplify then.

pkg/controller/stacks/install/stackinstall.go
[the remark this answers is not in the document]
displague: StackDefinition is unconditionally added to the Stack-Manager:
https://github.com/crossplaneio/crossplane/blob/6c95310778574ac012046acdcc497da4b2a6bc39/apis/stacks/v1alpha1/register.go#L75-L81


I think we are safe here.

apis/stacks/v1alpha1/stackdefinition.go
```diff
@@ -113,3 +113,27 @@ type StackDefinitionList struct {
...
+}
+
+// DeepCopyIntoStackDefinition copies a Stack to a StackDefinition
+// TODO(displague) should this reside in types.go with the Stack type
+func (s *Stack) DeepCopyIntoStackDefinition(sd *StackDefinition) {
+	s.Spec.AppMetadataSpec.DeepCopyInto(&sd.Spec.AppMetadataSpec)
+	s.Spec.CRDs.DeepCopyInto(&sd.Spec.CRDs)
+	s.Spec.Controller.DeepCopyInto(&sd.Spec.Controller)
+	s.Spec.Permissions.DeepCopyInto(&sd.Spec.Permissions)
+	sd.ObjectMeta.SetLabels(s.ObjectMeta.GetLabels())
```
suskin: If we are copying labels in both deep copy methods, it seems like we may want to copy them using the same approach. The other method uses individual copying, whereas this one uses replacement.

docs/install-crossplane.md
```diff
@@ -269,6 +269,7 @@ The following tables lists the configurable parameters of the Crossplane chart a
 | `clusterStacks.rook.version`     | Rook stack version to deploy                                    | `<latest released version>`   
 | `personas.deploy`                | Install roles and bindings for Crossplane user personas         | `true`     
 | `templateStacks.enabled`         | Enable experimental template stacks support                     | `false`     
+| `templateStacks.controllerImage` | Template Stack controller image                                 | `crossplane/templating-controller:0.1.0`
```
suskin: I didn't realize this table lived in more than one place o:

pkg/controller/stacks/install/installjob.go
```diff
@@ -328,3 +331,49 @@ func isCRDObject(obj runtime.Object) bool {
...
+		Value: name,
+	}}
+
+	// use convert functions because SetUnstructuredContent is unwieldy
+	sd, err := convertToStackDefinition(obj)
+	if err != nil {
+		return err
+	}
+
+	ts := &sd.Spec.Controller.Deployment.Spec.Template.Spec
```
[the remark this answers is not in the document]
suskin: Could `Spec.Controller` be `nil` too?

pkg/controller/stacks/install/stackinstall_test.go
```diff
@@ -49,6 +49,7 @@ const (
 	uid                     = types.UID(uidString)
 	resourceName            = "cool-stackinstall"
 	stackPackageImage       = "cool/stack-package:rad"
+	tsControllerImage       = "crossplane/ts-controller:0.0.0"
```
suskin: Will this ever pull a real image? Would it make sense to use an image which is more obviously fake? At first I thought it was supposed to be a real image version.

pkg/controller/stacks/stack/stack.go
```diff
@@ -670,30 +670,53 @@ func (h *stackHandler) prepareHostAwareJob(j *batch.Job, tokenSecret string) err
...
+func (h *stackHandler) prepareDeployment() *apps.Deployment {
 	controllerDeployment := h.ext.Spec.Controller.Deployment
 	if controllerDeployment == nil {
 		return nil
 	}
 
-	// ensure the deployment is set to use this stack's service account that we created
+	// force the deployment to use stack opinionated names and service account
+	name := h.ext.Name + "-controller"
+	matchLabels := map[string]string{"app": name}
```
suskin: What are we using this [selector](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#deploymentspec-v1-apps) for?

pkg/controller/stacks/stack/stack.go
```diff
@@ -685,30 +685,53 @@ func (h *stackHandler) prepareHostAwareJob(j *batch.Job, tokenSecret string) err
...
+			Name:      name,
 			Namespace: h.ext.Namespace,
 			Labels:    labels,
 		},
 		Spec: deploymentSpec,
 	}
+}
+
+func (h *stackHandler) processDeployment(ctx context.Context) error {
+	d := h.prepareDeployment()
```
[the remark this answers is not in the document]
suskin: It seems like it's related to [separation of concerns](https://en.wikipedia.org/wiki/Separation_of_concerns) . . . in this case, `processDeployment` doesn't really care where the deployment comes from, so it probably shouldn't be calling into the method to generate the deployment. It would make it simpler in the future if we ever want to pass parameters into `prepareDeployment` from the code which is calling `processDeployment`

pkg/controller/stacks/stack/stack.go
```diff
@@ -670,30 +670,53 @@ func (h *stackHandler) prepareHostAwareJob(j *batch.Job, tokenSecret string) err
...
+
+	if d == nil {
+		return nil
+	}
+
+	gvk := schema.GroupVersionKind{
+		Group:   apps.GroupName,
+		Kind:    "Deployment",
```
suskin: Nitpick: Does `d` have a `Kind` we could use instead of a hard-coded string?

pkg/controller/stacks/templates/stackdefinition_controller.go
```diff
@@ -79,24 +83,25 @@ func (r *StackDefinitionReconciler) Reconcile(req ctrl.Request) (ctrl.Result, er
 }
 
 func (r *StackDefinitionReconciler) createStack(ctx context.Context, sd *v1alpha1.StackDefinition, s *v1alpha1.Stack) error {
-	deepCopyStackDefinitionToStack(sd, s)
+	sd.DeepCopyIntoStack(s)
 	s.SetOwnerReferences([]metav1.OwnerReference{
 		*metav1.NewControllerRef(sd, v1alpha1.StackDefinitionGroupVersionKind),
 	})
+
+	rule := rbacv1.PolicyRule{
```
suskin: Awesome - thanks! :squid:

pkg/controller/stacks/stack/stack.go
displague: The selector is required by the Deployment spec and without it the `StackDefinition` resource  (yes, that) is not valid and is rejected on create.

displague: I think I was answering the wrong question here.  I needed to include a selector in the emitted `StackDefinition` resource.

And later when the Stack controller finally assembles the deployment, the deployment must have a selector to identify the pods that it is responsible for.

pkg/controller/stacks/install/stackinstall_test.go
displague: It would never pull a real image. I can change this to something more obviously fake.

docs/install-crossplane.md
displague: See #1109 and #1239

apis/stacks/v1alpha1/stackdefinition.go
displague: I pulled that change in using a Github review `suggestion`.  I overlooked this. I'll fix this because as the name implies (`Into`), the labels should be merged not clobbered.

pkg/controller/stacks/stack/stack.go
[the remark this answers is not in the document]
displague: I see how this could be modified to offer this.

`processDeployment` would send `prepareDeployment()` a bare `&appsv1.Deployment{}` which `prepareDeployment` would apply the `spec` using `DeepCopyInto`.

The `SetName` and `SetNamespace` would be used and the original labels would be merged with the parent labels instead of clobbered.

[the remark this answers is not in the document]
displague: As far as making `processDeployment` take parameters, that would warrant an out of scope refactor of the Stack controller's `create` function: https://github.com/crossplaneio/crossplane/blob/4534223ff95e7fb1f3ff559a792d65f49d0b855e/pkg/controller/stacks/stack/stack.go#L197-L229

apis/stacks/v1alpha1/stackdefinition.go
displague: I just discovered `meta.AddLabels` in `crossplane-runtime` :dancer:

jbw976:
> templating-controller should be tagged 0.1.0 and pushed to DockerHub to match this PR

Done!

jbw976:
None of the items in the PR checklist have been checked off.  I really want us to get out of the habit of ignoring them and actually do them and check them off.  For example, `make reviewable`.

pkg/controller/stacks/stack/stack.go
```diff
@@ -734,6 +748,8 @@ func (h *stackHandler) processJob(ctx context.Context) error {
 	// ensure the job is set to use this stack's service account that we created
 	jobSpec := *controllerJob.Spec.DeepCopy()
 	jobSpec.Template.Spec.ServiceAccountName = h.ext.Name
+	name := h.ext.Name + "-job"
```
jbw976: i've been wondering if we need to keep maintaining this "job" (instead of deployment) support. i don't think it's really used anywhere at all, it was sort of theoretical when we first added it, right? something to think about more, i keep questioning it every time i see PRs having to match a change to this code path too, which increases our maintenance burden for functionality that I don't think is used at all 🤔

pkg/stacks/unpack.go
```diff
@@ -156,52 +170,50 @@ type StackPackage struct {
...
-	}
+	if sp.GotBehavior() {
+		sp.Stack.DeepCopyIntoStackDefinition(&sp.StackDefinition)
 
-	// Behaviors are optional, so we skip it if it doesn't exist
-	if len(sp.Behavior.Object) > 0 {
-		bb, err := yaml.Marshal(sp.Behavior.Object)
-		if err != nil {
-			return "", errors.Wrap(err, "could not marshal behavior")
+		// New Format, using 'behavior.yaml'
```
jbw976: wondering if we should be versioning our stack package format now, so it's clear to a parser which format they should expect to be parsing, instead of figuring it out more heuristically/implicitly like we are doing here.

pkg/stacks/unpack.go
```diff
@@ -44,6 +45,15 @@ const (
 	groupFileName           = "group.yaml"
 	appFileName             = "app.yaml"
 	stackFileName           = "stack.yaml"
+	behaviorFileName        = "behavior.yaml"
```
jbw976: i'm getting confused between app.yaml, stack.yaml, and behavior.yaml also now...it would be really nice to have the [stack package format "spec"](https://github.com/crossplaneio/crossplane/blob/master/design/design-doc-stacks.md#stack-package-format) updated with explanations of all these different files.

Similar to our stack CRD situation right now, some of these files may be transient/temporary, which only causes more confusion as well.  The process around "experimental" features has continued to make me feel somewhat uncomfortable, with really low code coverage, temporary types and code, etc.  I'm definitely looking forward to discussing and solidifying as a community what our process is here that we are all comfortable with in #1174. 

/cc @suskin @muvaf @negz

pkg/controller/stacks/stack/stack.go
displague: Stacks have moved away from the possible solutions jobs support was made to provide. I'll create an issue to remove them.

displague: The issue is #1249

pkg/controller/stacks/install/installjob.go
```diff
@@ -328,3 +331,49 @@ func isCRDObject(obj runtime.Object) bool {
...
+	ts.Containers[0].Env = append(ts.Containers[0].Env, env...)
```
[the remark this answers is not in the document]
muvaf: Containers have `Name` and since we're the ones generating the whole deployment section for template stacks, we know that name. We can filter the containers by `Name` of the one we want and add the env vars to make sure we're adding them to the correct one.

pkg/controller/stacks/install/installjob.go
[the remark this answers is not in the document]
muvaf: @displague just a general style suggestion, it's easier to follow the code if you check for error and return early instead of putting the logic inside the if block. See https://danp.net/2015/11/02/reducing-go-nesting.html

PR #1221, opened 2020-01-31
Rehome pod util functions from crossplane-runtime

negz:

https://github.com/crossplaneio/crossplane-runtime/issues/1

Per the above issue the util package is deprecated. As best I can tell the stacks package is the only user of these functions, so I suggest they live where they're used.

### Checklist

I have:
  appropriate.

[definition of done].


negz:
> Die, util.. Die.

It's German for "The, util, the".

PR #1226, opened 2020-01-31
Update Resource Status Designs

rathpc:
This eliminates issues with typing and various expected values. By
doing this we can consume this in languages that are strongly typed
and may not be able to handle OR conditions easily.

#### Related PRs
- https://github.com/crossplaneio/sample-stack/pull/6
- https://github.com/crossplaneio/stack-aws/pull/125
- https://github.com/crossplaneio/stack-azure/pull/106
- https://github.com/crossplaneio/stack-gcp/pull/160
- https://github.com/crossplaneio/stack-minimal-gcp/pull/10
- https://github.com/crossplaneio/stack-rook/pull/40


rathpc:
I just pushed some updates to this...

The main one is related to the structure of the `resourceStatus` key and how you can specify entire arrays or objects while optionally defining them as a whole more explicitly or with overrides.

The second large change is to remove the Stack status design as it is not needed or relevant to any work and essentially started as a misinformed version of what the resource status design has become.

I will be linking related PR's to remove the additional references in any other repos.

rathpc:
#### Related PRs
- https://github.com/crossplaneio/sample-stack/pull/6
- https://github.com/crossplaneio/stack-aws/pull/125
- https://github.com/crossplaneio/stack-azure/pull/106
- https://github.com/crossplaneio/stack-gcp/pull/160
- https://github.com/crossplaneio/stack-minimal-gcp/pull/10
- https://github.com/crossplaneio/stack-rook/pull/40

thephred:
So it seems that in the goal to avoid the structure of the schema from changing we were mixing user supplied values with system defined values.  For instance
```
paths:
  JSONPath: '...'
```
vs.
```
paths:
  Ready:
    JSONPath: '...'
```
Means that not only does the keys in `paths` sometimes contain a explicit value (`JSONPath`) but sometimes user supplied values like `Ready`.
To combat this I suggest that the user supplied values go under a key beside the `JSONPath` key. This both provides more consistency of structure and avoids system key name conflicts with user supplied values.

PR #1227, opened 2020-02-03
Pass loggers down from main.go

### LINKED ISSUE #331: Errors and failures should be logged
**Is this a bug report or feature request?**
* Feature Request

**What should the feature do:** Some of the Crossplane controllers do not log anything at all when there are errors or failures.  We do a good job of capturing errors in status conditions on CRDs, but there is sometimes zero logging to help troubleshoot an issue.  We should take a pass through the controllers and log when any errors or failures happen.

**What is use case behind this feature:** It greatly helps troubleshooting and helping out our users when they can simply share the crossplane pod logs that will show all failures that occurred.  Currently, we have to help them find the correct object to look up status conditions on and often this is a chain of objects, e.g. bucket claim -> concrete bucket -> provider object.  This requires a lot of back and forth with the user and it's not intuitive for them to figure out on their own.

See the following Slack conversation for a great example of this:
https://crossplane.slack.com/archives/CEF5N8X08/p1551900157039900

I understand avoiding log bloat, but especially in the early phases of a project we need more consolidated logging of errors to streamline troubleshooting with our users and to allow them to self-service debug as well.  We can always dial back logging later on as the project gets more stable.

negz:

This PR updates Crossplane to use the logging library introduced in https://github.com/crossplaneio/crossplane-runtime/pull/108. It follows the [recommended practice](https://dave.cheney.net/2017/01/23/the-package-level-logger-anti-pattern) of plumbing loggers down to the types that use them rather than using package level loggers. It also uses shorter controller names, like `stacks/stackconfiguration`, or `workload/kubernetesapplication`. This fixes https://github.com/crossplaneio/crossplane/issues/388, and brings the Crossplane controllers into line with the pattern now used by the infrastructure stack controllers.

I've tried to avoid too many changes to the actual log lines Crossplane emits in this PR. I believe we should update our controllers to emit events as well as logs, and possibly tweak how we log a little to ensure logs are useful and reasonably consistent across controllers. I believe the Stacks side of the codebase is changing or will change a lot over the coming weeks and the workload controllers could also use an update to some of our newer controller patterns, so I would prefer to have the owners of each controller do this in future PRs. Per https://github.com/crossplaneio/crossplane/issues/944 I intend to document our logging and eventing best practices so that folks can tackle this at their leisure.

Fixes #331 

### Checklist

I have:
  appropriate.

[definition of done].


cmd/crossplane/main.go
```diff
@@ -17,176 +17,121 @@ limitations under the License.
...
-			outFile = os.Stdout
-		} else {
-			openFile, err := os.Create(*extUnpackOutfile)
-			kingpin.FatalIfError(err, "Cannot create outfile")
-			defer closeOrError(openFile)
-			outFile = openFile
+		log := logging.NewLogrLogger(zl.WithName("stacks"))
+
+		outFile := os.Stdout
+		if *extUnpackOutfile != "" {
```
hasheddan: It looks like this flag is not required, do we need to check it nil?

negz: I'm fairly certain kingpin ensures that string flags are always non-nil. If I add:

```go
		fmt.Printf("extUnpackOutfile value: '%s'\n", *extUnpackOutfile)
```

I don't get a nil pointer issue:

```
$ go run cmd/crossplane/main.go stack unstack --content-dir /tmp
extUnpackOutfile value: ''
{"level":"info","ts":1580721234.870985,"logger":"stacks","msg":"Unpacking stack","to":"/dev/stdout"}
main: error: failed to unpack stacks: Stack does not contain an app.yaml file
exit status 1
```

hasheddan: Gotcha! Thanks for clarifying 🙂

pkg/controller/stacks/install/installjob.go
```diff
@@ -263,8 +264,7 @@ func (jc *stackInstallJobCompleter) createJobOutputObject(ctx context.Context, o
 
 	meta.AddLabels(obj, labels)
 
-	// TODO(displague) pass/inject a controller specific logger
```
displague: Thank you!

negz:
@displague Do you feel okay with the parts of this PR that affect your end of the codebase?

pkg/controller/stacks/stacks.go
```diff
@@ -19,41 +19,23 @@ package stacks
...
-	}
-
-	for _, creator := range creators {
-		if err := (&install.Controller{
-			StackInstallCreator: creator,
-		}).SetupWithManager(mgr); err != nil {
+// Setup Crossplane Stacks controllers.
+func Setup(mgr ctrl.Manager, l logging.Logger) error {
+	for _, setup := range []func(ctrl.Manager, logging.Logger) error{
+		install.SetupStackInstall,
```
suskin: Neat :squid:

pkg/stacks/templates/engines/helm2.go
```diff
@@ -72,16 +71,16 @@ func (her *Helm2EngineRunner) CreateConfig(claim *unstructured.Unstructured, hc
...
-	her.Log.V(logging.Debug).Info("Converting configuration", "spec", s)
+	her.Log.Debug("Converting configuration", "spec", s)
 	configContents, err := yaml.Marshal(s)
 
-	her.Log.V(logging.Debug).Info("Configuration contents as yaml", "configContents", configContents)
+	her.Log.Debug("Configuration contents as yaml", "configContents", configContents)
 
 	if err != nil {
-		her.Log.Error(err, "Error marshaling claim spec as yaml!", "claim", claim)
+		her.Log.Info("Error marshaling claim spec as yaml!", "claim", claim, "error", err)
```
suskin: I agree with downgrading the level of this message. In fact, I believe we'd be okay with it at `Debug`, since the caller in the render phase controller also logs any returned error. I see that line seems to also log the error at `Info` - what are we thinking for the guideline for when to log at `Error`?

negz: Never - https://godoc.org/github.com/crossplaneio/crossplane-runtime/pkg/logging#Logger doesn't have an `Error` method. :) There are links to articles describing this philosophy in the package doc. Instead the convention is to include `"error", err` in the structured log arguments.

negz:
@turkenh Could you take a look at the latest commit in this PR? I rebased it on your host aware stack manager work and found that we were both passing config down from main to (some of) the controllers in different ways. I've tried to rewire things to make them consistent.

Is there a good way I can test this? I've validated that the stack controllers at least start without crashing in both hosted and regular modes.

cmd/crossplane/main.go
```diff
@@ -17,190 +17,133 @@ limitations under the License.
...
+			f, err := os.Create(*extUnpackOutfile)
+			kingpin.FatalIfError(err, "Cannot create output file")
+			defer kingpin.FatalIfError(f.Close(), "Cannot close file")
+			outFile = f
 		}
+		log.Info("Unpacking stack", "to", outFile.Name())
```
turkenh: Justed tested in host aware mode and stack installation failed due to this line with:

```
failed to parse output from job default.stack-gcp: Object 'Kind' is missing in '{"level":"info","ts":1580980892.3018987,"logger":"stacks","msg":"Unpacking stack","to":"/dev/stdout"}'
```

I believe this is independent of hosted mode though.

turkenh:
> @turkenh Could you take a look at the commit in this PR that relates to the host aware stack manager? I rebased the PR on your host aware stack manager work and found that we were both passing config down from main to (some of) the controllers in different ways. I've tried to rewire things to make them consistent.

Hi @negz,

I tested and commented about an issue that I found: https://github.com/crossplaneio/crossplane/pull/1227/files#r375740009

> Is there a good way I can test this? I've validated that the stack controllers at least start without crashing in both hosted and regular modes.

Ideally, we need to setup two cluster side by side and configure one as host that other as tenant by providing appropriate parameters. However, this might be a complicated setup unless we automate this which requires some work.
Hence, I just created a script which bootstraps a local kind cluster and uses it both as host and tenant for some quick local testing. 
Please see: #1237

cmd/crossplane/main.go
negz: Ah good catch, thanks! I wonder if this was already an issue if run in debug mode - we used to log this at that level per - https://github.com/crossplaneio/crossplane/pull/1227/files#diff-8452f5d38b0987fceae61edde51d44ccL459. I'll revert this to debug mode for now.

negz: I've just confirmed I can install `stack-gcp` successfully now that the log line is at debug.

PR #1231, opened 2020-02-04
template stacks: add user-friendly enable flag to helm chart

### LINKED ISSUE #1185: template stacks: add opt-in to helm chart

Now that we have a template stack controller in the codebase, we'll want to make it convenient for people to try it out once we've hardened the controller a little bit.

We already have a parameterized installation process for crossplane (via our helm chart). We should add a parameter to the helm chart to enable template stacks support.

suskin:

Fixes #1185 

Now that template stacks can be enabled by adventurous users by passing
a flag to the stack manager, we want to make it more convenient by
supporting it from within the helm chart.

Regression test:

* Built helm chart locally
* Installed helm chart locally with development script
* Checked to make sure deployments still started up

Template stack flag test:

* Built helm chart locally
* Installed helm chart locally with development script, adding a `--set templateStacks.enabled=true`
* Checked to make sure deployments still started up
* Looked for `Adding template controllers` in the logs of the stack manager pod

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/cluster/charts/crossplane-controllers/values.yaml.tmpl
+++ b/cluster/charts/crossplane-controllers/values.yaml.tmpl
@@ -22,3 +22,6 @@ hostedConfig:
   tenantKubernetesServicePort: ""
   controllerNamespace: ""
   crossplaneSATokenSecret: ""
+
+templateStacks:
+  enabled: false
--- a/cluster/charts/crossplane/values.yaml.tmpl
+++ b/cluster/charts/crossplane/values.yaml.tmpl
@@ -32,3 +32,6 @@ clusterStacks:
 
 hostedConfig:
   enabled: false
+
+templateStacks:
+  enabled: false
--- a/cluster/charts/crossplane-controllers/templates/stack-manager-deployment.yaml
+++ b/cluster/charts/crossplane-controllers/templates/stack-manager-deployment.yaml
@@ -33,6 +33,9 @@ spec:
         - --host-controller-namespace
         - {{ .Values.hostedConfig.controllerNamespace | quote }}
         {{- end }}
+        {{- if .Values.templateStacks.enabled }}
+        - --templates
+        {{- end }}
         {{- range $arg := .Values.args }}
         - {{ $arg }}
         {{- end }}
--- a/cluster/charts/crossplane/templates/stack-manager-deployment.yaml
+++ b/cluster/charts/crossplane/templates/stack-manager-deployment.yaml
@@ -33,6 +33,9 @@ spec:
         - --host-controller-namespace
         - {{ .Values.hostedConfig.controllerNamespace | quote }}
         {{- end }}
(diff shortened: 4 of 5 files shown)


cluster/charts/crossplane-controllers/values.yaml.tmpl
```diff
@@ -21,4 +21,7 @@ hostedConfig:
   tenantKubernetesServiceHost: ""
   tenantKubernetesServicePort: ""
   controllerNamespace: ""
-  crossplaneSATokenSecret: ""
\ No newline at end of file
+  crossplaneSATokenSecret: ""
+
+templateStacks:
```
jbw976: want to also add these new config values to the table in https://github.com/crossplaneio/crossplane/blob/master/docs/install-crossplane.md#configuration?

cluster/charts/crossplane/values.yaml.tmpl
```diff
@@ -31,4 +31,7 @@ clusterStacks:
     deploy: false
 
 hostedConfig:
-  enabled: false
\ No newline at end of file
+  enabled: false
+
+templateStacks:
```
muvaf: I am kind of confused. [Comment here](https://github.com/crossplaneio/crossplane/pull/1214#discussion_r373614885) made sense to me since templating is one of the things that this mechanism will do but we call it template stack in other places. Should we converge on a naming for everywhere?

cluster/charts/crossplane/values.yaml.tmpl
```diff
@@ -31,4 +31,7 @@ clusterStacks:
...
+  enabled: false
```
muvaf: As it doesn't affect any if the existing stuff, I suggest we enable it by default.

cluster/charts/crossplane-controllers/values.yaml.tmpl
suskin: Yeah definitely - thanks for the link! I was not aware of this document o:

cluster/charts/crossplane/values.yaml.tmpl
suskin: Yeah, this is a good point. It's a bit confusing. The reason this part is using the word "template" is because "template stacks" has been the working Product Name for the functionality we're adding. Since this part is external facing, I used template stacks here too. I agree we should converge on a name that we use consistently.

cluster/charts/crossplane/values.yaml.tmpl
suskin: I disagree with this. It does affect existing stuff since the stack definition controller (and stack configuration controller) run in the same process as the stack manager, and will kill the stack manager if they encounter a fatal error.

I agree with the general sentiment that we'll want to enable it by default soon though. I imagine us doing that once we've hardened it a bit more.

PR #1245, opened 2020-02-07
Support for PostgreSQL 11

### LINKED ISSUE #1244: Support PostgreSQL 11
Missing PostgreSQL 11 support

Allow PostgreSQL 11 support

vasartori:

Allow the version `11` on field `engineVersion` in resource `PostgreSQLInstance`
Fixes #1244

Create a PostgreSQLInstance spec with this:
```yaml
kind: PostgreSQLInstance
metadata:
  name: database-test
spec:
  classSelector:
    matchLabels:
      example: "true"
  writeConnectionSecretToRef:
    name: database-test
  engineVersion: "11"
```
### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/apis/database/v1alpha1/types.go
+++ b/apis/database/v1alpha1/types.go
@@ -66,7 +66,7 @@ type PostgreSQLInstanceSpec struct {
 	runtimev1alpha1.ResourceClaimSpec `json:",inline"`
 
 	// EngineVersion specifies the desired PostgreSQL engine version, e.g. 9.6.
-	// +kubebuilder:validation:Enum="9.6"
+	// +kubebuilder:validation:Enum="9.6";"11"
 	EngineVersion string `json:"engineVersion,omitempty"`
 }
 
--- a/apis/database/v1alpha1/types.go
+++ b/apis/database/v1alpha1/types.go
@@ -65,7 +65,7 @@ type MySQLInstanceList struct {
 type PostgreSQLInstanceSpec struct {
 	runtimev1alpha1.ResourceClaimSpec `json:",inline"`
 
-	// EngineVersion specifies the desired PostgreSQL engine version, e.g. 9.6.
+	// EngineVersion specifies the desired PostgreSQL engine version. Allowed Versions: 9.6 and 11.
 	// +kubebuilder:validation:Enum="9.6";"11"
 	EngineVersion string `json:"engineVersion,omitempty"`
 }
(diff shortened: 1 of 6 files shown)


docs/api/crossplaneio/crossplane/database-crossplane-io-v1alpha1.md
```diff
@@ -67,7 +67,7 @@ Appears in:
 
 Name | Type | Description
 -----|------|------------
-`engineVersion` | string | EngineVersion specifies the desired PostgreSQL engine version, e.g. 9.6.
+`engineVersion` | string | EngineVersion specifies the desired PostgreSQL engine version. 9.6 or 11 are allowed.
```
hasheddan: These docs are auto-generated using https://github.com/negz/crossdocs, so this should reflect the code comment on the `engineVersion` field. Unfortunately, it is a bit difficult for to setup `crossdocs` currently, so I see the following options here:
- You could remove this change to the docs and opt for the existing comment (it isn't wrong currently, but your update is an improvement IMO).
- You could setup `crossdocs` (I am more than happy to help here), and you could modify the code comment, then regenerate the docs here.
- You could remove this change, then I would be happy to follow-up with a PR to change the code comment and regenerate the docs. I could add you as an approver to make sure it matches your expectations.

This is a poor part of the development workflow right now and we need to move on getting a fix for it (related issue:  https://github.com/crossplaneio/crossplane/issues/1166). Apologies for the complications it is currently causing!

apis/database/v1alpha1/types.go
```diff
@@ -66,7 +66,7 @@ type PostgreSQLInstanceSpec struct {
 	runtimev1alpha1.ResourceClaimSpec `json:",inline"`
 
 	// EngineVersion specifies the desired PostgreSQL engine version, e.g. 9.6.
```
hasheddan: This is the code comment mentioned below from which the docs are generated.

vasartori: @hasheddan I think the best approach is setup crossdocs.
So, I tryed to run crossdocs without success... I folow the readme, but a lot of errors appears... 

I think my mistake with crossdocs was in this step: ```Update Gopkg.toml and doc.sh to ensure you're vendoring and running documentation generation for the versions of the Crossplane projects you care about.```

hasheddan: I'll follow-up with you offline to get you set up :+1:

vasartori:
Done!

negz:
Thanks @vasartori!

PR #1253, opened 2020-02-10
Adding default support to select control types in upbound-ui-spec

rathpc:


design/design-doc-upbound-ui-spec.md
```diff
@@ -143,6 +143,7 @@ initialConfig:
     Property | Description
     --- | ---
     `controlType` | **singleSelect**
+    `default` | An array containing a string to set as the selected value on render in the UI. This value should also appear in the `enum` property.
```
thephred: Is it really supposed to be an array?

rathpc: yes, it is the same for both single and multiple select

Issue #1254, opened 2020-02-11
Duplicate controller names break metrics


negz:
### What happened?

I noticed the following while running `stack-gcp` from master:

```
2020-02-10T16:04:28.317-0800    ERROR   controller-runtime.metrics      failed to register metric       {"name": "workqueue_depth", "queue": "claimdefaulting/kubernetescluster", 
"error": "duplicate metrics collector registration attempted"}                                                                                                                    
github.com/go-logr/zapr.(*zapLogger).Error                                                                                                                                        
        /Users/negz/control/go/pkg/mod/github.com/go-logr/zapr@v0.1.1/zapr.go:128                                                                                                 
sigs.k8s.io/controller-runtime/pkg/metrics.registerWorkqueueMetric                                                                                                                        /Users/negz/control/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.4.0/pkg/metrics/workqueue.go:37                                                                          
sigs.k8s.io/controller-runtime/pkg/metrics.workqueueMetricsProvider.NewDepthMetric                                                                                                
        /Users/negz/control/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.4.0/pkg/metrics/workqueue.go:50                                                                          
k8s.io/client-go/util/workqueue.(*queueMetricsFactory).newQueueMetrics                                                                                                                    /Users/negz/control/go/pkg/mod/k8s.io/client-go@v0.17.0/util/workqueue/metrics.go:243                                                                                     
k8s.io/client-go/util/workqueue.NewNamed                                                                                                                                          
        /Users/negz/control/go/pkg/mod/k8s.io/client-go@v0.17.0/util/workqueue/queue.go:44                                                                                        
k8s.io/client-go/util/workqueue.NewDelayingQueueWithCustomClock                                                                                                                           /Users/negz/control/go/pkg/mod/k8s.io/client-go@v0.17.0/util/workqueue/delaying_queue.go:50                                                                               
k8s.io/client-go/util/workqueue.NewNamedDelayingQueue                                                                                                                             
        /Users/negz/control/go/pkg/mod/k8s.io/client-go@v0.17.0/util/workqueue/delaying_queue.go:43                                                                               
k8s.io/client-go/util/workqueue.NewNamedRateLimitingQueue                                                                                                                                 /Users/negz/control/go/pkg/mod/k8s.io/client-go@v0.17.0/util/workqueue/rate_limiting_queue.go:46                                                                          
[...]

Issue #388, opened 2019-04-11
Consider shortening logger names


jbw976:
As discussed in the following PR comments, our logging names are somewhat verbose:
https://github.com/crossplaneio/crossplane/pull/385#discussion_r274086143
https://github.com/crossplaneio/crossplane/pull/385#discussion_r274089142

e.g. `controller.account.storage.azure.crossplane.io`

We can consider shortening these logger names somehow, without losing the knowledge of what part of the code base each logger belongs to.  The rook project uses abbreviated names instead of fully qualified type names, such as `edgefs-op-cluster`.

negz:
I like the idea of copying the controller-runtime pattern in which the loggers are named simply `crossplane.controller`, but have a value with the controller name. For example:

```
# A Crossplane log
2019-06-25T18:58:11.560Z        DEBUG   crossplane.controller   Reconciling     {"controller": "resourceclaim.crossplane.io", "request": "default/ekscluster"}

# A controller-runtime log
2019-06-25T18:58:11.065Z        DEBUG   kubebuilder.controller  Successfully Reconciled {"controller": "kubernetescluster.eks.compute.aws.crossplane.io", "request": "default/ekscluster"}
```

negz:
I'm experimenting with this as part of https://github.com/crossplaneio/crossplane-runtime/issues/104.

PR #1256, opened 2020-02-11
Use groupkind, not kind, for controller names

negz:

#1254

Per the above issue we attempted to use kind only. Unfortunately there are some cases in which we have two controllers of the same type (e.g. "managed") for the same kind in different groups. This means using a "type/kind" controller name like "claimbinding/rediscluster" can result in duplicate controller names, which breaks metric registration.

This commit switches Crossplane core to "type/groupkind", e.g. "claimbinding/rediscluster.cache.crossplane.io". This is longer than we'd like (per #388), but at least now the controller name is stored under a key ("controller") by convention, rather than literally being the name of the logger. I chose "type/groupkind" because in practice there can be only one controller per GroupKind.

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/pkg/controller/workload/kubernetes/scheduler/scheduler.go
+++ b/pkg/controller/workload/kubernetes/scheduler/scheduler.go
@@ -105,7 +105,7 @@ func UpdatePredicate(e event.UpdateEvent) bool {
 
 // Setup adds a controller that schedules KubernetesApplications.
 func Setup(mgr ctrl.Manager, l logging.Logger) error {
-	name := "scheduler/" + strings.ToLower(workloadv1alpha1.KubernetesApplicationKind)
+	name := "scheduler/" + strings.ToLower(workloadv1alpha1.KubernetesApplicationGroupKind)
 
 	return ctrl.NewControllerManagedBy(mgr).
 		Named(name).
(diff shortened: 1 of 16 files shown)


pkg/controller/workload/kubernetes/scheduler/scheduler.go
```diff
@@ -105,7 +105,7 @@ func UpdatePredicate(e event.UpdateEvent) bool {
 
 // Setup adds a controller that schedules KubernetesApplications.
 func Setup(mgr ctrl.Manager, l logging.Logger) error {
-	name := "scheduler/" + strings.ToLower(workloadv1alpha1.KubernetesApplicationKind)
+	name := strings.ToLower(workloadv1alpha1.KubernetesApplicationGroupKind)
```
hasheddan: Does this not conflict with the KubernetesApplication controller name above?

negz: Ah - the removal of the prefixes like `scheduler/` in this PR was unintentional. Will fix.

negz: Fixed - PTAL.

PR #1258, opened 2020-02-11
stacks: only labeled namespaces enroll in persona managment

### LINKED ISSUE #1124: Stack Manager should watch namespaces for persona enabling label

Part of #1036

https://github.com/crossplaneio/crossplane/blob/5b48149d758ceeb73c0b5f2e1d83d0af6e7a1776/design/one-pager-stacks-security-isolation.md#clusterroles-for-a-given-app-namespace

> Each namespace provides a unit of isolation with namespaced stacks, and thus a unique set of resources, based on the stacks that have been installed into that namespace. If someone were allowed to install stacks, it can't be assumed that they are cluster-admin, so the stack-manager will install and manage a set of roles that will aggregate the roles for the installed stacks into namespace specific roles for self-service. These roles should be bound to subjects to grant access to a particular namespace for end-user integration. For any namespace that is annotated with `rbac.crossplane.io/managed-roles: true` the stack-manager will ensure that ClusterRoles in the following format are created for the annotated namespace.
> ```
> crossplane:ns:{ns-name}:admin
> crossplane:ns:{ns-name}:edit
> crossplane:ns:{ns-name}:view
> ```

"StackManager should create workspace specific clusterroles for any properly labelled namespace."

Building on #1113, Stack Manager should begin to watch namespaces for the `managed-roles` label.

The stack-manager currently creates `crossplane:ns:{name}:{role}` clusterroles upon installation of a Stack.  This was a first step towards introducing managed roles for namespaces.

* Stop creating `crossplane:ns:{name}:{role}` roles on stack install
* Watch namespaces with the known label
  * When the label is added or removed, create or delete the `crossplane:ns:{name}:{role}` clusterroles
* Inventory `crossplane:ns:{name}:{role}` clusterroles and role-managed namespaces. Sync these periodically.

The clusterroles currently use the namespace records as owner references.  This ensures that the clusterroles will be deleted when the related namespaces are deleted.

displague:

Persona ClusterRoles are the ClusterRoles Crossplane creates when a Stack is installed in a namespace.  Given any namespace-scoped stack installed in namespace `foo`, Crossplane's Stack Manager will create the following clusterroles:

* `crossplane:ns:foo:view`
* `crossplane:ns:foo:edit`
* `crossplane:ns:foo:admin`

These persona roles allow for admins to administer `RoleBindings` granting their users the ability to work with these Stacks in various ways, including the ability to install more stacks within that namespace.  The problem with creating these roles at Stack install time is that the initial stack that led to the creation of these roles must be installed by the admin, not a user with `crossplane:ns:foo:admin`, since that clusterrole does not yet exist.

This PR moves the ClusterRole creation away from Stack creation time.  With this PR, an admin can now label a namespace `rbac.crossplane.io/managed-roles:"true"` to have these clusterroles created.  A user with a binding granting the crossplane-namespace-admin persona can then install the first stack in that namespace.

* Removes Persona ClusterRole (`crossplane:ns:foo:{view|edit|admin}`) creation from namespaced Stacks reconciliation
  * removes Stacks controller tests that looked for these clusterroles to be created
* Watches namespaces
  * determines if namespace is `rbac.crossplane.io/managed-roles:"true"` labeled
  * creates Persona ClusterRoles with the namespace as an OwnerReference (for GC cleanup)
  * deletes Persona ClusterRoles if the label is not present (or the resource has deletion timestamps)
  * adds tests to verify these changes

I determined that this PR does not need to be host-aware. 
 [RBAC rules are created with the tennant](https://github.com/crossplaneio/crossplane/pull/1138#discussion_r364281011) (not host) client that `h.kube.client` refers to. The code for the new persona controller was taken from the Stack controller where the same client was used.

Fixes #1124

1. `kubectl scale -n crossplane-system deployment --replicas=0 crossplane-stack-manager`
1. Run this branch `go run cmd/main.go stack manage --debug`
1. `kubectl create namespace foo`
  a. Observe there are no clusterroles for foo `kubectl get clusterrole | grep foo`
1. `kubectl label namespace foo rbac.crossplane.io/managed-roles=true`
  a. Observe there are now clusterroles for foo `kubectl get clusterrole | grep foo`
1. Delete label `kubectl label namespace foo rbac.crossplane.io/managed-roles-`
  a. Observe there are no clusterroles for foo `kubectl get clusterrole | grep foo`
1. Repeat the create+delete process by setting a non-true value for the label
1. Repeat the create+delete process by deleting the namespace
1. Break the `go run` (ctrl+c)
1. `kubectl scale -n crossplane-system deployment --replicas=1 crossplane-stack-manager`

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/design/one-pager-stacks-security-isolation.md
+++ b/design/one-pager-stacks-security-isolation.md
@@ -375,7 +375,7 @@ Each namespace provides a unit of isolation with namespaced stacks, and thus a u
  for the installed stacks into namespace specific roles for self-service. These `ClusterRoles` should be bound to subjects
  using a `RoleBinding` (not `ClusterRoleBinding`) to grant access to a particular namespace for end-user integration.
 
-This design calls for the stack-manager to act on any namespace that is labeled with `rbac.crossplane.io/managed-roles: true`,
+This design calls for the stack-manager to act on any namespace that is labeled with `rbac.crossplane.io/managed-roles: "true"`,
 ensuring that ClusterRoles in the following format are created for the labeled namespace.
 
  * `crossplane:ns:{ns-name}:admin`
@@ -384,11 +384,11 @@ ensuring that ClusterRoles in the following format are created for the labeled n
 
 For discoverability of the roles for a given namespace we will add the following labels:
  * `crossplane.io/scope: "namespace"`
- * `namespace.crossplane.io/{namespace-name}: true`
+ * `namespace.crossplane.io/{namespace-name}: "true"`
 
 Match Roles:
  * `rbac.crossplane.io/aggregate-to-namespace-{role}: "true"`
- * `namespace.crossplane.io/{namespace-name}: true // to match specific namespace.`
+ * `namespace.crossplane.io/{namespace-name}: "true" // to match specific namespace.`
 
 |ClusterRole Name | Kubernetes Counterpart | Permissions |
 |:---------------------|:----------|:----------|
@@ -421,7 +421,7 @@ Labels for these roles will be for aggregation purposes:
  * `rbac.crossplane.io/aggregate-to-{install-scope}-{role}: "true"`
  * `namespace.crossplane.io/{namespace-name}: "true"` // Only included on namespaced stacks install
 
-The format for the namespace label `namespace.crossplane.io/{namespace-name}: true`, which may seem unintuitive
+The format for the namespace label `namespace.crossplane.io/{namespace-name}: "true"`, which may seem unintuitive
  is required to support ClusterRole aggregation. Since we will have some ClusterRoles which aggregate to more than
  multiple namespaces and label keys must be unique, we need to encode the key/value in the key.
 
@@ -563,7 +563,7 @@ metadata:
   creationTimestamp: 2019-10-16T18:25:10Z
   name: app-team-1
   labels:
-    rbac.crossplane.io/managed-roles: true # Stack manager creates and will manage namespace specific ClusterRoles when it see's this.
+    rbac.crossplane.io/managed-roles: "true" # Stack manager creates and will manage namespace specific ClusterRoles when it see's this.
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 # This role binding allows "jane" to admin the Crossplane app-team-1 namespace.
--- a/design/one-pager-stacks-security-isolation.md
+++ b/design/one-pager-stacks-security-isolation.md
@@ -375,8 +375,8 @@ Each namespace provides a unit of isolation with namespaced stacks, and thus a u
  for the installed stacks into namespace specific roles for self-service. These `ClusterRoles` should be bound to subjects
  using a `RoleBinding` (not `ClusterRoleBinding`) to grant access to a particular namespace for end-user integration.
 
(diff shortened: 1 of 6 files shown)


pkg/controller/stacks/persona/persona.go
```diff
@@ -0,0 +1,260 @@
...
+}
+
+// delete ClusterRoles for namespace personas
+func (h *nsPersonaHandler) delete(ctx context.Context) error {
+	// Logging that clusterroles are attempting to be deleted would
+	// either be noisy (logged on unmanaged namespaces) or cost a
+	// lookup for existing clusterroles
+
+	labels := stacks.ParentLabels(h.ns)
+	if err := h.kube.DeleteAllOf(ctx, &rbacv1.ClusterRole{}, client.MatchingLabels(labels)); err != nil {
```
turkenh: I don't think DeleteAllOf returns `NotFound` if already deleted, but worths checking. We would need to ignore `NotFound` in that case.

pkg/controller/stacks/persona/persona.go
```diff
@@ -0,0 +1,260 @@
...
+	kube client.Client
+	log  logging.Logger
+	factory
+}
+
+// Setup adds a controller that reconciles Namespaces.
+func Setup(mgr ctrl.Manager, l logging.Logger) error {
+	r := &Reconciler{
+		kube:    mgr.GetClient(),
+		factory: &nsPersonaHandlerFactory{},
```
jbw976: is a factory strictly necessary in this reconciler? i've used that pattern before when the "handler" can only be generated at runtime with specific config or environment information, but the handler in this reconciler looks like it just takes in some objects and does no other init work that you'd want to avoid during tests (e.g. making a cloud provider call, or looking up something in the k8s api).

you could consider removing this factory and just use a handler directly (instead of delegating the creation of the handler to the factory) to reduce complexity. I believe that would work. Your choice though, it's not a big deal.

design/one-pager-stacks-security-isolation.md
```diff
@@ -562,7 +562,7 @@ kind: Namespace
 metadata:
   creationTimestamp: 2019-10-16T18:25:10Z
   name: app-team-1
-  annotations:
+  labels:
     rbac.crossplane.io/managed-roles: true # Stack manager creates and will manage namespace specific ClusterRoles when it see's this.
```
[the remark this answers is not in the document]
displague: updated in 15c03167

pkg/controller/stacks/persona/persona.go
displague: I took a quick pass at removing the factory but I ended up with a very similar complexity as I still needed a way to inject a handler into the tests.  If you are ok with leaving the handler in I think we can revisit this in future refactors.

pkg/controller/stacks/persona/persona.go
displague: I've verified that if I delete the clusterroles before removing the label there is no error condition here.

I believe I was wary of this same possible error value in a past PR and may have included a `NotFound` check.  I'll remove that when I next encounter that code.

pkg/controller/stacks/persona/persona.go
jbw976: yes sure, if it isn't an obvious win then no problem leaving it in!

PR #1264, opened 2020-02-14
StackInstall source should be used in unqualified Stack deployments

displague:

When a StackInstall (or ClusterStackInstall) includes a `spec.source`, this value will be prepended to the Stack deployment so long as no other source string is detected.
    
StackDefinition resources also get this change, affecting where they locate behavior images and the generic templating-controller.
    
Fixes crossplaneio/crossplane#1246

_Approach is based on https://github.com/crossplaneio/crossplane/issues/1246#issuecomment-583498263_

* Install a stack using no source:
  `kubectl crossplane stack install --cluster --namespace crossplane-system crossplane/stack-aws:v0.5.0 stack-aws`
  * Observe the deployment in the Stack resource, and observe the deployment's final controller image
* Install a stack using a source:
  `kubectl crossplane stack install --cluster --namespace crossplane-system crossplane/stack-aws:v0.5.0 stack-aws registry.hub.docker.com`
  * Observe the deployment in the Stack resource, and observe the deployment's final controller image
  
### Checklist

I have:
  appropriate.

[definition of done].


apis/stacks/v1alpha1/types.go
```diff
@@ -85,22 +93,116 @@ type StackInstallStatus struct {
...
+	Scheme string
+	Host   string
+	Port   int
+	Prefix string
+	Path   string
+	Tag    string
+}
+
+// String builds a string URI from a StackImage source and path
+func (s *StackImage) String() string {
```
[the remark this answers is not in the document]
displague: I agree with this sentiment. I was originally looking around in `docker/docker` and found relevant pieces that were not exported, related to this function https://godoc.org/github.com/docker/docker/client#ParseHostURL

`docker/distribution` does seem to offer the functions I was hoping to find, https://godoc.org/github.com/docker/distribution/reference#SplitHostname for one.

I'll give this a try.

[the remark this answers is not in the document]
displague: The `reference` package seems relatively small, I don't see a concern with importing it if a matching function exists there.

apis/stacks/v1alpha1/types_test.go
```diff
@@ -29,35 +32,69 @@ var (
...
+	type want struct {
+		url string
+		err error
+	}
+
 	tests := []struct {
 		name string
 		spec StackInstallSpec
-		want string
+		want want
```
negz: Not a blocker, but I've been experimenting with including a human readable `reason` string that is printed when a test case fails. I find this allows succinct test/subtest names, because it removes the temptation to try to encode the full context of the test in the test name.

apis/stacks/v1alpha1/types.go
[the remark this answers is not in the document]
displague: @jbw976 https://github.com/crossplaneio/crossplane/pull/1264/commits/b9d3705ddd56b954a5b411f93043e9f56c0fac61 defers this logic to docker/distribution reference

apis/stacks/v1alpha1/types.go
```diff
@@ -85,22 +86,54 @@ type StackInstallStatus struct {
...
 
-	return fmt.Sprintf("%s/%s", spec.Source, spec.Package)
+	// image contained a source
+	return image, nil
 }
 
-// Image gets the Spec.Image of the StackInstall
-func (si *StackInstall) Image() string { return si.Spec.Image() }
+// GetSpec gets the Spec of the StackInstall
+func (si *StackInstall) GetSpec() *StackInstallSpec {
```
jbw976: does this need to return a pointer type? when it's a pointer it makes callers want/need to implement nil checks so make sure they don't have a nil pointer, which is extra work that could hopefully be avoided

apis/stacks/v1alpha1/types_test.go
```diff
@@ -28,35 +28,79 @@ var (
...
-		want string
+		name   string
+		reason string
+		spec   StackInstallSpec
+		want   want
 	}{
 		{
-			name: "NoPackageSource",
+			name:   "NoPackageSource",
+			reason: "Packages without StackInstall source should be unchanged",
```
jbw976: i love this "reason" field for each test case that @negz suggested.  it makes the purpose/intent of the test case so much more clear without having to encode it into the test name.

pkg/controller/stacks/install/installjob.go
```diff
@@ -72,6 +72,12 @@ func createInstallJob(i v1alpha1.StackInstaller, executorInfo *stacks.ExecutorIn
 		namespace = o.Namespace
 	}
 
+	img, err := i.GetSpec().ImageWithSource(i.GetSpec().Package)
```
jbw976: here's a place where it's not clear if the caller of `i.GetSpec()` should be checking for nil before proceeding.  Knowing how the code is implemented it shouldn't ever be nil but the contract isn't clear just by looking at the interface. Does this `i.GetSpec()` need to return a pointer?

also, you can probably just call `i.GetSpec()` once instead of twice on this line and another time on 78 below

pkg/controller/stacks/install/installjob.go
```diff
@@ -72,6 +72,12 @@ func createInstallJob(i v1alpha1.StackInstaller, executorInfo *stacks.ExecutorIn
...
+	if err != nil {
+		// Applying the source is best-effort
```
jbw976: debug log statement about this scenario occuring?

pkg/controller/stacks/install/installjob.go
```diff
@@ -344,6 +420,28 @@ func convertToUnstructured(o *v1alpha1.StackDefinition) (*unstructured.Unstructu
 	return &unstructured.Unstructured{Object: u}, nil
 }
 
+// convertStackToUnstructured takes a Kubernetes object and converts
+// it into *unstructured.Unstructured that can be used as KubernetesApplication
```
jbw976: copy/paste typo here for KubernetesApplication?

apis/stacks/v1alpha1/types_test.go
```diff
@@ -28,35 +28,79 @@ var (
...
+			reason: "Sources should honor host, port, and prefix",
+			spec: StackInstallSpec{
+				Source:  "insecure:3000/prefix/",
+				Package: "cool/tagless-package",
+			},
+			want: want{
+				url: "insecure:3000/prefix/cool/tagless-package",
+			},
+		},
+		{
```
jbw976: how about another test case where the package is already a fully qualified image with a source?  e.g.:
```go
name:   "PackageContainsSourceAlready",
reason: "Packages that contain a source already should have that source honored",
spec: StackInstallSpec{
	Source:  "foo.bar",
	Package: "my.registry/cool/tagless-package",
},
want: want{
	url: "my.registry/cool/tagless-package",
},
```

apis/stacks/v1alpha1/types.go
displague: It can be avoid here. The only case that the pointer is useful now is for tests that provide StackInstall modifiers which I would like to add in #1281 (or could bring here).

The alternative is that I add `SetSource`, `SetImagePullPolicy`, `SetImagePullSecrets` which I am fine with adding.  `Get*` might remove my need for `GetSpec`.

displague: I've pushed this approach in the latest.

displague: Same change (SetSource, GetPackage) is now here in ef9be762. GetSpec is no longer introduced.

PR #1272, opened 2020-02-18
v0.8.0: prep docs and references

hasheddan:

Updates references and docs to point to new versions of Crossplane and stacks.

n/a

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/cluster/charts/crossplane/README.md
+++ b/cluster/charts/crossplane/README.md
@@ -126,7 +126,7 @@ metadata:
   name: stack-aws
   namespace: aws
 spec:
-  package: "crossplane/stack-aws:master"
+  package: "crossplane/stack-aws:v0.6.0"
 ```
 
 Then you can install the AWS stack into Crossplane in the `aws` namespace with the following command:
@@ -150,7 +150,7 @@ metadata:
   name: stack-rook
   namespace: rook
 spec:
-  package: "crossplane/stack-rook:master"
+  package: "crossplane/stack-rook:v0.4.0"
 ```
 
 Then you can install the Rook stack into Crossplane in the `rook` namespace with the following command:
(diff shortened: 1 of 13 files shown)


cluster/charts/crossplane/README.md
```diff
@@ -150,7 +150,7 @@ metadata:
   name: stack-rook
   namespace: rook
 spec:
-  package: "crossplane/stack-rook:master"
+  package: "crossplane/stack-rook:v0.4.0"
```
jbw976: i would expect all the other provider stacks to need to be updated in this file as well.  it looks like even in master, something funny is going on with the build automation to [copy the docs/install-crossplane.md over to this file](https://github.com/crossplaneio/crossplane/blob/master/Makefile#L146) in the helm charts.  we can investigate that in master though.

jbw976: oh i think that `sed` command intends to wipe out the markdown file's front matter, but unintentionally wipes out some example yamls too :(

i'll open an issue for that

hasheddan: Gotcha, thanks! We can backport updates :+1: Since this is a straight copy do you want me to handle manually here or no?

jbw976: yes, may as well just copy the file over manually in this PR and update it's refs manually also.

hasheddan: @jbw976 I just realized that will cause it to fail `make check-diff` though :disappointed:

hasheddan: I have a fix ready for the `sed` command though.

jbw976: ah darn and jenkins calls `make check-diff`.  good point.  well let's just merge this PR as is then without any manual modifications.

jbw976: and we can consider backporting the fix for the `sed` call later on

hasheddan: SGTM :heavy_check_mark:

jbw976:
haha apparently i can approve a PR more than once 🤔

PR #1276, opened 2020-02-18
Define core OAM API types

negz:

This PR adds the core OAM types to Crossplane. Type schema is inferred from https://github.com/oam-dev/spec/pull/304, pending a formal update of the OAM specification. Relates to https://github.com/crossplaneio/crossplane/issues/1268.

I'll also need to run https://github.com/negz/crossdocs to generate the Crossplane API documentation for the new `oam` API package. It's easier to do that in a separate PR, once this one is merged

### Checklist

I have:
  appropriate.

[definition of done].


negz:
CC [OAM maintainers](https://github.com/oam-dev/spec/blob/be06c2f3/OWNERS.md): @technosophos @vturecek @resouer @hongchaodeng

apis/oam/v1alpha1/register.go
```diff
@@ -0,0 +1,92 @@
...
+	"reflect"
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/scheme"
+)
+
+// Package type metadata.
+const (
+	Group   = "core.oam.dev"
+	Version = "v1alpha1"
```
wonderflow: maybe `v1alpha2` is better as we have make a big difference with current `v1alpha1` version

negz: Good call! I'll update to v1alpha2.

apis/oam/v1alpha1/types.go
```diff
@@ -0,0 +1,663 @@
...
+type TraitDefinitionList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []TraitDefinition `json:"items"`
+}
+
+// A ScopeDefinitionSpec defines the desired state of a ScopeDefinition.
+type ScopeDefinitionSpec struct {
+	// Reference to the CustomResourceDefinition that defines this scope kind.
+	Reference DefinitionReference `json:"definitionRef"`
```
hongchaodeng: Shouldn't it be type of [ObjectReference(https://github.com/kubernetes/api/blob/9e5ffd1f1320950b238cfce291b926411f0af722/core/v1/generated.proto#L2144)?

negz: No. Quoting [API Conventions](https://github.com/kubernetes/community/blob/017dbe20/contributors/devel/sig-architecture/api-conventions.md#references-to-related-objects):

> References to specific objects, especially specific resource versions and/or specific fields of those objects, are specified using the ObjectReference type (or other types representing strict subsets of it).

The reasons we prefer to define our own types (that are strict subsets of ObjectReference) are explained in https://github.com/crossplaneio/crossplane-runtime/issues/49.

apis/oam/v1alpha2/types.go
```diff
@@ -0,0 +1,857 @@
...
+
+	// FieldPaths specifies an array of fields within this Component's workload
+	// that will be overwritten by the value of this parameter. The type of the
+	// parameter (e.g. int, string) is inferred from the type of these fields;
+	// All fields must be of the same type. Fields are specified as JSON field
+	// paths without a leading dot, for example 'spec.replicas'.
+	FieldPaths []string `json:"fieldPaths"`
+
+	// TODO(negz): Use +kubebuilder:default marker to default Required to false
+	// once we're generating v1 CRDs.
```
hasheddan: This could also be changed from a `*bool` to a `bool` if we wanted default to `false`. However, that would be a divergence from the Kubernetes [api conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required) so I would support manually handling this until default tag is available as you suggest :+1:

apis/oam/oam.go
```diff
@@ -0,0 +1,38 @@
+/*
+Copyright 2019 The Crossplane Authors.
```
hasheddan: Not sure if this is why, but it might fix SonarQube to update the year here?

hasheddan: RIP sonar

apis/oam/v1alpha2/register.go
```diff
@@ -0,0 +1,140 @@
...
+var (
+	ServerKind             = reflect.TypeOf(Server{}).Name()
+	ServerGroupKind        = schema.GroupKind{Group: Group, Kind: ServerKind}.String()
+	ServerKindAPIVersion   = ServerKind + "." + SchemeGroupVersion.String()
+	ServerGroupVersionKind = SchemeGroupVersion.WithKind(ServerKind)
+)
+
+// SingletonServer type metadata.
+var (
+	SingletonServerKind             = reflect.TypeOf(SingletonServer{}).Name()
```
hongchaodeng: We are in the process of shrinking core types. We only need _Server_ as a container workload and _singleton_, _daemonset_, etc. are just developer traits (we called policies) of the workload.

Proposal: https://docs.google.com/document/d/1d3kVBFyf7yf-YzmVqkL1hr_qcIVXf6BUJLWKu9NJwvE/edit?usp=sharing

hongchaodeng: My suggestion is to just define _Server_ for the comment.

apis/oam/v1alpha2/types.go
```diff
@@ -0,0 +1,857 @@
...
+	// +optional
+	ParameterValues []ComponentParameterValue `json:"parameterValues,omitempty"`
+
+	// Traits of the specified component.
+	// +optional
+	Traits []ComponentTrait `json:"traits,omitempty"`
+
+	// Scopes in which the specified component should exist.
+	// +optional
+	Scopes []ComponentScope `json:"scopes,omitempty"`
```
ryanzhang-oss: Is it necessary to add the ComponentScope structure? Can we just use [] ScopeReference?

apis/oam/v1alpha2/types.go
```diff
@@ -0,0 +1,857 @@
...
+// A TraitDefinition registers a kind of Kubernetes custom resource as a valid
+// OAM trait kind by referencing its CustomResourceDefinition. The CRD is used
+// to validate the schema of the trait when it is embedded in an OAM
+// ApplicationConfiguration.
+// +kubebuilder:printcolumn:JSONPath=".spec.definitionRef.name",name=DEFINITION-NAME,type=string
+type TraitDefinition struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec TraitDefinitionSpec `json:"spec,omitempty"`
```
ryanzhang-oss: I wonder why there is no status here?

apis/oam/v1alpha2/types.go
```diff
@@ -0,0 +1,857 @@
...
+// A ScopeDefinition registers a kind of Kubernetes custom resource as a valid
+// OAM scope kind by referencing its CustomResourceDefinition. The CRD is used
+// to validate the schema of the scope when it is embedded in an OAM
+// ApplicationConfiguration.
+// +kubebuilder:printcolumn:JSONPath=".spec.definitionRef.name",name=DEFINITION-NAME,type=string
+type ScopeDefinition struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec ScopeDefinitionSpec `json:"spec,omitempty"`
```
ryanzhang-oss: I wonder why there is no status here?

apis/oam/v1alpha2/register.go
```diff
@@ -0,0 +1,140 @@
...
+	SingletonTaskGroupVersionKind = SchemeGroupVersion.WithKind(SingletonTaskKind)
+)
+
+func init() {
+	SchemeBuilder.Register(&WorkloadDefinition{}, &WorkloadDefinitionList{})
+	SchemeBuilder.Register(&TraitDefinition{}, &TraitDefinitionList{})
+	SchemeBuilder.Register(&ScopeDefinition{}, &ScopeDefinitionList{})
+	SchemeBuilder.Register(&Component{}, &ComponentList{})
+	SchemeBuilder.Register(&ApplicationConfiguration{}, &ApplicationConfigurationList{})
+	SchemeBuilder.Register(&Server{}, &ServerList{})
```
ryanzhang-oss: Also, is it possible to separate the core workload/trait/scope into another file/place? Conceptually, they are not the same type as the resources defined above.

apis/oam/v1alpha2/types.go
```diff
@@ -0,0 +1,857 @@
+/*
+Copyright 2019 The Crossplane Authors.
```
ryanzhang-oss: We might need to revisit this part after the governance/trademark agreement is signed.

ryanzhang-oss: also, it's 2020 now

apis/oam/v1alpha2/types.go
```diff
@@ -0,0 +1,857 @@
...
+// +kubebuilder:object:root=true
+
+// A Server is a long-running, scalable workload that has a network endpoint
+// with a stable name.
+// +kubebuilder:subresource:status
+type Server struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   ContainerisedWorkloadSpec   `json:"spec,omitempty"`
```
ryanzhang-oss: Since a component only inline a CR for this CRD,  the OAM controller won't even know if the component is referring to a "server" workload or a "singletonServer" anymore.  I wonder how can this work?

ryanzhang-oss: There might be another issue. The ContainerisedWorkloadSpec doesn't have GVK in it. I guess "server" is the schema instead of a workload then? How does this work? Do we actually have a "server" workload defined (which has a definitionRef in it) somewhere?

apis/oam/v1alpha2/types.go
negz: According to [API conventions](https://github.com/kubernetes/community/blob/bdec08/contributors/devel/sig-architecture/api-conventions.md#spec-and-status):

> All objects that represent a physical resource whose state may vary from the user's desired intent SHOULD have a "spec" and a "status". Objects whose state cannot vary from the user's desired intent MAY have only "spec", and MAY rename "spec" to a more appropriate name.

I don't think workload, trait, or scope will ever vary from the user's desired intent so it seemed appropriate to avoid including a status. I don't imagine there will be any controller actively reconciling (for example) `kind: TraitDefinition`, but rather the `ApplicationConfiguration` reconciler will use these kinds to validate workloads, traits, and scopes.

apis/oam/v1alpha2/types.go
negz: See https://github.com/crossplane/crossplane/pull/1276#discussion_r382811756

apis/oam/v1alpha2/register.go
negz: They should live in the same package because they're both the same API group and version. I've moved them into a separate file. I'd be happy to move them into a separate package if we'd be happy to change their API group - e.g. `workload.core.oam.dev` or similar.

apis/oam/v1alpha2/types.go
negz: I'll defer to @bassam here. :)

apis/oam/v1alpha2/types.go
negz: It's not strictly necessary, but I've nested it for future extensibility. The nesting allows us to add other keys at the same level as the `scopeRef` - this is the same principle as @hongchaodeng suggesting each trait CR be enveloped in a `trait` key, rather than simply an array of trait CRs.

apis/oam/v1alpha2/register.go
```diff
@@ -0,0 +1,140 @@
...
+// ApplicationConfiguration type metadata.
+var (
+	ApplicationConfigurationKind             = reflect.TypeOf(ApplicationConfiguration{}).Name()
+	ApplicationConfigurationGroupKind        = schema.GroupKind{Group: Group, Kind: ApplicationConfigurationKind}.String()
+	ApplicationConfigurationKindAPIVersion   = ApplicationConfigurationKind + "." + SchemeGroupVersion.String()
+	ApplicationConfigurationGroupVersionKind = SchemeGroupVersion.WithKind(ApplicationConfigurationKind)
+)
+
+// Server type metadata.
+var (
```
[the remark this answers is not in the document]
negz: Done.

apis/oam/v1alpha2/register.go
negz: I believe we agreed on `ContainerisedWorkload` for the name; this is done.

apis/oam/v1alpha2/types.go
```diff
@@ -0,0 +1,857 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
+
+	runtimev1alpha1 "github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
+)
+
+// A DefinitionReference refers to a CustomResourceDefinition by name.
+type DefinitionReference struct {
+	// Name of the referenced CustomResourceDefinition.
+	Name string `json:"name"`
```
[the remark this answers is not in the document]
negz: No, my intention was that this refers to a CRD by name. For example:

```yaml
apiVersion: apiextensions.k8s.io/v1beta1
kind: CustomResourceDefinition
metadata:
  name: containerisedworkloads.core.oam.dev
spec:
  group: core.oam.dev
  names:
    kind: ContainerisedWorkload
    listKind: ContainerisedWorkloadList
    plural: containerisedworkloads
    singular: containerisedworkload
  scope: Namespaced
  subresources:
    # etc...
```

We'd refer to the above CRD by name, so:

```yaml
kind: WorkloadDefinition
metadata:
  name: containerisedworkloads.core.oam.dev
spec:
  definitionRef:
    # Name of the above CRD. We can omit apiVersion and kind here
    # because definition is always kind: CustomResourceDefinition in Kubernetes.
    name: containerisedworkloads.core.oam.dev
```

I believe this makes sense in Kubernetes because when multiple versions of a resource kind exist (e.g. `v1alpha2` and `v1beta1`) they must both be defined by one CRD, and they're both served at the same time. That is, when you create a `v1beta1` `ContainerisedWorkload` an equivalent `v1alpha2` `ContainerisedWorkload` is also served.

If we _really_ wanted to prevent `Component` authors from embedding any version other than `v1beta1`, despite the fact that any other supported versions would implicitly be served by the API server we could instead write something like:

```yaml
kind: WorkloadDefinition
metadata:
  name: containerisedworkloads.core.oam.dev
spec:
  definitionRef:
    apiVersion: core.oam.dev/v1beta1
    kind: ContainerisedWorkload
```

I prefer to explicitly reference the CRD by name, because it makes it easier for our controllers to find and get the CRD for a particular kind of workload in order to validate components that embed said workload kind.

apis/oam/v1alpha2/types.go
negz: This changed in my latest push of this PR because we agreed to use a single workload type named `ContainerisedWorkload`. At the time you made this comment there were multiple workload kinds - e.g. `Server`, `SingletonServer`, etc that all happened to use the same struct for their spec and status. Does that make sense?

apis/oam/v1alpha2/types.go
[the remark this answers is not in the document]
ryanzhang-oss: Is this valid?

```yaml
kind: WorkloadDefinition
metadata:
  name: containerisedworkloads.core.oam.dev
spec:
  definitionRef:
    name: core.oam.dev/v1beta1.ContainerisedWorkload
```

```yaml
kind: WorkloadDefinition
metadata:
  name: containerisedworkloads.core.oam.dev
spec:
  definitionRef:
    name: ContainerisedWorkload.core.oam.dev/v1beta1
```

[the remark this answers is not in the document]
negz: No; those appear to be encodings of type metadata rather than names. If we wanted to use type metadata I would suggest the `apiVersion` and `kind` style reference from my previous comment. Per that comment, I think referring to a CRD by name is preferable.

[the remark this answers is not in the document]
wonderflow: I think we do really need a version, we could define like this:
```
kind: WorkloadDefinition
metadata:
  name: containerisedworkloads.core.oam.dev
spec:
  definitionRef:
    name: containerisedworkloads.core.oam.dev
    version: v1beta1
```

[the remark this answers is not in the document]
ryanzhang-oss: Thanks, Nic. I think we need to include the version as 

- I am not sure how to pinpoint a version without it
- Not all definitions Ref are CRD which includes version
- We don't want to enforce GKV style

How about 

```yaml
kind: WorkloadDefinition
metadata:
  name: containerisedworkloads.core.oam.dev
spec:
  definitionRef:
    name: ContainerisedWorkload.core.oam.dev
   version: v1beta1
```

apis/oam/v1alpha2/register.go
```diff
@@ -0,0 +1,95 @@
...
+import (
+	"reflect"
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/scheme"
+)
+
+// Package type metadata.
+const (
+	Group   = "core.oam.dev"
```
ryanzhang-oss: After we separate out the definition of core types out of the spec, it seems that it will make sense to separate the OAM spec level objects from the core level objects. It was kinda confusing.  Does it make sense?

apis/oam/v1alpha2/types.go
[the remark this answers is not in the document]
resouer: See my comment: https://github.com/oam-dev/spec/pull/316/files#r383564313

I really like to use CRD name here, I just propose we have an optional filed `version` in case multiple CRDs are there.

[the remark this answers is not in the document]
negz: Perhaps you could help me understand why we'd like to pinpoint one version of a kind defined by a CRD. It's not clear to me why we'd want to do this.

Let's say we have a `kind: CustomResourceDefinition` that defines CR `kind: Example`. The CRD defines the schema of two versions; `v1alpha1` and `v1beta1` - note that all versions of a particular GroupKind must be defined via the same CRD. If I author a new `v1beta1` instance of `kind: Example`  a `v1alpha1` compatible representation of that instance is automatically served. That is to say, I can access my new `kind: Example` CR via both the `v1alpha1` and the `v1beta1` API endpoints and schemas. This is discussed at https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning/#writing-reading-and-updating-versioned-customresourcedefinition-objects.

So given this, why would we want to bless only one of these versions as a valid workload? Why not allow any valid version? They'll all be translated to one internal version schema and handled by the same controller once they're instantiated anyhow.

apis/oam/v1alpha2/register.go
negz: Sure - I'm not opinionated about what API group these live in.

apis/oam/v1alpha2/types.go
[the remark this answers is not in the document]
ryanzhang-oss: @negz , we (Alibaba) discussed internally and agreed that we don't need to add version for now.
As for the name of this field, do you think `schemaName` is more informative?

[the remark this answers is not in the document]
negz: > As for the name of this field

Which field are you proposing we rename? `spec.definitionRef` or `spec.definitionRef.name`?

```yaml
spec:
  definitionRef:
    # This violates Kubernetes API conventions for referencing a resource.
    # References must be a strict subset of the ObjectReference type.
    schemaName: foo
```

```yaml
spec:
  # Presuming we remove spec.definitionRef and instead use schemaName
  # like this we'll be compliant with Kubernetes API conventions. I'd prefer
  # definitionName though.
  schemaName: somecrd.example.org
```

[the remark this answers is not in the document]
ryanzhang-oss: the first one, nm then.

apis/oam/v1alpha2/core_workload_types.go
```diff
@@ -0,0 +1,398 @@
...
+
+	// Resources managed by this containerised workload.
+	Resources []ResourceReference `json:"resources,omitempty"`
+}
+
+// +kubebuilder:object:root=true
+
+// A ContainerisedWorkload is a workload that runs OCI containers.
+// +kubebuilder:subresource:status
+type ContainerisedWorkload struct {
```
hasheddan: Nitpick: can we call this `ContainerizedWorkload` to match updates to spec in https://github.com/oam-dev/spec/pull/316? I think this may actually be necessary as it is a core type and must be implemented as compliant with spec. Alternatively we could make the spec call it `ContainerisedWorkload`. The internet did not seem favor one or the other 🤷‍♂

negz: Yeah, will do.

apis/oam/v1alpha2/core_types.go
```diff
@@ -0,0 +1,375 @@
...
+type TraitDefinitionSpec struct {
+	// Reference to the CustomResourceDefinition that defines this trait kind.
+	Reference DefinitionReference `json:"definitionRef"`
+
+	// AppliesToWorkloads specifies the list of workload kinds this trait
+	// applies to. Workload kinds are specified in kind.group/version format,
+	// e.g. server.core.oam.dev/v1alpha2. Traits that omit this field apply to
+	// all workload kinds.
+	// +optional
+	AppliesToWorkloads []string `json:"appliesToWorkloads,omitempty"`
```
ryanzhang-oss: Since we plan to allow trait to apply to "policy" later, should this still be called `appliesTo` as in alpha1?

apis/oam/v1alpha2/core_types.go
```diff
@@ -0,0 +1,375 @@
...
+	// +optional
+	Scopes []ComponentScope `json:"scopes,omitempty"`
+}
+
+// An ApplicationConfigurationSpec defines the desired state of a
+// ApplicationConfiguration.
+type ApplicationConfigurationSpec struct {
+	// Components of which this ApplicationConfiguration consists. Each
+	// component will be used to instantiate a workload.
+	Components []ApplicationConfigurationComponent `json:"components"`
```
ryanzhang-oss: https://github.com/oam-dev/spec/pull/318.    Actually has an []scopes object at the root of appConfig too. What do you think of that? The main reason to have that is to avoid duplications in the components as there can have many components in a single scope.

apis/oam/v1alpha2/core_types.go
```diff
@@ -0,0 +1,375 @@
...
+	// A Trait that will be created for the component
+	// +kubebuilder:validation:EmbeddedResource
+	// +kubebuilder:pruning:PreserveUnknownFields
+	Trait runtime.RawExtension `json:"trait"`
+}
+
+// A ScopeReference refers to an OAM scope resource.
+type ScopeReference struct {
+	// APIVersion of the referenced scope.
+	APIVersion string `json:"apiVersion"`
```
ryanzhang-oss: I assume that this is the apiversion of a scope CR, not the scopeDefinition, right?

apis/oam/v1alpha2/core_types.go
```diff
@@ -0,0 +1,375 @@
...
+	Status WorkloadStatus `json:"status,omitempty"`
+}
+
+// An ApplicationConfigurationStatus represents the observed state of a
+// ApplicationConfiguration.
+type ApplicationConfigurationStatus struct {
+	runtimev1alpha1.ConditionedStatus `json:",inline"`
+
+	// Workloads created by this ApplicationConfiguration.
+	Workloads []ApplicationConfigurationWorkload `json:"workloads,omitempty"`
```
ryanzhang-oss: Are they workloads or components?

apis/oam/v1alpha2/core_types.go
```diff
@@ -0,0 +1,375 @@
...
+// An ApplicationConfigurationSpec defines the desired state of a
+// ApplicationConfiguration.
+type ApplicationConfigurationSpec struct {
+	// Components of which this ApplicationConfiguration consists. Each
+	// component will be used to instantiate a workload.
+	Components []ApplicationConfigurationComponent `json:"components"`
+}
+
+// A WorkloadReference refers to an OAM workload resource.
+type WorkloadReference struct {
```
ryanzhang-oss: do we need to define 3 identical references struct? There is another identical ResourceReference struct in the  core_workload_types.go

apis/oam/v1alpha2/core_types.go
```diff
@@ -0,0 +1,375 @@
...
+
+// ScopeDefinitionList contains a list of ScopeDefinition.
+type ScopeDefinitionList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []ScopeDefinition `json:"items"`
+}
+
+// A ComponentParameter defines a configurable parameter of a component.
+type ComponentParameter struct {
```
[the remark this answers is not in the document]
negz: I think we should remove `default` because you can set a default by simply including the value(s) the parameter would set in `spec.workload`. The `description` is still valuable though - will fix that.

apis/oam/v1alpha2/core_types.go
negz: Right!

apis/oam/v1alpha2/core_types.go
negz: I recall being on the fence about this. I think I went with three identical types because the docstring here ("A WorkloadReference refers to an OAM workload resource.") is used to generate CRD and API type documentation (e.g. https://doc.crds.dev/github.com/crossplane/crossplane and https://crossplane.io/docs/master/api.html).

On reflection, the docstring for the actual field that uses this type will be included too, so we can probably use a single type here.

apis/oam/v1alpha2/core_types.go
negz: This is an array of the CRs (like `ContainerizedWorkload`) that were rendered by the AppConfig. I believe we should call these workloads. This way the mental model is "an AppConfig uses components to produce workloads", rather than "an AppConfig uses components to produce components".

apis/oam/v1alpha2/core_types.go
negz: I don't feel super strongly, but I like that `appliesToWorkloads` is a little more explicit in the current implementation. I don't think what we name this field now really has to affect what we name any fields that configure policy around trait-to-workload compatibility.

apis/oam/v1alpha2/core_types.go
```diff
@@ -0,0 +1,375 @@
...
+	// Value to set.
+	Value intstr.IntOrString `json:"value"`
+}
+
+// A ComponentTrait specifies a trait that should be applied to a component.
+type ComponentTrait struct {
+	// A Trait that will be created for the component
+	// +kubebuilder:validation:EmbeddedResource
+	// +kubebuilder:pruning:PreserveUnknownFields
+	Trait runtime.RawExtension `json:"trait"`
```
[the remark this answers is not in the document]
negz: Yeah, this embeds a trait (e.g. `kind: ManualScalerTrait`), not a `kind: TraitDefinition`, so we don't have access to the `appliesTo` here. I think we need a little more thought to figure out how we should enforce `appliesTo`; I imagine the options will be constrained by how exactly we plan to apply traits to workloads.

apis/oam/v1alpha2/core_types.go
negz: I responded to the related thread at https://github.com/oam-dev/spec/pull/318/files#r384912298. I think this is a good idea. I might wait until that PR is approved before doing a final pass to bring this one into compliance with the spec. I'd also be happy to merge this PR sooner, and make any final corrections in a future PR/

apis/oam/v1alpha2/core_types.go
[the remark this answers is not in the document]
hasheddan: It appears both `default` and `description` are removed in https://github.com/oam-dev/spec/pull/316 so will need to update there.

apis/oam/v1alpha2/core_types.go
ryanzhang-oss: > I'd also be happy to merge this PR sooner, and make any final corrections in a future PR/

I don't mind getting this PR merged first as long as this won't affect the implementation of the AppConfig controller.

apis/oam/v1alpha2/core_types.go
ryanzhang-oss: In the future,  I envision a trait "appliesTo" a policy. Does this not conflict with this name?

apis/oam/v1alpha2/core_types.go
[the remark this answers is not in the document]
ryanzhang-oss: Is it possible to add an OwnerReference  to the CRD `kind: ManualScalerTrait`  to point back to the CR of `kind: TraitDefinition`?

apis/oam/v1alpha2/core_types.go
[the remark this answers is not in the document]
ryanzhang-oss: Updated, I added back `description` only.

apis/oam/v1alpha2/core_types.go
```diff
@@ -0,0 +1,375 @@
...
+type DefinitionReference struct {
+	// Name of the referenced CustomResourceDefinition.
+	Name string `json:"name"`
+}
+
+// A WorkloadDefinitionSpec defines the desired state of a WorkloadDefinition.
+type WorkloadDefinitionSpec struct {
+	// Reference to the CustomResourceDefinition that defines this workload
+	// kind.
+	Reference DefinitionReference `json:"definitionRef"`
```
ryanzhang-oss: I wonder if there is any reason to not keep the same name "definitionRef" as the json field name?

apis/oam/v1alpha2/core_types.go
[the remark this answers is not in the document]
wonderflow: make sense

apis/oam/v1alpha2/core_types.go
negz: It's Kubernetes convention for actual types and fields to use the full `Reference` name, rather than the abbreviated `Ref` that is used in the JSON fields.

apis/oam/v1alpha2/core_types.go
negz: I don't think so. The policies still describe workloads, right? The trait isn't applying to a policy, it's using policies to determine which workload it applies to.

Regardless of what we name this field now it will be a breaking API change (presuming OAM mirrors Kubernetes API versioning policy) to switch its meaning from a list of workload kinds to a list of policies. It would also be a breaking API change to remove this field and replace its functionality with a differently named field. Fortunately (again presuming we mirror Kubernetes policy) alpha APIs are allowed to make breaking changes without warning.

negz:
I'm going to close this PR in favor of https://github.com/crossplane/crossplane/pull/1276, which also includes the ApplicationConfiguration reconciler.

Issue #1238, opened 2020-02-06
Stacks pullPolicy for package and deployments images should be configurable


displague:
## What problem are you facing?

In development environments, it may be preferable for the `imagePullPolicy` of the `stack unpack` process and the stack deployment controller to be `Never` or `IfNotPresent`.

In [hosted environments](https://github.com/crossplaneio/crossplane/pull/1138), on the other hand, `Never` and `IfNotPresent` are not preferable.  `Always` would be a better choice.  This is apparent when considering that updates to a channeled StackInstall package (`:master`, `:latest`) could change remotely.

### How could Crossplane help solve your problem?

The requirements for this change and the UX pros and cons of this should be weighed.

Crossplane currently exposes a helm value which is similar in intent, [`image.pullPolicy`](https://github.com/crossplaneio/crossplane/blob/ffa2a85c44604b060d924a1fdf70d8f2fcc3cf08/cluster/charts/crossplane-controllers/values.yaml.tmpl#L10).  This value applies to the Crossplane and Stack Manager deployments.

One way to proceed would be for Crossplane to include a helm value that sets the default behavior for Stacks.  This could be a single value that applies to both [unpacks](https://github.com/crossplaneio/crossplane/blob/ffa2a85c44604b060d924a1fdf70d8f2fcc3cf08/pkg/controller/stacks/install/installjob.go#L89) and [stack controller images](https://github.com/crossplaneio/crossplane/blob/ffa2a85c44604b060d924a1fdf70d8f2fcc3cf08/pkg/controller/stacks/stack/stack.go#L689-L711).

Furthermore, it may be necessary to toggle this value per Stack.  This would be revealed as a StackInstall spec parameter, it does not need to be presented in the Stack because the unpack operation is a level above the Stack and the Stack deployment image is a byproduct of the unpack and stackinstall handling of that operation.

Related: https://github.com/crossplaneio/crossplane/issues/974, https://github.com/crossplaneio/crossplane/issues/1126, #505

displague:
I'm working on this as part of #505.

The current thinking is:

kind: StackInstall
spec:
  serviceAccount:
    imagePullSecrets:
    - foo
These imagePullSecrets will be included in the Stack resource and the Stack unpack job.

I've not yet considered how these imagePullSecrets will make their way into KubernetesApplication resources, per #505.

displague:
Closed by #1281

Issue #505, opened 2019-06-03
Support installation of stacks from private registries


jbw976:
**Is this a bug report or feature request?**
* Feature Request

**What should the feature do:** Currently, users can only install extensions from public registries.  They should be able to publish and install from private registries too, perhaps with `imagePullSecrets`.  Note that this would apply to multiple levels of packages/images, since all of the following can be different images:

* the Stack package (run in control plane)
* the controller (`Deployment` image) or other implementation "body" (run in control plane)
* workloads (`KubernetesApplication` component resources) that the controller generates (run in target clusters)

To put into a more concrete example using say Gitlab, this **could** be 3 separate images that could all come from a private registry:
* gitlab-package (that installs)
* gitlab-controller (that generates)
* gitlab-workload (that is propagated to a target cluster)

Note that the workload is propagated to a target cluster, so the private registry credentials would need to be propagated there also.

**What is use case behind this feature:** . Some users will not want to publish their extension package to a public registry, so we should support private registries as well.

glyn:
Crossplane may wish to take a look at [CNAB](https://cnab.io/) and its reference implementation as a standard alternative to some of the features of stacks.

In relation to this issue, CNAB supports the notion of _image relocation_ where a "bundle" of software referring to docker/OCI images in public or private registries is relocated to a (typically private) registry. For more on image relocation, see this [gentle overview](https://github.com/deislabs/cnab-spec/blob/master/805-airgap.md) in the spec and this more detailed [guide](https://github.com/deislabs/duffle/blob/ce81bb239c0b851e1997af3136b52957bdd2c0f5/docs/guides/relocation-guide.md) in the reference implementation.

glyn:
Let's take any discussion relating to my [earlier comment](https://github.com/crossplaneio/crossplane/issues/505#issuecomment-517811617) over to https://github.com/crossplaneio/crossplane/issues/666.

muvaf:
This is blocking Upbound registry to be used. Is there a timeline on when we could get that feature implemented? @jbw976

displague:
@muvaf I'll be taking this on over the next couple of weeks.
I'm tagging the relationship between this issue and https://github.com/crossplaneio/crossplane/issues/974 because we may need changes to the StackInstall or Stack type to accomodate this.

Documentation updates will be the primary output of this effort, whether or not Stack/StackInstall changes are necessary.

I expect that one path for using private registries is to take advantage of namespace wide configurations for `imagePullSecrets`:
[...]

PR #1281, opened 2020-02-19
Some Image and SA Stack properties configurable from StackInstall

### LINKED ISSUE #1242: Allow annotations to be passed from ClusterStackInstall to stack deployment's ServiceAccount

Some managed Kubernetes providers allow for authentication by associating IAM roles with Kubernetes `ServiceAccounts`. This is usually done by enabling the functionality in the Kubernetes cluster, then annotating the `ServiceAccount`. There are examples of how to do so on [AWS](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) and [GCP](https://cloud.google.com/blog/products/containers-kubernetes/introducing-workload-identity-better-authentication-for-your-gke-applications).

Currently, when a stack is installed with Crossplane, it is given a `ServiceAccount` and bound to appropriate `ClusterRole`. If a user wanted to provide an annotation on the `ServiceAccount`, they would need to manually access the cluster, provide the annotation, then potentially restart the pod that the stack controller is running in.

While this functionality is enabled through the steps outlined above, it would be useful to be able to pass down an annotation from the `ClusterStackInstall` to `ServiceAccount` that Crossplane creates for the stack. In clusters where this functionality is enabled, the stacks would be able to automatically authenticate to the provider without any additional credentials being required.

Ref: https://github.com/crossplaneio/stack-aws/pull/126

displague:

This PR makes the following `StackInstall` possible:

```yaml
apiVersion: stacks.crossplane.io/v1alpha1
kind: StackInstall
spec:
  # existing properties
  source: unchanged.com
  package: unchanged/unchanged
  #
  # everything below is new
  #
  imagePullSecrets:
  - name: this-is-new
  imagePullPolicy: Always
  serviceAccount:
    annotations:
      this.is.a: "new feature"
```

The Stack created by this StackInstall will be different in the following ways:
```yaml
apiVersion: stacks.crossplane.io/v1alpha1
kind: Stack
spec:
  controller:
    deployment:
      spec:
        template:
          spec:
            containers:
            - imagePullPolicy: Always
            imagePullSecrets:
            - name: this-is-new
  serviceAccount:
    annotations:
      this.is.a: "new feature"  
```
  
The `ServiceAccount` created to run the controller deployment will include the specified annotations.

The `Deployment` created for the Stack controller will include the specified imagePullPolicy and imagePullSecrets.

The `initContainer` of the `Job` used by the StackManager, which copies the stack package to a shared volume, will use the `imagePullPolicy` and `imagePullSecrets` specified in the StackInstall to access the stack image.

The `stack unpack` processing container of this `Job` will, however, rely on the same `imagePullPolicy` used on the StackManager.

Fixes #1242, #1238

Related to #505 

* Created a private repo on Docker Hub.
* Pushed the stack-wordpress image to this repo
* Created a Docker token and created a secret in the cluster for this token
  ```console
  kubectl create secret docker-registry regcred --docker-server=docker.io --docker-username=displague --docker-password=... --docker-email=marques@example.com
  ```
* applied the following yaml:
  ```yaml
  apiVersion: stacks.crossplane.io/v1alpha1
  kind: StackInstall
  metadata:
    name: stack-wordpress
  spec:
    package: "displague/stack-wordpress:0.1.0"
    imagePullSecrets:
    - name: regcred
    imagePullPolicy: Always
    serviceAccount:
      annotations:
        foo: bar
  ```

* Observed that the Job initContainer included the imagePullPolicy from the StackInstall (Always)
* Observed that the Job Container included the imagePullPolicy from Crossplane itself (Never)
* Observed the StackDefinition included the StackInstall fields under `controller`
* Observed the Stack included the StackInstall fields under `controller`
* Observed that the SA included the `foo:bar` annotation
* Observed that the Deployment included the StackInstall imagePullPolicy and imageSecrets

### Checklist

I have:
  appropriate.

[definition of done].


apis/stacks/v1alpha1/types.go
```diff
@@ -244,6 +270,29 @@ type StackSpec struct {
...
+			annotations[k] = v
+		}
+	}
+	return annotations
+}
+
+// ImagePullSecrets guarantees a slice of LocalObjectReference from the
+// StackSpec's ServiceAccount.ImagePullSecrets ready for use in Deployment and
+// ServiceAccount objects
+func (spec StackSpec) ImagePullSecrets() []corev1.LocalObjectReference {
```
displague: I could avoid defining `ImagePullSecrets` for both StackSpec and StackInstallSpec if the method acts on `ServiceAccountOptions`.  Doing so might invigorate the cyclomatic complexity check that I was trying to avoid by making this method available on the higher level Spec.  The `ServiceAccount` properties on these specs are nil'able, and the `if != nil` check increases complexity in the functions that rely on this.

apis/stacks/v1alpha1/types.go
```diff
@@ -118,6 +133,16 @@ func (si *ClusterStackInstall) SetConditions(c ...runtimev1alpha1.Condition) {
 	si.Status.SetConditions(c...)
 }
 
+// GetSpec gets the Spec of the StackInstall
+func (si *StackInstall) GetSpec() *StackInstallSpec {
```
displague: `GetSpec` is borrowed from #1264 and will be removed from this PR if that is merged first.

pkg/controller/stacks/install/stackinstall.go
```diff
@@ -373,7 +373,7 @@ func (h *stackInstallHandler) delete(ctx context.Context) (reconcile.Result, err
 
 	// Delete all StackDefintions created by this StackInstall or
 	// ClusterStackInstall
-	if err := h.kube.DeleteAllOf(ctx, &v1alpha1.StackDefinition{}, client.InNamespace(h.ext.GetNamespace()), client.MatchingLabels(labels)); runtimeresource.IgnoreNotFound(err) != nil {
+	if err := h.kube.DeleteAllOf(ctx, &v1alpha1.StackDefinition{}, client.InNamespace(h.ext.GetNamespace()), client.MatchingLabels(labels)); err != nil {
```
displague: The `DeleteAllOf` cleanup is in a separate commit and can be removed if it creates too much noise.

This is a response to https://github.com/crossplane/crossplane/pull/1258#discussion_r379293273 and 
https://github.com/crossplane/crossplane/pull/1217#discussion_r376425771

apis/stacks/v1alpha1/types.go
displague: I've borrowed a few more features from #1264 for modifying the unstructured Stack and StackDefinition resources. 

These will need to be rebased here once that is merged. 

In this PR, I've applied an improved pattern for making these modifications:
https://github.com/crossplane/crossplane/pull/1281/files#diff-46f16c3aed495fb04c3696c537ce7501R233-R437

pkg/controller/stacks/install/installjob.go
```diff
@@ -72,6 +73,12 @@ func createInstallJob(i v1alpha1.StackInstaller, executorInfo *stacks.ExecutorIn
 		namespace = o.Namespace
 	}
 
+	imagePullSecrets := i.GetSpec().ImagePullSecrets
+	imagePullPolicy := i.GetSpec().ImagePullPolicy
```
hasheddan: For both of these I believe you will either need to check `GetSpec()` for `nil` or make the `spec` required on `StackInstall` and `ClusterStackInstall`. We have opted to make it required on a number of are other CRDs, and it seems like it may make sense for both of these as well.

apis/stacks/v1alpha1/stackdefinition.go
```diff
@@ -27,11 +27,9 @@ import (
 
 // StackDefinitionSpec defines the desired state of StackDefinition
 type StackDefinitionSpec struct {
-	AppMetadataSpec `json:",inline"`
-	CRDs            CRDList         `json:"customresourcedefinitions,omitempty"`
-	Controller      ControllerSpec  `json:"controller,omitempty"`
-	Permissions     PermissionsSpec `json:"permissions,omitempty"`
-	Behavior        Behavior        `json:"behavior,omitempty"`
+	StackSpec `json:",inline"`
```
suskin: Has this already been tested? If I recall correctly, when I tried this in the past there was an issue with nesting multiple levels of `",inline"`.

pkg/controller/stacks/install/installjob.go
```diff
@@ -72,6 +73,12 @@ func createInstallJob(i v1alpha1.StackInstaller, executorInfo *stacks.ExecutorIn
...
+	if imagePullPolicy == "" {
```
suskin: Do we want this default? This seems like it will change the default behavior of image pulls to be different from what it was before, especially for [implicit update cases](https://kubernetes.io/docs/concepts/containers/images/#updating-images).

pkg/controller/stacks/install/installjob.go
```diff
@@ -213,29 +223,41 @@ func (jc *stackInstallJobCompleter) readPodLogs(namespace, name string) (*bytes.
...
 		return nil
 	}
 
-	// when the current object is a Stack object, make sure the name and namespace are
-	// set to match the current StackInstall (if they haven't already been set). Also,
-	// set the owner reference of the Stack to be the StackInstall.
-	if isStackObject(obj) || isStackDefinitionObject(obj) {
+	// Modify Stack and StackDefinition resources based on StackInstall
+	isStack := isStackObject(obj)
+	isStackDefinition := !isStack && isStackDefinitionObject(obj)
```
suskin: Do we need the `!isStack &&`? Can both be true?

pkg/stacks/executorinfo.go
```diff
@@ -54,23 +67,27 @@ func (eif *KubeExecutorInfoDiscoverer) Discover(ctx context.Context) (*ExecutorI
 	eif.mutex.Lock()
 	defer eif.mutex.Unlock()
 
-	if eif.Image != "" {
+	if eif.Image != "" && eif.ImagePullPolicy != "" {
```
suskin: What's this do? What's the benefit of having both of the checks?

pkg/stacks/pod_test.go
```diff
@@ -183,57 +183,93 @@ func TestGetSpecContainerImage(t *testing.T) {
...
 			if diff := cmp.Diff(got, tt.want.image); diff != "" {
-				t.Errorf("GetSpecContainerImage() got != want:\n%v", diff)
+				t.Errorf("GetContainerImage() got != want:\n%v", diff)
 			}
 		})
 	}
 }
 
-func TestGetContainerImage(t *testing.T) {
+func TestGetContainerImagePullPolicy(t *testing.T) {
```
suskin: I feel like we may want a test for when the image is valid and no pull policy is specified. I don't think I see one

pkg/controller/stacks/install/installjob.go
displague: I'll give `corev1.PullPolicy("")` a try. This should result in the value being omitted.

pkg/controller/stacks/install/installjob.go
displague: The unstructured `obj` can only have one `apiVersion` + `Kind` combination.  These functions are exclusive.

apis/stacks/v1alpha1/stackdefinition.go
displague: StackDefinition and the resulting Stack stack-wordpress both reconciled for me with this.

pkg/stacks/executorinfo.go
displague: ExecutorInfo is the image (and now imagepullpolicy) that the `stack unpack` job must run with. The Discoverer gathers this information from the running environment (presumably the `stack manage` process) through some self-identification requiring queries to the K8s API.

In debug settings, with no running pods, the Discoverer can be told (through environment variables) what the image (and now imagepullpolicy) should be.  Whether discovered from the running pod or provided through environment variables, these values become fields in the `stack unpack` `Job`.

The check at line 70 is a cache for the already discovered container information.

The idea with this change was to enforce that both Image and ImagePullPolicy are set.  I'll need to remove the ImagePullPolicy test since it makes sense to omit that for a default behavior, thanks for pointing this out through your comments.

pkg/controller/stacks/install/installjob.go
displague: I'd like to eventually clean this up by removing the Stack vs StackDefinition vs CRD guesswork.

https://github.com/crossplane/crossplane/compare/master...displague:unpack-with-structure?expand=1 (WIP)

pkg/controller/stacks/install/installjob.go
displague: GetSpec has been replaced with getters and setters that into spec.

pkg/controller/stacks/stack/stack.go
```diff
@@ -589,6 +590,16 @@ func (h *stackHandler) prepareDeployment(d *apps.Deployment) {
 	d.Spec.Template.SetName(name)
 	meta.AddLabels(&d.Spec.Template, matchLabels)
 	d.Spec.Selector = &metav1.LabelSelector{MatchLabels: matchLabels}
+	d.Spec.Template.Spec.ImagePullSecrets = h.ext.Spec.Controller.ImagePullSecrets
```
jbw976: one day i want to take a pass and remove all these lingering `ext` references in the code/structs. it's a lingering reminder of when stacks were called extensions :(

jbw976:
LGTM, I think @hasheddan comment was resolved, so hopefully he can change to "approved" also.

The codecov diff is failing now, is there some easy/obvious test cases that can added to increase coverage and make that pass again?  How are we treating codecov failures since adding it to the repo very recently? Is that a must pass check?

negz:
> How are we treating codecov failures since adding it to the repo very recently? Is that a must pass check?

I recently disabled the patch check; Codecov should now only fail if a PR reduces the overall project coverage by more than 2%. It could be that rerunning Jenkins will clear the failed check.

jbw976:
Cool thanks @negz! I am replaying the jenkins build to see what the overall project coverage looks like from this PR

jbw976:
looks like `codecov/project` is failing with -1.52%. where is that threshold configured? I thought 2% was the threshold value? I need to learn more about our new codecov setup. @negz?

jbw976:
nevermind @negz, @rathpc pointed me at the yaml config that is at the account level.  That's actually set to 1% now, so I have an understanding of why it's failing.

displague:
I got the test coverage up through more E2E testing. I would like to improve the smaller scoped E2E tests, like those that verify that unpack-job pod log yamls are ultimately set to the k8s API with the expected additions.

We can easily test that labels and names were set, but verifying that a deep value, like the map at `.spec.controller.deployment.spec.template.spec.containers[0].env` (I couldn't make this up) would greatly benefit from crossplane-runtime's pavement package, which will not be available until the next release.  

I also spotted some tests (TestCreateJobOutputObject for one) that verify `tt.obj == tt.want.obj` when `obj` was not directly modified. I see little value in this test aside from verifying that the function did not mutate obj values when an error condition was met.  These tests could be modified to not require `tt.want.obj` to be defined.

I would like to revisit the current tests in subsequent PRs.

PR #1286, opened 2020-02-19
Fix to parse only region from aws config file

sayanchowdhury:
Signed-off-by: Sayan Chowdhury <sayan.chowdhury2012@gmail.com>

The `cluster/examples/aws-setup-provider/setup.sh` script fails if the `~/.aws/config` file contains anything more than just region, for example my config file had `region` as well as `output`. This PR would just parse the region value and use it in `provider.yaml`

By successfully running the `cluster/examples/aws-setup-provider/setup.sh` script.

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/cluster/examples/aws-setup-provider/setup.sh
+++ b/cluster/examples/aws-setup-provider/setup.sh
@@ -36,7 +36,7 @@ kubectl cluster-info > /dev/null || echo "KUBECONFIG is not configured properly"
 aws_profile="${aws_profile:-default}"
 
 # if region is not provided, retrieve aws profile region from config
-AWS_REGION=$(awk '/["$aws_profile"]/ {getline; print $3}' ${HOME}/.aws/config)
+AWS_REGION=$(aws configure get region --profile $aws_profile)
 
 # retrieve aws profile credentials, save it under 'default' profile, and base64 encode it
 AWS_CREDS_BASE64=$(cat ${HOME}/.aws/credentials | awk '/["$aws_profile"]/ {getline; print $0}' | awk 'NR==1{print "[default]"}1' | base64 | tr -d "\n")


muvaf:
My config file is the following:
```
[default]
region = us-west-2
[profile crossplane-user]
region = us-east-1
```

Both versions print the following:
```
us-west-2
us-east-1
```

Which probably results in erroneous situation. It's not changed in this PR but @sayanchowdhury  would you like to add a fix to solve that too?

cluster/examples/aws-setup-provider/setup.sh
```diff
@@ -36,7 +36,7 @@ kubectl cluster-info > /dev/null || echo "KUBECONFIG is not configured properly"
 aws_profile="${aws_profile:-default}"
 
 # if region is not provided, retrieve aws profile region from config
-AWS_REGION=$(awk '/["$aws_profile"]/ {getline; print $3}' ${HOME}/.aws/config)
+AWS_REGION=$(cat ${HOME}/.aws/config | awk '/["$aws_profile"]/ {getline; print $0}' | awk '$1=="region"{print $3}')
```
muvaf: I just realized we use `aws configure get region --profile ${aws_profile}` for this in docs. I think that could be better. See https://crossplane.io/docs/v0.8/cloud-providers/aws/aws-provider.html

sayanchowdhury: @muvaf I've updated the code with your suggestion (which was brilliant). I did not understand what you meant by using it in the docs. Did you mean code?

muvaf: Oh I meant we're already using that command in the docs. For some reason, it's different how we populate `AWS_REGION` here in this script and in the docs that let you do these manually.

Issue #1291, opened 2020-02-20
template stacks: stacks should be able to specify controller version


muvaf:
### What problem are you facing?

If I deploy a template stack, the version of the generic templating-controller image comes from the Crossplane helm chart. However, I might want my stack to be pinned to use a fork/master version/older version no matter what Crossplane version is deployed.

### How could Crossplane help solve your problem?

It should be possible to embed the templating-controller image in the template stack `.registry/behavior.yaml` and this should be used, if it doesn't exist, then Crossplane's value should be used as fallback.

suskin:
> I might want my stack to be pinned to use a fork/master version/older version

Who is deciding which version to use? Is it a stack author, or the stack consumer? Would I want to be able to specify the image pull policy too?

muvaf:
> Who is deciding which version to use? Is it a stack author, or the stack consumer?

It's the stack author.

> Would I want to be able to specify the image pull policy too?

I don't think that's very necessary at this point. Image pull policies are usually used either in dev mode to use the cached images or to save disk space in the cluster; for both cases our default `IfNotExists` should be sufficient IMO. If you have a use-case, would love to hear.

PR #1294, opened 2020-02-21
template stacks: allow authors to specify controller image

muvaf:

This PR allows template stack authors to pin down the templating controller image they'd like. If not provided, it falls back to the default value on Crossplane helm chart.

Fixes https://github.com/crossplane/crossplane/issues/1291

* Crossplane master with minimal azure with and without `engine.controllerImage`
* This PR with minimal azure with and without `engine.controllerImage`

It's backward compatible.

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/apis/stacks/v1alpha1/stackdefinition.go
+++ b/apis/stacks/v1alpha1/stackdefinition.go
@@ -61,10 +61,16 @@ type BehaviorCRD struct {
 
 // StackResourceEngineConfiguration represents a configuration for a resource engine, such as helm2 or kustomize.
 type StackResourceEngineConfiguration struct {
+	// ControllerImage is the image of the generic controller used to reconcile
+	// the instances of the given CRDs. If empty, it is populated by stack manager
+	// during unpack with default value.
+	// +optional
+	ControllerImage string `json:"controllerImage,omitempty"`
 	// Type is the engine type, such as "helm2" or "kustomize"
 	Type string `json:"type"`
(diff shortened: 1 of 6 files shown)


muvaf:
> Do we have a concrete use-case for this?

Yes, as author of the stack, I'd like to be sure of the behavior that I want is exactly what happens. Think of it like go mod dependencies; we're pinning them down for getting the same binary everywhere. Without this PR, the deciding factor is whatever Crossplane version happens to be deployed in the cluster, which is not something you'd want because it could be old/new/buggy version of templating controller, not the one you tested against.

negz:
Would this allows folks who otherwise might not have RBAC permissions to run pods to do so? Or put otherwise, what could a malicious party do if they were able to specify their own template controller image?

muvaf:
> Would this allows folks who otherwise might not have RBAC permissions to run pods to do so? Or put otherwise, what could a malicious party do if they were able to specify their own template controller image?

@negz The RBACs are applied like any other stack. Currently, they would be able to use whatever image they want when they build a normal stack, meaning someone with `*StackInstall` permissions can actually run any pod they want but it'd be constrained with RBAC they defined in `app.yaml`. In template stacks, instead of `install.yaml`, users would specify the image in `behavior.yaml`. So, there is not much of a change in terms of RBACs.

muvaf:
@negz if you're onboard, could you merge this?

PR #1300, opened 2020-02-24
stacks: introduce packageType to app.yaml and Stack spec

displague:
### Description of your changes

Adds a new meta descriptor for stacks: `packageType`.

This property can be provided in `app.yaml` and will be represented in the Stack and StackDefinition resources.

This field is another descriptor, like Category or Keyword, but defines the nature of the package as belonging to one of three classes: Provider, Stack, or Application.

### Checklist

I have:
- [x] Run `make reviewable` to ensure this PR is ready for review.
- [x] Ensured this PR contains a neat, self documenting set of commits.
- [x] Updated any relevant [documentation] and [examples].
- [ ] Reported all new error conditions into the log or as an event, as
  appropriate.

For more about what we believe makes a pull request complete, see our
[definition of done].

[documentation]: https://github.com/crossplane/crossplane/tree/master/docs
[examples]: https://github.com/crossplane/crossplane/tree/master/cluster/examples
[definition of done]: https://github.com/crossplane/crossplane/tree/master/design/one-pager-definition-of-done.md

PR #1303, opened 2020-02-24
stacks: include apiVersion field in app.yaml

### LINKED ISSUE #900: Stack package format version

As Stacks evolve, the filesystem hierarchy and the contents of the `app.yaml` (the presence and significance of that file at all) are prone to change.  New metadata files may be introduced.

To future-proof the Crossplane Stack format, a version of the format would be nice so that we can know which version of the format a Stack package was built against.

1. The Stacks design should be updated with the definition of this Stack Format version 
  a. This should include details about what process (if any) is known or intended for introducing new versions and migrating between versions
1. Declare the version of the current stack format (should this be pinned to the design version?)
1. Add a new field that indicates the version of the actual markdown.
  a. Include a revision entry in the Stacks design for this new field (as should be done for all format changes)

displague:

This PR introduces the `apiVersion` field of `app.yaml`, expressing the [revision of the Stacks
spec](https://github.com/crossplane/crossplane/blob/master/design/design-doc-stacks.md#revisions) that the package conforms to.  Changes to the spec that affect the contents of the app.yaml file, including the replacement of the app.yaml file, will be denoted in the Stack design spec (or its successors). 

As stack packages adopt new standards they will embed the Stack design version in their metadata.

The change introduced here merely creates a new type, `PackageMetadataSpec`, extending the `AppMetadataSpec` with an additional `APIVersion` field.  The `app.yaml` file conforms to the
`PackageMetadataSpec` while the CRD `Stack` `Spec` conforms to the embedded `AppMetadataSpec`.

The public representation of a Stack does not need to include the internal packaging format from which that the Stack was derived.

No additional changes are included in this change as there are no conflicting Stack versions to warrant content conversion or alternate processing.

Fixes #900

_Names are hard. The `designVersion` field name here is not the winner of a popularity poll.  Some alternate choices were rejected for various reasons. I'm open to revisiting the name. 'designRevision', anyone?_

_* `schemaVersion` the version being described in this PR expresses changes to the app.yaml contents (where schema would apply) and the outer layers of packaging_
_* `packageVersion` while this would fit with `packageType` naming (#1300), `packageVersion` could be easily confused with the existing `version` field_

*Update:* After cycling through `designVersion` and `formatVersion`, `APIVersion` won.

### Checklist

I have:
  appropriate.

[definition of done].


apis/stacks/v1alpha1/types.go
```diff
@@ -305,6 +305,14 @@ type StackStatus struct {
 	ControllerRef                     *corev1.ObjectReference `json:"controllerRef,omitempty"`
 }
 
+// PackageMetadataSpec defines metadata about the stack application
+// and package contents
+type PackageMetadataSpec struct {
+	DesignVersion string `json:"designVersion,omitempty"`
```
suskin: What about `metaVersion`? o: it's clearly not for the content because of `meta`, but it doesn't tie us directly to the design. And it can encompass more than just the `app.yaml` because `meta` isn't specific.

muvaf:
Isn't this exactly what `apiVersion` in Kubernetes resources is for? It's used by [Helm](https://helm.sh/docs/topics/charts/#the-chart-yaml-file) and [Kustomize](https://github.com/kubernetes-sigs/kustomize/blob/master/docs/versioningPolicy.md#default-values-for-k8s-kind-and-apiversion), too.

displague:
@muvaf It is interesting that they both use `apiVersion` and that the Kustomize `apiVersion: kustomize.config.k8s.io/v1beta1` uses a URL format.

I'm interested in @suskin's take on this.

I don't think the name matters too much, which is why I threw this PR up with a stand-in name. But I know that we will be living with this name for a while and there are a few competing concepts in the crossplane ecosystem space regarding versions.  Hopefully our decision here will reduce confusion.

apis/stacks/v1alpha1/types.go
jbw976: i'd like to put a vote in for `formatVersion`, since it feels like that would cover the full spectrum of app metadata and package matadata versioning...the entire package **format version** :)

displague: `formatVersion` was on my original poll, so it has my blessing. I think it, like `metaVersion`, conveys that there is more to this than the app.yaml file.  
`designVersion` has the weakness that it alludes to `design` document (compared to a specification).
`apiVersion` also works, but I think it is a mixed blessing that both Helm and Kustomize already use the same name.

I'll update the PR to use `formatVersion` and see how that goes. :-)

suskin:
@displague @muvaf I was thinking about `apiVersion` too. In terms of representing the version of the schema, it seems like it has pretty much the same purpose in the context of kubernetes resources. There is a subtle difference for our use-case in the sense that we are also meaning for it to represent the version of the entire schema **and** the packaging format, and there's no packaging format for kubernetes resources. I thought Muvaffak's point about the [Helm](https://helm.sh/docs/topics/charts/#the-apiversion-field) `apiVersion` covering both the schema and the packaging format was a good point though. That seems the closest to what we're trying to represent, so that seems like an argument for using `apiVersion` (fwiw `apiVersion` is one of my top choices).

I think the strongest argument against using `apiVersion` is that we may eventually represent the stack metadata as an actual kubernetes object, at which point there would be an `apiVersion` which was **only** representing the schema of the kubernetes object, and not necessarily for the folder layout and package format. But maybe it'd be okay to tightly couple the schema format version and the package format version; maybe we don't need to be able to specify them independently.

jbw976:
I'm on board with everyone else's suggestion to use `apiVersion`.  I think that captures our intent fairly well and its a familiar term within the ecosystem. Thanks for the idea @muvaf!

can you make that change in this PR @displague?

jbw976:
Actually, one further question about `apiVersion`.  Would that make us expected to follow familiar api versioning naming, such as `v1alpha1`, instead of a `semver`? that may be a little odd for this use case of a packaging/schema/format version.

displague:
@jbw976 The apiVersion fields are not using the same format between Kustomize (`kustomize.config.k8s.io/v1beta1`) and Helm (`v2`). We would be introducing our own version scheme for use in `apiVersion`.

Do you have thoughts on the use of `semver`? We are not using semver in our design document on Stacks and by using `designVersion` (and `formatVersion`) I set the target version to match the design (`1.1`).

Should I change the design document to use a semantic version of `0.1.1`, or use this non-semantic version for the `apiVersion` field?

jbw976:
My suggestion would be:

* the version specified in `app.yaml` should be decoupled from the design doc.  I can see design doc changes happening that don't have direct impact on the api/schema/format of the package itself, so I don't think they should be explicitly coupled.
* Using a semver makes sense to me, it feels reasonable to start with an initial minor version of `0.1.0` for `apiVersion`.

Somewhat unrelated, but just wanted to clarify with you:

> We are not using semver in our [design document](https://github.com/crossplane/crossplane/blob/master/design/design-doc-stacks.md)

`1.1` is a semver, is it not?

displague:
> `1.1` is a semver, is it not? edit: maybe it's not since patch field MUST be specified

I'll concede that it is, but it means `1.1.0`.

https://semver.org/#spec-item-8
https://semver.org/#spec-item-7

> Minor version ...  It MAY include patch level changes.

displague:
Updated to use `APIVersion` (linter wants `APIVersion` over `ApiVersion`).

apis/stacks/v1alpha1/types.go
```diff
@@ -305,6 +305,14 @@ type StackStatus struct {
 	ControllerRef                     *corev1.ObjectReference `json:"controllerRef,omitempty"`
 }
 
+// PackageMetadataSpec defines metadata about the stack application
+// and package contents
+type PackageMetadataSpec struct {
```
jbw976: this is a new type that isn't specifically included in the StackSpec, since it doesn't have meaning outside of the context of a package's app.yaml.  That makes sense to not include this data in a Stack CR (which has it's own CRD versioning process, independent of app.yaml schema/format/apiVersion), although that was counterintuitive to me for a second.

Issue #1268, opened 2020-02-17
Render OAM ApplicationConfigurations


negz:
### What problem are you facing?

The [OAM Runtime Architecture document](https://docs.google.com/document/d/11AYNGhvry_B3l_tO3Yyv1f9tDoezZOpGIYV56oZwb5s/edit#heading=h.x458enhyy1sq) has tentatively been approved by the OAM maintainers. We'd like Crossplane to start adding OAM support to Crossplane, with a goal of Crossplane eventually becoming a production quality OAM runtime.

### How could Crossplane help solve your problem?

 The first engineering task in this process is to render OAM `ApplicationConfiguration`, outputting a series of distinct workload and trait CRs. The OAM second working draft proposal at https://github.com/oam-dev/spec/pull/304 is the authoritative source for the `ApplicationConfiguration` schema until the actual spec is updated.

PR #1305, opened 2020-02-26
Proposed POC OAM workflow design

hasheddan:

This one-pager is a very rough overview for a basic workflow that would happen in Crossplane after the implementation of the initial controllers for `ApplicationConfiguration` and `ContainerizedWorkload` resources. It is meant to be used as a reference for anyone who is not familiar with #1276 and #1268, as well as a collaboration document for those working on controllers in parallel.

n/a

[skip ci]

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/design/one-pager-oam-workflow.md
+++ b/design/one-pager-oam-workflow.md
@@ -1,0 +1,228 @@
+# OAM POC Workflow
+* Owner: Dan Mangum (@hasheddan)
+* Reviewers: Crossplane Maintainers
+* Status: Draft
+
+This document describes a basic workflow for utilizing OAM types as implemented
+in [1276](https://github.com/crossplane/crossplane/pull/1276). The controllers
+described are for an initial POC and will be moved out of core Crossplane in
+later iterations.
+
+## 1. Crossplane Installation
+
+The following OAM CRDs will be installed with Crossplane:
+
+- `ApplicationConfiguration`
+- `Component`
+- `TraitDefinition`
+- `ScopeDefinition`
+- `WorkloadDefinition`
+- `ContainerizedWorkload`
+
+Crossplane will also create a `WorkloadDefinition` *instance* that looks as
+follows:
+
+```yaml
+apiVersion: core.oam.dev/v1alpha2
+kind: WorkloadDefinition
+metadata:
+  name: containerizedworkloads.core.oam.dev
+spec:
+  definitionRef:
+    name: containerizedworkloads.core.oam.dev
+```
+
+This makes the `ContainerizedWorkload` core workload type available to be used
+by OAM `Components`. Other workload types can be created and made available by
+creating a CRD for the workload and a `WorkloadDefinition` that references it.
+
+The OAM-related controllers will also be started on installation:
+
+- **Application Configuration controller**: watches `ApplicationConfiguration`
+  resources and creates the corresponding workloads based on the inline
+  parameters and referenced `Components`
+- **Containerized Workload controller**: watches `ContainerizedWorkload`
+  resources and packages them into `KubernetesApplications` for scheduling to a
+  target cluster 
+
+## 2. Component Creation
+
+Users create a `Component` to define a unit of deployment that can be included
+in an `ApplicationConfiguration`. A `Component` must specify a
+`WorkloadDefinition`.
+
+```yaml
+apiVersion: core.oam.dev/v1alpha2
+kind: Component
+metadata:
+  name: myapp
+spec:
+  workload:
+    apiVersion: core.oam.dev/v1alpha2
+    kind: ContainerizedWorkload
+    metadata:
+      name: my-workload
+    spec:
+      osType: linux
+      containers:
+      - name: tbs11-app
+        image: hasheddan/tbs11:latest
+        resources:
+          cpu:
+            required: 1.0
+          memory:
+            required: 100MB
+        env:
+          - name: DB_HOST
+            valueFrom:
+              secretKeyRef:
+                name: mysqlconn
+                key: endpoint
+          - name: DB_USER
+            valueFrom:
+              secretKeyRef:
+                name: mysqlconn
+                key: username
+          - name: DB_PASSWORD
+            valueFrom:
+              secretKeyRef:
+                name: mysqlconn
+                key: password
+        ports:
+          - containerPort: 8080
+  parameters: 
+  - name: imageName
+    required: false
+    fieldPaths: 
+    - "spec.containers[0].image"
+```
+
+The creation of this `Component` does not trigger reconciliation for any
+Crossplane controller.
+
+
+## 3. Application Configuration Creation
+
+In order to deploy the `Component` described above, a user must create an
+`ApplicationConfiguration` that references it.
+
+```yaml
+apiVersion: core.oam.dev/v1alpha2
+kind: ApplicationConfiguration
+metadata:
+  name: myapp-dev
+spec:
+  components:
+    - componentName: myapp
+      parameterValues:
+        - name: imageName
+          value: hello-world:latest
+```
+
+The creation of an `ApplicationConfiguration` controller will queue a reconcile
+for its controller. The controller will create instances of the workload defined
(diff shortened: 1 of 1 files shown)


design/one-pager-oam-workflow.md
```diff
@@ -0,0 +1,227 @@
+# OAM POC Workflow
+* Owner: Dan Mangum (@hasheddan)
+* Reviewers: Crossplane Maintainers
+* Status: Draft
+
+This document describes a basic workflow for utilizing OAM types as implemented
+in [1276](https://github.com/crossplane/crossplane/pull/1276). The controllers described are for an initial POC and will be moved out of core
+Crossplane in later iterations.
+
+1. Crossplane Installation
```
hongchaodeng: Consider changing its format to Header 2?

hongchaodeng: Like
```suggestion
## 1. Crossplane Installation
```

hasheddan: Will do! Thanks for suggestion :)

design/one-pager-oam-workflow.md
```diff
@@ -0,0 +1,228 @@
...
+          secretKeyRef:
+            name: mysqlconn
+            key: password
+    ports:
+      - containerPort: 8080
+```
+
+The creation of the `ContainerizedWorkload` will also queue a reconcile of its
+controller. The controller will take the `ContainerizedWorkload` and package it
+into a `KubernetesApplication` with the following configuration:
```
hongchaodeng: I'm under the impression that AppConfig will replace KubernetesApplication. Why not take ContainerizedWorkload to create Deployment here?

hasheddan: As shown here, the only component in the `KubernetesApplication` is a `Deployment`. However, some traits may result in other components being created (i.e. an `Ingress`) that should be grouped with the `Deployment`. `KubernetesApplication` is a convenient way to group these resources and schedule them collectively. I *think* this may change in the future, but it appears to be a straight-forward workflow for POC. I'll tag in @negz here to add to / correct any of my assumptions here.

negz: > I'm under the impression that AppConfig will replace KubernetesApplication. 

@hongchaodeng That's not the case. In short, `KubernetesApplication` becomes an optional delivery mechanism for OAM workloads. Take a look at the [OAM Runtime Architecture](https://docs.google.com/document/d/11AYNGhvry_B3l_tO3Yyv1f9tDoezZOpGIYV56oZwb5s/edit#) diagram for more details.

hongchaodeng: > KubernetesApplication is a convenient way to group these resources and schedule them collectively

Isn't AppConfig enough to group resources collectively? Why add another layer?

> KubernetesApplication becomes an optional delivery mechanism for OAM workloads. 

I totally agree with that. But that should be another workload type.
The point is to use _ContainerizedWorkload_  to create Deployment directly without a middle layer.

design/one-pager-oam-workflow.md
```diff
@@ -0,0 +1,228 @@
+# OAM POC Workflow
+* Owner: Dan Mangum (@hasheddan)
+* Reviewers: Crossplane Maintainers
+* Status: Draft
+
+This document describes a basic workflow for utilizing OAM types as implemented
+in [1276](https://github.com/crossplane/crossplane/pull/1276). The controllers
+described are for an initial POC and will be moved out of core Crossplane in
+later iterations.
```
negz: Could you drop a link to the [OAM Runtime Architecture doc](https://docs.google.com/document/d/11AYNGhvry_B3l_tO3Yyv1f9tDoezZOpGIYV56oZwb5s/edit#heading=h.x458enhyy1sq) somewhere in here? It might be good to point out that this doc only covers the "remote workload" variant.

design/one-pager-oam-workflow.md
```diff
@@ -0,0 +1,228 @@
...
+          cpu:
+            required: 1.0
+          memory:
+            required: 100MB
+        env:
+          - name: DB_HOST
+            valueFrom:
+              secretKeyRef:
+                name: mysqlconn
+                key: endpoint
```
negz: Does `ContainerizedWorkload` support `valueFrom`?

design/one-pager-oam-workflow.md
negz: > But that should be another workload type. 

`KubernetesApplication` is a low level concept without separation of concerns, like a `Deployment`. So `KubernetesApplication` should not be a kind of workload any more than `Deployment` should be a kind of workload. We could arguably create something like `RemoteContainerizedWorkload`, but that would only control _where_ the workload runs (the local cluster where Crossplane is installed vs a remote one). I think it's better to use a single workload kind - `ContainerizedWorkload` - and use a scope to configure where it's deployed.

> The point is to use ContainerizedWorkload to create Deployment directly without a middle layer.

Is this required by the OAM spec? I thought it was up to each runtime how they rendered a core workload.

design/one-pager-oam-workflow.md
hasheddan: It does not currently in #1276 and the implementation there differs from the current [OAM Spec](https://github.com/oam-dev/spec/blob/master/3.component_model.md#env) as it has a `fromParam` field (although it is marked as not required). If we do end up adopting the full k8s definition of a `Container` then these fields would be present.

design/one-pager-oam-workflow.md
hongchaodeng: @negz I totally agree with your points. But I am wondering what's the motivation and additional value to add another layer (KubernetesApplication, or RemoteContainerizedWorkload) between AppConfig and Deployment?

resouer: I recall the design is `KubernetesApplication` (should give a better name later) only exists in 'stack-kubernetes-remote' as an internal envelope to deliver K8s resources to target clusters. So it seems this POC workflow is describing remote kubernetes (i.e. multi-cloud) scenario, right?

What will be the `stack-kubernetes-local` POC workflow?

hasheddan: @resouer I am not speaking authoritatively here, but for the short-term `KubernetesApplication` could be useful for the local story as well because it can schedule resources to *any* cluster, including the one that it is in :) This would allow us to have the same controllers handle both local and remote in the short-term. Does that like a useful solution to you?

resouer: @hasheddan I think @ryanzhang-oss has already begun to work on the local one which is nothing but a oam->k8s translator :-) An envelope will not make it simpler in my mind. 

We have no concern about two controllers in different scenario, it's by design according to our previous architecture meeting.

ryanzhang-oss: @hasheddan,  one thing that I still need to understand is if both the local and remote cluster handle `KubernetesApplication` then I am not sure what 's the difference between them. It seems that they have the same functionality in that case.

hasheddan: @resouer @ryanzhang-oss I think it is absolutely fine for the the local controller to translate directly to deployment. I think I added confusion by stating that a KubernetesApplication could be used to schedule resources to the local cluster. While this is true, it does not mean it is the right answer, especially if we are sticking to the proposed design of separate local and remote controllers.

I think the question that remains is how we want to handle this for the POC situation. This is why I mentioned that KubernetesApplication could be used to schedule both locally and remotely: we could essentially get both the local and remote story in the POC without having to build both parts (i.e. local is just a remote that is scheduled to the local cluster). We then could move the POC controller out into “remote stack” and use the controller @ryanzhang-oss is working on for “local stack”.

Certainly open to your thoughts and feedback here 🙂

ryanzhang-oss: @hasheddan  Sounds good, I think I will continue to work on the local controller that will translate OAM resources to Kubernetes native resources while you can work on the remote controller.

hasheddan: @negz if this sound good to you I think we can merge :+1:

PR #1306, opened 2020-02-28
stacks: fix stack pull secrets and policy propagation to deployment

displague:

Follow-up to #1281

The Stack controller was creating deployments using an extra piece of Stack spec defining imagePullSecrets and imagePullPolicy.

These extra pieces were not being set when the StackInstall handler created the Stack resource, which created a bug that prevented the Deployment from using the StackInstall specified imagePullSecrets and imagePullPolicy.

These fields can be considered extraneous because the StackInstall handler already embeds these values into deep Stack (and StackDefinition) Controller properties.

A user creating a Stack resource by hand should not need to specify these values outside of the Stack Controller fields where they are used.

The fix in this PR is to remove the extra fields from the Stack resource and simply rely on the already prepared DeploymentSpec in the Stack Controller.

Tests have been added to ensure that these Stack values are present in the Deployment.

Same as #1281.  I overlooked the final step (verifying the Deployment) in later builds of 1281 and missed the intra-PR-regression fixed and tested by this PR.

### Checklist

I have:
  appropriate.

[definition of done].


apis/stacks/v1alpha1/types.go
```diff
@@ -465,7 +465,9 @@ type ContributorSpec struct {
 // ControllerSpec defines the controller that implements the logic for a stack,
 // which can come in different flavors.
 type ControllerSpec struct {
-	StackControllerOptions `json:",inline"`
+	// ServiceAccount options allow for changes to the ServiceAccount
+	// the Stack Manager creates for the Stack's controller
+	ServiceAccount *ServiceAccountOptions `json:"serviceAccount,omitempty"`
```
muvaf: I don't think the field should be named as `serviceAccount` when it's actually `ServiceAccountOptions`, not actual `ServiceAccount`.

displague: @muvaf This is post merge but I still wanted to respond. I agree that the name "serviceAccount" may suggest that we have a complete service account object beneath this field, but counter examples are "Stack.Controller.Deployment" which is a DeploymentSpec.  There are other examples of a `*Spec` type being used with the field label omitting "spec" from the name.

muvaf: Understood, though I think we should change that one, too, in the future.

PR #1311, opened 2020-03-01
stacks: multiple namespace support

### LINKED ISSUE #1051: Support multiple instances of the same StackInstall / Stack in different namespaces

Part of #1036, Related to #1014

https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-stacks-security-isolation.md#multiple-versions: 

> As previously mentioned, since namespace stacks will be installed per namespace, multiple instances of each namespace stack can be installed in the control plane. This also means that multiple versions of the namespace stack can be installed side by side. Because of the namespaced isolation, this is fine for the namespace stack's controllers, but it does cause a potential difficulty for the CRDs.
> 
> The custom resource definitions themselves are always cluster scoped. If two versions of a namespace stack are installed in the control plane and they have identical CRD versions, there could be a conflict.

Support multiple instances of the same StackInstall / Stack in different namespaces.

https://github.com/crossplaneio/crossplane/issues/526 is a likely part of this solution.

displague:
Fixes #1051 

When the Stack controller reconciles a Stack it will look for the expected CRDs.  If the CRDs are not present, the reconcile will be requeued.

Any of the matching CRDs which contain the `app.kubernetes.io/managed-by: stack-manager` label (which StackInstall controller's `unpack` provides) will receive additional labels:
* `namespace.crossplane.io/{ns}: "true"` an existing label being deferred from unpack to stack reconciliation
* `parent.stacks.crossplane.io/{ns}-{name}: "true"` a new label to support multiple parents of a CRD

When a Stack is installed into a namespace the existing CRDs will gain additional copies of these labels, for each namespace or stack within that namespace.

When a Stack is deleted from a namespace the CRDs will have `parent.stacks.crossplane.io/{ns}-{name}` labels removed by the Stack controller. The `namespace.crossplane.io/{ns}` label will also be removed if there are no remaining `parent.stacks.crossplane.io/{ns}*` labels. 

When a Stack is deleted and a CRD with the `app.kubernetes.io/managed-by: stack-manager` label contains none of the these two types of labels, the CRD will be deleted by the StackInstall controller.

Honoring the existing `namespace.crossplane.io/{ns}: "true"` discoverability labels prevents this PR from deleting existing Stack CRDs.

We may also want to revisit the need for single parent labels which are currently used for detecting and deleting owned resources that cross the namespace/cluster scope boundary.

(Updated comment) Single parent labeling is modified in this PR. CRDs no longer receive these labels on unpack. When a StackInstall is removed, any CRD parent labels referring to that StackInstall are also removed from any matching CRDs.

If two StackInstalls sharing a common CRD are installed and deleted within a reconciliation window a race could have the newly reused CRD deleted before the new Stack can label it.

#### Upgrading Stack-Manager
1. Install the Wordpress and GCP Stack before this version of Stack-Manager is applied.
2. Update Stack Manager to this version.
   * observe existing CRDs are not deleted

#### Multiple Stacks

1. Install this version of the Stack Manager
1. Install Stack GCP into crossplane-system
1. Install Stack Wordpress into two namespaces
   
   * observe CRDs are given the labels mentioned in the description above
   ```
   $ kubectl get crd -o jsonpath='{.metadata.labels}' wordpressinstances.wordpress.samples.stacks.crossplane.io ; echo 
   map[app.kubernetes.io/managed-by:stack-manager crossplane.io/scope:namespace namespace.crossplane.io/default:true namespace.crossplane.io/precipice:true parent.stacks.crossplane.io/default-stack-wordpress:true parent.stacks.crossplane.io/precipice-stack-wordpress:true]
   ```
1. Delete one of these StackInstalls

   * observe CRDs lose the appropriate labels
   ```
   $ kubectl delete stackinstall -n default stack-wordpress
   $ kubectl get crd -o jsonpath='{.metadata.labels}' wordpressinstances.wordpress.samples.stacks.crossplane.io ; echo 
   map[app.kubernetes.io/managed-by:stack-manager crossplane.io/scope:namespace namespace.crossplane.io/precipice:true parent.stacks.crossplane.io/precipice-stack-wordpress:true]
   ```
1. Delete the other StackInstall
   
   * observe that the CRD has been deleted

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/pkg/controller/stacks/install/installjob.go
+++ b/pkg/controller/stacks/install/installjob.go
@@ -232,13 +232,12 @@ func (jc *stackInstallJobCompleter) createJobOutputObject(ctx context.Context, o
 		return nil
 	}
 
-	ns := i.GetNamespace()
-
 	// Modify Stack and StackDefinition resources based on StackInstall
 	isStack := isStackObject(obj)
 	isStackDefinition := !isStack && isStackDefinitionObject(obj)
 
 	if isStack || isStackDefinition {
+		ns := i.GetNamespace()
 		name := i.GetName()
 		if obj.GetName() == "" {
 			obj.SetName(name)
@@ -265,23 +264,9 @@ func (jc *stackInstallJobCompleter) createJobOutputObject(ctx context.Context, o
 		}
 	}
 
-	// We want to clean up any installed CRDS when we're deleted. We can't rely
-	// on garbage collection because a namespaced object (StackInstall) can't
-	// own a cluster scoped object (CustomResourceDefinition), so we use labels
-	// instead.
+	// TODO(displague) parentlabels can only express a single parent, CRDs
+	// may have multiple contributing "parents"
 	labels := stacks.ParentLabels(i)
-
-	// CRDs are labeled with the namespaces of the stacks they are managed by.
-	// This will allow for a single Namespaced stack to be installed in multiple
-	// namespaces, or different stacks (possibly only differing by versions) to
-	// provide the same CRDs without the risk that a single StackInstall removal
-	// will delete a CRD until there are no remaining namespace labels.
-	if isCRDObject(obj) {
-		labelNamespace := fmt.Sprintf(stacks.LabelNamespaceFmt, ns)
-
-		labels[labelNamespace] = "true"
-	}
-
 	meta.AddLabels(obj, labels)
 
 	jc.log.Debug(
--- a/pkg/controller/stacks/install/installjob.go
+++ b/pkg/controller/stacks/install/installjob.go
@@ -251,6 +251,9 @@ func (jc *stackInstallJobCompleter) createJobOutputObject(ctx context.Context, o
 			saAnnotationSetter(i.GetServiceAccountAnnotations()),
 		}
 
+		labels := stacks.ParentLabels(i)
+		meta.AddLabels(obj, labels)
+
 		// StackDefinition controllers need the name of the StackDefinition
 		// which, by design, matches the StackInstall
 		if isStackDefinition {
@@ -263,11 +266,6 @@ func (jc *stackInstallJobCompleter) createJobOutputObject(ctx context.Context, o
 		}
 	}
 
-	// TODO(displague) parentlabels can only express a single parent, CRDs
-	// may have multiple contributing "parents"
-	labels := stacks.ParentLabels(i)
-	meta.AddLabels(obj, labels)
-
 	jc.log.Debug(
 		"creating object from job output",
 		"job", job.Name,
@@ -275,7 +273,6 @@ func (jc *stackInstallJobCompleter) createJobOutputObject(ctx context.Context, o
 		"namespace", obj.GetNamespace(),
 		"apiVersion", obj.GetAPIVersion(),
 		"kind", obj.GetKind(),
-		"labels", labels,
 	)
 	if err := jc.client.Create(ctx, obj); err != nil && !kerrors.IsAlreadyExists(err) {
 		return errors.Wrapf(err, "failed to create object %s from job output %s", obj.GetName(), job.Name)
--- a/pkg/controller/stacks/install/installjob.go
+++ b/pkg/controller/stacks/install/installjob.go
@@ -28,7 +28,6 @@ import (
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
-	apiextensions "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -308,16 +307,6 @@ func isStackDefinitionObject(obj stacks.KindlyIdentifier) bool {
 
 	return gvk.Group == v1alpha1.Group && gvk.Version == v1alpha1.Version &&
 		strings.EqualFold(gvk.Kind, v1alpha1.StackDefinitionKind)
-}
-
-func isCRDObject(obj runtime.Object) bool {
-	if obj == nil {
-		return false
-	}
-	gvk := obj.GetObjectKind().GroupVersionKind()
-
-	return apiextensions.SchemeGroupVersion == gvk.GroupVersion() &&
-		strings.EqualFold(gvk.Kind, "CustomResourceDefinition")
 }
 
 func setupStackDefinitionController(obj *unstructured.Unstructured, modifiers ...stackSpecModifier) error {
--- a/pkg/controller/stacks/install/stackinstall.go
+++ b/pkg/controller/stacks/install/stackinstall.go
@@ -298,8 +298,11 @@ func (h *stackInstallHandler) create(ctx context.Context) (reconcile.Result, err
 		// there is no install job created yet, create it now
 		job := h.createInstallJob()
 
-		// remove any stale jobs that would prevent the stack from installing
-		// correctly
+		// if an install job with our name already exists, compare the labels
+		// (specifically parent labels). If they match, adopt this job - we must
+		// have failed to update the jobref on a previous reconciliation.
+		// if the install job does not belong to this stackinstall, block
+		// reconciliation and report it until the problem is resolved.
 		existingJob := &batchv1.Job{ObjectMeta: metav1.ObjectMeta{Name: job.Name, Namespace: job.Namespace}}
 
 		switch err := h.hostKube.Get(ctx, types.NamespacedName{Name: job.Name, Namespace: job.Namespace}, existingJob); {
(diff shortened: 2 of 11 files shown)


displague:
I'm opening this up for review. I'll be adding some more tests and docs but the code is in the expected final state (pending review feedback).

displague:
Griping about "GoLangCI: Processing Timeout." here: https://github.com/golangci/golangci-worker/issues/56#issuecomment-596196843

pkg/controller/stacks/stack/stack.go
```diff
@@ -45,7 +46,11 @@ import (
 )
 
 const (
-	stacksFinalizer = "finalizer.stacks.crossplane.io"
+	stacksFinalizer    = "finalizer.stacks.crossplane.io"
+	namespaceMember    = "true"
```
jbw976: nit: when i see these consts later on in the code, they look like variables that could contain any type of value that isn't always super clear from just reading the code.  perhaps a naming scheme like `labelValue*` could make that more clear? e.g., `labelValueActiveParentStack`. 

definitely a nit pick, not blocking.

pkg/controller/stacks/stack/stack.go
```diff
@@ -235,102 +257,237 @@ func copyLabels(labels map[string]string) map[string]string {
...
+}
+
+// createPersonaClusterRolesCRDHandler provides a handler which creates admin,
+// edit, and view clusterroles that are namespace+stack+version specific
+func (h *stackHandler) createPersonaClusterRolesCRDHandler() crdHandler {
+	labels := stacks.ParentLabels(h.ext)
+
+	return func(ctx context.Context, crds []apiextensions.CustomResourceDefinition) error {
+
+		for persona := range roleVerbs {
```
jbw976: this func looks like a lot of new logic from the diff, but I think that's probably not the case and it's mostly moved/refactored logic.  This section of permissions management is the one i'd be the most concerned about for regression risk, so hopefully we have done the testing validation to ensure this still works as expected.

pkg/controller/stacks/stack/stack_test.go
```diff
@@ -1688,6 +1701,657 @@ func Test_stackHandler_prepareHostAwarePodSpec(t *testing.T) {
...
+			wantErr: fmt.Errorf(`Missing CRD with APIVersion %q and Kind %q`, apiVersion, kind+"z"),
+		},
+		{
+			name:    "Success",
+			args:    args{v1alpha1.CRDList{metav1.TypeMeta{Kind: kind, APIVersion: apiVersion}}, []apiextensionsv1beta1.CustomResourceDefinition{crd(withCRDGroupKind(group, kind), withCRDVersion(version))}},
+			wantErr: nil,
+		},
+		{
+			name:    "SuccessMultiple",
+			args:    args{v1alpha1.CRDList{metav1.TypeMeta{Kind: kind, APIVersion: apiVersion}, metav1.TypeMeta{Kind: kind + "z", APIVersion: apiVersion}}, []apiextensionsv1beta1.CustomResourceDefinition{crd(withCRDGroupKind(group, kind), withCRDVersion(version)), crd(withCRDGroupKind(group, kind+"z"), withCRDVersion(version))}},
```
jbw976: consider making these really long lines to be multi-line for easier readability, it's hard to tell what objects are involved in each test case

pkg/controller/stacks/stack/stack_test.go
```diff
@@ -1688,6 +1701,657 @@ func Test_stackHandler_prepareHostAwarePodSpec(t *testing.T) {
...
+func clusterRole(name string, crm ...crModifier) rbac.ClusterRole {
+	c := rbac.ClusterRole{}
+	c.SetName(name)
+	for _, m := range crm {
+		m(&c)
+	}
+	return c
+}
+
+func Test_crdListFulfilled(t *testing.T) {
```
jbw976: lots of test cases and coverage in this file, thank you!!

pkg/controller/stacks/install/installjob.go
```diff
@@ -265,23 +263,9 @@ func (jc *stackInstallJobCompleter) createJobOutputObject(ctx context.Context, o
 		}
 	}
 
-	// We want to clean up any installed CRDS when we're deleted. We can't rely
-	// on garbage collection because a namespaced object (StackInstall) can't
-	// own a cluster scoped object (CustomResourceDefinition), so we use labels
-	// instead.
+	// TODO(displague) parentlabels can only express a single parent, CRDs
```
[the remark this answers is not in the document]
displague: Per our conversation, I'm going to not add the "single parent" labels on CRDs here.  We will continue to use the other resources that unpack handles (Stack, StackDefinition).  We may revisit the use of single parent labels in that scenario later.

I'm also going to remove the existing "single parent" labels on StackInstall, to avoid leaking labels.  The StackInstall controller is the referent. When a StackInstall is being deleted it can search for any CRDs that match its single-parent labels (similar to the DeleteAll CRD call that is being removed by this PR) and then remove that set of labels.

This can be optimized to only take place after CRDs lacking multiple-parent labels are deleted.  This label cleanup code can be removed in the future since future stackinstalled CRDs will not contain this set of single parent labels.

pkg/controller/stacks/stack/stack.go
displague: I agree that the refactor here is hard to follow by diff.

#### Before

`create` calls `processRBAC`
`processRBAC` was calling `createPersonaClusterRoles`
`createPersonaClusterRoles` was fetching all CRDs using `crdsFromStack` before doing role specific work
`crdsFromStack` called `crdListsDiffer` which was a naive `len(fetched) == len(stack.spec.crds)` check

#### After

`create` calls `processRBAC` (no change)
`create` calls `processCRDs` (new)
`processRBAC` no longer calls `createPersonaClusterRoles`
`processCRDs` fetches all CRDs using `crdsFromStack` (improved for multi-parented and unmanaged CRDs)
`processCRDs` walks a list of handlers which include:
* `createPersonaClusterRoles` (inner crd-loop is unmodified, outer crd fetch + loop + len-check is not needed)
* multi-parent labeller (new)
* namespace discovery label (moved here away from unpack, because the managed crd may already exist)

`crdsFromStack` now verifies that the expected version of the CRD exists the fetched CRD
`crdsFromStack` is now usable for deletion where some CRDs could already be missing
`processCRDs` uses `crdListFulfilled` to verify that all expected CRDs are present (managed or not)

pkg/controller/stacks/install/stackinstall.go
```diff
@@ -297,7 +298,21 @@ func (h *stackInstallHandler) create(ctx context.Context) (reconcile.Result, err
 		// there is no install job created yet, create it now
 		job := h.createInstallJob()
 
-		if err := h.hostKube.Create(ctx, job); err != nil {
+		// remove any stale jobs that would prevent the stack from installing
```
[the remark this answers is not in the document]
displague: I changed the behavior without updating the comment.

The StackInstall will adopt an existing job if it matched the parent labels of our StackInstall (the job was created but the SI controller errored while updated the StackInstall).

The StackInstall will fail to reconcile until the user deletes any conflicting install job that may exist.

I'll revise the comment to say as much.

pkg/controller/stacks/install/stackinstall_test.go
```diff
@@ -622,3 +631,225 @@ func TestHandlerFactory(t *testing.T) {
...
+	got.SetResourceVersion(want.GetResourceVersion())
+
+	if diff := cmp.Diff(want, got, test.EquateConditions()); diff != "" {
+		t.Errorf("-want, +got:\n%s", diff)
+	}
+}
+
+func assertNoKubernetesObject(t *testing.T, g *GomegaWithT, got runtime.Object, unwanted metav1.Object, kube client.Client) {
+	n := types.NamespacedName{Name: unwanted.GetName(), Namespace: unwanted.GetNamespace()}
+	g.Expect(kube.Get(ctx, n, got)).To(HaveOccurred())
```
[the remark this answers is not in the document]
displague: I took a stab at this, but within this function I can't replicate the parameters needed to match `kerrors.NewNotFound()`.  And after looking into how to do this with Gometa, I started wondering why we are introducing Gomega at all.  This `assertNoKubernetesObject` mirrors `assertKubernetesObject` which were already present in either stackinstall_test or stack_test (I don't recall which came first).

In a future PR I think we can remove this because I don't think we need the Gomega semantics and detailed reported over what is already available and used elsewhere with `Diff`.

pkg/controller/stacks/stack/stack.go
```diff
@@ -116,6 +121,13 @@ func Setup(mgr ctrl.Manager, l logging.Logger, hostControllerNamespace string) e
 		log:          l.WithValues("controller", name),
 	}
 
+	if err := mgr.GetFieldIndexer().IndexField(&apiextensions.CustomResourceDefinition{}, crdAPIGroupField, func(rawObj runtime.Object) []string {
```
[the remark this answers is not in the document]
displague: It may be possible to create a pseudo-index (like a SQL composite index) here which could be used to avoid the extra field filters happening when we do use the MatchingFields query. I'm not sure if that is possible though and I kept this simple for now.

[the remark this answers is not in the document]
displague: Adding `//	Use of MatchingFields on fields not indexed by the API requires a cache index`

pkg/controller/stacks/stack/stack_test.go
displague: I may have wrapped some lines that were not in my PR in an attempt to clean this up visually.

pkg/controller/stacks/stack/stack_test.go
displague: I think this latest commit may bump the coverage up even higher (by hitting the added code and code that was not previously covered due to the refactoring needed to placate the cyclomatic complexity of the stackinstall delete function).

pkg/controller/stacks/install/stackinstall_test.go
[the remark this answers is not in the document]
jbw976: I am always in support of removing Gomega usage.

PR #1314, opened 2020-03-02
azure provider: replace UI instructions in provider setup with admin grant command

muvaf:

Our docs has UI instructions for this process but apparently there is a command in Azure CLI to do this.

 I've tested it in https://github.com/crossplane/stack-minimal-azure/pull/7 via Wordpress Stack.

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/docs/cloud-providers/azure/azure-provider.md
+++ b/docs/cloud-providers/azure/azure-provider.md
@@ -14,20 +14,20 @@ This assumes that you have already [set up the Azure CLI client](https://docs.mi
 
 Create a JSON file that contains all the information needed to connect and authenticate to Azure:
 
-```console
+```bash
 # create service principal with Owner role
 az ad sp create-for-rbac --sdk-auth --role Owner > crossplane-azure-provider-key.json
 ```
 
 Take note of the `clientID` value from the JSON file that we just created, and save it to an environment variable:
 
-```console
+```bash
 export AZURE_CLIENT_ID=<clientId value from json file>
 ```
 
 Now add the required permissions to the service principal that will allow it to manage the necessary resources in Azure:
 
-```console
+```bash
 # add required Azure Active Directory permissions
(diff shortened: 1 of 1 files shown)


docs/cloud-providers/azure/azure-provider.md
```diff
@@ -41,26 +41,17 @@ You might see an error similar to the following, but that is OK, the permissions
...
 ```
 
-After these steps are completed, you should have the following file on your local filesystem:
-
-* `crossplane-azure-provider-key.json`
-
-## Grant Consent to Application Permissions
+Finally, you need to grant admin permissions on the Azure Active Directory to the service principal because it will need to create other service principals for your `AKSCluster`:
+```bash
+# gra
```
jbw976: what is this `gra` comment supposed to mean? is it truncated for a more useful comment that starts with `grant...`? :)

docs/cloud-providers/azure/azure-provider.md
```diff
@@ -41,26 +41,17 @@ You might see an error similar to the following, but that is OK, the permissions
...
+az ad app permission admin-consent --id "${AZURE_CLIENT_ID}"
```
jbw976: this single step using `az` appears to be much less complicated than the UI sequence, which is awesome.

In your testing, did you do this with a new account that didn't already have these perms enabled? i'm hoping we have confidence that this step is reliable for everyone, as opposed to being vetted only by the account we normally use for testing that already had this set.

muvaf: By account, do you mean user account or service principal that is created in the first step?

I think the account I used is an old one with `Global Administrator` role but I created a new service principal in the first step and granted admin-consent to that new service principal with this command. As long as the account that user logged in for `az` command has the `Global Administrator` role, which is mentioned as a note, I think it shouldn't matter. Do you have a different experience?

docs/cloud-providers/azure/azure-provider.md
muvaf: Thanks for catching! Fixed.

docs/cloud-providers/azure/azure-provider.md
jbw976: I think you're right about that distinction between user account and service principal, i think your instructions are the streamlined CLI equivalent to what we had previously for GUI steps.  Great!! 🎉

PR #1315, opened 2020-03-02
Move TargetRef to K8s App Spec and update workload controllers to modern patterns

hasheddan:

Because `targetRef` cannot be reliably reproduced for `KubernetesApplication` and `KubernetesApplicationResource`, it should be in the `spec` instead of the `status`. While making this update, I also noticed that we had not moved to using `subresource:status` for our `workload` types. The types have been updated and the controllers are refactored to accommodate and implement patterns that are more consistent with rest of the Crossplane ecosystem.

As a byproduct of moving the `targetRef` to the `spec`, users and systems can now pick a specific `KubernetesTarget` directly as opposed to previously only being able to rely on the scheduler matching based on labels.

These controllers have significant unit testing implemented, but further manual testing will be included here as it is executed.

### Checklist

I have:
  appropriate.

[definition of done].


hasheddan:
@negz it would probably be good to add eventing and some other more modern patterns here as well. I am happy to either include in this PR or follow-up with those changes.

hasheddan:
@negz SGTM! Thanks for the review :)

muvaf:
Is this tested manually? Just wanted to check because spec/status call stuff is hard to catch in unit tests.

hasheddan:
@muvaf yep! I will also make sure to get the follow-up PR in soon and show some additional testing there 👍

Issue #1308, opened 2020-02-29
Stacks: Combine stack image references (unpack and controller image should be the same)


suskin:
### What problem are you facing?

As we have been using our stacks and managing them, we have discovered that having two references to each stack image is making things unnecessarily error-prone for us, and we aren't using the feature at all.

The feature works as follows. A stack image is specified in two places:

* On install, for the unpack phase
* In configuration (either `install.yaml` or `behavior.yaml`, depending on the type of stack), for the controller phase

Theoretically decoupling these two allows the user to pull the envelope (the image used for unpacking) from a different place or a different image from where the controller is pulled from, and that seems useful. But we haven't been using it yet, so we would like to not use that pattern in our codebase and revisit it later. And, since we plan to remove it from our codebase everywhere, we should probably just disable the feature until we revisit it later.

### How could Crossplane help solve your problem?

* Always pull the behavior source image (in the case of `behavior.yaml`) or the controller source image (in the case of `install.yaml`) from the stack install object, if the image is not specified in the configuration. These are changes in the stack manager.
* Add basic unit tests (one for each path) to check that this works
* Remove the source image name from `behavior.yaml` for all the stacks that we maintain
* Remove the source image name from `install.yaml` for all the stacks that we maintain
* Update the stack documentation to match the new behavior
* Update any relevant crossplane-cli scaffolding

prasek:
@suskin looks like we'll need to remove [image: "crossplane/stack-gcp:master"](https://github.com/crossplane/provider-gcp/blob/4dc7426f1e10542b19de1e8069ba6e35665e335e/config/stack/manifests/install.yaml#L20) from the install.yaml of the various providers as well.

this will simplify things quite a bit so looking forward to getting this in the near future, thanks!

//cc @rathpc @thephred @jbw976

prasek:
@displague sounds like `Stack Manager` could simply overwrite the controller image e.g. `image: "crossplane/stack-gcp:master"` in the [install.yaml](https://github.com/crossplane/provider-gcp/blob/4dc7426f1e10542b19de1e8069ba6e35665e335e/config/stack/manifests/install.yaml#L20) automatically for all stacks (essentially forcing all stacks to use the stack image as the controller image), and this change would be fairly isolated.

suskin:
[...]

PR #1319, opened 2020-03-02
unpack: set app metadata version from docker tag if not present

### LINKED ISSUE #1307: Stack manager: Support reading stack version from docker tag

As part of simplifying how we manage versions for our stacks, we would like to omit the stack version from the `app.yaml`, and just use the version from the docker tag if it exists. It appears that the stack manager does not currently support this, in the sense that it does not pull the version from the docker tag and put it on the `Stack` or `StackDefinition` object (I was testing with a locally built `stack-minimal-gcp`).

* On install, the stack manager should grab the stack version from the docker tag if it is not present in the app.yaml
* The stack manager should place this grabbed version on the `Stack` or `StackDefinition` object that it creates
* We should add a unit test to verify that this behavior works
* We should update the stack design doc to reflect this behavior

suskin:

Fixes #1307 . We want our stack metadata's version to come from the
docker tag if it isn't set in the stack's `app.yaml`. This makes it more
convenient to manage lots of versions and release channels as a stack
author.

I took a look at the stack design documentation, and it already describes this as the intended behavior for the version field, so I didn't update any of the documentation.

* I added a unit test case to exercise this logic inside the unpack logic
* I have tested ad-hoc locally to exercise the environment variable injection code

### Checklist

I have:
  appropriate.

[definition of done].


pkg/stacks/unpack_test.go
```diff
@@ -1176,7 +1243,10 @@ func TestUnpack(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			got := &bytes.Buffer{}
 			rd := &walker.ResourceDir{Base: tt.root, Walker: afero.Afero{Fs: tt.fs}}
-			err := Unpack(rd, got, tt.root, "Namespaced", "crossplane/ts-controller:0.0.0")
+
+			os.Setenv(StackImageEnv, "crossplane/sample-stack:0.0.1")
+			err := Unpack(rd, got, tt.root, "Namespaced", "crossplane/ts-controller:0.0.0", logging.NewLogrLogger(zap.Logger(true)))
```
displague: Unless we are testing that the logger received a log message, I think it is safe to replace `logging.NewLogrLogger(zap.Logger(true))` with `logging.NewNopLogger()`

pkg/stacks/unpack_test.go
```diff
@@ -1176,7 +1243,10 @@ func TestUnpack(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			got := &bytes.Buffer{}
 			rd := &walker.ResourceDir{Base: tt.root, Walker: afero.Afero{Fs: tt.fs}}
-			err := Unpack(rd, got, tt.root, "Namespaced", "crossplane/ts-controller:0.0.0")
+
+			os.Setenv(StackImageEnv, "crossplane/sample-stack:0.0.1")
```
displague: I think the tests would benefit from `StackImageEnv`'s value being set in the `tests` struct. We can toggle different values for each row in the the same test table.  As a string, the default value (for the existing tests that don't specify `tt.stackImageEnv`) will be "", which is the same as being unset.

pkg/stacks/unpack_test.go
```diff
@@ -973,6 +976,54 @@ var (
 	_ StackPackager = &StackPackage{}
 )
 
+func simpleAppFileNoVersion(permissionScope, packageType string) string {
```
displague: What do you think about making `simpleAppFile` take a `version` parameter and including the version snippet only when version is not `""`?

pkg/stacks/unpack.go
```diff
@@ -519,8 +541,9 @@ func NewStackPackage(baseDir, tmplCtrlImage string) *StackPackage {
 //
 // baseDir is expected to be an absolute path, i.e. have a root to the path,
 // at the very least "/".
-func Unpack(rw walker.ResourceWalker, out io.StringWriter, baseDir, permissionScope string, tsControllerImage string) error {
-	sp := NewStackPackage(filepath.Clean(baseDir), tsControllerImage)
+func Unpack(rw walker.ResourceWalker, out io.StringWriter, baseDir, permissionScope string, tsControllerImage string, log logging.Logger) error {
+	l := log.WithValues("operation", "unpack")
```
displague: I was thinking about this earlier today! Glad to see unpack getting logs.

Is there any configuration that would result in these log messages disrupting the yaml in the pod log that we unpacked?

pkg/stacks/unpack.go
```diff
@@ -220,6 +229,18 @@ func (sp *StackPackage) IsNamespaced() bool {
 func (sp *StackPackage) SetApp(app v1alpha1.AppMetadataSpec) {
 	app.DeepCopyInto(&sp.Stack.Spec.AppMetadataSpec)
 
+	if sp.Stack.Spec.AppMetadataSpec.Version == "" {
+		iv := os.Getenv(StackImageEnv)
```
displague: It will likely be a pain, but did you look for ways to plumb `StackImageEnv` through `StackPackage` and other parent types?

I'm not blocking on this, just curious.  The precedence is in your favor with `os.Getenv` already being used in `pkg/stacks/executorinfo.go` and `pkg/stacks/pod.go`. Ideally we would only mess with environment variables in main.go, via kingpin.

pkg/stacks/unpack.go
jbw976: in general, logging from the unpacking job will make the creation of the unpacked objects fail.  The expectation by the stack manager is that it can take the entirety of the unpack job's logs (stdout) and send them directly to the k8s api server.  If log statements are intermixed with that output of object yaml's, then something in the create flow will fail.

This seems to already been known and "documented" in https://github.com/crossplane/crossplane/blob/master/pkg/controller/stacks/install/installjob.go#L109-L111, so perhaps the policy that all logging in the unpack job should be at Debug level will work for us.

pkg/controller/stacks/install/installjob.go
```diff
@@ -122,6 +122,12 @@ func prepareInstallJob(p prepareInstallJobParams) *batchv1.Job {
 									MountPath: "/ext-pkg",
 								},
 							},
+							Env: []v1.EnvVar{
+								{
+									Name:  "STACK_IMAGE",
```
jbw976: could this not use the same const `StackImageEnv`?

pkg/stacks/unpack.go
displague: Is there a way to configure the logger to prepend all message newlines with `#` so they are ignored by the yaml parser?

displague: https://github.com/kubernetes/kubernetes/issues/28167 seems related (asks for the log output to be stderr vs stdout identifiable) but I don't see any answers there.

displague: https://blog.sandipb.net/2018/05/03/using-zap-creating-custom-encoders/ has an example of using a custom log encoder.

pkg/stacks/unpack_test.go
```diff
@@ -1176,7 +1243,10 @@ func TestUnpack(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			got := &bytes.Buffer{}
 			rd := &walker.ResourceDir{Base: tt.root, Walker: afero.Afero{Fs: tt.fs}}
-			err := Unpack(rd, got, tt.root, "Namespaced", "crossplane/ts-controller:0.0.0")
+
+			os.Setenv(StackImageEnv, "crossplane/sample-stack:0.0.1")
+			err := Unpack(rd, got, tt.root, "Namespaced", "crossplane/ts-controller:0.0.0", logging.NewLogrLogger(zap.Logger(true)))
+			os.Unsetenv("MY_IMAGE")
```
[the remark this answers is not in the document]
suskin: Yeah - good catch

pkg/controller/stacks/install/installjob.go
suskin: Hmm good point

pkg/stacks/unpack_test.go
suskin: Hmm yeah. When I was originally considering this I was imagining a boolean for including the version, and I got hung up by the version being in the middle of the file for some reason, but now I think we could just put it at the end instead.

pkg/stacks/unpack_test.go
suskin: Ah, yeah, that's a better approach

pkg/stacks/unpack_test.go
suskin: I considered this, but I decided to put a real logger in so that the log output could be used to help debug test failures.

pkg/stacks/unpack.go
suskin: Yeah - it'd be great to have a way to log and have the job be functional, but having logs we can enable is a good first step. I do not plan to do a longer-term solution in this PR

pkg/stacks/unpack.go
```diff
@@ -220,6 +229,18 @@ func (sp *StackPackage) IsNamespaced() bool {
...
+
+		if strings.Contains(iv, ":") {
```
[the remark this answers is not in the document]
suskin: Oo. I did not know about these; thanks! That seems like a reasonable utility to use

pkg/stacks/unpack.go
suskin: I did not, because of the precedents, but I agree with the principle of not touching environment variables down in the code. Yeah, ideally we'd pull them out in main and pass around some sort of configuration instead.

suskin:
> So to make sure I understand fully, this change essentially saves the stack image tag version into the stack CR, if no explicit value is provided in app.yaml.

Yep! Exactly.

> This PR (intentionally) does not address using this same version for the stack's controller also, correct? We expect that to be addressed with another PR for #1308?

Correct. Since they can be done separately, I opted for smaller PRs so we can make incremental progress faster.

pkg/stacks/unpack_test.go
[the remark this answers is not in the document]
suskin: This is done; resolving

pkg/stacks/unpack_test.go
suskin: This is done; resolving

pkg/stacks/unpack.go
```diff
@@ -220,6 +229,18 @@ func (sp *StackPackage) IsNamespaced() bool {
...
+			// If iv is ":", the result of Split will be ["", ""]
+			ivVersion := strings.Split(iv, ":")[1]
+			sp.log.Debug("No stack version found in app metadata; reading version from environment instead", "versionFromEnvironment", ivVersion)
+			sp.Stack.Spec.AppMetadataSpec.Version = ivVersion
+
```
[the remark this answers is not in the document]
suskin: Done; resolving

pkg/stacks/unpack.go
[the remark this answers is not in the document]
suskin: I changed the implementation to be essentially the same, but to use the `reference` library instead.

pkg/controller/stacks/install/installjob.go
suskin: This is changed; resolving

pkg/stacks/unpack_test.go
```diff
@@ -1023,7 +1021,18 @@ permissionScope: %q
 
 # License SPDX name: https://spdx.org/licenses/
 license: Apache-2.0
-`, packageType, permissionScope)
+`
+
+	if includeVersion {
+		appFile = appFile + `
```
golangcibot: assignOp: replace `appFile = appFile + `
# Version of project (optional)
# If omitted the version will be filled with the docker tag
# If set it must match the docker tag
version: 0.0.1
    `` with `appFile += `
# Version of project (optional)
# If omitted the version will be filled with the docker tag
# If set it must match the docker tag
version: 0.0.1
    `` (from `gocritic`)

jbw976: the formatting is screwy here with it being rendered as markdown, but the bot (and [failed GolangCI run](https://golangci.com/r/github.com/crossplane/crossplane/pulls/1319)) is saying to change `appFile = appFile +` to `appFile +=`

PR #1328, opened 2020-03-06
Add support for single stack image

suskin:

Related to #1308 - see that ticket for more background.

This change adds the ability to omit the container image name for the controller image (in the case of standard stacks) or behavior image (in the case of template stacks). We want to do this to simplify container image management, because we usually use the same container image for both the envelope image and for the controller/behavior image, and it's inconvenient for the envelope image to have a file in it with an explicit reference to the envelope image tag.

The changes are fully backward-compatible.

Unit tests have been added:

* Unpack functionality is tested with the image omitted, to make sure it still functions well
* Stack/StackDefinition object creation is tested with the image omitted from the input, to make sure that the image is injected properly

I have also tested this with a running Crossplane locally:

* With a locally-built version of stack-gcp with the image removed from the `install.yaml`. The Stack showed up with the image in the right place as expected.
* With a locally-built version of stack-minimal-gcp with the image removed from the `behavior.yaml` `source`. The Stack showed up with the image in the right place, as expected.

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/pkg/controller/stacks/install/installjob.go
+++ b/pkg/controller/stacks/install/installjob.go
@@ -247,14 +247,7 @@ func (jc *stackInstallJobCompleter) createJobOutputObject(ctx context.Context, o
 			obj.SetNamespace(ns)
 		}
 
-		// The package isn't *quite* what we will use as the image
-		pkg := i.GetPackage()
-		stackImg, err := i.ImageWithSource(pkg)
-		if err != nil {
-			// Applying the source is best-effort
-			jc.log.Debug("not applying stack envelope image source to stack controller image due to error", "pkg", pkg, "err", err)
-			stackImg = pkg
-		}
+		stackImg := i.GetPackage()
 
 		modifiers := []stackSpecModifier{
 			controllerImageInjector(stackImg),
@@ -447,6 +440,12 @@ func controllerImageSourcer(src imageWithSourcer) stackSpecModifier {
 // if there are two sources of truth instead of a single source of truth.
 func controllerImageInjector(stackImage string) stackSpecModifier {
 	return func(spec *v1alpha1.StackSpec) error {
+		// If the stack image is empty, we don't need to propagate an empty string
+		// down into more fields
+		if stackImage == "" {
+			return nil
+		}
+
 		if d := spec.Controller.Deployment; d != nil {
 			spec := &d.Spec.Template.Spec
 
--- a/pkg/controller/stacks/install/installjob.go
+++ b/pkg/controller/stacks/install/installjob.go
@@ -247,7 +247,17 @@ func (jc *stackInstallJobCompleter) createJobOutputObject(ctx context.Context, o
 			obj.SetNamespace(ns)
 		}
 
+		// The package isn't *quite* what we will use as the image
+		pkg := i.GetPackage()
+		stackImg, err := i.ImageWithSource(pkg)
+		if err != nil {
+			// Applying the source is best-effort
+			jc.log.Debug("not applying stack envelope image source to stack controller image due to error", "pkg", pkg, "err", err)
+			stackImg = pkg
+		}
+
 		modifiers := []stackSpecModifier{
+			controllerImageInjector(stackImg),
 			controllerPullSetter(i.GetImagePullPolicy(), i.GetImagePullSecrets()),
 			controllerImageSourcer(i),
 			saAnnotationSetter(i.GetServiceAccountAnnotations()),
@@ -425,6 +435,34 @@ func controllerImageSourcer(src imageWithSourcer) stackSpecModifier {
 				cs[i].Image = img
 			}
 		}
+		return nil
+	}
+}
+
+// controllerImageInjector adds a stack image to a stack or stack definition
+// spec, if there isn't one specified. The reason this exists is so that a stack
+// author can use the same image for their stack envelope (the image which is unpacked)
+// and their stack controller (the image which runs in the cluster), because that
+// is a common pattern, and it's inconvenient to manage the image names separately
+// if there are two sources of truth instead of a single source of truth.
+func controllerImageInjector(stackImage string) stackSpecModifier {
+	return func(spec *v1alpha1.StackSpec) error {
+		if d := spec.Controller.Deployment; d != nil {
+			spec := &d.Spec.Template.Spec
+
+			for i := range spec.InitContainers {
+				if spec.InitContainers[i].Image == "" {
+					spec.InitContainers[i].Image = stackImage
+				}
+			}
+
+			for i := range spec.Containers {
+				if spec.Containers[i].Image == "" {
+					spec.Containers[i].Image = stackImage
+				}
+			}
+		}
+
 		return nil
 	}
 }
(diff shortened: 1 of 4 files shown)


pkg/controller/stacks/install/stackinstall_test.go
```diff
@@ -91,6 +91,19 @@ func withStackRecord(stackRecord *corev1.ObjectReference) resourceModifier {
 	return func(r v1alpha1.StackInstaller) { r.SetStackRecord(stackRecord) }
 }
 
+// withPackage allows a test to set a StackInstaller's package
+// Another option would have been to modify the interface to allow this,
+// but it is preferable if we treat the package field as immutable.
+func withPackage(pkg string) resourceModifier {
+	return func(r v1alpha1.StackInstaller) {
+		if si, ok := r.(*v1alpha1.StackInstall); ok {
```
displague: `StackInstaller` has `GetPackage`, but not `SetPackage`.  Most of the methods it has today have Get+Set, but we haven't needed SetPackage until now.

You could add `SetPackage` to the interface and implement it for `StackInstall` and `ClusterStackInstall` using either side of your conditional here, and then `withPackage` just calls `SetPackage`.

pkg/stacks/unpack_test.go
```diff
@@ -969,122 +957,174 @@ status:
...
-
-	if includeVersion {
-		appFile += `
-# Version of project (optional)
-# If omitted the version will be filled with the docker tag
-# If set it must match the docker tag
-version: 0.0.1
-    `
+	if controllerImage != "" {
+		// The spaces are used for formatting the next line. This is a quick and dirty way
```
displague: If you replace:
```yaml
              %sname: sample-stack-controller
```
with 
```yaml
              %s#
              name: sample-stack-controller
```

Then this could become:
```go
imageTmpl := ""
if controllerImage != "" {
  imageTmpl = fmt.Sprintf("image: %s\n", controllerImage)
}
return fmt.Sprintf(tmpl, imageTmpl)
```

which would render the following (yaml doesn't care about the indentation of comment lines, right?):

```yaml
              image: foo/bar
#
              name: sample-stack-controller
```

pkg/controller/stacks/install/installjob_test.go
```diff
@@ -251,6 +355,85 @@ status:
  conditionedStatus: {}
 `
 
+	stackRawNoControllerImage = `---
```
displague: Did you want to make these functions too?  I'm fine with the blobs of text if not. It's messy but status quo.

pkg/controller/stacks/install/installjob.go
```diff
@@ -247,7 +247,17 @@ func (jc *stackInstallJobCompleter) createJobOutputObject(ctx context.Context, o
 			obj.SetNamespace(ns)
 		}
 
+		// The package isn't *quite* what we will use as the image
+		pkg := i.GetPackage()
+		stackImg, err := i.ImageWithSource(pkg)
```
displague: Do we really need to `ImageWithSource` here? I think this is redundant to the `controllerImageSourcer` step a few lines below.

displague: It is possible that `GetPackage` is empty (unrealistic now, but was the intent for crd-lookup stackinstalls).  I don't know if this is the first place that would be making the assertion that all StackInstalls include a `package`, but we may want to avoid pushing the `image` deeper into the controller spec if the `package` is empty.

This may be undefined behavior that happens to result in the same outcome, so perhaps there is nothing to really do about this, now.

suskin: Good point - this does seem redundant. I can't remember why I originally felt like this was needed now. I'll make a couple adjustments.

pkg/controller/stacks/install/installjob_test.go
suskin: I was thinking about it, but ended up not doing it because there was less to modify in this file. I will think about it a little more.

pkg/controller/stacks/install/stackinstall_test.go
suskin: I considered adding `SetPackage`, but I decided not to because I liked the idea of treating more fields as immutable. I also did not like the idea of adding methods to the interface just to make the testing more convenient, so I figured I would do this until we came up with a non-testing use-case for changing the interface.

pkg/stacks/unpack_test.go
suskin: Yeah, that's true. I think that would work too. I decided to do it the way I've done it because I felt like I preferred the approach that kept the output looking nice, even if the formatting code looks a little weird.

pkg/controller/stacks/install/installjob_test.go
suskin: I decided to do this; resolving.

pkg/controller/stacks/install/installjob.go
suskin: This is removed now; resolving

muvaf:
With this PR, we recommend to people to write an invalid `Deployment` object as `install.yaml` because image fields won't be given, right? If this is the case, I'm not sure if this is the right approach.

Currently, `install.yaml` almost never changes if you ship one image(all of our normal stacks). I'd imagine that authors wouldn't create it at all in normal cases and we can generate that `Deployment` during installation. But if they do want to specify their own `install.yaml` (maybe they want the controller to run as `StatefulSet`), we could provide some env var-like parameters they can use in their `install.yaml`, such as `{{ .StackImage }}`, that we fill during the installation.

We can go with a similar approach for template stacks, too; let them use a placeholder like `{{ .StackImage }}` that Crossplane replaces.

displague:
@muvaf Another approach that was being considered was treating an omitted install.yaml as deferring to the stack image and tag for a deployment.
https://github.com/crossplane/crossplane/issues/1308#issuecomment-595255880

The push that has been driving this is to simplify publishing, complexity, and perhaps ultimately security surface. It started by removing the need to include a tag (inheriting from the package), here the image, next could be the ability to use the Deployment spec's environment variables, volumes, and command line arguments.

Supporting a full deployment spec permits use of valueFrom fields and perhaps other unwanted capabilities.  We've been winding back the Stacks controller controls to give us more environment control.

That said, if we would parameterize the image name, we should also parameterize the image tag, which would justify reverting #1319. This would make the default controller image `latest`, which is less surprising, and would make the inherited package tag only available through explicit templating.

muvaf:
Just discovered that all this is already discussed in https://github.com/crossplane/crossplane/issues/1308#issuecomment-595581664

I'll keep my earlier comment there for future iterations.

pkg/stacks/unpack_test.go
```diff
@@ -673,167 +645,210 @@ status:
...
+
+	if sourceImage != "" {
+		return fmt.Sprintf(tmpl, fmt.Sprintf("image: %s", sourceImage))
+
+	}
+
+	return fmt.Sprintf(tmpl, "")
+}
+
+func simpleAppFile(permissionScope, packageType string, includeVersion bool) string {
```
jbw976: i like this change to have functions to generate the largely common output and inject in the variables, so we aren't duplicating huge chunks of yaml everywhere.

jbw976:
Ah yes, looks like you'll follow up on the checklist in https://github.com/crossplane/crossplane/issues/1308#issuecomment-596010800. thanks @suskin!

PR #1329, opened 2020-03-07
illustrate trait ownership reference

ryanzhang-oss:
Illustrate the idea to register any trait CRD with the traitDefinition CR which contains the appliesTo field. 

Break the OAM kernel resources into their own type definition files

This PR is not supposed to compile but I am not sure why `make reviewable` throws an error on me. I was expecting some compile errors but I got something like `no rule to make target generate`

### Checklist

I have:
  appropriate.

[definition of done].


apis/oam/v1alpha2/component_types.go
```diff
@@ -0,0 +1,90 @@
...
+	Workload runtime.RawExtension `json:"workload"`
+
+	// Parameters exposed by this component. ApplicationConfigurations that
+	// reference this component may specify values for these parameters, which
+	// will in turn be injected into the embedded workload.
+	// +optional
+	Parameters []ComponentParameter `json:"parameters,omitempty"`
+}
+
+// A ComponentStatus represents the observed state of a Component.
```
hasheddan: CI is failing because `ComponentStatus` is not embedded in the `Component` type, so the conditioned status getters and setters are failing in `method.go`: https://github.com/crossplane/crossplane/pull/1329/files#diff-6d36f2c62d0bed5dae064d7385c56346R55-R63

```
/home/upbound/go/src/github.com/crossplane/crossplane/apis/oam/v1alpha2/methods.go:62:5: cm.Status undefined (type *Component has no field or method Status)
```

pkg/controller/oam/traitdefinition/traitdefinition_controller.go
```diff
@@ -0,0 +1,89 @@
...
+	newCRD := traitCRD.DeepCopy()
+	newCRD.ObjectMeta.ManagedFields = nil
+	newCRD.ObjectMeta.OwnerReferences = []metav1.OwnerReference{
+		{
+			APIVersion: traitDefinition.APIVersion,
+			Kind:       traitDefinition.Kind,
+			Name:       traitDefinition.Name,
+			UID:        traitDefinition.UID,
+			Controller: &falseVar,
+		},
```
wonderflow: what benifits from doing this?

apis/oam/v1alpha2/component_types.go
ryanzhang-oss: This is not supposed to even compile. This just illustrates the idea (per discussion with @negz)

pkg/controller/oam/traitdefinition/traitdefinition_controller.go
ryanzhang-oss: We lose any information stored in the "TraitDefinition". For example, the "appliesTo" info cannot be retrieved from the application configuration through a trait CR.

wonderflow: I see, make sense. 👍

wonderflow: But I'm not sure whether it's a proper way. \cc @hongchaodeng

ryanzhang-oss: I know K8s has some corky convention but I only know how to express the idea in a general `computer science` way.

pkg/controller/oam/traitdefinition/traitdefinition_controller.go
```diff
@@ -0,0 +1,89 @@
...
+	ctx := context.Background()
+	log := r.Log.WithValues("traitDefinition definition", req.NamespacedName)
+
+	log.Info("reconciling trait definition")
+
+	var traitDefinition corev1alpha2.TraitDefinition
+	if err := r.Get(ctx, req.NamespacedName, &traitDefinition); err != nil {
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+	log.Info("Get a traitDefinition", "definition ref", traitDefinition.Spec.Reference.Name)
```
hongchaodeng:
```suggestion
	log.Info("Get a traitDefinition", "definition_ref", traitDefinition.Spec.Reference.Name)
```

pkg/controller/oam/traitdefinition/traitdefinition_controller.go
```diff
@@ -0,0 +1,89 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	corev1alpha2 "github.com/crossplane/crossplane/apis/oam/v1alpha2"
+)
+
+// TraitDefinitionReconciler reconciles a TraitDefinition object
+type TraitDefinitionReconciler struct {
+	client.Client
```
hongchaodeng: Better add a `Client` field. Embed is only good for _inheritance_ scenario.

pkg/controller/oam/traitdefinition/traitdefinition_controller.go
```diff
@@ -0,0 +1,89 @@
...
+
+func (r *TraitDefinitionReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
+	ctx := context.Background()
+	log := r.Log.WithValues("traitDefinition definition", req.NamespacedName)
+
+	log.Info("reconciling trait definition")
+
+	var traitDefinition corev1alpha2.TraitDefinition
+	if err := r.Get(ctx, req.NamespacedName, &traitDefinition); err != nil {
+		return ctrl.Result{}, client.IgnoreNotFound(err)
```
hongchaodeng: If ignoring not found err, we need to add Finalizer and handle delete event.

pkg/controller/oam/traitdefinition/traitdefinition_controller.go
```diff
@@ -0,0 +1,89 @@
...
+	newCRD.ObjectMeta.OwnerReferences = []metav1.OwnerReference{
+		{
+			APIVersion: traitDefinition.APIVersion,
+			Kind:       traitDefinition.Kind,
+			Name:       traitDefinition.Name,
+			UID:        traitDefinition.UID,
+			Controller: &falseVar,
+		},
+	}
+	traitCRD.Annotations["lastModifiedTime"] = time.Now().String()
```
hongchaodeng: It's more idiomatic to use CustomResourceDefinitionCondition in status to write lastModifiedTime.

negz:
>  I was expecting some compile errors but I got something like no rule to make target generate

Our `Makefile` depends on a git submodule of make libraries. 🙃I believe running plain old `make` will setup that submodule for you, after which `make reviewable` should work.

ryanzhang-oss:
> Secondly, I don't think it helps us relate an _instance_ of (e.g.) a Trait back to the appropriate `TraitDefinition` very much. As far as I know the only way to get the CRD for a particular kind of resource is to look up the CRD by its well known name, which MUST be `<plural>.<apigroup>`, e.g. `manualscalertraits.core.oam.dev`. Under this PR we'd look up the CRD by this well known name, then follow the CRD's owner reference back to the corresponding `TraitDefinition`.

I am glad that you get the point of this illustration PR.  AFAIK, the name of the CRD corresponding to a CR is defined by the K8s convention (given a CR of type G.V.K,  the CRD is called  Ks(pural) and its G.V = `CustomResourceDefinition` and K = `CustomResourceDefinition` ).  Is that the `well known` name you refer to here?  I

ryanzhang-oss:
> The v1alpha2 OAM spec currently _recommends_ that a `WorkloadDefinition` `metadata.name` match its `spec.definitionRef.name`. If we could _require_ this property for `WorkloadDefinition`, `TraitDefinition`, and `ScopeDefinition`

I guess that would also work although it kind of feels like a "weak" link instead. Is there any other way to link the definition of a CRD back to the CR that refers to it?

pkg/controller/oam/traitdefinition/traitdefinition_controller.go
ryanzhang-oss: This is auto-generated by kubebuilder.... What tools do u use to scaffolding?

negz:
> Is that the well known name you refer to here?

Yes. Given a custom resource of `kind: FooTrait` in any version of API group `core.oam.dev` the CRD that defines that custom resource will always be named `footraits.core.oam.dev`, which is the "well known name" I'm referring to.

> Is there any other way to link the definition of a CRD back to the CR that refers to it?

Not that I'm aware of. I've [asked the API machinery folks](https://kubernetes.slack.com/archives/C0EG7JC6T/p1583796550166200) if there is another way.

negz:
I believe this PR was opened just to illustrate a concept, and was never really intended to be merged. I'm going to close it.

PR #1332, opened 2020-03-11
KubernetesApplication controller should always attempt to sync all resources

hasheddan:

Currently, if any `KubernetesApplicationResource` that is owned by the `KubernetesApplication` being reconciled is unable to be synced, the controller will not attempt to sync subsequent resources. This could result in a "healthy" resource never being synced because an "unhealthy" one causes the controller to error early.

Added additional unit tests, and deployed a `KubernetesApplication` into a remote cluster where I knew some of the resources would fail. Prior to the change, I would get varied number of submitted resources on each reconcile and `STATUS` was always `Failed`:

```
NAME        CLUSTER                                STATUS               DESIRED   SUBMITTED
app-48081   a0a7ad39-d26b-4682-a982-aac77f10da1c   Failed   69        13
```

After the change, I got consistent number of submitted resources and `STATUS` was always `PatiallySubmitted`:

```
NAME        CLUSTER                                STATUS               DESIRED   SUBMITTED
app-48081   a0a7ad39-d26b-4682-a982-aac77f10da1c   PartiallySubmitted   69        59

```

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/pkg/controller/workload/kubernetes/application/application.go
+++ b/pkg/controller/workload/kubernetes/application/application.go
@@ -18,6 +18,7 @@ package application
 
 import (
 	"context"
+	"fmt"
 	"strings"
 	"time"
 
@@ -100,6 +101,8 @@ func (c *localCluster) sync(ctx context.Context, app *v1alpha1.KubernetesApplica
 	app.Status.DesiredResources = len(app.Spec.ResourceTemplates)
 	app.Status.SubmittedResources = 0
 
+	var errs []error
+
 	// Garbage collect any resource we control but no longer have templates for.
 	if err := c.gc.process(ctx, app); err != nil {
 		return v1alpha1.KubernetesApplicationStateFailed, errors.Wrap(err, errGarbageCollect)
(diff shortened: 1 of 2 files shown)


pkg/controller/workload/kubernetes/application/application.go
```diff
@@ -281,3 +288,10 @@ func getControllerName(obj metav1.Object) string {
 
 	return c.Name
 }
+
+func condenseErrors(errs []error) error {
+	if len(errs) == 0 {
+		return nil
+	}
+	return fmt.Errorf("%s", errs)
+}
```
hasheddan: I think this could be moved to `crossplane-runtime`. I have wanted similar functionality elsewhere before.

negz: Aren't there a handful of relatively popular Go libraries that handle this already? I'm not the biggest fan of condensing multiple error together, and tend to try to avoid it as much as possible.

PR #1333, opened 2020-03-11
stacks: add core APIGroup restriction option for Stack controller

### LINKED ISSUE #899: Restrict a Stack's access to core types when the stack claims they "depend on" them

As per the [stacks security and isolation design](https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-stacks-security-isolation.md#allowed-resource-access), it is intended to restrict access to the resources that a Stack can perform operations on.  Specifically, a stack only gets access to the CRDs that it owns or declares that it depends on.

However, we don't discriminate at all about what resources a Stack claims it depends on.  I believe that if a stack declares it "depends on" volumes or pods, then we'd be OK with that and create the necessary RBAC to allow that access.

**Important**: Must support a run mode compatible with running Crossplane directly in the target cluster such that Crossplane stacks can manage Kubernetes core types including `Deployment`, `Ingress`, `Services`.

We should consider black or white listing the API groups or resources that stacks can get access to, even if they claim they depend on them.  If a stack says it depends on volumes or pods, we should not grant them RBAC access to them.

The code that would need to be updated lives here:
https://github.com/crossplaneio/crossplane/blob/master/pkg/stacks/unpack.go#L359-L376

displague:

Fixes #899

I originally wrote this supporting documentation https://gist.github.com/displague/3b6368c7eca459961d018b5a9408f5c8
 into the Stacks design doc as a way to guide this PR.  I have since found that it would be more at home in the Stacks Isolation design doc or the Crossplane.io Stacks docs.  I'm looking for ideas on where to weave this in.

This PR adds the new `stack manage --restrict-core-apigroups` argument which when enabled prevents certain rules signatures from being used in Stack `spec.permissions.rules`:
* restricts `*.k8s.io` as an APIGroup
* restricts dot-less API groups (`""`, `"*"`, `"batch"`, `"api"`, etc)
* restricts `NonResourceURL` use in permission rules

`main.go` needed to be refactored as the cyclomatic complexity finally hit the roof.  Apologies in advance.

The old behavior of failing a StackInstall unpack if an app.yaml `dependsOn` field did not contain an API Group has been removed.  This will allow the use of `pod/v1`, for example, as a dependency so long as `--restrict-core-apigroups` has not been provided.  While `StackInstall` resources were prevented from using the `""` API group, `Stack` resources were not enforcing the same policy.

Unit tests cover the added logic and stack handler validation.

1. I took the yaml of an already installed Stack (stack-wordpress)
1. removed the status and creation metadata
1. then deleted the resource. 

With a crossplane image built from this branch and with the restricted command line arguments:
1.  I applied the stack-wordpress.yaml unmodified
1.  and waited for that Stack to be successful,
1.  and then repeated the process with bad values
1.  and waited for the Stack to fail reconciliation with the expected error

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/cmd/crossplane/main.go
+++ b/cmd/crossplane/main.go
@@ -19,7 +19,9 @@ package main
 import (
 	"os"
 	"path/filepath"
+	"time"
 
+	"github.com/go-logr/logr"
 	"github.com/spf13/afero"
 	"gopkg.in/alecthomas/kingpin.v2"
 	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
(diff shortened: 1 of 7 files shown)


cmd/crossplane/main.go
```diff
@@ -38,106 +40,166 @@ import (
...
 	if *debug {
 		// The controller-runtime runs with a no-op logger by default. It is
 		// *very* verbose even at info level, so we only provide it a real
 		// logger when we're running in debug mode.
-		ctrl.SetLogger(zl)
+		ctrl.SetLogger(a.zl)
+	}
+
+	knownCmd := false
+	for _, cmd := range commands {
```
suskin: How did we decide to take this approach, rather than have a `case` statement? On one hand I like that it makes it easy to add new commands and handle them in the same way, but on the other hand, I'm not sure about having the `break`.

cmd/crossplane/main.go
```diff
@@ -38,106 +40,166 @@ import (
...
 
 		kingpin.FatalIfError(apis.AddToScheme(mgr.GetScheme()), "Cannot add core Crossplane APIs to scheme")
 		kingpin.FatalIfError(oam.Setup(mgr, log), "Cannot setup OAM controllers")
 		kingpin.FatalIfError(workload.Setup(mgr, log), "Cannot setup workload controllers")
 		kingpin.FatalIfError(mgr.Start(ctrl.SetupSignalHandler()), "Cannot start controller manager")
+	}}}
+}
+
+// stackCmd provides the fullCommand for the "stack" command
+func (a appReq) stackCmd() []fullCommand {
```
suskin: Do we need this method? What's the advantage of bundling the stack commands together in a method in this way? As a potential future maintainer, I would personally prefer that they weren't bundled, since there are only two, and because this adds another level of indirection to following the full set of commands around to see how they're generated and bound.

cmd/crossplane/main.go
displague: Because the commands are a slice I can't offer a switch over an unknown quantity of command cases.

The `for` loop was envisioned while breaking up the cyclomatically complex main.  Now that it has been broken up I can see how this could be turned back into a switch statement.  That said, we have three commands to process now.  If this switch statement doesn't trigger the complexity warnings the next command (or the next one) could do it and we would be revisiting this again.

cmd/crossplane/main.go
displague: We've been adding lots of changes to `stackManageCmd` lately and I don't think that is going to stop. I kind of like that we don't have to get involved with the other commands while editing the stack manager.

The variables are kept in their function scope so they could be renamed to simplify them because there is no longer ambiguity. (remove `stackManager` prefix on variables in the `stackManageCmd`).

displague: My attempt to appeal to a higher authority doesn't seem to put much weight on either argument:

https://github.com/golang/go/wiki/CodeReviewComments#line-length

> This is, actually, exactly the same advice about how long a function should be. There's no rule "never have a function more than N lines long", but there is definitely such a thing as too long of a function, and of too stuttery tiny functions, and the solution is to change where the function boundaries are, not to start counting lines.

I was looking for something along the lines of "keep a function short and focused on one thing".

pkg/stacks/unpack.go
```diff
@@ -498,16 +498,15 @@ func (sp *StackPackage) applyRules() error {
 			}
 
 			gk := schema.ParseGroupKind(crd)
-			if gk.Group == "" || gk.Kind == "" {
```
jbw976: help me understand this change, why do we no longer check for `gk.Group == ""` here?

displague: We are no longer preventing the use of the corev1 apigroup during stackinstall's unpack. We defer those decisions to the stack controller.

Unpack could he aware of the restriction mode but we would have to snake that setting through the stackinstalls controller and the unpack command. Would we repeat (pre-peat?) all of the restriction checking at unpack time?

pkg/controller/stacks/stack/stack.go
```diff
@@ -488,6 +492,81 @@ func (h *stackHandler) createPersonaClusterRolesCRDHandler() crdHandler {
...
+		if a[i] != b[i] {
+			return false
+		}
+	}
+	return true
+}
+
+// policiesEqual verifies that two RBAC policy rules match
+//
+// TODO(displague) Any difference in the order of the policies array fields
```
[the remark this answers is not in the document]
displague: That makes sense. I could permit subset matches too if I change this from `policyEqual` to `policyIsSubset`. 

This would allow a Stack resource to state, 'I don't need secrets but do need configmaps and events'.

Any thoughts on taking that direction now vs later?

pkg/stacks/unpack.go
jbw976: ha oh right, duh :)

so you were right, we did have a tiny bit of core api group checking code already in the code base! i think this is the right move to consolidate the enforcement in the stack controller, before generating any rbac policies on behalf of the stack.  this is good.

pkg/controller/stacks/stack/stack.go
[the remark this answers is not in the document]
jbw976: I think we can proceed with what you have implemented and see if we need to expand out support/functionality in response to real use cases.

Issue #1337, opened 2020-03-12
Resource names and labels produced by Crossplane may exceed name lengths


displague:
# What happened?

Using Stack Manager in a host-aware environment, various resource names and labels exceeded the max length supported by the Kubernetes API.

### How can we reproduce it?

In a host-aware environment, create a host config whose name along with the namespace and name of a controller exceed 253 characters.

The same long names will effect labels created elsewhere.

### What environment did it happen in?
Crossplane version:

displague:
Closed by #1338

PR #1338, opened 2020-03-12
Stacks: Truncate resource names and labels for host-aware environments

displague:

Addresses #1337

There may be other scenarios requiring truncation, but this PR addresses the known problem encountered when creating a tenant `StackInstall` resource with a long name in a long namespace that will be processed by a host.

Resource names in the host environment may not exceed 253 characters. Since these names are based on tenant namespace+name they could easily exceed this limitation.
    
There are also some resources created on the host which invoke admission controller behaviors adding resource labels that match resource names. Because labels values are limited to 63 characters the resource names created on the host should not exceed 63 characters.
    
Because the names have been truncated, annotations have been added on the host to help identify the related tenant resource when debugging.
    
This PR adds a truncation package with helpers for specific use-cases. Truncation is performed in a predictable way to make it possible to find resources by label knowing only the original resource names.  The suffix of the truncation is a subset of a lowercased base32 checksum of a SHA1 digest of the original string.  SHA1's security limitations are not a factor here because this is not being used for security purposes, it is chosen for quick and even distribution in the checksum.
    
The suffix is lowercase because resource names must be lowercase. Base32 was chosen to avoid "/" and a conflicting set of alphabetic characters (upper and lower case are used in base64).

```sh
cluster/local/hosted_test.sh
export KUBECONFIG=~/.kube/kind-config-kind

# 63 characters
ns="r23456789012345678901234567890123456789012345678901234567890123"

# 63.63.63 characters
resource="s23456789012345678901234567890123456789012345678901234567890123.t23456789012345678901234567890123456789012345678901234567890123.u23456789012345678901234567890123456789012345678901234567890123.u23456789012345678901234567890123456789012345678901234567890"

kubectl create namespace $ns
kubectl crossplane stack install --namespace $ns --cluster 'crossplane/stack-gcp:master' $resource

kind delete cluster --name local-hosted-test
```

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/docs/troubleshoot.md
+++ b/docs/troubleshoot.md
@@ -11,6 +11,7 @@ indent: true
 * [Crossplane Logs](#crossplane-logs)
 * [Pausing Crossplane](#pausing-crossplane)
 * [Deleting a Resource Hangs](#deleting-a-resource-hangs)
+* [Host-Aware Resource Debugging](#host-aware-resource-debugging)
 
 ## Using the trace command
 
@@ -131,3 +132,16 @@ For example, for a Workload object (`workloads.compute.crossplane.io`) named
 ```console
 kubectl patch workloads.compute.crossplane.io test-workload -p '{"metadata":{"finalizers": []}}' --type=merge
 ```
+
+## Host-Aware Resource Debugging
+
+Stack resources (including the Stack, service accounts, deployments, and jobs) are usually easy to identify by name. These resource names are based on the name used in the StackInstall or Stack resource.
+
+In some cases, the full name of a Stack resource, which could be up to 253 characters long, can not be represented in the created resources. For example, jobs and deployment names may not exceed 63 characters because these names are turned into resource label values which impose a 63 character limit. Stack created resources whose names would otherwise not be permitted in the Kubernetes API will be truncated with a unique suffix.
+
+When running the Stack Manager in host-aware mode, tenant stack resources created in the host controller namespace generally reuse the Stack names: "{tenant namespace}.{tenant name}".  In order to avoid the name length restrictions, these resources may be truncated at either or both of the namespace segment or over the entire name.  In these cases resource labels, owner references, and annotations should be consulted to identify the responsible Stack.
+
+* [Relationship Labels](https://github.com/crossplane/crossplane/blob/master/design/one-pager-stack-relationship-labels.md)
+* [Owner References](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#owners-and-dependents)
+* Annotations: `tenant.crossplane.io/{singular}-name` and `tenant.crossplane.io/{singular}-namespace` (_singular_ may be `stackinstall`, `clusterstackinstall` or `stack`)
+
--- a/docs/troubleshoot.md
+++ b/docs/troubleshoot.md
@@ -135,13 +135,78 @@ kubectl patch workloads.compute.crossplane.io test-workload -p '{"metadata":{"fi
 
 ## Host-Aware Resource Debugging
 
-Stack resources (including the Stack, service accounts, deployments, and jobs) are usually easy to identify by name. These resource names are based on the name used in the StackInstall or Stack resource.
+Stack resources (including the Stack, service accounts, deployments, and jobs)
+are usually easy to identify by name. These resource names are based on the name
+used in the StackInstall or Stack resource.
 
-In some cases, the full name of a Stack resource, which could be up to 253 characters long, can not be represented in the created resources. For example, jobs and deployment names may not exceed 63 characters because these names are turned into resource label values which impose a 63 character limit. Stack created resources whose names would otherwise not be permitted in the Kubernetes API will be truncated with a unique suffix.
+In some cases, the full name of a Stack resource, which could be up to 253
+characters long, can not be represented in the created resources. For example,
+jobs and deployment names may not exceed 63 characters because these names are
+turned into resource label values which impose a 63 character limit. Stack
+created resources whose names would otherwise not be permitted in the Kubernetes
+API will be truncated with a unique suffix.
 
-When running the Stack Manager in host-aware mode, tenant stack resources created in the host controller namespace generally reuse the Stack names: "{tenant namespace}.{tenant name}".  In order to avoid the name length restrictions, these resources may be truncated at either or both of the namespace segment or over the entire name.  In these cases resource labels, owner references, and annotations should be consulted to identify the responsible Stack.
+When running the Stack Manager in host-aware mode, tenant stack resources
+created in the host controller namespace generally reuse the Stack names:
+"{tenant namespace}.{tenant name}".  In order to avoid the name length
+restrictions, these resources may be truncated at either or both of the
+namespace segment or over the entire name.  In these cases resource labels,
+owner references, and annotations should be consulted to identify the
+responsible Stack.
 
 * [Relationship Labels](https://github.com/crossplane/crossplane/blob/master/design/one-pager-stack-relationship-labels.md)
 * [Owner References](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#owners-and-dependents)
-* Annotations: `tenant.crossplane.io/{singular}-name` and `tenant.crossplane.io/{singular}-namespace` (_singular_ may be `stackinstall`, `clusterstackinstall` or `stack`)
+* Annotations: `tenant.crossplane.io/{singular}-name` and
+  `tenant.crossplane.io/{singular}-namespace` (_singular_ may be `stackinstall`,
+  `clusterstackinstall` or `stack`)
 
+### Example
+
+Long resource names may be present on the tenant.
+
+```console
+$ name=stack-with-a-really-long-resource-name-so-long-that-it-will-be-truncated
+$ ns=this-is-just-a-really-long-namespace-name-at-the-character-max
+$ kubectl create ns $ns
+$ kubectl crossplane stack install --namespace $ns crossplane/sample-stack-wordpress:0.1.1 $name
+```
+
+When used as host resource names, the stack namespace and stack are concatenated
+ to form host names, as well as labels.  These resource names and label values
+ must be truncated to fit the 63 character limit on label values.
+
+```console
+$ kubectl --context=crossplane-host -n tenant-controllers get job -o yaml
+apiVersion: v1
+items:
+- apiVersion: batch/v1
+  kind: Job
+  metadata:
+    annotations:
+      tenant.crossplane.io/stackinstall-name: stack-with-a-really-long-resource-name-so-long-that-it-will-be-truncated
+      tenant.crossplane.io/stackinstall-namespace: this-is-just-a-really-long-namespace-name-at-the-character-max
+    creationTimestamp: "2020-03-20T17:06:25Z"
+    labels:
+      core.crossplane.io/parent-group: stacks.crossplane.io
+      core.crossplane.io/parent-kind: StackInstall
+      core.crossplane.io/parent-name: stack-with-a-really-long-resource-name-so-long-that-it-wi-alqdw
+      core.crossplane.io/parent-namespace: this-is-just-a-really-long-namespace-name-at-the-character-max
+      core.crossplane.io/parent-uid: 596705e4-a28e-47c9-a907-d2732f07a85e
+      core.crossplane.io/parent-version: v1alpha1
+    name: this-is-just-a-really-long-namespace-name-at-the-characte-egoav
+    namespace: tenant-controllers
+  spec:
+    backoffLimit: 0
+    completions: 1
+    parallelism: 1
+    selector:
+      matchLabels:
+        controller-uid: 8f290bf2-8c91-494a-a76b-27c2ccb9e0a8
+    template:
+      metadata:
+        creationTimestamp: null
+        labels:
+          controller-uid: 8f290bf2-8c91-494a-a76b-27c2ccb9e0a8
+          job-name: this-is-just-a-really-long-namespace-name-at-the-characte-egoav
+  ...
+```
--- a/pkg/controller/stacks/hosted/config.go
+++ b/pkg/controller/stacks/hosted/config.go
@@ -22,9 +22,20 @@ import (
 
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
+
+	"github.com/crossplane/crossplane/pkg/stacks/truncate"
 )
 
 const (
+	// AnnotationTenantNameFmt with a resource singular applied provides the
+	// annotation key used to identify tenant resources by name on the host side
+	AnnotationTenantNameFmt = "tenant.crossplane.io/%s-name"
+
+	// AnnotationTenantNamespaceFmt with a resource singular applied provides
+	// the annotation key used to identify tenant resources by namespace on the
+	// host side
+	AnnotationTenantNamespaceFmt = "tenant.crossplane.io/%s-namespace"
+
 	errMissingOption = "host aware mode activated but %s is not set"
 )
 
@@ -59,11 +70,27 @@ func NewConfig(hostControllerNamespace, tenantAPIServiceHost, tenantAPIServicePo
 	}, nil
 }
 
-// ObjectReferenceOnHost maps object with given name and namespace into single controller namespace on Host Cluster.
+// ObjectReferenceOnHost maps object with given name and namespace into single
+// controller namespace on Host Cluster.
+// The resource name on the host cluster is truncated to label value length
+// because the name may be used in labels defined by an admission controller, as
+// is the case for jobs and deployments.
 func (c *Config) ObjectReferenceOnHost(name, namespace string) corev1.ObjectReference {
 	return corev1.ObjectReference{
-		Name:      fmt.Sprintf("%s.%s", namespace, name),
+		Name:      truncate.LabelValue(fmt.Sprintf("%s.%s", namespace, name)),
 		Namespace: c.HostControllerNamespace,
+	}
+}
+
+// ObjectReferenceAnnotationsOnHost returns a map for use as annotations on the
+// host to identify the named tenant resource
+func ObjectReferenceAnnotationsOnHost(singular, name, namespace string) map[string]string {
+	nameLabel := fmt.Sprintf(AnnotationTenantNameFmt, singular)
+	nsLabel := fmt.Sprintf(AnnotationTenantNamespaceFmt, singular)
+
+	return map[string]string{
+		nameLabel: name,
+		nsLabel:   namespace,
 	}
 }
 
--- a/pkg/controller/stacks/hosted/config.go
+++ b/pkg/controller/stacks/hosted/config.go
@@ -27,13 +27,15 @@ import (
 )
 
 const (
-	// AnnotationTenantNameFmt with a resource singular applied provides the
+	// AnnotationTenantNameFmt with a CR `singular` name applied provides the
 	// annotation key used to identify tenant resources by name on the host side
+	// Example: tenant.crossplane.io/stackinstall-name
 	AnnotationTenantNameFmt = "tenant.crossplane.io/%s-name"
 
-	// AnnotationTenantNamespaceFmt with a resource singular applied provides
+	// AnnotationTenantNamespaceFmt with a CR `singular` name applied provides
 	// the annotation key used to identify tenant resources by namespace on the
 	// host side
+	// Example: tenant.crossplane.io/stack-namespace
 	AnnotationTenantNamespaceFmt = "tenant.crossplane.io/%s-namespace"
 
 	errMissingOption = "host aware mode activated but %s is not set"
@@ -70,11 +72,13 @@ func NewConfig(hostControllerNamespace, tenantAPIServiceHost, tenantAPIServicePo
 	}, nil
 }
 
-// ObjectReferenceOnHost maps object with given name and namespace into single
-// controller namespace on Host Cluster.
-// The resource name on the host cluster is truncated to label value length
-// because the name may be used in labels defined by an admission controller, as
-// is the case for jobs and deployments.
+// ObjectReferenceOnHost maps objects with a given name and namespace into a
+// single controller namespace on the Host Cluster.
+//
+// The resource name on the host cluster may be truncated from the original
+// tenant name to fit label value length.  The resource name may be used as a
+// label, as is the case for jobs and deployments where the admission controller
+// generates labels based on the resource name.
 func (c *Config) ObjectReferenceOnHost(name, namespace string) corev1.ObjectReference {
 	return corev1.ObjectReference{
 		Name:      truncate.LabelValue(fmt.Sprintf("%s.%s", namespace, name)),
@@ -83,7 +87,13 @@ func (c *Config) ObjectReferenceOnHost(name, namespace string) corev1.ObjectRefe
 }
 
 // ObjectReferenceAnnotationsOnHost returns a map for use as annotations on the
-// host to identify the named tenant resource
+// host to identify the named tenant resource. This annotation is used for
+// reference purposes to define a relationship to a single resource of a
+// specific kind. For example, this could be used to declare the tenant
+// stackinstall resource that is related to a host install job.
+//
+// On a host the original tenant resource name may be truncated away.
+// Annotations provide a way to store the original name without truncation.
 func ObjectReferenceAnnotationsOnHost(singular, name, namespace string) map[string]string {
 	nameLabel := fmt.Sprintf(AnnotationTenantNameFmt, singular)
 	nsLabel := fmt.Sprintf(AnnotationTenantNamespaceFmt, singular)
--- a/pkg/controller/stacks/install/installjob.go
+++ b/pkg/controller/stacks/install/installjob.go
@@ -70,15 +70,17 @@ type prepareInstallJobParams struct {
 	stackManagerPullPolicy corev1.PullPolicy
 	imagePullPolicy        corev1.PullPolicy
 	labels                 map[string]string
+	annotations            map[string]string
 	imagePullSecrets       []corev1.LocalObjectReference
 }
 
 func prepareInstallJob(p prepareInstallJobParams) *batchv1.Job {
 	return &batchv1.Job{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      p.name,
-			Namespace: p.namespace,
-			Labels:    p.labels,
+			Name:        p.name,
+			Namespace:   p.namespace,
+			Labels:      p.labels,
+			Annotations: p.annotations,
 		},
 		Spec: batchv1.JobSpec{
 			BackoffLimit: &jobBackoff,
--- a/pkg/controller/stacks/install/stackinstall.go
+++ b/pkg/controller/stacks/install/stackinstall.go
@@ -346,7 +346,11 @@ func (h *stackInstallHandler) createInstallJob() *batchv1.Job {
 	namespace := i.GetNamespace()
 
 	if hCfg != nil {
-		annotations = hosted.ObjectReferenceAnnotationsOnHost("stackinstall", name, namespace)
+		singular := "stackinstall"
+		if i.PermissionScope() == string(apiextensionsv1beta1.ClusterScoped) {
+			singular = "clusterstackinstall"
+		}
+		annotations = hosted.ObjectReferenceAnnotationsOnHost(singular, name, namespace)
 
 		// In Hosted Mode, we need to map all install jobs on tenant Kubernetes into a single namespace on host cluster.
 		o := hCfg.ObjectReferenceOnHost(name, namespace)
--- a/pkg/controller/stacks/install/stackinstall.go
+++ b/pkg/controller/stacks/install/stackinstall.go
@@ -336,6 +336,8 @@ func (h *stackInstallHandler) create(ctx context.Context) (reconcile.Result, err
 }
 
 func (h *stackInstallHandler) createInstallJob() *batchv1.Job {
+	var annotations map[string]string
+
 	i := h.ext
 	executorInfo := h.executorInfo
 	hCfg := h.hostAwareConfig
@@ -344,6 +346,8 @@ func (h *stackInstallHandler) createInstallJob() *batchv1.Job {
 	namespace := i.GetNamespace()
 
 	if hCfg != nil {
+		annotations = hosted.ObjectReferenceAnnotationsOnHost("stackinstall", name, namespace)
+
 		// In Hosted Mode, we need to map all install jobs on tenant Kubernetes into a single namespace on host cluster.
 		o := hCfg.ObjectReferenceOnHost(name, namespace)
 		name = o.Name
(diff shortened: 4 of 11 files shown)


pkg/controller/stacks/hosted/config.go
```diff
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"net/url"
 
+	"github.com/crossplane/crossplane/pkg/stacks/truncate"
```
golangcibot: File is not `goimports`-ed with -local github.com/crossplane/crossplane (from `goimports`)
```suggestion

```

pkg/stacks/relationship.go
```diff
@@ -17,8 +17,10 @@ limitations under the License.
 package stacks
 
 import (
+	"fmt"
 	"strings"
 
+	"github.com/crossplane/crossplane/pkg/stacks/truncate"
```
golangcibot: File is not `goimports`-ed with -local github.com/crossplane/crossplane (from `goimports`)
```suggestion

```

pkg/stacks/truncate/truncate.go
```diff
@@ -0,0 +1,97 @@
...
+import (
+	// nolint:blacklist
+	"crypto/sha1"
+
+	"encoding/base32"
+	"fmt"
+)
+
+const (
+	LabelNameLength    = 63
```
golangcibot: exported const LabelNameLength should have comment (or a comment on this block) or be unexported (from `golint`)

pkg/controller/stacks/stack/stack.go
```diff
@@ -722,6 +726,9 @@ func (h *stackHandler) prepareHostAwareJob(j *batch.Job, tokenSecret string) err
 	j.Name = o.Name
 	j.Namespace = o.Namespace
 
+	a := hosted.ObjectReferenceAnnotationsOnHost("stack", h.ext.GetName(), h.ext.GetNamespace())
```
jbw976: i think the function this is inside of, `prepareHostAwareJob` is a missed remnant of the old job support we used to have and we recently removed.  I think this entire function can be removed.  It's not called from anywhere in the codebase except for tests.

pkg/controller/stacks/install/installjob.go
```diff
@@ -70,15 +70,17 @@ type prepareInstallJobParams struct {
...
 	imagePullSecrets       []corev1.LocalObjectReference
 }
 
 func prepareInstallJob(p prepareInstallJobParams) *batchv1.Job {
 	return &batchv1.Job{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      p.name,
-			Namespace: p.namespace,
-			Labels:    p.labels,
+			Name:        p.name,
```
jbw976: the name that will be used on this job also needs to be truncated to label (<=63) length too I think, if I'm reading this error message correctly:

```
message: 'Job.batch "crossplane-system.crossplane-stack-minimal-gcpv0.2.2-14-g96f534a"
        is invalid: spec.template.labels: Invalid value: "crossplane-system.crossplane-stack-minimal-gcpv0.2.2-14-g96f534a":
        must be no more than 63 characters'
```
I'm interpreting that to mean that the name given to the job, `crossplane-system.crossplane-stack-minimal-gcpv0.2.2-14-g96f534a` is being automatically applied as a label in the job's `spec.template.labels`, so it needs to be truncated.

Here's a snippet of the job and its relevant `spec.template.labels` that seems to confirm that the job's name gets inserted as a `job-name` label inside it's `spec.template.labels`:
```
spec:
  template:
    metadata:
      creationTimestamp: null
      labels:
        controller-uid: 643d1d2f-d6c9-4738-bd86-e8a478268146
        job-name: crossplane-system.crossplane-provider-gcpv0.6.0-rc.29.g762643b
```

The full job yaml output for your convenience can be seen in this gist: https://gist.github.com/jbw976/2c9def142b4e24fd1e39b9f3f439856e

pkg/stacks/truncate/truncate.go
```diff
@@ -0,0 +1,105 @@
...
+limitations under the License.
+*/
+
+// Package truncate provides functions for truncating Kubernetes values in a
+// predictable way offering mildly collision safe values usable in
+// deterministic field searches.
+package truncate
+
+import (
+	"crypto/sha1" // nolint:blacklist
```
jbw976: add a comment for why we're skipping the blacklist lint check here?

pkg/stacks/truncate/truncate_test.go
```diff
@@ -0,0 +1,200 @@
...
+		{
+			name: "max",
+			args: args{str: str63},
+			want: str63,
+		},
+		{
+			name: "truncate",
+			args: args{str: str253},
+			want: strings.Repeat("1234567890", 5) + "1234567-ij43w",
+		},
```
jbw976: nit (not blocking): it could be nice to add another sanity (and easy to read) test case for one of the real life scenarios we saw that we are hoping to fix with this new truncation behavior.  These cases test the major scenarios we care about, but they may be a bit hard to read since the last case here has just a little bit of "magic" in them.

pkg/stacks/truncate/truncate_test.go
```diff
@@ -0,0 +1,200 @@
...
+	"strings"
+	"testing"
+)
+
+var (
+	str63  = strings.Repeat("1234567890", 6) + "123"
+	str253 = strings.Repeat("1234567890", 25) + "123"
+)
+
+func Test_truncate(t *testing.T) {
```
jbw976: great coverage for a lot of possibilities these truncation functions will need to deal with :)

pkg/controller/stacks/install/stackinstall.go
```diff
@@ -344,6 +346,8 @@ func (h *stackInstallHandler) createInstallJob() *batchv1.Job {
 	namespace := i.GetNamespace()
 
 	if hCfg != nil {
+		annotations = hosted.ObjectReferenceAnnotationsOnHost("stackinstall", name, namespace)
```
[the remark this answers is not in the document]
displague: This was intended as a debugging annotation only.

In the world of edge cases, you got me.

Both a stackinstall and a clusterstackinstall could reside in the same namespace with the same name. I'll have to track "clusterstackinstall" with a different label name.

pkg/controller/stacks/stack/stack.go
displague: This brings me great joy.

pkg/controller/stacks/install/installjob.go
displague: `prepareInstallJob` is called with the name already truncated:
https://github.com/crossplane/crossplane/pull/1338/files#diff-a2cd537f6531321fbef9a92d96b2c425R351-R367

The truncation happens within `ObjectReferenceOnHost`:

https://github.com/crossplane/crossplane/pull/1338/files#diff-33a4f973d281ef627a1c398d9941fc68R73-R84

pkg/stacks/truncate/truncate.go
displague: Will do. It is also worth pointing out that I think our golangci-lint config may not include the blacklist linter, so this `nolint` is for other environments where this linter is included.

I've noticed warnings from `make` that this linter is not known.

pkg/stacks/truncate/truncate_test.go
displague: I'll add a test case we can refer to as an example.

The testing example in the PR description is similarly obtuse.  The digits make it easy to visually count the character length but don't reasonably convey how this affects "my-really-really-long-namespace.stack-name-is-also-long-version-one-point-oh"

pkg/stacks/truncate/truncate.go
displague: `WARN [runner/nolint] Found unknown linters in //nolint directives: blacklist `

displague:
@jbw976 I believe I've addressed all of your feedback. The added troubleshooting tips are here: https://github.com/crossplane/crossplane/pull/1338/commits/f77f0d5411f324043bea2ef4a017002009a8b0b4

What I added is somewhat generic.  I think we can add more detail to that as cases come up and we find useful patterns.

pkg/stacks/truncate/truncate.go
jbw976: not sure if i fully follow this, if our `make` shows a warning for this linter, and our golangci-lint config doesn't use this linter, should we just remove it? that makes more sense to me, instead of living with a new warning in our default environment.

pkg/controller/stacks/install/installjob.go
jbw976: Ahh, now I see where/how that happens, thanks for pointing that out!

docs/troubleshoot.md
```diff
@@ -131,3 +132,16 @@ For example, for a Workload object (`workloads.compute.crossplane.io`) named
...
+
+When running the Stack Manager in host-aware mode, tenant stack resources created in the host controller namespace generally reuse the Stack names: "{tenant namespace}.{tenant name}".  In order to avoid the name length restrictions, these resources may be truncated at either or both of the namespace segment or over the entire name.  In these cases resource labels, owner references, and annotations should be consulted to identify the responsible Stack.
+
+* [Relationship Labels](https://github.com/crossplane/crossplane/blob/master/design/one-pager-stack-relationship-labels.md)
+* [Owner References](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#owners-and-dependents)
+* Annotations: `tenant.crossplane.io/{singular}-name` and `tenant.crossplane.io/{singular}-namespace` (_singular_ may be `stackinstall`, `clusterstackinstall` or `stack`)
```
jbw976: Could we add just a quick (real) concrete example here so people will have more of an understanding of what this is talking about?

docs/troubleshoot.md
```diff
@@ -131,3 +132,16 @@ For example, for a Workload object (`workloads.compute.crossplane.io`) named
 ```console
 kubectl patch workloads.compute.crossplane.io test-workload -p '{"metadata":{"finalizers": []}}' --type=merge
 ```
+
+## Host-Aware Resource Debugging
+
+Stack resources (including the Stack, service accounts, deployments, and jobs) are usually easy to identify by name. These resource names are based on the name used in the StackInstall or Stack resource.
+
+In some cases, the full name of a Stack resource, which could be up to 253 characters long, can not be represented in the created resources. For example, jobs and deployment names may not exceed 63 characters because these names are turned into resource label values which impose a 63 character limit. Stack created resources whose names would otherwise not be permitted in the Kubernetes API will be truncated with a unique suffix.
```
jbw976: want to wrap these lines? the rest of the doc appears to be wrapped.

pkg/controller/stacks/stack/stack.go
```diff
@@ -707,20 +708,8 @@ func (h *stackHandler) prepareHostAwareDeployment(d *apps.Deployment, tokenSecre
 	d.Name = o.Name
 	d.Namespace = o.Namespace
 
-	return nil
-}
-func (h *stackHandler) prepareHostAwareJob(j *batch.Job, tokenSecret string) error {
```
suskin: Was this just not being used anywhere? The only other references I'm seeing are in tests.

pkg/controller/stacks/stack/stack.go
```diff
@@ -734,7 +723,9 @@ func (h *stackHandler) prepareDeployment(d *apps.Deployment) {
 	controllerDeployment.Spec.DeepCopyInto(&d.Spec)
 
 	// force the deployment to use stack opinionated names and service account
-	name := h.ext.Name + "-controller"
+	suffix := "-controller"
+	name, _ := truncate.Truncate(h.ext.Name, truncate.LabelValueLength-len(suffix), truncate.DefaultSuffixLength)
```
suskin: Nitpick: I would normally have expected `truncate.LabelValueLength-len(suffix)` to be more like `truncate.LabelValueLength - len(suffix)` so the operator is easier to see, but maybe this is a go idiom I'm not familiar with yet.

pkg/controller/stacks/stack/stack_test.go
```diff
@@ -1339,6 +1356,107 @@ func TestProcessDeployment(t *testing.T) {
...
+				withControllerSpec(defaultControllerSpec())),
+			initObjs: []runtime.Object{sa(
+				withSANamespacedName(types.NamespacedName{
+					Name:      resourceName + dashAlphabet,
+					Namespace: namespace + dashAlphabet,
+				}),
+				withTokenSecret(corev1.ObjectReference{
+					Name:      resourceName + dashAlphabet,
+					Namespace: namespace + dashAlphabet,
+				})), saSecret(resourceName+dashAlphabet, namespace+dashAlphabet)},
```
suskin: Nitpick: noticing the pattern again of having no spaces around the operator

pkg/stacks/relationship.go
```diff
@@ -33,9 +36,19 @@ const (
...
 	LabelParentUID       = "core.crossplane.io/parent-uid"
 
-	LabelMultiParentPrefix   = "parent.stacks.crossplane.io/"
+	LabelMultiParentPrefix = "parent.stacks.crossplane.io/"
+
 	LabelMultiParentNSFormat = "parent.stacks.crossplane.io/%s"
-	LabelMultiParentFormat   = LabelMultiParentNSFormat + "-%s"
+
+	// LabelMultiParentFormat defines the format for combining a
+	// LabelMultiParentNSFormat with a named resource
```
suskin: Which input is mapped to which placeholder?

pkg/stacks/relationship.go
```diff
@@ -33,9 +36,19 @@ const (
...
+	LabelMultiParentFormat = "%s-%s"
+
+	// preserveNSLength is the number of characters using the label name that
+	// will be dedicated to identifying the namespace. This length will include
+	// truncation characters if the namespace exceeds this length.
+	// example: parent.stacks.crossplane.io/{up to 32 chars of NS}-{Name}
```
suskin: Excellent detailed comment with example. This helps me reason about the intended behavior of the implementation. :squid:

pkg/stacks/relationship.go
```diff
@@ -48,16 +61,54 @@ type KindlyIdentifier interface {
 	GroupVersionKind() schema.GroupVersionKind
 }
 
+// MultiParentLabelPrefix returns the NS specific prefix of a multi-parent label
+// for resources co-owned by a set of Stacks.
```
suskin: Originally when I read this, I interpreted it as saying that it would parse a label and return its namespace prefix. Maybe an example of an input and output would be helpful in clearly explaining the reason this method exists.

pkg/stacks/relationship.go
```diff
@@ -48,16 +61,54 @@ type KindlyIdentifier interface {
...
+//
+// This prefix is suitable for identifying resources labeled within a given
+// namespace. The prefix may include a predictable truncation suffix if the
+// namespace exceeds 32 characters. This truncation length permits another 32
+// characters for a (potentially truncated) resource name to be appended to the
+// label.
+func MultiParentLabelPrefix(stackParent metav1.Object) string {
+	ns := stackParent.GetNamespace()
+
+	// guaranteed not to error based on the lengths of values we are supplying
```
suskin: What if a future maintainer changes the lengths?

pkg/stacks/relationship.go
```diff
@@ -48,16 +61,54 @@ type KindlyIdentifier interface {
...
+// MultiParentLabel returns a label name identifying the namespaced name of the
+// stack resource that co-owns another resource
+//
+// The label returned is based on the MultiParentLabelPrefix, which may include
+// a truncation suffix, and is then potentially truncated again to fit in the
+// complete label length restrictions.
+func MultiParentLabel(stackParent metav1.Object) string {
+	prefix := MultiParentLabelPrefix(stackParent)
+
+	// guaranteed 2 parts based on LabelMultiParentNSFormat
```
suskin: Are we using that format here? Since we accept input after it's been formatted, and since this is a public method, I'm not sure we can assume that in the long run. But it may be okay to make the assumption today and revisit it later.

pkg/stacks/relationship_test.go
```diff
@@ -0,0 +1,310 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+var (
+	namespace             = "cool-namespace"
+	uidString             = "definitely-a-uuid"
+	resourceName          = "cool-resource"
+	thirtyTwoAs           = strings.Repeat("a", 32)
```
suskin: Nitpick: originally I read it as the word `as`

pkg/stacks/relationship_test.go
```diff
@@ -0,0 +1,310 @@
...
+				LabelParentName:      sixtyThreeAs,
+				LabelParentUID:       uidString,
+				LabelParentGroup:     "",
+				LabelParentKind:      "",
+				LabelParentVersion:   "",
+			},
+		},
+		{
+			name: "Truncated",
+			i:    resource(sixtyThreeAs+"z", sixtyThreeAs+"z", uidString),
```
suskin: I  see - because the limit is `63`

pkg/stacks/relationship_test.go
```diff
@@ -0,0 +1,310 @@
...
+		},
+		{
+			name:        "NotTruncatedNS",
+			stackParent: resource(thirtyTwoAs, "z", uidString),
+			want:        "parent.stacks.crossplane.io/" + thirtyTwoAs + "-z",
+		},
+		{
+			name:        "TruncatedNS",
+			stackParent: resource(sixtyThreeAs+"z", resourceName, uidString),
+			// notice ukl4i is also the suffix with 64 a's because the sum is
```
suskin: I didn't follow that - we can get the same suffix for different strings?

pkg/stacks/truncate/truncate.go
```diff
@@ -0,0 +1,106 @@
...
+	sha1Length = 40
+)
+
+// Truncate replaces the suffixLength number of trailing characters from str
+// with a consistent hash fragment based on that string. The suffix will include
+// a leading hyphen.
+// An error will be returned if the truncation length is less than the
+// suffixLength, or truncation length is greater than sha1Length, or the suffix
+// length is less than 2.
+func Truncate(str string, length, suffixLength int) (string, error) {
```
suskin: I have a really dumb question. Did we already consider using or copying Kubernetes's implementation of this same behavior (that they use for pods, for example)? It seems like maybe the implementation is not long enough to be worth it.

pkg/stacks/truncate/truncate.go
```diff
@@ -0,0 +1,106 @@
...
+		return "", fmt.Errorf("truncate suffixLength is less than minimum: %d < 2", suffixLength)
+	}
+
+	// See the import comment regarding use of sha1
+	// nolint:gosec
+	checksum := sha1.Sum([]byte(str))
+	retained := length - suffixLength
+
+	// base64 includes "/" which can only occur once in label names
+	b32 := base32.StdEncoding.WithPadding(base32.NoPadding).EncodeToString(checksum[:])
```
suskin: There are multiple variants to base64, including a [URL-safe variant](https://tools.ietf.org/html/rfc4648#section-5) which does not use `/`. It looks like it's [supported in go](https://golang.org/pkg/encoding/base64/#pkg-variables) too. I would recommend using that instead, unless there's an additional reason to use base32, since base64 is more common.

pkg/stacks/truncate/truncate_test.go
```diff
@@ -0,0 +1,211 @@
...
+		},
+		{
+			name:    "SuffixExceedsStr",
+			args:    args{str: "a", length: 0, suffixLength: 2},
+			want:    "",
+			wantErr: true,
+		},
+		{
+			name:    "SuffixExceedsSum",
+			args:    args{str: strings.Repeat("12345678890", 21), length: 200, suffixLength: 100},
```
suskin: This is a dumb question, but did we consider just padding out to the desired suffix length? : )

pkg/stacks/truncate/truncate.go
```diff
@@ -0,0 +1,106 @@
...
+
+	sha1Length = 40
+)
+
+// Truncate replaces the suffixLength number of trailing characters from str
+// with a consistent hash fragment based on that string. The suffix will include
+// a leading hyphen.
+// An error will be returned if the truncation length is less than the
+// suffixLength, or truncation length is greater than sha1Length, or the suffix
+// length is less than 2.
```
suskin: How was this minimum chosen?

pkg/stacks/truncate/truncate.go
displague: Like Kubernetes, we are using labels to identify the related resources (matchLabels in a deployment refer to the labels in the pod spec).

Our resource names are extra long because we are taking two 253 characters strings (tenant namespace and tenant resource name) and appending them into at most a 253 character slot (the host resource name).  Compounding this further, we need to track these potentially 506 character names as label values, which can be no longer than 63 characters. 

Kubernetes does not offer a solution for this. If you try to create a deployment with a name > 63 characters, the admission controller will reject it.  To apply that same resolution here, if the namespace and name of the tenant Stack resource exceeded 62 characters (losing one for a separator), it would need to be rejected.

This PR seeks to fix the current implementation and avoid changing the existing behavior. Since resources with long name concatenations are not currently supported, this PR should not affect existing resources with the new truncation behavior.  A broader change to how resource names are stored on the host could result in orphaned resources in existing host environments.

Annotations offer sufficient length to store the tenant resource names on the host resource without modification.  The tenant resource names are being stored on the host for debug purposes only.

pkg/controller/stacks/hosted/config.go
```diff
@@ -59,14 +70,30 @@ func NewConfig(hostControllerNamespace, tenantAPIServiceHost, tenantAPIServicePo
...
 func (c *Config) ObjectReferenceOnHost(name, namespace string) corev1.ObjectReference {
 	return corev1.ObjectReference{
-		Name:      fmt.Sprintf("%s.%s", namespace, name),
+		Name:      truncate.LabelValue(fmt.Sprintf("%s.%s", namespace, name)),
 		Namespace: c.HostControllerNamespace,
 	}
 }
 
+// ObjectReferenceAnnotationsOnHost returns a map for use as annotations on the
+// host to identify the named tenant resource
```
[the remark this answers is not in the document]
displague: As an addendum to https://github.com/crossplane/crossplane/pull/1338#discussion_r395618270

Annotations offer sufficient length to store the tenant resource names on the host resource without modification.  The tenant resource names are being stored on the host for debug purposes only.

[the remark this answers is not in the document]
displague: I'll weave some of the PR description and comment text into the code comments.  Thanks for pointing out the lack of clarity and telepathy.

pkg/stacks/truncate/truncate.go
displague: Suffix length is an argument to `Truncate`. Each of the `Label*` and `ResourceName` helper functions have a predefined length that I pulled from the ether.  36^5 = 60,466,176.  While this may exceed (or not, I couldn't find a reference for it) the max resources per namespace, that is not the goal here.  The goal is to prevent name collisions on truncated resource names, "{really long namespace name}.stack-ms-print" should not be confused with "{same really long namespace name}.stack-ms-paint" when fetching host resources using a source resource as a reference.

any reference to "sha1" should be replaced with hash, digest, or sum.  I don't want to reflect that implementation choice in the comment.

minimum suffix length of 2 offers "-[a-z0-9]"

pkg/stacks/truncate/truncate.go
displague: While label values can include mixed case, resource names can not contain upper-case characters, so a lower-cased Base32 is `[a-z0-7]` fits the job nicely.  I considered mixing the use based on the job, but often the resource name becomes a label value.

I thought one character pattern would be sufficient here and avoid complexity.

pkg/stacks/truncate/truncate_test.go
displague: I didn't consider that. For our purposes the truncation suffix is locked in.  I added this test and the error response it invokes for logical completeness.

pkg/stacks/truncate/truncate.go
displague: You may have been referring to the Kubernetes suffix pattern that uses random characters as a suffix (deployment `foo`, replica scale 2 resulting in pod `foo-abc` and `foo-zyx`). 

In order to provide host resource names that could be determined given tenant resource names, randomization was avoided.

Could tenant resource labels have been used to to fetch host resources? Perhaps, but existing resources are not labeled.
Are host resource labels (referring to the tenant) sufficient? Perhaps, but existing resources are not labeled.
Are host resource annotations sufficient? They are not indexed, so a query to find a matching deployment on a host, given a tenant Stack name would need to query all stacks (from all namespaces in a tenant cluster).  Again, these annotations are not present on existing resources.

pkg/stacks/relationship_test.go
displague: I'll revisit this comment.  The intent was to express the following:

If the base32 sum of a digest of  "aaaaaaaaaaa" is "foo" (with a suffix length of 4).

Truncating this string to a length of 11 would result in "aaaaaaaaaaa"
Truncating this string to a length of 8 would result in "aaaa-foo"
Truncating this string to a length of 7 would result in "aaa-foo"
Truncating this string to a length of 5 would result in "a-foo"
 
"aaaaaaaaaaa"

pkg/controller/stacks/stack/stack.go
displague: In a resolved comment @jbw976 pointed out that this function was in support of `job` based Stack controllers, which we removed support for in a past release.

pkg/controller/stacks/hosted/config.go
```diff
@@ -59,14 +70,30 @@ func NewConfig(hostControllerNamespace, tenantAPIServiceHost, tenantAPIServicePo
 	}, nil
 }
 
-// ObjectReferenceOnHost maps object with given name and namespace into single controller namespace on Host Cluster.
+// ObjectReferenceOnHost maps object with given name and namespace into single
+// controller namespace on Host Cluster.
+// The resource name on the host cluster is truncated to label value length
```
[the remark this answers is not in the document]
displague: Yes. I'll defer a deeper explanation to the other review comments. In short, a 253 character tenant resource name (505 with namespace) needs to be representable in a 63 character host label value.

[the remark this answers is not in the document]
displague: And that 63 character name should be predictable based on the tenant (source) resource name.

pkg/controller/stacks/hosted/config.go
```diff
@@ -22,9 +22,20 @@ import (
 
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
+
+	"github.com/crossplane/crossplane/pkg/stacks/truncate"
 )
 
 const (
+	// AnnotationTenantNameFmt with a resource singular applied provides the
+	// annotation key used to identify tenant resources by name on the host side
```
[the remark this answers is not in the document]
displague: CRDs have multiple names. Kind (Foo), plural (foos), and singular (foo) can differ. I am referring to the singular name here.

The expected uses are:

* `tenant.crossplane.io/stackinstall-name: foo"
* `tenant.crossplane.io/stack-name: foo"

pkg/controller/stacks/hosted/config.go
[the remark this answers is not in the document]
displague: While adding some debug examples, I learned namespaces can only be 63 characters. ~505~ 315.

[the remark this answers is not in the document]
displague: I discovered some caveats to the name length restrictions while doing some testing in response to @jbw976's request for debugging examples.

Namespaces can not exceed 63 characters. 
Deployments and Jobs, the resources that we intend to create on the host, are limited to 63 characters because their names are used in label values.

Of note for #1354, Secrets (as well as ServiceAccounts) abide by the 253 character limit (with 63 character labels) described in https://kubernetes.io/docs/concepts/overview/working-with-objects/names/

Stack and StackInstall don't seem to abide by this naming policy as I was able to create one with 70+ characters in the name, without dots.  Since Stack and StackInstall resources generate deployments and jobs with matching names, they are limited, in practice to 63 characters, which this PR addresses for both hosted and local stack deployments and jobs.

pkg/controller/stacks/stack/stack.go
displague: VS code is collapsing the spaces here. :man_shrugging: 

I'll split this into two lines. VS code is not changing the spacing when I do that.

pkg/controller/stacks/stack/stack_test.go
displague: VSCode is collapsing these spaces. I'm going to let this stand since I can not easily split these into separate declarations in the test cases.  If I make each combination into a new constant the names would be even longer: `resourceNamePlusDashAlphabet` :stuck_out_tongue:

displague: `go fmt` seems to be doing this when there are multiple arguments to a function:
https://play.golang.org/p/xVTwhIh6eAb

jbw976: VS Code collapsing the spaces implies to me that that is the idiomatic thing to do :)

pkg/stacks/relationship.go
displague: I've left a warning on preserveNSLength :+1:

pkg/stacks/relationship.go
displague: The `MultiParentLabelPrefix` function relies on the `LabelMultiParentPrefix` constant which includes a `/`.  Namespaces can not contain `/` in their resource names. 

If that ever changes, without the rules of label prefixes changing (they allow a single `/`) we would have problems. We have to agree that the world is flat or we have nothing to stand on.

pkg/stacks/relationship_test.go
displague: I could come up with another character that can't be confused as a word or abbreviation in this context, but I would also have to update all of the suffixes.  I could add `Letter` to the variable name, but it becomes lengthy.

I think looking at the assignment disambiguates the meaning of the name of this limited use testing variable.

pkg/stacks/truncate/truncate.go
displague: I tried removing that `nolint` directive but then `gosec` fired.  I've replaced `blacklist` with `gosec` here. `make` has been appeased.

displague:
@jbw976 @suskin The latest commit responds to your feedback with documentation and comment changes.

displague:
Force pushed a few more comment updates that I missed in the last commit, https://github.com/crossplane/crossplane/compare/e62afa38c168c7403fa1a8b38dc0652928a3a4f1..a6b201aee9de31cbd78c091c24c12703df44cf72

docs/troubleshoot.md
```diff
@@ -131,3 +132,81 @@ For example, for a Workload object (`workloads.compute.crossplane.io`) named
...
+
+### Example
+
+Long resource names may be present on the tenant.
+
+```console
+$ name=stack-with-a-really-long-resource-name-so-long-that-it-will-be-truncated
+$ ns=this-is-just-a-really-long-namespace-name-at-the-character-max
+$ kubectl create ns $ns
+$ kubectl crossplane stack install --namespace $ns crossplane/sample-stack-wordpress:0.1.1 $sn
```
jbw976: what is `$sn` at the end of this line, should that actually be `$name`?

displague: Force pushed the last commit with this tweak.  I used `$sn` locally and replaced it with `$name` in the doc for readability.

PR #1341, opened 2020-03-16
[Bug] fix KAR remote resource patch and switch templates to raw.Extension

hasheddan:

This fixes a bug in the `KubernetesApplicationResource` controller where it was attempting to patch the remote resource with the remote resource's current config (i.e. no-op). It also updates the templates to use `raw.Extension` instead of `unstructured.Unstructured`.

*Note: further improvements should still be made for this controller, but this fixes a vital bug in the short term.*

Created `KubernetesApplication` with templates for a `Deployment` and `Service`. After the resources were successfully created in the remote cluster, I updated the `Replicas` field on the `Deployment` template and watched the remote resource be updated as well.

### Checklist

I have:
  appropriate.

[definition of done].


pkg/controller/workload/kubernetes/resource/resource.go
```diff
@@ -153,12 +165,14 @@ type remoteCluster struct {
 func (c *remoteCluster) sync(ctx context.Context, ar *v1alpha1.KubernetesApplicationResource, secrets []corev1.Secret) (v1alpha1.KubernetesApplicationResourceState, error) {
 	meta.AddFinalizer(ar, finalizerName)
 
-	// Our CRD requires template to be specified, but just in case...
-	if ar.Spec.Template == nil {
-		return v1alpha1.KubernetesApplicationResourceStateFailed, errMissingTemplate
+	// We copy the resource template here so we can modify its namespace and
+	// remote controller annotations without persisting those changes back to
+	// the KubernetesApplicationResource.
```
negz: Nit: this comment doesn't make as much sense now that we're rendering the template rather than copying it.

pkg/controller/workload/kubernetes/resource/resource.go
```diff
@@ -244,31 +257,39 @@ type unstructuredClient struct {
...
+
 	key, err := client.ObjectKeyFromObject(remote)
 	if err != nil {
-		return nil, errors.Wrap(err, "cannot take object key from resource")
+		return nil, errors.Wrap(err, errGetKey)
 	}
+
+	// TODO(hasheddan): this Get operation and subsequent controller check can
+	// be eliminated once resource.Apply is refactored to accept arbitrary
+	// ApplyOption functions.
```
negz: Note for anyone else reading this: it wasn't immediately obvious to me why resource.Apply doesn't accept arbitrary ApplyOption functions today, until I recalled that ApplyOptions are applied to an options struct, which controls flow within the function.

PR #1347, opened 2020-03-17
[Bug] Fix KAR deletion

hasheddan:

It appears deletion was messed up by my changes in #1341. This fixes the the problem. Not sure how deletion was successful there but I have re-tested and it is working here.

Created `KubernetesApplication` with `Deployment` and `Service`. Modified it and saw changes propagate to remote cluster. Deleted and saw resources deleted in remote cluster.

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/pkg/controller/workload/kubernetes/resource/resource.go
+++ b/pkg/controller/workload/kubernetes/resource/resource.go
@@ -132,11 +132,11 @@ type syncdeleter interface {
 }
 
 type unstructuredSyncer interface {
-	sync(ctx context.Context, data []byte) (*v1alpha1.RemoteStatus, error)
+	sync(ctx context.Context, template *unstructured.Unstructured) (*v1alpha1.RemoteStatus, error)
 }
 
 type unstructuredDeleter interface {
-	delete(ctx context.Context, data []byte) error
+	delete(ctx context.Context, template *unstructured.Unstructured) error
 }
 
 type unstructuredSyncDeleter interface {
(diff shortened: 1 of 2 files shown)


pkg/controller/workload/kubernetes/resource/resource.go
```diff
@@ -216,7 +216,11 @@ func (c *remoteCluster) delete(ctx context.Context, ar *v1alpha1.KubernetesAppli
 	ensureNamespace(template)
 	setRemoteController(ar, template)
 
-	if err := c.unstructured.delete(ctx, ar.Spec.Template.Raw); err != nil {
+	raw, err := json.Marshal(template)
+	if err != nil {
+		return v1alpha1.KubernetesApplicationResourceStateFailed, errors.Wrap(err, errUnmarshalTemplate)
+	}
+	if err := c.unstructured.delete(ctx, raw); err != nil {
```
negz: I'm not a huge fan of needing to marshal to unstructured, set some metadata, then marshal back to YAML, which is again unmarshalled to unstructured in the unstructuredClient. Would it make more sense for the unstructuredClient to accept and operate on `*unstructured.Unstructured` and unmarshal once in the `remoteCluster`?

hasheddan: @negz I agree that it is a bit cumbersome. However, it feels like it makes the point of using `raw.Extension` a little less impactful if we are still passing around `*unstructured.Unstructured`. That being said, I think the point of passing around `[]bytes` is to guard against direct modification of the object, and that may not be as important in this situation where there are not external consumers of these methods. What do you think? Am I off base in how I am thinking about the value of using `raw.Extension`?

negz: @hasheddan I believe the point of using `raw.Extension` is mostly to avoid embedding a real, functional `runtime.Object` inside another `runtime.Object`. When we embed a real `Object` we must always remember to deepcopy it to avoid any changes we make being propagated back to our "template".

I believe we can fulfil this goal by simply unmarshaling a `raw.Extension` early on in the reconcile, close to where we parse the KAR struct. We'll then have an *unstructured.Unstructured` we can plumb down into our various helpers, clients, etc without those methods having to worry about deepcopying before modifying.

Apologies that I didn't catch this in the previous PR!

hasheddan: @negz no worries! Could you give this another look? I am going to run manual testing again to confirm operations :+1:

PR #1348, opened 2020-03-18
Allow package type Addon in app.yaml

hasheddan:

Allows `PackageType: Addon` for stack installs.

This has not been formally tested with an `Addon` installation, but there should be one ready soon at https://github.com/crossplane/addon-oam-kubernetes-remote. This is very small diff in that it simply allows for `Addon` in the enum for validation, so it should not cause any problems. I will update with testing nonetheless.

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/apis/stacks/v1alpha1/types.go
+++ b/apis/stacks/v1alpha1/types.go
@@ -443,7 +443,7 @@ type AppMetadataSpec struct {
 	// drives the RBAC generation process.
 	DependsOn []StackInstallSpec `json:"dependsOn,omitempty"`
 
-	// +kubebuilder:validation:Enum=Provider;Stack;Application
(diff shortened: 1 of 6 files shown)


hasheddan:
@displague the definition is still not 100% clear, but an `addon` broadly means something that integrates with crossplane to add / modify functionality. It is different from a `provider` in that it is not introducing communication with a new API and provisioning resources on other platforms (thought it may provision resources in remote k8s clusters). You can think of them as similar to [kubernetes addons](https://kubernetes.io/docs/concepts/cluster-administration/addons/).

hasheddan:
@jbw976 good catch! I have updated the doc :)

PR #1364, opened 2020-03-24
release doc: add template stacks

suskin:

We maintain a set of template stacks, and we cut semver releases of our
stacks as part of the full release process for crossplane. So we want to
release the template stacks that we maintain also. They have a slightly
different process from existing stacks, so we need to add some guidance
to our documentation.

* I have done a dry run of the steps with `stack-minimal-gcp` and its Jenkins jobs.
* We will want to consider the first run-through for a real release as part of the final testing.

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/docs/release/release-process.md
+++ b/docs/release/release-process.md
@@ -23,6 +23,17 @@ with a very high level sequential overview for how to run the release process.
        semver
     1. **build/publish**: Run build pipeline to publish build with official
        semver
+1. **release template stacks**: Run the release process for each
+   **template stack** that we maintain. Note that the process for
+   template stacks is slightly different from the stack release process.
+    1. **test** Test builds from the release branch (typically
+       `master`), fix any critical bugs that are found
+    1. **version**: Update all version information in the docs, as
+       appropriate
+    1. **tag**: Run the tag pipeline to tag the release branch with an
+       official semver
+    1. **build/publish**: Run the publish pipeline to publish build with
+       official semver
 1. **docs/examples**: In Crossplane's release branch, update all examples, docs,
    and integration tests to reference the new versions
 1. **tag**: Run the tag pipeline to tag Crossplane's release branch with an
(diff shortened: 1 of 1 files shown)


docs/release/release-process.md
```diff
@@ -192,6 +201,77 @@ success, all release artifacts will be officially published.
...
+### Template Stack Tag And Publish Pipeline
+
+Here is the list of all template stacks:
+
+* `stack-minimal-gcp`
+* `stack-minimal-aws`
+* `stack-minimal-azure`
+* `app-wordpress`
+
+Each one should be released as part of a complete release.
```
hasheddan: I don’t think we need a change here, but despite us saying the releases of providers / stacks / apps, we still talk about them as though they are coupled. I don’t think this is necessarily a bad thing, but I would like to get to a point in the future where we just do new releases of non-core Crossplane repos either on their own cadences or just when needed 👍

docs/release/release-process.md
```diff
@@ -18,6 +18,15 @@ level sequential overview for how to run the release process.
...
     1. **tag**: Run the tag pipeline to tag the release branch with an official semver
     1. **build/publish**: Run build pipeline to publish build with official semver
+1. **release template stacks**: Run the release process for each
+   **template stack** that we maintain. Note that the process for
+   template stacks is slightly different from the stack release process.
+    1. **test** Test builds from the release branch, fix any critical
+       bugs that are found
+    1. **version**: Update all version information in the docs, as
+       appropriate
+    1. **tag**: Run the tag pipeline to tag the release branch with an official semver
```
jbw976: i think you mentioned the template stacks aren't using release branches? if so, we could remove release branch from this sentence

on the bigger topic of release branches, what's the thinking/justification for releasing directly from master? simplicity and streamlining?

if we want to do a patch release for the last release after master has moved on, will we create a release branch that starts at the commit we released from master? (i.e. retroactively create the release branch)

docs/release/release-process.md
```diff
@@ -18,6 +18,15 @@ level sequential overview for how to run the release process.
        tests in the release branch
     1. **tag**: Run the tag pipeline to tag the release branch with an official semver
     1. **build/publish**: Run build pipeline to publish build with official semver
+1. **release template stacks**: Run the release process for each
+   **template stack** that we maintain. Note that the process for
+   template stacks is slightly different from the stack release process.
+    1. **test** Test builds from the release branch, fix any critical
+       bugs that are found
+    1. **version**: Update all version information in the docs, as
```
jbw976: which docs get updated for template stacks? is that mentioned in more detail later on?

docs/release/release-process.md
```diff
@@ -192,6 +201,77 @@ success, all release artifacts will be officially published.
...
+strategy, but for now we are not using branches because it is simpler.
+
+Note that Template Stacks **do not** require any code changes to update
+the version.
+
+### Template Stack Tag And Publish Pipeline
+
+Here is the list of all template stacks:
+
+* `stack-minimal-gcp`
```
jbw976: can you link each of these to their repo, similar to how we are doing it in https://github.com/crossplane/crossplane/blob/master/docs/release/release-process.md#scope?

docs/release/release-process.md
```diff
@@ -192,6 +201,77 @@ success, all release artifacts will be officially published.
...
+#### Publish the Template Stack
+
+After the tag pipeline has been run and the repository has been tagged,
+you can run the `publish` job for the template stack. For example,
+here's a [link to the stack-minimal-gcp publish
+job](https://jenkinsci.upbound.io/job/crossplane/job/stack-minimal-gcp/job/publish/job/master/).
+This will kick off the official release build and upon success, all
+release artifacts will be officially published. This should also be run
+from the `master` branch in most cases. Or, if a release branch was
+used, from the release branch. The tag parameter should be used, and the
```
jbw976: help me understand in more detail (or point me to the appropriate doc) how the version is generated for the built and published template stack image.  We tagged the release commit in master in the previous step, and in this publishing step we are providing the version tag again.  How does the publishing step use that input? Does it use it directly for the version of the image it will publish, or does it instead use it to find the right commit to build/publish?

Along those lines, are variables like this `STACK_VERSION` for app-wordpress still needed? or are those used in the automated pipelines? https://github.com/crossplane/app-wordpress/blob/master/stack.env#L4

docs/release/release-process.md
jbw976: looks like you partially answer this below with:

> For Template Stacks, we do not use release branches unless we need to
make a patch release. In the future we may need a more robust branching
strategy, but for now we are not using branches because it is simpler.

but still curious about doing a patch later on, if we need to retroactively create a release branch 🤔

suskin: I'm now noticing that there is already a patch release section to this document which tells the reader to go to the instructions in each other section, so I will think a little bit more about whether we can add anything here to help.

docs/release/release-process.md
hasheddan: That is my understanding as well :)

hasheddan:
@suskin yep that is true! I am happy to do the rebase either way no matter which one gets merged first if it gets messy :)

docs/release/release-process.md
```diff
@@ -192,6 +201,77 @@ success, all release artifacts will be officially published.
...
+controller-based stacks that we maintain. Their processes are similar
+but a little simpler. This section will walk through how to release the
+Template Stacks themselves, and does not directly apply to core
+Crossplane.
+
+For Template Stacks, we do not use release branches unless we need to
+make a patch release. In the future we may need a more robust branching
+strategy, but for now we are not using branches because it is simpler.
+
+Note that Template Stacks **do not** require any code changes to update
```
muvaf: Actually if we release a new version of [`templating-controller`](https://github.com/crossplane/templating-controller), we'd probably want to update the [`controllerImage`](https://github.com/crossplane/stack-minimal-gcp/blob/master/.registry/behavior.yaml#L8) in the template stack to the latest version.

docs/release/release-process.md
```diff
@@ -192,6 +201,77 @@ success, all release artifacts will be officially published.
...
+
+### Template Stack Tag And Publish Pipeline
+
+Here is the list of all template stacks:
+
+* `stack-minimal-gcp`
+* `stack-minimal-aws`
+* `stack-minimal-azure`
+* `app-wordpress`
+
```
muvaf: Could you add `templating-controller` release, too?

docs/release/release-process.md
displague: Can "release stacks" and "release template stacks" merge? perhaps:

"release stacks: Run the release process for each stack and template stack that we maintain"
"version: Update all version information in the docs and integration tests, where appropriate in the release branch"

docs/release/release-process.md
displague: We should also be noting that a release of the `templating-controller` should result in an update to `cluster/charts/crossplane-controllers/values.yaml.tmpl` and `docs/install-crossplane.md`.

displague: In terms of timing, it may be advisable to release templating-controller ahead of crossplane when coordinating same-day releases.

docs/release/release-process.md
suskin: > Can "release stacks" and "release template stacks" merge?

I considered this . . . they **could** merge, but I decided to keep them separate because the way the jobs are structured is a little different, and the way versions are updated is also different between the two. It seemed like it would be more clear if they are two separate sections.

docs/release/release-process.md
suskin: Yeah. This is all true, but it's something that I'm thinking should be documented on the templating-controller side. We've done a pretty good job of decoupling templating-controller releases from crossplane releases already, so I'm thinking we should keep it that way. So I would put this in documentation on the templating-controller side so that we see it when we're releasing the templating-controller.

docs/release/release-process.md
suskin: I considered this, and I decided not to because we've been doing a pretty good job of keeping the releases decoupled. I think it could make sense to mention templating-controller here for completeness, but I'm thinking it makes sense for us to continue releasing the templating-controller separately. It makes it easier for us to manage the dependency if we're able to release it separately so that we don't need to update any versions when we release crossplane and the stacks.

docs/release/release-process.md
suskin: I've added some notes about creating patch releases for template stacks

docs/release/release-process.md
suskin: This is done

docs/release/release-process.md
suskin: I've added some notes discussing the templating-controller project

docs/release/release-process.md
suskin: I've added some notes discussing the templating-controller project

muvaf: Yeah, makes sense.

docs/release/release-process.md
muvaf: Thank you for mentioning `templating-controller` but would you mind adding a note to say something like `check the latest version of templating-controller and update to it in behavior.yaml` ? At least, the releaser would make sure that template stacks do use the latest version before the release.

suskin: Hmm double checking that is a good idea . . . we may want them to check it in the crossplane helm chart too.

jbw976: i'll add a note for this in my #1367 PR @suskin and @muvaf

PR #1367, opened 2020-03-25
release: update release process doc to formalize process in use for the past couple releases

### LINKED ISSUE #1187: Update release process guide to include instructions used in v0.7

The v0.7 release used a modified release process flow to address the following issues:

* Crossplane and stacks have not been using a pinned semantic version of crossplane-runtime
* Stacks have not been using a pinned semantic version of crossplane

In order to get a semantic version of Crossplane to use in the Stacks' go.mod files, we updated all the docs/examples/guides in Crossplane's release branch and then tagged that commit with the official v0.7.0 release tag, but did not run the final release build yet.

This is a shorthand flow/sequence that we used and should be incorporated into the [release process guide](https://github.com/crossplaneio/crossplane/blob/master/docs/release/release-process.md):

```
- Regenerate API Docs / Do any docs updates in Crossplane
- Release crossplane-runtime as v0.4.0
- Pin Crossplane to new crossplane-runtime release
- Create RC tag on Crossplane master
- Cut Crossplane release branch
- Do release branch prep for Crossplane (including updating refs to yet to be released stacks) on release-0.7 branch
- Tag Crossplane as v0.7.0 on release-0.7 branch
- Pin stacks to new crossplane-runtime release v0.4.0
- Pin stacks to Crossplane v0.7.0
- Create RC tag on stacks master
- Cut stacks release branches
- Do release branch prep for stacks
- Release stacks
- Release Crossplane
- Promote Crossplane and stacks
```

jbw976:

This PR updates the release process to capture the actual process we've been using in the past couple releases.  The reasons for this are captured in #1187, which this PR fixes.

Also included in this PR is an explanation of the new tagging/versioning strategy that now uses semver pre-release tags correctly.  When branching for a release, master will be tagged with a pre-release tag for the **next** version that master will be working towards (with more pre-releases).

I am using the [Rewrap extension](https://marketplace.visualstudio.com/items?itemName=stkb.rewrap) for VS Code to wrap all markdown lines at 80 chars, while maintaining syntax integrity.  Some automatic wrapping updates were made since it runs every time I save, it would be good to standardize on that within the community to avoid extra whitespace/wrapping changes in PRs.

Fixes #1187 

We will walk through these steps for the v0.9 release.  Until this PR is merged, the latest can be found at https://github.com/jbw976/crossplane/blob/rel-proc-updates/docs/release/release-process.md

### Checklist

I have:
  appropriate.

[definition of done].

[skip ci]

### DIFF
--- a/docs/release/release-process.md
+++ b/docs/release/release-process.md
@@ -10,41 +10,54 @@ with a very high level sequential overview for how to run the release process.
 
 1. **feature freeze**: Merge all completed features into master branches of all
    repos to begin "feature freeze" period
-1. **RC tag**: Run tag pipeline to tag the release candidate (RC) build in
-   master for each repo
-1. **branch**: Create a new release branch using the GitHub UI for each repo
-1. **release stacks**: Run the release process for each **stack** that we
+1. **API docs/user guides**: Regenerate API docs and update all user guides with
+   current content for scenarios included in the release
+1. **release crossplane-runtime**: Tag and release a new version of
+   crossplane-runtime using the GitHub UI.
+1. **pin crossplane dependencies**: Update the go modules of core crossplane in
+   master to depend on the newly released version of crossplane-runtime.
+1. **pre-release tag crossplane**: Run tag pipeline to tag the start of
+   pre-releases in master in the crossplane repo
+1. **branch crossplane**: Create a new release branch using the GitHub UI for
+   the crossplane repo
+1. **crossplane release branch prep**: In Crossplane's release branch, update
+   all examples, docs, and integration tests to update references and versions,
+   including the yet to be released versions of providers and stacks.
+1. **tag**: Run the tag pipeline to tag Crossplane's release branch with an
+   official semver
+1. **release providers**: Run the release process for each **provider** that we
    maintain
+    1. **pin dependencies**: Update the go modules of each provider repo to
+       depend on the new version of crossplane and crossplane-runtime.
+    1. **pre-release tag**: Run tag pipeline to tag the start of pre-releases in
+       **master** of each provider repo
+    1. **branch**: Create a new release branch using the GitHub UI for the
+       provider repo
+    1. **release branch prep**: In the release branch, update all examples,
+       docs, and integration tests to update references and versions
     1. **test** Test builds from the release branch, fix any critical bugs that
        are found
-    1. **version**: Update all version information in the docs, and integration
-       tests in the release branch
     1. **tag**: Run the tag pipeline to tag the release branch with an official
        semver
     1. **build/publish**: Run build pipeline to publish build with official
        semver
-1. **release template stacks**: Run the release process for each
-   **template stack** that we maintain. Note that the process for
-   template stacks is slightly different from the stack release process.
-    1. **test** Test builds from the release branch (typically
-       `master`), fix any critical bugs that are found
-    1. **version**: Update all version information in the docs, as
-       appropriate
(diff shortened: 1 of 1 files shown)


docs/release/release-process.md
```diff
@@ -178,26 +250,52 @@ and collect feedback.  Do **not** hit "Publish release" yet.
...
+#### Pin Provider Dependencies
 
-At this point, we should have a release branch created in all repos with all
-issues resolved that were triaged to be release blockers.  This section will
-walk through how to release the Stacks themselves and does not directly apply to
-core Crossplane.
+Similar to core crossplane, each provider should have its crossplane related
+dependencies pinned to the versions that we are releasing. In the **master**
+branch of each provider repo, update the crossplane and crossplane-runtime
+dependencies to the versions we are releasing.
```
hasheddan: Just a note here, we ran into issues in the `v0.9` release process because of an outdated version of `crossplane-tools`. We do not currently version `crossplane-tools` so it is difficult to determine what version you should be pinned to, but in-practice it is usually the latest commit on `master`. Not sure if we want to call that out here, but wanted to make sure it was mentioned :+1:

docs/release/release-process.md
```diff
@@ -140,6 +171,29 @@ The `master` branch can now be opened for new features since we have a safe
...
+* Documentation, such as [Installation
+  instructions](https://github.com/crossplane/crossplane/blob/release-0.9/docs/install.md#installing-infrastructure-providers),
+  and
+  [Stack](https://github.com/crossplane/crossplane/blob/release-0.9/docs/stack.md)
+  and
+  [App](https://github.com/crossplane/crossplane/blob/release-0.9/docs/app.md)
+  guides.
+  * searching for `:master` will help a lot here
+  * `kubectl -k` statements will contain `?ref=master`, which should be updated
+    to the release branch name instead, e.g. `?ref=release-0.4`
```
hasheddan: These have been removed so one less thing to update :tada:

docs/release/release-process.md
```diff
@@ -140,6 +171,29 @@ The `master` branch can now be opened for new features since we have a safe
 release branch to continue bug fixes and improvements for the release itself.
 Essentially, `master` is free to now diverge from the release branch.
 
+### Release Branch Prep
```
hasheddan: Given my slip-up on opening PRs against `master` in the `v0.9` release, it would be nice if we could establish some safe guards here. I think following these new more detailed notes helps a ton, an usually it will get caught in review, but I have seen some projects handle this sort of thing with a separate tool that is used for release that basically just wraps a Github client and enforces certain policies. Out-of-scope for this document right now, but always good to improve process where possible :)

docs/release/release-process.md
jbw976: I've added a note about `crossplane-tools` as well, and how in practice we use latest from master.

docs/release/release-process.md
jbw976: Nice, I've removed this note about `?ref=master` hits in the docs during release branch prep

docs/release/release-process.md
jbw976: yes i definitely agree! more automation is definitely desired from this process to not only reduce opportunities for human mistakes but also save time and energy too.  Hopefully you're getting good ideas from being in sig release upstream :)

jbw976:
Thanks for the reviews @hasheddan! I've incorporated/responded to all your feedback, so I think this is ready for a final look.

PR #1369, opened 2020-03-25
[Release] Update docs and references prior to v0.9 release

hasheddan:

This turned into a more unwieldy change than expected because I realized that the new docs directories and file names flow through to the urls (i.e. https://crossplane.io/docs/v0.9/1_infra_operators/1_manually_adding_clusters.html) which is pretty awful. I removed all number prefixes and updated links for this release. We can determine how we want to handle them on `master` in the future. I think we will likely remove them there but I do like how they keep the directories organized. We could potentially modify the files names in the build pipeline that publishes them.

`make reviewable`

### Checklist

I have:
  appropriate.

[definition of done].


hasheddan:
Note: I tested that this rendered correctly by running the website locally :+1:

cluster/charts/crossplane/README.md
```diff
@@ -144,7 +144,7 @@ metadata:
   name: provider-gcp
   namespace: gcp
 spec:
-  package: "crossplane/provider-gcp:master"
+  package: "crossplane/provider-gcp:v0.7.0"
```
negz: Should these be in here?

hasheddan: Do you mean in the Helm `README` specifically? This is copied here as part of `make generate` and I think we do want it there. If not I will have to adjust our `Makefile` or we will fail on `check-diff`.

PR #1376, opened 2020-03-27
[release-0.9] Unwrap inline links for Jekyll html conversion

hasheddan:

This unwraps in-line links in the docs such that they will render correctly in the Jekyll site. Currently the https://github.com/benbalter/jekyll-relative-links does not handle line breaks in markdown in-line links, despite being valid markdown. This is **not** evident when running the docs site locally, so this change is only being made on `release-0.9` as it is not a long-term sustainable solution for `master`. Possible solutions in master:

- Using only named links. We have done this before, but I think it is harder to maintain links in this way. It is not a good experience to edit a doc that uses named links and it harder to tell in review if the right link is used or exists.
- Using Jekyll [permalinks](https://jekyllrb.com/docs/permalinks/). This is a potential solution, but I think there may be a blocker in that if we force links to `.md` then the links that get converted to `.html` will no longer work (i.e. we will reverse the problem in that *only* line-breaked links will work). However, I think the plugin will force non-wrapped links to match their permalink, so it may work afterall.
- Adding a fix to https://github.com/benbalter/jekyll-relative-links to handle line-wrapped links

*Note: the Github UI is helpful for review here because it shows the links as blue when they are not line wrapped.*

n/a

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -10,9 +10,7 @@ features:
   Kubernetes API.
 - Management and scheduling of configuration data across multiple Kubernetes
   clusters.
-- Separation of concern between [infrastructure
-  owners](infra_operators/overview.md), [application
-  owners](app_operators/overview.md), and [developers](developers/overview.md).
+- Separation of concern between [infrastructure owners](infra_operators/overview.md), [application owners](app_operators/overview.md), and [developers](developers/overview.md).
 - Infrastructure agnostic packaging of applications and their dependencies.
 - Scheduling applications into different clusters, zones, and regions.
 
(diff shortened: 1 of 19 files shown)


negz:
> Using only named links. We have done this before, but I think it is harder to maintain links in this way. It is not a good experience to edit a doc that uses named links and it harder to tell in review if the right link is used or exists.

My vote is for named links. Your points are totally valid, but I feel like the improved flow of reading the document in markdown is worth it, particularly given that we typically review and edit docs in their markdown form.

hasheddan:
@negz that is fair. I very much prefer the in-line links and have proposed a potential fix to the maintainer of the Jekyll plugin we use: https://github.com/benbalter/jekyll-relative-links/issues/61

That being said, it is obviously more important that the links work and the majority of folks who are editing like the format, so I am more than happy to comply with the named links :)

PR #1377, opened 2020-03-27
stacks: add stack-manager option to force image pull policy

### LINKED ISSUE #1346: Stack Manager option to force ImagePullPolicy

In multi-tenant environments, images fetched for one Stack unpack or Stack controller become available in the node level docker registry.  This could allow other tenants to access those images without the appropriate imagePullSecrets.

The Stack Manager should offer an option to force an imagePullPolicy. Users should not be trusted to provide `imagePullPolicy: Always` in their `StackInstall` resources.

See https://kubernetes.io/docs/concepts/containers/images/#use-cases Case 3 and 4

displague:

In multi-tenant environments, images fetched for one Stack unpack or Stack
controller become available in the node level docker registry. This could
allow other tenants to access those images without the appropriate
imagePullSecrets.

This change introduces a stack-manager parameter: `--force-image-pull-policy`,
which when set to a non-empty string, will override the imagePullPolicy set in
any StackInstall spec (affecting the install job and upack) and will force an
imagePullPolicy for each pod in a Stack deployment.

Fixes #1346

Unit tests verify that Deployments and Jobs receive the correct pull policy.

This was run in the cluster/local/hosted-test.sh environment with the stack-manager deployment including the `--force-image-pull-policy Always` argument. A StackInstall was created with an `imagePullPolicy: Never` and the resulting Job and Deployment correctly adopted the imagePullPolicy of `Always`.

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/cluster/charts/crossplane/templates/stack-manager-deployment.yaml
+++ b/cluster/charts/crossplane/templates/stack-manager-deployment.yaml
@@ -40,6 +40,10 @@ spec:
         - {{ .Values.templateStacks.controllerImage | quote }}
         {{- end }}
         {{- end }}
+        {{- if .Values.forceImagePullPolicy }}
+        - --force-image-pull-policy
+        - {{ .Values.forceImagePullPolicy | quote }}
+        {{- end }}
         {{- range $arg := .Values.args }}
         - {{ $arg }}
         {{- end }}
--- a/pkg/controller/stacks/install/installjob_test.go
+++ b/pkg/controller/stacks/install/installjob_test.go
@@ -419,6 +419,54 @@ func withJobSource(src string) jobModifier {
 	}
 }
 
+// withJobPullPolicy sets the image pull policy on the InitContainer, whose
+// container is derived from the stack. The non-init containers are not affected
+// since they reflect the imagePullPolicy of the stack-manager itself
+func withJobPullPolicy(pullPolicy corev1.PullPolicy) jobModifier {
+	return func(j *batchv1.Job) {
+		ics := j.Spec.Template.Spec.InitContainers
+		for i := range ics {
+			ics[i].ImagePullPolicy = pullPolicy
+		}
+	}
+}
+
+// withJobExpectations sets default job expectations
+func withJobExpectations() jobModifier {
+	return func(j *batchv1.Job) {
+		zero := int32(0)
+		j.SetResourceVersion("1")
+		j.SetGroupVersionKind(batchv1.SchemeGroupVersion.WithKind("Job"))
+		j.SetLabels(stacks.ParentLabels(resource()))
+		j.Spec.BackoffLimit = &zero
+		j.Spec.Template.Spec.Volumes = []corev1.Volume{{
+			Name:         "package-contents",
+			VolumeSource: corev1.VolumeSource{EmptyDir: &corev1.EmptyDirVolumeSource{}},
+		}}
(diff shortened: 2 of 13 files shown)


pkg/controller/stacks/install/installjob_test.go
```diff
@@ -418,6 +419,54 @@ func withJobSource(src string) jobModifier {
...
+		j.Spec.Template.Spec.Containers = []corev1.Container{
+			{
+				Name:  "stack-unpack-and-output",
+				Image: stackPackageImage,
+				Args: []string{
+					"stack",
+					"unpack",
+					"--content-dir=/ext-pkg/.registry",
+					"--permission-scope=Namespaced",
+					"--templating-controller-image=",
```
suskin: Do we need a value for this parameter?

pkg/controller/stacks/stack/stack.go
```diff
@@ -818,6 +822,18 @@ func (h *stackHandler) prepareDeployment(d *apps.Deployment) {
 	meta.AddLabels(d, labels)
 
 	d.Spec.Template.Spec.ServiceAccountName = h.ext.Name
+
+	if h.forceImagePullPolicy != "" {
+		for _, c := range [][]corev1.Container{
+			d.Spec.Template.Spec.Containers,
```
suskin: Does this change any behaviors for template stacks using the templating-controller?

pkg/controller/stacks/install/installjob_test.go
displague: In the test environment we would only have a value here if we supplied it in the harnessing or fixture.  This particular set of tests isn't concerned with that value.  This is the first place that I am aware of where we are checking the final shape of an install job, so we may revisit this in the future as we add more tests and composition helpers (`withTemplatingControllerImage()`).

pkg/controller/stacks/stack/stack.go
displague: The templating controller is treated like any controller. Since the change that introduced `imagePullPolicy` as a StackInstall parameter, the templating controller adopts the supplied pull policy.  If the StackManager wants to enforce a value, that value should be enforced on the templating controller as well.

cluster/charts/crossplane/templates/stack-manager-deployment.yaml
```diff
@@ -40,6 +40,10 @@ spec:
         - {{ .Values.templateStacks.controllerImage | quote }}
         {{- end }}
         {{- end }}
+        {{- if .Values.forceImagePullPolicy }}
```
jbw976, line 43: this tests for non empty string values?

jbw976:
LGTM after this has been rebased and the conflicts are resolved!

jbw976:
@displague please @ mention me on this PR when it's ready to merge, thanks!

cluster/charts/crossplane/templates/stack-manager-deployment.yaml
displague: Yes - the pattern is repeated from the previous block: 
`{{- if .Values.templateStacks.controllerImage }}`

displague:
@jbw976 pushed with the merge conflicts resolved https://github.com/crossplane/crossplane/compare/e598f1f9f298c8d112838a4c8162e015fe3d0c93..555b54707fb38a4c14c471dd3c1653d7b0bd3cbf

PR #1381, opened 2020-03-30
Remove KAR owner UID annotation on remote resources

hasheddan:

This removes the owner `UID` annotation on remote Kubernetes resources in order to allow a lost `KubernetesApplication` to re-establish control of its remote resources. However, this also means that any`KubernetesApplicationResource` with the same name and namespace of the owner annotations on existing resources in a remote Kubernetes cluster can establish control of them. This is not a huge issue if the original `KubernetesApplicationResource` still exists because the name / namespace conflict would serve as a guard, but in the case that remote resources are orphaned and a user does not wish for a `KubernetesApplication` to ever be able to establish control over them again, they should make sure to remove the owner annotations.

Keeping in draft in case we need further discussion before merge.

Unit tests only right now. Manual testing will be done prior to merge.

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/pkg/controller/workload/kubernetes/resource/resource.go
+++ b/pkg/controller/workload/kubernetes/resource/resource.go
@@ -559,5 +559,12 @@ func haveSameController(a, b metav1.Object) bool {
 		return false
 	}
 
-	return a.GetAnnotations()[RemoteControllerUID] == b.GetAnnotations()[RemoteControllerUID]
+	// NOTE(hasheddan): we set an annotation for the UID of the remote
+	// controller but do not check that it matches because it would prohibit a
+	// lost KubernetesApplication from re-establishing ownership of its remote
+	// resources.
+	if a.GetAnnotations()[RemoteControllerNamespace] != b.GetAnnotations()[RemoteControllerNamespace] {
+		return false
+	}
+	return a.GetAnnotations()[RemoteControllerName] == b.GetAnnotations()[RemoteControllerName]
 }
--- a/pkg/controller/workload/kubernetes/resource/resource_test.go
+++ b/pkg/controller/workload/kubernetes/resource/resource_test.go
@@ -1686,7 +1686,7 @@ func TestHaveSameController(t *testing.T) {
 			want: false,
 		},
 		{
-			name: "HasDifferentController",
+			name: "HasDifferentUID",
 			a:    service,
 			b: &corev1.Service{
 				ObjectMeta: metav1.ObjectMeta{
@@ -1697,7 +1697,7 @@ func TestHaveSameController(t *testing.T) {
 					},
 				},
 			},
-			want: false,
+			want: true,
 		},
 	}
 


pkg/controller/workload/kubernetes/resource/resource.go
```diff
@@ -531,7 +530,6 @@ func setRemoteController(ctrl metav1.Object, obj metav1.Object) {
 	meta.AddAnnotations(obj, map[string]string{
 		RemoteControllerNamespace: ctrl.GetNamespace(),
 		RemoteControllerName:      ctrl.GetName(),
-		RemoteControllerUID:       string(ctrl.GetUID()),
```
negz: @hasheddan I was thinking this morning that perhaps it would be worth setting, but not checking the UID. This would give us some debuggability around which resource originally set the annotations. On the other hand it could be confusing that we set it but don't check it. What do you think? If you do elect to keep it we should probably add a comment near the namespace and name check explaining why we don't check it.

negz:
Some thoughts around guaranteeing ownership:

* Make the current behaviour (checking UID) optional, based on some field like `spec.adoptable`
* The UID mostly prevents the case in which two different Crossplanes are trying to control a remote resource. Perhaps we can find some other way to uniquely identify a Crossplane instance?

hasheddan:
@negz I like the idea of being able to designate something as "adoptable", even outside of just this context. I wonder if it would make sense to assign a UID to every crossplane instance and pass it around to all of the controllers it manages. We could then do unique naming on some concatenation of the name / namespace / crossplane ID.

negz:
@hasheddan I think that's worth looking into, but for now I'd prefer to just remove the UID check and merge this. Direction LGTM; happy to merge once you've tried it out and addressed (or not) the one comment.

hasheddan:
@negz I have updated with a note and tested that the following `KubernetesApplication` creates and cleans up its remote resources correctly:

```yaml
apiVersion: workload.crossplane.io/v1alpha1
kind: KubernetesApplication
metadata:
  name: helloworld
  namespace: cp-quickstart
  labels:
    app: helloworld
spec:
  resourceSelector:
    matchLabels:
      app: helloworld
  resourceTemplates:
    - metadata:
        name: helloworld-namespace
        labels:
          app: helloworld
      spec:
        template:
          apiVersion: v1
          kind: Namespace
          metadata:
            name: helloworld
            labels:
              app: helloworld
    - metadata:
        name: helloworld-deployment
        labels:
          app: helloworld
      spec:
        template:
          apiVersion: apps/v1
          kind: Deployment
          metadata:
            name: helloworld-deployment
            namespace: helloworld
          spec:
            selector:
              matchLabels:
                app: helloworld
            replicas: 1
            template:
              metadata:
                labels:
                  app: helloworld
              spec:
                containers:
                - name: hello-world
                  image: gcr.io/google-samples/node-hello:1.0
                  ports:
                  - containerPort: 8080
                    protocol: TCP
    - metadata:
        name: helloworld-service
        labels:
          app: helloworld
      spec:
        template:
          kind: Service
          metadata:
            name: helloworld-service
            namespace: helloworld
          spec:
            selector:
              app: helloworld
            ports:
            - port: 80
              targetPort: 8080
            type: LoadBalancer
```

pkg/controller/workload/kubernetes/resource/resource.go
```diff
@@ -559,5 +559,12 @@ func haveSameController(a, b metav1.Object) bool {
 		return false
 	}
 
-	return a.GetAnnotations()[RemoteControllerUID] == b.GetAnnotations()[RemoteControllerUID]
+	// NOTE(hasheddan): we set an annotation for the UID of the remote
```
hasheddan: @negz note that I am still checking in `hasController` that a remote UID annotation exists, it just does not have to match. Sound ok?

PR #1385, opened 2020-03-31
Set namespace on resources rendered from OAM containerizedworkload

hasheddan:

This follows #1361 and sets the `namespace` of the `Deployment` rendered by a `ContainerizedWorkload` to the `namespace` of the `ContainerizedWorkload`. The `Service` is set to the same `namespace` as the `Deployment` for which it is being injected. Note that these are the namespaces for the remote resources, not the `KubernetesApplicationResources` (although they will match).

Tested these changes successfully with https://github.com/crossplane/addon-oam-kubernetes-remote/pull/4 using the following configuration:

<details>
 <summary>App Config</summary>

```yaml
---
kind: TraitDefinition
metadata:
  name: manualscalertraits.core.oam.dev
spec:
  definitionRef:
    name: manualscalertraits.core.oam.dev
---
kind: WorkloadDefinition
metadata:
  name: containerizedworkloads.core.oam.dev
spec:
  definitionRef:
    name: containerizedworkloads.core.oam.dev
---
kind: Component
metadata:
  name: example-component
spec:
  workload:
    kind: ContainerizedWorkload
    spec:
      containers:
      - name: wordpress
        image: wordpress:4.6.1-apache
        # TODO(negz): Allow environment variables to be set from a secret? OAM
        # doesn't have a concept of Kubernetes secrets, so we will likely need
        # to abstract this somehow.
        ports:
        - containerPort: 80
          name: wordpress
  parameters:
  - name: instance-name
    required: true
    fieldPaths:
    - metadata.name
  - name: image
    fieldPaths:
    - spec.containers[0].image
---
kind: ApplicationConfiguration
metadata:
  name: example-appconfig
spec:
  components:
  - componentName: example-component
    parameterValues:
    - name: instance-name
      value: example-appconfig-workload
    - name: image
      value: wordpress:php7.2
    traits:
    - trait:
        kind: ManualScalerTrait
        metadata:
          # TODO(negz): This name can be omitted and generated automatically if
          # each trait kind may apply only once to a component/workload.
          name:  example-appconfig-trait
        spec:
          replicaCount: 3
```

</details>

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/pkg/oam/workload/containerized/containerized.go
+++ b/pkg/oam/workload/containerized/containerized.go
@@ -36,6 +36,8 @@ const (
 	errNotContainerizedWorkload = "object is not a containerized workload"
 )
 
+const defaultNamespace = "default"
+
 const labelKey = "containerizedworkload.oam.crossplane.io"
 
 var (
@@ -58,6 +60,11 @@ func Translator(ctx context.Context, w resource.Workload) ([]resource.Object, er
 		},
 		ObjectMeta: metav1.ObjectMeta{
 			Name: cw.GetName(),
+			// NOTE(hasheddan): we always create the Deployment in the default
+			// namespace because there is not currently a namespace scheduling
+			// mechanism in the Crossplane OAM implementation. It is likely that
+			// this will be addressed in the future by adding a Scope.
+			Namespace: defaultNamespace,
 		},
 		Spec: appsv1.DeploymentSpec{
 			Selector: &metav1.LabelSelector{
--- a/pkg/oam/workload/containerized/containerized_test.go
+++ b/pkg/oam/workload/containerized/containerized_test.go
@@ -65,7 +65,8 @@ func deployment(mod ...deploymentModifier) *appsv1.Deployment {
 			APIVersion: deploymentAPIVersion,
 		},
 		ObjectMeta: metav1.ObjectMeta{
-			Name: cwName,
+			Name:      cwName,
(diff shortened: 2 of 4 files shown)


negz:
@hasheddan Does this imply that when an `ApplicationConfiguration` deployed to namespace `foo` in cluster A, namespace `foo` must also exist in cluster B?

hasheddan:
@negz yes that is correct. Before, we were just creating things in default namespace, which we could hardcode now if we wanted. It seems like a better idea to give some optionality about the namespace though. However, we can always guarantee that default namespace will exist. What do you think?

negz:
@hasheddan Yeah, I like the safety of just deploying to `default`. I wonder if we could somehow influence the namespace in future using the hypothetical scheduling scope?


```yaml
apiVersion: oam.crossplane.io/v1alpha1
kind: SchedulingScope
metadata:
  name: coolcluster
spec:
  clusterSelector:
    matchLabels:
      cool: "true"
  namespace: coolest
```

hasheddan:
I was thinking we likely would but guessing we won’t be doing that immediately, I’ll update to schedule to default for now 👍

hasheddan:
@negz updated to set `Deployment` namespace to default, and `Service` namespace to that of the `Deployment`. Also added note for potential future use of `Scope`. :+1:

PR #1388, opened 2020-04-01
Remove name from bucket spec

muvaf:

Remove name from spec as external name annotation is used for that purpose now. Updates are being done on provider repositories as part of backup/restore compatibility effort.

Unit tests.

### Checklist

I have:
  appropriate.

[definition of done].


muvaf:
@negz it seems like mostly gitlab, which I'm not sure whether it works anymore, but I updated all. I think we can wait until all three providers merge PRs that do not use this field.

hasheddan:
@negz just noticed this has not been merged, and we will probably want it to be for backup and restore purposes :+1:

PR #1400, opened 2020-04-07
stacks: controller ownerrefs require finalizer subresource rbac in openshift

displague:
For the stack controller to set a controller owner reference on CRs
that it owns, in some settings (OpenShift 4.3), it is necessary to
give the controller access to a finalizers subresource, even if the
crd does not present this subresource. External controllers will look
for this rule. The error produced without it is: "cannot set
blockOwnerDeletion if an ownerReference refers to a resource you
can't set finalizers on"

Similar to https://github.com/openshift/origin/issues/22958

Give Stack controllers RBAC rules that include full access to the `/finalizers` sub-resource for all resources that the Stack owns.

Fixes https://github.com/crossplane/stack-aws-sample/issues/17

`make test` has been updated with these expectations.

While the code has not been run in Openshift, the change in rules has been tested in OpenShift 4.3.

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/pkg/stacks/unpack.go
+++ b/pkg/stacks/unpack.go
@@ -486,6 +486,16 @@ func (sp *StackPackage) applyRules() (v1alpha1.PermissionsSpec, error) {
 				kinds = append(kinds, crd.Spec.Names.Plural+"/scale")
 			}
 		}
+
+		// For the stack controller to set a controller owner reference on CRs
+		// that it owns, in some settings (OpenShift 4.3), it is necessary to
+		// give the controller access to a finalizers subresource, even if the
+		// crd does not present this subresource. External controllers will look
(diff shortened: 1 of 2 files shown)


displague:
> Has this also been tested in upstream vanilla kubernetes? I am expecting we don't regress there with this fix to be compatible with OpenShift.

Good thinking, @jbw976. I have used arbitrary groups, verbs, and resources in test resources before so I didn't expect any fallback from this.  I have run this branch in kind (vanilla) and the created Stack was installed with the `/finalizers` clusterrole rule and attained a `Ready:True` state before being successfully deleted (with all the sub-resources resources that the finalizer handles also being deleted).

Issue #247, opened 2018-12-08
Skip CI not working correctly


bassam:
For some reason jenkins is not skipping correctly when `[skip ci]` is added.

PR #1406, opened 2020-04-08
stacks: permit packages to replace CRDS under proper conditions

displague:
# Description of your changes

Addresses #247

Currently, the StackManager creates CRDs without concern for existing CRDs. When creating a CRD resource when a CRD by the same name already exists, the existing CRD is left as-is and the stackinstall will (barring other problems) produce a stack resource. The stack controller will add labels to that CRD when reconciling the stack resource, reflecting a shared parentage of this CRD.

Problems with ignoring existing CRDs are represented in #247.

This PR blocks a `StackInstall` / `ClusterStackInstall`, mid-install, while:
1. the stack is attempting to install a CRD that already exists but is marked for deletion
2. the stack is attempting to install a CRD that does not include all of the versions of an existing CRD

If an existing CRD is present, and is not blocked above, the existing CRD will be updated to the new CRD:
* preserving the labels of the existing CRD
* preserving the existing CRD annotations
* preserving existing versions (by requiring the new CRD to include existing versions)
* storageVersion from the new CRD will be applied

`make test` (with new tests)

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/pkg/controller/stacks/install/installjob.go
+++ b/pkg/controller/stacks/install/installjob.go
@@ -28,10 +28,12 @@ import (
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
+	apiextensions "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/yaml"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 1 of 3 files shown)


pkg/controller/stacks/install/installjob_test.go
```diff
@@ -1486,21 +1490,100 @@ func TestCreateJobOutputObject(t *testing.T) {
...
+			g := NewGomegaWithT(t)
+
 			gotErr := tt.jobCompleter.createJobOutputObject(ctx, tt.obj, tt.stackInstaller, tt.job)
 
 			if diff := cmp.Diff(tt.want.err, gotErr, test.EquateErrors()); diff != "" {
 				t.Errorf("createJobOutputObject(): -want error, +got error:\n%s", diff)
 			}
 
-			if diff := cmp.Diff(tt.want.obj, tt.obj, test.EquateConditions()); diff != "" {
-				t.Errorf("createJobOutputObject(): -want obj, +got obj:\n%v", diff)
```
displague, line 1502: This test did not verify that the object that went into `createJobOutputObject` was the object that came out of `createJobOutputObject`.

The existing tests were told to expect that what went into the function is what went into the function (which would verify no mutations).

The change in testing here fetches the created resource and compares the expected object against that, which is more meaningful imho.

pkg/controller/stacks/install/installjob.go
```diff
@@ -279,13 +281,78 @@ func (jc *stackInstallJobCompleter) createJobOutputObject(ctx context.Context, o
...
+		return errors.Wrapf(err, "failed to convert unstructured crd from job output")
+	}
+
+	if !crdIsVersionsInclusive(existing, crd) {
+		return errors.New("replacement crd removes crd versions")
+	}
+
+	// TODO(displague) REVIEWER QUESTION:
+	// capture unpacked crd labels/annotations and give them
+	// highter priority than existing labels/annotations?
```
displague, line 328: The unpacking process does not add any labels to CRDs, currently.  However, a CRD may have included labels within the package.

When we want to replace an existing CRD with one from a package (because it provides all of the old CRD versions and then some), we want to persist the existing labels (which denote the various stacks that share the responsibility of providing this CRD).

Annotations on the other hand, represent the UI metadata contained in the package.

The behavior of this PR is:
1. use the unpack annotations
2. merge any existing annotations from the existing CRD

It may be better (necessary?) to:
1. use the unpack annotations
2. merge any non-conflicting annotations from the existing CRD

Does it make more sense to:
1. use the unpack annotations
2. capture existing CRD annotations
3. remove from the existing CRD any annotations know to crossplane/stackmanager
2. merge the existing CRD annotations onto the new CRD annotations

PR #1407, opened 2020-04-08
stacks: remove parent-uid from parent labeling

### LINKED ISSUE #1389: Allow Stack, StackInstall, StackDefinition to be backed up and restored

Client-side backup and restore tools for Kubernetes, such as a Velero, typically recreate resources on restore.  The UID of these resources will differ from the original because UID is ignored on resource creation.

The StackManager and the resources it creates are not prepared to handle a change of UIDs in the resources that the StackManager creates and otherwise manages.

A mismatched UID could result in garbage collection by Kubernetes, or deletion (or other forms of disconnect) by the Stack Manager.

The stack manager moved away from owner references (which demand UID) for many of the object references, but UIDs are used in some of the label based approaches.  In some of the places where a OwnerReference (with UID) is used, we'll have to decide what action to take. The loss of an install job should not be a problem if the stack is fully installed.  We may want to clean those jobs up sooner ourselves.  The loss of a Deployment should be auto-recovered as an 'update' reconcile operation (but isn't), so we should make the stack-manager capable of handling that situation (and ignoring the UID for comparisons - possibly throwing the UID out or keeping the UID updated).

Since 0.8, CRDs installed from the stack manager should be immune to this, but restored Stack resources may be orphaned when their StackInstall is removed (because a UID in the parent labels is used for matching).

The ClusterRoles, ServiceAccounts, Roles, ClusterRoleBindings, and RoleBindings may also be lost on restore (check these).￼

Related to #1042, #1151

displague:

Fixes #1389

### Checklist

I have:
  appropriate.

[definition of done].


displague:
Closing this, pending further investigation.

PR #1408, opened 2020-04-09
stacks: truncated resource names can not contain ".-"

displague:
# Description of your changes

As discovered by @rathpc :

> When long names get truncated we are not explicitly checking what the last character is resulting in a possible issue as shown in the code block below. A . is next to a - which is considered INVALID for a job (possible other things)

This PR offers a fix by reducing the number of characters of the overall truncation by right trimming any `.` characters before the hyphenated suffix.

The DNS label (regex) restriction that triggers the error is not a name restriction in all of the places where we are using `Truncate`, but I think it is reasonable to make this change in all truncations.

`make test` (new test)

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/pkg/stacks/truncate/truncate.go
+++ b/pkg/stacks/truncate/truncate.go
@@ -56,8 +56,11 @@ const (
 // suffixLength, or truncation length is greater than sha1Length, or the suffix
 // length is less than 2.
 //
-// Example:
-// If the base32 sum of a digest of "aaaaaaaaaaa" is "ovo", with a suffix length of 4:
+// Any final "." characters in the truncated text will be removed to satisfy
+// DNS-1123 labeling.
+//
+// Example: If the base32 sum of a digest of "aaaaaaaaaaa" is "ovo", with a
+// suffix length of 4:
 //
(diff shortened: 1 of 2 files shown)


pkg/stacks/truncate/truncate.go
```diff
@@ -84,14 +87,15 @@ func Truncate(str string, length, suffixLength int) (string, error) {
 	// See the import comment regarding use of sha1
 	// nolint:gosec
 	checksum := sha1.Sum([]byte(str))
-	retained := length - suffixLength
+	retainedLength := length - suffixLength
+	retained := strings.TrimRight(str[0:retainedLength], ".")
```
jbw976, line 91: are there other chars we need to worry about? `-` or others?

rathpc: @jbw976 
This is the regex that the specific issue is colliding against:

`[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*`

Based on that it seems that as long as periods are immediately followed by any alphanumeric it is alright, it was the lack of an alphanumeric prior to the dash causing the reported issue

PR #1409, opened 2020-04-09
Support backing up and restoring Kubernetes targets

negz:

This switches automatically generated targets from using the UID of their claim to the name of their claim. We use the new crossplane-runtime applicator logic that will adopt uncontrolled targets and refuse to adopt controlled targets.

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/pkg/controller/workload/kubernetes/target/target_test.go
+++ b/pkg/controller/workload/kubernetes/target/target_test.go
@@ -33,10 +33,12 @@ import (
 
 	runtimev1alpha1 "github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
 	"github.com/crossplane/crossplane-runtime/pkg/logging"
+	"github.com/crossplane/crossplane-runtime/pkg/meta"
 	"github.com/crossplane/crossplane-runtime/pkg/resource"
 	"github.com/crossplane/crossplane-runtime/pkg/test"
 
 	computev1alpha1 "github.com/crossplane/crossplane/apis/compute/v1alpha1"
+	workloadv1alpha1 "github.com/crossplane/crossplane/apis/workload/v1alpha1"
 )
 
 const (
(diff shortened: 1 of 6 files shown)


negz:
Tested by deploying `app-wordpress` with `stack-azure-sample` and validating that a target appears:

```
$ kubectl get kubernetestarget
NAME             CLUSTER
testme-cluster   
messremb(negz@crossplane)$ kubectl describe kubernetestarget
Name:         testme-cluster
Namespace:    default
Labels:       app=testme
              core.crossplane.io/parent-group=wordpress.apps.crossplane.io
              core.crossplane.io/parent-kind=WordpressInstance
              core.crossplane.io/parent-name=testme
              core.crossplane.io/parent-namespace=default
              core.crossplane.io/parent-uid=cfceabab-826f-4d71-8826-49b5f229986e
              core.crossplane.io/parent-version=v1alpha1
Annotations:  <none>
API Version:  workload.crossplane.io/v1alpha1
Kind:         KubernetesTarget
Metadata:
  Creation Timestamp:  2020-04-09T06:10:24Z
  Generation:          1
  Owner References:
    API Version:     compute.crossplane.io/v1alpha1
    Controller:      true
    Kind:            KubernetesCluster
    Name:            testme-cluster
    UID:             fc6c1202-d053-4cce-8025-16ff4abda132
  Resource Version:  2084
  Self Link:         /apis/workload.crossplane.io/v1alpha1/namespaces/default/kubernetestargets/testme-cluster
  UID:               a1b6a265-712f-4cb8-857a-84f2a1aa7b66
Spec:
  Connection Secret Ref:
    Name:  testme-cluster
Status:
Events:  <none>
```

pkg/controller/workload/kubernetes/target/target_test.go
```diff
@@ -232,165 +191,18 @@ func TestReconcile(t *testing.T) {
...
-								},
-							},
-						),
-							withConnectionSecretRef(&runtimev1alpha1.LocalSecretReference{Name: "super-secret"}),
-							withTargetLabels(map[string]string{"dev": "true"}))
-						if diff := cmp.Diff(want, got); diff != "" {
-							t.Errorf("-want, +got:\n%s", diff)
-						}
+					Applicator: resource.ApplyFn(func(_ context.Context, _ runtime.Object, _ ...resource.ApplyOption) error {
 						return nil
```
hasheddan, line 206: We could check to make sure that the `KubernetesTarget` actually points to the right secret and has the correct labels here. Otherwise, there isn't much of a point in setting the labels, etc. that we want propagated on the `KubernetesCluster` we return from the `MockGet`. Happy to defer this to a follow-up though 👍

hasheddan:
@negz One thing I want to point out here is that if someone manually creates a `KubernetesTarget` in a namespace, it will have no owner references. This means that if a `KubernetesCluster` claim with the same name is created in the same namespace, then I believe it will take control of the uncontrolled manually created `KubernetesTarget` and change its the `Secret` it points to. This is not necessarily wrong, but you can imagine how it could be harmful for a `KubernetesApplication` that is scheduled to target to then suddenly connect to a different cluster and create the in-inline resources there. I think this can just be a noted behavior for the time being, and we can potentially establish a safeguard later down the line. What do you think?

pkg/controller/workload/kubernetes/target/target_test.go
negz: I should be able to address this now.

negz:
> if someone manually creates a KubernetesTarget in a namespace, it will have no owner references.

@hasheddan Good point - I'll add a label to autogenerated targets, and require that label to exist before we adopt them.

PR #1410, opened 2020-04-09
Support backing up and restoring ApplicationConfigurations

negz:

When an ApplicationConfiguration, its workloads, and its traits are backed up and subsequently restored using Velero the resources are allocated new UIDs. Owner references and resource status are also removed during the restore. This commit updates the ApplicationConfiguration controller to avoid reliance on UIDs for garbage collection. The controller will also 'adopt' any existing resources that it would otherwise create, as long as they're not controlled by another controller (i.e. as long as they have no controller reference).

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/pkg/controller/workload/kubernetes/resource/resource.go
+++ b/pkg/controller/workload/kubernetes/resource/resource.go
@@ -267,7 +267,8 @@ func (c *unstructuredClient) sync(ctx context.Context, template *unstructured.Un
 
 	rs := getRemoteStatus(remote)
 
-	return rs, errors.Wrap(resource.Apply(ctx, c.kube, template), errSyncResource)
+	// TODO(negz): Update to resource.APIPatchingApplicator.
+	return rs, errors.Wrap(resource.Apply(ctx, c.kube, template), errSyncResource) // nolint:staticcheck
 }
 
 func getRemoteStatus(u runtime.Unstructured) *v1alpha1.RemoteStatus {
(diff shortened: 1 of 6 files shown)


negz:
Leaving this in draft until I'm able to confirm that it works as expected. @hasheddan I think we'll also need to update the workload and trait reconcilers in crossplane-runtime.

pkg/controller/workload/kubernetes/resource/resource.go
```diff
@@ -267,7 +267,8 @@ func (c *unstructuredClient) sync(ctx context.Context, template *unstructured.Un
 
 	rs := getRemoteStatus(remote)
 
-	return rs, errors.Wrap(resource.Apply(ctx, c.kube, template), errSyncResource)
+	// TODO(negz): Update to resource.APIPatchingApplicator.
+	return rs, errors.Wrap(resource.Apply(ctx, c.kube, template), errSyncResource) // nolint:staticcheck
```
hasheddan, line 271: Just a heads up I believe this change is also made in https://github.com/crossplane/crossplane/pull/1409 but shouldn't matter which comes through first 👍

negz:
@hasheddan Just tested this out by adding and removing components and traits to an appconfig to ensure owner references were set and removed workloads and traits are garbage collected. I then backed up and restored everything in default namespace (i.e. the appconfig, components, workloads, and traits), then repeated the test to ensure garbage collection still worked and refs were set.

PR #1425, opened 2020-04-15
Do not sync KARs if KA is deleted

hasheddan:

In the case that a KubernetesApplication has been deleted, but is blocked on a subset of its KubernetesApplicationResources existing with controllerRefs to it, we do not want to attempt to create or update any KARs that have been deleted or are in the process of being deleted.

Potential fix for https://github.com/crossplane/app-wordpress/issues/48, which I have been unable to reproduce.

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/pkg/controller/workload/kubernetes/application/application.go
+++ b/pkg/controller/workload/kubernetes/application/application.go
@@ -98,15 +98,24 @@ type localCluster struct {
 }
 
 func (c *localCluster) sync(ctx context.Context, app *v1alpha1.KubernetesApplication) (v1alpha1.KubernetesApplicationState, error) {
-	app.Status.DesiredResources = len(app.Spec.ResourceTemplates)
-	app.Status.SubmittedResources = 0
-
 	var errs []error
+
+	// If App was deleted, do not attempt to sync. The KubernetesApplication is
+	// blocked on deletion of all KubernetesApplicationResources that have
+	// controller references to it. If we attempt to create or update while a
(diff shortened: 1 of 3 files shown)


pkg/controller/workload/kubernetes/application/application.go
```diff
@@ -98,16 +98,25 @@ type localCluster struct {
...
 func (c *localCluster) sync(ctx context.Context, app *v1alpha1.KubernetesApplication) (v1alpha1.KubernetesApplicationState, error) {
-	app.Status.DesiredResources = len(app.Spec.ResourceTemplates)
-	app.Status.SubmittedResources = 0
-
 	var errs []error
 
 	// Garbage collect any resource we control but no longer have templates for.
 	if err := c.gc.process(ctx, app); err != nil {
 		return v1alpha1.KubernetesApplicationStateFailed, errors.Wrap(err, errGarbageCollect)
 	}
```
negz, line 106: Should we return before running GC if we're being deleted? Kubernetes GC should take care of all KARs we made in this case.

hasheddan: Yes! Since to be garbage collected a KAR must be owned by the KA, and if it is owned by the KA then it will be cleaned up when the KA is deleted and all finalizers are removed from the KAR. Thanks for catching that, will update :+1:

hasheddan:
@negz fixed :+1:

PR #1433, opened 2020-04-21
Re-enable updating CRDs on new StackInstalls

hasheddan:

Previously, the stack controller was updated to include the ability to install multiple stacks that depend on the same CRD, and update it to the newly installed version. This was broken when CRDs were changed to be updated without directly modifying the existing CRD (#1406). This re-enables the intended functionality, but is serving as a stop-gap fix in a larger refactoring effort.

Created an `app-wordpress` install in `crossplane-system`:
```
kubectl apply -f wordpress.yaml -n crossplane-system
```

Controller started successfully and observed labels on `WordpressInstance` CRD:
```
🤖 (crossplane) kubectl get customresourcedefinitions wordpressinstances.wordpress.apps.crossplane.io -o=jsonpath={.metadata.labels}

map[app.kubernetes.io/managed-by:stack-manager crossplane.io/scope:namespace namespace.crossplane.io/crossplane-system:true parent.stacks.crossplane.io/crossplane-system-wp-sample:true]
```

Created an `app-wordpress` install in `default`:
```
kubectl apply -f wordpress.yaml -n default
```

Controller started successfully and observed labels on `WordpressInstance` CRD:
```
🤖 (crossplane) kubectl get customresourcedefinitions wordpressinstances.wordpress.apps.crossplane.io -o=jsonpath={.metadata.labels}

map[app.kubernetes.io/managed-by:stack-manager crossplane.io/scope:namespace namespace.crossplane.io/crossplane-system:true namespace.crossplane.io/default:true parent.stacks.crossplane.io/crossplane-system-wp-sample:true parent.stacks.crossplane.io/default-wp-sample:true]
```

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/pkg/controller/stacks/install/installjob.go
+++ b/pkg/controller/stacks/install/installjob.go
@@ -39,6 +39,7 @@ import (
 
 	"github.com/crossplane/crossplane-runtime/pkg/logging"
 	"github.com/crossplane/crossplane-runtime/pkg/meta"
+	"github.com/crossplane/crossplane-runtime/pkg/resource"
 	"github.com/crossplane/crossplane/apis/stacks/v1alpha1"
 	"github.com/crossplane/crossplane/pkg/stacks"
 )
@@ -328,7 +329,7 @@ func (jc *stackInstallJobCompleter) replaceCRD(ctx context.Context, obj *unstruc
 	meta.AddLabels(obj, existing.GetLabels())
 	meta.AddAnnotations(obj, existing.GetAnnotations())
 
-	return jc.client.Update(ctx, obj)
+	return resource.NewAPIPatchingApplicator(jc.client).Apply(ctx, obj)
 }
 
 // TODO(displague) this is copied from stacks. centralize.
--- a/pkg/controller/stacks/install/installjob_test.go
+++ b/pkg/controller/stacks/install/installjob_test.go
@@ -66,6 +66,9 @@ metadata:
   name: ` + crdName + `
 spec:
   group: samples.upbound.io
+  conversion:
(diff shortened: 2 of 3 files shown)


pkg/controller/stacks/install/installjob.go
```diff
@@ -328,7 +329,7 @@ func (jc *stackInstallJobCompleter) replaceCRD(ctx context.Context, obj *unstruc
 	meta.AddLabels(obj, existing.GetLabels())
 	meta.AddAnnotations(obj, existing.GetAnnotations())
 
-	return jc.client.Update(ctx, obj)
```
jbw976, line 331: Can you help me understand why the `Update` call was insufficient in this scenario?

hasheddan: @jbw976 sure! The issue here is that as you can see we above, we are not getting the object and then modifying it, we are getting the object, then trying to update it with the CRD that is supplied by the new stack. This means that we are not including a resource version on the object supplied to the update, so we will always fail because the api server can not determine if the object has been modified between when we got it and when we issued the update. We can actually simplify some of this logic a bit (for example, passing the crdVersionsInclusive function as an apply option), but I opted for the smallest fix with the least churn as we determine as path forward for some refactor here :)

jbw976: ah yes, that makes sense because the incoming `obj` here is output from the unpacking job.  It was not retrieved form the api server.  Thank you for taking the time to explain this to me @hasheddan 🙇‍♂️

PR #1436, opened 2020-04-22
Import oam-kubernetes-runtime for controller and type implementations

hasheddan:

This updates core crossplane to use the types and controllers that reside in `oam-kubernetes-runtime`. This was a necessary change as core crossplane was unable to update its `crossplane-runtime` dependency due to the removal of OAM functionality there, and was encountering a circular dependency when attempting to import `oam-kubernetes-runtime`.

~Note: this must wait for https://github.com/crossplane/oam-kubernetes-runtime/pull/5~

`make build`

No functional changes here, just moving of code.

### Checklist

I have:
  appropriate.

[definition of done].


hasheddan:
@negz I think this is ready for review :+1:

cmd/crossplane/core/core.go
```diff
@@ -60,6 +62,10 @@ func (c *Command) Run(log logging.Logger) error {
 		return errors.Wrap(err, "Cannot add core Crossplane APIs to scheme")
 	}
 
+	if err := oamapis.AddToScheme(mgr.GetScheme()); err != nil {
+		return errors.Wrap(err, "Cannot add core OAM APIs to scheme")
+	}
+
 	if err := oam.Setup(mgr, log); err != nil {
```
negz, line 69: Should we just call the oam-kubernetes-runtime setup function and get rid of our own?

cluster/charts/crossplane-types/crds/core.oam.dev_applicationconfigurations.yaml
```diff
@@ -1,242 +0,0 @@
-
```
negz, line 1: Hrm, how are we going to get these OAM CRDs installed?

hasheddan: hmm we could create a helm chart for `oam-kubernetes-runtime` and make it a dependency

hasheddan: For posterity: we now run `go mod vendor` before generating our CRDs and point generation to the vendored `oam-kubernetes-runtime` directory :+1:

hasheddan:
@negz I think this is ready now :+1:

hasheddan:
It is concerning to me that integration testing passed before these CRDs were added back and installed :thinking:. I would think the controller manager would have panicked without the CRDs that were registered in its scheme being present... I'll look into if these integration tests are still giving appropriate signal https://jenkinsci.upbound.io/blue/organizations/jenkins/crossplane%2Fcrossplane%2Fbuild/detail/PR-1436/5/pipeline/48

wonderflow:
cool

PR #1440, opened 2020-04-23
Add an InfrastructurePublication type and controller

negz:

Leaving this in draft for now, given there's no actual InfrastructurePublication controller. :) I refactored the InfrastructureDefinition controller and CRD generation in the hope that I can mostly reuse them for the InfrastructurePublication controller.

### Checklist

I have:
  appropriate.

[definition of done].


apis/apiextensions/v1alpha1/ccrd/schemas.go
```diff
@@ -0,0 +1,186 @@
...
+  properties:
+    name:
+      type: string
+  required:
+  - name
+  - namespace
+  type: object
+writeConnectionSecretToRef:
+  properties:
+    name:
```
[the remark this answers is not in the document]
negz: Not for the requirement, which always uses the namespace of the requirement.

apis/apiextensions/v1alpha1/crd_types.go
```diff
@@ -1,158 +0,0 @@
-/*
```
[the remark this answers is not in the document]
negz: Yeah, sorry about that. :( Usually git figures this out magically, but I guess it doesn't when the diff gets too big.

negz:
Tested with these manifests:

```yaml
---
kind: InfrastructurePublication
metadata:
  name: mysqls.database.crossplane.io
spec:
  infrastructureDefinitionRef:
    name: mysqls.database.crossplane.io
---
kind: InfrastructureDefinition
metadata:
  name: mysqls.database.crossplane.io
spec:
  connectionSecretKeys:
    - username
    - password
    - endpoint
  crdSpecTemplate:
    group: database.crossplane.io
    version: v1alpha1
    names:
      kind: MySQL
      listKind: MySQLList
      plural: mysqls
      singular: mysql
    validation:
      openAPIV3Schema:
        type: object
        properties:
          spec:
            type: object
            properties:
              engineVersion:
                description: EngineVersion specifies the desired MySQL engine version,
                  e.g. 5.7.
                enum:
                  - "5.6"
                  - "5.7"
                type: string
              storageGB:
                type: integer
---
kind: Composition
metadata:
  name: mysqlcomposition
  labels:
    connectivity: private
spec:
  reclaimPolicy: Delete
  from:
    kind: MySQL
  to:
    - base:
        apiVersion: azure.crossplane.io/v1alpha3
        kind: ResourceGroup
        spec:
          location: West US 2
          reclaimPolicy: Delete
          providerRef:
            name: example
    - base:
        apiVersion: database.azure.crossplane.io/v1beta1
        kind: MySQLServer
        spec:
          forProvider:
            administratorLogin: myadmin
            resourceGroupNameSelector:
              matchControllerRef: true
            location: West US 2
            sslEnforcement: Disabled
            version: "5.6"
            sku:
              # Basic tier does not support VNet connection rule.
              tier: GeneralPurpose
              capacity: 2
              family: Gen5
            storageProfile:
              storageMB: 8192
              backupRetentionDays: 7
              geoRedundantBackup: Disabled
          writeConnectionSecretToRef:
            namespace: crossplane-system
          providerRef:
            name: example
          reclaimPolicy: Delete
      patches:
      - fromFieldPath: "spec.engineVersion"
        toFieldPath: "spec.forProvider.version"
      - fromFieldPath: "spec.storageGB"
        toFieldPath: "spec.forProvider.storageMB"
        transforms:
          - type: math
            math:
              multiply: 1024
      connectionDetails:
    - base:
        apiVersion: database.azure.crossplane.io/v1alpha3
        kind: MySQLServerVirtualNetworkRule
        spec:
          name: virtualnetworkrule
          serverNameSelector:
            matchControllerRef: true
          resourceGroupNameSelector:
            matchControllerRef: true
          properties:
            virtualNetworkSubnetIdRef:
              name: subnet
          reclaimPolicy: Delete
          providerRef:
            name: example
      patches:
      - fromFieldPath: "metadata.name"
        toFieldPath: "spec.name"
```

```yaml
---
kind: MySQLRequirement
metadata:
  namespace: default
  name: example-sql
spec:
  engineVersion: "5.7"
  storageGB: 2
  compositionSelector:
    matchLabels:
      connectivity: private
  writeConnectionSecretToRef:
    name: example-sql
```

apis/apiextensions/v1alpha1/infrastructurepublication_types.go
```diff
@@ -0,0 +1,61 @@
...
+
+// +kubebuilder:object:root=true
+
+// An InfrastructurePublication publishes a defined kind of composite
+// infrastructure resource. Published infrastructure resources may be bound to
+// an application via an infrastructure requirement.
+// +kubebuilder:resource:categories={crossplane}
+// +kubebuilder:printcolumn:name="AGE",type="date",JSONPath=".metadata.creationTimestamp"
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope=Cluster
```
hasheddan, line 45: For cluster-scoped resources the category will not be added to the generated CRD unless it is listed in-line the the `resource:scope` declaration. Could you also update the `Composition` and `InfrastructureDefinition` types to this pattern as well (I didn't catch those in earlier PR). Example: https://github.com/crossplane/crossplane/blob/134e1d8190ab0a82aaabf4a7f37a25a9cbbc78c4/apis/kubernetes/v1alpha1/types.go#L40

pkg/controller/apiextensions/requirement/reconciler.go
```diff
@@ -0,0 +1,453 @@
...
+	// NOTE(negz): This method is well over our cyclomatic complexity goal.
+	// Be wary of adding additional complexity.
+
+	log := r.log.WithValues("request", req)
+	log.Debug("Reconciling")
+
+	ctx, cancel := context.WithTimeout(context.Background(), requirementReconcileTimeout)
+	defer cancel()
+
+	requirement := r.newRequirement()
```
muvaf, line 276: Collision with package name `requirement`

pkg/controller/apiextensions/requirement/reconciler.go
```diff
@@ -0,0 +1,453 @@
...
+	}
+
+	record := r.record.WithAnnotations("external-name", meta.GetExternalName(requirement))
+	log = log.WithValues(
+		"uid", requirement.GetUID(),
+		"version", requirement.GetResourceVersion(),
+		"external-name", meta.GetExternalName(requirement),
+	)
+
+	composite := r.newComposite()
```
muvaf, line 291: Collision with package name `composite`

pkg/controller/apiextensions/requirement/reconciler.go
```diff
@@ -0,0 +1,453 @@
...
+		record.Event(requirement, event.Normal(reasonWaitingToBind, "Composite resource is not yet ready"))
+
+		// TODO(negz): Watch the kind of composite resource we're concerned with
+		// so that we don't need to requeue here.
+		requirement.SetConditions(Waiting(), v1alpha1.ReconcileSuccess())
+		return reconcile.Result{RequeueAfter: aShortWait}, errors.Wrap(r.client.Status().Update(ctx, requirement), errUpdateRequirementStatus)
+	}
+
+	if composite.GetRequirementReference() == nil {
+		if err := r.composite.PropagateConnection(ctx, requirement, composite); err != nil {
```
muvaf, line 406: I think propagation should happen only after it successfully binds to the composite.

pkg/controller/apiextensions/requirement/api.go
```diff
@@ -0,0 +1,120 @@
...
+// NewAPIBinder returns a new APIBinder.
+func NewAPIBinder(c client.Client, t runtime.ObjectTyper) *APIBinder {
+	return &APIBinder{client: c, typer: t}
+}
+
+// Bind the supplied requirement to the supplied composite.
+func (a *APIBinder) Bind(ctx context.Context, rq resource.Requirement, cp resource.Composite) error {
+
+	cp.SetRequirementReference(meta.ReferenceTo(rq, resource.MustGetKind(rq, a.typer)))
+	if err := a.client.Update(ctx, cp); err != nil {
```
muvaf, line 88: How about using `Apply` to patch?

pkg/controller/apiextensions/requirement/api.go
```diff
@@ -0,0 +1,120 @@
...
+		return errors.Wrap(err, errUpdateComposite)
+	}
+
+	if meta.GetExternalName(cp) == "" {
+		return nil
+	}
+
+	// Propagate back the final name of the composite resource to the requirement.
+	meta.SetExternalName(rq, meta.GetExternalName(cp))
+	return errors.Wrap(a.client.Update(ctx, rq), errUpdateRequirement)
```
muvaf, line 98: How about using `Apply` to patch?

pkg/controller/apiextensions/requirement/api.go
```diff
@@ -0,0 +1,120 @@
...
+	client client.Client
+	typer  runtime.ObjectTyper
+}
+
+// NewAPICompositeCreator returns a new APICompositeCreator.
+func NewAPICompositeCreator(c client.Client, t runtime.ObjectTyper) *APICompositeCreator {
+	return &APICompositeCreator{client: c, typer: t}
+}
+
+// TODO(negz): We should render and patch a composite resource on each
```
muvaf, line 51: Do you plan to address this in this PR?

pkg/controller/apiextensions/requirement/configurator.go
```diff
@@ -0,0 +1,66 @@
...
+	if !ok {
+		return nil
+	}
+
+	// YOLO. This seems to work; it copies over some unknown fields which are
+	// dropped and not stored due to the CRD being configured to do so.
+	spec, _ := fieldpath.Pave(urq.Object).GetValue("spec")
+	_ = fieldpath.Pave(ucp.Object).SetValue("spec", spec)
+
+	// TODO(negz): Set reclaim policy and connection secret somehow? Mostly the
```
muvaf, line 55: Do you plan to address this in this PR?

pkg/controller/apiextensions/requirement/reconciler.go
```diff
@@ -0,0 +1,453 @@
...
+		log.Debug("Composite resource is not yet ready")
+		record.Event(requirement, event.Normal(reasonWaitingToBind, "Composite resource is not yet ready"))
+
+		// TODO(negz): Watch the kind of composite resource we're concerned with
+		// so that we don't need to requeue here.
+		requirement.SetConditions(Waiting(), v1alpha1.ReconcileSuccess())
+		return reconcile.Result{RequeueAfter: aShortWait}, errors.Wrap(r.client.Status().Update(ctx, requirement), errUpdateRequirementStatus)
+	}
+
+	if composite.GetRequirementReference() == nil {
```
muvaf, line 405: It should continuously propagate the secret or there should be a connection propagator controller. Right now, composite doesn't expose `Ready` condition, hence, the first propagation is actually empty secret. But even if it did, credentials are rotated so it should be refreshed somehow.

pkg/controller/apiextensions/publication/reconciler.go
```diff
@@ -0,0 +1,329 @@
...
+		}
+
+		// NOTE(muvaf): When user deletes InfrastructureDefinition object the deletion
+		// signal does not cascade to the owned resource until owner is gone. But
+		// owner has its own finalizer that depends on having no instance of the CRD
+		// because it cannot go away before stopping the controller.
+		// So, we need to delete all instances of CRD manually here.
+		obj := &kunstructured.Unstructured{}
+		obj.SetGroupVersionKind(Published(d.GetDefinedGroupVersionKind()))
+		if err := r.client.DeleteAllOf(ctx, obj); err != nil && !kmeta.IsNoMatchError(err) && !kerrors.IsNotFound(err) {
```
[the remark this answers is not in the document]
muvaf: This needs to be fixed before merge. Although if you tested it will iterate in all namespaces without that option, it should be OK.

pkg/controller/apiextensions/requirement/configurator.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+	ucp, ok := cp.(*composite.Unstructured)
+	if !ok {
+		return nil
+	}
+
+	// YOLO. This seems to work; it copies over some unknown fields which are
+	// dropped and not stored due to the CRD being configured to do so.
+	spec, _ := fieldpath.Pave(urq.Object).GetValue("spec")
+	_ = fieldpath.Pave(ucp.Object).SetValue("spec", spec)
```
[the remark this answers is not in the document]
muvaf: I think this should be fixed, too, before merging even though it looks safe I don't feel comfortable with full copy. Should be trivial do `delete()` the keys from the map since we already know the keys that shouldn't make it to composite.

pkg/controller/apiextensions/requirement/reconciler.go
muvaf: In my test, the secret never got created. So, I think that needs a fix, too. Maybe blindly propagate the secret all the time for now?

pkg/controller/apiextensions/publication/reconciler.go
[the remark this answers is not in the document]
negz: Last time I tested this it appeared to delete CRs across all namespaces. This meshes with the code:

https://github.com/kubernetes-sigs/controller-runtime/blob/1c83ff/pkg/client/options.go#L504
https://github.com/kubernetes-sigs/controller-runtime/blob/1c83ff/pkg/client/options.go#L334

```go
// Namespace represents the namespace to list for, or empty for
// non-namespaced objects, or to list across all namespaces.
Namespace string
```

pkg/controller/apiextensions/requirement/configurator.go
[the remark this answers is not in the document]
negz: Honestly I don't love either approach. I worry about the maintainability of having to update a blacklist of fields that should be deleted from. The alternative would be to require that all user specified fields were nested in (for example) `spec.parameters` or require that all Crossplane specific fields were nested under something like `spec.requirement`. I think this might be worth considering in future because it clearly distinguishes user specified fields from injected, Crossplane specific fields.

apis/apiextensions/v1alpha1/ccrd/crd.go
```diff
@@ -0,0 +1,276 @@
...
+		for k, v := range *status {
+			crd.Spec.Validation.OpenAPIV3Schema.Properties["status"].Properties[k] = v
+		}
+		return nil
+	}
+}
+
+// PublishesCompositeInfrastructure adds the validation fields required by all
+// published infrastructure resources to a CustomResourceDefinition.
+func PublishesCompositeInfrastructure() Option {
```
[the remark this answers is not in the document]
negz: Yeah, I agree these should probably be merged. I think I originally thought they wouldn't always be used together, but I'm less confident of this now.

pkg/controller/apiextensions/composite/reconciler.go
```diff
@@ -93,8 +100,8 @@ type Resolver interface {
 	ResolveSelector(ctx context.Context, cr resource.Composite) error
 }
 
-// compositeReconciler reconciles the generic CRD that is generated via InfrastructureDefinition.
-type compositeReconciler struct {
+// Reconciler reconciles the generic CRD that is generated via InfrastructureDefinition.
+type Reconciler struct {
```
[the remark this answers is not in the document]
negz: https://github.com/golang/go/wiki/CodeReviewComments#interfaces
https://mycodesmells.com/post/accept-interfaces-return-struct-in-go

This was mostly a reflexive nitpick refactoring, but I think it's a good practice to "accept interfaces and return concrete types" in Go. The current implementation was returning an interface because the concrete type was not public, so the change was more about not returning an interface than it was about making the type public.

pkg/controller/apiextensions/requirement/api.go
negz: No. I'd like to address this in a future PR.

pkg/controller/apiextensions/requirement/configurator.go
negz: No. I'd like to address this in a future PR.

pkg/controller/apiextensions/requirement/api.go
negz: Typically we use Apply when we need to avoid resetting fields that we don't know about, right? Given this is all unstructured it's not obvious to me what benefit using `Apply` would get us.

pkg/controller/apiextensions/requirement/api.go
negz: Typically we use Apply when we need to avoid resetting fields that we don't know about, right? Given this is all unstructured it's not obvious to me what benefit using `Apply` would get us.

pkg/controller/apiextensions/publication/reconciler.go
```diff
@@ -0,0 +1,329 @@
...
+	// InfrastructurePublication, or just require the InfrastructurePublication
+	// to be named foorequirements.example.org rather than foos.example.org.
+	d := &v1alpha1.InfrastructureDefinition{}
+	if err := r.client.Get(ctx, req.NamespacedName, d); err != nil {
+		log.Debug(errGetInfraDef)
+		p.Status.SetConditions(runtimev1alpha1.ReconcileError(errors.New(errGetInfraDef)))
+		return reconcile.Result{RequeueAfter: shortWait}, errors.Wrap(r.client.Status().Update(ctx, p), errUpdateInfraPubStatus)
+	}
+
+	crd, err := ccrd.New(ccrd.PublishesInfrastructureDefinition(d, p), ccrd.PublishesCompositeInfrastructure())
```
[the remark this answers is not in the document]
negz: We'd still need to use some logic to infer our requirement name from the ref name though. i.e. If the ref name is `examples.example.org` we'll have to extract the fact that we want to make `kind: ExampleReference`, from the string `examples`. This is _probably_ doable for almost all cases, but I'm actually thinking we might want to ensure the the InfraDef always exists while the InfraPub exists.

I'd like the InfraPub to watch for the resource kind defined by the InfraDef so we can reconcile the published kind when the defined kind is changed, so we probably don't want to allow the defined kind to be deleted while the InfraPub controller is still running. One approach might be to have the InfraPub controller add a finalizer to the InfraDef.

pkg/controller/apiextensions/requirement/reconciler.go
negz: Confirmed I have a secret after switching to blind propagation:

```
$ kubectl get -o json secret example-sql|jq '.data'
{
  "endpoint": "REDACTED",
  "password": "REDACTED",
  "username": "REDACTED"
}
```

negz:
There are a handful of remaining known issues with this PR, and we're still lacking tests, but in the interest of allowing @muvaf and I to iterate on this code ASAP and in parallel I'm going to merge this and follow up on the outstanding issues in separate PRs.

PR #1443, opened 2020-04-24
Add CRD management permissions to Crossplane ClusterRole

muvaf:

Bad side effects of doing the tests in IDE using admin kubeconfig. I was able to see the error only when the controller actually started to query CRDs, so, we'd probably be able to catch this with an integration test that actually creates resources and check their conditions.

Manually.

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/cluster/charts/crossplane-types/templates/clusterrole.yaml
+++ b/cluster/charts/crossplane-types/templates/clusterrole.yaml
@@ -18,6 +18,12 @@ rules:
   - patch
   - delete
 - apiGroups:
+  - apiextensions.k8s.io
+  resources:
+  - customresourcedefinitions
+  verbs:
+  - "*"
+- apiGroups:
   - ""
   resources:
   - secrets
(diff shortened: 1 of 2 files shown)


cluster/charts/crossplane-types/templates/clusterrole.yaml
```diff
@@ -17,6 +17,12 @@ rules:
   - update
   - patch
   - delete
+- apiGroups:
+  - apiextensions.k8s.io
+  resources:
+  - customresourcedefinitions
+  verbs:
+  - "*"
```
jbw976, line 25: can you help me understand the full set of actions that we need to do on CRDs? the PR description said we need to query them (i.e. GET, LIST), but we also need to create them too when someone defines a new CRD via composition?

hasheddan: @jbw976 I can confirm `CREATE` permission is required: https://github.com/crossplane/crossplane/blob/d3e017b94fb24a139739eb71c5f89827d5e85e6a/pkg/controller/apiextensions/definition/reconciler.go#L265

muvaf: It's essentially all operations since we create/update/delete.

Issue #1441, opened 2020-04-23
install.yaml allows a full Deployment spec to be declared which can be problematic


jbw976:
### What problem are you facing?

Currently, stack packages allow the author to specify an entire `Deployment` within their `install.yaml` file.  This is helpful for flexibility for the author to customize and provide additional configuration for their stack controller, but presents some security and reliability concerns, first described by @turkenh here: https://github.com/crossplane/crossplane/issues/1317#issuecomment-593832706

Perhaps we can limit the amount of config that authors can provide for their controller. @lukeweber suggests:

>  seems like we shouldn't allow people to set more than image, args, env, readiness probe, liveness probe

negz:
This is no longer the case now that the OG package manager has been removed. We now have the opposite problem (temporarily), in that you can't. influence how a provider is run at all. :)

PR #1444, opened 2020-04-25
Restrict stacks deployment security context by default

hasheddan:

This updates the default behavior to restrict core API groups permissions on stacks and also makes it default to deny root security context on stacks deployments.

Ref: #1441 

Unit testing, manual tests below:

<details>
 <summary>Manual testing:</summary>

Created a KIND cluster, then started a tenant kubernetes api server and controller manager using [kindred](https://github.com/crdsdev/kindred).
```
🤖 (crossplane) kindred tenant create
🤖 (crossplane) kubectl get pods -A
NAMESPACE            NAME                                                         READY   STATUS      RESTARTS   AGE
kube-system          coredns-6955765f44-2v4cn                                     1/1     Running     0          51m
kube-system          coredns-6955765f44-j7g74                                     1/1     Running     0          51m
kube-system          etcd-kind-control-plane                                      1/1     Running     0          51m
kube-system          kindnet-9lv48                                                1/1     Running     0          51m
kube-system          kube-apiserver-kind-control-plane                            1/1     Running     0          51m
kube-system          kube-controller-manager-kind-control-plane                   1/1     Running     2          51m
kube-system          kube-proxy-7bfxc                                             1/1     Running     0          51m
kube-system          kube-scheduler-kind-control-plane                            1/1     Running     3          51m
kube-tenant-27656    tenant-kube-apiserver-27656                                  1/1     Running     0          50m
kube-tenant-27656    tenant-kube-controller-manager-27656                         1/1     Running     3          50m
local-path-storage   local-path-provisioner-7745554f7f-tglnq                      1/1     Running     2          51m
```

I then obtained the kubeconfig for the tenant api server and started the stack maanger running against it.
```
🤖 (crossplane) kindred tenant config 27656 > tenant.conf

🤖 (crossplane) kubectl --kubeconfig=tenant.conf apply -f cluster/charts/crossplane/crds -R

🤖 (crossplane) kubectl --kubeconfig=tenant.conf apply -f cluster/charts/crossplane/templates/crds -R

🤖 (crossplane) go run cmd/crossplane/main.go stack manage -d --templates --templating-controller-image v0.3.0-rc --tenant-kubeconfig=/home/dan/code/github.com/crossplaneio/crossplane/tenant.conf --host-controller-namespace=kube-tenant-27656
```

The stack manager started successfully against the tenant API server. I then created a `ClusterStackInstall` for `provider-gcp` against the tenant cluster.
```
🤖 (crossplane) kubectl --kubeconfig=tenant.conf apply -f cluster/examples/provider/provider-gcp.yaml
```

The stack manager successfully started the GCP provider controller against the tenant api server, including installing and labelling its CRDs on the tenant:
```
🤖 (crossplane) kubectl get pods -A
NAMESPACE            NAME                                                         READY   STATUS      RESTARTS   AGE
kube-system          coredns-6955765f44-2v4cn                                     1/1     Running     0          51m
kube-system          coredns-6955765f44-j7g74                                     1/1     Running     0          51m
kube-system          etcd-kind-control-plane                                      1/1     Running     0          51m
kube-system          kindnet-9lv48                                                1/1     Running     0          51m
kube-system          kube-apiserver-kind-control-plane                            1/1     Running     0          51m
kube-system          kube-controller-manager-kind-control-plane                   1/1     Running     2          51m
kube-system          kube-proxy-7bfxc                                             1/1     Running     0          51m
kube-system          kube-scheduler-kind-control-plane                            1/1     Running     3          51m
kube-tenant-27656    crossplane-system.provider-gcp-8htjd                         0/1     Completed   0          11m
kube-tenant-27656    crossplane-system.provider-gcp-controller-76867f9799-wnvj8   1/1     Running     0          10m
kube-tenant-27656    tenant-kube-apiserver-27656                                  1/1     Running     0          50m
kube-tenant-27656    tenant-kube-controller-manager-27656                         1/1     Running     3          50m
local-path-storage   local-path-provisioner-7745554f7f-tglnq                      1/1     Running     2          51m

🤖 (crossplane) kubectl --kubeconfig=tenant.conf get crds | grep gcp
bucketclasses.storage.gcp.crossplane.io                   2020-04-27T15:41:33Z
buckets.storage.gcp.crossplane.io                         2020-04-27T15:41:33Z
cloudmemorystoreinstanceclasses.cache.gcp.crossplane.io   2020-04-27T15:41:35Z
cloudmemorystoreinstances.cache.gcp.crossplane.io         2020-04-27T15:41:35Z
cloudsqlinstanceclasses.database.gcp.crossplane.io        2020-04-27T15:41:33Z
cloudsqlinstances.database.gcp.crossplane.io              2020-04-27T15:41:33Z
connections.servicenetworking.gcp.crossplane.io           2020-04-27T15:41:33Z
gkeclusterclasses.compute.gcp.crossplane.io               2020-04-27T15:41:34Z
gkeclusterclasses.container.gcp.crossplane.io             2020-04-27T15:41:33Z
gkeclusters.compute.gcp.crossplane.io                     2020-04-27T15:41:35Z
gkeclusters.container.gcp.crossplane.io                   2020-04-27T15:41:33Z
globaladdresses.compute.gcp.crossplane.io                 2020-04-27T15:41:34Z
networks.compute.gcp.crossplane.io                        2020-04-27T15:41:34Z
nodepoolclasses.container.gcp.crossplane.io               2020-04-27T15:41:34Z
nodepools.container.gcp.crossplane.io                     2020-04-27T15:41:34Z
providers.gcp.crossplane.io                               2020-04-27T15:41:33Z
serviceaccounts.iam.gcp.crossplane.io                     2020-04-27T15:41:33Z
subnetworks.compute.gcp.crossplane.io                     2020-04-27T15:41:34Z
```

I then created a `Secret` and a `Provider` object with my GCP credentials against the tenant API server. Lastly, I statically provisioned a `GKECluster` against the tenant API server. The GCP controller successfully provisioned the GKE cluster.
```
🤖 (crossplane) cat gkecluster.yaml 
apiVersion: compute.gcp.crossplane.io/v1alpha3
kind: GKECluster
metadata:
  name: gkecluster-standard
  labels:
    oam: test
spec:
  writeConnectionSecretToRef:
    name: k8scluster
    namespace: crossplane-system
  machineType: n1-standard-1
  numNodes: 3
  zone: us-central1-b
  providerRef:
    name: gcp-provider
  reclaimPolicy: Delete

🤖 (crossplane) kubectl --kubeconfig=tenant.conf apply -f gkecluster.yaml

🤖 (crossplane) kubectl --kubeconfig=tenant.conf get -f gkecluster.yaml
NAME                  STATUS    STATE     CLUSTER-NAME   ENDPOINT         CLUSTER-CLASS   LOCATION        RECLAIM-POLICY   AGE
gkecluster-standard   Unbound   RUNNING                  35.226.118.186                   us-central1-b   Delete           3m28s
```

</details>

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/pkg/controller/stacks/stack/stack.go
+++ b/pkg/controller/stacks/stack/stack.go
@@ -86,6 +86,10 @@ var (
 	}
 
 	disableAutoMount = false
+
+	allowPrivilegeEscalation = false
+	privileged               = false
+	runAsNonRoot             = true
 )
 
 // Reconciler reconciles a Instance object
@@ -98,13 +102,14 @@ type Reconciler struct {
 	hostKube             client.Client
 	hostedConfig         *hosted.Config
 	allowCore            bool
+	allowFullDeployment  bool
 	forceImagePullPolicy string
 	log                  logging.Logger
 	factory
 }
 
 // Setup adds a controller that reconciles Stacks.
-func Setup(mgr ctrl.Manager, l logging.Logger, hostControllerNamespace string, allowCore bool, forceImagePullPolicy string) error {
+func Setup(mgr ctrl.Manager, l logging.Logger, hostControllerNamespace string, allowCore, allowFullDeployment bool, forceImagePullPolicy string) error {
 	name := "stacks/" + strings.ToLower(v1alpha1.StackGroupKind)
 
 	hostKube, _, err := hosted.GetClients()
@@ -122,6 +127,7 @@ func Setup(mgr ctrl.Manager, l logging.Logger, hostControllerNamespace string, a
 		hostKube:             hostKube,
 		hostedConfig:         hc,
 		allowCore:            allowCore,
+		allowFullDeployment:  allowFullDeployment,
 		forceImagePullPolicy: forceImagePullPolicy,
 		factory:              &stackHandlerFactory{},
 		log:                  l.WithValues("controller", name),
@@ -156,7 +162,7 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error) 
 		return fail(ctx, r.kube, stack, err)
 	}
 
-	handler := r.factory.newHandler(r.log, stack, r.kube, r.hostKube, r.hostedConfig, r.allowCore, r.forceImagePullPolicy)
+	handler := r.factory.newHandler(r.log, stack, r.kube, r.hostKube, r.hostedConfig, r.allowCore, r.allowFullDeployment, r.forceImagePullPolicy)
 
 	if meta.WasDeleted(stack) {
 		return handler.delete(ctx)
@@ -181,23 +187,25 @@ type stackHandler struct {
 	hostKube             client.Client
 	hostAwareConfig      *hosted.Config
 	allowCore            bool
+	allowFullDeployment  bool
 	forceImagePullPolicy string
 	ext                  *v1alpha1.Stack
 	log                  logging.Logger
 }
 
 type factory interface {
-	newHandler(logging.Logger, *v1alpha1.Stack, client.Client, client.Client, *hosted.Config, bool, string) handler
+	newHandler(logging.Logger, *v1alpha1.Stack, client.Client, client.Client, *hosted.Config, bool, bool, string) handler
 }
 
 type stackHandlerFactory struct{}
 
-func (f *stackHandlerFactory) newHandler(log logging.Logger, ext *v1alpha1.Stack, kube client.Client, hostKube client.Client, hostAwareConfig *hosted.Config, allowCore bool, forceImagePullPolicy string) handler {
+func (f *stackHandlerFactory) newHandler(log logging.Logger, ext *v1alpha1.Stack, kube client.Client, hostKube client.Client, hostAwareConfig *hosted.Config, allowCore, allowFullDeployment bool, forceImagePullPolicy string) handler {
 	return &stackHandler{
 		kube:                 kube,
 		hostKube:             hostKube,
 		hostAwareConfig:      hostAwareConfig,
 		allowCore:            allowCore,
+		allowFullDeployment:  allowFullDeployment,
 		forceImagePullPolicy: forceImagePullPolicy,
 		ext:                  ext,
 		log:                  log,
@@ -217,6 +225,10 @@ func (h *stackHandler) sync(ctx context.Context) (reconcile.Result, error) {
 
 func (h *stackHandler) create(ctx context.Context) (reconcile.Result, error) {
 	h.ext.Status.SetConditions(runtimev1alpha1.Creating())
+
+	if err := h.validateStackPermissions(); err != nil {
+		return fail(ctx, h.kube, h.ext, err)
+	}
 
 	// create RBAC permissions
 	if err := h.processRBAC(ctx); err != nil {
@@ -649,10 +661,6 @@ func (h *stackHandler) processRBAC(ctx context.Context) error {
 
 	labels := stacks.ParentLabels(h.ext)
 
-	if err := h.validateStackPermissions(); err != nil {
-		return err
-	}
-
 	clusterRoleName, err := h.createDeploymentClusterRole(ctx, labels)
 	if err != nil {
 		return err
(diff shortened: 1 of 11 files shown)


pkg/controller/stacks/stack/stack_test.go
```diff
@@ -428,8 +447,40 @@ func TestCreate(t *testing.T) {
 		name       string
 		r          *v1alpha1.Stack
 		clientFunc func(*v1alpha1.Stack) client.Client
+		allowCore  bool
 		want       want
 	}{
+		{
+			name: "FailRestricedPermissions",
```
jbw976, line 454: nit typo: `FailRestricedPermissions` => `FailRestrictedPermissions`

hasheddan: @jbw976 will fix and also update design doc :) Manual testing results coming shortly!

hasheddan:
@jbw976 @negz note that this change will require running with insecure flag to enable using AWS IAM Role for SA functionality (ref: https://github.com/crossplane/provider-aws/blob/master/AUTHENTICATION.md)

pkg/controller/stacks/stack/stack.go
```diff
@@ -86,6 +86,11 @@ var (
 	}
 
 	disableAutoMount = false
+
+	allowPrivilegeEscalation       = false
+	privileged                     = false
+	runAsNonRoot                   = true
+	runAsUser                int64 = 1000
```
jbw976, line 93: what is user 1000? is that standard k8s practice? a comment may be helpful for maintaining this code

edit: ah, it's called out in the design doc.  maybe a quick mention of that idea here in the code too?

hasheddan: Definitely will add a comment :+1: I do not know if this is typical k8s practice, mostly because I don't know of too many systems that are doing what we are doing with stacks. I know that running a container as `USER 1000` is a common practice, but typically you would know the container you are running instead of an arbitrary one in this case. @lukeweber or @negz any thoughts on this?

jbw976: Without having any background on this idea, It "feels" a bit brittle and non-portable to use what seems to me as a random user ID.  From searching around for more information on it this morning, it seems like it's commonly suggested, but it's harder to find a good justification for doing so.  It almost feels like one of those "examples" that becomes more idiomatic through copy/paste in the ecosystem 😉 

This article gets into a bit more description, in its "Do Not Run As Root" section: https://www.bluematador.com/blog/kubernetes-security-essentials

What happens if there is no user 1000 in the uid namespace the container is running in? Is 1000 always guaranteed to exist, or does that not actually matter?

This article indicates you can run a container as a totally random user ID and in fact that's what OpenShift does, see section "Use Case #2 – Running API Simulator Container Using Random Non-root User":
https://apisimulator.io/run-docker-containers-non-root-users-random-user-ids/

So while to me this seemed non-intuitive or a bit fragile, perhaps this is a fairly acceptable approach to "force" the container to not run as root, which in many cases it will try to do, for instance when no specific user was specified in the container build process.  Any other thoughts?

hasheddan: This pretty much sums up my thoughts as well. I also don't love setting to a user ID explicitly for a general "installer". However, we are setting to a user ID rather than a user, so we do not need to be concerned about the user not existing.

We could alternatively say that the onus falls on the creator of the Docker image to ensure that it runs as non-root, something that we should do in Crossplane-maintained images anyway. This might make it a bit easier to debug as the container will just fail rather than potentially being run without root permissions that it requires for some reason. Basically what we are trying to do here is make it such that installations do not fail for mis-configured Docker images, which I don't know if this is something we should do or not. I could go either way here.

hasheddan: For posterity: decision here is to not set user and document that containers must be able to run as non-root.

hasheddan:
@jbw976 I think we can merge here if this looks good to you. The PR's that are still open on other repos will pin to the crossplane release that will include this.

PR #1163, opened 2020-01-13
Composing apps and their infra dependencies

negz:
### Description of your changes

Fixes https://github.com/crossplane/crossplane/issues/1105
Fixes https://github.com/crossplane/crossplane/issues/1106
Fixes https://github.com/crossplane/crossplane/issues/1107
Fixes https://github.com/crossplane/crossplane/issues/1344

This pull request adds a design for the _composition_ of both applications and infrastructure in Crossplane.

Composition allows an application developer to model their application as a kind of Kubernetes resource (e.g. `Wordpress`) and to describe one or more _compositions_ that may be used to reconcile their application into the various namespaced resources that implement it. For example a `Wordpress` could be rendered as a `KubernetesApplication` and scheduled to a remote Kubernetes cluster, or run as a `Deployment` and a `Service` on the cluster where Crossplane is running.

Similarly, an infrastructure operator (or provider) may compose the "primitive" managed resources that are implemented by an infrastructure provider into more advanced, functional, composite managed resources. For example an Azure `MySQLServer` and a `MySQLServerVirtualNetworkRule` could be composed into a `PrivateMySQLServer`, allowing infrastructure operators to simply and consistently provision private MySQL servers.

Finally, an infrastructure operator may publish a kind of infrastructure resource that an application may require. Application operators may specify a requirement for an existing instance of a published kind of infrastructure resource, or provision one on demand. Publishing `PrivateMySQLServer` for use by applications allows an application operator to author a namespaced `PrivateMySQLServerRequirement` in order for their application to be allocated a cluster scoped `PrivateMySQLServer`.

Putting this all together:

1. An infrastructure operator installs `provider-azure` to add support for Azure primitive managed resources, including `MySQLServer` and `VirtualNetworkRule`.
1. The infrastructure operator defines `PrivateMySQLServer`, along with a composition that renders a `PrivateMySQLServer` to a `MySQLServer` and a `MySQLServerVirtualNetworkRule`.
1. The infrastructure operator publishes `PrivateMySQLServer` so that applications may use it.
1. An application developer defines `Wordpress`, along with a composition that renders a `kind: Wordpress` to a `Deployment`, `Server`, and `PrivateMySQLServerRequirement`.
1. An application operator creates a `Wordpress`, thus deploying wordpress as a Kubernetes `Deployment` with a `Service` for network ingress and a `PrivateMySQLServer` for data storage.

PR #1447, opened 2020-04-27
Crossplane ClusterRole expansion for composition

muvaf:

Using composition, Crossplane pod now provisions the managed resources directly and it needs the necessary ClusterRole to do that. There is already a `crossplane-admin` cluster role that when stacks are installed, the clusterrole to manage their CRs are aggregated to that persone. This PR expands the ClusterRole of Crossplane pod to include the ClusterRoles that `crossplane-admin` has so that the CRs of stacks can be part of composition.

In the end, we'll end up with Crossplane being able to deploy everything everywhere [as per the composition desig](https://github.com/crossplane/crossplane/pull/1163)n but this change only covers the case where managed resources are composed.

Manually tested with Azure stack.

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/cluster/charts/crossplane-types/templates/clusterrole.yaml
+++ b/cluster/charts/crossplane-types/templates/clusterrole.yaml
@@ -7,6 +7,25 @@ metadata:
     chart: {{ template "chart" . }}
     release: {{ .Release.Name }}
     heritage: {{ .Release.Service }}
+aggregationRule:
+  clusterRoleSelectors:
+  - matchLabels:
+      rbac.crossplane.io/aggregate-to-crossplane-admin: "true"
+  - matchLabels:
+      rbac.crossplane.io/aggregate-to-crossplane: "true"
+rules: []
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ template "name" . }}:system:aggregate-to-crossplane
+  labels:
+    app: {{ template "name" . }}
+    chart: {{ template "chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+    crossplane.io/scope: "system"
+    rbac.crossplane.io/aggregate-to-crossplane: "true"
 rules:
 - apiGroups:
   - ""
@@ -33,6 +52,7 @@ rules:
   - watch
   - create
   - update
+  - patch
 - apiGroups:
   - cache.crossplane.io
   - compute.crossplane.io
--- a/cluster/charts/crossplane/templates/clusterrole.yaml
+++ b/cluster/charts/crossplane/templates/clusterrole.yaml
@@ -7,6 +7,25 @@ metadata:
     chart: {{ template "chart" . }}
     release: {{ .Release.Name }}
     heritage: {{ .Release.Service }}
+aggregationRule:
+  clusterRoleSelectors:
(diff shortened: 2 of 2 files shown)


cluster/charts/crossplane-types/templates/clusterrole.yaml
```diff
@@ -7,6 +7,25 @@ metadata:
     chart: {{ template "chart" . }}
     release: {{ .Release.Name }}
     heritage: {{ .Release.Service }}
+aggregationRule:
+  clusterRoleSelectors:
+  - matchLabels:
+      rbac.crossplane.io/aggregate-to-crossplane-admin: "true"
+  - matchLabels:
+      rbac.crossplane.io/aggregate-to-crossplane: "true"
```
negz, line 15: Are you thinking that we could eventually have the stack manager add this label to newly installed roles, so we can differentiate between what `crossplane-admin` should grant and what `crossplane` should grant?

negz:
> we are already trusting it to install the stacks which themselves have RBAC control over their CRDs

That's done by the stack-manager pod, right? I believe we decided to run stack-manager as a separate deployment in order to restrict the access the Crossplane service account gets. I wonder if that still makes sense? I will note that the stack-manager appears to run as `cluster-admin`, so even with these expanded Crossplane permissions the stack-manager likely has more.

cluster/charts/crossplane-types/templates/clusterrole.yaml
muvaf: Actually, I don't see a case where there is a ClusterRole that Crossplane creates with more permissions than Crossplane pod itself anymore (same treatment as stack-manager). So, the SA of Crossplane will have the most permissions regarding stacks and `crossplane-admin` would probably have the same level of permissions as it's _admin_. So I don't immediately see a case where we want to differentiate on stack level.

The differentiation exists in this file to some extent; aggregation `rbac.crossplane.io/aggregate-to-crossplane` is the minimal set that Crossplane needs without any stack. But as the stacks get installed, both `crossplane-admin` and `crossplane` end up having the same rules.

hasheddan:
@negz I was meaning the trust a user was giving to crossplane in general terms, but it is a good point that it may not make sense to keep the core and stack manager separate any more

jbw976:
I like this approach in that (I believe) it is essentially aggregating permissions to only types that the packages contain. It's not willy nilly taking cluster-admin permissions to `*`.  This keeps the scope of permissions in crossplane to a superset of all packages that have been installed.  

I had to refresh my memory on why stack-manager would be different and needs cluster-admin, which I'll leave here for convenience from the [security design doc](https://github.com/crossplane/crossplane/blob/master/design/one-pager-stacks-security-isolation.md#stack-manager):
> Currently, the Stack Manager that installs stacks into the cluster runs with a ClusterRoleBinding to the cluster-admin ClusterRole because it needs to create RBAC permissions for the arbitrary CRDs that are owned and depended on by arbitrary stacks. A service account cannot grant RBAC permissions that it doesn't already have itself, so to support the arbitrary permissions needs of all stacks, we are currently using the cluster-admin role.

muvaf:
> This keeps the scope of permissions in crossplane to the set of all packages that have been installed.

@jbw976 Yeah I'm glad that we've been doing this for the clusterroles. It's a nice way to stay away from `*`.

PR #1452, opened 2020-04-29
Unit tests for composite package

muvaf:

Unit tests for `pkg/composite`. I'll keep this in draft until all is complete but keep pushing whatever I have so that others don't overlap.

Fixes https://github.com/crossplane/crossplane/issues/1472

Unit test.

### Checklist

I have:
  appropriate.

[definition of done].


muvaf:
It's funny that codecov fails in cases where you decrease the coverage by a lot but also when you increase it by a lot 🤦

go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/crossplane/crossplane
 
 go 1.13
 
+replace github.com/crossplane/crossplane-runtime => github.com/muvaf/crossplane-runtime v0.0.0-20200429141134-1ddc2c5d8d89
```
negz, line 5: Mental note: remove this before merge. I believe the changes are in crossplane-runtime now.

pkg/controller/apiextensions/composite/composed/composed.go
```diff
@@ -0,0 +1,181 @@
...
+			return Observation{}, errors.Wrap(err, errGetComposed)
+		}
+	}
+
+	if !meta.WasCreated(cd) {
+		// Configure is called only if Composed resource is not created yet
+		// because the fields that are not referred in the patches should be
+		// mutable by the user, i.e. dependency on the Composition should be
+		// limited to the field propagation.
+		if err := r.Configure(cp, cd, t); err != nil {
```
muvaf, line 145: @negz Are you sure that we should unmarshal from template in each reconcile? Most of the time, this will make the composed resource immutable since any changes to fields that are not propagated from composite will be overwritten. We took this approach in templating-controller but it results in some problems, for example, if the composition had an array of tags and a late-init process wants to add something, the whole array is overwritten to the old values that exist in the composition, so, you get two controllers keep changing the same fields. This would happen for the propagated fields as well but that's more discoverable and I'd expect that usually much small number of fields would be propagated anyway compared to number of fields in the template.

I think, in general, it's desirable that we have the least number of spec fields being overwritten so that the resources are kept interactable, either by user or by another controller. Continuously enforcing the template greatly increases that number.

Other point I mentioned in the meeting is that changing a field on the primitive managed resource will become harder as it will require a new `Composition` resource to be created and pointed at; just for one field. When you consider that the person who creates the `Composition`s in the cluster, i.e. infra ops, the users of the resource will need them to do the change and `Composition` creation permission is an important one since it essentially enables people to provision anything, so infra-ops will always bear that cost of manual operation for the users. In other case, they just give the permission to work on specific composed resource kind(s). There could even be an RBAC controller which scans the cluster and allows SAs in the specific namespace to own the composed resources that are bound to requirements in that namespace to automate this. But there isn't much you can do if you have to have a `Composition` creation permission.

muvaf: That's probably subjective but it's easier for me to know that there is one single source of truth in my composed resources: requirement in my namespace. If its `Composition` is another source of truth, then I have to make sure the field I want to change does not appear in the template in `Composition` resource.

negz: Let me think about this and get back to you. I feel that the `kind: Composition` being considered on every reconcile is conceptually simpler (it seems like _fewer_ sources of truth to me), but I can see your point regarding conflicts with late initialization.

pkg/controller/apiextensions/composite/composed/composed.go
```diff
@@ -0,0 +1,181 @@
...
+	// Overlay is applied to the Composed resource in all cases so that we can
+	// keep Composed resource up-to-date with the changes in Composite resource.
+	if err := r.Overlay(cp, cd, t); err != nil {
+		return Observation{}, errors.Wrap(err, errOverlay)
+	}
+
+	// Connection details are fetched in all cases in a best-effort mode, i.e.
+	// it doesn't return error if the secret does not exist or the resource
+	// does not publish a secret at all.
+	conn, err := r.Fetch(ctx, cd, t)
```
negz, line 159: I think this would read better if it weren't an embedded interface. It's not obvious here what is being fetched.

pkg/controller/apiextensions/composite/reconciler.go
```diff
@@ -39,6 +39,7 @@ import (
 	"github.com/crossplane/crossplane-runtime/pkg/resource/unstructured/composite"
 
 	"github.com/crossplane/crossplane/apis/apiextensions/v1alpha1"
+	composedctrl "github.com/crossplane/crossplane/pkg/controller/apiextensions/composite/composed"
```
negz, line 42: What's the thinking behind moving the composer into its own sub-package?

muvaf: One reason was that the interfaces/structs would collide with `Composite` resource reconciler, like both having `Configurator` so I'd have to prefix like `ComposedConfigurator`. Also I'd specify the struct name to say it's `Composed*` because they depend on being used for composed resources rather generic things. Then I realized it'd make sense for it to have its own package so that it's easier to see which interface/object is used by what exactly.

pkg/controller/apiextensions/composite/composed/composed.go
muvaf: Yeah, thanks for the suggestion!

muvaf:
@negz This is ready for final review. I am writing an issue about our discussion related to using the template in each reconcile or not, will cc you once opened.

pkg/controller/apiextensions/composite/composed/composed.go
negz: Just capturing our decision from 1:1 today - I'm okay with this being merged as is for now because don't have a strong decision either way. I do still think it should be considered on every reconcile, but I want to write out my justification in a separate issue to avoid blocking this PR.

negz: Here's that issue I promised: https://github.com/crossplane/crossplane/issues/1481.

pkg/controller/apiextensions/composite/api_test.go
```diff
@@ -21,20 +21,29 @@ import (
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/pkg/errors"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
negz, line 24: Nit: We typically import this as `metav1`.

pkg/controller/apiextensions/composite/composed/api.go
```diff
@@ -114,10 +115,17 @@ func (cdf *APIConnectionDetailsFetcher) Fetch(ctx context.Context, cd resource.C
 	// iteration.
 	s := &corev1.Secret{}
 	nn := types.NamespacedName{Namespace: sref.Namespace, Name: sref.Name}
-	if err := cdf.client.Get(ctx, nn, s); resource.IgnoreNotFound(err) != nil {
+	err := cdf.client.Get(ctx, nn, s)
```
negz, line 118: Nit: This could be abbreviated:

```go
if err := ...; err != nil {
    return nil, errors.Wrap(resource.IgnoreNotFound(err), errGetSecret)
}
```

pkg/controller/apiextensions/composite/composed/composed.go
negz: @muvaf After writing up #1481 and thinking on it for a few days I'm feeling like I would prefer to consider the composition on every reconcile for the time being. I'm hoping to merge https://github.com/crossplane/crossplane/pull/1163 this week, and follow up with a design PR proposing the behaviour outlined in #1481.

muvaf:
> This all looks good to me, but per #1452 (comment) I'd prefer to revert to applying the composition on every reconcile (despite its shortcomings) for the time being.

Sounds good. I have updated the PR to apply from the template in each reconcile and left a note in the comment.

pkg/controller/apiextensions/composite/composed/composed.go
```diff
@@ -150,14 +150,11 @@ func (r *Composer) Compose(ctx context.Context, cp resource.Composite, cd resour
 		}
 	}
```
negz, line 151: Is there a benefit to getting the current resource while we're continuously configuring it?

muvaf: Not anymore since now we always unmarshal from the template.

muvaf:
@negz Removed the unnecessary `Get`, should be ready to merge now.

Issue #1448, opened 2020-04-27
[Design Proposal] Add a generic mechanism for traits to work with workloads in OAM runtime


ryanzhang-oss:
### Problem
Traits and workloads are two major types of resources in OAM.  Traits usually affect how a workload operate either directly (through spec change) or indirectly (add ingress or sidecar). However, the current OAM implementation does not contain a generic mechanism for traits to locate and apply the desired changes to the resources. Therefore, we propose to address this problem in the OAM Kubernetes runtime.

### Design principles
Before we lay out the detailed design.  Let's first state the overall design objections.
1. We want to design an extensible mechanism that allows a `trait` to apply to any eligible `workload`. This means that this mechanism allows a trait developer to write the controller code once and the controller will work for any new `workloads` that this `trait` can apply to in the future.  

2. The mechanism cannot put any limit on the `trait` or `workload` CRDs. This means that we cannot pre-define any common CRD fields beyond Kubernetes conventions. This is because we want to accommodate any existing CRD to be imported as `workload` or `trait`. 

### Mechanism explained in detail
Here is the step by step workflow
1.  Add a `childResources` field in the  WorkloadDefinition. 
Currently, a workloadDefinition is nothing but a shim of a real workload CRD. We propose to add an **optional** field called `childResources` to the spec of the workloadDefinition. We encourage workload owners to fill in this field when they register their controllers to the OAM system. In this way, they can declare the types of the Kubernetes resources the workload controller actually generates.  For example,
```yaml
apiVersion: core.oam.dev/v1alpha2
kind: WorkloadDefinition
metadata:
   name: schema.example.oam.dev
spec:
  definitionRef:
     name: schema.example.oam.dev
  childResources:
     - groupVersion: apps/v1
       kind: deployment
     - groupVersion: core/v1
       kind: service      
```

2.  ApplicationConfig controller add labels containing the workload information to the emitted trait CR.  For example,
```yaml
apiVersion: core.oam.dev/v1alpha2
kind: ManualScalerTrait
metadata:
  labels:
    core.oam.dev/workloadGroupVersion: core.oam.dev/v1alpha2
    core.oam.dev/workloadKind:  ContainerizedWorkload
    core.oam.dev/workloadName: foo
```

3.  The trait controller can use the following stps in the Reconcile logic to find out all the resources that it should try to modify.

-  Fetch the corresponding `workload` instance by combining the label fields following Kubernetes convention (gvk)
[...]

PR #1454, opened 2020-04-29
add trait workload interaction mechanism

ryanzhang-oss:
The one-page design doc for OAM trait workload interaction

### Checklist

I have:
  appropriate.

[definition of done].


design/one-pager-trait-workload-interaction-mechanism.md
```diff
@@ -0,0 +1,112 @@
...
+name of the `workloadDefinition` CR is the name of the `workload` CRD it refers to. For example
+, the name of the `workloadDefinition` CR that refers to a `containerizedworkloads.core.oam.dev`
+CRD is exactly `containerizedworkloads.core.oam.dev` as well. For example, the following workload CR
+have all its child resources GVK in its 
+    ```yaml
+    apiVersion: standard.oam.dev/v1alpha2
+    kind: Mydb
+    metadata:
+      annotations:
+        childResources: "{ apps/v1.deployment, core/v1.service}"
```
wonderflow, line 87: add `resource name` annotation if a workload can launch multiple k8s resource with same type

design/one-pager-trait-workload-interaction-mechanism.md
```diff
@@ -0,0 +1,179 @@
...
+## Terminology
+
+* **CRD (Custom Resource Definition)** : A standard Kubernetes Custom Resource Definition
+* **CR (Custom Resource)** : An instance of a Kubernetes type that was defined using a CRD
+* **GVK (Group Version Kind)** : The API Group, Version, and Kind for a type of Kubernetes
+ resource (including CRDs)
+* **Workload child resources** : The Kubernetes resources generated by a workload controller. They
+ should all have a owner reference pointing to the parent workload instance.
+
+## Background
```
hongchaodeng, line 16: The background section is way too simplified. It should also cover some user scenarios as discussed in https://github.com/crossplane/crossplane/issues/1448#issuecomment-620323239.

design/one-pager-trait-workload-interaction-mechanism.md
```diff
@@ -0,0 +1,179 @@
...
+sidecar). However, the current OAM implementation does not contain a generic mechanism for traits
+ to locate the corresponding resource to modify. The problem is two folds
+1. A trait controller needs a way to find the workload that it is applied to
+2. A trait controller needs to know the exact resources it should modify. Note that these
+resources are most likely not the workload itself.
+
+## Goals
+In order to maximize the extensibility of our OAM implementation. We also operate under the
+following design objections.
+1. We want to allow a `trait` to apply to any eligible `workload`. This means that we want
```
hongchaodeng, line 28: These points are too abstract. We should provide one or two examples to demonstrate the end use cases.

design/one-pager-trait-workload-interaction-mechanism.md
ryanzhang-oss: mentioned it in the proposal but not sure if we need to implement it right away.

design/one-pager-trait-workload-interaction-mechanism.md
```diff
@@ -0,0 +1,197 @@
...
+1. A trait controller needs a way to find the workload that it is applied to. 
+   - In the example, the manualScalar trait needs to know that it is supposed to scale the
+    example-db workload. However, since the applicationConfiguration controller just emit a
+     `ManualScalerTrait` CR, it does not contain any information on the `example-db` workload.
+2. A trait controller needs to know the exact resources it should modify. Note that these
+resources are most likely not the workload itself.
+    - Use the same example, just knowing the `example-db` workload is not enough for the
+     ManualScalerTrait to work. The trait controller does not work with the `example-db` workload
+      directly. It needs to find the actual Kubernetes resources that the `example-db`
+     workload generates and then it can modify the `replica` field in its Pod spec.
```
hongchaodeng, line 84: Well, Pod spec has no `replica` field.
```suggestion
     workload generates and then it can modify the `replica` field in its spec.
```

design/one-pager-trait-workload-interaction-mechanism.md
```diff
@@ -0,0 +1,197 @@
...
+
+## Goals
+In order to maximize the extensibility of our OAM implementation. We also operate under the
+following design objections.
+1. We want to allow a `trait` to apply to any eligible `workload`. This means that we want
+to empower a trait developer to write the controller code once and the it will work for
+any new `workload` that this `trait` can apply to in the future.
+    - Using the example again, the `ManualScalerTrait` should work with any workload that
+     generates a Kubernetes resource that has a `replica` field in its Pod spec even if the
```
hongchaodeng, line 93: Pod spec has no replica field.

design/one-pager-trait-workload-interaction-mechanism.md
```diff
@@ -0,0 +1,197 @@
...
+     `ManualScalerTrait` CR, it does not contain any information on the `example-db` workload.
+2. A trait controller needs to know the exact resources it should modify. Note that these
+resources are most likely not the workload itself.
+    - Use the same example, just knowing the `example-db` workload is not enough for the
+     ManualScalerTrait to work. The trait controller does not work with the `example-db` workload
+      directly. It needs to find the actual Kubernetes resources that the `example-db`
+     workload generates and then it can modify the `replica` field in its Pod spec.
+
+## Goals
+In order to maximize the extensibility of our OAM implementation. We also operate under the
```
hongchaodeng, line 87: > In order to maximize the extensibility of our OAM implementation. We also operate under the following design objections.

Need to fix grammar?

design/one-pager-trait-workload-interaction-mechanism.md
```diff
@@ -0,0 +1,197 @@
...
+following design objections.
+1. We want to allow a `trait` to apply to any eligible `workload`. This means that we want
```
hongchaodeng, line 89:
```suggestion
1.  **Trait portability**: We want to allow a `trait` to apply to any eligible `workload`. This means that we want
```

design/one-pager-trait-workload-interaction-mechanism.md
```diff
@@ -0,0 +1,197 @@
...
+to empower a trait developer to write the controller code once and the it will work for
+any new `workload` that this `trait` can apply to in the future.
+    - Using the example again, the `ManualScalerTrait` should work with any workload that
+     generates a Kubernetes resource that has a `replica` field in its Pod spec even if the
+      workload does not exist when the `ManualScalerTrait` is implemented.
+2. The mechanism cannot put any limit on the `trait` or `workload` CRDs. This means that we
```
hongchaodeng, line 95:
```suggestion
2. **Adopting existing CRDs**: The mechanism cannot put any limit on the `trait` or `workload` CRDs. This means that we
```

hongchaodeng:
fix https://github.com/crossplane/crossplane/issues/1448

design/one-pager-trait-workload-interaction-mechanism.md
```diff
@@ -0,0 +1,198 @@
...
+ applicationConfiguration controller.
+- Workload: This mechanism does not affect workload implementation.
+- Trait: This mechanism is optional so all existing trait controller still works. This mechanism
+requires modification for any existing traits implementation that wants to take advantage of
+extensibility of OAM. Any traits that only applies to a certain type of workload don't need to
+use this mechanism.
+- WorkloadDefinition: We can modify the existing workloadDefinition if needed.
+
+## Alternative approach
+1. One alternative approach is that we can make the applicationConfiguration controller watch all
```
artursouza, line 185: I found this proposal a little confusing to follow. Specially because it adds complexity to AppConfig. Can we have an alternative approach where the controller automatically annotates the child resources with a reference to the parent resource via an annotations? This way, the generated child can be retrieved via querying the annotations. This way, there is no change in the CRDs and this becomes a spec for the controller only. OAM users are not even aware that these annotations are needed since they are auto-generated.

I might be missing something here but it feels like this proposal is making implementation simpler on the expense of increasing the user's language surface.

artursouza: Ryan and I had a call on this. Here are the clarifying points for this proposal:

1. We want to allow existing operators to be used with little or no change, so adding requirements of automatically generating annotations is against that principle.

2. Only the WorkloadDefinition will be “against” the spec by adding childResources property, the other changes are in the generated resources. So, the OAM resources are virtually unchanged.

3. We have a path forward in case there is a more complex generated resources and labeling can be used - no need to implement that now but we have it as a way out if that becomes an issue.

Given the points above. I am fine with this proposal.

Special thanks to Ryan for driving this change and creating clarity as we review it.

design/one-pager-trait-workload-interaction-mechanism.md
```diff
@@ -0,0 +1,198 @@
...
+     because OAM platform builders or infrastructure teams can install any existing CRDs they
+     maintained to the platform as a `workload` or `trait`. Therefore, our solution needs to
+     accommodate pre-defined CRDs that OAM users bring.
+
+## Proposal
+The overall idea is for the applicationConfiguration controller to annotate critical information
+in the workload and trait CR it emits. In addition, we will provide a helper library so that
+trait controller developers can locate the resources they need with a simple function call. 
+Here is the list of changes that we propose.
+1. Add a `childResources` field in the  WorkloadDefinition.
```
hasheddan, line 110: My immediate thought here was that it feels a little weird to be defining on the `WorkloadDefinition` something that is enforced directly by code of the workload controller. However, I actually think this is a fine pattern as the `WorkloadDefinition` will likely be installed by the system that is implementing the workload controllers, similar to how a Kubernetes operator installs its CRDs 👍

design/one-pager-trait-workload-interaction-mechanism.md
```diff
@@ -0,0 +1,198 @@
...
+    metadata:
+       name: mydb.standard.oam.dev
+    spec:
+      definitionRef:
+         name: mydb.standard.oam.dev
+      childResources:
+         - groupVersion: apps/v1
+           kind: deployment
+         - groupVersion: core/v1
+           kind: service
```
hasheddan, line 129: Nitpick:
```suggestion
         - apiVersion: apps/v1
           kind: Deployment
         - apiVersion: v1
           kind: Service      
```

design/one-pager-trait-workload-interaction-mechanism.md
```diff
@@ -0,0 +1,198 @@
...
+    metadata:
+      name: mydb-example
+      annotations:
+        core.oam.dev/childResources: "{ apps/v1.deployment, core/v1.service}"
+    ```
+
+3. ApplicationConfig controller add annotations containing the workload information to the emitted
+`trait` CR. The application controller actually generates the workload CR so it has all the
+information. In our example, the applicationConfig controller will emit the following `trait` CR with
+the GVK and name of the workload the `ManualScalerTrait` is applied to.
```
hasheddan, line 153: I wonder about the best way to handle situations in which a trait is defined for a workload in the `ApplicationConfiguration`, but the trait is not actually able to modify the rendered child resources. For instance, the manual scaler trait can only modify workload types that render out a `Deployment` right now. The model described here would render out the workload components then the `ManualScalerTrait` would try to modify the `Deployment`. If the workload did not produce a `Deployment`, the `ManualScalerTrait` controller would likely just not do anything here. This might be the right strategy (i.e. if you use a trait that does not apply to the workload it is basically a no-op). However, we could also do a similar thing that you propose for `WorkloadDefinition` in `TraitDefinition` by adding a `appliesTo` field (not sure that is the best name) with a slice of object types it knows how to modify. So for the `ManualScalerTrait` it could look like this:

```yaml
kind: TraitDefinition
metadata:
  name: manualscalertraits.core.oam.dev
spec:
  definitionRef:
    name: manualscalertraits.core.oam.dev
  appliesTo:
    - apiVersion: apps/v1
      kind: Deployment
    - apiVersion: apps/v1
      kind: StatefulSet # not currently supported but useful for illustration purposes
```

Following this pattern would allow the `ApplicationConfiguration` controller to decide if a configuration is valid, and it could refuse to render out workloads / traits if it is not. I am not sure if this is more desirable or not, but it could guard against situations where a user thinks they are deploying one configuration, but another one is actually being created (i.e. we won't render a workload if the trait is not compatible).

design/one-pager-trait-workload-interaction-mechanism.md
```diff
@@ -0,0 +1,198 @@
...
+2. Another approach is to implement a separate type for binding traits to workloads. This would
+ work but it seems that label/annotation is a natural place to record the information. Otherwise
+ , we need a way for the trait to discover the binding instance first.   
+
+## Extra labels
+There might be cases that a workload generates more than one resources with the same GVK and only
+want to expose a subset of them to traits. In this case, we can add a pre-defined label such as
+`core.oam.dev/expose=true` for the workload owner to indicate what resources to expose. This
+section is just to illustrate that this is a solvable problem and it's beyond the scope of this
+proposal for now .
```
hasheddan, line 198:
```suggestion
proposal for now.
```

design/one-pager-trait-workload-interaction-mechanism.md
```diff
@@ -0,0 +1,198 @@
...
+correspoinding workload CR. I would not recommend this approach as it increases the complexity
+of the applicationConfiguration controller and makes more of availability liability. 
+2. Another approach is to implement a separate type for binding traits to workloads. This would
+ work but it seems that label/annotation is a natural place to record the information. Otherwise
+ , we need a way for the trait to discover the binding instance first.   
+
+## Extra labels
+There might be cases that a workload generates more than one resources with the same GVK and only
+want to expose a subset of them to traits. In this case, we can add a pre-defined label such as
+`core.oam.dev/expose=true` for the workload owner to indicate what resources to expose. This
```
hasheddan, line 196: Ah nice! This addresses my comment above about how a trait will modify multiple resources of the same kind rendered by a workload. I still wonder about listing a resource multiple times in the `childResources` section. Maybe this is not necessary though as the trait reconciler should theoretically list all of every type of resource it cares about with owner references to the workload it cares about 👍

design/one-pager-trait-workload-interaction-mechanism.md
```diff
@@ -0,0 +1,198 @@
...
+    apiVersion: core.oam.dev/v1alpha2
+    kind: ManualScalerTrait
+    metadata:
+      annotations:
+        core.oam.dev/workloadGroupVersion: standard.oam.dev/v1alpha2
+        core.oam.dev/workloadKind: Mydb
+        core.oam.dev/workloadName: mydb-example
+    ```
+4. OAM runtime will provide a helper library. The library follows the following logic to help a
+ trait developer locate the resources for the trait to modify
```
hasheddan, line 164: This is essentially the function of the shared reconcilers written for traits and workloads in `oam-kubernetes-runtime`. Slight modifications would make them conformant with these new patterns :)

design/one-pager-trait-workload-interaction-mechanism.md
ryanzhang-oss: I like this idea and it plays well with the overall trait <-> workload compatibility scheme although the `traitDefinition` CRD already has an `appliesToWorkloads ` field.
To be more specific, I think this is part of the other problem we ought to solve. Namely, how can we know if a `trait` can be applied to a `workload`.  I think we should definitely incorporate this idea when we get to implement https://github.com/oam-dev/spec/issues/153.  The overall idea is that we will add some annotation keys to a component to indicate its runtime characteristics. We can then use the `appliesToWorkloads ` field in the `traitDefinition` to see if the trait can apply to the component instance. I think we can combine your idea with that scheme.  We will probably use a validatingMutator to implement it.

design/one-pager-trait-workload-interaction-mechanism.md
ryanzhang-oss: In this case, is it ok for me to modify that function? Will it break the remote add-on?

design/one-pager-trait-workload-interaction-mechanism.md
```diff
@@ -0,0 +1,198 @@
...
+    1. Get the corresponding `workload` instance from the Kubernetes cluster with the information
+     contained in the annotation field inserted by the application controller in the `trait` CR.
+    2. Fetch all the `childResources` values inserted by the application controller in the
+     annotation field in the `workload` instance.
```
negz, line 168: Would it be feasible to lookup the `WorkloadDefinition` that corresponds to the workload and read its `spec.childResources`, rather than requiring an annotation on the workload CR itself? This would avoid us having to repeat that information many times, and also avoid us storing it as unvalidated, opaque structured data (presumably JSON) inside an annotation.

design/one-pager-trait-workload-interaction-mechanism.md
```diff
@@ -0,0 +1,198 @@
...
+information. In our example, the applicationConfig controller will emit the following `trait` CR with
+the GVK and name of the workload the `ManualScalerTrait` is applied to.
+    ```yaml
+    apiVersion: core.oam.dev/v1alpha2
+    kind: ManualScalerTrait
+    metadata:
+      annotations:
+        core.oam.dev/workloadGroupVersion: standard.oam.dev/v1alpha2
+        core.oam.dev/workloadKind: Mydb
+        core.oam.dev/workloadName: mydb-example
```
negz, line 161:
```suggestion
        workload.oam.dev/apiVersion: standard.oam.dev/v1alpha2
        workload.oam.dev/kind: Mydb
        workload.oam.dev/name: mydb-example
```

How do you feel about the above structure? Similar to my previous comment, it's a little closer to the [`ObjectReference`](https://github.com/kubernetes/api/blob/a4f2d87006da76427d0d3d613f2ffec13d8e49d3/core/v1/types.go#L5053) convention that is common in Kubernetes.

design/one-pager-trait-workload-interaction-mechanism.md
```diff
@@ -0,0 +1,198 @@
...
+- Trait: This mechanism is optional so all existing trait controller still works. This mechanism
+requires modification for any existing traits implementation that wants to take advantage of
+extensibility of OAM. Any traits that only applies to a certain type of workload don't need to
+use this mechanism.
+- WorkloadDefinition: We can modify the existing workloadDefinition if needed.
+
+## Alternative approach
+1. One alternative approach is that we can make the applicationConfiguration controller watch all
+the possible workload child resources. It also inserts the child resources GVK and name to the
+correspoinding workload CR. I would not recommend this approach as it increases the complexity
```
negz, line 187:
```suggestion
corresponding workload CR. I would not recommend this approach as it increases the complexity
```

design/one-pager-trait-workload-interaction-mechanism.md
```diff
@@ -0,0 +1,198 @@
...
+of the applicationConfiguration controller and makes more of availability liability. 
+2. Another approach is to implement a separate type for binding traits to workloads. This would
+ work but it seems that label/annotation is a natural place to record the information. Otherwise
+ , we need a way for the trait to discover the binding instance first.
```
negz, line 191: I have to admit I'm a little curious what this (a separate type) would look like. I'm always a little wary of modelling required or structured data in annotations, which are opaque to the Kubernetes API server.

For example in this proposal the API server couldn't validate that the proposed `core.oam.dev/childResources` annotation contained valid JSON, or that all three of the proposed `core.oam.dev/workload*` annotations were present and non-zero. Introducing new types to model these things may seem a little heavyweight, but I do like that we'd be able to describe the schema of those types via a CRD and thus leverage API server validation.

design/one-pager-trait-workload-interaction-mechanism.md
```diff
@@ -0,0 +1,198 @@
...
+
+3. ApplicationConfig controller add annotations containing the workload information to the emitted
+`trait` CR. The application controller actually generates the workload CR so it has all the
+information. In our example, the applicationConfig controller will emit the following `trait` CR with
+the GVK and name of the workload the `ManualScalerTrait` is applied to.
+    ```yaml
+    apiVersion: core.oam.dev/v1alpha2
+    kind: ManualScalerTrait
+    metadata:
+      annotations:
```
negz, line 158: It seems like we're taking the contemporary `spec.workloadRef` of OAM-aware traits like `ManualScalerTrait` and moving them into annotations. I believe the idea is that we want to allow existing Kubernetes custom resources (that are blissfully unaware of OAM) to act as traits, similar to how existing OAM-unaware custom resources can act as OAM workloads.

How does this mesh with the below proposal that trait developers will use an OAM-specific library to lookup the workload to which a trait applies? It sounds like this data will only be used by developers who are intending to write an OAM trait. Presumably an existing OAM-unaware Kubernetes CR would not be aware of these annotations.

My suspicion is that if it's possible for an existing Kubernetes CR to act as an OAM trait, then that CR will already have some way of indicating what resource it operates on, for example a label selector in its spec.

design/one-pager-trait-workload-interaction-mechanism.md
```diff
@@ -0,0 +1,198 @@
...
+           name:  example-appconfig-trait
+         spec:
+           replicaCount: 3
+```
+
+The problem is two folds
+1. A trait controller needs a way to find the workload that it is applied to. 
+   - In the example, the manualScalar trait needs to know that it is supposed to scale the
+    example-db workload. However, since the applicationConfiguration controller just emit a
+     `ManualScalerTrait` CR, it does not contain any information on the `example-db` workload.
```
negz, line 78: Currently the AppConfig controller assumes all traits will support a `spec.workloadRef` object, and writes said object to each rendered trait.

https://github.com/crossplane/oam-kubernetes-runtime/blob/6b9e04/pkg/controller/oam/applicationconfiguration/apply.go#L68

wonderflow:
> If the trait is OAM-aware I think it's better to encode this reference in a well known schemafied spec field, rather than opaque annotations, so the API server can validate it.  If the trait is OAM-unaware, I don't see how it could be aware of the annotations. 

@negz  I think the question here is that the AppConfig controller (oam-k8s-runtime) won't understand the schema of these different kind of traits, while the annotation is always the same.

ryanzhang-oss:
> but I'm not convinced there's a real need to move the current `spec.workloadRef` into an annotation on traits.

I think the problem is if we don't move it to annotation then we are going to force **ALL** the traits to have this `workloadRef` fields in their CRD. Are you saying that we won't allow OAM user to bring in any existing CRDs? Isn't the whole purpose of spec v1alpha2 to seamlessly integrate existing operators?

negz:
> I think the question here is that the AppConfig controller (oam-k8s-runtime) won't understand the schema of these different kind of traits, while the annotation is always the same.

@wonderflow To be clear, the current implementation of the AppConfig controller doesn't understand the whole schema of a trait either - it doesn't know anything about `ManualScalerTrait` for example. It requires the following two things for a CR to be a trait:

1. A `TraitDefinition` exists and points at the CRD that defines the CR.
2. The CR has a `spec.workloadRef` field.

> I think the problem is if we don't move it to annotation then we are going to force ALL the traits to have this workloadRef fields in their CRD. Are you saying that we won't allow OAM user to bring in any existing CRDs?

@ryanzhang-oss My intuition is that it will be less common for folks to bring an existing CR to use as a trait than it will be to bring an existing CR to use as a workload. That said, I do think it is good to allow folks to bring an existing CR to use as a trait if that use case does exist for them.

What I am having trouble following is that it seems to me like this proposal is saying:

1. We can't assume that trait CRs have a `spec.workloadRef`, because the trait might be an existing Kubernetes CR that was not designed with OAM in mind. This CR is reconciled by a controller that knows nothing about OAM.
1. We'll solve this problem by providing a library that will "help a trait developer locate the resources for the trait to modify", and that said library will "get the corresponding `workload` instance from the Kubernetes cluster with the information contained in the annotation field inserted by the application controller in the `trait` CR."

How can both of the above be true? Let's use the upstream [`HorizontalPodAutoscaler`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#horizontalpodautoscaler-v1-autoscaling) as an example of an existing Kubernetes CR that may make sense to use as a trait. The HPA controller doesn't know it's a trait, so it can't use the library mentioned in point 2 to determine what workload it applies to. It can only use its `spec.scaleTarget` to pick what it applies to.

Can you help me understand what it would look like to bring an existing Kubernetes CR like an HPA as an OAM trait? I feel like I may be missing something that would be much easier to understand if this proposal was grounded in some real world use cases demonstrating how one or two Kubernetes CRs that exists today (HPA? Ingress?) could be used as an OAM trait.

wonderflow:
@negz  Yeah, it makes sense. I think we could make `spec.workloadRef` in our mechanism and make it optional. So existing CRD can upgrade and support this feature as an OAM trait, just like they support the annotation.

resouer:
>  how one or two Kubernetes CRs that exists today (HPA? Ingress?) could be used as an OAM trait.

@negz I think this has already been answered in original issue: https://github.com/crossplane/oam-kubernetes-runtime/issues/4#issuecomment-617845649

In general, there're many trait CR does not always explicitly refer to workload by GVK, they could use endpoint URL, label selector, revision name etc instead. A very quick example is a Service CR which is a natural candidate of trait:

```
apiVersion: v1
kind: Service
metadata:
  name: my-service
spec:
  selector:
    app: MyApp
  ports:
    - protocol: TCP
      port: 80
      targetPort: 9376
```

So the first issue is we don't want to have WorkloadRef as a requirement. Seems we all agree on this?

The second issue is WorkloadRef vs annotations in the case of implementing your own trait. We don't have strong opinion here and let's pick the one with best experience.

negz:
> I think this has already been answered in original issue: crossplane/oam-kubernetes-runtime#4 (comment)

@resouer Right, I have seen this example but it doesn't help with what I am trying to understand, which is the value of moving `spec.workloadRef` into an annotation. I can follow the general idea of "it would be nice for any CR to be able to be a trait", but I cannot follow the specific technical detail of how the annotation helps with that, which is why I am asking for that to be illustrated.

I'm not very familiar with the etcd operator, but I would presume from reading your example that the `EtcdBackup` controller would ignore any OAM related annotations and just use the `spec.etcdEndpoints` field to figure out which cluster (i.e. "workload") to backup.

resouer:
@negz let's clarify the two issues clearer:

(1) allow "bring your own trait"

This is the original [issue](https://github.com/crossplane/oam-kubernetes-runtime/issues/4#issuecomment-617845649) mainly about, and seems we now agree its a valid use case and we won't enforce all trait have WorkloadRef field or anything similar.

(2) a helper for "implement your trait"

This is the proposal mainly about. The argument is around wether we should use WorkloadRef vs annotation as indexer for workload in trait CR. It's more like a UX design to me. If we prefer the structured approach, WorkloadRef is better and we will probably update the spec doc a bit to reflect this design as well. Otherwise, we can go to annotation which is more light-weighted but less structured. 

I think we can ask for more trait implementers (Alibaba, ByteDance, Tencent etc) for some more input. /cc @ryanzhang-oss

ryanzhang-oss:
> This is the original [issue](https://github.com/crossplane/oam-kubernetes-runtime/issues/4#issuecomment-617845649) mainly about, and seems we now agree its a valid use case and we won't enforce all trait have WorkloadRef field or anything similar. Examples need to be added in the proposal.

Echo this. I wonder if we can agree on this point?

> If we prefer the structured approach, WorkloadRef is better and we will probably update the spec doc a bit to reflect this idea as well.

I think we have talked about that in the last OAM community meeting. I would prefer to start with annotation and then move to more structured approaches later (and probably into spec too)

resouer:
For trait to workload ref, we need to make final decision at day 0 since it’s a essential implementation detail for crossplane, not a oam feature. The rule of “api promotion” does not apply here, even we go to workloadRef, it’s essentially a practice for k8s implementation, not part of spec, we just need to mention it somewhere.

negz:
> I wonder if we can agree on this point?

@ryanzhang-oss @resouer Yes, I see no issue with folks using existing Kubernetes CRs as traits.

> If we prefer the structured approach, WorkloadRef is better and we will probably update the spec doc a bit to reflect this idea as well.

I suggest we require `spec.workloadRef` for CRs that are written as OAM traits. I don't see any advantage of annotations over 'real' fields in this case. This is because:

* We can always influence the schema of these trait CRs.
* We can describe `spec.workloadRef` using a CRD, allowing us to document the fields via schema, and allowing the API server to validate them.

design/one-pager-trait-workload-interaction-mechanism.md
ryanzhang-oss: The annotation on the workload CR itself is for a `trait`. It seems that every trait needs to repeat what an appConfig controller does without it. We are hoping to save this part of the logic for the OAM core. Does it make sense?

negz: I agree it would be better to avoid having this logic distributed across all of the traits. I like the idea of having a library in `oam-kubernetes-runtime` that can help with this. My question is whether that library could grab this metadata from the `WorkloadDefinition` instead of an annotation.

I imagine this would look something like:

4. OAM runtime will provide a helper library. The library follows the following logic to help a trait developer locate the resources for the trait to modify.
    1. Read the `spec.workloadRef` of the trait to determine what kind of workload it applies to and fetch the `WorkloadDefinition` that defines the kind of workload.
    1. Read the `spec.childWorkloadKinds` of the `WorkloadDefinition`
    1. List each child resource by its GVK and filter by controller reference.

Does that make sense?

design/one-pager-trait-workload-interaction-mechanism.md
```diff
@@ -0,0 +1,203 @@
...
+        core.oam.dev/childResourceKinds: "[{"apiVersion":"apps/v1","kind":"Deployment"}]"
+    ```
+
+3. ApplicationConfig controller no longer assumes that all `trait` CRD contains a "spec
+.workloadRef" field conforms to the OAM definition. 
+
+4. OAM runtime will provide a helper library. The library follows the following logic to help a
+ trait developer locate the resources for the trait to modify.
+    1. Get the corresponding `workload` instance from the Kubernetes cluster with the information
+     contained in the annotation field inserted by the application controller in the `trait` CR.
```
negz, line 171: Are we okay with these being a spec field rather than annotations?

design/one-pager-trait-workload-interaction-mechanism.md
wonderflow: I think what @negz  says is fine, but the workloadRef should be optional.

ryanzhang-oss: yeah, that works. The only concern is that the workload to workloadDefinition part can be confusing so we wanted to hide that part from the user.

wonderflow:
image

wonderflow:
@jbw976  @negz @hasheddan All comments seem to be fixed, would you mind get this PR merge since we are developing this feature.

hongchaodeng:
After a second thought, this design proposal seems to be more related to https://github.com/crossplane/oam-kubernetes-runtime. Let's move this proposal into that repo.

ryanzhang-oss:
> After a second thought, this design proposal seems to be more related to https://github.com/crossplane/oam-kubernetes-runtime. Let's move this proposal into that repo.

I agree. https://github.com/crossplane/oam-kubernetes-runtime/pull/23

SevenYoung:
two problems:
1.  workloadDefinition name convention should be MUST, not RECOMMEND
2. trait may need "SpecReplicasPath" JSONPath like /scale subresouce

ryanzhang-oss:
> 1. workloadDefinition name convention should be MUST, not RECOMMEND
>    e.g: trait find resources to manipulate through component workload's WorkloadDefinition, if workloadDefinition name convention not meet, this won't work.

This is optional so the workloadDefinition author can choose to not follow the convention. Thus, it's not a MUST. I agree that this mechanism won't work if they don't follow but this is not the only way for a `trait` to work with a `workload`

ryanzhang-oss:
> 2\. trait may need "SpecReplicasPath" JSONPath like /scale subresource
>     e.g: ManualScaler can manipulate all workload with "spec.replicas" JSONPath, this is not explicit in spec.

I don't fully get what is "SpecReplicasPath" for.  Maybe you are talking about duck typing?

SevenYoung:
> > 2. trait may need "SpecReplicasPath" JSONPath like /scale subresource
> > e.g: ManualScaler can manipulate all workload with "spec.replicas" JSONPath, this is not explicit in spec.
> 
> I don't fully get what is "SpecReplicasPath" for. Maybe you are talking about duck typing?

i think trait need explicitly define what fieldPath it can control

Issue #1457, opened 2020-04-29
Connection secret and reclaim policy are only replicated during composition selection


negz:
### What happened?

I created a composite resource that explicitly specified a composition by setting its `spec.compositionRef`. I expected my composite resource to derive its `spec.reclaimPolicy` and `spec.writeConnectionSecretToRef` from its `Composition`, but it did not. I believe this is because [the code is never reached](https://github.com/crossplane/crossplane/blob/7ced24/pkg/controller/apiextensions/composite/api.go#L100) when a composition is explicitly specified.

### How can we reproduce it?

Define and instantiate a composite resource with an explicit `spec.compositionRef`.

### What environment did it happen in?
Crossplane version:

PR #1458, opened 2020-04-29
Replicate conn. secret and reclaim policy from composition

negz:

Fixes https://github.com/crossplane/crossplane/issues/1457

Previously we were only replicating these during composition selection, so they were not replicated if a composition was referenced explicitly.

Tested manually using the examples in https://github.com/crossplane/crossplane/pull/1459; confirmed that reclaim policy and connection secret ref are replicated even when the composite reference is specified explicitly.

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/pkg/controller/apiextensions/composite/api.go
+++ b/pkg/controller/apiextensions/composite/api.go
@@ -116,13 +116,39 @@ func (r *APISelectorResolver) ResolveSelector(ctx context.Context, cr resource.C
 		}
 
 		cr.SetCompositionReference(meta.ReferenceTo(comp.DeepCopy(), v1alpha1.CompositionGroupVersionKind))
-		cr.SetReclaimPolicy(comp.Spec.ReclaimPolicy)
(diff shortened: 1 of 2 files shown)


pkg/controller/apiextensions/composite/api.go
```diff
@@ -116,13 +116,39 @@ func (r *APISelectorResolver) ResolveSelector(ctx context.Context, cr resource.C
...
+}
+
+// An APIConfigurator configures a composite resource using its
+// composition.
+type APIConfigurator struct {
+	client client.Client
+}
+
+// Configure the supplied composite resource using its composition.
+func (c *APIConfigurator) Configure(ctx context.Context, cr resource.Composite, cp *v1alpha1.Composition) error {
```
hasheddan, line 138: This is close to something like [APIConnectionPropagator](https://github.com/crossplane/crossplane-runtime/blob/05ca13f8ea7b9b9270978337ad815b76a75101ee/pkg/resource/api.go#L78), I wonder if it would be worth adding this as a more generic implementation in `crossplane-runtime`?

PR #1486, opened 2020-05-06
Add priorityClassName option to helm chart

turkenh:
Signed-off-by: Hasan Turken <turkenh@gmail.com>

This PR introduces priorityClassName variable to the crossplane helm chart.

https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/

> Pods can have priority. Priority indicates the importance of a Pod relative to other Pods. If a Pod cannot be scheduled, the scheduler tries to preempt (evict) lower priority Pods to make scheduling of the pending Pod possible.

https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/#priorityclass

> A PriorityClass is a non-namespaced object that defines a mapping from a priority class name to the integer value of the priority.

With this option, we will be able to assign higher priorities to crossplane and stack-manager pods.

1. Create a priority class named `crossplane-system-critical`

```
apiVersion: scheduling.k8s.io/v1
kind: PriorityClass
metadata:
  name: crossplane-system-critical
value: 1000000
globalDefault: false
description: "Priority class for crossplane and stack-manager pods."
```

2. Install crossplane helm chart with `--set priorityClassName: "crossplane-system-critical"`

3. Verify that, both crossplane and stack-manager pods running with this priority class set.

### Checklist

I have:
  appropriate.

### DIFF
--- a/cluster/charts/crossplane-controllers/templates/deployment.yaml
+++ b/cluster/charts/crossplane-controllers/templates/deployment.yaml
@@ -21,6 +21,9 @@ spec:
         app: {{ template "name" . }}
         release: {{ .Release.Name }}
     spec:
+      {{- if .Values.priorityClassName }}
+      priorityClassName: {{ .Values.priorityClassName  | quote }}
+      {{- end }}
       {{- if not .Values.hostedConfig.enabled }}
       serviceAccountName: {{ template "name" . }}
       {{- end }}
--- a/cluster/charts/crossplane/templates/stack-manager-deployment.yaml
+++ b/cluster/charts/crossplane/templates/stack-manager-deployment.yaml
@@ -21,6 +21,9 @@ spec:
         app: {{ template "name" . }}-stack-manager
         release: {{ .Release.Name }}
     spec:
+      {{- if .Values.priorityClassName }}
+      priorityClassName: {{ .Values.priorityClassName | quote }}
+      {{- end }}
       serviceAccountName: stack-manager
       containers:
       - image: {{ .Values.image.repository }}:{{ .Values.image.tag }}
--- a/docs/getting-started/install.md
+++ b/docs/getting-started/install.md
@@ -310,6 +310,7 @@ and their default values.
 | `personas.deploy`                | Install roles and bindings for Crossplane user personas         | `true`
 | `templateStacks.enabled`         | Enable experimental template stacks support                     | `true`
 | `templateStacks.controllerImage` | Template Stack controller image                                 | `crossplane/templating-controller:v0.2.1`
+| `priorityClassName      `        | Priority class name for crossplane and stack manager pods       | `""`
 | `resourcesCrossplane.limits.cpu`        | CPU resource limits for Crossplane                       | `100m`
 | `resourcesCrossplane.limits.memory`     | Memory resource limits for Crossplane                    | `512Mi`
 | `resourcesCrossplane.requests.cpu`      | CPU resource requests for Crossplane                     | `100m`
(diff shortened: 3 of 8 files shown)


cluster/charts/crossplane/templates/stack-manager-deployment.yaml
```diff
@@ -21,6 +21,9 @@ spec:
         app: {{ template "name" . }}-stack-manager
         release: {{ .Release.Name }}
     spec:
+      {{- if .Values.priorityClassName }}
+      priorityClassName: {{ .Values.priorityClassName }}
```
hasheddan, line 25: Nitpick: I don't really know why we do it, but it looks like other strings we conditionally add get piped to `quote`. Do we want to do that here also? Or maybe we should remove it in other places?

docs/getting-started/install.md
```diff
@@ -310,6 +310,7 @@ and their default values.
 | `personas.deploy`                | Install roles and bindings for Crossplane user personas         | `true`
 | `templateStacks.enabled`         | Enable experimental template stacks support                     | `true`
 | `templateStacks.controllerImage` | Template Stack controller image                                 | `crossplane/templating-controller:v0.2.1`
+| `priorityClassName      `        | Priority class name for crossplane and stack manager pods       | `""`
```
hasheddan, line 313: Nitpick: Would you mind updating `forceImagePullPolicy` default value to also be `""` below for consistency? It looks much nicer than `` :)

turkenh: Sure, done.

hasheddan:
@turkenh looks like you need another `make generate` here

cluster/charts/crossplane/templates/stack-manager-deployment.yaml
turkenh: We are doing that if that field could get a value which could be parsed as `int` by yaml wheres actual type should be a `string`.  It might not really be needed where you don't expect an int as parameter but would not hurt to have it. So, I'll add it.

hasheddan: Awesome, thanks!

turkenh:
> @turkenh looks like you need another `make generate` here

@hasheddan just pushed it. Is it ok now?

hasheddan:
Looks good to me!

cluster/charts/crossplane-controllers/templates/deployment.yaml
```diff
@@ -21,6 +21,9 @@ spec:
         app: {{ template "name" . }}
         release: {{ .Release.Name }}
     spec:
+      {{- if .Values.priorityClassName }}
+      priorityClassName: {{ .Values.priorityClassName }}
```
jbw976, line 25: should we be consistent with the ` | quote` usage? this core crossplane deployment doesn't have it but the stack manager deployment does.

negz:
Nit: It's probably super obvious to folks who are up to speed with Kubernetes concepts, but it would be nice for folks like me who don't know exactly what a priority class does to include some context in the PR and/or commit description around _why_ we're adding this to the Helm chart.

cluster/charts/crossplane-controllers/templates/deployment.yaml
turkenh: Ah, I missed that. Done.

turkenh:
> Nit: It's probably super obvious to folks who are up to speed with Kubernetes concepts, but it would be nice for folks like me who don't know exactly what a priority class does to include some context in the PR and/or commit description around _why_ we're adding this to the Helm chart.

@negz added some context.

PR #1490, opened 2020-05-07
Make check-diff work across operating systems

hasheddan:

This builds on #1487 by guarding against the case where [some operating systems](https://stackoverflow.com/questions/30927590/wc-on-osx-return-includes-spaces/30927885#30927885) print leading whitespace for output of `wc -c`. In practice this target is usually only invoked in CI, but it useful to support it across all OS regardless.

Tested on both Linux and Mac OSX with modified / staged / untracked files.

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -161,7 +161,7 @@ reviewable: generate lint
 # Ensure branch is clean.
 check-diff: reviewable
 	@$(INFO) checking that branch is clean
-	@if [ $$(git status -s | wc -c) != "0" ]; then $(FAIL); fi;
+	@test -z "$$(git status --porcelain)" || $(FAIL)
 	@$(OK) branch is clean
 
 # integration tests


Makefile
```diff
@@ -161,7 +161,7 @@ reviewable: generate lint
 # Ensure branch is clean.
 check-diff: reviewable
 	@$(INFO) checking that branch is clean
-	@if [ $$(git status -s | wc -c) != "0" ]; then $(FAIL); fi;
+	@if [ $$(git status -s | wc -c | sed -e 's/^[ \t]*//') != "0" ]; then $(FAIL); fi;
```
negz, line 164: This is getting a little arcane. Could we take advantage of `git status --porcelain`, which is intended to be parsed by scripts?

```shell
$ git status
On branch rc
nothing to commit, working tree clean
$ git status --porcelain
$ touch moo
$ git status --porcelain
?? moo
$ echo HI > README.md 
$ git status --porcelain
 M README.md
?? moo
$ test -z "$(git status --porcelain)" && echo clean
$ git reset --hard
HEAD is now at 8ab0cb15 Final proof-reading pass
$ rmd moo
$ test -z "$(git status --porcelain)" && echo clean
clean
```

hasheddan: Using `--porcelain` seems fine. `git status -s` is [almost the same thing](https://www.stefanjudis.com/today-i-learned/the-short-version-of-git-status-and-the-close-but-different-porcelain-mode/), but definitely think using `test` is a big improvement.

hasheddan:
@negz can you take another look here?

PR #1493, opened 2020-05-09
Shorter composite and composed resource names

muvaf:

Currently, the name of the composed resources get to be way too long for some provider validations, like `crossplane-system-myrequirement-sdas2-dsdh4` while it's not necessary to include the namespace as prefix and also yet another random string as postfix. With this change:
* Requirement name: `myrequirement`
* Composite name: `myrequirement-dshn3`
* A composed resource name: `myrequirement-hdsn2`

However, in Kubernetes, children resources have the name of their parents as prefix like:
* Deployment: `deploymentname`
* ReplicaSet: `deploymentname-dsjn3ndshh`
* Pod: `deploymentname-dsjn3ndshh-idsjn`

So, I think we should keep following the convention for composed resource name and make it `myrequirement-dshn3-hdsn2`, however, this increases the likeliness of failure in name validation of managed resources since it's longer; see https://github.com/crossplane/provider-gcp/issues/222 and https://github.com/crossplane/stack-azure-sample/issues/26 . In practice, it's fine if we have only one rand string set but it's against the convention.

I think we should do the truncation for `external-name` annotation generation as proposed in https://github.com/crossplane/provider-gcp/issues/222#issuecomment-617680890 . So, I'll keep this in draft until we resolve that discussion. If we do the truncation, then it should be fine to keep the convention.

Fixes https://github.com/crossplane/crossplane/issues/1598

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/pkg/controller/apiextensions/composite/composed/api.go
+++ b/pkg/controller/apiextensions/composite/composed/api.go
@@ -39,10 +39,11 @@ const (
 	errNamePrefix = "name prefix is not found in labels"
 )
 
+// Label keys.
 const (
-	// LabelKeyNamePrefixForComposed is the prefix that will be used by composed
-	// resources.
 	LabelKeyNamePrefixForComposed = "crossplane.io/root-composite"
+	LabelKeyRequirementName       = "crossplane.io/requirement-name"
+	LabelKeyRequirementNamespace  = "crossplane.io/requirement-namespace"
 )
 
 // ConfigureFn is a function that implements Configurator interface.
@@ -70,7 +71,11 @@ func (*DefaultConfigurator) Configure(cp resource.Composite, cd resource.Compose
 		return errors.New(errNamePrefix)
 	}
 	// This label will be used if composed resource is yet another composite.
-	meta.AddLabels(cd, map[string]string{LabelKeyNamePrefixForComposed: cp.GetLabels()[LabelKeyNamePrefixForComposed]})
+	meta.AddLabels(cd, map[string]string{
+		LabelKeyNamePrefixForComposed: cp.GetLabels()[LabelKeyNamePrefixForComposed],
+		LabelKeyRequirementName:       cp.GetLabels()[LabelKeyRequirementName],
+		LabelKeyRequirementNamespace:  cp.GetLabels()[LabelKeyRequirementNamespace],
+	})
 	// Unmarshalling the template will overwrite any existing fields, so we must
 	// restore the existing name, if any. We also set generate name in case we
 	// haven't yet named this composed resource.
--- a/pkg/controller/apiextensions/composite/composed/api.go
+++ b/pkg/controller/apiextensions/composite/composed/api.go
@@ -41,7 +41,7 @@ const (
 
 // Label keys.
 const (
-	LabelKeyNamePrefixForComposed = "crossplane.io/root-composite"
+	LabelKeyNamePrefixForComposed = "crossplane.io/composite"
 	LabelKeyRequirementName       = "crossplane.io/requirement-name"
 	LabelKeyRequirementNamespace  = "crossplane.io/requirement-namespace"
 )
--- a/pkg/controller/apiextensions/composite/composed/api.go
+++ b/pkg/controller/apiextensions/composite/composed/api.go
@@ -25,6 +25,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/json"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/crossplane/crossplane-runtime/pkg/meta"
 	"github.com/crossplane/crossplane-runtime/pkg/reconciler/managed"
 	"github.com/crossplane/crossplane-runtime/pkg/resource"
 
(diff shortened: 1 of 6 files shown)


pkg/controller/apiextensions/composite/composed/api.go
```diff
@@ -51,17 +51,23 @@ type DefaultConfigurator struct{}
...
-	// restore the existing name, if any. We also set generate name in case we
-	// haven't yet named this composed resource.
+	// If composite resource was created via a requirement, it will already have
+	// a random string appended. Appending yet another random string increases
+	// the odds that the composed managed resource won't be provisioned because
+	// of long name that the provider wouldn't accept.
 	cd.SetGenerateName(cp.GetName() + "-")
+	if cp.GetRequirementReference() != nil && cp.GetRequirementReference().Name != "" {
+		cd.SetGenerateName(cp.GetRequirementReference().Name + "-")
+	}
```
hasheddan, line 68: Agree that this feels a little weird to sometimes name it one way and sometimes name it another.

pkg/controller/apiextensions/composite/composed/api.go
```diff
@@ -51,17 +51,23 @@ type DefaultConfigurator struct{}
...
 	cd.SetName(name)
+	cd.SetNamespace(namespace)
```
hasheddan, line 70: Is this change related to naming or is this just updating so we are not creating in the `default` namespace?

muvaf: It's not related to naming, that just enables namespaced resources to be included in the composition. I noticed it while making that change.

negz:
@muvaf and I discussed this out of band today, and I wanted to capture an idea we had here.

Historically we've automatically named dynamically provisioned managed resources (which are cluster scoped) `claimnamespace-claimname-randm`. The thinking being that it makes it easier to tell where dynamically provisioned resources came from at a glance when running (e.g.) `kubectl list cloudsqlinstances` or browsing the list of CloudSQL resources in the GCP console.

This makes sense in some ways because the name reads left-to-right and thus sorts from least to most specific. Unfortunately it makes the name more likely to need truncation (because including the namespace makes it longer), and harder to truncate cleanly (because we'd truncate the most specific part of the name from the right).

With composition we face the added concern that composites may be nested - is a composite resource can compose a composite resource. So if we stuck with Kubernetes convention we could end up with a name like `reqnamespace-reqname-djm2f-os90d-m12kf-mdns2`.

I wonder if we can make use of labels to help with this? Perhaps something like:

* Statically provisioned composite resources are named whatever the user chooses.
* Dynamically provisioned composite resources are named `requirementname-randm`.
* All composed resources are labelled `crossplane.io/composite: compositename` where `compositename` is the name of the root/outermost composite resource.
* All composed resources are named `compositename-randm` where `compositename` is the name of the root/outermost composite resource.
* All composed resources of a dynamically provisioned composite resource are labelled `crossplane.io/requirement: reqnamespace-reqname`.

I believe this would give us the following properties:

* Composite and composed resource names would never collide (at the API server level).
* It would be possible to use `kubectl get -l crossplane.io/requirement` and `kubectl get -l crossplane.io/composite` to list resources that were composed by a particular composite and/or requirement.
* Composed resource names would include at most 12 random chars. A requirement named `mycooldb` would result in a composite named something like `mycooldb-fk1nd`, whose composed resources (no matter how nested) would be named something like `mycooldb-fk1nd-3d01d`.

pkg/controller/apiextensions/composite/composed/api.go
```diff
@@ -25,16 +25,24 @@ import (
...
+	errUnmarshal  = "cannot unmarshal base template"
+	errFmtPatch   = "cannot apply the patch at index %d"
+	errGetSecret  = "cannot get connection secret of composed resource"
+	errNamePrefix = "name prefix is not found in labels"
+)
+
+const (
+	// LabelKeyNamePrefixForComposed is the prefix that will be used by composed
+	// resources.
+	LabelKeyNamePrefixForComposed = "crossplane.io/root-composite"
```
muvaf, line 45: I'm on the fence about the value of this. `composite` didn't make much sense to me as the leafs are actually member of multiple composites but what we'd like to say it's the one that started it all. Made it `name-prefix` at one point and that's what it is technically but didn't seem like a very good name from UX perspective.

muvaf:
> Composed resource names would include at most 12 random chars. A requirement named mycooldb would result in a composite named something like mycooldb-fk1nd, whose composed resources (no matter how nested) would be named something like mycooldb-fk1nd-3d01d.

I have implemented this. All composite resources will first check whether they got the prefix label or not. If not, they'll label themselves with their name and use that label as prefix. This enables recursive application of the prefix.

muvaf:
Now all of the members of a composition has got `crossplane.io/requirement-name` and `crossplane.io/requirement-namespace` labels. @negz should be ready for another look.

All manually tested.

pkg/controller/apiextensions/composite/composed/api.go
```diff
@@ -54,14 +63,25 @@ func (*DefaultConfigurator) Configure(cp resource.Composite, cd resource.Compose
 	// Any existing name will be overwritten when we unmarshal the template. We
 	// store it here so that we can reset it after unmarshalling.
 	name := cd.GetName()
+	namespace := cd.GetNamespace()
```
negz, line 66: Will a composed resource ever have a namespace?

pkg/controller/apiextensions/composite/composed/api.go
negz: I lean toward keeping this as just `crossplane.io/composite` and living with the fact that it will always be the outermost composite. I'm not yet sure how frequent nested composites will even be. That said, I don't feel super strongly about the name.

muvaf: > That said, I don't feel super strongly about the name.

Yeah, me neither. I'll just make it `crossplane.io/composite`

pkg/controller/apiextensions/composite/composed/api.go
muvaf: It may or may not but I wanted that logic to be complete; we're setting the identifier in case template has something else and the identifier that can be overwritten is `namespace + name` here, not necessarily name only. UID could be part of the template but that's a really far away case to cover.

PR #1494, opened 2020-05-09
Add kind as one of the local Kubernetes development environment tool

muvaf:

It adds a script that has the same interface with `minikube.sh` to make `kind` easier to use for testing Crossplane.

I wanted to use kind for a long time because it's faster than Minikube and doesn't require a VM. However, it wasn't suitable for local manual Crosssplane tests since the restart/sleep of the computer would kill the cluster(see https://github.com/kubernetes-sigs/kind/issues/148), hence loose CRs which results in leaked cloud resources. In [v0.8.0](https://github.com/kubernetes-sigs/kind/releases/tag/v0.8.0), this has been fixed. @BenTheElder thank you so much for the fix!

`make`
`cluster/local/kind.sh up`
`cluster/local/kind.sh helm-install`

Crossplane came up. Created a few CRs.

*restarted the computer*

Crossplane and CRs are still there!

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/cluster/local/README.md
+++ b/cluster/local/README.md
@@ -1,29 +1,51 @@
 # Crossplane Local Deployment and Test
 
-The Local Framework is used to run end to end and integration tests on Crossplane. The framework depends on a running instance of Kubernetes.
-The framework also provides scripts for starting Kubernetes using `minikube` so users can
-quickly spin up a Kubernetes cluster.The Local framework is designed to install Crossplane, run tests, and uninstall Crossplane.
-
-## Requirements
-1. Docker version => 1.2 && < 17.0
-2. Ubuntu 16, Fedora 29 (the framework has only been tested on these versions of linux distributions)
-3. Kubernetes with kubectl configured
-4. Crossplane
+The Local Framework is used to run end to end and integration tests on Crossplane.
+The framework depends on a running instance of Kubernetes. The framework also
+provides scripts for starting a local Kubernetes using `kind`, `minikube` and `microk8s`
+so users can quickly spin up a Kubernetes cluster.The Local framework is designed
+to install Crossplane, run tests, and uninstall Crossplane.
 
 ## Instructions
 
 ## Setup
 
 ### Install Kubernetes
-You can choose any Kubernetes flavor of your choice.  The test framework only depends on kubectl being configured.
-The framework also provides scripts to install Kubernetes.
+You can choose any Kubernetes flavor of your choice.  The test framework only
+depends on kubectl being configured. The framework also provides scripts to install Kubernetes.
 
-- **Minikube** (recommended for MacOS): Run [minikube.sh](./minikube.sh) to setup a single-node Minikube Kubernetes.
+- **kind**: Run [kind.sh](./kind.sh) to setup a single-node Minikube Kubernetes.
+  - kind v0.8.0 and higher is supported
+- **Minikube**: Run [minikube.sh](./minikube.sh) to setup a single-node Minikube Kubernetes.
   - Minikube v0.28.2 and higher is supported
-- **MicroK8s** (recommended for Linux): Run [microk8s.sh](./microk8s.sh) to setup a single-node Microk8s Kubernetes.
+- **MicroK8s**: Run [microk8s.sh](./microk8s.sh) to setup a single-node Microk8s Kubernetes.
   - MicroK8s v1.14 (move from dockerd to containerd) and higher is supported
 
-#### Minikube (recommended for MacOS)
+#### Kind
+Starting the cluster on kind is as simple as running:
+```console
+cluster/local/kind.sh up
+```
+You can check the default Kubernetes version in the script and override it.
+
+To copy Crossplane image generated from your local build into the local registry in
+the kind cluster and install it via Helm, run the following commands after
+`kind.sh up` succeeded:
+```console
+cluster/local/kind.sh helm-install
+```
+
+Stopping the cluster and destroying the kind cluster can be done with:
+```console
+cluster/local/kind.sh clean
+```
+
+For complete list of subcommands supported by `kind.sh`, run:
+```console
+cluster/local/kind.sh
+```
+
+#### Minikube
 Starting the cluster on Minikube is as simple as running:
 ```console
 cluster/local/minikube.sh up
@@ -44,7 +66,7 @@ For complete list of subcommands supported by `minikube.sh`, run:
 cluster/local/minikube.sh
 ```
 
-#### MicroK8s (recommended for Linux)
+#### MicroK8s
 Starting the cluster on MicroK8s is as simple as running:
 ```console
 cluster/local/microk8s.sh up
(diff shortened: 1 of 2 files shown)


cluster/local/README.md
```diff
@@ -93,20 +115,24 @@ deleting or scaling down the `crossplane` deployment in the namespace
 `crossplane-system`.
 
 ## Run Tests
-The following sections provide commands helpful for development environments. `minikube.sh` may be replaced with [`local.sh`](./local.sh) or [`microk8s.sh`](./microk8s.sh) in those environments. These commands expect to be run from the project root.
+The following sections provide commands helpful for development environments.
+`kind.sh` may be replaced with [`local.sh`](./local.sh), [`microk8s.sh`](./microk8s.sh),
+or [`minikube.sh`](./kind.sh) in those environments. These commands expect to be
```
hasheddan, line 120:
```suggestion
or [`minikube.sh`](./minikube.sh) in those environments. These commands expect to be
```

cluster/local/README.md
```diff
@@ -119,17 +145,20 @@ make test
...
 #### 7. Stop Kubernetes
 ```
-cluster/local/minikube.sh down
+cluster/local/kind.sh down
 ```
 
 ## Run Integration Tests
-In addition to the `minikube.sh` tests described above which gives some flexibility on interacting with the cluster, one could also test the integration of the built artifacts with the cluster in a single command.
-This will create a new Kubernetes cluster using `kind` tool, and then installs Crossplane and checks a few assertions, and finally destructs the cluster.
+In addition to the `minikube.sh` tests described above which gives some
```
hasheddan, line 157:
```suggestion
In addition to the `kind.sh` tests described above which gives some
```

cluster/local/kind.sh
```diff
@@ -0,0 +1,107 @@
...
+}
+
+# Deletes pods with application prefix. Namespace is expected as the first argument
+function delete_pods() {
+    for pod in $(kubectl get pods -n "$2" -l "app=$1" --no-headers -o custom-columns=NAME:.metadata.name); do
+        kubectl delete pod "$pod" -n "$2"
+    done
+}
+
+# current kubectl context == minikube, returns boolean
```
hasheddan, line 32:
```suggestion
# current kubectl context == kind, returns boolean
```

cluster/local/kind.sh
```diff
@@ -0,0 +1,107 @@
...
+function check_context() {
+    if [ "$(kubectl config view 2>/dev/null | awk '/current-context/ {print $NF}')" = "kind-kind" ]; then
+        return 0
+    fi
+
+    return 1
+}
+
+# configure minikube
```
hasheddan, line 41:
```suggestion
# configure kind
```

cluster/local/kind.sh
```diff
@@ -0,0 +1,107 @@
...
+KUBE_IMAGE=${KUBE_IMAGE:-"kindest/node:v1.15.11@sha256:6cc31f3533deb138792db2c7d1ffc36f7456a06f1db5556ad3b6927641016f50"}
```
hasheddan, line 42: It looks like you hardcode 1.15 image below so I don’t think setting this variable would actually change anything

cluster/local/kind.sh
```diff
@@ -0,0 +1,107 @@
...
+scriptdir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+# shellcheck disable=SC1090
+projectdir="${scriptdir}/../.."
+
+# get the build environment variables from the special build.vars target in the main makefile
+eval $(make --no-print-directory -C ${scriptdir}/../.. build.vars)
+
+BUILD_IMAGE="${BUILD_REGISTRY}/${PROJECT_NAME}-amd64"
+MINIKUBE_IMAGE="${DOCKER_REGISTRY}/${PROJECT_NAME}:master"
```
hasheddan, line 14: Not a functional change, but kind of confusing that this is called MINIKUBE_IMAGE

muvaf:
@hasheddan Thanks for the review! Addressed all your comments.

cluster/local/kind.sh
```diff
@@ -0,0 +1,103 @@
...
+
+# configure kind
+KUBE_IMAGE=${KUBE_IMAGE:-"kindest/node:v1.15.11@sha256:6cc31f3533deb138792db2c7d1ffc36f7456a06f1db5556ad3b6927641016f50"}
+
+case "${1:-}" in
+  up)
+    kind create cluster --image "${KUBE_IMAGE}" --wait 5m
+
+    kubectl apply -f ${scriptdir}/helm-rbac.yaml
+    ${HELM} init --service-account tiller
```
hasheddan, line 49: Note to self: update `build` with helm 3

cluster/local/README.md
```diff
@@ -1,29 +1,51 @@
...
 
 ### Install Kubernetes
-You can choose any Kubernetes flavor of your choice.  The test framework only depends on kubectl being configured.
-The framework also provides scripts to install Kubernetes.
+You can choose any Kubernetes flavor of your choice.  The test framework only
+depends on kubectl being configured. The framework also provides scripts to install Kubernetes.
 
-- **Minikube** (recommended for MacOS): Run [minikube.sh](./minikube.sh) to setup a single-node Minikube Kubernetes.
+- **kind**: Run [kind.sh](./kind.sh) to setup a single-node Minikube Kubernetes.
+  - kind v0.8.0 and higher is supported
```
hasheddan, line 18: Was about to comment on that we use `kind` v0.7.0 in `build`, but then I noticed that you are using whatever `kind` version the user decides to install for these cases, which I think makes sense for local dev scenarios. I also wonder if we could clean-up how we vendor tools in https://github.com/upbound/build/blob/e806fbf89e0c6e13cac0331f4cbdc1bf53ba06c6/makelib/k8s_tools.mk. I would like to move to pinning tools in the same way that we do `controller-gen` if possible. Anyway, that is a conversation for another day :sweat_smile:

cluster/local/kind.sh
muvaf: Actually I though about this and then just made the call that as developers it might be a better idea to use helm2 and kubernetes 1.15 since we're still supporting them so that we can see if our changes break that compatibility. Or maybe it was just finding a reason for procrastination haha.

cluster/local/README.md
muvaf: Yeah, theoretically all versions should be OK but for local dev 0.8.0 is much safer IMO thanks to not wiping out the cluster after restarts.

negz:
> Add kind as one of the local Kubernetes development environment tool

Would you say kind was one _kind_ of local Kubernetes development environment tool? :trollface:

negz:
Nice! Looking forward to trying this out.

muvaf:
> Would you say kind was one kind of local Kubernetes development environment tool?

@negz I actually find myself pronouncing it like `kin-dee` since it's kubernetes-in-docker. Then correct myself with some embarrassment and say `kind`..

PR #1498, opened 2020-05-12
Add master CIDR block field to GKE composition example 

muvaf:

Add master CIDR block field to GKE composition example to enable multiple clusters to be created. We can't embed into the composition because it shouldn't overlap with other master CIDR blocks.

@negz FWIW, I think we've been taking the approach of specifying the provider on class, i.e. composition, but in this example we take it from the namespaced user. Given infra-ops are the ones who make decisions about which account to use, I'd say it's better in `Composition` but I didn't want to block this PR, so it's not included.

Manually by created 2 instances.

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/cluster/examples/composition/gcp-gkecluster/cluster.yaml
+++ b/cluster/examples/composition/gcp-gkecluster/cluster.yaml
@@ -11,9 +11,10 @@ spec:
   - us-central1-a
   - us-central1-b
   cidrs:
-    nodes: 192.168.0.0/24
-    pods: 10.128.0.0/20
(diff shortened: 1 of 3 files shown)


muvaf:
@negz should be ready to merge, I'm able to connect to the cluster.

muvaf:
@negz I have disabled the private access since it blocked me from accessing the cluster via kubeconfig and added `masterAuth.username: admin` so that kubeconfig is meaningful for local tests and confirmed that I can connect to the cluster with that kubeconfig.

PR #1503, opened 2020-05-14
Document defining and publishing composite infrastructure

negz:

This PR documents the process of defining, publishing, and consuming composite infrastructure resources for the impending Crossplane 0.11 release. I took a few liberties compared to [the proposed outline](https://docs.google.com/document/d/16mLVuOK-T-hPA7QFMuCJg2W7CnwFghuEERuq1L47l70/edit#heading=h.hz7neu6xuv3), but it's followed in spirit.

Note that you can see this document rendered correctly (including the diagram) at https://github.com/negz/crossplane/blob/compositionexposition/docs/composition.md

I've applied all of the YAML examples in the documentation to confirm they'd actually work if someone applied them - presuming that someone had an Azure provider named `example` configured.

### Checklist

I have:
  appropriate.

[definition of done].


docs/composition.md
```diff
@@ -0,0 +1,882 @@
...
+> [Current Limitations] for information on functionality that is planned but not
+> yet implemented.
+
+## Concepts
+
+![Infrastructure Composition Concepts]
+
+A _Composite_ infrastructure resource is composed of other resources. Its
+configuration schema is user-defined. The `MySQLInstance` resources in the above
+diagram are composite infrastructure resources.
```
kasey, line 41: Why is `MySQLInstance` pink ("Authored by App Ops") on the Azure side and lavender ("Authored by Infra Ops") on the AWS side?

kasey: Speaking to Nic just now in a zoom call we talked about the pink v lavender boxes denoting dynamic vs static provisioning. I recommended creating a second diagram to get that point across so that this diagram can stay more focused on showing the relationships between the different layers of the composition structure.

docs/composition.md
```diff
@@ -0,0 +1,882 @@
...
+
+A `Composition` specifies a set of resources of which a composite infrastructure
+resource may be composed. The Azure and AWS `Composition` resources in the above
```
kasey, line 44: The `Composition` CRD connects the "composite infrastructure resource", eg `MySQLInstance`, to the provider-specific set of managed resources needed to fulfill the promise of that resource, eg in the case of AWS, `DBSubnetGroup` + `RDSInstance`. The Composition also specifies how to copy values needed by these managed resources from the "composite infrastructure resource", using the `fromFieldPath` -> `toFieldPath` keywords.

docs/composition.md
```diff
@@ -0,0 +1,882 @@
...
+diagram configure two possible compositions of a `MySQLInstance`.
+
+A _Requirement_ for an infrastructure resource declares that an application
```
kasey, line 47: I think you want to convey that a Requirement is how an *application* specifies what infrastructure it needs, but with `for an infrastructure resource` confuses that point, I'd just snip it out leaving: A _Requirement_ declares that an application requires a particular kind and configuration of composite infrastructure resource.

kasey: It sounds like `MySQLInstanceRequirement` is also providing the values for the `MySQLInstance` it requires, but `configuration` almost makes it sound like there is yet another component. Would it make sense to say that the Requirement *configures* the composite infrastructure resource?

docs/composition.md
```diff
@@ -0,0 +1,882 @@
...
+requires a particular kind and configuration of composite infrastructure
+resource. The `MySQLInstanceRequirement` resources in the above diagram declare
+that the application pods each require a `MySQLInstance`.
+
+An `InfrastructureDefinition` defines a new kind of composite infrastructure
```
kasey, line 52: Could we use the term `InfrastructureDefinition` in place of "composite infrastructure resource" in this document? Could make the connections more clear and give us proper nouns to build understanding around.
"A _Composite_ infrastructure resource is composed of other resources. Its
configuration schema is user-defined" by an `InfrastructureDefinition`. In the diagram above, `MySQLInstance` has been defined by the `MySQLInstance` `InfrastructureDefinition`."

docs/composition.md
```diff
@@ -0,0 +1,882 @@
...
+resource. The `InfrastructureDefinition` in the above diagram defines the
+existence and schema of the `MySQLInstance` composite infrastructure resource.
+
+An `InfrastructurePublication` publishes a defined composite infrastructure
+resource, allowing application operators to declare a requirement for it. The
+`InfrastructurePublication` in the above diagram allows application operators to
+author a `MySQLInstanceRequirement`.
```
kasey, line 59: This paragraph leaves me wondering why Publications exist? Maybe we could explain why creating an `InfrastructureDefinition` doesn't inherently publish it, or why there isn't just a `published` flag on the definition.

docs/composition.md
```diff
@@ -0,0 +1,882 @@
...
+  composite infrastructure resource onto each "base".
+
+Multiple compositions may satisfy a particular kind of composite infrastructure
+resource, and the author of a composite resource (or resource requirement) may
+select which composition will be used. This allows an infrastructure operator to
+offer their application operators a choice between multiple opinionated classes
+of infrastructure, allowing them to explicitly specify only some configuration.
+An infrastructure operator may offer their application operators the choice
+between an "Azure" and a "GCP" composition when creating a `MySQLInstance` for
+example, Or they may offer a choice between a choice between a "production" and
```
kasey, line 250: a choice between ~a choice between~

docs/composition.md
```diff
@@ -0,0 +1,882 @@
...
+> Note that composition is an **experimental** feature of Crossplane. Refer to
+> [Current Limitations] for information on functionality that is planned but not
+> yet implemented.
+
+## Concepts
+
+![Infrastructure Composition Concepts]
+
+A _Composite_ infrastructure resource is composed of other resources. Its
+configuration schema is user-defined. The `MySQLInstance` resources in the above
```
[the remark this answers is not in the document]
negz: Yeah, I struggle with this too. Something like:

"the schema of the configuration"
"the configuration fields the resource supports"
"the fields it supports are user defined"

Any of those an improvement?

docs/composition.md
negz: Pushed a change to address this.

docs/composition.md
```diff
@@ -0,0 +1,882 @@
...
+infrastructure resources like Azure SQL servers and AWS ElastiCache clusters.
+Infrastructure composition allows infrastructure operators to define, share, and
+reuse new kinds of infrastructure resources that are _composed_ of other
+infrastructure resources. Infrastructure operators may configure one or more
+compositions of any defined resource, and may publish any defined resource to
+their application operators, who may then declare that their application
+requires that kind of resource.
+
+Composition can be used to build a catalogue of kinds and configuration classes
+of infrastructure that fit the needs and opinions of your organisation. As an
```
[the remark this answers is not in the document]
negz: Working as intended. 🇬🇧

docs/composition.md
```diff
@@ -0,0 +1,882 @@
+---
+title: Publishing Infrastructure
+toc: true
+weight: 400
+indent: true
+---
+
+# Publishing Infrastructure
```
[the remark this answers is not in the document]
negz: I believe it was @bassam's preferred term of the "define, compose, publish" trio, but I don't think any of us are super tied to it.

Possibly the tricky thing here is that we're covering several related ideas:

* Crossplane lets you (an infra op) define your own infrastructure resource types.
* Crossplane lets you choose which types you publish to your users (app ops).
* The resources you define are reconciled by composing one or more other resources.
* You can offer your users more than one composition of resources.

Would "Publish Infrastructure Abstractions" or "Define Infrastructure" be an improvement?

docs/composition.md
negz: What about "compositions configure composite composition?" :trollface:

docs/composition.md
negz: We've had similar thoughts, especially because using a separate publication resource adds some hairy implementation details - e.g. you need to delete the publication before deleting the definition.

We've already found some things that we want to configure at the InfrastructureDefinition level that aren't just a CRD template - e.g. the connection secret details a particular kind of infrastructure needs. No such things exist for publications today, but if we think they might in future it's probably worth keeping it as a distinct resource. If we don't, I'd err toward making publication a flag on the InfrastructureDefinition. @muvaf has pointed out that we could let RBAC decide whether folks should be able to author a requirement, but I like the UX of being able to `kubectl get infrastructurepublications` to discover which resources are published and which are not. `kubectl get infrastructuredefinitions` with a `PUBLISHED` column would likely be just as good.

Either way, I'm not sure user facing documentation is the right place for us to be justifying the value/existence of a concept. It exists, and it will exist for at least Crossplane 0.11, so I think we should document what it does today.

docs/composition.md
negz: I like the idea of including an example of how a `MySQLInstance` gets reconciled with a particular set of composed resources - will push an update.

docs/composition.md
negz: True - pushed an update that tries to make this more explicit.

docs/composition.md
negz: In short, I'd rather not. In this case the thing and the definition of the thing are distinct concepts that I'd rather avoid conflating. I do take your point that "composite infrastructure resource" is a mouthful and that something more pithy would be ideal, though.

docs/composition.md
```diff
@@ -0,0 +1,887 @@
...
+
+Crossplane allows infrastructure operators to define new kinds of infrastructure
+resources and publish them for the application operators they support to use,
+all without writing any code.
+
+Infrastructure providers enable Crossplane to manage a wide array of
+infrastructure resources like Azure SQL servers and AWS ElastiCache clusters.
+Infrastructure composition allows infrastructure operators to define, share, and
+reuse new kinds of infrastructure resources that are _composed_ of other
```
prasek, line 17:
```suggestion
new kinds of infrastructure resources that can hide infrastructure complexity and include policy guardrails, so applications can easily and safely consume the infrastructure they need, that conforms to your organizational best-practices.
```

docs/composition.md
```diff
@@ -0,0 +1,887 @@
...
+infrastructure resources. Infrastructure operators may configure one or more
```
prasek, line 18:
```
Infrastructure operators may configure one or more
```

docs/composition.md
```diff
@@ -0,0 +1,887 @@
...
+toc: true
+weight: 400
+indent: true
+---
+
+# Publishing Infrastructure
+
+Crossplane allows infrastructure operators to define new kinds of infrastructure
+resources and publish them for the application operators they support to use,
+all without writing any code.
```
prasek, line 12:
```suggestion
all without writing any code. These composite infrastructure resources can be consumed alongside any Kubernetes application to provision and manage the cloud services they require.
```

docs/composition.md
```diff
@@ -0,0 +1,887 @@
...
+
+Infrastructure providers enable Crossplane to manage a wide array of
+infrastructure resources like Azure SQL servers and AWS ElastiCache clusters.
```
prasek, line 15:
```suggestion
infrastructure resources like Azure SQL servers and AWS ElastiCache clusters, however infrastructure primitives (managed resources) represent a single resource and as such don't support grouping related resources together so they can be provisioned as a single unit.
```

docs/composition.md
```diff
@@ -0,0 +1,887 @@
...
+title: Publishing Infrastructure
+toc: true
+weight: 400
+indent: true
+---
+
+# Publishing Infrastructure
+
+Crossplane allows infrastructure operators to define new kinds of infrastructure
+resources and publish them for the application operators they support to use,
```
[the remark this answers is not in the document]
negz: Went with a hybrid of the original wording and your suggestions:

> Crossplane allows infrastructure operators to define and compose new kinds of
infrastructure resources then publish them for the application operators they
support to use, all without writing any code.

I like that this makes the subtle distinction that you don't _have to_ publish the infrastructure you define and compose.

docs/composition.md
```diff
@@ -0,0 +1,887 @@
...
+all without writing any code.
+
+Infrastructure providers enable Crossplane to manage a wide array of
```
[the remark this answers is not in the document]
negz: I feel like the available kinds of providers are implied by the two examples (Azure, AWS) that immediately follow. I did reword this to clarify that "provider" in this context means our extensions, not the actual clouds.

PR #1504, opened 2020-05-14
Core InfrastructureDefinitions

muvaf:

Adds correspondents of today's claims as `InfrastructureDefinition`s so that we have a base set of definitions that common/basic scenarios are handled. It'd also help app devs to package those base `*Requirement` resources to specify the type of the infra they need and the app operators could change them with their own definition of that type of infra.

It has been tested with Helm 2 and install/uninstall works fine. I am still figuring out how to make Helm3 delete the `InfrastructureDefinition`s and `InfrastructurePublication`s before deleting the controllers.

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/cluster/charts/kustomize_patches.yaml
+++ b/cluster/charts/kustomize_patches.yaml
@@ -1,0 +1,11 @@
+# We need InfrastructurePublications to be deleted first because if two resources
+# with pre-delete hooks have the same hook weight, they are deleted in alphabetic
+# order. So, Helm tries to delete InfrastructureDefinitions but they cannot be
+# deleted if they are published. Hence publications need to be deleted first.
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    "helm.sh/hook-weight": "-1"
+  name: infrastructurepublications.apiextensions.crossplane.io
--- a/Makefile
+++ b/Makefile
@@ -95,7 +95,7 @@ CRD_DIR = $(CROSSPLANE_TYPES_CHART_DIR)/crds
 CROSSPLANE_CHART_HELM2_CRD_DIR = $(CROSSPLANE_CHART_DIR)/templates/crds
 CROSSPLANE_CHART_HELM3_CRD_DIR = $(CROSSPLANE_CHART_DIR)/crds
 
-TYPE_MANIFESTS = $(wildcard $(CROSSPLANE_TYPES_CHART_DIR)/templates/*.yaml)
+TYPE_MANIFESTS = $(filter-out $(wildcard $(CROSSPLANE_TYPES_CHART_DIR)/templates/*.yaml), $(wildcard $(CROSSPLANE_TYPES_CHART_DIR)/templates/definitions/*.yaml))
 CONTROLLER_MANIFESTS = $(filter-out $(wildcard $(CROSSPLANE_CONTROLLERS_CHART_DIR)/templates/stack-manager-host-*.yaml), $(wildcard $(CROSSPLANE_CONTROLLERS_CHART_DIR)/templates/*.yaml))
 
 # This target copies manifests in crossplane-controllers and crossplane-types chart into crossplane chart.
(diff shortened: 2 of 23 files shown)


cluster/charts/kustomize_patches.yaml
```diff
@@ -0,0 +1,11 @@
+# We need InfrastructurePublications to be deleted first because if two resources
+# with pre-delete hooks have the same hook weight, they are deleted in alphabetic
+# order. So, Helm tries to delete InfrastructureDefinitions but they cannot be
+# deleted if they are published. Hence publications need to be deleted first.
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    "helm.sh/hook-weight": "-1"
```
muvaf, line 10: `InfrastructurePublication`s need to be deleted before the corresponding definitions because definitions are locked by a publication via finalizer. @negz we got some thoughts around this dependency like either making publication a field on definition or remove the dependency completely but they are not yet implemented.

muvaf:
I would like to list some of the challenges around deploying those core set to draw the full picture and decide whether we should do it in this release.
* `InfrastructureDefinition` and `InfrastructurePublication` CRDs should be registered before all of their instances. This is easy to achieve as we're already doing this for `PackageInstall` and `ClusterPackageInstall`. However, this pre-installation does not exist for the case where you want to deploy hosted mode with Helm2. Though it should be doable.
* `ClusterRole`s that are needed by each instances of those resources should be registered before or at the same time with the instances of `InfrastructureDefinition`s and `InfrastructurePublication`s. However, they can't be deleted before or at the same time because Crossplane needs to delete every instance of those new defined CRDs. The pre-install hack that we already have leaves whatever pre-installed in the cluster after uninstall ([Helm constraint](https://helm.sh/docs/topics/charts_hooks/#hook-resources-are-not-managed-with-corresponding-releases) and I don't think leaving aggregate `ClusterRole`s in the cluster is a good idea.
* Although conceptually it should be possible, with current implementation `InfrastructurePublication`s should be deleted at the same time or before their `InfrastructureDefinition`s because existence of a publication blocks the deletion of its correspondent definition. Helm does the deletion in alphabetic order and in a blocking manner, so, once it hits a `InfrastructureDefinition` to delete it stays there forever since Crossplane won't allow its deletion because it's published.
* I have tried using hooks and leaving everything in the cluster but there is some funny behavior of Helm2 where it deletes some of those resources in random manner and they keep get recreated. Besides, leaving them in cluster was already a bad idea due to `ClusterRole`s anyway.

In its essence, with our current packaging mechanism that has to satisfy both Helm2 and Helm3, I have not discovered a **safe** and reliable way to deploy those core sets. I can spend more time and somehow figure it out but this close to release, whatever I find working would be a risky solution. In short, I don't think we should do this for v0.11 in an automatic way.

My proposal is that in Composition documentation we can point to a folder `cluster/composition` that contain those `InfrastructureDefinition`s and `InfrastructurePublication`s together with their `ClusterRole` and recommend users to do `kubectl apply -f`. Due to mechanics of `ClusterRole`s, I would prefer having:
* `cluster/composition/clusterroles`
* `cluster/composition/definitions`

as separate folders so that uninstallation is smooth for the user. For example:
* Install with
  * `kubectl apply -f https://raw.githubusercontent.com/crossplane/crossplane/master/cluster/composition`
* Uninstall with
  * `kubectl delete -f https://raw.githubusercontent.com/crossplane/crossplane/master/cluster/composition/definitions` and then
  * `kubectl delete -f https://raw.githubusercontent.com/crossplane/crossplane/master/cluster/composition/clusterroles`

It's obviously not the best experience but I think it should be OK as an initial step. I really wish Helm had proper ways to handle ordered install/uninstall considering there exists hard dependencies such as RBAC and things like `CustomResourceDefinition`s.


After the release, I will come back to it.

negz:
@muvaf This SGTM. As I've mentioned before, I'm not super convinced that we really want to be installing a generic set of composite resources alongside Crossplane like we historically did with resource claims, so defining them but not installing them by default actually works quite well in my opinion.

muvaf:
Closing in favor of https://github.com/crossplane/crossplane/pull/1508

PR #1508, opened 2020-05-15
Add core InfrastructureDefinitions

muvaf:

This PR adds a set of `InfrastructureDefinition`s and `InfrastructurePublication`s that correspond to contemporary claims.

Manually via `kubectl apply -k`

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/cluster/composition/clusterroles/bucket.yaml
+++ b/cluster/composition/clusterroles/bucket.yaml
@@ -2,12 +2,12 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: buckets.definitions.crossplane.io
+  name: buckets.common.crossplane.io
   labels:
     rbac.crossplane.io/aggregate-to-crossplane: "true"
 rules:
   - apiGroups:
-      - definitions.crossplane.io
+      - common.crossplane.io
     resources:
       - buckets
       - buckets/status
--- a/cluster/composition/clusterroles/bucket.yaml
+++ b/cluster/composition/clusterroles/bucket.yaml
@@ -1,0 +1,17 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: buckets.definitions.crossplane.io
+  labels:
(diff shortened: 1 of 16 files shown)


cluster/composition/clusterroles/bucket.yaml
```diff
@@ -0,0 +1,17 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: buckets.definitions.crossplane.io
```
negz, line 5: I'm not sure `definitions` makes sense from the perspective of the folks who will be authoring these defined infrastructure resources. I'd prefer another API group. Some ideas:

* `buckets.infra.crossplane.io`
* `buckets.core.crossplane.io`
* `buckets.standard.crossplane.io`
* `buckets.common.crossplane.io`

Or some combination thereof? `buckets.common.infra.crossplane.io`?

hasheddan: I have been following a pattern of `buckets.storage.<org>`

muvaf: I wanted to stay away from grouping into categories like `storage` since it greatly varies among providers. `infra` as acronym is useful while speaking and in short chats but doesn't feel right in a domain although `infrastructure` is also too long. `standard` is kind of suggesting your own abstractions are exceptions and these are recommended, which is not the case.

I'm kind of leaning towards `buckets.common.crossplane.io`. `common` sounds more close to what they are compared to `core`.

kasey: +1 to `buckets.common.crossplane.io`. I am tempted to add something else between `common` and `crossplane` like `infra` but implying infra is in tune with our general focus on infra resources.

negz: Let's go with `buckets.common.crossplane.io` - @muvaf if you want to push an update I'm happy to merge.

muvaf:
@negz update all to be in `common.crossplane.io` instead of `definitions.crossplane.io`. It should be ready to merge.

PR #1514, opened 2020-05-15
Update Quick Start Guide

hasheddan:

WIP quick start for v0.11 docs

All examples need manual testing.

[skip ci]

### Checklist

I have:
  appropriate.

[definition of done].


docs/getting-started/provision-infrastructure.md
```diff
@@ -0,0 +1,212 @@
+---
+title: Provision Infrastructure
+toc: true
+weight: 3
+indent: true
+---
+
+# Provision Infrastructure
+
+Crossplane's most basic feature is provisioning infrastructure anywhere using
```
negz, line 10: Nit: Perhaps "core" or "foundation" rather than "most basic"?

negz:
@hasheddan Here's a Pod you can use to demonstrate that a PostgreSQL database is reachable:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: see-db
  namespace: default
spec:
  containers:
  - name: see-db
    image: postgres:9.6
    command: ['psql']
    args: ['-c', 'SELECT current_database();']
    env:
    - name: PGDATABASE
      value: postgres
    - name: PGHOST
      valueFrom:
        secretKeyRef:
          name: db-conn
          key: endpoint
    - name: PGUSER
      valueFrom:
        secretKeyRef:
          name: db-conn
          key: username
    - name: PGPASSWORD
      valueFrom:
        secretKeyRef:
          name: db-conn
          key: password
```

```console
$ kubectl apply -f pod.yaml 
pod/see-db created

$ kubectl logs see-db
 current_database 
------------------
 postgres
(1 row)
```

docs/getting-started/provision-infrastructure.md
```diff
@@ -0,0 +1,224 @@
...
+title: Provision Infrastructure
+toc: true
+weight: 3
+indent: true
+---
+
+# Provision Infrastructure
+
+Crossplane's core feature is provisioning infrastructure anywhere using the
+Kubernetes API. Once you have [installed a provider] and [configured your
```
ryanzhang-oss, line 11: I wonder how does this statement work with the adoption of an application model such as OAM in crossplane?

docs/getting-started/provision-infrastructure.md
```diff
@@ -0,0 +1,224 @@
...
+  name: rdsmysql
+spec:
+  forProvider:
+    engine: mysql
+    engineVersion: "10.0"
+    dbInstanceClass: rds.mysql.s1.small
+    dbInstanceStorageInGB: 20
+    securityIPList: "0.0.0.0/0"
+    masterUsername: "test123"
+  writeConnectionSecretsToRef:
```
ryanzhang-oss, line 181: where is this defined? I can't seem to see in the [Alibaba provider](https://github.com/crossplane/provider-alibaba/tree/master/apis) types.

hasheddan: @ryanzhang-oss this was adapted from https://github.com/crossplane/provider-alibaba/blob/master/cluster/examples/database/resource_class.yaml

docs/getting-started/publish-infrastructure.md
```diff
@@ -0,0 +1,560 @@
...
+
+<!-- Named Links -->
+
+[last section]: provision-infrastructure.yaml
+[composition]: ../composition.md
+[AWS]: #aws
+[GCP]: #gcp
+[Azure]: #azure
+[Alibaba]: #alibaba
+[next section]: consume-infrastructure.md
```
hasheddan, line 559: Note: dependent on @negz naming of this doc

docs/getting-started/provision-infrastructure.md
```diff
@@ -0,0 +1,224 @@
...
+
+Create the `RDSInstance`:
+```
+kubectl apply -f aws-rds.yaml
+```
+
+You can view the progress of provisioning the database with the following
+command:
+```
+kubectl get -f aws-rds.yaml
```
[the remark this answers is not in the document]
hasheddan: @muvaf `get` will only show the printer columns right?

docs/getting-started/publish-infrastructure.md
```diff
@@ -0,0 +1,560 @@
...
+    name: db-conn
+```
+
+Create the `MySQLRequirement`:
+```
+kubectl apply -f mysql-req.yaml
+```
+
+Once the database is provisioned on your provider of choice, you should see
+`STATUS: Bound` in the output when you run:
```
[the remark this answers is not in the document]
hasheddan: Oh dear 🤦‍♂️ haha

docs/getting-started/publish-infrastructure.md
```diff
@@ -0,0 +1,560 @@
...
+
+Because connection secrets are written as a Kubernetes `Secret` they can easily
+be consumed by Kubernetes primitives. The most basic building block in
+Kubernetes is the `Pod`. Let's define a `Pod` that will show that we are able to
+connect to our newly provisioned database.
+
+Create a file named `pod.yaml` with the following contents:
+```yaml
+apiVersion: v1
+kind: Pod
```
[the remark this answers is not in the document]
hasheddan: Hmm yeah I think that would make sense, except for I’m not sure if this `Pod` will get into a long `CrashLoopBackoff` waiting for the DB.

docs/getting-started/provision-infrastructure.md
```diff
@@ -0,0 +1,224 @@
...
+```
+
+When provisioning is complete, you should see `STATUS: Unbound` in the output.
```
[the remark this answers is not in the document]
hasheddan: Yeah I agree that would be more appropriate I don’t think it is shown in the printer columns for these resources :/ ... I could use `STATE` and just update it for each resource to whatever the cloud provider gives back

docs/getting-started/publish-infrastructure.md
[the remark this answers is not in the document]
hasheddan: Side note: we should remove binding status from printer columns and maybe from resources entirely in the future

hasheddan:
@muvaf that would be great! Thank you :)

docs/getting-started/publish-infrastructure.md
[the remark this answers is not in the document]
muvaf: If we create it after the DB is ready like composition example, why would it get into `CrashLoopBackoff`?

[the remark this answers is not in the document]
hasheddan: Aren’t you saying applying it all as one set of YAML?

docs/getting-started/provision-infrastructure.md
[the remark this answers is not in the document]
muvaf: I've started dropping PRs in all providers to include `READY` and `SYNCED` printer columns in all resources.

docs/getting-started/provision-infrastructure.md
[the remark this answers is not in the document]
muvaf: Oh sorry I thought it was YAML output, which is what I do usually.

docs/getting-started/publish-infrastructure.md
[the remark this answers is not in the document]
muvaf: Future is here waiting your review in respective repos.

docs/getting-started/publish-infrastructure.md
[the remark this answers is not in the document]
muvaf: I was saying that since Azure requires at least 3 resources for public access, we can deploy those in one big YAML. Others would stay same. After all is done, we can instruct user to create this `Pod` after they see all resources they deployed are ready.

docs/getting-started/provision-infrastructure.md
negz: @ryanzhang-oss It comes from the embedded `ResourceSpec`, which is common to all Crossplane managed resources.

* https://github.com/crossplane/provider-alibaba/blob/b4b04d/apis/database/v1alpha1/rds_types.go#L94
* https://github.com/crossplane/crossplane-runtime/blob/290de/apis/core/v1alpha1/resource.go#L175

docs/getting-started/provision-infrastructure.md
negz: I suspect this might be meaning core in the foundational/fundamental/lowest level sense. We'll update the wording to ensure equal footing is given to application models.

docs/getting-started/provision-infrastructure.md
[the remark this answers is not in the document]
negz: Is it worth including the `kubectl describe` output anywhere in this guide? I like the succinctness of `kubectl get`, especially when you're just waiting for a thing, but `kubectl describe` (with status, events, etc) communicates much more of a "hey this actually got provisioned" experience in my opinion.

That all said, I could be convinced that `kubectl describe` output would be too overwhelming for this guide.

docs/getting-started/publish-infrastructure.md
```diff
@@ -0,0 +1,560 @@
...
+  - mysqlinstancerequirements
+  - mysqlinstancerequirements/status
+  verbs:
+  - "*"
+```
+
+Create the `ClusterRole`:
+```
+kubectl apply -f mysql-cr.yaml
+```
```
negz, line 68: Could we store this `ClusterRole` as `quick-start-role.yaml` or something and `kubectl apply -f https://raw.url/quick-start-role.yaml`, without showing its contents here. Obviously this step needs to happen for things to work, but I don't think we add much value by demonstrating RBAC in motion.

docs/getting-started/publish-infrastructure.md
```diff
@@ -0,0 +1,560 @@
...
+typically provisioned at the *namespace scope* using Kubernetes primitives such
+as `Pod` or `Deployment`. To make infrastructure resources available to be
+provisioned at the namespace scope, they can be *published*. This consists of
+creating the following resources:
+
+- `InfrastructureDefinition`: defines a new kind of composite resource
+- `InfrastructurePublication`: makes an `InfrastructureDefinition` available at
+  the namespace scope
+- `ClusterRole`: allows Crossplane to manage this newly defined infrastructure
+  type
```
negz, line 26: I think we can gloss over (i.e. omit) this one here, since it's not a Crossplane concept.

docs/getting-started/publish-infrastructure.md
```diff
@@ -0,0 +1,560 @@
...
+```
+kubectl apply -f alibaba-mysqlinstance.yaml
+```
+
+## Create Requirement
+
+Now that we have defined our new type of infrastructure (`MySQLInstance`) and
+created at least one composition that can satisfy it, we can create a
+`MySQLInstanceRequirement` in the namespace of our choosing. In each
+`Composition` we defined we added a `provider: <name-of-provider` label. In the
```
negz, line 438:
```suggestion
`Composition` we defined we added a `provider: <name-of-provider>` label. In the
```

docs/getting-started/publish-infrastructure.md
hasheddan: I personally feel like it’s kind of weird to see everything then just `kubectl apply` this one thing. That being said, it is not a core concept and I get that the idea is simplicity here. It seems like we are already assuming folks are familiar with k8s concepts though so maybe it’s not adding much complexity?

docs/getting-started/provision-infrastructure.md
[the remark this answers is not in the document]
hasheddan: Would you want to `kubectl describe` and show the output here?

docs/getting-started/provision-infrastructure.md
```diff
@@ -0,0 +1,224 @@
...
+weight: 3
+indent: true
+---
+
+# Provision Infrastructure
+
+Crossplane's core feature is provisioning infrastructure anywhere using the
+Kubernetes API. Once you have [installed a provider] and [configured your
+credentials], you can create any infrastructure resource currently supported by
+the provider. Let's start by provisioning a database on your provider of choice.
```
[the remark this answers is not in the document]
hasheddan: Potentially yes.. I am not sure each of the providers can be boiled down to a few lines to configure though

docs/getting-started/provision-infrastructure.md
```diff
@@ -0,0 +1,224 @@
...
+    namespace: crossplane-system
+    name: aws-rdsmysql-conn
+  providerRef:
+    name: aws-provider
+  reclaimPolicy: Delete
+```
+
+Create the `RDSInstance`:
+```
+kubectl apply -f aws-rds.yaml
```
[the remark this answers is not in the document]
hasheddan: SGTM

docs/getting-started/provision-infrastructure.md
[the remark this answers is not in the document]
hasheddan: We could have some output that is collapsible so we could say something like "For more information run `kubectl describe` and you will see output like the following:" <collapsed-content>. That being said, I feel like we keep going back and forth between "all users will know how to use the Kubernetes API" and "some users may not know how to use the Kubernetes API". If we are saying the former then I think users will naturally `kubectl describe` if they want more information.

docs/getting-started/provision-infrastructure.md
hasheddan: Definitely, I updated it to say "Crossplane allows you to provision infrastructure anywhere using the Kubernetes API." since that is what this section is about :+1:

hasheddan:
@negz @muvaf I have addressed most of the comments here either with changes or with a response. Since it can be somewhat difficult to see what the tabs will look like, here are some screenshots:
tabs-1
tabs-2

These are dependent on https://github.com/crossplane/crossplane.github.io/pull/43

docs/getting-started/provision-infrastructure.md
```diff
@@ -0,0 +1,194 @@
...
+</div>
+<div class="tab-pane fade" id="gcp-tab-1" markdown="1">
+The GCP provider supports provisioning a [CloudSQL] instance with the
+`CloudSQLInstance` CRD it installs into your cluster.
+```yaml
+apiVersion: database.gcp.crossplane.io/v1beta1
+kind: CloudSQLInstance
+metadata:
+  name: cloudsqlmysql
+specTemplate:
```
prasek, line 72:
```suggestion
spec:
```

docs/getting-started/publish-infrastructure.md
```diff
@@ -0,0 +1,506 @@
...
+</div>
+
+## Create Requirement
+
+Now that we have defined our new type of infrastructure (`MySQLInstance`) and
+created at least one composition that can satisfy it, we can create a
+`MySQLInstanceRequirement` in the namespace of our choosing. In each
+`Composition` we defined we added a `provider: <name-of-provider>` label. In the
+`MySQLInstanceRequirement` below we use the `compositionSelector` with
+`provider: aws` to match our AWS `Composition`. This can be adjusted to select
```
prasek, line 400: Suggest tabs for each provider, since I just grabbed the content and ran it without reading the instructions, or maybe have the `matchLabels.provider`: REPLACE_ME in the example below.

docs/getting-started/publish-infrastructure.md
```diff
@@ -0,0 +1,506 @@
...
+  writeConnectionSecretsToNamespace: crossplane-system
+  reclaimPolicy: Delete
+  from:
+    apiVersion: database.example.org/v1alpha1
+    kind: MySQLInstance
+  to:
+    - base:
+        apiVersion: database.gcp.crossplane.io/v1beta1
+        kind: CloudSQLInstance
+        specTemplate:
```
prasek, line 212:
```suggestion
        spec:
```

docs/getting-started/publish-infrastructure.md
```diff
@@ -0,0 +1,506 @@
...
+            region: us-central1
+            ipConfiguration:
+              ipv4Enabled: true
+              authorizedNetworks:
+                - value: "0.0.0.0/0"
+            settings:
+              tier: db-n1-standard-1
+              dataDiskType: PD_SSD
+              dataDiskSizeGb: 10
+          writeConnectionSecretsToRef:
```
prasek, line 224:
```suggestion
          writeConnectionSecretToRef:
```

docs/getting-started/publish-infrastructure.md
```diff
@@ -0,0 +1,506 @@
...
+        kind: RDSInstance
+        spec:
+          forProvider:
+            dbInstanceClass: db.t2.small
+            masterUsername: masteruser
+            allocatedStorage: 20
+            engine: mysql
+            skipFinalSnapshotBeforeDeletion: true
+            publiclyAccessible: true
+          writeConnectionSecretsToRef:
```
prasek, line 174:
```suggestion
          writeConnectionSecretToRef:
```

haven't tested AWS yet, but docs say so this:
https://doc.crds.dev/github.com/crossplane/provider-aws/config/crd/database.aws.crossplane.io_rdsinstances.yaml#spec-writeConnectionSecretToRef

docs/getting-started/publish-infrastructure.md
```diff
@@ -0,0 +1,506 @@
...
+            location: West US 2
+            sslEnforcement: Disabled
+            version: "5.7"
+            sku:
+              tier: GeneralPurpose
+              capacity: 2
+              family: Gen5
+            storageProfile:
+              storageMB: 20480
+          writeConnectionSecretsToRef:
```
prasek, line 289:
```suggestion
          writeConnectionSecretToRef:
```

docs/getting-started/publish-infrastructure.md
```diff
@@ -0,0 +1,506 @@
...
+        kind: RDSInstance
+        spec:
+          forProvider:
+            engine: mysql
+            engineVersion: "10.0"
+            dbInstanceClass: rds.mysql.s1.small
+            dbInstanceStorageInGB: 20
+            securityIPList: "0.0.0.0/0"
+            masterUsername: "myuser"
+          writeConnectionSecretsToRef:
```
prasek, line 371:
```suggestion
          writeConnectionSecretToRef:
```

hasheddan:
@prasek Thanks for your review! Just pushed update to address each of your comments. Here is screenshot of tabs for `MySQLInstanceRequirement`:

tab-4

docs/getting-started/provision-infrastructure.md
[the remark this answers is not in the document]
hasheddan: @negz feel free to merge this at-will: https://github.com/crossplane/provider-alibaba/pull/11

docs/getting-started/publish-infrastructure.md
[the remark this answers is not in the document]
hasheddan: > I was saying that since Azure requires at least 3 resources for public access, we can deploy those in one big YAML. 

@muvaf I think I may not be following here. The Azure `Composition` has all 3 of these resources.

docs/getting-started/publish-infrastructure.md
hasheddan: Will address this depending on final resolution of https://github.com/crossplane/crossplane/pull/1514#discussion_r426226676

docs/getting-started/publish-infrastructure.md
```diff
@@ -0,0 +1,567 @@
...
+        secretKeyRef:
+          name: db-conn
+          key: password
+```
+
+This `Pod` simply connects to a PostgreSQL database and prints its name, so you
+should see the following output after creating it if you run `kubectl logs
+see-db`:
+```
+ current_database
```
muvaf, line 536: I don't see any output for a while when I follow AWS. Then the following after 2-3 min:
```
psql: could not connect to server: Connection refused
	Is the server running on host "default-my-db-zdcs5-g6bvd.csguarq5weag.us-east-1.rds.amazonaws.com" (54.160.126.136) and accepting
	TCP/IP connections on port 5432?
```
Could it be related to database being `MySQL` and client container image postgres? We don't mount a port and MySQL's default is 3306.

docs/getting-started/publish-infrastructure.md
```diff
@@ -0,0 +1,567 @@
...
+            sslEnforcement: Disabled
+            version: "5.7"
+            sku:
+              tier: GeneralPurpose
+              capacity: 2
+              family: Gen5
+            storageProfile:
+              storageMB: 20480
+          writeConnectionSecretToRef:
+            name: crossplane-system
```
muvaf, line 290:
```suggestion
            namespace: crossplane-system
```

docs/getting-started/publish-infrastructure.md
```diff
@@ -0,0 +1,567 @@
...
+            serverNameSelector:
+              matchControllerRef: true
+            resourceGroupNameSelector:
+              matchControllerRef: true
+            properties:
+              startIpAddress: 0.0.0.0
+              endIpAddress: 0.0.0.0
+          reclaimPolicy: Delete
+          providerRef:
+            name: example
```
muvaf, line 325:
```suggestion
            name: azure-provider
```

docs/getting-started/publish-infrastructure.md
```diff
@@ -0,0 +1,567 @@
...
+        - fromFieldPath: "spec.storageGB"
+          toFieldPath: "spec.forProvider.storageProfile.storageMB"
+          transforms:
+            - type: math
+              math:
+                multiply: 1024
+      connectionDetails:
+        - fromConnectionSecretKey: username
+        - fromConnectionSecretKey: password
+        - fromConnectionSecretKey: endpoint
```
muvaf, line 342: As far as I can see, you don't need `patches` or `connectionDetails`. In fact, it causes error to have `spec.writeConnectionSecretToRef.name` without namespace.

docs/getting-started/publish-infrastructure.md
```diff
@@ -0,0 +1,567 @@
...
+  writeConnectionSecretsToNamespace: crossplane-system
+  reclaimPolicy: Delete
+  from:
+    apiVersion: database.example.org/v1alpha1
+    kind: MySQLInstance
+  to:
+    - base:
+        apiVersion: azure.crossplane.io/v1alpha3
+        kind: ResourceGroup
+        spec:
```
muvaf, line 268: `spec.location` is a required field of `ResourceGroup`. We don't seem to have this validation in CRD.
```
'create failed: cannot create ResourceGroup: resources.GroupsClient#CreateOrUpdate:
      Invalid input: autorest/validation: validation failed: parameter=parameters.Location
      constraint=Null value=(*string)(nil) details: value can not be null; required
      parameter'
```

docs/getting-started/publish-infrastructure.md
muvaf: I see the same behavior in Azure, too.

docs/getting-started/publish-infrastructure.md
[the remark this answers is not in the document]
muvaf: I meant to add the `Pod` to the other document that is about provisioning infrastructure. The thing that would hold us back from that is Azure's extra resource requirement and we can create those together with MySQLServer, manually, as opposed to Composition.

docs/getting-started/provision-infrastructure.md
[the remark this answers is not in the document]
negz: Yeah, I agree it's safe to assume folks will know about `kubectl describe`. My suggestion is not so much because I'm worried the reader won't know they _can_ use `kubectl describe`, but because I think `kubectl describe` makes a stronger impression because (at least to me) there's more "wow factor" in seeing the spec and status of the resource updated to reflect the state in the cloud. That all said, it is quite a lot of output for a quickstart so I'm happy as long as it's demonstrated elsewhere in the docs.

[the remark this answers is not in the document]
negz: Let's stick with `kubectl get` for now - we can show `describe` in the more detailed guides.

docs/getting-started/publish-infrastructure.md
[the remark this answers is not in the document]
negz: > Should we create this Pod in infra provisioning step, too?

No, I don't think we should. Let's keep that section super simple; you apply YAML, you get infra in your cloud console. In reality I'm not sure we should be encouraging folks to statically provision something like a MySQLInstance and use it from a pod without ever creating a `Requirement`.

docs/getting-started/publish-infrastructure.md
negz: If I understand correctly the purpose of this guide is take folks on a tour of Crossplane concepts, getting them up and running as quickly as possible. This `ClusterRole` is needed for the Crossplane concepts we're introducing to work, but it's not a Crossplane concept, so to me it doesn't seem worth the extra lines of YAML and cognitive burden on the reader. We show the ClusterRole inline and discuss it in more detail in the longer composition guide.

Given this is a quick start I'd err toward including an obvious but succinct note like:

> Note that Crossplane must be granted an RBAC role allowing it to manage your newly defined and published infrastructure. Run `kubectl apply -f https://github.com/quick-start-role.yaml` to grant Crossplane access to the resources we'll cover in this guide.

docs/getting-started/publish-infrastructure.md
hasheddan: Ah yes at some point I switched from Postgres to MySQL, will update to use a MySQL client

docs/getting-started/publish-infrastructure.md
hasheddan: It was a typo to have `name` instead of `namespace` in the base, I am not sure why we would need no patches here?

docs/getting-started/publish-infrastructure.md
hasheddan: That sounds reasonable to me! Will update 👍

docs/getting-started/publish-infrastructure.md
```diff
@@ -0,0 +1,567 @@
...
+              matchControllerRef: true
+            location: West US 2
+            sslEnforcement: Disabled
+            version: "5.7"
+            sku:
+              tier: GeneralPurpose
+              capacity: 2
+              family: Gen5
+            storageProfile:
+              storageMB: 20480
```
negz, line 288: Typically when a field like `storageMB` is required by the definition I prefer to omit it from the base, since it will always be overwritten by the value of the composite/requirement.

docs/getting-started/publish-infrastructure.md
negz: I think there is a copy pasta bug here - all of the patches and connection details here apply to the MySQLServerFirewallRule, but it looks like they were intended to apply to the MySQLServer.

docs/getting-started/publish-infrastructure.md
```diff
@@ -0,0 +1,567 @@
...
+            name: azure-provider
+          reclaimPolicy: Delete
+      patches:
+        - fromFieldPath: "metadata.uid"
+          toFieldPath: "spec.writeConnectionSecretToRef.name"
+          transforms:
+            - type: string
+              string:
+                fmt: "%s-mysql"
+        - fromFieldPath: "spec.storageGB"
```
negz, line 301:
```suggestion
        - fromFieldPath: "spec.parameters.storageGB"
```

docs/getting-started/publish-infrastructure.md
```diff
@@ -0,0 +1,567 @@
...
+<br>
+<div class="tab-content">
+<div class="tab-pane fade in active" id="aws-tab-1" markdown="1">
+```yaml
+apiVersion: apiextensions.crossplane.io/v1alpha1
+kind: Composition
+metadata:
+  name: mysqlinstances.aws.database.example.org
+  labels:
+    provider: aws
```
negz, line 155: Could you label all of these Compositions with something consistent like `guide: quickstart`? That way the OAM section that follows this one will be able to select on that consistent label and get a valid composition for whatever cloud guide the reader followed.

docs/getting-started/publish-infrastructure.md
```diff
@@ -0,0 +1,567 @@
...
+        kind: MySQLServerFirewallRule
+        spec:
+          forProvider:
+            serverNameSelector:
+              matchControllerRef: true
+            resourceGroupNameSelector:
+              matchControllerRef: true
+            properties:
+              startIpAddress: 0.0.0.0
+              endIpAddress: 0.0.0.0
```
negz, line 322:
```suggestion
              endIpAddress: 255.255.255.254
```

docs/getting-started/publish-infrastructure.md
```diff
@@ -0,0 +1,567 @@
...
+    - name: PGUSER
+      valueFrom:
+        secretKeyRef:
+          name: db-conn
+          key: username
+    - name: PGPASSWORD
+      valueFrom:
+        secretKeyRef:
+          name: db-conn
+          key: password
```
negz, line 529:
```suggestion
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: see-db
  namespace: default
spec:
  containers:
  - name: see-db
    image: mysql:5.7
    command: ['mysql']
    args: ['-u', '$(MYSQL_USER)', '-e', 'SELECT CURRENT_USER();']
    env:
    - name: MYSQL_HOST
      valueFrom:
        secretKeyRef:
          name: db-conn
          key: endpoint
    - name: MYSQL_USER
      valueFrom:
        secretKeyRef:
          name: db-conn
          key: username
    - name: MYSQL_PWD
      valueFrom:
        secretKeyRef:
          name: db-conn
          key: password
```

This should do the trick:

```
$ kubectl logs -f see-db
CURRENT_USER()
myadmin@%
```

docs/getting-started/publish-infrastructure.md
negz: See https://github.com/crossplane/crossplane/pull/1514/files#r426487230 for a MySQL compatible example.

docs/getting-started/publish-infrastructure.md
hasheddan: Ah my apologies! The code snippet with this comment did not show whether it was with the `MySQLServerFirewallRule` or the `MySQLServer`, but I now see that the patch and connection details were duplicated :+1:

docs/getting-started/publish-infrastructure.md
hasheddan: sure thing!

docs/getting-started/publish-infrastructure.md
hasheddan: Thanks @negz. Apologies for getting this mixed up the first time, and thanks for providing another solution :)

docs/getting-started/provision-infrastructure.md
```diff
@@ -0,0 +1,194 @@
...
+
+When provisioning is complete, you should see `READY: True` in the output.
+You can then delete the `CloudSQLInstance`:
+```
+kubectl delete cloudsqlinstances.database.gcp.crossplane.io cloudsqlmysql
+```
+</div>
+<div class="tab-pane fade" id="azure-tab-1" markdown="1">
+The Azure provider supports provisioning an [Azure Database for MySQL] instance
+with the `MySQLServer` CRD it installs into your cluster.
```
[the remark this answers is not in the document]
hasheddan: Good idea :)

docs/getting-started/publish-infrastructure.md
```diff
@@ -0,0 +1,567 @@
...
+      provider: alibaba
+  writeConnectionSecretToRef:
+    name: db-conn
+```
+</div>
+</div>
+
+After creating the `MySQLInstanceRequirement` Crossplane will provision a
+database instance on your provider of choice. Once provisioning is complete, you
+should see `READY: True` in the output when you run:
```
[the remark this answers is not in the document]
hasheddan: Added a note with some helpful commands to view various groups of resources :+1:

prasek:
@hasheddan this looks awesome in the rendered docs, but is there a way for this to render more cleanly on GitHub when viewing the docs pages there?

image

hasheddan:
@prasek I think this was solved in most recent update when I added additional linebreaks. This is how it appears for me: 
tabs-nice

Could you try refreshing and let me know if it looks better for you?

prasek:
@hasheddan looks great, thanks!

hasheddan:
Just ran through `provider-alibaba` (with this fix: https://github.com/crossplane/provider-alibaba/pull/13) and made some updates. Everything working as expected for alibaba now :+1:

docs/getting-started/provision-infrastructure.md
```diff
@@ -0,0 +1,218 @@
...
+
+## Next Steps
+
+Now that you have seen how to provision individual infrastructure resources,
+let's take a look at how we can compose infrastructure resources together and
+publish them as a single unit to be consumed in the [next section].
+
+<!-- Named Links -->
+
+[installed a provider]: install.md
```
kasey, line 211: Should this link to the quickstart version instead? ( docs/getting-started/install-configure.md )

PR #1515, opened 2020-05-15
Documentation: Providers section

muvaf:
# Description of your changes

Adds `Providers` section to new documentation.

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/docs/providers.md
+++ b/docs/providers.md
@@ -7,14 +7,11 @@ indent: true
 
 # Providers
 
-Providers are external infrastructure providers such as Amazon Web Services,
-Microsoft Azure, Google Cloud and Alibaba Cloud.
-
-In order to provision a resource in the provider, a Custom Resource Definition(CRD)
+Providers extend Crossplane to enable infrastructure resource provisioning.
+In order to provision a resource, a Custom Resource Definition(CRD)
 needs to be registered in your Kubernetes cluster and its controller should
-be watching instances of that Custom Resource Definition. Provider packages
-contain many Custom Resource Definitions and their controllers to extend
-functionality of Crossplane.
+be watching the Custom Resources those CRDs define. Provider packages
+contain many Custom Resource Definitions and their controllers.
 
 Here is the list of current providers:
 
@@ -31,7 +28,7 @@ Here is the list of current providers:
 
 ## Installation
 
-The core Crossplane controller can install Provider controllers and CRDs for you
+The core Crossplane controller can install provider controllers and CRDs for you
 through its own provider packaging mechanism, which is triggered by the application
 of a `ClusterPackageInstall` resource. For example, in order to request
 installation of the `provider-gcp` package, apply the following resource to the
@@ -51,7 +48,7 @@ The field `spec.package` is where you refer to the container image of the
 provider. Crossplane Package Manager will unpack that container, register
 CRDs and set up necessary RBAC rules and then start the controllers.
 
-There are a few other ways to install to trigger the installation of provider
+There are a few other ways to to trigger the installation of provider
 packages:
 
 * As part of Crossplane Helm chart by adding the following
@@ -59,18 +56,19 @@ packages:
   It will install the default version hard-coded in that release of Crossplane
   Helm chart but if you'd like to specif an exact version, you can add:
   `--set clusterPackages.gcp.version=master`.
-* Using Crossplane kubectl plugin:
+* Using [Crossplane kubectl plugin][crossplane-cli]:
   `kubectl crossplane package install --cluster -n crossplane-system 'crossplane/provider-gcp:master' provider-gcp`
 
-You can delete them by deleting the `ClusterPackageInstall` resource you've created.
+You can uninstall a provider by deleting the `ClusterPackageInstall` resource
+you've created.
 
 ## Configuration
 
-In order to authenticate with the provider API, the provider controllers
-need to have access to credentials. It could be an AWS IAM User, GCP Service
-Account or Azure Service Principal. Every provider has a type called `Provider`
-that has information about how to authenticate to the provider API. A `Provider`
-resource for Azure looks like the following:
+In order to authenticate with the external provider API, the provider controllers
+need to have access to credentials. It could be an IAM User for AWS, a Service
+Account for GCP or a Service Principal for Azure. Every provider has a type called
+`Provider` that has information about how to authenticate to the provider API.
+An example `Provider` resource for Azure looks like the following:
 
 ```yaml
 apiVersion: azure.crossplane.io/v1alpha3
@@ -89,7 +87,7 @@ of that key should contain the credentials that the controller will use. The
 documentation of each provider should give you an idea of how that credentials
 blob should look like. See [Getting Started][getting-started] guide for more details.
 
-The following is an example usage of Azure `Provider`, referred by a `MySQLServer`:
+The following is an example usage of Azure `Provider`, referenced by a `MySQLServer`:
 
 ```yaml
 apiVersion: database.azure.crossplane.io/v1beta1
@@ -103,7 +101,7 @@ spec:
 
 The Azure provider controller will use that provider for this instance of `MySQLServer`.
 Since every resource has its own reference to a `Provider`, you can have multiple
-`Provider` resources in your cluster referred by different resources.
+`Provider` resources in your cluster referenced by different resources.
 
 
 <!-- Named Links -->
@@ -119,3 +117,4 @@ Since every resource has its own reference to a `Provider`, you can have multipl
 [provider-alibaba]: https://github.com/crossplane/provider-alibaba
 [alibaba-reference]: https://doc.crds.dev/github.com/crossplane/provider-alibaba
 [getting-started]: getting-started/configure.md
+[crossplane-cli]: https://github.com/crossplane/crossplane-cli
--- a/docs/providers.md
+++ b/docs/providers.md
@@ -31,9 +31,11 @@ Here is the list of current providers:
 
 ## Installation
 
-Since all managed resources are all cluster-scoped, provider packages are
-bundled as `ClusterPackage`s. In order to request installation of a provider,
-the following `ClusterPackageInstall` needs to be created:
+The core Crossplane controller can install Provider controllers and CRDs for you
+through its own provider packaging mechanism, which is triggered by the application
+of a `ClusterPackageInstall` resource. For example, in order to request
+installation of the `provider-gcp` package, apply the following resource to the
+cluster where Crossplane is running:
 
 ```yaml
 apiVersion: packages.crossplane.io/v1alpha1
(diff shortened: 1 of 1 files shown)


docs/providers.md
```diff
@@ -0,0 +1,114 @@
...
+Keep in mind that it's not a requirement that the authentication should be done
+through a credentials blob in a `Secret`. For example, in AWS, `Provider` has
+a field called `useServiceAccount` to imply that the provider controller
+should use workload identity feature of AWS instead of a `Secret` that contains
+IAM credentials.
+
+Since each managed resource refers to a `Provider` resource, you can have multiple
+`Provider` resources in your cluster and different managed resources using them.
+
+
```
muvaf, line 102: @prasek There are following sections listed in the doc but I couldn't find much that is not addressed above:
```
- Versions
- Multiple Provider Instances
- Referring to a Specific Provider
```

docs/providers.md
```diff
@@ -0,0 +1,114 @@
...
+* [Microsoft Azure][provider-azure]
+  * [API Reference][azure-reference]
+* [Rook][provider-rook]
+  * [API Reference][rook-reference]
+* [Alibaba Cloud][provider-alibaba]
+  * [API Reference][alibaba-reference]
+
+## Installation
+
+Since all managed resources are all cluster-scoped, provider packages are
```
kasey, line 34: Since all managed resources are ~all~ cluster-scoped

docs/providers.md
```diff
@@ -0,0 +1,114 @@
...
+bundled as `ClusterPackage`s. In order to request installation of a provider,
```
kasey, line 35: This might be confusing if you don't know what a ClusterPackage is. Consider just introducing the `ClusterPackageInstall` type like:
The core Crossplane controller can install Provider controllers and CRDs for you through its own provider packaging mechanism, which is triggered by the application of a `ClusterPackageInstall` resource. For example, in order to request installation of the `provider-gcp` package, apply the following resource to the cluster where Crossplane is running:

docs/providers.md
```diff
@@ -0,0 +1,114 @@
...
+spec:
+  package: "crossplane/provider-gcp:master"
+```
+
+The field `spec.package` is where you refer to the container image of the
+provider. Crossplane Package Manager will unpack that container, register
+CRDs and set up necessary RBAC rules and then start the controllers.
+
+There are a few ways that are more convenient than creating a YAML to install
+the providers.
```
kasey, line 53: You imply that the `ClusterPackageInstall` is less convenient but it's not clear why. I'd consider structuring this whole section like "You can install a Crossplane infrastructure provider via helm, the crossplane kubectl plugin, or by apply a Crossplane `ClusterPackageInstall` resource." Then you can just bullet point these 3 options out and people can check out the option that sounds like it fits their style. Do we have a preferred means of installing?

docs/providers.md
```diff
@@ -0,0 +1,114 @@
...
+  credentialsSecretRef:
+    namespace: crossplane-system
+    name: azure-prod-creds
+    key: credentials
+```
+
+You can see that there is a reference to a key in a specific `Secret`. The value
+of that key should contain the credentials that the controller will use. The
+documentation of each provider should give you an idea of how that credentials
+blob should look like.
```
kasey, line 91: I think you meant to say "what that credentials blob should look like" or "how that credentials blob should look".

docs/providers.md
```diff
@@ -0,0 +1,114 @@
...
+
+Keep in mind that it's not a requirement that the authentication should be done
+through a credentials blob in a `Secret`. For example, in AWS, `Provider` has
+a field called `useServiceAccount` to imply that the provider controller
+should use workload identity feature of AWS instead of a `Secret` that contains
+IAM credentials.
```
kasey, line 97: Link? https://github.com/crossplane/provider-aws/blob/b3608d666b95b65387dc3802117d01a15959e61e/AUTHENTICATION.md

Do we have equivalent workload identity options for other providers?

docs/providers.md
```diff
@@ -0,0 +1,114 @@
...
+
+Since each managed resource refers to a `Provider` resource, you can have multiple
+`Provider` resources in your cluster and different managed resources using them.
```
kasey, line 100: Is it worth reiterating that this is done via `spec.providerRef`?

docs/providers.md
kasey: I'm not seeing anything in `provider-gcp` or `provider-azure` that explains how to set up credentials. `provider-aws` has a detailed authentication doc, maybe we should make an `AUTHENTICATION.md` doc a requirement for providers.

docs/providers.md
negz: I suggest we pick our preferred method here and not worry too much about documenting the alternatives. To my knowledge submitting a `ClusterPackageInstall` is our preferred method. I'd slightly prefer recommending the `crossplane package install` command, but that would require the user to have `crossplane-cli` installed.

docs/providers.md
negz: https://crossplane.io/docs/v0.10/getting-started/configure.html

The current docs include this, but I do like the idea of moving that into an `AUTHENTICATION.md` file in each provider in a future release.

docs/providers.md
negz: I believe we only support this for AWS today. It's a somewhat underdeveloped experience that we put into place quickly because our several of users were asking for it. I'd suggest including it as a footnote - i.e. one sentence with a link. Something like:

> The AWS provider includes experimental support for authenticating using a Kubernete service account. See [the documentation for details](https://github.com/crossplane/provider-aws/blob/b3608d666b95b65387dc3802117d01a15959e61e/AUTHENTICATION.md).

docs/providers.md
prasek: If we can do tabs or similar would be nice to document the 3 install options, so new users are exposed to the crossplane-cli while also understanding that everything can be done with CRs against the Kubernetes API, and that we also offer the convenience of installing optional cluster-scoped packages when installing the Crossplane helm chart.

hasheddan: Supporting tabs in the markdown appears to be a little more difficult given our jekyll setup than I originally expected. However, if we get the content to a good place then I could circle back around on Monday and try and move some of the stuff to tabs. For the time-being I have just been doing separate sections but with links to jump to each. I also want to check how jekyll handles `<summary>` in markdown :+1:

docs/providers.md
hasheddan: +1 to moving auth to each provider. Also think it would be nice to potentially have some of this workflow in the `crossplane-cli`

docs/providers.md
hasheddan: This also currently requires some additional work due to not being able to run containers as `root`, which is required to access the token that AWS EKS webhook injects. It can still be done but requires doing your own build of `provider-aws`. The current thinking is to wait until this is fixed with projected service account volume in upstream k8s. More info here: https://github.com/crossplane/provider-aws/issues/211

muvaf: If it's not working as is, I'd prefer removing it. What do you think?

docs/providers.md
muvaf: I was actually thinking to even include an example small `MySQLServer` managed resource that refers to this provider.

muvaf:
So there seems to be 2 main comments:
- Whether more detail should be included per provider since secret data format changes.
- Whether all three installation instructions should be mentioned.

My main approach for this guide was more about talking about how the stuff works rather than a walk-through. In that sense, I think it makes sense to leave the actual credential configuration to getting-started/<provider>-config.md and link from this page instead of going into details for each. AFAIK, the getting started will include a simple installation instruction for each provider, too, so it looks more complete if we talk about all other possible ways here.

However, if the goal is more like a walk-through, getting user to a working state fast, then I can revise those sections to be more practical.

What do you guys think?

muvaf:
> Main feedback is I wonder if we should have guides focused on install+auth per-provider which we could link from here, but that's probably too much on our current timeline, let's think about it for next iteration.

I think that's a good idea. Provider README.md files can have an install+configure instructions assuming Crossplane is already installed. There, they can have more freedom to talk about ins and bits of that specific provider.

docs/providers.md
hasheddan: @muvaf yeah I think that sounds good for now :+1:

docs/providers.md
```diff
@@ -0,0 +1,121 @@
...
+title: Providers
+toc: true
+weight: 400
+indent: true
+---
+
+# Providers
+
+Providers are external infrastructure providers such as Amazon Web Services,
+Microsoft Azure, Google Cloud and Alibaba Cloud.
```
negz, line 11: Should we mention that providers extend Crossplane? I suspect people who are not familiar with Crossplane could interpret "provider" in this opening sentence to just mean "cloud provider", without realising that we're talking about providers as a Crossplane concept.

docs/providers.md
```diff
@@ -0,0 +1,121 @@
...
+
+In order to provision a resource in the provider, a Custom Resource Definition(CRD)
+needs to be registered in your Kubernetes cluster and its controller should
+be watching instances of that Custom Resource Definition. Provider packages
```
negz, line 15: Ranty nitpick: I'd really love to move away from the "instance of a CRD" terminology. To me an "instance of a CRD" sounds like we're discussing a resource of `kind: CustomResourceDefinition`, not an instance of the custom resource defined by the CRD.

docs/providers.md
```diff
@@ -0,0 +1,121 @@
...
+There are a few other ways to install to trigger the installation of provider
+packages:
+
+* As part of Crossplane Helm chart by adding the following
+  statement to your `helm install` command: `--set clusterPackages.gcp.deploy=true`
+  It will install the default version hard-coded in that release of Crossplane
+  Helm chart but if you'd like to specif an exact version, you can add:
+  `--set clusterPackages.gcp.version=master`.
+* Using Crossplane kubectl plugin:
+  `kubectl crossplane package install --cluster -n crossplane-system 'crossplane/provider-gcp:master' provider-gcp`
```
negz, line 63: If we're going to keep this detail we should probably include a link to the crossplane-cli install docs.

docs/providers.md
```diff
@@ -0,0 +1,121 @@
...
+  namespace: crossplane-system
+spec:
+  package: "crossplane/provider-gcp:master"
+```
+
+The field `spec.package` is where you refer to the container image of the
+provider. Crossplane Package Manager will unpack that container, register
+CRDs and set up necessary RBAC rules and then start the controllers.
+
+There are a few other ways to install to trigger the installation of provider
```
negz, line 54:
```suggestion
There are a few other ways to trigger the installation of provider
```

docs/providers.md
```diff
@@ -0,0 +1,121 @@
...
+
+* As part of Crossplane Helm chart by adding the following
+  statement to your `helm install` command: `--set clusterPackages.gcp.deploy=true`
+  It will install the default version hard-coded in that release of Crossplane
+  Helm chart but if you'd like to specif an exact version, you can add:
+  `--set clusterPackages.gcp.version=master`.
+* Using Crossplane kubectl plugin:
+  `kubectl crossplane package install --cluster -n crossplane-system 'crossplane/provider-gcp:master' provider-gcp`
+
+You can delete them by deleting the `ClusterPackageInstall` resource you've created.
```
negz, line 65:
```suggestion
You can uninstall a provider by deleting the `ClusterPackageInstall` resource you've created.
```

docs/providers.md
```diff
@@ -0,0 +1,121 @@
...
+    name: azure-prod-creds
+    key: credentials
+```
+
+You can see that there is a reference to a key in a specific `Secret`. The value
+of that key should contain the credentials that the controller will use. The
+documentation of each provider should give you an idea of how that credentials
+blob should look like. See [Getting Started][getting-started] guide for more details.
+
+The following is an example usage of Azure `Provider`, referred by a `MySQLServer`:
```
negz, line 92:
```suggestion
The following is an example usage of Azure `Provider`, referenced by a `MySQLServer`:
```

docs/providers.md
muvaf: > we should probably include a link to the crossplane-cli install docs.

Updated to link CLI repo.

> If we're going to keep this detail

See my comment above. I'm open to go either way depending on what we want this page to be about.

PR #1523, opened 2020-05-18
Install/configure quickstart with tabs

prasek:
Install and configure provider quickstart with tabs.

### Checklist

I have:
  appropriate.

[skip ci]

### DIFF
--- a/docs/getting-started/install-configure.md
+++ b/docs/getting-started/install-configure.md
@@ -1,0 +1,384 @@
+---
+title: Install & Configure
+toc: true
+weight: 1
+indent: true
+---
+
+# Install & Configure Quickstart
+
+Crossplane can be easily installed into any existing Kubernetes cluster using
+the regularly published Helm chart. The Helm chart contains all the custom
+resources and controllers needed to deploy and configure Crossplane.
+
+See [Install] and [Configure] docs for installing alternate versions and more
+detailed instructions.
+
+## Get a Kubernetes Cluster
+
+<ul class="nav nav-tabs">
+<li class="active"><a href="#setup-mac-brew" data-toggle="tab">macOS via Homebrew</a></li>
+<li><a href="#setup-mac-linux" data-toggle="tab">macOS / Linux</a></li>
+<li><a href="#setup-windows" data-toggle="tab">Windows</a></li>
+</ul>
+<br>
+<div class="tab-content">
+<div class="tab-pane fade in active" id="setup-mac-brew" markdown="1">
+For macOS via Homebrew use the following:
+
+```
+brew upgrade
+brew install kind
+brew install kubectl
+brew install helm
+
+kind create cluster --image kindest/node:v1.17.5 --wait 5m
+```
+
+</div>
+<div class="tab-pane fade" id="setup-mac-linux" markdown="1">
+For macOS / Linux use the following:
+
+* [Kubernetes cluster]
+  * [Kind]
+  * [Minikube], minimum version `v0.28+`
+  * etc.
+
+* [Helm], minimum version `v2.12.0+`.
+  * For Helm 2, make sure Tiller is initialized with sufficient permissions to
+    work on `crossplane-system` namespace.
+
+</div>
+<div class="tab-pane fade" id="setup-windows" markdown="1">
+For Windows use the following:
+
+* [Kubernetes cluster]
+  * [Kind]
+  * [Minikube], minimum version `v0.28+`
+  * etc.
+
+* [Helm], minimum version `v2.12.0+`.
+  * For Helm 2, make sure Tiller is initialized with sufficient permissions to
+    work on `crossplane-system` namespace.
+
+</div>
+</div>
+
+## Install Crossplane
+<ul class="nav nav-tabs">
+<li class="active"><a href="#install-tab-helm3" data-toggle="tab">Helm 3 (alpha)</a></li>
+<li><a href="#install-tab-helm2" data-toggle="tab">Helm 2 (alpha)</a></li>
+<li><a href="#install-tab-helm3-master" data-toggle="tab">Helm 3 (master)</a></li>
+<li><a href="#install-tab-helm2-master" data-toggle="tab">Helm 2 (master)</a></li>
+</ul>
+<br>
+<div class="tab-content">
+<div class="tab-pane fade in active" id="install-tab-helm3" markdown="1">
+Use Helm 3 to install the latest official `alpha` release of Crossplane, suitable for community use and testing:
+
+```
+kubectl create namespace crossplane-system
+
+helm repo add crossplane-alpha https://charts.crossplane.io/alpha
+helm install crossplane --namespace crossplane-system crossplane-alpha/crossplane
+```
+
+</div>
+<div class="tab-pane fade" id="install-tab-helm2" markdown="1">
+Use Helm 2 to install the latest official `alpha` release of Crossplane, suitable for community use and testing:
+
+```
+kubectl create namespace crossplane-system
+
+helm repo add crossplane-alpha https://charts.crossplane.io/alpha
+helm install --name crossplane --namespace crossplane-system crossplane-alpha/crossplane
+```
+
+</div>
+<div class="tab-pane fade" id="install-tab-helm3-master" markdown="1">
+Use Helm 3 to install the latest `master` pre-release version of Crossplane:
+
+```
+kubectl create namespace crossplane-system
+
+helm repo add crossplane-master https://charts.crossplane.io/master/
+helm search repo crossplane-master --devel
+
+# Kubernetes 1.15 and newer versions
+helm install crossplane --namespace crossplane-system crossplane-master/crossplane --version <version> --devel
+```
+
+For example:
+```
+helm install crossplane --namespace crossplane-system crossplane-master/crossplane --version 0.11.0-rc.100.gbc5d311 --devel
+```
+
+</div>
+<div class="tab-pane fade" id="install-tab-helm2-master" markdown="1">
+Use Helm 2 to install the latest `master` pre-release version of Crossplane, which is suitable for testing pre-release versions:
+
+```
+kubectl create namespace crossplane-system
+
+helm repo add crossplane-master https://charts.crossplane.io/master/
+helm search crossplane-master --devel
+
+helm install --name crossplane --namespace crossplane-system crossplane-master/crossplane --version <version>
+```
+
+For example:
+
+```
+helm install --name crossplane --namespace crossplane-system crossplane-master/crossplane --version 0.11.0-rc.100.gbc5d311
+```
+
+</div>
+</div>
+
+## Install Crossplane CLI
+The [Crossplane CLI] adds a set of `kubectl crossplane` commands to simplify common tasks:
+```
+curl -sL https://raw.githubusercontent.com/crossplane/crossplane-cli/master/bootstrap.sh | bash
+```
+
+## Select Provider
+Install and configure a provider for Crossplane to use for infrastructure provisioning:
+<ul class="nav nav-tabs">
+<li class="active"><a href="#provider-tab-aws" data-toggle="tab">AWS</a></li>
+<li><a href="#provider-tab-gcp" data-toggle="tab">GCP</a></li>
+<li><a href="#provider-tab-azure" data-toggle="tab">Azure</a></li>
+<li><a href="#provider-tab-alibaba" data-toggle="tab">Alibaba</a></li>
+</ul>
+<br>
+<div class="tab-content">
+<div class="tab-pane fade in active" id="provider-tab-aws" markdown="1">
+
+### Install AWS Provider
+
+```
+PACKAGE=crossplane/provider-aws:master
+NAME=provider-aws
+
+kubectl crossplane package install --cluster --namespace crossplane-system ${PACKAGE} ${NAME}
+```
+
+### Get AWS Account Keyfile
+
+Using an AWS account with permissions to manage RDS databases:
+```
+AWS_PROFILE=default && echo -e "[default]\naws_access_key_id = $(aws configure get aws_access_key_id --profile $AWS_PROFILE)\naws_secret_access_key = $(aws configure get aws_secret_access_key --profile $AWS_PROFILE)" > creds.conf
+```
+
+### Create a Provider Secret
+
+```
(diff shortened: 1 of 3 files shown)


docs/getting-started/install-configure.md
```diff
@@ -0,0 +1,183 @@
...
+GCP_PROJECT_ID=TODO_REPLACE
+curl -s https://raw.githubusercontent.com/crossplane/stack-gcp-sample/master/hack/gcp-credentials.sh | PROJECT_ID=${GCP_PROJECT_ID} bash -
+```
+
+Save the resulting secret JSON to `gcp-account-creds.json` and be sure to delete it later!
+
+
+### Create a Provider Secret
+```
+PROVIDER_SECRET_NAMESPACE=crossplane-system
```
negz, line 121: Can we just use crossplane-system inline here and in the example below? I'm sure folks with Kubernetes experience will be able to spot that they could change it.

docs/getting-started/install-configure.md
```diff
@@ -0,0 +1,183 @@
...
+<br>
+<div class="tab-content">
+<div class="tab-pane fade in active" id="setup-mac-brew" markdown="1">
+```
+brew upgrade
+brew install kind
+brew install kubectl
+brew install helm
+
+kind create cluster --image kindest/node:v1.15.11 --wait 5m
```
negz, line 33: We may want to recommend at least 1.17 here; the OAM local addon docs state they require it (and I'm inclined to believe them).

docs/getting-started/install-configure.md
```diff
@@ -0,0 +1,183 @@
...
+
+### Create gcp-account-creds.json
+The following `curl | bash` script that creates a service account and grants
+access to infrastructure services. Alternatively you use the [Configure] docs
+to manually generate account credentials.
+```
+# replace this with your own gcp project id
+GCP_PROJECT_ID=TODO_REPLACE
+curl -s https://raw.githubusercontent.com/crossplane/stack-gcp-sample/master/hack/gcp-credentials.sh | PROJECT_ID=${GCP_PROJECT_ID} bash -
+```
```
negz, line 114: I would prefer not to try to script this for folks. I think at best it's a scary proposition to run someone's script with your cloud provider credentials, and at worst it will lead to a poor UX when we inevitably forget to update this script when new resources that need new IAM roles are added.

I'm wondering whether we really have to automate and/or teach the process of creating a new project and granting IAM roles to a service account. I can't find any examples of similar infra-automation projects including those in their documentation. If we left this out the total process would be something like two commands: create the service account, and save it as a secret.

docs/getting-started/install-configure.md
```diff
@@ -0,0 +1,183 @@
...
+  name: gcp-provider
+spec:
+  # replace this with your own gcp project id
+  projectID: ${GCP_PROJECT_ID}
+  credentialsSecretRef:
+    namespace: ${PROVIDER_SECRET_NAMESPACE}
+    name: gcp-account-creds
+    key: credentialsKeyBody
+EOF
+```
```
negz, line 141: This format is pasteable, but we've found it to be imposing and/or less readable in the past compared to something like:

```yaml
apiVersion: gcp.crossplane.io/v1alpha3
kind: Provider
metadata:
  name: gcp-provider
spec:
  # Replace this with your own gcp project id
  projectID: my-project
  credentialsSecretRef:
    namespace: crossplane-system
    name: gcp-account-creds
    key: credentialsKeyBody
```

docs/getting-started/install-configure.md
```diff
@@ -0,0 +1,183 @@
...
+
+See [Install] and [Configure] docs for installing alternate versions and more
+detailed instructions.
+
+## Get a Kubernetes Cluster
+
+<ul class="nav nav-tabs">
+<li class="active"><a href="#setup-mac-brew" data-toggle="tab">macOS via Homebrew</a></li>
+<li><a href="#setup-mac-linux" data-toggle="tab">macOS / Linux</a></li>
+<li><a href="#setup-mac-linux" data-toggle="tab">Windows</a></li>
```
hasheddan, line 22:
```suggestion
<li><a href="#setup-windows" data-toggle="tab">Windows</a></li>
```

docs/getting-started/install-configure.md
```diff
@@ -0,0 +1,183 @@
...
+<li><a href="#provider-tab-aws" data-toggle="tab">AWS</a></li>
+<li><a href="#provider-tab-azure" data-toggle="tab">Azure</a></li>
+<li><a href="#provider-tab-alibaba" data-toggle="tab">Alibaba</a></li>
+</ul>
+<br>
+<div class="tab-content">
+<div class="tab-pane fade in active" id="provider-tab-gcp" markdown="1">
+### Install GCP Provider
+```
+PACKAGE=crossplane/provider-gcp:v0.9.0
```
hasheddan, line 101:
```suggestion
PACKAGE=crossplane/provider-gcp:master
```

Guessing we will probably update the version tag for each provider on the release branch 👍

docs/getting-started/install-configure.md
```diff
@@ -0,0 +1,183 @@
...
+helm repo add crossplane-alpha https://charts.crossplane.io/alpha
+helm install --name crossplane --namespace crossplane-system crossplane-alpha/crossplane
+```
+</div>
+</div>
+
+## Install Crossplane CLI
+The Crossplane CLI adds a set of `kubectl crossplane` commands to simplify common tasks:
+```
+curl -sL https://raw.githubusercontent.com/crossplane/crossplane-cli/master/bootstrap.sh | bash
```
hasheddan, line 84: @prasek do you think we should do another release of `crossplane-cli` prior to these docs going out? Or maybe just using `master` is fine?

prasek: @hasheddan IMO yes we should cut another crossplane-cli if that is how stable versions are released.

hasheddan: SGTM :+1:

docs/getting-started/install-configure.md
prasek: fixed

docs/getting-started/install-configure.md
prasek: fixed

docs/getting-started/install-configure.md
prasek: fixed

docs/getting-started/install-configure.md
prasek: fixed

docs/getting-started/install-configure.md
prasek: done

docs/getting-started/install-configure.md
prasek: done

kasey:
This doc overlaps quite a lot with https://github.com/crossplane/crossplane/blob/master/docs/getting-started/install.md

I wonder if it would simplify things to consolidate the two into install.md, that way all the links to install instructions in existing documentation will take people to the most up to date instructions?

docs/getting-started/install-configure.md
```diff
@@ -0,0 +1,299 @@
...
+
+```
+helm install --name crossplane --namespace crossplane-system crossplane-master/crossplane --version 0.11.0-rc.100.gbc5d311
+```
+
+</div>
+</div>
+
+## Install Crossplane CLI
+The [Crossplane CLI] adds a set of `kubectl crossplane` commands to simplify common tasks:
```
kasey, line 139: Should this be part of the quickstart? I think a ton of kubectl users don't install plugins, feels like something that could create friction if I think I have to install the plugin before I can try crossplane.

docs/getting-started/install-configure.md
```diff
@@ -0,0 +1,299 @@
...
+<div class="tab-content">
+<div class="tab-pane fade in active" id="provider-tab-aws" markdown="1">
+
+### Install AWS Provider
+
+```
+PACKAGE=crossplane/provider-aws:master
+NAME=provider-aws
+
+kubectl crossplane package install --cluster --namespace crossplane-system ${PACKAGE} ${NAME}
```
kasey, line 162:
```
$ kubectl crossplane package install --cluster --namespace crossplane-system ${PACKAGE} ${NAME}
error: unable to recognize "STDIN": no matches for kind "ClusterStackInstall" in version "stacks.crossplane.io/v1alpha1"
```
Presumably due to stacks->packages renaming?

negz:
> I wonder if it would simplify things to consolidate the two into install.md, that way all the links to install instructions in existing documentation will take people to the most up to date instructions?

I think that might have been @prasek's original approach, however last night on Slack I said:

> I lean toward including the MVP “install and configure” steps in the actual quick start guide and then leaving the install and configure guides as is.

In retrospect I don't feel too strongly either way. My concern is that the current install and configure guides could overwhelm folks. They go into quite a lot of detail about all the various ways you could install Crossplane, and in some cases walk folks through the process of creating a new cloud project (which is, IMO, a little overkill). Whatever we do, I want to make sure it doesn't feel someone opening the getting started guide that step one of the guide is go read this giant document on installing and configuring Crossplane. I could see two ways to do that:

1. Duplicate the _bare minimum_ amount of detail as part of the getting-started guide. Just enough to get you up and running enough to follow the guide and nothing more.
1. Update the existing install and configure guides to have the bare minimum up front, with all the scary details hidden away somewhere that you don't need to look unless you really do want to learn how to make a GCP project or whatever.

docs/getting-started/install-configure.md
```diff
@@ -0,0 +1,299 @@
...
+<div class="tab-pane fade in active" id="setup-mac-brew" markdown="1">
+For macOS via Homebrew use the following:
+
+```
+brew upgrade
+brew install kind
+brew install kubectl
+brew install helm
+
+kind create cluster --image kindest/node:v1.17.5 --wait 5m
```
negz, line 35: We've reworked the OAM addon such that it should (still validating) only need 1.16, because most cloud providers don't yet support 1.17 per https://github.com/crossplane/addon-oam-kubernetes-local/issues/30.

docs/getting-started/install-configure.md
```diff
@@ -0,0 +1,299 @@
...
+```
+kubectl create namespace crossplane-system
+
+helm repo add crossplane-alpha https://charts.crossplane.io/alpha
+helm install --name crossplane --namespace crossplane-system crossplane-alpha/crossplane
+```
+
+</div>
+<div class="tab-pane fade" id="install-tab-helm3-master" markdown="1">
+Use Helm 3 to install the latest `master` pre-release version of Crossplane:
```
negz, line 99: Are we planning on keeping the existing, more detailed install and configure docs? If so can we assume folks will be using alpha for the getting started guide and leave the detail on how to install pre-release master versions (which are very likely to break this guide) to the more detailed variants?

docs/getting-started/install-configure.md
```diff
@@ -0,0 +1,299 @@
...
+NAME=provider-aws
+
+kubectl crossplane package install --cluster --namespace crossplane-system ${PACKAGE} ${NAME}
+```
+
+### Get AWS Account Keyfile
+
+Using an AWS account with permissions to manage RDS databases:
+```
+AWS_PROFILE=default && echo -e "[default]\naws_access_key_id = $(aws configure get aws_access_key_id --profile $AWS_PROFILE)\naws_secret_access_key = $(aws configure get aws_secret_access_key --profile $AWS_PROFILE)" > creds.conf
```
kasey, line 169: Could we guide the user to create a new access key via `aws iam create-access-key` instead? Leading a new user into pushing their current credentials into a (possibly non-local) cluster feels like a potential `record-needle-scratch.wav` moment.

docs/getting-started/install-configure.md
```diff
@@ -0,0 +1,299 @@
...
+```
+
+### Create a Provider Secret
+
+```
+kubectl create secret generic aws-creds -n crossplane-system --from-file=key=./creds.conf
+```
+
+### Configure the Provider
+Run the following to generate the `provider.yaml`:
```
kasey, line 179: "Run the following" -> "Create a file called `provider.yaml` with the following contents"

docs/getting-started/install-configure.md
prasek: sounds good, happy to adjust down to 1.16 after you've validated all is well

docs/getting-started/install-configure.md
prasek: I'm hoping that detailed install and config docs can go away with this.

docs/getting-started/install-configure.md
prasek: yes that's a better way IMO, but suggest we do that with a follow-on PR

docs/getting-started/install-configure.md
prasek: fixed

docs/getting-started/install-configure.md
prasek: The alternative to to create the `ClusterPackageInstall` yaml and apply it, but happy to change it back, so crossplane-cli can be more optional.

prasek: The alternative to to create the `ClusterPackageInstall` yaml and apply it, but happy to change it back, so crossplane-cli can be more optional.

docs/getting-started/install-configure.md
```diff
@@ -0,0 +1,384 @@
...
+PACKAGE=crossplane/provider-alibaba:master
+NAME=provider-alibaba
+
+kubectl crossplane package install --cluster --namespace crossplane-system ${PACKAGE} ${NAME}
+```
+
+### Create a Provider Secret
+
+```
+kubectl create secret generic alibaba-creds --from-literal=accessKeyId=<your-key> --from-literal=accessKeySecret=<your-secret>
```
hasheddan, line 333: There might be an easier way to get these values but I don't know it immediately. They are stored in `~/.aliyun/config.json` but I think what we have here is good :+1:

docs/getting-started/install-configure.md
prasek: - aks supports 1.16.7
- gke supports 1.16.8
- eks supports 1.16.8

docs/getting-started/install-configure.md
prasek: although IIRC they limit you to 2 access keys, so wouldn't want to force that:
https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_iam-limits.html

PR #1539, opened 2020-05-19
Use postgres port provided in connection secret

hasheddan:
Signed-off-by: hasheddan <georgedanielmangum@gmail.com>

### Description of your changes

Adds using port from connection secret in case it differs from the default.

### How has this code been tested?

Tested with `alibaba` and `see-db` pod.

[skip ci]

### Checklist

I have:
- [x] Run `make reviewable` to ensure this PR is ready for review.
- [x] Ensured this PR contains a neat, self documenting set of commits.
- [x] Updated any relevant [documentation] and [examples].
- [ ] Reported all new error conditions into the log or as an event, as
  appropriate.

For more about what we believe makes a pull request complete, see our
[definition of done].

[documentation]: https://github.com/crossplane/crossplane/tree/master/docs
[examples]: https://github.com/crossplane/crossplane/tree/master/cluster/examples
[definition of done]: https://github.com/crossplane/crossplane/tree/master/design/one-pager-definition-of-done.md

PR #1546, opened 2020-05-19
Support injecting connection secrets to composite resources

negz:

Fixes https://github.com/crossplane/crossplane/issues/1550
Closes https://github.com/crossplane/crossplane/pull/1539

This PR allows a composed resource to specify the name and value of a connection detail that should be aggregated up to the composite resource. One example of this pattern being useful is for our getting started guide, which creates a PostgreSQL database using Alibaba, AWS, Azure, and GCP. Of these clouds only AWS and Alibaba include a port in their connection secret.

As well as helping compositions satsify a definition in a portable fashion, this allows composition authors to expose fixed, non-sensitive connection details (e.g. ports, regions, protocols, database versions) about their composed resources via a secret that can be loaded by a pod as env vars or config files.

I've run through "Publish Infrastructure" and "Run an Application" sections of the getting started guide on AWS, GCP, and Azure and confirmed that in all cases I can spin up a database and connect to it from the Service Tracker.

### Checklist

I have:
  appropriate.

[definition of done].


docs/getting-started/publish-infrastructure.md
```diff
@@ -410,6 +413,8 @@ spec:
         - fromConnectionSecretKey: username
         - fromConnectionSecretKey: password
         - fromConnectionSecretKey: endpoint
+        - name: port
+          value: NTQzMg==  # The string 5432 base64 encoded.
```
negz, line 417: It's arguably a little odd to specify these values here next to the composed resource, since they really just bubble up to the composite resource regardless of which composed resource they came from. I figured it was nice to add it here for two reasons:

* It's nice to conceptually model which composed resource an injected key is associated with - i.e. in this case the ports are for the underlying PostgreSQL instances, not the supporting infrastructure.
* It was doable without a breaking API change.

apis/apiextensions/v1alpha1/composition_types.go
```diff
@@ -302,7 +302,16 @@ type ConnectionDetail struct {
...
+	// +optional
+	FromConnectionSecretKey *string `json:"fromConnectionSecretKey,omitempty"`
+
+	// Value that will be propagated to the connection secret of the composition
+	// instance. Typically you should use FromConnectionSecretKey instead, but
+	// an explicit value may be set to inject a fixed, non-sensitive connection
+	// secret values, for example a well-known port. Supercedes
+	// FromConnectionSecretKey when set.
+	// +optional
+	Value []byte `json:"value,omitempty"`
```
muvaf, line 314: Why `[]byte`?

negz: I figured it was more consistent with the kind of data it could be used for, which will be written to a secret and live alongside things read from a secret. It is a pretty gross experience to have to write what will very often be strings as base64 encoded data though; I could be convinced to switch it to `*string` to make it friendlier.

muvaf: Yeah, I think `*string` would be much more user friendly. AFAIK, the reason secrets have base64-encoded value is about the ability to store binary data. While in ConfigMaps, they introduced `binaryData` field for base64-encoded data alongside normal string data. I doubt it, but if in the future we need to store binary data, it could be another field.

docs/getting-started/publish-infrastructure.md
negz: I suspect the last key wins.

The logic here probably needs some attention. We may want to raise an error if a key is specified twice. Also, I believe the InfraDef's set of connection secret keys are currently used as a filter, to ensure _at most_ those keys bubble up from the composed resources. I'd prefer if it was more of a contract, requiring that each of its specified keys was supplied exactly once by a composed resource.

apis/apiextensions/v1alpha1/composition_types.go
negz: Switched to `*string`.

docs/getting-started/publish-infrastructure.md
```diff
@@ -485,6 +490,8 @@ spec:
...
```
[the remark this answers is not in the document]
negz: I ended up switching it back to `*string` - I don't think the symmetry with secrets is worth the poor/weird UX.

negz:
@hasheddan @muvaf Switched to using a `*string` value for injected secrets, and validated again on GCP (with port injected) and AWS (with port from composed secret).

docs/getting-started/publish-infrastructure.md
hasheddan: Agreed, I think that this can be addressed in a follow-up and we can move ahead with this PR :+1:

PR #1549, opened 2020-05-20
better quickstart for application

resouer:
Minor fix to make the documentation simpler and easier to understand.

### DIFF
--- a/docs/getting-started/run-applications.md
+++ b/docs/getting-started/run-applications.md
@@ -33,12 +33,38 @@ on-demand by Crossplane.
 > published a PostgreSQLInstance with at least one working Composition in order
 > to create the OAM application we'll use in this guide.
 
-## Install the OAM Addon
+## Infrastructure Operator
+
+### Install workloads and traits
 
 As the infrastructure operator our work is almost done - we defined, published,
-and composed the infrastructure  that our application developer and operator
-team-mates will use in the previous guide. One task remains for us, which is to
-install and configure Crossplane's OAM addon.
+and composed the infrastructure that our application developer and operator
+teammates will use in the previous guide. One task remains is, which is to define 
+the [_workloads_] and [_traits_] that our platform supports. 
+ 
+OAM applications consist of workloads, each of which may be modified by traits. 
+The infrastructure operator may choose which workloads and traits by creating 
+or deleting `WorkloadDefinitions` and `TraitDefinitions` like below:
+
+```yaml
+---
+apiVersion: core.oam.dev/v1alpha2
+kind: WorkloadDefinition
+metadata:
+  name: containerizedworkloads.core.oam.dev
+spec:
+  definitionRef:
+    name: containerizedworkloads.core.oam.dev
+``` 
+Run the following command to add support for all the workloads and traits required
+by this guide:
+
+```console
+kubectl apply -f https://raw.githubusercontent.com/crossplane/crossplane/master/docs/snippets/run/definitions.yaml
+```
+
+Among them, OAM core workloads and traits will need to install corresponding controllers 
+as OAM addon:
 
 <ul class="nav nav-tabs">
 <li class="active"><a href="#install-tab-helm3" data-toggle="tab">Helm 3</a></li>
@@ -74,30 +100,9 @@ helm install --name addon-oam-kubernetes-local --namespace crossplane-system cro
 </div>
 </div>
 
-```yaml
----
-apiVersion: core.oam.dev/v1alpha2
-kind: WorkloadDefinition
-metadata:
-  name: containerizedworkloads.core.oam.dev
-spec:
-  definitionRef:
-    name: containerizedworkloads.core.oam.dev
-```
+## Application Developer
 
-Once the addon is installed we must define the [_workloads_] and [_traits_] that
-our platform supports. OAM applications consist of workloads, each of which may
-be modified by traits. The infrastructure operator may choose which workloads
-and traits their platform supports by creating or deleting `WorkloadDefinitions`
-and `TraitDefinitions`. We'll discuss workloads and traits in more detail below.
-Run the following command to add support for the workloads and traits required
-by this guide:
-
-```console
-kubectl apply -f https://raw.githubusercontent.com/crossplane/crossplane/master/docs/snippets/run/definitions.yaml
-```
-
-## Publish Application Components
+### Publish Application Components
 
 Now we'll play the role of the application developer. Our Service Tracker
 application consists of a UI service, four API services, and a PostgreSQL
@@ -319,7 +324,9 @@ All OAM components configure a kind of workload, and any kind of Kubernetes
 resource may act as an OAM workload as long as an infrastructure operator has
 allowed it to by authoring a `WorkloadDefinition`.
 
-## Run The Application
+## Application Operator
+
+### Run The Application
 
 Finally, we'll play the role of an application operator and tie together the
 application components and infrastructure that our application developer and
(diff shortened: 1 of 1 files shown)


resouer:
/cc @negz @hasheddan 

I adjusted sequence of several sentences so it reads better (at least to me) than before.

docs/getting-started/run-applications.md
```diff
@@ -22,7 +22,7 @@ native application. In this getting started guide:
 * _Application Operators_ compose, deploy, and run application configurations.
 
 We'll play the roles of each of these team members as we deploy an application -
-Service Tracker - that consists of several services. One of these services, the
+Service Tracker - that consists of several micro services. One of these services, the
```
negz, line 25: I'd prefer to just say "services" without putting any comment on the size or shape of those services. There's a recent trend away from assuming all distributed systems are made of _micro_ services.

docs/getting-started/run-applications.md
```diff
@@ -33,12 +33,38 @@ on-demand by Crossplane.
...
+  definitionRef:
+    name: containerizedworkloads.core.oam.dev
+``` 
+Run the following command to add support for the build-in workloads and traits required
+by this guide:
+
+```console
+kubectl apply -f https://raw.githubusercontent.com/crossplane/crossplane/master/docs/snippets/run/definitions.yaml
+```
+
```
negz, line 66: Can you help me understand why you feel that it's better to define workloads before installing the addon, rather than after?

docs/getting-started/run-applications.md
```diff
@@ -33,12 +33,38 @@ on-demand by Crossplane.
...
+apiVersion: core.oam.dev/v1alpha2
+kind: WorkloadDefinition
+metadata:
+  name: containerizedworkloads.core.oam.dev
+spec:
+  definitionRef:
+    name: containerizedworkloads.core.oam.dev
+``` 
+Run the following command to add support for the build-in workloads and traits required
+by this guide:
```
negz, line 61: This should be "built-in", not "build-in". I suggest we remove or rephrase "built-in" though - it's not clear to me what it means in the context of this guide.

docs/getting-started/run-applications.md
```diff
@@ -33,12 +33,38 @@ on-demand by Crossplane.
...
-## Install the OAM Addon
+## Infrastructure Operator
+
+### Install OAM build-in workloads and traits
 
 As the infrastructure operator our work is almost done - we defined, published,
 and composed the infrastructure  that our application developer and operator
-team-mates will use in the previous guide. One task remains for us, which is to
-install and configure Crossplane's OAM addon.
+team-mates will use in the previous guide. One task remains is, we must define
```
negz, line 42: The original "One task remains for us, which is to" is more grammatically correct than the proposed "One task remains is, we must".

docs/getting-started/run-applications.md
```diff
@@ -33,12 +33,38 @@ on-demand by Crossplane.
...
 and composed the infrastructure  that our application developer and operator
-team-mates will use in the previous guide. One task remains for us, which is to
-install and configure Crossplane's OAM addon.
+team-mates will use in the previous guide. One task remains is, we must define 
+the [_workloads_] and [_traits_] that our platform supports. We'll discuss workloads 
+and traits in more detail later.
+
+The infrastructure operator could choose which workloads
+and traits their platform supports by creating or deleting `WorkloadDefinitions`
+and `TraitDefinitions` like below:
```
negz, line 48: Is the goal of your proposed sentence to replace the below paragraph, which you've removed further down? Can you explain what your thinking is here?

> Once the addon is installed we must define the [_workloads_] and [_traits_] that our platform supports. OAM applications consist of workloads, each of which may be modified by traits. The infrastructure operator may choose which workloads and traits their platform supports by creating or deleting `WorkloadDefinitions` and `TraitDefinitions`. We'll discuss workloads and traits in more detail below.

docs/getting-started/run-applications.md
```diff
@@ -33,12 +33,38 @@ on-demand by Crossplane.
 > published a PostgreSQLInstance with at least one working Composition in order
 > to create the OAM application we'll use in this guide.
 
-## Install the OAM Addon
+## Infrastructure Operator
```
negz, line 36: I was considering adding headings like this too - I think they're a good idea.

docs/getting-started/run-applications.md
resouer: ACK, I thought it's a typo.

docs/getting-started/run-applications.md
kasey: Incidentally I think the preexisting hyphenation of "team-mate" is old-fashioned/British. "teammate" would be more conventional :)

docs/getting-started/run-applications.md
resouer: No. That paragraph is just split and place to different place.

docs/getting-started/run-applications.md
resouer: These controllers are for [core](https://github.com/oam-dev/spec/blob/master/3.workload.md#core-workload) and [standard](https://github.com/oam-dev/spec/blob/master/3.workload.md#standard-workload) OAM workloads/traits/scopes. Feel free to find a better word or I will just use core and standard instead.

docs/getting-started/run-applications.md
resouer: If you think `Definitions` as CRD (which is the intention), "addon" as controllers (which indeed are), it would be easier to get my point for this change.

docs/getting-started/run-applications.md
negz: The controllers are, but keep in mind we're also adding a workload definition for `MySQLInstanceRequirement`, which is not built-in/core/standard.

docs/getting-started/run-applications.md
negz: I have no feelings either either way on team-mate/teammate. I think I copied the hypenated example from another document then stuck with that for consistency.

docs/getting-started/run-applications.md
negz: Can you help me understand where and why?

As meta feedback it's a little hard to review when the context of this PR is "make the guide better", without explaining what you felt needed improving about the guide or how you feel you've improved it. What is or is not good documentation is highly subjective. I'd definitely open to improving this guide, but some of these changes feel a little arbitrary without context.

docs/getting-started/run-applications.md
resouer: > The controllers are,

The controllers are what we installed below, no?

docs/getting-started/run-applications.md
resouer: Can we migrate this comment to https://github.com/crossplane/crossplane/pull/1549#discussion_r427740956?

These two comments are duplicated essentially.

resouer:
@negz It seems Github diff will actually mess up the sequence of comments so it's hard for you to review with consistency. Let's put it here.

There's only one idea in this change (ref https://github.com/crossplane/crossplane/pull/1549#discussion_r427740956): I propose it's more natural to install definitions first and then controllers. Just like we normally install CRDs first and then controllers.

As for the name of [these controllers]( https://github.com/crossplane/addon-oam-kubernetes-local), they are controllers OAM core + standard workloads/traits/scopes. Open for any name proposal here.

docs/getting-started/run-applications.md
resouer: Oh, built-in should be deleted from this line. The built-in is for name of controllers.

docs/getting-started/run-applications.md
```diff
@@ -33,12 +33,38 @@ on-demand by Crossplane.
...
+``` 
+Run the following command to add support for all the workloads and traits required
+by this guide:
+
+```console
+kubectl apply -f https://raw.githubusercontent.com/crossplane/crossplane/master/docs/snippets/run/definitions.yaml
+```
+
+Among them, OAM core workloads and traits will need to install corresponding controllers 
+as OAM addon:
```
negz, line 67:
```suggestion
Now that we've defined our workloads and traits, we must install Crossplane's OAM addon. This addon
packages the controllers that reconcile core OAM workloads and traits.
```

resouer:
Much appreciated! And thank you for fixing the gramma nits (it's hard for me to catch them indeed).

PR #1556, opened 2020-05-22
Block deletion of local KAR until remote resource is gone

muvaf:

There are cases, especially in AWS, where the deletion call goes to `Service` object but it takes a while until all LB infra is detached. KAR should be kept in the local until `Service` resource is confirmed to be gone so that users don't leak it since the cluster is immediately gone when its deletion call is made to AWS.

Will test it.

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/pkg/controller/workload/kubernetes/resource/resource.go
+++ b/pkg/controller/workload/kubernetes/resource/resource.go
@@ -57,16 +57,17 @@ const (
 )
 
 var (
-	errUnmarshalTemplate = "cannot unmarshal template"
-	errUpdateStatusFmt   = "cannot update status %s %s"
-	errGetKey            = "cannot take object key from resource"
-	errDeleteResource    = "cannot delete resource"
-	errGetResource       = "cannot get resource"
-	errSyncResource      = "cannot sync resource"
-	errCreateResource    = "cannot create resource"
-	errDeleteSecret      = "cannot delete secret"
-	errGetSecret         = "cannot get secret"
-	errSyncSecret        = "cannot sync secret"
+	errUnmarshalTemplate  = "cannot unmarshal template"
+	errUpdateStatusFmt    = "cannot update status %s %s"
+	errGetKey             = "cannot take object key from resource"
+	errDeleteResource     = "cannot delete resource"
+	errGetResource        = "cannot get resource"
+	errSyncResource       = "cannot sync resource"
+	errCreateResource     = "cannot create resource"
+	errDeleteSecret       = "cannot delete secret"
+	errGetSecret          = "cannot get secret"
+	errSyncSecret         = "cannot sync secret"
+	errWaitingForDeletion = "waiting for resource to be deleted in remote cluster"
 )
 
 // Ownership annotations
@@ -299,6 +300,9 @@ func (c *unstructuredClient) delete(ctx context.Context, template *unstructured.
 		return errors.Wrap(err, errGetKey)
 	}
 
+	// We declare deletion as successful only if the remote resource is actually
+	// deleted so that we don't delete the local resource in case deletion
+	// results in error.
 	if err := c.kube.Get(ctx, key, remote); err != nil {
 		if kerrors.IsNotFound(err) {
 			return nil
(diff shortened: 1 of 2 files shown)


pkg/controller/workload/kubernetes/resource/resource.go
```diff
@@ -311,8 +315,10 @@ func (c *unstructuredClient) delete(ctx context.Context, template *unstructured.
 		return nil
 	}
 
-	// The object exists and we own it. Delete it.
-	return errors.Wrap(c.kube.Delete(ctx, remote), errDeleteResource)
+	if err := c.kube.Delete(ctx, remote); err != nil {
+		return errors.Wrap(err, errDeleteSecret)
+	}
+	return errors.New(errWaitingForDeletion)
```
negz, line 321: I'm not a huge fan of using an error as control flow for a seemingly non-error condition here. Would it be a lot riskier or more work to change the signature of these types to return some information? Perhaps:

```go
type Deletion struct {
    Done bool
}

# Maybe just (bool, error) if we don't think we're likely to need to return more information in future?
func (c *unstructuredClient delete(ctx context.Context, template *unstructured.Unstructured) (Deletion, error) {
  # ...
}
```

negz:
@muvaf I noticed @hasheddan mentioned that he's not confident this will solve 100% of the leak issues we see, but it seems like it will at least help with the problem. Is that a fair assessment?

hasheddan:
@negz that is my thought. I don’t think it is going fix _all_ the problems, but at the very least will be better logic and likely will fix some cases. I would like to do a bit more review here though.

muvaf:
> he's not confident this will solve 100% of the leak issues we see, but it seems like it will at least help with the problem. Is that a fair assessment?

We reduced the probability with ordered deletion by making sure we make the `Delete` call to the `Service` in remote, which is an improvement over racing deletion of `KubernetesCluster` but it still misses the case where deletion takes long.

I think there was an observation by @hasheddan  that LB `Service` resource in the remote cluster was gone but AWS LBs were still leaked but I didn't see that case happening in my tests. So, unless AWS leaks LBs even though `Service` resource is gone from the cluster, this PR should fix the remaining cases.

I'm testing this right now and will report back but I believe it's a pretty rare case that AWS leaks LB if its service controller allowed the `Service` resource to be gone.

pkg/controller/workload/kubernetes/resource/resource.go
muvaf: I just went for the cheapest option but will look at improving it after testing that it solves the problem.

muvaf:
@hasheddan I observed the case where everything in remote cluster is deleted fine but VPC still didn't delete and got stuck in dependency violation. The problem for that case is that there are 3 route tables in AWS when you install AWS Sample Stack:
* Main one, created automatically by AWS
* Custom, associated with 3 subnets we created. Created by sample stack.
* Custom, non-main, automatically created.

The third one is not known by Crossplane and not related to Wordpress as far as I can see because it's created before Wordpress app is even deployed. When you delete the sample stack, the third one causes the dependency violation because no one deletes it. Tested creating only a vpc and it didn't get created. So, I suspect there is some funny business going on in routetable controller but I will get to that after solving the problem in this PR.

pkg/controller/workload/kubernetes/resource/resource.go
muvaf: Tested and it does block KAR deletion until remote resources are gone.

> but will look at improving it after testing that it solves the problem.

@negz In the whole tree of calls, there is the assumption that deletion is made once and as long as it doesn't fail, it succeeds and it succeeds immediately. Since this will go into a patch release, I'd prefer shipping it like this and do another iteration for 0.12. Maybe a big refactor of this KA and KAR controllers.

pkg/controller/workload/kubernetes/resource/resource.go
```diff
@@ -365,8 +375,10 @@ func (c *secretClient) delete(ctx context.Context, template *corev1.Secret) erro
 		return nil
 	}
 
-	// We own the existing object. delete it.
-	return errors.Wrap(c.kube.Delete(ctx, remote), errDeleteSecret)
+	if err := c.kube.Delete(ctx, remote); err != nil {
+		return errors.Wrap(err, errDeleteResource)
+	}
+	return errors.New(errWaitingForDeletion)
```
hasheddan, line 381: @muvaf this is likely inconsequential, but just wondering if we need to do this for secrets as well. It is unlikely that they hang or are tied to any external infrastructure. Seems like we could avoid some additional errors if we just were okay with making a delete call and not doing the check that they no longer exist. Open to leaving this how you have it if you feel strongly though. Might be easier to reason about later if we do the same thing for secrets and other resources

hasheddan:
@muvaf integration tests failed due to failure to download `kubectl` which is likely a result of GH outages. Restarted :heavy_check_mark:

muvaf:
> integration tests failed due to failure to download kubectl which is likely a result of GH outages. Restarted 

@hasheddan thanks!

I added the comment.

hasheddan:
@muvaf looks like you may have to do a force push to get DCO to stop hanging

muvaf:
@negz should be good to go

PR #1567, opened 2020-05-22
KubernetesApplication manages the deletion of KARs

muvaf:

KA needs to delete KARs before it's removed from the cluster so that it guarantees the remote resources are removed.

The original approach of this controller was letting Kubernetes GC to handle that, however, that only works if the deletion policy is `Foreground`, which is not default in kubectl or controller-runtime clients. This PR allows the deletion to work as expected in the default `Background` mode.

Manually tested.

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/pkg/controller/workload/kubernetes/application/application.go
+++ b/pkg/controller/workload/kubernetes/application/application.go
@@ -44,12 +44,17 @@ const (
 	shortWait        = 30 * time.Second
 	longWait         = 1 * time.Minute
 
-	errGarbageCollect = "failed to garbage collect KubernetesApplicationResources"
-	errSyncTemplate   = "failed to sync template with KubernetesApplicationResource"
+	finalizerName = "finalizer.kubernetesapplication.workload.crossplane.io"
+
+	errGarbageCollect  = "failed to garbage collect KubernetesApplicationResources"
+	errSyncTemplate    = "failed to sync template with KubernetesApplicationResource"
+	errAddFinalizer    = "failed to add finalizer"
+	errRemoveFinalizer = "failed to remove finalizer"
 )
 
 type syncer interface {
 	sync(ctx context.Context, app *v1alpha1.KubernetesApplication) (v1alpha1.KubernetesApplicationState, error)
+	delete(ctx context.Context, app *v1alpha1.KubernetesApplication) (int, error)
 }
 
 // CreatePredicate accepts KubernetesApplications that have been scheduled to a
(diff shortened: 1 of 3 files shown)


pkg/controller/workload/kubernetes/application/application.go
```diff
@@ -271,6 +286,29 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error)
 		return reconcile.Result{Requeue: false}, errors.Wrapf(err, "cannot get %s %s", v1alpha1.KubernetesApplicationKind, req.NamespacedName)
 	}
 
+	if meta.WasDeleted(app) {
```
hasheddan, line 289: @muvaf since the KARs have owner references back to the KA, do we actually need to manually call delete, or can we just add the finalizer and keep checking if the KARs have been deleted here?

muvaf: Yes, because:
* background deletion: delete child resources after the parent disappears
* foreground deletion: delete child resources immediately, if all is gone then make the parent disappear

and the default is background deletion. We don't let the parent to disappear via finalizer, so, children never get deletion calls if we don't call deletion on them.

muvaf:
I have tested adding `foregroundDeletion` finalizer to enforce foreground deletion policy. It didn't have any effect on the policy; KA got removed immediately and KARs got deletion calls but their existence didn't block KA's deletion. `foregroundDeletion` finalizer didn't even have a normal finalizer effect. It was removed right when you make the `kubectl delete` call.

@negz @hasheddan So, I'd suggest us moving forward with the idea in this PR.

pkg/controller/workload/kubernetes/application/application.go
```diff
@@ -271,6 +286,29 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error)
...
+		app.Status.State = v1alpha1.KubernetesApplicationStateDeleting
+		remaining, err := r.local.delete(ctx, app)
```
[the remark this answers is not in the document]
muvaf: That's what I did initially but then the returned state gets messy and deletion is better handled at the top level reconcile, so I did it here.

pkg/controller/workload/kubernetes/application/application.go
```diff
@@ -146,6 +145,21 @@ func (c *localCluster) sync(ctx context.Context, app *v1alpha1.KubernetesApplica
 	return v1alpha1.KubernetesApplicationStateSubmitted, nil
 }
 
+func (c *localCluster) delete(ctx context.Context, app *v1alpha1.KubernetesApplication) (int, error) {
```
negz, line 148: @muvaf Would it make sense for this to be a method on `applicationResourceSyncer`? That way I believe we'd keep the `client.Client` contained there, rather than needing to add it to the `localCluster` type as well.

negz: Actually, I see now that that probably doesn't make sense. Could we instead follow the pattern of `applicationResourceSyncer` with an `applicationResourceDeleter`, i.e. make this a simple injected dependency of `localCluster` rather than a method? I know we don't love the logic that's there now, but it would be ideal to be consistent with the existing patterns and separation of concerns if possible.

muvaf: There are a few abstraction layer here. As far as I see, `applicationResourceSyncer.sync` is called for an individual KAR while `localCluster.sync` is called for the whole KA, hence it's running a `for` loop to iterate through the KARs.

For deletion, we need to call the following per resource:
```
err := c.kube.Delete(ctx, kar)
```

If I understand what you're saying, you'd like `applicationResourceDeleter` on `localCluster` with a `delete` function that only has `c.kube.Delete(ctx, kar)` and then we'd have `localCluster.delete` to call that function like `c.ad.delete` for every KAR, right?

Isn't that kind of too much for a function with one delete call?

negz: > Isn't that kind of too much for a function with one delete call?

I agree it's too many layers of abstraction. I have a PR (pending Github 500 issues) that removes a layer of abstraction.

PR #1568, opened 2020-05-22
Remove an unnecessary layer of abstraction from the KubeApp controller

negz:

In https://github.com/crossplane/crossplane/pull/1567 we added explicit `delete` functionality. @muvaf pointed out that it felt like too much abstraction to push deletion down into the `applicationResourceSyncer`, but I also didn't like that `localCluster` went from being a layer that only interacted with the Kubernetes API via other types, to interacting with the Kubernetes API both directly and through other types. Removing a pointless layer of abstraction (`applicationResourceClient`) felt like the best path forward here. I suspect this client only existed in the first place to let us stub out `controllerutil.CreateOrUpdate`, which we no longer use.

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/pkg/controller/workload/kubernetes/application/application.go
+++ b/pkg/controller/workload/kubernetes/application/application.go
@@ -46,7 +46,6 @@ const (
 
 	finalizerName = "finalizer.kubernetesapplication.workload.crossplane.io"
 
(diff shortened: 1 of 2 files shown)


muvaf:
@negz Tested this together with https://github.com/crossplane/crossplane/pull/1556 and it's working as expected.

PR #1571, opened 2020-05-23
Database engine fixes for the getting started guide

negz:

This PR fixes the AWS RDS engine version (there's an inconsistency in that most options are "foo__sql__", but PostgreSQL is just "postgres". It also speculatively changes the Alibaba engine from postgresql to PostgreSQL, which the API docs state is a valid value.

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/docs/snippets/provision/alibaba.yaml
+++ b/docs/snippets/provision/alibaba.yaml
@@ -4,7 +4,7 @@ metadata:
   name: rdspostgresql
 spec:
   forProvider:
-    engine: postgresql
+    engine: PostgreSQL
     engineVersion: "9.4"
     dbInstanceClass: rds.pg.s1.small
     dbInstanceStorageInGB: 20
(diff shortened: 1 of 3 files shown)


docs/snippets/provision/alibaba.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: rdspostgresql
 spec:
   forProvider:
-    engine: postgresql
+    engine: PostgreSQL
```
hasheddan, line 7: I _think_ either of these will work, but a new contributor had a successful run with existing today I believe

negz: I managed to come across some credentials and validated that `PostgreSQL` works. It's possible lowercase worked too, but either way we may as well be consistent between the two guides.

PR #1577, opened 2020-05-25
Allow passing non-zero fsGroup in package deployments

hasheddan:

Setting the fsGroup for a pod means that any volumes that support ownership management are to be modified to be owned and writable by the GID specified. In addition, the GID will be added as a supplementary group for containers in that pod. This is specifically useful in the case where a projected Service Account token is mounted, as Kubernetes currently sets the mode to 0600 always, which may cause non-root containers to be unable to read the file.

Fixes https://github.com/crossplane/provider-aws/issues/211

/cc @Niksko @mper0003

Draft until can be verified with steps in https://github.com/crossplane/provider-aws/blob/master/AUTHENTICATION.md

### Checklist

I have:
  appropriate.

[definition of done].


pkg/controller/packages/pkg/pkg.go
```diff
@@ -87,9 +87,10 @@ var (
 
 	disableAutoMount = false
 
-	allowPrivilegeEscalation = false
-	privileged               = false
-	runAsNonRoot             = true
+	allowPrivilegeEscalation       = false
+	privileged                     = false
+	runAsNonRoot                   = true
+	fsGroup                  int64 = 65534
```
turkenh, line 93: AFAIU, nobody user is expected to have least privilege possible: https://en.wikipedia.org/wiki/Nobody_(username).

If this is being added as supplementary group for containers, do we really have to use `nobody` user? Wouldn't it still work if we just set this to another value (e.g. 2000) other than 65534 and be more secure?

hasheddan: > Wouldn't it still work if we just set this to another value (e.g. 2000) other than 65534 and be more secure?

@turkenh can you expound on why setting to `2000` is more secure in this context than the `nobody` user?

turkenh: This is what I understand after reading about that special user.

https://askubuntu.com/a/329716/740255

> It's there to run things that don't need any special permissions. It's usually reserved for vulnerable services (httpd, etc) so that if they get hacked, they'll have minimal damage on the rest of the system.
> 
> Contrast this with running something as a real user, if that service were compromised (web servers are occasionally exploited to run arbitrary code), it would run as that user and have access to everything that user had. In most cases, this is as bad as getting root.

hasheddan: We are in a tough position here because if we went the route proposed here then we are basically saying that if you attach a volume to your controller `Deployment` then `nobody` will be given access. While this may not be detrimental in many cases, it does feel like a bad pattern to apply across the board. We find ourselves in this situation because we are overwriting the `SecurityContext` by default to make sure containers do not run as root. One thing we could do here is let an `fsGroup` set in an `install.yaml` still flow through. So we would take an `install.yaml`, overwrite the `SecurityContext`, then re-insert any specified `fsGroup` into the `SecurityContext`. If the `install.yaml` includes an `fsGroup` that would give a container root privileges then it will just fail to run. This places the onus on the provider to make sure their controller deployment is configured directly, which is the same path we followed when restricting running as root but not explicitly setting the `USER`.

How does this sound @turkenh?

hasheddan:
Note: this will likely resolve itself in 1.19 (see my comment here: https://github.com/aws/amazon-eks-pod-identity-webhook/issues/8#issuecomment-636888074). That being said, it will likely be quite a while before most users are on 1.19. So it would be good to go ahead and address this.

pkg/controller/packages/pkg/pkg.go
lukeweber: I feel like @hasheddan last solution seems reasonable. Will the installer error if fsGroup is set to 0 or are you assuming the other config that we've set would make 0 fail due to runAsRoot=false?

hasheddan: I was assuming the `runAsRoot=false` would catch it, but we could explicitly check as well. From what I could tell in looking through the source code today, I believe that k8s does in fact make sure this case does not occur, but I would want to test to make sure.

turkenh: Yes, this sound like a better solution as long as we test the possible edge cases like above.

hasheddan: So I don't think that `runAsRoot=false` will restrict setting `fsGroup: 0`, so we would need to explicitly restrict. Note that being added to the root group is not the same as being a root user (there are not actually specific kernel privileges given to GID 0). That being said, it is not uncommon for PSPs to restrict using `fsGroup: 0` so I think it would be fine for us to do so here as well.

hasheddan:
@muvaf we would theoretically only need to update `provider-aws` in the short-term. I will move this out of draft after some additional testing. It is also worth noting that this model may change in the future if the packaging format changes, but I think this is a useful workaround for now.

hasheddan:
Marking this as ready for review as I have tested it successfully with https://github.com/crossplane/provider-aws/pull/242 👍

hasheddan:
cc @negz @jbw976

hasheddan:
Ready for final review here. /cc @negz @jbw976

pkg/controller/packages/pkg/pkg_test.go
```diff
@@ -1412,7 +1421,88 @@ func TestProcessDeployment(t *testing.T) {
...
+						Namespace: namespace,
+						Labels:    packagespkg.ParentLabels(resource(withControllerSpec(defaultControllerSpec()))),
+					},
+					Spec: *deploymentSpec(
+						withDeploymentTmplMeta(controllerDeploymentName, "", nil),
+						withDeploymentMatchLabels(map[string]string{"app": controllerDeploymentName}),
+						withDeploymentSA(resourceName),
+						withDeploymentContainer(controllerContainerName, controllerImageName),
+						withDeploymentSecurityContext(&corev1.PodSecurityContext{
+							RunAsNonRoot: &runAsNonRoot,
```
jbw976, line 1504: the `FSGroup` is not set on the `want` object in this test case because we do not allow rootGroup 0 to be set unless the user has set the insecure allow full deployment flag, right?

pkg/controller/packages/pkg/pkg.go
```diff
@@ -856,9 +856,13 @@ func (h *packageHandler) prepareDeployment(d *apps.Deployment) {
 	d.Spec.Template.Spec.ServiceAccountName = h.ext.Name
 
 	if !h.allowFullDeployment {
-		d.Spec.Template.Spec.SecurityContext = &corev1.PodSecurityContext{
+		psc := &corev1.PodSecurityContext{
 			RunAsNonRoot: &runAsNonRoot,
 		}
+		if d.Spec.Template.Spec.SecurityContext != nil && d.Spec.Template.Spec.SecurityContext.FSGroup != nil && *d.Spec.Template.Spec.SecurityContext.FSGroup != 0 {
```
jbw976, line 862: nit: there was great discussion in the PR around why/how we landed on blocking fsgroup 0 here also.  consider including a brief summary as a comment so that decision/policy lives with the code as well.

hasheddan: Sounds good, will update :+1:

hasheddan: @jbw976 this is now updated. If you wouldn't mind taking a look at the note and making sure that it matches your understanding as well that would be much appreciated :)

PR #1580, opened 2020-05-27
Correct Postgresql engine version

zzxwill:

Fix issue [https://github.com/crossplane/provider-alibaba/issues/23](https://github.com/crossplane/provider-alibaba/issues/23)

### Checklist

I have:
  appropriate.

[definition of done].


zzxwill:
Please hold on, this modification might affect the engine version for AWS/GCP/Azure, sorry. Let me think it over.

cluster/charts/crossplane-types/crds/database.crossplane.io_postgresqlinstances.yaml
```diff
@@ -146,10 +146,12 @@ spec:
               type: object
             engineVersion:
               description: 'EngineVersion specifies the desired PostgreSQL engine
-                version. Allowed Versions: 9.6 and 11.'
+                version. Allowed Versions: 9.4, 10.0, 11.0 and 12.0.'
```
hasheddan, line 149: @zzxwill just a heads up, if we want to modify this we would do so in the go code, then it will be generated here with `make generate`. That being said, we want to use the lowest common denominator values here across all providers, so we likely would not support all of these versions. This is something claims run into frequently, so they will likely be deprecated soon because composition provides a much more robust experience :)

zzxwill: +1. Deprecation of the `engineVersion` values is great. On the one hand, composition provider will return more accurate exception information, on the other hand, these values diff from each those engine versions from Cloud Providers.

muvaf:
As @hasheddan pointed out, claims are to be deprecated. But we're still supporting them. I think we can get rid of this whole list altogether. If an unsupported version is used, provider will return an error anyway. @zzxwill removing [the line here](https://github.com/crossplane/crossplane/blob/master/apis/database/v1alpha1/types.go#L70) and running `make reviewable` should take care of this. You can do so for `MySQLInstance` as well by removing [this line](https://github.com/crossplane/crossplane/blob/master/apis/database/v1alpha1/types.go#L30).

zzxwill:
> As @hasheddan pointed out, claims are to be deprecated. But we're still supporting them. I think we can get rid of this whole list altogether. If an unsupported version is used, provider will return an error anyway. @zzxwill removing [the line here](https://github.com/crossplane/crossplane/blob/master/apis/database/v1alpha1/types.go#L70) and running `make reviewable` should take care of this. You can do so for `MySQLInstance` as well by removing [this line](https://github.com/crossplane/crossplane/blob/master/apis/database/v1alpha1/types.go#L30).

Got it. I will push another commit. Thanks.

cluster/charts/crossplane-types/crds/database.crossplane.io_postgresqlinstances.yaml
```diff
@@ -146,12 +146,10 @@ spec:
...
             engineVersion:
               description: 'EngineVersion specifies the desired PostgreSQL engine
-                version. Allowed Versions: 9.4, 10.0, 11.0 and 12.0.'
+                version. Allowed Versions: 9.6 and 11.'
               enum:
-              - "9.4"
-              - "10.0"
-              - "11.0"
-              - "12.0"
+              - "9.6"
```
muvaf: Did you run `make reviewable` after removing the validation line? I'd expect the `enum` array here to go away automatically since CRDs are deleted and regenerated via `make reviewable`

zzxwill: No. But I did just now following with another commit. Thank you, now I learnt the mechanism.

Issue #1471, opened 2020-04-30
Allow specifying a default composition in definition


muvaf:
### What problem are you facing?

There is no way to mark a `Composition` as default of a given `InfrastructureDefinition`.

### How could Crossplane help solve your problem?

Enable users to specify a `Composition` as the default on the `InfrastructureDefinition` resource.

negz:
@muvaf Could you make sure the documentation reflects this new functionality?

muvaf:
@negz done in https://github.com/crossplane/crossplane/pull/1607

Issue #1470, opened 2020-04-30
Enforced composition ref on the definition


muvaf:
### What problem are you facing?

Currently, there is no way to enforce a specific `Composition` resource for the instances of `InfrastructureDefinition`. Without this, the permission of creation an instance of the definition (`MySQLRequirement` or `MySQL`) is a vague one since the referenced `Composition` could include anything.

### How could Crossplane help solve your problem?

Optionally, make it possible to refer to a `Composition` in the `InfrastructureDefinition` so that the resulting resources of instantiation of a composition are always known and RBAC permission of that instantiation resources can be given to a `ServiceAccount` with known limits.

negz:
@muvaf Could you make sure the documentation reflects this new functionality?

muvaf:
@negz done in https://github.com/crossplane/crossplane/pull/1607

PR #1585, opened 2020-06-03
Default and Enforced Compositions

muvaf:

This PR allows infra-operators to choose a default composition to be selected in case no selector is given as well as an enforced composition to be selected no matter what.

Fixes https://github.com/crossplane/crossplane/issues/1471
Fixes https://github.com/crossplane/crossplane/issues/1470

Manually tested.

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/apis/apiextensions/v1alpha1/infrastructuredefinition_types.go
+++ b/apis/apiextensions/v1alpha1/infrastructuredefinition_types.go
@@ -17,7 +17,7 @@ limitations under the License.
 package v1alpha1
 
 import (
-	v1 "k8s.io/api/core/v1"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 
@@ -37,13 +37,13 @@ type InfrastructureDefinitionSpec struct {
 	// DefaultCompositionRef refers to the Composition resource that will be used
 	// in case no composition selector is given.
 	// +optional
-	DefaultCompositionRef *v1.ObjectReference `json:"defaultCompositionRef,omitempty"`
+	DefaultCompositionRef *corev1.ObjectReference `json:"defaultCompositionRef,omitempty"`
 
 	// EnforcedCompositionRef refers to the Composition resource that will be used
 	// by all composite instances whose schema is defined by this definition.
 	// +optional
 	// +immutable
-	EnforcedCompositionRef *v1.ObjectReference `json:"enforcedCompositionRef,omitempty"`
+	EnforcedCompositionRef *corev1.ObjectReference `json:"enforcedCompositionRef,omitempty"`
 
 	// CRDSpecTemplate is the base CRD template. The final CRD will have additional
 	// fields to the base template to accommodate Crossplane machinery.
--- a/apis/apiextensions/v1alpha1/infrastructuredefinition_types.go
+++ b/apis/apiextensions/v1alpha1/infrastructuredefinition_types.go
@@ -17,6 +17,7 @@ limitations under the License.
 package v1alpha1
 
 import (
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 
@@ -32,6 +33,11 @@ type InfrastructureDefinitionSpec struct {
 	// ConnectionSecretKeys is the list of keys that will be exposed to the end
 	// user of the defined kind.
 	ConnectionSecretKeys []string `json:"connectionSecretKeys,omitempty"`
+
+	// DefaultCompositionRef refers to the Composition resource that will be used
+	// in case no composition selector is given.
+	// +optional
+	DefaultCompositionRef *v1.ObjectReference `json:"defaultCompositionRef,omitempty"`
 
 	// CRDSpecTemplate is the base CRD template. The final CRD will have additional
 	// fields to the base template to accommodate Crossplane machinery.
--- a/apis/apiextensions/v1alpha1/infrastructuredefinition_types.go
+++ b/apis/apiextensions/v1alpha1/infrastructuredefinition_types.go
@@ -17,7 +17,6 @@ limitations under the License.
 package v1alpha1
 
 import (
-	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 
@@ -37,13 +36,13 @@ type InfrastructureDefinitionSpec struct {
 	// DefaultCompositionRef refers to the Composition resource that will be used
 	// in case no composition selector is given.
 	// +optional
-	DefaultCompositionRef *corev1.ObjectReference `json:"defaultCompositionRef,omitempty"`
+	DefaultCompositionRef *v1alpha1.Reference `json:"defaultCompositionRef,omitempty"`
 
 	// EnforcedCompositionRef refers to the Composition resource that will be used
 	// by all composite instances whose schema is defined by this definition.
 	// +optional
 	// +immutable
-	EnforcedCompositionRef *corev1.ObjectReference `json:"enforcedCompositionRef,omitempty"`
+	EnforcedCompositionRef *v1alpha1.Reference `json:"enforcedCompositionRef,omitempty"`
 
 	// CRDSpecTemplate is the base CRD template. The final CRD will have additional
 	// fields to the base template to accommodate Crossplane machinery.
--- a/apis/apiextensions/v1alpha1/infrastructuredefinition_types.go
+++ b/apis/apiextensions/v1alpha1/infrastructuredefinition_types.go
@@ -39,6 +39,11 @@ type InfrastructureDefinitionSpec struct {
 	// +optional
 	DefaultCompositionRef *v1.ObjectReference `json:"defaultCompositionRef,omitempty"`
 
+	// EnforcedCompositionRef refers to the Composition resource that will be used
+	// by all composite instances whose schema is defined by this definition.
+	// +optional
+	EnforcedCompositionRef *v1.ObjectReference `json:"enforcedCompositionRef,omitempty"`
+
 	// CRDSpecTemplate is the base CRD template. The final CRD will have additional
 	// fields to the base template to accommodate Crossplane machinery.
 	CRDSpecTemplate CRDSpecTemplate `json:"crdSpecTemplate,omitempty"`
--- a/apis/apiextensions/v1alpha1/infrastructuredefinition_types.go
+++ b/apis/apiextensions/v1alpha1/infrastructuredefinition_types.go
@@ -34,6 +34,9 @@ type InfrastructureDefinitionSpec struct {
 	// user of the defined kind.
 	ConnectionSecretKeys []string `json:"connectionSecretKeys,omitempty"`
 
+	// TODO(muvaf): A validating webhook on InfrastructureDefinition that will make
+	// sure referenced Composition is compatible could make more sense.
+
 	// DefaultCompositionRef refers to the Composition resource that will be used
 	// in case no composition selector is given.
 	// +optional
@@ -42,6 +45,7 @@ type InfrastructureDefinitionSpec struct {
 	// EnforcedCompositionRef refers to the Composition resource that will be used
 	// by all composite instances whose schema is defined by this definition.
 	// +optional
+	// +immutable
 	EnforcedCompositionRef *v1.ObjectReference `json:"enforcedCompositionRef,omitempty"`
 
 	// CRDSpecTemplate is the base CRD template. The final CRD will have additional
--- a/apis/apiextensions/v1alpha1/infrastructuredefinition_types.go
+++ b/apis/apiextensions/v1alpha1/infrastructuredefinition_types.go
@@ -34,9 +34,6 @@ type InfrastructureDefinitionSpec struct {
 	// user of the defined kind.
 	ConnectionSecretKeys []string `json:"connectionSecretKeys,omitempty"`
 
-	// TODO(muvaf): A validating webhook on InfrastructureDefinition that will make
-	// sure referenced Composition is compatible could make more sense.
-
 	// DefaultCompositionRef refers to the Composition resource that will be used
 	// in case no composition selector is given.
 	// +optional
--- a/pkg/controller/apiextensions/composite/api.go
+++ b/pkg/controller/apiextensions/composite/api.go
@@ -112,19 +112,19 @@ func (r *CompositionSelectorChain) SelectComposition(ctx context.Context, cp res
 	return nil
 }
 
-// NewAPISelectorResolver returns a SelectorResolver for composite resource.
-func NewAPISelectorResolver(c client.Client) *APISelectorResolver {
-	return &APISelectorResolver{client: c}
+// NewAPILabelSelectorResolver returns a SelectorResolver for composite resource.
+func NewAPILabelSelectorResolver(c client.Client) *APILabelSelectorResolver {
+	return &APILabelSelectorResolver{client: c}
 }
 
-// APISelectorResolver is used to resolve the composition selector on the instance
+// APILabelSelectorResolver is used to resolve the composition selector on the instance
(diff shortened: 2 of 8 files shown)


pkg/controller/apiextensions/composite/api.go
```diff
@@ -125,6 +147,45 @@ func (r *APISelectorResolver) ResolveSelector(ctx context.Context, cp resource.C
...
+
+// APIDefaultCompositionSelector selects the default composition referenced in
+// the definition of the resource if neither a reference nor selector is given
+// in composite resource.
+type APIDefaultCompositionSelector struct {
+	client client.Client
+	defRef v1.ObjectReference
+}
+
+// SelectComposition selects the default compositionif neither a reference nor
```
hasheddan, line 163:
```suggestion
// SelectComposition selects the default composition if neither a reference nor
```

pkg/controller/apiextensions/composite/api.go
```diff
@@ -82,6 +83,27 @@ func (a *APIFilteredSecretPublisher) UnpublishConnection(_ context.Context, _ re
...
+// CompositionSelectorChain calls the given list of CompositionSelectors in order.
+type CompositionSelectorChain struct {
+	list []CompositionSelector
+}
+
+// SelectComposition calls all SelectComposition functions of CompositionSelectors
+// in the list.
+func (r *CompositionSelectorChain) SelectComposition(ctx context.Context, cp resource.Composite) error {
+	for _, cs := range r.list {
+		if err := cs.SelectComposition(ctx, cp); err != nil {
```
hasheddan, line 100: Wondering if we could return a reference here and then we would avoid calling all of the composition selectors passed in. We could then pass in the `APISelectorResolver` and if a selector is specified then we will never call the `DefaultCompositionSelector`. We also would not have to check for composition references or selectors in the defaulter because they would have already been used and the chain would have exited. What do you think?

muvaf: That would work for selector resolver and default selector but not much for enforced selector, where it overrides whatever is set as reference on the composite resource no matter what.

So, we're not only delegating the choice of reference but also the responsibility of setting the reference to the `CompositionSelector` so that individual implementations of `CompositionSelector`s could incorporate their choice about override and possibly other operations on the CR.

muvaf:
I've added enforced composition as well since it's pretty similar to default one. All manually tested, ready for review.

pkg/controller/apiextensions/composite/api.go
hasheddan: Does this enforced selector exist already or are you just saying if it did then this model would not work? 

In my proposed model:
1. If you pass the `EnforcedSelector` first and a selector exists then it will always return a reference , and no other chained `CompositionSelector`s will be called.
2. If you pass the `EnforcedSelector` first and a selector does not exist then it will not return a reference, and other chained `CompositionSelector`s will be called.

In the current model:
1. If you pass the `EnforcedSelector` first and a selector exists then it will set the reference, and all other chained `CompositionSelector`s will be called.
2. If you pass the `EnforcedSelector` first and a selector does not exist then it will not modify the reference, and all other chained `CompositionSelector`s will be called.

I get what you mean when the `EnforcedSelector` is not passed first because then it will override any of the other `CompositionSelector`s, but that does not feel like a chain to me, it sounds more life modifiers. Generally with a chain I think of an ordered set of operations that are applied sequentially until one is satisfied. For instance, in some sort of authentication chain it will usually not check for a config file, find a valid one, then check for an environment variable after that and decide to toss out the config file if the environment variable exists (for example). In practice, the present `SelectorResolver`s will not run into this issue because of the way they check for a reference and selectors, but future override selectors that a user might define (such as something like `EnforceDefault`, which would override any references and selectors with the default) could cause surprising behavior when used with existing `CompositionSelector`s.

All of that being said, this is a UX thing that is not changing the actual functionality at this time, so if you feel strongly I would be willing to defer. It seems like the current model could have more complex behavior, but also be more powerful, while my proposal would be simpler to reason about but would limit the ability to use `CompositionSelector`s together in the same way.

hasheddan: @muvaf note that I wrote this comment above before the update you just pushed, which I see includes `EnforcedCompositionSelector` :) I think my points still stand though if you want to weigh in further :+1:

muvaf: I think the difference between two models is about keeping some common logic in the `CompositionSelectorChain` object or not. I think we shouldn't return a reference in `Reconcile` object and set it, so, in order to accommodate that suggestion we'd need to end up with 2 interfaces; one for `CompositionSelectorChain` to be called by `Reconcile` and one for each of the selectors that are passed in.

As you said it's a UX thing and I think one interface and having the chain caller dumb seems simpler to me. Also, I pushed the enforced selector, too. You can see each selector has all of its logic in the function in an idempotent way so that the order of the selectors doesn't actually matter.

hasheddan: > You can see each selector has all of its logic in the function in an idempotent way so that the order of the selectors doesn't actually matter.

Yep, I definitely appreciate that about these defined `CompositionSelector`s, my worry is the cognitive burden for someone adding a new `CompositionSelector` that may not be idempotent. However, maybe it is an okay expectation for us to have of users that they would know to write them in an idempotent way if they don't want consumers to worry about order. As I mentioned, I don't think there is a right way or a wrong way necessarily, just my personal preference. You have done more work on these controllers, so I will defer to your judgement :)

apis/apiextensions/v1alpha1/infrastructuredefinition_types.go
```diff
@@ -17,6 +17,7 @@ limitations under the License.
 package v1alpha1
 
 import (
+	v1 "k8s.io/api/core/v1"
```
negz, line 20: I think we mostly import this as `corev1` elsewhere. Can we do that here (and elsewhere in this PR) for consistency?

pkg/controller/apiextensions/composite/api.go
```diff
@@ -125,6 +151,93 @@ func (r *APISelectorResolver) ResolveSelector(ctx context.Context, cp resource.C
...
+	if cp.GetCompositionReference() != nil || cp.GetCompositionSelector() != nil {
+		return nil
+	}
+	def := &v1alpha1.InfrastructureDefinition{}
+	if err := r.client.Get(ctx, meta.NamespacedNameOf(&r.defRef), def); err != nil {
+		return errors.Wrap(err, errGetInfraDef)
+	}
+	if def.Spec.DefaultCompositionRef == nil {
+		return nil
+	}
```
negz, line 179: Could we pass the default and enforced compositions down from the InfraDef controller to its child composite controllers, rather than having the child controllers query their parent InfraDef on every reconcile? Presumably this would require us to handle the case in which the default/enforced composition changed - i.e. by restarting the composition controllers.

This leads me to wonder whether we should support the enforced composition changing at all. Would changing the enforced composition [retcon](https://en.wikipedia.org/wiki/Retroactive_continuity) any existing composite resources to use the newly enforced composition?

pkg/controller/apiextensions/composite/api.go
```diff
@@ -125,6 +151,93 @@ func (r *APISelectorResolver) ResolveSelector(ctx context.Context, cp resource.C
...
+
+	// TODO(muvaf): A validating webhook on InfrastructureDefinition that will make
+	// sure referenced Composition is compatible could make more sense.
```
negz, line 182: I agree a webhook would be the better approach. Do we need to check this in this controller? What happens today if a composite explicitly references a composition that is not compatible with it?

pkg/controller/apiextensions/composite/api_test.go
```diff
@@ -300,12 +301,370 @@ func TestResolveSelector(t *testing.T) {
...
+	}
+}
+
+func TestAPIDefaultCompositionSelector(t *testing.T) {
+	a, k := schema.EmptyObjectKind.GroupVersionKind().ToAPIVersionAndKind()
+	tref := v1alpha1.TypeReference{APIVersion: a, Kind: k}
+	comp := &v1alpha1.Composition{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "foo",
+			Namespace: "bar",
```
negz, line 321: Compositions don't have a namespace, right?

pkg/controller/apiextensions/composite/api.go
muvaf: That was my initial approach, too; pass the `InfrastructureDefinition` object to both composition selectors instead of a reference to it. I see no problem making enforced composition ref `+immutable` but I believe we should support changing default composition ref. Not so sure how restarting the controller at every change would look like.

Right now, it's only the enforced selector that queries the infradef at every reconcile (default selector doesn't query anything after the first pass) because it cannot know whether the reference on the composite resource drifted or not without knowing what the enforced value is. The safe but less-performant approach is the current one. But I believe what we can do is to mark the enforced composition ref as `+immutable` and make `NewAPIEnforcedCompositionSelector` accept the enforced composition reference directly, relying on the fact that it never changes. It'd be a weak guarantee for now but when we get the `+immutable` webhook implemented, it'd be a strong guarantee.

pkg/controller/apiextensions/composite/api.go
muvaf: > Do we need to check this in this controller?

Actually, InfraDef controller makes more sense since the change happens there, not on composite resource.

> What happens today if a composite explicitly references a composition that is not compatible with it?

Nothing, really. The compatibility is merely a documentation plus the guarantee in the selection, which got me thinking maybe it should be a lower level guarantee, i.e. check the reference in main `Reconcile` for compatibility. Though we'd still need the same check in label selector because it fetches a list of `Composition` and then do the filtering.

I think what I'll do is to make sure infradef controller returns error if an incompatible default or enforced composition is given, don't check compatibility in these two composition selectors, check the compatibility of the actual reference in `Reconcile` of composite resource. That'd fix the bug where the given direct reference is incompatible.

muvaf: I ended up implementing the check only for the composition reference. While testing, I realized that InfraDef doing a check is not really meaningful because it doesn't block the usage of that reference on composite level. In fact, I think it shouldn't even block CRD rendering, for example, even though default composition is not compatible, that default comp ref may never be used. So, it just seemed like not very meaningful check. If it's wrong you'll see error on composite which is the instance where this error should actually block. It could still be nice to have it but I'm not sure if additional logic for fetching composition at every reconcile would be worth the benefit.

I also added events when enforced or default composition has been selected so that people don't go crazy about where that composition reference came from.

Also, I removed the TODO there, because if definition and referenced composition is created at the same time with `kubectl apply`, then validating webhook might reject it since the referenced composition is yet to be created.

muvaf: Making it `+immutable` comes with a caveat actually. If I create infradef with an enforced composition ref and then I might want to remove that enforcement. I can't do that right now because the controller would have to restarted; even if it checks InfraDef at every reconcile, `+immutable` would lead the same behavior. Are we OK with `once enforced, you need to delete infradef to not enforce it` approach? When I manually tested, I did `kubectl apply` with infradef that doesn't have enforcement anymore and hit this. But I'm not sure if that'd be a common thing.

muvaf:
Tested manually with the latest changes.

pkg/controller/apiextensions/composite/api.go
negz: > Are we OK with once enforced, you need to delete infradef to not enforce it approach?

I feel okay with this limitation.

apis/apiextensions/v1alpha1/infrastructuredefinition_types.go
```diff
@@ -33,6 +34,17 @@ type InfrastructureDefinitionSpec struct {
...
+	// DefaultCompositionRef refers to the Composition resource that will be used
+	// in case no composition selector is given.
+	// +optional
+	DefaultCompositionRef *corev1.ObjectReference `json:"defaultCompositionRef,omitempty"`
+
+	// EnforcedCompositionRef refers to the Composition resource that will be used
+	// by all composite instances whose schema is defined by this definition.
+	// +optional
+	// +immutable
+	EnforcedCompositionRef *corev1.ObjectReference `json:"enforcedCompositionRef,omitempty"`
```
negz, line 46: Can we avoid proliferating the use of ObjectReference here? See the rationale in https://github.com/crossplane/crossplane-runtime/issues/49.

pkg/controller/apiextensions/composite/api.go
```diff
@@ -125,6 +155,68 @@ func (r *APISelectorResolver) ResolveSelector(ctx context.Context, cp resource.C
...
+	}
+	def := &v1alpha1.InfrastructureDefinition{}
+	if err := s.client.Get(ctx, meta.NamespacedNameOf(&s.defRef), def); err != nil {
+		return errors.Wrap(err, errGetInfraDef)
+	}
+	if def.Spec.DefaultCompositionRef == nil {
+		return nil
+	}
+	cp.SetCompositionReference(def.Spec.DefaultCompositionRef)
+	s.recorder.Event(cp, event.Normal(reasonCompositionSelection, "Default composition has been selected"))
```
negz, line 186: Is this information really worth plumbing a recorder down here for? Would we still be able to record "A composition has been selected" without specifically communicating that it was a default or enforced composition if we did this further up in the stack?

pkg/controller/apiextensions/composite/api.go
```diff
@@ -125,6 +155,68 @@ func (r *APISelectorResolver) ResolveSelector(ctx context.Context, cp resource.C
...
+	// because enforced composition ref is immutable as opposed to default
+	// composition ref.
+	if s.def.Spec.EnforcedCompositionRef == nil {
+		return nil
+	}
+	// If the composition is already chosen, we don't need to check for compatibility
+	// as its target type reference is immutable.
+	if cp.GetCompositionReference().String() == s.def.Spec.EnforcedCompositionRef.String() {
+		return nil
+	}
```
negz, line 214: Do we need this branch? It seems like if we removed it we'd have a no-op `SetCompositionReference` below, so the only difference is we might emit an event despite the enforced composition already being set.

apis/apiextensions/v1alpha1/infrastructuredefinition_types.go
muvaf: I've been using it just because most meta methods that work with object references support it, including ours. Though I agree with the rationale behind it.

muvaf: I actually wanted to refactor the tooling in crossplane-runtime that uses `ObjectReference`, like set/get composition reference functions. However, the types we have don't exactly meet this use case.

`v1alpha1.TypedReference` could make sense but `apiVersion` and `kind` fields are required in that one, which doesn't for us right now since we have only `Composition` kind and it's verbose to enforce people to write its `apiVersion` and `kind` right now.

`v1alpha1.Reference` has only `Name`, which works for now but I know that we'll want different `kind`s of `Composition`s in the future and that will require churn on Go code level (not on API level really, since `name` field will stay).

I will go with `v1alpha1.Reference` for now but save that refactoring for later.

pkg/controller/apiextensions/composite/api.go
muvaf: I think what composition is selected is already clear from its field on `spec` but I find it valuable to see how it's selected to avoid situations where it's relatively hard to figure out what's happening. This could happen in this situation because the person who is creating a requirement or composite resource is kind of far from seeing what default/enforced composition is easily, it's like you have to check the CRD to see it which we never really do. It's also good to educate people who are exploring how stuff works at the place where they do that exploration directly. So, I found it helpful to say how it made it here, if it's exceptional like default/enforced, and what is selected is already shown in spec when you do `kubectl describe`

pkg/controller/apiextensions/composite/api.go
muvaf: Yeah, I believe the event should be fired only when it's actually fired to keep the chronological order of events digestable. I'd like to avoid a situation where there are events on the object but one of every two event is `enforced composition is selected` whereas it's actually selected only if there is a drift.

muvaf:
@negz Thanks for the review! It's ready for another look.

PR #1594, opened 2020-06-05
Add kustomization.yaml for installing all CRDs

hasheddan:

This adds a Kustomization configuration that will allow all CRDs to be installed using kubectl apply -k.

Tested with https://github.com/crossplane/provider-gcp/pull/238

### Checklist

I have:
  appropriate.

[definition of done].

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,7 +83,7 @@ fallthrough: submodules
 manifests:
 	@$(WARN) Deprecated. Please run make generate instead.
 
-generate: $(KUSTOMIZE) go.vendor go.generate manifests.prepare manifests.annotate
+generate: $(KUSTOMIZE) go.vendor go.generate manifests.prepare manifests.annotate local-kustomize
 	@$(OK) Finished vendoring and generating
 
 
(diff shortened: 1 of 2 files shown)


negz:
I typically apply the local CRDs by running `kubectl apply -f path/to/crd/dir`. Does this Kustomize variant improve on that process/experience?

hasheddan:
@negz I think so because you can do it remotely with as shown in the GCP PR linked in body

Makefile
```diff
@@ -147,6 +147,16 @@ manifests.annotate:
 	sed '1,7d' $(SOURCE_DOCS_DIR)/reference/install.md > $(CROSSPLANE_CHART_DIR)/README.md
 	@$(OK) Copied and modified chart README.md from Crossplane docs
 
+local-kustomize:
+	@$(INFO) Adding all CRDs to Kustomize file for local development
+	@rm cluster/kustomization.yaml
```
negz, line 152: Should this be gitignored if we're deleting and regenerating when we run `make local-kustomize`?

hasheddan: @negz do you mean the `cluster/kustomization.yaml`? `local-kustomize` is called as part of `make generate` so if there are differences then we will fail `check-diff`. We cannot `gitignore` this file because it is what we are adding to enable `kubectl apply -k` here.

hasheddan: @negz could you circle back for another review here?

kasey:
I understand that the reason you're adding this is to allow all CRDs to be installed without cloning the repo locally. This LGTM, but suggestions for possible improvement are:
1) giving the `local-kustomize` make target a name that clarifies it's intended use is to generate, not apply, the kustomize file. Maybe something like `generate-crd-kustomization`?
2) Add a comment to the generated file indicating the intention, like "This kustomization can be used to install Crossplane CRDs from the crossplane github repository by running eg `kubectl apply -k https://raw.githubusercontent.com/crossplane/crossplane/master/cluster/kustomization.yaml`"

hasheddan:
@kasey those are great suggestions! @negz please hold off on merging until I implement :+1:

hasheddan:
@negz okay this is ready now :)

hasheddan:
@negz @kasey I have implemented changes suggested here and am going to go ahead and merge given that you have both approved :heavy_check_mark:

Issue #1586, opened 2020-06-03
Migrate sample stacks to composition


muvaf:
### What problem are you facing?

They are still using claims and classes.

### How could Crossplane help solve your problem?

We can migrate them to composition.

muvaf:
Depends on https://github.com/crossplane/crossplane/issues/1587 so that all sample stacks and Wordpress app target the same definitions.

muvaf:
Depends on https://github.com/crossplane/crossplane/issues/1605 so that Wordpress can still schedule to the target cluster.

muvaf:
Here is the list of PRs and issues that are required for Wordpress to work on composition e2e:
* KAR needs to work with just a Secret  - PR https://github.com/crossplane/crossplane/pull/1604
* There is no way to include InfraDef  in a package, which is required if we'd like to have stack define its own API for network resources; the alternative is to dump network resources just like today and only change the classes with Composition s (which is what the earlier commit in stack-gcp-sample does) - Issue https://github.com/crossplane/crossplane/issues/1610
* Composite resource deletion should be blocked until all composed resources are gone - Issue https://github.com/crossplane/crossplane/issues/1612
* Making Wordpress work with new KAR field and use common types: PRs https://github.com/crossplane/app-wordpress/pull/68 and https://github.com/crossplane/app-wordpress/pull/69
* Using composition in stack-gcp-sample (once this works e2e, others will be opened): PR https://github.com/crossplane/stack-gcp-sample/pull/55

There are a few smaller issues as well like https://github.com/crossplane/crossplane/issues/1611 but the ones that are listed above are blockers for the open PRs to be merged. I think we should be able to include them in 0.13 but we haven't yet done the planning for 0.13 yet.

muvaf:
We've synced and here is the current status on that migration.

The biggest hurdle to do this migration is to be able to bundle an `InfrastructureDefinition` in a package that will deploy it [only once](https://github.com/crossplane/crossplane/issues/1610), as opposed per CR which is how template stacks work right now. Essentially, we need a treatment for `InfrastructureDefinition` as if it's a CRD that the package deploys and this will be addressed in [package manager refactor design](https://github.com/crossplane/crossplane/pull/1616).

[...]

PR #1604, opened 2020-06-22
KubernetesApplicationResource to schedule to a cluster with only secret

### LINKED ISSUE #1605: Kubernetes Application cannot schedule to the types defined via InfrastructureDefinition

Kubernetes Application cannot schedule to the types defined via InfrastructureDefinition. This is required for [Wordpress](https://github.com/crossplane/app-wordpress) app to work with composition.

Allow `KubernetesApplication` or `KubernetesApplicationResource` to accept a secret to connect to the target cluster.

muvaf:

Claims are being deprecated and `KubernetesTarget` is tightly coupled to it. With composition, Crossplane will not know whether the type that is defined by `InfrastructureDefinition` has/is a Kubernetes cluster, hence, it could not schedule. The common denominator of all types that include a Kubernetes cluster is the credentials secret and this PR makes it possible to schedule to a target cluster by supplying a secret with kubeconfig.

With this PR, the actual connection is made with the `Secret` referenced in `spec.credentialsRef`. If only a `target` is given, then the controller extracts the secret reference from the target and writes it to `spec.credentialsRef` of KAR so that the change is backward-compatible.

Fixes #1605 

I have:

Manually with https://github.com/crossplane/app-wordpress/pull/68


muvaf:
Here is the latest update for the scenario that this PR wants to enable: https://github.com/crossplane/crossplane/issues/1586#issuecomment-648696616

@negz Making `KubernetesApplicationResource` work with bare `Secret`s will be beneficial for the point where we have new packages but not new app deployment model, so, I still think we should merge this one and make WP work with it using claim's secret.

apis/workload/v1alpha1/types.go
```diff
@@ -170,12 +170,26 @@ type KubernetesApplicationResourceSpec struct {
 	// +optional
 	Target *KubernetesTargetReference `json:"targetRef,omitempty"`
 
+	// TargetCredentialsRef references the secret that has the credentials to
```
hasheddan, line 173:
```suggestion
	// CredentialsRef references the secret that has the credentials to
```

muvaf:
> Do you just mean for when a k8s cluster is provisioned using composition so it doesn't get bound and so no KubernetesTarget is automatically created for it?

Yeah, sort of. There will be a point where we still have to use `KubernetesApplication` but be blocked because of this. The agent stuff doesn't even have a design doc yet.

muvaf:
What do you guys think about merging this? @negz @hasheddan

muvaf:
@hasheddan The [issue](https://github.com/crossplane/crossplane/issues/1605) this PR fixes is closed. I will close this PR now and if we need it, we can re-open it.

Issue #967, opened 2019-10-23
Proposal: Weighted random composition selection


negz:
### What problem are you facing?

The [simple resource class selection](https://github.com/crossplaneio/crossplane/blob/31ba7418/design/one-pager-simple-class-selection.md) design introduces the ability for resource claims to select the resource classes they use by labels. When multiple resource classes match a claim's label selector a random matching resource class is selected.

Weighting the resource classes would allow us to more intelligently select a resource class when multiple match the class selector. With weighted random resource class selection when multiple resource classes matched a resource claim's label selector the class with the heaviest weight would win. If multiple classes matching the label selector shared the highest weight they would be deemed equally desirable and one would be chosen at random.

A weighted resource class would look something like:

```yaml
apiVersion: cache.gcp.crossplane.io/v1alpha2
kind: CloudMemorystoreInstanceClass
metadata:
  name: gcp-dev
  labels:
    engineVersion: "3.2"
    region: us-west
    replicated: "true"
weight: 100  # Represents the weight of the class.
specTemplate:
  writeConnectionSecretsToNamespace: crossplane-system
  tier: STANDARD_HA
  region: us-west2
  memorySizeGb: 1
  redisVersion: REDIS_3_2
```

Weighted random resource class selection would provide the foundation for cost based scheduling - i.e. the cheapest resource classes get the heaviest weight. A standalone admission webhook or controller could watch for newly created or updated resource classes, price them, and allocate them a weight based on their price.

### How could Crossplane help solve your problem?

The challenge with implementing weighted random resource class selection is that as described in the [technical implementation](https://github.com/crossplaneio/crossplane/blob/31ba7418/design/one-pager-simple-class-selection.md#technical-implementation) there is not a single controller handling each newly found resource claim, but rather several racing controllers. Each controller assesses the available resource classes of a particular kind, then uses random jitter to ensure a random controller wins if multiple controllers find matching resource classes. These controllers would need to coordinate to support weighted random scheduling.

Claim scheduling controllers could coordinate via a new `ClassWeight` kind:

```yaml
apiVersion: scheduling.crossplane.io/v1alpha1
kind: ClassWeight
metadata:
  # The UID of the associated class. ClassWeights are cluster scoped, like classes.
  name: eb86ac7a-f5cb-11e9-b82c-5aa538984bd8
  labels:
    # This label is used to filter ClassWeights to those pertaining to classes that
    # could satisfy a particular kind of resource claim.
[...]

PR #1618, opened 2020-06-25
Select a random Composition when several match

negz:

A composite resource may select a Composition using labels. This can be used to describe the desired Composition in general terms; i.e. 'a Composition of prod resources' or 'a Composition of east coast resources'. It's possible that many Compositions will match these labels. If this happens, we should select one of the matching Compositions at random. This allows infrastructure operators to ensure Composite resources will be spread evenly across multiple Compositions (which may mean multiple regions, or cloud providers, etc).

I'd like to make this weighted random selection at some point in the future, per https://github.com/crossplane/crossplane/issues/967#issuecomment-637835664

I have:

I ran the existing unit tests and confirmed full code coverage. I don't think there's much value in testing or mocking out the randomness, and I think this change is tiny enough that I feel confident without manual testing.

### DIFF
--- a/pkg/controller/apiextensions/composite/api.go
+++ b/pkg/controller/apiextensions/composite/api.go
@@ -150,7 +150,7 @@ func (r *APILabelSelectorResolver) SelectComposition(ctx context.Context, cp res
 	v, k := cp.GetObjectKind().GroupVersionKind().ToAPIVersionAndKind()
 
 	for _, comp := range list.Items {
-		if comp.Spec.From.APIVersion == v || comp.Spec.From.Kind == k {
+		if comp.Spec.From.APIVersion == v && comp.Spec.From.Kind == k {
 			// This composition is compatible with our composite resource.
 			candidates = append(candidates, comp.Name)
 		}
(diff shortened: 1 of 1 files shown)


pkg/controller/apiextensions/composite/api.go
```diff
@@ -143,17 +145,25 @@ func (r *APILabelSelectorResolver) SelectComposition(ctx context.Context, cp res
...
 	}
-	apiVersion, kind := cp.GetObjectKind().GroupVersionKind().ToAPIVersionAndKind()
+
+	candidates := make([]string, 0, len(list.Items))
+	v, k := cp.GetObjectKind().GroupVersionKind().ToAPIVersionAndKind()
+
 	for _, comp := range list.Items {
-		if comp.Spec.From.APIVersion != apiVersion || comp.Spec.From.Kind != kind {
-			continue
+		if comp.Spec.From.APIVersion == v || comp.Spec.From.Kind == k {
```
muvaf, line 153:
```suggestion
		if comp.Spec.From.APIVersion == v && comp.Spec.From.Kind == k {
```

negz: Good catch, thanks!

PR #1621, opened 2020-06-29
add make install command to install crds to cluster

wonderflow:
Signed-off-by: 天元 <jianbo.sjb@alibaba-inc.com>

Add a `make install` command to install install crds to cluster, this is for convenience like `make run`. Mainly for debug and develop.

Fixes #

I have:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -190,6 +190,14 @@ submodules:
 	@git submodule sync
 	@git submodule update --init --recursive
 
+# Install CRDs into a cluster. This is for convenience.
+install: reviewable
+	kubectl apply -f cluster/charts/crossplane-types/crds/
+
+# Uninstall CRDs from a cluster. This is for convenience.
+uninstall:
+	kubectl delete -f cluster/charts/crossplane-types/crds/
+
 # This is for running out-of-cluster locally, and is for convenience. Running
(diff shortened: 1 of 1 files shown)


negz:
I like this target @wonderflow, but I wonder if if should have a different name? `make install` implies to me that it would install all of Crossplane, not just the CRDs. Is the idea here that you could do something like `make install run` to avoid needing to use Helm to develop against an API server?

wonderflow:
> Is the idea here that you could do something like make install run to avoid needing to use Helm to develop against an API server?

Yes, what about `make installcrd` ?

Makefile
```diff
@@ -190,6 +190,14 @@ submodules:
 	@git submodule sync
 	@git submodule update --init --recursive
 
+# Install CRDs into a cluster. This is for convenience.
+install: reviewable
+	kubectl apply -f cluster/charts/crossplane-types/crds/
```
hasheddan, line 195:
```suggestion
install: $(KUBECTL) reviewable
	$(KUBECTL) apply -f cluster/charts/crossplane-types/crds/
```

I believe this change would allow folks to run the command even if they hadn't already installed `kubectl`

wonderflow:
@hasheddan  Thanks, followed your advice

PR #1633, opened 2020-06-30
Crossplane Agent design doc

### LINKED ISSUE #1617: Design for Crossplane Agent

There are 2 main modes of deployment of Crossplane that folks are mostly using:
* Local scenario.
  * In this model, you have a single beefy cluster with infrastructure CRs and you create Kubernetes primitive resources to deploy your apps (`Deployment`, `Pod` etc) that will consume that infrastructure.
* Remote scenario.
  * You have a single dedicated Crossplane cluster that you use to provision & manage all of your infrastructure from a single point, including the Kubernetes cluster that your app will run. This is challenging than the first scenario because secrets & your app will need to be delivered to another cluster and for that we have workload API, i.e. `KubernetesApplication` and `KubernetesApplicationResource`s.

Here we're focusing on the second scenario. For reference, here is how the workload API looks like:
<details>
<summary>KubernetesApplication</summary>

```yaml
apiVersion: workload.crossplane.io/v1alpha1
kind: KubernetesApplication
metadata:
  name: wp-app
  labels:
    app: wp
spec:
  resourceSelector:
    matchLabels:
      app: wp
  targetSelector:
    matchLabels:
      app: wp
  resourceTemplates:
    - metadata:
        name: wp-namespace
        labels:
          app: wp
      spec:
        template:
          apiVersion: v1
          kind: Namespace
          metadata:
            name: wp
            labels:
              app: wordpress
    - metadata:
        name: wp-deployment
        labels:
          app: wp
      spec:
        secrets:
          # This must match the writeConnectionSecretToRef field
          # on the database claim; it is the name of the secret to
          # pull from the crossplane cluster, from this Application's namespace.
          - name: sql
        template:
          apiVersion: apps/v1
          kind: Deployment
          metadata:
            namespace: wp
            name: wordpress
            labels:
              app: wordpress
          spec:
            selector:
              matchLabels:
                app: wordpress
            template:
              metadata:
                labels:
                  app: wordpress
              spec:
                containers:
                  - name: wordpress
                    image: "wordpress:4.6.1-apache"
                    env:
                      - name: WORDPRESS_DB_HOST
                        valueFrom:
                          secretKeyRef:
                            # This is the name of the secret to use to consume the secret
                            # within the managed cluster. The reason it's different from the
                            # name of the secret above is because within the managed cluster,
                            # a crossplane-managed secret is written as '{metadata.name}-{secretname}'.
                            # The metadata name is specified above for this resource, and so is
                            # the secret name.
                            name: wp-deployment-sql
                            key: endpoint
                      - name: WORDPRESS_DB_USER
                        valueFrom:
                          secretKeyRef:
                            name: wp-deployment-sql
                            key: username
                      - name: WORDPRESS_DB_PASSWORD
                        valueFrom:
                          secretKeyRef:
                            name: wp-deployment-sql
                            key: password
                    ports:
                      - containerPort: 80
                        name: wordpress
    - metadata:
        name: wp-service
        labels:
          app: wp
      spec:
        template:
          apiVersion: v1
          kind: Service
          metadata:
            namespace: wp
            name: wordpress
            labels:
              app: wordpress
          spec:
            ports:
              - port: 80
            selector:
              app: wordpress
            type: LoadBalancer
```

</details>

Alternatively:

<details>
<summary>3 KubernetesApplicationResources</summary>

```yaml
---
apiVersion: workload.crossplane.io/v1alpha1
kind: KubernetesApplicationResource
metadata:
  name: wp-namespace
spec:
  targetSelector:
    matchLabels:
      app: wp
  template:
    apiVersion: v1
    kind: Namespace
    metadata:
      name: wp
      labels:
        app: wordpress
---
apiVersion: workload.crossplane.io/v1alpha1
kind: KubernetesApplicationResource
metadata:
  name: wp-deployment
spec:
  targetSelector:
    matchLabels:
      app: wp
  secrets:
    - name: wp-sql
  template:
    apiVersion: apps/v1
    kind: Deployment
    metadata:
      namespace: wp
      name: wordpress
      labels:
        app: wordpress
    spec:
      selector:
        matchLabels:
          app: wordpress
      template:
        metadata:
          labels:
            app: wordpress
        spec:
          containers:
            - name: wordpress
              image: "wordpress:4.6.1-apache"
              env:
                - name: WORDPRESS_DB_HOST
                  valueFrom:
                    secretKeyRef:
                      # This is the name of the secret to use to consume the secret
                      # within the managed cluster. The reason it's different from the
                      # name of the secret above is because within the managed cluster,
                      # a crossplane-managed secret is written as '{metadata.name}-{secretname}'.
                      # The metadata name is specified above for this resource, and so is
                      # the secret name.
                      name: wp-deployment-wp-sql
                      key: endpoint
                - name: WORDPRESS_DB_USER
                  valueFrom:
                    secretKeyRef:
                      name: wp-deployment-wp-sql
                      key: username
                - name: WORDPRESS_DB_PASSWORD
                  valueFrom:
                    secretKeyRef:
                      name: wp-deployment-wp-sql
                      key: password
              ports:
                - containerPort: 80
                  name: wordpress
---
apiVersion: workload.crossplane.io/v1alpha1
kind: KubernetesApplicationResource
metadata:
  name: wp-service
spec:
  targetSelector:
    matchLabels:
      app: wp
  template:
    apiVersion: v1
    kind: Service
    metadata:
      namespace: wp
      name: wordpress
      labels:
        app: wordpress
    spec:
      ports:
        - port: 80
      selector:
        app: wordpress
      type: LoadBalancer
```

</details>

There are a few issues with the workload propagation mechanics we're providing to the users.
* You cannot interact with what you deploy directly, i.e. always have to use `KubernetesApplicationResource` as a proxy and that has its own set of challenges:
  * It's a template and gets deployed after you make the edit, so, you loose the admission check rejections in case something went wrong. Instead, you'll see them in the status but you won't be prevented from making the change as opposed to directly interacting.
  * Late initialization.
    * Let's say you deployed a `Pod` and `spec.node` is late-initialized. You will not see that because we only propagate the status back, not spec because the template is not strong-typed and it's hard to differentiate between user's actual desired spec and what's only a late-inited value.
    * If you have an element in an array that is late-inited or some elements are added after the creation, `PATCH` command will replace the whole array with what you got in your template. If the type is well-constructed to provide its own merge mechanics, this could be avoidable but that is usually not the case. For example, in some cases an element of an array in spec is late-inited for bookkeeping the IP and removing this causes its controller to provision new ones each time.
* Migration from a Helm chart that has a few `Deployment`s and a `StatefulSet` for DB is harder.
  * You actually need to change each and every element to be in a `KubernetesApplicationResource` as well as `StatefulSet` of DB with a DB requirement CR.
* Operation experience. This is related to the first point. If you have an app using OAM or some other app model, then you always have that intermediate proxy of workload CRs. You're losing on some of the value that these models provide because of that proxy and in some cases it could be functionally detrimental.

Surely, it has its own advantages as well. For example, you can manage all of your apps from single point via `KubernetesApplication`s targeting the right clusters. But as we see more usage patterns, we're more convinced that the thing users want to manage from a single central point is infrastructure (including policy, audit etc) but not necessarily applications.

In the current model, the general approach is that main cluster pushes some applications together with some secrets that are populated by infrastructure CRs. What we'd like to achieve is to keep the experience of single control plane that you can view, manage, audit your infrastructure and define your own APIs to be used by different people as well as to improve the mechanics of how apps can consume that infrastructure from different clusters in a way that they don't have to make big changes and/or sacrifices from their original UX experience.

In its essence, the proposed model is that there will be an agent that you deploy into _any_ cluster, configure it to talk to a cluster that has Crossplane installed and make the APIs and infrastructure that exists in this single Crossplane cluster available for consumption in the remote cluster so that people directly create `Deployment`, `MySQLInstanceRequirement` and mount latter's secret to the former. It could sound that Crossplane gets a bit harder to operate with a new component and that is true for setup phase. However, the trade-off is that it gets a little bit harder to setup but once you do it as platform-owner, many of your users won't need to think about workload APIs anymore; they will just look at what infrastructure is defined and made available to them and use them in their YAMLs without having to migrate all to be `KubernetesApplicationResource`. So, the aim is to make end-users/app developers/app operators more happy while keeping the power of central control plane that lets you see & manage all the infrastructure used in the organization.

muvaf:

Crossplane agent design document to address remote mode.

Fixes #1617 

[skip ci]

I have:

In my dreams.


design/design-doc-agent.md
```diff
@@ -0,0 +1,586 @@
...
+> Though the override behavior also paves the way to target a completely
+> different central cluster, too, which may break the setup since CRDs in that
+> other central cluster could be different. We accept this risk as it would have
+> to be a deliberate choice of the user.
+
+#### Authorization
+
+##### Remote Cluster
+
+Since there will be one agent for the whole cluster, its own mounted
```
lukeweber, line 298: It feels like agent could connect safely to one cluster without becoming a mess, but could have n scoped connections for different namespace mappings, or could have one privileged connection with n mappings.

design/design-doc-agent.md
```diff
@@ -0,0 +1,586 @@
...
+  * Because you have to run applications in the same central cluster with
+    others, you'll have to share the same Kubernetes resources like nodes and
+    your needs in terms of instance types could differ greatly depending on your
+    workloads, like some need GPU-powered machines and others memory-heavy ones.
+* Security.
+    * All applications are subject to the same user management domain, i.e. same
+      api-server. This could be managed to be safe, but it's not physically
+      impossible to have a `ServiceAccount` in another namespace to have access
+      to resources in your namespace. So, you wouldn't really trust to have
+      production in one namespace and dev in the other.
```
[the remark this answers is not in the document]
muvaf: Actually I was meaning to refer to the native Kubernetes resources here instead of Crossplane ones because the comparison is about local and remote scenarios, and that comparison doesn't really change separation of namespace-scoped and cluster-scoped objects being more secure or not. It's more about multi-namespace usage of Kubernetes as workload scheduler. Obviously, it works well for a lot of cases but still the possibility of escape does exist, like it's not an isolation level that GCP Project provides or isolation level that you'd have in VMs that run on the same machine. That's why people use different projects for prod workloads for example.

design/design-doc-agent.md
```diff
@@ -0,0 +1,586 @@
...
+mechanisms.
+
+#### Authenticating to The Central Cluster
+
+In order to execute any operation, a `ServiceAccount` needs to exist in the
+central cluster with appropriate permissions to read during pull and write
+during push operations while synchronizing with central cluster. Since the agent
+is running in the remote cluster, the credentials of this `ServiceAccount` will
+be stored in a `Secret` in the remote cluster. Alongside the credentials, the
+agent needs to know the namespace that it should sync to in the central cluster.
```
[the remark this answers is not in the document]
muvaf: You mean two different remote clusters with their own agents configured to target the same namespace in the central cluster, right? That's a good point. I see two options there:
* Annotate requirements in the central cluster with the UID of `kube-system` namespace of the remote cluster that requested it.
* Allow conflict to happen. There could be cases where several clusters use the same cloud services provisioned in the central cluster.
  * Maybe we can have a read-only mode for the agent? It would only pull from the central cluster instead of actively pushing.


I am kind of leaning towards the second option partly because same cloud service (like network or message queue) being used by several remote cluster sounds like a valid use case and partly because preventing this would be making a decision to restrict the user's ability to have different scenarios that we don't anticipate, i.e. the benefit of preventing it may not outweigh its downside, which is accidentally messing up the requirements with different remote clusters.

design/design-doc-agent.md
```diff
@@ -0,0 +1,586 @@
...
+
+The way to have the agent find the right `Secret` to use in order to
+authenticate to the central cluster for synchronization of a given requirement
+will consist of two steps:
+
+* Look for specific annotations on the requirement type to specify what `Secret`
+  needs to be used:
+  * `agent.crossplane.io/credentials-secret-name: sa-secret` for the full
+    kubeconfig to connect to the central cluster.
+  * `agent.crossplane.io/namespace: my-app-1` for the namespace it should sync
```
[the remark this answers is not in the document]
muvaf: > Is this annotation on the *Requirement CRD or is this added on a *Requirement CR to effect it's sync behavior?

It's on the `*Requirement` custom resource, not on its definition.

> I think this annotation is based on explicit agent configuration. i.e. the CR is written to a agent cluster namespace, and the agent resolves the remote namespace for the local, defaulting to the same name if no-override is configured.

Do you mean if I create a requirement in namespace `foo` in my remote cluster, then by default, it should sync to namespace `foo` in the central cluster? I'd be wary of such pairing because the identification space of the central cluster is made up of several remote clusters and their namespaces, so, default wouldn't really be useful. For example, I can see the case where several remote clusters just using `default` namespace initially and they'd conflict by default. So, that exact pairing may start to not work well really quickly.

design/design-doc-agent.md
```diff
@@ -0,0 +1,586 @@
...
+    requirements and their secrets.
+* If no annotation is given, then use the defaults.
+  * Installation procedure of the agent will allow to specify a default secret
+    and namespace during installation.
+
+By this override behavior, it is possible to supply credentials of another
+namespace or `ServiceAccount` for a specific `*Requirement` resource. So, users
```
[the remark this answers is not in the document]
muvaf: Yeah, I think that's an accurate description. In its essence, we always require user to tell which namespace to sync to; either through agent setup or annotation on each requirement. The assumption is that the `ServiceAccount` that is provided should have the necessary permissions in whatever namespace you target to.

design/design-doc-agent.md
muvaf: Sorry, I couldn't get what you mean. Could you elaborate?

design/design-doc-agent.md
```diff
@@ -0,0 +1,586 @@
...
+the following permissions for agent to do its operations:
+
+* Read
+  * `CustomResourceDefinition`s
+  * `InfrastructureDefinition`s
+  * `InfrastructurePublication`s
+  * `Composition`s
+  * `Secret`s in the given namespace.
+* Write
+  * `*Requirement` types that you'd like to allow in given namespace.
```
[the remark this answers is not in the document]
muvaf: > To be clear this seems like n namespaced RoleBindings to a Roles for a given requirement, and might be hard in practice to calculate, so if the agent is adding annotations dynamically.

I think you can have one `Role` in the target namespace of the central cluster that includes all of your `*Requirement` types. In the example below, we add all `kind`s that exist under `database.acmeco.org` and `acmeco.org` groups. I couldn't get the annotation part though; agent doesn't annotate any resource on its own.

> I think something that declares the usage of a requirement might be helpful here to simplify the model.

Could you elaborate that with an example?

muvaf:
> For each event that the agent receives from the requirement types it watches, it'd look for the annotation of the namespace and work with that information to decide how to authenticate to the central cluster and what namespace it should use to sync the resources in that namespace.

Actually, can write that annotations to the requirements whose event it received so that in the consequent reconciles, it wouldn't need to query namespaces again. The annotation on the namespace would be the default value that will be propagated to all requirements in that namespace if they do not already have that annotation.

design/design-doc-agent.md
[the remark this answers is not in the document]
lukeweber: Yeah, I think you call out below that only one agent is running per cluster which my initial question. i.e. it wouldn't make any sense to run one agent in foo, and another in baz namespaces because they both operate on the cluster scope and seems like a headache to manage.

design/design-doc-agent.md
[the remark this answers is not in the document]
lukeweber: If I'm setting this agent up as an infra-owner defining the mapping seems totally reasonable as well. If we allow arbitrary cross mapping via annotations, it seems like it could get very confusing, very quickly. I might argue that we enforce a ns:ns mapping so we enforce a bit of organization on how things are provisioned and avoid maybe opening us up to some odd security issues.

It seems that sa-secret has to be a property of every local namespace and there's no way for infra-owner to set a default at the infra level that works for a set of namespaces which is what I was imagining.

This is what I was imaging not being a good behavior that we would by default not want to happen:
RequirementSync

Do you see as a use case an infra-owner who wants to setup a set of app namespaces to connect to a set of crossplane namespaces without needing separate service accounts for each namespace? I was kind of imaging, as an infra owner I want to provision an app cluster, and then install the agent on it to connect back to crossplane for all namespaces.

design/design-doc-agent.md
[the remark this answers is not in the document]
lukeweber: As an app-operator on an agent cluster it would be nice to clearly know what requirements I can operate on. In my experience with CRDs and RBAC this is not a user-friendly interface per se. I'll leave it as a conceptual idea whether the agent should consider a higher level concept of a list of requirements that are allowed.

design/design-doc-agent.md
```diff
@@ -0,0 +1,586 @@
...
+`MySQLInstanceRequirement` custom resource, a `Pod` and they would share the
+same name for the secret so that one fills that `Secret` with credentials and
+the other one mounts it for the containers to consume.
+
+> For brevity, application will be assumed to have only one `Pod`.
+
+This consumption model works well in cases where you deploy your application in
+a cluster where Crossplane runs, i.e. local mode. However, there are many valid
+use cases where this is not really feasible. Here is a subset of those use
+cases:
```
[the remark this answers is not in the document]
muvaf: Yeah, that's true. I'll update the section.

design/design-doc-agent.md
```diff
@@ -0,0 +1,586 @@
...
+you might want to see all infrastructure to be managed & represented from one
+place and this has various benefits like common published APIs, cost overview,
+tracking lost/forgotten resources etc. But you would want to enable application
+teams to self-serve and have certain level of freedom about on a certain
+infrastructure architecture they'd like to have their applications run inside.
+
+What we need to do is to enable a platform team to have this central
+infrastructure management ability while not imposing hard restrictions on
+application teams. In the end, the goal of the platform teams is to increase the
+velocity of development while keeping everything manageable.
```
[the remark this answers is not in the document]
muvaf: thanks!

design/design-doc-agent.md
```diff
@@ -0,0 +1,586 @@
...
+with this approach:
+
+* You cannot interact with what you deploy directly, i.e. always have to use
+  `KubernetesApplicationResource` as a proxy and that has its own set of
+  challenges:
+  * It's a template and gets deployed after you make the edit, so, you loose the
+    admission check rejections in case something went wrong. Instead, you'll see
+    them in the status, but you won't be prevented from making the change as
+    opposed to directly interacting.
+  * Late initialization.* You cannot interact with what you deploy directly,
```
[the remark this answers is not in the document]
muvaf: It seems there is some duplication as well here, fixed.

design/design-doc-agent.md
```diff
@@ -0,0 +1,586 @@
...
+  using OAM, or some other app model, then you always have that intermediate
+  proxy of workload CRs. You're losing on some value that these models
+  provide because of that proxy and in some cases it could be functionally
+  detrimental.
+
+Surely, it has its own advantages as well. For example, you can manage all of
+your apps from single point via `KubernetesApplication`s targeting the right
+clusters. But as we see more usage patterns, we're more convinced that the thing
+users want to manage from a single central point is infrastructure (including
+policy, audit etc) but not necessarily applications.
```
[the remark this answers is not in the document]
muvaf: Yeah, it does sound a bit too much to claim.

design/design-doc-agent.md
```diff
@@ -0,0 +1,586 @@
...
+      source of truth will be the remote cluster.
+    * `InfrastructureDefinition`, `InfrastructurePublication` and `Composition`.
+  * `Composition`s to discover what's available to choose. CRs of this type will
+    be read-only, and the source of truth will be the central cluster.
+  * `InfrastructurePublication`s to discover what's availabe as published API.
+    Read-only.
+  * `InfrastructureDefinition`s to discover how the secret keys are shaped.
+    Read-only.
+  * `Secret`s that are result of the infrastructure that is provisioned so that
+    it can be mounted to `Pod`s. Read-only.
```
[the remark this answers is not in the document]
muvaf: > I imagine we could pull only the IDs and IPs, and have the agent create the appropriate CRDs in the remote cluster.

I wouldn't really bind agent to the generation logic as it'd imply a manually enforced parity between the generation logic here and in the Crossplane itself. I think just copying the CRDs of the resources listed in the first bullet would be sufficient and it'd guarantee that the source of truth is one and it's Crossplane in the central cluster.

design/design-doc-agent.md
```diff
@@ -0,0 +1,586 @@
...
+* Push
+  * `*Requirement` custom resources so that infrastructure can be requested.
+    Read and write permissions in a specific namespace in the central cluster will
+    be needed.
+
+### RBAC
+
+As we have two different Kubernetes cluster, there will be two separate security
```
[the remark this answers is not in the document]
muvaf: I feel like `API` is kind of overloaded since we're using it here for definition types as well. Changed it to `clusters`.

design/design-doc-agent.md
```diff
@@ -0,0 +1,586 @@
...
+`apiGroup` is less than one would imagine thanks to the ability of allowing a
+whole group; most frequently, the platform operators will be adding new kinds to
+the existing groups.
+
+In the light of this assumption, the initial approach will be that the `Role`
+bound to the agent will be populated by a static list of the groups of the
+requirement types during the installation like shown above and if a new group is
+introduced, then an addition to this `Role` will be needed. A separate
+controller to dynamically manage the `Role` is mentioned in the [Future
+Considerations](#future-considerations) section.
```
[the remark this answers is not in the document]
muvaf: Yeah, we're focusing on the new CRDs requiring new permissions for agent in the app cluster. So, the idea is similar.

design/design-doc-agent.md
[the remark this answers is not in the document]
muvaf: Got it. The UX of CRDs and RBAC could be improved with higher level abstractions but I'd tackle that problem with another one-pager/issue instead where we can consider many options and think carefully about not complicating the system.

design/design-doc-agent.md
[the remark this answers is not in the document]
negz: That makes sense, but I do think we should create a record (i.e. IP) on the remote cluster so that folks can have the same `kubectl get infrastructurepublications` experience on remote clusters as they can on the central cluster.

design/design-doc-agent.md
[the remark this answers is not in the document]
negz: I still have a slight preference for something like "Kubernetes API servers", mostly because "Kubernetes cluster" implies to me a cluster that is handling container orchestration. Just a nitpick though; feel free to ignore.

design/design-doc-agent.md
```diff
@@ -0,0 +1,627 @@
...
+agent needs to know the namespace that it should sync to in the central cluster.
+
+The easiest configuration would be the one where we specify the `Secret` and a
+target namespace via installation commands of the agent. However, the namespace
+pairing list could change dynamically after the installation. So, we need to be
+able to supply the pairings after the installation. In order to specify which
+namespace in the remote cluster should be synced to which namespace in the
+central cluster, `Namespace` resource needs to be annotated.
+
+```yaml
```
ryanzhang-oss, line 274: is this installed on the control plane cluster or the remote cluster?

design/design-doc-agent.md
```diff
@@ -0,0 +1,627 @@
...
+   like the following:
+
+  ```yaml
+  apiVersion: v1
+  kind: Namespace
+  metadata:
+    name: foo
+    annotations:
+      "agent.crossplane.io/target-namespace": bar
+  ```
```
ryanzhang-oss, line 512: Again, is this on the remote or the crossplane cluster? My guess is that it's on the remote but why is the ns name "foo" which is the same as the on the central cluster above.

design/design-doc-agent.md
```diff
@@ -0,0 +1,627 @@
...
+Since there will be one agent for the whole cluster, its own mounted
+`ServiceAccount` in that remote cluster needs to have read & write permissions
+for all of the following kinds in the remote cluster listed below:
+
+* `CustomResourceDefinition`
+* `Composition`
+* `InfrastructureDefinition`
+* `InfrastructurePublication`
+* `Secret`
+* All `*Requirement` types
```
ryanzhang-oss, line 335: I assume that the agent in the remote cluster somehow 'read' all those CRDs from the crossplane cluster and then apply them in the remote cluster? or it's something else?

design/design-doc-agent.md
```diff
@@ -0,0 +1,627 @@
...
+
+The `ServiceAccount` that will be created in the central cluster needs to have
+the following permissions for agent to do its operations:
+
+* Read
+  * `CustomResourceDefinition`s
+  * `InfrastructureDefinition`s
+  * `InfrastructurePublication`s
+  * `Composition`s
+  * `Secret`s in the given namespace.
```
ryanzhang-oss, line 388: Just to clarify, are those CRs or CRDs?

design/design-doc-agent.md
```diff
@@ -0,0 +1,627 @@
...
+* Write
+  * `*Requirement` types that you'd like to allow in given namespace.
```
ryanzhang-oss, line 390: How does the crossplane cluster know those types? Again, are those CRDs or CRs? I assumed that those are from the remote cluster.

design/design-doc-agent.md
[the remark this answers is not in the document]
negz: > something less flexible and more predictable would be ideal

@muvaf Pushed a change to this PR since I left this comment that aligns pretty closely with what I was thinking. The relationship between remote and central resources are now defined by annotations at the namespace level rather than annotations on the CRs themselves. I suspect this is about the right level of flexibility and granularity.

design/design-doc-agent.md
```diff
@@ -0,0 +1,627 @@
...
+
+* Pull
+  * `CustomResourceDefinition`s of all types that we want the applications to be
+    able to manage and view:
+    * Requirements that are published via `InfrastructurePublication`s. The
+      source of truth will be the remote cluster.
+    * `InfrastructureDefinition`, `InfrastructurePublication` and `Composition`.
+  * `Composition`s to discover what's available to choose. CRs of this type will
+    be read-only, and the source of truth will be the central cluster.
+  * `InfrastructurePublication`s to discover what's availabe as published API.
```
negz, line 234:
```suggestion
  * `InfrastructurePublication`s to discover what's available as published API.
```

design/design-doc-agent.md
negz: @muvaf should confirm, but we chatted earlier today and my understanding is that the annotation would be on the namespace on the remote (agent side) cluster.

design/design-doc-agent.md
```diff
@@ -0,0 +1,627 @@
...
+namespace in the remote cluster should be synced to which namespace in the
+central cluster, `Namespace` resource needs to be annotated.
+
+```yaml
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: foo
+  annotations:
+    "agent.crossplane.io/target-namespace": bar
```
negz, line 280: Would this annotation be mandatory, or optional?

Per our out of band conversation I suggest:

* The annotation be optional, and the agent default to trying to sync to the same namespace in the central cluster if the annotation is omitted. So if a requirement was authored in remote namespace `example` the agent would assume that an `example` namespace also existed in the central cluster and would try to create a corresponding requirement there.
* We avoid conflicts by using `metadata.generateName` when we create a corresponding requirement in the central cluster. That is, if the remote requirement is `{namespace: example, name: coolreq}` (and there were no annotations on the namespace) the corresponding central requirement would be `{namespace: example, generateName: coolreq}`.

We'd need to ensure there was some way for the remote requirement to regain control of the corresponding central requirement under this approach; presumably annotations creating a relationship at each end.

design/design-doc-agent.md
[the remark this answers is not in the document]
muvaf: > but I do think we should create a record (i.e. IP) on the remote cluster so that folks can have the same kubectl get infrastructurepublications experience on remote clusters as they can on the central cluster.

Yeah, that's the current proposal. I'll try to make the wording more clear. All CRDs as well as IPs and IDs will be copied, it's just there won't be any controllers watching IP/IDs, their corresponding CRDs will be brought over by the agent.

design/design-doc-agent.md
muvaf: Yes, this is in the remote cluster.

design/design-doc-agent.md
muvaf: > The annotation be optional, and the agent default to trying to sync to the same namespace in the central cluster if the annotation is omitted. So if a requirement was authored in remote namespace example the agent would assume that an example namespace also existed in the central cluster and would try to create a corresponding requirement there.

I have thought having that namespace pairing. I have a feeling that doing exact pairing of the namespaces could be invasive for one remote cluster's effect on that common central cluster. Let's say we have 100 remote clusters and they all have their own namespaces. Besides from conflicts of the general names that we all use when we create a namespace, I think an individual who is creating a namespace in one of those remote cluster may not know what he's signing up to that is claiming yet another full namespace in the central cluster. It just seems very in-directional information that they need to be aware of. So, I thought that if we let the admin of the remote cluster choose one namespace to be the default, then that individual will have less things to think/be careful about. That obviously increases the chance of conflict of the requirements but I'd say that's less of a concern compared to namespace conflicts which seems unrelated to a person setting up their own namespace. Also, changing the name of the requirement is much easier than changing the name of the namespace. In all cases, they can annotate the namespace in the remote cluster to avoid conflicts altogether.

> We avoid conflicts by using metadata.generateName when we create a corresponding requirement in the central cluster. That is, if the remote requirement is {namespace: example, name: coolreq} (and there were no annotations on the namespace) the corresponding central requirement would be {namespace: example, generateName: coolreq}.

I still couldn't convince myself for this; rejecting to sync and writing about it in the status sounds like a simpler approach without the complexity introduced by that name difference that will be caused by `generateName` and its second order effects like making the name even longer (original name + 18 chars will be the length of managed resource name). I'm not sure if we should make agent that smart just because of the conflicts. I mean, you can prevent all kinds of conflicts just by annotating a namespace and you'll explicitly know when a conflict happens so you can take action. So, I'm leaning towards keeping the name pairing between requirements and annotate them in the central cluster with UID of the namespace it exists in the remote cluster to uniquely identify them in the world of custom resources. UID of the `kube-system` could be worth adding, too, to discover in what cluster that namespace exists.

design/design-doc-agent.md
muvaf: The remote cluster is the authority over all fields of the `*Requirement` types except `status`; `status` field will be pulled from Crossplane cluster. The rest of the CRs will be only pulled from the central cluster

design/design-doc-agent.md
muvaf: Both CRDs and their instances

design/design-doc-agent.md
muvaf: They are all defined in the crossplane cluster via InfrastrucutureDefinition and InfrastructurePublications. The agent makes sure both clusters have the same set of CRDs so that the sync is lossless. The actual `*Requirement`s will come from remote cluster.

design/design-doc-agent.md
muvaf: Yeah, it's on the remote. I corrected the names of the namespaces, thanks!

design/design-doc-agent.md
negz: > Let the admin of the remote cluster choose one namespace to be the default

A default that can be overridden at the namespace level seems reasonable. This would promote a pattern in which all namespaces of a remote cluster map to one namespace in the central cluster by default, but enable other patterns with administrator intervention.

It does seem like we could also support the one-to-one mapping by default if we wanted to though - perhaps this could be enabled as part of the agent's config? i.e. Whether to map every remote namespace to one central namespace if there was no annotation, or to map every remote namespace to the central namespace with the same name if there was no annotation. I would suggest that even if we choose not to support the latter case we should structure whatever flags/config file controls this such that we could add support for both if there was user demand.

> I still couldn't convince myself for this

Fair enough - I'm happy to go with identical names that will be rejected if they happen to conflict.

design/design-doc-agent.md
```diff
@@ -0,0 +1,665 @@
...
+* `Secret`
+* All `*Requirement` types
+
+The last one is a bit tricky because the exact list of `*Requirement` types on
+kind level is not known during installation and it's not static; new published
+APIs should be available in the remote cluster dynamically. One option is to
+allow agent to grant `Role` and `RoleBinding`s to itself as it creates the
+necessary `CustomResourceDefinition`s in the remote cluster. However, an entity
+that is able to grant permissions to itself could greatly increase the security
+posture.
```
negz, line 354: Nit: "increase the security posture" reads to me like "makes it more secure", whereas I think this means "makes it less secure". Perhaps "impact the security posture" would better imply the latter?

design/design-doc-agent.md
```diff
@@ -0,0 +1,665 @@
...
+the central cluster and write the credentials to a secret. Then the user can
+copy this secret into their remote cluster and refer to it.
+
+### Dynamic Updates to Agent's Role
+
+In the remote cluster, we provide the `Role` that has the static set of
+`apiGroup`s we'd like agent to be able to manage in the remote cluster. There
+could be a controller that is separately installed and it could add new
+`apiGroup`s as they appear as `InfrastructurePublication`s in the remote
+cluster.
```
negz, line 610: Would this be something like https://github.com/crossplane/crossplane/issues/1637, but at the remote cluster?

design/design-doc-agent.md
```diff
@@ -0,0 +1,665 @@
...
+resources:
+
+* Pull
+  * `CustomResourceDefinition`s of all types that we want the applications to be
+    able to manage and view:
+    * Requirements that are published via `InfrastructurePublication`s. The
+      source of truth will be the remote cluster.
+    * `InfrastructureDefinition`, `InfrastructurePublication` and `Composition`.
+  * `Composition`s to discover what's available to choose. CRs of this type will
+    be read-only, and the source of truth will be the central cluster.
```
negz, line 233: How should we enforce the read-only nature of Compositions, InfraDefs, and InfraPubs?

design/design-doc-agent.md
```diff
@@ -0,0 +1,665 @@
...
+team, which is `*Requirement` types and consume the infrastructure by mounting
+the secret whose name they specified on the `*Requirement` custom resource. To
+keep this experience, we need to have a synchronization loop for the following
+resources:
+
+* Pull
+  * `CustomResourceDefinition`s of all types that we want the applications to be
+    able to manage and view:
+    * Requirements that are published via `InfrastructurePublication`s. The
+      source of truth will be the remote cluster.
```
negz, line 230: Could you add a little context somewhere in this doc that explains why we'd pull the require CRDs from the central cluster rather than mirroring the logic that generates a CRD from an InfraPub in the agent? I'm not entirely convinced that it wouldn't be more flexible for the agent to mirror the logic (e.g. if we wanted the CRD gen process to diverge slightly at some point in future).

design/design-doc-agent.md
lukeweber: > So, I'm leaning towards keeping the name pairing between requirements and annotate them in the central cluster with UID of the namespace it exists in the remote cluster to uniquely identify them in the world of custom resources. UID of the `kube-system` could be worth adding, too, to discover in what cluster that namespace exists.

A quick note - we've run into problems with this on restore since the uid of a namespace isn't durable since it's generated on create. As long as this is updated actively if the object is restored, this seems like a good approach.

design/design-doc-agent.md
lukeweber: Sorry for the confusion. Originally when we were talking about supporting annotations that pointed to a secret and remote namespace, I was commenting on how this model might be quite complex if the agent connected arbitrarily many times between any central and remote namespaces. I think it's clarified now already with the namespace annotation updates you've made.

design/design-doc-agent.md
negz: > the uid of a namespace isn't durable

Agreed. I suspect we won't be able to use the UID of `kube-system` as this would prevent us from replacing a remote cluster that was completely destroyed and recreated. I suspect assigning each agent/remote cluster a UID manually would be the better path.

design/design-doc-agent.md
muvaf: The main reason is that I don't want to have 2 different source of truth for the resulting CRD, i.e. the agent and the Crossplane. For example, let's say Crossplane is updated to a new version in the central cluster but agents connected to the namespaces are still on an old version. Suddenly, CRDs could be different and sync could face unexpected errors because the schemas are different. Also by design, InfraPub and InfraDef resources are read-only in the remote cluster so there isn't much gain to have the generation in remote cluster as well because in the end, the schemas that are of importance are the ones in the central cluster.

design/design-doc-agent.md
muvaf: The agent will blindly `Update` those resources in the remote cluster, meaning even if the user changes a value, it will be overridden right away.

design/design-doc-agent.md
muvaf: > I would suggest that even if we choose not to support the latter case we should structure whatever flags/config file controls this such that we could add support for both if there was user demand.

@negz A flag to pass `matchNamespaces: true` to enable the pairing makes sense here as an option.

Hmm restoration mechanics are interesting. The agent will have three scenarios:
* Each namespace is annotated with the target namespace.
  * No conflict for single remote cluster.
  * Could conflict if more than one cluster is connected and its namespaces have common annotations with other namespaces in other clusters.
* Each namespace targets a namespace with the same name in the central cluster.
  * No conflict for single remote cluster.
  * Could conflict if more than one cluster is connected and its namespaces have same names.
* All namespaces roll up to a default namespace.
  * Could conflict for single remote cluster if different namespaces have requirements with same name.
  * Could conflict if more than one cluster is connected to the same namespace but it's less likely to be in that situation since the default namespace selection is done explicitly during agent setup, probably by an admin.

UID of the namespace of the requirement in the remote cluster is unique enough to prevent all these conflict cases, however, it's not durable as mentioned. The name of the namespace could be considered more durable since restoration of the remote cluster will bring the namespaces back with the same name but it doesn't guarantee enough uniqueness among the N namespace * M clusters. We can actually have a unique identifier of a namespace via `name of the namespace + unique identifier of the cluster` but the problem with finding a unique identifier for the cluster is that there is none except UID of `kube-system` which doesn't meet durability standard of the restoration.

I don't love it but @negz 's suggestion of having the user manually enter a cluster identifier makes sense to achieve the uniqueness although it's not automatic as UID, i.e. user has to be careful about it. So, I'm proposing that:
* Agent will accept a cluster identifier as an installation flag like `agentIdentifier: string`
* If the flag is not provided, it will use UID of `kube-system`.

I imagine that user won't really bother supplying an identifier but in case the remote cluster was to be replaced, they can enter the UID of the `kube-system` as cluster identifier during the setup of the new remote cluster. Let's say a disaster happened and remote cluster is gone suddenly, they can look up the annotations of the requirements created by that cluster to discover what's used as cluster identifier. So we don't force them to enter an identifier, it'll have a sane default.

design/design-doc-agent.md
muvaf: Not exactly, this separate controller would not only create the new roles but also bind them.

muvaf:
> I suspect that folks will grow from running a single Kubernetes cluster that runs Crossplane (or 2-3 clusters that each run Crossplane) to wanting a dedicated, central Crossplane cluster connected to their various Kubernetes clusters via the agent. I'd love it if there was a really smooth transition from the initial 'addon' deployment model to the agent deployment model.

@negz There could be 2 possibilities there; you have 1 cluster with everything in it and then remote clusters start to connect with an agent or you decide that the initial cluster should be a remote cluster. The former is possible as is but the latter has some caveats although I can see that one being more popular since the initial cluster would tend to be a beefy one while IMO central clusters will usually be rather small.

In order for the migration to work in the latter case:
* Crossplane and providers will be deleted from the initial cluster eventually and this shouldn't result in CRD or CR deletion as that'd trigger the actual deletion of those resources.
* An entity needs to copy the requirements, composites and managed resources. Just like backup/restore.

I think the agent sync mechanism could somehow handle the second bullet-point but for the first one there needs to be some changes in Crossplane Helm chart as well as package manager mechanics.

With all that said, I feel like it's a bit stretch to include this in the initial design as there won't be major changes to the agent itself anyway. So, I would like to open a separate issue for migration topic and come back to it after the MVP.

design/design-doc-agent.md
muvaf: @negz Added a paragraph explaining the reasoning.

design/design-doc-agent.md
negz: Oh - I missed that this was granting access to the agent itself.

design/design-doc-agent.md
negz: Got it. I guess RBAC could be used as well to ensure that no one has access to update them in the first place.

negz:
> There could be 2 possibilities there; you have 1 cluster with everything in it and then remote clusters start to connect with an agent or you decide that the initial cluster should be a remote cluster. The former is possible as is but the latter has some caveats although I can see that one being more popular since the initial cluster would tend to be a beefy one while IMO central clusters will usually be rather small.

@muvaf So far we've only discussed the latter use case. I agree it's likely to be more popular and I think it's the only one we need to support, at least to begin with.

> I would like to open a separate issue for migration topic and come back to it after the MVP.

I'd prefer there to be some mention of this in this design, even if it's just a 'future considerations' section. I'm okay deferring the full design for the migration path, but I want to make sure we're building the agent in a way that we expect to support really smooth migrations.

design/design-doc-agent.md
```diff
@@ -0,0 +1,730 @@
...
+    requirements with same name.
+  * Could conflict if more than one cluster is connected to the same namespace
+    but it's less likely to be in that situation since the default namespace
+    selection is done explicitly during agent setup, probably by an admin.
+
+In order to prevent conflicts, the agent will add two annotations to the
+requirements it syncs:
+* `agent.crossplane.io/source-namespace`: The name of the namespace in the
+  remote cluster.
+* `agent.crossplane.io/source-uid`: The unique identifier of the remote
```
negz, line 380: `source-cluster-uid`? `source-uid` sounds like it could mean the UID of the remote requirement.

muvaf:
@negz I've added the migration at the top of the future consideration section. It'll likely be the first thing to tackle after the MVP.

@lukeweber I'm planning to merge this today. Does everything look good to you, too?

design/design-doc-agent.md
```diff
@@ -0,0 +1,586 @@
...
+connect to the central cluster. Additionally, it will need some permissions to
+execute operations in the remote cluster like `Secret` and
+`CustomResourceDefinition` creation. We will look at how the agent will be
+granted permissions to do its job in two separate domains with different
+mechanisms.
+
+#### Authenticating to The Central Cluster
+
+In order to execute any operation, a `ServiceAccount` needs to exist in the
+central cluster with appropriate permissions to read during pull and write
```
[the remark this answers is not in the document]
muvaf: I actually wanted not to use actual component names because I think it makes it harder to understand sentences like `the agent in the agent cluster will connect to Crossplane in the Crossplane cluster` etc. Having them named with *where* they sit in the archutecture instead of *what* they contain sounds easier to understand to me.

PR #1660, opened 2020-07-21
Use CONTAINER_NAME env var to get image if no name is set

### LINKED ISSUE #1659: Crossplane provider installation fails
### What happened?
Following the documentation to install crossplane on an already existing cluster with istio, installing of a provider (in my case gcp 0.10.0) failed.

```
$ kubectl crossplane package install --cluster --namespace crossplane-system crossplane/provider-gcp:v0.10.0 provider-gcp
clusterpackageinstall.packages.crossplane.io/provider-gcp created
apiVersion: packages.crossplane.io/v1alpha1
kind: ClusterPackageInstall
metadata:
  annotations:
      {"apiVersion":"packages.crossplane.io/v1alpha1","kind":"ClusterPackageInstall","metadata":{"annotations":{},"name":"provider-gcp","namespace":"crossplane-system"},"spec":{"package":"crossplane/provider-gcp:v0.10.0"}}
  creationTimestamp: "2020-07-21T19:56:28Z"
  finalizers:
  - finalizer.packageinstall.crossplane.io
  generation: 1
  name: provider-gcp
  namespace: crossplane-system
  resourceVersion: "27639621"
  selfLink: /apis/packages.crossplane.io/v1alpha1/namespaces/crossplane-system/clusterpackageinstalls/provider-gcp
  uid: 8ed1fc10-f2bc-4c57-97d6-26939b683408
spec:
  package: crossplane/provider-gcp:v0.10.0
status:
  conditionedStatus:
    conditions:
    - lastTransitionTime: "2020-07-21T19:56:28Z"
      message: 'failed to get image for pod: failed to find image for container '
      reason: Encountered an error during resource reconciliation
      status: "False"
      type: Synced
```

A describe on the package-manager looks like this:
```
Name:         crossplane-package-manager-6598cb864b-92qt5
Namespace:    crossplane-system
Priority:     0
Node:         shoot--dxp--dxp-core-cpu-worker-8-z1-7d76dc7cc-jvf5s/REDACTED
Start Time:   Tue, 21 Jul 2020 12:40:14 -0700
Labels:       app=crossplane-package-manager
              pod-template-hash=6598cb864b
              release=crossplane
              security.istio.io/tlsMode=istio
Annotations:  cni.projectcalico.org/podIP: REDACTED
              cni.projectcalico.org/podIPs: REDACTED
              kubernetes.io/psp: gardener.privileged
              sidecar.istio.io/status:
                {"version":"e4fecf157fd68413c68551d81bc33530d2d5bf8c350a7b9d365185c2247587da","initContainers":["istio-init"],"containers":["istio-proxy"]...
Status:       Running
IP:           REDACTED
IPs:
  IP:           REDACTED
Controlled By:  ReplicaSet/crossplane-package-manager-6598cb864b
Init Containers:
  istio-init:
    Container ID:  docker://922715f808ab02259db1822fc182c4180315fe223b46d0dc99c943845f2fefc7
    Image:         docker.io/istio/proxyv2:1.4.7-distroless
    Image ID:      docker-pullable://istio/proxyv2@sha256:8ba1681e84fef93b26cf24318dbca395c30954ec39dc1fb70c09878ce4b6b9b2
    Port:          <none>
    Host Port:     <none>
    Command:
      istio-iptables
      -p
      15001
      -z
      15006
      -u
      1337
      -m
      REDIRECT
      -i
      *
      -x
      -b
      *
      -d
      15020
    State:          Terminated
      Reason:       Completed
      Exit Code:    0
      Started:      Tue, 21 Jul 2020 12:40:16 -0700
      Finished:     Tue, 21 Jul 2020 12:40:17 -0700
    Ready:          True
    Restart Count:  0
    Limits:
      cpu:     100m
      memory:  50Mi
    Requests:
      cpu:        10m
      memory:     10Mi
    Environment:  <none>
    Mounts:
      /var/run/secrets/kubernetes.io/serviceaccount from package-manager-token-5fdjq (ro)
Containers:
  crossplane:
    Container ID:  docker://2987fa22388520fb7c75685139a75eff3e69cce7b1180000c42c6205e9444bb8
    Image:         crossplane/crossplane:v0.12.0
    Image ID:      docker-pullable://crossplane/crossplane@sha256:66666e6963af5c8fe5005de83cb6dfa614a00def40765c3bdafad7f1f7fde237
    Port:          <none>
    Host Port:     <none>
    Args:
      package
      manage
      --templates
      --templating-controller-image
      crossplane/templating-controller:v0.2.1
      --debug
    State:          Running
      Started:      Tue, 21 Jul 2020 12:40:40 -0700
    Last State:     Terminated
      Reason:       Error
      Exit Code:    1
      Started:      Tue, 21 Jul 2020 12:40:25 -0700
      Finished:     Tue, 21 Jul 2020 12:40:25 -0700
    Ready:          True
    Restart Count:  2
    Limits:
      cpu:     100m
      memory:  512Mi
    Requests:
      cpu:     100m
      memory:  256Mi
    Environment:
      POD_NAME:       crossplane-package-manager-6598cb864b-92qt5 (v1:metadata.name)
      POD_NAMESPACE:  crossplane-system (v1:metadata.namespace)
    Mounts:
      /var/run/secrets/kubernetes.io/serviceaccount from package-manager-token-5fdjq (ro)
  istio-proxy:
    Container ID:  docker://7f7e739183157dac9d84e2d956e85bcb62b60bdd6d23adc906a2add9acbb24b1
    Image:         docker.io/istio/proxyv2:1.4.7-distroless
    Image ID:      docker-pullable://istio/proxyv2@sha256:8ba1681e84fef93b26cf24318dbca395c30954ec39dc1fb70c09878ce4b6b9b2
    Port:          15090/TCP
    Host Port:     0/TCP
    Args:
      proxy
      sidecar
      --domain
      $(POD_NAMESPACE).svc.cluster.local
      --configPath
      /etc/istio/proxy
      --binaryPath
      /usr/local/bin/envoy
      --serviceCluster
      crossplane-package-manager.$(POD_NAMESPACE)
      --drainDuration
      45s
      --parentShutdownDuration
      1m0s
      --discoveryAddress
      istio-pilot.istio-system:15011
      --datadogAgentAddress
      $(HOST_IP):8126
      --proxyLogLevel=warning
      --proxyComponentLogLevel=misc:error
      --connectTimeout
      10s
      --proxyAdminPort
      15000
      --concurrency
      2
      --controlPlaneAuthPolicy
      MUTUAL_TLS
      --dnsRefreshRate
      300s
      --statusPort
      15020
      --applicationPorts
      --trust-domain=cluster.local
    State:          Running
      Started:      Tue, 21 Jul 2020 12:40:22 -0700
    Ready:          True
    Restart Count:  0
    Limits:
      cpu:     200m
      memory:  128Mi
    Requests:
      cpu:      20m
      memory:   32Mi
    Readiness:  http-get http://:15020/healthz/ready delay=1s timeout=1s period=2s #success=1 #failure=30
    Environment:
      POD_NAME:                          crossplane-package-manager-6598cb864b-92qt5 (v1:metadata.name)
      POD_NAMESPACE:                     crossplane-system (v1:metadata.namespace)
      INSTANCE_IP:                        (v1:status.podIP)
      SERVICE_ACCOUNT:                    (v1:spec.serviceAccountName)
      HOST_IP:                            (v1:status.hostIP)
      ISTIO_META_POD_PORTS:              [
                                         ]
      ISTIO_META_CLUSTER_ID:             Kubernetes
      ISTIO_META_POD_NAME:               crossplane-package-manager-6598cb864b-92qt5 (v1:metadata.name)
      ISTIO_META_CONFIG_NAMESPACE:       crossplane-system (v1:metadata.namespace)
      SDS_ENABLED:                       false
      ISTIO_META_INTERCEPTION_MODE:      REDIRECT
      ISTIO_META_INCLUDE_INBOUND_PORTS:  
      ISTIO_METAJSON_ANNOTATIONS:        {"kubernetes.io/psp":"gardener.kube-system.calico"}
      ISTIO_METAJSON_LABELS:             {"app":"crossplane-package-manager","pod-template-hash":"6598cb864b","release":"crossplane"}
      ISTIO_META_WORKLOAD_NAME:          crossplane-package-manager
      ISTIO_META_OWNER:                  kubernetes://apis/apps/v1/namespaces/crossplane-system/deployments/crossplane-package-manager
      ISTIO_META_MESH_ID:                cluster.local
      ISTIO_KUBE_APP_PROBERS:            {}
    Mounts:
      /etc/certs/ from istio-certs (ro)
      /etc/istio/proxy from istio-envoy (rw)
      /var/run/secrets/kubernetes.io/serviceaccount from package-manager-token-5fdjq (ro)
Conditions:
  Type              Status
  Initialized       True 
  Ready             True 
  ContainersReady   True 
  PodScheduled      True 
Volumes:
  package-manager-token-5fdjq:
    Type:        Secret (a volume populated by a Secret)
    SecretName:  package-manager-token-5fdjq
    Optional:    false
  istio-envoy:
    Type:       EmptyDir (a temporary directory that shares a pod's lifetime)
    Medium:     Memory
    SizeLimit:  <unset>
  istio-certs:
    Type:        Secret (a volume populated by a Secret)
    SecretName:  istio.package-manager
    Optional:    true
QoS Class:       Burstable
Node-Selectors:  <none>
Tolerations:     node.kubernetes.io/not-ready:NoExecute for 300s
                 node.kubernetes.io/unreachable:NoExecute for 300s
Events:          <none>
```

### How can we reproduce it?
Not sure, it might be related to istio sidecar injection:
https://crossplane.slack.com/archives/CEG3T90A1/p1595366947224600

Crossplane version: 0.12

Running on Gardener on top of GKE.
Kubernetes version 1.16.10

hasheddan:

This updates the logic when finding a container in a pod spec to check
the CONTAINER_NAME env var when no name is supplied and there is more
than one container in the Pod.

Fixes #1659 

I have:

Manual testing will be done with updated helm chart before this is moved out of draft.

### DIFF
--- a/cluster/charts/crossplane-controllers/templates/package-manager-deployment.yaml
+++ b/cluster/charts/crossplane-controllers/templates/package-manager-deployment.yaml
@@ -72,6 +72,9 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
+        # The container name
+        - name: CONTAINER_NAME
+          value: {{ .Chart.Name }}
         resources:
           {{- toYaml .Values.resourcesPackageManager | nindent 12 }}
       {{- if .Values.hostedConfig.enabled }}
--- a/pkg/packages/pod.go
+++ b/pkg/packages/pod.go
@@ -31,6 +31,8 @@ const (
 	PodNameEnvVar = "POD_NAME"
 	// PodNamespaceEnvVar is the env variable for getting the pod namespace via downward api
 	PodNamespaceEnvVar = "POD_NAMESPACE"
+	// ContainerNameEnvVar is the env variable for getting the container name if provided
+	ContainerNameEnvVar = "CONTAINER_NAME"
 )
 
 // GetRunningPod will get the pod object for the currently running pod.  This assumes that the
@@ -91,16 +93,20 @@ func GetSpecContainer(spec v1.PodSpec, name string, initContainer bool) (v1.Cont
 	return image, nil
 }
 
-// GetMatchingContainer returns the container from the given set of containers that matches the
-// given name.  If the given container list has only 1 item then the name field is ignored and
-// that container is returned.
+// GetMatchingContainer returns the container from the given set of containers
+// that matches the given name. If the given container list has only 1 item and
+// no name is specified then that container is returned. If more than one
+// container exists and no name is specified it attempts to access a name from
+// env vars, before then checking for the existence of the container.
 func GetMatchingContainer(containers []v1.Container, name string) (v1.Container, error) {
 	var result *v1.Container
-	if len(containers) == 1 {
-		// if there is only one pod, use its image rather than require a set container name
+	if len(containers) == 1 && name == "" {
+		// if there is only one container and no name is specified, return it
 		result = &containers[0]
 	} else {
-		// if there are multiple pods, we require the container to have the expected name
+		if name == "" {
+			name = os.Getenv(ContainerNameEnvVar)
+		}
 		for _, container := range containers {
 			container := container // pin range scoped container var
 			if container.Name == name {
--- a/cluster/charts/crossplane/templates/package-manager-deployment.yaml
+++ b/cluster/charts/crossplane/templates/package-manager-deployment.yaml
@@ -72,6 +72,9 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
+        # The container name
+        - name: CONTAINER_NAME
+          value: {{ .Chart.Name }}
         resources:
           {{- toYaml .Values.resourcesPackageManager | nindent 12 }}
       {{- if .Values.hostedConfig.enabled }}
--- a/pkg/packages/executorinfo_test.go
+++ b/pkg/packages/executorinfo_test.go
@@ -70,10 +70,11 @@ func TestExecutorInfoDiscoverer_Discover(t *testing.T) {
 	}
 
 	tests := []struct {
-		name      string
-		imageName string
-		d         ExecutorInfoDiscoverer
-		want      want
+		name         string
+		imageName    string
+		imageNameEnv string
+		d            ExecutorInfoDiscoverer
+		want         want
 	}{
 		{
 			name: "FailedGetRunningPod",
@@ -105,14 +106,15 @@ func TestExecutorInfoDiscoverer_Discover(t *testing.T) {
 			},
 		},
 		{
-			name: "SuccessfulDiscovery",
+			name:         "SuccessfulDiscovery",
+			imageNameEnv: "foo",
 			d: &KubeExecutorInfoDiscoverer{
 				Client: &test.MockClient{
 					MockGet: func(ctx context.Context, key client.ObjectKey, obj runtime.Object) error {
 						*obj.(*corev1.Pod) = corev1.Pod{
 							ObjectMeta: metav1.ObjectMeta{Name: key.Name, Namespace: key.Namespace},
 							Spec: corev1.PodSpec{
-								Containers: []corev1.Container{{Name: "foo", Image: "foo-image"}},
+								Containers: []corev1.Container{{Name: "foo", Image: "foo-image", Env: []corev1.EnvVar{{Name: "CONTAINER_NAME", Value: "foo"}}}},
 							},
 						}
 						return nil
@@ -150,6 +152,7 @@ func TestExecutorInfoDiscoverer_Discover(t *testing.T) {
 
 			os.Setenv(PodNameEnvVar, "podName")
 			os.Setenv(PodNamespaceEnvVar, "podNamespace")
+			os.Setenv(ContainerNameEnvVar, tt.imageNameEnv)
 			os.Setenv(PodImageNameEnvVar, tt.imageName)
 			os.Setenv(PodImagePullPolicyEnvVar, "Always")
 


hasheddan:
@turkenh tagged you for review here since this can affect host aware mode. I believe it will play nicely and will make sure to test it, but thought you might catch anything I missed :+1:

pkg/packages/pod.go
```diff
@@ -91,16 +93,20 @@ func GetSpecContainer(spec v1.PodSpec, name string, initContainer bool) (v1.Cont
...
 	var result *v1.Container
-	if len(containers) == 1 {
-		// if there is only one pod, use its image rather than require a set container name
+	if len(containers) == 1 && name == "" {
+		// if there is only one container and no name is specified, return it
 		result = &containers[0]
 	} else {
-		// if there are multiple pods, we require the container to have the expected name
+		if name == "" {
+			name = os.Getenv(ContainerNameEnvVar)
```
turkenh, line 108: `GetMatchingContainer` function already takes `name` as an argument, wouldn't it be cleaner if we leave setting the name of the container properly to the caller instead? Tracking the name parameter, I found it is set as empty here:  https://github.com/crossplane/crossplane/blob/0d1aeec8b505bf80de46f16bf25aaed5921a3cfa/pkg/packages/executorinfo.go#L87

hasheddan: @turkenh I totally agree. However, this package is already in a bit of a weird pattern IMO because it uses environment variables for `GetRunningPod` so I was attempting to mirror that here. I would be fine going either way. It think some of this code should be replaced soon anyway.

cluster/charts/crossplane-controllers/templates/package-manager-deployment.yaml
```diff
@@ -72,6 +72,9 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
+        # The container name
+        - name: CONTAINER_NAME
+          value: {{ .Chart.Name }}
```
muvaf, line 77: Do you think it'd cause a problem if we used a hard-coded value as container name, like `manager` or sth, both here and `name` field of the container? I couldn't really get why `{{ .Chart.Name }} ` had been chosen.

hasheddan: In the general case no, but if someone was running in host aware mode then they may have multiple manager deployments in the same name namespace but running against different API servers. If the manager deployment name was hardcoded then they would have collisions.

muvaf: > If the manager deployment name was hardcoded then they would have collisions.

That's not the deployment name though, it's the container's identifier in the `Pod`. Also, `.Chart.Name` is actually same for all Crossplane installations as opposed to `.Release.Name`.

hasheddan: @muvaf good call! Yeah this shouldn't end up conflicting if hard-coded so that may be a better solution 👍

pkg/packages/pod.go
displague: The `PACKAGE_MANAGER_IMAGE` is used in debugging contexts, where the image name can not be determined (there is no running pod, just a process).  The image name is needed so `manage` can create `unpack` jobs.

One other scenario where this is used is: 

https://github.com/operator-framework/community-operators/blob/master/community-operators/crossplane/0.11.0/crossplane.v0.11.0.clusterserviceversion.yaml#L403-L404

In this scenario, the package manager is run with an OLM side-car, and the `manage` process could not determine which pod container represented Crossplane's package manager.

hasheddan: @displague thanks for that context! Sounds like we could just set https://github.com/crossplane/crossplane/pull/1660/files#diff-5024fc0cbab4a78ee75f7db9c2d696dcR76 to `PACKAGE_MANAGER_IMAGE` as in the OLM case and then pass that from the caller of this function. Does that sound good to you?

displague: My concern wasn't about this PR but how external uses could benefit from it.

I think the chart name is fine with how you are matching it now, since it is the same here: https://github.com/crossplane/crossplane/pull/1660/files#diff-5024fc0cbab4a78ee75f7db9c2d696dcR63.

In the OLM manifest, you could replace the `PACKAGE_MANAGER_IMAGE` definition with `CONTAINER_NAME` (`crossplane`) at:
https://github.com/operator-framework/community-operators/blob/master/community-operators/crossplane/0.11.0/crossplane.v0.11.0.clusterserviceversion.yaml#L403-L404

hasheddan: @displague I _think_ the changes as proposed here would not actually affect the OLM case at all. If we keep `PACKAGE_MANAGER_IMAGE` as is then `GetMatchingContainer` is never called. Therefore, we would have the following cases:
1. If `PACKAGE_MANAGER_IMAGE` is supplied then we will always use it (useful for overriding in debugging as you mentioned, or specifying a different unpack image than package manager image).
2. If `CONTAINER_NAME` is supplied then we will use it to find the container and it's corresponding image (safer than not supplying in the case that a sidecar is injected).
3. If neither are supplied, if there is only one container then we will use it. Otherwise we will fail.

displague: It's one less line to tweak in the OLM manifest if that manifest adopts the new `CONTAINER_NAME` option.

hasheddan:
I am going to close this because I anticipate that there could be meaningful changes to the deployment model here with the new package manager in our next release. This problem will likely be overwritten.

PR #1662, opened 2020-07-24
local dev: preload locally cached Helm image to dev cluster

muvaf:

It's unnecessary to download Tiller image each time so we load it before initialization of Helm.

Fixes #

I have:

Manually.

### DIFF
--- a/cluster/local/microk8s.sh
+++ b/cluster/local/microk8s.sh
@@ -33,7 +33,7 @@ function wait_for_ready() {
 function copy_image_to_cluster() {
     local build_image=$1
     local final_image=$2
-    docker tag "${build_image}" "${final_image}" && docker push "${final_image}"
+    docker tag "${build_image}" "localhost:32000/${final_image}" && docker push "${final_image}"
     echo "Tagged image: ${final_image}"
 }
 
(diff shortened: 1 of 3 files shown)


cluster/local/microk8s.sh
```diff
@@ -52,6 +52,10 @@ case "${1:-}" in
 
     microk8s.enable registry
 
+    # We'll use locally cached image instead of having it downloaded by kind
+    # cluster.
+    docker pull "gcr.io/kubernetes-helm/tiller:${HELM_VERSION}"
+    copy_image_to_cluster "gcr.io/kubernetes-helm/tiller:${HELM_VERSION}" "gcr.io/kubernetes-helm/tiller:${HELM_VERSION}"
```
hasheddan, line 58: I believe you are going to want to make the second image name here correspond to the microk8s registry. I _think_ this will attempt to just push it back to gcr https://microk8s.io/docs/registry-built-in

cluster/local/local.sh
```diff
@@ -23,6 +23,10 @@ function copy_image_to_cluster() {
 
 case "${1:-}" in
   up)
+    # We'll use locally cached image instead of having it downloaded by kind
+    # cluster.
+    docker pull "gcr.io/kubernetes-helm/tiller:${HELM_VERSION}"
+    copy_image_to_cluster "gcr.io/kubernetes-helm/tiller:${HELM_VERSION}" "gcr.io/kubernetes-helm/tiller:${HELM_VERSION}"
```
hasheddan, line 29: I think this is a no-op in the local case, it is just tagging the image with its own name

cluster/local/minikube.sh
```diff
@@ -63,6 +63,10 @@ case "${1:-}" in
 
     minikube addons enable ingress
 
+    # We'll use locally cached image instead of having it downloaded by kind
+    # cluster.
```
hasheddan, line 67: Did you intend for each of these comments to say "downloaded by kind cluster"?

cluster/local/microk8s.sh
muvaf: Thanks for catching! I've changed the copy_image_to_cluster function to prepend `localhost:32000` as documented in the link you provided.

muvaf: Me neither and I really suspect that even anyone is using this.

PR #1674, opened 2020-08-13
Remove reclaim policy from composite resources

negz:

Composite resources will always be deleted when their bound requirement is deleted. See https://github.com/crossplane/crossplane-runtime/issues/179#issuecomment-671686000 for details.

I have:

### DIFF
--- a/pkg/controller/apiextensions/composite/api.go
+++ b/pkg/controller/apiextensions/composite/api.go
@@ -260,26 +260,22 @@ type APIConfigurator struct {
 	client client.Client
 }
 
-// Configure the supplied composite resource using its composition.
+// Configure any required fields that were omitted from the composite resource
+// by copying them from its composition.
 func (c *APIConfigurator) Configure(ctx context.Context, cp resource.Composite, comp *v1alpha1.Composition) error {
 	apiVersion, kind := cp.GetObjectKind().GroupVersionKind().ToAPIVersionAndKind()
 	if comp.Spec.From.APIVersion != apiVersion || comp.Spec.From.Kind != kind {
 		return errors.New(errCompositionNotCompatible)
 	}
 
-	if cp.GetReclaimPolicy() != "" && cp.GetWriteConnectionSecretToReference() != nil {
+	if cp.GetWriteConnectionSecretToReference() != nil {
 		return nil
 	}
 
-	if cp.GetReclaimPolicy() == "" {
-		cp.SetReclaimPolicy(comp.Spec.ReclaimPolicy)
-	}
-	if cp.GetWriteConnectionSecretToReference() == nil {
-		cp.SetWriteConnectionSecretToReference(&runtimev1alpha1.SecretReference{
-			Name:      string(cp.GetUID()),
-			Namespace: comp.Spec.WriteConnectionSecretsToNamespace,
-		})
-	}
+	cp.SetWriteConnectionSecretToReference(&runtimev1alpha1.SecretReference{
+		Name:      string(cp.GetUID()),
+		Namespace: comp.Spec.WriteConnectionSecretsToNamespace,
+	})
 
 	return errors.Wrap(c.client.Update(ctx, cp), errUpdateComposite)
 }
--- a/pkg/controller/apiextensions/requirement/reconciler.go
+++ b/pkg/controller/apiextensions/requirement/reconciler.go
@@ -61,7 +61,7 @@ const (
 // Event reasons.
 const (
 	reasonBind      event.Reason = "BindCompositeResource"
-	reasonUnbind    event.Reason = "UnbindCompositeResource"
+	reasonDelete    event.Reason = "DeleteCompositeResource"
 	reasonConfigure event.Reason = "ConfigureCompositeResource"
 	reasonPropagate event.Reason = "PropagateConnectionSecret"
 )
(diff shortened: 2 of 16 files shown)


pkg/controller/apiextensions/requirement/reconciler.go
```diff
@@ -337,25 +342,25 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error)
...
 		log = log.WithValues("deletion-timestamp", rq.GetDeletionTimestamp())
 
-		if err := r.requirement.Unbind(ctx, rq, cp); err != nil {
+		if err := r.composite.Delete(ctx, rq, cp); err != nil {
 			// If we didn't hit this error last time we'll be requeued
 			// implicitly due to the status update. Otherwise we want to retry
 			// after a brief wait, in case this was a transient error.
-			log.Debug("Cannot unbind requirement", "error", err, "requeue-after", time.Now().Add(aShortWait))
-			record.Event(rq, event.Warning(reasonUnbind, err))
+			log.Debug("Cannot delete requirement", "error", err, "requeue-after", time.Now().Add(aShortWait))
```
hasheddan, line 349:
```suggestion
			log.Debug("Cannot delete composite resource", "error", err, "requeue-after", time.Now().Add(aShortWait))
```

pkg/controller/apiextensions/requirement/reconciler.go
```diff
@@ -337,25 +342,25 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error)
...
+			record.Event(rq, event.Warning(reasonDelete, err))
 			rq.SetConditions(v1alpha1.Deleting(), v1alpha1.ReconcileError(err))
 			return reconcile.Result{RequeueAfter: aShortWait}, errors.Wrap(r.client.Status().Update(ctx, rq), errUpdateRequirementStatus)
 		}
 
 		log.Debug("Successfully unbound composite resource")
```
hasheddan, line 355:
```suggestion
		log.Debug("Successfully deleted composite resource")
```

negz:
Good catches @hasheddan - I squashed the fixes.

pkg/controller/apiextensions/composite/api.go
```diff
@@ -260,26 +260,22 @@ type APIConfigurator struct {
 	client client.Client
 }
 
-// Configure the supplied composite resource using its composition.
+// Configure any required fields that were omitted from the composite resource
+// by copying them from its composite.
```
muvaf, line 264:
```suggestion
// by copying them from its composition.
```

pkg/controller/apiextensions/requirement/reconciler.go
```diff
@@ -337,25 +342,25 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error)
...
 			rq.SetConditions(v1alpha1.Deleting(), v1alpha1.ReconcileError(err))
 			return reconcile.Result{RequeueAfter: aShortWait}, errors.Wrap(r.client.Status().Update(ctx, rq), errUpdateRequirementStatus)
 		}
 
-		log.Debug("Successfully unbound composite resource")
-		record.Event(rq, event.Normal(reasonUnbind, "Successfully unbound composite resource"))
+		log.Debug("Successfully deleted composite resource")
+		record.Event(rq, event.Normal(reasonDelete, "Successfully deleted composite resource"))
 
 		if err := r.requirement.RemoveFinalizer(ctx, rq); err != nil {
```
muvaf, line 358: This assumes that in case `Delete` call is successful, we'll remove the finalizer even though the composite resource is still there. I think a blocking requirement deletion until composite is gone could be more assuring. For example, if the deletion of composite resource fails, no one will really know because consumer thinks what he created (requirement) is deleted. So, operator will need to scan the cluster for hanging composite resources to make sure they are not costing money. On the other hand, as a consumer, I delete my app via `helm delete` and deletion of cloud resources could take a long time like Kubernetes clusters. So, deletion could time out.

So, there is a trade-off there and I believe users will get much more sad if they have to deal with set of hanging composite resources to make sure they don't waste money. In the app deletion case, I feel like it's fair to expect consumers to know that cloud resources they created take time to delete and to account for that. Not great, but better than hanging composites.

negz: This may be true, but this PR doesn't introduce this behaviour. Previously we'd delete the composite resource during the unbind call if the reclaim policy was `Delete`. All this PR does is rename `r.requirement.Unbind` to `r.composite.Delete` and remove the part that returned early before deleting if the reclaim policy was `Retain`.

I'd also point out that we use this pattern in the resource claim reconciler; we don't double check that the Delete call actually worked before removing the finalizer. We assume that if we successfully ask the Kubernetes API to delete the managed resource (in the claim reconciler case) that it will eventually be deleted. Adding our own logic to ensure this smells a little like pre-emptive optimisation to me - I'd rather add complexity that if and when we get reports that we're leaking composites in practice.

pkg/controller/apiextensions/composite/api.go
negz: Squashed a fix.

negz:
Just tested this out using the GCP example from our docs; appears to work as expected.

PR #1675, opened 2020-08-13
Implement initial v2 Package Manager

hasheddan:

Implementation of the new package manager.

TODO:
- [x] Complete unit tests (this PR)
- [x] Clean up commits (this PR)
- [x] Restrict what a `ProviderRevision` and `ConfigurationRevision` can install (this PR -- linting packages)
- [x] Update helm chart with required permissions (this PR -- related to #1706)
- [x] Remove `replace` stanza for `crossplane-runtime` (this PR)
- [x] Add more descriptive conditions (this PR)
- Update design doc (follow-up PR)
- Set extra permission requests in revision status (follow-up PR?)
- Restrict based on crossplane version restriction (follow-up PR)
- Add hosted mode support (follow-up PR -- tracked by https://github.com/crossplane/crossplane/issues/1691)

Fixes https://github.com/crossplane/crossplane/issues/1690

I have:

Manual testing artifacts will be added below.

### DIFF
--- a/pkg/controller/pkg/manager/reconciler.go
+++ b/pkg/controller/pkg/manager/reconciler.go
@@ -1,0 +1,354 @@
+/*
+Copyright 2020 The Crossplane Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package manager
+
+import (
+	"context"
+	"math"
+	"strings"
+	"time"
+
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	runtimev1alpha1 "github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
+	"github.com/crossplane/crossplane-runtime/pkg/event"
+	"github.com/crossplane/crossplane-runtime/pkg/logging"
+	"github.com/crossplane/crossplane-runtime/pkg/meta"
+	"github.com/crossplane/crossplane-runtime/pkg/resource"
+
+	"github.com/crossplane/crossplane/apis/pkg/v1alpha1"
+)
+
+const (
+	parentLabel      = "pkg.crossplane.io/package"
+	reconcileTimeout = 1 * time.Minute
+
+	shortWait     = 30 * time.Second
+	veryShortWait = 5 * time.Second
+)
+
+const (
+	errGetPackage            = "cannot get package"
+	errListRevisions         = "cannot list revisions for package"
+	errCreatePackageRevision = "cannot create package revision"
+	errGCPackageRevision     = "cannot garbage collect old package revision"
+	errGCInstallPod          = "cannot garbage collect old package revision install pod"
+
+	errUpdateStatus                  = "cannot update package status"
+	errUpdateInactivePackageRevision = "cannot update inactive package revision"
+	errUpdateActivePackageRevision   = "cannot update active package revision"
+)
+
+// Event reasons.
+const (
+	reasonList               event.Reason = "ListRevision"
+	reasonUnpack             event.Reason = "UnpackPackage"
+	reasonTransitionRevision event.Reason = "TransitionRevision"
+	reasonGarbageCollect     event.Reason = "GarbageCollect"
+	reasonInstall            event.Reason = "InstallPackageRevision"
+)
+
+// ReconcilerOption is used to configure the Reconciler.
+type ReconcilerOption func(*Reconciler)
+
+// WithNewPackageFn determines the type of package being reconciled.
+func WithNewPackageFn(f func() v1alpha1.Package) ReconcilerOption {
+	return func(r *Reconciler) {
+		r.newPackage = f
+	}
+}
+
+// WithNewPackageRevisionFn determines the type of package being reconciled.
+func WithNewPackageRevisionFn(f func() v1alpha1.PackageRevision) ReconcilerOption {
+	return func(r *Reconciler) {
+		r.newPackageRevision = f
+	}
+}
+
+// WithNewPackageRevisionListFn determines the type of package being reconciled.
+func WithNewPackageRevisionListFn(f func() v1alpha1.PackageRevisionList) ReconcilerOption {
+	return func(r *Reconciler) {
+		r.newPackageRevisionList = f
+	}
+}
+
+// WithPodManager specifies how the Reconciler should manage install pods.
+func WithPodManager(m PodManager) ReconcilerOption {
+	return func(r *Reconciler) {
+		r.podManager = m
+	}
+}
+
+// WithLogger specifies how the Reconciler should log messages.
+func WithLogger(log logging.Logger) ReconcilerOption {
+	return func(r *Reconciler) {
+		r.log = log
+	}
+}
+
+// WithRecorder specifies how the Reconciler should record Kubernetes events.
+func WithRecorder(er event.Recorder) ReconcilerOption {
+	return func(r *Reconciler) {
+		r.record = er
+	}
+}
+
+// Reconciler reconciles packages.
+type Reconciler struct {
+	client     resource.ClientApplicator
+	podManager PodManager
+	log        logging.Logger
+	record     event.Recorder
+
+	newPackage             func() v1alpha1.Package
+	newPackageRevision     func() v1alpha1.PackageRevision
+	newPackageRevisionList func() v1alpha1.PackageRevisionList
+}
+
+// SetupProvider adds a controller that reconciles Providers.
+func SetupProvider(mgr ctrl.Manager, l logging.Logger, namespace string) error {
+	name := "packages/" + strings.ToLower(v1alpha1.ProviderGroupKind)
+	np := func() v1alpha1.Package { return &v1alpha1.Provider{} }
+	nr := func() v1alpha1.PackageRevision { return &v1alpha1.ProviderRevision{} }
+	nrl := func() v1alpha1.PackageRevisionList { return &v1alpha1.ProviderRevisionList{} }
+
+	r := NewReconciler(mgr,
+		WithNewPackageFn(np),
+		WithNewPackageRevisionFn(nr),
+		WithNewPackageRevisionListFn(nrl),
+		WithPodManager(NewPackagePodManager(mgr.GetClient(), namespace)),
+		WithLogger(l.WithValues("controller", name)),
+		WithRecorder(event.NewAPIRecorder(mgr.GetEventRecorderFor(name))),
+	)
+
+	return ctrl.NewControllerManagedBy(mgr).
+		Named(name).
+		For(&v1alpha1.Provider{}).
+		Owns(&v1alpha1.ProviderRevision{}).
+		Complete(r)
+}
+
+// SetupConfiguration adds a controller that reconciles Configurations.
+func SetupConfiguration(mgr ctrl.Manager, l logging.Logger, namespace string) error {
+	name := "packages/" + strings.ToLower(v1alpha1.ConfigurationGroupKind)
+	np := func() v1alpha1.Package { return &v1alpha1.Configuration{} }
+	nr := func() v1alpha1.PackageRevision { return &v1alpha1.ConfigurationRevision{} }
+	nrl := func() v1alpha1.PackageRevisionList { return &v1alpha1.ConfigurationRevisionList{} }
+
+	r := NewReconciler(mgr,
+		WithNewPackageFn(np),
+		WithNewPackageRevisionFn(nr),
+		WithNewPackageRevisionListFn(nrl),
+		WithPodManager(NewPackagePodManager(mgr.GetClient(), namespace)),
+		WithLogger(l.WithValues("controller", name)),
+		WithRecorder(event.NewAPIRecorder(mgr.GetEventRecorderFor(name))),
+	)
+
+	return ctrl.NewControllerManagedBy(mgr).
+		Named(name).
+		For(&v1alpha1.Configuration{}).
+		Owns(&v1alpha1.ConfigurationRevision{}).
+		Complete(r)
+}
+
+// NewReconciler creates a new package reconciler.
+func NewReconciler(mgr ctrl.Manager, opts ...ReconcilerOption) *Reconciler {
+	r := &Reconciler{
+		client: resource.ClientApplicator{
+			Client:     mgr.GetClient(),
+			Applicator: resource.NewAPIPatchingApplicator(mgr.GetClient()),
+		},
+		podManager: NewPackagePodManager(mgr.GetClient(), "crossplane-system"),
+		log:        logging.NewNopLogger(),
+		record:     event.NewNopRecorder(),
+	}
+
+	for _, f := range opts {
+		f(r)
+	}
+
+	return r
+}
+
+// Reconcile package.
+func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error) { // nolint:gocyclo
+	log := r.log.WithValues("request", req)
+	log.Debug("Reconciling")
+
+	ctx, cancel := context.WithTimeout(context.Background(), reconcileTimeout)
+	defer cancel()
+
+	p := r.newPackage()
+	if err := r.client.Get(ctx, req.NamespacedName, p); err != nil {
+		// There's no need to requeue if we no longer exist. Otherwise we'll be
+		// requeued implicitly because we return an error.
+		log.Debug(errGetPackage, "error", err)
+		return reconcile.Result{}, errors.Wrap(resource.IgnoreNotFound(err), errGetPackage)
+	}
+
+	log = log.WithValues(
+		"uid", p.GetUID(),
+		"version", p.GetResourceVersion(),
+		"name", p.GetName(),
+	)
+
+	// Get existing package revisions.
+	prs := r.newPackageRevisionList()
+	if err := r.client.List(ctx, prs, client.MatchingLabels(map[string]string{parentLabel: p.GetName()})); resource.IgnoreNotFound(err) != nil {
+		log.Debug(errListRevisions, "error", err)
+		r.record.Event(p, event.Warning(reasonList, err))
+		return reconcile.Result{RequeueAfter: shortWait}, nil
+	}
+
+	p.SetConditions(v1alpha1.Unpacking())
+
+	hash, err := r.podManager.Run(ctx, p)
+	if err != nil {
+		log.Debug(errListRevisions, "error", err)
+		r.record.Event(p, event.Warning(reasonUnpack, err))
+		return reconcile.Result{RequeueAfter: shortWait}, errors.Wrap(r.client.Status().Update(ctx, p), errUpdateStatus)
+	}
+
+	if hash == "" {
+		r.record.Event(p, event.Normal(reasonUnpack, "waiting for unpack to complete"))
+		return reconcile.Result{RequeueAfter: veryShortWait}, errors.Wrap(r.client.Status().Update(ctx, p), errUpdateStatus)
+	}
+
+	// Set the current revision to match the install job for image.
+	p.SetCurrentRevision(packNHash(p.GetName(), hash))
+
+	pr := r.newPackageRevision()
+	maxRevision := int64(0)
+	oldestRevision := int64(math.MaxInt64)
+	oldestRevisionIndex := -1
+	revisions := prs.GetRevisions()
+
+	// Check to see if revision already exists.
+	for index, rev := range revisions {
+		revisionNum := rev.GetRevision()
+
+		// Set max revision to the highest numbered existing revision.
+		if revisionNum > maxRevision {
+			maxRevision = revisionNum
+		}
+
+		// Set oldest revision to the lowest numbered revision and record its
+		// index.
+		if revisionNum < oldestRevision {
+			oldestRevision = revisionNum
+			oldestRevisionIndex = index
+		}
+		// If revision name is same as current revision, then revision already exists.
+		if rev.GetName() == p.GetCurrentRevision() {
+			pr = rev
+			// If current revision exists, is not active, and package has an
+			// automatic activation policy, we should set the revision to
+			// active.
+			p.SetConditions(v1alpha1.Active())
+			if p.GetActivationPolicy() == v1alpha1.AutomaticActivation {
+				pr.SetDesiredState(v1alpha1.PackageRevisionActive)
+			}
+			// Finish iterating through all revisions to make sure all
+			// non-current revisions are inactive.
+			continue
+		}
+		if rev.GetDesiredState() == v1alpha1.PackageRevisionActive {
+			// If revision is not the current revision, set to inactive. This
+			// should always be done, regardless of the package's revision
+			// activation policy.
+			rev.SetDesiredState(v1alpha1.PackageRevisionInactive)
+			if err := r.client.Apply(ctx, rev, resource.MustBeControllableBy(p.GetUID())); err != nil {
+				log.Debug(errUpdateInactivePackageRevision, "error", err)
+				r.record.Event(p, event.Warning(reasonTransitionRevision, err))
+				return reconcile.Result{RequeueAfter: shortWait}, errors.Wrap(r.client.Status().Update(ctx, p), errUpdateStatus)
+			}
+		}
+	}
+
+	// The current revision should always be the highest numbered revision.
+	if pr.GetRevision() < maxRevision || maxRevision == 0 {
+		pr.SetRevision(maxRevision + 1)
+	}
+
+	// Check to see if there are revisions eligible for garbage collection.
+	if len(revisions) > (int(p.GetRevisionHistoryLimit())+1) && p.GetRevisionHistoryLimit() != 0 {
+		gcRev := revisions[oldestRevisionIndex]
+		// Find the oldest revision and delete it.
+		if err := r.client.Delete(ctx, gcRev); err != nil {
+			log.Debug(errGCPackageRevision, "error", err)
+			r.record.Event(p, event.Warning(reasonGarbageCollect, err))
+			return reconcile.Result{RequeueAfter: shortWait}, errors.Wrap(r.client.Status().Update(ctx, p), errUpdateStatus)
+		}
+
+		// Clean up the oldest revision's Pod as well.
+		if err := r.podManager.GarbageCollect(ctx, gcRev.GetSource(), p); err != nil {
+			log.Debug(errGCInstallPod, "error", err)
+			r.record.Event(p, event.Warning(reasonGarbageCollect, err))
+			return reconcile.Result{RequeueAfter: shortWait}, errors.Wrap(r.client.Status().Update(ctx, p), errUpdateStatus)
+		}
+	}
+
+	// If the name of the package revision matches the current revision of the
+	// package, update and return.
+	if pr.GetName() == p.GetCurrentRevision() {
+		// Make sure install pod matches.
+		pr.SetInstallPod(runtimev1alpha1.Reference{Name: imageToPod(p.GetSource())})
+		if err := r.client.Apply(ctx, pr); err != nil {
+			log.Debug(errUpdateActivePackageRevision, "error", err)
+			r.record.Event(p, event.Warning(reasonInstall, err))
+			return reconcile.Result{RequeueAfter: shortWait}, nil
+		}
+
+		// Update Package status to match that of revision.
+		if pr.GetCondition(v1alpha1.TypeHealthy).Status == corev1.ConditionTrue {
+			p.SetConditions(v1alpha1.Healthy())
+		} else {
+			p.SetConditions(v1alpha1.Unhealthy())
+		}
+		return reconcile.Result{}, errors.Wrap(r.client.Status().Update(ctx, p), errUpdateStatus)
+	}
+
+	// Create the non-existent package revision.
+	pr.SetName(packNHash(p.GetName(), hash))
+	pr.SetLabels(map[string]string{parentLabel: p.GetName()})
+	pr.SetInstallPod(runtimev1alpha1.Reference{Name: imageToPod(p.GetSource())})
+	pr.SetSource(p.GetSource())
+
+	pr.SetDesiredState(v1alpha1.PackageRevisionInactive)
+	p.SetConditions(v1alpha1.Inactive())
+	if p.GetActivationPolicy() == v1alpha1.AutomaticActivation {
+		pr.SetDesiredState(v1alpha1.PackageRevisionActive)
+		p.SetConditions(v1alpha1.Active())
+	}
+
+	meta.AddOwnerReference(pr, meta.AsController(meta.TypedReferenceTo(p, p.GetObjectKind().GroupVersionKind())))
+	if err := r.client.Apply(ctx, pr, resource.MustBeControllableBy(p.GetUID())); err != nil {
+		log.Debug(errCreatePackageRevision, "error", err)
+		r.record.Event(p, event.Warning(reasonInstall, err))
+		return reconcile.Result{RequeueAfter: shortWait}, errors.Wrap(r.client.Status().Update(ctx, p), errUpdateStatus)
+	}
+
+	// NOTE(hasheddan): when the first package revision is created for a
+	// package, the health of the package is not set until the revision reports
+	// its health. If updating from an existing revision, the package health
+	// will match the health of the old revision until the next reconcile.
+	return reconcile.Result{}, errors.Wrap(r.client.Status().Update(ctx, p), errUpdateStatus)
+}
--- a/pkg/controller/pkg/manager/reconciler_test.go
+++ b/pkg/controller/pkg/manager/reconciler_test.go
@@ -1,0 +1,656 @@
+/*
+Copyright 2020 The Crossplane Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package manager
+
+import (
+	"context"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/pkg/errors"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	"github.com/crossplane/crossplane-runtime/pkg/event"
+	"github.com/crossplane/crossplane-runtime/pkg/logging"
+	"github.com/crossplane/crossplane-runtime/pkg/resource"
+	"github.com/crossplane/crossplane-runtime/pkg/test"
+	"github.com/crossplane/crossplane/apis/pkg/v1alpha1"
+)
+
+var _ PodManager = &MockPodManager{}
+
+type MockPodManager struct {
+	MockRun            func() (string, error)
+	MockGarbageCollect func() error
+}
+
+func NewMockRunFn(hash string, err error) func() (string, error) {
+	return func() (string, error) {
+		return hash, err
+	}
+}
+
+func NewMockGarbageCollectFn(err error) func() error {
+	return func() error { return err }
+}
+
+func (m *MockPodManager) Run(context.Context, v1alpha1.Package) (string, error) {
+	return m.MockRun()
+}
+func (m *MockPodManager) GarbageCollect(context.Context, string, v1alpha1.Package) error {
+	return m.MockGarbageCollect()
+}
+
+func TestReconcile(t *testing.T) {
+	errBoom := errors.New("boom")
+
+	type args struct {
+		req reconcile.Request
+		rec *Reconciler
+	}
+	type want struct {
+		r   reconcile.Result
+		err error
+	}
+
+	cases := map[string]struct {
+		reason string
+		args   args
+		want   want
+	}{
+		"PackageNotFound": {
+			reason: "We should not return and error and not requeue if package not found.",
+			args: args{
+				req: reconcile.Request{NamespacedName: types.NamespacedName{Name: "test"}},
+				rec: &Reconciler{
+					newPackage: func() v1alpha1.Package { return &v1alpha1.Configuration{} },
+					client: resource.ClientApplicator{
+						Client: &test.MockClient{MockGet: test.NewMockGetFn(kerrors.NewNotFound(schema.GroupResource{}, ""))},
+					},
+					log: logging.NewNopLogger(),
+				},
+			},
+			want: want{
+				r: reconcile.Result{},
+			},
+		},
+		"ErrGetPackage": {
+			reason: "We should return an error if getting package fails.",
+			args: args{
+				req: reconcile.Request{NamespacedName: types.NamespacedName{Name: "test"}},
+				rec: &Reconciler{
+					newPackage: func() v1alpha1.Package { return &v1alpha1.Configuration{} },
+					client: resource.ClientApplicator{
+						Client: &test.MockClient{MockGet: test.NewMockGetFn(errBoom)},
+					},
+					log: logging.NewNopLogger(),
+				},
+			},
+			want: want{
+				r:   reconcile.Result{},
+				err: errors.Wrap(errBoom, errGetPackage),
+			},
+		},
+		"ErrListRevisions": {
+			reason: "We should requeue after short wait if listing revisions for a package fails.",
+			args: args{
+				req: reconcile.Request{NamespacedName: types.NamespacedName{Name: "test"}},
+				rec: &Reconciler{
+					newPackage:             func() v1alpha1.Package { return &v1alpha1.Configuration{} },
+					newPackageRevisionList: func() v1alpha1.PackageRevisionList { return &v1alpha1.ConfigurationRevisionList{} },
+					client: resource.ClientApplicator{
+						Client: &test.MockClient{
+							MockGet:  test.NewMockGetFn(nil),
+							MockList: test.NewMockListFn(errBoom),
+						},
+					},
+					log:    logging.NewNopLogger(),
+					record: event.NewNopRecorder(),
+				},
+			},
+			want: want{
+				r: reconcile.Result{RequeueAfter: shortWait},
+			},
+		},
+		"SuccessfulNoExistingRevisionsAutoActivate": {
+			reason: "We should be active not requeue on successful creation of the first revision with auto activation.",
+			args: args{
+				req: reconcile.Request{NamespacedName: types.NamespacedName{Name: "test"}},
+				rec: &Reconciler{
+					newPackage:             func() v1alpha1.Package { return &v1alpha1.Configuration{} },
+					newPackageRevision:     func() v1alpha1.PackageRevision { return &v1alpha1.ConfigurationRevision{} },
+					newPackageRevisionList: func() v1alpha1.PackageRevisionList { return &v1alpha1.ConfigurationRevisionList{} },
+					client: resource.ClientApplicator{
+						Client: &test.MockClient{
+							MockGet: test.NewMockGetFn(nil, func(o runtime.Object) error {
+								p := o.(*v1alpha1.Configuration)
+								p.SetName("test")
+								p.SetGroupVersionKind(v1alpha1.ConfigurationGroupVersionKind)
+								return nil
+							}),
+							MockList: test.NewMockListFn(kerrors.NewNotFound(schema.GroupResource{}, "")),
+							MockStatusUpdate: test.NewMockStatusUpdateFn(nil, func(o runtime.Object) error {
+								want := &v1alpha1.Configuration{}
+								want.SetName("test")
+								want.SetGroupVersionKind(v1alpha1.ConfigurationGroupVersionKind)
+								want.SetCurrentRevision("test-1234567")
+								want.SetConditions(v1alpha1.Active())
+								if diff := cmp.Diff(want, o); diff != "" {
+									t.Errorf("-want, +got:\n%s", diff)
+								}
+								return nil
+							}),
+						},
+						Applicator: resource.ApplyFn(func(_ context.Context, _ runtime.Object, _ ...resource.ApplyOption) error {
+							return nil
+						}),
+					},
+					podManager: &MockPodManager{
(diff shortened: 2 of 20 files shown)


apis/pkg/v1alpha1/provider_types.go
```diff
@@ -0,0 +1,105 @@
...
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   ProviderSpec   `json:"spec,omitempty"`
+	Status ProviderStatus `json:"status,omitempty"`
+}
+
+// ProviderSpec specifies details about a request to install a provider to
+// Crossplane.
+type ProviderSpec struct {
+	ControllerOptions `json:",inline"`
```
hasheddan, line 46: `ProviderSpec` differs from `ConfigurationSpec` in that it uses this `ControllerOptions` struct, which is also part of the `PackageRevisionSpec`. The `ConfigurationSpec` has the `ImagePullSecrets` and `ImagePullPolicy` parts of the `ControllerOptions`, but not the `ServiceAccountOptions` because the `Configuration` doesn't produce a controller that needs an SA.

We are going to need to allow configuration of the Provider's packaged controller Deployment, but I think it should be completely separate from the configuration of the unpack Job. The first step to get here would be to add `ImagePullSecrets` and `ImagePullPolicy` to `ProviderSpec` outside of the `ControllerOptions` struct. These would be the specific fields for configuring the unpack Job, just like they are in `ConfigurationSpec`.

This leaves the need to determine how to allow a user to pass configuration to the `Deployment`. We have discussed a few approaches in the past, but haven't landed on a clear direction, mostly because we don't have enough data on how folks will want to configure providers. I think we can confidently say the following:

1. Passing configuration to a Provider deployment is a desirable feature.
2. Allowing passing of configuration to a Provider deployment is a potential major security risk to anyone hosting a Crossplane. This likely means the package manager needs to be able to turn this feature on or off, similar to how it does now.
3. The default Provider deployment should be as secure as possible given the typical use case.
4. Configuration should be passed at package "install" time rather than package "build" time. If there is a configuration that is required for every deployment of a specific Provider, the package author should specify that in the documentation, and users should provide that configuration. This does require a bit of overhead, but results in a much simpler model where we don't have to worry about reconciling packaged configuration with install-time configuration. This could always be expanded in the future.
5. Configuring Provider deployments _could_ be deferred to after this initial implementation. The only known use-case right now is with IAM Roles for SA when provider-aws is installed into a Crossplane cluster running on EKS, and we are currently planning on keeping the `ClusterPackageInstall` around for a few releases, so `provider-aws` could still be installed to work in that way. Also, though it would be a degraded experience, a Provider can always be installed manually without the package manager.

With these in mind, I think I would support starting without configurability, then talking to some users and deciding the best way to support it. Alternatively, we could allow a full `DeploymentSpec` in the `Provider` install Spec, then just consider or ignore that based on how the package manager is configured, but I would prefer we bias towards not adding fields until we know how we want them to be used. The cost of adding is much less than the cost of removing.

How does all of this sound? /cc @lukeweber @negz

pkg/controller/pkg/revision/deployment.go
```diff
@@ -0,0 +1,73 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/crossplane/crossplane-runtime/pkg/meta"
+	metav1alpha1 "github.com/crossplane/crossplane/apis/pkg/meta/v1alpha1"
+	"github.com/crossplane/crossplane/apis/pkg/v1alpha1"
+)
+
+func buildProviderDeployment(provider *metav1alpha1.Provider, revision v1alpha1.PackageRevision, namespace string) (*corev1.ServiceAccount, *appsv1.Deployment) {
+	s := &corev1.ServiceAccount{
```
hasheddan, line 30: Note: will need to be named in accordance with the RBAC design in #1706

pkg/controller/pkg/manager/reconciler.go
```diff
@@ -0,0 +1,428 @@
...
+	// is that the job will restart a failed pod automatically. However, if the
+	// job completes and the pod is then deleted, the job will not automatically
+	// recreate it. We handle this case below.
+	j := &batchv1.Job{}
+
+	// NOTE(hasheddan): we have to create the job to get the image ID of the
+	// package. If we move to parsing the image manifest from registry directly
+	// then we can instead create the package revision and let it manage the
+	// full lifecycle of its job.
+	jobName := strings.Replace(strings.Replace(p.GetSource(), "/", "-", -1), ":", "-", -1)
```
hasheddan, line 236: This is a bit gross but gets the job done. In other places we have deferred to naming the "thing" associated with a package revision the hash of its image. However, we can't do that here because we don't know the hash yet. If we adjust the names of things like a Provider deployment or SA, would be good to follow that convention here.

hasheddan: (Obviously only to the extent that we can without using the hash)

pkg/controller/pkg/revision/reconciler.go
```diff
@@ -0,0 +1,565 @@
...
+
+	clientset, err := kubernetes.NewForConfig(cfg)
+	if err != nil {
+		return errors.Wrap(err, "failed to initialize host clientset with in cluster config")
+	}
+
+	installNamespace := os.Getenv(envPodNamespace)
+	if installNamespace == "" {
+		return errors.New("cannot initialize controller with empty install namespace")
+	}
```
negz, line 178: Would it be possible to take this as a CLI flag rather than an environment variable? I feel that flags are somewhat more discoverable, and kingpin lets us set them from an env var so they're sort of the best of both worlds. They're slightly annoying to plumb down, but there is precedent; https://github.com/crossplane/crossplane/blob/0b4227e/cmd/crossplane/package/manage/manage.go#L95

hasheddan: @negz yep! Works for me. I went with env var for testing purposes, but if we are plumbing it down then that works even better 👍

pkg/controller/pkg/revision/reconciler.go
```diff
@@ -0,0 +1,565 @@
...
+
+	cfg, err := config.GetConfig()
+	if err != nil {
+		return errors.Wrap(err, "failed to initialize host config with in cluster config")
+	}
+
+	clientset, err := kubernetes.NewForConfig(cfg)
+	if err != nil {
+		return errors.Wrap(err, "failed to initialize host clientset with in cluster config")
+	}
```
negz, line 173: Do we need to pass a clientset to `NewReconciler`, or could it build one using `mgr.GetConfig()`?

pkg/controller/pkg/revision/reconciler.go
```diff
@@ -0,0 +1,565 @@
...
+			log.Debug(errDeleteProviderDeployment, "error", err)
+			pr.SetConditions(runtimev1alpha1.Unavailable(), runtimev1alpha1.ReconcileSuccess())
+			return reconcile.Result{RequeueAfter: shortWait}, errors.Wrap(r.client.Status().Update(ctx, pr), errUpdateStatus)
+		}
+		if err := r.client.Delete(ctx, s); err != nil {
+			log.Debug(errDeleteProviderSA, "error", err)
+			pr.SetConditions(runtimev1alpha1.Unavailable(), runtimev1alpha1.ReconcileSuccess())
+			return reconcile.Result{RequeueAfter: shortWait}, errors.Wrap(r.client.Status().Update(ctx, pr), errUpdateStatus)
+		}
+	}
```
negz, line 335: I forget whether we discussed during design what would happen to managed resources that are 'orphaned' when a provider revision becomes inactive. I recall that their CRDs and CRs would continue to exist and be owned by an inactive package revision, but I suspect we'd be unable to delete any dangling CRs without a controller running to reconcile them since nothing would remove the finalizers.

pkg/controller/pkg/revision/reconciler.go
```diff
@@ -0,0 +1,565 @@
...
+
+	// NOTE(hasheddan): the following logic enforces the present guarantees of
+	// the package manager, which includes:
+	// 1. We will not create or update any
+	// objects unless we can create or update all objects 2. We will not create
+	// or update any objects unless we can gain control of all objects 3. We
+	// will only modify owner references when the package revision is inactive
+	// 4. If objects are reconciled by a packaged controller, we will not remove
+	// owner references until the controller has been stopped
```
negz, line 344: Nit: Looks like this list might have been wrapped.

negz (commented):
I didn't manage to make it all the way through this PR tonight, so I'll publish a partial review for now. It's looking good so far, though I'd like it if it was possible to lower the length and complexity of the reconcile loops by breaking some of the logic out into composable pieces.

pkg/controller/pkg/manager/reconciler.go
hasheddan: @negz yep, but we would also collide at the `_____Revision` level, so likely okay to go ahead and hit that here.

apis/pkg/v1alpha1/provider_types.go
hasheddan: > While I agree that folks could work around the issue by manually deploying the provider, I suspect it would be a better UX if they didn't have to.

yes, we definitely need to make this a good experience, but it seems that there are some barriers to doing so right now that may take time to work out in a long-term sustainable way. One other thing to mention here as a use-case is that the `imagePullPolicy` and `imagePullSecrets` also flow down to the `Deployment` right now. This allows you to use an image from a private registry for the controller `Deployment`. The uses of this are likely fairly limited right now because this is only applicable for Providers and I am not sure how many proprietary Providers are in existence. However, we could allow this to continue to flow down for now, and just document the behavior change in the future.

apis/pkg/v1alpha1/configuration_types.go
```diff
@@ -0,0 +1,114 @@
...
+	RevisionHistoryLimit *int64 `json:"revisionHistoryLimit,omitempty"`
+
+	// ImagePullSecrets are named secrets in the same workspace that can be used
+	// to fetch Configurations from private repositories.
+	ImagePullSecrets []corev1.LocalObjectReference `json:"imagePullSecrets,omitempty"`
+
+	// ImagePullPolicy defines the pull policy for all images used during
+	// Configuration extraction.
+	// https://kubernetes.io/docs/concepts/configuration/overview/#container-images
+	ImagePullPolicy corev1.PullPolicy `json:"imagePullPolicy,omitempty"`
```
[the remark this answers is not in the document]
hasheddan: That sounds fine to me. It makes it a bit more abstract for the user, but since we are already intentionally introducing abstraction, I would be fine to continue that trend with these fields.

apis/pkg/v1alpha1/revision_types.go
```diff
@@ -0,0 +1,124 @@
...
+	// and a semantic version range. Either Package, CustomResourceDefinition,
+	// or Crossplane can be specified.
+	Package string `json:"package,omitempty"`
+
+	// CustomResourceDefinition is the full name of a CRD that is owned by the
+	// package being requested. This can be a convenient way of installing a
+	// package when the desired CRD is known, but the package name that contains
+	// it is not known. Either Package, CustomResourceDefinition, or Crossplane
+	// can be specified.
+	CustomResourceDefinition string `json:"crd,omitempty"`
```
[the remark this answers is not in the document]
hasheddan: I would be in favor of omitting until this functionality is actually available 👍

apis/pkg/v1alpha1/revision_types.go
```diff
@@ -0,0 +1,124 @@
...
+
+	// DependsOn is the list of packages and CRDs that this package depends on.
+	DependsOn []Dependency `json:"dependsOn,omitempty"`
+
+	// ControllerOptions is the options applied to packaged controller deployment.
+	ControllerOptions *ControllerOptions `json:"controllerOptions,omitempty"`
+
+	// Permissions are additional permissions that should be added to a packaged
+	// controller's service account.
+	Permissions *PermissionsSpec `json:"permissionSpec,omitempty"`
```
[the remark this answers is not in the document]
hasheddan: This is copied over from existing package types, and I think you are right that they should be `permissionRequests`. In practice, the RBAC controller would look at these and add them to the role it creates for the SA. Since they are part of the `Spec`, I would expect them to be populated by the `Package` controller, rather than the `PackageRevision` controller, but that involves parsing the package contents in the `Package` controller. How would you feel about this being moved to the status of `___Revision` like we did with Objects?

pkg/controller/packages/hosted/sync.go
```diff
@@ -59,10 +60,9 @@ func SyncImagePullSecrets(ctx context.Context, tenantKube, hostKube client.Clien
 	name := hostObj.GetName()
 	hostNS := hostObj.GetNamespace()
 
-	ref := &corev1.ObjectReference{
+	ref := &v1alpha1.TypedReference{
 		APIVersion: v,
 		Kind:       k,
-		Namespace:  hostNS,
```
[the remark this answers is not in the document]
hasheddan: I am going to move this out into a separate PR to reduce noise here 👍

pkg/controller/pkg/manager/reconciler.go
```diff
@@ -0,0 +1,428 @@
...
+	reconcileTimeout = 1 * time.Minute
+
+	shortWait     = 30 * time.Second
+	veryShortWait = 5 * time.Second
+
+	envPodNamespace = "POD_NAMESPACE"
+)
+
+// variableWait selects a shorter wait time if passed condition is true.
+func variableWait(short bool) time.Duration {
```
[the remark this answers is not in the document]
hasheddan: SGTM, I was attempting to achieve this with naming the parameter `short`, but I agree the name of the function is better practice. I wonder if a general function that took a boolean expression and two `time.Duration` would be useful across controllers 🤔

pkg/controller/pkg/manager/reconciler.go
```diff
@@ -0,0 +1,428 @@
...
+		log.Debug(errListRevisions, "error", err)
+		return reconcile.Result{}, errors.Wrap(resource.IgnoreNotFound(err), errListRevisions)
+	}
+
+	// Identify or create an install job.
+	// NOTE(hasheddan): the only advantage to using a job instead of just a pod
+	// is that the job will restart a failed pod automatically. However, if the
+	// job completes and the pod is then deleted, the job will not automatically
+	// recreate it. We handle this case below.
+	j := &batchv1.Job{}
```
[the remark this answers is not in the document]
hasheddan: I like the idea of just using a pod (hence this comment), but we do lose the functionality described in this comment.

pkg/controller/pkg/manager/reconciler.go
```diff
@@ -0,0 +1,428 @@
...
+
+// NewReconciler creates a new package reconciler.
+func NewReconciler(mgr ctrl.Manager, opts ...ReconcilerOption) *Reconciler {
+	r := &Reconciler{
+		client: resource.ClientApplicator{
+			Client:     mgr.GetClient(),
+			Applicator: resource.NewAPIPatchingApplicator(mgr.GetClient()),
+		},
+		log:    logging.NewNopLogger(),
+		record: event.NewNopRecorder(),
```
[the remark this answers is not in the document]
hasheddan: Definitely, meant to include this in `TODO` along with more descriptive conditions. I figured these were somewhat trivial (relative to things like the API of the types and the controller design), so I deferred them to after initial review.

pkg/controller/pkg/manager/reconciler.go
```diff
@@ -0,0 +1,428 @@
...
+	jobName := strings.Replace(strings.Replace(p.GetSource(), "/", "-", -1), ":", "-", -1)
+	if err := r.client.Get(ctx, types.NamespacedName{Name: jobName, Namespace: r.namespace}, j); err != nil {
+		if kerrors.IsNotFound(err) {
+			j = buildParserJob(jobName, r.namespace, p.GetSource(), map[string]string{parentLabel: p.GetName()}, p.GetImagePullSecrets(), p.GetImagePullPolicy())
+			// NOTE(hasheddan): install jobs are owned by the
+			meta.AddOwnerReference(j, meta.AsOwner(meta.TypedReferenceTo(p, p.GetObjectKind().GroupVersionKind())))
+			if err := r.client.Create(ctx, j); err != nil {
+				log.Debug(errCreateParserJob, "error", err)
+			}
+			p.SetConditions(runtimev1alpha1.Unavailable(), runtimev1alpha1.ReconcileSuccess())
```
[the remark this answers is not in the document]
hasheddan: Likely yes, but I think custom conditions may be in order here.

pkg/controller/pkg/manager/reconciler.go
```diff
@@ -0,0 +1,428 @@
...
+		if err := r.client.Delete(ctx, j, client.PropagationPolicy(metav1.DeletePropagationForeground)); err != nil {
+			log.Debug(errDeleteParserJob, "error", errors.New("failed to delete job that needs to restart"))
+		}
+		p.SetConditions(runtimev1alpha1.Unavailable(), runtimev1alpha1.ReconcileSuccess())
+		return reconcile.Result{RequeueAfter: veryShortWait}, errors.Wrap(r.client.Status().Update(ctx, p), errUpdateStatus)
+	}
+
+	// If length of pod list is not 1, we are likely waiting for job to create pod.
+	if len(podList.Items) != 1 {
+		log.Debug(errGetParserPod, "error", errors.New("wrong number of pods for job"))
```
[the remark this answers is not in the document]
hasheddan: It is not, so would likely be good to keep this as a debug message but not as an error

pkg/controller/pkg/manager/reconciler.go
```diff
@@ -0,0 +1,428 @@
...
+		// longer wait as manual intervention is required.
+		return reconcile.Result{RequeueAfter: variableWait(len(pod.Status.ContainerStatuses) == 0)}, errors.Wrap(r.client.Status().Update(ctx, p), errUpdateStatus)
+	}
+
+	imageIDSlice := strings.Split(pod.Status.ContainerStatuses[0].ImageID, ":")
+	hash := imageIDSlice[len(imageIDSlice)-1]
+
+	// Do not create a revision if the image ID has not yet been populated.
+	if hash == "" {
+		log.Debug(errGetParserContainer, "error", errors.New("invalid image hash for container in parser pod"))
```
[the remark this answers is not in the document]
hasheddan: Same as comment above 👍

pkg/controller/pkg/revision/deployment.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+	d := &appsv1.Deployment{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:            revision.GetName(),
+			Namespace:       namespace,
+			OwnerReferences: []metav1.OwnerReference{meta.AsController(meta.TypedReferenceTo(revision, v1alpha1.ProviderRevisionGroupVersionKind))},
+		},
+		Spec: appsv1.DeploymentSpec{
+			Selector: &metav1.LabelSelector{
+				MatchLabels: map[string]string{"revision.pkg.crossplane.io": "true"},
```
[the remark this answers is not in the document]
hasheddan: SGTM

pkg/controller/pkg/manager/reconciler.go
```diff
@@ -0,0 +1,428 @@
...
+	}
+	p.SetCurrentRevision(hash)
+
+	pr := r.newPackageRevision()
+	maxRevision := int64(0)
+	oldestRevision := int64(10)
+	oldestRevisionIndex := -1
+	revisions := prs.GetRevisions()
+
+	// Check to see if revision already exists.
```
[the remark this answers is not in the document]
hasheddan: I'll look at doing so. However, I was a bit hesitant as the existing package manager is broken out into lots of smaller pieces and it personally feels hard to follow for me. I could see potentially breaking out some of the garbage collection and job / pod creation. I could also see having separate validation and create / update functions in the revision controller, similar to how we pass Applicators to our reconcilers. Let me refactor a bit and let you take another look 👍

pkg/controller/pkg/revision/reconciler.go
hasheddan: So as long as the revision exists, its objects will also exist (the inactive revision is now an Owner instead of Controller). You are correct that in order to clean up CRs after that, you would need to manually remove the finalizer. However, you could also just delete the revision, in which case the CRD would be GC'ed by k8s (no more owner refs), and all CRs for that CRD will also be cleaned up.

I think this is a good pattern as it makes use of kubernetes inherent functionality, rather than adding a bunch of logic on top. The one case where this could be difficult is when you update a package and then want to go and delete some but not all of the CRs for CRDs only owned by the old package revision (this would be a strange case because you would be electing to keep around CRs that were not being reconciled). In this case, I think we should be okay saying that it takes some manual work, and we could even add some utilities to `crank` to help with removing finalizers for things owned by an inactive package revision.

apis/pkg/v1alpha1/revision_types.go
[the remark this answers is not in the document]
negz: Regardless of what controller sets the requests, they do feel a little more like the desired state of the system to me. I can see why we might put them in status but I lean slightly toward keeping them in spec.

pkg/controller/pkg/manager/reconciler.go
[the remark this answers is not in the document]
negz: Yeah. I didn’t think very hard about it, but I’m wondering if it would be simpler for us overall to just handle that functionality ourselves.

pkg/controller/pkg/manager/reconciler.go
negz: Would that be true even if a config and a package both tried to use the same image? It could be that that’s an obscure enough edge case that we shouldn’t bother with it.

pkg/controller/pkg/manager/reconciler.go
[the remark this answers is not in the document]
negz: That sounds like a good approach. It’s definitely tough to strike the balance between making the logic composeable and keeping it east to follow.

pkg/controller/pkg/revision/reconciler.go
negz: I agree with the approach you propose, but are you sure that deleting a CRD will delete the CRs it defines, even if they have finalizers that never get removed?

pkg/controller/pkg/manager/reconciler.go
hasheddan: So there are a few cases here (I think no. 4 is what you are getting at):
1. Two providers use the same image name/tag with the same contents (hash). Right now they would collide on the `Job` and the second would be unable to install. If we made the `Job` unique, they would not collide, but would collides when the second tried to create the `ProviderRevision.
2. Two providers use the same image name/tag with different contents (hash). Right now they would collide on the `Job` and the second would be unable to install. If we made the `Job` unique, they would not collide and the `ProviderRevision` would not collide, making the second able to be installed as well (assuming no conflicting objects). However, this could be a somewhat confusing scenario and I am not sure we want to support it.
3. A provider and configuration use the same image name/tag with the same contents (hash). Right now they would collide on `Job`. If we made the `Job` unique, they would not collide until they got to installing objects. One of them would have to have a mismatch with the meta package type in the package, which would cause it to fail to install.
4. A provider and configuration use the same image name/tag with different contents (hash). Right now they would collide on `Job`. If we made `Job` unique, they would not collide, and they would not collide on `___Revision` or when installing objects (assuming the contents fundamentally changed the package type). This would technically be a valid use-case, but I once again am skeptical that we should support it. Using image hashes allow us to accommodate situations in which a package's contents change for the image with the same tag, and a user wants to update to using the new image contents (this could be done by deleting the existing `Job`). However, I don't think this is going to be a common use-case, and I don't think we should further complicate the model by allowing the use of the same image with different contents outside the scope of a single `Provider` or `Configuration`.

apis/pkg/v1alpha1/revision_types.go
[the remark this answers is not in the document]
hasheddan: The issue with putting them in `spec` is that if both the `Package` and `PackageRevision` controllers are watching the `PackageRevision`, then the `PackageRevision` controller updating the `spec` will trigger the the `Package` controller, which will then trigger the `PackageRevision` controller.... and so on. We are currently using the `APIPatchingApplicator`, and we could just say "if certain fields match, don't do a patch", but I would prefer if we were able to make it such that the `PackageRevision` controller didn't touch the `spec` at all. I do see your point that it feels like these belong in the `spec`, but it kind of depends on how you look at the revision. If the `spec` is indicating how it should parse and install a package, then you could argue that these don't belong in the `spec` since they are the _result_ of parsing the package, rather than indicating _how_ to parse a package.

pkg/controller/packages/hosted/sync.go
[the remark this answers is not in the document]
hasheddan: This has been factored out in https://github.com/crossplane/crossplane/pull/1718

pkg/controller/pkg/revision/reconciler.go
hasheddan: You are correct, the CRD will hang around as long as the CR still has a finalizer, so folks would need to go back and manually remove that finalizer (ref: https://github.com/kubernetes/kubernetes/issues/60538#issuecomment-680279481). This is not straightforward for us to address though. If you leave behind CRs when deleting a Revision (or its parent package), I'm not sure we can assume that you want us to remove arbitrary finalizers that the provider controller may have added to the CRs, and I definitely don't think we should ever explicitly delete CRs. Do you have any thoughts here?

apis/pkg/v1alpha1/provider_types.go
hasheddan: Ping @turkenh. We should discuss what you see is minimum requirements for initial implementation here.

pkg/controller/pkg/manager/job.go
```diff
@@ -0,0 +1,74 @@
...
+
+package manager
+
+import (
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var (
```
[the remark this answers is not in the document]
hasheddan: We have to take their address, so they cannot be `const`

pkg/controller/pkg/revision/reconciler.go
hasheddan: This was preemptive as we are going to need to manipulate building the clientset for host-aware vs regular mode. We _could_ pass information to `NewReconciler` to do that for us, but not sure that would be much of an optimization.

pkg/controller/pkg/revision/deployment.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+	d := &appsv1.Deployment{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:            revision.GetName(),
+			Namespace:       namespace,
+			OwnerReferences: []metav1.OwnerReference{meta.AsController(meta.TypedReferenceTo(revision, v1alpha1.ProviderRevisionGroupVersionKind))},
+		},
+		Spec: appsv1.DeploymentSpec{
+			Selector: &metav1.LabelSelector{
+				MatchLabels: map[string]string{"pkg.crossplane.io/revision": revision.GetName()},
```
hasheddan, line 47: Very annoyingly, the limit on a k8s label value is 63 chars, while a sha256 is 64 characters :roll_eyes:

hasheddan:
@negz I am going to attempt to summarize the updates I have made since your last review to ease the burden of you next review pass:
- Switched to using `Pod` directly instead of `Job` for unpacking. The package controller acts sort of like an opinionated `Job` in that it will delete and recreate a `Pod` any time it enters an unhealthy state. I was trying to think of if we should always try to restart or if in some cases we should just say we tried and failed. I landed on always trying to restart but with a not very aggressive requeue. That way we shouldn't thrash too hard, but I suspect someone will either take action to update the package to a different version or take manual steps in the cluster to fix the install issue, so we might as well keep trying until we are told not to.
- Removed all fields for configuring a provider's controller for now from the `PackageRevisionSpec`
- Moved all fields that come from actually parsing a package to the status of the package revision to establish clear separation of responsibility between package controller and package revision controller (this include permissions and dependencies)
- Added custom conditions. I kind of struggled with what to use for conditions, but they are currently as follows:
  - `Installed`: this only applies to package types (i.e. `Provider` / `Configuration`). It is true if there the current `PackageRevision` is `Active`
  - `Healthy`: this applies to both package and package revision types. It is true for a package revision if all steps of installation for a package revision complete successfully. It is true for a package if the current package revision is healthy.
- Added events for errors and successful installation

There are still some unresolved comments above that I think we need to discuss a bit more. Outside of that, the following things remain:
- Finishing up unit tests (mostly waiting on this until the implementation is signed off on to avoid churn)
- Any factoring of logic into smaller functions. This will likely go hand-in-hand with finishing unit tests. It also could be partially addressed then followed up in a subsequent PR since it doesn't directly affect functionality.
- Rebase and clean up commits

Let me know if there are any other steps to make this PR a bit more manageable to review :)

apis/pkg/v1alpha1/provider_types.go
```diff
@@ -0,0 +1,116 @@
...
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   ProviderSpec   `json:"spec,omitempty"`
+	Status ProviderStatus `json:"status,omitempty"`
+}
+
+// ProviderSpec specifies details about a request to install a provider to
+// Crossplane.
+type ProviderSpec struct {
```
hasheddan, line 47: With `ControllerOptions` removed for now, I think we can move `ProviderSpec` and `ConfigurationSpec` to `PackageSpec` (similar to how we do with `PackageRevisionSpec`) and that can be the base that any future package type has to support (at minimum).

pkg/controller/pkg/manager/reconciler.go
```diff
@@ -0,0 +1,401 @@
...
+	// NOTE(hasheddan): we have to create the pod to get the image ID of the
+	// package. If we move to parsing the image manifest from registry directly
+	// then we can instead create the package revision and let it manage the
+	// full lifecycle of its pod.
+	podName := strings.NewReplacer("/", "-", ":", "-").Replace(p.GetSource())
+	if err := r.client.Get(ctx, types.NamespacedName{Name: podName, Namespace: r.namespace}, pod); err != nil {
+		if kerrors.IsNotFound(err) {
+			pod = buildInstallPod(podName, r.namespace, p.GetSource(), map[string]string{parentLabel: p.GetName()}, p.GetPackagePullSecrets(), p.GetPackagePullPolicy())
+			meta.AddOwnerReference(pod, meta.AsOwner(meta.TypedReferenceTo(p, p.GetObjectKind().GroupVersionKind())))
+			if err := r.client.Create(ctx, pod); err != nil {
```
hasheddan, line 223: @kasey brought up a good question this morning about gc of completed Pods, which I wanted to add some context around...

`Succeeded` pods will be garbage collected according to the [`terminated-pod-gc-threshold` flag](https://kubernetes.cn/docs/concepts/workloads/pods/pod-lifecycle/#pod-garbage-collection) on the `kube-controller-manager`, which has a [default value of 12500](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/podgc/config/v1alpha1/defaults.go#L34). This means that (with this current implementation), if an install Pod gets garbage collected (or manually deleted for that matter), the package revision will report that it is unhealthy, which will trigger a reconcile from the package controller, which will see that the Pod does not exist, and will recreate it. The package revision controller will then be successful on its next read of the Pod logs, will update to healthy, and that will be updated in the package as well.

This is assuming the current package revision though. If an _old_ package revision pod gets garbage collected, that revision will become unhealthy. This isn't a _huge_ problem, but it is a problem because an old current revision does still check that it has owner references on all of its installed objects, so it will fail to do that and the package controller will not recreate its Pod.

One way to get around this is to read the list of `Objects` we have on the revision (that get populated after reading Pod logs and modifying the objects), and if the package revision is inactive we can just use those to set the owner references (we are just adding them to the current object in the API server, so we don't need to know what it looks like in the package).

That being said, besides the old package revision being unhealthy, the fact that it is unhealthy is likely to not have effect on the system because nothing should be removing the owner references it has on an object, so it won't be updating them, but the object should still be garbage collected when the revision is deleted.

But don't worry, there's more complexity! If someone was to do a backup & restore of their cluster using something like Velero, the UID of the old package revision could change, and if it was the only owner reference on an object, that object would be GC'ed because it would see that none of its owners exist, and the Pod would not be created for the old package revision so it would not be able to go and update the owner references.

> aside: if there is no finalizer on the objects created by the package revision controller, could they be garbage collected on a backup and restore in the time it takes for the owner / controller references to get updated, even for the current package revision?

One thing that this all reaffirms in my mind is that storing objects in pod logs after initial unpack is a less than ideal situation. I hope that we can gain traction in a future iteration to unpack packages and either store them in volumes or some other medium.

hasheddan: In regards to that last comment, another thing I would be in support of is getting the hash in the package controller (using something like `go-containerregistry`), but then deferring all unpacking and management of pods to the package revision controller.

pkg/controller/pkg/manager/reconciler.go
negz: If we go with the `providername-abbreviatedhash` naming scheme, will that take care of all of these cases?

apis/pkg/v1alpha1/configuration_types.go
```diff
@@ -0,0 +1,114 @@
...
+
+	Spec   ConfigurationSpec   `json:"spec,omitempty"`
+	Status ConfigurationStatus `json:"status,omitempty"`
+}
+
+// ConfigurationSpec specifies details about a request to install a
+// configuration to Crossplane.
+type ConfigurationSpec struct {
+	// Package is the name of the package that is being requested, e.g.,
+	// myapp. Either Package or CustomResourceDefinition can be specified.
```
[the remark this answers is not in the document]
negz: This was marked resolved, but no change was made. Was that intentional?

apis/pkg/v1alpha1/provider_types.go
negz: I believe after discussion with @turkenh we're good to proceed without exposing any options to configure the provider (for now), right?

apis/pkg/v1alpha1/revision_types.go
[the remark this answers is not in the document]
negz: > it kind of depends on how you look at the revision.

Yeah, on reflection I think your approach of putting these in status makes sense. If I think of "`spec.installPodRef` as being the desired state - or a pointer to the actual source of desired state (the install pod's logs) - then modelling all these as actual state in status makes sense.

apis/pkg/v1alpha1/configuration_types.go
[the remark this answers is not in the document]
hasheddan: Ah my apologies! Forgot to push those changes 👍

apis/pkg/v1alpha1/provider_types.go
hasheddan: Correct!

pkg/controller/pkg/manager/reconciler.go
hasheddan: Yes 👍

negz (commented):
Thanks @hasheddan! I've left some comments, but in general I think this PR is ready to be cleaned up for final approve and merge. I'd still like to make the reconcile logic a little more composable, but I think the priority at this stage should be adding tests and getting this merged for us to iterate on. Please make executive decisions as to whether you address my comments, choose to disregard them, or choose to defer them for future (though in the latter case please be sure to add a TODO or raise an issue).

While you wrap this PR up, would you mind opening a separate PR that adds the new package types? I think they're in pretty good shape, and getting them into master would make it a little easier to add provider support to the RBAC manager.

pkg/controller/pkg/manager/pod.go
```diff
@@ -0,0 +1,68 @@
...
+				{
+					Name:            "unpack",
+					Image:           img,
+					ImagePullPolicy: pullPolicy,
+					SecurityContext: &corev1.SecurityContext{
+						RunAsUser:                &runAsUser,
+						RunAsGroup:               &runAsGroup,
+						AllowPrivilegeEscalation: &allowPrivilegeEscalation,
+						Privileged:               &privileged,
+						RunAsNonRoot:             &runAsNonRoot,
```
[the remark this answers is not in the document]
hasheddan: We do not need to. A container will inherit from Pod if it is not set explicitly on the container. However, not all fields are available on the Pod, such as `Privileged` and `AllowPrivilegeEscalation`. If we wanted to strip some out here we should remove them from the Pod security context. I actually kind of like having them explicitly defined at both levels though because if someone is adjusting them in the future it feels like a stronger check if they have to change a value rather than add something net new (once again, we would have them defined at the container level and adding something to Pod security context would not override, so not super risky). However, this is mostly just a philosophical thing, so happy to simplify if desired.

pkg/controller/pkg/manager/reconciler.go
```diff
@@ -0,0 +1,402 @@
...
+		pr.SetRevision(maxRevision + 1)
+	}
+
+	// Check to see if there are revisions eligible for garbage collection.
+	if len(revisions) > (int(p.GetRevisionHistoryLimit())+1) && p.GetRevisionHistoryLimit() != 0 {
+		gcRev := revisions[oldestRevisionIndex]
+		// Find the oldest revision and delete it.
+		if err := r.client.Delete(ctx, gcRev); err != nil {
+			log.Debug(errGCPackageRevision, "error", err)
+			return reconcile.Result{RequeueAfter: shortWait}, errors.Wrap(r.client.Status().Update(ctx, p), errUpdateStatus)
```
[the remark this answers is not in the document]
hasheddan: That's correct

pkg/controller/pkg/manager/reconciler.go
```diff
@@ -0,0 +1,402 @@
...
+		return reconcile.Result{RequeueAfter: shortWait}, errors.Wrap(r.client.Status().Update(ctx, p), errUpdateStatus)
+	}
+	if err := r.client.Applicator.Apply(ctx, pr, resource.MustBeControllableBy(p.GetUID())); err != nil {
+		log.Debug(errCreatePackageRevision, "error", err)
+		return reconcile.Result{RequeueAfter: shortWait}, errors.Wrap(r.client.Status().Update(ctx, p), errUpdateStatus)
+	}
+
+	// Set package to unavailable until package revision is available. Reconcile
+	// will be queued on changes to revisions, so we do not need to explicitly
+	// requeue here.
```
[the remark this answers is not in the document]
hasheddan: We previously were, although now we are not setting the `Healthy` status at all and just waiting to set it until the revision reports its health. Seems like having a blank `Healthy` status is better than showing `Unhealthy` before the revision reports back, but am open to suggestions here. Also, I'll update this comment.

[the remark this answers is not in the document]
hasheddan: One caveat here: if it is not the first revision, then we will have the `Healthy` status of the old revision until the new one reports back.

pkg/controller/pkg/revision/reconciler.go
```diff
@@ -0,0 +1,600 @@
...
+				pr.SetConditions(v1alpha1.Unhealthy())
+				return reconcile.Result{RequeueAfter: shortWait}, errors.Wrap(r.client.Status().Update(ctx, pr), errUpdateStatus)
+			}
+			if err := r.client.Create(ctx, d, client.DryRunAll); err != nil {
+				log.Debug(errCreateDryRunObj, "error", err)
+				r.record.Event(pr, event.Warning(errCreateDryRunObj, err))
+				pr.SetConditions(v1alpha1.Unhealthy())
+				return reconcile.Result{RequeueAfter: shortWait}, errors.Wrap(r.client.Status().Update(ctx, pr), errUpdateStatus)
+			}
+			continue
```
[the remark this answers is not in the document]
hasheddan: Yep! Although I assume we would want to check the error on the `meta.AsController`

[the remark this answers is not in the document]
hasheddan: Ah nevermind, I see that is only for `AddControllerReference`. This works well for `Create` :+1:

pkg/controller/pkg/revision/reconciler.go
```diff
@@ -0,0 +1,600 @@
...
+		if !ok {
+			return nil, errors.New("cannot assert object with provider GVK to provider type")
+		}
+		return provider, nil
+	}
+	if pkgMetaGVK == pkgmeta.ConfigurationGroupVersionKind {
+		return nil, nil
+	}
+	return nil, errors.New("unrecognized package meta type")
+}
```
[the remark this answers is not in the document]
hasheddan: This doesn't check that there aren't multiple package meta defined, which I think we should do. However, I am adding in linting so this function will go away anyway.

pkg/controller/pkg/revision/reconciler.go
```diff
@@ -0,0 +1,600 @@
...
+
+// BuildMetaScheme builds the default scheme used for identifying metadata in a
+// Crossplane package.
+func BuildMetaScheme() (*runtime.Scheme, error) {
+	metaScheme := runtime.NewScheme()
+	if err := pkgmeta.SchemeBuilder.AddToScheme(metaScheme); err != nil {
+		return nil, err
+	}
+	return metaScheme, nil
+}
```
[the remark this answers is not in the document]
hasheddan: I think I would prefer to bubble them up to the entrypoint, which would eventually call `FatalIfError` I believe.

pkg/controller/pkg/manager/pod.go
```diff
@@ -0,0 +1,68 @@
...
+package manager
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var (
+	runAsUser                = int64(2000)
+	runAsGroup               = int64(2000)
```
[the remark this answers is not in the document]
hasheddan: Only that they are typically not privileged. This can be helpful in the case that someone does not set their UID / GID in their container image, so it defaults to root when they don't really need it

pkg/controller/pkg/manager/reconciler.go
```diff
@@ -0,0 +1,402 @@
...
+
+	p := r.newPackage()
+	if err := r.client.Get(ctx, req.NamespacedName, p); err != nil {
+		// There's no need to requeue if we no longer exist. Otherwise we'll be
+		// requeued implicitly because we return an error.
+		log.Debug(errGetPackage, "error", err)
+		return reconcile.Result{}, errors.Wrap(resource.IgnoreNotFound(err), errGetPackage)
+	}
+
+	// Get existing package revisions.
```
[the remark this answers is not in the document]
hasheddan: Yep, I believe we can just defer to the GC here.

pkg/controller/pkg/manager/reconciler.go
```diff
@@ -0,0 +1,402 @@
...
+	prs := r.newPackageRevisionList()
+	if err := r.client.List(ctx, prs, client.MatchingLabels(map[string]string{parentLabel: p.GetName()})); err != nil {
+		log.Debug(errListRevisions, "error", err)
+		return reconcile.Result{}, errors.Wrap(resource.IgnoreNotFound(err), errListRevisions)
```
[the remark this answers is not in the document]
hasheddan: I believe `kmeta.IsNoMatchError` is returned if the API types is not able to be recognized by the client. The doc on `List` says:
```
// List retrieves list of objects for a given namespace and list options. On a
// successful call, Items field in the list will be populated with the
// result returned from the server.
``` 
Which sound to me like it should return an error if none are found. However, I actually don't know how what I have right here would be working if that was the case because we would never get past this point because we would start with no package revisions, and then we would never create any :thinking:

pkg/controller/pkg/revision/reconciler.go
hasheddan: I really like this idea. I'll get it written up in an issue for tracking :+1:

hasheddan: Opened https://github.com/crossplane/crossplane/issues/1732 to track

pkg/controller/pkg/revision/linter.go
```diff
@@ -0,0 +1,166 @@
...
+const (
+	errMultipleMeta         = "not exactly one package meta type"
+	errNotMetaProvider      = "package meta type is not Provider"
+	errNotMetaConfiguration = "package meta type is not Configuration"
+	errNotCRD               = "object is not a CRD"
+	errNotXRD               = "object is not an XRD"
+	errNotComposition       = "object is not a Composition"
+)
+
+// A Linter lints packages.
```
hasheddan, line 41: I anticipate moving much of this (excluding the functions specific to Crossplane types) to `crossplane-runtime` at some point

pkg/controller/pkg/revision/establisher.go
```diff
@@ -0,0 +1,178 @@
...
+			if err := e.create(ctx, d, parent, control, client.DryRunAll); err != nil {
+				return err
+			}
+			continue
+		}
+
+		c, ok := current.(resource.Object)
+		if !ok {
+			return errors.New(errAssertObj)
+		}
```
negz, line 114: I'm pretty sure this branch should never be hit, since we check at line 83.

pkg/controller/pkg/revision/establisher.go
```diff
@@ -0,0 +1,178 @@
...
+
+		if err := e.update(ctx, c, d, parent, control, client.DryRunAll); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// Establish establishes control or ownership of resources in the package.
```
negz, line 131: Could be worth clarifying that the caller must call Check before calling Establish.

pkg/controller/pkg/revision/reconciler.go
```diff
@@ -0,0 +1,398 @@
...
+	}
+	return objScheme, nil
+}
+
+// Reconciler reconciles packages.
+type Reconciler struct {
+	namespace    string
+	client       resource.ClientApplicator
+	establisher  Establisher
+	podLogClient *kubernetes.Clientset
```
negz, line 160: Nit: Could this be a `kubernetes.Interface`?

pkg/controller/pkg/revision/reconciler.go
negz: @hasheddan I recall you mentioned you were on board with this approach. Are you planning to address that in this PR?

pkg/controller/pkg/revision/reconciler.go
```diff
@@ -0,0 +1,398 @@
...
+			return reconcile.Result{RequeueAfter: shortWait}, errors.Wrap(r.client.Status().Update(ctx, pr), errUpdateStatus)
+		}
+		pr.SetControllerReference(&runtimev1alpha1.Reference{Name: d.GetName()})
+	}
+
+	// Update object list in package revision status and set ready condition to
+	// true.
+	pr.SetObjects(r.establisher.GetResourceRefs())
+	// TODO(hasheddan): set remaining status fields (dependencies, crossplane
+	// version, permission requests)
```
negz, line 394: Does this TODO need addressing before we merge, or do you want to handle it in a future PR?

negz (approved):
🎉

pkg/controller/pkg/revision/reconciler.go
hasheddan: Future PR :+1: Related to how RBAC controller will create roles for controllers

pkg/controller/pkg/manager/reconciler_test.go
```diff
@@ -0,0 +1,713 @@
...
+									t.Errorf("-want, +got:\n%s", diff)
+								}
+								return nil
+							}),
+						},
+						Applicator: resource.ApplyFn(func(_ context.Context, _ runtime.Object, _ ...resource.ApplyOption) error {
+							return nil
+						}),
+					},
+					podManager: NewPackagePodManager(&test.MockClient{
```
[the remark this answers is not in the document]
hasheddan: Agreed :+1:

pkg/controller/pkg/revision/reconciler.go
hasheddan: @negz ah I had updated `SetupProviderRevision` but not `SetupConfigurationRevision` to use `mgr.GetConfig` :+1:

pkg/controller/pkg/manager/reconciler_test.go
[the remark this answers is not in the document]
hasheddan: Updated both here and for the package revision establisher :+1:

PR #1696, opened 2020-08-27
Add Package and Configuration meta package configuration types

hasheddan:

Adds `Provider` and `Configuration` meta types. They are to be used to define package configuration and will be parsed by the package manger and any other package tooling. They are NOT installed as CRDs and do not have CRDs generated on their behalf.

Depends on #1681 

I have:

`make generate`


apis/pkg/meta/v1alpha1/provider_types.go
```diff
@@ -0,0 +1,75 @@
...
+	Provider *string `json:"provider,omitempty"`
+
+	// Configuration is the name of a Configuration package image.
+	Configuration *string `json:"configuration,omitempty"`
+
+	// Version is the semantic version of the dependency image.
+	Version string `json:"version"`
+}
+
+// Path is a filesystem path.
```
negz, line 62: Nit: I could imagine this having non-paths to ignore, like perhaps a GVK or regex.

hasheddan: Good point, how about just calling the struct `Ignore`?

Issue #1639, opened 2020-07-08
Composite resources should support bidirectional patching


negz:
### What problem are you facing?

Currently a `Composition` configures how the values of a composite resource's fields should be used to influence composed resources; the composite resource's spec values may be 'overlaid' or 'patched' onto the composed resources. For example the below `Composition` patches the external name annotation from the composite resource (the `KubeCluster`) to the composed resource (the `NodePool`):

```yaml
apiVersion: apiextensions.crossplane.io/v1alpha1
kind: Composition
metadata:
  name: mycomposition
spec:
  from:
    apiVersion: example.org/v1alpha1
    kind: KubeCluster
  to:
  - base:
      apiVersion: container.gcp.crossplane.io/v1alpha3
      kind: NodePool
      spec:
        # ...
  - patches:
    - fromFieldPath: metadata.annotations[crossplane.io/external-name]
      toFieldPath: metadata.annotations[crossplane.io/external-name]
      transforms:
        type: string
        string:
          fmt: "%s-a"
```

This allows a composite resource to control the desired state of its composed resources. Ideally it would also be possible to reflect the actual state of composed resources on the composite resource. For example a `KubeCluster`  resource that composed a `GKECluster` might want to reflect the `status.atProvider.status` (`RUNNING`, `PROVISIONING`, etc) of the `GKECluster` in its own status.

### How could Crossplane help solve your problem?

Crossplane could support bidirectional patching - both from composite to composed resource and from composed to composite resource. Presumably this would require us to rethink the `from` and `to` convention in the `Composition` schema.

This could look something like:

```yaml
apiVersion: apiextensions.crossplane.io/v1alpha1
kind: Composition
metadata:
  name: mycomposition
spec:
  composite:
    apiVersion: example.org/v1alpha1
    kind: KubeCluster
  composed:
  - base:
      apiVersion: container.gcp.crossplane.io/v1beta1
      kind: GKECluster
      spec:
        # ...
  - patches:
    - fromFieldPath: metadata.annotations[crossplane.io/external-name]
      toFieldPath: metadata.annotations[crossplane.io/external-name]
      direction: CompositeToComposed  # Perhaps this is the default?
      transforms:
        type: string
        string:
          fmt: "%s-a"
    - fromFieldPath: status.atProvider.status
      toFieldPath: status.state
      direction: ComposedToComposite
```

muvaf:
I feel like `direction` is too big of a property to let other fields unaffected. To elaborate:
```yaml
[...]

PR #1697, opened 2020-08-30
Make top-level spec fields in Composition more descriptive

### LINKED ISSUE #1685: Composition field name updates
A `Composition` looks like:
```yaml
kind: Composition
metadata:
  name: clusters.cnp.example.org
spec:
  reclaimPolicy: Delete
  compositeResource:
    apiVersion: cnp.example.org/v1alpha1
    kind: AllocatedCluster
  composedResources:
  - base:
      apiVersion: clusters.cnp.example.org/v1alpha1
      kind: EKS
    patches:
    - fromFieldPath: metadata.annotations[crossplane.io/external-name]
      toFieldPath: metadata.annotations[crossplane.io/external-name]
    - fromFieldPath: metadata.uid
      toFieldPath: spec.writeConnectionSecretToRef.name
      transforms:
      - type: string
        string:
          fmt: "%s-eks"
...
```

* `compositeResource`: I think from the first look `compositeResource` could feel like an actual resource but it's actually a type reference, i.e. it has only `apiVersion` and `kind`. So, `compositeKind`, `compositeType`, `targetKind` or `targetType` or even just `type` or `definition`/`targetDefinition` could give more clues about what it is immediately.

* `composedResources`: it could be `resourceTemplates` or just `resources`/`templates` since being editing a `Composition` kind already implies that these are the composed resources.

* `fromFieldPath`: Since we'll be doing bidirectional patches, it could make sense to imply the direction in the field names. For example, `fromComposite`/`toResource` and `fromResource`/`toComposite`. I believe `*FieldPath` is not worth to include since it's already implied that it's a field path. In cases it's not, it's very clear like `value`.
  * We may end up with a separate top field instead of reusing `patches` but I think this would still apply.

hasheddan:

Changes spec.from to spec.compositeTypeRef and spec.to to
spec.resources. This makes both of these fields more reflective of their
purpose in a Composition.

Fixes #1685 

I have:

Installed Crossplane and applied updated `Compositions` in docs.


apis/apiextensions/v1alpha1/composition_types.go
```diff
@@ -48,14 +48,14 @@ var (
...
+	// CompositeTypeRef refers to the type that this composition is compatible.
+	// The values for the underlying resources will be fetched from the
+	// instances of the referenced composite type.
 	// +immutable
-	From TypeReference `json:"from"`
+	CompositeTypeRef TypeReference `json:"compositeTypeRef"`
 
-	// To is the list of target resources that make up the composition.
-	To []ComposedTemplate `json:"to"`
+	// Resources is the list of target resources that make up the composition.
```
muvaf, line 57:
```suggestion
	// Resources is the list of resource templates that will be used when a composite resource referring to this composition is created.
```

I think somehow we should say these are resource _templates_

cluster/charts/crossplane-types/crds/apiextensions.crossplane.io_compositions.yaml
```diff
@@ -171,8 +172,8 @@ spec:
                 provisioned using this composition will be created.
               type: string
           required:
-          - from
-          - to
+          - compositeTypeRef
+          - resources
           - writeConnectionSecretsToNamespace
```
muvaf, line 177: Not directly related but I think we can make `writeConnectionSecretsToNamespace` optional since there are compositions without any connection secret being published like networks.

hasheddan: Good point 👍 I think someone mentioned that in slack recently

hasheddan: @muvaf I updated this to optional, but wasn't sure how we want to handle when it is not supplied. I opted to keep it as a `string` instead of `*string`, which means that we will just pass along a `""` to the managed resource if it is not set explicitly. In that case, we are basically deferring the error handling of passing an empty namespace to the managed resource controller. What do you think about that?

muvaf: I believe this is used only for choosing the namespace of connection secret of the composite resource [here](https://github.com/crossplane/crossplane/blob/master/pkg/controller/apiextensions/composite/api.go#L275). The conn secret ref of managed resources are left to the user to configure either in the template or via patches.

So, if you'd like to tackle that in this PR, I'd suggest making it `*string` since it's optional and put a check in that [APIConfigurator](https://github.com/crossplane/crossplane/blob/master/pkg/controller/apiextensions/composite/api.go#L270) that will return early if that field is not given. I can also tackle it in another PR if you'd like to.

apis/apiextensions/v1alpha1/composition_types.go
```diff
@@ -48,14 +48,14 @@ var (
 
 // CompositionSpec specifies the desired state of the definition.
 type CompositionSpec struct {
-	// From refers to the type that this composition is compatible. The values
-	// for the underlying resources will be fetched from the instances of the
-	// From.
+	// CompositeTypeRef refers to the type that this composition is compatible.
+	// The values for the underlying resources will be fetched from the
+	// instances of the referenced composite type.
```
negz, line 53:
```suggestion
	// CompositeTypeRef specifies the type of composite resource that this
	// composition is compatible with.
```

I'd prefer to remove the "values will be fetched from the composite" part while we're in here, since we're planning to allow bidirectional patching per https://github.com/crossplane/crossplane/issues/1639.

apis/apiextensions/v1alpha1/composition_types.go
```diff
@@ -48,19 +48,21 @@ var (
...
+	// Resources is the list of resource templates that will be used when a
+	// composite resource referring to this composition is created.
+	Resources []ComposedTemplate `json:"resources"`
 
 	// WriteConnectionSecretsToNamespace specifies the namespace in which the
 	// connection secrets of composite resource dynamically provisioned using
 	// this composition will be created.
-	WriteConnectionSecretsToNamespace string `json:"writeConnectionSecretsToNamespace"`
+	// +optional
+	WriteConnectionSecretsToNamespace string `json:"writeConnectionSecretsToNamespace,omitempty"`
```
negz, line 65: I can't comment on the existing thread about this for some reason, but I tend to agree with @muvaf that we should make this a `*string` since it's now optional. I believe it's only used in one place, so updating the corresponding controller code should hopefully be straightforward.

cluster/charts/crossplane-types/crds/apiextensions.crossplane.io_compositions.yaml
hasheddan: @muvaf yep that's what I was looking at. So to be clear, we would just skip setting `WriteConnectionSecretToReference` instead of setting it with a `""` namespace. The other option would be returning an error, but that would obviously make this change useless if we didn't want to set `WriteConnectionSecretToReference` for a resource.

apis/apiextensions/v1alpha1/composition_types.go
hasheddan: Agreed, responded in thread above.

hasheddan:
@muvaf @negz could y'all take another look here?

cluster/charts/crossplane-types/crds/apiextensions.crossplane.io_compositions.yaml
muvaf: That's exactly what I had in mind, too. Thanks!

PR #1703, opened 2020-09-01
Generate clients for stable types

hasheddan:

Adds clientsets for types in apiextensions group. Client generation for
additional groups can be expanded in the future, but most other types
are relatively unstable at current time.

I have:

`make generate`
`make build`

Anecdotally, I also tested against a live cluster with a quickly built CLI.


negz:
@hasheddan Could you expand this PR with a little context around why we prefer clientsets in some contexts (i.e. CLI tools)? Presumably the alternative would be to use the controller-runtime `client.Client.

hasheddan:
@negz yep, the controller-runtime client or the client-go dynamic client could be used. That being said, the generated clientset feels a lot better to work with and implements methods like `Watch`, which can be useful outside the context of a reconcile loop. I would not be surprised if other tools built in the community also consumed the clientsets if we were offering them. Do you have any reservations about generating these?

negz:
> Do you have any reservations about generating these?

I don't, but it wasn't obvious to me from the PR description why we wanted to do this. Understanding why we're making this change helps me review it.

> I would not be surprised if other tools built in the community also consumed the clientsets if we were offering them.

How likely do you think this is? Such tools would have to take a dependency on c/c just to get the clientsets. If we weren't trending toward building the `crank` CLI in this repo I'd be tempted to suggest we make these clientsets a distinct Go module.

hasheddan:
> If we weren't trending toward building the crank CLI in this repo I'd be tempted to suggest we make these clientsets a distinct Go module.

I assume that module would also include the types? Is there an advantage to those being versioned separately from the core Crossplane controllers?

muvaf:
@hasheddan I see that methods are pretty similar to what we're using with controller-runtime across all the codebases. Is it the `Watch` function that makes you think these clientsets are needed? I assume that you'd like that function for CLI in cases where you'd want to have the user wait for a condition.

hasheddan:
@muvaf yes, and also the potential to have expanded methods that may do type specific operations, such as `client.PkgV1alpha1().Provider("provider-aws").Revisions()`.

muvaf:
```golang
// Watch returns a watch.Interface that watches the requested compositeResourceDefinitions.
func (c *compositeResourceDefinitions) Watch(ctx context.Context, opts v1.ListOptions) (watch.Interface, error) {
	var timeout time.Duration
	if opts.TimeoutSeconds != nil {
		timeout = time.Duration(*opts.TimeoutSeconds) * time.Second
	}
	opts.Watch = true
	return c.client.Get().
		Resource("compositeresourcedefinitions").
		VersionedParams(&opts, scheme.ParameterCodec).
		Timeout(timeout).
		Watch(ctx)
}
```

Would it be possible to have some tooling for the `Watch` operation in the CLI's package for all types it'd possibly work? TBH, I'm sort of hesitant to have & maintain the clientsets given we're all in on controller-runtime in all codebases, which has served well IMO. (I don't really like the general syntax of clientset as well but that's not really a big deal.)

So, I'd say maybe we should investigate an alternative like CLI having its own tooling for the `Watch` operation instead of bringing in full clientsets for all core Crossplane types mainly because of the divergence of the tooling for almost same operations.

hasheddan:
> I'm sort of hesitant to have & maintain the clientsets given we're all in on controller-runtime in all codebases

> So, I'd say maybe we should investigate an alternative like CLI having its own tooling for the Watch operation instead of bringing in full clientsets for all core Crossplane types mainly because of the divergence of the tooling for almost same operations.

@muvaf I hear what you are saying, but these two points feel somewhat incompatible. These clientsets are auto-generated, whereas it seems that you are suggesting we write our own client operations wrapper, which feels like a heavier maintenance burden?

muvaf:
> These clientsets are auto-generated, whereas it seems that you are suggesting we write our own client operations wrapper, which feels like a heavier maintenance burden?

@hasheddan What I'm saying is that a generic single `Watch` function could be written inside the CLI package that can be used for all types instead of generating full clientsets for every core Crossplane type. Though it may not be as straightforward because of some scheme stuff etc.

hasheddan:
@muvaf yep, I get what you are saying, but generating clientsets just feels like a pretty low maintenance cost to me. If you feel strongly I can try to put together a generic `Watch` function and we can see how big the hurdles are for that. One other thing to take into consideration is that folks in the k8s community are generally pretty familiar with these generated clientsets, so they might prefer to work with them. That is strictly speculation though, and we don't have any other consumers at this time other than the proposed CLI :slightly_smiling_face:

jbw976:
We've been using generated clientsets in the Rook project for probably a couple years now, and they have not given us many problems in terms of maintenance.  I would prefer using the upstream supported, community adopted, automatically generated clientsets if that makes parts of our codebase easier instead of writing any libraries or functions that we maintain.

As @hasheddan mentioned, there's benefit beyond just a Watch function.  Essentially you're getting a strongly typed model to program with, which can feel nice when you're programming against a known/expected type.

negz:
> I assume that module would also include the types?

Yes, that would make sense to me. Similar to https://github.com/kubernetes/api. We wanted to head in this direction by breaking out our standard resource claims, because we didn't want (for example) all providers depending on crossplane/crossplane. This became a moot point once we deprecated those claims.

> Is there an advantage to those being versioned separately from the core Crossplane controllers?

I don't feel the advantage is separate versioning, but rather a smaller target to depend on. If I'm building a client for Crossplane I probably just want my project to depend on the API types and clientsets, not the entire crossplane/crossplane Go module.

Given that in practice the only known client of these types is `crank`, and that you intend to build `crank` in this repo, I think it's reasonable to add the clientsets here for now. However I'd feel more comfortable if we at least raised an issue to track breaking them out before any client outside this repo depended on them. I would consider going so far as to put them in an `internal` package to prevent anyone from taking a dependency on them.

muvaf:
> yep, I get what you are saying, but generating clientsets just feels like a pretty low maintenance cost to me. If you feel strongly I can try to put together a generic Watch function and we can see how big the hurdles are for that.

@hasheddan No, I wouldn't really block this effort. It's just that having two alternatives is sort of divergence in the codebase that may not be necessary, for example `crank` code could look different and we may need additional utilities here and there on top of what we've been developing for controller-runtime. But I'm not as knowledgeable about crank so I trust your judgement here.

> As @hasheddan mentioned, there's benefit beyond just a Watch function. Essentially you're getting a strongly typed model to program with, which can feel nice when you're programming against a known/expected type.

I'm sure there are benefits to each approaches, my only reservation is about having both of them with functionality overlap with high percentage.

negz (approved):

Issue #1678, opened 2020-08-19
Enable cdk8s users to offer constructs as native CRDs via Crossplane CustomComposition


prasek:
Part of #1955 First-class multi-language support 🚀 

### Background
[Crossplane Composition](https://crossplane.github.io/docs/v0.12/introduction/composition.html) (with [associated refinements](https://github.com/crossplane/example-cnp)) supports authoring `CompositeResourceDefinitions` (XRDs) that publish CRDs to the Kubernetes API and `Composition` resources that provides a simple built-in YAML templating language for composing the cloud service primitives in Crossplane Providers for AWS, GCP, Azure, Alibaba, Packet, and more. You can define, compose, and offer new CRDs entirely in yaml without writing  any code, that can be consumed using `kubectl`, GitOps, or any tool that work with the Kubernetes API. This is suitable for many use cases and offers a Kubernetes-native authoring experience.

However in not everyone prefers authoring YAML, so having a toolkit like [cdk8s](https://github.com/awslabs/cdk8s) lets you compose cdk8s `Constructs` with your language of choice (Javascript, Python) with loops, conditionals, and a broad ecosystem of packages and libraries. A cdk8s `synth` command is run client-side, emits the yaml for a cdk8s program, and then `kubectl apply` can then be run to apply it to your cluster. 

cdk8s `Constructs` can't currently be published as native CRDs, so they can consumed via the Kubernetes API or composed into larger composition.

### Proposal
Enable cdk8s users to offer `Constructs` as native CRDs via Crossplane `CompositeResourceDefinitions` (XRDs) and `CustomComposition` resources. A new `cdk8s pack` command would generate a sidecar image that bundles the cdk8s program as a simple web server that can be used as the `source` in a Crossplane `CustomComposition` resource (e.g. https://github.com/eladb/cdk8s-pack-prototype). The yaml resources emitted by the sidecar would be would be labelled so `kubectl apply --prune` semantics can be applied by the Crossplane custom-composition controller.

### Notes
- Want to enable CDK8s users to offer constructs as native CRDs via Crossplane Composition
- Idea is to author your definition logic in one of the languages that cdk8s supports and compose it as part of a Crossplane Composition.
- DX: `cdk8s pack`: take a construct and produce a sidecar container image that serves this construct that serves this construct, that can be invoked from a Crossplane composition using a new toCdk8s field (final name TBD)
[...]

PR #1705, opened 2020-09-02
Custom compositions design doc

muvaf:
## Description of your changes

This is the very first pass. Its first target is to have cdk8s as composition engine as mentioned in https://github.com/crossplane/crossplane/issues/1678#issuecomment-678347186 and https://github.com/awslabs/cdk8s/issues/24 . I will fill the details after syncing with cdk8s folks. cc @negz @prasek @eladb

Fixes #

I have:

### DIFF
--- a/design/design-doc-custom-composition.md
+++ b/design/design-doc-custom-composition.md
@@ -1,0 +1,179 @@
+# Custom Composition Types
+
+* Owner: Muvaffak Onuş (@muvaf)
+* Reviewers: Crossplane Maintainers
+* Status: Draft
+
+## Background
+
+## Proposal
+
+### Authorship Experience
+
+Let's say we have the following `CompositeResourceDefinition` that defines a `PrivateNetwork`
+type.
+
+```yaml
+apiVersion: apiextensions.crossplane.io/v1alpha1
+kind: CompositeResourceDefinition
+metadata:
+  name: privatenetworks.aws.example.org
+spec:
+  crdSpecTemplate:
+    group: aws.example.org
+    version: v1alpha1
+    names:
+      kind: PrivateNetwork
+      listKind: PrivateNetworkList
+      plural: privatenetworks
+      singular: privatenetwork
+    validation:
+      openAPIV3Schema:
+        type: object
+        properties:
+          spec:
+            type: object
+            properties:
+              region:
+                type: string
+                description: Geographic region in which nodes will exist.
+              providerConfigRef:
+                type: object
+                description: Crossplane AWS provider credentials to use.
+                properties:
+                  name:
+                    type: string
+                required:
+                - name
+            required:
+            - region
+            - providerConfigRef
+```
+
+User can have multiple `Composition` or custom compositions that target this new
+type. Let's assume they would like to write a cdk8s app and use it as the composition
+engine. They'd need to write the cdk8s app that will take the following resources
+as input:
+  * `PrivateNetwork` custom resource.
+  * `CustomComposition` custom resource.
+  It will return all the resulting resources as output.
+
+This conversion function will be wrapped by an HTTP server and pushed as an image
+whose `ENTRYPOINT` is to start that HTTP server. The input will come as POST
+request and the body of the result will consist of all the output resources.
+
+At this point, we have our cdk8s app that will act as composition engine. Now,
+let's go through how the consumers of that app will be able to use it.
+
+### Consumption Experience
+
+Now that our composition engine is ready, we'll instruct Crossplane to install
+that composition engine and send the appropriate requests. In order to do that,
+the following `CustomComposition` will be created:
+
+```yaml
+apiVersion: apiextensions.crossplane.io/v1alpha1
+kind: CustomComposition
+metadata:
+  name: acmenetwork.acme.org
+spec:
+  writeConnectionSecretsToNamespace: crossplane-system
+  reclaimPolicy: Delete
+  compositeTypeRef:
+    apiVersion: aws.example.org/v1alpha1
+    kind: PrivateNetwork
+  image: acme/network-cdk8s:1.0.0
+```
+
+Once this `CustomComposition` is created, a new `Deployment` will be created by
+Crossplane with the following properties:
+
+```yaml
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: acmenetwork.acme.org
+  labels:
+    app: acmenetwork.acme.org
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: acmenetwork.acme.org
+  template:
+    metadata:
+      labels:
+        app: acmenetwork.acme.org
+    spec:
+      containers:
+      - name: custom-composition
+        image: crossplane/custom-composition:0.0.1
+        env:
+        # COMPOSITE_TYPE will let the custom-composition controller know what type
+        # it should watch.
+        - name: COMPOSITE_TYPE
+          value: privatenetworks.aws.example.org
+        # CUSTOMCOMPOSITION_NAME is the name of the CustomComposition that will
+        # be sent over to composition engine alongside composite instance. 
+        - name: CUSTOMCOMPOSITION_NAME
+          value: acmenetwork.acme.org
+      - name: composition-engine
+        image: acme/network-cdk8s:1.0.0
+        ports:
+        - containerPort: 80
+```
+
+After this `Deployment` is up, the container named `custom-composition` will start
+watching `PrivateNetwork` types and it will reconcile only the ones with the following
+reference:
+```yaml
+apiVersion: acme.example.org
+kind: PrivateNetwork
+metadata:
+  name: something
+spec:
+  region: us-east-1
+  providerConfigRef:
+    name: my-creds
+  compositionRef:
+    apiVersion: apiextensions.crossplane.io/v1alpha1
+    kind: CustomComposition
+    name: acmenetwork.acme.org
+```
+
+In each of its reconciliation pass:
+1. It will make an HTTP request to `cdk-app` container
+with the `PrivateNetwork` instance and `CustomComposition` instance.
+  * The reason we include `CustomComposition` is that the cdk8s app might be configured
+    with some additional fields in the future or metadata of `CustomComposition`.
+2. The response from the cdk-app container is expected to be a set of valid
+    Kubernetes YAMLs.
+3. `custom-composition` will add owner references to all objects and run the
+   equivalent of `kubectl apply --prune -f <response YAMLs>`.
+
+At this point, we got all of our output deployed to the cluster.
+
+## Alternatives Considered
+
+### Separate Type for Each Custom Composition Type
+
+We could have `CDKComposition`, `HelmComposition` types with their own additional
+fields. But since for each instance of those types, a new `Deployment` is created,
+the entity that creates this `Deployment` couldn't be Crossplane because it'd not
+know of those types. So, for each type, users would have to install a CRD and a
+separate controller that will do this `Deployment` creation operation.
+
+We could possibly include those custom composition types in Crossplane with the
+assumption that the number of types will not be that big anyway. However, some users
+would prefer the flexibility of use of their in-house custom compositions and having
+to have it merged to Crossplane would be a big barrier.
+
+Another possibility is simply have each type be almost completely decoupled from
+Crossplane by having them install that `Deployment` creator entity, too. For instance,
+they could install their CRD and a `Deployment` that has only a controller which creates
+a `Deployment` for every instance of that custom composition type. But this approach
+expands the attack surface since now they have to give `Deployment` permission to
+yet another `ServiceAccount` besides from Crossplane itself. Also, they'd have to
+write a controller to have a custom composition, which could be alleviated by
+having a generic one since the operation itself is not a complex one, but that's
+yet another thing to maintain.


design/design-doc-custom-composition.md
```diff
@@ -0,0 +1,179 @@
+# Custom Composition Types
+
+* Owner: Muvaffak Onuş (@muvaf)
+* Reviewers: Crossplane Maintainers
+* Status: Draft
+
+## Background
```
prasek, line 7:
```suggestion
## Background
[Crossplane Composition](https://crossplane.github.io/docs/v0.12/introduction/composition.html) (with [associated refinements](https://github.com/crossplane/example-cnp)) supports authoring `CompositeResourceDefinitions` (XRDs) that publish CRDs to the Kubernetes API and `Composition` resources that provides a simple built-in YAML templating language for composing the cloud service primitives in Crossplane Providers for AWS, GCP, Azure, Alibaba, Packet, and more. You can define, compose, and offer new CRDs entirely in yaml without writing  any code, that can be consumed using `kubectl`, GitOps, or any tool that work with the Kubernetes API. This is suitable for many use cases and offers a Kubernetes-native authoring experience.

However in not everyone prefers authoring YAML, so having a toolkit like [cdk8s](https://github.com/awslabs/cdk8s) lets you compose cdk8s `Constructs` with your language of choice (Javascript, Python) with loops, conditionals, and a broad ecosystem of packages and libraries. A cdk8s `synth` command is run client-side, emits the yaml for a cdk8s program, and then `kubectl apply` can then be run to apply it to your cluster. 

cdk8s `Constructs` can't currently be published as native CRDs, so they can consumed via the Kubernetes API or composed into larger composition.
```

design/design-doc-custom-composition.md
```diff
@@ -0,0 +1,179 @@
...
+
+## Proposal
```
prasek, line 9:
```suggestion
## Proposal
Enable cdk8s users to offer `Constructs` as native CRDs via Crossplane `CompositeResourceDefinitions` (XRDs) and `CustomComposition` resources that can use sidecar images generated by the new `cdk8s pack` command that bundles the cdk8s program as a simple web server that can be used from a Crossplane `Composition` resource to emit the yaml for a Crossplane `CustomComposition` (e.g. https://github.com/eladb/cdk8s-pack-prototype). The emitted yaml resources would be labelled so `kubectl apply --prune` semantics can be observed.
```

design/design-doc-custom-composition.md
negz: Nit: The first paragraph here reads a bit like a sales pitch for composition, which isn't the tone we typically take in a design doc.

Perhaps we should try to incorporate this content while also making it clear that this type would be used for more than just CDK8s based compositions (e.g. for other external tools like Helm, Kustomize, etc) if that is indeed the case?

design/design-doc-custom-composition.md
negz: Should we move this under a goals heading?

design/design-doc-custom-composition.md
```diff
@@ -0,0 +1,179 @@
...
+type. Let's assume they would like to write a cdk8s app and use it as the composition
+engine. They'd need to write the cdk8s app that will take the following resources
+as input:
+  * `PrivateNetwork` custom resource.
+  * `CustomComposition` custom resource.
+  It will return all the resulting resources as output.
+
+This conversion function will be wrapped by an HTTP server and pushed as an image
+whose `ENTRYPOINT` is to start that HTTP server. The input will come as POST
+request and the body of the result will consist of all the output resources.
```
negz, line 63: Would this pattern work for other 'custom' tools besides cdk8s? Presumably we could (for example) write a Docker image that invoked Kustomize or similar via a compatible HTTP API.

design/design-doc-custom-composition.md
```diff
@@ -0,0 +1,179 @@
...
+kind: CustomComposition
+metadata:
+  name: acmenetwork.acme.org
+spec:
+  writeConnectionSecretsToNamespace: crossplane-system
+  reclaimPolicy: Delete
+  compositeTypeRef:
+    apiVersion: aws.example.org/v1alpha1
+    kind: PrivateNetwork
+  image: acme/network-cdk8s:1.0.0
```
negz, line 85: I wonder if (eventually) we would do anything to validate that this image we're running is in fact serving custom composition logic. We probably don't want to become a vector to deploy arbitrary containers.

design/design-doc-custom-composition.md
```diff
@@ -0,0 +1,179 @@
...
+          value: acmenetwork.acme.org
+      - name: composition-engine
+        image: acme/network-cdk8s:1.0.0
+        ports:
+        - containerPort: 80
+```
+
+After this `Deployment` is up, the container named `custom-composition` will start
+watching `PrivateNetwork` types and it will reconcile only the ones with the following
+reference:
```
negz, line 128: This seems to imply that the custom composition controller would be separate from core Crossplane. Why is this? Would the custom composition controller change or remain the same if the composition engine changed (e.g. from cdk8s to kustomize)?

negz (commented):
Nice start on this @muvaf! I like the idea of a single `CustomComposition` that decouples the reconcile logic by deferring it to an external composition engine image.

design/design-doc-custom-composition.md
muvaf: > Presumably we could (for example) write a Docker image that invoked Kustomize or similar via a compatible HTTP API.

Yeah, that's what I had in mind. Something need to be the wrapper server and give the composite CR as input to these tools.

design/design-doc-custom-composition.md
muvaf: Honestly, that's the part I'm struggling with most. I am hesitant to add yet another reason for Crossplane to be super user. One option is to have the user create the `Deployment` alongside `CustomComposition`:
```yaml
kind: CustomComposition
metadata:
  name: acmenetwork.acme.org
spec:
  reclaimPolicy: Delete
  compositeTypeRef:
    apiVersion: aws.example.org/v1alpha1
    kind: PrivateNetwork
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: acmenetwork.acme.org
  labels:
    app: acmenetwork.acme.org
spec:
  replicas: 1
  selector:
    matchLabels:
      app: acmenetwork.acme.org
  template:
    metadata:
      labels:
        app: acmenetwork.acme.org
    spec:
      containers:
      - name: custom-composition
        image: crossplane/custom-composition:0.0.1
        env:
        # COMPOSITE_TYPE will let the custom-composition controller know what type
        # it should watch.
        - name: COMPOSITE_TYPE
          value: privatenetworks.aws.example.org
        # CUSTOMCOMPOSITION_NAME is the name of the CustomComposition that will
        # be sent over to composition engine alongside composite instance. 
        - name: CUSTOMCOMPOSITION_NAME
          value: acmenetwork.acme.org
      - name: composition-engine
        image: acme/network-cdk8s:1.0.0
        ports:
        - containerPort: 80
```

`CustomComposition` wouldn't have any controllers and its only goal would be to be referenced so that composite can demand this `CustomComposition`. It saves Crossplane from `Deployment` permission but UX is getting a hit since user has to write that `Deployment`.

> I wonder if (eventually) we would do anything to validate that this image we're running is in fact serving custom composition logic. We probably don't want to become a vector to deploy arbitrary containers.

I'm not sure if we could. In fact, the `Deployment` needs a service account with capabilities to create the output of the cdk8s app and there isn't much way for Crossplane to know what cdk8s needs until actual creation.

Thinking through this while writing, I think that might be the right approach. The full YAML that cdk8s can print could have:
* 1 `CompositeResourceDefinition` (optional)
* 1 `CustomComposition`
* 1 `Deployment`
* 1 `ServiceAccount`, 1 `ClusterRole`, 1 `ClusterRoleBinding` (these will need to be created by the user in all cases, otherwise Crossplane would have to give `cluster-admin` role to the `Deployment`)

(having a spider-sense saying we'll land on making this a package, probably a `Configuration` or a new type)

@negz what's your take?

design/design-doc-custom-composition.md
muvaf: It'd remain the same. Having both of them in the same `Pod` saves us from in-cluster networking, which could require TLS handling and expand the attack surface, thought that's manageable. But the key point is what entity will create the output YAMLs. If it's core Crossplane pod, then that means cdk8s app can do anything it wants in the cluster since Crossplane has `cluster-admin` and I'd feel really uncomfortable doing that because then `CustomComposition` permission would be equal to root, which is not immediately clear to the user. So, having the deployer together with cdk8s app could give the ability to user to limit what cdk8s app can ever deploy via RBAC specific to its `ServiceAccount`.

design/design-doc-custom-composition.md
muvaf: I have given this some more thought. To some extent, this is very similar to a provider in the sense that it sees a type and creates some objects in an API, though API in this case is Kubernetes and the created objects are not limited to CRDs it brings with itself.

So, an option alternative to the earlier [suggestion](https://github.com/crossplane/crossplane/pull/1705/files#r483710125) is that we can have a type, like `ExternalComposition.pkg.crossplane.io` (not sure about the name), that will create the RBAC resources according to what's given in `crossplane.yaml` ([permission](https://github.com/crossplane/crossplane/issues/1767) requests to certain kinds it can manage) and also create the necessary `Deployment` with two containers (one is `external-composition-controller` the other one is the app itself). Users would choose the `ExternalComposition` in their `compositionRef`, Crossplane itself wouldn't reconcile it at all and the `external-composition-controller` would reconcile and call the endpoints of the app running as sidecar container.

The overall DX using cdk8s would be that I write some JavaScript, a `crossplane.yaml` for metadata and permissions and a `Dockerfile`. Then `build` it similar to provider builds & push (image path is given in `crossplane.yaml`). Crossplane package manager will know what operations to do when it deals with this `kind` of package.

In the earlier [suggestion](https://github.com/crossplane/crossplane/pull/1705/files#r483710125) we can have the CLI generate those extra YAMLs and let user do the `kubectl apply`. Developer could make sure that each build runs that generation and package them up in a Helm chart for external usage, similar to controllers out in the wild. But instead, if we do the generation in Package Manager and use the standard Crossplane packaging format, it's less cumbersome while we take on all the responsibility regarding RBAC management of that package. The main difference between providers and this package would be that the final `Deployment` is controller+server and that may require some customization.

I feel closer to a new packaging type, it feels like even if we do start with generation of those YAMLs using CLI we'll end up deciding to make a new package type since the overall operation is pretty similar to what package manager does. But I'm not yet fully sold on the idea.

What do you guys think? Is this feasible at all and is it a good API to interact with? @hasheddan @negz @kasey @prasek

hasheddan: @muvaf I am thinking along the same lines as you and I see a lot of parallels to some of the work in https://github.com/crossplane-contrib/crisscross. The difference between `crisscross` and this scenario is that we can't use generic controller logic (i.e. the managed reconciler) and just define certain endpoints when the controller reaches out for the custom logic. However, I would like to explore if we could bring these two ideas closer together and potentially implement a single package type that satisfies both (right now I am thinking something along the lines of `Function.pkg.crossplane.io`.

We could go about bringing the two ideas together by encoding more of the reconcile logic into the functions themselves, and, in the case of crisscross, potentially abstracting that away with a library / base image so users still _feel_ like they are just writing endpoints. In Crossplane, this would look like dynamically spinning up controllers that only make an http request to a the function, where all logic is encompassed.

As you mentioned, installing a `Function` would allow us to make sure the appropriate RBAC is provided, as well as setup certificates for mutual TLS, etc. What do you think about this?

design/design-doc-custom-composition.md
```diff
@@ -0,0 +1,179 @@
...
+                  name:
+                    type: string
+                required:
+                - name
+            required:
+            - region
+            - providerConfigRef
+```
+
+User can have multiple `Composition` or custom compositions that target this new
```
mpetason, line 53:
```suggestion
Users can have multiple `Composition` or custom compositions that target this new
```

design/design-doc-custom-composition.md
```diff
@@ -0,0 +1,179 @@
...
+as input:
+  * `PrivateNetwork` custom resource.
+  * `CustomComposition` custom resource.
+  It will return all the resulting resources as output.
+
+This conversion function will be wrapped by an HTTP server and pushed as an image
+whose `ENTRYPOINT` is to start that HTTP server. The input will come as POST
+request and the body of the result will consist of all the output resources.
+
+At this point, we have our cdk8s app that will act as composition engine. Now,
```
mpetason, line 65:
```suggestion
At this point, we have our cdk8s app that will act as a composition engine. Now,
```

design/design-doc-custom-composition.md
```diff
@@ -0,0 +1,179 @@
...
+let's go through how the consumers of that app will be able to use it.
+
+### Consumption Experience
+
+Now that our composition engine is ready, we'll instruct Crossplane to install
+that composition engine and send the appropriate requests. In order to do that,
```
mpetason, line 71:
```suggestion
it and send the appropriate requests. In order to do that,
```

design/design-doc-custom-composition.md
```diff
@@ -0,0 +1,179 @@
...
+  region: us-east-1
+  providerConfigRef:
+    name: my-creds
+  compositionRef:
+    apiVersion: apiextensions.crossplane.io/v1alpha1
+    kind: CustomComposition
+    name: acmenetwork.acme.org
+```
+
+In each of its reconciliation pass:
```
mpetason, line 144:
```suggestion
In each of its reconciliation passes:
```

design/design-doc-custom-composition.md
```diff
@@ -0,0 +1,179 @@
...
+1. It will make an HTTP request to `cdk-app` container
+with the `PrivateNetwork` instance and `CustomComposition` instance.
+  * The reason we include `CustomComposition` is that the cdk8s app might be configured
+    with some additional fields in the future or metadata of `CustomComposition`.
+2. The response from the cdk-app container is expected to be a set of valid
+    Kubernetes YAMLs.
+3. `custom-composition` will add owner references to all objects and run the
+   equivalent of `kubectl apply --prune -f <response YAMLs>`.
+
+At this point, we got all of our output deployed to the cluster.
```
mpetason, line 154:
```suggestion
At this point, we have all of our output deployed to the cluster.
```

design/design-doc-custom-composition.md
```diff
@@ -0,0 +1,179 @@
...
+
+## Alternatives Considered
+
+### Separate Type for Each Custom Composition Type
+
+We could have `CDKComposition`, `HelmComposition` types with their own additional
+fields. But since for each instance of those types, a new `Deployment` is created,
+the entity that creates this `Deployment` couldn't be Crossplane because it'd not
+know of those types. So, for each type, users would have to install a CRD and a
```
mpetason, line 163:
```suggestion
fields. For each instance of those types, a new `Deployment` is created, but
the entity that creates this `Deployment` couldn't be Crossplane because it would not
know about those types. So, for each type, users would have to install a CRD and a
```

design/design-doc-custom-composition.md
```diff
@@ -0,0 +1,179 @@
...
+separate controller that will do this `Deployment` creation operation.
+
+We could possibly include those custom composition types in Crossplane with the
+assumption that the number of types will not be that big anyway. However, some users
+would prefer the flexibility of use of their in-house custom compositions and having
```
mpetason, line 168:
```suggestion
would prefer the flexibility to use their in-house custom compositions and having
```

design/design-doc-custom-composition.md
```diff
@@ -0,0 +1,179 @@
...
+to have it merged to Crossplane would be a big barrier.
+
+Another possibility is simply have each type be almost completely decoupled from
+Crossplane by having them install that `Deployment` creator entity, too. For instance,
```
mpetason, line 172:
```suggestion
Another possibility is to simply have each type be almost completely decoupled from
Crossplane by having them install that `Deployment` creator entity. For instance,
```

design/design-doc-custom-composition.md
```diff
@@ -0,0 +1,179 @@
...
+they could install their CRD and a `Deployment` that has only a controller which creates
+a `Deployment` for every instance of that custom composition type. But this approach
+expands the attack surface since now they have to give `Deployment` permission to
+yet another `ServiceAccount` besides from Crossplane itself. Also, they'd have to
```
mpetason, line 176:
```suggestion
yet another `ServiceAccount` in addition to Crossplane itself. Also, they'd have to
```

mpetason (commented):
small updates that may read a little bit better. I need to learn more about the project to update other parts with suggestions.

negz:
I think we should still proceed with the custom composition concept, but this draft design is a little focused on CDK specifically and is currently missing the context I'd like to provide to the Crossplane community when we propose "custom compositions" as an alternative to expanding the existing `Composition` type with support for things like conditionals and iteration. I've raised https://github.com/crossplane/crossplane/issues/2524 to track this general idea, and cross referenced this design document for future reference. After discussion with @muvaf I'm going to close this design in favor of using that issue to track the idea.

PR #1706, opened 2020-09-04
Crossplane RBAC Manager design doc

negz:

This PR introduces a design document for an "RBAC manager" process that runs alongside Crossplane, creating and managing (in some cases opinionated) RBAC roles for providers, composite resources, and composite resource claims. This RBAC manager process would replace similar functionality available in the contemporary package manager.

I have:

### DIFF
--- a/design/design-doc-rbac-manager.md
+++ b/design/design-doc-rbac-manager.md
@@ -234,14 +234,23 @@ rules:
 ```
 
 Crossplane's user-facing cluster roles are inspired by the [user-facing roles]
-of Kubernetes - `admin`, `edit`, and `view`. Note that Kubernetes distinguishes
-the `cluster-admin` cluster role, which is intended to be granted at cluster
-scope via a cluster role binding, from the `admin`, `edit`, and `view` cluster
-roles, which are intended to be granted within a particular namespace using a
-role binding. Crossplane does not make this distinction. The `crossplane-admin`,
-`crossplane-edit`, and `crossplane-view` roles are intended to be granted at the
-cluster scope, and thus grant admin, edit, or view access at the cluster scope.
-Crossplane provides distinct edit and view roles for each namespace.
+of Kubernetes - `admin`, `edit`, and `view`, albeit with a few differences:
+
+* Kubernetes distinguishes the `cluster-admin` cluster role, which is intended
+  to be granted at cluster scope via a cluster role binding, from the `admin`,
+  `edit`, and `view` cluster roles, which are intended to be granted within a
+  particular namespace using a role binding. Crossplane does not make this
+  distinction. The `crossplane-admin`, `crossplane-edit`, and `crossplane-view`
+  roles are intended to be granted at the cluster scope, and thus grant admin,
+  edit, or view access at the cluster scope.
+* Crossplane includes a `browse` role, which grants view access to composite
+  resources and compositions (not providers and managed resources). This role
+  may be granted to subjects who can author resource claims in a namespace (or
+  namespaces), and who should be able to browse the compositions or composite
+  resources that they may reference or select.
+* Crossplane provides distinct edit and view roles for each namespace. This
+  allows the admin role (for example) to grant access to different resource
+  claims from one namespace to another.
 
 The `crossplane-admin` role is automatically bound to the `crossplane:masters`
 group for convenience. It grants the following rules:
@@ -382,6 +391,35 @@ rules:
   verbs: [get, list, watch]
 ```
 
+The `crossplane-browse` role allows read-only access to composite resources and
+compositions.
+
+```yaml
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: crossplane-view
+rules:
+# Crossplane browsers have access to view events.
+- apiGroups: [""]
+  resources: [events]
+  verbs: [get, list, watch]
+# Crossplane browsers have read-only access to compositions and XRDs. This
+# allows them to discover and select an appropriate composition when creating a
+# resource claim.
+- apiGroups:
+  - apiextensions.crossplane.io
+  resources: ["*"]
+  verbs: [get, list, watch]
+# Crossplane browsers have read-only access to all of the composite resources it
+# defines. This allows them to claim a statically provisioned composite
+# resource.
+- apiGroups: [xr.example.org]
+  resources:
+  - examplecomposites
+  verbs: [get, list, watch]
+```
+
 The RBAC manager creates `edit` and `view` cluster roles for each namespace.
 These cluster roles are 'namespace aligned' - they are intended to grant access
 to a particular namespace via a `RoleBinding` - but not namespace scoped. This
@@ -423,12 +461,6 @@ rules:
 - apiGroups: [""]
   resources: [secrets]
   verbs: ["*"]
-# Crossplane editors have read-only access to composite resource definitions and
-# compositions. This allows them to discover and select an appropriate
-# composition when creating a resource claim.
-- apiGroups: [apiextensions.crossplane.io]
-  resources: ["*"]
-  verbs: [get, list, watch]
 # Crossplane editors have full access to all of the composite resource claims
 # that an admin has chosen to enable in their namespace.
 - apiGroups: [xr.example.org]
--- a/design/design-doc-rbac-manager.md
+++ b/design/design-doc-rbac-manager.md
@@ -577,7 +577,7 @@ cluster role aggregates to the `crossplane-view` cluster role, and any `view`
 role for a namespace in which the composite resource may be claimed.
 
 Composite resources may be claimed in any namespace that has an annotation with
-the key `rbac.crossplane.io/composites.example.org`, and the value `xrd-enabled`
+the key `rbac.crossplane.io/composites.example.org`, set to `xrd-claim-accepted`
 (where `composites.example.org` is the name of the XRD). When the RBAC manager
 encounters a namespace with one or more such annotations, it creates `edit` and
 `view` cluster roles for that namespace.
@@ -588,8 +588,8 @@ kind: Namespace
 metadata:
   name: example
   annotations:
-    rbac.crossplane.io/composites.example.org: xrd-enabled
-    rbac.crossplane.io/examplecomposites.xr.example.org: xrd-enabled
+    rbac.crossplane.io/composites.example.org: xrd-claim-accepted
+    rbac.crossplane.io/examplecomposites.xr.example.org: xrd-claim-accepted
 ```
 
 The above namespace, for example, will result in the creation of the following
--- a/design/design-doc-rbac-manager.md
+++ b/design/design-doc-rbac-manager.md
@@ -36,7 +36,7 @@ that it can reconcile.
 
 When Crossplane defines new custom resources some entities (or subjects, in
 Kubernetes RBAC terminology) must be granted access to those resources in order
-for them to be useful. These subjects can be categorised as software, or humans.
+for them to be useful. These subjects can be categorised as software, or users.
 
 Software, in this context, corresponds to the service accounts that Crossplane
 or a provider run as. When the package manager reconciles a `Provider`, it runs
@@ -45,7 +45,7 @@ managed resources unless its service account is granted access to do so.
 Furthermore, the core Crossplane controllers cannot create the provider's
 managed resources, and thus cannot compose them into a higher level resource.
 
-Humans are the people who use Crossplane to orchestrate external systems.
+Users are the people who use Crossplane to orchestrate external systems.
 Crossplane typically categorises these people into three roles, which may
 overlap:
 
@@ -64,7 +64,7 @@ able to grant this access because the API server requires that a subject be
 granted a particular RBAC role before they are able to grant that role to
 another subject.
 
-In either case - software or human - it is possible for a superuser to manually
+In either case - software or user - it is possible for a superuser to manually
 author the appropriate RBAC roles and bind them to the appropriate subjects to
 make Crossplane work. A superuser in this context is either a subject bound to
 the built in `cluster-admin` RBAC `ClusterRole`, or a subject bound to a role
@@ -233,7 +233,7 @@ rules:
   verbs: [get, list, watch, update, patch]
 ```
 
-Crossplane's human-facing cluster roles are inspired by the [user-facing roles]
+Crossplane's user-facing cluster roles are inspired by the [user-facing roles]
 of Kubernetes - `admin`, `edit`, and `view`. Note that Kubernetes distinguishes
 the `cluster-admin` cluster role, which is intended to be granted at cluster
 scope via a cluster role binding, from the `admin`, `edit`, and `view` cluster
--- a/design/design-doc-rbac-manager.md
+++ b/design/design-doc-rbac-manager.md
@@ -732,7 +732,7 @@ the RBAC manager without compromising any core Crossplane functionality. In
 addition to completely disabling the RBAC manager it will be possible to limit
 the scope of the RBAC manager via a flag. The following levels will exist:
 
-* `--manage=all` - Include all functionality described in this design.
+* `--manage=all` - Include all functionality described in this design. (Default)
 * `--manage=serviceaccounts` - Manage only the 'software' roles - i.e. the
   `crossplane` role and the roles for each provider.
 * `--manage=basic` - Manage the software roles plus the `crossplane-admin`,
--- a/design/design-doc-rbac-manager.md
+++ b/design/design-doc-rbac-manager.md
@@ -1,0 +1,777 @@
+# Crossplane RBAC Manager
+
+* Owner: Nic Cope (@negz)
+* Reviewers: Crossplane Maintainers
+* Status: Draft
+
+## Background
+
+Crossplane, as a project, consists of three key building blocks:
+
+1. A provider is a controller manager that can orchestrate an external system
+   (like a cloud provider). Each external system primitive is represented as a
+   'managed resource' - a kind of custom resource.
+1. The API extensions controllers allow Crossplane to compose managed resources
+   into higher level 'composite resources'.
+1. The package manager controllers allow Crossplane to fetch and install a
+   provider or a configuration of composition from a package - an OCI image
+   containing metadata and custom resources.
+
+Each provider is a distinct process that is typically deployed as a pod. The API
+extensions and package manager controllers are part of the 'core' Crossplane
+controller manager process. The core controller manager is therefore responsible
+for _extending Crossplane_. Its controllers add and and remove Custom Resource
+Definitions (CRDs) to and from the API server. The core Crossplane controllers
+define custom resources (CRs) that represent:
+
+* Provider configuration.
+* The managed resources a provider may orchestrate within an external system.
+* Composite resources and composite resource claims.
+
+These controllers are novel in that they define custom resources and start
+controllers in response to the creation of other custom resources. Creating a
+`Provider` custom resource, for example, would cause the package manager to
+deploy a new provider controller manager, and to define the custom resources
+that it can reconcile.
+
+When Crossplane defines new custom resources some entities (or subjects, in
+Kubernetes RBAC terminology) must be granted access to those resources in order
+for them to be useful. These subjects can be categorised as software, or humans.
+
+Software, in this context, corresponds to the service accounts that Crossplane
+or a provider run as. When the package manager reconciles a `Provider`, it runs
+the provider pod as a certain service account. The provider cannot reconcile its
+managed resources unless its service account is granted access to do so.
+Furthermore, the core Crossplane controllers cannot create the provider's
+managed resources, and thus cannot compose them into a higher level resource.
+
+Humans are the people who use Crossplane to orchestrate external systems.
+Crossplane typically categorises these people into three roles, which may
+overlap:
+
+* Platform builders define composite resources and compositions. They may
+  package these composite resources and compositions as configurations.
+* Platform operators install providers, install configurations, and manage
+  composite resources. They may also interact with the managed resources of
+  which a composite resource is composed.
+* Platform consumers claim composite resources. The platform builder specifies
+  which composite resources are offered to platform consumers.
+
+These people are unable to configure and use Crossplane to orchestrate the
+systems they wish to orchestrate unless they (i.e. their RBAC user or group) are
+granted access to perform the tasks they need to perform. Furthermore, no one is
+able to grant this access because the API server requires that a subject be
+granted a particular RBAC role before they are able to grant that role to
+another subject.
+
+In either case - software or human - it is possible for a superuser to manually
+author the appropriate RBAC roles and bind them to the appropriate subjects to
+make Crossplane work. A superuser in this context is either a subject bound to
+the built in `cluster-admin` RBAC `ClusterRole`, or a subject bound to a role
+that grants the `escalate` verb on RBAC `Roles` and/or `ClusterRoles`. Manually
+authoring and binding RBAC roles is the most flexible and secure approach, but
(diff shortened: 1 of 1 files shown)


design/design-doc-rbac-manager.md
```diff
@@ -0,0 +1,706 @@
...
+roles, using [`ClusterRole` aggregation][Aggregated ClusterRoles] to form a set
+of cluster roles that may change over time as providers and composite resources
+are added to and removed from the API server. The RBAC manager would manage the
+following cluster roles. Note that these roles are aggregated - they consist of
+several more granular roles that are not covered here. **The following examples
+are 'flattened' aggregations represented as a single fixed `ClusterRole` to
+demonstrate specifically what permissions each role will have.**
+
+The Crossplane `ClusterRole` is bound to the service account the Crossplane pod
+runs as. It grants the following rules:
```
hasheddan, line 140: We will also need the following permissions for the existing package manager, as well as the new package manager (which we may be able to be paired down after removal of the existing):
```yaml
rules:
  - apiGroups: ["extensions", "apps"]
    resources: ["deployments"]
    verbs: ["get", "list", "create", "update", "patch", "delete", "deletecollection"]
  - apiGroups: ["batch", "extensions"]
    resources: ["jobs"]
    verbs: ["get", "list", "create", "update", "patch", "delete", "deletecollection"]
  - apiGroups: [""]
    resources: ["secrets"]
    verbs: ["get", "list", "create", "update", "patch", "delete"]
  - apiGroups: [""]
    resources: ["pods"]
    verbs: ["get", "list"]
  - apiGroups: [""]
    resources: ["pods/log"]
    verbs: ["get"]
```
However, these are only needed for a `Role` and can be bound with a `RoleBinding` as the package manager only operates in the namespace Crossplane is running in. Note that these are the [same privileges the package manager has in hosted mode](https://github.com/crossplane/crossplane/blob/master/cluster/charts/crossplane-controllers/templates/package-manager-host-role.yaml). The package manager also will need cluster-scoped permissions, but that is included in the `ClusterRole` you defined below :+1:

design/design-doc-rbac-manager.md
```diff
@@ -0,0 +1,706 @@
...
+      rbac.crossplane.io/aggregate-to-namespace-edit: "true"
+      rbac.crossplane.io/xrd: examplecomposites.xr.example.org
+```
+
+### Provider RBAC Mechanics
+
+To create and bind a cluster role for a `Provider` (package):
+
+1. Find its active `ProviderRevision`.
+1. List all CRDs, filter down to those controlled by the active revision.
```
hasheddan, line 605: Just adding a note from out meeting that I would like for this to be discoverable on the revision itself, but it works either way

design/design-doc-rbac-manager.md
```diff
@@ -0,0 +1,706 @@
...
+1. Create cluster roles that grant RBAC access to said CRDs.
+1. List all service accounts, filter them similarly. There can be only one.
```
hasheddan, line 607: I think we will name the SA the same name as the revision, so you may not need to list them all

hasheddan (commented):
Thanks for this write-up @negz! I was thinking about how as long as the existing package manager is present that we will still be running it as privileged, but I think a simple solution would be to keep the existing package manager as its own `Deployment`, but make it optional, then build the new package manager into the existing Crossplane `Deployment`. Then we could eventually just remove the existing package manager altogether after a few cycles of deprecation. How does that sound?

design/design-doc-rbac-manager.md
negz: Yeah, this occurred to me this morning as well. I like your idea of keeping the existing package manager as a distinct deployment, rolling the new package manager into the core deployment, and granting the core deployment the access it needs (jobs, deployments, pods, etc).

design/design-doc-rbac-manager.md
negz: What do you think about naming it `crossplane:provider:7f63e3661` (where `7f63e3661` is the revision name) to match the prefix of the RBAC roles the RBAC manager would generate for it?

design/design-doc-rbac-manager.md
negz: I've updated this role; let me know how it looks. I removed a few things like `deletecollection` and `list` that I suspect aren't really used, but I could be wrong.

negz:
I think this design is ready for review - I've moved it out of draft.

design/design-doc-rbac-manager.md
hasheddan: That sounds fine to me, although we can't include `:` as SA's must adhere to DNS-1123 subdomain naming. We could do `crossplane-provider-7f63e3661` or `crossplane.provider.7f63e3661` though.

design/design-doc-rbac-manager.md
```diff
@@ -0,0 +1,777 @@
...
+- apiGroups: [""]
+  resources: [events]
+  verbs: [get, list, watch]
+# Crossplane namespace editors may need to read or otherwise interact with
+# resource claim connection secrets.
+- apiGroups: [""]
+  resources: [secrets]
+  verbs: ["*"]
+# Crossplane editors have read-only access to composite resource definitions and
+#  compositions. This allows them to discover and select an appropriate
```
hasheddan, line 415:
```suggestion
# compositions. This allows them to discover and select an appropriate
```

design/design-doc-rbac-manager.md
```diff
@@ -0,0 +1,777 @@
...
+    rbac.crossplane.io/xrd: examplecomposites.xr.example.org
+rules:
+- apiGroups: [xr.example.org]
+  resources:
+  - exampleclaims
+  - exampleclaims/status
+  # Composite resources are cluster scoped. Binding this cluster role at the
+  # namespace scope will not actually grant access to these composite resources.
+  # Including them in this cluster role allows one crossplane:composite role to
+  # aggregate to both the crossplane-ns-example-edit and crossplane-edit roles.
```
hasheddan, line 639: This would be deferred to a Crossplane user to build right?

hasheddan (commented):
Added a few comments below, but the overall design LGTM

design/design-doc-rbac-manager.md
negz: Ah, fair enough. This was just a minor optimisation so I'm happy to go with whatever makes the most sense to you. Using just the revision name seems fine to me. If we do go for a prefix I have a slight preference for hyphen separators.

design/design-doc-rbac-manager.md
```diff
@@ -0,0 +1,777 @@
...
+  name: example-provider
+rules:
+# Providers create events to report their status.
+- apiGroups: [""]
+  resources: [events]
+  verbs: [create]
+# Providers read their credentials from secrets.
+- apiGroups: [""]
+  resources: [secrets]
+  verbs: [get]
```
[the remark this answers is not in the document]
negz: Good point, they do. Will update.

design/design-doc-rbac-manager.md
```diff
@@ -0,0 +1,777 @@
...
+kind: ClusterRoleBinding
+metadata:
+  name: crossplane:provider:7f63e3661:system
+roleRef:
+  apiVersion: rbac.authorization.k8s.io/v1
+  kind: ClusterRole
+  name: crossplane:provider:7f63e3661:system
+subjects:
+- kind: ServiceAccount
+  name: crossplane:provider:7f63e3661
```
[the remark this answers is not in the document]
negz: Any preference on whether I name this just `7f63e3661` or `crossplane-provider-7f63e3661`? (Or something else?)

design/design-doc-rbac-manager.md
```diff
@@ -0,0 +1,777 @@
...
+- apiGroups: [provider.example.org]
+  resources:
+  - examplemanageds
+  - exampleproviderconfigs
+  verbs: ["*"]
+# Crossplane uses jobs to unpack packages.
+- apiGroups: [batch, extensions]
+  resources: [jobs]
+  verbs: [get, create, update, delete]
+# Crossplane reads pod logs to unpack packages.
```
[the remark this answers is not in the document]
negz: Yeah I think this got lost when I moved some things around; originally this comment applied to both the the pod rules. Will fix.

design/design-doc-rbac-manager.md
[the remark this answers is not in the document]
hasheddan: I am currently just naming it the hash, but it may be confusing if someone is looking at all of the SA's present in their cluster to just see a bunch of long hash names. I like `crossplane-provider-*`. If we adjust to a different name we can always circle back and update here :+1:

design/design-doc-rbac-manager.md
```diff
@@ -0,0 +1,777 @@
...
+  resources: [secrets]
+  verbs: [get]
+# Providers have access to the resources they define.
+- apiGroups: [provider.example.org]
+  resources:
+  - examplemanageds
+  - examplemanageds/status
+  - exampleproviderconfigs
+  - exampleproviderconfigs/status
+  verbs: [get, list, watch, update, patch]
```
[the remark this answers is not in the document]
negz: My goal was to run providers with the least privileges they need to operate, and I don't think they need to be able to create or delete their managed resources or providers.

That said, I also tried to strike a balance between running with least privilege and writing super granular RBAC roles. For example if I was strictly going for least privilege I could remove all but RO access to provider configs, and remove Crossplane's ability to create RBAC roles.

design/design-doc-rbac-manager.md
```diff
@@ -0,0 +1,777 @@
...
+- apiGroups: [""]
+  resources: [pods]
+  verbs: [get, list]
+# Crossplane uses deployments to run providers.
+- apiGroups: [extensions, apps]
+  resources: [deployments]
+  verbs: [get, create, update, delete]
+- apiGroups: [""]
+  resources: [pods/log]
+  verbs: [get]
```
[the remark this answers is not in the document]
negz: Agreed. I'll break the `Role` out in this example to make things a little clearer. I might gloss over the differences between regular and hosted mode though - if I understand correctly the RBAC manager can be ignorant of those differences, which are handled by the Helm chart.

jbw976 (commented):
This is a great write-up @negz! Should we also update https://github.com/crossplane/crossplane/blob/master/design/one-pager-packages-security-isolation.md to indicate that functionality is now obsolete (or at least out of date) and to now point to this document from there as the source of truth?

negz:
> Should we also update https://github.com/crossplane/crossplane/blob/master/design/one-pager-packages-security-isolation.md to indicate that functionality is now obsolete (or at least out of date) and to now point to this document from there as the source of truth?

I don't think we have a super consistent approach to old design docs, but I agree that we should at least mark it `Status: Defunct`. I'd prefer to do that in a separate PR since we actually have quite a few docs that should be marked defunct at this stage. I've added https://github.com/crossplane/crossplane/issues/1711 to the 0.13 milestone to make sure we do that.

design/design-doc-rbac-manager.md
```diff
@@ -0,0 +1,777 @@
...
+  resources:
+  - examplemanageds
+  - exampleproviderconfigs
+  verbs: [get, list, watch]
+```
+
+The RBAC manager creates `edit` and `view` cluster roles for each namespace.
+These cluster roles are 'namespace aligned' - they are intended to grant access
+to a particular namespace via a `RoleBinding` - but not namespace scoped. This
+is because (while these examples are flattened into single roles) they use role
```
muvaf, line 376: Is the namespace-specific aggregation the only reason? Do you expect those roles to have cluster-scoped permissions at some point? (I'm not sure if we discussed this, probably did). If you don't expect cluster-scoped permissions to be added in the future, maybe RBAC manager could do the aggregation manually, i.e. look at annotations on the `Namespace` and update `Role`s in that namespace.

design/design-doc-rbac-manager.md
```diff
@@ -0,0 +1,777 @@
...
+encounters a namespace with one or more such annotations, it creates `edit` and
+`view` cluster roles for that namespace.
+
+```yaml
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: example
+  annotations:
+    rbac.crossplane.io/composites.example.org: xrd-enabled
```
muvaf, line 590: Nit: maybe something with `offer` owrd, like `xrd-offered`?

design/design-doc-rbac-manager.md
```diff
@@ -0,0 +1,777 @@
...
+
+### Opting out of RBAC Management
+
+Building and deploying the RBAC manager as a distinct process allows Crossplane
+users to completely opt-out of RBAC management. Users may choose not to deploy
+the RBAC manager without compromising any core Crossplane functionality. In
+addition to completely disabling the RBAC manager it will be possible to limit
+the scope of the RBAC manager via a flag. The following levels will exist:
+
+* `--manage=all` - Include all functionality described in this design.
```
muvaf, line 734:
```suggestion
* `--manage=all` - Include all functionality described in this design. This is the default.
```

`--manage=all` will be the default, right?

design/design-doc-rbac-manager.md
muvaf, line 590: Nit: maybe something with `offer` owrd, like `xrd-offered`?

muvaf (approved):
LGTM! I have a suggestion about doing the aggregation manually but I have a feeling that we already discussed it in one of many of our discussion, though I don't remember the arguments. So, if that's the case, maybe adding a few lines about it in the `Alternatives Considered` could be good.

design/design-doc-rbac-manager.md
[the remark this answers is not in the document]
hasheddan: Yep that's correct :+1:

design/design-doc-rbac-manager.md
negz: I don't like using the term offered here - it conflates the act of creating a namespaced CRD with the act of managing RBAC roles that all that defined CR to be used in a namespace. "Offers a claim" is a platform builder concern, while "enables a claim in a namespace" is a platform operator concern.

design/design-doc-rbac-manager.md
negz: Right.

design/design-doc-rbac-manager.md
negz: I could also go for something like `rbac.crossplane.io/composites.example.org: claim-accepted`. I feel like this would fit pretty well, i.e.:

1. Platform builder offers a claim for this XRD.
1. Platform operator accepts the claim in this namespace.

negz: I think this comment is a dupe - resolving.

design/design-doc-rbac-manager.md
negz: Sorry, I'm not quite sure what you're asking.

design/design-doc-rbac-manager.md
negz: > Is the namespace-specific aggregation the only reason? Do you expect those roles to have cluster-scoped permissions at some point.

It is, and I don't.

> maybe RBAC manager could do the aggregation manually

I was on the fence about this too. On one hand I appreciate and share the desire to use a namespaced `Role` to represent a role that applies to a specific namespace. On the other hand I'm reluctant to 'reinvent' aggregation functionality that already exists. I might sketch out what it would look like if we used a `Role` after I address the other comments in this doc.

design/design-doc-rbac-manager.md
```diff
@@ -0,0 +1,789 @@
...
+# resource claim connection secrets.
+- apiGroups: [""]
+  resources: [secrets]
+  verbs: ["*"]
+# Crossplane editors have read-only access to composite resource definitions and
+# compositions. This allows them to discover and select an appropriate
+# composition when creating a resource claim.
+- apiGroups: [apiextensions.crossplane.io]
+  resources: ["*"]
+  verbs: [get, list, watch]
```
negz, line 431: On reflection I'm fairly sure this won't work. XRDs and Compositions are cluster scoped, and I believe rules pertaining to cluster scoped resources are ignored when a `ClusterRole` is bound to a subject in a particular namespace via a `RoleBinding`.

negz: @thephred I've introduced a new `crossplane-browse` cluster role to address this. I'm hopeful that this could be integrated into Upbound Cloud either by allowing granting 'browse' (along with admin, edit, or view) at the cluster scope or by automatically granting 'browse' at the cluster scope to anyone granted `edit` at the namespace scope.

design/design-doc-rbac-manager.md
negz: Take a look at the latest commit - I decided to go with having the RBAC manager manage namespaced roles by emulating/approximating an opinionated form of cluster role aggregation. I went back and forth a lot, but it seems like it's not really any more difficult to manually aggregate RBAC role rules (as opposed to manually managing aggregation rules), and it feels a little more 'idiomatic' than creating a cluster role for each namespace.

design/design-doc-rbac-manager.md
```diff
@@ -0,0 +1,843 @@
...
+- apiGroups: [""]
+  resources: [pods]
+  verbs: [get, list]
+- apiGroups: [""]
+  resources: [pods/log]
+  verbs: [get]
+# Crossplane uses deployments to run providers.
+- apiGroups: [extensions, apps]
+  resources: [deployments]
+  verbs: [get, create, update, delete]
```
hasheddan, line 205: I have ended up needing some additional permissions here (such as `watch`), but these could change depending on final details of package manager implementation, so I would defer to me updating this after that is merged.

design/design-doc-rbac-manager.md
```diff
@@ -0,0 +1,843 @@
...
+# Crossplane uses a fixed (at runtime) namespace to unpack packages and run
+# providers. This role allows it to do so. Note that it is included for
+# illustrative purposes; it is static and is thus managed by the Crossplane Helm
+# chart rather than the RBAC manager.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  namespace: crossplane-system
+  name: crossplane
+rules:
```
hasheddan, line 190: nitpick: also need to be able to create SA, but this can be updated later. I also considered if we might want the RBAC controller to create the SA for a `ProviderRevision`, then we could watch for its existence and know that when it existed the correct ClusterRole would already be bound and we would be safe to create the Deployment :thinking:

design/design-doc-rbac-manager.md
```diff
@@ -0,0 +1,843 @@
...
+```
+
+The `crossplane-browse` role allows read-only access to composite resources and
+compositions.
+
+```yaml
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: crossplane-view
```
hasheddan, line 401:
```suggestion
  name: crossplane-browse
```

design/design-doc-rbac-manager.md
```diff
@@ -0,0 +1,843 @@
...
+# Crossplane browsers have read-only access to compositions and XRDs. This
+# allows them to discover and select an appropriate composition when creating a
+# resource claim.
+- apiGroups:
+  - apiextensions.crossplane.io
+  resources: ["*"]
+  verbs: [get, list, watch]
+# Crossplane browsers have read-only access to all of the composite resources it
+# defines. This allows them to claim a statically provisioned composite
+# resource.
```
hasheddan, line 416: I admittedly don't have as much knowledge of the `browse` role as the others, but I wonder if we would want a separate `ClusterRole` for each namespace such that a user can only "browse" the cluster-scoped types for which their is a claim type enabled for their namespace. Or maybe that is the intention here and I am misinterpreting? I would also be comfortable saying that this would be nice but the complexity can be deferred for the time-being.

hasheddan (approved):
One change needed, the rest are just comments that may or may not require changes. This doc will likely need some clean up after implementation is complete, so I am going ahead and giving approval :heavy_check_mark: Thanks for describing this complex controller in a digestible manner @negz!

design/design-doc-rbac-manager.md
muvaf: Nice! I love that we're able to do that. There will be some complexity added to the controller code but I believe that's manageable given the cost of the alternative. Thanks!

thephred:
LGTM

I believe that a user that has been given rights to view/edit a CompositeResource claim should be able to see the CompositeResource that is created on its behalf as well as list/view CompositeResourceDefinitions.

design/design-doc-rbac-manager.md
```diff
@@ -0,0 +1,843 @@
...
+  - examplecomposites
+  - examplecomposites/status
+  verbs: ["*"]
+```
+
+Composite resources may be claimed in any namespace that has an annotation with
+the key `rbac.crossplane.io/composites.example.org`, set to `xrd-claim-accepted`
+(where `composites.example.org` is the name of the XRD). When the RBAC manager
+encounters a namespace with one or more such annotations, it creates `admin`,
+`edit` and `view` roles in that namespace.
```
lukeweber, line 636: Suggestion: If one happens to uninstall or upgrade packages these annotations might become out of date. As well, if there's a typo, it wouldn't be obvious that it might fail to provision permissions for a type, or that it was an out of date permission grant.

For someone building a UI around this like Upbound cloud, listing a particular Kind, and seeing the status of whether the permission could be applied, or if it's deprecated/no longer existing would be a lot easier to work with, than a totally different component reading the namespace annotations, parsing the xrd out of them, and comparing these names with what's installed on the server to know what's still valid.

lukeweber (approved):
Generally looks good to me. One small comment that I think a separate type is probably more useful for Upbound for building a UI around these permissions as the RBAC controller could give useful feedback, but we can make it work either way.

design/design-doc-rbac-manager.md
negz: @lukeweber Could you sketch me some YAML representing the resource you're thinking of?

lukeweber: The only point was just that the calculation of whether you were able to find such an xrd will technically already be known by the RBAC controller and might make for a better DX, but as you noted it's another type. I'll defer to your judgement here. Just a suggestion :)

```yaml
apiVersion: rbac.authorization.crossplane.io/v1
kind: XRDGrant
metadata:
  name: composites.example.org
  namespace: foo
spec:
  xrd: composites.example.org
status:
  phases: ready , failed
  conditions: xrdMissing, createdRBAC, noPermissions
```

negz: Got it. It seems like the key value here is that we could use the new type's status to reflect whether the intent to accept/grant a claim in a namespace was successful. I'd like to stick with the current design for now, but we can introduce this type if annotations prove to be a poor UX. I believe it's fairly common in Kubernetes for something to start as an annotation and later grow into a real field/type.

design/design-doc-rbac-manager.md
negz: Yeah, let's avoid that complexity for now. I'd like to avoid generating "namespace aligned ClusterRoles" if possible. I'm hopeful that this broad 'browse' role (or even the view role) will be appropriate for 80% of folks - keep in mind that there's always the fallback option of authoring and granting your own roles.

design/design-doc-rbac-manager.md
negz: I'll add list and watch for now. I agree we can update this later to more accurately reflect reality as we understand it better.

design/design-doc-rbac-manager.md
negz: I've updated these rules to add create SA.

Regarding the RBAC manager creating SAs, I think the separation of concerns is perhaps a little cleaner if the package manager owns that task. That way you can still install a and run provider if you're not running the RBAC manager - you just have to make sure some other entity grants RBAC access (or that RBAC is disabled).

PR #1713, opened 2020-09-08
Configurable readiness probe for composed resources

### LINKED ISSUE #1687: Make composed resource readiness detection configurable

Composite resources were designed to compose managed resources, and other composite resources. In practice the composite controller expects that the resources it composes:

* Are cluster scoped
* Can eventually become "ready", in the sense that their `type: Ready` status condition can transition to `status: True`.

The latter expectation is limiting on two fronts:

* In some cases it may be useful to include a `ProviderConfig` (fka `Provider`) in a `Composition`. For example a composition may want to create an EKS cluster and make it consumable by other providers (e.g. the [Helm provider](https://github.com/crossplane-contrib/provider-helm) or the [Rook provider](https://github.com/crossplane/provider-rook)) by composing a Kubernetes `ProviderConfig`.
* Per https://github.com/crossplane/crossplane/pull/1677 we plan to start introducing managed resources that are shims around Terraform providers. Terraform resources don't have a generic concept of 'readiness'.

Crossplane could allow the `Composition` author to describe what it means for a resource to be "ready". I suspect a reasonable approach would be:

1. If the composition specifies how to determine readiness, those conditions determine whether the resource is ready.
1. If the resource has a `Ready` status condition and it's not `True`, the resource is not ready.
1. If the resource has a `Ready` status condition and it is `True`, the resource is ready.
1. If the resource has no `Ready` status condition but `Synced` status condition is `True`, the resource is ready.

I imagine a `Composition` would allow authors to specify whether a resource is ready via something like:

```yaml
kind: Composition
metadata:
  name: clusters.cnp.example.org
spec:
  compositeTypeRef:
    kind: CoolComposite
  resources:
  - base:
      kind: CoolManaged
    patches:
    - fromFieldPath: metadata.annotations[crossplane.io/external-name]
      toFieldPath: metadata.annotations[crossplane.io/external-name]
    readiness:
    - fieldPath: status.atProvider.state
      type: Match
      match: "ONLINE"
```

muvaf:

This PR introduces `readinessProbe` field to base templates in `Composition` resource so that users can configure how they would like that specific composed resource should be considered ready or not.

Fixes #1687 

Fixes #

I have:

Manually.


muvaf:
Keeping it as draft until unit tests are written.

apis/apiextensions/v1alpha1/composition_types.go
```diff
@@ -87,6 +87,26 @@ type ComposedTemplate struct {
 	// resource to the composition instance connection secret.
 	// +optional
 	ConnectionDetails []ConnectionDetail `json:"connectionDetails,omitempty"`
+
+	// ReadinessProbe allows users to use a custom readiness check. The default
+	// readiness probe is to have the "Ready" condition to be "True".
+	ReadinessProbe *ReadinessProbe `json:"readinessProbe,omitempty"`
```
negz, line 93: Naming nit: "probe" to me implies that we'd be "probing" the resource whose readiness we're concerned with - i.e. doing something a little more active like running an exec or making a network request. I'd prefer something like `readinessConstraint`, `readinessRequirement`, or `readinessCheck` here.

negz: Also, I think we should support an array of these. I'm very skeptical that inspecting only one field to determine that something is ready will scale over time.

apis/apiextensions/v1alpha1/composition_types.go
```diff
@@ -87,6 +87,26 @@ type ComposedTemplate struct {
...
+// ReadinessProbe is used to indicate how to tell whether a resource is ready
+// for consumption
+type ReadinessProbe struct {
+
+	// FieldPath shows the path of the field whose value will be used.
+	FieldPath string `json:"fieldPath,omitempty"`
+
+	// Type indicates the type of probe you'd like to use.
+	// +kubebuilder:validation:Enum="Match";"NonEmpty"
+	Type string `json:"type"`
```
negz, line 105: Can we make a subtype of `string` - i.e. `ReadinessProbeType` - with the possible values as constants in this file?

pkg/controller/apiextensions/composite/composed/api.go
```diff
@@ -164,3 +167,36 @@ func (cdf *APIConnectionDetailsFetcher) Fetch(ctx context.Context, cd resource.C
 
 	return conn, nil
 }
+
+// DefaultReadinessProber is a readiness prober which returns whether the composed
+// resource is ready or not.
+type DefaultReadinessProber struct{}
```
negz, line 173: Nit: Could we use a `ReadinessProberFn`, rather than using an empty struct named `Default<T>` to satisfy our interface? It doesn't seem like this needs to be a struct.

Edit: Ah, I see this is an established convention in this file. It's not my preferred convention, but I'm happy to stick to it.

muvaf: That would've been nicer indeed. I want to defer it to later though.

muvaf:
@negz addressed all of your comments. It's ready to merge but we need https://github.com/crossplane/crossplane-runtime/pull/199 merged and update `go.mod` of this file. I've manually tested all the functionality with the composition guide we have at crossplane.io

go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/crossplane/crossplane
 
 go 1.13
 
+replace github.com/crossplane/crossplane-runtime => github.com/muvaf/crossplane-runtime v0.0.0-20200910111027-446c12e8ee41
```
negz, line 5: Please remove and update to master before merging.

negz (approved):
Thanks @muvaf this looks great. Feel free to merge once the `replace` statement has been removed.

PR #1716, opened 2020-09-09
fix: kubectl command

frezbo:
Signed-off-by: Noel Georgi <git@frezbo.com>

Fixes the kubectl command to list aws provider

### DIFF
--- a/docs/cloud-providers/aws/aws-provider.md
+++ b/docs/cloud-providers/aws/aws-provider.md
@@ -41,7 +41,7 @@ resources.
 You can confirm the existense of the  AWS `ProviderConfig` by running:
 
 ```bash
-kubectl -n crossplane-system get provider/aws-provider
+kubectl -n crossplane-system get provider aws-provider
 ```
 
 ## Optional: Setup AWS Provider Manually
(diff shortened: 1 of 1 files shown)


docs/cloud-providers/aws/aws-provider.md
```diff
@@ -41,7 +41,7 @@ resources.
 You can confirm the existense of the  AWS `ProviderConfig` by running:
 
 ```bash
-kubectl -n crossplane-system get provider/aws-provider
+kubectl -n crossplane-system get provider aws-provider
```
hasheddan, line 44: Thanks for looking into this! I _think_ this might should actually look like the following with recent updates:
```suggestion
kubectl get providerconfig aws-provider
```

hasheddan (changes requested):

docs/cloud-providers/aws/aws-provider.md
frezbo: @hasheddan Seems it's still not in the alpha release. I'm on `0.12.0`

```bash
❯ kubectl get providerconfig aws-provider
error: the server doesn't have a resource type "providerconfig"
```

hasheddan: Yep, this commit is to master though, meaning it will be in the next release. Also, you’ll have to manually add DCO to that commit because the Github UI doesn’t do it for you :/

frezbo: yup, fixed 👍

hasheddan (approved):

hasheddan:
This doc (and some others) need additional clean up, but we can move forward with this fix as it is an improvement 👍

PR #1721, opened 2020-09-10
Remove 'Synced' status condition from apiextensions controllers

negz:

See https://github.com/crossplane/crossplane-runtime/issues/198 for context. This came to mind when I started adding a third controller that reconciled XRDs (by creating RBAC roles). I noticed that the two existing controllers were fighting over the synced status condition, and that said condition was now redundant.

I have:


pkg/controller/apiextensions/claim/reconciler.go
```diff
@@ -303,8 +303,7 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error)
 					// after a brief wait, in case this was a transient error.
 					log.Debug("Cannot remove finalizer", "error", err, "requeue-after", time.Now().Add(aShortWait))
 					record.Event(cm, event.Warning(reasonDelete, err))
-					cm.SetConditions(v1alpha1.Deleting(), v1alpha1.ReconcileError(err))
```
hasheddan, line 306: I noticed there are a few places where you removed setting the `Deleting` condition also, but didn't notice that explicitly called out in the PR or the issue (may have missed it). Do you think we the event / log in deleting scenarios should replace setting this condition?

negz: Good catch - I forgot to mention this. It's another case where it probably doesn't make sense to have N controllers (composite, claim, RBAC, etc) all trying to set the one condition - in this case `Deleting()` tries to set `Established: False` in both XRD controllers.

We could address this by using two different conditions - perhaps `Terminating()` (like CRD) for the definition controller that's cleaning up the composite's CRD, and `Redacting()` for the offered controller that's cleaning up the claim's CRD. I opted to take the lazy way out and just not set a condition when we're being deleted, because we're already emitting an event and because in the happy path the XRD won't stick around very long after being deleted. Do you think there's value in keeping these "deleting" conditions?

hasheddan: I like the two different conditions here I think, but am kind of indifferent on whether they need to be part of this PR.

muvaf:
> This came to mind when I started adding a third controller that reconciled XRDs (by creating RBAC roles). I noticed that the two existing controllers were fighting over the synced status condition, and that said condition was now redundant.

FWIW, I had the same problem of `Synced` being used by multiple controllers in Crossplane agent. I went for making each controller that want to write a condition to be the exclusive owner of it. The agent has `AgentSynced` type, so, for example we can have different kind for RBAC manager or offered XRD controller, and that granularity should help debugging, too.

negz:
> I went for making each controller that want to write a condition to be the exclusive owner of it.

I initially started going down a path like that - e.g. having something like `CompositeSynced`, `ClaimSynced`, and `RBACSynced` for the XRD controllers. These names didn't feel right though, because it's not really the case that "the composite is synced". It's more like "the controller that handles creating composite resources reconciled without an error". I'm still feeling pretty confident that we don't need a 'synced' condition per https://github.com/crossplane/crossplane-runtime/issues/198.

hasheddan (approved):
I think this works well with the split conditions. I wonder if we could just confine synced to resources that talk to APIs outside of the cluster? Anyway, we can discuss on the PR in crossplane-runtime

PR #1734, opened 2020-09-16
Unpack tool

muvaf:

I started working on this as a command for `crank` but I believe it'll have its own repo. I wrote some parts of the current unpack tool that does metadata operations but I believe our approach will be different than the existing mechanism. So, I removed those parts.

Right now, the PR is pretty lean. It just reads and emits after I removed the other stuff. @hasheddan is this all the functionality we need from this tool? I'd like to make sure the functionality is there and then I'll create a new repo or move it out of `crank` command.

Fixes #

I have:


hasheddan (commented):
@muvaf thanks for working on this! I agree that it should be moved out to a separate repo and part of the review there will be on how the base image is built for the binary.

One thing to go ahead and mention here though is that I am somewhat skeptical that we should actually parse and validate the package contents in the unpack image. Theoretically, we would be even more lean here if we just literally walk the fs and emit everything. I think if we included a flag to emit metadata in a specific dir in the future then we would obviously skip that dir at the main entrypoint, but otherwise I think we may want to emit everything. The motivation for this would be to force a package author to explicitly exclude things (via the `ignore` stanza) when building their package, which in turn would lead to more predictable behavior from the install process. For example, if the Crossplane package manager started accepting a new package type, let's call it `Policy` for now, I would prefer that users would not have to update the base image they build with to be able to create a `Policy` package.

The primary argument against this is that it would be more likely for users to build a package that the package manager will not accept, but doing the parsing in the unpack image and returning an error doesn't actually help with this very much (i.e. you would still have to install the package and the package manager would just fail because unpack failed rather than getting some YAML or other file it won't accept). I think the best option here is to provide a `lint` command in `crank` that invokes a previously built package container and does the validation on its output. This decouples the unpack image from depending on Crossplane and defers that to `crank`, which is built and versioned with Crossplane anyway.

How does that sound to you?

muvaf:
> One thing to go ahead and mention here though is that I am somewhat skeptical that we should actually parse and validate the package contents in the unpack image. Theoretically, we would be even more lean here if we just literally walk the fs and emit everything. I think if we included a flag to emit metadata in a specific dir in the future then we would obviously skip that dir at the main entrypoint, but otherwise I think we may want to emit everything.

If I understand correctly, you mean emitting all files under `/` dir right? If so, I'm really not sure how feasible this is. Even alpine image has some dirs that are completely unrelated. So, people will have to be careful about that stuff. What I was thinking is to have the path `/crossplane` as default and if anyone else wants to use a different path then they could pass it in as a flag.

hasheddan:
> If I understand correctly, you mean emitting all files under / dir right? If so, I'm really not sure how feasible this is. Even alpine image has some dirs that are completely unrelated.

The `WORKDIR` can be set in this base image. `crank pkg build` will copy files to the working dir and then this base image will emit everything there. This is good behavior for `crank` too because we are saying that if you want to use a different base image, it should set the `WORKDIR` to the directory in that image it want the package contents to be in. Does that sound right to you?

muvaf:
> The WORKDIR can be set in this base image.

I'd have preferred a more opinionated approach for such low level details, but yeah that'd work, too.

> Theoretically, we would be even more lean here if we just literally walk the fs and emit everything.

At that point, do we even need a tool? Is it not simply `cat *`? So maybe a simple `Dockerfile` in crank repo to be used as base image would be sufficient?

hasheddan:
> At that point, do we even need a tool? Is it not simply cat *? So maybe a simple Dockerfile in crank repo to be used as base image would be sufficient?

We could start with that, but it does not have the logic to handle passing flags and changing behavior, so we would likely need a very small binary in the future if we didn't create one now.

muvaf:
Closing in favor of https://github.com/crossplane-contrib/pkg-unpack/pull/1

PR #1741, opened 2020-09-17
Remove ignore stanza and permission requests from pkg meta types

hasheddan:

File paths will be ignored by specifying a flag during build such that
the pkg meta types are only consumed at runtime by the package manager
and do not inform any phases of the build process.

The first iteration of the RBAC manager will not honor permissions
requested in a provider package, so we defer introducing the field as a
valid option until it is supported

I have:

`make generate`
`make reviewable`
`make build`


kasey (approved):
lgtm!

negz (approved):

PR #1746, opened 2020-09-18
Build binaries for darwin and windows

hasheddan:

Adds support for building and publishing go binaries for darwin and
windows, which is now more important that the crossplane CLI is built
within this repo. Note that this will not affect images that are built.

I have:

`make build`

### DIFF
--- a/cluster/images/crossplane/Makefile
+++ b/cluster/images/crossplane/Makefile
@@ -12,6 +12,10 @@ include ../../../build/makelib/image.mk
 # ====================================================================================
 # Targets
 
+ifeq ($(PLATFORM),$(filter $(PLATFORM),darwin_amd64 windows_amd64)) 
+$(info Skipping image build for $(PLATFORM))
+img.build:
+else
 img.build:
 	@$(INFO) docker build $(IMAGE)
 	@cp Dockerfile $(IMAGE_TEMP_DIR) || $(FAIL)
@@ -23,3 +27,4 @@ img.build:
 		-t $(IMAGE) \
 		$(IMAGE_TEMP_DIR) || $(FAIL)
 	@$(OK) docker build $(IMAGE)
+endif
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@
 PROJECT_NAME := crossplane
 PROJECT_REPO := github.com/crossplane/$(PROJECT_NAME)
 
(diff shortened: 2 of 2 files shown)


hasheddan:
oof.. looks like `img.build` is called as part of `build.all` --> `do.build.artifacts`, so the include we have in the image `Makefile` does not preclude the platforms that are specified in the main `Makefile` but not the image `Makefile`. We could skip explicitly in the `img.build` if we don't want to build for specific platforms..

cluster/images/crossplane/Makefile
```diff
@@ -12,6 +12,10 @@ include ../../../build/makelib/image.mk
 # ====================================================================================
 # Targets
 
+ifeq ($(PLATFORM),$(filter $(PLATFORM),darwin_amd64 windows_amd64)) 
+$(info Skipping image build for $(PLATFORM))
+img.build:
+else
```
hasheddan, line 18: I don't _love_ this, but I think I actually prefer it to doing a larger refactor in the `build` submodule or trying to work around it in our main `Makefile`. You can see that the correct [images](https://releases.crossplane.io/build/PR-1746/v0.13.0-rc.153.gd2399b5/images/) and [binaries](https://releases.crossplane.io/build/PR-1746/v0.13.0-rc.153.gd2399b5/bin/) have been built for this PR.

muvaf:
Why do we need these targets?

cluster/images/crossplane/Makefile
muvaf: I don't really know the details of `Makefile` syntax. Is it possible to have the if/else under `img.build` and print `Skipping image build since the target is ...` in case it's skipped?

hasheddan: It is possible to exit earlier, but from what I could tell while trying to determine the best way to do this, it is considered better practice to override the target

hasheddan:
> It seems that the target list is for all builds happening from this repo.

@muvaf do you mean that the crossplane binary also is built for each of the 4 platforms? If so, I think that is fine.

jbw976 (approved):
I reviewed the build submodule flow again @hasheddan and this looks like a reasonable way to have separate PLATFORMS for binaries and images, without having to add new logic to the build submodule.  LGTM :)

PR #1748, opened 2020-09-20
Allow the RBAC Manager to manage provider roles and bindings

### LINKED ISSUE #1637: Generate RBAC roles for composite resources and claims

A Crossplane `InfrastructureDefinition` defines a new API - a new CR that Crossplane may reconcile by composing other CRs. An `InfrastructurePublication` publishes this API at the namespace scope such that applications can easily consume it. Currently these concepts are blissfully ignorant of RBAC; a platform builder must first define and publish their API, then create and bind RBAC roles such that folks can actually use that API. I suspect the RBAC roles platform builders would define would often (but not always) be fairly standard; perhaps an RBAC role that granted `*` verbs to the newly defined resources, and an RBAC role that granted read-only verbs to the newly defined resources.

Crossplane could automatically create (but not bind) RBAC roles when new APIs were defined. For example if an `InfrastructureDefinition` defined a new `mysqlinstance.example.org` composite resource, Crossplane might automatically create a `mysqlinstance.example.org-editor` and a `mysqlinstance.example.org-viewer` `ClusterRole` that granted RO and RW access to the newly defined API. Perhaps these roles could be aggregated up to a general `crossplane-admin` superuser role.

This would leave the decision of whether and how to actually _grant_ these roles up to the folks deploying Crossplane, but would automate the need to actually author roles for a lot of people. It would emulate the experience of the "baked in" IAM roles that cloud providers provide, while still allowing platform builders to use RBAC to author and bind bespoke RBAC roles if the generated roles did not fit their needs.

negz:

Fixes #1637

This PR follows onto https://github.com/crossplane/crossplane/pull/1728 by allowing the RBAC manager (when using either the Basic or All management policy) to:

* Create edit, view, and system cluster roles for each provider installed via the package manager.
* Aggregate the edit and view cluster roles to the crossplane-admin, crossplane-edit, and crossplane-view cluster roles.
* Aggregate the edit cluster role to the crossplane cluster role, so that Crossplane may compose the provider's resources.
* Bind the system cluster role to the provider deployment's service account, so that it may reconcile its resources.

Note that this implementation differs slightly from the [RBAC manager design](https://github.com/crossplane/crossplane/blob/15d496349f366685dbe1d53d8726562874722a5d/design/design-doc-rbac-manager.md), in that we create a set of roles and a binding for each provider revision, not for each provider. See the relevant commit for details.

I have:

I've built and deployed the RBAC manager on this branch and used https://github.com/hasheddan/xp-package-ports/tree/master/provider-aws to deploy v0.10, upgrade to v0.11, then return to v0.10. I ensured all roles and bindings were as expected, and that the revisions were being updated with events as expected.

### DIFF
--- a/apis/pkg/v1alpha1/revision_types.go
+++ b/apis/pkg/v1alpha1/revision_types.go
@@ -17,8 +17,6 @@ limitations under the License.
 package v1alpha1
 
 import (
-	rbac "k8s.io/api/rbac/v1"
-
 	runtimev1alpha1 "github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
 )
 
@@ -73,8 +71,4 @@ type PackageRevisionStatus struct {
 
 	// References to objects owned by PackageRevision.
 	ObjectRefs []runtimev1alpha1.TypedReference `json:"objectRefs,omitempty"`
-
-	// PermissionRequests are additional permissions that should be added to a
-	// packaged controller's service account.
-	PermissionRequests []rbac.PolicyRule `json:"permissionRequests,omitempty"`
 }
--- a/pkg/controller/pkg/manager/reconciler.go
+++ b/pkg/controller/pkg/manager/reconciler.go
@@ -48,6 +48,7 @@ const (
 const (
 	errGetPackage            = "cannot get package"
 	errListRevisions         = "cannot list revisions for package"
+	errUnpack                = "cannot unpack package"
 	errCreatePackageRevision = "cannot create package revision"
 	errGCPackageRevision     = "cannot garbage collect old package revision"
 	errGCInstallPod          = "cannot garbage collect old package revision install pod"
@@ -214,7 +215,7 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error) 
 	prs := r.newPackageRevisionList()
 	if err := r.client.List(ctx, prs, client.MatchingLabels(map[string]string{parentLabel: p.GetName()})); resource.IgnoreNotFound(err) != nil {
 		log.Debug(errListRevisions, "error", err)
-		r.record.Event(p, event.Warning(reasonList, err))
+		r.record.Event(p, event.Warning(reasonList, errors.Wrap(err, errListRevisions)))
 		return reconcile.Result{RequeueAfter: shortWait}, nil
 	}
 
@@ -222,8 +223,8 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error) 
 
 	hash, err := r.podManager.Sync(ctx, p)
 	if err != nil {
-		log.Debug(errListRevisions, "error", err)
-		r.record.Event(p, event.Warning(reasonUnpack, err))
+		log.Debug(errUnpack, "error", err)
+		r.record.Event(p, event.Warning(reasonUnpack, errors.Wrap(err, errUnpack)))
 		return reconcile.Result{RequeueAfter: shortWait}, errors.Wrap(r.client.Status().Update(ctx, p), errUpdateStatus)
 	}
 
@@ -277,7 +278,7 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error) 
 			rev.SetDesiredState(v1alpha1.PackageRevisionInactive)
 			if err := r.client.Apply(ctx, rev, resource.MustBeControllableBy(p.GetUID())); err != nil {
 				log.Debug(errUpdateInactivePackageRevision, "error", err)
-				r.record.Event(p, event.Warning(reasonTransitionRevision, err))
+				r.record.Event(p, event.Warning(reasonTransitionRevision, errors.Wrap(err, errUpdateInactivePackageRevision)))
 				return reconcile.Result{RequeueAfter: shortWait}, errors.Wrap(r.client.Status().Update(ctx, p), errUpdateStatus)
 			}
 		}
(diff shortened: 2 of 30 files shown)


apis/pkg/v1alpha1/revision_types.go
```diff
@@ -73,8 +71,4 @@ type PackageRevisionStatus struct {
 
 	// References to objects owned by PackageRevision.
 	ObjectRefs []runtimev1alpha1.TypedReference `json:"objectRefs,omitempty"`
-
-	// PermissionRequests are additional permissions that should be added to a
-	// packaged controller's service account.
-	PermissionRequests []rbac.PolicyRule `json:"permissionRequests,omitempty"`
```
hasheddan, line 79: Since we are taking the approach of removing anything that isn't currently used, do you think we should keep `Crossplane` and `DependsOn` here? I suspect they will likely both be supported in next release (v0.14) for context

pkg/controller/pkg/manager/reconciler.go
```diff
@@ -214,16 +215,16 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error)
...
 		return reconcile.Result{RequeueAfter: shortWait}, nil
 	}
 
 	p.SetConditions(v1alpha1.Unpacking())
 
 	hash, err := r.podManager.Sync(ctx, p)
 	if err != nil {
-		log.Debug(errListRevisions, "error", err)
-		r.record.Event(p, event.Warning(reasonUnpack, err))
+		log.Debug(errUnpack, "error", err)
```
hasheddan, line 226: thanks for the cleanup :)

pkg/controller/pkg/revision/reconciler.go
```diff
@@ -62,9 +62,9 @@ const (
 
 // Event reasons.
 const (
-	reasonParse   event.Reason = "ParsePackage"
-	reasonLint    event.Reason = "LintPackage"
-	reasonInstall event.Reason = "InstallPackage"
+	reasonParse  event.Reason = "ParsePackage"
+	reasonLint   event.Reason = "LintPackage"
+	reasonConfig event.Reason = "ConfigurePackage"
```
hasheddan, line 67: nitpick: I don't love `ConfigurePackage` as a condition here, especially given our overloading of the term `Configuration`. Would certainly be open to alternatives to `InstallPackage` though

pkg/controller/rbac/provider/binding/reconciler.go
```diff
@@ -0,0 +1,218 @@
...
+	l := &corev1.ServiceAccountList{}
+	if err := r.client.List(ctx, l); err != nil {
+		log.Debug(errListSAs, "error", err)
+		r.record.Event(pr, event.Warning(reasonBind, errors.Wrap(err, errListSAs)))
+		return reconcile.Result{RequeueAfter: shortWait}, nil
+	}
+
+	// Filter down to the ServiceAccounts that are owned by this
+	// ProviderRevision. Each revision should control at most one, but it's easy
+	// and relatively harmless for us to handle there being many.
```
hasheddan, line 173: Currently we are naming the SA the same name as the revision and they all exist in a single namespace for a given Crossplane deployment. Do you anticipate needing this more robust strategy for a future scenario?

pkg/controller/rbac/provider/binding/reconciler.go
```diff
@@ -0,0 +1,218 @@
...
+			}
+		}
+	}
+
+	n := roles.SystemClusterRoleName(pr.GetName())
+	ref := meta.AsController(meta.TypedReferenceTo(pr, v1alpha1.ProviderRevisionGroupVersionKind))
+	rb := &rbacv1.ClusterRoleBinding{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:            n,
+			OwnerReferences: []metav1.OwnerReference{ref},
```
hasheddan, line 192: Do we want to create these for inactive provider revisions? They will not have a controller running so they will not be made use of

hasheddan: One benefit of this strategy is that we don't have to handle any cleanup, we just defer to GC. However, we also will be keeping around some extraneous cluster roles. Does it cause an issue to create a cluster role with an SA subject that does not exist anymore?

pkg/controller/rbac/provider/roles/reconciler.go
```diff
@@ -0,0 +1,220 @@
...
+	crds := make([]v1beta1.CustomResourceDefinition, 0)
+	for _, crd := range l.Items {
+		for _, ref := range crd.GetOwnerReferences() {
+			if ref.UID == pr.GetUID() {
+				crds = append(crds, crd)
+			}
+		}
+	}
+
+	for _, cr := range r.rbac.RenderClusterRoles(pr, crds) {
```
hasheddan, line 203: I _do_ think we need the cluster roles that are created for users to view managed resource to exist for inactive provider revisions as those CRDs will continue to exist in the cluster even if they are not being reconciled :+1:

pkg/controller/rbac/provider/roles/roles.go
```diff
@@ -0,0 +1,148 @@
...
+	},
+}
+
+// SystemClusterRoleName returns the name of the 'system' cluster role - i.e.
+// the role that a provider's ServiceAccount should be bound to.
+func SystemClusterRoleName(revisionName string) string {
+	return namePrefix + revisionName + nameSuffixSystem
+}
+
+// RenderClusterRoles returns ClusterRoles for the supplied XRD.
```
hasheddan, line 72:
```suggestion
// RenderClusterRoles returns ClusterRoles for the supplied CRD.
```

hasheddan (commented):
Looking pretty good @negz! A few questions below :+1:

apis/pkg/v1alpha1/revision_types.go
negz: If they won't be functional in v0.13 I think we should remove them. I'd prefer that to be done in a separate PR, since neither of them relate to the RBAC manager.

pkg/controller/rbac/provider/binding/reconciler.go
negz: Not really - it just seemed more future proof to discover the SA rather than rely on assumptions about how other controllers behave. If listing and filtering all SAs proves to be an issue I'd be fine with changing the behaviour.

pkg/controller/rbac/provider/binding/reconciler.go
negz: > One benefit of this strategy is that we don't have to handle any cleanup, we just defer to GC.

Right - that's a big part of why I prefer this approach. I initially went with something more like what you're suggesting, but creating roles and bindings for all revisions means:

* Crossplane and the cluster scoped roles (`crossplane-admin`, etc) has access to edit the resource of an inactive provider - e.g. to cleanup finalizers before removing the revision.
* Roles and bindings work instantaneously when updating to an existing revision.
* No need to manually GC roles and bindings for revisions that still exist.

> Does it cause an issue to create a cluster role with an SA subject that does not exist anymore?

No, you can create a ClusterRoleBinding to any subject whether they exist or not.

pkg/controller/pkg/revision/reconciler.go
negz: I'm open to alternatives too - `InstallPackage` seemed inappropriate because it's sometimes used to effectively uninstall a revision - i.e. by removing a provider revision's deployment and configuration. I think elsewhere we've just used something like `ApplyPackage` if that seems any better.

hasheddan: @negz we could say something like `SyncPackage`?

hasheddan (approved):
LGTM! Thanks @negz :tada:

Issue #1670, opened 2020-08-11
Remove resource claims, classes and the package manager


negz:
### What problem are you facing?

In Crossplane 0.11 we introduced a new feature that we typically call 'composition'. This feature allows Crossplane users to define and publish their own custom resources - known as composite resources - that are composed of one or more underlying managed infrastructure resources. Composition can be used to implement the same functionality as Crossplane's resource claims and classes, but is more flexible. We deprecated resource claims and classes in Crossplane 0.12 per https://github.com/crossplane/crossplane/issues/1479.

### How could Crossplane help solve your problem?

Crossplane should eventually remove all traces of the resource claim and class pattern from its APIs and codebase. This issue will serve as a tracking issue for that work.

hasheddan:
@negz given that claims have already been deprecated (#1600), should this issue instead track their removal?

negz:
One thing that will probably be confusing during this transition period - we've decided to reuse the name 'resource claim' to replace the composition concept of a 'resource requirement'. Further context in https://github.com/crossplane/crossplane-runtime/pull/187.

muvaf:
@negz I can tackle this for 0.13 if you'd like to. I believe we have many breaking stuff already so this could be a good time for the removal.

negz:
I'd like us to consider removing these in the v0.13 release of Crossplane (and corresponding provider releases). We're aiming to make these releases on September 23rd. Claims were marked deprecated approximately three months ago in v0.12 both in the release notes and CRD docs. They're also no longer mentioned by our documentation.

It's important to note that the _functionality_ of the deprecated resource claims can be implemented using composition. @muvaf [created these XRDs](https://github.com/crossplane/crossplane/tree/master/cluster/composition/definitions), which will be usable to replace the original resource claims in v0.13.

negz:
It's been a week, and we haven't had any feedback that indicates folks would be particularly troubled by removing the resource claims in v0.13 so I'm inclined to go ahead with this removal. CC @bassam @jbw976.

bassam:
remove them !

Issue #1726, opened 2020-09-11
Allow installing Crossplane Providers in Helm chart


hasheddan:
### What problem are you facing?

Currently, Crossplane's Helm chart allows you to create a `ClusterPackageInstall` for various providers (AWS, Azure, GCP, Rook) when you `helm install` Crossplane with the `--set clusterPackages.gcp.deploy=true --set clusterPackages.gcp.version=master` flags (for example). `ClusterPackageInstall` will be replaced by `Provider` in this context in the v0.13 release, and installing the old package manager and types will likely be "off" by default.

### How could Crossplane help solve your problem?

We should provide the same functionality for the new package manager in the Helm chart by creating a `Provider` for the given provider on install.

negz:
> Currently, Crossplane's Helm chart allows you to create a ClusterPackageInstall for various providers (AWS, Azure, GCP, Rook) when you helm install Crossplane with the --set clusterPackages.gcp.deploy=true --set clusterPackages.gcp.version=master flags (for example). 

Note that I removed this functionality recently, so this issue is now scoped more as "allow installing _any_ providers in the Helm chart".

PR #1749, opened 2020-09-20
Remove superseded features ahead of v0.13 release

### LINKED ISSUE #1702: Stop generating and installing OAM CRDs

We are currently [generating](https://github.com/crossplane/crossplane/blob/365a7c3ad900ff6cf8da646fedb99e894e07c41e/apis/generate.go#L28) OAM CRDs and [installing](https://github.com/crossplane/crossplane/tree/master/cluster/charts/crossplane/templates/crds) them as part of the Crossplane helm chart. This will cause issues when the oam-kubernetes-runtime helm chart installs the CRDs as part of its helm chart.

Remove the OAM CRDs from core Crossplane.

negz:

Fixes #1702
Updates #1670

This PR removes:

* The resource claim type definitions, which were deprecated in v0.12.
* The 'autotarget' controller, which created a `KubernetesTarget` for each `KubernetesCluster` claim.
* The OAM controllers and CRDs, which will now be installed by the [oam-kubernete-runtime](https://github.com/crossplane/oam-kubernetes-runtime/) Helm chart.
* The original package manager, which has been replaced in v0.13.

The removal of the original package manager was not originally planned for v0.13, but keeping it around makes little sense given the removal of resource claims. See the relevant commit message for more context.

I have:

I've built and deployed this branch, along with https://github.com/crossplane/provider-gcp/pull/254 (which is built against this branch and https://github.com/crossplane/crossplane-runtime/pull/201). All seems to be working as expected when I run through the GCP composition examples.


cluster/charts/crossplane/README.md
```diff
@@ -272,30 +112,12 @@ and their default values.
...
-| `clusterPackages.azure.deploy`     | Deploy Azure package                                              | `false`
-| `clusterPackages.azure.version`    | Azure provider version to deploy                                   | `<latest released version>`
-| `clusterPackages.rook.deploy`      | Deploy Rook package                                               | `false`
-| `clusterPackages.rook.version`     | Rook provider version to deploy                                    | `<latest released version>`
-| `personas.deploy`                | Install roles and bindings for Crossplane user personas         | `true`
-| `templateStacks.enabled`         | Enable experimental template stacks support                     | `true`
-| `templateStacks.controllerImage` | Template Stack controller image                                 | `crossplane/templating-controller:v0.2.1`
-| `priorityClassName      `        | Priority class name for crossplane and package manager pods       | `""`
+| `priorityClassName`        | Priority class name for crossplane and package manager pods       | `""`
```
hasheddan, line 115:
```suggestion
| `priorityClassName`        | Priority class name for crossplane pod       | `""`
```

go.mod
```diff
@@ -4,19 +4,15 @@ go 1.13
 
 require (
 	github.com/alecthomas/kong v0.2.11
+	github.com/alecthomas/template v0.0.0-20190718012654-fb15b899a751 // indirect
+	github.com/alecthomas/units v0.0.0-20190924025748-f65c72e2690d // indirect
```
hasheddan, line 8: Do you know what introduced these?

Makefile
```diff
@@ -112,41 +111,6 @@ manifests.prepare:
 	cp $(TYPE_MANIFESTS) $(CROSSPLANE_CHART_DIR)/templates
 	@$(OK) Copied type manifests to Crossplane chart
 
-
-# Add "helm.sh/hook: crd-install" and "helm.sh/hook-delete-policy:
```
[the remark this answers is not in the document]
negz: I'd prefer to approach that in a separate PR - I'd prefer this PR to be strictly removing functionality as it's quite large.

go.mod
negz: iirc they appeared when the dependency on `oam-kubernetes-runtime` was removed.

Makefile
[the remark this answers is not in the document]
hasheddan: SGTM

go.mod
hasheddan: Hmm that is somewhat surprising to me :thinking: I guess if they don't flag `make check-diff` then they are accurate though...

cluster/charts/crossplane-types/templates/clusterrole.yaml
```diff
@@ -66,15 +62,8 @@ rules:
   - list
   - watch
 - apiGroups:
-  - cache.crossplane.io
-  - compute.crossplane.io
-  - database.crossplane.io
   - kubernetes.crossplane.io
```
muvaf, line 65: I remember a PR removing Kubernetes Provider type usage from other repos. Do we need to keep it here?

cluster/charts/crossplane/README.md
```diff
@@ -88,166 +88,6 @@ helm install crossplane --namespace crossplane-system crossplane-master/crosspla
 helm install crossplane --namespace crossplane-system crossplane-master/crossplane --version <version> --devel --disable-openapi-validation
 ```
 
-## Installing Infrastructure Providers
```
muvaf, line 91: Why are these parts removed?

cluster/charts/crossplane/README.md
```diff
@@ -272,30 +112,12 @@ and their default values.
 | `imagePullSecrets`               | Names of image pull secrets to use                              | `dockerhub`                                            |
 | `replicas`                       | The number of replicas to run for the Crossplane operator       | `1`                                                    |
 | `deploymentStrategy`             | The deployment strategy for the Crossplane operator             | `RollingUpdate`                                        |
-| `clusterPackages.aws.deploy`       | Deploy AWS package                                                | `false`
```
muvaf, line 275: Do we remove the feature that allows providers to be installed together with Crossplane?

cluster/charts/crossplane-controllers/values.yaml.tmpl
```diff
@@ -41,14 +37,6 @@ resourcesCrossplane:
...
-  limits:
-    cpu: 100m
-    memory: 512Mi
-  requests:
-    cpu: 100m
-    memory: 256Mi
-
 resourcesRBACManager:
   limits:
     cpu: 100m
```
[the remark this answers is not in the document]
muvaf: And do we capture their functionality with some other mechanism?

examples/README.md
```diff
@@ -1,17 +0,0 @@
-# Crossplane Examples
```
muvaf, line 1: Can we link to `examples` folder of providers somewhere? Either in `README.md` or `examples/README.md`? I believe the shortest path to the actual examples should be as short as possible. An `examples` folder in the root of the repo is pretty good in that sense but if we remove it, at least let's make sure we have references to examples in READMEs and documentation website.

cluster/charts/crossplane/README.md
hasheddan: @muvaf they will be changing with new package manager, I'll circle back to update them in a subsequent PR 👍

cluster/charts/crossplane/README.md
hasheddan: We are replacing it with the new packages, tracked: https://github.com/crossplane/crossplane/issues/1726

examples/README.md
hasheddan: There are also some docs links that I think we may need to clean up as a result of this

cluster/charts/crossplane-types/templates/clusterrole.yaml
negz: I don't think so; I can remove it.

cluster/charts/crossplane-controllers/values.yaml.tmpl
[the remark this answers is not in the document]
negz: I believe they are used by the original package manager, which is removed by this commit: https://github.com/crossplane/crossplane/blob/0382319a83822d2e0b011c62d9adc018f68b9ac7/cluster/charts/crossplane-controllers/templates/package-manager-deployment.yaml

I don't think they're relevant under the new package manager design, which is opinionated about how provider controllers deployments are formed.

examples/README.md
muvaf: What do you guys think about having only `examples/README.md` that points to `examples` folder of the providers one by one?

negz:
I’d like to merge it soon if others feel comfortable. It’s almost entirely a red diff - there are some things (examples, installing providers via the Helm chart) that I’d like us to handle as follow-on work.

examples/README.md
negz: This sounds good to me, but I'd like to keep this diff mostly red and handle that in a follow-on PR.

hasheddan (approved):
Thanks for taking care of this @negz! I feel good about merging then following up 👍

examples/README.md
negz: Just cleaned up any references from the docs to the examples, by moving the referenced examples (scripts, actually) into docs/snippets.

Issue #1750, opened 2020-09-21
Can I install crossplan via a big template or operator


gyliu513:
I can see from the document of crossplan, I have to install crossplan via helm https://crossplane.io/docs/v0.12/getting-started/install-configure.html which request me to download a helm binary, any plan to provide a big yaml template or operator to let me install the crossplan via a kubectl command?

muvaf:
Hi @gyliu513 ! Not checked, but I believe that should be possible wit bare YAMLs. Even if not, we can use kustomize overlays so it'd be `kubectl apply -k`. Would you be interested in contributing those YAMLs? A folder named `cluster/deploy` could be a good home for that.

gyliu513:
sure, let me check.

BTW: I found crossplan already have a operator at https://github.com/operator-framework/community-operators/tree/master/community-operators/crossplane , but I cannot find the source code for this operator, @muvaf do you know where is the source code of this operator?

hasheddan:
@gyliu513 you can find the manifests for install using OLM here: https://github.com/operator-framework/community-operators/blob/34e8415f2a3f0d86e0f5bc56a8dc66969d34d572/community-operators/crossplane/0.11.0/crossplane.v0.11.0.clusterserviceversion.yaml#L300

These are a bit dated, but we plan to update with new releases :)

gyliu513:
Good, then I think we should also maintain the CSV in this repo as well, and sync up to community-operators. Suggest put the CSV files under `cluster/deploy/olm`, comments @hasheddan

hasheddan:
@gyliu513 it certainly sounds reasonable to me that we would maintain the manifest in this repo. I am going to defer to @kasey for specifics as he has worked on the OLM deployment in the past as part of https://github.com/crossplane/crossplane/issues/1553

haoqing0110:
@hasheddan is there a plan to upgrade operator to v0.14.0? since it's the latest release.

negz:
We (the Crossplane community) are not planning on maintaining a Crossplane entry in OperatorHub. However, putting on my @upbound employee hat, we are maintaining an OperatorHub entry for our distribution of Crossplane - see https://operatorhub.io/operator/universal-crossplane.

PR #1753, opened 2020-09-21
Added a bare YAML to deploy crossplane

gyliu513:
Signed-off-by: Guang Ya Liu <gyliu513@gmail.com>

Fixes https://github.com/crossplane/crossplane/issues/1750

I have:

### DIFF
--- a/cluster/deploy/yaml/crossplane.yaml
+++ b/cluster/deploy/yaml/crossplane.yaml
@@ -1,0 +1,10942 @@
+---
+# Source: crossplane/templates/package-manager-serviceaccount.yaml
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: package-manager
+  labels:
+    app: crossplane
+    chart: crossplane-0.12.0
+    release: crossplane
+    heritage: Helm
+---
+# Source: crossplane/templates/serviceaccount.yaml
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: crossplane
+  labels:
+    app: crossplane
+    chart: crossplane-0.12.0
+    release: crossplane
+    heritage: Helm
+imagePullSecrets:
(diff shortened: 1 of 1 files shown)


gyliu513:
This file was generated by following command:

```
helm install crossplane --namespace crossplane-system crossplane-alpha/crossplane --dry-run
```

hasheddan (commented):
Thanks for taking this on @gyliu513! A few thoughts on maintaining a "static" YAML deployment manifest:
- I generally am hesitant to maintain multiple ways to deploy something because it increases the opportunity for them to diverge. However, it doesn't seem entirely unreasonable that someone does not want to use Helm or does not have access to in their operating environment, so I think there is some merit here.
- If we were to maintain this, do you think it should be a `kustomize` chart so that there is still some available configuration? I would potentially buy an argument against it because it could be harder (potentially impossible) to maintain compatible configuration. However, `kustomize` allows users to build overlays on a base, so it seems like we could just provider the base and allow consumers to overlay as they see fit. I believe this is what @muvaf was suggesting on the linked issue.
- Regardless of direction, I think this manifest should definitely be generated automatically by our code generation / build processes. Happy to help out on that if needed :)

muvaf:
> This file was generated by following command
> `helm install crossplane --namespace crossplane-system crossplane-alpha/crossplane --dry-run`

Agree with @hasheddan that it'd be great if we had this command as part of our `make generate` target that will generate those YAMLs. We'd probably need to add version tag to that command so that it generates from the correct chart in `master` and `alpha` channels. Separately, I'd opt for `cluster/deploy` as folder instead of `cluster/deploy/yaml` because it's kind of apparent that they are `yaml` files.

gyliu513:
Thanks, +1 to leverage `make generate` to enable this, let me close my PR.

PR #1758, opened 2020-09-22
Clean up package revision and pass pull secrets and policy

hasheddan:

This PR cleans up a few things:
- Removes `Crossplane` and `Dependencies` from package revision `Status` as they are not yet honored
- Sets security context on packaged provider `Deployment` (same restrictions as install `Pod`)
- Passes `PackagePullSecrets` and `PackagePullPolicy` to package revision so that they can be applied in provider `Deployment`

Read individual commits for more detail.

Passing the secrets and pull policy has specific behavior as implemented here. If you change either field and then roll back to a previous existing inactive revision, the fields will not be applied. Similarly, if you add the fields without change the `package` field of the `Provider` or `Configuration`, the existing revision will not be modified. This could be unexpected, but has the nice quality that rolling back to a previous revision truly means rolling back to how that revision was configured. This scenario can be difficult to explain, so I have provided an example below to illustrate the behavior.

Please see manual testing regarding the behavior of changing `PackagePullSecrets` and `PackagePullPolicy` then rolling back to a previous revision.

I have:

<details>
  <summary>Click to expand!</summary>
  

1. Created a `Provider`

```yaml
kind: Provider
metadata:
  name: provider-aws
spec:
  package: hasheddan/xp3-provider-aws:v0.10.0
```

2. Install `Pod` created with `ImagePullPolicy` `IfNotPresent` because that is the default.

```yaml
spec:
  containers:
  - image: hasheddan/xp3-provider-aws:v0.9.0
    imagePullPolicy: IfNotPresent
    name: unpack
    resources: {}
    securityContext:
      privileged: false
      runAsGroup: 2000
      runAsNonRoot: true
      runAsUser: 2000
```

3. Provider `Deployment` created with `ImagePullPolicy` `IfNotPresent`.

```yaml
containers:
        - image: crossplane/provider-aws:v0.9.0
          imagePullPolicy: IfNotPresent
          name: provider-aws
          resources: {}
          securityContext:
            privileged: false
            runAsGroup: 2000
            runAsNonRoot: true
            runAsUser: 2000
```

4. Update `Provider` to new version and add `PackagePullPolicy` `Always`.

```yaml
kind: Provider
metadata:
  name: provider-aws
spec:
  package: hasheddan/xp3-provider-aws:v0.10.0
  packagePullPolicy: Always
```

5. Install `Pod` created with `ImagePullPolicy` `Always`.

```yaml
spec:
  containers:
  - image: hasheddan/xp3-provider-aws:v0.10.0
    imagePullPolicy: Always
    name: unpack
    resources: {}
    securityContext:
      privileged: false
      runAsGroup: 2000
      runAsNonRoot: true
      runAsUser: 2000
```

6. Provider `Deployment` created with `ImagePullPolicy` `Always`.

```yaml
containers:
        - image: crossplane/provider-aws:v0.10.0
          imagePullPolicy: Always
          name: provider-aws
          resources: {}
          securityContext:
            privileged: false
            runAsGroup: 2000
            runAsNonRoot: true
            runAsUser: 2000
```

7. Roll back to older version of `Provider`.

```yaml
kind: Provider
metadata:
  name: provider-aws
spec:
  package: hasheddan/xp3-provider-aws:v0.9.0
  packagePullPolicy: Always
```

8. Existing install `Pod` is used

```yaml
spec:
  containers:
  - image: hasheddan/xp3-provider-aws:v0.9.0
    imagePullPolicy: IfNotPresent
    name: unpack
    resources: {}
    securityContext:
      privileged: false
      runAsGroup: 2000
      runAsNonRoot: true
      runAsUser: 2000
```

9. Provider `Deployment` created with the existing `ProviderRevision` spec.

```yaml
containers:
        - image: crossplane/provider-aws:v0.9.0
          imagePullPolicy: IfNotPresent
          name: provider-aws
          resources: {}
          securityContext:
            privileged: false
            runAsGroup: 2000
            runAsNonRoot: true
            runAsUser: 2000
```

</details>


pkg/controller/pkg/revision/hook.go
```diff
@@ -121,13 +119,13 @@ func NewConfigurationHooks() *ConfigurationHooks {
 
 // Pre sets status fields based on the configuration package.
 func (h *ConfigurationHooks) Pre(ctx context.Context, pkg runtime.Object, pr v1alpha1.PackageRevision) error {
-	pkgConfig, ok := pkg.(*pkgmeta.Configuration)
+	_, ok := pkg.(*pkgmeta.Configuration)
```
negz, line 122: Should we just make this a complete no-op method for now?

negz (approved):

pkg/controller/pkg/revision/hook.go
hasheddan: My reasoning for keeping the check here is that I think we will need it in the short term future and if we did a no-op here then it would match `NopHooks`, and then it would strictly be duplicate code and in that case it seems like we should just remove `ConfigurationHooks` entirely for now. Since we would be reintroducing this soon, the lowest churn path seemed like to keep it around with minimal functionality, then fill it in with more as necessary.

PR #1761, opened 2020-09-23
Remove crossplane-types and crossplane-controllers Helm charts

negz:

These charts were used to enable the host-aware package manager, which is defunct as of v0.13.

I have:

I've installed Crossplane using the new chart and ensured it works as expected in combination with https://github.com/crossplane/provider-gcp/pull/254.


hasheddan (approved):
LGTM 👍

cluster/charts/crossplane-controllers/values.yaml.tmpl
```diff
@@ -1,54 +0,0 @@
...
-
-rbacManager:
-  deploy: true
-  managementPolicy: All
-
-imagePullSecrets:
-- dockerhub
-
-hostedConfig:
-  enabled: false
```
turkenh, line 22: hostedConfig.enabled needs to be cleaned up from top level values.yaml.tmpl as well: https://github.com/negz/crossplane/blob/hellm/cluster/charts/crossplane/values.yaml.tmpl#L19

turkenh (approved):
May be we should drop helm2 support with this PR as well?

negz:
> May be we should drop helm2 support with this PR as well?

We're definitely planning to drop Helm 2 support with the release of v0.13. I'm not sure what is needed to do that though, so I'd prefer to handle it in a follow on PR. (Unless it's super simple?)

negz:
Just noting that the chart README (and our other, similar install docs) are both fairly stale at this point. I intend to address that in a separate PR, too.

PR #1763, opened 2020-09-23
remove CRDs in templates that are kept for helm2 support

muvaf:
Signed-off-by: Muvaffak Onus <onus.muvaffak@gmail.com>

https://github.com/crossplane/crossplane/pull/1761 removed the install-hook hack that we had for Helm2 support but without that hook stuff, Helm3 won't install the CRDs twice. This PR removes the unnecessary ones.

Fixes #

I have:


hasheddan (commented):
@muvaf I believe this fails because integration tests are still being run with helm2. It looks like we have support in `build` for helm3 (https://github.com/upbound/build/blob/7bad6710447022c7f40c5cff6c230e05af107df4/makelib/k8s_tools.mk#L44) so I think updating https://github.com/crossplane/crossplane/blob/6e6cee50d87f390164d484c4d7ab840338817391/Makefile#L125 to use `$(HELM3)` and then updating the integration testing script appropriately should work :+1:

hasheddan:
@muvaf this was breaking some of the work I was doing in repackaging providers so I went ahead and did full update to helm3 in #1771 and added you as co-author on commit that removes the CRDs. Is that okay with you?

PR #1764, opened 2020-09-23
Remove ignore and ordering constraints on package metadata

negz:

In practice we're using a CLI flag to ignore files when building a package, and the requirement to have the metadata at the beginning of the stream seems to be a complication that we don't really need.

I have:

### DIFF
--- a/design/one-pager-package-format-v2.md
+++ b/design/one-pager-package-format-v2.md
@@ -3,7 +3,14 @@
 
 * Owner: Nic Cope (@negz)
 * Reviewers: Crossplane Maintainers
-* Status: Draft
+* Status: Accepted
+
+## Revisions
+
+* 1.1
+  * Removed the requirement for the `Provider` or `Configuration` metadata to be
+    the first resource in the YAML stream.
+  * Removed the `ignore` metadata field.
 
 ## Background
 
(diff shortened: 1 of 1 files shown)


design/one-pager-package-format-v2.md
```diff
@@ -256,7 +255,7 @@ that will, when invoked:
 
 1. Emit the content of `crossplane.yaml` to stdout.
```
hasheddan, line 256: nitpick: this still somewhat indicates that the `crossplane.yaml` will be emitted first

negz: Switched to an unordered list.

hasheddan (approved):

design/one-pager-package-format-v2.md
```diff
@@ -70,17 +77,16 @@ package consumers.
 All software that consumes a package must do so via its output stream - a YAML
 stream emitted to stdout when the package's OCI image entrypoint is invoked. The
 output stream will consist of a `Provider` or `Configuration` YAML document
-(described below), followed by a series of Crossplane-specific YAML documents,
-which must be either:
+(described below) and a series of Crossplane-specific YAML documents, which must
+be either:
 
 * A `CustomResourceDefinition` that defines a Crossplane managed resource.
```
kasey, line 83: Does restricting packages to these types assume that any additional metadata that may be desired in a package, like metadata for a package registry, will be defined in annotations? We've talked about use case involving an icon or readme, are we ok with those being represented as annotation values?

kasey (approved):
removal of `ignore` and ordering requirement lgtm, the question about additional metadata existing only in the restricted subset of package types isn't a new question so shouldn't gate a merge.

design/one-pager-package-format-v2.md
negz: > Does restricting packages to these types assume that any additional metadata that may be desired in a package, like metadata for a package registry, will be defined in annotations?

Correct - this document proposes all additional metadata should be annotations on the output stream.

> We've talked about use case involving an icon or readme, are we ok with those being represented as annotation values?

Yes, the idea is those would be annotation values. There were some ideas around using the `--context` flag to the unpack binary to control whether such annotations were actually emitted or not; they're mostly only interesting when indexing a package, and they could push our API objects over the 1MB etcd limit.

PR #1766, opened 2020-09-24
Remove cargo-culted reconciler scaffold from RBAC manager

negz:

The RBAC manager controllers never use the mgr, and don't need a special Kubernetes client.

I've just run the tests, and made sure Crossplane builds and starts when deployed via Helm. This change only removes an unused variable and and unused layer of client indirection, so I'm pretty confident it won't break anything.


hasheddan (approved):
LGTM 👍

PR #1771, opened 2020-09-25
Drop support for Helm 2

### LINKED ISSUE #1100: Drop Helm 2 support

Helm 3 GA recently released and has major improvements against Helm 2. However, it has breaking changes and as we try to keep supporting both, we will need to apply not-so-nice [workarounds](https://github.com/crossplaneio/crossplane/pull/1077).

Drop Helm 2 support and set minimum Helm version requirement as 3.0.0.

### LINKED ISSUE #1731: Can't Skip Helm CRDs
### What happened?
I am unable to use the SkipCRDs flag with Helm v3. With Helm v3 you can simply add crds to the crd folder in the chart and then Helm will handle them in a special way and you can skip them too. 

I do agree the templates/crds should still exist for backwards compatibility with Helm v2. However, a lot of other charts will have a value in the `values.yaml` file like below to have helm create the CRDs through the templates like all of the other resources if using Helm v2 or lower.

```yaml
# set to true if using Helm v2 or below
createCRDs: false 
```

Since I am using kustomize, I want to deploy the CRDs by simply using kustomizations like this one ya'll made https://github.com/crossplane/crossplane/blob/master/cluster/charts/kustomization.yaml

Then I can just say `--skip-crds` and manage them on my own. 

### How can we reproduce it?
Run the helm chart with `--skip-crds` and they will still be created.

hasheddan:

Fixes #1731 
Fixes #1100 

Closes #1763 

We had already effectively broken helm 2 support in #1761, but had not removed nested helm charts. This does so and drops all other references to helm 2. It also updates the `build` submodule to pick up helm 3 support and updates integration tests to use helm 3.

I have:

Installing helm chart into cluster, as demonstrated by the updated integration tests.


docs/reference/install.md
```diff
@@ -64,203 +44,20 @@ installed, then you can use the `master` channel.
...
 helm search repo crossplane-master --devel
 
-# Kubernetes 1.15 and newer versions
 helm install crossplane --namespace crossplane-system crossplane-master/crossplane --version <version> --devel
-
-# Kubernetes 1.14 and older versions
-helm install crossplane --namespace crossplane-system crossplane-master/crossplane --version <version> --devel --disable-openapi-validation
-```
-
-## Installing Infrastructure Providers
```
hasheddan, line 98: Went ahead and removed this section and will replace it as part of https://github.com/crossplane/crossplane/issues/1726

Makefile
```diff
@@ -83,7 +83,7 @@ fallthrough: submodules
 manifests:
 	@$(WARN) Deprecated. Please run make generate instead.
 
-generate: $(KUSTOMIZE) go.vendor go.generate gen-kustomize-crds
+generate: $(HELM) $(KUSTOMIZE) go.vendor go.generate gen-kustomize-crds
```
hasheddan, line 86: Note that we still lint / build with helm2. This is due to the structure of the `build` submodule. Making changes there would cause significant churn and is not required to move to only supporting helm3 install.

docs/reference/install.md
```diff
@@ -279,30 +76,12 @@ and their default values.
...
-| `priorityClassName      `        | Priority class name for crossplane and package manager pods       | `""`
+| `priorityClassName`        | Priority class name for crossplane and package manager pods       | `""`
 | `resourcesCrossplane.limits.cpu`        | CPU resource limits for Crossplane                       | `100m`
 | `resourcesCrossplane.limits.memory`     | Memory resource limits for Crossplane                    | `512Mi`
 | `resourcesCrossplane.requests.cpu`      | CPU resource requests for Crossplane                     | `100m`
 | `resourcesCrossplane.requests.memory`   | Memory resource requests for Crossplane                  | `256Mi`
-| `resourcesPackageManager.limits.cpu`      | CPU resource limits for PackageManager                     | `100m`
-| `resourcesPackageManager.limits.memory`   | Memory resource limits for PackageManager                  | `512Mi`
-| `resourcesPackageManager.requests.cpu`    | CPU resource requests for PackageManager                   | `100m`
-| `resourcesPackageManager.requests.memory` | Memory resource requests for PackageManager                | `256Mi`
```
negz, line 301: I missed putting some of the new values in here - we'll need to do that at some point. Can be a follow-on PR.

```yaml
rbacManager:
  deploy: true
  managementPolicy: All

resourcesRBACManager:
  limits:
    cpu: 100m
    memory: 512Mi
  requests:
    cpu: 100m
    memory: 256Mi
```

hasheddan: @negz I can go ahead and add those here :+1:

negz (approved):

hasheddan:
@negz could you circle back and take a look at most recent commit and give a thumbs up if still looks good to you?

negz (approved):
LGTM - thank you!

PR #1773, opened 2020-09-26
Add an explicit 'None' readiness check

negz:

The default behaviour when no readiness checks are supplied is to derive a composed resource's readiness from its 'Ready' status condition. A small subset of composable resources (e.g. ProviderConfigs) do not have status conditions, and do not have _any_ status from which we could reasonably determine their readiness. This commit introduces a new 'None' readiness check that can be used to 'turn off' readiness checking, and thus indicate that a composed resource should always be considered ready.

For example:

```yaml
spec:
  resources:
  - base:
    # ...
  - readinessChecks:
    - type: None
```

It's worth noting that this functionality is possible to emulate with the current implementation by writing something like the below example, but I feel this approach is more explicit and readable.

```yaml
spec:
  resources:
  - base:
    # ...
  - readinessChecks:
    - type: NonEmpty
      fieldPath: metadata.name
```

I have:

### DIFF
--- a/apis/apiextensions/v1alpha1/composition_types.go
+++ b/apis/apiextensions/v1alpha1/composition_types.go
@@ -103,17 +103,19 @@ const (
 	ReadinessCheckNonEmpty     TypeReadinessCheck = "NonEmpty"
 	ReadinessCheckMatchString  TypeReadinessCheck = "MatchString"
 	ReadinessCheckMatchInteger TypeReadinessCheck = "MatchInteger"
+	ReadinessCheckNone         TypeReadinessCheck = "None"
 )
 
 // ReadinessCheck is used to indicate how to tell whether a resource is ready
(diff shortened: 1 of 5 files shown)


hasheddan (approved):
LGTM! I wonder if we ever more explicitly support arbitrary resources in compositions if it would make more sense to have the default be none rather than the `Ready` status condition. However, it would only be for convenience, and maybe it would be even less convenient as it would be more typical for a resource to be a Crossplane-flavor managed resource, in which case the `Ready` condition would be most applicable as a default.

negz:
Thanks for the approval @hasheddan! I'm going to wait to hear from @muvaf since he worked on this functionality originally.

muvaf (approved):
LGTM! That looks simpler to use than `NonEmpty -> metadata.uid`.

PR #1777, opened 2020-09-28
Focus local dev scripts on kind and Helm 3

negz:

This PR updates kind.sh to work with Helm 3 (it was previously failing in master due to HELM_VERSION being unset). It also removes the Minikube and Microk8s variants. I'm hoping this is uncontroverisal, and will allow us to maintain one local dev script going forward rather than three.

I have:


hasheddan (approved):
I think this is a good move, primarily because we install Kind for you, which we do not do for the tools. I also wouldn't be opposed to splitting the Kind setup and Crossplane install into two different scripts that get called sequentially by a make target such that the latter script could be used with any other configured kubeconfig. We can follow up this PR with that implementation though if others are in favor.

cluster/local/kind.sh
```diff
@@ -77,24 +67,29 @@ case "${1:-}" in
     copy_image_to_cluster ${BUILD_IMAGE} "${DOCKER_REGISTRY}/${PROJECT_NAME}:${helm_tag}" "${KIND_NAME}"
 
     [ "$2" ] && ns=$2 || ns="${DEFAULT_NAMESPACE}"
-    echo "installing helm package(s) into \"$ns\" namespace"
-    ${HELM} install --name ${PROJECT_NAME} --namespace ${ns} ${projectdir}/cluster/charts/${PROJECT_NAME} --set image.pullPolicy=Never,imagePullSecrets=''
+    echo "installing helm package into \"$ns\" namespace"
+    ${KUBECTL} get namespace ${ns} 2>/dev/null||${KUBECTL} create namespace ${ns}
```
jbw976, line 71: i had been working on this same issue this weekend :)

another option with helm3 is to use `--create-namespace`, from https://stackoverflow.com/a/60817927/467462

small world though, that stackoverflow post mentions the upstream helm issue on this, opened by none other than illya 🙇‍♂️ 
https://github.com/helm/helm/issues/6794

PR #1778, opened 2020-09-28
Allow providers RBAC access to KubernetesTarget

negz:

This was producing a lot of RBAC access errors in the provider logs. As far as I can tell, it was also preventing the providers from working as expected - I was surprised to find that when installing provider-gcp via the new package manager it wouldn't reconcile anything until I fixed this issue.

I have:

```yaml
kind: Provider
metadata:
  name: provider-gcp-master
spec:
  package: crossplane/provider-gcp:master
```

I've applied the above provider, and confirmed that the GCP example under `docs/snippets/compose` works with this change (but not without it).


hasheddan (approved):
Thanks @negz!

negz:
Thanks for the quick review @hasheddan. There's actually a few more tweaks needed here - I'm going to close this PR and open a new one.

Issue #1782, opened 2020-09-28
Handle dependencies via connection secrets


negz:
### What problem are you facing?

Crossplane creates the resources an XR composes in the order they are specified in their composition. The order in which Crossplane deletes the composed resources is undefined and presumably non-deterministic; it relies on Kubernetes garbage collection to delete resources.

In most cases this is not an issue:

* Cross resource references are not processed when a managed resource is being deleted, so deleting (for example) a VPC network before a subnetwork that references it works just fine. Crossplane waits for confirmation that the external resource has actually been deleted, which typically won't happen until all of its external dependencies have been deleted. i.e. GCP won't let a network be deleted until all the subnetworks in it have been deleted, and Crossplane reflects this.
* A special case exists for the relationship between managed resources and their `ProviderConfig`. Work is ongoing in https://github.com/crossplane/crossplane/issues/1737 to ensure a `ProviderConfig` cannot be deleted while it is in use.

The one case where this is known to be an issue is when a composed resource depends on another composed resource _via its connection secret_. The [Helm provider](https://github.com/crossplane-contrib/provider-helm) commonly faces this issue, because it's typical for a composite resource to:

1. Compose a Kubernetes cluster of some kind (e.g. a `GKECluster`)
1. Compose a Helm `ProviderConfig` that uses said cluster's connection secret as its credentials.
1. Compose a Helm `Release` that uses said `ProviderConfig`.

Once https://github.com/crossplane/crossplane/issues/1737 has been solved the `ProviderConfig` will remain until the `Release` is deleted, but nothing stops the underlying `GKECluster` from being deleted (thus making it impossible to connect to the cluster and delete the `Release`).

### How could Crossplane help solve your problem?

I can think of two ways we could handle this:

1. Something like the approach proposed in https://github.com/crossplane/crossplane-runtime/pull/204. Under this approach connection secret consumers would be expected to add some hint (a finalizer, annotation, etc) to the connection secret(s) they consumed. The managed resource reconciler would be refactored to refuse to delete resources whose connection secrets had such hints.
1. Support ordered, serial deletion of composed resources (or ideally groups of composed resources).

negz:
CC @muvaf, who I believe was investigating option 2 (ordered deletion).

negz:
[...]

Issue #1737, opened 2020-02-12
ProviderConfigs should not be deleted until all resources referring to it are deleted


muvaf:
### What problem are you facing?

When I delete an infra-stack, all CRDs get deletion requests, including `Provider` ones. Since there is no `finalizer` on `Provider` resources, they are deleted immediately. This leads to hanging resources with deletion timestamp because their controllers cannot make deletion calls anymore since they'd have to have `Provider` resource to give them the credentials.

### How could Crossplane help solve your problem?

Crossplane could block the deletion of a `Provider` resources until all resources that used this `Provider` are deleted from the cluster.

One mechanism could be that every resource that is successfully created could add its identification info to `finalizers` array and remove its entry when the resource is deleted.

The other alternative would be to have a `Provider` controller that scans all the resources who refer to this `Provider` instance and block the deletion until it finds none. We could somehow make use of labels for easy listing instead of `providerRef`.

negz:
We've deprecated infra stacks, but we're now allowing ProviderConfigs inside Composite resources so we're running into the same issue.

muvaf:
IMO, we have two options for how we'd like to proceed:
* In composition, specifically leak the resources whose kind is `ProviderConfig`. Fast and dirty hack for 0.13.
* Write controller for `ProviderConfig` kinds that will check whether there is any resource using that `ProviderConfig`.
  * The check logic is pretty generic but fetching the CRs belonging to that provider could be a bit tricky. In case of deletion;
    * Either the controller could query `aws` category (`kubectl get aws`) and all AWS managed resources will be returned.
    * Or we can have a list of types supplied during the initialization of the `ProviderConfig` controller so that it lists all of them. However, this would require each managed resource addition to register to that controller and it'd make _many_ `GET` calls, i.e. one for each type.

I feel like the category querying is simple and we can get it working fast enough compared to the hack. Though depending on category of CRD in actual controller flow feels a bit unsafe. But it's an explicit field on `CustomResourceDefinition` just like kind name so that feeling is probably because we just never used it.

negz:
Another approach could be to have each provider controller add some metadata to the ProviderConfigs they use to authenticate - a finalizer, owner ref, annotation etc. Refuse to let the ProviderConfig be deleted while such annotations still exist. This would probably get pretty messy pretty fast though with many managed resource controllers fighting to update ProviderConfig metadata.

[...]

Issue #1612, opened 2020-06-23
Composite and claim deletion should be blocked while composed resources are being deleted


muvaf:
### What problem are you facing?

When you delete a composite resource, it disappears after making deletion calls to all composed resources. However, there are many cases where actual deletion of the composed resources result in error. To account for that, admins should scan the cluster time to time to see if there are any orphan composed resources.

Also there are systems like stacks that depend on everything being gone so that they can delete the `Provider` object; when composite resource is instantly deleted, `Provider` gets deleted, too, before all composed resources that refer to that `Provider` finished deleting.

### How could Crossplane help solve your problem?

Composite reconciler can confirm that all the resources under `spec.resourceRefs` are gone before removing the finalizer from the composite resource so that it stays in the cluster. Additionally, we can block the deletion of claim as well so that the claim would disappear only if everything related to is gone for sure.

negz:
Could we leave this up to the Kubernetes garbage collector? I believe it we set `blockOwnerDeletion: true` in the controller references we add to all composed resources we'll be able to use regular Kubernetes machinery for this rather than implementing it ourselves.

https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#controlling-how-the-garbage-collector-deletes-dependents

muvaf:
> 	// If true, AND if the owner has the "foregroundDeletion" finalizer, then
	// the owner cannot be deleted from the key-value store until this
	// reference is removed.
	// Defaults to false.

@negz I believe that property applies only if the deletion is of type `foregroundDeletion` which is not the default of `kubectl` or normal standard `Delete` calls we make with controller-runtime. I remember trying to use `blockOwnerDeletion ` for template stacks, too, and didn't decide to go for it for this reason.

samidbb:
Is there any plans to progress on solving this issue? I've found a challenge with this when trying to delete cloudfront distribution resources using claims. The claim and the composite resources will be deleted but the managed resource is still running in k8s with a sync error. The user will think that the resource have been deleted on AWS and everything is fine while in fact there is an error in one of the managed resources..

muvaf:
You can configure the behavior with a flag when you delete the composite resource but I don't think this is the case for claim since we don't have ownership relation there. The controller just makes sure that the composite resource gets the deletion call before removing the finalizer on claim.

[...]

PR #1783, opened 2020-09-28
Custom deletion priority for composed resources

muvaf:

This PR introduces the notion of deletion priority we had in template stacks. Its main use case is the hard dependency of composed resources between each other. For example, we don't want the `EKSCluster` we have in composition to be deleted before `Release` resource of provider-helm because we'd like Helm to delete that release.

Fixes https://github.com/crossplane/crossplane/issues/1737 by allowing user to specify a negative deletion priority for `ProviderConfig` object they have in their `Composition`.
Fixes https://github.com/crossplane/crossplane/issues/1782 by allowing user to assign a deletion priority to Helm release that is higher than the priority of the cluster it points to.
Fixes https://github.com/crossplane/crossplane/issues/1612 by making sure all composed resources are gone before removing the finalizer on composite resource.

@negz has been investigating two mechanisms to fix the first two issues while this feature seems to be solving both. However, this approach applies only to composition scenarios while the secret locking & [`ProviderConfigUsage`](https://github.com/crossplane/crossplane-runtime/pull/206) work on managed resource level. See discussion in https://github.com/crossplane/crossplane/issues/1737#issuecomment-699966637

Fixes #

I have:

Create the following YAML:
```yaml
metadata:
spec:
  claimNames:
    kind: PostgreSQLInstance
    plural: postgresqlinstances
  connectionSecretKeys:
    - username
    - password
    - endpoint
    - port
  crdSpecTemplate:
    group: database.example.org
    version: v1alpha1
    names:
    validation:
      openAPIV3Schema:
        type: object
        properties:
          spec:
            type: object
            properties:
              parameters:
                type: object
                properties:
                  storageGB:
                    type: integer
                required:
                  - storageGB
            required:
              - parameters
---
kind: Composition
metadata:
  labels:
    provider: gcp
    guide: quickstart
spec:
  compositeTypeRef:
  resources:
    - base:
        kind: CloudSQLInstance
        spec:
          forProvider:
            databaseVersion: POSTGRES_9_6
            region: us-central1
            settings:
              tier: db-custom-1-3840
              dataDiskType: PD_SSD
              ipConfiguration:
                ipv4Enabled: true
                authorizedNetworks:
                  - value: "0.0.0.0/0"
          writeConnectionSecretToRef:
            namespace: crossplane-system
          providerConfigRef:
            name: example
          reclaimPolicy: Delete
      patches:
          transforms:
            - type: string
              string:
                fmt: "%s-postgresql"
      connectionDetails:
        - name: port
          value: "5432"
      # The default deletion priority is 0, so CloudSQLInstance will be deleted
      # before Network resource.
      deletionPriority: 1
    - base:
        apiVersion: compute.gcp.crossplane.io/v1beta1
        kind: Network
        spec:
          forProvider:
            autoCreateSubnetworks: false
            routingConfig:
              routingMode: REGIONAL
          reclaimPolicy: Delete
          providerConfigRef:
            name: example
```

Then request a claim via:
```yaml
kind: PostgreSQLInstance
metadata:
  name: my-db
  namespace: crossplane-system
spec:
  parameters:
    storageGB: 20
  compositionSelector:
    matchLabels:
      provider: gcp
  writeConnectionSecretToRef:
    name: db-conn
```

Now delete `my-db` claim and see that `Network` is not deleted until `CloudSQLInstance` is gone.

### DIFF
--- a/pkg/controller/apiextensions/composite/api.go
+++ b/pkg/controller/apiextensions/composite/api.go
@@ -18,11 +18,14 @@ package composite
 
 import (
 	"context"
+	"math"
 	"math/rand"
 	"time"
 
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	runtimev1alpha1 "github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
@@ -30,6 +33,7 @@ import (
 	"github.com/crossplane/crossplane-runtime/pkg/meta"
 	"github.com/crossplane/crossplane-runtime/pkg/reconciler/managed"
 	"github.com/crossplane/crossplane-runtime/pkg/resource"
+	runtimecomposed "github.com/crossplane/crossplane-runtime/pkg/resource/unstructured/composed"
 
 	"github.com/crossplane/crossplane/apis/apiextensions/v1alpha1"
 	"github.com/crossplane/crossplane/pkg/controller/apiextensions/composite/composed"
@@ -300,4 +304,50 @@ func (c *APINamingConfigurator) Configure(ctx context.Context, cp resource.Compo
 	}
 	meta.AddLabels(cp, map[string]string{composed.LabelKeyNamePrefixForComposed: cp.GetName()})
 	return errors.Wrap(c.client.Update(ctx, cp), errUpdateComposite)
+}
+
+// NewAPIPrioritizedDeleter returns a new *APIPrioritizedDeleter.
+func NewAPIPrioritizedDeleter(client client.Client) *APIPrioritizedDeleter {
+	return &APIPrioritizedDeleter{client: client}
+}
+
+// APIPrioritizedDeleter deletes the composed objects in layers that are specified
+// as deletion priority in the *Composition object. Deletion priority of a composed
+// object determines its layer and deletion waits for higher layers to complete
+// before proceeding to the lower one.
+type APIPrioritizedDeleter struct {
+	client client.Client
+}
+
+// Delete makes sure the current highest priority is deleted and the remaining
+// ojbects are returned.
+func (d *APIPrioritizedDeleter) Delete(ctx context.Context, cr resource.Composite, comp *v1alpha1.Composition) ([]resource.Composed, error) {
+	var del []resource.Composed
+	hp := int64(math.MinInt64)
+	for i, ref := range cr.GetResourceReferences() {
+		u := runtimecomposed.New(runtimecomposed.FromReference(ref))
+		nn := types.NamespacedName{Name: ref.Name, Namespace: ref.Namespace}
+		err := d.client.Get(ctx, nn, u)
+		if resource.IgnoreNotFound(err) != nil {
+			return nil, errors.Wrap(err, "cannot get resource")
+		}
+		if kerrors.IsNotFound(err) {
+			continue
+		}
+		p := comp.Spec.Resources[i].DeletionPriority
+		switch {
+		case p > hp:
+			hp = p
+			del = []resource.Composed{u}
+		case p == hp:
+			del = append(del, u)
+		}
+	}
+	for _, u := range del {
+		err := d.client.Delete(ctx, u)
+		if resource.IgnoreNotFound(err) != nil {
+			return nil, errors.Wrap(err, "cannot delete composed resource")
+		}
+	}
+	return del, nil
 }
--- a/pkg/controller/apiextensions/composite/api.go
+++ b/pkg/controller/apiextensions/composite/api.go
@@ -48,6 +48,8 @@ const (
 	errUpdateComposite          = "cannot update composite resource"
 	errCompositionNotCompatible = "referenced composition is not compatible with this composite resource"
 	errGetXRD                   = "cannot get composite resource definition"
+	errGetComposed              = "cannot get composed resource"
+	errDeleteComposed           = "cannot delete composed resource"
 )
 
 // Event reasons.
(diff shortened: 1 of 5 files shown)


pkg/controller/apiextensions/composite/api.go
```diff
@@ -301,3 +307,49 @@ func (c *APINamingConfigurator) Configure(ctx context.Context, cp resource.Compo
...
+	for i, ref := range cr.GetResourceReferences() {
+		u := runtimecomposed.New(runtimecomposed.FromReference(ref))
+		nn := types.NamespacedName{Name: ref.Name, Namespace: ref.Namespace}
+		err := d.client.Get(ctx, nn, u)
+		if resource.IgnoreNotFound(err) != nil {
+			return nil, errors.Wrap(err, errGetComposed)
+		}
+		if kerrors.IsNotFound(err) {
+			continue
+		}
```
hasheddan, line 338: Not sure if worth it, especially initially, but it would be nice to not have to `Get` resources on subsequent reconciles we could know to have already been deleted.

hasheddan: Also, we `Get` every resource here even if the known deletion priority of the resource will not result in it being deleted on this reconcile. We could sort resources beforehand by deletion priority, then iterate through and as soon as we `Get` a resource that exists then we can exit the loop if the next resource's deletion priority is less than that of the current. An even simpler solution to start with would be to move the check below above the `Get` call so that if `hp` is already greater then this resource's `p` (i.e. we have gotten a resource that exists and has a higher deletion priority), then we can skip getting this resource for this reconcile.

muvaf: > but it would be nice to not have to Get resources on subsequent reconciles we could know to have already been deleted.

Between reconciles, we have only the custom resource itself as the state store that we can store at which level we were in the former reconcile, but I'm not sure if adding a field on composite resource is worth it. Keep in mind that most of these `Get` calls are fulfilled by the cache. Also, we can't remove an entry from `spec.resourceRefs` since the order of those resources correspond to the order of the templates in the `Composition` due to the fact that we don't name the templates in `Composition`, so index is used as identifier.

> We could sort resources beforehand by deletion priority, then iterate through and as soon as we Get a resource that exists then we can exit the loop if the next resource's deletion priority is less than that of the current.

Yeah I've tried sorting and it looks like this in practice:
```go
layers := map[int64][]corev1.ObjectReference{}
for i, ref := range cr.GetResourceReferences() {
	layers[comp.Spec.Resources[i].DeletionPriority] = append(layers[comp.Spec.Resources[i].DeletionPriority], ref)
}
type layer struct {
	refs     []corev1.ObjectReference
	priority int64
}
var list []layer
for key, val := range layers {
	list = append(list, layer{refs: val, priority: key})
}
sort.Slice(list, func(i, j int) bool {
	return list[i].priority > list[j].priority
})
cp := int64(math.MaxInt64)
var del []corev1.ObjectReference
for _, layer := range list {
	for _, ref := range layer.refs {
		u := runtimecomposed.New(runtimecomposed.FromReference(ref))
		err := d.client.Delete(ctx, u)
		if resource.IgnoreNotFound(err) != nil {
			return nil, errors.Wrap(err, errDeleteComposed)
		}
		if kerrors.IsNotFound(err) {
			continue
		}
		cp = layer.priority
	}
	if cp == layer.priority {
		del = layer.refs
		break
	}
}
return del, nil

```

So, it's less api-server calls but I think the logic is more complicated with many loops. Also, as soon as you iterate with `Get` calls, I don't think there is a stark difference between going through each object vs only until the current priority level given that the length of resources array is small most of the time.

> An even simpler solution to start with would be to move the check below above the Get call so that if hp is already greater then this resource's p (i.e. we have gotten a resource that exists and has a higher deletion priority), then we can skip getting this resource for this reconcile.

I'm not sure I got that. How are we able to know that resource exists without making either `Get` or `Delete` call? The reason I do a `Get` is that a resource with higher priority could always be chosen even if it's been accounted for and deleted, so, we filter those via the `Get` call.

muvaf: BTW, feel free to post pseudo code if I didn't understand what you mean correctly.

hasheddan: for the simple solution I just mean something like this prior to the `Get` call:
```go
p := comp.Spec.Resources[i].DeletionPriority
if p < hp { // if we have already found a resource that exists and has higher deletion priority we don't need to GET this one
  continue
}
```

pkg/controller/apiextensions/composite/reconciler.go
```diff
@@ -257,6 +272,41 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error)
 		"composition-name", comp.GetName(),
 	)
 
+	if meta.WasDeleted(cr) {
```
negz, line 275: Is there any chance we could detect when everything is at the same deletion priority and simply use Kubernetes garbage collection? I feel like that's the least surprising behaviour for users.

negz:
How would you feel about holding off until after the v0.13 release to assess whether we want to pursue this functionality @muvaf?

My thinking is that:

* I think we should proceed with `ProviderConfigUsage` regardless of whether we add ordered deletion for the reasons outlined in https://github.com/crossplane/crossplane/issues/1737#issuecomment-700199699, and because I feel it provides worthwhile functionality (easily discovering the users of a `ProviderConfig`) even if it was not involved in the deletion process.
* Per https://github.com/crossplane/crossplane/issues/1782#issuecomment-700203185 there's still a case in which a composed managed resource in depends on the secret of a composed resource outside of its composite that this approach does not address.
* Finally, I'd really like to avoid having to override the normal Kubernetes GC behaviour, if possible.

This all isn't to say that we shouldn't pursue the ordered deletion approach - I'd just like a little more time to consider it before we proceed. My testing of https://github.com/crossplane-contrib/provider-helm/pull/31 shows that it addresses the immediate issue we were facing for the v0.13 release - i.e. it's possible to cleanly delete a composite that composes a Kube cluster and a Helm provider config.

negz:
I'm going to close this draft for the time being. We ended up fixing https://github.com/crossplane/crossplane/issues/1737 by introducing `ProviderConfigUsage`. This PR does fix some other issues that we're subject to (e.g. the XR disappearing before its composed resources are gone), but I imagine if it was rebased on master it would have a pretty big merge conflict to resolve and in practice we aren't (AFAIK) seeing too many complaints about these problems at the moment.

PR #1787, opened 2020-09-29
Add oam-kubernetes-runtime as optional crossplane helm chart dependency

hasheddan:

Adds `oam-kubernetes-runtime` as an optional helm chart dependency that can be installed by specifying `oam.enabled=true`. Keeping in draft to ensure other maintainers understand this functionality and any implications (particularly on how we version the dependency) before merge.

I have:

`make build`
`make e2e.run`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,7 @@ gitlab/
 # ignore vscode files (debug config etc...)
 /.vscode
 /.idea
+
+# helm chart dependencies
+**/charts/*.tgz
+**/charts/**/requirements.lock
--- a/cluster/charts/crossplane/requirements.yaml
+++ b/cluster/charts/crossplane/requirements.yaml
@@ -1,0 +1,5 @@
+dependencies:
+- name: oam-kubernetes-runtime
+  version: 0.1.1-1.g8cde76d
+  repository: "https://charts.crossplane.io/master"
+  condition: alpha.oam.enabled
--- a/docs/getting-started/run-applications.md
+++ b/docs/getting-started/run-applications.md
@@ -72,43 +72,6 @@ by this guide:
 kubectl apply -f https://raw.githubusercontent.com/crossplane/crossplane/master/docs/snippets/run/definitions.yaml
 ```
 
-Now that we've defined our workloads and traits, we must install Crossplane's OAM addon. This addon
-packages the controllers that reconcile core OAM workloads and traits.
-
-<ul class="nav nav-tabs">
-<li class="active"><a href="#install-tab-helm3" data-toggle="tab">Helm 3</a></li>
-<li><a href="#install-tab-helm2" data-toggle="tab">Helm 2</a></li>
-</ul>
-<br>
-<div class="tab-content">
-<div class="tab-pane fade in active" id="install-tab-helm3" markdown="1">
-Use Helm 3 to install the latest official `alpha` release of Crossplane OAM
-Addon, suitable for community use and testing:
-
-```console
-kubectl create namespace crossplane-system
-helm repo add crossplane-alpha https://charts.crossplane.io/alpha
-helm install addon-oam-kubernetes-local --namespace crossplane-system crossplane-alpha/oam-core-resources
-```
-
-> Note that the OAM Addon requires at least Kubernetes 1.16.
-
-</div>
-<div class="tab-pane fade" id="install-tab-helm2" markdown="1">
-Use Helm 2 to install the latest official `alpha` release of the Crossplane OAM
-Addon, suitable for community use and testing:
-
-```console
-kubectl create namespace crossplane-system
-helm repo add crossplane-alpha https://charts.crossplane.io/alpha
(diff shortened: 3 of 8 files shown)


.gitignore
```diff
@@ -16,4 +16,8 @@ gitlab/
 
 # ignore vscode files (debug config etc...)
 /.vscode
-/.idea
\ No newline at end of file
+/.idea
+
+# helm chart dependencies
+**/charts/*.tgz
+**/charts/**/requirements.lock
```
hasheddan, line 23: when we build our helm chart it will download the `oam-kubernetes-runtime` chart tarball and create a `requirements.lock` that has its digest. At first I considered committing this as I think we should treat this dependency as we do go modules, etc., but unfortunately the date time in the `requirements.lock` gets updated every time we `help dependency update` as part of the build. Not committing the `requirements.lock` would only be an issue in the case that a PR ran integration tests pinned to a version of `oam-kubernetes-runtime` then after merge the contents of the `oam-kubernetes-runtime` chart at that version changed (this should never happen) and the Crossplane helm chart was built with the new contents. This is an edge case that we should not encounter.

cluster/charts/crossplane/requirements.yaml
```diff
@@ -0,0 +1,5 @@
+dependencies:
+- name: oam-kubernetes-runtime
+  version: 0.1.1-1.g8cde76d
```
hasheddan, line 3: There are issues with how the `oam-kubernetes-runtime` chart is being promoted on official releases (https://github.com/crossplane/oam-kubernetes-runtime/issues/226), so we will start with it just pinned to the latest version.

hasheddan (commented):
@negz @kasey @muvaf this is ready for review :+1:

.gitignore
muvaf, line 23:
```suggestion
cluster/charts/crossplane/charts/
cluster/charts/crossplane/requirements.lock
```

I think that would be easier to understand.

muvaf (approved):
I have a small comment but it looks good to me. Tested manually and it's working as intended.

.gitignore
hasheddan: We would theoretically want to guard against vendoring of a dependency of any chart that existed in this repo though right?

muvaf: > We would theoretically want to guard against vendoring of a dependency of any chart that existed in this repo though right?

Yes, I think so, too. They would all go into `cluster/charts/crossplane`. If we have a completely separate chart, then many changes will be due anyway.

hasheddan: @muvaf oh I was saying using `**/` seemed more future proof :sweat_smile:

muvaf: Yeah, got it 😄  I was just saying that if there will be another directory than `cluster/charts/crossplane`, then many changes will be needed anyway, so it isn't worth the asterisk. But it's a very small detail anyway, you can merge it as is.

muvaf:
I was going through the docs and realized OAM installation should be updated with this PR. Likely, we'll do a `helm upgrade` with `oam.enabled=true` at that step but I'm not fully certain about how that works with chart dependency.

hasheddan:
@muvaf good point :+1: I'll get those updated and check if the `helm update` strategy works

hasheddan:
I updated the installation instructions for `OAM`. I believe some further refinements may be needed for the walk-through, but I will defer those to folks from c/oam-kubernetes-runtime :+1:

cluster/charts/crossplane/requirements.yaml
ryanzhang-oss: just curious, is this `requirements.yaml` identified by the helm automatically as the dependencies or you can put the `dependencies` field anywhere in the directory?

docs/getting-started/run-applications.md
```diff
@@ -35,6 +35,16 @@ on-demand by Crossplane.
 
 ## Infrastructure Operator
 
+### Install OAM Controllers
+
+Use Helm 3 to update our Crossplane installation to install OAM controllers:
+
+```console
+helm upgrade crossplane --namespace crossplane-system crossplane-alpha/crossplane --set oam.enabled=true
```
muvaf, line 43: I think we need a tab for `Helm 3 (master)` since if you installed Crossplane using `master` then the upgrade should be using the same chart version.

hasheddan: Good call :+1:

cluster/charts/crossplane/requirements.yaml
hasheddan: @ryanzhang-oss we are currently building with helm 2, which requires using `requirements.yaml`. Helm 3 allows you to optionally just put them in-line in the `Chart.yaml` :+1:

docs/getting-started/run-applications.md
negz: Should we just remove this section and advise folks to use the `--set oam.enabled=true` in the mainline install docs?

negz (approved):

PR #1790, opened 2020-09-30
Add separate args param for rbacManager

turkenh:

Adds a separate helm value for RBAC Manager deployment args (`rbacManager.args`). Without this fix, I cannot set `--namespace` flag for `Crossplane` as setting `args` value causes the following error in RBAC manager:

```
crossplane: error: unknown long flag '--namespace', try --help
```

I have:

With `helm template` command by using various configuration of related parameters


hasheddan (approved):
Good catch, thanks @turkenh!

PR #1791, opened 2020-09-30
Bump golangci-lint timeout in gh actions

hasheddan:

Bumps the timeout for golangci-lint from the default 1m duration to
match the build submodule lint timeout, which is 10m.

I have noticed this happening across a number of our repos since the GH action was added. Will bump them to match as well :+1: 

I have:

n/a

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,5 +1,5 @@
 run:
-  deadline: 2m
+  timeout: 10m
 
   skip-files:
   - "zz_generated\\..+\\.go$"


negz:
https://github.com/golangci/golangci-lint/blob/master/.golangci.example.yml#L9 it looks like the config file supports this. Should we add it there instead?

hasheddan:
@negz SGTM

hasheddan:
@negz updated :+1: looks like deadline was deprecated for timeout a while back (https://github.com/golangci/golangci-lint/pull/793)

negz (approved):

PR #1792, opened 2020-10-01
Switch to a static package format design

negz:

This PR updates the v2 package metadata design to reflect our decision that packages should be static - i.e. should be files in an OCI image rather than the product of executing an OCI image's entrypoint.

I have:

### DIFF
--- a/design/one-pager-package-format-v2.md
+++ b/design/one-pager-package-format-v2.md
@@ -5,13 +5,6 @@
 * Reviewers: Crossplane Maintainers
 * Status: Accepted
 
-## Revisions
-
-* 1.1
-  * Removed the requirement for the `Provider` or `Configuration` metadata to be
-    the first resource in the YAML stream.
-  * Removed the `ignore` metadata field.
-
 ## Background
 
 Packages extend Crossplane with new functionality. [Historically][packages-v1]
@@ -64,21 +57,19 @@ manager might unpack, parse, and install the aforementioned configuration. This
 document therefore proposes that two key 'views' of package metadata exist:
 
 * The 'on disk format(s)'. The format authored by a platform operator.
-* The 'output stream'. The single supported machine readable format.
+* The 'packaged format'. The single supported machine readable format.
 
-This document proposes that the output stream be a [YAML stream] emitted to
-stdout when the package's OCI image [`entrypoint`][entrypoint] is invoked. The
-'on disk format' could take many forms - for example static files or a Python
-script - as long as that format is rendered as an output stream when invoked by
-package consumers.
+This document proposes that the packaged format be a [YAML stream] stored at
+`/package.yaml` at the root of an OCI image. The 'on disk format' could take
+many forms - for example static files or a Python script - as long as that
+format is compiled to a `/package.yaml` file stored in an OCI image.
 
-### The Output Stream
+### The Packaged Format
 
-All software that consumes a package must do so via its output stream - a YAML
-stream emitted to stdout when the package's OCI image entrypoint is invoked. The
-output stream will consist of a `Provider` or `Configuration` YAML document
-(described below) and a series of Crossplane-specific YAML documents, which must
-be either:
+All software that consumes a package must do so via its packaged format - a YAML
+stream stored as `/package.yaml` in the base layer of an OCI image. The packaged
+format will consist of a `Provider` or `Configuration` YAML document (described
+below) and a series of Crossplane-specific YAML documents, which must be either:
 
 * A `CustomResourceDefinition` that defines a Crossplane managed resource.
 * A `CustomResourceDefinition` that defines a Crossplane provider configuration.
@@ -102,7 +93,7 @@ metadata:
   name: provider-example
   # Optional. Must comply with Kubernetes API conventions. Annotations are
   # opaque to Crossplane, which will replicate them to the annotations of a
-  # PackageRevision when this package is unpacked, but otherwise ignore them.
+  # ProviderRevision when this package is unpacked, but otherwise ignore them.
   # Systems such as package registries may extend this specification to require
   # or encourage specific annotations.
   annotations:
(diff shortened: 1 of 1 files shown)


design/one-pager-package-format-v2.md
```diff
@@ -219,64 +211,89 @@ spec:
...
+
+Metadata that is not consumed by Crossplane and that does not need to be passed
+through to the API server is known as supplementary metadata. This metadata may
+be included in any format its author and consumer(s) see fit, as long as it does
+not interfere with `/package.yaml`.
+
+Note that Crossplane is currently unopinionated about what additional metadata a
+package should contain. In future Crossplane may recommend standard additional
+metadata in order to ensure consumers of such metadata (e.g. package registries
+and user interfaces) use consistent metadata.
```
negz, line 278: @rbwsam @rathpc I feel like we could probably take the existing detail about package metadata from the v1 package manager design and the UI schema design and refactor it into something like `design-doc-recommended-additional-metadata.md`. What do you think?

design/one-pager-package-format-v2.md
```diff
@@ -64,21 +57,19 @@ manager might unpack, parse, and install the aforementioned configuration. This
...
 
-This document proposes that the output stream be a [YAML stream] emitted to
-stdout when the package's OCI image [`entrypoint`][entrypoint] is invoked. The
-'on disk format' could take many forms - for example static files or a Python
-script - as long as that format is rendered as an output stream when invoked by
-package consumers.
+This document proposes that the packaged format be a [YAML stream] stored at
+`/package.yaml` at the root of an OCI image. The 'on disk format' could take
+many forms - for example static files or a Python script - as long as that
+format is compiled to a `/package.yaml` file in an OCI image.
```
hasheddan, line 65: nitpick:
```suggestion
format is compiled to a `/package.yaml` file that is then placed in an OCI image.
```

design/one-pager-package-format-v2.md
```diff
@@ -64,21 +57,19 @@ manager might unpack, parse, and install the aforementioned configuration. This
...
-### The Output Stream
+### The Packaged Format
 
-All software that consumes a package must do so via its output stream - a YAML
-stream emitted to stdout when the package's OCI image entrypoint is invoked. The
-output stream will consist of a `Provider` or `Configuration` YAML document
-(described below) and a series of Crossplane-specific YAML documents, which must
-be either:
+All software that consumes a package must do so via its packaged format - a YAML
+stream stored as `/package.yaml` in a single-layer OCI image. The packaged
```
hasheddan, line 70: Do we have a strict requirement that it be single-layer?

design/one-pager-package-format-v2.md
```diff
@@ -219,64 +211,89 @@ spec:
...
+that resource. Additional metadata must not be attached to `/package.yaml` in
+any other way, including labels or additional resources.
+
+Metadata that is not consumed by Crossplane and that does not need to be passed
+through to the API server is known as supplementary metadata. This metadata may
+be included in any format its author and consumer(s) see fit, as long as it does
+not interfere with `/package.yaml`.
+
+Note that Crossplane is currently unopinionated about what additional metadata a
+package should contain. In future Crossplane may recommend standard additional
```
hasheddan, line 276: nitpick:
```suggestion
package should contain. In the future Crossplane may recommend standard additional
```

design/one-pager-package-format-v2.md
```diff
@@ -219,64 +211,89 @@ spec:
...
+* Discover and append all Crossplane resources contained to `package.yaml`
+* Create a single-layer OCI image containing `package.yaml`.
+
+The aforementioned software may also be responsible for reading additional
+metadata from the package directory and including it in the OCI image.
+
+### Additional Metadata
+
+The Crossplane package format allows for two forms of additional metadata -
+passthrough metadata and supplmentary metadata.
```
rathpc, line 251:
```suggestion
passthrough metadata and supplementary metadata.
```

design/one-pager-package-format-v2.md
rathpc: That makes sense to me. I also think some updates need to be made to https://github.com/crossplane/crossplane/blob/master/design/design-doc-upbound-ui-spec.md but that will be independent of that refactoring

rathpc (changes requested):

design/one-pager-package-format-v2.md
lukeweber: I think package.yaml being the first layer is fine requirement for the future. Images would be well suited maybe in an extra layer, but this is kind of digging into implementation and maybe best to wait until ideas like oras can land with more time in the future to be more opinionated if it makes sense.

design/one-pager-package-format-v2.md
```diff
@@ -184,7 +176,7 @@ metadata:
   name: configuration-example
   # Optional. Must comply with Kubernetes API conventions. Annotations are
   # opaque to Crossplane, which will replicate them to the annotations of a
-  # PackageRevision when this package is unpacked, but otherwise ignore them.
+  # config revision when this package is unpacked, but otherwise ignore them.
```
lukeweber, line 179: `Configuration` instead of `config` just for consistency

design/one-pager-package-format-v2.md
negz: Yeah, I was going more for 'first layer' than 'single layer'. I was mostly intending to communicate 'you don't need to include an entire OS in your package'. Will rephrase.

design/one-pager-package-format-v2.md
negz: This is correct, albeit British-leaning, grammar.

design/one-pager-package-format-v2.md
negz: It is important that our metadata be supple.

design/one-pager-package-format-v2.md
negz: I got lazy because my automatic 80 char line wrapping tool won't wrap comments inside YAML snippets for some reason. Fixed.

negz:
All feedback addressed - PTAL.

hasheddan (approved):

lukeweber (approved):

rbwsam (approved):

Issue #1793, opened 2020-10-01
Add Configuration / Provider create and get commands


hasheddan:
### What problem are you facing?

There is [scaffolding present](https://github.com/crossplane/crossplane/blob/d9497019129dc50a0e4111bd5f25bc32a80f9d83/cmd/crank/configuration/configuration.go#L95) for `Create` and `Get` commands for `Providers` and `Configurations`, but they have not yet been implemented.

### How could Crossplane help solve your problem?

These commands should be implemented. It should be fairly straightforward given that the args and flags are already defined and we are [generating clients](https://github.com/crossplane/crossplane/tree/master/pkg/client/clientset/versioned/typed/pkg/v1alpha1) for these types.

PR #1798, opened 2020-10-02
cli: install command for provider and configuration types

muvaf:

Implements `install` command.

Fixes https://github.com/crossplane/crossplane/issues/1793

Fixes #

I have:

```bash
$ go run cmd/crank/main.go provider install provider-gcp crossplane/provider-gcp:master
provider.pkg.crossplane.io/provider-gcp is created
```

### DIFF
--- a/cmd/crank/configuration/configuration.go
+++ b/cmd/crank/configuration/configuration.go
@@ -34,11 +34,10 @@ import (
 
 // Cmd is the root command for Configurations.
 type Cmd struct {
-	Build  BuildCmd    `cmd:"" help:"Build a Configuration."`
-	Lint   LintCmd     `cmd:"" help:"Lint the contents of a Configuration Package."`
-	Push   pkg.PushCmd `cmd:"" help:"Push a Configuration Package."`
-	Create CreateCmd   `cmd:"" help:"Install a Configuration."`
-	Get    GetCmd      `cmd:"" help:"Get installed Configurations."`
+	Build   BuildCmd    `cmd:"" help:"Build a Configuration."`
+	Lint    LintCmd     `cmd:"" help:"Lint the contents of a Configuration Package."`
+	Push    pkg.PushCmd `cmd:"" help:"Push a Configuration Package."`
+	Install InstallCmd  `cmd:"" help:"Install a Configuration."`
 }
 
 // Run runs the Configuration command.
@@ -98,17 +97,17 @@ func (pc *LintCmd) Run() error {
 	return pc.LintCmd.Lint(ctx, docker, linter)
 }
 
-// CreateCmd creates a Configuration in the cluster.
-type CreateCmd struct {
+// InstallCmd creates a Configuration in the cluster.
+type InstallCmd struct {
 	Name    string `arg:"" name:"name" help:"Name of Configuration."`
 	Package string `arg:"" name:"package" help:"Image containing Configuration package."`
 
-	RevisionHistoryLimit int64 `help:"Revision history limit."`
-	ManualActivation     bool  `short:"a" help:"Enable manual revision activation policy."`
+	RevisionHistoryLimit int64 `short:"rl" help:"Revision history limit."`
+	ManualActivation     bool  `help:"Enable manual revision activation policy."`
 }
 
-// Run the CreateCmd.
-func (p *CreateCmd) Run() error {
+// Run the InstallCmd.
+func (p *InstallCmd) Run() error {
 	ctx := context.TODO()
 	rap := v1alpha1.AutomaticActivation
 	if p.ManualActivation {
@@ -132,17 +131,5 @@ func (p *CreateCmd) Run() error {
 		return errors.Wrap(err, "cannot create configuration")
 	}
 	fmt.Printf("%s/%s is created\n", strings.ToLower(v1alpha1.ConfigurationGroupKind), res.GetName())
-	return nil
-}
-
-// GetCmd gets one or more Configurations in the cluster.
-type GetCmd struct {
-	Name string `arg:"" optional:"" name:"name" help:"Name of Configuration."`
-
-	Revisions bool `short:"r" help:"List revisions for each Configuration."`
-}
-
-// Run the Get command.
-func (b *GetCmd) Run() error {
 	return nil
 }
--- a/cmd/crank/configuration/configuration.go
+++ b/cmd/crank/configuration/configuration.go
@@ -18,10 +18,17 @@ package configuration
 
 import (
 	"context"
+	"fmt"
+	"strings"
 
 	"github.com/docker/docker/client"
+	"github.com/pkg/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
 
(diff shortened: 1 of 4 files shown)


muvaf:
However, while I'm doing this, I realized the style of kubectl is:
```
kubectl get pod
```
while our CLI prefers a different order:
```
kubectl crossplane provider get
```

I'd personally prefer `kubectl` style but deferring to @hasheddan here.

Another thing I realized is the meaning of `get` command. `kubectl get providers` would work and in fact I looked up whether I can plumb in all the arguments we get from user to `kubectl get` to give the same experience but not sure if it's worth it to have `get` at all.

cmd/crank/configuration/configuration.go
```diff
@@ -96,12 +103,35 @@ type CreateCmd struct {
 	Name    string `arg:"" name:"name" help:"Name of Configuration."`
 	Package string `arg:"" name:"package" help:"Image containing Configuration package."`
 
-	History  int  `help:"Revision history limit."`
-	Activate bool `short:"a" help:"Enable automatic revision activation policy."`
+	RevisionHistoryLimit int64 `help:"Revision history limit."`
+	ManualActivation     bool  `short:"a" help:"Enable manual revision activation policy."`
```
muvaf, line 107: I switched this because it seems the default is automatic in the reconciler while if user skips giving `Activate` a value, it'd actually mean `false` hence manual activation will be chosen, which is a different behavior than YAML apply.

cmd/crank/configuration/configuration.go
```diff
@@ -96,12 +103,35 @@ type CreateCmd struct {
...
 }
 
 // Run the CreateCmd.
 func (p *CreateCmd) Run() error {
```
hasheddan, line 111: Should this also be `InstallCmd`?

cmd/crank/configuration/configuration.go
```diff
@@ -96,12 +103,35 @@ type CreateCmd struct {
 	Name    string `arg:"" name:"name" help:"Name of Configuration."`
 	Package string `arg:"" name:"package" help:"Image containing Configuration package."`
 
-	History  int  `help:"Revision history limit."`
-	Activate bool `short:"a" help:"Enable automatic revision activation policy."`
+	RevisionHistoryLimit int64 `help:"Revision history limit."`
```
hasheddan, line 106: Should we maybe add a short flag here given that changing the name from `History` makes the verbose flag much longer?

hasheddan (commented):
@muvaf point taken with the `Get` command. However, I think the purpose there is that it can do things like get a `Provider` / `Configuration` _and_ all of its revisions and their health / activity just by supplying a flag. This would be a fine thing to cut from scope though in first release IMO. As far as matching the `kubectl` commands in order, I really think that CLI's should be <noun> --> <verb>, but `kubectl` is not and since `crank` is going to be a `kubectl` plugin then maybe we should match that. I presume we will want to do that across the board. That will require somewhat significant restructuring, so if you want to go ahead and merge this and let me do the juggling of commands with my PR that would be fine with me.

cmd/crank/configuration/configuration.go
muvaf: How does `rl` sound?

muvaf:
> That will require somewhat significant restructuring, so if you want to go ahead and merge this and let me do the juggling of commands with my PR that would be fine with me.

@hasheddan I believe you said a refactor was already due. So, I want to merge this as is and let you change the order while doing that refactor.

muvaf:
@hasheddan @negz @kasey  Not sure if it's worth it but when I was going through the docs and also testing the install command here, the decision for a name for `kind: Provider` always felt unnecessary. In YAML, we have to do that. But, in CLI, do you think if it's a smoother experience for the user if we made the name an optional flag and if it's not given, assign a name like `provider-gcp-<randstr>`, `provider-gcp` being the string after `/` and before `:` in the package path? Alternatively, we could do what docker/helm does as well, give them a random but human understandable name if a name isn't supplied, like `vetoed-kiwi`.

hasheddan:
@muvaf that sounds pretty good to me. However, just want to make sure that in the commands you mention there it would still be `kubectl crossplane install provider ...` because we need to know whether to create a `Provider`or `Configuration` 👍 I think for the name we could just use the image name directly because you like will not have multiple of a `Provider` or `Configuration` for the same image. I am hesitant to append the random string because it would look similar to our revision naming strategy.

cmd/crank/configuration/configuration.go
hasheddan: SGTM

cmd/crank/provider/provider.go
```diff
@@ -92,28 +98,40 @@ func (pc *LintCmd) Run() error {
...
-type CreateCmd struct {
+// InstallCmd creates a Provider in the cluster.
+type InstallCmd struct {
 	Name    string `arg:"" name:"name" help:"Name of Provider."`
 	Package string `arg:"" name:"package" help:"Image containing Provider package."`
 
-	History  int  `help:"Revision history limit."`
-	Activate bool `short:"a" help:"Enable automatic revision activation policy."`
+	RevisionHistoryLimit int64 `short:"rl" help:"Revision history limit."`
+	ManualActivation     bool  `help:"Enable manual revision activation policy."`
```
hasheddan, line 107: How about `short:"m"` here?

muvaf:
> I think for the name we could just use the image name directly because you like will not have multiple of a Provider or Configuration for the same image.

I am not sure. There could be `personal/config-aws` and `company/config-aws` which are totally different. In fact, we can't guarantee even if we take the full name of `company/config-aws` because different image repositories could have different images with that tag.

> I am hesitant to append the random string because it would look similar to our revision naming strategy.

That's hash based though, right? I think petnames would be different enough. (pushed this update to PR before realizing your comment didn't want random names, both ways are straight-forward though) The experience looks like:
```console
$ kubectl crossplane provider install crossplane/provider-gcp:master
provider.pkg.crossplane.io/touched-walleye is created
```
```console
kubectl crossplane provider install crossplane/provider-aws:master --name my-aws
provider.pkg.crossplane.io/my-aws is created
```

muvaf:
Hmm, the pods look like this after installation:
```
NAME                                            READY   STATUS      RESTARTS   AGE
crossplane-84968cc767-7dzss                     1/1     Running     0          16m
crossplane-provider-aws-master                  0/1     Completed   0          15m
crossplane-provider-gcp-master                  0/1     Completed   0          16m
crossplane-rbac-manager-649c6bfb76-gpcmt        1/1     Running     0          16m
my-aws-f7c1890ac997-77578cbf57-2q69n            1/1     Running     0          15m
touched-walleye-d1b010d96b27-6cbfd676bd-zwtmk   1/1     Running     0          15m
```

I think we do make an assumption about `crossplane/provider-gcp:master` (including tag) being unique already given that unpack pod is named `crossplane-provider-gcp-master` though it will be gone I believe with your open PR. But I realized, I don't really like it much to see `touched-walleye` instead of `provider-gcp`. So, your suggestion, `provider-gcp` as default and returning error in case of conflict sounds good, users can always supply their custom names with `--name`.

muvaf:
@hasheddan ready for another review

PR #1804, opened 2020-10-05
Install providers during Crossplane installation

### LINKED ISSUE #1726: Allow installing Crossplane Providers in Helm chart

muvaf:

You can install any number of arbitrary providers together with Crossplane by supplying their package. Experience looks like the following:
```bash
helm install crossplane --namespace crossplane-system crossplane-alpha/crossplane \
  --set provider.packages="{crossplane/provider-gcp:master,crossplane/provider-aws:master}"
```

Fixes #1726 

Fixes #

I have:

Manually with
```bash
$ cluster/local/kind.sh helm-install # to see if I broke anything
$ helm install crossplane --namespace crossplane-system cluster/charts/crossplane --set image.pullPolicy=Never,imagePullSecrets='',provider.packages="{crossplane/provider-gcp:master,crossplane/provider-aws:master}"
```


cluster/charts/crossplane/values.yaml.tmpl
```diff
@@ -9,6 +9,9 @@ image:
 
 args: {}
 
+provider:
+  packages: []
```
hasheddan, line 13: nitpick: what would you think about switching these (i.e. `packages.provider`)? Then we could also support installing other package types in the future by just adding another field to this stanza

cluster/charts/crossplane/templates/provider.yaml
```diff
@@ -0,0 +1,11 @@
+{{- range $.Values.provider.packages }}
+{{ if ne . "" }}
+apiVersion: pkg.crossplane.io/v1alpha1
+kind: Provider
+metadata:
+  name: {{ . | trim | replace "/" "-" | replace ":" "-" }}
+spec:
+  package: {{ . | trim }}
```
hasheddan, line 8: I like that this is general now as opposed to us manually adding providers to a list that could be installed with Crossplane! 👍

cluster/charts/crossplane/values.yaml.tmpl
muvaf: I went back and forth about this and am still on the fence. At one side, as you point out a `Provider` is a package, so it makes to have `packages.providers` and/or `packages.configurations`. On the other hand, package is what goes to `spec.package` of those resources and users don't really know about the concept that `Provider`s and `Configuration`s are package types, `package` for them means `spec.package` which is actually `image`. So, `provider.packages` makes sense if you treat `package` as the image because the values are in fact image paths and also `provider.packages` read like what we want it to be.

hasheddan: Hmm thinking of the CLI semantics though we say `kubectl crossplane install <package-type> <package-image>`. So that would be (package) --> (package type) --> (package image name), which I feel like matches `packages.providers` and `packages.configurations`.

negz: I slightly lean toward the current `provider.packages` here, for the reasons @muvaf points out. This seems a little more like a UX of "installing a provider, which is in package foo", vs "installing a package, which is provider foo". The former meshes more with our desired UX of the user "installing a provider", not "installing a package".

negz (approved):

cluster/charts/crossplane/values.yaml.tmpl
hasheddan: What would you think of just `providers`? I don't think we will have other fields under `provider` here, which is why is made more sense to me to have `packages.___` as we could have multiple fields under than (i.e. `providers`, `configurations`, etc.). Not a huge deal so fine to defer if y'all like it as is :+1:

negz: I'd be fine with just `providers` too. I'll leave this for @muvaf to decide before he merges.

muvaf: There is also the option of having:
```
providers:
  - package: crossplane/provider-gcp:master
  - package: crossplane/provider-aws:master
```

which paves the way for further customization. But I don't like the way users have to pass in the values:
```
--set providers[0].package='crossplane/provider-gcp:master',providers[1].package='crossplane/provider-aws:master'
```

Tried a few things but it seems that's the support Helm has for passing in an array of objects. I'll merge this PR but let me know if you find a nice UX for that structure, we can revisit.

PR #1807, opened 2020-10-06
Allow for package images to be pre-cached

hasheddan:

Updates package manager to support "pre-cached" packages and mounting cache from PVC. Please see manual testing walk through and individual commit messages for more info.

I have:

1. Created single-node `kind` cluster with host path mount to a local dir

```yaml
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
- role: control-plane
  extraMounts:
  - hostPath: "/home/dan/code/github.com/crossplane/crossplane/cache"
    containerPath: /cache
```

2. Created a PV for that mount on the `kind` "host" (docker container)

```yaml
apiVersion: v1
kind: PersistentVolume
metadata:
  name: package-cache
  labels:
    type: local
spec:
  storageClassName: manual
  capacity:
    storage: 5Mi
  accessModes:
    - ReadWriteOnce
  hostPath:
    path: "/cache"
```

3. Created a PVC for that PV

```yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: package-cache
  namespace: crossplane-system
spec:
  accessModes:
    - ReadWriteOnce
  volumeName: package-cache
  storageClassName: manual
  resources:
    requests:
      storage: 1Mi
```

4. Copied a locally build `provider-aws.xpkg` to `"/home/dan/code/github.com/crossplane/crossplane/cache"`
5. Loaded locally built Crossplane docker image into `kind`
6. Installed locally built Crossplane helm chart with: `helm install crossplane -n crossplane-system _output/charts/crossplane-0.13.0-rc.312.ga64f4778-dirty.tgz --set packageCache.pvc=package-cache,image.pullPolicy=Never`
7. Created the following `Provider`:

```yaml
kind: Provider
metadata:
  name: provider-aws
spec:
  package: provider-aws # remember name of pre-cached package was provider-aws.xpkg
  # when the packagePullPolicy is Never then we assume the package is already cached and you have provided the name
  packagePullPolicy: Never
```
8. Observed successful creation of all CRDs and the `provider-aws` controller
9. Deleted `Provider` and observed that CRDs and controller were removed, but `provider-aws.xpkg` was still present in the cache.

NOTE: behavior when `packagePullPolicy` is not `Never` is unchanged, but I ran through the same tests with pulling a real image from a registry and it worked as expected :+1: 


pkg/controller/pkg/manager/reconciler.go
```diff
@@ -233,7 +233,7 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error)
 
 	p.SetConditions(v1alpha1.Unpacking())
 
-	hash, err := r.digest.Fetch(ctx, p)
+	hash, err := r.digester.Digest(ctx, p)
```
negz, line 236: Nit: `r.package.Digest(ctx, p)` would be more in line with the typical (albeit possibly controversial) prose-like struct field names we use in other reconcilers.

pkg/controller/pkg/revision/imageback.go
```diff
@@ -93,48 +103,13 @@ func (i *ImageBackend) Init(ctx context.Context, bo ...parser.BackendOption) (io
...
-		if !ok {
-			return
-		}
-		i.secrets = v1alpha1.RefNames(s)
-	}
-}
-
-// Cache sets the cache for the ImageBackend.
-func Cache(cache xpkg.Cache) parser.BackendOption {
+// PackageRevision sets the name of the package revision for ImageBackend.
```
negz, line 106: Nit: This comment looks a bit misleading.

negz (approved):

pkg/controller/pkg/manager/reconciler.go
hasheddan: Went with `pkg` because I couldn't use `package` in the struct definition as it collides `package` special word

PR #1808, opened 2020-10-07
Give providers create permissions on their resources

hasheddan:

Expands the provider system role to include create permissions on the
resources it installs. Previously, providers had no need to create
instances of resources they installed, they only reacted to the creation
of instances. However, the introduction of the ProviderConfigUsage makes
it necessary for providers to be able to also create instances of
resources it installs, in this case to track the usage of a
ProviderConfig by a managed resource.

I have:

1. Install Crossplane using latest Helm chart.
2. Install `provider-aws` using `hasheddan/provider-aws:master`.
3. Attempt to create a managed resource. Before this change there is an error, after there is not.

### DIFF
--- a/pkg/controller/rbac/provider/roles/roles.go
+++ b/pkg/controller/rbac/provider/roles/roles.go
@@ -50,7 +50,7 @@ const (
 var (
 	verbsEdit   = []string{rbacv1.VerbAll}
 	verbsView   = []string{"get", "list", "watch"}
-	verbsSystem = []string{"get", "list", "watch", "update", "patch"}
+	verbsSystem = []string{"get", "list", "watch", "update", "patch", "create"}
 )
 
 // Extra rules that are granted to all provider pods.


muvaf:
I wonder if there is a way to limit this to `ProviderConfigUsage` but I'm not very familiar with that part of the codebase. Maybe a filtering based on an annotation on CRD? Not too worried about giving `create` for all resources of providers but conceptually, they shouldn't really need it.

hasheddan:
@muvaf we could certainly have a special case for `ProviderConfigUsage`. It could be as simple as saying anything named `ProviderConfigUsage` gets `create` perms. However, as you mentioned, I'm not sure we need to restrict `create` perms for other resources, and if that is the case then it would likely be more future proof just to give it to everything installed :+1:

negz (approved):
Thanks for fixing this!

PR #1813, opened 2020-10-08
Add CLI install script

hasheddan:

Adds an installation script that can be used with curl to install any
version of crank and save it as kubectl-crossplane, making it able to be
invoked as kubectl crossplane <command>.

I have:

I have not tested this extensively, so keeping it in draft for now.


install.sh
```diff
@@ -0,0 +1,60 @@
...
+        unsupported_arch $OS $ARCH
+        ;;
+    esac
+    ;;
+  *)
+    unsupported_arch $OS $ARCH
+    ;;
+esac
+
+url="https://releases.crossplane.io/${CHANNEL}/${VERSION}/bin/linux_amd64/${BIN}"
```
muvaf, line 55:
```suggestion
url="https://releases.crossplane.io/${CHANNEL}/${VERSION}/bin/${os_arch}/${BIN}"
```

Shouldn't it be like this?

hasheddan: Yep! Thanks for catching :)

muvaf (approved):
Tested it on MacOS with `CHANNEL=master` and `VERSION=v0.13.0-rc.326.g41076fc` and it works as expected. However, I had to look up version from DockerHub. Is there way to make it fetch the latest master? Either a flag or `if channel is master and version is not given, get the latest master`?

muvaf:
@hasheddan In fact, what would you think about making `VERSION` optional for both channels? If no version is given, the latest on the given channel is used. If it's given then that's the version that'll be fetched.

hasheddan:
@muvaf thanks for testing this out! :) Unfortunately we don't publish the binaries in the same way as helm charts (i.e. there isn't an `index.yaml` that specifies `latest`, etc.), so even specifying `alpha` channel will require a version to be set. What we should do it publish a `latest.txt` file on each build for `master` and `alpha` so that this script will check for the latest, then download that version. I am working on incorporating this into our Jenkins pipelines :+1:

muvaf:
> Unfortunately we don't publish the binaries in the same way as helm charts (i.e. there isn't an index.yaml that specifies latest, etc.), so even specifying alpha channel will require a version to be set.

Hmm, is it possible to sort the folders [here](https://releases.crossplane.io/master) in sh script by their version and get the latest? Though probably the `latest.txt` is a safer bet. Hope it's not much of a headache for you to set up.

hasheddan:
I removed the `chmod +x` and download to `/usr/local/bin` so that users don't have have root to install. This makes it such that the install directions will match installing `kubectl`: https://kubernetes.io/docs/tasks/tools/install-kubectl/#install-kubectl-on-linux

1. `curl <script-url>`
2. `chmod +x kubectl-crossplane`
3. `mv kubectl-crossplane /usr/local/bin/`

muvaf (approved):

Issue #1815, opened 2020-10-08
Poorly formed patch transform causes nil ref and crash loop in crossplane


jbw976:
### What happened?

I authored a poorly formed patch transform, which resulted in a nil ref and a crash loop of the crossplane pod. Ideally, invalid compositions should not result in crashes, they should be handled more gracefully with a helpful error message.

```yaml
      patches:
      - fromFieldPath: "metadata.uid"
        toFieldPath: "spec.writeConnectionSecretToRef.name"
        transforms:
          - type: string
            fmt: "%s-postgresql"
```

The transform is invalid because it's specifying the `type: string` as a map when it should be just a field.

This resulted in the following relevant error message and stack trace:

```
E1008 03:32:08.384765       1 runtime.go:78] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
goroutine 659 [running]:
...
	/usr/local/go/src/runtime/panic.go:969 +0x166
github.com/crossplane/crossplane/apis/apiextensions/v1alpha1.(*StringTransform).Resolve(0x0, 0x171b4c0, 0xc0013afeb0, 0xc001497090, 0xc000988ac0, 0x2, 0x2)
	/Users/jared/dev/go/src/github.com/crossplane/crossplane/apis/apiextensions/v1alpha1/composition_types.go:317 +0x42
github.com/crossplane/crossplane/apis/apiextensions/v1alpha1.(*Transform).Transform(0xc001497098, 0x171b4c0, 0xc0013afeb0, 0x0, 0x0, 0x0, 0x0)
	/Users/jared/dev/go/src/github.com/crossplane/crossplane/apis/apiextensions/v1alpha1/composition_types.go:241 +0x15c
github.com/crossplane/crossplane/apis/apiextensions/v1alpha1.(*Patch).Apply(0xc001497188, 0x7f8b625a24c0, 0xc00048a6d0, 0x7f8b62627458, 0xc00009aab8, 0x1, 0x19a2020)
	/Users/jared/dev/go/src/github.com/crossplane/crossplane/apis/apiextensions/v1alpha1/composition_types.go:172 +0x1ea
[...]

PR #1817, opened 2020-10-08
Fix crash caused by invalid transform in composition

muvaf:

We need to check the value of the interface instead of itself directly because interfaces are `nil` only if both their type and value are `nil`. In this case, we assign a `nil` struct but the resulting interface is not `nil` since it carries the type information of that struct even though that struct was `nil`. This PR lets the code check only the value of the interface against `nil`.

Fixes https://github.com/crossplane/crossplane/issues/1815

Fixes #

I have:

Used the example in #1815 . Instead of crash, we get the following error in events:
```
  Warning  ComposeResources   28s (x2 over 58s)    defined/compositeresourcedefinition.apiextensions.crossplane.io  cannot compose resource at index 1: cannot apply overlay: cannot apply the patch at index 0: transform at index 0 returned error: given type string requires configuration
```

### DIFF
--- a/apis/apiextensions/v1alpha1/composition_types.go
+++ b/apis/apiextensions/v1alpha1/composition_types.go
@@ -21,10 +21,9 @@ import (
 	"fmt"
 	"reflect"
 
-	"k8s.io/apimachinery/pkg/runtime"
-
 	"github.com/pkg/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 
 	"github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
 	"github.com/crossplane/crossplane-runtime/pkg/fieldpath"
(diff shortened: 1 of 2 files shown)


apis/apiextensions/v1alpha1/composition_types.go
```diff
@@ -235,7 +234,10 @@ func (t *Transform) Transform(input interface{}) (interface{}, error) {
 	default:
 		return nil, errors.New(errTypeNotSupported(string(t.Type)))
 	}
-	if transformer == nil {
+	// An interface equals nil only if both the type and value are nil. Above,
+	// even if t.<Type> is nil, its type is assigned to "transformer" but we're
+	// interested in whether only the value is nil or not.
+	if reflect.ValueOf(transformer).IsNil() {
```
hasheddan, line 240: This seems reasonable to me 👍 One other option I considered was expanding the `transformer` interface such that it had something like `IsValid()` that could do the underlying `nil` check (you can call functions on a nil pointer in go) and potentially any other validation needed for that specific transform. See an example here: https://play.golang.org/p/3h1vXVL0K4o

muvaf: I think in other possible places that these structs are used, the developers will need to do the `nil` check on the transform type they would like to use anyway since marking it as pointer implies that it can be empty. In this specific case, `nil` check on an interface needs some logic rather than the structs themselves. So, I'd prefer to keep that logic contained in this function where it's needed to keep the structs simple.

hasheddan: @muvaf fair enough, looks like some linter errors to correct here

muvaf: just fixed, thanks!

hasheddan (approved):

jbw976:
Thanks for this quick fix @muvaf! 💪

PR #1823, opened 2020-10-09
Default to current for crank install version

### LINKED ISSUE #1820: Publish index.yaml version file (or similar) in releases bucket

Currently, users are able to get the latest stable Crossplane helm chart when they install from `alpha` channel (https://charts.crossplane.io/alpha/) because we have an `index.yaml`. However, if they want to download the latest version of a binary, such as `crank`, they have to explicitly specify the version because we do not have any indicator in the bucket (https://releases.crossplane.io/alpha/).

As part of the promotion process, we should populate an `index.yaml` (or similar) that can be parsed by something like [install.sh](https://github.com/crossplane/crossplane/blob/master/install.sh) to retrieve the latest binary. It would also be nice to support the same functionality for `master` by updating a file on each pushed build.

hasheddan:

Updates the install.sh script to default to current rather than a pinned
stable version of crossplane. This means that if CHANNEL is master and
no VERSION is supplied then we will get latest master build. If CHANNEL
is alpha and no VERSION is supplied then we will get latest stable.

Fixes #1820 

I have:

Tested running locally.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -3,7 +3,7 @@
 set -eu
 
 CHANNEL=${CHANNEL:-alpha}
-VERSION=${VERSION:-v0.13.0}
+VERSION=${VERSION:-current}
 
 os=$(uname -s)
 arch=$(uname -m)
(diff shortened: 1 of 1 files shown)


negz (approved):

hasheddan:
@prasek ICYMI :)

PR #1836, opened 2020-10-12
Polish docs

muvaf:

Small fixes in docs. Take a look at individual commits for details.

Fixes #

I have:

Manually.


docs/getting-started/install-configure.md
```diff
@@ -250,7 +248,7 @@ spec:
 Then apply it:
 
 ```console
-kubectl apply -f provider.yaml
+kubectl apply -f providerconfig.yaml
```
hasheddan, line 251: Can this one also be a snippet URL?

docs/getting-started/package-infrastructure.md
```diff
@@ -58,197 +58,6 @@ into a package.
 First we'll create a `CompositeResourceDefinition` (XRD) to define the schema of
 our `CompositePostgreSQLInstance` and its `PostgreSQLInstance` resource claim.
 
-<ul class="nav nav-tabs">
```
hasheddan, line 61: Nice catch 👍 This shows off composition better with the tabs removed for the definition!

hasheddan (commented):
One question, otherwise LGTM! Will want to make sure this gets backported to `release-0.13` after merge 🙂

docs/getting-started/install-configure.md
muvaf: GCP `ProviderConfig` object has `spec.projectId` field which user has to edit before applying. It'd have to be something like `curl <snippet> | sed <change things> | kubectl apply -f -` so I just kept it as is because it looked simpler to understand.

hasheddan: I wonder if we could just set the env var then do something like we do here: https://github.com/crossplane/provider-gcp/blob/d97211eb06898c04cdc8b1f8af9be5a9bab0ade5/cluster/local/integration_tests.sh#L273

If not, this seems fine to me :+1:

hasheddan (approved):

docs/getting-started/compose-infrastructure.md
```diff
@@ -267,9 +267,9 @@ Try the following command to watch your provisioned resources become ready:
 kubectl get crossplane -l crossplane.io/claim-name=my-db
 ```
 
-You should also see a `Secret` in the `default` namespace named `db-conn` that
-contains keys that we defined in XRD. If they are filled by the composition, then
-they should appear:
+Once `PostgreSQLInstance` is ready, you should see a `Secret` in the `default`
```
negz, line 270:
```suggestion
Once the `PostgreSQLInstance` is ready, you should see a `Secret` in the `default`
```

Or "once your".

docs/getting-started/compose-infrastructure.md
```diff
@@ -267,9 +267,9 @@ Try the following command to watch your provisioned resources become ready:
...
+namespace named `db-conn` that contains keys that we defined in XRD. If they are
```
negz, line 271:
```suggestion
namespace named `db-conn` that contains keys that we defined in XRD. If they were
```

docs/getting-started/package-infrastructure.md
negz: Duplicating this `definition.yaml` file was actually intentional. It allows each `doc/snippets/package` subdirectory to be a valid package that includes everything it needs (definition, composition, and meta). This makes it quick and easy for us to publish the "getting-started" packages used in this guide.

Perhaps we could make the definition in each package directory a symlink to this single definition, to get the best of both worlds?

hasheddan: Can we just keep all the existing ones plus the new one at `https://raw.githubusercontent.com/crossplane/crossplane/master/docs/snippets/package/definition.yaml`?

negz: I'd be okay with that too. My thinking is that iirc Git supports symlinks, and that linking them all to one "source" file (probably the new one muvaf created) could help avoid them diverging.

muvaf: @negz Symlinking is supported by Git but I was suspicious whether `kubectl crossplane build configuration` supported it and from my experiment, it seems that symlinks are skipped but hardlinks are OK. So, I just went for hardlink.

muvaf: Well, hardlinks aren't supported by Git since it doesn't save node information of files.

hasheddan: @muvaf yeah I feel like just maintaining multiple copies may be best option here...

negz (approved):

PR #1840, opened 2020-10-13
Explicitly avoid leaking file descriptors when caching packages

hasheddan:

File descriptors can be leaked when a file is opened and not subsequently
closed. In the case of the package manager, there would likely need to be a
large number of packages installed before this became an issue as there
should only be one open per revision when it is initially stored. Files
are opened and closed correctly currently on Get calls.

_Note that I don't believe we are actually leaking file descriptors because os.File is implicitly closed when garbage collected. See testing below for more information._

Also adds some unit tests for cache package.

I have:

I ran this and installed 4 packages and compared `lsof` output vs current `master`, and it didn't appear that we were leaking file descriptors. However, it is better practice to close explicitly.

### DIFF
--- a/pkg/xpkg/cache.go
+++ b/pkg/xpkg/cache.go
@@ -81,7 +81,10 @@ func (c *ImageCache) Store(tag, id string, img v1.Image) error {
 	if err != nil {
 		return err
 	}
-	return tarball.Write(ref, img, cf)
+	if err := tarball.Write(ref, img, cf); err != nil {
+		return err
+	}
+	return cf.Close()
 }
 
 // Delete removes an image from the ImageCache.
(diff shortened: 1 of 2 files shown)


negz (approved):
LGTM, though we should probably handle that error.

https://groups.google.com/g/golang-nuts/c/Hj7-HV-W_iU/m/ZqlBiz0REpIJ
https://www.joeshaw.org/dont-defer-close-on-writable-files/

I'm usually of the "fsync and ignore the close error" school of thought when you know you're dealing with an `os.File`, but iirc in this case we're dealing with an `io.ReadCloser` so for all we know the underlying implementation could return legitimate errors on close.

hasheddan:
@negz done :+1: Could you give this a thumbs up if still looks good to you?

negz (approved):

Issue #1435, opened 2020-04-22
Use v1 CustomResourceDefinitions


hasheddan:
`v1beta1` `CustomResourceDefinitions` are [scheduled for removal](https://github.com/kubernetes/apiextensions-apiserver/blob/99685ba23021e88fc9a771f029157b348fd92bab/pkg/apis/apiextensions/v1beta1/types.go#L380) in the `v1.19` cycle. We should start using `v1` prior to the release of `v1.19`, which is [tentatively scheduled](https://github.com/kubernetes/sig-release/pull/1058) for Tuesday, August 25, 2020.

muvaf:
I think we need to decide on a version interval for supported Kubernetes versions. For example, if `v1.CustomResourceDefinition` doesn't work in k8s `1.14`, I'd not use it since 1.14 is still widely used as default by cloud providers. But if the removal will render our CRDs useless in 1.19, we should document that Crossplane is not compatible with 1.19.

negz:
Does upgrading our core providers to use v1 CRDs fall within the scope of this issue? If safe to do so it would be great to enable defaulting as part of those PRs - I believe we could use it instead of an initializer to handle the default `providerConfigRef`, `deletionPolicy`, and perhaps some other things.

hasheddan:
@negz yes I was thinking they would be encompassed by this issue. Agreed on `providerConfigRef` and `deletionPolicy`, I wonder if there will be any for individual managed resources as well. We can always update to v1 then switch to use defaulting to any fields that we currently manually default to later if needed.

negz:
I believe this is now done.

negz:
Scratch that - we also need to use v1 CRDs in our providers.

negz:
Alright, this is actually done now!

PR #1850, opened 2020-10-15
Use v1 CRDs

hasheddan:

Updates controller-tools to generates v1 CRDs by default.

_Note: we cannot update to latest controller-tools version due to https://github.com/kubernetes-sigs/controller-tools/pull/505_

xref https://github.com/crossplane/crossplane/issues/1435 (going to keep that issue open to track updating the CRDs we create in composition as well as updating providers)

I have:

1. Installed `crossplane v0.13.0`
2. Installed `provider-aws`
3. `helm upgrade` locally built chart
4. Verified that crossplane pods were replaced but `provider-aws` kept running without interruption


negz (approved):
I see there's a `check-diff` failure, but otherwise LGTM.

PR #1853, opened 2020-10-16
Aggregate edit and view roles to admin

### LINKED ISSUE #1852: crossplane-view rules are not aggregated up for crossplane-admin
### What happened?
Rules for `rbac.crossplane.io/aggregate-to-view` are not aggregated up for `crossplane-admin`

### How can we reproduce it?
1. Create a cluster role with label, `rbac.crossplane.io/aggregate-to-view: "true"`

Example:
```
kind: ClusterRole
metadata:
  name: another-role
  labels: 
    rbac.crossplane.io/aggregate-to-view: "true"
rules:
- apiGroups:
  - ""
  resources:
  - pods
  - pods/log
  verbs:
  - get
  - list
  - watch
```

2. Check `crossplane-admin` cluster role to see if aggregated

```
kubectl get clusterrole crossplane-admin -o yaml
```

Crossplane version: 0.14.0-rc.18.g99c02be

hasheddan:

This updates the crossplane-admin role to be given any roles that are
given to the edit or view roles. We still explicitly aggregate to the
admin when the RBAC manager creates roles, so these aggregation lables
are convenience for if a user want to manually create a role for edit or
view that should roll up to admin.

Fixes #1852 

Keeping this in draft for discussion. I actually don't love this change as I generally prefer being explicit about the aggregation of a role. If we do go this route, it might be worth examining if we want to clean up some of the logic in the code that explicitly labels edit roles with the aggregate to admin and/or assigning the base role of the admin explicitly then _only_ aggregating edit and view to admin.

I have:

`make e2e.run`

### DIFF
--- a/cluster/charts/crossplane/templates/rbac-manager-managed-clusterroles.yaml
+++ b/cluster/charts/crossplane/templates/rbac-manager-managed-clusterroles.yaml
@@ -26,6 +26,8 @@ aggregationRule:
   clusterRoleSelectors:
   - matchLabels:
       rbac.crossplane.io/aggregate-to-admin: "true"
+      rbac.crossplane.io/aggregate-to-edit: "true"
+      rbac.crossplane.io/aggregate-to-view: "true"
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole


cluster/charts/crossplane/templates/rbac-manager-managed-clusterroles.yaml
```diff
@@ -26,6 +26,8 @@ aggregationRule:
   clusterRoleSelectors:
   - matchLabels:
       rbac.crossplane.io/aggregate-to-admin: "true"
+      rbac.crossplane.io/aggregate-to-edit: "true"
```
turkenh, line 29: Thank a lot for quick response @hasheddan 👍

However, I was more expecting something like https://github.com/crossplane/crossplane/blob/release-0.11/cluster/charts/crossplane/templates/environment-personas-clusterroles.yaml#L119

See, how it bubbles up from view -> edit -> admin

hasheddan: Ah, I see. I think this would have the same effect (minus view roles going to edit), but the referenced model feel more composable :+1: I am going to leave as-is for the moment and wait for @negz to weigh in on the general pattern.

hasheddan:
@turkenh did you end up being able to work around this or are you still looking for this functionality to be included?

turkenh:
> @turkenh did you end up being able to work around this or are you still looking for this functionality to be included?

@hasheddan yes, I've worked around by adding all labels. So, nothing urgent, thanks!

hasheddan:
Closing this out as I don't think this is something we want to do at this point.

PR #1854, opened 2020-10-16
Introduce Crossplane version constraints in packages

### LINKED ISSUE #1843: Support Crossplane Version Constraints in Packages

`Provider` and `Configuration` packages have fields to specify `Crossplane` version constraints (i.e. whether they are compatible with the version of Crossplane that they are being installed into) but the package manager does not currently honor the constraints.

Refuse to install a package that is not compatible with running Crossplane version.

[Placeholder issue, more info to follow]

hasheddan:

Adds Crossplane version constraints to packages with the following behavior:
- Packages that have constraints that do not match the current Crossplane version will not be installed successfully (i.e. no resources will be created)
- Crank will check the `crossplane` spec field during build to ensure the semver constrains provided are valid ranges
- Specifying `ignoreCrossplaneConstraints: true` on `Provider` or `Configuration` spec will cause package revision controllers to skip checking crossplane constrains (useful for dev when using `go run` and version is empty)
- Revision controllers will refuse to install a package with invalid version constraints, even when `ignoreCrossplaneConstraints: true` (note the package would have had to be built with tool other than crank given that crank will validate version constraints)

For documentation on semver constraint format, see https://github.com/Masterminds/semver

Fixes #1843 

I have:

1. `make build`
2. Install locally built Crossplane helm chart
3. Install package with following `crossplane.yaml`

```yaml
apiVersion: meta.pkg.crossplane.io/v1alpha1
kind: Provider
metadata:
  name: test-configuration
spec:
  crossplane: ">=1.0.0"
````

4. Observe failure to install and the following event:

```yaml
  Type     Reason       Age              From                                              Message
  ----     ------       ----             ----                                              -------
  Warning  LintPackage  8s (x3 over 9s)  packages/configurationrevision.pkg.crossplane.io  package is not compatible with Crossplane version (v0.14.0-rc.26.gfa9e89fd)
```

5. Delete package revision and install with `ignoreCrossplaneConstraints: true`

```yaml
kind: Configuration
metadata:
  name: test-configuration
spec:
  package: hasheddan/test-configuration:latest
```

6. Observe successful install

```yaml
NAME                                                 INSTALLED   HEALTHY   PACKAGE                               AGE
configuration.pkg.crossplane.io/test-configuration   True        True      hasheddan/test-configuration:latest   11s
```

7. Also installed some providers, which currently don't have any Crossplane version constraints and they installed successfully.

---

Crank behavior:

1. Build package with `crossplane.yaml` and locally built crank

```yaml
apiVersion: meta.pkg.crossplane.io/v1alpha1
kind: Configuration
metadata:
  name: test-configuration
spec:
  crossplane: ">=a1.0.0"
```

2. Observe following error:

```
kubectl crossplane: error: failed to build package: failed to lint package: package version constraints are poorly formatted: improper constraint: >=a1.0.0
```


apis/pkg/meta/v1alpha1/interfaces.go
```diff
@@ -0,0 +1,49 @@
...
+type Pkg interface {
+	GetCrossplaneVersion() *string
+	GetDependencies() []Dependency
+}
+
+// GetCrossplaneVersion gets the Configuration package's Crossplane version
+// constraints.
+func (c *Configuration) GetCrossplaneVersion() *string {
+	return c.Spec.MetaSpec.Crossplane
+}
```
negz, line 33: It seems like we're referring to this in different places as 'crossplane', 'crossplane version', and 'crossplane constraints'. Is there a way we can be consistent? It seems like maybe `spec.crossplaneVersion` or `spec.crossplaneVersionConstraints` might make the most sense for the meta pkg if we feel okay about renaming that field (which I believe exists, but is non functional in v0.13).

apis/pkg/v1alpha1/interfaces.go
```diff
@@ -140,6 +143,19 @@ func (p *Provider) SetRevisionHistoryLimit(l int64) {
 	p.Spec.RevisionHistoryLimit = &l
 }
 
+// GetIgnoreCrossplaneConstraints of this Provider.
+func (p *Provider) GetIgnoreCrossplaneConstraints() bool {
+	if p.Spec.IgnoreCrossplaneConstraints == nil {
+		return false
+	}
```
negz, line 150: Nit: I'm a little wary of these methods automatically handling the conversion to and from a bool pointer. It seems convenient and I'm sure is pretty safe in this context, but it could be a little surprising to users of the (Go) API given that AFAIK we don't do this in our interface getters and setters anywhere else.

pkg/controller/pkg/revision/reconciler.go
```diff
@@ -345,6 +355,18 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error)
 		return reconcile.Result{RequeueAfter: longWait}, errors.Wrap(r.client.Status().Update(ctx, pr), errUpdateStatus)
 	}
 
+	// Lint Package using generic package linter if version constraints apply.
+	if !pr.GetIgnoreCrossplaneConstraints() {
+		if err := r.generic.Lint(pkg); err != nil {
```
negz, line 360: This seems like a breakdown in the `generic` abstraction; we state that `r.generic` is a "generic" linter (it's not immediately clear to me how that's different from the existing linter", but in practice this logic appears to assume that the generic linter will be exactly `xpkg.PackageCrossplaneCompatible`.

pkg/xpkg/lint.go
```diff
@@ -71,6 +76,42 @@ func IsConfiguration(o runtime.Object) error {
...
 
+// PackageCrossplaneCompatible checks that the current Crossplane version is
+// compatible with the package constraints.
+func PackageCrossplaneCompatible(v version.Operations) parser.ObjectLinterFn {
+	return func(o runtime.Object) error {
+		p, ok := o.(pkgmeta.Pkg)
+		if !ok {
+			return errors.New(errNotMeta)
+		}
+		if p.GetCrossplaneVersion() != nil {
```
negz, line 87: Nit: You could return early here and below `if p.GetCrossplaneVersion() == nil `to avoid indenting the bulk of this function.

pkg/controller/pkg/revision/reconciler.go
hasheddan: The existing linter has checks that are specific to a _type of package_ while the generic linter is for _all package types_. This enables us to gate checks on qualities that are applicable to all package types (in this case `ignoreCrossplaneConstraints`), which can only be determined at reconcile time (i.e. based on the specific PR we are reconciling, not the type of package).

apis/pkg/v1alpha1/interfaces.go
hasheddan: That seems like a reasonable concern. I really was struggling to see why we really even need this field to be a `*bool` instead of a `bool`, but was trying to follow API conventions. Note that we also can do defaulting with v1 CRDs so there may be some areas where we have traditionally used a pointer because we wanted to default to a value that is not the zero value of the type (so we needed to distinguish between zero value and no value) that could be updated.

apis/pkg/meta/v1alpha1/interfaces.go
hasheddan: I feel like `crossplaneVersionConstraints` is most descriptive, but it is also very long :thinking:. I guess we can also have `crossplane` and then a nested `version` field...

negz: Yeah, `spec.crossplaneVersion` seems reasonable to me. `spec.crossplane.version` would be more future proof if we suspect we might introduce more fields under that object in future.

pkg/controller/pkg/revision/reconciler.go
negz: This still seems fishy to me. If I follow correctly what we're doing - running the generic linter - is gated on a more narrowly scoped condition - is ignore crossplane constraints not set. What would happen if someone configured the reconciler to use a different generic linter, that had nothing to do with Crossplane version constraints?

apis/pkg/meta/v1alpha1/interfaces.go
hasheddan: Yeah I like `spec.crossplane.version` :+1:

apis/pkg/v1alpha1/package_types.go
```diff
@@ -44,6 +44,12 @@ type PackageSpec struct {
 	// Default is IfNotPresent.
 	// +optional
 	PackagePullPolicy *corev1.PullPolicy `json:"packagePullPolicy,omitempty"`
+
+	// IgnoreCrossplaneConstraints indicates to the package manager whether to
+	// honor Crossplane version constrains specified by the package.
+	// Default is false.
+	// +optional
+	IgnoreCrossplaneConstraints *bool `json:"ignoreCrossplaneConstraints,omitempty"`
```
muvaf, line 52: @hasheddan Given that it's assumed to be a rare case, would it make sense to make this an annotation instead of a first class spec field?

pkg/controller/pkg/revision/reconciler.go
hasheddan: @negz good point, I think we can just check this directly instead of using a linter :+1:

apis/pkg/v1alpha1/package_types.go
hasheddan: I am pretty indifferent. The fact that it will be present for all package types kind of makes me lean towards making it a field, but either way could be fine

hasheddan: @negz do you have any opinion on this? I am kind of leaning towards it being a dedicated spec field because it is more discoverable, but I don't feel _super_ strongly.

hasheddan:
@negz I believe this is ready for final review. Not sure if we should consider this a breaking API change with the meta crossplane version field change? I tend to think no because the old format was not being honored and it will be stripped out now, so users will see the same behavior when supplying the same package meta file.

hasheddan:
@negz rebased to pick GH actions :+1:

apis/pkg/v1alpha1/package_types.go
negz: Yeah, I'd lean slightly toward the optional field here too.

pkg/controller/pkg/revision/reconciler.go
```diff
@@ -155,18 +156,26 @@ func WithLinter(l parser.Linter) ReconcilerOption {
 	}
 }
 
+// WithVersioner specifies how the Reconciler should versionerally parse a package.
```
negz, line 159: Is "versionerally" a real word? 🤔

negz: Oh - I think this is probably from `s/generic/version` haha.

negz (approved):
Thanks @hasheddan!

pkg/controller/pkg/revision/reconciler.go
hasheddan: I don't believe it is a word... but I like it haha.. cleaning up 👍

Issue #1630, opened 2020-06-30
Crossplane Core should support only Packages and Composition


negz:
### What problem are you facing?

We're focusing Crossplane on defining, composing, and publishing infrastructure APIs. To this end we'd like to remove any functionality that does not aid this from Crossplane core. We see the core of the project as doing a small handful of things:

* Letting you define, compose, and publish infrastructure APIs.
* Letting you package, distribute, and install those APIs.
* Letting you package, distribute, and install the providers that power those APIs.

### How could Crossplane help solve your problem?

Crossplane core could remove or rehome:

* Support for resource claims and classes (https://github.com/crossplane/crossplane/issues/1479)
* Support for workloads (KubernetesApplications) and scheduling (https://github.com/crossplane/crossplane/issues/1595).
* Support for OAM AppliationConfigurations (https://github.com/crossplane/oam-kubernetes-runtime/issues/55).
* Support for (template) StackDefinitions.

Tsumida:
## crossplane+OAM动向
目前通过crossplane使用oam的流程是

 1. 安装crossplane，同时自动创建相关的crd
 2. 为某namespace安装与oam相关的controller，这个可以通过这个[addon-oam-k8s-local](https://github.com/crossplane/addon-oam-kubernetes-local)安装
 3. 在对应的namespace中部署component、AppConf等

奇怪的是，在crossplane下有一个[oam-k8s-runtime](https://github.com/crossplane)仓库，似乎是一个applicationConfiguration controller，这与上面那个仓库有什么差别？结果对比发现:

 1. 两个仓库的目录结构十分相似，对比./charts/Chart.yaml：
![8](https://user-images.githubusercontent.com/29515339/86589968-34951400-bfc1-11ea-93e5-b2fcc124c316.png)
 可见addon-oam-k8s-local这个仓库的controller版本更新。
 2. 根据[slack上的记录](https://app.slack.com/client/TEGUU4LMD/CTHADJCEN/thread/CTHADJCEN-1591859126.351700), maintaier建议不要用oam-k8s-runtime这个库:
 
![9](https://user-images.githubusercontent.com/29515339/86589985-3bbc2200-bfc1-11ea-9e10-10ed5cec560f.png)

[...]

Issue #1755, opened 2020-09-21
Remove KubernetesApplication and KubernetesTarget


negz:
### What problem are you facing?

We deprecated these in https://github.com/crossplane/crossplane/issues/1595. Once we've released v0.13 with the deprecation notice we should move forward with removing them in v0.14 or so.

### How could Crossplane help solve your problem?

Remove the KubernetesApplication, KubernetesApplicationResource, and KubernetesTarget types and their controllers.

prasek:
@negz looks like this is all wrapped up, closing it out 👍

PR #1855, opened 2020-10-17
Remove deprecated workload support

negz:

Fixes https://github.com/crossplane/crossplane/issues/1630
Partially addresses https://github.com/crossplane/crossplane/issues/1755

This removes KubernetesApplication, KubernetesTarget, and the Kubernetes Provider. See the above issues for context.

I have:

It hasn't, apart from verifying that the code compiles.


hasheddan (commented):
@negz I believe we can also remove some of the API groups from crossplane cluster role :) https://github.com/crossplane/crossplane/blob/99c02be53bced648a154b452062ee389f686cf63/cluster/charts/crossplane/templates/clusterrole.yaml#L61

hasheddan (approved):
🎉

PR #1857, opened 2020-10-18
Enable leader election for controller managers

khos2ow:

This PR enables leader election for Crossplane and RBAC controller manager. ~~Note that this has a hard dependency on https://github.com/crossplane/crossplane-runtime/pull/220 and a new release of `crossplane-runtime`~~.

~~But it demonstrates, currently as it stands, the idea of how leader election can be enabled from a central place for any and all controller manager within crossplane ecosystem.~~

Note that this PR still uses controller-runtime v0.6.x which doesn't have [Lease API](https://v1-18.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#lease-v1-coordination-k8s-io). Originally we wanted to enable it right out of the bat, but controller-runtime is still in alpha state.

Partial fix for #5 

I have:

Manually ran the following steps and checked for various points:

- ran `./cluster/local/kind.sh up`
- ran `./cluster/local/kind.sh helm-install` (with additional `--set replicas=2`)
- watched for both pods to come up
- tailed logs of both pods on separate windows, simultaneously
- ran `kubectl crossplane install provider crossplane/provider-aws:v0.13.0`
- looked for the logs of two pods and only **one** of them should pick up this operation

```bash
☸ kind-kind (crossplane-system) ➜ k get pods
NAME                                        READY   STATUS    RESTARTS   AGE
crossplane-8566f7fdcd-ql45n                 1/1     Running   0          5m43s
crossplane-8566f7fdcd-s4c97                 1/1     Running   0          5m43s
crossplane-rbac-manager-84b5dc7cd5-dv2r5    1/1     Running   0          5m43s
provider-aws-2302358baab6-868cb8b97-n6svh   1/1     Running   0          4m32s


☸ kind-kind (crossplane-system) ➜ k logs -f crossplane-8566f7fdcd-ql45n
I1029 17:09:49.160081       1 leaderelection.go:242] attempting to acquire leader lease  crossplane-system/crossplane-leader-election-core...
I1029 17:10:06.566954       1 leaderelection.go:252] successfully acquired lease crossplane-system/crossplane-leader-election-core
E1029 17:10:40.827300       1 aws_credentials.go:77] while getting AWS credentials NoCredentialProviders: no valid providers in chain. Deprecated.
        For verbose messaging see aws.Config.CredentialsChainVerboseErrors


☸ kind-kind (crossplane-system) ➜ k logs -f crossplane-8566f7fdcd-s4c97
I1029 17:09:49.160044       1 leaderelection.go:242] attempting to acquire leader lease  crossplane-system/crossplane-leader-election-core...
E1029 17:10:06.570526       1 leaderelection.go:356] Failed to update lock: Operation cannot be fulfilled on configmaps "crossplane-leader-election-core": the object has been modified; please apply your changes to the latest version and try again

```

### DIFF
--- a/cluster/charts/crossplane/templates/clusterrole.yaml
+++ b/cluster/charts/crossplane/templates/clusterrole.yaml
@@ -77,3 +77,14 @@ rules:
   - patch
   - delete
   - watch
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  verbs:
+  - get
+  - list
+  - create
+  - update
+  - patch
+  - watch
--- a/cluster/charts/crossplane/templates/deployment.yaml
+++ b/cluster/charts/crossplane/templates/deployment.yaml
@@ -47,6 +47,8 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: metadata.namespace
+          - name: LEADER_ELECTION
+            value: "{{ .Values.leaderElection }}"
         volumeMounts:
           - mountPath: /cache
             name: package-cache
--- a/cluster/charts/crossplane/values.yaml.tmpl
+++ b/cluster/charts/crossplane/values.yaml.tmpl
@@ -7,6 +7,7 @@ image:
   tag: %%VERSION%%
   pullPolicy: Always
 
+leaderElection: true
 args: {}
 
 provider:
@@ -17,7 +18,9 @@ imagePullSecrets:
 
 rbacManager:
   deploy: true
+  replicas: 1
   managementPolicy: All
+  leaderElection: true
   args: {}
 
 priorityClassName: ""
--- a/cmd/crossplane/core/core.go
+++ b/cmd/crossplane/core/core.go
@@ -17,6 +17,7 @@ limitations under the License.
 package core
 
 import (
+	"fmt"
 	"time"
 
 	"github.com/pkg/errors"
@@ -35,10 +36,11 @@ import (
 
 // Command configuration for the core Crossplane controllers.
 type Command struct {
-	Name      string
-	Namespace string
-	CacheDir  string
-	Sync      time.Duration
+	Name           string
+	Namespace      string
+	CacheDir       string
+	LeaderElection bool
+	Sync           time.Duration
 }
 
 // FromKingpin produces the core Crossplane command from a Kingpin command.
@@ -47,6 +49,7 @@ func FromKingpin(cmd *kingpin.CmdClause) *Command {
 	cmd.Flag("namespace", "Namespace used to unpack and run packages.").Short('n').Default("crossplane-system").OverrideDefaultFromEnvar("POD_NAMESPACE").StringVar(&c.Namespace)
 	cmd.Flag("cache-dir", "Directory used for caching package images.").Short('c').Default("/cache").OverrideDefaultFromEnvar("CACHE_DIR").ExistingDirVar(&c.CacheDir)
 	cmd.Flag("sync", "Controller manager sync period duration such as 300ms, 1.5h or 2h45m").Short('s').Default("1h").DurationVar(&c.Sync)
+	cmd.Flag("leader-election", "Use leader election for the conroller manager.").Short('l').Default("false").OverrideDefaultFromEnvar("LEADER_ELECTION").BoolVar(&c.LeaderElection)
 	return c
 }
 
@@ -59,7 +62,11 @@ func (c *Command) Run(log logging.Logger) error {
 		return errors.Wrap(err, "Cannot get config")
 	}
 
-	mgr, err := ctrl.NewManager(cfg, ctrl.Options{SyncPeriod: &c.Sync})
+	mgr, err := ctrl.NewManager(cfg, ctrl.Options{
+		LeaderElection:   c.LeaderElection,
+		LeaderElectionID: fmt.Sprintf("crossplane-leader-election-%s", c.Name),
+		SyncPeriod:       &c.Sync,
+	})
 	if err != nil {
 		return errors.Wrap(err, "Cannot create manager")
 	}
--- a/cluster/charts/crossplane/README.md
+++ b/cluster/charts/crossplane/README.md
@@ -67,8 +67,9 @@ and their default values.
 | `image.tag` | Image tag | `master` |
 | `image.pullPolicy` | Image pull policy | `Always` |
 | `imagePullSecrets` | Names of image pull secrets to use | `dockerhub` |
-| `replicas` | The number of replicas to run for the Crossplane and RBAC Manager (if enabled) pods | `1` |
+| `replicas` | The number of replicas to run for the Crossplane pods | `1` |
 | `deploymentStrategy` | The deployment strategy for the Crossplane and RBAC Manager (if enabled) pods | `RollingUpdate` |
+| `leaderElection` | Enable leader election for Crossplane Managers pod | `true` |
 | `priorityClassName` | Priority class name for Crossplane and RBAC Manager (if enabled) pods | `""` |
 | `resourcesCrossplane.limits.cpu` | CPU resource limits for Crossplane | `100m` |
 | `resourcesCrossplane.limits.memory` | Memory resource limits for Crossplane | `512Mi` |
@@ -82,6 +83,8 @@ and their default values.
 | `resourcesRBACManager.requests.cpu` | CPU resource requests for RBAC Manager | `100m` |
 | `resourcesRBACManager.requests.memory` | Memory resource requests for RBAC Manager | `256Mi` |
 | `rbacManager.deploy` | Deploy RBAC Manager and its required roles | `true` |
+| `rbacManager.replicas` | The number of replicas to run for the RBAC Manager pods | `1` |
+| `rbacManager.leaderElection` | Enable leader election for RBAC Managers pod | `true` |
 | `rbacManager.managementPolicy`| The extent to which the RBAC manager will manage permissions. `All` indicates to manage all Crossplane controller and user roles. `Basic` indicates to only manage Crossplane controller roles and the `crossplane-admin`, `crossplane-edit`, and `crossplane-view` user roles. | `All` |
 | `alpha.oam.enabled` | Deploy the `crossplane/oam-kubernetes-runtime` Helm chart | `false` |
 
--- a/cluster/charts/crossplane/templates/rbac-manager-clusterrole.yaml
+++ b/cluster/charts/crossplane/templates/rbac-manager-clusterrole.yaml
@@ -77,4 +77,15 @@ rules:
   - clusterrolebindings
   verbs:
   - "*"
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  verbs:
+  - get
+  - list
+  - create
+  - update
+  - patch
+  - watch
 {{- end}}
--- a/cluster/charts/crossplane/templates/rbac-manager-deployment.yaml
+++ b/cluster/charts/crossplane/templates/rbac-manager-deployment.yaml
@@ -9,7 +9,7 @@ metadata:
     release: {{ .Release.Name }}
     heritage: {{ .Release.Service }}
 spec:
-  replicas: {{ .Values.replicas }}
+  replicas: {{ .Values.rbacManager.replicas }}
   selector:
     matchLabels:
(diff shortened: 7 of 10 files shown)


cmd/crossplane/core/core.go
```diff
@@ -59,7 +62,15 @@ func (c *Command) Run(log logging.Logger) error {
 		return errors.Wrap(err, "Cannot get config")
 	}
 
-	mgr, err := ctrl.NewManager(cfg, ctrl.Options{SyncPeriod: &c.Sync})
+	mgr, err := controller.NewManager(cfg, controller.Options{
+		Options: &ctrl.Options{
```
khos2ow, line 66: @negz I've already updated this here based on the corresponding `crossplane-runtime` PR. Although at this current state of the PR nothing will even get compiled, but this was for the sake of demonstration of the idea.

cmd/crossplane/core/core.go
```diff
@@ -59,7 +69,35 @@ func (c *Command) Run(log logging.Logger) error {
...
+		return errors.New(errManagerName)
+	}
+	if c.Namespace == "" {
+		return errors.New(errManagerNamespace)
+	}
+	mgr, err := ctrl.NewManager(cfg, ctrl.Options{
+		LeaderElection:             c.LeaderElection,
+		LeaderElectionID:           fmt.Sprintf("crossplane-leader-election-%s", c.Name),
+		LeaderElectionNamespace:    c.Namespace,
+		LeaderElectionResourceLock: resourcelock.LeasesResourceLock,
```
negz, line 98: I recall that we discussed wanting to use lease based locks so that folks, but I didn't realise until now that they're only available in controller-runtime 0.7.0, which is still in alpha. Given that we don't want to upgrade to 0.7.0 until it's out of alpha, I think we should just go with ConfigMap based locks for now.

cmd/crossplane/core/core.go
```diff
@@ -59,7 +69,35 @@ func (c *Command) Run(log logging.Logger) error {
...
+	if c.Name == "" {
+		return errors.New(errManagerName)
+	}
+	if c.Namespace == "" {
+		return errors.New(errManagerNamespace)
+	}
+	mgr, err := ctrl.NewManager(cfg, ctrl.Options{
+		LeaderElection:             c.LeaderElection,
+		LeaderElectionID:           fmt.Sprintf("crossplane-leader-election-%s", c.Name),
+		LeaderElectionNamespace:    c.Namespace,
```
negz, line 97: https://github.com/kubernetes-sigs/controller-runtime/blob/v0.6.3/pkg/leaderelection/leader_election.go#L61

It's not documented very well, but it seems like if we omit this we'll use the namespace where the pod is running by default. I think it might be simpler overall to just do that, so I'd prefer to not set `LeaderElectionNamespace` unless we get requests for it to be configurable.

cmd/crossplane/core/core.go
```diff
@@ -59,7 +69,35 @@ func (c *Command) Run(log logging.Logger) error {
...
+	// 		LeaderElectionNamespace: c.Namespace,
+	// 		SyncPeriod:              &c.Sync,
+	// 	},
+	// 	ManagerName:      c.Name,
+	// 	ManagerNamespace: c.Namespace,
+	// })
+
+	// approach 2:
+	// keep calling `ctrl.NewManager` with additional config here
+	// ----------------------------------------------------------
```
negz, line 87: This is my preferred approach; I feel a little bit of repetition is simpler overall than attempting to provide a helper library for this.

cmd/crossplane/core/core.go
```diff
@@ -59,7 +69,35 @@ func (c *Command) Run(log logging.Logger) error {
...
+	if c.Name == "" {
+		return errors.New(errManagerName)
+	}
+	if c.Namespace == "" {
+		return errors.New(errManagerNamespace)
+	}
```
negz, line 93: I believe if we omit the `LeaderElectionNamespace` per my comment below we'll be able to avoid these empty string checks; I'm pretty sure `Name` will never be empty unless there's a programming error.

negz (commented):
Thanks @khos2ow! I I like approach 2 here. Could you please close https://github.com/crossplane/crossplane-runtime/pull/220 and update this PR to address the comments I've left.

cmd/crossplane/core/core.go
khos2ow: Yes, I agree. What I can suggest is either of the following options:

1) keep this PR as is until controller-runtime graduates out of alpha and then update dependencies and merge this
2) revert back to v0.6.0 and merge with this PR and in another one, prior to v0.14.0 release update to v0.7.0

I prefer the latter depending on the release time of controller-runtime, but generally prefer to directly go to Lease API in Crossplane v0.14.0, if possible.

cmd/crossplane/core/core.go
khos2ow: That's true, it might have been slight over-engineering on my side to expose the ability to provide any arbitrary namespace for leader election resource, which in reality there might never a case for this separation. And if so it can easily be added back again.

cmd/crossplane/core/core.go
khos2ow: In the meantime, I can open another PR (similar to https://github.com/crossplane/crossplane-runtime/pull/210) to update `controller-runtime` to latest v0.7.0-alpha and fix potential issues, compilation errors, and such there so when the controller-runtime was released we can bring it over here as soon as possible (maybe even just in time for v0.14.0)

cluster/charts/crossplane/templates/deployment.yaml
```diff
@@ -27,8 +27,9 @@ spec:
       serviceAccountName: {{ template "name" . }}
       containers:
       - image: {{ .Values.image.repository }}:{{ .Values.image.tag }}
-        {{- if .Values.args }}
         args:
+        - --leader-election={{ .Values.leaderElection }}
+        {{- if .Values.args }}
         {{- range $arg := .Values.args }}
         - {{ $arg }}
         {{- end }}
```
hasheddan, line 35: Any idea what happens if someone manually provided `--leader-election` as an arg as well?

cmd/crossplane/core/core.go
khos2ow: Actually, I think we should revert this back, without `LeaderElectionNamespace: c.Namespace` there won't be any way to manually run this out of a cluster (i.e. `go run cmd/crossplane/main.go core ...` won't work) with following error:

```
main: error: cannot run crossplane: Cannot create manager: unable to find leader election namespace: not running in-cluster, please specify LeaderElectionNamespace
```

cluster/charts/crossplane/templates/deployment.yaml
khos2ow: That I believe will be handled by `kingpin`, not entirely sure yet. I mean errors out or being able to handle duplicate occurrence of a flag. Trying to replicate now.

khos2ow: So it did error out. `main: error: flag 'leader-election' cannot be repeated, try --help`

cmd/crossplane/core/core.go
negz: I wonder how likely it is that folks will want to run Crossplane out of cluster _and_ also need leader election? For the `go run` use case it seems like we'd be developing Crossplane, and thus probably would not need leader election enabled.

cluster/charts/crossplane/values.yaml.tmpl
```diff
@@ -1,4 +1,4 @@
-replicas: 1
+replicas: 2
```
negz, line 1: I feel like we can probably leave this at 1 replica by default - but maybe make this configurable? The nature of Crossplane is such that we don't really need to fail over instantly to a second replica; we can usually tolerate being offline for a few seconds while a replacement spins up.

khos2ow: I had a similar idea and actually wanted to add a new variable too `rbacManager.replicas` to effectively distinguish the two.

cmd/crossplane/core/core.go
khos2ow: It is most probably very less likely, and for the latter you are correct, us running `go run` we won't be needing leader election but without that extra attribute for `ctrl.Options` it will never work.

cluster/charts/crossplane/templates/deployment.yaml
khos2ow: Moving this into deployment's `env` could solve but, but I've noticed another thing. It appears kingpin is not able to process negative boolean flags, i.e. `--leader-election=false` doesn't work, but `export LEADER_ELECTION=false` does.

khos2ow: Moved this to environment variable of deployment which prevents collision if `--leader-election` was explicitly provided in `args`. Note that explicit arg will override the environment variable, so the intention of the user will be preserved.

cmd/crossplane/core/core.go
khos2ow: In 38d7b36 I effectively reversed the behavior, it means leader election is disabled by default (`go run` continues to work as before and still can be enabled with `--leader-election` flag or `LEADER_ELECTION=true` environment variable), and in the helm chart it gets enabled by default upon deployment.

negz:
Thanks @khos2ow! This pattern is looking good to me, though it seems like the e2e tests are failing for reasons that aren't immediately obvious to me.

cluster/charts/crossplane/templates/deployment.yaml
negz: > It appears kingpin is not able to process negative boolean flags, i.e. --leader-election=false doesn't work, but export LEADER_ELECTION=false does.

I vaguely recall that `kingpin` handles this by using a `--no` prefix, i.e. `--no-leader-election`.

khos2ow:
Yeah, e2e tests were failing because I had forgotten to escape boolean values I put in deployment's env variables. Should be good now.

cluster/charts/crossplane/templates/deployment.yaml
khos2ow: Ah interesting, I wasn't aware of `--no` convention. But anyways I reversed the logic for `--leader-election` to be `false` by default and gets enabled by choice.

cluster/charts/crossplane/templates/clusterrole.yaml
```diff
@@ -77,3 +77,12 @@ rules:
...
   - watch
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  verbs:
+  - get
+  - create
+  - update
+  - watch
```
khos2ow, line 88: @negz These verbs are needed for leader lease resource, technically we can add the rest of the verbs too, but I think omitting those should be fine.

negz: Sounds good! Will we need to add these to the RBAC manager's cluster role template too?

https://github.com/crossplane/crossplane/blob/master/pkg/controller/rbac/provider/roles/roles.go#L54

We'll probably need to add ConfigMaps to the above extra rules that the RBAC manager creates for all provider deployments, too.

negz (commented):
This is looking good to me @khos2ow! One more comment to address, then please squash your commits and rebase on master and I think we'll be good to merge. Once this is merged will you be able to replicate this pattern to our various providers?

cluster/charts/crossplane/templates/clusterrole.yaml
khos2ow: Yes that's a good catch, but it's kind of important when provider is also enabling leader election process. To make it simple for now I think it's fine to just enable it by default irrelevant of what provider wants, but maybe it will be good idea to extract this from provier itself (maybe new field `LeaderElection bool` in `PackageSpec`?)

Also to be in parity I'm gonna update `clusterrole.yaml` to include all the verbs of `verbsSystem` (i.e. `"get", "list", "watch", "update", "patch", "create"`).

khos2ow:
This should be good now. Yes definitely after this gets merged I'll start with one of the providers to get more familiar and after the first one is done the rest of them should be even faster.

negz (approved):
Awesome, thanks for working on this @khos2ow! It's great to see one of our oldest issues being addressed at last.

negz:
Looks like this needs another rebase on master; I'll merge once that's ready to go.

PR #1858, opened 2020-10-19
PreserveUnknownFields on RawExtension fields

hasheddan:

Adds CRD generation directive to specify preserve unknown fields in
RawExtension types so that information provided is not stripped during
pruning, which is default for v1 CRDs.

Follow-up to #1850 
xref #1435 

See https://github.com/kubernetes/kubernetes/issues/88252 for more info about why this change is needed.

I have:

Applied an XRD and Composition, then created an instance of the XR. Before the change of the XRD preserve unknown fields, validation failed because the `openAPIV3Schema` was being stripped of the XRD. Before the change of the Composition preserve unknown fields, rendering resources failed because `base` was getting stripped off resource templates. After both changes, full composition workflow worked as expected.

### DIFF
--- a/apis/apiextensions/v1alpha1/composition_types.go
+++ b/apis/apiextensions/v1alpha1/composition_types.go
@@ -76,6 +76,7 @@ type TypeReference struct {
 // should be processed.
 type ComposedTemplate struct {
 	// Base is the target resource that the patches will be applied on.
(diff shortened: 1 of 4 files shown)


muvaf (approved):
LGTM! Thanks for digging in and finding the problem!

PR #1861, opened 2020-10-20
Have controllers deal in v1 CRDs

negz:

Partial fix for https://github.com/crossplane/crossplane/issues/1435

This PR follows on from https://github.com/crossplane/crossplane/pull/1850 by updating the various core Crossplane controllers (API extensions, packages, and RBAC) to deal in v1 CRDs. It also contains a small commit that requires composition bases to be valid Kubernetes objects.

I have:

I've tested it by following https://crossplane.io/docs/v0.13/getting-started/compose-infrastructure.html using the master version of the Azure provider, and the master (i.e. v0.13 compatible) version of the Azure getting started configuration. Everything appears to work as expected.

I also validated that the following `Composition` is rejected:

```yaml
kind: Composition
metadata:
  labels:
    guide: quickstart
    provider: azure
  name: test
spec:
  compositeTypeRef:
  resources:
  - base:
      apiVersion: database.azure.crossplane.io/v1alpha3
      # kind: PostgreSQLServerFirewallRule
      spec:
        cool: 900
```

```console
$ kubectl apply -f cmp.yaml 
The Composition "test" is invalid: spec.resources[0].base.kind: Required value: must not be empty
```


negz:
@hasheddan Sorry if this duplicates work you had already started on. I originally intended only to make a few tweaks to our `Composition` validation, but then ended down a massive rabbit hole that took me most of the way to this PR. It didn't occur to me until I was almost done that I had not declared my intent to work on this. 😬

cluster/charts/crossplane/crds/apiextensions.crossplane.io_compositeresourcedefinitions.yaml
```diff
@@ -108,7 +108,7 @@ spec:
                           description: type is an OpenAPI type definition for this column. See https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#data-types for details.
                           type: string
                       required:
-                      - JSONPath
+                      - jsonPath
```
negz, line 111: Note that this is the only reason this PR is labelled as a breaking change - `additionalPrinterColumns.JSONPath` changed to `additionalPrinterColumns.jsonPath` with v1 CRDs. If we want to avoid the breaking change we could define our own `AdditionalPrinterColumns` struct with the old struct tag that doesn't comply with API conventions.

I'm inclined to make this break because I'd like to make sure we comply with API conventions when we hit v1beta1, and because I suspect relatively few folks are using this functionality. Presumably there's an upgrade path for those who are, i.e.:

* Remove printer columns from your XRDs
* Upgrade
* Re-add printer columns to your XRDs

apis/apiextensions/v1alpha1/xrd_types.go
```diff
@@ -89,35 +89,7 @@ type CRDSpecTemplate struct {
 	// Top-level and per-version columns are mutually exclusive.
 	// If no top-level or per-version columns are specified, a single column displaying the age of the custom resource is used.
 	// +optional
-	AdditionalPrinterColumns []v1beta1.CustomResourceColumnDefinition `json:"additionalPrinterColumns,omitempty"`
-}
-
-// CustomResourceDefinitionVersion describes a version for CRD.
-type CustomResourceDefinitionVersion struct {
```
negz, line 96: Note that this type wasn't used anywhere.

apis/apiextensions/v1alpha1/ccrd/crd.go
```diff
@@ -42,130 +41,127 @@ const (
...
-				Status: &v1beta1.CustomResourceSubresourceStatus{},
-			},
-			Validation: &v1beta1.CustomResourceValidation{
-				OpenAPIV3Schema: &v1beta1.JSONSchemaProps{
-					Type:       "object",
-					Properties: BaseProps(),
-				},
+// ForCompositeResource derives the CustomResourceDefinition for a composite
+// resource from the supplied CompositeResourceDefinition.
+func ForCompositeResource(xrd *v1alpha1.CompositeResourceDefinition) (*extv1.CustomResourceDefinition, error) {
```
negz, line 53: There's some arguably drive-by refactoring in this file. I was attempting to update the existing implementations to output v1 CRDs and found that I was struggling to understand what they were doing. I'm hoping that the updated implementation makes a reasonable tradeoff of being a little less DRY in exchange for a little more readability.

apis/apiextensions/v1alpha1/ccrd/crd.go
```diff
@@ -16,7 +16,7 @@ limitations under the License.
 
 // Package ccrd generates CustomResourceDefinitions from Crossplane definitions.
 //
-// v1beta1.JSONSchemaProps is incompatible with controller-tools (as of 0.2.4)
+// v1.JSONSchemaProps is incompatible with controller-tools (as of 0.2.4)
 // because it is missing JSON tags and uses float64, which is a disallowed type.
 // We thus copy the entire struct as CRDSpecTemplate. See the below issue:
 // https://github.com/kubernetes-sigs/controller-tools/issues/291
```
muvaf, line 22: It seems like this issue is closed with https://github.com/kubernetes-sigs/controller-tools/pull/427 . Would you mind testing if the new controller-tools can work for us here?

apis/apiextensions/v1alpha1/ccrd/crd.go
```diff
@@ -42,130 +41,127 @@ const (
...
+func ForCompositeResourceClaim(xrd *v1alpha1.CompositeResourceDefinition) (*extv1.CustomResourceDefinition, error) {
+	if err := validateClaimNames(xrd); err != nil {
+		return nil, errors.Wrap(err, errInvalidClaimNames)
+	}
 
-		if err := validateClaimNames(d); err != nil {
-			return errors.Wrap(err, errInvalidClaimNames)
-		}
+	vr := extv1.CustomResourceDefinitionVersion{
+		Name:                     xrd.Spec.CRDSpecTemplate.Version,
```
muvaf, line 114: Are we keeping this? The `v1.CRD`s have only `versions` array instead of `version` as opposed to our `CompositeResourceDefinition`. I think now could be a good time for a breaking change here and adopt the current v1 CRD schema since we plan on making XRD `v1beta1`.

apis/apiextensions/v1alpha1/ccrd/crd.go
muvaf: The new structure looks much simpler to understand, thanks!

apis/apiextensions/v1alpha1/ccrd/schemas.go
```diff
@@ -154,8 +154,8 @@ func CompositeResourceClaimSpecProps() map[string]v1beta1.JSONSchemaProps {
 // CompositeResourceStatusProps is a partial OpenAPIV3Schema for the status
 // fields that Crossplane expects to be present for all defined or published
 // infrastructure resources.
-func CompositeResourceStatusProps() map[string]v1beta1.JSONSchemaProps {
-	return map[string]v1beta1.JSONSchemaProps{
+func CompositeResourceStatusProps() map[string]v1.JSONSchemaProps {
+	return map[string]v1.JSONSchemaProps{
 		// TODO(negz): Remove composedResources, readyResources, and
```
muvaf, line 159: Should we remove these fields now?

apis/apiextensions/v1alpha1/composition_types.go
```diff
@@ -77,6 +77,7 @@ type TypeReference struct {
 type ComposedTemplate struct {
 	// Base is the target resource that the patches will be applied on.
 	// +kubebuilder:pruning:PreserveUnknownFields
+	// +kubebuilder:validation:EmbeddedResource
```
muvaf, line 80: > An embedded resource is a value that has apiVersion, kind and metadata fields. They are validated implicitly according to the semantics of the currently running apiserver. It is not necessary to add any additional schema for these field, yet it is possible.

Will this allow `base`s without `metadata` field?

pkg/xpkg/lint.go
```diff
@@ -73,10 +74,14 @@ func IsConfiguration(o runtime.Object) error {
 
 // IsCRD checks that an object is a CustomResourceDefinition.
 func IsCRD(o runtime.Object) error {
-	if _, ok := o.(*v1beta1.CustomResourceDefinition); !ok {
+	switch o.(type) {
+	case *extv1beta1.CustomResourceDefinition:
+		return nil
+	case *extv1.CustomResourceDefinition:
+		return nil
```
muvaf, line 81:
```suggestion
	case *extv1beta1.CustomResourceDefinition, *extv1.CustomResourceDefinition:
		return nil
```

Would this work?

apis/apiextensions/v1alpha1/ccrd/crd.go
hasheddan: @muvaf unfortunately we can't update to `v0.4.0`, which is the version that the referenced issue is fixed in because of the bug described in https://github.com/crossplane/crossplane/pull/1850 😞

apis/apiextensions/v1alpha1/ccrd/crd.go
hasheddan: Agreed, although I believe that is tracked in a separate issue so maybe a follow-up to this PR?

apis/apiextensions/v1alpha1/ccrd/crd.go
negz: Yeah, this was the rabbit hole I went down. I was hoping we might get some kind of validation of this validation schema, but had no luck.

apis/apiextensions/v1alpha1/composition_types.go
negz: It does. I think I left a comment about this in the commit message because I also read that it would require metadata.

apis/apiextensions/v1alpha1/ccrd/crd.go
negz: I don't think we currently have an issue tracking this for XRs (just for Crossplane CRs in general). I'm actually not quite convinced that we should (or perhaps rather can cleanly) support serving multiple versions of an XR, so I'd like to raise a distinct issue to discuss the topic and get some community feedback.

pkg/xpkg/lint.go
negz: I believe so. Will update.

apis/apiextensions/v1alpha1/ccrd/crd.go
negz: Issue raised: https://github.com/crossplane/crossplane/issues/1871

negz:
I believe I've addressed all feedback - PTAL.

muvaf (approved):
Thanks!

hasheddan (approved):

Issue #1583, opened 2020-06-02
Fork Upbound build submodule


negz:
### What problem are you facing?

Currently Crossplane relies on a CI/CD system that could be opaque to community members who do not work for Upbound.

* The majority of repos under the crossplane org use [a git submodule of Make libraries](https://github.com/upbound/build/) that lives under the upbound org. This submodule is also use by several private repositories inside Upbound. This makes it difficult for community members to update, because their updates must account for potential impact on projects they could not possibly be aware of.
* The build submodule is invoked by a Jenkins instance hosted at https://jenkinsci.upbound.io/. While this instance has some level of GitHub integration (i.e. GitHub org members may login and run and restart builds) it is hosted and administered by Upbound employees.

### How could Crossplane help solve your problem?

At a minimum we should ensure Crossplane's CI/CD system is associated with "Crossplane" rather than "Upbound". This would involve forking the build submodule to github.com/crossplane/build, and hosting Jenkins at https://jenkins.crossplane.io.

We may consider taking this a step further and replacing Jenkins with [GitHub Actions](https://github.com/features/actions). This "GitHub native" CI/CD system may be less opaque to community members, and integrate more tightly with GitHub's native permissions model.

negz:
https://crossplane.github.io/docs/master/release/release-process.html

Note that Jenkins (somewhat) automates the above release process for Crossplane, and any proposal to change to a different CI/CD system will need to account for automating Crossplane releases (i.e. tagging, Cross compiling, and publishing artifacts).

negz:
Not strictly related to CI/CD openness, but if we do move to Github actions it would be good to replace our previous configuration that added comments to PRs for linter violations per https://github.com/crossplane/crossplane/issues/1411.

negz:
We're now using GitHub Actions which has made a big difference, but the Upbound build submodule being used internally still means it's hard for folks who don't work for Upbound to understand the blast radius of their changes to the build submodule.

jeanduplessis:
I'm re-opening this issue for further consideration

negz:
/fresh

I'm pretty sure _Upbound_ decided to fork the Upbound build submodule. So internal Upbound projects are moving away from it. I still think it would be worth forking it under the Crossplane org to clarify that it's a "Crossplane thing", that we could optimize for use only by Crossplane projects.

PR #1865, opened 2020-10-21
Migrate core CI pipeline to GitHub Actions

negz:

Fixes https://github.com/crossplane/crossplane/issues/1583

This PR migrates Crossplane's CI/CD system from Jenkins to GitHub Actions. Its logic is largely implemented by the https://github.com/upbound/build/ `make` library, and thus unchanged from the Jenkins pipeline. The big differences in this implementation are:

* We use the [Docker Buildx](https://github.com/docker/setup-buildx-action) and [Setup Go](https://github.com/actions/setup-go) actions - not the Upbound cross container - to ensure repeatable builds. This allows us to avoid the ~5 minute cross container build time penalty.
* We run the key parts of the build (lint, unit test, integration test, build artifacts) in parallel on distinct workers.

The net result of these changes is that we can [build and verify a pull request in ~6.5 minutes](https://github.com/negz/crossplane/actions/runs/319197909), where Jenkins has been [averaging ~20 minutes](https://jenkinsci.upbound.io/job/crossplane/job/crossplane/job/build/job/master/) lately.

In addition to being faster, migrating to GitHub Actions means:

* No particular entity (i.e. Upbound) needs to own and operate the CI/CD infrastructure.
* We don't need to pay for Jenkins and its runners anymore. Actions is free for OSS.
* Setting up CI for new repos is easier. Any repo that includes `.github/workflows/ci.yml` will have CI/CD enabled.
* GitHub Actions is arguably friendlier, more idiomatic, and more tightly integrated for GitHub users than Jenkins.
* We always use a cached Go binary, and often used cached Go dependencies. Failures downloading these has been a common cause of failed builds on Jenkins.

Take a look at https://github.com/negz/crossplane/actions/runs/319197909 to see this pipeline in action.

I have:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,198 @@
...
+          GIT_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      
+      - name: Promote Artifacts
+        if: github.ref == 'refs/heads/master' && env.AWS_ACCESS_KEY_ID != '' && env.AWS_SECRET_ACCESS_KEY != ''
+        run: make -j2 promote
+        env:
+          BRANCH_NAME: master
+          CHANNEL: master
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
```
negz, line 197: Note that I haven't tested the publish and promote steps, except to verify that they don't run on my fork. The idea here is that:

* Publish will run on any build that has valid AWS secrets set.
* Promote will run only on the master branch as long as valid AWS secrets are set.

Note that unlike Jenkins, "any build" here can include forks. In practice forks should run this workflow, but skip the publish and promote stages because they don't have the AWS credential secrets set. If a fork did somehow acquire and set these credentials (in its GitHub Settings) it would publish and promote artifacts, but this seems unlikely.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,198 @@
...
+          # builds by default. Specifying --load does so.
+          BUILD_ARGS: "--load"
+      
+      - name: Publish Artifacts
+        run: make -j2 publish BRANCH_NAME=${GITHUB_REF##*/}
+        if: env.AWS_ACCESS_KEY_ID != '' && env.AWS_SECRET_ACCESS_KEY != ''
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          GIT_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
negz, line 188: This secret is provided automatically - https://docs.github.com/en/free-pro-team@latest/actions/reference/authentication-in-a-workflow

.github/workflows/ci.yml
```diff
@@ -0,0 +1,198 @@
...
+          # We're using docker buildx, which doesn't actually load the images it
+          # builds by default. Specifying --load does so.
+          BUILD_ARGS: "--load"
+      
+      - name: Publish Artifacts
+        run: make -j2 publish BRANCH_NAME=${GITHUB_REF##*/}
+        if: env.AWS_ACCESS_KEY_ID != '' && env.AWS_SECRET_ACCESS_KEY != ''
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
```
negz, line 187: I suggest we set these two secrets at the org level, to make setting up new repos easy - https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets#creating-encrypted-secrets-for-an-organization

.github/workflows/ci.yml
```diff
@@ -0,0 +1,235 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - master
```
negz, line 6: Should we include the release branches here @hasheddan and @jbw976?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,249 @@
...
+        with:
+          name: output
+          path: _output/**
+      
+      - name: Login to Docker
+        uses: docker/login-action@v1
+        if: env.DOCKER_USR != ''
+        with:
+          username: ${{ secrets.DOCKER_USR }}
+          password: ${{ secrets.DOCKER_PSW }}
```
negz, line 231: I suggest we set these and the AWS secrets at the org level, to make setting up new repos easy - https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets#creating-encrypted-secrets-for-an-organization

.github/workflows/ci.yml
```diff
@@ -0,0 +1,245 @@
...
+          key: ${{ runner.os }}-pkg-${{ hashFiles('**/go.sum') }}
+          restore-keys: ${{ runner.os }}-pkg-
+
+      - name: Vendor Dependencies
+        run: make vendor vendor.check
+
+      # This action uses its own setup-go, which always seems to use the latest
+      # stable version of Go. We could run 'make lint' to ensure our desired Go
+      # version, but we prefer this action because it leaves 'annotations' (i.e.
+      # it comments on PRs to point out linter violations).
```
hasheddan, line 53: This seems fine to me, we would be more concerned about build errors due to go version but I think linting is fine with whatever is being used here. Kind of surprised it is not exposed as configurable though

.github/workflows/ci.yml
hasheddan: I believe so 👍

.github/workflows/ci.yml
```diff
@@ -0,0 +1,245 @@
...
+        run: make vendor vendor.check
+
+      - name: Run Unit Tests
+        run: make -j2 test
+
+      - name: Publish Unit Test Coverage
+        uses: codecov/codecov-action@v1
+        with:
+          flags: unittests
+          file: _output/tests/linux_amd64/coverage.txt
```
hasheddan, line 120: nitpick: can this be made general like we do in the current Jenkinsfile? `_output/tests/**/coverage.txt`

hasheddan (commented):
This is awesome to see @negz! Thanks so much for investing time in this :)

It looks like we are building multiple times here in order to be able to run our different steps concurrently. Had you considered doing a serial build, then splitting out to concurrent running of unit tests / e2e tests / etc.? Is that even possible with GH actions?

.github/workflows/ci.yml
hasheddan: @negz I believe that we currently do publish artifacts for to https://releases.crossplane.io/build/ for any successful build on a PR. I don't think we necessarily have to do that though, especially since it looks like we now upload to GH directly (which is super cool!)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,245 @@
...
+
+      - name: Vendor Dependencies
+        run: make vendor vendor.check
+
+      - name: Run Unit Tests
+        run: make -j2 test
+
+      - name: Publish Unit Test Coverage
+        uses: codecov/codecov-action@v1
+        with:
```
khos2ow, line 118: I think it's better to add the following line here as well. To continue publishing the code coverage even if some test fails.

```diff
      - name: Publish Unit Test Coverage
        uses: codecov/codecov-action@v1
+       if: always()
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,245 @@
...
+  # a step 'if env.AWS_USR' != ""', so we copy these to succinctly test whether
+  # credentials have been provided before trying to run steps that need them.
+  DOCKER_USR: ${{ secrets.DOCKER_USR }}
+  AWS_USR: ${{ secrets.AWS_USR }}
+
+jobs:
+
+  lint:
+    runs-on: ubuntu-latest
+
```
khos2ow, line 33: I usually add this line right after `runs-on` in all the jobs too, that gives us a native way (without code change) to skip ci, for whatever reason.

```yaml
if: "!contains(github.event.head_commit.message, '[ci skip]')"
```

Note that `[ci skip]` is by convention and can be changed to any arbitrary text and can be put anywhere in the commit message (title or body).

.github/workflows/ci.yml
```diff
@@ -0,0 +1,245 @@
...
+      
+      - name: Upload Artifacts (to GitHub)
+        uses: actions/upload-artifact@v2
+        with:
+          name: output
+          path: _output/**
+      
+      - name: Login to Docker
+        uses: docker/login-action@v1
+        if: env.DOCKER_USR != ''
```
khos2ow, line 224: I think this additionally can get enforced even further (and to prevent running on forks) by modifying `if` to the following:

```yaml
if: github.repository == ‘crossplane/crossplane’ && env.DOCKER_USR != ''
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,245 @@
...
+        with:
+          username: ${{ secrets.DOCKER_USR }}
+          password: ${{ secrets.DOCKER_PSW }}
+      
+      - name: Publish Artifacts (to S3 and Docker Hub)
+        run: make -j2 publish BRANCH_NAME=${GITHUB_REF##*/}
+        if: env.AWS_USR != '' && env.DOCKER_USR != ''
```
khos2ow, line 231: The same `if` condition above can be applied here as well.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,245 @@
+name: CI
```
khos2ow, line 1: I think it might worth considering to split this file into smaller pieces two, that can help to run them in parallel even more and maybe put more _ref_ related restriction on them. May be something like this:

```
$ tree .github/workflows/
.github/workflows/
├── prerelease.yaml   (if needed)
├── publish.yaml      (for building, publishing, promoting)
├── release.yaml      (if needed)
└── validate.yaml     (for lint, vendor check, tests)
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,245 @@
...
+      - 'docs/**'
+      - 'design/**'
+
+env:
+  # Common versions
+  GO_VERSION: '1.14'
+  GOLANGCI_VERSION: 'v1.31'
+  DOCKER_BUILDX_VERSION: 'v0.4.2'
+
+  # Common users. We can't run a step 'if secrets.AWS_USR != ""' but we can run
```
khos2ow, line 23: Check my comment below on modifying `if` condition to prevent the run on forks.

khos2ow (commented):
This looks pretty amazing, if I may suggest some further changes.

.github/workflows/ci.yml
negz: Right - this implementation will also publish (but not promote) artifacts to Docker Hub and S3 on every build, as long as the appropriate secrets are set to do so.

.github/workflows/ci.yml
negz: Our current Jenkins pipeline uses a similar convention, but in practice we mostly only use it to skip CI when we know we're making changes that don't need it to run (e.g. editing markdown). This pipeline excludes some files (markdown, images) via `paths-ignore` so I'm hoping we can get away without needing to explicitly skip CI now.

.github/workflows/ci.yml
negz: I might try `github. repository_owner == 'crossplane'` here so that we can replicate the same `ci.yml` to our provider repos without needing to make changes.

.github/workflows/ci.yml
negz: It's currently challenging to break some of these up into distinct workflows (or even jobs) because we're using the Docker daemon to pass state between steps; i.e. one step builds a Docker image, then the next step publishes it, etc. This could be an area for improvement in future though.

.github/workflows/ci.yml
negz: Our code coverage publishing is quite slow (~2.5 mins), so I personally lean toward failing fast and not spending the time publishing coverage if the tests don't pass.

.github/workflows/ci.yml
negz: Doing so actually slows down publishing coverage; I believe some detection of files is skipped when we give an explicit file path. I'd prefer to keep this fixed unless we're concerned about `coverage.txt` moving.

.github/workflows/ci.yml
negz: FWIW this is the case today as well (we're currently running this action with Go v1.15) so it's at least not a regression.

Jenkinsfile
```diff
@@ -1,151 +0,0 @@
-pipeline {
```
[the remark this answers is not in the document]
negz: Right - I intend to clean all of that up once this is approved.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,245 @@
...
+        env:
+          # We're using docker buildx, which doesn't actually load the images it
+          # builds by default. Specifying --load does so.
+          BUILD_ARGS: "--load"
+      
+      - name: Upload Artifacts (to GitHub)
+        uses: actions/upload-artifact@v2
+        with:
+          name: output
+          path: _output/**
```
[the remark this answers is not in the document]
negz: It does - as `output.zip`.

negz:
> Had you considered doing a serial build, then splitting out to concurrent running of unit tests / e2e tests / etc.? Is that even possible with GH actions?

It may be possible, but I suspect it would get tricky. We'd need to model the fan out as separate jobs, which would run on distinct VMs each with a pristine build environment. So we'd need to use the cache or artifacts to pass our Go build cache from the initial build onto the jobs that run the unit tests, e2e tests, etc. I don't think the complexity is worth it, given that it I don't believe it would actually speed up builds.

.github/workflows/ci.yml
negz: After experimenting with this, I think I prefer just gating on the env vars. This way we don't need to update the workflow to use it in our other orgs (e.g. `crossplane-contrib`).

.github/workflows/ci.yml
```diff
@@ -0,0 +1,245 @@
...
+          GIT_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      
+      - name: Promote Artifacts (in S3 and Docker Hub)
+        if: github.ref == 'refs/heads/master' && env.AWS_USR != '' && env.DOCKER_USR != ''
+        run: make -j2 promote
+        env:
+          BRANCH_NAME: master
+          CHANNEL: master
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_USR }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_PSW }}
```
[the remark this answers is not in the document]
negz: I experimented with this, but decided I preferred the way it was originally implemented. As mentioned elsewhere, it's a little complicated to pass state between jobs (and thus workflows) so to do this cleanly I think we'd need to:

1. Have one workflow that ran on branches and PRs to lint and test.
2. Have one workflow that only ran on PRs to ensure we can build artifacts for all platforms.
3. Have one workflow that only ran on master and release branches to upload artifacts for all platforms.

The second and third workflows would be almost identical apart from where they published; they'd both have the same build stages. I feel like there's less chance of them diverging if we keep them all in the same workflow.

hasheddan:
> I don't think the complexity is worth it, given that it I don't believe it would actually speed up builds.

True, we would still be constrained by the longest running action. That being said, it would be nice to not needlessly build three times. Definitely can be a future optimization though :+1:

hasheddan (approved):
LGTM

.github/workflows/ci.yml
khos2ow: Yes for sure, this is a really good candidate for future improvements when the initial phase of unblocking the faulty CI has passed. :+1:

.github/workflows/ci.yml
khos2ow: I cannot mention a solid use-case of this at the moment in the scope of crossplane codebases, specially when in a PR, but once in a while I had ended up in a situation that I needed to skip a pipeline (mostly publishing of some sort) on the master branch in which this became really handy.

Issue #1873, opened 2020-10-22
Migrate tag and promote Jenkins jobs to GitHub Actions


negz:
### What problem are you facing?

https://github.com/crossplane/crossplane/pull/1865 (and similar PRs to other repos under this org) switched Crossplane's CI from Jenkins to GitHub Actions. The only Jenkins jobs that now remain are the 'tag' and 'promote' jobs for each repo, as per the (slightly outdated - https://github.com/crossplane/crossplane/issues/1831) [release process](https://github.com/crossplane/crossplane/blob/master/docs/release/release-process.md#release-process). Keeping these jobs in Jenkins means we need to keep maintaining and paying for Jenkins infrastructure.

### How could Crossplane help solve your problem?

Crossplane could add equivalent GitHub Actions.

negz:
I believe the release process is currently roughly:

1. Get the master branch ready to cut release `V` (i.e. pin dependencies, etc)
1. Tag master as `V+1-rc`
1. Create release branch `release-V`
1. Create a PR to prepare branch `release-V`
1. Run the tag job on branch `release-V`
1. Run the CI workflow (i.e. GitHub Action) on branch `release-V`
1. Run the promote job on branch `release-V` to promote it to master.

https://github.com/crossplane/crossplane/blob/1ddb11/Jenkinsfile.tag#L30
https://github.com/crossplane/crossplane/blob/1ddb11/Jenkinsfile.promote#L24

The current tag and promote jobs simply run `make tag` and `make promote` with a given set of inputs (i.e. release version and channel). We should be able to add [manual triggers](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/) that faithfully replicate those jobs.

PR #1874, opened 2020-10-22
Migrate tag and promote jobs to GitHub Actions

negz:

Fixes https://github.com/crossplane/crossplane/issues/1873

I'm not sure how to test these, but I pretty much took the existing pipelines below, and adapted them per [this guide](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/). I believe once this is merged we should get a 'run workflow' button with a branch selector for these jobs in the 'Actions' tab.

https://github.com/crossplane/crossplane/blob/1ddb11/Jenkinsfile.tag#L30
https://github.com/crossplane/crossplane/blob/1ddb11/Jenkinsfile.promote#L24

I have:

`¯\_(ツ)_/¯`


.github/workflows/tag.yml
```diff
@@ -0,0 +1,27 @@
...
+        uses: actions/checkout@v2
+        with:
+          submodules: true
+
+      - name: Fetch History
+        run: git fetch --prune --unshallow
+
+      - name: Create Tag
+        run: make -j2 tag
+        env:
```
hasheddan, line 25: I think we optionally allow a commit hash to be specified in current job: https://github.com/crossplane/crossplane/blob/1ddb11225ea70c4ac425b50d13ead1d8fc7ed507/Jenkinsfile.tag#L32

hasheddan: Looks like github gives us the commit hash we can use if default if desired also: https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#workflow_dispatch

.github/workflows/promote.yml
```diff
@@ -0,0 +1,42 @@
+name: Promote
+
+on:
+  workflow_dispatch:
```
hasheddan, line 4: I checked to see if we could put branch restrictions on this, but it seems as though it can be run on any branch where the workflow is defined. This is likely fine as we need these jobs available on all branches (release and master)

.github/workflows/tag.yml
negz: I omitted this intentionally, under the assumption that it was a feature we never used. Do you think it's useful/necessary?

hasheddan: Does `make tag` use latest commit on branch if not specified? I am guessing yes, but just checking that we don't need to provide our own default.

negz: I believe it does - in the previous `Jenkinsfile.tag` it looks like COMMIT_HASH would be empty by default.

hasheddan (approved):
LGTM

PR #1883, opened 2020-10-23
Pin getting started configs to v0.13

negz:

These versions correspond to the release of Crossplane, not the provider versions. Pinning these allows us to modify the master tags while developing v0.14.

I have:

I've validated that I can install the new `:v0.13` packages, and that each of them installs the `Composition` I would expect. I haven't validated actually applying the compositions, but this seems low risk given the new tags are simply a result of pulling, retagging, and pushing the existing `:master` tags.

### DIFF
--- a/docs/getting-started/compose-infrastructure.md
+++ b/docs/getting-started/compose-infrastructure.md
@@ -73,7 +73,7 @@ We will now install a `Configuration` that:
 > an RDS instance that will allow traffic from the internet.
 
 ```console
-kubectl crossplane install configuration crossplane/getting-started-with-aws:master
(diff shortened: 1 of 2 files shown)


muvaf (approved):
LGTM! Assuming build & publish steps of the release pipeline include these configuration packages.

negz:
> Assuming build & publish steps of the release pipeline include these configuration packages.

Hrm, good point. The package infra steps don't exactly target these images (they use a dummy `my-package-repo/` by default, not `crossplane/`) but I think updating them to `:v0.13` makes sense to tie the examples together, and so that we can easily use the package infra steps to build the packages used in the compose infra steps.

PR #1888, opened 2020-10-23
DO NOT MERGE: Demonstrate no-op CI false positive

wonderflow:
How to reproduce?

1. give a broke commit (without signoff)
2. signoff and force push

### DIFF
--- a/apis/apis.go
+++ b/apis/apis.go
@@ -32,7 +32,7 @@ func init() {
 	)
 }
 
-// AddToSchemes may be used to add all resources defined in the project to a Scheme
+// Broke test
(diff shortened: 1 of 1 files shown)


wonderflow:
@negz

hasheddan:
Interesting.. seems like `detect-noop` should return with the status of the identical run :thinking:

negz:
Thanks @wonderflow! I'll see if we can fix this behaviour in the GitHub action upstream, or at least raise a bug.

negz:
A fix for this has been merged upstream. I've raised https://github.com/crossplane/crossplane/issues/1893 to track deploying the fix once it is released.

PR #1904, opened 2020-10-28
Make HEAD request when fetching a package digest

### LINKED ISSUE #1903: Make HEAD request when fetching a package digest

Currently we get the entire image manifest when checking for the package digest in the package manager controller. Since this is done semi-frequently (depending on the `packagePullPolicy`), the pull counts are driven up, which could cause rate-limiting issues (especially with [free-tier users](https://www.docker.com/pricing)).

`go-cotainerregistry` offers a [`remote.Head()`](https://github.com/google/go-containerregistry/issues/769) method which will make a `HEAD` request to dockerhub (or any other registry) when descriptor information is all that is needed. We should use it for this case.

hasheddan:

Updates the package manager controller to make a `HEAD` request instead of `GET` to acquire an image digest.

Fixes #1903 

I have:

Followed a typical package installation flow:
1. Install Crossplane
2. Install `provider-aws`: `k crossplane install provider crossplane/provider-aws:master`
3. See successful package install with correct digest

```
🤖 (crossplane) k get pkg
NAME                                      INSTALLED   HEALTHY   PACKAGE                          AGE
provider.pkg.crossplane.io/provider-aws   True        True      crossplane/provider-aws:master   119s

🤖 (crossplane) k get providerrevision
NAME                        HEALTHY   REVISION   IMAGE                            STATE    AGE
provider-aws-4ef40e146f8b   True      1          crossplane/provider-aws:master   Active   117s

🤖 (crossplane) k get pods -n crossplane-system
NAME                                                 READY   STATUS    RESTARTS   AGE
crossplane-5dd4f6f599-d877q                          1/1     Running   0          3m20s
crossplane-rbac-manager-55d4b59c7-mnzj6              1/1     Running   0          3m20s
oam-kubernetes-runtime-crossplane-67f4689d7f-4bjb8   1/1     Running   0          3m20s
provider-aws-4ef40e146f8b-7866d6b44c-d95n4           1/1     Running   0          113s
```

### DIFF
--- a/pkg/controller/pkg/manager/revisioner.go
+++ b/pkg/controller/pkg/manager/revisioner.go
@@ -28,7 +28,7 @@ import (
 )
 
 const (
-	errFetchPackage = "failed to fetch package from remote"
+	errFetchPackage = "failed to fetch package digest from remote"
 )
 
 // Revisioner extracts a revision name for a package source.
@@ -63,15 +63,11 @@ func (r *PackageRevisioner) Revision(ctx context.Context, p v1alpha1.Package) (s
 	if err != nil {
 		return "", err
 	}
-	img, err := r.fetcher.Fetch(ctx, ref, v1alpha1.RefNames(p.GetPackagePullSecrets()))
-	if err != nil {
+	d, err := r.fetcher.Head(ctx, ref, v1alpha1.RefNames(p.GetPackagePullSecrets()))
+	if err != nil || d == nil {
 		return "", errors.Wrap(err, errFetchPackage)
 	}
-	h, err := img.Digest()
-	if err != nil {
-		return "", err
-	}
-	return xpkg.FriendlyID(p.GetName(), h.Hex), nil
+	return xpkg.FriendlyID(p.GetName(), d.Digest.Hex), nil
 }
 
 // NopRevisioner returns an empty revision name.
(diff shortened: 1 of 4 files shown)


pkg/controller/pkg/manager/revisioner.go
```diff
@@ -63,15 +63,11 @@ func (r *PackageRevisioner) Revision(ctx context.Context, p v1alpha1.Package) (s
 	if err != nil {
 		return "", err
 	}
-	img, err := r.fetcher.Fetch(ctx, ref, v1alpha1.RefNames(p.GetPackagePullSecrets()))
-	if err != nil {
+	d, err := r.fetcher.Head(ctx, ref, v1alpha1.RefNames(p.GetPackagePullSecrets()))
+	if err != nil || d == nil {
```
negz, line 67: In what circumstance would `d` and `err` both be `nil`? If this is only covering the `NopFetcher`, perhaps that could return `&v1.Descriptor{}, nil` instead?

negz (approved):

pkg/controller/pkg/manager/revisioner.go
hasheddan: @negz they should not ever both be `nil`. However, `Head` does return a `*v1.Descriptor` so I think we should check for `nil` regardless even if in practice it should never occur

PR #1905, opened 2020-10-28
Introduce package ControllerConfig

hasheddan:
### Description of your changes

Introduces a cluster-scoped `ControllerConfig` type in the `pkg.crossplane.io/v1alpha1` group. `Provider` packages can reference a `ControllerConfig` to override some values in the packaged controller's `Deployment`. All labels and annotations on a `ControllerConfig` are also propagated to the `ServiceAccount` and `Deployment`.

Fixes #974 

I have:

- [x] Read and followed Crossplane's [contribution process].
- [x] Run `make reviewable test` to ensure this PR is ready for review.

### How has this code been tested

<details>
  <summary>Click to expand!</summary>
  
Tested with the following:
`ControllerConfig`
```
apiVersion: pkg.crossplane.io/v1alpha1
kind: ControllerConfig
metadata:
  name: aws-config
  labels:
    cool-label: cool-value
  annotations:
    cool-annotation: cool-value
spec:
  podSecurityContext:
    runAsUser: 1000
    runAsGroup: 1000
    fsGroup: 1000
  securityContext: {}
  env:
    - name: TEST
      value: "This is a test environment variable"
  resources:
    limits:
      memory: 200Mi
      cpu: 500m
```

`Provider`
```
apiVersion: pkg.crossplane.io/v1alpha1
kind: Provider
metadata:
  name: provider-aws
spec:
  package: crossplane/provider-aws:master
  packagePullPolicy: IfNotPresent
  controllerConfigRef:
    name: aws-config
```

Observed expected configuration of `Deployment` and `ServiceAccount` and controller running successfully (some fields removed for brevity:

`Deployment`
```
apiVersion: apps/v1
kind: Deployment
metadata:
  annotations:
    cool-annotation: cool-value
    deployment.kubernetes.io/revision: "2"
  creationTimestamp: "2020-10-28T22:25:08Z"
  generation: 2
  labels:
    cool-label: cool-value
  name: provider-aws-9d0c753df426
  namespace: default
  ownerReferences:
  - apiVersion: pkg.crossplane.io/v1alpha1
    controller: true
    kind: ProviderRevision
    name: provider-aws-9d0c753df426
    uid: e321c09c-e5f1-4716-af0b-856656a25f3e
  resourceVersion: "1329"
  selfLink: /apis/apps/v1/namespaces/default/deployments/provider-aws-9d0c753df426
  uid: e4aabc94-c1dc-45ee-a464-6fb1bb5f01bb
spec:
  progressDeadlineSeconds: 600
[...]

PR #1908, opened 2020-10-29
Add replicas and args to package ControllerConfig spec

### LINKED ISSUE #1638: Env var support for packages to enable users to supply proxy credentials

All cloud SDKs support proxies through environment variables. For example, AWS SDK expects you to have `HTTP_PROXY` env var defined, which is the standard of Go http library. However, Crossplane package manager doesn't allow entry of any kind of environment variable during installation of the providers. 

https://github.com/crossplane/provider-aws/issues/199#issuecomment-649773923 is a community member requesting this.

Allow specification of environment variables for the controller that will be installed via package manager so that users can define proxy env vars in corporate networks with no direct access to the internet.

khos2ow:

Following up #1905 in regards to enabling leader election at #1857, adding `replicas` to `ControllerConfig` spec in order for the downstream providers to be able to pass this value if they chose to and specially if they need to enable leader election on provider controller managers.

Also adding `args` to `ControllerConfig` spec too, which can be used to pass additional arguments for the deployment by the downstream providers.

Fixes #1638

I have:

- Create a new `ControllerConfig`:
  ```yaml
  kind: ControllerConfig
  metadata:
    name: aws-config
  spec:
    replicas: 2
    args:
      - --no-leader-election
  ```
- with the `Provider`:
  ```yaml
  kind: Provider
  metadata:
    name: provider-aws
  spec:
    package: crossplane/provider-aws:master
    controllerConfigRef:
      name: aws-config
  ```

- look for provider's `Deployment` to have relevant information:
  ```yaml
  # omitted items for brevity
  apiVersion: apps/v1
  kind: Deployment
  metadata:
    name: provider-aws-56bfa7d9d4ce
  spec:
    replicas: 2
    ...
    template:
      spec:
        containers:
        - args:
          - --no-leader-election
          image: crossplane/provider-aws-controller:v0.14.0-rc.0.4.g0b1c4b04
          name: provider-aws
  ...
  ```


hasheddan (commented):
This is awesome @khos2ow! Thanks for adding 🙂 have you done any manual testing for this? I don't immediately see any issues with running multiple replicas 👍

khos2ow:
@hasheddan No, not yet, I actually wanted to open this as a draft and update the description after my test but clicked the wrong button!

> I don't immediately see any issues with running multiple replicas

I believe if we run this as is without the leader election enabled within the provider's controller manager itself (which I'll be opening corresponding PRs for core providers soon), this will produce issues.

hasheddan:
> I believe if we run this as is without the leader election enabled within the provider's controller manager itself (which I'll be opening corresponding PRs for core providers soon), this will produce issues.

For sure, might be worth calling that out in the field description 👍

khos2ow:
That's actually a great idea. Beforehand I'll try to run this as is and see what happens for replicas of more than one without leader election capability baked into the providers.

pkg/controller/pkg/revision/deployment.go
```diff
@@ -133,5 +146,20 @@ func buildProviderDeployment(provider *metav1alpha1.Provider, revision v1alpha1.
 			d.Spec.Template.Spec.Containers[0].Env = cc.Spec.Env
 		}
 	}
-	return s, d
+	if err := ensureLeaderElection(d); err != nil {
+		return nil, nil, err // replicas > 1 but LEADER_ELECTION is not set or is not 'true'
+	}
```
khos2ow, line 151: @negz @hasheddan I've decided to error out if the provided replicas is more than `1` and leader election is not enabled. This prevents a user to make a mistake and leaves a log trace of what happened for them to view and fix. What do you think?

hasheddan: Hmm I'm not sure we should enforce this requirement. There are a variety of ways a user could misconfigure a controller deployment here and it feels more future proof to not lock in provider authors to a specific environment variable.

khos2ow: I'm not sure if I got your point completely. This restriction is only for number of replica (which is 1 by default). It means if a user intentionally wants to bump it for a provider they need to use the new `ControllerConfig` you added, and by convention they need to also pass the extra environment variable needed for this.

```yaml
kind: ControllerConfig
metadata:
  name: aws-config
spec:
  replicas: 2
  env:
    - name: "LEADER_ELECTION"
      value: "true"
```

I originally had the idea to auto-populate the environment (if missing) here if `replicas > 1`, but thought this might feel more like a automagic to the user rather than throwing an actual error for it. This can be described (e.g. in a README or https://doc.crds.dev/ maybe).

khos2ow: Considering the CRD reads:

```go
type ControllerConfigSpec struct {
	// Number of desired pods. This is a pointer to distinguish between explicit
	// zero and not specified. Defaults to 1.
	// Note: If number more than 1 is set, environment variable "LEADER_ELECTION"
	// must exist and must be true.
	// +optional
	Replicas *int32 `json:"replicas,omitempty"`
...
}
```

hasheddan: > // Note: If number more than 1 is set, environment variable "LEADER_ELECTION"
	// must exist and must be true.

My point is that I don't think we should tie replicas to a specific environment variable being set. If a provider wanted to use a different environment variable, or they had leader election enabled by default, then a user may just want to set a replica number without the environment variable. It is true that a user could accidentally supply a bad configuration that caused controllers to fight with each other, but I don't think we should explicitly restrict that from happening. Ensuring this environment variable is set doesn't even guarantee that the configuration is valid (i.e. a provider may not honor `LEADER_ELECTION`), and in that case, enforcing this could actually cause additional confusion.

khos2ow: Oh I see your point! That makes sense. By this approach I was trying to implement a way to pass input corresponding to [this](https://github.com/crossplane/provider-aws/pull/408/files#diff-36b6d20eb5aea66cf39f8d94111bd96513626ef7f61459f0d9e8e9507ded1d17R39) line. I've looked around a bit and the only options I could have thought were:

- pass it as environment variable to container
- pass it as `args` to container

Both cases needed a change of provider's `Deployment` and environment variable was a safer approach.

Now, I'm thinking maybe we should remove the flag from providers and move the leader election enablement to the providers (i.e. enabled by default or disabled altogether).

negz: I agree with @hasheddan here; we should allow `ControllerConfig` authors to specify their desired number of replicas regardless of whether or not they have enabled leader election (either by setting the `LEADER_ELECTION` env var or some other means). I think this is a place where we should rely on convention (i.e. expect users to provide correct configuration) rather than enforcement.

So, I like the pattern you've established in https://github.com/crossplane/provider-aws/pull/408 (and similar PRs) that adds a flag and env var that can be used to enable leader election; I think we should keep that. I think the only change we should make is to remove the `ensureLeaderElection` check here.

khos2ow: Great, will update. On the same subject, to be more feature complete and flexible, maybe we can add `Args` to `Spec.Template.Spec.Containers` too? This will not be a blocker of this PR but since the file is being change I can add it too and rename the title of the PR too.

hasheddan: @khos2ow I would support adding `Args` :) Would you like to do it in this PR or in a follow-up?

apis/pkg/v1alpha1/config.go
```diff
@@ -25,6 +25,12 @@ import (
 // Values provided will override package manager defaults. Labels and
 // annotations are passed to both the controller Deployment and ServiceAccount.
 type ControllerConfigSpec struct {
+	// Number of desired pods. This is a pointer to distinguish between explicit
+	// zero and not specified. Defaults to 1.
+	// Note: If number more than 1 is set, environment variable "LEADER_ELECTION"
+	// must exist and must be true.
```
hasheddan, line 31: Could you adjust this note to just mention that if more than 1 replica is set for a controller and leader election is not enabled then controllers could conflict?

pkg/controller/pkg/revision/deployment.go
khos2ow: In this particular case, with only one line change needed in the same file of this PR I prefer to do it here, otherwise opening a follow-up PR is completely fine by me if you prefer that.

hasheddan: Here is great :)

apis/pkg/v1alpha1/config.go
```diff
@@ -89,6 +96,16 @@ type ControllerConfigSpec struct {
...
+	// Arguments to the entrypoint.
+	// The docker image's CMD is used if this is not provided.
+	// Variable references $(VAR_NAME) are expanded using the container's environment. If a variable
+	// cannot be resolved, the reference in the input string will be unchanged. The $(VAR_NAME) syntax
+	// can be escaped with a double $$, ie: $$(VAR_NAME). Escaped references will never be expanded,
+	// regardless of whether the variable exists or not.
+	// Cannot be updated.
+	// More info: https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#running-a-command-in-a-shell
+	// +optional
+	Args []string `json:"args,omitempty" protobuf:"bytes,4,rep,name=args"`
```
hasheddan, line 108:
```suggestion
	Args []string `json:"args,omitempty"`
```

hasheddan (changes requested):

apis/pkg/v1alpha1/config.go
khos2ow: Nice catch! copy-paste issue :sweat_smile:

hasheddan:
@khos2ow could you rebase this on `master` and we can get it in for tomorrow's release :)

hasheddan:
Note: this fixes https://github.com/crossplane/crossplane/issues/1638

hasheddan (approved):

Issue #1910, opened 2020-10-30
Composed resources are leaked when there is XR write contention


negz:
### What happened?

While working on https://github.com/crossplane/crossplane/issues/1649 I've noticed that the XR controller leaks composed resources when there is higher than usual write contention (e.g. because an update to the XR triggers both the claim and composite resource controllers to update the XR). This is because the composite controller first creates a managed resource using `metadata.generateName` then, records the generated name by updating the XR's `spec.resourceRefs`. If the update fails the composite controller doesn't know it has already created the composed resource, and will do so again on the next reconcile.

### How can we reproduce it?

Presumably by triggering many updates to a newly created XR in a very tight loop.

### What environment did it happen in?
Crossplane version: https://github.com/crossplane/crossplane/commit/28224d92322a9caeaea12e6a2df97632358184ff

muvaf:
I was thinking that `generateName` actually guarantees uniqueness but from what I can tell, that's [not the case](https://github.com/kubernetes/apiserver/blob/e9e4beec4b157a40f6c742782d74faf45c2b280e/pkg/storage/names/generate.go#L53). I think we can generate a reference, save it to `resourceRefs` and then create the resource with that name that we're sure saved to the composite resource.

negz:
I haven't completely joined the dots, but I believe there is middleware that implements the described behaviour of `generateName`:

> If this field is specified and the generated name exists, the server will NOT return a 409 - instead, it will either return 201 Created or 500 with Reason ServerTimeout indicating a unique name could not be found in the time allotted, and the client should retry (optionally after the time indicated in the Retry-After header).

https://github.com/kubernetes/apiserver/blob/release-1.19/pkg/registry/generic/registry/store.go#L367
https://github.com/kubernetes/apiserver/blob/release-1.19/pkg/registry/rest/create.go#L137
https://github.com/kubernetes/apiserver/blob/release-1.19/pkg/registry/generic/registry/dryrun.go#L38

It seems like we can actually dry-run create in order to generate a name. Efficiency wise it's about the same as generating a name locally and dry-run creating it to see whether it's available, but I prefer to use GenerateName to ensure that our name generation logic never diverges from the API server's.

muvaf:
Dry-run sounds good. As mentioned in https://github.com/crossplane/crossplane/pull/1916 , it gives a clue to users about the errors before creating anything, which is a nice UX plus.

PR #1916, opened 2020-10-31
Persist references to composed resources before creating them

negz:

Fixes https://github.com/crossplane/crossplane/issues/1910

This commit refactors the XR reconciler such that new composed resources are named (by the API server via a dry-run create) before they are created. This allows us to persist a reference the new composed resource before it is created, which in turn ensures we don't leak composed resources if we fail to update our references. Instead we'll notice that we already named the composed resource in question (by reading our references) and use that name.

I have:

I've run it with the GCP and AWS (with VPC) composition examples and confirmed it works as expected in both cases.


pkg/controller/apiextensions/composite/reconciler.go
```diff
@@ -265,43 +260,65 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error)
 	// array with the same length. Then copy the already provisioned ones into
 	// that array to not create new ones because composed reconciler assumes that
 	// if the reference is empty, it needs to create the resource.
+
+	// TODO(negz): This approach means that the resources of a Composition are
+	// effectively append only. We may want to reconsider this per
+	// https://github.com/crossplane/crossplane/issues/1909
 	refs := make([]corev1.ObjectReference, len(comp.Spec.Resources))
```
muvaf, line 267: Not related to this PR, but we should get rid of `ObjectReference` here due to https://github.com/crossplane/crossplane-runtime/issues/49

muvaf (commented):
Looks pretty good to me. As I said yesterday in the meeting, I was leaning towards just generating the name ourselves instead of dry-run since that's what api-server does. However, dry-run actually gives a chance for surfacing errors in the template without actually creating anything. So, users would be able to make corrections before any creation, which is a nice improvement.

And thanks for getting rid of `composed` package. I've been meaning to do a refactor there but couldn't get around it. At the time, I was imagining more functionality for that package but that never got realized. Having it in a separate file seems like a better level of separation.

hasheddan (approved):
LGTM!

pkg/controller/apiextensions/composite/reconciler.go
negz: Yeah, I'm definitely keen to do that. I suspect it might be best done in its own PR though - I considered doing it in this one for a split second but decided it would add too much unrelated churn.

negz:
PTAL! No real changes since this was first opened; I've rebased it to accommodate recent changes to master and added some more tests.

hasheddan (approved):
Still LGTM! Thanks @negz!

Issue #1917, opened 2020-11-01
managed resource connection secret updates do not get propagated to the claim namespace


jbw976:
### What happened?

I have a cluster composition that uses EKS as the underlying managed resource. The auth tokens given out by EKS have a [TTL of 15 min](https://aws.github.io/aws-eks-best-practices/iam/#:~:text=The%20token%20has%20a%20time,each%20time%20the%20token%20expires.), so the connection secret for the EKS managed resources gets updated often.

It appears that updates to the connection secret do not get propagated back up the composition hierarchy and to the claim's namespace.  So when trying to use the claim's connection secret, it is stale and results in `error: You must be logged in to the server (Unauthorized)`.

### How can we reproduce it?

The [AWS reference platform](https://github.com/upbound/platform-ref-aws) is an example of this behavior.  The composition hierarchy is:

* Cluster claim (namespaced)
  * CompositeCluster (composite resource)
    * `eks.aws.platformref.crossplane.io` (composite resource)
      * `cluster.eks.aws.crossplane.io` (managed resource)

When examining the connection secrets at those 4 levels, the only diff is between the claim and the CompositeCluster, so it appears the managed resource secret update is getting propagated through the compositions, but not the final hop to the namespaced claim.

### What environment did it happen in?
Crossplane version: `v0.14.0-rc.102.g6e1b4ce2`

negz:
As you point out, the propagation from the actual credentials in AWS to the claim secret works as follows:

1. The EKS credentials change inside AWS.
2. Every 30 to 60 seconds the provider-aws EKS controller polls the AWS API for new changes. If the credentials have changed they will be propagated to the EKS cluster's connection secret.
3. Every 30 to 60 seconds the composite resource controller polls all of its composed resources. If any composed resource credential secret keys have changed (and those secret keys are mapped to the composite's connection details) they will be propagated to the composite's connection secret.
4. On every reconcile the claim controller propagates its composite resource's connection secret.

I believe the culprit here is the fact that we don't queue claim reconciles every 30-60 seconds, because the controller watches both the claim and composite and thus assumes a reconcile will be queued if anything changes about those resources. Unfortunately we don't watch the composite's connection secret.

There's a few fixes here:

1. Configure the claim reconciler to speculatively requeue every 60 seconds. This would result in an almost 3 minute propagation delay from the AWS API to the claim secret in the pathological case; i.e. at each propagation step we happen to not requeue until almost a minute after the previous step's credentials changed.
[...]

PR #1918, opened 2020-11-03
Reconcile XRCs when XR connection secrets change

negz:

Fixes https://github.com/crossplane/crossplane/issues/1917

Enqueueing an XRC reconcile when an XR connection secret changes ensures we always propagate XR connection secrets to XRC connection secrets when they change.

I have:

I've tested this by building and installing https://github.com/upbound/platform-ref-aws/tree/6137abb642d9ad6e336f29c94f08ebece1f98ee9 and confirming that as the underlying EKS cluster's connection secret changes over time, the `Cluster` claim's connection secret changes to match it - I can use the claim's connection secret to connect to the EKS cluster long after the EKS token should have rotated;

```console
$ kubectl get cluster.eks
NAME                                   READY   SYNCED   AGE
platform-ref-aws-cluster-q4hdw-k6t8r   True    True     29m

$ kubectl get -o json secret platform-ref-aws-cluster |jq -r '.data.kubeconfig|@base64d'> claim.kcfg
$ kubectl --kubeconfig claim.kcfg -n kube-system get po
NAME                       READY   STATUS    RESTARTS   AGE
aws-node-9v26x             1/1     Running   0          14m
aws-node-s2dt7             1/1     Running   0          14m
aws-node-st6gj             1/1     Running   0          14m
coredns-5946c5d67c-bjkb8   1/1     Running   0          18m
coredns-5946c5d67c-xgzk6   1/1     Running   0          18m
kube-proxy-9jb7r           1/1     Running   0          14m
kube-proxy-r827r           1/1     Running   0          14m
kube-proxy-wmkr5           1/1     Running   0          14m
```

### DIFF
--- a/pkg/controller/apiextensions/offered/watch.go
+++ b/pkg/controller/apiextensions/offered/watch.go
@@ -17,9 +17,15 @@ limitations under the License.
 package offered
 
 import (
+	"context"
+
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/workqueue"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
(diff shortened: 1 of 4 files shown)


pkg/controller/apiextensions/offered/watch.go
```diff
@@ -84,3 +90,61 @@ func addClaim(obj runtime.Object, queue adder) {
...
+		return
+	}
+
+	nn := types.NamespacedName{Name: ref.Name}
+	cp := &composite.Unstructured{}
+	cp.SetAPIVersion(ref.APIVersion)
+	cp.SetKind(ref.Kind)
+	if err := e.client.Get(context.TODO(), nn, cp); err != nil {
+		return
+	}
```
negz, line 145: I don't love that we need a client here to get the secret's XR in order to resolve its claim ref - especially since we have no context and no opportunity to return an error here -  but at least we should be reading from cache. 

I keep going back and forth on whether we should instead annotate XR connection secrets with their claim's GVK, namespace, and name. This would allow us to skip the intermediate get of the XR and just enqueue a request per the secret's annotations. Performance wise they're about the same - the main reason to go with this approach is that it lets us make the connection with the information we already have rather than adding new information (i.e. annotations).

CC @muvaf what's your opinion here?

muvaf: Yeah, I think that sounds like a good compromise.

Ideally, I'd want XRC controller to be notified when XR changes, which would be caused by a change in secret so that the information chain is still straight and XR secret doesn't need to know about XRC. But there doesn't seem to be a sane way to do achieve that.

negz: > Ideally, I'd want XRC controller to be notified when XR changes, which would be caused by a change in secret so that the information chain is still straight and XR secret doesn't need to know about XRC. But there doesn't seem to be a sane way to do achieve that.

One approach here could be to modify the XR somehow when its connection secret changes. i.e.:

1. XRC controller is watching XR.
1. XR controller is watching XR connection secret.
1. When XR connection secret changes, we set `status.connectionDetails.lastRefreshTime` or something.

This way the change to the secret would trigger a change to the XR which would trigger a change to the XRC.

muvaf: > When XR connection secret changes

That's kind of hard to detect though unless we're willing to assume _every_ event is a change, which could potentially render too frequent reconciliations.

FWIW, we'll call `GET` for every connection secret with an owner, which is most of what we publish, so it will be called very frequently in different go routines. Hope the cache helps for not hitting the api-server for those calls. It could really hog it after a few composite resources.

hasheddan (approved):
LGTM

pkg/controller/apiextensions/offered/watch.go
negz: > That's kind of hard to detect though unless we're willing to assume every event is a change, which could potentially render too frequent reconciliations.

Yeah. I'm thinking we'd have the XR reconciler report this, i.e. update the connection secret publishing behaviour to report whether the secret actually changed. That should be doable, but it's definitely a bigger diff to do it. I'm still on the fence so I might take a shot at that implementation in a competing PR so we can compare.

negz: Okay, I've opened https://github.com/crossplane/crossplane/pull/1927 which uses the implementation I described. In the end I much prefer that implementation (despite it being a larger diff):

* It feels more elegant than having to make an API call (even to cache) in order to map secrets to XRs.
* Including the time we last published connection details in the status seems handy for debugging.
* Watching connection secret changes indirectly via their associated XR is more future proof, as it would allow us to 'watch' for connection detail changes even if those details were stored in an external system such as Vault, rather than a Secret.

muvaf: Yeah, I like that implementation more, too. It seems you overcame the frequent event problem by comparing the content of the secret. So, we'll get an event only if the content of the secret changes, which is precisely what we care.

PR #1919, opened 2020-11-03
Watch for CRDs and ServiceAccounts owned by ProviderRevisions

negz:

I recently encountered a ClusterRoleBinding for a ProviderRevision that had no subject. I suspect this is because there was enough of a delay between when the ProviderRevision was created and when its ServiceAccount was created that this controller listed ServiceAccounts before the ProviderRevision's ServiceAccount was created. This commit should cause the ProviderRevision to be queued for a reconcile when its ServiceAccount is added.

I have:

I've tested this by:

1. Installing `crossplane/provider-aws:master`.
2. Confirming the `ProviderRevision`'s `ServiceAccount` was added by the RBAC manager as a subject of its `ClusterRole`.
3. Deleting said `ServiceAccount`.
4. Confirming that the `ClusterRole` was updated to have no subjects.
5. Confirming that the `ClusterRole` was updated to once again include the `ServiceAccount` once the `ProviderRevision` controller noticed it was gone and recreated it.

### DIFF
--- a/pkg/controller/rbac/provider/binding/reconciler.go
+++ b/pkg/controller/rbac/provider/binding/reconciler.go
@@ -27,8 +27,10 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	kcontroller "sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	"github.com/crossplane/crossplane-runtime/pkg/event"
 	"github.com/crossplane/crossplane-runtime/pkg/logging"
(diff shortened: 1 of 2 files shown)


pkg/controller/rbac/provider/binding/watch.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+func addRevision(obj runtime.Object, queue adder) {
+	sa, ok := obj.(*corev1.ServiceAccount)
+	if !ok {
+		return
+	}
+	for _, ref := range sa.GetOwnerReferences() {
+		if ref.APIVersion != v1alpha1.ProviderRevisionGroupVersionKind.GroupVersion().String() {
+			return
```
muvaf, line 70:
```suggestion
			continue
```
Is the assumption that there will ever be only one ownerref feasible?

negz: There will most likely only be one owner ref and it will be a ProviderRevision, but my intent here was to assess all of them just in case so I agree `continue` would be better.

hasheddan (commented):
This looks good to me in regards to addressing the observed problem. I wonder if we could more generally address it by just having the RBAC controller watch for service accounts with a specific label or annotation, rather than watching for `ProviderRevision` explicitly. This would allow for more flexibility if we introduce an additional package type in the future that needs permissions granted.

negz:
Back to draft - it occurred to me at the last second that I could just use `handlers.EnqueueRequestForOwner{}` from controller-runtime here, so I'm going to do that.

negz:
Okay, vastly simplified and re-tested per the process in the description.

hasheddan (approved):

PR #1925, opened 2020-11-04
Initial commit of Tilt for local development

khos2ow:

This PR introduces usage of [Tilt](https://github.com/tilt-dev/tilt). Using Tilt will improve developer's experience significantly while working on Crossplane. The workflow will be as simple as:

- spin up a local cluster (using new `kind-with-registry.sh` script)
- `tilt up` on the root of repository
    
Tilt will watch the changes in required folders (e.g. `cmd`, `pkg`, ... and ignore for example `docs`, `hacks`, ...) and build and live-reload the resources needed to be deployed.

~~Note that as of now, this will only work on `crossplane/crossplane`, consecutively I'll try to add the support for providers as well. So that one can clone Crossplane and provider locally and _bundle_ them up together for local rapid deployment.~~

Consecutively if you are working on a Crossplane provider and you want it to get deployed with Tilt you can add the Provider detail to a `gitignore`-d file called `tilt-providers.json` in the root of `crossplane/crossplane` repository. Example:

`tilt-providers.json`:

```json
{
    "provider-aws": {
        "enabled": true,
        "context": "/path/to/local/clone/provider-aws"
    },
    "provider-gcp": {
        "enabled": false,
        "package_owner": "crossplane",
        "package_ref": "alpha"
    }
}
```

You also would be able to set some configurations of overall state of deployment in `tilt-settings.json`.

`tilt-settings.json`:

```json/text
{
    "args": [],                        // args to pass to pod
    "debug": false,                    // enable debug mode
    "namespace": "crossplane-system"   // namespace to deploy provider into
}
```

This will deploy two providers as part of running Tilt:

- `provider-gcp` from `crossplane` Dockerhub namespace of `alpha` channel
- `provider-aws` built locally from `/path/to/local/clone/provider-aws`
  - note that `context` can also be relative to crossplane local cloned folder

**NOTE:** this PR depends on https://github.com/upbound/build/pull/127 and must be merged after that one.

I have:

### DIFF
--- a/cluster/local/README.md
+++ b/cluster/local/README.md
@@ -2,5 +2,91 @@
 
 This directory contains scripts that automate common local development flows for
 Crossplane, allowing you to deploy your local build of Crossplane to a `kind`
-cluster. Run [kind.sh](./kind.sh) to setup a single-node kind Kubernetes
-cluster.
+cluster. [kind-with-registry.sh] is being used to setup a single-node
+kind Kubernetes cluster.
+
+## Using Tilt
+
+[Tilt] can be used to reduce the repetative tasks you might have to do to keep your
+local deployment up to date. It watches required folders locally and takes action
+accordingly (e.g. build crossplane binary and image, deploy it on local cluster, etc).
+
+In order to use tilt, the following steps are needed:
+
+1. Prepare local `.work/` folder (only once)
+
+   ```sh
+   make local.prepare
+   ```
+
+2. Spin up a kind cluster:
+
+    ```sh
+    USE_TILT=true make kind.up
+    ```
+
+3. Update helm chart dependency, this must be done [outside of Tilt]:
+
+    ```sh
+    make local.helmdep
+    ```
+
+    This is needed once or when `cluster/charts/crossplane/charts` is missing or any of the
+    dependencies of Crossplane chart have been changed, otherwise you'd see this error:
+
+    ```text
+    Error: found in Chart.yaml, but missing in charts/ directory
+    ```
+
+4. Use tilt!
+
+    ```sh
+    make tilt.up
+    ```
+
+    If you've encountered `Error: notify.Add(...): no space left on device` check out
+    [this comment] for a solution.
+
+To see all available options for local developement look at [build/makelib/local.mk].
+
+### Working with Providers
+
+If you are working on a Crossplane provider and you want it to get deployed with Tilt
+you can add the provider details to a `gitignore`-d file called `tilt-providers.json`
+in the root of `crossplane/crossplane` repository.
+
+```json
+{
+    "provider-aws": {
+        "enabled": true,
+        "context": "/path/to/local/clone/provider-aws"
+    },
+    "provider-gcp": {
+        "enabled": true,
+        "package_owner": "crossplane",
+        "package_ref": "alpha"
+    }
+}
+```
+
+You also would be able to set some of the configuration options for the overall state of the deployment in `tilt-settings.json`.
+
+```json
+{
+    "args": [],
+    "debug": false,
+    "namespace": "crossplane-system"
+}
+```
+
+This will deploy two providers as part of running Tilt:
+
+- `provider-gcp` from `crossplane` Dockerhub namespace of `alpha` channel
+- `provider-aws` built locally from `/path/to/local/clone/provider-aws`
+  - note that `context` can also be relative to crossplane local cloned folder
+
+[kind-with-registry.sh]: ./kind-with-registry.sh
+[Tilt]: https://tilt.dev
+[outside of Tilt]: https://docs.tilt.dev/helm.html#sub-charts-and-requirementstxt
+[this comment]: https://github.com/tilt-dev/tilt/issues/2079#issuecomment-632226113
+[build/makelib/local.mk]: build/makelib/local.mk
--- a/cluster/local/provider.Tiltfile
+++ b/cluster/local/provider.Tiltfile
@@ -1,0 +1,196 @@
+####################################################################################
+# Crossplane Providers helper functions
+#
+# These are helper functions to build, deploy, and live-reload providers.
+# Note that these functions are shared with both:
+#
+# 1) core crossplane repo
+# 2) any standalone provider
+####################################################################################
+
+# Register CRDs which expose `image` for Tilt to be able to find them and
+# build the corresponding docker image for it.
+# https://docs.tilt.dev/custom_resource.html#does-it-contain-images
+k8s_kind('ControllerConfig', image_json_path = '{.spec.image}')
+
+# Build a crossplane provider
+#
+# 1. build and apply CRDs
+# 2. build and deploy controller manager
+def build_provider(name, provider, settings):
+    _generate_crds(provider, settings)
+    _build_controller(name, provider, settings)
+
+# Generate and apply crossplane provider's CRDs
+def _generate_crds(provider, settings):
+    context = provider.get('context')
+
+    # Prefix each dependency with context so Tilt can watch the correct paths
+    # for changes. At the same time keep track of 'zz_generated.*' files to
+    # ignore (this is needed to prevent being stuck in an infinite loop.)
+    crd_deps = []
+    ignore_deps = []
+    for d in provider.get('crd_deps', []):
+        crd_deps.append(context + '/' + d)
+        ignore_deps.append(context + '/' + d + '/**/zz_generated.*')
+        
+    # Generate crossplane provider CRDs out of API definition
+    local_resource(
+        'gen-crds-' + provider['short_name'],
+        'cd ' + context + ' && make generate',
+        deps = crd_deps,
+        ignore = ignore_deps
+    )
+
+    # Find all the CRDs manifests files
+    crds_folder = context + '/' + provider.get('crds_folder', 'package/crds')
+    files = listdir(crds_folder)
+
+    # Extract name of the CRDs to show on Tilt UI
+    crds_name = []
+    for f in files:
+        crds_name.append(read_yaml(f)['metadata']['name'] + ':customresourcedefinition')
+
+    # Apply CRD manifests
+    k8s_yaml(files)
+    k8s_resource(
+        new_name = 'crds-' + provider['short_name'],
+        objects = crds_name,
+        resource_deps = settings.get('resource_deps')
+    )
+
+# Build crossplane provider binary and docker image. Binary is built for
+# linux-amd64 (to be only used inside a container.) The binary is stored
+# in crossplane cloned folder at _output/tilt/
+def _build_controller(name, provider, settings):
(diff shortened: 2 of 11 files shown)


khos2ow:
Converting this to draft for now as I've thought of more enhancement for it.

khos2ow:
This should be good to try out and review.

cluster/charts/crossplane/values.yaml.tmpl
```diff
@@ -33,6 +33,12 @@ resourcesCrossplane:
     cpu: 100m
     memory: 256Mi
 
+securityContextCrossplane:
+  runAsUser: 2000
+  runAsGroup: 2000
+  allowPrivilegeEscalation: false
+  readOnlyRootFilesystem: true
```
hasheddan, line 40: @khos2ow could you break these parameters out into a separate PR? I would definitely like for these values to be configurable such that crossplane can run on platforms that restrict setting them, but would like for the change to be tracked as a distinct PR :)

khos2ow: Sure, #1983 should do it. I'm gonna wait for that to land and then rebase it here.

cluster/local/README.md
```diff
@@ -2,5 +2,91 @@
...
 Crossplane, allowing you to deploy your local build of Crossplane to a `kind`
-cluster. Run [kind.sh](./kind.sh) to setup a single-node kind Kubernetes
-cluster.
+cluster. [kind-with-registry.sh] is being used to setup a single-node
+kind Kubernetes cluster.
+
+## Using Tilt
+
+[Tilt] can be used to reduce the repetative tasks you might have to do to keep your
+local deployment up to date. It watches required folders locally and take action
```
mpetason, line 11:
```suggestion
local deployment up to date. It watches required folders locally and takes action
```

cluster/local/README.md
```diff
@@ -2,5 +2,91 @@
...
+accordingly (e.g. build crossplane binary and image, deploy it on local cluster, etc).
+
+In order to use tilt, following steps are needed:
```
mpetason, line 14:
```suggestion
In order to use tilt, the following steps are needed:
```

cluster/local/README.md
```diff
@@ -2,5 +2,91 @@
...
+
+    If you've encountered `Error: notify.Add(...): no space left on device` check out
+    [this comment] for a solution.
+
+To see all available options for local developement look at [build/makelib/local.mk].
+
+### Working with Providers
+
+If you are working on a Crossplane provider and you want it to get deployed with Tilt
+you can add the provider detail to a `gitignore`-d file called `tilt-providers.json`
```
mpetason, line 55:
```suggestion
you can add the provider details to a `gitignore`-d file called `tilt-providers.json`
```

cluster/local/README.md
```diff
@@ -2,5 +2,91 @@
...
+    },
+    "provider-gcp": {
+        "enabled": true,
+        "package_owner": "crossplane",
+        "package_ref": "alpha"
+    }
+}
+```
+
+You also would be able to set some configurations of overall state of deployment in `tilt-settings.json`.
```
mpetason, line 72:
```suggestion
You also would be able to set some of the configuration options for the overall state of the deployment in `tilt-settings.json`.
```
or maybe "of the overall state" - or if it's all of the configuration then "you would also be able to set the configuration for the overall state of the deployment"

mpetason (commented):
small updates that may read a little bit better. overall looks pretty good.

cluster/local/provider.Tiltfile
```diff
@@ -0,0 +1,196 @@
...
+
+    # Build crossplane provider docker image
+    dockerfile_contents = '\n'.join([
+        'FROM alpine:3.7',
+        'RUN apk --no-cache add ca-certificates bash',
+        'ADD ' + build_output + ' /usr/local/bin/crossplane-provider',
+        # TODO(khos2ow): change this URL right before
+        # merging the PR in crossplane/crossplane!
+        #
+        # https://github.com/crossplane/crossplane/pull/1925
```
hasheddan, line 95: @khos2ow does this need to be changed?

cluster/local/provider.Tiltfile
```diff
@@ -0,0 +1,196 @@
...
+        'cd ' + context + ' && CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o ' + core_path + '/' + build_output + ' ' + go_main,
+        deps = cmd_deps,
+        ignore = ignore_deps
+    )
+
+    # Build crossplane provider docker image
+    dockerfile_contents = '\n'.join([
+        'FROM alpine:3.7',
+        'RUN apk --no-cache add ca-certificates bash',
+        'ADD ' + build_output + ' /usr/local/bin/crossplane-provider',
```
hasheddan, line 91: Can we use the same `Dockerfile` we use for actually building the images?

cluster/local/provider.Tiltfile
```diff
@@ -0,0 +1,196 @@
...
+    ignore_deps = []
+    for d in provider.get('cmd_deps', []):
+        cmd_deps.append(context + '/' + d)
+        ignore_deps.append(context + '/' + d + '/**/zz_generated.*')
+
+    # Build crossplane provider
+    build_output = '_output/tilt/' + name
+    local_resource(
+        'build-' + provider['short_name'],
+        'cd ' + context + ' && CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o ' + core_path + '/' + build_output + ' ' + go_main,
```
hasheddan, line 82: Is there an advantage here to just using `make build` and copying over to the necessary directory if necessary? That would ensure consistency between development and build.

cluster/local/provider.Tiltfile
```diff
@@ -0,0 +1,196 @@
...
+
+# Build a crossplane provider
+#
+# 1. build and apply CRDs
+# 2. build and deploy controller manager
+def build_provider(name, provider, settings):
+    _generate_crds(provider, settings)
+    _build_controller(name, provider, settings)
+
+# Generate and apply crossplane provider's CRDs
```
hasheddan, line 24: This may no longer be needed as Crossplane has an `initContainer` that manages CRDs.

hasheddan (commented):
@khos2ow thanks for working on this and apologies for the delayed reviews :/ This looks pretty good, but I am also a bit hesitant as I imagine it will require ongoing maintenance due to the fact that it duplicates functionality present in our build system and would need to be updated accordingly when changes are made there. Are you interested in maintaining this functionality in the long-term? If so, I think we are pretty close to moving forward.

cluster/local/provider.Tiltfile
khos2ow: Yes, they need to be changed to `https://raw.githubusercontent.com/crossplane/crossplane/master/...` right before merge. The reason for current one is the make the functionality _workable_ for testing, demo, or example usage in [provider-digitalocean](https://github.com/crossplane-contrib/provider-digitalocean/blob/main/Tiltfile#L17).

cluster/local/provider.Tiltfile
khos2ow: Specifically in this file (i.e. `cluster/local/provider.Tiltfile`), no. Because this `dockerfile_contents` is intended for building and live-reloading. To clarify a bit more, this file is being used in two different contexts:

- inside `crossplane/crossplane` when someone wants to work on Crossplane *and* a provider at the same time
- inside `org/provider-foo`, when someone is only working on a provider with the assumption of crossplane is running

In the first case there is no visibility of `Dockerfile` for provider at all to do anything with it.

But in the second case you could argue the `Dockerfile` for provider already exists in the context, but the main (missing) enhancement point would be `tilt_start.sh` and `tilt_restart.sh` to live reload the process inside the running pod without bouncing the pod itself.

cluster/local/provider.Tiltfile
khos2ow: That would make an assumption that `make build` exists and it builds with `CGO_ENABLED=0 GOOS=linux GOARCH=amd64` (to be run inside an `alpine` Docker image). Assuming if `Makefile` and `build` target exist is not very far fetched (to be honest), but not so much for the `GOOS` and `GOARCH`.

Also `make build` (as it stands now) runs some slightly irrelevant checks (for Tilt context) such as module dependencies verification, golangci-lint, to name a few. I mean they are completely important but not necessarily in Tilt when you want to live reload as fast as possible (under a second or so) when you change a Go file.

cluster/local/provider.Tiltfile
khos2ow: Good to know, I haven't tested this recently but one thing gives me pause (just by looking at the code) is that you say this runs in `initContainer` but the `provider.Tiltfile` code (as it stands) doesn't care about it and even when the Tilt is running and if it detects a change in `api/` folder it will regenerate and reapply CRDs anyways (as opposed to one-off operations of `initContainer`).

But for sure this can be revisited and if we can remove it, all the better!

khos2ow:
@hasheddan yes for sure it needs maintenance, but realistically the content of `Tiltfile` and `provider.Tiltefile` should remain relatively stable, unless new features would be needed. But all in all I'd be happy to help on the maintenance anyways.

negz:
How are we feeling about this PR folks? It's been open for over a year (our bad!). It seems to me that the main concern is that it duplicates existing functionality from our build submodule, right? For example I see that the `Tiltfile` appears to invoke `go generate` and builds Crossplane without using `make build`. I definitely share these concerns, though I'm also curious about Tilt and don't want to have a chilling effect on changes that might make development easier.

My hunch is that we probably should not merge this, and instead lump it into the [greater conversations](https://github.com/crossplane/crossplane/issues?q=is%3Aopen+is%3Aissue+label%3Abuild) we have going on about improving build - e.g. starting to replace some of the `Make` we use elsewhere with tools like Tilt and Hermit rather than maintaining both.

khos2ow:
> For example I see that the Tiltfile appears to invoke go generate and builds Crossplane without using make build.

@negz the main reason I didn't reuse make targets (specifically `make build` in this case) was the different nature of Makefile targets and Tilt. Make targets, generally, are intended to handle dependencies and some sort off abstraction (e.g. if you want to _build_, we have to first _generate_ stuff). But one of the main benefits of Tilt is the fast turnaround (e.g. when you make a change and save a file, it will rebuild the _things_, repopulate/inject the docker image/layer, restart the pod, etc), somewhere in the neighbourhood of 1 seconds.

TBH, I don't exactly remember why I didn't use a make target, namely for the example you pointed out, but my objective was to reuse Makefile as much as possible otherwise copy the _thing_ over directly from corresponding target to `Tiltfile` for faster turnaround of execution within tilt lifecycle.

negz:
This has been sitting open for a long time now and we haven't moved on it so I'm going to close it out. We appreciate the idea nonetheless!

Issue #1923, opened 2020-11-03
Should package revisions watch the objects they control


negz:
### What problem are you facing?

I believe the `ProviderRevision` and `ConfigurationRevision` controllers currently do not watch the objects they control (i.e. Deployment, ServiceAccount, CRD, XRD, and Compositions), and thus do not react immediately to reset changes made to those objects. The controllers instead speculatively requeue every 60 seconds.

### How could Crossplane help solve your problem?

The revision controllers could watch the objects they are a controller reference of such that a reconcile would be queued for the revision when any of those objects changed. I believe this would allow us to avoid having the revision controllers speculatively reconcile every 60 seconds.

negz:
That said, we might not want to duplicate the knowledge of what types might be packaged (that knowledge is currently cleanly contained in `objScheme`. If there was some way to extract instances of all types known to that scheme we could presumably do something like:

```go
	b := ctrl.NewControllerManagedBy(mgr).
		Named(name).
		For(&v1alpha1.ProviderRevision{})

	for _, o := range GetTypes(objScheme) {
		b = b.Owns(o, &handler.EnqueueRequestForOwner{OwnerType: &v1alpha1.ProviderRevision{}, IsController: true})
	}

	return b.Complete(r)
```

hasheddan:
> That said, we might not want to duplicate the knowledge of what types might be packaged

I actually don't think this is an issue as this information will be passed at controller setup time, and we are already doing package-specific things there, such as setting hooks to run.

PR #1928, opened 2020-11-05
Do not enqueue on changes to roles or CRDs in ProviderRevision RBAC manager

hasheddan:

In order to avoid constant reconciliation of RBAC roles, we rely on a
ProviderRevision reporting unhealthy to enqueue a request for its roles
and role bindings to be reconciled.

I have:

[will update with manual tests]

### DIFF
--- a/pkg/controller/rbac/provider/roles/reconciler.go
+++ b/pkg/controller/rbac/provider/roles/reconciler.go
@@ -26,10 +26,8 @@ import (
 	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	kcontroller "sigs.k8s.io/controller-runtime/pkg/controller"
-	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
-	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	"github.com/crossplane/crossplane-runtime/pkg/event"
 	"github.com/crossplane/crossplane-runtime/pkg/logging"
@@ -78,7 +76,6 @@ func Setup(mgr ctrl.Manager, log logging.Logger) error {
 		Named(name).
 		For(&v1alpha1.ProviderRevision{}).
 		Owns(&rbacv1.ClusterRole{}).
-		Watches(&source.Kind{Type: &extv1.CustomResourceDefinition{}}, &handler.EnqueueRequestForOwner{OwnerType: &v1alpha1.ProviderRevision{}}).
 		WithOptions(kcontroller.Options{MaxConcurrentReconciles: maxConcurrency}).
 		Complete(NewReconciler(mgr,
 			WithLogger(log.WithValues("controller", name)),


pkg/controller/rbac/namespace/reconciler.go
```diff
@@ -118,10 +118,11 @@ func WithRoleRenderer(rr RoleRenderer) ReconcilerOption {
 // NewReconciler returns a Reconciler of Namespaces.
 func NewReconciler(mgr manager.Manager, opts ...ReconcilerOption) *Reconciler {
 	r := &Reconciler{
-		// TODO(negz): Is Updating appropriate here? Probably.
+		// NOTE(hasheddan): Patching applicator is used to avoid conflict with
+		// Kubernetes RBAC aggregation controller.
 		client: resource.ClientApplicator{
 			Client:     mgr.GetClient(),
-			Applicator: resource.NewAPIUpdatingApplicator(mgr.GetClient()),
+			Applicator: resource.NewAPIPatchingApplicator(mgr.GetClient()),
```
hasheddan, line 125: I think we may actually be running into issues because we are both "owning" and "watching" cluster roles. I am going to remove the "owning" and test again.

hasheddan: Okay so I have improved the situation by having the `ProviderRevision` RBAC controllers not own `ClusterRole` and `ClusterRoleBinding` respectively. This addresses the issue of continuous updates as it appears the namespace and providerrevision RBAC controllers were triggering each other repetitively. However, we are still "boiling the ocean" every 60 seconds due to https://github.com/crossplane/crossplane/issues/1923. I'll try to address that now.

pkg/controller/rbac/definition/reconciler.go
```diff
@@ -117,10 +117,11 @@ func WithClusterRoleRenderer(rr ClusterRoleRenderer) ReconcilerOption {
 // NewReconciler returns a Reconciler of CompositeResourceDefinitions.
 func NewReconciler(mgr manager.Manager, opts ...ReconcilerOption) *Reconciler {
 	r := &Reconciler{
-		// TODO(negz): Is Updating appropriate here? Probably.
+		// NOTE(hasheddan): Patching applicator is used to avoid conflict with
+		// Kubernetes RBAC aggregation controller.
 		client: resource.ClientApplicator{
 			Client:     mgr.GetClient(),
-			Applicator: resource.NewAPIUpdatingApplicator(mgr.GetClient()),
+			Applicator: resource.NewAPIPatchingApplicator(mgr.GetClient()),
```
negz, line 124: If I recall correctly the patching applicator here given that it's additive; i.e. we'll never remove any field that was previously set on a Role or ClusterRole. That said, I believe the rules array will be set to our desired state (i.e. we'd delete any rules that we no longer wanted) due to the fact that we're using a JSON merge patch that always completely rewrites arrays. I'm a little wary of depending on this behaviour for security purposes; if we do need to stick with the patching applicator I think we should at least add a comment explaining what it is doing and why it's safe.

hasheddan: I do have a comment above, but I actually think we can stick with updating. The second commit is what fixed the issue from what I have been able to tell

hasheddan:
Updated PR title and body to reflect the current solution.

pkg/controller/rbac/provider/binding/reconciler.go
```diff
@@ -67,8 +65,6 @@ func Setup(mgr ctrl.Manager, log logging.Logger) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		Named(name).
 		For(&v1alpha1.ProviderRevision{}).
-		Owns(&rbacv1.ClusterRoleBinding{}).
-		Watches(&source.Kind{Type: &corev1.ServiceAccount{}}, &handler.EnqueueRequestForOwner{OwnerType: &v1alpha1.ProviderRevision{}}).
```
negz, line 71: This seems like it reverts https://github.com/crossplane/crossplane/pull/1919, which was put in place to address a bug that I observed. Do we have a reason to believe we won't see that bug again?

hasheddan: No I think that could occur again. We should be able to add back the requeue safely for SA as it is not watched by the Namespace controller. Does that sound good to you?

hasheddan:
Closing in favor of #1931 and #1932

Issue #1929, opened 2020-11-05
RBAC Manager constantly applies roles


negz:
### What happened?

We've observed that the RBAC Manager constantly applies roles, causing a lot of thrash and memory usage in the API server to the point that it uses >2GB of memory.

### How can we reproduce it?

More details TBD.

### What environment did it happen in?
Crossplane version:

negz:
I believe this is now fixed since https://github.com/crossplane/crossplane/pull/1932 was merged (in addition to the PRs already linked).

PR #1931, opened 2020-11-06
Apply RBAC roles less frequently

negz:

Partial fix for https://github.com/crossplane/crossplane/issues/1929

This PR ensures that we don't update RBAC Roles and ClusterRoles unless our update would meaningfully change the role. Based on my testing this should be a partial fix for the above issue. Each time a CRD or XRD is updated by the ProviderRevision or ConfigurationRevision controllers the RBAC controller has a reconcile queued. That can be up to 40 reconciles for a provider like AWS that owns 40 CRDs, though the reconcile workqueue appears to do some deduping for us here. With this PR in place updates to CRDs and XRDs won't result in a fanout of updates to RBAC ClusterRoles and Roles unless something meaningfully changed (i.e. unless we'd change rules, annotations, or labels).

@hasheddan is working on a corresponding PR that will prevent the package revision controllers from updating the objects owned by provider and configuration revisions unless something meaningful about them has changed. I expect the two PRs combined should eliminate the API server load issues being caused by RBAC manager churn.

I have:

I've tested this by installing provider-aws:master on a Kubernetes 1.19 kind cluster running this PR. I see that every 60 seconds all CRDs change. This triggers a lot of reconciles (around 12 reconciles every 60 seconds, when the ProviderRevision controller updates its 40 AWS CRDs), but those reconciles don't interact with the API server. They generate new RBAC roles, then read the existing RBAC roles from cache and determine that nothing has changed.

### DIFF
--- a/pkg/controller/rbac/namespace/reconciler.go
+++ b/pkg/controller/rbac/namespace/reconciler.go
@@ -20,9 +20,11 @@ import (
 	"context"
 	"time"
 
+	"github.com/google/go-cmp/cmp"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	kcontroller "sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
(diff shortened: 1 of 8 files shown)


pkg/controller/rbac/provider/roles/reconciler.go
```diff
@@ -215,3 +224,12 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error)
 	// There's no need to requeue explicitly - we're watching all PRs.
 	return reconcile.Result{Requeue: false}, nil
 }
+
+// ClusterRolesDiffer returns true if the supplied objects are different
+// ClusterRoles. We consider ClusterRoles to be different their labels and rules
+// do not match.
```
hasheddan, line 230:
```suggestion
// ClusterRoles. We consider ClusterRoles to be different if their labels and rules
// do not match.
```

pkg/controller/rbac/namespace/reconciler.go
```diff
@@ -189,17 +191,31 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error)
...
-		log.Debug("Applied RBAC Role", "role-name", rl.GetName())
+		log.Debug("Applied RBAC Role", "role-name")
 	}
 
 	r.record.Event(ns, event.Normal(reasonApplyRoles, "Applied RBAC Roles"))
 	return reconcile.Result{Requeue: false}, nil
 }
+
+// RolesDiffer returns true if the supplied objects are different Roles. We
+// consider Roles to be different their annotations and rules do not match.
```
hasheddan, line 216:
```suggestion
// consider Roles to be different if their annotations and rules do not match.
```

hasheddan (approved):
Two minor nitpicks but LGTM!

negz:
I believe this is now fixed since https://github.com/crossplane/crossplane/pull/1932 was merged (in addition to the PRs already linked).

PR #1932, opened 2020-11-06
Only requeue package revision controller on changes to installed objects

### LINKED ISSUE #1923: Should package revisions watch the objects they control

hasheddan:

Updates the package revision controller to only requeue when an installed object changes or when the `____Revision` itself is modified.

Fixes #1923 

I have:

manual testing steps to be added

### DIFF
--- a/pkg/controller/pkg/revision/establisher.go
+++ b/pkg/controller/pkg/revision/establisher.go
@@ -19,6 +19,8 @@ package revision
 import (
 	"context"
 
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/pkg/errors"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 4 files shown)


pkg/controller/pkg/revision/reconciler.go
```diff
@@ -217,6 +224,10 @@ func SetupProviderRevision(mgr ctrl.Manager, l logging.Logger, cache xpkg.Cache,
 	return ctrl.NewControllerManagedBy(mgr).
 		Named(name).
 		For(&v1alpha1.ProviderRevision{}).
+		Watches(&source.Kind{Type: &extv1.CustomResourceDefinition{}}, &handler.EnqueueRequestForOwner{OwnerType: &v1alpha1.ProviderRevision{}}).
+		Watches(&source.Kind{Type: &extv1beta1.CustomResourceDefinition{}}, &handler.EnqueueRequestForOwner{OwnerType: &v1alpha1.ProviderRevision{}}).
```
negz, line 228: Do we need to watch both of these versions? I'm fairly sure that when one changes we'll see an update for the other.

pkg/controller/pkg/revision/establisher.go
```diff
@@ -131,30 +133,42 @@ func (e *APIEstablisher) Establish(ctx context.Context, objs []runtime.Object, p
...
-func (e *APIEstablisher) create(ctx context.Context, obj resource.Object, parent resource.Object, control bool, opts ...client.CreateOption) error {
+func (e *APIEstablisher) create(ctx context.Context, obj runtime.Object, parent resource.Object, control bool, opts ...client.CreateOption) error {
+	o := obj.DeepCopyObject().(resource.Object)
 	ref := meta.AsController(meta.TypedReferenceTo(parent, parent.GetObjectKind().GroupVersionKind()))
 	if !control {
 		ref = meta.AsOwner(meta.TypedReferenceTo(parent, parent.GetObjectKind().GroupVersionKind()))
 	}
 	// Overwrite any owner references on the desired object.
-	obj.SetOwnerReferences([]metav1.OwnerReference{ref})
+	o.SetOwnerReferences([]metav1.OwnerReference{ref})
```
negz, line 143: Won't this be a no-op? It looks like we create a deep copy of obj, then change its owner references, then apply obj. Do we need to make the deepcopy, or can we just assert obj to a `resource.Object` (or `metav1.Object`)?

pkg/controller/pkg/revision/establisher.go
```diff
@@ -131,30 +133,42 @@ func (e *APIEstablisher) Establish(ctx context.Context, objs []runtime.Object, p
...
 	return e.client.Create(ctx, obj, opts...)
 }
 
-func (e *APIEstablisher) update(ctx context.Context, current, desired resource.Object, parent resource.Object, control bool, opts ...client.UpdateOption) error {
+func (e *APIEstablisher) update(ctx context.Context, current, desired runtime.Object, parent resource.Object, control bool, opts ...client.UpdateOption) error {
+	c := current.DeepCopyObject().(resource.Object)
+	d := desired.DeepCopyObject().(resource.Object)
```
negz, line 149: Ditto the above comment; do we need to deepcopy here?

pkg/controller/pkg/revision/establisher.go
```diff
@@ -131,30 +133,42 @@ func (e *APIEstablisher) Establish(ctx context.Context, objs []runtime.Object, p
...
 	if !control {
-		meta.AddOwnerReference(current, meta.AsOwner(meta.TypedReferenceTo(parent, parent.GetObjectKind().GroupVersionKind())))
-		return e.client.Update(ctx, current, opts...)
+		ownerRef := meta.AsOwner(meta.TypedReferenceTo(parent, parent.GetObjectKind().GroupVersionKind()))
+		for _, ref := range c.GetOwnerReferences() {
+			if ref == ownerRef {
```
negz, line 153: Will struct equality work as expected when the struct has pointer fields? (e.g. `ref.Controller`)?

negz: Also, this logic looks pretty similar to what `meta.AddOwnerReference` does. Presumably we could also copy the original, then call `meta.AddOwnerReference`, then use `cmp` to determine whether they changed, but I suspect your current implementation would be more efficient.

pkg/controller/pkg/revision/reconciler.go
hasheddan: You are correct :slightly_smiling_face: 

> // OwnerType is the type of the Owner object to look for in OwnerReferences.  Only Group and Kind are compared.

pkg/controller/pkg/revision/establisher.go
hasheddan: > Won't this be a no-op?

This is to guard against a case where a package attempts to install a resource with owner references already present. The reason why I have moved to deep copying in both the create and update methods is that we call them multiple times (once for dry run, once for actual action), and when we modify the owner references on the first pass, they may be inaccurate on the second (e.g. I noticed pointer values being modified, which can be unexpected if you have already stored the object into a different struct, such as `allObjs` here).

pkg/controller/pkg/revision/establisher.go
hasheddan: See comment above. The added tests also demonstrate some of this behavior.

pkg/controller/pkg/revision/establisher.go
hasheddan: > Will struct equality work as expected when the struct has pointer fields?

This comparison is only used on owners, never controllers, so we don't actually end up comparing any pointers that are not `nil`.

pkg/controller/pkg/revision/reconciler.go
negz: I see! So should we remove `extv1beta1` here?

pkg/controller/pkg/revision/establisher.go
negz: I'm still not quite following here; are you intending to be applying the copy (`o`) below, or the original (`obj`)?

hasheddan: Ah yes sorry, I was looking at the comment view here, that omits that line.. will fix

hasheddan:
#1923 is still relevant, but this PR is stale.

Issue #1930, opened 2020-11-05
Bump all core Crossplane APIs to v1beta1


negz:
### What problem are you facing?

We feel good about the stability of our APIs, and would like to bump all core APIs (i.e. APIs defined in this repo) from v1alpha1 to v1beta1.

PR #1937, opened 2020-11-06
Bump core APIs to v1beta1

negz:

Fixes https://github.com/crossplane/crossplane/issues/1930

This PR bumps all core APIs (except those in the meta.pkg.crossplane.io group) to v1beta1. It maintains support for the existing v1alpha1 APIs without conversion, because the two versions are identical in their schemas. All controllers have been ported to use the new v1beta1 versions, which are marked as the storage version. This goes against the [backward compatibility guidelines](https://github.com/kubernetes/community/blob/46f114/contributors/devel/sig-architecture/api_changes.md#backward-compatibility-gotchas), which state:

> Do not add a new API version of an existing resource and make it the preferred version in the same release, and do not make it the storage version. The latter is necessary so that a rollback of the apiserver doesn't render resources in etcd undecodable after rollback.

I suspect this guidance pertains mostly to cases in which we'd need to roll back to a storage version with a different schema. In this case I'm hoping the change is boilerplate enough that we don't need to worry about this.

I have:

I have:

* Installed a build of this PR into a `kind` cluster, and used the new `crank` for all interaction.
* Installed all the core providers.
* Installed the master GCP getting-started configuration (which install v1alpha1 XRDs and Compositions).
* Applied the GCP claim from `docs/snippets/compose`.
* Inspected the RBAC manager and core Crossplane logs to make sure nothing is awry.

I'm fairly satisfied based on this cursory testing that I haven't broken anything significant with this huge (but largely boilerplate) change.


hasheddan (approved):
Thanks @negz!

PR #1950, opened 2020-11-09
Do not execute lint job where it's not relevant

khos2ow:

This PR removes lint jobs (go and helm) from normal build targets and CI jobs. Notable changes:

- `make reviewable`: remained as previous
- `make check-diff`: does not execute lint (as lint doesn't impose any dirtiness to the repo)
- `make build`: does not execute `make build.check` anymore (this is really debatable!)
- `make helm.check`: new target added with corresponding CI step in `lint` job

This PR depends on https://github.com/upbound/build/pull/126 and should be reviewed as a pair.

**NOTE:** This PR should not get merged as is. I added two demo commits in this PR which should be reverted after review is done. Also to make the demo work I updated `build/` submodule pointing to my own fork, which should be reverted to HEAD of `upboud/build:master` when the corresponding PR over there gets merged.

I have:

n/a


cluster/local/integration_tests.sh
```diff
@@ -43,7 +43,7 @@ BUILD_IMAGE="${BUILD_REGISTRY}/${PROJECT_NAME}-${HOSTARCH}"
 
 helm_tag="$(cat ${projectdir}/_output/version)"
 CROSSPLANE_IMAGE="${DOCKER_REGISTRY}/${PROJECT_NAME}:${helm_tag}"
-K8S_CLUSTER="${K8S_CLUSTER:-${BUILD_REGISTRY}-INTTESTS}"
+K8S_CLUSTER="${K8S_CLUSTER:-${BUILD_REGISTRY}-inttests}"
```
khos2ow, line 46: This change was needed due to new constraint in kind that uppercase characters are not allowed anymore.

khos2ow:
@negz @hasheddan this should be good for review now.

go.mod
```diff
@@ -4,6 +4,7 @@ go 1.13
 
 require (
 	github.com/Masterminds/semver v1.5.0
+	github.com/Masterminds/semver/v3 v3.1.0 // indirect
```
hasheddan, line 7: This is surprising to see in this set of changes 🤔

cmd/crossplane/main.go
```diff
@@ -19,11 +19,9 @@ package main
 import (
 	"os"
 	"path/filepath"
-
```
hasheddan, line 22: Looks like these newlines were inadvertently removed and are causing errors.

go.mod
khos2ow: This is a demo commit to show the error in `check-diff` doesn't leak somewhere else in the CI. I will revert them after review is done.

cmd/crossplane/main.go
khos2ow: This is a demo commit to show the error in `lint` doesn't leak somewhere else in the CI. I will revert them after review is done.

.github/workflows/ci.yml
```diff
@@ -68,11 +67,14 @@ jobs:
...
 
+      - name: Lint Helm Chart
+        run: make helm.lint
+
       # This action uses its own setup-go, which always seems to use the latest
       # stable version of Go. We could run 'make lint' to ensure our desired Go
       # version, but we prefer this action because it leaves 'annotations' (i.e.
       # it comments on PRs to point out linter violations).
-      - name: Lint
+      - name: Lint Go
```
negz, line 77: I'm a little wary of the fact that this would mean we're no longer running the build submodule variant of `golangci-lint` anywhere during our CI runs; i.e. `make reviewable` could subtly drift from this action (due to different GoLang CI versions, etc).

Do you have an idea of how much this speeds up builds? If it's only a slight gain, perhaps the duplication is preferable.

Makefile
```diff
@@ -109,11 +109,10 @@ cobertura:
 		$(GOCOVER_COBERTURA) > $(GO_TEST_OUTPUT)/cobertura-coverage.xml
 
 # Ensure a PR is ready for review.
-reviewable: generate lint
-	@go mod tidy
+reviewable: generate lint go.vendor.tidy
 
 # Ensure branch is clean.
-check-diff: reviewable
+check-diff: generate go.vendor.tidy
```
negz, line 115: One thing to note here is that theoretically at least running `golangci-lint` can result in changes. It has a `--fix` flag (that we do not enable by default) that allows it to automatically fix a small set of the linter rules it knows.

.github/workflows/ci.yml
khos2ow: I haven't benchmarked it that much, but to be frank it will be a matter of seconds (maybe ~10-15 sec?), but the speed was not the main factor behind the change rather than preventing seeing false negative in the CI.

A recent example of that is #1925, `e2e-tests` is failing but because of lint issues. This, just by looking at the status of the PR, gives an impression that the change actually broke something fundamental, and might make author or reviewer waste some time trying to figure out what has happened.

But this is completely in relation of your point https://github.com/upbound/build/pull/126#discussion_r520788815 , the remediation of this was to change upstream `build/` submodule and the concerns around it.

Makefile
khos2ow: That's a great point. I had considered it and exactly because `--fix` is not enabled in the config or CLI, I thought this could be fine to drop from `check-diff`.

On the other hand, this can be reverted back to ensure we always run `golang-ci` variant of `build/` submodule at least somewhere on the CI (with a very minimal time penalty).

hasheddan:
@khos2ow I believe this will be accomplished in https://github.com/crossplane/crossplane/pull/2334, which incorporates some changes in the build submodule. Are you okay with closing this one out?

khos2ow:
@hasheddan technically, no. The idea behind this PR is to decouple running `lint` from `build`. So if there's a lint issue for example `CI / unit-tests` won't fail. Note that `reviewable` target was not changed as part of this. So the workflow could be:

- on dev machine: you can run `make reviewable` to ensure the branch will be _reviewable_
- on CI: if there's a lint issue only `CI / lint` fails and there won't be any false positive failure on other jobs (e.g. `unit-tests` or `e2e-tests` which technically shouldn't be concerned about lint issues; at CI job level not the whole PR)

hasheddan:
> That said, letting lint target to run as part of all ci jobs is really not very expensive and overall this PR can be discarded, but personally I think it still has some benefit for the separation.

I agree that it is valuable to reduce resource usage. However, I am not sure that just removing the lint step from parallel jobs is the most critical thing in this effort. For instance, It would likely be a much larger gain from a resource usage perspective to not build the binaries and images in every job, and instead introduce [job dependencies](https://docs.github.com/en/actions/learn-github-actions/managing-complex-workflows#creating-dependent-jobs). I believe the initial reason for not doing this was a mix of simplicity and speed in the happy path (i.e. if all jobs are going to pass it could theoretically take more time to use dependent jobs). I would certainly be in favor of reducing much of this duplication :+1:

negz:
I appreciate you working on this @khos2ow, but I'm going to close it. In practice I don't think the duplicative linting has really been an issue either false positives or CI time wise.

PR #1951, opened 2020-11-10
Support overriding image in ControllerConfig

### LINKED ISSUE #1938: Support Overriding the Image in ControllerConfig

We have some clusters which aren't allowed to access https://hub.docker.com/ directly. They can only pull images via a registry mirror (e.g. `remote-docker.artifactory.example.com/<org>/<image>:<tag>`). The mirror acts as a pull-through cache and serves the exact same images as Docker hub.
Creating a `Provider` we can specify the correct provider image:
```yaml
kind: Provider
metadata:
  name: provider-helm
spec:
  package: remote-docker.artifactory.example.com/crossplane/provider-helm:v0.3.6
```
The provider meta config in this image references the Docker hub again though and there's currently no way to override this. The only workaround currently would be to copy the provider image and change the meta config accordingly. This imposes unnecessary image copies and maintenance though.

Using the new controller config (#974) it could be possible to override the controller image for a provider:
```yaml
kind: ControllerConfig
metadata:
  name: provider-helm
spec:
  image: remote-docker.artifactory.example.com/crossplane/provider-helm-controller
```

srueg:

Fixes #1938 

I have:

I tested it locally: `make run` against `kind` cluster and installed `provider-helm` with a `ControllerConfig` to override the image:
```yaml
kind: ControllerConfig
metadata:
  name: provider-helm
spec:
  args:
  - --debug
  image: docker.io/vshn/provider-helm:v0.3.6
```

### DIFF
--- a/apis/pkg/v1alpha1/config.go
+++ b/apis/pkg/v1alpha1/config.go
@@ -32,6 +32,12 @@ type ControllerConfigSpec struct {
 	// used to enable leader election process.
 	// +optional
 	Replicas *int32 `json:"replicas,omitempty"`
+	// Docker image name.
+	// More info: https://kubernetes.io/docs/concepts/containers/images
+	// This field is optional to allow higher level config management to default or override
+	// container images in workload controllers like Deployments and StatefulSets.
+	// +optional
+	Image *string `json:"image,omitempty"`
 	// NodeSelector is a selector which must be true for the pod to fit on a node.
 	// Selector which must match a node's labels for the pod to be scheduled on that node.
 	// More info: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/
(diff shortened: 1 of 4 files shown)


apis/pkg/v1alpha1/config.go
```diff
@@ -32,6 +32,12 @@ type ControllerConfigSpec struct {
 	// used to enable leader election process.
 	// +optional
 	Replicas *int32 `json:"replicas,omitempty"`
+	// Docker image name.
+	// More info: https://kubernetes.io/docs/concepts/containers/images
+	// This field is optional to allow higher level config management to default or override
+	// container images in workload controllers like Deployments and StatefulSets.
+	// +optional
+	Image string `json:"image,omitempty"`
```
hasheddan, line 40: nitpick: let's make this a `*string` so we evaluate if it is present by check on `nil` rather than length of string 👍

hasheddan (approved):
LGTM! Thanks @srueg, would you mind rebasing on `master` and we can merge this in?

PR #1968, opened 2020-11-18
Block owner deletion for package revisions

hasheddan:

Makes the inability to delete a package revision block deletion of its
parent package. There should never be a case in which a revision exists
but has no parent.

I have:

Tested by running with a package revision controller that never removes its finalizer.

### DIFF
--- a/pkg/controller/pkg/manager/reconciler.go
+++ b/pkg/controller/pkg/manager/reconciler.go
@@ -25,6 +25,7 @@ import (
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/utils/pointer"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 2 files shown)


pkg/controller/pkg/manager/reconciler.go
```diff
@@ -47,6 +47,10 @@ const (
 	pullWait      = 1 * time.Minute
 )
 
+var (
+	blockOwnerDeletion = true
+)
```
negz, line 52: Nit: Could this be a function local var? https://godoc.org/k8s.io/utils/pointer#BoolPtr could work too.

negz (approved):

pkg/controller/pkg/manager/reconciler.go
hasheddan: Good call, done :+1: Was actually thinking about adding this to the helpers for owner refs we have in crossplane-runtime, but introducing a function that only sets a single field seems like over-optimizing

Issue #1649, opened 2020-07-13
Updating a claim does not update its composite resource


muvaf:
### What problem are you facing?

Updating a field on a composite resource claim should propagate to the corresponding composite resource field, but does not.

### How could Crossplane help solve your problem?

Automatically propagate fields of requirement to all composed resources.

PR #1970, opened 2020-11-18
Propagate XRC updates to their corresponding XR

negz:

Fixes https://github.com/crossplane/crossplane/issues/1649

This PR also includes a handful of updates to prevent spurious reconciles that were discovered as a byproduct of reconciling composite resources more frequently.

I have:

I've followed the AWS (with VPC) getting started with composition guide, and ensured that when I update the storage GB of my `PostgreSQLInstance` claim the changes are propagated to the corresponding `CompositePostgreSQLInstance`) XR (and thus onto the underlying `RDSInstance`).

### DIFF
--- a/pkg/controller/apiextensions/claim/api.go
+++ b/pkg/controller/apiextensions/claim/api.go
@@ -34,65 +34,17 @@ import (
 
 // Error strings.
 const (
-	errCreateComposite      = "cannot create composite resource"
-	errUpdateClaim          = "cannot update composite resource claim"
-	errUpdateComposite      = "cannot update composite resource"
-	errDeleteComposite      = "cannot delete composite resource"
-	errBindConflict         = "cannot bind composite resource that references a different claim"
-	errGetSecret            = "cannot get composite resource's connection secret"
-	errSecretConflict       = "cannot establish control of existing connection secret"
-	errCreateOrUpdateSecret = "cannot create or update connection secret"
+	errUpdateClaim           = "cannot update composite resource claim"
+	errUpdateComposite       = "cannot update composite resource"
+	errBindClaimConflict     = "cannot bind claim that references a different composite resource"
+	errBindCompositeConflict = "cannot bind composite resource that references a different claim"
+	errGetSecret             = "cannot get composite resource's connection secret"
+	errSecretConflict        = "cannot establish control of existing connection secret"
+	errCreateOrUpdateSecret  = "cannot create or update connection secret"
 )
 
-// An APICompositeCreator creates resources by submitting them to a Kubernetes
-// API server.
-type APICompositeCreator struct {
-	client client.Client
-	typer  runtime.ObjectTyper
-}
-
-// NewAPICompositeCreator returns a new APICompositeCreator.
-func NewAPICompositeCreator(c client.Client, t runtime.ObjectTyper) *APICompositeCreator {
-	return &APICompositeCreator{client: c, typer: t}
-}
-
-// TODO(negz): We should render and patch a composite resource on each
-// reconcile, rather than just creating it once.
-
-// Create the supplied composite using the supplied claim.
-func (a *APICompositeCreator) Create(ctx context.Context, cm resource.CompositeClaim, cp resource.Composite) error {
-
-	cp.SetClaimReference(meta.ReferenceTo(cm, resource.MustGetKind(cm, a.typer)))
-	if err := a.client.Create(ctx, cp); err != nil {
-		return errors.Wrap(err, errCreateComposite)
-	}
-	// Since we use GenerateName feature of ObjectMeta, final name of the
-	// resource is calculated during the creation of the resource. So, we
-	// can generate a complete reference only after the creation.
-	cpr := meta.ReferenceTo(cp, resource.MustGetKind(cp, a.typer))
-	cm.SetResourceReference(cpr)
-
-	return errors.Wrap(a.client.Update(ctx, cm), errUpdateClaim)
-}
-
-// An APICompositeDeleter deletes composite resources from the API server.
-type APICompositeDeleter struct {
-	client client.Client
-}
-
-// NewAPICompositeDeleter returns a new APICompositeDeleter.
-func NewAPICompositeDeleter(c client.Client) *APICompositeDeleter {
-	return &APICompositeDeleter{client: c}
-}
-
-// Delete the supplied composite resource from the API server.
-func (a *APICompositeDeleter) Delete(ctx context.Context, _ resource.CompositeClaim, cp resource.Composite) error {
-	return errors.Wrap(resource.IgnoreNotFound(a.client.Delete(ctx, cp)), errDeleteComposite)
-}
-
 // An APIBinder binds claims to composites by updating them in a Kubernetes API
-// server. Note that APIBinder does not support objects that do not use the
-// status subresource; such objects should use APIBinder.
+// server.
 type APIBinder struct {
 	client client.Client
 	typer  runtime.ObjectTyper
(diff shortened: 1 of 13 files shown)


pkg/controller/apiextensions/claim/api.go
```diff
@@ -105,25 +57,32 @@ func NewAPIBinder(c client.Client, t runtime.ObjectTyper) *APIBinder {
...
 // Bind the supplied claim to the supplied composite.
 func (a *APIBinder) Bind(ctx context.Context, cm resource.CompositeClaim, cp resource.Composite) error {
-	existing := cp.GetClaimReference()
-	proposed := meta.ReferenceTo(cm, resource.MustGetKind(cm, a.typer))
-
-	if existing != nil && (existing.Namespace != proposed.Namespace || existing.Name != proposed.Name) {
-		return errors.New(errBindConflict)
+	existing := cm.GetResourceReference()
+	proposed := meta.ReferenceTo(cp, resource.MustGetKind(cp, a.typer))
+	if existing != nil && !cmp.Equal(existing, proposed, cmpopts.IgnoreFields(corev1.ObjectReference{}, "UID")) {
```
hasheddan, line 62: This would be a nice helper to have in `crossplane-runtime` as a helper at some point :)

pkg/controller/apiextensions/composite/reconciler.go
```diff
@@ -430,17 +432,15 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error)
 		r.record.Event(cr, event.Normal(reasonPublish, "Successfully published connection details"))
 	}
 
-	// TODO(muvaf): Report which resources are not ready.
-
-	// TODO(muvaf): If a resource becomes Unavailable at some point, should we still
-	// report it as Creating?
-	wait := longWait
-	cr.SetConditions(runtimev1alpha1.Available())
```
hasheddan, line 438: Ah was this causing needless reconciles when not ready? I ran into this when implementing package manager as well

pkg/controller/apiextensions/claim/reconciler.go
```diff
@@ -294,44 +281,7 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error)
...
-			// handler can only enqueue reconciles for composite resources that
-			// have their claim reference set, so we can't expect to be queued
-			// implicitly when the composite resource we want to bind to appears.
-			log.Debug("Referenced composite resource not found", "requeue-after", time.Now().Add(aShortWait))
-			record.Event(cm, event.Warning(reasonBind, err))
-			cm.SetConditions(Waiting())
-			return reconcile.Result{RequeueAfter: aShortWait}, errors.Wrap(r.client.Status().Update(ctx, cm), errUpdateClaimStatus)
-		}
-		if err != nil {
+		if err := r.client.Get(ctx, meta.NamespacedNameOf(ref), cp); resource.IgnoreNotFound(err) != nil {
```
hasheddan, line 284: Are we still guarding against the case that we have a reference but it has been deleted? If I am seeing this correctly, it looks like we will always try to get if we have a reference, and if that reference does not exist then we will requeue.

pkg/controller/apiextensions/claim/api.go
negz: Possibly, though ideally we'll be rid of ObjectReference soon per https://github.com/crossplane/crossplane-runtime/issues/49. I don't think our own equivalent types include a UID.

pkg/controller/apiextensions/composite/reconciler.go
negz: Yeah, exactly. It was causing us to update the `lastTransitionTime` of the condition on every reconcile.

hasheddan: Nice catch 👍

pkg/controller/apiextensions/claim/reconciler.go
negz: Previously I believe we were guarding against the case in which we had a reference to an XR, but that XR didn't yet exist. i.e. The case in which an XRC wants to claim an existing XR, and the claim is created before the XR.

With this new logic if the XRC explicitly references an XR that does not yet (or anymore) exist, we'll simply continue and make a new composite resource. When Configure is called we'll overwrite the resource reference to point to the new composite resource.

I actually forget whether this change was intentional, or accidental. 😬 In retrospect I think we should probably maintain the previous behaviour; I'll update.

negz: I've pushed a commit that reverts to the old behaviour and also addresses the old TODO about not replicating the deletion logic. Thanks for catching this!

hasheddan (approved):
LGTM! Thanks @negz 🎉

pkg/controller/apiextensions/claim/configurator.go
```diff
@@ -35,10 +37,20 @@ const (
...
 )
 
+const errUnsupportedClaimSpec = "composite resource claim spec was not an object"
+
 // Configure the supplied composite resource. The composite resource name is
 // derived from the supplied claim, as {name}-{random-string}. The claim's
 // external name annotation, if any, is propagated to the composite resource.
 func Configure(_ context.Context, cm resource.CompositeClaim, cp resource.Composite) error {
+	cp.SetGenerateName(fmt.Sprintf("%s-", cm.GetName()))
+	meta.AddAnnotations(cp, cm.GetAnnotations())
```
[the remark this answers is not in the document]
muvaf: Well, do we really need a separate `Bind` function now that we do an `Apply` operation?

muvaf (commented):
Sorry, forgot to click `Submit review`

pkg/controller/apiextensions/claim/api.go
```diff
@@ -105,25 +57,32 @@ func NewAPIBinder(c client.Client, t runtime.ObjectTyper) *APIBinder {
...
+		return errors.New(errBindClaimConflict)
 	}
 
-	cp.SetClaimReference(proposed)
-	if err := a.client.Update(ctx, cp); err != nil {
-		return errors.Wrap(err, errUpdateComposite)
+	// Propagate the actual external name back from the composite to the claim.
+	meta.SetExternalName(cm, meta.GetExternalName(cp))
+
+	// We set the claim's resource reference first in order to reduce the chance
```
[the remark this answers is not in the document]
negz: Yeah, I had the same thought. Out of scope for this PR though, given this code was already there (I think I just moved it around a bit).

pkg/controller/apiextensions/claim/api.go
```diff
@@ -105,25 +57,32 @@ func NewAPIBinder(c client.Client, t runtime.ObjectTyper) *APIBinder {
...
+	cm.SetResourceReference(proposed)
+	if err := a.client.Update(ctx, cm); err != nil {
+		return errors.Wrap(err, errUpdateClaim)
 	}
 
-	if meta.GetExternalName(cp) == "" {
-		return nil
+	existing = cp.GetClaimReference()
+	proposed = meta.ReferenceTo(cm, resource.MustGetKind(cm, a.typer))
+	if existing != nil && !cmp.Equal(existing, proposed, cmpopts.IgnoreFields(corev1.ObjectReference{}, "UID")) {
```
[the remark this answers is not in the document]
negz: Agreed, but out of scope for this PR given we're already using ObjectRef here.

pkg/controller/apiextensions/claim/configurator.go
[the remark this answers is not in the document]
negz: Would you mind raising an issue for this since it was caught after merge? I'll address when I get time, probably early next week.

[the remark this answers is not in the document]
negz: > Well, do we really need a separate Bind function now that we do an Apply operation?

Apply only updates the composite, whereas Bind updates both the composite and claim (and also makes the changes to those types to indicate that they're bound - i.e. setting references). It's possible this all could be done behind one interface, but I don't know that we'd benefit much from doing so.

Issue #1909, opened 2020-10-30
Compositions should allow insertion, deletion, and reordering of resources


negz:
### What problem are you facing?

Each time Crossplane reconciles a Composite Resource (XR) it fetches the Composition the XR references and uses it to render a series of composed resources. The names of these composed resources are usually non-deterministic; Crossplane sets `metadata.generateName` to request that the API server name the composed resource at create time.  The the types and names of composed resources are recorded as an array of references (`spec.resourceRefs`) on the XR.

Currently Crossplane’s composition logic assumes that the `spec.resources` array of composed resource templates in a Composition is append-only. This is because after the composed resources are initially created Crossplane matches templates to existing composed resources by matching their indexes; the template at `spec.resources[0]` of the Composition is assumed to always correspond to `spec.resourceRefs[0]` in the XR. This means:

* It's not possible to reorder a Composition's `spec.resources` array.
* It's not possible to delete an element from the `spec.resources` array.
* It's not possible to add a new template to the `spec.resources` array, except at the end.

Note that above would all still be true under the CompositionRevision proposal in https://github.com/crossplane/crossplane/issues/1481, which simply snapshots each update to a Composition as a distinct revision to avoid XRs being forced to update whenever a Composition is updated.

### How could Crossplane help solve your problem?

I can see a few ways to address this:

1. We could declare that `Composition` resources and the `spec.compositionRef` of an XR are immutable. This would simplify our implementation at the expense of reduced flexibility for composition users. It would not be possible to change how an XR should be composed without deleting and recreating it.
1. We could formalise the current behaviour by declaring that the `spec.resources` array of a Composition is append only. Presumably an XR's `spec.compositionRef` would either need to be immutable, or required to only ever be updated to a Composition with an appended version of the current Composition's `spec.resources`.
1. We could ask Composition authors to supply a deterministic identifier for each entry in the Composition's `spec.resources`, and use that to correlate templates to each of the XR's `spec.resourceRefs`.

negz:
My preference is to go with option 3 - supplying a deterministic identifier at the Composition level. For example [the GCP Composition](https://github.com/crossplane/crossplane/blob/28224d92322a9caeaea12e6a2df97632358184ff/docs/snippets/package/gcp/composition.yaml) from our current documentation would become:

```yaml
apiVersion: apiextensions.crossplane.io/v1alpha1
kind: Composition
metadata:
  name: compositepostgresqlinstances.gcp.database.example.org
  labels:
[...]

PR #1972, opened 2020-11-18
Implement patchSets on Composition to allow deduplication of patches

### LINKED ISSUE #1960: Deduplicate repeated patches within a composition
When creating a `Composition` with a large number of base resources, it seems common to use the same patch definitions repeatedly for settings which apply to multiple resources, for example applying the following to pretty much every base within a `Composition`:

```yaml
patches:
  toFieldPath: metadata.labels
  toFieldPath: metadata.annotations
```

It would be really cool to allow patches to be defined once, and then referenced where necessary from each base.

@negz suggested something like this in Slack, which I like:

```yaml
kind: Composition
metadata:
  # ...
spec:
  # Sets of patches that can be referenced by id so you don't need to repeat yourself.
  patchSets:
  - id: metadata
    patches:
      toFieldPath: metadata.labels
      toFieldPath: metadata.labels
  - id: coolness
    patches:
    - fromFieldPath: spec.coolness
      toFieldPath: spec.amountOfCool
  resources:
  - base:
      # ...
    # Includes the patchSets above; use instead of or in addition to patches
    patchSets:
    - id: metadata
    - id: coolness
```

I'm not averse to submitting a PR to do this (on the assumption that I can work out how to implement, being relatively unfamiliar with the codebase 😄 )

benagricola:
Implements a new `patchSets` field on `CompositionSpec` for users to define named lists of `Patch`es at the composite level. 

`Patch`es are given a new `Type` field that can currently be either `FromCompositeFieldPath` (default if not specified - existing patch definitions on base resources continue to work)  or `PatchSet`. 

`PatchSet` patches have a new `patchSetName` field which, when applied at the resource level, will substitute the patches from the relevant `patchSet` on the composition. `PatchSet` patches cannot be defined at the `Composition` level.

Fixes #1960 

I have:

Tests have been added for the `InlinePatchSets()` method which is called by the `Composite` reconciler. Input and output of the relevant functions are compared similarly to the other tested methods in `composition_types_test.go`. There are specific tests to trigger each of the new error types.

I've applied the generated CRD's to a local `kind` cluster, and confirmed their functionality manually using a test XRD + Composition definition:

```yaml 
---
metadata:
  name: compositeobjectstoragebuckets.example.org
spec:
  group: example.org
  names:
    kind: CompositeObjectStorageBucket
    plural: compositeobjectstoragebuckets
  claimNames:
    kind: ObjectStorageBucket
    plural: objectstoragebuckets
  defaultCompositionRef:
    name: aws-object-storage-bucket
  versions:
    - name: v1alpha1
      referenceable: true
      served: true
      schema:
        openAPIV3Schema:
          type: object
          properties:
            spec:
              type: object
              properties:
                parameters:
                  type: object
                  properties:
                    location:
                      description: Bucket Location
                      enum:
                        - us
                        - uk
                        - au
                      type: string
                      default: us
                    namePrefix:
                      description: Bucket Name Prefix
                      type: string
                  required:
                    - namePrefix
              required:
                - parameters
---
kind: ClusterRole
metadata:
  name: compositeobjectstoragebuckets.example.org
  labels:
    rbac.oam.dev/aggregate-to-controller: "true"
rules:
- apiGroups:
  - example.org
  resources:
  - objectstoragebuckets
  - objectstoragebuckets/status
  - compositeobjectstoragebuckets
  - compositeobjectstoragebuckets/status
  verbs:
  - '*'
---
kind: Composition
metadata:
  labels:
    provider: aws
  name: aws-object-storage-bucket 
spec:
  compositeTypeRef:
    kind: CompositeObjectStorageBucket
  patchSets:
    - name: metadata
      patches:
        toFieldPath: metadata.labels

    - name: parameters
      patches:
      - fromFieldPath: spec.parameters.location
        toFieldPath: spec.forProvider.locationConstraint
        transforms:
          - type: map
            map:
              us: us-east-1
              uk: eu-west-2
              au: ap-southeast-2

      - fromFieldPath: spec.parameters.namePrefix
        toFieldPath: metadata.generateName
        transforms:
          - type: string
            string:
              fmt: 'added-by-parameters-set-%s'
  resources:
    - base:
        kind: Bucket
        spec:
          deletionPolicy: Delete
          forProvider:
            acl: private
            locationConstraint: unknown
              rules:
                    sseAlgorithm: aws:kms
      patches:
        # Obviously this is more useful the more base resources you have :)
        - type: PatchSet
          patchSetName: metadata
          toFieldPath: metadata.annotations
        - type: PatchSet
          patchSetName: parameters

---
kind: CompositeObjectStorageBucket
metadata:
  name: test-bucket
  annotations:
    woo: lalala
  labels:
    patch-this-label: wahoo
spec:
  parameters:
    namePrefix: test
    location: us
```

Which produces the following `Bucket` definition on-cluster:

```yaml
kind: Bucket
metadata:
  annotations:
    woo: lalala
  creationTimestamp: "2020-11-18T17:11:58Z"
  generateName: added-by-parameters-set-test
  generation: 1
  labels:
    crossplane.io/composite: test-bucket
    patch-this-label: wahoo
  name: added-by-parameters-set-test4s4s2
  ownerReferences:
  - apiVersion: example.org/v1alpha1
    controller: true
    kind: CompositeObjectStorageBucket
    name: test-bucket
    uid: 05c370d9-b152-4d5c-a0e7-f971a9ce94eb
  resourceVersion: "247129"
  selfLink: /apis/s3.aws.crossplane.io/v1beta1/buckets/added-by-parameters-set-test4s4s2
  uid: 5d5d4346-0860-45aa-b3e9-f31a93654eb6
spec:
  deletionPolicy: Delete
  forProvider:
    acl: private
    locationConstraint: us-east-1
      rules:
          sseAlgorithm: aws:kms
```


pkg/controller/apiextensions/composite/reconciler.go
```diff
@@ -366,7 +366,15 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error)
...
 	for i := range refs {
 		cd := composed.New(composed.FromReference(refs[i]))
-		if err := r.composed.Render(ctx, cr, cd, comp.Spec.Resources[i]); err != nil {
+
+		ct := comp.Spec.Resources[i]
+
+		if err := ct.InjectCompositionData(comp.Spec); err != nil {
+			log.Debug(errInjectCompData, "error", err)
+			r.record.Event(cr, event.Warning(reasonCompose, err))
+		}
```
negz, line 375: Do we want to return early from the reconcile if this injection fails? At the moment it seems like we'll proceed and attempt to do what we can with the presumably partially complete patches. If so we should add a test to cover this case.

apis/apiextensions/v1beta1/composition_types.go
```diff
@@ -133,14 +206,36 @@ type ReadinessCheck struct {
...
+type patchSetMapType map[string][]Patch
+
+// PatchType is type of the patch to be chosen.
+type PatchType string
+
+// Accepted TransformTypes.
+const (
+	PatchTypeStandard  PatchType = "std"
+	PatchTypeReference PatchType = "ref"
+)
```
negz, line 225: We have a handful of additional use cases where patch types could be used - I'd like to make sure we consider all of them before we introduce any. I'll write up an issue tracking the ones I know of so we can make sure to pick types that will be as future proof as possible.

negz (commented):
Awesome work @benagricola! My comments are mostly nitpicks - I think the main thing we need to figure out before merging this is going to be what patch types we want to support.

apis/apiextensions/v1beta1/composition_types.go
negz: Here's that issue: https://github.com/crossplane/crossplane/issues/1974

apis/apiextensions/v1beta1/composition_types.go
```diff
@@ -102,7 +112,70 @@ type ComposedTemplate struct {
 	ReadinessChecks []ReadinessCheck `json:"readinessChecks,omitempty"`
 }
 
-// TypeReadinessCheck is used for readiness check types
+// mapPatchSets processes a list of PatchSets and returns a map of PatchSet
+// ID to Patches.
+func (c *ComposedTemplate) mapPatchSets(patchSets []PatchSet) (patchSetMapType, error) {
```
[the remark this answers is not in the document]
benagricola: Nope. Split this out into a separate function to avoid cyclomatic complexity on `InjectCompositionData()` but it doesn't need access to `c` so can be a standalone private function. Will fix.

apis/apiextensions/v1beta1/composition_types.go
```diff
@@ -102,7 +112,70 @@ type ComposedTemplate struct {
...
+	var patchSetMap = patchSetMapType{}
```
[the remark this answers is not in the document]
benagricola: Roger, will rename some of my long vars along with the above

apis/apiextensions/v1beta1/composition_types.go
```diff
@@ -133,14 +206,36 @@ type ReadinessCheck struct {
...
-// after piping the value that is at FromFieldPath of the target resource through
-// transformers.
+// PatchSet defines a named list of patches on a composition.
+type PatchSet struct {
+	ID      string  `json:"id"`
+	Patches []Patch `json:"patches"`
+}
+
+// patchSetMap type stores a list of patches from a set, against the set ID.
+type patchSetMapType map[string][]Patch
```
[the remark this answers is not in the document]
benagricola: Name conflict with the long var name in the `Inject...` and `mapPatchSets` functions, after var rename I'll rename it back to `patchSetMap`

pkg/controller/apiextensions/composite/reconciler.go
benagricola: Yes. I think I even wrote some blurb about that in one of the tests, or one of the comments, and then forgot to actually implement it 🤦 Will Fix

benagricola:
@negz Thanks 👍 I've committed some fixes for the nits outlined, and the reconciler short circuit. I've also added a test to confirm that functionality when trying to reconcile with an invalid patch.

Agree about the patch types and your reasoning there. I'll have a think about this and post my thoughts over on #1974

apis/apiextensions/v1beta1/composition_types.go
```diff
@@ -133,14 +206,36 @@ type ReadinessCheck struct {
...
+// A standard patch is field based, patching the field on the base resource at
+// ToFieldPath after piping the value that is at FromFieldPath of the target
+// resource through transformers.
 type Patch struct {
 
 	// FromFieldPath is the path of the field on the upstream resource whose value
 	// to be used as input.
-	FromFieldPath string `json:"fromFieldPath"`
+	// +optional
+	FromFieldPath string `json:"fromFieldPath,omitempty"`
```
negz, line 238: Ah - I missed this earlier, but API conventions call for this to be a `*string` now that it's optional. Ditto for `Type` and `ID` below.

https://github.com/kubernetes/community/blob/69c5a105d684e8de893f3a7a9765e935410e6ed1/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required

apis/apiextensions/v1beta1/composition_types.go
```diff
@@ -152,6 +247,24 @@ type Patch struct {
...
+	// +optional
+	ID string `json:"id,omitempty"`
+}
+
+// SetDefaults makes sure patches have the correct default values applied.
+func (c *Patch) SetDefaults() {
+	if c.Type == "" {
+		c.Type = PatchTypeStandard
+	}
 }
```
negz, line 268: We recently updated the project to use v1 CRDs, which support default values as part of the CRD schema. It seems like kubebuilder has undocumented support for this, which may allow us to avoid this `SetDefaults` method. See https://github.com/kubernetes-sigs/controller-tools/pull/337/files

negz (commented):
This is looking good to me. I noticed one or two things that I missed on my first review, but otherwise I think this will be good to merge once we've bikeshedded our first set of patch type names.

docs/introduction/composition.md
```diff
@@ -307,6 +307,16 @@ spec:
   compositeTypeRef:
     apiVersion: example.org/v1alpha1
     kind: CompositeMySQLInstance
+
```
muvaf, line 310: Could you add a one-liner update description at the top of this document similar to https://github.com/crossplane/crossplane/blob/master/design/one-pager-managed-resource-api-design.md#revisions

apis/apiextensions/v1beta1/composition_types.go
```diff
@@ -152,6 +247,24 @@ type Patch struct {
...
+	// Type can be set to refer to a named patch defined against the parent
+	// composition. If a Patch has a type set, then all other fields will be
+	// ignored and the referenced patch will be used instead.
+	// +optional
+	Type PatchType `json:"type,omitempty"`
+
+	// ID can be set on a patch of type ref to refer to a specific patch set by
+	// name.
+	// +optional
+	ID string `json:"id,omitempty"`
```
muvaf, line 260:
```suggestion
	Name string `json:"name,omitempty"`
```

I think `name` is usually what's used in Kubernetes API as identifier if it's human-written. Also, `Patch` uses `name` field by default to identify resources in an array if I remember correctly.

apis/apiextensions/v1beta1/composition_types.go
```diff
@@ -152,6 +247,24 @@ type Patch struct {
 	// input to be transformed.
 	// +optional
 	Transforms []Transform `json:"transforms,omitempty"`
+
+	// Type can be set to refer to a named patch defined against the parent
+	// composition. If a Patch has a type set, then all other fields will be
+	// ignored and the referenced patch will be used instead.
+	// +optional
+	Type PatchType `json:"type,omitempty"`
```
muvaf, line 255: Let's say we have only the `Type` field. The logic users would think that there is a global pool of patches that I can choose from and I can contribute to that pool via `spec.patchSets`. Instead, we have two layers of classification; type of the patch, which could be patchSet or something else and also ID of a patch in that pool. To me, it feels like having the global pool is a simpler model for users to tinker when they write their compositions because number of custom compositions will almost never reach 10. So, having two layer identification pools feels kind of unnecessary.

We can have only `Type` field (or `Name`), some hard-coded types like `FromComposite`, `FromHTTPServer` etc. and the ones that users write in `spec.patchSets` (or preferably `spec.customPatchTypes` if you go with `Type`) like `MyRegionPatch`. Then when user chooses one, we'd first go over the hard-coded ones and if we can't find it, we'll iterate over the ones they supplied.

apis/apiextensions/v1beta1/composition_types.go
```diff
@@ -102,7 +112,70 @@ type ComposedTemplate struct {
...
+	}
+	return m, nil
+}
+
+// InjectCompositionData performs processing tasks on a composed template
+// prior to rendering. It is is executed as part of the reconciliation
+// loop, just after the ComposedTemplate is instantiated, and is passed
+// the Composition Spec - this allows data to be passed between the composition
+// and the composed, such as PatchSets.
+func (c *ComposedTemplate) InjectCompositionData(comp CompositionSpec) error {
```
muvaf, line 139: `InjectCompositionData` feels like we're bringing in a blob of data to be used in spec of the managed resources but in reality we just inject patches. Maybe `DereferencePatches` or `FlattenPatches`?

muvaf (commented):
Thanks for the PR @benagricola ! Code looks close to merge but I have suspicions about having both `type` and `name` to identify the patch.

apis/apiextensions/v1beta1/composition_types.go
benagricola: This is the approach I tried to take first and failed, so went the simpler route to start with 🙃 - partly as I've never worked with the Kubernetes CRD spec before and confused myself with recursive patch types.

I think I have a good handle on the implementation for this now, but it means that when defining a patch in `spec.customPatchTypes`, we either need to define a `Type` and a `Name` (since the base type of the custom patch needs to be known), or we need to restrict the patches defined under `spec.customPatchTypes` to be a single `Patch` type - in this case I would use the `PatchSet` type as other patch types can be nested within it.


Assume we *use* the predefined and custom patches like this:

```yaml
  resources:
  - base:
      # ...
    patches:
    - type: Metadata # Dereferenced
    - type: PatchSet # Inlined PatchSet, handled as-is (pointless but here as an example)
      patches:
      - fromFieldPath: spec.parameters.namePrefix
        toFieldPath: metadata.generateName
    - fromFieldPath: spec.parameters.blah # Inlined FromCompositeFieldPath
      toFieldPath: spec.forProvider.providerField
    - type: Parameters # Dereferenced
```

We have two custom patch types - `Metadata` and `Parameters`.

If we allow type definition in `customPatchTypes` they can be defined like this:

```yaml
kind: Composition
spec:
  customPatchTypes:
  # Custom PatchSet called Metadata
  - name: Metadata # Dereferenced as PatchSet
    type: PatchSet
    patches:
      fromFieldPath: metadata.labels
    - fromFieldPath: metadata.annotations # Type defaulted to FromCompositeFieldPath

  # Custom FromCompositeFieldPath called Parameters
  - name: Parameters # Dereferenced as individual Patch
    type: FromCompositeFieldPath
    fromFieldPath: spec.parameters.location
    toFieldPath: spec.forProvider.locationConstraint
```

Where `Metadata` is dereferenced to a `PatchSet` and `Parameters` is dereferenced to a single `FromCompositeFieldPath` Patch.

If we restrict `spec.customPatchTypes` to only defining `PatchSet` instances:

```yaml
kind: Composition
spec:
  customPatchTypes:
  # Custom PatchSet called Metadata
  - type: Metadata # Dereferenced as PatchSet
    patches:
      fromFieldPath: metadata.labels
    - fromFieldPath: metadata.annotations # Type defaulted to FromCompositeFieldPath

  # Custom FromCompositeFieldPath nested inside a PatchSet
  - type: Parameters # Dereferenced as PatchSet
    patches:
    - fromFieldPath: spec.parameters.location
      toFieldPath: spec.forProvider.locationConstraint
```

In practice the differences are small - but I _think_ if we restrict `spec.customPatchTypes` to being **only** `PatchSet` instances, then we avoid the need of the additional name field, we get consistency between the `type` field on `resource` patches as well as `spec.customPatchTypes` and we have just as much flexibility, it just means if you want to use a different patch type in `spec.customPatchTypes` you have to nest it in the `patches` field.

Does this make sense? 

From an implementation perspective, `PatchSet` just defines a list of `patches` and an `Apply` method which loops the `patches` and runs their `Apply` method.

benagricola:
The most recent commit takes into account the previous feedback, and implements the second `customPatchTypes` format from my above comment, where `customPatchTypes` can contain only `PatchSet`.

I've tested with the following composition on a `kind` cluster, as well as using the tests from the previous commits.

Given that the `Patch.Apply()` methods now need to validate their own input settings, I've tried to add some additional tests for those but haven't been able to make the patches apply correctly to the `fake.Composed` resources (`FromFieldPath` does not seem to match an existing field). Tests for `Patch.Apply()` will follow if I can work that out 😄 

```yaml
kind: Composition
metadata:
  labels:
    provider: aws
  name: aws-object-storage-bucket 
spec:
  compositeTypeRef:
    kind: CompositeObjectStorageBucket
  customPatchTypes:
    - type: Metadata
      patches:
        toFieldPath: metadata.labels

    - type: Parameters
      patches:
      - fromFieldPath: spec.parameters.location
        toFieldPath: spec.forProvider.locationConstraint
        transforms:
          - type: map
            map:
              us: us-east-1
              uk: eu-west-2
              au: ap-southeast-2

      - fromFieldPath: spec.parameters.namePrefix
        toFieldPath: metadata.generateName
        transforms:
          - type: string
            string:
              fmt: 'added-by-parameters-set-%s'
  resources:
    - base:
      # ...
      patches:
        - type: Metadata
          toFieldPath: metadata.annotations
        - type: Parameters
```

pkg/controller/apiextensions/composite/reconciler_test.go
```diff
@@ -36,6 +36,13 @@ import (
 	"github.com/crossplane/crossplane/apis/apiextensions/v1beta1"
 )
 
+// Various fields on our types take a *string, but
+// We want to pass them literals or values from constants.
+// Pass them through this first
+func strPtr(x string) *string {
+	return &x
+}
```
negz, line 44: There's an upstream util for this we could use - https://github.com/kubernetes/utils/blob/master/pointer/pointer.go#L91

apis/apiextensions/v1beta1/composition_types.go
negz: I had a similar thought, but preferred to stick with `InjectCompositionData` since that name would make sense if this method ever did more than inject patch sets.

apis/apiextensions/v1beta1/composition_types.go
negz: The `customPatchSets` pattern feels like a regression to me compared to the `patchSets` API where the type of patch the and the identifier/name were distinct.

It seems to me that we are overloading the `type` field to cover both the "built in" types of patch and the extended types of patch (from the `customPatchSets` array). As far as I can tell the main advantage to this approach is needing one fewer field (i.e. `type` does double duty and covers the `name` or `id` field). The main disadvantage is that it makes it harder to distinguish what the type field is referring to at a glance. Is `type: Widget` a native type or defined by `customPatchTypes`? What would a user new to composition think if four of of five patch types were declared in the `Composition`, but the fifth was some magic value they couldn't figure out the source of?

It's also a little odd to me that a custom "type of patch" is actually a custom _set of patches_.

apis/apiextensions/v1beta1/composition_types.go
negz: I agree that `name` is more common in this context (identifying an element in an array of objects). However I would be cautious as to whether `name` could be mistaken or confused with the `metadata.name` of the composed resource, rather than the name of the patch set.

This could be fine in this context, but I will almost certainly use `id` rather than `name` to identify elements within the `resources` array in https://github.com/crossplane/crossplane/issues/1909 (where `name` is at a higher risk of being interpreted to mean "the name of the resource" rather than "the name of the template"), and thus  prefer `id` to keep the `Composition` API internally consistent.

apis/apiextensions/v1beta1/composition_types.go
benagricola: Yeah my original reasoning behind `InjectCompositionData` was that it applied a new integration point between the composition and the composed resources, that _at the moment_ only does `PatchSet` injection.

Thinking about this more, I wonder if in future there would ever be movement in the reverse direction, in which case `InjectCompositionData` might also be too specific in a single direction.

pkg/controller/apiextensions/composite/reconciler_test.go
benagricola: TIL 😃 Switching over to use that

apis/apiextensions/v1beta1/composition_types.go
benagricola: I kinda liked the succinctness of overloading the type field but you definitely have a solid point there - the `type` field either referred to a `PatchType` (of which `PatchSet` is one) but also an _instance_ of `PatchSet` with a particular name. I can definitely see that being confusing for new users when working out where patches are coming from.

Also, if the type field can only contain known values, then it can be validated using the `+kubebuilder:validation:Enum=` setting, which avoids a situation where you typo an inbuilt patch type but don't get an error until the composite resource is created, because that is the point where `PatchSet` items from the Composition are injected into the composed resource.

I've changed the API back to using `patchSets` since this is effectively what they were anyway, and no longer override the type field within the defined patches, but use the name instead.

Now when applying an invalid Composition I get: `The Composition "aws-object-storage-bucket" is invalid: spec.resources.patches.type: Unsupported value: "PatchxxXY": supported values: "FromCompositeFieldPath", "PatchSet"`

muvaf: > The main disadvantage is that it makes it harder to distinguish what the type field is referring to at a glance. Is type: Widget a native type or defined by customPatchTypes?

How much would they be concerned with knowing from where it comes from though? Even if they are, it's a top-level field. The non-author readers would see the custom patch types immediately as well since it's a top-level field.

> What would a user new to composition think if four of of five patch types were declared in the Composition, but the fifth was some magic value they couldn't figure out the source of?

The fact that the name of the field is `customPatchTypes` already implies that there is an existing set of patch types that author made addition to. So, I feel like it's not that much of a surprise.

> It's also a little odd to me that a custom "type of patch" is actually a custom set of patches.

I think I can see that being confusing. So, I'd like us examine a similar case in Kubernetes API where users first define the set of `volumes` and then use them in the `containers` array of the pod, under `volumeMounts`:
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: configmap-pod
spec:
  containers:
    - name: test
      image: busybox
      volumeMounts:
        - name: config-vol
          mountPath: /etc/config
  volumes:
    - name: config-vol
      configMap:
        name: log-config
        items:
          - key: log_level
            path: log_level
```

IMO, `volumes` maps to the custom patches array and `volumeMounts` could map to `patches` in the `resources` array. So, instead of trying so hard to reuse `fromFieldPath` and `toFieldPath` for many things, we can have a type distinction in the field name and let each type has the diverse set of fields and mechanism available to them. The examples above would look like the following:
```yaml
kind: Composition
spec:
  customPatches:
  - name: Metadata
    resourcePatches: # The mechanism of the patch is identified by the field used here.
  - name: Location
    resourcePatch:
      fromFieldPath: spec.parameters.location
      toFieldPath: spec.forProvider.locationConstraint
resources:
- base:
   # ...
  patches:
  - name: Metadata # Name is given only if you want to use the custom patches you named.
  - name: Location
  # compositePatch means we're patching the composite.
  - compositePatch: # No name is necessary if you will use built-in patches. 
      fromFieldPath: status.atProvider.ARN
      toFieldPath: status.rdsARN
```

One difference between volumes semantics is that they don't have any built-in volumes so they always use `name` in the `containers[*].volumeMounts[*]`. In our case, user will give a name in `resources[*].patches[*]` if they'd like to use the custom patches they defined. If the patch is unique to that specific resource, then they won't need to use `name` field, they will just configure the patch to define the mechanism. This would also make it clear from where the patch comes from.

benagricola:
As of this most recent commit I think most of the previous nits should be cleared up. I haven't managed to get a working test for `Patch.Apply()` yet due to issues getting it to find the `from` path for the patch on a `fake.Composite`, so there's a bit of a gap there where we're not appropriately testing the handoff of the `Apply()` method to the individual patch types.

Aside from that, there's obviously still some questions / concerns around the user-facing API style, but with the code as it currently is it wasn't too difficult to switch from one to the other - so I think I'll 'tools down' until next week - thanks for all the feedback so far, guys 👍

apis/apiextensions/v1beta1/composition_types.go
negz: @muvaf I think the example of volumes makes sense, but we're going to need patch 'types' as well as 'names' that are references into what you've called the `customPatches` array. So to me, this takes us back to the API in which a reference to the array of reusable patches (by name or ID) is a distinct type of patch. In general, this feels like a "more is more" situation. I can't see what we gain by overloading the type/name concept except for one fewer field for folks to type, at the expense of more cognitive load figuring out what that field does.

apis/apiextensions/v1beta1/composition_types.go
negz: I'm happy to defer to your judgement on naming here @benagricola. I don't think we need anything too future proof - we can always rename it if its function changes in future. :)

apis/apiextensions/v1beta1/composition_types.go
```diff
@@ -133,35 +212,145 @@ type ReadinessCheck struct {
...
+		patcher = c.PatchFromCompositeFieldPath
+	case PatchTypePatchSet:
+		patcher = c.PatchSet
+	default:
+		return errors.Errorf(errInvalidPatchType, c.Type)
+	}
+
+	// An interface equals nil only if both the type and value are nil. Above,
+	// even if t.<Type> is nil, its type is assigned to "transformer" but we're
+	// interested in whether only the value is nil or not.
```
negz, line 269: Nit: Would it be simpler to check whether the `c.PatchSet` or `c.PatchFromCompositeFieldPath` structs are nil inside the `switch` to avoid needing to use reflection here?

apis/apiextensions/v1beta1/composition_types.go
```diff
@@ -133,35 +212,145 @@ type ReadinessCheck struct {
...
+}
+
+// PatchSet defines a list of patches as a single grouping.
+// It can contain all other patch types except itself.
+type PatchSet struct {
+	// ID is the unique identifier of the PatchSet. ID is required when
+	// defining a new PatchSet, and also refers to a PatchSet from within
+	// the list of Patches on a resource.
+	// +optional
+	ID *string `json:"id,omitempty"`
```
negz, line 289: Is there any case where `id` is optional? Could we require it?

apis/apiextensions/v1beta1/composition_types.go
```diff
@@ -133,35 +212,145 @@ type ReadinessCheck struct {
...
+)
+
+// Patch objects are applied between the composite and composed resources.
+// Their exact behaviour depends on the Type selected.
+// If the given type does not match one of the standard patch types, then
+// it may refer to a customPatchSet defined on the Composition. In this case,
+// the referenced patch will be dereferenced into the patch list in its' place.
 type Patch struct {
+	// Type can be a standard PatchSet value, or a 'Name' which refers
+	// to a custom patch type defined on the composition.
```
negz, line 234: I think this needs updating now that we're back at the `PatchSets` approach.

docs/introduction/composition.md
```diff
@@ -7,6 +7,13 @@ indent: true
 
 # Composing Infrastructure
 
+## Revisions
+
+* 1.1 - Ben Agricola (@benagricola)
+  * Updated [Specify How Your Resource May Be Composed](#specify-how-your-resource-may-be-composed) with examples outlining how to define and consume patch sets.
```
negz, line 13: Could you please wrap this at 80 chars to fit the rest of the doc.

pkg/controller/apiextensions/composite/reconciler_test.go
```diff
@@ -20,15 +20,17 @@ import (
...
 	corev1 "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/utils/pointer"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
+	"github.com/google/go-cmp/cmp"
+	"github.com/pkg/errors"
```
negz, line 32: Nit: These look like they snuck out into their own group; they can be merged with the above one.

pkg/controller/apiextensions/composite/reconciler.go
```diff
@@ -56,8 +56,8 @@ const (
 	errConfigure    = "cannot configure composite resource"
 	errPublish      = "cannot publish connection details"
 	errRender       = "cannot render composed resource"
-
-	errFmtRender = "cannot render composed resource at index %d"
+	errFmtRender    = "cannot render composed resource at index %d"
+	errInjectCD     = "cannot inject Composition data onto composite resource"
```
negz, line 60: Nit: I think the intent here was to keep `errFmt` and `err` strings separated, so I'd put this up under `errRender`.

apis/apiextensions/v1beta1/composition_types_test.go
```diff
@@ -22,9 +22,203 @@ import (
...
+		},
+		"UndefinedPatchSet": {
+			args{
+				comp: CompositionSpec{},
+				patches: []Patch{
+					{Type: PatchTypePatchSet, PatchSet: &PatchSet{ID: pointer.StringPtr("patch-set-1")}},
+				},
+			},
+			want{
+				// We expect the internal Patches field to be unmodified
```
negz, line 87: We have a convention (in some tests) of including a `reason` string in each test case that captures this kind of data, both as documentation for test readers and as something that can be printed if the test fails.

negz (approved):
Thanks @benagricola - this looks great! I'm excited to have you adding a new composition feature. I've left a few comments - mostly nits but one or two should be addressed before we merge this (mostly the `Type` docstring).

apis/apiextensions/v1beta1/composition_types.go
benagricola: The `PatchSet` type is used to both define a patch set in the `CompositionSpec` as well as reference it from the `Patch`. AIUI it needs to be optional when defined in the `Patch` otherwise it'll need to be set even when the `Type` is not `PatchSet`.

The `ID` field is currently checked at line 126 when the patch sets are mapped, but I suppose we could have another type def (`PatchSetDefinition` 🤷‍♂️) that requires the ID and is then converted to a normal `PatchSet` when mapped?

apis/apiextensions/v1beta1/composition_types.go
benagricola: Rather than in the `switch`, what if it becomes the concern of the actual patch type `Apply` method? 

The patch types already need to check that their own required fields are set correctly, so I wonder if just moving the pointer nil check there makes sense - leaving the `Patch.Apply` as a dispatcher?

apis/apiextensions/v1beta1/composition_types.go
muvaf: > I can't see what we gain by overloading the type/name concept except for one fewer field for folks to type, at the expense of more cognitive load figuring out what that field does.

I'm leaning towards succinctness because I think custom patches is an advanced use case where we can expect some experience. And even in cases where you don't define your own custom patches, this allows each built-in type to have its own set of fields separately from others. For making it more clear, the following pattern where we explicitly reference a custom patch could be easier to grasp.

```yaml
kind: Composition
spec:
  customPatches: # The struct here has 'name' field and inlined patch struct.
  - name: Metadata
    resourcePatches: # The mechanism of the patch is identified by the field used here.
  - name: Location
    resourcePatch:
      fromFieldPath: spec.parameters.location
      toFieldPath: spec.forProvider.locationConstraint
resources:
- base:
   # ...
  patches:
  - customPatchRef: # One of patch types in the patch struct.
      name: Metadata
  - customPatchRef:
      name: Location
  - compositePatch: # compositePatch means we're patching the composite, it's a type of patch.
      fromFieldPath: status.atProvider.ARN
      toFieldPath: status.rdsARN
  - resourcePatchList: # resourcePatchList is a list of resource patches where composed is patched.
    - fromFieldPath: spec.size
      toFieldPath: spec.forProvider.dbSize
```

Anyway, I believe this API is more inline with Kubernetes API, but I can see that the suggested top-level name/type separation approach is also acceptable.

apis/apiextensions/v1beta1/composition_types.go
negz: Got it - I think leaving it as is makes sense.

apis/apiextensions/v1beta1/composition_types.go
negz: That SGTM!

apis/apiextensions/v1beta1/composition_types.go
negz: @muvaf I responded to your similar comment in https://github.com/crossplane/crossplane/issues/1974. In short, it seems like your latest proposal is a breaking API change and would also benefit from a `type` discriminator field.

benagricola:
Force pushed because I forgot to `-s` the previous commit and the DCO help said to 🤦 

Again, this fixes the previously mentioned nits - adds patch apply testing, adds reasons to each of the tests, updated some comments and wrapped the doco changelog at 80 chars, moves the patch field sanity checking into the `Apply` method of each type, and defines a `PatchTypeCompatiblePatch` type which is what is nested under a `PatchSet` - this is done to avoid nesting `PatchSets` inside `PatchSets`, and avoid a recursive type def if we were to allow `Patch` to be defined under `PatchSet`

apis/apiextensions/v1beta1/composition_types.go
muvaf: Resolved in https://github.com/crossplane/crossplane/issues/1974

benagricola:
Previous commits fix an issue I found testing in `kind` related to the `ToFieldPath` being optional. I've addeed a test for this condition as well, and I _think_, barring any further nits, this resolves everything above.

apis/apiextensions/v1beta1/composition_types.go
```diff
@@ -133,35 +211,158 @@ type ReadinessCheck struct {
...
+	// PatchSets are defined on a Composition and can be referenced here,
+	// using their ID.
+	// +optional
+	*PatchSet `json:",inline"`
+
+	// PatchFromCompositeFieldPath is the 'standard' patch type that maps a field
+	// path from the composite resource to the composed resource, with optional
+	// transformations.
+	// +optional
+	*PatchFromCompositeFieldPath `json:",inline"`
```
negz, line 246: I'm not certain that we'll be able to cleanly embed/inline these different patch types in future, for example if `toFieldPath` and `fromFieldPath` were used by both `type: FromCompositeFieldPath` and `type: FromComposedFieldPath` (per the sketch at https://github.com/crossplane/crossplane/issues/1974).

apis/apiextensions/v1beta1/composition_types.go
```diff
@@ -133,35 +211,158 @@ type ReadinessCheck struct {
...
+}
+
+// PatchSet defines a list of patches as a single grouping.
+// It can contain all other patch types except itself.
+type PatchSet struct {
+	// ID is the unique identifier of the PatchSet. ID is required when
+	// defining a new PatchSet, and also refers to a PatchSet from within
+	// the list of Patches on a resource.
+	// +optional
+	ID *string `json:"id,omitempty"`
```
negz, line 293: I'm afraid that per our API sketch at https://github.com/crossplane/crossplane/issues/1974#issuecomment-733146966 @muvaf convinced me that `name` was preferable to `id` here after all. Would you mind updating?

We'd also like to call this `patchSetName` when referring to a patchset from a resource's array of patches to make it very clear what the field is for, so we probably can't use a single type for both cases.

apis/apiextensions/v1beta1/composition_types.go
```diff
@@ -133,35 +211,158 @@ type ReadinessCheck struct {
...
+	// (necessary because with multiple patch types, the input fields
+	// must be +optional).
+	if c == nil || c.FromFieldPath == nil {
+		return errors.New(errInvalidPatch)
+	}
+
+	// Default to patching the same field on the composed resource
+	if c.ToFieldPath == nil {
+		c.ToFieldPath = c.FromFieldPath
+	}
```
negz, line 358: Is this functionality new? I like it, but we should make sure to document it.

negz (changes requested):
Thanks for persisting with this @benagricola! I'm really impressed by the quality of your work here. @muvaf and I have agreed on an API per https://github.com/crossplane/crossplane/issues/1974#issuecomment-733146966 - I believe the only difference from your current implementation is renaming `id` to either `name` or `patchSetName` depending on the context in which it's used. Once that change is made this should be good to go.

Sorry for all the churn here as we figured out an API we can all live with. :)

apis/apiextensions/v1beta1/composition_types.go
negz: Per https://github.com/crossplane/crossplane/issues/1974#issuecomment-733146966 we decided to go with `name` and `patchSetName` here. :)

negz:
PS. Once this is good to go could you please update the PR description to match the updated API?

apis/apiextensions/v1beta1/composition_types.go
rozcietrzewiacz: @negz `ToFieldPath` is actually already advertised as working this way (see description of ToFieldPath in Patch), but not implemented. I was literally just about to make a separate PR with just that :smiley:

benagricola: Yeah - I actually implemented this because I already believed from the doco that it worked that way, and I'd only broken it by making all of the fields optional + string pointers 😆

negz: Well, that's convenient! Thanks for the update. :)

apis/apiextensions/v1beta1/composition_types.go
benagricola: I'm all aboard with the API described in #1974 - however I've been looking at implementing this now as 2 different types, and I'm having trouble getting my head around how to allow this to work without having extraneous fields that appear in the CRD definition.

Basically, if we define a `PatchSetDefinition` on the Composition, at some point that `PatchSetDefinition` has to be 'copied' to <somewhere> on the `ComposedTemplate` to allow it to be defined at some point in the future. Using the one `PatchSet` type which was embedded into `Patch`, it was possible to simply create a new `Patch` with the `PatchSet` struct configured.

With the additional type to allow the Name field to be changed in JSON (which has its' own benefits from a validation perspective), if we embed both into `Patch` then we end up with the following field definitions in the CRD:

```yaml
resources:
    items:
      properties:
        ...
        patches:
          items:
            properties:
              # Defined on CompositeFieldPath
              fromFieldPath:
                type: string
              # Defined on PatchSetDefinition
              name:
                description: Name is the unique identifier of the PatchSet. It is required when defining a new PatchSet, and used to refer to a PatchSet from within the list of Patches on a resource.
                type: string
              # Defined on PatchSet
              patchSetName:
                description: Name is the unique identifier for the PatchSet. It is used to refer to a PatchSet from within the list of Patches on a resource.
                type: string
              # Defined on PatchSetDefinition
              patches:
                # ...
              toFieldPath:
                type: string
              transforms:
                # ...
              type:
                default: FromCompositeFieldPath
                description: Type sets the patching behaviour to be used. Each patch type may require its' own fields to be set on the Patch object.
                enum:
                - FromCompositeFieldPath
                - PatchSet
                type: string
            required:
            - name
            - patches
```

Which won't work because for one, it defines `patches` as a required field (because it is, for a `PatchSetDefinition`).

I can't tell if I'm missing something and it's possible to 'hide' a field like `patches` from the CRD (i.e. make it internal) or I'm overthinking this and I should just add a `patches` field to the `PatchSet` definition (the one that refers to a `PatchSetDefinition` which in theory could be set by the user, but would be overwritten if they named a valid patchset to refer to.

Thoughts appreciated 🙂

benagricola:
^ changes the field names to match the API sketch as requested. I've also updated the PR description.

I'm still unsure about the definition of the `patches` list on `PatchSet`. It is currently set `+immutable` but that isn't really accurate, ideally it wouldn't be possible to set that value as an API user as it should always be populated by the Composition `PatchSetDefinition`.

Any input there appreciated as per my previous comment on the diff!

apis/apiextensions/v1beta1/composition_types.go
benagricola: Given that `FromCompositeFieldPath` is the only patch type right now (prior to this PR), ISTM that inlining the fields only makes sense from a backwards-compatibility perspective, which is why I did it - I assume the existing patch style in `v1beta1` needs to continue working as-is unless a new API version was used.

My assumption was that any new patch types that require the same field names would work in the same manner as `Transform` currently does.

But I now realise I have already violated that as the `PatchSet` type is also inlined 🤦 Does it make sense for `PatchFromCompositeFieldPath` to be the only inlined patch type, for backwards compatibility, and everything else (including `PatchSet`) becomes a map item of the same name? (e.g. `type: PatchSet` -> `patchSet: { patchSetName: ... }`)

negz (commented):
https://github.com/crossplane/crossplane/commit/584c66abb623b29da7ea2f1d99086218fd024d1e

@benagricola what do you think of an approach something like the above sketch? I believe it keeps things relatively simple - the idea is the controller would call `InlinePatchSets()` then proceed as usual. The main downside is that to anyone looking at the resulting Composition CRD it would look like patch sets could reference further patch sets, but doing so would throw an error. I feel comfortable with that; we could probably just document it in the `Type` docstring.

benagricola:
> [584c66a](https://github.com/crossplane/crossplane/commit/584c66abb623b29da7ea2f1d99086218fd024d1e)
> 
> @benagricola what do you think of an approach something like the above sketch? I believe it keeps things relatively simple - the idea is the controller would call `InlinePatchSets()` then proceed as usual. The main downside is that to anyone looking at the resulting Composition CRD it would look like patch sets could reference further patch sets, but doing so would throw an error. I feel comfortable with that; we could probably just document it in the `Type` docstring.

Yep - this is definitely cleaner. It also makes sense to me that input fields on `Patch` are not 'owned' by a particular type of patch, as they might be used across multiple 'types' (Thinking about `FromCompositeFieldPath` and e.g. `FromComposedFieldPath` which might use the same field names).

I'll have a look at implementing this in my fork and see how it looks.

apis/apiextensions/v1beta1/composition_types.go
```diff
@@ -53,6 +60,12 @@ type CompositionSpec struct {
 	// +immutable
 	CompositeTypeRef TypeReference `json:"compositeTypeRef"`
 
+	// PatchSets define a named patch that can be referenced from base resources
```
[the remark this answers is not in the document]
benagricola: Fixed in the most recent commit.

apis/apiextensions/v1beta1/composition_types.go
```diff
@@ -64,6 +74,50 @@ type CompositionSpec struct {
...
+		for _, p := range s.Patches {
+			if p.Type == PatchTypePatchSet {
+				return errors.New(errPatchSetType)
+			}
+		}
+		pn[s.Name] = s.Patches
+	}
+
+	for i, r := range cs.Resources {
+		po := []Patch{}
```
benagricola, line 91: @negz I've changed this slightly from the sketch, as `PatchSet` dereferenced patches would always be applied in-order _after_ all of the normal patches applied on the resource. 

This way maintains order between the standard patches and the ordering of the patches inside the referenced `PatchSets`, so e.g. you could have a `PatchSet` which copies all of `metadata.annotations` from the `Composition` and then have a later `Patch` definition which overwrites a single annotation.

pkg/controller/apiextensions/composite/reconciler.go
```diff
@@ -363,6 +363,13 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error)
 	refs := make([]corev1.ObjectReference, len(comp.Spec.Resources))
 	copy(refs, cr.GetResourceReferences())
 
+	// Inline PatchSets from Composition Spec before rendering
```
benagricola, line 366: @negz Wasn't 100% sure if this is where you meant by 
> the controller would call InlinePatchSets() then proceed as usual

But it works here as tested locally, and I'm not sure there's anywhere else that the `CompositionSpec` is available to have the method called on it 🙂

benagricola:
`check-diff` failure appears to be transient, ssl error attempting to download `helm`. Just reran locally to confirm.

hasheddan:
@benagricola sorry about that! Just kicked off jobs again so hopefully it passes this time :)

apis/apiextensions/v1beta1/composition_types.go
```diff
@@ -238,130 +208,52 @@ type Patch struct {
...
-// PatchFromCompositeFieldPath is a standard patch - taking an input value
-// from the composite resource, and applying it to the composed resource,
-// optionally transforming it along the way.
-type PatchFromCompositeFieldPath struct {
 	// FromFieldPath is the path of the field on the upstream resource whose value
-	// to be used as input.
+	// to be used as input. Required when type is FromCompositeFieldPath.
 	// +optional
-	FromFieldPath *string `json:"fromFieldPath,omitempty"`
+	FromFieldPath *string `json:"fromFieldPath"`
```
negz, line 214: I think we should keep the `omitempty` tag here.

pkg/controller/apiextensions/composite/reconciler.go
negz: This matches about what I was thinking.

apis/apiextensions/v1beta1/composition_types.go
negz: Good call - it makes sense to me to keep the ordering as close to original as possible.

negz (approved):
Awesome, thanks for persisting with this @benagricola! I left one commit about an `omitempty` tag that I missed in my sketch. If you can add that back in and tidy up your commit history I'll merge this today.

benagricola:
So.. I in hindsight I made a horrible hash of my git commit history implementing then modifying the various API types - I wasn't even able to separate out the documentation changes into a single commit due to the way it was done 😭 

I've rebased onto master and re-applied the diff in stages so there's one commit for the imp, one for tests and one for documentation - I hope that's acceptable now that this is approved, otherwise please let me know and I'll do my best to fixup / squash and push the previous history of the branch 🙂

muvaf:
Congrats @benagricola ! This was a lot to take on and I'm happy to see it's merged! Thank you!

PR #1983, opened 2020-11-21
Move securityContext items into chart's input values

khos2ow:

The purpose of this move is to be able to override them on the fly for any platforms that restrict setting them or in situations where they need to be completely disabled, such as development environments.

I have:

- `make build`
- `make helm.prepare`
- `./cluster/local/kind.sh up`
- modify `cluster/charts/crossplane/values.yaml` as follow:
  ```yaml
  ...
  securityContextCrossplane:
    runAsUser: 3000
    runAsGroup: 3000
  ...
  ```
- make sure the corresponding values do exist in the deployment:
  ```bash
  $ k -n crossplane-system get deployments.apps crossplane -o yaml
  apiVersion: apps/v1
  kind: Deployment
  metadata:
    labels:
      app: crossplane
      app.kubernetes.io/managed-by: Helm
      chart: crossplane-0.0.1
      heritage: Helm
      release: crossplane
    name: crossplane
    namespace: crossplane-system
  spec:
    ...
    template:
      ...
      spec:
        containers:
        - image: crossplane/crossplane:v1.0.0-rc.0.27.g73352a7a-dirty
          ...
          securityContext:
            readOnlyRootFilesystem: true
            runAsGroup: 3000
            runAsUser: 3000
  ...
  ```  

### DIFF
--- a/cluster/charts/crossplane/README.md
+++ b/cluster/charts/crossplane/README.md
@@ -75,6 +75,10 @@ and their default values.
 | `resourcesCrossplane.limits.memory` | Memory resource limits for Crossplane | `512Mi` |
 | `resourcesCrossplane.requests.cpu` | CPU resource requests for Crossplane | `100m` |
 | `resourcesCrossplane.requests.memory` | Memory resource requests for Crossplane | `256Mi` |
+| `securityContextCrossplane.runAsUser` | Run as user for Crossplane | `2000` |
+| `securityContextCrossplane.runAsGroup` | Run as group for Crossplane | `2000` |
+| `securityContextCrossplane.allowPrivilegeEscalation` | Allow privilege escalation for Crossplane | `false` |
+| `securityContextCrossplane.readOnlyRootFilesystem` | ReadOnly root filesystem for Crossplane | `true` |
 | `packageCache.medium` | Storage medium for package cache. `Memory` means volume will be backed by tmpfs, which can be useful for development. | `""` |
 | `packageCache.sizeLimit` | Size limit for package cache. If medium is `Memory` then maximum usage would be the minimum of this value the sum of all memory limits on containers in the Crossplane pod. | `5Mi` |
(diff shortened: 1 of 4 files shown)


cluster/charts/crossplane/README.md
```diff
@@ -75,13 +75,20 @@ and their default values.
 | `resourcesCrossplane.limits.memory` | Memory resource limits for Crossplane | `512Mi` |
 | `resourcesCrossplane.requests.cpu` | CPU resource requests for Crossplane | `100m` |
 | `resourcesCrossplane.requests.memory` | Memory resource requests for Crossplane | `256Mi` |
+| `securityContextCrossplane.runAsUser` | Run as user for Crossplane | `2000` |
+| `securityContextCrossplane.runAsGroup` | Run as group for Crossplane | `2000` |
+| `securityContextCrossplane.allowPrivilegeEscalation` | Allow privilege escalation for Crossplane | `false` |
+| `securityContextCrossplane.readOnlyRootFilesystem` | ReadOnly root filesystem for Crossplane | `true` |
```
hasheddan, line 81: @khos2ow do you think it would make sense to just allow a user to provide any of the `securityContext` values, rather than scoping it to 4 here and 2 below? I think the change in the helm chart below would actually allow them to set values not defined here, but the documentation doesn't indicate that

khos2ow: That's a good point and I think it comes down to provide explicit documentation what user can do. But technically Crossplane depends on these 4 and 2 items. That said I think it makes sense to leave them as is (as default needed values) but user can override them or pass additional items of `securityContext` for helm to use.

hasheddan: Ah, that is a good point. This documentation is not only showing what the user can set, it is also showing what we set by default 👍 LGTM, thanks!

hasheddan (approved):

PR #1988, opened 2020-11-25
composition.md: XRC->XR change propagation fixed

rozcietrzewiacz:
https://github.com/crossplane/crossplane/pull/1970 fixed the issue in master

### DIFF
--- a/docs/introduction/composition.md
+++ b/docs/introduction/composition.md
@@ -786,10 +786,6 @@ Events:
 Composite resources are an alpha feature of Crossplane. At present the below
 functionality is planned but not yet implemented:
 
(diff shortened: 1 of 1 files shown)


hasheddan (changes requested):
@rozcietrzewiacz since this change is in `master` we should add the docs update to the `master` branch for next release. This document is accurate for `release-0.14`.

Also, looks like we need DCO here :)

rozcietrzewiacz:
@hasheddan Fixed ;)

hasheddan (approved):
LGTM! Thanks @rozcietrzewiacz!

Issue #1856, opened 2020-10-18
Composition unable to map enum to bool


prasek:
### What happened?
Tried mapping an enum to a bool but composition didn't allow it.

With the XRD property below:
```
      connectivity:
        type: string
        description: network connectivity
        enum:
        - public
        - private
```

and a Composition that tried to map the enum to a bool
```
      patches:
      - fromFieldPath: "spec.parameters.connectivity"
        toFieldPath: "spec.forProvider.publiclyAccessible"
        transforms:
        - type: map
          map:
            public: true
            private: false
```
`kubectl crossplane build` returned the following:
```
kubectl crossplane: error: failed to build package: failed to parse package: v1alpha1.Composition.Spec: v1alpha1.CompositionSpec.Resources: []v1alpha1.ComposedTemplate: v1alpha1.ComposedTemplate.v1alpha1.ComposedTemplate.Patches: []v1alpha1.Patch: v1alpha1.Patch.Transforms: []v1alpha1.Transform: v1alpha1.Transform.Type: Map: unmarshalerDecoder: json: cannot unmarshal bool into Go value of type string, error found in #10 byte of ...|lic":true},"type":"m|..., bigger context ...|ransforms":[{"map":{"private":false,"public":true},"type":"map"}]}]},{"base":{"apiVersion":"database|...
```

I then tried mapping as string values, hoping it would get coerced:
```
      patches:
      - fromFieldPath: "spec.parameters.connectivity"
        toFieldPath: "spec.forProvider.publiclyAccessible"
        transforms:
        - type: map
          map:
            public: "true"
            private: "false"
```

But in this case it was silently ignored and didn't set the values.

<details><summary>definition.yaml</summary>

```
apiVersion: apiextensions.crossplane.io/v1alpha1
kind: CompositeResourceDefinition
metadata:
  name: compositepostgresqlinstances.aws.platformref.crossplane.io
spec:
  claimNames:
    kind: PostgreSQLInstance
    plural: postgresqlinstances
  connectionSecretKeys:
    - username
    - password
    - endpoint
    - port
  crdSpecTemplate:
    group: aws.platformref.crossplane.io
    version: v1alpha1
    names:
      kind: CompositePostgreSQLInstance
      plural: compositepostgresqlinstances
    validation:
      openAPIV3Schema:
        type: object
        properties:
          spec:
            type: object
            properties:
              parameters:
                type: object
                properties:
                  storageGB:
                    type: integer
                  networkRef:
                    type: object
[...]

PR #1994, opened 2020-12-01
Add type conversion transform

muvaf:

Adds a new transformer type to do type conversions.

Fixes https://github.com/crossplane/crossplane/issues/1856

I have:

```yaml
metadata:
spec:
  group: database.example.org
  names:
  claimNames:
    kind: PostgreSQLInstance
    plural: postgresqlinstances
  connectionSecretKeys:
    - username
    - password
    - endpoint
    - port
  versions:
  - name: v1alpha1
    served: true
    referenceable: true
    schema:
      openAPIV3Schema:
        type: object
        properties:
          spec:
            type: object
            properties:
              parameters:
                type: object
                properties:
                  storageGB:
                    type: integer
                  public:
                    type: string
                required:
                  - storageGB
            required:
              - parameters
---
kind: Composition
metadata:
  labels:
    provider: gcp
    guide: quickstart
spec:
  compositeTypeRef:
  resources:
    - base:
        kind: CloudSQLInstance
        spec:
          forProvider:
            databaseVersion: POSTGRES_9_6
            region: us-central1
            settings:
              tier: db-custom-1-3840
              dataDiskType: PD_SSD
              ipConfiguration:
                ipv4Enabled: true
                authorizedNetworks:
                  - value: "0.0.0.0/0"
          writeConnectionSecretToRef:
            namespace: crossplane-system
      patches:
          transforms:
            - type: string
              string:
                fmt: "%s-postgresql"
        - fromFieldPath: "spec.parameters.public"
          toFieldPath: "spec.forProvider.settings.ipConfiguration.ipv4Enabled"
          transforms:
          - type: convert
            convert:
              toType: bool
      connectionDetails:
        - name: port
          value: "5432"
```

The creation of the following YAML results in a `CloudSQLInstance` whose `spec.forProvider.settings.ipConfiguration.ipv4Enabled` field is `false`:
```yaml
kind: PostgreSQLInstance
metadata:
  name: my-db
  namespace: crossplane-system
spec:
  parameters:
    storageGB: 20
    public: "false"
  compositionSelector:
    matchLabels:
      provider: gcp
  writeConnectionSecretToRef:
    name: db-conn
```


apis/apiextensions/v1beta1/composition_types.go
```diff
@@ -217,10 +221,16 @@ type Transform struct {
...
-	// Map uses the input as a key in the given map and returns the value.
+	// Map uses the input as a key in the given map whose values can only be
+	// strings and returns the value.
 	// +optional
 	Map *MapTransform `json:"map,omitempty"`
 
+	// BoolMap uses the input as a key in the given map whose values can only
+	// be booleans and returns the value.
+	// +optional
+	BoolMap *MapTransform `json:"boolMap,omitempty"`
```
hasheddan, line 232: I appreciate the eye for backwards compatibility here, but do you think we could just adapt the existing implementation to infer a boolean if value in `map[string]string` is `true` or `false`? I believe this could be done without breaking API compatability

negz: I had a similar thought. Maybe we could add a flag that controlled that behaviour? I'm rusty on how this API works, but I was thinking something like:

```yaml
transforms:
- type: map
  # When this is set the strings 'true' and 'false' are
  # interpreted as boolean true/false.
  inferBooleanValues: true
  map:
    key1: true
```

This would maintain API and behaviour compatibility for existing folks without needing to create a new kind of map.

muvaf: I'm thinking the cases where we'd expand this to other types like `small: 1024, big: 4096` where the value is integer. In that case, we'd have a separate `inferIntegerValues`. Maybe we can have one field named `valueType: [int, bool, string, []int, ....]` (possibly use an enum for possible types or use reflect package for arbitrary input) and we'd cast the values into that type. We could have only one field and allow users to tell the type with `string` being implicit default to keep the existing behavior. Though we can also just say it wouldn't go much further than `string, int, bool` types anyway, so two additional fields would be good enough. What do you think?

hasheddan: I like the idea of `valueType` 👍

negz: Yeah, that sounds like a more future proof approach.

muvaf: Actually it seems like it's harder to get an option field for MapTransform. In order to do the following:
```yaml
transforms:
- type: map
  anOptionField: aValue
  map:
    key1: true
```

We'd have to add that option to the same level as transforms [here](https://github.com/crossplane/crossplane/blob/e2d7278/apis/apiextensions/v1beta1/composition_types.go#L210). And I wanted to cheated by adding it to under `map` like the following:
```
transforms:
- type: map
  map:
    anOptionField: aValue # maybe api-server reserves that "anOptionField" keyword for us
    key1: true
```

But we get the following error as expected during the registration of `Composition` CRD:
```
Error: failed to install CRD crds/apiextensions.crossplane.io_compositions.yaml: CustomResourceDefinition.apiextensions.k8s.io "compositions.apiextensions.crossplane.io" is invalid: spec.versions[1].schema.openAPIV3Schema.properties[spec].properties[resources].items.properties[patches].items.properties[transforms].items.properties[map].additionalProperties: Forbidden: additionalProperties and properties are mutual exclusive
```

But when you think about it, maybe this `valueType` could actually go for any transform? So, we'd do the following:
```yaml
- type: map
  valueType: bool
  map:
    key1: true
- type: map
  valueType: float64
  math:
    multiply: "1.2"
```

And then there wouldn't be a change in `map` transform specifically; we'd cast any value returned from the chosen transform resolver into the given `valueType`. Is everyone cool about this?

muvaf: We have discussed this a bit more offline and @negz brought up a great idea. We'd just add a new transform type to do a type conversion so that output from any transform could be input to that conversion transform. I'll update the PR to do that.

apis/apiextensions/v1beta1/composition_types_test.go
```diff
@@ -64,10 +71,72 @@ func TestMapResolve(t *testing.T) {
...
+			want: want{
+				err: errors.New(fmt.Sprintf(errFmtMapValueCastFailed, boolType)),
+			},
+		},
 	}
 	for name, tc := range cases {
 		t.Run(name, func(t *testing.T) {
-			got, err := (&MapTransform{Pairs: tc.m}).Resolve(tc.i)
+			n := name
+			fmt.Println(n)
```
hasheddan, line 138: looks like this may have been accidentally left behind

apis/apiextensions/v1beta1/composition_types.go
```diff
@@ -429,6 +435,98 @@ func (s *StringTransform) Resolve(input interface{}) (interface{}, error) {
...
+// The list of supported ConvertTransform input and output types.
+const (
+	ConvertTransformTypeString  = "string"
+	ConvertTransformTypeBool    = "bool"
+	ConvertTransformTypeInt     = "int"
+	ConvertTransformTypeFloat64 = "float64"
+)
+
+var (
+	conversions = map[string]func(interface{}) (interface{}, error){
```
muvaf, line 447: I went back and forth between having this map or nested `switch` statements for pairs. I feel like this is easier to write tests for and doesn't increase the cyclomatic complexity of the `Resolve` function like `switch` does (though we'd skip since it's not really more complex). You can see the `switch` structure in the earlier commit.

muvaf:
I'm planning on moving all this patch business into `internal` after this PR. What do you think?

hasheddan:
@muvaf that sounds good to me!

apis/apiextensions/v1beta1/composition_types.go
```diff
@@ -429,6 +435,98 @@ func (s *StringTransform) Resolve(input interface{}) (interface{}, error) {
...
+		"string.string": func(i interface{}) (interface{}, error) {
+			return i, nil
+		},
```
negz, line 450: Could we detect and handle the no-op case in which the from and to types are the same down in `Resolve`, so that we don't need to include the handful of no-op conversions in this map?

negz (commented):
A few comments, but looks good in general to me.

apis/apiextensions/v1beta1/composition_types.go
muvaf: yeah, just added that.

apis/apiextensions/v1beta1/composition_types.go
muvaf: `reflect.Kind` is just an `iota` and to get the actual string, we use `.String()` method. However, it doesn't work backwards, i.e. there is no function to take `int` and convert it to `reflect.Kind`. We can implement it but `reflect` doesn't expose the array so we'd have to duplicate those bindings, too. So, I just opted for using string as type for those fields:
```golang
type conversionPair struct {
  From string
  To string
}
```

muvaf:
image

I'm running `gofmt` repeatedly but couldn't figure out what the warning is about.

negz:
> I'm running gofmt repeatedly but couldn't figure out what the warning is about.

@muvaf Yeah the `gofmt -s` linter rule is always hard to diagnose. At a guess, I suspect it might want you to omit the `conversionPair` qualifier of each map key since it can be inferred from the map key type. e.g.:

```go
var conversions = map[conversionPair]func(interface{}) (interface{}, error) {
    {From: ConvertTransformTypeString, To: ConvertTransformTypeInt}: func(interface{}) (interface{}, error) { ... },
}

apis/apiextensions/v1beta1/composition_types.go
negz: >  i.e. there is no function to take int and convert it to reflect.Kind

Ah, this confused me for a bit but I think you're saying there's no way to take the string `"int"` and convert it to a `reflect.Kind`, right? I was initially thinking you were saying it wasn't possible to get the `reflect.Kind` of an `int`.

muvaf:
@negz thanks! this is ready for a final review.

negz (approved):

PR #1998, opened 2020-12-03
Move everything in pkg to internal directory

hasheddan:

In preparation for the v1.0 release of crossplane, all packages are
being moved to the internal directory as anything that should be able to
be imported by other packages should move to crossplane-runtime.

I have:

`make build`


negz:
@hasheddan Any idea why coverage dropped so much? I vaguely recall that the makefile explicitly runs tests in a specific test of dirs - one of which is probably `pkg`.

hasheddan:
@negz good catch! We do manually declare static packages, I'll update in `Makefile` 👍

khos2ow:
@hasheddan I believe you missed this section in `.golangci.yml`:

```yaml
...
  skip-dirs:
  - "pkg/client/clientset"
...
```

hasheddan:
Thanks @khos2ow! Surprised that it didn't pop up with any failures :sweat_smile: Since it didn't I'll just remove the skip rather than updating it :+1:

khos2ow:
That's true, and the fact that those code are being generated there's a good chance that there won't be any warning/error anyways. But I think it's still safe to keep them around to prevent any random surprises. For instance [this](https://github.com/hasheddan/crossplane/blob/internalized/internal/client/clientset/versioned/scheme/register.go#L33) doesn't have a GoDoc and technically a violation (being ignored by golangci).

negz (approved):

PR #2001, opened 2020-12-04
Remove (most) v1alpha1 APIs

negz:

This PR removes all v1alpha1 APIs that have now reached v1beta1. The two exceptions are the ControllerConfiguration and Lock types in pkg.crossplane.io, which continue to be considered alpha features.

I have:

I followed the getting started with GCP guide to composition, and ensured everything worked as expected. I also double checked the Crossplane and RBAC manager pods had no errors in their logs, and that I could install a configuration using crank built from this PR.


apis/pkg/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -21,162 +21,10 @@ limitations under the License.
 package v1alpha1
 
 import (
-	commonv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
 	"k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/runtime"
+	runtime "k8s.io/apimachinery/pkg/runtime"
```
hasheddan, line 25: not a big deal, but this was a little bit unexpected since I don't think it is aliased elsewhere and the controller-gen version didn't change

internal/xcrd/schemas.go
```diff
@@ -230,7 +230,7 @@ func CompositeResourceClaimPrinterColumns() []extv1.CustomResourceColumnDefiniti
 		},
 		{
 			Name:     "AGE",
-			Type:     "Date",
+			Type:     "date",
```
hasheddan, line 233: wasn't sure why this changed but I assume you know what is happening :)

hasheddan (approved):
Thanks @negz! I left two comments below, but they are non-critical. Feel free to merge when you are ready

internal/xcrd/schemas.go
negz: See https://github.com/crossplane/crossplane/pull/2001/commits/7150d363326736a44fb8291aec40f2d12700308b

apis/pkg/v1alpha1/zz_generated.deepcopy.go
negz: Yeah, I have no idea why this changed but it seems relatively innocuous.

PR #2004, opened 2020-12-04
Add dependencies and constraints to docs configurations

hasheddan:

Updates configuration packages used in docs to include crossplane
constraints and provider dependencies.

I have:

Manual build, push, install. Note that in the current documentation the providers will already be installed, so we just check that the present version is valid :+1: 

### DIFF
--- a/docs/snippets/package/alibaba/crossplane.yaml
+++ b/docs/snippets/package/alibaba/crossplane.yaml
@@ -5,3 +5,9 @@ metadata:
   annotations:
     guide: quickstart
     provider: alibaba
+spec:
+  crossplane:
+    version: ">=v0.14.0-0"
+  dependsOn:
+    - provider: crossplane/provider-alibaba
+      version: ">=v0.4.0-0"
--- a/docs/snippets/package/aws-with-vpc/crossplane.yaml
+++ b/docs/snippets/package/aws-with-vpc/crossplane.yaml
@@ -6,3 +6,9 @@ metadata:
     guide: quickstart
     provider: aws
     vpc: new
+spec:
+  crossplane:
+    version: ">=v0.14.0-0"
+  dependsOn:
+    - provider: crossplane/provider-aws
+      version: ">=v0.14.0-0"
--- a/docs/snippets/package/aws/crossplane.yaml
+++ b/docs/snippets/package/aws/crossplane.yaml
@@ -6,3 +6,9 @@ metadata:
     guide: quickstart
     provider: aws
     vpc: default
+spec:
+  crossplane:
+    version: ">=v0.14.0-0"
+  dependsOn:
+    - provider: crossplane/provider-aws
+      version: ">=v0.14.0-0"
--- a/docs/snippets/package/azure/crossplane.yaml
+++ b/docs/snippets/package/azure/crossplane.yaml
@@ -5,3 +5,9 @@ metadata:
   annotations:
     guide: quickstart
     provider: azure
+spec:
+  crossplane:
+    version: ">=v0.14.0-0"
+  dependsOn:
+    - provider: crossplane/provider-azure
+      version: ">=v0.13.0-0"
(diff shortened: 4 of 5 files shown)


docs/snippets/package/aws-with-vpc/crossplane.yaml
```diff
@@ -6,3 +6,9 @@ metadata:
     guide: quickstart
     provider: aws
     vpc: new
+spec:
+  crossplane:
+    version: ">=v0.14.0-0"
+  dependsOn:
+    - provider: crossplane/provider-aws
+      version: ">=v0.14.0-0"
```
jbw976, line 14: should these AWS ones be 0.15?

docs/snippets/package/alibaba/crossplane.yaml
```diff
@@ -5,3 +5,9 @@ metadata:
   annotations:
     guide: quickstart
     provider: alibaba
+spec:
+  crossplane:
+    version: ">=v0.14.0-0"
```
jbw976, line 10: is the expectation that we'll update these in the release branch for each release?

hasheddan: @jbw976 no I don't think so, only when these packages themselves get updated. This constraint will use the latest version that is above v0.14.0, so we will roll forward to any new releases.

docs/snippets/package/aws-with-vpc/crossplane.yaml
hasheddan: They could be, but I don't these packages explicitly depend on anything in v0.15.0, but they will use the latest version if aws isn't already installed. So keeping the constraint here makes us compatible with older versions, while defaulting to the newest.

docs/snippets/package/alibaba/crossplane.yaml
hasheddan: It might make sense to pin them to v1.0.0 when it is available, but they would functionally operate the same

jbw976: does using the trailing `-0` pre-release identifier in the constraint imply that we'd pick up the latest pre-release? if so, that would make sense in master, but perhaps not from the release branch docs.

hasheddan: hmm I think it would be good on release branches as well. The release branches have the documentation for both installing Crossplane from stable or from master. It is likely that folks will be installing stable, but if they do want to try from master I would prefer that work as well. What do you think?

jbw976: i think that makes sense, i just wasn't sure how the dependency resolution aspect of this works...so say someone is following the release branch documentation and they haven't yet installed provider-alibaba yet when they install this snippet.  would dependency resolution give them the latest bleeding edge build from master, or would it give them the latest official release or provider-alibaba?

hasheddan: @jbw976 so that is switching gears a bit because the constraint highlighted here is for `crossplane` version, rather than the dependencies. I think `crossplane` constraint should keep the prerelease `-0`. For the dependencies, there is a trade-off:
- Typically folks will have already installed the providers when going through the docs sequentially (at least for now, maybe we will update to just having them auto-installed by dep resolution in the future). This means that they will have chosen their version, typically the latest stable. The version constraints allowing a pre-release for the dependency will work with this installed stable release, but also allow for other non-stable versions if someone wanted to test them out.
- If dependency resolution is used, we will pull the in the latest from master for the provider, which could not be the experience folks want.

For now, while we are not using dependency resolution, it seems like it may be optimal to allow pre-release versions. We could potentially update versions on the release branch as you mentioned to just not include `-0`.

jbw976: Got it, OK that thinking makes sense to me.  Thank you for explaining Dan!

jbw976 (approved):

hasheddan:
Holding this one for #2009

PR #2005, opened 2020-12-04
Ensure valid DNS label for FriendlyID 

hasheddan:

Updates FriendlyID package revision naming function to always return a
valid DNS label made up of a concatenation of the package name and the
revision digest. This will not result in changes for the typical naming
scenario, but can guard against issues when a revision image has been
manually loaded into the cache with an unexpected name.

I have:

This fixes the case we run into here: https://github.com/crossplane-contrib/provider-in-cluster/pull/9.. but I also just think its cleaner in gernal to always produce a valid DNS label as opposed to _usually_ producing a valid one.

### DIFF
--- a/internal/xpkg/name_test.go
+++ b/internal/xpkg/name_test.go
@@ -73,6 +73,14 @@ func TestFriendlyID(t *testing.T) {
 			},
 			want: "provider-aws-plusabunchofothernonsensethatisgoingt-1234-5678912",
 		},
+		"DigestIsName": {
+			reason: "A valid DNS label should be returned when package digest is a name.",
+			args: args{
+				pkg:  "provider-in-cluster",
+				hash: "provider-in-cluster",
+			},
+			want: "provider-in-cluster-provider-in",
+		},
 	}
 
 	for name, tc := range cases {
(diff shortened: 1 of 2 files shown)


hasheddan:
cc @krishchow

internal/xpkg/name_test.go
```diff
@@ -73,6 +73,14 @@ func TestFriendlyID(t *testing.T) {
...
 			want: "provider-aws-plusabunchofothernonsensethatisgoingt-1234-5678912",
 		},
+		"DigestIsName": {
+			reason: "A valid DNS label should be returned when package digest is a name.",
+			args: args{
+				pkg:  "provider-in-cluster",
+				hash: "provider-in-cluster",
+			},
+			want: "provider-in-cluster-provider-in",
+		},
```
kasey, line 83: presumably `ToDNSLabel` handles the replace `.`->`-` now?

kasey: if so lgtm. fwiw if that was shown in the included test this would be an even more ez +1 👍

kasey (approved):

internal/xpkg/name_test.go
hasheddan: @kasey test above this one already demonstrates :)

kasey: Fair enough! LGTM

negz (approved):

Issue #1846, opened 2020-10-14
Test CPU and memory usage of Crossplane and providers


muvaf:
### What problem are you facing?

If I want to deploy Crossplane with some resource limits, I don't have much information other than guessing what it needs.

### How could Crossplane help solve your problem?

We can test Crossplane & providers for resource usage and write down what affects the usage (how much a new XRD increases CPU usage?) so that users can reliably deploy it.

hasheddan:
xref: https://github.com/crossplane/crossplane/issues/314

jbw976:
I think it's a really good idea to understand our resource usage needs, especially with memory consumption.  We've seen some provider pods (e.g. provider-aws) get OOM killed recently, so I'm wondering if memory usage has increased in recent releases.

We should at least do a [memory profile](https://golang.org/pkg/runtime/pprof/#hdr-Profiling_a_Go_program) to understand what our memory usage/needs are, and to see if there is any low hanging fruit of obvious fixes that we should put in as part of our v1.0.0 hardening efforts to be less resource intensive.

/cc @prasek @lukeweber

khos2ow:
@jbw976 In #1940 we're enabling metrics by tapping into metrics server `controller-runtime` provides. There's also a way to extend that and add customized metrics (which I don't remember off the top of my head), but at the bare minimum some good handful of metrics are available (can be found [here](https://gist.github.com/khos2ow/6ab83bd0bb9b0823935bd8bf8998850d)). That being said even proceeding with #1940 doesn't enable metrics gathering from providers, but it's gonna be very simple to add that (essentially a port definition and couple of annotation is needed).

On the other hand, neither of these will cover memory profiling, which I've done previously for some other projects and I can either squeeze it in the same PR or in a new one. Something like exposing additional `/debug/pprof/` endpoint to deployments.

muvaf:
https://github.com/crossplane/crossplane/pull/2983 will tackle the provider part of this by providing a tool to run standardized tests against common metrics.

PR #2007, opened 2020-12-04
Enable memory profiling in debug mode

khos2ow:

When in debug mode, start a standalone HTTP server on port 9090
dedicated for Memory Profiling. It exposes the following endpoints:

- /debug/pprof/
- /debug/pprof/cmdline
- /debug/pprof/profile
- /debug/pprof/symbol
- /debug/pprof/trace
- /debug/pprof/goroutine
- /debug/pprof/heap
- /debug/pprof/threadcreate
- /debug/pprof/block

This PR is blocked by crossplane/crossplane-runtime#233
Partial fix for #1846 

I have:

- run `./cluster/local/kind.sh up`
- run `HELM3_FLAGS="--set debug=true" ./cluster/local/kind.sh helm-install`
- run `kubectl -n crossplane-system port-forward CROSSPLANE_POD 9090`
- open above endpoints in the browser: e.g. http://localhost:9090/debug/pprof/symbol

### DIFF
--- a/cluster/charts/crossplane/README.md
+++ b/cluster/charts/crossplane/README.md
@@ -63,6 +63,7 @@ and their default values.
 
 | Parameter | Description | Default |
 | --- | --- | --- |
+| `debug` | Enable debug mode for Crossplane and RBAC Manager | `false` |
 | `image.repository` | Image | `crossplane/crossplane` |
 | `image.tag` | Image tag | `master` |
 | `image.pullPolicy` | Image pull policy | `Always` |
--- a/cluster/charts/crossplane/templates/deployment.yaml
+++ b/cluster/charts/crossplane/templates/deployment.yaml
@@ -33,11 +33,14 @@ spec:
       serviceAccountName: {{ template "name" . }}
       containers:
       - image: {{ .Values.image.repository }}:{{ .Values.image.tag }}
-        {{- if .Values.args }}
+        {{- if or (.Values.args) (.Values.debug) }}
         args:
+        {{- end }}
+        {{- if or .Values.debug }}
+        - --debug
+        {{- end }}
         {{- range $arg := .Values.args }}
         - {{ $arg }}
-        {{- end }}
         {{- end }}
         imagePullPolicy: {{ .Values.image.pullPolicy }}
         name: {{ .Chart.Name }}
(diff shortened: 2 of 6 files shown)


khos2ow:
The general problem with this is that it doesn't take providers into consideration and this functionality needs to be enabled, individually, for each provider. I would prefer to move this to `crossplane-runtime` to be shared among projects, but I wanted to get the discussion going here before start doing that.

internal/pprof/pprof.go
```diff
@@ -0,0 +1,92 @@
...
+func (s *server) Start() {
+	s.router.Handle("/debug/pprof/", http.HandlerFunc(pprof.Index))
+	s.router.Handle("/debug/pprof/cmdline", http.HandlerFunc(pprof.Cmdline))
+	s.router.Handle("/debug/pprof/profile", http.HandlerFunc(pprof.Profile))
+	s.router.Handle("/debug/pprof/symbol", http.HandlerFunc(pprof.Symbol))
+	s.router.Handle("/debug/pprof/trace", http.HandlerFunc(pprof.Trace))
+	s.router.Handle("/debug/pprof/goroutine", pprof.Handler("goroutine"))
+	s.router.Handle("/debug/pprof/heap", pprof.Handler("heap"))
+	s.router.Handle("/debug/pprof/threadcreate", pprof.Handler("threadcreate"))
+	s.router.Handle("/debug/pprof/block", pprof.Handler("block"))
```
hasheddan, line 67: Any reason not to use the default mux? It will expose each of these for you if you do an anonymous pprof import (ref: https://golang.org/pkg/net/http/pprof/)

hasheddan (commented):
I really like this @khos2ow! Thanks for adding this :) One question below.

internal/pprof/pprof.go
khos2ow: The reason was essentially twofold 1) spin up a dedicated debug server on port `9090` and 2) pass additional profilers which don't exist in [`init()` of that package](https://golang.org/src/net/http/pprof/pprof.go#L79).

khos2ow: And also what do you think of moving this to `crossplane-runtime` to be able to DRY it with providers as well (if they choose to utilize it)?

hasheddan: @khos2ow adding this to crossplane-runtime feels pretty reasonable to me :)

khos2ow: @hasheddan I've opened up https://github.com/crossplane/crossplane-runtime/pull/233 and after that gets merged we can proceed with this one with a mandatory change to `go.mod`.

negz:
This has been sitting open for a long time now and we haven't moved on it so I'm going to close it out. We appreciate the idea nonetheless!

PR #2009, opened 2020-12-04
Do not default registry name when resolving dependencies

### LINKED ISSUE #2008: Dependencies are not found if provided without registry
### What happened?

`go-containerregistry` defaults to `index.docker.io` as registry if none is provided. We account for this when a package identifies itself in the `Lock`, but not for identifying dependencies. This means that having a dependency on something like `crossplane/provider-aws` will always fail to find a matching dependency, but `registry.upbound.io/crossplane/provider-aws` or `index.docker.io/crossplane/provider-aws` will work.

The simplest solution to this is to update the dependency searching to use registry defaulting. However, this will break functionality in the future if we change to a different default. Alternatively, we could have packages add themselves to the `Lock` without the default registry added. (i.e. if I install `crossplane/provider-aws` it would be added to the `Lock` as `crossplane/provider-aws` rather than `<default-registry>/crossplane/provider-aws`).

Regardless of the direction we choose, we should always encourage using full registry paths rather than falling back on a default.

### How can we reproduce it?

1. Install `crossplane`
2. `kubectl crossplane install hasheddan/getting-started-with-aws:v0.1.0`
3. See that `provider-aws` is installed successfully, but `getting-started-with-aws` never becomes healthy and reports missing dependency

```
🤖 (crossplane) k get pkg
NAME                                                       INSTALLED   HEALTHY   PACKAGE                                     AGE
configuration.pkg.crossplane.io/getting-started-with-aws   True        Unknown   hasheddan/getting-started-with-aws:v0.1.0   49s

NAME                                                 INSTALLED   HEALTHY   PACKAGE                                             AGE
provider.pkg.crossplane.io/crossplane-provider-aws   True        True      crossplane/provider-aws:v0.16.0-rc.0.57.g641be176   40s
```

```
Events:
  Type     Reason               Age                From                                              Message
  ----     ------               ----               ----                                              -------
  Warning  ResolveDependencies  29s (x4 over 61s)  packages/configurationrevision.pkg.crossplane.io  missing dependencies: [crossplane/provider-aws]
```

4. Observe packages in `Lock`

```
packages:
  - dependencies:
    - constraints: '>=v0.14.0-0'
      package: crossplane/provider-aws
      type: Provider
    name: getting-started-with-aws-b269422271fa
    source: index.docker.io/hasheddan/getting-started-with-aws
    type: Configuration
    version: v0.1.0
  - dependencies: []
    name: crossplane-provider-aws-9d325be8ac27
    source: index.docker.io/crossplane/provider-aws
    type: Provider
    version: v0.16.0-rc.0.57.g641be176

```

Crossplane version: `1.0.0-rc.0.81.gdce235c8`

hasheddan:

This updates the logic that adds package revisions to the Lock to avoid
defaulting to a registry. Registry defaulting still takes place when
actually pulling an image, but it remains opaque to users. All
dependencies and revisions will be added to the Lock in the same format
they were declared.

Fixes #2008 

I have:

Testing steps used #2008 now produce expected results:

```
🤖 (crossplane) k get pkg
NAME                                                       INSTALLED   HEALTHY   PACKAGE                                     AGE
configuration.pkg.crossplane.io/getting-started-with-aws   True        True      hasheddan/getting-started-with-aws:v0.1.0   84s

NAME                                                 INSTALLED   HEALTHY   PACKAGE                                             AGE
provider.pkg.crossplane.io/crossplane-provider-aws   True        True      crossplane/provider-aws:v0.16.0-rc.0.57.g641be176   73s
```

```
  packages:
  - dependencies:
    - constraints: '>=v0.14.0-0'
      package: crossplane/provider-aws
      type: Provider
    name: getting-started-with-aws-b269422271fa
    source: hasheddan/getting-started-with-aws
    type: Configuration
    version: v0.1.0
  - dependencies: []
    name: crossplane-provider-aws-9d325be8ac27
    source: crossplane/provider-aws
    type: Provider
    version: v0.16.0-rc.0.57.g641be176
```


internal/controller/pkg/revision/dependency.go
```diff
@@ -205,7 +202,7 @@ func (m *PackageDependencyManager) Resolve(ctx context.Context, pkg runtime.Obje
 
 // RemoveSelf removes a package from the lock.
 func (m *PackageDependencyManager) RemoveSelf(ctx context.Context, pr v1beta1.PackageRevision) error {
-	prRef, err := name.ParseReference(pr.GetSource())
+	prRef, err := name.ParseReference(pr.GetSource(), name.WithDefaultRegistry(""))
```
kasey, line 205: My understanding of this change is that, were this option not specified, `go-containerregistry` will default to `index.docker.io` (src: https://github.com/google/go-containerregistry/blob/master/pkg/name/options.go#L37)

So we are effectively deferring the job of picking a default domain for an unqualified image tag to the code that ultimately does the pull, which I think in this case is going to be the package manager. What I don't know without looking deeper is what part of the package management system or libraries will pick the domain to use in the request?

hasheddan: There are three places where we do this:
1. Getting the digest of the current package image: https://github.com/crossplane/crossplane/blob/dce235c83c2990bb8a521c86223efd07c63001a2/internal/controller/pkg/manager/revisioner.go#L62
2. Actually downloading the package into the cache: https://github.com/crossplane/crossplane/blob/dce235c83c2990bb8a521c86223efd07c63001a2/internal/controller/pkg/revision/imageback.go#L78
3. Getting all tags for a dependency: https://github.com/crossplane/crossplane/blob/dce235c83c2990bb8a521c86223efd07c63001a2/internal/controller/pkg/resolver/reconciler.go#L220

kasey: SGTM

kasey (approved):

negz (approved):

PR #2011, opened 2020-12-05
Fix CRD Required Schema

### LINKED ISSUE #2003: Generated CRDs are Missing the `required` Field in the Structural Schema
### What happened?

I just hit a small issue with `openAPIV3Schema` for XRDs: In the code we only copy the `.Properties` field of the `spec` schema to the final CRD schema.
This makes it impossible to make properties of the `spec` schema required because the `.Required` field at this level is ignored.
Also the `spec` field itself is not required and could therefore be completely missing.

Same goes for the generated Claim CRDs.

https://github.com/crossplane/crossplane/blob/master/internal/xcrd/crd.go#L94
https://github.com/crossplane/crossplane/blob/master/internal/xcrd/crd.go#L81

### How can we reproduce it?

Using the XRD example from the [docs](https://crossplane.io/docs/v0.14/introduction/composition.html#define-your-composite-resource), the generated CRD won't have the fields `.spec` and `.spec.parameters` marked as required, even though the schema in the XRD does so.

Crossplane version: v0.14.0

srueg:

Fixes #2003

I have:

`make test`

### DIFF
--- a/internal/xcrd/crd.go
+++ b/internal/xcrd/crd.go
@@ -83,16 +83,19 @@ func ForCompositeResource(xrd *v1beta1.CompositeResourceDefinition) (*extv1.Cust
 			},
 		}
 
-		p, err := getSpecProps(vr.Schema)
+		p, required, err := getSpecProps(vr.Schema)
 		if err != nil {
 			return nil, errors.Wrap(err, errGetSpecProps)
 		}
+		specProps := crd.Spec.Versions[i].Schema.OpenAPIV3Schema.Properties["spec"]
+		specProps.Required = append(specProps.Required, required...)
 		for k, v := range p {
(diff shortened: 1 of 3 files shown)


negz:
Thanks @srueg! Would you mind doing a little manual testing on this one; i.e. make sure you can create a composite resource claim and composite resource with and without required fields?

negz (approved):
This looks good to me - happy to merge once you ack that it works fine when tested manually.

srueg:
Alright, I did a quick test based on the examples (XRD, XR & XRC) in the [docs](https://crossplane.io/docs/v0.14/introduction/composition.html) and it all worked as expected. 🎉

hasheddan (approved):

negz:
@srueg Thanks! Looks like this needs a rebase - once that's done I'll merge.

Issue #1767, opened 2020-09-24
Allow Provider packages to request additional RBAC permissions


negz:
### What problem are you facing?

Per the [package manager design](https://github.com/crossplane/crossplane/blob/18324aa15a191bbb1a71a0161af5edd9f7bddfe6/design/design-doc-packages-v2.md) we expect that some providers may need additional RBAC permissions. The design proposes that providers be able to specify an array RBAC rules as `spec.permissions` in their package metadata, which would be surfaced as `spec.permissions` on a `ProviderRevision`. We've since decided this field should be named `permissionRequests`.

https://github.com/crossplane/crossplane/pull/1616

### How could Crossplane help solve your problem?

Crossplane should support and respect the `permissionRequests` of a `ProviderRevision`. This would likely be implemented by the RBAC manager, which presumably would need some way for the operator to configure which permissions a provider was and was not allowed to request.

PR #2015, opened 2020-12-09
Allow provider packages to request extra RBAC permissions

negz:

Fixes https://github.com/crossplane/crossplane/issues/1767

This PR allows Providers to include permission requests (in the form of an optional array of RBAC rules) in their `crossplane.yaml` file. These requests are propagated by the package manager to the ProviderRevision. The RBAC manager rejects all permission requests by default, but it can be configured to assess permission requests against a named ClusterRole - providers will be allowed to request any permission enumerated by the ClusterRole.

I have:

See https://github.com/crossplane/crossplane/pull/2015#issuecomment-742208049.


internal/controller/pkg/revision/hook.go
```diff
@@ -77,6 +80,18 @@ func (h *ProviderHooks) Pre(ctx context.Context, pkg runtime.Object, pr v1beta1.
...
+	if !ok {
+		return errors.New(errNotProviderRevision)
+	}
+
+	if !cmp.Equal(provRev.Spec.PermissionRequests, pkgProvider.Spec.Controller.PermissionRequests) {
+		provRev.Spec.PermissionRequests = pkgProvider.Spec.Controller.PermissionRequests
+		if err := h.client.Update(ctx, provRev); err != nil {
+			return errors.Wrap(err, errUpdateProviderRevision)
+		}
+	}
```
negz, line 93: @hasheddan I don't love that I update the ProviderRevision post-create to propagate the requests rather than setting it at create time. Is there a better option here?

internal/controller/rbac/provider/roles/requests.go
```diff
@@ -0,0 +1,198 @@
...
+			if c.allowed {
+				return true
+			}
+			if c.Allowed(p[1:]) {
+				return true
+			}
+		}
+	}
+	return false
+}
```
negz, line 90: Credit to @kasey for sketching out this tree implementation - does this look sound to you?

internal/controller/rbac/provider/roles/roles.go
```diff
@@ -53,19 +49,6 @@ var (
...
-// RBAC manager to be configured such that it knows which rules may and may not
-// be requested.
-var rulesSystemExtra = []rbacv1.PolicyRule{
-	{
-		APIGroups: []string{""},
-		Resources: []string{pluralSecrets, pluralConfigmaps, pluralEvents},
-		Verbs:     verbsEdit,
-	},
-}
-
```
negz, line 68: Note that we no longer grant providers any extra rules by default; they get only the things they define. In practice pretty much all providers will need these rules. I was thinking we should update all providers packages to request them, though it occurs to me that would be a breaking change for anyone who didn't update their provider packages.

cmd/crossplane/rbac/rbac.go
```diff
@@ -38,17 +38,19 @@ const (
...
+	ManagementPolicy    string
+	ProviderClusterRole string
 }
 
 // FromKingpin produces the RBAC manager command from a Kingpin command.
 func FromKingpin(cmd *kingpin.CmdClause) *Command {
 	c := &Command{Name: cmd.FullCommand()}
 	cmd.Flag("sync", "Controller manager sync period duration such as 300ms, 1.5h or 2h45m").Short('s').Default("1h").DurationVar(&c.Sync)
 	cmd.Flag("manage", "RBAC management policy.").Short('m').Default(ManagementPolicyAll).EnumVar(&c.ManagementPolicy, ManagementPolicyAll, ManagementPolicyBasic)
+	cmd.Flag("provider-clusterrole", "A ClusterRole enumerating the permissions provider packages may request.").StringVar(&c.ProviderClusterRole)
```
negz, line 53: This is unset by default and thus all permission requests will be rejected. I'm thinking we should create a default (e.g. allowing secrets, events, and configmaps) via the Helm chart and pass that in here.

internal/controller/pkg/revision/hook.go
hasheddan: We _could_ move pull logic to the package controller. That would be a pretty large change, and we would still need to access the package contents in the revision controller (this would be fine, the package controller and the revision controller would just share the cache), but it would allow the package manager to set revision fields that are specific to package contents at creation time. The original motivation for not doing this was just a separation of concern for the package manager to only worry about the lifecycle of revisions and for the revision itself to act in its own interest and have all "revision-specific" operations take place in its controller.

We could choose to annotate / label the service account we create with any additional permissions that we are requesting from the RBAC manager. Any thoughts on that?

apis/pkg/v1beta1/revision_types.go
```diff
@@ -78,6 +79,13 @@ type PackageRevisionSpec struct {
 	// +optional
 	// +kubebuilder:default=false
 	SkipDependencyResolution *bool `json:"skipDependencyResolution,omitempty"`
+
+	// PermissionRequests for RBAC rules required for this provider's controller
+	// to function. The RBAC manager is responsible for granting requested
+	// permissions. It uses a whitelist to determine what permissions a provider
+	// may be granted.
+	// +optional
+	PermissionRequests []rbacv1.PolicyRule `json:"permissionRequests,omitempty"`
```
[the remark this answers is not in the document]
hasheddan: Also see my comment below about potentially annotating / labelling the SA we create with additional permission requests.

internal/controller/rbac/provider/roles/reconciler.go
```diff
@@ -198,6 +250,28 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error)
...
+		log.Debug(errRejectedPermission, "rule", rule)
+		r.record.Event(pr, event.Warning(reasonApplyRoles, errors.Errorf("%s %s", errRejectedPermission, rule)))
+	}
+
+	// We return early and don't grant _any_ RBAC permissions if we would reject
+	// any requested permission. It's better for the provider to be completely
+	// and obviously broken than for it to be subtly broken in a way that may
+	// not surface immediately, i.e. due to missing an RBAC permission it only
+	// occasionally needs. There's no need to requeue - the revisions requests
+	// won't change, and we're watching the ClusterRole of allowed requests.
```
hasheddan, line 270: Somewhat related: how would you feel about us updating providers to panic if types are missing or we don't have RBAC? It can be confusing when the pods are running normally but in reality they can't perform any operations.

hasheddan (commented):
Thanks for this work @negz! Implementation looks pretty solid. I dropped some comments with thoughts around how we might make permission requests in a cleaner manner. Let me know what you think :)

internal/controller/pkg/revision/hook.go
negz: > We could move pull logic to the package controller.

I think I prefer this implementation to that one - I can convince myself we're "late intializing" the revision. :)

> We could choose to annotate / label the service account we create with any additional permissions that we are requesting from the RBAC manager. Any thoughts on that?

I prefer the revision to this alternative too - especially because it would be tricky to add RBAC rules as unstructured labels/annotations.

hasheddan: > especially because it would be tricky to add RBAC rules as unstructured labels/annotations

Do you think marshaling the struct and adding it as an annotation, then unmarshaling in the RBAC manager would be error prone?

negz: It's not so much that I think it would be error prone so much as I don't see the advantage of doing so, especially when we can pass that data using a structured, schemafied object (i.e. the ProviderRevision's spec) instead.

internal/controller/rbac/provider/roles/roles.go
negz: Reverted this in a subsequent commit - I think we need to give more notice if we decide to remove these extra rules.

internal/controller/rbac/provider/roles/reconciler.go
negz: Yes, we should refuse to run if we know we can't run properly. I thought this was the case already when CRDs were missing, but I know we don't crash for missing RBAC permissions.

apis/pkg/v1beta1/revision_types.go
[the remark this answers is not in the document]
negz: I didn't think much about this - the spec just seemed to be the intuitive place given that requesting permissions seems to be a form of expressing desired state.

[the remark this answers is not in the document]
hasheddan: Yeah I see what you are saying. I think my rationale is that if something like a `Pod` has a container image in the spec, when it pulls the image and runs it the digest and other information about that image is put in the status because pulling that image and running it is a result of operations performed based on the desired spec. This feels kind of like that, the difference here is just that we have another part of the system explicitly depending on a status field which feels weird.

internal/controller/pkg/revision/hook.go
hasheddan: Gotcha 👍 I guess its more about that I feel like it doesn't belong in the spec, but given that it isn't intended to be a user-facing object then maybe it is fine.

apis/pkg/v1beta1/revision_types.go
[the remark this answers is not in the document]
negz: To me the issue isn't so much whether another part of the system depends on this information, but rather whether we consider the permission requests to be an input to how the provider revision is reconciled (i.e. its desired state), or an output of how it was reconciled (i.e. observed state). This isn't so clear cut in this case.

My thinking was that the creation of Deployments, CRDs, RBAC roles, etc happen in response to the existence of the ProviderRevision. The ProviderRevision is thus responsible for declaring the desired state of those things and thus the permission requests fit in its spec.

However another way of looking at it - which I think probably fits better with how ProviderRevisions work today - is that a ProviderRevision is basically a pointer to something else (the provider package) that captures the desired state. So the relationship isn't `package tarball -> revision -> CRDs etc` but rather `revision -> package tarball -> CRDs etc`. Under this view I can see how the revisions desired state is "go unpack this package and do what it says", and thus "the package said to do this" is fair game as status.

TL;DR Putting this in status still feels kinda weird to me, but I think on reflection you're right that it does fit better with the current implementation to put it there. I forget - would we recover from lost status here? i.e. Would we unpack the package again and recreate the status if it were lost?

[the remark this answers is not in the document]
hasheddan: @negz yep! we check the package contents on every reconcile because we must guarantee installed objects still match.

negz:
I've run some tests and confirmed that:

When running without the `--provider-clusterrole` flag:
* I can install a provider with no permission requests without issue.
* The RBAC manager refuses to grant any requested permissions.

When running with the `--provider-clusterrole` flag:
* I can install a provider with no permission requests without issue.
* The RBAC manager refuses to grant any permissions that aren't in the cluster role.
* Updating the cluster role to allow the permissions causes the RBAC manager to grant them.


```console
$ kubectl describe clusterrole crossplane:provider:provider-gcp-8a7e46df4367:system
Name:         crossplane:provider:provider-gcp-8a7e46df4367:system
Labels:       <none>
Annotations:  <none>
PolicyRule:
  Resources   Non-Resource URLs  Resource Names  Verbs
  ---------   -----------------  --------------  -----
  configmaps  []                 []              [*]
  events      []                 []              [*]
  secrets     []                 []              [*]
```

```console
Events:
  Type     Reason             Age                 From                                         Message
  ----     ------             ----                ----                                         -------
  Normal   ApplyClusterRoles  23m (x3 over 23m)   rbac/providerrevision.pkg.crossplane.io      Applied RBAC ClusterRoles
  Warning  SyncPackage        23m                 packages/providerrevision.pkg.crossplane.io  cannot run post establish hook for package: provider package deployment is unavailable: Deployment does not have minimum availability.
  Warning  ApplyClusterRoles  23m (x4 over 23m)   rbac/providerrevision.pkg.crossplane.io      refusing to apply any RBAC roles due to request for disallowed permission {apiGroup: "", resource: "deployments", resourceName: "*", verb: "get"}
  Warning  ApplyClusterRoles  23m (x4 over 23m)   rbac/providerrevision.pkg.crossplane.io      refusing to apply any RBAC roles due to request for disallowed permission {apiGroup: "", resource: "deployments", resourceName: "*", verb: "list"}
  Warning  ApplyClusterRoles  23m (x4 over 23m)   rbac/providerrevision.pkg.crossplane.io      refusing to apply any RBAC roles due to request for disallowed permission {apiGroup: "", resource: "deployments", resourceName: "*", verb: "watch"}
  Normal   BindClusterRole    23m (x8 over 23m)   rbac/providerrevision.pkg.crossplane.io      Bound system ClusterRole to provider ServiceAccount(s)
  Normal   SyncPackage        35s (x26 over 23m)  packages/providerrevision.pkg.crossplane.io  Successfully configured package revision
```

It is a little misleading that if you look at the events you see "Applied RBAC ClusterRoles", followed by several refusals to apply any ClusterRoles. I believe this is because a few reconciles happen while the ProviderRevision is still being formed; i.e. as its status is updated with the objects it owns and its spec is updated with its permission requests. I'm not sure how worthwhile it is to try to improve this; if we do I'd prefer to do so in a future PR.

apis/pkg/v1beta1/revision_types.go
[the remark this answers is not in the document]
negz: Okay, I pushed a commit that moves these to the status.

internal/controller/pkg/revision/hook.go
negz: We set the requests in the status now, which feels like a better fit in this context.

internal/controller/rbac/provider/roles/requests.go
```diff
@@ -0,0 +1,198 @@
...
+				for _, v := range r.Verbs {
+					out = append(out, Rule{NonResourceURL: u, Verb: v})
+				}
+			}
+			continue
+		}
+
+		names := r.ResourceNames
+		if len(names) < 1 {
+			names = []string{"*"}
```
hasheddan, line 142: nitpick: could use `wildcard` const here

hasheddan (approved):
LGTM :heavy_check_mark: thanks for this @negz!

kasey (approved):

internal/controller/rbac/provider/roles/requests.go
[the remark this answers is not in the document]
negz: I don't follow. Do you mean the comment on `PolicyRule.ResourceNames`?

internal/controller/rbac/provider/roles/requests.go
negz: Yeah I was on the fence about that too, but was too lazy to break this into a public method wrapping a private method for recursion. Arguably this whole implementation should be private.

internal/controller/rbac/provider/roles/reconciler.go
```diff
@@ -44,16 +45,36 @@ const (
...
 	reasonApplyRoles event.Reason = "ApplyClusterRoles"
 )
 
+// A PermissionRequestsValidator validates requested RBAC rules.
+type PermissionRequestsValidator interface {
+	// ValidatePermissionRequests validates the supplied slice of RBAC rules. It
+	// returns a slice of any rejected (i.e. disallowed) rules. It returns an
+	// error if it is unable to validate permission requests.
+	ValidatePermissionRequests(ctx context.Context, requested ...rbacv1.PolicyRule) ([]Rule, error)
+}
```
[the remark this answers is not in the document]
kasey: It was a generic golang design question, as an aside, definitely not feedback that should impact whether this gets merged. Why is `PermissionRequestsValidator` an interface, and not just a callback fn?

internal/controller/rbac/provider/roles/requests.go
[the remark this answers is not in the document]
kasey: yes

internal/controller/rbac/provider/roles/reconciler.go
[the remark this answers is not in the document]
negz: Ah, got it. It's probably mostly just a result of what felt more idiomatic to me when we started doing this. In some cases these interfaces have more than one method, or have implementations that could be cleaner as a more complex type (i.e. a struct with some state) rather than an enclosure callback.

internal/controller/rbac/provider/roles/requests.go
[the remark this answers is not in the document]
negz: Added a comment.

internal/controller/rbac/provider/roles/requests.go
```diff
@@ -0,0 +1,198 @@
...
+func Expand(rs ...rbacv1.PolicyRule) []Rule {
+	out := make([]Rule, 0, len(rs))
+	for _, r := range rs {
+		if len(r.NonResourceURLs) > 0 {
+			for _, u := range r.NonResourceURLs {
+				for _, v := range r.Verbs {
+					out = append(out, Rule{NonResourceURL: u, Verb: v})
+				}
+			}
+			continue
```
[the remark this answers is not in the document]
negz: I ended up refactoring to be less opinionated here.

internal/controller/rbac/provider/roles/requests.go
negz: Done!

PR #2022, opened 2020-12-10
Fix panic in crank with missing filters

hasheddan:

FilterFns were being copied incorrectly into the slice that was passed
to the fsReader backend, causing it to call nil functions and panic
paths are ignored. This updates it to correctly copy the default
functions and path skips into the FilterFn slice.

I have:

`kubectl crossplane build configuration --name package.xpkg --ignore "examples/*,hack/*"` with https://github.com/upbound/platform-ref-aws

### DIFF
--- a/cmd/crank/build.go
+++ b/cmd/crank/build.go
@@ -94,12 +94,15 @@ func (c *buildCmd) Run(child *buildChild) error {
 // default build filters skip directories, empty files, and files without YAML
 // extension in addition to any paths specified.
 func buildFilters(root string, skips []string) []parser.FilterFn {
-	opts := make([]parser.FilterFn, len(skips)+3)
-	opts[0] = parser.SkipDirs()
-	opts[1] = parser.SkipNotYAML()
-	opts[2] = parser.SkipEmpty()
+	defaultFns := []parser.FilterFn{
+		parser.SkipDirs(),
+		parser.SkipNotYAML(),
+		parser.SkipEmpty(),
+	}
+	opts := make([]parser.FilterFn, len(skips)+len(defaultFns))
+	copy(opts, defaultFns)
 	for i, s := range skips {
-		opts[i+2] = parser.SkipPath(filepath.Join(root, s))
+		opts[i+len(defaultFns)] = parser.SkipPath(filepath.Join(root, s))
 	}
 	return opts
 }


prasek (approved):
lgtm, thx!

cmd/crank/build.go
```diff
@@ -94,12 +94,15 @@ func (c *buildCmd) Run(child *buildChild) error {
...
+		parser.SkipDirs(),
+		parser.SkipNotYAML(),
+		parser.SkipEmpty(),
+	}
+	opts := make([]parser.FilterFn, len(skips)+len(defaultFns))
+	copy(opts, defaultFns)
 	for i, s := range skips {
-		opts[i+2] = parser.SkipPath(filepath.Join(root, s))
+		opts[i+len(defaultFns)] = parser.SkipPath(filepath.Join(root, s))
 	}
```
negz, line 106:
```go
defaultFns := []parser.FilterFn{
	parser.SkipDirs(),
	parser.SkipNotYAML(),
	parser.SkipEmpty(),
}

skipFns := make([]parser.FilterFn, len(skips))
for i := range skips {
	skipFns[i] = parser.SkipPath(filepath.Join(root, skips[i])
}

return append(defaultFns, skipsFns...)
```

@hasheddan FWIW I found it a little tricky to parse what this was doing. The above approach could be worth considering if you feel it's safer/more readable.

negz (approved):

cmd/crank/build.go
hasheddan: @negz I agree that your implementation looks a bit simpler, but I believe it will cause more overhead in that we create the `defaultFns` slice, the `skipFns` slice, then allocate a new slice for the append and copy all of the elements over. However, you could also say my change in this PR is less optimal than the existing (with the index offset corrected) because it uses the `defaultFn` slice when we could just do assignment as we were doing before. Anyway, I think I'll stick with the update I have here, but feel free to follow up if you feel that the proposed implementation is an improvement :+1: I anticipate refactoring a fair portion of the `crank` code following the v1.0 release.

negz: @hasheddan I'm not particularly concerned about which implementation we go with here, but I want to give the general feedback that given a choice between a safer, more readable implementation and a more efficient, performant implementation we should bias toward the safer, more readable implementation 9 times out of 10.

Safer and more readable is perhaps subjective so don't take this as me asserting that my suggestion was objectively that. Rather, my point is that if you identify a more readable or safer approach I suggest preferring that approach over performance optimisations - at least until we start to notice performance becoming an issue.

hasheddan: @negz thanks for the feedback! I agree with the sentiment and will make sure I am incorporating it across my changes :)

PR #2027, opened 2020-12-11
Adding documentation for how to use Crossplane offered by a service

grantgumina:
Today the documentation in Crossplane describes how to setup the project by installing it into a Kubernetes cluster. Community members can also deploy it using a managed Crossplane service like Upbound Cloud. This update calls out that option for the community, and also makes sure any community member who's built another Crossplane managed service has a mechanism for requesting we highlight their offering as well.

These changes have been reviewed with a markdown editor confirming they render properly.

### DIFF
--- a/docs/getting-started/install-configure.md
+++ b/docs/getting-started/install-configure.md
@@ -144,7 +144,11 @@ curl -sL https://raw.githubusercontent.com/crossplane/crossplane/master/install.
 
 ## Select Provider
 
-Install and configure a provider for Crossplane to use for infrastructure provisioning:
+If you've decided to use a Hosted Crossplane from Upbound Cloud, make sure you've gone ahead and 
+[connected locally](https://cloud.upbound.io/docs/getting-started/connect-to-your-platform) to the remote 
+cluster. 
+
+Now from your terminal, install and configure a provider for Crossplane to use for infrastructure provisioning:
 <ul class="nav nav-tabs">
 <li class="active"><a href="#provider-tab-aws" data-toggle="tab">AWS</a></li>
 <li><a href="#provider-tab-gcp" data-toggle="tab">GCP</a></li>
--- a/docs/getting-started/install-configure.md
+++ b/docs/getting-started/install-configure.md
@@ -31,7 +31,7 @@ Once you've completed thse two steps, skip down to
 
 <i>Want see another hosted Crossplane service listed? Please 
 <a href="http://slack.crossplane.io/">reach out on Slack</a> and our community will 
-highlight it here!</a>
+highlight it here!</a></i>
 
 ## Start with a Self-Hosted Crossplane
 Installing Crossplane into an existing Kubernetes cluster will require a bit more
--- a/docs/getting-started/install-configure.md
+++ b/docs/getting-started/install-configure.md
@@ -1,12 +1,15 @@
 ---
-title: Install & Configure toc: true weight: 2 indent: true
+title: Install & Configure 
+toc: true 
+weight: 2 
+indent: true
 ---
 
 # Choosing Hosted or Self-Hosted Crossplane
 
-Users looking to use Crossplane for the first time have two options avaiable to
+Users looking to use Crossplane for the first time have two options available to
 them today. The first way is to use a hosted Crossplane service like [Upbound
-Cloud]("https://upbound.io"). Alternatively, users looking for some more
+Cloud][Upbound Cloud]. Alternatively, users looking for some more
 flexability can install Crossplane into their own Kubernetes cluster.
 
 Crossplane will be installed using the regularly published Helm chart. The Helm
@@ -421,3 +424,4 @@ kubectl delete namespace crossplane-system
 [Kind]: https://kind.sigs.k8s.io/docs/user/quick-start/ 
 [Crossplane packages]:../introduction/packages.md 
 [Slack]: http://slack.crossplane.io/
+[Upbound Cloud]: https://upbound.io
--- a/docs/getting-started/install-configure.md
+++ b/docs/getting-started/install-configure.md
@@ -5,16 +5,39 @@ weight: 2
 indent: true
 ---
 
-# Install & Configure Crossplane
+# Choosing Hosted vs. Self-Hosted Crossplane
 
-Crossplane can be easily installed into any existing Kubernetes cluster using
-the regularly published Helm chart. The Helm chart contains all the custom
-resources and controllers needed to deploy and configure Crossplane.
+Users looking to use Crossplane for the first time have two options avaiable to them 
+today. The easiest way to start using Crossplane quickly is to use a hosted Crossplane 
+service like Upbound Cloud. Setup should take 10-15 minutes total.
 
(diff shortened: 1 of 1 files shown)


negz:
@grant Looks like you need to sign these commits - see https://github.com/crossplane/crossplane/pull/2027/checks?check_run_id=1539764606.

Also, could you please wrap the markdown at 80 chars to fit the rest of the document.

docs/getting-started/install-configure.md
```diff
@@ -5,16 +5,39 @@ weight: 2
...
+and then [connect](https://cloud.upbound.io/docs/getting-started/connect-to-your-platform) 
+to your hosted Crossplane cluster.
 
-## Get a Kubernetes Cluster
+Once you've completed thse two steps, skip down to 
+[Install Crossplane CLI](#install-crossplane-cli) for further setup instructions.
+
+<i>Want see another hosted Crossplane service listed? Please 
+<a href="http://slack.crossplane.io/">reach out on Slack</a> and our community will 
+highlight it here!</a></i>
```
negz, line 34: I think you can use markdown for this, i.e.

```markdown
Want see another hosted Crossplane service listed? Please [reach out on Slack][Slack] and our community will 
highlight it here!
````

Note that you'll need to add `[Slack]: http://slack.crossplane.io/` to the links at the bottom.

File: docs/getting-started/install-configure.md [hunk 2]
```diff
@@ -5,16 +5,39 @@ weight: 2
...
 
-# Install & Configure Crossplane
+# Choosing Hosted vs. Self-Hosted Crossplane
 
-Crossplane can be easily installed into any existing Kubernetes cluster using
-the regularly published Helm chart. The Helm chart contains all the custom
-resources and controllers needed to deploy and configure Crossplane.
+Users looking to use Crossplane for the first time have two options avaiable to them 
+today. The easiest way to start using Crossplane quickly is to use a hosted Crossplane 
+service like Upbound Cloud. Setup should take 10-15 minutes total.
```
negz, line 12: > Setup should take 10-15 minutes total.

How would you feel about removing this part? Partly because it crosses the line into feeling a bit like it's selling Upbound Cloud, and partly because 10-15 mins is probably about the same as (or longer than) the time most folks with an existing Kubernetes cluster would need to get setup with Crossplane.

jbw976, line 12: you could link this first mention of Upbound Cloud to upbound.io

docs/getting-started/install-configure.md
```diff
@@ -5,16 +5,39 @@ weight: 2
...
+
+<i>Want see another hosted Crossplane service listed? Please 
+<a href="http://slack.crossplane.io/">reach out on Slack</a> and our community will 
+highlight it here!</a></i>
+
+## Start with a Self-Hosted Crossplane
+Installing Crossplane into an existing Kubernetes cluster will require a bit more
+setup, but can provide slightly more configurability today.
+
+### Get a Kubernetes Cluster
```
hasheddan, line 40: What if we just add tabs here for hosted vs self-hosted, similar to how we do for different providers.

docs/getting-started/install-configure.md
```diff
@@ -5,16 +5,39 @@ weight: 2
...
+
+## Start with a Hosted Crossplane
+Upbound Cloud is a managed service of Crossplane created by the founders of Crossplane. 
+You can [create an account](https://cloud.upbound.io/register) to get started. Once 
+logged in, you can [create](https://cloud.upbound.io/docs/getting-started/set-up-upbound-cloud) 
+and then [connect](https://cloud.upbound.io/docs/getting-started/connect-to-your-platform) 
+to your hosted Crossplane cluster.
 
-## Get a Kubernetes Cluster
+Once you've completed thse two steps, skip down to
```
jbw976, line 29: typo: thse --> these


i recommend a spell checker for your IDE that can catch this stuff early on, e.g. https://github.com/streetsidesoftware/vscode-spell-checker

docs/getting-started/install-configure.md
grantgumina: That's very fair and good feedback. Would something like this be more appropriate instead?

> Users looking to use Crossplane for the first time have two options avaiable to
> them today. The first way is to use a hosted Crossplane service like Upbound
> Cloud. Alternatively, users looking for some more flexability can install
> Crossplane into their own Kubernetes cluster.
> 
> Crossplane will be installed using the regularly published Helm chart. The Helm
> chart contains all the custom resources and controllers needed to deploy and
> configure Crossplane.
>

docs/getting-started/install-configure.md
negz: This was the approach I was thinking we'd take too.

grantgumina: That was my first inclination as well, but became concerned about having nested tabs. Attempting to add this in now so we can see how it renders.

docs/getting-started/install-configure.md
```diff
@@ -1,20 +1,54 @@
 ---
-title: Install & Configure
-toc: true
-weight: 2
-indent: true
+title: Install & Configure toc: true weight: 2 indent: true
```
negz, line 2: Looks like this got wrapped unexpectedly. :(

docs/getting-started/install-configure.md
```diff
@@ -1,20 +1,54 @@
...
 ---
 
-# Install & Configure Crossplane
+# Choosing Hosted or Self-Hosted Crossplane
 
-Crossplane can be easily installed into any existing Kubernetes cluster using
-the regularly published Helm chart. The Helm chart contains all the custom
-resources and controllers needed to deploy and configure Crossplane.
+Users looking to use Crossplane for the first time have two options avaiable to
```
negz, line 7:
```suggestion
Users looking to use Crossplane for the first time have two options available to
```

docs/getting-started/install-configure.md
```diff
@@ -1,20 +1,54 @@
...
+them today. The first way is to use a hosted Crossplane service like [Upbound
+Cloud]("https://upbound.io"). Alternatively, users looking for some more
```
negz, line 9: Nit: This doc uses reference style links - it would be good to be consistent.

https://www.markdownguide.org/basic-syntax/#reference-style-links

docs/getting-started/install-configure.md
grantgumina: Apologies for these mistakes. My editor's autoformat for line width consolidated these lines. Pushing to fix.

docs/getting-started/install-configure.md
grantgumina: Done.

docs/getting-started/install-configure.md
grantgumina: Took Jared's advice and installed a spellchecker. :)

negz (approved):
Thanks for persisting with this @grantgumina!

grantgumina:
> Thanks for persisting with this @grantgumina!

And thank you for being so forgiving as you put up with my rookie mistakes!

PR #2025, opened 2020-12-11
Add v1beta1 package meta type

hasheddan:
### Description of your changes

Adds v1beta1 package meta types and updates crank and package manager to accept them.

I have:

- [x] Read and followed Crossplane's [contribution process].
- [x] Run `make reviewable test` to ensure this PR is ready for review.

### How has this code been tested

Will report testing steps here and move out of draft when ready :+1: 

[contribution process]: https://git.io/fj2m9

PR #2029, opened 2020-12-12
Add support for meta.pkg.crossplane.io/v1

negz:

Closes https://github.com/crossplane/crossplane/pull/2025

This PR enables support for `meta.pkg.crossplane.io/v1` package metadata types. The types have an identical schema to their v1alpha1 equivalents, which continue to be supported. "User-facing" API types are converted to an internal type that is used in our controller logic. We use the same controller-runtime pattern for conversion as would be used for a conversion webhook.

I'm conscious that this is a somewhat sizeable (though hopefully not _too_ sophisticated) change coming in at the last hour before our 1.0 release. I've tried to ensure it has good unit test coverage, and have manually tested all of the relevant scenarios I can think of - please let me know if there are other scenarios I should test to help build confidence in the change.

If we deem this change too risky, our alternative options are to:

1. Make no changes, and accept that Crossplane v1.0 will depend on packages that use a v1alpha1 metadata format.
1. Bump the v1alpha1 metadata format to v1 in place, thus dropping support for v1alpha1. All existing v1alpha1 packages would need to be rebuilt and pushed to work with Crossplane 1.0.

I have:

I have tested that:

* I can install `registry.upbound.io/xp/getting-started-with-gcp:latest`, a v1alpha1 Configuration that depends on a v1alpha1 Provider. Both the Configuration and the Provider install successfully. I installed this using the crank binary produced by this PR.
* I can use the build of crank produced by this PR to build a v1 variant of `registry.upbound.io/xp/getting-started-with-gcp:latest` from `doc/snippets/package`, then install that variant successfully with both the build of crank produced by this PR, and an older build.
* I can install a v1alpha1 Provider (`crossplane/provider-azure:master`).
* I can build and install a v1 Provider (`negz/provider-aws:v1`).


apis/pkg/meta/configuration_types.go
```diff
@@ -0,0 +1,40 @@
+/*
```
negz, line 1: Note that the newly introduced `pkg/meta` and `pkg/meta/v1` types are identical to the existing `pkg/meta/v1alpha1` types.

apis/pkg/meta/configuration_types.go
```diff
@@ -0,0 +1,40 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ConfigurationSpec specifies the configuration of a Configuration.
+type ConfigurationSpec struct {
+	MetaSpec `json:",inline"`
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:skipversion
```
negz, line 29: This annotation marks this version as internal - we don't include it in our generated CRD 'API docs'.

apis/pkg/meta/interfaces.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package v1alpha1
+package meta
```
negz, line 17: The `Pkg` interface now lives (only) in the internal `meta` package.

apis/pkg/meta/v1/configuration_types.go
```diff
@@ -0,0 +1,103 @@
...
+// +kubebuilder:storageversion
+
+// A Configuration is the description of a Crossplane Configuration package.
+type Configuration struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec ConfigurationSpec `json:"spec"`
+}
+
```
negz, line 47: The below conversion functions are new.

apis/pkg/meta/v1/provider_types.go
```diff
@@ -0,0 +1,135 @@
...
+// +kubebuilder:storageversion
+
+// A Provider is the description of a Crossplane Provider package.
+type Provider struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec ProviderSpec `json:"spec"`
+}
+
```
negz, line 64: The below conversion functions are new.

apis/pkg/meta/v1alpha1/configuration_types.go
```diff
@@ -34,3 +43,61 @@ type Configuration struct {
 
 	Spec ConfigurationSpec `json:"spec"`
 }
+
```
negz, line 46: The below conversion functions are new. They're identical to the equivalent v1 conversion functions.

apis/pkg/meta/v1alpha1/provider_types.go
```diff
@@ -51,3 +60,75 @@ type Provider struct {
 
 	Spec ProviderSpec `json:"spec"`
 }
+
```
negz, line 63: The below conversion functions are new. They're identical to the equivalent v1 conversion functions.

internal/controller/pkg/revision/hook.go
```diff
@@ -161,13 +170,6 @@ func NewConfigurationHooks() *ConfigurationHooks {
 
 // Pre sets status fields based on the configuration package.
 func (h *ConfigurationHooks) Pre(ctx context.Context, pkg runtime.Object, pr v1.PackageRevision) error {
-	_, ok := pkg.(*pkgmeta.Configuration)
-	if !ok {
-		return errors.New(errNotConfiguration)
-	}
-
-	// TODO(hasheddan): update any status fields relevant to package revisions
-
```
negz, line 170: I noticed we were checking this type assertion in the pre hook but not the post hook, where both are no-ops. I figured it was safe to remove this rather than updating it given that we already perform this check in the linter.

hasheddan (commented):
Thanks for tackling this @negz. I have left comments below on the implementation. I don't think we necessarily need the internal type here, but I could be convinced if you think it is cleaner to import and use it. I do think calling the `xpkg.ConvertToPkg` once in the reconciler instead of in each sub-action would be an improvement though.

apis/pkg/meta/v1/provider_types.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	p.Spec.DependsOn = make([]Dependency, len(in.Spec.DependsOn))
+	for i := range in.Spec.DependsOn {
+		p.Spec.DependsOn[i] = Dependency{
+			Provider:      in.Spec.DependsOn[i].Provider,
+			Configuration: in.Spec.DependsOn[i].Configuration,
+			Version:       in.Spec.DependsOn[i].Version,
+		}
+	}
```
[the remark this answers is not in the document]
negz: Good point, but I'd like to address this in a future PR. That way we can take a little more time to establish type conversion patterns, since we'll presumably be doing this a fair bit once we introduce conversion webhooks.

apis/pkg/meta/register.go
```diff
@@ -0,0 +1,62 @@
...
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/scheme"
+)
+
+// Package type metadata.
+const (
+	Group   = "meta.pkg.crossplane.io"
+	Version = "meta"
+)
```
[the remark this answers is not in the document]
negz: I have a hunch that it could make our lives easier as we get deeper into conversions. I like that it would give us fixed types to build our actual code around. Versioned 'user facing' types could be simple types with no logic and our getters and setters, composition methods, etc could be associated with only the internal type. We'd also never need to go through the process of bumping our controllers to a wholesale new type (e.g. `s/v1beta1/v1/g` on the codebase). I suspect there could be some more advantages for roundtrippability such as those laid out in ASO's [research on this](https://github.com/Azure/k8s-infra/blob/611763e4ae99501b69372535020774ed46529125/hack/generator/spec/versioning.md).

That all said, I agree that it's not the right time to introduce such a pattern so I've pushed a commit that removes the internal types and uses `v1` as the conversion hub.

internal/controller/pkg/revision/dependency.go
```diff
@@ -67,7 +68,7 @@ func NewPackageDependencyManager(c client.Client, nd dag.NewDAGFn, t v1alpha1.Pa
 
 // Resolve resolves package dependencies.
 func (m *PackageDependencyManager) Resolve(ctx context.Context, pkg runtime.Object, pr v1.PackageRevision) (found, installed, invalid int, err error) { // nolint:gocyclo
-	pack, ok := pkg.(pkgmeta.Pkg)
+	pack, ok := xpkg.ConvertToPkg(pkg, &pkgmeta.Provider{}, &pkgmeta.Configuration{})
```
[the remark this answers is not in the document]
negz: Yeah, I started out with the call being there then realised we also needed conversion in the linter calls which happen before this. I've now updated it so that we make a call here and everywhere else we might need conversion, but we only convert once (redundant calls will skip conversion).

negz:
@hasheddan PTAL when you get time - I believe I've addressed most of the feedback and I've run all the manual tests again.

hasheddan (approved):
LGTM - I think the Configuration build will probably fail on merge until the new CLI version is available, but I'll monitor to make sure.

PR #2034, opened 2020-12-13
Build configuration packages with v1alpha1 meta types

hasheddan:

Reverts update to build configuration packages with v1 meta types and
uses v1alpha1 for the time being, which are still supported in the
package manager.

I have:

`kubectl crossplane build configuration ...`

### DIFF
--- a/docs/snippets/package/alibaba/crossplane.yaml
+++ b/docs/snippets/package/alibaba/crossplane.yaml
@@ -1,4 +1,4 @@
-apiVersion: meta.pkg.crossplane.io/v1
+apiVersion: meta.pkg.crossplane.io/v1alpha1
 kind: Configuration
 metadata:
   name: getting-started-with-alibaba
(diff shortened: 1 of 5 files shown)


negz (commented):
@hasheddan is it possible to temporarily decouple the Configuration push from our docs? i.e. To (manually?) build and push `v1alpha1` packages that the Upbound Registry will accept, but keep the docs referring to `v1` metadata?

hasheddan:
@negz yep we could do that! We could also keep pushing them and just pin the install command to previously built version. Do you have a preference?

negz:
> Do you have a preference?

I don't think so - happy to go with whatever is easiest.

That said, will we need to pin the provider deps of these configs in the v1 release branch such that they won't resolve their provider dependencies to RC providers? I imagine that would require a manual package build.

hasheddan:
We could just update the action to only on workflow dispatch and do manual builds if needed. We should probably make all versions just depend on stable versions of their providers though, in which case continuing to build on merges would work fine I think.

hasheddan:
@negz all the configurations' dependencies have now been updated to require a stable version :+1:

negz (approved):

PR #2036, opened 2020-12-14
Touchup the troubleshooting documentation

negz:

* crossplane trace hasn't been ported to the new CLI
* Most core resources don't use the 'synced' condition
* Events are our primary info source
* What's a 'stack'?

I have:

N/A

### DIFF
--- a/docs/reference/troubleshoot.md
+++ b/docs/reference/troubleshoot.md
@@ -7,34 +7,11 @@ indent: true
 
 # Troubleshooting
 
-* [Using the trace command]
 * [Resource Status and Conditions]
+* [Resource Events]
 * [Crossplane Logs]
 * [Pausing Crossplane]
 * [Deleting a Resource Hangs]
-
-## Using the trace command
-
-The [Crossplane CLI] trace command provides a holistic view for a particular
-object and related ones to ease debugging and troubleshooting process. It finds
-the relevant Crossplane resources for a given one and provides detailed
-information as well as an overview indicating what could be wrong.
-
-Usage:
-```
-kubectl crossplane trace TYPE[.GROUP] NAME [-n| --namespace NAMESPACE] [--kubeconfig KUBECONFIG] [-o| --outputFormat dot]
-```
-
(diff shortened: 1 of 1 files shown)


docs/reference/troubleshoot.md
```diff
@@ -80,9 +68,13 @@ namespace. To get the current Crossplane logs, run the following:
...
 
-Remember that much of Crossplane's functionality is provided by Stacks. You can
-use `kubectl logs` to view Stack logs too, though Stacks may not run in the
-`crossplane-system` namespace.
+Remember that much of Crossplane's functionality is provided by providers. You
+can use `kubectl logs` to view provider logs too.
+
+> Note that Crossplane emits few logs by default - events are typically the best
+> place to look for information about what Crossplane is doing. You may need to
+> restart Crossplane (or your provider) with the `--debug` flag if you can't
```
muvaf, line 76: Is this possible? Won't package manager override the `Deployment`?

muvaf (approved):
LGTM, one correction maybe.

docs/reference/troubleshoot.md
hasheddan: @muvaf I believe a flag should persist as we only patch the deployment in the provider revision controller:

https://github.com/crossplane/crossplane/blob/59620600805c4d8297a9a696ddf46104a09371f5/internal/controller/pkg/revision/reconciler.go#L220

That being said, I would recommend that a user reference a `ControllerConfig` in this situation.

hasheddan (approved):

docs/reference/troubleshoot.md
muvaf: @hasheddan users would need to add the flag to `command: ` array. Do we have this array right now in the default case? If we do, I believe patch will replace the whole list once it sees any difference.

negz: I think this note has value regardless of how difficult it may be to actually restart the controller if it's under package manager management.

PR #2053, opened 2020-12-23
Update build submodule and use helm v2.17.0

hasheddan:

Helm recently changed the location of stable charts, which is
automatically handled in the latest patch release. This updates the
build submodule to pick up the latest patch and get our CI pipelines
working again.

I have:

`make build`


jbw976 (commented):
a few of the CI checks failed, and they look related to the helm stable charts issue.  do we expect these to start passing again AFTER this is merged?  I know how Jenkins used to behave during CI builds for PRs that change build files, but I'm not sure how GH Actions behaves.

hasheddan:
@jbw976 I believe it is because we are still falling back to helm 2 here, which was not updated. I am going to bump to helm 3.

hasheddan:
Note that we will be blocked on update to helm 3 until https://github.com/helm/helm/pull/9118 is merged.

hasheddan:
@jbw976 this should be good to go now :+1:

Makefile
```diff
@@ -38,6 +38,7 @@ GO111MODULE = on
 # ====================================================================================
 # Setup Kubernetes tools
 
+HELM_VERSION=v2.17.0
```
jbw976, line 41: do we still need to update the build submodule if we are using this newer version of helm2?

cluster/local/integration_tests.sh
```diff
@@ -43,7 +43,7 @@ BUILD_IMAGE="${BUILD_REGISTRY}/${PROJECT_NAME}-${HOSTARCH}"
 
 helm_tag="$(cat ${projectdir}/_output/version)"
 CROSSPLANE_IMAGE="${DOCKER_REGISTRY}/${PROJECT_NAME}:${helm_tag}"
-K8S_CLUSTER="${K8S_CLUSTER:-${BUILD_REGISTRY}-INTTESTS}"
+K8S_CLUSTER="${K8S_CLUSTER:-${BUILD_REGISTRY}-inttests}"
```
jbw976, line 46: what caused this casing to need to change? something in the build submodule set of updates?

Makefile
hasheddan: We don't, but I think it is good to continue to update frequently. In reality I think we should have the build submodule versioned, but right now the latest commit it essentially the latest version.

cluster/local/integration_tests.sh
hasheddan: kind version bump in build submodule, we saw the same change required in other repos 👍

jbw976 (approved):

PR #2069, opened 2021-01-11
Create podmonitor for helm chart

duncaan:

Creates a prometheus service monitor with the helm chart.
Fixes #

I have:

I rendered the templates locally and they look good.


cluster/charts/crossplane/values.yaml.tmpl
```diff
@@ -62,3 +62,5 @@ alpha:
 
 metrics:
   enabled: false
+  monitor:
+    enabled: true
```
hasheddan, line 66: I think we should probably make this `false` by default.

cluster/charts/crossplane/README.md
```diff
@@ -95,6 +95,7 @@ securityContextRBACManager:
 | `rbacManager.managementPolicy`| The extent to which the RBAC manager will manage permissions. `All` indicates to manage all Crossplane controller and user roles. `Basic` indicates to only manage Crossplane controller roles and the `crossplane-admin`, `crossplane-edit`, and `crossplane-view` user roles. | `All` |
 | `alpha.oam.enabled` | Deploy the `crossplane/oam-kubernetes-runtime` Helm chart | `false` |
 | `metrics.enabled` | Expose Crossplane and RBAC Manager metrics endpoint | `false` |
+| `metrics.monitor.enabled` | Create prometheus Service Monitor and Service for scraping the metrics endpoint | `false` |
```
hasheddan, line 98: Might be worth mentioning that this requires kube-prometheus to be installed.

hasheddan (commented):
Thanks for thinking of this @duncaan! Could you provide some information around your testing of this with prometheus?

cluster/charts/crossplane/values.yaml.tmpl
duncaan: I forgot to disable this after testing.

duncaan:
> Thanks for thinking of this @duncaan! Could you provide some information around your testing of this with prometheus?

For _reasons_ we've disabled prometheus scraping by using the annotations. This just gives another way to monitor the application, which is supported by most charts.

cluster/charts/crossplane/README.md
```diff
@@ -95,6 +95,7 @@ securityContextRBACManager:
 | `rbacManager.managementPolicy`| The extent to which the RBAC manager will manage permissions. `All` indicates to manage all Crossplane controller and user roles. `Basic` indicates to only manage Crossplane controller roles and the `crossplane-admin`, `crossplane-edit`, and `crossplane-view` user roles. | `All` |
 | `alpha.oam.enabled` | Deploy the `crossplane/oam-kubernetes-runtime` Helm chart | `false` |
 | `metrics.enabled` | Expose Crossplane and RBAC Manager metrics endpoint | `false` |
+| `metrics.monitor.enabled` | Create prometheus Service Monitor and Service for scraping the metrics endpoint (Requires Prometheus to be installed) | `false` |
```
negz, line 98: Does this require the Prometheus Operator specifically to be installed? Should we call that out explicitly, or is it pretty much implied these days? It's been a while since I've deployed Prometheus.

hasheddan: @negz I believe it depends on `kube-prometheus` because the core os `ServiceMonitor` CRD must be present https://github.com/crossplane/crossplane/pull/2069#discussion_r555168499

hasheddan:
@duncaan just wondering, is there a specific motivation for using `ServiceMonitor` over `PodMonitor`?

duncaan:
> @duncaan just wondering, is there a specific motivation for using `ServiceMonitor` over `PodMonitor`?

No specific motivation other than not knowing about a `PodMonitor` :)
I tested a PodMonitor yesterday and seemed to work fine.  I'll add that to this PR if that's cool.

hasheddan:
@duncaan sounds good :) let's switch out `ServiceMonitor` for `PodMonitor` then this should be good to go 👍

hasheddan (approved):
@duncaan need a quick rebase here but otherwise looks good :+1:

hasheddan:
@duncaan could you squash commits here / remove merge commit? I usually use something like:

```
git checkout master && git fetch upstream && git merge upstream/master
git checkout create-servicemonitor
git rebase -i HEAD~4
# mark commits to be squashed
# update commit message
git push -f
```

PR #2073, opened 2021-01-12
Composition dependency

### LINKED ISSUE #2072: Composition Dependency and Ordered Creation
Although my default Crossplane creates all resources in a composition in order, this typically does not include the entire reconciliation lifecycle for provisioning some abstract resource. Just applying a resource does not mean that has been created, and we may have other resources in the composition rely on a previous resource. Currently, within Crossplane the main mechanism for handling these issues is the reconciliation loop, however, this is not always feasible if we are working with resources outside of the Crossplane ecosystem. Similar issues also arise if we want to make stronger claims about the resources we are creating (e.g., if we want to block the creation of an RDS database until we are certain that all of the subnets were created).

This is a use case that was previously outlined in a [discussion](https://crossplane.slack.com/archives/CEG3T90A1/p1610110813347800) on the Crossplane Slack. 

I think it would be a good idea to have a depends_on field at the top level of each composition, similar to readinessChecks, where users can input fields to assign dependency to some other resource. The easiest kind would be a dependency on the previous resource in the composition, but we could also expand this to dependency on any other resource and utilize the same base API types as the readinessChecks. I have a draft PR which I will open up soon for this issue.

krishchow:

Fixes #2072 

I have:

The code has been manually tested, but more testing is required still. This is a draft PR and should not be merged yet. 


negz:
I don't think we're likely to proceed with this for the time being, so I'm going to close this PR. I'll leave the issue open though since we've had some good discussion there. Thanks for opening this to illustrate though @krishchow!

PR #2076, opened 2021-01-13
Warn user that kubectl plugin might be outdated

khos2ow:

When outdated kubectl crossplane plugin is used to install provider or
configuration, user will face a generic and a bit misleading error
message of "the server could not find the resource." At this point user
won't have more context of what happened when in reality they might be
using older version of crossplane CLI plugin all along.

https://crossplane.slack.com/archives/CEG3T90A1/p1610142999359200 for a bit more context

I have:

- deploy the latest version of crossplane using `crossplane-master` chart
- manually check out tag `v0.13.0` and apply this patch and build a new crank binary
- use the binary to install `provider-aws:master` and make sure the warning is being shown properly

### DIFF
--- a/cmd/crank/install.go
+++ b/cmd/crank/install.go
@@ -19,17 +19,20 @@ package main
 import (
 	"context"
 	"fmt"
+	"net/http"
 	"strings"
 
 	"github.com/alecthomas/kong"
 	"github.com/google/go-containerregistry/pkg/name"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 
 	v1 "github.com/crossplane/crossplane/apis/pkg/v1"
 	typedclient "github.com/crossplane/crossplane/internal/client/clientset/versioned/typed/pkg/v1"
+	"github.com/crossplane/crossplane/internal/version"
 	"github.com/crossplane/crossplane/internal/xpkg"
 
 	// Load all the auth plugins for the cloud providers.
@@ -97,7 +100,7 @@ func (c *installConfigCmd) Run(k *kong.Context) error {
 	kube := typedclient.NewForConfigOrDie(ctrl.GetConfigOrDie())
 	res, err := kube.Configurations().Create(context.Background(), cr, metav1.CreateOptions{})
 	if err != nil {
-		return errors.Wrap(err, "cannot create configuration")
+		return errors.Wrap(warnIfNotFound(err), "cannot create configuration")
 	}
 	_, err = fmt.Fprintf(k.Stdout, "%s/%s created\n", strings.ToLower(v1.ConfigurationGroupKind), res.GetName())
 	return err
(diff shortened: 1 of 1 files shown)


cmd/crank/install.go
```diff
@@ -141,8 +145,19 @@ func (c *installProviderCmd) Run(k *kong.Context) error {
...
 	return err
 }
+
+func warnOutdated(err error) {
+	serr, ok := err.(*apierrors.StatusError)
+	if !ok {
+		return
+	}
+	if serr.ErrStatus.Code == http.StatusNotFound {
+		fmt.Printf("kubectl crossplane: warning: kubectl-crossplane plugin might be out of date (installed version: %s)\n", version.New().GetVersionString())
```
hasheddan, line 161: @khos2ow can we send this to `k.Stdout`?

khos2ow: I cannot think of a reason why we couldn't. I had thought about it but wanted to get the general feel of this approach before trying that.

hasheddan: Oh yes I know we can :) It is nice to do so because you can plumb down a different stdout for testing :+1: I think the solution you have here in general sounds good to me for now. I still feel the implementation of the various CLI methods is a little sloppy (my fault), but it works as expected

khos2ow: Refactored (might be a bit too much!)

cmd/crank/install.go
```diff
@@ -141,8 +148,22 @@ func (c *installProviderCmd) Run(k *kong.Context) error {
 	kube := typedclient.NewForConfigOrDie(ctrl.GetConfigOrDie())
 	res, err := kube.Providers().Create(context.Background(), cr, metav1.CreateOptions{})
 	if err != nil {
+		if msg := warnIfNotFound(err); msg != "" {
+			fmt.Fprintf(k.Stdout, warnPluginOutdated, msg, version.New().GetVersionString())
+		}
 		return errors.Wrap(err, "cannot create provider")
```
hasheddan, line 154: @khos2ow how about just passing `warnIfNotFound(err)` here and passing in the alternate message?
```suggestion
		return errors.Wrap(err, warnIfNotFound(err, "cannot create provider"))
```

khos2ow: I originally tried similar approach (but with `errors.WithMessage()`) and I personally didn't like the output message. It reads as:

```
kubectl crossplane: error: cannot create provider: kubectl-crossplane plugin might be out of date: the server could not find the requested resource (post providers.pkg.crossplane.io)
```

The order is wrong, IMO. `kubectl crossplane: error: <title>: <some speculation>: <what actually went wrong>`. I mean it's kinda ok too. What do you think?

If you like the output error message then I'd actually convert your suggestion to this:

```suggestion
		return errors.Wrap(warnIfNotFound(err), "cannot create provider")
```

hasheddan: hmm yeah it is kind of a peculiar looking error. However, I do think it is better than printing the `kubectl-crossplane plugin might be out of date` on a line before the error message. The heavy-handed solution would be to implement a custom wrapper, but I would be fine to what your change here in the near term.

khos2ow: I wanted to show the warning after the error (which wasn't possible due to current implementation of logging) and settled for the next thing (to show it before). But now to think of it, I prefer a one liner better. Will update.

hasheddan (approved):
@khos2ow LGTM, can you rebase :)

khos2ow:
@hasheddan this should be good to go, but essentially this is future proofing and doesn't do anything for existing outdated plugins out there being used. Maybe worth dropping a line in docs FAQ or somewhere as part of this PR too?

hasheddan:
@khos2ow let's do that in a follow-up, maybe add an entry in https://crossplane.io/docs/v1.0/reference/troubleshoot.html

PR #2077, opened 2021-01-14
Add bidirectional patches to composite resources

mcavoyk:

Fixes https://github.com/crossplane/crossplane/issues/1639.

Adds new patch type `ToCompositeFieldPath`. Adds composite render to composite reconciler to apply patches which modify composite from composed. Also adds ClaimConfigurator to claim reconciler to late initialize spec values from composite and update status fields from composite. 

Fixes https://github.com/crossplane/crossplane/issues/2081, which is strongly related, previously status fields from XRD's were not persisted but status fields within claims and compositions can now be patched from composed resources.

I have:

Tested with an XRD which creates a deployment. Patches parameters to deployment spec and status back from deployment.

Tested both Composite and CompositeClaim resource.

 <details>
    <summary>xrd-deploy.yaml</summary>
  <br></br>
<head class="lang-yaml">

    metadata:
      creationTimestamp: null
      name: compositedeploys.example.com
    spec:
      claimNames:
        kind: Deploy
        plural: deploys
        singular: deploy
      group: example.com
      names:
        kind: CompositeDeploy
        plural: compositedeploys
        singular: compositedeploy
      versions:
        - name: v1alpha1
          referenceable: true
          served: true
          schema:
            openAPIV3Schema:
              properties:
                spec:
                  type: object
                  properties:
                    parameters:
                      type: object
                      properties:
                        namespace:
                          type: string
                        image:
                          type: string
                        replicas:
                          type: integer
                status:
                  type: object
                  properties:
                    readyReplicas:
                      type: integer
                    updatedReplicas:
                      type: integer
</head>
</details>

<details>
    <summary>composition-deploy.yaml</summary>
  <br></br>
<head class="lang-yaml">

    kind: Composition
    metadata:
      name: compositedeploys
    spec:
      compositeTypeRef:
        apiVersion: example.com/v1alpha1
        kind: CompositeDeploy
      resources:
        - base:
            apiVersion: apps/v1
            kind: Deployment
            metadata:
              name: xrd-test
              namespace: default
            spec:
              replicas: 1
              template:
                spec:
                  containers:
                    - name: default
                      image: busybox:latest
                      command:
                        - /bin/sh
                        - "-c"
                      args:
                        - "echo starting...; sleep infinity"
          patches:
            - fromFieldPath: metadata.name
              toFieldPath: metadata.name
              type: FromCompositeFieldPath
            - fromFieldPath: metadata.name
              toFieldPath: spec.selector.matchLabels[app]
              type: FromCompositeFieldPath
            - fromFieldPath: metadata.name
              toFieldPath: spec.template.metadata.labels[app]
              type: FromCompositeFieldPath
    
            # inputs
            - fromFieldPath: spec.parameters.namespace
              toFieldPath: metadata.namespace
              type: FromCompositeFieldPath
            - fromFieldPath: spec.parameters.image
              toFieldPath: spec.template.spec.containers[0].image
              type: FromCompositeFieldPath
            - fromFieldPath: spec.parameters.replicas
              toFieldPath: spec.replicas
              type: FromCompositeFieldPath
    
            # outputs
            - fromFieldPath: spec.replicas
              toFieldPath: spec.parameters.replicas
              type: ToCompositeFieldPath
            - fromFieldPath: status.readyReplicas
              toFieldPath: status.readyReplicas
              type: ToCompositeFieldPath
            - fromFieldPath: status.updatedReplicas
              toFieldPath: status.updatedReplicas
              type: ToCompositeFieldPath
          readinessChecks:
            - type: MatchString
              fieldPath: status.conditions[0].status
              matchString: "True"
</head>
</details>

 <details>
    <summary>composite-deploy.yaml</summary>
  <br></br>
<head class="lang-yaml">

    apiVersion: example.com/v1alpha1
    kind: CompositeDeploy
    metadata:
      name: composite-test
    spec:
      parameters:
        namespace: default
        image: busybox:1
        replicas: 1
</head>
</details>

 <details>
    <summary>claim-deploy.yaml</summary>
  <br></br>
<head class="lang-yaml">

    apiVersion: example.com/v1alpha1
    kind: Deploy
    metadata:
      name: claim-test
    spec:
      parameters:
        namespace: default
        image: busybox:1
        replicas: 2
</head>
</details>

### DIFF
--- a/apis/apiextensions/v1/composition_types_test.go
+++ b/apis/apiextensions/v1/composition_types_test.go
@@ -24,7 +24,6 @@ import (
 	"github.com/google/go-cmp/cmp"
 	"github.com/pkg/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-
 	"k8s.io/utils/pointer"
 
 	"github.com/crossplane/crossplane-runtime/pkg/resource/fake"
@@ -481,6 +480,7 @@ func TestPatchApply(t *testing.T) {
 		patch Patch
 		cp    *fake.Composite
 		cd    *fake.Composed
+		only  []PatchType
 	}
 	type want struct {
 		cp  *fake.Composite
@@ -494,7 +494,7 @@ func TestPatchApply(t *testing.T) {
 		want
 	}{
 		"InvalidCompositeFieldPathPatch": {
-			reason: "Should return error when required fields not passed to applyFromCompositeFieldPatch",
+			reason: "Should return error when required fields not passed to applyFromFieldPathPatch",
 			args: args{
 				patch: Patch{
 					Type: PatchTypeFromCompositeFieldPath,
@@ -546,9 +546,76 @@ func TestPatchApply(t *testing.T) {
 			},
 			want: want{
 				cd: &fake.Composed{
-					ObjectMeta: metav1.ObjectMeta{Name: "cd", Labels: map[string]string{
-						"Test": "blah",
-					}},
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "cd",
+						Labels: map[string]string{
+							"Test": "blah",
+						},
+					},
+				},
+				err: nil,
+			},
+		},
+		"FilterExcludeCompositeFieldPathPatch": {
+			reason: "Should not apply the patch as the PatchType is not present in filter.",
+			args: args{
+				patch: Patch{
+					Type:          PatchTypeFromCompositeFieldPath,
+					FromFieldPath: pointer.StringPtr("objectMeta.labels"),
+					ToFieldPath:   pointer.StringPtr("objectMeta.labels"),
+				},
+				cp: &fake.Composite{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "cp",
+						Labels: map[string]string{
+							"Test": "blah",
+						},
+					},
+					ConnectionDetailsLastPublishedTimer: lpt,
+				},
+				cd: &fake.Composed{
+					ObjectMeta: metav1.ObjectMeta{Name: "cd"},
+				},
+				only: []PatchType{PatchTypePatchSet},
+			},
+			want: want{
+				cd: &fake.Composed{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "cd",
+					},
+				},
+				err: nil,
+			},
+		},
+		"FilterIncludeCompositeFieldPathPatch": {
+			reason: "Should apply the patch as the PatchType is present in filter.",
+			args: args{
+				patch: Patch{
+					Type:          PatchTypeFromCompositeFieldPath,
+					FromFieldPath: pointer.StringPtr("objectMeta.labels"),
+					ToFieldPath:   pointer.StringPtr("objectMeta.labels"),
+				},
+				cp: &fake.Composite{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "cp",
+						Labels: map[string]string{
+							"Test": "blah",
+						},
+					},
+					ConnectionDetailsLastPublishedTimer: lpt,
+				},
+				cd: &fake.Composed{
+					ObjectMeta: metav1.ObjectMeta{Name: "cd"},
+				},
+				only: []PatchType{PatchTypeFromCompositeFieldPath},
+			},
+			want: want{
+				cd: &fake.Composed{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "cd",
+						Labels: map[string]string{
+							"Test": "blah",
+						}},
 				},
 				err: nil,
 			},
@@ -570,14 +637,70 @@ func TestPatchApply(t *testing.T) {
 					ConnectionDetailsLastPublishedTimer: lpt,
 				},
 				cd: &fake.Composed{
-					ObjectMeta: metav1.ObjectMeta{Name: "cd"},
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "cd",
+					},
 				},
 			},
 			want: want{
+				cp: &fake.Composite{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "cp",
+						Labels: map[string]string{
+							"Test": "blah",
+						},
+					},
+					ConnectionDetailsLastPublishedTimer: lpt,
+				},
 				cd: &fake.Composed{
-					ObjectMeta: metav1.ObjectMeta{Name: "cd", Labels: map[string]string{
-						"Test": "blah",
-					}},
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "cd",
+						Labels: map[string]string{
+							"Test": "blah",
+						}},
+				},
+				err: nil,
+			},
+		},
+		"ValidToCompositeFieldPathPatch": {
+			reason: "Should correctly apply a ToCompositeFieldPath patch with valid settings",
+			args: args{
+				patch: Patch{
+					Type:          PatchTypeToCompositeFieldPath,
+					FromFieldPath: pointer.StringPtr("objectMeta.labels"),
+					ToFieldPath:   pointer.StringPtr("objectMeta.labels"),
+				},
+				cp: &fake.Composite{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "cp",
+					},
+					ConnectionDetailsLastPublishedTimer: lpt,
+				},
+				cd: &fake.Composed{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "cd",
+						Labels: map[string]string{
+							"Test": "blah",
+						},
+					},
+				},
+			},
+			want: want{
+				cp: &fake.Composite{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "cp",
+						Labels: map[string]string{
+							"Test": "blah",
+						},
+					},
+					ConnectionDetailsLastPublishedTimer: lpt,
+				},
+				cd: &fake.Composed{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "cd",
+						Labels: map[string]string{
+							"Test": "blah",
+						}},
 				},
 				err: nil,
 			},
@@ -586,7 +709,7 @@ func TestPatchApply(t *testing.T) {
 	for name, tc := range cases {
 		t.Run(name, func(t *testing.T) {
 			ncp := tc.args.cp.DeepCopyObject()
-			err := tc.args.patch.Apply(ncp, tc.args.cd)
+			err := tc.args.patch.Apply(ncp, tc.args.cd, tc.args.only...)
 
 			if tc.want.cp != nil {
 				if diff := cmp.Diff(tc.want.cp, ncp); diff != "" {
--- a/go.mod
+++ b/go.mod
@@ -11,6 +11,7 @@ require (
 	github.com/docker/docker v17.12.0-ce-rc1.0.20200926000217-2617742802f6+incompatible // indirect
 	github.com/google/go-cmp v0.5.2
 	github.com/google/go-containerregistry v0.2.1
+	github.com/imdario/mergo v0.3.11
 	github.com/pkg/errors v0.9.1
 	github.com/spf13/afero v1.4.1
 	gopkg.in/alecthomas/kingpin.v2 v2.2.6
--- a/internal/controller/apiextensions/claim/configurator.go
+++ b/internal/controller/apiextensions/claim/configurator.go
@@ -20,9 +20,10 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/imdario/mergo"
 	"github.com/pkg/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
-	"github.com/crossplane/crossplane-runtime/pkg/fieldpath"
 	"github.com/crossplane/crossplane-runtime/pkg/meta"
 	"github.com/crossplane/crossplane-runtime/pkg/resource"
 	"github.com/crossplane/crossplane-runtime/pkg/resource/unstructured/claim"
@@ -30,19 +31,19 @@ import (
 	"github.com/crossplane/crossplane/internal/xcrd"
 )
 
-// Label keys.
 const (
-	LabelKeyNamePrefixForComposed = "crossplane.io/composite"
-	LabelKeyClaimName             = "crossplane.io/claim-name"
-	LabelKeyClaimNamespace        = "crossplane.io/claim-namespace"
+	errUnsupportedClaimSpec = "composite resource claim spec was not an object"
+	errUnsupportedDstObject = "destination object was not valid object"
+	errUnsupportedSrcObject = "source object was not valid object"
+
+	errMergeClaimSpec   = "unable to merge claim spec"
+	errMergeClaimStatus = "unable to merge claim status"
 )
 
-const errUnsupportedClaimSpec = "composite resource claim spec was not an object"
-
-// Configure the supplied composite resource. The composite resource name is
-// derived from the supplied claim, as {name}-{random-string}. The claim's
+// ConfigureComposite configures the supplied composite resource. The composite resource name
+// is derived from the supplied claim, as {name}-{random-string}. The claim's
 // external name annotation, if any, is propagated to the composite resource.
-func Configure(_ context.Context, cm resource.CompositeClaim, cp resource.Composite) error {
+func ConfigureComposite(_ context.Context, cm resource.CompositeClaim, cp resource.Composite) error {
 	// It's possible we're being asked to configure a statically provisioned
 	// composite resource in which case we should respect its existing name and
 	// external name.
@@ -74,20 +75,21 @@ func Configure(_ context.Context, cm resource.CompositeClaim, cp resource.Compos
 		return nil
 	}
 
-	// TODO(negz): Consider nesting all user-specified spec values under a
-	// predictable object like spec.parameters so we can propagate _only_ user
-	// specified fields. Maintaining a set of keys to delete here seems error
-	// prone. Note that deleting these keys isn't always necessary; the CRD will
-	// drop them if preserveUnknownFields is false, but that is not the default
-	// for pre-v1 CRDs.
-	i, _ := fieldpath.Pave(ucm.Object).GetValue("spec")
-	spec, ok := i.(map[string]interface{})
+	icmSpec := ucm.Object["spec"]
+	spec, ok := icmSpec.(map[string]interface{})
 	if !ok {
 		return errors.New(errUnsupportedClaimSpec)
 	}
 
-	_ = fieldpath.Pave(ucp.Object).SetValue("spec", filter(spec, xcrd.FilterClaimSpecProps...))
+	// Delete base claim fields when configuring composite spec
+	baseClaimSpec := xcrd.CompositeResourceClaimSpecProps()
 
+	// keep any necessary fields between claim and composite
+	for _, field := range xcrd.KeepClaimSpecProps {
+		delete(baseClaimSpec, field)
+	}
+	claimSpecFilter := xcrd.GetPropFields(baseClaimSpec)
+	ucp.Object["spec"] = filter(spec, claimSpecFilter...)
 	return nil
 }
 
@@ -106,4 +108,94 @@ func filter(in map[string]interface{}, keys ...string) map[string]interface{} {
 		out[k] = v
 	}
 	return out
+}
+
+// APIClaimConfigurator configures the supplied claims with fields
+// from the composite. This includes late-initializing spec values
+// and updating status fields in claim.
+type APIClaimConfigurator struct {
+	client client.Client
+}
+
+// NewAPIClaimConfigurator returns a APIClaimConfigurator.
+func NewAPIClaimConfigurator(client client.Client) *APIClaimConfigurator {
+	return &APIClaimConfigurator{client: client}
+}
+
+// Configure the supplied claims with fields from the composite.
+// This includes late-initializing spec values and updating status fields in claim.
+func (c *APIClaimConfigurator) Configure(ctx context.Context, cr resource.CompositeClaim, cp resource.Composite) error {
+	ucr, ok := cr.(*claim.Unstructured)
+	if !ok {
+		return nil
+	}
+	ucp, ok := cp.(*composite.Unstructured)
+	if !ok {
+		return nil
+	}
+
+	if err := merge(ucr.Object["status"], ucp.Object["status"],
+		// Status fields from composite overwrite non-empty fields in claim
+		withMergeOptions(mergo.WithOverride),
+		withSrcFilter(xcrd.GetPropFields(xcrd.CompositeResourceStatusProps())...)); err != nil {
+		return errors.Wrap(err, errMergeClaimStatus)
+	}
+
+	if err := c.client.Status().Update(ctx, cr); err != nil {
+		return errors.Wrap(err, errUpdateClaimStatus)
+	}
+
+	if err := merge(ucr.Object["spec"], ucp.Object["spec"],
+		withSrcFilter(xcrd.GetPropFields(xcrd.CompositeResourceSpecProps())...)); err != nil {
+		return errors.Wrap(err, errMergeClaimSpec)
+	}
+
+	return errors.Wrap(c.client.Update(ctx, cr), errUpdateClaim)
+}
+
+type mergeConfig struct {
+	mergeOptions []func(*mergo.Config)
+	srcfilter    []string
+}
+
+// withMergeOptions allows custom mergo.Config options
+func withMergeOptions(opts ...func(*mergo.Config)) func(*mergeConfig) {
+	return func(config *mergeConfig) {
+		config.mergeOptions = opts
+	}
+}
+
+// withSrcFilter filters supplied keys from src map before merging
+func withSrcFilter(keys ...string) func(*mergeConfig) {
+	return func(config *mergeConfig) {
+		config.srcfilter = keys
+	}
+}
+
+// merge a src map into dst map
+func merge(dst, src interface{}, opts ...func(*mergeConfig)) error {
+	if dst == nil || src == nil {
+		// Nothing available to merge if dst or src are nil.
+		// This can occur early on in reconciliation when the
+		// status subresource has not been set yet.
+		return nil
+	}
+
+	config := &mergeConfig{}
+
+	for _, opt := range opts {
+		opt(config)
+	}
+
+	dstMap, ok := dst.(map[string]interface{})
+	if !ok {
+		return errors.New(errUnsupportedDstObject)
+	}
+
+	srcMap, ok := src.(map[string]interface{})
+	if !ok {
+		return errors.New(errUnsupportedSrcObject)
+	}
+
+	return mergo.Merge(&dstMap, filter(srcMap, config.srcfilter...), config.mergeOptions...)
 }
--- a/internal/controller/apiextensions/claim/configurator_test.go
+++ b/internal/controller/apiextensions/claim/configurator_test.go
@@ -25,6 +25,7 @@ import (
 	"github.com/pkg/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/crossplane/crossplane-runtime/pkg/meta"
 	"github.com/crossplane/crossplane-runtime/pkg/resource"
@@ -35,7 +36,7 @@ import (
 	"github.com/crossplane/crossplane/internal/xcrd"
 )
 
-func TestConfigure(t *testing.T) {
+func TestCompositeConfigure(t *testing.T) {
 	ns := "spacename"
 	name := "cool"
 	now := metav1.Now()
@@ -152,6 +153,10 @@ func TestConfigure(t *testing.T) {
 							"spec": map[string]interface{}{
 								"coolness": 23,
 
+								// These should be preserved.
+								"compositionRef":      "ref",
+								"compositionSelector": "ref",
+
 								// These should be filtered out.
 								"resourceRef":                "ref",
 								"writeConnectionSecretToRef": "ref",
@@ -173,7 +178,9 @@ func TestConfigure(t *testing.T) {
 								},
 							},
 							"spec": map[string]interface{}{
-								"coolness": int64(23),
+								"coolness":            23,
+								"compositionRef":      "ref",
+								"compositionSelector": "ref",
 							},
 						},
 					},
@@ -229,7 +236,7 @@ func TestConfigure(t *testing.T) {
 							"spec": map[string]interface{}{
 								// This should be overridden with the value of
 								// the equivalent claim field.
-								"coolness": int64(42),
+								"coolness": 42,
 							},
 						},
 					},
@@ -256,7 +263,7 @@ func TestConfigure(t *testing.T) {
 								},
 							},
 							"spec": map[string]interface{}{
-								"coolness": int64(23),
+								"coolness": 23,
 							},
 						},
 					},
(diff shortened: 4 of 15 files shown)


internal/controller/apiextensions/composite/composed.go
```diff
@@ -133,6 +135,48 @@ func (r *APIDryRunRenderer) Render(ctx context.Context, cp resource.Composite, c
...
+// NewAPIClaimConfigurator returns a APIClaimConfigurator that uses
+// the API server to get or patch the referenced composite claim.
+func NewAPIClaimConfigurator(c client.Client) *APIClaimConfigurator {
+	return &APIClaimConfigurator{client: c}
+}
+
+// GetClaim returns the composite claim referenced from the composite or nil if no
+// claim is referenced.
+func (r *APIClaimConfigurator) GetClaim(ctx context.Context, cp resource.Composite) (resource.CompositeClaim, error) {
+	claimRef := cp.GetClaimReference()
```
mcavoyk, line 153: One oddity my testing found was that this is not currently guaranteed to be set for all composites which have a claim. Unsure of whether this is a bug in the claim controller or the assumptions here, but the initial creation of a composite from the claim controller does not have the claim ref set and is instead updated later in the claim reconcile.

This leads to an extra reconcile by the claim and composite on initial claim creation.

internal/controller/apiextensions/composite/reconciler.go
```diff
@@ -399,6 +471,17 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error)
 			return reconcile.Result{RequeueAfter: shortWait}, nil
 		}
 
+		// DeepCopy composed resource to ensure changes to composed are not persisted
+		// when fetching connection details or readiness
+		cdCopy := cd.DeepCopyObject()
+		for _, p := range comp.Spec.Resources[i].Patches {
+			if err := p.Apply(patchObject, cdCopy); err != nil {
```
mcavoyk, line 478: Patches from composed to composite currently can take up to 60s to apply due to the long wait reconcile. One option if this is not desirable is we could detect if the used composition has any patches which may patch the composite and shorten the longWait reconcile for these resources.

negz:
Thanks for working on this @mcavoyk!

internal/controller/apiextensions/composite/reconciler.go
```diff
@@ -399,6 +471,17 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error)
...
+				log.Debug(fmt.Sprintf(errFmtPatch, i), "error", err)
+				r.record.Event(cr, event.Warning(reasonCompose, err))
+				return reconcile.Result{RequeueAfter: shortWait}, nil
+			}
+		}
```
negz, line 483: What do you think about:

1. Moving this patch loop into a new `Renderer`
1. Adding an optional patch type filter to `p.Apply`.

This would keep all of the patch application at the same 'scope' (e.g. in methods called by the reconciler, rather than partly in the reconciler and partly in said methods) and allow us to avoid spending time calculating patches that we don't need.

I'm thinking something like:

```go
// Apply executes a patching operation between the from and to resources.
// If applies all patch types unless an 'only' filter is supplied.
func (c *Patch) Apply(from, to runtime.Object, only ...PatchType) error {
	// ...
}
```

With the reconciler being something like:

```go
// Render each composed resource - applies only the composite -> composed patches.
for i := range refs {
	cd := composed.New(composed.FromReference(refs[i]))
	_ = r.composed.Render(ctx, cr, cd, comp.Spec.Resources[i])
}

// Persist references
cr.SetResourceReferences(refs)
_ := r.client.Update(ctx, cr)

// Similar to the current loop, but we also 're-render' the composite resource based on each
// composed resource.
for i, cd := range cds {
	_ = r.client.Apply(ctx, cd, resource.MustBeControllableBy(cr.GetUID()))
	c, _ := r.composed.FetchConnectionDetails(ctx, cd, comp.Spec.Resources[i])
	rdy, _ := r.composed.IsReady(ctx, cd, comp.Spec.Resources[i])

	// Note composite.Render, rather than composed.Render. Applies only the
	// composed -> composite patches.
	_ = r.composite.Render(ctx, cr, cd, comp.Spec.Resources[i])

}

```

internal/controller/apiextensions/composite/reconciler.go
```diff
@@ -384,12 +443,25 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error)
 	}
 
 	cr.SetResourceReferences(refs)
-	if err := r.client.Update(ctx, cr); err != nil {
-		log.Debug(errUpdate, "error", err)
+
+	// Create generic runtime object to apply patches from composed resources to either composite
+	// or composite claim. If a claimRef is specified, the composite claim must be patched rather than
+	// the composite, as the composite claim is the owner of the composite specification.
```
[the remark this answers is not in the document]
mcavoyk: This was my initial thought as well, but it was not clear to me how the claim reconciler could differentiate between inputs on the claim which should be propagated to the composite, and patches from the composed to the composite which should be propagated to the claim. Server side apply was the main idea I considered but that has its own drawbacks in-terms of API server version requirements.

Updating the claim if one exists allows the claim to remain the source of truth for parameters.

internal/controller/apiextensions/composite/reconciler.go
mcavoyk: A composite render is a good idea, I tried to re-use the composite render initially but that path seemed unsuccessful 😄

internal/controller/apiextensions/composite/reconciler.go
```diff
@@ -425,6 +508,22 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error)
...
+		return reconcile.Result{RequeueAfter: shortWait}, nil
+	}
+
+	// Update composite with resource reference changes and potential
+	// patches from composed resource
+	if err := r.client.Patch(ctx, cr, client.MergeFrom(crActual)); err != nil {
+		log.Debug(errUpdate, "error", err)
+		r.record.Event(cr, event.Warning(reasonCompose, err))
+		return reconcile.Result{RequeueAfter: shortWait}, nil
+	}
```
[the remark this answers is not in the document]
mcavoyk: Since the reconcile already performs a Get on the composite, it seemed like a patch with a merge diff offered a smaller patch to reduce field conflict between composite and claim reconciler.

internal/controller/apiextensions/composite/reconciler.go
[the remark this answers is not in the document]
negz: > This was my initial thought as well, but it was not clear to me how the claim reconciler could differentiate between inputs on the claim which should be propagated to the composite, and patches from the composed to the composite which should be propagated to the claim.

Yeah I agree this is tricky, but I'd like to do a bit more investigation to see if it's possible since it would allow us to keep a clean separation of responsibilities (i.e. allow the composite reconciler to remain blissfully unaware of claims).

One naive (but relatively simple) approach might be to assume that `spec` is propagated claim -> composite, but `status` is propagated composite -> claim. I suspect bidirectional patching will be most often used to derive the status of the composite resource from the statuses of its composed resources.

Otherwise (possibly a dumb idea - haven't thought this through) could we just patch from the claim to the composite, then patch from the composite back to the claim? Presumably the first patch would ensure the composite was updated with the claim's desired state. 🤔 

> Server side apply was the main idea I considered but that has its own drawbacks in-terms of API server version requirements.

We might consider making the cut if it would help - we had been holding off due to the API server version requirements but if I recall correctly we might already have those requirements due to having migrated to v1 CRDs (iirc both v1 CRDs and SSA require v1.16).

internal/controller/apiextensions/composite/reconciler.go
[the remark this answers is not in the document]
negz: Got it. Though it occurs to me now that we were just calling `Update` before, since we should keep a relatively fresh view on the composite throughout the reconcile. Did you go with Patch to avoid resource version conflicts?

internal/controller/apiextensions/composite/reconciler.go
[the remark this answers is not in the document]
mcavoyk: > One naive (but relatively simple) approach might be to assume that spec is propagated claim -> composite, but status is propagated composite -> claim. I suspect bidirectional patching will be most often used to derive the status of the composite resource from the statuses of its composed resources.

I created https://github.com/crossplane/crossplane/issues/2081 to track that currently status fields in XRD's are overridden with the default claim/composite status. 

Although that approach would be simpler, propagating back values which were transformed in compositions or values which may have different defaults per composition (rather than the default in XRD) back into the spec seems valuable.

> Otherwise could we just patch from the claim to the composite, then patch from the composite back to the claim? Presumably the first patch would ensure the composite was updated with the claim's desired state. 

If some field's initial value is a zero value, and receives a value from a patch from the composed to the composite, the claim reconciler could patch this value back. But now future reconciles on the claim controller can't differentiate between spec drift between fields on the claim -> composite or patches which need to be propagated back to the claim.

I think this approach could work with server-side apply, my thinking is the composite controller does a server-side apply with force for changes on itself, the claim controller can create a patch diff between the claim and the composite and apply without force this diff on the composite. This would allow spec changes which are not affected by bidirectional patches to persist to the composite. Then the claim is updated based on the results from patching the composite which would be a merge of claim spec and bi-directional patch.

Not sure if that works or makes sense 😄 , which is why my initial draft implementation avoided this and just had the composite reconciler patch the claim instead if it existed.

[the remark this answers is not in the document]
negz: > If some field's initial value is a zero value, and receives a value from a patch from the composed to the composite, the claim reconciler could patch this value back. But now future reconciles on the claim controller can't differentiate between spec drift between fields on the claim -> composite or patches which need to be propagated back to the claim.

Hrm, this is making me think that we might want to treat status as a special case.

In general we follow the [late initialization](https://github.com/kubernetes/community/blob/1d2f05/contributors/devel/sig-architecture/api-conventions.md#late-initialization) convention across our managed resources - i.e. the managed resource is always the source of truth, but we will update a spec field if it is set in the external API (i.e. the cloud provider) and not set in the managed resource. Once the field is set on the managed resource it becomes the source of truth and we never touch it.

I think it makes sense for us to do something similar with XRs - if we only allow composed resources to patch to fields that are not currently set we ensure the XR (and thus claim) are always the source of truth. This pattern would fall apart if applied to status fields though - the source of truth is flipped there and thus you typically want the composed values to overwrite whatever is at the XR/claim level.

I'm wondering if this behaviour should be configurable at the XRD level? i.e. Perhaps we allow only late initialization by default, but you can supply an array of field paths (which would default to only `status`) under which we assume propagation should _always_ happen in reverse - i.e. `composed -> XR -> claim` - even when doing so would overwrite existing values.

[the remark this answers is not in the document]
mcavoyk: Hmm yeah late initialization is an interesting way to consider this. If the composite could late initialize spec values with patches from composed, and always own status fields this would solve most of our use-cases around bidirectional patches.

This also solves some of the issues I've been thinking about here where cycles could occur between spec input on the claim and a composition patch which overrode that same field. Without a clear source of truth the claim controller and composite controller would just keep patching that field back and forth.

muvaf:
@negz @mcavoyk have you guys considered having this at the top level alongside `patches`? Like `propagations`? Personally, I would find it easier to understand if I know which way the i/o flow is at a higher level since the field names in the structs are pretty similar (`fromFieldPath`/`toFieldPath`) already.

mcavoyk:
> Have you guys considered having this at the top level alongside `patches`? Like `propagations`? Personally, I would find it easier to understand if I know which way the i/o flow is at a higher level since the field names in the structs are pretty similar (`fromFieldPath`/`toFieldPath`) already.

I agree that `FromCompositeFieldPath` and `FromComposedFieldPath` are too similiar which hurts readability of compositions. I'm not sure I see exactly how this work at a top level, but `FromCompositeFieldPath` and `ToCompositeFieldPath` might read better.

negz:
> have you guys considered having this at the top level alongside patches? Like propagations? Personally, I would find it easier to understand if I know which way the i/o flow is at a higher level since the field names in the structs are pretty similar (fromFieldPath/toFieldPath) already.

I think we discussed it a little on https://github.com/crossplane/crossplane/issues/1639. I personally prefer it just being a patch type rather than a distinct array of patches, but I feel like this is a pretty subjective thing. I'd struggle to name the new array - I've heard `outputs` and `propagations` suggested but to me they're all still _patches_, just with the source and destination inverted.

> I agree that FromCompositeFieldPath and FromComposedFieldPath are too similiar which hurts readability of compositions. I'm not sure I see exactly how this work at a top level, but FromCompositeFieldPath and ToCompositeFieldPath might read better.

This naming seems reasonable to me - I agree it's easier to distinguish at a glance.

apis/apiextensions/v1/composition_types_test.go
```diff
@@ -570,14 +570,70 @@ func TestPatchApply(t *testing.T) {
...
+				err: nil,
+			},
+		},
+		"ValidComposedFieldPathPatch": {
+			reason: "Should correctly apply a ComposedFieldPathPatch with valid settings",
+			args: args{
+				patch: Patch{
+					Type:          PatchTypeFromComposedFieldPath,
+					FromFieldPath: pointer.StringPtr("objectMeta.labels"),
+					ToFieldPath:   pointer.StringPtr("objectMeta.labels"),
```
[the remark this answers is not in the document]
mcavoyk: `objectMeta.labels` seems correct here because the fake structs do not have struct tags on these fields.

internal/controller/apiextensions/claim/configurator.go
```diff
@@ -74,20 +85,20 @@ func Configure(_ context.Context, cm resource.CompositeClaim, cp resource.Compos
...
+	spec, ok := icmSpec.(map[string]interface{})
 	if !ok {
 		return errors.New(errUnsupportedClaimSpec)
 	}
 
-	_ = fieldpath.Pave(ucp.Object).SetValue("spec", filter(spec, xcrd.FilterClaimSpecProps...))
+	// Delete base claim fields when configuring composite spec
+	baseClaimSpec := xcrd.CompositeResourceClaimSpecProps()
+	// keep composition selection fields
+	delete(baseClaimSpec, "compositionRef")
```
mcavoyk, line 97: It seemed slightly clearer here to filter all keys and then specifically not filter the two composition selection fields.

negz: This seems reasonable, but it did take me a few reads to wrap my head around. Perhaps a slightly more detailed comment before this logic would be useful? That said, I just tried writing one up and it didn't say much more than your commentary here already does.

internal/controller/apiextensions/claim/configurator.go
```diff
@@ -74,20 +85,20 @@ func Configure(_ context.Context, cm resource.CompositeClaim, cp resource.Compos
...
+	delete(baseClaimSpec, "compositionSelector")
```
negz, line 98: What do you think about using something like `xcrd.KeepClaimSpecProps := []string{"compositionRef", "compositionSelector"}`? This might help with readability a bit, and keep the set of fields that we want to keep / propagate close to where the schemas are defined in the xcrd package, similar to the previous `xcrd.FilterClaimSpecProps`.

internal/controller/apiextensions/claim/configurator.go
```diff
@@ -107,3 +118,87 @@ func filter(in map[string]interface{}, keys ...string) map[string]interface{} {
...
+
+	if err := merge(ucr.Object["spec"], ucp.Object["spec"],
+		withSrcFilter(xcrd.GetPropFields(xcrd.CompositeResourceSpecProps())...)); err != nil {
+		return errors.Wrap(err, errMergeClaimSpec)
+	}
+
+	return errors.Wrap(c.client.Update(ctx, cr), errUpdateClaim)
+}
+
+// Note(mcavoyk): merge package might be a better fit in crossplane-runtime
```
negz, line 165: Historically I'd agree with you, but we have a few libs in runtime now that are _only_ used by the Composition controllers and it's kind of a pain to have to update runtime just to get something done in core. I'm starting to come around to the idea that we shouldn't put things in runtime until they have more than one consumer (e.g. they're used by multiple providers).

internal/controller/apiextensions/claim/configurator_test.go
```diff
@@ -278,3 +286,319 @@ func TestConfigure(t *testing.T) {
...
+
+	for name, tc := range cases {
+		t.Run(name, func(t *testing.T) {
+			c := NewAPIClaimConfigurator(tc.args.client)
+			got := c.Configure(context.Background(), tc.args.cm, tc.args.cp)
+			if diff := cmp.Diff(tc.want.err, got, test.EquateErrors()); diff != "" {
+				t.Errorf("b.Bind(...): %s\n-want error, +got error:\n%s\n", tc.reason, diff)
+			}
+			if diff := cmp.Diff(tc.want.cm, tc.args.cm); diff != "" {
+				t.Errorf("b.Bind(...): %s\n-want, +got:\n%s\n", tc.reason, diff)
```
negz, line 599: Looks like a copy paste issue in these error strings. :)

internal/controller/apiextensions/claim/reconciler.go
```diff
@@ -60,10 +60,11 @@ const (
...
 // Event reasons.
 const (
-	reasonBind      event.Reason = "BindCompositeResource"
-	reasonDelete    event.Reason = "DeleteCompositeResource"
-	reasonConfigure event.Reason = "ConfigureCompositeResource"
-	reasonPropagate event.Reason = "PropagateConnectionSecret"
+	reasonBind               event.Reason = "BindCompositeResource"
+	reasonDelete             event.Reason = "DeleteCompositeResource"
+	reasonCompositeConfigure event.Reason = "ConfigureCompositeResource"
+	reasonClaimConfigure     event.Reason = "COnfigureClaimResource"
```
negz, line 66:
```suggestion
	reasonClaimConfigure     event.Reason = "ConfigureClaim"
```

negz: We tend to just call it a "claim", "composite resource claim", or "XRC" rather than a "claim resource".

internal/controller/apiextensions/claim/reconciler.go
```diff
@@ -384,6 +393,13 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error)
 		return reconcile.Result{RequeueAfter: aShortWait}, errors.Wrap(r.client.Status().Update(ctx, cm), errUpdateClaimStatus)
 	}
 
+	if err := r.claim.Configure(ctx, cm, cp); err != nil {
+		log.Debug("Cannot configure claim resource", "error", err, "requeue-after", time.Now().Add(aShortWait))
```
negz, line 397:
```suggestion
		log.Debug("Cannot configure composite resource claim", "error", err, "requeue-after", time.Now().Add(aShortWait))
```

internal/controller/apiextensions/composite/composed.go
```diff
@@ -133,6 +133,27 @@ func (r *APIDryRunRenderer) Render(ctx context.Context, cp resource.Composite, c
 	return errors.Wrap(r.client.Create(ctx, cd, client.DryRunAll), errName)
 }
 
+// An CompositeRenderer renders the composite resource
+type CompositeRenderer struct{} // nolint: golint
```
negz, line 137: What was golint complaining about here?

internal/xcrd/crd.go
```diff
@@ -156,9 +167,20 @@ func ForCompositeResourceClaim(xrd *v1.CompositeResourceDefinition) (*extv1.Cust
 			specProps.Properties[k] = v
 		}
 		crd.Spec.Versions[i].Schema.OpenAPIV3Schema.Properties["spec"] = specProps
+
+		statusP, statusRequired, err := getProps("status", vr.Schema)
+		if err != nil {
+			return nil, errors.Wrap(err, errGetSpecProps)
```
negz, line 173: Should we replace `errGetSpecProps` with something that covers status (or arbitrary fields)?

internal/controller/apiextensions/composite/reconciler.go
```diff
@@ -423,6 +426,25 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error)
...
+	if err := r.client.Status().Update(ctx, cr); err != nil {
+		log.Debug(errUpdate, "error", err)
+		r.record.Event(cr, event.Warning(reasonCompose, err))
+		return reconcile.Result{RequeueAfter: shortWait}, nil
+	}
+
+	if err := r.client.Update(ctx, cr); err != nil {
+		log.Debug(errUpdate, "error", err)
+		r.record.Event(cr, event.Warning(reasonCompose, err))
+		return reconcile.Result{RequeueAfter: shortWait}, nil
```
negz, line 447: You might need to test that this is working as expected. We've often had issues in the past with updating status and then updating the rest of an object. If I recall correctly this is because when you update the status the controller-runtime client loads the updated object from the API server, thus reverting any pending changes to the object's spec or metadata. :(

This is kind of a pain, because it might mean we can't neatly render out the new composite resource and then update both its spec and status. We may instead need to render the spec, persist that, then do the status, then persist that (or vice versa).

internal/controller/apiextensions/composite/composed.go
```diff
@@ -133,6 +133,27 @@ func (r *APIDryRunRenderer) Render(ctx context.Context, cp resource.Composite, c
...
+
+// NewCompositeRenderer returns a Renderer composite resource
+func NewCompositeRenderer() *CompositeRenderer {
+	return &CompositeRenderer{}
+}
+
+// Render the supplied composite resource using the supplied composed resource
+// and template.
+func (r *CompositeRenderer) Render(_ context.Context, cp resource.Composite, cd resource.Composed, t v1.ComposedTemplate) error {
+	compositePatches := []v1.PatchType{v1.PatchTypeToCompositeFieldPath}
```
negz, line 147: Nit: I'd be tempted to call this `only` to clarify how it's used by `p.Apply` below.

negz (commented):
This is looking really promising - thanks for persisting with this @mcavoyk!

internal/controller/apiextensions/composite/composed.go
mcavoyk: Package stutter, `composite.CompositeRender`, but the package also contains a few other renderers.

internal/controller/apiextensions/composite/reconciler.go
mcavoyk: I'll verify again with a debugger that this is being a persisted, but I thought I recall the proper order being status update, then update. The update itself I believe reverts status.

negz: I'll definitely be happy if I'm wrong - I like how the code is laid out now.

CC @muvaf you usually remember the details of this quirk better than I do.

muvaf: @negz your assessment is correct. Any update operation returns the object from the api-server and unmarshalls it into the object you passed, in this case `cr`. What we usually do is to issue a spec update if there is a change in spec and only then update the status.

mcavoyk: Yeah, after testing it seems either update, spec or status, resets the other sub-field to values from the api-server. In this case updating status first works properly because each sub-field is being set first. So status is set then updated, then spec is set and updated.

negz: @mcavoyk Apologies if I'm belabouring the point here, but I'm still struggling to grok how this is working.

Reading the code I would expect:

1. We iterate over all composed resources calling `r.composite.Render` which potentially patches from the composed resource to either the status or the spec of the composite resource (i.e. spec fields that a claim - if any - will allow us to late init).
2. We update the status, which resets any pending spec changes from 1.
3. Any pending spec changes were reset in 2, so this is always a no-op.

What am I missing?

mcavoyk: My mistake, I thought this was reviewing the section in the claim controller which updates status and spec of the claim. Yes from looking at this, it must be only working in my tests after multiple reconciles. This should be updated to preserve the values in status, update the spec, then update status instead of the current logic.

mcavoyk: This should be resolved now, one of the reasons I did not notice this was spec propagated in my testing was from the base composed, and not a later initialization of the composed. Before, spec updates to composite were not applied and lost when the status was updated.

internal/controller/apiextensions/composite/reconciler.go
```diff
@@ -423,6 +426,27 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error)
...
+	// Update composite spec and status with patches from composed resources
+	// DeepCopy composite to preserve rendered status
+	crSpecObject := cr.DeepCopyObject()
+	if err := r.client.Update(ctx, crSpecObject); err != nil {
+		log.Debug(errUpdate, "error", err)
+		r.record.Event(cr, event.Warning(reasonCompose, err))
+		return reconcile.Result{RequeueAfter: shortWait}, nil
+	}
+
+	if err := r.client.Status().Update(ctx, cr); err != nil {
```
negz, line 446: Have you confirmed this works? I would expect this update to always fail due to a stale `metadata.resourceVersion`. i.e.

1. `cr` is at `metadata.resourceVersion` `N`.
2. We make a deepcopy of `cr`.
3. We call `client.Update()`, passing in the deepcopy. This causes the API server to bump its expected resource version to `N+1`.
4. We call `client.Status().Update()`, passing in the original `cr`. Its `metadata.resourceVersion` is still `N`, so I would expect the API server to reject this update with an error like "the object has been modified; please apply your changes to the latest version and try again".

mcavoyk: If this spec update occurs, then later status updates will receive `resourceVersion` conflicts, but these occur when the spec has been updated, so this will lead to an another reconciliation of the resource and eventual consistency.

We discussed this some on slack - but because the composite needs both the spec and status updated here, the options are to update one sub-field and overwrite the resourceVersion for the other update, or handle a second reconcile. Choosing the path of a further reconcile leaves slightly clearer logic for the controller, so I've updated the PR and removed the explicit status update here as most* of the paths in the remaining reconcile logic leads to to updating the status regardless.

go.mod
```diff
@@ -6,11 +6,12 @@ require (
 	github.com/Masterminds/semver v1.5.0
 	github.com/alecthomas/kong v0.2.11
 	github.com/alecthomas/units v0.0.0-20190924025748-f65c72e2690d // indirect
-	github.com/crossplane/crossplane-runtime v0.12.1-0.20210121145519-c674db318cdb
+	github.com/crossplane/crossplane-runtime v0.12.1-0.20210128000612-240e79c8a694
```
hasheddan, line 9: Note: this fixes a bug that was introduced in the bump to controller-runtime v0.8.0 that was fixed in https://github.com/crossplane/crossplane-runtime/pull/238

negz (commented):
I tested this out in a branch (including my small proposed change) and it seems to be working as expected, but I'm noticing (spurious?) warnings on the claim reconciler:

```console
Events:
  Type     Reason                      Age                     From                                                             Message
  ----     ------                      ----                    ----                                                             -------
  Warning  ConfigureClaim              9m20s                   offered/compositeresourcedefinition.apiextensions.crossplane.io  unable to merge claim status: destination object was not valid object
  Warning  ConfigureClaim              9m20s (x5 over 9m20s)   offered/compositeresourcedefinition.apiextensions.crossplane.io  unable to merge claim status: source object was not valid object
  Warning  ConfigureCompositeResource  9m20s                   offered/compositeresourcedefinition.apiextensions.crossplane.io  cannot patch object: Operation cannot be fulfilled on compositepostgresqlinstances.database.example.org "my-db-qfqjh": the object has been modified; please apply your changes to the latest version and try again
  Normal   ConfigureCompositeResource  8m49s (x12 over 9m20s)  offered/compositeresourcedefinition.apiextensions.crossplane.io  Successfully applied composite resource
  Normal   BindCompositeResource       8m49s (x6 over 9m19s)   offered/compositeresourcedefinition.apiextensions.crossplane.io  Composite resource is not yet ready
```

I tested using the GCP example from getting started, with slight modifications:

```yaml
---
metadata:
spec:
  group: database.example.org
  names:
  claimNames:
    kind: PostgreSQLInstance
    plural: postgresqlinstances
  connectionSecretKeys:
    - username
    - password
    - endpoint
    - port
  versions:
  - name: v1alpha1
    served: true
    referenceable: true
    schema:
      openAPIV3Schema:
        type: object
        properties:
          spec:
            type: object
            properties:
              parameters:
                type: object
                properties:
                  storageGB:
                    type: integer
                  diskType:
                    type: string
                required:
                  - storageGB
            required:
              - parameters
          status:
            type: object
            properties:
              selfLink:
                type: string
---
kind: Composition
metadata:
  labels:
    provider: gcp
    guide: quickstart
spec:
  compositeTypeRef:
  resources:
    - base:
        kind: CloudSQLInstance
        spec:
          forProvider:
            databaseVersion: POSTGRES_9_6
            region: us-central1
            settings:
              tier: db-custom-1-3840
              ipConfiguration:
                ipv4Enabled: true
                authorizedNetworks:
                  - value: "0.0.0.0/0"
          writeConnectionSecretToRef:
            namespace: crossplane-system
      patches:
          fromFieldPath: "metadata.uid"
          transforms:
            - type: string
              string:
                fmt: "%s-postgresql"
          fromFieldPath: "spec.parameters.storageGB"
          fromFieldPath: "spec.parameters.diskType"
          toFieldPath: "spec.forProvider.settings.dataDiskType"
        - type: ToCompositeFieldPath
          fromFieldPath: "status.atProvider.selfLink"
          toFieldPath: "status.selfLink"
        - type: ToCompositeFieldPath
          fromFieldPath: "spec.forProvider.settings.dataDiskType"
          toFieldPath: "spec.parameters.diskType"
      connectionDetails:
        - name: port
          value: "5432"
---
kind: PostgreSQLInstance
metadata:
  name: my-db
  namespace: default
spec:
  parameters:
    storageGB: 20
  compositionSelector:
    matchLabels:
      provider: gcp
  writeConnectionSecretToRef:
    name: db-conn
```

internal/controller/apiextensions/composite/reconciler.go
```diff
@@ -423,6 +427,21 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+	}
+
+	// Update composite with patches from composed resources
+	// DeepCopy composite to preserve rendered status
+	crSpecObject := cr.DeepCopyObject().(client.Object)
+	if err := r.client.Update(ctx, crSpecObject); err != nil {
+		log.Debug(errUpdate, "error", err)
+		r.record.Event(cr, event.Warning(reasonCompose, err))
+		return reconcile.Result{RequeueAfter: shortWait}, nil
 	}
```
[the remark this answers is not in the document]
mcavoyk: This seems reasonable, after more thought I'm curious what your opinions are on overwriting the `resourceVersion` here is. The spec is already updated here, so we are not forcing changes on legitimate changes, the `resourceVersion` overwrite would just be against any other status changes, which the controller should already own.

Other than overwriting, the early exit makes sense as another option to remove unneeded logging/events.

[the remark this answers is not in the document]
negz: I (very) slightly prefer the early exit because it eliminates the small chance we'd accidentally reset any status changes that had been persisted to the API server (by another party) if we overwrite the resource version.

For example:

1. We make a deepcopy.
2. Some other entity updates the XR's status.
3. We update the spec and metadata using our deepcopy.
4. We copy the deepcopy's resource version to our original object
5. We update the status using our (now stale) original object, thus overwriting the status changes.

That all said, I hadn't considered until you mentioned it that only status would be affected and in practice we're the only controller that updates status - so this is much less of a big deal than resetting a spec update for example. We do have some instances of multiple controllers updating status (e.g. two different controllers update XRD status) but it is admittedly quite an edge case.

I'd typically optimise for simplicity and readability over correctness in cases like this in which the correctness issue is so trivial, but to my mind both options are of similar complexity/readability.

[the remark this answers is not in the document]
mcavoyk: Makes sense, early exit is what I've gone with here after your suggestion 👍

internal/controller/apiextensions/composite/reconciler.go
```diff
@@ -423,6 +427,21 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		if rdy {
 			ready++
 		}
+
+		if err := r.composite.Render(ctx, cr, cd, comp.Spec.Resources[i]); err != nil {
+			log.Debug(errRenderCR, "error", err)
+			r.record.Event(cr, event.Warning(reasonCompose, err))
+			return reconcile.Result{RequeueAfter: shortWait}, nil
+		}
```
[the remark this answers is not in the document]
mcavoyk: Added

negz (approved):
Awesome work on this PR @mcavoyk! You bit off a challenging first PR and implemented it really well - we appreciate it!

muvaf:
Thanks a lot @mcavoyk ! That's a big contribution!

PR #2082, opened 2021-01-18
Update controller-runtime to v0.8.0

khos2ow:

This updates controller-runtime to v0.8.0 and corresponding
crossplane-runtime to latest version.

This PR depends on https://github.com/crossplane/crossplane-runtime/pull/210.

I have:

TBA


internal/controller/apiextensions/claim/reconciler.go
```diff
@@ -252,14 +252,14 @@ func NewReconciler(m manager.Manager, of resource.CompositeClaimKind, with resou
...
-func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error) { // nolint:gocyclo
+func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) { // nolint:gocyclo
 	// NOTE(negz): This method is well over our cyclomatic complexity goal.
 	// Be wary of adding additional complexity.
 
 	log := r.log.WithValues("request", req)
 	log.Debug("Reconciling")
 
-	ctx, cancel := context.WithTimeout(context.Background(), reconcileTimeout)
+	ctx, cancel := context.WithTimeout(ctx, reconcileTimeout)
```
khos2ow, line 262: A `context` is being passed to `Reconcile()` and that is being used instead of `context.Background()`.

internal/controller/pkg/revision/establisher.go
```diff
@@ -79,7 +79,7 @@ func (e *APIEstablisher) Establish(ctx context.Context, objs []runtime.Object, p
 
 		// Make a copy of the desired object to be populated with existing
 		// object, if it exists.
-		current := res.DeepCopyObject()
+		current := d
 		err := e.client.Get(ctx, types.NamespacedName{Name: d.GetName(), Namespace: d.GetNamespace()}, current)
```
khos2ow, line 83: `e.client.Get` has been refactored and now expects a `client.Object` rather than `runtime.Object` as last argument, as such using `res.DeepCopyObject()` (which returns a `runtime.Object`) is not possible. I've kept `current` around because it's used in line `103` too, but in reality both of them can be changed to use `d` and `current` can be removed altogether.

khos2ow (commented):
There are two notable changes in this PR (see below).

go.mod
```diff
@@ -5,29 +5,22 @@ go 1.13
 require (
 	github.com/Masterminds/semver v1.5.0
 	github.com/alecthomas/kong v0.2.11
-	github.com/alecthomas/template v0.0.0-20190718012654-fb15b899a751 // indirect
 	github.com/alecthomas/units v0.0.0-20190924025748-f65c72e2690d // indirect
 	github.com/crossplane/crossplane-runtime v0.12.0
```
hasheddan, line 9: @khos2ow would you mind adding a replace to use @saschagrunert's fork from https://github.com/crossplane/crossplane-runtime/pull/210 so we can get this compiling?

khos2ow: Not at all, will do in a minute.

hasheddan:
@khos2ow crossplane-runtime PR has been merged so we should be able to update deps here now :)

khos2ow:
@hasheddan updated `go.mod` and rebased on master also (just in case).

internal/controller/pkg/revision/establisher.go
hasheddan: @khos2ow unfortunately, I don't think this can be simplified in this manner. `res` is an interface here, which may have an underlying pointer type. In that case, instantiating `current` with `d` and then "getting into" `current` will actually change the underlying data of `d` as well. It looks like we will have to attempt to assert `current` to `client.Object` after the deep copy.

Here is a small toy example to illustrate the situation: https://play.golang.org/p/VeK7BHiKsad

cluster/charts/crossplane/crds/pkg.crossplane.io_controllerconfigs.yaml
```diff
@@ -553,6 +553,18 @@ spec:
                         description: User is a SELinux user label that applies to the container.
                         type: string
                     type: object
+                  seccompProfile:
```
hasheddan, line 556: Note for others: this is a non-breaking API change to `ControllerConfig`. Seccomp was promoted to security context / pod security context from annotations upstream, which is now reflected here 👍

hasheddan (approved):
This looks good to me @khos2ow! Thanks :) Could you rebase on `master`?

hasheddan:
@khos2ow my apologies, looks like this slipped behind master after your rebase :/ could you update one more time?

khos2ow:
@hasheddan yeah sure, no worries, should be good now :)

hasheddan:
Thanks again @khos2ow !

PR #2087, opened 2021-01-20
Make dependency status fields optional in revisions

hasheddan:

In rare edge cases, having required status fields can cause problems
when trying to clean up old objects by manually removing finalizers.
kubectl is not able to edit the status subresource, meaning that if
these fields are not set then a user is stuck trying to remove the
finalizer because the patch is technically invalid. This would only be
the case if a user was updating from a pre-v1.0.0 Crossplane version and
had manually upgraded the revision CRDs to v1.

_**Note: this is technically an API change, but it is backwards compatible.**_

I have:

See https://crossplane.slack.com/archives/CEG3T90A1/p1611152392084200 for context on how this addresses the described issue.

### DIFF
--- a/apis/pkg/v1/revision_types.go
+++ b/apis/pkg/v1/revision_types.go
@@ -90,9 +90,9 @@ type PackageRevisionStatus struct {
 	ObjectRefs []xpv1.TypedReference `json:"objectRefs,omitempty"`
 
 	// Dependency information.
-	FoundDependencies     int64 `json:"foundDependencies"`
-	InstalledDependencies int64 `json:"installedDependencies"`
-	InvalidDependencies   int64 `json:"invalidDependencies"`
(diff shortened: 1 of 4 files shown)


jack1902:
@hasheddan Looks great, thanks for pushing this upstream and helping us get out of the mess we ended up in 😆

negz (approved):

mf-lit:
Hey @hasheddan thanks for squashing this so quickly.

> This would only be
> the case if a user was updating from a pre-v1.0.0 Crossplane version and
> had manually upgraded the revision CRDs to v1.

Out of interest, is there a recommended way we should have done this?

hasheddan:
@mf-lit we have a tiny guide for doing so, but I think y'all may have encountered an edge case https://crossplane.io/docs/v1.0/guides/upgrading-to-v1.0.html

Moving forward, we will not break any backwards compatibility, so this should not be an issue 👍

Issue #1481, opened 2020-05-01
Proposal: Introduce CompositionRevision


negz:
### What problem are you facing?

The [experimental composition design](https://github.com/crossplane/crossplane/pull/1163) introduces the concept of a Crossplane resource that composes other resources. Under this design there is [an open question](https://github.com/crossplane/crossplane/pull/1452#discussion_r417671831) around which resource(s) should represent the authoritative desired state for an external resource.

There are a few moving parts under the current design:

* A composite infrastructure resource. A (mostly) arbitrary Kubernetes custom resource that is roughly equivalent to a Helm values file or a Terraform module `inputs.tf` file. From the user's perspective the composite infrastructure resource _is_ the resource. A [`NetworkGroup`](https://github.com/crossplane/crossplane/blob/86d4d7/cluster/examples/composition/azure-sqlserver/networkgroup.yaml) is an example of a composite resource; it's a concept defined by an infrastructure operator.

* An optional infrastructure requirement. An infrastructure requirement is a lightweight proxy for a composite resource; their schemas are for all intents and purposes identical. Requirements exist to enable separation of concerns. Application operators may not typically operate directly on a composite infrastructure resource, but may operate on it indirectly via its corresponding infrastructure requirement. A [`PrivatePostgreSQLServerRequirement`](https://github.com/crossplane/crossplane/blob/86d4d/cluster/examples/composition/azure-sqlserver/privatepostgresqlserver.yaml) is an example of a requirement; it is a proxy for a `PrivatePostgreSQLServer` composite infrastructure resource.

* A `Composition`. A template used to "fill in the blanks" - to specify the "base layer" over which the composite resource's spec is overlaid to render one or more composed resources. A ` Composition` is analogous to a set of Helm moustache templates or a Terraform module `main.tf`. One novel difference between Crossplane and the aforementioned tools is that multiple compositions may be able to satisfy a particular kind of composite resource. [This composition](https://github.com/crossplane/crossplane/blob/86d4d7/cluster/examples/composition/azure-sqlserver/definitions/networkgroup.yaml#L64) can satisfy the earlier `NetworkGroup` composite resource.

* The composed infrastructure resources. The outputs of the composition process. These are either primitive (atomic) managed infrastructure resources, or nested composite infrastructure resources. They're analogous to the (custom) resources rendered by a Helm chart, or the Terraform resources rendered by a Terraform module.

[...]

PR #2090, opened 2021-01-22
Enable more kinds of updates to the Composition resource template array

negz:

Fixes https://github.com/crossplane/crossplane/issues/1909, which in turn unblocks https://github.com/crossplane/crossplane/issues/1481

Currently we use 'anonymous' resource templates. These templates rely on matching the order of the composition resource template array relative to the composite resource references array in order to match composed resources to the templates used to render them. This means we can't reorder the templates within a composition (or change to using a Composition with a different template order) without affecting composed resources.

This commit introduces optional 'named' templates. If all templates within a Composition are named Crossplane will use the names to relate templates to composed resources. This allows us to reorder templates, and to delete composed resources when their corresponding template is deleted.

I have:

I've spun up the contemporary "AWS with VPC" PostgreSQLInstance example from our docs, and validated that Compositions with anonymous resource templates work as they do today. I've then updated the Composition such that all templates have unique template names. When I do so the template names are propagated to the existing composed resources, and I can subsequently reorder and delete templates. When a delete a template, its corresponding composed resources are deleted.


internal/controller/apiextensions/composite/composed.go
```diff
@@ -238,3 +278,324 @@ func IsReady(_ context.Context, cd resource.Composed, t v1.ComposedTemplate) (bo
...
+			return CompositionResult{}, errors.Wrap(err, errReadiness)
+		}
+
+		if rdy {
+			ready++
+		}
+	}
+
+	return CompositionResult{ConnectionDetails: conn, DesiredResources: len(comp.Spec.Resources), ReadyResources: ready}, nil
+}
```
negz, line 464: Note that this logic is pretty much directly extracted from the current Reconcile loop.

internal/controller/apiextensions/composite/composed.go
```diff
@@ -39,16 +39,42 @@ import (
...
+// reconcile a composed resource as an annotation.
+func SetCompositionTemplateName(o metav1.Object, name string) {
+	meta.AddAnnotations(o, map[string]string{AnnotationKeyCompositionTemplateName: name})
+}
+
+// GetCompositionTemplateName gets the name of the composition template used to
+// reconcile a composed resource from its annotations.
+func GetCompositionTemplateName(o metav1.Object) string {
+	return o.GetAnnotations()[AnnotationKeyCompositionTemplateName]
+}
```
negz, line 76: I considered persisting the template name in the XR's resourceRefs array, rather than as an annotation on the composed resource. Technically doing so would be a violation of API conventions, which require that references be "a strict subset" of the ObjectReference type.

I feel like there was another reason to persist the name to the actual resource rather than the reference, but it's not immediately coming to mind.

negz:
I'm leaving this in draft for now because it needs more test coverage, but I want to get https://github.com/crossplane/crossplane/pull/2077 merged first so I can rebase on those changes.

I don't love how complex this implementation is, but we need to maintain our v1.0 behaviour and provide a supported path to migrate from anonymous to named templates. Some other options I considered but discarded were:

1. Use two different controllers; one for named and one for anonymous compositions. This would perhaps be cleaner, but would result in a lot of duplication and require us to remember to make changes to two different controllers.
1. Have a controller automatically generate missing template names in a Composition, allowing us to assume all Compositions use named templates. We'd still need to use something like the existing anonymous composer logic to propagate the template name to the composed resources, and any existing Compositions would get stuck with randomly generated template names if they upgraded from Crossplane 1.0 to 1.1.

mcavoyk:
> 2. Have a controller automatically generate missing template names in a Composition, allowing us to assume all Compositions use named templates. We'd still need to use something like the existing anonymous composer logic to propagate the template name to the composed resources, and any existing Compositions would get stuck with randomly generated template names if they upgraded from Crossplane 1.0 to 1.1.

This route could be improved by leaving a Crossplane version between which has the `templateName` field available, but unused so compositions could be updated if generated names was not desired. Also generated names could use the kind and index of that kind to for improved readability of the auto-generated `templateNames`, e.g. `cluster-0`, `vpc-0`, `subnet-0`, `subnet-1`.

That being said, this route which attempts to use the names composer and falls back to anonymous seems promising.

internal/controller/apiextensions/composite/composed.go
```diff
@@ -238,3 +278,324 @@ func IsReady(_ context.Context, cd resource.Composed, t v1.ComposedTemplate) (bo
...
+		return CompositionResult{}, errors.Wrap(err, errUpdate)
+	}
+
+	// The resources map should now contain resources that were rendered using
+	// the current array of templates, as well as any existing resources that no
+	// longer have a corresponding template. The former should be applied, while
+	// the latter should be deleted.
+	conn := managed.ConnectionDetails{}
+	ready := 0
+	for name, cd := range resources {
```
mcavoyk, line 568: Iterating over the map of resources will result in a different order each reconciliation. In practice, this might be low impact, but seems like a difference that was noteworthy.

This may allow an XR resource to work around a stuck resource after a few reconcile loops, as the composed resource introducing the error may fall in a different position when iterating on the map. But if applying all composed resources besides the ones which experience an error is desirable, it is probably worth doing so explicitly with continuation and an error array, rather than using the reconciler back-off and randomness.

internal/controller/apiextensions/composite/composed.go
```diff
@@ -238,3 +278,324 @@ func IsReady(_ context.Context, cd resource.Composed, t v1.ComposedTemplate) (bo
...
+
+	cr.SetResourceReferences(refs)
+	if err := r.client.Update(ctx, cr); err != nil {
+		return CompositionResult{}, errors.Wrap(err, errUpdate)
+	}
+
+	conn := managed.ConnectionDetails{}
+	ready := 0
+	for i, cd := range cds {
+		if err := r.client.Apply(ctx, cd, resource.MustBeControllableBy(cr.GetUID())); err != nil {
```
mcavoyk, line 437: This logic feels duplicated between the two composers, but unsure whether something like `cdStatus, err := r.composed.Reconcile(ctx, cd, cr, tpl)` reduces the complexity.

negz: I agree - this whole loop is quite similar between the two Composers. In much of it the only difference is that we lookup the template by index vs by map key. I suspect we can do better here.

internal/controller/apiextensions/composite/composed.go
negz: Good observation. That said, I'm not overly concerned about optimising for either applying managed resources in the order of their templates or making the most progress possible in a single reconcile. Arbitrary order of operation and retrying when we can't do something both seem fairly in line with the eventually consistent nature of the Kubernetes API to me.

apis/apiextensions/v1/composition_types.go
```diff
@@ -134,6 +134,17 @@ func TypeReferenceTo(gvk schema.GroupVersionKind) TypeReference {
...
+	// TODO(negz): TemplateName should be a required field in v2 of this API.
+
+	// A TemplateName uniquely identifies this resource template within its
+	// Composition. Template names are strongly recommended. Either all or no
+	// templates within a Composition must be named. Compositions that use
+	// anonymous templates match templates to composed resources by order, and
+	// must therefore treat the order and length of their spec.resources array
+	// as immutable.
+	// +optional
+	TemplateName *string `json:"templateName,omitempty"`
```
muvaf, line 146: Why not use `name` or `id`?

internal/controller/apiextensions/composite/reconciler.go
```diff
@@ -240,41 +218,52 @@ func WithConnectionPublisher(p ConnectionPublisher) ReconcilerOption {
...
+
 	r := &Reconciler{
-		client: resource.ClientApplicator{
-			Client:     kube,
-			Applicator: resource.NewAPIPatchingApplicator(kube),
-		},
+		client:       kube,
 		newComposite: nc,
 
+		composition: ValidationChain{
```
muvaf, line 242: The current implementation uses the array index as template name. What do you think about using one mutator instead of the validators that will update the `Composition` resource with templates having their index as `templateName` if it's `nil`? We'd not need the `AnonymousComposer` since all will be guaranteed to have a `templateName`.

internal/controller/apiextensions/composite/reconciler.go
```diff
@@ -240,41 +218,52 @@ func WithConnectionPublisher(p ConnectionPublisher) ReconcilerOption {
 type compositeResource struct {
 	CompositionSelector
 	Configurator
+	Composer
```
muvaf, line 221: I like that we separated `Composer`. We might be able to use the same reconciler in `CustomComposition` by changing the `Composer` implementation.

negz: That's good to know - I was worried it might be going back to too many layers of abstraction but it seemed like a clean-ish way to have two different implementations.

apis/apiextensions/v1/composition_types.go
negz: You convinced me that `name` was more idiomatic than `id`, but I was worried that `name` alone would easily be mistaken with `metadata.name` in this context so `templateName` seemed like a good way to make it clear that it's the name of the template, not the name of the resulting managed resource.

PR #2093, opened 2021-01-23
Implement multiple-source-field patch type

### LINKED ISSUE #1973: Composition transform patch: Introduce a way to getValues from an array of fieldPaths
Currently the `string` transform allows a single placeholder substitution in the `fmt` string. This allows pre-/in-/suffixing. Here's a use case where it would make more sense to consume a list of input `fromFieldPaths` for subsequent `%s` placeholders.

Consider a Composition that provisions an s3 bucket on AWS along with a set of IAM objects that reference it, resulting in a tight IRSA setup:

```yaml
kind: Composition
metadata:
  name: irsabuckets.xxx.com
  namespace: crossplane-system
spec:
  compositeTypeRef:
    apiVersion: xxx.com/v1alpha1
    kind: IRSABucket
  resources:
    - base:
        kind: Bucket
        spec:
          forProvider:
            acl: private
      patches:
      - fromFieldPath: "spec.parameters.region"
        toFieldPath: "spec.forProvider.locationConstraint"
    - base:
        apiVersion: identity.aws.crossplane.io/v1beta1
        kind: IAMRole
        spec:
          forProvider:
           assumeRolePolicyDocument: |
                 {
                   "Statement": [
                     {
                       "Action": "sts:AssumeRoleWithWebIdentity",
                       "Condition": {
                         "StringLike": {
                           "oidc.eks.AWS_REGION.amazonaws.com/id/XXXXX:sub": "system:serviceaccount:NAMESPACE:BUCKETNAME-access"
                         }
                       },
                     "Effect": "Allow",
                     "Principal": {
                       "Federated": "arn:aws:iam::AWS_ACCOUNT:oidc-provider/oidc.eks.AWS_REGION.amazonaws.com/id/XXXXX"
                       },
                     "Sid": ""
                     }
                   ],
                   "Version": "2012-10-17"
                 }
    - base:
        apiVersion: identity.aws.crossplane.io/v1alpha1
        kind: IAMPolicy
        patches:
        # (construct the policy document in the similar manner to th eIAMRole)
    - base:
        apiVersion: identity.aws.crossplane.io/v1beta1
        kind: IAMRolePolicyAttachment
        # (...)
```

In order to fill in all the necessary fields in `Policy` and `Role` documents, one would need to either use external templating (with Helm for example) or wrestle with some weird stack of partial patches. 

By having an array equivalent of  `fromFieldPath`

benagricola:
This PR adds a new patch type, `FromManyCompositeFieldPaths`, which accepts a list of composite field paths.

When the patch is applied, these multiple values are retrieved from the composite. If any value cannot be retrieved, the patch is aborted.

The `Transform` system has been modified to allow Transforms to run on multiple values. Standard transforms will output the same number of values as they are passed.

The `String` transform will now take any number of input values and output a single value. This is backwards compatible with 1.0.

Existing tests pass given the Transform modifications to process multiple values. Additional tests have been added for testing the multi input patch type and the string transform multiple input to single output.

### TODO
- [x] Add tests for multi-patch value retrieval
- [x] Add tests for multi-transform functionality
- [ ] Document added functionality (can work on this in combination with #2128) as 'advanced composition'

Fixes #1973

I have:

Existing tests pass. Code has been compiled and run in a `kind` cluster locally, using the following resource definitions to prove out the multi-field / multi-transform / combine pipeline:

<details>
<summary>claim.yaml</summary>
<p>

```yaml
kind: Test
metadata:
  name: test-one
  annotations:
    woo: lalala
  labels:
    label: test
spec:
  parameters:
    fieldOne: one
    fieldTwo: two
```
</p>
</details>

<details>
<summary>test-xrd.yaml</summary>
<p>

```yaml
# Create XRD with some default source values
metadata:
  name: compositetests.example.org
spec:
  group: example.org
  names:
    kind: CompositeTest
    plural: compositetests
  claimNames:
    kind: Test
    plural: tests
  defaultCompositionRef:
    name: test
  versions:
    - name: v1alpha1
      referenceable: true
      served: true
      schema:
        openAPIV3Schema:
          type: object
          properties:
            spec:
              type: object
              properties:
                parameters:
                  type: object
                  properties:
                    fieldOne:
                      type: string
                    fieldTwo:
                      type: string
              required:
                - parameters
```
</p>
</details>

<details>
<summary>test-composition.yaml</summary>
<p>

```yaml
kind: Composition
metadata:
  name: test 
spec:
  compositeTypeRef:
    kind: CompositeTest
  patchSets:
    - name: Metadata
      patches:
      - fromFieldPath: metadata.labels["crossplane.io/claim-namespace"]
        toFieldPath: metadata.namespace

    - name: Parameters
      patches:
      - type: FromMultipleCompositeFieldPaths
        # Retrieve two values from XRD parameters
        fromMultipleFieldPaths: 
          - spec.parameters.fieldOne
          - spec.parameters.fieldTwo
        toFieldPath: metadata.annotations["description"]
        transforms:
          # Map each field value to another value
          # ["one","two"] -> ["1","2"]
          - type: map
            map:
              one: "1"
              two: "2"

          # Format all values into a single output
          # ["1","2"] -> "Field One: 31 - Field Two: 32"
          - type: combine
            combine:
              type: string
              string:
                fmt: "Field One: %x - Field Two: %x"

  resources:
    - base:
        apiVersion: v1
        kind: Secret # Lazy testing, should not use native resources
        metadata:
          namespace:
        type: kubernetes.io/opaque
        data: {}
      patches:
        - type: PatchSet
          patchSetName: Metadata
        - type: PatchSet
          patchSetName: Parameters
```
</p>
</details>

Which produces the following `Secret` output on reconciliation:

<details>
<summary>secret.yaml</summary>
<p>

```yaml
apiVersion: v1
  kind: Secret
  type: kubernetes.io/opaque
  metadata:
    annotations:
      description: 'Field One: 31 - Field Two: 32'
    generateName: test-one-v55gp-
    labels:
      crossplane.io/claim-name: test-one
      crossplane.io/composite: test-one-v55gp
      label: test
    name: test-one-v55gp-lt7j7
    namespace: default
...
```
</p>
</details>


mcavoyk:
I don't dislike `FromMultipleCompositeFieldPaths`, but the type name seems to be getting quite lengthy, not sure what a solution here is, `FromMany` is shorter but not necessarily clearer.

I had a hard time parsing the example composite and understanding what the transform list did, took a few attempts for me to understand the inputs/outputs. I prefer keeping transform either very simple, or allowing them to be complex and nearly a full language, middle ground here seems like it has the problems from either side but without any of the benefits.

A simple transform solution would be not to have a separate `combine` transform sub-type, and just allow transform functions to take `$transform(input ....interface{}) (interface{}, error)`. This would mean the first transform could take the list, but chained transforms always were one input and one output. Anything more complex than this makes me want named inputs/outputs and the ability to fully functionalize the transform ability.

I lean towards keeping the transform's simple, and then continuing effort on `CustomCompositions` as the effort to allow a language to generate the set of resources rather than trying to create a language with the `transform` functionality. I have been struggling a bit with IRSA on the aws provider,  I will post a snippet of the design I was exploring, but the complexity reached makes me want to use a `CustomComposition` instead.

<details>
    <summary>simple-transform.yaml</summary>
  <br></br>
 
    patches:
      - type: FromManyCompositeFieldPaths
        fromManyCompositeFieldPaths:
        - "spec.parameters.someField"
        - "spec.parameters.otherField"
        toFieldPath: "spec.forProvider.exampleField"
        transforms:
          - type: string
            string:
              fmt: "%s-%s"

</head>
</details>

<details>
    <summary>complex-eks-irsa.yaml</summary>
  <br></br>
<head class="lang-yaml">

    patches:
      - type: FromManyCompositeFieldPaths
        fromManyCompositeFieldPaths:
          - name: oidcIssuer
            path: "status.atProvider.identity.oidc.issuer"
          - name: roleArn
            value: "spec.forProvider.roleArn"
        toFieldPath: "spec.forProvider.assumeRolePolicyDocument"
        transforms:
          - type: function
            function:
              name: trimPrefix
              parameters:
                - type: Variable
                  name: oidcIssuer
                - type: Constant
                  value: "https://"
              output:
                type: Variable
                name: oidcIssuerTrimmed

          - type: function
            function:
              name: regexFind
              parameters:
                - type: Variable
                  name: roleArn
                - type: Constant
                  value: '.*iam::([0-9]+).*'
              output:
                type: Variable
                name: accountId

          - type: function
            function:
              name: "printf"
              parameters:
                - type: Constant
                  value: |
                    {
                      "Version": "2012-10-17",
                      "Statement": [
                        {
                          "Sid": "",
                          "Effect": "Allow",
                          "Principal": {
                            "Federated": "arn:aws:iam::%s:oidc-provider/%s"
                          },
                          "Action": "sts:AssumeRoleWithWebIdentity",
                          "Condition": {
                            "StringEquals": {
                              "%s:sub": "system:serviceaccount:kube-system:cluster-autoscaler"
                            }
                          }
                        }
                      ]
                    }
                - type: Variable
                  name: accountId
                - type: Variable
                  name: oidcIssuerTrimmed
                - type: Variable
                  name: oidcIssuerTrimmed
</head>
</details>

benagricola:
> A simple transform solution would be not to have a separate `combine` transform sub-type, and just allow transform functions to take `$transform(input ....interface{}) (interface{}, error)`. This would mean the first transform could take the list, but chained transforms always were one input and one output. Anything more complex than this makes me want named inputs/outputs and the ability to fully functionalize the transform ability.
> 
> I lean towards keeping the transform's simple, and then continuing effort on `CustomCompositions` as the effort to allow a language to generate the set of resources rather than trying to create a language with the `transform` functionality.

I agree with this, I'd like to keep transforms as simple as possible and I think `CustomCompositions` is the right way to allow 'programmable' composition. 

My reasoning for this being a separate transform here is that it avoids any unintended behaviour along the lines of 

> "what happens when we give multiple input values to a `math`, `convert` or `map` transform?"

and tries to explicitly say -

> "this is the transform that turns your multiple values into a single output value, and these are the combination methods we support".

If there's a way we can express this without needing the extra transform type and without ending up with inconsistent behaviour between the transform types (maybe the answer is to just not allow multiple inputs for transforms that can't reduce them to a single output?), then I'm happy to rework this to keep things simple 👍

benagricola:
I've had a bit of an experiment with removing the combine transform type and relying solely on the existing `string` transform, as its' behaviour is essentially identical, without the added complexity of the additional types. 

This has the caveat that string transform behaviour should be well documented so it is clear that multiple inputs will be formatted into a single output value.

I think it is _probably_ safe to assume that if a user applies a transform with multiple input values, they are expecting the same number of output values _unless_ there is clear documentation to state otherwise.


The idea here is:
 * All transforms accept multiple values.
 * All transforms behave identically to 1.0 when passed a single value.
 * The `string` transform will take any number of input values and always output a single value. This is new behaviour that allows formatting of multiple values into a complex field.
 * All other existing transforms will output the same number of values as they were given.
 * There is no restriction to the number of values that can be outputted.
 * Future transforms are able to define their own behaviour when passed one or many values.

There are a couple of benefits that I can see to this:
 * `map`, `convert` and `math` transforms can be used to process many values, individually, before they're patched.
 * multiple values can be patched to an output field path as a slice. This means you can do something like this:

```yaml
# This patch would take the input parameters "one" and "two", and output []string{"three","four"}.
- type: FromManyCompositeFieldPaths
  fromManyFieldPaths:
    - spec.parameters.fieldOne # "one"
    - spec.parameters.fieldTwo # "two"
  toFieldPath: spec.parameters.multiField # This is a single field in another resource that takes []string
  transforms:
    # Map each field value to another value
    - type: map
      map:
        one: "three"
        two: "four"
```

You will receive an error from the API if trying to patch multiple values to a field type that does not support them.

_(NOTE: I haven't worked out how to write a test for patching multiple values to a single field, as the mock `Composed` resource does not contain an appropriate field to patch against - any suggestions here welcome)_


I can see this being enhanced to allow slice _inputs_ from a single field to be transformed and outputted (values would be dereferenced into the list of values to transform similarly to what `PatchSet` inlining currently does):

```yaml
# This patch would take the input slice []string{"one","two"} and output []string{"three","four"}.
  fromFieldPath: spec.parameters.fieldOne # []string{"one","two"}
  toFieldPath: spec.parameters.multiField # This is a single field in another resource that takes []string
  transforms:
    # Map each field value to another value
    - type: map
      map:
        one: "three"
        two: "four"
```

Interested to know if this is grokkable or if it's better to only allow multiple input values for the first transform, and block processing of multiple inputs for transforms that can't turn multiple inputs into a single output.

benagricola:
Rebased against master due to conflicts, and implemented the patching policy changes that were brought in with 1.1

muvaf:
An update for everyone who didn't make it to the last community meeting. We have landed on a bit different API that will allow this patch type to be self-contained without requiring a change in how transforms work. The following is an example:
```yaml
- type: CombineFromComposite
  combine:
    variables:
    - fromFieldPath: "metadata.name"
    - fromFieldPath: "spec.somefield"
    strategy: string # maybe "template" in future?
    string:
      fmt: "%s-somestuff-%s"
  toFieldPath: "spec.policy"
  transforms:
  - type: string
    string:
      fmt: "%s-value"
```

tohjustin:
> An update for everyone who didn't make it to the last community meeting. We have landed on a bit different API that will allow this patch type to be self-contained without requiring a change in how transforms work.

I have a particular use case that requires "conditional" patching (i.e. patch from field B only if field A exists) which seems to be achievable with this multiple-source field patch type + `Optional` patch policy. Here's what I'm trying to do:

1. I'm creating a composition with the [Queue](https://doc.crds.dev/github.com/crossplane/provider-aws/sqs.aws.crossplane.io/Queue/v1beta1@v0.18.1) managed resource from the aws-provider.
2. To implement namespace isolation (i.e. a `Queue` provisioned by an XRC cannot reference another `Queue` with different claim namespace as its dead-letter queue), I patch the XR's `spec.forProvider.redrivePolicy.deadLetterTargetArnSelector.matchLabels[crossplane.io/claim-namespace]` field to ensure that the selector can only match within the same claim namespace.
3. However this patch should only happen when the XRC creator wants their `Queue` managed resource to have a dead-letter queue (i.e.`spec.redrivePolicy.deadLetterTargetRef` field in the XRC isn't empty).
4. Here's my attempt to achieve this:

    ```yaml
    kind: Composition
    spec:
      resources:
      - patches:
        - fromFieldPath: spec.redrivePolicy.deadLetterTargetRef
          toFieldPath: spec.forProvider.redrivePolicy.deadLetterTargetArnSelector.matchLabels[crossplane.io/claim-name]
        # The patch below only happens when `spec.redrivePolicy.deadLetterTargetRef` exists
        - type: CombineFromComposite
          combine:
            variables:
            - fromFieldPath: spec.redrivePolicy.deadLetterTargetRef
            - fromFieldPath: metadata.labels[acme.co/namespace]
            strategy: string
            string:
              # `%.s` below is to omit the `spec.redrivePolicy.deadLetterTargetRef` value
              # since we only want the value of `metadata.labels[acme.co/namespace]`
              # for patching `deadLetterTargetArnSelector.matchLabels[crossplane.io/claim-namespace]`
              fmt: "%.s%s"
          toFieldPath: spec.forProvider.redrivePolicy.deadLetterTargetArnSelector.matchLabels[crossplane.io/claim-namespace]
          transforms:
          - type: string
            string:
              fmt: “%s”
      # ...
    ```

I guess there's another way to avoid using `%.s`, but I would still like to know whether the example above is considered a valid usage & isn't "discouraged" 😅

```yaml
kind: Composition
spec:
  resources:
  - patches:
    - type: CombineFromComposite
      combine:
        variables:
        - fromFieldPath: spec.redrivePolicy.deadLetterTargetRef
        - fromFieldPath: metadata.labels["acme.co/namespace"]
        strategy: string
        string:
          fmt: |-
            {
              "matchLabels": {
                "crossplane.io/claim-name": "%s",
                "crossplane.io/claim-namespace": "%s"
              }
            }
      toFieldPath: spec.forProvider.redrivePolicy.deadLetterTargetArnSelector
      transforms:
      - type: string
        string:
          fmt: “%s”
  # ...
```

negz:
@tohjustin I wouldn't consider what you're doing an anti-pattern, but it does seem like something that perhaps we could have a nicer API for. Would you mind copying your comment into a new issue that we could use to discuss further?

tohjustin:
> @tohjustin I wouldn't consider what you're doing an anti-pattern, but it does seem like something that perhaps we could have a nicer API for. Would you mind copying your comment into a new issue that we could use to discuss further?

Hi @negz, really appreciate you for following up on this! 🙏 
I've copied over my comment into a new issue https://github.com/crossplane/crossplane/issues/2364 as requested.

PR #2097, opened 2021-01-25
Establish and document release cadence and support cycle

### LINKED ISSUE #2066: Establish LTS schedule around releases

With the release of crossplane v1.0.0 all minor versions moving forward should be backwards compatible. However, it is likely that we will want to retire service for versions when a sufficient number of subsequent versions have been released.

We should publish an official schedule for versions that are currently supported and what the scope of a supported version is (likely that we will backport any fixes to the release branch).

hasheddan:

Adds documentation around Crossplane's release cycle and LTS. Leaves some room for process to be adjusted in the future, specifically around intra-release timeline.

Fixes #2066

I have:

n/a


docs/reference/release-cycle.md
```diff
@@ -0,0 +1,109 @@
...
+Alpha, Beta, and RC releases are cut for an upcoming release on an as-needed
+basis. As a policy, at least one pre-release will be cut prior to any minor
+release. Pre-releases will not be made on release branches.
+
+### Provider Releases
+
+The Crossplane release process is not required to be adhered to by any other
+Crossplane projects, but a similar cadence is encouraged. Maintainers listed in
+each repositories `OWNERS.md` file are responsible for determining and
+publishing the release process for their project.
```
negz, line 67: Do providers count as "other Crossplane projects" under this section? (I think yes, but it's a little ambiguous.)

docs/reference/release-cycle.md
```diff
@@ -0,0 +1,109 @@
...
+The following stages are the main milestones in a Crossplane release.
+
+### Active Development
+
+During active development, any code that meets the requisite criteria (i.e.
+passing appropriate tests, approved by a maintainer, etc.) will be merged into
+the main development branch. At present, there is no requirement to formally
+submit an enhancement proposal prior to the start of the release cycle, but
+contributors are encouraged to open an issues and gather feedback before
+starting work on a major implementation.
```
negz, line 80: Is it worth a link to CONTRIBUTING.md (and perhaps a backlink to this document?)

docs/reference/release-cycle.md
```diff
@@ -0,0 +1,109 @@
...
+
+### Feature Freeze
```
negz, line 82: Do you have any ideas around how we can keep ourselves accountable for not violating the feature and code freezes? Perhaps a label we can use if and when we make exceptions?

negz (approved):

docs/reference/release-cycle.md
hasheddan: Yep, will add note!

docs/reference/release-cycle.md
hasheddan: I think that is a good idea. I made sure to make this document indicate that our exception process is mostly informal right now because it doesn't make a ton of sense to implement a hard line before we really understand how the community will function with contributors / maintainers at large scale. We should eventually have some sort of automated merge blocker that requires an explicit invocation to allow an exception to go through in feature / code freeze.

docs/reference/release-cycle.md
hasheddan: @negz actually I think I may leave as-is since section title here is `Provider Releases`.

docs/reference/release-cycle.md
hasheddan: Another thing we should consider in the future it whether we want to go ahead and cut release branch at feature freeze and then backport any fixes. This would fall more under [release process](https://crossplane.io/docs/master/contributing/release-process.html) than release cycle though, and I am not sure it is worth the extra effort at this point with our current flow of contributions.

PR #2100, opened 2021-01-26
Fix v1beta1 packge import typo

khos2ow:

This PR fixes an API package import typo.

(see the discussion below)

~~This PR enables staticcheck job as part of github action CI, and it actually found an issue!~~

~~From their website:~~

~~> Staticcheck is a state of the art linter for the Go programming~~
~~> language. Using static analysis, it finds bugs and performance issues,~~
~~> offers simplifications, and enforces style rules.~~

I have:

not applicable.

### DIFF
--- a/apis/pkg/pkg.go
+++ b/apis/pkg/pkg.go
@@ -21,8 +21,8 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 
 	v1 "github.com/crossplane/crossplane/apis/pkg/v1"
-	v1beta1 "github.com/crossplane/crossplane/apis/pkg/v1"
 	v1alpha1 "github.com/crossplane/crossplane/apis/pkg/v1alpha1"
+	v1beta1 "github.com/crossplane/crossplane/apis/pkg/v1beta1"
 )
 
 func init() {


hasheddan (changes requested):
@khos2ow I believe we already run staticcheck using golangci-lint, where it is enabled by default (ref: https://golangci-lint.run/usage/linters/). However, it does looks like you have discovered a bug here in that `v1beta1` `pkg` types are not being added to scheme. In practice, I'm not sure how much impact this will have as I believe that `v1beta1` types will be stored as `v1` and they are using default conversion strategy of `None` (https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#overview), and we only request the `v1` flavor in our controllers (example: https://github.com/crossplane/crossplane/blob/d9d9836fae45bc505081136e65c612b67cbe0f27/internal/controller/pkg/revision/reconciler.go#L38). One thing that would be good to investigate is why this wasn't already caught by golangci-lint / staticcheck. I am guessing it is using an older version, so it may be worth trying to update. Could you investigate that?

khos2ow:
Yes I agree that this doesn't impact (maybe at all), and since I've noticed it I explicitly added it as part of ci and not depend on golangci-lint (which is strange on how it hadn't caught it).

I just tried it with latest version of golangci-lint and still doesn't catch it. I'll try to see what can be done.

khos2ow:
A bit of update, it seems the behavior of running staticcheck natively and through golangci-lint are different. I explicitly added a faulty code to make sure staticcheck is being run by golangci-lint at all. The result is as follow:

```
$ .cache/tools/linux_amd64/golangci-lint-v1.31.0 run ./...
cmd/crossplane/main.go:39:13: SA1004: sleeping for 1 nanoseconds is probably a bug; be explicit if it isn't (staticcheck)
        time.Sleep(1)
                   ^
$ .cache/tools/linux_amd64/golangci-lint-v1.36.0 run ./...
cmd/crossplane/main.go:39:13: SA1004: sleeping for 1 nanoseconds is probably a bug; be explicit if it isn't (staticcheck)
        time.Sleep(1)
                   ^
cmd/crank/install.go:154:14: type assertion on error will fail on wrapped errors. Use errors.As to check for specific errors (errorlint)
        serr, ok := err.(*apierrors.StatusError)
                    ^
$ go run honnef.co/go/tools/cmd/staticcheck -- ./...
apis/pkg/pkg.go:23:2: package "github.com/crossplane/crossplane/apis/pkg/v1" is being imported more than once (ST1019)
        apis/pkg/pkg.go:24:2: other import of "github.com/crossplane/crossplane/apis/pkg/v1"
cmd/crossplane/main.go:39:13: sleeping for 1 nanoseconds is probably a bug; be explicit if it isn't (SA1004)
exit status 1
```

In total there are three issues:

- golangci-lint@v1.31.0: only sees 1 (misses one staticcheck, the other one is due to being not latest version)
- golangci-lint@v1.36.0 (latest version): sees 2 (but still misses one from staticcheck)
- staticcheck: sees both issues (related to staticcheck)

To summarize, the state of codebase -running against staticcheck natively- is in pretty good shape (no major issues, even for couple providers which I tested manually), but still a bit of different behavior. I'm still a bit in favor of this PR to be on the very safe side of things but maybe that's being too much!

hasheddan:
@khos2ow do you know what version of `staticcheck` that `golangci-lint v1.36.0` is using? I am guessing that is what is causing the difference. Regardless, I think I would opt not to add a separate `staticcheck` job as golangci-lint is getting us most of the way there and will likely close any gaps in future releases.

khos2ow:
`golangci-lint v1.36.0` uses `honnef.co/go/tools @2020.1.6` which is two versions and ~150 commits behind master and latest release, but I went ahead and built HEAD of golangci-lint with HEAD of staticcheck and it was still the same, i.e. that particular error doesn't show up with golangci-lint.

But as I said earlier, that seems to be a very low impacting type of lint error and since the rest of the codebase (and other projects) behave very normally and don't have any major issues maybe this one specific case can be ignored altogether.

hasheddan:
@khos2ow yeah I think I would stick with golangci-lint only for now. Want to get this PR updated and we can get it merged?

khos2ow:
Wait what? I'm confused :sweat_smile:
I thought we agreed that this is too much and we can only rely on golangci-lint and ignore this one particular issue!

hasheddan:
@khos2ow right, but we still need to fix the bug :+1:

khos2ow:
Although it might worth to strip out the CI step and go ahead with `v1beta1` import change.

hasheddan (approved):

hasheddan:
This could be a cherry pick candidate, though we don't have our patch schedule well-defined at this point. If we did have an urgent patch I would like this to not be left off the branch, even though it is not really consequential from a functionality perspective. How do folks feel?

@kasey @muvaf @negz

PR #2102, opened 2021-01-27
Give Crossplane and installed providers access to Lease API for leader election

hasheddan:

Adds RBAC permissions for Crossplane / RBAC manager and provider to be able to perform leader election using the Kubernetes Lease API, which is used by default in controller-runtime v0.7.0+.

I have:

Ran locally and verified install + provider install:

```
🤖 (crossplane) k get pkg

NAME                                      INSTALLED   HEALTHY   PACKAGE                           AGE
provider.pkg.crossplane.io/provider-aws   True        True      crossplane/provider-aws:v0.16.0   33s
```

```
🤖 (crossplane) k get pods -n crossplane-system
NAME                                         READY   STATUS    RESTARTS   AGE
crossplane-5cb4c69d4f-zk9px                  1/1     Running   0          2m23s
crossplane-rbac-manager-dc5988cd8-kqqqp      1/1     Running   0          2m23s
provider-aws-6b36ca882d8e-6bf947b897-zlk5m   1/1     Running   0          68s
```

```
🤖 (crossplane) k logs -n crossplane-system crossplane-5cb4c69d4f-zk9px
I0127 22:42:11.833353       1 leaderelection.go:243] attempting to acquire leader lease crossplane-system/crossplane-leader-election-core...
I0127 22:42:11.943696       1 leaderelection.go:253] successfully acquired lease crossplane-system/crossplane-leader-election-core
```

### DIFF
--- a/cluster/charts/crossplane/templates/clusterrole.yaml
+++ b/cluster/charts/crossplane/templates/clusterrole.yaml
@@ -79,8 +79,10 @@ rules:
   - watch
 - apiGroups:
   - ""
+  - coordination.k8s.io
   resources:
   - configmaps
+  - leases
   verbs:
   - get
   - list
@@ -88,3 +90,4 @@ rules:
   - update
   - patch
   - watch
+  - delete
--- a/cluster/charts/crossplane/templates/rbac-manager-clusterrole.yaml
+++ b/cluster/charts/crossplane/templates/rbac-manager-clusterrole.yaml
@@ -79,8 +79,10 @@ rules:
   - "*"
 - apiGroups:
   - ""
+  - coordination.k8s.io
   resources:
   - configmaps
+  - leases
   verbs:
   - get
   - list
@@ -88,4 +90,5 @@ rules:
   - update
   - patch
   - watch
+  - delete
 {{- end}}
--- a/internal/controller/rbac/provider/roles/roles.go
+++ b/internal/controller/rbac/provider/roles/roles.go
@@ -45,6 +45,7 @@ const (
 	pluralEvents     = "events"
 	pluralConfigmaps = "configmaps"
 	pluralSecrets    = "secrets"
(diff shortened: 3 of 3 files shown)


khos2ow (approved):
Sorry my bad! I should have opened this up immediately after merging the crossplane-runtime upgrade PR.

Maybe worth mentioning this in some part of the doc? Specially for users who want to upgrade crossplane in place. I believe they should get rid of lease based on configmap prior to upgrade.

negz (approved):

hasheddan:
@khos2ow from my understanding, the default here should support both strategies for now https://github.com/kubernetes-sigs/controller-runtime/pull/1144#discussion_r480173688. Is that your read as well? That being said, we should certainly at least mention it in release notes and monitor for when it changes. I am also setting up some automated upgrade tests 👍

khos2ow:
@hasheddan sorry for the delay, yes that comment and the corresponding inline comments in relevant files are exactly the point here. In other words the `MultiLock` implementation will most probably will be good for majority of cases, in other words if you deploy crossplane (the upcoming release) for the very first time the _lock_ will be a `Lease` resource. But I personally haven't yet tested the migration path from current crossplane version to upcoming release.

That being said I don't expect to be causing any problem for the users, but definitely has to be pointed out in release note, or documentation if it's proven to be problematic (in some way or other).

hasheddan:
@khos2ow agreed. Users would hit issues with this if we switched to just using `Lease` in the future, and someone upgraded from `v1.0.0` (which uses `ConfigMap`) to the version that uses just `Lease`. So basically we need to make sure that folks upgrading from `v1.0.0` move to an intermediate release that uses the `MultiLock` before going to a release that only uses `Lease`. We could just support `Multilock` in perpetuity, in which case this wouldn't be an issue, but something to track for sure.

PR #2104, opened 2021-01-29
Support setting Pod Annotations in ControllerConfig

hasheddan:

Adds support for setting annotations on a provider Pod via the
referenced ControllerConfig. This is especially useful in cases where a
mutating webhook acts on Pods based on annotations present.

I have:

1. Install Crossplane
2. Create `Provider` that references `ControllerConfig` with `metadata.annotations` set
3. Observe annotations present on `Pod`


apis/pkg/v1alpha1/config.go
```diff
@@ -126,6 +130,17 @@ type ControllerConfigSpec struct {
...
 
+// PodObjectMeta is metadata that is added to the Pods in a provider's
+// Deployment.
+type PodObjectMeta struct {
+	// Annotations is an unstructured key value map stored with a resource that may be
+	// set by external tools to store and retrieve arbitrary metadata. They are not
+	// queryable and should be preserved when modifying objects.
+	// More info: http://kubernetes.io/docs/user-guide/annotations
+	// +optional
+	Annotations map[string]string `json:"annotations,omitempty"`
```
muvaf, line 141: Should we add `Labels`, too?

internal/controller/pkg/revision/deployment.go
```diff
@@ -96,6 +96,9 @@ func buildProviderDeployment(provider *pkgmetav1.Provider, revision v1.PackageRe
 		s.Annotations = cc.Annotations
 		d.Labels = cc.Labels
 		d.Annotations = cc.Annotations
```
muvaf, line 98: I think we should not propagate metadata anymore now that they can give it via spec. Unrelated to this PR, we had a problem regarding claim->composite direct metadata propagation where ArgoCD used labels to identify claim but since composite had the same labels, it did its operations on that, too. So, I'd avoid direct propagation of metadata if possible due to its usage patterns in the ecosystem.

apis/pkg/v1alpha1/config.go
hasheddan: I opted not to add `Labels` due to the semantics of `matchLabels` on the `Deployment` and `labels` on the pod spec template. We currently set a custom Crossplane label and I don't really want to enable adding additional ones or overwriting it until we have a use-case.

internal/controller/pkg/revision/deployment.go
hasheddan: Hmm this metadata is different than that we now include in the spec. These go to the `ServiceAccount` / `Deployment`, while the ones in the spec go to the `PodSpecTemplate`. We need to maintain the ability to add labels on the `ServiceAccount` because it is required for IRSA. I could be convinced to move these to the spec, but I currently like the model of keeping them separate because they act at kind of a "higher level". What do you think?

hasheddan:
@muvaf could you take another look here? :)

kasey (approved):
LGTM

apis/pkg/v1alpha1/config.go
muvaf: Sounds good.

internal/controller/pkg/revision/deployment.go
muvaf: Got it. It just feels like it's kind of surprising to get those propagated because when you add a metadata to an object, it belongs to that specific object and you use that metadata as [identifier](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/) in some context or just for [documentation](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/) purposes for other tools. So, having them propagated down to different resources feels more like actual configuration, rather than metadata. From what I've seen, the operators that deploy multiple resources usually include specific configurations under spec, like `spec.serviceAccount`/`spec.serviceAccountTemplate` or `spec.deployment` etc. We could do a similar thing here as well if there is use case.

But as I pointed out, this isn't really relevant to this specific PR.

muvaf (approved):
LGTM! I have one small concern about copying metada of `ControllerConfig` to everything it deploys but it's not directly related to this PR.

internal/controller/pkg/revision/deployment.go
hasheddan: @muvaf I tend to agree with you :+1: I'll see about addressing that in a subsequent PR

PR #2111, opened 2021-02-02
Add support for kubectl crossplane update

### LINKED ISSUE #1978: Add kubectl crossplane upgrade command for packages

Currently, packages can be installed using the Crossplane CLI, but if a user wants to upgrade they must manually edit the YAML.

Add commands for Providers and Configurations to upgrade. It should take the Provider / Configuration name and a tag or hash. It would also be nice to optionally allow for upgrade to the latest semver tag if no tag is provided.

weastel:
Signed-off-by: Drumil Patel <drumilpatel720@gmail.com>

Adds support for `kubectl crossplane update`
Fixes #1978 

I have:

For testing this PR I ran following steps:
1. `go build` in cmd/crank repository
2. Install any previous provider using `./crank install provider crossplane/provider-aws:v0.15.0 prov_name`
3. Checking provider has been installed via `kubectl get provider.pkg`
4. Updating provider using `./crank install provider prov_name v0.16.0`
5. Checking provider has been updated via `kubectl get provider.pkg`

### DIFF
--- a/cmd/crank/update.go
+++ b/cmd/crank/update.go
@@ -1,0 +1,106 @@
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"github.com/alecthomas/kong"
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/pkg/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+
+	v1 "github.com/crossplane/crossplane/apis/pkg/v1"
+	typedclient "github.com/crossplane/crossplane/internal/client/clientset/versioned/typed/pkg/v1"
+
+	// Load all the auth plugins for the cloud providers.
+	_ "k8s.io/client-go/plugin/pkg/client/auth"
+)
+
+// updateCmd updates a package.
+type updateCmd struct {
+	Configuration updateConfigCmd   `cmd:"" help:"Update a Configuration package."`
+	Provider      updateProviderCmd `cmd:"" help:"Update a Provider package."`
+}
+
+// Run runs the update cmd.
+func (c *updateCmd) Run(b *buildChild) error {
+	return nil
+}
+
+// updateConfigCmd updates a Configuration.
+type updateConfigCmd struct {
+	Name string `arg:"" help:"Name of Configuration."`
+	Tag  string `arg:"" help:"Updated tag for Configuration package."`
+}
+
+// Run runs the Configuration update cmd.
+func (c *updateConfigCmd) Run(k *kong.Context) error { // nolint:gocyclo
+	kube := typedclient.NewForConfigOrDie(ctrl.GetConfigOrDie())
+	prevConf, err := kube.Configurations().Get(context.Background(), c.Name, metav1.GetOptions{})
+	if err != nil {
+		return errors.Wrap(warnIfNotFound(err), "cannot update configuration")
+	}
+	pkg := prevConf.Spec.Package
+	pkgReference, err := name.ParseReference(pkg)
+	if err != nil {
+		return errors.Wrap(warnIfNotFound(err), "cannot update configuration")
+	}
+	newPkg := ""
+	if strings.HasPrefix(c.Tag, "sha256") {
+		newPkg = pkgReference.Context().Digest(c.Tag).Name()
+	} else {
+		newPkg = pkgReference.Context().Tag(c.Tag).Name()
+	}
+	prevConf.Spec.Package = newPkg
+	req, err := json.Marshal(prevConf)
+	if err != nil {
+		return errors.Wrap(warnIfNotFound(err), "cannot update configuration")
+	}
+	res, err := kube.Configurations().Patch(context.Background(), c.Name, types.MergePatchType, req, metav1.PatchOptions{})
+	if err != nil {
+		return errors.Wrap(warnIfNotFound(err), "cannot update configuration")
+	}
+	_, err = fmt.Fprintf(k.Stdout, "%s/%s updated\n", strings.ToLower(v1.ConfigurationGroupKind), res.GetName())
+	return err
+}
+
+// updateProviderCmd update a Provider.
+type updateProviderCmd struct {
+	Name string `arg:"" help:"Name of Provider."`
+	Tag  string `arg:"" help:"Updated tag for Provider package."`
+}
+
+// Run runs the Provider update cmd.
+func (c *updateProviderCmd) Run(k *kong.Context) error { // nolint:gocyclo
+	kube := typedclient.NewForConfigOrDie(ctrl.GetConfigOrDie())
+	preProv, err := kube.Providers().Get(context.Background(), c.Name, metav1.GetOptions{})
+	if err != nil {
+		return errors.Wrap(warnIfNotFound(err), "cannot update provider")
+	}
+	pkg := preProv.Spec.Package
+	pkgReference, err := name.ParseReference(pkg)
+	if err != nil {
+		return errors.Wrap(warnIfNotFound(err), "cannot update provider")
+	}
+	newPkg := ""
+	if strings.HasPrefix(c.Tag, "sha256") {
+		newPkg = pkgReference.Context().Digest(c.Tag).Name()
+	} else {
+		newPkg = pkgReference.Context().Tag(c.Tag).Name()
+	}
+	preProv.Spec.Package = newPkg
+	req, err := json.Marshal(preProv)
+	if err != nil {
+		return errors.Wrap(warnIfNotFound(err), "cannot update provider")
+	}
+	res, err := kube.Providers().Patch(context.Background(), c.Name, types.MergePatchType, req, metav1.PatchOptions{})
+	if err != nil {
+		return errors.Wrap(warnIfNotFound(err), "cannot update provider")
+	}
+	_, err = fmt.Fprintf(k.Stdout, "%s/%s updated\n", strings.ToLower(v1.ProviderGroupKind), res.GetName())
+	return err
+}
--- a/cmd/crank/main.go
+++ b/cmd/crank/main.go
@@ -48,6 +48,7 @@ var cli struct {
 
 	Build   buildCmd   `cmd:"" help:"Build Crossplane packages."`
 	Install installCmd `cmd:"" help:"Install Crossplane packages."`
+	Update  updateCmd  `cmd:"" help:"Update Crossplane packages."`
 	Push    pushCmd    `cmd:"" help:"Push Crossplane packages."`
 }
 


weastel:
Hi everyone, This pull request is WIP(Work In Progress). I have a doubt related to the implementation of `Tag-based Update`. Can we use this thread a way to help me resolve this conflict?

cmd/crank/update.go
```diff
@@ -0,0 +1,128 @@
...
+			Name: s,
+		}
+	}
+	cr := &v1.Configuration{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: pkgName,
+		},
+		Spec: v1.ConfigurationSpec{
+			PackageSpec: v1.PackageSpec{
+				Package:                  c.Package,
```
hasheddan, line 70: I think the `update` command should only modify the tag. So the command would look something like:

```
kubectl crossplane provider update <name-of-provider-k8s-object> <tag-to-update-spec.package-to>
```

An example would be:

```
kubectl crossplane provider update provider-aws v0.16.0
```

It will take some consideration to handle digest (`@`) vs tag (`:`) correctly, but `github.com/google/go-containerregistry/pkg/name` has some utilities for parsing.

hasheddan (changes requested):

cmd/crank/update.go
```diff
@@ -0,0 +1,146 @@
...
+			PackageSpec: v1.PackageSpec{
+				Package:                  newPkg,
+				RevisionActivationPolicy: &rap,
+				RevisionHistoryLimit:     &c.RevisionHistoryLimit,
+				PackagePullSecrets:       packagePullSecrets,
+			},
+		},
+	}
+	kube := typedclient.NewForConfigOrDie(ctrl.GetConfigOrDie())
+	res, err := kube.Configurations().Update(context.Background(), cr, metav1.UpdateOptions{})
```
hasheddan, line 87: Likely want to just `Patch` instead of `Update`

cmd/crank/update.go
hasheddan: I think we can just use `@` if "tag" starts with `sha256` and otherwise use `:`

cmd/crank/update.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+// Run runs the update cmd.
+func (c *updateCmd) Run(b *buildChild) error {
+	return nil
+}
+
+// updateConfigCmd updates a Configuration.
+type updateConfigCmd struct {
+	Package string `arg:"" help:"Image containing Configuration package."`
```
hasheddan, line 39: Package arg should be removed and Name should come first and not be optional

cmd/crank/update.go
```diff
@@ -0,0 +1,151 @@
...
+	cr := &v1.Configuration{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: pkgName,
+		},
+		Spec: v1.ConfigurationSpec{
+			PackageSpec: v1.PackageSpec{
+				Package:                  newPkg,
+				RevisionActivationPolicy: &rap,
+				RevisionHistoryLimit:     &c.RevisionHistoryLimit,
+				PackagePullSecrets:       packagePullSecrets,
```
hasheddan, line 84: This will likely overwrite these fields to their zero values if a user does not provide values for the flags. We should only update if the user provides args, which could be identified by using references and checking for nil

cmd/crank/update.go
```diff
@@ -0,0 +1,144 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name: pkgName,
+		},
+		Spec: v1.ConfigurationSpec{
+			PackageSpec: v1.PackageSpec{
+				Package: newPkg,
+			},
+		},
+	}
+	if &c.RevisionHistoryLimit != nil {
```
hasheddan, line 66: Use pointers as flag types instead of taking address of variable :+1:

hasheddan (changes requested):

cmd/crank/update.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+// updateConfigCmd updates a Configuration.
+type updateConfigCmd struct {
+	Name string `arg:"" help:"Name of Configuration."`
+	Tag  string `arg:"" help:"Updated tag for Configuration package."`
+
+	RevisionHistoryLimit *int64      `short:"rl" default:"-1" help:"Revision history limit."`
+	ManualActivation     *bool       `short:"m" help:"Enable manual revision activation policy."`
+	PackagePullSecrets   *([]string) `help:"List of secrets used to pull package."`
```
hasheddan, line 43:
```suggestion
	PackagePullSecrets   []string `help:"List of secrets used to pull package."`
```
Just check length on this :+1:

cmd/crank/update.go
```diff
@@ -0,0 +1,144 @@
...
+	pkgReference, _ := name.ParseReference(c.Name)
+	newPkg := ""
+	if strings.HasPrefix(c.Tag, "sha256") {
+		newPkg = pkgReference.Context().Digest(c.Tag).Name()
+	} else {
+		newPkg = pkgReference.Context().Tag(c.Tag).Name()
+	}
+	cr := &v1.Configuration{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: pkgName,
```
hasheddan, line 58:
```suggestion
			Name: c.name,
```

cmd/crank/update.go
```diff
@@ -0,0 +1,144 @@
...
+
+	RevisionHistoryLimit *int64      `short:"rl" default:"-1" help:"Revision history limit."`
+	ManualActivation     *bool       `short:"m" help:"Enable manual revision activation policy."`
+	PackagePullSecrets   *([]string) `help:"List of secrets used to pull package."`
+}
+
+// Run runs the Configuration update cmd.
+func (c *updateConfigCmd) Run(k *kong.Context) error {
+	pkgName := c.Name
+	pkgReference, _ := name.ParseReference(c.Name)
```
hasheddan, line 49: `c.Name` is the name of the configuration, not the package image. We will have to get the `Configuration` / `Provider` first, extract the current package image, then replace the tag or digest with the one specified.

cmd/crank/update.go
```diff
@@ -0,0 +1,144 @@
...
+		cr.Spec.PackageSpec.RevisionHistoryLimit = c.RevisionHistoryLimit
+	}
+	if c.ManualActivation != nil {
+		rap := v1.AutomaticActivation
+		if *c.ManualActivation {
+			rap = v1.ManualActivation
+		}
+		cr.Spec.PackageSpec.RevisionActivationPolicy = &rap
+	}
+	if c.PackagePullSecrets != nil {
```
hasheddan, line 76:
```suggestion
	if len(c.PackagePullSecrets) > 0 {
```

cmd/crank/update.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+// updateProviderCmd update a Provider.
+type updateProviderCmd struct {
+	Name string `arg:"" help:"Name of Provider."`
+	Tag  string `arg:"" help:"Updated tag for Provider package."`
+
+	RevisionHistoryLimit *int64 `short:"rl" help:"Revision history limit."`
+	ManualActivation     *bool  `short:"m" help:"Enable manual revision activation policy."`
+}
```
hasheddan, line 103: Looks like package pull secrets were dropped here.

cmd/crank/update.go
```diff
@@ -0,0 +1,144 @@
...
+
+// Run runs the Provider update cmd.
+func (c *updateProviderCmd) Run(k *kong.Context) error {
+	pkgName := c.Name
+	pkgReference, _ := name.ParseReference(c.Name)
```
hasheddan, line 108: Need to check error

cmd/crank/update.go
```diff
@@ -0,0 +1,144 @@
...
+	if c.ManualActivation != nil {
+		rap := v1.AutomaticActivation
+		if *c.ManualActivation {
+			rap = v1.ManualActivation
+		}
+		cr.Spec.PackageSpec.RevisionActivationPolicy = &rap
+	}
+	kube := typedclient.NewForConfigOrDie(ctrl.GetConfigOrDie())
+	req := new(bytes.Buffer)
+	_ = json.NewEncoder(req).Encode(cr)
```
hasheddan, line 137: `json.Marshal` will give you `bytes` back so it is likely simpler to use in this case

cmd/crank/update.go
weastel: `installProviderCmd` didn't consist of `PackagePullSecrets` , So I did same for it. If it is required than I will add this.

hasheddan: Ah good catch, would you mind adding it there?

cmd/crank/update.go
```diff
@@ -0,0 +1,171 @@
...
+	return nil
+}
+
+// updateConfigCmd updates a Configuration.
+type updateConfigCmd struct {
+	Name string `arg:"" help:"Name of Configuration."`
+	Tag  string `arg:"" help:"Updated tag for Configuration package."`
+
+	RevisionHistoryLimit *int64   `short:"rl" default:"-1" help:"Revision history limit."`
+	ManualActivation     *bool    `short:"m" help:"Enable manual revision activation policy."`
```
hasheddan, line 41: @weastel my apologies, it appears that pointers are not valid types in the `kong` parse tree, which is what is currently causing the unit test failures right now. Using a `default:"-1"` value as you have done here is not a bad idea at all. Another thing I could imagine trying is using a hook. However, for now would you be okay with just doing `Name` and `Tag` and we can revisit any of the flags in a follow-up if needed?

hasheddan (approved):
@weastel this looks pretty good to me! Thanks for your work :) Could you please squash commits and rebase + add testing steps in PR body?

cmd/crank/update.go
```diff
@@ -0,0 +1,124 @@
...
+			PackageSpec: v1.PackageSpec{
+				Package: newPkg,
+			},
+		},
+	}
+	req, err := json.Marshal(cr)
+	if err != nil {
+		return errors.Wrap(warnIfNotFound(err), "cannot update configuration")
+	}
+	res, err := kube.Configurations().Patch(context.Background(), cr.Name, types.StrategicMergePatchType, req, metav1.PatchOptions{})
```
hasheddan, line 72: @weastel I pulled down your fork here and tried this out and remembered that `StrategicMergePatch` is not supported for CRDs (ref https://github.com/kubernetes/kubernetes/issues/58414). We should be fine to use just `MergePatch` here.

cmd/crank/update.go
```diff
@@ -0,0 +1,124 @@
...
+	if err != nil {
+		return errors.Wrap(warnIfNotFound(err), "cannot update configuration")
+	}
+	newPkg := ""
+	if strings.HasPrefix(c.Tag, "sha256") {
+		newPkg = pkgReference.Context().Digest(c.Tag).Name()
+	} else {
+		newPkg = pkgReference.Context().Tag(c.Tag).Name()
+	}
+	cr := &v1.Configuration{
```
hasheddan, line 58: We don't need to construct an entire new object, we can just modify the `spec.package` of the one we already got

hasheddan (changes requested):

hasheddan (approved):
I validated this locally and works as expected, thanks for your work on this @weastel! I'll need to follow this up by bumping to a newer version of go-containerregistry which includes https://github.com/google/go-containerregistry/pull/824 because we are hitting the `index.docker.io` here again.

PR #2112, opened 2021-02-03
Add CodeQL security scanning to CI

### LINKED ISSUE #2098: Add CodeQL Security Scanning

CodeQL is a good security practice, and is a graduation requirement of the CNCF. It's easy to enable via GitHub Actions.

We should enable CodeSQL scanning. :) There is prior art in https://github.com/crossplane/crossplane/pull/1999 that just needs to be adapted to match our existing CI/CD conventions (see commentary on that PR).

khos2ow:

Enable CodeQL security scanning as part of the CI. This is essentially a copy/paste of #1999.

Fixes #2098

I have:

Let the CI run.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,42 @@
+name: CodeQL
+
+on:
+  push:
+    branches:
+      - master
+      - release-*
+  workflow_dispatch: {}
+
+jobs:
+  detect-noop:
+    runs-on: ubuntu-18.04
+    outputs:
+      noop: ${{ steps.noop.outputs.should_skip }}
+    steps:
+      - name: Detect No-op Changes
+        id: noop
+        uses: fkirc/skip-duplicate-actions@v2.1.0
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          paths_ignore: '["**.md", "**.png", "**.jpg"]'
+          do_not_skip: '["workflow_dispatch", "schedule", "push"]'
+          concurrent_skipping: false
+
+  analyze:
+    runs-on: ubuntu-18.04
+    needs: detect-noop
+    if: needs.detect-noop.outputs.noop != 'true'
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          submodules: true
+
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@v1
+        with:
+          languages: go
+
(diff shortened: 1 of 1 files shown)


khos2ow:
Originally it was suggested not to run this for PRs (https://github.com/crossplane/crossplane/pull/1999#discussion_r535545120), due to amount of time it takes to finish, as such I've set this to draft to see if I can manage to bring down the execution to some reasonable time or not.

On a brief observation I think we can incorporate CI cache and explicit building to reduce time; TBD.

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,45 @@
+name: CodeQL
+
+on:
+  push:
+    branches:
+      - master
+      - release-*
+  pull_request: {}
```
hasheddan, line 8: Happy for you to optimize here, but I think only running it on merges is a perfectly acceptable solution for now :)

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,45 @@
...
+  workflow_dispatch: {}
+  schedule:
+    - cron: "37 23 * * 6"
```
hasheddan, line 11: If we run on merges lets drop the periodic schedule here

.github/workflows/codeql.yml
khos2ow: I'm gonna try to see if it can be reduced to some reasonable time with reasonable effort, otherwise I agree, on merge should be good enough.

hasheddan: @khos2ow gotcha 👍 I would mention here though that running it at all on every PR feels a little wasteful to me, so I wouldn't invest too much time in optimization

hasheddan:
@khos2ow did you want to finish out the changes here? Or would you prefer to pass off to someone else? Either way is great!

khos2ow:
@hasheddan to give a little bit of update, I benchmarked running scanning a bit and overall it looks something like:

- `CI / unit-tests` (the longest running job): ~5-8 minutes
- `CodeQL / analyze`: ~8-10 minutes

I mean the difference is not very huge and expensive, but might be logically repetitive. I think it eventually comes down to preferences to keep running them on each PR or just leave them on `master` and `release-*` branches. To be honest I'm fine with both.

hasheddan (commented):
@khos2ow I would just run them on `master` and `release-*` and strip off the PR trigger :+1:

hasheddan (approved):
Thanks @khos2ow! LGTM :)

Issue #1974, opened 2020-11-18
Introduce composition patch types.


negz:
### What problem are you facing?

Crossplane Compositions consist of an array of 'base' resources, each with an array of zero or more patches. Currently these patches support patching from a field path within an XR to a field path within the base (composed) resource. Each patch may also include zero or more simple transforms. Currently `fromFieldPath` and `toFieldPath` are required fields.

For example:
```yaml
# ...
resources:
    - base:
        apiVersion: s3.aws.crossplane.io/v1beta1
        kind: Bucket
        spec:
          deletionPolicy: Delete
          forProvider:
            acl: private
            locationConstraint: unknown
            serverSideEncryptionConfiguration:
              rules:
                - applyServerSideEncryptionByDefault:
                    sseAlgorithm: aws:kms
        patches:
          - fromFieldPath: metadata.annotations
            toFieldPath: metadata.annotations
            transforms:
            - # ...
```

Since this API was introduced we've identified a handful of features that could fit as part of the `Patch` struct (the type that represents an entry in the `patches` array). These include:

* https://github.com/crossplane/crossplane/issues/1960 - Allowing references to sets of predefined patches.
* https://github.com/crossplane/crossplane/issues/1639 - Patching from a composed resource back to the XR.
* https://github.com/crossplane/crossplane/issues/1973 - Mapping an array of 'from' field paths mapped to a single 'to' path.

Supporting all of these functions would likely mean making all or most of the `Patch` struct fields optional, and relying on the user to set the appropriate fields for their use case (e.g. don't set both `fromFieldPath` and `fromFieldPaths`). It's good practice in Kubernetes API design to use a "differentiator" field (often called `type`) to allow the user to clearly specify their intent when they intend to set a subset of optional fields. For example:

```yaml
patches:
# The user wants to patch from one field of the XR to one field of
# the composed resource.
- type: FromCompositeField
  fromFieldPath: spec.parameters.foo
  toFieldPath: spec.forProvider.bar
# The user wants to include a patch set; this type lets Crossplane
# know that it's okay for them to omit the field paths, but that they
# must include an id.
- type: PatchSet
  id: CustomNS
```

I'd like us to consider the use cases we're aware of today before we introduce the patch `type` field, to increase the chances we pick a future proof naming scheme.

### How could Crossplane help solve your problem?

We could sketch out some example YAML that covers the above use cases in order to arrive at a user friendly and extensible set of patch types!

Here's a straw man example:

```yaml
[...]

PR #2120, opened 2021-02-05
fromFieldPath in secret

### LINKED ISSUE #1609: Composition to support taking values from members to fill the connection secret

In some cases, applications need non-sensitive data that is exposed through spec.forProvider or status.atProvider. But we composition only supports propagating values from the connection secret that the managed resource exposes, which has pre-defined keys that may not work for all cases.

See the discussion on Slack https://crossplane.slack.com/archives/CEG3T90A1/p1592583874432900

Similar to fromConnectionSecretKey, we can have fromFieldPath to refer to a value in the CR to be included in the final connection secret.

### LINKED ISSUE #2119: FieldPath reference for ConnectionDetails
Sometimes I would like to expose a field in the CR within the connectionDetails. Currently, there is no way to achieve this use we are using a constant value. 

Crossplane should include support in the composition to target a fieldPath for a secret.

krishchow:
Fixes #1609 

This PR adds a FromFieldPath value in the ConnectionDetail struct and also cleans up and adds functionality to parse the fieldPath into a value from CR.

I have:

Manual testing, unit testing needs to be added.

### DIFF
--- a/apis/apiextensions/v1/composition_types.go
+++ b/apis/apiextensions/v1/composition_types.go
@@ -543,6 +543,16 @@ func (s *ConvertTransform) Resolve(input interface{}) (interface{}, error) {
 	return f(input)
 }
 
+// A ConnectionDetailType is a type of connection detail.
+type ConnectionDetailType string
+
+// ConnectionDetailType types.
+const (
+	ConnectionDetailFromConnectionSecretKey ConnectionDetailType = "FromConnectionSecretKey" // Default
+	ConnectionDetailFromFieldPath           ConnectionDetailType = "FromFieldPath"
+	ConnectionDetailValue                   ConnectionDetailType = "Value"
+)
+
 // ConnectionDetail includes the information about the propagation of the connection
 // information from one secret to another.
 type ConnectionDetail struct {
@@ -552,15 +562,22 @@ type ConnectionDetail struct {
 	// +optional
 	Name *string `json:"name,omitempty"`
 
+	// Type sets the connection detail fetching behaviour to be used. Each connection detail type may require
+	// its' own fields to be set on the ConnectionDetail object.
+	// +optional
+	// +kubebuilder:validation:Enum=FromConnectionSecretKey;FromFieldPath;Value
+	// +kubebuilder:default=FromConnectionSecretKey
+	Type ConnectionDetailType `json:"type,omitempty"`
+
 	// FromConnectionSecretKey is the key that will be used to fetch the value
 	// from the given target resource's secret
 	// +optional
 	FromConnectionSecretKey *string `json:"fromConnectionSecretKey,omitempty"`
 
-	// FromResourceFieldPath is the path of the field on the composed resource whose value
-	// to be used as input. Name must be specified.
+	// FromFieldPath is the path of the field on the composed resource whose value
+	// to be used as input. Name must be specified if the type is FromFieldPath is specified.
 	// +optional
-	FromResourceFieldPath *string `json:"fromResourceFieldPath,omitempty"`
+	FromFieldPath *string `json:"fromFieldPath,omitempty"`
 
 	// Value that will be propagated to the connection secret of the composition
 	// instance. Typically you should use FromConnectionSecretKey instead, but
--- a/apis/apiextensions/v1/composition_types.go
+++ b/apis/apiextensions/v1/composition_types.go
@@ -208,7 +208,7 @@ type Patch struct {
 	// +kubebuilder:default=FromCompositeFieldPath
 	Type PatchType `json:"type,omitempty"`
 
-	// FromFieldPath is the path of the field on the upstream resource whose value
+	// FromFieldPath is the path of the field on the composed resource whose value
 	// to be used as input. Required when type is FromCompositeFieldPath.
 	// +optional
 	FromFieldPath *string `json:"fromFieldPath,omitempty"`
@@ -553,9 +553,14 @@ type ConnectionDetail struct {
 	Name *string `json:"name,omitempty"`
 
 	// FromConnectionSecretKey is the key that will be used to fetch the value
-	// from the given target resource.
+	// from the given target resource's secret
 	// +optional
 	FromConnectionSecretKey *string `json:"fromConnectionSecretKey,omitempty"`
+
+	// FromResourceFieldPath is the path of the field on the composed resource whose value
+	// to be used as input. Name must be specified.
+	// +optional
+	FromResourceFieldPath *string `json:"fromResourceFieldPath,omitempty"`
 
 	// Value that will be propagated to the connection secret of the composition
 	// instance. Typically you should use FromConnectionSecretKey instead, but
--- a/internal/controller/apiextensions/composite/composed.go
+++ b/internal/controller/apiextensions/composite/composed.go
@@ -23,6 +23,7 @@ import (
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/json"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -159,7 +160,7 @@ func NewAPIConnectionDetailsFetcher(c client.Client) *APIConnectionDetailsFetche
 }
 
 // FetchConnectionDetails of the supplied composed resource, if any.
-func (cdf *APIConnectionDetailsFetcher) FetchConnectionDetails(ctx context.Context, cd resource.Composed, t v1.ComposedTemplate) (managed.ConnectionDetails, error) {
+func (cdf *APIConnectionDetailsFetcher) FetchConnectionDetails(ctx context.Context, cd resource.Composed, t v1.ComposedTemplate) (managed.ConnectionDetails, error) { // nolint:gocyclo
 	sref := cd.GetWriteConnectionSecretToReference()
 	if sref == nil {
 		return nil, nil
@@ -178,28 +179,49 @@ func (cdf *APIConnectionDetailsFetcher) FetchConnectionDetails(ctx context.Conte
 	}
 
 	for _, d := range t.ConnectionDetails {
-		if d.Name != nil && d.Value != nil {
-			conn[*d.Name] = []byte(*d.Value)
-			continue
-		}
+		switch {
+		case d.Value != nil:
+			if d.Name != nil {
+				conn[*d.Name] = []byte(*d.Value)
+			}
+		case d.FromConnectionSecretKey != nil && len(s.Data[*d.FromConnectionSecretKey]) != 0:
+			key := *d.FromConnectionSecretKey
+			if d.Name != nil {
+				key = *d.Name
+			}
 
-		if d.FromConnectionSecretKey == nil {
-			continue
+			conn[key] = s.Data[*d.FromConnectionSecretKey]
+		case d.FromResourceFieldPath != nil && d.Name != nil:
+			_ = extractFieldPathValue(cd, d, conn)
 		}
-
-		if len(s.Data[*d.FromConnectionSecretKey]) == 0 {
-			continue
-		}
-
-		key := *d.FromConnectionSecretKey
-		if d.Name != nil {
-			key = *d.Name
-		}
-
-		conn[key] = s.Data[*d.FromConnectionSecretKey]
 	}
 
 	return conn, nil
+}
+
+func extractFieldPathValue(from runtime.Object, detail v1.ConnectionDetail, conn managed.ConnectionDetails) error {
+	fromMap, err := runtime.DefaultUnstructuredConverter.ToUnstructured(from)
+	if err != nil {
+		return err
+	}
+
+	str, err := fieldpath.Pave(fromMap).GetString(*detail.FromResourceFieldPath)
+	if err == nil {
+		conn[*detail.Name] = []byte(str)
+		return nil
+	}
+
+	in, err := fieldpath.Pave(fromMap).GetValue(*detail.FromResourceFieldPath)
+	if err != nil {
+		return err
+	}
+
+	buffer, err := json.Marshal(in)
+	if err != nil {
+		return err
+	}
+	conn[*detail.Name] = buffer
+	return nil
 }
 
 // IsReady returns whether the composed resource is ready.
--- a/internal/controller/apiextensions/composite/composed.go
+++ b/internal/controller/apiextensions/composite/composed.go
@@ -40,14 +40,17 @@ import (
 
 // Error strings
 const (
-	errApply       = "cannot apply composed resource"
-	errFetchSecret = "cannot fetch connection secret"
-	errReadiness   = "cannot check whether composed resource is ready"
-	errUnmarshal   = "cannot unmarshal base template"
-	errFmtPatch    = "cannot apply the patch at index %d"
-	errGetSecret   = "cannot get connection secret of composed resource"
-	errNamePrefix  = "name prefix is not found in labels"
-	errName        = "cannot use dry-run create to name composed resource"
+	errApply          = "cannot apply composed resource"
+	errFetchSecret    = "cannot fetch connection secret"
+	errReadiness      = "cannot check whether composed resource is ready"
+	errUnmarshal      = "cannot unmarshal base template"
+	errFmtPatch       = "cannot apply the patch at index %d"
+	errGetSecret      = "cannot get connection secret of composed resource"
+	errNamePrefix     = "name prefix is not found in labels"
+	errName           = "cannot use dry-run create to name composed resource"
+	errConnDetailKey  = "connection detail of type %s key is not set"
+	errConnDetailVal  = "connection detail of type %s value is not set"
+	errConnDetailPath = "connection detail of type %s fromFieldPath is not set"
 )
 
 // Observation is the result of composed reconciliation.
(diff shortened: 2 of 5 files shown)


internal/controller/apiextensions/composite/composed.go
```diff
@@ -178,28 +179,49 @@ func (cdf *APIConnectionDetailsFetcher) FetchConnectionDetails(ctx context.Conte
 	}
 
 	for _, d := range t.ConnectionDetails {
-		if d.Name != nil && d.Value != nil {
-			conn[*d.Name] = []byte(*d.Value)
-			continue
-		}
+		switch {
```
negz, line 182: Given we now have a few mutually exclusive secret sources (e.g. value, from connection secret key, and from field path) I think it's probably time to add a [discriminator field](https://github.com/kubernetes/kubernetes/issues/35345) here, similar to https://github.com/crossplane/crossplane/issues/1974. This would let us switch on that field rather than trying to infer intent from which values were and were not set.

I'd suggest something like:

```yaml
resources:
- base:
    # YAML..
  connectionDetails:
  - type: FromConnectionSecretKey  # The default, since this field needs to be optional
    fromConnectionSecretKey: somekey
  - type: FromFieldPath
    fromFieldPath: "status.interestingField"
```

I wonder if we should use something more specificly named than `FromFieldPath` to make it clear that the path is within the composed resource, not the composite. We've previously shied away from `FromComposedFieldPath` because it's easy for folks to confuse "composed" and "composite" at a glance.

internal/controller/apiextensions/composite/composed_test.go
```diff
@@ -263,6 +264,43 @@ func TestFetch(t *testing.T) {
...
+				t: v1.ComposedTemplate{ConnectionDetails: []v1.ConnectionDetail{
+					{
+						// Entries with only a name are silently ignored.
+						Name:          pointer.StringPtr("name"),
+						FromFieldPath: pointer.StringPtr("objectMeta.name"),
+					},
+					{
+						// Entries with only a value are silently ignored.
+						Value: pointer.StringPtr("missingname"),
+					},
```
negz, line 295: Is it worth including this entry? iirc it's tested in another case.

internal/controller/apiextensions/composite/composed.go
```diff
@@ -178,28 +179,49 @@ func (cdf *APIConnectionDetailsFetcher) FetchConnectionDetails(ctx context.Conte
...
+	in, err := fieldpath.Pave(fromMap).GetValue(*detail.FromFieldPath)
+	if err != nil {
+		return err
+	}
+
+	buffer, err := json.Marshal(in)
+	if err != nil {
+		return err
+	}
+	conn[*detail.Name] = buffer
```
negz, line 223: I would expect `json.Marshal` of a string value to return a byte array containing a quoted string - e.g. `[]byte("\"value\"")` - but the unit tests say otherwise. What am I missing? 🤔

negz: Ah - just realised this was a fallback for cases in which the value isn't a string. Should we add a test for this?

negz (commented):
Thanks for tackling this @krishchow! I have a few small comments but generally this looks great.

internal/controller/apiextensions/composite/composed.go
cheewaio: How about `FromResourceFieldPath` or `ValueFromResourceFieldPath`? I think it should also note that `name` must be specified when set.

internal/controller/apiextensions/composite/composed.go
```diff
@@ -159,7 +160,7 @@ func NewAPIConnectionDetailsFetcher(c client.Client) *APIConnectionDetailsFetche
 }
 
 // FetchConnectionDetails of the supplied composed resource, if any.
-func (cdf *APIConnectionDetailsFetcher) FetchConnectionDetails(ctx context.Context, cd resource.Composed, t v1.ComposedTemplate) (managed.ConnectionDetails, error) {
+func (cdf *APIConnectionDetailsFetcher) FetchConnectionDetails(ctx context.Context, cd resource.Composed, t v1.ComposedTemplate) (managed.ConnectionDetails, error) { // nolint:gocyclo
 	sref := cd.GetWriteConnectionSecretToReference()
 	if sref == nil {
```
cheewaio, line 165: When using `fromFieldPath`, does `writeConnectionSecretToRef` still required on the managed resource? Wouldn't it be reasonable to expect `writeConnectionSecretToRef` only needed when using with `fromConnectionSecretKey`

internal/controller/apiextensions/composite/composed.go
krishchow: Sounds good to me, I'll make the relevant changes shortly :)

internal/controller/apiextensions/composite/composed_test.go
```diff
@@ -271,6 +309,8 @@ func TestFetch(t *testing.T) {
 			if diff := cmp.Diff(tc.want.err, err, test.EquateErrors()); diff != "" {
 				t.Errorf("\n%s\nFetch(...): -want, +got:\n%s", tc.reason, diff)
 			}
+			fmt.Printf("%+v\n", tc.want.conn)
+			fmt.Printf("%+v\n", conn)
```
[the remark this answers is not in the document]
krishchow: Yup, these were for debugging, both have been removed

apis/apiextensions/v1/composition_types.go
```diff
@@ -553,10 +553,15 @@ type ConnectionDetail struct {
 	Name *string `json:"name,omitempty"`
 
 	// FromConnectionSecretKey is the key that will be used to fetch the value
-	// from the given target resource.
+	// from the given target resource's secret
 	// +optional
 	FromConnectionSecretKey *string `json:"fromConnectionSecretKey,omitempty"`
 
+	// FromFieldPath is the path of the field on the upstream resource whose value
```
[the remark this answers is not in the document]
krishchow: fixed

internal/controller/apiextensions/composite/composed.go
krishchow: I would think writeConnectionSecretToRef is still required since we still need to write the values to a specific secret.

cheewaio: In my experiment, crossplane will write the values to a working connection secret with uid name to the namespace specified in `writeConnectionSecretsToNamespace`. I don't need to specify `writeConnectionSecretToRef` when `fromConnectionSecretKey` is not used. Same goes to fixed `name` and `value` pair

krishchow: Ahh yes, I am not sure, let me test out this behavior

krishchow: I understand now, I think it would be okay to make the GetWriteConnectionSecretToReference optional, this way we can still publish values/FromResourceFieldPath even if there is no secret for a resource.

How does that sound to you @negz?

negz: That sounds good to me!

internal/controller/apiextensions/composite/composed.go
```diff
@@ -159,47 +160,83 @@ func NewAPIConnectionDetailsFetcher(c client.Client) *APIConnectionDetailsFetche
...
+		// It's possible that the composed resource does want to write a
+		// connection secret but has not yet. We presume this isn't an issue and
+		// that we'll propagate any connection details during a future
+		// iteration.
+		s := &corev1.Secret{}
+		nn := types.NamespacedName{Namespace: sref.Namespace, Name: sref.Name}
+		if err := cdf.client.Get(ctx, nn, s); client.IgnoreNotFound(err) != nil {
+			return nil, errors.Wrap(err, errGetSecret)
+		}
+		data = s.Data
```
negz, line 181: What do you think about moving this whole secret-getting block inside the `v1.ConnectionDetailFromConnectionSecretKey` case? It looks like that's the only place that uses `data`, so we could probably just return/continue early if the user asked for a particular connection secret key but there was no connection secret. e.g.:

```go
switch d.Type {
case v1.ConnectionDetailFromConnectionSecretKey:
	sref := cd.GetWriteConnectionSecretToReference()
	if sref == nil {
		continue
	}
	# ...
}
```

internal/controller/apiextensions/composite/composed.go
```diff
@@ -159,47 +160,83 @@ func NewAPIConnectionDetailsFetcher(c client.Client) *APIConnectionDetailsFetche
...
 	for _, d := range t.ConnectionDetails {
-		if d.Name != nil && d.Value != nil {
-			conn[*d.Name] = []byte(*d.Value)
-			continue
+		switch d.Type {
+		case v1.ConnectionDetailValue:
+			if d.Name != nil && d.Value != nil {
+				// Name must be set if value type
+				conn[*d.Name] = []byte(*d.Value)
+			}
```
negz, line 191: I wonder whether it would make sense for this case to return an error if name or value are missing. We now have a clear understanding of the author's intent - they're declaring that this is a `type: Value` connection secret, which requires a name and a value. It might be useful to return an error if one or both are missing rather than silently proceeding without the requested connection detail.

Note that this case is different from the following two because in those cases the connection secret key or composed resource field value are likely to be absent at first, before eventually being populated. In those cases we can't tell the difference between "the value hasn't appeared yet" and "the value will never appear", but in this one we know that the value will never show up without human intervention.

internal/controller/apiextensions/composite/composed.go
```diff
@@ -159,47 +160,83 @@ func NewAPIConnectionDetailsFetcher(c client.Client) *APIConnectionDetailsFetche
...
 		}
+	}
 
-		if d.FromConnectionSecretKey == nil {
-			continue
-		}
+	if len(conn) == 0 {
+		return nil, nil
+	}
```
negz, line 212: Does it matter whether we return `nil` vs an empty `managed.ConnectionDetails{}` here?

apis/apiextensions/v1/composition_types.go
```diff
@@ -543,6 +543,16 @@ func (s *ConvertTransform) Resolve(input interface{}) (interface{}, error) {
 	return f(input)
 }
 
+// A ConnectionDetailType is a type of connection detail.
+type ConnectionDetailType string
+
+// ConnectionDetailType types.
+const (
+	ConnectionDetailFromConnectionSecretKey ConnectionDetailType = "FromConnectionSecretKey" // Default
+	ConnectionDetailFromResourceFieldPath   ConnectionDetailType = "FromResourceFieldPath"
```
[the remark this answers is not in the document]
cheewaio: The thought process was to align with composition resource bases. 
```
resources:
 - base
 - base
```
Perhaps, `FromBaseFieldPath`? Just throwing out some suggestion for your consideration.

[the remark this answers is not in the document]
negz: Hrm, technically it's not really from the base though - it's decorated onto the base. I would suggest we use either `FromComposedFieldPath` or the original `FromFieldPath`.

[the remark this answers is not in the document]
krishchow: Switching it back to `FromFieldPath` since that term is used elsewhere in the API for the composition

apis/apiextensions/v1/composition_types.go
```diff
@@ -552,11 +562,23 @@ type ConnectionDetail struct {
 	// +optional
 	Name *string `json:"name,omitempty"`
 
+	// Type sets the connection detail fetching behaviour to be used. Each connection detail type may require
+	// its' own fields to be set on the ConnectionDetail object.
+	// +optional
+	// +kubebuilder:validation:Enum=FromConnectionSecretKey;FromResourceFieldPath;Value
+	// +kubebuilder:default=FromConnectionSecretKey
+	Type ConnectionDetailType `json:"type,omitempty"`
```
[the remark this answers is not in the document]
krishchow: I based this on the PatchType field which is not a pointer

```
// Type sets the patching behaviour to be used. Each patch type may require
// its' own fields to be set on the Patch object.
// +optional
// +kubebuilder:validation:Enum=FromCompositeFieldPath;PatchSet;ToCompositeFieldPath
// +kubebuilder:default=FromCompositeFieldPath
Type PatchType `json:"type,omitempty"`
```

internal/controller/apiextensions/composite/composed.go
krishchow: If I am understanding this correctly, we would need to get the secret each time we loop for every connectionDetail of type `v1.ConnectionDetailFromConnectionSecretKey`, which seems less efficient

internal/controller/apiextensions/composite/composed.go
krishchow: Hmm that makes a lot of sense, I'll add in the corresponding error + test now

internal/controller/apiextensions/composite/composed.go
krishchow: Yup, if we return an empty managed.ConnectionDetails{}, it creates an empty secret.

krishchow: The current behavior is that no secret is created, so I added this check to ensure that aspect of the connectionDetails does not change

internal/controller/apiextensions/composite/composed.go
krishchow: Similarly, should I return an error if the name is not set for `v1.ConnectionDetailFromFieldPath`, since this also expects the name to be set?

krishchow: Done, let me know what you think :)

apis/apiextensions/v1/composition_types.go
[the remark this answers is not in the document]
negz: Got it. I feel good about these being non-pointers since in practice the API server will make sure they always have a value set thanks to the `default`.

internal/controller/apiextensions/composite/composed.go
negz: Ah - good point. I missed the loop.

Another option here would be:

```go
data := map[string][]byte{}
if sref := cd.GetWriteConnectionSecretToReference(); sref != nil {
	s := &corev1.Secret{}
	nn := types.NamespacedName{Namespace: sref.Namespace, Name: sref.Name}
	if err := cdf.client.Get(ctx, nn, s); client.IgnoreNotFound(err) != nil {
		return nil, errors.Wrap(err, errGetSecret)
	}
	data = s.Data
}
```

That said, I'm just nitpicking because I have a weird aversion to `else` statements, so feel free to ignore this suggestion.

internal/controller/apiextensions/composite/composed.go
```diff
@@ -159,47 +163,96 @@ func NewAPIConnectionDetailsFetcher(c client.Client) *APIConnectionDetailsFetche
...
-		if d.Name != nil && d.Value != nil {
-			conn[*d.Name] = []byte(*d.Value)
-			continue
-		}
+		switch d.Type {
+		case v1.ConnectionDetailValue:
+			// Name, Value must be set if value type
+			switch {
+			case d.Name != nil && d.Value != nil:
+				conn[*d.Name] = []byte(*d.Value)
```
negz, line 194: Nit: I think this could be the `default` case, or just live below the `switch`. Switch cases are evaluated in order, so we'd know the name and value weren't nil if we got past the below two cases.

internal/controller/apiextensions/composite/composed.go
```diff
@@ -159,47 +163,96 @@ func NewAPIConnectionDetailsFetcher(c client.Client) *APIConnectionDetailsFetche
...
+				key := *d.FromConnectionSecretKey
+				if d.Name != nil {
+					key = *d.Name
+				}
+				if key != "" {
+					conn[key] = data[*d.FromConnectionSecretKey]
+				}
+			} else if d.FromConnectionSecretKey == nil {
+				return nil, fmt.Errorf(errConnDetailKey, d.Type)
+			}
```
negz, line 212:
```suggestion
			if d.FromConnectionSecretKey == nil {
				return nil, fmt.Errorf(errConnDetailKey, d.Type)
			}
			if  data[*d.FromConnectionSecretKey] == nil {
				// We don't consider this an error because it's possible the
				// key will still be written at some point in the future.
				continue
			}
			key := *d.FromConnectionSecretKey
			if d.Name != nil {
				key = *d.Name
			}
			if key != "" {
				conn[key] = data[*d.FromConnectionSecretKey]
			}
```

I'd prefer to make the `d.FromConnectionSecretKey == nil` condition a guard clause/early return rather than having it trail the main logic here. We may also consider doing the same with the `data[*d.FromConnectionSecretKey] == nil` condition per my suggestion above.

internal/controller/apiextensions/composite/composed.go
```diff
@@ -159,47 +163,96 @@ func NewAPIConnectionDetailsFetcher(c client.Client) *APIConnectionDetailsFetche
...
+		case v1.ConnectionDetailFromFieldPath:
+			switch {
+			case d.Name != nil && d.FromFieldPath != nil:
+				_ = extractFieldPathValue(cd, d, conn)
```
negz, line 216: Similar to my suggestion above, we can infer that Name and FromFieldPath are not nil by just moving this below the following two cases.

negz (approved):
Thanks @krishchow! This looks good overall, but I've left a few style/nitpick suggestions. Take a look and see if any make sense to accept, then let me know when you're ready for me to merge. :)

internal/controller/apiextensions/composite/composed.go
krishchow: I like this way more! Looks so much cleaner haha

internal/controller/apiextensions/composite/composed.go
krishchow: Done!

internal/controller/apiextensions/composite/composed.go
krishchow: Much cleaner :)

PR #2124, opened 2021-02-09
Allow ability to expose metric endpoints for providers

smcavallo:
We know the base crossplane exposes prometheus metrics when metrics.enabled=true with helm.
What is missing is the ability to exposing metrics for crossplane-aws (or other providers.)
Monitoring is an important component for the package providers and there is currently no way to enable it for providers.  Even when they are natively instrumented with prometheus listening on `:8080` and `/metrics` this port is not exposed outside of the container itself.
This change allows the configuration of container ports to expose metrics outside the pod and enable end users to add monitors around package providers.

Fixes #

I have:

```
kind: ControllerConfig
metadata:
  name: aws-config
spec:
  podSecurityContext:
    fsGroup: 2000
  ports:  
  - containerPort: 8080
    name: metrics
```
results in deployment and resulting pod ->
```
spec:
  containers:
  - image: crossplane/provider-aws-controller:v0.16.0
    imagePullPolicy: IfNotPresent
    name: provider-aws
    ports:
    - containerPort: 8080
      name: metrics
      protocol: TCP
    resources: {}
    securityContext:
      privileged: false
      runAsGroup: 2000
      runAsNonRoot: true
      runAsUser: 2000
    terminationMessagePath: /dev/termination-log
    terminationMessagePolicy: File
```


internal/controller/pkg/revision/deployment.go
```diff
@@ -105,6 +105,9 @@ func buildProviderDeployment(provider *pkgmetav1.Provider, revision v1.PackageRe
 		if cc.Spec.Image != nil {
 			d.Spec.Template.Spec.Containers[0].Image = *cc.Spec.Image
 		}
+		if cc.Spec.Ports != nil {
```
hasheddan, line 108: It won't make a difference in this case, but would you mind using `len(cc.spec.Ports) > 0` for consistency with other slice types here?

hasheddan (approved):
Thanks @smcavallo! Two minor nitpicks, but otherwise LGTM!

PR #2130, opened 2021-02-11
Return consistent connection secret errors

negz:

Some trivial tweaks to the errors we return when handling connection secrets. I didn't notice these until after I'd merged https://github.com/crossplane/crossplane/pull/2120 and was rebasing my branch on the changes. 

* Use the errors package rather than fmt.Errorf (adds stack trace)
* Use %q (quoted values) in error format strings
* Distinguish errFmt strings from plain old err strings

I have:

Just the unit tests - it's a trivial change.

### DIFF
--- a/internal/controller/apiextensions/composite/composed.go
+++ b/internal/controller/apiextensions/composite/composed.go
@@ -40,17 +40,18 @@ import (
 
 // Error strings
 const (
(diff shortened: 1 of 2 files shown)


hasheddan (approved):
LGTM! Thanks @negz!

PR #2132, opened 2021-02-11
Fix liquid template rendering in vault guide

hasheddan:

Fixes an issue where the vault injection template was being rendered
incorrectly due to jekyll interpreting it as a liquid template.

See https://crossplane.io/docs/master/guides/vault-injection.html#install-provider-gcp

I have:

Rendered locally:

raw-liquid

### DIFF
--- a/docs/guides/vault-injection.md
+++ b/docs/guides/vault-injection.md
@@ -227,6 +227,7 @@ injected into the container filesystem by assuming role `crossplane-providers`.
 There is also so template formatting added to make sure the secret data is
 presented in a form that `provider-gcp` is expecting.
 
+{% raw  %}
 ```console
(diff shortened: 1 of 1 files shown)


negz (approved):

negz:
Unrelated to this PR, but it didn't occur to me until now - would it be worth publishing snippets for this guide like we do for the majority of the docs? I think the copy-and-paste `echo` style is unique to this guide within our documentation.

hasheddan:
@negz we definitely could, there actually is some areas where we use the `echo` style elsewhere because it allows for env var expansion (ref https://crossplane.io/docs/master/guides/vault-injection.html#install-provider-gcp)

Issue #2094, opened 2021-01-23
Optionally mark certain patches as required


negz:
### What problem are you facing?

https://github.com/crossplane/crossplane/blob/65e42d5/apis/apiextensions/v1/composition_types.go#L266

Currently Crossplane will happily allow you to patch from a non-existent field path - treating the patch as a no-op. This enables 'opportunistic' patching - i.e. "patch from this field if it exists". This us useful for some use cases, but in others it would be better to return an error (and retry again later) when a patch's source is not available. For example in https://github.com/crossplane/crossplane/issues/2091 the goal is to patch from the XR's `spec.claimRef.name`, which may not exist for a brief period when the XR is first created. Ideally in this case we'd return an error and try again, eventually succeeding when the claim was specified.

### How could Crossplane help solve your problem?

Crossplane could allow Composition authors to specify whether a particular patch is required or not. i.e.:

```yaml
      patches:
        - fromFieldPath: "spec.claimRef.name"
          toFieldPath: "metadata.name"
          transforms:
            - type: string
              string:
                fmt: "example-%s"
          # Better names welcome. I'm going for something like
          # "required: true" but using an enum to keep things open
          # ended per API conventions.
          strategy: Required
```

The default behaviour would match today's (i.e. the patch would be a no-op), but if `strategy: Required` we'd instead return an error and retry.

muvaf:
I'm wondering whether it should be `Required` all the time. Are there cases where you'd need it to create the resource even if the source field path doesn't have the information? For example, using https://github.com/crossplane/crossplane/pull/2077 I would use this feature as a lock mechanism, i.e. create the composed resource only if `status.subnet` has a value, similar to cross resource references where the resource isn't reconciled if there is no value in the referenced object.

negz:
>  I'm wondering whether it should be Required all the time.

I don't think that's an option now that we've hit 1.0. It's a behaviour change that I think could have impact on existing Compositions - e.g. a Composition that doesn't block on missing fields in Crossplane 1.0 shouldn't start blocking in 1.1.

> Are there cases where you'd need it to create the resource even if the source field path doesn't have the information? 

IMO there are. I touched on this in https://github.com/crossplane/crossplane/issues/1455, which was the issue that lead to the condition. The most concrete example I can think of would be things like labels, annotations, and tags where you might want to propagate something opportunistically. i.e. "Copy the foo annotation from the XR only _if_ it exists".

[...]

Issue #2091, opened 2021-01-22
Claim created two instances of the same resource when metadata.name is set using a transformation


bandesz:
### What happened?

I have a claim which would create a single RDSInstance resource.

In the composition the metadata.name was set using a transformation - the idea being that the Kubernetes object would have the same name as the real resource.

Normally only one underlying resource is created, but one time I had two created with a one second difference.

The claim stays not ready forever.

See the resources below, only one of the RDSInstance objects has been reconciled (and the database is indeed created), and both belong to the same composite object.

If I delete the claim, both objects are deleted.

### How can we reproduce it?

Composition:
```yaml
---
apiVersion: apiextensions.crossplane.io/v1
kind: Composition
metadata:
  name: aws-rds-postgresql-database-default
  labels:
    crossplane-configuration: crossplane-aws-rds
    provider: aws
spec:
  writeConnectionSecretsToNamespace: example-crossplane
  compositeTypeRef:
    apiVersion: aws.crossplane.example.io/v1alpha1
    kind: PostgreSQLDatabase
  resources:
    - base:
        apiVersion: database.aws.crossplane.io/v1beta1
        kind: RDSInstance
        spec:
          forProvider:
            [...]
          writeConnectionSecretToRef:
            namespace: example-crossplane
      patches:
        - fromFieldPath: "spec.claimRef.name"
          toFieldPath: "metadata.name"
          transforms:
            - type: string
              string:
                fmt: "example-%s"
```

The resources created:
```yaml
- apiVersion: database.aws.crossplane.io/v1beta1
  kind: RDSInstance
  metadata:
    annotations:
      crossplane.io/external-name: example-e2e-poqcfy2l
    creationTimestamp: "2021-01-22T11:18:43Z"
    generateName: e2e-poqcfy2l-4vdt4-
    generation: 80
    labels:
      crossplane.io/claim-name: e2e-poqcfy2l
      crossplane.io/claim-namespace: crossplane-test
      crossplane.io/composite: e2e-poqcfy2l-4vdt4
    name: example-e2e-poqcfy2l
    ownerReferences:
    - apiVersion: aws.crossplane.example.io/v1alpha1
      controller: true
      kind: MySQLDatabase
      name: e2e-poqcfy2l-4vdt4
      uid: 56b61692-3e13-41b6-b399-13379275d0b6
    resourceVersion: "9486478"
[...]

Issue #2099, opened 2021-01-26
Support patching from common data sources?


negz:
### References

Design: #3008
Implementation: #3007 

### What problem are you facing?

When platform teams design their XRs they may want to use environment scoped data to render out composed resources. 
It's possible that this data varies enough that it's not ideal for it to be a fixed value in a Composition, but that it's also not something that the platform team desired claim owners (i.e. their API consumers) to have control over.

Some concrete examples:

1. Which VPC an RDS instance should be created in. Imagine a platform team that manages 10 Kubernetes clusters. Each cluster runs Crossplane, and each cluster is in its own VPC. Any databases used by each cluster should be created in its VPC. Currently this platform team could not use a single database Composition across all clusters unless they exposed the target VPC as an XR (and thus claim) field - i.e. left it up to their API consumers to set.
2. Which cloud provider credentials should be used to create managed resources. Imagine a platform team that runs one Kubernetes cluster. The cluster runs Crossplane, and each of the teams the platform team supports have their own namespace within said cluster. Each team also has their own set of cloud credentials. All teams should be able to create database XRs, but they must do so using their own credentials. Currently this is impossible. The platform team could create a ProviderConfig for each team, but they would then need to either let their API consumers specify which PlatformConfig to use (without restriction) or fork their database Composition to reflect each team's ProviderConfig (teams could again select which Composition to use without restriction).

### How could Crossplane help solve your problem?

Crossplane could allow Compositions to patch from common Kubernetes data sources, possibly including ConfigMaps, Secrets, and (claim) Namespaces. Doing so would enable the above use cases:

1. The platform team can add a well known ConfigMap or Secret to each Kubernetes cluster - perhaps a ConfigMap named `cluster-info` in the `crossplane-system` namespace, then write a Composition that patched from `data.vpc-name` within that ConfigMap.
2. The platform team can annotate each team's namespace with their ProviderConfig name, then write a Composition that patched from `metadata.annotations[crossplane.io/provider-config` on the claim's namespace through to `spec.providerConfigRef.name` on each composed resource.

Note that this enhancement needs a little more thought. Specifically I'm wondering:

1. Do we need this at the Composition layer if we allow arbitrary cross resource references at the managed resource layer per  https://github.com/crossplane/crossplane/issues/1770. i.e. Could you cover at least the first use case by writing your Composition such that each managed resource's `spec.forProvider.vpcName` was a cross-resource-reference to a ConfigMap?
[...]

PR #2133, opened 2021-02-12
Add support for required fromFieldPath patches

negz:

Fixes https://github.com/crossplane/crossplane/issues/2094
Fixes https://github.com/crossplane/crossplane/issues/2091
Enables https://github.com/crossplane/crossplane/issues/2099#issuecomment-768301167

Current a fromFieldPath patch is a no-op if the specified 'from field path' does not actually exist. This commit allows such patches to be marked as 'Required', causing them to return an error if the field path they're patching from does not (yet) exist. This can be used to block the creation of a composed resource until the field path exists on the composite resource.

I have:

I've created an XR using a Composition with a mix of Required and Optional fieldpath patches. When I require a non-existent field path I see the following:

```console
Events:
  Type     Reason             Age              From                                                             Message
  ----     ------             ----             ----                                                             -------
  Normal   SelectComposition  2s (x4 over 2s)  defined/compositeresourcedefinition.apiextensions.crossplane.io  Successfully selected composition
  Warning  ComposeResources   2s (x4 over 2s)  defined/compositeresourcedefinition.apiextensions.crossplane.io  cannot render composed resource at index 0: cannot apply the patch at index 2: spec.team: no such field
```


apis/apiextensions/v1/composition_types.go
```diff
@@ -213,6 +222,16 @@ type Patch struct {
...
 
+	// FromFieldPathStrategy determines how to patch from a field path. The
+	// default strategy is 'Optional', which means the patch will be a no-op if
+	// the specified fromFieldPath does not exist. Use the 'Required' strategy
+	// if the patch should fail if the specified fromFieldPath does not exist.
+	// The Required strategy can be useful to ensure that a composed resource
+	// is not rendered until the specified fromFieldPath is populated.
+	// +kubebuilder:validation:Enum=Optional;Required
+	// +optional
+	FromFieldPathStrategy *FromFieldPathStrategy `json:"fromFieldPathStrategy,omitempty"`
```
hasheddan, line 233: This is kind of related to @benagricola question on https://github.com/crossplane/crossplane/pull/2134#discussion_r575077878, but have you thought about how this works patching back the other direction? Do we want to also error if the `FromFieldPath` on the `Composition` that we are trying to patch back to the XR / XRC is not present? I don't think it is necessarily a bad idea to error in that case, but it would have different implications because the managed resources could exist but we would still be erroring out.

apis/apiextensions/v1/composition_types.go
```diff
@@ -213,6 +222,16 @@ type Patch struct {
 	// +optional
 	FromFieldPath *string `json:"fromFieldPath,omitempty"`
 
+	// FromFieldPathStrategy determines how to patch from a field path. The
```
muvaf, line 225: Is there a specific reason this isn't named `Strategy`? I can see this strategy being applied for all operations of the patch, i.e. not only fetching the value from field path but also not being able to write the value to the destination.

negz: I think there's value in distinguishing the source and destination here. For example I could imagine "to field path" strategies like "Replace" vs "Merge". I considered trying to model that all under a `strategy` key too, like:

```yaml
  fromFieldPath: metadata.annotations
  toFieldPath: metadata.annotations
  strategy:
    fromFieldPath: Required
    toFieldPath: Merge
```

I figured in this case that would look instead like:

```yaml
  fromFieldPath: metadata.annotations
  fromFieldPathStrategy: Required
  toFieldPath: metadata.annotations
  toFieldPathStrategy: Merge
```

I don't feel strongly which one we go with though.

apis/apiextensions/v1/composition_types.go
negz: I'm not quite following you here.

A `ToCompositeFieldPath` type patch will patch from the composed resources (i.e. the managed resources) back to the XR - not from the `Composition`. This PR will therefore let you declare that the composed resource field path you're patching from is required to exist, just the same as if you were patching in the other direction. We're not changing a default or anything, so if you don't want the field to be required you can just stick with it being optional.

hasheddan: Sorry when I said `Composition` I meant the managed resource that was rendered out. My point here is just that it is a different case when you are patching back to the XR and could be more difficult to debug than patching to managed resource. Maybe I am missing some context about when patching actually takes place when patching to the XR though. Is it before the managed resource actually gets created? Once again, not saying we shouldn't do this, just trying to understand the implications.

apis/apiextensions/v1/composition_types.go
benagricola: I agree that strategy could be different between the source and the destination, so separating them makes sense. 

I find the first example a little easier to read as the fields look very similar to the existing fields in the second example.

Is there a possibility that future patch types might not require either the `fromFieldPath` or `toFieldPath` settings but still want to define a strategy for whatever their source or destination is?

If so, maybe the field names could be more generic (e.g. `sourceStrategy` or `fromStrategy`).

The only example I could think of there would be a patch type that retrieves something from entirely outside the k8s API, or something that reads an entire resource (so does not use `fromFieldPath`), but you might still want to define a strategy based on the existence of that source. I don't know how likely that is but thought it worth mentioning just in case 😃

muvaf: It feels like `Optional|Required` choices are not in the same class with `Merge|Replace`. Like I might want a patch `toFieldPath` to be `Required` as well as `Merge`. Also, I'm just realizing that `patch strategy` is actually used in other contexts, like patch strategies in kubectl. So, it feels like a different name and field for `Optional|Required` choice could make sense. The `strategy` would be preserved for the actual strategy of the patch, i.e. _how_ the patch should be done instead of whether to enforce it or not.

negz: > I find the first example a little easier to read as the fields look very similar to the existing fields in the second example.

I'm happy to take that approach instead.

> Is there a possibility that future patch types might not require either the fromFieldPath or toFieldPath settings but still want to define a strategy for whatever their source or destination is?

There is, but I figured we could add strategies as appropriate, rather than trying to come up with a generic concept before we know what exactly it's for. I prefer a "from [field path] strategy" over a "source strategy" because the similar naming makes the relationship between the two fields obvious. I don't think it would be so bad that if in future we added a `type: FromWidget` patch we add a `strategy.fromWidget` next to the `strategy.fromFieldPath`.

> It feels like Optional|Required choices are not in the same class with Merge|Replace. Like I might want a patch toFieldPath to be Required as well as Merge.

"Strategy" is a pretty open ended word. To me it can fit both "what is our strategy for reading the value" (e.g. is it required) and "what is our strategy for writing the value" (e.g. should it be merged). The concepts "read" and "write" in our API are usually something like "from" and "to".

I'm going to update the API to the "nested" option above for now. I don't want to quash debate around the best API here, but I would love to get this merged before our feature freeze on Monday if possible. Let me know if anyone has any concerns with the proposed approach - ideally with a counterproposal as to how this might work. :)

apis/apiextensions/v1/composition_types.go
hasheddan: @negz I spent some more time looking into this today and I believe I have confirmed by assumption about how this works. It appears that we will error out rendering composed managed resources in the case we have a missing required from field that is being patched back to the XR:
https://github.com/crossplane/crossplane/blob/5f239dbb7c399a8d544518be2be23ad16f98a71d/internal/controller/apiextensions/composite/reconciler.go#L471

I believe this means that we would not be able to do any required patching from the composed managed resources to the XR in the case where the from field was in the status or is late initialized. I guess we are just saying that these types of fields should not be marked as required patches, but it may be nice to instead say something like a resource cannot be considered ready until the field exists. Although I guess a readiness check should be used in that case. Anyway, you likely already considered this but may be worth documenting that it is a rare case when someone would want a required patch from the managed resource to the XR.

negz: Got it. The concern is that we might get stuck in a 'chicken and egg' situation in which we require a patch from a late initialized composed resource field, and thus we'll error out before we ever create the managed resource. I've pushed a commit that should address this - let me know if you think this needs more attention.

apis/apiextensions/v1/composition_types.go
negz: > It feels like Optional|Required choices are not in the same class with Merge|Replace. Like I might want a patch toFieldPath to be Required as well as Merge.

It just occurred to me that we might be interpreting this differently. Are you thinking that `FromFieldPathStrategy` would be the strategy used for all patches of `type: FromCompositeFieldPath`, while `ToFieldPathStrategy` would be the strategy used for all patches of `type: ToCompositeFieldPath`?

My intention was that these strategies would correspond more the `fromFieldPath:` and `toFieldPath:` fields, regardless of which patch type is used. If you look at it this way I think the current design satisfies your criteria, e.g.:

```yaml
# The 'from' path is required, and we merge into the 'to' path.
  fromFieldPath: metadata.annotations
  toFieldPath: metadata.annotations
  strategy:
    fromFieldPath: Required
    toFieldPath: Merge
# Same as the above, but in the other direction.
- type: ToCompositeFieldPath
  fromFieldPath: metadata.annotations
  toFieldPath: metadata.annotations
  strategy:
    fromFieldPath: Required
    toFieldPath: Merge
```

apis/apiextensions/v1/composition_types.go
hasheddan: @negz I think your commit has solved the primary issue, but my initial concern was actually around a more subtle one. Your commit solves the case where we would refuse to create _any_ resources, but I was thinking of the case where we just refuse to create _some_ resource because of an error with a missing required patch from a different managed resource to the XR. Let me outline the logic to see if I am understanding this correctly:
1. First we render the composed managed resources. Your commit addresses the case where we error out here from a missing from field that is required but is not yet populated.

https://github.com/crossplane/crossplane/blob/5f239dbb7c399a8d544518be2be23ad16f98a71d/internal/controller/apiextensions/composite/reconciler.go#L422

2. Next we go through a loop of those rendered managed resources.

https://github.com/crossplane/crossplane/blob/5f239dbb7c399a8d544518be2be23ad16f98a71d/internal/controller/apiextensions/composite/reconciler.go#L441

3. In that loop, we first `Apply` the rendered managed resource, then fetch connection details, then check if it is ready, then finally.. we render the composite based on any patches from the managed resource to the XR. This is where we can still error on a missing required from field.

https://github.com/crossplane/crossplane/blob/5f239dbb7c399a8d544518be2be23ad16f98a71d/internal/controller/apiextensions/composite/reconciler.go#L471

4. This makes sense, but importantly we don't just let the user know about that error, we _stop creating subsequent composed managed resources_. This means that order of the templates in a `Composition` now matters. If a managed resource with a required from field patch to the XR depends on (i.e. has a cross-resource reference to or just on the provider needs it to be created) a subsequent managed resource, that subsequent managed resource will never get created. For example, this is very contrived, but if we had a `Subnetwork` that references a `Network` and exists before that `Network` in a `Composition` and we had a required from field path patch from the `Subnetwork` to the `XR` on a field that gets populated after the `Subnetwork` is created successfully, we would never resolve because we would error out in the `Apply` loop on the missing from field, meaning the `Network` does not get created, so the `Subnetwork` will never have the field populated, and we will be stuck.

As mentioned before, this is a specific case, and it is not necessarily _wrong_, but it is a design decision. If we keep it as-is we should likely document a best practice of always having a resource that is dependent on another come after it in a `Composition` (users seem to already do this by default, but I don't think we don't specifically say to do so). Alternatively, it seems like we could do the patching from composed managed resources back to the XR in a separate loop (i.e. not coupled to the creation of the managed resources) and we would avoid this issue.

apis/apiextensions/v1/composition_types.go
muvaf: > It just occurred to me that we might be interpreting this differently. Are you thinking that FromFieldPathStrategy would be the strategy used for all patches of type: FromCompositeFieldPath, while ToFieldPathStrategy would be the strategy used for all patches of type: ToCompositeFieldPath?

Not really, I think we're on the same page.

My worry is that we might want `toFieldPath` to have both `Required` and `Merge` strategies so to say, because the possible `strategy` types we're expecting to exist even now (merge, required, optional, replace) are not exclusive; that's why having those types under one umbrella could limit us.

The following is an example of how we can make `strategy` specific to patching operation strategy (`setValue`) we'll follow and have more room for details about the chosen strategy. This specific example shows how to merge arrays and what to use as identifier to decide which element should merge to which. The field `required` would then capture the whole pipeline, including fetching, transform operations and setting the final value; if any of them fails, `required` dictates that we can't continue.
```yaml
  fromFieldPath: spec.someArray
  toFieldPath: spec.some.otherArray
  required: true
  strategy:
    type: Merge
    merge:
      identifier: ".name"
```

Additionally, in this way we tie these properties to the common operations of all kinds of patches we do (get, transform, set) rather than fields (`fromFieldPath`, `toFieldPath` or even `fromMultipleFieldPaths`) which may change with the type.

negz: Let me try take a step back.

My aim with this PR is to allow Composition authors to mark the `fromFieldPath` of a patch as required. This is something that we're confident would be helpful to people, for example by ensuring we wait until an XR's `spec.claimRef.namespace` is set before we try to patch from it.

I didn't pick the name "strategy" with the API server concept of a "patch strategy" (e.g. merge patch) in mind, but I can see the confusion there. I picked it because it's a pretty generic way to say "how to do a thing", and [API conventions recommend](https://github.com/kubernetes/community/blob/6254e11696ac0886d554021a27c8179b459502c6/contributors/devel/sig-architecture/api-conventions.md#primitive-types) avoiding the obvious route of adding a `required: true` boolean:

> Think twice about bool fields. Many ideas start as boolean but eventually trend towards a small set of mutually exclusive options. Plan for future expansions by describing the policy options explicitly as a string type alias (e.g. TerminationMessagePolicy).

Now as we try to figure out how we should model whether or not the `fromFieldPath` is required we're factoring in some other functionality we _may_ want to add in future, including allowing Composition authors to control:

1. How the `toFieldPath` should be patched - e.g. merged or replaced.
1. Whether the `toFieldPath` should be required. (I'm guessing this means "only patch this field if it already exists"?)
1. Whether other patch-related-stuff like transforms are required.

I think we have varying levels of confidence that folks will want to control these things. Personally, 1 seems pretty likely to me but I can't immediately think of any reason someone would want to control 2 or 3. Given this, I'd like to find a way to let folks control whether the `fromFieldPath` is required that doesn't conflate the requiredness of the `fromFieldPath` with the requiredness of other parts of the patch, but that also leaves a path open to modelling everything we've touched on in this issue if and when we need to.

Perhaps we can move forward here by calling the 'strategy' field 'policy' or something equally generic that folks won't associate with API server patch strategies. So something like:

```yaml
  fromFieldPath: spec.claimRef.namespace
  toFieldPath: spec.providerConfigRef.name
  policy:
    fromFieldPath: Required
```

This avoids conflating 'the from field path is required' with 'everything about this patch is required'. It also leaves open space to separately model everything else we've discussed, e.g.:

```yaml
  fromFieldPath: spec.claimRef.namespace
  toFieldPath: spec.providerConfigRef.name
  policy:
    fromFieldPath: Required
    toFieldPath: Required
    transforms: Required
    someOtherThingWeHaventThoughtOfYet: Optional
  strategy:
    type: Merge
    merge:
      identifier: ".name"
```

apis/apiextensions/v1/composition_types.go
negz: Fair point.

My hunch is that required `ToCompositeFieldPath` patches will be rare or non-existent. The only use case that I'm aware of for required patches is to block the creation of the patched resource in order to ensure the `toFieldPath` isn't late initialised to an undesirable value.  e.g. We want to make sure a composed resource's `spec.forProvider.name` doesn't get defaulted to `default` if we're patching from a late-initialized XR field like `spec.claimRef.namespace`. Given that the XR will always exist already when we're patching from a composed resource I can't think of any reason someone would want to use `ToCompositeFieldPath`.

So I suspect that this case in which a required `ToCompositeFieldPath` patch prevents the XR from ever rendering successfully is fairly hypothetical. That said - your suggested fix of breaking up the apply-then-assess loop into multiple loops seems like a low-risk way to ensure we never need to worry about it.

apis/apiextensions/v1/composition_types.go
muvaf: > I'd like to find a way to let folks control whether the fromFieldPath is required that doesn't conflate the requiredness of the fromFieldPath with the requiredness of other parts of the patch, but that also leaves a path open to modelling everything we've touched on in this issue if and when we need to.

The cases where transform interact with external systems or many inputs could be candidate for this. I can't really come up with an example for requiredness of `setValue`, i.e. `toFieldPath`, other than schema error like you try to set an array to a string field. So, my thought was that if I want `fromFieldPath` to be required, chances are pretty high that I want other operations to be required as well because requiredness for me is about having the final value in the managed resource and blocking if that doesn't happen, so transform/`setValue` failures should also block the rendering.

image

This looks pretty good to me. My main worry was that strategies and requiredness are not mutually exclusive to be in the same field and this separation seems to solve it pretty well. I can't really think of much scenarios but more granularity about which operations are required could also come in handy.

hasheddan (approved):
LGTM! Thanks for this @negz :)

muvaf (approved):

muvaf:
@negz Thank you for getting that in v1.1, it'll be very helpful for multi-tenant users!

PR #2134, opened 2021-02-12
Cleanup Composition 'Type' discriminator fields

negz:

This PR:

* Copies a few backward compatible changes over from apiextensions/v1 to apiextensions/v1beta1
* Makes sure all Composition `FooType` types are named `FooType`, not `TypeFoo`.
* Renames the "Value" `ConnectionDetailType` to "ValueFrom".
* Marks 'value' connection details in our examples as such

I have:

I haven't - they're all pretty trivial.

### DIFF
--- a/apis/apiextensions/v1/composition_types.go
+++ b/apis/apiextensions/v1/composition_types.go
@@ -562,7 +562,7 @@ type ConnectionDetailType string
 const (
 	ConnectionDetailTypeFromConnectionSecretKey ConnectionDetailType = "FromConnectionSecretKey" // Default
 	ConnectionDetailTypeFromFieldPath           ConnectionDetailType = "FromFieldPath"
-	ConnectionDetailTypeValue                   ConnectionDetailType = "Value"
+	ConnectionDetailTypeFromValue               ConnectionDetailType = "FromValue"
 )
 
 // ConnectionDetail includes the information about the propagation of the connection
@@ -578,7 +578,7 @@ type ConnectionDetail struct {
 	// connection detail type may require its' own fields to be set on the
 	// ConnectionDetail object.
 	// +optional
-	// +kubebuilder:validation:Enum=FromConnectionSecretKey;FromFieldPath;Value
+	// +kubebuilder:validation:Enum=FromConnectionSecretKey;FromFieldPath;FromValue
 	// +kubebuilder:default=FromConnectionSecretKey
 	Type ConnectionDetailType `json:"type,omitempty"`
 
--- a/apis/apiextensions/v1/composition_types.go
+++ b/apis/apiextensions/v1/composition_types.go
@@ -560,7 +560,8 @@ type ConnectionDetailType string
 
 // ConnectionDetailType types.
 const (
-	ConnectionDetailTypeFromConnectionSecretKey ConnectionDetailType = "FromConnectionSecretKey" // Default
+	ConnectionDetailTypeUnknown                 ConnectionDetailType = "Unknown"
+	ConnectionDetailTypeFromConnectionSecretKey ConnectionDetailType = "FromConnectionSecretKey"
 	ConnectionDetailTypeFromFieldPath           ConnectionDetailType = "FromFieldPath"
 	ConnectionDetailTypeFromValue               ConnectionDetailType = "FromValue"
 )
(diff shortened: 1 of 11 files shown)


docs/getting-started/package-infrastructure.md
```diff
@@ -390,7 +390,8 @@ spec:
         - fromConnectionSecretKey: username
         - fromConnectionSecretKey: password
         - fromConnectionSecretKey: endpoint
-        - name: port
+        - type: FromValue
+          name: port
```
negz, line 394:
```console
  Warning  ComposeResources   29m (x12 over 34m)    defined/compositeresourcedefinition.apiextensions.crossplane.io  connection detail of type "FromConnectionSecretKey" key is not set
```

I noticed this change was needed because when I applied one of our doc snippets it was giving me the above warning, which means technically introducing a `type` here is a breaking behaviour change. I wonder if we're willing to tolerate it? I'm slightly tempted to, because my guess is that 'FromValue' connection details are not super common.

I believe the impact of including this change would be that XRs that used a `FromValue` connection detail would cease to successfully compose resources until their Composition was updated to be specific about its connection detail type.

apis/apiextensions/v1beta1/composition_types.go
```diff
@@ -144,11 +145,11 @@ type Patch struct {
...
 	// +optional
-	// +kubebuilder:validation:Enum=FromCompositeFieldPath;PatchSet
+	// +kubebuilder:validation:Enum=FromCompositeFieldPath;PatchSet;ToCompositeFieldPath
 	// +kubebuilder:default=FromCompositeFieldPath
 	Type PatchType `json:"type,omitempty"`
 
-	// FromFieldPath is the path of the field on the upstream resource whose value
+	// FromFieldPath is the path of the field on the composed resource whose value
 	// to be used as input. Required when type is FromCompositeFieldPath.
 	// +optional
```
benagricola, line 154: `FromFieldPath` and `ToFieldPath` are used bi-directionally so could each refer to a field on either the composite or composed resource.

```suggestion
	// FromFieldPath is the path of the source field on the composed or 
	// composite resource whose value is to be used as input. Required 
	// when type is FromCompositeFieldPath or ToCompositeFieldPath.
	// +optional
```

docs/getting-started/package-infrastructure.md
muvaf: This is because now the default is `FromConnectionSecretKey`, right? I agree not many people use it but I still don't really feel comfortable breaking this in `v1`, it just doesn't feel like something important enough that's worth breaking our contract with the community. Can we work around it by having the defaulting mechanism in the code, instead of kubebuilder? Like when `type` is not given: `if value is given, the type is FromValue, else assume type is FromConnectionSecretKey`. It's hacky, but it'd be a fairly contained hack in the codebase.

krishchow: This was the behavior we had previously @muvaf, but I think there was a discussion about removing this behavior in order to make the connection detail more explicit. I think this is overall a positive change since trying to imply the type a user is intended could lead to unintended behavior.

The other breaking change here is the move to `FromValue` instead of `Value`. Although this is more consistent with the names of the other fields, I would argue it is not as valuable of a chance if we are considered breaking/updating the API version contract.

negz: > The other breaking change here is the move to FromValue instead of Value

I don't think this is a breaking change. The field for specifying a value is still `value:` - I just renamed one of the possible values of the `type:` field from `Value` to `FromValue`. This type field only exists in `master` (it's unreleased) so I don't think we need to consider renaming it to be a breaking change.

> Can we work around it by having the defaulting mechanism in the code, instead of kubebuilder?

I was thinking something similar. I'll add a commit to this PR.

apis/apiextensions/v1beta1/composition_types.go
negz: Good point - I'll fix that.

docs/getting-started/package-infrastructure.md
negz: This is now done.

apis/apiextensions/v1beta1/composition_types.go
negz: Done!

apis/apiextensions/v1/composition_types.go
```diff
@@ -346,6 +347,7 @@ const (
 type Transform struct {
 
 	// Type of the transform to be run.
+	// +kubebuilder:validation:Enum=map;math;string;convert
```
hasheddan, line 350: Just noting here while this is technically restricting the allowed value for this field at the API server, we are currently erroring out before creating _any_ composed managed resources, so this change should not break any current consumers of this functionality 👍

hasheddan (approved):
LGTM! Thanks @negz! Left a note (mainly for myself), that may be worth including in release notes :)

PR #2137, opened 2021-02-13
Tinker with the XR Reconciler

negz:

Depends on https://github.com/crossplane/crossplane/pull/2133 - only the most recent commit is unique to this PR.

This PR refactors the XR reconciliation flow a little. It's functionally identical to the current logic, but tries to tie a few recent changes from @benagricola, @mcavoyk, and myself together a little more cohesively. 

I have:

I've given this a quick spin by applying the "getting started with GCP" configuration and claim.


apis/apiextensions/v1/composition_types.go
```diff
@@ -73,38 +74,44 @@ type CompositionSpec struct {
 	WriteConnectionSecretsToNamespace *string `json:"writeConnectionSecretsToNamespace,omitempty"`
 }
 
-// InlinePatchSets dereferences PatchSets and includes their patches inline. The
-// updated CompositionSpec should not be persisted to the API server.
-func (cs *CompositionSpec) InlinePatchSets() error {
+// ComposedTemplates processes and returns a Composition's resource templates,
+// defereferncing and inlining any PatchSets.
+func (cs *CompositionSpec) ComposedTemplates() ([]ComposedTemplate, error) {
```
negz, line 79: The idea here is to derive the merged templates from the Composition rather than sort of using the Composition as a 'transport' for them. Mostly just to avoid mutating the Composition and thus the (negligible) risk that we'd commit it back to the API server unintentionally.

apis/apiextensions/v1/composition_types.go
```diff
@@ -197,6 +204,29 @@ type ReadinessCheck struct {
 	MatchInteger int64 `json:"matchInteger,omitempty"`
 }
 
+// Patches that may be applied to a composite or composed resource.
+type Patches []Patch
+
+// Apply a set of patches, optionally filtered by patch type.
+func (p Patches) Apply(composite runtime.Object, composed runtime.Object, only ...PatchType) error {
```
negz, line 211: The idea here is to lift the filtering of patch types up to the 'array of patches' level, rather than potentially processing the filters on each patch operation.

apis/apiextensions/v1/composition_types.go
```diff
@@ -267,37 +297,18 @@ type PatchPolicy struct {
 
 // Apply executes a patching operation between the from and to resources.
 // Applies all patch types unless an 'only' filter is supplied.
-func (c *Patch) Apply(from, to runtime.Object, only ...PatchType) error {
-	if c.filterPatch(only...) {
-		return nil
-	}
-
+func (c *Patch) Apply(composite, composed runtime.Object) error {
```
negz, line 300: Sadly "from and to" doesn't make sense in a world of bidirectional patching.

negz (commented):
I'd like to test a few more scenarios (i.e. patchsets, bidir patching) before (and if) this is merged but I think it's ready for review.

negz:
This is pretty stale and not super important. I might try tackle some of it next time I'm working on composition things.

PR #2146, opened 2021-02-17
Pedantically re-re-refactor the fieldpath.IsNotFound check when patching

negz:

I refactored this check to account for the new patch policy, and broke it while doing so. Dan refactored it to fix my refactor. This commit just removes the need to return a tuple, pretty much returning to the original API.

I have:

I haven't manually tested this.


negz:
No need to backport this (if it's worth merging at all). It's mostly just a result of my strange aversion to `else` clauses.

hasheddan (approved):
This looks like a simpler, clearer implementation. Thanks for the improvement @negz!

PR #2153, opened 2021-02-21
compress images for fast load Readme

airdry:
I wanted to suggest a little optimization of the images in the project, for more loading.
At the moment, when you open Readme, the logo is 300 KB. After compression we get the same logo but with a weight of 50 KB.


hasheddan (changes requested):
Hey @airdry! Thanks for this optimization! If we are going to do this I would like to make sure we have a consistent workflow (and optimally enforcement of it in CI) so that all of our images are stored in the same fashion. Would you mind adding the steps you went through to make these changes? Also, it would be preferable for the provided PR template to be used as well :)

airdry:
This is a minor improvement for the project and documentation.
Optimization procedure:
- Check content in devtool under Network tab and img parameter.
- Compress all files larger than 200KB in the convector (https://www.websiteplanet.com/webtools/imagecompressor/)
- Replaced current files with new ones.

Now there are not so many files, and in order not to configure this process in CI, I made compression for all the pictures in the project.
In the future, it would be great to automate everything.

Issue #2156, opened 2021-02-23
Init container to upgrade Crossplane CRDs


muvaf:
### What problem are you facing?

Helm does not upgrade CRDs when you run `helm upgrade` but we need the new CRDs to ship new features.

### How could Crossplane help solve your problem?

Current workaround is to `kubectl apply` new CRDs and then do `helm upgrade` but that's OK only when there isn't a breaking change.

We can have an init container that will update the CRDs of Crossplane. We are able to do this because Crossplane already has RBAC necessary for managing CRDs. This init container will also be used in webhook server TLS certificate generation for both core Crossplane CRD conversions as well as provider CRDs.

PR #2160, opened 2021-02-23
Crossplane to manage its own CRDs

muvaf:

This PR adds initialization functionality to core Crossplane process. The steps are:
* Install (or update) all Crossplane CRDs.
* Create `Lock` object.
* Create all `Provider` and `Configuration` packages that are given for preinstallation.

While core Crossplane is doing all these things, RBAC manager has to wait for a few seconds until `CompositeResourceDefinition` and `ProviderRevision` CRDs exist. To do that, it uses an initializer that has only one step that waits for these CRDs.

This will allow Crossplane not depend on Helm to handle CRDs especially during the update. Otherwise, we'd require users to run a `kubectl apply`. Additionally, this initializer mechanism will be useful for generating the TLS Cert for webhook servers.

Fixes https://github.com/crossplane/crossplane/issues/2156

Fixes #

I have:

* Normal installation.
* Install v1.0.0, install provider-aws, provision a `VPC`. Then upgrade using `cluster/local/kind.sh helm-upgrade`.
* Install v1.0.0, follow quick start [composition](https://crossplane.io/docs/v1.0/getting-started/compose-infrastructure.html), then upgrade using `cluster/local/kind.sh helm-upgrade`. Nothing seems broken. Deletion works as well.
* Installation with the following `values.yaml` to make sure pre-install is not broken:
```yaml
provider:
  packages:
    - crossplane/provider-aws:v0.16.0
    - crossplane/provider-azure:v0.14.0
configuration:
  packages:
    - registry.upbound.io/upbound/platform-ref-gcp:v0.0.1
```


internal/initializer/crds.go
```diff
@@ -0,0 +1,87 @@
...
+
+// CoreCRDs makes sure the CRDs are installed.
+type CoreCRDs struct {
+	Path string
+}
+
+// Run applies all CRDs in the given directory.
+func (c *CoreCRDs) Run(ctx context.Context, kube resource.ClientApplicator) error { // nolint:gocyclo
+	var crds []*v1.CustomResourceDefinition
+	err := filepath.Walk(c.Path, func(path string, info os.FileInfo, err error) error {
```
hasheddan, line 47: @muvaf can we reuse the parser implementation [here](https://github.com/crossplane/crossplane-runtime/blob/30a941c3c3c67e4dc064a2a170cb725de40b0271/pkg/parser/parser.go#L93)? It has support for a filesystem backend that would accommodate this use-case. We use it currently to build the packages in the CLI: https://github.com/crossplane/crossplane/blob/d9afc165cd5335703483b4671b0ca0c751038d63/cmd/crank/build.go#L64

muvaf: I think we can. Let me see.

muvaf: We'd be able to use the backend which reads the files but not the actual parser right? We'd still do the YAML decoding ourselves here?

hasheddan: You can use the parser also, you just need to add the proper versions to the `Scheme`. ref: https://github.com/crossplane/crossplane/blob/f6ebabeda19dd31d39adfc71c0c2110ee74032ab/internal/xpkg/scheme.go#L46

cmd/crossplane/core/core.go
```diff
@@ -50,11 +54,22 @@ func FromKingpin(cmd *kingpin.CmdClause) *Command {
 	cmd.Flag("cache-dir", "Directory used for caching package images.").Short('c').Default("/cache").OverrideDefaultFromEnvar("CACHE_DIR").ExistingDirVar(&c.CacheDir)
 	cmd.Flag("sync", "Controller manager sync period duration such as 300ms, 1.5h or 2h45m").Short('s').Default("1h").DurationVar(&c.Sync)
 	cmd.Flag("leader-election", "Use leader election for the conroller manager.").Short('l').Default("false").OverrideDefaultFromEnvar("LEADER_ELECTION").BoolVar(&c.LeaderElection)
+	cmd.Flag("provider", "Pre-install a Provider by giving its image URI. This argument can be repeated.").StringsVar(&c.Providers)
+	cmd.Flag("configuration", "Pre-install a Configuration by giving its image URI. This argument can be repeated.").StringsVar(&c.Configurations)
```
negz, line 58: Is there any value in making this a separate subcommand? This way folks could decouple initialising Crossplane from actually running Crossplane if they wanted to? I can't think of many use cases for doing so but it "feels" a little cleaner to me to keep them separate.

internal/initializer/initializer.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/pkg/errors"
+	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/crossplane/crossplane-runtime/pkg/resource"
+)
+
+// NewInitializer returns a new *Initializer.
```
negz, line 33: How about just `New`, so it's called as `initializer.New()` rather than `initializer.NewInitializer`, which stutters?

internal/initializer/initializer.go
```diff
@@ -0,0 +1,78 @@
...
+// Step is a blocking step of the initialization process.
+type Step interface {
+	Run(ctx context.Context, kube resource.ClientApplicator) error
+}
+
+// Initializer makes sure the CRDs Crossplane reconciles are ready to go before
+// starting main Crossplane routines.
+type Initializer struct {
+	Steps []Step
+}
```
negz, line 50: Nit - could this type be a slice? e.g.:

```go
type Initializer []Step
```

internal/initializer/installer.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/pkg/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/crossplane/crossplane-runtime/pkg/resource"
+
+	v1 "github.com/crossplane/crossplane/apis/pkg/v1"
+)
+
+var replaceSet = map[string]string{"(": "", ":": "", "|": "", "@": "", ")": "", ".": "-", "*": "", "/": "-"}
```
negz, line 32: Nit: Maybe this could be a https://golang.org/pkg/strings/#Replacer (or two)?

If not, perhaps let the map take up multiple lines - it's a little hard to read like this.

internal/initializer/crds.go
muvaf: Oh OK, when I looked at this I saw parser returns a `*Package` struct and thought it's package-specific but it seems it's just two lists of objects. I'll use this implementation.

cmd/crossplane/core/core.go
muvaf: You're saying that we'd have to make two calls right? So, `crossplane initialize && crossplane start` in `Deployment` YAML?

For context, I added these flags just to handle pre-install mechanism we have in the Helm chart. So, they need to be part of the arguments in `Deployment` that starts Crossplane.

internal/initializer/initializer.go
muvaf: I actually made some changes to this. It used to figure out the client config on its own, now `Initializer` structs needs a client passed in. So, now it has another field.

internal/initializer/crds.go
muvaf: Just pushed a new commit that uses the parser.

muvaf:
I have completed all of the manual tests I wanted to do and updated the PR description with details.

cmd/crossplane/core/core.go
negz: Ah, right. Probably not the best place to drop my comment but yes I was thinking `crossplane initialize` and `crossplane [core]` would be separate commands. That way we could either chain them in the one container or possibly let an init container handle the initialization.

Separately, is the idea behind moving the provider and config install logic into Crossplane core that we need the Lock to exist before we do that? Could the Helm chart continue to handle installing them?

muvaf: > Probably not the best place to drop my comment but yes I was thinking crossplane initialize and crossplane [core] would be separate commands.

If we do that, we'd probably need `crossplane core initialize` and `crossplane rbac initialize` separately because `rbac` can start only if some of the CRDs exist, so it needs an initialization period.

Separating the initialization from core would help local debugging as you would not want to run the initialization logic that looks to a certain path for CRDs, that is `/crds`. Also once we have the separate command, I think it makes sense to have an init container for a more proper readiness signaling. I'll do that in this PR.

> Separately, is the idea behind moving the provider and config install logic into Crossplane core that we need the Lock to exist before we do that? Could the Helm chart continue to handle installing them?

Not sure about `Lock`, but for Helm to be able to create `Provider` and `Configuration` objects, their CRDs need to present. I guess we could keep these two CRDs in the Helm chart and still update them in initializer but I'd not want to manage CRDs in two different places. I think if we go with your suggestion of separating `initialize` command, these pre-install commands will make more sense to have as flag.

muvaf:
@negz @hasheddan added unit tests. It should be ready to go.

muvaf:
@negz @hasheddan I believe all comments are addressed. Both pods have init containers and installation of packages are done via flag on that init command.

Makefile
```diff
@@ -85,11 +85,16 @@ fallthrough: submodules
...
 
-generate: $(HELM) $(KUSTOMIZE) go.vendor go.generate gen-kustomize-crds gen-install-doc
-	@$(OK) Finished vendoring and generating
+CRD_DIR = cluster/crds
 
+crds.clean:
+	@$(INFO) cleaning generated CRDs
+	@find $(CRD_DIR) -name '*.yaml' -exec sed -i.sed -e '1,2d' {} \; || $(FAIL)
+	@find $(CRD_DIR) -name '*.yaml.sed' -delete || $(FAIL)
+	@$(OK) cleaned generated CRDs
```
negz, line 94: Will these `sed` commands work on both OS X and Linux?

cmd/crossplane/main.go
```diff
@@ -48,12 +53,26 @@ func main() {
...
+	// Note that the controller managers scheme must be a superset of the
+	// package manager's object scheme; it must contain all object types that
+	// may appear in a Crossplane package. This is because the package manager
+	// uses the controller manager's client (and thus scheme) to create packaged
+	// objects.
+	s := runtime.NewScheme()
+	utilruntime.Must(scheme.AddToScheme(s))
+	utilruntime.Must(extv1.AddToScheme(s))
+	utilruntime.Must(extv1beta1.AddToScheme(s))
+	utilruntime.Must(apis.AddToScheme(s))
```
negz, line 65: Does `utilruntime.Must` panic if these `AddToScheme` calls fail? Could we use `kingpin.FatalIfError` instead? I'd prefer to be consistent with our existing patterns.

cmd/crossplane/rbac/init.go
```diff
@@ -0,0 +1,64 @@
...
+
+	cl, err := client.New(cfg, client.Options{Scheme: s})
+	if err != nil {
+		return errors.Wrap(err, "cannot create new kubernetes client")
+	}
+	// NOTE(muvaf): The plural form of the kind name is not available in Go code.
+	i := initializer.New(cl,
+		initializer.NewCRDWaiter([]string{
+			fmt.Sprintf("%s.%s", "compositeresourcedefinitions", v1.Group),
+			fmt.Sprintf("%s.%s", "providerrevisions", pkgv1.Group),
```
negz, line 56: Do we have constants we could use (I think they're near the type definitions) instead of using `Sprintf` here?

internal/initializer/crds.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+// CoreCRDs makes sure the CRDs are installed.
+type CoreCRDs struct {
+	Path   string
+	Scheme *runtime.Scheme
+}
+
+// Run applies all CRDs in the given directory.
+func (c *CoreCRDs) Run(ctx context.Context, kube client.Client) error { // nolint:gocyclo
```
negz, line 44: Is this nolint comment needed here? The method doesn't look super complex.

internal/initializer/waiter.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+// Run continuously checks whether the list of CRDs whose names are given are
+// present in the cluster.
+func (cw *CRDWaiter) Run(ctx context.Context, kube client.Client) error {
+	beginning := time.Now()
+	ending := beginning.Add(cw.Timeout)
+	current := beginning
+	cw.log.Info(fmt.Sprintf("Started waiting for the following CRDs to be present: %v", cw.Names))
+	for {
```
negz, line 59: I think it may be possible to use a `select` here - e.g.

```go
timeout := time.After(cw.Timeout)
ticker := time.NewTicker(cw.Period)
defer ticker.Stop()

for {
	select {
	case <-ticker.C:
		// Main logic goes here. Executed every cw.Period.
	case <-timeout:
		return errors.Errorf(errFmtTimeoutExceeded, cw.Timeout.Seconds())
	}
}
```

internal/initializer/waiter.go
```diff
@@ -0,0 +1,83 @@
...
+			if kerrors.IsNotFound(err) {
+				break
+			}
+			present++
+		}
+		if present == len(cw.Names) {
+			return nil
+		}
+		if current.After(ending) {
+			return errors.New(fmt.Sprintf(errFmtTimeoutExceeded, cw.Timeout.Seconds()))
```
negz, line 77: You should be able to use `errors.Errorf` here, rather than `New` + `Sprintf`.

internal/initializer/waiter.go
```diff
@@ -0,0 +1,83 @@
...
+	log     logging.Logger
+}
+
+// Run continuously checks whether the list of CRDs whose names are given are
+// present in the cluster.
+func (cw *CRDWaiter) Run(ctx context.Context, kube client.Client) error {
+	beginning := time.Now()
+	ending := beginning.Add(cw.Timeout)
+	current := beginning
+	cw.log.Info(fmt.Sprintf("Started waiting for the following CRDs to be present: %v", cw.Names))
```
negz, line 58: I'd prefer to avoid using `Sprintf` in our log lines - we should instead leverage structured logging, e.g.:

```go
cw.log.Info("Waiting for required CRDs to be present", "names" cw.Names, "poll-interval", cw.Period)
```

I believe if we state the interval in this way we can just log at the beginning of each loop iteration.

Makefile
hasheddan: @negz yep these were verified (after some trial and error :sweat_smile:) in the providers https://github.com/crossplane/provider-aws/blob/c7c185f2dc480396d86c1342af82af89151bba55/Makefile#L70

Hoping to also fix this in the parser -- but it is unfortunately a bit less straightforward than it seems.

cmd/crossplane/rbac/init.go
muvaf: There is a comment in line 52; there is no constant for plural forms of the name of the kinds and that's what we need since CRD objects are named like `<plural kind name>.<group>`.

internal/initializer/waiter.go
muvaf: Ugh I do that every time. Thanks for catching!

internal/initializer/waiter.go
muvaf: Nice! I was not aware of that ticker utility. Changed the implementation.

internal/initializer/crds.go
muvaf: I guess it's left there because of the earlier implementation, removed.

hasheddan (approved):
This is looking good to me @muvaf, thanks for working on it! I did notice a flake with a restart on e2e tests, but it appears to be unrelated. I'll see if I can reproduce but should not block here :+1:

PR #2166, opened 2021-02-26
Update documentation to lead with composition

hasheddan:

Refactors documentation, specifically the "Getting Started" guide to lead with composition and configurations rather than providers and managed resources. Other highlights include:
- Update configuration packages to depend on Crossplane `v1.0.0`
- Update `crossplane.yaml` for each configuration package to match snippets
- Added dedicated uninstall guide (based on traffic in slack on the topic)

See individual commits for additional details.

**Some commits are CP candidates for release-1.0, all are for release-1.1**

I have:

Rendered locally with `make run_docs_local`.


docs/getting-started/install-configure.md
```diff
@@ -263,17 +265,67 @@ kubectl apply -f https://raw.githubusercontent.com/crossplane/crossplane/master/
...
+### Create a Provider Secret
+
+```console
+kubectl create secret generic aws-creds -n crossplane-system --from-file=key=./creds.conf
+```
+
+### Configure the Provider
+
+We will create the following `ProviderConfig` object to configure credentials
+for AWS Provider:
```
hasheddan, line 298: I considered leaving out all of the `ProviderConfig` steps and putting them into the configuration packages themselves so that users would only need to create the provider secret. This would make the guide slightly more streamlined, but I opted against doing this for now since it would hide an important concept. Definitely open to more thoughts on this though.

docs/getting-started/install-configure.md
```diff
@@ -201,29 +201,31 @@ curl -sL https://raw.githubusercontent.com/crossplane/crossplane/master/install.
...
+<li><a href="#azure-tab-1" data-toggle="tab">Azure</a></li>
+<li><a href="#alibaba-tab-1" data-toggle="tab">Alibaba</a></li>
 </ul>
 <br>
 <div class="tab-content">
-<div class="tab-pane fade in active" id="provider-tab-aws" markdown="1">
+<div class="tab-pane fade in active" id="aws-tab-1" markdown="1">
 
-### Install AWS Provider
+### Install Configuration Package
```
muvaf, line 221: Can we have a short description of what `Configuration` means and what it contains?

docs/getting-started/provision-infrastructure.md
```diff
@@ -1,258 +1,333 @@
...
-Crossplane allows you to provision infrastructure anywhere using the Kubernetes
-API. Once you have [installed a provider] and [configured your credentials], you
-can create any infrastructure currently supported by the provider. Let's start
-by provisioning a database on your provider of choice.
-
-Each provider below offers their own flavor of a managed database. When you
-install a provider it extends Crossplane by adding support for several "managed
-resources". A managed resource is a cluster-scoped Kubernetes custom resource
-that represents an infrastructure object, such as a database instance.
+# Compose Infrastructure
```
muvaf, line 8:
```suggestion
# Provision Infrastructure
```

docs/getting-started/provision-infrastructure.md
```diff
@@ -1,258 +1,333 @@
...
+
+In the [last section] we installed packages that extend Kubernetes with
```
muvaf, line 10: These two paragraphs don't match the current structure. For example, we did not actually `used them to provision resources on a cloud`. Maybe we can move `Configuration`-related parts to the former doc and keep a short description of what claim means and how it's similar to PVC, then point the full composition doc.

docs/getting-started/provision-infrastructure.md
```diff
@@ -1,258 +1,333 @@
...
+may be managed by application operators. The application operators are
+restricted to their team's namespace, but their SQL instances should all be
+attached to the VPC network that the infrastructure operator manages. Crossplane
+enables scenarios like this  by allowing the infrastructure operator to offer
+their application operators a _composite resource claim_ (XRC). An XRC is a
+namespaced proxy for an XR; the schema of an XRC is identical to that of its
+corresponding XR. When an application operator creates an XRC, a corresponding
+backing XR is created automatically.
+
+We use two special Crossplane resources to define and configure new XRs and
```
muvaf, line 32: These resources do not appear in the steps, so I think we can remove this section and rely on composition doc for details.

docs/getting-started/provision-infrastructure.md
```diff
@@ -1,258 +1,333 @@
...
+installed to Crossplane by creating a declarative `Configuration` resource, or
+by using `kubectl crossplane install configuration`. In the examples below we
+will install a configuration that defines a new `CompositePostgreSQLInstance` XR
+that takes a single `storageGB` parameter, and creates a connection `Secret`
+with keys for `username`, `password`, and `endpoint`. A `Configuration` exists
+for each provider that can satisfy a `PostgreSQLInstance`. Let's get started!
+
+## Claim Your Infrastructure
+
+The `Configuration` package we installed in the last section:
```
muvaf, line 52: Maybe this part would be more eligible for the step where we install the `Configuration` to explain what's going to happen when they run `kubectl crossplane install`

docs/getting-started/package-infrastructure.md
```diff
@@ -1,13 +1,13 @@
 ---
 title: Package Infrastructure
 toc: true
-weight: 6
+weight: 3
 indent: true
 ---
 
 # Package Infrastructure
```
muvaf, line 8: I feel like what we package is not infrastructure, but templates for the infrastructure. In other words, `packaging infrastructure` would mean literally zipping the managed resource YAMLs for me whereas here what we package is `Composition`s and XRDs

muvaf (commented):
@hasheddan Thanks for getting this up before the release! It greatly simplifies the quick start guide.

I'm wondering if we can do better about composition parts. When people ask about `Composition`, I had had to reference the `Package Infrastructure` section because that was the only place where they get to see a simple XRD and simple `Composition`s, however, it was from packaging perspective so it didn't have claim or usage instructions. If I point them to `Compose Infrastructure`, well, it doesn't really show how to compose but how to use a ready-made composed infrastructure.

From that point of view, I think we still lack an end-to-end *and* simple guide for composition. `Provision Infrastructure` is geared towards consumption, `Compose Infrastructure` is geared towards packaging even though it has the composition YAMLs. The actual composition concept guide is very long purposefully because it acts like an API reference to some extent.

I know we're trying to optimize for magical experience and `Configuration` installation seems to be the fastest lane to get there, so I don't really have a good proposal while  keeping that magical experience. Just wanted to point out a pain point that I've been experiencing for quite a while.

docs/getting-started/install-configure.md
muvaf: I think that's a good call.

muvaf:
@negz I think this might be in the class of PRs that needs your attention since it's where all new people come to try Crossplane, hence has high impact.

hasheddan:
> I know we're trying to optimize for magical experience and Configuration installation seems to be the fastest lane to get there, so I don't really have a good suggestion while keeping that magical experience. Just wanted to point out a pain point that I've been experiencing for quite a while.

@muvaf I agree with this sentiment. I think it might be nice to eventually have a longer version and let folks choose which they prefer, or annotate the shorter one so folks can click "tell me more about this" in each section and it will expand with more content.

docs/getting-started/provision-infrastructure.md
hasheddan: I think just eliminating `used them to provision resources on a cloud` would be sufficient to make this work, but I am open to moving these to another doc if desired.

docs/getting-started/package-infrastructure.md
hasheddan: Hmm I see what you mean.. I am not sure what a good alternative is here, maybe something like `Package Your Platform`? I think the motivation for `Package Infrastructure` is packaging your infrastructure *APIs*. Maybe we could literally say `Package Infrastructure APIs`

docs/getting-started/provision-infrastructure.md
hasheddan: I am kind of torn on this because it feels super critical to explaining the rationale and machinery around cluster-scoping. I agree we are being a little inconsistent in regards to wanting the "magical" experience then doing a fair amount of under the hood explaining here.

docs/getting-started/provision-infrastructure.md
hasheddan: How do we feel about just moving this entire intro section to: https://github.com/crossplane/crossplane/pull/2166/files#diff-fd6af16be5d1d97ecd20ef62757aed21bc0ddc1a30a4b320e2d4b343af453610R204

(which I think is roughly what @muvaf is suggesting here)

docs/getting-started/package-infrastructure.md
negz: I think I named this to stick with the `<Verb> Infrastructure` theme. If we want to break that theme something like "Create a Configuration" or "Create a Composition" could make sense.

docs/getting-started/provision-infrastructure.md
```diff
@@ -1,258 +1,333 @@
 ---
 title: Provision Infrastructure
 toc: true
-weight: 3
+weight: 2
```
negz, line 4: It looks like "Install & Configure" is also at `weight: 2`. Should it be at `weight: 1` to ensure it appears before this?

docs/getting-started/provision-infrastructure.md
negz: I tend to agree that explanation of what XRDs and Compositions are belongs close to where we install them for the first time; I don't think it ruins the magic to give a little detail with regard to what exactly the Configurations will be installing.

I'd introduce the concept of a claim here, close to where we install them, if we can factor that out without a huge rewrite. So I think that would mean pretty much everything except the "XRs are always cluster scoped - ..." paragraph moves back to the Install & Configure docs.

docs/guides/direct-managed.md
```diff
@@ -0,0 +1,280 @@
+---
+title: Managed Resources
+toc: true
+weight: 250
+indent: true
+---
+
+# Using Managed Resources Directly
```
negz, line 8: I think we need a little more detail around _why_ managed resources are cluster scoped and in which cases we expect folks to use them directly. To the best of my understanding:

* We expect that application operators will _always_ consume infra via an XR and a claim.
* The only case in which we expect someone might create infra directly would be an infra operator creating supporting infrastructure like a VPC at the cluster scope.

I worry that if we keep this guide without that context folks are going to be confused about why our resources are cluster scoped. There may be some content we can copy and paste from https://blog.crossplane.io/crossplane-vs-cloud-infrastructure-addons/ to help explain this.

docs/reference/uninstall.md
```diff
@@ -0,0 +1,92 @@
+---
+title: Uninstall
+toc: true
+weight: 303
+indent: true
+---
```
negz, line 6: Thanks for adding this!

negz (commented):
Thanks for tackling this @hasheddan! Two thoughts that didn't fit near any of the actual diff:

* Should we add a little section at the top of the getting started flow that explains what we'll do to get started? i.e. Install Crossplane, configure Crossplane to offer an opinionated PostgreSQL instance API that abstracts your favourite cloud, then create one?
* Should we move detail about installing with OAM support into the OAM guide? I could go either way.

muvaf:
image
I think we need to make changes in the welcome doc as well; OAM is not included in quick start anymore so we might as well remove that section and we don't really use the terms `Infra Ops`, `App Ops` and `App Dev` anymore. I think `App Ops` was introduced by OAM, so if we want to keep these terms we can remove that one.

hasheddan:
@muvaf I agree, though that comes from the project `README.md` so I think we should update there as well. I think @negz was going to take a pass at that but I am happy to pick that up if desired 👍

docs/getting-started/install-configure.md
hasheddan: This has been addressed by moving the paragraphs from https://github.com/crossplane/crossplane/pull/2166#discussion_r585092430 to here

hasheddan:
@muvaf @negz I believe I have addressed all comments and tried to commit changes in a way the shows how they were addressed. This is ready for final review.

docs/getting-started/install-configure.md
```diff
@@ -201,29 +201,110 @@ curl -sL https://raw.githubusercontent.com/crossplane/crossplane/master/install.
...
 
+> If you prefer to see the contents of this configuration package and how it is
+> constructed prior to install, skip ahead to the [create a configuration]
+> section.
```
hasheddan, line 245: Added this note to each of the configuration package install commands so folks to jump ahead to see how these exact packages were constructed.

docs/getting-started/install-configure.md
```diff
@@ -201,29 +201,110 @@ curl -sL https://raw.githubusercontent.com/crossplane/crossplane/master/install.
...
+  `Composition` options for each composite resource.
+
+XRDs and Compositions may be packaged and installed as a _configuration_. A
+configuration is a [package] of composition configuration that can easily be
+installed to Crossplane by creating a declarative `Configuration` resource, or
+by using `kubectl crossplane install configuration`. In the examples below we
+will install a configuration that defines a new `CompositePostgreSQLInstance` XR
+that takes a single `storageGB` parameter, and creates a connection `Secret`
+with keys for `username`, `password`, and `endpoint`. A `Configuration` exists
+for each provider that can satisfy a `PostgreSQLInstance`. Let's get started!
```
muvaf, line 229:
```suggestion
for each provider that can satisfy a `CompositePostgreSQLInstance`. Let's get started!
```

This is the first time we mention the claim type and readers don't know where it comes from yet. We can stick to `CompositePostgreSQLInstance` until we talk about claims or maybe mention the claim concept here.

docs/getting-started/install-configure.md
```diff
@@ -201,29 +201,110 @@ curl -sL https://raw.githubusercontent.com/crossplane/crossplane/master/install.
...
+- A `CompositeResourceDefinition` (XRD) _defines_ a new kind of composite
+  resource, including its schema. An XRD may optionally _offer_ a claim (XRC).
+- A `Composition` specifies which resources a composite resource will be
+  composed of, and how they should be configured. You can create multiple
+  `Composition` options for each composite resource.
+
+XRDs and Compositions may be packaged and installed as a _configuration_. A
+configuration is a [package] of composition configuration that can easily be
+installed to Crossplane by creating a declarative `Configuration` resource, or
+by using `kubectl crossplane install configuration`. In the examples below we
```
muvaf, line 225: Maybe a separate paragraph starting with `In the examples ...` could be a good separation.

docs/getting-started/install-configure.md
```diff
@@ -387,120 +476,39 @@ spec:
...
-kubectl apply -f https://raw.githubusercontent.com/crossplane/crossplane/master/docs/snippets/configure/alibaba/providerconfig.yaml
-```
-
 </div>
 </div>
 
 ## Next Steps
 
-Now that you have a provider configured, you can [provision infrastructure].
+Now that you have configured Crossplane with support for `PostgreSQLInstance`,
```
muvaf, line 490: We don't know what `PostgreSQLInstance` type is yet.

docs/getting-started/provision-infrastructure.md
```diff
@@ -7,252 +7,274 @@ indent: true
...
+Composite resources (XRs) are always cluster scoped - they exist outside of any
+namespace. This allows an XR to represent infrastructure that might be consumed
+from several different namespaces. This is often true for VPC networks - an
+infrastructure operator may wish to define a VPC network XR and an SQL instance
+XR, only the latter of which may be managed by application operators. The
+application operators are restricted to their team's namespace, but their SQL
+instances should all be attached to the VPC network that the infrastructure
+operator manages. Crossplane enables scenarios like this  by allowing the
+infrastructure operator to offer their application operators a _composite
+resource claim_ (XRC). An XRC is a namespaced proxy for an XR; the schema of an
```
muvaf, line 19: I think it's worth mentioning similarity with PVC and PV. I use that a lot when I'm explaining the relationship between XRC and XR and it really helps.

docs/getting-started/create-configuration.md
```diff
@@ -1,35 +1,24 @@
...
+desired level of abstraction, and can be installed into any Crossplane cluster.
 
 ## Create a Configuration Directory
 
-Our configuration will consist of three files:
+We are going to build the same configuration package that we previously
+installed. It will consist of three files:
 
 * `crossplane.yaml` - Metadata about the configuration.
 * `definition.yaml` - The XRD.
```
muvaf, line 24: Not changed in this PR, but I think we can remove the last sentence of the next paragraph that starts with `Note that a configuration ...` because we already say `one or more XRDs or Compositions.` earlier.

docs/getting-started/create-configuration.md
```diff
@@ -657,6 +666,12 @@ metadata:
...
     provider: azure
+spec:
+  crossplane:
+    version: ">=v1.0.0-0"
+  dependsOn:
+    - provider: crossplane/provider-azure
+      version: ">=v0.13.0"
 ```
 
 ```console
```
muvaf, line 677: I think we can separate `push` command from this snippet to have its own because pushing requires credentials to be set. Also, `build` doesn't let you specify a tag like `docker build`, hence only way to push the configuration is to be in that folder. It might be worth mentioning that you tag the configuration when you push it, not during the build. We might also mention it never ends up in your local docker registry, but that may be too much information, though helpful to know if you use local cluster like `kind` to test the package.

docs/getting-started/create-configuration.md
```diff
@@ -716,7 +705,7 @@ rm -rf crossplane-config
```
muvaf, line 705: Maybe we can have an ending here with a link to `Concepts`, guiding people who want to understand everything in detail after finishing the flow here.

muvaf (commented):
Looking pretty good! I've left a few comments but I think it's ready to go after that. @negz when do you plan on opening a PR for the welcome page? I think that's also a release blocker and not addressed in this PR.

docs/getting-started/provision-infrastructure.md
hasheddan: I think it might be useful for folks to draw comparisons, but I also am cognizant of the fact that the mechanics of how XR / XRC works diverges pretty significantly from PV / PVC.

docs/getting-started/create-configuration.md
hasheddan: These are excellent points, I think they probably necessitate a separate document or additions to the package doc in `Concepts`. Would be happy to follow up with that 👍

docs/getting-started/create-configuration.md
hasheddan: Great idea!

negz (approved):

docs/getting-started/provision-infrastructure.md
negz: I could go either way on this; @hasheddan feel free to make the call.

muvaf:
Thank you a lot for getting this done in a short time with such high quality @hasheddan !!

Issue #2157, opened 2021-02-23
Use proper backoff across core controllers


hasheddan:
### What problem are you facing?

We recently [updated](https://github.com/crossplane/crossplane-runtime/pull/243) the `crossplane-runtime` [managed reconciler](https://github.com/crossplane/crossplane-runtime/blob/30a941c3c3c67e4dc064a2a170cb725de40b0271/pkg/reconciler/managed/reconciler.go#L473) to make use of the controller's configured rate limiting strategy, and provided some sane [default implementations](https://github.com/crossplane/crossplane-runtime/blob/30a941c3c3c67e4dc064a2a170cb725de40b0271/pkg/ratelimiter/default.go#L42) for controllers that are reconciling managed resources. We do not currently make use of rate limiting in most cases (other than [failed status updates](https://github.com/crossplane/crossplane/blob/d9afc165cd5335703483b4671b0ca0c751038d63/internal/controller/pkg/manager/reconciler.go#L357)) across the Crossplane core controllers. In some cases, a proper rate limiting / backoff strategy would be advantageous. A good example would be [dependency resolution](https://github.com/crossplane/crossplane/blob/d9afc165cd5335703483b4671b0ca0c751038d63/internal/controller/pkg/revision/reconciler.go#L411), which currently uses a 30s `shortWait`. This frequently leads to long waiting for a `Configuration` package to become ready because `30s` is either far too long to wait, or it resolution takes just over `30s` and we don't check again for another `30s`. An exponential backoff strategy with a higher base delay (potentially a value around `~5s`) and a relatively low capped max delay (`30s`) would likely be more appropriate for this scenario.

### How could Crossplane help solve your problem?

We should examine our current short / long delay settings and see if there are cases where we should use rate limiting / backoff.

negz:
@hasheddan adding myself as an assignee, as I'd like to help with this. Let me know if you've started anything and/or really want to tackle it yourself.

muvaf:
Is this still relevant?

PR #2167, opened 2021-02-26
Use default max concurrency for all controllers

srueg:

Use default max concurrency for all controllers.
The composite and claim reconcilers didn't configure the `MaxConcurrentReconciles`.

I have:

Locally against kind cluster.

I did some test runs where I measured the time it took from creating the composite to it being reay (all managed resources created).
Tested with 520 composite objects.
* Using a max concurrency of 1: 82s, 42s, 45s
* Using a max concurrency of 5: 66s, 82s, 45s
* Using a max concurrency of 10: 21s, 45s, 44s

So it doesn't seem to make a big difference.

### DIFF
--- a/internal/controller/apiextensions/definition/reconciler.go
+++ b/internal/controller/apiextensions/definition/reconciler.go
@@ -384,7 +384,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		)),
 		composite.WithLogger(log.WithValues("controller", composite.ControllerName(d.GetName()))),
 		composite.WithRecorder(recorder),
-	)}
+	), MaxConcurrentReconciles: maxConcurrency}
 
 	u := &kunstructured.Unstructured{}
 	u.SetGroupVersionKind(d.GetCompositeGroupVersionKind())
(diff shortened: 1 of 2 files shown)


hasheddan (approved):
This seems reasonable to me, although as @srueg mentioned in slack, it would be good to make this configurable. Performance related improvements can be tracked with https://github.com/crossplane/crossplane/issues/2157. Also would be good to have some periodic scale testing.

negz:
What's the practical difference here - i.e. what's the default when `MaxConcurrentReconciles` is unspecified?

hasheddan:
@negz I believe the default is `1` https://github.com/kubernetes-sigs/controller-runtime/blob/e388e1e1f5a745166955442294bb8aede1f761e2/pkg/controller/controller.go#L37

hasheddan:
@srueg were you able to see if this helped out in your large scale case?

srueg:
No, sorry. Didn't get around to test the change so far.

hasheddan:
@srueg you would likely need to increase max concurrent reconciles significantly for it to have a large impact. This is also subject to failed reconciles, etc.. I think this is something we should expose in configuration, but I would be fine to go ahead and bump these numbers to be consistent. Sound good?

negz:
Let's bump these. I agree configurability would be nice, we but can address that separately.

negz (approved):

PR #2180, opened 2021-03-02
Do not block composed resources on failed neighbor

### LINKED ISSUE #2175: Required patch policy aborts rendering of entire composite
Regarding discussion in Slack around the `Required` patch policy in combination with bidirectional patching.

This appears to be [intentional behaviour](https://github.com/crossplane/crossplane/blob/master/internal/controller/apiextensions/composite/reconciler.go#L418) instead of an actual _bug_ (and I suspect this is how Compositions have always behaved) but seems somewhat counter-intuitive to me when used with bi-directional patching and the `Required` policy.

### What happened?
I was attempting to use a bi-directional patch within a single composition so I could use the ARN of one base resource in the body of another base resource (`Bucket` ARN in an `IAMPolicy` document).

My belief was that I would be able to use the following XRD / Composition:

```yaml
---
metadata:
  annotations:
  name: compositeobjectstoragebuckets.resources.company.systems
spec:
  group: resources.company.systems
  names:
    kind: CompositeObjectStorageBucket
    plural: compositeobjectstoragebuckets
  claimNames:
    kind: ObjectStorageBucket
    plural: objectstoragebuckets
  defaultCompositionRef:
    name: object-storage-bucket-aws
  versions:
    - name: v1alpha1
      referenceable: true
      served: true
      schema:
        openAPIV3Schema:
          type: object
          properties:
            status:
              type: object
              properties:
                uid:
                  description: >
                    UID of the underlying bucket. On AWS for example,
                    this would contain the ARN of the generated bucket.
                  type: string

---
kind: Composition
metadata:
  name: object-storage-bucket-aws
spec:
  compositeTypeRef:
    apiVersion: resources.company.systems/v1alpha1
    kind: CompositeObjectStorageBucket

  patchSets:
    - name: BucketSyncStatus
      patches:
        # Sync the bucket ARN back to the UID field on the composite
        - type: ToCompositeFieldPath
          toFieldPath: status.uid
          # Optional because status.atProvider won't exist until the bucket
          # is reconciled.
          policy:
            fromFieldPath: Optional

    - name: S3ObjectRwPolicy
      patches:
        # Use the synced bucket UID to generate a valid policyDocument
        - fromFieldPath: status.uid
          toFieldPath: spec.forProvider.document
          # Required because without a source ARN, the policy document is invalid
          policy:
            fromFieldPath: Required
          transforms:
            - type: string
              string:
                fmt: |
                  {
                    "Version": "2012-10-17",
                    "Statement": [
                        {
                            "Action": [
                                "s3:GetObject*",
                                "s3:DeleteObject*",
                                "s3:PutObject*"
                            ],
                            "Resource": "%s/*",
                            "Effect": "Allow"
                        }
                    ]
                  }

  resources:
    - name: Bucket
      base:
        kind: Bucket
        spec:
          deletionPolicy: Delete
          forProvider:
            acl: private
            locationConstraint: us-east-1
      patches:
        - type: PatchSet
          patchSetName: BucketSyncStatus

    - name: S3RwPolicy
      base:
        apiVersion: identity.aws.crossplane.io/v1alpha1
        kind: IAMPolicy
        spec:
          deletionPolicy: Delete
      patches:
        - type: PatchSet
          patchSetName: S3ObjectRwPolicy
```

I thought what would happen here is that the patch on the `Bucket` would be a no-op because the `fromField` doesn't exist, the `Bucket` resource would be created, and then the patch on the `Policy` would be rejected (and only the Policy resource would not be created) as the `fromField` does not exist yet and is set `Required`.

What actually happens is the error from the `Required` patch on the `Policy` (`cannot render composed resource from resource template at index 1: cannot apply the patch at index 0: status: no such field`) causes the entire composition to be aborted and requeued, with no resources applied to the cluster - the Bucket is never created, so can never populate its' `status` field for the `Policy` patch to work on.

Removing the `Required` policy from the patch allows both resources to be created but until the `Bucket` is ready and its' value is patched back, the `Policy` resource is invalid (in this case, contains an empty format string placeholder from the string transform).

If this invalid resource were rejected by the provider (e.g. there was a policy check before creation on the `provider-aws` side that threw an error) then I don't think it would be possible to bypass and the resources would never be created.

### How can we reproduce it?
Create a Composition and XRD based on the above code, and create an instance of it. Watch as the Composition never completes first reconciliation, instead aborting due to the patching failure of the `IAMPolicy`.

Crossplane version: v1.2.0-rc.0.5.ge4491ffd

cc @negz @muvaf @mcavoyk

hasheddan:

Updates the composite resource reconciler to render all resources it is
able to even if it is not able to render all resources in the specified
composition. This is a change in behavior from refusing to render any
composed resources if not able to render all.

Fixes #2175 

I have:

Using the manifests in #2175:
1. Create an `ObjectStorageBucket`
```
🤖 (crossplane) k get objectstoragebucket
NAME   READY   CONNECTION-SECRET   AGE
test   False                       119s
```
2. Observed that only the `Bucket` was created
```
🤖 (crossplane) k get bucket
NAME               READY   SYNCED   AGE
test-j4jhd-7kvk8   True    True     60m
```

3. See that both references are create but `IAMPolicy` does not have a `name`

```
- apiVersion: resources.company.systems/v1alpha1
  kind: CompositeObjectStorageBucket
  metadata:
    annotations:
      crossplane.io/external-name: ""
        {"apiVersion":"resources.company.systems/v1alpha1","kind":"ObjectStorageBucket","metadata":{"annotations":{},"name":"test","namespace":"default"},"spec":{}}
    creationTimestamp: "2021-06-03T22:14:21Z"
    generateName: test-
    generation: 4
    labels:
      crossplane.io/claim-name: test
      crossplane.io/composite: test-j4jhd
   ...
   name: test-j4jhd
  spec:
    claimRef:
      apiVersion: resources.company.systems/v1alpha1
      kind: ObjectStorageBucket
      name: test
      namespace: default
    compositionRef:
      name: object-storage-bucket-aws
    resourceRefs:
    - apiVersion: s3.aws.crossplane.io/v1beta1
      kind: Bucket
      name: test-j4jhd-7kvk8
    - apiVersion: identity.aws.crossplane.io/v1alpha1
      kind: IAMPolicy
  status:
    conditions:
    - lastTransitionTime: "2021-06-03T22:16:51Z"
      reason: Creating
      status: "False"
      type: Ready
    uid: arn:aws:s3:::test-j4jhd-7kvk8
```

4. Now that `uid` is populated in `status`, we can create the `IAMPolicy`

```
🤖 (crossplane-runtime) k get iampolicy
NAME               ARN                                           READY   SYNCED   AGE
test-j4jhd-5phzf   arn:aws:iam::609897127049:policy/test-j4jhd   True    True     6m23s
```

5. Name of `IAMRole` is now present in resource refs on the composite

```
- apiVersion: resources.company.systems/v1alpha1
  kind: CompositeObjectStorageBucket
  metadata:
    annotations:
      crossplane.io/external-name: ""
        {"apiVersion":"resources.company.systems/v1alpha1","kind":"ObjectStorageBucket","metadata":{"annotations":{},"name":"test","namespace":"default"},"spec":{}}
    creationTimestamp: "2021-06-03T22:14:21Z"
    generateName: test-
    generation: 5
    labels:
      crossplane.io/claim-name: test
      crossplane.io/composite: test-j4jhd
    ...
    name: test-j4jhd
    ...
  spec:
    claimRef:
      apiVersion: resources.company.systems/v1alpha1
      kind: ObjectStorageBucket
      name: test
      namespace: default
    compositionRef:
      name: object-storage-bucket-aws
    resourceRefs:
    - apiVersion: s3.aws.crossplane.io/v1beta1
      kind: Bucket
      name: test-j4jhd-7kvk8
    - apiVersion: identity.aws.crossplane.io/v1alpha1
      kind: IAMPolicy
      name: test-j4jhd-5phzf
  status:
    conditions:
    - lastTransitionTime: "2021-06-03T23:16:37Z"
      reason: Available
      status: "True"
      type: Ready
    uid: arn:aws:s3:::test-j4jhd-7kvk8
```

6. Deleting the claim cleans up the composite and all managed resources.

### DIFF
--- a/internal/controller/apiextensions/composite/reconciler.go
+++ b/internal/controller/apiextensions/composite/reconciler.go
@@ -341,6 +341,13 @@ type Reconciler struct {
 	record event.Recorder
 }
 
+// composedRendered is a wrapper around a composed resource that tracks whether
+// it was successfully rendered or not.
+type composedRendered struct {
+	resource resource.Composed
+	rendered bool
+}
+
 // Reconcile a composite resource.
 func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) { //nolint:gocyclo
 	// NOTE(negz): Like most Reconcile methods, this one is over our cyclomatic
@@ -415,20 +422,25 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		return reconcile.Result{RequeueAfter: shortWait}, nil
 	}
 
-	// We want to ensure we can render all of our composed resources before we
-	// apply any of them. We prefer to avoid creating or updating any composed
-	// resources if we know we won't be able to create or update all of them.
+	// We optimistically render all composed resources that we are able to with
+	// the expectation that any that we fail to render will subsequently have
+	// their error corrected by manual intervention or propagation of a required
+	// input.
 	refs := make([]corev1.ObjectReference, len(tas))
-	cds := make([]resource.Composed, len(tas))
+	cds := make([]composedRendered, len(tas))
 	for i, ta := range tas {
 		cd := composed.New(composed.FromReference(ta.Reference))
+		rendered := true
 		if err := r.composed.Render(ctx, cr, cd, ta.Template); err != nil {
 			log.Debug(errRenderCD, "error", err, "index", i)
 			r.record.Event(cr, event.Warning(reasonCompose, errors.Wrapf(err, errFmtRender, i)))
-			return reconcile.Result{RequeueAfter: shortWait}, nil
+			rendered = false
 		}
 
-		cds[i] = cd
+		cds[i] = composedRendered{
+			resource: cd,
+			rendered: rendered,
+		}
 		refs[i] = *meta.ReferenceTo(cd, cd.GetObjectKind().GroupVersionKind())
 	}
 
@@ -448,7 +460,12 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 	// issues observing and processing one composed resource won't block the
 	// application of another.
 	for _, cd := range cds {
-		if err := r.client.Apply(ctx, cd, resource.MustBeControllableBy(cr.GetUID())); err != nil {
+		// If we were unable to render the composed resource we should not try
+		// and apply it.
+		if !cd.rendered {
+			continue
+		}
+		if err := r.client.Apply(ctx, cd.resource, resource.MustBeControllableBy(cr.GetUID())); err != nil {
 			log.Debug(errApply, "error", err)
 			r.record.Event(cr, event.Warning(reasonCompose, err))
 			return reconcile.Result{RequeueAfter: shortWait}, nil
@@ -460,13 +477,19 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 	for i, tpl := range comp.Spec.Resources {
 		cd := cds[i]
 
-		if err := r.composite.Render(ctx, cr, cd, tpl); err != nil {
+		// If we were unable to render the composed resource we should not try
+		// and to observe it.
+		if !cd.rendered {
+			continue
+		}
+
+		if err := r.composite.Render(ctx, cr, cd.resource, tpl); err != nil {
 			log.Debug(errRenderCR, "error", err)
 			r.record.Event(cr, event.Warning(reasonCompose, err))
 			return reconcile.Result{RequeueAfter: shortWait}, nil
 		}
 
-		c, err := r.composed.FetchConnectionDetails(ctx, cd, tpl)
+		c, err := r.composed.FetchConnectionDetails(ctx, cd.resource, tpl)
 		if err != nil {
 			log.Debug(errFetchSecret, "error", err)
 			r.record.Event(cr, event.Warning(reasonCompose, err))
@@ -477,7 +500,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 			conn[key] = val
 		}
 
-		rdy, err := r.composed.IsReady(ctx, cd, tpl)
+		rdy, err := r.composed.IsReady(ctx, cd.resource, tpl)
 		if err != nil {
 			log.Debug(errReadiness, "error", err)
 			r.record.Event(cr, event.Warning(reasonCompose, err))
--- a/internal/controller/apiextensions/composite/composed.go
+++ b/internal/controller/apiextensions/composite/composed.go
@@ -230,6 +230,10 @@ func (a *GarbageCollectingAssociator) AssociateTemplates(ctx context.Context, cr
 	}
 
 	for _, ref := range cr.GetResourceReferences() {
+		// If reference does not have a name then we haven't rendered it yet.
+		if ref.Name == "" {
+			continue
+		}
 		cd := composed.New(composed.FromReference(ref))
 		nn := types.NamespacedName{Namespace: ref.Namespace, Name: ref.Name}
 		err := a.client.Get(ctx, nn, cd)
--- a/internal/controller/apiextensions/composite/reconciler_test.go
+++ b/internal/controller/apiextensions/composite/reconciler_test.go
@@ -260,7 +260,12 @@ func TestReconcile(t *testing.T) {
 								}
 								return nil
 							}),
+							MockUpdate:       test.NewMockUpdateFn(nil),
+							MockStatusUpdate: test.NewMockStatusUpdateFn(nil),
 						},
+						Applicator: resource.ApplyFn(func(c context.Context, r client.Object, ao ...resource.ApplyOption) error {
+							return nil
+						}),
 					}),
 					WithCompositionSelector(CompositionSelectorFn(func(_ context.Context, cr resource.Composite) error {
 						cr.SetCompositionReference(&corev1.ObjectReference{})
@@ -271,6 +276,12 @@ func TestReconcile(t *testing.T) {
 					})),
 					WithRenderer(RendererFn(func(ctx context.Context, cp resource.Composite, cd resource.Composed, t v1.ComposedTemplate) error {
 						return errBoom
+					})),
+					WithConnectionDetailsFetcher(ConnectionDetailsFetcherFn(func(ctx context.Context, cd resource.Composed, t v1.ComposedTemplate) (managed.ConnectionDetails, error) {
+						return nil, nil
+					})),
+					WithReadinessChecker(ReadinessCheckerFn(func(ctx context.Context, cd resource.Composed, t v1.ComposedTemplate) (ready bool, err error) {
+						return false, nil
 					})),
 				},
 			},
--- a/internal/xcrd/crd_test.go
+++ b/internal/xcrd/crd_test.go
@@ -257,7 +257,7 @@ func TestForCompositeResource(t *testing.T) {
 													"name":       {Type: "string"},
 													"kind":       {Type: "string"},
 												},
-												Required: []string{"apiVersion", "kind", "name"},
+												Required: []string{"apiVersion", "kind"},
 											},
 										},
 									},
--- a/internal/xcrd/schemas.go
+++ b/internal/xcrd/schemas.go
@@ -106,7 +106,7 @@ func CompositeResourceSpecProps() map[string]extv1.JSONSchemaProps {
 						"name":       {Type: "string"},
 						"kind":       {Type: "string"},
 					},
-					Required: []string{"apiVersion", "kind", "name"},
+					Required: []string{"apiVersion", "kind"},
 				},
 			},
 		},


hasheddan:
`check-diff` is know bug that is fixed in #2166

hasheddan:
I removed the `BucketPolicy` from the composition and that resulted in the `cannot access object metadata` error going away, but the UID field was not being patched back to the XR or XRC correctly :thinking:

internal/controller/apiextensions/composite/reconciler.go
```diff
@@ -422,7 +422,12 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 	cds := make([]resource.Composed, len(tas))
 	for i, ta := range tas {
 		cd := composed.New(composed.FromReference(ta.Reference))
-		if err := r.composed.Render(ctx, cr, cd, ta.Template); err != nil {
+		err := r.composed.Render(ctx, cr, cd, ta.Template)
+		if v1.IsPatchRequiredFieldNotFoundErr(err) {
```
muvaf, line 426: Should we tie this to a specific error? This case is probably where this will be needed most but as a principle, I think Crossplane should create everything it can and report back the ones it couldn't to provide eventual consistency in all error cases. The main reason is that while the resources are in the same `Composition`, there is no requirement for them to be bound or related to each other. So, I think we should respect this property for all cases.

Another practical example I can think of is where you have a type in one of the templates. While we can't create that specific resource, there isn't a reason to not create the rest of the resources. If there is any hard relation between them, that's already handled by the specific resources, like referencer mechanism.

hasheddan: @muvaf this is a fair point, but I think this would actually be a divergence from the initial intention, which I believe was to block creation of all resources in the case that one was misconfigured. The missing required patch field is an edge case that we definitely should enable IMO, but we should likely have a larger discussion about our philosophy on partial creation for other cases.

muvaf: I think the decision of moving forward with other resources depends on whether it'd affect the desired state of those resources. For example, there are two possibilities:
* First and second resources can be rendered, create both of them.
* First resource cannot be rendered but move on and create the second one. Then in the next reconcile, the first resource is created.

If we see that at at the end of these two paths, the desired state of the resources are different, then we should block. For example, managed resource reference resolver mechanism is safe from that perspective because it blocks in the lower level. For this specific patch policy, it's safe because users are able to toggle whether it should block or not. What I'm trying to do is to come up with an example case in that it's unsafe to move on. And if there isn't, having the eventual consistency and deploying whatever it can would be more aligned with Kubernetes API in my opinion.

>The missing required patch field is an edge case that we definitely should enable IMO, but we should likely have a larger discussion about our philosophy on partial creation for other cases.

Agreed, we need input from @negz specifically.

hasheddan: > For example, managed resource reference resolver mechanism is safe from that perspective because it blocks in the lower level.

This wouldn't cause a render error so I'm not sure it is applicable. I think most misconfigurations (outside of this required patch case) that would cause a render error are going to require editing of the composition, and so users may prefer that creating resources is "all or nothing". Maybe this is even a setting on XR / XRC? Something like `renderAllOrFail: true/false`.

negz: I don't feel super strongly about this. I suspect this past design decision was mostly a hunch that more conservative was better here. Particularly in the case where we're updating existing infrastructure it seems safer to me to stop the world than to make what we _know_ ahead of time will be a partial update. I haven't thought through this deeply though, and I grant that leaning into eventual consistency feels more idiomatic in the Kubernetes world.

Would we need to refactor or somehow account for the default `AssociateByOrder` template association logic if we skipped resources that didn't render? At a glance I think the current draft would break that logic, because we'd potentially omit a resource from the middle of the resource refs array if it didn't render, then inject it later if it did.

muvaf: >This wouldn't cause a render error so I'm not sure it is applicable.

@hasheddan what I meant was to go through the mechanisms where not being able to deploy one composed resource would affect another one and it's one of the mechanisms that relates the composed resources to each other and it's safe, i.e. if referenced object doesn't exist initially, that wouldn't cause change in the final desired state. The other one is the reverse patching which relates them indirectly via composite resource.

> It seems like we'd potentially have empty elements in an XR's resourceRefs array?

Since we keep anonymous templates, I think so.

I guess it'd help the discussion if we list the render errors and how we expect them to be fixed:
* Schema errors in base
  * Manual edit of `Composition`
* Patch/transform errors during `Patch`
  * Manual edit of `Composition`
  * A change in the value retrieved `fromFieldPath`

For schema errors, deploying the rest and waiting for a fix for that specific resource doesn't seem like would cause any problems because resources do not refer each other at `Composition` level, all patches are either from or to the composite resource. I think one place where this could affect was the optional-by-default patching mechanism where not having the resource at the right time would cause differences, but that's solved by patch policies, i.e. if you think you'd not want to deploy resource A unless specific patch(es) are successful, then you are able to mark them as such.

In case of patch/transform errors, we already have the patch policy to toggle the behavior.

> Particularly in the case where we're updating existing infrastructure it seems safer to me to stop the world than to make what we know ahead of time will be a partial update.

@negz I'm trying to think of cases for this. I guess there is the case where you have let's say `vpcId` field in the claim that is mutable and optional and it's empty. Then user adds a `vpcId` and it goes through the composition. This could mean that some of the patches are run only now, hence might cause schema errors. So, the new `vpcId` might end up in only some of the resource it's supposed to be propagated. Depending on the effect of the field, it could cause disruption of service. So, where the patch is optional, not used initially and misconfigured in a way that'll cause schema error is a case where we might want all-or-none.

When I'm trying to think of how other tools solve this, there are different opinions. For example, `kubectl` doesn't concern itself at all if one of the YAMLs has an error; it's always `BestEffort`. Another example could be replicas; even if one of them couldn't be applied the rest still makes it to the cluster. On the other hand, `helm` stops the world if it couldn't apply one of the resources. Though `helm` has a very heavy templating machinery so they might be subject to schema errors much more frequently than us because we're mostly behind the API line and after webhook validation, the patches will be the only things not validated for schema errors.

Personally, my hunch is that eventual consistency behavior is worth more than handling that case. Especially with the dependency resolution provided by patch policy, there is a real effect of this behavior.

hasheddan: > For schema errors, deploying the rest and waiting for a fix for that specific resource doesn't seem like would cause any problems because resources do not refer each other at Composition level, all patches are either from or to the composite resource.

@muvaf I was writing a response to this along the lines of "yes, it is technically possible, but is that the expected behavior?", but mid-way through I think I changed to agreeing and thinking that we could address the case where someone wanted "all or nothing" by adding support for dry-run functionality in a CLI or other tool. That way we could reflect Kubernetes eventual consistency model with the composition machinery, but a user could also test before applying whether they may enter a case where only some of the resources are rendered successfully. What do you think?

hasheddan: @negz do you want to weigh in here as well? Sounds like current proposal is rendering all resources we can.

negz: I'm okay making this behaviour change if you both feel it's the better approach and if we have a solution to the 'gaps' it could cause in our resourceRefs array. Will the API allow us to just have an 'empty' entry in the resourceRefs array if an anonymous template fails to render?

hasheddan: Sounds good, I'll test it out and add the results here before proceeding.

muvaf:
Are we feeling OK with including this in v1.2.0?

hasheddan:
@muvaf yep! Need to do a bit more testing, but I'll get this fixed up and ready to go!

internal/controller/apiextensions/composite/reconciler.go
```diff
@@ -415,21 +415,22 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
-	// resources if we know we won't be able to create or update all of them.
+	// We optimistically render all composed resources that we are able to with
+	// the expectation that any that we fail to render will subsequently have
+	// their error corrected by manual intervention or propagation of a required
+	// input.
 	refs := make([]corev1.ObjectReference, len(tas))
 	cds := make([]resource.Composed, len(tas))
 	for i, ta := range tas {
 		cd := composed.New(composed.FromReference(ta.Reference))
+		refs[i] = *meta.ReferenceTo(cd, cd.GetObjectKind().GroupVersionKind())
```
hasheddan, line 426: This isn't going to work as we have to add the reference after render -- I am likely going to use a different var to track if there were resources we were unable to render. Update coming.

muvaf:
The `append` will mess up with the ordered association code [here](https://github.com/crossplane/crossplane/blob/a6eebfe/internal/controller/apiextensions/composite/composed.go#L175) because it is coupled with the order very tightly.

As we talked offline, we need either an empty entry or placeholder. My preference would be towards empty entry. It seems like we need to remove the requirements [here](https://github.com/crossplane/crossplane/blob/ec0b383/internal/xcrd/schemas.go#L109) and there is also a workaround [here](https://github.com/crossplane/crossplane-runtime/blob/0b27607/pkg/resource/unstructured/composite/composite.go#L124) in the code that sets the references (I don't remember why it existed precisely, but I think it might have to do with using `ObjectReference` schema, which we don't do now even though the struct in Go code is `ObjectReference`).

I think it's worth testing these changes. I can't really see a reason why it'd not be able to have an array entry with all the values being `""`. But if it can't, maybe having `apiVersion` and `kind` but leaving `name` empty could be helpful as placeholder.

A different approach could be that we could mark the entry with the name or index of the template that's used, i.e. adding a `baseName` to [ref struct](https://github.com/crossplane/crossplane/blob/ec0b383/internal/xcrd/schemas.go#L104) here. But I guess that ship has sailed in https://github.com/crossplane/crossplane/pull/2131 where we decided to use an annotation on the composed resource.

negz:
> A different approach could be that we could mark the entry with the name or index of the template that's used, i.e. adding a baseName to ref struct here. But I guess that ship has sailed in #2131 where we decided to use an annotation on the composed resource.

Just dropping some breadcrumbs to this specific comment - https://github.com/crossplane/crossplane/pull/2131#discussion_r574375852, which touches on why I didn't take that option.

hasheddan:
@muvaf @negz I have updated here to only reference composed resources by GVK when we are unable to render them (see updated testing results in PR body) and everything appears to be behaving correctly. I think this is ready for another review, but I am still testing out a few scenarios -- I am specifically concerned with situations where we fail to render a resource after we have already rendered it successfully in the past.

internal/controller/apiextensions/composite/reconciler.go
```diff
@@ -341,6 +341,13 @@ type Reconciler struct {
 	record event.Recorder
 }
 
+// composedRendered is a wrapper around a composed resource that tracks whether
+// it was successfully rendered or not.
+type composedRendered struct {
+	composed resource.Composed
```
negz, line 347:
```suggestion
	resource resource.Composed
```

Nit: I read `cd` as an abbreviation for "composed", so `cd.resource` makes more sense to me than `cd.composed`.

internal/controller/apiextensions/composite/reconciler.go
```diff
@@ -527,7 +552,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 	// * Report which resources are not ready.
 	// * If a resource becomes Unavailable at some point, should we still report
 	//   it as Creating?
-	if ready != len(refs) {
+	if ready != len(refs) || renderFail {
```
negz, line 555: Won't `ready != len(refs)` always be true if `renderFail` is true? Or are there cases where we need both?

negz (commented):
One question and a nit, but this broadly LGTM.

internal/controller/apiextensions/composite/reconciler.go
hasheddan: You are correct :+1: `renderFail` was leftover from a previous iteration, removing now

muvaf (approved):
LGTM! Thanks @hasheddan !

PR #2183, opened 2021-03-03
Make concepts overview match content

hasheddan:

The OAM section has been moved to guides so we must clean up the
concepts section with references to its prior location.

I have:

n/a

### DIFF
--- a/docs/concepts/overview.md
+++ b/docs/concepts/overview.md
@@ -88,31 +88,6 @@ operator facing, but may optionally offer an application developer facing
 composite resource claim that acts as a proxy for a composite resource. You can
 learn more about all of these concepts in the [composition documentation].
 
-## OAM
-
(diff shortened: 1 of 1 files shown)


docs/concepts/overview.md
```diff
@@ -88,31 +88,6 @@ operator facing, but may optionally offer an application developer facing
 composite resource claim that acts as a proxy for a composite resource. You can
```
muvaf, line 88: There are mentions of OAM and its CRDs in the introduction section. Would you like to remove them as well?

hasheddan: @muvaf no I don't think so because those notes are alluding the the fact that helm will install OAM CRDs even if the subchart is not enabled

jbw976 (approved):

PR #2196, opened 2021-03-08
Move community day meetings to Thursday mornings

negz:

I'd like to propose we move our community day meetings from Monday to Thursday mornings. I'd also like to offset the week by one - so for example we'd meet on Thursday 11th, Thursday 25th, etc rather than Monday 15th, Monday 29th, etc. This proposed time slot would be more convenient for those of us on the steering committee and maintainership team that work at Upbound.

I have:

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,7 +122,7 @@ Kubernetes API.
 * Join our regular community meetings.
 * Provide feedback on our [roadmap](ROADMAP.md).
 
-The Crossplane community meeting takes place every other [Monday at 10:00am
+The Crossplane community meeting takes place every other [Thursday at 10:00am
 Pacific Time]. Anyone who wants to discuss the direction of the project, design
 and implementation reviews, or raise general questions with the broader
 community is encouraged to join.


README.md
```diff
@@ -122,7 +122,7 @@ Kubernetes API.
 * Join our regular community meetings.
 * Provide feedback on our [roadmap](ROADMAP.md).
 
-The Crossplane community meeting takes place every other [Monday at 10:00am
+The Crossplane community meeting takes place every other [Thursday at 10:00am
 Pacific Time]. Anyone who wants to discuss the direction of the project, design
 and implementation reviews, or raise general questions with the broader
 community is encouraged to join.
```
negz, line 128: Note to self: I'll need to update the meeting link below if this change is accepted.

hasheddan (approved):
+1 from me

muvaf (approved):
+1

krishchow:
10:00am on Thursdays conflicts on my end with a SIG Storage meeting. Would it make more sense for the meeting to be at 9:00am on Thursdays since it is already alternating with TBS which runs at the same time?

negz:
> Would it make more sense for the meeting to be at 9:00am on Thursdays

I wonder if an earlier 9:00am PST start would make it less likely for folks in PST to attend? I know that's quite early for some folks (including myself). Perhaps this isn't too much of an issue if the majority of regular attendees aren't in PST?

> since it is already alternating with TBS which runs at the same time?

@hasheddan Do I recall that mentioned you might move TBS to a different time slot (or already had)?

hasheddan:
@negz yep that is correct :+1: It can also be moved again if needed.

README.md
jbw976: I can update the zoom meeting itself so that the link below doesn't need to change at all

i can also update the calendar appointment

jbw976 (approved):

negz:
Thanks folks! Let's discuss this at the Monday 15th meeting so that everyone is aware of the pending change.

jbw976:
Added this topic to the community meeting agenda for Monday!

pdettori:
+1 for me

negz:
We just discussed this on the community call. The next Crossplane community meeting will be Thursday 25th.

jbw976:
The calendar appointment and zoom meeting time have been updated for this new Thurs time slot :)

PR #2198, opened 2021-03-09
Update README and Overview with current project features and mission

hasheddan:

Updates the Crossplane `README.md` (as well as the `Overview` section in the docs) to reflect current project features and mission.

I have:

n/a

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,46 +2,46 @@
 
 ![Crossplane](docs/media/banner.png)
 
-Crossplane, a [Cloud Native Computing Foundation][CNCF] sandbox project, is an
-open source Kubernetes add-on that extends any cluster with the ability to
-provision and manage cloud infrastructure, services, and applications using
-kubectl, GitOps, or any tool that works with the Kubernetes API.
+Crossplane is an open source Kubernetes add-on that enables platform teams to
+assemble infrastructure from multiple vendors, and expose higher level
+self-service APIs for application teams to consume. Crossplane effectively
+enables platform teams to quickly put together their own opinionated platform
+declaratively without having to write any code, and offer it to their
+application teams as a self-service Kubernetes-style declarative API.
 
-With Crossplane you can:
+Both the higher level abstractions as well as the granular resources they are
+composed of are represented simply as objects in the Kubernetes API, meaning
+they can all be provisioned and managed by kubectl, GitOps, or any tools that
+can talk with the Kubernetes API. To facilitate reuse and sharing of these APIs,
+Crossplane supports packaging them in a standard OCI image and distributing via
+any compliant registry.
 
-* **Provision & manage cloud infrastructure with kubectl**
-  * [Install Crossplane][documentation] to provision and manage cloud infrastructure and
-    services from any Kubernetes cluster.
-  * Provision infrastructure primitives from any provider ([GCP], [AWS],
-    [Azure], [Alibaba], on-prem) and use them alongside existing application
-    configurations.
-  * Version, manage, and deploy with your favorite tools and workflows that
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -4,44 +4,41 @@
...
+
+* **Expose Custom APIs**
+  * Crossplane standardizes on the Kubernetes RBAC permission model, meaning
+    that custom APIs can be exposed to developers on a per-object basis using
+    familiar Kubernetes APIs, such as `Role` and `ClusterRole`. Combining this
+    permission model with Crossplane's composition functionality provides the
+    ability to define access for developers at the same level of abstraction
+    they are exposed to.
+
+* **Extend the Kubernetes API with New Providers**
```
jbw976, line 36: another good addition to this list is the XRM in general and its role in providing a universal/consistent API across multiple vendors and providers.  here's some prose from the draft CNCF proposal on that topic that we should consider including here:

> While bringing this diverse set of resources into the Kubernetes based control plane, Crossplane recognized the need for a single consistent API across all of their vendors and providers. To this end, we have created the Crossplane Resource Model (XRM).  XRM extends the Kubernetes Resource Model (KRM) in an opinionated way, resulting in a universal experience for managing resources external to the control plane itself. While interacting with the XRM, things like credentials, workload identity, connection secrets, status conditions, deletion policies, and references to other resources work the same no matter which provider they are backed by.

jbw976 (commented):
Thanks for this effort @hasheddan! In general, I'd like us to find alignment with the CNCF proposal messaging too, so I've added a couple comments towards that.

The CNCF proposal is essentially 2 main feature areas (managing infra, and composing your own APIs) plus the XRM that provides the universal API experience.

I think it makes sense to go into more details in this README (it's a different audience), like you have done with packaging and exposing, so I would keep those, but we should get a bit more synced on the other higher level main value messaging too.

Feel free to sync with me offline to workshop more together :)

docs/README.md
```diff
@@ -2,40 +2,43 @@
...
 
-* **Provision & manage cloud infrastructure with kubectl**
-  * [Install Crossplane] to provision and manage cloud infrastructure and
-    services from any Kubernetes cluster.
-  * Provision infrastructure primitives from any provider ([GCP], [AWS],
-    [Azure], [Alibaba], on-prem) and use them alongside existing application
-    configurations.
-  * Version, manage, and deploy with your favorite tools and workflows that
-    you’re using with your clusters today.
+* **Compose Custom APIs**
```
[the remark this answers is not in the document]
hasheddan: @jbw976 I agree that this needs to be worded better, and maybe switching provisioning infrastructure to be first would be correct. However, the reasoning for putting composition first is because it matches how we have been moving towards introducing the project in the documentation. As you can see with `v1.1` docs, the [Provisioning Infrastructure](https://crossplane.io/docs/v1.1/getting-started/provision-infrastructure.html) section uses composition, rather than managed resources directly, which is a reflection of the fact that we think most Crossplane users should never create managed resources except through XRs / XRCs. That being said, we are still provisioning infrastructure, just at a higher abstraction level - I wonder if there is a way to lead in the `README.md` with provisioning, but at the higher level as well?

README.md
hasheddan: This is great! It would be awesome if we could link to a longer document on XRM from here also 👍

jbw976 (approved):
Thank you for making these updates @hasheddan!

jbw976:
@hasheddan can we backport this to 1.1 also? That way, when people go to the docs, where the default is 1.1, they will also see this messaging.

hasheddan:
@jbw976 you read my mind 🙂 coming right up!

Issue #2201, opened 2021-03-09
convert transform could not resolve: input type int64 is not supported


ytsarev:
### What happened?

`transforms` fails with
```
  Warning  ComposeResources   22s (x4 over 22s)  defined/compositeresourcedefinition.apiextensions.crossplane.io  cannot render composed resource from resource template at index 0: cannot apply the patch at index 1: transform at index 0 returned error: convert transform could not resolve: input type int64 is not supported
```
in an attempt to convert an integer to string

### How can we reproduce it?

In a CompositeResourceDefinition create integer parameter like
```
                  CostCenter:
                    type: integer
```
in Composition define the transformation:
```
        - fromFieldPath: "spec.parameters.CostCenter"
          toFieldPath: "spec.forProvider.tagging.tagSet[1].value"
          transforms:
            - type: convert
              convert:
                toType: string
```
Apply the claim which contains the integer parameter specified
```
    CostCenter: 12345
```
Observe
```
  Warning  ComposeResources   116s (x12 over 5m56s)  defined/compositeresourcedefinition.apiextensions.crossplane.io  cannot render composed resource from resource template at index 0: cannot apply the patch at index 1: transform at index 0 returned error: convert transform could not resolve: input type int64 is not supported
```
in Composite resource event log

### What environment did it happen in?
Crossplane version: v1.1

PR #2203, opened 2021-03-10
composition: add support for int64 to convert transform

muvaf:

Adds support for `int64` type to convert transform because it's not uncommon to get int64 from CRs.

Fixes https://github.com/crossplane/crossplane/issues/2201

I have:

Not tested yet. @ytsarev would you be able to run this PR locally and test with your setup?


ytsarev:
@muvaf sure thing, will be happy to test. What is the best way to make a controller image with the test build?

muvaf:
@ytsarev I think the easiest way would be to set the image in `Deployment` of Crossplane to `muvaf/crossplane:2203` that I just built and pushed. Or you can set the replica of `Deployment` to 0 and run `go run cmd/crossplane/main.go core start` if you have this PR in your local.

apis/apiextensions/v1/composition_types.go
```diff
@@ -510,6 +510,7 @@ const (
 	ConvertTransformTypeString  = "string"
 	ConvertTransformTypeBool    = "bool"
 	ConvertTransformTypeInt     = "int"
+	ConvertTransformTypeInt64   = "int64"
```
hasheddan, line 513: Copying over some info from Slack:
```
@hasheddan  does fieldpath GetValue of an integer always return an int64?
@negz The fact that that becomes an int64 is an implementation detail of the test, because we use json.Unmarshal to unmarshal those bytes into a map[string]interface{} and json.Unmarshal decodes numbers as int64. Hrm scratch that - it returns float64. It’s actually "k8s.io/apimachinery/pkg/util/json", which iirc has special sauce to use int64 instead of float64.
@hasheddan I’ve kind of been asking leading questions here because I noticed through chatting with @Yury Tsarev  that we only support conversion of int types and I’m thinking that might actually never work because we are getting int64 out each time when we get value using fieldpath https://github.com/crossplane/crossplane/blob/54217a20c9595548f5406d7433772a51a2205f2c/apis/apiextensions/v1/composition_types.go#L512
```
What I believe is happening here is that we are always using unstructured objects when performing these patches, and so under the hood the following is happening:
1. We [start the composite controller with an unstructured type](https://github.com/crossplane/crossplane/blob/54217a20c9595548f5406d7433772a51a2205f2c/internal/controller/apiextensions/definition/reconciler.go#L389).
2. On reconcile we `GET` into the unstructured object, which unmarshals the json returned by the API server [using the `UnstructuredJSONScheme`](https://github.com/kubernetes/apimachinery/blob/1f1bc58a1c7983ab067dac7b23ff9fdc5081d5ba/pkg/apis/meta/v1/unstructured/unstructured.go#L125).
3. The `UnstructuredJSONScheme` decoder is [using the an underlying JSON serializer](https://github.com/kubernetes/apimachinery/blob/1f1bc58a1c7983ab067dac7b23ff9fdc5081d5ba/pkg/apis/meta/v1/unstructured/helpers.go#L388) from apimachinery (this is what @negz was alluding to).
4. The apimachinery serializer is [converting numbers to either `int64` or `float64`](https://github.com/kubernetes/apimachinery/blob/1f1bc58a1c7983ab067dac7b23ff9fdc5081d5ba/pkg/util/json/json.go#L79). It will always try to [use `int64` first](https://github.com/kubernetes/apimachinery/blob/1f1bc58a1c7983ab067dac7b23ff9fdc5081d5ba/pkg/util/json/json.go#L152), then [attempt `float64`](https://github.com/kubernetes/apimachinery/blob/1f1bc58a1c7983ab067dac7b23ff9fdc5081d5ba/pkg/util/json/json.go#L156).

All of this seems to be supported by conversations I had with @ytsarev about [changing the integer `format`](https://swagger.io/docs/specification/data-models/data-types/) of fields in an XRD schema and still seeing that we error on `int64` not supported. So I am wondering if we should just drop the `int` conversion support (as it is never possible), or alias it to work with `int64` as well for a more friendly UX. What are your thoughts?

muvaf: That's a great summary! I think we can replace `int` with `int64` since we are able to get enum validation, `int64` is just as fine as `int` from UX perspective IMO.

hasheddan: @muvaf I guess this would be an API change, but it would only break a case that was already not working if I understand correctly.

muvaf: Yes, unless they used one of core CRDs as composed resource which is not a supported scenario anyway.

muvaf: Hmm, yeah that's true. I was still thinking about schema knowing those types but it seems unstructured is not concerned with schema anywhere.

muvaf:
@ytsarev were you able to test?

apis/apiextensions/v1/composition_types.go
negz: Would it be easy to alias the `int` type to `int64`? i.e. Introduce support for `int64` but keep `int` (which would use `int64` under the hood). I get the rationale that `int` was broken anyway, but:

* For folks who are thinking about the YAML/JSON APIs rather than Go code, I doubt they care whether it's an int or int64 (there's no difference in JSON), so `int` seems like nice shorthand to keep.
* It would be nice if anyone who had been trying to use an `int` conversion had it now 'just work'.

ytsarev:
@muvaf just tested it - conversion did work as expected. Thanks a lot!

apis/apiextensions/v1/composition_types.go
muvaf: Yeah, should be doable.

negz (approved):

negz (changes requested):
Whoops - I hit approve but then noticed a unit test is failing. Looks good to me once that's fixed.

negz (approved):

PR #2227, opened 2021-03-21
cli: make the message after install to do a sudo move more obvious

jbw976:

This PR adds a splash of color to the success output message in our `install.sh` script that installs the Crossplane CLI.  The command printed there to move the downloaded binary to where kubectl lives is really easy to miss when you quickly glance at the output.  This change would make the need to run this command more obvious.

I have:

I have run the script directly with `./install.sh` locally on my Mac.

This [stackoverflow article](https://stackoverflow.com/questions/5947742/how-to-change-the-output-color-of-echo-in-linux) indicates that the `tput` approach is portable, but having a friendly reviewer of this PR that is using Linux test this change also would be appreciated :)

#### Before:

Screen Shot 2021-03-20 at 6 41 53 PM

#### After:

Screen Shot 2021-03-20 at 6 42 45 PM

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -59,10 +59,14 @@ fi
 
 chmod +x kubectl-crossplane
 
-echo "kubectl plugin downloaded successfully! Run the following commands to finish installing it:"
-echo 
-echo sudo mv kubectl-crossplane $(dirname $(which kubectl))
-echo kubectl crossplane --help
+YELLOW_FOREGROUND=`tput setaf 3`
+RESET=`tput sgr0`
+
+echo "kubectl plugin downloaded successfully!"
+echo
+echo "${YELLOW_FOREGROUND}Run the following commands to finish installing it:"
+echo "sudo mv kubectl-crossplane $(dirname $(which kubectl))"
+echo "kubectl crossplane --help${RESET}"
 echo
 echo "Visit https://crossplane.io to get started. 🚀"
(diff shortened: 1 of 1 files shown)


hasheddan (commented):
Thanks @jbw976, this is a great update! I'll test it out on my Ubuntu machine 👍

hasheddan (approved):
LGTM!
cli-colors

jbw976:
Thanks for testing this out too @hasheddan! 🙇‍♂️

install.sh
```diff
@@ -59,10 +59,14 @@ fi
...
-echo sudo mv kubectl-crossplane $(dirname $(which kubectl))
-echo kubectl crossplane --help
+YELLOW_FOREGROUND=`tput setaf 3`
+RESET=`tput sgr0`
+
+echo "kubectl plugin downloaded successfully!"
+echo
+echo "${YELLOW_FOREGROUND}Run the following commands to finish installing it:"
+echo "sudo mv kubectl-crossplane $(dirname $(which kubectl))"
+echo "kubectl crossplane --help${RESET}"
```
khos2ow, line 69: I believe to workaround the error `printf` can be used:

```
YELLOW_FOREGROUND="\033[33m"
RESET="\033[0m"

printf "%sRun the following commands to finish installing it:%s", YELLOW_FOREGROUND, RESET
...
```

PR #2228, opened 2021-03-22
Revert "Merge pull request #2227 from jbw976/cli-install-colors"

jbw976:
This reverts commit 98237f718f699e5cdd028b934fe5e170259766d0, reversing changes made to b07187078e0fff03b7f620a7aa0c77a3c44fce17 as part of #2227.

This is being reverted because the github actions env is not compatible by default with the way that #2227 was calling `tput` (`$TERM` is not set in that environment).  Rather than floundering around trying to get a fix that works in all environments, we're simply reverting the change for now.

I have:

Manually ran `./install.sh` locally to ensure behavior is back to before #2227 :(

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -59,14 +59,10 @@ fi
 
 chmod +x kubectl-crossplane
 
-YELLOW_FOREGROUND=`tput setaf 3`
-RESET=`tput sgr0`
-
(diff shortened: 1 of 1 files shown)


hasheddan (approved):

jbw976:
Ah, using `git revert` didn't sign-off the commit

PR #2230, opened 2021-03-23
Fix "Getting Started with AWS (with VPC)" configuration

negz:

The "AWS with VPC" guide appears to be broken as of Crossplane 1.1.

I have:

I haven't built and installed this configuration, but I have installed provider-aws v0.16, applied the XRD and Composition from this configuration, and validated that I can create a `CompositePostgreSQLInstance`.


hasheddan (approved):
Thanks @negz! Probably should keep tabs on each provider in case any of the resources used in the getting started guide are changed 👍

Issue #1033, opened 2019-11-06
Implement integration testing across Crossplane ecosystem


hasheddan:
### What problem are you facing?

Currently, all integration testing across the Crossplane ecosystem is performed in a manual ad-hoc manner that results in a significant amount of inefficient effort and uncaught errors leading up to new releases. It is desirable to run automated integration tests on a more frequent basis such that errors are caught faster and releases are delivered more smoothly.

### How could Crossplane help solve your problem?

Developing a framework that can be used across Crossplane projects would standardize the implementation and execution of integration tests. The initial implementation should be broad and flexible, deferring much of the logic to the written tests. Over time, commonly used tests and functionality can be incorporated into the framework itself.

### Steps

- [x] Create design doc for integration testing framework #1032 
- [x] Create integration testing framework (work can be tracked [here](https://github.com/hasheddan/athodyd))
- [ ] Implement initial integration tests in stack-aws https://github.com/crossplaneio/stack-aws/issues/73
- [ ] Implement initial integration tests in stack-azure https://github.com/crossplaneio/stack-azure/issues/61
- [x] Implement initial integration tests in stack-gcp https://github.com/crossplaneio/stack-gcp/issues/87
- [ ] Implement initial integration tests in stack-rook

suskin:
Summarizing some discussion on this from Slack with @hasheddan:

As a developer, I'd like to run fast, narrowly scoped "integration" tests which test my controller logic (for example). I'd like to be able to run them as part of my PR test suite, for ones which are fast enough and which don't use external resources or infrastructure.

This would likely be accomplished by allowing us to easily use the local mode of the `envtest` library, which depends on local apiserver binaries being available. Solving it in this way would mean updating the build images to include those binaries.

This came up in the context of a test in #1214

negz:
@hasheddan, @rahulgrover99  do you know if this issue should still be "In Progress" in the v1.4 timeframe? Is it being actively worked on?

hasheddan:
@negz is it not to my knowledge. I feel like it would it make sense to close this one out as much of the context is stale, and more specific issues related to this effort would serve as better items for folks to work on.

PR #2231, opened 2021-03-23
Expose typed clients for external usage.

rahulgrover99:
Allows typed clients package to be used for external usage. This would also allow writing tests using Crossplane Resources.
The files from `internal/client` have been moved to `pkg/client` and the paths for the same have also been fixed in the required files. 

This will also be useful for me to write [Provider Upgrade tests](https://github.com/crossplane/test/issues/7) and [Integration Tests](https://github.com/crossplane/crossplane/issues/1033). 

Signed-off-by: Rahul Grover <rahulgrover99@gmail.com>

I have:

N/A


negz:
I feel like I've asked @hasheddan this before and forgotten the answer, but is there a benefit to using these generated typed clients compared to the `controller-runtime` `client.Client` that we typically use in our controller code?

apis/generate.go
```diff
@@ -30,9 +30,9 @@ limitations under the License.
 //go:generate go run -tags generate sigs.k8s.io/controller-tools/cmd/controller-gen object:headerFile=../hack/boilerplate.go.txt paths=./pkg/meta/... crd:crdVersions=v1 output:artifacts:config=../docs/api-docs/crds
 
 // Generate clientset for types.
-//go:generate rm -rf ../internal/client
-//go:generate go run -tags generate k8s.io/code-generator/cmd/client-gen --clientset-name "versioned" --build-tag="ignore_autogenerated" --go-header-file "../hack/boilerplate.go.txt" --output-package "github.com/crossplane/crossplane/internal/client/clientset" --input-base "github.com/crossplane/crossplane/apis" --output-base "../tmp-clientgen" --input "pkg/v1alpha1,pkg/v1beta1,pkg/v1,apiextensions/v1"
-//go:generate cp -r ../tmp-clientgen/github.com/crossplane/crossplane/internal/client ../internal/client
+//go:generate rm -rf ../pkg/client
```
negz, line 33: If we do expose these clients, does it make sense to drop the `pkg` directory and have their import path be `github.com/crossplane/crossplane/clients`? The `pkg` directory has questionable value - see https://github.com/golang-standards/project-layout/issues/10 for some debate around it.

hasheddan:
@negz not specifically that I know of, though I don't see a huge issue in making them consumable now that APIs are v1. @rahulgrover99 apologies for the delay on getting test repo set up! I'll push an example of how to use the dynamic client.

negz:
> I don't see a huge issue in making them consumable now that APIs are v1

I'm always wary of increasing the API surface exposed by a high profile Go module like this one. I'm not saying we shouldn't do it, but I do think we should have a coherent reason to do so.

hasheddan:
@rahulgrover99 do you think we can close this out since the dynamic client is serving this purpose for the tests you are working on?

rahulgrover99:
> @rahulgrover99 do you think we can close this out since the dynamic client is serving this purpose for the tests you are working on?

Yes, I think we can close this. :)

PR #2239, opened 2021-03-26
adding extra env vars logic in helm chart to be injected in deployment yml

### LINKED ISSUE #2238: Allow additional env vars to be set via helm for the crossplane deployment
currently there is no ability to set and manage crossplane deployment env vars via the helm chart which causes the additional overhead of editing things after install.

BigSammich:
…s to be set in deployment.yml

Signed-off-by: Kyle Zimmermann <kyle.s.zimmermann@gmail.com>

This change exposes an extraEnvironmentVars field in the crossplane helm chart which can be used to
set/manage additional env vars that might be needed for a crossplane deployment.

Fixes #2238

I have:

Locally executed a `helm template` command and validated that deployment.yml was created
properly with the additional env vars set.

### DIFF
--- a/cluster/charts/crossplane/templates/deployment.yaml
+++ b/cluster/charts/crossplane/templates/deployment.yaml
@@ -76,6 +76,10 @@ spec:
                 fieldPath: metadata.namespace
           - name: LEADER_ELECTION
             value: "{{ .Values.leaderElection }}"
+        {{- range $key, $value := .Values.extraEnvironmentVars }}
+          - name: {{ $key | replace "." "_" }}
+            value: {{ $value | quote }}
+        {{- end}}
         volumeMounts:
           - mountPath: /cache
             name: package-cache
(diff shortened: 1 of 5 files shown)


BigSammich:
I'm currently limited on running make execution due to developing on an m1 mac

hasheddan (commented):
Thanks for adding this @BigSammich! Would you mind adding these values to https://github.com/crossplane/crossplane/blob/master/cluster/charts/crossplane/README.md and run `make generate` to update the docs?

cluster/charts/crossplane/templates/deployment.yaml
```diff
@@ -76,6 +76,10 @@ spec:
                 fieldPath: metadata.namespace
           - name: LEADER_ELECTION
             value: "{{ .Values.leaderElection }}"
+        {{- range $key, $value := .Values.extraEnvironmentVars }}
```
hasheddan, line 79: @BigSammich I wonder if we should also allow for env vars to be set on RBAC manager for consistency (though I know it isn't relevant for your use case). I think we should at least call these `extraEnvironmentVarsCrossplane` for consistency with other inputs, and consider adding `extraEnvironmentVarsRBACManager`. (Feel free to shorten to `extraEnvVars____` if you like also)

BigSammich: Agree with the consistency. Changes made

hasheddan (approved):
Thanks @BigSammich!

PR #2259, opened 2021-04-09
Update go-containerregistry to v0.4.1

### LINKED ISSUE #2202: Update to newer version of go-containerregistry

We are currently on an older version of `go-containerreigstry` (`v0.2.1` while latest is `v0.4.1`). There are a number of fixes we should take advantage of, notably https://github.com/google/go-containerregistry/pull/824, which allows us to better handle registry defaulting. `k8schain` has also been moved to its own go module.

Update to `go-containerregistry` `v0.4.1`. It would be good to do this early in the `v1.2` release cycle to have longer bake time.

hasheddan:

Updates `go-containerregistry` to latest `v0.4.1` release. Most notable change is the breaking of `k8schain` into a separately versioned module. 

Fixes #2202

I have:

Ran through integration tests and getting started guide. Also confirmed that `kubecl crossplane update provider/configuration` did not append default registry if not provided.


go.mod
```diff
@@ -3,14 +3,17 @@ module github.com/crossplane/crossplane
 go 1.13
 
 require (
+	github.com/Azure/go-autorest/autorest/validation v0.2.0 // indirect
```
negz, line 6: Are these Azure and AWS changes expected? I'm guessing they relate to k8schain?

negz (approved):

go.mod
hasheddan: @negz yep! We actually don't use them at build time (we exclude the functionality with build tags), so it is kind of unfortunate that they show up here.

PR #2269, opened 2021-04-15
rbac manager: make it optional to deploy the aggregated clusterroles

muvaf:

I'd like to make my Crossplane security footprint as low as possible but these aggregated `ClusterRole`s are deployed no matter what even though I don't really use the `ClusterRole`s they aggregate to. We can make it optional to deploy them. I chose not to make it optional to deploy the actual `ClusterRoles` these ones aggregate to so that users can just replace the aggregated one with their own and keep using the actual ones.

The Helm value is `false` by default so there will be no change to users unless they explicitly specify that they want not to deploy them via the added Helm value.

Fixes #

I have:

```
$ kubectl create ns crossplane-system
namespace/crossplane-system created
$ helm install crossplane --namespace crossplane-system cluster/charts/crossplane --set rbacManager.skipAggregatedClusterRoles=true
```

Confirmed that the list of `ClusterRole`s does not include the ones I skipped here.

### DIFF
--- a/cluster/charts/crossplane/values.yaml.tmpl
+++ b/cluster/charts/crossplane/values.yaml.tmpl
@@ -25,6 +25,7 @@ imagePullSecrets:
 
 rbacManager:
   deploy: true
+  skipAggregatedClusterRoles: false
   replicas: 1
   managementPolicy: All
   leaderElection: true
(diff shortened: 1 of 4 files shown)


hasheddan (approved):
Though we are close to the v1.2 release, I think this change is reasonable as it only introduces optional functionality that is off by default 👍

cluster/charts/crossplane/values.yaml.tmpl
```diff
@@ -25,6 +25,7 @@ imagePullSecrets:
 
 rbacManager:
   deploy: true
+  skipAggregatedClusterRoles: false
```
muvaf, line 28: Anyone's got a better name for this field?

PR #2270, opened 2021-04-19
Lint with same go version as build

hasheddan:

Updates lint workflow to setup go version beforehand so that it uses the
same as all build and publish workflows.

I have:

n/a

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,6 +47,11 @@ jobs:
         with:
           submodules: true
 
+      - name: Setup Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: ${{ env.GO_VERSION }}
+
       - name: Find the Go Build Cache
         id: go
         run: echo "::set-output name=cache::$(go env GOCACHE)"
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -68,10 +73,9 @@ jobs:
       - name: Vendor Dependencies
         run: make vendor vendor.check
 
-      # This action uses its own setup-go, which always seems to use the latest
-      # stable version of Go. We could run 'make lint' to ensure our desired Go
-      # version, but we prefer this action because it leaves 'annotations' (i.e.
-      # it comments on PRs to point out linter violations).
```
negz, line 74: Is the part of this comment that was removed no longer true, or was I mistaken when I wrote it?

hasheddan: @negz it is still true, but we are now disabling it with `skip-go-installation: true` (which I just added).

negz: Ah, got it. It seems like that option didn't exist back when I wrote this comment (it was added inhttps://github.com/golangci/golangci-lint-action/pull/144).

negz (approved):

PR #2271, opened 2021-04-21
helm: change image pull policy to IfNotPresent to make use of kubelet…

muvaf:

Change image pull policy to IfNotPresent to make use of kubelet cache. There isn't really much of a reason for pulling every time.

I have:

N/A.


hasheddan (changes requested):
@muvaf docs need to be updated also if we make this change

hasheddan (approved):
Nice! Thanks for updating the description as well :)

PR #2274, opened 2021-04-26
Add Crossplane multi-tenant guide to documentation

### LINKED ISSUE #2116: docs: User guide for enabling multi-tenant support
This question has come up a couple of times, so it would be great to write a user guide or doc about how to enable it.

Scenario: A platform team is running a single crossplane cluster.  They want to enable multi-tenancy where each application team (or customer) that they service gets their own namespace and isolation from other teams/customers. An important part of this is that each namespace gets its own `ProviderConfig` so they are each using credentials/permissions in the cloud provider that are specific for them.  Team A should NOT be able to use the credentials (`ProviderConfig`) for Team B.

@hasheddan had some great suggestions on this:
1. Hardcode the providerConfig in a composition. This is not optimal in some way because it means that anywhere that composition is used will use those same credentials. However,  if you are thinking in an "object-based" auth model rather than a "user-based" auth model, it actually makes sense. Least privilege looks like making the provider use credentials scoped to the operations it needs for that resource or group of resources. Then you control who can create those resources with k8s RBAC. IMO this is a good model because you are not crossing the responsibilities of k8s user control and cloud provider user control. K8s is how you manage users RBAC, the cloud provider is how you handle provider RBAC. This is the area I really feel like we need to educate more folks on.
1. You can expose a providerConfig field on the XRD / XRC then write OPA (or other policy) around what values can be supplied in what namespace (i.e. the claim is created at the namespace scope, so you can write a policy that says for `MyCoolClusterClaim` in namespace `team-1` the only values acceptable for `spec.providerConfig.name` are `team-1-pc`)
1. We have also talked about adding support for "patching from data sources" (@negz is working on this area: https://github.com/crossplane/crossplane/issues/2099). This would mean you could create a composition that had a path that did something like interpolating the namespace into the providerConfig name. @muvaf describes how this is already possible with current mechanisms here: https://github.com/crossplane/crossplane/issues/2099#issuecomment-768301167

@muvaf suggestion in https://github.com/crossplane/crossplane/issues/2099#issuecomment-768301167 is pretty slick and works with current v1.0 functionality and without any additional tooling.

hasheddan:

Adds a guide for various strategies to deploy Crossplane in multi-tenant
environments, both within a single cluster and across multiple clusters.

Fixes #2116 

I have:

n/a

### DIFF
--- a/docs/guides/multi-tenant.md
+++ b/docs/guides/multi-tenant.md
@@ -1,0 +1,299 @@
+---
+title: Multi-Tenant Crossplane
+toc: true
+weight: 240
+indent: true
+---
+
+This guide describes how to use Crossplane effectively in multi-tenant
+environments by utilizing Kubernetes primitives and compatible policy
+enforcement projects in the cloud-native ecosystem.
+
+- [Background](#background)
+    - [Cluster Scoped Managed Resources](#cluster-scoped-managed-resources)
+    - [Namespace Scoped Claims](#namespace-scoped-claims)
+- [Single Cluster Multi Tenancy](#single-cluster-multi-tenancy)
+  - [Composition as an Isolation Mechanism](#composition-as-an-isolation-mechanism)
+  - [Namespaces as an Isolation Mechanism](#namespaces-as-an-isolation-mechanism)
+  - [Policy Enforcement with Open Policy Agent](#policy-enforcement-with-open-policy-agent)
+- [Multi Cluster Multi Tenancy](#multi-cluster-multi-tenancy)
+  - [Reproducible Platforms with Configuration Packages](#reproducible-platforms-with-configuration-packages)
+  - [Control Plane of Control Planes](#control-plane-of-control-planes)
+
+## Background
+
+Crossplane is designed to run in multi-tenant environments where many teams are
+consuming the services and abstractions provided by infrastructure operators in
+the cluster. This functionality is facilitated by two major design patterns in
+the Crossplane ecosystem. 
+
+### Cluster-Scoped Managed Resources
+
+Typically, Crossplane providers, which supply granular [managed resources] that
+reflect an external API, authenticate by using a `ProviderConfig` object that
+points to a credentials source (such as a Kubernetes `Secret`, the `Pod`
+filesystem, or an environment variable). Then, every managed resource references
+a `ProviderConfig` that points to credentials with sufficient permissions to
+manage that resource type.
+
+For example, the following `ProviderConfig` for `provider-aws` points to a
+Kubernetes `Secret` with AWS credentials.
+
+```yaml
+apiVersion: aws.crossplane.io/v1beta1
+kind: ProviderConfig
+metadata:
+  name: cool-aws-creds
+spec:
+  credentials:
+    source: Secret
+    secretRef:
+      namespace: crossplane-system
+      name: aws-creds
+      key: key
+```
+
+If a user desired for these credentials to be used to provision an
+`RDSInstance`, they would reference the `ProviderConfig` in the object manifest:
+
+```yaml
+apiVersion: database.aws.crossplane.io/v1beta1
+kind: RDSInstance
+metadata:
+  name: rdsmysql
+spec:
+  forProvider:
+    region: us-east-1
+    dbInstanceClass: db.t3.medium
+    masterUsername: masteruser
+    allocatedStorage: 20
+    engine: mysql
+    engineVersion: "5.6.35"
+    skipFinalSnapshotBeforeDeletion: true
+  providerConfigRef:
+    name: cool-aws-creds # name of ProviderConfig above
+  writeConnectionSecretToRef:
+    namespace: crossplane-system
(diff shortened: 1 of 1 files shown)


docs/guides/multi-tenant.md
```diff
@@ -0,0 +1,299 @@
...
+namespace scope using a `Role`. This allows for giving users the ability to
+create and and manage `MySQLInstances` in their given namespace, but not the
+ability to see those defined in other namespaces.
+
+Futhermore, because the `metadata.namespace` is a field on the XRC, patching can
+be utilized to configure managed resources based on the namespace in which the
+corresponding XRC was defined. This is especially useful if a platform builder
+wants to designate specific credentials or a set of credentials that users in a
+given namespace can utilize when provisioning infrastructure using an XRC. This
+can be accomplished today by creating one or more `ProvideConfig` objects that
```
tobru, line 187:
```suggestion
can be accomplished today by creating one or more `ProviderConfig` objects that
```

docs/guides/multi-tenant.md
```diff
@@ -0,0 +1,299 @@
...
+include the name of the namespace in the `ProviderConfig` name. For example, if
+any `MySQLInstance` created in the `team-1` namespace should use specific AWS
+credentials when th provider controller creates the underlying `RDSInstance`,
```
tobru, line 190:
```suggestion
credentials when the provider controller creates the underlying `RDSInstance`,
```

tobru (approved):
Very good write-up which absolutely helps to better understand the Crossplane concepts. Just a few typos which need correction.

I wonder if it would make sense to point to the [Crossplane Service Broker](https://github.com/vshn/crossplane-service-broker) together with the [Service Catalog](https://github.com/kubernetes-sigs/service-catalog) as another form of cross-cluster multi-tenancy? What do you think?

hasheddan:
@tobru Thanks for the review!

> I wonder if it would make sense to point to the Crossplane Service Broker together with the Service Catalog as another form of cross-cluster multi-tenancy? What do you think?

I think this could definitely be a nice improvement, especially if y'all have actual experience with doing it. Would you be interested with following up on this guide with the additions?

docs/guides/multi-tenant.md
```diff
@@ -0,0 +1,299 @@
...
+indent: true
+---
+
+This guide describes how to use Crossplane effectively in multi-tenant
+environments by utilizing Kubernetes primitives and compatible policy
+enforcement projects in the cloud-native ecosystem.
+
+- [Background](#background)
+    - [Cluster Scoped Managed Resources](#cluster-scoped-managed-resources)
+    - [Namespace Scoped Claims](#namespace-scoped-managed-resources)
```
jbw976, line 14: this link is broken, i think the current subsection title would make it `#namespace-scoped-claims`

docs/guides/multi-tenant.md
```diff
@@ -0,0 +1,299 @@
...
+`ProviderConfig`, obtaining the credentials it points to, and using those
+credentials to reconcile the `RDSInstance`. This means that anyone who has been
+given [RBAC] to manage `RDSInstance` objects can use any credentials to do so.
+In practice, Crossplane assumes that only folks acting as infrastructure
+administrators or platform builders will interact directly with cluster-scoped
+resources.
+
+### Namespace Scoped Claims
+
+While managed resources exist at the cluster scope, composite resource, which
```
jbw976, line 91: composite resources (plural)?

jbw976 (approved):
This is great @hasheddan! i like how a number of concepts are pulled together here to make the scenarios and options more robust.

You can consider creating a TL;DR section that is a really high level and short summary of the approach - that could help frame this in an easily consumable way, in addition to all the great details contained within

docs/guides/multi-tenant.md
```diff
@@ -0,0 +1,299 @@
...
+...
+resources:
+- base:
+    apiVersion: database.aws.crossplane.io/v1beta1
+    kind: RDSInstance
+    spec:
+      forProvider:
+      ...
+  patches:
+  - fromFieldPath: spec.claimRef.namespace
```
muvaf, line 222: It might be worth mentioning the limitation of this approach, i.e. `ProviderConfig` has to be named same as the namespace (or some other, using `map` transform) and you can use only one `ProviderConfig` per namespace.

muvaf (approved):
Looks pretty good! I left one comment and agree with @jbw976 that a TLDR that talks about the `spec.providerConfigRef.name` patch would be nice.

hasheddan:
@jbw976 @muvaf thanks for y'all's reviews! Your suggestions have been incorporated :)

tobru:
> Would you be interested with following up on this guide with the additions?

Yes, I'll prepare a PR for it soon.

peterbosalliandercom:
@hasheddan In this documentation the provider-config is linked to aws-creds, is it also possible to do multitenancy with iam-role instead (as described here? : https://github.com/crossplane/provider-aws/blob/master/AUTHENTICATION.md)

Issue #2284, opened 2021-05-03
Can't delete XRD due to missing "DeleteOptions"


negz:
### What happened?

When deleting an XRD I'm seeing the following event, and my XRD refuses to delete.

```console
Warning  TerminateComposite  2m37s                   defined/compositeresourcedefinition.apiextensions.crossplane.io  cannot delete defined composite resources: no kind "DeleteOptions" is registered for version "admissionregistration.k8s.io/v1" in scheme "k8s.io/apiextensions-apiserver/pkg/apiserver/apiserver.go:52"
```

Interestingly I never actually created any XRs using this XRD.

### How can we reproduce it?

Here's the XRD I'm using:

```yaml
apiVersion: apiextensions.crossplane.io/v1
  kind: CompositeResourceDefinition
  metadata:
    creationTimestamp: "2021-05-03T22:41:16Z"
    deletionGracePeriodSeconds: 0
    deletionTimestamp: "2021-05-03T22:41:26Z"
    finalizers:
    - defined.apiextensions.crossplane.io
    generation: 2
    name: clusterconformances.test.crossplane.io
    resourceVersion: "950166"
    uid: f43c8d95-25d0-4b0d-99d2-57961b8ebf19
  spec:
    claimNames:
      kind: Conformance
      plural: conformances
    group: test.crossplane.io
    names:
      kind: ClusterConformance
      plural: clusterconformances
    versions:
    - name: v1alpha1
      referenceable: true
      schema:
        openAPIV3Schema: {}
      served: true
  status:
    conditions:
    - lastTransitionTime: "2021-05-03T22:41:26Z"
      reason: TerminatingCompositeResource
      status: "False"
      type: Established
    - lastTransitionTime: "2021-05-03T22:41:26Z"
      reason: TerminatingCompositeResourceClaim
      status: "False"
      type: Offered
    controllers:
      compositeResourceClaimType:
        apiVersion: test.crossplane.io/v1alpha1
        kind: Conformance
      compositeResourceType:
        apiVersion: test.crossplane.io/v1alpha1
        kind: ClusterConformance
```

This is the first time I'm testing composition with a Kubernetes cluster newer than 1.16. Perhaps this is due to the newer Kubernetes version I'm running?

### What environment did it happen in?
Crossplane version: 

```console
$ kubectl version
[...]

PR #2285, opened 2021-05-04
Use a slimmer scheme, in part to fix an obscure scheme behaviour

negz:

Fixes https://github.com/crossplane/crossplane/issues/2284

This PR includes two commits. The first appears to be the minimum change we can make to fix the above bug. I don't have a complete understanding of what exactly is happening in that bug, but it appears that if we don't emulate [this `metav1.AddToGroupVersion` call](https://github.com/kubernetes/client-go/blob/8c8fa70f/kubernetes/scheme/register.go#L138) that the client-go `scheme.Scheme` makes (including calling it _before_ calling `scheme.AddToScheme`) we see weird errors. Specifically when we try to call the controller-runtime client's `DeleteAllOf` method on an arbitrary type of `*unstructured.Unstructured` resource we see:

```console
no kind "DeleteOptions" is registered for version "admissionregistration.k8s.io/v1" in scheme "k8s.io/apiextensions-apiserver/pkg/apiserver/apiserver.go:52"
```

I suspect `admissionregistration.k8s.io/v1` is only relevant because it's the first thing in the scheme. I still have no idea why [the scheme that lives at `apiserver.go:52`](https://github.com/kubernetes/apiextensions-apiserver/blob/v0.20.1/pkg/apiserver/apiserver.go#L52) specifically is sad, because that appears to be a completely different scheme from any I'd expect us to actually be using.

The second commit fixes the issue in a slightly different way - by avoiding using `scheme.Scheme` at all. Instead we add only the types that we actually use in Crossplane, avoiding adding superfluous API groups like `admissionregistration`. This should help to avoid red herrings like this in future, but will require us to remember to add any Kubernetes types we start using to our scheme.

I have:

Primarily I've confirmed that I can successfully delete XRDs with these commits in place. However, I've also run the nascent https://github.com/crossplane/conformance test suite, which currently mostly ensures the package manager is behaving as expected (e.g. creating XRDs, provider deployments etc)

### DIFF
--- a/cmd/crossplane/main.go
+++ b/cmd/crossplane/main.go
@@ -23,7 +23,9 @@ import (
 	"gopkg.in/alecthomas/kingpin.v2"
 	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	extv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/kubernetes/scheme"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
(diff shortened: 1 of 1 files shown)


hasheddan (approved):
LGTM! Thanks for tracking this down!

PR #2298, opened 2021-05-07
Add FromConstantValue Composition patch type

### LINKED ISSUE #2013: Add constant value patch type in Composition
### Use Case

While building a resource catalog, we are leveraging Helm to template Compositions and CompositeResourceDefinitions for individual resources. On the input of such chart, we define upstream (provider) resource kind/apiVersion, some resource defaults (those go to `spec.forProvider` dierctly), some configurables (those go to `patches`). On top of that, we specify values that are global to the cluster, such as `accoundID` and `region`, that are required as part of some individual parameters for `base` resoures. The cluster wide parameters should be fixed in a Composition instead of being set at the creation of each CompositeResource.

We stumbled upon an issue with setting a cluster-wide value `region` in a unified manner. Most of the cloud resources have that setting under `spec.forProvider.region`, but not all - so we wanted to allow ourselves to define an override for the target resource spec path, where the Composition would inject the hadcoded `region` value. One example of such exception is AWS S3 bucket, that expects region restriction defined through ` spec.forProvider.locationConstraint`.

To allow an override, we tried something like this in composition template:
```yaml
{{- range .Values.supportedResources }}
---
kind: Composition                                                            
metadata:                                                                    
  name: x{{ include "xxx.resourcePlural" . }}.xxx.com
spec:
  compositeTypeRef:                                      
    apiVersion: xxx.com/v1alpha1               
    kind: X{{- .kind }}                                  
  resources:                                             
    - base:                                              
        apiVersion: {{ .apiVersion }}                    
        kind: {{ .kind }}                                
        spec:                                            
          providerConfigRef:                             
            name: {{ $.Values.defaultProvider | quote }} 
          forProvider:                                   
            {{- if (empty .overrideRegionPath) }}        
            region: {{ $.Values.aws.region | quote }}
            {{- end }}
      patches:
        {{- with .overrideRegionPath }}                                                              
        - fromFieldPath: "kind" # Whatever, just put it there!                                       
          toFieldPath: {{ . | quote }}                                                               
          transforms:                                                                                
            - type: string                                                                           
              string:                                                                                
                fmt: {{ $.Values.aws.region | quote }}                                               
        {{- end }}
#... 
{{- end }}
```

This does not work however, because the `string.fmt` syntax does not include any `%s`, which results in an error from `Sprintf`.
Neither leaving out `fromFieldPath` nor setting it to an empty string worked either. 

Provide a means for Composition to feed a raw value to a specified JSONPath of the target base.

stevendborrelli:

This implements a `FromConstantValue` patch type for compositions. 

Fixes #2013 : "Add ConstantValue patch in composition" 

I have:

I have created a Composition test at https://github.com/stevendborrelli/composition-examples/tree/main/FromConstantValuePatch and tested it against a 1.2.1 Crossplane build.

This results in a DB being created with the properties:
```yaml
kind: CloudSQLInstance
metadata:
  annotations:
    crossplane.io/composition-resource-name: cloudsqlinstance
    crossplane.io/external-name: cv-test-db-tbzwp-x2dqn
    test.crossplane.io/constant-value: constant annotation
  creationTimestamp: "2021-05-07T18:19:04Z"
  finalizers:
  generateName: cv-test-db-tbzwp-
  generation: 4
  labels:
    crossplane.io/claim-name: cv-test-db
    crossplane.io/composite: cv-test-db-tbzwp
  name: cv-test-db-tbzwp-x2dqn
  ownerReferences:
  - apiVersion: database.example.org/v1alpha1
    controller: true
    name: cv-test-db-tbzwp
    uid: acd55391-8df9-4f69-835a-e12f8b17371c
  resourceVersion: "266474"
  uid: 096bb318-caac-4c80-9a0f-e0c6c0f1ee18
spec:
  forProvider:
    databaseVersion: POSTGRES_9_6
    gceZone: us-west1-b
    instanceType: CLOUD_SQL_INSTANCE
    region: us-west1
    settings:
      activationPolicy: ALWAYS
      availabilityType: ZONAL
      backupConfiguration:
        startTime: "10:00"
      dataDiskSizeGb: 20
      dataDiskType: PD_SSD
      ipConfiguration:
        authorizedNetworks:
        - value: 0.0.0.0/0
        ipv4Enabled: true
      locationPreference:
        zone: us-west1-b
      pricingPlan: PER_USE
      replicationType: SYNCHRONOUS
      storageAutoResize: true
      tier: db-custom-1-3840
      userLabels:
        crossplane-kind: cloudsqlinstance_database_gcp_crossplane_io
        crossplane-name: cv-test-db-tbzwp-x2dqn
        crossplane-providerconfig: default
  providerConfigRef:
    name: default
  writeConnectionSecretToRef:
    name: acd55391-8df9-4f69-835a-e12f8b17371c-postgresql
    namespace: crossplane-system
``

### CODE REVIEW

File: apis/apiextensions/v1/composition_types.go [hunk 1]
```diff
@@ -47,6 +47,11 @@ const (
 	errFmtTransformTypeFailed          = "%s transform could not resolve"
 	errFmtMapTypeNotSupported          = "type %s is not supported for map transform"
 	errFmtMapNotFound                  = "key %s is not found in map"
+
+	errConstantValue                 = "type %s requires constantValue to be set"
+	errConstantValueTypeNotDefined   = "constantValue type not defined"
+	errConstantValueTypeNotSupported = "constantValue type %s is not supported"
+	errRequiredValue                 = "constant value for for type %s is missing"
```
negz, line 54: Nit: We have a convention of using an `errFmt` prefix for anything that is a format string, rather than a plain string.

apis/apiextensions/v1/composition_types.go
```diff
@@ -265,6 +276,64 @@ type PatchPolicy struct {
...
+const (
+	ConstantTypeBool   ConstantType = "bool"
+	ConstantTypeString ConstantType = "string"
+	ConstantTypeInt    ConstantType = "int"
+)
+
+// A ConstantValue is constant value to patch
+type ConstantValue struct {
+	// Type of the Constant, defaults to string.
+	// +kubebuilder:validation:Enum=bool;int;string
```
negz, line 292: I recently learned that you can associate this particular validation with the type declaration. I have a slight preference for keeping it there since it's perhaps a little closer to where the allowed values are defined.

apis/apiextensions/v1/composition_types.go
```diff
@@ -277,6 +346,8 @@ func (c *Patch) Apply(from, to runtime.Object, only ...PatchType) error {
 		return c.applyFromFieldPathPatch(from, to)
 	case PatchTypeToCompositeFieldPath:
 		return c.applyFromFieldPathPatch(to, from)
+	case PatchTypeFromConstantValue:
+		return c.applyFromConstantValuePatch(to)
```
negz, line 350: Nothing to do in this particular PR, but we should get rid of this "from" and "to" notation sometime soon. It made sense originally, but less so now per https://github.com/crossplane/crossplane/pull/2137/files#r593597291.

apis/apiextensions/v1/composition_types.go
```diff
@@ -607,7 +709,7 @@ const (
 	ConnectionDetailTypeUnknown                 ConnectionDetailType = "Unknown"
 	ConnectionDetailTypeFromConnectionSecretKey ConnectionDetailType = "FromConnectionSecretKey"
 	ConnectionDetailTypeFromFieldPath           ConnectionDetailType = "FromFieldPath"
-	ConnectionDetailTypeFromValue               ConnectionDetailType = "FromValue"
+	ConnectionDetailTypeFromValue               ConnectionDetailType = "From Value"
```
negz, line 712: This seems like an unintentional change.

apis/apiextensions/v1/composition_types_test.go
```diff
@@ -489,6 +503,127 @@ func TestConvertResolve(t *testing.T) {
 	}
 }
 
+func TestGetConstantValue(t *testing.T) {
+	type args struct {
+		ct ConstantValue
+	}
```
negz, line 509: Nit: Usually when there's only one arg we eschew the `args` struct, so in this case we'd have ` ct` field directly in the `map[string]struct{}` of cases. That said, there's no harm in keeping it.

apis/apiextensions/v1/composition_types.go
```diff
@@ -205,6 +210,7 @@ const (
 	PatchTypeFromCompositeFieldPath PatchType = "FromCompositeFieldPath" // Default
 	PatchTypePatchSet               PatchType = "PatchSet"
 	PatchTypeToCompositeFieldPath   PatchType = "ToCompositeFieldPath"
+	PatchTypeFromConstantValue      PatchType = "FromConstantValue"
```
negz, line 213: Currently I believe we support patching from a constant value to a composed resource. I wonder if there's any value / use cases in patching from a constant value back to the composite resource. Similar to how we have `FromCompositeFieldPath` and `ToCompositeFieldPath`.

It seems like it would be fairly straightforward to support this in either "direction" (we'd just need to additionally support calling `applyFromConstantValuePatch(from)`). The tricky part is figuring out what we'd call the two patch types.

My strawman proposal would be:

* `CompositeFieldPathFromComposedValue` patches the composite resource.
* `ComposedFieldPathFromConstantValue` patches the composed resource.

I'd love to hear better options though, because these strings are a little too long for my liking and we've previously heard that it's hard to distinguish between `Composite` and `Composed` in strings like this.

negz (commented):
This looks awesome, thanks @stevendborrelli! I'd like to see if we can make this patch work in both directions, but other than that and a few trivial nitpicks the implementation looks good to go

apis/apiextensions/v1/composition_types.go
stevendborrelli: yes it was a inadvertent space.  I've pushed a fix in 21a616de.

apis/apiextensions/v1/composition_types.go
stevendborrelli: I'm a little confused with this one, did you mean to reference the Enum or the bool/int/string constants, where I am using a kubebuilder marker for the type.

```go
// Bool is a constant boolean value
	// +kubebuilder:validation:Type=boolean
	// +optional
	Bool *bool `json:"bool,omitempty"`
```

Assuming it is the latter, I will remove the unnecessary marker.

stevendborrelli: markers for bool and int removed in b440ebc9

rozcietrzewiacz:
I think there is a general problem in the fact that `PatchType` field is quite overused to indicate different concepts, don't you think @stevendborrelli & @negz? It is already used to describe two different characteristics of the patch:
 1. Patch direction: `ToCompositeFieldPath` vs `FromCompositeFieldPath`
 2. Patch definition location: `PatchSet` (when not defined "in place")

I would personally rather see those split into separate fields, like
1. `direction` with possible values `Up`, `Down`, `ToComposite` and `FromComposite` - where `Up` is an alias for `FromComposite`, while `Down` means `ToComposite`. It could be just me, but `ToCompositeFieldPath` and `FromCompositeFieldPath` always require a long pause when reading yaml... I just started appending those with either `#UP` or `#DOWN` comments in my manifests :laughing: 
2. `fromPatchSet`, which - if set - would indicate the patchSet name, therefore removing the need for separate `patchSetName` field. Then "including" a previously defined patch would only require a single line instead of two.

So the (now proposed) `FromConstantValue` seems like yet another category: it describes the type of the source for the *value* to be used. Currently this characteristics is also covered by `FromCompositeFieldPath`, which creates a kind of conflict with the "direction" category... And then there is also the pending #2093, introducing `FromManyCompositeFieldPaths`, adding a fourth dimension into `type` - the indication of plurality of sources...

Which brings me to some crazy revolutionary suggestion which breaks everything :rofl:
What if the `patches` looked like this?

```yaml
- # Equivalent of the current "ToCompositeFieldPath" type
  from:
    childFieldPath: "some.base.resource.path"
  to:
    parentFieldPath: "some.composite.path"

- # Equivalent of the current "FromCompositeFieldPath" type
  from:
    parentFieldPath: "some.composite.path"
  to:
    childFieldPath: "some.base.resource.path"

- from:
    constantValue:
      type: "int"
      int: 100
  to:
    childFieldPath: "spec.forProvider.settings.maxDiskSize"  # Equivalent of "ComposedFieldPathFromConstantValue"
# OR parentFieldPath: "..." - equivalent of "CompositeFieldPathFromComposedValue"

- # Equivalent of the FromManyCompositeFieldPaths type as proposed in #2093 
  from:
    parentFieldPaths:
      - spec.parameters.fieldOne
      - spec.parameters.fieldTwo
  to:
    childFieldPath: "spec.forProvider.settings.maxDiskSize"
  transforms:
    - type: combine
      combine:
        type: string
          string:
            fmt: "Field One: %x - Field Two: %x"

- # Equivalent of `type: PatchSet` + `patchSet: someName`. 
  fromPatchSet: someName
  # ^- this and `to/from` should be mutually exclusive. Maybe should not even be part of `patches`, but rather referenced at a higher level...
```

Apart from my subjective opinion about readability, this approach would decouple the concepts of retrieving data and moving it around, so `parentFieldPaths` could become `childFieldPaths` to reference composed instead of composite. 

But I know, I'm coming way too late with these ideas, as they would call for a serious refactor in both the api and code. 

However I don't see how we can otherwise move forward without adding `UnreadableNewVerySpecificTypeNames` for each possible combination of the concepts mixed together inside the `type` :man_shrugging:

apis/apiextensions/v1/composition_types.go
negz: I was referring to the enum - apologies for the vaguery. What I was aiming to communicate was that this is an option too:

```go
// ConstantType is the type of constant
// +kubebuilder:validation:Enum=bool;int;string
type ConstantType string
```

Very much a nitpick, but I do think there's a little value in keeping the validation closer to where the allowed values are defined.

negz:
> But I know, I'm coming way too late with these ideas, as they would call for a serious refactor in both the api and code.

Yeah, unfortunately we can't introduce ideas like these without going to v2 of the Composition API. :( That said, we should capture ideas like this for when we are ready to make that breaking change.

Side note: The original idea was in fact to have a `direction` field per https://github.com/crossplane/crossplane/issues/1639, but somewhere amidst the implementation it was replaced with different patch types. I suspect this means we'll be stuck with defining two types for most patch concepts going forward - e.g. "from constant value to composite" and "from constant value to composed".

apis/apiextensions/v1/composition_types.go
negz: Hrm, actually, on reflection I don't think it really makes sense to patch from a constant value back to the composite resource - at least not in this particular context. 🤔 We're patching from an array of patches that are associated with a specific composed resource, but I can't imagine that a constant value will really be associated with a particular composed resource.

I wonder if it makes sense to support an array of patches from the Composition in general back to the XR? In some cases we can just use OpenAPI defaulting at the XR level (e.g. if there should be a constant default value regardless of Composition) but I could imagine wanting to set a constant XR value based on which Composition was selected.

Note - we don't need to bring this idea into scope for this PR. What I'd like to determine is whether we really think we'd need to patch from a constant back to the XR _from a particular composed resource's templates array_. If we don't think this is a real use case we can just ignore this thread and open a separate issue to track patching from a constant value to the XR.

PR #2299, opened 2021-05-07
Propose a few consistent conventions for Crossplane annotation metadata

negz:

These annotations are primarily relevant to package metadata, but could be used elsewhere too (e.g. on CRDs) in future. The idea here is to recommend conventions, rather than _requiring_ or being strongly opinionated about metadata.

In practice we're using very similar annotations to a lot of these already - see for example https://github.com/crossplane/provider-gcp/blob/75ccf4bc/package/crossplane.yaml. If this design is ratified I'll take a pass over all known providers and configurations and _add_ this new metadata, leaving any existing metadata in place to give consuming systems time to change over (if they want to).

I have:

I've created a `Secret` with the exact annotations proposed by this design to ensure that they're valid and can be created.

### DIFF
--- a/design/one-pager-package-annotations.md
+++ b/design/one-pager-package-annotations.md
@@ -63,8 +63,9 @@ metadata:
       provider-example is a really great Crossplane provider...
 
     # The provider's SVG icon URI. The SVG should be optimized for 24, 48, 65px
-    # and 2x versions.
-    meta.crossplane.io/iconData: data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciLz4KBIWXMA
+    # and 2x versions. Consumers are encouraged to support at least data URIs.
+    # https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs
+    meta.crossplane.io/iconURI: data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciLz4KBIWXMA
 
     # A mapping of concepts to friendly, human centric names.
     friendly-name.meta.crossplane.io/database.example.org: "Databases"
--- a/design/one-pager-package-annotations.md
+++ b/design/one-pager-package-annotations.md
@@ -1,0 +1,82 @@
+# Package Annotations
+
+- Owner: Nic Cope (@negz)
+- Reviewers: Crossplane Maintainers
+- Status: Proposed
+
+## Background
+
+[Packages][packages-v2] extend Crossplane with new functionality. The `Provider`
+and `Configuration` Crossplane resources are powered by packages; they represent
+a declarative intent to install a provider or configuration package.
+
+Crossplane packages are OCI images that deliver a payload of valid Kubernetes
+resources. Provider packages deliver CRDs for the managed resources a provider
+defines, while Configurations deliver XRDs and Compositions. All packages
+contain general package metadata in the form of a `crossplane.yaml` file, per
+the [package format one pager][format-v2].
+
+The `crossplane.yaml` file contains Kubernetes-like YAML, and is of kind
+`Provider` or `Configuration` in API group `meta.pkg.crossplane.io`. The package
+manager propagates any annotations from the `crossplane.yaml` file to its
+corresponding `ProviderRevision` or `ConfigurationRevision` in the Kubernetes
+API server. Such annotations are referred to as 'passthrough metadata' about a
+package, because they are not meaningful to Crossplane or the package manager.
+
+The package format one pager provides an example of a `crossplane.yaml` with
+annotations such as `company` and `maintainer`, but is not opinionated about
+the format of the annotations, or which annotations should exist.
+
+## Goals
+
+The goal of this document is to encourage - but not require - consistent
+package annotations. e.g. To increase the likelihood that all package authors
+will align on using the `maintainer` annotation to indicate the package's
+maintainer.
+
+## Proposal
+
+This document recommends that packages include the following annotations:
+
+```yaml
+apiVersion: meta.pkg.crossplane.io/v1alpha1
+kind: Provider
+metadata:
+  name: example
+  annotations:
+    # The package's maintainers, as a short opaque text string.
+    meta.crossplane.io/maintainer: Example T. Maintainer <example@example.org>
+
+    # The URL at which the package's source can be found.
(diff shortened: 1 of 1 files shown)


design/one-pager-package-annotations.md
```diff
@@ -0,0 +1,81 @@
...
+    meta.crossplane.io/description: |
+      The Google Cloud Platform (GCP) Crossplane provider adds support for
+      managing GCP resources in Kubernetes.
+
+    # A longer description, documentation, etc.
+    meta.crossplane.io/readme: |
+      provider-example is a really great Crossplane provider...
+
+    # The provider's icon as a base64 encoded SVG.
+    metadata.crossplane.io/iconData: PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciLz4KBIWXMA
```
hasheddan, line 66: I noticed the other annotations use `meta.crossplane.io`, should this one be the same?

hasheddan (approved):
This seems like a relatively non-controversial change as it is just proposing suggested patterns that I think are already being used in some places. Thanks @negz! One minor comment below :)

design/one-pager-package-annotations.md
negz: Ah yes - good catch!

design/one-pager-package-annotations.md
```diff
@@ -0,0 +1,81 @@
...
+    meta.crossplane.io/description: |
+      The Google Cloud Platform (GCP) Crossplane provider adds support for
+      managing GCP resources in Kubernetes.
+
+    # A longer description, documentation, etc.
+    meta.crossplane.io/readme: |
+      provider-example is a really great Crossplane provider...
+
+    # The provider's icon as a base64 encoded SVG.
+    meta.crossplane.io/iconData: PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciLz4KBIWXMA
```
grantgumina, line 66: Are there icon size recommendations users should adhere to? I know Upbound Registry, for example, expects package icons to be 65px x 65px.

negz: Does that apply to SVGs? 🤔 I thought the 'scalable' bit meant they worked at any size.

negz:
See https://github.com/crossplane/provider-azure/pull/244 for an example of a provider being updated to use this format.

design/one-pager-package-annotations.md
negz: More generally though, I'm happy to recommend whatever conventions we have today. If there's an emergent convention of 130 x 130px icons I'm happy to encode that in a recommendation.

thephred: They have a native/default width and height and a viewBox but they are scalable.

negz: Updated; let me know if the new wording does not make sense.

design/one-pager-package-annotations.md
```diff
@@ -0,0 +1,82 @@
...
+    # A longer description, documentation, etc.
+    meta.crossplane.io/readme: |
+      provider-example is a really great Crossplane provider...
+
+    # The provider's SVG icon URI. The SVG should be optimized for 24, 48, 65px
+    # and 2x versions.
+    meta.crossplane.io/iconData: data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciLz4KBIWXMA
+
+    # A mapping of concepts to friendly, human centric names.
+    friendly-name.meta.crossplane.io/database.example.org: "Databases"
```
muvaf, line 70: Do we need singular/plural?

design/one-pager-package-annotations.md
```diff
@@ -0,0 +1,82 @@
...
+    friendly-name.meta.crossplane.io/database.example.org/ExampleDBInstance: "ExampleDB Instance"
```
muvaf, line 71: What does this exactly refer to?

design/one-pager-package-annotations.md
```diff
@@ -0,0 +1,82 @@
...
+      The Google Cloud Platform (GCP) Crossplane provider adds support for
+      managing GCP resources in Kubernetes.
+
+    # A longer description, documentation, etc.
+    meta.crossplane.io/readme: |
+      provider-example is a really great Crossplane provider...
+
+    # The provider's SVG icon URI. The SVG should be optimized for 24, 48, 65px
+    # and 2x versions.
+    meta.crossplane.io/iconData: data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciLz4KBIWXMA
```
muvaf, line 67: Comment says URI but this seems like base64-encoded SVG.

negz: I'll clarify this in the commentary too.

design/one-pager-package-annotations.md
muvaf: Hmm got it. Would the format used in CRD names be more familiar to users, i.e. `friendly-name.meta.crossplane.io/exampledbinstances.database.example.org`? Though we'd probably need a differentiator between group friendly-name key and kind friendly-name key to make it easier for clients to parse.

Also, the group name is plural while this is singular. Should we go with the same choice for both?

design/one-pager-package-annotations.md
muvaf: Hmm got it. An example of URI value would also help.

muvaf (approved):
LGTM, a few nitpicks. Thanks @negz !

design/one-pager-package-annotations.md
muvaf: I really don't like having the upper-case as differentiator there since it's neglected in most of interactions of k8s users. Maybe `friendly-group-name.meta.crossplane.io` and `friendly-kind-name.meta.crossplane.io` or `kind-name...` and `group-name...`?

negz: I `friendly-group-name` and `friendly-kind-name` are good ideas, but we already have one consumer (Upbound Cloud) using the syntax proposed here, so my preference is to stick with this unless we think it's likely to be an issue.

FWIW despite `kubectl` etc using lowercase, I believe actual API types enforce title case kinds (i.e. they're case sensitive in the actual API). We expect these annotations to be written and consumed by fairly savvy folks so I'm don't personally see too much risk in requiring the kind case in the annotation to match the API.

negz: Okay, I've updated the PR to incorporate this feedback. :)

PR #2308, opened 2021-05-15
Add FromFn random string generator

### LINKED ISSUE #1895: Create Random Generator Transformer for Composition

Using the [provider-helm](https://github.com/crossplane-contrib/provider-helm), many Helm charts require the definition of a password. If no password is defined a random one is generated by the Helm templating engine which changes on each `helm upgrade` and might break the deployed application.
Asking for the password in a `parameter` is insecure because the pw won't be stored in a secret.

By providing a random string transformer such passwords could be generated on CompositeInstance creation and then be exposed in a connection secret (together with other info).

psinghal20:
This PR adds a new PatchType `FromFn` to patch the composed resource values through the specified function. One instance of this type of Patch is a `RandomString` generator added with this change. It generates a random string using the input length and Charset. 

Currently, I have disabled the `gosec 404` lint, which notifies the use of `math/rand` as a weak random number generator. Let me know if we should move towards the `crypto/rand` random number generator instead.

Fixes #1895 

I have:

Currently, I have tested these changes through the unit tests written alongside this PR. I will take a stab to deploy and try out with a sample composition.

Thanks for your help with this change @stevendborrelli! :smile: 


apis/apiextensions/v1/composition_types.go
```diff
@@ -349,6 +425,42 @@ func (c *Patch) applyFromFieldPathPatch(from, to runtime.Object) error { // noli
...
+	}
+
+	out := fnValue
+	for i, f := range c.Transforms {
+		if out, err = f.Transform(out); err != nil {
+			return errors.Wrapf(err, errFmtTransformAtIndex, i)
+		}
+	}
+
+	toMap, err := runtime.DefaultUnstructuredConverter.ToUnstructured(to)
```
stevendborrelli, line 452: I think with this PR and https://github.com/crossplane/crossplane/pull/2298, we are repeating this code pattern. I wonder if we have an opportunity to refactor out the field mapping & conversion (and maybe document what is happening internally a bit bette).

stevendborrelli:
@psinghal20 Thank you for submitting this! It looks good on first glance. Could you add test results from a composition showing the results?

psinghal20:
Hi @stevendborrelli, This is the composition file I used(which is similar to the getting started and your example setup)
```
---
kind: Composition
metadata:
  labels:
    provider: crossplane-provider-gcp
    guide: quickstart
spec:
  compositeTypeRef:
  resources:
    - name: cloudsqlinstance
      base:
        kind: CloudSQLInstance
        spec:
          forProvider:
            databaseVersion: POSTGRES_9_6
            region: us-west1
            settings:
              tier: db-custom-1-3840
              dataDiskType: PD_SSD
              ipConfiguration:
                authorizedNetworks:
                  - value: "0.0.0.0/0"
          writeConnectionSecretToRef:
            namespace: crossplane-system
      patches:
          transforms:
            - type: string
              string:
                fmt: "%s-postgresql"
        - type: FromFn
          fromFn:
            type: randomString
          toFieldPath: "metadata.annotations[test/random]"
        - type: FromFn
          fromFn:
            type: randomString
            randomString:
              length: 10
          toFieldPath: "metadata.annotations[test/random-only-length]"
        - type: FromFn
          fromFn:
            type: randomString
            randomString:
              charset: abcd
          toFieldPath: "metadata.annotations[test/random-only-charset]"
        - type: FromFn
          fromFn:
            type: randomString
            randomString:
              charset: abcd
              length: 10
          transforms:
            - type: string
              string:
                fmt: "%s-test-postgres"
          toFieldPath: "metadata.annotations[test/random-charset-length-transformer]"
      connectionDetails:
        - type: FromValue
          name: port
          value: "5432"

```
and here are the resulting annotations results:
```
Name:         random-test-db-62rzh-pg2ff
Namespace:    
Labels:       crossplane.io/claim-name=random-test-db
              crossplane.io/claim-namespace=default
              crossplane.io/composite=random-test-db-62rzh
Annotations:  crossplane.io/composition-resource-name: cloudsqlinstance
              crossplane.io/external-name: random-test-db-62rzh-pg2ff
              test/random: 8**F%%pL9jNA1p&4gMEhoqqiF9j7897A
              test/random-charset-length-transformer: bdabdacaba-test-postgres
              test/random-only-charset: abbbbcdaccdadacbabaabdbaaccadbdb
              test/random-only-length: CdH25o2iIq
API Version:  database.gcp.crossplane.io/v1beta1
Kind:         CloudSQLInstance
Metadata:
  Creation Timestamp:  2021-05-17T17:52:17Z
  Generate Name:       random-test-db-62rzh-
  Generation:          3
  Owner References:
    API Version:     database.example.org/v1alpha1
    Controller:      true
    Kind:            CompositePostgreSQLInstance
    Name:            random-test-db-62rzh
    UID:             d0fd6fc4-420a-4eb5-938b-e418392c1d9d
  Resource Version:  9419
  Self Link:         /apis/database.gcp.crossplane.io/v1beta1/cloudsqlinstances/random-test-db-62rzh-pg2ff
  UID:               2c77afed-3bee-4b2e-972a-625dea84abd3
```

I had to change the type of `RandomString.Charset` from `*string` to `string` to enable the default value.

apis/apiextensions/v1/composition_types.go
```diff
@@ -205,6 +211,7 @@ const (
 	PatchTypeFromCompositeFieldPath PatchType = "FromCompositeFieldPath" // Default
 	PatchTypePatchSet               PatchType = "PatchSet"
 	PatchTypeToCompositeFieldPath   PatchType = "ToCompositeFieldPath"
+	PatchTypeFromFn                 PatchType = "FromFn"
```
negz, line 214: https://github.com/crossplane/crossplane/pull/2298#discussion_r628663555

Cross-linking the above thread, which could be relevant in this context too. Would there ever be a case in which we wanted to patch "FromFn" back to the XR, not to the composed resource? If so, would we want to do that from the context of a particular composed resource - would the "FromFn" function ever take properties of the composed resource as inputs?

apis/apiextensions/v1/composition_types.go
```diff
@@ -243,6 +250,73 @@ type Patch struct {
 	// Policy configures the specifics of patching behaviour.
 	// +optional
 	Policy *PatchPolicy `json:"policy,omitempty"`
+
+	// FromFn is a function whose output is patched instead of a reference to a
+	// field path
+	// +optional
+	FromFn *FromFn `json:"fromFn,omitempty"`
```
negz, line 257: I know it was me who suggested `fromFn` as a pattern in https://github.com/crossplane/crossplane/issues/1895, but I feel like this API needs some more thought. In particular, it's a little confusing that we would have "fromFn" (which takes no inputs) and "transform functions" that do take inputs. 🤔

apis/apiextensions/v1/composition_types.go
psinghal20: I am not sure, but I think at least in the current scenario with `FromFn` having just the `RandomString` utility, I don't think there would be a need to communicate back a random string from the Composed resource to the XR.

apis/apiextensions/v1/composition_types.go
```diff
@@ -19,6 +19,7 @@ package v1
 import (
 	"encoding/json"
 	"fmt"
+	"math/rand" // #nosec G404
```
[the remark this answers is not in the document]
psinghal20: Yeah, sorry. I just silenced the underlying gosec lint. I will change this to golangci style linter comment.

apis/apiextensions/v1/composition_types.go
psinghal20: I got a little confused too. I saw transformers like `FromFn` but with the restriction of taking a `FromField` input, if remove that restriction `RandomString` can just be a transformer applied without input.

muvaf:
@psinghal20 Does that not generate a new random number in every reconciliation pass and set that since patches always go through regardless of whether the target field has a value? Have you tested it manually in a k8s cluster?

negz:
This unfortunately turned out to be more complicated than we thought. I don't think this will work as currently implemented and I think @psinghal20 has moved on to other things so I'm going to close it out. We'd still like this feature if we can figure out a way to avoid generating a new random number on every reconcile.

PR #2311, opened 2021-05-19
Remove OAM from Helm chart and guides

muvaf:

Crossplane charter no more includes OAM and KubeVela seems to replace OAM. This PR removes all optional OAM pieces from the Helm chart. Users can still use OAM/KubeVela to consume Crossplane CRDs.

I have:

`make reviewable` and normal installation. 


hasheddan (approved):
@muvaf LGTM, will this remove the subchart if you upgrade from a previous version?

negz (approved):

muvaf:
I've done additional tests.
* Installed 1.2.1, upgraded to this PR; no change.
* Installed 1.2.1 with `--set alpha.oam.enabled=true`, upgraded to this PR; CRDs are still in the cluster but OAM deployment is removed.

So, for people who installed Crossplane with OAM, I don't think there will be a disruptive change since OAM CRDs are unaffected. They will need to install other OAM projects to reconcile those CRDs separately.

Issue #2307, opened 2021-05-15
Propagate annotations from package meta to revision


negz:
### What problem are you facing?

https://github.com/crossplane/crossplane/blob/c5aa27/design/one-pager-package-format-v2.md#additional-metadata

Past negz told me through the medium of design documents that annotations in a package's `crossplane.yml` file would be propagated to the corresponding `ProviderRevision` or `ConfigurationRevision`, but that is in fact a lie. There's no code in Crossplane that actually does this propagation.

### How could Crossplane help solve your problem?

We could propagate the annotations from the package metadata to its corresponding revision. I'm pretty sure that will just involve asserting the `pkgMeta` in the revision reconciler to a `metav1.Accessor` (iirc) and calling `pr.SetAnnotations(pkgMeta.GetAnnotations())`.

https://github.com/crossplane/crossplane/blob/c5aa27/internal/controller/pkg/revision/reconciler.go#L390

PR #2321, opened 2021-05-21
Remove finalizer and do not requeue if lock is empty 

### LINKED ISSUE #2320: Remove / do not add finalizer from Lock if empty

Currently we [add a finalizer](https://github.com/crossplane/crossplane/blob/1f692aaf09d2d03162b38007e8c9ae01964dfc8b/internal/controller/pkg/resolver/reconciler.go#L178) to the singleton `Lock` resource that tracks our dependencies and we _never_ remove it. This guards against a user accidentally deleting the lock and changing dependency ordering, but makes it difficult to clean up a Crossplane installation as the `Lock` CRD and its extant singleton CR will never be deleted.

Revisions remove themselves from the `Lock` when marked inactive or deleted. @muvaf made the astute observation that if there are no packages installed in a Crossplane cluster, the `Lock` will be empty, and in this case, the finalizer can be removed from the `Lock` as deleting and recreating would not cause any problems. This would allow users more cleanly uninstall Crossplane after all packages have been successfully cleaned up.

hasheddan:

Updates dependency resolver controller to remove finalizer on lock and
exit early without requeue if the lock package list is empty. This
allows for cleanup of the lock when uninstalling Crossplane after all
packages have already been uninstalled.

Fixes #2320 

**If we consider this a bug fix, which I am inclined to do, per our [stated release policy](https://crossplane.io/docs/v1.2/reference/release-cycle.html#patch-releases) this should be a patch release for all active branches where applicable, which includes `v1.2`, `v1.1`, `v1.0` for this fix.**

I have:

1. Install Crossplane
2. Observe empty `Lock` with no finalizer
```yaml
🤖 (crossplane) k get lock -o yaml
apiVersion: v1
items:
  kind: Lock
  metadata:
    creationTimestamp: "2021-05-21T19:37:11Z"
    generation: 1
    name: lock
    resourceVersion: "692"
    uid: aaa25d0b-0e68-44af-ac6e-ec67d7414420
kind: List
metadata:
  resourceVersion: ""
  selfLink: ""
```
3. Install `provider-gcp`
4. Observe `Lock` with one entry and finalizer
```yaml
🤖 (crossplane) kxp install get lock -o yaml
apiVersion: v1
items:
  kind: Lock
  metadata:
    creationTimestamp: "2021-05-21T19:37:11Z"
    finalizers:
    - lock.pkg.crossplane.io
    generation: 2
    name: lock
    resourceVersion: "981"
    uid: aaa25d0b-0e68-44af-ac6e-ec67d7414420
  packages:
  - dependencies: []
    name: crossplane-provider-gcp-efb628debc28
    source: crossplane/provider-gcp
    type: Provider
    version: v0.16.0
kind: List
metadata:
  resourceVersion: ""
  selfLink: ""
```
5. Delete `provider-gcp`
6. Observe `Lock` with no entries and no finalizer
```yaml
🤖 (crossplane) k delete proget lock -o yaml
apiVersion: v1
items:
  kind: Lock
  metadata:
    creationTimestamp: "2021-05-21T19:37:11Z"
    generation: 3
    name: lock
    resourceVersion: "1597"
    uid: aaa25d0b-0e68-44af-ac6e-ec67d7414420
kind: List
metadata:
  resourceVersion: ""
  selfLink: ""
```
7. Uninstall Crossplane
8. Delete all Crossplane CRDs 
```
kubectl get crd -o name | grep crossplane.io | xargs kubectl delete
```
9. Observe that they are cleaned up successfully
```
🤖 (crossplane) k get crds
No resources found
```

### DIFF
--- a/internal/controller/pkg/resolver/reconciler.go
+++ b/internal/controller/pkg/resolver/reconciler.go
@@ -55,6 +55,7 @@ const (
 
 	errGetLock              = "cannot get package lock"
 	errAddFinalizer         = "cannot add lock finalizer"
+	errRemoveFinalizer      = "cannot remove lock finalizer"
 	errBuildDAG             = "cannot build DAG"
 	errSortDAG              = "cannot sort DAG"
 	errMissingDependencyFmt = "missing package (%s) is not a dependency"
@@ -172,9 +173,18 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		return reconcile.Result{}, errors.Wrap(resource.IgnoreNotFound(err), errGetLock)
 	}
 
-	// NOTE(hasheddan): we add a finalizer and never remove it to guard against
-	// a user accidentally deleting the lock, which should never be done.
-	// Consider instead recreating the lock if it is not present.
+	// If no packages exist in Lock then we remove finalizer and wait until a
+	// package is added to reconcile again. This allows for cleanup of the Lock
+	// when uninstalling Crossplane after all packages have already been
+	// uninstalled.
+	if len(lock.Packages) == 0 {
+		if err := r.lock.RemoveFinalizer(ctx, lock); err != nil {
+			log.Debug(errRemoveFinalizer, "error", err)
+			return reconcile.Result{RequeueAfter: shortWait}, nil
+		}
+		return reconcile.Result{}, nil
+	}
+
 	if err := r.lock.AddFinalizer(ctx, lock); err != nil {
 		log.Debug(errAddFinalizer, "error", err)
 		return reconcile.Result{RequeueAfter: shortWait}, nil
--- a/docs/reference/uninstall.md
+++ b/docs/reference/uninstall.md
@@ -82,7 +82,6 @@ Helm does not delete CRD objects. You can delete the ones Crossplane created
 with the following commands:
 
 ```console
-kubectl patch lock lock -p '{"metadata":{"finalizers": []}}' --type=merge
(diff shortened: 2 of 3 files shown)


internal/controller/pkg/resolver/reconciler.go
```diff
@@ -172,9 +173,18 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+	// If no packages exist in Lock then we remove finalizer and wait until a
+	// package is added to reconcile again. This allows for cleanup of the Lock
+	// when uninstalling Crossplane after all packages have already been
+	// uninstalled.
+	if len(lock.Packages) == 0 {
+		if err := r.lock.RemoveFinalizer(ctx, lock); err != nil {
+			log.Debug(errRemoveFinalizer, "error", err)
+			return reconcile.Result{RequeueAfter: shortWait}, nil
+		}
+		return reconcile.Result{}, nil
```
muvaf, line 185: Wouldn't this make that object never reconcile because the array has no element when the object is created?

muvaf:
I agree it's a bug fix but I'm not sure if it's critical enough to warrant a patch release. But I can be convinced that this fix and https://github.com/crossplane/crossplane/issues/2307 together could be worth a patch release.

internal/controller/pkg/resolver/reconciler.go
muvaf: Oh nevermind, I always think that the list is constructed by this controller but it's not, so it will come as filled by other controllers.

muvaf (approved):
LGTM if the manual tests pass.

PR #2331, opened 2021-05-24
Create Lock if not exist in revision controller

### LINKED ISSUE #2323: Lock can be lost on upgrades (or a while after upgrades)
### What happened?

In #2321 we added functionality to remove the `finalizer` on the singleton `Lock` when the number of installed packages == 0. This makes it simpler to uninstall Crossplane and delete its CRDs without having to manually remove a finalizer. However, as exposed in the test mentioned below, upgrading from a version prior to `v1.2.0` (where we switched to managing our CRDs via init container in https://github.com/crossplane/crossplane/pull/2160) can cause issues (**even if you first upgrade to `v1.2.x`**) because when the singleton `Lock` went from being created by Helm to not being present in the next chart version, meaning Helm deletes it (i.e. its `deletionTimestamp` is set). The reason why this has not been a problem thus far is because we _never_ removed the finalizer, so the `Lock` singleton was existing forever even though its `deletionTimestamp` was set. However, as soon as we remove that finalizer, as we now do in #2321, the `Lock` will get cleaned up, making all subsequent package revision reconciliations fail.

The simple solution for an end user that experiences this is to delete the `core` Crossplane `Pod` and let the `Deployment` recreate it, at which point our `initContainer` will recreate the `Lock`. However, it would be nice to automatically recover from this in case the `Lock` gets deleted in any case, which can be accomplished by having any revision that notices that the `Lock` does not exist recreate it. 

### How can we reproduce it?

The exact scenario is demonstrated in https://github.com/crossplane/test/runs/2645918064?check_suite_focus=true.

Crossplane version: `v1.1.0` --> `v1.3.0-rc.0.40.gd005048b`

hasheddan:

Updates revision controllers to create the lock if it does not exist
when resolving dependencies. If a package revision is inactive it will
not create the lock as its only purpose would be to remove itself from
it. If a package opts out of dependency management it will not create
the lock as it does not attempt to consider dependencies.

Fixes #2323 

I have:

1. Install Crossplane built from this PR
2. Delete the singleton `Lock`
3. Install `platform-ref-aws` and see `Lock` created with dependency info
```yaml
apiVersion: v1
items:
  kind: Lock
  metadata:
    creationTimestamp: "2021-05-24T18:24:11Z"
    finalizers:
    - lock.pkg.crossplane.io
    generation: 4
    name: lock
    resourceVersion: "8443"
    selfLink: /apis/pkg.crossplane.io/v1alpha1/locks/lock
    uid: 16434666-8535-4570-b17d-a127f3e58f64
  packages:
  - dependencies:
    - constraints: '>=v0.14.0-0'
      package: registry.upbound.io/crossplane/provider-aws
      type: Provider
    - constraints: '>=v0.3.6-0'
      package: registry.upbound.io/crossplane/provider-helm
      type: Provider
    name: upbound-platform-ref-aws-8f35c853335d
    source: registry.upbound.io/upbound/platform-ref-aws
    type: Configuration
    version: v0.1.2
  - dependencies: []
    name: crossplane-provider-aws-3b7ef8509173
    source: registry.upbound.io/crossplane/provider-aws
    type: Provider
    version: v0.18.1
  - dependencies: []
    name: crossplane-provider-helm-accc448deaad
    source: registry.upbound.io/crossplane/provider-helm
    type: Provider
    version: v0.7.0
kind: List
metadata:
  resourceVersion: ""
  selfLink: ""
```
4. Uninstall `provider-ref-aws` and see dependency information removed as well as finalizer.
```yaml
apiVersion: v1
items:
  kind: Lock
  metadata:
    creationTimestamp: "2021-05-24T18:24:11Z"
    generation: 11
    name: lock
    resourceVersion: "12300"
    selfLink: /apis/pkg.crossplane.io/v1alpha1/locks/lock
    uid: 16434666-8535-4570-b17d-a127f3e58f64
kind: List
metadata:
  resourceVersion: ""
  selfLink: ""
```

### DIFF
--- a/internal/controller/pkg/revision/dependency.go
+++ b/internal/controller/pkg/revision/dependency.go
@@ -22,6 +22,7 @@ import (
 	"github.com/Masterminds/semver"
 	"github.com/google/go-containerregistry/pkg/name"
 	"github.com/pkg/errors"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -37,7 +38,7 @@ const (
 	lockName = "lock"
 
 	errNotMeta                   = "meta type is not a valid package"
-	errGetLock                   = "cannot get lock"
+	errGetOrCreateLock           = "cannot get or create lock"
 	errIncompatibleDependencyFmt = "incompatible dependencies: %+v"
 	errMissingDependenciesFmt    = "missing dependencies: %+v"
 	errDependencyNotInGraph      = "dependency is not present in graph"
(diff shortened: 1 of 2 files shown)


jbw976 (approved):

internal/controller/pkg/revision/dependency.go
```diff
@@ -92,8 +93,18 @@ func (m *PackageDependencyManager) Resolve(ctx context.Context, pkg runtime.Obje
 
 	// Get the lock.
 	lock := &v1alpha1.Lock{}
-	if err := m.client.Get(ctx, types.NamespacedName{Name: lockName}, lock); err != nil {
-		return found, installed, invalid, errors.Wrap(err, errGetLock)
+	err = m.client.Get(ctx, types.NamespacedName{Name: lockName}, lock)
```
muvaf, line 96: Can we use `Apply` here?

hasheddan: @muvaf we perform operations based on the content of the Lock in the case it exists here so I don’t think so, but maybe I am not understanding your suggestion.

muvaf (approved):

PR #2345, opened 2021-06-02
feat(cli/crank): add wait flag in pkg install cmd in crossplane cli

### LINKED ISSUE #2182: CLI option to wait for installation to complete

We frequently see users reporting `no matches for kind "ProviderConfig" in version "aws.crossplane.io/v1beta1"` due to installation of provider not being completed yet or having some problems. However, since the command `kubectl crossplane install x` exits successfully in all cases as long as the CR is created (which is in line with usual kubectl behavior), they don't see the problem right away.

I always use `kind` with `kind create cluster --wait 5m` so that I don't try to use the cluster before it's ready. I think a similar flag for our CLI could be very helpful. It'd exit successfully only if the installation is healthy and ready to be used. This would also take the burden from users to check if the installation is ready by another kubectl command in their CI/CD pipelines.

psinghal20:
This PR introduces a wait flag in the install cmd to wait for the installed package to be healthy.

Fixes #2182 

I have:

Currently, none of the unit tests cover this CLI. This change has been mainly tested locally.

### DIFF
--- a/cmd/crank/install.go
+++ b/cmd/crank/install.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"net/http"
 	"strings"
+	"time"
 
 	"github.com/alecthomas/kong"
 	"github.com/google/go-containerregistry/pkg/name"
@@ -47,10 +48,18 @@ const (
 	errKubeClient    = "failed to create kube client"
 )
 
+type waitFlag bool
+
 // installCmd installs a package.
 type installCmd struct {
+	WaitFlag      waitFlag           `name:"wait" help:"Wait for installation of package"`
 	Configuration installConfigCmd   `cmd:"" help:"Install a Configuration package."`
 	Provider      installProviderCmd `cmd:"" help:"Install a Provider package."`
+}
+
+func (w waitFlag) BeforeApply(ctx *kong.Context) error { // nolint:unparam
+	ctx.Bind(5 * time.Minute)
+	return nil
 }
 
 // Run runs the install cmd.
@@ -69,7 +78,7 @@ type installConfigCmd struct {
 }
 
 // Run runs the Configuration install cmd.
-func (c *installConfigCmd) Run(k *kong.Context, logger logging.Logger) error {
+func (c *installConfigCmd) Run(k *kong.Context, logger logging.Logger, waitDuration time.Duration) error { //nolint:gocyclo
 	rap := v1.AutomaticActivation
 	if c.ManualActivation {
 		rap = v1.ManualActivation
@@ -120,6 +129,30 @@ func (c *installConfigCmd) Run(k *kong.Context, logger logging.Logger) error {
 		logger.Debug("Failed to create configuration", "error", warnIfNotFound(err))
 		return errors.Wrap(warnIfNotFound(err), "cannot create configuration")
 	}
+	if err = waitUntil(waitDuration, func() error {
+		logger.Debug("Waiting for the configuration to be ready")
+		res, err := kube.Configurations().Get(context.Background(), pkgName, metav1.GetOptions{})
+		if err != nil {
+			logger.Debug("Failed to fetch created configuration", "error", err)
+			return errors.Wrap(err, "failed to fetch configuration")
+		}
+		var configStatus corev1.ConditionStatus
+		for _, condition := range res.Status.Conditions {
+			if condition.Type == v1.TypeHealthy {
+				configStatus = condition.Status
+			}
+		}
+		if configStatus == corev1.ConditionTrue {
+			logger.Debug("Configurtion is ready")
+			return nil
+		}
+
+		logger.Debug("Configuration is not ready")
+		return errors.Errorf("Configuration is not ready")
+	}); err != nil {
+		logger.Debug("Configuration not ready in wait duration", "error", err)
+		return err
+	}
 	_, err = fmt.Fprintf(k.Stdout, "%s/%s created\n", strings.ToLower(v1.ConfigurationGroupKind), res.GetName())
 	return err
 }
@@ -136,7 +169,7 @@ type installProviderCmd struct {
 }
 
 // Run runs the Provider install cmd.
-func (c *installProviderCmd) Run(k *kong.Context, logger logging.Logger) error {
+func (c *installProviderCmd) Run(k *kong.Context, logger logging.Logger, waitDuration time.Duration) error { //nolint:gocyclo
 	rap := v1.AutomaticActivation
 	if c.ManualActivation {
 		rap = v1.ManualActivation
@@ -192,6 +225,30 @@ func (c *installProviderCmd) Run(k *kong.Context, logger logging.Logger) error {
 		logger.Debug("Failed to create provider", "error", warnIfNotFound(err))
 		return errors.Wrap(warnIfNotFound(err), "cannot create provider")
 	}
+	if err = waitUntil(waitDuration, func() error {
+		logger.Debug("Waiting for the provider to be ready")
+		res, err := kube.Providers().Get(context.Background(), pkgName, metav1.GetOptions{})
+		if err != nil {
+			logger.Debug("Failed to fetch created provider", "error", err)
+			return errors.Wrap(err, "failed to fetch provider")
+		}
+		var configStatus corev1.ConditionStatus
+		for _, condition := range res.Status.Conditions {
+			if condition.Type == v1.TypeHealthy {
+				configStatus = condition.Status
+			}
+		}
+		if configStatus == corev1.ConditionTrue {
+			logger.Debug("Provider is ready")
+			return nil
+		}
+
+		logger.Debug("Provider is not ready")
+		return errors.Errorf("Provider is not ready")
+	}); err != nil {
+		logger.Debug("Provider not ready in wait duration", "error", err)
+		return err
+	}
 	_, err = fmt.Fprintf(k.Stdout, "%s/%s created\n", strings.ToLower(v1.ProviderGroupKind), res.GetName())
 	return err
 }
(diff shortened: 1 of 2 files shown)


cmd/crank/install.go
```diff
@@ -120,6 +129,30 @@ func (c *installConfigCmd) Run(k *kong.Context, logger logging.Logger) error {
...
+			return errors.Wrap(err, "failed to fetch configuration")
+		}
+		var configStatus corev1.ConditionStatus
+		for _, condition := range res.Status.Conditions {
+			if condition.Type == v1.TypeHealthy {
+				configStatus = condition.Status
+			}
+		}
+		if configStatus == corev1.ConditionTrue {
+			logger.Debug("Configurtion is ready")
```
hasheddan, line 146:
```suggestion
			logger.Debug("Configuration is ready")
```

hasheddan: Would also be good to move some of these messages to `const` values. Since they are used across both commands you could consider making them format strings so that you can pass "Configuration" or "Provider"

cmd/crank/install.go
```diff
@@ -120,6 +129,30 @@ func (c *installConfigCmd) Run(k *kong.Context, logger logging.Logger) error {
 		logger.Debug("Failed to create configuration", "error", warnIfNotFound(err))
 		return errors.Wrap(warnIfNotFound(err), "cannot create configuration")
 	}
+	if err = waitUntil(waitDuration, func() error {
```
hasheddan, line 132: Does this just wait for the whole duration in the happy path? I believe we should be stopping the wait as soon as our condition (package is healthy) is met. I recommend using something like [wait.PollImmediate](https://github.com/kubernetes/apimachinery/blob/aec8116c445f5fa9d277e64ca432d978afae4ae3/pkg/util/wait/wait.go#L508) :+1:

cmd/crank/install.go
```diff
@@ -120,6 +129,30 @@ func (c *installConfigCmd) Run(k *kong.Context, logger logging.Logger) error {
...
+		logger.Debug("Waiting for the configuration to be ready")
+		res, err := kube.Configurations().Get(context.Background(), pkgName, metav1.GetOptions{})
+		if err != nil {
+			logger.Debug("Failed to fetch created configuration", "error", err)
+			return errors.Wrap(err, "failed to fetch configuration")
+		}
+		var configStatus corev1.ConditionStatus
+		for _, condition := range res.Status.Conditions {
+			if condition.Type == v1.TypeHealthy {
+				configStatus = condition.Status
```
hasheddan, line 142: after switching to conditional wait, you can just return here

cmd/crank/install.go
```diff
@@ -47,12 +48,20 @@ const (
...
+
 // installCmd installs a package.
 type installCmd struct {
+	WaitFlag      waitFlag           `name:"wait" help:"Wait for installation of package"`
 	Configuration installConfigCmd   `cmd:"" help:"Install a Configuration package."`
 	Provider      installProviderCmd `cmd:"" help:"Install a Provider package."`
 }
 
+func (w waitFlag) BeforeApply(ctx *kong.Context) error { // nolint:unparam
+	ctx.Bind(5 * time.Minute)
```
hasheddan, line 61: This should be configurable by the user

cmd/crank/install.go
psinghal20: No, it actually returns in case of the happy path. If you take a look at this `waitUntil` function ([here](https://github.com/crossplane/crossplane/pull/2345/files#diff-373f6bfdec661e4a082de6049df5e7c0da1ae221df32c76cdbe61e7d6463f5e9R272)) if the `try` function returns `nil` then it returns immediately. For each try method, we return nil as soon as the resource is ready([here](https://github.com/crossplane/crossplane/pull/2345/files#diff-373f6bfdec661e4a082de6049df5e7c0da1ae221df32c76cdbe61e7d6463f5e9R243))

I believe this does kind of the same thing as `wait.PollImmediate` but we can replace this with `PollImmediate` which will give us a standardized approach.

hasheddan: Ah I see, we are returning an error while condition is not yet met. I prefer `wait.PollImmediate` because we return a `(bool, error)` in the condition function, meaning we have the ability to:
1. Continue waiting when condition is not met: `(false, nil)`
2. Exit early when we encounter an error that is not related to the package not being healthy: `(false, err)`
3. Exit early when the package is healthy: `(true, nil)`

psinghal20: Right, it is a more flexible pattern and allows early exit in case of unrelated errors. I have made the above change.

cmd/crank/install.go
```diff
@@ -120,6 +136,30 @@ func (c *installConfigCmd) Run(k *kong.Context, logger logging.Logger) error {
...
+		if configStatus == corev1.ConditionTrue {
+			logger.Debug(fmt.Sprintf(msgFmtPkgReady, "configuration"))
+			return true, nil
+		}
+
+		logger.Debug(fmt.Sprintf(msgFmtPkgNotReady, "configuration"))
+		return false, nil
+	}); err != nil {
+		logger.Debug(fmt.Sprintf(errFmtPkgNotReadyTimeout, "Configuration"), "error", err)
+		return errors.Wrap(err, fmt.Sprintf(errFmtPkgNotReadyTimeout, "Configuration"))
```
hasheddan, line 161:
```suggestion
		return errors.Wrapf(err, errFmtPkgNotReadyTimeout, "Configuration")
```

cmd/crank/install.go
```diff
@@ -192,6 +233,30 @@ func (c *installProviderCmd) Run(k *kong.Context, logger logging.Logger) error {
...
+		if configStatus == corev1.ConditionTrue {
+			logger.Debug(fmt.Sprintf(msgFmtPkgReady, "provider"))
+			return true, nil
+		}
+
+		logger.Debug(fmt.Sprintf(msgFmtPkgNotReady, "provider"))
+		return false, nil
+	}); err != nil {
+		logger.Debug(fmt.Sprintf(errFmtPkgNotReadyTimeout, "Provider"), "error", err)
+		return errors.Wrap(err, fmt.Sprintf(errFmtPkgNotReadyTimeout, "Provider"))
```
hasheddan, line 258:
```suggestion
		return errors.Wrapf(err, errFmtPkgNotReadyTimeout, "Provider")
```

cmd/crank/install.go
hasheddan: Looks like this one still needs to be addressed. You can also just use [`GetCondition()`](https://github.com/crossplane/crossplane/blob/d04a059fe341a941760a3a94ef07085ae7158365/apis/pkg/v1/interfaces.go#L90)

cmd/crank/install.go
```diff
@@ -120,6 +136,30 @@ func (c *installConfigCmd) Run(k *kong.Context, logger logging.Logger) error {
...
+			return false, errors.Wrap(err, fmt.Sprintf(errFmtFetchPkg, "configuration"))
+		}
+		var configStatus corev1.ConditionStatus
+		for _, condition := range res.Status.Conditions {
+			if condition.Type == v1.TypeHealthy {
+				configStatus = condition.Status
+			}
+		}
+		if configStatus == corev1.ConditionTrue {
+			logger.Debug(fmt.Sprintf(msgFmtPkgReady, "configuration"))
```
hasheddan, line 153: Apologies for the conflicting feedback here, but I think it may actually make sense to have a set of errors for `Configuration` and a set for `Provider`, both as `const` (i.e. `msgConfigurationReady` and `msgProviderReady`). This will keep us from having to `fmt.Sprintf` in these log messages. My apologies for going back and forth on this!

hasheddan (approved):
Nice work here @psinghal20! I checked this out locally and ran it with success :)

```
🤖 (crossplane) kubectl crossplane install configuration registry.upbound.io/xp/getting-started-with-aws:v1.2.2 -w 5m --verbose
2021-06-14T08:48:20.099-0400	DEBUG	Found kubeconfig	{"configurationName": "xp-getting-started-with-aws"}
2021-06-14T08:48:20.099-0400	DEBUG	Created Kubernetes client	{"configurationName": "xp-getting-started-with-aws"}
2021-06-14T08:48:20.112-0400	DEBUG	Waiting for the Configuration to be ready	{"configurationName": "xp-getting-started-with-aws"}
2021-06-14T08:48:20.114-0400	DEBUG	Configuration is not ready	{"configurationName": "xp-getting-started-with-aws"}
2021-06-14T08:48:30.115-0400	DEBUG	Waiting for the Configuration to be ready	{"configurationName": "xp-getting-started-with-aws"}
2021-06-14T08:48:30.123-0400	DEBUG	Configuration is not ready	{"configurationName": "xp-getting-started-with-aws"}
2021-06-14T08:48:40.115-0400	DEBUG	Waiting for the Configuration to be ready	{"configurationName": "xp-getting-started-with-aws"}
2021-06-14T08:48:40.117-0400	DEBUG	Configuration is not ready	{"configurationName": "xp-getting-started-with-aws"}
2021-06-14T08:48:50.115-0400	DEBUG	Waiting for the Configuration to be ready	{"configurationName": "xp-getting-started-with-aws"}
2021-06-14T08:48:50.117-0400	DEBUG	Configuration is not ready	{"configurationName": "xp-getting-started-with-aws"}
2021-06-14T08:49:00.115-0400	DEBUG	Waiting for the Configuration to be ready	{"configurationName": "xp-getting-started-with-aws"}
2021-06-14T08:49:00.116-0400	DEBUG	Configuration is not ready	{"configurationName": "xp-getting-started-with-aws"}
2021-06-14T08:49:10.115-0400	DEBUG	Waiting for the Configuration to be ready	{"configurationName": "xp-getting-started-with-aws"}
2021-06-14T08:49:10.122-0400	DEBUG	Configuration is not ready	{"configurationName": "xp-getting-started-with-aws"}
2021-06-14T08:49:20.115-0400	DEBUG	Waiting for the Configuration to be ready	{"configurationName": "xp-getting-started-with-aws"}
2021-06-14T08:49:20.121-0400	DEBUG	Configuration is not ready	{"configurationName": "xp-getting-started-with-aws"}
2021-06-14T08:49:30.115-0400	DEBUG	Waiting for the Configuration to be ready	{"configurationName": "xp-getting-started-with-aws"}
2021-06-14T08:49:30.122-0400	DEBUG	Configuration is ready	{"configurationName": "xp-getting-started-with-aws"}
configuration.pkg.crossplane.io/xp-getting-started-with-aws created
```

PR #2346, opened 2021-06-02
Add initial e2e tests for compositions

rahulgrover99:
Signed-off-by: Rahul Grover <rahulgrover99@gmail.com>

This introduces e2e tests for the composition engine using provider-nop. It adds two test cases: 
1. Check if the specified composite resource gets ready after some time or not.
2. Check if the nop resources specified in the spec get ready after the specified time or not. 

I have:

Tested manually and with supplied e2e tests with latest build of crossplane.

### DIFF
--- a/test/e2e/apiextensions/composition_test.go
+++ b/test/e2e/apiextensions/composition_test.go
@@ -1,0 +1,735 @@
+// +build e2e
+
+/*
+Copyright 2020 The Crossplane Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package apiextensions
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/utils/pointer"
+
+	corev1 "k8s.io/api/core/v1"
+	kextv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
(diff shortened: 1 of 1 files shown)


test/e2e/apiextensions/composition_test.go
```diff
@@ -0,0 +1,735 @@
...
+
+				t.Logf("Created nop-example %q", res.GetName())
+
+				if err := wait.PollImmediate(10*time.Second, 90*time.Second, func() (done bool, err error) {
+					d, err := dc.Resource(nopRes).Namespace("default").Get(context.TODO(), "nop-example", metav1.GetOptions{})
+					if err != nil {
+						t.Fatalf("Get nop-res %q: %v", nopresource.GetName(), err)
+					}
+
+					conditionList, found, err := unstructured.NestedSlice(d.Object, "status", "conditions")
```
hasheddan, line 324: One thing you could do here to make these unstructured objects easier to work with is use [the `composed.Unstructured` wrapper struct in crossplane-runtime](https://github.com/crossplane/crossplane-runtime/blob/36fc69eff96ecb5856f156fec077ed3f3c3b30b1/pkg/resource/unstructured/composed/composed.go#L60). Once you construct it, you can utilize the methods it offers, such as [`GetCondition`](https://github.com/crossplane/crossplane-runtime/blob/36fc69eff96ecb5856f156fec077ed3f3c3b30b1/pkg/resource/unstructured/composed/composed.go#L70).

hasheddan (commented):
This is looking really great @rahulgrover99, nice work! I have one suggestion below that I think could make a few of the steps a little bit easier :) Let me know if you have any questions or thoughts!

test/e2e/apiextensions/composition_test.go
jbw976: Great pointer @hasheddan, thanks for sharing that! that programming model looks nice :)

rahulgrover99: Thank you @hasheddan! That suggestion really made the code look concise and clean. 🎉

test/e2e/apiextensions/composition_test.go
```diff
@@ -0,0 +1,700 @@
...
+
+				t.Cleanup(func() {
+					t.Logf("Cleaning up provider %q.", prv.GetName())
+					if err := c.Get(ctx, types.NamespacedName{Name: prv.GetName()}, prv); resource.IgnoreNotFound(err) != nil {
+						t.Fatalf("Get provider %q: %v", prv.GetName(), err)
+					}
+					if err := c.Delete(ctx, prv); resource.IgnoreNotFound(err) != nil {
+						t.Fatalf("Delete provider %q: %v", prv.GetName(), err)
+					}
+					t.Logf("Deleted provider %q", prv.GetName())
```
hasheddan, line 102: We could run into flakiness in this test because packages are blocked from deletion by their revisions which could cause installation of the provider in a subsequent test to fail if the previous hasn't been completely cleaned up yet, but in reality cleanup should be almost instantaneous, so I feel comfortable with this strategy and we can revisit if this is causing problems in our integration suite :+1:

hasheddan (approved):
Thanks for this awesome work @rahulgrover99! Because `go test` [may run tests from different packages in parallel](https://golang.org/cmd/go/#hdr-Testing_flags) by default, we could run into some issues with these composition tests running alongside the existing package dependency tests, but that can be addressed by setting `-p 1` on the test invocation side. I also think the package dependency tests and these composition tests actually may not interfere with each other since they are using different packages and resources :+1:

PR #2353, opened 2021-06-04
Prevent superfluous external-name annotation on XRs

### LINKED ISSUE #2350: Composite reconciler initializes with empty string for external name
### What happened?

When you create a claim you'll see that the external-name annotation has `""` value in both claim and composite. When you propagate `metadata.annotations[crossplane.io/external-name]` or `metadata.annotations` in `Composition` and left the annotation empty in claim/composite, you'll see that resources with late-assigned IDs like AWS VPC will be created over and over again due to having annotation empty.

The cause of the problem is that [here](https://github.com/crossplane/crossplane/blob/ec0b383/internal/controller/apiextensions/claim/configurator.go#L66) we set an empty string if the composite is not yet created, then the claim reconciler [propagates](https://github.com/crossplane/crossplane/blob/b9a8f4b/internal/controller/apiextensions/claim/api.go#L67) that to claim as late-initialization. Doing these operations only if the value is not empty solves the problem.

### How can we reproduce it?

Create a claim/composite and look at annotations of both.

Crossplane version: 1.2.2

ulucinar:
Signed-off-by: Alper Rifat Ulucinar <ulucinar@users.noreply.github.com>

Fixes #2350

I have:

- [X ] Read and followed Crossplane's [contribution process].

Using getting-started guide's aws (with default VPC) configuration, I provisioned an XR via its claim. The superfluous external-name annotation is not observed. Please see the comment below for the semantics of this change.

### DIFF
--- a/internal/controller/apiextensions/claim/api.go
+++ b/internal/controller/apiextensions/claim/api.go
@@ -64,7 +64,11 @@ func (a *APIBinder) Bind(ctx context.Context, cm resource.CompositeClaim, cp res
 	}
 
 	// Propagate the actual external name back from the composite to the claim.
-	meta.SetExternalName(cm, meta.GetExternalName(cp))
+	if en, enExists := cp.GetAnnotations()[meta.AnnotationKeyExternalName]; enExists { // we may also add this as a utility to crossplane-runtime
+		meta.SetExternalName(cm, en)
+	} else {
+		meta.RemoveAnnotations(cm, meta.AnnotationKeyExternalName)
+	}
 
 	// We set the claim's resource reference first in order to reduce the chance
 	// of leaking newly created composite resources. We want as few calls that
--- a/internal/controller/apiextensions/claim/api.go
+++ b/internal/controller/apiextensions/claim/api.go
@@ -64,11 +64,10 @@ func (a *APIBinder) Bind(ctx context.Context, cm resource.CompositeClaim, cp res
 	}
 
 	// Propagate the actual external name back from the composite to the claim
-	// if the claim does not have one yet.
+	// if it's set.
 	// For dynamically provisioned composites, claim's external name is the
-	// authoritative source.
-	en := meta.GetExternalName(cp)
-	if meta.GetExternalName(cm) == "" && en != "" {
+	// source initially (if set) as expected.
+	if en := meta.GetExternalName(cp); en != "" {
 		meta.SetExternalName(cm, en)
 	}
 
--- a/internal/controller/apiextensions/claim/api.go
+++ b/internal/controller/apiextensions/claim/api.go
@@ -64,10 +64,9 @@ func (a *APIBinder) Bind(ctx context.Context, cm resource.CompositeClaim, cp res
 	}
 
 	// Propagate the actual external name back from the composite to the claim.
-	if en, enExists := cp.GetAnnotations()[meta.AnnotationKeyExternalName]; enExists { // we may also add this as a utility to crossplane-runtime
+	en := meta.GetExternalName(cp)
+	if meta.GetExternalName(cm) == "" && en != "" {
 		meta.SetExternalName(cm, en)
-	} else {
-		meta.RemoveAnnotations(cm, meta.AnnotationKeyExternalName)
 	}
 
 	// We set the claim's resource reference first in order to reduce the chance
--- a/internal/controller/apiextensions/claim/api.go
+++ b/internal/controller/apiextensions/claim/api.go
@@ -63,7 +63,10 @@ func (a *APIBinder) Bind(ctx context.Context, cm resource.CompositeClaim, cp res
 		return errors.New(errBindClaimConflict)
 	}
 
-	// Propagate the actual external name back from the composite to the claim.
+	// Propagate the actual external name back from the composite to the claim
+	// if the claim does not have one yet.
+	// For dynamically provisioned composites, claim's external name is the
+	// authoritative source.
 	en := meta.GetExternalName(cp)
 	if meta.GetExternalName(cm) == "" && en != "" {
 		meta.SetExternalName(cm, en)
--- a/internal/controller/apiextensions/claim/api_test.go
+++ b/internal/controller/apiextensions/claim/api_test.go
@@ -28,6 +28,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
+	"github.com/crossplane/crossplane-runtime/pkg/meta"
 	"github.com/crossplane/crossplane-runtime/pkg/resource"
 	"github.com/crossplane/crossplane-runtime/pkg/resource/fake"
 	"github.com/crossplane/crossplane-runtime/pkg/test"
@@ -53,11 +54,59 @@ func TestBind(t *testing.T) {
 	}
 
 	cases := map[string]struct {
-		reason string
-		fields fields
-		args   args
-		want   error
+		reason    string
+		fields    fields
+		args      args
+		want      error
+		wantClaim resource.CompositeClaim
 	}{
+		"ReconcileXRCExtNameFromXR": {
+			reason: "If existing XR already has an external-name, XRC's external-name should be set from it",
+			fields: fields{
+				c: &test.MockClient{
+					MockUpdate: test.NewMockUpdateFn(nil),
+				},
+				t: fake.SchemeWith(&fake.Composite{}, &fake.CompositeClaim{}),
+			},
+			args: args{
+				cm: &fake.CompositeClaim{
+					ObjectMeta: metav1.ObjectMeta{
+						Annotations: map[string]string{
+							meta.AnnotationKeyExternalName: "name-from-claim",
+						},
+					},
+					CompositeResourceReferencer: fake.CompositeResourceReferencer{
+						Ref: &corev1.ObjectReference{
+							APIVersion: fake.GVK(&fake.Composite{}).GroupVersion().String(),
+							Kind:       fake.GVK(&fake.Composite{}).Kind,
+							Name:       "wat",
+						},
+					},
+				},
+				cp: &fake.Composite{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "wat",
+						Annotations: map[string]string{
+							meta.AnnotationKeyExternalName: "name-from-composite",
+						},
+					},
+				},
+			},
+			wantClaim: &fake.CompositeClaim{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: map[string]string{
+						meta.AnnotationKeyExternalName: "name-from-composite",
+					},
+				},
+				CompositeResourceReferencer: fake.CompositeResourceReferencer{
+					Ref: &corev1.ObjectReference{
+						APIVersion: fake.GVK(&fake.Composite{}).GroupVersion().String(),
+						Kind:       fake.GVK(&fake.Composite{}).Kind,
+						Name:       "wat",
+					},
+				},
+			},
+		},
 		"CompositeRefConflict": {
 			reason: "An error should be returned if the claim is bound to another composite resource",
 			fields: fields{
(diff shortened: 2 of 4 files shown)


ulucinar:
The semantics of this change are as follows. The context is the interaction between a claim and the corresponding XR. external-name below refers to the external-name annotation:
- For already existing XRs (statically provisioned), the source of truth for the external-name is the XR. Even if an external-name is provided for the claim, claim's external-name is reconciled to (statically provisioned) XR's claim, and it's removed if the (statically provisioned) XR does not contain an external-name. This is motivated by https://github.com/crossplane/crossplane/blob/bab887dbc67bfc7937b70dac8bb53de13c86a23c/internal/controller/apiextensions/claim/configurator.go#L47
- We no longer add a superfluous external-name if the claim does not have one (and XR is dynamically-provisioned (via the claim))
- For dynamically-provisioned XRs (via their claims) if the claim contains an external-name annotation, it's also used for the XR. And as explained above, external-name on the claim is reconciled from the XR's external-name. This is motivated by:
https://github.com/crossplane/crossplane/blob/bab887dbc67bfc7937b70dac8bb53de13c86a23c/internal/controller/apiextensions/claim/api.go#L66

internal/controller/apiextensions/claim/api.go
```diff
@@ -64,7 +64,11 @@ func (a *APIBinder) Bind(ctx context.Context, cm resource.CompositeClaim, cp res
 	}
 
 	// Propagate the actual external name back from the composite to the claim.
-	meta.SetExternalName(cm, meta.GetExternalName(cp))
+	if en, enExists := cp.GetAnnotations()[meta.AnnotationKeyExternalName]; enExists { // we may also add this as a utility to crossplane-runtime
+		meta.SetExternalName(cm, en)
+	} else {
+		meta.RemoveAnnotations(cm, meta.AnnotationKeyExternalName)
+	}
```
muvaf, line 71:
```suggestion
	if meta.GetExternalName(cm) == "" && meta.GetExternalName(cp) != "" {
		meta.SetExternalName(cm, en)
	}
```

@negz can confirm but I believe the case mentioned here is actually just a late-initialization, i.e. copy the annotation from composite to claim if claim doesn't have a value for it. The bug is that it doesn't respect this order of propagation and overwrites claim's external name no matter what. We'd like to do that only if the claim doesn't have a value for it and composite does.

I'd also suggest using `meta.GetExternalName` because accounting for empty/nil string difference is useful in cases where empty string is actually a valid configuration. In this case, I'm not sure if it's a valid external name in any context. If I remember correctly, that's why we didn't implement it as `meta.GetExternalName(runtime.Object) *string`.

internal/controller/apiextensions/claim/configurator.go
```diff
@@ -63,7 +64,11 @@ func ConfigureComposite(_ context.Context, cm resource.CompositeClaim, cp resour
 	// external name (even if that external name was empty) in order to ensure
```
muvaf, line 64: I'm a bit confused about why we needed this logic at all. @negz `meta.AddAnnotations(cp, cm.GetAnnotations())` will not override the external name annotation if the claim doesn't have it. If it does, shouldn't it be respected? Let's say I have `bucket-name-test` as my external-name annotation in my claim so that I can get a bucket with that name, should I even be paired with a composite resource whose external name is different than that?

muvaf: My other comment also depends on this one. If we don't respect claim's annotation, then it's not late-init, hence keeping the functionality as is and only checking for `""` or `nil`(current change) would be sufficient.

internal/controller/apiextensions/claim/api.go
negz: Right - iirc that's the goal. We want to propagate the external name back from XR -> XRC _if_ there isn't an external name on the claim already. If there was an external name on the XRC we'd instead propagate it to the XR.

internal/controller/apiextensions/claim/configurator.go
negz: I think this was addressing the (exceedingly rare) edge case in which:

1. An XR is statically provisioned.
2. An XRC later explicitly claims the XR by setting its `spec.resourceRef`.
3. The XR and XRC both have different external-name annotations set at claim time.

In this case the XR's external name has probably already been applied. We can't just propagate the XRC's external name over the top of the XR's because doing so would presumably trickle down to managed resources and attempt to rename them in place, which almost certainly wouldn't work in most cases. So the idea here is that we basically in this edge case say "nope, we're ignoring the XRC's requested external name because the XR already has one".

In this specific edge case we probably do want to overwrite the XRC's external name with the real external name - the one from the XR. Alternately, I could get behind your suggestion that we should just refuse to let an XRC claim an XR if they have conflicting external names. Or going even further, generally don't allow a claim for a statically provisioned XR to ever specify an external name, since it would realistically never be honored.

muvaf: Hmm I can see two options then:
* Seems like this is the only place we account for external name annotation at this level. We could possibly not do anything special for external-name annotation on XRC-XR level, treat it just like any other annotation.
* Late initialize external-name annotation if the claim has empty external name annotation AND refuse to bind if both XRC and XR have non-empty external-names that are not same.

I can go either way but feel closer to the second option because we'd like composite to be similar to managed resources so that it's easier to treat it like one in multi-layered `Composition`s. @ulucinar what are your thoughts?

> Or going even further, generally don't allow a claim for a statically provisioned XR to ever specify an external name, since it would realistically never be honored.

Hmm I'd say that if user manually chose an XR with `resourceRef` and the external-name annotations of XRC and XR matches, it would be safe to bind the two.

negz: > I can go either way but feel closer to the second option.

I agree.

ulucinar: I don't have a strong opinion here, but in the light of https://github.com/crossplane/crossplane/issues/2350#issuecomment-856203139
my understanding is that if an XRC explicitly claims a statically provisioned XR by referencing it (via its `spec.resourceRef`) and the external-name annotations in XRC and XR are in conflict, it seems to me that we had better honor the binding request communicated by setting `spec.resourceRef` both because of a desire to behave in a backwards-compatible manner and because this is an explicit communication of an intent although it's conflicting with what's specified in the XRC's external-name annotation. What's more, considering the common use cases of annotations, it would, in my opinion, make sense to reconcile the annotation in XRC to reflect the annotation specified in the referenced (statically provisioned) XR, i.e., let the referenced XR override external-name annotation specified in XRC.

Please note that if we choose to do so, for a dynamically provisioned XR, the external-name annotation will initially be set to what's specified in the claim, assuming one is set. Later in the runtime, if an external entity explicitly sets the external-name on the XR to something else, we would be propagating it back to XR.

ulucinar:
We now have a behavior change in case there is a mismatch between a claim's external-name and the bound XR's external name: Controller now returns an error and refuses to reconcile the XR with the following error message:
```
mismatch of external names between claim and managed resource
```

internal/controller/apiextensions/claim/api.go
```diff
@@ -64,7 +64,10 @@ func (a *APIBinder) Bind(ctx context.Context, cm resource.CompositeClaim, cp res
 	}
 
 	// Propagate the actual external name back from the composite to the claim.
```
negz, line 66: It may be worth expanding on this comment to mention the fact that we do this conditionally in order to allow the claim's external name to be authoritative in cases where the XR does not yet have one.

internal/controller/apiextensions/claim/configurator.go
```diff
@@ -63,7 +65,15 @@ func ConfigureComposite(_ context.Context, cm resource.CompositeClaim, cp resour
...
 	// we don't try to rename anything after the fact.
 	if meta.WasCreated(cp) {
-		meta.SetExternalName(cp, en)
+		enCm := meta.GetExternalName(cm)
+		// if there is a mismatch between claim's external-name and composite's external-name
+		if enCm != "" && enCm != en {
+			return errors.New(errExternalNameMismatch)
+		}
+		// fix(2353): do not introduce a superfluous extern-name (empty external-names are treated as invalid)
```
negz, line 73: A few comment style nits - I believe the convention in this file is to:

1. Capitalize the first letter.
2. Wrap them at 80 chars.

Do you think we should establish a convention of linking to the PR number in comments using this `fix()` syntax? We've historically included full links to relevant issues/PRs when it's useful to do so.

internal/controller/apiextensions/claim/configurator_test.go
```diff
@@ -201,7 +202,7 @@ func TestCompositeConfigure(t *testing.T) {
 									// composite resource value, since it has
 									// most likely already taken effect and
 									// cannot be updated retroactively.
```
negz, line 204: I don't think the above comment is relevant anymore. Do we need this test case at all?

internal/controller/apiextensions/claim/configurator_test.go
```diff
@@ -270,6 +271,90 @@ func TestCompositeConfigure(t *testing.T) {
...
+					Unstructured: unstructured.Unstructured{
+						Object: map[string]interface{}{
+							"metadata": map[string]interface{}{
+								"namespace": ns,
+								"name":      name,
+								"annotations": map[string]interface{}{
+									// This should be reset to the equivalent
+									// composite resource value, since it has
+									// most likely already taken effect and
+									// cannot be updated retroactively.
```
negz, line 287: Is this comment true? I think we'll return an error rather than resetting this annotation.

negz (approved):
This looks good to me, apart from a few small nits.

WRT whether we return an error or overwrite the claim's external name when they conflict, I don't feel strongly either way. You make a good point at https://github.com/crossplane/crossplane/pull/2353#discussion_r650031397 @ulucinar and I'd also be happy to approve the variant you suggest if you feel that is the better path.

internal/controller/apiextensions/claim/configurator.go
ulucinar: It may be useful to standardize a concise format for such purposes similar to a `TODO(<user>)`.

internal/controller/apiextensions/claim/configurator_test.go
ulucinar: I'd like to keep this test case because it asserts that if the external names on a statically provisioned XR and the claim match, we should still be able to reconcile without any errors from `claim.ConfigureComposite`.  I have corrected the comment above it to reflect current expectations.

internal/controller/apiextensions/claim/configurator_test.go
ulucinar: Copy/paste error. Thanks for pointing it out. I've rewritten it to reflect current expectations.

internal/controller/apiextensions/claim/api_test.go
```diff
@@ -27,6 +27,8 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/crossplane/crossplane-runtime/pkg/meta"
+
```
negz, line 31: This whitespace should be removed so that all the runtime packages are grouped together.

internal/controller/apiextensions/claim/api.go
```diff
@@ -63,8 +63,14 @@ func (a *APIBinder) Bind(ctx context.Context, cm resource.CompositeClaim, cp res
...
 	}
 
-	// Propagate the actual external name back from the composite to the claim.
-	meta.SetExternalName(cm, meta.GetExternalName(cp))
+	// Propagate the actual external name back from the composite to the claim
+	// if it's set.
+	// For dynamically provisioned composites, claim's external name is the
+	// source initially (if set) as expected.
+	en := meta.GetExternalName(cp)
+	if en != "" {
```
negz, line 71:
```suggestion
	if en := meta.GetExternalName(cp); en != "" {
```

I just noticed we could do this to reduce the scope of the `en` variable. Just a nit pick though - feel free to ignore.

negz (approved):
This looks great @ulucinar! Thank you. Let me know when you're ready for me to merge this.

ulucinar:
Hi @negz,
After the backport changes are merged into their respective release branches, do we follow the same release procedure as `wesaas` for instance? I think I have read something in the lines of accumulating and releasing things in batches for crossplane, is this correct? Do I need to do anything else after this is merged?

negz:
@ulucinar I think we will want to include these in patch release "soon", but whether we release patch releases just for this bug or wait for a few more to build up is pretty open ended. We should definitely get the backports done, but we might want to hold off a little to see if we can get a few more bugs fixed before we cut patch releases.

Issue #1770, opened 2020-09-25
Generic Cross-Resource References


kasey:
### What problem are you facing?
As mentioned in the design doc for terraform-based controller codegen(https://github.com/crossplane/crossplane/pull/1677/) , we will need to implement a generic cross-resource referencing scheme to compensate for the fact that terraform does not provide the necessary metadata for our code generator to understand which fields are cross-referenceable. 

### How could Crossplane help solve your problem?

A design document would allow us to collect feedback on the syntax and behavior of these references.

negz:
Here's a strawman sketch of what generic cross-resource references might look like.

Today we have this for example:

```yaml
---
apiVersion: network.aws.crossplane.io/v1alpha2
kind: VPC
metadata:
  name: my-vpc
spec:
  ...
---
apiVersion: network.aws.crossplane.io/v1alpha2
kind: Subnet
metadata:
  name: my-subnet
spec:
  forProvider
    vpcId: vpc-33fk3fjdf2  # The VPC's actual external name
    # Used to resolve the VPC's actual external name.
    # The resolution logic is written in Go as part of
    # adding support for a new managed resource.
    vpcIdRef:
      name: my-vpc
  ...
```

With generic cross-resource references this might look like:
```yaml
---
apiVersion: network.aws.crossplane.io/v1alpha2
kind: VPC
metadata:
  name: my-vpc
spec:
  ...
---
apiVersion: network.aws.crossplane.io/v1alpha2
kind: Subnet
metadata:
  name: my-subnet
spec:
  # Open ended naming so we could extend this with other kinds
  # of field modifiers. Perhaps this is overthinking it?
  fieldModifiers:
  - fromRef:
      apiVersion: v1alpha2
      kind: VPC
      name: my-vpc
      fieldPath: metadata.annotations[crossplane.io/external-name]
    # Perhaps this should be scoped/limited to spec.forProvider?
    toFieldPath: spec.forProvider.vpcId
  forProvider:
    vpcId: vpc-33fk3fjdf2  # The VPC's actual external name
  ...
```

muvaf:
That looks pretty good to me. Do we want to allow taking values from namespaced resources?

negz:
Unstale please. I'm still pretty sure we're going to need this.

negz:
Throwing out another idea (based on a demo of something related @apelisse) put together - I could see a distinct `Reference` type being in some ways better than an array of references embedded in a resource.

So instead of...

```yaml
---
[...]

Issue #2384, opened 2021-06-16
Consider giving providers Secret and ConfigMap access only in their own namespace


muvaf:
### What problem are you facing?

In the `ClusterRole` of providers, we add `Secret`, `ConfigMap`, `Lease` and `Event` [access](https://github.com/crossplane/crossplane/blob/master/internal/controller/rbac/provider/roles/roles.go#L66) that's not limited to any namespace. While we have to have this kind of power in core Crossplane due to how claims work, but it may not be necessary for providers. The less arbitrary access we have to `Secret`s in the cluster, the better in general due to how it can be used to get tokens of other `ServiceAccount`s.

With the generic references on managed resources, this will mean edit permission on any managed resource will get you access to any `Secret` in the cluster.

### How could Crossplane help solve your problem?

It should be possible to use providers whose managed resources write their connection `Secret`s into only single namespace. For users who consume managed resources via composition, it doesn't really matter where the connection secret of the managed resources ends up. For others who provision managed resources directly and would like connection secrets to be in the same namespace as their `Pod`, this will mean a serious limitation unless they manually grant the permissions. We can consider an option in `ControllerConfig` that will toggle whether the provider has multi-namespace access to those objects or only the namespace where the provider `Pod` is running. Since it'd be breaking, the default would be to allow it. But we could consider defaulting to single namespace for new installations.

turkenh:
Extending `ControllerConfig` in a backward-compatible way for this sounds like a good idea to me. 

Considering, users might request [additional permissions](https://github.com/crossplane/crossplane/pull/2385#discussion_r654680659) at the configuration level, what do you think an API like the following? 

```
type ControllerConfigSpec struct {
        ...
	// +optional
	PermissionConfig PermissionConfig `json:"permissionConfig,omitempty"`
}

type PermissionConfig struct {
	DisableClusterScopedSecretAccess bool `json:"disableClusterScopedSecretAccess,omitempty"`
	DisableClusterScopedConfigmapAccess bool `json:"disableClusterScopedConfigmapAccess,omitempty"`

	ExtraPermissionRequests []PermissionRequest `json:"extraPermissionRequests,omitempty"`
}

type PermissionRequest struct {
	ClusterScoped bool `json:"clusterScoped,omitempty"`
	rbacv1.PolicyRule
}
```

hasheddan:
@turkenh this sounds fairly reasonable, but there are two main areas of concern from me:
[...]

PR #2385, opened 2021-06-16
Generic references design doc

muvaf:

The design doc for generic references which also addresses common data source problem.

Fixes https://github.com/crossplane/crossplane/issues/1770
Fixes https://github.com/crossplane/crossplane/issues/2099

I have:

N/A.

### DIFF
--- a/design/design-generic-references.md
+++ b/design/design-generic-references.md
@@ -33,10 +33,9 @@ put it in the referencer object.
 
 ## Proposal
 
-All managed resources and composite resources will have a top-level spec field
-that lists the generic references. They will be resolved in order and failure of
-any will cause reconciliation to fail. Inspired from the
-[latest developments in API conventions](https://github.com/kubernetes/community/pull/5748)
+All managed resources will have a top-level spec field that lists the generic
+references. They will be resolved in order and failure of any will cause
+reconciliation to fail. Inspired from the [latest developments in API conventions](https://github.com/kubernetes/community/pull/5748)
 regarding multi-kind references, the API will look like the following:
 ```yaml
 # A generic reference to another managed resource.
@@ -75,15 +74,14 @@ have to create necessary RBAC resources manually to grant the permissions.
 
 ### Implementation
 
-The field will exist in all managed resources, the API struct and the generic
-resolver will live in Crossplane Runtime and be integrated into managed reconciler.
-Providers will only have to update their runtime dependency and regenerate their
-CRDs to get the feature in.
+The new field will exist in all managed resources and the cross-resource reference
+fields will stay as is and we'll continue adding them wherever possible. They will
+still be the main way to refer to another resource. The generic references will
+help in cases where that's not possible or not implemented yet.
 
-In core Crossplane, the generic composite reconciler will use the same generic
-reference resolver and it will run it before creating any composed resource
-because unresolved fields may cause managed resources to provision resources
-with missing configuration.
+The API struct and the generic resolver will live in Crossplane Runtime and be
+integrated into managed reconciler. Providers will only have to update their
+runtime dependency and regenerate their CRDs to get the feature in.
 
 One of the caveats with following the new conventions is that controller-runtime
 utilities mostly assume that you have Group-Version-Kind of the resource you're
@@ -100,6 +98,41 @@ be issued. In line with cross-resource references, if the field on the reference
 object, pointed by `toFieldPath`, already has a value then the resolver will
 skip it, meaning it will resolve only once and users will need to delete that
 value for resolver to set it again.
+
+## Future Considerations
+
+### Transforms
+
+The retrieved value may not be in the form that's useful for the referencer
+resource directly and may need some formatting. If we see a great need for that,
+we can move the transform machinery from Crossplane to Crossplane Runtime and
+reuse it.
+
+### More Limited RBAC
+
+Currently, providers have access to all `Secret`s in the cluster. While this will
+provide a nice UX for generic reference users, it may not be desirable for security
+minded folks since it'd cause giving managed resource permission mean giving access
+to any `Secret` in the cluster. See [#2384](https://github.com/crossplane/crossplane/issues/2384).
+
+We can consider new settings in `Provider` or `ControllerConfig` to let users
+control this behavior with more granularity.
+
+### Composite Resources
+
+We could possibly have this on composite resources as well to keep them similar
+to managed resources, hence improving nested composition construction experience.
+But while its implementation would be straight-forward, there are some issues like
+whether we should let claim authors propagate this field as well and would that
+mean letting them access everything provider has access to, which is not few.
+
+### Obscure References
+
+There could be cases where you'd like to fetch a value from another resource but
+don't want to expose it on the managed resource object, i.e. make it available to
+controller to be inclduded in the calls but not visible on the CR. We could think
+of ways letting controller know that by either giving `toFieldPath: OBSCURED` or
+`makeAvailableAs: <some key in a map>`.
 
 ## Alternatives Considered
 
@@ -172,26 +205,7 @@ spec:
     toFieldPath: "spec.forProvider.directoryArn"
 ```
 
-The advantage of this option is that it will allow composite/managed resource to
+The advantage of this option is that it will allow composite/managed resources to
 get information from multiple resources to construct a single value, which is not
 possible today in any way. However, it's questionable how necessary it is since
 we haven't seen many users reporting that they need this functionality.
-
-## Future Considerations
-
-### Transforms
-
-The retrieved value may not be in the form that's useful for the referencer
-resource directly and may need some formatting. If we see a great need for that,
-we can move the transforms from composition to Crossplane Runtime and reuse it
-reference struct as well.
-
-### More Limited RBAC
-
-Currently, providers have access to all `Secret`s in the cluster. While this will
-provide a nice UX for generic reference users, it may not be desirable for security
-minded folks since it'd cause giving managed resource permission mean giving access
-to any `Secret` in the cluster. See [#2384](https://github.com/crossplane/crossplane/issues/2384).
-
-We can consider new settings in `Provider` or `ControllerConfig` to let users
-control this behavior with more granularity.
--- a/design/design-generic-references.md
+++ b/design/design-generic-references.md
@@ -1,0 +1,200 @@
+# Crossplane Agent for Consumption
+
+* Owner: Muvaffak Onuş (@muvaf)
+* Reviewers: Crossplane Maintainers
+* Status: Accepted
+
+## Background
+
+In many cases, Crossplane managed resources require information that is not
+available when the user creates the resource. One example is that if you'd like to create
+a `Subnet`, you need to supply a VPC ID but if you create the `VPC` at the same
+time then the ID will be available only after the creation, so you need to extract
+the ID and use it in `Subnet`. Today, it's possible to let the provider do this
+operation using cross-resource references. However, this requires knowing the kind of
+the referenced resource and how to extract the value during development time of
+the provider. There are several cases where this is not possible:
+
+* The referencer field may not always target a single kind. For example, a URL
+  field in Route53 could get populated by a URL of an S3 bucket or access endpoint
+  of an RDS Instance.
+* Instead of hard-coding region for all managed resources, you may want to refer
+  to a `ConfigMap` and let the managed resource take that information from there.
+* In Composition, you can trade values of the fields only between the children of
+  the same composite instance, but there are cases where you'd like to use a separate
+  composite-independent resource to store the information and serve all composites.
+
+In order to enable such use cases, we need a generic version of the cross-resource
+references where you can specify all metadata required to find the referenced
+resource as well as details about how to extract the value you need and where to
+put it in the referencer object.
+
+## Proposal
+
+All managed resources and composite resources will have a top-level spec field
+that lists the generic references. They will be resolved in order and failure of
+any will cause reconciliation to fail. Inspired from the
+[latest developments in API conventions](https://github.com/kubernetes/community/pull/5748)
+regarding multi-kind references, the API will look like the following:
+```yaml
+# A generic reference to another managed resource.
+spec:
+  references:
+  - group: ec2.aws.crossplane.io
+    version: v1alpha1
+    resource: vpcs
+    name: main-vpc
+    fromFieldPath: spec.forProvider.cidrBlock
+    toFieldPath: spec.cidrBlock
+```
+```yaml
+# A generic reference to a namespaced Kubernetes resource.
+spec:
+  references:
+  - version: v1
+    # group is omitted since it is empty for ConfigMap.
+    resource: configmaps
+    name: common-settings
+    namespace: crossplane-system
+    fromFieldPath: data.region
+    toFieldPath: spec.forProvider.region
+```
+
+The syntax of `fromFieldPath` and `toFieldPath` fields will be similar to what we
+use in Composition, in line with [Kubernetes API Conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#selecting-fields).
+
+The provider pods will resolve the references via managed reconciler, hence they
+will need at least read-only permissions for any resource that's been referenced
+by an instance. Currently, every provider [has read access](https://github.com/crossplane/crossplane/blob/master/internal/controller/rbac/provider/roles/roles.go#L66)
+to all of its own kinds, `Secret`s, `ConfigMap`s, `Lease`s and `Event`s in all
+namespaces. In order to reference resources other than the ones listed, users will
+have to create necessary RBAC resources manually.
+
+### Implementation
+
+Since the field will exist in all managed resources, the API struct and the generic
+resolver will live in Crossplane Runtime and be integrated into managed reconciler.
+Providers will only have to update their runtime dependency and regenerate their
+CRDs to get the feature in.
+
+The API struct will look roughly as the following:
+```go
+type GenericReference struct {
+    Resource      string `json:"resource"`
+    Name          string `json:"name"`
+    FromFieldPath string `json:"fromFieldPath"`
+    ToFieldPath   string `json:"toFieldPath"`
+    
+    // +optional
+    Group *string `json:"group,omitempty"`
+    // +optional
+    Version *string `json:"version,omitempty"`
+    // +optional
+    Namespace *string `json:"namespace,omitempty"`
+}
+```
+
+One of the caveat with following the new conventions is that controller-runtime
+utilities mostly assume that you have Group-Version-Kind of the resource you're
+operating with. In order to use Group-Version-Resource, we will make use of
+`EquivalentResourceRegistry` object as a translator between the two, implemented
+as an alternative to [`MustCreateObject`](https://github.com/crossplane/crossplane-runtime/blob/406fe0b/pkg/resource/resource.go#L145).
+After getting a `runtime.Object`, all controller-runtime utilities should be
+available to use.
+
+The resolver will make use of fieldpath library we're using for composition to
+get and set the values unknown types and once it is done. In line with cross-resource
+references, if the field on the referenced object, pointed by `toFieldPath`, already
+has a value then the resolver will skip it, meaning it will resolve only once and
+users will need to delete that value for resolver to set it again.
+
+## Alternatives Considered
+
+### More Open-ended API
+
+A variation of the following suggestion was made in the initial issue [#1770](https://github.com/crossplane/crossplane/issues/1770).
+
+```yaml
+spec:
+  fieldModifiers:
+    - fromRef:
+        apiVersion: ec2.aws.crossplane.io/v1beta1
+        kind: VPC
+        name: my-vpc
+        fieldPath: metadata.annotations[crossplane.io/external-name]
+      toFieldPath: spec.forProvider.vpcId
+```
+
+This API would allow future source additions to `fromRef` by exposing a type
+discriminator. However, it's likely that any additions will start to compete with
+what you can do with composition patches. Converting this to the new API convention,
+it's a pretty good alternative to the proposal above since it's safer but I'd like
+to see more use cases where we'd like to implement different source types.
+
+### Multiple Source References
+
+Composition allows information to flow from resource to another due to the nature
+of how patches are designed. However, there could be cases where you'd like to
+construct a single value using information from multiple resources, like an ARN.
+An API that would allow this could look like the following:
+
+```yaml
+spec:
+  references:
+  - combine:
+    - version: v1
+      resource: configmaps
+      name: common-settings
+      namespace: crossplane-system
+      fieldPath: data.region
+    - version: v1
+      resource: secrets
+      name: accounts
+      namespace: crossplane-system
+      fieldPath: data.prodAWSAccountID
+    fmt: "arn:aws:clouddirectory:%s:%s:schema/published/cognito/1.0"
+    toFieldPath: "spec.forProvider.directoryArn"
+```
+Or alternatively:
+```yaml
+spec:
+  references:
+  - variables:
+    - name: region
+      fromRef:
+        version: v1
+        resource: configmaps
+        name: common-settings
+        namespace: crossplane-system
+        fieldPath: data.region
+    - name: accountId
+      fromRef:
+        version: v1
+        resource: secrets
+        name: accounts
+        namespace: crossplane-system
+        fieldPath: data.prodAWSAccountID
+    template: "arn:aws:clouddirectory:{{ .region }}:{{ .accountId }}:schema/published/cognito/1.0"
+    toFieldPath: "spec.forProvider.directoryArn"
+```
+
+The advantage of this option is that it will allow composite resource to get
+information from multiple resources to construct a single value, which is not
+possible today. However, it's questionable how much necessary this is since
+we haven't seen many users reporting that they need this functionality.
+
+## Future Considerations
+
+### Transforms
+
+The retrieved value may not be in the form that's useful for the target resource
+directly and may need some formatting. If we see a great need for that, we can
+move the transforms from composition to Crossplane Runtime and use it in
+`GenericReference` object as well.
+
+### More Limited RBAC
+
+Currently providers have access to all `Secret`s in the cluster. While this will
+provide a nice UX for generic reference users, it may not be desirable for security
+minded folks since it'd cause giving managed resource permission mean giving access
+to any `Secret` in the cluster. See [#2384](https://github.com/crossplane/crossplane/issues/2384).
--- a/design/design-generic-references.md
+++ b/design/design-generic-references.md
@@ -1,4 +1,4 @@
-# Crossplane Agent for Consumption
+# Generic Resource References
 
 * Owner: Muvaffak Onuş (@muvaf)
 * Reviewers: Crossplane Maintainers
@@ -19,10 +19,12 @@ the provider. There are several cases where this is not possible:
   field in Route53 could get populated by a URL of an S3 bucket or access endpoint
   of an RDS Instance.
 * Instead of hard-coding region for all managed resources, you may want to refer
-  to a `ConfigMap` and let the managed resource take that information from there.
-* In Composition, you can trade values of the fields only between the children of
-  the same composite instance, but there are cases where you'd like to use a separate
-  composite-independent resource to store the information and serve all composites.
+  to a `ConfigMap` or a `Secret` and let the managed resource take that
+  information from there.
+* In Composition, you can use values of the fields only between the children of
+  the same composite instance, but there are cases where you'd like to use a
+  separate resource that is not bound to any composite instance to store the
+  information and serve all composites.
 
 In order to enable such use cases, we need a generic version of the cross-resource
 references where you can specify all metadata required to find the referenced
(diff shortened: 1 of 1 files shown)


AaronME (approved):

design/design-generic-references.md
```diff
@@ -0,0 +1,197 @@
...
+  information and serve all composites.
+
+In order to enable such use cases, we need a generic version of the cross-resource
+references where you can specify all metadata required to find the referenced
+resource as well as details about how to extract the value you need and where to
+put it in the referencer object.
+
+## Proposal
+
+All managed resources and composite resources will have a top-level spec field
```
muvaf, line 36: I'm not 100% sure whether we should do it for composite resources as well. It's useful at that level, too, and it'd make composite resources more similar to managed resources, which is something we want but there are questions to be answered like would we allow claims to be able to edit that? Wouldn't that mean cross-namespace information leak through core crossplane pod? If we won't allow claims to do that, then it's useful for only cases where you want to use composite as part of another composition. Has anyone doing that asked for it? The more I write, I feel inclined to move it to `Future Considerations` section and keep this specific to managed resources, which covers the use cases laid out under background.

design/design-generic-references.md
```diff
@@ -0,0 +1,197 @@
...
+    template: "arn:aws:clouddirectory:{{ .region }}:{{ .accountId }}:schema/published/cognito/1.0"
+    toFieldPath: "spec.forProvider.directoryArn"
+```
+
+The advantage of this option is that it will allow composite/managed resource to
+get information from multiple resources to construct a single value, which is not
+possible today in any way. However, it's questionable how necessary it is since
+we haven't seen many users reporting that they need this functionality.
+
+## Future Considerations
```
hasheddan, line 180: One feature that I think could be immediately useful to folks is "obscured references". This would mean that a sensitive value from a `Secret` (or other resource) could be used in a field, but it wouldn't actually be exposed on the resource. I imagine we could just populate the field in the `toFieldPath` with something like `OBSCURED`, but that could be subject to schema type constraints for resources...

design/design-generic-references.md
```diff
@@ -0,0 +1,197 @@
...
+The provider pods will resolve the references via managed reconciler, hence they
+will need at least read-only permissions for any resource that's been referenced
+by an instance. Currently, every provider [has read access](https://github.com/crossplane/crossplane/blob/d8f57a8/internal/controller/rbac/provider/roles/roles.go#L66)
+to all of its own kinds, `Secret`s, `ConfigMap`s, `Lease`s and `Event`s in all
+namespaces. In order to reference resources other than the ones listed, users will
+have to create necessary RBAC resources manually to grant the permissions.
+
+### Implementation
+
+The field will exist in all managed resources, the API struct and the generic
```
hasheddan, line 78: Do you imagine that we will preserve (and continue to support on new resources) the existing typed references? If so, could you call that out in this proposal?

design/design-generic-references.md
```diff
@@ -0,0 +1,197 @@
...
+
+The syntax of `fieldPath` and `toFieldPath` fields will be similar to what we
+use in Composition, in line with [Kubernetes API Conventions](https://github.com/kubernetes/community/blob/744e270/contributors/devel/sig-architecture/api-conventions.md#selecting-fields).
+
+The provider pods will resolve the references via managed reconciler, hence they
+will need at least read-only permissions for any resource that's been referenced
+by an instance. Currently, every provider [has read access](https://github.com/crossplane/crossplane/blob/d8f57a8/internal/controller/rbac/provider/roles/roles.go#L66)
+to all of its own kinds, `Secret`s, `ConfigMap`s, `Lease`s and `Event`s in all
+namespaces. In order to reference resources other than the ones listed, users will
+have to create necessary RBAC resources manually to grant the permissions.
```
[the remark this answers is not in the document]
muvaf: Yeah I was feeling good about this since I knew `Secret`s have to be available but saw `ConfigMap`s as well, at least in the same namespace. So most use cases should be covered with today's permissions and we can consider making additions to `ControllerConfig` in the future to allow toggling this [if we decide](https://github.com/crossplane/crossplane/issues/2384) to remove some of the permissions by default.

design/design-generic-references.md
```diff
@@ -0,0 +1,197 @@
...
+
+All managed resources and composite resources will have a top-level spec field
+that lists the generic references. They will be resolved in order and failure of
+any will cause reconciliation to fail. Inspired from the
+[latest developments in API conventions](https://github.com/kubernetes/community/pull/5748)
+regarding multi-kind references, the API will look like the following:
+```yaml
+# A generic reference to another managed resource.
+spec:
+  externalValues:
```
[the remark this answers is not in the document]
muvaf: Because of the shape of the API, it may not be referencers anymore, i.e. we might decide to add another source type. If you look at the alternatives considered, there is an option there that I considered first that has object ref directly but then I changed my mind to make it at least extendible in the future by putting the ref under `fromObject`. One possible extension I was thinking was giving the ability to have `fromMultipleObjects` and reduce multiple values from different resources into one.

design/design-generic-references.md
```diff
@@ -0,0 +1,197 @@
...
+use in Composition, in line with [Kubernetes API Conventions](https://github.com/kubernetes/community/blob/744e270/contributors/devel/sig-architecture/api-conventions.md#selecting-fields).
+
+The provider pods will resolve the references via managed reconciler, hence they
+will need at least read-only permissions for any resource that's been referenced
+by an instance. Currently, every provider [has read access](https://github.com/crossplane/crossplane/blob/d8f57a8/internal/controller/rbac/provider/roles/roles.go#L66)
+to all of its own kinds, `Secret`s, `ConfigMap`s, `Lease`s and `Event`s in all
+namespaces. In order to reference resources other than the ones listed, users will
+have to create necessary RBAC resources manually to grant the permissions.
+
+### Implementation
```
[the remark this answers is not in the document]
muvaf: > It is possible we just say "you can do this, but you're going to likely have a bad time"

I think that'd be a good decision because I'm sure that people will have legitimate needs for such possible things that we cannot think of right now. If it's possibly hazardous but also worth doing for them, I wouldn't want to stand in the way.

design/design-generic-references.md
muvaf: Hmm yeah makes sense. We're making sensitive fields secret reference but there might be cases where a thing is not sensitive on the surface but is for users. I'll add it.

design/design-generic-references.md
muvaf: Yes, added. They will stay as the recommended approach for those fields. Generic references are just for cases where that's not possible or not implemented yet.

design/design-generic-references.md
```diff
@@ -0,0 +1,197 @@
...
+[latest developments in API conventions](https://github.com/kubernetes/community/pull/5748)
+regarding multi-kind references, the API will look like the following:
+```yaml
+# A generic reference to another managed resource.
+spec:
+  externalValues:
+  - fromObject:
+      group: ec2.aws.crossplane.io
+      version: v1alpha1
+      resource: vpcs
```
muvaf, line 48: I'm open to keeping this like the old pattern, i.e. `apiVersion, kind` fields since the convention is very new, though that means referencing subresources might be cumbersome, which is one of the reasons why they went with that shape.

design/design-generic-references.md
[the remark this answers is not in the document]
muvaf: I don't love `externalValues` by the way. Open to suggestions.

design/design-generic-references.md
```diff
@@ -0,0 +1,211 @@
...
+have to create necessary RBAC resources manually to grant the permissions.
+
+### Implementation
+
+The new field will exist in all managed resources and the cross-resource reference
+fields will stay as is and we'll continue adding them wherever possible. They will
+still be the main way to refer to another resource. The generic references will
+help in cases where that's not possible or not implemented yet.
+
+The API struct and the generic resolver will live in Crossplane Runtime and be
```
chlunde, line 82: I think the previous version made it clear that references behaved like "required patches" and were resolved by core crossplane and the provider. Is it now only in the provider? I wonder how this will work if the managed resource is not valid before resolving the reference (schema/webhook).

I wonder if partially creating the managed resources (those without any unresolved dependencies) will be required in some cases.

chlunde (approved):
Looks very cool and you covered everything I thought of while reading it :+1:

design/design-generic-references.md
```diff
@@ -0,0 +1,197 @@
...
+have to create necessary RBAC resources manually to grant the permissions.
+
+### Implementation
+
+The field will exist in all managed resources, the API struct and the generic
+resolver will live in Crossplane Runtime and be integrated into managed reconciler.
+Providers will only have to update their runtime dependency and regenerate their
+CRDs to get the feature in.
+
+In core Crossplane, the generic composite reconciler will use the same generic
```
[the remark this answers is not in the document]
muvaf: I actually removed adding it to composite resources, see future considerations section. There are some other questions need to be answered if we add generic referencing to composites. For managed resources, the behavior will be same as cross-resources references, i.e. do not call any external client methods before resolving all references.

design/design-generic-references.md
```diff
@@ -0,0 +1,211 @@
...
+  field in Route53 could get populated by a URL of an S3 bucket or access endpoint
+  of an RDS Instance.
+* Instead of hard-coding region for all managed resources, you may want to refer
+  to a `ConfigMap` or a `Secret` and let the managed resource take that
+  information from there.
+* In Composition, you can use values of the fields only between the children of
+  the same composite instance, but there are cases where you'd like to use a
+  separate resource that is not bound to any composite instance to store the
+  information and serve all composites.
+
```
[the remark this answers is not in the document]
muvaf: Sounds like a good use case, though it feels like too specific to include in this section and sort of an advanced case (using claimRef.namespace in a patch of composition that will go to generic reference). I think the last use case covers it to some extent.

design/design-generic-references.md
```diff
@@ -0,0 +1,211 @@
...
+### More Limited RBAC
+
+Currently, providers have access to all `Secret`s in the cluster. While this will
+provide a nice UX for generic reference users, it may not be desirable for security
+minded folks since it'd cause giving managed resource permission mean giving access
+to any `Secret` in the cluster. See [#2384](https://github.com/crossplane/crossplane/issues/2384).
+
+We can consider new settings in `Provider` or `ControllerConfig` to let users
+control this behavior with more granularity.
```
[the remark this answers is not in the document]
muvaf: That's possible today, too, when you give a `Secret` reference as RDS instance password. If the provider has access to secrets in that namespace, it'll be able to get it for you. I opened https://github.com/crossplane/crossplane/issues/2384 proposing a restriction as I believe the installation namespace to store the managed resource connections secret should be enough for most use cases using composition.

design/design-generic-references.md
muvaf: I think it's helpful to think of this closer to cross-resource references like `bucketRef` on the managed resources rather than patches on composition. Today, you can create a managed resource with `bucketRef` whose name doesn't correspond to anything. You'll see an error in the status but the resource creation will go through. The generic references in the managed resources will work the same way in the sense that you'll be able to create the resource and provider pod will try to resolve all references, both cross-resource and generic refs, and only after doing that successfully it will start calling external client functions, `Connect`, `Observe` etc.

design/design-generic-references.md
```diff
@@ -0,0 +1,211 @@
...
+to all of its own kinds, `Secret`s, `ConfigMap`s, `Lease`s and `Event`s in all
+namespaces. In order to reference resources other than the ones listed, users will
+have to create necessary RBAC resources manually to grant the permissions.
+
+### Implementation
+
+The new field will exist in all managed resources and the cross-resource reference
+fields will stay as is and we'll continue adding them wherever possible. They will
+still be the main way to refer to another resource. The generic references will
+help in cases where that's not possible or not implemented yet.
```
negz, line 80: I like the idea of maintaining the existing references for the time being in order to avoid a breaking change, but if we adopt this new model my preference would be to deprecate the old explicitly-coded method. That way there is one clear way for users to define references.

design/design-generic-references.md
```diff
@@ -0,0 +1,211 @@
...
+      resource: vpcs
+      name: main-vpc
+      fieldPath: spec.forProvider.cidrBlock
+    toFieldPath: spec.cidrBlock
+```
+```yaml
+# A generic reference to a namespaced Kubernetes resource.
+spec:
+  externalValues:
+  - fromObject:
```
negz, line 56: Currently we support both explicit references and selectors. This enables scenarios like "use the subnets created by such and such XR, even though I don't know their Kubernetes names ahead of time". I'd prefer this generic reference approach to have similar functionality (especially if we deprecate the existing approach).

design/design-generic-references.md
negz: I mostly want to avoid a split in our API patterns, so I suggest we either:

1. Stick with the old pattern.
2. Form a plan to move our other object references over to the new pattern (e.g. in v2).

Perhaps we could do both; i.e. stick with the old pattern for now but formally capture that we want to use the new pattern everywhere in our v2 APIs?

design/design-generic-references.md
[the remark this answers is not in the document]
negz: Perhaps `patches`, for symmetry with Compositions?

design/design-generic-references.md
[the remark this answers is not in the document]
negz: Today the RBAC resources a provider can access is (iirc) a contract between the provider author and operator. i.e.:

1. The provider author defines the extra RBAC permissions it needs to operate.
2. The provider operator defines what RBAC permissions a provider may request and be granted.

This change almost raises the problem of what RBAC permissions a provider needs up to the configuration level. The provider author no longer knows what permissions it will need; it's possible a Composition author could decide that an RDSInstance should derive a field from some managed resource the provider-aws authors have never even heard of.

design/design-generic-references.md
```diff
@@ -0,0 +1,211 @@
...
+In order to enable such use cases, we need a generic version of the cross-resource
+references where you can specify all metadata required to find the referenced
+resource as well as details about how to extract the value you need and where to
+put it in the referencer object.
+
+## Proposal
+
+All managed resources will have a top-level spec field that lists the generic
+references. They will be resolved in order and failure of any will cause
+reconciliation to fail. Inspired from the [latest developments in API conventions](https://github.com/kubernetes/community/pull/5748)
```
negz, line 38: > They will be resolved in order and failure of any will cause reconciliation to fail.

If we fail on the second of three references would we still try to resolve the third? I'm wondering what the chances are that we'll run into something like the class of issues we addressed with https://github.com/crossplane/crossplane/pull/2180.

design/design-generic-references.md
```diff
@@ -0,0 +1,211 @@
...
+[`MustCreateObject`](https://github.com/crossplane/crossplane-runtime/blob/406fe0b/pkg/resource/resource.go#L145).
+After getting a `runtime.Object`, all controller-runtime utilities should be
+available to use.
+
+The resolver will make use of `fieldpath` library we're using for composition to
+get and set the values unknown types and once it is done, an update request will
+be issued. In line with cross-resource references, if the field on the referenced
+object, pointed by `toFieldPath`, already has a value then the resolver will
+skip it, meaning it will resolve only once and users will need to delete that
+value for resolver to set it again.
```
negz, line 100: We have a lot of feedback on https://github.com/crossplane/crossplane-runtime/issues/250 requesting that references be resolved on every reconcile, i.e. that the reference take precedence over the underlying field's value.

design/design-generic-references.md
```diff
@@ -0,0 +1,211 @@
...
+resource directly and may need some formatting. If we see a great need for that,
+we can move the transform machinery from Crossplane to Crossplane Runtime and
+reuse it.
+
+### More Limited RBAC
+
+Currently, providers have access to all `Secret`s in the cluster. While this will
+provide a nice UX for generic reference users, it may not be desirable for security
+minded folks since it'd cause giving managed resource permission mean giving access
+to any `Secret` in the cluster. See [#2384](https://github.com/crossplane/crossplane/issues/2384).
```
negz, line 116: I think we'll need an answer for this before we proceed with this design. If we keep allowing providers to access any secret then implementing this design would mean anyone with access to author any managed resource now has access to indirectly read a value from any secret.

design/design-generic-references.md
```diff
@@ -0,0 +1,211 @@
...
+# A generic reference to a namespaced Kubernetes resource.
+spec:
+  externalValues:
+  - fromObject:
+      version: v1
+      resource: configmaps
+      name: common-settings
+      namespace: crossplane-system
+      fieldPath: data.region
+    toFieldPath: spec.forProvider.region
```
negz, line 62: I wonder whether there would be any value in limiting the scope of `toFieldPath`, e.g. only allowing it to patch inside of `spec.forProvider`?

design/design-generic-references.md
```diff
@@ -0,0 +1,211 @@
...
+We could certainly have gone with this approach as well but having everything at
+the same level increases the risk of having to break the API to add new features.
+Although it's probably safe that we will not see many new types of sources be
+added, because a lot of use cases are already covered by composition patch types,
+it could be desirable to have multiple source objects to construct a single value
+for example and that'd be a new type. So, we're taking the safer approach of
+letting the API be more open-ended for future developments and pay a small UX
+fee in exchange.
+
+### Multiple Source References
```
negz, line 161: I agree we don't want to build this out the gate, but history shows that it's definitely wise to structure our API such that we could cleanly add support in future if we needed to. :)

negz (commented):
Thanks for getting this started @muvaf!

design/design-generic-references.md
```diff
@@ -0,0 +1,211 @@
...
+regarding multi-kind references, the API will look like the following:
+```yaml
+# A generic reference to another managed resource.
+spec:
+  externalValues:
+  - fromObject:
+      group: ec2.aws.crossplane.io
+      version: v1alpha1
+      resource: vpcs
+      name: main-vpc
```
morningspace, line 48: I'm wondering if this is always the case where the name here is resolvable at design phase, i.e. the time when people write this spec. Will it be possible if that's resolved at runtime, i.e. something generated dynamically, in which case, we might have ways to look for(or query) it elegantly.

@muvaf @negz

design/design-generic-references.md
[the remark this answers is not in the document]
muvaf: @negz Yeah `patches` are what they are but I was sort of hesitant to not make it too similar to composition. But I guess we don't have another word that matches is as good as `patches`. I'll run with that.

design/design-generic-references.md
muvaf: Sticking with the old pattern makes me feel better. Opened https://github.com/crossplane/crossplane/issues/2450

design/design-generic-references.md
muvaf: Though I agree having to add strongly typed references has been a challenge for writing/generating new managed resources, I actually lean towards keeping the existing references as first class. Because its API is much simpler for the cases where we know all the information we need during build time, i.e. references to other managed resources in the same provider, which is the case most of the time in my experience. Though this may change in the future if generated providers choose not to implement strongly-typed references for some reason, so we can come back and change the statement to make generic patches first class in such cases.

design/design-generic-references.md
muvaf: Do you have a specific case in mind where users would like to have that limitation?

design/design-generic-references.md
muvaf: Yeah I think we get that flexibility by putting the source under `fromObject` field instead of having it at the top. This way source fields can be added, only `toFieldPath` would be static which has been fine so far in composition.

design/design-generic-references.md
muvaf: @negz had the same concern, requesting to have a label selector which is something I'm adding to the latest revision.

design/design-generic-references.md
muvaf: Good idea! Added `matchLabels` and made `name` optional. Had a bit of back and forth with having `ref`/`selector` under `fromObject` but feels like all fields under `fromObject` are in the same class of reducing the set of possibilities for the selection hence put them directly there but not feeling strongly about either way.

design/design-generic-references.md
muvaf: By answer, do you mean https://github.com/crossplane/crossplane/issues/2384 should block the implementation of this design?

design/design-generic-references.md
```diff
@@ -0,0 +1,217 @@
...
+      apiVersion: ec2.aws.crossplane.io/v1alpha1
+      kind: VPC
+      name: main-vpc # optional, matchLabels can be used instead.
+      fieldPath: spec.forProvider.cidrBlock
+    toFieldPath: spec.cidrBlock
+```
+```yaml
+# A generic selector to another managed resource.
+spec:
+  patches:
```
morningspace, line 54: Regarding the naming, I saw there was a discussion in this PR before. I am curious whether it could be made a bit more generic so this design can cover https://github.com/crossplane/crossplane/issues/2072#issuecomment-764001935 as well.

In that scenario, instead of patching, what people want is to defer the managed resource handing till its dependent resources are resolvable. I like the idea that @negz suggested in the above comment for https://github.com/crossplane/crossplane/issues/2072 since it can use the same solution to address two different scenarios, patching vs. dependency check.

With that, I guess using patching in the above scenario would be a bit misleading. May be `depends` is better? Also, both `fieldPath` and `toFieldPath` could be optional. When `toFieldPath` is omitted, it does not patch anything. Instead, it will try to resolve the dependent resource(s) when `fieldPath` is omitted, or the `fieldPath` value(s) when `fieldPath` is specified.

Does it make sense?

@muvaf

design/design-generic-references.md
muvaf: @morningspace just added more details regarding the selector.

design/design-generic-references.md
muvaf: It's a bit different than #2180 due to two-way dependency. I'd expect relying on eventual consistency in this case would cover most cases already, so I say we start with this limitation and look for community feedback on whether we need a policy here as well. If so, the addition would be non-breaking.

design/design-generic-references.md
muvaf: @negz I kind of want either both (CRR and generic refs) or none of the mechanism to resolve at every reconcile so that the resolver behavior is consistent. Do you think both should resolve every time?

design/design-generic-references.md
```diff
@@ -0,0 +1,222 @@
...
+references. They will be resolved in order and failure of any will cause
+reconciliation to fail. Inspired from the [latest developments in API conventions](https://github.com/kubernetes/community/pull/5748)
+regarding multi-kind references, the API will look like the following:
+```yaml
+# A generic reference to another managed resource.
+spec:
+  patches:
+  - fromObject:
+      apiVersion: ec2.aws.crossplane.io/v1alpha1
+      kind: VPC
```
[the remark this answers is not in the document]
muvaf: That [was](https://github.com/crossplane/crossplane/pull/2385/commits/dc273491b0fc53f7081cf1ec85699cb1c210f982) the case initially, see discussion [here](https://github.com/crossplane/crossplane/pull/2385#discussion_r654065773).

design/design-generic-references.md
muvaf: > I agree that existing references have a simpler API but I would also prefer to have one way of doing referencing and deprecate the old way unless we are losing some functionality. 

In terms of functionality, we need to have more advanced transforms to cover some of the referencing mechanisms. For example, when resource-a requires ARN of resource-b, that ARN is not always available and you have to construct it. We do these constructions in Go today with the strongly typed references and users need to give only the name of the target resource. When you'd like to do the same with generic references, you need to give bunch of identification information as well as a set of transforms, so it gets pretty ugly in terms of UX and we can keep that from happening in build time. I think for users the flow will be that they will look at CRD schema and likely see the ref/selector field, if not, they know they can patch anything on the resource and will do that. But if they have to write generic ref patch for 7 fields on the same CR, which could be base template in the composition, the error surface increases and you get to know it after applying

negz: I remain convinced that we should not maintain multiple different ways of cross-resource references. As @turkenh says, having multiple ways to do one thing makes things complicated for our users. FWIW I also prefer the current reference implementation, which I feel is a more intuitive API than what we're likely to land on here. The issue is how manual that API is to support.

Is it possible for us to keep working on this design until we have one that we feel it can handle all the cases covered by the current XRR implementation in a reasonably elegant fashion?

design/design-generic-references.md
```diff
@@ -0,0 +1,222 @@
...
+
+All managed resources will have a top-level spec field that lists the generic
+references. They will be resolved in order and failure of any will cause
+reconciliation to fail. Inspired from the [latest developments in API conventions](https://github.com/kubernetes/community/pull/5748)
+regarding multi-kind references, the API will look like the following:
+```yaml
+# A generic reference to another managed resource.
+spec:
+  patches:
+  - fromObject:
```
negz, line 44:
```suggestion
  - from:
```

Would just `from` make sense? `fromObject` is correct, but we don't typically use the term "object" much in the Crossplane world (we're kind of committed to conflating "resource" to mean both "API object" and "API resource").

design/design-generic-references.md
```diff
@@ -0,0 +1,222 @@
...
+# A generic selector to another managed resource.
+spec:
+  patches:
+  - fromObject:
+      apiVersion: ec2.aws.crossplane.io/v1alpha1
+      kind: VPC
+      # selector resolver will populate name and namespace fields.
+      matchLabels:
+        class: main-vpc
+      fieldPath: spec.forProvider.cidrBlock
```
negz, line 77: Hrm, typically `<T>Selector` and `<T>Reference` fields are distinct. The API conventions used to recommend this (stating that `<T>Reference` fields should be a subset of the `ObjectReference` type), but that requirement was removed recently. How do we feel about conjoining the two here?

design/design-generic-references.md
negz: > I kind of want either both (CRR and generic refs) or none of the mechanism to resolve at every reconcile so that the resolver behavior is consistent.

I agree. Though that said I'm also arguing for there to only be one reference implementation.

> Do you think both should resolve every time?

I think it's a tradeoff. :) I've made my case over in https://github.com/crossplane/crossplane-runtime/issues/250 for the current logic (which I believe to be the common way selectors work upstream) but I can also see that I'm in the minority and that users are finding the current implementation to be unintuitive.

design/design-generic-references.md
negz: Right - or should at least proceed in parallel. i.e. I don't think we can really treat solving this as a "future consideration".

design/design-generic-references.md
muvaf: > Is it possible for us to keep working on this design until we have one that we feel it can handle all the cases covered by the current XRR implementation in a reasonably elegant fashion?

We can keep discussing the API shape but I'm not sure how much simplification we can ever get. Once the identification happens in runtime, the search space is `all kinds * all resources * all fields * field transformation` and we need to ask for information for each of those variables whereas current XRR search space is only `all resources` because kind and field and field transformations needed are known in build time, which makes it very easy to use.

> FWIW I also prefer the current reference implementation, which I feel is a more intuitive API than what we're likely to land on here. The issue is how manual that API is to support.

The manual support of that API could actually be automated with the code generation as long as this identification information that is to be written in build time is provided to the generator by the developer. In fact, I had started [working on that for ACK](https://github.com/aws-controllers-k8s/code-generator/pull/11) a while ago but postponed due to other reasons related to how ACK type generation works but I believe it's doable.

In fact, while our initial use case is generic referencing I believe we should think of this mechanism something similar to composition patches rather than a new referencing mechanism. Even though XRR was our initial use case, there is https://github.com/crossplane/crossplane/issues/2099 and possibly other use cases like implicit dependency management between managed resources that this feature will solve. I can see for example new types of patches here, like `fromVault` or `fromHTTP` that's used to fetch information from other sources and patch the resource.

design/design-generic-references.md
muvaf: The naming depends on the main thing that the mechanism is trying to achieve, right? In that sense, its main goal is to be able to gather information from other sources and put it as value to a field, i.e. patching. It provides a dependency graph as well because it's blocking but I'm not sure that's a big enough goal of this design that it should affect what it's called.

muvaf: > both fieldPath and toFieldPath could be optional. When toFieldPath is omitted, it does not patch anything. Instead, it will try to resolve the dependent resource(s) when fieldPath is omitted, or the fieldPath value(s) when fieldPath is specified.

You can achieve this by getting `metadata.uid` of the source and patching an arbitrary label/annotation of the resource. I'd like to look for more feedback from the community to make it a first citizen use case that we should account for.

design/design-generic-references.md
muvaf: I feel like `from` covers too much ground that may keep us adding different types of sources. Maybe `fromRef` **and** `fromSelector` which could also address [this discussion](https://github.com/crossplane/crossplane/pull/2385#discussion_r679586587) as well? cc @turkenh @hasheddan

design/design-generic-references.md
morningspace: > I'm not sure that's a big enough goal of this design that it should affect what it's called.
> I'd like to look for more feedback from the community to make it a first citizen use case that we should account for.

Thanks @muvaf, yeah, I can understand your point. The thing is, if we decide to leave this door open to wait for feedback, and when we see more requests on this, we may have to deprecate something and use another more appropriate term.

On the other hand, if we use a more generic term here, `dependencies`, `references` or something better, even we do not want to support the "creation deferral" case at the moment, this will give us a bit more flexible to make turnaround.

> You can achieve this by getting metadata.uid of the source and patching an arbitrary label/annotation of the resource.
I can see that works. It's more like a workaround for now :-D

BTW: I was trying to experiment "creation deferral" using provider-kubernetes by following the spec defined in this proposal, hopefully I can share more on this later based on my findings.

muvaf:
Even though figuring out which field is referencer is still a manual process, we have an easy way to generate the reference resolvers now with https://github.com/crossplane/crossplane-tools/pull/35 and there is a possibility that we might merge TF-based resources into native providers so that there would be single provider repo with some resources TF-based and some being native. These things eliminate the main needs for this feature.

There are other use cases like patching from common data sources, but for now I'm thinking of putting this on hold until we figure out what we want to do for references and maybe come back with a design more optimized for other use cases if we think referencing shouldn't be in the scope of this design.

design/design-generic-references.md
```diff
@@ -0,0 +1,222 @@
...
+  field in Route53 could get populated by a URL of an S3 bucket or access endpoint
+  of an RDS Instance.
+* Instead of hard-coding region for all managed resources, you may want to refer
+  to a `ConfigMap` or a `Secret` and let the managed resource take that
+  information from there.
+* In Composition, you can use values of the fields only between the children of
+  the same composite instance, but there are cases where you'd like to use a
+  separate resource that is not bound to any composite instance to store the
+  information and serve all composites.
+
```
AaronME, line 28: Wanted to offer another use-case that we've run into:

* Enabling a Workload Identity on a GKE Cluster requires supplying a value which contains the Project ID which hosts the cluster.  Because compositions are intended to be re-used, this requires passing in the Project ID at render time. It _is_ possible to add this as a parameter to the Composition, but this is information that a Claim Author is not expected to manage or know. A Platform Operator would want to supply this data " behind-the-scenes." In this case, allowing Composition Authors to source a `ConfigMap` or `Secret` outside the Composition would allow Platform Operators to add manage these values without Claim Authors needing to take on extra context.

sivarajp:
Hi, Does the cross referencing takes consideration of the status of the cross referencing object? I.e will it wait for ready status. Before fetching the value

negz:
I'm going to close this PR for the time being - there's still interest in this feature but we've paused work for the time being.

PR #2401, opened 2021-06-25
replace kingpin with kong cli framework

### LINKED ISSUE #1953: Use a consistent CLI framework

When the new Crossplane CLI was introduced we chose to use [kong](https://github.com/alecthomas/kong), which is the successor to [kingpin](https://github.com/alecthomas/kingpin). However, core Crossplane and the RBAC managed still use kingpin (as well as all providers). 

Using a consistent CLI framework requires less contextual knowledge for maintainers and new contributors, and kingpin is now in maintenance mode. We should move Crossplane and the RBAC manager to using kong. We could also consider using kong in each of the providers, but that is less of a priority.

psinghal20:

This PR replaces the kingpin CLI framework with its successor Kong. 

Logically there is no change in the CLI structure. One thing to note though, I have replaced the `zl.WithName(<NAME>)` to `log.WithValues("CmdName", "<NAME>")`. This allowed binding the single common logger to Kong and use that for each command. 

Fixes #1953 

I have:

Tested manually.

### DIFF
--- a/cmd/crossplane/core/core.go
+++ b/cmd/crossplane/core/core.go
@@ -37,7 +37,9 @@ type Command struct {
 	Init  initCommand  `cmd:"" help:"Make cluster ready for Crossplane controllers."`
 }
 
-// Run runs the core command
+// Run is the no-op method required for kong call tree
+// Kong requires each node in the calling path to have associated
+// Run method.
 func (c *Command) Run() error {
 	return nil
 }
@@ -52,7 +54,6 @@ type startCommand struct {
 
 // Run core Crossplane controllers.
 func (c *startCommand) Run(s *runtime.Scheme, log logging.Logger) error {
-	log.WithValues("CmdName", "core start")
 	cfg, err := ctrl.GetConfig()
 	if err != nil {
 		return errors.Wrap(err, "Cannot get config")
--- a/cmd/crossplane/core/core.go
+++ b/cmd/crossplane/core/core.go
@@ -19,10 +19,8 @@ package core
 import (
 	"time"
 
-	"github.com/google/go-containerregistry/pkg/name"
 	"github.com/pkg/errors"
 	"github.com/spf13/afero"
-	"gopkg.in/alecthomas/kingpin.v2"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 
@@ -33,34 +31,28 @@ import (
 	"github.com/crossplane/crossplane/internal/xpkg"
 )
 
-// Command configuration for the core Crossplane controllers.
+// Command runs the core crossplane controllers
 type Command struct {
-	Name           string
-	Namespace      string
-	Registry       string
-	CacheDir       string
-	LeaderElection bool
-	Sync           time.Duration
+	Start startCommand `cmd:"" help:"Start Crossplane controllers."`
+	Init  initCommand  `cmd:"" help:"Make cluster ready for Crossplane controllers."`
 }
 
-// FromKingpin produces the core Crossplane command from a Kingpin command.
-func FromKingpin(cmd *kingpin.CmdClause) (*Command, *InitCommand) {
-	startCmd := cmd.Command("start", "Start Crossplane controllers.")
-	c := &Command{Name: startCmd.FullCommand()}
-	cmd.Flag("namespace", "Namespace used to unpack and run packages.").Short('n').Default("crossplane-system").OverrideDefaultFromEnvar("POD_NAMESPACE").StringVar(&c.Namespace)
-	cmd.Flag("registry", "Default registry used to fetch packages when not specified in tag.").Short('r').Default(name.DefaultRegistry).Envar("REGISTRY").StringVar(&c.Registry)
-	cmd.Flag("cache-dir", "Directory used for caching package images.").Short('c').Default("/cache").OverrideDefaultFromEnvar("CACHE_DIR").StringVar(&c.CacheDir)
-	cmd.Flag("sync", "Controller manager sync period duration such as 300ms, 1.5h or 2h45m").Short('s').Default("1h").DurationVar(&c.Sync)
-	cmd.Flag("leader-election", "Use leader election for the conroller manager.").Short('l').Default("false").OverrideDefaultFromEnvar("LEADER_ELECTION").BoolVar(&c.LeaderElection)
-	initCmd := cmd.Command("init", "Make cluster ready for Crossplane controllers.")
-	init := &InitCommand{Name: initCmd.FullCommand()}
-	initCmd.Flag("provider", "Pre-install a Provider by giving its image URI. This argument can be repeated.").StringsVar(&init.Providers)
-	initCmd.Flag("configuration", "Pre-install a Configuration by giving its image URI. This argument can be repeated.").StringsVar(&init.Configurations)
-	return c, init
+// Run runs the core command
+func (c *Command) Run() error {
+	return nil
+}
+
+type startCommand struct {
+	Namespace      string        `short:"n" help:"Namespace used to unpack and run packages." default:"crossplane-system" env:"POD_NAMESPACE"`
+	CacheDir       string        `short:"c" help:"Directory used for caching package images." default:"/cache" env:"CACHE_DIR"`
+	LeaderElection bool          `short:"l" help:"Use leader election for the conroller manager." default:"false" env:"LEADER_ELECTION"`
+	Registry       string        `short:"r" help:"Default registry used to fetch packages when not specified in tag." env:"REGISTRY"`
+	Sync           time.Duration `short:"s" help:"Controller manager sync period duration such as 300ms, 1.5h or 2h45m" default:"1h"`
 }
 
 // Run core Crossplane controllers.
-func (c *Command) Run(s *runtime.Scheme, log logging.Logger) error {
+func (c *startCommand) Run(s *runtime.Scheme, log logging.Logger) error {
+	log.WithValues("CmdName", "core start")
 	cfg, err := ctrl.GetConfig()
 	if err != nil {
 		return errors.Wrap(err, "Cannot get config")
--- a/cmd/crossplane/core/core.go
+++ b/cmd/crossplane/core/core.go
@@ -40,7 +40,7 @@ type Command struct {
 }
 
 // KongVars represent the kong variables associated with the CLI parser
-// required for the RBAC enum interpolation.
+// required for the Registry default variable interpolation.
 var KongVars = kong.Vars{
 	"default_registry": name.DefaultRegistry,
 }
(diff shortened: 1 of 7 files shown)


cmd/crossplane/core/core.go
```diff
@@ -32,32 +31,27 @@ import (
...
-	cmd.Flag("sync", "Controller manager sync period duration such as 300ms, 1.5h or 2h45m").Short('s').Default("1h").DurationVar(&c.Sync)
-	cmd.Flag("leader-election", "Use leader election for the conroller manager.").Short('l').Default("false").OverrideDefaultFromEnvar("LEADER_ELECTION").BoolVar(&c.LeaderElection)
-	initCmd := cmd.Command("init", "Make cluster ready for Crossplane controllers.")
-	init := &InitCommand{Name: initCmd.FullCommand()}
-	initCmd.Flag("provider", "Pre-install a Provider by giving its image URI. This argument can be repeated.").StringsVar(&init.Providers)
-	initCmd.Flag("configuration", "Pre-install a Configuration by giving its image URI. This argument can be repeated.").StringsVar(&init.Configurations)
-	return c, init
+// Run runs the core command
+func (c *Command) Run() error {
+	return nil
```
negz, line 42: Is this no-op function typical for kingpin? It might be worth adding a comment explaining why it's a no-op.

cmd/crossplane/main.go
```diff
@@ -37,48 +36,57 @@ import (
...
+	)
+	ctx.FatalIfErrorf(corev1.AddToScheme(s), "cannot add core v1 Kubernetes API types to scheme")
+	ctx.FatalIfErrorf(appsv1.AddToScheme(s), "cannot add apps v1 Kubernetes API types to scheme")
+	ctx.FatalIfErrorf(rbacv1.AddToScheme(s), "cannot add rbac v1 Kubernetes API types to scheme")
+	ctx.FatalIfErrorf(coordinationv1.AddToScheme(s), "cannot add coordination v1 Kubernetes API types to scheme")
+	ctx.FatalIfErrorf(extv1.AddToScheme(s), "cannot add apiextensions v1 Kubernetes API types to scheme")
+	ctx.FatalIfErrorf(extv1beta1.AddToScheme(s), "cannot add apiextensions v1beta1 Kubernetes API types to scheme")
+	ctx.FatalIfErrorf(apis.AddToScheme(s), "cannot add Crossplane API types to scheme")
+	err := ctx.Run(s)
+	ctx.FatalIfErrorf(err)
```
negz, line 91:
```suggestion
	ctx.FatalIfErrorf(ctx.Run(s))
```

Nit: I'd prefer to inline this.

cmd/crossplane/main.go
```diff
@@ -37,48 +36,57 @@ import (
...
-		kingpin.FatalIfError(r.Run(s, logging.NewLogrLogger(zl.WithName("rbac"))), "cannot run RBAC manager")
-	case ri.Name:
-		kingpin.FatalIfError(ri.Run(s, logging.NewLogrLogger(zl.WithName("rbac init"))), "cannot initialize RBAC manager")
-	default:
-		kingpin.FatalUsage("unknown command %s", cmd)
-	}
+	ctx := kong.Parse(&cli,
+		kong.Name("crossplane"),
+		kong.Description("An open source multicloud control plane."),
+		kong.BindTo(logging.NewLogrLogger(zl), (*logging.Logger)(nil)),
```
negz, line 71: This `BindTo` pattern looks kind of arcane. Could we include a comment explaining what's happening?

cmd/crossplane/main.go
```diff
@@ -37,48 +36,57 @@ import (
...
+		kong.UsageOnError(),
+		kong.Vars{
+			"rbac_manage_default_var": rbac.ManagementPolicyAll,
+			"rbac_manage_enum_var": strings.Join(
+				[]string{
+					rbac.ManagementPolicyAll,
+					rbac.ManagementPolicyBasic,
+				},
+				", "),
+		},
```
negz, line 81: Is there any way to attach these vars to the context closer to where they're used, in the RBAC start command?

cmd/crossplane/core/core.go
```diff
@@ -32,32 +31,27 @@ import (
...
+	Namespace      string        `short:"n" help:"Namespace used to unpack and run packages." default:"crossplane-system" env:"POD_NAMESPACE"`
+	CacheDir       string        `short:"c" help:"Directory used for caching package images." default:"/cache" env:"CACHE_DIR"`
+	LeaderElection bool          `short:"l" help:"Use leader election for the conroller manager." default:"false" env:"LEADER_ELECTION"`
+	Sync           time.Duration `short:"s" help:"Controller manager sync period duration such as 300ms, 1.5h or 2h45m" default:"1h"`
 }
 
 // Run core Crossplane controllers.
-func (c *Command) Run(s *runtime.Scheme, log logging.Logger) error {
+func (c *startCommand) Run(s *runtime.Scheme, log logging.Logger) error {
+	log.WithValues("CmdName", "core start")
```
negz, line 54: I think `log.WithValues` returns a logger that must be used subsequently, rather than modifying `log`, so I don't think this call is having any effect. That said, I'd prefer to keep using logger names rather than values. It would probably be okay for all loggers to just have a name like `crossplane` rather than including a separate one for init, rbac, etc etc.

cmd/crossplane/main.go
```diff
@@ -37,48 +36,57 @@ import (
...
+type debugFlag bool
+
+var cli struct {
+	Debug debugFlag `short:"d" help:"Print verbose logging statements."`
+
+	Core core.Command `cmd:"" help:"Start core Crossplane controllers." default:"1"`
+	Rbac rbac.Command `cmd:"" help:"Start Crossplane RBAC Manager controllers."`
+}
+
+func (d debugFlag) BeforeApply(ctx *kong.Context) error { // nolint:unparam
```
negz, line 48: Could you include a comment explaining why we need to disable the `unparam` linter here please.

cmd/crossplane/main.go
```diff
@@ -37,48 +36,57 @@ import (
 	"github.com/crossplane/crossplane/cmd/crossplane/rbac"
 )
 
+type debugFlag bool
```
negz, line 39: It's perhaps not so bad in this case but we typically try to avoid package scoped variables. Is there some way to avoid using one here?

cmd/crossplane/main.go
```diff
@@ -37,48 +36,57 @@ import (
...
-		app   = kingpin.New(filepath.Base(os.Args[0]), "An open source multicloud control plane.").DefaultEnvars()
-		debug = app.Flag("debug", "Run with debug logging.").Short('d').Bool()
-	)
-	c, ci := core.FromKingpin(app.Command("core", "Start core Crossplane controllers.").Default())
-	r, ri := rbac.FromKingpin(app.Command("rbac", "Start Crossplane RBAC Manager controllers."))
-	cmd := kingpin.MustParse(app.Parse(os.Args[1:]))
+	// // NOTE(negz): We must setup our logger after calling kingpin.MustParse in
+	// // order to ensure the debug flag has been parsed and set.
+	zl := zap.New(zap.UseDevMode(false))
+	ctrl.SetLogger(zl)
```
negz, line 59: This call was previously only made if the debug flag was true. Did you intend for it to be called now in all cases?

negz (commented):
Thanks for tackling this @psinghal20!

cmd/crossplane/core/core.go
psinghal20: Yes, Kong requires each node in the path of the command to have an associated Run method. I have added the same as part of a comment.

cmd/crossplane/main.go
psinghal20: The Vars are options to the parser and attach the vars to the parsed command. I have moved these vars to the RBAC package. Let me know if that would work.

cmd/crossplane/core/core.go
psinghal20: Yeah, missed that. I have these `withValues` calls and just added a common Crossplane name to the logger.

cmd/crossplane/main.go
psinghal20: No, I don't think so. I have removed this call from main, it should only be called in case of when the debug flag is passed, which is handled by the `BeforeApply` method. I removed your initial comments by mistake too, added them back with a new commit.

cmd/crossplane/main.go
psinghal20: Done.

cmd/crossplane/main.go
psinghal20: Hmm, the preferred way to add hooks such as this one with Kong is generally using such package scoped variables. I am not sure of any workaround for defining hooks.

cmd/crossplane/core/core.go
```diff
@@ -32,32 +31,28 @@ import (
...
+// Kong requires each node in the calling path to have associated
+// Run method.
+func (c *Command) Run() error {
+	return nil
+}
+
+type startCommand struct {
+	Namespace      string        `short:"n" help:"Namespace used to unpack and run packages." default:"crossplane-system" env:"POD_NAMESPACE"`
+	CacheDir       string        `short:"c" help:"Directory used for caching package images." default:"/cache" env:"CACHE_DIR"`
+	LeaderElection bool          `short:"l" help:"Use leader election for the conroller manager." default:"false" env:"LEADER_ELECTION"`
```
negz, line 50:
```suggestion
	LeaderElection bool          `short:"l" help:"Use leader election for the controller manager." default:"false" env:"LEADER_ELECTION"`
```

This typo already existed, but it would be great to fix it while we're in here.

negz (approved):
This looks good to me - pending the merge conflict. Would you mind adding a comment showing the new output of `--help` for the commands with these changes in place so I can make sure that looks good before we merge?

psinghal20:
Hi @negz, fixed the conflict and the typo. Adding the updated help messages below for your reference.
```
Usage: crossplane <command>

An open source multicloud control plane.

Flags:
  -h, --help     Show context-sensitive help.
  -d, --debug    Print verbose logging statements.

Commands:
  core start
    Start Crossplane controllers.

  core init
    Make cluster ready for Crossplane controllers.

  rbac start
    Start Crossplane RBAC controllers.

  rbac init
    Initialize RBAC Manager.

Run "crossplane <command> --help" for more information on a command.
```
```
Usage: crossplane core start

Start Crossplane controllers.

Flags:
  -h, --help                             Show context-sensitive help.
  -d, --debug                            Print verbose logging statements.

  -n, --namespace="crossplane-system"    Namespace used to unpack and run packages ($POD_NAMESPACE).
  -c, --cache-dir="/cache"               Directory used for caching package images ($CACHE_DIR).
  -l, --leader-election                  Use leader election for the controller manager ($LEADER_ELECTION).
  -r, --registry="index.docker.io"       Default registry used to fetch packages when not specified in tag ($REGISTRY).
  -s, --sync=1h                          Controller manager sync period duration such as 300ms, 1.5h or 2h45m
```
```
Usage: crossplane core init

Make cluster ready for Crossplane controllers.

Flags:
  -h, --help                                 Show context-sensitive help.
  -d, --debug                                Print verbose logging statements.

      --providers=PROVIDERS,...
      --configurations=CONFIGURATIONS,...
```
```
Usage: crossplane rbac start

Start Crossplane RBAC controllers.

Flags:
  -h, --help                           Show context-sensitive help.
  -d, --debug                          Print verbose logging statements.

      --provider-clusterrole=STRING    A ClusterRole enumerating the permissions provider packages may request.
  -l, --leader-election                Use leader election for the conroller manager ($LEADER_ELECTION).
  -s, --sync=1h                        Controller manager sync period duration such as 300ms, 1.5h or 2h45m
  -m, --manage="All"                   RBAC management policy.
```

negz:
Thank you @psinghal20!

PR #2409, opened 2021-06-29
Support merging instead of replacing in patches

### LINKED ISSUE #2211: Support merging, rather than replacing objects

We're seeing a common pattern emerging in `Compositions`:

```yaml
  patchSets:
  - name: Metadata
    patches:
 ```

This patchset will replicate the XR's labels and annotations to each of its composed resources. A potential issue with this approach is that the [`fieldpath` package](https://pkg.go.dev/github.com/crossplane/crossplane-runtime@v0.13.0/pkg/fieldpath) package replaces objects - it does not merge them. Crossplane can store some important information about each composed resource in their annotations - for example the `crossplane.io/external-name` annotation controls the actual name of a resource in an external system in cases where that name should not be derived from the managed resource's Kubernetes `metadata.name`. It's possible for a managed resource to accidentally orphan an external resource if this information is patched over.

Crossplane could support configurable merge semantics when patching objects, perhaps using something like https://github.com/imdario/mergo or a JSON merge patch (https://tools.ietf.org/html/rfc7386). Ideally we would:

* Maintain the current behaviour by default, and replace the entire destination object.
* Support optionally merging in only keys that don't exist in the destination object.
* Support optionally replacing any key that exists in the destination object if the same key exists in the source object.

### PR DESCRIPTION
**Pull request by ulucinar**

On top of the new `fieldpath.MergeOptions` type and the `resource.WithMergeOptions` `ApplyOption` constructor provided in https://github.com/crossplane/crossplane-runtime/pull/270,  this PR extends the existing patch policies with support for [merging rather than replacing objects](https://github.com/crossplane/crossplane/issues/2211). Here is an example snippet that specifies a patch on `metadata.annotations` with merge semantics:

```yaml
kind: Composition
...
spec:
  patchSets:
  - name: metadata
    patches:
      policy:
        mergeOptions:
          keepMapValues: true
      type: FromCompositeFieldPath
```

The existing behavior of replacing objects is preserved: If `mergeOptions` has zero-value, then the default behavior is to replace the object rather than merging them.

This PR depends on https://github.com/crossplane/crossplane-runtime/pull/270.

Fixes #2211

I have:

Please see the discussion in https://github.com/crossplane/crossplane-runtime/pull/270 for how this PR has manually been tests. I have also used the following manifests to manually test this PR:

```yaml
---
metadata:
spec:
  group: database.example.org
  names:
  claimNames:
    kind: PostgreSQLInstance
    plural: postgresqlinstances
  connectionSecretKeys:
    - username
    - password
    - endpoint
    - port
  versions:
  - name: v1alpha1
    served: true
    referenceable: true
    schema:
      openAPIV3Schema:
        type: object
        properties:
          spec:
            type: object
            properties:
              parameters:
                type: object
                properties:
                  vpcSecurityGroupIDSelector:
                    description: VPCSecurityGroupIDSelector selects references to VPCSecurityGroups used to set the VPCSecurityGroupIDs.
                    properties:
                      matchControllerRef:
                        description: MatchControllerRef ensures an object with the same controller reference as the selecting object is selected.
                        type: boolean
                      matchLabels:
                        additionalProperties:
                          type: string
                        description: MatchLabels ensures an object with matching labels is selected.
                        type: object
                    type: object
                  storageGB:
                    type: integer
                  dbSecurityGroups:
                    description: 'DBSecurityGroups is a list of DB security groups to associate with this DB instance. Default: The default DB security group for the database engine.'
                    items:
                      type: string
                    type: array
                required:
                  - storageGB
            required:
              - parameters


---
kind: Composition
metadata:
  name: composition-new
  labels:
    provider: aws
    guide: quickstart
    vpc: default
spec:
  compositeTypeRef:
  patchSets:
  - name: selector
    patches:
    - toFieldPath: spec.parameters.vpcSecurityGroupIDSelector
      fromFieldPath: spec.forProvider.vpcSecurityGroupIDSelector
      type: ToCompositeFieldPath
      policy:
        mergeOptions:
          keepMapValues: true
  - name: metadata
    patches:
      policy:
        mergeOptions:
          keepMapValues: true
      type: FromCompositeFieldPath
  - name: dbsecuritygroups
    patches:
    - fromFieldPath: spec.parameters.dbSecurityGroups
      toFieldPath: spec.forProvider.dbSecurityGroups
      policy:
        mergeOptions:
          appendSlice: true
  resources:
    - name: rdsinstance
      base:
        kind: RDSInstance
        spec:
          forProvider:
            region: us-east-1
            dbInstanceClass: db.t2.small
            masterUsername: masteruser
            engine: postgres
            engineVersion: "9.6"
            skipFinalSnapshotBeforeDeletion: true
            publiclyAccessible: true
            allocatedStorage: 20
      patches:
         - type: PatchSet
           patchSetName: selector
         - type: PatchSet
           patchSetName: metadata
         - type: PatchSet
           patchSetName: dbsecuritygroups
         - fromFieldPath: "spec.writeConnectionSecretToRef.namespace"
           toFieldPath: "spec.writeConnectionSecretToRef.namespace"
           transforms:
             - type: string
               string:
                 fmt: "%s-postgresql"
           toFieldPath: "spec.forProvider.allocatedStorage"
      connectionDetails:
        - fromConnectionSecretKey: port


---
kind: PostgreSQLInstance
metadata:
  name: my-db
  namespace: default
  annotations:
    a: a-from-claim
spec:
  parameters:
    dbSecurityGroups:
    - a-from-claim
    vpcSecurityGroupIDSelector:
      matchLabels:
        l1: l1-from-claim
    storageGB: 20
  compositionSelector:
    matchLabels:
      provider: aws
      vpc: default

```


apis/apiextensions/v1/composition_types.go
```diff
@@ -843,3 +830,20 @@ type CompositionList struct {
...
+	if err != nil {
+		return err
+	}
+
+	if err := paved.MergeValue(fieldPath, value, mo); err != nil {
+		return err
+	}
+
+	return runtime.DefaultUnstructuredConverter.FromUnstructured(paved.UnstructuredContent(), to)
+}
```
negz, line 849: Can this function live outside our API type definition files? i.e. Somewhere under `internal/`? As currently implemented it doesn't appear to actually have any dependency on the API types defined in `package v1`. I'd actually personally prefer for our API type definitions to be as close to plain structs without any logic as possible, but I haven't yet found a good way to do that.

Another option here would be to just keep this function private and duplicate it in `merge.go`.

internal/controller/apiextensions/composite/reconciler.go
```diff
@@ -465,7 +465,8 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		if !cd.rendered {
 			continue
 		}
-		if err := r.client.Apply(ctx, cd.resource, resource.MustBeControllableBy(cr.GetUID())); err != nil {
+		if err := r.client.Apply(ctx, cd.resource,
+			append(mergeOptions(tas), resource.MustBeControllableBy(cr.GetUID()))...); err != nil {
```
negz, line 469: Style nit (feel free to ignore) - personally I'd prefer the longer line here than wrapping like this, because wrapping makes it a bit harder to visually distinguish the wrapped argument to `Apply` from the start of the nested block.

apis/apiextensions/v1/composition_types.go
```diff
@@ -273,6 +276,7 @@ type PatchPolicy struct {
 	// +kubebuilder:validation:Enum=Optional;Required
 	// +optional
 	FromFieldPath *FromFieldPathPolicy `json:"fromFieldPath,omitempty"`
+	MergeOptions  *xpv1.MergeOptions   `json:"mergeOptions,omitempty"`
```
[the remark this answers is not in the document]
negz: Oh whoops - I failed to realise this was already nested nuder the `PatchPolicy` type haha. In that context I think `mergeOptions` makes sense.

internal/controller/apiextensions/composite/reconciler.go
```diff
@@ -558,3 +561,25 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 	cr.SetConditions(xpv1.Available())
 	return reconcile.Result{RequeueAfter: longWait}, errors.Wrap(r.client.Status().Update(ctx, cr), errUpdateStatus)
 }
+
+func filterToXRPatches(tas []TemplateAssociation) []v1.Patch {
+	filtered := make([]v1.Patch, 0, len(tas))
```
ulucinar, line 566: `len(tas)` is not an upper bound for the number of elements in the result to be returned, but linter asked me pre-allocate this array and I think it's better than initialing with 0 capacity.

internal/controller/apiextensions/composite/composed.go
```diff
@@ -344,8 +344,8 @@ func (r *APIDryRunRenderer) Render(ctx context.Context, cp resource.Composite, c
 	cd.SetNamespace(namespace)
 
 	onlyPatches := []v1.PatchType{v1.PatchTypeFromCompositeFieldPath, v1.PatchTypeCombineFromComposite}
-	for i, p := range t.Patches {
-		if err := p.Apply(cp, cd, onlyPatches...); err != nil {
+	for i := 0; i < len(t.Patches); i++ {
+		if err := t.Patches[i].Apply(cp, cd, onlyPatches...); err != nil {
```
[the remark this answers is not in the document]
ulucinar: I was using `Patch.Filtered` as a cache to store the filtering result computed during a call to `Patch.Apply`. If I remember correctly, `p.Apply` was storing the filtering result in the copied loop variable, not in the actual `Patch` object. So, I had switched to indexing `t.Patches` to have `Filtered` set on the actual object. I have reverted this because I'm no longer using a `Patch.Filtered` to cache filtering computation.

apis/apiextensions/v1/composition_types.go
```diff
@@ -253,6 +253,9 @@ type Patch struct {
 	// Policy configures the specifics of patching behaviour.
 	// +optional
 	Policy *PatchPolicy `json:"policy,omitempty"`
+
+	// Filtered holds whether this patch has been filtered
+	Filtered bool `json:"-"`
```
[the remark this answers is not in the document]
ulucinar: Thanks @negz.
Yes, that's correct: The purpose was to cache the filtering status of the `Patch` object computed in `Patch.Apply`, and to avoid producing superfluous merge options for filtered patches. 
I was not either comfortable with introducing a field that's not to be serialized, and as I mentioned in the above comment, I have now removed this field. Instead, the current version recomputes filtering status.

apis/apiextensions/v1/composition_types.go
ulucinar: I have duplicated it although, in my opinion, preventing a duplicate is a stronger incentive. Nowhere, except probably the `apiextensions/composite` package (another place this is used) looked suitable for me but I think it would not be suitable to import something from `apiextensions/composite` in `apiextensions/v1`. So I preferred duplication.

internal/controller/apiextensions/composite/reconciler.go
ulucinar: Done.

internal/controller/apiextensions/composite/composed.go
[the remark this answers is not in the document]
ulucinar: Ooops, it turns out that we were also caching the computation of the default value of `Patch.ToFieldPath` if it's not set. I did not duplicate it elsewhere, but reverted back to indexing of `t.Patches`.

internal/controller/apiextensions/composite/reconciler.go
```diff
@@ -558,3 +561,25 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+	for _, ta := range tas {
+		filtered = append(filtered, filterPatches(ta.Template.Patches,
+			v1.PatchTypeToCompositeFieldPath, v1.PatchTypeCombineToComposite)...)
```
ulucinar, line 569: We will need to maintain this as new patch types are added...

internal/controller/apiextensions/composite/reconciler.go
```diff
@@ -440,6 +441,8 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		cds[i] = composedRendered{
 			resource: cd,
 			rendered: rendered,
+			filteredPatches: filterPatches(ta.Template.Patches, v1.PatchTypeFromCompositeFieldPath,
```
ulucinar, line 444: We will need to maintain this as new patch types are added.

internal/controller/apiextensions/composite/reconciler.go
```diff
@@ -558,3 +561,25 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+		filtered = append(filtered, filterPatches(ta.Template.Patches,
+			v1.PatchTypeToCompositeFieldPath, v1.PatchTypeCombineToComposite)...)
+	}
+	return filtered
+}
+
+func filterPatches(pas []v1.Patch, onlyTypes ...v1.PatchType) []v1.Patch {
+	filtered := make([]v1.Patch, 0, len(pas))
+	for _, p := range pas {
+		for _, t := range onlyTypes {
```
ulucinar, line 577: We can also consider exporting `Patch.filterPatch`.

apis/apiextensions/v1/composition_types.go
negz: > I have duplicated it although, in my opinion, preventing a duplicate is a stronger incentive.

Interestingly this is one of many areas where Go takes a bit of an atypical approach - in fact "a little copying is better than a little dependency" is one of the "Go proverbs" per https://go-proverbs.github.io/ and https://www.youtube.com/watch?v=PAAkCSZUG1c&t=568s.

internal/controller/apiextensions/composite/reconciler.go
negz: Perhaps we should use "constant" package level variables with docstrings to help make it more obvious that we need to keep these up to date? e.g.:

```go
var (
	// Types of patches that are _from_ a composite resource to a composed resource.
	fromXRTypes = []v1.PatchType{...}

	// Types of patches that are from a composed resource _to_ a composite resource.
	toXRTypes = []v1.PatchType{...}
)
```

internal/controller/apiextensions/composite/reconciler.go
```diff
@@ -515,7 +518,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 	// We pass a deepcopy because the update method doesn't update status,
 	// but calling update resets any pending status changes.
 	updated := cr.DeepCopyObject().(client.Object)
-	if err := r.client.Update(ctx, updated); err != nil {
+	if err := r.client.Apply(ctx, updated, mergeOptions(filterToXRPatches(tas))...); err != nil {
```
negz, line 521: What are the implications of changing from `Update` to `Apply` (i.e. patch) here?

negz:
Thanks for bearing with me @ulucinar! I like extending the existing `composedRendered` type much more than the previous approach of using a 'hidden' field on the user-facing API type. In general most of my commentary is around naming and documentation - I must admit I've been finding it a little difficult to understand what is going on with this PR so I encourage "over documenting" even if it may seem obvious. :)

internal/controller/apiextensions/composite/reconciler.go
ulucinar: Done. Added `composed.patchTypesToXR` & `composed.patchTypesFromXR` functions to define XR-targeting & composed-targeting patch types, respectively.

internal/controller/apiextensions/composite/reconciler.go
```diff
@@ -344,8 +344,9 @@ type Reconciler struct {
 // composedRendered is a wrapper around a composed resource that tracks whether
 // it was successfully rendered or not.
 type composedRendered struct {
-	resource resource.Composed
-	rendered bool
+	resource        resource.Composed
+	rendered        bool
+	filteredPatches []v1.Patch
```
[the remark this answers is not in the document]
ulucinar: Renamed `composite.composedRendered` to `composite.composedRenderState` and renamed field `filteredPatches` as `appliedPatches`. Also updated the docstring as:
```
// composedRenderState is a wrapper around a composed resource that tracks whether
// it was successfully rendered or not, together with a list of patches defined
// on its template that have been applied (not filtered out).
```

internal/controller/apiextensions/composite/reconciler.go
ulucinar: Added the following comments for clarification:
```
	// call Apply so that we do not just replace fields on existing XR but
	// merge fields for which a merge configuration has been specified.
	// For fields for which a merge configuration does not exist, the
	// behavior will be a replace from updated.
```
The idea is we now have control over whether fields of the XR get replaced or merged according to the rules specified on a patch targeting them (XR-targeting patches, if any).

ulucinar:
> Thanks for bearing with me @ulucinar! I like extending the existing `composedRendered` type much more than the previous approach of using a 'hidden' field on the user-facing API type. In general most of my commentary is around naming and documentation - I must admit I've been finding it a little difficult to understand what is going on with this PR so I encourage "over documenting" even if it may seem obvious. :)

It's my pleasure @negz to have quality & detailed reviews, as always. I also like the PR better now. Thank you very much for your time.

internal/controller/apiextensions/composite/reconciler.go
negz: Thank you!

I think there is a small behaviour change here in that we will now patch rather than update, regardless of the supplied merge options. That is, if an XR field was unset we'd previously persist that but now won't. That said, I don't think this will matter in practice because what we're doing here is persisting the result of the above `r.composite.Render` calls, and I believe the patches applied by those calls are purely additive (i.e. they never unset fields).

negz (approved):

PR #2423, opened 2021-07-06
Preserve rendered annotations & labels on composed resources

### LINKED ISSUE #2416: Rendered Annotations on Managed Resources Get Lost if a Patch on the root Annotations Object is Specified
### What happened?

This is one of the three issues that surfaced when discussing #2211. 
As discussed [here](https://github.com/crossplane/crossplane/issues/2211#issuecomment-853889836), if a patch from XR to MR (managed resource) is defined on the root `metadata.annotations` object, and if the XR has a non-zero `metadata.annotations` map, then for instance, the `crossplane.io/composition-resource-name` annotation, which is added during the rendering of the composed from its template base, is lost. 

Conceptually this is similar to #2211, however, in my opinion, this is a bug in the XR reconciler when it invokes `APIDryRunRenderer.Render` to render a composed. `APIDryRunRenderer.Render` loses the `crossplane.io/composition-resource-name` annotation it adds [here](https://github.com/crossplane/crossplane/blob/589f8c337f516a4f3f8e47e73aeaeee8d0158ca7/internal/controller/apiextensions/composite/composed.go#L345) when applying the patches [here](https://github.com/crossplane/crossplane/blob/589f8c337f516a4f3f8e47e73aeaeee8d0158ca7/internal/controller/apiextensions/composite/composed.go#L356), if there exists a patch from XR to composed on `metadata.annotations` and if XR has non-zero annotations. 

I propose to unconditionally merge the paths `metadata.annotations` and `metadata.labels` disallowing overrides of already existing values when applying patches. This is different than adding a new patch specification API that will control how objects are replaced/merged. 

Another question is should we allow a patch to be defined on certain other paths like `metadata.name`, `metadata.namespace`. etc. that also get rendered at the same site? My proposal would be just to make such patch specifications ineffective, and even with an admission controller disallow them.

Without #2353, this problem is amplified because we get an empty superfluous external-name annotation on the XR that satisfies one of the preconditions of the issue reported here.

### How can we reproduce it?

As a result of a claim such as this:
```
kind: PostgreSQLInstance
metadata:
  annotations:
    crossplane.io/external-name: name-from-claim
...
```
and a selected composition such as:
```
kind: Composition
...
spec:
  patchSets:
  - name: Metadata
    patches:
      type: FromCompositeFieldPath
...
```
you get a composed such as this:
```
kind: RDSInstance
metadata:
  annotations:
    crossplane.io/external-name: name-from-claim
  generation: 1
...
```
where the `crossplane.io/composition-resource-name` annotation gets lost. Of course, the same problem exists for other rendered fields.

Crossplane version: 1.3

ulucinar:

As detailed in #2416, we would like to keep any rendered annotations or labels on composed resources when patches on `metada.annotations` or `metadata.labels` are applied. Notably, if there is a patch (from composition to composed) is defined on `metadata.annotations` in the composition:
```
...
  patchSets:
  - name: metadata
    patches:
      type: FromCompositeFieldPath
```

and the composite has non-zero `metadata.annotations`, we lose the rendered `crossplane.io/composition-resource-name: rdsinstance` annotation on the composed.

Fixes #2416 

I have:

With a composition specifying a patch on `metadata.annotations` (exemplified above) selected, a dynamically provisioned XR with the following annotations:
```
...
metadata:
  annotations:
    a: b
  labels:
    crossplane.io/claim-name: my-db
    crossplane.io/composite: my-db-p7vk9
...
```  
now renders the following composed resource:
```
...
metadata:
  annotations:
    a: b
    crossplane.io/composition-resource-name: rdsinstance
  labels:
    crossplane.io/claim-name: my-db
    crossplane.io/composite: my-db-p7vk9
...
```

Without the proposed changes, it renders a composed like the following:
```
...
metadata:
  annotations:
    a: b
  labels:
    crossplane.io/claim-name: my-db
    crossplane.io/composite: my-db-p7vk9
...
```

### DIFF
--- a/internal/controller/apiextensions/composite/composed.go
+++ b/internal/controller/apiextensions/composite/composed.go
@@ -35,6 +35,7 @@ import (
 	"github.com/crossplane/crossplane-runtime/pkg/reconciler/managed"
 	"github.com/crossplane/crossplane-runtime/pkg/resource"
 	"github.com/crossplane/crossplane-runtime/pkg/resource/unstructured/composed"
+
 	v1 "github.com/crossplane/crossplane/apis/apiextensions/v1"
 	"github.com/crossplane/crossplane/internal/xcrd"
 )
@@ -335,6 +336,14 @@ func (r *APIDryRunRenderer) Render(ctx context.Context, cp resource.Composite, c
 		return errors.New(errNamePrefix)
 	}
 
+	onlyPatches := []v1.PatchType{v1.PatchTypeFromCompositeFieldPath, v1.PatchTypeCombineFromComposite}
+	for i, p := range t.Patches {
+		if err := p.Apply(cp, cd, onlyPatches...); err != nil {
+			return errors.Wrapf(err, errFmtPatch, i)
+		}
+	}
+
+	// Fix(2416): composed labels and annotations should be rendered after patches are applied
 	meta.AddLabels(cd, map[string]string{
 		xcrd.LabelKeyNamePrefixForComposed: cp.GetLabels()[xcrd.LabelKeyNamePrefixForComposed],
 		xcrd.LabelKeyClaimName:             cp.GetLabels()[xcrd.LabelKeyClaimName],
@@ -351,13 +360,6 @@ func (r *APIDryRunRenderer) Render(ctx context.Context, cp resource.Composite, c
 	cd.SetGenerateName(cp.GetLabels()[xcrd.LabelKeyNamePrefixForComposed] + "-")
 	cd.SetName(name)
 	cd.SetNamespace(namespace)
-
-	onlyPatches := []v1.PatchType{v1.PatchTypeFromCompositeFieldPath, v1.PatchTypeCombineFromComposite}
-	for i, p := range t.Patches {
-		if err := p.Apply(cp, cd, onlyPatches...); err != nil {
-			return errors.Wrapf(err, errFmtPatch, i)
-		}
-	}
 
 	// We do this last to ensure that a Composition cannot influence owner (and
 	// especially controller) references.
--- a/internal/controller/apiextensions/composite/composed.go
+++ b/internal/controller/apiextensions/composite/composed.go
@@ -336,6 +336,13 @@ func (r *APIDryRunRenderer) Render(ctx context.Context, cp resource.Composite, c
 		return errors.New(errNamePrefix)
 	}
 
+	// Unmarshalling the template will overwrite any existing fields, so we must
+	// restore the existing name, if any. We also set generate name in case we
+	// haven't yet named this composed resource.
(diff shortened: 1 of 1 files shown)


internal/controller/apiextensions/composite/composed.go
```diff
@@ -352,13 +361,6 @@ func (r *APIDryRunRenderer) Render(ctx context.Context, cp resource.Composite, c
 	cd.SetName(name)
 	cd.SetNamespace(namespace)
```
hasheddan, line 362: @ulucinar would this change mean that it is now impossible to patch over the `metadata.name` and `metadata.namespace` values? If so, I think this may be a breaking change that we are not willing to make. The existing behavior is desirable in cases where you want to provision a resource like a `ProviderConfig` with an identifying name that is used to group all the resources in a composition: https://github.com/upbound/platform-ref-aws/blob/63e724fc104ca1147c61787acc8a2891aced0393/cluster/eks/composition.yaml#L180

ulucinar: Hi @hasheddan,
Yes it would prevent such use-cases, thanks for pointing to the example use-case.

ulucinar: With the second commit, we now allow such patches.

hasheddan (approved):
This seems reasonable to me -- we are essentially mandating that our opinionated labels always be present. I would like for @negz to sign off on this though as it is somewhat subjective and I believe he originally added this functionality.

internal/controller/apiextensions/composite/composed.go
```diff
@@ -359,6 +350,17 @@ func (r *APIDryRunRenderer) Render(ctx context.Context, cp resource.Composite, c
 		}
 	}
 
+	// Fix(2416): composed labels and annotations should be rendered after patches are applied
```
negz, line 353: This `Fix(NNNN)` practice is new to me. Do you think it's something we should adopt? Typically I've mostly trusted git to create the audit trail we need through commit messages, git blame, etc such that code can be traced back to a PR. We do specifically link to a PR or via its full URL in some cases when we feel that the reader would benefit from additional context.

negz (approved):

internal/controller/apiextensions/composite/composed.go
ulucinar: The idea is to use a concise format similar to GitHub's [autolinked references](https://docs.github.com/en/github/writing-on-github/working-with-advanced-formatting/autolinked-references-and-urls) to refer to a related issue (in case of a `Fix`, to refer to an issue reporting a bug) in order to add some more context on the source listing itself. I think it draws attention to the specific site and warns future maintainers about a potential pitfall on the site. But we can of course embed full URLs or just leave comments (+ git blame) to achieve the same goal.

hasheddan:
@ulucinar it looks like we will need manual backports to `release-1.1` and `release-1.2`, do you want to get those in motion?

PR #2437, opened 2021-07-14
A one-pager describing composition revisions

negz:

This PR proposes a design for "composition revisions", which were initially described in https://github.com/crossplane/crossplane/issues/1481. You can see a partial implementation of the proposed design in https://github.com/crossplane/crossplane/pull/2451.

I propose adding a new type to Crossplane - `CompositionRevision`. This proposal is intended to address the fact that today when you update a `Composition` all XRs that use that `Composition` will immediately be affected by the update. The idea behind the proposed `CompositionRevision` type is that XRs can optionally pin to a particular revision of a `Composition`, and update to a later revision when they’re ready to do so (or roll back to an older revision)

I have:

N/A


design/one-pager-composition-revisions.md
```diff
@@ -0,0 +1,104 @@
...
+  parameters:
+    storageGB: 20
+  compositionSelector:
+    matchLabels:
+      provider: gcp
+  compositionRef:
+    name: example-gcp
+  # A new optional field, typically set automatically to the latest revision of
+  # the chosen composition.
+  compositionRevisionRef:
```
OmerKahani, line 77: Why not adding it under `compositionRef`?

OmerKahani:
Started experimenting with crossplane today, and immediately thought about how to update the Composition. Looking forward for this issue :)

design/one-pager-composition-revisions.md
negz: Because `compositionRef` is a reference to a Composition, while this is a reference to a CompositionRevision; two different resources.

The Kubernetes [API conventions](https://github.com/kubernetes/community/blob/6042eed74ab108c617af1162b85f7fdb3f9c2c27/contributors/devel/sig-architecture/api-conventions.md#references-to-related-objects) until recently recommended that `Ref` fields be a subset of the `ObjectReference` type, which would also preclude nesting reference under this field. That guidance was recently removed.

OmerKahani: I missed it's a new object, thanks!

design/one-pager-composition-revisions.md
```diff
@@ -0,0 +1,300 @@
...
+  # Composition is deleted all of its CompositionRevisions are garbage
+  # collected.
+  - apiVersion: apiextensions.crossplane.io/v1
+    kind: Composition
+    name: example
+    controller: true
+spec:
+  # Each revision includes an integer revision number. This number increases
+  # monotonically as the associated composition is updated.
+  revision: 1
```
muvaf, line 128: Can we have this revision number in the name as well for easier browsing? i.e. `example-1-jfjn3` instead of `example-jfjn3`

muvaf (approved):
LGTM! I'd have preferred webhooks for creating revisions because it's completely action-based; reconciliation without user action is unnecessary but we don't have them yet. Maybe we can add a statement that says webhook would be preferred if they exist at the time of the implementation?

design/one-pager-composition-revisions.md
```diff
@@ -0,0 +1,300 @@
...
+are designed such that only one revision will be 'active' at any one time. For
+example there is only ever one active `ProviderRevision` for each Crossplane
+`Provider` resource. This means it's generally fairly safe and easy to garbage
+collect old revisions; you need only avoid garbage collecting the active one.
+
+The design proposed by this document requires a more nuanced approach to
+garbage collecting old revisions, because many revisions may be in use at one
+time. The policy would likely need to be "garbage collect revisions older than N
+if they're not in use", which would require the garbage collector to keep track
+of which revisions are in use across an arbitrary set of XR types. This document
```
hasheddan, line 263: I wonder if this reference counting could be handled using built-in functionality (namely, `finalizers`). It would require adding a finalizer for each XR that utilized a `CompositionRevision`, but it would make the deletion logic simpler as we would simply always try to delete a `CompositionRevision` that was outside of the GC window, and k8s would be responsible for making sure it stuck around. Also has the nice side-effect of exposing what XRs are blocking deletion of a given `CompositionRevision`.

design/one-pager-composition-revisions.md
```diff
@@ -0,0 +1,300 @@
...
+```
+
+This will be replaced with a call to a new `CompositionFetcher` interface, e.g.:
+
+```go
+// A CompositionFetcher fetches an appropriate Composition for the supplied
+// composite resource.
+type CompositionFetcher interface {
+  Fetch(ctx context.Context, cr resource.Composite) (*v1.Composition, error)
+}
```
hasheddan, line 212: I appreciate this detail in a one-pager :)

hasheddan (commented):
@muvaf makes an interesting point of using webhooks, though I believe either a controller triggered on updates or webhooks could result in a situation where we miss a revision (i.e. controller or webhook server is down), and any action-based response is moving away from declarative patterns. This may already be part of the plan, but I wonder if we could compute a digest for a composition and check it against the existing set of revisions. This would allow us to ensure that a revision exists for any given state of a `Composition` over time (with the exception that if two updates to a `Composition` happened during a single window where the controllers were down we could still miss the first).

muvaf:
> I believe either a controller triggered on updates or webhooks could result in a situation where we miss a revision (i.e. controller or webhook server is down)

If the webhook server is down, the change request on the composition would be rejected. We'll likely have validations there as well, so all problems would be apparent right away. But either way works good enough and don't affect the users too much.

> This may already be part of the plan, but I wonder if we could compute a digest for a composition and check it against the existing set of revisions. This would allow us to ensure that a revision exists for any given state of a Composition over time (with the exception that if two updates to a Composition happened during a single window where the controllers were down we could still miss the first).

Good point. The comparison could be a bit challenging with the raw templates. For example, a map in the raw template could have its order changed on the CR and theoretically no new revision should be issued but if the comparison is hash based, we'd issue a new one. Some details covering the comparison would be helpful, probably a YAML utility.

design/one-pager-composition-revisions.md
negz: That sounds like a good approach. Honestly we could probably avoid the random suffix if we just used the revision number. 🤔

design/one-pager-composition-revisions.md
negz: Is this pattern of using many finalizers common? It always seems like a bit of a misuse of finalizers to me when I hear it proposed - probably because I think of finalizers as being one-per-reconciler rather than one-per-related-object.

negz:
> This may already be part of the plan, but I wonder if we could compute a digest for a composition and check it against the existing set of revisions.

That's the plan - I didn't go into much detail about how we'd create revisions here. Perhaps I should.

> I'd have preferred webhooks for creating revisions because it's completely action-based

Hrm, webhooks doesn't seem like an intuitive fit to me but I suppose it would work. I wonder if other revision implementations use webhooks rather than a controller? Given this will be our second revision implementation (after package revisions) perhaps we can reassess whether all revisions should be created by webhooks once we've built out support.

negz:
> That's the plan

@hasheddan On reflection I'm thinking I will actually allow the creation of duplicate revisions - i.e. I won't use hashes to dedupe them.

Using hashes to dedupe makes sense on other contexts like package revisions because we can just mark an older revision 'active'. Composition revisions don't have that concept. I think it could be confusing that some updates to compositions create a revision while others don't (because they're identical to an older revision). Instead I'm thinking I might decorate revisions with a hash of their compositions spec. This way every update results in a new revision with a monotonically increasing revision number, but you can at least identify identical revisions by looking at the hashes (which could be a printer column).

design/one-pager-composition-revisions.md
negz: Updated the doc to reflect that names will just be `compname-revnum`, like `example-1`.

design/one-pager-composition-revisions.md
```diff
@@ -0,0 +1,308 @@
...
+  # updated automatically to reflect the latest available revision, or whether
+  # it must be updated manually. Valid values are Automatic or Manual. Defaults
+  # to Automatic.
+  compositionUpdatePolicy: Automatic
+```
+
+This design is backward compatible with Crossplane's current behaviour. An XR or
+XRC author may be ignorant of the new `CompositionRevision` type. The XR will
+default to `compositionUpdatePolicy: Automatic`, and thus always select the
+latest revision of the desired `Composition`. Only when the XR or XRC author
```
chlunde, line 190: will `Automatic` include any safe guards, or is that (currently) out of scope? For example,
* rate limiting the number of upgrades (this could be difficult without configuration - what is an acceptable number of upgrades per minute for a cluster with 30 XRCs, and one with 10 000?)
* requiring an upgrade to be ready before upgrading the next component (again, I guess this would be difficult, when do we know that the upgrade is done? and for the 10 000 XRC cluster, you probably want upgrades in parallell)

Or will safe guards be `Manual` an external component/process? One helpful thing for an external upgrade process would be to have a status field with the number of ready/unready/synced claims, but I guess this might be difficult to implement?

negz: I wasn't planning on building any new safeguards into the `Automatic` strategy. The idea (at least at first) is for `Automatic` to function similarly to the current implementation where there is an implicit "automatic" update of all XRs when a Composition is updated.

That said, the logic that updates the Composition used by each XR will be be built into the XR reconciler. Each reconciler has its own reconcile requeue rate limit, and iirc we have a global rate limit as well. So I'd expect `Automatic` updates to enjoy some safeguards that they did not have previously under this design.

design/one-pager-composition-revisions.md
hasheddan: @negz I'm not sure, it does feel somewhat unconventional, but based on the [description of the purpose of a finalizer](https://kubernetes.io/blog/2021/05/14/using-finalizers-to-control-deletion/#understanding-finalizers), it doesn't necessarily feel like a misuse:
> Finalizers are keys on resources that signal pre-delete operations. They control the garbage collection on resources, and are designed to alert controllers what cleanup operations to perform prior to removing a resource. However, they don’t necessarily name code that should be executed; finalizers on resources are basically just lists of keys much like annotations. Like annotations, they can be manipulated.

hasheddan (approved):
@negz looks good to me! Thanks for working in the section about computing the hash!

PR #2443, opened 2021-07-21
Add a page documenting Crossplane terminology

negz:

We don't currently have a single place to find details about common Crossplane concepts. I've added this under 'concepts' rather than 'reference' in the hope it might be more discoverable there, though I could go either way.

I have:

I've asked a few Upbounders to review this, and would like to have the community take a look before merging.

### DIFF
--- a/docs/concepts/terminology.md
+++ b/docs/concepts/terminology.md
@@ -1,0 +1,216 @@
+---
+title: Terminology
+toc: true
+weight: 51
+indent: true
+---
+
+# Terminology
+
+## A Note on Style
+
+Each type of Kubernetes resource has a ‘Pascal case’ name - i.e. a title case
+name with no spaces between each word. Examples include ‘DaemonSet’ and
+‘PersistentVolumeClaim’. Often these names are written using fixed width fonts
+to draw attention to the fact that they’re a concrete type of resource within
+the API - e.g. `PersistentVolumeClaim`.
+
+Crossplane follows this convention. We often use names like RDSInstance or
+CompositeResourceDefinition when discussing Crossplane types. Crossplane also
+has “classes of types” - i.e. concepts that aren’t a distinct type of API
+resource, but rather describe a group of conceptually similar types. For example
+there is no ManagedResource type in Crossplane - instead types like RDSInstance
+and GKECluster are said to be “a managed resource”.
+
+Use your discretion as to whether you use pascal case when writing about a
+distinct type - e.g. “RDS Instance” and “RDSInstance” are both fine. The pascal
+case form makes more sense in contexts like documentation where you’re referring
+to Crossplane’s RDSInstance managed resource rather than the general concept of
+“an RDS instance”. Avoid using Pascal case when talking about classes of types -
+i.e. always write “managed resource”, not “ManagedResource”. Each of the below
+terms clarify whether they correspond to a single type, or a class of types.
+
+### Why 'X'?
+
+You may notice that Crossplane uses “X” as shorthand for “Crossplane” and/or
+“Composite”. This is because some of our concepts - specifically Composite
+Resources (XRs) and Composite Resource Definitions (XRDs) are modelled on
+similar Kubernetes concepts - Custom Resources (CRs) and Custom Resource
+Definitions (CRDs). We chose to abbreviate to (e.g.) XRD instead of CRD to avoid
+confusion.
+
+## Crossplane Terms
+
+The below terms are commonly used in the Crossplane ecosystem.
+
+### Composition
+
+The term Composition has two related but distinct meanings.
+
+“Composition” refers broadly to the feature of Crossplane that allows teams to
+define their own opinionated platform APIs.
+
+“A Composition” or `Composition` (fixed width) refers to the key Crossplane API
+type that configures how Crossplane should compose resources into a higher level
+“composite resource”. A Composition tells Crossplane “when someone creates
+composite resource X, you should respond by creating resources Y and Z”.
+
+The latter use of Composition represents a distinct Crossplane API type so
+Pascal case and fixed width fonts are appropriate. We also tend to capitalise
+the former use, representing the feature in general, but fixed width fonts are
+not appropriate in that context.
+
+> Folks accustomed to Terraform might think of a Composition as a Terraform
+> module; the HCL code that describes how to take input variables and use them
+> to create resources in some cloud API. Folks accustomed to Helm might think of
+> a Composition as a Helm chart’s templates; the moustache templated YAML files
+> that describe how to take Helm chart values and render Kubernetes resources.
+
+### Composite Resource
+
+A “Composite Resource” or “XR” is an API type defined using Crossplane. A
+composite resource’s API type is arbitrary - dictated by the concept the author
+wishes to expose as an API, for example an “AcmeCoDB”. A common convention is
+for types to start with the word “Composite” - e.g. “CompositeAcmeCoDB”.
+
+We talk about Crossplane being a tool teams can use to define their own
+opinionated platform APIs. Those APIs are made up of composite resources; when
+you are interacting with an API that your platform team has defined, you’re
+interacting with composite resources.
+
+A composite resource can be thought of as the interface to a Composition. It
+provides the inputs a Composition uses to compose resources into a higher level
+concept. In fact, the composite resource _is_ the high level concept.
+
+The term “Composite Resource” refers to a class of types, so avoid using Pascal
+case - “Composite Resource” not CompositeResource. Use pascal case when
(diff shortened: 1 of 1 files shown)


docs/concepts/terminology.md
```diff
@@ -0,0 +1,215 @@
...
+Managed resources are granular, high fidelity Crossplane representations of a
+resource in an external system - i.e. resources that are managed by Crossplane.
+Managed resources are what Crossplane enables platform teams to compose into
+higher level composite resources, forming an opinionated platform API. They're
+the building blocks of Crossplane.
+
+You’ll often hear three related terms used in the Crossplane ecosystem; composed
+resource, managed resource, and external resource. While there are some subtle
+contextual differences, these all broadly refer to the same thing. Take an
+RDSInstance for example; it is a managed resource. A distinct resource within
```
hasheddan, line 161: nitpick:
```suggestion
RDSInstance for example; it is a managed resource - a distinct resource within
```

docs/concepts/terminology.md
```diff
@@ -0,0 +1,215 @@
...
+resources and claims, or support for new kinds of managed resources. There are
+two types of Crossplane package; configurations and providers.
+
+A package is not a distinct type in the Crossplane API, but rather a class of
+types. Therefore Pascal case is not appropriate.
+
+### Configuration
+
+A configuration extends Crossplane by installing conceptually related groups of
+XRDs and Compositions. Put otherwise, it configures the opinionated platform API
```
hasheddan, line 188: Wouldn't mind the addition of specifying dependencies in this definition, but I think that could also fall into a detail that is out of scope for this document.

docs/concepts/terminology.md
```diff
@@ -0,0 +1,215 @@
...
+that Crossplane exposes.
+
+A `Configuration` is a distinct type in the Crossplane API, therefore Pascal
+case and fixed width fonts are appropriate.
+
+### Provider
+
+A provider extends Crossplane by installing controllers for new kinds of managed
+resources. Providers typically group conceptually related managed resources; for
```
hasheddan, line 197: maybe? feel free to keep as is if this also is out of scope
```suggestion
resources. Providers typically group conceptually related managed resources and their controllers; for
```

hasheddan (approved):
A few comments but none of them blocking, thanks for adding this @negz! I think this will be very useful for folks :)

docs/concepts/terminology.md
```diff
@@ -0,0 +1,215 @@
...
+case - “Composite Resource” not CompositeResource. Use pascal case when
+referring to a distinct type of composite resource - e.g. a CompositeAcmeCoDB.
+
+> Folks accustomed to Terraform might think of a composite resource as a
+> `tfvars` file that supplies values for the variables a Terraform module uses
+> to create resources in some cloud API. Folks accustomed to Helm might think of
+> a composite resource as the `values.yaml` file that supplies inputs to a Helm
+> chart’s templates.
+
+### Composite Resource Claim
```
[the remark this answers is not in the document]
negz: Yeah, my preference is to keep this a little higher level. More about what and why rather than how.

docs/concepts/terminology.md
negz: I think I avoided this because mixing the earlier semicolon and the hypen felt... odd.

hasheddan: ha fair point :) I was just a little tripped up with what I think is an incomplete sentence

docs/concepts/terminology.md
negz: Controllers are mentioned above here - I think that's sufficient.

docs/concepts/terminology.md
negz: I like the idea of mentioning dependencies - done.

docs/concepts/terminology.md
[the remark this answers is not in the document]
rotemjac: @negz 
I think that it really helpful to specify the comparisons to Terraform and Helm like you did - thanks

docs/concepts/terminology.md
```diff
@@ -0,0 +1,216 @@
...
+resource, but rather describe a group of conceptually similar types. For example
+there is no ManagedResource type in Crossplane - instead types like RDSInstance
+and GKECluster are said to be “a managed resource”.
+
+Use your discretion as to whether you use pascal case when writing about a
+distinct type - e.g. “RDS Instance” and “RDSInstance” are both fine. The pascal
+case form makes more sense in contexts like documentation where you’re referring
+to Crossplane’s RDSInstance managed resource rather than the general concept of
+“an RDS instance”. Avoid using Pascal case when talking about classes of types -
+i.e. always write “managed resource”, not“ManagedResource”. Each of the below
```
thephred, line 30:
```suggestion
i.e. always write “managed resource”, not “ManagedResource”. Each of the below
```

docs/concepts/terminology.md
```diff
@@ -0,0 +1,216 @@
...
+You may notice that Crossplane uses “X” as shorthand for “Crossplane” and/or
+“Composite”. This is because some of our concepts - specifically Composite
+Resources (XRs) and Composite Resource Definitions (XRDs) are modelled on
+similar Kubernetes concepts - Custom Resources (CRs) and Custom Resource
+Definitions (CRDs). We chose to abbreviate to (e.g.) XRD instead of CRD to avoid
+confusion.
+
+## Crossplane Terms
+
+The below terms are commonly used in the Crossplane ecosystem.
```
thephred, line 44: This might read better.
```suggestion
The terms below are commonly used in the Crossplane ecosystem.
```

docs/concepts/terminology.md
```diff
@@ -0,0 +1,216 @@
...
+
+The latter use of Composition represents a distinct Crossplane API type so
+Pascal case and fixed width fonts are appropriate. We also tend to capitalise
+the former use, representing the feature in general, but fixed width fonts are
+not appropriate in that context.
+
+> Folks accustomed to Terraform might think of a Composition as a Terraform
+> module; the HCL code that describes how to take input variables and use them
+> to create resources in some cloud API. Folks accustomed to Helm might think of
+> a Composition as a Helm chart’s templates; the moustache templated YAML files
```
thephred, line 66:
```suggestion
> a Composition as a Helm chart’s templates; the mustache templated YAML files
```

docs/concepts/terminology.md
```diff
@@ -0,0 +1,216 @@
...
+
+We talk about Crossplane being a tool platform teams can use to offer
+opinionated platform APIs to the application teams they support. The platform
+team offers those APIs using claims. It helps to think of the claim as an
+application team’s interface to a composite resource. You could also think of
+claims as the public (app team) facing part of the opinionated platform API,
+while composite resources are the private (platform team) facing part.
+
+A common convention is for a claim to be of the same type as its corresponding
+composite resource, but without the "Composite" prefix. So an "AcmeCoDB" would
```
thephred, line 110: Maybe better?
```suggestion
A common convention is for a claim to have the same type name as its corresponding
composite resource, but without the "Composite" prefix. So an "AcmeCoDB" would
```

docs/concepts/terminology.md
```diff
@@ -0,0 +1,216 @@
...
+make up a platform API.
+
+XRDs are often conflated with composite resources (XRs) - try to avoid this.
+When someone uses the platform API to create infrastructure they’re not creating
+XRDs but rather creating composite resources (XRs). It may help to think of a
+composite resource as a database entry, while an XRD is a database schema. For
+those familiar with Kubernetes, the relationship is very similar to that between
+a Custom Resource Definition (CRD) and a Custom Resource (CR).
+
+A CompositeResourceDefinition is a distinct Crossplane API type, so Pascal case
```
thephred, line 142:
```suggestion
A `CompositeResourceDefinition` is a distinct Crossplane API type, so Pascal case
```

thephred (commented):
Some nit picks and suggestions. This looks great @negz!

muvaf (approved):
Love it. Thank you @negz !

docs/concepts/terminology.md
negz: Apparently this is a somewhat controversial point but it seems like both forms are valid so I'm going to be lazy and leave it. :) https://english.stackexchange.com/questions/74182/above-below-before-after-a-noun

docs/concepts/terminology.md
negz: https://en.wikipedia.org/wiki/Moustache 🇬🇧

docs/concepts/terminology.md
negz: Hrm, I want to avoid "type name" here since strictly it's the "kind" which makes up part of a type. I think I'll leave this as is.

PR #2451, opened 2021-07-29
Add support for composition revisions

negz:

Fixes https://github.com/crossplane/crossplane/issues/1481

This PR contains the implementation of the design proposed in https://github.com/crossplane/crossplane/pull/2437. It depends on https://github.com/crossplane/crossplane-runtime/pull/273.

I have:

I've tested manually that:

- That Composition works as it did previously without the feature flag.
- Updating revisions manually.
- Updating revisions automatically.

In more detail:

* I confirmed that Composition without revisions works as expected by running Crossplane locally without the new `--enable-alpha-composition-revisions` flag and applying the GCP getting started manifests. I then updated the `Composition` to add a new flag (`spec.forProvider.settings.storageAutoResize`) and observed that the composed `CloudSQLInstance` was updated accordingly soon thereafter.
* I confirmed that Composition with revisions and the `Automatic` update policy work as expected by running Crossplane with the new `--enable-alpha-composition-revisions` flag and applying the GCP getting started  and running the above test. I see that the `CompositePostgreSQLInstance` tracks the most current revision as I toggle the `storageAutoResize` field on the `Composition`, and that those changes propagate to the actual `CloudSQLInstance`.
* I confirmed that Composition with revisions and the `Manual` update policy work as expected by running Crossplane with the new `--enable-alpha-composition-revisions` flag and applying the GCP getting started  and running the above test. I see that the `CompositePostgreSQLInstance` does not track the most current revision as I toggle the `storageAutoResize` field on the `Composition`. When I update the XR's revision manually I see the changes propagate to the actual `CloudSQLInstance`.


negz:
Some further detail on testing:

* I confirmed that Composition without revisions works as expected by running Crossplane locally without the new `--enable-alpha-composition-revisions` flag and applying the GCP getting started manifests. I then updated the `Composition` to add a new flag (`spec.forProvider.settings.storageAutoResize`) and observed that the composed `CloudSQLInstance` was updated accordingly soon thereafter.

internal/controller/apiextensions/claim/configurator.go
```diff
@@ -86,14 +86,17 @@ func ConfigureComposite(_ context.Context, cm resource.CompositeClaim, cp resour
...
+	// first we must filter out any well-known fields that are unique to
+	// claims. We do this by:
+	// 1. Grabbing a map whose keys represent all well-known claim fields.
+	// 2. Deleting any well-known fields that we want to propagate.
+	// 3. Using the resulting map keys to filter the claim's spec.
+	wellKnownClaimFields := xcrd.CompositeResourceClaimSpecProps()
+	for _, field := range xcrd.PropagateClaimSpecProps {
+		delete(wellKnownClaimFields, field)
+	}
+	claimSpecFilter := xcrd.GetPropFields(wellKnownClaimFields)
```
negz, line 99: There's no logic change here - I just got stuck on this code for a bit and wanted to try make it a bit more readable.

internal/controller/apiextensions/composite/api.go
```diff
@@ -113,6 +119,107 @@ func (a *APIFilteredSecretPublisher) UnpublishConnection(_ context.Context, _ re
...
+	for i := range rl.Items {
+		if !metav1.IsControlledBy(&rl.Items[i], comp) {
+			continue
+		}
+
+		// We use the hash, not the revision number, to determine which
+		// revision is current. This is because the Composition may have
+		// been reverted to an older state. When this happens we don't
+		// create a new CompositionRevision; rather the prior revision
+		// becomes effectively 'current'.
```
negz, line 202: I think it would be useful to mark the current revision as such. I'd like to do that in a future PR though - I'll open an issue to track.

internal/controller/apiextensions/composite/composed.go
```diff
@@ -181,15 +181,15 @@ func AssociateByOrder(t []v1.ComposedTemplate, r []corev1.ObjectReference) []Tem
 
 // A CompositionTemplateAssociator returns an array of template associations.
 type CompositionTemplateAssociator interface {
-	AssociateTemplates(context.Context, resource.Composite, *v1.Composition) ([]TemplateAssociation, error)
+	AssociateTemplates(context.Context, resource.Composite, []v1.ComposedTemplate) ([]TemplateAssociation, error)
```
negz, line 184: This refactoring to operate on `[]v1.ComposedTemplate` is spurious. I had originally thought this would help with this PR because both compositions and revisions have an identical `ComposedTemplate` type. I neglected to realise that the `CompositionRevision` type is different though, due to being in a different package.

I decided to leave the refactoring in because these types don't actually operate on the whole `*v1.Composition` - just the templates. Ideally I'd break it out into its own commit but it got a little bit mixed up with the actual `CompositionRevision` implementation so I'd prefer to just leave it as is if no one minds too much. 😬

internal/controller/apiextensions/composite/reconciler.go
```diff
@@ -379,27 +401,15 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+		log.Debug(errFetchComp, "error", err)
 		r.record.Event(cr, event.Warning(reasonCompose, err))
 		return reconcile.Result{RequeueAfter: shortWait}, nil
 	}
 
-	log = log.WithValues(
-		"composition-uid", comp.GetUID(),
-		"composition-version", comp.GetResourceVersion(),
-		"composition-name", comp.GetName(),
-	)
```
negz, line 401: This was removed because if we derive our composition from a revision it won't actually have a UID, version, or name.

internal/controller/apiextensions/composite/reconciler.go
```diff
@@ -408,14 +418,21 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		return reconcile.Result{RequeueAfter: shortWait}, nil
 	}
 
+	if err := r.composite.Configure(ctx, cr, comp); err != nil {
+		log.Debug(errConfigure, "error", err)
+		r.record.Event(cr, event.Warning(reasonCompose, err))
+		return reconcile.Result{RequeueAfter: shortWait}, nil
+	}
```
negz, line 425: Another bit of drive-by refactoring. I moved this down from where it was at line 391 because we should validate the Composition _before_ we use it to configure the XR - not after.

negz (commented):
I've added a few comments to hopefully make reviewing this giant PR a little easier. I'd suggest reviewing it on a commit by commit basis.

internal/controller/apiextensions/composite/api.go
```diff
@@ -113,6 +119,107 @@ func (a *APIFilteredSecretPublisher) UnpublishConnection(_ context.Context, _ re
...
+		if rl.Items[i].GetLabels()[v1alpha1.LabelCompositionSpecHash] == currentHash {
+			latestIdx = i
+		}
+	}
+
+	if latestIdx == -1 {
+		return nil, errors.New(errNoCompatibleCompositionRevision)
+	}
+
+	latest := &rl.Items[latestIdx]
```
turkenh, line 212: After reading the code here, I'm wondering if there is any functionality of the `revision` field in the `CompositionRevision` spec other than just showing how many revisions were created (excepts reverts to an older state). 
I was expecting it to be somehow used to find the latest but at the same time, I was aware that wouldn't work in case of reverts. Now, I am wondering if we could drop that `revision` field completely or make it functional (i.e. rely on it to find the latest here) by also increasing it in case of reverts.

internal/controller/apiextensions/composition/reconciler.go
```diff
@@ -0,0 +1,175 @@
...
+		"uid", comp.GetUID(),
+		"version", comp.GetResourceVersion(),
+		"name", comp.GetName(),
+		"spec-hash", currentHash,
+	)
+
+	rl := &v1alpha1.CompositionRevisionList{}
+	if err := r.client.List(ctx, rl, client.MatchingLabels{v1alpha1.LabelCompositionName: comp.GetName()}); err != nil {
+		log.Debug(errListRevs, "error", err)
+		r.record.Event(comp, event.Warning(reasonCreateRev, err))
```
turkenh, line 145: nit: shouldn't the event also include information that this error happened during listing revisions (same for get above)? 

For example:
```suggestion
		r.record.Event(comp, event.Warning(reasonCreateRev, errors.Wrap(err, errListRevs)))
```

turkenh (approved):
Looking pretty good with just a few non-blocking comments.

Also thank you very much for all the details you've provided in the [proposal](https://github.com/crossplane/crossplane/issues/1481), [design doc](https://github.com/crossplane/crossplane/blob/master/design/one-pager-composition-revisions.md), and code comments which extremely helped to understand the logic and helped a lot for reviewing.

apis/apiextensions/v1alpha1/revision_types.go
```diff
@@ -0,0 +1,413 @@
...
+
+	// WriteConnectionSecretsToNamespace specifies the namespace in which the
+	// connection secrets of composite resource dynamically provisioned using
+	// this composition will be created.
+	// +optional
+	WriteConnectionSecretsToNamespace *string `json:"writeConnectionSecretsToNamespace,omitempty"`
+
+	// Revision number. Newer revisions have larger numbers.
+	// +immutable
+	Revision int64 `json:"revision"`
```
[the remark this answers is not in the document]
negz: > Couldn't we find another way other than copying all types and fields from composition_types.go here, like an inline struct?

I think an inline struct would be the way to go once they're all in the same package(s) (e.g. v1beta1 and v1). As it stands with this new type being at v1alpha1, it "felt" a little strange to me for an alpha package to import types from a v1 package, but I could be convinced to do so in the name of simplification. One downside would be that if the v1alpha1 package ever did diverge more from the v1 package we'd need to redo all this copying work again.

> Also wondering, how can we ensure that they will always stay identical?

I don't have a good answer, apart from perhaps adding a comment? It's worth noting that we already have this problem for our v1beta1 and v1 types, which must remain identical.

cmd/crossplane/core/core.go
```diff
@@ -58,6 +59,8 @@ type startCommand struct {
 	LeaderElection bool          `short:"l" help:"Use leader election for the controller manager." default:"false" env:"LEADER_ELECTION"`
 	Registry       string        `short:"r" help:"Default registry used to fetch packages when not specified in tag." default:"${default_registry}" env:"REGISTRY"`
 	Sync           time.Duration `short:"s" help:"Controller manager sync period duration such as 300ms, 1.5h or 2h45m" default:"1h"`
+
+	EnableAlphaCompositionRevisions bool `help:"Enable alpha support for CompositionRevisions."`
```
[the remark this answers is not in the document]
negz: Yeah, I thought about something like this too. Personally I feel that the UX of distinct flags is better, but I agree that makes them tricky to deprecate. One thing we could do would be to keep the flags around for a few releases even after the feature has been promoted to beta.

negz:
Assigning @muvaf and @hasheddan because I need a maintainer approve (in addition to @turkenh's) before I can merge.

cluster/crds/pkg.crossplane.io_controllerconfigs.yaml
```diff
@@ -323,10 +323,71 @@ spec:
                                         The requirements are ANDed.
                                       type: object
                                   type: object
+                                namespaceSelector:
```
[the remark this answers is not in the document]
hasheddan: @negz I believe it is from the k8s version bumps (which I think is what you are saying) -- looks strictly additive and only enabled via alpha flag, so I'm not concerned 👍🏻

cmd/crossplane/core/core.go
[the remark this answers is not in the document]
hasheddan: > One thing we could do would be to keep the flags around for a few releases even after the feature has been promoted to beta.

Just noting here that I believe we _must_ keep them for beta, as that would mean they are "on by default", but can still be turned off. I also like the idea of having distinct flags because a configuration value silently not modifying behavior is negative attribute IMO (i.e. I want to break if I tell a service to behave in a way it does not understand) 👍🏻 

@negz I do wonder if it would make sense to add `default:"false"` to the flag so it will show in the help output (I know typically a bool flag will always be `false` if unspecified, but if we follow the k8s "on by default" strategy this bool flag will actually default to `true` in the future). You can also make the help output group all feature flags together using a group (example: https://github.com/upbound/up/blob/8e5965a42eff0285820247abed47cd91f2d26848/cmd/up/uxp/uxp.go#L40)

design/one-pager-composition-revisions.md
```diff
@@ -116,9 +116,9 @@ spec:
 apiVersion: apiextensions.crossplane.io/v1alpha1
 kind: CompositionRevision
 metadata:
-  # The revision's name is derived by appending the revision number to the
-  # Composition's name.
-  name: example-1
+  # The revision's name is derived from the composition's name using the API
+  # server's GenerateName support.
+  name: example-fklem2
```
hasheddan, line 121: Was there discussion around this change? I seem to remember someone mentioning it would be nice to be able to easily identify the revision number, but I am in support of this change because revision is in the spec and, more importantly for UX, defined as a printer column 👍🏻

internal/controller/apiextensions/composite/api.go
```diff
@@ -113,6 +119,107 @@ func (a *APIFilteredSecretPublisher) UnpublishConnection(_ context.Context, _ re
...
+		}
+
+		// We use the hash, not the revision number, to determine which
+		// revision is current. This is because the Composition may have
+		// been reverted to an older state. When this happens we don't
+		// create a new CompositionRevision; rather the prior revision
+		// becomes effectively 'current'.
+		if rl.Items[i].GetLabels()[v1alpha1.LabelCompositionSpecHash] == currentHash {
+			latestIdx = i
+		}
```
hasheddan, line 205: @negz there should only ever be one revision that matches `currentHash` correct? If so, can we break out of our range early here?

internal/feature/feature_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+		want := true
+		got := f.Enabled(cool)
+
+		if diff := cmp.Diff(want, got); diff != "" {
+			t.Errorf("f.Enabled(...): -want, +got:\n%s", diff)
+		}
+	})
+
+	t.Run("EnabledOnZeroValue", func(t *testing.T) {
```
hasheddan, line 40: Major Nitpick: aren't we checking here that a flag is _not_ enabled on the `Flags` zero value?

hasheddan (approved):
Awesome work @negz! I don't think any of my comments are blocking 👍🏻 I would support making the proposed updates to flag, but that could be accomplished in a follow-up if desired.

design/one-pager-composition-revisions.md
negz: The original discussion was in https://github.com/crossplane/crossplane/pull/2437#discussion_r679028275.

I could have sworn I made this change in a distinct commit with the rationale, but apparently not. I ended up going back to this approach because I realised it was more consistent with what we do for `ProviderRevision` and `ConfigurationRevision`. I like the idea of including the revision in the name, but I prefer consistency and as you say we show the revision pretty prominently as a printer column.

internal/controller/apiextensions/composite/api.go
negz: Yes - good catch.

cmd/crossplane/core/core.go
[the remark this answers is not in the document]
negz: > I do wonder if it would make sense to add default:"false" to the flag so it will show in the help output 

Unfortunately while `default` does appear to actually influence the default value it doesn't change the `--help` output in any way. That said, I do think there's a clean way to handle promoting features to beta. Specifically:

* We don't mention `beta` or `alpha` in the flag name. I've changed this to just `--enable-composition-revisions`.
* We use the `group` as you suggested to make it clear whether a feature is alpha or beta.
* When a feature reaches beta we add `default:"true" negatable:""` to the tags.

This means anyone who is enabling the feature by passing `--enable-composition-revisions` at alpha will not be affected when the feature hits beta. Anyone who wants to disable the feature when it hits beta can pass `--no-enable-composition-revisions`.

PR #2455, opened 2021-08-03
Allow user control over pod labels using controllerconfig

### LINKED ISSUE #2398: Add static label to providers identifying the package
We have installed `provider-sql` using the `Provider` CRD. I would like to restart the deployment using a predictable name, and add security groups to the pods, but the names of the pods/deployments are generated. One way to achieve this would be to use label selectors.

```
kind: Provider
metadata:
  name: crossplane-provider-sql
spec:
  package: crossplane/provider-sql:v0.2.0
```
```
    labels:
      pkg.crossplane.io/revision: crossplane-provider-sql-16bdf1af9403
      pod-template-hash: 997d559fb
```

Add a label at least for the package name, without tags. It might be useful to add more labels too as they will show up in prometheus etc.

chlunde:
To be able to select pods using PodSecurityPolicy in EKS, we need to
have predictable labels on the pods. Currently only an unpredictable
hash label is provided on the pods, and the user can only add
annotations.

This change adds a label map for the pod template, similar to the
existing annotation map.

To ensure future extensibility labels with crossplane.io prefixes are
forbidden in the documentation, and the code is structured to ensure
that the template will match the pod selector, and that any user
provided labels which overlaps with the selector will be overriden.

Fixes #2398

I have:

* existing tests work, but coverage is low for ControllerConfig
* checked with kind cluster


chlunde:
```yaml
kind: ControllerConfig
metadata:
  name: sql-config
  # https://github.com/crossplane/crossplane/issues/2398#issuecomment-867570603
  labels:
    provider: sql
spec:
  metadata:
    labels:
      foo: bar
```

```yaml
apiVersion: v1
items:
- apiVersion: v1
  kind: Pod
  metadata:
    creationTimestamp: "2021-08-03T11:12:16Z"
    generateName: crossplane-provider-sql-16bdf1af9403-546677cfb5-
    labels:
      foo: bar
      pkg.crossplane.io/revision: crossplane-provider-sql-16bdf1af9403
      pod-template-hash: 546677cfb5
```

negz:
@turkenh @ulucinar would one of you mind owning review for this PR? I'm happy to merge if and when it looks good to you. Thank you!

chlunde:
I would also be happy with two alternative implementations:
* just add another fixed label with the package name
* copy the labels from the controllerconfig

apis/pkg/v1alpha1/config.go
```diff
@@ -142,6 +142,15 @@ type PodObjectMeta struct {
 	// More info: http://kubernetes.io/docs/user-guide/annotations
 	// +optional
 	Annotations map[string]string `json:"annotations,omitempty"`
+
+	// Map of string keys and values that can be used to organize and
+	// categorize (scope and select) objects. This will only affect
+	// labels on the pod, not the pod selector. Labels will be merged
+	// with internal labels used by crossplane, and labels with the
+	// crossplane.io/ namespaces might be overwritten.
```
hasheddan, line 150: nitpick
```suggestion
	// with internal labels used by crossplane, and labels with a
	// crossplane.io key might be overwritten.
```

chlunde:
@turkenh can you trigger CI run here? "First-time contributors need a maintainer to approve running workflows"

turkenh:
> @turkenh can you trigger CI run here? "First-time contributors need a maintainer to approve running workflows"

@chlunde I am afraid I don't have that right yet. Maybe @muvaf can help?

internal/controller/pkg/revision/deployment.go
```diff
@@ -99,6 +98,16 @@ func buildProviderDeployment(provider *pkgmetav1.Provider, revision v1.PackageRe
 		if cc.Spec.Metadata != nil {
 			d.Spec.Template.Annotations = cc.Spec.Metadata.Annotations
 		}
+		templateLabels := make(map[string]string)
+		if cc.Spec.Metadata != nil {
+			for k, v := range cc.Spec.Metadata.Labels {
+				templateLabels[k] = v
+			}
+		}
+		for k, v := range d.Spec.Selector.MatchLabels { // ensure the template matches the selector
```
turkenh, line 107: How would the template matches the selector if no controller config provided?

internal/controller/pkg/revision/deployment.go
```diff
@@ -63,7 +63,6 @@ func buildProviderDeployment(provider *pkgmetav1.Provider, revision v1.PackageRe
 				ObjectMeta: metav1.ObjectMeta{
 					Name:      provider.GetName(),
 					Namespace: namespace,
-					Labels:    map[string]string{"pkg.crossplane.io/revision": revision.GetName()},
```
turkenh, line 66: Similar/related to the other comment, we would lose this label if no controller config is provided.

internal/controller/pkg/revision/deployment.go
chlunde: Thanks! Fixed

internal/controller/pkg/revision/deployment_test.go
```diff
@@ -0,0 +1,159 @@
+package revision
```
negz, line 1: Thank you for adding these tests @chlunde!

internal/controller/pkg/revision/deployment_test.go
```diff
@@ -0,0 +1,159 @@
...
+	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
+	pkgmetav1 "github.com/crossplane/crossplane/apis/pkg/meta/v1"
+	v1 "github.com/crossplane/crossplane/apis/pkg/v1"
+	"github.com/crossplane/crossplane/apis/pkg/v1alpha1"
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
```
negz, line 15: The linter comment is a little vague here; it's asking you to separate out all the `crossplane/crossplane` (i.e. "local") imports into a group separated by a line of whitespace. We typically separate out runtime imports too, so it looks something like:

```go
import (
	"stdlib-imports"

	"third-party-imports"

	"crossplane-runtime-imports"

	"crossplane-imports"
)

Comment (chlunde) [reply internal/controller/pkg/revision/deployment_test.go:15]: :+1: 

forgot `make reviewable` after adding the test.

Running `goimports -w -local github.com/crossplane/crossplane internal/controller/pkg/revision/deployment_test.go` now

File: internal/controller/pkg/revision/deployment_test.go (hunk 4 above)
Comment (chlunde) [reply internal/controller/pkg/revision/deployment_test.go:1]: Lesson learned, even for a "trivial" change, add tests even if the file does not have any already :)

PS! This doesn't add full coverage for all the existing branches here

**Review by ulucinar — APPROVED** [review]
LGTM. Thank you @chlunde!

File: internal/controller/pkg/revision/deployment_test.go [hunk 6]
```diff
@@ -0,0 +1,160 @@
...
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
+	pkgmetav1 "github.com/crossplane/crossplane/apis/pkg/meta/v1"
```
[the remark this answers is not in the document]
chlunde: right, running `goimports -w -local github.com/crossplane/crossplane internal/controller/pkg/revision/deployment_test.go` now wasn't enough...

internal/controller/pkg/revision/deployment_test.go
```diff
@@ -0,0 +1,161 @@
+package revision
```
hasheddan, line 1: @chlunde thanks again for adding tests here :) could you make sure to add the appropriate license header to this file?

hasheddan (approved):

PR #2459, opened 2021-08-06
Remove Composition status

negz:

This status was added by accident, and has never been used. It doesn't really hurt, but it's a little misleading given that a Composition never has status conditions in practice.

Removing this is technically a breaking API change, but none of our controllers have ever set it so I can't imagine anyone is actually depending on it.

I have:

N/A


negz:
I'm assigning this to @hasheddan for a sanity check, since you usually have a conservative approach to breaking changes. :)

hasheddan (approved):
Given that these are never populated today, I agree that they can be safely removed without a version bump or deprecation schedule :) Probably would be good to note in the 1.4 release notes though 👍🏻

PR #2460, opened 2021-08-06
Remove v1beta1 Package interfaces

negz:

I believe we only use the v1 versions of these interfaces, and that the v1beta1 versions are now vestigial; we only need the struct type definitions.

I have:

N/A


hasheddan (approved):
Good catch! Thanks @negz :)

PR #2453, opened 2021-07-29
Adds support for building/running using darwin_arm64

tnthornton:
### Description of your changes

These changes allow individuals using Apple Silicon to build/run crossplane locally.
Changes outlined:
1. bring in update to build submodule that updates kind to latest version
2. update hardcoded references to `arm64` for the crossplane image tag
3. updates the helper functions to work with arm64 in addition to amd64

I have:

- [X] Read and followed Crossplane's [contribution process].
- [X] Run `make reviewable` to ensure this PR is ready for review.
- [ ] Added `backport release-x.y` labels to auto-backport this PR if necessary.

### How has this code been tested
1. `make`
2. `make test`
3. `make e2e SAFEHOSTARCH=arm64` -  for now, you'll need to override `SAFEHOSTARCH` in this manner when running on darwin_arm64
4. `make reviewable`
5. `./cluster/local/kind.sh up`
6. `./cluster/local/kind.sh helm-install`

[contribution process]: https://git.io/fj2m9

PR #2466, opened 2021-08-10
output darwin_amd64 for arm64 (m1)

benofben:
Signed-off-by: Ben Lackey <ben.lackey@outlook.com>

I'm running on an MacBook Air (M1, 2020).  The arch on that comes back as "arm64."  Strictly speaking that's true.  However, there's no build for arm64.  There is a build for amd64 that runs just fine in the Mac's emulation.  So, the install script should probably grab that.

Interestingly this exposed a second issue with the script.  It's spitting out this url: https://releases.crossplane.io/stable/current/bin/darwin_arm64/crank.  It probably ought to be generated a unsupported architecture message.  So, it looks like the logic in install.sh isn't detecting that right.

This PR aims to fix both those things without introducing any fun new bugs.

The cause of the issue seems to be that there's a case statement for arm64 but no binary to match it.  So, this PR drops the case statement and downloads Darwin amd64 on the Darwin arm64 system.  I assume the arm64 case is a branch for a not yet existing build that made it into the repo too fast.

I also did a "chmod +x install.sh" as the script is unrunnable in its current state.  Though, it's conceivable there's some reason that wasn't done before.

I have:

I tried to change nothing but the issue I saw and then ran the install script on my local machine.  I can't think of any way this can break things but I don't know how this install.sh is used (other than in the documentation I've been following).

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -32,11 +32,8 @@ case $OS in
     ;;
   Darwin)
     case $ARCH in
-      x86_64|amd64)
+      x86_64|amd64|arm64)
         OS_ARCH=darwin_amd64
-        ;;
-      arm64)
-        OS_ARCH=darwin_arm64
         ;;
       *)
         unsupported_arch $OS $ARCH


hasheddan (commented):
Thanks for opening this @benofben! This functionality was recently added, as was support for building for `darwin_arm64` (https://github.com/crossplane/crossplane/pull/2453). However, until the next release (v1.4 at the end of August), this build will not be available on `stable`. The script should currently work if passed with `CHANNEL=master` or `ARCH=amd64`. I don't love that this currently fails on `stable`, but I do like that it allows for folks on M1 macs to get binaries from `master` with the script today. I think we could either go with your revert here, or we could denote the current limitations in the docs until the v1.4 release. Do you have a preference as an M1 user? :)

benofben:
Yeah, kinda figured this was the install script getting ahead of the releases.  The behavior of s3 is a little weird where it doesn't 404, rather generating html, so this thing fails silently.  What got me was the silent failure.  You could always have the script throw an error if the download is ascii rather than binary (indicating an autogenerated s3 page rather than the binary you were supposed to get).  That (or something similar) might fix the issue longer terms as archs come and go.

I don't have a strong feeling about the solution, was more just trying to do a good deed.

tnthornton:
Thanks for putting this together @benofben! 

There are a couple of byproducts of my changes in #2453 that you both call out that standout to me:
> However, until the next release (v1.4 at the end of August), this build will not be available on stable. The script should currently work if passed with CHANNEL=master or ARCH=amd64. I don't love that this currently fails on stable

and

> The behavior of s3 is a little weird where it doesn't 404, rather generating html, so this thing fails silently. What got me was the silent failure. 

I think given the above, I'd vote for taking the path of reverting the case statement and maybe opening an issue where we tracked the things that need to be added once we get to the next release v1.4 that would keep these pain points from happening again (adding arm64 as it's own case again, etc).

What're your thoughts @hasheddan?

hasheddan (approved):
@tnthornton I agree with your analysis 👍🏻 @benofben would you mind opening an issue tracking the desired improvements you noted here?

benofben:
Here you go.  https://github.com/crossplane/crossplane/issues/2474

hasheddan:
Thanks @benofben !

Issue #2464, opened 2021-08-06
Race condition while setting XR's `spec.claimRef`


ulucinar:
### What happened?

Previously while working on some experiments with compositions to answer a [question](https://crossplane.slack.com/archives/C01718T2476/p1624587843169500) in the Crossplane providers slack channel, I have observed a race condition that causes an AWS `Bucket` MR (and the corresponding S3 bucket) to leak. If a bucket leaks and when the corresponding claim is deleted, the leaked `Bucket` MR and the corresponding S3 bucket are not deleted. 

Please note that this case involves patching from XR's `spec.claimRef.name` to composed's external-name annotation and the race condition seems to be between the composite reconciler (while rendering the composed resources) and the claim reconciler (when a claim is bound to the XR). Please see the comments below for some further context.

### How can we reproduce it?

Back then, I could reproduce the issue with the following manifests. Please note that the race condition does not always occur by definition. 
```yaml
---
apiVersion: apiextensions.crossplane.io/v1
kind: CompositeResourceDefinition
metadata:
  name: compositebuckets.aws.example.org
spec:
  group: aws.example.org
  names:
    kind: CompositeBucket
    plural: compositebuckets
  claimNames:
    kind: CompositeBucketClaim
    plural: compositebucketclaims
  versions:
  - name: v1alpha1
    served: true
    referenceable: true

---
apiVersion: aws.example.org/v1alpha1
kind: CompositeBucketClaim
metadata:
  name: my-bucket-with-policy
  namespace: default
spec:
  compositionSelector:
    matchLabels:
      provider: aws
      policy: test

---
apiVersion: apiextensions.crossplane.io/v1
kind: Composition
metadata:
  name: bucket-with-policy
  labels:
    provider: aws
    policy: test
spec:
  compositeTypeRef:
    apiVersion: aws.example.org/v1alpha1
    kind: CompositeBucket
  resources:
    - name: bucket
      base:
        apiVersion: s3.aws.crossplane.io/v1beta1
        kind: Bucket
        spec:
          forProvider:
            acl: private
            locationConstraint: us-east-1
      patches:
        - fromFieldPath: "spec.claimRef.name"
          toFieldPath: "metadata.annotations['crossplane.io/external-name']"
        - fromFieldPath: "status.atProvider.arn"
          toFieldPath: "metadata.annotations['bucket-arn']"
          type: ToCompositeFieldPath
    - name: bucket-policy
      base:
[...]

Issue #2078, opened 2021-01-14
Streamline claim binding logic


negz:
### What problem are you facing?

@mcavoyk points out in https://github.com/crossplane/crossplane/pull/2077/files#r557603114 that our claim reconciler:

1. Creates an XR.
1. Updates the claim to reference the XR.
1. Updates the XR to reference the claim.

This means that we make multiple calls to create and bind a new XR, and that XRs exist for a brief period of time without referencing a claim. There's a bunch of subtle things that lead to this logic - it could probably be explained a little better in the comments.

I think this is partly the result of deriving this reconciler logic from an older iteration of the claim concept. We support what we think of as "static provisioning" - you can create an XR then later claim it, as opposed to provisioning it "dynamically" in response to the creation of the claim. This caused us to keep XR creation and binding separate, since sometimes we only wanted to bind. This is less relevant in the contemporary claim reconciler though, since we always configure the XR, apply the XR, then bind the XR whether it exists already or not.

I suspect another thing going on here is that we're trying to mitigate the impact of leaked XRs. The creation and binding order is important because the claim reconciler is configured to watch both claims and XRs, and it resolves XRs to a claim to reconcile via their claim ref. If we found ourselves in a situation in which we created an XR with a reference to a claim, but then hit an error before we updated the claim to reference the XR we would likely create duplicate XRs. We'd enqueue a request for the XR's claim, which would not know about the XR and thus create a new one.

### How could Crossplane help solve your problem?

It may be worth adopting [this pattern](https://github.com/crossplane/crossplane/blob/e0767ae5/internal/controller/apiextensions/composite/composed.go#L128) from the XR reconciler here. If we knew the XR's name before it was created we could:

1. Generate a name for the XR.
2. Persist that name to the claim's ref.
3. Have `Configure` handle configuring the XR to reference the claim.

PR #2467, opened 2021-08-10
Persist claim resourceRefs before creating XRs

negz:

Fixes https://github.com/crossplane/crossplane/issues/2078
Fixes https://github.com/crossplane/crossplane/issues/2464
Fixes https://github.com/crossplane/crossplane/issues/2473

This PR updates the claim reconciler to persist a claim's `spec.resourceRef` before dynamically provisioning an XR. It uses a similar approach to the XR reconciler in that it now uses a dry-run create to generate an XR name (and validate the XR) without actually creating it. 'Binding' a claim to an XR is now broken up into three stages.

1. We set the XR's claimRef during Configure.
2. We set (and persist) the XRC's resourceRef during Bind
3. We persist the XR's claimRef during Apply.

This approach has a handful of benefits:

1. It prevents 'leaking' XRs. Previously if anything went wrong between when we first created an XR and when we persisted its resourceRef we'd return from the Reconcile and requeue a new one. The new reconcile would have no way of knowing that we'd already created an XR for this claim so it would create a new one.
2. It ensures dynamically provisioned XRs never exist without a claimRef - it's set when the XR is created. This can help avoid issues like #2464. Note that XRs should still be designed to support static binding - i.e. they should not assume that the claimRef will always be set at XR creation time.
3. It eliminates one superfluous API server write, because we only update the XR once, rather than applying it then updating it to bind it.

Note that this commit changes the behaviour of the (AFAIK _exceedingly_ obscure) edge case in which someone attempts to claim an XR tha does not exist - i.e. by explicitly setting their claim's resourceRef. Previously we'd assume they did in fact want to claim an XR that hadn't yet been created, and wait around for it to exist. Now we assume that the resourceRef was populated by a previous iteration of the reconciler and proceed to dynamically provision an XR of that name. One interesting side effect here is that this means claim authors can now influence the (Kubernetes) name of the XR they would like to provision, though we'll still bail out if they try to 'claim over' an XR that is claimed by some other entity.

This PR also contains a small additional fix for https://github.com/crossplane/crossplane/issues/2473, which I (re)discovered while addressing the original issue. The original refactoring done in this PR is a partial fix for this issue because we now detect 'misbound' claims earlier and refuse to make any mutations on composite resources we don't own, so I only needed to fix the delete flow to ensure a claim could not delete an XR it was not bound to.

I have:

I've tested this by walking through the GCP getting started guide and confirming that dynamic provisioning works as expected. I've also followed similar steps to those outlined in https://github.com/crossplane/crossplane/issues/2473 to ensure I can't reproduce that issue with this PR.

### DIFF
--- a/internal/controller/apiextensions/claim/api.go
+++ b/internal/controller/apiextensions/claim/api.go
@@ -56,8 +56,18 @@ func NewAPIBinder(c client.Client) *APIBinder {
 func (a *APIBinder) Bind(ctx context.Context, cm resource.CompositeClaim, cp resource.Composite) error {
 	existing := cm.GetResourceReference()
 	proposed := meta.ReferenceTo(cp, cp.GetObjectKind().GroupVersionKind())
-	if existing != nil && !cmp.Equal(existing, proposed, cmpopts.IgnoreFields(corev1.ObjectReference{}, "UID")) {
+	equal := cmp.Equal(existing, proposed, cmpopts.IgnoreFields(corev1.ObjectReference{}, "UID"))
+
+	// We refuse to 're-bind' a claim that is already bound to a different
+	// composite resource.
+	if existing != nil && !equal {
 		return errors.New(errBindClaimConflict)
+	}
+
+	// There's no need to call update if the claim already references this
+	// composite resource.
+	if equal {
+		return nil
 	}
 
 	cm.SetResourceReference(proposed)
--- a/internal/controller/apiextensions/claim/api.go
+++ b/internal/controller/apiextensions/claim/api.go
@@ -34,13 +34,11 @@ import (
 
 // Error strings.
 const (
-	errUpdateClaim           = "cannot update composite resource claim"
-	errUpdateComposite       = "cannot update composite resource"
-	errBindClaimConflict     = "cannot bind claim that references a different composite resource"
-	errBindCompositeConflict = "cannot bind composite resource that references a different claim"
-	errGetSecret             = "cannot get composite resource's connection secret"
-	errSecretConflict        = "cannot establish control of existing connection secret"
-	errCreateOrUpdateSecret  = "cannot create or update connection secret"
+	errUpdateClaim          = "cannot update composite resource claim"
+	errBindClaimConflict    = "cannot bind claim that references a different composite resource"
+	errGetSecret            = "cannot get composite resource's connection secret"
+	errSecretConflict       = "cannot establish control of existing connection secret"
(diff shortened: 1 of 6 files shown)


negz:
A little more testing - I ran through the slightly more complex AWS (with VPC) getting started guide and that also appears to work as expected.

internal/controller/apiextensions/claim/api.go
```diff
@@ -34,74 +34,56 @@ import (
...
 func (a *APIBinder) Bind(ctx context.Context, cm resource.CompositeClaim, cp resource.Composite) error {
 	existing := cm.GetResourceReference()
-	proposed := meta.ReferenceTo(cp, resource.MustGetKind(cp, a.typer))
-	if existing != nil && !cmp.Equal(existing, proposed, cmpopts.IgnoreFields(corev1.ObjectReference{}, "UID")) {
+	proposed := meta.ReferenceTo(cp, cp.GetObjectKind().GroupVersionKind())
+	equal := cmp.Equal(existing, proposed, cmpopts.IgnoreFields(corev1.ObjectReference{}, "UID"))
+
+	// We refuse to 're-bind' a claim that is already bound to a different
+	// composite resource.
+	if existing != nil && !equal {
```
hasheddan, line 63: nitpick: I believe if `existing == nil` we could return early with claim update, rather than checking for equality with a nil object, but I think what you have here may actually be more readable.

internal/controller/apiextensions/claim/configurator_test.go
```diff
@@ -123,6 +111,89 @@ func TestCompositeConfigure(t *testing.T) {
 				},
 				cp: &composite.Unstructured{},
 			},
+			want: want{
+				cp:  &composite.Unstructured{},
+				err: errors.New(errUnsupportedClaimSpec),
+			},
+		},
+		"AlreadyClaimedError": {
+			reason: "We should return an error if we appear to be configuring a composite resource claimed by a different... claim.",
```
hasheddan, line 120: 😅

hasheddan (approved):
This LGTM! Thanks for this improvement @negz!

internal/controller/apiextensions/claim/api.go
negz: Ah, good point. I don't feel super strongly about either approach being significantly better so laziness wins out and I'm going to keep it as is. ;)

negz:
/backport

(Again, because I was too lazy to rebase the previous PRs.)

Issue #2366, opened 2021-06-11
Support for external secret stores


negz:
### What problem are you facing?

All conformant Crossplane managed resource support publishing their "connection details" - things like API endpoints and credentials - to a Kubernetes secret. All resources offer a `spec.writeConnectionSecretToRef` field. The managed resource reconciler will always write the referenced secret if specified, though what connection details to write (if any) is up to each managed resource implementation.

An increasing number of organisations prefer to avoid Kubernetes secrets altogether in favour of external secret stores. The most common external store we see folks using is Vault, per https://github.com/crossplane/crossplane/issues/1424. Unfortunately there's currently no way for Crossplane to emit connection details to these stores (at least not without doing it by writing a Kubernetes secret and letting something else propagate that secret to the relevant store).

### How could Crossplane help solve your problem?

https://github.com/crossplane/crossplane-runtime/blob/ded1778/pkg/reconciler/managed/reconciler.go#L417

Crossplane could support pluggable secret stores. The `crossplane-runtime` `managed.Reconciler` implementation is actually already fairly decoupled from Secrets - you can pass in an alternative `ConnectionPublisher` implementation when initialising the reconciler. The key issues here are:

* Only one `ConnectionPublisher` (that manages  a Kubernetes secret) exists.
* The XRM currently only offers `spec.writeConnectionSecretToRef` - no `spec.writeConnectionDetailstoVault` or similar.

I believe we should start approaching this by sketching out what our connection detail publishing API would look like if we supported pluggable secret stores.

## Implementation

I (@turkenh) would like to behave this ticket as an epic for ESS (External Secret Store) and would like to divide the work into multiple phases for better manageability and traceability. Here is my plan:

### Phase 1: Laying the Foundation

**Deliverable:**  An MR writes connection details to a **K8s secret** using [the new API](https://github.com/crossplane/crossplane/blob/master/design/design-doc-external-secret-stores.md#api).

- [x] In Crossplane Runtime: [Implement the Foundation for External Secret Store Support](https://github.com/crossplane/crossplane-runtime/issues/315) => https://github.com/crossplane/crossplane-runtime/pull/321
- [x] In Provider GCP: Define "SecretStoreConfig" and manually add new methods to satisfy new interfaces _to test / demonstrate the deliverable_. => https://github.com/crossplane/provider-gcp/pull/417

### Phase 2: Vault as an ESS

**Deliverable:**  An MR writes connection details to a **Vault secret**.

[...]

PR #2494, opened 2021-08-18
Add Design for External Secret Stores

turkenh:

This PR adds a design doc for supporting external secret stores (e.g. Vault) in Crossplane as described in [this ticket](https://github.com/crossplane/crossplane/issues/2366).

I have:

N/A


design/design-doc-external-secret-stores.md
```diff
@@ -0,0 +1,416 @@
...
+`writeConnectionSecretsToNamespace` in favor of the same 
+`publishConnectionDetailsTo` field. This would be used for any partial input
+that needs to be enforced/provided by the composition authors.
+
+```
+apiVersion: apiextensions.crossplane.io/v1
+kind: Composition
+metadata:
+  name: compositeclusters.aws.platformref.crossplane.io
+spec:
```
turkenh, line 98: Comment from @negz 

The writeConnectionSecretsToNamespace field on a Composition today controls which namespace XRs (that use that Composition) will write secrets to. So it's a partial input to the XR's writeConnectionSecretsToRef. We needed to add this because we wanted platform operators, not claim authors, to choose which namespace the XR secret was written to.

I'm not sure this makes sense in the Vault context. In fact we probably need to think more about the relationship between MR, XR, and XRC connection details when vault is in use. Do we need to write all three like we do when we use Secrets?

turkenh: I think we could still pass some partial input in the Vault context like which config to use or even a parent path to use in Vault which could be enforced by platform operators. This would however require us to adapt our interface in a way that we get `parentPath` as a separate argument.

 - type: Vault 
   vault:
     parentPath: /path/to/parent
     name: my-cool-managed-resource-connection 
     configRef: 
       name: vault-primary


However, I am not sure how this is applicable for other secret stores that we would want to support in the feature.

design/design-doc-external-secret-stores.md
```diff
@@ -0,0 +1,416 @@
...
+if requested in the managed resource API (`publishConnectionDetailsTo` includes
+`Vault` as store).
+
+![Secret Store Plugin Example][secret-store-plugin-example]
+
+#### Option 3: Propagate from K8S secrets
+
+_In this option we assume that it is ok to keep storing connection details as
+Kubernetes secrets (temporarily or permanently) as long as they will also be
+propagated to external secret stores if configured/requested._
```
turkenh, line 271: Comment from @negz 

This is my preferred approach from an architectural perspective, but I think we will find that we can't make this assumption. I'm pretty sure this option will be a non-starter because anecdotally I've heard multiple orgs state that their security teams deem Kubernetes Secrets to be insecure, and thus they can't be used to store sensitive data at all.

design/design-doc-external-secret-stores.md
```diff
@@ -0,0 +1,416 @@
...
+Crossplane’s existing extensibility solution, which is providers, if 
+communication over k8s api (via secrets) is not acceptable. **Option 2** tries
+to find a solution by introducing a new plugin solution while also trying to
+keep things simple by not introducing an additional authentication and
+authorization problem by suggesting a communication over unix domain sockets.
+However, since all providers and Crossplane itself require this functionality,
+deploying plugins as sidecars introduces another complexity which could be
+solved with a sidecar injecting mutating webhook.
+
+**Option 1**, as the most straightforward solution, suffers from the need for
```
turkenh, line 359: Comment from @negz 

Is there a possibility to start with option 1, but keep the option to evolve toward option 2 without breaking API changes? If so, that would be my preference.

My thinking is that building the plugin approach upfront is a little heavyweight if we find that maybe Secrets, Vault, and one or two cloud KMS services are all we need to support at first. It also would be subject to a less 'strongly typed' API. We could wait until it was becoming painful to support many backends in runtime before we invested in the plugin model. Perhaps we would find that we actually never got to that point and doing it all in tree worked pretty well.

I could imagine having 'plugin' be a distinct type of secret store that we could add later - something like https://gist.github.com/negz/b018eb03a55be2697549327d8490786c

turkenh: I think that sounds reasonable. 
Once we gather some more feedback, I will consider moving **option 2** and **option 3** under the **alternatives considered** section and extend **option 1** with what you're suggesting.

braghettos: I also like the Option 1 approach! I didn't know the option for plugins in Crossplane, I was thinking more to implement a specific provider for Vault or other secret managers.

design/design-doc-external-secret-stores.md
```diff
@@ -0,0 +1,414 @@
...
+  name: my-cool-managed-resource
+spec:
+  forProvider:
+    coolness: 24
+  publishConnectionDetailsTo:
+  - store: KubernetesSecret
+    kubernetesSecret:
+      secretRef:
+        name: my-cool-managed-resource-connection
+        namespace: secrets
```
AaronME, line 80:
```suggestion
  - kubernetesSecretRef:
        name: my-cool-managed-resource-connection
        namespace: secrets
```

It seems like we could drop the `store` field and opt, instead, to detect the type from the name of the field submitted. I've seen this pattern already in use with our Refs and Selectors in Crossplane.

design/design-doc-external-secret-stores.md
AaronME: Based on the description in the `Configuring External Secret Stores` section, we might be double-configuring the secret store at this point.

If we assume that SecretStores will always be enabled/disabled by platform operators using a StoreConfig, then we need not identify the store type when we access the secret store.  We could do something like this for vault:

```
apiVersion: secrets.crossplane.io/v1alpha1
kind: StoreConfig
metadata:
  name: vault-primary
Spec:
  type: Vault
  credentials:
    source: Secret
    secretRef:
      namespace: crossplane-system
      name: vault-creds
      key: config
```

Then write to secrets with:
```
publishConnectionDetailsTo:
- storeName: vault-primary
  secretName: /path/to/my/secrets/my-secrets
```

Or, with AWS Secrets Manager:
```
apiVersion: secrets.crossplane.io/v1alpha1
kind: StoreConfig
metadata:
  name: aws-secrets-primary
Spec:
  type: AWSSecret
  credentials:
    source: Secret
    secretRef:
      namespace: crossplane-system
      name: aws-creds
      key: config
```
Leveraged by:
```
publishConnectionDetailsTo:
- storeName: aws-secrets-primary
  secretName: my-secrets
```

Kubernetes could be supported as a `default` storeConfig which is always a kubernetes secret.
```
publishConnectionDetailsTo:
- storeName: default
  secretNamespace: claim-namespaces
  secretName: my-secrets
```

design/design-doc-external-secret-stores.md
AaronME: This comment can be ignored in light of the discussion lower down about leaving Store Type determination to the StoreConfig.

design/design-doc-external-secret-stores.md
hasheddan: @muvaf I came here to bring this up and see that you beat me to it :) I have a working demo of this, and you are right that at first glance the read operation is not as clean as the write. That being said, I worked around this by changing composition to not actually read the managed resource `Secret`, but rather write the "composed" `Secret` with annotations that inform the webhook how to aggregate a new external secret from the managed resource external secrets it already knows about. I believe this was working well, and it actually reduces some of our permissions surface area because core Crossplane doesn't ever need to actually know the contents of any managed resource `Secret` it just instructs the mutating webhook to use another the contents of another `Secret` it knows about compose a new one.

muvaf: @hasheddan But we have a restriction that the sensitive information shouldn't end up in a secret. If I understood your suggestion correctly, there will be a secret with content, right?

hasheddan: @muvaf there will not be -- the webhook just has a notion of the mapping between the (empty) managed resource `Secrets` and their external representations. Composition machinery creates a composite `Secret` that has annotations referencing the (empty) managed resource `Secrets`, and the webhook intercepts and composes the new external composed secret by fetching the external managed resource secret data and combining them appropriately. The caveat here is that composition would have to be okay with writing an empty composite `Secret` when the managed resource `Secrets` are also empty.

turkenh: @hasheddan that's a neat approach. 
I am definitely happy to discuss further and am wondering if we could make this completely transparent to Crossplane, requiring no changes in composition logic at all. I believe this is not possible because webhooks cannot intercept read requests but wondering if we could extend that approach (intercepting secret requests) with a k8s API server proxy (similar to how we did in upbound agent proxy) instead of webhooks 🤔 
This would be a solution that could work for any Kubernetes application and not only for Crossplane.

negz: One thing I'm curious about with this approach is how folks would find the UX. Wouldn't (e.g) 'writeConnectionSecretToRef' be a little confusing if we're actually never writing a Secret? My hunch is that our types declaring the _actual_ secret sink in their APIs would be more intuitive and self documenting.

This problem arguably also exists in the case in which we're writing a Secret that is then propagated to some other system, but somehow the fact that the 'intermediary' Secret actually exists makes that seem more tolerable to me.

negz: That all said - this Secret intercept idea does sound cool and may be worth proving out further even if we don't use it in this specific context.

hasheddan: > I believe this is not possible because webhooks cannot intercept read requests but wondering if we could extend that approach (intercepting secret requests) with a k8s API server proxy (similar to how we did in upbound agent proxy) instead of webhooks

@turkenh this is an interesting approach that I had also considered, but I think we can alleviate the need to do so with writing an empty `Secret` with annotations as mentioned, which would mean that Crossplane (composition) would never try to read a `Secret`, it would just be essentially creating "secret pointers" that the webhook knew how to resolve.

> Wouldn't (e.g) 'writeConnectionSecretToRef' be a little confusing if we're actually never writing a Secret? (Or rather writing an empty one.) 

@negz this could absolutely be the case, and IMO is the biggest issue with this direction. However, I do think it is possible that it could actually be useful to have a marker that says "this thing exists externally" as it kind of creates a similar abstraction layer over secret stores that managed resources do over external APIs. We could make this even more clear by having the mutating webhook add an annotation that indicates what secret store backend the secret contents were actually written to, as well as (potentially) what fields are present in external secret.

design/design-doc-external-secret-stores.md
```diff
@@ -0,0 +1,414 @@
...
+
+```
+apiVersion: widgets.crossplane.io/v1beta1
+kind: Example
+metadata:
+  name: my-cool-managed-resource
+spec:
+  forProvider:
+    coolness: 24
+  publishConnectionDetailsTo:
```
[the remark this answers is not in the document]
merlindorin: Just a single use case, but at Renault, we have many Vault with different purposes (subsidiary, legal reasons, connectivity, etc.).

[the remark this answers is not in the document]
muvaf: @merlindorin do you see a case where you push the same credentials to multiple Vault stores?

braghettos:
Hi there! Are there any updates regarding this topic?

turkenh:
> Hi there! Are there any updates regarding this topic?

I couldn't find time to get back to this yet but on my to-do list.

design/design-doc-external-secret-stores.md
```diff
@@ -0,0 +1,655 @@
...
+```
+
+#### Separation of Concerns: Compositions, Composites and Claims
+
+`StoreConfig` resource will be a _namespaced_ resource to allow platform
+consumers to configure their own secret store of choice. This is aligned with
+the existing approach since Kubernetes secrets are namespaced resources and
+platform operators could specify where should composite resource secrets land
+using the partial input `writeConnectionSecretsToNamespace`. We will similarly
+deprecate this field in favor of `publishConnectionDetailsToStore`.
```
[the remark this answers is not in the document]
turkenh: I believe it would be important to support multi-tenant scenarios, where, platform operators share a control plane between multiple teams where each team has its own cloud account (e.g. want to use AWS secret store under their account), which could be quite common for big enterprises to manage/track cloud costs. And I remember a very recent discussion with a Platform engineer working with that setup (cc @luebken). The discussion was indeed around `ProviderConfig` and the question was how can I let my platform consumer teams use their own cloud accounts, but it also applies here.

Putting consistency concerns aside, I believe a namespaced `StoreConfig` and a cluster scoped `ClusterStoreConfig` would be ideal. And I am wondering if we can find a _not so great_ solution with some wording trick like:
Cluster scoped: `StoreConfig` and `ProviderConfig`
Namespaced: `TenantStoreConfig` or `NamespacedStoreConfig` and `TenantProviderConfig` or `NamespacedProviderConfig`

design/design-doc-external-secret-stores.md
```diff
@@ -0,0 +1,655 @@
...
+
+**Claim** spec for writing connection secret in the same Kubernetes namespace:
+
+```yaml
+spec:
+  publishConnectionDetailsTo:
+    name: database-creds 
+    # when neither `kubernetes` nor `externalStore` configs
+    # provided, it defaults to "kubernetes in the same
+    # namespace", same as "kubernetes: {}"
```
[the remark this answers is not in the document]
turkenh: We will lose the ability to write both to a Kubernetes secret and to an external one: https://github.com/crossplane/crossplane/pull/2494#discussion_r733602782

Other than that this will definitely be a simpler API.

design/design-doc-external-secret-stores.md
```diff
@@ -0,0 +1,655 @@
...
+          arbitrary-config: true
+```
+
+### User Experience
+
+With the API definition above, it is mostly clear how the user interaction would
+be. But one important point that worth mentioning is, both Crossplane core pod
+and provider pods would need to access to the external secret stores. The
+credentials that is made available needs to be authorized to read, write and
+delete the secrets living at the configured scope.
```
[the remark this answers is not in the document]
turkenh: I am not happy with introducing the only dependency between provider -> core Crossplane, but, UX-wise, I don't feel comfortable regarding the duplication. I wouldn't be too concerned regarding Provider or Crossplane not supporting a new store (yet) as long as we return a meaningful error message that users could easily find (i.e. via events). 
Putting implementation details aside, this sounds like a global configuration rather than a provider-specific one (different than `ProviderConfig`).
Also, I am wondering if we could just use the unstructured client to get `StoreConfig` in the providers not to have an explicit dependency and only look for it if a StoreConfig referenced (otherwise store as a k8s secret with some defaulting as you suggested). Or somehow put this type to crossplane-runtime so that we can register its schema from providers but install its CRD as part of Core Crossplane.

In any case, we can leverage being an alpha feature and revisit the decision we made here based on the community feedback after some baking period.

turkenh:
> 1. I'm not sold on `StoreConfig` being namespaced. I see your point that some orgs might want their consumers to configure the available stores, but I also think some orgs might want to limit that to their platform teams. I believe the same is true for `ProviderConfig`, and (for better or worse) we went with cluster scope for that. I'd prefer to be consistent.

As I mentioned in my response [here](https://github.com/crossplane/crossplane/pulls#discussion_r740794429), I believe having both cluster-scoped and namespaced flavors of _store configuration_ would be the most flexible and powerful option. I see this similar to platform consumers being able to specify the name of their connection secret within their scope (i.e. tenant namespace). 
However, I see your point regarding consistency but wondering if we could work around that with some naming trick as I mentioned in the above response, e.g. cluster-scoped `StoreConfig` and namespaced `TenantStoreConfig`.

> 2. I'm on the fence about there only being one kind of `StoreConfig` that is shared by core Crossplane and all providers. I could imagine for example it being confusing if the `StoreConfig` schema appeared to support (e.g.) AWS KMS but only some of the installed providers actually had that support at a code level. While I don't think `StoreConfig` will vary from provider to provider very much (apart from support for new stores potentially being added) my hunch is that maybe the looser coupling of defining one per provider (and one for core Crossplane) would be worth the duplication it would incur.

Please see my response [here](https://github.com/crossplane/crossplane/pulls#discussion_r740956181).

> 3. I'd like to consider merging writing Kubernetes Secrets into a `StoreConfig` too per my inline comment. I feel like this would result in a simpler, cleaner API and allow for more flexible configurable defaults.

Honestly, I liked the idea of being able to write both to a k8s secret and to an external store as I mentioned [here](https://github.com/crossplane/crossplane/pulls#discussion_r733602782). However, this would definitely be something that we could give up for a cleaner API given it is not an explicit goal of this design and there are already tooling available to achieve this with some additional work.

negz:
> However, I see your point regarding consistency but wondering if we could work around that with some naming trick as I mentioned in the above response, e.g. cluster-scoped StoreConfig and namespaced TenantStoreConfig.

Could we start with just the cluster scoped `StoreConfig` and wait for user feedback before adding a namespace scoped equivalent?

> Please see my response here.

I understand your concerns, but please let's start with provider scoped StoreConfigs. As you mention, we can always switch to a single one based on user feedback during the alpha. As much as I agree with the UX concerns I'm more concerned with tight coupling.

design/design-doc-external-secret-stores.md
```diff
@@ -0,0 +1,655 @@
...
+spec:
+  publishConnectionDetailsTo:
+    name: my-db-connection
+    kubernetes:
+      namespace: crossplane-system
+      labels:
+        environment: production
+      annotations:
+        acme.example.io/secret-type: infrastructure
+```
```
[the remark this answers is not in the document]
chlunde: Hi, I was going to create a new issue for specifying type and annotations, and noticed this comment. So I'll just add this to this thread instead:

Tekton secrets requires specific types specified, such as `type: kubernetes.io/basic-auth`, and also annotations:

```yaml
apiVersion: v1
kind: Secret
metadata:
  annotations:
    tekton.dev/git-0: https://github.com
    tekton.dev/git-1: https://gitlab.com
    tekton.dev/docker-0: https://gcr.io
type: kubernetes.io/basic-auth
stringData:
  username: <cleartext username>
  password: <cleartext password>
```

https://tekton.dev/vault/pipelines-v0.28.2/auth/#overview

Make it possible to define annotations and secret type for connection secrets.

turkenh:
> > However, I see your point regarding consistency but wondering if we could work around that with some naming trick as I mentioned in the above response, e.g. cluster-scoped StoreConfig and namespaced TenantStoreConfig.
> 
> Could we start with just the cluster scoped `StoreConfig` and wait for user feedback before adding a namespace scoped equivalent?
> 
> > Please see my response here.
> 
> I understand your concerns, but please let's start with provider scoped StoreConfigs. As you mention, we can always switch to a single one based on user feedback during the alpha. As much as I agree with the UX concerns I'm more concerned with tight coupling.

Alright, I'll update the design accordingly.

design/design-doc-external-secret-stores.md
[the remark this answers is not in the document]
turkenh: Done.

design/design-doc-external-secret-stores.md
```diff
@@ -0,0 +1,632 @@
...
+```yaml
+spec:
+  publishConnectionDetailsToStore: vault-production
+```
+
+**Claim** resources requires some additional logic since they are namespaced but
+need a Cluster scoped `StoreConfig`. They will always use the `default`
+`StoreConfig` and if it is configured for local Kubernetes, secret will be
+written into the **Claim** namespace instead of namespace configured in `default`
+`StoreConfig`.
```
turkenh, line 235: @negz this is the best way I could come to handle referring a cluster scoped `StoreConfig` from a namespaced `Claim` and I cannot say that I am happy with it.

Apart from having some implicit logic that differs for Kubernetes and other secret stores, the main limitation of this approach is we are losing the scoping done at the namespace level. As a concrete example, if the `default` store is configured as `Vault`, secrets in namespace `team-1` and `team-2` would conflict. 

I believe we could solve this by introducing a _scope_ concept to Store implementations. Usage of scope would differ for each store implementation, for example, for Kubernetes it will just be used as the namespace of the secret but for vault, it will be appended to the parentPath of the secret. This way, we could define a common `defaultScope` parameter for `StoreConfig` resource which would be used whenever a scope is not available (i.e. cluster scoped).

```
apiVersion: secrets.crossplane.io/v1alpha1
kind: StoreConfig
metadata:
  name: default
spec:
  type: Kubernetes
  kubernetes:
    ...
```

```
apiVersion: secrets.crossplane.io/v1alpha1
kind: StoreConfig
metadata:
  name: default
spec:
  type: Vault
  vault:
    parentPath: "secret/my-cloud/dev/"
    ...
```

or we can just name is a `defaultNamespace`. What do you think?

turkenh (commented):
@negz I've updated the design as you suggested.

design/design-doc-external-secret-stores.md
```diff
@@ -0,0 +1,632 @@
...
+```
+
+Example StoreConfig:
+
+```yaml
+apiVersion: secrets.crossplane.io/v1alpha1
+kind: StoreConfig
+metadata:
+  name: kubernetes-cluster-1
+  namespace: team-a
```
negz, line 462: Nit: we're not going to have namespaced StoreConfig types (yet).

design/design-doc-external-secret-stores.md
negz: I think your proposal sounds good.

My only question would be whether we're confident every secret store we'll support has a concept of scope? If there was any concern about that we could move the `defaultScope` into the more specific store type blocks, like:

```yaml
apiVersion: secrets.crossplane.io/v1alpha1
kind: StoreConfig
metadata:
  name: default
spec:
  type: Kubernetes
  kubernetes:
    # Used for cluster scoped types. Namespaced types always use their own namespace.
    defaultNamespace: crossplane-system
```

```yaml
apiVersion: secrets.crossplane.io/v1alpha1
kind: StoreConfig
metadata:
  name: default
spec:
  type: Vault
  vault:
    parentPath: "secret/my-cloud-dev"
    # Used for cluster scoped types. Namespaced types always use their own namespace.
    defaultParentPathSuffix: crossplane-system
```

I don't feel strongly about this though - if you're confident scope will always be relevant making it applicable to all secret stores makes sense. I suppose even if a secret store didn't have a native concept of scope we could always prefix it to the secret name in that store, so name the secret something like `crossplane-system-foo`.

negz (approved):
This looks ready to build to me @turkenh! I'm excited to see it happen. Thank you for persisting with this - I know my feedback loop has been very slow. I dropped a few final thoughts and comments, but none of them need to block this design. Let me (or another maintainer) know when you're ready for this to be merged.

design/design-doc-external-secret-stores.md
turkenh: Yeah that's a good point. I believe any secret store we are adding support for, needs to somehow handle secrets coming from multiple namespaces with some sort of scoping which could be just prefixing to secret name as you suggested. So, I will add a `spec.defaultScope` to `StoreConfig`.

design/design-doc-external-secret-stores.md
```diff
@@ -0,0 +1,632 @@
...
+A `StoreConfig` named `default` will be created during installation which is
+configured to write secret to a Kubernetes cluster in the Crossplane
+installation namespace. 
+
+`StoreConfig.configRef` will be an optional parameters and **in its absence**:
+
+- For cluster-scoped resources (e.g. `MR` and `XR`), it will be late initialized
+as `configRef.name=default`.
+- For namespaced resources (e.g. `XRC`), secrets published to the same
+namespace as the resource.
```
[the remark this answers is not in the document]
turkenh: Yes, reworded was a bit messed up.

design/design-doc-external-secret-stores.md
```diff
@@ -0,0 +1,632 @@
...
+      name: vault-dev
+```
+
+Publish Connection details to a _Kubernetes_ secret with some labels/annotations:
+
+```yaml
+spec:
+  publishConnectionDetailsTo:
+    name: my-db-connection
+    attributes:
```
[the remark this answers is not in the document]
turkenh: `attributes` applies for any StoreConfig and expects arbitrary key/value pairs and not only for Kubernetes.

design/design-doc-external-secret-stores.md
```diff
@@ -0,0 +1,632 @@
...
+
+Publish Connection details to an _AWS Secret Manager_ secret with some tags:
+
+```yaml
+spec:
+  publishConnectionDetailsTo:
+    name: my-db-connection
+    attributes:
+      tags:
+        environment: production
```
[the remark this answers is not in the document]
turkenh: Hmm, I was always thinking that the author of this spec would be aware of the secret store to be used, but this might not be the case 🤔 But not sure how to support cases like [this](https://github.com/crossplane/crossplane/pull/2494#discussion_r754145633) in a generic way where we need to configure multiple attributes (labels/annotations/types) of the secrets. 
One solution just coming to my mind is, instead of arbitrary key value pairs we stick to Kubernetes ones, but each store config implementation maps them internally.

For example, for the following updated spec:

```
spec:
  publishConnectionDetailsTo:
    name: my-db-connection
    labels:
      environment: production
    annotations:
      tekton.dev/git-0: https://github.com
    type: kubernetes.io/basic-auth
    configRef:
      name: default
```

AWS secret store implementation would use `labels` as `tags` but ignore others (annotations and type).
What do you think?

design/design-doc-external-secret-stores.md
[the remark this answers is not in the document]
turkenh: See https://github.com/crossplane/crossplane/pull/2494#discussion_r770470777.

design/design-doc-external-secret-stores.md
```diff
@@ -0,0 +1,632 @@
...
+
+**Examples:**
+
+**Composition** configuring to publish to default Store which is created during
+installation time, i.e. publish to a Kubernetes secret in the namespace where
+crossplane installed.
+
+```yaml
+spec:
+  publishConnectionDetailsToStore: default
```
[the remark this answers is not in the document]
turkenh: I prefer it to be a bit long instead of having it nested, updated.

design/design-doc-external-secret-stores.md
```diff
@@ -0,0 +1,632 @@
...
+    configRef:
+      name: aws-secret-manager-prod
+```
+
+#### Separation of Concerns: Compositions, Composites and Claims
+
+Currently, platform operators could specify where should composite resource
+secrets land using the partial input `writeConnectionSecretsToNamespace`. We
+will similarly deprecate this field in favor of
+`publishConnectionDetailsToStore`.
```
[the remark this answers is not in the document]
turkenh: Sounds reasonable to me, but I would prefer to update the design later (e.g. during implementation).

design/design-doc-external-secret-stores.md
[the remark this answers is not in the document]
turkenh: If we agree on this, I think I will also rename `defaultScope` as `defaultNamespace` and it would be up to store config implementation how to map namespace in k8s.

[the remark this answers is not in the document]
negz: Those both sound good to me. I'd suggest moving labels and annotations out from under attribs to the same level as name for maximum symmetry. Or rename attribs to 'metadata'.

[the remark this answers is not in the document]
turkenh: Renamed `attributes` as `metadata` 👍

design/design-doc-external-secret-stores.md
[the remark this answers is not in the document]
turkenh: Created a ticket to follow up: https://github.com/crossplane/crossplane/issues/2774

turkenh:
@negz this should be ready for merge, thanks a lot for all your feedback so far!

Issue #2493, opened 2021-08-18
Highlight the 'current' composition revision


negz:
### What problem are you facing?

We recently added support for a new feature we call 'composition revisions'. This feature is in alpha (behind a feature flag), and allows platform teams and their customers to roll out new compositions at a controlled pace.

#2437 - One pager
#2451 - Implementation

Each unique new state of a `Composition` triggers the creation of a `CompositionRevision` that an XR may consume. Revisions are numbered monotonically. When a `Composition` returns to a previous state we do not create a new revision, which makes it difficult to discern which revision matches the 'latest' or 'current' state of the `Composition`.

### How could Crossplane help solve your problem?

We should add a field to the `CompositionRevision` type (probably in the status) that reflects whether or not it is currently... current. This field would be exposed as a printer column to make it easy to identify the current revision when running `kubectl get compositionrevisions`.

PR #2500, opened 2021-08-19
Indicate which `CompositionRevision` is 'current'

negz:

Fixes https://github.com/crossplane/crossplane/issues/2493

The current CompositionRevision is the one with a spec that matches the Composition's spec at any point in time. There should only ever be one current CompositionRevision, and it won't necessarily be the one with the highest revision number, e.g. if the Composition has reverted to a state that matches a prior revision.

I have:

I've tested this by applying the "getting started with GCP" Composition from our documentation, and editing it a few times. Each time I edit I see that the `CURRENT` printer column changes to match the revision I would expect to be current. This is true even when I make edits that 'revert' to an older revision rather than creating a new one.

```console
$ kubectl get compositionrevision
NAME                                                          REVISION   CURRENT   AGE
compositepostgresqlinstances.gcp.database.example.org-2hql4   3          False     53s
compositepostgresqlinstances.gcp.database.example.org-2qtx4   1          True      14m
compositepostgresqlinstances.gcp.database.example.org-xmm7h   2          False     5m10s
```


hasheddan (approved):
LGTM, thanks @negz! I briefly considered whether this should be `Active` instead of `Current` to match package revisions, but then realized this is representing a state that is pretty different from our `Active` assertion for package revisions so I like that the name is different / it is represented in conditions rather than spec 👍🏻

PR #2503, opened 2021-08-19
Do not requeue on successful revision reconcile

### LINKED ISSUE #2449: Too many update calls for resource customresourcedefinitions
### What happened?
We have recently installed crossplane Things are working fine. The only observation is crossplane making around 178,554 update call on resource `customresourcedefinitions` in 24 hrs. Is this expected? 

crossplane-update

### How can we reproduce it?
Enable k8s audit log and check how many `update` call are made by crossplane for `customresourcedefinitions`

Crossplane version: v1.3.0

* Cloud provider: aws
* Kubernetes version: 1.19.4

hasheddan:

This is a fairly substantial change in the package revision controller
behavior, but should drastically reduce the number of calls we make to
the API Server. We previously reconciled package revisions every minute
after a successful reconcile, but we now opt for only requeing if a
reconcile is not successful. This means that installed resources could
drift from what is specified in the package, but the odds of this
causing negative impact on an environment are somewhat mininal.

We also update to watch owned Deployments for a ProviderRevision as we
need to be notified immediately if a provider Deployment is unhealthy.
This does have the consequence that we will update all installed
resources on a regular cadence in the case that a provider Deployment
remains unhealthy. That being said, an unhealthy provider Deployment
should call for urgent intervention, either by the controller or the
cluster administrator.

~A potential negative side-effect of not regularly reconciling
ProviderRevisions is that an updated ControllerConfig will not flow
through to the provider Deployment within a minute, which was the
previous guarantee. That being said, any update to the Provider itself
will trigger a reconcile, which would cause ControllerConfig changes to
also be picked up.~

Update: we are now enqueing reconciles for ProviderRevisions when
an event on a ControllerConfig it references is observed.

Fixes #2449 

_Note: I could go either way on this being backported. Typically I would not want to backport a behavior change, but this feels like it could be considered a "bug fix" given that the excessive API calls are causing issues for folks. Because I don't foresee if having negative impact on any existing users, I have opted to backport._

I have:

Tested this by installing `provider-aws` and the `getting-started-with-gcp` Configuration. Once resources were established and `Deployment` was healthy, I observed no further reconciles taking place. I also observed that when a `Deployment` became unhealthy, we reconcile on our short wait cadence (i.e. `30s`) until it becomes healthy again, at which point we cease requeing.

### DIFF
--- a/internal/controller/pkg/revision/watch.go
+++ b/internal/controller/pkg/revision/watch.go
@@ -1,0 +1,87 @@
+/*
+Copyright 2021 The Crossplane Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package revision
+
+import (
+	"context"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/workqueue"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	v1 "github.com/crossplane/crossplane/apis/pkg/v1"
+	"github.com/crossplane/crossplane/apis/pkg/v1alpha1"
+)
+
+type adder interface {
+	Add(item interface{})
+}
+
+// EnqueueRequestForReferencingProviderRevisions enqueues a request for all
+// provider revisions tha reference a ControllerConfig when the given
+// ControllerConfig changes.
+type EnqueueRequestForReferencingProviderRevisions struct {
+	client client.Client
+}
+
+// Create enqueues a request for all provider revisions that reference a given
+// ControllerConfig.
+func (e *EnqueueRequestForReferencingProviderRevisions) Create(evt event.CreateEvent, q workqueue.RateLimitingInterface) {
+	e.add(evt.Object, q)
+}
+
+// Update enqueues a request for all provider revisions that reference a given
+// ControllerConfig.
+func (e *EnqueueRequestForReferencingProviderRevisions) Update(evt event.UpdateEvent, q workqueue.RateLimitingInterface) {
+	e.add(evt.ObjectOld, q)
+	e.add(evt.ObjectNew, q)
+}
+
+// Delete enqueues a request for all provider revisions that reference a given
+// ControllerConfig.
+func (e *EnqueueRequestForReferencingProviderRevisions) Delete(evt event.DeleteEvent, q workqueue.RateLimitingInterface) {
+	e.add(evt.Object, q)
+}
+
+// Generic enqueues a request for all provider revisions that reference a given
+// ControllerConfig.
+func (e *EnqueueRequestForReferencingProviderRevisions) Generic(evt event.GenericEvent, q workqueue.RateLimitingInterface) {
+	e.add(evt.Object, q)
+}
+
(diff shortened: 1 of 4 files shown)


hasheddan (commented):
I have some general concerns around the fact that we will still hit the API server a bunch of times in the event that the `Deployment` is unhealthy or a case where we consistently had issues with "Object has been modified" errors -- i.e. I don't love that we have to check _all_ of the installed resources for a package revision every time we reconcile as it means any reconcile is fairly expensive (depending on the number of resources in the package). That being said, I am also not a fan of conditionally checking resources on reconcile because I want to preserve our guarantee that we won't create a `Deployment` for a `Provider` unless all resources can be controlled, and conditionally checking resources would mean there could be a case in which we don't check, but also may create or update a `Deployment`.

turkenh:
> A potential negative side-effect of not regularly reconciling
ProviderRevisions is that an updated ControllerConfig will not flow
through to the provider Deployment within a minute, which was the
previous guarantee. That being said, any update to the Provider itself
will trigger a reconcile, which would cause ControllerConfig changes to
also be picked up.

@hasheddan Isn't it an option to watch ControllerConfig objects as well to fix this?

> I have some general concerns around the fact that we will still hit the API server a bunch of times in the event that the `Deployment` is unhealthy or a case where we consistently had issues with "Object has been modified" errors -- i.e. I don't love that we have to check _all_ of the installed resources for a package revision every time we reconcile as it means any reconcile is fairly expensive (depending on the number of resources in the package). That being said, I am also not a fan of conditionally checking resources on reconcile because I want to preserve our guarantee that we won't create a `Deployment` for a `Provider` unless all resources can be controlled, and conditionally checking resources would mean there could be a case in which we don't check, but also may create or update a `Deployment`.

In such a case, would we hit API server with exponential backoff? If that is the case, I wouldn't be concerned too much given it will be less frequent than the current case.

hasheddan:
> Isn't it an option to watch ControllerConfig objects as well to fix this?

@turkenh good point -- we could and use a predicate to only enqueue for `ProviderConfig` that reference a given `ControllerConfig` -- Will update 👍🏻

> In such a case, would we hit API server with exponential backoff? If that is the case, I wouldn't be concerned too much given it will be less frequent than the current case.

We actually won't hit backoff here because we aren't returning an error. We could opt to return an error, but would likely want to tune backoff because we don't want to retry a bunch of times immediately and reach max backoff in the case that a `Deployment` is restarting a `Pod` or something. I agree that this change is an improvement over status quo though. I am planning on looking into the backoff we use across all controllers as a separate effort.

internal/controller/pkg/revision/watch.go
```diff
@@ -0,0 +1,87 @@
+/*
```
hasheddan, line 1: Thank you @negz for basically writing all of this code for me :)

(copied with slight modifications from `EnqueueRequestForAllRevisionsWithRequests`)

hasheddan (commented):
@turkenh updated here if you want to take another look 👍🏻

internal/controller/pkg/revision/watch.go
```diff
@@ -0,0 +1,87 @@
...
+	v1 "github.com/crossplane/crossplane/apis/pkg/v1"
+	"github.com/crossplane/crossplane/apis/pkg/v1alpha1"
+)
+
+type adder interface {
+	Add(item interface{})
+}
+
+// EnqueueRequestForReferencingProviderRevisions enqueues a request for all
+// provider revisions tha reference a ControllerConfig when the given
```
turkenh, line 38: typo:
```suggestion
// provider revisions that reference a ControllerConfig when the given
```

turkenh (approved):
LGTM👍

internal/controller/pkg/revision/watch.go
```diff
@@ -0,0 +1,87 @@
...
+)
+
+type adder interface {
+	Add(item interface{})
+}
+
+// EnqueueRequestForReferencingProviderRevisions enqueues a request for all
+// provider revisions that reference a ControllerConfig when the given
+// ControllerConfig changes.
+type EnqueueRequestForReferencingProviderRevisions struct {
```
muvaf, line 40: I wonder if we can come up with something like [this](https://github.com/crossplane/crossplane-runtime/blob/master/pkg/resource/predicates.go#L34) (maybe one that accepts `add` function?) which would allow us not duplicate code but it's definitely non-blocking.

muvaf (approved):
LGTM! Thanks @hasheddan , I think the tradeoff makes sense here. IMO, it's fine to rely on apiserver eventing for resources that do not get actively changed by the user, like managed resources or composition.

Issue #2492, opened 2021-08-18
Document composition revisions


negz:
### What problem are you facing?

We recently added support for a new feature we call 'composition revisions'. This feature is in alpha (behind a feature flag), and allows platform teams and their customers to roll out new compositions at a controlled pace. At the time of writing the feature is not documented.

https://github.com/crossplane/crossplane/pull/2437 - One pager
https://github.com/crossplane/crossplane/pull/2451 - Implementation

### How could Crossplane help solve your problem?

We could document the new feature, as well as the newly added feature flag support.

PR #2510, opened 2021-08-21
Add a guide to CompositionRevisions

negz:

Fixes https://github.com/crossplane/crossplane/issues/2492

I imagine this will be rolled into the broader Composition documentation at some point, and probably framed more around the actual use case it solves, i.e. 'updating XRs' or 'evolving Compostions'. However for the time being while the feature is in alpha I'd like to keep this documentation separate and frame it as a guide to the specific feature.

I have:

With my eyeballs.

### DIFF
--- a/docs/guides/composition-revisions.md
+++ b/docs/guides/composition-revisions.md
@@ -1,0 +1,155 @@
+---
+title: Composition Revisions
+toc: true
+weight: 260
+indent: true
+---
+
+This guide discusses the use of "Composition Revisions" to safely make and roll
+back changes to a Crossplane [`Composition`][composition-type]. It assumes
+familiarity with Crossplane, and particularly with
+[Composition][composition-term].
+
+> Composition Revisions are an __alpha feature__. They are not yet recommended
+> for production use, and are disabled by default.
+
+A `Composition` configures how Crossplane should reconcile a Composite Resource
+(XR). Put otherwise, when you create an XR the selected `Composition` determines
+what managed resources Crossplane will create in response. Let's say for example
+that you define a `PlatformDB` XR, which represents your organisation's common
+database configuration of an Azure MySQL Server and a few firewall rules. The
+`Composition` contains the 'base' configuration for the MySQL server and the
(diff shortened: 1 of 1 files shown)


jbw976 (commented):
This is great to have @negz!  Also awesome live stream today :)

Is there any guidance or caveats we should also be giving users in this guide about safely updating compositions to get the intended result? e.g. are there known scenarios that the machinery doesn't handle well, or where the user should take extra steps to be careful?

docs/guides/composition-revisions.md
```diff
@@ -0,0 +1,155 @@
...
+`Composition` contains the 'base' configuration for the MySQL server and the
+firewall rules that is extended by the configuration fo the `PlatformDB`.
+
+There is a one-to-many relationship between a `Composition` and the XRs that use
+it. You might define a `Composition` named `big-platform-db` that is used by ten
+different `PlatformDB` XRs. Usually, in the interest of self-service, the
+`Composition` is managed by a different team from the actual `PlatformDB` XRs.
+For example the `Composition` may be written and maintained by a platform team
+member, while individual application teams create `PlatformDB` XRs that use said
+`Composition`.
```
hasheddan, line 30: I feel like it probably isn't worth introducing an additional concept here, but just wanted to note that we typically talk about app teams authoring XRCs rather than XRs directly. That being said, "creating an XR" could be thought of here in a general sense as "via a Claim, or directly". Don't think we need to make any change, just thought I'd mention :)

hasheddan (approved):

docs/guides/composition-revisions.md
negz: Yeah, I had a similar thought here. I figured it was better to keep it simple in this context. As you say, it's not incorrect - the creation is just indirect.

negz:
> Is there any guidance or caveats we should also be giving users in this guide about safely updating compositions to get the intended result? e.g. are there known scenarios that the machinery doesn't handle well, or where the user should take extra steps to be careful?

Nothing immediately comes to mind. I'm sure we'll find something eventually and can update appropriately.

PR #2525, opened 2021-08-24
Support for custom/untrusted package registries (fixes 2318)

### LINKED ISSUE #2318: Support package registries with custom certificate authorities

Currently, we use the `go-containerregistry` [remote pkg](https://github.com/google/go-containerregistry/tree/main/pkg/v1/remote) as our client to fetch package manifests and images from OCI registries. It uses the `http.DefaultTransport`, meaning that it loads the default system certificates. This precludes fetching packages from private registries that may use certs from a custom CA.

The default transport can [easily be overridden](https://github.com/google/go-containerregistry/blob/8b535fa90f33fcb5121d422e3a51b37e25e46096/pkg/v1/remote/options.go#L95) with one that passes additional certs to the `TLSConfig`. These could be provided by mounting them on the Crossplane `Pod`, then pointed at by passing a flag or environment variable. This would allow for custom certificate authorities to be trusted when a user is running their own.

zonybob:

Fixes #2318 

Allows a custom CA certificate bundle to be mounted from a ConfigMap (configurable via helm install) to enable fetching packages from registries with unknown/untrusted certificates.

**NOTE:**
This could have been resolved with zero code change and only helm changes by mounting an entire, large bundle over `/etc/ssl/certs/ca-certificates.crt` from the distroless image, but the preference in #2318 seemed to be "extend" the root CAs rather than "overwrite". The "overwrite" option would have sufficed for my needs, so let me know if you prefer the "no-code-change" option.

I have:

This code was tested in 2 ways:

1. Using `make run` in the local `kind` cluster, a configuration package was pulled from a localhost `registry:2` container with a self-signed certificate by having run with `CA_BUNDLE_PATH=/path/to/registry.crt` 
2. The updated helm chart was published to my own s3 bucket and then installed on an OpenShift cluster. I then set up a public-facing `registry:2` container with a self-signed certificate and observed a configuration package could be pulled if the chart was deployed with the proper config (pulling the `ca.crt` from a named `ConfigMap`)

I did not update unit tests (yet). I was unsure how/if to do that for this particular change. I was looking for any other tests for `fetch.go` and did not see any. I'm open to suggestions!


negz (commented):
This broadly looks good to me, though @hasheddan is the expert on this part of the codebase. Unfortunately we just missed the feature freeze for v1.4 so we won't be able to ship this until v1.5. That said, this is a super simple change so I could be convinced to make (our first?) exception.

zonybob:
@negz thanks for the feedback...
I pushed changes to pass/wrap errors from CA parsing (rather than trying to log).
Having this sneak into 1.4.0 would be fantastic for my own timeline/requirements, but I do understand the importance of freezes. Let me know either way if there's anything else you need from me.

internal/xpkg/fetch.go
```diff
@@ -38,14 +44,46 @@ type Fetcher interface {
...
+		return nil, errors.Wrapf(err, "Failed to append %q to RootCAs", caBundlePath)
+	}
+
+	// Append our cert to the system pool
+	if ok := rootCAs.AppendCertsFromPEM(certs); !ok {
+		return nil, errors.Errorf("No certificates could be parsed from %q", caBundlePath)
+	}
+
+	var customCATransport http.RoundTripper = http.DefaultTransport
+	customCATransport.(*http.Transport).TLSClientConfig = &tls.Config{RootCAs: rootCAs, MinVersion: tls.VersionTLS12}
```
hasheddan, line 80: @zonybob while I think encountering it would be an edge case (because hopefully any registry being used is on at least TLS 1.2), I am somewhat inclined to just fallback to default TLS version min defined in `tls` pkg (which is currently `1.0`). Do you see a compelling reason to not leave this to std library?

internal/xpkg/fetch.go
```diff
@@ -38,14 +44,46 @@ type Fetcher interface {
 type K8sFetcher struct {
 	client    kubernetes.Interface
 	namespace string
+	transport http.RoundTripper
 }
 
 // NewK8sFetcher creates a new K8sFetcher.
-func NewK8sFetcher(client kubernetes.Interface, namespace string) *K8sFetcher {
+func NewK8sFetcher(client kubernetes.Interface, namespace, caBundlePath string) (*K8sFetcher, error) {
```
hasheddan, line 51: nitpick: since the `caBundlePath` is optional and I could also see additional optional features being added here, I wonder if it would make sense to do something like this (as an example):
```

type FetcherOpt func(k *K8sFetcher) error

func WithCustomCA(path string) FetcherOpt {
	return func(k *K8sFetcher) error {
		// Get the SystemCertPool, continue with an empty pool on error
		rootCAs, _ := x509.SystemCertPool()
		if rootCAs == nil {
			rootCAs = x509.NewCertPool()
		}

		// Read in the cert file
		certs, err := ioutil.ReadFile(filepath.Clean(path))
		if err != nil {
			return errors.Wrapf(err, "Failed to append %q to RootCAs", path)
		}

		// Append our cert to the system pool
		if ok := rootCAs.AppendCertsFromPEM(certs); !ok {
			return errors.Errorf("No certificates could be parsed from %q", path)
		}

		// NOTE: could optionally just require that k.transport is
		// *http.Transport, but that would restrict adding WithRoundTripper
		// option in the future unless package was modified. Or just override
		// the transport entirely as you were already doing :)
		t, ok := k.transport.(*http.Transport)
		if !ok {
			return errors.New("Fetcher transport is not an HTTP transport")
		}
		t.TLSClientConfig = &tls.Config{RootCAs: rootCAs, MinVersion: tls.VersionTLS12}
		return nil
	}
}

// NewK8sFetcher creates a new K8sFetcher.
func NewK8sFetcher(client kubernetes.Interface, namespace string, opts ...FetcherOpt) (*K8sFetcher, error) {
	k := &K8sFetcher{
		client:    client,
		namespace: namespace,
		transport: http.DefaultTransport,
	}

	for _, o := range opts {
		if err := o(k); err != nil {
			return nil, err
		}
	}

	return k, nil
}
```
Would also allow for adding some unit tests to `NewK8sFetcher` (though they would be fairly trivial at this point) as well as potentially the `WithCustomCA` option if we wanted to pass in the system cert pool / additional certs instead of parsing them in function body (I imagine in that case we would want to parse the certs as early as possible, i.e. pass them down through setup functions, which would avoid parsing them for each controller that requires the k8s fetcher). Anyway, let me know if you have any thoughts on this pattern!

hasheddan (commented):
@zonybob thanks so much for implementing this! It will unfortunately have to wait for v1.5, but I'm sure it'll be a welcome addition for many users! Left a few comments below, none of them critical, but could be options for slightly adjusting implementation.

hasheddan:
@zonybob are you still interested in driving this work to completion? This would be a much appreciated feature :)

zonybob:
@hasheddan in reference to tls version... the lint/gosec process is not happy with removing that min tls version...
```
internal/xpkg/fetch.go:83:24: G402: TLS MinVersion too low. (gosec)
                t.TLSClientConfig = &tls.Config{RootCAs: rootCAs}
                                     ^
```

Would you prefer that I keep the minTLS version in there? or have gosec ignore that via `// #nosec G402` comment?

PR #2536, opened 2021-08-28
Design doc for generated providers that are based on Terraform

muvaf:

This design doc has been in draft state for a while but the implementation has already started. Code generation tools and common utilities are in [terrajet](https://github.com/crossplane-contrib/terrajet) repository. [`provider-tf-aws`](https://github.com/crossplane-contrib/provider-tf-aws) is the first terraform-based provider that is compatible with (most) Crossplane Resource Model requirements.

I have:

N/A

### DIFF
--- a/design/design-doc-terrajet.md
+++ b/design/design-doc-terrajet.md
@@ -507,6 +507,20 @@ exceptional cases that need to be handled and customizations that should be
 implemented per resource. Terraform community has done that for years with
 hand-crafted code and no other tool has as much coverage as Terraform.
 
+## Prior Art
+
+There are many projects in infrastructure space that builds on top of Terraform.
+Each of the projects have their own limitations, additional features and different
+license restrictions.
+
+* [Crossplane: Terraform Provider Runtime](https://github.com/crossplane/crossplane/blob/e2d7278/design/design-doc-terraform-provider-runtime.md)
+* [Crossplane: provider-terraform](https://github.com/crossplane-contrib/provider-terraform)
+* [Hashicorp Terraform Cloud Operator](https://github.com/hashicorp/terraform-k8s)
+* [Rancher Terraform Controller](https://github.com/rancher/terraform-controller)
+* [OAM Terraform Controller](https://github.com/oam-dev/terraform-controller)
+* [Kubeform](https://github.com/kubeform/kubeform)
+* [Terraform Operator](https://github.com/isaaguilar/terraform-operator)
+
 [jsoniter]: https://github.com/scoutapp/jsoniter-go
 [json-multipletagkey]: https://github.com/scoutapp/jsoniter-go/blob/ca39e5a/config.go#L22
 [schema-schema]: https://github.com/hashicorp/terraform-plugin-sdk/blob/9321fe1/helper/schema/schema.go#L37
--- a/design/design-doc-terrajet.md
+++ b/design/design-doc-terrajet.md
@@ -114,7 +114,9 @@ CLI.
 Additionally, the mechanics of providers naturally include core Terraform code
 structs that we'd need to develop logic to work on and that'd increase
 our bug surface since they are not intended to be called by something other than
-Terraform CLI code.
+Terraform CLI code. This fact has some consequences around how the authors of the
+providers organize their code as well. For example, Azure provider stores everything
+under `internal` package which would force us to fork the provider.
 
 ### Talk to Provider Server
 
@@ -178,6 +180,12 @@ For example, [`ResourceDiff`][resource-diff] struct here has many fields and
 functions that are not exposed. So the implementation we'd be doing if we went
 with other two options will be very similar to how Terraform CLI implemented these
 functions already.
+
+Additionally, we see that some providers hide everything under `internal` package.
+This makes it essential to fork the code for the importing option. In other options,
+we still need to import the provider to get the schema to be used for generating
+the CRD types, but that happens during development time, so we can have some
+workarounds making the generator access the schema.
 
 #### Optionality for the Future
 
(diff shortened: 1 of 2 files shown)


design/design-doc-terrajet.md
```diff
@@ -0,0 +1,526 @@
...
+diff'ing, importing, conversion of JSON/HCL to `cty` and such. Providers do give
+you the customization functions whenever needed, but the main pipeline of execution
+lives in Terraform CLI. So, we would have to reimplement those parts of Terraform
+CLI.
+
+Additionally, the mechanics of providers naturally include core Terraform code
+structs that we'd need to develop logic to work on and that'd increase
+our bug surface since they are not intended to be called by something other than
+Terraform CLI code.
+
```
displague, line 118: Some providers have begun moving their implementation to `internal/` directories. This may make importing provider go code more complex.

design/design-doc-terrajet.md
```diff
@@ -0,0 +1,526 @@
...
+well-tested by the community, and not spend cycles.
+
+#### Unfriendly Programming Interfaces
+
+Since the providers are designed to be consumed only by Terraform CLI, they are
+designed for its specific needs and whatever the CLI doesn't need is not exposed.
+For example, [`ResourceDiff`][resource-diff] struct here has many fields and
+functions that are not exposed. So the implementation we'd be doing if we went
+with other two options will be very similar to how Terraform CLI implemented these
+functions already.
```
displague, line 180: It's also worth pointing out somewhere in this design that one of the principles in Crossplane's providers is a faithful representation of the external provider API.  

By adopting Terraform APIs, intermediary purpose-driven APIs are adopted, with the inclusion of behavioral fields that may (?) operate differently when executed in Crossplane.

negz: > By adopting Terraform APIs, intermediary purpose-driven APIs are adopted, with the inclusion of behavioral fields that may (?) operate differently when executed in Crossplane.

I think I follow what you're saying here, but could you give an example or two?

design/design-doc-terrajet.md
muvaf: Yeah we noticed that with @ulucinar , Azure provider keeps it all under `internal/`. Though I didn't know it was a trend.

We still do some importing to get the schema to build CRD types but that happens in build time so we can put some hacks in place to make it work, like clone-rename to pkg-use.

design/design-doc-terrajet.md
displague: The Crossplane interface to Terraform-derived providers will not be a faithful representation of the cloud provider API. Rather they will be faithful representations of the Terraform provider API.

The update of `ForceNew` behavior of a resource could be toggled by resource fields:
- https://registry.terraform.io/providers/hashicorp/aws/latest/docs/guides/resource-tagging#ignoring-changes-in-all-resources
- `instance_interruption_behavior` (as another example) https://github.com/hashicorp/terraform-provider-aws/blob/687cdcaa7a8c1b6f11e2f63431014533dd25aaec/aws/resource_aws_spot_instance_request.go#L140-L158

These fields are not part of the AWS API, they are helpers to achieve the desired Terraform behavior. How should this design doc address the shift from faithful API representation and how will this design be impacted or aided by these types of fields.

negz: Got it. My hunch is that we're effectively just going to 'leak' these Terraform details in our representation of cloud APIs. That's okay for the alpha stage, though eventually I suspect it will be a priority for the Terraform-derived MRs to match our 'native' MRs.

@muvaf perhaps we should consider a manually created filter of fields that don't map to the upstream APIs?

negz (approved):
I reviewed an earlier version of this design out of band with @muvaf, so I'm happy to approve this without comment. Let's leave this open for a bit to get community input though.

design/design-doc-terrajet.md
muvaf: We will have a way for provider authors to provide exception information here, like removal/addition of certain fields, manipulation of field tags/comments etc. So, this would be addressed by the provider authors, unfortunately, per resource if it's not generic enough. @displague thanks for the detailed explanation here, I'll add a section addressing this concern.

muvaf: Done!

design/design-doc-terrajet.md
```diff
@@ -413,6 +413,22 @@ and the sensitive attributes will be stored in the connection detail secret of t
...
+does that for most resources, there are occasionally fields in the Terraform schema
+of the resource that configures the behavior of Terraform execution rather than
+only the request made to the provider API. For example, `force_destroy` is a field
+in S3 Bucket Terraform schema but there is no corresponding API call - Terraform
+just deletes all objects in the bucket before calling deletion.
+
+For such cases, we'll expose configuration points for provider authors to provide
+per-resource exception information. They will be able to remove/add specific fields,
+manipulate their JSON tags or code comments. Since authors will call the pipeline
+in a Go context, they will be able to reuse same exception information for many
```
displague, line 429: I'm not sure if you meant `the same exception information` or `some exception information`. Subtle difference.

muvaf: I meant the cases where there is something common in many resources that you'd like to have the exception at provider level. For example, we add `region` to all AWS resources top-level schema.

design/design-doc-terrajet.md
```diff
@@ -0,0 +1,562 @@
+# Terrajet: Code Generation for Terraform-based Providers
+* Owner: Muvaffak Onus (@muvaf)
+* Reviewers: Crossplane Maintainers
+* Status: Draft
```
negz, line 4: @muvaf Are you feeling ready to merge this? If so, please change this to approved and merge away.

PR #2537, opened 2021-08-29
Promote `Lock` to `v1beta1` and mark `v1alpha1` as deprecated

### LINKED ISSUE #2469: Promote `Lock` to v1beta1

The (package) `Lock` type was added back in [v1.0.0](https://github.com/crossplane/crossplane/releases/tag/v1.0.0). We need to make a decision on whether it is ready to promote to v1beta1.

We should determine what we want to do with `Lock` - e.g. update it to v1beta1 as is, remove it, or leave it as is. If we do decide to leave it as is it we should define what we would want from the type in order for it to reach v1beta1.

hasheddan:

Introduces the `v1beta1` `Lock` API and marks the `v1alpha1` version as deprecated. The `v1alpha1` version is scheduled to be removed in the `v1.7` Crossplane release.

Fixes #2469 

I have:

Tested by installing Crossplane `v1.3.1`, installing `xp/getting-started-with-aws:v1.3.1`, then upgrading to the build from this PR and verifying successful CRD install and no disruption of service.


negz:
Mostly LGTM, but I think we should also update the controller code to use the new `v1beta1` type.

hasheddan:
@negz will do 👍🏻

hasheddan (commented):
@negz updated!

negz (approved):

Issue #2523, opened 2021-08-23
Cut v1.4.0 on August 31, 2021.


negz:
### Steps

This issue can be closed when we have:

- [x] Created the [release branch][release-branch].
- [x] Created and merged an empty commit to the `master` branch ([PR][empty-commit-pr]).
- [x] Run the [Tag workflow][rc-tag-workflow] on the `master` branch with the next release candidate tag ([Tag][rc-tag]).
- [x] Updated the current release version in the [Crossplane docs website repo] ([link][website-pr]).
- [x] Updated all version information in the documentation on the relevant release branch ([link][docs-update-pr]).
- [x] Run the [Tag workflow][tag-workflow] on the relevant release branch with the proper release version ([link][tag]).
- [x] Run the [CI workflow][ci-workflow] on the release branch and verified that the tagged build version exists on the [releases.crossplane.io] `build` channel. ([link][release-build]).
- [x] Run the [Configurations workflow][configurations-workflow] on the release branch and verified  that version exists on [registry.upbound.io] for all getting started packages.
    - [x] `xp/getting-started-with-aws` ([link][xp/getting-started-with-aws])
    - [x] `xp/getting-started-with-with-vpc` ([link][xp/getting-started-with-aws-with-vpc])
    - [x] `xp/getting-started-with-azure` ([link][xp/getting-started-with-azure])
    - [x] `xp/getting-started-with-gcp` ([link][xp/getting-started-with-gcp])
- [x] Run the [Promote workflow][promote-workflow] with channel `stable` on the release branch and verified that the tagged build version exists on the [releases.crossplane.io] `stable` channel ([link][stable-build]).
- [x] Published a [new release] for the tagged version, with the same name as the version and descriptive release notes ([link][release]).
- [x] Updated the [releases table] in the `README.md` on `master` ([link][release-table-pr]).
- [x] Ensured that users have been notified of the release on all communication channels:
    - [ ] Slack ([link][slack])
    - [ ] Twitter ([link][twitter])

[release-branch]: https://github.com/crossplane/crossplane/tree/release-1.4
[empty-commit-pr]: https://github.com/crossplane/crossplane/pull/2540
[rc-tag-workflow]: https://github.com/crossplane/crossplane/actions/runs/1186014852
[rc-tag]: https://github.com/crossplane/crossplane/releases/tag/v1.5.0-rc.0
[website-pr]: https://github.com/crossplane/crossplane.github.io/pull/120

[...]

PR #2540, opened 2021-08-31
Empty commit to advance rc tag.

muvaf:

Empty commit to advance rc tag.


hasheddan (approved):
xref https://github.com/crossplane/crossplane/issues/2523

Issue #2219, opened 2021-03-17
Consider allowing new package revisions to own objects even if they cannot control


hasheddan:
### What problem are you facing?

This feature is based on a scenario a community member recently shared with me. This write-up will hopefully illustrate some details of how the package manager works, as well as how to avoid any "catastrophe" situations.

The issue in this particular example was that the upgrade of `provider-aws` from `v0.16.0` to `v0.17.0` required manual intervention due to the dropping of support for an `alpha` CRD version as described in https://github.com/crossplane/crossplane/issues/2165. The inability for the `v0.17.0` `ProviderRevision` to gain control of CRDs meant that none of its CRDs were installed / updated, so the previous `ProviderRevision` (`v0.16.0`) was the only owner/controller of the `provider-aws` CRDs that were present in the cluster.

> Note: if a new revision cannot control _all_ of the CRDs it reconciles in the cluster, we don't let it control _any_. As you will see in this example, one option we could pursue is letting a new revision go ahead and _own_ the CRDs even if it can't establish control.

The community member decided at this point to try deleting the old `ProviderRevision` (`v0.16.0`) to try to fix the problem. This meant that the installed CRDs had only one owner reference (the `v0.16.0` `ProviderRevision`), and that object was now deleted, so Kubernetes marked all of the CRDs for [garbage collection](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/). CRDs with instances (CRs) existing in the cluster were not immediately deleted because the instances block owner deletion, and the instances themselves were being blocked on deletion by the presence of the [managed reconciler finalizer](https://github.com/crossplane/crossplane-runtime/blob/fe13ff9a1c2173dfb9d8db7dd4bbada7a33626d6/pkg/reconciler/managed/reconciler.go#L41). The finalizer was not being removed because the old `ProviderRevision` controller was stopped due to the fact that it had been transitioned to `Inactive` and the new `ProviderRevision` controller had not been started because it could not gain control of its objects. So the instances had a deletion timestamp set, but could not be cleaned up and were not being deleted externally.

[...]

PR #2547, opened 2021-09-01
Do not couple lifecycle of CRDs/XRDs to revisions

muvaf:

Per the discussion in https://github.com/crossplane/crossplane/issues/2219 , now all CRDs have two owner references: one controlling `ProviderRevision` and another one non-controlling `Provider` (or `Configuration` flavor). This allows users to keep CRDs/XRDs in the cluster even if `ProviderRevision`s are deleted. They will be gone only if you decide to delete `Provider`.

Fixes https://github.com/crossplane/crossplane/issues/2219

I have:

Manually. I've installed AWS Provider and validated that deleting `ProviderRevision` doesn't affect the existing CRDs; it only removes that `ProviderRevision` from the owner references list. Then new revision is created immediately and gets added in the place of the deleted one.

The previous behavior was that when you delete `ProviderRevision`, all CRDs would be gone, new revision would be created and CRDs would be installed again. Now they are only _updated_.

### DIFF
--- a/internal/controller/pkg/revision/establisher.go
+++ b/internal/controller/pkg/revision/establisher.go
@@ -18,10 +18,10 @@ package revision
 
 import (
 	"context"
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/pkg/errors"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -150,7 +150,7 @@ func (e *APIEstablisher) create(ctx context.Context, obj resource.Object, parent
 	// get deleted when the new revision doesn't include it in order not to lose
 	// user data, such as custom resources of an old CRD.
 	falseVal := false
-	refs := make([]v1.OwnerReference, len(parent.GetOwnerReferences())+1)
+	refs := make([]metav1.OwnerReference, len(parent.GetOwnerReferences())+1)
 	for i, ownerRef := range parent.GetOwnerReferences() {
 		nonControllerRef := ownerRef.DeepCopy()
 		nonControllerRef.Controller = &falseVal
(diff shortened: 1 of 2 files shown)


internal/controller/pkg/revision/establisher.go
```diff
@@ -146,15 +146,38 @@ func (e *APIEstablisher) Establish(ctx context.Context, objs []runtime.Object, p
 }
 
 func (e *APIEstablisher) create(ctx context.Context, obj resource.Object, parent resource.Object, opts ...client.CreateOption) error {
-	ref := meta.AsController(meta.TypedReferenceTo(parent, parent.GetObjectKind().GroupVersionKind()))
+	// We add the parent as `owner` of the resources so that the resource doesn't
+	// get deleted when the new revision doesn't include it in order not to lose
+	// user data, such as custom resources of an old CRD.
+	falseVal := false
+	refs := make([]metav1.OwnerReference, len(parent.GetOwnerReferences())+1)
+	for i, ownerRef := range parent.GetOwnerReferences() {
```
hasheddan, line 154: @muvaf do you think we should copy all owner refs instead of just the parent package? It can be identified via the `pkg.crossplane.io/package` [label we set on the revision](https://github.com/crossplane/crossplane/blob/2215d0bf53671c2477addee3072b6e59f7c57653/internal/controller/pkg/manager/reconciler.go#L337).

muvaf: Done!

apis/pkg/v1/interfaces.go
```diff
@@ -46,6 +53,18 @@ func RefNames(refs []corev1.LocalObjectReference) []string {
 	return stringRefs
 }
 
+// GetPackageOwnerReference returns the owner reference that points to the owner
+// package of given revision, if it can find one.
+func GetPackageOwnerReference(rev resource.Object) (metav1.OwnerReference, bool) {
```
muvaf, line 58: I thought of having `GetPackageOwnerReference` and `SetPackageOwnerReference` for convenience but realized we use `resource.Object` in the places I need these functions and didn't want to do a big change. So I opted for making it a helper function here.

apis/pkg/v1/interfaces.go
```diff
@@ -18,12 +18,19 @@ package v1
...
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
 	"github.com/crossplane/crossplane-runtime/pkg/resource"
 )
 
+const (
+	// ParentLabel is used as key for the owner package label we add to the
+	// revisions. Its corresponding value should be the name of the owner package.
+	ParentLabel = "pkg.crossplane.io/package"
```
negz, line 31: Nit: I believe our typical convention with these constants is to name them something like `LabelFoo` or `LabelKeyFoo` where `Foo` matches the actual label content. So this would be something like `LabelKeyPackage`. The rationale is that if we add more labels we can group them all together and they'll all start with `Label`.

apis/pkg/v1/interfaces.go
negz: I like this approach of using a helper function, but I'd prefer it to live near where it's used in the controller code rather than with the API type definitions.

negz (approved):

apis/pkg/v1/interfaces.go
muvaf: I felt like it could be useful in other contexts as well but we can move it there once we know that's needed. Just moved it to controller and added tests.

PR #2553, opened 2021-09-04
Fix helm chart init flags

### LINKED ISSUE #2549: Helm Chart passes wrong flags to install providers and configurations
### What happened?
I'm using crossplane installed using helm charts in version 1.3.1, I have upgraded crossplane to recent version 1.4.0 but it does not work. Crossplane init container fails with : 

```
crossplane: error: unknown flag --provider, did you mean "--providers"?
Usage: crossplane core init

Make cluster ready for Crossplane controllers.

Flags:
  -h, --help                       Show context-sensitive help.
  -d, --debug                      Print verbose logging statements.

      --providers=PROVIDERS,...
      --configurations=CONFIGURATIONS,...
 
```

I have GCP provider `crossplane/provider-gcp:v0.17.1`

### How can we reproduce it?

1. Create k8s cluster (for reproducing I used k3d for simplicity `k3d cluster create test`)
2. Install helm with 1.3.1 `helm install crossplane --namespace crossplane-system crossplane-stable/crossplane --set 'provider.packages={crossplane/provider-gcp:v0.17.1}'` 
3. Upgrade helm repo `helm repo update`
4. Upgrade crossplane with helm : `helm upgrade crossplane  crossplane-stable/crossplane --set 'provider.packages={crossplane/provider-gcp:v0.17.1}' --namespace crossplane-system`
5. Check new crossplane controller pod failing on init container.

Crossplane version: 
1.3.1 then. upgraded to  1.4.0

GKE (1.20.8-gke.2100) and local k3d

If flag indeed changed than in the current helm chart flag is still `--provider`

jgillich:
Fixes #2549

I have:

On an empty cluster:
```
make
helm install crossplane ./_output/charts/crossplane-1.5.0-rc.0.9.g99643d04.dirty.tgz  --set provider.packages[0]=crossplane/provider-helm:master,provider.packages[1]=crossplane/provider-kubernetes:main --set image.tag=master
```
-> check that crossplane was installed and providers are healthy.

Does the auto backport label work with the helm charts? Getting the existing chart updated would definitely come in handy

### DIFF
--- a/cmd/crossplane/core/init.go
+++ b/cmd/crossplane/core/init.go
@@ -31,8 +31,8 @@ import (
 
 // initCommand configuration for the initialization of core Crossplane controllers.
 type initCommand struct {
-	Providers      []string
-	Configurations []string
+	Providers      []string `name:"provider" help:"Pre-install a Provider by giving its image URI. This argument can be repeated."`
+	Configurations []string `name:"configuration" help:"Pre-install a Configuration by giving its image URI. This argument can be repeated."`
 }
 
 // Run starts the initialization process.


hasheddan (commented):
@jgillich thank you for fixing this! Looks like this got changed in https://github.com/crossplane/crossplane/pull/2401 because we switched from setting the flag names from `provider` and `configuration` directly to inferring their names from the init cmd struct fields. I am wondering if it may make sense to instead change the flag name back to what it was before, rather than change to the new name in the helm charts. It likely is not incredibly consequential because the helm chart is folks main interface to Crossplane, but I think I would prefer to make the flags aligned with what we had before so that someone upgrading from v1.3.x to v1.4.1 will not have the breaking flag name change at all. Would you be willing to go that route?

And the backport label should work for you when ready :)

jgillich:
@hasheddan Agreed! I've restored the original flags and their help text

```
Usage: crossplane core init

Make cluster ready for Crossplane controllers.

Flags:
  -h, --help                               Show context-sensitive help.
  -d, --debug                              Print verbose logging statements.

      --provider=PROVIDER,...              Pre-install a Provider by giving its image URI. This argument can be repeated.
      --configuration=CONFIGURATION,...    Pre-install a Configuration by giving its image URI. This argument can be repeated.
```

hasheddan (approved):
Thank you @jgillich!

PR #2108, opened 2021-02-01
Do not build with cloud provider keychains for pkg manager

hasheddan:
### Description of your changes

Currently Crossplane does not allow for configuring a pod to use the
cloud provider keychain to fetch package images. Including the
functionality results in confusing errors being logged by the
anonymously imported cloud provider credential providers in k8schain.
This updates the build to exclude modules.

Signed-off-by: hasheddan <georgedanielmangum@gmail.com>

In addition to this, we don't formally document the ability to use these credential providers, and they can cause issues, such as those described in https://github.com/kubernetes/kubernetes/pull/100686. `go-containerregistry` maintains a fork of these providers so that they don't have to take a dependency on k/k (https://github.com/vdemeester/k8s-pkg-credentialprovider), but this makes any modifications subject to long release cycles and lots of rounds of approval.

I would prefer to eliminate this dependency until we officially document support for it (if ever).

I have:

- [x] Read and followed Crossplane's [contribution process].
- [x] Run `make reviewable test` to ensure this PR is ready for review.

### How has this code been tested

Tested and confirmed that this eliminates the spurious logs _and_ eliminates issues with http client getting hung in network proxies.

[contribution process]: https://git.io/fj2m9

PR #2559, opened 2021-09-07
Reeanble K8s credential providers

### LINKED ISSUE #2561: Crossplane does not use K8s credential providers to pull packages
### What happened?

Since #2108 Crossplane does not use the credential providers for k8s in `go-contrainerregistry` that would use the identity of the kubelet to fetch packages from a private registry. A current workaround is using `packagePullSecrets` but that's quite cumbersome because it requires a uploading a token manually (at least for AWS).

The providers were deactivated because some issues that seem resolved now: https://github.com/kubernetes/kubernetes/pull/100686

Would be great to see them reactivated in Crossplane.

### How can we reproduce it?

Deploy Crossplane on an EKS cluster and let it fetch packages from a private ECR.

Crossplane version: 

Every version since 1.1

MisterMX:

Fixes #2561

This PR reactivates K8s keychain credential providers that have been disabled in #2108 because of some issues that have now been fixed.

This PR bumps `go-containerregistry` and its dependencies to the latest version.

I have:

Manually on an EKS cluster

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,8 +32,8 @@ GO_TEST_PARALLEL := $(shell echo $$(( $(NPROCS) / 2 )))
 GO_STATIC_PACKAGES = $(GO_PROJECT)/cmd/crossplane $(GO_PROJECT)/cmd/crank
 GO_LDFLAGS += -X $(GO_PROJECT)/internal/version.version=$(VERSION)
 GO_SUBDIRS += cmd internal apis
-# disables credential providers for pulling package images
-GO_TAGS += disable_gcp disable_aws disable_azure
+# uncomment to disable credential providers for pulling package images
+# GO_TAGS += disable_gcp disable_aws disable_azure
 GO111MODULE = on
 -include build/makelib/golang.mk
 


negz:
@MisterMX could you please update the PR description with some more context about the issues that were fixed (ideally a link to the issue).

hasheddan (commented):
This looks pretty good to me -- the problem that we had before was the Azure credential provider hanging forever with no timeout, which was subsequently fixed in k8s upstream (https://github.com/kubernetes/kubernetes/pull/100686). This has now also been reflected in the credential provider fork that go-containerregistry uses: https://github.com/vdemeester/k8s-pkg-credentialprovider/blob/master/azure/azure_acr_helper.go#L83

However, I do think there are still some unaddressed issues, such as the spurious errors in logs when this is enabled per observation in https://github.com/crossplane/crossplane/pull/2108. @MisterMX have you noticed any of those errors since re-enabling? I have a TODO in go-containerregistry to change how importing works with k8schain (https://github.com/google/go-containerregistry/issues/1042), so it may make sense to try and tackle that first.

That being said, I am happy to see this coming early in a release cycle as we will have more bake time before v1.5. I would love to get @seanmalloy and team to test this change out as they were the ones originally impacted by the missing timeout when running in a firewalled environment.

MisterMX:
I ran a test in an EKS environment with `--debug`. I haven't encountered any strange logs so far that were not Crossplane related.

MisterMX:
I rebased the branch on master and removed the changes to `go.mod` as they are not necessary for the credential providers to work.

hasheddan (commented):
@MisterMX I apologize for the long delay on this one -- I would still like to clean up the way k8schain works, but this is an active blocker for the community at this point so I don't want to block moving forward. Would you mind updating the comment as suggested and then I think we can get this in for v1.6 :)

MisterMX:
> Would you mind updating the comment as suggested

@hasheddan  Done

hasheddan (approved):
Thanks @MisterMX!

PR #500, opened 2019-06-01
Remove Provider validation checks, and thus Provider controllers

negz:
**Description of your changes**
Most resource controllers check that their Provider is 'valid' before attempting to connect to it. This `IsValid()` check simply reflects whether the Provider's `Ready` condition is currently true. The `Ready` condition is set by the Provider controllers, which make a naive request to the cloud provider to validate that their provider secrets are functional and mark the Provider CR ready if they are.

This approach is flawed in two ways.

Firstly it relies on the recorded rather than current actual state of the world to determine whether the provider is `Ready`. Recently we've preferred that controllers make decisions based on real rather that recorded observations. Put otherwise, a resource controller (e.g. the CloudMemorystore controller) should validate that its credentials work, not check whether the Provider controller indicated the credentials worked some time in the recent past.

Secondly, there's no way the Provider controller can ensure its credentials are valid for all the potential controllers that depend on it. Most providers test their validity by using their credentials to list buckets, or some similarly common managed resource. Unfortunately credentials being valid to list buckets does not mean they're valid to CRUD CloudMemorystore instances.

In short, it's preferable to determine whether credentials are valid for a particular task (i.e. CRUD-ing CloudMemorystore instances) by actually trying to use them to perform that task. Given this, I've removed the Provider controllers and their status objects in this commit.

**Checklist:**
- [ ] Documentation has been updated, if necessary.
- [ ] Pending release notes updated with breaking and/or notable changes, if necessary.
- [ ] Upgrade from previous release is tested and upgrade user guide is updated, if necessary.
- [x] Code generation (`make generate`) has been run to update object specifications, if necessary.
- [x] CRD manifests generation (`make manifests`) has been run to update CRD manifests yaml file specifications, if necessary.
- [x] Update dependencies (`make vendor`) has been run to update `Gopkg.lock` file, if necessary
- [ ] RBAC permissions in `clusterrole.yaml` have been updated to include new types, if necessary
- [x] All related commits have been squashed to improve readability.
- [x] Comments have been added or updated based on the standards set in [CONTRIBUTING.md](../blob/master/CONTRIBUTING.md#comments)

Issue #2411, opened 2021-06-29
Consider allowing provider partition on ProviderConfig


hasheddan:
### What problem are you facing?

As users continue to adopt Crossplane, installed Providers have become responsible for an increasing number of managed resources. As the number of managed resources climbs, Providers are more likely to run into resource constraints (i.e. CPU / memory), as well as controller rate-limiting, and any rate-limiting imposed by the external API. To address the first issue, infrastructure admins can always allocate more and more resources to the Provider `Deployment`, but the latter two issues are more complicated.

External rate limiting generally is based on factors such as the credentials being used to authenticate to the API, or characteristics of the machine making the request (e.g. IP address). Controller rate-limiting can be adjusted, but only at the per-controller and global level. We [have explored](https://github.com/crossplane/crossplane-runtime/issues/253#issue-830211294) exposing the polling interval for managed resources on their referenced `ProviderConfig`, which would make the drift detection frequency (not the rate-limiting) configurable, and we are [in the process](https://github.com/crossplane/provider-aws/pull/741) of exposing the global drift detection interval at the Provider `Pod` level. In practice, however, setting polling interval on the `ProviderConfig` as Providers currently are deployed today could become a rather complicated system where managed resources have many different polling intervals, but are all governed by the same global and per-controller rate-limiting (backoff) restrictions.

In addition to performance and latency implications, as Crossplane deployments scale within organizations, various permission models are required to satisfy the constraints of each organization. Having a single Provider `Pod` that reconciles all managed resources can make these constraints difficult to operate within, especially if consumers are using techniques such as IRSA or some other form of workload identity.

### How could Crossplane help solve your problem?

As dictated by [Provider conformance testing](https://github.com/crossplane/conformance/blob/fce75e1aed26a057d7fa166ba56787b8e9182f97/provider/provider_test.go#L197), any conformant Crossplane Provider must supply and support the usage of a `ProviderConfig`. As evidenced by some of the thinking around exposing the polling interval on `ProviderConfig`, as well as being the fact that a `ProviderConfig` is a set of, well, configuration for a Provider, it makes sense to use them as a unit of partitioning for Provider deployments.

This could be supported as an opt-in feature when installing a Provider, and Providers could even indicate whether they support the feature or not in their package manifest (i.e. `crossplane.yaml` -- this could be implicitly denoted by setting a Crossplane constraint to >= the version where this functionality was introduced). When a consumer opts-in to Provider partitioning, I imagine the functionality providing the following experience:

[...]

Issue #2122, opened 2021-02-08
Ability to deploy only the selected API groups of a provider


muvaf:
### What problem are you facing?

When you deploy a provider like provider-aws, CRDs of all groups are [deployed](https://github.com/crossplane/provider-aws/tree/483058c/package/crds) and their controllers are [started](https://github.com/crossplane/provider-aws/blob/483058c/pkg/controller/aws.go#L84). But if you'd like to use only IAM resources like `IAMRole`, `IAMPolicy` etc., you can't really do it.

One of the use cases that needs this granularity could be the use of granular IRSA. Today, you can choose how controller should authenticate to provider API by using `spec.providerConfigRef`, however, the controller pod itself has ability to use every auth mechanism that's defined, including IRSA, if it wants to. So, I wanted to open this issue to see how folks feel about this in terms of security.

### How could Crossplane help solve your problem?

We could deploy a separate provider instance for each group. For example:
```console
kubectl crossplane install crossplane/provider-aws:v0.15.0 \
  --groups rds.aws.crossplane.io,someothergroup.aws.crossplane.io
  # if --groups is not given, assume all
```

Then you can configure IRSA of this provider pod specifically using `ControllerConfig` and then if you'd like to deploy other groups with different IRSA configuration, you can do so. The resulting pods won't share the same cache and each controller manager would have a bit of overhead, so in terms of CPU/memory usage I'd expect higher numbers in total.

bassam:
it might be helpful to have a whitelist and blacklist of supported groups. This would enable folks to disable IAM for example.

muvaf:
We had a quick sync with @ulucinar about this issue. I believe the use cases (partition, hence stronger IRSA story) described in the issue will be mostly covered by https://github.com/crossplane/crossplane/issues/2411 . So, it would make sense to limit this to just being able to install a provider with limited set of resources with no guarantee about whether you can install another provider with another set or conflicting sets. The rough design is to have a new field under `Provider` resource:
```yaml
apiVersion: pkg.crossplane.io/v1
kind: Provider
metadata:
  name: provider-aws
spec:
  image: crossplane/provider-aws:v0.19.2
  apiGroups: # optional. if empty, it will work just like today.
  - ec2.aws.crossplane.io
  - rds.aws.crossplane.io
```

The `spec.apiGroups` would propagate to `ProviderRevision` and provider revision controller would install only allowed set and let the provider binary know about the list via a command line flag.

[...]

PR #2646, opened 2021-10-20
Add "enabledAPIs" to support selectively deploying package objects

### LINKED ISSUE #2649: API Server (and clients) becomes unresponsive with too many CRDs

As part of the ongoing [Terrajet-based] providers effort, we have observed that registering 100s of CRDs has some performance impacts on the cluster. Observations from two sets of experiments are described [here](https://github.com/crossplane-contrib/terrajet/issues/47#issuecomment-920316141) and [here](https://github.com/crossplane-contrib/terrajet/issues/47#issuecomment-920441882). As discussed in the results of the first experiment, Kubernetes scalability thresholds document currently does not consider #-of-CRDs (per cluster) as a dimension. However, `sig-api-machinery` folks suggest a maximum limit of 500 CRDs. And the reason for this suggested limit is not API call latency SLOs but rather, as we also identified in our experiments, due to the latency in the OpenAPI spec publishing. As the results of the second experiment demonstrate, the marginal cost of adding a new CRD increases as more and more CRDs exist in the cluster.

Although not considered as a scalability dimension officially yet, it looks like we need to be careful for the #-of-CRDs we install in a cluster. And with Terrajet-based providers we would like to be able to ship 100s of CRDs per provider package. Currently, for the initial releases of these providers, we are including a small subset (less than 10% of the total count) of all the managed resources we can generate. We would like to be able to ship all supported resources in a provider package.

[Terrajet-based]: https://github.com/crossplane-contrib/terrajet

As part of the `v1.Provider`, `v1.ProviderRevision`, `v1.Configuration` and `v1.ConfigurationRevision` specs, we could have a new configuration field that tells the GVKs of package objects to be installed onto the cluster. To behave backwards-compatible, if this new API is not used, all objects defined in the package manifest get installed. If the new field has a non-zero value, then it's enabled and only selected objects get installed. In order to make the UX around installing packages using the new configuration field easier, we can also add new options for the `install provider` and `install configuration` commands of the Crossplane `kubectl` plugin.

As described in the experiment results, when a large #-of-CRDs are registered in a relatively short period of time, the background task that prepares the OpenAPI spec to be published from the `/openapi/v2` endpoint may cause a spike in the API server's CPU utilization, and this may saturate the CPU resources allocated to the API server. Similar to what controller-runtime client does, we may also consider implementing a throttling mechanism in the package revision controller to prevent this. However, because of the reasons discussed above and in the experiment results (especially the latency introduced in OpenAPI spec publishing), it looks like we will need a complementary mechanism like the one suggested above in addition to a throttling implementation in the package revision reconciler.

ulucinar:
<!--
Thank you for helping to improve Crossplane!

Please read through https://git.io/fj2m9 if this is your first time opening a
Crossplane pull request. Find us in https://slack.crossplane.io/messages/dev if
you need any help contributing.
-->

<!--
Briefly describe what this pull request does. Be sure to direct your reviewers'
attention to anything that needs special consideration.

We love pull requests that resolve an open Crossplane issue. If yours does, you
can uncomment the below line to indicate which issue your PR fixes, for example
"Fixes #500":

-->
Fixes #2649

This PR proposes a change that introduces the field `enabledAPIs` for both package and package revisions to support deploying package objects selectively. Each object contained in the package manifest is checked against the regular expressions provided in the `enabledAPIs` list. If no match is found, package object is not established and skipped. If `enabledAPIs` is empty, then all objects are established. 

Changing the `enabledAPIs` list results in a new package revision.  However, duplicates are eliminated and the order of the regular expressions in the `enabledAPIs` do not matter. Thus the following specifications are all the same and will not result in a package revision (and corresponds to the same package revision):
```yaml
  enabledAPIs:
  - virtual.*
  - lb
```

```yaml
  enabledAPIs:
  - lb
  - virtual.*
```

```yaml
  enabledAPIs:
  - lb
  - lb
  - virtual.*
  - virtual.*
```
. However, the following enabled APIs declaration belongs to a different package revision although with the current `provider-tf-azure` it selects the same set of package objects:
```yaml
  enabledAPIs:
  - lb.*
  - virtual.*
```

Each item of `enabledAPIs` is a regular expression that's matched with a string representation of the `schema.GVK` of each package object. For this purpose, the package objects GVK is formatted as:
```
<API Group>/<API Version>.<Kind>
```
For instance, `provider-tf-azure`s `v1alpha1.ResourceGroup` kind's GVK is formatted as the following and then is matched against every element of `enabledAPIs` if a non-zero valued `enabledAPIs` is supplied:
```
resource.azure.tf.crossplane.io/v1alpha1.ResourceGroup
```
This allows us to define fine-grained selectors to include/exclude APIs. 

Also in this scheme, `apiextensions.k8s.io/v1.CustomResourceDefinition`s are treated in a special way. If the package object is a CRD, then it's not only the CRD that's matched against items in `enabledAPIs` but also all versions of the defined kind by that CRD. This allows us, for example, to enable the whole `virtual` API group in `provider-tf-azure` with a compact specification such as `virtual` or `virtual.*`. 

Active package revision is determined by the package image's digest and a stable digest computed from the value of the `enabledAPIs` field and not from the actual set of package objects selected. I think this is a good enough approximation for practical purposes and also reflects the usual behavior of similar APIs.

This PR also changes the package revision name format if `enabledAPIs` is non-empty. A hash computed from the set of enabled APIs is appended to the package revision name so that the new format is `<package name>-<image hash>-<enabled APIs hash>`. If `enabledAPIs` has zero-length, then package revision format is still `<package name>-<image hash>`. Thus, Crossplane upgrade is not expected to affect existing provider installations unless the new API is explicitly used in the `v1.Provider` or `v1.Configuration` object defining the package.

Another important aspect of the changes proposed here is that the provider revision reconciler passes a new command-line option, `--enabled-apis`, to the provider container (defined in the provider deployment corresponding to the revision) to communicate the enabled APIs, the provider is expected to honor this command-line option and add controllers only for the enabled APIs. It may fail if it does not properly handle the new command-line option, as any CRDs not selected by a non-zero valued `enabledAPIs` field will not have been registered via the package revision controller. Or it may fail because of being passed an unknown command-line option. Example manifest for a provider deployment when `enabledAPIs` has a non-zero value is given in the "How has this code been tested" section. If `enabledAPIs` has a zero-length, then provider revision controller does not produce the `--enabled-apis` command-line option thus the behavior is backwards-compatible if the new API is not employed.

I have:

- [ ] Add unit tests

<!--
Before reviewers can be confident in the correctness of this pull request, it
needs to tested and shown to be correct. Briefly describe the testing that has
already been done or which is planned for this change.
-->
There are two additional PRs accompanying this one, in terrajet repo for generating the required code for properly handling the communicated set of enabled APIs and another one in `provider-tf-azure` that has been generated using the new terrajet version and that introduces the `--enabled-apis` command-line option to the provider. This PR has been manually tested with these three PRs using the following provider manifest. The package referred in this manifest has been built and pushed to the registry. I can also produce an amd64 version (or maybe one will have already been produced in the CI pipeline of the corresponding change) if requested:
```yaml
kind: Provider
metadata:
  name: crossplane-provider-tf-azure
spec:
  enabledAPIs:
  - Provider.*
  - virtual.*
  - lb
  - resource.azure.tf.crossplane.io/v1alpha1.ResourceGroup
  package: ulucinar/provider-tf-azure-arm64:build-83b9bfc0
```

Here is the partial `status` of the corresponding `v1.Provider` object when the above manifest is applied:
```yaml
kind: Provider
metadata:
  name: crossplane-provider-tf-azure
...
status:
  conditions:
  - reason: HealthyPackageRevision
    status: "True"
    type: Healthy
    ...
  - reason: ActivePackageRevision
    status: "True"
    type: Installed
    ...
  currentAPIs:
  - Provider.*
  - virtual.*
  - lb
  - resource.azure.tf.crossplane.io/v1alpha1.ResourceGroup
  currentIdentifier: ulucinar/provider-tf-azure-arm64:build-83b9bfc0
  currentRevision: crossplane-provider-tf-azure-75613241815f-bc37ddfabb7c
```

Here is the corresponding (partial) provider revision manifest generated by the package revision controller:
```yaml
kind: ProviderRevision
metadata:
  ...
  labels:
    pkg.crossplane.io/package: crossplane-provider-tf-azure
  name: crossplane-provider-tf-azure-75613241815f-bc37ddfabb7c
  ownerReferences:
  ...
spec:
  desiredState: Active
  enabledAPIs:
  - Provider.*
  - virtual.*
  - lb
  - resource.azure.tf.crossplane.io/v1alpha1.ResourceGroup
  image: ulucinar/provider-tf-azure-arm64:build-83b9bfc0
  ...
status:
  objectRefs:                                                                                                                                                                                                                                   
  - apiVersion: apiextensions.k8s.io/v1
    kind: CustomResourceDefinition
    name: virtualhubconnections.virtual.azure.tf.crossplane.io
    uid: b16c4a9b-f71c-421a-9b7a-5e143fa701bf
  - apiVersion: apiextensions.k8s.io/v1
    kind: CustomResourceDefinition
    name: virtualdesktopapplications.virtual.azure.tf.crossplane.io
    uid: 7ef33b35-9139-4e76-bc03-88ff9f22af6b
  - apiVersion: apiextensions.k8s.io/v1
    kind: CustomResourceDefinition
    name: resourcegrouppolicyassignments.resource.azure.tf.crossplane.io
    uid: 08bf2ffc-1e6b-4324-a571-bcc7d0674113
  - apiVersion: apiextensions.k8s.io/v1
    kind: CustomResourceDefinition
    name: virtualhubs.virtual.azure.tf.crossplane.io
    uid: a0cef493-a876-4ef3-9372-6da631ee17cf
  - apiVersion: apiextensions.k8s.io/v1
    kind: CustomResourceDefinition
    name: virtualdesktopworkspaces.virtual.azure.tf.crossplane.io
    uid: 90f5d120-2926-4ad3-b14e-ad785660b296
  - apiVersion: apiextensions.k8s.io/v1
    kind: CustomResourceDefinition
    name: providerconfigs.azure.tf.crossplane.io
    uid: c883fc76-38d2-4244-86d2-ec57849f1abc
  - apiVersion: apiextensions.k8s.io/v1
    kind: CustomResourceDefinition
    name: providerconfigusages.azure.tf.crossplane.io
    uid: 792dc394-7e1e-4479-bd9c-12448f2592d5
  - apiVersion: apiextensions.k8s.io/v1
    kind: CustomResourceDefinition
    name: virtualnetworkpeerings.virtual.azure.tf.crossplane.io
    uid: dc88f728-fae0-46dd-af1a-a9996f06015a
  - apiVersion: apiextensions.k8s.io/v1
    kind: CustomResourceDefinition
    name: loadbalancers.lb.azure.tf.crossplane.io
    uid: d7b70a98-23aa-4c3e-ac09-71fb2edaf359
  - apiVersion: apiextensions.k8s.io/v1
    kind: CustomResourceDefinition
    name: resourcegroups.resource.azure.tf.crossplane.io
    uid: bd63b829-58e9-4a43-9f34-530ec67accad
  - apiVersion: apiextensions.k8s.io/v1
    kind: CustomResourceDefinition
    name: virtualdesktophostpools.virtual.azure.tf.crossplane.io
    uid: 6f389165-5513-440b-b7b0-4c9b667e8fc6
  - apiVersion: apiextensions.k8s.io/v1
    kind: CustomResourceDefinition
    name: virtualwans.virtual.azure.tf.crossplane.io
    uid: 4cef9530-150d-433a-b628-383d81182a5b
  - apiVersion: apiextensions.k8s.io/v1
    kind: CustomResourceDefinition
    name: virtualnetworkgateways.virtual.azure.tf.crossplane.io
    uid: 070f34a1-a477-4b35-a6f1-530f3b28ada3
  - apiVersion: apiextensions.k8s.io/v1
    kind: CustomResourceDefinition
    name: virtualhubroutetables.virtual.azure.tf.crossplane.io
    uid: 154ccdcd-445a-416b-b8f5-493881cae7cb
  - apiVersion: apiextensions.k8s.io/v1
    kind: CustomResourceDefinition
    name: resourcegrouptemplatedeployments.resource.azure.tf.crossplane.io
    uid: dc616558-7b8d-416a-8c4c-d2bf89cb503f
  - apiVersion: apiextensions.k8s.io/v1
    kind: CustomResourceDefinition
    name: virtualhubsecuritypartnerproviders.virtual.azure.tf.crossplane.io
    uid: 38942ac7-5efd-4f78-bbbf-5db9f523bf4c
  - apiVersion: apiextensions.k8s.io/v1
    kind: CustomResourceDefinition
    name: virtualhubips.virtual.azure.tf.crossplane.io
    uid: eabfabe9-bc24-46dc-a236-f6d87582a212
  - apiVersion: apiextensions.k8s.io/v1
    kind: CustomResourceDefinition
    name: virtualhubbgpconnections.virtual.azure.tf.crossplane.io
    uid: 8404b3b4-4790-44c2-a51b-ef604e1165fd
  - apiVersion: apiextensions.k8s.io/v1
    kind: CustomResourceDefinition
    name: virtualnetworks.virtual.azure.tf.crossplane.io
    uid: 886fca2c-0565-4169-bfa1-ba3407793e36
  - apiVersion: apiextensions.k8s.io/v1
    kind: CustomResourceDefinition
    name: virtualnetworkgatewayconnections.virtual.azure.tf.crossplane.io
    uid: d49d9327-4809-45b9-a6df-84c5419548cf
```
Please note that CRDs not selected by `enabledAPIs` are not available in `status.objectRefs`. 

Here is the (partial) provider deployment corresponding to that revision:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: crossplane-provider-tf-azure-75613241815f-bc37ddfabb7c
  ...
spec:
  ...
  template:
    ...
    spec:
      containers:
      - args:
        - --enabled-apis
        - Provider.*,virtual.*,lb,resource.azure.tf.crossplane.io/v1alpha1.ResourceGroup
      ...
```


negz:
I'd like to understand more about the use cases behind this change before we proceed with it.

I see in https://github.com/crossplane/crossplane/issues/2122 we discuss that it may help scope authz permissions for strategies like IRSA. I'm particularly curious how that would interplay with the ideas put forward in https://github.com/crossplane/crossplane/issues/2411. Would a user create multiple `Provider.pkg` each running (for example) `provider-tf-azure` with a different set of enabled API groups?

I also suspect that despite the above use case being what we have written down the more immediate motivation is that we're seeing issues scaling on number of CRDs. If that's really what this is solving for us today I'd like us to have an issue opened against crossplane/crossplane detailing it. I've heard us discuss potentially rate limiting CRD installs (to smooth out OpenAPI processing) and bumping REST mapper client rate limits as other options to alleviate this undocumented problem. Do we think this option is the better solution or is it a temporary workaround?

negz:
We decided to hold this functionality for now; per https://github.com/crossplane/crossplane/issues/2649 we're addressing the performance issues we've seen by working with the Kubernetes folks to improve performance there. That said, we may revisit something like this functionality in future.

PR #2692, opened 2021-11-05
[helm] Add podSecurityContext to values to allow setting securityContext on Pod level

### LINKED ISSUE #2694: Denied permission when mounting dynamic volume to the cache on GKE
### What happened?

I would like to use crossplane with multiple providers, which means that cache size will be increased.
I used a PVC on GKE to make crossplane save caches in a volume instead of in memory.
 
Using crossplane values to create a pvc

```yaml
packageCache:
  medium: "crossplane"
  sizeLimit: 1Gi
  pvc: "crossplane"
```

this will cause an issue, if you check the error in the pkgrev

```
cannot initialize parser backend: failed to store package in cache: open /cache/crossplane-provider-helm-d29c7a1e7fbb.xpkg: permission denied
```

### How can we reproduce it?

Deploy crossplane on GKE and use pvc that provision dynamic pv, then deploy any provider.

Crossplane version: 

* Cloud provider or hardware configuration: GCP
* Kubernetes version (use `kubectl version`): "v1.21.2"
* Kubernetes distribution (e.g. Tectonic, GKE, OpenShift): GKE

mcbenjemaa:
Signed-off-by: Mohamed Chiheb <mc.benjemaa@gmail.com>

Adding PVC (dynamic pv) for the cache on GKE  cluster, causes this issue:

```
cannot initialize parser backend: failed to store package in cache: open /cache/crossplane-provider-helm-d29c7a1e7fbb.xpkg: permission denied
```
That's because the volume is mounted without proper permissions.

To make the pod able to set ownership for the mounted volume! we need to set, `fsGroup` on the [PodSecurityContext ](https://pkg.go.dev/k8s.io/api@v0.22.3/core/v1#PodSecurityContext)

e.g.
```
spec:
  securityContext:
     fsGroup: 2000
```

Current crossplane values doesn't have option to set securityContext on the pod level.

In this PR I added an option to allow setting POD level `securityContext`.
While I believe that by design this is fine,  as long as container level `securityContext` will override pod level ones.

Fixes #2694

I have:

/cc @negz @hasheddan 

I think this is worth to have a patch release.
Can you create a patch release to adapt this? 

### DIFF
--- a/cluster/charts/crossplane/templates/deployment.yaml
+++ b/cluster/charts/crossplane/templates/deployment.yaml
@@ -27,6 +27,8 @@ spec:
         app: {{ template "name" . }}
         release: {{ .Release.Name }}
     spec:
+      securityContext:
+        {{- toYaml .Values.podSecurityContextCrossplane | nindent 8 }}
       {{- if .Values.priorityClassName }}
       priorityClassName: {{ .Values.priorityClassName  | quote }}
       {{- end }}
--- a/cluster/charts/crossplane/templates/rbac-manager-deployment.yaml
+++ b/cluster/charts/crossplane/templates/rbac-manager-deployment.yaml
@@ -28,6 +28,8 @@ spec:
         app: {{ template "name" . }}-rbac-manager
         release: {{ .Release.Name }}
     spec:
+      securityContext:
+        {{- toYaml .Values.podSecurityContextRBACManager | nindent 8 }}
       {{- if .Values.priorityClassName }}
       priorityClassName: {{ .Values.priorityClassName | quote }}
       {{- end }}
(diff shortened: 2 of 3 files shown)


hasheddan (approved):
Thanks for opening this @mcbenjemaa! I do think this is worthy of path release (feature not functioning properly should be considered a bug), and is probably just a good thing to make configurable in general. That being said, I believe this _can_ be worked around today by specifying the GID on the PV using the beta annotation (https://kubernetes.io/docs/tasks/configure-pod-container/configure-persistent-volume-storage/#access-control).

mcbenjemaa:
Yes! In my case I would rather prefer the gid annotation in the PVC(But that's not supported.), as it's a dynamic pv provisioned.
I think we gotta make crossplane handle this behaviour!

Issue #1152, opened 2020-01-07
One-pager for pluggable webhooks


muvaf:
### What problem are you facing?

In Crossplane, we need validating and mutating webhooks in several places. But the certificate handling and the structure of webhooks have made it hard to get something working and extensible quickly. A one-pager that introduces a mechanism (possibly in crossplane-runtime) for developers to add their webhooks just like adding a new controller would help a lot.

The key point is the stack developers should be able to register their webhooks as easy as registration of a new controller. 

### How could Crossplane help solve your problem?

The first candidate use-cases that come to my mind:

Validating webhook:
* Reject modifications to fields marked `// +immutable`
* Validate objects in `spec.templates` array against their `kind`

Mutating webhook:
* All scheduling controllers could be mutating webhooks
  * defaulting resource class controller
  * generic resource class selector
  * KubernetesApplication scheduler

Related issues:
* https://github.com/crossplaneio/crossplane/issues/169 
* https://github.com/crossplaneio/crossplane/issues/950
* https://github.com/crossplaneio/crossplane/issues/967
* https://github.com/crossplaneio/crossplane/issues/727
* https://github.com/crossplaneio/crossplane/issues/427
* https://github.com/crossplane/crossplane/issues/1476
* https://github.com/crossplane/crossplane/issues/1497
* https://github.com/crossplane/crossplane/issues/1584

muvaf:
Assigning to myself just to not lose track of the issue. There isn't a plan to implement this yet.

displague:
I am also finding that this will be necessary to offer conversion webhooks to support Stack versioning. 

https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning/#webhook-conversion

The stack-manager will need to manage the deployment of the webhook service and app.yaml will need ways to express this.

displague:
A conversion webhook would resolve https://github.com/crossplaneio/crossplane/issues/533

negz:
It seems like it's increasingly common for controller managers to include built in support for managing their own webhook certs - e.g. https://github.com/open-policy-agent/cert-controller.

Issue #2671, opened 2021-10-26
[RFC] Provider Runtime Interface (PRI)


hasheddan:
### What problem are you facing?

Crossplane supports two types of packages: `Configurations` and `Providers`. In practice, both of these packages need only be a single `package.yaml` file in a tarball in the first layer of an OCI image that contains a [stream of YAML documents](https://yaml.org/spec/1.2.2/#92-streams). However, `Providers` are unique in that they also [point to another image](https://doc.crds.dev/github.com/crossplane/crossplane/meta.pkg.crossplane.io/Provider/v1@v1.4.2#spec-controller-image) that is contains the controller for the CRDs in the YAML stream. When a `Provider` is installed in a Crossplane cluster, Crossplane unpacks the package image, installs all of the CRDs in the YAML stream, then creates a `Deployment` using the the referenced controller image (which can be further [configured by referencing a `ControllerConfig`](https://doc.crds.dev/github.com/crossplane/crossplane/pkg.crossplane.io/ControllerConfig/v1alpha1@v1.4.2)).

However, the fact the the controllers are run as a `Deployment` is merely an implementation detail. They could be provisioned and run as another Kubernetes primitive, a higher-level construct (e.g. a function), or a runtime completely external to the cluster (e.g. a server, a managed service, another Kubernetes cluster etc.). Enabling each of these use cases opens up a set of unique deployment models for Crossplane itself:

- Utilizing external runtimes means that Crossplane could run with a minimized Kubernetes API server, perhaps even one that is bundled into Crossplane itself. For example, running a Crossplane instance could be as simple as `docker run crossplane/crossplane:latest`, and providers could be run by using a Docker implementation of the "Provider Runtime Interface (PRI)" that deployed the provider controllers as containers that shared a network namespace with Crossplane.
- Decoupling the control plane from the provider runtime means that the two components can be operated independently, potentially by different parties. This enables scenarios where credentials may need to be managed externally (at the provider runtime), but package management and composition can be managed by a third party.
- A PRI implementation could opt to handle some aspects of the data plane. For example, something like [crisscross](https://github.com/crossplane-contrib/crisscross) could be implemented by the PRI implementation provisioning an intermediate controller that brokers communication to smaller functions that can be written an any language (essentially simplifying the interface to be notified of Kubernetes events). (Note: I am not entirely sold on this being the way to do `crisscross`, but seems a least related to this effort).

### How could Crossplane help solve your problem?

[...]

PR #2719, opened 2021-11-26
design: webhooks in crossplane

muvaf:

This design aims to cover high level design of webhooks in Crossplane ecosystem. While it wasn't as urgent in the past because the improvements were mostly about UX, now that we have many `v1beta1` APIs and candidates for `v1`, it becomes necessity to tackle this problem.

Fixes https://github.com/crossplane/crossplane/issues/1152

I have:

I walked it through the corridors of my brain.


design/design-doc-webhooks.md
```diff
@@ -0,0 +1,383 @@
...
+  packages as lean as possible so that external users of those Go packages do
+  not have to depend on things that are irrelevant to them. However, there are
+  some cases that might require including cloud provider SDK and/or `client`
+  package from controller-runtime:
+  * Let's say GCP decides to add a new field in the new version of SDK and we
+    want to include it in the new `apiVersion`. However, in order to make
+    queries the value of that field needs to be known, hence we need to make a
+    `GET` call to the provider API, which means we'd have to import GCP Go SDK.
+    We update SDKs very infrequently so we have not encountered this but it's
+    possible.
```
negz, line 204: I'm having a tough time wrapping my head around this case. Could you elaborate on it a little?

design/design-doc-webhooks.md
```diff
@@ -0,0 +1,383 @@
...
+  return PatchResponseFromRaw(req.Object.Raw, marshalled)
+}
+```
+
+The conversion webhook implementation is a bit [more
+complex](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/webhook/conversion/conversion.go#L22)
+due to the marshalling and version checks that need to happen. However, it seems
+that some of the cases it needs to handle [are not
+necessary](https://github.com/kubernetes-sigs/controller-runtime/issues/1388) in
+v1 CRD mechanics.
```
negz, line 191: Is this really true? It would be surprising to me if controller-runtime had built superfluous functionality and no one else had challenged that until now. It doesn't look like the linked issue really got to any resolution.

design/design-doc-webhooks.md
```diff
@@ -0,0 +1,383 @@
...
+While the provider installations will be handled this way, Crossplane doesn't
+orchestrate the installation of itself. But we will again utilize the fact that
+we have package manager, hence necessary RBAC to create the registration YAMLs.
+So, what we will do is that Crossplane will register its own webhook
+configurations right after it makes sure the TLS secret is there during the
+initialization phase of the process, similar to how we deploy
+`CustomResourceDefinition`s of Crossplane in its init container.
+
+Since TLS `Secret` is required, no webhook functionality will be enabled if it's
+not provided.
```
negz, line 285: https://github.com/knative/pkg/tree/main/webhook/certificates

I vaguely recall that the Knative folks addressed this by having Knative automatically provision its own webhook certificates (presumably with some facility for a user to override this and supply one explicitly). Could we do something similar?

design/design-doc-webhooks.md
```diff
@@ -0,0 +1,383 @@
...
+When a controller-runtime manager starts, the webhook server is also started
+automatically. So, as long as necessary implementations are there and hooked up
+with the main manager object, we have the server up and running. But in order to
+expose it to cluster, we need to create a `Service` resource similar to
+`Deployment` of the controller.
+
+The package manager will create an opinionated `Service` resource for the
+providers it installs and this will require Crossplane to have necessary RBAC
+for managing `Service` objects, too. Though it could be limited to which
+namespace it's installed, again similar to `Deployment` RBAC.
```
hasheddan, line 255: This seems like it could me a prime candidate for sitting behind the [Provider Runtime Interface](https://github.com/crossplane/crossplane/issues/2671) so that we are not explicitly tied to using a `Service` since [deploying outside a cluster](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#deploy-the-admission-webhook-service) is an option.

design/design-doc-webhooks.md
```diff
@@ -0,0 +1,383 @@
...
+certificates or having `cert-manager` installed.
+
+We will utilize the fact that we are orchestrating the installation of provider
+controllers by our package manager. Crossplane will accept the TLS `Secret` as
+input to its installation. Then in every provider installation, it will mount
+this `Secret` to the provider container to use as the certificate & key. Since
+it also installs the necessary YAMLs that register the webhooks to API Server,
+it will inject the CA Bundle before the creation of those resources, which are
+`MutatingWebhookConfiguration`, `ValidatingWebhookConfiguration` and the
+`spec.webhook` part of `CustomResourceDefinition`s.
```
hasheddan, line 274: This also feels like it could be part of PRI -- I do acknowledge that it does not exist yet though, so this feedback isn't super helpful. If we think that is the direction we want to go though, this proposal could be a motivating use case.

design/design-doc-webhooks.md
muvaf: Yes, for example, let's say we're migrating from a Jet CRD to native or vice versa, and the resource definitions do not match, i.e. `RouteTable` has `ingress` array field in one but not in the other because it has `RouteTableIngress`. In such cases, we'd need to make calls to AWS to get the information about the ingresses so that the conversion is loss-less.

design/design-doc-webhooks.md
```diff
@@ -0,0 +1,383 @@
...
+
+The conversion webhook implementation is a bit [more
+complex](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/webhook/conversion/conversion.go#L22)
+due to the marshalling and version checks that need to happen. However, it seems
+that some of the cases it needs to handle [are not
+necessary](https://github.com/kubernetes-sigs/controller-runtime/issues/1388) in
+v1 CRD mechanics.
+
+These high level abstractions are fairly simple but they have a few drawbacks:
+* They require implementation in API type packages and we want to keep API type
```
muvaf, line 194: I discovered that with https://github.com/kubernetes-sigs/controller-runtime/pull/1676 now it's possible to register custom handlers that are not part of the API definition, so this concern is mostly gone. I'll update this PR once I'm done with prototyping to see how it all looks like.

design/design-doc-webhooks.md
muvaf: Though that's adressed with custom handlers now https://github.com/crossplane/crossplane/pull/2719/files#r817234614

muvaf: I've removed this section completely, proposing that we build layer only to have an ordered execution of multiple functions in a given webhook call; no re-implementation since untyped webhook registration is now possible.

design/design-doc-webhooks.md
muvaf: I removed that section because I changed the proposal so that the layer is built on top of the existing high level abstraction, so no need to discuss this in the doc anymore.

design/design-doc-webhooks.md
muvaf: I have decided to change the proposal from `controller-runtime high level abstraction does not work for us so we will build one` to `we will build a lean layer on top of what is in controller-runtime so that we can append multiple function that will be called in a single webhook call` given that the biggest concern about their abstraction layer is eliminated with custom handlers.

design/design-doc-webhooks.md
muvaf: If we do end up with that runtime interface, I agree that it could be a good candidate but right now, I'm planning to integrate to the existing mechanisms with as little change as possible.

design/design-doc-webhooks.md
negz: @muvaf and I chatted about this out of band, and I believe we're thinking we could handle this.

negz (approved):
Apologies for the last minute approval, given this is pretty much implemented at this point. 😬 It might be worth capturing the fact that we did decide auto-generate the secret required to secure webhook communication after all.

elohmrow:
> I walked it through the corridors of my brain.
🤣

PR #2727, opened 2021-11-29
add note to package doc

fahedouch:
Signed-off-by: fahed dorgaa <fahed.dorgaa@gmail.com>

Why adding this comment? 
I spent a lot of time to understand that the reason for rejecting permissions is that they are not declared in the provider-clusterrole
Fixes #Documentation

I have:

no need

### DIFF
--- a/docs/concepts/packages.md
+++ b/docs/concepts/packages.md
@@ -108,7 +108,10 @@ all types that its package installs, as well as `Secrets`, `ConfigMaps`, and
 
 > Note that the Crossplane RBAC manager can be configured to reject permissions
 > for certain API groups. If a package requests permissions that Crossplane is
-> configured to reject, the package will fail to be installed. 
+> configured to reject, the package will fail to be installed.
+> Authorized permissions should be aggregated to the rbac manager clusterrole 
+> (the cluster role defined by the provider-clusterrole flag in the rbac manager) 
+> by using the label `rbac.crossplane.io/aggregate-to-allowed-provider-permissions: "true"`
 
 The `spec.crossplane.version` field specifies the version constraints for core
 Crossplane that the `Provider` is compatible with. It is advisable to use this


docs/concepts/packages.md
```diff
@@ -108,7 +108,9 @@ all types that its package installs, as well as `Secrets`, `ConfigMaps`, and
 
 > Note that the Crossplane RBAC manager can be configured to reject permissions
 > for certain API groups. If a package requests permissions that Crossplane is
-> configured to reject, the package will fail to be installed. 
+> configured to reject, the package will fail to be installed.
+> Authorized permissions should be added to the rbac manager provider-clusterrole
+> (the cluster role defined by the provider-clusterrole flag in the rbac manager)
```
hasheddan, line 113: Thanks so much for adding the note of clarity here @fahedouch! One of the nice features of Kubernetes RBAC is [`ClusterRole` aggregation](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#aggregated-clusterroles), which we employ with the allowed permissions for providers, by using a label:
https://github.com/crossplane/crossplane/blob/2cb935ac62650d2932e73c4e2957ac55f02bd3af/cluster/charts/crossplane/templates/rbac-manager-allowed-provider-permissions.yaml#L13

I believe this is our recommended strategy for adding allowed permissions for providers, rather than mutating the `ClusterRole` we create directly. Would you mind updating the note to reflect that?

hasheddan (changes requested):

docs/concepts/packages.md
fahedouch: @hasheddan WDYT ?

hasheddan: @fahedouch looks good! Although it looks like there may be an unintended build submodule update included now?

fahedouch: Sorry @hasheddan . fixed

hasheddan (approved):

PR #2740, opened 2021-12-02
Use dedicated Upbound Registry robots for Configurations and Crossplane

hasheddan:

Updates the configurations workflow to use UPBOUND_XP_ROBOT, which has
credentials to push repositories in the XP org on Upbound Registry.

Moves to using UPBOUND_CROSSPLANE_ROBOT for pushing the crossplane image
to Upbound Registry. This robot account has permissions to push to
repositories in the Crossplane org.

I have:

I have tested registry operations with both of these sets of credentials, but we won't know for sure until CI runs.

### DIFF
--- a/.github/workflows/configurations.yml
+++ b/.github/workflows/configurations.yml
@@ -8,7 +8,7 @@ on:
   workflow_dispatch: {}
 
 env:
-  DOCKER_USR: ${{ secrets.UPBOUND_ROBOT_USR }}
+  DOCKER_USR: ${{ secrets.UPBOUND_XP_ROBOT_USR }}
 
 jobs:
   getting-started-with-aws:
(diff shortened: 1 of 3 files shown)


.github/workflows/configurations.yml
```diff
@@ -29,8 +29,8 @@ jobs:
         if: env.DOCKER_USR != ''
         with:
           registry: registry.upbound.io
-          username: ${{ secrets.UPBOUND_ROBOT_USR }}
-          password: ${{ secrets.UPBOUND_ROBOT_PSW }}
+          username: ${{ secrets.UPBOUND_XP_ROBOT_USR }}
+          password: ${{ secrets.UPBOUND_XP_ROBOT_PSW }}
```
negz, line 33: Would it make sense to just use the existing xp org for everything? Both Crossplane itself, as well as configurations (and in future providers)? Having two orgs that both mean kind of the same thing ("crossplane" and abbreviation of "crossplane") seems a little confusing.

negz (approved):
Looks good, though I question whether we should have two orgs.

.github/workflows/configurations.yml
hasheddan: Just noting here as we discussed offline: `crossplane` org already exists as well and that is where providers are manually synced today, so we want to keep using it. We may opt to consolidate in the future, but for now `crossplane` org = offical and `xp` org = supplementary artifacts (experimental / docs / etc.)

PR #2742, opened 2021-12-02
Actually cache the go build cache between CI runs

hasheddan:

The build submodule currently overrides XDG_CACHE_HOME in
order to force the Helm 3 to use the .work/helm directory. This causes Go on
Linux machines to use that directory as the build cache as well. We should
adjust this behavior in the build submodule because it is also causing Linux
users to duplicate their build cache, but for now we just make it easier to
identify its location in CI so that we cache between builds.

Updates CI workflow to cache based on go.cachedir to work around the
fact that we override the GOCACHE in our make context due to the
inclusion of helm.mk.

I have:

I have tested this out locally, and anecdotally, I am a Linux user and discovered that all of my projects that consume the `build` submodule are maintaining their individual caches in `.work/helm`. We should fix this in the `build` submodule, but it will likely have farther reaching effects, and this solution gets us caching in CI in the short term, and is resilient to any changes made in the `build` submodule as we are entering the `make` context and then invoking `go env GOCACHE`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -161,6 +161,15 @@ e2e-tests-compile:
 # Compile e2e tests for each platform.
 build.code.platform: e2e-tests-compile
 
+# NOTE(hasheddan): the build submodule currently overrides XDG_CACHE_HOME in
+# order to force the Helm 3 to use the .work/helm directory. This causes Go on
+# Linux machines to use that directory as the build cache as well. We should
+# adjust this behavior in the build submodule because it is also causing Linux
+# users to duplicate their build cache, but for now we just make it easier to
+# identify its location in CI so that we cache between builds.
+go.cachedir:
+	@go env GOCACHE
+
 # This is for running out-of-cluster locally, and is for convenience. Running
 # this make target will print out the command which was used. For more control,
 # try running the binary directly with different arguments.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -101,7 +101,7 @@ jobs:
 
       - name: Find the Go Build Cache
         id: go
-        run: echo "::set-output name=cache::$(go env GOCACHE)"
+        run: echo "::set-output name=cache::$(make go.cachedir)"
 
       - name: Cache the Go Build Cache
         uses: actions/cache@v2


hasheddan (commented):
A few relevant links:
- Where `XDG_CACHE_HOME` is being overwritten in `build`: https://github.com/upbound/build/blob/70b7766c9cc2e1f4fc36d5b8a2b3ffd4661a6df5/makelib/helm.mk#L48
- Where Go sources cache dir: https://cs.opensource.google/go/go/+/master:src/cmd/go/internal/cache/default.go;drc=master;l=70
- Where Go use `XDG_CACHE_HOME` if set: https://cs.opensource.google/go/go/+/refs/tags/go1.17.3:src/os/file.go;l=431

Makefile
```diff
@@ -161,6 +161,15 @@ e2e-tests-compile:
 # Compile e2e tests for each platform.
 build.code.platform: e2e-tests-compile
 
+# NOTE(hasheddan): the build submodule currently overrides XDG_CACHE_HOME in
+# order to force the Helm 3 to use the .work/helm directory. This causes Go on
+# Linux machines to use that directory as the build cache as well. We should
+# adjust this behavior in the build submodule because it is also causing Linux
+# users to duplicate their build cache, but for now we just make it easier to
+# identify its location in CI so that we cache between builds.
```
hasheddan, line 169: It is possible that we do want to replicate the build cache across projects, and this is actually working for speeding up builds locally because we are pointing at the same location every time. However, we are not caching this between builds in CI because `go env GOCACHE` outside of our `make` context is not giving us the same path as inside of our `make` context.

Makefile
```diff
@@ -161,6 +161,15 @@ e2e-tests-compile:
...
+go.cachedir:
+	@go env GOCACHE
```
negz, line 171: Dumb make question - how is this different from just calling `go env GOCACHE` directly like we did before?

negz (approved):
Was this a regression introduced when we switched to Helm 3? I vaguely recall builds getting a bunch slower recently - wonder if this was the cause?

Makefile
hasheddan: @negz this calls it in our `make` context, which means we have already evaluated the helm 3 override and export of `XDG_CACHE_HOME`. This also gives us the nice property of not breaking things when we fix this upstream in the `build` submodule (it will still just evaluate to whatever it then equals in the new `make` context). And yep it broke with helm 3 :)

PR #2748, opened 2021-12-08
Update Azure credentials step to not require AD access

hasheddan:

The permissions we grant in this step are highly privileged and are
currently only required if we are provisioning AKS clusters, which we no
longer do in the getting started guide. Many users will not be able to
set the permissions, and may be deterred from continuing if they
encounter an error. This removes them as they are not required for the
operations performed in subsequent steps.

I have:

Went through getting started guide without adding permissions that are removed in this commit.

### DIFF
--- a/docs/getting-started/install-configure.md
+++ b/docs/getting-started/install-configure.md
@@ -421,22 +421,6 @@ watch kubectl get pkg
 ```console
 # create service principal with Owner role
 az ad sp create-for-rbac --sdk-auth --role Owner > "creds.json"
-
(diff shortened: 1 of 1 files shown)


jbw976 (approved):
This makes a lot of sense to streamline and remove a common cause for errors/issues with folks in the quick start - it isn't needed for the resources provisioned in the quick start at all.

For future readers: Both provider-jet-azure and https://github.com/crossplane/provider-azure/issues/292 can further help with AKS clusters currently needing these extra permissions.

PR #2756, opened 2021-12-10
Make crossplane ServiceAccount creation optional

### LINKED ISSUE #2755: Provide a bring-your-own ServiceAccount option in the crossplane helm chart

The latest version of the helm chart (1.5.1) unconditionally creates a ServiceAccount. I would like to manage the ServiceAccount outside the helm chart, so it would be nice to be able to disable the ServiceAccount creation in the helm chart.

I intend to create a PR that does what I'm asking for, so I suppose the way Crossplane could help solve my problem is by having a look at my PR when it is ready and helping me to get it into the helm chart officially.

jeremyspykerman:
For anyone who wants to manage the crossplane ServiceAccount externally
this defines a flag to allow disabling the creation of this specific
ServiceAccount in the Helm chart.

Fixes #2755 

I have:

I tried to run `make reviewable`:
~~~
make reviewable
make: *** No rule to make target `reviewable'.  Stop.
~~~

I have used the existing values.yaml.tmpl file to create a valid values.yaml file. I then used my own override of the new setting.

Below shows the steps I used other than having to provide a valid value to replace `%%VERSION%%` from the template.

~~~
$ cp values.yaml.tmpl values.yaml
$ helm template . > /tmp/default.yaml
$ helm template . -f myvalues.yaml > /tmp/serviceaccountdisabled.yaml
$ diff /tmp/default.yaml /tmp/serviceaccountdisabled.yaml 
17,33d16
< # Source: crossplane/templates/serviceaccount.yaml
< apiVersion: v1
< kind: ServiceAccount
< metadata:
<   name: crossplane
<   labels:
<     app: crossplane    
<     helm.sh/chart: crossplane-0.0.1
<     app.kubernetes.io/managed-by: Helm
<     app.kubernetes.io/component: cloud-infrastructure-controller
<     app.kubernetes.io/part-of: crossplane
<     app.kubernetes.io/name: crossplane
<     app.kubernetes.io/instance: RELEASE-NAME
<     app.kubernetes.io/version: "0.0.1"
< imagePullSecrets:
< - name: dockerhub
< ---
~~~

This shows that the flag works as intended.

### DIFF
--- a/cluster/charts/crossplane/README.md
+++ b/cluster/charts/crossplane/README.md
@@ -87,6 +87,7 @@ and their default values.
 | `securityContextCrossplane.runAsGroup` | Run as group for Crossplane | `65532` |
 | `securityContextCrossplane.allowPrivilegeEscalation` | Allow privilege escalation for Crossplane | `false` |
 | `securityContextCrossplane.readOnlyRootFilesystem` | ReadOnly root filesystem for Crossplane | `true` |
+| `serviceAccount.crossplane.create` | Setting to `false` means the crossplane ServiceAccount is managed outside this chart | `true` |
 | `provider.packages` | The list of Provider packages to install together with Crossplane | `[]` |
 | `configuration.packages` | The list of Configuration packages to install together with Crossplane | `[]` |
 | `packageCache.medium` | Storage medium for package cache. `Memory` means volume will be backed by tmpfs, which can be useful for development. | `""` |
--- a/cluster/charts/crossplane/templates/serviceaccount.yaml
+++ b/cluster/charts/crossplane/templates/serviceaccount.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.serviceAccount.crossplane.create }}
 apiVersion: v1
 kind: ServiceAccount
 metadata:
@@ -11,3 +12,4 @@ imagePullSecrets:
 - name: {{ $secret }}
 {{- end }}
 {{ end }}
+{{- end }}
--- a/cluster/charts/crossplane/values.yaml.tmpl
+++ b/cluster/charts/crossplane/values.yaml.tmpl
@@ -55,6 +55,10 @@ securityContextCrossplane:
   allowPrivilegeEscalation: false
   readOnlyRootFilesystem: true
 
+serviceAccount:
+  crossplane:
+    create: true
+
 packageCache:
   medium: ""
   sizeLimit: 5Mi


cluster/charts/crossplane/README.md
```diff
@@ -87,6 +87,7 @@ and their default values.
 | `securityContextCrossplane.runAsGroup` | Run as group for Crossplane | `65532` |
 | `securityContextCrossplane.allowPrivilegeEscalation` | Allow privilege escalation for Crossplane | `false` |
 | `securityContextCrossplane.readOnlyRootFilesystem` | ReadOnly root filesystem for Crossplane | `true` |
+| `serviceAccount.crossplane.create` | Setting to `false` means the crossplane ServiceAccount is managed outside this chart | `true` |
```
muvaf, line 90: Instead of an on-off switch, can we accept the name of the `ServiceAccount` you'd like to use? Crossplane could create its own if no name is given.

jeremyspykerman: That will not address my specific use case. I want to create it outside of the helm chart completely because it needs to be created with AWS arn specific annotations for rbac purposes.

I suppose if there was a way to add annotations to the service account that would help, but as we already have the code for creating the service account I'd prefer to just switch it off completely in this helm chart. I accept that this is not my codebase, so others may not agree.

muvaf: What I meant was that if you provide the name of the ServiceAccount you created, the chart would make Crossplane use it - with whatever anootations/labels you deem appropriate for that ServiceAccount. If a name isn't provided, only then it'd create its own. Wouldn't that work for you?

dwerder: @jeremyspykerman, with the merge of #2788 your usecase is handled. In release 1.6.0 the helm charts allows you to set annotation for the service account, to be used for AWS IRSA stuff.

jeremyspykerman: I agree that what you've proposed is a better solution @muvaf . I just haven't had time to refactor to take this approach.

Also now that the option mentioned by @dwerder is available, I'll have to reconsider the approach we're using. It could be quite some time, but once we make a decision regarding which approach we want to take I'll either rework the PR, or adapt our code to use the annotation features from #2788

andycmaj: chiming in here RE https://crossplane.slack.com/archives/CEG3T90A1/p1643853900373079 on slack...

i would still like to see the option to pass a custom SA name. reason being i use `eksctl` to manage well-known service accounts, like `ecr-access`. so i would love to re-use the existing account/trust-relationship it manages for me.

otherwise, i need to add additional IAM automation to set up the trust relationship for the `crossplane` SA.

muvaf:
@jeremyspykerman are you still pursuing this PR?

negz:
We haven't heard back in a while so I'm going to close this.

PR #2309, opened 2021-05-16
Getting started doc: fix API version for AWS resource RouteTable.

Smana:
In the getting started documentation, when using a new AWS VPC I noticed
the following error:

```console
cannot render composed resource from resource template at index 6:
cannot use dry-run create to name composed resource: no matches for kind
"RouteTable" in version "ec2.aws.crossplane.io/v1alpha4"
```

Crossplane v1.2.1

Updating the composition yaml to make use of the API version v1beta1 for
AWS RouteTable.

Signed-off-by: Smaine Kahlouch <smainklh@gmail.com>

I have:

- [x] Read and followed Crossplane's [contribution process].
- [ ] Run `make reviewable test` to ensure this PR is ready for review.

### How has this code been tested

I was able to run through the getting started documentation

PR #2771, opened 2021-12-16
feat(bump-aws-versions): #2709 bump versions for provider-aws

### LINKED ISSUE #2709: Getting started docs depend on provider-aws:v0.18.2
### What happened?

The getting started images for AWS are still depending on an older version of provider-aws.  I suspect this was done because v0.18.2 had a necessary fix in it, but since we've moved on to newer versions, we should remove this exact version constraint.  This work was done in PR https://github.com/crossplane/crossplane/pull/2309.

I believe we could just update the version constraints in the [crossplane.yaml's](https://github.com/crossplane/crossplane/blob/master/docs/snippets/package/aws/crossplane.yaml#L14) (there are multiple of them, as per these [search results](https://github.com/crossplane/crossplane/search?q=%22v0.18.2%22)) so they pick up the newest, as opposed to this specific version.

### How can we reproduce it?

Follow the getting started guide for [AWS](https://crossplane.io/docs/v1.5/getting-started/install-configure.html), which instructs you to install `registry.upbound.io/xp/getting-started-with-aws:v1.5.0`.  This will result in `provider-aws:v0.18.2` being installed, when there are newer versions that would be preferred: https://github.com/crossplane/provider-aws/releases

Crossplane version: v1.5 docs have this issue, and previous docs will also.  We need to think about which versions to backport this too.

haarchri:
Signed-off-by: haarchri <chhaar30@googlemail.com>

set provider-aws to version `>=v0.22.0` because we stuck for getting started on `v0.18.2`
set crossplane version to `>=v1.4.0-0` as of `v1.3 EOL December 2021`

Fixes #2709

I have:

think we need backport for 1.4, 1.5 ?


docs/getting-started/create-configuration.md
```diff
@@ -571,10 +571,10 @@ metadata:
     vpc: default
 spec:
   crossplane:
-    version: ">=v1.2.0-0"
+    version: ">=v1.4.0-0"
   dependsOn:
     - provider: crossplane/provider-aws
-      version: "v0.18.2"
+      version: ">=v0.22.0"
```
hasheddan, line 577: @haarchri were you able to test out with these changes? I believe that there was an issue that required modifying the Compositions in order to upgrade to newer provider-aws versions.

haarchri: Tested clean-rollout what issue you mean ?

hasheddan: @haarchri just wanted to check that everything provisioned correctly with `v0.22.0` -- it looks like we pinned to `v0.18.2` before because we just needed higher than `v0.18.1` (ref https://github.com/crossplane/crossplane/pull/2309#discussion_r705852208). In practice, we really should only need to specify the "earliest version that works" because it allows more flexibility for folks who already have `provider-aws` installed and may have constraints imposed by other packages that prevent them from upgrading. Meanwhile, folks installing fresh will still get the latest version as long as we use a `>=` constraint. Is there anything in these packages that suggest a `>=v0.18.2` constraint would not be sufficient?

dwerder: As an newbe/enduser I would expect to get the latest supported version of the provider. Furthermore as far as my knowledge goes provider-aws just supports the latest version, without major backports. At the time being thats 0.22.0. So people should also be "enforced" to use that one, so they see the and can use all features.
And there is the question of breaking changes, between versions.

hasheddan: @dwerder thanks for the feedback! I respectfully disagree -- we should not enforce users to use `v0.22.0` if older versions are also compatible. New users that come along will still get the latest supported version of the provider because it will be the "latest image in the semantic version range" with a constraint like `>=v0.18.2`. We have to keep in mind that folks may have other packages that do not support newer versions of `provider-aws` and we should not force rebuild of those packages when this one would work with the `provider-aws` version they already have installed. More information on package installs [here](https://crossplane.io/docs/v1.5/concepts/packages.html#installing-a-package) :)

haarchri: @hasheddan  changed to` >=v0.18.2`

haarchri:
@hasheddan did we need something more here ?

haarchri:
@jbw976  anything i can do here ?

jbw976 (approved):
This looks reasonable to me @haarchri! Thank you for the reminder to get this merged in :)

jbw976:
I'm adding the backport labels since I think it woudl affect all currently supported docs (1.4, 1.5, 1.6)

jbw976:
I'm going to take for this for a quick test run before merging 🤪

jbw976:
LGTM, I built these changes locally and uploaded the package to UbC registry.  After installing the package (that has the nice new version constraint), I have the most fresh provider-aws pulled in as a dependency 💪 

```
> k get pkg
NAME                                                                INSTALLED   HEALTHY   PACKAGE                                                        AGE
configuration.pkg.crossplane.io/jaredorg-getting-started-with-aws   True        True      registry.upbound.io/jaredorg/getting-started-with-aws:v0.0.1   46s

NAME                                                 INSTALLED   HEALTHY   PACKAGE                           AGE
provider.pkg.crossplane.io/crossplane-provider-aws   True        True      crossplane/provider-aws:v0.24.1   44s
```

jbw976:
> Backport failed for release-1.4, because it was unable to cherry-pick the commit(s).

Looking at the merge conflicts here, it looks like 1.4 was using a specific older version due to compatibility issues with composition features in 1.4. - I will leave those 1.4 docs as is.

Issue #2262, opened 2021-04-13
[v2] Introduce a `ManagedResourceDefinition` type


negz:
### What problem are you facing?

An unintended (but nice) property of Crossplane's `CompositeResourceDefinition` (XRD) type is that it gives us a Crossplane specific artifact representing the existence of a Crossplane specific class of resource (i.e. Composite Resources or XRs). This means it's possible for a subject to walk the tree of Crossplane resources from a `Configuration` to a `ConfigurationRevision` to a `CompositeResourceDefinition` to the XRs defined by that XRD without ever needing to be granted RBAC access to anything but Crossplane-specific types. In practice an XRD is a proxy for a CRD, but Crossplane users can get all the information they need from the XRD without ever needing RBAC access to read CRDs.

Managed resources (and provider configs) are a different story. In order to walk the tree from a provider to a particular managed resource you need to traverse CRDs - e.g. `Provider` to `ProviderRevision` to `CustomResourceDefinition` to the managed resources defined by that CRD. This implies that browsing the Crossplane Resource Model (XRM) requires access to get and list CRDs. In any particular Kubernetes cluster running Crossplane it's likely that some subset of the CRDs don't have anything to do with Crossplane, and that the cluster operator might be reluctant to grant access to _all_ CRDs when in fact they really only want to enable a subject to map a particular `ProviderRevision` to the managed resources and provider configs it delivers.

### How could Crossplane help solve your problem?

Crossplane could support specific `ManagedResourceDefinition` and `ProviderConfigDefinition` types, which would be proxies for CRDs. This would allow cluster operators to grant access to the `ManagedResourceDefinition` type (for example) to allow a subject to discover what Crossplane managed resources exist without necessarily having access to discover _all_ custom resources that exist.

negz:
This proposal has a bunch of implications, e.g.:

* Would we have a provider package up a bunch of Managed Resource Definitions (MRDs) instead of CRDs?
* Would we need a controller that created a CRD for every MRD it saw?
* Would we need to make controller-tools generate MRDs from Crossplane types, instead of CRDs?

muvaf:
I'm trying to gauge whether it's worth taking on the added mental complexity of additional proxy types. Is the following list of benefits for MRDs complete?
* Instead of `list, get all CRDs` RBAC rules, they will be able to work with `list, get all MRDs`.
[...]

PR #2775, opened 2021-12-17
Add specification for xpkg images and extensions

hasheddan:

Adds a formal specification for xpkg images and extension mechanisms
they offer. The xpkg spec imposes a superset of the requirements of the
OCI image spec.

Fixes https://github.com/crossplane/crossplane/issues/2778

I have:

n/a docs -- though the structure described has been exercised through the usage of `xpkg`s in Crossplane for some time

### DIFF
--- a/docs/reference/xpkg.md
+++ b/docs/reference/xpkg.md
@@ -18,6 +18,16 @@ requirements detailed in [OCI image specification]. It is divided into two broad
 sections: requirements related to OCI image format and requirements related to
 Crossplane `package.yaml` contents.
 
+- [OCI Image Format](#oci-image-format)
+  - [Indexes](#indexes)
+  - [Manifests](#manifests)
+  - [Configuration](#configuration)
+  - [Layers](#layers)
+- [package.yaml Contents](#packageyaml-contents)
+  - [Configuration Package Requirements](#configuration-package-requirements)
+  - [Provider Package Requirements](#provider-package-requirements)
+  - [Object Annotations](#object-annotations)
+
 ## OCI Image Format
 
 OCI images are comprised of [manifests], [configuration], and [layers].
@@ -80,7 +90,12 @@ If no layer descriptors have an annotation in the form `io.crossplane.xpkg:
 base`, the resultant filesystem from [applying changesets] from all layers will
 be used.
 
-> Crossplane prefers the usage of layers to define additive package content as it provides a clean mechanism to build an `xpkg` through a series of stages. A valid `xpkg` can be produced and later modified while verifying that the integrity of the existing content is not violated, which ensures that Crossplane's package manager will process the resulting `xpkg` in the same manner as the it would prior to modification.
+> Crossplane prefers the usage of layers to define additive package content as
+> it provides a clean mechanism to build an `xpkg` through a series of stages. A
+> valid `xpkg` can be produced and later modified while verifying that the
+> integrity of the existing content is not violated, which ensures that
+> Crossplane's package manager will process the resulting `xpkg` in the same
+> manner as the it would prior to modification.
 
 ### Configuration
 
@@ -126,7 +141,7 @@ requirements:
   objects MAY be defined in the YAML stream.
 - Zero (0) or more `Composition.apiextensions.crossplane.io` objects MAY be
   defined in the YAML stream.
-- No other object types may be defined in the YAML stream.
+- Zero (0) other object types may be defined in the YAML stream.
 
 ### Provider Package Requirements
 
@@ -172,6 +187,7 @@ other objects in a package are propagated to their in-cluster representation
 unmodified.
 
 <!-- Named Links -->
+
 [packages]: ../concepts/packages.md
 [OCI images]: https://github.com/opencontainers/image-spec
 [OCI image specification]: https://github.com/opencontainers/image-spec/blob/main/spec.md
--- a/docs/reference/xpkg.md
+++ b/docs/reference/xpkg.md
@@ -88,14 +88,29 @@ may be specific to a given consumer.
 
 If no layer descriptors have an annotation in the form `io.crossplane.xpkg:
 base`, the resultant filesystem from [applying changesets] from all layers will
-be used.
+be used. It is preferred to use layer descriptor annotations.
 
-> Crossplane prefers the usage of layers to define additive package content as
-> it provides a clean mechanism to build an `xpkg` through a series of stages. A
-> valid `xpkg` can be produced and later modified while verifying that the
-> integrity of the existing content is not violated, which ensures that
-> Crossplane's package manager will process the resulting `xpkg` in the same
-> manner as the it would prior to modification.
+#### Motivation
+
+Crossplane prefers the usage of annotated layer descriptors because it allows
+for fetching and processing individual layers, rather than all layers in the
+image. In the event that the image contains a single layer, this overhead is
+minimal. However, larger images with many layers, whether they contain
+third-party `xpkg` content or unrelated data, will result in multiple network
+calls and more data to process.
+
+Crossplane also prefers the usage of annotated layer descriptors to define
+additive package content (i.e. third-party `xpkg` content) as it provides a
+clean mechanism to build an `xpkg` through a series of stages. A valid `xpkg`
+can be produced and later modified while verifying that the integrity of the
+existing content is not violated, which ensures that Crossplane's package
+manager will process the resulting `xpkg` in the same manner as the it would
+prior to modification.
+
+While not explicitly forbidden, modifying content from a preceding layer with
+the `io.crossplane.xpkg` annotation in any subsequent layers is discouraged, as
+it may lead to confusion if a third-party is consuming content from the
+flattened filesystem.
 
 ### Configuration
 
--- a/docs/reference/xpkg.md
+++ b/docs/reference/xpkg.md
@@ -170,12 +170,6 @@ annotations by third-party consumers of Crossplane packages:
   released.
 - `meta.crossplane.io/description`: A one sentence description of the package.
 - `meta.crossplane.io/readme`: A longer description, documentation, etc.
-- `friendly-name.meta.crossplane.io`: A 'human-friendly' name for this package.
-- `friendly-group-name.meta.crossplane.io/<group>`: A "human-friendly" name for
-  an API group defined by this package.
-- `friendly-kind-name.meta.crossplane.io/<kind>.<group>`: A "human-friendly"
-  name for a an API kind defined by this package. Note that the kind is
-  singular, not plural (i.e. instance not instances).
 
 Third party consumers may define additional arbitrary annotations with any key
 and value on any object in a package. All annotations on "meta" types (i.e.
--- a/docs/reference/xpkg.md
+++ b/docs/reference/xpkg.md
@@ -1,0 +1,185 @@
+---
+title: xpkg Specification
+toc: true
+weight: 305
+indent: true
+---
+
+# Overview
+
+Crossplane supports two types of [packages]: Providers and Configurations. These
+packages are distributed as generic [OCI images], which contain [YAML] content
+informing the Crossplane package manager how to alter the state of a cluster by
+installing objects that configure new resource types, and starting controllers
+to reconcile them. An OCI image that contains valid Crossplane package content
+is commonly referred to as an `xpkg` ("ex-package"). This document provides the
+specification for a valid `xpkg`, which can be considered a superset of the
+requirements detailed in [OCI image specification]. It is divided into two broad
+sections: requirements related to OCI image format and requirements related to
+Crossplane `package.yaml` contents.
+
+## OCI Image Format
+
+OCI images are comprised of [manifests], [configuration], and [layers].
+Additionally, an image reference could refer to an image [index], which may
+reference multiple image manifests and is frequently used for multi-platform
+images. A valid Crossplane `xpkg` imposes various requirements on the components
+of an OCI, each of which are described in the following sections.
+
+### Indexes
+
+The components of an `xpkg` that Crossplane interacts with do not contain any
+platform-specific information, so Crossplane is broadly agnostic to the
+formatting of an image index. Crossplane does impose the following requirements
+on an image index:
+
+- At least one (1) manifest MUST be referenced in the manifest descriptor array
+  for a package to be successfully fetched and processed.
+
+> The OCI image specification allows for zero-length manifest descriptor arrays
+> in an index.
+
+The following default behavior when interacting with image indexes is
+implemented in the Crossplane package manager:
+
+- If one manifest is referenced in the image index, the image it points to will
+  be used.
+- If multiple manifests are referenced in the image index, Crossplane will use
+  the `linux/amd64` variant by default.
+
+> It is important to note that the platform of the package image that is used by
+> Crossplane does not necessarily mean that the same platform will be used for
+> the controller if the package is a Provider. The decision of selecting a
+> platform for a Provider controller image is deferred to the configured
+> container runtime.
+
+### Manifests
+
+A manifest defines the layers and configuration of a specific image. Crossplane
+is only concerned with the layer descriptors array in an image manifest and does
+not impose additional requirements on any other portions of the manifest. The
+following requirements are imposed on the layer descriptors array:
+
+- One (1) layer descriptor in the array MAY have an [annotation] with key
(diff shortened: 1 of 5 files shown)


docs/reference/xpkg.md
```diff
@@ -0,0 +1,201 @@
...
+- `meta.crossplane.io/license`: The license under which the package's source is
+  released.
+- `meta.crossplane.io/description`: A one sentence description of the package.
+- `meta.crossplane.io/readme`: A longer description, documentation, etc.
+- `friendly-name.meta.crossplane.io`: A 'human-friendly' name for this package.
+- `friendly-group-name.meta.crossplane.io/<group>`: A "human-friendly" name for
+  an API group defined by this package.
+- `friendly-kind-name.meta.crossplane.io/<kind>.<group>`: A "human-friendly"
+  name for a an API kind defined by this package. Note that the kind is
+  singular, not plural (i.e. instance not instances).
```
hasheddan, line 178: I considered moving this to the objects they are describing directly, but having them on the meta type (and thus the revision) make it much easier for a consumer to build a list of types and their friendly name in a given package.

docs/reference/xpkg.md
```diff
@@ -0,0 +1,201 @@
...
+  after all layer changesets are applied.
+- The `package.yaml` file MUST contain a valid [YAML stream].
+- All other content in either the `xpkg` base layer, or the full image
+  filesystem is ignored by Crossplane.
+
+> The ability to use the image's flattened filesystem is primarily for backwards
+> compatibility and is not encouraged, especially in the event that an image
+> contains more than just `xpkg` related content, due to the fact that
+> accidentally overwriting or modifying the `xpkg` layer contents in subsequent
+> layers when constructing an image could cause the package to be invalid.
```
hasheddan, line 124: I strongly considered dropping this and instead making the "backwards compatibility" requirement that the image only consists of a single (1) layer. This would technically still be a breaking change because today we will honor a package with multiple layers and a `package.yaml` defined at any layer, but in practice, every meaningful `xpkg` we are aware of just includes the single layer. So while it would be a breaking change, it is unlikely to have negative consequences and the existing behavior could be considered "unspecified" rather than "explicitly supported". The real benefit of using the single layer for backwards compatibility is that third-party consumers who wanted to honor the spec for backwards compatibility would still only have to download a single layer for parsing.

docs/reference/xpkg.md
```diff
@@ -0,0 +1,201 @@
...
+The `package.yaml` for Provider packages must adhere to the following
+requirements:
+
+- One (1) and only one `Provider.meta.pkg.crossplane.io` object MUST be defined
+  in the YAML stream.
+- Zero (0) or more `CustomResourceDefinition.apiextensions.k8s.io` objects MAY
+  be defined in the YAML stream.
+- Zero (0) other object types may be defined in the YAML stream.
+
+### Object Annotations
```
hasheddan, line 157: This notably drops the `iconURI` annotation that is defined in https://github.com/crossplane/crossplane/blob/master/design/one-pager-package-annotations.md#proposal

docs/reference/xpkg.md
```diff
@@ -0,0 +1,201 @@
...
+than `base`, but does require the following of the `xpkg` base layer:
+
+- A single file with name `package.yaml` MUST exist in the root directory of the
+  `xpkg` base layer if distinguished, or in the root of the image filesystem
+  after all layer changesets are applied.
+- The `package.yaml` file MUST contain a valid [YAML stream].
+- All other content in either the `xpkg` base layer, or the full image
+  filesystem is ignored by Crossplane.
+
+> The ability to use the image's flattened filesystem is primarily for backwards
```
muvaf, line 120: What do we gain by taking a specific layer into consideration instead of using the resulting filesystem all the time? We've got the tools to work with a specific layer but treating the image as a single filesystem is easier for people to reason about and work with, like just `docker run` and copy something etc.

docs/reference/xpkg.md
```diff
@@ -0,0 +1,201 @@
...
+The `package.yaml` for Configuration packages must adhere to the following
+requirements:
+
+- One (1) and only one `Configuration.meta.pkg.crossplane.io` object MUST be
+  defined in the YAML stream.
+- Zero (0) or more `CompositeResourceDefinition.apiextensions.crossplane.io`
+  objects MAY be defined in the YAML stream.
+- Zero (0) or more `Composition.apiextensions.crossplane.io` objects MAY be
+  defined in the YAML stream.
+- Zero (0) other object types may be defined in the YAML stream.
```
muvaf, line 144: Is there a use case in mind that we'll need to allow arbitrary types in a configuration package?

docs/reference/xpkg.md
```diff
@@ -0,0 +1,201 @@
...
+### Provider Package Requirements
+
+The `package.yaml` for Provider packages must adhere to the following
+requirements:
+
+- One (1) and only one `Provider.meta.pkg.crossplane.io` object MUST be defined
+  in the YAML stream.
+- Zero (0) or more `CustomResourceDefinition.apiextensions.k8s.io` objects MAY
+  be defined in the YAML stream.
+- Zero (0) other object types may be defined in the YAML stream.
```
muvaf, line 155: Is there a use case in mind that we'll need to allow arbitrary types in a provider package?

docs/reference/xpkg.md
muvaf: This would impose a requirement that for any add/remove/update there are two places to make changes, which is fine, but something to keep in mind because it might lead to forgetting to update this file when you add a new resource which is happening today. Though I'm not sure if we have a good way of adding an annotation to CRD through kubebuilder.

muvaf (commented):
My main concern is about us working with a specific layer instead of just treating the image as a filesystem which is what they are used for most of the time.

docs/reference/xpkg.md
hasheddan: @muvaf I think this is called out in the following sentences, but the things we gain include:
- If subsequent layers overwrite content the package is still valid. This means an if image is built incrementally, there is less chance of ending up with an image that does not work as a Crossplane package. While it may seem like this is unlikely, in the event the `package.yaml` in the root is overwritten it would be somewhat difficult to debug, and if you are using extensions there is potentially "3rd party package content" outside of the `package.yaml`.
- Specifying a single layer means that we don't have to download the entire image to parse the package contents. This becomes especially useful if we move to using combined controller / package images more frequently, and/or with using the extension mechanisms described.

It is also worth pointing out that this specification explicitly allows for _not_ specifying a single layer, which means the path that you are talking about will still work as expected. For the reasons mentioned above, using the flattened filesystem approach may result in more inefficient parsing (only at the beginning, we can opt to only cache the package.yaml content).

docs/reference/xpkg.md
hasheddan: There is not today, so I would opt to explicitly disallow it until we have a use case.

docs/reference/xpkg.md
hasheddan: There is not today, so I would opt to explicitly disallow it until we have a use case.

docs/reference/xpkg.md
hasheddan: @muvaf yeah I am not super excited about this pattern, though its worth noting that we do already employ it across providers today. These fields are optional so package functionality would not be impacted by forgetting to add. Tooling could also check for types that do not have friendly names and emit warnings if desired.

docs/reference/xpkg.md
```diff
@@ -0,0 +1,201 @@
...
+- One (1) layer descriptor in the array MAY have an [annotation] with key
+  `io.crossplane.xpkg` and value `base`.
+- Any number of layer descriptors in the array MAY have an annotation with key
+  `io.crossplane.xpkg` and arbitrary value. Whether multiple layer descriptors
+  may have the same value is left to the specification of the consumer of those
+  layers.
+
+> As evidenced by the fact that annotations are provided as a map of
+> _string-string_, no single descriptor will contain multiple
+> `io.crossplane.xpkg` annotations.
```
negz, line 82: Is something like `io.crossplane.xpkg/type` allowed as a key? If so, would that level of specificity make sense to allow for other xpkg related annotations?

docs/reference/xpkg.md
negz: I feel slightly better about what is proposed here - while I take your point that in practice pretty much everything is today a single layer I'd rather optimise for broader compatibility here (e.g. supporting theoretical existing packages w/multiple layers) vs optimising for guaranteeing the ability to always download only one layer.

docs/reference/xpkg.md
negz: Could we advise, or require, that the `package.yaml` is never mutated or removed in subsequent layers? I like that we can keep it in an easily identified layer for downloads and caching, but would prefer to make sure it's always visible there in the resulting FS to avoid potentially surprising users.

docs/reference/xpkg.md
negz: Breadcrumbs for https://github.com/crossplane/crossplane/issues/2262. Obviously this doesn't exist today so no action needed, but if we did proceed with this idea I imagine provider packages would deliver MRDs rather than CRDs at some point in the future.

docs/reference/xpkg.md
negz: Does it make sense for this document to absorb or replace the package annotations one pager?

docs/reference/xpkg.md
negz: I don't love this pattern of keeping all the friendly name annotations in the package metadata either. One thing that makes it a _little_ more tolerable in my mind is that in theory we might be able to derive friendly names in a lot of cases, so perhaps they wouldn't be needed everywhere.

negz (approved):
Dropped a few thoughts/comments but this looks good to me overall. I imagine we might need to track some work to make `kubectl crossplane` and the package manager honor the new OCI annotations?

docs/reference/xpkg.md
hasheddan: @negz would you advocate for dropping these on the CRDs / XRDs themselves and prohibiting use on the package metadata, or just supporting both? I ask because starting off with them on the package metadata does not prohibit us from eventually also supporting them on the CRDs / XRDs. However, I imagine we would only want to support `friendly-kind-name` there as we wouldn't want the same group to be defined with different friendly names on multiple CRDs / XRDs.

docs/reference/xpkg.md
hasheddan: @negz yes I think so 👍🏻

docs/reference/xpkg.md
muvaf: > Specifying a single layer means that we don't have to download the entire image to parse the package contents.

Thank you for explaining, makes a lot of sense to me.

What do you think about having a `Motivation` section to put those reasonings so that if people discuss changing it, they can have a clearer picture of why we do this?

docs/reference/xpkg.md
muvaf: I agree with keeping the compatibility. A lot of tools and some registries do assume as if the image is just a file blob, which is fair IMO.

muvaf (approved):
Thank you!

docs/reference/xpkg.md
hasheddan: > Could we advise, or require, that the package.yaml is never mutated or removed in subsequent layers?

@negz I think I would push back on requiring that, and I think this gets to the broader purpose of this spec: defining what a valid package is, rather than how to build one. From that perspective, mandating that the `package.yaml` not be changed would mean that we would need to reject a package in Crossplane in the following scenario:
- None of the layers are annotated
- There are multiple layers
- The flattened file system contains a valid `package.yaml` at root
- The `package.yaml` was touched in multiple layers

Actually determining that the `package.yaml` was touched in multiple layers would be a fairly cumbersome process that I don't see providing a lot of value. I would advise customers to use tooling that doesn't allow that to happen, and I think any tooling in the Crossplane community should take that approach, but one of the reasons that we are maintaining the ability to use the flattened filesystem at all is that folks may use something like Docker to build their packages and mandating that the `package.yaml` is put into that filesystem in a certain manner makes that process more complicated IMO.

I do think warning against potentially unexpected behavior in documentation on building packages makes a lot of sense though 👍🏻 

> What do you think about having a Motivation section to put those reasonings so that if people discuss changing it, they can have a clearer picture of why we do this?

@muvaf great idea! Will add 👍🏻

docs/reference/xpkg.md
hasheddan: @negz thanks for pointing that out! I am a pretty big fan of the MRD idea and that would definitely have an impact on this spec in the future -- I imagine it would be possible to support both CRDs and MRDs alongside each other in a package, but the complexity of doing so would likely not be worth it. I _could_ see a world in which this spec was adjusted as part of that effort to say "Zero (0) or more CRDs OR Zero (0) or more MRDs", but as you said we can cross that bridge when we come to it :)

docs/reference/xpkg.md
hasheddan: @negz my reading of [the spec](https://github.com/opencontainers/image-spec/blob/main/annotations.md#rules) is no -- the key must be a "reverse domain name". The value on the other hand is free form. In practice I doubt any tooling rejects annotation keys of a different form (i.e. I would be that they are treated as opaque strings), but I would not want to deviate from the spec due to the fact that Crossplane package images could potentially be deemed non-conforming by tooling in the future.

docs/reference/xpkg.md
hasheddan: FWIW, I would not be opposed to dropping annotations entirely and saying that any extra metadata is defined in 3rd-party _layer extension_ mechanism.

negz: > @negz would you advocate for dropping these on the CRDs / XRDs themselves and prohibiting use on the package metadata, or just supporting both?

Not necessarily. I don't particularly like either approach.

docs/reference/xpkg.md
negz: @hasheddan I'm concerned about it being able to do things that seem clever/surprising to those (presumably majority of) folks who don't have as deep a knowledge of OCI images as yourself. I would certainly find it surprising if I were to inspect a Crossplane package naively (i.e. by looking at the 'complete' set of layers) and find that either:

* There was no `package.yaml`.
* There was a `package.yaml` that was different than what Crossplane was actually using.

Perhaps the requirement should be that the `package.yaml` in the annotated layer must match the 'final' `package.yaml` in the flattened filesystem?

I feel comfortable about having the spec say not to do something without necessarily writing code to enforce that the thing is not done - especially within the actual package manager (as opposed to a linter where having to have all the layers in order to validate it seems less onerous).

hasheddan: > I feel comfortable about having the spec say not to do something without necessarily writing code to enforce that the thing is not done

I don't think I am aligned here as it could be viewed as Crossplane not adhering to its own spec if there is explicit behavior defined that is not honored. I would be surprised as a user if a tool didn't let me build something because it was adhering to the spec, but Crossplane would let me install it. Also keep in mind the following points:
- If someone is less familiar with OCI images and they are using a tool like Docker, they likely are not annotating a layer and the `package.yaml` in the flattened filesystem will be what is used by default.
- If someone is using Crossplane tooling that annotated the layer then they likely are either building a fresh image that is only a package (i.e. single layer) or they are modifying an existing image. In the former case there is no chance for the `package.yaml` to be altered so it will be reflected in flattened filesystem (which is the same as the single layer filesystem). In the latter case, the tooling can determine exactly how it wants to add the `package.yaml`, potentially being more specific than Crossplane requires.

With this in mind, I would prefer for the spec to be exact about what will be accepted as a valid package, and allow users to dictate the UX they want for building.

docs/reference/xpkg.md
hasheddan: After some offline discussion, I am going to opt to drop the friendly name annotations for the time-being as Crossplane does not use them directly for any purpose today and their inclusion on the meta package type feels somewhat clunky. If we can justify a valid usage and design in the future we can opt to add them.

docs/reference/xpkg.md
hasheddan: After some offline discussion I am going to keep the spec as-is, but will add a note mentioning that mutating the `package.yaml` in subsequent layers could lead to confusion for users. Crossplane will not explicitly disallow it (i.e. packages will not be rejected if they are valid according to the requirements outlines in this document).

Thanks for the discussion on this one @negz and @muvaf!

PR #2786, opened 2021-12-23
Apply merge options from patch policy

### LINKED ISSUE #2690: MergeOption: appendSlice is overwriting instead of appending
### What happened?

Hi all, i guess the mergeOption: appendSlice is broken.
Seems like the appendSlice is not respected and the tagset is overwritten.
Is this a known issue?
Thanks

### How can we reproduce it?
I have the following composition with patches

```yaml
  resources:
    - base: 
        kind: Bucket
        metadata:
          name: METADATA_UID
        spec:
          deletionPolicy: Delete
          providerConfigRef:
            name: aws-provider-config
          forProvider:
            acl: private
            locationConstraint: REGION
            versioningConfiguration:
              status: Enabled
              rules:
                    sseAlgorithm: AES256
            publicAccessBlockConfiguration:
              blockPublicAcls: true
              blockPublicPolicy: true
              ignorePublicAcls: true
              restrictPublicBuckets: true
            tagging:
              tagSet: 
                - key: namespace
                  value: MY_NAMESPACE
          writeConnectionSecretToRef:
            name: METADATA_NAME-MR-s3bucket-managed-resource
            namespace: crossplane-system
      patches:
          fromFieldPath: spec.parameters.tags
          toFieldPath: spec.forProvider.tagging.tagSet
          policy:
            mergeOptions:
              appendSlice: true
              keepMapValues: true
```

My claim has the following:
```yaml
    tags:
      - key: another-tag
        value: with-value
```

So the following would be expected result:
```yaml
            tagging:
              tagSet: 
                - key: namespace
                  value: MY_NAMESPACE
                - key: another-tag
                  value: with-value
```

But i get the following:
```yaml
    tagging:
      tagSet:
      - key: another-tag
        value: with-value
```

Crossplane version: 

Crossplane version: 1.4.1

Cloud provider or hardware configuration - aws-provider v0.20.0-rc
Kubernetes version (use kubectl version) - v1.21.2-eks-0389ca3
Kubernetes distribution (e.g. Tectonic, GKE, OpenShift) - EKS
OS (e.g. from /etc/os-release) - Amazon Linux

EdgeJ:

Prior to this commit, merge options were not applied to the
paved.MergeValue method and were effectively thrown away. This commit
passes in any options from Policy.MergeOptions, passing in nil if none
were supplied so that fields can be correctly preserved and merged.

Fixes #2690

I have:

Unit test added for preserving map values and arraySlice merges tested with locally running Crossplane and applying XRs to create MRs with appended fields.

### DIFF
--- a/apis/apiextensions/v1/composition_patches.go
+++ b/apis/apiextensions/v1/composition_patches.go
@@ -166,13 +166,13 @@ func (c *Patch) applyTransforms(input interface{}) (interface{}, error) {
 // patchFieldValueToObject, given a path, value and "to" object, will
 // apply the value to the "to" object at the given path, returning
 // any errors as they occur.
-func patchFieldValueToObject(fieldPath string, value interface{}, to runtime.Object) error {
+func patchFieldValueToObject(fieldPath string, value interface{}, to runtime.Object, mergeOpts *xpv1.MergeOptions) error {
 	paved, err := fieldpath.PaveObject(to)
 	if err != nil {
 		return err
 	}
 
-	if err := paved.MergeValue(fieldPath, value, nil); err != nil {
+	if err := paved.MergeValue(fieldPath, value, mergeOpts); err != nil {
 		return err
 	}
 
@@ -205,13 +205,21 @@ func (c *Patch) applyFromFieldPathPatch(from, to runtime.Object) error {
 		return err
 	}
 
+	var mergeOpts *xpv1.MergeOptions
+
+	if c.Policy == nil {
+		mergeOpts = nil
+	} else {
+		mergeOpts = c.Policy.MergeOptions
+	}
+
 	// Apply transform pipeline
 	out, err := c.applyTransforms(in)
 	if err != nil {
 		return err
 	}
 
-	return patchFieldValueToObject(*c.ToFieldPath, out, to)
+	return patchFieldValueToObject(*c.ToFieldPath, out, to, mergeOpts)
 }
 
 // applyCombineFromVariablesPatch patches the "to" resource, taking a list of
@@ -276,7 +284,7 @@ func (c *Patch) applyCombineFromVariablesPatch(from, to runtime.Object) error {
 		return err
 	}
 
-	return patchFieldValueToObject(*c.ToFieldPath, out, to)
+	return patchFieldValueToObject(*c.ToFieldPath, out, to, nil)
 }
 
 // IsOptionalFieldPathNotFound returns true if the supplied error indicates a
--- a/apis/apiextensions/v1/composition_patches.go
+++ b/apis/apiextensions/v1/composition_patches.go
@@ -166,13 +166,13 @@ func (c *Patch) applyTransforms(input interface{}) (interface{}, error) {
 // patchFieldValueToObject, given a path, value and "to" object, will
 // apply the value to the "to" object at the given path, returning
 // any errors as they occur.
-func patchFieldValueToObject(fieldPath string, value interface{}, to runtime.Object, mergeOpts *xpv1.MergeOptions) error {
+func patchFieldValueToObject(fieldPath string, value interface{}, to runtime.Object, mo *xpv1.MergeOptions) error {
 	paved, err := fieldpath.PaveObject(to)
 	if err != nil {
 		return err
 	}
 
-	if err := paved.MergeValue(fieldPath, value, mergeOpts); err != nil {
+	if err := paved.MergeValue(fieldPath, value, mo); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


apis/apiextensions/v1/composition_patches.go
```diff
@@ -276,7 +284,7 @@ func (c *Patch) applyCombineFromVariablesPatch(from, to runtime.Object) error {
 		return err
 	}
 
-	return patchFieldValueToObject(*c.ToFieldPath, out, to)
+	return patchFieldValueToObject(*c.ToFieldPath, out, to, nil)
```
negz, line 287: I imagine we might want to pass merge options here too at some point; in fact it's a little misleading that `mergeOptions` are currently ignored by patches of `type: CombineFromComposite` and `type: CombineToComposite`. I don't think it makes sense to implement now though, given that AFAICT it's not actually possible for the combine patch types to produce a map or array.

apis/apiextensions/v1/composition_patches.go
```diff
@@ -166,13 +166,13 @@ func (c *Patch) applyTransforms(input interface{}) (interface{}, error) {
 // patchFieldValueToObject, given a path, value and "to" object, will
 // apply the value to the "to" object at the given path, returning
 // any errors as they occur.
-func patchFieldValueToObject(fieldPath string, value interface{}, to runtime.Object) error {
+func patchFieldValueToObject(fieldPath string, value interface{}, to runtime.Object, mergeOpts *xpv1.MergeOptions) error {
```
negz, line 169: I'd prefer `mo` here. It's what the argument to `MergeValue` is named, and I think the type gives enough context for what they are per https://github.com/crossplane/crossplane/blob/b67e7e0/CONTRIBUTING.md#use-descriptive-variable-names-sparingly


```suggestion
func patchFieldValueToObject(fieldPath string, value interface{}, to runtime.Object, mo *xpv1.MergeOptions) error {
```

apis/apiextensions/v1/composition_patches.go
```diff
@@ -166,13 +166,13 @@ func (c *Patch) applyTransforms(input interface{}) (interface{}, error) {
...
 	paved, err := fieldpath.PaveObject(to)
 	if err != nil {
 		return err
 	}
 
-	if err := paved.MergeValue(fieldPath, value, nil); err != nil {
+	if err := paved.MergeValue(fieldPath, value, mergeOpts); err != nil {
```
negz, line 175:
```suggestion
	if err := paved.MergeValue(fieldPath, value, mo); err != nil {
```

apis/apiextensions/v1/composition_patches.go
```diff
@@ -205,13 +205,21 @@ func (c *Patch) applyFromFieldPathPatch(from, to runtime.Object) error {
...
+	}
+
 	// Apply transform pipeline
 	out, err := c.applyTransforms(in)
 	if err != nil {
 		return err
 	}
 
-	return patchFieldValueToObject(*c.ToFieldPath, out, to)
+	return patchFieldValueToObject(*c.ToFieldPath, out, to, mergeOpts)
```
negz, line 222:
```suggestion
	return patchFieldValueToObject(*c.ToFieldPath, out, to, mo)
```

apis/apiextensions/v1/composition_patches.go
```diff
@@ -205,13 +205,21 @@ func (c *Patch) applyFromFieldPathPatch(from, to runtime.Object) error {
 		return err
 	}
 
+	var mergeOpts *xpv1.MergeOptions
+
+	if c.Policy == nil {
+		mergeOpts = nil
+	} else {
+		mergeOpts = c.Policy.MergeOptions
+	}
```
negz, line 214:
```suggestion
	var mo *xpv1.MergeOptions
	if c.Policy != nil {
		mo = c.Policy.MergeOptions
	}
```

Nit: I believe this is a slightly more succinct version of the same thing.

negz (commented):
Thanks for catching this @EdgeJ! I wonder if this was a regression or just never worked. 🤔

apis/apiextensions/v1/composition_patches.go
EdgeJ: Agreed that passing them in here will make sense once the combine types support arrays/maps.  As you pointed out, I didn't implement it with this PR because of the fact that it's basically a no-op at this point.

negz (approved):
Thanks for fixing this @EdgeJ!

hasheddan (commented):
@negz should this also be backported to `release-1.5` and `release-1.4`?

Issue #2335, opened 2021-05-28
Composition patches: Introduce a method to iterate over arrays of objects


rozcietrzewiacz:
### What problem are you facing?

Some managed resources include arrays of `Object` type. Some examples:
 - `ResourceRecordSet` in `route53.aws.crossplane.io/v1alpha1` at [`spec.forProvider.resourceRecords`](https://doc.crds.dev/github.com/crossplane/provider-aws/route53.aws.crossplane.io/ResourceRecordSet/v1alpha1@v0.18.1#spec-forProvider-resourceRecords) (single field "value")
 - `FargateProfile` in `eks.aws.crossplane.io/v1alpha1`  at  [`spec.forProvider.selectors`](https://doc.crds.dev/github.com/crossplane/provider-aws/eks.aws.crossplane.io/FargateProfile/v1alpha1@v0.18.1#spec-forProvider-selectors) - with `labels` and `namespace` fields
 - `CosmosDBAccount` in `database.azure.crossplane.io/v1alpha3` at [`spec.forProvider.properties.locations`](https://doc.crds.dev/github.com/crossplane/provider-azure/database.azure.crossplane.io/CosmosDBAccount/v1alpha3@v0.16.1#spec-forProvider-properties-locations) (3 fields)

Since Platform Builders want to simplify and/or restrict the interface for Platform Consumers, they should be able to produce `Compositions` that would consume arrays of simple types to fill in just one of the fields in each such `Object`.

For instance: To allow consumers create a set of route53 DNS records, the consumer should be able to supply those as array of `Strings` instead of an array of `Objects` with a single `value` field.

An even more obvious example could be drawn in case of `FargateProfile` selectors mentioned above. Platform Builders could have a solid reason to allow Consumers define `labels`, but not `namespace` in their `CompositeResource`.

To achieve this, the Composition currently has to do things like 
```
- fromFieldPath: "spec.records[0]"
  toFieldPath: "spec.forProvider.resourceRecords[0].value"
- fromFieldPath: "spec.records[1]"
  toFieldPath: "spec.forProvider.resourceRecords[1].value"
- fromFieldPath: "spec.records[2]"
  toFieldPath: "spec.forProvider.resourceRecords[2].value"
- fromFieldPath: "spec.records[3]"
  toFieldPath: "spec.forProvider.resourceRecords[3].value"
- fromFieldPath: "spec.records[4]"
  toFieldPath: "spec.forProvider.resourceRecords[4].value"
[...]

PR #2791, opened 2021-12-30
Patch multiple array fields when passed a wildcarded path

### LINKED ISSUE #2784: Patch a field into multiple arrays

There are times where an array of objects have a common data point that is either cumbersome to code by hand, or must be derived from some other place.  A specific example I am working on is creating an S3 BucketPolicy, using the ARN derived from a Bucket created by the same composition:

```yaml
  resources:
  - name: bucket
    base:
      kind: Bucket
      spec:
        forProvider:
          acl: private
          locationConstraint: us-east-1
          publicAccessBlockConfiguration:
            # block all public access by default
            blockPublicAcls: true
            blockPublicPolicy: true
            ignorePublicAcls: true
            restrictPublicBuckets: true
            rules:
                sseAlgorithm: AES256
          tagging:
            tagSet:
            - key: createdBy
              value: crossplane
    patches:
    # get the bucket ARN to supply to the policy resource
    - type: ToCompositeFieldPath
      toFieldPath: metadata.annotations[crossplane.io/bucket-arn]
  - name: bucketPolicy
    base:
      apiVersion: s3.aws.crossplane.io/v1alpha3
      kind: BucketPolicy
      spec:
        forProvider:
          bucketNameSelector:
            matchControllerRef: true  # select the bucket created by this composition
          policy:
            version: "2012-10-17"
            statements:
            - action:
              - s3:ListBucket*
              - s3:GetBucket*
              - s3:PutBucket*
              effect: Allow
              principal:
                awsPrincipals: []
              resource: []
          region: us-east-1
    patches:
    - fromFieldPath: spec.parameters.bucketPolicy.allowedIAMRoleArn
      toFieldPath: spec.forProvider.policy.statements[0].principal.awsPrincipals[0].iamRoleArn
    - fromFieldPath: metadata.annotations[crossplane.io/bucket-arn]
      toFieldPath: spec.forProvider.policy.statements[0].resource[0]
```

In this case, it would be useful to allow users to specify additional policy statements to append to the statements created by default in the composition, but there is no good way to apply the calculated bucket ARN value to multiple array items.  

Add a patch type that can apply a single object element to an indeterminate number of object arrays.  For example:

```yaml
resources:
- name: sample-bucket
  spec:
    forProvider:
      policy:
        statements:
        - resource: []
  patches:
  # append array elements of policy statements to the existing default
  - fromFieldPath: spec.parameters.bucketPolicy.additionalPolicies
    toFieldPath: spec.forProvider.policy.statements
    policy:
      mergeOptions:
        appendSlice: true
  # insert the annotation added to the XR into each array element in statements
  - type: ArrayInsert
    fromFieldPath: metadata.annotations[crossplane.io/bucket-arn]
    toFieldPath: spec.forProvider.policy.statements.resource[0]
```

I think this is _somewhat_ related to #2335 but I think the functionality differs somewhat with that issue being more about programmatically looping over input arrays and this being more the case of straightforward patching, but on multiple items.

EdgeJ:
This change allows patches of the FromCompositeFieldPath and
ToCompositeFieldPath types to patch from one field in the "from"
resource to multiple object fields nested inside an array in the "to" resource.
This allows an indefinite number of arrays to be patched without
needing to specify a patch per each individual element.

Fixes #2784

I have:

Unit tests written to check the new functionality and I ran Crossplane locally to deploy real resources using the multiple wildcard patching.

### DIFF
--- a/apis/apiextensions/v1/composition_patches.go
+++ b/apis/apiextensions/v1/composition_patches.go
@@ -19,6 +19,7 @@ package v1
 import (
 	"fmt"
 	"reflect"
+	"strings"
 
 	"github.com/pkg/errors"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -37,6 +38,7 @@ const (
 	errFmtCombineStrategyNotSupported = "combine strategy %s is not supported"
 	errFmtCombineConfigMissing        = "given combine strategy %s requires configuration"
 	errFmtCombineStrategyFailed       = "%s strategy could not combine"
+	errExpandingArrayFieldPaths       = "Cannot expand ToFieldPath %s"
 )
 
 // A PatchType is a type of patch.
@@ -163,6 +165,33 @@ func (c *Patch) applyTransforms(input interface{}) (interface{}, error) {
 	return input, nil
 }
 
+// patchFieldValueToMultiple, given a path with wildcards in an array index,
+// expands the arrays paths in the "to" object and patches the value into each
+// of the resulting fields, returning any errors as they occur.
+func patchFieldValueToMultiple(fieldPath string, value interface{}, to runtime.Object) error {
+	paved, err := fieldpath.PaveObject(to)
+	if err != nil {
+		return err
+	}
+
+	arrayFieldPaths, err := paved.ExpandWildcards(fieldPath)
+	if err != nil {
+		return err
+	}
+
+	if len(arrayFieldPaths) == 0 {
+		return errors.Errorf(errExpandingArrayFieldPaths, fieldPath)
+	}
+
+	for _, field := range arrayFieldPaths {
+		if err := paved.MergeValue(field, value, nil); err != nil {
+			return err
+		}
+	}
+
+	return runtime.DefaultUnstructuredConverter.FromUnstructured(paved.UnstructuredContent(), to)
+}
+
 // patchFieldValueToObject, given a path, value and "to" object, will
 // apply the value to the "to" object at the given path, returning
 // any errors as they occur.
@@ -214,6 +243,11 @@ func (c *Patch) applyFromFieldPathPatch(from, to runtime.Object) error {
 	out, err := c.applyTransforms(in)
 	if err != nil {
 		return err
+	}
+
+	// Patch all expanded fields if the ToFieldPath contains wildcards
+	if strings.Contains(*c.ToFieldPath, "[*]") {
+		return patchFieldValueToMultiple(*c.ToFieldPath, out, to)
 	}
 
 	return patchFieldValueToObject(*c.ToFieldPath, out, to, mo)
--- a/apis/apiextensions/v1/composition_patches.go
+++ b/apis/apiextensions/v1/composition_patches.go
@@ -168,7 +168,7 @@ func (c *Patch) applyTransforms(input interface{}) (interface{}, error) {
 // patchFieldValueToMultiple, given a path with wildcards in an array index,
 // expands the arrays paths in the "to" object and patches the value into each
 // of the resulting fields, returning any errors as they occur.
-func patchFieldValueToMultiple(fieldPath string, value interface{}, to runtime.Object) error {
+func patchFieldValueToMultiple(fieldPath string, value interface{}, to runtime.Object, mo *xpv1.MergeOptions) error {
 	paved, err := fieldpath.PaveObject(to)
 	if err != nil {
 		return err
@@ -184,7 +184,7 @@ func patchFieldValueToMultiple(fieldPath string, value interface{}, to runtime.O
 	}
 
 	for _, field := range arrayFieldPaths {
-		if err := paved.MergeValue(field, value, nil); err != nil {
+		if err := paved.MergeValue(field, value, mo); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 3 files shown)


cdenneen:
@negz Since this seems to cover 2 open issues #2784 and #2335 can this be reviewed and implemented?
@EdgeJ looks like needs to be rebased to resolve current conflicts?

negz:
@turkenh Could you take a look at this one when you get a chance? I think you implemented `fieldpath.ExpandWildcards`.

A few thoughts/questions:

* Does `ExpandWildcards` support expanding objects as well as arrays? The `[*]` syntax implies that it does.
* What happens when both `FromFieldPath` and `ToFieldPath` have wildcards?

EdgeJ:
I'm not sure about expanding objects vs arrays, but I'm pretty sure for the second question we're only looking at `ToFieldPath` here: https://github.com/crossplane/crossplane/pull/2791/files#diff-fdd289c27b8b59d09910862c76402dd0be37b082140e139f9b36afeb6e987a21R249-R251

So, I believe that wildcards in `FromFieldPath` would be interpreted literally.

apis/apiextensions/v1/composition_patches.go
```diff
@@ -216,6 +245,11 @@ func (c *Patch) applyFromFieldPathPatch(from, to runtime.Object) error {
 		return err
 	}
 
+	// Patch all expanded fields if the ToFieldPath contains wildcards
+	if strings.Contains(*c.ToFieldPath, "[*]") {
```
turkenh, line 249: [ExpandWildcards returns the same path](https://github.com/crossplane/crossplane-runtime/blob/edc27bfc3ed3c672c0cd5302c80ef5c49a6a7e81/pkg/fieldpath/paved_test.go#L844-L875) if no wildcard exists, so, we could always use the same code path without this check.

However, I am on the fence regarding performance. If this `strings.Contains` check works well for all cases, I agree that this is a good optimization.

apis/apiextensions/v1/composition_patches.go
```diff
@@ -37,6 +38,7 @@ const (
 	errFmtCombineStrategyNotSupported = "combine strategy %s is not supported"
 	errFmtCombineConfigMissing        = "given combine strategy %s requires configuration"
 	errFmtCombineStrategyFailed       = "%s strategy could not combine"
+	errExpandingArrayFieldPaths       = "Cannot expand ToFieldPath %s"
```
turkenh, line 41:
```suggestion
	errFmtExpandingArrayFieldPaths       = "cannot expand ToFieldPath %s"
```

Errors should not be capitalized and to follow the `errFmt` convention since it is a formatted string.

docs/reference/composition.md
```diff
@@ -405,6 +405,34 @@ field.
   toFieldPath: status.zone
 ```
 
+Both `FromCompositeFieldPath` and `ToCompositeFieldPath` can take a wildcarded
+field path containing arrays and patch each array element on a specified
+field.
```
turkenh, line 410: Looks like this is not correct. `fromFieldPath` cannot take a wildcard field path as you also stated in https://github.com/crossplane/crossplane/pull/2791#issuecomment-1127853801

UX-wise, we should try supporting this, but I don't think it makes sense to have wildcards in `fromFieldPath` in practice. 

In addition to documenting this properly, I would also suggest adding a check that throws an error if `fromFieldPath` has a wildcard in it since this could be quite confusing.

turkenh (changes requested):
@EdgeJ thank you for your contribution and sorry for the late review. 

I believe this is not far to get merged, just left a couple of comments which require some small changes.

docs/reference/composition.md
EdgeJ: Yep this language is totally wrong. Fixed.

apis/apiextensions/v1/composition_patches.go
EdgeJ: Fixed.

apis/apiextensions/v1/composition_patches.go
EdgeJ: I didn't do any performance benchmarking, but my presumption was that using a check to avoid using the multiple patching would be more performant. Another factor that makes it good to do this is that the current implementation of `patchFieldValuetoMultiple` only works if the supplied field already exists in the defined MR unlike the implementation of `patchFieldvalueToObject`, which inserts a new field if it doesn't already exist.

docs/reference/composition.md
EdgeJ: Ah, I figured out why it didn't seem to make any sense. Wildcards _do_ work for both `FromCompositeFieldPath` and `ToCompositeFieldPath` type patches, but _only_ for the `toFieldPath` parameter in either type. (Revisiting months old work is difficult when trying to remember what you were doing). I'll clarify the language to make it clear that wildcards in `toFieldPath` are the only valid input in both of these cases

apis/apiextensions/v1/composition_patches.go
negz: Performance isn't of particular concern here; we should optimize for simpler and more readable code. We can always tune for performance once we notice an issue.

apis/apiextensions/v1/composition_patches.go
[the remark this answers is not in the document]
EdgeJ: Already fixed in a previous commit. Probably a typo on my part, but the context is interesting. I hadn't thought of messages being concatenated. I mostly left uncapitalized from cargo-culting other examples.

shanproofpoint:
This is very much anticipated. Thank you so much for your contribution @EdgeJ !

apis/apiextensions/v1/composition_patches.go
turkenh: > Another factor that makes it good to do this is that the current implementation of patchFieldValuetoMultiple only works if the supplied field already exists in the defined MR unlike the implementation of patchFieldvalueToObject, which inserts a new field if it doesn't already exist.

Ok, noticed this as well. If we always used expand wildcards, we wouldn't be able to patch labels unless the object to be patched has already some labels.

turkenh (commented):
Thanks for your patience @EdgeJ, one last comment before approval.

turkenh (approved):

PR #2811, opened 2022-01-05
Make release template (arguably) less onerous

negz:

When running a release I prefer to be able to simply check off boxes via the GitHub UI (i.e. rendered, not in markdown). I find that having to keep the issue open in markdown mode while I insert links to each step to be quite onerous. In practice GitHub automatically cross references most things (e.g. PRs and commits) as long as they're opened with a reference back to the release issue. Anything else (e.g. CI runs, Tweets etc) would (in my opinion) be quicker to drop in as a comment on the release issue.

Specifically this PR:

* Removes the notes section. In my experience we rarely use it, and we can always add one in if needed.
* Removes the list of example artifacts in favor of referencing an older release issue.
* Removes all the links that the author is expected to fill in.

I have:

N/A

### DIFF
--- a/.github/ISSUE_TEMPLATE/release.md
+++ b/.github/ISSUE_TEMPLATE/release.md
@@ -16,13 +16,10 @@ For example:
 Please assign the release manager to the issue.
 -->
 
-### Steps
-<!--
-Please complete the following steps in order. Links should be populated at the
-bottom of this section.
--->
-
-This issue can be closed when we have:
+This issue can be closed when we have completed the following steps (in order).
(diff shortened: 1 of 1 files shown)


muvaf (approved):
I had similar thoughts in 1.5 release but couldn't get around to it, thank you!

hasheddan (approved):
Thanks @negz! Agree this will be easier for folks 👍🏻 I was considering trying out the new [template issues](https://github.blog/changelog/2021-06-23-issues-forms-beta-for-public-repositories/), but it looks like you only get the template view at issue creation time, and I like that we create these issues prior to the steps actually being completed today

PR #2816, opened 2022-01-05
Docs: Added all active providers with Github stars >= 10

luebken:
Signed-off-by: Matthias Luebken <matthias.luebken@gmail.com>

Ton increase visibility of provider I've added the new Terrajet providers and all providers with  Github stars >= 10

I have:

### DIFF
--- a/docs/concepts/providers.md
+++ b/docs/concepts/providers.md
@@ -35,6 +35,7 @@ Here is the list of prominent providers:
 | Provider IBM Cloud | [GitHub Repo][provider-ibm-cloud] | [API Reference][provider-ibm-cloud-api]  |  |
 | Provider Argocd | [GitHub Repo][provider-argocd] | [API Reference][provider-argocd-api]  |  |
 | Provider Styra | [GitHub Repo][provider-styra] | [API Reference][provider-styra-api]  |  |
+| Provider Cloudflare | [GitHub Repo][provider-cloudflare] | [API Reference][provider-cloudflare-api]  |  |
 
 
 ## Installing Providers
@@ -163,3 +164,5 @@ will attempt to use a `ProviderConfig` named `default`.
 [provider-argocd-api]: https://doc.crds.dev/github.com/crossplane-contrib/provider-argocd
 [provider-styra]: https://github.com/crossplane-contrib/provider-styra
 [provider-styra-api]: https://doc.crds.dev/github.com/crossplane-contrib/provider-styra
+[provider-cloudflare]: https://github.com/crossplane-contrib/provider-cloudflare
+[provider-cloudflare-api]: https://doc.crds.dev/github.com/crossplane-contrib/provider-cloudflare
--- a/docs/concepts/providers.md
+++ b/docs/concepts/providers.md
@@ -7,41 +7,34 @@ indent: true
 
 # Providers
 
-Providers extend Crossplane to enable infrastructure resource provisioning. In
-order to provision a resource, a Custom Resource Definition (CRD) needs to be
-registered in your Kubernetes cluster and its controller should be watching the
-Custom Resources those CRDs define. Provider packages contain many Custom
-Resource Definitions and their controllers.
+Providers are Crossplane packages that bundle a set of [Managed
+Resources][managed-resources] and their respective controllers to allow
+Crossplane to provision the respective infrastructure resource.
 
 Here is the list of prominent providers:
 
-### AWS Provider
+|   |   |   |   |
+|---|---|---|---|
+| Provider AWS  |  [GitHub Repo](https://github.com/crossplane/provider-aws) | [API Reference](https://doc.crds.dev/github.com/crossplane/provider-aws)  |  [Amazon Web Services (AWS) IAM User] |
+| Provider Jet AWS  |  [GitHub Repo](https://github.com/crossplane-contrib/provider-jet-aws) | [API Reference](https://doc.crds.dev/github.com/crossplane-contrib/provider-jet-aws) |   | 
+| Provider GCP |  [GitHub Repo](https://github.com/crossplane/provider-gcp) | [API Reference](https://doc.crds.dev/github.com/crossplane/provider-gcp) |   [Google Cloud Platform (GCP) Service Account] | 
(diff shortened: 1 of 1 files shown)


docs/concepts/providers.md
```diff
@@ -7,41 +7,34 @@ indent: true
...
+| Provider Jet AWS  |  [GitHub Repo](https://github.com/crossplane-contrib/provider-jet-aws) | [API Reference](https://doc.crds.dev/github.com/crossplane-contrib/provider-jet-aws) |   | 
+| Provider GCP |  [GitHub Repo](https://github.com/crossplane/provider-gcp) | [API Reference](https://doc.crds.dev/github.com/crossplane/provider-gcp) |   [Google Cloud Platform (GCP) Service Account] | 
+| Provider Jet GCP  |  [GitHub Repo](https://github.com/crossplane-contrib/provider-jet-gcp) | [API Reference](https://doc.crds.dev/github.com/crossplane-contrib/provider-jet-gcp)  |   | 
+| Provider Azure  | [GitHub Repo](https://github.com/crossplane/provider-azure) | [API Reference](https://doc.crds.dev/github.com/crossplane/provider-azure)  |  [Microsoft Azure Service Principal] | 
+| Provider Jet Azure  |  [GitHub Repo](https://github.com/crossplane-contrib/provider-jet-azure) | [API Reference](https://doc.crds.dev/github.com/crossplane-contrib/provider-jet-azure) |   | 
+| Provider Alibaba |  [GitHub Repo](https://github.com/crossplane/provider-alibaba) | [API Reference](https://doc.crds.dev/github.com/crossplane/provider-alibaba)  |   | 
+|   |   |   |   |
```
hasheddan, line 25: Is this line an intentional separation of providers?

docs/concepts/providers.md
```diff
@@ -7,41 +7,34 @@ indent: true
...
-Custom Resources those CRDs define. Provider packages contain many Custom
-Resource Definitions and their controllers.
+Providers are Crossplane packages that bundle a set of [Managed
+Resources][managed-resources] and their respective controllers to allow
+Crossplane to provision the respective infrastructure resource.
 
 Here is the list of prominent providers:
 
-### AWS Provider
+|   |   |   |   |
```
hasheddan, line 16: Wonder if it would be worth adding headers here for each column -- they are pretty self-explanatory, but the auth mechanism part is somewhat confusing (I do realize these are persisted from the original content though). Any thoughts on how we could make this clearer? Might be worth requiring an `AUTHENTICATION.md` or similar in each repo.

docs/concepts/providers.md
```diff
@@ -7,41 +7,34 @@ indent: true
...
+|---|---|---|---|
+| Provider AWS  |  [GitHub Repo](https://github.com/crossplane/provider-aws) | [API Reference](https://doc.crds.dev/github.com/crossplane/provider-aws)  |  [Amazon Web Services (AWS) IAM User] |
```
hasheddan, line 18: I believe there was desire to standardized on named links instead of inline links -- I don't have a strong preference, but it might be good to be consistent

hasheddan (commented):
Thanks @luebken! Left a few questions below :)

docs/concepts/providers.md
negz: I prefer named links, both in general and because it's consistent with what we're doing today.

docs/concepts/providers.md
```diff
@@ -7,41 +7,34 @@ indent: true
...
-Providers extend Crossplane to enable infrastructure resource provisioning. In
-order to provision a resource, a Custom Resource Definition (CRD) needs to be
-registered in your Kubernetes cluster and its controller should be watching the
-Custom Resources those CRDs define. Provider packages contain many Custom
-Resource Definitions and their controllers.
+Providers are Crossplane packages that bundle a set of [Managed
+Resources][managed-resources] and their respective controllers to allow
+Crossplane to provision the respective infrastructure resource.
 
 Here is the list of prominent providers:
```
negz, line 14: Would it make sense to make this list its own page (somewhere) rather than being under concepts? It feels a little hidden here to me.

docs/concepts/providers.md
luebken: In this case, I find using named links make it actually harder to read and maintain. I'm happy to comply for consistency reasons. Is there any other reason other than consistency?

docs/concepts/providers.md
luebken: Certainly. Actually the list shouldn't be here in the first place it should be derived from the registry. At this point I would like to keep the list here and refactor at a later stage. Main purpose for me was to have the Terrajet providers mentioned in the docs.

docs/concepts/providers.md
negz: I think other than being consistent it's just subjective.

luebken: Fixed with https://github.com/crossplane/crossplane/pull/2816/commits/c5b9bcc33b311c7b6d82f501d3bd0fc3eda07dca

docs/concepts/providers.md
luebken: Yes. It separates the 3(4) big cloud providers from the others.

docs/concepts/providers.md
luebken: I don't see a big issue with the current table TBH. I don't know if an Authentication.md is really needed for every repo. And if it does we should do it in a new repo.

docs/concepts/providers.md
hasheddan: Makes sense 👍🏻 I wonder what our criteria for this distinction would be -- for instance, I could see if being providers under Crossplane org (rather than crossplane-contrib), but that would exclude the new jet providers. The reason why this comes to mind is that many of the ones below the line are more actively maintained than `provider-alibaba`. I don't think anything necessarily needs to change here, but could be some motivation for defining tiers of support more explicitly in the future :)

janwillies:
Just noting here that [provider-styra](https://github.com/crossplane-contrib/provider-styra) has >= 10 GH stars as well

luebken:
@janwillies Thanks. The last time I checked, Styra had less than 10.

docs/concepts/providers.md
hasheddan: @luebken fair assessment -- I would like for us to have some more standardization on how folks discover authentication methods, but I don't think this PR is necessarily the place to accomplish that 👍🏻

hasheddan (approved):
Thanks @luebken!

PR #2819, opened 2022-01-05
Adding distro information to the getting started guide

grantgumina:
In the current Crossplane documentation, Upbound's hosted control plane feature is presented like it's the same thing as a hosted upstream Crossplane deployment. While this is somewhat true, in reality, it serves very different use cases. 

Users reading the getting started guide see hosed control planes, create an Upbound account, and attempt to go through the Crossplane quickstart using a hosted control plane in Upbound. Because of the underlying differences, they quickly run into trouble, and Upbounders find themselves having to explain why a hosted control plane isn't exactly upstream Crossplane managed by Upbound. For example, a common issue is hosted Crossplane uses the `upbound-system` namespaces vs. the `crossplane-system` one mentioned in the Crossplane docs.

The changes in this PR attempt to do the following:

- Clarify that community members have a choice in distribution when using Crossplane
- Route users who would like to use the Upbound Crossplane distribution to the Upbound docs site in order to avoid confusion
- Make the nested tab structure more clear by commenting begin/end sections in the code

Fixes #

I have:

- [x ] Read and followed Crossplane's [contribution process].
- [x ] Run `make reviewable` to ensure this PR is ready for review.

### DIFF
--- a/docs/getting-started/install-configure.md
+++ b/docs/getting-started/install-configure.md
@@ -12,9 +12,9 @@ them today. The first way option to use the upstream distribution of Crossplane
 which is maintained and released by the community. 
 
 Your second option is to use a vendor supported Crossplane distribution. These 
-distributions are fully compatible with upstream Crossplane, but may include 
-additional features or tooling around it that makes it easier to use in 
-production environments. 
+distributions are [certified by the CNCF] to be conformant with upstream 
+Crossplane, but may include additional features or tooling around it that makes 
+it easier to use in production environments. 
 
 <ul class="nav nav-tabs"> 
 <li class="active"><a href="#using-upstream-crossplane" data-toggle="tab">Upstream Crossplane</a></li>
@@ -539,3 +539,4 @@ you can [provision infrastructure].
 [Providers]: ../concepts/providers.md
 [Universal Crossplane]: https://cloud.upbound.io/docs/uxp
 [Get started with Universal Crossplane]: https://cloud.upbound.io/docs/getting-started/
+[certified by the CNCF]: https://github.com/cncf/crossplane-conformance
--- a/docs/getting-started/install-configure.md
+++ b/docs/getting-started/install-configure.md
@@ -17,8 +17,8 @@ may include additional features or tooling around it that makes it easier to use
 in production environments.
 
 <ul class="nav nav-tabs"> 
-<li class="active"><a href="#using-upstream-crossplane" data-toggle="tab">Crossplane</a></li>
-<li><a href="#using-a-downstream-distro" data-toggle="tab">Downstream Distribution</a></li>
+<li class="active"><a href="#using-upstream-crossplane" data-toggle="tab">Crossplane (upstream)</a></li>
+<li><a href="#using-a-downstream-distro" data-toggle="tab">Downstream Distributions</a></li>
 </ul>
 <br>
 <!-- Begin Distro Tabs -->
--- a/docs/getting-started/install-configure.md
+++ b/docs/getting-started/install-configure.md
@@ -5,44 +5,27 @@ weight: 2
 indent: true
 ---
 
-# Choosing Hosted or Self-Hosted Crossplane
+# Choosing Your Crossplane Distribution
 
 Users looking to use Crossplane for the first time have two options available to
-them today. The first way is to use a hosted Crossplane service like [Upbound
-Cloud][Upbound Cloud]. Alternatively, users looking for some more
-flexibility can install Crossplane into their own Kubernetes cluster.
+them today. The first way option to use the upstream distribution of Crossplane 
+which is maintained and released by the community. 
 
-Crossplane will be installed using the regularly published Helm chart. The Helm
-chart contains all the custom resources and controllers needed to deploy and
-configure Crossplane.
+Your second option is to use a vendor supported Crossplane distribution. These 
+distributions are fully compatible with upstream Crossplane, but may include 
+additional features or tooling around it that makes it easier to use in 
+production environments. 
 
-Users choosing the self-hosted option can reference our [Install] and
-[Configure] docs for installing alternate versions and more detailed
-instructions.
-
-<ul class="nav nav-tabs">
-<li class="active"><a href="#using-hosted-crossplane" data-toggle="tab">Hosted Crossplane</a></li>
-<li><a href="#using-self-hosted-crossplane" data-toggle="tab">Self-Hosted Crossplane</a></li>
+<ul class="nav nav-tabs"> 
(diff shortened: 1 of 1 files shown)


docs/getting-started/install-configure.md
```diff
@@ -5,44 +5,27 @@ weight: 2
...
-</div>
-
-<div class="tab-pane fade" id="using-self-hosted-crossplane" markdown="1">
-
-## Start with a Self-Hosted Crossplane
+<!-- Begin Distro Tabs -->
+<div class="tab-content"> 
+<!-- Begin Upstream Tab -->
+<div class="tab-pane fade in active" id="using-upstream-crossplane" markdown="1">
+## Start with Upstream Crossplane
```
hasheddan, line 28:
```suggestion
<div class="tab-pane fade in active" id="using-upstream-crossplane" markdown="1">

## Start with Upstream Crossplane
```

I think not having a blank line between these is causing rendering issues.

docs/getting-started/install-configure.md
```diff
@@ -447,12 +468,41 @@ spec:
...
+</div>
+<!-- End Cloud Provider Tabs -->
+
+</div>
+<!-- End Upstream Tab -->
+
+<!-- Begin Downstream Tab -->
+<div class="tab-pane fade" id="using-a-downstream-distro" markdown="1">
+
+## Start with a Downstream Distribution
```
hasheddan, line 488: Is the intention for the Crossplane getting started guide to work with downstream distributions as well (as they should be compatible)? If so, should this section come before installing configurations?

docs/getting-started/install-configure.md
```diff
@@ -5,44 +5,27 @@ weight: 2
...
-them today. The first way is to use a hosted Crossplane service like [Upbound
-Cloud][Upbound Cloud]. Alternatively, users looking for some more
-flexibility can install Crossplane into their own Kubernetes cluster.
+them today. The first way option to use the upstream distribution of Crossplane 
+which is maintained and released by the community. 
 
-Crossplane will be installed using the regularly published Helm chart. The Helm
-chart contains all the custom resources and controllers needed to deploy and
-configure Crossplane.
+Your second option is to use a vendor supported Crossplane distribution. These
```
hasheddan, line 14:
```suggestion
The second option is to use a vendor supported Crossplane distribution. These 
```

Indifferent to "Your" of "The" but just want to be consistent between this sentence and the previous one.

docs/getting-started/install-configure.md
grantgumina: What I'm proposing here is to route users to the docs site for whatever downstream distribution they want to use vs. put getting started documentation for that distro in the Crossplane documentation site itself. So the path would be "use upstream and continue reading, or choose a distro and view the distro's quickstart". It feels like today there's a lot of jumping back and forth between Upbound docs and Crossplane docs with the way things are structured now.

docs/getting-started/install-configure.md
```diff
@@ -5,44 +5,28 @@ weight: 2
...
 
-# Choosing Hosted or Self-Hosted Crossplane
+# Choosing Your Crossplane Distribution
 
 Users looking to use Crossplane for the first time have two options available to
-them today. The first way is to use a hosted Crossplane service like [Upbound
-Cloud][Upbound Cloud]. Alternatively, users looking for some more
-flexibility can install Crossplane into their own Kubernetes cluster.
+them today. The first way is to use the upstream distribution of Crossplane 
+which is maintained and released by the community.
```
negz, line 12: Nit: Do we want to call upstream a distribution? I think of distributions of adding extra 'stuff' (like UXP adds the ability to work with Upbound) or otherwise being a divergence from 'upstream' personally, but I guess it's not exact.

docs/getting-started/install-configure.md
```diff
@@ -5,44 +5,28 @@ weight: 2
...
+them today. The first way is to use the upstream distribution of Crossplane 
+which is maintained and released by the community. 
 
-Crossplane will be installed using the regularly published Helm chart. The Helm
-chart contains all the custom resources and controllers needed to deploy and
-configure Crossplane.
+The second option is to use a vendor supported Crossplane distribution. These 
+distributions are fully compatible with upstream Crossplane, but may include 
+additional features or tooling around it that makes it easier to use in 
+production environments.
```
negz, line 17: Should we mention something about distributions being conformant? e.g. https://github.com/cncf/crossplane-conformance/pull/16

docs/getting-started/install-configure.md
negz: "Upstream Crossplane" feels weird to me in this context to - a bit like using "upstream" to refer to the place in the stream where you're currently standing. Perhaps "vanilla Crossplane", or just unqualified "Crossplane"?

negz (commented):
Few nits, but I generally agree with the change.

docs/getting-started/install-configure.md
grantgumina: That's a great point. Added here. https://github.com/crossplane/crossplane/pull/2819/commits/270b8608fe7f908e538727f1ee4db605d1594de3

docs/getting-started/install-configure.md
grantgumina: Addressed with this. PTAL https://github.com/crossplane/crossplane/pull/2819/commits/8039a4a6bbab4725624bdc17cb2f5303bbf8b623

docs/getting-started/install-configure.md
hasheddan: @grantgumina that makes a lot of sense 👍🏻 I think conformant "downstream" distributions should all be able to support the Crossplane getting started guide as well, but it is not _technically_ required that this getting started guide "works" (depending on how you define that) in every downstream distribution, so it makes sense to defer the the downstream docs. In practice we could advise any folks who want to add their downstream Crossplane distros to this section to make them work with the Crossplane getting started guide based on Upbound's experience of users getting confused if it does not work 👍🏻

grantgumina: > In practice we could advise any folks who want to add their downstream Crossplane distros to this section to make them work with the Crossplane getting started guide based on Upbound's experience of users getting confused if it does not work

I like that

grantgumina:
@hasheddan @negz seems like everything has been resolved. Are we good to merge?

hasheddan:
@grantgumina commits will need sign off to merge: https://github.com/crossplane/crossplane/pull/2819/checks?check_run_id=4723317285

hasheddan (approved):
Fixed up the corrupted git history -- LGTM, but I'll let @negz also verify that his nits were addressed 👍🏻

PR #2822, opened 2022-01-06
Update go and crossplane-runtime to the latest

muvaf:

I had to make a change in Makefile for it to delete only the first line since they don't add empty line in the beginning of YAMLs anymore.

I have:

`make reviewable`


cluster/crds/pkg.crossplane.io_controllerconfigs.yaml
```diff
@@ -1219,6 +1230,17 @@ spec:
                         description: GMSACredentialSpecName is the name of the GMSA
                           credential spec to use.
                         type: string
+                      hostProcess:
```
hasheddan, line 1233: This is a new field so we'll either need to support translating it or craft our own struct with it omitted.

go.mod
```diff
@@ -1,27 +1,114 @@
...
+	k8s.io/apimachinery v0.23.0
+	k8s.io/client-go v0.23.0
+	k8s.io/code-generator v0.23.0
+	k8s.io/utils v0.0.0-20210930125809-cb0fa318a74b
+	sigs.k8s.io/controller-runtime v0.11.0
+	sigs.k8s.io/controller-tools v0.8.0
+	sigs.k8s.io/yaml v1.3.0
+)
+
+require (
```
negz, line 27: Is this a Go 1.17 thing?

muvaf: Yes, see `Pruned module graphs in go 1.17 modules` section in https://go.dev/doc/go1.17

cluster/crds/pkg.crossplane.io_controllerconfigs.yaml
muvaf: It's part of pod security context and that field seems to be set as is [here](https://github.com/crossplane/crossplane/blob/8c48d4c/internal/controller/pkg/revision/deployment.go#L151). So, I think it's handled?

hasheddan: Ah good call -- thanks!

hasheddan (approved):

PR #2841, opened 2022-01-13
Consume the new k8schain implementation in ggcr

### LINKED ISSUE #2839: Crossplane v1.6.1 not able to handle package deployments
### What happened?

Updating to `1.6.1` makes crossplane not able to handle new package deployments (providers, configurations).
No information in the describe mode, events, and no `pkgrev`
no logs even in debug mode!

Although, it was working in `1.5.1` once updating to `1.6.1` the packages are being marked as unhealthy and could not handle any new packages..

same applies when using public pkg e.g.

```
k crossplane install provider crossplane/provider-kubernetes:v0.1.0
```

The only thing I can see is this e.g. (no pkgrev)

```
(⎈ |dev-tec-myadm-sa:default)☁  ~  k get provider
NAME                  INSTALLED   HEALTHY   PACKAGE                                                    AGE
provider-kubernetes                         some-registry/provider-kubernetes:v0.1.0                   63m
```

Here's more information of our setup:

- We are using a DTR(Docker trusted registry) and we will migrate to Google artifact registry.

- We have a PROXY, so we set proxy entries in crossplane core POD (public communication).

```yaml
env: 
- name: http_proxy
  value: http://something:80
- name: https_proxy
  value: http://something:80
- name: no_proxy
  value: something-different, 127.0.0.1, localhost, metadata, metadata.google.internal, *.googleapis.com

```

- We use `imagePullSecrets` 

Ref: [slack-discussion](https://crossplane.slack.com/archives/CEG3T90A1/p1642001970110600)

### How can we reproduce it?

Update crossplane to v1.6.1 while using a PV for cache.
TBD

Crossplane version: v1.6.1
We are using our own registry!
Environment GKE.
K8s version: v1.21.5
We are using a PV for cache.

hasheddan:

Updates the go-containerregistry and k8schain deps to consume the newly
refactored k8schain, which moves away from consuming the upstream
kubernetes credential providers. The upstream providers cause issues
where long timeouts cause packages to silently fail when our controllers
have context cancelled on timeout.

Fixes #2839 
xref https://github.com/google/go-containerregistry/pull/1234

I have:

In progress. This PR will not be merged until we van verify this implementation works for folks who are pulling packages using IRSA credentials on AWS (for example) _and_ this doesn't break for folks that are pulling through a proxy.

Updates:
- I have run a variety of installations using a default environment (i.e. public registry, using `packagePullSecrets`, not using a proxy).
- @MisterMx has verified this is working correctly with IRSA in their environment.
- @haarchri is testing this out IRSA functionality in their environment (results will be added when testing completed).
- @mcbenjemaa is testing out in their environment with pull through proxy.


mcbenjemaa:
An optional question:

How do you manage transitive dependencies in crossplane?

hasheddan (commented):
@mcbenjemaa the specific answer to your question is that we use the default Go behavior of pinning dependencies, introducing explicit transitive dependency pins when required (such as what we are doing here), but I'm not sure that is the information you are looking for?

haarchri:
it is working in our Environment with IRSA and pull from ECR

hasheddan (commented):
Noting here that with this change we do start to get some superfluous logs again from the ECR cred helper when we are not pulling from ECR explicitly:
```
time="2022-01-14T17:15:14Z" level=error msg="Error parsing the serverURL" error="docker-credential-ecr-login can only be used with Amazon Elastic Container Registry." serverURL=index.docker.io/crossplane/provider-aws
```

hasheddan:
This is likely due to azure creds helper not honoring timeout correctly (or checking whether it is actually an ACR image): https://github.com/chrismellard/docker-credential-acr-env/pull/2

mcbenjemaa:
Results: 

- [X]  `hasheddan/crossplane-k8screds:none` --> contains no cloud creds helpers, we verified this works
- [X]  `hasheddan/crossplane-k8screds:google` --> only google creds helper
- [X]  `hasheddan/crossplane-k8screds:aws` --> only aws creds helper
- [ ]  `hasheddan/crossplane-k8screds:azure` --> only azure creds helper ⚠️

hasheddan (commented):
ACR fix is in -- we'll need to consume https://github.com/google/go-containerregistry/pull/1247

go.mod
```diff
@@ -7,12 +7,13 @@ require (
 	github.com/alecthomas/kong v0.2.17
 	github.com/crossplane/crossplane-runtime v0.15.1-0.20220106140106-428b7c390375
 	github.com/google/go-cmp v0.5.6
-	github.com/google/go-containerregistry v0.7.0
-	github.com/google/go-containerregistry/pkg/authn/k8schain v0.0.0-20210330174036-3259211c1f24
+	github.com/google/go-containerregistry v0.8.1-0.20220120151853-ac864e57b117
+	github.com/google/go-containerregistry/pkg/authn/k8schain v0.0.0-20220120151853-ac864e57b117
```
hasheddan, line 11: I would typically try to stay pinned to a release version here, but I am very familiar with this dependency and feel comfortable using this incremental version to move off the problematic credentials providers.

internal/xpkg/fetch.go
```diff
@@ -29,11 +30,21 @@ import (
...
 )
 
+func init() {
+	// NOTE(hasheddan): we set the logrus package-level logger to discard output
+	// due to the fact that the AWS ECR credential helper uses it to log errors
+	// when parsing registry server URL, which happens any time a package is
+	// pulled from a non-ECR registry.
+	// https://github.com/awslabs/amazon-ecr-credential-helper/issues/308
+	logrus.SetOutput(io.Discard)
+}
```
hasheddan, line 46: I don't love this, but the error log messages being emitted on every package install is pretty much unacceptable IMO. I have already seen some folks who have tested this out get tripped up by seeing them, so I think this is worth it for the short term.

hasheddan (commented):
Updated with the ACR fix and @mcbenjemaa has verified this build works in their environment. This one is ready for final review.

negz:
I wonder if this will fix https://github.com/crossplane/crossplane/issues/2849, too.

go.mod
negz: Is it worth a TODO/note about this?

(Side note - I realised shortly after releasing v1.6.0 that I had forgotten to release and pin crossplane-runtime.)

internal/xpkg/fetch.go
negz: Agreed on both counts. The only alternative I can think of would be disabling this closer to `main()` to avoid the `init()`, but I think I prefer it here closer to the code it affects.

negz (approved):

hasheddan:
@negz perhaps, though @MisterMX mentioned that they had not been having issues prior to this, and also that this fix still worked for them 👍🏻

go.mod
hasheddan: I'll add one 👍🏻

PR #2844, opened 2022-01-13
Do not build unpublished architectures

AaronME:
Signed-off-by: Aaron Eaton <aaron@upbound.io>

CI is building all supported architectures, though only publishing linux docker images. This PR removes unpublished architectures from CI.

I have:

Running `make build` locally confirmed that the local architecture always builds.
Running `make build.all` locally confirmed that only published architectures are built for CI.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@
 PROJECT_NAME := crossplane
 PROJECT_REPO := github.com/crossplane/$(PROJECT_NAME)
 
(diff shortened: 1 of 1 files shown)


hasheddan (changes requested):
@AaronME we build these architectures for the CLI and don’t currently have a great way to disable also building Crossplane core for them, so we can’t just remove them at this time.

negz:
It sounds like we can't disable these and it's been a few months so I'm going to close the PR.

PR #2852, opened 2022-01-21
[release-1.6] Consume the new k8schain implementation in ggcr

hasheddan:

Updates the go-containerregistry and k8schain deps to consume the newly
refactored k8schain, which moves away from consuming the upstream
kubernetes credential providers. The upstream providers cause issues
where long timeouts cause packages to silently fail when our controllers
have context cancelled on timeout.

Backport of #2841 that also brings along https://github.com/crossplane/crossplane/pull/2822 and https://github.com/crossplane/crossplane/pull/2814 for the ride

I have:

See steps in #2841, which I have reproduced with the build from this backport.


go.mod
```diff
@@ -1,29 +1,133 @@
...
+	github.com/sirupsen/logrus v1.8.1
+	github.com/spf13/afero v1.8.0
+	golang.org/x/tools v0.1.8
+	k8s.io/api v0.23.0
+	k8s.io/apiextensions-apiserver v0.23.0
+	k8s.io/apimachinery v0.23.0
+	k8s.io/client-go v0.23.0
+	k8s.io/code-generator v0.23.0
+	k8s.io/utils v0.0.0-20210930125809-cb0fa318a74b
+	sigs.k8s.io/controller-runtime v0.11.0
```
negz, line 27: Hrm, do we really want to go from 0.9.x -> 0.11.x of this fairly impactful dependency in a patch release? 🤔 Hypothetically there are similar concerns for crossplane-runtime, but I'm fairly sure not much has changed in that bump.

hasheddan: @negz fair point -- I included it as it was grouped with the `1.17` change, but I believe we could drop https://github.com/crossplane/crossplane/pull/2852/commits/5bf6faad9da505bbcd2e9ed02c4b3ff0e1b241b5 and it wouldn't complicate much.

hasheddan: @negz it does introduce some conflicts -- with them resolved it looks like this: https://github.com/hasheddan/crossplane/tree/rel-1.6-k8schain-rebase

Definitely agree that bumping two versions of `controller-runtime` feels out of place for a patch. I did look through the releases and the changes mentioned do not look overly concerning to me (https://github.com/kubernetes-sigs/controller-runtime/releases -- the fairly significant refactor of `Manager` does give me pause). That being said, I think I might opt to drop the backport strategy altogether so we don't have to pick up the `1.17` changes and instead just make the same changes from https://github.com/crossplane/crossplane/pull/2841 on `release-1.6`. What do you think?

hasheddan: @negz this turned out to be more of a challenge than expected as the `go-containerregistry` updates bump [k8s deps](https://github.com/google/go-containerregistry/blob/ac864e57b117d74a16269906dcbbd8db366eaadd/pkg/authn/kubernetes/go.mod#L9), which in turn break logging in our pinned version of `controller-runtime` (due to `client-go` changes). We could try to use a `replace` and find a common version that works, but that honestly feels more brittle to me than just moving ahead with the bumps.

negz: @hasheddan If you feel comfortable I could be convinced to learn to stop worrying and love the bump. ;) Maybe we can kick the tyres a bit before we release it, just to make sure there's nothing weird with the manager refactor? The composition controllers in particular use some fairly obscure features of the `Manager` code.

hasheddan: @negz I am feeling pretty good about it (well, relative to the alternatives). The `controller-runtime` release has had about a month of bake time and I have at least given a cursory review of most of the major changes. Definitely agree that we should do some further testing before release. I specifically want to test some of the heavy load scenarios and will likely enlist some community members to try to stress it in their org's dev environments. I think a release early next week would be plenty of time for me to feel confident in the stability of these bumps.

negz (approved):

Issue #2524, opened 2021-08-24
Proposal: Custom Compositions


negz:
### What problem are you facing?

Composition is the Crossplane feature that lets teams define their own opinionated APIs (i.e. Kubernetes CRs). We call these APIs "Composite Resources", or XRs for short. When a user creates an XR, Crossplane uses another kind of resource - a `Composition` - to determine what to do. For example a `Composition` might tell Crossplane that when a user creates an `AcmeCoDatabase` XR it should respond by creating a GCP `CloudSQLInstance` and a `ServiceAccount`. From our [terminology documentation](https://crossplane.io/docs/v1.3/concepts/terminology.html#composition):

> Folks accustomed to Terraform might think of a Composition as a Terraform module; the HCL code that describes how to take input variables and use them to create resources in some cloud API. Folks accustomed to Helm might think of a Composition as a Helm chart’s templates; the moustache templated YAML files that describe how to take Helm chart values and render Kubernetes resources.

A Crossplane `Composition` consists of an array of one ore more 'base' resources. Each of these resources can be 'patched' with values derived from the XR. The functionality enabled by a `Composition` is intentionally limited - for example there is no support for conditionals (e.g. only create this resource if the following conditions are met) or iteration (e.g. create N of the following resource, where N is derived from an XR field). These limits:

* Attempt to avoid [the pitfalls of configuration domain-specific languages](https://github.com/kubernetes/community/blob/8956bcd54dc6f99bcb681c79a7e5399289e15630/contributors/design-proposals/architecture/declarative-application-management.md#pitfalls-of-configuration-domain-specific-languages-dsls).
* Allow us to express "composition logic" as a custom resource that may be entirely stored within and validated at create time by the Kubernetes API server.
* Allow the `Composition` type to be relatively simple, avoiding it becoming a programming language expressed as YAML.

Limiting the functionality of our `Composition` type allows us to reduce Crossplane's learning curve and implementation complexity. At the same time it limits Crossplane's ability to address certain use cases. It's not uncommon for Crossplane users to resort to 'composing' Crossplane managed resources using more featureful tools such as Helm or jsonnet when they hit the limitations imposed by the `Composition` type. Furthermore, some Crossplane users simply prefer to express their intent in a language (or with a tool) they already know and feel comfortable with, such as Helm, jsonnet, Python, HCL, Typescript, etc.

### How could Crossplane help solve your problem?

[...]

PR #2886, opened 2022-02-09
Add a design document for 'Composition Functions'

negz:

Fixes https://github.com/crossplane/crossplane/issues/3001
Fixes https://github.com/crossplane/crossplane/issues/2959
Fixes https://github.com/crossplane/crossplane/issues/2958

This commit adds a design document for 'Composition Functions', which
we've previously referred to as 'Custom Compositions'. See https://github.com/crossplane/crossplane/issues/2524 for the
full context.

I have:

@sergenyalcin and I have built a few proof of concepts, notably https://github.com/crossplane/crossplane/issues/3001
and https://github.com/crossplane/crossplane/issues/2959 to inform this design.

### DIFF
--- a/design/design-doc-composition-functions.md
+++ b/design/design-doc-composition-functions.md
@@ -237,7 +237,8 @@ spec:
       # Containers are run by an external process listening at the supplied
       # endpoint. Specifying an endpoint is optional; the endpoint defaults to
       # the below value.
-      endpoint: unix:///@crossplane/fn/default.sock
+      runner:
+        endpoint: unix:///@crossplane/fn/default.sock
     # An x-kubernetes-embedded-resource RawExtension (i.e. an unschemafied
     # Kubernetes resource). Passed to the function as the config block of its
     # FunctionIO.
@@ -417,7 +418,7 @@ In order to provide flexibility and choice of tradeoffs in running containers
 (e.g. speed, scalability, security) this document proposes Crossplane defer
 containerized functions to an external runner. Communication with the runner
 would occur via a gRPC API, with the runner expected to be listening at the
-`endpoint` specified via the function's `container` configuration block. This
+`endpoint` specified via the function's `runner` configuration block. This
 endpoint would default to `unix:///@crossplane/fn/default.sock` - an abstract
 [Unix domain socket][unix-domain-sockets].
 
--- a/design/design-doc-composition-functions.md
+++ b/design/design-doc-composition-functions.md
@@ -164,7 +164,9 @@ support for the feature to Crossplane.
 This document proposes that a new `functions` array be added to the existing
 `Composition` type. This array of functions would be called either instead of or
 in addition to the existing `resources` array in order to determine how an XR
-should be composed.
+should be composed. The array of functions acts as a pipeline; the output of
+each function is passed as the input to the next. The output of the final
+function tells Crossplane what must be done to reconcile the XR.
 
 ```yaml
 apiVersion: apiextensions.crossplane.io/v2alpha1
@@ -233,8 +235,8 @@ spec:
       # How long the function may run before it's killed. Defaults to 10s.
       timeout: 30s
     # An x-kubernetes-embedded-resource RawExtension (i.e. an unschemafied
-    # Kubernetes resource). Passed to the function as the functionConfig block
-    # of its ResourceList.
+    # Kubernetes resource). Passed to the function as the config block of its
+    # FunctionIO.
     config:
       apiVersion: database.example.org/v1alpha1
       kind: Config
@@ -246,53 +248,53 @@ spec:
 
 ### Function API
 
-This document proposes that each function uses a `ResourceList` type derived
-from the [KRM function specification][krm-fn-spec] as its input and output.
+This document proposes that each function uses a `FunctionIO` type as its input
+and output. This method is inspired by [KRM functions][krm-fn-spec].
 
 ```yaml
-apiVersion: fn.apiextensions.crossplane.io/v1
-kind: ResourceList
-functionConfig:
+apiVersion: apiextensions.crossplane.io/v1alpha1
+kind: FunctionIO
+config:
   apiVersion: database.example.org/v1alpha1
   kind: Config
   metadata:
     name: cloudsql
   spec:
     version: POSTGRES_9_6
-items:
-- apiVersion: database.example.org/v1alpha1
+composite:
+  apiVersion: database.example.org/v1alpha1
   kind: XPostgreSQLInstance
   metadata:
     name: my-db
-    annotations:
-      fn.apiextensions.crossplane.io/type: "CompositeResource"
   spec:
     parameters:
       storageGB: 20
     compositionSelector:
       matchLabels:
         provider: gcp
-    writeConnectionSecretToRef:
-      name: db-conn
 ```
 
-Like a KRM function, a Composition function uses a `ResourceList` that consists
-of the following top-level fields:
+A Composition Function uses a `FunctionIO` that consists of the following
+top-level fields:
 
 * The `apiVersion` and `kind` (required).
-* A `functionConfig` (optional). This is a [Kubernetes resource][rawextension]
+* A `config` object (optional). This is a [Kubernetes resource][rawextension]
   with an arbitrary schema that may be used to provide additional configuration
   to a function. For example a `render-helm-chart` function might use its
-  `functionConfig` to specify which Helm chart to render.
-* An `items` array (required). The XR and zero or more composed resources.
-* A `results` object (optional). Used to communicate information about the
-  result of a function, including warnings and errors.
+  `config` to specify which Helm chart to render.
+* A `composite` object (required). This is the XR being reconciled.
+* A `resources` array (optional). Zero or more composed resources.
+* A `results` array (optional). Used to communicate information about the result
+  of a function, including warnings and errors.
 
-Each function in the array would be executed as a pipeline, with each function:
+Each function:
 
-1. Taking the XR and zero or more composed resources as input.
-1. Producing the optionally mutated XR, optionally mutated composed resources,
-   and newly rendered composed resources as output.
+1. Takes its config, XR, and zero or more composed resources as input.
+1. Optionally mutates the XR's metadata, spec, and/or status.
+1. Optionally mutates the array of composed resources. This may involve adding
+   or deleting composed resources, as well as mutating the metadata and spec of
+   any entry in the array. Updates to the status of composed resources are not
+   allowed and will be silently discarded.
 
 This allows the output of one function to be the input to the next. The first
 function in the pipeline would be supplied with either:
@@ -301,75 +303,101 @@ function in the pipeline would be supplied with either:
    did not include a `resources` array.
 1. The XR and one or more resources. This would be the case if either:
    1. The Composition included both a `resources` array and a `functions` array.
-      In this case the `resources` array is computed first then passed to
-      `functions`.
+      In this case the `resources` array is processed first, then passed to
+      `functions` for further processing.
    1. This was not a newly created XR, but instead an XR that had already
       created its composed resources and was now being updated.
 
 In the case of `Container` functions this input and output would correspond to
 stdin and stdout. This may differ for future function implementations; for
-example a hypothetical webhook function implementation might take the input as
-an HTTP PUT request body and return the output as a response body. Crossplane
-would be responsible for reading stdout from the final function and applying its
-changes to the relevant XR and subsequent composed resources.
+example a webhook function implementation might take the input as an HTTP PUT
+request body and return the output as a response body. Crossplane would be
+responsible for reading stdout from the final function and applying its changes
+to the relevant XR and subsequent composed resources.
 
 An example function that responded by asking Crossplane to create (compose) a
-`CloudSQLInstance` would do so by returning the following `ResourceList`:
+`CloudSQLInstance` would do so by returning the following `FunctionIO`:
 
 ```yaml
-apiVersion: fn.apiextensions.crossplane.io/v1alpha1
-kind: ResourceList
-functionConfig:
+apiVersion: apiextensions.crossplane.io/v1alpha1
+kind: FunctionIO
+config:
   apiVersion: database.example.org/v1alpha1
   kind: Config
   metadata:
     name: cloudsql
   spec:
     version: POSTGRES_9_6
-items:
-- apiVersion: database.example.org/v1alpha1
+composite:
+  apiVersion: database.example.org/v1alpha1
   kind: XPostgreSQLInstance
   metadata:
     name: my-db
-    annotations:
-      fn.apiextensions.crossplane.io/type: "CompositeResource"
   spec:
     parameters:
       storageGB: 20
     compositionSelector:
       matchLabels:
         provider: gcp
-    writeConnectionSecretToRef:
-      name: db-conn
-- apiVersion: database.gcp.crossplane.io/v1beta1
-  kind: CloudSQLInstance
-  metadata:
-    name: cloudsqlpostgresql
-    annotations:
-      fn.apiextensions.crossplane.io/type: "ComposedResource"
-      fn.apiextensions.crossplane.io/name: cloudsqlinstance
-  spec:
-    forProvider:
-      databaseVersion: POSTGRES_9_6
-      region: us-central1
-      settings:
-        tier: db-custom-1-3840
-        dataDiskType: PD_SSD
-        dataDiskSizeGb: 20
-    writeConnectionSecretToRef:
-      namespace: crossplane-system
-      name: cloudsqlpostgresql-conn
-results:
-- severity: Error
-  message: "Could not render Database.postgresql.crossplane.io/v1alpha1`
+resources:
+- name: cloudsqlinstance
+  resource:
+    apiVersion: database.gcp.crossplane.io/v1beta1
+    kind: CloudSQLInstance
+    spec:
+      forProvider:
+        databaseVersion: POSTGRES_9_6
+        region: us-central1
+        settings:
+          tier: db-custom-1-3840
+          dataDiskType: PD_SSD
+          dataDiskSizeGb: 20
+      writeConnectionSecretToRef:
+        namespace: crossplane-system
+        name: cloudsqlpostgresql-conn
+  connectionDetails:
+  - name: hostname
+    fromConnectionSecretKey: hostname
+  readinessChecks:
+  - type: None
 ```
 
-Note the Crossplane specific `fn.apiextensions.crossplane.io/` annotations:
+Note that a the `resources` array of the `FunctionIO` type is very similar to
+the `resources` array of the `Composition` type. In comparison:
 
-1. `type` distinguishes an XR from a composed resource. XRs are the first
-   element in the `items` array by convention.
-1. `name` associates a composed resource entry in the `items` array with its
-   corresponding entry in the `resources` array (if any).
+* `name` works the same across both types, but is required by `FunctionIO`.
+* `connectionDetails` and `readinessChecks` work the same across both types.
+* `FunctionIO` does not support `base` and `patches`. Instead a function should
+  configure the `resource` field accordingly.
+
+A function that could not complete successfully could do so by returning the
+following `FunctionIO`:
+
+```yaml
+apiVersion: apiextensions.crossplane.io/v1alpha1
+kind: FunctionIO
+config:
+  apiVersion: database.example.org/v1alpha1
+  kind: Config
+  metadata:
+    name: cloudsql
+  spec:
+    version: POSTGRES_9_6
+composite:
+  apiVersion: database.example.org/v1alpha1
+  kind: XPostgreSQLInstance
+  metadata:
+    name: my-db
+  spec:
+    parameters:
+      storageGB: 20
+    compositionSelector:
+      matchLabels:
+        provider: gcp
+results:
+- severity: Error
+  message: "Could not render Database.postgresql.crossplane.io/v1beta1`
+```
 
 While KRM-function-like this API is not KRM function compatible. See the
 [Alternatives Considered](#alternatives-considered) section for details on why.
@@ -553,7 +581,7 @@ present time but likely will in future.
 
 Crossplane could invoke functions by calling a webhook rather than running an
 OCI container. In this model function input and output would still take the form
-of a `ResourceList`, but would be HTTP request and response bodies rather than a
+of a `FunctionIO`, but would be HTTP request and response bodies rather than a
 container's stdin and stdout.
 
 The primary detractor of this approach is the burden it puts on function authors
@@ -610,13 +638,6 @@ ideal fit. This is because:
 1. Crossplane needs additional metadata to distinguish which resource in the
    `ResourceList` is the composite resource and which are the composed
    resources.
-
-[Work is ongoing][krm-fn-runtimes] within the `kpt` project as well as
-Kubernetes sig-cli to make the KRM function specification 'server-side'
-compatible; it may be possible to achieve compatibility in future. Compatibility
-with the existing [catalog][krm-fn-catalog] of KRM functions would be nice to
-have for functions such as `set-annotations`, `gatekeeper`, and
-`render-helm-chart`.
 
 ### gVisor
 
--- a/design/design-doc-composition-functions.md
+++ b/design/design-doc-composition-functions.md
@@ -155,7 +155,7 @@ this document could serve as a test bed for new features in the contemporary
 
 The user experience around authoring and maintaining Composition Functions is
 out of scope for this proposal, which focuses only on adding foundational
-support for the feature to Crossplane.
+support for the feature to Crossplane. 
 
 ## Proposal
 
@@ -184,7 +184,7 @@ spec:
 
 Under this proposal each function is the entrypoint of an OCI image, though the
 API is designed to support different function implementations (such as webhooks)
-in th future. The updated API would affect only the `Composition` type - no
+in the future. The updated API would affect only the `Composition` type - no
 changes would be required to the schema of `CompositeResourceDefinitions`, XRs,
 etc.
 
@@ -232,9 +232,9 @@ spec:
       network: Accessible
       # How long the function may run before it's killed. Defaults to 10s.
       timeout: 30s
-    # An x-kubernetes-embedded-resource RawExtension (i.e. an unschemafied)
-    # Kubernetes resource. Passed to the function as the functionConfig block of
-    # its ResourceList.
+    # An x-kubernetes-embedded-resource RawExtension (i.e. an unschemafied
+    # Kubernetes resource). Passed to the function as the functionConfig block
+    # of its ResourceList.
     config:
       apiVersion: database.example.org/v1alpha1
       kind: Config
--- a/design/design-doc-composition-functions.md
+++ b/design/design-doc-composition-functions.md
@@ -163,6 +163,22 @@ spec:
     container:
       # The OCI image to pull and run.
       image: xkpg.io/my-cool-function:0.1.0
+      # Whether to pull the function Never, Always, or IfNotPresent.
+      imagePullPolicy: IfNotPresent
+      # Secrets used to pull from a private registry.
+      imagePullSecrets:
+      - namespace: crossplane-system
+        name: my-xpkg-io-creds
+      # Note that only resource limits are supported - not requests.
+      # The function will be run with the specified resource limits.
+      resources:
+        limits:
+          memory: 64Mi
+          cpu: 250m
+      # Defaults to 'Isolated' - i.e an isolated network namespace.
+      network: Accessible
+      # How long the function may run before it's killed. Defaults to 10s.
+      timeout: 30s
     # An x-kubernetes-embedded-resource RawExtension (i.e. an unschemafied)
     # Kubernetes resource. Passed to the function as the functionConfig block of
     # its ResourceList.
--- a/design/design-doc-composition-functions.md
+++ b/design/design-doc-composition-functions.md
@@ -425,7 +425,7 @@ endpoint would default to `unix:///@crossplane/fn/default.sock` - an abstract
 Communication between Crossplane and a containerized function runner would use
 the following API:
 
-```proto3
+```protobuf
 syntax = "proto3";
 
 // This service defines the APIs for a containerized function runner.
--- a/design/design-doc-composition-functions.md
+++ b/design/design-doc-composition-functions.md
@@ -153,6 +153,10 @@ While not an explicit goal, it would also be ideal if the solution put forth by
 this document could serve as a test bed for new features in the contemporary
 'resources array' based form of Composition.
 
+The user experience around authoring and maintaining Composition Functions is
+out of scope for this proposal, which focuses only on adding foundational
+support for the feature to Crossplane.
+
 ## Proposal
 
 ### Overview
--- a/design/design-doc-composition-functions.md
+++ b/design/design-doc-composition-functions.md
@@ -234,6 +234,10 @@ spec:
       network: Accessible
       # How long the function may run before it's killed. Defaults to 10s.
       timeout: 30s
+      # Containers are run by an external process listening at the supplied
+      # endpoint. Specifying an endpoint is optional; the endpoint defaults to
+      # the below value.
+      endpoint: unix:///@crossplane/fn/default.sock
     # An x-kubernetes-embedded-resource RawExtension (i.e. an unschemafied
     # Kubernetes resource). Passed to the function as the config block of its
     # FunctionIO.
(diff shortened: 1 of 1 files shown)


negz:
We're currently prototyping two approaches to actually implement this.

1. https://github.com/crossplane/crossplane/issues/3001
2. https://github.com/crossplane/crossplane/issues/2959

The second approach also involves https://github.com/crossplane/crossplane/issues/2958. Each approach has tradeoffs and we may pick either one or allow a choice.

The first approach would likely result in

* Pro: Faster reconcile times - no need to schedule and run a pod.
* Pro: No need for a wrapper process - the controller could pass stdin/stdout etc between functions.
* Pro: Likely a better integration with web hook based functions (a function of the two above pros).
* Con: Noisy neighbors (multiple function processes in the same 'runner' container)
* Con: No isolation (network/process, etc wise) between function processes in the same container.
* Con: "Reinventing the wheel" (pulling, caching function images)

The second approach is roughly the inverse:

* Pro: Leverages existing Kubernetes infra (Pods) - caching, reruns, isolation, etc.
* Con: No native ordering of containers within a Pod (needs a wrapper process ala Argo Emissary).
* Con: Likely scheduling issues delays - need to create, schedule, and run a pod for each reconcile loop iteration.

negz:
This is now ready for review. I'd like to get approvals from @sergenyalcin and @turkenh - perhaps @sergenyalcin could do a first pass since he has the most context on this design.

design/design-doc-composition-functions.md
```diff
@@ -0,0 +1,439 @@
...
+  name: example
+spec:
+  compositeTypeRef:
+    apiVersion: database.example.org/v1alpha1
+    kind: XPostgreSQLInstance
+  functions:
+  - name: my-cool-function
+    type: Container
+    container:
+      image: xkpg.io/my-cool-function:0.1.0
```
turkenh, line 130: Considering the input will be a list of XRs and Composed resources, would it be possible to have some generic/reusable function images? Unless I am missing something, there will be a tight coupling between function image(s) and XRD / Composition spec in practice (for most of the cases?) which could complicate packaging (configurations and functions). Any possibility to package them together as a single image?

turkenh (commented):
Excited to see this coming 🔥

design/design-doc-composition-functions.md
```diff
@@ -0,0 +1,439 @@
...
+```
+
+Under this proposal each function is the entrypoint of an OCI image, though the
+API is designed to support different function implementations (such as webhooks)
+in future. The updated API would affect only the `Composition` type - no changes
```
ron1, line 135: in the future. The updated API would affect only the `Composition` type - no changes

ron1, line 135:
```suggestion
in the future. The updated API would affect only the `Composition` type - no changes
```

design/design-doc-composition-functions.md
```diff
@@ -0,0 +1,439 @@
...
+spec:
+  compositeTypeRef:
+    apiVersion: database.example.org/v1alpha1
+    kind: XPostgreSQLInstance
+  functions:
+  - name: my-cool-function
+    type: Container
+    container:
+      image: xkpg.io/my-cool-function:0.1.0
+```
```
ron1, line 131: The functionConfig in KRM seems to play a key role in making functions reusable. In particular it ensures function parameters are specified in a consistent way. Is there a reason you decided to capture function meta info using an array instead of adopting the KRM declarative ConfigMap-based style of functionConfig specification described [here](https://googlecontainertools.github.io/kpt/guides/consumer/function/#declarative-run)?

ron1, line 131: Is it possible for a function container arg to be patched from a field in the XR?

design/design-doc-composition-functions.md
```diff
@@ -0,0 +1,439 @@
...
+      case the `resources` array is computed first then passed to `functions`.
+   1. This was not a newly created XR, but instead an XR that already had
+      already created its composed resources and was now being updated.
+
+In the case of `Container` functions this input and output would correspond to
+stdin and stdout. This may differ for future function implementations; for
+example a hypothetical webhook function implementation might take the input as
+an HTTP PUT request body and return the output as a response body. Crossplane
+would be responsible for reading stdout from the final function and applying its
+changes to the relevant XR and subsequent composed resources.
```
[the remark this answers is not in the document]
negz: I'm not quite sure what you're asking.

Today we support deleting composed resources as long as all entries in the `resources` array have a `name`. That is, if you remove a named entry from the `resources` array of a `Composition` Crossplane will notice on the next reconcile and delete the corresponding composed resource. This should work the same with functions - if a composed resource that previously existed is no longer present in the final `ResourceList` we'll delete it.


https://github.com/crossplane/crossplane/blob/4cee13fb/internal/controller/apiextensions/composite/composed.go#L212

negz (commented):
Thanks for the review folks!

design/design-doc-composition-functions.md
ron1: Introducing support for KRM-like function https://catalog.kpt.dev/render-helm-chart/v0.1/ might be a good use case here. It requires a `functionConfig` to specify parameters and requires optional network-enabling to support access to remote charts or remote valuesFiles.

Are you thinking about using the Kpt declarative ConfigMap approach described [here](https://googlecontainertools.github.io/kpt/guides/consumer/function/#declarative-run)? How would this work with the functions array that is currently proposed?

negz: > A POC for Kpt-like function [render-helm-chart](https://catalog.kpt.dev/render-helm-chart/v0.1/) might be a good use case here.

@sergenyalcin is actually working on this - we should probably have an issue tracking that. I do imagine our take on this function would use the XR as the input to `helm template` though - i.e. convert the XR to a `values.yaml`.

> WDYT about using the Kpt declarative ConfigMap approach for functionConfigs described [here](https://googlecontainertools.github.io/kpt/guides/consumer/function/#declarative-run)? I'm not sure how this would this work with the functions array that is currently proposed? Would you consider including the ConfigMaps in the ResourceList and removing the functions array altogether?

If I'm following what kit is doing correctly I don't think it makes sense for us. It seems to be a way for a kpt to detect any type (not just ConfigMaps) within a directory that is annotated with a particular function and then run that function with the type as its functionConfig.

> Do you have any initial thoughts about achieving interoperability between Kpt functions and Crossplane functions?

I touched on this in the design doc. It seems like it would be nice to have but not particularly important. It may be more feasible once the KRM functions spec is updated to consider server-side implementations per   https://github.com/GoogleContainerTools/kpt/issues/2567.

negz:
I noticed while testing https://github.com/crossplane/crossplane/pull/3062 that the default seccomp profile in Docker and containerd reject the `unshare` syscall that we need to make rootless containers. Kubernetes runs with the Unconstrained seccomp profile by default so this shouldn't be a problem for most folks, but that may change with https://github.com/kubernetes/kubernetes/pull/101943.

I'm not too worried about this - I believe that even if `RuntimeDefault` becomes the default seccomp profile in Kubernetes we'd only be asking folks to give us two syscalls that aren't included by default - `unshare` and optionally `mount`.

design/design-doc-composition-functions.md
ron1: How are you planning to capture the `functionConfig`s? A Kpt-like solution might be a list of ConfigMaps, each with an annotated function definition as described in the link above. To make the functions reusable, it would seem useful for the ConfigMaps to be patchable from the XR.

design/design-doc-composition-functions.md
```diff
@@ -0,0 +1,439 @@
...
+The disadvantages of running the pipeline inside the Crossplane container are
+scale and reinvention of the wheel. The resources available to the Crossplane
+container will bound how many functions it can run at any one time, and
+Crossplane will need to handle features that the Kubelet already offers such as
+pull secrets, caching etc. In future Crossplane may support other ways to invoke
+the container pipeline - refer to the [Alternatives
+Considered](#alternatives-considered) section for more information.
+
+[Rootless containers][rootless] appear to be the most promising way to run
+functions as containers inside the Crossplane container:
```
negz, line 270: I did a bit of a survey this weekend to make sure rootless containers (as we use them) will work. I mostly tested by running `unshare -rUm` inside an unprivileged container to see whether I could get a new user and mount namespace. I found:

* They work fine on EKS out of the box.
* They work on regular GKE, but you need to disable AppArmor for the relevant container by adding the `container.apparmor.security.beta.kubernetes.io/${CONTAINER_NAME}=unconfined` annotation.
* They don't work on GKE with gVisor, which doesn't fully support the `unshare` syscall yet.
* They don't work on GKE Autopilot, which requires gVisor.

It is worth noting that the default Docker and containerd seccomp profiles will reject the `unshare` syscall too. Kubernetes (including GKE and EKS) runs without seccomp by default today, but that may change in future. At that point we'd have to configure the Crossplane pod to either not use seccomp or to use a slightly less restrictive seccomp profile (i.e. one that allowed unshare).

My hunch is that we'll likely want to add a `type: Chroot` function that uses a simple chroot to run a function in future. This would work in more places (i.e. more restrictive container runtimes) but would give us less chance to isolate the functions (i.e. no network isolation) and always require making a fresh copy of the 'container' filesystem.

design/design-doc-composition-functions.md
sergenyalcin: > @sergenyalcin is actually working on this - we should probably have an issue tracking that. I do imagine our take on this function would use the XR as the input to helm template though - i.e. convert the XR to a values.yaml.

In [helm-based XRM reference function issue], I commented my some [observations and ideas]. They can be related about this point.

[helm-based XRM reference function issue]: https://github.com/crossplane/crossplane/issues/3067
[observations and ideas]: https://github.com/crossplane/crossplane/issues/3067#issuecomment-1114212053

design/design-doc-composition-functions.md
```diff
@@ -0,0 +1,439 @@
...
+```yaml
+apiVersion: fn.apiextensions.crossplane.io/v1alpha1
+kind: ResourceList
+items:
+- apiVersion: database.example.org/v1alpha1
+  kind: XPostgreSQLInstance
+  metadata:
+    name: my-db
+    annotations:
+      fn.apiextensions.crossplane.io/type: "CompositeResource"
```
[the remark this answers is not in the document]
negz: I'm going to resolve this thread - feel free to reopen if needed.

design/design-doc-composition-functions.md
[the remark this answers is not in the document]
negz: I'm hoping this answered your question - I'm going to resolve this thread. Please reopen if I misunderstood.

design/design-doc-composition-functions.md
negz: Just pushed a commit that touches on this in the doc.

design/design-doc-composition-functions.md
negz: > I could be convinced to add a section to this document or follow up separately.

I just added a brief section on this. I went back and forth, but in the end I'm leaning toward not supporting (or at least not optimizing for) functions being packaged together with a configuration. My thinking is:

* Functions will be pulled and cached separately from configurations, so there's not much benefit at the 'consumer' (i.e. Crossplane) end to packaging them in the same OCI image.
* It's tough to avoid collisions when multiple 'package-ish' things are in the same OCI image (e.g. multiple `/package.yaml` files). This is especially true if we want to package more than one function with a configuration.
* Any configuration that includes a Composition implicitly declares its dependency on containerized functions by simply including references to them in its `functions` array.

I think then the only real convenience of packaging functions with a configuration then is at the producer (i.e. packager) end, in that they might be able to run one build and one push command to push both a configuration and one or more functions.

design/design-doc-composition-functions.md
```diff
@@ -0,0 +1,439 @@
...
+
+Notably the functions would not be responsible for interacting with the API
+server to created, update, or delete composed resources. Instead they instruct
+Crossplane which resources should be created, updated, or deleted.
+
+Under the proposed design functions could also be used for purposes besides
+rendering composed resources, for example validating the results of the
+`resources` array or earlier functions in the `functions` array. Furthermore, a
+function could also be used to implement 'side effects' such as triggering a
+replication or backup.
```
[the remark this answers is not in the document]
negz: I've pushed a commit that details the usage of the `functionConfig` block. This can be used to pass configuration/environment details to a function.

That said, `functionConfig` as its currently proposed is not good for passing in _credentials_ since it currently requires the config to be specified inline as part of a `Composition`. I have some thoughts on this but I'm going to resolve this thread and move the conversation closer to the new `config` block I've suggested.

design/design-doc-composition-functions.md
negz: @ron1 I've pushed a commit that details how I imagine `functionConfig` would work with Composition functions. @turkenh had some related thoughts so I'm going to resolve this thread (and his related thread) and start a new one closer to the relevant example.

design/design-doc-composition-functions.md
```diff
@@ -0,0 +1,583 @@
...
+    # An x-kubernetes-embedded-resource RawExtension (i.e. an unschemafied)
+    # Kubernetes resource. Passed to the function as the functionConfig block of
+    # its ResourceList.
+    config:
+      apiVersion: database.example.org/v1alpha1
+      kind: Config
+      metadata:
+        name: cloudsql
+      spec:
+        version: POSTGRES_9_6
```
negz, line 175: @turkenh @ron1 this is how I imagine per-function configuration would work. The `config` block specified here would be passed as the `functionConfig` to the relevant function. Therefore the input to each function is:

1. The `items` array - i.e. the XR and any composed resources.
2. The `functionConfig`, as specified in the `Composition`.

Taking the Helm chart based "reference function" @sergenyalcin is working on in https://github.com/crossplane/crossplane/issues/3067 for example, I imagine:

1. Which Helm chart to use would be specified in the `functionConfig`.
2. The inputs to the Helm chart (i.e. the `values.yaml`) would be the `ResourceList`, or at least the XR from said list.

Note that this assumes we're using a Helm chart expressly written to do the job of deriving composed resources from an XR; not a generic Helm chart. We already have `provider-helm` for rendering generic Helm charts.

One gap in this design is passing sensitive details to functions. As @turkenh pointed out a function might require credentials to operate, for example a username and password for a Helm repository. In this design those sensitive credentials would have to be provided inline in the Composition.

We could fairly easily allow config to be loaded from a Kubernetes Secret, for example:

```yaml
apiVersion: apiextensions.crossplane.io/v2alpha1
 kind: Composition
 metadata:
   name: example
 spec:
   compositeTypeRef:
     kind: XPostgreSQLInstance
   functions:
   - name: my-cool-function
     type: Container
     container:
       # The OCI image to pull and run.
       image: xkpg.io/my-cool-function:0.1.0
     configSource: Secret
     # Mutually exclusive with the config block.
     configSecretRef:
       namespace: crossplane-system
       name: my-cool-function-config
```

In this example the entire referenced `Secret` would be passed as the `functionConfig`.

I'm not sure just supporting Kubernetes secrets is really enough though. We've already begun adding support to Crossplane for other secret stores such as Vault, so I imagine it's just as likely someone would want to load function config from them. @turkenh suggested we could pass in connection details, but in that case:

1. Where are the connection details coming from? The XR? What sets them if rendering the XR is dependent on the function?
2. What if we need to pass in connection details _and_ other function config?

I'm tempted to leave this out of scope for the first pass of this feature and address it later once we have a few more concrete use cases.

design/design-doc-composition-functions.md
```diff
@@ -0,0 +1,599 @@
...
+  `functionConfig` to specify which Helm chart to render.
+* An `items` array (required). The XR and zero or more composed resources.
+* A `results` object (optional). Used to communicate information about the
+  result of a function, including warnings and errors.
+
+Each function in the array would be executed as a pipeline, with each function:
+
+1. Taking the XR and zero or more composed resources as input.
+1. Producing the optionally mutated XR, optionally mutated composed resources,
+   and newly rendered composed resources as output.
```
st3v, line 242: **question**: Are there any assumptions for these newly composed resources to adhere to XRM? More specifically, can my function render arbitrary, namespace-scoped resources?

design/design-doc-composition-functions.md
ron1: In the above example, the `my-cool-function` accepts parameter `spec.version`. Rather than hardcoding this parameter into the functionConfig, I would like to capture this value in the XR and then have it patched into the functionConfig. Is it possible to set the functionConfig parameter values in this dynamic fashion?

design/design-doc-composition-functions.md
negz: They're just rendering the resources that Crossplane should reconcile, so they're under the same restrictions as a contemporary Compositon - rendering XRM resources. To be fair I don't think this is actually enforced in either case (resources array style Composition or functions array style), but it's unsupported and the behavior is undefined.

design/design-doc-composition-functions.md
```diff
@@ -0,0 +1,599 @@
...
+functionConfig:
+  apiVersion: database.example.org/v1alpha1
+  kind: Config
+  metadata:
+    name: cloudsql
+  spec:
+    version: POSTGRES_9_6
+items:
+- apiVersion: database.example.org/v1alpha1
+  kind: XPostgreSQLInstance
```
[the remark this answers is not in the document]
negz: Yes, you'll get an XR this its `claimRef` and labels set appropriately.

design/design-doc-composition-functions.md
negz: > Rather than hardcoding this parameter into the functionConfig, is it possible to capture this value in the XR and then have it patched into the functionConfig? WDYT about setting functionConfig parameter values in this dynamic fashion?

Why pass it in via the `functionConfig` when the XR itself will always be available in the `items` array?

ron1: Most kpt functions are reusable. An XRM function that depends on a specific XR schema will not be reusable. So providing an easy way to map XR field values to functionConfig values seems important to enable reusable functions.

The functions in the kpt catalog typically accept a few simple inputs. I would expect a catalog of XRM functions to be similar. Many XRs have custom types which prevents them from being directly consumed from reusable XRM functions.

negz: > Most kpt functions are reusable

I am assuming most XRM functions - say ~70% - will not be reusable. The primary motivation behind this design is to provide a way to take an XR as input and output composed resources. I'd consider "generic" functions (e.g. those for validation etc) to be a secondary concern.

Perhaps there are use cases I am not seeing here - I'd be happy to hear them.

ron1: For example, say an existing reasonably complex XR has a spec field and value that I want to add as a label to all my composed resources. Kpt function https://catalog.kpt.dev/set-labels/v0.1/ uses the functionConfig to accomplish this. A reusable XRM function could do likewise. I would not want to rewrite this set-labels function for every different source XR. Rather, I would like to pick the field value out of the XR and patch it into the set-labels functionConfig to allow my XR and function to remain decoupled.

design/design-doc-composition-functions.md
```diff
@@ -0,0 +1,599 @@
...
+
+Under this proposal each function is the entrypoint of an OCI image, though the
+API is designed to support different function implementations (such as webhooks)
+in th future. The updated API would affect only the `Composition` type - no
+changes would be required to the schema of `CompositeResourceDefinitions`, XRs,
+etc.
+
+Notably the functions would not be responsible for interacting with the API
+server to create, update, or delete composed resources. Instead they instruct
+Crossplane which resources should be created, updated, or deleted.
```
salaboy, line 140: 100% on this, something that I am not liking too much is the fact that Crossplane seems to be owning the lifecycle of these functions. The function definition in the previous example showed the container image for the function and I wouldn't go that far. The composition can check if the functions are available by checking by reference. Check comment above.

design/design-doc-composition-functions.md
```diff
@@ -0,0 +1,599 @@
...
+kind: Composition
+metadata:
+  name: example
+spec:
+  compositeTypeRef:
+    apiVersion: database.example.org/v1alpha1
+    kind: XPostgreSQLInstance
+  functions:
+  - name: my-cool-function
+    type: Container
```
salaboy, line 127: This implies the Composition owns the function lifecycle and I really get scared when I see this much coupling. Why not rely on references to the function resources? For example, you can have a reference to a Kubernetes Service which is pointing to a running function, if the Service is referenced in the composition is not available, then the composition is not ready to be used.

design/design-doc-composition-functions.md
```diff
@@ -0,0 +1,599 @@
...
+    # Configuration specific to `type: Container` functions.
+    container:
```
salaboy, line 163: same as before.. this feels like mixing logical composition with runtime parameters of how the functions will run.

design/design-doc-composition-functions.md
```diff
@@ -0,0 +1,599 @@
...
+1. The XR and one or more resources. This would be the case if either:
+   1. The Composition included both a `resources` array and a `functions` array.
+      In this case the `resources` array is computed first then passed to
+      `functions`.
+   1. This was not a newly created XR, but instead an XR that had already
+      created its composed resources and was now being updated.
+
+In the case of `Container` functions this input and output would correspond to
+stdin and stdout. This may differ for future function implementations; for
+example a hypothetical webhook function implementation might take the input as
```
salaboy, line 258: This is also a bit confusing, why wouldn't the main kind of function be HTTP based? It looks like by default the Container function reads from stdin and write to the stdout. I believe that having a webhook approach by default will make everything easier as there are tools solving 98% of all the needs here.

design/design-doc-composition-functions.md
```diff
@@ -0,0 +1,599 @@
...
+   element in the `items` array by convention.
+1. `name` associates a composed resource entry in the `items` array with its
+   corresponding entry in the `resources` array (if any).
+
+While KRM-function-like this API is not KRM function compatible. See the
+[Alternatives Considered](#alternatives-considered) section for details on why.
+
+### Running Container Function Pipelines
+
+Crossplane is almost always deployed as a container in a Kubernetes Pod. This
```
salaboy, line 326: Going with my previous comments, Crossplane running functions or pipelines of functions shouldn't be a thing in my opinion. Crossplane doing a sequence of HTTP requests to orchestrate the functions required by a composition sounds like a much simpler and cleaner approach.

design/design-doc-composition-functions.md
```diff
@@ -0,0 +1,599 @@
...
+present time but likely will in future.
+
+### Using Webhooks to Run Functions
+
+Crossplane could invoke functions by calling a webhook rather than running an
+OCI container. In this model function input and output would still take the form
+of a `ResourceList`, but would be HTTP request and response bodies rather than a
+container's stdin and stdout.
+
+The primary detractor of this approach is the burden it puts on function authors
```
salaboy, line 506: While I see this as a valid concern, the complexities of running and owning the lifecycle of a whole function approach in Kubernetes inside Crossplane sounds like an overkill. I am of course biased because I am working on Knative Functions, but if I have a clear interface from Crossplane with inputs and outputs, the functions side of things is already optimized by other projects, so running and operating these functions is already the concern of these other frameworks. The approach that the project MetaController took follows this webhook approach and it took me 10 minutes to create a function to create a metacontroller controller. Notice the big overlap of APIs between metacontroller (https://metacontroller.github.io/metacontroller/guide/create.html) and this proposal, because I do see that implementing what this proposal is describing around running functions and pipelines of functions seems extremely hard to implement and messy. I will be happy to have a call about this in more detail if needed.

salaboy:
@negz @st3v @turkenh folks.. I hope my comments here are not a problem. I am hoping to be able to help with this as I am quite passionate about functions. As I've mentioned in one of my comments, as a user I believe that the metacontroller approach is the way to go: https://metacontroller.github.io/metacontroller/guide/create.html, and I will be more than happy to build a PoC around that with Crossplane.. because at least in my head adding function references to compositions to run custom logic shouldn't be that complicated :)

design/design-doc-composition-functions.md
negz: What you're describing sounds close to the webhook function variant that we envision adding at some point in the future. In the meantime I think it's important to support OCI based functions because they're dramatically easier to build and deploy than spinning up something that can serve a webhook (e.g. Deployment + Service + Certs etc).

> I really get scared when I see this much coupling.

Why?

design/design-doc-composition-functions.md
negz: See comment above.  There's a lot less involved to package function logic as an OCI containerized process than there is in spinning up something to serve webhooks.

design/design-doc-composition-functions.md
negz: I'm going to resolve this one since you have a handful of other comment threads around the same concern.

design/design-doc-composition-functions.md
negz: > Running and owning the lifecycle of a whole function approach [...] inside Crossplane

FWIW we'll likely iterate on this design to reflect the function runner being a swappable sidecar with Crossplane speaking e.g. gRPC to invoke functions, rather than something that literally happens inside the Crossplane process. I do still feel strongly that said sidecar should be invoking OCI images though. We also considered having a webhook that would run OCI based functions but the `Composition` API is weaker in this situation when Crossplane itself can't speak in OCI image terms.

negz:
@salaboy Thanks for your comments. I don't immediately share your concerns - my feeling is that what's proposed here should make it easier for folks to build and deploy functions vs requiring them to do so as separate long running processes. Consider especially:

* Folks who are not Kubernetes experts
* Configuring secure HTTP based authentication between Crossplane and functions.
* Providing SDKs to make writing functions easier.
* Functions that leverage / are light wrappers on other tools (e.g. Helm) rather than complete bespoke logic.

I'm just getting back around to working on this - happy to chat with you about it some time.

design/design-doc-composition-functions.md
negz: I'm not opposed to a feature that patches from XR -> function config to allow for functions that are unaware of the XR's schema. That said, my inclination is to avoid building that in the first pass. We can add it later once clear use cases arise.

negz:
https://gist.github.com/negz/c5a5b4f82b06882906b08af377a609a4

Drawing attention to the above gist. I used this to sketch out the `FunctionIO` API, and it sparked a good discussion between @turkenh and I in the comments.

negz:
From @turkenh at https://gist.github.com/negz/c5a5b4f82b06882906b08af377a609a4?permalink_comment_id=4329791#gistcomment-4329791:

> Actually I am wondering if we need to pass this connectionDetails field at all? If we want to pass existing connection secrets as input and expect XR connection secret as output (as we discussed in a separate thread), I believe outputting the final XR secret should be enough.

I was thinking the same originally, but I don't think we can rely on Functions outputting the XR secret. Consider that:

* Many (most?) Functions will simply be logic that takes an XR as input and produces (the manifests of) one or more MRs as output. They won't actually create those MRs, so they won't know what their connection details are.
* Sometimes a contemporary P&T style `resources` array will be computed before the array of `functions`. How would the functions know to include any connection detail (mappings) from the `resources` array?

I think we need Functions to be able to handle at least what P&T style Composition can do, with regard to connection details. So that would be:

* After specifying a new MR that Crossplane should create, tell Crossplane how to derive XR connection details from that MR's connection details.
* Support returning an explicit XR connection detail (i.e. don't derive it, just use this value). This could use the same API as P&T `FromValue` style connection details, but the advantage of doing so in a Function is that the value could be somehow computed (e.g. by calling an external system).

design/design-doc-composition-functions.md
```diff
@@ -0,0 +1,753 @@
...
+        settings:
+          tier: db-custom-1-3840
+          dataDiskType: PD_SSD
+          dataDiskSizeGb: 20
+      writeConnectionSecretToRef:
+        namespace: crossplane-system
+        name: cloudsqlpostgresql-conn
+  connectionDetails:
+  - name: hostname
+    fromConnectionSecretKey: hostname
```
negz, line 365: From @turkenh at https://gist.github.com/negz/c5a5b4f82b06882906b08af377a609a4?permalink_comment_id=4329791#gistcomment-4329791:

> Actually I am wondering if we need to pass this connectionDetails field at all? If we want to pass existing connection secrets as input and expect XR connection secret as output (as we discussed in a separate thread), I believe outputting the final XR secret should be enough.

I was thinking the same originally, but I don't think we can rely on Functions outputting the XR secret. Consider that:

* Many (most?) Functions will simply be logic that takes an XR as input and produces (the manifests of) one or more MRs as output. They won't actually create those MRs, so they won't know what their connection details are.
* Sometimes a contemporary P&T style `resources` array will be computed before the array of `functions`. How would the functions know to include any connection detail (mappings) from the `resources` array?

I think we need Functions to be able to handle at least what P&T style Composition can do, with regard to connection details. So that would be:

* After specifying a new MR that Crossplane should create, tell Crossplane how to derive XR connection details from that MR's connection details.
* Support returning an explicit XR connection detail (i.e. don't derive it, just use this value). This could use the same API as P&T `FromValue` style connection details, but the advantage of doing so in a Function is that the value could be somehow computed (e.g. by calling an external system).

negz: An alternative to including `connectionDetails` and `readinessChecks` would be to provide functions with enough context to figure this out themselves - i.e. providing functions with the status and connection details of composed resources.

Let's say for example we returned to an API more like `ResourceList` (so a big array of resources, with some way to differentiate between the XR, composed resources, and other types). We could:

First reconcile of a new XR:

1. Crossplane passes the XR as the only element in the `ResourceList` to a Function.
2. Function responds by adding one or more composed resources to the `ResourceList`, returns it to Crossplane.
3. Crossplane creates those composed resources.

Second reconcile of the above XR:

1. Crossplane passes the XR, and composed resources as elements in a `ResourceList` to a Function. It also passes the XR and composed resources' connection details - let's say as `kind: Secret`.
2. Function responds by mapping connection details from composed resource Secrets to XR Secret as appropriate.
3. Function updates XR status (probably derived from composed resource status) as appropriate.

I don't hate this alternative. I think the key upside is that it's much more flexible than the current `FunctionIO` approach, but it does have a few downsides (or at least quirks):

* Connection details aren't actually a `kind: Secret` if you use an external secret store (e.g. Vault). Representing them in functions as `kind: Secret` could be misleading or confusing to Function authors.
* Always having to include logic in your Function to derive XR connection details and status from composed resources could become repetitive and painful fast.
* It's more work for function authors to 'find things' in the array. For example you can't just look at the `composite` field - you need to iterate over all resources and find the one that's the XR.
* We'd need to create associations/relationships within the array - e.g. "this Secret represents the connection details for the composed resource named `foo`".
* We'd have to rely on annotations to represent types (e.g. XR vs composed vs connection details), names, and relationships

turkenh: I am on the same page regarding that `FunctionIO` approach works better than `ResourceList` approach.

If we need a more flexible solution while eliminating the quirks you've mentioned except the second one (unless it introduces others), a combined solution could be passing connection secret data under a `connectionData` field and expecting XR secret from `composite.connectionData`, instead of passing **how** to build it with `connectionDetails`.

```yaml
kind: FunctionIO
config:
  kind: Config
  metadata:
    name: cloudsql
  spec:
    version: POSTGRES_9_6
composite:
  resource:
    kind: XPostgreSQLInstance
    metadata:
      name: my-db
    spec:
      parameters:
        storageGB: 20
      compositionSelector:
        matchLabels:
          provider: gcp
  connectionData:
    dbURI: mysql://foo:bar@hostname:3306
resources:
- name: cloudsqlinstance
  resource:
    kind: CloudSQLInstance
    spec:
      forProvider:
        databaseVersion: POSTGRES_9_6
        region: us-central1
        settings:
          tier: db-custom-1-3840
          dataDiskType: PD_SSD
          dataDiskSizeGb: 20
      writeConnectionSecretToRef:
        namespace: crossplane-system
        name: cloudsqlpostgresql-conn
  connectionData:
    username: foo
    password: bar
    host: hostname
    port: 3306
  readinessChecks:
  - type: None
```

> Always having to include logic in your Function to derive XR connection details and status from composed resources could become repetitive and painful fast.

Not sure if it would be painful for connection secrets with an approach like above. But I can kind of get that it could be repetitive to build the readiness condition calculation in each function by checking all composed resources statuses instead of having a common logic. But I believe this might be acceptable as a trade off for more flexibility.

negz: @turkenh Maybe there's a best of both worlds approach? Something like:

```yaml
kind: FunctionIO
config:
  kind: Config
  metadata:
    name: cloudsql
  spec:
    version: POSTGRES_9_6
composite:
  resource:
    kind: XPostgreSQLInstance
    metadata:
      name: my-db
    spec:
      parameters:
        storageGB: 20
      compositionSelector:
        matchLabels:
          provider: gcp
  # XR connection details can be specified by an entry in the resources array
  # too. Allowing them to be specified here decouples them from a specific
  # resources array entry - e.g. it allows a function to 'build' an connection
  # detail for an XR by merging multiple resource connection details, or from
  # some entirely unrelated source.
  connectionDetails:
  - type: FromValue
    name: uri
    value: postgresql://admin:password@db.example.org:5432
  # Connection details that already exist - i.e. have been persisted to and read
  # back from a secret store. Read-only to the function.
  existingConnectionDetails:
  - name: uri
    value: postgresql://admin:password@db.example.org:5432
resources:
- name: cloudsqlinstance
  resource:
    kind: CloudSQLInstance
    spec:
      forProvider:
        databaseVersion: POSTGRES_9_6
        region: us-central1
        settings:
          tier: db-custom-1-3840
          dataDiskType: PD_SSD
          dataDiskSizeGb: 20
      writeConnectionSecretToRef:
        namespace: crossplane-system
        name: cloudsqlpostgresql-conn
  connectionDetails:
  - type: FromConnectionSecretKey
    name: hostname
    fromConnectionSecretKey: hostname
  - type: FromValue
    name: username
    value: admin
  # Connection details that already exist - i.e. have been persisted to and read
  # back from a secret store. Read-only to the function.
  existingConnectionDetails:
  - name: hostname
    value: db.example.org
  - name: username
    value: admin
  readinessChecks:
  - type: None
```

Under this approach functions can set `connectionDetails` to configure how XR connection details are set (either `FromConnectionSecretKey` or `FromValue`), but they also get to see any existing (already persisted) connection details - which they could use to create new `FromValue` entries in the `connectionDetails` array.

negz: @turkenh The main thing I don't like about my most recent suggestion above is that it starts to feel like it's mixing desired state and actual state without clear differentiation.

One way to address that would be to nest everything in `FunctionIO` under spec and status, like a typical Kubernetes object. The `status` object would be read-only to functions, and exist strictly to provide them with context about the actual state of the world (as opposed to the desired state, as specified by the Composition's `resources` array, and/or the logic of any previous functions that have run before them.

```yaml
kind: FunctionIO
spec:
  config:
    kind: Config
    metadata:
      name: cloudsql
    spec:
      version: POSTGRES_9_6
  composite:
    resource:
      kind: XPostgreSQLInstance
      metadata:
        name: my-db
      spec:
        parameters:
          storageGB: 20
        compositionSelector:
          matchLabels:
            provider: gcp
    connectionDetails:
    - type: FromValue
      name: uri
      value: postgresql://admin:password@db.example.org:5432
  resources:
  - name: cloudsqlinstance
    resource:
      kind: CloudSQLInstance
      spec:
        forProvider:
          databaseVersion: POSTGRES_9_6
          region: us-central1
          settings:
            tier: db-custom-1-3840
            dataDiskType: PD_SSD
            dataDiskSizeGb: 20
        writeConnectionSecretToRef:
          namespace: crossplane-system
          name: cloudsqlpostgresql-conn
    connectionDetails:
    - name: hostname
      fromConnectionSecretKey: hostname
    readinessChecks:
    - type: None
status:
  composite:
    resource:
      kind: XPostgreSQLInstance
      metadata:
        name: my-db
      spec:
        parameters:
          storageGB: 20
        compositionSelector:
          matchLabels:
            provider: gcp
      status:
        state: ACTIVE
        conditions:
        - type: Ready
          status: True
    connectionDetails:
    - name: uri
      value: postgresql://admin:password@db.example.org:5432
  resources:
  - name: cloudsqlinstance
    resource:
      kind: CloudSQLInstance
      spec:
        forProvider:
          databaseVersion: POSTGRES_9_6
          region: us-central1
          settings:
            tier: db-custom-1-3840
            dataDiskType: PD_SSD
            dataDiskSizeGb: 20
        writeConnectionSecretToRef:
          namespace: crossplane-system
          name: cloudsqlpostgresql-conn
      status:
        atProvider:
          state: ACTIVE
        conditions:
        - type: Ready
          status: True
    connectionDetails:
    - name: hostname
      value: db.example.org
```

turkenh: I think we are getting there 🙂

Looking better than what I suggested above, but I am a bit confused who reads/writes specs vs status, i.e. what is input to the function and what is output of the function. I am wondering if we could make it more clear with the following:

Input to the function:

```yaml
kind: FunctionIO
config:
  kind: Config
  metadata:
    name: cloudsql
  spec:
    version: POSTGRES_9_6
observed:
  composite:
    resource:
      kind: XPostgreSQLInstance
      metadata:
        name: my-db
      spec:
        parameters:
          storageGB: 20
        compositionSelector:
          matchLabels:
            provider: gcp
      status:
        state: ACTIVE
        conditions:
        - type: Ready
          status: True
    connectionDetailsData:
    - name: uri
      value: postgresql://admin:password@db.example.org:5432
  resources:
  - name: cloudsqlinstance
    resource:
      kind: CloudSQLInstance
      spec:
        forProvider:
          databaseVersion: POSTGRES_9_6
          region: us-central1
          settings:
            tier: db-custom-1-3840
            dataDiskType: PD_SSD
            dataDiskSizeGb: 20
        writeConnectionSecretToRef:
          namespace: crossplane-system
          name: cloudsqlpostgresql-conn
      status:
        atProvider:
          state: ACTIVE
        conditions:
        - type: Ready
          status: True
    connectionDetailsData:
    - name: hostname
      value: db.example.org
```

Output of the function:

```yaml
kind: FunctionIO
desired:
  composite:
    resource:
      kind: XPostgreSQLInstance
      metadata:
        name: my-db
      spec:
        parameters:
          storageGB: 20
        compositionSelector:
          matchLabels:
            provider: gcp
    connectionDetails:
    - type: FromValue
      name: uri
      value: postgresql://admin:password@db.example.org:5432
  resources:
  - name: cloudsqlinstance
    resource:
      kind: CloudSQLInstance
      spec:
        forProvider:
          databaseVersion: POSTGRES_9_6
          region: us-central1
          settings:
            tier: db-custom-1-3840
            dataDiskType: PD_SSD
            dataDiskSizeGb: 20
        writeConnectionSecretToRef:
          namespace: crossplane-system
          name: cloudsqlpostgresql-conn
    connectionDetails:
    - name: hostname
      fromConnectionSecretKey: hostname
    readinessChecks:
    - type: None
```

What I did was:
- Moved `spec.config` up to top level as `config`.
- Renamed `spec` as `desired`.
- Renamed `status` as `observed`.
- Renamed `status.connectionDetails` as `observed.ConnectionDetailsData` (not mandatory but to indicate they are different types).

turkenh: Also with the above, `status` could still be used for error reporting by the function. `input`/`output` could be another alternative to `observed`/`desired`.

negz: @turkenh I considered `observed` and `desired` too - I actually like them more than `spec` and `status` but I'm not sure they're "better" enough to warrant diverging from the typical `spec` and `status` names.

> I am a bit confused who reads/writes specs vs status, i.e. what is input to the function and what is output of the function. 

FWIW I don't think `input` and `output` make sense in the context of a pipeline of functions, where the output of one function is the input to another. Instead I think of all functions in the pipeline as working together to "accumulate" a desired state, by each mutating the spec.

So, under my proposal:

* All functions may write to `spec` - the "desired state" as accumulated by the functions.
* No functions may write to `status` - the "observed state" used to inform the functions.

I do like the idea of moving `config` to the top level, since it's different from the function's desired state. As I write this out I'm thinking I do like `observed` and `desired` better than `spec` and `status` because they have different meaning from the typical `spec` and `status` - it's probably confusing that `status` isn't actually the status of a function.

I'll push a commit to reflect the changes we've discussed here.

negz: > I'll push a commit to reflect the changes we've discussed here.

@turkenh Done. Let me know what you think (feel free to resolve this thread if it looks good.

design/design-doc-composition-functions.md
```diff
@@ -0,0 +1,778 @@
...
+        compositionSelector:
+          matchLabels:
+            provider: gcp
+      status:
+        conditions:
+        - type: Ready
+          status: True
+    connectionDetails:
+    - name: uri
+      value: postgresql://db.example.org:5432
```
negz, line 290: @turkenh We can currently get, but not list connection details, right? Is it feasible to extend the connection details API with list support? I believe we'd need to do so to build this.

turkenh: Do you mean listing secrets or connection details (i.e. k/v pairs in a given secret) ?

We get a [secret](https://github.com/crossplane/crossplane-runtime/blob/master/pkg/connection/store/store.go#L43) with [FetchConnection](https://github.com/crossplane/crossplane-runtime/blob/bce61005a175d286ef5a7a75c4f83fcd273ec6dd/pkg/connection/manager.go#L128) which contains all connection details of an MR/XR as key values. 

This should be enough here I guess.

design/design-doc-composition-functions.md
```diff
@@ -266,115 +268,151 @@ config:
...
   with an arbitrary schema that may be used to provide additional configuration
   to a function. For example a `render-helm-chart` function might use its
-  `config` to specify which Helm chart to render.
-* A `composite` object (required). This is the XR being reconciled.
-* A `resources` array (optional). Zero or more composed resources.
+  `config` to specify which Helm chart to render. Functions need not return
+  their `config`, and any mutations will be ignored.
+* An `observed` object (required). This reflects the observed state of the XR,
+  any existing composed resources, and their connection details. Functions must
+  return the `observed` object unmodified.
```
turkenh, line 303: > Functions must return the `observed` object unmodified.

Should Crossplane do something to ensure that, e.g. ignore changes in `observed` before passing to next function in the pipeline?

negz: Yeah I think we should - I'm going to hold off on explicitly stating that for the moment though.

turkenh (approved):
Thank you @negz! Really excited to see this coming 🚀

Issue #2898, opened 2022-02-14
Documentation page for CRD migrations


muvaf:
### What problem are you facing?

We have breaking and non-breaking changes that change the API version of CRDs. So far, a migration guide is created for each release, like this [one](https://github.com/crossplane/provider-aws/blob/master/cluster/UPGRADE.md), but the steps are very similar between providers.

### How could Crossplane help solve your problem?

We can have a documentation page that we can point to from release notes as the migration guide that covers a set of known cases.

plumbis:
This would be specific to how to upgrade providers, not Crossplane pod upgrades, correct?

jeanduplessis:
I agree this feels like its related to details of the provider, not core crossplane.

PR #2907, opened 2022-02-15
docs: add migration guide for providers

muvaf:

We're having cases where we need to provide a migration guide in providers like [this one](https://github.com/crossplane/provider-aws/blob/master/cluster/UPGRADE.md) and this doc will help provider developers to just point to the case here when they need to have their users go through the migration.

Fixes https://github.com/crossplane/crossplane/issues/2898

I have:

Tested the commands.

### DIFF
--- a/docs/guides/migrations.md
+++ b/docs/guides/migrations.md
@@ -1,0 +1,242 @@
+---
+title: Migrations
+toc: true
+weight: 200
+indent: true
+---
+
+# Migrations
+
+Crossplane is built on CustomResourceDefinition (CRD) objects that expose an API
+with a certain schema. As the capabilities of the controller reconciling a CRD
+changes, so does its schema, which is what the users and their systems interact
+with. You will find that each schema is enumerated by a [version](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/).
+
+Kubernetes provides a way to make sure users of CRDs are able to migrate
+to the new versions without breaking their existing systems in a way that give
+them enough time to make the changes to adopt the new version. The versions in
+this context have certain implications about the API stability guarantees that
+the owner of the CRD promises. Crossplane community follows those guarantees.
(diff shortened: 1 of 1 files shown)


docs/guides/migrations.md
```diff
@@ -0,0 +1,242 @@
...
+they need to provide it.
+
+In this case, your systems are expected to continue to work after the upgrade
+without any immediate changes. In most cases, developers would deprecate the old
+version and warn you about its removal schedule in the release notes. The steps
+described below need to be taken before that removal date.
+
+The important thing to keep in mind here is that Kubernetes API server is able to
+serve multiple versions equally well but the object stored in its persistence
+layer has a single shape. The point of the steps below is that we'll have the
```
jbw976, line 56: is this true? i expected that when a new version is released, that it will typically become the new stored version, so we're telling users to update their manifests to use the new version, and that new version is what will be the 1 stored version in the api-server too.

I have that wrong? 🤔

muvaf: This is what we do but we don't have to, i.e. you can choose an old or even non-served schema as the storage one. The only restriction is that you can only ever have a single storage version. You can see the toggle [here](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#customresourcedefinitionversion-v1-apiextensions-k8s-io).

jbw976: Thanks @muvaf - so when we bump something from say v1alpha1 -> v1beta1, we leave `storage: false` on the new v1beta1 version?  When does the new v1beta1 version eventually becoming the `storage: true` version?

Maybe a little table example could be helpful in this doc, that very clearly shows the matrix of expected served/storage for each version involved in the migration? 😊

negz:
@muvaf Should we move forward with getting this merged? Maybe @turkenh (or someone else from your team) could do a review pass?

negz:
@muvaf This has been in draft for a while and I'm doing a bit of PR backlog cleanup so I'm going to close it. Feel free to reopen if/when you get back around to it.

PR #2918, opened 2022-02-21
One-pager for CRD scaling discussions in Crossplane

### LINKED ISSUE #2895: Consolidation of CRD Scaling Issues

Previously in https://github.com/crossplane/terrajet/issues/47, #2649 and https://github.com/kubernetes/kubernetes/issues/105932, we have observed high resource consumption in the Kubernetes API Server when over 500 CRDs are installed. Especially with local `kind` clusters, installing [provider-jet-aws](https://github.com/crossplane-contrib/provider-jet-aws) with over 700 CRDs could render the cluster unresponsive on machines with relatively low CPU resources, due to the high CPU consumption. And with the managed Kubernetes offerings such as GKE, EKS and AKS, with varying levels dependent on the cluster configurations, we had observed API service disruptions as we installed 100s or 1000s of CRDs. While discussing those issues as the Crosplane community, we had tried a couple of different workarounds detailed in #2649, such as installing CRDs in smaller batches and giving the API Server some time to finish aggregating their OpenAPI specs.

Meanwhile after getting in touch with the upstream Kubernetes maintainers, we learned that they already were working on https://github.com/kubernetes/kube-openapi/pull/251, which would introduce lazy marshalling for the aggregated OpenAPI specs of the installed CRDs served from the `/openapi/v2` endpoint. Collecting profiling data and builing `kind` images with that PR and testing them in local `kind` clusters, we observed substantial [improvements](https://github.com/crossplane/crossplane/issues/2649#issuecomment-956914007) in peak resource consumption. However, as discussed in #2649 there were still some open points left:
- Discovery client throttles itself (client-side throttling) while discovering the exposed API, if its cache is unestablished (first time discovery) or is invalidated (subsequent discoveries)
- Metrics collected from the API server reveal a continued memory consumption of ~ 2.75 GiB for ~600 CRDs and memory consumption increases with the number of CRDs

Although we could not actually observe the effects of the aggregated OpenAPI spec lazy-marshalling [optimization](https://github.com/kubernetes/kube-openapi/pull/251) in a managed Kubernetes offering [1], we expected it to alleviate control plane service disruptions. Starting with Kubernetes versions `v1.20.13`, `v1.21.7`, `v1.22.4`, `v1.23.0` this fix is available and as of this writing:
- `v1.23.1` and `v1.22.4` are available for most GKE regions via the `rapid` [GKE release channel](https://cloud.google.com/kubernetes-engine/docs/concepts/release-channels), and
- `v1.22.4` and `v1.21.7` are available for most AKS regions.

This allowed us to test the hypothesis with the lazy-marshalling optimization and we have recently made a number of tests with the managed Kubernetes offerings. *Please note that the following are reports from single experiments, and only server-side issues are reported.* For all clusters even if there is no API service disruption, we always experience client-side throttling.

### Related issues

 - [ ] #2869

<hr/>

#### GKE Zonal
On a GKE `v1.23.1` zonal cluster with three worker nodes of type `e2-medium`, it took `56 min` for the `ProviderRevision` to acquire the `Healthy` condition for a [provider-jet-aws](https://github.com/crossplane-contrib/provider-jet-aws/releases/tag/v0.4.0-preview) installation with 763 CRDs. During this period, the cluster needed at least one repair, and there was over 50 min of service disruption. After a slight confusion regarding worker-node HA and control-plane HA, [this thread](https://issuetracker.google.com/u/1/issues/192000009?pli=1) suggests using regional clusters for control-plane high availability, and states that control-plane service disruptions are expected for zonal GKE clusters as GKE is scaling the managed control-plane under load. However, as discussed next, the situation with a regional cluster was not good either.

<hr/>

#### GKE Regional
A `v1.23.1` cluster was provisioned in the `us-east1` region with one `e2-medium` node in each availability zone, with a total of 3 worker nodes in 3 zones. Although it took only `~ 150 s` for the `ProviderRevision` to acquire the `Healthy` condition for the `provider-jet-aws@v0.4.0-preview` installation, the regional GKE cluster went through the repairing mode at least three times afterwards. In between these "RUNNING" and "RECONCILING" states of the regional cluster, we have observed different kinds of errors in response to the `kubectl` commands run, notably connection errors and I/O timeouts while reaching the API server. It took over an hour for the cluster to stabilize. However, the control-plane was intermittently available for short periods during this period.

<hr/>

#### EKS
When we performed these tests, none of the Kubernetes versions consuming the lazy-marshalling optimization were available for EKS clusters. However, we used a `v1.21.5-eks-bc4871b` 3-worker node cluster to test a `provider-jet-aws@v0.4.0-preview` installation. Grafana dashboards indicate a control-plane with two members from the very beginning before `provider-jet-aws` package is installed into the system. It took the `ProviderRevision` about 2 min to acquire the `Healthy` condition. The cluster stayed stable after the provider installation. A reexamination of the Grafana dashboards after 3 hours reveals that the two-member control-plane was scaled up and down during this period. Apart from the client-side throttling issues, control-plane is stable with a `provider-jet-aws` installation.

<hr/>

#### AKS
A `v1.22.4` cluster was provisioned with two worker nodes. It took `~112 s` for the `ProviderRevision` to acquire the `Healthy` condition for a `provider-jet-aws@v0.4.0-preview` installation. The cluster stayed stable after the provider installation. Also installed [`provider-jet-azure@v0.7.0-preview`](https://github.com/crossplane-contrib/provider-jet-azure/releases/tag/v0.7.0-preview) on this cluster making the total CRD count `1430`. It took `91 s` for `provider-jet-azure`'s ProviderRevision to acquire the `Healthy` condition. It takes `~ 40 s` for the discovery client to refresh its cache. And we observe some pod restarts due to timeouts. But the cluster stays stable. However, adding [`provider-jet-gcp@v0.2.0-preview`](https://github.com/crossplane-contrib/provider-jet-gcp/releases/tag/v0.2.0-preview) as a third provider has finally "breaks" the cluster. Crossplane is having a hard time installing the package. 

We also repeated these experiments with a two worker node AKS `v1.20.13` with similar results: It took `107 s` for the ProviderRevision to acquire the `Healthy` condition. The cluster stayed stable after the provider installation. Reaching a total of 780 CRDs, the cluster has no severe issues. Client-side throttling kicks in as expected.

<hr/>

We also need a deeper understanding on which metrics are being employed by the Cloud providers when deciding to scale up their control planes. Usual metrics are CPU/memory consumption and/or utilization of the control-plane components but they could as well be depending on metrics such as the number of installed CRDs and/or API Server response time SLIs, etc. 

Another dimension where we need clarification is why we observe control plane service disruptions when the managed control plane is scaling up. It's perfectly expected, at some point as we continue adding more CRDs to the cluster, the control plane will need to scale up but if it has high-availability features, then why are we not isolated from the effects?

We are now at a point where there are several upstream Kubernetes issues and PRs, Crossplane issues and discussions happening at different places, which makes it hard to figure out the exact problems and root causes. Also under #2649, there is still some discussion going on. This motivates us to open this issue so that we can decide how to move on.

[1]: Currently, we are not aware of a configuration option or another mechanism that would allow us to replace the `kube-openapi` with a custom build for an AKS/EKS/GKE cluster.

We can have a one-pager that includes:
- The set of problems we are still observing like the continuos memory consumption and client-side throttling issues with the discovery client. We also need a deeper understanding of these issues and their mechanics. 
- A deeper understanding of how Cloud providers are scaling their Kubernetes control planes and which metrics they are depending on. And also a deeper understanding on why we experience service disruptions even with HA control planes when installing 1000s of CRDs. We can achieve this understanding by opening issues to the Cloud providers for them to investigate and having discussions with them.
- A definition for the ideal state like reasonable response times on certain kubectl requests (cached and uncached) when there are 100s of CRDs and memory usage per 100s CRDs.
- It would be helpful to have scalability goal(s) for Crossplane in terms of the number of (active) managed resource **types** (CRDs) to support in a cluster with an appropriate size. This can be input for a Crossplane scalability guide and/or for advocating Crossplane use-cases when CRD scalability thresholds are being discussed upstream, etc.
- The tooling to quickly reproduce the cases and test the progress against the ideal state. Go/Bash tools to run these tests for a given cluster and report the comparison. References to all known issues and PRs about the topic.

This would help all stakeholders to see all the data we have, reproduce the problems quickly and make more informed decisions about how to move forward. The tooling could be part of the Kubernetes Conformance tests at one point but it’s not in the scope of this issue.

Another goal we should work towards is to get CRDs as a scalability dimension in the Kubernetes [thresholds document](https://github.com/kubernetes/community/blob/master/sig-scalability/configs-and-limits/thresholds.md) as discussed [here](https://github.com/kubernetes/kubernetes/issues/105932#issuecomment-954650339).

ulucinar:

Fixes #2895

With this one-pager proposal, we would like to establish a common understanding in the Crossplane community on the issues around CRD scaling, and the paths to possible solutions to these issues.

I have:

N.A.


design/one-pager-crd-scaling.md
```diff
@@ -0,0 +1,150 @@
+# Performance Issues of Deploying Thousands of CRDs
+* Owner: Alper Rifat Uluçınar (@ulucinar)
+* Reviewers: Crossplane Maintainers
+* Status: Draft
+
+## Background
+With the release of the [Terrajet](https://github.com/crossplane/terrajet) based providers, the Crossplane community has become more aware of some upstream scaling issues related to custom resource definitions. We did some early analysis such as [[1]] and [[2]] to get a better understanding of these issues and as we will discuss in more detail in the “Issues” section, the broader K8s community has already been aware of especially the client-side throttling problems for some time. It’s also not Crossplane alone. [Azure Service Operator](https://github.com/Azure/azure-service-operator), or [GCP Config Connector](https://github.com/GoogleCloudPlatform/k8s-config-connector) are projects that rely on Kubernetes [custom resources](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/) as an extension mechanism and have many CRDs representing associated Cloud resources.
```
negz, line 7: Could you please wrap this markdown at 80 chars? In part because it's consistent with our other designs, and in part because it allows more 'targeted' comments in GitHub - currently it's only possible to comment on entire paragraphs (rather than a line within a paragraph) because each paragraph is one long line.

There's a few tools that can do this automatically - I use https://marketplace.visualstudio.com/items?itemName=stkb.rewrap

negz: FWIW I reached out to the ASO and ACK folks and that haven't seen these scaling issues yet, but I fully expect they will in time. ACK is somewhat insulated since they're bundles of smaller controllers, but I believe ASO and KCC take the some "one controller manager for all of a cloud's resources" approach we do.

design/one-pager-crd-scaling.md
```diff
@@ -0,0 +1,150 @@
...
+
+Kubernetes is a complex ecosystem with many moving parts and we need a deeper understanding for the issues around scaling in the dimension of the total number of CRDs per-cluster. This dimension is not yet officially considered in the scalability thresholds document [[3]] but it will be with good probability. So, as the Crossplane community, we would like to have our use cases considered in relevant contexts, and we would like to gain a good understanding so that we can:
```
negz, line 9: > This dimension is not yet officially considered in the scalability thresholds document [[3]] but it will be with good probability.

I think you asked upstream about this right? Is there an issue tracking getting CRDs added to that doc?

design/one-pager-crd-scaling.md
```diff
@@ -0,0 +1,150 @@
...
+`kubectl` maintains a discovery cache for the discovered server-side resources under the (default) filesystem path of `$HOME/.kube/cache/discovery/<host_port>/`. Here, `<host_port>` is a string derived from the API server host and the port number it’s listening on. An example path would be `$HOME/.kube/cache/discovery/exampleaks_8e092dad.hcp.eastus.azmk8s.io_443`, or `$HOME/.kube/cache/discovery/EB788B3B801893B684B4579B2ADF0171.gr7.us_east_1.eks.amazonaws.com`. Under this cache, we have the `servergroups.json` file, which is a JSON-serialized [`v1.APIGroupList`](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#APIGroupList) object. Thus, the cache file `$HOME/.kube/cache/discovery/<host_port>/servergroups.json` holds all of the discovered API GroupVersions (GVs) together with their preferred versions from that API service. And for each discovered API GroupVersion, we have a `serverresources.json` that caches metadata about the discovered resources under that GroupVersion, JSON-serialized as a [`v1.APIResourceList`](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#APIResourceList). This metadata about resources is crucial for various tasks, such as:
```
negz, line 26: I want to note that that while fixing `kubectl` would be very impactful, we may want to go down a level and try get this fixed in `client-go` so that all (Go based) clients that do discovery are fixed.

design/one-pager-crd-scaling.md
```diff
@@ -0,0 +1,150 @@
...
+```
+
+A more realistic example that also performs a final HTTP GET to fetch a `PodList` resource:
+```bash
+rm -fR  ~/.kube/cache/discovery/<host_port> && time (kubectl get pods > /dev/null)
+```
+
+## Action Items
+- We can consider cherry-picking [[5]] to all active release branches `v1.23`, `v1.22`, `v1.21`, as the anticipated release date for the `v1.24` release is in April, 2022.
+- Open issues regarding API service disruptions for managed control-planes (GKE regional, AKS, EKS, etc.), where we expect high-availability.
```
negz, line 117: Some folks on today's community meeting (namely @haarchri and Jillian Hill from Guidewire) mentioned that their AWS professional services teams have noticed increase resource usage even when EKS appears to be working well and reached out to investigate. It sounds like in some cases they're also seeing EKS clusters choke when installing providers with a lot of CRDs - one theory is that some regions have smaller EKS control planes.

Either way it sounds like we might be able to reach out either to our own contacts at AWS or through a company in the Crossplane community to get some insight into how EKS is handling this load.

negz (commented):
Thanks for working on this @ulucinar! Great start.

ulucinar:
`release-1.23` cherry-pick for the kubectl configuration fix is here: https://github.com/kubernetes/kubernetes/pull/108401

design/one-pager-crd-scaling.md
ulucinar: Thanks Nic for the tool suggestion! It was very helpful. I have also removed most of the inline references.

design/one-pager-crd-scaling.md
ulucinar: Not yet Nic. It would be good to have Crossplane use cases clarified to have them on the table for those discussions, and probably use them as a guideline in those discussions. Also, as you mentioned above, it would be great to involve ASO and KCC folks. There was a previous community survey asking for CRD use cases and I believe its results were incorporated in the GA scalability targets for CRDs. It looks like projects like Crossplane, ASO, KCC have more demanding new use cases which could be the motivation in those discussions.

design/one-pager-crd-scaling.md
ulucinar: Agreed. Cache busting for the discovery client is under discussion now. Maybe we can extend this one-pager by mentioning some options considered (cache busting, disabling client-side throttling, etc.).

design/one-pager-crd-scaling.md
ulucinar: That would be great Nic. The support issue we opened for GKE regional clusters was not fruitful, the outcome was it works as expected. What we wanted to learn was basically the metrics GKE autoscalers are using when deciding to scale their managed control-planes up (CPU/memory usage/utilization of control-plane components, kube-apiserver or other component failures, # of CRDs, or some other SLIs, etc.?). One thing I still wonder is how [SLAs](https://cloud.google.com/kubernetes-engine/sla) come into play as Crossplane users install three big providers in their clusters. @AaronME, do you know if it's possible, in cloud consoles or by some other means (like a support ticket), we can inquiry about these SLAs?

design/one-pager-crd-scaling.md
```diff
@@ -0,0 +1,563 @@
...
+  components such as increased resource consumption in `kube-apiserver`
+  proportional to the number of CRDs we install in the cluster. While we expect
+  more resources to be allocated as we install more CRDs, some detailed analysis
+  presented in the [API Server Resource
+  Consumption](#api-server-resource-consumption) section reveals some possible
+  improvement points and some other points such as OpenAPI v3 schema sizes where
+  we should be careful.
+
+
+### Client-side Throttling
```
jbw976, line 64: is it possible to very clearly state (maybe with a table?) which versions of kubectl are released (or targeting releases) that have all the client-side throttling fixes and we can recommend to the community?

jbw976: e.g. it's not super obvious to me (as a casual reader) which versions of kubectl (and also k8s-api) i would need to have a good experience with jet providers.  Maybe we can make that super obvious at the top of the doc so folks coming to this document just looking for guidance on what versions to use to have a good experience can easily find that info without getting lost in the details.  what do you think @ulucinar?

ulucinar: I think it's a good idea @jbw976. Added a `kubectl` version table with their release dates as a recommendation at the beginning of the `Client-side Throttling` section.

ulucinar: Thanks @jbw976. I've also added a table of kube-apiserver versions to the beginning of the `API Server Resource Consumption` section explicitly mentioning about the OpenAPI v2 spec lazy-marshaling change and the related upstream issues.

jbw976: that is super helpful @ulucinar, thank you for making this document more accessible for a broader audience!! 🙇‍♂️

design/one-pager-crd-scaling.md
```diff
@@ -0,0 +1,590 @@
...
+would like to support and that involve large numbers of CRDs:
+1. Installation of a single provider with hundreds of CRDs: As discussed above,
+   installation of a single provider, such as `provider-jet-aws` preview
+   edition, results in ~190 GVs being served by the API server. With the current
+   (`v1.23`) set of client-side throttling parameters of `kubectl`, this has
+   adverse effects on the perceived user performance. However, situation
+   improves as the burstiness and the fill rate of the token bucket rate limiter
+   are increased in `kubectl@v1.24`.
+2. Installation of multiple providers, such as `provider-jet-{aws,gcp,azure}`
+   preview editions, into the same cluster results in ~370 GVs being served by
```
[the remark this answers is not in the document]
ulucinar: As discussed in the `Client-side Throttling` section, experiments done with a custom build of `kubectl` that allows us to specify discovery-client's tbrl parameters reveal that, with tbrl(b=400, r=50.0 qps), client-side throttling is no longer a bottleneck for our current GV counts (despite a delay of ~18s. on a cold cache). This is subject to change if the number of GVs increase because of different combinations of providers or because of API regroupings etc. However, the upstream community sees bumping these limits to cover certain use-cases as increasing the debt ceiling.

design/one-pager-crd-scaling.md
```diff
@@ -0,0 +1,563 @@
...
+  Crossplane pod restarts (again due to timeouts while communicating with the
+  control-plane).
+- And according to our measurements on an EKS cluster with these three large
+  providers installed, `kubectl@v1.24.0`'s discovery client tbrl parameters of
+  `(b=300, r=50.0 qps)` still result in client-side throttling with a cold
+  cache, but using a `b=400` prevents client-side throttling, as expected (due
+  to the total API GV count in the cluster). Also, even if we completely disable
+  discovery client's rate limiter, it takes ~18s to complete the discovery.
+
+## Action Items
```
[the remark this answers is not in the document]
ulucinar: Done. Opened some new issues and added references to these new issues as well as to the existing ones.

design/one-pager-crd-scaling.md
```diff
@@ -0,0 +1,590 @@
...
+- We can consider cherry-picking [[5]] to all active release branches `v1.23`,
+  `v1.22`, `v1.21`, as the anticipated release date for the `v1.24` release is
+  in April, 2022.
+- Open issues regarding API service disruptions for managed control-planes (GKE
```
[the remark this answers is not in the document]
ulucinar: I've opened the tracking issues and referenced them in the document.

design/one-pager-crd-scaling.md
```diff
@@ -0,0 +1,563 @@
...
+  These clients can throttle the requests they make to the API server solely as
+  a function of the [rate limiter] implementation they are using with no
+  feedback from the server. Per our observations, the problematic client-side
+  component that’s responsible for the high number of requests in a unit time
+  (high request rate) is the [discovery client]. In the [Client-side
+  Throttling](#client-side-throttling) section below, we will examine how
+  `kubectl` behaves when there is a large number of CRDs installed in the
+  cluster and how the discovery cache and the discovery client affect the
+  perceived performance of the `kubectl` commands run. 
+- Server-side issues: We also have some observations on the control-plane
```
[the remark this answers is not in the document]
ulucinar: Hi @muvaf,
Thank you for the comments.
I have extended the paragraph where we discuss this change with more details.

design/one-pager-crd-scaling.md
```diff
@@ -0,0 +1,590 @@
...
+  heap space is allocated and `zapcore.newCounters` is no longer the largest
+  allocation site anymore, as expected. This profiling data better emphasizes
+  allocation sites which are inflated with large (deserialized) OpenAPI v3 CRD
+  schemas. One can generate such CRDs using the following command:
+```
+github.com/ulucinar/terrajet-scale/cmd/apigen generate -g 200 -c 200 --property-count 5000
+```
+
+
+## Criteria Set for Ideal State
```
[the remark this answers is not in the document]
ulucinar: I've added a table of benchmark provider installation scenarios summarizing the criteria set under the `Criteria Set for Ideal State` section.

design/one-pager-crd-scaling.md
```diff
@@ -0,0 +1,590 @@
...
+  `v1.22`, `v1.21`, as the anticipated release date for the `v1.24` release is
+  in April, 2022.
+- Open issues regarding API service disruptions for managed control-planes (GKE
+  regional, AKS, EKS, etc.), where we expect high-availability.
+- With some insight on server-side issues, we will also need to pursue a set of
+  issues in kube-apiserver and possibly in other control-plane components.
+- Initiate further discussions with Kubernetes [sig-scalability] community
+  regarding CRD-scalability and bring agreed-upon Crossplane scenarios into
+  their attention.
+
```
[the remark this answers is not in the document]
ulucinar: I've added a new action item for bumping the burst limit of the default discovery client in `client-go`.

muvaf (approved):
Looking great! @ulucinar thank you for tackling such a problem that expands to multiple components in the upstream and tell the story in a consumable way. This is really helpful for both Crossplane and Kubernetes communities!

design/one-pager-crd-scaling.md
matthchr: Dropping by to say: @negz was correct: https://github.com/Azure/azure-service-operator/issues/2920

Issue #2915, opened 2022-02-17
Proposal for merging Jet AWS, Jet Azure and Jet GCP providers with their existing counterparts


muvaf:
### What problem are you facing?

In provider strategy doc, the decision for initial releases of the first three Jet-based providers (AWS, Azure and GCP) was to have them as separate providers for the reasons stated [here](https://github.com/crossplane/crossplane/blob/master/design/design-doc-provider-strategy.md#decision-for-initial-releases). We called out that we'll revisit this decision after more usage and community feedback, which was lacking most probably because we were asking people (in Slack and in https://github.com/crossplane/crossplane/pull/2701) for their opinions about providers that they have not used yet.

Now that we have more and more users using them, it has become apparent that the confusion cost that was called out in the document is very big that almost every conversation regarding Terrajet starts with `Well, which provider should we use now?`. Also, the CRDs of those providers (the ones in non-preview version, i.e. `v1alpha2`) are fully [XRM-compliant](https://github.com/crossplane/crossplane/blob/master/design/one-pager-managed-resource-api-design.md) and have been on alpha for more than a release, meaning they are due to be bumped to beta; so there will need to be migration guarantees, which would make it challenging to make changes afterwards.

### How could Crossplane help solve your problem?

I think it's time that we have another discussion in the community about how to move forward. We can have a short one-pager that is specifically proposing [Option C](https://github.com/crossplane/crossplane/blob/master/design/design-doc-provider-strategy.md#proposal), the addition of Jet-based CRDs and controllers to the existing providers to fill the gaps, i.e. `provider-aws` would have Terrajet-generated CRDs that are reconciled by generic terrajet controller. An addition to that proposal could be to provide migration tooling for users that use jet providers today, which should be straight-forward to tackle for the non-preview CRDs. If we land on keeping them separate like today, then we'd start bumping resources in jet providers from `v1alpha2` to `v1beta1` and move the provider repositories to `crossplane` org.

Note that all of the options are available today in Terrajet for all provider builders. We are only deciding for [Jet AWS](https://github.com/crossplane-contrib/provider-jet-aws), [Jet Azure](https://github.com/crossplane-contrib/provider-jet-azure/) and [Jet GCP](https://github.com/crossplane-contrib/provider-jet-gcp/).

I suggest the proposal PR to be open for 4 weeks to gather the feedback.

muvaf:
See https://github.com/crossplane/crossplane/pull/2930 for context.

PR #2930, opened 2022-02-28
Proposal to merge Jet AWS, Jet Azure and Jet GCP with their classic counterparts

muvaf:

The proposal to merge:
* provider-jet-aws with provider-aws
* provider-jet-gcp with provider-gcp
* provider-jet-azure with provider-azure

Please feel free to leave your thoughts as users, contributors, maintainers or people who still didn't make up their mind about adopting Crossplane.

I'm going to cc the maintainers of these providers and the steering committee of Crossplane.

cc @chlunde @haarchri @Feggah @jbw976 @artursouza @bassam @negz @hasheddan @krishchow @turkenh @ulucinar  @AaronME 

Fixes https://github.com/crossplane/crossplane/issues/2915

I have:

N/A

### DIFF
--- a/design/one-pager-merge-jets-proposal.md
+++ b/design/one-pager-merge-jets-proposal.md
@@ -15,10 +15,10 @@ of coverage in the matter of days. The API of the generated resources are fully
 compliant with [Crossplane Resource Model (XRM)][xrm-doc]. You can read more
 about its internals in [the deep dive blog post series][deep-dive-blogs].
 
-Before releasing the first Terrajet-based providers, we have had a discussion
+Before releasing the first Terrajet-based providers, we have had a discussion on
 whether we'd like to have them as separate providers or have the
 Terrajet-generated CRDs added to the existing classic providers in the [provider
-strategy doc][strategy-doc]. The decision for initial releases of the first
+strategy doc][strategy-doc]. The decision for the initial releases of the first
 three Jet-based providers ([AWS][jet-aws], [Azure][jet-azure] and
 [GCP][jet-gcp]) was to have them as separate providers for the reasons stated
 [here](https://github.com/crossplane/crossplane/blob/master/design/design-doc-provider-strategy.md#decision-for-initial-releases).
@@ -33,7 +33,7 @@ AWS, Azure and GCP is due.
 
 ## Goals
 
-We have the following goals for _the big three_ providers:
+We have the following goals for providers:
 
 * The cloud vendors should maintain their Crossplane provider.
 * The users should have the ability to provision all of their infrastructure
@@ -41,7 +41,7 @@ We have the following goals for _the big three_ providers:
 * The providers should be mature enough for users to depend on them in
   production, i.e. maturity.
 
-Today, we have two Crossplane providers for each of the big three clouds.
+This document aims for a decision for the following providers:
 * AWS
   * Classic with [124 CRDs][provider-aws-crds] with 37 beta CRDs.
   * Jet with [763 CRDs][jet-aws-crds], no beta CRDs.
@@ -52,32 +52,33 @@ Today, we have two Crossplane providers for each of the big three clouds.
   * Classic with [27 CRDs][provider-gcp-crds] with 11 beta CRDs.
   * Jet with [438 CRDs][jet-gcp-crds], no beta CRDs.
 
-Both variants of each reach those goals at different levels but none of them are
-maintained by their cloud vendor. The aim of this document is to capture a
-decision about the future of these provider implementations with the goals
-stated above in mind.
+Both variants of each cloud reach those goals at different levels but none of
+them are maintained by their cloud vendor. The aim of this document is to
+capture a decision about the future of these provider implementations with the
+goals stated above in mind.
 
 ## Proposal
 
-There were three options in the provider strategy doc:
+There were three options in the [provider strategy doc][strategy-doc]:
 * Option A: Completely separate providers.
+  * The initial decision.
 * Option B: One provider but Jet-based resources in their own API group.
 * Option C: One provider and no Jet vs classic difference. Only API version
   would be bumped if the implementation changes.
 
 We should go with the Option C, i.e. use Terrajet in
-[`provider-aws`][provider-aws], provider-azure and provider-gcp as another way
-of generating CRDs and controllers so that they can reach the goal of coverage
-really fast. Then the community can keep investing in that single provider and
-make each CRD mature enough to depend on in production. In this way, we can
-increase the usage by increasing both the coverage and the maturity of the CRDs
-in a consolidated way. This increase in usage can attract cloud vendors to come
-and join the maintainership of these providers. Once they do, they can either
-progressively convert the implementation of the controllers to their choice of
-technology or start with new providers which we can suggest people to use and we
-can provide migration tooling from the community-maintained one. In either case,
-the users would see a clear path forward and we'd not waste any efforts by the
-community.
+[`provider-aws`][provider-aws], [`provider-azure`][provider-azure] and
+[`provider-gcp`][provider-gcp] as another way of generating CRDs and controllers
+so that they can reach the goal of coverage really fast. Then the community can
+keep investing in that single provider and make each CRD mature enough to depend
+on in production. In this way, we can increase the usage by increasing both the
+coverage and the maturity of the CRDs in a consolidated way. This increase in
+usage can attract cloud vendors to come and join the maintainership of these
+providers. Once they do, they can either progressively convert the
+implementation of the controllers to their choice of technology or start with
+new providers which we can suggest people to use and we can provide migration
+tooling from the community-maintained one. In either case, the users would see a
+clear path forward and we'd not waste any efforts by the community.
 
 Note that we are making two decisions here:
 * There will be a single provider.
@@ -105,7 +106,8 @@ following:
   and bump the API version.
 * If an existing classic CRD (hand-written or ACK-generated) has a bug that
   cannot be fixed easily or has a blocker or we continuously get bugs which
-  makes it hard to maintain, consider replacing it with Terrajet.
+  makes it hard to maintain, consider replacing it with Terrajet and if you do
+  bump the API version.
 
 The main idea behind these guidelines is that we'd like to make use of the best
 tooling for a given resource in terms of its maintenance cost and stability. We
@@ -145,16 +147,16 @@ ARM in their TF provider and they've shown interest in generating Crossplane
 based on ARM and Google is using Magic Modules and DCL to generate TF provider,
 so it's likely that they will choose to do that for their Crossplane provider as
 well. [Provider Strategy doc][strategy-doc] goes into more detail about these
-tools. In summary, Terrajet will be a temporary solution that we put in place to
-get more coverage and at some point we'll need to discuss how to change the
-implementation to those tools.
+tools. As a summary from that document, Terrajet-generated CRDs will not change
+but the controller implementation approach may differ once the vendors step in
+and we'll provide the necessary tooling to make this as smooth as possible.
 
-With that in mind, it's a hard choice to invest in separate Jet-based providers
-knowing that they will, even though after years, be deprecated at one point and
-users will have to migrate. At the same time, classic providers are very far
-from having the same coverage level. This creates confusion and reluctance to
-use either of the providers, which also results in decreased pace of
-contributions because of the divergence of efforts.
+With that in mind, it's a hard choice to invest in standalone Jet-based
+providers knowing that they will, probably after years, be deprecated at one
+point and users will have to migrate. At the same time, classic providers are
+very far from having the same coverage level. This creates confusion and
+reluctance to use either of the providers, which also results in decreased pace
+of contributions because of the divergence of efforts.
 
 With a single mixed provider, the story is much clearer; we fill the gaps with
 Terrajet and over time the implementation of the resources will change to be
@@ -210,7 +212,12 @@ have to consider. Overall, the decision of how to go from Terrajet to
 cloud-vendor tooling is an independent decision from how we manage Terrajet and
 classic provider mix since it's the step after we decide that, if ever, the
 provided tooling is better than we have. For example, today, calling AWS API
-directly is definitely not better than Terrajet.
+directly in a controller is less resource intensive compared to Terraform CLI
+calls, however, the process of adding and maintaining those controllers simply
+does not scale to hundreds of CRDs in a meaningful timeframe so we choose
+Terrajet over that. It's the trade-offs like these that will be specific to the
+tools they will come up with that will allow the maintainers to make the best
+choice.
 
 In all cases, our best bet on getting them to maintain their provider is to
 increase usage so that their customers demand it and the single mixed provider
@@ -235,7 +242,7 @@ decreased usage or Jet one which will be deprecated at one point, even if years
 away. If there is a single mixed provider, then everyone would know the path
 forward and be assured that the next step for that provider is only additive
 towards the end goals; not deprecation or removal of their contributions in
-favor of a temporary approach.
+favor of an approach that is not the end goal.
 
 ### Cross-resource References
 
@@ -347,17 +354,17 @@ and DCL. Azure seems to have a single provider that is built with ARM. The
 common theme of all these TF providers is that the users have a linear choice to
 make; it's either a single provider or they know which provider is going to be
 maintained in the foreseeable future. In our case, we have separate Jet and
-classic providers, but it's hard to convince people that Terrajet is a temporary
-solution but also Jet providers will be maintained long enough to outlive the
-project they are currently building, hence the confusion.
+classic providers, but it's hard to convince people that standalone Jet
+providers will be maintained long enough to outlive the project they are
+currently building, hence the confusion.
 
 For Pulumi, the story is a bit different. They had started with providers
 generated from TF equivalents, a similar approach to Terrajet. But then they
 [bootstrapped][pulumi-blog] new providers that use cloud vendor tooling
 directly, without Terraform. At all points, their users had a clear idea of
 what's next and the choice is usually clear enough. In our case, we started with
-cloud vendor tooling and then introduced the temporary solution and that make
-people hesitant to choose either.
+cloud vendor tooling and then introduced the TF solution and that make people
+hesitant to choose either.
 
 Even though there are similarities, our trajectory of starting with cloud vendor
 tooling and introducing a provider with more coverage using a middle-layer, i.e.
@@ -374,6 +381,8 @@ point to either of them as similar cases.
 [jet-gcp]: https://github.com/crossplane-contrib/provider-jet-gcp/
 [jet-azure]: https://github.com/crossplane-contrib/provider-jet-azure/
 [provider-aws]: https://github.com/crossplane/provider-aws/
+[provider-gcp]: https://github.com/crossplane/provider-gcp/
+[provider-azure]: https://github.com/crossplane/provider-azure/
 [gcp-tf-contributing]:
     https://github.com/hashicorp/terraform-provider-google/blob/610f62b/.github/CONTRIBUTING.md#generated-resources
 [classic-setup]:
--- a/design/one-pager-merge-jets-proposal.md
+++ b/design/one-pager-merge-jets-proposal.md
@@ -116,26 +116,58 @@ With that in mind, it's a hard choice to invest in separate Jet-based providers
 knowing that they will, even though after years, be deprecated at one point and
 users will have to migrate. At the same time, classic providers are very far
 from having the same coverage level. This creates confusion and reluctance to
-use and contribute to either of the providers, which also results in decreased
-pace of contributions because of the divergence of efforts.
+use either of the providers, which also results in decreased pace of
+contributions because of the divergence of efforts.
 
 With a single mixed provider, the story is much clearer; we fill the gaps with
 Terrajet and over time the implementation of the resources will change to be
-more _native_; be it cloud vendor generation tools or handwritten API calls. As
-long as the CRD schema and the API behavior is same, they can make the best
-choice they'd like. In case they decide to start their own provider, then the
-story is still clear; there is a new provider maintained by its vendor and you
-should migrate to that one once you see fit as opposed to Jet vs classic choice
-where you need to think about different trade-offs with different timelines.
+more _native_; be it cloud vendors using their own approaches or handwritten API
+calls community writes that may be necessary to overcome a bug in TF provider.
+As long as the CRD schema and the API behavior is same, the maintainers can make
+the best choice depending on the situation, communicate any breaking changes and
+provide a smooth experience to the users. In case cloud vendor steps in and
+decides to start its own provider, then the story is still clear; there is a new
+provider maintained by its vendor and you should migrate to that one once you
+see fit as opposed to Jet vs classic choice where you need to think about
+different trade-offs with different and uncertain timelines.
+
+### Ownership by The Cloud Vendors
+
+As stated above, we hope that cloud vendors will come in and maintain their
+Crossplane providers. It seems like Terraform has achieved this goal only for
+GCP; Azure and AWS Terraform providers are not maintained by the vendors. In our
+case, Crossplane is a CNCF project with an open governance model, so we can't
+really compare it with Terraform, a property of HashiCorp, in that respect but
+it still provides a data point.
+
+The strategy we choose should take both possibilities into account:
+* They may choose to bootstrap their own Crossplane provider, maybe even not
+under Crossplane organization.
+* They might decide to convert the underlying implementation to their own
+technology, similar to GCP Terraform provider that is [undergoing a
+process][gcp-tf-contributing] where they change implementation from handwritten
+API calls to Magic Modules and DCL.
+
+In the first case, regardless of whether we go with separate or mixed providers,
+we will likely suggest users to migrate to the vendor-maintained provider so
+that they can get commercial grade support for the bugs in the provider. One
+could argue that if users are on a single provider once that happens, it's
+probably easier to handle the migration.
+
+In the second case, mixed provider wouldn't require any manual migration. The
+conversion webhooks that Kubernetes API provides will take care of the API
+changes as described in [provider strategy doc][strategy-doc] in detail. In the
+case of separate providers though, users of one of the provider would have to
+migrate.
 
 ### Convergence of Efforts
 
 The classic providers have the years of usage experience and contributions from
 the community that makes them mature enough to be used in production. While Jet
 providers are working on top of the mature TF providers that have been
-maintained longer than Crossplane's existence in some cases, they will still need
-contributions from the community to get to a certain level of maturity. If there
-is a single provider, then the community wouldn't have to deal with two
+maintained longer than Crossplane's existence in some cases, they will still
+need contributions from the community to get to a certain level of maturity. If
+there is a single provider, then the community wouldn't have to deal with two
 implementations of the same API, trying to cover ground on both. The bugs would
 be fixed once and for everyone.
 
@@ -223,14 +255,18 @@ work. So, even though the structure is same, you'll see that [the generic
 controller][generic-controller] makes extensive use of the functions defined in
 that interface where as classic implementations work with the type directly.
 
-### Code Debugging Experience Difference
+### Maintenance Challenges
 
 This is probably the biggest difference between classic and Jet providers. If
 you run the debugger and put breakpoints in the code, you'll realize that Jet
 providers run the Terraform CLI in a temporary TF workspace folder instead of
 making HTTP calls to the cloud vendor. This makes it more challenging to debug
 Jet providers because you need to have a rough idea of how Terraform works, i.e.
-concepts like TF state and TF configuration.
+concepts like TF state and TF configuration. In either case though, if you use a
+Jet-generated CRD in Jet provider or mixed provider, you'll need to bear, this
+cost. Mixing the provider doesn't really make that harder or easier but what to
+expect when you start debugging could be different, i.e. not every CRD is
+debugged in the same way.
 
 One thing to keep in mind is that it's not really a common case for Crossplane
 users to locally debug the code to see what's happening - usually the error in
@@ -246,6 +282,8 @@ much change to the errors it gets from the cloud vendor API.
 [jet-aws]: https://github.com/crossplane-contrib/provider-jet-aws
 [jet-gcp]: https://github.com/crossplane-contrib/provider-jet-gcp/
 [jet-azure]: https://github.com/crossplane-contrib/provider-jet-azure/
+[gcp-tf-contributing]:
+    https://github.com/hashicorp/terraform-provider-google/blob/610f62b/.github/CONTRIBUTING.md#generated-resources
 [classic-setup]:
     https://github.com/crossplane/provider-aws/blob/475fd24/pkg/controller/acm/controller.go#L62
 [jet-setup]:
--- a/design/one-pager-merge-jets-proposal.md
+++ b/design/one-pager-merge-jets-proposal.md
@@ -275,6 +275,39 @@ the case for Jet providers, in fact, most of the errors you see still look
 similar to classic providers since Terraform, just like Crossplane, doesn't make
 much change to the errors it gets from the cloud vendor API.
 
+## What Others Did
+
+The concept of providers are common with projects like Terraform and Pulumi.
+These projects have had similar challenges as well regarding the technology
+choices that they made to build their providers.
+
+In Terraform, it seems that the maintainers (not affiliated with AWS) has
+decided to bootstrap a new provider when they decided to move to AWS Cloud
+Control whereas GCP Provider (maintained by GCP) had [decided to
+mix][gcp-tf-contributing] the manually written TF code with their Magic Modules
+and DCL. Azure seems to have a single provider that is built with ARM. The
+common theme of all these TF providers is that the users have a linear choice to
+make; it's either a single provider or they know which provider is going to be
+maintained in the foreseeable future. In our case, we have separate Jet and
+classic providers, but it's hard to convince people that Terrajet is a temporary
+solution but also Jet providers will be maintained long enough to outlive the
+project they are currently building, hence the confusion.
+
+For Pulumi, the story is a bit different. They had started with providers
+generated from TF equivalents, a similar approach to Terrajet. But then they
+[bootstrapped][pulumi-blog] new providers that use cloud vendor tooling
+directly, without Terraform. At all points, their users had a clear idea of
+what's next and the choice is usually clear enough. In our case, we started with
+cloud vendor tooling and then introduced the temporary solution and that make
+people hesitant to choose either.
+
+Even though there are similarities, our trajectory of starting with cloud vendor
+tooling and introducing a provider with more coverage using a middle-layer, i.e.
+Terrajet, is unique and makes a big difference in how people should think about
+the choice they make. So, we have one of the three in Terraform going with mixed
+approach and none in Pulumi, but the trajectories don't really map enough to
+point to either of them as similar cases.
+
 [Terrajet]: https://github.com/crossplane/terrajet
 [deep-dive-blogs]: https://blog.crossplane.io/deep-dive-terrajet-part-i/
 [xrm-doc]: one-pager-managed-resource-api-design.md
@@ -307,3 +340,4 @@ much change to the errors it gets from the cloud vendor API.
     https://doc.crds.dev/github.com/crossplane/provider-gcp@v0.20.0
 [jet-gcp-crds]:
     https://doc.crds.dev/github.com/crossplane-contrib/provider-jet-gcp@v0.2.0-preview
+[pulumi-blog]: https://www.pulumi.com/blog/announcing-aws-native/
--- a/design/one-pager-merge-jets-proposal.md
+++ b/design/one-pager-merge-jets-proposal.md
@@ -1,8 +1,8 @@
 # Merging AWS, Azure and GCP Providers
 
 * Owner: Muvaffak Onus (@muvaf)
-* Reviewers: Crossplane Maintainers
-* Status: Draft
+* Reviewers: Crossplane Maintainers and The Steering Committee
+* Status: Accepted
 
 ## Background
 
@@ -141,49 +141,53 @@ The generic Terrajet controller calls Terraform CLI which in turn calls its
 Terraform provider to operate. Each CRD has its own completely independent
 reconciler thread configured to work with the schema of that resource.
 
-This dependence causes the Crossplane provider to be exposed to the bugs of the
-underlying Terraform provider. While we strive to fix the bug there first, it's
-not always feasible for us to land that change in a timely manner. In addition,
-there are cases where we are locked to a version with a bug where downgrading or
-upgrading is not an option because of another bug or behavior change.
+This dependence allows us to leverage the years' of work in these mature
+Terraform providers but also causes the Crossplane provider to be exposed to
+their bugs. While we strive to fix the bug there first, it's not always feasible
+for us to land that change in a timely manner. In addition, there are cases
+where we are locked to a version with a bug where downgrading or upgrading is
+not an option because of another bug or behavior change.
 
-With the mixed provider, we can change the underlying implementation to a
-hand-written controller whenever we face such a bug. With the Kubernetes API
-conversion tooling, we can manage the possible API changes as well. This would
-also mean that over time, we'd have less and less Terraform dependence in our
-ecosystem.
+With the mixed provider approach, we can change the underlying implementation to
+a hand-written controller whenever we face such a bug. With the Kubernetes API
+conversion tooling, we can manage the possible API changes as well.
 
 ### Confusion for Users and Maintainers
 
 Terraform is a great tool but in an ideal state, we'd like the cloud vendors to
 step in and maintain those providers. They may choose to keep using their TF
-provider but it's unlikely; AWS does not maintain their TF provider, Azure uses
-ARM in their TF provider and they've shown interest in generating Crossplane
-based on ARM and Google is using Magic Modules and DCL to generate TF provider,
-so it's likely that they will choose to do that for their Crossplane provider as
-well. [Provider Strategy doc][strategy-doc] goes into more detail about these
-tools. As a summary from that document, Terrajet-generated CRDs will not change
-but the controller implementation approach may differ once the vendors step in
-and we'll provide the necessary tooling to make this as smooth as possible.
+provider but it's unlikely;
+* AWS and Azure do not maintain their TF provider.
+* Azure uses ARM in their TF provider and they've shown interest in generating
+Crossplane based on ARM.
+* Google is using Magic Modules and DCL to generate TF provider. It's likely
+  that they will choose to do that for their Crossplane provider as well.
+
+[Provider Strategy doc][strategy-doc] goes into more detail about these tools.
+As a summary from that document, Terrajet-generated CRDs will not change but the
+controller implementation approach may differ once the vendors step in and we'll
+provide the necessary tooling to make this as smooth as possible.
 
 With that in mind, it's a hard choice to invest in standalone Jet-based
-providers knowing that they will, probably after years, be deprecated at one
-point and users will have to migrate. At the same time, classic providers are
-very far from having the same coverage level. This creates confusion and
+providers knowing that they will, even if it might take years, be deprecated at
+one point and users will have to migrate. At the same time, classic providers
+are very far from having the same coverage level. This creates confusion and
 reluctance to use either of the providers, which also results in decreased pace
 of contributions because of the divergence of efforts.
 
 With a single mixed provider, the story is much clearer; we fill the gaps with
-Terrajet and over time the implementation of the resources will change to be
-more _native_; be it cloud vendors using their own approaches or handwritten API
-calls community writes that may be necessary to overcome a bug in TF provider.
-As long as the CRD schema and the API behavior is same, the maintainers can make
-the best choice depending on the situation, communicate any breaking changes and
-provide a smooth experience to the users. In case cloud vendor steps in and
-decides to start its own provider, then the story is still clear; there is a new
-provider maintained by its vendor and you should migrate to that one once you
-see fit as opposed to Jet vs classic choice where you need to think about
-different trade-offs with different and uncertain timelines.
+Terrajet and over time the implementation of all of the resources will change to
+Terrajet. Once vendors come in, they can start using their own tooling. There
+are two possibilities:
+
+* Being maintainers in these providers.
+  * As long as the CRD schema and the API behavior is same, the maintainers can
+    make the best choice of tooling, communicate any breaking changes and
+    provide a smooth experience to the users.
+* Bootstrapping new providers.
+  * Users should migrate to that one once you see fit as opposed to Jet vs
+    classic choice where you need to think about different trade-offs with
+    different and uncertain timelines.
 
 ### Ownership by The Cloud Vendors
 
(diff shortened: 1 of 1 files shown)


design/one-pager-merge-jets-proposal.md
```diff
@@ -0,0 +1,370 @@
...
+different trade-offs with different and uncertain timelines.
+
+### Ownership by The Cloud Vendors
+
+As stated above, we hope that cloud vendors will come in and maintain their
+Crossplane providers. It seems like Terraform has achieved this goal only for
+GCP; Azure and AWS Terraform providers are not maintained by the vendors. In our
+case, Crossplane is a CNCF project with an open governance model, so we can't
+really compare it with Terraform, a property of HashiCorp, in that respect but
+it still provides a data point.
```
negz, line 147: Are there other data-points apart from Terraform providers we could look to? Perhaps ACK, ASO, KCC, etc?

design/one-pager-merge-jets-proposal.md
```diff
@@ -0,0 +1,370 @@
...
+opinions about providers that they have not used yet. Today we have more data
+and more awareness in the community, and having two different implementations
+has been confusing for a lot of users, so a revisit of the decision regarding
+AWS, Azure and GCP is due.
+
+## Goals
+
+We have the following goals for _the big three_ providers:
+
+* The cloud vendors should maintain their Crossplane provider.
```
negz, line 38: A note for readers: those of us on the Crossplane Steering Committee are working on a charter for the project. We've talked about wanting vendors to own their own providers for a while but never really written it down. Expect it to be codified in the charter.

design/one-pager-merge-jets-proposal.md
```diff
@@ -0,0 +1,370 @@
...
+under Crossplane organization.
+* They might decide to convert the underlying implementation to their own
+technology, similar to GCP Terraform provider that is [undergoing a
+process][gcp-tf-contributing] where they change implementation from handwritten
+API calls to Magic Modules and DCL.
+* They would provide the tools but not take part in maintenance of the codebase.
+
+In the first case, regardless of whether we go with separate or mixed providers,
+we will likely suggest users to migrate to the vendor-maintained provider so
+that they can get commercial grade support for the bugs in the provider. One
```
negz, line 160: Who is giving commercial grade support here? The vendor? I'm not sure we know enough to presume that will happen.

design/one-pager-merge-jets-proposal.md
```diff
@@ -0,0 +1,370 @@
...
+In the third case, if we go with single mixed provider, we can choose to use the
+cloud vendor tooling for the resources we decide to convert. If we go with
+separate providers, we could keep working with those tools, like AWS CC, in the
+existing classic provider or bootstrap a new one but either case would either be
+similar or worse than the single mixed provider in terms of choices users would
+have to consider. Overall, the decision of how to go from Terrajet to
+cloud-vendor tooling is an independent decision from how we manage Terrajet and
+classic provider mix since it's the step after we decide that, if ever, the
+provided tooling is better than we have. For example, today, calling AWS API
+directly is definitely not better than Terrajet.
```
negz, line 182: > For example, today, calling AWS API directly is definitely not better than Terrajet.

Is this that much of a foregone conclusion? Is there something readers can reference to demonstrate it? I would presume that like anything there are trade-offs here. Off the top of my head:

* Provider maintainers can move faster without a dependency on Terraform (an upstream project).
* Invoking Terraform binaries is presumably more resource intensive than making an SDK call.

design/one-pager-merge-jets-proposal.md
```diff
@@ -0,0 +1,370 @@
...
+
+In all cases, our best bet on getting them to maintain their provider is to
+increase usage so that their customers demand it and the single mixed provider
+makes people more comfortable by reducing the decision overhead and providing
+confidence to users about what they can expect to happen to the tool they build
+their infrastructure on top of.
+
+### Convergence of Efforts
```
negz, line 190: Honestly I don't think mixing providers alone makes a big (or really any) difference here. Contributors still have to account for N different implementations (TJ, classic, etc) when adding a feature or functionality that touches many MRs, those N different implementations just all live in one repo instead of several.

design/one-pager-merge-jets-proposal.md
```diff
@@ -0,0 +1,370 @@
...
+  matchLabels:
+    app: ola
+```
+
+While this referencing style is useful for many scenarios and being discussed
+[here][generic-references], it's not the same as how Kubernetes resources
+reference each other and requires more upfront information. We could possibly
+make it work like native with a default and let users optionally give different
+`kind`, `apiVersion` and `fieldPath`, there are other concerns with the approach
+listed in its discussion.
```
negz, line 257: I have a hunch that we'll want generic references eventually, and I'd prefer for us to align on one way to handle references (i.e. deprecate our current way of doing cross resource refs once we figure out something better). That said, I do appreciate that what this document proposes allows us to kick that can down the road a bit instead of blocking on solving it now.

negz:
@zzxwill I'm curious to get your opinion on this document too, as someone who has maintained `provider-alibaba` and also recently worked on building `provider-jet-alibaba`.

design/one-pager-merge-jets-proposal.md
```diff
@@ -0,0 +1,370 @@
...
+knowing that they will, even though after years, be deprecated at one point and
+users will have to migrate. At the same time, classic providers are very far
+from having the same coverage level. This creates confusion and reluctance to
+use either of the providers, which also results in decreased pace of
+contributions because of the divergence of efforts.
+
+With a single mixed provider, the story is much clearer; we fill the gaps with
+Terrajet and over time the implementation of the resources will change to be
+more _native_; be it cloud vendors using their own approaches or handwritten API
+calls community writes that may be necessary to overcome a bug in TF provider.
```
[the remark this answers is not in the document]
muvaf: > I'm betting option 2 will waste hundreds of contributor hours building MRs a vendor will eventually throw away.

Could you elaborate on this? I don't really set a goal to converge on classic; either vendors will convert them to whatever they'd like or we will convert but only when we **have to** due to a bug - no explicit goal of getting to full classic. So, we'll add new CRDs to the classic providers via Terrajet and that's pretty much it. We'll do any sort of conversion only if it's deemed necessary by its maintainers; which could be the vendor or community maintainers. No one will build new native MRs, Terrajet will be the first choice.

If I understood correctly, your preference of option 3 suggests that we set a goal to convert SDK implementations to Jet. Assuming the conversion webhooks are in place, hence we're able to make breaking API changes, this would mean replacing the controllers of 127 CRDs in provider-aws. The costs of doing this in my mind are:
* Risk. Each replacement of controller, be it in either way, carries some amount of risk of bugs. Even though you get the API conversion right, controller behavior may not exactly match. We'd have to be careful about this risk for especially v1beta1 resources.
  * The proposal suggests that we do this only if we have to due to a bug or vendor decides to do so.
* Effort. We can use Terrajet to generate these CRDs right away but the schema conversion functions will have to be handwritten and we need to test each of them to make sure existing users are not affected. Especially v1beta1 ones will make the testing more cumbersome.
  * The proposal suggests to do that only if we can't fix a bug in a Jet resource and its direction would be Jet -> classic, so we're going from less transparency to full control which means we have more tooling available to make the controller behave exactly like its Jet variant.
* Community backlash. These resource have been in place since the existence of Crossplane and many people in the community contributed to them in one way or another. Replacing the controller code with a temporary solution which depends on Terraform may not be well-received.

> As we've identified, each of the big three vendors seems to have their own preferred way to build providers (e.g. AWS Cloud Control API, Azure ARM, GCP DCL).

I agree that when the vendors step in to maintain their providers, they'll try to make the best decision to make it more maintainable for themselves. However, this doesn't directly mean that it will result in a different provider. For example, GCP DCL does not have full coverage, so they use it only for a portion of the provider. As a counter example, TF folks decided to start another provider for CC.

IMO, the reason they will step in will be the popularity of the community-maintained providers and they would possibly decide to be maintainers on those repos and change the implementation to their tools by using conversion mechanisms we have in place.

In my opinion, the best we can do is to strive to make such replacements as less risky as possible so that they take part in the existing provider. If they decide to bootstrap another one, like CC, then that's also perfectly fine because in the meantime, the community will have coverage **and** maturity option already and they can decide to move to that provider whenever they'd like to, which is similar to how [AWS CC explains](https://github.com/hashicorp/terraform-provider-awscc/blob/main/contributing/FAQ.md#which-provider-should-i-use) their two provider approach.

design/one-pager-merge-jets-proposal.md
muvaf: I am assuming a level of support similar to their SDKs. Do you think no such support exists?

design/one-pager-merge-jets-proposal.md
muvaf: My point about in this section is actually more like usage divergence which would result contribution divergence. For example, contributor A fixes an EKS bug in Jet provider but classic provider users don't get that fix until someone fixes it there, too. Or that bug doesn't exist in classic at all, so if contributor A were using classic, then they'd not have to spend that effort in the first place. It's essentially about the upsides of having a common single controller code for a single API resource.

> Contributors still have to account for N different implementations (TJ, classic, etc) when adding a feature or functionality that touches many MRs, those N different implementations just all live in one repo instead of several.

That is true but I think that makes up a really low amount of the contributions to providers and we had tried to make Jet ones as similar as possible to the classic ones to reduce the cost of difference. In my mind, I think the API-specific features/fixes are more critical to think about in that respect.

design/one-pager-merge-jets-proposal.md
negz: I would not assume it does - it may be worth looking to open source projects like ASO and ACK to determine whether vendors deem those to be "commercially supported", which to me implies it's possible to pay for support for those tools.

design/one-pager-merge-jets-proposal.md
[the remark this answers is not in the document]
muvaf: @negz I just pushed an update adding the tangible steps we'll take to merge the providers and also the guidelines to follow when you add or convert a controller from one approach to another. Let me know if you'd like to see more detailed guidelines or want to change them.

> I agree with these costs, but I think they are worth paying to provide users and contributors a more consistent experience going forward.

Minding the costs, I still don't really think we should set a goal of converting everything to Terrajet but tried to list a few cases where you'd like to do that and it's totally fine, i.e. the golden standard is not native, it's whatever best for that specific resource. Though I believe we will end up with more and more converted-to-Jet controllers in the long run, especially with the APIs that are really cumbersome to manage like DynamoDB or CloudFront Distribution; but not much for EKS cluster or VPC.

unixdaddy:
The performance in jet with all the CRDs has been really quite difficult. How will this be addressed?

design/one-pager-merge-jets-proposal.md
zzxwill: So Crossplane Steering Committee or a provider generator will maintain the code base, and some mechanism will be responsible for patches from the new release of a Terraform Provider. Did I get the point?

design/one-pager-merge-jets-proposal.md
```diff
@@ -0,0 +1,402 @@
...
+* Option A: Completely separate providers.
+* Option B: One provider but Jet-based resources in their own API group.
+* Option C: One provider and no Jet vs classic difference. Only API version
+  would be bumped if the implementation changes.
+
+We should go with the Option C, i.e. use Terrajet in
+[`provider-aws`][provider-aws], provider-azure and provider-gcp as another way
+of generating CRDs and controllers so that they can reach the goal of coverage
+really fast. Then the community can keep investing in that single provider and
+make each CRD mature enough to depend on in production. In this way, we can
```
zzxwill, line 72: Can `a consolidated way` be specified a little bit? I learned that the Alibaba Terraform provider team will focus on unit-test coverage. How will Terrajet increase the coverage?
Besides, the maturity might depend on some regression test in some way. How will all the e2e tests be composed for hundreds, even thousands of CRDs?

design/one-pager-merge-jets-proposal.md
```diff
@@ -0,0 +1,402 @@
...
+[`provider-aws`][provider-aws] as an example:
+1. Add Terrajet pipeline to the provider repository.
+1. Copy the manual resource configurations from provider-jet-aws and introduce
+  the CRDs of those configurations in their new API group, without changing
+  their version, i.e. `v1alpha2`.
+1. Generate the full set of CRDs in `release-x.x-preview` branch similar to jet
+  repository and include those non-configured ones only in `-preview` suffixed
+  version.
+1. Write up a guide or tool to help the current users of non-preview version of
+  Jet providers to migrate to [`provider-aws`][provider-aws].
```
zzxwill, line 97: The idea of migration is brilliant.
Besides, a smooth upgrade of the provider is essential for production users. Some of KubeVela uses classic Crossplane Alibaba Cloud CRDs in the business applications.

design/one-pager-merge-jets-proposal.md
```diff
@@ -0,0 +1,402 @@
...
+underlying Terraform provider. While we strive to fix the bug there first, it's
+not always feasible for us to land that change in a timely manner. In addition,
+there are cases where we are locked to a version with a bug where downgrading or
+upgrading is not an option because of another bug or behavior change.
+
+With the mixed provider, we can change the underlying implementation to a
+hand-written controller whenever we face such a bug. With the Kubernetes API
+conversion tooling, we can manage the possible API changes as well. This would
+also mean that over time, we'd have less and less Terraform dependence in our
+ecosystem.
```
zzxwill, line 137: Does a hand-written controller mean calling API/SDK to manage a cloud resource or its operation?

design/one-pager-merge-jets-proposal.md
```diff
@@ -0,0 +1,370 @@
...
+Terraform is a great tool but in an ideal state, we'd like the cloud vendors to
+step in and maintain those providers. They may choose to keep using their TF
+provider but it's unlikely; AWS does not maintain their TF provider, Azure uses
+ARM in their TF provider and they've shown interest in generating Crossplane
+based on ARM and Google is using Magic Modules and DCL to generate TF provider,
+so it's likely that they will choose to do that for their Crossplane provider as
+well. [Provider Strategy doc][strategy-doc] goes into more detail about these
+tools. In summary, Terrajet will be a temporary solution that we put in place to
+get more coverage and at some point we'll need to discuss how to change the
+implementation to those tools.
```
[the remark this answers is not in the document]
zzxwill: SDKs for various languages and the Terraform provider of Alibaba Cloud is also auto-generated by an API-level orchestration machine.

design/one-pager-merge-jets-proposal.md
```diff
@@ -0,0 +1,402 @@
...
+maintained by their cloud vendor. The aim of this document is to capture a
+decision about the future of these provider implementations with the goals
+stated above in mind.
+
+## Proposal
+
+There were three options in the provider strategy doc:
+* Option A: Completely separate providers.
+* Option B: One provider but Jet-based resources in their own API group.
+* Option C: One provider and no Jet vs classic difference. Only API version
```
jbw976, line 65: it's not immediately clear to me what happens in option C here when there are conflicting resources between classic and jet.  i.e. if provider-aws already has `foo.aws.crossplane.io/v1beta1` and jet has `foo.jet.aws.crossplane.io/v1alpha2`, what happens when they get merged? is there only 1 that wins since they are merged into the same provider and api group?

maybe this is describe later in the doc, but it's the very first question that came to my mind and should perhaps be addressed more up front.

danielinclouds:
I think the performance with Jet GCP provider issues should be addressed before this is merged. Pure GCP provider works like a charm so combining it with Jet GCP would ruin that experience.

design/one-pager-merge-jets-proposal.md
andrzej-natzka: This is even worse in Azure, mainly AKS implementation, which in classic is very basic and missing many features. If the deciding criteria will be: use jet if classic resource does not exist functionality will be degraded. This implies that for merging via option C each resource has to be crossed check about functionality and this will require substantial work.

muvaf: @jbw976 I've added a list of tangible steps about how it's going to be, let me know if that doesn't cover your concern. The idea is that whatever exists already will stay as is and new CRDs will be added with Terrajet without the `jet` grouping, just like adding another CRD using ACK.

@andrzej-natzka Initially, no such conversion will take place. But there are cases where we might want to do Jet->classic or classic->Jet and I dropped a guideline about those. @negz has suggested we do classic->Jet for all existing ones at some point by setting a goal [here](https://github.com/crossplane/crossplane/pull/2930#discussion_r817263352), you can follow that discussion there and add your thoughts.

design/one-pager-merge-jets-proposal.md
muvaf: For AWS, it specifically means calling API/SDK directly and without using ACK codegen pipeline. For GCP and Azure, there is no such pipeline anyway, so it's just classic implementation.

design/one-pager-merge-jets-proposal.md
muvaf: The existing CRDs in classic providers will not be affected at all, we won't change their controller or schema with this plan today. The migration is mostly about users of Jet providers because once the merger is completed, those repositories will be archived so they will need to move, which will involve only API group change for most resources but also some schema changes if the CRD already existed in the classic provider (we don't replace the existing CRDs/controllers).

design/one-pager-merge-jets-proposal.md
muvaf: The consolidation here refers to reducing number of providers for a cloud to 1 and note that the whole proposal is only about AWS, Azure and GCP. Alibaba provider maintainers or others can follow the track but they definitely don't have to.

> I learned that the Alibaba Terraform provider team will focus on unit-test coverage. How will Terrajet increase the coverage?

From that perspective, you can think of it like whatever investment you make on your Terraform provider will benefit you in Crossplane provider as well, including e2e and unit tests. For the providers mentioned here, the best we have today is to make sure examples work before we introduce the CRD and test against that if there is a change to that controller. I don't really expect much change in that regarding what tool is used to generate the CRD/controller.

design/one-pager-merge-jets-proposal.md
muvaf: I have updated this section to point to some of the trade-offs. Do you think having an explicit comparison would help?

design/one-pager-merge-jets-proposal.md
muvaf: The goal is for vendor to maintain it in a way that works best for them, which could be under `crossplane` or `crossplane-contrib` or their own organization in Github.

design/one-pager-merge-jets-proposal.md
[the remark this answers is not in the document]
muvaf: I have reworded those sections to remove `temporary` and make sure it conveys that it's just not guaranteed to be the end goal, which is cloud vendor tooling.

design/one-pager-merge-jets-proposal.md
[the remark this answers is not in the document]
Feggah: I got a little confused with this discussion. Before this proposal, I had in my mind that hand-written MR implementation was preferred over TJ-generated ones. And TJ was only serving as a fast way to increase coverage for providers, but in the future, we should strive to have all MRs to be hand-written.

Now I'm getting a little confused, because if the true goal is to have vendors to maintain their own providers (and they can simply throw away the community-one to create another implementation for themselves), contributors would end up "wasting" a lot of time implementing hand-written MRs/transforming TJ to classic. And this wouldn't happen with TJ-generated implementation, because it is significantly faster/effortless then the hand-written option.

So I think it's important to set a goal here, just to be clear for contributors what should we do until vendors owns the maintainership of their providers. 

When we need to add a new resource, its clear to me what we have to do when deciding about TJ/classic.

But how do we should **preferably** iterate on these resources? Trying to be more clear in my question, if we have all providers with 100% coverage and in a stable version in the future, what should this provider looks like, 99% TJ or 99% classic? Which is the best/preferable implementation for MRs?

design/one-pager-merge-jets-proposal.md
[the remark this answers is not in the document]
luebken: We are talking exclusively about AWS, Azure and GCP here. I would like to point out that a Terrajet created provider is valid long-term strategy for many other use-cases. Especially in cases where the Terraform provider is well maintained.

blut:
I think the focus should be the value this adds for the user. 
Sure, maintainability must be possible, but shouldn't the Jet provider be generated and therefore (almost) self-maintained? 

Using a generated provider was meant as a work around to increase coverage momentarily, which it does, but merging a generated provider with the native equivalent does not increase the value. 

If the case exists, where one immediately requires resources not implemented in the native provider, it is still possible to run both providers on the same controlplane. Compositions can reference either resource, therefore merging does not increase usability. 

Instead, a merge brings numerous new disadvantages 
- Merged providers bring the performance issues of the Jet provider to the resulting version 
- Arbitration on the prevailing version 
- Providers maintenance becomes more complex and any new native implementation would be expected to implement the same behaviour as the previous Jet equivalent 
- Relying on Terraform (might) results in slower release cycles (e.g. Bug Fixes in Terraform library must still be propagated to a new provider version) 

Even if we were to plan for providers to be maintained by the API owner, I'd expect them to use native code, especially since equivalent Terraform providers are not maintained by the Cloud Providers. 
 
From a user point of view, merging the two providers does not provide any value over the current setup.

design/one-pager-merge-jets-proposal.md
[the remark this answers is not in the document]
negz: > I would like to point out that a Terrajet created provider is valid long-term strategy for many other use-cases. Especially in cases where the Terraform provider is well maintained.

I think TJ is a valid strategy for basically all cases. Hypothetically if (for example) Google wanted to 'own' their own Crossplane provider to the extent that they owned and maintained a TJ based provider generated from the Terraform provider they already owned, we (the Crossplane maintainers) would be fine with that.

What seems to complicate discussions of TJ based providers is different understandings of what "long term" means. On a sufficiently long timescale I could imagine Crossplane (or something else) becoming so successful that Terraform falls out of favor, at which point Crossplane providers being built from Terraform providers may make less sense. That's likely far enough off that we don't really need to think of what we're doing here as "temporary" in any meaningful way, though.

negz:
> @danielinclouds @unixdaddy I have added a note about CRD scaling issues.

@muvaf et al I believe we should try to decouple this design from the scaling issues. We need to address both - let's use this document to align on what the best path forward is for the "big three" providers under the assumption that we can fix the scaling issues.

negz:
> I think the focus should be the value this adds for the user.
> Sure, maintainability must be possible, but shouldn't the Jet provider be generated and therefore (almost) self-maintained?

@blut our focus is always on the user. Part of that is optimizing the contributor experience to ensure it's as simple as possible for contributors to build a great user experience.

> If the case exists, where one immediately requires resources not implemented in the native provider, it is still possible to run both providers on the same controlplane. Compositions can reference either resource, therefore merging does not increase usability.

There are some gotchas in practice - for example cross resource references only work "within" a provider, so you couldn't ask a Terrajet based resource to reference a non-Terrajet based resource that wasn't part of the same Composition. https://github.com/crossplane/crossplane/issues/1770 could address this but we haven't yet made headway there.

> From a user point of view, merging the two providers does not provide any value over the current setup.

FWIW this was my position until recently and I do see where you're coming from. I still go back and forth myself. We have seen concrete evidence that having two distinct providers for the "big three" confuses things for users though:

* Having two variants inherently raises the question of "which one is better". We see folks asking this all the time.
* We see features (like identity auth for GCP) being added to one provider but not the other.

design/one-pager-merge-jets-proposal.md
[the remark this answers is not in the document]
muvaf: > But how do we should preferably iterate on these resources? Trying to be more clear in my question, if we have all providers with 100% coverage and in a stable version in the future, what should this provider looks like, 99% TJ or 99% classic? Which is the best/preferable implementation for MRs?

I think reducing the whole thing to which one is better does not serve well for all resources. Even though we standardize the cloud provider APIs behind XRM, each (service * cloud) API combination has its own quirks and some of them are handled better with full transparency of what's happening, i.e. classic implementation, and some of them are way better off with Terrajet because we'd repeat what Terraform has done for them anyway. In addition, the conversion cost is non-zero; we have the conversion webhooks for API changes but each conversion needs to be tested manually to see if the behavior matches or TF is doing unexpected things.

In summary, I see that these providers will go closer to 95% TJ over a long period of time but meanwhile it will be a mix and I don't think we should rush to get to that level because of the costs associated are not worth it for every resource. For example, why would you need to convert VPC in AWS to TJ? All features are there, it's beta and there is almost no issue and it's faster to provision than TJ.

In the light of all these, I proposed the following guidelines (added to the proposal) for resource iterations:
* If you encounter a bug that cannot be solved with Terrajet in a meaningful
  timeframe, consider changing its controller to classic hand-written controller
  and bump the API version.
* If an existing classic CRD (hand-written or ACK-generated) has a bug that
  cannot be fixed easily or has a blocker or we continuously get bugs which
  makes it hard to maintain, consider replacing it with Terrajet and if you do
  bump the API version.

@Feggah Does that clear the confusion for you? Would you look for more concrete conditions?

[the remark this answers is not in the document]
Feggah: > Does that clear the confusion for you? Would you look for more concrete conditions?

Makes sense, thank you. I thought TJ wasn't the perfect implementation for MRs, but actually it is. No more doubts 👍🏼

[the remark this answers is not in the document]
negz: > In summary, I see that these providers will go closer to 95% TJ over a long period of time but meanwhile it will be a mix and I don't think we should rush to get to that level because of the costs associated are not worth it for every resource. For example, why would you need to convert VPC in AWS to TJ? All features are there, it's beta and there is almost no issue and it's faster to provision than TJ.

This doesn't sound aggressive enough in my opinion. The "if it's not broke don't fix it" approach you mention with VPCs don't resonate with me because it leaves us in a dual implementation where maintainers and contributors need to understand how to work on a native implementation (with all its complexities) for potentially months just because we don't want to commit the time to port it to the more common (TJ) approach and validate that that approach works well.

I'm okay being in that dual implementation state for a significant number of managed resources (~120 MRs in AWS) for a limited time (say 6 months) as long as there's an explicit goal to align as many managed resources as possible on one implementation. I do not feel good about 20+ managed resources using a 'deprecated' implementation indefinitely.

negz:
@muvaf @jbw976  I wonder if it would be useful to setup a community session to discuss this proposal and see if we can reach alignment? Perhaps a follow-on session after the next community meeting? Or maybe we could make sure to carve out enough time in the community meeting itself? I think we'd need at least 30 mins for this topic.

chlunde:
Both as a maintainer and user, I think community efforts would be more focused with a single provider for AWS. As I don't have much experience with the jet providers I don't have any input on the timing.

jbw976:
>  Or maybe we could make sure to carve out enough time in the community meeting itself? 

@negz I've added this to the agenda for tomorrow community meeting - we'll try to move fast through the first half to get to this topic with sufficient time.

iAnomaly:
Whew, this was a lot to digest. I want to tell my story as a user in an Engineering organization of around 71 (including leadership/management) at a well known mindfulness and meditation app where I have evangelized Crossplane as a replacement for Terraform as we grow and mature from a monolithic service to a poly and maybe eventually micro service paradigm.

My initial impression was that introducing Terrajet was bad in the sense that it introduced choice and confusion that didn't exist before but with the benefit of adding rapid coverage that also didn't exist before. The latter is a critical accelerator for replacing Terraform at any organization with Crossplane because parity gaps become blockers during migration work. @luebken's [response to my concern in Slack was](https://crossplane.slack.com/archives/CEG3T90A1/p1635164840075200?thread_ts=1634916000.048100&cid=CEG3T90A1):
> [@Cameron Boulton](https://crossplane.slack.com/team/U01GCP5MXHR) thanks for the feedback. We are super excited about the upcoming releases. With these we will also clarify of how we see these different providers will be used. But to prevent any kind of confusion: provider-tf-aws will not replace provider-aws. The “native” providers will always stay an option and in many cases be the default. We will continue to invest in them just like the [big release](https://github.com/crossplane/provider-aws/releases/tag/v0.20.0) [@muvaf](https://crossplane.slack.com/team/UK2G8F7EY) has cut last week and we are welcoming and support everyone to continue adding new resources.
We also see the challenge of possibly incompatibilities between these different providers.  And we hope to resolve these with [Generic references](https://github.com/crossplane/crossplane/pull/2385) or in the composition layer. Do you have a specific use-case in mind which we could discuss? 

As I've introduced XRDs for IAM, S3, SQS and now working on Aurora at this point, I'm actually aligning more in more with @negz opinions that he calls **Option 3**. This may seem radical but I actually think it would be most beneficial to **officially deprecate `provider-aws` in favor of `provider-jet-aws` as quickly as possible** and add what's missing to `provider-jet-aws`. From my limited resource experience so far, there are more issues with `provider-aws` around AWS API edge cases on which fields should be optional or absent depending on usage of other fields. https://github.com/crossplane/provider-aws/issues/1225 is one recent example of this. The Terraform community has already solved these edge cases so I think TJ wins not only on coverage but also on distributed knowledge and learning of actual use if that makes sense. Really, the key feature gap I have found in TJ so far is the lack of the Selector/Ref field types which are important when composing n > 1 resources which dependencies within those resources.

On the topic of performance for `provider-jet-aws`'s 763 resources I agree with @negz we should decouple that from this decision entirely and solve it separately. The upstream fixes in Kubernetes and kubectl already look promising.

Sorry to have missed the community meeting for this topic on March 10th but I'll try to make April's if this is still being openly discussed.

design/one-pager-merge-jets-proposal.md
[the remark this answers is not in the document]
muvaf: I am fine with being as aggressive as possible but there is a limit that's set by our capacity and the nature of contributions. For example, I don't think it's practical to set a timeline like 6 months without having people dedicated to this work, which is quite hard since it's less of a concern for users and contributors but more for maintainers which is a very small set.

I changed the guideline to say `consider converting to Terrajet before fixing the bug` and added a sentence to say that we actively try to converge on Terrajet but still dismissing a planned push that will cost us engineer cycles today.

> We will also update the guidelines about how to add a resource roughly with the
following:
> * If you are adding a new resource, use Terrajet pipeline unless there is a
  blocker.
> * If you encounter a bug that cannot be solved with Terrajet in a meaningful
  timeframe, consider converting its controller to classic hand-written
  controller and bump the API version.
> * If an existing classic CRD (hand-written or ACK-generated) has a bug, you
  should consider converting its controller to Terrajet first.

> The end goal for the conversions is to have more homogeneity in the codebase so
we should convert as many managed resources as possible to Terrajet. Today,
Terrajet is the technology that these three Crossplane providers decide as the
best available for most cases, hence we'll converge on it. Tomorrow, it might be
AWS Cloud Control API for provider-aws, ARM for provider-azure etc. so we'll
have a goal of converging to those technologies. But we won't have a big push to
convert everything to Terrajet immediately because depending on the complexity
of the resource, the conversions require extensive testing to make sure the
behavior matches and it carries risk for existing users.

muvaf:
@iAnomaly I think deprecating existing classic providers is definitely a viable option. At the end of the day, this proposal also targets an end result that is +95% Terrajet but the way to get there is different. When I think of a comparison of the process, the following concerns come to my mind:
* No conversion webhooks, so users would have to manually migrate every resource from one provider to the other.
* Maintenance burden of the classic providers we need to keep around for a while since they have beta resources and many existing users.
* The end result would be people using `provider-jet-aws` rather than `provider-aws`.
* The API Group domains cannot conflict with provider-aws, so they'd have to keep the `jet` prefix, i.e. `ec2.jet.aws.crossplane.io` instead of `ec2.aws.crossplane.io`

The main concern for me is how we can't really have a standardized and well-documented tools for smooth migration for users. Because in either case it has to happen so it comes down to what tools we can make use of to make this smoother without having production users conducting risky surgeries. To be fair though, this concern scales with the number of CRDs and users hence it's much less of a problem for provider-azure and provider-gcp for example.

design/one-pager-merge-jets-proposal.md
```diff
@@ -0,0 +1,428 @@
...
+We will also update the guidelines about how to add a resource roughly with the
+following:
+* If you are adding a new resource, use Terrajet pipeline unless there is a
+  blocker.
+* If you encounter a bug that cannot be solved with Terrajet in a meaningful
+  timeframe, consider converting its controller to classic hand-written
+  controller and bump the API version.
+* If an existing classic CRD (hand-written or ACK-generated) has a bug, you
+  should consider converting its controller to Terrajet first.
+* Existing classic CRDs that are on alpha version will be converted to Terrajet
```
Feggah, line 109: I don't quite understand this. If we have a classic CRD that has been well tested by users and we are thinking in bumping it to beta (when do we should bump to beta?), we should convert it to terrajet? In my opinion, that would "throw away" the well tested part in the alpha version. Looks like we are creating a jet resource in beta instead of alpha. Am I missing something?

muvaf: > In my opinion, that would "throw away" the well tested part of the alpha version.

I'm on the fence about this but yes, this is one of the implications of aggressiveness in converging on Terrajet, which is discussed in [this thread](https://github.com/crossplane/crossplane/pull/2930#discussion_r838672541). Even beta ones are suggested to convert after extensive testing. The initial proposal was along the lines of `do not convert to Terrajet unless there is a bug` but now it's more like trying everything we can to push to Terrajet. This one is more on the technical enforcement side.  Do you find this too aggressive?

> Looks like we are creating a jet resource in beta instead of alpha.

I'd imagine something like `v1alpha1 (classic)` -> `v1alpha2 (terrajet)` -> and continue with terrajet.

muvaf: Oh great 😄  Yes it definitely involves less effort to start and also maintain since TF providers are very mature but the conversion does take some effort due to behavior tests. Please feel free to approve or decline the PR as you are a maintainer in one of the affected repositories 🙂

design/one-pager-merge-jets-proposal.md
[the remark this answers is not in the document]
muvaf: @negz I added the following to the guideline as we discussed offline to make it more aggressive.

> Existing classic CRDs that are on alpha version will be converted to Terrajet before bumping to beta unless there is a blocker.

Also updated the rest of the sections to be inline with converging on Terrajet instead of staying mixed.

Feggah (approved):
LGTM!

I agree with merging the providers and the documentation giving the directions that we will follow in the near future is clear enough to me. Good job

I think that we will need to be extra careful when testing and creating migration documentation for the users, which may have resources in production

design/one-pager-merge-jets-proposal.md
```diff
@@ -0,0 +1,432 @@
...
+* There will be a single provider.
+* The single provider will be the classic one and we'll add Terrajet CRDs on
+  top.
+
+The order of execution will be roughly as following, taking
+[`provider-aws`][provider-aws] as an example:
+1. Add Terrajet pipeline to the provider repository.
+1. Copy the manual resource configurations from provider-jet-aws and introduce
+  the CRDs of those configurations in their new API group, without changing
+  their version, i.e. `v1alpha2`.
```
turkenh, line 93: I expect some conflicts due to different resource grouping while merging the two. Should we clarify how to solve such conflicts in this document or discuss case by case basis?

For example, Elasticache Replication Group: [native](https://github.com/crossplane/provider-aws/blob/master/apis/cache/v1beta1/register.go#L28) under **cache** vs [jet](https://github.com/crossplane-contrib/provider-jet-aws/blob/release-0.4-preview/apis/elasticache/v1alpha2/zz_replicationgroup_types.go) under **elasticache** group.

design/one-pager-merge-jets-proposal.md
```diff
@@ -0,0 +1,432 @@
...
+since its implementation lives in `terrajet` repository. In either case, the
+controller implementation satisfies the same
+[`ExternalClient`][externalclient-interface] interface from crossplane-runtime.
+
+The biggest difference in that regard is that the CRD structs in Jet providers
+need to satisfy an additional interface called
+[`Terraformed`][terraformed-interface] in order for the generic controller to
+work. So, even though the structure is same, you'll see that [the generic
+controller][generic-controller] makes extensive use of the functions defined in
+that interface where as classic implementations work with the type directly.
```
turkenh, line 341: I couldn't get why this is a drawback, given that most Terrajet code is auto-generated. I can understand that maintainers should understand how Terrajet works due to this difference but it looks like this is already covered below.

turkenh (approved):
The proposal sounds good to me, just added a couple of nonblocking comments.

Thanks for driving this @muvaf 🙌

design/one-pager-merge-jets-proposal.md
AaronME: Just a brief note:

This decision tree might get more complicated for provider-aws. Original provider-aws CRDs that do not align with ack-generated code are currently kept under a `manualv1alpha1` apiVersion. (It looks like this primarily impacts ec2 and eks). So we would be doing a 3-way evaluation, there, for "manual code," "ack-generated," and "jet-generated."

design/one-pager-merge-jets-proposal.md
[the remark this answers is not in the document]
AaronME: I appreciate the huge win that `jet` offers in getting resources to be managed by Crossplane. I want to call out some of the non-technical consequences of moving to a Jet as the default technology underlying crossplane providers. Namely, some of the gaps we have already encountered using Jet providers. As adoption increases, crossplane should expect to invest more time and effort in documentation, resource behavior incompatibility, resource interoperability issues and error-handling. 

1. Documentation: Not all Terraform behavior is intuitive. A good example of this can be found on the [google_project_iam](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/google_project_iam) resources. The implications of choosing, for example, `google_project_iam_policy` over `google_project_iam_member` need to be surfaced to users of provider-jet-gcp in the provider-jet-gcp (and not Terraform) docs. [Current Provider-Jet-GCP docs](https://doc.crds.dev/github.com/crossplane-contrib/provider-jet-gcp/cloudplatform.gcp.jet.crossplane.io/ProjectIAMPolicy) do not represent these notes. 
1. Resource Behavior: Some Terraform behavior does not fit nicely into Crossplane's Resource Model. See [this issue](https://github.com/crossplane-contrib/provider-jet-gcp/issues/46) where secret data that _can_ be updated by terraform requires a currently missing event from Kubernetes/Crossplane to get the Update pipeline to run. Jet-providers which need to detect secret updates will require machinery to include this in the "observe" step. Successful migration to a fully jet-based system will require building support around issues like this.
1. Resource Interoperability: Terraform supports several [built-in functions](https://www.terraform.io/language/functions) that ease the friction of chaining provider resources. For providers which require data in a [specific format](https://github.com/crossplane-contrib/provider-jet-vault/issues/1), crossplane providers will need to support this as well. This could be mean extending the transformation capabilities of composition or extending the ability to control a resource's inputs in [terrajet](https://github.com/crossplane/terrajet/issues/260).
1. Error-Handling: Errors output from terraform will need to be complete and easily readable. A recent error encountered by one user in my org looked like the following:
```
accessNotConfigured. If you received a 403 error, make sure you have the `roles/resourcemanager.projectCreator` permission: : File name: main.tf.json
```
This error was incomplete. Rendering the resource in terraform and manually running it revealed the following, which included more debugging information:
```
│ Error: error creating project platform-test-project (Platform Test Project): googleapi: Error 403: Cloud Resource Manager API has not been used in project 295269043414 before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/cloudresourcemanager.googleapis.com/overview?project=295269043414 then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry.
│ Details:
│ [
│   {
│     "@type": "type.googleapis.com/google.rpc.Help",
│     "links": [
│       {
│         "description": "Google developers console API activation",
│         "url": "https://console.developers.google.com/apis/api/cloudresourcemanager.googleapis.com/overview?project=295269043414"
│       }
│     ]
│   },
│   {
│     "@type": "type.googleapis.com/google.rpc.ErrorInfo",
│     "domain": "googleapis.com",
│     "metadata": {
│       "consumer": "projects/295269043414",
│       "service": "cloudresourcemanager.googleapis.com"
│     },
│     "reason": "SERVICE_DISABLED"
│   }
│ ]
│ , accessNotConfigured. If you received a 403 error, make sure you have the `roles/resourcemanager.projectCreator` permission
│
│   with google_project.platform_test,
│   on main.tf line 7, in resource "google_project" "platform_test":
│    7: resource "google_project" "platform_test" {
```

Another example recently encountered: providing expired GCP credentials to a provider-jet-gcp can result in this error:
```
{"error":"invalid_grant","error_description":"Invalid JWT Signature."}
```

Currently, the only way to debug an error like that is to search the web for "'Invalid JWT Signature' terraform provider gcp"

This is not meant to be a comprehensive list of the issues that terrajet adoption implies. It is simply offered to provide context on how far-reaching this decision will become. Adopting terrajet as the default implementation offers rapid coverage and alleviates the technical burden of building up the library of crossplane-managed resources. At the same time, it creates pressure on downstream support efforts. As the Crossplane Team builds tooling and community around this approach, decisions forced by terrajet adoption will become long-term support requirements for the project.

AaronME (commented):
Moving to jet as the default implementation for all resources will definitely accelerate provider coverage. But there are downstream consequences to adopting the current set of terraform bugs, owning the interoperability of providers, implementing Crossplane-like behavior on top of terraform, and designing Documentation and Error Message handling around current terraform design decisions which could be difficult to unwind as the project matures.

design/one-pager-merge-jets-proposal.md
```diff
@@ -0,0 +1,432 @@
...
+
+### Maintenance Challenges
+
+This is probably the biggest difference between classic and Jet providers. If
+you run the debugger and put breakpoints in the code, you'll realize that Jet
+providers run the Terraform CLI in a temporary TF workspace folder instead of
+making HTTP calls to the cloud vendor. This makes it more challenging to debug
+Jet providers because you need to have a rough idea of how Terraform works, i.e.
```
snarlysodboxer, line 349: I do get the desire to increase adoption by increasing coverage, but this proposal is **incredibly disappointing and disheartening**. I've evangelized Crossplane at my company and we've been so interested in it and started implementing it because we've all experienced the suckage of automating Terraform and troubleshooting it and fixing it's state management.

I've worked at companies that would vito the continued use of Crossplane if it's going to shell out as that has negative security implications, not to mention it's an ugly smelly pattern, and like you said is a phenomenally awful debugging experience. Another company I've worked at would vito it _just_ because it's aiming for Terraform.

I would rather have to write my own where I need coverage than have to debug Terraform being used by Crossplane.

I can see combining the providers, but I can't believe we're considering moving towards Terraform and away from native. We will be looking for another solution if that's the case.

muvaf: >I've evangelized Crossplane at my company and we've been so interested in it and started implementing it because we've all experienced the suckage of automating Terraform and troubleshooting it and fixing it's state management.

I understand the frustration you've experienced with Terraform but it would be really helpful if you can point to specific details that we can relate to our usage of Terraform in Terrajet. I think quite a good portion of them may not be relevant since we're interacting with Terraform and its providers in a single and consistent way via its stable interfaces.

> I would rather have to write my own where I need coverage than have to debug Terraform being used by Crossplane.

While Terraform and Crossplane have architectural differences, some of the problems users experience with Terraform are because of the problem that both projects try to solve - building a declarative API on top of imperative calls of cloud vendors. So I think the classic providers are likely to have some users with experience similar to TF providers that are written manually once they get to that coverage level. To be fair, the set of problems are not limited to those common ones but without the specifics, it's hard to dismiss that the worries you've got will appear in classic providers as well.

> I can see combining the providers, but I can't believe we're considering moving towards Terraform and away from native.

We need to define what _native_ is in this context and I assume it's making direct SDK calls in the controller, i.e. what's defined as _classic_ in this doc. And as Crossplane maintainers, I think we it's safe to say that we're seeing manual implementation of the controllers as not sustainable due to the sheer number of APIs exposed by these cloud vendors. In fact, the cloud vendors themselves also come up with tooling to scale as mentioned in the doc like AWS Cloud Control, Azure Resource Manager. So, what we're betting on is that the next step after Terrajet will not be _classic_ but more likely these tools that, similar to Terrajet, sit in the middle of the controller and the final API calls. So we're moving away from classic is because vendors don't prefer it either and there isn't a really good way to scale their maintenance without a middle layer - be it Terrajet or vendor tools. At that point, it's hard to tell that making direct API calls is _native_ and others are not. More details [here](https://github.com/crossplane/crossplane/blob/master/design/design-doc-provider-strategy.md) about these tools in the strategy doc.

negz:
Hey folks - thank you for all the great feedback here. We really appreciate having such an involved community helping us make sure we make the right call. Those of us on the Crossplane steering committee have been working on an updated project charter that may affect the decision the provider maintainers make WRT merging these providers; we're hoping to share it publicly before the next community meeting. In the meantime we're going to keep this proposal open but not approve/merge it just yet.

design/one-pager-merge-jets-proposal.md
snarlysodboxer: @muvaf Thanks for the level headed response. I definitely spoke from a place of little knowledge of how Crossplane uses Terraform, that's on me.

> it would be really helpful if you can point to specific details that we can relate to our usage of Terraform in Terrajet. I think quite a good portion of them may not be relevant since we're interacting with Terraform and its providers in a single and consistent way via its stable interfaces.

It's been a while since I've been deep in the weeds with Terraform, but the kinds of problems I remember are:
* The troubleshooting and contribute-ablity issue: When I've automated Terraform it always involved shelling out to the binary. The problem is not just about running a debugger, it's also about following the logic path mentally. When there's a problem I'll often just go straight to reading the code, but the reasonableness of doing that shifts significantly with the issues surrounding having to jump to a new codebase, and I've repeatedly found myself feeling it's easier to stab in the dark until it works with Terraform than track logic through it's code and then track that back to the automation. It's not just about finding the right place in the Terraform code to pick up reading, it involves tracking down Terraform and Module versions and context switching between them as well. You can't "GoToDefinition". IMO, the path to debugging has crossed a critical threshold of maintainability.
* The lack of comparing the real-world to the intended-world before running operations, (instead, comparing to the state file.): - This may not be the case with Crossplane/Terrajet's usage of Terraform? - It sounds like you're probably familiar with the issues surrounding that state file getting out of sync due to an interrupted operations, etc.

> So I think the classic providers are likely to have some users with experience similar to TF providers that are written manually once they get to that coverage level.

I agree there'll be bugs in both forms of providers, but to me it's about the approachability of providing fixes. With one approach it's all right there in one place in the code, with the other it's a whole different paradigm.

> I think we it's safe to say that we're seeing manual implementation of the controllers as not sustainable

OK. I'm shifting my thinking...

> So, what we're betting on is that the next step after Terrajet will not be classic but more likely these tools that, similar to Terrajet, sit in the middle of the controller and the final API calls. So we're moving away from classic is because vendors don't prefer it either and there isn't a really good way to scale their maintenance without a middle layer - be it Terrajet or vendor tools. At that point, it's hard to tell that making direct API calls is native and others are not.

Oh, OK, I didn't followed this part. The consideration of using those middle layer APIs is kind of key though, right? - **That's hugely different** than what I understood this proposal to be, because the vendors will maintain those middle APIs (AWS Cloud Control, etc), which means we won't have to dubug them, we won't have to follow logic from Crossplane code through to Terrajet through to Terraform through to a Module's API calls. - We'll only have to follow logic inside Crossplane's code and imported libraries, and report issues with a vendor's API to that vendor.

muvaf:
Hey folks! With the updated charter, the classic providers will be moved out of crossplane org and [be part of crossplane-contrib](https://github.com/crossplane/crossplane/issues/3054) so this decision is completely up to the maintainers of those repositories now instead of the Crossplane steering committee. In addition, there are some valid concerns stated in this PR about having Terraform called in these providers alongside direct SDK calls. So, I am closing my proposal and let maintainers of each repo discuss and decide what's best for each individual provider.

It's been great to see this fruitful discussion and how collaborative the Crossplane community is. Thank you!

PR #2932, opened 2022-03-01
Support package base layer descriptor in package images

### LINKED ISSUE #2888: Support layer descriptor annotations in package manager

The new [`xpkg` specification](https://github.com/crossplane/crossplane/blob/master/docs/reference/xpkg.md#manifests) defines the `io.crossplane.xpkg` annotation, which can be added to a layer descriptor to inform Crossplane that only that layer should be pulled and considered for package contents.

The package manager should honor the behavior as described in the specification. The Crossplane CLI should also support building packages in this manner, but given the fact that it currently only supports single layer images anyway, that work can be decoupled.

hasheddan:

Updates the package manager (primarily the package revision controller) to honor `io.crossplane.xpkg: base` layers in Crossplane packages, as described in [the `xpkg` specification](https://github.com/crossplane/crossplane/blob/master/docs/reference/xpkg.md#manifests). A few notes on the implementation:
- The `ImageBackend` used for the package parser, which IMO was pretty bloated (and still is to some extent, but further improvements require restructuring parser interface in `crossplane-runtime`), is refactored to _only_ deal with fetching package images and extracting the YAML stream. It does not handle caching.
- If one layer descriptor is annotated with `io.crossplane.xpkg: base`, the `ImageBackend` will _only_ download it for extracting the YAML stream. If no layer descriptor is annotated, the _whole_ image is downloaded and flattened to extract the YAML stream, but only the stream contents are cached.
- The revision reconciler is refactored to handle all caching directly, rather than deferring some of it to the `ImageBackend`, which created a leaky separation of concern.
- In the event that package contents are not in the cache, we read the YAML stream contents while also passing them along to the cache to be written to a gzipped file concurrently.
- In the event that package contents are in the cache, we stream them through the gzip reader to the parser.

> Note: potential future optimizations can be made around the use of `afero.TarFs`, which reads file contents into a map. The simplest improvement would be to just return the `tar.Reader` in annotated packages since we require that the `package.yaml` is the only file in that layer. I am running some benchmarks to illustrate the change in performance before implementing.

Fixes #2888 

> Note: as detailed in the issue, additional machinery will be added to `crank` / `crossplane-runtime` to support writing annotated package layer descriptors.

I have:

Tested with both existing packages (`registry.upbound.io/upbound/platform-ref-azure:v0.1.0`) and one with annotated layer descriptors ([hasheddan/cool-layers:v0.0.3](https://explore.ggcr.dev/?image=hasheddan%2Fcool-layers%3Av0.0.3)). Cache used a local directory mount:

```
🤖 (crossplane) k get pkgrev
NAME                                                                              HEALTHY   REVISION   IMAGE                                                   STATE    DEP-FOUND   DEP-INSTALLED   AGE
configurationrevision.pkg.crossplane.io/upbound-platform-ref-azure-2941a425862c   True      1          registry.upbound.io/upbound/platform-ref-azure:v0.1.0   Active   2           2               18s

NAME                                                                        HEALTHY   REVISION   IMAGE                                                   STATE    DEP-FOUND   DEP-INSTALLED   AGE
providerrevision.pkg.crossplane.io/crossplane-provider-azure-4ce0c0d45ac9   True      1          registry.upbound.io/crossplane/provider-azure:v0.18.1   Active                               14s
providerrevision.pkg.crossplane.io/crossplane-provider-helm-295bcd0e6dc3    True      1          registry.upbound.io/crossplane/provider-helm:v0.9.0     Active                               8s
providerrevision.pkg.crossplane.io/hasheddan-cool-layers-0a2ec46c4ef1       True      1          hasheddan/cool-layers:v0.0.3                            Active                               3m14s

🤖 (crossplane) ls -la .work/inttest-package-cache/ --block-size=K
total 420K
drwxrwxrwx 2 dan   dan     4K Mar  1 14:37 .
drwxrwxr-x 6 dan   dan     4K Mar  1 14:31 ..
-rw-r--r-- 1 65532 65532  33K Mar  1 14:37 crossplane-provider-azure-4ce0c0d45ac9.gz
-rw-r--r-- 1 65532 65532  13K Mar  1 14:37 crossplane-provider-helm-295bcd0e6dc3.gz
-rw-r--r-- 1 65532 65532 347K Mar  1 14:35 hasheddan-cool-layers-0a2ec46c4ef1.gz
-rw-r--r-- 1 65532 65532   9K Mar  1 14:37 upbound-platform-ref-azure-2941a425862c.gz
```

Also verified that unzipped contents of items in the package cache matched expected YAML stream.

### DIFF
--- a/internal/controller/pkg/revision/imageback.go
+++ b/internal/controller/pkg/revision/imageback.go
@@ -23,7 +23,6 @@ import (
 
 	"github.com/google/go-containerregistry/pkg/name"
 	"github.com/google/go-containerregistry/pkg/v1/mutate"
-	"github.com/spf13/afero/tarfs"
 
 	"github.com/crossplane/crossplane-runtime/pkg/errors"
 	"github.com/crossplane/crossplane-runtime/pkg/parser"
@@ -110,6 +109,11 @@ func (i *ImageBackend) Init(ctx context.Context, bo ...parser.BackendOption) (io
 		if a, ok := l.Annotations[layerAnnotation]; !ok || a != baseAnnotationValue {
 			continue
 		}
+		// NOTE(hasheddan): the xpkg specification dictates that only one layer
+		// descriptor may be annotated as xpkg base. Since iterating through all
+		// descriptors is relatively inexpensive, we opt to do so in order to
+		// verify that we aren't just using the first layer annotated as xpkg
+		// base.
 		if foundAnnotated {
 			return nil, errors.New(errMultipleAnnotatedLayers)
 		}
@@ -123,14 +127,31 @@ func (i *ImageBackend) Init(ctx context.Context, bo ...parser.BackendOption) (io
 			return nil, errors.Wrap(err, errGetUncompressed)
 		}
 	}
-	// If we still don't have content then we need to flatten image and
-	// extract its package file in filesystem root.
+
+	// If we still don't have content then we need to flatten image filesystem.
 	if !foundAnnotated {
 		tarc = mutate.Extract(img)
 	}
-	fs := tarfs.New(tar.NewReader(tarc))
-	f, err := fs.Open(xpkg.StreamFile)
-	return f, errors.Wrap(err, errOpenPackageStream)
+
+	// The ReadCloser is an uncompressed tarball, either consisting of annotated
+	// layer contents or flattened filesystem content. Either way, we only want
+	// the package YAML stream.
+	t := tar.NewReader(tarc)
+	for {
+		h, err := t.Next()
+		if err != nil {
+			return nil, errors.Wrap(err, errOpenPackageStream)
+		}
+		if h.Name == xpkg.StreamFile {
+			break
+		}
+	}
+
+	// NOTE(hasheddan): we return a JoinedReadCloser such that closing will free
+	// resources allocated to the underlying ReadCloser. See
+	// https://github.com/google/go-containerregistry/blob/329563766ce8131011c25fd8758a25d94d9ad81b/pkg/v1/mutate/mutate.go#L222
+	// for more info.
+	return xpkg.JoinedReadCloser(t, tarc), nil
 }
 
 // nestedBackend is a nop parser backend that conforms to the parser backend
--- a/internal/controller/pkg/revision/imageback.go
+++ b/internal/controller/pkg/revision/imageback.go
@@ -22,10 +22,8 @@ import (
 	"io"
 
 	"github.com/google/go-containerregistry/pkg/name"
-	regv1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/mutate"
 	"github.com/spf13/afero/tarfs"
-	corev1 "k8s.io/api/core/v1"
 
 	"github.com/crossplane/crossplane-runtime/pkg/errors"
 	"github.com/crossplane/crossplane-runtime/pkg/parser"
@@ -35,17 +33,23 @@ import (
 )
 
 const (
-	errPullPolicyNever   = "failed to get pre-cached package with pull policy Never"
-	errBadReference      = "package tag is not a valid reference"
-	errFetchPackage      = "failed to fetch package from remote"
-	errCachePackage      = "failed to store package in cache"
-	errOpenPackageStream = "failed to open package stream file"
+	errBadReference            = "package tag is not a valid reference"
+	errFetchPackage            = "failed to fetch package from remote"
+	errGetManifest             = "failed to get package image manifest from remote"
+	errFetchLayer              = "failed to fetch annotated base layer from remote"
+	errGetUncompressed         = "failed to get uncompressed contents from layer"
+	errMultipleAnnotatedLayers = "package is invalid due to multiple annotated base layers"
+	errOpenPackageStream       = "failed to open package stream file"
+)
+
+const (
+	layerAnnotation     = "io.crossplane.xpkg"
+	baseAnnotationValue = "base"
 )
 
 // ImageBackend is a backend for parser.
 type ImageBackend struct {
 	registry string
-	cache    xpkg.Cache
 	fetcher  xpkg.Fetcher
 }
 
@@ -60,9 +64,8 @@ func WithDefaultRegistry(registry string) ImageBackendOption {
 }
 
 // NewImageBackend creates a new image backend.
-func NewImageBackend(cache xpkg.Cache, fetcher xpkg.Fetcher, opts ...ImageBackendOption) *ImageBackend {
+func NewImageBackend(fetcher xpkg.Fetcher, opts ...ImageBackendOption) *ImageBackend {
 	i := &ImageBackend{
-		cache:   cache,
 		fetcher: fetcher,
 	}
 	for _, opt := range opts {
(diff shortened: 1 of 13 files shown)


hasheddan (commented):
@muvaf @negz I updated to eliminate the `tarfs` usage as mentioned previously and when using a non-annotated fat package (using `hasheddan/provider-aws-fat:v0.0.2` for testing), the reduction in allocations is significant as we are no longer reading the whole filesystem into memory just to get the `package.yaml`.

> Note: these benchmarks are run only on `ImageBackend.Init`, not on the subsequent parsing, which is still quite inefficient due to how we are consuming the upstream k8s tooling.

With `tarfs`:
```
🤖 (revision) go test -bench=. -benchmem -memprofile memprofiletarfs.out -benchtime=20s
goos: linux
goarch: amd64
pkg: github.com/crossplane/crossplane/internal/controller/pkg/revision
cpu: Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
BenchmarkInit-12    	      15	1536714280 ns/op	291899693 B/op	  237475 allocs/op
PASS
ok  	github.com/crossplane/crossplane/internal/controller/pkg/revision	55.692s
```

Without:
```
🤖 (revision) go test -bench=. -benchmem -memprofile memprofile.out -benchtime=20s
goos: linux
goarch: amd64
pkg: github.com/crossplane/crossplane/internal/controller/pkg/revision
cpu: Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
BenchmarkInit-12    	      69	 336935786 ns/op	 4640937 B/op	    5012 allocs/op
PASS
ok  	github.com/crossplane/crossplane/internal/controller/pkg/revision	29.939s
```

internal/controller/pkg/revision/imageback.go
```diff
@@ -85,45 +88,70 @@ func (i *ImageBackend) Init(ctx context.Context, bo ...parser.BackendOption) (io
...
+		if a, ok := l.Annotations[layerAnnotation]; !ok || a != baseAnnotationValue {
+			continue
+		}
+		// NOTE(hasheddan): the xpkg specification dictates that only one layer
+		// descriptor may be annotated as xpkg base. Since iterating through all
+		// descriptors is relatively inexpensive, we opt to do so in order to
+		// verify that we aren't just using the first layer annotated as xpkg
+		// base.
+		if foundAnnotated {
+			return nil, errors.New(errMultipleAnnotatedLayers)
```
negz, line 118: Would it make sense to fall back to using the artifacts from the 'flattened' image in this case? I'm thinking no; it would perhaps be more compatible, but also less predictable and in violation of the spec.

internal/xpkg/cache.go
```diff
@@ -17,95 +17,103 @@ limitations under the License.
...
+	Get(id string) (io.ReadCloser, error)
+	Store(id string, content io.ReadCloser) error
 	Delete(id string) error
 }
 
-// ImageCache stores and retrieves OCI images in a filesystem-backed cache in a
-// thread-safe manner.
-type ImageCache struct {
+// PackageCache stores and retrieves package content in a filesystem-backed
+// cache in a thread-safe manner.
```
negz, line 45: I imagine changing from a cache of packages to a cache of YAML streams has some implications:

1. We're slightly more committed to "the content of a Crossplane package" as far as the things the package manager needs to operate on a package being a single YAML stream. Probably not a problem?
2. Folks updating to this implementation will presumably have their package caches invalidated. Is It worth adding logic to cleanup the old image based cache? IMO probably not.

internal/xpkg/cache.go
```diff
@@ -17,95 +17,103 @@ limitations under the License.
...
+type PackageCache struct {
```
negz, line 46: Naming nit: Aren't all hypothetical instances of `Cache` (the interface) "a package cache"? What do you think about:

* `type Cache interface` -> `type PackageCache interface`
* `type PackageCache struct` -> `type FilesystemPackageCache struct` (or similar)

internal/xpkg/cache.go
```diff
@@ -17,95 +17,103 @@ limitations under the License.
...
-// A Cache caches OCI images.
+const cacheContentExt = ".gz"
+
+// A Cache caches package content.
 type Cache interface {
-	Get(tag string, id string) (v1.Image, error)
-	Store(tag string, id string, img v1.Image) error
+	Has(id string) bool
+	Get(id string) (io.ReadCloser, error)
+	Store(id string, content io.ReadCloser) error
```
negz, line 40: Should we clarify what format `content io.ReadCloser` is expected to contain - i.e. a gzipped xpkg compliant YAML stream? In particular the fact that it's gzipped wasn't obvious to me until I read the implementation.

internal/xpkg/reader.go
```diff
@@ -0,0 +1,118 @@
...
+var _ io.ReadCloser = &gzipFileReader{}
+
+// gzipFileReader reads compressed contents from a file.
+type gzipFileReader struct {
+	f    afero.File
+	gzip *gzip.Reader
+}
+
+// GzipFileReader constructs a new gzipFileReader from the passed file.
+func GzipFileReader(f afero.File) (io.ReadCloser, error) {
```
negz, line 35: Does this need to take `afero.File` as an argument? At a glance it seems like an `io.ReadCloser` would work.

internal/controller/pkg/revision/reconciler.go
```diff
@@ -386,22 +391,87 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+	id := pr.GetName()
+	// If packagePullPolicy is Never, the identifier is the package source and
+	// contents must be in the cache.
+	if pr.GetPackagePullPolicy() != nil && *pr.GetPackagePullPolicy() == corev1.PullNever {
+		pullPolicyNever = true
+		id = pr.GetSource()
+	}
+
+	var rc io.ReadCloser
+	var err error
```
negz, line 404: Could this `err` be more tightly scoped? I think it's only used in the `if r.cache.Has` block, right?

negz (approved):
A few thoughts, but nothing blocking. I like the refactoring to pull the cache out of the image backend. Thanks @hasheddan!

internal/controller/pkg/revision/reconciler.go
hasheddan: Absolutely, I think this is a relic of a previous iteration, thanks for pointing out!

internal/xpkg/cache.go
hasheddan: The `content` here is actually not gzipped. Content comes into the cache uncompressed, `Store` wraps the file being written to in a gzip writer, and `Get` wraps the file being read in a gzip reader.

internal/xpkg/cache.go
hasheddan: 1. I don't think moving to caching just the YAML stream _necessarily_ prohibits us from extending the package format in the future -- all packages would just need to be pulled again, which shouldn't be a huge issue.
2. I thought about this, but given that there is not a possibility for conflict with content previously written by the cache, I agree that the added complexity of trying to clean up is not worth it. One area I do want to note in the release notes is that if you had manually populated the cache, you'll need to repopulate it again with a different process. I don't think this is a common use case today though.

internal/xpkg/cache.go
hasheddan: Great call out -- I think I'll opt for `PackageCache` and `FsPackageCache` 👍🏻

internal/xpkg/reader.go
hasheddan: Nope -- good catch! This can just be a `GzipReadCloser` 👍🏻

PR #2933, opened 2022-03-02
all: fix some typos

cuishuang:


hasheddan (commented):
Thanks for these updates @cuishuang! Could you add sign-off to your commit as described [here](https://github.com/crossplane/crossplane/pull/2933/checks?check_run_id=5389506180) so that we can get the DCO check passing?

cuishuang:
> Thanks for these updates @cuishuang! Could you add sign-off to your commit as described [here](https://github.com/crossplane/crossplane/pull/2933/checks?check_run_id=5389506180) so that we can get the DCO check passing?

Done

hasheddan (approved):
LGTM! Adding backports to active branches since one of these fixes is for live documentation on crossplane.io 👍🏻

PR #2938, opened 2022-03-03
feat(composition): Add patch fromObjectFieldPath

### LINKED ISSUE #2099: Support patching from common data sources?

MisterMX:

Adds the ability to patch from any K8s object using `type: FromObjectFieldPath`.

The changes are purely additive and do not break the existing API.

Internally, the controller tries to find the object using `fromObjecRef` and passes the found object to the already existing `applyFromFieldPath`.

Fixes #2099

A patch would look like this: 

```yaml
      patches:
        - type: FromObjectFieldPath
          fromObjectRef:
            apiVersion: v1
            kind: ConfigMap
            name: sample-config
            namespace: sample-ns
          fromFieldPath: data.value
          toFieldPath: spec.forProvider.sampleField
          policy:
            fromFieldPath: Required # Dont render if referenced resource does not exist
```

_I know there are a few open design discussions but they seam to have been gone stale for quite some time. So I wanted to give it a bit of kickstart ;-)_

I have:

Unit tests


negz:
Just ACKing that I've seen this and appreciate the prompt to think about this problem. 😉 I'm not sure whether this functionality fits well as part of a `Composition` though. Would we support patching an XR in a similar fashion? Or would building this at the MR level (e.g. via something like https://github.com/crossplane/crossplane/issues/1770) make sense?

MisterMX:
@negz thanks for taking a look at this.

> I'm not sure whether this functionality fits well as part of a Composition though.

From our perspective it makes sense to have this feature supported by compositions because

1. All patching is done in compositions, so it feels "natural" from a user perspective to have it there
2. We could bring this feature globally to users with one release of Crossplane
    * We don't depend on provider maintainers to add this to each provider and release
    * We would be able to use this with older provider versions
    * We would ensure consistency across the API. (I am imagining running different providers which do patching from objects differently or don't support it at all).

> Would we support patching an XR in a similar fashion?

Although I don't have a good use case in mind this feature could be implemented quite easily:

Currently, applying a `FromObjectFieldPath` patch is done by getting the referenced object and passing it to
```golang
c.applyFromFieldPathPatch(ref, mr)
```
where `ref` is the referenced object and `mr` is the managed resource.

Patching to XR would be done by simply swapping `mr` with `xr`.

Does this makes sense?

MisterMX:
While doing some testing I noticed that sometimes `fromObjectRef` needs to be dynamic. For example you may want your namespace to match `metadata.labels[crossplane.io/claim-namespace]`.

I added two additional fields `nameFromFieldPath` and `namespaceFromFieldPath` to extract a value from an arbitrary composite field.

negz:
We discussed this in last Thursday's community meeting, and unfortunately the maintainers aren't feeling ready to merge this as is:

* At least @muvaf and I are still unsure whether Composition (vs managed resources) is the right place for this.
* There are RBAC implications here that need to be worked through - this could be used to read anything Crossplane could read (and Crossplane would need RBAC access to read anything it was asked to reference).

My hunch is that a one-pager might be the best way to address this (and possibly its interactions with https://github.com/crossplane/crossplane/issues/1770). In the mean-time we're making good progress on Custom Compositions (https://github.com/crossplane/crossplane/pull/2886) and hope to have them implemented in Crossplane v1.8. My hope is that they might lower the urgency of this feature, since it would become possible to decorate a Composition with a "function" that pulled in state from an external object. In my mind this would be a good way to experiment with this functionality before we make it a "first class" part of Composition.

MisterMX:
@negz thanks for your reply. We discussed this internally as well and we understand your security concerns. With this you could use compositions to escalate privileges on the cluster. This would most likely be a concern if generic referencers would be implemented on MR level.

On the other hand, however, we think that RBAC and security should not be a concern of the composition author.

Therefore we would propose to make this feature to be enabled via a feature flag similar to composition revisions, i.e. `--enable-generic-referencers`. This way the cluster operator would have to make the explicit decision to enable it and is therefore responsible for the appropriate RBAC setup.

Regarding #1770: Why don't support both? Both features are not exclusive to each other imo.

I also had a look at #2886: If I understood it right it seems to be a bit overcomplicated for us. We basically just want to fetch values from a config map. And it also seems to be quite a long journey until it arrives. And again - why not support both?

negz:
I'm going to close this one, since @MisterMX is taking a new approach in https://github.com/crossplane/crossplane/pull/3007.

Issue #2608, opened 2021-09-30
Conversion support for XRD versions?


turkenh:
### What problem are you facing?

Crossplane allows you [to specify multiple versions in an XRD](https://github.com/crossplane/crossplane/pull/1884) but does not support providing a conversion strategy.

> Note that all versions must (by convention) have identical schemas; generated CRDs always use the 'None' conversion strategy to convert between the schemas of different CR versions. Crossplane may support other conversion strategies (e.g. webhooks or declarative conversion) in future.

See the related discussion: https://github.com/crossplane/crossplane/discussions/2605

### How could Crossplane help solve your problem?

Consider adding support for a conversion strategy for XRD versions.

negz:
Thanks for raising this @turkenh - I didn't realise we didn't have an issue tracking it. As you pointed out, we deferred this for now because it seems like the only option would be to have folks write conversion webhooks (presumably in Go), which would somewhat defeat the "no code" nature of XRs. I think the first step here would be exploring what a declarative conversion strategy would look like. Some prior thought on this can be found upstream [here](https://github.com/kubernetes/community/blob/3c61bb0db08277d6f23d457927d3db628dbba350/contributors/design-proposals/api-machinery/customresource-conversion-webhook.md#alternatives-considered).

yhrn:
Hi,

We are evaluating Crossplane for our internal platform and this limitation might be a blocker for us. Has there been any further thinking or progress? Is webhook conversion definitely off the table?

I understand you want to stay true to the "no code" idea but if such a conversion strategy is so hard to implement that it just won't happen it really means that XRs misses out on one of the most fundamental and important features of k8s APIs.

negz:
There's tentative plans within Kubernetes API machinery to explore CEL based declarative conversion (after the recent CEL based validation work). I expect this could get us this functionality "for free". I'll try find a link.

In the meantime we're not opposed to allowing webhook based conversion to bridge the gap.

negz:
Supporting multiple versions in a declarative way - i.e. without writing conversion code - seems like a pretty hard problem. Tackling that problem is not on the horizon for us at the moment. I do recall that @ncdc prototyped CEL based conversion but I can't immediately find a reference to that.

[...]

PR #2948, opened 2022-03-07
feat(xrd): Add support for version deprecation

### LINKED ISSUE #2945: XRD versions do not support deprecation
### What happened?

XRDs do not support the `deprecated` and `deprecationWarning` fields that [are available for CRDs](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/).

### How can we reproduce it?

1. Define a XRD with a version. Set `spec.versions.deprecated` to true.
2. Run `kubectl apply`

Crossplane version: 1.6.3

MisterMX:

This adds two new fields `deprecated` and `deprecationWarning` to XRDs `spec.versions[]` that match their CRD counterparts.

Fixes #2945 

I have:

Manually


apis/apiextensions/v1/xrd_types.go
```diff
@@ -98,6 +98,14 @@ type CompositeResourceDefinitionVersion struct {
 	// Served specifies that this version should be served via REST APIs.
 	Served bool `json:"served"`
 
+	// The deprecated field specifies that this version is deprecated and should not be used.
+	// +optional
+	Deprecated *bool `json:"deprecated,omitempty"`
+
+	// DeprecationWarning specifies the message that should be shown to the user when using this version.
```
hasheddan, line 105: Are these comments wrapped at 80 characters? If not could you make sure to wrap them! 🙂

hasheddan (approved):
LGTM -- I'm a fan of this change even though in practice I believe we can still only offer one XRD version at a time. Would like to get @negz thoughts before merge.

apis/apiextensions/v1/xrd_types.go
MisterMX: Fixed

MisterMX:
@hasheddan thanks for the review! I fixed the line lengths. Let me know if something is missing.

Regarding serving multiple XRD versions: AFAIK Crossplane _does_ support multiple versions at the same time.

However, our main use case is to let users know when they use a deprecated version. In some cases we do deprecate complete XRDs so we might need this for every version.

negz:
> AFAIK Crossplane does support multiple versions at the same time.

Kind of. It supports it as long as all the versions are identical. :) So, that said, this feature makes sense. e.g.:

1. You start with a v1alpha1 API.
2. You 'promote' the API by introducing an identical v1beta1 variant.
3. You mark the v1alpha1 API deprecated.

Eventually we'll support introducing new API versions with different schemas per https://github.com/crossplane/crossplane/issues/2608.

negz (approved):

PR #2956, opened 2022-03-10
[DO NOT MERGE] Add one-pager for Terraform registry scrapers and registry metadata

ulucinar:

Fixes crossplane/terrajet/issues/203

With this one-pager, we would like to discuss how metadata about Terraform native resources can be scraped and processed from the [Terraform registry](https://registry.terraform.io/) and used in [Terrajet codegen pipelines](https://github.com/ulucinar/terrajet/blob/fix-48/pkg/pipeline/example.go) to produce [example manifests](https://github.com/crossplane-contrib/provider-jet-azure/tree/main/examples-generated) or CRD documentation and in various other use cases. 

We believe such metadata offers large potential to improve the user experience with the Crossplane providers generated using Terrajet as well as the quality of them (e.g., auto-generated CRD documentation).

We aim to gather community feedback and stir up some discussions to discover new use cases with such metadata in Terrajet-based providers.

I have:

N.A.

**NOTE**: We are considering to move this proposal to https://github.com/crossplane/terrajet after the review process is completed.

### DIFF
--- a/design/one-pager-terrajet-metadata-extraction.md
+++ b/design/one-pager-terrajet-metadata-extraction.md
@@ -3,7 +3,7 @@
 * Reviewers: Crossplane Maintainers
 * Status: Draft
 
-### Background
+## Background
 For providers generated using [Terrajet], the number of managed resources can
 exceed [several hundreds](provider-jet-aws-preview), and especially for the big
 three Terrajet-based providers ([provider-jet-aws], [provider-jet-gcp] and
@@ -40,7 +40,7 @@ accepting the scraped metadata from the Terraform registry.
 - Extension of existing Terrajet codegen pipelines to also generate
   documentation on `struct`s and fields.
 
-### Goals
+## Goals
 We would like to achieve the following goals with this proposal:
 - Terrajet [resource configuration][resource configuration API] framework allows
   us to customize/adjust the code generation pipeline invoked for generating the
@@ -78,6 +78,41 @@ We would like to achieve the following goals with this proposal:
   Terrajet pipelines will always be working on a well defined format regardless
   of how those metadata are scraped.
 
+
+## Proposal
+We propose to scrape metadata about the Terraform providers and their resources
+from the [Terraform registry] and to define an easily consumable format for
+storing the scraped metadata. We can use the extracted metadata later in the
+various automated pipelines (in the existing Terrajet code-generation pipeline, and
+further in the newly added pipelines, such as an example-manifest generation
+pipeline). We can scrape the example Terraform configurations represented in
+[HCL], store them in the common metadata format, and then with a new
+example-manifest generation pipeline run right after the existing Terrajet
+code-generation pipeline, we can generate the YAML example manifests for the
+generated managed resources. Or, we can incorporate Terraform resource
+documentation, which is extracted with these scrapers from the Terraform registry and
+stored in the common metadata format, in the existing Terrajet code-generation
+pipeline to generate the CRD documentation (embedded as Go comments in the CRD
+`struct` types and their corresponding fields). There are also other use cases
+that we can enable using this extracted registry metadata, such as standardizing
+API groups of the generated managed resources.
+
+There is a plethora of metadata that we can extract from the Terraform registry.
+HCL-formatted example resource configurations in the registry can be used for
+generating YAML-formatted managed resource example manifests (this has already
+been accomplished in [[1]]). Documentation of the Terraform resources can be
+incorporated as the CRD documentation for the generated Crossplane managed
+resources. Subcategory information or path components extracted from Terraform
+import statements are candidates to be used as the API group names of the
+generated managed resources. 
+
+In the following sections, we describe the proposed [common metadata
+format][Proposed Metadata Format] for storing the extracted registry metadata.
+Then we discuss how we can [scrape metadata][Metadata scrapers] from the
+registry and how we can incorporate this metadata in new and existing
+[code-generation pipelines][Terrajet Codegen Pipelines Consuming Metadata]. We
+conclude with a discussion on some [alternatives considered][Alternatives
+Considered] and [future considerations][Future Considerations]. 
 
 ### Proposed Metadata Format
 The proposed syntax for scraped metadata documents is YAML as we would also like
@@ -258,6 +293,12 @@ Because these ID strings appear in the import statements of most
 `terraform-provider-azurerm` resources, using such metadata enables us to have a
 consistent, repo-wide defaulting for the API group names of the generated resources. 
 
+
+### Alternatives Considered
+
+### Future Considerations
+
+[1]: https://github.com/crossplane/terrajet/issues/48
 [Terrajet]: https://github.com/crossplane/terrajet
 [provider-jet-aws-preview]:
     https://doc.crds.dev/github.com/crossplane-contrib/provider-jet-aws@v0.4.0-preview
@@ -282,3 +323,4 @@ consistent, repo-wide defaulting for the API group names of the generated resour
     https://github.com/crossplane-contrib/provider-jet-azure/blob/main/config/apigroup_config.go
 [resource configuration API]:
     https://github.com/crossplane/terrajet/blob/main/pkg/config/resource.go
+[HCL]: https://github.com/hashicorp/hcl
--- a/design/one-pager-terrajet-metadata-extraction.md
+++ b/design/one-pager-terrajet-metadata-extraction.md
@@ -364,8 +364,48 @@ consistent, repo-wide defaulting for the API group names of the generated resour
 
 
 ### Alternatives Considered
+#### Manually adding example manifests and CRD documentation:
+We have already been manually maintaining the example manifests and we could
+introduce improved support for adding CRD documentation in the [resource
+configuration framework][resource configuration API]. But especially for large
+providers (in terms of number of supported managed resources), manually
+preparing example manifests and documentation is time consuming. 
+
+#### Generating just Kubernetes Object Metadata and GVK as example manifest:
+During our discussions, one of the alternatives we considered was to generate
+just the object `metadata` and GroupVersionKind (GVK) as the example manifest and have the `spec` later
+populated manually. An example would be:
+```yaml
+apiVersion: iothub.azure.jet.crossplane.io/v1alpha1
+kind: IOTHub
+metadata:
+  name: example
+```
+This is a viable approach as we generally require an example manifest with which
+the Cloud resource was provisioned, in PRs introducing support for new managed
+resources (or for Terrajet-based providers, PRs that configure new resources).
+However, having example manifests [already available][provider-jet-azure-examples] in the Crossplane provider
+repo [greatly simplifies][azurerm-iothub] authoring the `spec` of the example resource manifest.
 
 ### Future Considerations
+#### Maintenance of metadata:
+As we start maintaining scraped metadata in the terrajet-based provider
+repositories, we may occasionally need to override certain aspects of it for
+various reasons like fixing upstream typos, rebranding, or improving example
+manifests and the like. Although we will need to run the scrapers after a
+version bump in the native Terraform provider consumed, we may want to easily
+apply the metadata overrides to the metadata scraped for the newer version of
+the native provider. In order to support such cases, we can consider maintaining
+the overrides themselves as separate patch files to the actual per-resource
+scraped metadata. Then we can leverage a tool like [Kustomize] to apply these
+patches to the relevant resource metadata. Due to upstream changes in the
+Terraform registry, this approach may not always be easily applicable (what
+happens, for instance, if there is a change in the argument names for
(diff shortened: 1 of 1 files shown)


design/one-pager-terrajet-metadata-extraction.md
```diff
@@ -0,0 +1,272 @@
...
+  scraper is not immediately available. This will also allow us, if needed, to
+  have different scrapers that produce output in the same metadata format. For
+  instance, we may have a relatively complex scraper for extracting metadata
+  from the Terraform registry, and another relatively simple one that just adds
+  example HCL configurations by reading them from their respective
+  [files][aws-example-configurations]. This will allow different scraper
+  implementations to be able to fetch metadata from different sources but the
+  Terrajet pipelines will always be working on a well defined format regardless
+  of how those metadata are scraped.
+- We would like to have the scrapers run as needed, produce their output in the
```
muvaf, line 70: I think this specific item is leaking the proposal detail into the goal section. IMO, the goals section as a whole should be a bit higher level and allow alternative proposal to achieve the same goals.

design/one-pager-terrajet-metadata-extraction.md
```diff
@@ -0,0 +1,272 @@
...
+  common metadata format, and to have the metadata documents added to their
+  respective repositories. However, we can then have the corresponding pipelines
+  run each time with a `make generate`, just like the existing codegen pipelines
+  we have. This would allow us to separate the lifecycles of metadata-scraping
+  and code generation.
+
```
muvaf, line 76: A `Proposal` section could be helpful for readers here to understand the whole solution before jumping to the metadata format which is one of the implementation details of the solution.

design/one-pager-terrajet-metadata-extraction.md
```diff
@@ -0,0 +1,272 @@
...
+### Metadata Format
+The proposed syntax for scraped metadata documents is YAML as we would also like
+the metadata to be human readable, searchable and maintainable, if needed. A
```
muvaf, line 79: What is meant by `maintainable`?

design/one-pager-terrajet-metadata-extraction.md
```diff
@@ -0,0 +1,272 @@
...
+concrete example of a scraped registry metadata document for a resource named
+`azurerm_analysis_services_server` of the native Terraform provider
+[terraform-provider-azurerm] could be as follows:
+
+```yaml
```
muvaf, line 84: Having an example definitely helps to understand what goes where but I wonder if we can have an API reference before the examples to show the schema of the YAML.

design/one-pager-terrajet-metadata-extraction.md
muvaf: Listing what's available in TF registry as metadata information would also be very valuable for having readers think about the future expansions.

design/one-pager-terrajet-metadata-extraction.md
```diff
@@ -0,0 +1,272 @@
...
+                create: '- (Defaults to 30 minutes) Used when creating the Analysis Services Server.'
+                delete: '- (Defaults to 30 minutes) Used when deleting the Analysis Services Server.'
+                read: '- (Defaults to 5 minutes) Used when retrieving the Analysis Services Server.'
+                update: '- (Defaults to 30 minutes) Used when updating the Analysis Services Server.'
+        # import statement scraped from the Terraform registry, if available
+        # Can be used for advanced purposes, such as constructing resource config "ExternalName.GetIDFn" functions, etc.
+        importStatements:
+            - terraform import azurerm_analysis_services_server.server /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/resourcegroup1/providers/Microsoft.AnalysisServices/servers/server1
+```
+
```
muvaf, line 165: Could you elaborate on why the chosen method is better than these alternatives? This would help us reason about them more deliberately when we want to make changes to the pipeline.

design/one-pager-terrajet-metadata-extraction.md
```diff
@@ -0,0 +1,272 @@
...
+- For `terraform-provider-aws`:
+  https://github.com/hashicorp/terraform-provider-aws/tree/main/website/docs/r
+- For `terraform-provider-azurerm`:
+  https://github.com/hashicorp/terraform-provider-azurerm/tree/main/website/docs/r
+- For `terraform-provider-google`:
+  https://github.com/hashicorp/terraform-provider-google/tree/main/website/docs/r
+
+Thus a common metadata scraper implementation can extract metadata from these
+well-formatted per-resource markdown documents. Any spotted errors can then
+potentially be corrected manually in the scraped YAML metadata document.
```
muvaf, line 195: How would we prevent the scraper from overriding the manual changes made to the YAML metadata document?

design/one-pager-terrajet-metadata-extraction.md
```diff
@@ -0,0 +1,272 @@
...
+Scrapers can optionally be chained: If desired, another scraper can append
+example HCL configurations read from a different source (such as the `examples`
+folder found in some of the Terraform native provider repositories as discussed
+above).
+
+For most Terraform native providers, we anticipate that Terraform registry
+scrapers will **not** run on HTTP, as the resource markdown files are part of
+their corresponding provider repositories. They can just read those markdowns
+from a pointed directory in the local filesystem, which is specified as a
```
muvaf, line 204: Do you mean it will fetch the `.md` files from the Go module directory or the repositories will be cloned manually to a folder in the filesystem?

design/one-pager-terrajet-metadata-extraction.md
```diff
@@ -0,0 +1,272 @@
...
+command-line argument, for instance. 
+
+As already indicated, if it turns out that a common registry scraper
```
muvaf, line 207: I love that optionality.

design/one-pager-terrajet-metadata-extraction.md
```diff
@@ -0,0 +1,272 @@
...
+and produce example manifests, CRD documentation, etc. (as discussed above). The
+configured pipelines should not fail if the necessary metadata is missing: For
+instance, the [example manifest generation pipeline] should simply not generate
+an example manifest for a managed resource, if no sample HCL configuration is
+available for the corresponding Terraform resource in the metadata. Or, the
+metadata-enhanced CRD generation pipeline should simply skip doc comments if
+none or some are not available in the corresponding metadata document. 
+
+Metadata is valuable; the scrapers should capture as much metadata as possible
+and store them in the common format, even for future use cases we do not yet
```
muvaf, line 230: I'm a bit on the fence about storing everything we can because we have to have a format for each of those information that we don't use and once we do use them, we may want to change the format, i.e. where it lives in the YAML and also since we provide manual input ability, the format become one of our public APIs. So, I wonder if we should, for example, just keep the examples since that's all we use today and then once we use other information, we could add them.

design/one-pager-terrajet-metadata-extraction.md
```diff
@@ -0,0 +1,272 @@
...
+envision. New Terrajet pipelines can be added, or existing ones can be enhanced
+to support advanced use cases. One such proposal could be to extend the CRD
```
muvaf, line 232: A `Future Considerations` section could work well to list all the use cases you have in mind at the moment.

design/one-pager-terrajet-metadata-extraction.md
muvaf: If it's the latter, I wonder if we can go with HTTP just to have it fully automated.

design/one-pager-terrajet-metadata-extraction.md
```diff
@@ -0,0 +1,272 @@
...
+resources. For example, a provider implementation may opt to use the category
+names to group respective APIs. Or the examples provided in the Terraform
+registry hint at reference fields that can help us in auto-generating
+cross-resource references (that appear in those HCL configurations).
+
+While working on generating example manifests for the big three Terrajet-based
+providers `provider-jet-aws`, `provider-jet-gcp` and `provider-jet-azure` in the
+context of the corresponding [Terrajet issue #48], we have seen utility in
+extracting such metadata from the Terraform registry and use it to generate
+example manifests and documentation. In this document, we would like to propose:
```
muvaf, line 32: I think `Background` section could be limited to the problem statement in general and the actual proposal could be kept for its own section.

design/one-pager-terrajet-metadata-extraction.md
```diff
@@ -0,0 +1,272 @@
+# Metadata Extraction from Terraform Registry for Terrajet-based providers
```
muvaf, line 1:
```suggestion
# Metadata Extraction from Terraform Registry in Terrajet
```

Since we're removing the difference between Terrajet vs SDK calls at repository level, I think removing `Terrajet-based provider` or `Terrajet-based repository` terms would be more future-proof for readers.

negz:
Would it make sense to move this design (and perhaps the Terrajet design doc too?) to a new `design/` directory under https://github.com/crossplane/terrajet?

design/one-pager-terrajet-metadata-extraction.md
ulucinar: Thanks @muvaf for the detailed review. 
I have restructured the `Goals` section and removed the proposal hints from it.

design/one-pager-terrajet-metadata-extraction.md
ulucinar: I have added a top-level proposal section giving a high-level overview and made the existing proposal sections subsections of this top-level section. 
I have also extended the discussion on available metadata in the Terraform registry.

design/one-pager-terrajet-metadata-extraction.md
ulucinar: I have added example maintenance tasks for the scraped metadata in the document.

design/one-pager-terrajet-metadata-extraction.md
ulucinar: I think using a full example (an example that contains all keys) with comments instead of a formal notation (like BNF or OpenAPI schema) serves our purposes better here. This is not intended to be a formal specification but rather the purpose is to convey the idea.

design/one-pager-terrajet-metadata-extraction.md
ulucinar: Added clarifications on the choices we made here.

design/one-pager-terrajet-metadata-extraction.md
ulucinar: Added a discussion to the relevant section but also discussed this in the `Future Considerations` section.

design/one-pager-terrajet-metadata-extraction.md
ulucinar: I think we can selectively git-checkout the repo path that contains the markdown files, as part of a `Makefile` target. I have also extended the discussion on alternative transports for the scrapers.

design/one-pager-terrajet-metadata-extraction.md
ulucinar: I have added example use cases for the proposed fields. In fact (apart from documentation) most have been utilized in provider-jet-azure. I think this also serves as an opportunity to show the potential we have here.

design/one-pager-terrajet-metadata-extraction.md
ulucinar: Added a `Future Considerations` section.

design/one-pager-terrajet-metadata-extraction.md
ulucinar: Removed proposal hints from the `Background` section as suggested.

ulucinar:
> 

Hi @negz,
I think what you propose makes sense. Let me have the PR approved here (to preserve the existing context) and then I can reproduce it in https://github.com/crossplane/terrajet referring to the review here. I will also mark this PR with do-not-merge. Thank you!

muvaf:
> Would it make sense to move this design (and perhaps the Terrajet design doc too?) to a new design/ directory under https://github.com/crossplane/terrajet?

I think having all design docs under `crossplane/crossplane` helps with discoverability of those docs for the community. Maybe something like `crossplane/enhancements` or `crossplane/design-docs` that could contain all design docs of Crossplane community? It could be a place for non-crossplane org stuff as well, like provider or other extension designs.

negz:
> I think having all design docs under crossplane/crossplane helps with discoverability of those docs for the community. Maybe something like crossplane/enhancements or crossplane/design-docs (similar to [kubernetes](https://github.com/kubernetes/enhancements)) that could contain all design docs of Crossplane community? It could be a place for non-crossplane org stuff as well, like provider or other extension designs.

I don't feel super strongly, but it does seem a little odd to me that a design pertaining completely to Terrajet is not in the Terrajet repo. We can just stick with everything going in the design folder here if folks feel differently.

ulucinar:
Closing this PR as we would like to keep code generation in terrajet only scoped to generating CRDs.

PR #2968, opened 2022-03-14
feat(claim): Add SYCNED condition to XRs and XRCs

### LINKED ISSUE #2850: Add SYNCED status column to XRC and XR

Currently Crossplane adds a `READY` column to a XRC and XR that is either `True` or `False` depending on the state of the managed resources:

```yaml
- additionalPrinterColumns:
    - jsonPath: .status.conditions[?(@.type=='Ready')].status
      name: READY
      type: string
```

However, it is difficult to distinguish between MR and composition issues without looking inside the XRC or XR. This can be especially troublesome if you have a lot of XRCs.

Crossplane should add an additional column called `SYNCED` to every XRC and XR similar to the `SYNCED` column of managed resources:

```yaml
    - jsonPath: .status.conditions[?(@.type=='Synced')].status
      name: SYNCED
      type: string
```

This would require adding an additional `Synced` condition to `status.conditions`, i.e:

```yaml
  - lastTransitionTime: "2021-12-17T12:50:27Z"
    reason: ReconcileSuccess
    status: "True"
    type: Synced
```

MisterMX:

Fixes #2850

* Add `Synced` condition that can be used to track reconcile errors
(similar to managed resources in providers)
* Add an additional printer column to XR CRDs that references the
`Synced` condition

I have:

Manually and unit tests

### DIFF
--- a/internal/controller/apiextensions/claim/reconciler.go
+++ b/internal/controller/apiextensions/claim/reconciler.go
@@ -56,6 +56,7 @@ const (
 	errGetClaim           = "cannot get composite resource claim"
 	errGetComposite       = "cannot get referenced composite resource"
 	errDeleteComposite    = "cannot delete referenced composite resource"
+	errDeleteUnbound      = "refusing to delete composite resource that is not bound to this claim"
 	errDeleteCDs          = "cannot delete connection details"
 	errRemoveFinalizer    = "cannot remove composite resource claim finalizer"
 	errAddFinalizer       = "cannot add composite resource claim finalizer"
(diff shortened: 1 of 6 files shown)


internal/controller/apiextensions/claim/reconciler.go
```diff
@@ -294,6 +294,18 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
 
+	res, err := r.reconcileClaim(ctx, log, record, cm)
+	if err != nil {
+		cm.SetConditions(xpv1.ReconcileError(err))
+	} else {
+		cm.SetConditions(xpv1.ReconcileSuccess())
+	}
+
+	return res, errors.Wrap(r.client.Status().Update(ctx, cm), errUpdateClaimStatus)
+}
```
negz, line 305: I'd prefer to update the reconciler to set `ReconcileError` as part of the existing `SetConditions` calls, rather than moving the bulk of the logic into a wrapped `reconcile` method. The former approach is more in line with how we approach this in the managed reconciler, for example:

https://github.com/crossplane/crossplane-runtime/blob/988c9ba9c255d7cade32e2c0135ce47b083eaf95/pkg/reconciler/managed/reconciler.go#L684

negz (approved):

negz:
Apparently I can't merge this PR - I think because it hasn't been approved by any CODEOWNERS (which would in this case be a member of the reviewers group or muvaf).

hasheddan (approved):
Stamping to unblock merge based on @negz approval.

bobh66 (approved):

Issue #2947, opened 2022-03-07
Package manager should ignore unknown types instead of stopping the installation


muvaf:
### What problem are you facing?

With https://github.com/crossplane/crossplane/issues/2894 , provider packages will come with `ValidatingWebhookConfiguration` and `MutatingWebhookConfiguration` types. However, if you have an older version of Crossplane before https://github.com/crossplane/crossplane/pull/2919 then the installation/upgrade will not go through at all. This would mean that the providers with webhooks would impose a minimum Crossplane version requirement.

### How could Crossplane help solve your problem?

The best we can do at this point for specific problem of webhooks is to release patches that will make the linter pass the validation and it'd ignore any non-CRD type during the installation of the objects, which will be the same behavior when you don't enable webhooks by not providing a TLS secret.

I think to make it future-proof, the linter should allow all objects as long as they are valid Kubernetes objects and installation should install only the types it needs so that if a new type is introduced, we don't need a code change and existing users wouldn't be affected.

hasheddan:
> This would mean that the providers with webhooks would impose a minimum Crossplane version requirement.

@muvaf I think this is an entirely reasonable requirement. Webhooks are a new feature so patching them back to old release branches would go against are stated semver patch strategy. If we are concerned about upgrading providers to use webhooks, and that requiring a Crossplane version bump, which some users may not want to make initially, but may still want unreleatd fixes included in later versions of the provider, I think options could include:
1. Institute formal support schedules for provider, especially highly used ones (this is probably overdue anyway).
2. Publish webhook and "non-webhook" versions of provider for a period of time.

> I think to make it future-proof, the linter should allow all objects as long as they are valid Kubernetes objects and installation should install only the types it needs so that if a new type is introduced, we don't need a code change and existing users wouldn't be affected.

I actually think requiring users to upgrade to get new feature support in various packages is a good outcome here. I am very hesitant to allow of _arbitrary_ kubernetes object in packages. This would result in a case where a single package could be installed differently based on the version of Crossplane it is being installed with. I would much rather a package say that it can't be installed than to silently ignore items I expect it to install.

muvaf:
> Webhooks are a new feature so patching them back to old release branches would go against are stated semver patch strategy.
> I actually think requiring users to upgrade to get new feature support in various packages is a good outcome here.

Yes, we're on the same page here. The case I'm concerned is the following:
1. I'm a happy user of Crossplane v1.6 and provider-aws v0.24.1
[...]

PR #2982, opened 2022-03-17
[release-1.5] Allow packages with webhook configs to be installed in v1.6 without webhook support

muvaf:

Webhook configs are new types that providers have to include in the package alongside `CustomResourceDefinition`s. Today, you can use those packages with v1.7 and after with `--set webhooks.enabled=false` and you'd see no change. However, if you don't want to use webhooks and are on a version older than v1.7, you cannot install those packages even though they can run without webhook support. See the discussion in https://github.com/crossplane/crossplane/issues/2947

This PR relaxes the linter restriction and ignores webhook config types in v1.5. Users will still need to upgrade to a new version but they will have the choice to update to a patch version instead of minor.

I have:

Manually. Usual installation of Crossplane and installation of `muvaf/provider-aws:v0.25.0-rc.0.22.gac09cd47.dirty` package which contains webhook configs.

### DIFF
--- a/internal/controller/pkg/revision/establisher.go
+++ b/internal/controller/pkg/revision/establisher.go
@@ -19,6 +19,7 @@ package revision
 import (
 	"context"
 
+	admv1 "k8s.io/api/admissionregistration/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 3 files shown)


hasheddan (approved):
Thanks @muvaf!

PR #2986, opened 2022-03-18
Guide for using Vault as an External Secret Store

### LINKED ISSUE #2961: Document External Secret Store feature

With External Secret Store support being available soon, we need to document how to enable/configure the feature.

Add documentation and have a guide for writing secrets to Vault.

turkenh:

This PR adds a guide for configuring and using Vault as an External Secret Store. It also adds `customAnnotations` helm parameter to set custom annotations to the Crossplane pod which is required for injecting the Vault agent as part of the steps in the guide.

Fixes #2961 

TODO:
- [x] Update provider GCP image in the guide once a release with ESS support is available.

I have:

~~By running the steps using a local built provider gcp image, `turkenh/provider-gcp:v0.21.0-rc.13.g273ebf2`, from [this](https://github.com/crossplane/provider-gcp/pull/421) branch.~~

By running the guide as is.

### DIFF
--- a/docs/guides/vault-as-secret-store.md
+++ b/docs/guides/vault-as-secret-store.md
@@ -1,0 +1,452 @@
+---
+title: Vault as an External Secret Store 
+toc: true
+weight: 230
+indent: true
+---
+
+# Using Vault as an External Secret Store
+
+This guide walks through the steps required to configure Crossplane and
+its Providers to use [Vault] as an [External Secret Store]. For the sake of
+completeness, we will also include steps for Vault installation and setup,
+however, you can skip those and use your existing Vault.
+
+> External Secret Stores are an alpha feature. They are not yet recommended for
+> production use, and are disabled by default.
+
+Crossplane consumes and also produces sensitive information to operate which
+could be categorized as follows:
+
+1. **Provider credentials:** These are the credentials required for Providers
+to authenticate against external APIs. For example, AWS Access/Secret keys, GCP
+service account json, etc.
+2. **Connection Details:** Once an infrastructure provisioned, we usually
+need some connection data to consume it. Most of the time, this
+information includes sensitive information like usernames, passwords or access
+keys. 
+3. **Sensitive Inputs to Managed Resources:** There are some Managed resources
+which expect input parameters that could be sensitive. Initial password of a
+managed database is a good example of this category.
+
+It is already possible to use Vault for the 1st category (i.e. Provider
+Credentials) as described in [the previous guide]. 3rd use case is a relatively
+rare and being tracked with [this issue].
+
+In this guide we will focus on the 2nd category, which is storing Connection
+Details for managed resources in Vault.
+
+## Steps
+
+> Some steps in this guide duplicates [the previous guide] on Vault injection.
+> However, for convenience, we put them here as well with minor
+> changes/improvements.
+
+At a high level we will run the following steps:
+
+- Install and Unseal Vault.
+- Configure Vault with Kubernetes Auth.
+- Install and Configure Crossplane by enabling the feature.
+- Install and Configure Provider GCP by enabling the feature.
+- Deploy a Composition and CompositeResourceDefinition.
+- Create a Claim.
+- Verify all secrets land in Vault as expected.
+
+For simplicity, we will deploy Vault into the same cluster as Crossplane,
+however, this is not a requirement as long as Vault has Kubernetes auth enabled
+for the cluster where Crossplane is running.
+
+### Prepare Vault
+
+1. Install Vault Helm Chart
+
+```shell
+kubectl create ns vault-system
+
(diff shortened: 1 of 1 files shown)


docs/guides/vault-as-secret-store.md
```diff
@@ -0,0 +1,452 @@
...
+
+- Enabling `External Secret Stores` feature. 
+- Annotating for [Vault Agent Sidecar Injection]
+
+```shell
+kubectl create ns crossplane-system
+
+helm repo add crossplane-stable https://charts.crossplane.io/stable --force-update
+
+helm --namespace crossplane-system upgrade --install crossplane crossplane-stable/crossplane --namespace crossplane-system \
```
turkenh, line 144: This wouldn't work until we make the release.

Feggah, line 144: I think here is an additional `--namespace` flag in the command:

```suggestion
helm upgrade --install crossplane crossplane-stable/crossplane --namespace crossplane-system \
```

Feggah, line 144: Should we provide a `--version` flag to this command? I tried with the `--version` unspecified and it caused an `Init:Error` in crossplane Pod startup because the default version used was `v1.2` and the flag `--enable-external-secret-stores` isn't available.

If we are already using the provider-gcp image `turkenh/provider-gcp:v0.21.0-rc.13.g273ebf2` in this guide, can't we set the `--version` to something like `v1.8.0-rc.0.5.ge78e6a7a` (the latest today)?

docs/guides/vault-as-secret-store.md
```diff
@@ -0,0 +1,452 @@
...
+
+3. Verify all resources SYNCED and READY:
+
+```shell
+kubectl get managed
+
+kubectl -n default get claim
+
+kubectl get composite
+```
```
Feggah, line 348: I got stuck at this point in the guide (this may be because I'm not familiar with Vault at all).

When I created the claim, it wasn't getting READY, even when the managed resources were:
image

When I `kubectl describe` the composite, I received this error:
```
Events:
  Type     Reason                   Age                  From                                                             Message
  ----     ------                   ----                 ----                                                             -------
  Normal   SelectComposition        13m (x11 over 18m)   defined/compositeresourcedefinition.apiextensions.crossplane.io  Successfully selected composition
  Warning  ComposeResources         13m (x11 over 18m)   defined/compositeresourcedefinition.apiextensions.crossplane.io  cannot fetch connection secret: cannot fetch connection secret: cannot connect to secret store: cannot extract token: open /vault/secrets/token: no such file or directory
```

Is this path `vault/secrets/token` valid? It seems not to be anywhere else in the guide.

After this error, I tried to delete the claim and the `kubectl delete` was hanging. Should this happen? Also, this event appeared in the XR:
```
  Warning  DeleteCompositeResource  3m9s (x11 over 12m)  defined/compositeresourcedefinition.apiextensions.crossplane.io  cannot unpublish connection details: cannot connect to secret store: cannot extract token: open /vault/secrets/token: no such file or directory
```

Then I needed to delete each the managed resource individually and `kubectl patch` to remove the finalizers from the claim and composite. Is this the desired behavior?

docs/guides/vault-as-secret-store.md
turkenh: I guess you didn't update your local helm repo cache, because I would expect you to have the latest v1.6 release. I was expecting the following will do the trick, did you run that command as well? (or skipped thinking you already have it added):

```
helm repo add crossplane-stable https://charts.crossplane.io/stable --force-update
```

turkenh: > If we are already using the provider-gcp image turkenh/provider-gcp:v0.21.0-rc.13.g273ebf2 in this guide, can't we set the --version to something like v1.8.0-rc.0.5.ge78e6a7a (the latest today)?

Honestly, I am not thinking to merge this PR until we have the release. So, I will update those images before merging.

docs/guides/vault-as-secret-store.md
turkenh: Having managed resources as ready is a good indicator meaning it could write its connection secret to Vault. 

For XR/XRC, it looks like the Vault agent didn't get injected as a sidecar which is expected to write the token at `/vault/secrets/token`, and [here](https://www.vaultproject.io/docs/platform/k8s/injector/annotations#vault-hashicorp-com-agent-inject-token) is where that path is coming from. I believe this is expected if you've used latest master build of Crossplane since [this was](https://github.com/crossplane/crossplane/pull/2994) missing, hence required annotations were not added to Crossplane pod to have Vault agent injected and configured.

That change was originally part of this PR but your test reminded me that we need to have it in the release different than the guide itself and opened the PR after seeing your test results.  

So, thanks for testing @Feggah, it really helped a lot 🙌

negz (approved):

PR #3004, opened 2022-03-23
Pin snippet links and configurations to v1.7.0

turkenh:

See https://github.com/crossplane/crossplane/issues/2977

I have:

N/A

### DIFF
--- a/docs/reference/troubleshoot.md
+++ b/docs/reference/troubleshoot.md
@@ -21,7 +21,7 @@ indent: true
 ## Requested Resource Not Found
 
 If you use the kubectl Crossplane plugin to install a `Provider` or `Configuration`
-(e.g. `kubectl crossplane install provider crossplane/provider-aws:master`) and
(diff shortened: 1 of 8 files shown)


hasheddan (commented):
@turkenh could you also set the tags to match on the configuration installs in the getting started guide?

docs/reference/troubleshoot.md
```diff
@@ -21,7 +21,7 @@ indent: true
 ## Requested Resource Not Found
 
 If you use the kubectl Crossplane plugin to install a `Provider` or `Configuration`
-(e.g. `kubectl crossplane install provider crossplane/provider-aws:master`) and
+(e.g. `kubectl crossplane install provider crossplane/provider-aws:v1.7.0`) and
```
hasheddan, line 24: I think this tag may need to remain unchanged since v1.7.0 is not valid for provider-aws

turkenh: ah, good catch! fixed.

hasheddan (approved):

PR #3007, opened 2022-03-24
feat(compositions): Support patching from Environment

MisterMX:
### Description of your changes

Design: https://github.com/crossplane/crossplane/pull/3008

Fixes: #2099 

This adds a new cluster scope CRD `EnvironmentConfig` that is similar to a standard `ConfigMap` but allows storing arbitrary values instead of just strings.

`EnvironmentConfig`s can be used in compositions using the new `FromEnvironmentFieldPath` patch type.

The main use case of this PR is to use to build a generic Crossplane package that can run in multiple environment without having to hardcode environment specific values in the compositions.

This is a replacement for #2938. Based on an idea briefly mentioned by @negz in the community meeting.

It avoids the discussed security and RBAC issues by not requiring access to secrets and config maps and relying on Crossplane native resources instead.

I have:

- [x] Read and followed Crossplane's [contribution process].
- [x] Run `make reviewable` to ensure this PR is ready for review.
- [ ] Added `backport release-x.y` labels to auto-backport this PR if necessary.

### How has this code been tested

Manually and unit tests

[contribution process]: https://git.io/fj2m9

Issue #2836, opened 2022-01-08
CompositionRevision does not work when Composition has a Map transform


ArunRamakani:
Crossplane Version 1.6.1

Issue: CompositionRevision does not work when Composition has a Map transform

Logs Info: Map transform expects a string but it should be an object

```
2022-01-08T06:41:33.479Z	ERROR	crossplane.controller-runtime.manager.controller.revisions/composition.apiextensions.crossplane.io	Reconciler error	{"reconciler group": "apiextensions.crossplane.io", "reconciler kind": "Composition", "name": "mysql", "namespace": "", "error": "cannot create CompositionRevision: CompositionRevision.apiextensions.crossplane.io \"mysql-gp5c7\" is invalid: spec.resources.patches.transforms.map.Pairs: Invalid value: \"object\": spec.resources.patches.transforms.map.Pairs in body must be of type string: \"object\""}
```

ArunRamakani:
Below is the Composition 

apiVersion: apiextensions.crossplane.io/v1
kind: Composition
metadata:
  name: mysql
  labels:
    crossplane.io/xrd: xgcpdbs.alpha-beta.imarunrk.com
spec:
  compositeTypeRef:
    apiVersion: alpha-beta.imarunrk.com/v1
    kind: XGCPdb
  writeConnectionSecretsToNamespace: crossplane-system
  resources:
  - name: cloudsqlinstance
    base:
      apiVersion: database.gcp.crossplane.io/v1beta1
      kind: CloudSQLInstance
      spec:
        providerConfigRef:
          name: gcp-credentials-project-1
        forProvider:
          databaseVersion: MYSQL_5_7
          region: us-central1
          settings:
            tier: db-g1-small
            dataDiskSizeGb: 20
    patches:
    - type: FromCompositeFieldPath
      fromFieldPath: spec.parameters.size
      toFieldPath: spec.forProvider.settings.tier
      transforms:
      - type: map
        map:
          BIG: db-n1-standard-1
          SMALL: db-g1-small
      policy:
        fromFieldPath: Required
    - type: FromCompositeFieldPath
      fromFieldPath: spec.parameters.size
      toFieldPath: spec.forProvider.settings.dataDiskSizeGb
      transforms:
      - type: map
        map:
          BIG: "40"
          SMALL: "20"
      - type: convert
        convert:
          toType: int
      policy:
        fromFieldPath: Required
    - type: ToCompositeFieldPath
[...]

PR #3008, opened 2022-03-24
Design: Add proposal to patch from a generic datasource

MisterMX:

This design proposal aims to provide a solution for adding a third data source for composition patches next to claim/composite and composed resource by introducing a new resource `EnvironmentConfig`.

It is _not_ the goal of this design document to provide a way for patching from any Kubernetes object (aka generic referencers).

CC @negz 

Related to https://github.com/crossplane/crossplane/issues/2099

Implementation: https://github.com/crossplane/crossplane/pull/3007

I have:

n.a.

### DIFF
--- a/design/one-pager-composition-environment.md
+++ b/design/one-pager-composition-environment.md
@@ -1,0 +1,267 @@
+# Patch from Generic Datasources
+
+* Owner: Maximilian Blatt (maximilian.blatt@accenture.com, @mistermx)
+* Reviewers: Crossplane Maintainers
+* Status: Draft
+
+## Background
+
+Crossplane currently supports patching from composite resources to composed
+resources and vice versa. It is also possible to copy values between two
+composed resources inside one composition by patching the value from one
+composed back to the composite and then to the second composed resource.
+
+However, Crossplane currently does not provide a way to patch from
+environment-dependent data sources. 
+Compositions behave the same regardless in which environment they are deployed.
+
+This becomes a big issue once you want to use the same composition in
+multiple environments. For example deploying AWS resources in two different 
+accounts requires different subnet IDs, VPC Ids, OIDC issuers etc.
+
+## Current workarounds
+
+### Build one composition per environment
+
+The simplest solution is to build one composition per environment.
+For example by using a helm chart that renders differently for each `values.yaml`.
+While this indeed works, it does not scale well with the number of environments
+and becomes increasingly more complex.
+
+### Use generic provider referencers
+
+While compositions itself are not environment-aware, some providers are by
+pulling data from generic resources like secrets. For example:
+
+* `provider-helm`: using [`spec.forProvider.valuesFrom.secretKeyRef`](https://doc.crds.dev/github.com/crossplane-contrib/provider-helm/helm.crossplane.io/Release/v1beta1@v0.10.0#spec-forProvider-valuesFrom-secretKeyRef)
+* `provider-kubernetes`: using [`spec.forProvider.references[].patchesFrom`](https://doc.crds.dev/github.com/crossplane-contrib/provider-kubernetes/kubernetes.crossplane.io/Object/v1alpha1@v0.3.0#spec-references-patchesFrom)
+
+However, none of these methods is native to Crossplane native and they are not
+available for every provider. Furthermore, they would require an additional
+managed resource (here `Object` or `Release`) to deploy the composed resource
+making the composition much more complex.
+
+## Goals
+
+The goal of this document is to define an API design for extracting environment
+specific data considering the following requirements:
+
+1. Allow the creation of generic compositions that render composed resources
+based on the environment they are executed in.
+2. Allows users to define this environment in Crossplane native way.
+3. Prevent escalation of priveleges by disallowing users to read out fields
+from objects they shouldn't have access to.
+
+## Proposed Solution
+
+Introduce a new Crossplane resource `EnvironmentConfig` that can store any 
+kind of  data similar to a K8s `ConfigMap` but supports complex data as well
+using the [`JSON` struct](https://pkg.go.dev/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1#JSON)
+which can store anything that serializes to valid JSON.
+
+```yaml
+apiVersion: apiextensions.crossplane.io/v1alpha1
+kind: EnvironmentConfig
+metadata:
+  name: example-environment
+data:
+  simple: value
+  int: 123
+  bool: false
+  complex:
+    a: b
+    c:
+      d: e
+  list:
+    - a
+    - b
+    - c
+```
+
+`EnvironmentConfig`s are referenced in the composition spec. They can be
+referenced directly as names or through labels in `spec.environment.environmentConfigs`.
+
+```yaml
+apiVersion: apiextensions.crossplane.io/v1
+kind: Composition
+metadata:
+  name: example
+spec:
+  # This feature will be introduced as alpha (behind a flag). Composition is a
+  # v1 API so we'll need to add a note that the environment block is alpha and
+  # may change without notice.
+  environment:
+    environmentConfigs:
+    # The list of the EnvironmentConfig selectors used to generate
+    # spec.environmentConfigRefs in the XR.
+    # These selectors are only executed if spec.environmentConfigRefs is null,
+    # similar to referencers in providers.
+    # 
+    # During reconcile, the data of the XR's EnvironmentConfigs is merged into
+    # a single 'computed' environment object. This computed environment
+    # is what all environment patches operate upon. For debugging, the computed
+    # environment written to the controller logs. Note: It is possible to expand
+    # this later and store the computed environment somewhere else, i.e. in the
+    # XR's status.
+    - type: Reference
+      reference:
+        name: example-environment
+    - type: Reference
+      reference:
+        name: other-environment
+    - type: Selector
+      selector: 
+      # Select one EnvironmentConfig with matching "stage" AND "my-label" labels.
+      - matchLabels:
+        # Matches an EnvironmentConfig whose "stage" label matches the value read
+        # from the Composite's field path.
+        - type: FromCompositeFieldPath
+          key: stage
+          valueFromFieldPath: spec.parameters.stage
+        # Matches an EnvironmentConfig whose "my-label" label matches the supplied
+        # static value.
+        - type: Value
+          key: my-label
+          value: metadata.labels[my-label]
+   # This is where we specify patches "between" the XR and (computed) environment
+    patches:
+    # A FromCompositeFieldPath patches from XR -> computed environment.
+    - type: FromCompositeFieldPath
+      fromFieldPath: spec.widgets
+      toFieldPath: widgets.count
+    # A FromEnvironmentFieldPath patches from computed environment -> XR.
+    - type: FromEnvironmentFieldPath
+      fromFieldPath: spec.widgets
+      toFieldPath: widgets.count
+```
+
+All found `EnvironmentConfig`s are merged together using strategic merging
+in the order they are listed. Similar to Helm value files.
+
+The selected environment configs refs are stored in the XR under
+`spec.resourceRefs`:
+
+```yaml
+apiVersion: demo.org/v1alpha1
+kind: XExample
+spec:
+  environmentConfigRefs:
+    - name: example-environment
+    - name: other-environment
+    - name: label-environment
+```
+
+Similar to referencers in providers, `environmentConfigRefs` is only going to be
+updated if it is null. Otherwise the list is going to be reused on consecutive
+reconciles.
+
+Composed resourced can be patched using the new `FromEnvironmentFieldPath` and
+`CombineFromEnvironment` patch types:
+
+```yaml
+        - type: FromEnvironmentFieldPath
+          fromFieldPath: key
+          toFieldPath: spec.forProvider.manifest.data.key
+        - type: CombineFromEnvironment
+          combine:
+            variables:
+              - fromFieldPath: key
+              - fromFieldPath: key
+          toFieldPath: spec.forProvider.manifest.data.key
+```
+
+It is also possible to patch the (in-memory) environment itself using 
+`ToEnvironmentFieldPath` and `CombineToEnvironment` patches:
+
+```yaml
+        - type: ToEnvironmentFieldPath
+          fromFieldPath: metadata.name
+          toFieldPath: tmp.name
+        - type: CombineToEnvironment
+          combine:
+            variables:
+              - fromFieldPath: metadata.namespace
+              - fromFieldPath: metadata.name
+          toFieldPath: tmp.namespacedName
+```
+
+**Advantages:**
+
+* Centralized solution that can be used with any provider.
+* API extensions are purely additive. No breaking changes.
+* No security risks through the usage of a new CRD. Composition writers cannot
+use this solution to extract data from resources they shouldn't have access to
+- except `EnvironmentConfig`, of course.
+* `EnvironmentConfig` is more powerful than standard `ConfigMap` since they
+can store complex data types instead of just `map[string]string`. This allows
+patching whole objects or arrays during compose.
+
+**Drawbacks:**
+
+* Further extending the patching API making it more complex.
+
+## Alternatives considered
+
+Some other solutions to patch from generic data sources:
+
+### Patch from any object
+
+Introduce a new `fromObjectFieldPath` patch type that can extract values from
+any object Crossplane has access to:
+
+```yaml
+      patches:
+        - type: FromObjectFieldPath
+          fromObjectRef:
+            apiVersion: v1
+            kind: ConfigMap
+            name: sample-config
+            namespace: sample-ns
+          fromFieldPath: data.value
+          toFieldPath: spec.forProvider.sampleField
+          policy:
+            fromFieldPath: Required # Dont render if referenced resource does not exist
+```
+
+See https://github.com/crossplane/crossplane/pull/2938 for more details.
+
+**Drawbacks:**
+
+* Further extends the patching API making it more complex.
+* Possible security issue by allowing escalation of privileges:
+A user without cluster admin access can create a composition to read out any value
+from an object they do not have access to.
+
+### Referencers on managed resource (MR) level
+
+Generic resource referencers could be implemented on MR level. Here every
+provider is responsible for implementing and supporting this feature.
+
+See https://github.com/crossplane/crossplane/issues/1770 for details.
+
+However, the security issues mentioned in [Patch from any Object](#patch-from-any-object)
+would occure here as well.
+One could potentially use a managed resource to extract data from a secret
+within another namespace.
+
+Additionally, this solution would require every provider to implement this
+seperately. It would therefore exclude every provider version that does not
+implement this feature.
+
+### Custom compositions
+
+Custom compositions are a proposed way of generating compositions on-the-fly
+using XRM functions which are similar to KRM function.
+
+See https://github.com/crossplane/crossplane/pull/2886/files for details.
+
+XRM functions are definitely a very powerful way to generate or modify
+compositions and might become the standard way of using compositions in the
+future.
+
+However, they also come with a huge increase in complexity compared to plain
+YAML which might be too high if you just want to patch some environment specific
+fields. For this use case, Crossplane should provide a more simpler method
+out-of-the-box. It would also allow teams to stick with their existing YAML
+compositions without having to migrate to XRM functions and a (potentially
+API breaking) composition v2.


madhureddy08 (approved):

design/one-pager-generic-datasource.md
```diff
@@ -0,0 +1,171 @@
...
+* Status: Draft
+
+## 1 Background
+
+Crossplane currently supports patching from composite resources to composed
+resources and vice versa. It is also possible to copy values between two
+composed resources inside one compositiion by patching the value from one
+composed back to the composite and then to the second composed resource.
+
+However, Crossplane currently does not provide a way to patch values from
```
muvaf, line 14: Could you expand this with more specifics, i.e. what is _dynamic_ resource? An example use case would also be great.

design/one-pager-generic-datasource.md
```diff
@@ -0,0 +1,171 @@
...
+
+## 3 Goals
+
+The goal of this document is to define a proposal for extracting environment
+specific data considering the following requirements:
+
+1. Allow the creation of generic compositions that are able to load environment 
+specific parameters during compose time.
+2. Allow users to define these values on cluster level.
+3. Prevent users from escalting their rights on the cluster and read out fields
```
muvaf, line 38:
```suggestion
3. Prevent users from escalating their rights on the cluster and read out fields
```

design/one-pager-generic-datasource.md
```diff
@@ -0,0 +1,171 @@
...
+Introduce a new Crossplane CRD `ValueSet` that can store arbitrary data similar
+to K8s `ConfigMaps` but supports more complex data types:
+
+```yaml
+---
+apiVersion: apiextensions.crossplane.io/v1alpha1
+kind: ValueSet
+metadata:
+  name: sample-value-set
+data:
```
muvaf, line 54: I think the proposed structure may not be flexible enough to address with field path. For example, will they use `data.complex.a`? What if there are deeply nested, i.e. the case where you have an array in a `complex` map?

One idea you may want to consider is to use `map[string]runtime.RawExtension` type where you can have an arbitrary YAML object in `RawExtension`. We [use](https://github.com/crossplane/crossplane/blob/79631b8/apis/apiextensions/v1/composition_types.go#L104) that for `resources[*].base` field in `Composition` type. This way anything that can be represented by YAML can be put there and would be addressable via field path (I haven't tried this but fairly sure).

design/one-pager-generic-datasource.md
```diff
@@ -0,0 +1,171 @@
...
+
+TODO: See https://github.com/crossplane/crossplane/pull/3007 for a draft
+implementation.
+
+**Advantages:**
+
+* API extensions are purely additive. No breaking changes.
+* No security risks through the usage of a new CRD. Composition writers cannot
+use this solution to extract data from resources they shouldn't have access to
+- expect `ValuesSets`, of course.
```
muvaf, line 99:
```suggestion
- expect `ValueSet`, of course.
```

design/one-pager-generic-datasource.md
```diff
@@ -0,0 +1,171 @@
...
+* `ValueSet` is more powerfull than standard `ConfigMaps` since they can store
```
muvaf, line 100:
```suggestion
* `ValueSet` is more powerfull than standard `ConfigMap` since they can store
```

design/one-pager-generic-datasource.md
```diff
@@ -0,0 +1,171 @@
...
+
+## 4 Proposed Solution
+
+Introduce a new Crossplane CRD `ValueSet` that can store arbitrary data similar
+to K8s `ConfigMaps` but supports more complex data types:
+
+```yaml
+---
+apiVersion: apiextensions.crossplane.io/v1alpha1
+kind: ValueSet
```
muvaf, line 51: I think it's always good to keep the naming close to the upstream so that people would know what it does immediately. I'm wondering if we can come up with something closer to upstream rather than a complete new name.

The concept is very similar to `ConfigMap` except that it's common to the whole cluster and every composition author can access it. I don't have any suggestion that I feel strongly but some possible ones could be `ConfigSet`, `CommonConfigMap`, `ConfigMap` (same name, its own group), `CommonMap`, `ClusterConfigMap`.

design/one-pager-generic-datasource.md
```diff
@@ -0,0 +1,171 @@
+# Patch from Generic Datasources
+
+* Owner: Maximilian Blatt (maximilian.blatt@accenture.com, @mistermx)
+* Reviewers: Crossplane Maintainers
+* Status: Draft
+
+## 1 Background
```
muvaf, line 7:
```suggestion
## Background
```

Since the design docs and one pagers usually aren't very long, we don't really use index numbers.

design/one-pager-generic-datasource.md
```diff
@@ -0,0 +1,171 @@
...
+2. Allow users to define these values on cluster level.
+3. Prevent users from escalting their rights on the cluster and read out fields
+from objects they shouldn't have access to.
+4. (Bonus) Allow package writers to include default values in configuration
+packages.
+
+## 4 Proposed Solution
+
+Introduce a new Crossplane CRD `ValueSet` that can store arbitrary data similar
+to K8s `ConfigMaps` but supports more complex data types:
```
muvaf, line 46:
```suggestion
to K8s `ConfigMap` but supports more complex data types:
```

muvaf (commented):
@MisterMX Thanks! I think this is a great way to solve the common config problem.

design/one-pager-generic-datasource.md
MisterMX: I am not sure how flexible `RawExtension` actually is because the data gets parsed into a `runtime.Object`. I am not sure if you can store primitive values in there.

In either case, the [example implementation[(https://github.com/crossplane/crossplane/pull/3007/files#diff-755442f10dbfe9f790b1843c9f660718dcb1bff01a151c41d5ecc74f33cdf6dbR23) mentioned uses the `JSON` struct which allows storing everything that is valid JSON, including complex paths like `data.complex.a`.

design/one-pager-generic-datasource.md
MisterMX: I thought `ValueSet` to be a good choice because it just contains an arbitrary set of values defined by the user. Other than that `ConfigSet` would work for me as well.

design/one-pager-generic-datasource.md
MisterMX: Removed the indices.

design/one-pager-generic-datasource.md
MisterMX: I refer to everything that is hardcoded in the composition as *static*. This includes patches between the composite and the composed resources. Its *static* because they behave the same regardless of the environment.

By *dynamic* I mean environment-dependent values. I.e. deploying AWS resources in two different VPCs requires different subnet IDs, VPC Ids, OIDC issuers etc.

design/one-pager-generic-datasource.md
```diff
@@ -0,0 +1,186 @@
...
+    - a
+    - b
+    - c
+```
+
+Extend the composition CRD to allow patching from `ConfigSet`:
+
+```yaml
+        - type: FromConfigSetFieldPath
+          fromConfigSetRef:
```
muvaf, line 83: I'm a little bit on the fence about how to represent the reference. If we draw some similarities to `Combine` functionality, an alternative could be like the following:
```yaml
- type: FromConfigSet
  configSet:
    ref:
      name: sample-config-set
    fromFieldPath: data.key
  toFieldPath: spec.forProvider.manifest.data.static
  policy:
    fromFieldPath: Required
```
or keep using the same `fromFieldPath` at the top level.
```yaml
- type: FromConfigSet
  configSet:
    ref:
      name: sample-config-set
  fromFieldPath: data.key
  toFieldPath: spec.forProvider.manifest.data.static
  policy:
    fromFieldPath: Required
```
closer to the existing one
```yaml
- type: FromConfigSet
  configSetRef:
    name: sample-config-set
  fromFieldPath: data.key
  toFieldPath: spec.forProvider.manifest.data.static
  policy:
    fromFieldPath: Required
```

I'm essentially looking for ways to reduce the repetition of the words while keeping the structs contained to themselves and flexible enough for future iterations. @negz what are your thoughts?

MisterMX: Good thought. I also felt that the design is pretty repetitive.

I would propose referencing the `ConfigSets` at the top of the Composition's spec:

```yaml
kind: Composition
metadata:
  name: cnp-virtualcontrolplane
spec:
  environment:
    configSetRefs:
      - name: sample-config-set
    # Alternatively, get it via labels
    configSetSelectors:
      - matchLabels:
          stage: prod
```

All found `ConfigSet`s are merged into one using strategic merging. Similar to Helm values.

Then you would just have to change the `type` in your patches:

```yaml
        - type: FromConfigSetFieldPath
          fromFieldPath: data.key
          toFieldPath: spec.forProvider.manifest.data.static
          policy:
            fromFieldPath: Required # Dont render if referenced field does not exist
```

design/one-pager-generic-datasource.md
MisterMX: Makes sense. I dropped `dynamic` and `static` and use `environment-dependent patching` instead.

design/one-pager-generic-datasource.md
muvaf: Once we want to use more than a singe configset ref, we'd need to make `environment` able to hold more than one, then I guess we'd have to ask which one to use under the patch instance, right? I feel like having `the identification of which ConfigSet to use` contained in the patch instance could allow us to be more flexible and less intrusive of other fields/features of `Composition`.

I feel like the following is the most similar to existing API conventions we have:
```yaml
- type: FromConfigSet # Or FromConfigSetToComposed? FromConfigSetToComposite in the future if users ask for it. Maybe FromConfigSet
  configSetRef:
    name: sample-config-set
  fromFieldPath: data.key
  toFieldPath: spec.forProvider.manifest.data.static
  policy:
    fromFieldPath: Required
```

muvaf: The similarity between `PatchSet` word used in `Composition` and the `ConfigSet` is something I just realized, though not sure if it's necessarily a bad thing.

design/one-pager-generic-datasource.md
```diff
@@ -0,0 +1,189 @@
...
+data:
+  key: value
+  complex:
+    a: b
+    c: d
+  list:
+    - a
+    - b
+    - c
+```
```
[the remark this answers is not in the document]
negz: Apologies for arriving late with this suggestion - I see you've already gone through one round of discussing naming. 😄

negz (commented):
Thanks for opening this @MisterMX! I'm really pleased to see this coming together in design doc form, and I like the approach.

design/one-pager-generic-datasource.md
```diff
@@ -0,0 +1,189 @@
...
+2. Allow users to define these values on cluster level.
+3. Prevent users from escalating their rights on the cluster and read out fields
+from objects they shouldn't have access to.
+4. (Bonus) Allow package writers to include default values in configuration
+packages.
+
+## Proposed Solution
+
+Introduce a new Crossplane CRD `ConfigSet` that can store arbitrary data similar
+to K8s `ConfigMap` but supports complex data using the [`JSON` type](https://pkg.go.dev/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1#JSON) that can be anything that is valid JSON:
```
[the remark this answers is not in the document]
negz: @liggitt just reminded me that the whole point of Secrets and ConfigMaps was to expose their data as files and/or env vars. Given we don't have that constraint here supporting complex data makes sense to me.

design/one-pager-generic-datasource.md
```diff
@@ -0,0 +1,189 @@
...
+    configSetRefs:
+      - name: sample-config-set
+    # Alternatively, get it via labels
+    configSetSelectors:
+      - matchLabels:
+          stage: prod
+```
+
+All found `ConfigSet`s are merged into one using strategic merging. Similar to
+Helm values.
```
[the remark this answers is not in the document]
MisterMX: I did not yet take deeper look into that but I guess the easiest solution would be marshal everything into JSON and then use a library that can do the merging, such as [the4thamigo-uk/conflate](https://github.com/the4thamigo-uk/conflate).

design/one-pager-generic-datasource.md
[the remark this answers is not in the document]
MisterMX: > Can you outline some use cases for storing complex data, rather than a map[string][]byte?

For example, in our case, we want to patch a list of subnet IDs directly into a resource without picking them one by one. Additionally, have a hierarchical structure allows you to order stuff that belongs together, i.e `aws.subnetsIds`  and `azure.subnetIds` instead of `aws_subnet_1` ,`azure_subnet_2` etc.

> it does have a somewhat more predictable, self-documenting schema

My idea behind `ConfigSets`s it is to _not_ restrict the user with a schema but rather let him chose the data structure on his own.

design/one-pager-generic-datasource.md
```diff
@@ -0,0 +1,189 @@
...
+The goal of this document is to define a proposal for extracting environment
+specific data considering the following requirements:
+
+1. Allow the creation of generic compositions that are able to load environment 
+specific parameters during compose time.
+2. Allow users to define these values on cluster level.
+3. Prevent users from escalating their rights on the cluster and read out fields
+from objects they shouldn't have access to.
+4. (Bonus) Allow package writers to include default values in configuration
+packages.
```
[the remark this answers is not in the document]
MisterMX: I agree that providing default values is just an optional feature that can be included a at later time if needed.

design/one-pager-generic-datasource.md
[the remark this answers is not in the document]
MisterMX: @negz I rewrote the [draft implementation](https://github.com/crossplane/crossplane/pull/3007/files#diff-5d7910f2ba87282df3f14400d5ec4d39c262e3b7b444b7889c9887c2dbf92162R104) and it turns out, merging is actually pretty straightforward.

design/one-pager-generic-datasource.md
MisterMX: I updated the [draft implementation](https://github.com/crossplane/crossplane/pull/3007/files#diff-5d7910f2ba87282df3f14400d5ec4d39c262e3b7b444b7889c9887c2dbf92162R104) that includes merging all referenced `ConfigSets` into one and make the patch agnostic of the source. This feels the most natural to me and I would propose to go forward with this.

design/one-pager-generic-datasource.md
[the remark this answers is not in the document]
negz: Sounds good - let's proceed with unstructured complex data.

design/one-pager-generic-datasource.md
negz: @MisterMX I'm onboard with merging referenced ConfigSets into one set of data (e.g. when the `environmentConfigSelector` matches multiple `EnvironmentConfigs`). I think that's a related but separate concern from whether we'd want to let folks build multiple named 'config sets' within a Composition.

Allowing multiple named config sets is probably unnecessary complexity when each set can be merged from multiple inputs, but I'm tempted to push for it anyhow mostly because it makes the API work very similarly to the existing "patch sets" functionality.

design/one-pager-generic-datasource.md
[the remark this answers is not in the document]
negz: Sounds good - in that case I like the idea of merging. Do you have any idea how you'll handle conflicts?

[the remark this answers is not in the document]
MisterMX: I would merge them together in the order they are listed and let values be overwritten by ConfigSets of higher order.

design/one-pager-generic-datasource.md
MisterMX: >  (e.g. when the environmentConfigSelector matches multiple EnvironmentConfigs)

Do we want to select multiple ConfigSets through one label selector? I think this might be untransparent to the user because the order of items returned by `kube.List()` can be quite different. I would suggest to only select the first and require the user to make another selection if he needs more.

negz: > Do we want to select multiple ConfigSets through one label selector?

Oh, I got mixed up and thought this was what you had been proposing. I agree that it could be confusing.

So would something like this be what you were thinking:

```yaml
kind: Composition
metadata:
  name: example
spec:
  patchSets:
  - name: metadata
    patches:
      fromFieldPath: metadata.labels[some-important-label]
  configSets:
    # The 'name' field is now just informational, to distinguish entries.
    # We could omit it if we wanted to.
  - name: staging
    environmentConfigRef:
    - name: staging
  - name: west-coast
    environmentConfigRef:
    - name: west-coast
  resources:
  - name: example
    base:
      apiversion: example.org/v1
      kind: Foo
      spec:
        forProvider:
          bars: 42
    patches:
    - type: PatchSet
      patchSetName: metadata
    # FromConfigSetFieldPath _only_ patches from ConfigSet -> Composed resource.
    # It doesn't make sense for patching from ConfigSet -> Composite resource to
    # be associated with a particular composed resource template.
    - type: FromConfigSetFieldPath
      fromFieldPath: data.bazes
      toFieldPath: spec.forProvider.bazes
      policy:
        fromFieldPath: Required
  # Introduce a new top-level 'composite' field. This field can be used to
  # mutate the composite resource in ways that don't have anything to do with
  # a particular composed resource. In theory we could avoid this by patching
  # from configSet -> composed -> composite but that seems awkward.
  composite:
    patches:
    - type: FromConfigSetFieldPath
      fromFieldPath: data.bangs
      toFieldPath: status.bangs
    # We could move this recently added alpha field here too
    publishConnectionDetailsWithStoreConfigRef:
      name: default
```

Note that there's no longer a reference to a particular `configSetName`, it just references the merged data.

MisterMX:
I made a few more adjustments to the design:

* Renamed `ConfigSet` to `CompositionEnvironment` to indicate its connection to `Composition`
* Moved environment selection to `spec.environment.sources`

There are one open point that need to be solved:

#### Patching the composite

@negz you proposed to add a new field `spec.composite.patches` to hold a list of patches just like in resources.

I am not so sure about this anymore. How would this be implemented? I am not sure if you can reuse the existing `Patch` type because you would have to handle the patch types differently: Default behavior is patching to composed, but in this case there is none. Do you want to create a new `CompositePatch` type? The patch logic is pretty hardwired to the `Patch` type. You would have to rewrite a lot - and test it.

I see two alternatives to solve this:

1. Extend the list of existing patch types: `FromEnvironmentToComposite` and `CombineFromEnvironmentToComposite`.
  The changes to the API are absolutely minimal, only a few lines of code need to be changed. The control flow would not be changed since they are just regular patches.
2. Introduce a new list of patches to `spec.environment.patches` that are run before the resource patches. In this case, the "resource" to be patched would be the Environment.

@negz do you think we can move on with one of this solutions?

design/one-pager-generic-datasource.md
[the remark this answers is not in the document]
negz: Bumping this one. I continue to prefer `EnvironmentConfig` to `CompositionEnvironment`. I think being in the `apiextensions.crossplane.io` API group is sufficient to link it to a `Composition` conceptually.

design/one-pager-generic-datasource.md
negz: I'm going to resolve this thread and respond to https://github.com/crossplane/crossplane/pull/3008#issuecomment-1135756546, which touches on the same topic.

negz:
> How would this be implemented?

@MisterMX I'm empathetic to the fact this might require a lot of code changes, but we should prioritize API design here. We have more flexibility to change code in future; we can't change the API once we've committed to something (until 2.x).

> Extend the list of existing patch types: FromEnvironmentToComposite and CombineFromEnvironmentToComposite.

Where would these patch types be used? It seems odd to have to associate them with an element in the `resources` array when in fact they don't pertain to that element at all.

> Introduce a new list of patches to spec.environment.patches that are run before the resource patches. In this case, the "resource" to be patched would be the Environment.

How does this differ from the `spec.composite.patches` approach? I'm probably missing something but it seems like it would just be the same thing at a different location within the document.

MisterMX:
> Where would these patch types be used? It seems odd to have to associate them with an element in the resources array when in fact they don't pertain to that element at all.

You are right. I feel the same thinking about this. I just made a list of possible solutions I could think of. I would prefer 2, personally.

> How does this differ from the spec.composite.patches approach? I'm probably missing something but it seems like it would just be the same thing at a different location within the document.

My idea behind this is to clarify which object (composite or environment) is "from" and "to".

The term `spec.composite.patches` implies to me that I am patching *to* the composite by default. This does not play nicely with the existing `FromComposite*` and `ToComposite*` patch types. So it would require a new Patch Struct that has a different type enum. (Yes, most of the code would be copy paste, but it still increases the code size, making things more complicated.)

`spec.environment.patches` allows reusing the existing patch types because we can just replace the composed resource with the environment. If the user wants to patch the composite then he can do it with `ToCompositeFieldPath` just as he is to used to from regular resource patching. 

Furthermore the user could use the environment to store temporary values during the rendering process.

(And it would save us from rewriting/extending the patch API)

design/one-pager-generic-datasource.md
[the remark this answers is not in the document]
MisterMX: `EnvironmentConfig` is fine for me 👍

negz:
> The term spec.composite.patches implies to me that I am patching to the composite by default. This does not play nicely with the existing FromComposite* and ToComposite* patch types.

Good point.

> So it would require a new Patch Struct that has a different type enum.

Unfortunately I'm currently liking this approach best, despite agreeing with the code level downsides you point out.

Even with the `spec.environment.patches` option it seems odd that we'd presumably default to patching "`FromCompositeFieldPath`" to "the environment". What would that mean? Would the patch be mutating the actual `EnvironmentConfig` custom resource?

MisterMX:
> Even with the spec.environment.patches option it seems odd that we'd presumably default to patching "FromCompositeFieldPath" to "the environment".

I think that depends on the use case. If users use the composition environment primarily as a temporal data storage to pre-process variables it makes sense to patch to the environment by default.

>> So it would require a new Patch Struct that has a different type enum.

> Unfortunately I'm currently liking this approach best, despite agreeing with the code level downsides you point out.

Thinking about it more, it might actually be necessary. In this we could think about sharing the patch structs for `Composition` and `CompositionRevision` to avoid redundant code. This lead to some issues in the past, i.e #2836.

> Would the patch be mutating the actual EnvironmentConfig custom resource?

I would handle it like shell variables passed to subprocesses: A container in a pod cannot modify a ConfigMap by changing the environment variable.

MisterMX:
Another use case appeared during more testing:

In the current version of the design, the environment(s) would be referenced like so:
```yaml
spec:
  environment:
    sources:
      - selector:
          matchLabels:
            my-label: static
```

What if the environment should be picked by something provided by the claim, for example a label `stage: test`? It would require some sort of extraction before loading the environment.

This could be handled by letting the user define a set of patches:
```yaml
    sources:
      - selector:
          matchLabels:
            stage: # patched
        patches:
          - fromFieldPath: metadata.labels.stage
            toFieldPath: selector.matchLabels.stage
```

Feggah (commented):
Good job writing this design doc, @MisterMX !

I think the proposed solution is **way** more complex when compared to "Patch from any object". 

The drawback in "Patch from any object" is really a problem? I don't think users (platform clients) should have enough RBAC to create Compositions, in general, only platform owners should have this permission -- and platform owners would probably have admin permissions or something really close to it in the control plane cluster.

The learning curve to understand about XRM is already really complex, I think this is a feature that will be widely used by platform owners and the "Patch from any object" solution is more straight forward in my opinion. It's easier to understand

MisterMX:
@Feggah "Patch from any object" is not that simple. While patching from a static object is trivial, it becomes much more complex once the object name and labels are parameterized and passed via the claim. Furthermore, you would have to do the object selection in every patch instead of fetching just one EnvironmentConfig. This is bad performance- and design-wise.

shanproofpoint:
hi can i get a status of this? this would be really useful. thank you!

negz:
> I would handle it like shell variables passed to subprocesses: A container in a pod cannot modify a ConfigMap by changing the environment variable.

That sounds reasonable.

> In this we could think about sharing the patch structs for Composition and CompositionRevision to avoid redundant code. This lead to some issues in the past, i.e https://github.com/crossplane/crossplane/issues/2836.

Yeah, this has been hurting us a bunch e.g. https://github.com/crossplane/crossplane/pull/3158. We still need code to sync between Compositions and CompositionRevisions too. Ideally I'd prefer to keep separate types and generate code to keep them in sync, though.

> In the current version of the design, environment referencers are static

A special selector type might work here too, something like:

```yaml
spec:
  environment:
    sources:
      - selector:
          # Match environments with a my-label label that matches the value of the XR's my-label label.
          matchCompositeLabels:
          - my-label

**Comment by negz** [comment]
> How are configs grabbed via environmentConfigRefs and environmentConfigSelectors merged together?

It seems we might have different understandings of how this would work. Are you thinking that selectors would select some environment configs and refs would reference some others?

I instead imagine this working the same way existing selector/ref pairs do in Crossplane, which is modeled after the node selector / ref on pods. The only function of a selector is to select and set refs. This means the only environment configs used are the referenced ones. Typically once the refs are set we ignore selectors on future reconciles (you can clear the refs if you want the selectors to be re-processed).

Does that help?

> Regarding storing the computed environment in the XR

I thought our proposal was to store the computed environment on the _Composition_, not the XR. This is why I said it probably doesn't makes sense, since it's unique to the XR+Composition combination.

Storing the environment on the XR on the other hand does seem feasible, but I'm a little wary of each XR having a potentially quite large amount of extra information in its status, especially given that some of that information might not actually be relevant to/used by the XR.

Emitting the computed environment config as a debug log seems like a good start for now. I'm not too worried about log flooding given that it would only be enabled when debug logging was enabled.

**Comment by MisterMX** [comment]
> It seems we might have different understandings of how this would work. Are you thinking that selectors would select some environment configs and refs would reference some others?

@negz I don't think that we are that far away from each other. I, too, think that the list of resource refs should be computed once and reused later until the ref field is reset.

However, I would store the references in the XR, rather than the Composition because the list of references is unique to the XR+Composition combination.

My only concern is how the composition writer is able to reference environments. I would give him the ability to refer by name or labels individually. The computed reference list should then be stored in the XR.

In other words, this
```yaml
# composition.yaml
spec:
  environment:
    environmentConfigs:
    - type: Reference
      reference:
        name: static-env
    - type: Selector
      selector:
      - matchLabels:
        - type: Value
          key: my-label
          value: metadata.labels[my-label] // selects dynamic-env
    - type: Reference
      reference:
        name: another-static-env
```

Should resolve to something like this:

```yaml
# xr.yaml
spec:
  environmentConfigRefs:
    - name: static-env
    - name: dynamic-env
    - name: another-static-env
```

Does this make sense?

> I thought our proposal was to store the computed environment on the Composition, not the XR. This is why I said it probably doesn't makes sense, since it's unique to the XR+Composition combination.
>
> Storing the environment on the XR on the other hand does seem feasible, but I'm a little wary of each XR having a potentially quite large amount of extra information in its status, especially given that some of that information might not actually be relevant to/used by the XR.
>
> Emitting the computed environment config as a debug log seems like a good start for now. I'm not too worried about log flooding given that it would only be enabled when debug logging was enabled.

I agree with you. For now it should be enough to log the environment. Maybe we can find a solution once we have gathered some experience. Maybe we can hide the environment behind an alpha feature flag, for example `--debug-dump-composition-environment`?

shanproofpoint:
I really appreciate the discourse here!  I understand it is safer to restrict it to a crossplane native object, but how can we add to the EnvironmentConfig dynamically ? I guess your answer to that is inject a few precomputed permutations..

btw an XR can result in a really large composition. In order to break the composition down in stages to not make a mess, i produced intermediate XRs that have less data in them from the original XR. Breaking things down into smaller compositions allow faster debug cycle for that next layer of resources.  I would like to somehow not copy any data and reference the original XR from the new XR+Composition.  EnvironmentConfig wont allow me to do that, but referencing a generic resource would.

negz:
> I agree with you. For now it should be enough to log the environment. Maybe we can find a solution once we have gathered some experience.

💯 

> Maybe we can hide the environment behind an alpha feature flag, for example --debug-dump-composition-environment?

Like any new feature I think this whole thing will need to be introduced at v1alpha1 and behind a feature flag like `--enable-composition-environments` or something. I think that one flag will be sufficient at least while it's in alpha.

negz:
> However, I would store the references in the XR, rather than the Composition because the list of references is unique to the XR+Composition combination.

As yes, good point I hadn't considered that. Having the selectors on one object (the composition) and the pinned references on another (the XR) is starting to feel a little more complex than I'd like.

Could we perhaps get away with just making all environment selection the XR's responsibility? That is, all selectors and refs would exist at the XR level? We could support default or enforced Environments at the XRD level, similar to how today we support default or enforced Compositions at the XRD level.

One thing we'd lose with the approach I'm proposing here would be the ability to 'automatically' derive an environment from a particular field path within an XR in a way that's largely transparent to the XR author.

negz:
> I understand it is safer to restrict it to a crossplane native object, but how can we add to the EnvironmentConfig dynamically ? I guess your answer to that is inject a few precomputed permutations..

@shanproofpoint Can you expand on what you mean by "dynamically" in this context? Under this design it is possible to update the EnvironmentConfig over time, and those updates would be reflected on subsequent XR reconciles.

> btw an XR can result in a really large composition. In order to break the composition down in stages to not make a mess, i produced intermediate XRs that have less data in them from the original XR. Breaking things down into smaller compositions allow faster debug cycle for that next layer of resources. I would like to somehow not copy any data and reference the original XR from the new XR+Composition. EnvironmentConfig wont allow me to do that, but referencing a generic resource would.

Can you raise an issue describing this use case? I'm not sure I want to bring it into scope for this particular design but it does sound interesting.

MisterMX:
> Could we perhaps get away with just making all environment selection the XR's responsibility? That is, all selectors and refs would exist at the XR level? We could support default or enforced Environments at the XRD level, similar to how today we support default or enforced Compositions at the XRD level.

I think environment selection should be part of the composition because it depends on the implementation which environment values are used. For example, you might need different EnvironemntConfigs for Azure and AWS compositions.

I don't think it belongs to the XRD either because the XRD only defines the frontend API for the user who usually does not know about composition internals.

> Like any new feature I think this whole thing will need to be introduced at v1alpha1 and behind a feature flag like --enable-composition-environments or something. I think that one flag will be sufficient at least while it's in alpha.

Works for me. Lets go forward with that.

MisterMX:
I rebased the branch and added a section about storing `spec.environmentConfigRefs` in the XR.

@negz do you feel confident about the design? I don't have anything to add from my side.

design/one-pager-composition-environment.md
```diff
@@ -0,0 +1,266 @@
...
+    - type: Reference
+      reference:
+        name: example-environment
+    - type: Reference
+      reference:
+        name: other-environment
+    # Each selector in this array matches at most one EnvironmentConfig. If a
+    # selector matches multiple we pick the first. (This is how cross resource
+    # ref selectors work.) Each matching EnvironmentConfig is 'pinned' as a
+    # reference below. Selecting is optional; you can just reference one or more
```
negz, line 108: I think we now pin these to the list of references on an XR, not "below" on the Composition.

negz (approved):
Two nits where the document is out-of-date relative to what we've discussed in the comments, but in general I approve. Thanks for persisting with this for so long @MisterMX!

design/one-pager-composition-environment.md
MisterMX: That's corrent. I removed this section.

design/one-pager-composition-environment.md
```diff
@@ -0,0 +1,266 @@
...
+  name: example
+spec:
+  # This feature will be introduced as alpha (behind a flag). Composition is a
+  # v1 API so we'll need to add a note that the environment block is alpha and
+  # may change without notice.
+  environment:
+    environmentConfigs:
+    # The EnvironmentConfigs this Composition will use. The data of these
+    # EnvironmentConfigs is merged into a single 'computed' environment object,
+    # that is written to status.environment.computed. This computed environment
```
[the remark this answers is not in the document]
MisterMX: Yes, the description is outdated. I rewrote it and added a note that the list of used environments is stored in XR's `spec.environmentConfigRefs`.

shanproofpoint:
> > I understand it is safer to restrict it to a crossplane native object, but how can we add to the EnvironmentConfig dynamically ? I guess your answer to that is inject a few precomputed permutations..
> 
> @shanproofpoint Can you expand on what you mean by "dynamically" in this context? Under this design it is possible to update the EnvironmentConfig over time, and those updates would be reflected on subsequent XR reconciles.

I misunderstood sorry, btw having EnvironmentConfig be dynamically updated may resulted in harder to follow compositions if the value were not easily traceable. 

Ok then given we can update the EnvironmentConfig at will:

How did EnvironmentConfig come into existence? When I allocate a set of clusters, each cluster gets its own configmap that is unique. This configmap is generated from the XR. I would then need to translate that into a EnvironmentConfig having no trace in github at all. This is done for DRY.  This means I have to generate EnvironmentConfig using crossplane as a first superficial layer before choosing the actual patches in another layer using it.  In fact I would always need crossplane first before I am able to gather data from other objects using kubernetes provider into this EnvironmentConfig.

negz:
> Having EnvironmentConfig be dynamically updated may result in harder to follow compositions if the value were not easily traceable

Agreed. I'm a little worried about this but happy to move to alpha and try it out. We can always change the feature if it proves to be a problem.

> How did EnvironmentConfig come into existence?

I imagine the same way as anything else - e.g. by hand, by a gitops pipeline, composed as part of an XR, etc.

> When I allocate a set of clusters, each cluster gets its own configmap that is unique. This configmap is generated from the XR. I would then need to translate that into a EnvironmentConfig having no trace in github at all. 

Could you just create an EnvironmentConfig instead of a ConfigMap?

shanproofpoint:
> > Having EnvironmentConfig be dynamically updated may result in harder to follow compositions if the value were not easily traceable
> 
> Agreed. I'm a little worried about this but happy to move to alpha and try it out. We can always change the feature if it proves to be a problem.
> 
> > How did EnvironmentConfig come into existence?
> 
> I imagine the same way as anything else - e.g. by hand, by a gitops pipeline, composed as part of an XR, etc.
> 
> > When I allocate a set of clusters, each cluster gets its own configmap that is unique. This configmap is generated from the XR. I would then need to translate that into a EnvironmentConfig having no trace in github at all.
> 
> Could you just create an EnvironmentConfig instead of a ConfigMap?

I guess I was hesitant due to the amount of change on our side lmao. But you are correct,  starting with a EnvironmentConfig objects would be more self descriptive than a bunch of kubernetes primitives cobbled together.

shanproofpoint:
I believe the EnvironmentConfig would still not allow us to merge variable number of them into a composition. the composition would still need to reference envConfig1, envConfig2 vs forEachNewClusterDiscovered add them into the gcp.BackendService

PR #3029, opened 2022-04-06
Add One Pager for Pluggable Secret Stores

### LINKED ISSUE #2993: Proposal: Implement a plugin mechanism for out-of-tree External Secret Stores

We recently [merged](https://github.com/crossplane/crossplane/pull/2940) External Secret Stores as an _alpha_ feature with an initial support of [Vault](https://www.vaultproject.io/) as a Secret Store. As proposed in [the design doc](https://github.com/crossplane/crossplane/blob/master/design/design-doc-external-secret-stores.md), this was implemented _in-tree_ in Crossplane Runtime as a first step. 

However, this implementation introduces [additional dependencies](https://github.com/crossplane/crossplane-runtime/pull/322/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R52) in Crossplane Runtime and would add more as we add more Secret Stores. We initially chose this approach to avoid the upfront complexity of a [plugin mechanism](https://github.com/crossplane/crossplane/blob/master/design/design-doc-external-secret-stores.md#out-of-tree-support-with-a-plugin-api) and wanted to focus on the feature itself, however, with the gained confidence, we can now iterate with a more scalable option.

Consider building a [plugin mechanism](https://github.com/crossplane/crossplane/blob/master/design/design-doc-external-secret-stores.md#out-of-tree-support-with-a-plugin-api) that would enable adding out-of-tree Secret Stores without introducing additional dependencies to Crossplane Runtime.

turkenh:

This PR adds a one-pager design document for Pluggable Secret Stores extending the [proposal issue](https://github.com/crossplane/crossplane/issues/2993).

Fixes #2993 

I have:

N/A


luebken (dismissed):
I'm not able to comment on the implementation too much. Maybe just ensuring compatibility / similarity to other extension mechanism. 

As for the background and goals. They look good. 

For user feedback we could sketch out an example on how swapping a secret store could look like. But I'm not sure if that would be informative.

Thanks a lot.

design/one-pager-pluggable-secret-stores.md
```diff
@@ -0,0 +1,233 @@
...
+
+The goal of this document is to propose a pluggable secret store mechanism
+that enables the development and consumption of out-of-tree secret store
+plugins.
+
+This new mechanism should:
+- Allow building/using new secret stores for Crossplane without making any
+changes in Crossplane and/or Provider repositories.
+- Provide a smooth development and consumption experience. 
+- Have secure communication with the plugin process.
```
negz, line 34: Do we need to support more than one kind of external secret store being deployed?

design/one-pager-pluggable-secret-stores.md
```diff
@@ -0,0 +1,233 @@
...
+
+## Proposal
+
+This document proposes introducing a plugin mechanism for communicating with
+External Secret Stores. We will follow a similar approach as [KMS plugins] used
+by Kubernetes API Server for encryption at rest. A Secret Store plugin will
+be serving a well-defined API via gRPC. For the sake of security, plugins will
+be deployed as sidecars to Crossplane and Provider pods and communication will
+be over Unix domain sockets.
```
negz, line 43: It may be worth adding a little information about how two containers within a pod can communicate via a Unix domain socket - I had to lookup some documentation to refresh my memory. I'm pretty sure it doesn't require any special privileges, right? The two containers just need to share the volume where the socket exists.

design/one-pager-pluggable-secret-stores.md
```diff
@@ -0,0 +1,233 @@
...
+metadata:
+  name: vault
+spec:
+  type: External
+  defaultScope: crossplane-system
+  external:
+    configRef:
+      apiVersion: secrets.crossplane.io/v1alpha1
+      kind: VaultConfig
+      name: vault-internal
```
negz, line 75: If we need to support multiple external secret stores, will we need some way to tell Crossplane which one of them to connect to?

I could imagine adding an `address` field to the `external` block here which would default to a well known unix domain socket path, but which could be overridden to point to a different socket (maybe even TCP/IP in future?).

Something like this:

```yaml
 apiVersion: gcp.crossplane.io/v1alpha1
 kind: StoreConfig
 metadata:
   name: vault
 spec:
   type: External
   address: unix:///crossplane/secret/default.sock  # The default, can be changed.
   external:
     configRef:
       apiVersion: secrets.crossplane.io/v1alpha1
       kind: VaultConfig
       name: vault-internal
```

negz (commented):
This looks great - I'm pretty much ready to approve but I want to align on whether we support multiple external secret stores or not first.

design/one-pager-pluggable-secret-stores.md
turkenh: As mentioned somewhere below, 

```
A _Secret Store Plugin_:

- May support one or more Secret Stores.
- ...
```

All `rpc` methods have configReference as their first parameter. So, one secret store plugin may support more than one Secret Store type, e.g. Vault, AWS, GCP, Azure ... all in a single plugin binary, and related implementation could be called depending on a given config. 

Added a new section with the title: **Plugin Binary vs Secret Store Implementation**.

design/one-pager-pluggable-secret-stores.md
turkenh: Please see https://github.com/crossplane/crossplane/pull/3029#discussion_r866559614

design/one-pager-pluggable-secret-stores.md
turkenh: Added a new section with the title: `Communication over Unix Domain Sockets`.

turkenh:
@negz thanks for the review. 
Addressed your comments and this is now ready for another pass.

negz:
> [negz](https://github.com/negz) requested a review from [luebken](https://github.com/luebken) [32 seconds ago](https://github.com/crossplane/crossplane/pull/3029#event-6696122307)

Sorry @luebken - I didn't mean to request a new review.

design/one-pager-pluggable-secret-stores.md
```diff
@@ -0,0 +1,269 @@
...
+one Secret Store Plugin may support more than one Secret Store implementation,
+e.g. Vault, AWS, GCP, Azure ... all in a single Plugin binary, and related
+implementation could be called depending on a given config similar to how it is
+being done in [RuntimeStoreBuilder] today. Implementation of individual Secret
+Stores still need to satisfy the same [Store] interface today and could live in
+separate repositories. However, they would need to be packaged and served by a
+Secret Store Plugin binary to be used with Crossplane.
+
+This approach allows us to support more than one Secret Store implementation
+with a single plugin sidecar simplifying the deployment model.
```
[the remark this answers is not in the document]
turkenh: I was imagining each secret store implemented and released separately supporting only one type.
Then, there could be another wrapper project(s), that imports existing secret stores and supports multiple. Not perfect but I feel like a good optimization to simplify the deployment model.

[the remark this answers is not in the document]
negz: Is there a reason we couldn't just deploy multiple secret stores each using a different path for their API socket (e.g. https://github.com/crossplane/crossplane/pull/3029/files#r862208395)? Is there a reason to prefer this wrapper approach?

bassam:
This is great @turkenh - and big +1 for deprecating and removing the in-tree Vault implementation.

design/one-pager-pluggable-secret-stores.md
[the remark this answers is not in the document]
turkenh: No strong reasons other than thinking we could achieve the same with 1 sidecar vs `n` number of sidecars. However, I would agree this could in practice end up having one monolithic sidecar binary which is not good. 
I will update accordingly.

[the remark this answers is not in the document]
turkenh: Done: https://github.com/crossplane/crossplane/pull/3029/commits/8120ba623f3902a42440be2ecb78757440475848

design/one-pager-pluggable-secret-stores.md
```diff
@@ -0,0 +1,267 @@
...
+    bool keep_empty_secret = 1;
+}
+```
+
+A _Secret Store Plugin_:
+
+- Will be a binary packaged as a container image.
+- Should run a gRPC server listening at Unix domain socket.
+- May support one or more Secret Stores.
+- Should deploy CRD(s) for supported store configurations at startup.
```
muvaf, line 159: Deploying own CRDs feels almost like a package 🤔

design/one-pager-pluggable-secret-stores.md
```diff
@@ -0,0 +1,267 @@
...
+be serving a well-defined API via gRPC. For the sake of security, plugins will
+be deployed as sidecars to Crossplane and Provider pods and communication will
+be over Unix domain sockets.
+
+![secret-store-plugin.png](images/secret-store-plugin.png)
+
+### API
+
+In addition to the existing Secret Store types, `Kubernetes` and `Vault`, we
+will define a new type named as `External`. We may consider deprecating type
```
muvaf, line 50: Should we make an explicit decision about whether we'll deprecate all built-in plugins except `Kubernetes`?

design/one-pager-pluggable-secret-stores.md
```diff
@@ -0,0 +1,267 @@
...
+- Will be a binary packaged as a container image.
+- Should run a gRPC server listening at Unix domain socket.
+- May support one or more Secret Stores.
+- Should deploy CRD(s) for supported store configurations at startup.
+- Will run as a sidecar next to Crossplane and Provider containers.
+
+#### Plugin Binary vs Secret Store Implementation
+
+All `rpc` methods have `ConfigReference` as their first parameter. So,
+one Secret Store Plugin may support more than one Secret Store implementation,
```
muvaf, line 165: I like that we're deferring the decision of how many places the secret content will end up in to the plugin. I assume their CRDs would likely allow toggling that configuration.

design/one-pager-pluggable-secret-stores.md
```diff
@@ -0,0 +1,269 @@
...
+spec:
+  type: External
+  defaultScope: crossplane-system
+  external:
+    configRef:
+      apiVersion: secrets.crossplane.io/v1alpha1
+      kind: VaultConfig
+      name: vault-internal
+---
+apiVersion: secrets.crossplane.io/v1alpha1
```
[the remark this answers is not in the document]
turkenh: Yes, intentional. 
It was meant to be installed by the plugin process itself during startup. However, considering RBAC, it might make more sense to install it during plugin installation via the package manager if we go that route.

design/one-pager-pluggable-secret-stores.md
```diff
@@ -0,0 +1,267 @@
...
+e.g. Vault, AWS, GCP, Azure ... all in a single Plugin binary, and related
+implementation could be called depending on a given config similar to how it is
+being done in [RuntimeStoreBuilder] today. Implementation of individual Secret
+Stores still need to satisfy the same [Store] interface today and could live in
+separate repositories. However, they would need to be packaged and served by a
+Secret Store Plugin binary to be used with Crossplane.
+
+### Deployment
+
+Crossplane helm chart will accept an optional container image for a secret
```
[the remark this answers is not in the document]
turkenh: yes

negz (approved):
This looks great, thanks @turkenh! I'd like @muvaf to sign off too since he's been reviewing.

design/one-pager-pluggable-secret-stores.md
muvaf: Their own CRD could have a config with multiple endpoint, i.e. the shape of `VaultConfig` is up to the plugin author. So, if there is a use case for multiple places, then the plugin could have that list in its own config CR instance.

PR #3031, opened 2022-04-06
Prevent flux from accidentally deleting serviceaccount when using ControllerConfig

### LINKED ISSUE #3030: Flux compatibility - service account annotations
### What happened?
We sometimes see bursts of Unauthorized for calls from crossplane providers to the kubernetes apiserver. After some debugging we've found that the UID in the service account does not match the token on the file system. We noticed that the service account has a "fluxcd.io" annotation, even though it is not managed by flux. This is because crossplane copies all annotations from the `ControllerConfig` to the `ServiceAccount` object. This is probably required for ArgoCD, and it is used for IRSA in AWS.

But for flux, this does not work. We believe flux will consider this as a situation where a former `Kustomization` managed this object, and delete it as it is no longer managed. crossplane will quickly recreate it, but the token mounted in the container will for some time have the old UID and the API server will reject the requests. I'm not sure if this is a race condition in the kubelet when a token is rotated while a pod is created, or if it is simply not a supported way to rotate the tokens by deleting the service account while it is in use, but Kubernetes does recover from this situation after a while.

The specific annotation is `"kustomize.toolkit.fluxcd.io/checksum": "3583e44b78adb421bd53fcb43edfcd5d39b287b8"`

### How can we reproduce it?
Flux-managed `ControllerConfig`, see above

### Suggested fix

I think we should exclude annotations with keys `fluxcd.io/` or prefix `kustomize.toolkit.fluxcd.io/` when creating the deployment and serviceaccount in

https://github.com/crossplane/crossplane/blob/1344a86018f03fcbba6b9365c63696dcd47ebcf6/internal/controller/pkg/revision/deployment.go#L174-L177

We might also want to skip key `kubectl.kubernetes.io/last-applied-configuration` or prefix `kubectl.kubernetes.io/` as it is fairly misleading to copy it.

Long term crossplane should probably implement "bring your own service account" instead (see issue https://github.com/crossplane/crossplane/issues/2295 etc)

Crossplane version: 1.6.3, EKS 1.21, flux 0.17

chlunde:
Flux will garbage collect the ServiceAccount if we copy the flux
annotations from the ControllerConfig, when the ControllerConfig is managed by flux and the ServiceAccount is not.

Exclude known bad and meaningless annotations when copying annotations.

Fixes #3030

I have:

Unit tests

### DIFF
--- a/internal/controller/pkg/revision/deployment.go
+++ b/internal/controller/pkg/revision/deployment.go
@@ -38,6 +38,11 @@ var (
 	allowPrivilegeEscalation = false
 	privileged               = false
 	runAsNonRoot             = true
+	doNotPropagate           = []string{
+		// Prevent flux from accidentally garbage collecting generated objects
+		"kustomize.toolkit.fluxcd.io/",
+		"kubectl.kubernetes.io/",
+	}
 )
 
 // Providers are expected to use port 8080 if they expose Prometheus metrics,
@@ -277,9 +282,13 @@ func buildProviderDeployment(provider *pkgmetav1.Provider, revision v1.PackageRe
 // filterAnnotations excludes annotations that are meaningless or unsafe to copy
 func filterAnnotations(in map[string]string) map[string]string {
 	ret := make(map[string]string)
+
+nextLabel:
 	for label, value := range in {
-		if strings.HasPrefix(label, "kustomize.toolkit.fluxcd.io/") || strings.HasPrefix(label, "kubectl.kubernetes.io/") {
-			continue
+		for _, labelPrefix := range doNotPropagate {
+			if strings.HasPrefix(label, labelPrefix) {
+				continue nextLabel
+			}
 		}
 		ret[label] = value
 	}
--- a/internal/controller/pkg/revision/deployment.go
+++ b/internal/controller/pkg/revision/deployment.go
@@ -17,6 +17,8 @@ limitations under the License.
 package revision
 
 import (
+	"strings"
+
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


jjcallis (approved):

haarchri (approved):
tested this in our flux environment - works as expected now - LGTM

internal/controller/pkg/revision/deployment.go
```diff
@@ -269,3 +273,15 @@ func buildProviderDeployment(provider *pkgmetav1.Provider, revision v1.PackageRe
 	}
 	return s, d, svc
 }
+
+// filterAnnotations excludes annotations that are meaningless or unsafe to copy
+func filterAnnotations(in map[string]string) map[string]string {
+	ret := make(map[string]string)
+	for label, value := range in {
+		if strings.HasPrefix(label, "kustomize.toolkit.fluxcd.io/") || strings.HasPrefix(label, "kubectl.kubernetes.io/") {
```
negz, line 281: Does it make sense to make these a "constant" (i.e. a `var doNotPropagate []string = ...`)?

negz: Following up here. I think this is mergeable if you have time to address this. :)

chlunde: @negz sorry, busy times! doNotPropagate at function level, or package?

negz: I was thinking package (sorry also for the huge delay in response times).

chlunde: @negz At least we have a stable cadence :)

wwentland:
It would be fantastic if this could be reviewed again! I believe that @negz suggestion has been addressed.

nullable-eth (approved):
Agree that @negz comment was addressed, no other objections.

chlunde:
@nullable-eth FYI I don't have a merge button, a maintainer must merge

nullable-eth:
> @nullable-eth FYI I don't have a merge button, a maintainer must merge

I think this still requires another code owner to review.

wwentland:
> > @nullable-eth FYI I don't have a merge button, a maintainer must merge
> 
> I think this still requires another code owner to review.

@nullable-eth Thank you! Who might be best placed to help us with that?

turkenh (approved):
Looking good to me, but needs a rebase before merge!

hasheddan (commented):
I am currently hesitant to continue to evolve this set of functionality (i.e. Provider deployment and configuration) given a desire to move towards customer provider runtimes as defined in https://github.com/crossplane/crossplane/issues/2671. In that model, our current functionality of creating a `Deployment` with certain attributes would be moved into a separate component, where folks could customize and or alter the behavior / level of customization as they see fit. @chlunde would you mind taking a peek at that proposal and weighing in if you have any thoughts?

negz:
Should we close this PR?

PR #3040, opened 2022-04-12
Add support for base64 encode/decode in Convert transform

### LINKED ISSUE #2920: Support base64 encode/decode within Composition patches using Transform Types

I want to be able to use [Transform Types](https://crossplane.io/docs/v1.6/reference/composition.html#transform-types) within the Composition to encode a string into base64/decode the base64 to a string.

I have a managed resource that exposes on Status its ID base64-encoded. Some resources receives this ID as an input (spec) as decoded base64, and other resources receive the value encoded. So I need to be able to properly patch this value as needed (both encoded or decoded)

Adding a new Transform Type that receives a string to be decoded from/encoded to base64

bobh66:

Added ToBase64 and FromBase64 Convert types to encode/decode strings.   There are a number of resources that require base64 encoded strings, so having the ability to encode the string as part of constructing the composition is critical to supporting these resources.

Fixes #2920 

I have:

Unit tests added and tested in a local kind cluster.


apis/apiextensions/v1/composition_transforms_test.go
```diff
@@ -248,6 +248,26 @@ func TestStringResolve(t *testing.T) {
...
+		"ConvertFromBase64": {
+			args: args{
+				stype:   StringTransformConvert,
+				convert: &frombase64,
+				i:       "Q3Jvc3NQbGFuZQ==",
+			},
+			want: want{
+				o: "CrossPlane",
+			},
+		},
```
negz, line 270: Could you add a test for the error case here (i.e. malformed base64 input) please.

apis/apiextensions/v1/composition_transforms.go
```diff
@@ -250,6 +253,11 @@ func stringConvertTransform(input interface{}, t *StringConversionType) (interfa
 		return strings.ToUpper(str), nil
 	case ConversionTypeToLower:
 		return strings.ToLower(str), nil
+	case ConversionTypeToBase64:
+		return base64.StdEncoding.EncodeToString([]byte(str)), nil
+	case ConversionTypeFromBase64:
+		s, err := base64.StdEncoding.DecodeString(str)
+		return string(s), err
```
negz, line 260: Can you `errors.Wrap` this error with some context (an error string constant) please - something like `errors.Wrap(err, errDecodeString)`.

apis/apiextensions/v1/composition_transforms.go
```diff
@@ -210,7 +213,7 @@ type StringTransform struct {
 
 	// Convert the type of conversion to Upper/Lower case.
 	// +optional
-	// +kubebuilder:validation:Enum=ToUpper;ToLower
+	// +kubebuilder:validation:Enum=ToUpper;ToLower;ToBase64;FromBase64
```
negz, line 216: These changes (just the type definitions, not the methods) need to be reflected in the v1alpha1 composition revision types as well. It actually looks like that file is already out-of-date (it doesn't seem to include ToUpper and ToLower for example). I think we'll need to bring it up-to-date in a separate commit so that we can back port that commit without back porting the rest of this feature.

negz (commented):
Thanks for working on this! A few comments but generally this looks good.

apis/apiextensions/v1/composition_transforms_test.go
bobh66: Done

apis/apiextensions/v1/composition_transforms.go
bobh66: Done

apis/apiextensions/v1/composition_transforms.go
bobh66: OK, so two commits - one to add the ToUpper/ToLower which can be backported, and a second to add the ToBase64/FromBase64 that will not be backported?

I'm not entirely sure what the scope of the changes should be, but this is what I came up with:

```
diff --git a/apis/apiextensions/v1alpha1/revision_types.go b/apis/apiextensions/v1alpha1/revision_types.go
index bbddaf5c..a7d3018f 100644
--- a/apis/apiextensions/v1alpha1/revision_types.go
+++ b/apis/apiextensions/v1alpha1/revision_types.go
@@ -408,12 +408,34 @@ func (m *MapTransform) Resolve(input interface{}) (interface{}, error) {
        }
 }
 
+// StringTransformType is type of the string transform function to be executed fmt/convert.
+type StringTransformType string
+
+// StringConversionType is the type of string conversion, ToUpper/ToLower/ToBase64/FromBase64
+type StringConversionType string
+
 // A StringTransform returns a string given the supplied input.
 type StringTransform struct {
+
+       // Type of the string transform to be run.
+       // +optional
+       // +kubebuilder:validation:Enum=Format;Convert;TrimPrefix;TrimSuffix
+       // +kubebuilder:default=Format
+       Type StringTransformType `json:"type,omitempty"`
+
        // Format the input using a Go format string. See
        // https://golang.org/pkg/fmt/ for details.
        // +immutable
-       Format string `json:"fmt"`
+       Format string `json:"fmt,omitempty"`
+
+       // Convert the type of conversion to Upper/Lower case.
+       // +optional
+       // +kubebuilder:validation:Enum=ToUpper;ToLower
+       Convert *StringConversionType `json:"convert,omitempty"`
+
+       // Trim the prefix or suffix from the input
+       // +optional
+       Trim *string `json:"trim,omitempty"`
 }
 
 // A ConvertTransform converts the input into a new object whose type is supplied.
```

Is that right?   I wasn't sure if Format should still be "immutable", but making it "optional" changed the type to *string.

Thanks!

negz: > I assume it needs two commits - one to add the ToUpper/ToLower which can be backported, and a second to add the ToBase64/FromBase64 that will not be backported?

Yes exactly, and I appreciate you tackling that! The diff you posted looks good to me.

bobh66: Thanks!  These changes are pushed too.

muvaf (approved):
LGTM! Thanks @bobh66 ! Great first contribution!

PR #3045, opened 2022-04-15
Add a project charter

negz:

The Crossplane steering committee hopes that establishing a charter will help our community better understand our intentions for the project - in particular what is and isn’t in scope. As a byproduct of this, we hope it will help vendors better understand where Crossplane ends and they start.

The document is informed by the CNCF’s [guidance] around project charters, which they recommend consist of a brief mission statement, supplemented by scope and principles. It’s also heavily inspired by the “What is Kubernetes?” [page] (which appears to be the closest thing the Kubernetes project as a whole has to a charter). Like said document this one attempts to describe what Crossplane is for and why it exists without using or introducing specific Crossplane terminology (e.g. “provider”, XR, etc).

[guidance]: https://contribute.cncf.io/maintainers/governance/charter/
[page]: https://kubernetes.io/docs/concepts/overview/what-is-kubernetes/

I have:

The steering committee did an internal review before opening this PR.

### DIFF
--- a/CHARTER.md
+++ b/CHARTER.md
@@ -51,9 +51,9 @@ to control anything - not just containers. The Crossplane project consists of:
 * The Crossplane Resource Model (XRM). An API model for Crossplane extensions.
 * The ability to compose bespoke control plane APIs for any compute plane.
 * A package format and manager to extend your control plane with:
-* The ability to orchestrate any kind of compute plane provider.
-* New ways to compose control plane APIs.
-* New integration with supporting services such as secret stores.
+  * The ability to orchestrate any kind of compute plane provider.
+  * New ways to compose control plane APIs.
+  * New integration with supporting services such as secret stores.
 * Tooling to create and test new Crossplane extensions.
 * A conformance program for Crossplane extensions.
 
--- a/CHARTER.md
+++ b/CHARTER.md
@@ -1,0 +1,117 @@
+# Project Charter
+
+## Overview
+
+Crossplane is a framework for building cloud native control planes without
+needing to write code. It has a highly extensible backend that enables you to
+build a control plane that can orchestrate applications and infrastructure no
+matter where they run, and a highly configurable frontend that puts you in
+control of the schema of the declarative API it offers.
+
+## Control Planes
+
+Control planes first emerged in network packet switching, where devices must
+forward packets as fast as possible, but also be highly configurable. Breaking
+out the control plane from the data plane allows the former to be optimized for
+flexibility while the latter is optimized for speed. It also decouples the
+failure domains - a control plane failure does not prevent the data plane from
+functioning. Control planes are rooted in control theory, in particular closed
+loop control where the actual state of a system is observed in order to
+determine how best to drive it toward a desired state.
+
+Control planes power cloud computing. Cloud providers like AWS are built with
+control planes, as are projects like Kubernetes. In cloud computing the data
+plane is instead a “compute plane” consisting of applications and the
+infrastructure they run on - VMs, containers, databases, caches, queues, etc.
+
+Control planes are popular in cloud computing because they:
+
+* Are “always-on”, constantly correcting drift by driving observed state toward
+  desired state.
+* Are API-driven, which enables discovery, collaboration, and integration with
+  other systems.
+* Don’t need to be installed on every laptop and CI/CD system, making them
+  easier to operate.
+* Keep control in a separate failure domain from the plane under control (e.g.
+  compute).
+
+Cloud native organizations increasingly prefer control planes to approaches such
+as imperative scripts or Infrastructure-as-Code (IaC), but historically building
+a control plane has required a significant investment to design and code it from
+scratch. This is where Crossplane comes in.
+
+## What Crossplane Is
+
+Crossplane is a neutral place for vendors and individuals to come together in
+enabling control planes. It offers a framework to build control planes for cloud
+native computing without needing to write code (unless you want to). It shares a
+foundation with Kubernetes, but supports many powerful extensions that enable it
+to control anything - not just containers. The Crossplane project consists of:
+
+* The Crossplane Resource Model (XRM). An API model for Crossplane extensions.
+* The ability to compose bespoke control plane APIs for any compute plane.
+* A package format and manager to extend your control plane with:
+* The ability to orchestrate any kind of compute plane provider.
+* New ways to compose control plane APIs.
+* New integration with supporting services such as secret stores.
+* Tooling to create and test new Crossplane extensions.
+* A conformance program for Crossplane extensions.
+
+As a framework, Crossplane:
+
+* Avoids building features specific to any particular compute plane in its core.
+  Instead features pertaining to any particular cloud provider, project, or tool
+  are enabled through extension points.
+* Focuses on APIs. Crossplane provides and recommends “golden path” tools to
+  build conformant extensions but does not require them. Anything with
+  conformant API schema and behavior may be considered a Crossplane extension -
+  regardless of how it is built. 
+* Optimizes for “vertical integration”. That is, Crossplane extensions are
+  intended and designed to be used in conjunction with Crossplane and each
+  other.
+* Prioritizes enabling a “separation of concerns” between control plane curators
+  and consumers. Curators deploy, extend, and configure control planes, while
+  consumers use them to orchestrate compute planes.
+
+## What Crossplane Is Not
+
+Crossplane is not intended to be a “batteries included” control plane for any
+particular use case. Instead it is a framework that you can deploy, configure,
+and extend to power control planes for your own highly bespoke use cases, often
+without writing any code.
+
+The project is a neutral place for vendors and individuals to come together in
+enabling control planes. To that extent, it does not police Crossplane
+extensions. Any extension that passes the appropriate Crossplane conformance
+suite may consider itself to be “a Crossplane extension” - regardless of how it
+is built, how it is licensed, who maintains it, how it is governed, or where it
+is hosted. Conformance suites are typically concerned with behavior and (where
+applicable) API shape, not internal implementation details.
+
+Concrete examples of things that are not in scope for the project include (but
+are not limited to):
+
+* Providing or mandating abstractions around specific use cases such as
+  provisioning applications (e.g. a PaaS API) or provisioning Kubernetes
+  clusters. 
+* Providing or mandating a configuration language (for example, CUE). Crossplane
+  provides a declarative API that may be targeted by arbitrary configuration
+  languages, and extension points that may be powered by arbitrary languages.
+* Enabling the use of Crossplane extensions without using Crossplane and vice
+  versa; this is not supported and may result in undefined behavior.
+* Providing or mandating opinions around control plane hosting and tenancy
+  models.
+* Offering ‘in-tree’ support for particular compute planes - e.g. AWS. All
+  features specific to any specific compute plane provider are built as
+  extensions.
+* Offering a package specification or manager for anything other than Crossplane
+  extensions.
+* Offering API clients such as command-line interfaces (CLIs) or web consoles.
+
+## Updating this Document
+
+This is a living document. Changes to the scope, principles, or mission
+statement of the Crossplane project require a [majority vote][sc-voting] of the
+steering committee.
+
+[sc-voting]: https://github.com/crossplane/crossplane/blob/master/GOVERNANCE.md#updating-the-governance
--- a/CHARTER.md
+++ b/CHARTER.md
@@ -51,7 +51,7 @@ to control anything - not just containers. The Crossplane project consists of:
 * The Crossplane Resource Model (XRM). An API model for Crossplane extensions.
 * The ability to compose bespoke control plane APIs for any compute plane.
 * A package format and manager to extend your control plane with:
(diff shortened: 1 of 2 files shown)


CHARTER.md
```diff
@@ -0,0 +1,117 @@
...
+out the control plane from the data plane allows the former to be optimized for
+flexibility while the latter is optimized for speed. It also decouples the
+failure domains - a control plane failure does not prevent the data plane from
+functioning. Control planes are rooted in control theory, in particular closed
+loop control where the actual state of a system is observed in order to
+determine how best to drive it toward a desired state.
+
+Control planes power cloud computing. Cloud providers like AWS are built with
+control planes, as are projects like Kubernetes. In cloud computing the data
+plane is instead a “compute plane” consisting of applications and the
```
negz, line 24: The term "compute plane" has been somewhat divisive when I've tested it with folks so far. I'm not married to it, but I would love a succinct way to address the "other plane" - i.e. what Crossplane is a control plane _for_.

Two alternatives that I don't love are:

* Call it a "data plane". This is the typical term but seems odd outside of a network context.
* Use "infrastructure plane" or "application plane". I would prefer a term inclusive of both.

I'm open to suggestions here.

Piotr1215: How about "runtime plane"? It sounds a bit more application-focused, but it can also represent infrastructure or data. It also emphasises the fact that the control process and environment itself are dynamic.

CHARTER.md
```diff
@@ -0,0 +1,117 @@
...
+to control anything - not just containers. The Crossplane project consists of:
+
+* The Crossplane Resource Model (XRM). An API model for Crossplane extensions.
+* The ability to compose bespoke control plane APIs for any compute plane.
+* A package format and manager to extend your control plane with:
+* The ability to orchestrate any kind of compute plane provider.
+* New ways to compose control plane APIs.
+* New integration with supporting services such as secret stores.
+* Tooling to create and test new Crossplane extensions.
+* A conformance program for Crossplane extensions.
```
muvaf, line 58:
```suggestion
  * The ability to orchestrate any kind of compute plane provider.
  * New ways to compose control plane APIs.
  * New integration with supporting services such as secret stores.
  * Tooling to create and test new Crossplane extensions.
  * A conformance program for Crossplane extensions.
```

CHARTER.md
```diff
@@ -0,0 +1,117 @@
...
+* Providing or mandating a configuration language (for example, CUE). Crossplane
+  provides a declarative API that may be targeted by arbitrary configuration
+  languages, and extension points that may be powered by arbitrary languages.
+* Enabling the use of Crossplane extensions without using Crossplane and vice
+  versa; this is not supported and may result in undefined behavior.
+* Providing or mandating opinions around control plane hosting and tenancy
+  models.
+* Offering ‘in-tree’ support for particular compute planes - e.g. AWS. All
+  features specific to any specific compute plane provider are built as
+  extensions.
```
muvaf, line 106:
```suggestion
  extensions such as providers.
```

nit to make that more clear.

CHARTER.md
```diff
@@ -0,0 +1,117 @@
...
+* Offering a package specification or manager for anything other than Crossplane
+  extensions.
+* Offering API clients such as command-line interfaces (CLIs) or web consoles.
```
muvaf, line 109: Does that mean Crossplane CLI (and [crank](https://github.com/crossplane/crossplane/tree/master/cmd/crank)) will be deprecated?

CHARTER.md
```diff
@@ -0,0 +1,117 @@
+# Project Charter
```
muvaf, line 1: One thing I'm struggling is whether _Crossplane_ mentioned here entails everything under `crossplane` org or `crossplane/crossplane` repository. For example, I know that this document means providers hosted under `crossplane` may no longer be but it's hard to get that this is changing. If that shouldn't be part of the charter, maybe we can include changes to `GOVERNANCE.md` resulting from this change so that it's more clear to folks what's changing in practice.

muvaf (commented):
Love the new definition!

CHARTER.md
negz: I'm kind of torn on this one - I was trying to avoid using terminology specific to the project.

CHARTER.md
negz: Personally I lean toward grandfathering them in as a bit of an exception. I don't think we'd add any more functionality to them though.

CHARTER.md
negz: Yeah I considered capturing that here too, but as you mentioned I think GOVERNANCE.md is the better place. I believe the current thinking is:

* Core, runtime, tools (incl Terrajet) all live under `github.com/crossplane`. Basically things that aren't extensions.
* Extensions _can_ live under `github.com/crossplane-contrib` if they want to be under project governance.

The immediate practical implication being that we'd move all providers to contrib.

CHARTER.md
negz: Ah good catch, but I think the last two bullet points should not be indented.

CHARTER.md
bobh66: Maybe "resource plane"?  Since crossplane "controls" (or "manages") resources, both composite and "managed" or native?

CHARTER.md
```diff
@@ -0,0 +1,117 @@
...
+Crossplane is a neutral place for vendors and individuals to come together in
+enabling control planes. It offers a framework to build control planes for cloud
+native computing without needing to write code (unless you want to). It shares a
+foundation with Kubernetes, but supports many powerful extensions that enable it
+to control anything - not just containers. The Crossplane project consists of:
+
+* The Crossplane Resource Model (XRM). An API model for Crossplane extensions.
+* The ability to compose bespoke control plane APIs for any compute plane.
+* A package format and manager to extend your control plane with:
+  * The ability to orchestrate any kind of compute plane provider.
```
hasheddan, line 54: This is really the only place where "compute plane" felt a little strange, or perhaps restrictive, to me -- do we feel that all providers fall into this category?

CHARTER.md
```diff
@@ -0,0 +1,117 @@
...
+* Providing or mandating abstractions around specific use cases such as
+  provisioning applications (e.g. a PaaS API) or provisioning Kubernetes
+  clusters. 
+* Providing or mandating a configuration language (for example, CUE). Crossplane
+  provides a declarative API that may be targeted by arbitrary configuration
+  languages, and extension points that may be powered by arbitrary languages.
+* Enabling the use of Crossplane extensions without using Crossplane and vice
+  versa; this is not supported and may result in undefined behavior.
+* Providing or mandating opinions around control plane hosting and tenancy
+  models.
```
hasheddan, line 103: Thinking about this in the context of our current "multi-tenancy" features, do this mean something like provider partitioning in core Crossplane would be out of scope? It is related to multi-tenancy, and saying multi-tenancy features will not be "provided" feels like a pretty strong statement around what we won't do. Perhaps what I'm looking for is whether "interfaces the enable multi-tenancy" may be in scope, rather than concrete implementations. An interface could be considered an "opinion" about multi-tenancy.

CHARTER.md
hasheddan: I started out wanting a more concrete definition for "extension", but found myself pleased with the generality by the time I got to this point :)

CHARTER.md
hasheddan: FWIW I would like to drop Crossplane CLI / crank. The most commonly used functionality is installing a package, and one could argue that it is an anti-pattern to use a CLI to imperatively carry out that operation (i.e. rather than applying a manifest).

hasheddan (commented):
Thank you @negz and steering committee members for working to define scope and mission here -- overall very pleased with the direction :)

CHARTER.md
negz: Thanks folks! I personally like "runtime plane" and "resource plane" about equal to "compute plane". I'll keep it open for suggestions and perhaps the other steering committee members can chime in.

CHARTER.md
negz: Right. I'd love for this document to be specific enough to create some real bounds on project scope, but at the same time open ended enough that we don't need to update it very often as we tweak the features and architecture of the project.

CHARTER.md
negz: Sounds good. Maybe we can deprecate it around the time that we ratify this PR.

That said, do we have a good alternative for the creation of packages? That seems like something that should live in the Crossplane realm (it would be odd to be able to install them but not create them).

CHARTER.md
negz: FWIW I meant provider in the general sense, not a Crossplane provider. So AWS for example is a "compute provider". Whether all provider (SQL, Terraform, Slack, GitHub, etc) fall into that category depends how far we want to stretch the definition of "compute".
 
```suggestion
  * The ability to orchestrate any kind of compute resource.
```

Perhaps the above would make more sense - which makes me lean a little more toward @bobh66's suggestion of "resource plane".

CHARTER.md
negz: I don't love this bullet point as it stands - I've struggled to succinctly capture our thinking around tenancy. This is complicated because "tenants" can mean different things to different people. Different tenants could be different apps, different teams, or different paying customers.

Today I believe Crossplane (like Kubernetes) offers a kind of "soft" or "leaky" multi-tenancy. Claims are namespaced, but anyone who wants to discover available Compositions or "see how the sausage is made" (i.e. how their claim is composed) needs at least read-only RBAC access to Compositions and/or some kinds of MR at the cluster scope. I'd judge this to be fairly tolerable for _team based_ tenancy - it's probably not so bad that you can see what infrastructure another team has - but not acceptable for paying customers. I'd say Kubernetes is pretty similar - you'd probably let many teams have `kubectl` access to a single cluster but not multiple paying customers.

As you allude to, Crossplane has further tenancy constraints in the sense that some orgs want to use pod based identity (e.g. AWS IRSA) _and_ multiple AWS accounts for multiple tenants.

With all that context in place, the spirit of what I'm trying to say here is that enabling a stronger tenancy model is out of scope - I think that would include adding interfaces to enable that model. We feel that (again similar to Kubernetes) the Crossplane tenancy model is good enough for many use cases, and that where it isn't good enough the best approach is to deploy multiple Crossplanes.

thejosephstevens:
This makes a lot of sense! I'm fairly new to the community, but for my ten cents: the explicit distinction between IaC versus Control Plane models gives a really clear language for how Crossplane is different than Terraform and Pulumi with clear alignment with other modern technologies and practices. I think this makes it easier to align people on how to talk about Crossplane as a tool, and improve adoption by making the value proposition more clear.

In particular I think this is a very simple way of getting this point across
>Cloud native organizations increasingly prefer control planes to approaches such
as imperative scripts or Infrastructure-as-Code (IaC), but historically building
a control plane has required a significant investment to design and code it from
scratch. This is where Crossplane comes in.

CHARTER.md
muvaf: nitpick: I'd suggest to make that change to `GOVERNANCE.md` in this PR to make this clear to folks.

CHARTER.md
```diff
@@ -0,0 +1,117 @@
...
+native computing without needing to write code (unless you want to). It shares a
+foundation with Kubernetes, but supports many powerful extensions that enable it
+to control anything - not just containers. The Crossplane project consists of:
+
+* The Crossplane Resource Model (XRM). An API model for Crossplane extensions.
+* The ability to compose bespoke control plane APIs for any compute plane.
+* A package format and manager to extend your control plane with:
+  * The ability to orchestrate any kind of compute plane provider.
+  * New ways to compose control plane APIs.
+  * New integration with supporting services such as secret stores.
```
turkenh, line 56: FWIW, here we are assuming secret stores plugins to be installed via package manager which is at least not the case with the [initial design](https://github.com/crossplane/crossplane/pull/3029).

negz: The idea is to give examples of non-provider extensions the package manager might include in future and Secret Stores was the "most real" example (webhooks and functions are others that aren't as far along). I'm not meaning to imply that this works today, but rather that doing something like this would be within the project's charter.

CHARTER.md
muvaf: I think the usual `docker build` or any other OCI image builder does the job, right? As long as the files exist in the right places.

jbw976 (approved):
Thanks for driving this @negz and helping bring some clarity to the scope and purpose of the Crossplane project! 💪

CHARTER.md
negz: After discussion with the steering committee we're going to just stick with "compute plane".

bassam (approved):
Looks great

CHARTER.md
negz: I'm going to leave 'providers' out here. I'm hoping that over time 'extensions' will be as meaningful as 'providers' (i.e. once it's more common to have extensions that are not providers).

CHARTER.md
negz: I'm going to resolve this conversation and leave it open ended for now. :) Personally I could go either way WRT grandfathering `crank` in or removing it - perhaps we can raise an issue to follow up on that once the charter is merged.

CHARTER.md
negz: I kind of agree, but since this has been approved I'm going to fast follow instead. :)

PR #3082, opened 2022-05-06
Update latest 1.6.x and 1.7.x releases

negz:

Fixes https://github.com/crossplane/crossplane/issues/3075
Fixes https://github.com/crossplane/crossplane/issues/3076

I have:

N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,8 +27,8 @@ documentation].
 You can subscribe to the [community calendar] to track all release dates.
 
 [v1.5.3]: https://github.com/crossplane/crossplane/releases/tag/v1.5.3
-[v1.6.5]: https://github.com/crossplane/crossplane/releases/tag/v1.6.5
-[v1.7.0]: https://github.com/crossplane/crossplane/releases/tag/v1.7.0
+[v1.6.6]: https://github.com/crossplane/crossplane/releases/tag/v1.6.6
+[v1.7.1]: https://github.com/crossplane/crossplane/releases/tag/v1.7.1
 
 ## Get Involved
 


README.md
```diff
@@ -27,8 +27,8 @@ documentation].
 You can subscribe to the [community calendar] to track all release dates.
 
 [v1.5.3]: https://github.com/crossplane/crossplane/releases/tag/v1.5.3
-[v1.6.5]: https://github.com/crossplane/crossplane/releases/tag/v1.6.5
-[v1.7.0]: https://github.com/crossplane/crossplane/releases/tag/v1.7.0
+[v1.6.6]: https://github.com/crossplane/crossplane/releases/tag/v1.6.6
+[v1.7.1]: https://github.com/crossplane/crossplane/releases/tag/v1.7.1
```
negz, line 31: Does maintaining (and having to update this list) really help anyone? Can't they figure out the latest releases pretty easily by browsing to the 'Releases' page on GitHub?

hasheddan: @negz I think it is useful given that if we have multiple patches on a branch in a row it can bury the latest patches for other minor versions. We'll also need to update the `Current Patch` column above to match these new links 👍🏻

negz: I'm sure it's of non-zero use, but I'm (very) skeptical that it's worth the extra step in the release process. GitHub releases may not have the ideal UX for a project that releases patches on multiple release branches but it's quite a common pattern nonetheless and I don't see too many other projects going out of their way to clarify which one is latest. Speaking personally (as a user of other projects) I'd find this to be a minor convenience at best.

PR #3083, opened 2022-05-06
Don't link to latest patch releases in README.md

negz:

This PR removes the releases table's links to patch releases.

I'm sure it's of non-zero use, but I'm (very) skeptical that it's worth the extra step in the release process.

GitHub releases may not have the ideal UX for a project that releases patches on multiple release branches but it's quite a common pattern nonetheless and I don't see too many other projects going out of their way to clarify which one is latest. Speaking personally (as a user of other projects) I'd find this to be a minor convenience at best.

I have:

### DIFF
--- a/.github/ISSUE_TEMPLATE/release.md
+++ b/.github/ISSUE_TEMPLATE/release.md
@@ -21,22 +21,12 @@ Please ensure all artifacts (PRs, workflow runs, Tweets, etc) are linked from
 this issue for posterity. Refer to this [prior release issue][release-1.7] for
 examples of each step.
 
+<!-- Uncomment the following block only if cutting a minor release. -->
 <!--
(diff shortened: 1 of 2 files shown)


muvaf (approved):
LGTM!

hasheddan (commented):
I personally find this table fairly useful, especially when we have a series of patches on one branch that is not present on the others, but I also acknowledge that:
1. This is extra work per release.
2. This is not authoritative / automated, so it has potential to actually introduce confusion rather than eliminate it.

Perhaps we can add an automated way to display latest on each branch in the future.

negz:
> Perhaps we can add an automated way to display latest on each branch in the future.

That sounds good to me if possible. My main motivation here is removing a PR from the release flow.

PR #3085, opened 2022-05-11
Add lifecycle hooks to ControllerConfig

### LINKED ISSUE #3084: Add lifecycle configuration to ControllerConfig
provider-terraform needs to be able to gracefully shutdown any running terraform commands when before the pod is terminated.  This is needed to avoid locks being left behind on Lease objects in the kubernetes back end.

One way to do this would be to use the pod lifecycle configruation to define a preStop hook that sends a HUP signal to all running terraform processes so they can shut down cleanly before the pod exits.

```
  lifecycle:
    preStop:
      exec:
        command: ["/bin/sh","-c","while killall -HUP terraform; do sleep 1; done"]
```

Add support for the lifecycle configuration to the ControllerConfig definition so those attributes can be copied into the pod definition at runtime.

See https://github.com/crossplane-contrib/provider-terraform/issues/46

bobh66:

Add support for lifecycle hooks to the ControllerConfig resource

Fixes #3084 

I have:

Tested in a local crossplane deployment and verified that the provider-terraform pod is now deployed with a preStop lifecycle hook


chlunde:
FWIW I added a kind of review/some alternative ideas at the original issue: https://github.com/crossplane-contrib/provider-terraform/issues/46#issuecomment-1125855555

negz:
@bobh66 Do you want to move forward with this PR given @chlunde's proposed alternatives? I tend to agree with @chlunde  that it would be nicer to fix the specific issue you're facing "in code" rather than requiring folks to use a ControllerConfig.

bobh66:
@negz I'm fine with abandoning this change and moving forward with a solution in the provider.  I'd like to get some direction on how to implement a signal handler within the provider-terraform code to catch the signal and send SIGHUP to the running terraform processes.   I know it's not guaranteed to work 100% but it can't hurt.

bobh66:
@negz @chlunde I believe I have a way to catch the context.Cancelled and context.DeadlineExceeded events in the provider-terraform code, and send a SIGTERM to the running terraform processes  (https://github.com/crossplane-contrib/provider-terraform/issues/46#issuecomment-1180484879),   That eliminates the need for this change, however I can envision a scenario where we would like to be able to set terminationGracePeriodSeconds to a value greater than 30 in the ControllerConfig.

I'd like to close this PR and open a new one that adds support for terminationGracePeriodSeconds  in ControllerConfig if that is acceptable?

Thanks

PR #3089, opened 2022-05-17
Enchance multiple of transform math type to support float64

### LINKED ISSUE #3059: Enhance `multiply` of transform `math` type with support of `float64`

If we try the constructs like

```
      transforms:
      - type: convert
        convert:
          toType: float64
      - type: math
        math:
          multiply: 90
```

```
```

edith007:

Fixes #3059 

I have:


ytsarev (changes requested):
@edith007 i am not sure if it's a draft or you are waiting for review? In any case please follow DCO by adding Sign-off: statement to the commit message

PR #3099, opened 2022-05-24
Add regexp support for string transform

### LINKED ISSUE #3028: Regular expression group capture string transformation

Crossplane currently supports a limited number of string transformations:
* fmt
* trim
* convert

However, I feel these are quite limiting. Crossplane could additionally support a regular expression syntax along with group capture, which would be much more flexible. It would even obsolete the need for TrimPrefix/TrimSuffix.

Imagine we had something like this:

```yaml
        - type: ToCompositeFieldPath
          toFieldPath: status.aws.accountNumber
          transforms:
            - type: string
              string:
                regexp: "arn:aws:iam::(\d+):.*"
                group: 1
```

So if the the ARN was: `arn:aws:iam::12345678910:role/role-name` it would extract `12345678910` from it. The account number could then be fed into other resources like `assumeRolePolicyDocument` of a Role.

kiote:

Adds a basic possibility of regexp usage in string transformations for patches.

Usage example:

```yaml
        - type: ToCompositeFieldPath
          toFieldPath: status.aws.accountNumber
          transforms:
            - type: string
              string:
                regexp: "arn:aws:iam::(\d+):.*"
```

Fixes #3028

I have:

I've added unit tests.


docs/reference/composition.md
```diff
@@ -555,6 +556,13 @@ Currently only `multiply` is supported.
   string:
      type: TrimSuffix
      trim: '-test'
+
+# If the value of the 'from' field is my-string-1-test, the value of the 'to' field will
+# be set to my-string
```
negz, line 561: Is this true?

apis/apiextensions/v1/composition_transforms.go
```diff
@@ -220,6 +225,11 @@ type StringTransform struct {
 	// Trim the prefix or suffix from the input
 	// +optional
 	Trim *string `json:"trim,omitempty"`
+
+	// Extract first match from the input using a Go regex string. See
+	// https://pkg.go.dev/regexp/ for details.
+	// +optional
+	Regexp *string `json:"regexp,omitempty"`
```
negz, line 232: You'll need to mirror these changes (just the types, no logic) in `apis/apiextensions/v1alpha1/revision_types.go`. I suspect some changes will also be needed to replicate the values in `internal/controller/apiextensions/composition/revision.go`.

negz (commented):
Thank you! This API looks good to me.

docs/reference/composition.md
kiote: fixed

apis/apiextensions/v1/composition_transforms.go
negz: Honestly I think we forgot to update it in revision.

kiote: haha, I'll dig a bit more then

negz: That appears to be only copying the `Format` field of the `StringTransform` struct - it's missing the existing `Type` and `Trim` fields, and the new `Regexp` field.

negz: Given you're now cleaning up stuff we missed earlier I'd be happy to merge this PR as is and follow up to add this all myself. You will need to figure out the `check-diff` check failing though - typically that means that some change created by running `make reviewable` should be part of this PR but isn't.

kiote: Sounds like a good option for me, if not too much to ask, can you please somehow show me that follow-up after it's done? would be nice to understand it after all.

I did `make reviewable` magic again, which is indeed has changed some files. So I think it should be good now?

apis/apiextensions/v1/composition_transforms.go
```diff
@@ -275,6 +287,23 @@ func stringTrimTransform(input interface{}, t StringTransformType, trim string)
...
+	}
+
+	re, err := regexp.Compile(r)
+
+	if err != nil {
+		return nil, errors.Errorf(errStringTransformTypeRegexFailed, err.Error())
+	}
+
+	str := fmt.Sprintf("%v", input)
+	result := re.Find([]byte(str))
```
negz, line 302: I was just doing a final re-read before approving and it occurred to me that `re.Find` will either return `[]byte` or `nil`. I'm not sure how these values will be handled by the rest of the transform/patch code - in particular I suspect `[]byte` will be rendered as base64 encoded arbitrary bytes, not the expected value (presumably a string).

My suggestion here is:

* The transform should fail (i.e. return an error) if there is no match (i.e. `re.Find` returns `nil`).
* The transform should always type assert the returned `[]byte` into a ` string` before returning it.

negz: Given that I took so long to spot this I'm inclined to merge this and fix it in a follow-up PR. I'll do that once https://github.com/crossplane/crossplane/pull/3158 is merged. :)

PR #3103, opened 2022-05-25
Fix logger panic in dependency resolution and guard against future logger panic bugs

### LINKED ISSUE #3102: Invalid Package version causes XP to panic
### What happened?

Crossplane panics when trying to log about a lock issue:
```
1.6534635870410209e+09	DPANIC	crossplane	odd number of arguments passed as key-value pairs for logging	{"controller": "packages/lock.pkg.crossplane.io", "request": "/lock", "uid": "268a27ef-0551-4689-a769-4784e731e74b", "version": "126095196", "name": "lock", "ignored key": "dependency (xpkg.upbound.io/crossplane/provider-aws) does not have version in constraints (>=v0.27.0-0)"}
```

Crossplane had recently been upgraded from 1.6.x to 1.7.0. 
There was a single lock on the system that existed before XP was upgraded. 

```
apiVersion: v1
items:
- apiVersion: pkg.crossplane.io/v1beta1
  kind: Lock
  metadata:
    creationTimestamp: "2022-01-26T19:49:35Z"
    generation: 9873
    name: lock
    resourceVersion: "130572258"
    uid: 268a27ef-0551-4689-a769-4784e731e74b
kind: List
metadata:
  resourceVersion: ""

```

### How can we reproduce it?

Possibly upgrade Crossplane with a package lock that does not include a value version.

Crossplane version: 1.7.0-up1.
K8s: EKS, v1.21.5-eks-9017834

hasheddan:

Fixes a panic in the logger when no version is supplied in the lock.

Updates all reconciler tests to use a test logger that mirrors
Crossplane's logger in debug mode. All logs are discarded, but passing
the logger will reveal any panics that result from any keys that are
passed without values.

> Note: passing the logger in the reconciler tests should guard against these kind of bugs showing up in the future. We have had multiple instances where a log panic in debug mode has resulted in us needing to run a patch release, which is an unfortunate outcome that this will help avoid.

Fixes #3102 

I have:

Verified that all tests are passing successfully and that without the fix to the resolver reconciler test was failing with log panic.

### DIFF
--- a/internal/controller/apiextensions/composite/reconciler_test.go
+++ b/internal/controller/apiextensions/composite/reconciler_test.go
@@ -18,6 +18,7 @@ package composite
 
 import (
 	"context"
+	"io"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
@@ -27,10 +28,12 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/utils/pointer"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	"github.com/crossplane/crossplane-runtime/pkg/errors"
+	"github.com/crossplane/crossplane-runtime/pkg/logging"
 	"github.com/crossplane/crossplane-runtime/pkg/reconciler/managed"
 	"github.com/crossplane/crossplane-runtime/pkg/resource"
 	"github.com/crossplane/crossplane-runtime/pkg/resource/fake"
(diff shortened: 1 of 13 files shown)


muvaf (approved):
LGTM!

internal/controller/apiextensions/composite/reconciler_test.go
```diff
@@ -42,6 +45,7 @@ import (
 
 func TestReconcile(t *testing.T) {
 	errBoom := errors.New("boom")
+	testLog := logging.NewLogrLogger(zap.New(zap.UseDevMode(true), zap.WriteTo(io.Discard)).WithName("testlog"))
```
negz, line 48: Could be handy to define this in the logging package in future.

hasheddan: @negz agree 👍🏻 was planning on opening an issue on `crossplane-runtime`

negz: I forgot about this PR and was wondering why this pattern was everywhere. Definitely still feel like this would be good to have in crossplane-runtime. In fact we could probably make the NopLogger that everything uses by default use this pattern so that we'd detect panics without having to explicitly plumb it in everywhere.

PR #3129, opened 2022-06-08
Adds JSON string encoding for `transforms` API

### LINKED ISSUE #2943: Support a 'json' transformation type to render connection secrets in proper json
I am working with a terrajet provider based on [hashicorp/vault](https://registry.terraform.io/providers/hashicorp/vault/latest/docs). The `vault_generic_secret` resource requires that secret data be submitted as a json-formatted string.

Currently, emitting a json-formatted string requires manually constructing proper json using the string "fmt" transform, which could quickly become unreadable and unworkable.

Implement a 'json' transform which could be used to wrap any string in proper json formatting.

For example:
provider-gcp currently offers a connection secret for a GKE cluster which contains `clusterCA`, `endpoint`, and `kubeconfig.`

It is easy to render `endpoint` in json with:
```
transforms:
- type: string
  string:
    type: Format
    fmt: '{"clusterCA":"%s"}'
```
 
However, both `clsuterCA` and `kubeconfig` are multi-line strings and would not survive this simple form of json encoding.

Another example is a serviceaccount from the same provider:

The privateKey for a ServiceAccount is already a json. When passed to a resource expecting json input, the fields are interpreted as individual fields instead of treating the full json as a string value.

Crossplane could provider a format which would accept any of these fields and allow them to be encoded into safe json that would look like the following:

```yaml
transforms:
- type: string
   string:
    type: Encode
    encode: JSON
  ```

### PR DESCRIPTION
**Pull request by jpmcb**

👋🏼  - first time Crossplane contributor here! Feel free to let me know if this is missing anything!

- Accepts a transform with the encoding specified as "JSON" in the format:

```yaml
transforms:
- type: string
   string:
    type: Encode
    encode: JSON
```

- Takes in the raw string, marshals it to ensure it's valid JSON
  and than returns the valid formatted JSON string

Fixes #2943

I have:

Added two unit tests to cover this. Still spinning up an environment to give this a full e2e spin.


negz:
Thanks @jpmcb! This looks good, but similar to the thread at https://github.com/crossplane/crossplane/pull/3099#discussion_r881080373 you'll need to reflect these changes in a few other places to account for our v1alpha1 CompositionRevision feature.

jpmcb:
Done! Also updated the docs in regards to this new transform as well!

negz:
I was just about to approve this, but it occurred to me that we already support `ToBase64` in the "convert" type of string transform. I think this might be a better fit there for consistency rather than adding another new type of string transform (even though I think "encode" is more accurate for both this case and `ToBase64`).

So this would instead make the API:


```yaml
transforms:
- type: string
  string:
    type: Convert
    convert: ToJSON
```

sboschman:
Can you please provide an example composition as well on how to use this json transformation to solve the example given in the issue? I must be missing something obvious, but I don't understand how to apply this json transformation to the connection details from the connection secret.

Example based upon [upbound/platform-ref-gcp](https://github.com/upbound/platform-ref-gcp/blob/main/cluster/gke/composition.yaml):
```
kind: Composition
metadata:
  name: gke.gcp.platformref.crossplane.io
  labels:
    provider: GCP
spec:
  writeConnectionSecretsToNamespace: upbound-system
  compositeTypeRef:
    apiVersion: gcp.platformref.crossplane.io/v1alpha1
    kind: GKE
  resources:
    - base:
        apiVersion: container.gcp.crossplane.io/v1beta2
        kind: Cluster
        spec:
          forProvider:
            ...
      connectionDetails:
        - fromConnectionSecretKey: clusterCA
        - fromConnectionSecretKey: kubeconfig
    - base:
        apiVersion:
        kind:
        spec:
          forProvider:
             jsonData: # patched from connection secret
        patches:
          - type: ?
             fromFieldPath: <<< how to reference the connection secret key 'clusterCA'?
             toFieldPath: spec.forProvider.jsonData
             transforms:
              - type: string
                 string:
                   fmt: '{"clusterCA":"%s"}'
              - type: string
                 string:
                   type: Convert
                   convert: ToJSON
```

The examples which reference connection secrets I can find rely on the provider implementing a xxxSecretRef field, e.g.
```
apiVersion: helm.crossplane.io/v1beta1
kind: ProviderConfig
spec:
  credentials:
    source: Secret
    secretRef:
      key: kubeconfig
```

or

```
apiVersion: generic.vault.jet.crossplane.io/v1alpha1
kind: Secret
metadata:
  name: example
spec:
  forProvider:
    path: "secret/foo"
    dataJsonSecretRef:
      key: data_json
      name: example-data
      namespace: default
```

This SecretRef reference seems to be bypassing the patch+transform flow.

An example on how to construct a custom json blob from connection secret details and normalize it with ToJSON would be much appreciated.

jpmcb:
Thanks @negz for the detailed review and thoughts! I'm unfortunately no longer in a position to dedicate my own personal time or resources to moving this forward. I'd be happy for anyone else on the team in in the community to pick up my contribution and finish it. Please feel free to close this once a replacement PR is open! ❤️

braghettos:
@negz I think it would be a nice feature to add to the crossplane functionalities, what do you think?

PR #3144, opened 2022-06-15
feat(xrd): Add composite name generation via go templates

### LINKED ISSUE #3120: Allow deterministic Composite Resource Names

Crossplane composite resource names are generated by the kube API server through setting `generatedName` to `<claim-name>-`. The server will append a random string as suffix to make the object's name unique:

https://github.com/crossplane/crossplane/blob/master/internal/controller/apiextensions/claim/configurator.go#L134

```go
        if !meta.WasCreated(cp) {
		// The API server returns an available name derived from
		// generateName when we perform a dry-run create. This name is
		// likely (but not guaranteed) to be available when we create
		// the composite resource. If the API server generates a name
		// that is unavailable it will return a 500 ServerTimeout error.
		cp.SetGenerateName(fmt.Sprintf("%s-", cm.GetName()))
		return errors.Wrap(c.client.Create(ctx, cp, client.DryRunAll), errName)
	}
```

Because of this randomness, creating the same claim multiple times will result in different composite names which makes migration tasks and traceability very difficult.

Crossplane should offer a second, deterministic name generator.

For example, the generated name could be combination of the claim's name and namespace. The folks at [loft-sh/vcluster](https://github.com/loft-sh/vcluster) do something similar. There the name is a combination of name, namespace and cluster name:

```
<SERVICE_ACCOUNT>-x-<SERVICE_ACCOUNT_NAMESPACE>-x-<VIRTUAL_CLUSTER_NAME>
```

If the name is too long, it is trimmed:

```
<52 Characters of previous mentioned string>-<First 10 Characters SHA256 Hash of previous mentioned string>
```

To ensure backwards compatibility, the feature should be enabled via a feature flag, i.e. `--enable-deterministic-composite-names`.

MisterMX:

This adds the ability to generate custom composite names by passing a mini helm template via the new XRD field `spec.composite.name.template`.

Fixes #3120 

To ensure backward compatibility, the reconciler will fall back to let the Kube API choose an object name via `metadata.generateName`.

An example XRD could look like this (truncated):
```yaml
metadata:
  name: compositetests.test.com
spec:
  composite:
    name:
      template: '{{ printf "%s.%s-%s.%s" .Claim.Namespace .Claim.Kind (sha1sum .Claim.GVK | substr 0 5) .Claim.Name  }}'
```

I have:

Manually, unit tests.


MisterMX:
@negz I understand your reasoning and it makes indeed sense. Go templates are quite complex and can fail (or even panic) for some reason.

Would you accept a different but simpler solution? For example
1. Named string formats, similar to Python, i.e. `"%{ClaimNamespace}.%{ClaimKind}.%{ClaimName}"`
2. Format with [placeholder similar to Git](https://git-scm.com/docs/pretty-formats#Documentation/pretty-formats.txt-emHem): `"%cns.%ck.%cn"`

negz:
> Would you accept a different but simpler solution?

Unfortunately these don't seem much better to me, given they're still kind of templating.

I suppose in a way the Go string format transform is prior art here. In general though I'm wondering if there's some way to allow deterministic names without using templating (or making a breaking change).

MisterMX:
I am closing this for now. We decided internally that we will calculate the name of our managed resources by combining name and namespace, something like

```yaml
- type: CombineFromComposite
  combine:
    variables:
      - metadata.namespace
      - metadata.name
    strategy: string
    string:
      fmt: "%s.%s"
```

This works for us as we can ensure internally, that names don't collide.

PR #3167, opened 2022-07-04
feat(transform): Allow mapping to any JSON value

### LINKED ISSUE #3150: Transform: Support mapping to arbitrary JSON values

Similar to #2887, #2698 and #2588.

Currently, Crossplane's map transform does only allow mapping to string values. It should support any kind of valid JSON / YAML which includes strings, numbers, boolean, objects and arrays.

Instead of a `map[string]string` it should map to a `map[string]extv1.JSON` where [JSON](https://pkg.go.dev/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1#JSON) is a K8s native type that can store any kind of data.

MisterMX:

This modifies the implementation for map transforms to allow mapping to any kind of JSON value (string, number, bool, slice, object).

The API changes are fully backward compatible since a single string does also count as valid JSON.

Fixes #3150 

I have:

Unit tests

### DIFF
--- a/apis/apiextensions/v1/composition_transforms.go
+++ b/apis/apiextensions/v1/composition_transforms.go
@@ -26,6 +26,7 @@ import (
 	"strings"
 
 	"github.com/pkg/errors"
+	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/utils/pointer"
 )
 
@@ -41,6 +42,7 @@ const (
 	errFmtTransformTypeFailed          = "%s transform could not resolve"
 	errFmtMapTypeNotSupported          = "type %s is not supported for map transform"
 	errFmtMapNotFound                  = "key %s is not found in map"
+	errFmtMapInvalidJSON               = "value for key %s is not valid JSON"
 
 	errStringTransformTypeFailed        = "type %s is not supported for string transform type"
 	errStringTransformTypeFormat        = "string transform of type %s fmt is not set"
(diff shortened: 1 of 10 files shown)


MisterMX:
@negz can you take a look at this?

bobh66 (approved):
LGTM - thanks!

nullable-eth (approved):
This looks great, thanks!

apis/apiextensions/v1/composition_transforms.go
```diff
@@ -172,10 +174,14 @@ func (m MapTransform) MarshalJSON() ([]byte, error) {
 func (m *MapTransform) Resolve(input any) (any, error) {
 	switch i := input.(type) {
 	case string:
-		val, ok := m.Pairs[i]
+		raw, ok := m.Pairs[i]
```
negz, line 177:
```suggestion
		p, ok := m.Pairs[i]
```

Nit, just because `raw.Raw` below stutters.

negz (approved):
One nit, but otherwise LGTM!

apis/apiextensions/v1/composition_transforms.go
MisterMX: Good catch. Thanks.

shanproofpoint:
hi can i get a status on this?

jbw976:
hey there @shanproofpoint - looks like this was included in the v1.10 release, release notes can be found at https://github.com/crossplane/crossplane/releases/tag/v1.10.0.  Let us know if you are not seeing the functionality available in v1.10+ releases 🙇

shanproofpoint:
sorry i apologize... my testing method is not quite right. something wrong with my schema. thank you @jbw976

PR #3175, opened 2022-07-07
Update composition.md

MMartyn:
Updating the docs to indicate that your XR definition needs to have the definition for any fields you are planning on writing back to it to share values between resources in your composition.

Making this tip/trick a little clearer as I was caught by this gotcha.

Fixes #

I have:

### DIFF
--- a/docs/reference/composition.md
+++ b/docs/reference/composition.md
@@ -787,9 +787,10 @@ so:
 
 1. Use a `ToCompositeFieldPath` patch to patch from your source composed
    resource to the XR. Typically you'll want to patch to a status field or an
-   annotation. Note: The field you are patching to needs to be defined in the 
-   XR definition. For instance, if you are patching to `.status.foo` then the definition
-   for the XR must define `status.foo`.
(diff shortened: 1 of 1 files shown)


docs/reference/composition.md
```diff
@@ -787,7 +787,9 @@ so:
 
 1. Use a `ToCompositeFieldPath` patch to patch from your source composed
    resource to the XR. Typically you'll want to patch to a status field or an
-   annotation.
+   annotation. Note: The field you are patching to needs to be defined in the 
+   XR definition. For instance, if you are patching to `.status.foo` then the definition
```
bobh66, line 791: This is true for status fields but not for metadata.labels and metadata.annotations which are very often used as "scratchpads" to copy information from one composite to another.   Maybe clarify that this note only applies to status fields and not to annotations or labels.

MMartyn: Good point. I'll update.

negz (approved):
Thanks @MMartyn! I think this is a useful tidbit of information.

PR #3176, opened 2022-07-07
Parallelize package API establisher

### LINKED ISSUE #2564: Installing packages with many CRDs causes reconciler to exceed context deadline
### What happened?

Cannot install a provider package containing 750+ CRDs. 

Trying to install [provider-tf-aws](https://github.com/crossplane-contrib/provider-tf-aws) built from [this commit](https://github.com/crossplane-contrib/provider-tf-aws/tree/daf1e9f7fcbd7959c25ccdcc11e0800d17e904fe), but package is never installed.

I am observing context deadline errors in `providerrevisions.pkg.crossplane.io` events:

```
Spec:
  Desired State:                  Active
  Ignore Crossplane Constraints:  false
  Image:                          turkenh/provider-tf-aws:daf1e9f7-2
  Package Pull Policy:            IfNotPresent
  Revision:                       1
  Skip Dependency Resolution:     false
Status:
  Conditions:
    Last Transition Time:  2021-09-10T09:14:52Z
    Reason:                UnhealthyPackageRevision
    Status:                False
    Type:                  Healthy
  Controller Ref:
    Name:
Events:
  Type     Reason             Age                  From                                         Message
  ----     ------             ----                 ----                                         -------
  Warning  SyncPackage        76s                  packages/providerrevision.pkg.crossplane.io  cannot establish control of object: context deadline exceeded
  Normal   BindClusterRole    16s (x4 over 2m16s)  rbac/providerrevision.pkg.crossplane.io      Bound system ClusterRole to provider ServiceAccount(s)
  Normal   ApplyClusterRoles  16s (x3 over 2m15s)  rbac/providerrevision.pkg.crossplane.io      Applied RBAC ClusterRoles
  Warning  SyncPackage        16s                  packages/providerrevision.pkg.crossplane.io  cannot establish control of object: rate: Wait(n=1) would exceed context deadline
```

### How can we reproduce it?

`kubectl crossplane install provider turkenh/provider-tf-aws:daf1e9f7-2`

Crossplane version: v1.3.1
Kubernetes Cluster: Kind - v1.21.1

Possibly related: https://github.com/crossplane-contrib/terrajet/issues/47

### LINKED ISSUE #2849: ConfigurationRevision: cannot establish control of object: context deadline exceeded
### What happened?

Crossplane runs into a timeout when reconciling a `ConfigurationRevision`:

```
Warning  SyncPackage  50m   packages/configurationrevision.pkg.crossplane.io   cannot establish control of object: context deadline exceeded
```

### How can we reproduce it?

We don't have a clear way of reproducing it, but it might be a scaling issue.

Our configuration package contains roundabout 30 compositions and as XRDs. In addition, we have about 800 composite resources and 250 CRDs installed on our cluster.

Crossplane version: 1.6.1

avalanche123:

Fixes #2564
Fixes #2849
Closes #3168

I have:

installed Crossplane into a kind cluster and I see client side throttling:

image

### DIFF
--- a/internal/controller/pkg/revision/establisher.go
+++ b/internal/controller/pkg/revision/establisher.go
@@ -128,132 +128,129 @@ func (e *APIEstablisher) validate(ctx context.Context, objs []runtime.Object, pa
 	}
 	g, ctx := errgroup.WithContext(ctx)
 	g.SetLimit(maxConcurrentEstablishers)
-	out := make(chan currentDesired, maxConcurrentEstablishers)
-	go func() {
-		for _, res := range objs {
-			res := res // Pin the range variable before using it in a Goroutine.
-			g.Go(func() error {
-				// Assert desired object to resource.Object so that we can access its
-				// metadata.
-				d, ok := res.(resource.Object)
-				if !ok {
-					return errors.New(errAssertResourceObj)
-				}
+	out := make(chan currentDesired, len(objs))
+	for _, res := range objs {
+		res := res // Pin the range variable before using it in a Goroutine.
+		g.Go(func() error {
+			// Assert desired object to resource.Object so that we can access its
+			// metadata.
+			d, ok := res.(resource.Object)
+			if !ok {
+				return errors.New(errAssertResourceObj)
+			}
 
-				// The generated webhook configurations have a static hard-coded name
-				// that the developers of the providers can't affect. Here, we make sure
-				// to distinguish one from the other by setting the name to the parent
-				// since there is always a single ValidatingWebhookConfiguration and/or
-				// single MutatingWebhookConfiguration object in a provider package.
-				// See https://github.com/kubernetes-sigs/controller-tools/issues/658
-				switch conf := res.(type) {
-				case *admv1.ValidatingWebhookConfiguration:
+			// The generated webhook configurations have a static hard-coded name
+			// that the developers of the providers can't affect. Here, we make sure
+			// to distinguish one from the other by setting the name to the parent
+			// since there is always a single ValidatingWebhookConfiguration and/or
+			// single MutatingWebhookConfiguration object in a provider package.
+			// See https://github.com/kubernetes-sigs/controller-tools/issues/658
+			switch conf := res.(type) {
+			case *admv1.ValidatingWebhookConfiguration:
+				if len(webhookTLSCert) == 0 {
+					return nil
+				}
+				if pkgRef, ok := GetPackageOwnerReference(parent); ok {
+					conf.SetName(fmt.Sprintf("crossplane-%s-%s", strings.ToLower(pkgRef.Kind), pkgRef.Name))
+				}
+				for i := range conf.Webhooks {
+					conf.Webhooks[i].ClientConfig.CABundle = webhookTLSCert
+					if conf.Webhooks[i].ClientConfig.Service == nil {
+						conf.Webhooks[i].ClientConfig.Service = &admv1.ServiceReference{}
+					}
+					conf.Webhooks[i].ClientConfig.Service.Name = parent.GetName()
+					conf.Webhooks[i].ClientConfig.Service.Namespace = e.namespace
+					conf.Webhooks[i].ClientConfig.Service.Port = pointer.Int32(webhookPort)
+				}
+			case *admv1.MutatingWebhookConfiguration:
+				if len(webhookTLSCert) == 0 {
+					return nil
+				}
+				if pkgRef, ok := GetPackageOwnerReference(parent); ok {
+					conf.SetName(fmt.Sprintf("crossplane-%s-%s", strings.ToLower(pkgRef.Kind), pkgRef.Name))
+				}
+				for i := range conf.Webhooks {
+					conf.Webhooks[i].ClientConfig.CABundle = webhookTLSCert
+					if conf.Webhooks[i].ClientConfig.Service == nil {
+						conf.Webhooks[i].ClientConfig.Service = &admv1.ServiceReference{}
+					}
+					conf.Webhooks[i].ClientConfig.Service.Name = parent.GetName()
+					conf.Webhooks[i].ClientConfig.Service.Namespace = e.namespace
+					conf.Webhooks[i].ClientConfig.Service.Port = pointer.Int32(webhookPort)
+				}
+			case *extv1.CustomResourceDefinition:
+				if conf.Spec.Conversion != nil && conf.Spec.Conversion.Strategy == extv1.WebhookConverter {
 					if len(webhookTLSCert) == 0 {
-						return nil
+						return errors.New(errConversionWithNoWebhookCA)
 					}
-					if pkgRef, ok := GetPackageOwnerReference(parent); ok {
-						conf.SetName(fmt.Sprintf("crossplane-%s-%s", strings.ToLower(pkgRef.Kind), pkgRef.Name))
+					if conf.Spec.Conversion.Webhook == nil {
+						conf.Spec.Conversion.Webhook = &extv1.WebhookConversion{}
 					}
-					for i := range conf.Webhooks {
-						conf.Webhooks[i].ClientConfig.CABundle = webhookTLSCert
-						if conf.Webhooks[i].ClientConfig.Service == nil {
-							conf.Webhooks[i].ClientConfig.Service = &admv1.ServiceReference{}
-						}
-						conf.Webhooks[i].ClientConfig.Service.Name = parent.GetName()
-						conf.Webhooks[i].ClientConfig.Service.Namespace = e.namespace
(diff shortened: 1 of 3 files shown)


negz:
Thanks for opening this @avalanche123! Does it appear to help with the underlying issue - i.e. help packages install before they hit the context deadline? I've found that installing the three Terrajet providers is a good way to tax the package manager:

```console
kubectl crossplane install provider crossplane/provider-jet-gcp:v0.2.0-preview
kubectl crossplane install provider crossplane/provider-jet-azure:v0.11.0-preview
kubectl crossplane install provider crossplane/provider-jet-aws:v0.4.0-preview
```

In particular provider-aws usually hits the deadline a lot while it tries to unpack its ~900 CRDs.

negz:
WRT the rate limiting, we can probably tune that: 

https://github.com/negz/crossplane/blob/1c57786e3/cmd/crossplane/core/core.go#L86

I'm pretty sure `ratelimiter.LimitRESTConfig` will set the client-side rate limits to 2x the supplied reconcile per second, so giving a higher `--max-reconcile-rate` should bump it.

avalanche123:
after bumping `--max-reconcile-rate=100` all 3 providers install in ~2 minutes without bumping the reconcile timeout:

image

avalanche123:
@negz parallelization allow installing the providers within the timeout, however, not without bumping `--max-reconcile-rate` as well since otherwise client-side throttling slows the process down. we might want to merge #3168 as well. wdyt?

internal/controller/pkg/revision/establisher.go
```diff
@@ -94,7 +101,7 @@ type currentDesired struct {
 // Establish checks that control or ownership of resources can be established by
 // parent, then establishes it.
 func (e *APIEstablisher) Establish(ctx context.Context, objs []runtime.Object, parent v1.PackageRevision, control bool) ([]xpv1.TypedReference, error) { // nolint:gocyclo
-	allObjs := []currentDesired{}
+	allObjs := make(chan currentDesired, len(objs))
```
negz, line 104:
```suggestion
	allObjs := make(chan currentDesired, maxConcurrentReconciles)
```

Would this make sense, or is there an advantage to letting the entire set of objects into the buffer?

internal/controller/pkg/revision/establisher.go
```diff
@@ -108,136 +115,165 @@ func (e *APIEstablisher) Establish(ctx context.Context, objs []runtime.Object, p
...
 
-	for _, cd := range allObjs {
-		if !cd.Exists {
-			// Only create a missing resource if we are going to control it.
-			// This prevents an inactive revision from racing to create a
-			// resource before an active revision of the same parent.
-			if control {
-				if err := e.create(ctx, cd.Desired, parent); err != nil {
-					return nil, err
+	refsCh := make(chan xpv1.TypedReference, len(allObjs))
```
negz, line 242:
```suggestion
	refsCh := make(chan xpv1.TypedReference, maxConcurrentReconciles)
```

Similar to my comment above, would this make sense?

internal/controller/pkg/revision/establisher_test.go
```diff
@@ -402,7 +403,12 @@ func TestAPIEstablisherEstablish(t *testing.T) {
 			if diff := cmp.Diff(tc.want.err, err, test.EquateErrors()); diff != "" {
 				t.Errorf("\n%s\ne.Check(...): -want error, +got error:\n%s", tc.reason, diff)
 			}
-			if diff := cmp.Diff(tc.want.refs, refs, test.EquateErrors()); diff != "" {
+			trans := cmp.Transformer("Sort", func(refs []xpv1.TypedReference) []xpv1.TypedReference {
+				out := append([]xpv1.TypedReference(nil), refs...) // Copy input to avoid mutating it
+				sort.SliceStable(out, func(i, j int) bool { return out[i].Name < out[j].Name })
+				return out
+			})
```
negz, line 410: Could you use the built in SortSlices option instead?

https://pkg.go.dev/github.com/google/go-cmp/cmp/cmpopts#SortSlices

negz (commented):
Awesome work @avalanche123 - this implementation using errgroup is very clean and 2 minutes to get everything installed is a big improvement.

> @negz parallelization allows installing the providers within the timeout, however, not without bumping --max-reconcile-rate as well since otherwise client-side throttling slows the process down. we might want to merge https://github.com/crossplane/crossplane/pull/3168 as well. wdyt?

I'd prefer to tune Crossplane such that we don't hit rate limits with the default of `--max-reconcile-rate=10`. I'm pretty sure the only thing setting a higher `--max-reconcile-rate` is doing for us in this scenario is raising the client-side rate limits, which are derived from that flag. Specifically we allow 2x the value per https://github.com/crossplane/crossplane-runtime/blob/a520b60f/pkg/ratelimiter/default.go#L75.

Looking at that code, it's almost certainly broken. It seems like we set burst on the token bucket rate limiter to exactly `3` - I'm pretty sure it's supposed to be `--max-reconcile-rate` x 3. In general I'm happy to be allow for much higher rate limits - we should be able to rely on https://kubernetes.io/docs/concepts/cluster-administration/flow-control/ instead of client-side rate limiting.

internal/controller/pkg/revision/establisher.go
```diff
@@ -39,6 +40,12 @@ import (
 	v1 "github.com/crossplane/crossplane/apis/pkg/v1"
 )
 
+const (
+	// maxConcurrentReconciles specifies the maximum number of goroutines to use
+	// for updating resources.
+	maxConcurrentReconciles = 10
```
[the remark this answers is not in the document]
avalanche123: agreed. I'll rename it to `maxConcurrentEstablishers`

internal/controller/pkg/revision/establisher.go
avalanche123: since we're validating all resources before syncing them, we have to buffer the results otherwise the `g.Wait()` below will end up blocking forever.

internal/controller/pkg/revision/establisher.go
```diff
@@ -108,136 +115,165 @@ func (e *APIEstablisher) Establish(ctx context.Context, objs []runtime.Object, p
...
-				continue
-			}
-			if pkgRef, ok := GetPackageOwnerReference(parent); ok {
-				conf.SetName(fmt.Sprintf("crossplane-%s-%s", strings.ToLower(pkgRef.Kind), pkgRef.Name))
-			}
-			for i := range conf.Webhooks {
-				conf.Webhooks[i].ClientConfig.CABundle = webhookTLSCert
-				if conf.Webhooks[i].ClientConfig.Service == nil {
-					conf.Webhooks[i].ClientConfig.Service = &admv1.ServiceReference{}
+	{
```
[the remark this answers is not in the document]
avalanche123: I've extracted both into private methods. I used it to avoid name conflicts for 2 `errgroup`s in this method both named `g`

internal/controller/pkg/revision/establisher_test.go
avalanche123: nice! done!

internal/controller/pkg/revision/establisher.go
```diff
@@ -108,138 +126,185 @@ func (e *APIEstablisher) Establish(ctx context.Context, objs []runtime.Object, p
...
-		// Assert desired object to resource.Object so that we can access its
-		// metadata.
-		d, ok := res.(resource.Object)
-		if !ok {
-			return nil, errors.New(errAssertResourceObj)
-		}
+	g, ctx := errgroup.WithContext(ctx)
+	g.SetLimit(maxConcurrentEstablishers)
+	out := make(chan currentDesired, maxConcurrentEstablishers)
+	go func() {
```
negz, line 132: It wasn't clear to me what this extra "layer" of goroutine was doing at first, then I realized that it's helping avoid a deadlock:

* I had suggested we cap the output channel buffer at `maxConcurrentEstablishers`.
* Therefore we need to range over the channel so that the Goroutines won't deadlock on channel writes.
* Therefore we need to close the channel before our `range` will complete.

In retrospect I think you had the right idea to begin with, sorry - my bad! I found this code much easier to follow when you let `len(objs)` results accumulate in the output channel. That way we could:

1. Call `g.Go` as many times as we need.
2. Call `g.Wait()` to let them all finish
3. `close` the output channel and range over it.

I like that you now have everything broken out into methods but would you mind removing the middle layer of Goroutine? Once that's done I'm ready to approve.

FWIW I drafted up a commit that does just this as part of trying to understand what was happening here - https://github.com/crossplane/crossplane/commit/5a665eb

negz (changes requested):

internal/controller/pkg/revision/establisher.go
avalanche123: applied your commit

PR #3183, opened 2022-07-14
Bump crossplane-runtime and build submodule

negz:

I believe this will get us to sub 2 mins to install the "big three" jet providers (~1,800 CRDs).

Note that the `xpv1.Reference` type changed in the latest crossplane-runtime to include a `Policy` field. We don't use this field in core Crossplane so I've replaced our handful of `xpv1.Reference` uses with local structs. These reference structs are simple enough that it seemed better to define our own than to depend on (and have to name!) a new kind of reference in crossplane-runtime.

I have:

I've successfully deployed the "getting started with GCP" configuration from the https://crossplane.io documentation (with composition revisions enabled).


negz:
CC @avalanche123

hasheddan (approved):

negz:
Setting this to draft - looks like the below PR updated some reference types to add a policy field. We use those reference types in some places in core that won't honor the policy. I don't think we can use these types in core.


https://github.com/crossplane/crossplane-runtime/pull/328/files#diff-0d8ebf4c1879297c0b02781d58ee25833e3e21e993939eda922e3ef49a7e417cR120

MisterMX:
Can we also include https://github.com/upbound/build/pull/192 once its merged?

negz:
> Can we also include https://github.com/upbound/build/pull/192 once its merged?

Yes, will do.

apis/apiextensions/v1/composition_types.go
```diff
@@ -53,11 +51,19 @@ type CompositionSpec struct {
...
 	// provisioned using this composition will be published.
 	// +optional
 	// +kubebuilder:default={"name": "default"}
-	PublishConnectionDetailsWithStoreConfigRef *xpv1.Reference `json:"publishConnectionDetailsWithStoreConfigRef,omitempty"`
+	PublishConnectionDetailsWithStoreConfigRef *StoreConfigReference `json:"publishConnectionDetailsWithStoreConfigRef,omitempty"`
+}
+
+// A StoreConfigReference references a secret store config that may be used to
+// write connection details.
+// secrets
```
hasheddan, line 63:
```suggestion
```

apis/apiextensions/v1alpha1/revision_types.go
```diff
@@ -74,17 +74,25 @@ type CompositionRevisionSpec struct {
...
+	PublishConnectionDetailsWithStoreConfigRef *StoreConfigReference `json:"publishConnectionDetailsWithStoreConfigRef,omitempty"`
 
 	// Revision number. Newer revisions have larger numbers.
 	// +immutable
 	Revision int64 `json:"revision"`
 }
 
+// A StoreConfigReference references a secret store config that may be used to
+// write connection details.
+// secrets
```
hasheddan, line 90:
```suggestion
// secrets
```
```suggestion
```

hasheddan (approved):
@negz two small nitpicks (which I _think_ were unintentional but perhaps this is an annotation I'm not familiar with) -- otherwise LGTM!

apis/apiextensions/v1/composition_types.go
negz: Thanks! Yes this was unintentional.

PR #3213, opened 2022-07-29
Add controller reference to existing owner references

### LINKED ISSUE #3211: Rendering of composite removes external ownerReferences
### What happened?
The metadata.ownerReferences attribute of all composed objects is restricted to contain a single entry for the crossplane controlling resource:

https://github.com/crossplane/crossplane/blob/master/internal/controller/apiextensions/composite/composed.go#L366

This eliminates the ability to represent multiple owner relationships between resources and implement functionality like foreground cascading deletion.

### How can we reproduce it?
Create a composition with two resources and create an entry in each resource definition metadata.ownerReferences that points at the other resource (as a trivial example)

Observe that when the composition is rendered the specified ownerReferences are missing and the only owner is the parent composite resource.

Crossplane version: 1.9

This is an element of discussion #3116 

As discussed with @negz on the community call, this is likely a bug and can be relaxed to allow other owners as long as the only Controller resource remains the same as it is today.

I'll push a PR for review and discussion.

bobh66:

Modified composed.Render() to add the controller reference to the existing list of ownerReferences rather than replacing the existing references with the controller reference.   This will allow users to specify ownerReferences in compositions or via external entities such as Kyverno policies.

This is also one step toward supporting foreground cascading deletion.

Note that if the user should try to specify an ownerReference with controller: true the render will fail with an error.

Also note that the required error check for this change increased the complexity of the function to 11, so I disabled the lint check.   If there is a better solution that leaves the lint check enabled I can refactor it.

Fixes #3211 

I have:

Added a unit test and also verified in our environment that a composition can now specify an ownerReference for a different resource and it will not be removed when the composition is rendered.

### DIFF
--- a/internal/controller/apiextensions/composite/composed.go
+++ b/internal/controller/apiextensions/composite/composed.go
@@ -59,6 +59,7 @@ const (
 	errFmtConnDetailKey  = "connection detail of type %q key is not set"
 	errFmtConnDetailVal  = "connection detail of type %q value is not set"
 	errFmtConnDetailPath = "connection detail of type %q fromFieldPath is not set"
+	errSetControllerRef  = "cannot set controller reference"
 )
 
 // Annotation keys.
@@ -315,7 +316,7 @@ func NewAPIDryRunRenderer(c client.Client) *APIDryRunRenderer {
 // Render the supplied composed resource using the supplied composite resource
 // and template. The rendered resource may be submitted to an API server via a
 // dry run create in order to name and validate it.
-func (r *APIDryRunRenderer) Render(ctx context.Context, cp resource.Composite, cd resource.Composed, t v1.ComposedTemplate) error {
+func (r *APIDryRunRenderer) Render(ctx context.Context, cp resource.Composite, cd resource.Composed, t v1.ComposedTemplate) error { // nolint:gocyclo
 	kind := cd.GetObjectKind().GroupVersionKind().Kind
 	name := cd.GetName()
 	namespace := cd.GetNamespace()
@@ -360,10 +361,11 @@ func (r *APIDryRunRenderer) Render(ctx context.Context, cp resource.Composite, c
 		SetCompositionResourceName(cd, *t.Name)
 	}
 
-	// We do this last to ensure that a Composition cannot influence owner (and
-	// especially controller) references.
+	// We do this last to ensure that a Composition cannot influence controller references.
 	or := meta.AsController(meta.TypedReferenceTo(cp, cp.GetObjectKind().GroupVersionKind()))
-	cd.SetOwnerReferences([]metav1.OwnerReference{or})
+	if err := meta.AddControllerReference(cd, or); err != nil {
+		return errors.Wrap(err, errSetControllerRef)
+	}
 
 	// We don't want to dry-run create a resource that can't be named by the API
 	// server due to a missing generate name. We also don't want to create one
--- a/internal/controller/apiextensions/composite/composed_test.go
+++ b/internal/controller/apiextensions/composite/composed_test.go
@@ -38,6 +38,7 @@ import (
 	"github.com/crossplane/crossplane-runtime/pkg/resource/fake"
 	"github.com/crossplane/crossplane-runtime/pkg/resource/unstructured/composed"
 	"github.com/crossplane/crossplane-runtime/pkg/test"
+
 	v1 "github.com/crossplane/crossplane/apis/apiextensions/v1"
 	"github.com/crossplane/crossplane/internal/xcrd"
 )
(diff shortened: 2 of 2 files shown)


muvaf (approved):
LGTM from code perspective but I know you chatted with @negz regarding the implications, so I'll leave it to him for final approval & merge.

internal/controller/apiextensions/composite/composed.go
```diff
@@ -360,10 +361,11 @@ func (r *APIDryRunRenderer) Render(ctx context.Context, cp resource.Composite, c
 		SetCompositionResourceName(cd, *t.Name)
 	}
 
-	// We do this last to ensure that a Composition cannot influence owner (and
-	// especially controller) references.
```
negz, line 364: This comment (which I'm pretty sure I wrote) make it sound like I actively didn't want a Composition to be able to set owner references. I can't remember why, though. 🤔

internal/controller/apiextensions/composite/composed_test.go
```diff
@@ -234,6 +235,35 @@ func TestRender(t *testing.T) {
...
+					GenerateName: "ola-",
+					Labels: map[string]string{
+						xcrd.LabelKeyNamePrefixForComposed: "ola",
+						xcrd.LabelKeyClaimName:             "rola",
+						xcrd.LabelKeyClaimNamespace:        "rolans",
+					},
+					OwnerReferences: []metav1.OwnerReference{{Controller: &ctrl,
+						UID: "random_uid"}},
+				}},
+				err: errors.Wrap(errors.Errorf("cd is already controlled by   (UID random_uid)"), errSetControllerRef),
```
negz, line 264: The format of this error returned by `meta.AddControllerReference` seems like it might be a bug. 🤔

negz:
@bobh66 Thank you for the deep exploration and context in #3116!

I feel okay making this change given that I don't recall why we had this limitation in the first place (i.e. whether it was intentional or accidental). Some quick questions before we merge:

> Create a composition with two resources and create an entry in each resource definition metadata.ownerReferences that points at the other resource (as a trivial example)

Does this result in owner refs without UIDs set, or are you able to patch them between composed resources? I ask because I wonder whether not having a UID would confuse the Kubernetes garbage collector.

> After some more research and testing, it's clear that foreground cascading deletion on it's own is not a complete solution, as it results in all of the "leaf" resources being marked deleted, with no foregroundDeletion finalizer, regardless of the dependencies of the higher level composites. In addition to the changes described above it's necessary to manage dependencies on composite resources using finalizers on the entire tree of dependencies to prevent deletion of any of the component resources of the composite before the dependency on the composite is removed.

If I follow correctly this change is still valuable despite the above, right?

negz (approved):

bobh66:
> > Create a composition with two resources and create an entry in each resource definition metadata.ownerReferences that points at the other resource (as a trivial example)
> 
> Does this result in owner refs without UIDs set, or are you able to patch them between composed resources? I ask because I wonder whether not having a UID would confuse the Kubernetes garbage collector.

The UIDs can be patched into the ownerReferences.  UID is required so the ownerReference is not created until the UID is available, which means the trivial example might not be quite so trivial, but only one ownerReference is required to prove the point.  I'm not sure why I described two.

> > After some more research and testing, it's clear that foreground cascading deletion on it's own is not a complete solution, as it results in all of the "leaf" resources being marked deleted, with no foregroundDeletion finalizer, regardless of the dependencies of the higher level composites. In addition to the changes described above it's necessary to manage dependencies on composite resources using finalizers on the entire tree of dependencies to prevent deletion of any of the component resources of the composite before the dependency on the composite is removed.
> 
> If I follow correctly this change is still valuable despite the above, right?

Yes - in fact this change is foundational to the rest of the changes that are needed for #3116 and it's related issues.    #3225 builds on this change by adding support for "external" finalizers, and I will have a proposal soon that builds on #3225.

Thanks!

internal/controller/apiextensions/composite/composed.go
bobh66: I can see where you would be concerned about a user constructing an ownerReference that would interfere with the Crossplane controller's ownerReference, so it made sense to me that you might want to block it if no one was using it.

internal/controller/apiextensions/composite/composed_test.go
bobh66: Would you like me to reformat?

negz: No - just thinking out loud. I'm pretty sure this would need to be fixed in crossplane-runtime. You're welcome to make that fix if you like but it definitely doesn't need to block this PR.

PR #3223, opened 2022-08-05
Retry fetching package manifest using a `GET` request after a `HEAD` …

### LINKED ISSUE #3222: Failed to pull package from harbor registry
### What happened?

Install custom configuration from internal harbor registry failed:
```
Events:
  Type     Reason         Age                     From                                      Message
  ----     ------         ----                    ----                                      -------
  Warning  UnpackPackage  2m57s (x1249 over 19h)  packages/configuration.pkg.crossplane.io  cannot unpack package: failed to fetch package digest from remote: HEAD https://registry.xxx.com/v2/iac_test/k8s-iac/crossplane-iac-provider/manifests/main: response did not include Docker-Content-Digest header
```
But `docker pull` works fine
```
docker pull registry.xxx.com/iac_test/k8s-iac/crossplane-iac-provider
Using default tag: latest
latest: Pulling from iac_test/k8s-iac/crossplane-iac-provider
cxxxxxb9a2: Pull complete
Digest: sha256:cxxxxx5
Status: Downloaded newer image for registry.xxx.com/iac_test/k8s-iac/crossplane-iac-provider:latest
registry.xxx.com/iac_test/k8s-iac/crossplane-iac-provider:latest
```
### How can we reproduce it?

I find out some issues that may be related, #1957 , [NEXUS-12684](https://issues.sonatype.org/browse/NEXUS-12684), [NEXUS-27619](https://issues.sonatype.org/browse/NEXUS-27619) and [NEXUS-29640](https://issues.sonatype.org/browse/NEXUS-29640)

Crossplane version: v1.10.0-rc.0.40.g76fab9fe

ryeyao:
…request failure. For some registry servers may not properly support a `HEAD` request.

Signed-off-by: Rye Yao <rye.y.cn@gmail.com>

Added a retry in function `K8sFetcher.Head`, according to what the go-containerregistry function `remote.Head` comment says
```
// Head returns a v1.Descriptor for the given reference by issuing a HEAD
// request.
//
// Note that the server response will not have a body, so any errors encountered
// should be retried with Get to get more details.
func Head(ref name.Reference, options ...Option) (*v1.Descriptor, error) {
...
```
Fixes #3222 

I have:

Verified on our internal environment, and installing package from our internal harbor registry worked as expected.

### DIFF
--- a/internal/xpkg/fetch.go
+++ b/internal/xpkg/fetch.go
@@ -136,7 +136,15 @@ func (i *K8sFetcher) Head(ctx context.Context, ref name.Reference, secrets ...st
 	if err != nil {
 		return nil, err
 	}
-	return remote.Head(ref, remote.WithAuthFromKeychain(auth), remote.WithTransport(i.transport), remote.WithContext(ctx))
+	d, err := remote.Head(ref, remote.WithAuthFromKeychain(auth), remote.WithTransport(i.transport), remote.WithContext(ctx))
+	if err != nil || d == nil {
+		rd, gErr := remote.Get(ref, remote.WithAuthFromKeychain(auth), remote.WithTransport(i.transport), remote.WithContext(ctx))
+		if gErr != nil {
+			return nil, errors.Wrapf(gErr, "failed to fetch package descriptor with a GET request after a previous HEAD request failure: %v", err)
+		}
+		return &rd.Descriptor, nil
+	}
+	return d, nil
 }
 
 // Tags fetches a package's tags.


hasheddan (commented):
Thanks for working on this @ryeyao! Is this documented behavior of Harbor? I am surprised to see this is not supported as it is defined in the [OCI distribution spec](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#checking-if-content-exists-in-the-registry).

ryeyao:
> Thanks for working on this @ryeyao! Is this documented behavior of Harbor? I am surprised to see this is not supported as it is defined in the [OCI distribution spec](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#checking-if-content-exists-in-the-registry).

Thank you for your reply @hasheddan ! This is not documented behavior of Harbor. It's more of a client-side legacy compatibility thing according to the [OCI distribution spec](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#legacy-docker-support-http-headers), 

> Legacy Docker support HTTP headers
Because of the origins this specification, the client MAY encounter Docker-specific headers, such as Docker-Content-Digest, or Docker-Distribution-API-Version. These headers are OPTIONAL and clients SHOULD NOT depend on them.

negz:
> It's more of a client-side legacy compatibility thing according to the [OCI distribution spec](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#legacy-docker-support-http-headers),

I'm far from an expert on this area, but what you quoted appears to be discussing Docker HTTP headers, while your change appears to be something different - falling back to a GET if HEAD fails. Can you explain how they're related?

ryeyao:
> > It's more of a client-side legacy compatibility thing according to the [OCI distribution spec](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#legacy-docker-support-http-headers),
> 
> I'm far from an expert on this area, but what you quoted appears to be discussing Docker HTTP headers, while your change appears to be something different - falling back to a GET if HEAD fails. Can you explain how they're related?

Hi @negz , I'm no expert on this area either:). Simply speaking, falling back to a GET if HEAD fails follows what the [google-containerregistry's API reference](https://pkg.go.dev/github.com/google/go-containerregistry/pkg/v1/remote#Head) suggests.
> Head returns a v1.Descriptor for the given reference by issuing a HEAD request
> 
> Note that the server response will not have a body, so any errors encountered should be retried with Get to get more details.

And in my case(#3222 ), the HEAD fails because the server response did not include Docker-Content-Digest header, which is optional(as we discussed above), so it should be retried with a GET according to the API spec. 

I've also found similar issue & fix in repo containerd https://github.com/containerd/containerd/issues/3238,  https://github.com/containerd/containerd/pull/3245

nullable-eth:
> > > It's more of a client-side legacy compatibility thing according to the [OCI distribution spec](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#legacy-docker-support-http-headers),
> > 
> > 
> > I'm far from an expert on this area, but what you quoted appears to be discussing Docker HTTP headers, while your change appears to be something different - falling back to a GET if HEAD fails. Can you explain how they're related?
> 
> Hi @negz , I'm no expert on this area either:). Simply speaking, falling back to a GET if HEAD fails follows what the [google-containerregistry's API reference](https://pkg.go.dev/github.com/google/go-containerregistry/pkg/v1/remote#Head) suggests.
> 
> > Head returns a v1.Descriptor for the given reference by issuing a HEAD request
> > Note that the server response will not have a body, so any errors encountered should be retried with Get to get more details.
> 
> And in my case(#3222 ), the HEAD fails because the server response did not include Docker-Content-Digest header, which is optional(as we discussed above), so it should be retried with a GET according to the API spec.
> 
> I've also found similar issue & fix in repo containerd [containerd/containerd#3238](https://github.com/containerd/containerd/issues/3238), [containerd/containerd#3245](https://github.com/containerd/containerd/pull/3245)

I agree it's pretty clear that remote.HEAD [incorrectly relies on an OPTIONAL legacy header](https://github.com/google/go-containerregistry/blob/e2d575cf2c7deaf939cccfa8bd88045e232f29bc/pkg/v1/remote/descriptor.go#L338) and should be retried with a GET request.

nullable-eth (approved):

internal/xpkg/fetch.go
```diff
@@ -136,7 +136,15 @@ func (i *K8sFetcher) Head(ctx context.Context, ref name.Reference, secrets ...st
 	if err != nil {
 		return nil, err
 	}
-	return remote.Head(ref, remote.WithAuthFromKeychain(auth), remote.WithTransport(i.transport), remote.WithContext(ctx))
+	d, err := remote.Head(ref, remote.WithAuthFromKeychain(auth), remote.WithTransport(i.transport), remote.WithContext(ctx))
+	if err != nil || d == nil {
+		rd, gErr := remote.Get(ref, remote.WithAuthFromKeychain(auth), remote.WithTransport(i.transport), remote.WithContext(ctx))
+		if gErr != nil {
+			return nil, errors.Wrapf(gErr, "Failed to fetch package descriptor with a GET request after a previous HEAD request failure: %v", err)
```
hasheddan, line 143: Was initially wondering if `remote.Head` was returning an error on successful responses that just didn't include the digest header, but it looks like it is, so this error message should always have a valid `err`. 👍🏻 [ref](https://github.com/google/go-containerregistry/blob/87b3a7966622dbfdb8cbb9c4b5068884540d387e/pkg/v1/remote/descriptor.go#L340)

hasheddan:
```suggestion
			return nil, errors.Wrapf(gErr, "failed to fetch package descriptor with a GET request after a previous HEAD request failure: %v", err)
```

ref: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

hasheddan (approved):
Thanks @ryeyao! One small nitpick but otherwise LGTM 👍🏻 I think this fix is worthy of backporting as it is correcting a behavior that should be supported -- I'll add the necessary labels

PR #3250, opened 2022-08-19
Fix typo in registry self-signed certs guide

hasheddan:

Fixes a typo of Configure in the registry self-signed certs guide.

I have:

n/a -- docs

### DIFF
--- a/docs/guides/self-signed-ca-certs.md
+++ b/docs/guides/self-signed-ca-certs.md
@@ -17,7 +17,7 @@ Crossplane needs to be installed via the Helm chart with the
 `registryCaBundleConfig.name` and `registryCaBundleConfig.key` parameters 
 defined. See [Install Crossplane].
 
-## Conifgure
+## Configure
 
 1. Create a CA Bundle (A file containing your Root and Intermediate 
 certificates in a specific order). This can be done with any text editor or 


jbw976 (approved):

hasheddan (commented):
@jbw976 hmm interesting 🤔 looks like your review doesn't actually allow for merge since the steering group isn't in the codeowners here -- I'm going to see if I can update.

hasheddan:
Actually, I think this may be working as intended. `docs` is falling through to the fallback, which is `crossplane/reviewers` or `crossplane/maintainers`. It probably makes sense for the steering committee to not be getting assigned to random reviews, and it looks like the `crossplane/steering-committee` assigns all three members. However, I do wonder if some steering committee members who also want to be more active in reviewing (i.e. maybe @jbw976) should be added to the `crossplane/maintainers` group as well?

muvaf (approved):

jbw976:
>  I do wonder if some steering committee members who also want to be more active in reviewing (i.e. maybe @jbw976) should be added to the crossplane/maintainers group as well?

I'll try to earn it back over the next few months as I get more involved with community/reviews 💪 🙇‍♂️

PR #3251, opened 2022-08-20
Support float64 multiplication #3059

### LINKED ISSUE #3059: Enhance `multiply` of transform `math` type with support of `float64`

knowshan:
Enhance multiply of transform math type with support of float64
Fixes #3059 

I have:

* Added new test and ran it manually
* Ran `make reviewable` 

### DIFF
--- a/apis/apiextensions/v1/composition_transforms.go
+++ b/apis/apiextensions/v1/composition_transforms.go
@@ -137,6 +137,8 @@ func (m *MathTransform) Resolve(input any) (any, error) {
 		return *m.Multiply * i, nil
 	case int:
 		return *m.Multiply * int64(i), nil
+	case float64:
+		return float64(*m.Multiply) * i, nil
 	default:
 		return nil, errors.New(errMathInputNonNumber)
 	}
--- a/apis/apiextensions/v1/composition_transforms_test.go
+++ b/apis/apiextensions/v1/composition_transforms_test.go
@@ -132,6 +132,15 @@ func TestMathResolve(t *testing.T) {
 				o: 3 * m,
 			},
 		},
+		"SuccessFloat64": {
+			args: args{
+				multiplier: &m,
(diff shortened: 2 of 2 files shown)


apis/apiextensions/v1/composition_transforms.go
```diff
@@ -137,6 +137,8 @@ func (m *MathTransform) Resolve(input any) (any, error) {
 		return *m.Multiply * i, nil
 	case int:
 		return *m.Multiply * int64(i), nil
+	case float64:
+		return float64(*m.Multiply) * i, nil
```
negz, line 141: Given that actual inputs are never float64 (Kubernetes encoding requires they always be int64s), is this intended to cover the case in which `input` has been converted to a float64 in a previous transform?

ytsarev (commented):
This change needs to be tested with a Composition patch from #3059 

```
      transforms:
      - type: convert
        convert:
          toType: float64
      - type: math
        math:
          multiply: 90
```

and documented with example in a https://github.com/crossplane/crossplane/blob/master/docs/reference/composition.md#transform-types

apis/apiextensions/v1/composition_transforms.go
MisterMX: Is this actually guaranteed by the parser or just a frontend validation?

Given that the JSON spec does not distinguish between integers and floats and only defines an abstract `number` type (like in Javascript). This means in Go, all numbers that are parsed with `json.Unmarshall` (which is used by k8s-yaml internally) are converted to float64 by default if the type is not known, like in `Unstructured` that is used for XRs and XRCs.

See https://go.dev/play/p/0CI40vKXcnc

PR #3258, opened 2022-08-22
Ensure the stale workflow uses fewer operations

negz:

We only get 1,000 operations per hour for every workflow that uses GITHUB_TOKEN. We're now seeing the stale workflow use relatively few operations with few to no stale issues open.

In addition to lowering the workflow's operation limit this change has the workflow run once a day. It adds a new `/fresh` comment command that can be used to instantly mark an issue or PR as not stale, rather than waiting up to a day for the stale workflow to run.

I have:

This needs to be merged to test it.

### DIFF
--- a/.github/workflows/commands.yml
+++ b/.github/workflows/commands.yml
@@ -90,3 +90,37 @@ jobs:
         github_token: ${{ secrets.GITHUB_TOKEN }}
         github_workspace: ${{ github.workspace }}
         version: v0.0.4
+
+  fresh:
+    runs-on: ubuntu-20.04
+    if: startsWith(github.event.comment.body, '/fresh')
+
+    steps:
+    - name: Extract Command
+      id: command
+      uses: xt0rted/slash-command-action@v1
+      with:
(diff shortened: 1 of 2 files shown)


.github/workflows/commands.yml
```diff
@@ -90,3 +90,37 @@ jobs:
...
+        github-token: ${{ secrets.GITHUB_TOKEN }}
+        script: |
+          // Delete the stale label.
+          try {
+            await github.issues.deleteLabel({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              name: ['stale']
+            })
```
negz, line 121: I'm pretty sure this works for PRs too, but I need to doublecheck.

hasheddan (approved):

.github/workflows/commands.yml
negz: I... think this is the case based on the Octokit docs? I'm inclined to merge and test.

negz: Confirmed in https://github.com/crossplane/crossplane/pull/3259

PR #3265, opened 2022-08-24
Add basic metadata annotations and refresh getting-started packages

### LINKED ISSUE #3247: Attach more metadata to getting started packages

The [`Configuration` packages](https://github.com/crossplane/crossplane/tree/master/docs/snippets/package) used in the getting started guide are extremely minimal, and do not contain any additional metadata about their purpose or usage.

The [`xpkg` specification](https://crossplane.io/docs/v1.9/reference/xpkg.html#object-annotations) includes mechanisms for attaching metadata to packages. We should add at least the basic annotations to the getting started packages.

jastang:
This PR adds metadata annotations and basic API documentation in the community "getting started" Configurations that help new users of Crossplane. 

It also refreshes a few dependencies to be in-line with the latest published versions of the Crossplane Providers in the Upbound registry.

Fixes #3247

I have:

For each Configuration, the package was built using `up` (v0.12.2) and pushed to a private registry. Then, each Configuration and its dependencies were installed separately into a cluster running Crossplane (1.8.1-up.2), and verified they were healthy upon install. Note each package had to be installed alone, as the `xpostgresqlinstances.database.example.org` objects all share the same name and subsequent installs would not be able to establish control from the previous ConfigurationRevision.

```
➜  azure k get pkg
NAME                                                          INSTALLED   HEALTHY   PACKAGE                                                      AGE
configuration.pkg.crossplane.io/jason-getting-started-azure   True        True      xpkg.upbound.io/upbound/jason-getting-started-azure:v0.0.4   18s

NAME                                                   INSTALLED   HEALTHY   PACKAGE                             AGE
provider.pkg.crossplane.io/crossplane-provider-aws     True        True      crossplane/provider-aws:v0.30.1     29h
provider.pkg.crossplane.io/crossplane-provider-azure   True        True      crossplane/provider-azure:v0.19.0   24h
provider.pkg.crossplane.io/crossplane-provider-gcp     True        True      crossplane/provider-gcp:v0.21.0     25h
```

Provisioning was also tested to ensure the database instance comes up healthy, for example:
```
➜  gcp k get postgresqlinstance
NAME    READY   CONNECTION-SECRET   AGE
my-db   True    db-conn             25m

Events:
  Type    Reason                      Age                  From                                                             Message
  ----    ------                      ----                 ----                                                             -------
  Normal  BindCompositeResource       3m19s (x6 over 22m)  offered/compositeresourcedefinition.apiextensions.crossplane.io  Composite resource is not yet ready
  Normal  ConfigureCompositeResource  78s (x8 over 22m)    offered/compositeresourcedefinition.apiextensions.crossplane.io  Successfully applied composite resource
  Normal  BindCompositeResource       78s (x2 over 78s)    offered/compositeresourcedefinition.apiextensions.crossplane.io  Successfully bound composite resource
  Normal  PropagateConnectionSecret   78s                  offered/compositeresourcedefinition.apiextensions.crossplane.io  Successfully propagated connection details from composite resource
```

### DIFF
--- a/docs/snippets/package/aws-with-vpc/crossplane.yaml
+++ b/docs/snippets/package/aws-with-vpc/crossplane.yaml
@@ -1,14 +1,29 @@
-apiVersion: meta.pkg.crossplane.io/v1alpha1
+apiVersion: meta.pkg.crossplane.io/v1
 kind: Configuration
 metadata:
   name: getting-started-with-aws-with-vpc
   annotations:
+    meta.crossplane.io/maintainer: Crossplane Maintainers <info@crossplane.io>
+    meta.crossplane.io/source: github.com/crossplane/crossplane
+    meta.crossplane.io/license: Apache-2.0
+    meta.crossplane.io/description: |
+      An introductory example to Crossplane and Composition for AWS.
+    meta.crossplane.io/readme: |
+      An introductory example to Crossplane and Composition for AWS.
+      This will enable provisioning of an RDS database instance.
+      The example also illustrates how to specify a non-default VPC.
+      
+      [Install & 
+      Configure](https://crossplane.io/docs/master/getting-started/install-configure.html)
+
+      [Provision a PostgreSQL RDS 
+      Database](https://crossplane.io/docs/master/getting-started/provision-infrastructure.html)
     guide: quickstart
     provider: aws
     vpc: new
 spec:
   crossplane:
-    version: ">=v1.4.0-0"
+    version: ">=v1.8.0-0"
   dependsOn:
     - provider: crossplane/provider-aws
-      version: ">=v0.18.2"
+      version: ">=v0.24.1"
--- a/docs/snippets/package/aws/crossplane.yaml
+++ b/docs/snippets/package/aws/crossplane.yaml
@@ -1,14 +1,29 @@
-apiVersion: meta.pkg.crossplane.io/v1alpha1
+apiVersion: meta.pkg.crossplane.io/v1
 kind: Configuration
 metadata:
   name: getting-started-with-aws
   annotations:
+    meta.crossplane.io/maintainer: Crossplane Maintainers <info@crossplane.io>
+    meta.crossplane.io/source: github.com/crossplane/crossplane
+    meta.crossplane.io/license: Apache-2.0
+    meta.crossplane.io/description: |
+      An introductory example to Crossplane and Composition for AWS.
+    meta.crossplane.io/readme: |
+      An introductory example to Crossplane and Composition for AWS.
+      This will enable provisioning of an RDS database instance.
+
+      [Install & 
+      Configure](https://crossplane.io/docs/master/getting-started/install-configure.html)
+
+      [Provision a PostgreSQL RDS 
+      Database](https://crossplane.io/docs/master/getting-started/provision-infrastructure.html)
+  labels:
     guide: quickstart
     provider: aws
     vpc: default
 spec:
   crossplane:
-    version: ">=v1.4.0-0"
+    version: ">=v1.8.0-0"
   dependsOn:
     - provider: crossplane/provider-aws
-      version: ">=v0.18.2"
+      version: ">=v0.24.1"
--- a/docs/snippets/package/azure/composition.yaml
+++ b/docs/snippets/package/azure/composition.yaml
@@ -30,6 +30,8 @@ spec:
             location: West US 2
             sslEnforcement: Disabled
             version: "11"
+            storageProfile:
+              storageMB: 5120            
             sku:
               tier: GeneralPurpose
               capacity: 2
--- a/docs/snippets/package/azure/crossplane.yaml
+++ b/docs/snippets/package/azure/crossplane.yaml
@@ -1,13 +1,27 @@
-apiVersion: meta.pkg.crossplane.io/v1alpha1
+apiVersion: meta.pkg.crossplane.io/v1
 kind: Configuration
 metadata:
   name: getting-started-with-azure
   annotations:
+    meta.crossplane.io/maintainer: Crossplane Maintainers <info@crossplane.io>
+    meta.crossplane.io/source: github.com/crossplane/crossplane
+    meta.crossplane.io/license: Apache-2.0
+    meta.crossplane.io/description: |
+      An introductory example to Crossplane and Composition for Azure.
+    meta.crossplane.io/readme: |
+      An introductory example to Crossplane and Composition for Azure.
+      This will enable provisioning of an Azure SQL database instance.
+
+      [Install & 
+      Configure](https://crossplane.io/docs/master/getting-started/install-configure.html)
+
+      [Provision a PostgreSQL Azure SQL 
+      Database](https://crossplane.io/docs/master/getting-started/provision-infrastructure.html)
     guide: quickstart
     provider: azure
 spec:
   crossplane:
-    version: ">=v1.4.0-0"
+    version: ">=v1.8.0-0"
   dependsOn:
     - provider: crossplane/provider-azure
-      version: ">=v0.13.0"
+      version: ">=v0.18.1"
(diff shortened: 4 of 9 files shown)


docs/snippets/package/azure/composition.yaml
```diff
@@ -30,6 +30,8 @@ spec:
             location: West US 2
             sslEnforcement: Disabled
             version: "11"
+            storageProfile:
+              storageMB: 5120
```
jastang, line 34: This was a small bug that I found. The user will pass in the storage capacity in GB, but we do a multiplication transform in the patch.

negz:
@jastang Just FYI, this PR won't be automatically assigned any reviewers until it's marked ready to review.

jastang:
> @jastang Just FYI, this PR won't be automatically assigned any reviewers until it's marked ready to review.

@negz yup, wanted to check a few more things before doing that 👍

docs/snippets/package/aws/crossplane.yaml
```diff
@@ -1,14 +1,29 @@
...
+
+      [Install & 
+      Configure](https://crossplane.io/docs/v1.9/getting-started/install-configure.html)
+
+      [Provision a PostgreSQL RDS 
+      Database](https://crossplane.io/docs/v1.9/getting-started/provision-infrastructure.html)
+  labels:
     guide: quickstart
     provider: aws
-    vpc: default
```
bobh66, line 8: nit - trailing spaces

docs/snippets/package/aws-with-vpc/crossplane.yaml
```diff
@@ -11,4 +26,4 @@ spec:
     version: ">=v1.4.0-0"
   dependsOn:
     - provider: crossplane/provider-aws
-      version: ">=v0.18.2"
```
bobh66, line 14: latest aws provider is 0.30.1 - is there any reason not to use that version?  Driving users (especially new users) to later versions may save them time and aggravation when bringing things up for the first time.

docs/snippets/package/aws-with-vpc/crossplane.yaml
```diff
@@ -11,4 +26,4 @@ spec:
     version: ">=v1.4.0-0"
```
bobh66, line 26: I think I would update this to the latest crossplane release too - no reason to encourage the use of older versions

docs/snippets/package/aws/crossplane.yaml
```diff
@@ -1,14 +1,29 @@
...
+    vpc: default      
 spec:
   crossplane:
     version: ">=v1.4.0-0"
```
bobh66, line 26: same comment as above

docs/snippets/package/azure/crossplane.yaml
```diff
@@ -1,13 +1,27 @@
...
+      [Install & 
+      Configure](https://crossplane.io/docs/v1.9/getting-started/install-configure.html)
+
+      [Provision a PostgreSQL Azure SQL 
+      Database](https://crossplane.io/docs/v1.9/getting-started/provision-infrastructure.html)    
     guide: quickstart
     provider: azure
 spec:
   crossplane:
     version: ">=v1.4.0-0"
```
bobh66, line 24: same comment as above

docs/snippets/package/gcp/crossplane.yaml
```diff
@@ -1,13 +1,27 @@
...
+      [Install & 
+      Configure](https://crossplane.io/docs/v1.9/getting-started/install-configure.html)
+
+      [Provision a PostgreSQL Cloud SQL 
+      Database](https://crossplane.io/docs/v1.9/getting-started/provision-infrastructure.html)        
     guide: quickstart
     provider: gcp
 spec:
   crossplane:
     version: ">=v1.4.0-0"
```
bobh66, line 24: same comment as above

bobh66 (changes requested):
A couple of minor suggestions but otherwise it LGTM.  Thanks!

docs/snippets/package/aws-with-vpc/crossplane.yaml
jastang: it's not pinned, so we will still install the latest available versions from `docker.io/crossplane/...`:

```
➜  crossplane git:(jastang-update-xpkg-metadata) k get providers
NAME                        INSTALLED   HEALTHY   PACKAGE                             AGE
crossplane-provider-aws     True        True      crossplane/provider-aws:v0.30.1     2d3h
crossplane-provider-azure   True        True      crossplane/provider-azure:v0.19.0   47h
crossplane-provider-gcp     True        True      crossplane/provider-gcp:v0.21.0     47h
```

there's some work to simultaneously publish to Docker hub and Upbound registries, for example [here](https://github.com/crossplane-contrib/provider-gcp/issues/453), and by default the `dependsOn.provider` looks at the Upbound registry so I'm using the latest available version there. once the backfill is done and everything is in sync we can bump these up :)

docs/snippets/package/aws-with-vpc/crossplane.yaml
jastang: @bobh66 Yeah, I thought this was outdated as well, but latest might be too restrictive -- I was testing on a hosted Upbound control plane for convenience, which was running 1.8.1 so I think some slack feels reasonable. @negz any guidance here on what this should be?

nullable-eth (approved):
LGTM, thanks @jastang

docs/snippets/package/aws-with-vpc/crossplane.yaml
hasheddan: General philosophy on error constraints in Crossplane packages: usually a `>=` with the earliest known compatible version (e.g. MSV - Minimum Supported Version) is a good policy (both for the package and Crossplane version constraints). Otherwise, you are potentially excluding someone with an existing install of dependencies (`provider-aws` in this case) or Crossplane from using your package when in reality it is compatible. If the user does not have the dependencies installed, Crossplane is going to resolve to whatever the latest version is at that point (as @jastang mentions).

hasheddan: To avoid just providing feedback without any suggested action: I think the bumps as currently proposed in this PR are still okay given that these getting started packages are tagged alongside their Crossplane versions, and `v0.24.1` has been out for a sufficiently long period of time to not break most folks, and these packages are primarily used for new installs anyway 👍🏻

docs/snippets/package/azure/crossplane.yaml
```diff
@@ -1,13 +1,27 @@
...
+    meta.crossplane.io/source: github.com/crossplane/crossplane
+    meta.crossplane.io/license: Apache-2.0
+    meta.crossplane.io/description: |
+      An introductory example to Crossplane and Composition for Azure.
+    meta.crossplane.io/readme: |
+      An introductory example to Crossplane and Composition for Azure.
+      This will enable provisioning of an Azure SQL database instance.
+
+      [Install & 
+      Configure](https://crossplane.io/docs/v1.9/getting-started/install-configure.html)
```
hasheddan, line 16: @jastang if we link to the docs, we'll need to make sure to update these every time we cut a release. We do a find and replace on a few different places during that process, but in order to identify that these need to be replaced, I think we should use https://crossplane.io/docs/master/getting-started/install-configure.html on `master` so that when we cut the `v1.10` branch (for example) our find and replace `s/master/v1.10/g` will catch this. What do you think?

jastang: @hasheddan good call -- makes sense to me. I'll update throughout 👍

docs/snippets/package/aws-with-vpc/crossplane.yaml
bobh66: Sounds good to me - thanks @hasheddan !

docs/snippets/package/aws-with-vpc/crossplane.yaml
bobh66: Thanks - LGTM

bobh66 (approved):

hasheddan (approved):
Thank you @jastang!

hasheddan:
@jastang I know we had previously opted to not backport this for the time being, but we are queuing up some patch releases for next week (#3298 #3299 #3300) and I think backporting _just_ the annotation updates could make a lot of sense. Would you be up for doing that?

PR #3287, opened 2022-09-02
fix(claim): Reuse spec.resourceRef.name if it exists

### LINKED ISSUE #3286: Claim: Composite Name is regenerated if `spec.resourceRef.name` is set
### What happened?

The reference to the composite resource in the claim is stored at `spec.resourceRef`. When the composite is deleted by the user, Crossplane will try to create a new composite - **always with a new name**. This, however, will fail because the resource reference still exists with the old name the controller will fail with the message
```
cannot bind claim that references a different composite resource
```

### How can we reproduce it?

1. Claim a composition
2. Delete the composite resource

Crossplane version: 1.9.0

MisterMX:

Only generate a new composite name if the claim's `spec.resourceRef.name` is not empty.

Fixes #3286

I have:

Unit tests

### DIFF
--- a/internal/controller/apiextensions/claim/configurator.go
+++ b/internal/controller/apiextensions/claim/configurator.go
@@ -66,7 +66,7 @@ func NewAPIDryRunCompositeConfigurator(c client.Client) *APIDryRunCompositeConfi
 // composite may or may not have been created in the API server when passed to
 // this method. The configured composite may be submitted to an API server via a
 // dry run create in order to name and validate it.
-func (c *APIDryRunCompositeConfigurator) Configure(ctx context.Context, cm resource.CompositeClaim, cp resource.Composite) error {
+func (c *APIDryRunCompositeConfigurator) Configure(ctx context.Context, cm resource.CompositeClaim, cp resource.Composite) error { //nolint:gocyclo
 	ucm, ok := cm.(*claim.Unstructured)
 	if !ok {
 		return nil
@@ -126,13 +126,19 @@ func (c *APIDryRunCompositeConfigurator) Configure(ctx context.Context, cm resou
 	ucp.SetClaimReference(proposed)
 
 	if !meta.WasCreated(cp) {
-		// The API server returns an available name derived from
-		// generateName when we perform a dry-run create. This name is
-		// likely (but not guaranteed) to be available when we create
-		// the composite resource. If the API server generates a name
-		// that is unavailable it will return a 500 ServerTimeout error.
-		cp.SetGenerateName(fmt.Sprintf("%s-", cm.GetName()))
-		return errors.Wrap(c.client.Create(ctx, cp, client.DryRunAll), errName)
+		// If the claim already specifies a resource ref, use the existing name.
+		// Otherwise let the API server generate a new one.
+		if ref := cm.GetResourceReference(); ref != nil && ref.Name != "" {
+			cp.SetName(ref.Name)
+		} else {
+			// The API server returns an available name derived from
+			// generateName when we perform a dry-run create. This name is
+			// likely (but not guaranteed) to be available when we create
+			// the composite resource. If the API server generates a name
+			// that is unavailable it will return a 500 ServerTimeout error.
+			cp.SetGenerateName(fmt.Sprintf("%s-", cm.GetName()))
+			return errors.Wrap(c.client.Create(ctx, cp, client.DryRunAll), errName)
+		}
 	}
 
 	return nil
--- a/internal/controller/apiextensions/claim/configurator_test.go
+++ b/internal/controller/apiextensions/claim/configurator_test.go
@@ -280,6 +280,65 @@ func TestCompositeConfigure(t *testing.T) {
 				},
 			},
 		},
+		"ConfiguredNewXRWithExistingName": {
+			reason: "A dynamically provisioned composite resource should be configured according to the claim and reuse an already specified composite name",
+			c: &test.MockClient{
+				MockCreate: test.NewMockCreateFn(nil),
+			},
+			args: args{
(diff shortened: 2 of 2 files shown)


MisterMX:
Intersting for #3120

internal/controller/apiextensions/claim/configurator.go
```diff
@@ -126,13 +126,19 @@ func (c *APIDryRunCompositeConfigurator) Configure(ctx context.Context, cm resou
...
-		// The API server returns an available name derived from
-		// generateName when we perform a dry-run create. This name is
-		// likely (but not guaranteed) to be available when we create
-		// the composite resource. If the API server generates a name
-		// that is unavailable it will return a 500 ServerTimeout error.
-		cp.SetGenerateName(fmt.Sprintf("%s-", cm.GetName()))
-		return errors.Wrap(c.client.Create(ctx, cp, client.DryRunAll), errName)
+		// If the claim already specifies a resource ref, use the existing name.
+		// Otherwise let the API server generate a new one.
+		if ref := cm.GetResourceReference(); ref != nil && ref.Name != "" {
```
bobh66, line 131: One side effect of this change is that it allows the user to specify the name of the composite, which I think has been requested or at least discussed.  This could be a good thing, but there may be other consequences?

bobh66 (approved):
The changes LGTM - I don't know if there are any unintended consequences of allowing the claim to specify the composite name, meaning that the user could create the claim with the composite resource name included.

internal/controller/apiextensions/claim/configurator.go
MisterMX: AFAIK it is not possible to influence another composite resource this way, because Crossplane will check if the `ownerReference` matches the claim's UID.

negz:
Would it make more sense to clear out the claim's `resourceRef` so that it can create a new XR name in this scenario? I like that this would make it extra clear that the XR was deleted and recreated - i.e. that you essentially have a new XR.

MisterMX:
> Would it make more sense to clear out the claim's resourceRef so that it can create a new XR name in this scenario?

Not really, because the newly generated name is always different than the old one. And since the XR name is often used as base for the MR name, the existing resources wouldn't be picked up.

A workaround to recover from a deleted XR is to create a "blank" XR with nothing in the spec that is filled by Crossplane on the next reconcile. But you have to know how Crossplane does the reconcile internally to come up with this.

negz:
> since the XR name is often used as base for the MR name, the existing resources wouldn't be picked up.

I think I'm missing something here. Could you maybe elaborate on the problem/use case touched on in https://github.com/crossplane/crossplane/issues/3286 a little more? It's seeming like there's a case where:

1. The XR is deleted.
2. The claim is not deleted.
3. The managed resources are not deleted (or their external resources are orphaned).

Then you want to recreate the MR and have it 'reattach' to the existing MRs?

MisterMX:
> I think I'm missing something here. Could you maybe elaborate on the problem/use case touched on in https://github.com/crossplane/crossplane/issues/3286 a little more? It's seeming like there's a case where [...]

No you are not missing something. That's exactly the situation I wanted to describe :)

> Then you want to recreate the MR and have it 'reattach' to the existing MRs?

If you remove an MR that is generated from a composition, it will be recreated on next reconcile. I want the same to happen with XRs. I want to recreate XR with the name specified in the claim's `spec.resourceRefs`.

internal/controller/apiextensions/claim/configurator.go
```diff
@@ -126,13 +126,19 @@ func (c *APIDryRunCompositeConfigurator) Configure(ctx context.Context, cm resou
...
+			cp.SetName(ref.Name)
+		} else {
```
negz, line 133: Nit: I have a weird(?) bias against using `else` - you could also just `return nil` after line 132 and keep this "non-exceptional" case a little less nested.

negz (commented):
I'm on the fence about this change.

CC @muvaf for a second opinion from another maintainer.

MisterMX:
> I'm on the fence about this change. @MisterMX how frequently do you see it happening?

I have encountered this once where the composite was deleted in an unknown way and wasn't reattached to the claim.

However, I have encountered the issue of regenerated composite names every time I was doing a migration (i.e. from one cluster to another). I cannot just copy the claim alone because this would create new external resources with different names - I have to copy the corresponding composite as well.

I would like to see this issue fixed. Maintaining claims and composite is a very clumsy job and I always have my fingers crossed that I didn't miss something. This is one of the reasons.

eloo:
just run into the same issue
we have found a claim with a deleted composite (unknown why this is deleted)
but at least the system is not healing itself then..

not stale

andrecastro:
@muvaf  and @nullable-eth What is holding this PR? Is there any issue that could be caused by adding this functionality?

My team is exploring the terraform Crossplane provider to get the benefit of having the state of the resources outside the cluster (S3 bucket). This would facilitate things like migration between clusters and other operations...exactly to overcome issues described here before.

The problem is that Crossplane always generate random names for the claims, so even using the terraform provider this becomes an issue, as the Workspace resource would be created with a new name, and the provider would try to create new resources...

So what is the problem to have this feature added?

negz:
I dropped a few thoughts at https://github.com/crossplane/crossplane/issues/3120#issuecomment-1532352009. Perhaps we should discuss this at the community meeting (or in a smaller forum) some time?

I'm fairly convinced that 'proper' use of the `crossplane.io/external-name` annotation would alleviate a lot of the issues folks are having here.

negz (changes requested):
Just being explicit that I'm the blocker here. I'm not yet sure this is the right way to solve the problem at hand.

PR #3292, opened 2022-09-07
Require Steering Committee approval of README.md

negz:

This is one of the more public facing "entrypoints" for the project - it seems like something we probably want steering committee approval to change.

I have:

N/A


jbw976 (approved):
This seems reasonable to me, as it is the public facing entry point to the project for a lot of folks.  I doubt this will end up being too burdensome in practice, but we can keep an eye out.

PR #3314, opened 2022-09-13
feat(transform): Add match transform

### LINKED ISSUE #3199: Composition Patch map default value or regexp match
Hello

If possible, I'd like to set a default value for a composite resource patch of type map.

We have a value map for different environments.
With multiple dev-environments, I'd like to define the value for all dev enviroments through a map default value or a map regexp match. 

Currently I have to specify each environment in my patch map
image

Especially for dev environments I'd like to use a regexp or a default value
```
    - fromFieldPath: metadata.labels['crossplane.io/claim-namespace']
      policy:
        fromFieldPath: Required
      toFieldPath: spec.forProvider.tags[6].value
      transforms:
      - map:
          .*-dev: DEV
```

MisterMX:

This adds a new transform `match` that is similar to `map` but supports more complex patterns such as regular expressions and allows returning a fallback value if no pattern matches.

For the start the transform supports `literal` and `regexp` patterns but this list can be expanded in the future. The default pattern type is `literal`.

```yaml
- type: match
  match:
    patterns:
      - type: literal # Not needed. This is the default.
        literal: us-west
        result: West US
      - type: regexp
        regexp: '^af-.*'
        result: Somewhere in Africa
    fallbackValue: Unknown
```

Closes #3199 

I have:

Manually and unit tests.


docs/reference/composition.md
```diff
@@ -510,6 +510,22 @@ You can use the following types of transform on a value being patched:
     au-east: Australia East
 ```
 
+`match`. A more complex version of `map` that can match different kinds of
+patterns.
```
negz, line 514: Some docs thoughts:

* Should we provide guidance on when to use map vs match? I imagine this could supersede/deprecate match if we wanted.
* I'd like us to document what happens when there are multiple matches.

apis/apiextensions/v1/composition_transforms.go
```diff
@@ -60,6 +68,7 @@ type TransformType string
 // Accepted TransformTypes.
 const (
 	TransformTypeMap     TransformType = "map"
+	TransformTypeMatch   TransformType = "match"
```
negz, line 71: Don't forget to update the CompositionRevision types and conversion code. 😞

negz (commented):
This mostly LGTM, though I'd love a reviewer to also take a look.

apis/apiextensions/v1/composition_transforms.go
MisterMX: Is #3315 a solution for this?

docs/reference/composition.md
MisterMX: Good catch. I added to the description that `match` should be used if more advanced pattern matching techniques are required. I also stated that the result of the first matching pattern is picked as output of the transform.

apis/apiextensions/v1/composition_transforms.go
MisterMX: I added the API changes to CompositionRevision as well so we don't depend on other issues for this.

@negz can you take another look?

internal/controller/apiextensions/composite/revision.go
```diff
@@ -138,14 +138,29 @@ func AsCompositionPatch(rp v1alpha1.Patch) v1.Patch {
...
+				Literal: p.Literal,
+				Regexp:  p.Regexp,
+				Result:  p.Result,
+			}
+		}
+		t.Match = &v1.MatchTransform{
+			Patterns:      patterns,
+			FallbackValue: rt.Match.FallbackValue,
+		}
+	}
```
negz, line 163: I think similar changes are needed in `internal/controller/composition/revision.go`. Otherwise this change LGTM.

MisterMX: Added.

negz (approved):

muvaf (approved):
Thanks! Sorry for delaying the merge!

MisterMX:
@muvaf can this be merged or is there something blocking it?

Issue #2656, opened 2021-10-22
Promote `CompositionRevision` to v1beta1


negz:
### What problem are you facing?

Support for `CompositionRevision` was added at v1alpha1 in v1.4.0. 
### How could Crossplane help solve your problem?

We need a plan to promote it to v1beta1, which may include accepting it as is, changing it, or dropping the feature. Community feedback would help here.

### Related issues

- [ ] https://github.com/crossplane/crossplane/issues/2762
- [ ] https://github.com/crossplane/crossplane/issues/3274
- [x] https://github.com/crossplane/crossplane/issues/2544
- [x] https://github.com/crossplane/crossplane/issues/3011
- [x] https://github.com/crossplane/crossplane/issues/3263
- [x] https://github.com/crossplane/crossplane/issues/3015
- [x] https://github.com/crossplane/crossplane/issues/2954

samidbb:
I find this feature very useful to control rollout of updates to compositions. But the current implementation will require platform builders to communicate about the changes in compositions which can be implementation specific to the platform. An example could be changing the way of referencing the providerconfig in XRs. I doubt that the consumers will be bothered about such changes in the implementation as long as they do affect the API schema in most cases.

The idea is that I would like to keep the ability to do the changes as a platform builder and test them on a selected number of existing resources. Then when I'm done I would like to have the existing consumers upgraded to the new composition revision that I have marked as **stable**. This way I can do changes behind the scene without requiring consumers to change their claim manifests in a controlled manner. 

My suggestion is as follows:

1) Introduce a new field in the compositionRevision CRD. This could be called something like "Stable" with boolean value:
```
> kubectl get compositionRevision
NAME                                           REVISION   CURRENT   Stable  AGE
myeresouce.class.crossplane.dfds.cloud-ywbwz   3          False     True    41h
myeresouce.class.crossplane.dfds.cloud-6hbzj   2          False     False   44h
myeresouce.class.crossplane.dfds.cloud-q5w48   1          True      False   44h
```
2) Add support for a new field in claim/composite resources
```
apiVersion:class.crossplane.dfds.cloud/v1alpha1
kind: MyClaim
...
spec:
  parameters:
  ...
  compositionUpdatePolicy: Manual # Valid options: [Maunal, Automatic]. This is still the same 
  compositionUpdateStrategy: Latest # Valid options: [Latest, Stable]. Only will be activated if user set compositionUpdatePolicy to  Automatic
```  
[...]

PR #3315, opened 2022-09-13
fix(CompositionRevisions): Import composition types from v1

### LINKED ISSUE #3274: Ensure full compatibility between Composition and CompositionRevision
### What happened?

Compositions and CompositionRevisions are made of different Go structs in `apis/apiextensions/v1` and `apis/apiextensions/v1alpha1`. This frequently leads to a desynchronization between both because changes in Compositions are not propagated to CompositionRevisions:

* #2825
* #2836
* #3113
* #3130

Consistency with Compositions for CompositionRevisions should be ensured before bumping them to v1beta1 (#2656).

This gets relevant every time the Composition API is extended.

There are multiple ways to ensure compatibility between Composition and CompositionRevisions:

* Automatically copy Go types from `v1` to `v1alpha1` using a script Go code generator.
* Let `v1alpha1` import Composition structs from `v1`.
* Create a third API struct to which `v1` and `v1alpha1` types are converted to during reconcile. K8s does something similar with it's own [apiextensions](`https://pkg.go.dev/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions).

MisterMX:

This ensures compatibility between Compositions and CompositionRevisions by importing all structures from v1 into v1alpha1 instead of copying them.

Fixes #3274

I have:

Unit tests


bobh66 (commented):
I'm not proficient enough in go or in the codebase to make any specific comments, but strictly from a code style/programming standpoint, if a composition revision is just a composition with a revision number, then this makes sense to me, as it is clear and concise and maintains consistency between the Composition and the CompositionRevision.  I understand the reluctance to mix v1 and v1alpha1, but hopefully CompositionRevision will reach v1 and it will cease to be a problem.   From that point on I would expect them to proceed in lockstep version-wise. 

My .02 - I can make this an approval if you think it qualifies.

negz:
I'm going to "sleep on" this for the weekend but I'm inline to just go ahead and merge. It seems like a violation to share API types, but it would make life a lot easier (and avoid a lot of bugs), and as you both point out this is a bit of a special case where one type is a mirror of another.

hasheddan (commented):
I noticed in https://github.com/crossplane/crossplane/issues/3274 that the idea of generating copies of the structures in `v1` was discussed. While this would have the same effect today, I do like that any changes to `v1` would be surfaced as actual diffs in `v1alpha1`. Was there is a particular reason why we didn't pursue that path?

MisterMX:
> I noticed in #3274 that the idea of generating copies of the structures in `v1` was discussed. While this would have the same effect today, I do like that any changes to `v1` would be surfaced as actual diffs in `v1alpha1`. Was there is a particular reason why we didn't pursue that path?

Sadly, I don't know of any kind of utility to clone code from another package.

It would also make Revision -> Composition conversion more complex because with different types you cannot use `DeepCopy` for it.

negz:
> Maybe it makes sense to overthink the design of CompositionRevision again

@MisterMX Personally I would prefer not to make CompositionRevision an opaque bag of JSON - I dropped a comment on that issue.

FWIW I believe @ezgidemirel is planning to tackle https://github.com/crossplane/crossplane/issues/2656 within the next week or two. This won't obviate this issue completely, but if CompositionRevision becomes beta and on-by-default we can at least simplify a bit in that we won't need to round-trip from revision -> composition when reconciling. We'll only need to be able to make revisions from a composition, and all reconcile logic (patches and transforms etc) will move to the revision type.

negz:
I'm going to close this since we're reluctant to proceed with the approach. Definitely copying at build time would be ideal, but as far as I know there's no tool to do this.

PR #3316, opened 2022-09-14
feat(transform): Allow converting string to float as resource.Quantity

### LINKED ISSUE #3284: Transforms: Support parsing strings as `resource.Quanitity` to `float64`

K8s does not recommend using floats in APIs [[1]](https://github.com/kubernetes-sigs/controller-tools/issues/245#issuecomment-518465214) and instead promotes the use of integers when dealing with numbers.

However, some apps require numbers with decimal points, such as the (haha) official [Kubernetes autoscaler](https://github.com/kubernetes/autoscaler/blob/8ba18537807fb15a074d3587528157d441ac4389/cluster-autoscaler/main.go#L106).

Crossplane prevents composition writers from being forced to propagate bad design decisions in their custom API through XRDs.

One way would be to define the field as integer and convert it using a float `math` transform (see #3059). This would work but the range of decimal points is limited (i.e. 3 if multiplied with 1000).

A better way would be to make use of K8s own [`resource.Quantity`](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/resource#Quantity) which is a special type that allows defining decimal values as integers with a unit prefix (i.e. `m`, `GiB` and many more).

Crossplane should provide a transform to convert a string into a `float64` by parsing the string as `resource.Quanitity`:

```yaml
- type: convert
  convert:
   toType: float64
   format: quantity
```

MisterMX:

Add the ability to convert a string to a float64 by parsing it as a `resource.Quantity`.

Configuration is done through a new properties `convert.format` that can be set to `quantity`. If this property is not set the default string to float conversion is executed.

```yaml
- type: convert
  convert:
   toType: float64
   format: quantity
```

Fixes #3284

I have:

Manually and unit tests.


apis/apiextensions/v1/composition_transforms.go
```diff
@@ -221,9 +221,28 @@ const (
...
+
+	// The expected input format.
+	//
+	// * `quantity` - parses the input as a K8s [`resource.Quantity`](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/resource#Quantity).
+	// Only used during `string -> float64` conversions.
+	//
+	// If this property is null, the default conversion is applied.
+	//
+	// +kubebuilder:validation:Enum=quantity
+	Format *ConvertTransformFormat `json:"format,omitempty"`
```
turkenh, line 247: Please also include a docs update.

internal/controller/apiextensions/composite/revision.go
```diff
@@ -0,0 +1,225 @@
+/*
```
turkenh, line 1: Should the changes in this file appear as a diff?

internal/controller/apiextensions/composite/transforms.go
```diff
@@ -354,6 +364,18 @@ func ResolveConvert(t v1.ConvertTransform, input any) (any, error) {
 	if to == v1.ConvertTransformTypeInt {
 		to = v1.ConvertTransformTypeInt64
 	}
+
+	var format v1.ConvertTransformFormat
+	if t.Format != nil {
+		format = *t.Format
+	}
+	switch format {
+	case "", v1.ConvertTransformFormatQuantity:
```
turkenh, line 373: nit: Should we define a default convertTransformFormat instead of an empty string for better readability? 
`v1.ConvertTransformFormatNone` or `v1.ConvertTransformFormatRaw` ?

turkenh (commented):
The diff is more than it actually is since I believe the PR needs a rebase after the changes in https://github.com/crossplane/crossplane/pull/3479.

internal/controller/apiextensions/composite/revision.go
MisterMX: That was by mistake. I removed these changes.

apis/apiextensions/v1/composition_transforms.go
MisterMX: Done

internal/controller/apiextensions/composite/transforms.go
MisterMX: Good idea. I added a `none` value. That the controller will fall back to.

internal/controller/apiextensions/composite/api.go
```diff
@@ -198,16 +198,6 @@ func (f *APIRevisionFetcher) Fetch(ctx context.Context, cr resource.Composite) (
 	return AsComposition(current), nil
 }
 
-// AsComposition creates a new composition from the supplied revision. It
-// exists only as a temporary translation layer to allow us to introduce the
-// alpha CompositionRevision type with minimal changes to the XR reconciler.
-// Once CompositionRevision leaves alpha this code should be removed and the XR
-// reconciler should operate on CompositionRevisions instead.
-func AsComposition(cr *v1beta1.CompositionRevision) *v1.Composition {
```
turkenh, line 206: Is this expected?

MisterMX: No, that was by mistake. I removed these changes.

MisterMX:
@turkenh I fixed the issues. Can you check again?

apis/apiextensions/v1/composition_transforms.go
turkenh: @MisterMX doc updates need to go to https://github.com/crossplane/docs now. 
Could you make sure it is reflected there?

turkenh (approved):
@MisterMX looking good to me, thanks for your contribution 👍 

Please rebase on top of the latest master so that we can merge it.

MisterMX:
> @MisterMX doc updates need to go to [crossplane/docs](https://github.com/crossplane/docs) now.
 Could you make sure it is reflected there?

Here you go: https://github.com/crossplane/docs/pull/292

Issue #3225, opened 2022-08-11
Proposal: Support foreground cascading deletion


bobh66:
### What problem are you facing?

When deleting a Crossplane Claim, the "--cascade=foreground" option to kubectl delete has no impact on the delete process.
All Composite Resources get deleted immediately and all Managed Resources are deleted as soon as possible by their associated providers.

This can cause issues with hung resources as documented in many other issues including:

#1612 
#1782 
https://github.com/crossplane/crossplane/issues/2439
https://github.com/crossplane-contrib/provider-helm/issues/115
https://github.com/crossplane-contrib/provider-kubernetes/issues/38
https://github.com/crossplane-contrib/provider-terraform/issues/47

and in:  https://github.com/crossplane/crossplane/discussions/3116

### How could Crossplane help solve your problem?
This proposal is the first of two that deal with dependency management, and deals only with supporting foreground cascading deletion.  More complex dependency handling will be addressed in a second proposal that is dependent on the changes described here.

Kubernetes uses Owner References and Finalizers to represent and implement dependencies between resources.  When a resource has one or more Finalizers and is marked "deleted", Garbage Collection will not reap the resource until the last finalizer has been removed, indicating that all dependencies have been cleaned up by their respective Controllers.  Cascading foreground deletion uses the ownerReferences and blockOwnerDeletion flag to generate a dependency graph and adds finalizers to ensure that the set of resources is deleted from the "bottom" up.

https://kubernetes.io/docs/concepts/architecture/garbage-collection/#foreground-deletion

Crossplane has two limitations that cause it to not support foreground cascading deletion:

- the blockOwnerDeletion flag is not set to "true" in the controller owner references that Crossplane adds to the resources it creates

- Crossplane ignores all "external" finalizers and deletes it's resources immediately when it sees the deletionTimestamp attribute added to an object's metadata

These issues can be resolved by the following changes:

- Include the blockOwnerDeletion: true attribute in the controller ownerReferences that Crossplane adds to the resources it creates.  This will cause kubernetes to add the foregroundDeletion finalizer to the owner resource when the delete command is executed with "--cascade=foreground" and block it's deletion while there are still resources that reference it.  Note that most/all Kubernetes controllers (Deployments, Replicasets, Statefulsets, etc) set this flag for their ownerReferences on the resources they control, such as Pods.

[...]

PR #3320, opened 2022-09-14
Postpone the deletion of XRs and XRCs until their direct children are gone

MisterMX:

Fixes https://github.com/crossplane/crossplane/issues/1612

Add a check before `RemoveFinalizer` for both XR and XRC to postpone their deletion while their children are still alive.

Essentially
* postpone the deletion of an XRC while the referenced XR is still alive
* postpone the deletion of an XR while at least one composed resource is still live

Relates to #3225 (not sure if it fixes it, though)

I have:

Manually and unit tests

### DIFF
--- a/internal/controller/apiextensions/claim/reconciler.go
+++ b/internal/controller/apiextensions/claim/reconciler.go
@@ -401,6 +401,13 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		log.Debug("Successfully deleted composite resource")
 		record.Event(cm, event.Normal(reasonDelete, "Successfully deleted composite resource"))
 
+		// If cp has a creation timestamp, it still does exist.
+		if meta.WasCreated(cp) {
+			log.Debug("Postponed deletion of claim because composite resource claim is still alive")
+			cm.SetConditions(xpv1.ReconcileSuccess())
+			return reconcile.Result{Requeue: false}, errors.Wrap(r.client.Status().Update(ctx, cm), errUpdateClaimStatus)
+		}
+
 		if err := r.claim.RemoveFinalizer(ctx, cm); err != nil {
 			log.Debug(errRemoveFinalizer, "error", err)
 			err = errors.Wrap(err, errRemoveFinalizer)
--- a/internal/controller/apiextensions/composite/api.go
+++ b/internal/controller/apiextensions/composite/api.go
@@ -24,8 +24,11 @@ import (
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
 	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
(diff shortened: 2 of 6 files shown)


bobh66 (commented):
I have a set of changes for 3225 in progress in both crossplane and crossplane-runtime  - I'd like to defer this change until those changes are in place as they are closely related to that work which is already being tested.

bobh66:
I'm also concerned that this is a significant change in behavior to introduce without any controls attached to it.  I agree that it's a desirable behavior, but it has the potential to cause disruption in the way people are used to having things work, and I think it would be better to have it under the control of something like the compositeDeletePolicy: Foreground option that is in https://github.com/crossplane/crossplane/pull/3321

MisterMX:
> I agree that it's a desirable behavior, but it has the potential to cause disruption in the way people are used to having things work

@bobh66 do you have some potential disruption in mind? 

From my perspective it makes absolute sense to keep the XR and XRC around to allow the user to track potential deletion errors. You cannot do that if the object is gone instantly.

> I'm also concerned that this is a significant change in behavior to introduce without any controls attached to it.

This actually does not change the behaviour of the deletion process. Unlike the `Orphan` deletion policy where external resource are kept alive, the child resources are deleted in the same way as before.

It only takes longer until the XR object is actually deleted.

bobh66:
> From my perspective it makes absolute sense to keep the XR and XRC around to allow the user to track potential deletion errors. You cannot do that if the object is gone instantly.

That's what foreground cascading deletion is for.  It delays the delete of the resources pointed at by ownerReferences until all dependencies have been deleted.

> This actually does not change the behaviour of the deletion process. Unlike the Orphan deletion policy where external resource are kept alive, the child resources are deleted in the same way as before.
> It only takes longer until the XR object is actually deleted.

Yes, it does.  The changes to the claim reconciler effectively force ALL kubectl delete commands of Claim objects to function as if the user has provided the --cascade=foreground command line option, instead of leaving the default "background" value.  There would be NO WAY to execute background deletion on a Claim object even if the user wants to.

I don't believe that choice should be forced upon the user by the claim reconciler - it should be controllable by the user either in the command line option (which doesn't work for Claims because they are namespaced and Composites are not) or the compositeDeletePolicy.

#3225 is not intended to change any existing behavior - it is intended to allow the user to specify that they WANT foreground cascading deletion and to allow Crossplane to functionally implement that deletion process using standard kubernetes processing.

I am willing to extend the changes for #3225 to include this functionality - I think it makes sense from a user experience point of view, but it would still need to be controlled by the compositeDeletePolicy=Foreground option so that it is not changing the default background propagation policy.

internal/controller/apiextensions/claim/reconciler.go
```diff
@@ -401,6 +401,13 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		log.Debug("Successfully deleted composite resource")
 		record.Event(cm, event.Normal(reasonDelete, "Successfully deleted composite resource"))
 
+		// If cp has a creation timestamp, it still does exist.
+		if meta.WasCreated(cp) {
+			log.Debug("Postponed deletion of claim because composite resource claim is still alive")
```
negz, line 406:
```suggestion
			log.Debug("Postponed deletion of claim because composite resource is still alive")
```

internal/controller/apiextensions/composite/api.go
```diff
@@ -428,3 +432,37 @@ func (c *APINamingConfigurator) Configure(ctx context.Context, cp resource.Compo
...
+func (c *APIDeletableChecker) IsDeletable(ctx context.Context, cr resource.Composite) (bool, error) {
+	for i, ref := range cr.GetResourceReferences() {
+		nn := types.NamespacedName{
+			Namespace: ref.Namespace,
+			Name:      ref.Name,
+		}
+		obj := unstructured.Unstructured{}
+		// Don't delete the composite if at least one composed resource is still
+		// alive.
+		err := c.client.Get(ctx, nn, &obj)
```
negz, line 459: Does this work? 🤔 How does the client know what GVK to get? I think you might need to set them on `obj`.

negz (changes requested):
After a little more reading and thinking, I tend to agree with @bobh66 here. This PR is "manually" doing the same thing that foreground cascading deletion should do for us.

PR #3321, opened 2022-09-14
Add support for deleteCompositePolicy

### LINKED ISSUE #3225: Proposal: Support foreground cascading deletion

bobh66:
Added support for the compositeDeletePolicy attribute and modified claim and composite reconcilers to delay delete processing while external finalizers exist.   This is the coordinated PR for https://github.com/crossplane/crossplane-runtime/pull/355

Also updated tests to include the blockOwnerDeletion flag that is now set.

This PR contains the private version of crossplane-runtime from the above PR and will need to be updated before it is merged.

Fixes #3225

I have:

Verified in local environment that foreground deletion works as expected.

### DIFF
--- a/design/design-doc-foreground-deletion.md
+++ b/design/design-doc-foreground-deletion.md
@@ -1,0 +1,262 @@
+# Foreground Cascading Deletion Support
+
+* Owner: Bob Haddleton (@bobh66)
+* Reviewers: Crossplane Maintainers
+* Status: Draft
+
+## Background
+
+When a Claim is deleted, Crossplane deletes the associated Composite resource
+using the default "Background" kubernetes propagation policy. Crossplane
+delegates the deletion of all the Composed resources from that Composite to
+kubernetes via the ownerReference on each resource.  When the "root" Composite
+resource is deleted by the claim reconciler, kubernetes garbage collection
+deletes all the Composed resources for the entire graph, which sets the
+deletionTimestamp attribute on each resource.  That triggers essentially
+simultaneous deletion of all Composite and Managed Resources that are related to
+the root Composite.
+
+Crossplane places finalizers on the objects that it creates to ensure that the
+objects are processed by the Crossplane resource controllers before they are
+deleted.  Crossplane directly manages the deletion of the Composite and Managed
+resources that it creates, calling the appropriate Delete() function for the
+resources when the deletionTimestamp attribute shows up on an object.
+
+While all the Composed objects (Composite and Managed) have Controller
+ownerReferences that point to the composite that created them, there is no
+ownerReference on the "root" Composite that references the Claim.  This is due
+to the fact that the Composite object is cluster-scoped and the Claim is
+namespace-scoped.  Kubernetes does not allow a cluster-scoped
+object to be "owned" by a namespaced object.  This renders the
+_--cascade=foreground_ option non-functional for Claim objects. Kubernetes does
+not find any objects that refer to the Claim in an ownerReference, so it doesn't
+set the _foregroundDeletion_ finalizer, and it marks the Claim as deleted. The
+Claim reconciler sees the deletionTimestamp indicator and calls Delete() on the
+associated Composite, using the default Propagation Policy of "background".
+
+If a standalone Composite object is deleted with --cascade=foreground, all the
+"leaf" node Managed Resources are immediately deleted as described above, and
+the ConnectionDetails are removed from all the Composite resources, and then the
+Composite resources are deleted from etcd from the "bottom up" by garbage
+collection using the _foregroundDeletion_ finalizer.
+
+As a result of the above:
+- The _--cascade=foreground_ option to kubectl is effectively non-functional on
+  all Crossplane resources.
+- Deletion of the claim or composite returns immediately, while actual managed
+  resource deletion may take a considerable amount of time.
+- Failed deletion of managed resources is not visible to the user. [1612]
+
+### Goals
+
+We would like to come up with a design that:
+
+- Will not break any existing functionality.
+- Will support the use of Foreground Cascading Deletion on Composite objects.
+- Will support a simulated Foreground Cascading Deletion on Claim objects.
+
+## Out of Scope
+
+This design does not consider any changes to the Claim/Composite reference
+design that would allow _kubectl delete --cascade=foreground_ to work from the
+command line on a Claim resource.
+
+This design is not intended to block resource deletion requests to the
+Kubernetes API.
+
+This design does not address use of the Orphan propagation policy when deleting
+Claims or standalone Composite objects.
+
+## Design
+
+The intent of this design is to allow Foreground Cascading Deletion to work on
+Crossplane resources.
+
+Foreground Cascading Deletion requires the _blockOwnerDeletion_ attribute to be
+set to _true_ on the Controller ownerReference that Crossplane adds to all
+Composed resources.
+
+Given the restrictions on Claim/Composite references, in order to simulate
+Foreground Cascading Deletion on claims, a new attribute is added to the Claim
+which enables the claim controller to call _Delete()_ on the composite resource
+with propagation policy _Foreground_.
+
+### API
+
+#### Claim: compositeDeletePolicy
+
+A _compositeDeletePolicy_ attribute is added to the Claim spec to indicate to
+the claim reconciler that it should use a specific Propagation Policy when
+deleting the associated Composite,   This attribute has possible values of
+Background (default) and Foreground.  The value "Orphan" is intentionally not
+supported as there is no requirement to leave the Composite resource in place
+(and under reconciliation) while deleting the Claim.
+
+The Claim reconciler uses the value of compositeDeletePolicy in the Delete API
+call for the associated Composite.  If Background is used the Composite deletion
+will occur as it does today and all associated resources will be marked for
+deletion immediately and then reconciled.
+
+If the value Foreground is used then the Composite and all of it's composed
+resources will be deleted using Foreground Cascading Deletion and the delete
+process with start at the leaf nodes and proceed back to the composite.
+
+**Examples:**
+
+Delete the Composite resource using Background propagation policy (existing
+scenario):
+
+```yaml
+spec:
+  compositeDeletePolicy: Background
+```
+
+Delete the Composite resource using Foreground propagation policy:
+
+```yaml
+spec:
+  compositeDeletePolicy: Foreground
+```
+
+### User Experience - Before
+
+#### Claim Delete - Background
+- User executes _kubectl delete -n namespace <claim name>_.
+- Kubernetes adds the deletionTimestamp attribute to the claim's metadata.
+- Crossplane claim reconciler detects the deletion and executes a Kubernetes API
+  Delete on the associated Composite with the default Background propagation
+  policy.
+- Kubernetes generates a graph of resources using ownerReferences and marks all
+  resources as deleted with deletionTimestamp.
+- Crossplane composite and managed reconcilers detect the deletion:
+  - The Composite reconcilers delete the associated ConnectionDetails and
+    remove the Crossplane finalizer.
+  - The Managed reconcilers delete the remote external resources and remove the
+    Crossplane finalizer.
+- Kubernetes garbage collection removes the Composite/Managed objects after the
+  Crossplane finalizers are removed.
+
+#### Claim Delete - Foreground
+- User executes _kubectl delete -n namespace --cascade=foreground <claim name>_.
+- Kubernetes adds the deletionTimestamp attribute to the claim's metadata.
+- There are no resources with ownerReferences that point at the claim so no
+  finalizers are added.
+- Crossplane claim reconciler detects the deletion and executes a Kubernetes API
+  Delete on the associated Composite with the default Background propagation
+  policy.
+- Kubernetes generates a graph of resources using ownerReferences and marks all
+  resources as deleted with deletionTimestamp.
+- Crossplane composite and managed reconcilers detect the deletion.
+  - The Composite reconcilers delete the associated ConnectionDetails and
+    remove the Crossplane finalizer.
+  - The Managed reconcilers delete the remote external resources and remove the
+    Crossplane finalizer.
+- Kubernetes garbage collection removes the Composite/Managed objects after the
+  Crossplane finalizers are removed
+
+#### Standalone Composite Delete - Background
+- User executes _kubectl delete <composite name>_.
+- Kubernetes adds the deletionTimestamp attribute to the composite's metadata.
+- Kubernetes generates a graph of resources using ownerReferences and marks all
+  resources as deleted with deletionTimestamp.
+- Crossplane composite and managed reconcilers detect the deletion.
+  - The Composite reconcilers delete the associated ConnectionDetails and
+    remove the Crossplane finalizer.
+  - The Managed reconcilers delete the remote external resources and remove the
+    Crossplane finalizer.
+- Kubernetes garbage collection removes the Composite/Managed objects after the
+  Crossplane finalizers are removed.
+
+#### Composite Delete - Foreground
+- User executes _kubectl delete --cascade=foreground <composite name>_.
+- Kubernetes adds the deletionTimestamp attribute to the composite's metadata.
+- Kubernetes generates a graph of resources using ownerReferences and marks all
+  resources as deleted with deletionTimestamp.
+- Crossplane composite and managed reconcilers detect the deletion.
+  - The Composite reconcilers delete the ConnectionDetails and remove the
+    Crossplane finalizer.
+  - The Managed reconcilers delete the remote external resources and remove the
+    Crossplane finalizer.
+- Kubernetes garbage collection removes the Composite/Managed objects after the
+  Crossplane finalizers are removed.
+
+### User Experience - After
+
+#### Claim Delete - Background (no change)
+
+#### Claim Delete - Foreground
+- A Claim exists with compositeDeletePolicy: Foreground.
+- User executes _kubectl delete -n namespace <claim name>_.
+- Kubernetes adds the deletionTimestamp attribute to the claim's metadata.
+- There are no resources with ownerReferences that point at the claim so no
+  finalizers are added at this stage.
+- Crossplane claim reconciler detects the deletion and executes a Kubernetes API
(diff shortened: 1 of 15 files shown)


bobh66:
It may be useful to enhance this change to block deletion of the Claim until the Composite is completely deleted.  The claim Reconciler can requeue until the Composite is gone and then remove the finalizer to allow the Claim to be deleted.

internal/controller/apiextensions/claim/reconciler.go
```diff
@@ -378,7 +382,11 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
 
-			if err := r.client.Delete(ctx, cp); resource.IgnoreNotFound(err) != nil {
+			do := &client.DeleteOptions{}
+			if *cm.GetCompositeDeletePolicy() == xpv1.CompositeDeleteForeground {
+				client.PropagationPolicy(metav1.DeletePropagationForeground).ApplyToDelete(do)
+			}
+			if err := r.client.Delete(ctx, cp, do); resource.IgnoreNotFound(err) != nil {
 				log.Debug(errDeleteComposite, "error", err)
 				err = errors.Wrap(err, errDeleteComposite)
 				record.Event(cm, event.Warning(reasonDelete, err))
```
bobh66, line 392: If we want to preserve the Claim as long as the Composite is not yet deleted, thus simulating Foreground deletion on the Claim itself, we could return after calling Delete() and wait to remove the finalizer until after the Composite is gone.   That check could happen after line 364 - if meta.WasDeleted(cp) is true then Delete() was called but the composite has not finished deleting yet so the rest of the reconciliation should be skipped.

internal/controller/apiextensions/claim/reconciler.go
```diff
@@ -359,6 +359,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 
 	if meta.WasDeleted(cm) {
 		log = log.WithValues("deletion-timestamp", cm.GetDeletionTimestamp())
+		if len(cm.GetFinalizers()) > 1 {
```
MisterMX, line 362: I don't think this is how finalizers are to be used here

From the [k8s docs](https://kubernetes.io/docs/concepts/overview/working-with-objects/finalizers/):

> Finalizers are namespaced keys that tell Kubernetes to wait until specific conditions are met before it fully deletes resources marked for deletion. Finalizers alert [controllers](https://kubernetes.io/docs/concepts/architecture/controller/) to clean up resources the deleted object owned.

To me this sounds like the opposite of what we are trying to do here: Do nothing while there is more than one finalizer.

I believe this is an overly complicated solution for what we actually want: Postpone the deletion of XR and XRC while their children are still around (see #3320).

internal/controller/apiextensions/claim/reconciler.go
```diff
@@ -378,7 +382,11 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 				return reconcile.Result{Requeue: false}, errors.Wrap(r.client.Status().Update(ctx, cm), errUpdateClaimStatus)
 			}
 
-			if err := r.client.Delete(ctx, cp); resource.IgnoreNotFound(err) != nil {
+			do := &client.DeleteOptions{}
+			if *cm.GetCompositeDeletePolicy() == xpv1.CompositeDeleteForeground {
+				client.PropagationPolicy(metav1.DeletePropagationForeground).ApplyToDelete(do)
```
MisterMX, line 387: Does this mean the controller is stuck until delete is done? It is very likely that we will run in a timeout with this. It is better to stick with background deletion and check the state in the next reconcile.

internal/controller/apiextensions/claim/reconciler.go
bobh66: This is not intended to be a solution for #3320 - it is intended to be a solution for #3225 and provide a foundation for #3321.
It can easily be extended to support #3320 but that was not it's primary purpose.

This is exactly what finalizers are intended to do - delay the deletion/garbage collection of a resource while a dependency exists on that resource.  Crossplane is the controller for these resources and so it needs to delay it's delete processing while other resources have dependencies on them.

negz: I do think this is a bit of a novel use of finalizers, in that I'm not sure it's common for controller code to consider _all_ finalizers as opposed to only considering finalizers it added.

That's not to say we shouldn't proceed with this direction though. Two thoughts:

* Given it's somewhat novel I think an explanatory comment would make sense.
* Would supporting a single special Crossplane finalizer make sense here? i.e. We would not proceed with deletion if a finalizer like `crossplane.io/has-dependents` still exists, but would otherwise ignore other finalizers?

bobh66: > Typically I see finalizers used as a way for a controller to signal that it needs to do something before the API server should complete the deletion.

Exactly.  Now this idea is extended to the context of Crossplane, in it's role as an API server "on top of" the kubernetes API server.  Where the Kubernetes API server is typically managing native objects that have no existence outside of etcd, Crossplane is managing groups of objects that by definition are composed of other Crossplane objects AND objects that have manifestations outside of Kubernetes itself.  Therefore just like Kubernetes garbage collection uses finalizers to keep objects in etcd until "everyone is done with them", Crossplane can use the same concepts to keep it's collections of local/remote objects in existence until there are no more declared dependencies on them.

> Given it's somewhat novel I think an explanatory comment would make sense.

Definitely - I've also thought about maybe a design document just to save the background on the design for posterity in case someone ever needs to know why it is this way.  Not to mention people wanting to understand how it works and how to use it.

> Would supporting a single special Crossplane finalizer make sense here?

I actually did a lot of thinking about this for #3323 and the problem with a single finalizer is you don't know where it came from or when it's ok to delete it.

If we assume the case of an external entitiy declaring a dependency on a Claim or Composite X.  X now has 2 finalizers, it's own plus one for the remote dependency.  (provider-kubernetes does this today with it's dependsOn feature)  Unfortunately maintaining the finalizer only on the Claim/Composite X is not sufficient for the dependency, since even Foreground Cascading Deletion doesn't put a foregroundDeletion finalizer on the "leaf" or "edge" nodes which have no ownerReferences pointing at them.  Any deletion of X will still cause all of the composed/managed resources that make up X to be deleted by the Composite and Managed Reconcilers because there are no additional finalizers on those "child" objects to prevent their deletion.  X will remain with it's extra finalizer but it will have no child resources associate with it, so it's essentially useless.

That's where #3323 comes in.  It is necessary to reconcile the extra finalizers all the way down to the leaf/edge nodes so that deletion of the entire Claim/Composite structure is blocked, not just the Claim/Composite itself.  However, it is not necessary for the specific dependencies on X to be reconciled to the edges.  Only the fact that X has more than one finalizer is important.  So when the claim/composite reconcilers see more than one finalizer on X, they can put a unique finalizer on all of the child composite/composed resources that it is responsible for.  That increases the finalizer count on those objects and those will add their own finalizers to their children, all of the way out to the leaf/edge nodes which provides the desired block against resource deletion.  It also provides traceability of the
finalizers since we can put the object UID in the finalizer name to indicate where the lock originates from (it had better be the same as the ownerReference object, but still).  This can also handle multiple overlapping dependencies on the same resources without concern for finalizer name conflicts.

When the external finalizer is removed from X, the reconciliation sees only 1 finalizer remaining and removes the unique finalizer from it's children, which flows out to the edges.  They now have no finalizers and get deleted as expected.

I originally thought the implementation of #3323 was going to be dependent on #3321 and fairly complex so I kept them separate, which also simplifies the individual problem descriptions.  However once I did the full investigation it turns out that the changes required for #3323 are fairly trivial and decoupled from #3321.  They won't function correctly without #3321, but they are in different places.  So they can be combined into one PR if that's preferred.

Sorry for the novel.  I'll definitely put together a design document.

negz: > I've also thought about maybe a design document just to save the background on the design for posterity in case someone ever needs to know why it is this way.

+1 - I think a design doc to tie together the handful of proposals would be very useful.

> I actually did a lot of thinking about this for https://github.com/crossplane/crossplane/issues/3323 and the problem with a single finalizer is you don't know where it came from or when it's ok to delete it.

Yeah, this occurred to me in retrospect as well.

design/design-doc-finalizer-support.md
```diff
@@ -0,0 +1,332 @@
...
+spec:
+  compositeDeletePolicy: Background
+```
+
+Delete the Composite resource using Foreground propagation policy:
+
+```yaml
+spec:
+  compositeDeletePolicy: Foreground
+```
```
negz, line 166: I was thinking today - is it instead possible to infer that a claim was deleted with `propagationPolicy: Foreground` by looking for the presence of the `foregroundDeletion` finalizer on the claim? If it were this would seem like a more "transparent" implementation of cascading deletion than adding a new field (which may conflict with the policy supplied when deleting the claim).

I have a vague recollection that you looked into this already and the answer is "no", but I can't find where we discussed it.

bobh66: That was my initial plan but when I tested the changes, I placed a watch on the claim and found that the foregroundDeletion finalizer is placed on the claim and then immediately removed, presumably because there are no ownerReferences pointing at the claim, and by the time the reconciler retrieves the object the foregroundDeletion finalizer is gone.  So the reconciler has no way to know that the deletion was caused by foreground deletion.   I don't know of any other way to detect foreground cascading deletion has been requested.  It's too bad there isn't an annotation added with that information.

negz: Got it.

I'd love it if we could somehow detect foreground deletion of a claim without needing this separate policy, but I don't see any great options.

design/design-doc-finalizer-support.md
```diff
@@ -0,0 +1,332 @@
...
+
+This design does not address use of the Orphan propagation policy when deleting Claims or
+standalone Composite objects.
+
+## Design
+
+The intent of this design is to allow Crossplane objects to "behave" like native Kubernetes objects
+during delete processing.  A native Kubernetes object will remain in etcd, and all of it's
+content remains available to the user, while the object is marked deleted and waiting for
+finalizers to be removed by any controllers that are watching the resource.
```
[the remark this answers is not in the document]
bobh66: I think the question here is whether the resource(s) that are associated with the Composite/Managed resource should remain in existence until the `foregroundDeletion` finalizer is removed.   For Composites that just means the `ConnectionDetails` which has little impact. For Managed resources,  if there is an `ownerReference` with `blockOwnerDeletion: true` pointing at it, that Managed Resource is no longer a leaf node in the graph and will have a `foregroundDeletion` finalizer added to it when foreground cascading deletion is requested.  In the current design that finalizer is ignored and the external Delete() function is called immediately when the `deletionTimestamp` is added to the object, causing the external resource to be deleted before the object that is pointing at it can be deleted.  If that external object was relying on the external resource for something it will no longer be available.

design/design-doc-finalizer-support.md
```diff
@@ -0,0 +1,332 @@
...
+object prevents garbage collection from removing an object from etcd,
+leaving the object "intact" while it is marked as deleted, the existence of an "additional"
+finalizer on a Crossplane resource should block delete routines that would normally
+execute for a resource that has been marked deleted.
+
+Given that a Claim, Composite or Managed Resource in Crossplane is either a collection of other
+Crossplane resources or a reference to an external resource, any dependency represented by the
+existence of a finalizer is actually a dependency on ALL resources owned or represented by that
+resource.  This includes "local" Composite or Managed resources as well as "remote" resources that
+a Managed resource represents.
```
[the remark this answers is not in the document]
bobh66: I'll rework the document based on our discussions.

design/design-doc-finalizer-support.md
```diff
@@ -0,0 +1,332 @@
...
+The existence of a non-Crossplane finalizer on a Crossplane object does not prevent
+Crossplane from executing the associated Delete logic on the resource.  The most noticeable impact
+of this is that Foreground Cascading Deletion does not work on Crossplane Composite resources.
+This is caused by two factors, one is that the Controller ownerReferences that Crossplane creates
+do not set the _blockOwnerDeletion_ attribute to true.  The other reason is that the
+existence of the _foregroundDeletion_ finalizer on the Controller objects (composites) does not
+block the deletion of the resources associated with those composites or their composed resources,
+so the deletion process does not "cascade" from the edge/leaf nodes to the root node.  The
+composite and managed reconcilers detect object deletion and execute their delete routines,
+ignoring the external finalizers.
```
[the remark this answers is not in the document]
negz: Thanks for confirming @bobh66! That sounds like what I would expect.

internal/controller/apiextensions/claim/reconciler.go
bobh66: Design doc has been added, limited to foreground cascading deletion functionality.

negz (commented):
This looks good to me - will approve once we get the runtime changes merged and this PR updated to match.

I'd still like to consider adding some way to override the delete policy (e.g. set or force a default in the XRD like we do for composition refs) but that can happen in a future PR.

design/design-doc-foreground-deletion.md
```diff
@@ -0,0 +1,205 @@
+# Foreground Cascading Deletion Support
```
negz, line 1: Nit: We prefer to wrap docs at 80 chars - I'd appreciate it if you could do so here. There's a few plugins (e.g. "rewrap" for VS Code) that can do this automatically.

internal/controller/apiextensions/claim/reconciler.go
```diff
@@ -374,6 +374,12 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 
 		cm.SetConditions(xpv1.Deleting())
 		if meta.WasCreated(cp) {
+			if meta.WasDeleted(cp) {
+				if *cm.GetCompositeDeletePolicy() == xpv1.CompositeDeleteForeground {
```
negz, line 378: Would it make sense to check for the `foregroundDeletion` finalizer having been set on the composite here instead? I imagine what we're trying to detect is that the composite is currently undergoing foreground deletion, and the presence of that finalizer seems like a more 'direct' signal.

This is late in the game so I'm happy to ignore the thought (or follow up in a future PR).

negz (approved):
Thanks for persevering with this @bobh66!

design/design-doc-foreground-deletion.md
bobh66: Fixed - sorry about that.

internal/controller/apiextensions/claim/reconciler.go
bobh66: That would probably work but I would want to test it thoroughly.  I'm happy to put a PR in for 1.11 and test it before it releases.

negz (approved):

PR #3324, opened 2022-09-15
Add LICENSE file to crossplane Helm chart

### LINKED ISSUE #3283: Helm chart missing license
Hi I am working with an enterprise which requires helm charts to be onbarded to in internal repository. The onboarding process requires the chart to have a license declared as defined here:

https://helm.sh/docs/topics/charts/

The onboarding is complaining that a license is not included in the crossplane chart 1.10.0-rc.0.112.g83356388 I am onboarding

Looking at github, I do not see a license file in the chart:

https://github.com/crossplane/crossplane/tree/5bfba85669acbf8758d0698ce325533be44f4520/cluster/charts/crossplane

Could you consider including a license file in the chart to enable using crossplane in organizations that require license declaration.

Thanks.

jeanduplessis:
Signed-off-by: Jean du Plessis <jeandp@gmail.com>

Adds the same LICENSE file from the root of the repo to the crossplane Helm chart. 
This conforms to the Helm chart guidance for adding a license with a chart: https://helm.sh/docs/topics/charts/#chart-license-readme-and-notes

Fixes #3283

I have:

Not applicable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -95,7 +95,10 @@ crds.clean:
 	@find $(CRD_DIR) -name '*.yaml.sed' -delete || $(FAIL)
 	@$(OK) cleaned generated CRDs
 
-generate.run: gen-kustomize-crds gen-install-doc
+generate.run: gen-kustomize-crds gen-install-doc gen-chart-license
+
+gen-chart-license:
+    @cp LICENSE cluster/charts/crossplane/LICENSE
 
 generate.done: crds.clean
 
--- a/Makefile
+++ b/Makefile
@@ -98,7 +98,7 @@ crds.clean:
 generate.run: gen-kustomize-crds gen-install-doc gen-chart-license
 
 gen-chart-license:
-    @cp LICENSE cluster/charts/crossplane/LICENSE
+	@cp -f LICENSE cluster/charts/crossplane/LICENSE
 
 generate.done: crds.clean
 
--- a/cluster/charts/crossplane/LICENSE
+++ b/cluster/charts/crossplane/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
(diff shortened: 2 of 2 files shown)


hasheddan (commented):
Thanks @jeanduplessis! What would you think about adding a `cp` command attached to `generate.run` that copied the LICENSE from repo root into the chart directory to make sure that we keep these in-sync?

Something like:
```
generate.run: gen-kustomize-crds gen-install-doc gen-chart-license

gen-chart-license:
    @cp LICENSE cluster/charts/crossplane/LICENSE
```

This would run in CI and CI would fail if there was an uncommitted diff

jeanduplessis:
@hasheddan that sounds like a good way to ensure the licenses stay in sync, I'll update the PR.

bassam (approved):

hasheddan (approved):

jeanduplessis:
@hasheddan Not sure why the CI jobs are failing here... 
Seems to fail at running the Go build cache action: https://github.com/crossplane/crossplane/blob/master/.github/workflows/ci.yml#L59

hasheddan:
@jeanduplessis try rebasing on `master` then force pushing

jeanduplessis:
@hasheddan thanks, would you mind approving the workflow again please.

hasheddan (commented):
@jeanduplessis did you intentionally drop the Makefile machinery here?

jeanduplessis:
@hasheddan Urgh no... it was a mistake 🤦🏻‍♂️ Fixing now

jeanduplessis:
@hasheddan please approve the workflow again.

Makefile
```diff
@@ -95,7 +95,10 @@ crds.clean:
 	@find $(CRD_DIR) -name '*.yaml.sed' -delete || $(FAIL)
 	@$(OK) cleaned generated CRDs
 
-generate.run: gen-kustomize-crds gen-install-doc
+generate.run: gen-kustomize-crds gen-install-doc gen-chart-license
+
+gen-chart-license:
+    @cp LICENSE cluster/charts/crossplane/LICENSE
```
turkenh, line 101: Not seem to be relevant to the CI failure but, should add `-r` here? https://www.tecmint.com/force-cp-to-overwrite-without-confirmation/

turkenh:
> @hasheddan Not sure why the CI jobs are failing here... Seems to fail at running the Go build cache action: https://github.com/crossplane/crossplane/blob/master/.github/workflows/ci.yml#L59

@jeanduplessis checked out this PR and run `make go.cachedir` which is the command run the step before failure [here](https://github.com/crossplane/crossplane/blob/d815faa2cfbd62570d8d7aa4c80f2fb2b6345cca/.github/workflows/ci.yml#L57).

It failed with `Makefile:101: *** missing separator.  Stop.`

It looks like it is a [spaces instead of a tab](https://retrocomputing.stackexchange.com/questions/20292/why-does-make-only-accept-tab-indentation) problem.

Makefile
jeanduplessis: We shouldn't have a problem just using `cp`. (I've tested this locally).
That said I don't think there's any harm in using `-f` flag to force overwrite. 

`-r` would be needed if we were copying directories.

turkenh: Actually, that would only be a precaution to the following statement in the linked blog above: 

> By default, modern Linux distributions especially those in the Red Hat Enterprise Linux (RHEL) family come with an alias for the cp command which makes a user run the cp command in interactive mode. This may not be the case on Debian and Ubuntu derivatives.

turkenh: Ah, sorry, I meant `-f`, yes 🤦

turkenh (approved):

jeanduplessis:
@turkenh Thanks so much for picking up that issue. TIL about Makefiles and tabs vs space.

Issue #3294, opened 2022-09-07
Honor `imagePullSecrets` on the Crossplane `ServiceAccount` and propagate to dependencies and Provider `Deployment`s


hasheddan:
### What problem are you facing?

Crossplane has a dedicated `ServiceAccount` that [gets created on installation](https://github.com/crossplane/crossplane/blob/b01b17353198a8de28664cf1eec601aaaf2fd95a/cluster/charts/crossplane/templates/serviceaccount.yaml#L2). The Helm chart allows for attaching `imagePullSecrets`, but currently the package manager does [falls back](https://github.com/google/go-containerregistry/blob/f981b4c07e536a789fae2fbe5ce4b4269ee767fb/pkg/authn/kubernetes/keychain.go#L55) to only consider `imagePullSecrets` from the `default` SA in the namespace. This means that, unintuitively, secrets attached to the `default` SA will be considered when pulling packages, but those from the Crossplane SA will not.

Furthermore, secrets from any SA are not attached to Provider `Deployment` SA's, meaning that a Provider will be able to be pulled, but its `Deployment` will go into `CrashLoopBackoff` on trying to pull without credential.s  

### How could Crossplane help solve your problem?

We should honor the Crossplane SA `imagePullSecrets` _and_ propagate them to all Provider `Deployment`s. Attaching an `imagePullSecret` to SA will make it usable for installing any package. If more granular secret scoping is desired, `packagePullSecrets` can still be attached to individual `Provider` or `Configuration` installs.

PR #3332, opened 2022-09-20
Fetch package pull secrets from Crossplane SA and propagate to Provider Deployments

hasheddan:

This changeset updates to consume package pull secrets from the Crossplane `ServiceAccount`, rather than the previous behavior of falling back to the `default` `ServiceAccount`. It also propagates secrets from the `ServiceAccount` to the `ServiceAccount` of any Provider `Deployment`. This enables installing chains of private dependencies by attaching all required package pull secrets to the Crossplane `ServiceAccount`.

Lastly, any package pull secrets on a `ProviderRevision` are propagated to their `Deployment` as well, but they are not propagated to dependencies. We may opt to do this in the future, but there is some subtleties around shared dependents that make it more complex to do so. Because propagating the package pull secrets from the `ServiceAccount` makes it possible to install chains of private dependencies, we defer propagating revision secrets to dependencies until a future date.

Fixes https://github.com/crossplane/crossplane/issues/3294

I have:

`make build` && `make e2e.run skipcleanup=true`

Then created pull secret:
```
up ctp pull-secret create package-pull-secret -n crossplane-system -f creds.json
```

Patched Crossplane SA:
```
kubectl patch serviceaccount crossplane -n crossplane-system -p '{"imagePullSecrets": [{"name": "package-pull-secret"}]}'
```

Installed provider without pull secrets:
```
kxp install provider xpkg.upbound.io/upbound/provider-aws:v0.13.0
```

Verified installed and healthy:
```
🤖 (crossplane) k describe provider.pkg.crossplane.io/upbound-provider-aws
Name:         upbound-provider-aws
Namespace:    
Labels:       <none>
Annotations:  <none>
API Version:  pkg.crossplane.io/v1
Kind:         Provider
Spec:
  Ignore Crossplane Constraints:  false
  Package:                        xpkg.upbound.io/upbound/provider-aws:v0.13.0
  Package Pull Policy:            IfNotPresent
  Revision Activation Policy:     Automatic
  Revision History Limit:         0
  Skip Dependency Resolution:     false
Status:
  Conditions:
    Last Transition Time:  2022-09-21T13:53:22Z
    Reason:                HealthyPackageRevision
    Status:                True
    Type:                  Healthy
    Last Transition Time:  2022-09-21T13:52:17Z
    Reason:                ActivePackageRevision
    Status:                True
    Type:                  Installed
  Current Identifier:      xpkg.upbound.io/upbound/provider-aws:v0.13.0
  Current Revision:        upbound-provider-aws-fc98ece40acb
Events:
  Type     Reason                  Age                From                                 Message
  ----     ------                  ----               ----                                 -------
  Warning  InstallPackageRevision  48s (x3 over 89s)  packages/provider.pkg.crossplane.io  current package revision health is unknown
  Normal   InstallPackageRevision  24s (x2 over 24s)  packages/provider.pkg.crossplane.io  Successfully installed package revision
```

### DIFF
--- a/cluster/charts/crossplane/templates/deployment.yaml
+++ b/cluster/charts/crossplane/templates/deployment.yaml
@@ -65,6 +65,10 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: metadata.namespace
+          - name: POD_SERVICE_ACCOUNT
+            valueFrom:
+              fieldRef:
+                fieldPath: spec.serviceAccountName
           {{- if .Values.webhooks.enabled }}
           - name: "WEBHOOK_TLS_SECRET_NAME"
             value: webhook-tls-secret
--- a/internal/xpkg/fetch.go
+++ b/internal/xpkg/fetch.go
@@ -54,9 +54,10 @@ type Fetcher interface {
 
 // K8sFetcher uses kubernetes credentials to fetch package images.
 type K8sFetcher struct {
-	client    kubernetes.Interface
-	namespace string
-	transport http.RoundTripper
+	client         kubernetes.Interface
+	namespace      string
+	serviceAccount string
+	transport      http.RoundTripper
 }
 
 // FetcherOpt can be used to add optional parameters to NewK8sFetcher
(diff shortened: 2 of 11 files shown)


cluster/charts/crossplane/templates/deployment.yaml
```diff
@@ -65,6 +65,10 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: metadata.namespace
+          - name: POD_SERVICE_ACCOUNT
+            valueFrom:
+              fieldRef:
+                fieldPath: spec.serviceAccountName
```
hasheddan, line 71: Note: propagating the Crossplane SA name to the xpkg fetcher could be viewed as a breaking change (if someone was attaching package pull secrets to the `default` SA then it would now break). However, given that this would have to take place outside of the Crossplane install process (we only allow passing package pull secrets to Crossplane SA), it is quite unituitive behavior, and is not documented as supported, I feel comfortable including this in the upcoming minor release.

muvaf: Agreed that this was an unintended behavior anyway - using secret names from an SA that's not referenced during installation of Crossplane. Worth calling out in the release notes though.

internal/xpkg/fetch.go
```diff
@@ -98,11 +99,28 @@ func WithCustomCA(rootCAs *x509.CertPool) FetcherOpt {
...
+func WithServiceAccount(sa string) FetcherOpt {
+	return func(k *K8sFetcher) error {
+		k.serviceAccount = sa
+		return nil
+	}
+}
+
 // NewK8sFetcher creates a new K8sFetcher.
-func NewK8sFetcher(client kubernetes.Interface, namespace string, opts ...FetcherOpt) (*K8sFetcher, error) {
+func NewK8sFetcher(client kubernetes.Interface, opts ...FetcherOpt) (*K8sFetcher, error) {
```
muvaf, line 121: Why do we remove the namespace from arguments? Can it function without a namespace?

muvaf (approved):
LGTM after addressing my single comment. Thanks @hasheddan !

internal/xpkg/fetch.go
hasheddan: We are just moving to passing Namespace via option to simplify signature here. It is a similar situation as SA and will just fall back to `default` if not provided. (ref: https://github.com/google/go-containerregistry/blob/9a5c14ada6ffae772793231c186a50d911668e72/pkg/authn/kubernetes/keychain.go#L37)

PR #3333, opened 2022-09-20
Reconcile finalizers

### LINKED ISSUE #3323: Proposal:  Reconcile Finalizers
One way in which Kubernetes represents dependencies between resources is via finalizers.  A resource with a dependency on another resource can place a finalizer on it's dependency so that the dependency cannot be deleted until the dependent resource removes the finalizer.

Crossplane Composites are collections of other resources, so any dependency on a Claim or Composite is actually a dependency on all of the components of the Claim/Composite, down to the "edge" or "leaf" nodes or Managed resources.

That means that when a resource places a finalizer on a Claim or Composite resource, it is declaring a dependency on the entire tree of resources represented by the Claim/Composite, not just the Claim/Composite itself.

When kubectl delete is called on the top-level claim/composite, kubernetes uses the ownerReference relationships to traverse the tree and mark all of the references deleted at the same time.  Since the finalizer only exists on the Claim or Composite dependency, it does not prevent all of the "leaf" node Managed Resources from being deleted, even though the Composite itself cannot be deleted due to the finalizer.  This behavior is independent of Foreground Cascading Deletion, as discussed in #3225 but those changes will help to support this proposal.

Crossplane can reconcile the "external" or "added" finalizer(s) on Claims and Composite resources, cascading them all the way down to the edge nodes/Managed Resources, so that the declared dependency is enforced throughout the entire tree of resources, and not just on the single Claim/Composite.  This will ensure that the dependencies are maintained as long as the finalizer exists, and that the dependency Managed Resources are not deleted until the finalizer has been removed by the dependent resource.

The finalizer reconciliation should exclude all crossplane.io finalizers, and the foregroundDeletion finalizer used by kubernetes.

By ensuring that these dependencies are properly maintained, resource cleanup can be "orchestrated" by functionality such as the dependsOn feature in the Kubernetes Crossplane provider.  It also ensures that Crossplane behaves "as expected" by other Kubernetes ecosystem resources.

This change can be limited to the claim and composite reconcilers in the Crossplane project.  The Managed resource reconciler in the crossplane-runtime project should not require any changes.

While this feature should not affect any "normal" usage of Crossplane, it may be desirable to have this functionality controlled by a feature flag for the near term, until the feature is well understood and well tested.

This proposal is dependent on the crossplane-runtime changes proposed in https://github.com/crossplane/crossplane-runtime/pull/355 and the corresponding crossplane project changes proposed in https://github.com/crossplane/crossplane/pull/3321

Feedback and suggestions are welcome

bobh66:
This change adds reconciliation of finalizers to composite and composed resources.  If a Claim or Composite resource has more than one finalizer, then some other resource controller has declared a dependency on the Claim/Composite, and it is necessary to reflect that dependency on all of the resources that the Claim/Composite is composed of.

This will cause additional finalizers to be added to all of the "lower level" composite and managed resources so that they cannot be deleted by Kubernetes/Crossplane until the dependency has been removed from the Claim/Composite.

These changes are dependent on #3321 and https://github.com/crossplane/crossplane-runtime/pull/355

Also see #3225 for more background

Fixes #3323 

I have:

Unit tests have been added and testing in a development cluster is in progress.


negz (changes requested):
I'm not comfortable with this approach in which a "parent" controller manages the finalizers of its children. I'd prefer to keep controllers restricted to managing the finalizers of the types they reconcile (i.e. the claim controller should only be able to add/remove finalizers to/from claims to indicate that it must do work or assess conditions before allowing a delete to complete).

I believe the spirit of this change is to enable "in use" deletion protection. I have a few alternative approaches we might consider that I will share ASAP.

bobh66:
I've seen a variety of descriptions of finalizers, and there does not appear to be any consensus or documentation of what they are intended for and who they should be used by.  The PVC use is one example, others are in namespace and job handling.  Garbage collection uses them for it's own purposes with foregroundDeletion and it's not a controller.

I have seen a number of discussions like this - https://www.howtogeek.com/devops/what-are-finalizers-in-kubernetes-how-to-handle-object-deletions/ - which define finalizers as "a mechanism for enforcing certain conditions be met before an object can be deleted" and make no mention of restrictions on who can/should create and manage them.  Without a formal definition it's hard to know what constitutes a proper or improper usage.

I don't know that the changes in #3321 or the related crossplane-runtime changes are useful in practice without this PR, so I'll wait to see the alternatives before doing any more work on this option.

negz:
> I have a few alternative approaches we might consider that I will share ASAP.

See https://docs.google.com/document/d/1Yu3GxNMJOOMWMjD6gEw5QtT8Fz5qQSyo18ZZFKqcqKw/edit#

> Without a formal definition it's hard to know what constitutes a proper or improper usage.

This is true. There's little explicit official guidance around patterns here.

That said, I do strongly suspect this approach is at best novel/atypical - I can't find any prior art for a "dependent" controller adding finalizers to its dependencies apart from our own provider-kubernetes. It seems more typical for the controller for the resource that _is_ the dependency to be aware of this possibility, detect dependents, and manage its own finalizers accordingly. See for example:

* The [PV protection controller](https://github.com/kubernetes/kubernetes/blob/1294f557b90b5221e19d4fe3bffb06c09cd8ea81/pkg/controller/volume/pvprotection/pv_protection_controller.go) and [PVC protection controller](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/volume/pvcprotection/pvc_protection_controller.go).
* The Crossplane [ProviderConfig controller(s)](https://github.com/crossplane/crossplane-runtime/blob/619edbacb81eb9b4ccb6a5a7ba8f07d74cc07dc9/pkg/reconciler/providerconfig/reconciler.go#L188).
* [This article on the pattern](https://www.ibm.com/cloud/blog/preserve-kubernetes-api-objects-while-they-are-in-use).
* The cascading deletion implementation (where I'd consider the parent the "dependency" in a deletion-ordering sense).

My concern is that having the dependent controller add finalizers (or someday liens) to its dependencies breaks the "compartmentalization" of controller responsibility. I prefer to keep the code that detects whether a resource should be allowed to be deleted aligned with that resource. I like that this means the logic that determines whether (for example) an RDSInstance can or cannot be deleted is packaged alongside (or in) the RDSInstance controller (i.e. in provider-aws, not core Crossplane).

Issue #151, opened 2018-11-23
Add support for a default ResourceClass


bassam:
We should support a default resource class for a given Kind. This is a stepping stone on the path to something like #91, but can improve the overall experience immediately.

This can implemented similar to `StorageClass` in [Kubernetes](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#class-1) by setting an annotation:

```
apiVersion: core.crossplane.io/v1alpha1
kind: ResourceClass
metadata:
  name: standard
  namespace: crossplane-system
  annotations:
    storageclass.kubernetes.io/is-default-class: "true"
provisioner: mysql.rds.aws.crossplane.io
...
```

or we could make it part of the ResourceClass itself:

```
apiVersion: core.crossplane.io/v1alpha1
kind: ResourceClass
metadata:
  name: standard
  namespace: crossplane-system
default: true
provisioner: mysql.rds.aws.crossplane.io
...
```

By setting default to `true` that means if an abstract resource does not specify a classRef the default storage class will be selected for a given provisioner. There can only be a single default for a  provisioner.

PR #3337, opened 2022-09-22
Docs hugo migration

plumbis:

The work to migrate crossplane.github.io to Hugo has been completed in the site repo [PR #151](https://github.com/crossplane/crossplane.github.io/pull/151).

The content of the docs is pushed from /docs into the site repo via Make.

This PR provides the following updates:
1. Updates Makefile to use the correct directory for Hugo content (`/content`)
2. Updates all of the existing docs content to comply with Hugo's metadata (called `frontmatter`) and content organization expectations. This mainly involves moving files from `Overview.md` to `_index.md` at the root of all directories.
3. Creates a docs contributing guide to detail how to use Hugo, how we use `frontmatter` and our CoC and community guidelines. (in `docs/contributing/docs.md`)

I also verified that no changes seem to be required to`/build`.

Fixes # https://github.com/crossplane/crossplane.github.io/issues/149

This will need to be merged shortly after the site PR is merged.

I have:

Tested through manual verification in a Hugo environment. 

The changes to `Makefile` **haven't** been tested.

### DIFF
--- a/docs/api-docs/crossplane.md
+++ b/docs/api-docs/crossplane.md
@@ -1,7 +1,5 @@
 ---
-title: crossplane
-toc: true
-weight: 401
-indent: true
-redirect_to: https://doc.crds.dev/github.com/crossplane/crossplane
+title: Crossplane
+layout: redirect
+to: "https://crossplane.io/docs/v1.9/api-docs/crossplane.html"
 ---
--- a/docs/api-docs/crossplane.md
+++ b/docs/api-docs/crossplane.md
@@ -1,5 +1,5 @@
 ---
 title: Crossplane
 layout: redirect
-to: "https://crossplane.io/docs/v1.9/api-docs/crossplane.html"
+to: "https://doc.crds.dev/github.com/crossplane/crossplane"
 ---
--- a/docs/api-docs/provider-alibaba.md
+++ b/docs/api-docs/provider-alibaba.md
@@ -1,7 +1,5 @@
 ---
 title: provider-alibaba
-toc: true
-weight: 402
-indent: true
-redirect_to: https://doc.crds.dev/github.com/crossplane/provider-alibaba
+to: https://doc.crds.dev/github.com/crossplane/provider-alibaba
+layout: redirect
 ---
--- a/docs/concepts/_index.md
+++ b/docs/concepts/_index.md
@@ -1,10 +1,7 @@
 ---
 title: Concepts
-toc: true
 weight: 100
 ---
-
-# Overview
 
 Crossplane introduces multiple building blocks that enable you to provision,
 compose, and consume infrastructure using the Kubernetes API. These individual
@@ -49,10 +46,10 @@ learn more about all of these concepts in the [composition documentation].
 
 <!-- Named Links -->
 
-[Packages]: packages.md
+[Packages]: {{<ref "packages" >}}
 [CRDs]: https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/
 [controllers]: https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/#custom-controllers
-[providers documentation]: providers.md
+[providers documentation]: {{<ref "providers" >}}
 [doc.crds.dev]: https://doc.crds.dev
-[managed resources documentation]: managed-resources.md
-[composition documentation]: composition.md
+[managed resources documentation]: {{<ref "managed-resources" >}}
+[composition documentation]: {{<ref "composition" >}}
--- a/docs/contributing/docs.md
+++ b/docs/contributing/docs.md
@@ -1,0 +1,187 @@
+---
+title: "Crossplane Documentation"
+weight: 2000
+---
+
+- [Code of Conduct](#code-of-conduct)
+  - [Reporting](#reporting)
+- [Licensing](#licensing)
+  - [Directory structure](#directory-structure)
+  - [Run Hugo](#run-hugo)
+  - [Documentation issues and feature requests](#documentation-issues-and-feature-requests)
+- [Contributing](#contributing)
+  - [Front matter](#front-matter)
+    - [New Crossplane versions](#new-crossplane-versions)
+    - [Excluding pages from the Table of Contents](#excluding-pages-from-the-table-of-contents)
+  - [Links](#links)
+    - [Linking between docs pages](#linking-between-docs-pages)
+    - [Linking to external sites](#linking-to-external-sites)
+  - [Tabs](#tabs)
+
+## Code of Conduct
+We follow the [CNCF Code of Conduct](https://github.com/cncf/foundation/blob/main/code-of-conduct.md).
+
+Taken directly from the CNCF CoC:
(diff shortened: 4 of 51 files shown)


docs/contributing/docs.md
```diff
@@ -0,0 +1,187 @@
+---
+title: "Crossplane Documentation"
```
hasheddan, line 2: We typically try to wrap docs at 80 characters to make it easier to comment on smaller sections. Would you be open to doing that in this doc (and potentially add it as part of this guide as well)?

docs/api-docs/crossplane.md
```diff
@@ -1,7 +1,5 @@
 ---
-title: crossplane
-toc: true
-weight: 401
-indent: true
-redirect_to: https://doc.crds.dev/github.com/crossplane/crossplane
----
+title: Crossplane
+layout: redirect
+to: "https://crossplane.io/docs/v1.9/api-docs/crossplane.html"
```
hasheddan, line 4: Is this a vanity link? If so, should it use `master` on `master` branch?

docs/api-docs/provider-alibaba.md
```diff
@@ -1,7 +1,5 @@
 ---
 title: provider-alibaba
```
hasheddan, line 2: Just double checking, these provider docs will show in TOC right?

docs/concepts/_index.md
```diff
@@ -49,10 +46,10 @@ learn more about all of these concepts in the [composition documentation].
 
 <!-- Named Links -->
 
-[Packages]: packages.md
+[Packages]: {{<ref "packages" >}}
```
hasheddan, line 49: Is the `ref` syntax required? Not a _huge_ deal but is kind of a bummer that they don't render as markdown links on GitHub.

docs/contributing/docs.md
```diff
@@ -0,0 +1,187 @@
...
+
+### Reporting
+To report violations contact the Crossplane maintainers at `info@crossplane.io` or the CNCF at `conduct@cncf.io`.
+
+## Licensing
+The Crossplane documentation is under the [Creative Commons Attribution](https://creativecommons.org/licenses/by/4.0/) license. CC-BY allows reuse, remixing and republishing of Crossplane documentation with attribution to the Crossplane organization.
+
+
+### Directory structure
+The relevant directories of the docs repository are:
```
hasheddan, line 37: This is the directory structure for the docs site repository right? It might be worth mentioning that we are _not_ talking about `crossplane/crossplane`.

hasheddan (commented):
Thanks for this awesome effort @plumbis! Left a few comments / questions below :)

docs/guides/self-signed-ca-certs.md
```diff
@@ -51,6 +47,3 @@ in an `override.yaml` file would look like this:
     name: ca-bundle-config
     key: ca-bundle
 ```
-
-
-[Install Crossplane]: ../reference/install.md
```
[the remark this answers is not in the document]
plumbis: Yes, when using `{{<ref ... >}}` shortcode Hugo will find the markdown page that is associated with this path/title and put it in the markdown. The result _should_ be identical to the markdown engine.

The magic is that if the page moves or is renamed, the link breaks. With traditional markdown links this is just a silently broken link. By using the shortcode Hugo will fail to start. This draws attention to update the link.

99% of the time we're not trying to help the person writing the link but trying to notify the person moving pages to update the links pointing to that page.

docs/contributing/docs.md
plumbis: I don't think this works well for text and makes it harder to read in an editor. 

Not only is it unlikely to be viewed on a terminal where the line limit makes sense but it becomes very messy when you're including links. [For example](https://github.com/crossplane/crossplane/blob/432981c7e27063f89376421189d20b4acb7e643d/docs/guides/argo-cd-crossplane.md?plain=1#L18) the link alone is 70 characters.

docs/api-docs/crossplane.md
plumbis: This was an archive from testing, good catch.

docs/api-docs/provider-alibaba.md
plumbis: Yes, that's why this looks like a weird implementation. The pages have to be here with this metadata to appear in the menu but we can't enforce a redirect from the markdown page so there are a few hoops that have to be jumped through to make it happen. 

You can see in the preview site for the crossplane.github.io PR. https://master.d1bmjmu5ixie8d.amplifyapp.com/v1.9/api-docs/

docs/concepts/_index.md
plumbis: No, it's not, but to the earlier comment, it's an easy and fast way to detect broken links. 

Although it doesn't render on github I think you'll get more value by running hugo locally and using that as a method for review, since that provides the visuals as well.

docs/contributing/docs.md
plumbis: I updated this to be a link to `crossplane/crossplane.github.io` repo to clarify.

docs/guides/self-signed-ca-certs.md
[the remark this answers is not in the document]
hasheddan: Gotcha, thanks for the explanation! I think this is a reasonable outlook and ultimately will lead to us having higher quality documentation. We are paying a cost in review as folks cannot see the links in the GitHub rendered markdown, but that feels like the right trade-off to make (i.e. let's rely on Hugo machinery to not let us publish instead of human eyes noticing bad links). That being said, I don't think we need the whole Hugo pipeline in this repo necessarily, but we should run at least enough to verify that links are not broken in CI so we can avoid merging bad content.

docs/contributing/docs.md
hasheddan: > I don't think this works well for text and makes it harder to read in an editor.

We have had pretty strong consensus on doing this in the past due to the GitHub suggestion granularity, but I am inclined to defer to the person who literally writes documentation for a living :) If there is strong push back we can always change in the future, and we can preserve the practice of wrapping in design docs for now where we are perhaps going to give more granular feedback.

plumbis: +1 to being willing to change in the future and thanks for being flexible.

docs/guides/self-signed-ca-certs.md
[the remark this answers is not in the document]
plumbis: Right now the only thing we have is the content, none of the layout or "shortcodes" that render links like that. Those live exclusively in the Crossplane.io repo ([example](https://github.com/upbound/crossplane.github.io)) . 

Historically CI (for the links) has been done by just running hugo against the complete repo- if it compiles we pass, if it doesn't then there's something wrong (e.g., broken link). To do that we would have to bring in the rest of the hugo files and configurations.

Defining this process doesn't need to be a blocker for this PR but I think we can checkout the main/master branch of crossplane.io and copy the docs from _this_ repo into the clone and run hugo. I don't imagine this will be very complicated.

hasheddan (approved):
Thanks @plumbis! We are going to follow this up with some pre-merge checks that ensure content builds correctly 👍🏻

hasheddan (approved):
Thanks for all of your effort here @plumbis! 🙌🏻

negz (approved):
I didn't review closely - the Helm chart README change requires my steering committee stamp. I trust @hasheddan's review.

Issue #3339, opened 2022-09-26
Consider removing `aggregationRule` to support working in restricted environments


Piotr1215:
### What problem are you facing?

Crossplane requires admin privileges on the crossplane service account due to

```yaml
aggregationRule:
  clusterRoleSelectors:
  - matchLabels:
      [rbac.crossplane.io/aggregate-to-crossplane](http://rbac.crossplane.io/aggregate-to-crossplane): "true"
```
This prevents Crossplane from working in restricted environments.

### How could Crossplane help solve your problem?

Remove the `aggregationRule` to enable Crossplane to work in restricted environments.

negz:
> This prevents Crossplane from working in restricted environments.

Can you elaborate? What environments restrict aggregation rules? How do they do that?

Piotr1215:
This can be closed, addressed in this PR: https://github.com/crossplane/crossplane/pull/3343

PR #3343, opened 2022-09-28
Use aggregation in ClusterRole of Crossplane only if rbac-manager is deployed

muvaf:

Currently, rbac-manager deals with all `ClusterRole`s that Crossplane and providers need to have for CRDs introduced during installation and runtime, i.e. CRDs created for XRDs and CRDs of installed providers. The mechanism it does all that is role aggregation, which requires elevated permissions for the rbac-manager.

We have an option to disable rbac-manager for folks who'd like to be super strict about how they manage their RBAC, hence `rbacManager.deploy` option. However, even when this is `false`, we have one remaining `ClusterRole` that does role aggregation, which is Crossplane's `ClusterRole` itself. If rbac-manager is not deployed, then there is no other `ClusterRole`s to be aggregated to the original one other then the static one in the Helm chart. So, this change proposes that we use aggregation on that `ClusterRole` only if rbac-manager is deployed so that Crossplane doesn't require elevated permissions required by aggregation when we can get away with static `ClusterRole` definition.

Related to https://github.com/crossplane/crossplane/issues/3339

I have:

Manually tested via `helm install --set rbacManager.deploy=false` and saw that it was able to install CRDs and provider, which is what that main `ClusterRole` allows it to do, so it's effective.

### DIFF
--- a/cluster/charts/crossplane/templates/clusterrole.yaml
+++ b/cluster/charts/crossplane/templates/clusterrole.yaml
@@ -5,6 +5,7 @@ metadata:
   labels:
     app: {{ template "crossplane.name" . }}
     {{- include "crossplane.labels" . | indent 4 }}
+{{- if .Values.rbacManager.deploy }}
 aggregationRule:
   clusterRoleSelectors:
   - matchLabels:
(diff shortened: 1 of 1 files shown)


hasheddan (approved):
This seems reasonable to me, thanks @muvaf!

PR #3349, opened 2022-09-30
Add support for a pause annotation on composite resources & claims, which pauses reconciliations

ulucinar:

Similar to https://github.com/crossplane/crossplane-runtime/issues/351 with this PR, if a composite resource has the `crossplane.io/paused` annotation with its value set to `true`, then the reconciler emits an event indicating that further reconciliations on that composite are paused and returns early after setting a `Synced` status condition with `False` status and `ReconcilePaused` reason. If the annotation is removed or its value is changed, reconciliation resumes.

*NOTE*: This PR also updates the crossplane-runtime Go mod dependency. 

I have:

Manually tested using the following manifests:

```yaml
metadata:
  name: xmyresources.test.com
spec:
  group: test.com
  names:
    kind: XMyResource
    plural: xmyresources
  claimNames:
    kind: MyResource
    plural: myresources
  versions:
  - name: v1alpha1
    served: true
    referenceable: true
    schema:
      openAPIV3Schema:
        type: object
        properties:
          spec:
            type: object
            properties:
              parameters:
                type: object
                properties:
                  tagValue:
                    type: string
                required:
                - tagValue
            required:
            - parameters
```

```yaml
kind: Composition
metadata:
  name: example
  labels:
    purpose: example
spec:
  compositeTypeRef:
    apiVersion: test.com/v1alpha1
    kind: XMyResource
  resources:
    - name: myrole
      base:
        apiVersion: iam.aws.crossplane.io/v1beta1
        kind: Role
        metadata:
        spec:
          forProvider:
            assumeRolePolicyDocument: |
              {
                "Version": "2012-10-17",
                "Statement": [
                    {
                        "Effect": "Allow",
                        "Principal": {
                            "Service": [
                                "ec2.amazonaws.com",
                                "eks.amazonaws.com",
                                "eks-fargate-pods.amazonaws.com",
                                "lambda.amazonaws.com",
                                "s3.amazonaws.com",
                                "rds.amazonaws.com",
                                "dax.amazonaws.com"
                            ]
                        },
                        "Action": [
                            "sts:AssumeRole"
                        ]
                    }
                ]
              }
            tags:
              - key: Name
                value: alper-fix-351
              - key: patch
      patches:
        - fromFieldPath: "spec.parameters.tagValue"
          toFieldPath: spec.forProvider.tags[1].value
```

```yaml
apiVersion: test.com/v1alpha1
kind: MyResource
metadata:
  namespace: test
  name: my-resource
spec:
  parameters:
    tagValue: alper-test
  compositionRef:
    name: example
```

When the `crossplane.io/paused` annotation with value `true` is set on the composite resource of kind `XMyResource`, it acquires the following `Synced` status condition:

```
Status:
  Conditions:
    Last Transition Time:  2022-09-30T17:27:00Z
    Reason:                ReconcilePaused
    Status:                False
    Type:                  Synced
    Last Transition Time:  2022-09-30T16:01:10Z
    Reason:                Available
    Status:                True
    Type:                  Ready
```

and the following event with reason `ReconciliationPaused`:

```
Events:
  Type    Reason                Age                  From                                                             Message
  ----    ------                ----                 ----                                                             -------
  Normal  SelectComposition     26m (x52 over 100m)  defined/compositeresourcedefinition.apiextensions.crossplane.io  Successfully selected composition
  Normal  ReconciliationPaused  19m (x4 over 21m)    defined/compositeresourcedefinition.apiextensions.crossplane.io  Reconciliation is paused via the pause annotation
  Normal  ComposeResources      10m (x11 over 22m)   defined/compositeresourcedefinition.apiextensions.crossplane.io  Successfully composed resources
  Normal  SelectComposition     2m1s (x19 over 22m)  defined/compositeresourcedefinition.apiextensions.crossplane.io  Successfully selected composition
```

When the annotation is removed, reconciliation resumes and the `Synced` status condition either stays at `False` status with reason `ReconcileError` or transitions to `True` status with reason `ReconcileSuccess`:

```
Status:
  Conditions:
    Last Transition Time:  2022-09-30T17:32:29Z
    Reason:                ReconcileSuccess
    Status:                True
    Type:                  Synced
    Last Transition Time:  2022-09-30T16:01:10Z
    Reason:                Available
    Status:                True
    Type:                  Ready
```

The same set of experiments has also been performed on the claim. One notable observation regarding claims is that claim's annotations are normally passed on the associated XR. However:
- When we add the `crossplane.io/paused: true` annotation to the claim, because reconciliations on the claim are now paused, this new annotation does not get propagated to the XR, as expected. 
- If this value of the `crossplane.io/paused` annotation is changed from `true` to some other value, reconciliations on the claim resume, and thus the  `crossplane.io/paused` annotation is passed to the XR with the non-`true` value.
- If the `crossplane.io/paused` annotation is removed from the claim but has been set to a non-`true` value before being removed, this annotation stays on the XR but with a non-`true` value, thus not blocking reconciliations on the XR, as expected. 

### DIFF
--- a/docs/concepts/managed-resources.md
+++ b/docs/concepts/managed-resources.md
@@ -316,6 +316,25 @@ never deletes the external resource in the provider.
 > means Crossplane will allow immutable fields to be changed, but will not
 > actually make the desired change. This is tracked in [this issue][issue-727].
 
+#### Pausing Reconciliations
+If a managed resource being reconciled by the [managed reconciler], has the
+`crossplane.io/paused` annotation with its value set to `true` as in the
+following example, then further reconciliations are paused on that resource
+after emitting an event with the type `Synced`, the status `False`,
+and the reason `ReconcilePaused`:
+```yaml
+apiVersion: ec2.aws.upbound.io/v1beta1
+kind: VPC
+metadata:
+  name: paused-vpc
+  annotations:
+    crossplane.io/paused: "true"
+...
+```
+Reconciliations on the managed resource will resume once the 
+`crossplane.io/paused` annotation is removed or its value is set
+to anything other than `true`.
+
 ### External Name
 
 By default the name of the managed resource is used as the name of the external
@@ -468,3 +487,4 @@ including Velero.
 [issue-727]: https://github.com/crossplane/crossplane/issues/727
 [issue-1143]: https://github.com/crossplane/crossplane/issues/1143
 [managed-api-patterns]: https://github.com/crossplane/crossplane/blob/master/design/one-pager-managed-resource-api-design.md
+[managed reconciler]: https://github.com/crossplane/crossplane-runtime/blob/84e629b9589852df1322ff1eae4c6e7639cf6e99/pkg/reconciler/managed/reconciler.go#L637
--- a/docs/reference/composition.md
+++ b/docs/reference/composition.md
@@ -376,6 +376,45 @@ spec:
       fromFieldPath: metadata.labels[some-important-label]
 ```
 
+### Pause Annotation
+There is an annotation named `crossplane.io/paused` that you can use on
+Composite Resources and Composite Resource Claims to temporarily pause
+reconciliations of their respective controllers on them. An example
+for a Composite Resource Claim is as follows:
+```yaml
+apiVersion: test.com/v1alpha1
+kind: MyResource
+metadata:
+  annotations:
+     crossplane.io/paused: true
+  namespace: upbound-system
+  name: my-resource
+spec:
+  parameters:
+    tagValue: demo-test
+  compositionRef:
+    name: example
+```
+where `MyResource` is a Composite Resource Claim kind.
+When a Composite Resource or a Claim has the `crossplane.io/paused` annotation
+with its value set to `true`, the Composite Resource controller or the Claim
+controller pauses reconciliations on the resource until
+the annotation is removed or its value set to something other than `true`.
+Before temporarily pausing reconciliations, an event with the type `Synced`,
+the status `False`, and the reason `ReconcilePaused` is emitted
+on the resource.
+Please also note that annotations on a Composite Resource Claim are propagated
+to the associated Composite Resource but when the
+`crossplane.io/paused: true` annotation is added to a Claim, because
+reconciliations on the Claim are now paused, this newly added annotation
+will not be propagated. However, whenever the annotation's value is set to a
+non-`true` value, reconciliations on the Claim will now resume, and thus the
+annotation will now be propagated to the associated Composite Resource
+with a non-`true` value. An implication of the described behavior is that
+pausing reconciliations on the Claim will not inherently pause reconciliations
+on the associated Composite Resource.
+
+
 ### Patch Types
 
 You can use the following types of patch in a `Composition`:
--- a/internal/controller/apiextensions/claim/reconciler_test.go
+++ b/internal/controller/apiextensions/claim/reconciler_test.go
@@ -31,11 +31,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
-	"github.com/crossplane/crossplane-runtime/pkg/meta"
-
 	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
 	"github.com/crossplane/crossplane-runtime/pkg/errors"
 	"github.com/crossplane/crossplane-runtime/pkg/logging"
+	"github.com/crossplane/crossplane-runtime/pkg/meta"
 	"github.com/crossplane/crossplane-runtime/pkg/resource"
 	"github.com/crossplane/crossplane-runtime/pkg/resource/fake"
 	"github.com/crossplane/crossplane-runtime/pkg/resource/unstructured/claim"
--- a/internal/controller/apiextensions/claim/reconciler_test.go
+++ b/internal/controller/apiextensions/claim/reconciler_test.go
@@ -31,6 +31,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
+	"github.com/crossplane/crossplane-runtime/pkg/meta"
+
 	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
 	"github.com/crossplane/crossplane-runtime/pkg/errors"
 	"github.com/crossplane/crossplane-runtime/pkg/logging"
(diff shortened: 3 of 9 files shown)


ulucinar:
There might be a sporadic regarding status condition timestamps that I'll check later...

internal/controller/apiextensions/composite/reconciler.go
```diff
@@ -403,6 +404,17 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		"name", cr.GetName(),
 	)
 
+	// Check the pause annotation and return if it has the value "true"
+	// after logging, publishing an event and updating the SYNC status condition
+	if meta.IsPaused(cr) {
+		log.Debug("Reconciliation is paused via the pause annotation", "annotation", meta.AnnotationKeyReconciliationPaused, "value", "true")
+		r.record.Event(cr, event.Normal(reasonPaused, "Reconciliation is paused via the pause annotation"))
+		cr.SetConditions(xpv1.ReconcilePaused())
```
ytsarev, line 412: Should we also propagate `meta.AnnotationKeyReconciliationPaused` down to all composed MRs? Otherwise, if associated Managed Resources will not have pause annotation, they still will be reconciled by the associated provider controller, is that correct?

internal/controller/apiextensions/composite/reconciler.go
```diff
@@ -403,6 +404,17 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+		// If the pause annotation is removed, we will have a chance to reconcile again and resume
+		// and if status update fails, we will reconcile again to retry to update the status
+		return reconcile.Result{}, errors.Wrap(r.client.Status().Update(ctx, cr), errUpdateStatus)
+	}
+
 	if meta.WasDeleted(cr) {
```
bobh66, line 418: What happens when someone wants to delete a resource that has reconciliation paused?   It seems like the resource will be "stuck" until the paused annotation is removed?  I would think deletion should take precedence as the resource may be in the middle of a larger chain or someone may forget that they placed the paused annotation.

internal/controller/apiextensions/composite/reconciler.go
bobh66: That's a good point, and reconciling the paused annotation gets tricky because how does the reconciler know when to remove the annotation when the paused composite is in the middle of a larger graph?  Which resource is the "root" of the paused resource graph?  This is the same problem I found when I was looking at reconciling finalizers - I could not just add the same finalizer to all of the sub-resources because you quickly lose track of which resource is the "root" and there is no way to tell when to remove it.

I ended up adding a unique finalizer for the object to it's sub-resources if the object had more than one finalizer, and removing the unique finalizer from the sub-resources if the object had only 1 finalizer (it's own).  I think this would require something similar - if the paused annotation exists on an object, add a different annotation (paused-by-parent?) to the sub-resources so that reconciliation can figure out when to remove the paused-by-parent annotations, and then check for either annotation in the reconciler to skip reconciliation.

internal/controller/apiextensions/composite/reconciler.go
```diff
@@ -403,6 +404,17 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
 
+	// Check the pause annotation and return if it has the value "true"
+	// after logging, publishing an event and updating the SYNC status condition
+	if meta.IsPaused(cr) {
+		log.Debug("Reconciliation is paused via the pause annotation", "annotation", meta.AnnotationKeyReconciliationPaused, "value", "true")
+		r.record.Event(cr, event.Normal(reasonPaused, "Reconciliation is paused via the pause annotation"))
+		cr.SetConditions(xpv1.ReconcilePaused())
+		// If the pause annotation is removed, we will have a chance to reconcile again and resume
+		// and if status update fails, we will reconcile again to retry to update the status
+		return reconcile.Result{}, errors.Wrap(r.client.Status().Update(ctx, cr), errUpdateStatus)
```
bobh66, line 415: I assume that the reconciler will still run every _poll interval_ seconds/minutes and see that this is paused, so there will be a debug log message and an event on every interval.  Just verifying that this is the expected behavior - it's probably what we want to make it clear that something is not being reconciled.

internal/controller/apiextensions/composite/reconciler_test.go
```diff
@@ -942,6 +943,111 @@ func TestReconcile(t *testing.T) {
...
+					o.SetConditions(xpv1.ReconcilePaused())
+				}),
+			},
+		},
+		"ReconciliationResumes": {
+			reason: `If a composite resource has the pause annotation with some value other than "true" and the Synced=False/ReconcilePaused status condition, reconciliation should resume with requeueing.`,
+			args: args{
+				mgr: &fake.Manager{},
+				composite: withComposite(func(o resource.Composite) {
+					o.SetAnnotations(map[string]string{meta.AnnotationKeyReconciliationPaused: ""})
```
bobh66, line 990: Is it worth a test case for when the entire annotation is removed, not just empty?  Maybe it's the same logic either way.

bobh66 (changes requested):

internal/controller/apiextensions/composite/reconciler.go
muvaf: I think that would be expected behavior. The user tells to stop reconciling towards the desired state and resource being deleted is the desired state. So, what I'd look for in such "stuck" cases is whether we can do a good job of letting them know that's why they're stuck and event + condition pair seems like the best we could do there.

internal/controller/apiextensions/composite/reconciler.go
muvaf: I think it's important to align on what use cases we have for pause annotation at the moment. In https://github.com/crossplane/crossplane-runtime/issues/351 , my main motivation was to enable smooth migration operation and in that context you're dealing with individual resources directly, hence it'd be surprising to see that the pause annotation you put on one k8s resource gets propagated to others. It's essentially surgery and I think keeping the impact as scoped as possible would make operator feel more in control.

internal/controller/apiextensions/composite/reconciler.go
```diff
@@ -403,6 +404,17 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		"name", cr.GetName(),
 	)
 
+	// Check the pause annotation and return if it has the value "true"
+	// after logging, publishing an event and updating the SYNC status condition
+	if meta.IsPaused(cr) {
```
muvaf, line 409: @ulucinar I think there are a couple of other places we need to update:
* [XRM doc](https://github.com/crossplane/crossplane/blob/master/design/one-pager-managed-resource-api-design.md) should be updated to include this annotation.
* [This docs](https://github.com/crossplane/crossplane/blob/master/docs/concepts/managed-resources.md) page should document the behavior.
* Claims should also be stopped from reconciliation since this is part of XRM now.

internal/controller/apiextensions/composite/reconciler.go
ytsarev: @muvaf definitely fine to keep the scope narrow. But from a pure UX standpoint, pausing the Claim/XR that would result in pausing all related composed resources seems natural to me.

chlunde:
FYI, I created a related issue here: https://github.com/crossplane/crossplane-runtime/issues/360

internal/controller/apiextensions/composite/reconciler.go
ulucinar: I also support per-controller behavior here instead of defining an API (the pause annotation) with higher-level semantics, like controlling a tree of resources rooted at the XR. Instead, this annotation would then just pause the reconciliations on the XR.

internal/controller/apiextensions/composite/reconciler.go
ulucinar: As long as the status update succeeds here, there should be no further reconciliation requests, unless an external event (like editing of the XR by an external client) needs to be handled. The periodic reconciliations normally scheduled every poll interval are also stopped here because `Result.RequeueAfter` is 0.

It could still be the case that the workerqueue is not empty and in such cases, there will be a limited number of further reconciliations after the a paused event is posted but the number of such reconciliations is limited by the queue capacity and in practice you may observe only a few of such repeating events & logs.

internal/controller/apiextensions/composite/reconciler_test.go
ulucinar: Yes, you are right: we check the value corresponding to the key is `true` or not, and the cases in which the annotation is missing or exists but has a value other than `true` are handled in the same way. The case in which this annotation does not exist is already tested where we assert a `Synced` condition with a `True` status with reason `ReconcileSuccess`. But still we can add a test case that asserts the mere existence of the annotation will not be sufficient to pause reconciliations unless it has the value `true`. I will extend the test cases as you suggested.

internal/controller/apiextensions/composite/reconciler.go
ulucinar: Thank you @muvaf. Added & tested support for pause annotation for claims also. One observation is that claims pass their annotations to XRs. I will add more observations into the description related to this behavior.

internal/controller/apiextensions/claim/reconciler_test.go
```diff
@@ -31,6 +31,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
+	"github.com/crossplane/crossplane-runtime/pkg/meta"
+
```
muvaf, line 35: Could you merge this with the list below?

internal/controller/apiextensions/composite/reconciler.go
muvaf: @ulucinar I don't think the docs got an update. Could you try to add it today so that we merge it before the code freeze tomorrow?

internal/controller/apiextensions/claim/reconciler_test.go
ulucinar: Done.

docs/reference/composition.md
```diff
@@ -376,6 +376,35 @@ spec:
...
+     crossplane.io/paused: true
+  namespace: upbound-system
+  name: my-resource
+spec:
+  parameters:
+    tagValue: demo-test
+  compositionRef:
+    name: example
+```
+, where `MyResource` is a Composite Resource Claim kind.
```
muvaf, line 398: Is this comma a leftover?

docs/reference/composition.md
```diff
@@ -376,6 +376,35 @@ spec:
...
+    tagValue: demo-test
+  compositionRef:
+    name: example
+```
+, where `MyResource` is a Composite Resource Claim kind. 
+When a Composite Resource or a Claim has the `crossplane.io/paused` annotation
+with its value set to `true`, the Composite Resource controller or the Claim
+controller pauses reconciliations on the resource until
+the annotation is removed or its value set to something other than `true`.
+Before temporarily pausing reconciliations, an event with the type `Synched`,
```
muvaf, line 403:
```suggestion
Before temporarily pausing reconciliations, an event with the type `Synced`, 
```

docs/reference/composition.md
```diff
@@ -376,6 +376,35 @@ spec:
...
+  namespace: upbound-system
+  name: my-resource
+spec:
+  parameters:
+    tagValue: demo-test
+  compositionRef:
+    name: example
+```
+, where `MyResource` is a Composite Resource Claim kind. 
+When a Composite Resource or a Claim has the `crossplane.io/paused` annotation
```
muvaf, line 399: Could you also note that the annotations are propagated from claim to composite, hence adding it to claim will in practice pause composite, too?

muvaf (approved):
LGTM! A few small things before merging. Thanks @ulucinar !

muvaf:
@bobh66 Could you take a look? We can't merge with you requesting changes.

internal/controller/apiextensions/composite/reconciler.go
bobh66: It's possible to "pause" a controller by setting replicas to 0 in ControllerConfig - that would at least let you do surgery on the MRs without the controller interfering.  I think this needs more iterations but it's a good start - we can make changes as we learn how it's being used.

internal/controller/apiextensions/composite/reconciler.go
bobh66: I think this has the potential to cause confusion but as long as it's documented it's fine for now.  Thanks.

bobh66 (approved):

docs/concepts/managed-resources.md
```diff
@@ -316,6 +316,25 @@ never deletes the external resource in the provider.
 > means Crossplane will allow immutable fields to be changed, but will not
 > actually make the desired change. This is tracked in [this issue][issue-727].
 
+#### Pausing Reconciliations
+If a managed resource being reconciled by the [managed reconciler], has the
+`crossplane.io/paused` annotation with its value set to `true` as in the
+following example, then further reconciliations are paused on that resource
+after emitting an event with the type `Synched`, the status `False`,
```
negz, line 323:
```suggestion
after emitting an event with the type `Synced`, the status `False`,
```

negz (approved):

docs/reference/composition.md
ulucinar: Done. Removed the comma.

docs/reference/composition.md
ulucinar: Thank you :) Fixed the typo.

docs/concepts/managed-resources.md
ulucinar: Thank you :) Fixed the typo.

docs/reference/composition.md
ulucinar: Added the following text explaining the behavior:
```
Please also note that annotations on a Composite Resource Claim are propagated
to the associated Composite Resource but when the
`crossplane.io/paused: true` annotation is added to a Claim, because
reconciliations on the Claim are now paused, this newly added annotation
will not be propagated. However, whenever the annotation's value is set to a
non-`true` value, reconciliations on the Claim will now resume, and thus the
annotation will now be propagated to the associated Composite Resource
with a non-`true` value. An implication of the described behavior is that
pausing reconciliations on the Claim will not inherently pause reconciliations
on the associated Composite Resource.
```

PR #3363, opened 2022-10-10
update example mapping to match example json

tomas-mota:
fixed json mapping of example field fancinesslevel

### DIFF
--- a/docs/contributing/provider_development_guide.md
+++ b/docs/contributing/provider_development_guide.md
@@ -194,7 +194,7 @@ type FavouriteDBInstanceParameters struct {
     // conventions trump cloud provider fidelity.
 
     // FancinessLevel specifies exactly how fancy this instance is.
-    FancinessLevel int `json:"fancinessLevel"`
+    FancinessLevel int `json:"fanciness_level"`
 
     // Version specifies what version of FancySQL this instance will run.
     // +optional


docs/contributing/provider_development_guide.md
```diff
@@ -194,7 +194,7 @@ type FavouriteDBInstanceParameters struct {
     // conventions trump cloud provider fidelity.
 
     // FancinessLevel specifies exactly how fancy this instance is.
-    FancinessLevel int `json:"fancinessLevel"`
+    FancinessLevel int `json:"fanciness_level"`
```
hasheddan, line 197: @tomas-mota we use camel case in all of our fields, so I believe this is correct as-is

negz:
@tomas-mota I agree with @hasheddan's comment here. We might be missing something though - could you please update your PR description with more context around what problem you're trying to solve. You'll also need to sign the commit for us to be able to merge it.

docs/contributing/provider_development_guide.md
tomas-mota: maybe there is some implicit mapping, because the json it is referring to looks like this:
image

hasheddan: I think the purpose is to demonstrate that we’ll translate snake case fields into camel case. See note:
```
// Note that fanciness_level becomes fancinessLevel below. Kubernetes API
     // conventions trump cloud provider fidelity.

```

tomas-mota: totally missed that comment, my bad!

PR #3366, opened 2022-10-11
Backport docs hugo migration to v1.8

plumbis:
Updates v1.8 docs to support publishing to hugo.


hasheddan (commented):
Thanks @plumbis! I think we may be missing the `Makefile` and helm chart README.md updates made in https://github.com/crossplane/crossplane/pull/3337.

PR #3391, opened 2022-10-17
Create composition revision names with partial hash

### LINKED ISSUE #3011: CompositionRevision duplicates
### What happened?
Hi,
as we are using more and more composition revision we may have found a bug. 
Or at least a really really unexpected behaviour.

We are facing composition revision duplicates.
Like this:

```
s3bucket-composition-4jsjv             2          False     14d
s3bucket-composition-4ns75             2          False     14d
s3bucket-composition-67rnl             3          False     4d8h
s3bucket-composition-6zh2c             1          False     26d
s3bucket-composition-df5ts             4          False     4d5h
s3bucket-composition-lf2st             5          True      3d6h
```

You see here that the revision `s3bucket-composition-4jsjv` and `s3bucket-composition-4ns75` having the same revision number. 

Comparing the yaml output the only difference is 
```
metadata.name
metadata.resourceVersion
metadata.uid
```

I guess its not expected to have duplicates of revision with the same content.

### How can we reproduce it?
Not sure, but maybe the following is a good hint:

- Enable composition revisions
- Create a composition
- Modify it a couple of times and apply it using kubectl apply 

if this is not enough:
- Let fluxcd apply your changes

Crossplane version: 
crossplane/crossplane:v1.6.3
crossplane/provider-aws-controller:v0.24.1

Maybe its important that we use:
FluxCD
for applying our changes to the cluster using Git as source

Best regards

ezgidemirel:
Signed-off-by: ezgidemirel <ezgidemirel91@gmail.com>

Instead of using GenerateName, with this change, we'll start adding a partial hash suffix to prevent duplicate revision creations.   

Fixes #3011 

I have:

Created/updated compositions to verify that revisions are created with the same name for the same compositions. 

### DIFF
--- a/apis/apiextensions/v1/composition_hash.go
+++ b/apis/apiextensions/v1/composition_hash.go
@@ -17,15 +17,15 @@ limitations under the License.
 package v1
 
 import (
+	"crypto/sha256"
 	"fmt"
-	"hash/fnv"
 
 	"sigs.k8s.io/yaml"
 )
 
 // Hash of the CompositionSpec.
 func (cs CompositionSpec) Hash() string {
-	h := fnv.New64a()
+	h := sha256.New()
 	y, err := yaml.Marshal(cs)
 	if err != nil {
 		// I believe this should be impossible given we're marshalling a
@@ -33,5 +33,5 @@ func (cs CompositionSpec) Hash() string {
 		return "unknown"
 	}
 	h.Write(y) //nolint:errcheck // Writing to a hash never errors.
-	return fmt.Sprintf("%x", h.Sum64())
+	return fmt.Sprintf("%x", h.Sum(nil))
 }
--- a/internal/controller/apiextensions/composition/revision.go
+++ b/internal/controller/apiextensions/composition/revision.go
@@ -17,6 +17,8 @@ limitations under the License.
 package composition
 
 import (
+	"fmt"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/crossplane/crossplane-runtime/pkg/meta"
@@ -29,7 +31,7 @@ import (
 func NewCompositionRevision(c *v1.Composition, revision int64, compSpecHash string) *v1alpha1.CompositionRevision {
 	cr := &v1alpha1.CompositionRevision{
 		ObjectMeta: metav1.ObjectMeta{
-			GenerateName: c.GetName() + "-",
+			Name: fmt.Sprintf("%s-%s", c.GetName(), compSpecHash[0:5]),
 			Labels: map[string]string{
 				v1alpha1.LabelCompositionName:     c.GetName(),
 				v1alpha1.LabelCompositionSpecHash: compSpecHash,
--- a/internal/controller/apiextensions/composition/revision.go
+++ b/internal/controller/apiextensions/composition/revision.go
@@ -31,10 +31,12 @@ import (
 func NewCompositionRevision(c *v1.Composition, revision int64, compSpecHash string) *v1alpha1.CompositionRevision {
 	cr := &v1alpha1.CompositionRevision{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: fmt.Sprintf("%s-%s", c.GetName(), compSpecHash[0:5]),
+			Name: fmt.Sprintf("%s-%s", c.GetName(), compSpecHash[0:7]),
 			Labels: map[string]string{
-				v1alpha1.LabelCompositionName:     c.GetName(),
-				v1alpha1.LabelCompositionSpecHash: compSpecHash,
+				v1alpha1.LabelCompositionName: c.GetName(),
+				// We cannot have a label value longer than 63 chars
+				// https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set
+				v1alpha1.LabelCompositionSpecHash: compSpecHash[0:63],
 			},
 		},
 		Spec: NewCompositionRevisionSpec(c.Spec, revision),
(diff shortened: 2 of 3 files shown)


negz:
This looks good, thanks @ezgidemirel!

A few thoughts:

* This will be a breaking change, right? That's okay (its alpha), but we should probably include some documentation or release notes to let folks know that their composition revisions will all be recreated.
* Is it worth updating [this logic](https://github.com/crossplane/crossplane/blob/e8d1166/internal/controller/apiextensions/composition/reconciler.go#L170) to use the hash fragment from the name of the revision, rather than also adding that hash as a label?
* Would this be a good time to switch from using fnv1a to a cryptographic hash like sha256 to generate hashes of the Composition spec? I've historically preferred to avoid cryptographic hashes unless I knew they were needed (which is why I went with fnv1a), but someone recently made a compelling case that if you're not really worried about performance you may as well use a cryptographic hash "just in case". In this case I imagine using sha256 would protect against the case where someone maliciously crafted a composition spec that produced the same hash as an existing revision, which isn't something I'm really worried about.

ezgidemirel:
Hi @negz, thanks for your comment! You can find my answers below: 

> This will be a breaking change, right? That's okay (its alpha), but we should probably include some documentation or release notes to let folks know that their composition revisions will all be recreated.

This won't be a breaking change if we don't change the hash algorithm. I created composition revisions with generated names and update to this branch to see what will happen. Since, we decide to create a new revision based on the hash value stored in this [label](https://github.com/crossplane/crossplane/blob/e8d1166/internal/controller/apiextensions/composition/reconciler.go#L170), I didn't observe any new composition revisions. When I updated the composition, I saw that the new revision is created with the hash fragment suffix. 

> Is it worth updating [this logic](https://github.com/crossplane/crossplane/blob/e8d1166/internal/controller/apiextensions/composition/reconciler.go#L170) to use the hash fragment from the name of the revision, rather than also adding that hash as a label?

I think we should keep it. We want to reduce the number of collisions while comparing compositions and having a longer string as hashes will decrease collision probability. 

> Would this be a good time to switch from using fnv1a to a cryptographic hash like sha256 to generate hashes of the Composition spec? I've historically preferred to avoid cryptographic hashes unless I knew they were needed (which is why I went with fnv1a), but someone recently made a compelling case that if you're not really worried about performance you may as well use a cryptographic hash "just in case". In this case I imagine using sha256 would protect against the case where someone maliciously crafted a composition spec that produced the same hash as an existing revision, which isn't something I'm really worried about.

I also don't worry about a malicious attack with composition revisions. I would consider collision rates as a distinguisher. I did some research on the collision rates of cryptographic and non-cryptographic algorithms, specifically fnv1a and sha216, but couldn't find a good comparison. And even if I would, these experiment results usually depend on the characteristics of the used data set (like all numbers, lowercase or uppercase English words etc.). 

Since we start to use the first 5 characters of this hash to _prevent_ duplicate revision creations, I believe this will not guarantee that. We would have calculated different hashes but the first 5 characters might be the same.

negz:
> We want to reduce the number of collisions while comparing compositions and having a longer string as hashes will decrease collision probability.

Wouldn't we already have a collision at this point?

Let's say the Composition named `example` currently has hash `3ed5daaaaaaaaaaa`. We will take the first 5 chars of that hash and create a revision named `example-3ed5d`. If the same Composition was later edited such that it had hash `3ed5dbbbbbbbbbbb` we would attempt to create another revision with the same name - `example-3ed5d`. Despite the labels not having collided the abbreviated 5 character names do, so both revisions cannot exist at once.

FWIW avoiding situations like this is arguably another case for using a cryptographic hash, which typically have the property that small changes in plaintext (e.g. a small update to a Composition) have large changes to the resulting hash (see https://en.wikipedia.org/wiki/Avalanche_effect).

ezgidemirel:
> Wouldn't we already have a collision at this point?

That's correct. We have a higher chance of having collisions. We may increase the hash suffix length to reduce it (e.g., git uses the first 7 chars of the commit ID) but there is no guaranteed way of preventing it.  

> FWIW avoiding situations like this is arguably another case for using a cryptographic hash, which typically have the property that small changes in plaintext (e.g. a small update to a Composition) have large changes to the resulting hash (see https://en.wikipedia.org/wiki/Avalanche_effect).

That's a great point! If we change the hash algorithm, after deploying this change, we will have a new composition revision with the same spec but a different hash value. Claims/compositions with the manual update policy will not be affected and the ones with the automatic update policy will start to refer to the latest one. However, there will be no practical effects on them. Do you still consider this as a breaking change?

negz:
> Do you still consider this as a breaking change?

A little, but breaking changes are acceptable because this is an alpha feature.

MisterMX:
> That's correct. We have a higher chance of having collisions. We may increase the hash suffix length to reduce it (e.g., git uses the first 7 chars of the commit ID) but there is no guaranteed way of preventing it.

Considering that the collision chance with 7 characters is already __very__ low I think it is safe to go forward with it. But aside from that: What prevents us from using the full hash as part of the object name?

ezgidemirel:
Just after sending the comment above, I started to test sha265 algorithm and hit this limit :) 

```
1.667403221663029e+09   DEBUG   crossplane.events       Warning {"object": {"kind":"Composition","name":"eks.multik8s.platformref.crossplane.io","uid":"25a6f932-6175-4a02-8d4f-01c3cc1cd480","apiVersion":"apiextensions.crossplane.io/v1","resourceVersion":"221328"}, "reason": "CreateRevision", "message": "CompositionRevision.apiextensions.crossplane.io \"eks.multik8s.platformref.crossplane.io-39d4a8f\" is invalid: metadata.labels: Invalid value: \"39d4a8fc7c6d913af07eb652a1ae3e5d1dc8796d5fe8c724c809a695d4ae5943\": must be no more than 63 characters"}

```

turkenh (approved):
Looking good to me. 
@negz are you ok with merging this or do you have further concerns?

negz (approved):

Issue #169, opened 2018-11-28
Validation for dynamic provider allowed values


jbw976:
A lot of resources have a set of allowed values by the cloud provider that will change over time, such as the allowed machine types, or the list of regions where a service is available.  We can hard code these lists into the controller code and the CRD schema validation, but those lists will become stale over time as the providers make updates.

It would be ideal to fetch these allowed values at runtime so we do not need to make any code changes when the allowed values change.  The best user experience does validation through CRD schema validation, which is performed up front during CRD create.  If the user's value is no good then they get an error at create time as opposed to later on downstream.  This could be accomplished by:
* webhook validation for CRDs
* controllers update the CRD with schema validation at runtime after fetching the allowed values from the provider (is this possible to update schema validation?)

Not all provider values have an API to fetch the list of allowed values, but we should do this for any that do.

Originally discussed in https://github.com/crossplaneio/crossplane/pull/150/files#r236804934

jbw976:
Another great example for how this type of functionality could improve the user experience is the supported Kubernetes versions for Azure AKS clusters.  Those change over time, and if a resource class is specifying a version that is no longer supported, the AKS cluster will go to the following failed condition:

```
  status:
    Conditions:
    - LastTransitionTime: 2019-02-20T03:41:43Z
      Message: 'failed to start create operation for AKS cluster aks-2dad6b7d-34c1-11e9-935d-08002745fceb:
        containerservice.ManagedClustersClient#CreateOrUpdate: Failure sending request:
        StatusCode=400 -- Original Error: Code="InvalidParameter" Message="The value
        of parameter orchestratorProfile.OrchestratorVersion is invalid." Target="orchestratorProfile.OrchestratorVersion"'
      Reason: failed to create cluster
      Status: "True"
      Type: Failed
```
It would be much better if we could look up what the allowed versions are and provide earlier admissions based validation, e.g. when the resource class is first being created.

For AKS, I think we could fetch the allowed Kubernetes versions through the [List Orchestrators API](https://docs.microsoft.com/en-us/rest/api/container-service/container%20service%20client/listorchestrators).  Yes, that is for Container Service resources, but I think it's also shared with AKS Managed Clusters resources too.

muvaf:
[...]

PR #3396, opened 2022-10-19
Add AWS,GCP,Azure quickstarts and adjust section weights

plumbis:
Signed-off-by: Pete Lumbis <pete@upbound.io>

Contributes quickstart guides for AWS, GCP and Azure.

Each quickstart walks through:
* installing Crossplane via Helm or via upstream distribution
* installing the provider
* Creating cloud secrets and creating a providerconfig
* creating and verifying a simple managed resource

This also adjusts the weights of the existing "getting started guides" to come after these quick start guides.

The file `docs/getting-started/install-crossplane.md` is a "common include" to document installing Crossplane for any provider. This file is hidden from the table of contents with `toc_hide: true`. It is imported into a file with the `include` shortcode introduced via [site PR #169](https://github.com/crossplane/crossplane.github.io/pull/169).

### DIFF
--- a/docs/contributing/docs.md
+++ b/docs/contributing/docs.md
@@ -17,6 +17,9 @@ weight: 2000
     - [Linking between docs pages](#linking-between-docs-pages)
     - [Linking to external sites](#linking-to-external-sites)
   - [Tabs](#tabs)
+  - [Code blocks](#code-blocks)
+    - [Hover to highlight](#hover-to-highlight)
+    - [Copying code](#copying-code)
 
 ## Code of Conduct
 We follow the [CNCF Code of Conduct](https://github.com/cncf/foundation/blob/main/code-of-conduct.md).
@@ -184,4 +187,73 @@ A second example tab.
 
 
 Both `tab` and `tabs` require opening and closing tags. Unclosed tags causes Hugo to fail.
+
+### Code blocks
+Hugo supports standard markdown code fences with ` ``` `
+
+Specify the language of the code block after the code fence with no space, for example ` ```yaml `.
+
+{{< hint type="important" >}}
+Not providing a language renders no line numbers or styling.
+{{< /hint >}}
+
+Hugo has a [full list](https://gohugo.io/content-management/syntax-highlighting/#list-of-chroma-highlighting-languages) of supported languages.
+
+The language hint determines the styling and may not match the exact code represented.
+For example, using a `yaml` highlight hint provides greater readability for a block that is technically a `bash` command.
+
+Using the `yaml` hint:
+```yaml
+cat <<EOF | kubectl apply -f -
+apiVersion: aws.upbound.io/v1beta1
+kind: ProviderConfig
+metadata:
+  name: default
+```
+
+Using the `bash` hint:
+```bash
+cat <<EOF | kubectl apply -f -
+apiVersion: aws.upbound.io/v1beta1
+kind: ProviderConfig
+metadata:
+  name: default
+```
+
+
+#### Hover to highlight
+{{< hint type="warning" >}}
+This is a future capability for the Crossplane documentation and isn't supported today. You can add the `hover` tag and shortcode today to automatically enable highlighting when supported.
+{{< /hint >}}
+A single line in a code block can be highlighted on mouseover of another command with the `{{</* hover */>}}` shortcode.
+
+A command inside the shortcode highlights lines in an associated code block.
+
+To enable hover to highlight, first apply a `{label="<label>"}` tag on a code block.
+
+````markdown
+```shell {label="an_example"}
+echo hello
+hello
+```
+````
+
+Outside of the code block, reference the `label` and line number to provide highlighting.
+
+For example:
+````markdown
+```shell {label="an_example"}
+echo hello
+hello
+```
+````
+Hover over <code>&#123;&#123;< hover label="an_example" line="2">&#125;&#125;</code> hello <code>&#123;&#123;< hover >&#125;&#125;</code> to see it in action.
+
+The command must include a closing shortcode of <code>&#123;&#123;< hover >&#125;&#125;</code>.  
+
+#### Copying code
+{{< hint type="warning" >}}
+This is a future capability for the Crossplane documentation and isn't supported today. You can add the `copy-lines` tag today to automatically enable copying when supported.
+{{< /hint >}}
+By default the copy button copies the first line of the code block. Attach a `copy-lines=<start>-<end>` tag to the code block.
 
--- a/docs/getting-started/install-configure.md
+++ b/docs/getting-started/install-configure.md
@@ -17,78 +17,11 @@ in production environments.
 
 {{% tab "Crossplane (upstream)" %}}
 
-## Start with Upstream Crossplane
+{{< include file="install-crossplane.md" type="page" >}}
 
-Installing Crossplane into an existing Kubernetes cluster will require a bit
-more setup, but can provide more flexibility for users who need it.
+### Install a pre-release version of Crossplane
 
-### Get a Kubernetes Cluster
-<!-- inside Crossplane (upstream) -->
-{{% tabs "Kubernetes Clusters" %}}
-
-{{% tab "macOS via Homebrew" %}}
-
-For macOS via Homebrew use the following:
-
-```bash
-brew upgrade
-brew install kind
-brew install kubectl
-brew install helm
-kind create cluster --image kindest/node:v1.23.0 --wait 5m
-```
-<!-- close "macOS via Homebrew" -->
-{{% /tab  %}}
-
-{{% tab "macOS / Linux" %}}
-
-For macOS / Linux use the following:
-
-* [Kubernetes cluster](https://kubernetes.io/docs/setup/)
-* [Kind](https://kind.sigs.k8s.io/docs/user/quick-start/)
-* [Minikube](https://minikube.sigs.k8s.io/docs/start/), minimum version `v0.28+`
-* etc.
-* [Helm](https://helm.sh/docs/intro/using_helm/), minimum version `v3.0.0+`.
-
-<!-- close "macOS / Linux" -->
-{{% /tab %}}
-
-{{% tab "Windows" %}}
-For Windows use the following:
-
-* [Kubernetes cluster](https://kubernetes.io/docs/setup/)
-* [Kind](https://kind.sigs.k8s.io/docs/user/quick-start/)
-* [Minikube](https://minikube.sigs.k8s.io/docs/start/), minimum version `v0.28+`
-* etc.
-* [Helm](https://helm.sh/docs/intro/using_helm/), minimum version `v3.0.0+`.
-
-<!-- close "Windows" -->
-{{% /tab %}}
-
-<!-- close "Kubernetes Clusters" -->
-{{% /tabs %}}
-
-### Install Crossplane
-
-{{% tabs "install with helm" %}}
-
-{{% tab "Helm 3 (stable)" %}}
-Use Helm 3 to install the latest official `stable` release of Crossplane, suitable for community use and testing:
-
-```bash
-kubectl create namespace crossplane-system
-helm repo add crossplane-stable https://charts.crossplane.io/stable
-helm repo update
-
-helm install crossplane --namespace crossplane-system crossplane-stable/crossplane
-```
-
-<!-- close "Helm 3 (stable)" -->
-{{% /tab %}}
-
-{{% tab "Helm 3 (latest)" %}}
-<!-- fold start -->
-Use Helm 3 to install the latest pre-release version of Crossplane:
+To install a pre-release version of Crossplane use `--devel` with `helm install`.
 
 ```bash
 kubectl create namespace crossplane-system
@@ -107,11 +40,6 @@ For example:
 helm install crossplane --namespace crossplane-system crossplane-master/crossplane \
   --version 0.11.0-rc.100.gbc5d311 --devel
 ```
-<!-- close "Helm 3 (latest)" -->
-{{% /tab %}}
-<!-- close "install with helm" -->
-{{% /tabs %}}
-
 ### Check Crossplane Status
 
 ```bash
--- a/docs/getting-started/install-configure.md
+++ b/docs/getting-started/install-configure.md
@@ -1,6 +1,6 @@
 ---
 title: Install & Configure
-weight: 2
+weight: 10
 ---
 ## Choosing Your Crossplane Distribution
 
--- a/docs/getting-started/install-crossplane.md
+++ b/docs/getting-started/install-crossplane.md
@@ -6,21 +6,13 @@ toc_hide: true
 
 Crossplane installs into an existing Kubernetes cluster. 
 
-Install Crossplane either as an upstream open source distribution or from a [CNCF certified](https://github.com/cncf/crossplane-conformance) vendor distribution.
-
-The Crossplane community supports the upstream distribution.  
-The Crossplane vendor supports their specific Crossplane distribution. 
-
 {{< hint type="tip" >}}
 If you don't have a Kubernetes cluster create one locally with [minikube](https://minikube.sigs.k8s.io/docs/start/) or [kind](https://kind.sigs.k8s.io/).
 {{< /hint >}}
 
-{{< tabs "distros" >}}
-
-{{< tab "Community Supported" >}}
 ### Install the Crossplane Helm chart
 
-Helm enables Crossplane to install all its Kubernetes components through a single _Helm Chart_.
+Helm enables Crossplane to install all its Kubernetes components through a _Helm Chart_.
 
 Enable the Crossplane Helm Chart repository:
 
@@ -34,17 +26,19 @@ Helm supports a `dry-run` to see the changes Helm makes. Run the Helm dry-run to
 
 ```shell
 helm install crossplane \
+crossplane-stable/crossplane \
 --dry-run --debug \
 --namespace crossplane-system \
-crossplane-stable/crossplane
+--create-namespace
 ```
 
 Last, install the Crossplane components using `helm install`.
 
 ```shell
 helm install crossplane \
+crossplane-stable/crossplane \
 --namespace crossplane-system \
-crossplane-stable/crossplane
+--create-namespace
 
 NAME: crossplane
 LAST DEPLOYED: Sun Oct 16 19:11:27 2022
@@ -66,56 +60,13 @@ Kube Version: v1.25.2
 Verify Crossplane installed with `kubectl get pods`.
 
 ```shell
-kubectl get pods -n crossplane-system                                                                  
+kubectl get pods -n crossplane-system
 NAME                                      READY   STATUS    RESTARTS   AGE
 crossplane-d4cd8d784-ldcgb                1/1     Running   0          54s
 crossplane-rbac-manager-84769b574-6mw6f   1/1     Running   0          54s
 ```
 
-{{< /tab >}}
-
-{{< tab "Vendor Distributions" >}}
-
-## Upbound
-Upbound are the founders of Crossplane. Upbound maintains and supports an open source distribution of Crossplane called [Upbound Universal Crossplane](https://github.com/upbound/universal-crossplane) (_UXP_).
-
-More details about installing and configuring UXP are available in the [Upbound documentation](https://docs.upbound.io)
-
-### Install the Up command-line
-The Up command-line helps manage UXP and manage Upbound user accounts. 
-
-Download and install the Upbound `up` command-line.
-
-```shell {copy-lines="all"}
-curl -sL "https://cli.upbound.io" | sh
-sudo mv up /usr/local/bin/
-```
-
-### Install Upbound Universal Crossplane
-_UXP_ consists of upstream Crossplane and Upbound-specific enhancements and patches. It's [open source](https://github.com/upbound/universal-crossplane) and maintained by Upbound. 
-
-Install UXP with the Up command-line `up uxp install` command.
-
-```shell
-up uxp install
-UXP 1.9.1-up.2 installed
-```
-
-Verify all UXP pods are `Running` with `kubectl get pods -n upbound-system`. This may take up to five minutes depending on your Kubernetes cluster.
-
-```shell {label="pods"}
-kubectl get pods -n upbound-system
-NAME                                        READY   STATUS    RESTARTS      AGE
-crossplane-7fdfbd897c-pmrml                 1/1     Running   0             68m
-crossplane-rbac-manager-7d6867bc4d-v7wpb    1/1     Running   0             68m
-upbound-bootstrapper-5f47977d54-t8kvk       1/1     Running   0             68m
-xgql-7c4b74c458-5bf2q                       1/1     Running   3 (67m ago)   68m
-```
-{{< /tab >}}
-
-{{< /tabs >}}
-
-Installing Crossplane creates new Kubernetes API end-points. Take a look at the new API end-points with `kubectl api-resources  | grep crossplane`. In a later step you use the {{< hover label="grep" line="10">}}Provider{{< /hover >}} resource install the Official Provider.
+Installing Crossplane creates new Kubernetes API end-points. Look at the new API end-points with `kubectl api-resources  | grep crossplane`. In a later step you use the {{< hover label="grep" line="10">}}Provider{{< /hover >}} resource install the Official Provider.
 
 ```shell  {label="grep"}
 kubectl api-resources  | grep crossplane
--- a/docs/getting-started/install-crossplane.md
+++ b/docs/getting-started/install-crossplane.md
@@ -74,7 +74,7 @@ crossplane-rbac-manager-84769b574-6mw6f   1/1     Running   0          54s
 
 {{< /tab >}}
 
-{{< tab "Vendor distributions" >}}
+{{< tab "Vendor Distributions" >}}
 
 ## Upbound
 Upbound are the founders of Crossplane. Upbound maintains and supports an open source distribution of Crossplane called [Upbound Universal Crossplane](https://github.com/upbound/universal-crossplane) (_UXP_).
--- a/docs/getting-started/install-crossplane.md
+++ b/docs/getting-started/install-crossplane.md
@@ -1,0 +1,132 @@
+---
+toc_hide: true
+---
+
+## Install Crossplane
+
+Crossplane installs into an existing Kubernetes cluster. 
+
+Install Crossplane either as an upstream open source distribution or from a [CNCF certified](https://github.com/cncf/crossplane-conformance) vendor distribution.
+
+The Crossplane community supports the upstream distribution.  
+The Crossplane vendor supports their specific Crossplane distribution. 
+
+{{< hint type="tip" >}}
+If you don't have a Kubernetes cluster create one locally with [minikube](https://minikube.sigs.k8s.io/docs/start/) or [kind](https://kind.sigs.k8s.io/).
+{{< /hint >}}
+
+{{< tabs "distros" >}}
+
+{{< tab "Community Supported" >}}
+### Install the Crossplane Helm chart
+
+Helm enables Crossplane to install all its Kubernetes components through a single _Helm Chart_.
+
+Enable the Crossplane Helm Chart repository:
+
+```shell
+helm repo add \
+crossplane-stable https://charts.crossplane.io/stable
+helm repo update
+```
+
+Helm supports a `dry-run` to see the changes Helm makes. Run the Helm dry-run to see all the Crossplane components.
+
+```shell
+helm install crossplane \
+--dry-run --debug \
+--namespace crossplane-system \
+crossplane-stable/crossplane
+```
+
+Last, install the Crossplane components using `helm install`.
+
+```shell
+helm install crossplane \
+--namespace crossplane-system \
+crossplane-stable/crossplane
+
+NAME: crossplane
+LAST DEPLOYED: Sun Oct 16 19:11:27 2022
+NAMESPACE: crossplane-system
+STATUS: deployed
+REVISION: 1
+TEST SUITE: None
+NOTES:
+Release: crossplane
+
+Chart Name: crossplane
+Chart Description: Crossplane is an open source Kubernetes add-on that enables platform teams to assemble infrastructure from multiple vendors, and expose higher level self-service APIs for application teams to consume.
+Chart Version: 1.9.1
+Chart Application Version: 1.9.1
+
+Kube Version: v1.25.2
+```
+
+Verify Crossplane installed with `kubectl get pods`.
+
+```shell
+kubectl get pods -n crossplane-system                                                                  
+NAME                                      READY   STATUS    RESTARTS   AGE
+crossplane-d4cd8d784-ldcgb                1/1     Running   0          54s
+crossplane-rbac-manager-84769b574-6mw6f   1/1     Running   0          54s
+```
+
+{{< /tab >}}
+
+{{< tab "Vendor distributions" >}}
+
+## Upbound
+Upbound are the founders of Crossplane. Upbound maintains and supports an open source distribution of Crossplane called [Upbound Universal Crossplane](https://github.com/upbound/universal-crossplane) (_UXP_).
+
+More details about installing and configuring UXP are available in the [Upbound documentation](https://docs.upbound.io)
+
+### Install the Up command-line
+The Up command-line helps manage UXP and manage Upbound user accounts. 
+
+Download and install the Upbound `up` command-line.
+
+```shell {copy-lines="all"}
+curl -sL "https://cli.upbound.io" | sh
+sudo mv up /usr/local/bin/
+```
+
+### Install Upbound Universal Crossplane
+_UXP_ consists of upstream Crossplane and Upbound-specific enhancements and patches. It's [open source](https://github.com/upbound/universal-crossplane) and maintained by Upbound. 
+
+Install UXP with the Up command-line `up uxp install` command.
+
+```shell
+up uxp install
+UXP 1.9.1-up.2 installed
+```
+
+Verify all UXP pods are `Running` with `kubectl get pods -n upbound-system`. This may take up to five minutes depending on your Kubernetes cluster.
+
+```shell {label="pods"}
+kubectl get pods -n upbound-system
+NAME                                        READY   STATUS    RESTARTS      AGE
+crossplane-7fdfbd897c-pmrml                 1/1     Running   0             68m
+crossplane-rbac-manager-7d6867bc4d-v7wpb    1/1     Running   0             68m
+upbound-bootstrapper-5f47977d54-t8kvk       1/1     Running   0             68m
+xgql-7c4b74c458-5bf2q                       1/1     Running   3 (67m ago)   68m
+```
+{{< /tab >}}
+
+{{< /tabs >}}
+
+Installing Crossplane creates new Kubernetes API end-points. Take a look at the new API end-points with `kubectl api-resources  | grep crossplane`. In a later step you use the {{< hover label="grep" line="10">}}Provider{{< /hover >}} resource install the Official Provider.
+
+```shell  {label="grep"}
+kubectl api-resources  | grep crossplane
+compositeresourcedefinitions      xrd,xrds     apiextensions.crossplane.io/v1         false        CompositeResourceDefinition
+compositionrevisions                           apiextensions.crossplane.io/v1alpha1   false        CompositionRevision
+compositions                                   apiextensions.crossplane.io/v1         false        Composition
+configurationrevisions                         pkg.crossplane.io/v1                   false        ConfigurationRevision
+configurations                                 pkg.crossplane.io/v1                   false        Configuration
+controllerconfigs                              pkg.crossplane.io/v1alpha1             false        ControllerConfig
+locks                                          pkg.crossplane.io/v1beta1              false        Lock
+providerrevisions                              pkg.crossplane.io/v1                   false        ProviderRevision
+providers                                      pkg.crossplane.io/v1                   false        Provider
+storeconfigs                                   secrets.crossplane.io/v1alpha1         false        StoreConfig
+```
--- a/docs/getting-started/install-crossplane.md
+++ b/docs/getting-started/install-crossplane.md
@@ -7,7 +7,7 @@ toc_hide: true
 Crossplane installs into an existing Kubernetes cluster. 
 
 {{< hint type="tip" >}}
-If you don't have a Kubernetes cluster create one locally with [minikube](https://minikube.sigs.k8s.io/docs/start/) or [kind](https://kind.sigs.k8s.io/).
+If you don't have a Kubernetes cluster create one locally with [kind](https://kind.sigs.k8s.io/).
 {{< /hint >}}
 
 ### Install the Crossplane Helm chart
--- a/docs/getting-started/provider-aws.md
+++ b/docs/getting-started/provider-aws.md
@@ -1,0 +1,226 @@
+---
+title: AWS Quickstart
+weight: 2
+---
+
+Connect Crossplane to AWS to create and manage cloud resources from Kubernetes with the [AWS Official Provider](https://marketplace.upbound.io/providers/upbound/provider-aws).
+
+This guide walks you through the steps required to get started with the AWS Official Provider. This includes installing Crossplane, configuring the provider to authenticate to AWS and creating a _Managed Resource_ in AWS directly from your Kubernetes cluster.
+
+- [Prerequisites](#prerequisites)
+  - [Install the official AWS provider](#install-the-official-aws-provider)
+  - [Create a Kubernetes secret for AWS](#create-a-kubernetes-secret-for-aws)
+    - [Generate an AWS key-pair file](#generate-an-aws-key-pair-file)
+    - [Create a Kubernetes secret with the AWS credentials](#create-a-kubernetes-secret-with-the-aws-credentials)
+  - [Create a ProviderConfig](#create-a-providerconfig)
+  - [Create a managed resource](#create-a-managed-resource)
+  - [Delete the managed resource](#delete-the-managed-resource)
+- [Next steps](#next-steps)
+
+## Prerequisites
+This quickstart requires:
+* a Kubernetes cluster with at least 3 GB of RAM
+* permissions to create pods and secrets in the Kubernetes cluster
+* [Helm] version `v3.0.0` or later
+* an AWS account with permissions to create an S3 storage bucket
+* AWS [access keys](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-quickstart.html#cli-configure-quickstart-creds)
+
+{{< hint type="tip" >}}
+If you don't have a Kubernetes cluster create one locally with [minikube](https://minikube.sigs.k8s.io/docs/start/) or [kind](https://kind.sigs.k8s.io/).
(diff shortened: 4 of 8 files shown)


docs/getting-started/install-configure.md
```diff
@@ -1,6 +1,6 @@
 ---
 title: Install & Configure
```
hasheddan, line 2: Much of this document does not make as much sense with all of the provider guides preceding it. I'm not sure what our flow is now for a quick start. Are we expecting that folks install providers, provision infrastructure, delete infrastructure, uninstall providers, then install configurations? Or perhaps more likely, install providers, provision infrastructure, install configuration, provision more infrastructure but using composition? At that point it feels like we are getting a bit beyond a quickstart. The reason for the current structure is that it show off all features of Crossplane with relatively few commands. The danger of starting with just using providers is that users may drop off afterwards and never really get to understanding composition, etc. Counterpoint to that would be that it doesn't matter because they don't understand what is actually happening when they see the quickstart today, but it would be awesome if we could bridge the gap of:
1. keeping the quickstart minimal
2. showing off the full capabilities of Crossplane
3. not overwhelming users with concepts

Right now I feel like we have (1) and (2), and this PR would kind of take us to some version of (2) and (3).

docs/getting-started/install-crossplane.md
```diff
@@ -0,0 +1,132 @@
...
+kubectl get pods -n crossplane-system                                                                  
+NAME                                      READY   STATUS    RESTARTS   AGE
+crossplane-d4cd8d784-ldcgb                1/1     Running   0          54s
+crossplane-rbac-manager-84769b574-6mw6f   1/1     Running   0          54s
+```
+
+{{< /tab >}}
+
+{{< tab "Vendor distributions" >}}
+
```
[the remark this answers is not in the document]
plumbis: I would go the other way and say we should remove it from the other guide and centralize it into a section about other vendors or providers being added to the docs. This would tie into some of the changes in #3398

docs/getting-started/provider-aws.md
```diff
@@ -0,0 +1,226 @@
+---
+title: AWS Quickstart
+weight: 2
+---
+
+Connect Crossplane to AWS to create and manage cloud resources from Kubernetes with the [AWS Official Provider](https://marketplace.upbound.io/providers/upbound/provider-aws).
```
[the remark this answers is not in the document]
plumbis: Providers are outside of the scope of the charter but they are inextricably linked to the usage of Crossplane. As a result discussion of using providers becomes a requirement for user documentation. I would encourage anyone in the community or any provider author to build and contribute a quickstart for their provider. If/when we need guidelines around the docs contributor's guide would be a great home for it.

Regarding the maintenance burden, these guides should be version independent (or pin to a minimum version). The idea isn't to change a quickstart with every release. The goal is to show basic functionality. If the number of contributed provider quick starts grows too large, that's a good problem to have and we can solve it then.  For now, I believe it is in the best interest of the community to provide the quickstart guides directly in the crossplane docs. If a doc author wishes to link to marketplace.upbound.io or any other external resource, that is fine and up to the author.

docs/getting-started/install-configure.md
```diff
@@ -1,6 +1,6 @@
...
-weight: 2
+weight: 10
 ---
 ## Choosing Your Crossplane Distribution
```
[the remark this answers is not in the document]
plumbis: No. The goal of this PR is limited only to contributing the quickstart guides and changing the doc orders. I explicitly to not want to modify the existing content in this PR. They would be modified as part of #3398 

I understand this creates an inconsistency but we have to decide if we are okay with minor inconsistencies like this with a short-term goal of correcting them or do we want to have very large content reorg/rewrite PRs to address all of them at once?

docs/getting-started/install-configure.md
plumbis: This came up today in the community call, but I agree. 

Some of this is in #3398, specifically-

>Quick start vs Reference Implementations
>Quickstarts are specifically designed for people with little or no knowledge of Crossplane. The goal of Quickstart guides is not to show the full power of Crossplane but just to demonstrate how Crossplane connects a Kubernetes cluster to external resources.
>
>Reference Implementations are designed to be a blueprint for "how to do it right". This should be considered a (mostly) comprehensive reference for users that that have experience with Crossplane and are looking for practical examples on how to implement more advanced Crossplane features.

In the Oct 20 Community Call we discussed different goals for what you're describing based on personas:
**1.) A user completely new to Crossplane. May have limited Kubernetes knowledge.** 
This is the target persona of a provider quickstart. How do you demonstrate the core concept of Crossplane communicating with an external API endpoint, entirely through Kubernetes. The quickstart (should) also demonstrates a validation of security credentials between Crossplane and the Provider.

**2.) A user who is familiar with Crossplane or self-service infrastructure APIs** 
This persona is more interested in a more comprehensive example and "magic moment". The discussion today was around creating a new "Composition and Claim" quickstart delivered in a simple fashion (TBD, likely a config package). The user would then apply a small number of configurations or changes to finish the deployment and consume the resources. The main thing here is fast forwarding through all of the lower-level configurations and requirements (e.g., Provider) to focus on full value of Crossplane.

**3.) An end-to-end example using one or more providers**
This would be the "Reference Implementations' mentioned in #3398. These would be "production grade" ways to use Crossplane for a given provider (or set of providers). These should be considered "common/best practices" way to do things. 

Regarding the existing "Getting Started" examples I believe they should be pulled into the Concepts/Components sections to show the configuration and linkage of resources. It's a bit hard to explain this without just building the content

plumbis:
In code review I realized the `include` shortcode is pointing to the wrong file location (a result of the disconnect between content and presentation). This is fixed in the most recent push ([a65bde5](https://github.com/crossplane/crossplane/pull/3396/commits/a65bde53b1a2791f9320c75de6457ae1be674466))


Website [issue #170](https://github.com/crossplane/crossplane.github.io/issues/170) was filed to prevent it in the future.

docs/getting-started/install-crossplane.md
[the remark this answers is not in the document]
hasheddan: > I would go the other way and say we should remove it from the other guide and centralize it into a section about other vendors or providers being added to the docs.

I'm indifferent as long as it is present and consistent.

plumbis:
A new commit pulls Upbound and `up` CLI related info.

Based on conversations this patch set uses the following criteria:
* Crossplane does not have a "favored" provider. The use of the Upbound providers is a consequence of the contributor (me). Crossplane docs should/will welcome similar guides from other providers.
* Outside of picking a provider for demonstration purposes, these guides attempt to be vendor (i.e., Upbound) independent. 
* This introduces inconsistencies with the existing _Getting Started_ guides in the `master` branch. There is an immediate goal of resolving this after #3398 is settled. [note: `backport` tags were removed]


Tagging @hasheddan to resolve outstanding conversations if he feels they we have agreement.

docs/getting-started/install-crossplane.md
```diff
@@ -0,0 +1,83 @@
...
+Kube Version: v1.25.2
+```
+
+Verify Crossplane installed with `kubectl get pods`.
+
+```shell
+kubectl get pods -n crossplane-system
+NAME                                      READY   STATUS    RESTARTS   AGE
+crossplane-d4cd8d784-ldcgb                1/1     Running   0          54s
+crossplane-rbac-manager-84769b574-6mw6f   1/1     Running   0          54s
```
sftim, line 66: This is not valid Bourne shell.

docs/getting-started/provider-aws.md
```diff
@@ -0,0 +1,219 @@
...
+{{< /hint >}}
+
+
+{{< hint type="note" >}}
+All commands use the current `kubeconfig` context and configuration. 
+{{< /hint >}}
+
+{{< include file="/docs/master/getting-started/install-crossplane.md" type="page" >}}
+
+### Install the AWS provider
```
sftim, line 39: This shouldn't really be a subheading of prerequisites.

docs/getting-started/provider-aws.md
```diff
@@ -0,0 +1,219 @@
...
+  credentials:
+    source: Secret
+    secretRef:
+      namespace: crossplane-system
+      name: aws-secret
+      key: creds
+EOF
+```
+
+This attaches the AWS credentials, saved as a Kubernetes secret, as a {{< hover label="providerconfig" line="9">}}secretRef{{</ hover>}}.
```
sftim, line 158: Is that good practice? Any viable alternative?

docs/getting-started/provider-aws.md
```diff
@@ -0,0 +1,219 @@
...
+  name: $bucket
+spec:
+  forProvider:
+    region: us-east-1
+  providerConfigRef:
+    name: default
+EOF
+```
+
+Notice the {{< hover label="xr" line="3">}}apiVersion{{< /hover >}} and {{< hover label="xr" line="4">}}kind{{</hover >}} are from the `Provider's` CRDs.
```
sftim, line 186:
```suggestion
Notice the {{< hover label="xr" line="3">}}apiVersion{{< /hover >}} and {{< hover label="xr" line="4">}}kind{{</hover >}} are from the provider's CRDs.
```

docs/getting-started/provider-aws.md
```diff
@@ -0,0 +1,219 @@
...
+
+{{< hint type="tip" >}}
+Crossplane created the bucket when the values `READY` and `SYNCED` are `True`.  
+This may take up to 5 minutes.  
+{{< /hint >}}
+
+```shell
+kubectl get buckets
+NAME                          READY   SYNCED   EXTERNAL-NAME                 AGE
+crossplane-bucket-45eed4ae0   True    True     crossplane-bucket-45eed4ae0   61s
```
sftim, line 204: This isn't valid Bourne shell code.

docs/getting-started/provider-aws.md
```diff
@@ -0,0 +1,219 @@
...
+```
+
+### Delete the managed resource
+Before shutting down your Kubernetes cluster, delete the S3 bucket just created.
+
+Use `kubectl delete bucket <bucketname>` to remove the bucket.
+
+```shell
+kubectl delete bucket $bucket
+bucket.s3.aws.upbound.io "crossplane-bucket-45eed4ae0" deleted
```
sftim, line 214: This isn't valid Bourne shell code.

docs/getting-started/provider-azure.md
```diff
@@ -0,0 +1,216 @@
...
+{{< /hint >}}
+
+
+{{< hint type="note" >}}
+All commands use the current `kubeconfig` context and configuration. 
+{{< /hint >}}
+
+{{< include file="/docs/master/getting-started/install-crossplane.md" type="page" >}}
+
+### Install the Azure provider
```
sftim, line 39: This shouldn't really be a subheading of prerequisites.

docs/getting-started/provider-azure.md
```diff
@@ -0,0 +1,216 @@
...
+
+{{< hint type="tip" >}}
+All the supported CRDs are also available in the [Upbound Marketplace](https://marketplace.upbound.io/providers/upbound/provider-azure/latest/crds).
+{{< /hint >}}
+
+
+### Create a Kubernetes secret for Azure
+The provider requires credentials to create and manage Azure resources. Providers use a Kubernetes _Secret_ to connect the credentials to the provider.
+
+First generate a Kubernetes _Secret_ from your Azure JSON file and then configure the Provider to use it.
```
sftim, line 80: Is there an alternative, such as using a service identity?

docs/getting-started/provider-azure.md
```diff
@@ -0,0 +1,216 @@
...
+Name:         azure-secret
+Namespace:    crossplane-system
+Labels:       <none>
+Annotations:  <none>
+
+Type:  Opaque
+
+Data
+====
+creds:  629 bytes
```
sftim, line 140: This isn't a valid Bourne shell script?

docs/getting-started/provider-azure.md
```diff
@@ -0,0 +1,216 @@
...
+  name: default
+kind: ProviderConfig
+spec:
+  credentials:
+    source: Secret
+    secretRef:
+      namespace: crossplane-system
+      name: azure-secret
+      key: creds
+EOF
```
sftim, line 160: This isn't valid YAML; it's a snippet of shell.

docs/getting-started/provider-azure.md
```diff
@@ -0,0 +1,216 @@
...
+apiVersion: azure.upbound.io/v1beta1
+kind: ResourceGroup
+metadata:
+  name: example-rg
+spec:
+  forProvider:
+    location: "East US"
+  providerConfigRef:
+    name: default
+EOF
```
sftim, line 186: This isn't valid YAML; it's a snippet of shell.

docs/getting-started/provider-azure.md
```diff
@@ -0,0 +1,216 @@
...
+This example uses the name `example-rg`.
+
+The {{< hover label="xr" line="8">}}spec.forProvider.location{{< /hover >}} tells Azure which Azure region to use when deploying resources. The region can be any [Azure geography](https://azure.microsoft.com/en-us/explore/global-infrastructure/geographies/) code.
+
+Use `kubectl get resourcegroup` to verify Crossplane created the resource group.
+
+```shell
+kubectl get ResourceGroup
+NAME         READY   SYNCED   EXTERNAL-NAME   AGE
+example-rg   True    True     example-rg      4m58s
```
sftim, line 201: This isn't a valid snippet of Bourne shell.

docs/getting-started/provider-azure.md
```diff
@@ -0,0 +1,216 @@
...
+```
+
+### Delete the managed resource
+Before shutting down your Kubernetes cluster, delete the resource group just created.
+
+Use `kubectl delete resource-group` to remove the bucket.
+
+```shell
+kubectl delete resourcegroup example-rg
+resourcegroup.azure.upbound.io "example-rg" deleted
```
sftim, line 211: This isn't a valid snippet of Bourne shell.

docs/getting-started/provider-gcp.md
```diff
@@ -0,0 +1,227 @@
...
+Verify the provider installed with `kubectl get providers`. 
+
+{{< hint type="note" >}}
+It may take up to five minutes for the provider to list `HEALTHY` as `True`. 
+{{< /hint >}}
+
+```shell 
+kubectl get providers
+NAME                   INSTALLED   HEALTHY   PACKAGE                                        AGE
+upbound-provider-gcp   True        False     xpkg.upbound.io/upbound/provider-gcp:v0.15.0   8s
```
sftim, line 67: This isn't a valid snippet of Bourne shell.

docs/getting-started/provider-gcp.md
```diff
@@ -0,0 +1,227 @@
...
+Name:         gcp-secret
+Namespace:    crossplane-system
+Labels:       <none>
+Annotations:  <none>
+
+Type:  Opaque
+
+Data
+====
+creds:  2330 bytes
```
sftim, line 131: This isn't a valid snippet of Bourne shell.

docs/getting-started/provider-gcp.md
```diff
@@ -0,0 +1,227 @@
...
+  name: default
+spec:
+  projectID: <PROJECT_ID>
+  credentials:
+    source: Secret
+    secretRef:
+      namespace: crossplane-system
+      name: gcp-secret
+      key: creds
+EOF
```
sftim, line 157: This isn't a YAML document, it's a snippet of shell.

docs/getting-started/provider-gcp.md
```diff
@@ -0,0 +1,227 @@
...
+metadata:
+  name: $bucket
+spec:
+  forProvider:
+    location: US
+    storageClass: MULTI_REGIONAL
+  providerConfigRef:
+    name: default
+  deletionPolicy: Delete
+EOF
```
sftim, line 184: This isn't a YAML document; it's a snippet of shell.

docs/getting-started/provider-gcp.md
```diff
@@ -0,0 +1,227 @@
...
+
+{{< hint type="tip" >}}
+Crossplane created the bucket when the values `READY` and `SYNCED` are `True`.  
+This may take up to 5 minutes.  
+{{< /hint >}}
+
+```shell
+kubectl get bucket
+NAME                          READY   SYNCED   EXTERNAL-NAME                 AGE
+crossplane-bucket-cf2b6d853   True    True     crossplane-bucket-cf2b6d853   3m3s
```
sftim, line 207: This isn't a valid snippet of Bourne shell.

docs/getting-started/provider-gcp.md
```diff
@@ -0,0 +1,227 @@
...
+Optionally, log into the [GCP Console](https://console.cloud.google.com/) and see the storage bucket inside GCP.
+
+### Delete the managed resource
+Before shutting down your Kubernetes cluster, delete the S3 bucket just created.
+
+Use `kubectl delete bucket <bucketname>` to remove the bucket.
+
+```shell
+kubectl delete bucket $bucket
+bucket.storage.gcp.upbound.io "crossplane-bucket-b7cf6b590" deleted
```
sftim, line 219: This isn't a valid snippet of Bourne shell.

sftim (commented):
Thanks. I hope this feedback is useful.

docs/getting-started/provider-gcp.md
```diff
@@ -0,0 +1,227 @@
...
+
+
+{{< hint type="note" >}}
+All commands use the current `kubeconfig` context and configuration. 
+{{< /hint >}}
+
+
+{{< include file="/docs/master/getting-started/install-crossplane.md" type="page" >}}
+
+### Install the GCP provider
```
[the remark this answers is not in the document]
plumbis: Great catch! fixed in [34620cb](https://github.com/crossplane/crossplane/pull/3396/commits/34620cb98530a33bc216ec2814a5ef096b364a72)

docs/getting-started/install-crossplane.md
```diff
@@ -0,0 +1,83 @@
...
+```
+
+Installing Crossplane creates new Kubernetes API end-points. Look at the new API end-points with `kubectl api-resources  | grep crossplane`. In a later step you use the {{< hover label="grep" line="10">}}Provider{{< /hover >}} resource install the Official Provider.
+
+```shell  {label="grep"}
```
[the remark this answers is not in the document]
plumbis: Is the concern that the content in the code block isn't valid? 

The ```` ```shell ```` is a hint to the syntax highlighter. This is based on the [Hugo syntax highlighter](https://gohugo.io/content-management/syntax-highlighting/).

docs/getting-started/install-crossplane.md
```diff
@@ -0,0 +1,83 @@
+---
+toc_hide: true
```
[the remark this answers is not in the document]
plumbis: This is only intended to be an included page. This means it doesn't need a title to display (since it won't be a standalone page) and it uses the `toc_hide` frontmatter to keep it out of the left-side table of contents.

docs/getting-started/install-crossplane.md
plumbis: Same as before, this is just providing hints to the syntax highlighter.

docs/getting-started/provider-aws.md
plumbis: Fixed in [4b0853a](https://github.com/crossplane/crossplane/pull/3396/commits/4b0853ac386d19f2944727f1d6b30d3924d80cd0)

docs/getting-started/provider-aws.md
plumbis: No it's not good practice and the alternatives are more complex. The goal here is "easiest way to see Crossplane do something". There are two different callouts to alternative authentication methods, the first one is [here](https://github.com/crossplane/crossplane/pull/3396/files/91bfab79cfc02c7b1b9be09a49cf97a416820c77#diff-207d19c174999fbb3184ab046d25344852149f47c145e7b78f45aae74e84a2f5R83)

docs/getting-started/provider-gcp.md
plumbis: Same as before, this is a hint to the syntax highlighter, not an indication to the actual shell usage.

docs/getting-started/provider-gcp.md
plumbis: Same as before, the ```` ```yaml ```` is a hint to the highlighter. Using YAML highlighting here is much more readable than shell. 

The hint used is not exposed in the docs.

docs/getting-started/provider-gcp.md
plumbis: Hint to the syntax highlighter.

docs/getting-started/provider-gcp.md
plumbis: Syntax highlighter hint.

docs/getting-started/provider-gcp.md
plumbis: Syntax highlighter hint.

plumbis:
> Thanks. I hope this feedback is useful.

SUPER helpful!! A lot of the code block call outs are prep for upcoming stylesheet changes that will improve code formatting and add some other interesting features like copying lines and enabling the ability to hover over lines of text and have it highlight the associated code block.
Unfortunately since those features haven't gone live yet I haven't pushed the associated docs for the docs to explain why the code is that way.

Please, keep the feedback coming!

sftim:
Some of these mismatched highlighter hints (where the content isn't a valid example of the syntax being highlighted) content look fragile to me - I'd want a strong case made before I would allow them for https://k8s.io/ 

There's too much risk of action-at-a-distance breakage for a future upgrade to the static site generator, or similar.
It's great to let people hover over something but it's problematic when the rendering comes out wrong and that confuses a reader.

docs/getting-started/provider-gcp.md
sftim: Consider putting a (hidden) comment to explain your thinking here, to help future maintainers.

docs/getting-started/provider-gcp.md
sftim: For Kubernetes docs, we [separate commands from output](https://kubernetes.io/docs/contribute/style/style-guide/#separate-commands-from-output).

docs/getting-started/install-crossplane.md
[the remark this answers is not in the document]
sftim: Maybe mark it [headless](https://gohugo.io/content-management/page-bundles/#headless-bundle)?

docs/getting-started/provider-aws.md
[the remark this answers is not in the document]
plumbis: Dan, do you feel our offline discussion resolved this?

docs/getting-started/install-configure.md
plumbis: Let me know if there are any other questions or comments on this one.

docs/getting-started/provider-gcp.md
plumbis: I'm making a different editorial choice. It's easier to see the command and output in a single place. The main motivator for separating the command from output would be to simplify copy/paste. That will be address in the Crossplane docs in the near future with a copy button that allows the author to define which lines in the output the button copies.

sftim: Seems reasonable. It'll only really be a problem if you get a large team of contributors and need to codify things for a style guide.

docs/getting-started/provider-aws.md
sftim: (feel free to mark this resolved)

docs/getting-started/provider-aws.md
sftim: (feel free to mark this resolved)

docs/getting-started/provider-azure.md
sftim: (feel free to mark this resolved)

docs/getting-started/provider-gcp.md
sftim: (feel free to mark this resolved)

docs/getting-started/provider-gcp.md
sftim: (feel free to mark this resolved)

docs/getting-started/provider-gcp.md
sftim: (feel free to mark this resolved)

docs/getting-started/provider-gcp.md
sftim: (feel free to mark this resolved)

docs/getting-started/provider-aws.md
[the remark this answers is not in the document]
sftim: If it feels useful, you could de-emphasize provider docs in the `sitemap.xml` whilst still allowing them to be crawled and indexed.

docs/getting-started/provider-gcp.md
plumbis: Agreed! With some of the upcoming style changes I am planning an large expansion to the existing [Contributor's Guide](http://localhost:1313/docs/master/contributing/docs.html). You can see some of that expansion today in the [Upbound Docs Contributor's Guide](https://docs.upbound.io/contributing).

plumbis:
> Some of these mismatched highlighter hints (where the content isn't a valid example of the syntax being highlighted) content look fragile to me - I'd want a strong case made before I would allow them for https://k8s.io/
> 
> There's too much risk of action-at-a-distance breakage for a future upgrade to the static site generator, or similar. It's great to let people hover over something but it's problematic when the rendering comes out wrong and that confuses a reader.

This is valid feedback. There is a tradeoff that has to be made when providing syntax highlighting in the documentation. It's up to the highlighting engine (in this case [Chroma](https://github.com/alecthomas/chroma) via [Hugo](https://gohugo.io/content-management/syntax-highlighting/)) to apply the proper styles. Short of forking the highlighting engine or building our own we don't have many other options to create the styling that displays the given code. For example, with today's style sheets these three outputs are generated based on the hints passed.

1.) No hints, let the highlighter decide.
Screen Shot 2022-10-25 at 12 18 55 PM

2.) Use a `shell` hint. (this would be the same for any Chroma mapped alias: bash, sh, ksh, zsh, shell)
Screen Shot 2022-10-25 at 12 19 02 PM

3.) Use a `yaml` hint. 
Screen Shot 2022-10-25 at 12 19 10 PM

I'm of the opinion that we need to provide the best visual experience for the reader, even if it creates a minor discrepancy inside the source text. 

As far as concerns around changing SSGs or themes in the future, this would be universally true with any opinions that are taken. We can either fully abstract out the documentation into pure markdown or gain additional functionality by relying on things inside the tooling. The same could be said about relying on Hugo shortcodes. This creates a strict dependency on Hugo that would have to be unwound in the future if we moved to different tooling. 

I believe that detailing these decisions and reasoning within the contributor's guide is the solution. This ensures any customizations or dependencies are captured for future authors and maintainers. I've added this with [a726d5b](https://github.com/crossplane/crossplane/pull/3396/commits/a726d5b1385f6cbd12bf7d62b2a69ae1224b1b38)

docs/getting-started/install-crossplane.md
```diff
@@ -0,0 +1,83 @@
...
+---
+
+## Install Crossplane
+
+Crossplane installs into an existing Kubernetes cluster. 
+
+{{< hint type="tip" >}}
+If you don't have a Kubernetes cluster create one locally with [minikube](https://minikube.sigs.k8s.io/docs/start/) or [kind](https://kind.sigs.k8s.io/).
```
fuzz, line 10: This looks great! My concern--undoubtedly biased from just having wasted way too much time trying to get minikube working--is that this removes the previous `kind` happy path instructions for macOS/Homebrew users that were the default. I think new users would be better served--and Crossplane would receive fewer "I can't get this working" reports (and the ones it does receive would have a variable eliminated)--by only suggesting `kind` to users who don't yet have Kubernetes set up.

Maybe something llke "If you don't already have a local Kubernetes cluster (eg minikube) or run into trouble we recommend using kind."

plumbis: I'm 100% okay with that. I'll let @hasheddan and @sftim chime in too before I make the change-

Any objection to only recommending Kind due to recent minikube issues? ([minikube issues #14505](https://github.com/kubernetes/minikube/issues/14505) and [minikube issues  #14523](https://github.com/kubernetes/minikube/issues/14523))

fuzz: I made a thing while procrastinating this afternoon. https://gist.github.com/fuzz/40d345bda5c6e52b4b0a1635520dee37

sftim: > If you don't already have a local Kubernetes cluster (eg minikube) or run into trouble we recommend using kind."

Seems fair. The Kubernetes project tries to be extra neutral on recommending / not recommending cluster setup tools; projects like Crossplane might not need to be. Now that kind doesn't require Docker it's even more reasonable to recommend kind, IMO.

hasheddan: +1 I’m fine with recommending kind as the happy path here

docs/getting-started/install-configure.md
hasheddan: @plumbis how would you feel about starting with the managed resources (as you have proposed), but making the quickstart go all the way through composition and configuration packages as it does today? Would that be too long? Perhaps if there were clearly denoted drop-off points it would be tenable?

docs/getting-started/provider-aws.md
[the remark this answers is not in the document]
hasheddan: @plumbis I believe we already chatted about this offline, but if there was a _theoretical_ cloud provider that didn't require providing credentials (or distributed temporary creds for demoing), but gave you a unique URL where you could go see resources provisioned by Crossplane, would using it as the only provider in a quickstart be tenable in your mind? It would:
- Make the quickstart seamless and assuage any concerns about putting creds in a cluster
- Avoid needing to pick providers to use in the quickstart
- Avoid the maintenance burden of using multiple providers in the quickstart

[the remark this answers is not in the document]
plumbis: Possibly. A major value of having a quickstart for a specific provider instead of something abstract is it connects something the user knows (their cloud provider) to something new (Crossplane). It shows how the the new thing connects to and modifies the existing thing.  

This is why I don't think something like `provider-noop` would be effective. 

I would be 100% open to considering the theoretical provider you're describing but it needs to make a clear "magic moment" without requiring the user to learn another new technology.

docs/getting-started/install-configure.md
plumbis: I want that as the "call to action" at the end of these guides. Now that the user has connected Crossplane to their cloud provider and shown it works, the next step is to really highlight the value of Crossplane. 

Maybe this means extending the single guide or maybe it means a second guide, I think that'll become clear once the writing starts. My gut says it's a second "Getting started with Compositions" type guide.

docs/getting-started/provider-azure.md
plumbis: Fixed.

docs/getting-started/install-configure.md
hasheddan: Gotcha 👍🏻 I'm not sure I agree, but I defer to your expertise here. I do feel a bit uneasy about merging these guides as-is, which would effectively put our docs into a "broken" state (granted that they wouldn't be showing up as the default). Do you think there is potentially some middle ground where all of the linked issue does not have to be resolved in this PR, but we could be in a state where we would feel good about folks using the `master` docs in the interim?

docs/getting-started/install-crossplane.md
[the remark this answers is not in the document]
plumbis: Reasoning has been added to the [contributing guide](https://github.com/crossplane/crossplane/pull/3396/files#diff-93426d953f03b733adbbb574ffd2717d0892773a464c0d301d1b6c9095d67b15R191). 

Resolving this conversation (unresolve if you're not satisfied with this)

docs/getting-started/install-crossplane.md
plumbis: Reasoning has been added to the [contributing guide](https://github.com/crossplane/crossplane/pull/3396/files#diff-93426d953f03b733adbbb574ffd2717d0892773a464c0d301d1b6c9095d67b15R191). 

Resolving this conversation (unresolve if you're not satisfied with this)

docs/getting-started/provider-aws.md
plumbis: Call outs have been added for all 3 guides.

docs/getting-started/provider-azure.md
plumbis: Reasoning has been added to the [contributing guide](https://github.com/crossplane/crossplane/pull/3396/files#diff-93426d953f03b733adbbb574ffd2717d0892773a464c0d301d1b6c9095d67b15R191).

Resolving this conversation (unresolve if you're not satisfied with this)

docs/getting-started/provider-azure.md
plumbis: Reasoning has been added to the [contributing guide](https://github.com/crossplane/crossplane/pull/3396/files#diff-93426d953f03b733adbbb574ffd2717d0892773a464c0d301d1b6c9095d67b15R191).

Resolving this conversation (unresolve if you're not satisfied with this)

docs/getting-started/provider-gcp.md
plumbis: Reasoning has been added to the [contributing guide](https://github.com/crossplane/crossplane/pull/3396/files#diff-93426d953f03b733adbbb574ffd2717d0892773a464c0d301d1b6c9095d67b15R191).

Resolving this conversation (unresolve if you're not satisfied with this)

docs/getting-started/provider-gcp.md
plumbis: Reasoning has been added to the [contributing guide](https://github.com/crossplane/crossplane/pull/3396/files#diff-93426d953f03b733adbbb574ffd2717d0892773a464c0d301d1b6c9095d67b15R191).

Resolving this conversation (unresolve if you're not satisfied with this)

docs/getting-started/install-configure.md
hasheddan: @plumbis so mark all existing quickstart docs as draft and enable the new ones right? Is motivation for getting this PR in prior to other docs so that folks can see where the docs are headed and provide feedback?

plumbis: The specifics I think depend on how the maintainers like yourself prefer to work:

Option 1: **Drafts**
We mark these as drafts. They aren't published on the site but live in the repo. We have this PR finalized and start working on the next set of things. Each set is a "stand alone" topic/set of pages, all marked as drafts until we're ready to flip the switch on everything with one final "release" PR. This PR would fit into this model.

Option 2: **Monolithic PR**
Get it all together, written and ready for publication. The single PR could be 10s of pages of content and anywhere between 50 and 100+ commits. A single large PR is submitted for review containing everything. This PR would likely be closed without committing and re-submitted with the larger collection of content being discussed.

Either option works for me, so I'll leave it up to you on what you'd prefer.

docs/getting-started/install-crossplane.md
negz: What do you think about keeping commands separate from output? This is something the Kubernetes docs recommend. Ideally they'd say why, but I'm guessing it's to make it clear which bits are a command you can copy/paste and which bits are output.

https://kubernetes.io/docs/contribute/style/style-guide/#separate-commands-from-output

docs/getting-started/provider-aws.md
```diff
@@ -0,0 +1,219 @@
+---
+title: AWS Quickstart
```
negz, line 2: I know this is a bit of a controversial opinion, but I question whether https://crossplane.io/docs is the right place for guides that are focused on a particular provider. I wonder if we could get folks up-to-speed with what Crossplane is all about faster by focusing on a simpler scenario that demonstrated what Crossplane was all about.

I like how the Terraform docs approach this. They have getting started guides for the major cloud providers, but before you get into provider specific details they take you through a little "deploy a Docker container" scenario - e.g.:

* https://developer.hashicorp.com/terraform/tutorials/aws-get-started starts with "What is IaC with Terraform?" which takes me to an interactive tutorial to spin up a Docker container.
* (Oddly) the second entry in https://developer.hashicorp.com/terraform/tutorials/aws-get-started is https://developer.hashicorp.com/terraform/tutorials/aws-get-started/install-cli, which again shows me how to use Terraform to spin up a Docker container.

By following these tutorials I've run some Terraform commands, seen what the syntax looks like, and got a quick win (actually deployed something) without having to even have access to a cloud account, figure out credentials, etc.

If we adopted this approach it would potentially allow us to keep https://crossplane.io/docs "provider agnostic", with provider-specific quick start guides living closer to the provider implementations (e.g. in their Git repos and/or Marketplace entries). I feel this would encourage a broader set of quick start guides (not just the "big three" cloud providers and increase the chances of provider-specific docs stay up-to-date with their providers.

To be a little more concrete, what I imagine is a getting started flow something like:

1. Install Crossplane.
2. Create an MR that orchestrates some simple, credential-free external system.
3. Show that Crossplane actually changes the external system, corrects drift, etc.
4. Create a Configuration that builds an abstraction (claim, Composition, etc) around the MR from step 2.
5. Create a claim.

Obviously the gap here would be finding/building such a simple, credential-free external system. We can't deploy a Docker container on your laptop like Terraform does, and something like https://github.com/crossplane-contrib/provider-nop doesn't illustrate how Crossplane actually changes an external system.

docs/contributing/docs.md
```diff
@@ -185,3 +188,72 @@ A second example tab.
...
+For example, using a `yaml` highlight hint provides greater readability for a block that is technically a `bash` command.
+
+Using the `yaml` hint:
+```yaml
+cat <<EOF | kubectl apply -f -
+apiVersion: aws.upbound.io/v1beta1
+kind: ProviderConfig
+metadata:
+  name: default
+```
```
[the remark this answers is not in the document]
plumbis: This is a philosophical choice. From multiple user experience sessions there was a desire to have a "copy and paste experience" for these guides. 

This is explicitly different from guides that are just showing commands or describing technology. 

The instructions in a quickstart are effectively "copy and paste this and see the next step". That's the goal here. This is not creating a global standard for presenting code snippets.

docs/contributing/docs.md
```diff
@@ -185,3 +188,72 @@ A second example tab.
...
+````
+Hover over <code>&#123;&#123;< hover label="an_example" line="2">&#125;&#125;</code> hello <code>&#123;&#123;< hover >&#125;&#125;</code> to see it in action.
+
+The command must include a closing shortcode of <code>&#123;&#123;< hover >&#125;&#125;</code>.  
+
+#### Copying code
+{{< hint type="warning" >}}
+This is a future capability for the Crossplane documentation and isn't supported today. You can add the `copy-lines` tag today to automatically enable copying when supported.
+{{< /hint >}}
+By default the copy button copies the first line of the code block. Attach a `copy-lines=<start>-<end>` tag to the code block.
```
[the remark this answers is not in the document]
plumbis: This was discussed previously [here](https://github.com/crossplane/crossplane/pull/3337#discussion_r982471312).

The tl;dr is that 80 characters for text tends to make it more difficult to read and can create unintended formatting changes since two EOL spaces in markdown is rendered as a new line. 

If a text line is too long to provide precise commentary it's likely an indication that the line should be broken up anyhow. 

Regarding github specific reviews, I don't see us moving away from per-line commentary in the lines but as we build a rich documentation this becomes less important. The review process will become even better once we move to [Netlify](https://github.com/crossplane/crossplane.github.io/issues/98) and use [Deploy Previews.](https://www.netlify.com/products/deploy-previews/)

docs/getting-started/install-configure.md
plumbis: This has two challenges:

1. It's slow. Since all docs are loosely coupled it's hard to start on patch-2 until patch-1 is approved. 
2. It will likely create large number of merge conflicts and will require a force push when the branch is merged.

I don't have opinions on (2) but I want to make sure I'm in line with how the maintainers expect code contributions.

docs/getting-started/install-crossplane.md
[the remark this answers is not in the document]
plumbis: This is covered in the docs contributing guide:
https://github.com/crossplane/crossplane/blob/master/docs/contributing/docs.md?plain=1#L97

docs/getting-started/provider-aws.md
plumbis: Agreed. Discussions with @hasheddan and @bassam the conclusion was "We need _something_ to show how Crossplane works". In this case the _something_ is a provider. 

We are using the big-three cloud providers to get a large set of users. If other provider maintainers want to contribute quickstart guides in the future we should welcome this. If we have dozens of guides then it's a good problem to have and we can solve it then. 

Regarding your "flow" this is exactly what I envision. This may be in a single large quickstart or across two. Right now my preference is across two guides so we don't have a 40-page "quick" start.

If you have additional questions about using providers in the docs it's worth a meeting between you, me and @bassam  to discuss.

docs/getting-started/install-crossplane.md
plumbis: @sftim brought up a similar point [here](https://github.com/crossplane/crossplane/pull/3396/files#discussion_r1004680374).

Copying the previous answer:
> I'm making a different editorial choice. It's easier to see the command and output in a single place. The main motivator for separating the command from output would be to simplify copy/paste. That will be address in the Crossplane docs in the near future with a copy button that allows the author to define which lines in the output the button copies.

docs/getting-started/install-configure.md
negz: > This has two challenges.

I presume you're referring to the option of using a new docs branch? If so I'm happy to fall back to a monolithic PR. If I try to think about these changes as I would code we'd aim for incremental changes if possible, but if not fall back to a large PR. I don't think we have any precedent for merging draft anything (code, docs, etc) into master.

docs/getting-started/install-crossplane.md
negz: > That will be address in the Crossplane docs in the near future with a copy button that allows the author to define which lines in the output the button copies.

Will this also visually distinguish which parts are input from output?

It would be great to track these kinds of planned improvements publicly. Perhaps an issue on https://github.com/crossplane/crossplane.github.io?

docs/contributing/docs.md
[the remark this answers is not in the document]
negz: I'm not personally seeing how one alternative here is more copy-and-paste friendly than another?

With the proposed here doc style you either:

* Copy the whole thing (here doc and all) if you want to actually run the command.
* Avoid copying the bash parts if you want to copy just the YAML.

With the current style you either:

* Copy the command (immediately following the YAML file) if you want to actually run the command.
* Copy the plain YAML if you just want to copy the YAML.

I'm very reluctant to go back to this here doc style. My worry is that mixing YAML and shell in one command may be confusing/intimidating to newcomers. Is this a concern your planned "copy button" would obviate?

docs/contributing/docs.md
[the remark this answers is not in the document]
negz: Changing away from 80 chars would have a large blast radius. If your logic is true for docs, presumably its true for every markdown file, code comment, etc that we've historically wrapped at 80 chars. I would prefer to stick to the existing convention unless there's a strong case that we should drop it for markdown for the entire project.

plumbis:
Based on offline discussions, this PR will be closed and rewritten in smaller chunks.

PR #3405, opened 2022-10-21
Set crossplane user agent on package requests

hasheddan:

Sets `crossplane/<version>` user agent header on all requests made for packages. This is more explicit than the `go-containerregistry/<version>` that is utilized currently. 

I have:

Ran against local registry and verified that user agent is set to `crossplane/v1.11.0-rc.0.27.gf447ebe1.dirty go-containerregistry/v0.9.0`.

**Update:** `go-containerregistry` currently does not allow removal of the default `go-containerregistry/<version>` `User-Agent` header, so the Crossplane one is appended for the time-being.

### DIFF
--- a/internal/transport/transport.go
+++ b/internal/transport/transport.go
@@ -1,0 +1,48 @@
+/*
+Copyright 2022 The Crossplane Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package transport
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/crossplane/crossplane/internal/version"
+)
+
+// DefaultUserAgent is the default User-Agent header that is set when making
+// HTTP requests for packages.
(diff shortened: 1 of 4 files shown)


nullable-eth (approved):
Looks good and is a useful distinction.

internal/transport/transport.go
```diff
@@ -0,0 +1,48 @@
...
+func NewUserAgent(rt http.RoundTripper, userAgent string) *UserAgent {
+	return &UserAgent{
+		rt:        rt,
+		userAgent: userAgent,
+	}
+}
+
+// RoundTrip injects a User-Agent header into every request.
+func (u *UserAgent) RoundTrip(req *http.Request) (*http.Response, error) {
+	req.Header.Set("User-Agent", u.userAgent)
```
negz, line 46: I'm guessing this will always be applied after and thus override the default user agent, right?

negz (approved):

internal/transport/transport.go
hasheddan: @negz right 👍🏻

internal/transport/transport.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/crossplane/crossplane/internal/version"
+)
+
+// DefaultUserAgent is the default User-Agent header that is set when making
+// HTTP requests for packages.
+var DefaultUserAgent = fmt.Sprintf("%s/%s", "crossplane", version.New().GetVersionString())
```
ytsarev, line 28: Just to double-check, it should also expose UXP in the form of `up` in the version like `1.9.1-up.2` is that correct?

ytsarev (approved):

internal/transport/transport.go
hasheddan: @ytsarev that will be a downstream concern 👍🏻

PR #3409, opened 2022-10-24
More granular comparison of Roles annotations during reconciliation.

jmprusi:

This PR changes how roles are reconciled by comparing only the crossplane RBAC annotations (pre fixed by `rbac.crossplane.io/`) instead of all the annotations of the Role objects.

This change will allow the RBAC namespace reconciler to play along with other controllers that set annotations on the same resources, avoiding a reconciliation loop between them. 

I have:

1. Updated unit-tests to reflect the changes.
1. Deployed the change on a local cluster (image: jmprusi/crossplane-778e92a53dbc12731015f0486c2909a6@sha256:37b493d4854ae1946d922f6fea7524270f5282e9cf4aaf90530abd5ad4437fee)
1. Added a new annotation to the `default/crossplane-view` rbac role.
2. `crossplane-rbac-manager` shows in the log that no change has been detected: `Skipped no-op RBAC Role apply`.
1. Remove crossplane-view rbac.crossplane.io annotation.
1. Role is reconcile and `crossplane-rbac-manager` shows in the log "Applied RBAC Role".

### DIFF
--- a/internal/controller/rbac/namespace/reconciler.go
+++ b/internal/controller/rbac/namespace/reconciler.go
@@ -18,6 +18,7 @@ package namespace
 
 import (
 	"context"
+	"strings"
 	"time"
 
 	"github.com/google/go-cmp/cmp"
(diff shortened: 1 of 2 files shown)


stevekuznetsov:
Another approach might be to issue patch or SSA requests for intended state, but that seems more invasive.

negz (approved):

jmprusi:
force-pushed with commit sign-off for the DCO check.

Issue #3436, opened 2022-11-02
Ensure chart deploys to release namespace


AaronME:
### What problem are you facing?
Crossplane charts currently do not have a `namespace` defined on namespaced resources. 

### How could Crossplane help solve your problem?
Update the crossplane charts with `metadata.namespace: {Release.namespace}`  to ensure namespace is set during a `helm template` operation.

PR #3437, opened 2022-11-02
Explicitly setting release namespace on resources

AaronME:

Sets explicit namespaces on helm resources.

Fixes https://github.com/crossplane/crossplane/issues/3436

I have:

Deployed charts to a local kind cluster.

### DIFF
--- a/cluster/charts/crossplane/templates/deployment.yaml
+++ b/cluster/charts/crossplane/templates/deployment.yaml
@@ -2,6 +2,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: {{ template "crossplane.name" . }}
+  namespace: {{ .Release.Namespace }}
   labels:
     app: {{ template "crossplane.name" . }}
     release: {{ .Release.Name }}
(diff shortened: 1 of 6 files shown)


cluster/charts/crossplane/templates/deployment.yaml
```diff
@@ -2,6 +2,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: {{ template "crossplane.name" . }}
+  namespace: {{ .Release.Namespace }}
```
hasheddan, line 5: @AaronME does providing it explicitly change behavior from what currently happens with `helm install -n <namespace>`?

AaronME: The expected behavior is the same for `helm install -n <namespace>`

This change makes it possible to run `helm template -n <namespace> <name> <chart>` and get manifests with explicit namespaces declared. These can then be piped through other tools before applying them to the cluster. (Some checkers have flagged the lack of a namespace as "needing attention.")

hasheddan: Nice! Thanks for the explanation 👍🏻

hasheddan (approved):

hasheddan (commented):
@AaronME looks like commit needs sign off here 👍🏻

Issue #2647, opened 2021-10-20
Is it possible for package manager to use node image cache


muvaf:
### What problem are you facing?

We're trying to make Crossplane provider installation work in an offline environment. While we can load the controller images into the node cache of `kind` cluster and use `IfNotExist` policy to make it use that, package manager has its own client that hits the registries. This results in a few things:
* Users have to deploy their own in-cluster registry.
* In-cluster local registry should have trusted CA certs for SSL https://github.com/crossplane/crossplane/issues/2318
* Need to pull & tag & push with different image URL.
* Need to override controller image URL via `ControllerConfig`

### How could Crossplane help solve your problem?

We might be able to make package manager request the image from the container runtime of the node, using the same path with `kubelet` when it starts a `Pod`. This will let admins pre-load images and users won't even have to worry that there is no connection.

I know @hasheddan had concerns about this but I'd like us to discuss the possible ways of doing it to see the trade-off and justify the UX we have to provide here.

hasheddan:
Wanted to follow up on a few assertions here:
> Users have to deploy their own in-cluster registry

This is one option, but certainly not the only one. The [Crossplane package cache](https://crossplane.io/docs/v1.5/concepts/packages.html#the-package-cache) can be backed by any volume, which means that it can be [pre-populated](https://crossplane.io/docs/v1.5/concepts/packages.html#pre-populating-the-package-cache). In our integration tests for all `Providers` today, we actually use a [`PersistentVolume` that is mapped to a local directory on the host](https://github.com/crossplane/provider-aws/blob/34522c5e73b00e73eb0bd1e7ae0d127c33ab888b/cluster/local/integration_tests.sh#L100), meaning that the _only_ required step is to build the package (`kubectl crossplane configuration/provider build`). For development purposes, you can even use something like a [`Configmap` as a the package cache](https://danielmangum.com/posts/config-map-oci-image-cache/).

> We might be able to make package manager request the image from the container runtime of the node, using the same path with kubelet when it starts a Pod

[...]

PR #3438, opened 2022-11-04
feat(chart): Allow prepopulating cache from ConfigMap

MisterMX:

This adds an option to the Crossplane Helm chart that allows to load cached images from a `ConfigMap` as described [here](https://danielmangum.com/posts/config-map-oci-image-cache/).

Related to #2647

```yaml
packageCache:
  configMap: "example-configmap"
```

I have:

Manually

### DIFF
--- a/cluster/charts/crossplane/values.yaml.tmpl
+++ b/cluster/charts/crossplane/values.yaml.tmpl
@@ -68,6 +68,7 @@ packageCache:
   medium: ""
   sizeLimit: 5Mi
   pvc: ""
+  configMap: ""
 
 resourcesRBACManager:
   limits:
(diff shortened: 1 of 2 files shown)


cluster/charts/crossplane/values.yaml.tmpl
```diff
@@ -68,6 +68,7 @@ packageCache:
   medium: ""
   sizeLimit: 5Mi
```
hasheddan, line 69: Wishing I had stuck these under an `emptyDir` entry but alas, not worth breaking folks now 😞

hasheddan (approved):
I love it -- thanks @MisterMX!

MisterMX:
@hasheddan thanks for merging this.

One think I noticed during testing is that ConfigMap is always mounted as read-only file system. This means that Crossplane will not be able fetch packages from remote because it will fail writing them to disk.

PR #3444, opened 2022-11-08
Add finalizers for rbac provider and definition

### LINKED ISSUE #3443: RBAC permissions issue in OpenShift
### What happened?
I've got a permission issue that seems to only apply to OpenShift. 
Similar issues here:
https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/434
https://github.com/spotahome/redis-operator/issues/98
https://github.com/jaegertracing/jaeger-operator/issues/461

### How can we reproduce it?
1. Run crossplane on openshift
2. Try to change the version in `Provider` and `ControllerConfig`
3. Note errors logs on `crossplane` and `crossplane-rbac-manager`: 

`DEBUG	crossplane	cannot apply ClusterRole	{"controller": "rbac/providerrevision.pkg.crossplane.io", "request": "/provider-aws-edfa6e79bd65", "uid": "77a14918-0cab-4c77-9bd1-04dc085682f2", "version": "1592953204", "name": "provider-aws-edfa6e79bd65", "role-name": "crossplane:provider:provider-aws-edfa6e79bd65:aggregate-to-edit", "role-name": "crossplane:provider:provider-aws-edfa6e79bd65:aggregate-to-view", "role-name": "crossplane:provider:provider-aws-edfa6e79bd65:system", "error": "cannot update object: clusterroles.rbac.authorization.k8s.io \"crossplane:provider:provider-aws-edfa6e79bd65:system\" is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on: , <nil>"}`

`ERROR	crossplane.controller.rbac/providerrevision.pkg.crossplane.io	Reconciler error	{"reconciler group": "pkg.crossplane.io", "reconciler kind": "ProviderRevision", "name": "provider-aws-edfa6e79bd65", "namespace": "", "error": "cannot apply ClusterRole: cannot update object: clusterroles.rbac.authorization.k8s.io \"crossplane:provider:provider-aws-edfa6e79bd65:system\" is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on: , <nil>"}`

Crossplane version: v1.10.1
$ oc version
Client Version: v4.2.0-alpha.0-1570-g3c85519
Kustomize Version: v4.5.4
Server Version: 4.10.13 (Openshift)
Kubernetes Version: v1.23.5+b463d71

el-mail:
Sets `<resource>/finalizers` on RBAC for provider and definition

Fixes #3443

I have:

- Ran e2e tests with the following KIND_CONFIG: 
```
export KIND_CONFIG="kind: Cluster
  apiVersion: kind.x-k8s.io/v1alpha4
  nodes:
  - role: control-plane
    kubeadmConfigPatches:
    - |
      kind: ClusterConfiguration
      apiServer:
          extraArgs:
            enable-admission-plugins: OwnerReferencesPermissionEnforcement"
```
- Created Docker image from this branch. (https://hub.docker.com/u/qwzar)
- I've run this image on a private Opesnfhit (v4.10.*) cluster
- I tested by creating and deleting a Provider and seeing that the controllers logged and reconciled successfully.
- I've checked Providers, Managed Resources, and CRD's on existing errors from issue https://github.com/crossplane/crossplane/issues/3443#issuecomment-1306828892

### DIFF
--- a/cluster/charts/crossplane/templates/rbac-manager-clusterrole.yaml
+++ b/cluster/charts/crossplane/templates/rbac-manager-clusterrole.yaml
@@ -25,6 +25,18 @@ rules:
   - get
   - list
   - watch
+# The RBAC manager creates a series of RBAC roles for each namespace it sees.
+# These RBAC roles are controlled (in the owner reference sense) by the namespace.
+# The RBAC manager needs permission to set finalizers on Namespaces in order to
+# create resources that block their deletion when the
+# OwnerReferencesPermissionEnforcement admission controller is enabled.
+# See https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#ownerreferencespermissionenforcement
+- apiGroups:
+  - ""
+  resources:
+  - namespaces/finalizers
+  verbs:
+  - update
 - apiGroups:
   - apiextensions.crossplane.io
   resources:
@@ -33,6 +45,18 @@ rules:
   - get
   - list
   - watch
+# The RBAC manager creates a series of RBAC cluster roles for each XRD it sees.
+# These cluster roles are controlled (in the owner reference sense) by the XRD.
+# The RBAC manager needs permission to set finalizers on XRDs in order to
+# create resources that block their deletion when the 
+# OwnerReferencesPermissionEnforcement admission controller is enabled.
+# See https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#ownerreferencespermissionenforcement
+- apiGroups:
+  - apiextensions.crossplane.io
+  resources:
+  - compositeresourcedefinitions/finalizers
+  verbs:
+  - update
 - apiGroups:
   - pkg.crossplane.io
   resources:
@@ -41,6 +65,18 @@ rules:
   - get
   - list
   - watch
+# The RBAC manager creates a series of RBAC cluster roles for each ProviderRevision
+# it sees. These cluster roles are controlled (in the owner reference sense) by the
+# ProviderRevision. The RBAC manager needs permission to set finalizers on
+# ProviderRevisions in order to create resources that block their deletion when the 
+# OwnerReferencesPermissionEnforcement admission controller is enabled.
+# See https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#ownerreferencespermissionenforcement
+- apiGroups:
+  - pkg.crossplane.io
+  resources:
+  - providerrevisions/finalizers
+  verbs:
+  - update
 - apiGroups:
   - apiextensions.k8s.io
   resources:
--- a/cluster/local/integration_tests.sh
+++ b/cluster/local/integration_tests.sh
@@ -58,16 +58,20 @@ if [ "$skipcleanup" != true ]; then
     trap cleanup EXIT
 fi
 
-echo_step "creating k8s cluster using kind"
-"${KIND}" create cluster --name="${K8S_CLUSTER}"
+readonly DEFAULT_KIND_CONFIG="kind: Cluster
+apiVersion: kind.x-k8s.io/v1alpha4
+"
+echo_step "creating k8s cluster using kind with the following config:"
+kind_config="${KIND_CONFIG:-$DEFAULT_KIND_CONFIG}"
+echo "${kind_config}"
+echo "${kind_config}" | "${KIND}" create cluster --name="${K8S_CLUSTER}" --config=-
 
 # tag crossplane image and load it to kind cluster
 docker tag "${BUILD_IMAGE}" "${CROSSPLANE_IMAGE}"
 "${KIND}" load docker-image "${CROSSPLANE_IMAGE}" --name="${K8S_CLUSTER}"
 
 echo_step "installing helm package(s) into \"${CROSSPLANE_NAMESPACE}\" namespace"
-"${KUBECTL}" create ns "${CROSSPLANE_NAMESPACE}"
-"${HELM3}" install "${PROJECT_NAME}" --namespace "${CROSSPLANE_NAMESPACE}" "${projectdir}/cluster/charts/${PROJECT_NAME}" --set replicas=2,rbacManager.replicas=2,image.pullPolicy=Never,imagePullSecrets=''
+"${HELM3}" install --create-namespace -n "${CROSSPLANE_NAMESPACE}" "${PROJECT_NAME}" "${projectdir}/cluster/charts/${PROJECT_NAME}" --set replicas=2,args={'-d'},rbacManager.replicas=2,rbacManager.args={'-d'},image.pullPolicy=Never,imagePullSecrets=''
 
 echo_step "waiting for deployment ${PROJECT_NAME} rollout to finish"
 "${KUBECTL}" -n "${CROSSPLANE_NAMESPACE}" rollout status "deploy/${PROJECT_NAME}" --timeout=2m
@@ -131,6 +135,14 @@ for ((i = 1; i <= 5; i++)); do
         if $(echo "$pod_stat" | awk '{print $3}' | grep -ivq 'Running'); then
             echo_error "is not running"
             exit -1
+        else
+            echo_step_completed
+        fi
+
+        echo_info "check if has error in logs"
+        error_logs=$("${KUBECTL}" -n "${CROSSPLANE_NAMESPACE}" logs --all-containers=true --timestamps=true --tail=10 "${name}" | grep -w ERROR || true)
+        if ((${#error_logs} > 0)); then
+            echo_warn "${error_logs}"
         else
             echo_step_completed
         fi
--- a/internal/controller/rbac/definition/roles.go
+++ b/internal/controller/rbac/definition/roles.go
@@ -48,13 +48,15 @@ const (
 
 	valTrue = "true"
 
-	suffixStatus = "/status"
+	suffixStatus     = "/status"
+	suffixFinalizers = "/finalizers"
 )
 
 var (
 	verbsEdit   = []string{rbacv1.VerbAll}
 	verbsView   = []string{"get", "list", "watch"}
 	verbsBrowse = []string{"get", "list", "watch"}
+	verbsUpdate = []string{"update"}
 )
 
 // RenderClusterRoles returns ClusterRoles for the supplied XRD.
@@ -74,6 +76,17 @@ func RenderClusterRoles(d *v1.CompositeResourceDefinition) []rbacv1.ClusterRole 
 					d.Spec.Names.Plural + suffixStatus,
 				},
 				Verbs: verbsEdit,
+			},
+			{
+				// Crossplane reconciles an XR by creating one or more composed resources.
+				// These composed resources are controlled (in the owner reference sense) by the XR.
+				// Crossplane needs permission to set finalizers on XRs in order to create resources
+				// that block their deletion when the OwnerReferencesPermissionEnforcement admission controller is enabled.
+				APIGroups: []string{d.Spec.Group},
+				Resources: []string{
+					d.Spec.Names.Plural + suffixFinalizers,
+				},
+				Verbs: verbsUpdate,
 			},
 		},
 	}
--- a/internal/controller/rbac/provider/roles/roles.go
+++ b/internal/controller/rbac/provider/roles/roles.go
@@ -41,7 +41,8 @@ const (
 
 	valTrue = "true"
 
-	suffixStatus = "/status"
+	suffixStatus     = "/status"
+	suffixFinalizers = "/finalizers"
 
 	pluralEvents     = "events"
 	pluralConfigmaps = "configmaps"
@@ -53,6 +54,7 @@ var (
 	verbsEdit   = []string{rbacv1.VerbAll}
 	verbsView   = []string{"get", "list", "watch"}
 	verbsSystem = []string{"get", "list", "watch", "update", "patch", "create"}
+	verbsUpdate = []string{"update"}
 )
 
 // Extra rules that are granted to all provider pods.
@@ -104,6 +106,16 @@ func RenderClusterRoles(pr *v1.ProviderRevision, crds []extv1.CustomResourceDefi
 		})
 	}
 
+	// Provider pods may create Kubernetes secrets containing managed resource connection details.
+	// These secrets are controlled (in the owner reference sense) by the managed resource.
+	// Crossplane needs permission to set finalizers on managed resources in order to create secrets
+	// that block their deletion when the OwnerReferencesPermissionEnforcement admission controller is enabled.
+	ruleFinalizers := rbacv1.PolicyRule{
+		APIGroups: groups,
+		Resources: []string{rbacv1.ResourceAll + suffixFinalizers},
+		Verbs:     verbsUpdate,
+	}
+
 	edit := &rbacv1.ClusterRole{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: namePrefix + pr.GetName() + nameSuffixEdit,
@@ -137,7 +149,7 @@ func RenderClusterRoles(pr *v1.ProviderRevision, crds []extv1.CustomResourceDefi
 	// directly to the service account tha provider runs as.
 	system := &rbacv1.ClusterRole{
 		ObjectMeta: metav1.ObjectMeta{Name: SystemClusterRoleName(pr.GetName())},
-		Rules:      append(append(withVerbs(rules, verbsSystem), rulesSystemExtra...), pr.Status.PermissionRequests...),
+		Rules:      append(append(append(withVerbs(rules, verbsSystem), ruleFinalizers), rulesSystemExtra...), pr.Status.PermissionRequests...),
 	}
 
 	roles := []rbacv1.ClusterRole{*edit, *view, *system}
--- a/internal/controller/rbac/definition/roles_test.go
+++ b/internal/controller/rbac/definition/roles_test.go
@@ -74,6 +74,11 @@ func TestRenderClusterRoles(t *testing.T) {
 							Resources: []string{pluralXR, pluralXR + suffixStatus},
 							Verbs:     verbsEdit,
 						},
+						{
+							APIGroups: []string{group},
+							Resources: []string{pluralXR + suffixFinalizers},
+							Verbs:     verbsUpdate,
+						},
 					},
 				},
 				{
@@ -157,6 +162,11 @@ func TestRenderClusterRoles(t *testing.T) {
 							APIGroups: []string{group},
 							Resources: []string{pluralXR, pluralXR + suffixStatus},
 							Verbs:     verbsEdit,
+						},
+						{
+							APIGroups: []string{group},
+							Resources: []string{pluralXR + suffixFinalizers},
+							Verbs:     verbsUpdate,
 						},
 						{
 							APIGroups: []string{group},
--- a/internal/controller/rbac/provider/roles/roles_test.go
+++ b/internal/controller/rbac/provider/roles/roles_test.go
@@ -150,6 +150,11 @@ func TestRenderClusterRoles(t *testing.T) {
 							Resources: []string{pluralCRDC, pluralCRDC + suffixStatus},
 							Verbs:     verbsSystem,
 						},
+						{
+							APIGroups: []string{groupCRDA, groupCRDC},
+							Resources: []string{rbacv1.ResourceAll + suffixFinalizers},
+							Verbs:     verbsUpdate,
+						},
 					}, rulesSystemExtra...),
 				},
 			},


el-mail:
I'm not sure, but maybe we should update permissions rbac-manager on Helm chart.
https://github.com/el-mail/crossplane/blob/master/cluster/charts/crossplane/templates/rbac-manager-clusterrole.yaml

add rules, like this:
```
...
- apiGroups:
  - rbac.authorization.k8s.io
  resources:
  - clusterrolebindings/finalizers
  - clusterroles/finalizers
  - roles/finalizers
  verbs:
  - update```

**Comment by el-mail** [comment]
Aloh @chobostar! 
Thank you for explaining, I added this improvement for Helm chart 

https://github.com/crossplane/crossplane/pull/3444/commits/5092cb66ac9245fb06b6032a826cff7912695ee6

**Comment by chobostar** [comment]
Here a relevant fix for k8s upstream: https://github.com/kubernetes/kubernetes/pull/101186

So the point is "the clearer the intention, the better the execution", this PR consistent with k8s community about finalizers RBAC.

---
Alternative solution is disable `OwnerReferencesPermissionEnforcement` admission controller, however plugin prevents unauthorized users and pods from delaying arbitrary pods.
See:
- https://kubernetes.io/docs/admin/admission-controllers/#ownerreferencespermissionenforcement
- https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#foreground-cascading-deletion

**Comment by negz** [comment]
@el-mail Please make sure your commits are signed (see the DCO check) and update the PR description per the template (in particular detail how you've tested this change).

File: internal/controller/rbac/definition/roles.go [hunk 1]
```diff
@@ -48,7 +48,8 @@ const (
 
 	valTrue = "true"
 
-	suffixStatus = "/status"
+	suffixStatus     = "/status"
+	suffixFinalizers = "/finalizers"
```
negz, line 52: Does this sub resource exist? Is it a special OpenShift thing?

I thought it was something being proposed upstream, but not yet implemented. e.g. https://github.com/kubernetes/enhancements/issues/3285

el-mail: @negz thank you for involving
I'm not sure if exist subresource management in OpenShift, but It seems rights on finalizer CR's needed. 
I can check it

P.S. By default, OwnerReferencesPermissionEnforcement is enabled https://docs.openshift.com/container-platform/4.10/architecture/admission-plug-ins.html#admission-plug-ins-default_admission-plug-ins

negz: Right, I'm familiar with finalizers, what I've never seen before is the `someresource/finalizers` as a valid RBAC resource. I thought this had to be for [subresources](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#subresources) like `/status`.

I see from other PRs that this is a real thing, so I'm just curious whether it affects upstream Kubernetes or just OpenShift.

cluster/charts/crossplane/templates/rbac-manager-clusterrole.yaml
```diff
@@ -41,6 +41,12 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - pkg.crossplane.io
+  resources:
+  - providerrevisions/finalizers
+  verbs:
+  - update
```
negz, line 49: Could you add a comment here please. Something like:

> The RBAC manager creates an RBAC ClusterRole for each ProviderRevision it sees. It needs RBAC access to update the finalizers of ProviderRevisions in order to set blockOwnerDeletion on ClusterRoles owned by a ProviderRevision per https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#ownerreferencespermissionenforcement

negz: Also, I think a few other updates are needed to cover the case where the RBAC manager creates RBAC roles that block the deletion of their owners:

* https://github.com/crossplane/crossplane/blob/377239/internal/controller/rbac/provider/roles/roles.go#L145
* https://github.com/crossplane/crossplane/blob/377239/internal/controller/rbac/definition/roles.go#L168
* https://github.com/crossplane/crossplane/blob/377239/internal/controller/rbac/namespace/roles.go#L84

You covered the first one `providerrevisions/finalizers`, but I think it also needs `namespaces/finalizers` and `compositeresourcedefinitions/finalizers`.

internal/controller/rbac/definition/roles.go
```diff
@@ -72,6 +73,7 @@ func RenderClusterRoles(d *v1.CompositeResourceDefinition) []rbacv1.ClusterRole
 				Resources: []string{
 					d.Spec.Names.Plural,
 					d.Spec.Names.Plural + suffixStatus,
+					d.Spec.Names.Plural + suffixFinalizers,
```
negz, line 76: I'm not following why this is needed - can you add a comment (to the code) explaining?

This system role is going to allow Crossplane (the 'core' pod, not the RBAC manager pod) `*` access to any composite resource (XR) defined by an XRD. Crossplane does create resources that block the owner deletion of these resources (e.g. when creating a composed resource owned by the XR) but I would have thought the `*` verb already implied it could update finalizers for the XR.

negz: (Ditto for the claims variant below.)

cluster/local/integration_tests.sh
```diff
@@ -58,16 +58,26 @@ if [ "$skipcleanup" != true ]; then
...
+fi
 
 # tag crossplane image and load it to kind cluster
 docker tag "${BUILD_IMAGE}" "${CROSSPLANE_IMAGE}"
 "${KIND}" load docker-image "${CROSSPLANE_IMAGE}" --name="${K8S_CLUSTER}"
 
 echo_step "installing helm package(s) into \"${CROSSPLANE_NAMESPACE}\" namespace"
-"${KUBECTL}" create ns "${CROSSPLANE_NAMESPACE}"
-"${HELM3}" install "${PROJECT_NAME}" --namespace "${CROSSPLANE_NAMESPACE}" "${projectdir}/cluster/charts/${PROJECT_NAME}" --set replicas=2,rbacManager.replicas=2,image.pullPolicy=Never,imagePullSecrets=''
+"${KUBECTL}" create ns "${CROSSPLANE_NAMESPACE}" || $skipcleanup
```
chobostar, line 78: Starting from [3.2.0](https://github.com/helm/helm/pull/7648) helm has
```
--create-namespace    create the release namespace if not present
```
option
https://helm.sh/docs/helm/helm_install/#options

cluster/local/integration_tests.sh
```diff
@@ -58,16 +58,26 @@ if [ "$skipcleanup" != true ]; then
...
+$skipcleanup && "${HELM3}" uninstall "${PROJECT_NAME}" --namespace "${CROSSPLANE_NAMESPACE}" || true
+"${HELM3}" upgrade --install "${PROJECT_NAME}" --namespace "${CROSSPLANE_NAMESPACE}" "${projectdir}/cluster/charts/${PROJECT_NAME}" --set replicas=2,args={'--debug'},rbacManager.replicas=2,rbacManager.args={'--debug'},image.pullPolicy=Never,imagePullSecrets=''
 
 echo_step "waiting for deployment ${PROJECT_NAME} rollout to finish"
```
chobostar, line 82: It looks like stale code, currently helm supports `--wait` or `--atomic` option for this purpose:
>--wait                                       if set, will wait until all Pods, PVCs, Services, and minimum number of Pods of a Deployment, StatefulSet, or ReplicaSet are in a ready state before marking the release as successful. It will wait for as long as --timeout

the code of feature:
https://github.com/helm/helm/blob/3974136b6eba92f17a821aac52a80b59a9fa1202/pkg/kube/wait.go#L45-L62

However it's up to you

cluster/local/integration_tests.sh
```diff
@@ -135,6 +145,14 @@ for ((i = 1; i <= 5; i++)); do
             echo_step_completed
         fi
 
+        echo_info "check if has error in logs"
+        error_logs=$("${KUBECTL}" -n "${CROSSPLANE_NAMESPACE}" logs --all-containers=true --timestamps=true --tail=10 "${name}" | grep -w ERROR || true)
```
chobostar, line 149: 1. the script has no `pipefail` [option](https://gist.github.com/mohanpedala/1e2ff5661761d3abd0385e8223e16425#set--o-pipefail), so result of whole command could be unpredictable.

2. you put as `${name}` the name of deployment and didn't use `deployment/` prefix, read here examples:
https://kubernetes.io/docs/reference/kubectl/cheatsheet/#interacting-with-deployments-and-services
current result:
    ```bash
    root@build-01fa6cad-inttests-control-plane:/# kubectl -n crossplane-system get pods                                                         
    NAME                                       READY   STATUS    RESTARTS   AGE
    crossplane-d44d85996-8ktkm                 1/1     Running   0          3m26s
    crossplane-d44d85996-wsc5c                 1/1     Running   0          3m26s
    crossplane-rbac-manager-6d4d74cdf6-9hdmq   1/1     Running   0          3m26s
    crossplane-rbac-manager-6d4d74cdf6-wxftk   1/1     Running   0          3m26s

    root@build-01fa6cad-inttests-control-plane:/# kubectl -n crossplane-system logs crossplane --all-containers=true --timestamps=true --tail=10
    Error from server (NotFound): pods "crossplane" not found
    ```

3. Why you use `--tail=10` ? What if errors not in last 10 lines?

cluster/charts/crossplane/templates/rbac-manager-clusterrole.yaml
```diff
@@ -89,4 +89,22 @@ rules:
...
+- apiGroups:
+  - apiextensions.k8s.io
+  resources:
+  - customresourcedefinitions/finalizers
```
chobostar, line 107: Are you sure you mean `customresourcedefinitions/finalizers` instead of 
`compositeresourcedefinitions/finalizers` ? If so, could you explain why rbac-manager needs this?

cluster/local/integration_tests.sh
```diff
@@ -58,16 +58,26 @@ if [ "$skipcleanup" != true ]; then
     trap cleanup EXIT
 fi
 
-echo_step "creating k8s cluster using kind"
-"${KIND}" create cluster --name="${K8S_CLUSTER}"
+readonly DEFAULT_KIND_CONFIG="kind: Cluster
+apiVersion: kind.x-k8s.io/v1alpha4
+"
+if $skipcleanup && $(echo $("${KIND}" get clusters) | grep -wq "${K8S_CLUSTER}"); then
```
chobostar, line 64: lot of unnecessary code added just to support to output:
```
"existing kind cluster: <somecluster>
```
The script already has clear error text:
```
>>>>>>> creating k8s cluster using kind
ERROR: failed to create cluster: node(s) already exist for a cluster with the name "build-01fa6cad-inttests"
```
It could be useful only if you have to run integration tests multiple times during **one** pipeline.

negz:
@el-mail Can you clarify why you're updating the integration tests in this PR? Updates are welcome, but I would prefer them to be in a separate PR unless they're related to the rest of this PR (e.g. the finalizer permissions).

el-mail:
> @el-mail Can you clarify why you're updating the integration tests in this PR? Updates are welcome, but I would prefer them to be in a separate PR unless they're related to the rest of this PR (e.g. the finalizer permissions).

@negz, the integration test was improved for this case when need to create KIND cluster with enabled admission plugin configuration.
example usage:
```
$ KIND_CONFIG=$(cat .work/config.yaml) make e2e
12:30:59 [ .. ] running integration tests using kind v0.14.0

>>>>>>> creating k8s cluster using kind with the following config:
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
- role: control plane
   kubeadmConfigPatches:
   - |
     kind: ClusterConfiguration
     apiServer:
         extraArgs:
           enable-admission-plugins: OwnerReferencesPermissionEnforcement
```

Thank you for your comments on this PR. I've taken in develop
PR has been already moved to Draft.

cluster/local/integration_tests.sh
el-mail: I haven't added it because of backward compatibility reasons

el-mail:
@negz I think this PR is ready for testing, can you check it

how to run e2e test with admission plugin

```
$ KIND_CONFIG="kind: Cluster
  apiVersion: kind.x-k8s.io/v1alpha4
  nodes:
  - role: control-plane
    kubeadmConfigPatches:
    - |
      kind: ClusterConfiguration
      apiServer:
          extraArgs:
            enable-admission-plugins: OwnerReferencesPermissionEnforcement" make e2e
            
```
or 
```
$ cat > .work/config.yaml <<EOF
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
- role: control-plane
  kubeadmConfigPatches:
  - |
    kind: ClusterConfiguration
    apiServer:
        extraArgs:
          enable-admission-plugins: OwnerReferencesPermissionEnforcement

EOF
$ KIND_CONFIG=$(cat .work/config.yaml) make e2e
```

el-mail:
Hi @negz, happy new year to you 🎊 

have any update on this PR?

braghettos:
Is it possible to add this PR also for Crossplane version 1.10.2?

internal/controller/rbac/definition/roles.go
```diff
@@ -48,13 +48,15 @@ const (
...
 )
 
 var (
-	verbsEdit   = []string{rbacv1.VerbAll}
-	verbsView   = []string{"get", "list", "watch"}
-	verbsBrowse = []string{"get", "list", "watch"}
+	verbsEdit           = []string{rbacv1.VerbAll}
+	verbsView           = []string{"get", "list", "watch"}
+	verbsBrowse         = []string{"get", "list", "watch"}
+	verbsOwnerReference = []string{"update"}
```
negz, line 59:
```suggestion
	verbsUpdate = []string{"update"}
```

Please rename all instances of `verbsOwnerReference` to `verbsUpdate`. The naming convention is "verbs that allow X", not "verbs for X".

internal/controller/rbac/provider/roles/roles.go
```diff
@@ -41,7 +41,8 @@ const (
 
 	valTrue = "true"
 
-	suffixStatus = "/status"
+	suffixStatus     = "/status"
+	suffixFinalizers = "/finalizers"
```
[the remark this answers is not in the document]
negz: My request here still stands - not sure why this was resolved.

negz (changes requested):
There's a general theme here that I'm finding this change difficult to review because it's not explained in enough detail.

I understand that when the `OwnerReferencesPermissionEnforcement` admission controller is enabled an RBAC subject must have permission to update finalizers on a resource in order to create another resource that blocks its deletion. What is making review challenging is that for each place you've granted this permission I'm having to spend time figuring out why it is needed in that specific case.

I would very much appreciate it if you could add code comments explaining specifically why you're granting these permissions in each case. If the changes aren't self-explanatory to me as the author of the RBAC manager they're probably not self-explanatory to future readers either.

internal/controller/rbac/provider/roles/roles.go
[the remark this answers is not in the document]
el-mail: @negz interesting, for real, request was resolved, but my comment was not submitted.
can retry again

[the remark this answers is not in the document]
el-mail: @negz this is done
https://github.com/crossplane/crossplane/blob/a983f57874435d5347f89161e7efd4e931c728f4/internal/controller/rbac/provider/roles/roles.go#L138
https://github.com/crossplane/crossplane/blob/a983f57874435d5347f89161e7efd4e931c728f4/internal/controller/rbac/definition/roles.go#L104
can you check it

internal/controller/rbac/definition/roles.go
```diff
@@ -99,6 +101,15 @@ func RenderClusterRoles(d *v1.CompositeResourceDefinition) []rbacv1.ClusterRole
...
 			},
+			// The rule grants access to update the finalizers of resources that belong to the API group.
+			// If OwnerReferencesPermissionEnforcement is enabled, it allows after deleting the XRD to
+			// clean up resources that they no longer own, so that the system can be kept in a consistent state.
+			// See https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#ownerreferencespermissionenforcement
+			{
+				APIGroups: []string{d.Spec.Group},
+				Resources: []string{d.Spec.Names.Plural + suffixFinalizers},
+				Verbs:     verbsUpdate,
+			},
```
negz, line 112: Are you sure we want the edit role to have these permissions and not the system role?

* https://github.com/crossplane/crossplane/blob/0951db/cluster/charts/crossplane/templates/clusterrole.yaml#L12

I'm a bit rusty on the RBAC manager, but I'm pretty sure the Crossplane pod ran as the system role (not edit, which is for human users) - see `clusterrole.yaml` above. I'm guessing the intent here is that Crossplane should be allowed to create composed resources that block the deletion of their XRD.

internal/controller/rbac/definition/roles.go
```diff
@@ -99,6 +101,15 @@ func RenderClusterRoles(d *v1.CompositeResourceDefinition) []rbacv1.ClusterRole
 				Resources: []string{d.Spec.Names.Plural},
 				Verbs:     verbsEdit,
 			},
+			// The rule grants access to update the finalizers of resources that belong to the API group.
+			// If OwnerReferencesPermissionEnforcement is enabled, it allows after deleting the XRD to
+			// clean up resources that they no longer own, so that the system can be kept in a consistent state.
```
negz, line 106: It's probably not the XRD (the definition) that is being deleted, but rather the XR, right? I think the intention here is to allow the Crossplane pod to add a finalizer to XRs, so that it can create composed resources that block their deletion.

If that is the case, it's not clear to me why we need this rule at all. I see you observed issues with composition per https://github.com/crossplane/crossplane/issues/3443#issuecomment-1306828892, but I don't understand why that is happening. Crossplane already has `verbsEdit` (i.e. `*`) access to XRs, which I would assume included access to update XR finalizers. Perhaps `*` not apply to the `/finalizers` resource?

Crossplane definitely has RBAC access to add/remove XR finalizers today - see for example https://github.com/crossplane/crossplane/blob/0951db/internal/controller/apiextensions/composite/reconciler.go#L457

If we do need this rule, I'm pretty sure it should be on the `system` role rather than the `edit` role. The `system` role is what the Crossplane pod runs as - `edit` is for human users. I'd also prefer the comment to be a little more detailed, for example:

> Crossplane reconciles an XR by creating one or more composed resources. These composed resources are controlled (in the owner reference sense) by the XR. Crossplane needs permission to set finalizers on XRs in order to create resources that block their deletion when the OwnerReferencesPermissionEnforcement admission controller is enabled.

internal/controller/rbac/provider/roles/roles.go
```diff
@@ -123,6 +135,12 @@ func RenderClusterRoles(pr *v1.ProviderRevision, crds []extv1.CustomResourceDefi
 		Rules: withVerbs(rules, verbsEdit),
 	}
 
+	// The rule grants access to update the finalizers of resources that belong to the API group.
+	// If OwnerReferencesPermissionEnforcement is enabled, it allows after deleting the provider to
+	// clean up resources that they no longer own, so that the system can be kept in a consistent state.
+	// See https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#ownerreferencespermissionenforcement
+	edit.Rules = append(edit.Rules, withVerbs(finalizerRules, verbsUpdate)...)
```
negz, line 142: > it allows after deleting the provider to clean up resources that they no longer own

Is this comment accurate? It sounds like you're saying that this somehow helps with deleting a `Provider.pkg`?

I might be wrong, but I _think_ what you're actually trying to do here is allow Crossplane access to update the finalizers of a managed resource. The only reason I can think of that Crossplane would need this permission would be to create connection secrets that would block the deletion of a managed resource. If that's the case, I think this rule should apply to the `system` role rather than `edit`. The `system` role is what the provider pod runs as, and it's the provider pod that would be creating managed resource connection secrets.

If I'm right, I would prefer the comment to be more explicit, for example:

> Provider pods may create Kubernetes secrets containing managed resource connection details. These secrets are controlled (in the owner reference sense) by the managed resource. Crossplane needs permission to set finalizers on managed resources in order to create secrets that block their deletion when the OwnerReferencesPermissionEnforcement admission controller is enabled.

negz (changes requested):

el-mail:
Hi @negz!
Finally, I have updated PR and added only the needed rules to fix the issue https://github.com/crossplane/crossplane/issues/3443

How can check it:

1. Create Kind cluster with OwnerReferencesPermissionEnforcement plugin
```
$ cat > .work/config.yaml <<EOF
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
- role: control-plane
  kubeadmConfigPatches:
  - |
    kind: ClusterConfiguration
    apiServer:
        extraArgs:
          enable-admission-plugins: OwnerReferencesPermissionEnforcement
EOF
$ skipcleanup=true KIND_CONFIG=$(cat .work/config.yaml) make e2e
```
2. Install package:
`kubectl crossplane install configuration registry.upbound.io/xp/getting-started-with-aws:v1.10.2`
3. Create claim:
`kubectl --token=<crossplane token> apply -f https://raw.githubusercontent.com/crossplane/crossplane/release-1.10/docs/snippets/compose/claim-aws.yaml`
4. Check XR: 
`kubectl describe xpostgresqlinstances.database.example.org`
5. Profit! :)

el-mail:
> Hi @negz! Finally, I have updated PR and added only the needed rules to fix the issue #3443
> 
> How can check it:
> 
>     1. Create `kind` cluster with OwnerReferencesPermissionEnforcement plugin
> ...

P.S. This change is enough to fix, but in the future, we can also add rules like this for providers.


https://github.com/crossplane/crossplane/blob/06d23c35d0fda18fb79d7bfdce3f068694081d2e/internal/controller/rbac/provider/roles/roles.go#L99
```
	// The rule that is granted access to update the finalizers in all groups.
	ruleFinalizers := rbacv1.PolicyRule{
		APIGroups: groups,
		Resources: []string{rbacv1.ResourceAll + suffixFinalizers},
		Verbs:     verbsUpdate,
	}
```

internal/controller/rbac/provider/roles/roles.go
negz: > The only reason I can think of that Crossplane would need this permission would be to create connection secrets that would block the deletion of a managed resource.

@el-mail I see you've removed this rule entirely. Did you determine it wasn't needed? Have you confirmed in your testing that managed resource connection secrets are being written successfully?

internal/controller/rbac/definition/roles.go
negz: > I don't understand why that is happening. Crossplane already has verbsEdit (i.e. *) access to XRs, which I would assume included access to update XR finalizers. Perhaps * not apply to the /finalizers resource?

I see you've moved this rule to `system` - thank you. It's still not clear to me why/whether it's needed though per my above comment. Can you clarify?

el-mail: @negz yes,

When XR tries to create a managed resource with ownerReference them, these rules give to allow update finalizers to MR

without this permission, crossplane cannot create resources. I've got issues when I try to use my own XRD:
`cannot compose resources: cannot apply composed resource: cannot create
        object: buckets.s3.aws.crossplane.io "default-bucket-test1" is
        forbidden: cannot set blockOwnerDeletion if an ownerReference refers to
        a resource you can't set finalizers on: , <nil>`
this exception happened is here:
https://github.com/crossplane/crossplane/blob/master/internal/controller/apiextensions/composite/composition_pt.go#L244

and this is when I try to use  
`kubectl apply -f https://raw.githubusercontent.com/crossplane/crossplane/release-1.10/docs/snippets/compose/claim-aws.yaml`
error from XR:
`cannot publish connection details: cannot apply connection secret:
        cannot create object: secrets "" is
        forbidden: cannot set blockOwnerDeletion if an ownerReference refers to
        a resource you can't set finalizers on: , <nil>`

this exception happened is here:
https://github.com/crossplane/crossplane/blob/master/internal/controller/apiextensions/composite/reconciler.go#L533

p.s. s, you're absolutely right about set Finalizers in any resource like this https://github.com/crossplane/crossplane/blob/0951db/internal/controller/apiextensions/composite/reconciler.go#L457, maybe I should research how crossplane works with RBAC and resource, but this is needed more time for me. I noticed between Crossplane and any Providers have a gap in permissions

internal/controller/rbac/provider/roles/roles.go
el-mail: @negz this is needed.
Without this permission, I've got this error at MR:
`connect failed: cannot track ProviderConfig usage: cannot apply
        ProviderConfigUsage: cannot create object:
        providerconfigusages.aws.crossplane.io
        "1220bf14-83ce-484b-a321-dc75495c6227" is forbidden: cannot set
        blockOwnerDeletion if an ownerReference refers to a resource you can't
        set finalizers on: , <nil>`

https://github.com/crossplane-contrib/provider-aws/blob/c9ed477f32e6e22b2c7ac7ed251b53ca927b5054/pkg/clients/aws.go#L420

internal/controller/rbac/definition/roles.go
negz: Interesting. I guess it needs update on `resource`/finalizers` specifically. I would have thought that was implied by having `*` on `resource`. Thanks for checking!

negz (approved):
Thanks for working through this tricky PR @el-mail!

el-mail:
> Thanks for working through this tricky PR @el-mail!

Thank you @negz too!

migueladh:
Hi all!
Will this be backported to v1.11?

PR #3453, opened 2022-11-18
Promote `CompositionRevision` to v1beta1

### LINKED ISSUE #2656: Promote `CompositionRevision` to v1beta1

ezgidemirel:
Signed-off-by: ezgidemirel <ezgidemirel91@gmail.com>

This PR enables the feature flag `--enable-composition-revisions` by default. In addition to that, it promotes the composition revision version to `v1beta1`

Fixes #2656 

I have:

1) Installed Crossplane v1.10.1
2) Enabled composition revision feature
3) Created a composition
4) Scaled down the deployment
5) Applied new CRDs
6) `make run`
7) Validated that new composition revision is created with new hash suffix, existing composition revisions are updated to `v1beta1` automatically. 


muvaf (approved):
Looking great! @ezgidemirel Just to make sure, could you test this with a real `helm upgrade` just to make sure we're over the path of how users are going to upgrade? For both users who are already using `CompositionRevision` with enabled flag and who haven't enabled the flag at all.

ezgidemirel:
Thanks @muvaf for your comment. I followed the steps below to test with helm charts and verified that composition revision versions are updated to `v1beta1` and crossplane pod is running as expected:

```
helm install crossplane --namespace crossplane-system crossplane-stable/crossplane --set args='{"--enable-composition-revisions"}'

helm upgrade crossplane /Users/ezgidemirel/Projects/crossplane/_output/charts/crossplane-1.11.0-rc.0.82.g6d3d0842.dirty.tgz -f values.yaml
```
where `values.yaml` file is 

```
args: {}
```

cmd/crossplane/core/core.go
```diff
@@ -76,7 +76,6 @@ type startCommand struct {
 	PollInterval     time.Duration `help:"How often individual resources will be checked for drift from the desired state." default:"1m"`
 	MaxReconcileRate int           `help:"The global maximum rate per second at which resources may checked for drift from the desired state." default:"10"`
 
-	EnableCompositionRevisions bool `group:"Alpha Features:" help:"Enable support for CompositionRevisions."`
```
negz, line 79: For beta features I'd like to keep the feature flag in place, but move it into `group:"Beta Features:"` and have it enabled by default. That way folks can opt-out of the feature while its in beta if they so desire. This is in line with how Kubernetes feature maturity works.

apis/apiextensions/v1beta1/doc.go
```diff
@@ -0,0 +1,21 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package v1beta1 contains API types that extend the Crossplane API.
+// +kubebuilder:object:generate=true
+// +groupName=apiextensions.crossplane.io
+// +versionName=v1beta1
+package v1beta1
```
negz, line 21: A comment on the alpha and beta packages reminding folks that their schemas must be identical may help. It hasn't been super reliable in the past, but presumably better than nothing.

negz: Also, with the new beta types becoming the storage version (and generally "most mature" version) I think it makes sense to update all the controller code to reconcile the v1beta1 types.

negz (changes requested):

negz:
Is there anything we could do to build confidence in this feature before we ship it at v1beta1? It had a lot of time to soak at v1alpha1, but anything we can do to build confidence in the recent changes before committing to beta level stability would be worthwhile.

Note this doesn't need to block merging this - it's something we could do before the new APIs are released with v1.11. I'm thinking perhaps ask folks who are using CompositionRevisions to test out a pre-release build of Crossplane na make sure it works for them / addresses their problems.

MisterMX:
I want point to at #3274 again. CompositionRevisions will continue to break on API enhancement. And as long as there is no clear way of fixing that, I don't think they are stable enough to enter beta stage.

muvaf:
@MisterMX I think the schemas should definitely match when it's v1beta1 but the way we keep them in sync feels more about our maintenance concerns rather than the stability of the CRD or controller. We should get better at that by sharing a struct but I don't think it's blocking.

apis/apiextensions/v1beta1/doc.go
ezgidemirel: done

cmd/crossplane/core/core.go
ezgidemirel: Updated the flag name and enabled it by default.

ezgidemirel:
> Is there anything we could do to build confidence in this feature before we ship it at v1beta1? It had a lot of time to soak at v1alpha1, but anything we can do to build confidence in the recent changes before committing to beta level stability would be worthwhile.
> 
> Note this doesn't need to block merging this - it's something we could do before the new APIs are released with v1.11. I'm thinking perhaps ask folks who are using CompositionRevisions to test out a pre-release build of Crossplane na make sure it works for them / addresses their problems.

I'm working on a blogpost and tutorial material for this feature. I believe sharing those and asking for feedback will be helpful.

negz:
> We should get better at that by sharing a struct but I don't think it's blocking.

Per discussion in https://github.com/crossplane/crossplane/issues/3274 I still don't want to share a struct (e.g. have one version import another version's code) while Compositions are v1 and CompositionRevisions are not. Once they're all v1 (in the same package) I would feel more comfortable sharing code, though generating the revision type would still be ideal.

negz (approved):

PR #3458, opened 2022-11-22
feat(transform): Add ToJson string conversion

MisterMX:

This adds a new `ToJson` conversion option to the string transform which allows converting the input value into its raw JSON representation:

```yaml
- type: string
  string:
     type: Convert
     convert: ToJson
```

I have:

Manually and unit tests.

### DIFF
--- a/apis/apiextensions/v1/composition_transforms.go
+++ b/apis/apiextensions/v1/composition_transforms.go
@@ -168,6 +168,7 @@ type StringConversionType string
 const (
 	StringConversionTypeToUpper    StringConversionType = "ToUpper"
 	StringConversionTypeToLower    StringConversionType = "ToLower"
+	StringConversionTypeToJSON     StringConversionType = "ToJson"
 	StringConversionTypeToBase64   StringConversionType = "ToBase64"
 	StringConversionTypeFromBase64 StringConversionType = "FromBase64"
 )
@@ -186,9 +187,12 @@ type StringTransform struct {
 	// +optional
 	Format *string `json:"fmt,omitempty"`
 
-	// Convert the type of conversion to Upper/Lower case.
+	// Optional conversion method to be specified.
+	// `ToUpper` and `ToLower` change the letter case of the input string.
+	// `ToBase64` and `FromBase64` perform a base64 conversion based on the input string.
+	// `ToJson` converts any input value into its raw JSON representation.
 	// +optional
-	// +kubebuilder:validation:Enum=ToUpper;ToLower;ToBase64;FromBase64
+	// +kubebuilder:validation:Enum=ToUpper;ToLower;ToBase64;FromBase64;ToJson
 	Convert *StringConversionType `json:"convert,omitempty"`
 
 	// Trim the prefix or suffix from the input
--- a/docs/reference/composition.md
+++ b/docs/reference/composition.md
@@ -637,6 +637,13 @@ Currently only `multiply` is supported.
      type: Convert
      convert: FromBase64
 
+# If the value of the 'from' field is not nil, the value of the 'to' field will be
+# set to raw JSON representation of the 'from' field.
+- type: string
+  string:
+     type: Convert
+     convert: ToJson
+
 # If the value of the 'from' field is https://crossplane.io, the value of the 'to' field will
 # be set to crossplane.io
 - type: string
--- a/cluster/crds/apiextensions.crossplane.io_compositions.yaml
+++ b/cluster/crds/apiextensions.crossplane.io_compositions.yaml
@@ -266,13 +266,18 @@ spec:
                                     a string.
                                   properties:
                                     convert:
-                                      description: Convert the type of conversion
-                                        to Upper/Lower case.
+                                      description: Optional conversion method to be
+                                        specified. `ToUpper` and `ToLower` change
+                                        the letter case of the input string. `ToBase64`
+                                        and `FromBase64` perform a base64 conversion
+                                        based on the input string. `ToJson` converts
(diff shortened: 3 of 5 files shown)


nullable-eth (approved):
Looks great!

apis/apiextensions/v1/composition_transforms.go
```diff
@@ -61,6 +61,7 @@ const (
 	errStringConvertTypeFailed          = "type %s is not supported for string convert"
 
 	errDecodeString = "string is not valid base64"
+	errMarshalJson  = "failed to marshal value to JSON"
```
turkenh, line 64:
```suggestion
	errMarshalJson  = "cannot marshal value to JSON"
```

nitpick: just to be consistent with other error messages.

turkenh (approved):

ezgidemirel:
Hi @MisterMX, thanks for the PR! Would you mind sharing snippets of your test manifests?

MisterMX:
@ezgidemirel here is a working example:

`composition.yaml`
```yaml
---
kind: Composition
metadata:
  name: test-string-tojson
spec:
  compositeTypeRef:
    kind: XSecret
  resources:
    - name: secret
      base:
        apiVersion: secretsmanager.aws.crossplane.io/v1beta1
        kind: Secret
        metadata:
          name: # patched
        spec:
          forProvider:
            region: eu-central-1
            stringSecretRef:
              name: # patched
              namespace: # patched
              key: # patched
            resourcePolicy: # patched
          providerConfigRef:
            name: example
      patches:
        # Policy
        - fromFieldPath: "spec.parameters.resourcePolicy"
          toFieldPath: "spec.forProvider.resourcePolicy"
          transforms:
            - type: string
              string:
                type: Convert
                convert: ToJson
        # SecretRef
        - fromFieldPath: "spec.parameters.secretRef.name"
          toFieldPath: "spec.forProvider.stringSecretRef.name"
        - fromFieldPath: "spec.parameters.secretRef.key"
          toFieldPath: "spec.forProvider.stringSecretRef.key"
        - fromFieldPath: "metadata.labels[crossplane.io/claim-namespace]"
          toFieldPath: "spec.forProvider.stringSecretRef.namespace"
```

`xrd.yaml`
```yaml
---
metadata:
  name: xsecrets.example.org
spec:
  group: example.org
  claimNames:
    kind: Secret
    plural: secrets
  names:
    kind: XSecret
    plural: xsecrets
  versions:
    - name: v1alpha1
      served: true
      referenceable: true
      schema:
        openAPIV3Schema:
          type: object
          properties:
            spec:
              type: object
              properties:
                parameters:
                  type: object
                  properties:
                    resourcePolicy:
                      type: object
                      description: AWS Secret policy object.
                      x-kubernetes-preserve-unknown-fields: true
                    secretRef:
                      type: object
                      description: The Kubernetes Secret whose data will be sent as string to AWS.
                        The secret needs to exist in the same namespace as this claim.
                      properties:
                        key:
                          type: string
                          description: Optional key whose value should be uploaded. If not set, the
                            whole map in the Secret data will be used.
                        name:
                          type: string
                          description: Name of the Kubernetes Secret resource in the same namespace
                            as this claim.
                      required:
                        - name
              required:
                - parameters
```

`xrc.yaml`
```yaml
---
apiVersion: v1
kind: Secret
metadata:
  name: example-secret
  namespace: default
type: Opaque
stringData:
  password: passw0rd
---
kind: Secret
metadata:
  name: example-secret
  namespace: default
spec:
  parameters:
    secretRef:
      name: example-secret
    resourcePolicy:
      {
        "Version": "2012-10-17",
        "Statement": [
          {
            "Effect": "Allow",
            "Principal": {
              "Federated": [
                "*"
              ]
            },
            "Action": "secretsmanager:GetSecretValue",
            "Resource": "*"
          }
        ]
      }
```

MisterMX:
I can't get the checks to succeed. It seems like an internal error in the pipeline. @ezgidemirel do you have any ideas?

apis/apiextensions/v1/composition_transforms.go
```diff
@@ -61,6 +61,7 @@ const (
 	errStringConvertTypeFailed          = "type %s is not supported for string convert"
 
 	errDecodeString = "string is not valid base64"
+	errMarshalJson  = "cannot to marshal value to JSON"
```
hasheddan, line 64: @MisterMX looks like linter doesn't like this naming:
```
  Error: var-naming: const errMarshalJson should be errMarshalJSON (revive)
```

MisterMX: Thanks for the hint. I fixed it.

apj0nes:
@hasheddan do you know if this feature will go into the next crossplane release?

hasheddan (approved):
This is awesome -- thanks @MisterMX! I am +1 on inclusion in v1.11.0 release 👍🏻

bobh66:
If we could combine this with a FromJSON transform we could patch complex objects between resources in a composition.  Use ToJSON to patch a list/map into a string annotation, and then FromJSON to patch that annotation into another resource.  Maybe in a separate PR but it would be really useful.

docs/reference/composition.md
```diff
@@ -637,6 +637,13 @@ Currently only `multiply` is supported.
      type: Convert
      convert: FromBase64
 
+# If the value of the 'from' field is not nil, the value of the 'to' field will be
```
turkenh, line 640: @plumbis This PR has this doc change. Having this updated here is enough, or do we need a separate PR in the new [docs repo](https://github.com/crossplane/docs)?

PR #3462, opened 2022-11-23
Enable the `revive` and `nolintlint` linters.

negz:

This PR enables two linters:

* https://revive.run
* https://github.com/ashanbrown/nolintlint

Revive is intended to be a drop in replacement for `golint`, which was archived and disabled in recent history. We've previously relied on `golint` to (presumably among other things) require that all exported symbols have a godoc string. Revive appeared to not be working because of the small config block we were supplying; when there is no config supplied it works in `golint` compatibility mode, which is what we want.

While I was playing with linters I decided to enable one I recently discovered - `nolintlint`. This linter is useful in that:

* It requires folks who use a `//nolint` directive to explain why. I've frequently done this manually in code review.
* It can spot `//nolint` directives that aren't doing anything, e.g. because a function previously triggered the `gocyclo` linter was refactored such that it no longer does.

I have:

I've run `make reviewable`, which invokes the linters.


internal/xpkg/name.go
```diff
@@ -65,7 +65,7 @@ func FriendlyID(name, hash string) string {
 }
 
 // ToDNSLabel converts the string to a valid DNS label.
-func ToDNSLabel(s string) string { // nolint:gocyclo
+func ToDNSLabel(s string) string { //nolint:gocyclo // TODO(negz): Document the conditions in this function.
```
turkenh, line 68: nit: the `TODO` here does not seem to be answering the reasoning linter exception :)

turkenh (approved):
Looking good, thanks!

hasheddan (approved):

PR #3465, opened 2022-11-24
Support running Containerized Composition functions

negz:

Fixes https://github.com/crossplane/crossplane/issues/2524
Closes https://github.com/crossplane/crossplane/pull/3550

This PR implements Composition Functions per the [design document](https://github.com/crossplane/crossplane/blob/b9e4773dc431d974a44232b95500e6f815a9b8ac/design/design-doc-composition-functions.md).

It introduces a new OCI image - `crossplane/xfn` - with a corresponding `xfn` binary.

* `xfn start` starts a long-running process, listening on a gRPC API for requests to run . 
* `xfn run` is a test/helper utility that runs a single Composition Function.

I made a simple no-op Composition Function that can be used to test `xfn`:

```console
$ cat ~/fnio.yaml | xfn run negz/xfn-nop:latest -
```

In order to run this Composition Function, `xfn` will create a new user (and mount) namespace. It will then invoke `xfn spark` as a user that is root inside the new user namespace, but unprivileged relative to the 'real' parent namespace. Spark will:

1. Fetch and cache the Composition Function image.
2. Build an OCI runtime bundle for the Composition Function container run.
3. Use an OCI runtime (by default `crun`) to run the Composition Function container.
4. Clean everything up.

If `xfn` is run with CAP_SETUID and CAP_SETGID it's able to create a slightly "better" user namespace, in that the user namespace supports up to 65,536 UIDs and GIDs. It's possible to run `xfn` without any privileges at all,  but it will be limited to creating user namespaces where only one UID and GID (the user namespace root's) exist.

If `xfn spark` is running in an environment that supports overlayfs in user namespaces (i.e. Kernel 5.11 or above, with a cache directory that is not itself on an overlayfs) it will use overlayfs to create container root filesystems. Otherwise, it will fall back to an "uncompressed" image cache that extracts uncompressed tarballs each time a function is run. The overlay is faster, while the uncompressed approach is widely compatible.

This PR also implements the "Crossplane end" of Composition Functions - new Composition logic that is capable of running a pipeline of Composition Functions as well as contemporary Patch & Transform (P&T) style Composition. While this change required significant refactoring of the Composition logic I've tried to keep things _mostly_ the same unless `--enable-composition-functions` is set.

One notable place this implementation differs from the design document is that Composition Functions must include _all_ desired state. The design document had proposed they could omit desired state they were not concerned with. I couldn't find a good way to make that design work in practice - in particular it wasn't possible to differentiate "delete this resource (or field)"  from "I don't have an opinion about this resource (or field)".

Unless anything comes up in review, the only remaining blocker to ship this with v1.11 is writing documentation.

There's a few small things I may attempt to add, In separate PRs, before v1.11 if there is time:

- [ ] Implement image pull auth
- [ ]  Have RunFunction return errors with meaningful gRPC status codes
- [ ]  Refactor spark.Run for reduced complexity, add tests
- [ ]  Figure out broken FunctionIO CRD generation (maybe due to missing object metadata?)

Post v1.11 I would like to:

- [ ] Implement some kind of garbage collection for the image cache
- [ ]  Look into optionally deploying xfn as a deployment - it should scale horizontally pretty well.
- [ ]  Determine whether we need to do anything to support multi-platform OCI images.
- [ ]  Allow functions to access the environment (perhaps once its non-alpha?)
- [ ]  Optionally cache results of function runs with the same input for configurable duration

I have:

This PR needs much more testing, but as of right now Composition Functions are running end-to-end in kind for me. I've also validated that everything appears to work as normal when Composition Functions are not enabled by deploying https://marketplace.upbound.io/configurations/upbound/platform-ref-gcp/v0.3.0.

To test this you can:

1. Checkout this PR.
2. Run `make` to build it
3. Run `./cluster/local/kind.sh up` to get a kind cluster

You can then use the below command to deploy Crossplane with alpha Composition Functions support enabled.

```
HELM3_FLAGS="--set args={--debug,--enable-composition-functions} --set xfn.enabled=true --set xfn.args={--debug}" ./cluster/local/kind.sh helm-install upbound-system
```

Once Crossplane is running you can add a `functions` array to a Composition. I've been using a locally modified fork of platform-ref-gcp to do so.

```yaml
kind: Composition
metadata:
  name: xgke.gcp.platformref.upbound.io
  labels:
    provider: GCP
spec:
  writeConnectionSecretsToNamespace: upbound-system
  compositeTypeRef:
    apiVersion: gcp.platformref.upbound.io/v1alpha1
    kind: XGKE
  functions:
  - name: nop
    type: Container
    container:
      # This is a function that just returns the desired state that it is passed unmodified
      image: negz/xfn-nop:latest
```

If you want to try write your own function you can see the I/O specification in the [Composition Functions design doc](https://github.com/crossplane/crossplane/blob/9ee7a27dcf643a92df68d2a6de6e55ce95e7f907/design/design-doc-composition-functions.md#function-api). Note as above, and contrary to the design, that the `desired` object of a `FunctionIO` must include _all_ desired state. Your function will be passed a `desired` object containing the desired state according to Crossplane (e.g. from the `resources` array of the Composition if any) and any functions that ran before it. Your function may modify the desired state in any way it sees fit.

```console
$ cat ~/fnio.yaml | docker run \
    -i --rm -u nobody \
    --security-opt seccomp=unconfined \
    --mount type=tmpfs,destination=/xfn \
    build-c4cbe24f/xfn-arm64 run negz/xfn-nop:latest -
```

I've also validated that the above command runs successfully. Note that we don't really have to completely disable seccomp; we just need to allow some extra syscalls such as `unshare` that are not allowed by Docker's (and containerd's) default seccomp policy.

If folks want to help test things out, here are a few ideas:

- [ ] Make sure everything continues to work the same without the `--enable-composition-functions` flag.
- [x] Check that functions can't access the network with `networkPolicy: Isolated`.
- [x] Check that functions can access the network (if the runner can) with `networkPolicy: Runner`.
- [x] A pipeline of multiple functions.
- [ ] Many XRs using a pipeline of multiple functions.
- [x] Writing a function.


cluster/images/xfn/Dockerfile
```diff
@@ -0,0 +1,31 @@
+# TODO(negz): Download crun via the Makefile instead?
+FROM curlimages/curl@sha256:cfdeba7f88bb85f6c87f2ec9135115b523a1c24943976a61fbf59c4f2eafd78e AS curl
+
+ARG CRUN_VERSION=1.6
+ARG TARGETOS
+ARG TARGETARCH
+
+RUN curl -fSSL -o /tmp/crun https://github.com/containers/crun/releases/download/${CRUN_VERSION}/crun-${CRUN_VERSION}-${TARGETOS}-${TARGETARCH}-disable-systemd
+RUN chmod +x /tmp/crun
```
negz, line 9: This will break the build because `crun` doesn't work for (and thus doesn't have builds for) any OS other than Linux, and is only available for certain architectures (i.e. amd64 and aarch64). I'll need to figure out some alternative here - ideally not building xfn at all by default for OSes where it won't work.

Worth noting as well that the initial impetus to pick `crun` over `runc` no longer exists, so we could just use `runc` if need be. One reason to stick with `crun` though is that it has support for running WASM containers.

negz:
Here's the example `fnio.yaml` I tested with:

```yaml
kind: FunctionIO
config:
  kind: Config
  metadata:
    name: cloudsql
  spec:
    version: POSTGRES_9_6
observed:
  composite:
    resource:
      kind: XPostgreSQLInstance
      metadata:
        name: my-db
      spec:
        parameters:
          storageGB: 20
        compositionSelector:
          matchLabels:
            provider: gcp
      status:
        conditions:
        - type: Ready
          status: True
    connectionDetails:
    - name: uri
      value: postgresql://db.example.org:5432
```

Note that per https://github.com/negz/xfns/blob/39e6bdb/xfn-nop/nop.sh the `negz/xfn-nop:latest` function doesn't actually care whether it's passed a `FunctionIO`, but `xfn run` does because it currently deserializes and then re-serializes it before passing it to the function's entrypoint.

internal/xfn/container_linux.go
```diff
@@ -0,0 +1,161 @@
...
+		// 5. In the child process, call execve(2) to execute spark.
+		//
+		// Per user_namespaces(7) the child process created by clone(2) starts
+		// out with a complete set of capabilities in the new user namespace
+		// until the call to execve(2) causes them to be recalculated. This
+		// includes the CAP_SETUID and CAP_SETGID necessary to become UID 0 in
+		// the child user namespace, effectively dropping privileges to UID
+		// 100000 in the parent user namespace.
+		//
+		// https://github.com/golang/go/blob/1b03568/src/syscall/exec_linux.go#L446
```
negz, line 118: Note to self: this giant comment block should use `/* */` style comments to match the other giant one above.

cmd/xfn/spark/spark.go
```diff
@@ -0,0 +1,217 @@
...
+	if err != nil {
+		return errors.Wrap(err, errReadRequest)
+	}
+
+	req := &v1alpha1.RunFunctionRequest{}
+	if err := proto.Unmarshal(pb, req); err != nil {
+		return errors.Wrap(err, errUnmarshalRequest)
+	}
+
+	t := req.GetRunFunctionConfig().GetTimeout().AsDuration()
```
turkenh, line 92: Would this crash if timeout is not set, e.g. GetTimeout returns nil?

internal/oci/store/overlay/store_overlay.go
```diff
@@ -0,0 +1,426 @@
...
+// overlayfs where the upper dir is itself on an overlayfs (i.e. is on a
+// container's root filesystem).
+// https://github.com/torvalds/linux/commit/459c7c565ac36ba09ffbf
+func Supported(cacheRoot string) bool {
+	// We use NewLayerWorkdir to test because it needs to create an upper dir on
+	// the same filesystem as the supplied cacheRoot in order to be able to move
+	// it into place as a cached layer. NewOverlayBundle creates an upper dir on
+	// a tmpfs, and is thus supported in some cases where NewLayerWorkdir isn't.
+	w, err := NewLayerWorkdir(cacheRoot, "supports-overlay-test", []string{})
+	if err != nil {
```
turkenh, line 78: Would be helpful to capture/propagate this error message somehow for further investigation since there are a couple of cases where `NewLayerWorkdir` returns an error.

cmd/xfn/spark/spark.go
```diff
@@ -0,0 +1,217 @@
...
+
+	// We prefer to use an overlayfs bundler where possible. It roughly doubles
+	// the disk space per image because it caches layers as overlay compatible
+	// directories in addition to the CachingImagePuller's cache of uncompressed
+	// layer tarballs. The advantage is faster start times for containers with
+	// cached image, because it creates an overlay rootfs. The uncompressed
+	// bundler on the other hand must untar all of a containers layers to create
+	// a new rootfs each time it runs a container.
+	var s store.Bundler = uncompressed.NewBundler(c.CacheDir)
+	if overlay.Supported(c.CacheDir) {
```
turkenh, line 109: Assuming this wouldn't be changed at runtime, could this be moved to initialization somewhere and passed as a config to spark?

internal/xfn/container_linux.go
```diff
@@ -0,0 +1,172 @@
...
+	stdio, err := StdioPipes(cmd, r.rootUID, r.rootGID)
+	if err != nil {
+		return nil, errors.Wrap(err, errCreateStdioPipes)
+	}
+
+	if err := cmd.Start(); err != nil {
+		return nil, errors.Wrap(err, errStartSpark)
+	}
+
+	b, err := proto.Marshal(req)
```
turkenh, line 138: Should we move this up not to have cmd started in case this returns an error?

turkenh:
As a side note, 

On my M1 Mac, I can successfully run the command in the `How has this code been tested` with Docker Desktop, however was getting the following error when I use [Colima](https://github.com/abiosoft/colima) as container runtime:

```
cat ~/fnio.yaml | docker run \                                                                                  
    -i --rm -u nobody \
    --security-opt seccomp=unconfined \
    --mount type=tmpfs,destination=/xfn \
    build-58577482/xfn-arm64 run negz/xfn-nop:latest -
crossplane: error: function failed: exit status 1: open executable: Permission denied
                   crossplane: error: Running command: /usr/local/bin/crun --root=/xfn/runtime run --bundle=/xfn/c/22c8d322-021e-4c74-9245-c9b72c016313 22c8d322-021e-4c74-9245-c9b72c016313: cannot invoke OCI runtime: exit status 1
```

With the recent commits error message is as below: 

```
xfn: error: run.Command.Run(): cannot run function: exit status 1: xfn: error: spark.Command.Run(): OCI runtime error: exit status 1
```

negz:
This PR needs a bunch of cleanup and tests, but as of right now Composition Functions are running end-to-end in kind for me. I haven't had any time to polish so if you want to try it out YMMV, but you should be able to:

1. Checkout this PR.
2. Run `make` to build it
3. Run `./cluster/local/kind.sh up` to get a kind cluster

You can then use the below command to deploy Crossplane with alpha Composition Functions support enabled.

```
HELM3_FLAGS="--set args={--debug,--enable-composition-functions} --set xfn.enabled=true --set xfn.args={--debug}" ./cluster/local/kind.sh helm-upgrade upbound-system
```

Once Crossplane is running you can add a `functions` array to a Composition.

```yaml
kind: Composition
metadata:
  name: xgke.gcp.platformref.upbound.io
  labels:
    provider: GCP
spec:
  writeConnectionSecretsToNamespace: upbound-system
  compositeTypeRef:
    apiVersion: gcp.platformref.upbound.io/v1alpha1
    kind: XGKE
  functions:
  - name: nop
    type: Container
    container:
      # This is a function that just returns the desired state that it is passed unmodified
      image: negz/xfn-nop:latest
```

If you want to try write your own function you can see the I/O specification in the [Composition Functions design doc](https://github.com/crossplane/crossplane/blob/9ee7a27dcf643a92df68d2a6de6e55ce95e7f907/design/design-doc-composition-functions.md#function-api). Note that one major way the implementation differs from the proposed design is that the `desired` object of a `FunctionIO` must include _all_ desired state. Your function will be passed a `desired` object containing the desired state according to Crossplane (e.g. from the `resources` array of the Composition if any) and any functions that ran before it. Your function may modify the desired state in any way it sees fit.

negz:
Here's the remaining TODO list for getting this feature to alpha in v1.11:

Ship blockers

- [ ] Configure `composite.Reconciler` correctly when both Secret Stores and Comp Fns are enabled
- [ ] Parse and handle FunctionIO results array (e.g. for errors)
- [ ] Refactor PTFComposer to reduce complexity of Compose method
- [ ] Tests for PTFComposer, FallBackComposer, overlay.CachingBundler
- [ ] Figure out how to package an OCI runtime with xfn (curl crun fails for most platform/arch combos)

Nice to have

- [ ] Implement image pull auth
- [ ] Move FunctionIO types to `apis/apiextensions/fn/io/v1alpha1`
- [ ] Have `RunFunction` return errors with meaningful gRPC status codes
- [ ] Refactor spark.Run for reduced complexity, add tests
- [ ] Figure out broken FunctionIO CRD generation (maybe due to missing object metadata?)

Post v1.11

- [ ] Look into optionally deploying xfn as a deployment - it should scale horizontally pretty well. Would require our gRPC config to have secure (mTLS) communication over TCP and client-side load balancing.
- [ ] Determine whether we need to do we need to do anything to support multi-platform OCI images.
- [ ] Allow functions to access the environment (perhaps once its non-alpha?)
- [ ] Optionally cache results of function runs with the same input for configurable duration

apis/apiextensions/fn/v1alpha1/functionio_types.go
```diff
@@ -0,0 +1,266 @@
...
+	// +optional
+	Desired Desired `json:"desired,omitempty"`
+
+	// Items is a list of Crossplane resources - either XRs or MRs.
+	//
+	// A function will read this field in the input FunctionIO and populate
+	// this field in the output FunctionIO.
+	// +kubebuilder:validation:EmbeddedResource
+	// +kubebuilder:pruning:PreserveUnknownFields
+	Items []runtime.RawExtension `json:"items"`
```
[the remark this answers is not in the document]
negz: Yes, good catch. I'll delete it.

internal/xfn/container_nonlinux.go
```diff
@@ -0,0 +1,40 @@
...
+*/
+
+package xfn
+
+import (
+	"context"
+
+	"github.com/crossplane/crossplane-runtime/pkg/errors"
+
+	"github.com/crossplane/crossplane/apis/apiextensions/fn/v1alpha1"
```
ezgidemirel, line 26:
```suggestion
"github.com/crossplane/crossplane/apis/apiextensions/fn/proto/v1alpha1"
```
Looks like you moved proto files into a separate folder. `make` is broken right now.

ezgidemirel, line 26:
```suggestion
	"github.com/crossplane/crossplane/apis/apiextensions/fn/proto/v1alpha1"
```
Looks like you moved the proto files into a separate folder. `make` is broken right now

apis/apiextensions/v1/composition_types.go
```diff
@@ -182,31 +199,179 @@ type ConnectionDetail struct {
...
+
+// ContainerFunctionRunner represents runner configuration for a Composition
+// Function.
+type ContainerFunctionRunner struct {
+	// TODO(negz): url.Parse can't handle the @ in the below address.
+
+	// Endpoint specifies how and where Crossplane should reach the runner it
+	// uses to invoke containerized Composition Functions.
+	// +optional
+	// +kubebuilder:default="unix-abstract://@crossplane/fn/default.sock"
```
negz, line 371: I need to remove this TODO and update this to match `DefaultTarget`.

negz: Also, interesting to note that this kind of means we have "webhook" support already. Crossplane will send a gRPC request with a bunch of stuff instructing the function runner how to run a container function, but I suppose there's nothing stopping the server ignoring all of that and just doing something with the FunctionIO.

apis/apiextensions/fn/v1alpha1/functionio_types.go
```diff
@@ -0,0 +1,266 @@
...
+// Desired state of a function pipeline invocation.
+type Desired struct {
+	// Composite reflects the desired state of the XR this function reconciles.
+	// +optional
+	Composite DesiredComposite `json:"composite,omitempty"`
+
+	// Resources reflect the desired state of composed resources, including
+	// those that do not yet exist.
+	// +optional
+	Resources []DesiredResource `json:"resources,omitempty"`
```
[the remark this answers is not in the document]
negz: I'd prefer to stick to Kubernetes API conventions here. While this isn't a "real" custom resource, it is a KRM-like object (at least it has an `apiVersion` and `kind`). I know maps are often easier to work with, but we're pretty committed to lists of named subjects everywhere in the Crossplane APIs so I think it will be less surprising to be consistent.

https://github.com/kubernetes/community/blob/0b036b/contributors/devel/sig-architecture/api-conventions.md#lists-of-named-subobjects-preferred-over-maps

apis/apiextensions/fn/v1alpha1/functionio_types.go
```diff
@@ -0,0 +1,266 @@
...
+	// deletion timestamp, name, etc?
+
+	// Resource reflects the desired XR. Functions may update the metadata,
+	// spec, and status of an XR.
+	// +optional
+	Resource *runtime.RawExtension `json:"resource,omitempty"`
+
+	// ConnectionDetails reflects the desired connection details of the XR.
+	// +optional
+	ConnectionDetails []ExplicitConnectionDetail `json:"connectionDetails"`
```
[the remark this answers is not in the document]
negz: I'd prefer not to. Similar rationale to https://github.com/crossplane/crossplane/pull/3465#discussion_r1070456314, but I think it's good for the API here to be reasonably symmetrical with `DerivedConnectionDetail` (it would be strange to me for one variant to be a map and the other to be a list of named subobjects).

cmd/xfn/spark/spark.go
negz: No, I don't think so. Most Go code generated from protobufs is pretty good about nil checking for you. `GetTimeout()` returns a `*durationpb.Duration`. You can see that `AsDuration()` calls `GetSeconds()` and `GetNano()`, both of which will just return 0 if the `*Duration` is nil.

https://github.com/protocolbuffers/protobuf-go/blob/v1.28.1/types/known/durationpb/duration.pb.go#L176

internal/xfn/container_linux.go
```diff
@@ -0,0 +1,172 @@
...
+		best way to run code in a user namespace in Go is to execute a separate
+		binary; the unix.Unshare syscall affects only one OS thread, and the Go
+		scheduler might move the goroutine to another.
+
+		Therefore we execute a small shim - xfn spark - in a new user and mount
+		namespace. spark sets up the overlayfs if the Kernel supports it. It
+		then executes an OCI runtime which creates further namespaces in order
+		to actually execute the function.
+	*/
+	cmd := exec.CommandContext(ctx, os.Args[0], spark, "--cache-dir="+r.cache) //nolint:gosec // We're intentionally executing with variable input.
```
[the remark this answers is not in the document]
negz: Possibly. Can you elaborate on what the benefit would be?

internal/oci/store/overlay/store_overlay.go
negz: What would we do with the error?

The purpose of this function is to determine whether overlayfs is supported or not. If it doesn't appear to be supported we don't return an error, we just fall back to using the more widely supported `store/uncompressed` package. It seems to me that the best we could do is log that we're not using overlayfs and why before falling back.

cmd/xfn/spark/spark.go
negz: I'm not sure I understand what you're suggesting.

Is the idea that you'd prefer to avoid running this check every time `xfn spark` runs? If so, I agree in principal but I'm not sure its worthwhile (yet). This check needs to be done by something running in the same "environment" as spark (i.e. in the same namespaces, with the same capabilities) - it's not something we could move up into the long-running `xfn start` process for example.

Presumably we could do something like:

* Add an `xfn spark --store-type` flag to force it to use a particular type of store (overlay or uncompressed).
* Add an `xfn spark --check-capabilities` flag, which would cause spark to report what it can do and then exit.
* Have `xfn start` run `xfn spark --check-capabilities` once at startup to determine what stores it supports, then pass the desired `--store-type` flag accordingly.

This all seems pretty complex though - I'd prefer to wait and see whether we actually need to remove this check from the function run path. I believe the cost is 5-6 mkdir calls, one mount, and an unmount each time we run `xfn spark`. Right now it doesn't seem to be adding a lot of overhead to the time to run a function:

```console
# Cold image cache
$ time _output/bin/linux_arm64/xfn run negz/xfn-nop  ~/xp/fnio.yaml 1>/dev/null
_output/bin/linux_arm64/xfn run negz/xfn-nop ~/xp/fnio.yaml > /dev/null  0.20s user 0.12s system 22% cpu 1.413 total

# Hot image cache
$ time _output/bin/linux_arm64/xfn run negz/xfn-nop  ~/xp/fnio.yaml 1>/dev/null
_output/bin/linux_arm64/xfn run negz/xfn-nop ~/xp/fnio.yaml > /dev/null  0.10s user 0.03s system 107% cpu 0.117 total
```

internal/xfn/container_linux.go
negz: Sure. I think the chance of hitting an error here is quite low, but we may as well get our request ready before running the function rather than having it sit around waiting for us.

negz (commented):
This is now ready for review. I've just done a self-review pass and highlighted a few things that might be useful to reviewers.

internal/controller/apiextensions/composite/composition_ptf.go
```diff
@@ -0,0 +1,982 @@
...
+func UpdateResourceRefs(s *PTFCompositionState) {
+	refs := make([]corev1.ObjectReference, 0, len(s.ComposedResources))
+	for _, cd := range s.ComposedResources {
+		// Don't record references to resources that don't exist and that failed
+		// to render. We won't apply (i.e. create) these resources this time
+		// around, so there's no need to create dangling references to them.
+		if !meta.WasCreated(cd.Resource) && !cd.Rendered {
+			continue
+		}
+		ref := meta.ReferenceTo(cd.Resource, cd.Resource.GetObjectKind().GroupVersionKind())
```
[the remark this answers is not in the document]
negz: Yeah, this is because we're trying to deprecate the use of `corev1.ObjectReference` in general (iirc it's deprecated upstream). In practice our resourceRefs are currently of that type though so somewhat unavoidable.

internal/controller/apiextensions/composite/composition_ptf.go
```diff
@@ -0,0 +1,982 @@
...
+			return errors.Wrapf(errors.Errorf(errFmtUnsupportedFnType, fn.Type), errFmtRunFn, fn.Name)
+		}
+	}
+
+	// Results of fatal severity stop the Composition process. Normal or warning
+	// results are accumulated to be emitted as events by the Reconciler.
+	for _, rs := range r {
+		switch rs.Severity {
+		case iov1alpha1.SeverityFatal:
+			return errors.Wrap(errors.New(rs.Message), errFatalResult)
```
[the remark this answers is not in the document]
negz: I believe my logic was that a subsequent function could fix and remove the fatal error (I suppose maybe that means it's non-fatal?). That said, I also prefer that the code is simpler as-is so I may be justifying it to myself for that reason. :)

internal/controller/apiextensions/composite/composition_ptf.go
```diff
@@ -0,0 +1,982 @@
...
+
+	// Run P&T logic, updating the composition state accordingly.
+	if err := c.composition.PatchAndTransform(ctx, req, state); err != nil {
+		return CompositionResult{}, errors.Wrap(err, errPatchAndTransform)
+	}
+
+	// Build the initial desired state to be passed to our Composition Function
+	// pipeline. It's expected that each function in the pipeline will mutate
+	// this state. It includes any desired state accumulated by the P&T logic.
+	d, err := FunctionIODesired(state)
```
turkenh, line 308: During my experiments, I found out that the list of desired resources here includes the desired resources from previous runs (i.e. was desired -> created -> now observed -> now included in desired). This forces each function to track previously created desired resources in case they are no longer desired, I hope could be avoided.

Here is my experiment setup:

- I built a function the desires `n` (provider-kubernetes) objects wrapping a configmap where `n` is coming from XR spec.
- I started with n=1, and I correctly observed 1 configmap created.
- I bumped n to 3, and not correctly had 3 configmaps.
- I set n back to 1 expecting the 2 configmap to be deleted leaving 1 configmap again.

This didn't happen, and I still had 3 configmaps. Digging further, I realized that my function gets the 3 configmaps in the desired list, and even though my function only wants 1 configmap, the other two stay as desired. 

Is putting everything observed into desired of the first function io intentional? Wondering if we can avoid that and only put PT resources there in the beginning?

negz: I pushed https://github.com/crossplane/crossplane/pull/3465/commits/9cae4dbbb005eef964bc5841f7c3a6ddc63ddd91, which should fix this.

internal/controller/apiextensions/composite/composition_ptf.go
```diff
@@ -0,0 +1,982 @@
...
+	}
+
+	// We apply all of our composed resources before we observe them and update
+	// in the loop below. This ensures that issues observing and processing one
+	// composed resource won't block the application of another.
+	for _, cd := range state.ComposedResources {
+		// Don't try to apply this resource if we didn't render it successfully.
+		// Note that this doesn't mean this resource won't exist; it might have
+		// been created previously.
+		if !cd.Rendered {
```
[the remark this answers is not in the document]
negz: I pushed https://github.com/crossplane/crossplane/pull/3465/commits/5c3e6772fc39709b8c63ceb5df213668d0c7cb7e to fix this.

>  I think we can consider the function pipeline itself to be "rendering" composed resources, and potentially have `ParseDesiredResource` set `Rendered == true`.

This approach didn't work out for a few reasons. At the moment we pass all resources from the P&T process as the initial desired state of the function pipeline, even if those resource failed to render. So the function pipeline might be supplied with broken or partial desired state.

If `ParseDesiredResources` just set `Rendered == true` for every desired resource that was produced by the pipeline it would 'reset' these broken resources that were passed into it from P&T Composition back to `Rendered == true` and we'd try apply them.

I considered that maybe we shouldn't pass these partially rendered resources into the function pipeline in the first place. After all, functions won't have any way to distinguish an existing desired resource that is correct, fully formed, and will apply successfully from one that isn't. To be fair, this is something functions are going to have to tolerate regardless - for example they'd need to tolerate a previous function in the pipeline producing a desired composed resource that might not be able to be applied successfully.

The issue with not passing P&T composed resources that didn't render successfully to the function pipeline is that we currently consider any resource that didn't 'survive' the function pipeline to be eligible for garbage collection. Because we fetch any existing composed resources first this means it's possible that we'd erroneously garbage collect an existing composed resource that experienced a temporary P&T render error.

I don't particularly love the asymmetry between P&T and the function pipeline here:

* If P&T fails to render something, we remember and don't apply it.
* If a function fails to render something it's expected to omit it from its desired state.

I considered just removing the check in the apply loop entirely, and just trying to apply everything even if we know the render failed. In the likely event that the apply also failed we could just emit a warning event and try again on the next reconcile. This is the approach I like most, but I think where it falls apart is the edge case in which:

1. We partially render a composed resource, then hit an error.
2. The resource is fully formed enough to apply successfully.
3. We later successfully render it, but can't update it to its "successful" state (e.g. due to an immutable field).

internal/controller/apiextensions/composite/composition_ptf.go
turkenh: Confirming that now it works as expected 👍

internal/controller/apiextensions/composite/connection.go
```diff
@@ -186,3 +168,188 @@ func (c *SecretStoreConnectionDetailsConfigurator) Configure(ctx context.Context
...
+// ExtractConfigsFromDesired builds extract configs for the supplied Composition
+// Function desired state.
+func ExtractConfigsFromDesired(dr *iov1alpha1.DesiredResource) []ConnectionDetailExtractConfig {
+	if dr == nil {
+		return nil
+	}
+	out := make([]ConnectionDetailExtractConfig, len(dr.ConnectionDetails))
+	for i := range dr.ConnectionDetails {
+		out[i] = ConnectionDetailExtractConfig{
+			Type:                    ConnectionDetailType(dr.ConnectionDetails[i].Type),
```
turkenh, line 297: There is an inconsistency between `composite.ConnectionDetailType` and `iov1alpha1.ConnectionDetailTypeFromConnectionDetailKey`, i.e. `FromConnectionSecretKey` vs `FromConnectionDetailKey` which brakes extracting keys from connection details of managed resource in function pipelines. 

```suggestion
			Type:                    toConnectionDetailType(dr.ConnectionDetails[i].Type),
```

where

```
func toConnectionDetailType(detailType iov1alpha1.ConnectionDetailType) ConnectionDetailType {
	switch detailType {
	case iov1alpha1.ConnectionDetailTypeFromConnectionDetailKey:
		return ConnectionDetailTypeFromConnectionSecretKey
	default:
		return ConnectionDetailType(detailType)
	}
}
```

To continue with my experiments, I applied the above patch which may or may not what we want to do to fix the problem here. I would prefer being consistent with existing semantics but can also understand if we want to use this as an opportunity to use a better naming.

internal/controller/apiextensions/composite/composition_ptf.go
```diff
@@ -0,0 +1,993 @@
...
+// composed resources in order to extract XR connection details.
+func NewConnectionDetailsObserver(e ConnectionDetailsExtractor) *ConnectionDetailsObserver {
+	return &ConnectionDetailsObserver{details: e}
+}
+
+// ObserveComposedResources to extract XR connection details.
+func (o *ConnectionDetailsObserver) ObserveComposedResources(ctx context.Context, s *PTFCompositionState) error {
+	for _, cd := range s.ComposedResources {
+		ecfgs := append(ExtractConfigsFromTemplate(cd.Template), ExtractConfigsFromDesired(cd.Desired)...)
+		e, err := o.details.ExtractConnection(cd.Resource, cd.ConnectionDetails, ecfgs...)
```
turkenh, line 978: I am trying to set some connection secret keys using `FromFieldPath` and noticed that what we have as `cd.Resource` is what we desired rather than the observed state. The problem with this is, I cannot use a path that is filled by the controllers, e.g. atProvider, or resolved references.

internal/controller/apiextensions/composite/connection.go
negz: > I would prefer being consistent with the existing API

On reflection I'm inclined to agree. Pushed add67cbb to address this.

internal/controller/apiextensions/composite/connection.go
```diff
@@ -186,3 +168,188 @@ func (c *SecretStoreConnectionDetailsConfigurator) Configure(ctx context.Context
...
+			Value:                   dr.ConnectionDetails[i].Value,
+			FromConnectionDetailKey: dr.ConnectionDetails[i].FromConnectionDetailKey,
+			FromFieldPath:           dr.ConnectionDetails[i].FromFieldPath,
+		}
+
+		if dr.ConnectionDetails[i].Name != nil {
+			out[i].Name = *dr.ConnectionDetails[i].Name
+		}
+
+		if out[i].Type == ConnectionDetailTypeFromConnectionSecretKey && out[i].FromConnectionDetailKey != nil {
```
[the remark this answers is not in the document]
negz: You're right - good catch. I think I forgot to include a `continue` in the condition above.

I pushed 661528aa to fix this. It also adds a bunch of tests that I noticed I was missing.

internal/controller/apiextensions/composite/composition_ptf.go
negz: That should work on a subsequent reconcile though, right? Desired resources should be a superset of observed resources, because we start by loading up (existing) observed resources, then run composition to get our desired state.

I'm imagining a flow like:

1. Reconcile 1 composes the resources and applies them.
2. Reconcile 1 tries to extract connection details from a field path (or connection secret key) that doesn't yet exist.
3. Reconcile 1 finishes.
4. Reconcile 2 gets the resources composed by reconcile 1, which should now have status, conn details etc.
5. Reconcile 2 composes again, probably mostly no-ops this time.
6. Reconcile 2 tries to extract connection details, which are now there.

Am I missing something? Is the observer failing and returning an error, or never returning the connection details you're asking for?

turkenh: I tried with latest changes today and observed a completely different behavior that [FromFieldPath](https://github.com/negz/xfns/pull/1/files#diff-0d0c4c0d411d737742c3cce1e489ae1ed30923b56e4ded9217fb2a081652a30eR20) worked well but [FromConnectionDetailKey](https://github.com/negz/xfns/pull/1/files#diff-0d0c4c0d411d737742c3cce1e489ae1ed30923b56e4ded9217fb2a081652a30eR24) didn't. 

Unfortunately, I didn't have time to spent more time to dig this in but writing here just in case you have any idea why that could happen.

negz: @turkenh I think this is because I renamed `FromConnectionDetailKey` back to `FromConnectionSecretKey` per your suggestion in another comment thread -https://github.com/crossplane/crossplane/pull/3465#discussion_r1081968598.

turkenh (approved):
Awesome work @negz; I believe this will unlock a lot of innovation and take composing infrastructure experience to a whole new level 🚀 

I didn't have time to test functionality when the feature was not enabled rather mostly focused on the new functionality. It would be great if we could make sure that things working as before when the flag is not enabled if we haven't done so.

negz:
I just tested that the network policy works, using a Python Composition Function that annotates all composed resources with a quote from https://api.quotable.io/random - https://github.com/negz/xfns/pull/2/files.

By default I see the following warning emitted by the function when I describe my XR:

```console
  Warning  ComposeResources         9m34s               defined/compositeresourcedefinition.apiextensions.crossplane.io  Cannot get quote: HTTPSConnectionPool(host='api.quotable.io', port=443): Max retries exceeded with url: /random (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0xffff90f3f880>: Failed to establish a new connection: [Errno -3] Temporary failure in name resolution'))
```

However when I update the function to use network policy `Runner` (i.e. access to the function runner's network) I see my composed resources being annotated. 😄 

```yaml
functions:
- name: quotable
  type: Container
  container:
    image: negz/xfn-quotable:latest
    network:
      policy: Runner
```

```yaml
$ k get -o yaml cluster.container|head -n11
apiVersion: v1
items:
- apiVersion: container.gcp.upbound.io/v1beta1
  kind: Cluster
  metadata:
    annotations:
      crossplane.io/composition-resource-name: gke-cluster
      crossplane.io/external-name: platform-ref-gcp-cluster-d2qgn-2g4nq
      quotable.io/author: Only put off until tomorrow what you are willing to die
        having left undone.
      quotable.io/quote: Pablo Picasso
```

negz:
> I didn't have time to test functionality when the feature was not enabled rather mostly focused on the new functionality. It would be great if we could make sure that things working as before when the flag is not enabled if we haven't done so.

I completely agree. I've run some cursory tests (i.e. using https://marketplace.upbound.io/configurations/upbound/platform-ref-gcp/v0.3.0 and making sure everything works normally). I think we need more, but I'm inclined to merge this PR so that we can ask the community to make sure everything still works in master before we cut a release in a little over a week.

internal/oci/store/overlay/store_overlay.go
negz: That makes sense. I'm thinking in a followup PR it may be useful to add a flag to control this. That way we could default to the current behaviour, but also have a `--use-overlay-cache` (or something) that would actually return an error if it wasn't supported. This way folks could try force the overlay cache to find out whether it was going to work.

internal/xfn/container_linux.go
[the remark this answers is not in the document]
negz: Unfortunately it didn't seem to appease gosec when I applied it here. I did go ahead and apply it to the user supplied input that we get from kong arg parsing though - https://github.com/crossplane/crossplane/pull/3465/commits/c259fd7edb5853ce9e4ab9646d8f91907ad7247d.

PR #3470, opened 2022-11-28
fix all broken links in install-configure

csantanapr:

- Fixed all links inside a tab using reference 
- Removed dead references at the bottom.

I have:

I tested with hugo using the new content 

### DIFF
--- a/docs/getting-started/install-configure.md
+++ b/docs/getting-started/install-configure.md
@@ -123,7 +123,7 @@ kubectl get all -n crossplane-system
 ## Install Crossplane CLI
 
 The Crossplane CLI extends `kubectl` with functionality to build, push, and
-install [Crossplane packages]:
+install [Crossplane packages]({{<ref "../concepts/packages" >}}):
 
 {{% tabs "crossplane CLI" %}}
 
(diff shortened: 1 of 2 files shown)


plumbis (approved):

docs/getting-started/install-configure.md
```diff
@@ -460,23 +460,9 @@ Slack][Slack] and our community will highlight it here!</i>
   and Azure mentionded in this guide.
 
 <!-- Named Links -->
-
-[package]: {{<ref "../concepts/packages" >}}
-[provision infrastructure]: {{<ref "provision-infrastructure" >}}
-[create a configuration]: {{<ref "create-configuration" >}}
 [Install]: {{<ref "../reference/install" >}}
```
hasheddan, line 463: @plumbis it looks like we are now going to be in a place where we have mixed named and inline links. Can you share why `[package]` does not work, but something like `[Install]` does?

plumbis: This was a new one to me. The problem is a result of how Hugo renders shortcodes. The broken links are inside a `tabs` shortcode.

Hugo renders shortcodes from the inside to the outside. The contents of the shortcode are fully rendered and the result is passed back to the original page. 

With named links, the link inside a shortcode doesn't exist. Only the text `[package]`. The Markdown engine has no idea that`[package]` is actually a link. The named link context only exists on the parent page. 

The solution is to not use named links in shortcodes and put the link inline, like the PR does. I would vote to discourage named links across the docs not only because of this problem but it can also create unintended errors if a link name is accidentally reused.

hasheddan: Thanks for the explanation @plumbis! I am fine with either solution, but would advocate for consistency as well 👍🏻 I don't think we need to block this important update on that though.

hasheddan (approved):

PR #3472, opened 2022-11-29
Move all Patch and Transform logic to the XR controller

negz:

This PR moves all Patch and Transform logic to the composite resource controller. Previously this logic was all defined as methods on the API types. I (think I?) prefer defining this logic in the XR controller because:

* It's the only consumer of the logic, which it uses to compose resources.
* Future Composition logic (e.g. logic to invoke Composition Functions) will likely be too complex to be defined as methods on the API types, and would thus be implemented in the XR controller.
* I generally prefer to keep "business logic" in a separate package from API types, e.g. so that a hypothetical client of Compositions could import the apiextensions/v1 types without having to (or indeed being allowed to) also bring in all the Composition logic.
* I'm (very?) slightly hopeful that decoupling this logic from the API types will help folks remember they need to update more than one package to add new Composition functionality (e.g. they'll need to update the CompositionRevision types too).

I have:

I've tested this using https://github.com/upbound/platform-ref-gcp. This doesn't exercise every single Composition feature, but given this was mostly just moving code I have reasonable confidence everything is working as it was.


muvaf:
@negz I agree with your points. One thing to consider though is that these functions may be useful for folks implementing composition functions (if they use `go`), so keeping them under `internal` would prevent that from happening. On the other hand, I don't expect the composite functions to include much patching functionality since now it's the next step after all the patching is done. But it's worth considering whether there is some logic that can be generalized and moved to `crossplane-runtime`.

negz:
> One thing to consider though is that these functions may be useful for folks implementing composition functions (if they use go), so keeping them under internal would prevent that from happening.

That's a good point. Even so, I'd prefer to move them to internal for now so that we can be more conscious about offering them for use in building composition functions in future (maybe that would look like moving them to runtime, or a new SDK of some kind).

internal/controller/apiextensions/composite/patches.go
```diff
@@ -0,0 +1,308 @@
...
+	}
+	return true
+}
+
+// ResolveTransforms applies a list of transforms to a patch value.
+func ResolveTransforms(c v1.Patch, input any) (any, error) {
+	var err error
+	for i, t := range c.Transforms {
+		if input, err = Resolve(t, input); err != nil {
+			// TODO(negz): Including the type might help find the offending transform faster.
```
turkenh, line 86: @ezgidemirel fyi, I believe @negz preferred to leave this as a to-do to keep the PR focused, but this would be a DX improvement to identify composition errors.

turkenh (approved):

ezgidemirel (approved):

Issue #3461, opened 2022-11-23
Move docs out of the crossplane/crossplane repo


plumbis:
### What problem are you facing?
There has been a long standing policy of having docs next to code.  As a result the /docs folder contains the docs for that release. After a release is cut, updates to the docs requires a backport patch to the specific branch.

This creates four main negative side effects:
1. There is no (simple) method to do local development of documentation and see styling of a docs contribution.
2. Docs updates across multiple versions are extremely inconvenient. The docs differ enough across releases it's not safe to copy/paste files across branches and it requires manual updates and multiple PRs.
3. No way to see a rendered preview of the PR. 
4. It's not obvious that docs shouldn't be updated on the website repo. For example website [issue #208](https://github.com/crossplane/crossplane.github.io/pull/208)

### How could Crossplane help solve your problem?
I propose moving the docs out of this repo and having the [docs website repo](https://github.com/crossplane/crossplane.github.io) act as the source of truth. 

### How does this address the negative issues mentioned?
1. Cloning the website repo and editing is now self-contained. Docs updates are created, tested and committed to the same repo.
2. All of the docs would live in a single parent directory with versions acting as a single sub-directory. Making changes across all versions of docs could be contained in a single commit and PR
3. With the imminent [move to Netlify](https://github.com/crossplane/crossplane.github.io/issues/98) will give the ability to have [deploy previews](https://docs.netlify.com/site-deploys/deploy-previews/). Any PR will generate a unique staging URL for the change that can be viewed by reviewers. This includes [PR feedback and markup](https://www.netlify.com/blog/2021/05/19/give-meaningful-feedback-with-collaborative-deploy-previews/) against the page contents.
4. There is only as single "docs" location. 

### Why Now?
The motivation for this is primarily driven by the value in doc relate PR review that Netlify will give us. Crossplane also now has a docs maintainer to manage and focus on documentation. 

Other projects break this out including:
* [Kubernetes](https://github.com/kubernetes/website)
* [Helm](https://github.com/helm/helm-www)
* [etcd](https://github.com/etcd-io/website)
* [Prometheus](https://github.com/prometheus/docs)

### Downsides?
* Releases will require work in another repo. This can be coordinated with the docs maintainer(s) or an automated PR can be issued to the docs repo.
* It won't be obvious where docs are for people looking at the source. We will need to clearly state in the README that docs are in a different repo.
* Changes to the `make` setup will be required to remove the docs component.

hasheddan:
[...]

PR #3486, opened 2022-11-30
Update master docs content to conform to new docs.crossplane.io

plumbis:
This PR updates the docs content to conform with changes required for the new docs.crossplane site deployment. 
This includes changes to front matter and tab generation. 

Functional changes include:
* replacing markdown images with image shortcode to convert images to webp and lazy load at build time.
* `docs/faqs/related_projects.md` - Fixes broken k8s link
* `Makefile` -  moves docs into `/content` instead of `/content/docs` on the website
* `docs/contributing/release-process.md` - Removes reference to "complete"
* `docs/getting-started/create-configuration.md` removes incomplete Alibaba configuration
* `docs/api-docs/_index.md` - removes references to crds.dev
* Removes links that redirected to crds.dev
* `docs/concepts/providers.md` - removes outdated provider table
* Updates multiple packages to pull from Upbound marketplace
* `docs/contributing/docs.md` massive rewrite and expansion to cover all new features and functionality of the updated docs site

**Do not merge this PR until /after/ the new docs site has been deployed.**


plumbis:
Docs Validation fails because make is rendering the old site that doesn't include one or more shortcodes that are in the new site.

hasheddan (commented):
@plumbis you should be able to adjust that here: https://github.com/crossplane/crossplane/pull/3486/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R72

plumbis:
@hasheddan the repo isn't changing. Because the new site features aren't merged yet the test fails. We don't want to update the Makefile since what's there now will be accurate _after_ the site updates. 

If you want to test manually you can pull the [`theme-update` branch](https://github.com/crossplane/crossplane.github.io/tree/theme-update).

docs/concepts/managed-resources.md
```diff
@@ -25,14 +25,8 @@ Composite Resources or XRs - not used directly. See the
...
 of Crossplane managed resources. Following is an example of a managed resource:
 
-<ul class="nav nav-tabs">
-<li class="active"><a href="#aws-tab-1" data-toggle="tab">AWS</a></li>
-<li><a href="#gcp-tab-1" data-toggle="tab">GCP</a></li>
-<li><a href="#azure-tab-1" data-toggle="tab">Azure</a></li>
-</ul>
-<br>
-<div class="tab-content">
-<div class="tab-pane fade in active" id="aws-tab-1" markdown="1">
```
hasheddan, line 35: ah so nice to get rid of this!

docs/contributing/docs.md
```diff
@@ -2,186 +2,504 @@
...
 ## Contributing
 
-The `/content` directory contains all documentation pages.  
-Each version of Crossplane software maintains a unique directory within `content`.
+Documentation content contributions are always welcome.
 
-### Front matter
-Each page contains metadata called [front matter](https://gohugo.io/content-management/front-matter/). Each page requires front matter to render.
+The Crossplane documentation source is in the [Crossplane
+repository](https://github.com/crossplane/crossplane) `/docs` directory.
```
hasheddan, line 55: This appears to be repeated above.

docs/contributing/docs.md
```diff
@@ -2,186 +2,504 @@
...
-```yaml
----
-title: "Troubleshooting Crossplane"
-weight: 610
----
-```
+Crossplane recommends using the [website
+repository](https://github.com/crossplane/crossplane.github.io) for local development.
+When a contribution is ready, submit a pull request to the [Crossplane
+repository](https://github.com/crossplane/crossplane).
```
hasheddan, line 60: Why do we recommend this? You can run the docs site from this repo as outlined in https://github.com/crossplane/crossplane/issues/3461#issuecomment-1324536727. I believe the justification there was that folks may not have `make` installed, but it is going to be installed by default in many distributions. That means they only have to run one command, which feels much simpler than jumping between multiple repositories.

docs/contributing/docs.md
plumbis: This is intentional to try and avoid a mix up between the two repos.

docs/contributing/docs.md
plumbis: >default in many distributions.

But not on mac or Windows.

hasheddan: They are eventually having to come and interact with this repo, where we ask folks to run `make` commands prior to submitting a PR. I am all for reducing any barrier to entry, but I don't think that this is actually accomplishing that goal as it requires more steps rather than less.

plumbis: Suggestion removed.

docs/contributing/docs.md
hasheddan: ah i see -- this is also slightly different from what is said above 👍🏻

docs/contributing/docs.md
hasheddan: Thanks. I feel like the current status is still pointing folks towards using the Crossplane website repo, and is not suggesting using any of the machinery that is put in place there. We have single commands that ensure you have the correct version of hugo installed, etc. and allow for very quick iteration. Is the reason for not recommending any of this in order to avoid using `make`? If I was helping someone contribute I would say:
1. Clone the Crossplane repo
2. Make changes
3. Run `make docs.run`

This will ensure you have the latest version of the docs site and are using the correct version of hugo. Is there anything we could change about this machinery to make you more inclined to suggest it?

plumbis: You're a beat ahead of me. I was just walking through this workflow to add instructions but ran into https://github.com/crossplane/crossplane/issues/3487.

If this is resolved I'll put in detailed steps on the make process.

(also our make configuration [doesn't support Windows](https://github.com/upbound/build/blob/master/makelib/common.mk#L99))

hasheddan: @plumbis opened https://github.com/crossplane/crossplane/pull/3488 to fix -- you can consider be signed up to fix any issues with improving the docs workflow in this repo since I am pushing so hard for using it :)

docs/faqs/related_projects.md
```diff
@@ -75,7 +75,7 @@ plane.
 <!-- Named Links -->
 
 [Open Service Broker]: https://www.openservicebrokerapi.org/
-[Kubernetes Service Catalog]: https://kubernetes.io/docs/concepts/extend-kubernetes/service-catalog/
+[Kubernetes Service Catalog]: https://github.com/kubernetes-retired/service-catalog
```
negz, line 78: FWIW we could completely remove this - I haven't heard anyone talk about it in a while.

(Or not, happy to keep it w/the updated link.)

negz (approved):
LGTM, modulo the ongoing thread with @hasheddan getting resolved.

I'd prefer to squash/remove all the merge commits to cleanup the commits in this PR before merging.

docs/contributing/docs.md
plumbis: Development section rewritten to cover make and local builds. https://github.com/crossplane/crossplane/pull/3486/files#diff-93426d953f03b733adbbb574ffd2717d0892773a464c0d301d1b6c9095d67b15R58

docs/faqs/related_projects.md
plumbis: I'll keep it for this PR and remove it in a future update.

hasheddan (approved):
We will squash and merge here per @plumbis request 👍🏻

Issue #3514, opened 2022-12-06
Missing match as a type of transform 


portswigger-tim:
https://github.com/crossplane/crossplane/blob/4e4f57a80dc19d24bb712acaa853a836b6963dc9/apis/apiextensions/v1/composition_transforms.go#L83

It seems that there is a missing transform: `match` 😿

jbw976:
Thanks for reporting this @portswigger-tim! Did #3515 resolve this issue for you? 🤔

portswigger-tim:
It _should_ fix the problem (in `master`) - the problem being that the CRD missing `match` caused a validation error on trying to apply a composition with it in.

PR #3515, opened 2022-12-06
fix(transforms): add match to kubebuilder enum

pysysops:
Seems that `match` is missing from the kubebuilder-generated CRD. This fixes that.


negz (approved):

portswigger-tim:
Closes #3514

PR #3521, opened 2022-12-09
Add agent-inject to crossplane itself

jonnylangefeld:
If you only set `` on the composition, then the XRD shows the following warning:
```
cannot publish connection details: cannot connect to secret store: cannot extract token: open /vault/secrets/token: no such file or directory
```

That is because the crossplane container itself didn't get the agent injected and therefore `/vault/secrets/token` wasn't written.

Fixes #

I have:

### DIFF
--- a/docs/guides/vault-as-secret-store.md
+++ b/docs/guides/vault-as-secret-store.md
@@ -149,6 +149,7 @@ cat << EOF > values.yaml
 args:
 - --enable-external-secret-stores
 customAnnotations:
+  vault.hashicorp.com/agent-inject: \"true\"
   vault.hashicorp.com/agent-inject-token: "true"
   vault.hashicorp.com/role: "crossplane"
   vault.hashicorp.com/agent-run-as-user: "65532"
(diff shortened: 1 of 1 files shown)


docs/guides/vault-as-secret-store.md
```diff
@@ -149,6 +149,7 @@ cat << EOF > values.yaml
 args:
 - --enable-external-secret-stores
 customAnnotations:
+  vault.hashicorp.com/agent-inject: \"true\"
```
hasheddan, line 152: nitpick: do we need to escape the `"` here? Just looking to stay consistent with other annotations below 🙂

hasheddan (commented):
Thanks @jonnylangefeld ! One question below

docs/guides/vault-as-secret-store.md
jonnylangefeld: Very good point. I copied and pasted from the example a few lines below where it was escaped because it was part of an `echo ""`.

Fixed!

hasheddan (approved):
Awesome! Thanks @jonnylangefeld !

PR #3529, opened 2022-12-13
feat(revision): Store composition manifest instead of fixed API

### LINKED ISSUE #3274: Ensure full compatibility between Composition and CompositionRevision

MisterMX:

Proposal for a solution that fixes #3274.

Rework the composition revision type so that is stores the manifest of the composition in its spec instead of mirroring every field.

This makes the revision type much more flexible:
* Developers no longer need to make sure composition revision fields match those in composition.
* Composition provides support for potential breaking APIs in the future, i.e. a possible composition v2.
* Composition and CompositionRevision don't need to be converted into each other, reducing the amout of conversion code.

I have:

Unit tests


negz:
I don't think we're going to make this change.

I'm hopeful that in addition to the work we've already done, any remaining pain will be resolved when we promote to v1 per https://github.com/crossplane/crossplane/issues/3715 (since everything will then be in the same package and we can reuse types).

Issue #1722, opened 2020-09-10
Proposal: Observe-only Crossplane resources


turkenh:
### What problem are you facing?

I want to get configuration/status of an external resource that is not managed by Crossplane. This is especially useful while building a composition that relies on configuration/status of an already existing external resource, where resource could be managed:

- by other tools (terraform, ansible, helm ...)
- manually created from UI or a legacy one
- by another Crossplane control plane instance
- by another Crossplane composite resource

This is quite similar to [data source's in Terraform](https://www.terraform.io/docs/configuration/data-sources.html) or [fact/info modules in Ansible](https://docs.ansible.com/ansible/devel/user_guide/playbooks_vars_facts.html).

For my specific case, I am interested in [reading status of an existing Kubernetes Service (which is managed by helm)](https://github.com/crossplane-contrib/provider-helm/issues/25) to get LoadBalancer IP so that I can report in composite resource status. Even better, [I want to provision a DNS record with Crossplane](https://github.com/crossplane/crossplane/issues/1603#issuecomment-647452551) and report URL. This is quite similar to the following terraform code:

```
data "kubernetes_service" "example" {
  metadata {
    name = "terraform-example"
  }
}

resource "aws_route53_record" "example" {
  zone_id = "${data.aws_route53_zone.k8.zone_id}"
  name    = "example"
  type    = "CNAME"
  ttl     = "300"
  records = ["${data.kubernetes_service.example.load_balancer_ingress.0.hostname}"]
}

``` 

However, I feel like this would enable other possibilities like fetching any kind of data from external API's without making changes, like [looking up existing AMI images](https://github.com/hashicorp/terraform/pull/4396) or getting a [list of availability zones](https://github.com/hashicorp/terraform/pull/6669).

#### Related Issues

There were some previous discussions which seems to be related to this, but proposed solutions were around claim/class pattern which is being deprecated.

- Re-using a GCP CloudSQLInstance: https://github.com/crossplane/provider-gcp/issues/157

    From [description from @dsyer](https://github.com/crossplane/provider-gcp/issues/157#issue-559105058): 

    > There are potentially 2 separate use cases here:
    > 
    > - I would like to be able to bind to a SQL instance that already exists.
    > 
    > - I would like to be able to use Crossplane to create a SQL instance, and then have it re-used by multiple claims (e.g. I deploy the same app twice).
    > 

- Crossplane should refuse to operate on existing external resources: https://github.com/crossplane/crossplane-runtime/issues/27

[...]

Issue #3516, opened 2022-12-07
Granular management policies (e.g. ability to ignore changes etc.)


timnee:
### What problem are you facing?

Crossplane provides reconciliation of all parameters under `spec`.  There are cases where this is not desired post initial creation.  Such as when a parameter is managed outside of Crossplane.  For example [AutoScalingGroup DesireCapacity](https://doc.crds.dev/github.com/upbound/provider-aws/autoscaling.aws.upbound.io/AutoscalingGroup/v1beta1@v0.21.0#spec-forProvider-desiredCapacity) could be updated during a scaling event triggered by Cloudwatch.  

[Slack thread](https://crossplane.slack.com/archives/CEG3T90A1/p1670285697486339)

### How could Crossplane help solve your problem?

Add an equivalent to [Terraform's lifecycle ignore_changes](https://developer.hashicorp.com/terraform/language/meta-arguments/lifecycle#ignore_changes) for parameters that are updated outside of Crossplane and keep it from being reconciled.  

I agree with @turkenh [suggestion](https://crossplane.slack.com/archives/CEG3T90A1/p1670356012575189?thread_ts=1670285697.486339&cid=CEG3T90A1) 
```
spec:
  ignoreChanges:
    - desiredCapacity
 ```

**Comment by linaking-guidewire** [comment]
I have a similar issue, using the same provider. We create DynamoDB [Tables](https://doc.crds.dev/github.com/upbound/provider-aws/dynamodb.aws.upbound.io/Table/v1beta1@v0.24.0) with autoscaling ([Policy](https://doc.crds.dev/github.com/upbound/provider-aws/appautoscaling.aws.upbound.io/Policy/v1beta1@v0.24.0) + [Target](https://doc.crds.dev/github.com/upbound/provider-aws/appautoscaling.aws.upbound.io/Target/v1beta1@v0.24.0)). The autoscaler scales the read and write capacity, and then Crossplane resets it back to the `readCapacity` and `writeCapacity` defined on the Table, which effectively makes autoscaling useless.

@turkenh's proposed solution would solve this case as well.
```
spec:
  ignoreChanges:
    - readCapacity
    - writeCapacity
```

turkenh:
Another use case from [slack](https://crossplane.slack.com/archives/CEG3T90A1/p1674744295223779):

> I am creating nodegroup for eks with the below specs scalingConfig:
>  - minSize: 1
>    maxSize: 5
>    desiredSize: 1
>
[...]

PR #3531, opened 2022-12-14
Design Doc for Observe Only Resources

turkenh:

This PR extends [the initial proposal](https://github.com/crossplane/crossplane/issues/1722) for Observe Only resources with a design doc.

I have:

N/A 


design/design-doc-observe-only-resources.md
```diff
@@ -0,0 +1,293 @@
...
+This way, we will also err on the side of caution by leaving the actual resource. However, this comes at the cost of
+some possible confusion for users focusing on only one of the policies.
+
+Another solution could be simply throwing an error and preventing reconciliation during conflict. This would be more
+explicit but would require some manual actions and degraded UX for the usage of the feature, for example:
+
+- Creating an ObserveOnly resource will require both setting `managementPolicy` to `ObserveOnly` and `deletionPolicy`
+to `Orphan` .
+- If there are existing resources with `deletionPolicy: Orphan` when the feature is enabled, they will start failing to
+reconcile until their `managementPolicy`’s updated to `ObserveCreateUpdate`.
```
jeanduplessis, line 149: This section starts off by stating what the behavior will be when there is a conflicting case but then goes into an alternative solution, leaving the reader unclear as to what the actual recommendation of the design doc is. 

Propose that it is reframed so that it is clear which approach the design doc puts forward and lists the other as something that was given consideration but not chosen.

turkenh: I'll reframe accordingly and make it clear that we will go with _not deleting without throwing an error_  when there is a conflict.

design/design-doc-observe-only-resources.md
```diff
@@ -0,0 +1,293 @@
...
+as of Kubernetes 1.25. See the following diff for the required changes we need for making [CIDRBlock] parameter of
+AWS VPC required only if not `ObserveOnly`:
+
+```diff
+		// CIDRBlock is the IPv4 network range for the VPC, in CIDR notation. For
+        // example, 10.0.0.0/16.
+-       // +kubebuilder:validation:Required
+        // +immutable
+-       CIDRBlock string `json:"cidrBlock"`
++       CIDRBlock *string `json:"cidrBlock,omitempty"`
```
bobh66, line 170: This appears to indicate that in order to get the most benefit from this feature either k8s 1.25 or higher is required, or the validation rules feature gate needs to be enabled on k8s <= 1.24.  I don't think the feature is nearly as useful if the resource required fields are still required (e.g. VPC cidrBlock, Cluster resourcesVpcConfig).

It also means a lot of work for the (existing, manual) providers to rework the resources with these changes.

Does upjet provide a way to implement this automatically across all resources?  That may be another reason to move to the generated providers if they automatically support observe-only resources.

bobh66 (commented):
Overall I think this looks good.  My main concern is the dependency on CEL and the changes required in the providers to make the feature useful.  I don't know of any alternatives, but I think it will be important to clearly specify what the base requirements are and whether the provider resource(s) have been modified to support ObserveOnly without the required fields being required.  We have found that importing existing resources is not very useful if we have to specify the required fields, and I would expect the same for this feature.  It's good that CEL provides a way around this, but I can see it being confusing for people given the k8s feature dependency and the changes required in the providers.

design/design-doc-observe-only-resources.md
```diff
@@ -0,0 +1,293 @@
...
+apiVersion: ec2.aws.crossplane.io/v1beta1
+kind: VPC
+metadata:
+  annotations:
+    crossplane.io/external-name: vpc-12345678
+  name: observe-vpc
+spec:
+  managementPolicy: ObserveOnly
+  forProvider:
+    region: us-east-1
```
negz, line 85: I think there's a lot going for this implementation where we use the same CRs whether something is observe only or not, as opposed to introducing new CRs:

* Doing so should make migration smoother given that you update an existing CR to do so rather than replacing an existing CR with a new one.
* Existing cross-resource-references keep working.
* We avoid roughly doubling the number of CRDs we add to a system.

I will note that I don't think the observe only -> full control migration would be _dramatically_ worse if a different type was involved, and that I think we need to deprecate cross resource references in their current form regardless.

That all said, I'm struggling to get past the fact that `spec.forProvider` and `status.atProvider`, and arguably any distinction between `spec` and `status` just don't make sense when the management policy is configurable. Why make desired state (`forProvider`) distinct from observed state when desired state can be toggled to actually be observed state?

negz: Throwing out a possibly half-baked option that I haven't seen discussed.

Back when we designed the current managed resource schema (introducing `forProvider` etc) @muvaf proposed that `status.atProvider` in fact be a superset of `spec.forProvider`. That is:

* `spec.forProvider` include only fields that a human may configure.
* `status.atProvider` include all the fields of `spec.forProvider`, plus "output only" fields.

I believe @muvaf's rationale was that it's possible that our desired state (`forProvider`) can't actually be actuated, and that under today's schema you wouldn't have any way of knowing what the actual value was in that case. Say for example you have a database with 10GB of storage. You set `spec.forProvider.dbSize` to 20, but Crossplane can't make that change for some reason. Under the current schema you would have no way of knowing what the actual state of the database is - you'd only know Crossplane _wants_ it to have 20GB. If we had a corresponding `status.atProvider.dbSize` field you could see what was different.

At the time I argued against duplicating spec fields under status because it was not an 'idiomatic' Kubernetes CR pattern. I'm not sure that logic holds up anymore though - there's no precedence that I'm aware of for functionality like observe only resources either, but we still need to build it.

The reason I share all of this context is that I wonder whether duplicating spec fields under `status.atProvider` would be a good observe only UX:

* _Everything_ we can observe about a resource appears under `status.atProvider` regardless of the management policy of the resource. We update any cross resource references that currently read from `spec.forProvider` to instead read from the equivalent `status.atProvider` fields.
* If you want to specify desired state, include it in `spec.forProvider`. If you don't, don't include it. You could almost omit the `managementPolicy` under this approach, but I think it would still be useful to control the schema (requiredness) of `spec.forProvider` fields.

design/design-doc-observe-only-resources.md
```diff
@@ -0,0 +1,293 @@
...
+```go
+type ExternalClient interface {
+	Observe(ctx context.Context, mg resource.Managed) (ExternalObservation, error)
+	Create(ctx context.Context, mg resource.Managed) (ExternalCreation, error)
+	Update(ctx context.Context, mg resource.Managed) (ExternalUpdate, error)
+	Delete(ctx context.Context, mg resource.Managed) error
+}
+```
+
+We will leverage the fact that we have an already implemented Observe method for all managed resources by calling only
```
negz, line 102: Would we also need updates to the `Observe` implementations to handle observing fields that have historically been required?

design/design-doc-observe-only-resources.md
```diff
@@ -0,0 +1,293 @@
...
+
+The proposed approach here involves utilizing the same CR, hence schema, for both managing and observing resources.
+The caveat here is that some fields are required for creating the resources but not for observing. This means that it
+won’t be possible to create an observe-only resource without providing a value for any required fields, because
+the Kubernetes API checks for the presence of these fields before allowing the CR to be created.
+
+> Please note this is already the case with [importing existing managed resources] today.
+
+We will fix this by leveraging the [Common Expression Language (CEL)], which was graduated to beta (i.e. enabled by default)
+as of Kubernetes 1.25. See the following diff for the required changes we need for making [CIDRBlock] parameter of
```
negz, line 161: Would this make v1.25 the minimum supported version of Kubernetes for folks who wanted to use Crossplane? 

Typically we like to support all currently supported versions of Kubernetes (the three most recent minor versions). Would that be possible - I'm wondering if we could build a webhook that handled the CEL logic for older versions.

design/design-doc-observe-only-resources.md
```diff
@@ -0,0 +1,293 @@
...
+-       CIDRBlock string `json:"cidrBlock"`
++       CIDRBlock *string `json:"cidrBlock,omitempty"`
+
+        // The IPv6 CIDR block from the IPv6 address pool. You must also specify Ipv6Pool
+        // in the request. To let Amazon choose the IPv6 CIDR block for you, omit this
+@@ -170,6 +169,7 @@ type VPC struct {
+        metav1.TypeMeta   `json:",inline"`
+        metav1.ObjectMeta `json:"metadata,omitempty"`
+
++       // +kubebuilder:validation:XValidation:rule="self.managementPolicy == 'ObserveOnly' || has(self.forProvider.cidrBlock)",mess
```
negz, line 178: I'm excited to potentially apply this pattern to generic cross-resource references, which also need more advanced logic to specify whether a field is required or not per https://github.com/crossplane/crossplane/issues/1770#issuecomment-1227805874 (e.g. "required unless its computed by a reference").

design/design-doc-observe-only-resources.md
```diff
@@ -0,0 +1,293 @@
...
+
+
+**Option B: Defer this to the Composition layer, specifically, Compositions Functions:**
+
+- Compositions already operate as an upper layer by owning and managing the lifecycle of managed resources.
+- Querying and filtering are more like an imperative action that does not change the state of the external world and
+could be considered as part of auxiliary actions for compositing the infrastructure.
+- Authentication to the Cloud APIs is a problem that needs to be solved which is the biggest caveat of this approach.
+In the first pass of the composition functions design, even [passing sensitive configuration] to functions is not
+covered yet, and we would eventually need support for other authentication mechanisms.
```
negz, line 226: In some cases I could imagine stacking OO resources with Composition functions - e.g. have the Composition pass in a bunch of OO resources to a function which can then query them. I actually prefer this pattern because it keeps all communication with external systems flowing through cloud providers (keeping all auth etc in one place).

design/design-doc-observe-only-resources.md
turkenh: > This appears to indicate that in order to get the most benefit from this feature either k8s 1.25 or higher is required, or the validation rules feature gate needs to be enabled on k8s <= 1.24. I don't think the feature is nearly as useful if the resource required fields are still required (e.g. VPC cidrBlock, Cluster resourcesVpcConfig).

Once we do this change here for a given resource, Observe Only resources would immediately work without needing all required fields to be provided, whether CEL is available or not. What users with k8s <= 1.24 (without the alpha feature enabled) lose is that when, management policy is not ObserveOnly, they will no longer get a validation error at the API server level, during CR creation when they didn't provide a required field.

This would definitely be a UX degradation; however, it shouldn't be a huge deal assuming the reconciler will report the error from cloud API indicating that a required field is missing. 

The question is whether we can live with it until k8s 1.25 is ubiquitous or if should we consider providing a temporary solution like a validation webhook that understands CEL and apply the validation there?

 > It also means a lot of work for the (existing, manual) providers to rework the resources with these changes.

We don't have to add this support to all (manual) resources at once, but yes, all resource schemas needs to be updated accordingly.

> Does upjet provide a way to implement this automatically across all resources? That may be another reason to move to the generated providers if they automatically support observe-only resources.

Yes, I believe it would be straightforward to automate this with upjet, and we can it available for all resources at once.

design/design-doc-observe-only-resources.md
```diff
@@ -0,0 +1,293 @@
...
+
+- Referencing existing resources without managing them
+  - In your composition, you want to reference network resources like VPC and subnets that are managed by another tool
+  or team.
+- Fetching data from existing resources
+  - You need information about an existing VPC, such as its CIDR range and the subnets it contains.
+- Gradual migration of existing/legacy infrastructure to Crossplane
+  - You have existing infrastructures managed by Terraform, and you want to migrate them gradually to Crossplane.
+  - You have a legacy infrastructure that you want to migrate to Crossplane, but you want to test the managed resources
+  before taking ownership of the underlying resources.
```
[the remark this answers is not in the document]
turkenh: _Test_ not be a good word here, maybe _experiment_ could be better. 
You want to build something with Crossplane that has some dependency on the legacy infra. You don't want to take ownership of those, but you still need Crossplane resources representing them so you can reference them etc. You start with observe only, see the power of Crossplane and give full control.

design/design-doc-observe-only-resources.md
```diff
@@ -0,0 +1,293 @@
...
+  - For an existing resource, you don’t want to give some initial configuration that might override the actual
+  configuration. You want to late-initialize all fields, including the ones that would be required otherwise.
```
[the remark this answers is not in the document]
turkenh: Sure, I have a VPC created by click ops. I want to migrate it to Crossplane. 
Currently, I need to create a VPC with same configuration of your existing resource so that Crossplane does not change its configuration. This is trying to say, you create with observe only without giving any parameter, e.g. [cidrBlock](https://marketplace.upbound.io/providers/crossplane-contrib/provider-aws/v0.33.0/resources/ec2.aws.crossplane.io/VPC/v1beta1#doc:spec-forProvider-cidrBlock), verify the configuration filled in including cidrBlock, change policy to fullControl.

Realizing this is more of an issue with how we import today, where you need to provide required fields but assuming we solved that with CEL, starting with ObserveOnly gives you a chance to verify all look good before giving full control to XP.

design/design-doc-observe-only-resources.md
```diff
@@ -0,0 +1,293 @@
...
+multiple Compositions.
+
+In this document, we aim to introduce a solution to observe resources with Crossplane without taking ownership of them.
+This would allow users to integrate existing cloud resources with the Crossplane ecosystem without giving full ownership.
+
+## Goals
+
+- Introduce a way to observe existing resources without taking ownership of them in Crossplane.
+- Enable seamless integration of existing cloud resources with the Crossplane ecosystem.
+- Allow users to fetch data from existing resources without giving full ownership.
```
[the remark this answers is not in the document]
turkenh: You're right, the first one is generic enough to cover this one as well. I'll remove this.

design/design-doc-observe-only-resources.md
```diff
@@ -0,0 +1,293 @@
...
+
+### Non-goals
+
+- Partially managing a resource by observing a subset of fields.
```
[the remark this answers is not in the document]
turkenh: Yes. 
I need to add some explanation here on why this is a non-goal, but, here is some context from a related [slack discussion](https://crossplane.slack.com/archives/CEG3T90A1/p1670356012575189?thread_ts=1670285697.486339&cid=CEG3T90A1):

> I’ve put some thought into this; while it sounds like partially observing a resource initially, I believe it has a fundamental difference that is, here, we want to use those parameters during creation. Whereas we expect observe only to be a completely read-only operation that should never make any changes on the external system including the creation.

> I believe it deserves a dedicated issue in Crossplane, but, I can imagine something like the following could do the trick:

> ```yaml
> apiVersion: eks.aws.upbound.io/v1beta1
> kind: NodeGroup
> metadata:
>   name: sample-eks-ng
> spec:
>   ignoreChanges:
>     - scalingConfig[0].desiredSize
>   forProvider:
>     # other fields
>     scalingConfig:
>       - desiredSize: 1
>         maxSize: 1
>         minSize: 1
> ```

design/design-doc-observe-only-resources.md
```diff
@@ -0,0 +1,293 @@
...
+
+## Proposal
+
+To support observing resources without taking ownership, we will introduce a new spec named `managementPolicy` to the
+Managed Resources. We will also deprecate the existing `deletionPolicy` in favor of the new spec since they will be
+controlling the same behavior; that is, how should the changes on the CR affect the external cloud resource.
```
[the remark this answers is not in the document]
turkenh: I wouldn't feel comfortable with naming this as `managementPolicy` if we don't deprecate `deletionPolicy` since they are quite relevant regarding what they are doing, that is, controlling CRUD behavior on external resources.

If this would be the way we would like to go, then I would rather use a parameter like `observeOnly: true` in spec or may be an annotation, `crossplane.io/observe-only: true`. TBH, I am not as excited as current proposal with these options 🙂 

FWIW, I don't believe the logic for handling both would be too complex, and don't see a big problem with it staying around for a while.

design/design-doc-observe-only-resources.md
```diff
@@ -0,0 +1,293 @@
...
+
+- `FullControl`(Default): Crossplane will fully manage and control the external resource, including deletion when the CR
+is deleted (same as `deletionPolicy: Delete`).
+- `ObserveCreateUpdate`: Crossplane will observe and perform create and update operations on the external resource but
+will not make any deletions (same as `deletionPolicy: Orphan`).
+- `ObserveOnly`: Crossplane will only observe the external resource and will not make any changes or deletions.
+
+Please note while `ObserveCreateUpdate` may sound verbose, it accurately describes the actions that Crossplane will take
+on the external resource. This naming convention also focuses on what Crossplane will do, rather than what it won't do,
+making it more intuitive and extensible for future policy options, such as `ObserveDelete`.
```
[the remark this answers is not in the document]
turkenh: No, I agree.
I started with `managementPolicy: Orphan` but immediately noticed that it sounds like we will leave external resources as orphaned by not **_managing_** it. 
So, open to suggestions here, but if possible, would still like to stick to the convention of focusing on what Crossplane will do, rather than what it won't do as mentioned in the above paragraph.

design/design-doc-observe-only-resources.md
turkenh: I think we will also need to update late-init logic to include required fields as well which worth mentioning 👍 

FWIW, this could be done once instead of per resource with code generation solutions like upjet.

design/design-doc-observe-only-resources.md
turkenh: FWIW, XP will still be usable prior v1.25 at the cost of some UX degradation, as I detailed here: 

https://github.com/crossplane/crossplane/pull/3531#discussion_r1049608625

design/design-doc-observe-only-resources.md
turkenh: Do you mean option A by 'prefer this pattern' ?

design/design-doc-observe-only-resources.md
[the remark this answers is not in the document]
turkenh: Just to get it persisted here. 
As discussed in the community meeting, we may choose a proactive path and leverage a conversion webhook to drop the deletion policy by converting it to the corresponding management policy instead of leaving it as deprecated. This would require an API Version bump on affected resources (i.e., all managed resources) which we need to plan carefully, and we may consider dropping a couple of other deprecated fields altogether.

design/design-doc-observe-only-resources.md
jbw976: > it shouldn't be a huge deal assuming the reconciler will report the error from cloud API indicating that a required field is missing.

this sounds reasonable to me honestly - my initial concern was that folks wouldn't be able to run Crossplane on < 1.25 at all (i.e. Crossplane was taking a hard dependency on 1.25+), which would definitely be undesirable :)

design/design-doc-observe-only-resources.md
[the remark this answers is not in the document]
turkenh: Rephrased the statement in the doc.

design/design-doc-observe-only-resources.md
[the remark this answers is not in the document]
turkenh: Added some explanation to the doc on why this is a non-goal.

design/design-doc-observe-only-resources.md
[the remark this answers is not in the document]
turkenh: One possibility could be `managementPolicy: OrphanOnDelete` if it sounds better than `managementPolicy: ObserveCreateUpdate`.

design/design-doc-observe-only-resources.md
turkenh: I like having `status.atProvider` to include the full state. 
I cannot think of any immediate drawbacks of this approach, and makes sense considering the advantages you already have listed.

turkenh:
I have responded/resolved all the existing comments, and this is ready for another pass.

design/design-doc-observe-only-resources.md
turkenh: Hmm, one caveat is some possible friction with migration. 
Currently, the scenario is, to start with `managementPolicy: ObserveOnly`, and once you feel good, simply set `managementPolicy: Full` and expect XP automagically start managing.
If we have all observations in `status.atProvider`, the required ones need to be moved up to `spec.forProvider` in addition to changing the `managementPolicy` to `Full`.

design/design-doc-observe-only-resources.md
negz: No, something closer to option B.

However rather than a Composition Function talking directly to an external system (which I'd like to avoid as much as possible) I'm thinking that the Composition Function would "query" OO resources to derive information from them. I suppose a Composition Function pipeline could request the creation of OO resources that they need data from.

Consider for example an XR that needs to find an available subnet to use. I could imagine a pipeline of Composition Functions like:

1. Composition Function A requests an OO VPC resource be created.
2. Composition Function B reads the OO VPC resource to learn what subnets are in use, returns an available subnet.

negz:
@turkenh I like the direction this is going. I think adding a few example resources (i.e. YAML manifests) showing what OO resources would look like in practice would be useful to illustrate the idea.

design/design-doc-observe-only-resources.md
```diff
@@ -0,0 +1,293 @@
...
+When using Crossplane to manage resources, we typically create a Managed resource that represents the desired state of
+the resource at a provider. Crossplane takes ownership of the resource, starts acting as the source of truth, and
+ensures configuration matches the desired state.
+
+Sometimes, we want to "observe" an existing resource without taking ownership of it. This could be useful in several
+scenarios, which could be grouped as follows:
+
+- Referencing existing resources without managing them
+  - In your composition, you want to reference network resources like VPC and subnets that are managed by another tool
+  or team.
```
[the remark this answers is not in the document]
turkenh: I added an example to make it more clear.

design/design-doc-observe-only-resources.md
```diff
@@ -0,0 +1,293 @@
...
+- Fetching data from existing resources
+  - You need information about an existing VPC, such as its CIDR range and the subnets it contains.
```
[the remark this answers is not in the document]
turkenh: I added an example to make it more clear.

design/design-doc-observe-only-resources.md
turkenh: > Would we need a managementPolicy at all in this scenario?

`managementPolicy` would at least be needed to decide whether a field is required (with CEL), e.g. if the `managementPolicy: Full` you need to provide `spec.size`. 

I also have some concerns about moving from resource ownership to a field ownership model but I need to spend some time to materialize them with some examples.

design/design-doc-observe-only-resources.md
turkenh: I see your point, and such a flow could help for certain use cases.
However, not sure how it could help for a use case like I need to get the latest AMI image.

design/design-doc-observe-only-resources.md
turkenh: The primary concern I have with moving from _resource ownership_ to a _field ownership_ model is that it would be a bold move since it is changing something fundamental. One example consequence of such a change is, late-initialization logic would need to be revisited for all resources, and we would probably late-init to `status.atProvider` and do some merging with `spec.forProvider` before the update calls.

I rather prefer to defer it to a dedicated design, e.g. field level management, if it solves some real-world problems. So far, the only use case that could be relevant is [this issue](https://github.com/crossplane/crossplane/issues/3516), which could have more straightforward solutions like the one proposed in the issue or as you proposed [here](https://github.com/crossplane/crossplane/pull/3531#discussion_r1072945963). 

Even if we go with that route, I believe having a `managementPolicy` would be helpful since:
- Being explicit about marking a resource as Read-Only would make folks more comfortable while adopting existing resources.
- We can use it to decide whether a field is required (with CEL).

So, I would suggest continuing with the proposed design and considering moving to a more granular solution if we see such a need.

design/design-doc-observe-only-resources.md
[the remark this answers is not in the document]
turkenh: I like the `spec.initProvider` approach and I believe it could be good alternative to the one proposed in https://github.com/crossplane/crossplane/issues/3516. 

However, as I stated [here](https://github.com/crossplane/crossplane/pull/3531#discussion_r1087895081), I believe this should be tackled separately.

turkenh:
> @turkenh I like the direction this is going. I think adding a few example resources (i.e. YAML manifests) showing what OO resources would look like in practice would be useful to illustrate the idea.

@negz added some examples as you suggested and responded to your comments. Could you have another look?

design/design-doc-observe-only-resources.md
```diff
@@ -0,0 +1,515 @@
...
+This will not prevent the field from appearing in the schema of the managed
+resources. However, we will ignore the `spec.managementPolicy` when the feature
+is not enabled.
+
+#### Deprecation of `deletionPolicy`
+
+With the new `managementPolicy` covering the existing `deletionPolicy`, we will
+deprecate the latter in favor of the former.
+
+Until we drop the `deletionPolicy` from the schema altogether, we need to be
```
AlainRoy, line 163: I have a question about the future, technically out of scope of your proposal. You mention eventually dropping `deletionPolicy` from the schema. How will that affect customers who are updating to a new version of Crossplane? Will there be mechanism to help them migrate, or do they have to manually migrate from `deletionPolicy` to `managementPolicy`?

design/design-doc-observe-only-resources.md
```diff
@@ -0,0 +1,515 @@
...
+
+- `FullControl`(Default): Crossplane will fully manage and control the external
+resource, including deletion when the CR is deleted
+(same as `deletionPolicy: Delete`).
+- `ObserveCreateUpdate`: Crossplane will observe and perform create and update
+operations on the external resource but will not make any deletions
+(same as `deletionPolicy: Orphan`).
+- `ObserveOnly`: Crossplane will only observe the external resource and will not
+make any changes or deletions.
+
```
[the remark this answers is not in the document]
thephred: Or any other combination for that matter: ObserveCreate (create only), ObserveDelete (delete only) etc.

design/design-doc-observe-only-resources.md
negz: We basically can't drop it from any `v1beta1` resource. We could introduce `v1` (or `v1beta2`) and drop it there, then after a sufficient amount of time remove the beta APIs altogether.

design/design-doc-observe-only-resources.md
negz: I agree this is a big change, but if we think it's the better way to do things longer term I would rather start moving toward that than introduce an "intermediate" version of the feature . If we add a management policy that puts `spec.forProvider` into "observe only mode" that's behaviour (and potentially a few field/knob) that we'd have to deprecate and remove if we moved to including everything in `status.atProvider` eventually.

design/design-doc-observe-only-resources.md
```diff
@@ -0,0 +1,515 @@
...
+describes the actions that Crossplane will take on the external resource. This
+naming convention also focuses on what Crossplane will do, rather than what it
+won't do, making it more intuitive and extensible for future policy options,
+such as `ObserveDelete`.
+
+As indicated above, `FullControl` and `ObserveCreateUpdate` policies will behave
+precisely the same as the deletion policies we have today, including keeping the
+default behavior the same. We will introduce the new behavior with the
+`ObserveOnly` option, which would be pretty similar to what we have today to
+[import existing managed resources], but instead of starting to manage after
```
jbw976, line 102: does this also have the same requirement that import does, where in addition to `external-name` being set, all the required fields of the resource must also be set?

We often hear that is a burden for folks to use import, so will that same burden exist for observeOnly?

jbw976: nevermind, looks like this was addressed later on in the "Schema Changes" section - for k8s 1.25+, users won't have to include the required fields.  Hooray!!

design/design-doc-observe-only-resources.md
turkenh: Ok, after putting in some more thoughts, I think I can now articulate my concerns with an example.

The API without `managementPolicy` lacks information between the MR and the cloud resource despite it having that for individual fields. As an example, let's think about the reconciler logic operating on the following resource:

```yaml
apiVersion: eks.aws.crossplane.io/v1beta1
kind: Cluster
metadata:
  name: some-eks-cluster
  annotations:
    crossplane.io/external-name: some-eks-cluster
spec:
  forProvider:
    region: us-west-1
```

Our reconciler will read the MR from k8s API and try to get the EKS cluster named `some-eks-cluster` in `us-west-1` from AWS API. If it exists, it will check if all the spec matches what is provided by the user under `spec.forProvider`, and otherwise do an update call to AWS to get the desired state synced.

However, **if the cluster does NOT exist, it is not clear what to do**:
- If the user wants an EKS cluster with all the default parameters, **we need to create the cluster**.
- If the user's intent is to observe an existing EKS cluster, **we need to fail with a proper error**.

This is because we no longer know the relationship between MR and the resource itself, even though we can have it at the field level. 

The bottom line is we need to have the `managementPolicy` field to infer that relationship.

turkenh: > - If you're creating a new resource you must include the minimum required set of fields in spec.forProvider.
> - If you're observing/importing an existing resource, you can omit all fields from spec.forProvider.

Re-reading your thoughts above and I believe you have the assumption that all resources will always have at least one required field, which could indeed be used to distinguish create vs observe. However, this does not hold for most of the resources, and even if it were, it would be quite an implicit distinction.

negz: @turkenh You're right, I don't think we can assume an empty `spec.forProvider` means someone wants an observe-only resource. So we do need `managementPolicy` (or something like it).

I still feel quite strongly that it shouldn't be possible to put `spec.forProvider` into an "observe only mode" though. I would like `spec.forProvider` to _only_ ever be used for desired state, with all observed state in `status.atProvider`. If we keep the management policy, are there any outstanding issues with this approach?

I know you pointed out in https://github.com/crossplane/crossplane/pull/3531#discussion_r1068122286 that toggling from observe-only to not would require editing the MR move stuff up from `status.atProvider` to `spec.forProvider`. Presumably if we went the other way (Full -> OO) we'd have to remove stuff from `spec.atProvider` too. Do you consider this a show-stopper?

design/design-doc-observe-only-resources.md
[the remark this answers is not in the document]
bobh66: Something like ObserveCreateDelete would definitely be useful - we have a number of use cases where we want Crossplane to create a resource and then ignore any changes to it, but still delete it when requested.

design/design-doc-observe-only-resources.md
turkenh: @negz as we discussed offline, I've updated the design to have a full observed state under `status.atProvider`  https://github.com/crossplane/crossplane/pull/3531/commits/be9d560dda3aec3912d61c58fc5efbc8e93aa330

design/design-doc-observe-only-resources.md
```diff
@@ -0,0 +1,515 @@
...
+Sometimes, we want to "observe" an existing resource without taking ownership of
+it. This could be useful in several scenarios, which could be grouped as
+follows:
+
+- Referencing existing resources without managing them
+  - In your managed resource, you want to reference network resources like VPC
+  and subnets that are managed by another tool or team. _For example, you want
+  to create and RDS instance in Crossplane, but you want it to use an existing
+  Subnet Group managed by Terraform._
+- Fetching data from existing resources
```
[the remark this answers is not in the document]
turkenh: > observe a resource selected by labels

I would categorize this as querying and filtering, which is mentioned under Future Work with some possible solutions.

design/design-doc-observe-only-resources.md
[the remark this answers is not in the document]
turkenh: While not directly in the scope of this design doc, the new management policy would allow such extensions in the future whenever needed.

turkenh:
I've updated the design to reflect all the comments/discussions. 

I've also added a new section proposing an import procedure leveraging the feature and tackling the issues with the current import procedure. /cc @bobh66

design/design-doc-observe-only-resources.md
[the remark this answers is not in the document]
turkenh: Ok, I am going with `OrphanOnDelete` instead: https://github.com/crossplane/crossplane/pull/3531/commits/0a183485272df16f9109778cd55e5f8d106d1d15

design/design-doc-observe-only-resources.md
```diff
@@ -0,0 +1,703 @@
...
+- Return error if the resource does not exist.
+- Publish connection details.
+- Ignore late-initialization result and never call `client.Update` method to
+  update the resource spec.
+- Report success and return early.
+
+We will also need the following changes per resource:
+
+- Update the API schema to have all the fields under `spec.forProvider` under
+  `status.atProvider` as well.
```
negz, line 205: Would we populate these fields if the feature flag was off?

design/design-doc-observe-only-resources.md
```diff
@@ -0,0 +1,703 @@
...
+
+1. Create a new resource with `ObserveOnly` policy.
+   1. With external name annotation set to the external name of the resource to be imported.
+   2. Only provide the identifying arguments (e.g. `region`) in the spec of the
+      resource and skip all the other fields including the required ones ( which 
+      would no longer be required, see the previous section).
+2. Expect the existing resource to be observed successfully indicating that the
+   existing resource is found.
+3. Change the policy to `Full` and provide the required fields by copying them
+   from `status.atProvider` to give full control of the resource to Crossplane.
```
negz, line 315: I wonder if we could automate this.

If `spec.managementPolicy` is `Full`, `spec.forProvider` is empty, and `status.atProvider` is not Crossplane automatically copies everything (that can be changed) from `status.atProvider` to `spec.forProvider`.

Perhaps this would be too implicit/surprising?

design/design-doc-observe-only-resources.md
```diff
@@ -0,0 +1,703 @@
...
+apiVersion: ec2.aws.crossplane.io/v1beta1
+kind: VPC
+metadata:
+  annotations:
+    crossplane.io/external-name: vpc-12345678
+  name: observe-vpc
+spec:
+  managementPolicy: ObserveOnly
+  forProvider:
+    region: us-east-1
```
negz, line 120: It may be worth a small note pointing out that `spec.forProvider.region` is a bit special for OO resources. It's more an identifier than desired state.

negz (approved):
I have a few last comments/questions but largely this looks good to me. Thanks @turkenh!

design/design-doc-observe-only-resources.md
```diff
@@ -0,0 +1,703 @@
...
+| Orphan | ObserveOnly | Yes | No | No | No |
+
+For conflicting cases, we will decide based on the non-default configuration
+which means "not deleting the external resource" for all 3 conflicting cases.
+This way, we will also err on the side of caution by leaving the actual resource
+untouched, avoiding any accidental deletion or modification.
+
+> Another solution could be simply throwing an error and preventing
+> reconciliation during conflict. This would be more explicit but would require
+> some manual actions and degraded UX for the usage of the feature, for example:
```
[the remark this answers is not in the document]
turkenh: The only difference from the current proposal would be **deleting the external resource** when `DeletionPolicy: Orphan && ManagementPolicy: Full` instead of orphaning it. This could happen when someone had intentionally set `DeletionPolicy: Orphan` on existing resources and later enabled the feature flag without touching the ManagementPolicy field on existing resources (ending with its default value). I wouldn't be comfortable deleting external resources in this case and would rather orphan it obeying the non-default configuration reflecting the user's intent.

FWIW, while the table might look complicated, this is how it looks like when implemented: https://github.com/crossplane/crossplane-runtime/pull/384/files#diff-97300a2543f95f5a2ada3560bf47dd7334e237e27976574d15d1cddef2e66c01R1105

design/design-doc-observe-only-resources.md
turkenh: This is tricky since we cannot control schema changes with feature flags. 
With upjet, I know that populating them is much easier than not since we are simply unmarshaling terraform fields back to crd fields. Basically, we don't need any changes in the existing logic there.
This might be the opposite for native providers. So, I would prefer to leave this open and up to implementation.

design/design-doc-observe-only-resources.md
turkenh: Yes, I believe this could be automated as you suggested. We wouldn't even need everything but only the required fields and late-init would take care of the rest as usual. It would be great if we could even implement it at managed reconciler level (i.e. in a common place). 

> Perhaps this would be too implicit/surprising?

I don't think so. But we might be more confident after hearing some real use cases or need for this.

design/design-doc-observe-only-resources.md
turkenh: Good call, done: https://github.com/crossplane/crossplane/pull/3531/commits/d31602a09627bc44c10a33e48be5dbf08354126e

design/design-doc-observe-only-resources.md
jbw976: FWIW, this is the same conclusion we arrived at this morning when we took the OO feature for an early preview test drive :)

We first created a OO resource by specifying the external name annotation, the region, and managementPolicy of ObserveOnly.  Everything got synced successfully and status.atProvider was populated fully. We were hoping to complete the import by then just flipping managementPolicy to FullControl and have the controller take care of everything for us to update the spec to match the reality we had already observed.

So +1 data point for expectations being that the controller would automate this for us 🤪

bobh66:
One other scenario just for posterity - AWS creates a Security Group when it creates an EKS Cluster, and for some reason it will neglect to delete the SG when the Cluster is deleted, which blocks the VPC deletion until the SG is manually deleted.  It would be nice to be able to ObserveDelete the SG to prevent this blocking scenario.

ytsarev:
@bobh66 interesting, we do not hit it in https://github.com/upbound/platform-ref-aws , vpc is getting properly cleaned

design/design-doc-observe-only-resources.md
turkenh: Opened an issue to track this: https://github.com/crossplane/crossplane/issues/3999

PR #3550, opened 2022-12-28
Validate Composition and configure XR before fetching environment

negz:

This way if our Composition is invalid or we fail to configure the XR (often using the Composition) we'll fail early without hitting the environment select/fetch logic unnecessarily.

I have:

I haven't tested this, so I'm opening it as a draft to get feedback from folks who know the implementation better before I do. @MisterMX @ytsarev do you think this makes sense?


ytsarev:
@negz sorry, I didn't manage to look into this one before. Does it mean that https://github.com/crossplane/crossplane/pull/3465 inherently implements the above?

PR #3553, opened 2022-12-30
fuzzing: run fuzzers in CI

AdamKorcz:
Signed-off-by: AdamKorcz <adam@adalogics.com>

Add [CIFuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration/) workflow action to have fuzzers build and run on each PR.

This is a service offered by OSS-Fuzz where Crossplane already runs. CIFuzz can help detect regressions, bring fuzzing closer into the development workflow, and catch fuzzing build issues early. CIFuzz has a variety of features (see the URL above). In the current PR the fuzzers get built on a pull request and will run for 600 seconds.

Fixes #

I have:


negz:
@AdamKorcz is it possible to make fuzzing a job under the existing CI workflow?

negz (changes requested):
Just reiterating my existing request - unless there's a reason to prefer a new workflow I'd rather this be a new job within the existing CI workflow.

AdamKorcz:
> @AdamKorcz is it possible to make fuzzing a job under the existing CI workflow?

It is possible to run the fuzzers with `go test -fuzz=FuzzTest` which would run the fuzzers for a specified amount of time. CIFuzz has the following advantages:

- OSS-Fuzz uses its corpus from hundreds of hours in the CI. 
- OSS-Fuzz checks if found crashes in the CIFuzz are already known - thereby avoiding reporting an issue in a PR that is already in the bug tracker.
- CIFuzz is a dedicated tool for running the fuzzers in the CI and the OSS-Fuzz team are working continuously on making it smart for that purpose.

phisco:
@AdamKorcz, @negz meant if it was possible to move the `Fuzzing` job from a the file `cifuzz.yml` to be just one of the jobs in `.github/workflows/ci.yml`. Maybe between [unit-tests](https://github.com/crossplane/crossplane/blob/master/.github/workflows/ci.yml#L123) and [e2e-tests](https://github.com/crossplane/crossplane/blob/master/.github/workflows/ci.yml#L172)

negz:
> @negz meant if it was possible to move the Fuzzing job from the file cifuzz.yml to be just one of the jobs in .github/workflows/ci.yml. 

Right - these all run concurrently. It doesn't make much practical difference except that I like having everything CI related under the CI workflow.

AdamKorcz:
Sorry for the confusion. I have added the fuzzing job at the end of `ci.yml` now.

phisco (approved):
Maybe I'd lowercase the job's name, but that's fine

negz:
> Maybe I'd lowercase the job's name, but that's fine

That would be my preference, but I happen to be touching this file in https://github.com/crossplane/crossplane/pull/3644 so I can add a small cleanup commit there to do this.

negz (approved):

negz:
I notice fuzzing was quite slow - 25 minutes. I'd like to bias for merging this but it would be good to follow-up with anything we can do to cut down that time.

PR #3556, opened 2023-01-02
feat: Added commonLabel to label resources applied via Configuration

### LINKED ISSUE #3555: commonLables in configuration.pkg for Composition Revisions Labels
we want to add commonLables in configuration.pkg, that can be used for composition revisions lables, in a automatic way
example:

```
---
kind: Configuration
metadata:
  name: test
spec:
  package: sop:v1.193.0
  commonLabels:
    version: v1.193.0
    channel: beta
```

when using ci and gitops tooling to package and deploy configuration packages it is not an option to update labels manually - nor is it feasible to use calculated hashes from compositionrevisions

akesser:
Signed-off-by: André Kesser <andre.kesser@dkb.de>

Added commonLabel property to Configuration.pkg to label applied resources. The idea is to be able to add labels that can later be used by compositionRevisionSelector -> matchLabels

Fixes #3555 

I have:

We tested the commonLabel attribute by applying a configuration with this attribute on a local kind cluster and observed the applied objects. We testet to create a new configuration with commonLabels, update a configuration without this attribute to add commonLabels and we tested to remove single and all commonLabels from a configuration.

Example: 
```yaml
kind: Configuration
metadata:
  name: test
spec:
  commonLabels:
    revision: v0.3.0
    test: "4"
  package: xpkg.upbound.io/crossplane-contrib/x-generation:v0.3.0
  revisionActivationPolicy: Automatic
  revisionHistoryLimit: 1
  skipDependencyResolution: true
```
With these compositions:
```yaml
kind: Composition
metadata:
  labels:
    example.cloud/provider: example
  name: compositeinstanceprofile.iam.aws.example.cloud
...
---
kind: Composition
metadata:
  labels:
    example.cloud/provider: example
  name: compositealias.kms.aws.example.cloud
...
```
leads to 
```yaml
kind: Composition
metadata:
  labels:
    example.cloud/provider: example
    revision: v0.3.0
    test: "4"
  name: compositeinstanceprofile.iam.aws.example.cloud
...
---
kind: Composition
metadata:
  labels:
    example.cloud/provider: example
    revision: v0.3.0
    test: "4"
  name: compositealias.kms.aws.example.cloud
...
```
etc.


internal/controller/pkg/revision/reconciler.go
```diff
@@ -570,7 +572,24 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+		}
+		labels := d.GetLabels()
+		if labels != nil {
+			for key, value := range commonLabels {
+				labels[key] = value
+			}
+		} else {
+			d.SetLabels(commonLabels)
+		}
+	}
```
hasheddan, line 591: Any reason not to do this as part of the establisher?

akesser: I moved the labeling not a method of the establisher

ezgidemirel:
Hello @akesser , why do we need to add `commonLabels` field to CRDs other than configurations? My understanding was you want to propagate some labels to the resources created with a configuration. Am I missing something?

akesser:
@ezgidemirel We need commonLabels on the resources to be able to use labelSelectors for CompositionRevisions. With labels compositionRevisionSelector.matchLabels can be used to select the compositions in a Claim so developer do not need to look up the calculated revision hash that needs to be used otherwise.

haarchri:
@ezgidemirel  what you mean with with fields to CRDs other than configurations ? we tested configurations and added lables to crds - what "issue" you facing ?

akesser:
Ah ok, we want commonLabel in ConfigurationSpec. ConfigurationSpec is the same as PackageSpec, from the Configuration/Package, a ProviderRevision object gets applied. The PackageRevision is then used in its own Reconciler to create the objects, therefore ProviderRevision needs the commonLabel attribute too. Providers share the same interface with Configuration called Package, ProviderRevisions share the same interface with ConfigurationRevison, called PackageRevison, therefore the labels appear there too.

ezgidemirel:
Thanks @akesser for the explanation! The PR looks good to me. I think it's a good idea to have a common set of labels on resources that are coming from a configuration. It would definitely be useful with compositionSelectors as well.

ezgidemirel (approved):

turkenh:
@akesser could you add some details under `How has this code been tested` section?

akesser:
@turkenh I added details and Examples to the section

turkenh (approved):

PR #3568, opened 2023-01-11
publish to upbound marketplace

### LINKED ISSUE #3560: Publish Crossplane image to xpkg.upbound.io

A few months ago, many of the packages in the Crossplane ecosystem moved from Dockerhub to `xpkg.upbound.io` due to a few motivating factors:
- Rate limiting on pulls from Dockerhub.
- Automatic documentation support for all API types in a package on marketplace.upbound.io.
- Ability to search and discover packages based on their metadata on marketplace.upbound.io.

Crossplane itself is published as an OCI image and Helm chart, with the latter referencing the former. This differs from the structure of Crossplane packages, but there are a few commonalities, namely:
- Crossplane bundles its API types in its OCI image.
- Pulling the Crossplane image is still subsequent to rate limiting as imposed by whatever registry it is hosted on.

Moving the Crossplane image to be hosted on xpkg.upbound.io would ensure that the community is not negatively impacted by any rate limiting, and would create the opportunity for the Crossplane API documentation to live alongside Crossplane packages (and similarly be automatically updated when new versions are published).

Moving the image would likely have very minimal impact on existing users, as users install Crossplane by downloading the Helm chart, which would be updated to reference the new location by default. There are two cases where a user could potentially have difficulty in upgrading:
- They have Dockerhub on an allow list in their cluster and do not have xpkg.upbound.io on the list.
- They are automatically syncing Crossplane versions from Dockerhub to an internal registry and then accessing the internal registry from their cluster.

Both of these issues could be mitigated by proper community messaging and a reasonable transition window. I would propose that we dual publish to Dockerhub and xpkg.upbound.io and update to point at xpkg.upbound.io in the Crossplane helm chart by default for the next release (`v1.11.0`). This would be accompanied by messaging in release notes and Slack prior to the release. At the next release (`v1.12.0`), Dockerhub could be dropped as an official host. This would give a reasonable window for users that need to update any internal systems to pull from xpkg.upbound.io a full release cycle to do so (they would be required to upgrade with `docker.io` specified in the values file for the mean time), while going ahead and getting the vast majority of the community over to xpkg.upbound.io on `v1.11.0`.

**Amendment**

Instead of moving the default in the Helm chart in `v1.11.0`, we will add a note in the `v1.11.0` release notes that the default will be changing in `v1.12.0`, then stop publishing to Dockerhub and update the default as part of the `v1.12.0` release.

nullable-eth:
Signed-off-by: nullable.eth <2248325+nullable-eth@users.noreply.github.com>

Add xpkg.upbound.io to `REGISTRY_ORGS` to dual publish to Upbound Marketplace along with Docker Hub.

Fixes #3560 

I have:


Makefile
```diff
@@ -59,11 +59,19 @@ HELM_CHART_LINT_ARGS_crossplane = --set nameOverride='',imagePullSecrets=''
...
 -include build/makelib/imagelight.mk
 
+# ====================================================================================
+# Setup Docs
+
+SOURCE_DOCS_DIR = docs
+DEST_DOCS_DIR = content
+DOCS_GIT_REPO = https://$(DOCS_GIT_USR):$(DOCS_GIT_PSW)@github.com/crossplane/docs.git
+-include build/makelib/docs.mk
+
```
hasheddan, line 74: @nullable-eth looks like some of the recently removed docs functionality got re-added here. A rebase should fix it 👍🏻

nullable-eth: Sorry my fork was not up to date.

hasheddan (commented):
Thanks @nullable-eth! Can you update testing sections with your steps after completing?

.github/workflows/ci.yml
```diff
@@ -302,9 +302,17 @@ jobs:
...
-      - name: Publish Artifacts to S3, DockerHub
+      - name: Login to Upbound
+        uses: docker/login-action@v1
+        if: env.UPBOUND_MARKETPLACE_PUSH_ROBOT_USR != ''
+        with:
+          registry: xpkg.upbound.io
+          username: ${{ secrets.UPBOUND_MARKETPLACE_PUSH_ROBOT_USR }}
+          password: ${{ secrets.UPBOUND_MARKETPLACE_PUSH_ROBOT_PSW }}
+
+      - name: Publish Artifacts to S3, Marketplace, DockerHub
```
negz, line 313: Do we need to do any promoting (i.e. the next step in this job)?

hasheddan: Since `xpkg.upbound.io/crossplane` is added to `REGISTRY_ORGS` it will be automatically included in the promotion stage as well. However, mirroring the check for the env var in the step could make sense for consistency.

nullable-eth: Added the check

hasheddan (approved):
Thanks @nullable-eth! We should be able to go ahead and merge here since the checks are added, then re-run CI once the necessary secrets are added to the repo. @negz or @jbw976 could one of y'all take care of that?

Makefile
```diff
@@ -59,7 +59,7 @@ HELM_CHART_LINT_ARGS_crossplane = --set nameOverride='',imagePullSecrets=''
 # Due to the way that the shared build logic works, images should
 # all be in folders at the same level (no additional levels of nesting).
 
-REGISTRY_ORGS = docker.io/crossplane
+REGISTRY_ORGS = docker.io/crossplane  xpkg.upbound.io/crossplane
```
hasheddan, line 62: @nullable-eth is this supposed to be a double space here?
```suggestion
REGISTRY_ORGS = docker.io/crossplane xpkg.upbound.io/crossplane
```

.github/workflows/ci.yml
```diff
@@ -12,7 +12,7 @@ env:
   # Common versions
   GO_VERSION: '1.19'
   GOLANGCI_VERSION: 'v1.49'
-  DOCKER_BUILDX_VERSION: 'v0.8.2'
+  DOCKER_BUILDX_VERSION: 'v0.10.0'
 
   # Common users. We can't run a step 'if secrets.AWS_USR != ""' but we can run
   # a step 'if env.AWS_USR' != ""', so we copy these to succinctly test whether
```
hasheddan, line 18: Will need to add this for checks to work:
`UPBOUND_MARKETPLACE_PUSH_ROBOT_USR: ${{ secrets.UPBOUND_MARKETPLACE_PUSH_ROBOT_USR }}`

hasheddan (commented):
@nullable-eth there are a few more updates needed, including making the same changes in https://github.com/crossplane/crossplane/blob/master/.github/workflows/promote.yml

Makefile
nullable-eth: fixed

nullable-eth:
> @nullable-eth there are a few more updates needed, including making the same changes in https://github.com/crossplane/crossplane/blob/master/.github/workflows/promote.yml

I've made all the necessary changes for ENV and to the Promote flow.

hasheddan (approved):
Thanks @nullable-eth!

Issue #3047, opened 2022-04-18
Release manager program


negz:
### What problem are you facing?

https://github.com/crossplane/crossplane/blob/7d01448/docs/contributing/release-process.md

We currently have a fairly well documented release schedule and process, but folks really only think about this toward the end of a release cycle. We typically raise a release tracking issue early, but then find ourselves soliciting a volunteer to cut the release close to when it is due. We're missing "end to end ownership" of a release, including:

* Creating a project for each release and keeping it up to date.
* Keeping the community up-to-date with the state of the release at each community meeting.
* Following up as we close on code freeze to determine what will ship and what will slip.
* Running the release process.
* Owning backport (patch) releases?
* Proactively improving the release process and automation.

### How could Crossplane help solve your problem?

We might consider instituting a release manager position - I imagine this would be similar to what Kubernetes does but lighter weight. Such a position would be responsible for the things I listed above for the scope of a single release. I believe this could be a good way to avoid "bystander syndrome" while also avoiding relying on a small group of folks to run every release. Finally, it's a good way to get folks involved in the project without them needing to write code.

negz:
CC @jbw976 @hasheddan who I've discussed this with in the past.

jeanduplessis:
Its 2024, and here's the state of the world today.

We usually assign a maintainer to own doing the release of Crossplane.

> Creating a project for each release and keeping it up to date.
We have a GitHub project for roadmap, release issues, etc. https://github.com/orgs/crossplane/projects/20/views/1?query=sort:updated-desc+is:open

> Keeping the community up-to-date with the state of the release at each community meeting.
We have a section in the XP Community meeting agenda that covers the latest updates in a release cycle: https://docs.google.com/document/d/1q_sp2jLQsDEOX7Yug6TPOv7Fwrys6EwcF5Itxjkno7Y/edit#heading=h.bag869lp4lyz

> Following up as we close on code freeze to determine what will ship and what will slip.
This is something we can improve on. Currently we have the [#core-development](https://crossplane.slack.com/archives/CEF5N8X08) Slack channel for coordinating this. 

> Running the release process.
We have a dedicated release repo, https://github.com/crossplane/release, that has the process and issue templates in them and is being used for managing release. See the most recent release here: https://github.com/crossplane/release/issues/16

> Owning backport (patch) releases?
This is done by the release manager today.

> Proactively improving the release process and automation.
This has received some attention since this issue was opened, culminating in what we have in `crossplane/release` today.

[...]

PR #3575, opened 2023-01-13
ci: renovate configuration

### LINKED ISSUE #3562: CVEs reported requiring dependencies bump
### What happened?

Running a [trivy](https://github.com/aquasecurity/trivy) scan against `controlplane/controlplane:master` image reports the following CVEs:
```
usr/local/bin/crossplane:
        CRITICAL, CVE-2022-1996, github.com/emicklei/go-restful, v2.15.0+incompatible => 2.16.0
        HIGH, CVE-2022-21698, github.com/prometheus/client_golang, v1.11.0 => 1.11.1
        HIGH, CVE-2022-27664, golang.org/x/net, v0.0.0-20220516155154-20f960328961 => 0.0.0-20220906165146-f3363e06e74c
        MEDIUM, CVE-2022-41717, golang.org/x/net, v0.0.0-20220516155154-20f960328961 => 0.4.0
        HIGH, CVE-2022-32149, golang.org/x/text, v0.3.7 => 0.3.8
```

Dependencies should be updated in order to fix these CVEs.

### How can we reproduce it?

Running `trivy image controlplane/controlplane:master` to have a table formatted output or `trivy image --format template --template '{{- range . -}}{{ .Target }}:{{ printf "\n" }}{{ range .Vulnerabilities }}{{ printf "\t" }}{{- .Severity }}, {{ .VulnerabilityID }}, {{ .PkgName }}, {{ .InstalledVersion }} => {{ .FixedVersion }}{{ printf "\n"}}{{ end }}{{ end }}' crossplane/crossplane:master` to have the output above.

Crossplane version: 
Latest `crossplane/crossplane:master@sha256:81ef6dc0f34aa9c71bd4c0b0e843816a96baa502bcda36cbe0dc1a61cdf84a8c` image.

### LINKED ISSUE #3571: Pin to specific Go patch releases in CI

We currently [pin CI to Go minor version](https://github.com/crossplane/crossplane/blob/3b22fcba51b1a445146cce539a338d06f6ee9a79/.github/workflows/ci.yml#L13) via the [setup-go](https://github.com/actions/setup-go) action. The motivation for doing so is reducing the manual toil in updating to new patch  versions when released (though it is worth noting that we don't currently enable [check latest](https://github.com/actions/setup-go/tree/v2.2.0#check-latest-version) functionality so we likely are not updating as frequently as we would like). However, this also means that running CI on the same commit at different points in time may not produce identical builds. The only way to determine what version of Go our binaries were built with is by downloading and inspecting the binary.

This is particularly cumbersome when updating to a specific Go version is necessary due to vulnerabilities or a significant bug. The existing published binaries would need to be inspected, as would any subsequent patch to update the Go version (i.e. the patch would be subject to the action fetching the patch version of Go that we want).

Running our CI pipelines on the same commit should produce the same output, regardless of when they are run. We ensure this for dependencies by committing our `go.mod`/`go.sum` and tooling by pinning to specific versions in our `make` machinery. Likewise, we reference base images by digest in our Dockerfile to ensure we are getting the same content. Because our GitHub workflows are also committed to this repo, we could ensure the same reproducability by pinning to full patch release versions of Go, rather than just minor versions. Alternatively, we could move downloading Go into the `build` submodule machinery and pin the version in our `Makefile` similar to how we manage other tooling. This would provide the advantage of making the build reproducible outside of our specific CI as well.

This does create some additional toil on maintainers, but it is something we could automate via tools like Dependabot or Renovate -- something @phisco has already proposed introducing in https://github.com/crossplane/crossplane/pull/3566 (note: there may be some configuration necessary to have it propose an update to the workflow files).

phisco:

Fixes #3562
Fixes #3571 

I have:

Configured it on my fork and it looks like it's working properly: https://github.com/phisco/crossplane

### DIFF
--- a/.github/ISSUE_TEMPLATE/release.md
+++ b/.github/ISSUE_TEMPLATE/release.md
@@ -36,6 +36,7 @@ examples of each step.
 - [ ] Run the [Promote workflow][promote-workflow] with channel `stable` on the release branch and verified that the tagged build version exists on the [releases.crossplane.io] `stable` channel.
 - [ ] Published a [new release] for the tagged version, with the same name as the version and descriptive release notes.
 - [ ] Update the [`crossplane/test` repo test workflows][crossplane-test-workflows] to ensure the checkout release branch and helm install version(s) point at the new release
+- [ ] Update the `baseBranches` list in renovate's configuration file, `.github/renovate.json5`, with the new release branch, removing the old unsupported release one.
 - [ ] Ensured that users have been notified of the release on all communication channels:
   - [ ] Slack
   - [ ] Twitter
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -1,0 +1,138 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": [
+    "config:base"
+  ],
+// The maximum number of PRs to be created in parallel
+  "prConcurrentLimit": 5,
+// The branches renovate should target
+// PLEASE UPDATE THIS WHEN RELEASING.
+  "baseBranches": ["master","release-1.8","release-1.9","release-1.10"],
+  "ignorePaths": ["design/**"],
+  "postUpdateOptions": ["gomodTidy"],
+// By default renovate will auto detect whether semantic commits have been used
+// in the recent history and comply with that, we explicitly disable it
+  "semanticCommits": "disabled",
+// All PRs should have a label
+  "labels": ["automated"],
+  "regexManagers": [
+    {
+// We want a PR to bump Go versions used through env variables in any Github
+// Actions, taking it from the official Github repository.
+      "fileMatch": ["^\\.github\\/workflows\\/[^/]+\\.ya?ml$"],
+      "matchStrings": [
+        "GO_VERSION: '(?<currentValue>.*?)'\\n"
+      ],
+      "datasourceTemplate": "golang-version",
+      "depNameTemplate": "golang"
+    }, {
+// We want a PR to bump golangci-lint versions used through env variables in
+// any Github Actions, taking it from the official Github repository tags.
+      "fileMatch": ["^\\.github\\/workflows\\/[^/]+\\.ya?ml$"],
+      "matchStrings": [
+        "GOLANGCI_VERSION: '(?<currentValue>.*?)'\\n"
+      ],
+      "datasourceTemplate": "github-tags",
+      "depNameTemplate": "golangci/golangci-lint"
+    }
+  ],
+// PackageRules disabled below should be enabled in case of vulnerabilities
+  "vulnerabilityAlerts": {
+    "enabled": true
+  },
+  "packageRules": [
+    {
+// We need to ignore k8s.io/client-go older versions as they switched to
+// semantic version and old tags are still available in the repo.
+      "matchDatasources": [
+        "go"
+      ],
+      "matchDepNames": [
+        "k8s.io/client-go"
+      ],
+      "allowedVersions": "<1.0"
+    }, {
+// We want a single PR for all the patches bumps of kubernetes related
+// dependencies, as most of the times these are all strictly related.
+      "matchDatasources": [
+        "go"
+      ],
+      "groupName": "kubernetes patches",
+      "matchUpdateTypes": [
+        "patch"
+      ],
+      "matchPackagePrefixes": [
+        "k8s.io",
+        "sigs.k8s.io"
+      ]
+    }, {
+// We want dedicated PRs for each minor and major bumps to kubernetes related
+// dependencies.
+      "matchDatasources": [
+        "go"
+      ],
+      "matchUpdateTypes": [
+        "major",
+        "minor"
+      ],
+      "matchPackagePrefixes": [
+        "k8s.io",
+        "sigs.k8s.io"
+      ]
+    }, {
+// We want dedicated PRs for each bump to non-kubernetes Go dependencies, but
+// only if there are known vulnerabilities in the current version.
+      "matchDatasources": [
+        "go"
+      ],
+      "matchPackagePatterns": [
+        "*"
+      ],
+      "enabled": false,
+      "excludePackagePrefixes": [
+        "k8s.io",
+        "sigs.k8s.io"
+      ],
+      "matchUpdateTypes": [
+        "major",
+      ],
+    }, {
+// We want a single PR for all minor and patch bumps to non-kubernetes Go
+// dependencies, but only if there are known vulnerabilities in the current
+// version.
+      "matchDatasources": [
+        "go"
+      ],
+      "matchPackagePatterns": [
+        "*"
+      ],
+      "enabled": false,
+      "excludePackagePrefixes": [
+        "k8s.io",
+        "sigs.k8s.io"
+      ],
+      "matchUpdateTypes": [
+        "minor",
+        "patch"
+      ],
+      "groupName": "all non-major go dependencies"
+    }, {
+// We want a single PR for all minor and patch bumps of Github Actions
+      "matchDepTypes": [
+        "action"
+      ],
+      "matchUpdateTypes": [
+        "minor",
+        "patch"
+      ],
+      "groupName": "all non-major github action",
+      "pinDigests": true
+    },{
+// We want dedicated PRs for each major bump to Github Actions
+      "matchDepTypes": [
+        "action"
+      ],
+      "pinDigests": true
+    }
+  ]
+}


phisco:
Requires renovate to be installed and enabled by a maintainer via https://github.com/apps/renovate

.github/renovate.json
```diff
@@ -0,0 +1,62 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": [
+    "config:base"
+  ],
+  "prConcurrentLimit": 5,
+  "baseBranches": ["master","release-1.8","release-1.9","release-1.10"],
```
hasheddan, line 7: It is not immediately obvious to me how we could automate this list, but if we have it manual for now we should add a step to the minor release issue template to update this after new releases 👍🏻

.github/renovate.json
```diff
@@ -0,0 +1,62 @@
...
+  "ignorePaths": ["design/**"],
+  "postUpdateOptions": ["gomodTidy"],
+  "recreateClosed": true,
+  "regexManagers": [
+    {
+      "fileMatch": ["^.github/workflows/ci.yml$"],
```
hasheddan, line 13: We have a few other workflows that we'll also want to keep consistent on Go versions, such as https://github.com/crossplane/crossplane/blob/3b22fcba51b1a445146cce539a338d06f6ee9a79/.github/workflows/promote.yml#L16

.github/renovate.json
```diff
@@ -0,0 +1,62 @@
...
+      "matchDatasources": [
+        "go"
+      ],
+      "groupName": "kubernetes non-major packages",
+      "groupSlug": "kubernetes-go",
+      "matchUpdateTypes": [
+        "minor",
+        "patch"
+      ],
+      "matchPackagePrefixes": [
```
hasheddan, line 41: How does separating out the k8s deps from the other deps (both of which we want minor and patch bumps on) impact renovate? Feel free to point me to docs if more efficient!

hasheddan (commented):
I'm +1 on getting this set up -- thank you @phisco! Left a few comments below 👍🏻

.github/renovate.json
phisco: I'll grep around the repo to be sure to handle all of them 👍

.github/workflows/ci.yml
```diff
@@ -10,7 +10,9 @@ on:
 
 env:
   # Common versions
+  # renovate: datasource=repology depName=homebrew/go versioning=loose
```
negz, line 13: What does homebrew mean here?

This is probably a question best answered with a comment in the file explaining what these directives are doing and perhaps pointing to the renovate config file.

phisco: Renovate can be configured to check the latest version available for a package from different sources, see [docs](https://docs.renovatebot.com/modules/datasource/). The most commonly used seems to be "repology" which keeps track of all the versions available from almost all the available sources, e.g. homebrew, this means that renovate will check the latest version available from homebrew for go and golangci-lint. I don't have strong opinions regarding the repository chosen, homebrew felt like a safe bet. I just noticed there is also a golang-version datasource we could use, but still we'd need to handle differently golangci-lint, either through repology or directly from the GitHub repo, so I'm not sure it would be worth switching. I agree a comment could be useful, I'll add that.

phisco: Instead of adding a comment I moved the whole logic to the `renovate.json` file, comments in the workflows files felt too brittle. I also switched to `golang-version` and `github-tags` as datasources for `go` and `golangci-lint` versions respectively. Inspired by [this](https://github.com/timescale/promscale/blob/master/.github/renovate.json) and other examples found grepping through Github.

.github/renovate.json
```diff
@@ -0,0 +1,95 @@
...
+  "prConcurrentLimit": 5,
+  "baseBranches": ["master","release-1.8","release-1.9","release-1.10"],
+  "ignorePaths": ["design/**"],
+  "postUpdateOptions": ["gomodTidy"],
+  "recreateClosed": true,
+  "regexManagers": [
+    {
+      "fileMatch": ["^\\.github\\/workflows\\/[^/]+\\.ya?ml$"],
+      "matchStrings": [
+        "GOLANG_VERSION='(?<currentValue>.*?)'\\n"
```
phisco, line 15:
```suggestion
        "GO_VERSION='(?<currentValue>.*?)'\\n"
```

.github/ISSUE_TEMPLATE/release.md
```diff
@@ -36,6 +36,7 @@ examples of each step.
 - [ ] Run the [Promote workflow][promote-workflow] with channel `stable` on the release branch and verified that the tagged build version exists on the [releases.crossplane.io] `stable` channel.
 - [ ] Published a [new release] for the tagged version, with the same name as the version and descriptive release notes.
 - [ ] Update the [`crossplane/test` repo test workflows][crossplane-test-workflows] to ensure the checkout release branch and helm install version(s) point at the new release
+- [ ] Update the `baseBranches` list in renovate's configuration file, `.github/renovate.json`, with the new release branch, removing the old unsupported release one.
```
negz, line 39: This is fine for now, but I am feeling like the number of toil-ey manual steps in our release process is creeping up again. Something to think about as part of https://github.com/crossplane/crossplane/issues/3047, perhaps.

.github/renovate.json
negz: https://docs.renovatebot.com/configuration-options/#configuration-options.

They don't seem to state explicitly but it looks like this config file is actually https://json5.org (which I had never heard of before) - a superset of JSON that supports `//` style comments. I definitely encourage dropping as much context/rationale into config files like this as possible so that future readers can understand intent.

negz:
I just enabled the Renovate app on the crossplane org.

negz:
> I just enabled the Renovate app on the crossplane org.

Apparently this causes Renovate to open a PR to every repo - didn't realize that sorry.

.github/renovate.json
```diff
@@ -0,0 +1,94 @@
+{
```
negz, line 1: Nitpick: I'm not a huge fan of the (e.g) `chore(deps)` PR title style that Renovate uses by default. I _think_ there's settings to override this, e.g. https://docs.renovatebot.com/configuration-options/#prtitle.

My preference would be to communicate:

* What is the impact/urgency of the PR. Perhaps `chore` vs `fix` is trying to communicate this, but it's unclear.
* What branch the PR is against (if not master) - e.g. `[release-1.10] update foo to v1.0.0`
* That the PR is automated. I don't think we need to mention Renovate by name given the PR comes from @renovate.
*

phisco: Totally agree on the branch and highlighting it's an automated pr. About the chore/fix, I'll dig a little deeper in the logic it applies, I think that the grouping also affects it. Did you have something in mind instead of chore/fix? I'm used to conventional commits, but it looks like it's not commonly enforced on the project, so I'm fine setting renovate not to use it.

phisco: found in the doc a little bit more details: https://docs.renovatebot.com/semantic-commits/

negz: > Did you have something in mind instead of chore/fix? I'm used to conventional commits, but it looks like it's not commonly enforced on the project, so I'm fine setting renovate not to use it.

Ah - I didn't know about conventional commits. My guess was that Renovate might be using "fix" to mean "this PR fixes a security issues", and "chore" to mean "this PR just bumps to the latest version for reasons other than security issues". I could be wrong though - I haven't actually read up on Renovate and am basically just making assumptions.

In retrospect I think I'd personally prefer to _only_ have Renovate bump dependencies when there are security reasons to do so. We can bump for new features/functionality as part of our development process. I like the idea that Renovate only opens a PR when there's a security issue to remediate because it means that all Renovate PRs would be a priority.

phisco: There are different options, it looks like it's possible to only have PRs opened in case of Vulnerabilities, see [here](https://github.com/renovatebot/renovate/discussions/15490), it's also possible to have only PRs that have been explicitly approved through the dashboard issue, see [my fork](https://github.com/phisco/crossplane/issues/111). Personally, I'd rather have it enabled for everything, it will be super noisy in the beginning, but once things are updated I expect it to be manageable. Then, we can fine tune the configuration as we go obviously.

bobh66:
Somewhat related - I have been looking at updating the dependencies for k8s 0.26 in crossplane-runtime and there are some significant (to me) changes required in the kubernetes client and testing code in both crossplane-runtime and crossplane to get it to work.   I'm not sure whether I should continue that effort or let the PRs show up from Renovate and deal with it then?

.github/renovate.json
phisco: Added comments, thanks for pointing that out!

phisco:
@bobh66, yes, 0.26 is known to have caused more than a few headaches. You are more than welcome to continue working on that, once Renovate is configured properly we'll just either ignore its PR in favour of yours or just don't get one if yours has already been merged, which I hope will be the case 😉

phisco:
@negz in the latest change I explicitly disabled semantic commits and have added an `automated` label to all PRs, the result will be this:
image

.github/renovate.json
negz: > Personally, I'd rather have it enabled for everything

Why is that?

On one hand I suppose it could help with unknown/undisclosed/zero-day vulnerabilities. It does feel like it's in a way fighting against Go's minimal version selection though, and could lead to unexpected behaviour changes.

negz:
@phisco That looks great. I think if you rebase the PR the unit test issue should be solved - it's a flake in master that just started showing up recently.

.github/renovate.json
phisco: We can always deliberately ignore PRs by closing them, but at least we know new updates are there and we know we already discarded them for some reason.

negz:
@phisco Actually, one other (possibly dumb) question - why Renovate rather than Dependabot? I don't want to change direction at this point but it would be good to capture the rationale somewhere for posterity. My understanding is that they both do the same thing, but it seems like Dependabot maybe has some more native GitHub integration (I recall seeing Dependabot options that seemed native to the GitHub UX).

phisco:
I was leaning toward dependabot too in the beginning, as it's directly supported by GitHub, but renovate is much more configurable and more importantly it can handle multiple target branches, which we need to handle release branches, while dependabot does not support that AFAICT or at least not in a straight forward way.

.github/renovate.json
jbw976: I don't feel super strongly, but I am on the side of security focused changes only.  Without having any direct experience with this, my concern would be having too much churn in our dependencies if we are constantly picking up every change/update to them, which would increase the risk of regressions making their way in.

If we need new functionality, we could purposefully pick up those dependencies as needed.

.github/renovate.json5
```diff
@@ -0,0 +1,115 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": [
+    "config:base"
+  ],
+// The maximum number of PRs to be created in parallel
+  "prConcurrentLimit": 5,
+// The branches renovate should target
+// PLEASE UPDATE THIS WHEN RELEASING, see
```
negz, line 9: Looks like maybe a link is missing here.

negz (approved):
One final comment, but otherwise this looks great to me. Excited to have this functionality.

PR #3581, opened 2023-01-15
Restore packagePullPolicy Always polling behavior

hasheddan:

The packagePullPolicy Always behavior is intended to cause the package manager to poll for new package content every minute. However, it was previously updated to requeue immediately, which starts the process of exponential backoff. This restores the intended and documented behavior.

> Note: this is being backported to all active branches. It appears that this behavior was originally introduced in `v1.6.0`, but `v1.6` and `v1.7` release branches are no longer active.

I have:

Verified via unit tests and running locally, demonstrating that we are once again checking every minute rather than immediately.

### DIFF
--- a/internal/controller/pkg/manager/reconciler.go
+++ b/internal/controller/pkg/manager/reconciler.go
@@ -45,11 +45,16 @@ import (
 
 const (
 	reconcileTimeout = 1 * time.Minute
+
+	// pullWait is the time after which the package manager will check for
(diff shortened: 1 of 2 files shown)


hasheddan (commented):
Documented behavior here: https://docs.crossplane.io/v1.10/concepts/packages/#specpackagepullpolicy

ytsarev (approved):
LGTM, thanks, Dan!

PR #3599, opened 2023-01-19
Fix flaky unit test by using approximate time comparison

hasheddan:

We are experience intermittent unit test failures due to diff in connection secret published time. We should likely be mocking time in these reconcilers, but using approximate comparison allows us to cover the code path of setting the last published time with a minimal diff to these reconcilers.

Example flake: https://github.com/crossplane/crossplane/actions/runs/3948296294/jobs/6758125407

> I'll follow up on this by creating a tracking issue with `good first issue` label on mocking time in these reconcilers.

I have:

`make test`

### DIFF
--- a/internal/controller/apiextensions/claim/reconciler_test.go
+++ b/internal/controller/apiextensions/claim/reconciler_test.go
@@ -20,8 +20,10 @@ import (
 	"context"
 	"io"
 	"testing"
+	"time"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	corev1 "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


internal/controller/apiextensions/claim/reconciler_test.go
```diff
@@ -791,7 +793,7 @@ func TestReconcile(t *testing.T) {
 
 			got, err := r.Reconcile(context.Background(), reconcile.Request{})
 
-			if diff := cmp.Diff(tc.want.claim, tc.args.claim, test.EquateErrors()); diff != "" {
+			if diff := cmp.Diff(tc.want.claim, tc.args.claim, cmpopts.EquateApproxTime(30*time.Second)); diff != "" {
```
negz, line 796: Perhaps a comment explaining why we equate time approximately would be useful for future readers?

Also, I don't feel strongly but 30 secs feels like a long time. I've seen this twice now and each time it was a 1 second difference. I went with 3 seconds.

negz: Nevermind, I dropped these comments slightly too late and they're nothing serious so feel free to ignore.

PR #3597, opened 2023-01-18
Remove docs dir and move packages & meta CRDs to cluster dir

hasheddan:
### Description of your changes

Moves the configuration starter packages to the cluster directory so
that they live alongside other built artifacts.

Signed-off-by: hasheddan <georgedanielmangum@gmail.com>

Updates package meta CRD generation to target the cluster directory.
These types are not actually installed, but we generate them so that
they can inform API documentation.

Signed-off-by: hasheddan <georgedanielmangum@gmail.com>

Crossplane documentation content now lives in the crossplane/docs
repository. This removes the content from this repository.

Signed-off-by: hasheddan <georgedanielmangum@gmail.com>

Fixes #3569 

I have:

- [x] Read and followed Crossplane's [contribution process].
- [x] Run `make reviewable` to ensure this PR is ready for review.
- [ ] Added `backport release-x.y` labels to auto-backport this PR if necessary.

### How has this code been tested

Executed configuration building commands as used in the GitHub workflow.

[contribution process]: https://git.io/fj2m9

PR #3634, opened 2023-01-24
Resurrect contributing docs and README banner

negz:

We lost a few things in https://github.com/crossplane/crossplane/pull/3597 that it turns out we wanted to keep - namely our GitHub repo banner, and most of our contributing documentation.

I did a whole git dance of reverting the commit that deleted the docs directory and moving everything around in separate commits in the hopes of maintaining file history, but unfortunately everything under `/contributing` is showing up as new content. None of it is actually new - I only moved files around. I did edit the headers on all files under `/contributing` to remove front matter from the old docs setup.

I have:

I browsed the repo in my branch to make sure the links still worked.


contributing/guide-adding-external-secret-stores.md
```diff
@@ -0,0 +1,132 @@
+# Adding Secret Store Support
```
hasheddan, line 1: Are we planning to drop these from the docs site? I definitely don’t think we want to maintain them in two separate places, but if I was developing a provider I’d likely think to look at the website rather the the Crossplane repo.

plumbis: I can't find the reference but @negz and I discussed where "contributing docs" for crossplane/crossplane live, and the answer was this repo. This content will be removed from the docs with v1.11 and later.

hasheddan: Thanks for the context! @negz since the revert does result in the files being marked as moved do you think it still makes sense to bring the full directory back then delete it again?

negz: It seems like the approach I took does maintain the file history, despite it all showing green here, so yes.

negz: > If I was developing a provider I’d likely think to look at the website rather the the Crossplane repo.

It occurred to me after we made this decision that one (and a half) of these documents are mostly aligned with provider development. I'd still prefer to keep everything here for the time being - we can change later if we find there's a discoverability issue.

In particular I do think we want a `CONTRIBUTING.md` file in the repo where the code is, and that makes me want to keep everything "how to contribute code to the project" related in one place.

jbw976: @plumbis perhaps the docs site could have a very simple page for Contributing that then simply points at c/c/contributing?

jbw976 (approved):
Thanks for catching this @negz and cleaning it up! 🙇

contributing/guide-adding-external-secret-stores.md
hasheddan: > It seems like the approach I took does maintain the file history, despite it all showing green here, so yes.

Hmm yeah looks like it updated since last time I was looking 🤔 

> In particular I do think we want a CONTRIBUTING.md file in the repo where the code is, and that makes me want to keep everything "how to contribute code to the project" related in one place.

I agree on `CONTRIBUTING.md` as much of what it is describing is how to contribute to _this_ repository. The others feel less appropriate, but it sounds like y'all have already had discussion around this so I am happy to defer 👍🏻

plumbis: Already ahead of ya boss. https://deploy-preview-271--crossplane.netlify.app/contribute/

Screen Shot 2023-01-24 at 1 40 10 PM

PR #3639, opened 2023-01-24
Pin to Go 1.19.5

### LINKED ISSUE #3571: Pin to specific Go patch releases in CI

hasheddan:

Pins CI to Go 1.19.5, which is the latest patch release. Renovate should update when new releases are available.

Fixes #3571 

I have:

I did some digging in the renovate code base and I _think_ this should result in it starting to bump patch versions for us. I am not 100% sure that is the case though, so this may need to be adjusted.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ on:
 env:
   # Common versions
   GO_VERSION: '1.19.5'
-  GOLANGCI_VERSION: 'v1.49'
+  GOLANGCI_VERSION: 'v1.49.0'
   DOCKER_BUILDX_VERSION: 'v0.10.0'
 
   # Common users. We can't run a step 'if secrets.AWS_USR != ""' but we can run
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ on:
 
 env:
   # Common versions
-  GO_VERSION: '1.19'
+  GO_VERSION: '1.19.5'
   GOLANGCI_VERSION: 'v1.49'
   DOCKER_BUILDX_VERSION: 'v0.10.0'
 
--- a/.github/workflows/promote.yml
+++ b/.github/workflows/promote.yml
@@ -13,7 +13,7 @@ on:
 
 env:
   # Common versions
-  GO_VERSION: '1.19'
+  GO_VERSION: '1.19.5'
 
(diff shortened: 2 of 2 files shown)


phisco (approved):

.github/workflows/ci.yml
```diff
@@ -10,7 +10,7 @@ on:
 
 env:
   # Common versions
-  GO_VERSION: '1.19'
+  GO_VERSION: '1.19.5'
   GOLANGCI_VERSION: 'v1.49'
```
phisco, line 14:
```suggestion
  GOLANGCI_VERSION: 'v1.49.0'
```

PR #3663, opened 2023-01-25
ci: configure renovate to ignore oss-fuzz action updates

phisco:

Given that oss-fuzz is not tagging releases we'll default to referencing it by branch, pointing only to `master` in the ci workflow.

Closes #3649
Closes #3661
Closes #3664

I have:

Configured on my local repo.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -386,14 +386,14 @@ jobs:
       # seems to build Crossplane inside of a Docker image.
       - name: Build Fuzzers
         id: build
-        uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@700fd90d10cf656bd08a250c2ed19555a9cfc4ef
+        uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
         with:
           oss-fuzz-project-name: "crossplane"
           dry-run: false
           language: go
 
       - name: Run Fuzzers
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -386,14 +386,14 @@ jobs:
       # seems to build Crossplane inside of a Docker image.
       - name: Build Fuzzers
         id: build
-        uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@700fd90d10cf656bd08a250c2ed19555a9cfc4ef
+        uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
```
hasheddan, line 389: @phisco perhaps it is less important for this specific action (because fuzzing is already somewhat non-deterministic), but in general I like for CI workflows at a given commit to behave the same regardless of when they are executed. If we pin to `master` here I believe that multiple runs of the same workflow at the same commit would potentially use different versions of the action.

phisco: Totally agree, this should be the exception, not the norm. Given that they are not using tags, we'll get a PR for every commit they do, so either we pin it and disable updates, so we end up never updating that or we just use master and have implicit updates without the noise. Given that it's actually interacting with an external service I think it's better to have it updated one way or another in order to avoid falling behind and having it broken all of a sudden because some API changed on the other side.

hasheddan: @phisco that makes sense -- thanks for sharing your reasoning here!

hasheddan (approved):

PR #3445, opened 2022-11-08
Add fuzzer in Xcrd package

AdamKorcz:
Signed-off-by: AdamKorcz <adam@adalogics.com>

### Description of your changes

Moves over two fuzzers from cncf-fuzzing: https://github.com/cncf/cncf-fuzzing/blob/main/projects/crossplane/xcrd_fuzzer.go

I have:

- [x] Read and followed Crossplane's [contribution process].
- [x] Run `make reviewable` to ensure this PR is ready for review (but get the error `make: *** No rule to make target 'reviewable'. Stop.`)
- [ ] Added `backport release-x.y` labels to auto-backport this PR if necessary.

### How has this code been tested
Both fuzzers can be tested with:
```
go test -fuzz=FuzzForCompositeResourceXcrd
go test -fuzz=FuzzForCompositeResourceClaim
```

from the `internal/xcrd` dir

[contribution process]: https://git.io/fj2m9

PR #3671, opened 2023-01-26
tests(fuzz): move remaining test cases from cncf-fuzzing

phisco:
Move over remaining test cases from
https://github.com/cncf/cncf-fuzzing/tree/main/projects/crossplane.

Co-authored-by: AdamKorcz <adam@adalogics.com>

To continue the effort to move all the fuzz test cases from [cncf-fuzzing](https://github.com/cncf/cncf-fuzzing/tree/main/projects/crossplane) started with #3445, this PR moves the remaining test cases.

I have:

Test cases can be run in the folder of each test with:
```
go test -fuzz=Fuzz<name>
```


AdamKorcz:
Awesome! Can we also move the oss-fuzz build script over? https://github.com/cncf/cncf-fuzzing/blob/main/projects/crossplane/build.sh

phisco:
@AdamKorcz could you have a look at it too and check everything is fine? I run the custom build script as you suggested and it's working.

AdamKorcz:
LGTM. Could you also set up a PR on cncf-fuzzing that removes the fuzzers there and change the path to the build script here: https://github.com/google/oss-fuzz/blob/060f02a8daa5968f0bad258463279032cae74d26/projects/crossplane/build.sh#L18

phisco:
@AdamKorcz https://github.com/google/oss-fuzz/pull/9515

AdamKorcz:
> @AdamKorcz [google/oss-fuzz#9515](https://github.com/google/oss-fuzz/pull/9515) and [cncf/cncf-fuzzing#309](https://github.com/cncf/cncf-fuzzing/pull/309).

Awesome. In that case, let's get this PR merged, and then I will merge the two others.

internal/controller/apiextensions/claim/fuzz_test.go
```diff
@@ -0,0 +1,73 @@
...
+					*o.(*corev1.Secret) = *s
+					return nil
+				}),
+			},
+			Applicator: resource.ApplyFn(func(_ context.Context, o client.Object, _ ...resource.ApplyOption) error {
+				return nil
+			}),
+		}
+		api := &APIConnectionPropagator{client: c}
+		_ = api
```
negz, line 70: What is this for? A comment may help.

internal/controller/pkg/revision/fuzz_test.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"os"
+	"testing"
+
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/google/go-containerregistry/pkg/v1/mutate"
+	"github.com/google/go-containerregistry/pkg/v1/tarball"
+)
+
+// Tests mutate.Extract() - an API used by Crossplane.
```
negz, line 28: This file seems like its testing go-containerregistry, not Crossplane. I would prefer Crossplane tests to test Crossplane code - perhaps go-containerregistry should have its own fuzz tests?

internal/xpkg/fuzz_test.go
```diff
@@ -0,0 +1,78 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+
+	"github.com/crossplane/crossplane-runtime/pkg/parser"
+
+	"github.com/spf13/afero"
+
+	fuzz "github.com/AdaLogics/go-fuzz-headers"
+)
+
+func FuzzParse(f *testing.F) {
```
negz, line 34: This seems like it's testing code from crossplane-runtime, not crossplane. Similar to my comment above, I think this test belongs in crossplane-runtime.

negz (changes requested):

internal/controller/apiextensions/claim/fuzz_test.go
phisco: pretty sure it's just an error, right @AdamKorcz?

internal/xpkg/fuzz_test.go
phisco: Totally agree, @AdamKorcz if you don't have strong reasons to keep these, I'd just remove both `xpkg.FuzzParse` and `revision.FuzzGCRExtract`

phisco: On the other hand though, those are APIs we use and pass user provided inputs more or less directly into, and for example we can not leverage oss-fuzz for `crossplane-runtime`, nor we can push `go-containerregistry` to do fuzzing, so it might make sense to keep them

AdamKorcz: IMO if the APIs are important to Crossplane, then it makes sense to keep them for the purpose of being aware if any issues exist in these APIs.

internal/controller/apiextensions/claim/fuzz_test.go
AdamKorcz: > pretty sure it's just an error, right @AdamKorcz?

It is

internal/xpkg/fuzz_test.go
AdamKorcz: AFAIK, the GCR parsing function is being fuzzed on OSS-Fuzz.

phisco: I removed the test cases testing go-container registry and left the controller-runtime one, I'll have a look at whether we can lift FuzzParse to test `crossplane` logic instead of just `controller-runtime`.

phisco: I didn't find an obvious improvement. I don't have strong opinions on this, let me know if you want me to remove also this test case, @negz.

negz: Yes, please remove this. I do want the test, but it belongs in crossplane-runtime, not here.

test/fuzz/oss_fuzz_build.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash -eu
```
negz, line 1: Do we need this script? I don't see anything calling it.

phisco: It's called by oss-fuzz

phisco:
```suggestion
#!/bin/bash -eu
# DO NOT DELETE: this script is used from oss-fuzz.
# You can find more details in the official documentation: https://google.github.io/oss-fuzz/getting-started/new-project-guide/go-lang/
```

phisco: Added a comment directly

negz (approved):

PR #3686, opened 2023-01-30
Feature: Allow service account to be not be created and custom service account name

### LINKED ISSUE #2755: Provide a bring-your-own ServiceAccount option in the crossplane helm chart

markandersontrocme:
Signed-off-by: Mark Anderson-Trocme <mark.anderson@altitude-sports.com>

As discussed in https://github.com/crossplane/crossplane/pull/2756#discussion_r766759008 when `serviceAccount.name` is provided no serviceAccount is created.

Fixes #2755

I have:

 I ran `helm template . -f values.yaml --debug` with a `serviceAccount.name` set and not set and output was as expected.

### DIFF
--- a/cluster/charts/crossplane/templates/serviceaccount.yaml
+++ b/cluster/charts/crossplane/templates/serviceaccount.yaml
@@ -1,4 +1,4 @@
-{{- if not .Values.serviceAccount.name -}}
+{{- if .Values.serviceAccount.create -}}
 apiVersion: v1
 kind: ServiceAccount
 metadata:
--- a/cluster/charts/crossplane/templates/serviceaccount.yaml
+++ b/cluster/charts/crossplane/templates/serviceaccount.yaml
@@ -1,7 +1,8 @@
+{{- if not .Values.serviceAccount.name -}}
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  name: {{ template "crossplane.name" . }}
+  name: {{ template "crossplane.serviceAccountName" . }}
   namespace: {{ .Release.Namespace }}
   labels:
     app: {{ template "crossplane.name" . }}
@@ -15,3 +16,4 @@ imagePullSecrets:
 - name: {{ $secret }}
 {{- end }}
 {{ end }}
+{{- end -}}
--- a/cluster/charts/crossplane/README.md
+++ b/cluster/charts/crossplane/README.md
@@ -78,7 +78,8 @@ and their default values.
 | `nodeSelector` | Enable nodeSelector for Crossplane pod | `{}` |
 | `customLabels` | Custom labels to add into metadata | `{}` |
 | `customAnnotations` | Custom annotations to add to the Crossplane deployment and pod | `{}` |
-| `serviceAccount.name` | Custom name to give to the serviceaccount of Crossplane, defaults to Release name. Does not create the serviceAccount if set | `""` |
+| `serviceAccount.create` | Create the serviceaccount for Crossplane | `true` |
+| `serviceAccount.name` | Custom name to give to the serviceaccount of Crossplane, defaults to Release name | `""` |
 | `serviceAccount.customAnnotations` | Custom annotations to add to the serviceaccount of Crossplane | `{}` |
 | `priorityClassName` | Priority class name for Crossplane and RBAC Manager (if enabled) pods | `""` |
 | `resourcesCrossplane.limits.cpu` | CPU resource limits for Crossplane | `100m` |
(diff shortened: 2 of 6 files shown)


cluster/charts/crossplane/templates/serviceaccount.yaml
```diff
@@ -1,7 +1,8 @@
+{{- if not .Values.serviceAccount.name -}}
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  name: {{ template "crossplane.name" . }}
+  name: {{ template "crossplane.serviceAccountName" . }}
```
hasheddan, line 5: Does using `nameOverride` (see discussion: https://github.com/crossplane/crossplane/discussions/3680) not work for your use case?

markandersontrocme: The issue with this is that you are left with a dangling service account that is not used.

In our use case we create and manage all service accounts in a separate project, so being able to provide a SA name is clean and follows other projects way of managing SA (example [Keda](https://github.com/kedacore/charts/blob/main/keda/values.yaml#L127))

cluster/charts/crossplane/templates/serviceaccount.yaml
```diff
@@ -1,7 +1,8 @@
+{{- if not .Values.serviceAccount.name -}}
```
phisco, line 1: instead of relying on the name itself I'd add another boolean flag `.Values.serviceAccount.create`, it would allow to cover also the "I want the chart to create the serviceAccount, but it should have a specific name" use-case. You can see an example in the [kube-prometheus-stack helm chart](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/values.yaml), but it's pretty common practice across most well-known helm charts.

markandersontrocme: done!

hasheddan: I don't think I agree with this -- as @muvaf [pointed out](https://github.com/crossplane/crossplane/pull/2756#discussion_r766759008) in the first iteration of this PR, the presence of a name if enough for us to know whether to create or not.

cluster/charts/crossplane/templates/serviceaccount.yaml
hasheddan: > In our use case we create and manage all service accounts in a separate project, so being able to provide a SA name is clean and follows other projects way of managing SA (example [Keda](https://github.com/kedacore/charts/blob/main/keda/values.yaml#L127))

@markandersontrocme so is your goal here to allow for _not_ creating an SA rather than just changing its name? If so, this could lead to some surprising behavior. Are you still wanting the [aggregated ClusterRole to be bound](https://github.com/crossplane/crossplane/blob/master/cluster/charts/crossplane/templates/clusterrolebinding.yaml) to your manually managed SA?

markandersontrocme: > is your goal here to allow for not creating an SA
yes

Good point on the ClusterRole, I can update that to point to 

```
subjects:
- kind: ServiceAccount
  name: {{ template "crossplane.serviceAccountName" . }}
```

cluster/charts/crossplane/templates/serviceaccount.yaml
phisco: Sorry @hasheddan, I didn't go through the whole thread. I do agree it could be enough, but personally as a user I would have to read the template to be sure of its behavior, while the create flag is a common and recognizable pattern. Unfortunately helm charts are not easily configurable except for what the values provide out of the box, so not supporting a use-case usually leads users to either fork the chart or have to discard the chart.

markandersontrocme: I agree with @phisco that the create flag is more explicit and makes the UX much more clear. However if you want me to just do it based on name, I can just revert the commit back to what it was initially both give the same output.

markandersontrocme: @hasheddan has a decision been made on the implementation?

PR #3713, opened 2023-02-03
Remove configurations and workflow

negz:

These configurations were used by our documentation. The docs have now been updated such that they're no longer in use. I don't think there's value in continuing to maintain them.

Specifically these are superseded by part 2 and 3 of the AWS Quickstart at https://docs.crossplane.io/v1.11/getting-started/provider-aws/.

I have:

N/A


hasheddan (commented):
Our quick start guides no longer lead with installing these packages, but I do think there could be some value in keeping them around and adding an option to install them in part 3 to see the end result without having to build and push your own package.

hasheddan:
> Generally, I feel that keeping these around is not worth the maintenance burden to keep them updated.

That's fair -- mostly just would like to see if @plumbis is interested in utilizing pre-built packages and, if so, if we want to build them here or in c/docs. If the former, it might make sense to modify these rather than drop all of them and the corresponding machinery.

plumbis:
I'm totally open to packages with context. That is to say, if someone wants to provide a package and a doc that says "run this to install a package. Now let's describe what just happened....". 

For the updated quick start I'm okay with opening with a "want to skip all this and just see it live? run this command...." option, I just didn't add it yet.

negz:
It turns out there's still parts of the docs that try to use these, per https://github.com/crossplane/docs/issues/333.

I'm not a fan of the idea of keeping these around just in case we want to use them in future. I'd rather delete them and bring them back if/when we want to update them. That's what git history is for. 🙂 I don't think I'm personally likely to spend the time updating them to use official providers any time soon.

Issue #3714, opened 2023-02-03
Add a Composition Runner spec


negz:
### What problem are you facing?

We expect that there may be multiple Composition Function runner implementations. When designing Composition Functions we realized that it was tough to find a perfect solution considering all of the tradeoffs (running functions securely, running many functions, making guarantees around starting functions quickly, etc). We should provide guidance for folks who might want to implement their own Function runner.

We have a reference Composition Function runner (`xfn`) implementation, and strongly typed I/O in the form of a [`RunFunctionRequest`](https://github.com/crossplane/crossplane/blob/v1.11.0/apis/apiextensions/fn/proto/v1alpha1/run_function.proto). We're missing (at least):

* gRPC encryption and authentication requirements.
* gRPC status code meanings.
* How to turn a `RunFunctionRequest` into an OCI runtime spec.
* How to pass `FunctionIO` to and from an OCI container.
* Expectations of the OCI containers we run (i.e. they need a baked in `ENTRYPOINT`)

### How could Crossplane help solve your problem?

We might consider adding a specification to formally describe all these requirements.

jbw976:
/fresh we are tracking this in #3751

PR #3719, opened 2023-02-03
Fix broken docs related links and move the xpkg specification to this repo

hasheddan:

The xpkg spec has lived in the Crossplane docs for its entire lifetime. However, not that Crossplane docs live in a separate repo, it makes more sense to home the specification closer to the code that implements it. A general philosophy for whether a doc should live in c/c is:
- Is the primary audience folks who will be working on Crossplane and / or Crossplane-related projects?
- Is the documentation tightly tied to the code in c/c?

In this case, both are true.

I have:

Checked all broken links that were fixed in this changeset.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ Crossplane is under the Apache 2.0 license.
 <!-- Named links -->
 
 [Crossplane]: https://crossplane.io
-[release cycle documentation]: https://crossplane.io/docs/master/reference/release-cycle.html
+[release cycle documentation]: https://docs.crossplane.io/knowledge-base/guides/release-cycle
 [install]: https://crossplane.io/docs/latest
 [Slack]: https://slack.crossplane.io
 [developer mailing list]: https://groups.google.com/forum/#!forum/crossplane-dev
--- a/contributing/README.md
+++ b/contributing/README.md
@@ -547,7 +547,7 @@ make run
 [crossplane-runtime]: https://github.com/crossplane/crossplane-runtime
 [docs]: https://github.com/crossplane/docs
 [Effective Go]: https://golang.org/doc/effective_go
-[Observability Developer Guide]: docs/contributing/observability_developer_guide.md
+[Observability Developer Guide]: guide-observability.md
 [Dave Cheney's blog]: https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis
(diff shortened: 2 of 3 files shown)


hasheddan (commented):
xref https://github.com/crossplane/crossplane/issues/3714 which is another spec that could go in this new directory. I think a formal spec for external secret stores could also make since at some point (cc @ezgidemirel)

nullable-eth (approved):

jbw976 (approved):
Thanks for this clean-up work here @hasheddan!

For discoverability/visibility of this xpkg specification, are there any links that point to its new location? How will contributors find this spec easily now? (before, they could find it within the docs site itself)

hasheddan (commented):
> For discoverability/visibility of this xpkg specification, are there any links that point to its new location? How will contributors find this spec easily now? (before, they could find it within the docs site itself)

@jbw976 good question -- I don't think the new location is necessarily less discoverable than the previous location, but I would like to make it a little more visible. Perhaps a call out in the `contributing/README.md`? Or do you think it would warrant linking to the `contributing/specifications` directory in the main `README.md`?

jbw976:
> @jbw976 good question -- I don't think the new location is necessarily less discoverable than the previous location...

You can make the call here @hasheddan, but IMO having it within the docs before, with it's nice left hand navigation pane to find things, had higher discoverability than having it now underneath the contribution folder.  I'm fine with its new location, but I still think some linking to it would help people find it.

Your suggestion of linking it from contributing README sounds good (I wouldn't do the main README), and perhaps there is also a place within the docs where we can link to this spec?

e.g. maybe on https://docs.crossplane.io/v1.11/concepts/packages/, we can have a section that links to this formal spec?

hasheddan (commented):
> e.g. maybe on https://docs.crossplane.io/v1.11/concepts/packages/, we can have a section that links to this formal spec?

@jbw976 I think this is a great place to link to it -- ultimately the motivation for moving it to the c/c repo is to ensure that c/c maintainers are responsible for its maintenance and that code changes can be coupled with spec changes. I am concerned about if it only lives in the docs that it might get dropped or modified in a way that causes a divergence in specification and implementation.

jbw976 (approved):
LGTM @hasheddan, looking forward to the follow up PR in the docs to link to this wonderful spec of yours 😊

PR #3736, opened 2023-02-05
chore(renovate): set monthly digest updates for kubernetes deps and every two weeks for docker images

phisco:

With this patch Renovate will only open PRs to bump digest updates once a month and docker ones every two weeks, if required due to security reasons.
See [here](https://docs.renovatebot.com/key-concepts/scheduling/#schedule-when-to-update-specific-dependencies) for more information about Renovate's scheduling.

I have:

Configured on my fork, the existing PR #3683 won't drop the digest update as it is the one due for this month, but it should not open another one until next month.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -64,6 +64,21 @@
       ],
       "allowedVersions": "<1.0"
     }, {
+// We don't want digest updates for kubernetes dependencies, as they are too
+// noisy
+      "matchDatasources": [
+        "go"
+      ],
+      "groupName": "kubernetes patches",
+      "matchUpdateTypes": [
+        "digest",
+      ],
+      "extends": ['schedule:monthly'],
+      "matchPackagePrefixes": [
+        "k8s.io",
+        "sigs.k8s.io"
+      ]
+    }, {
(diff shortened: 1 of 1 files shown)


.github/renovate.json5
```diff
@@ -64,6 +68,21 @@
...
       "allowedVersions": "<1.0"
     }, {
+// We don't want digest updates for kubernetes dependencies, as they are too
+// noisy
+      "matchDatasources": [
+        "go"
+      ],
+      "groupName": "kubernetes patches",
+      "matchUpdateTypes": [
+        "digest",
```
hasheddan, line 78: Does this mean that we will get non-release versions of k8s deps?

phisco: No, it's just for packages that do not use tags, such as k8s.io/utils. Renovate would open a PR every time a new commit gets on main, this way it's just going to do so once a month.

phisco: Yes, it should, but given that we are using distroless images I expect only scheduled PRs 🤞

hasheddan (approved):
Thanks @phisco !

Issue #1476, opened 2020-04-30
Validate compositions via webhook


muvaf:
### What problem are you facing?

You can put any valid YAML under `Composition.spec.to` array. However, they have to be registered on the api-server, which means they can be validated against their registered OpenAPIv3 schemas.

### How could Crossplane help solve your problem?

Implement an admission webhook that will validate every template in `Composition` resource with correct OpenAPIv3 schemas.

wonderflow:
I think all validation work (including https://github.com/crossplane/oam-kubernetes-runtime/issues/14) can all be done by this admission webhook.

negz:
I've deduped https://github.com/crossplane/crossplane/issues/1497 into this issue. So we should validate both:

* The base templates in a composition.
* The fieldpaths in the patches array.

negz:
We actually already have some Composition validation code, for example:

https://github.com/crossplane/crossplane/blob/9850b98edc3362c41f66997a12d7b3adb916eace/internal/controller/apiextensions/composite/composed.go#L107

Currently this code (i.e. `RejectMixedTemplates` and `RejectDuplicateNames`) lives in the XR reconciler, and will fail to reconcile any XR that uses an invalid Composition. This was the best option available at the time, but is not ideal because:

* The person who created the XR (or claim) may not be the person who wrote the Composition. We should tell the person who wrote the Composition that it was invalid - not the person who later tried to use it.
* With CompositionRevisions we're going to make an invalid CompositionRevision of the invalid Composition, then later tell folks that revision is (and will always be) invalid when they actually try to use it.

It would be much better to have an admission control webhook run these validation checks and reject the creation of an invalid Composition in the first place. I think this could be a good "MVP" Composition validation webhook, because running these existing checks in a webhook should be easier to do than validating the schema of composed resource templates.

phisco:
Hello all, I've been putting some thought into this issue and I do agree it makes sense to move the current validation logic to a validating webhook and that it would be really useful to also validate bases and fieldpaths.

However, I didn't find a way to do so without querying the API server. The two options that came to my mind were either to server-side dry-run apply resources leveraging the existing logic in the `APIDryRunRenderer` or to lazily get and then cache the schemas for all the resources encountered. Still, any of these two options could lead to increased latencies and/or requests amplification. Client-side dry-run is not actually checking anything at all, so it's not an option.

[...]

PR #3756, opened 2023-02-10
docs(proposal): compositions validating webhook

phisco:

Proposal to address Composition Validation as started discussing in #1476.

I have:

N.A.

### DIFF
--- a/design/design-doc-composition-validating-webhook.md
+++ b/design/design-doc-composition-validating-webhook.md
@@ -1,0 +1,391 @@
+# Composition Validating Webhook
+
+* Owner: Philippe Scorsolini (@phisco)
+* Reviewers: Crossplane Maintainers
+* Status: Draft
+
+## Background
+
+Validating webhooks in Kubernetes allow users to immediately know whether the
+resources they are applying are correct or contain any issue that could result
+in an error later on.
+
+Validating webhooks have already been introduced in Crossplane for
+`CompositeResourceDefinition`s ([here][original-webhook-pr]) after having
+discussed them more broadly through a dedicated [design
+doc][original-webhook-design-doc]. In the original design document
+Composition’s validation was already mentioned and the conversation has
+continued in a dedicated issue,
+[#1476][original-composition-validation-webhook-issue].
+
+Compositions describe how Crossplane should compose resources into higher level
+Composite resources. From a user’s perspective, Compositions might be complex
+to write as they can embed many other resources at `.spec.resources` and it is
+relatively easy to introduce trivial errors, e.g. a misspelled field, either in
+the `base` or the `patches` part of each of them. For this reason, having a
+tighter feedback loop is of the utmost importance to guarantee a good Developer
+Experience.
+
+Existing implementations of Compositions’ validation are:
+
+- In Crossplane itself, Compositions are only validated once actually selected
+  for a Composite Resource. Validation is performed through a chain of
+  validation functions, performing only a few important checks. The actual
+  validation is done by rendering all the Composed resources and dry-run
+  applying them, leveraging the `APIDryRunRenderer` component. This allows it
+  to check the resources’ validity before actually applying them to reconcile
+  any change required.
+- A [Visual Studio Code plugin][vscode-plugin] was developed by Upbound
+  implementing the Language Server Protocol (LSP) as a sub-command of the
+  [upbound/up][upbound/up] CLI. The plugin is validating Compositions as part
+  of a Package, after having pulled down all the Package’s dependencies
+  declared in the `crossplane.yaml` manifest. It uses a forked version of
+  Crossplane’s Composite resource’s reconciliation loop, stripped of any side
+  effect, which allows it to generate all the Composed resources to be
+  validated against the retrieved schemas. It’s able to validate Composition’s
+  errors end to end, taking into consideration both `patches` and `bases`.
+- In [upbound/upjet][upbound/upjet] in order to properly migrate from
+  `crossplane-contrib/provider-` to `upbound/provider-` providers some
+  validation logic around patches’ fieldpaths was developed. Similarly to the
+  language server approach above, CRDs are available locally and fieldPaths are
+  validated against the proper schema.
+- [crossplane-contrib/crossplane-lint][crossplane-contrib/crossplane-lint] was
+  recently released. It pulls down all the dependencies specified in a
+  `.crossplane-lint.yaml` file so that it can validate all the resources in a
+  Package using their schema. The main difference with the LSP above is firstly
+  that it is implemented as a CLI tool that can be run locally or in CI
+  pipelines, but it also differs in the approach, as it implements only a few
+  specific validation rules, e.g. fieldPaths are correct according to the
+  schemas, but does not validate the end to end result, which means it’s not
+  reporting missing required fields in the generated Managed Resources for
+  example.
+
+## Goals
+
+The goal of this one-pager is not to discuss whether a Validating Webhook for
+Compositions would be useful, as it’s pretty clear and agreed that implementing
+it would improve the developer experience for users working with Compositions.
+Instead it should focus more on the definition and possible implementation of
+it, given that it is not at all trivial.
+
+## Proposal
+
+As we saw, there are different implementations of composition validations out
+in the wild, each focusing on different aspects of Compositions, so we need
+first to define what we mean by validating Compositions.
+
+### Validating Compositions
+
+A Composition is made of a list of `resources`, each specifying a `base` and a
+series of `patches`  to be applied on top of it. Different kinds of `patches`
+are available, which might be using information coming from different sources,
+e.g. values from the associated Composite Resource or EnvironmentConfigs.
+
+We could split each `resource`’s validation in the following parts:
+
+1. the `base` , which should be valid according to its schema.
+2. `patches` , of which each should be validated according to its type and
+   specified policies, but overall we can say that each of them should:
+    1. have a valid path as a `fromFieldPath` according to the schema of the
+       source object, if defined for the type of patch and required by the
+       `PatchPolicy` defined.
+    2. have a valid path as a `toFieldPath` according to the schema of the
+       destination object, if defined for the type of patch.
+    3. have correctly configured types, e.g. if `fromFieldPath` is a string,
+       `toFieldPath` is an integer, the patch should be rejected, but not if
+       there is a mapping correctly setting it.
+3. the result of all the `patches` being applied to the `base` should be valid
+   according to the Managed Resource schema.
+
+Compositions are also defining the following fields that we should take into
+consideration while validating:
+
+- `patchSets`: a list of named patches that can be referenced by resources.
+  Depending on the type of patch it could be impossible to validate them
+  independently from the resource they would be used on. We could think of only
+  validating them once they are dereferenced and associated to a specific
+  resource.
+- `functions`: a list of Composition Functions that can be used to generate or
+  modify the Composed Resources. Given that it’s an `alpha` feature, we can
+  think of not covering this in the first implementation, however we should
+  keep in mind that having at least a function defined would invalidate the
+  assumption that rendered Managed Resources should be valid according to the
+  schema. Moreover, taking them into consideration during the validation phase
+  would mean adding another external dependency, which could both reduce
+  reliability and add latency.
+- `environment`: contains a list of `EnvironmentConfigs` that can be referenced
+  by specific types of patches and a list of`EnvironmentPatches` . Also an
+  `alpha` feature, similar implications to what already said for `functions`,
+  would require additional information and can invalidate the assumption that
+  we render valid Composed Resources if defined.
+- `writeConnectionSecretsToNamespace` : the name of the namespace the `Secret`
+  containing the connection details should be created into. We could check
+  whether it exists already, but that would be against the behaviour Kubernetes
+  users are used to, e.g. a `Deployment` mounting a `Secret` which does not
+  exist yet is considered valid but remains pending until it’s not created.
+- `publishConnectionDetailsWithStoreConfigRef` : a reference by name to a
+  `StoreConfig`, we could check that it exists, but similarly to the namespace
+  it would go against the behaviour Kubernetes users might be expecting.
+
+### Additional Constraints for the ValidatingWebhook use case
+
+A few constraints that makes validating Compositions from a Validating Webhook
+harder w.r.t. the cases mentioned in the Background:
+
+1. The CRDs of the Managed Resources defined as `base` and traversed by
+   `patches` might not yet have been created when the Composition is being
+   validated.
+2. The CRD of the Composite Resources specified by `CompositeTypeRef` and
+   traversed by `patches` might not yet have been created when the Composition
+   is being validated.
+3. We do not have all the required CRDs locally, so we should get them from
+   the API server, if available.
+4. We do not have actual input values supplied from the Composite Resource that
+   are required to render valid and applicable Managed Resources.
+5. Some `alpha` features could require some additional interactions with
+   external systems, e.g. `CompositionFunctions`.
+
+### Solution Proposed
+
+In order to be able to have Compositions properly validated we should cover all
+the `resources` parts, while also taking into consideration all the constraints
+above.
+
+In order to address the possible absence of Managed Resources' CRDs, Composite
+Resources' CRDs or any other required external resource at validation time we
+should allow explicitly setting one of 2 modes, `strict` and `loose`(default) .
+These should be configurable through an annotation, e.g.
+`crossplane.io/composition-validation-mode`, on a Composition or maybe even
+globally via a flag, and would respectively imply that any missing resource
+should result in a direct rejection or a possibly incomplete validation.
+
+In `loose` mode we should emit warnings in case any of the preconditions for a
+check are not met, and skip further validation related to that check.
+
+Regarding the approach, I propose to implement a Composition Validating Webhook
+by validating each `resource` and final rendered `Composed
+Resource`, querying the API Server to get the needed schemas.
+
+I suggest we avoid validating `bases`. Doing that would require to ignore
+validating required fields, as those could be set at a later stage by patches
+or transformations. Assuming we implement full `patches` type coverage,
+validating `bases` per se can be avoided, as any error should either be caught
+validating the patches or can be attributed to the `base` itself.
+
+To validate `patches`, we will need to implement ad-hoc functions, because the
+current patching code is dealing directly with unstructured data and does not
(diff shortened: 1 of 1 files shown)


Piotr1215 (commented):
Apologies if this doesn't fit with the webhook theme, but I wanted to ask about a CLI based validation. 
Something that
- could run in a pipeline
- does not require Crossplane installed
- works with directories and YAML files directly in a file system
- validates against XRD schema but also provider MRs, composition schema and patches.

phisco:
@Piotr1215 as described in the "Future Work" section, the end goal should be to have as much as possible reusable logic, allowing us to leverage it for a linter, language server and any other possible usage.

design/design-doc-composition-validating-webhook.md
```diff
@@ -0,0 +1,359 @@
...
+schema related error. The code implemented for
+[upbound/upjet][upject-validation-code] or
+[crossplane-contrib/crossplane-lint][crossplane-lint-validation-code] could be
+good starting points for this.
+
+To render the Composed Resources, we will need to refactor the Composite
+resources reconciliation loop, extracting the rendering logic to remove any
+side effect it currently has. Which is similar to what has been done in the LSP
+[implementation][lsp-validation-code]. This step requires all `patches` to have
+been applied in order to have the final Managed Resource to validate.
```
turkenh, line 176: > This step requires all `patches` to have been applied in order to have the final Managed Resource to validate

Isn't this already be part of the rendering logic?

design/design-doc-composition-validating-webhook.md
```diff
@@ -0,0 +1,359 @@
...
+To validate `patches`,  we will need to implement ad-hoc validation functions,
+because the current patching logic is not taking into consideration schemas, as
+it’s dealing directly with unstructured data, therefore not reporting any
+schema related error. The code implemented for
+[upbound/upjet][upject-validation-code] or
+[crossplane-contrib/crossplane-lint][crossplane-lint-validation-code] could be
+good starting points for this.
+
+To render the Composed Resources, we will need to refactor the Composite
+resources reconciliation loop, extracting the rendering logic to remove any
```
turkenh, line 173: > extracting the rendering logic

This could be tricky with the `functions` support since rendering logic would also include running the functions.

design/design-doc-composition-validating-webhook.md
```diff
@@ -0,0 +1,359 @@
...
+side effect it currently has. Which is similar to what has been done in the LSP
+[implementation][lsp-validation-code]. This step requires all `patches` to have
+been applied in order to have the final Managed Resource to validate.
+
+In order to address the possible absence of Managed Resources,
```
turkenh, line 178: nit: 

```suggestion
In order to address the possible absence of Managed Resource Definitions (CRDs),
```

design/design-doc-composition-validating-webhook.md
```diff
@@ -0,0 +1,359 @@
...
+CompositeResourceDefinitions or any other required external resource at
+validation time we should allow explicitly setting one of 2 modes, `strict` and
+`loose`(default) . These should be configurable through an annotation, e.g.
+`crossplane.io/composition-validation-mode`, on a Composition or maybe even
+globally via a flag, and would respectively imply that missing resources should
+result in rejection or be ignored.
+
+I suggest, at least at the moment, to ignore and fail open if `alpha` features
```
turkenh, line 186: How would the validator report the failure when ignored?

design/design-doc-composition-validating-webhook.md
```diff
@@ -0,0 +1,359 @@
...
+good starting points for this.
+
+To render the Composed Resources, we will need to refactor the Composite
+resources reconciliation loop, extracting the rendering logic to remove any
+side effect it currently has. Which is similar to what has been done in the LSP
+[implementation][lsp-validation-code]. This step requires all `patches` to have
+been applied in order to have the final Managed Resource to validate.
+
+In order to address the possible absence of Managed Resources,
+CompositeResourceDefinitions or any other required external resource at
```
turkenh, line 179: > any other required external resource

Definition of Composed resources (whether it is an [MR](https://github.com/upbound/platform-ref-aws/blob/023be259acddd37b32a2e2ee52e0ea9819748c67/package/cluster/network/composition.yaml#L20) or [another XR](https://github.com/upbound/platform-ref-aws/blob/023be259acddd37b32a2e2ee52e0ea9819748c67/package/cluster/composition.yaml#L12)) would always be required if we want to validate their schemas, so MR CRDs or XRD will be needed for schema validation.

However, validation would also always require the `CompositeResource` since we want to render the composed resources and patches refer fields from it. I don't expect having `CompositeResource` instance while installing a `Composition` is possible; most of the time, we install XRD + Composition while setting up the platform but create `CompositeResource`s at a later stage while consuming the platform. So, if we want to validate the rendered resources, I feel like we somehow mock these values. Maybe I am missing something since the VS Code plugin somehow does this. 

Nevertheless, it would be helpful to elaborate a bit on what resources are required at what stage and how we can do the validation in the absence of them.

design/design-doc-composition-validating-webhook.md
phisco: What do you mean? I meant that in case of alpha features being used by a Composition, if such feature is known to change the final rendered resources, e.g. Functions, the validator should not validate anything and "fail open" in the sense that it should short-circuit and just accept the resource. At least in the first iteration.

design/design-doc-composition-validating-webhook.md
```diff
@@ -0,0 +1,359 @@
...
+  `StoreConfig`, we could check that it exists, but similarly to the namespace
+  it would go against the behaviour Kubernetes users might be expecting.
+
+### Additional Constraints for the ValidatingWebhook use case
+
+A few constraints that makes validating Compositions from a Validating Webhook
+harder w.r.t. the cases mentioned in the Background:
+
+1. Managed Resources defined as `base` and traversed by `patches` might not yet
+   have been deployed when the Composition is being validated.
```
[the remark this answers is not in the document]
turkenh: Having an MR always implies having its CRD deployed, but not having MRs does not give any hint about its CRDs existence. So, it would be more clear if we mentioned what we need.

design/design-doc-composition-validating-webhook.md
turkenh: The question is independent of alpha features indeed, when we use the `loose`(default) mode, are we planning to let user know that something is wrong (i.e. validation failed). With `strict`, it is clear that we reject the resource with proper error message and user knows what is wrong there.

phisco: We could emit warnings, such as "... CRD was not available" or "... used, so no validation performed".

design/design-doc-composition-validating-webhook.md
```diff
@@ -0,0 +1,359 @@
...
+So, I propose to implement a Composition Validating Webhook by validating each
+`resource`'s `patches` and final rendered `Composed Resource`, querying the API
+Server to get the needed schemas.
+
+I suggest we avoid validating `bases`, that would require to implement
+validation ignoring required fields, as those could be set at a later stage by
+patches. Assuming we implement full `patches` type coverage, validating `bases`
+per se can be avoided, as any error should either be caught validating the
+patches or can be attributed to the `base` itself. But this could be considered
+at a later stage.
```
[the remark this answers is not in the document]
phisco: Yes, I was just suggesting that now as I didn't want to be too specific with the implementation, but now that I have experimented a little bit more in the POC, I'd say we can state it explicitly, as validating bases per se would be a useless effort.

design/design-doc-composition-validating-webhook.md
phisco: > Nevertheless, it would be helpful to elaborate a bit on what resources are required at what stage and how we can do the validation in the absence of them.

I updated the doc to reflect a few learnings from the POC, I think at least initially we should just focus on the happy case, which is `strict` mode, as handling partially missing inputs would make the code much more complex.

> However, validation would also always require the `CompositeResource` since we want to render the composed resources and patches refer fields from it. I don't expect having `CompositeResource` instance while installing a `Composition` is possible; most of the time, we install XRD + Composition while setting up the platform but create `CompositeResource`s at a later stage while consuming the platform. So, if we want to validate the rendered resources, I feel like we somehow mock these values. Maybe I am missing something since the VS Code plugin somehow does this.

Yes, working on the POC I came to the same conclusion, mocking missing Composite resources with the minimal set of required fields is a requirement. I updated the doc to reflect that.

design/design-doc-composition-validating-webhook.md
phisco: Indeed, updated the doc to explicitly mention functions.

design/design-doc-composition-validating-webhook.md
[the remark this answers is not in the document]
phisco: Reworded

lsviben (commented):
Added some minor wording suggestions

design/design-doc-composition-validating-webhook.md
```diff
@@ -0,0 +1,391 @@
...
+  it would go against the behaviour Kubernetes users might be expecting.
+
+### Additional Constraints for the ValidatingWebhook use case
+
+A few constraints that makes validating Compositions from a Validating Webhook
+harder w.r.t. the cases mentioned in the Background:
+
+1. The CRDs of the Managed Resources defined as `base` and traversed by
+   `patches` might not yet have been created when the Composition is being
+   validated.
```
negz, line 137: This makes me wonder whether we should have both a validation webhook and a long-running validation controller.

It seems like a Composition that is valid at creation time (i.e. webhook time) could later become invalid. For example if the CRD for one of the MRs it composes is removed, or (in the case of alpha MRs) there's a breaking schema change.

Such a controller could toggle a `valid` status condition on the `Composition`, preventing it from being selected by XRs.

design/design-doc-composition-validating-webhook.md
```diff
@@ -0,0 +1,391 @@
...
+the `resources` parts, while also taking into consideration all the constraints
+above.
+
+In order to address the possible absence of Managed Resources' CRDs, Composite
+Resources' CRDs or any other required external resource at validation time we
+should allow explicitly setting one of 2 modes, `strict` and `loose`(default) .
+These should be configurable through an annotation, e.g.
+`crossplane.io/composition-validation-mode`, on a Composition or maybe even
+globally via a flag, and would respectively imply that any missing resource
+should result in a direct rejection or a possibly incomplete validation.
```
negz, line 160: Is there a reason to prefer an annotation to a 'real' field on the Composition (e.g. `spec.validation.policy` or similar)?

design/design-doc-composition-validating-webhook.md
```diff
@@ -0,0 +1,391 @@
...
+
+In `loose` mode we should emit warnings in case any of the preconditions for a
+check are not met, and skip further validation related to that check.
+
+Regarding the approach, I propose to implement a Composition Validating Webhook
+by validating each `resource` and final rendered `Composed
+Resource`, querying the API Server to get the needed schemas.
```
negz, line 167: Worth noting that in future I could imagine a composition function validating the final XR before apply, as long as it was the last function in the chain.

design/design-doc-composition-validating-webhook.md
```diff
@@ -0,0 +1,391 @@
...
+
+I suggest we avoid validating `bases`. Doing that would require to ignore
+validating required fields, as those could be set at a later stage by patches
+or transformations. Assuming we implement full `patches` type coverage,
+validating `bases` per se can be avoided, as any error should either be caught
+validating the patches or can be attributed to the `base` itself.
```
negz, line 173: I agree that full validation of bases can be out of scope (at least at first). Some validation does seem like it would be useful, though. For example:

* Making sure folks aren't setting a field that doesn't exist in the schema (sign of a typo or incorrect indentation).
* Making sure folks aren't setting an unsupported field value (e.g. wrong type, unsupported enum).

Anecdotally I think typos and incorrect indentations in bases are one of the more common things folks run into when writing Compositions.

design/design-doc-composition-validating-webhook.md
```diff
@@ -0,0 +1,391 @@
...
+
+The Validating Webhook should:
+
+- reject it if in `strict` mode and no `RDSInstance` Managed resource and/or
+  `XPostgreSQLInstance` Composite Type is deployed yet
+- accept it in any other case
+
+While, any of the errors highlighted in the following invalid example should
+result in the Composition being rejected, except if in `loose` mode and both
+`XPostgreSQLInstance` and `RDSInstance` have not yet been deployed.:
```
negz, line 271: Doesn't this contradict what is suggested previously?

> I suggest we avoid validating `bases`.

design/design-doc-composition-validating-webhook.md
```diff
@@ -0,0 +1,391 @@
...
+current patching code is dealing directly with unstructured data and does not
+take into consideration schemas at all, therefore not reporting any schema
+related error. The code implemented for [upbound/upjet][upject-validation-code]
+or [crossplane-contrib/crossplane-lint][crossplane-lint-validation-code] could
+be good starting points for this.
+
+To render the Composed Resources, we will need to reuse the Composite resources
+reconciliation loop as much as possible, extracting the rendering logic to
+remove any write operation it currently performs. Which is similar to what has
+been done in the LSP [implementation][lsp-validation-code].
```
negz, line 185: It's not clear to me whether this validation would happen at XR-creation time, or at Composition-creation time. If the latter, does this mean we'll use dummy inputs?

negz (commented):
This looks great. I'm close to approving, but I have a few questions first.

design/design-doc-composition-validating-webhook.md
phisco: I'd go with an annotation for the first iterations as it allows us not have to change the APIs and let us change our mind more easily, we could think of moving it to the spec at a later stage once the UX settles to make it more "official" and easily discoverable.

design/design-doc-composition-validating-webhook.md
phisco: Makes sense, that's why Functions should be considered non deterministic and more complex validation logic won't make sense in their presence as suggested below: https://github.com/crossplane/crossplane/pull/3756/files#diff-8dfb220f48da6208755c36731275d4d8ff51209399fedbffd3c2ea3e073fa55fR204

design/design-doc-composition-validating-webhook.md
phisco: Composition creation time. The assumption is that: "A Composition can be considered valid if all patches and transformations are valid per se, and given a minimal mocked XR according to the XRD schema, produces all valid rendered resources according to each resource's schema". See reference implementation for more details: https://github.com/crossplane/crossplane/pull/3902/files#diff-1681f727532a5c3b92fda0de12dd176ef3d9320676691daf5bb5cb0e4d0221b5R78.

So yes, we would need to mock inputs, mostly the XR, but we might think of mocking also other inputs, e.g. EnvironmentConfigs or decide to let the reconciler use actual ones if present, for read only operations. We did not handle EnvironmentConfigs in the reference implementation at the moment, only mocking the minimal required XR.

design/design-doc-composition-validating-webhook.md
phisco: Fair point, indeed given how complex and powerful Compositions are, there are a lot of possible bad scenarios. 

Right now we have no validation at all and errors are reported only once and if a Composition is selected by an XR. Having a validating Webhook would allow us and users to be at least sure that Compositions are valid at deploy time, if all requirements were met. Which paired with a GitOps approach also will let users also know whether Compositions became invalid for any reason.

Obviously later on we might want to address the run time too and we'll have to decide how to address questions like the following ones:
- what happens if the requirements were not met at deploy time, but are met later on and make the Composition invalid? Successive applies of the Composition will fail, but should we be able to notice it and report it in some way?
- what happens if all requirements were met at deploy time and therefore the resource was fully validated and later on one of the requirements is not satisfied anymore? Successive applies will still succeed, but we'll have less confidence and depending on whether there are already XRs using it or not how should these be handled?
- what if it was fully validated at deploy time and, as you are suggesting, one of the requirements changes and makes the Composition invalid? Successive applies of the Composition will fail, but what should happen for existing and new XRs using it? How are users going to be notified?

Ideally we would want to only accept fully validated Compositions, always know whether a Composition is valid or not, prevent any action that would invalidate a valid Composition and report any error on the Composition itself as soon as possible.

However, given that currently users expect not to have any validation at all, we can not be too aggressive on it, so I'd suggest we start from the Webhook and iterate from there one step at a time.

design/design-doc-composition-validating-webhook.md
negz: That sounds good to me.

I do recall though that upstream Kubernetes moved away from the practice you described for some reason. Might be worth checking why.

design/design-doc-composition-validating-webhook.md
negz: Got it. It wasn't obvious to me reading the design doc that we had decided to mock inputs. I'll re-read today and see whether I just missed it.

design/design-doc-composition-validating-webhook.md
negz: Got it. Per my below comment I didn't realize we had chosen to mock out the XR input. This makes more sense in that light.

design/design-doc-composition-validating-webhook.md
negz: > However, given that currently users expect not to have any validation at all, we can not be too aggressive on it, so I'd suggest we start from the Webhook and iterate from there one step at a time.

Absolutely - I should have been explicit I didn't mean to bring any of this into scope for this current design.

design/design-doc-composition-validating-webhook.md
```diff
@@ -0,0 +1,391 @@
...
+
+All `patches` will have to be applied in order to have the final Managed
+Resource properly validated, so we'll need to provide all the required inputs
+to the reconciler, to properly apply them. Inputs could be either actual
+resources available or mocked ones.
```
negz, line 190: It sounds like we've decided to use mocked inputs. I think it would be useful to be explicit about that in this document, and explicit about what inputs we'll mock.

negz (approved):
This looks good to me, thanks! One final suggestion to clarify that we'll be mocking (it wasn't obvious to me) but I'm on board with the design overall.

turkenh (approved):
This looks good to me 🙌 

One last suggestion, it could be helpful to group/mention different types of validations that we are planning to implement. You have already come up with the 3 groups, something like **logical validations**, **schema validations**, and **rendered validations** in the form of implementation phases. While I agree that this may sound like an implementation detail, it would still be helpful for conceptualizing the different types of validations we are planning to cover.

PR #3758, opened 2023-02-10
fix(chart): Satisfy the Kubernetes PSS (Restricted) by default

mikebryant:

These changes aim to satisfy https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted by default.

I have:

```
✗ make reviewable
make: *** No rule to make target `reviewable'.  Stop.
```

We're currently running with these settings in our installation of Crossplane.

### DIFF
--- a/cluster/charts/crossplane/values.yaml.tmpl
+++ b/cluster/charts/crossplane/values.yaml.tmpl
@@ -59,10 +59,16 @@ resourcesCrossplane:
     memory: 256Mi
 
 securityContextCrossplane:
+  capabilities:
+    drop:
+    - ALL
+  runAsNonRoot: true
   runAsUser: 65532
   runAsGroup: 65532
   allowPrivilegeEscalation: false
   readOnlyRootFilesystem: true
+  seccompProfile:
+    type: RuntimeDefault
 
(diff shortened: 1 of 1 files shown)


negz:
> `make: *** No rule to make target `reviewable'.  Stop.`

Try running just `make` first. I think that will pull down the git submodule required to run `make reviewable`.

cluster/charts/crossplane/values.yaml.tmpl
```diff
@@ -59,10 +59,16 @@ resourcesCrossplane:
     memory: 256Mi
 
 securityContextCrossplane:
+  capabilities:
+    drop:
+    - ALL
+  runAsNonRoot: true
```
negz, line 65: How does this interplay with `runAsUser` below?

negz:
@mikebryant Are you still interested in following up on this PR?

cluster/charts/crossplane/values.yaml.tmpl
phisco: And it makes sense to have them configured by default in the chart too, I don't see a reason not to.

PR #3762, opened 2023-02-12
Apply pollInterval to Claim and Composite reconcilers

### LINKED ISSUE #3761: crossplane ignores --poll-interval parameter
### What happened?
We deploy crossplane with the --poll-interval parameter set to 5m to reduce CPU usage and API traffic.  The PollInterval is not passed to the options for the controllers that do reconciliation so they use the default value of 1m.

### How can we reproduce it?
Start crossplane with --poll-interval=5m and observe that all reconciliation continues to be at 1m intervals

Crossplane version: 1.11.0

I tested a fix for the Claim and Composite reconcilers and verified that it works correctly.

Should the specified pollInterval apply to all reconcilers in the core crossplane pod:

CompositeResourceDefinitions
Compositions
CompositionRevisions
Configurations
ConfigurationRevisions
ControllerConfigs
Locks
Providers
ProviderRevisions

I think the short term answer is "yes" - reconciling these at 1m intervals is a lot of work that is probably unnecessary in many environments.  I think individual options for each type is probably overkill, but it would be useful to be able to have some control over the global poll interval.

I can submit a PR but I'd like to get some feedback on what the scope of the change should be.

bobh66:

Add pollInterval option to the Claim and Composite Reconcilers

Fixes #3761 

I have:

Tested in a development environment and verified that claim and composite resources are reconciled every 5m when the --poll-interval=5m parameter is included in the Crossplane args

### DIFF
--- a/internal/controller/apiextensions/definition/reconciler.go
+++ b/internal/controller/apiextensions/definition/reconciler.go
@@ -39,7 +39,6 @@ import (
 	"github.com/crossplane/crossplane-runtime/pkg/controller"
 	"github.com/crossplane/crossplane-runtime/pkg/errors"
 	"github.com/crossplane/crossplane-runtime/pkg/event"
-	"github.com/crossplane/crossplane-runtime/pkg/feature"
 	"github.com/crossplane/crossplane-runtime/pkg/logging"
 	"github.com/crossplane/crossplane-runtime/pkg/meta"
 	"github.com/crossplane/crossplane-runtime/pkg/ratelimiter"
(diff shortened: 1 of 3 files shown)


internal/controller/apiextensions/definition/reconciler.go
```diff
@@ -403,7 +403,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 			"desired-version", desired.APIVersion))
 	}
 
-	ro := CompositeReconcilerOptions(r.options.Features, d, r.client, r.log, r.record)
+	ro := CompositeReconcilerOptions(r.options.Features, d, r.client, r.log, r.record, r.options.PollInterval)
```
negz, line 406: Perhaps `CompositeReconcilerOptions` should just take `r.options` as an argument?

bobh66: I thought about that but wasn't sure how much to refactor.  I'll make that change.

bobh66: Of course now it has four of the six Reconciler components as arguments, so it almost seems more efficient to just make it 
```
func (r *Reconciler) CompositeReconcilerOptions(d *v1.CompositeResourceDefinition) []composite.ReconcilerOption {
```
but I don't know if that has other undesirable consequences.

negz: I don't feel strongly about whether we go with a method or a function here.

bobh66: There is less churn leaving it as a function so I went with that.

negz (approved):
Good catch, thanks @bobh66!

hasheddan (approved):

PR #3795, opened 2023-02-23
Composition validating webhook  POC

phisco:

**This is just a POC implementation** of the approach discussed in #3756.
I tried to reduce refactoring as much as possible, but I still had to move a few things around to avoid dependency cycles, ideally part of this should go in crossplane-runtime in the final implementation.

The current implementation:

- only covers `FromCompositeFieldPath` patches, but could be extended to other types of patches
- does not validate Compositions using Functions or unsupported patch types and allows them by default.
- is not reporting the whole error context properly which should be done in order to be able to use this also for a linter or language server.
- detects fields not defined in source or target resources
- detects fields' with mismatching types
- populates the referenced Composite resource's required fields in order to be able to successfully apply patches and detect missing required fields in the rendered resources
- detects required fields set using patches from fields that are not required in the current CompositeResource, which would result in a runtime error if not specified.
 
I have:

- [ ] Run `make reviewable` to ensure this PR is ready for review. ** not passing linting, but as said above, this should be seen as POC code only **

```bash
$ ./cluster/local/kind.sh up
... Cluster spin up ...
$ make
... build binaries ...
$ HELM3_FLAGS="--set webhooks.enabled=true" ./cluster/local/kind.sh helm-install
... install the built crossplane with webhooks enabled ...
$ kubectl apply -f - <<EOF
kind: Provider
metadata:
  name: provider-aws
spec:
  package: xpkg.upbound.io/crossplane-contrib/provider-aws:v0.36.1
---
metadata:
  name: xpostgresqlinstances.database.example.org
spec:
  group: database.example.org
  names:
    kind: XPostgreSQLInstance
    plural: xpostgresqlinstances
  claimNames:
    kind: PostgreSQLInstance
    plural: postgresqlinstances
  connectionSecretKeys:
    - username
    - password
    - endpoint
    - port
  versions:
  - name: v1alpha1
    served: true
    referenceable: true
    schema:
      openAPIV3Schema:
        type: object
        properties:
          spec:
            type: object
            description: "The OpenAPIV3Schema of this Composite Resource Definition."
            properties:
              parameters:
                type: object
                properties:
                  storageGB:
                    type: integer
                    description: "The desired storage capacity of the database, in GB."
                  engine:
                    type: string
                required:
                  - storageGB
            required:
              - parameters
EOF
```

Then, applying an invalid Composition as follows:
``` bash
$ kubectl apply -f - <<EOF
---
kind: Composition
metadata:
  name: xpostgresqlinstances.aws.database.example.org
  labels:
    provider: aws
    guide: quickstart
    vpc: default
spec:
  compositeTypeRef:
    kind: XPostgreSQLInstance
  resources:
    - name: rdsinstance
      base:
        kind: RDSInstance
        spec:
          forProvider:
            dbInstanceClass: db.t2.small
            masterUsername: masteruser
            engineVersion: "12"
            skipFinalSnapshotBeforeDeletion: true
            publiclyAccessible: true
      patches:
        - fromFieldPath: "spec.parameters.engine"
          toFieldPath: "spec.forProvider.engine"
        - fromFieldPath: "spec.parameters.engine"
          toFieldPath: "spec.forProvider.wrong"   ## <- missing field in the RDSInstance resource
        - fromFieldPath: "spec.parameters.region" ## <- missing field in CompositeResourceDefinition
          toFieldPath: "spec.forProvider.region"
        - fromFieldPath: "metadata.uid"    ## <- source field is a string
          toFieldPath: "spec.forProvider.allocatedStorage" ## <- target field is an integer
          toFieldPath: "spec.forProvider.allocatedStorage"
      connectionDetails:
        - fromConnectionSecretKey: port
EOF
```

Will result in the following error, showing it's actually detecting the errors defined above:
```yaml
Error from server (BadRequest): error when creating "STDIN": admission webhook "compositions.apiextensions.crossplane.io" denied the request: invalid composition
resource rdsinstance (0), patch 1: invalid toFieldPath: spec.forProvider.wrong: unable to find field: wrong
resource rdsinstance (0), patch 2: invalid fromFieldPath: spec.parameters.region: unable to find field: region
resource rdsinstance (0), patch 3: field path types do not match: string, integer
```


phisco:
I'm trying to reimplement the rendering logic using the available PTComposer, that would allow us to leverage fully the existing code and be sure to avoid drifting in the future

PR #3804, opened 2023-02-24
Update Pluggable Secret Store proposal

ezgidemirel:
This PR updates the existing pluggable secret store one-pager and changes the deployment part.

I have:

N/A


design/one-pager-pluggable-secret-stores.md
```diff
@@ -70,7 +69,7 @@ spec:
   type: External
```
turkenh, line 69: What do you think about renaming the type as I suggested [before](https://github.com/crossplane/crossplane-runtime/pull/379#discussion_r1111848656)? 

```suggestion
  type: Plugin
```

design/one-pager-pluggable-secret-stores.md
```diff
@@ -70,7 +69,7 @@ spec:
...
   defaultScope: crossplane-system
   external:
-    endpoint: unix:///@crossplane/secret/default.sock  # The default, can be changed.
+    endpoint: external-vault:4040  # Service endpoint of the plugin.
```
turkenh, line 72: nit:

```suggestion
    endpoint: ess-plugin-vault:4040  # Service endpoint of the plugin.
```

design/one-pager-pluggable-secret-stores.md
```diff
@@ -154,47 +153,29 @@ message DeleteOptions {
 
 A _Secret Store Plugin_:
 
-- Will be a binary packaged as a container image.
-- Should run a gRPC server listening at Unix domain socket.
-- May support one or more Secret Stores.
+- Will be a k8s deployment behind a service.
```
turkenh, line 156: > Will be a binary packaged as a container image.

This still holds, indeed.

design/one-pager-pluggable-secret-stores.md
```diff
@@ -154,47 +153,29 @@ message DeleteOptions {
...
+- Should run a gRPC server.
+- Should support one Secret Store.
 - Should deploy CRD(s) for supported store configurations at startup.
-- Will run as a sidecar next to Crossplane and Provider containers.
+- Should have server certificates signed by the kubernetes cluster root CA.
```
turkenh, line 160: Does the plugin really care who signed the certs? Otherwise, let's move this detail to the deployment section.

design/one-pager-pluggable-secret-stores.md
```diff
@@ -154,47 +153,29 @@ message DeleteOptions {
...
+deployment with one exception. We need to generate a server certificate 
+during deployment time and mount it to the plugin pod. This is required to enable
+mTLS communication.
 
-### Deployment
+In addition to the server certificates, we will also generate a client certificate
+for each provider and store it in a k8s secret. These certificates will be
+generated by package manager during the package installation. 
+
+Since Crossplane itself will be a gRPC client, we need to generate and mount
```
turkenh, line 174: nit:

> Since the core Crossplane communicates with the secret stores as well, it will also act as a gRPC client to communicate with the plugin. Hence we need to generate ...

design/one-pager-pluggable-secret-stores.md
```diff
@@ -154,47 +153,29 @@ message DeleteOptions {
...
+client certificates to the Crossplane pod as well. We will create an empty 
+k8s secret with the Crossplane helm chart and mount it to the pod. The init 
+container will then generate the client certificates and store them in the secret.
```
turkenh, line 177: ... similar to how webhook certificates generated today.

design/one-pager-pluggable-secret-stores.md
```diff
@@ -231,13 +212,20 @@ func (ss *SecretStore) WriteKeyValues(ctx context.Context, s *store.Secret, _ ..
...
+       i.e. Vault, and write the secret data.
 
 ## Alternatives Considered
 
+### Deploy Plugin as a Sidecar
+
+Instead of deploying plugins as separate deployments, it was initially proposed to
+deploy them as sidecars. The main motivation behind this was eliminating authentication
+complications and having a secure communication via sockets. However, adding sidecars
+to Crossplane and providers will make the packaging and deployment more complicated.
```
turkenh, line 227: Would you mind elaborating on this a bit? For example:
- We would need to inject plugin sidecars into the core crossplane and to all providers.
- When a new plugin installed at runtime, we would need to inject it to Core Crossplane probably requiring a helm upgrade which couldn't be handled by the package manager unless we build a solution like mutating webhook which diverges from our existing package deployment model.
- When multiple plugins are installed, we would need to have multiple sidecars on multiple nodes. For example, with 3 providers running, if someone want to enable ess plugin for 2 secret stores, we would end up `(3+1) * 2 = 8` plugin processes running as sidecars.

design/one-pager-pluggable-secret-stores.md
```diff
@@ -154,47 +153,29 @@ message DeleteOptions {
...
+- Will be a k8s deployment behind a service.
+- Should run a gRPC server.
+- Should support one Secret Store.
 - Should deploy CRD(s) for supported store configurations at startup.
-- Will run as a sidecar next to Crossplane and Provider containers.
+- Should have server certificates signed by the kubernetes cluster root CA.
+- Should require and verify client certificates.
 
-#### Plugin Binary vs Secret Store Implementation
+### Deployment
```
turkenh, line 163: Could you add some details for **Packaging** as well, e.g. we will define a new package type called `SecretStorePlugin` etc.

design/one-pager-pluggable-secret-stores.md
```diff
@@ -155,46 +148,40 @@ message DeleteOptions {
 A _Secret Store Plugin_:
 
 - Will be a binary packaged as a container image.
-- Should run a gRPC server listening at Unix domain socket.
-- May support one or more Secret Stores.
+- Will be a k8s deployment behind a service.
```
negz, line 151: Does it have to be a deployment behind a service? Could it just be any gRPC endpoint?

To be clear I think using deployments makes sense in practice, but I would not artificially limit Crossplane to _only_ be able to communicate with a secret store plugin via a Kubernetes service.

design/one-pager-pluggable-secret-stores.md
```diff
@@ -155,46 +148,40 @@ message DeleteOptions {
...
-sandboxed environments like [gVisor], this [may not work] depending on the
-isolation level.
-
-[Abstract Unix Domain Sockets] on the other hand does not require sharing a path
-on the filesystem and isolated at [network namespace] level. Given that
-containers in the same Pod uses a shared [network namespace], we don't need
-any specific setup and simply be able to communicate with a plugin sidecar over
-a socket like `@crossplane/secret/default.sock`.
+For deploying External Secret Store plugins, we will introduce a new package type
+`Secret Store Plugin`. This package type will be similar to the existing `Provider`
```
negz, line 181: Perhaps just `SecretStore`?

design/one-pager-pluggable-secret-stores.md
ezgidemirel: removed the limitation

ezgidemirel:
Thanks @turkenh and @negz for the review! It's ready for the second pass.

turkenh (approved):
Thank you @ezgidemirel!

design/one-pager-pluggable-secret-stores.md
```diff
@@ -155,46 +148,39 @@ message DeleteOptions {
...
-[Abstract Unix Domain Sockets] on the other hand does not require sharing a path
-on the filesystem and isolated at [network namespace] level. Given that
-containers in the same Pod uses a shared [network namespace], we don't need
-any specific setup and simply be able to communicate with a plugin sidecar over
-a socket like `@crossplane/secret/default.sock`.
+For deploying External Secret Store plugins, we will introduce a new package type
+`Secret Store`. This package type will be similar to the existing `Provider`
+type. It will contain a CRD to configure the secret store plugin (e.g., `VaultConfig`), 
+a k8s service and a deployment. Required RBAC resources will be generated by the
+package manager.
```
turkenh, line 183: There are some concerns about bringing `SecretStore` at the same level as `Providers` or `Configurations` as an extension type since it is not directly related to the _Platform API authoring experience_ but rather more about hosting/running the Control plane. For example, in a Platform Configuration package, you can define dependencies to other Configurations or Providers; however, it does not make sense to define a dependency to a Secret Store since compositions do not require specific secret store types, i.e. agnostic to where secrets are stored.

Stating this, I think it makes sense to start with some documented procedure to install secret store plugins and consider defining it as a package type (and automating this procedure with the package manager) after we feel confident about the above concern and see that it is worth the effort.

We can have a helm chart (instead of an XP package) installing the plugin server, service, and other manifests and document the installation steps accordingly.

turkenh (changes requested):

design/one-pager-pluggable-secret-stores.md
ezgidemirel: removed relevant part

turkenh (approved):

PR #3815, opened 2023-03-02
ci: scheduled trivy image scan

phisco:

This patch adds a nightly workflow, scanning the last 3 supported releases' last built images using trivy, pushing the generated sarif file in case of vulnerabilities reported to the Github security tab and uploading it as a pipeline's artefact for later consumption if needed.

I have:

Deployed to my fork.

### DIFF
--- a/.github/workflows/scan.yaml
+++ b/.github/workflows/scan.yaml
@@ -1,0 +1,137 @@
+name: Trivy Scan
+on:
+  workflow_dispatch: {}
+  schedule:
+    # run every day at 3:07am UTC
+    - cron: '7 3 * * *'
+
+env:
+  DOCKER_USR: ${{ secrets.DOCKER_USR }}
+
+jobs:
+  generate-matrix:
+    runs-on: ubuntu-latest
+    outputs:
+      versions: ${{ steps.get-releases.outputs.versions}}
+      supported_releases: ${{ steps.get-releases.outputs.supported_releases }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # v3
+        with:
+          fetch-depth: 0
+
+      - name: Get Last 3 Releases
+        id: get-releases
+        shell: bash
+        ## find the 3 latest supported releases and their latest patch releases, storing them in the steps' outputs
+        run: |
+          # get the last 3 releases in "vX.Y" form
+          supported_releases="$(git for-each-ref --sort='-committerdate' --format='%(refname:short)' --count=3 'refs/remotes/origin/release-*' | sed 's/.*release-/v/g')"
+          if [ -z "$supported_releases" ]
+          then
+            echo "DEBUG: No supported releases found"
+            echo "DEBUG: $(git for-each-ref 'refs/remotes')"
+            exit 1
+          fi
+
+          echo "DEBUG: ${supported_releases}"
+
+          # get the latest non-rc tag for each release
+          tags=""
+          while IFS= read -r version; do
+            tag="$(git for-each-ref --sort=-taggerdate --count=1 'refs/tags/'${version}'.[\!-rc.*]' --format='%(tag)')"
+            if [ -z "$tag" ]
+            then
+              echo "No tags found for version ${version}, ${tag}"
+              echo "DEBUG: $(git for-each-ref 'refs/tags')"
+              exit 1
+            fi
+            tags="${tags} ${version}=${tag}"
+          done <<< "${supported_releases}"
+
+          echo "DEBUG: ${tags}"
+
+          # build a JSON formatted list of all the supported releases for crossplane/crossplane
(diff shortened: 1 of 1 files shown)


phisco:
Here an example of pipeline run: https://github.com/phisco/crossplane/actions/runs/4315640586
Screenshot in case the above run disappears:
image
And here a screenshot of the security tab:
image
Each alert then will appear like this, the multiple lines in the `Analysis origins` box are only due to multiple tests, only the `crossplane/crossplane:v1.9.1` line should appear on a clean repo:
image

More details about GitHub's Code scanning alerts can be found [here](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/managing-code-scanning-alerts-for-your-repository), as you can see alerts can be acknowledged and either dismissed or handled.

phisco:
More details about GitHub's Code scanning alerts can be found [here](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/managing-code-scanning-alerts-for-your-repository), as you can see alerts can be acknowledged and either dismissed or handled.

** OPIONION REQUEST: ** I made it so that the pipeline fails in case of CRITICAL or HIGH vulnerabilities present in any of the scanned images, so to acknowledge and ignore specific vulnerabilities we would need to explicitly mark them as ignored via a config file, while if pipelines were always passing the full handling would be done via the Github Security tab. I like having a red pipeline in case of vulnerabilities found, but I would also prefer the handling mechanism the Security tab offers more than using a config file, so I'm not sure whether to keep the failing pipelines or remove that. Any opinion?

negz:
> I like having a red pipeline in case of vulnerabilities found, but I would also prefer the handling mechanism the Security tab offers more than using a config file, so I'm not sure whether to keep the failing pipelines or remove that. Any opinion?

Would this potentially cause master builds (not just PRs) to fail if a new vulnerability was found? Would that prevent us pushing vulnerable artifacts? That seems like a good thing. I don't feel strongly though.

.github/workflows/scan.yaml
```diff
@@ -0,0 +1,144 @@
+name: Trivy Scan
+on:
+  workflow_dispatch: {}
+  schedule:
+    # run every day at 3:07am UTC
+    - cron: '7 3 * * *'
+
+env:
+  GOMPLATE_VERSION: v3.11.4
```
negz, line 9: What's Gomplate? Do we need this?

.github/workflows/scan.yaml
```diff
@@ -0,0 +1,144 @@
...
+    outputs:
+      versions: ${{ steps.get-releases.outputs.versions}}
+      supported_releases: ${{ steps.get-releases.outputs.supported_releases }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # v3
+        with:
+          fetch-depth: 0
+
+      - name: Get last 3 supported releases and their latest tags
```
negz, line 24: Nit: I think all of our other steps use brief title case names.

.github/workflows/scan.yaml
phisco: that was a leftover of a previous attempt at generating some summary from trivy's output, removed 👍

.github/workflows/scan.yaml
phisco: Changed all of them to be shorter and title case 👍

phisco:
> I'd let the pipeline run all the way even if vulnerabilities are found, given my understanding that that gives us better insight/tooling to manage them.

Totally agree, changed it to always give a green light so that we can handle everything from the security tab 👍

.github/workflows/scan.yaml
```diff
@@ -0,0 +1,140 @@
...
+
+      - name: Run Trivy Vulnerability Scanner
+        uses: aquasecurity/trivy-action@8bd2f9fbda2109502356ff8a6a89da55b1ead252 # 0.9.1
+        with:
+          image-ref: ${{ matrix.image }}:${{ env.tag }}
+          format: 'sarif'
+          ignore-unfixed: true
+          severity: HIGH,CRITICAL
+          output: 'trivy-results.sarif'
+          limit-severities-for-sarif: true
```
phisco, line 126: We might consider removing the limits, it's probably better to handle them downstream in the Security tab either discarding them or handling them even if not fixed or lower serverity:
```suggestion
          image-ref: ${{ matrix.image }}:${{ env.tag }}
          format: 'sarif'
          output: 'trivy-results.sarif'
```

phisco:
updated pipelines screenshot:
image

lsviben (approved):
Went through how this looks like live with @phisco , and ran the commands locally to test them, all looks good. Seems like an useful addition, especially as we are building on our security processes.

turkenh (approved):
Looking good, great work @phisco 🙌

PR #3822, opened 2023-03-06
One-pager design for ignore changes on updates / granular management policies

lsviben:

Extended the [initial proposal](https://github.com/crossplane/crossplane/issues/3516) with an one-pager design doc. 

Iterating over the solution we came to a more generalized approach that introduces granular management policies through which features like ObserveOnly and ignore changes can be realized.

Fixes https://github.com/crossplane/crossplane/issues/3516

I have:

N/A

### DIFF
--- a/design/one-pager-ignore-changes.md
+++ b/design/one-pager-ignore-changes.md
@@ -1,0 +1,681 @@
+# Ignore changes
+
+* Owners: Lovro Sviben (@lsviben)
+* Reviewers: @turkenh, @negz
+* Status: Draft
+
+## Background
+
+When Crossplane manages a resource, it reconciles all the parameters
+under  `spec.forProvider` to the external providers both during the
+creation and subsequent updates to the managed resource. But in some cases,
+a parameter of the external resource could change due to it being managed
+outside Crossplane. In the current state, if that happens, Crossplane
+will try to "fix" the situation, and update the parameter back to the one it
+has in the `spec.forProvider` as it is designed to be the only source of truth.
+
+Some examples of this case:
+- [AutoScalingGroup DesireCapacity], where when using an autoscaler, the
+  parameter gets changed externally.
+- [DynamoDB Tables] with using autoscaling ([Policy] + [Target]). The
+  autoscaler scales the read and write capacity, and then Crossplane resets
+  it back to the `readCapacity` and `writeCapacity` defined on the Table,
+  which effectively makes autoscaling useless.
+- [NodeGroup] while using a node autoscaler. There is a race condition
+  between crossplane and node autoscaler. Node autoscaler increases the
+  `desiredSize` due to resource requirement but Crossplane decreases it to
+  `desiredSize` set in composition.
+- GCP [Node Pool], where when `management.autoUpgrade` is enabled. After
+  creation, the provider late initializes `spec.forProvider.version`,
+  which when the nodes get upgraded starts conflicting with the 
+  external nodes version. This causes an endless loop of downgrades
+  and upgrades.
+- Tags are a use case that affects multiple resources. Many cloud users
+  have external systems that add tags to resources for billing, tracking
+  or similar purposes. This can conflict with the tags set in 
+  `spec.forProvider.tags`.
+- There is an use case with [AzureAD Group] where initially the group is
+  created certain with `members` and `owners` but is then managed externally
+  over time as members need to be added or removed. For now the
+  workaround is to use terraform provider with [ignore_changes].
+
+In these cases, it would make sense to make Crossplane ignore select
+parameters during updates. Terraform addresses this use case with its
+[ignore_changes] functionality.
+
+The obvious way would be to just not set the field in the `spec.forProvider`.
+However, due to [Late Initialization], some fields are being set in
+`spec.forProvider` without any user input. Plus there are some cases
+where the field is required/wanted upon creation, but not on updates.
+
+So in this document, the aim will be to introduce a solution in Crossplane
+for ignoring some managed resource parameters during updates.
+
+## Goals
+
+* Design a functionality that allows for ignoring select fields during
+  resource updates.
+
+## Proposal
+
+Proposed solution is to rework the new [managementPolicy] feature which came
+with the [ObserveOnly] feature and transform it into a set of enum values
+representing what Crossplane should do with the managed resource.
+
+- `Create` - Create the external resource using `spec.forProvider`
+  and `spec.initProviders` fields.
+- `Update` - Update the external resource using `spec.forProvider` fields.
+- `Delete` - Delete the external resource when the managed resource is
+  deleted.
+- `Observe` - Update the `status.atProvider` to reflect the state of the
+  external resource.
+- `LateInitialize` - Update unspecified `spec.forProvider` fields to reflect
+  the state of the external resource, typically with the defaults from the cloud
+  provider.
+
+This will allow users to fine-tune how Crossplane manages the external
+resource, in a manner which is very explicit and easy to understand.
+
+Some examples on how the management policies would work and how they would
+replace the current `managementPolicy` and `deletionPolicy`:
+
+```yaml
+# Default
+spec:
+  managementPolicy: FullControl
+  deletionPolicy: Delete
+
+# would be replaced with:
+spec:
+  managementPolicy: ["Create", "Update", "Delete", "Observe", "LateInitialize"]
+  # or
+  managementPolicy: ["*"]
+
+# ObserveOnly
+spec:
+  managmentPolicy: ObserveOnly
+
+# would be replaced with:
+spec:
+  managementPolicy: ["Observe"]
+
+# OrphanOnDelete
+spec:
+  managementPolicy: OrphanOnDelete
+
+# would be replaced with:
+spec:
+  managementPolicy: ["Create", "Update", "Observe", "LateInitialize"]
+
+# pause can be achieved by setting managementPolicy to empty list instead of
+# using the annotation
+spec:
+  managementPolicy: []
+
+# Turn off late initialization
+spec:
+  managementPolicy: ["Create", "Update", "Delete", "Observe"]
+```
+
+In addition to the new management policy, we will also add a new field
+`spec.initProvider` which would contain parameters that should only be
+used at the creation time. This would allow users to specify fields that
+should be ignored on updates, but still be used during creation.
+
+## Implementation
+
+### Management policy
+
+The new management policy would be implemented in the `crossplane-runtime`.
+Most of the work will be in the [Managed Reconciler], where we will need to add
+management policy checks to sections of code where we create, update etc.
+
+For example:
+- Create - [Create Section]
+- Update - [Update Section]
+- Delete - [Delete Section]
+- Late Init - [Late Init Section]
+- Observe - this one is tricky because we still need to observe/get a resource,
+  so we just need to make sure that we don't update `status.atProvider`.
+
+### Deletion policy
+
+`deletionPolicy` was planned to be deprecated in favour of the new management
+policies according to [the ObserveOnly design doc.][ObserveOnly], but still
+retain some functionality if a non-default value was set. In practice, it
+meant that if the `deletionPolicy` was set to `Orphan`, and the
+`managementPolicy` set to `FullControl`, the external resource would be
+orphaned.
+
+In the new design, we could still follow this approach, by orphaning the
+resource even if the `managementPolicy` includes `Delete`, if the
+`deletionPolicy` is set to `Orphan`, until we entirely remove the deletion
+policy.
+
+Keep in mind that the `deletionPolicy` still keeps its full functionality if the
+management policy alpha feature is not enabled.
+
+### ["Create", "Update", "Delete", "Observe", "LateInitialize"] or ["*"] by default
+
+Except for the cosmetic value, this decision affects how future additions to
+the management policy would be handled. If we go with the `["*"]` approach,
+every new management policy would be automatically included in the resources,
+which may be what we want. Similarly, if we go with the `["Create", "Update",
+"Delete", "Observe", "LateInitialize"]` approach, every new management policy
+would need to be added manually or through a migration.
+
+So this is something to keep in mind when developing new management policies.
+For this proposal, we will go with the `["*"]` approach as its more
+future-proof.
+
+### Migrating existing resources
+
+The `managementPolicy` feature is alpha, so it should be ok to break the
+API. The combinations of `managementPolicy` and `deletionPolicy` would look
+like this in the new `managementPolicy` field.
+
+| managementPolicy | deletionPolicy | new managementPolicy                              |
+|------------------|----------------|---------------------------------------------------|
+| FullControl      | Delete         | ["*"]                                             |
+| FullControl      | Orphan         | ["Create", "Update", "Observe", "LateInitialize"] |
+| OrphanOnDelete   | Delete         | ["Create", "Update", "Observe", "LateInitialize"] |
+| OrphanOnDelete   | Orphan         | ["Create", "Update", "Observe", "LateInitialize"] |
+| ObserveOnly      | Delete         | ["Observe"]                                       |
+| ObserveOnly      | Orphan         | ["Observe"]                                       |
+
+As this will be a breaking change, if users want to keep the old
+`managementPolicy` behaviour, we suggest pausing the reconciliation of the MR,
+upgrading Crossplane, and then updating the `managementPolicy` to the desired
+value before unpausing the reconciliation.
+
+In reality this is only needed for the `ObserveOnly` and
+(`OrphanOnDelete` + `Delete`) combinations, as the `FullControl` as in other
+cases the default new management policy (`["*"]`) won't change the behaviour.
+
+### initProvider
+
+The idea here was to have a field that would be used only during creation,
+called `initProvider`. The `initProvider` field would have the same schema
+as `forProvider`. The `initProvider` fields would be merged with the
+`forProvider` fields during the `Create` step of the managed resource
+reconciliation loop. Fields that are specified in both `initProvider` and
+`forProvider` give precedence to the `forProvider` fields.
+
+```yaml
+apiVersion: eks.aws.crossplane.io/v1alpha1
+kind: NodeGroup
+metadata:
+  name: my-group
+spec:
+  initProvider:
+    scalingConfig:
+      desiredSize: 1
+  forProvider:
+    region: us-east-1
+    scalingConfig:
+      maxSize: 5
+      minSize: 1
+```
+
+This would allow users to specify fields that would only be used on creation,
+but not on updates.
+
+When it comes to the implementation, the handling would all need to be done
+on the provider level. The `initProvider` fields should only be used in the
+`Create` step of the managed resource reconciliation loop.
+
+Support for `initProvider` would need to be a manual change in each provider.
+From our side we could provide a helper function to merge the `initProvider`
+and `forProvider` fields which would be used in the `Create` step.
+
+We know that for now the `initProvider` field would be useful just for a 
+handful of cases that we are aware of. So provider developers do not need to
+support `initProvider` for all resources right away. Rather, they can add
+support for those resources that are known to require it, and later on based
+on a request basis.
+
+### Feature Gating
+
+The new management policy would use the same feature gate already introduced
+by `ObserveOnly`. So it would be an alpha feature that will be off by default
+and will be controlled by `--enable-alpha-management-policies` flag in
+Providers. It will also enable the `initProvider` feature.
+
+### Examples how the new functionality would affect the use cases
+
+Some use cases mentioned in the beginning of the document would be
+solved just by omitting the management policy `LateInitialize`. For
+others, the `initProvider` field would need to be used, mostly because
+the fields in question are required on creation.
+
+#### AutoScalingGroup
+
+The `desiredCapacity` is not required on creation, but populated by late
+initialization. It can be solved just by omitting `Late Initialize` management
+policy.
+
+```yaml
+spec:
+  managementPolicy: ["Create", "Update", "Delete", "Observe"]
+  forProvider:
+    maxSize: 5
+    minSize: 1
+    launchConfigurationNameRef:
+      name: sample-launch-config
+    ...
+```
+
+#### DynamoDB Table
+
+The `readCapacity` and `writeCapacity` are not required, but they are required
+if the `billingMode` is set to `PROVISIONED`. So the `initProvider` field
+would need to be used alongside omitting `LateInitialize` management policy.
+
+```yaml
+spec:
+  managementPolicy: ["Create", "Update", "Delete", "Observe"]
+  initProvider:
+    readCapacity: 1
+    writeCapacity: 1
+  forProvider:
+    billingMode: PROVISIONED
+...
+```
+
+#### EKS NodeGroup
+
+The `scalingConfig.desiredSize` is required so `initProvider` would need
+to be used alongside omitting `LateInitialize` management policy. This way
+the autoscaler would be able to control the `desiredSize` after creation.
+
+```yaml
+spec:
+  managementPolicy: ["Create", "Update", "Delete", "Observe"]
+  initProvider:
+    scalingConfig:
+      desiredSize: 1
+  forProvider:
+    region: us-east-1
+    scalingConfig:
+      maxSize: 5
+      minSize: 1
+```
+
+#### GCP Node Pool
+
+Just omitting the `LateInitialize` management policy would be enough as the
+`version` is not required on creation.
+
+```yaml
+spec:
+  managementPolicy: ["Create", "Update", "Delete", "Observe"]
+  forProvider:
+  ...
+```
+
+#### Azure AD Group
+
+The use case where `members` and `owners` want to be set on creation and
+then ignored on updates would be solved by using `initProvider` alongside
+omitting the `LateInitialize` management policy.
+
+Example:
+```yaml
+spec:
+  managementPolicy: ["Create", "Update", "Delete", "Observe"]
+  initProvider:
+    members:
+      - user1
+      - user2
+    owners:
+      - user3
+  forProvider:
+    displayName: my-group
+    securityEnabled: true
+```
+
+#### Tags
+
+Setting initial tags should work with `initProvider`. However, it seems that
+the providers are setting some default tags in the into the `forProvider`
+field in the [Initialize] step and are updating the resource, so this would
+need to be changed to use `initProvider` instead, or we can check if those
+tags are actually needed or just skip setting them similarly to how
+ObserveOnly does it.
+
+Ref: [Upjet Initialize] or [AWS community provider tag example].
+
+## Solution alternative considered
+
+### PartialControl management policy + initProvider
+
+Proposed solution is to use the new [managementPolicy] field which came with
+the [ObserveOnly] feature and add a new management policy that will
+skip late initialization. The loss the information that the
+late initialization was providing would be offset by the `status.atProvider`
+which contains the resource state on the provider side and is being
+added alongside [ObserveOnly].
+
+In addition to the new management policy, we will also add a new field
+`spec.initProvider` which would contain parameters that should only be
+used at the creation time. This would allow users to specify fields that
+should be ignored on updates, but still be used during creation. This can
+be added as a separate feature.
+
+#### Management policy
+
+Proposed name for the new management policy is `PartialControl`, which
+would fit with the current `FullControl` and insinuate that the resource
+is not fully managed by Crossplane. Some other proposals for the naming
+are: `ExplicitControl`, `SelectiveControl`, `IgnoreLateInit`.
+
+The new management policy would be implemented in the `crossplane-runtime`
+and would just ignore the [ResourceLateInitialize] condition and
+skip the managed resource update [here][Late Initialization Update] if
+the management policy is set to `PartialControl`.
+
+#### Deletion policy
+
+`DeletionPolicy` was planned to be deprecated in favour of the new management
+policies according to [the ObserveOnly design doc.][ObserveOnly] However,
+with `PartialControl` it's not clear if it should delete or orphan the
+external resources. So to avoid creating `PartialControl` and
+`PartialControlOrphanOnDelete` management policies, we will keep the
+`DeletionPolicy` to be able to specify the behaviour. In that case, we will
+need to remove the `OrphanOnDelete` management policy as it won't be needed
+anymore. As it was an alpha feature, this should not be a problem, but we
+should still make sure to communicate this change to the users.
+
+So the management policies would be:
+- `FullControl`
+- `PartialControl`
+- `ObserveOnly`
+
+#### initProvider
+
+The idea is the same as in the proposal.
+
+#### Examples how the new functionality would affect the use cases
+
+Some use cases mentioned in the beginning of the document would be
+solved just by setting the new management policy to `PartialControl`. For
+others, the `initProvider` field would need to be used, mostly because
+the fields in question are required on creation.
+
+#### AutoScalingGroup
+
+The `desiredCapacity` is not required on creation, but populated by late
+initialization. It can be solved just using the `PartialControl` management
+policy.
+
+```yaml
+spec:
+  managementPolicy: PartialControl
+  forProvider:
+    maxSize: 5
+    minSize: 1
+    launchConfigurationNameRef:
+      name: sample-launch-config
+...
+```
+
+#### DynamoDB Table
+
+The `readCapacity` and `writeCapacity` are not required, but they are required
+if the `billingMode` is set to `PROVISIONED`. So the `initProvider` field
+would need to be used alongside `PartialControl` management policy.
+
+```yaml
+spec:
+  managementPolicy: PartialControl
+  initProvider:
+    readCapacity: 1
+    writeCapacity: 1
+  forProvider:
+    billingMode: PROVISIONED
+...
+```
+
+#### EKS NodeGroup
+
+The `scalingConfig.desiredSize` is required so `initProvider` would need
+to be used alongside `PartialControl` management policy. This way the
+autoscaler would be able to control the `desiredSize` after creation.
+
+```yaml
+spec:
+  managementPolicy: PartialControl
+  initProvider:
+    scalingConfig:
+      desiredSize: 1
+  forProvider:
+    region: us-east-1
+    clusterNameRef:
+      name: sample-cluster
+    subnetRefs:
+      - name: sample-subnet1
+    nodeRoleRef:
+      name: somenoderole
+    scalingConfig:
+      maxSize: 5
+      minSize: 1
+```
+
+#### GCP Node Pool
+
+Just using the `PartialControl` management policy would be enough as the
+`version` is not required on creation.
+
+```yaml
+spec:
+  managementPolicy: PartialControl
+  forProvider:
+  ...
+```
+
+#### Azure AD Group
+
+The use case where `members` and `owners` want to be set on creation and
+then ignored on updates would be solved by using `initProvider` alongside
+`PartialControl` management policy.
+
+Example:
+```yaml
+spec:
+  managementPolicy: PartialControl
+  initProvider:
+    members:
+      - user1
+      - user2
+    owners:
+      - user3
+  forProvider:
+    displayName: my-group
+    securityEnabled: true
+```
+
+#### Tags
+
+Setting initial tags should work with `initProvider`. However, it seems that
+the providers are setting some default tags in the into the `forProvider`
+field in the `Initialize` step and are updating the resource, so this would
+need to be changed to use `initProvider` instead, or we can check if those
+tags are actually needed or just skip setting them similarly to how
+ObserveOnly does it.
+
+Ref: [Upjet Initialize] or [AWS community provider tag example].
+
+### ignoreChanges field
+
+The idea is to introduce a new `ignoreChanges` string array field in the
+`spec` of the managed resource, that would work similarly
+to Terraform's [ignore_changes]. The items in the array would be the field
+paths of the fields that should be ignored on updates.
+
+```yaml
+apiVersion: eks.aws.crossplane.io/v1alpha1
+kind: NodeGroup
+metadata:
+  name: my-group
+spec:
+  ignoreChanges:
+    - scalingConfig.desiredSize
+  forProvider:
+    region: us-east-1
+    clusterNameRef:
+      name: sample-cluster
+    subnetRefs:
+      - name: sample-subnet1
+    nodeRoleRef:
+      name: somenoderole
+    scalingConfig:
+      desiredSize: 1
+      maxSize: 5
+      minSize: 1
+    updateConfig:
+      maxUnavailablePercentage: 50
+      force: true
+  providerConfigRef:
+    name: example
+```
+
+The solution should affect just the Update part of the Crossplane lifecycle,
+so create and late initialization should work as before, even though a
+field is marked under `ignoreChanges`. Additionally, as Crossplane will soon
+support `status.atProvider`, users will be able to observe the actual values
+of the ignored fields.
+
+#### Ignoring required fields
+
+As there is no standard to distinguish which required fields are
+required for create and update and which just for creat, we can only
+advise to use caution when putting required fields into `ignoreChanges`.
+If such a need arises, it should preferably be handled as a special case
+on the provider side, and not through the `ignoreChanges` feature.
+
+#### Implementation
+
+The implementation would be done on the provider level, so we would need to
+update all the providers over time to use the new field.
+
+For providers using Upjet the solution will be to leverage Terraform's
+[ignore_changes] lifecycle field. So Upjet generation should be updated
+with a step that transfers the `spec.ignoreChanges` field into Terraform's
+`ignore_changes`. Something like this in the Upjet method that writes the
+main TF file:
+
+```go
+// WriteMainTF writes the content main configuration file that has the desired
+// state configuration for Terraform.
+func (fp *FileProducer) WriteMainTF() error {
+
+    fp.parameters["lifecycle"] = map[string]interface{}{
+        // If the resource is in a deletion process, we need to remove the
+        // deletion protection.
+        "prevent_destroy": !meta.WasDeleted(fp.Resource),
+        // Add fields which should be ignored on updates.
+        "ignore_changes": fp.Resource.GetIgnored(),
+    }
+...
+```
+
+For other providers we can implement a helper function in the
+[crossplane-runtime] repo that would use the existing code in
+[crossplane-runtime/fieldpath] to unset the fields of a managed resource
+that are under `ignoreChanges`. That way provider maintainers could use this
+function in the Update method without the need to invent their own solution.
+
+Helper function would look something this (WIP):
+```go
+func UnsetIgnored(managed resource.Managed) error {
+	p, err := PaveObject(managed)
+	if err != nil {
+		return err
+	}
+	
+	for _, s := range managed.GetIgnored() {
+		err = p.DeleteField(s)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+```
+
+#### Feature Gating
+
+Similar to all other new features being added to Crossplane, we will ship this
+new policy as an alpha feature that will be off by default and will be
+controlled by the `--enable-alpha-ignore-changes` flag in Providers.
+
+This will not prevent the field from appearing in the schema of the managed
+resources. However, we will ignore the `spec.ignoreChanges` when the feature
+is not enabled.
+
+#### Alternative feature gating approach
+
+To avoid changing the API, we could ship the `ignoreChanges` field as an
+optional annotation as a first solution. This would allow us to survey
+users and see how much this feature is useful before adding it to the API.
+
+#### Why this solution was not chosen
+
+This solution was originally chosen, but it was later decided that it would
+be better to have a clear distinction of what is being reconciled to the
+external provider and what is not. Imagine a case where a field is both set
+in `spec.forProvider` and `spec.ignoreChanges`. This could happen as an
+effect of late initialization or if the user sets the field in both places.
+In that case, the field would be ignored on updates, but could cause confusion
+as it would still be in the`spec.forProvider` field. Of course, we could also
+introduce something like `PartialControl` management policy that turns off
+late init to minimize that case.
+
+That's where the `initProvider` solution has a slight advantage, as it clear
+that the field is just used on create and not on updates. Therefore, we can
+keep `spec.forProvider` as the source of truth for the rest of the resource
+lifecycle.
+
+All in all, this solution is still valid and the decision to go with
+the `initProvider` is because it seems it would bring less confusion to the
+users.
+
+### Case by case
+
+This approach would just tackle the ignored fields on a case-by-case basis,
+instead of implementing a generic solution. The premise here is that not many
+resources would need to use this feature.
+
+For instance in the [NodeGroup] example, we would add a field `initialSize`,
+along with the current `desiredSize` and the updates could be handled in code.
+
+#### Why this solution was not chosen
+
+While this approach is valid, making changes case-by-case could lead to 
+bad user experience as they would need time to identify the issue and wait 
+for provider code change, instead of just setting a field. On the other hand,
+if this issue is not that widespread, we could have an easy fix.
+
+[AutoScalingGroup DesireCapacity]: https://doc.crds.dev/github.com/upbound/provider-aws/autoscaling.aws.upbound.io/AutoscalingGroup/v1beta1@v0.21.0#spec-forProvider-desiredCapacity
+[DynamoDB Tables]: https://doc.crds.dev/github.com/upbound/provider-aws/dynamodb.aws.upbound.io/Table/v1beta1@v0.24.0
+[Policy]: https://doc.crds.dev/github.com/upbound/provider-aws/appautoscaling.aws.upbound.io/Policy/v1beta1@v0.24.0
+[Target]: https://doc.crds.dev/github.com/upbound/provider-aws/appautoscaling.aws.upbound.io/Target/v1beta1@v0.24.0
+[NodeGroup]: https://doc.crds.dev/github.com/upbound/provider-aws/eks.aws.upbound.io/NodeGroup/v1beta1@v0.24.0
+[Node Pool]: https://doc.crds.dev/github.com/upbound/provider-gcp/container.gcp.upbound.io/NodePool/v1beta1@v0.28.0
+[AzureAD Group]: https://marketplace.upbound.io/providers/upbound/provider-azuread/v0.5.0/resources/groups.azuread.upbound.io/Group/v1beta1
+[Late Initialization]: https://docs.crossplane.io/v1.11/concepts/managed-resources/#late-initialization
+[Managed Reconciler]: https://github.com/crossplane/crossplane-runtime/blob/1316ae6695eec09cf47abdfd0bc6273aeaab1895/pkg/reconciler/managed/reconciler.go
+[Create section]: https://github.com/crossplane/crossplane-runtime/blob/1316ae6695eec09cf47abdfd0bc6273aeaab1895/pkg/reconciler/managed/reconciler.go#L943-L1031
+[Delete section]: https://github.com/crossplane/crossplane-runtime/blob/1316ae6695eec09cf47abdfd0bc6273aeaab1895/pkg/reconciler/managed/reconciler.go#L865-L922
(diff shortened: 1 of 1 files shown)


design/one-pager-ignore-changes.md
```diff
@@ -0,0 +1,199 @@
...
+users will be able to observe the actual values of the ignored fields.
+
+#### Ignoring required fields
+
+Required fields should not be put into `ignoreChanges`, as that could compromise the functionality of the provider. If such
+a need arises, it should be handled as a special case on the provider side, and not through the `ignoreChanges` feature.
+
+Of course, some fields could be marked as required, but they are actually required only for create. As there currently is
+no standard on differentiating them, it is advised to use `ignoreChanges` on required fields only if certain that it's not
+required for update.
```
negz, line 82: This seems unclear to me. Are we saying that required fields cannot have their changes ignored, or not?

design/one-pager-ignore-changes.md
```diff
@@ -0,0 +1,199 @@
...
+be ignored on updates. 
+
+```yaml
+apiVersion: eks.aws.crossplane.io/v1alpha1
+kind: NodeGroup
+metadata:
+  name: my-group
+spec:
+  ignoreChanges:
+    - scalingConfig.desiredSize
```
negz, line 51: Would this have any interaction with the new `managementPolicy` field?

It seems like a bit of a shame to introduce two new concepts (`managementPolicy` and `ignoreChanges`) to address what to me seem like two very similar / closely related problems (essentially OO resources and OO fields).

design/one-pager-ignore-changes.md
```diff
@@ -0,0 +1,199 @@
...
+    updateConfig:
+      maxUnavailablePercentage: 50
+      force: true
+  providerConfigRef:
+    name: example
+```
+
+The great thing about this solution is that it feels natural to add `initProvider` as we already have `forProvider`,
+`atProvider`. But on the other hand a thing that would be bothersome with this solution is that a value is needed in the 
+`initProvider`. A field cannot just be marked to be ignored, without setting the initial value.
```
negz, line 187: > A field cannot just be marked to be ignored, without setting the initial value.

In @turkenh's OO resources design we've extended `status.atProvider` to always contain the current values of all of an MR's fields. So presuming we had:

* `spec.initProvider` - state that is desired only at init/create time.
* `spec.forProvider` - state that is desired all the time.
* `status.atProvider` - observed actual state.

If what we're trying to say is "for this field we want to set an initial value then let the external system take over" then it seems like `spec.initProvider` covers that. If what we're trying to say is "for this field we never have desired state, just let the external system control it", wouldn't we just omit it from `spec.initProvider` _and_ `spec.forProvider` and let it only show up under `stats.atProvider`?

Is the issue (only) late initialization? i.e. That we'd need some way to say "don't late init this field", even though its not in `initProvider` or `forProvider`?

negz (commented):
Thanks for tackling this design @lsviben!

design/one-pager-ignore-changes.md
```diff
@@ -0,0 +1,199 @@
...
+Compositions managing a single external resource. This could open up some interesting new possibilities for Crossplane.
+
+So in this document, the aim will be to introduce a solution in Crossplane for ignoring some managed resource parameters
+during updates.
+
+## Goals
+
+* Design a functionality that allows for ignoring some fields during resource updates.
+* Try and make it as easy as possible to apply the ignoring functionality on existing Crossplane providers. In other words, to 
+design a standard way on how to update the providers code to support `ignore changes`.
```
muvaf, line 36: I think _a standard way_ is more like one of the solutions rather than a goal we should aim for. There might be solutions that are not shared across all MRs that will achieve the same goals.

design/one-pager-ignore-changes.md
```diff
@@ -0,0 +1,199 @@
...
+	}
+	return nil
+}
+```
+
+#### Feature Gating
+
+Similar to all other new features being added to Crossplane, we will ship this
+new policy as an alpha feature that will be off by default and will be
+controlled by the `--enable-alpha-ignore-changes` flag in Providers.
```
muvaf, line 134: The CRDs are statically generated and baked into the provider images so we can't really use a flag to add/remove a field. It could only affect whether the code that would process it should be activated or not.

design/one-pager-ignore-changes.md
```diff
@@ -0,0 +1,199 @@
...
+      maxSize: 5
+      minSize: 1
+    updateConfig:
+      maxUnavailablePercentage: 50
+      force: true
+  providerConfigRef:
+    name: example
+```
+
+The great thing about this solution is that it feels natural to add `initProvider` as we already have `forProvider`,
```
muvaf, line 185: FWIW, the idea behind that naming is that`spec.forProvider` would mean `specifications for provider` and `status.atProvider` would mean `status at provider`. So, `specifications initialization provider` does not really conform the pattern.

design/one-pager-ignore-changes.md
```diff
@@ -0,0 +1,199 @@
...
+would be passed to the Update method.
+
+While this solution would enable the functionality without the need to change all the providers, it has a few 
+drawbacks:
+* It doesn't allow for flexibility or handling special cases in the provider code, as the providers will just receive a 
+managed resource which does not have the ignored fields set. There could be cases where the wish is to not update the field
+on the external resource, but is useful in some other way.
+* Won't work well with Upjet as it uses Terraform underneath which would work better with its native [ignore_changes].
+
+## Solution alternative considered
```
muvaf, line 153: I think another solution to consider could be to handle this problem on case by case basis depending on how many cases we've seen. For example, `NodeGroup` could have `initialSize` along `desiredSize` where both are marked as optional instead of a generic solution and we can manage it in the code.

In my experience, it's quite seldom in Crossplane land compared to Terraform since `ignore_changes` is used way more frequently in Terraform especially to prevent resource deletion & recreation since that's the behavior of TF for fields that cannot be changed whereas we simply do nothing or reject the action via webhook. Hence comparison with how many times `ignore_changes` is seen in HCLs or how it's top-level field in Terraform may be misleading. Ideally, we would gather a ratio like `every 1 of 4 CRDs need this functionality` (I suspect it's way lower) and then decide but that's probably quite challenging to gather other than trying to think of counting the open issues. Maybe one way could be to implement this field as an optional annotation and survey users about how many CRDs they needed to use it in. This could also be a good way to ship it as alpha without a change in CRDs, which is similar to how upstream k8s usually introduces new fields.

design/one-pager-ignore-changes.md
lsviben: My point is that users should be careful with putting required fields into `ignoreChanges` as its not clear from the resource spec if the field is required only in create or both in create and update.

 I dont expect needing some validation for that, just put a note about it into the docs

design/one-pager-ignore-changes.md
lsviben: The interaction would be that if the `managementPolicy` : `ObserveOnly` is set, then there is no point to set `ignoreChanges`. 

You are right, these problems are similar, as you say OO resources and OO fields. `ignoreChanges` just gives a more fine-grained control, and tries to fix a bit different problem, where ownership of an external resource is shared. While ObserveOnly is either owned by Crossplane or by something else. 

Not sure currently how we could bring this concepts closer together, to avoid having different solutions for similar problems.

design/one-pager-ignore-changes.md
lsviben: Right. My plan was to introduce it in the CRD, but enable/disable it per flag in the providers. The option with annotations that you mentioned [in the comment below](https://github.com/crossplane/crossplane/pull/3822/files#r1127672514) would make sense as well if we want to avoid impact before we are sure about this solution

design/one-pager-ignore-changes.md
lsviben: You are right. The issue would mostly be late initialization, I need to check up a bit on late initialization to figure out how we could stop it from late init some fields. I wonder if in the end it turns into `ignoreLateInit` which brings us back to something like `ignoreChanges`

One other thing, I feel marking the field explicity ignored somewhere instead of just omitting it from `spec.initProvider` and `spec.forProvider` gives a type of "documentation" and "safety", so over time nobody decides to add this field to `spec.forProvider`.

design/one-pager-ignore-changes.md
```diff
@@ -0,0 +1,199 @@
+# Ignore changes
+
+* Owners: Lovro Sviben (@lsviben)
+* Reviewers: Crossplane Maintainers
+* Status: Draft
+
+## Background
+
+When Crossplane manages a resource, it reconciles all of the parameters under `spec.forProvider` to the external providers both
```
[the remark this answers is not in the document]
lsviben: Done!

design/one-pager-ignore-changes.md
lsviben: Right, removed the goal as its vague

design/one-pager-ignore-changes.md
stevendborrelli: The main use case I have seen for this functionality is with tagging. Many cloud users have external systems that add tags to resources for billing, tracking or similar purposes.

ferpizza:
I'm currently impacted by the issues we are trying to solve, shared [here](https://github.com/upbound/provider-gcp/issues/249#issuecomment-1465999020)

TL,DR; 3 different fields on the NodePool manifest of provider-gcp are generating issues with GKE's autoUpgrade and autoScaling functionalities.

These 3 fields have something in common. None of them are on my initial manifest and neither is a mandatory field.

So, I'm thinking that another approach to solving this is to avoid updating the original manifest `spec.forProvider` section on every reconcile cycle, and instead add a full manifest of current state on `status.atProvider`

This way, CrossPlane will ensure that only the values set by the user are sustained over time while allowing all other values to flow freely. Additionally, providers and other plugins will be able to get the full info on the resource from the `status.atProvider` section, so nothing will be missed.

lsviben:
> I'm currently impacted by the issues we are trying to solve, shared [here](https://github.com/upbound/provider-gcp/issues/249#issuecomment-1465999020)
> 
> TL,DR; 3 different fields on the NodePool manifest of provider-gcp are generating issues with GKE's autoUpgrade and autoScaling functionalities.
> 
> These 3 fields have something in common. None of them are on my initial manifest and neither is a mandatory field.
> 
> So, I'm thinking that another approach to solving this is to avoid updating the original manifest `spec.forProvider` section on every reconcile cycle, and instead add a full manifest of current state on `status.atProvider`
> 
> This way, CrossPlane will ensure that only the values set by the user are sustained over time while allowing all other values to flow freely. Additionally, providers and other plugins will be able to get the full info on the resource from the `status.atProvider` section, so nothing will be missed.

Thanks for the input! 

Basically your idea would be to turn off `late-initialization` and instead rely on `status.atProvider` which is a new feature coming with Observe-Only Resources. The "ignoring" would be done by just not setting the fields in the `spec.forProvider` and as there is no `late-initialization`, there is no undesired conflict between Crossplane and external resource managers. I hope I got it right?

Ill add your use case and alternative solution suggestion to the proposal

ferpizza:
Your interpretation sounds about right, but I'm a bit new to CrossPlane and I'm not fully familiar with what `late-initialization` is trying to achieve, TBH.

I'm getting some inspiration on ArgoCD, that keeps 2 manifests versions: the one that the user uploads (desired manifest), and the one that is actually running on the cluster (current manifest). 

Then, on every cycle, we would make sure that the current manifests includes all the values in the desired manifest, and reconcile only if there is a deviation between these two.

It's expected that the current manifest will have more values that the desired manifest. And this is OK.
If the user wants to take control over these other values, it only need to add them to their desired manifests and upload changes.

Regarding how this overlaps with [Observe-Only goals](https://github.com/crossplane/crossplane/blob/bd2c98514491bddf5844014c78c9b23a82cdf874/design/design-doc-observe-only-resources.md#goals), I believe this proposal will accomplishes both of them.

negz:
> Basically your idea would be to turn off late-initialization and instead rely on status.atProvider which is a new feature coming with Observe-Only Resources. 

This is along the lines of what I was thinking after our discussion a few weeks back. We can't turn off late-init by default because that would be a breaking API change. We could add a knob to disable it though - for example the new `managementPolicy` field being added in support of OO resources.

At that point we'd have:

* The full observed state in `status.atProvider`.
* (Optionally) The initial 'bootstrap' desired state in `spec.initProvider`.
* (Optionally) The constantly-reconciled desired state in `spec.forProvider`.

If you wanted to allow the external system (e.g. a GKE node pool autoscaler) to control a particular field you would:

* (Optionally) add an initial desired value to `spec.initProvider`.
* Don't specify a desired value under `spec.forProvider`.
* Set `spec.managementPolicy: DoNotLateInit` (hopefully with a better policy name)

One downside of this approach is that you have to opt-out of late init entirely (not just for the fields you care about). I'm not sure how much of an issue that would actually be in practice.

turkenh:
> Set spec.managementPolicy: DoNotLateInit (hopefully with a better policy name)

The default management policy is `FullControl`, meaning that XP will own/control each and every field, full resource control. This is indeed why we are late-initializing; to know the default values and take them under our control (i.e. source of truth).

Stating this, I like the idea of using the [managementPolicy](https://github.com/crossplane/crossplane/blob/master/design/design-doc-observe-only-resources.md#management-policy) to control the late-init behavior since disabling late-init would mean XP managing the resource but only the provided fields. We can find a suitable policy name considering this, just to throw out some ideas: `PartialControl`, `SelectiveControl`, `FieldControl`, `ManagedFields` or `CustomControl`.

negz:
Throwing out another potential policy name: `ExplicitControl` (i.e. only control the fields I explicitly tell you to).

lsviben:
> > Basically your idea would be to turn off late-initialization and instead rely on status.atProvider which is a new feature coming with Observe-Only Resources.
> 
> This is along the lines of what I was thinking after our discussion a few weeks back. We can't turn off late-init by default because that would be a breaking API change. We could add a knob to disable it though - for example the new `managementPolicy` field being added in support of OO resources.
> 
> At that point we'd have:
> 
> * The full observed state in `status.atProvider`.
> * (Optionally) The initial 'bootstrap' desired state in `spec.initProvider`.
> * (Optionally) The constantly-reconciled desired state in `spec.forProvider`.
> 
> If you wanted to allow the external system (e.g. a GKE node pool autoscaler) to control a particular field you would:
> 
> * (Optionally) add an initial desired value to `spec.initProvider`.
> * Don't specify a desired value under `spec.forProvider`.
> * Set `spec.managementPolicy: DoNotLateInit` (hopefully with a better policy name)
> 
> One downside of this approach is that you have to opt-out of late init entirely (not just for the fields you care about). I'm not sure how much of an issue that would actually be in practice.

I like this approach of using the `managementPolicy` and just adding a new policy. 

And in the start we don't need to support `spec.initProvider` as non of the use cases we got up till now really needs something like this. 

Just not sure if this should be optional:
> * (Optionally) The constantly-reconciled desired state in `spec.forProvider`.

IMO the fields that are in `spec.forProvider` still need to be constantly-reconciled by Crossplane. 

For the naming (hardest part as always) I narrowed down the suggestions to:
`PartialControl` - shows that Crossplane is only partially controlling the resource, meaning something else could be controlling other parts/fields.
`SelectiveControl` - Crossplane just controls selected fields of `spec.forProvider`
`ExplicitControl` - similar to above, Crossplane just controls the fields that are explicitly set in `spec.forProvider`.

I'm personally leaning towards `PartialControl` as it feels most unambiguous, and fits with the existing `FullControl`.

lsviben:
> for our use-case, we're creating an azuread group (https://marketplace.upbound.io/providers/upbound/provider-azuread/v0.5.0/resources/groups.azuread.upbound.io/Group/v1beta1), and we want to set `owners` and `members` only on group creation. the owners of the group then manage owners+members outside of crossplane, and also crossplane won't fight with other automation if owners/members leave the company. to achieve this, we had to switch to the terraform provider and use `lifecycle { ignore_changes = [ owners, members ] }`. it's sounding like there was no use-case for a `spec.initProvider` (if not just doing an ignoreChanges field), so wanted to make sure our use-case was known

Thanks for the input, I included it in the use cases :)

lsviben:
Updated the proposal with the solution we discussed above (https://github.com/crossplane/crossplane/pull/3822#issuecomment-1478920120), as it seems like we are converging to a common view.

Now the suggested solution is to add a new Management Policy(name still up for discussion) that would act in all ways as `FullControl` except for:
- skipping late initialization
- orphans the resource on delete

In addition, as there are use cases which need a way to initialize resources with some parameters, but not use them later, the `initProvider` field is added to handle these cases.  

Although, while writing this solution update, it seems to me like these can be divided into 2 features.
One is the new management policy which fixes the use cases where late initialization is conflicting with external management of certain fields.
Second is the `initProviders` which allow users to set some fields just during creation time (although probably this one needs to be used with the new management policy to work well).

Not sure if we should split these features? 

P.S Its also worth to note that the previous suggested solution with `ignoredChanges` solves both issues.

ferpizza:
> Now the suggested solution is to add a new Management Policy(name still up for discussion) that would act in all ways as `FullControl` except for:
>     * skipping late initialization
>     * orphans the resource on delete

Wondering why we are forcing the `Orphan` by design.

Following up on the GCP nodepool example having issues with autoUpgrade and autoScaler, I would want to avoid managing certain fields (eg. nodeCount, version) and still eliminate the Nodepool on delete.

While `Orphan` is a sane default value for these cases, the user should be able to change `spec.deletionPolicy` to `Delete` if it fits its needs better (TL,DR; should continue to work as today)

lsviben:
> > Now the suggested solution is to add a new Management Policy(name still up for discussion) that would act in all ways as `FullControl` except for:
> > * skipping late initialization
> > * orphans the resource on delete
> 
> Wondering why we are forcing the `Orphan` by design.
> 
> Following up on the GCP nodepool example having issues with autoUpgrade and autoScaler, I would want to avoid managing certain fields (eg. nodeCount, version) and still eliminate the Nodepool on delete.
> 
> While `Orphan` is a sane default value for these cases, the user should be able to change `spec.deletionPolicy` to `Delete` if it fits its needs better (TL,DR; should continue to work as today)

Agree that by default we should `Orphan` but leave an option to specify the resource to be deleted. Not sure how to do that yet though. It would be possible with the DeletionPolicy, but that is planned to be [deprecated in favor of ManagementPolicy.](https://github.com/crossplane/crossplane/blob/499895a25d1a1a0ba1604944ef98ac7a1a71f197/design/design-doc-observe-only-resources.md?plain=1#L223) 

But anyway, a valid point and I think we should keep this question open.

turkenh:
> Agree that by default we should Orphan but leave an option to specify the resource to be deleted. Not sure how to do that yet though. It would be possible with the DeletionPolicy, but that is planned to be deprecated in favor of ManagementPolicy.
> 
> But anyway, a valid point and I think we should keep this question open.

Yes, this was something that I realized independently and came here to comment on. 

> Now the suggested solution is to add a new Management Policy(name still up for discussion) that would act in all ways as FullControl except for:
> 
>  - skipping late initialization
>  - orphans the resource on delete

Using the new `ManagementPolicy` instead of the existing `DeletionPolicy` starts becoming a stretch if we want to introduce a management policy like `PartialControl`. To cover all cases, we may need to overload the management policy with extended policy names considering the deletion. `FullControl`, `FullControlOrphanOnDelete`, `PartialControl`, `PartialControlOrphanOnDelete`, `ObserveOnly`.

Considering all of these, I am wondering whether we should keep the `DeletionPolicy` to control management and deletion independently 🤔 The main motivation for deprecation was ManagementPolicy covering existing cases, and the new policy `ObserveOnly` should obviously orphan the resource. However, we can still keep the `DeletionPolicy` and ignore it when the `ManagementPolicy` is `ObserveOnly` as we operate read-only.

bobh66:
> Considering all of these, I am wondering whether we should keep the DeletionPolicy to control management and deletion independently 🤔 The main motivation for deprecation was ManagementPolicy covering existing cases, and the new policy ObserveOnly should obviously orphan the resource. However, we can still keep the DeletionPolicy and ignore it when the ManagementPolicy is ObserveOnly as we operate read-only.

I agree with this - keeping DepetionPolicy separate is a cleaner interface that is easier to understand.

ferpizza:
> To cover all cases, we may need to overload the management policy with extended policy names considering the deletion. `FullControl`, `FullControlOrphanOnDelete`, `PartialControl`, `PartialControlOrphanOnDelete`, `ObserveOnly`.

From my perspective `PartialControl` and `ObserveOnly` can become one. ObserveOnly seems an edge case of PartialControl

If we want to observe only, our manifest will include only the resource name (eg. as we do now when importing existing resources from a cloud provider)

If we add any other value to that manifest, then it behaves as partial control, ensuring informed values are always in sync.

It could be confusing to have values in an ObserveOnly manifest that are not kept in sync, so i assume that name only is the way to go for this case.

turkenh:
> From my perspective `PartialControl` and `ObserveOnly` can become one. ObserveOnly seems an edge case of PartialControl
> 
> If we want to observe only, our manifest will include only the resource name (eg. as we do now when importing existing resources from a cloud provider)
> 
> If we add any other value to that manifest, then it behaves as partial control, ensuring informed values are always in sync.

The main problem here is what to do regarding the creation of resources. `ObserveOnly` clearly indicates that XP should operate as read-only including the creation of the resource.

For example, what should the XP controller do after noticing there is no such bucket with the below manifest? Should it create it or error out assuming the user only wants to observe it? 

```yaml
apiVersion: s3.aws.upbound.io/v1beta1
kind: Bucket
metadata:
  annotations:
    crossplane.io/external-name: example-bucket
  name: example-bucket
spec:
  managementPolicy: PartialControl
  forProvider:
    region: us-west-1
```

We already had some similar discussions in the OO design PR, please check: https://github.com/crossplane/crossplane/pull/3531#discussion_r1101237894

turkenh:
> We wont need the `OrphanOnDelete` ManagementPolicy then? We would just have `FullControl` , `PartialControl`, `ObserveOnly`.

Correct.

> Im not sure if we should have the special case that we ignore the DeletionPolicy for ObserveOnly, seems not intuitive.

Yeah, this is something I am not happy with. However, we can document something like "`DeletionPolicy` will be ignored when `ManagementPolicy` is `ObserveOnly` and the external resource will always be Orphaned".

> Id rather have it Orphan by default when the ManagementPolicy is ObserveOnly.

I don't think setting a conditional default is technically possible, at least with the way we are currently using, i.e. API level defaulting.

lsviben:
Added the results of the latest discussion about DeletionPolicy. Now the proposal incorporates that we:
- keep the lately deprecated `DeletionPolicy`, management policy no longer affects deletion*
- drop the `OrphanOnDelete` management policy as its not needed anymore

\* `ObserveOnly` is a special case where we ignore the `DeletionPolicy` as we are in ReadOnly mode.

jamesdobson:
What are the semantics if the user does something like the following:

```yaml
spec:
  initProvider:
    scalingConfig:
      desiredSize: 1
  forProvider:
    scalingConfig:
      desiredSize: 2
```

Will it be considered an error or, more interestingly, are there any known or theoretical cases where it's useful to allow it?

Thanks!

negz:
> IMO the fields that are in `spec.forProvider` still need to be constantly-reconciled by Crossplane.

Agreed. If a field appears under `spec.forProvider` it should always be considered the user's desired state and enforced by Crossplane. By "optional" I meant the user could optionally omit a field from `spec.forProvider`, and instead include it in `spec.initProvider`. This would communicate that they care about the field's initial "bootstrap" value only.

negz:
> What are the semantics if the user does something like the following:

@jamesdobson Good question.

I don't think it should be an error. I could imagine starting out with the field only specified under `spec.initProvider`, then later wanting to add it under `spec.forProvider`. I don't think we should require the user to "clean up" the field from `initProvider` in that case.

Instead, I would suggest that `spec.forProvider` always supercedes/overrides `spec.initProvider`. If you specify both, we just ignore the `initProvider` value.

I can't immediately think of any use case where it would actually be useful to specify both. Is there a use case for saying "I want to create the resource with this value, but then use this other value on every subsequent update?"

design/one-pager-ignore-changes.md
```diff
@@ -0,0 +1,354 @@
...
+  `desiredSize` set in composition.
+- GCP [Node Pool], where when `management.autoUpgrade` is enabled. After
+  creation, the provider late initializes `spec.forProvider.version`,
+  which when the nodes get upgraded starts conflicting with the 
+  external nodes version. This causes an endless loop of downgrades
+  and upgrades.
+- Tags are a use case that affects multiple resources. Many cloud users
+  have external systems that add tags to resources for billing, tracking
+  or similar purposes. This can conflict with the tags set in 
+  `spec.forProvider.tags`.
```
negz, line 36: This use case seems a little different from the others. Tags are a set, so I could imagine folks wanting to configure something like "ensure these tags are always present, but also allow other tags to be added". This is a little different from "set these tags initially, but after that I don't care what the tags are". In the former case we'd always enforce the tags you specified, but not delete tags added by other entities. In the latter case, we'd only set the specified tags at creation time. If some other entity deleted them, we would not correct that.

I don't think this observation invalidates the design being put forward here - at the very least it's probably a "good enough" solution for the tags use case.

design/one-pager-ignore-changes.md
```diff
@@ -0,0 +1,354 @@
...
+  workaround is to use terraform provider with [ignore_changes].
+
+In these cases, it would make sense to make Crossplane ignore select
+parameters during updates. Terraform addresses this use case with its
+[ignore_changes] functionality.
+
+An additional benefit of ignoring some fields would be that it would be
+possible to have different Crossplane Compositions managing a single
+external resource. This could open up some interesting new possibilities
+for Crossplane.
```
negz, line 49: This sounds worrying to me. What possibilities does it open up?

design/one-pager-ignore-changes.md
```diff
@@ -0,0 +1,354 @@
...
+
+Proposed solution is to use the new [managementPolicy] field which came with
+the [ObserveOnly] feature and add a new management policy that will 
+skip late initialization. The loss the information that the
+late initialization was providing would be offset by the `status.atProvider`
+which contains the resource state on the provider side and is being
+added alongside [ObserveOnly].
+
+In addition to the new management policy, we will also add a new field
+`spec.initProviders` which would contain parameters that should only be
```
negz, line 74:
```suggestion
`spec.initProvider` which would contain parameters that should only be
```

Nit. This appears a few places in the document. I think it should always be `spec.initProvider` (singular).

design/one-pager-ignore-changes.md
```diff
@@ -0,0 +1,354 @@
...
+be added as a separate feature.
+
+## Implementation
+
+### Management policy
+
+Proposed name for the new management policy is `PartialControl`, which
+would fit with the current `FullControl` and insinuate that the resource
+is not fully managed by Crossplane. Some other proposals for the naming
+are: `ExplicitControl`, `SelectiveControl`, `IgnoreLateInit`.
```
negz, line 86: Throwing another option into the ring: `SharedControl`.

I like this option because it communicates the _why_ a little better. Typically when Crossplane does not have "full" control of a resource it's because it's sharing control of the resource with some other system or entity.

negz: It's interesting to me that we're looking at several enum values that mostly end in the word _control_. This makes me wonder whether `managementPolicy` should instead be `controlPolicy`. For example:

```yaml
spec:
  controlPolicy: Full
```

Perhaps in this world `managementPolicy: ObserveOnly` would be `controlPolicy: None`. As in Crossplane does not control this resource. This would leave us with:

* `controlPolicy: Full` - The default. Crossplane controls every field, even those that are late-initialized.
* `controlPolicy: Shared` - Crossplane only controls fields that were explicitly specified.
* `controlPolicy: None` - Crossplane does not control this resource. It only observes it.

cc @turkenh

design/one-pager-ignore-changes.md
```diff
@@ -0,0 +1,354 @@
...
+### Deletion policy
+
+`DeletionPolicy` was planned to be deprecated in favour of the new management
+policies according to [the ObserveOnly design doc.][ObserveOnly] However,
+with `PartialControl` it's not clear if it should delete or orphan the
+external resources. So to avoid creating `PartialControl` and
+`PartialControlOrphanOnDelete` management policies, we will keep the
+`DeletionPolicy` to be able to specify the behaviour. In that case, we will
+need to deprecate and remove the `OrphanOnDelete` management policy as it
+won't be needed anymore.
```
negz, line 102: Our alpha API contract allows us to just drop this enum value with no notice. We should definitely make it very obvious in the release notes, but we don't need a deprecation period.

design/one-pager-ignore-changes.md
```diff
@@ -0,0 +1,354 @@
...
+
+So the management policies would be:
+- `FullControl`
+- `PartialControl`
+- `ObserveOnly`
+
+With a note that for `ObserveOnly` the `DeletionPolicy` will be ignored,
+as we are in read-only mode.
```
negz, line 110: This seems implicit and surprising. I'd prefer we honor deletion policy even in observe-only mode. In practice I would expect almost everyone to set `deletionPolicy: Orphan` when they set `managementPolicy: ObserveOnly`.

design/one-pager-ignore-changes.md
```diff
@@ -0,0 +1,354 @@
...
+    name: example
+```
+
+This would allow users to specify fields that would only be used on creation,
+but not on updates.
+
+When it comes to the implementation, from Crossplane side we would need to
+change the API. But the handling would all need to be done on the 
+provider level. The `initProvider` fields should only be used in the `Create`
+step of the managed resource reconciliation loop.
```
negz, line 152: > from Crossplane side we would need to change the API

I don't follow. What changes would we need to make in Crossplane? Wouldn't this change be isolated to providers?

design/one-pager-ignore-changes.md
```diff
@@ -0,0 +1,354 @@
...
+resources. However, we will ignore the `spec.ignoreChanges` when the feature
+is not enabled.
+
+#### Alternative feature gating approach
+
+To avoid changing the API, we could ship the `ignoreChanges` field as an
+optional annotation as a first solution. This would allow us to survey
+users and see how much this feature is useful before adding it to the API.
+
+### initProviders
```
negz, line 284: Should this still appear under Alternatives Considered? It appears to be part of the proposed approach.

negz (approved):
I have a few nitpicks and a suggestion around the naming of `managementPolicy`, but overall this looks great to me. I'm happy to hop on a call this week if there's anything small you'd like my help in ironing out (i.e. field/enum naming suggestions).

design/one-pager-ignore-changes.md
turkenh: >Throwing another option into the ring: `SharedControl`.

I feel like the `SharedControl` causes a slight confusion by triggering the question of "shares control with whom/what?". From Crossplane POV, it `partially` or `selectively` controls the resource, so I am closer to `Partial` or `Selective` control.

> It's interesting to me that we're looking at several enum values that mostly end in the word control. This makes me wonder whether managementPolicy should instead be controlPolicy.

I like the simplification in general, but not sure because of the following:

- `managementPolicy: ObserveOnly` is more explicit than `controlPolicy: None`.
- `managementPolicy` sounds easier to extend if needed compared to `controlPolicy`.

design/one-pager-ignore-changes.md
lsviben: right, its not viable as a standalone solution for all the use cases so ill remove it.

design/one-pager-ignore-changes.md
lsviben: Good catch, actually it should not touch Crossplane API at all as the initProviders are only handled in providers, removing that line.

design/one-pager-ignore-changes.md
```diff
@@ -0,0 +1,354 @@
+# Ignore changes
+
+* Owners: Lovro Sviben (@lsviben)
+* Reviewers: Crossplane Maintainers
```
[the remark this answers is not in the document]
lsviben: Will note that for the future, I set you and @turkenh now as you are both reviewing this already.

design/one-pager-ignore-changes.md
turkenh: > `managementPolicy` sounds easier to extend if needed compared to `controlPolicy`.

The `managementPolicy` field was heavily inspired by the previous work for [enhanced provider-kubernetes](https://github.com/crossplane-contrib/provider-kubernetes/blob/main/docs/enhanced-provider-k8s.md), where management policy was controlling which of the CRUD operations should be enabled for a given resource. The policy there [supports](https://github.com/crossplane-contrib/provider-kubernetes/blob/main/docs/enhanced-provider-k8s.md#resource-management-policy) `ObserveDelete` in addition to what we have here and there is a [request](https://github.com/crossplane-contrib/provider-kubernetes/issues/115) for `ObserveUpdate`. 

Noticed that we are moving away from such a semantic with `PartialControl` and by keeping `deletionPolicy` together with this. This should be ok since I am not aware of any current need for behaviors like `ObserveDelete` or `ObserveDelete` for cloud resources in general, but still, we may want to keep this in mind as possible extensions.

design/one-pager-ignore-changes.md
lsviben: Removed the part about ignoring the `deletionPolicy` My worry was that the default `deletionPolicy` is `Delete` so users who do not set the `deletionPolicy: Orphan` could be surprised. But probably less intuitive it would be if it was ignored altogether. 

Not sure how to improve this experience except noting it very clearly in the documentation.

design/one-pager-ignore-changes.md
sttts: my two cents, from a similar topic in a different context: thinking in policies is very procedural. You have to think about what the providers do, how they work. In that other context, we came up with ownership of fields as an alternative view, plus exceptions (a list of json paths), i.e.:
```
ownership: External
exceptFields:
- foo.bar
```
or:
```
ownership: Crossplane
externalFields:
- a.b.c
```
Looking further down, `ignoreChanges:` is pretty much that. `externallyOwned:` would reflect the ownership idea more, but that's cosmetics.

design/one-pager-ignore-changes.md
```diff
@@ -0,0 +1,352 @@
...
+Setting initial tags should work with `initProvider`. However, it seems that
+the providers are setting some default tags in the into the `forProvider`
+field in the `Initialize` step and are updating the resource, so this would
+need to be changed to use `initProvider` instead, or we can check if those
+tags are actually needed or just skip setting them similarly to how
+ObserveOnly does it.
+
+Ref: [Upjet Initialize] or [AWS community provider tag example].
+
+## Solution alternative considered
```
turkenh, line 206: Could you also add some notes on why we didn't go with the alternative solutions?

design/one-pager-ignore-changes.md
```diff
@@ -0,0 +1,352 @@
...
+
+The new management policy would use the same feature gate as ObserveOnly. So
+it would be an alpha feature that will be off by default and will be
+controlled by `--enable-alpha-management-policies` flag in Providers.
+
+The `initProvider` field would require a separate feature gate. So it would
+also be disabled by default and be controlled by `--enable-alpha-init-provider`
+flag in the Provider.
+
+### Examples how the new functionality would affect the use cases
```
turkenh, line 160: nit: it would be helpful to see some (partial) yaml's here.

turkenh:
Triggered by the feedback by @bobh66 on [Crossplane slack](https://crossplane.slack.com/archives/CEG3T90A1/p1683813876442189?thread_ts=1683810301.274759&cid=CEG3T90A1), I am putting some thought into converging the policy here with the `managementPolicy` field in the `provider-kubernetes` .

After some unsuccessful experiments trying to extend [xpv1.ManagementPolicy](https://github.com/crossplane/crossplane-runtime/blob/1316ae6695eec09cf47abdfd0bc6273aeaab1895/apis/common/v1/policies.go#L22) with the [management policies in provider kubernetes](https://github.com/crossplane-contrib/provider-kubernetes/blob/2c5108bc296b239802c3db5bf32a9ea5913aebb7/apis/object/v1alpha1/types.go#L38-L46) with the intent of making some conversions between (e.g. if `Observe` return `ObserveOnly`), I am closer to using a conversion webhook that would convert `v1alpha1` to `v1alpha2` of provider kubernetes `Object` to deprecate policies there in favor of the ones in the Runtime. 

I imagined to build a conversion webhook as follows (`v1alpha1` => `v1alpha2`):
- `managementPolicy: Default` => `managementPolicy: FullControl && deletionPolicy: Delete`
- `managementPolicy: Observe` => `managementPolicy: ObserveOnly && deletionPolicy: Orphan`
- `managementPolicy: ObserveCreateUpdate` => `managementPolicy: FullControl && deletionPolicy: Orphan`
- `managementPolicy: ObserveDelete` => `managementPolicy: ObserveOnly && deletionPolicy: Delete`

All existing policies perfectly maps a combination of `managementPolicy` and `deletionPolicy` but with a caveat which is, how the `managementPolicy: ObserveOnly` would work together with `deletionPolicy: Delete` ?

As mentioned [above](https://github.com/crossplane/crossplane/pull/3822#issuecomment-1505054199), I believe it is not safe to delete the object when `managementPolicy: ObserveOnly && deletionPolicy: Delete` given we have the `deletionPolicy` defaulted as `delete`. So, I suggested making an exception here and ignoring `deletionPolicy` and never delete the resource when `managementPolicy: ObserveOnly`. 

After the above experiments with provider-kubernetes, I am now thinking whether we should not make such an exception but rather change how we default the deletionPolicy value. For example, it does not default at API level, and we default to `Orphan` if managementPolicy is `ObserveOnly` but `Delete` otherwise.

I am still not happy since `ObserveOnly` kind of implies that we will only observe (and never delete the resource), but this is a natural outcome of the decision we are making on keeping `deletionPolicy` together with `managementPolicy` in the scope of this design.

turkenh:
> I am now thinking whether we should not make such an exception but rather change how we default the deletionPolicy value. For example, it does not default at API level, and we default to Orphan if managementPolicy is ObserveOnly but Delete otherwise.
> 
> I am still not happy since ObserveOnly kind of implies that we will only observe (and never delete the resource), but this is a natural outcome of the decision we are making on keeping deletionPolicy together with managementPolicy in the scope of this design.

One caveat with above, even if we set the default of the `deletionPolicy`, a user might change the `managementPolicy` from `FullControl` to `ObserveOnly` on an existing resource that was initialized as `deletionPolicy: Delete` in the beginning. I cannot convince myself to delete an external resource with `managementPolicy: ObserveOnly` in any case.

turkenh:
This makes me wonder if we should stick with the decision of deprecating `deletionPolicy` in favor of `managementPolicy` but find another way to control the "late-initialization" behavior for this design. 

We had an offline discussion with @lsviben on the topic and wanted to record some alternative options as the outcome of that. 

Introduce another field for that:

```yaml
spec:
  managementPolicy: Default # or ObserveOnly, OrphanOnDelete, ObserveDelete
  fieldManagementPolicy: AllFields # or SpecifiedFieldsOnly
```

Or use a nested structure for `managementPolicy`:

```yaml
spec:
  managementPolicy:
    allowedActions: All  # or ObserveOnly, OrphanOnDelete, ObserveDelete
    fieldScope: AllFields  # or SpecifiedFieldsOnly
```

Deprecating `deletionPolicy` for both cases.

design/one-pager-ignore-changes.md
lsviben: > my two cents, from a similar topic in a different context: thinking in policies is very procedural. You have to think about what the providers do, how they work. In that other context, we came up with ownership of fields as an alternative view, plus exceptions (a list of json paths), i.e.:
> 
> ```
> ownership: External
> exceptFields:
> - foo.bar
> ```
> 
> or:
> 
> ```
> ownership: Crossplane
> externalFields:
> - a.b.c
> ```
> 
> Looking further down, `ignoreChanges:` is pretty much that. `externallyOwned:` would reflect the ownership idea more, but that's cosmetics.

Thanks for the feedback @sttts. Actually initially my proposed solution was to go with the `ignoreChanges:`, which is a valid idea and would solve the issue. But in the end, we opted for `initProviders` as it seems easier to grasp the concept to have an `initProviders` field that supplements the `forProviders` one at the creation/initialization time and later on the `forProviders` is the single source of truth. While with `ignoreChanges`, the field values that are in `forProviders` can become outdated if they are ignored, which can lead to confusion.

negz:
> I am now thinking whether we should not make such an exception but rather change how we default the deletionPolicy value. For example, it does not default at API level, and we default to Orphan if managementPolicy is ObserveOnly but Delete otherwise.

This seems like a good compromise to me.

I worry that if we overthink this too much we'll end up with something that is potentially safer, but harder to understand. Perhaps it would be sufficient to add a strong warning to `managementPolicy` that it does not control what happens when the MR is deleted, and that you must set `deletionPolicy` accordingly?

negz:
> The Initialize can be confused with the [MR Initialize](https://github.com/crossplane/crossplane-runtime/blob/master/pkg/reconciler/managed/reconciler.go#L742) which from what I see mostly just sets some default tags, while the late initialization update happens [elsewhere](https://github.com/crossplane/crossplane-runtime/blob/master/pkg/reconciler/managed/reconciler.go#L1033).

We could avoid this by calling it `LateInitialize`:

```yaml
spec:
  managementPolicy: [Observe, Create, LateInitialize, Update, Delete]
```

> So IMO we would still need a separate field that controls the actions on the managed resource,

I don't think I agree. I would argue that `Observe` is an action on the managed resource too - I think of `Observe` as being short-hand for "observe the external resource, and update the managed resource's `status.atProvider` accordingly. So `Observe` already affects how the MR is updated. I think it would be fine for `LateInitialize` to do the same, and it keeps everything in "one place" (fewer fields to control how the MR is reconciled).

Worth noting that there are even more slightly oddball combinations this would open up, like:


```spec:
  managementPolicy: [Create, Update, Delete]
```

I would interpret this as:

* Create the resource using the desired state specified in `spec.forProvider`
* Update the resource using the desired state specified in `spec.forProvider`
* Do not update the observed state in `status.atProvider`.

In practice we'd still be observing the ER in order to know whether we should create or update it, we just wouldn't be recording our observations. I also don't really know why anyone would use this particular policy, but it seems like we if we took this "array of enums" path we should support it for consistency.

bobh66:
I think using `managementPolicy` to directly control reconciliation and the external interfaces makes a lot of sense - it's better to be explicit about what each enum value is doing. `PartialControl` is vague and non-specific - using `Create`, `Update` etc is much easier to understand.

negz:
```yaml
spec:
  managementPolicy: [Observe, Create, LateInitialize, Update, Delete]
```

Let me try write documentation for this, as I imagine it would appear in the API docs: 

> `managementPolicy` controls how Crossplane reconciles this managed resource with its equivalent external resource. It is an array of operations Crossplane should perform. The following operations are supported:
>
> * `Create` - Use `spec.initProvider` and `spec.forProvider` to create the external resource.
> * `Update` - Use `spec.forProvider` to update the external resource.
> * `Delete` - Delete the external resource when the managed resource is deleted.
> * `Observe` - Update `status.atProvider` to reflect the state of the external resource.
> * `LateInitialize` - Update unspecified `spec.forProvider` fields to reflect the state of the external resource.
>
> Crossplane will only perform the operations that a resource's `managementPolicy` allows. The default `managementPolicy` is `[*]`. This is semantically equivalent to `[Create, Update, Delete, Observe, LateInitialize]`.

lsviben:
> ```yaml
> spec:
>   managementPolicy: [Observe, Create, LateInitialize, Update, Delete]
> ```
> 
> Let me try write documentation for this, as I imagine it would appear in the API docs:
> 
> > `managementPolicy` controls how Crossplane reconciles this managed resource with its equivalent external resource. It is an array of operations Crossplane should perform. The following operations are supported:
> > 
> > * `Create` - Use `spec.initProvider` and `spec.forProvider` to create the external resource.
> > * `Update` - Use `spec.forProvider` to update the external resource.
> > * `Delete` - Delete the external resource when the managed resource is deleted.
> > * `Observe` - Update `status.atProvider` to reflect the state of the external resource.
> > * `LateInitialize` - Update unspecified `spec.forProvider` fields to reflect the state of the external resource.
> > 
> > Crossplane will only perform the operations that a resource's `managementPolicy` allows. The default `managementPolicy` is `[*]`. This is semantically equivalent to `[Create, Update, Delete, Observe, LateInitialize]`.

For me this seems understandable and flexible. If there are no other concerns, I will update the proposal to reflect the proposed changes to the management policy.

turkenh:
> ```yaml
> spec:
>   managementPolicy: [Observe, Create, LateInitialize, Update, Delete]
> ```

I like the flexibility and explicitness of this API. My only concern is its verbosity, especially for some common use cases. For example:

```
spec:
  deletionPolicy: Orphan
```  

becomes

```
spec:
  managementPolicy: [Observe, Create, LateInitialize, Update]
```

But I don't think it is a significant problem and fine with moving forward with it.

design/one-pager-ignore-changes.md
```diff
@@ -0,0 +1,658 @@
...
+#### Tags
+
+Setting initial tags should work with `initProvider`. However, it seems that
+the providers are setting some default tags in the into the `forProvider`
+field in the [Initialize] step and are updating the resource, so this would
+need to be changed to use `initProvider` instead, or we can check if those
+tags are actually needed or just skip setting them similarly to how
+ObserveOnly does it.
+
+Ref: [Upjet Initialize] or [AWS community provider tag example].
```
lsviben, line 323: I am wondering if for this use case we should:
- add a new management policy enum "Initialize" which controls if this step of the reconciler loop should be executed.
- handle this in the providers, meaning remove forcing the default tags in `spec.forProvider` . I am not aware if these tags are required for something or are they informative. In any case this behavior is currently the issue with this use case. We could remove initializing them all togather, or initialize them in `spec.initProviders`

I am closer to solution no1, adding another enum to control the initialization. But that is in part because I am not sure how solution no2 would affect the providers.

lsviben:
> > ```yaml
> > spec:
> >   managementPolicy: [Observe, Create, LateInitialize, Update, Delete]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Let me try write documentation for this, as I imagine it would appear in the API docs:
> > > `managementPolicy` controls how Crossplane reconciles this managed resource with its equivalent external resource. It is an array of operations Crossplane should perform. The following operations are supported:
> > > 
> > > * `Create` - Use `spec.initProvider` and `spec.forProvider` to create the external resource.
> > > * `Update` - Use `spec.forProvider` to update the external resource.
> > > * `Delete` - Delete the external resource when the managed resource is deleted.
> > > * `Observe` - Update `status.atProvider` to reflect the state of the external resource.
> > > * `LateInitialize` - Update unspecified `spec.forProvider` fields to reflect the state of the external resource.
> > > 
> > > Crossplane will only perform the operations that a resource's `managementPolicy` allows. The default `managementPolicy` is `[*]`. This is semantically equivalent to `[Create, Update, Delete, Observe, LateInitialize]`.
> 
> For me this seems understandable and flexible. If there are no other concerns, I will update the proposal to reflect the proposed changes to the management policy.

Updated the proposal, please take another look.

lsviben:
I am not sure if this document should still be a proposal for `ignore_changes`. I feel that it went beyond that and now the `ignore_changes` are just one of the symptoms we are solving through a broader solution.

Maybe more fitting would be something like : "Crossplane management policy" or "Crossplane managed resource management policy" - but that has too many manage in it.

WDYT? I would change the background and goals a bit to fit more with the proposed solution.

bobh66:
> Actually, at least in some cases, that might well be desired. Perhaps some out-of-band change was made in a break-glass kind of scenario, and we don't want that to automatically be reverted.

This is one of the scenarios behind the "pause reconciliation" feature that uses annotation `crossplane.io/paused: "true"`

I agree with @negz that `managementPolicy: []` appears to be equivalent to  `crossplane.io/paused: "true"` and is (to me) easier to manage than the annotation.

design/one-pager-ignore-changes.md
```diff
@@ -0,0 +1,658 @@
...
+This would allow users to specify fields that would only be used on creation,
+but not on updates.
+
+When it comes to the implementation, the handling would all need to be done
+on the provider level. The `initProvider` fields should only be used in the
+`Create` step of the managed resource reconciliation loop.
+
+Support for `initProvider` would need to be a manual change in each provider.
+From our side we could provide a helper function to merge the `initProvider`
+and `forProvider` fields which would be used in the `Create` step.
```
turkenh, line 217: Considering we are only aware a limited set of use cases/resources, we could also add support for `spec.initProvider` only for those resources. So, a managed resource may or may not have `spec.initProvider` depending on its nature. 

Another questions is if `initProvider` should have all the fields from `forProvider` or would it be ok just to put the fields that are needed, e.g. `scalingConfig.desiredSize` ?

design/one-pager-ignore-changes.md
```diff
@@ -0,0 +1,658 @@
...
+The `scalingConfig.desiredSize` is required so `initProvider` would need
+to be used alongside omitting `LateInitialize` management policy. This way
+the autoscaler would be able to control the `desiredSize` after creation.
+
+```yaml
+spec:
+  managementPolicy: ["Create", "Update", "Delete", "Observe"]
+  initProvider:
+    scalingConfig:
+      desiredSize: 1
```
turkenh, line 278: It looks like we are disabling lateInit whenever we set `initProvider`. So, I am wondering if this is something that could be generalized as, if a field is provided under initProvider, never late initialize that specific parameter even managementPolicy contains lateInit. I think the underlying idea is providing a field under `initProvider` implies that it should not appear under `forProvider`.

design/one-pager-ignore-changes.md
turkenh: > add a new management policy enum "Initialize" which controls if this step of the reconciler loop should be executed.

Initialize contains other initializers like [NameAsExternalName](https://github.com/crossplane/crossplane-runtime/blob/1316ae6695eec09cf47abdfd0bc6273aeaab1895/pkg/reconciler/managed/reconciler.go#L479), I don't think exposing control of this to users is a good idea. 

> We could remove initializing them all together

They were added on purpose so, not sure if we should remove them unless we see them causing a significant issue that we cannot avoid.

> initialize them in `spec.initProviders`

This could be an option but as I commented above, I am wondering if we could support `spec.initProvider` on a case-by-case basis instead of supporting it for all managed resources.

Another idea could be to make an exception for tags for now. Are there any signals that they would cause an issue? May be I am missing the problem here.

design/one-pager-ignore-changes.md
lsviben: > Considering we are only aware a limited set of use cases/resources, we could also add support for `spec.initProvider` only for those resources. So, a managed resource may or may not have `spec.initProvider` depending on its nature.
> 

Added a paragraph below to better explain it. 

> We know that for now the `initProvider` field would be useful just for a 
handful of cases that we are aware of. So provider developers do not need to
support `initProvider` for all resources right away. Rather, they can add
support for those resources that are known to require it, and later on based
on a request basis.

So basically its up to the provider developers to support `spec.initProviders` for a particular MR, based on the needs. The important thing is that they have such an option now. 


> Another questions is if `initProvider` should have all the fields from `forProvider` or would it be ok just to put the fields that are needed, e.g. `scalingConfig.desiredSize` ?

I think it also depends on the provider developer. I suggested doing the same as `spec.forProvider` as it seems easier to maintain.

design/one-pager-ignore-changes.md
lsviben: We could add such a functionality that the late initialization code checks for `initProvider` fields when updating the `forProvider` fields. It already does it for `forProvider` fields, so it would not be a stretch.

design/one-pager-ignore-changes.md
lsviben: > Another idea could be to make an exception for tags for now. Are there any signals that they would cause an issue? May be I am missing the problem here.

The issue with the tags is that as they are always present in the `spec.forProvider`, users are not able to manage the tags externally, and the proposed solution here does not help with that because they are added in the Initialize step. 

Maybe the tags problem should be handled in another issue.

design/one-pager-ignore-changes.md
```diff
@@ -0,0 +1,658 @@
...
+So this is something to keep in mind when developing new management policies.
+For this proposal, we will go with the `["*"]` approach as its more
+future-proof.
+
+### Migrating existing resources
+
+The `managementPolicy` feature is alpha, so it should be ok to change the
+API. We should still offer a migration path for existing resources. The
+combinations of `managementPolicy` and `deletionPolicy` should be migrated
+to the new `managementPolicy` field.
```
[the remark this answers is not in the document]
lsviben: Renamed the new field to `managementPolicies`. 

We cant just change the field type as it would remove the existing data in the field, and although the ObserveOnly feature is alpha, its probably used by a bunch of people due to it being a very sought after feature.

So the plan would be to keep the deprecated `managementPolicy` field for a release, and provide a migration tool if we think its necessary.

jbw976:
> I am not sure if this document should still be a proposal for ignore_changes. I feel that it went beyond that and now the ignore_changes are just one of the symptoms we are solving through a broader solution.

Agreed on this @lsviben - do you want to go ahead and update the PR title to capture this broader scope too?

lsviben:
> Member

Done! Kept the ignore changes part for historic and intention reasons and added granular management policies

PR #3846, opened 2023-03-09
Update status in composition function

### LINKED ISSUE #3812: Can't use Composition Function to update XR status
### What happened?
Trying to update a status field in my XR, but it seems to alway be overwritten with conditions. I was trying to replicate a `ToCompositeFieldPath` patch, but it never seems to be written. Updating other fields in the desired state such as labels works fine, but status fields look to be uneditable.

### How can we reproduce it?
I have the following in my definition.yaml
```yaml
piVersion: apiextensions.crossplane.io/v1
metadata:
  name: xbuckets.catalog.projectx.mavenwave.dev
spec:
  group: catalog.projectx.mavenwave.dev
  names:
    kind: XBucket
    plural: xbuckets
  claimNames:
    kind: BucketClaim
    plural: bucketclaims
  defaultCompositionRef:
    name: gcp-bucket
  versions:
  - name: v1alpha1
    served: true
    referenceable: true
    schema:
      openAPIV3Schema:
        type: object
        properties:
...
          status:
            type: object
            properties:
              bucket:
                type: object
                properties:
                  id:
                    type: string
```

Here's a snippet of my Python code that updates the status field.
```python
try:
  Functionio = read_Functionio()
except yaml.parser.ParserError as err:
  sys.stdout.write("cannot parse FunctionIO: {}\n".format(err))
  sys.exit(1)
...
status = deepcopy(Functionio["observed"]["composite"]["resource"]["status"])
Functionio["desired"]["composite"]["resource"]["status"] = status
Functionio["desired"]["composite"]["resource"]["status"]["bucket"] = {"id": "bar"}
Functionio["desired"]["composite"]["resource"]["metadata"]["labels"] = {"id": "bar"}

write_Functionio(Functionio)
```

The `Functionio["desired"]["composite"]["resource"]["metadata"]["labels"] = {"id": "bar"}` works as expected while the `Functionio["desired"]["composite"]["resource"]["status"]["bucket"] = {"id": "bar"}` does not.

Crossplane version: 
v0.11.0, v0.11.1, 1.12.0-rc.0

* Cloud provider or hardware configuration
xpkg.upbound.io/upbound/provider-gcp:v0.28.0
* Kubernetes version (use `kubectl version`)
Client Version: version.Info{Major:"1", Minor:"25", GitVersion:"v1.25.6", GitCommit:"ff2c119726cc1f8926fb0585c74b25921e866a28", GitTreeState:"clean", BuildDate:"2023-01-18T19:22:09Z", GoVersion:"go1.19.5", Compiler:"gc", Platform:"darwin/amd64"}
Kustomize Version: v4.5.7
Server Version: version.Info{Major:"1", Minor:"25", GitVersion:"v1.25.3", GitCommit:"434bfd82814af038ad94d62ebe59b133fcb50506", GitTreeState:"clean", BuildDate:"2022-10-25T19:35:11Z", GoVersion:"go1.19.2", Compiler:"gc", Platform:"linux/amd64"}
* Kubernetes distribution (e.g. Tectonic, GKE, OpenShift)
Kind
* OS (e.g. from /etc/os-release)
MacOS Ventura 13.2
* Kernel (e.g. `uname -a`)
Darwin Kernel Version 22.3.0

bradkwadsworth-mw:
Fixes issue where composition functions do not update the defined status fields in the XR.
Fixes #3812 

I have:

Created a function container that just updates values in the definition status fields.

### DIFF
--- a/internal/controller/apiextensions/composite/composition_ptf.go
+++ b/internal/controller/apiextensions/composite/composition_ptf.go
@@ -337,6 +337,9 @@ func (c *PTFComposer) Compose(ctx context.Context, xr resource.Composite, req Co
 	// Subsequent attempts to update that object will therefore fail. This
 	// should be okay; the caller should keep trying until this is a no-op.
 	ao := mergeOptions(filterPatches(allPatches(state.ComposedResources), patchTypesToXR()...))
+	if err := c.client.Status().Update(ctx, state.Composite); err != nil {
+		return CompositionResult{}, errors.Wrap(err, errApplyXR)
+	}
 	if err := c.client.Apply(ctx, state.Composite, ao...); err != nil {
 		return CompositionResult{}, errors.Wrap(err, errApplyXR)
 	}
--- a/internal/controller/apiextensions/composite/composition_ptf_test.go
+++ b/internal/controller/apiextensions/composite/composition_ptf_test.go
@@ -245,7 +245,8 @@ func TestPTFCompose(t *testing.T) {
 			params: params{
 				kube: &test.MockClient{
 					// We test through the ClientApplicator - Get is called by Apply.
-					MockGet: test.NewMockGetFn(errBoom),
+					MockGet:          test.NewMockGetFn(errBoom),
+					MockStatusUpdate: test.NewMockSubResourceUpdateFn(nil),
 				},
 				o: []PTFComposerOption{
 					WithCompositeConnectionDetailsFetcher(ConnectionDetailsFetcherFn(func(ctx context.Context, o resource.ConnectionSecretOwner) (managed.ConnectionDetails, error) {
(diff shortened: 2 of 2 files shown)


internal/controller/apiextensions/composite/composition_ptf.go
```diff
@@ -337,6 +337,9 @@ func (c *PTFComposer) Compose(ctx context.Context, xr resource.Composite, req Co
 	// Subsequent attempts to update that object will therefore fail. This
 	// should be okay; the caller should keep trying until this is a no-op.
 	ao := mergeOptions(filterPatches(allPatches(state.ComposedResources), patchTypesToXR()...))
+	if err := c.client.Status().Update(ctx, state.Composite); err != nil {
+		return CompositionResult{}, errors.Wrap(err, errApplyXR)
+	}
```
negz, line 342: Are you sure this works? More specifically, are you sure it doesn't "undo" pending changes to other parts of the XR?

My recollection is that when you do a status update like this, the `spec` and `metadata` of `state.Composite` will be reset to whatever the current state is in the API server. i.e. Any pending state changes will be discarded before we reach `Apply`.

bradkwadsworth-mw: Looking into this a little further, is it actually possible to take values from a resource's `status.atProvider` in a desired resource and apply it to the XR?

negz: Do you need it from a desired resource specifically, or would an observed resource do?

Either way, it's definitely intended that you can update the XR from the status of a composed resource.

bradkwadsworth-mw: Here's an example use case. I would like to create a GCP VPC and x number of subnets. Each  subnet would be represented as an array of objects in my definition. My composition function would loop over this array and create a subnetwork for each item in the array and add the desired resources. How would I access the `status.atProvider` fields of those subnets in order to add some of them to the status fields of my definition?

Issue #3717, opened 2023-02-03
Implement `imagePullSecrets` for Composition Functions


negz:
### What problem are you facing?

The [Composition Functions design](https://github.com/crossplane/crossplane/blob/40649e2ce9cac78fc4e16f3366b78ccd4dfb46e8/design/design-doc-composition-functions.md) called for an array of `imagePullSecrets` that would be used to provide credentials to authenticate to a private OCI repository. This functionality was not implemented in time for the v1.11 release and had to be cut.

```yaml
apiVersion: apiextensions.crossplane.io/v2alpha1
kind: Composition
metadata:
  name: example
spec:
  compositeTypeRef:
    apiVersion: database.example.org/v1alpha1
    kind: XPostgreSQLInstance
  functions:
  - name: my-cool-function
    type: Container
    container:
      image: private.example.org/my-cool-function:0.1.0
      imagePullSecrets:
      - namespace: crossplane-system
        name: my-xpkg-io-creds
```

These secrets would be resolved to an [`ImagePullAuth`](https://github.com/crossplane/crossplane/blob/40649e2ce9cac78fc4e16f3366b78ccd4dfb46e8/apis/apiextensions/fn/proto/v1alpha1/run_function.proto#L42) message in the `RunFunctionRequest`:

```proto
// ImagePullAuth configures authentication to a remote OCI registry.
// It corresponds to go-containerregistry's AuthConfig type.
// https://pkg.go.dev/github.com/google/go-containerregistry@v0.11.0/pkg/authn#AuthConfig
message ImagePullAuth {
  string username = 1;
  string password = 2;
  string auth = 3;
  string identity_token = 4;
  string registry_token = 5;
}
```

### How could Crossplane help solve your problem?

We should implement `imagePullSecrets`. I believe to do so we need to:

* Update the `Composition` and `CompositionRevision` type definitions.
* Use the [`k8schain`](https://pkg.go.dev/github.com/google/go-containerregistry/pkg/authn/k8schain) package to resolve the secrets to an `ImagePullAuth`.

This would be a novel use of `k8schain` but I think it should work. Typically `k8schain` is passed as an argument to the go-containerregistry code that does the image pulling, but in our case that code exists inside `xfn` on the "other side" of the gRPC connection from Crossplane.

srlobo:
I'm trying to implement this, as we need testing composite functions on our setup and we are working with private repositories.

truongnht:
[...]

Issue #3807, opened 2023-02-26
cannot run composition function with crossplane/xfn:v1.11.1


duizabojul:
### What happened?

After upgrading to v1.11.1 (using latest helm chart) to avoid [flapping issues](https://github.com/crossplane/crossplane/issues/3747), I got errors when i am running my functions: 

```
cannot compose resources: cannot run Composition Function pipeline:
cannot run function "xnf-api": cannot run container: rpc error: code =
Unknown desc = exit status 1: xfn: error: spark.Command.Run(): OCI
runtime error: exit status 1
```

I thought it was a bug in my container script but even if I forward stdin to stdout, I got the above error.

Reverting to `crossplane/xfn:v1.11.0` fix this error.

### How can we reproduce it?

Use composition functions with xfn container running `crossplane/xfn:v1.11.1`. 

### What environment did it happen in?
Crossplane version: 
`crossplane/xfn:v1.11.1`

Kubernetes:
EKS 1.24

duizabojul:
I digged a bit in this issue and i found the working image is running crun 1.5 and the failing one crun 1.8

negz:
@duizabojul Thanks for the report! Are you able to share your Composition Function? I seem to be able to run functions in general using v1.11.1:

```yaml
$ cat ~/xp/fnio.yaml |docker run -i --security-opt=seccomp=unconfined crossplane/xfn:v1.11.1 run -c /tmp negz/xfn-quotable:latest -
apiVersion: apiextensions.crossplane.io/v1alpha1
config:
  apiVersion: database.example.org/v1alpha1
  kind: Config
  metadata:
    name: cloudsql
  spec:
    version: POSTGRES_9_6
kind: FunctionIO
observed:
  composite:
    connectionDetails:
    - name: uri
      value: postgresql://db.example.org:5432
    resource:
      apiVersion: database.example.org/v1alpha1
      kind: XPostgreSQLInstance
      metadata:
        name: my-db
      spec:
        compositionSelector:
          matchLabels:
            provider: gcp
        parameters:
          storageGB: 20
      status:
        conditions:
        - status: true
          type: Ready
```

negz:
As a side-note, it would be ideal to pin the `crun` version we're using.

duizabojul:
[...]

Issue #3718, opened 2023-02-03
Support passing credentials to Composition Functions


negz:
### What problem are you facing?

When reviewing the [Composition Functions design](https://github.com/crossplane/crossplane/blob/40649e2ce9cac78fc4e16f3366b78ccd4dfb46e8/design/design-doc-composition-functions.md) @turkenh pointed out that we don't currently have a secure way for a Composition Function to authenticate to an external system. We don't expect Composition Functions to _orchestrate_ external systems directly - we expect them to produce composed resources for Crossplane to orchestrate. We do however expect that Composition Functions might need to _query_ external systems in order to determine what Crossplane should do.

It's currently possible to specify an arbitrary `config` object for a function. This object will be copied verbatim to the config object of `FunctionIO`. This is not appropriate for passing credentials because the `Composition` is available as plaintext to any one with RBAC access to view them.

```yaml
apiVersion: apiextensions.crossplane.io/v1
kind: Composition
metadata:
  name: example
spec:
  compositeTypeRef:
    apiVersion: database.example.org/v1alpha1
    kind: XPostgreSQLInstance
  functions:
  - name: my-cool-Function
    type: Container
    container:
      image: xkpg.io/my-cool-Function:0.1.0
    config:
      apiVersion: database.example.org/v1alpha1
      kind: Config
      metadata:
        name: cloudsql
      spec:
        version: POSTGRES_9_6
```

I can't think of any other good alternatives, either.

* It's not a good idea to bake your credentials into your function's OCI image, especially because we don't yet support private OCI images per https://github.com/crossplane/crossplane/issues/3717.
* Composition Functions aren't run as Pods, so there's no way to inject credentials e.g. via a volume or environment variable.

### How could Crossplane help solve your problem?

We should make it possible to securely specify credentials. One option might be to support loading the entire `config` object from a (more) secure source such as a Kubernetes secret or a Secret Store (https://github.com/crossplane/crossplane/issues/3670). Something like:

```yaml
apiVersion: apiextensions.crossplane.io/v1
kind: Composition
metadata:
  name: example
spec:
  compositeTypeRef:
    apiVersion: database.example.org/v1alpha1
    kind: XPostgreSQLInstance
  functions:
  - name: my-cool-Function
    type: Container
    container:
      image: xkpg.io/my-cool-Function:0.1.0
    config:
      source: Secret
      secretRef:
        namespace: crossplane-system
        name: my-cool-config
```

negz:
[...]

PR #3869, opened 2023-03-15
added keystore support for custom functions auth

### LINKED ISSUE #3868: Add credential helpers support for composition functions
Using ECR for my customm Composition Functions. crossplane-xfn supports only static credentials, which are not applicable for ECR registries

Add go-containerregistry multi-keychain support to support default helpers for ECR, ACR, GCR, Docker Hub, etc

AndrewChubatiuk:

**crossplane-xfn** doesn't support pulling images from:
- registries such as ECR, ACR, which are using temporary credentials
- service account credentials
- kube secrets

Fixes #3868 #3717

I have:

Build container locally with precompiled crun v1.8.1 and tested in AWS EKS with and without an access to ECR


negz (commented):
This looks good to me at first glance. Relates loosely to:

*  https://github.com/crossplane/crossplane/issues/3717

Could you please fill out the "how has this been tested" section of the PR template.

jbw976:
Thanks a lot @AndrewChubatiuk for taking a stab at fixing these issues that have been blocking some early adopters progress on testing composition functions! That's very helpful 🤓 

> made it compatible with crun 1.8.1.

Do the changes to make it compatible with crun 1.8.1 have any impact to compatibility with older versions of crun? Should we be considering pinning the version of crun we use, as suggested by @negz [here](https://github.com/crossplane/crossplane/issues/3807#issuecomment-1450905774)?

bobh66:
Great work @AndrewChubatiuk !  Thanks for digging into this!
Probably a good idea to pin crun to 1.8.1 as part of these changes, if that interface is going to be fragile.

internal/oci/pull.go
```diff
@@ -141,11 +142,19 @@ type RemoteClient struct{}
...
 	opts := parse(o...)
+	if opts.pull == ImagePullPolicyNever {
+		return nil, errors.New(errPullNever)
+	}
 	if opts.auth != nil {
 		return remote.Image(ref, remote.WithContext(ctx), remote.WithAuth(opts.auth))
 	}
-	if opts.pull == ImagePullPolicyNever {
-		return nil, errors.New(errPullNever)
+	keychain, err := k8schain.NewInCluster(ctx, k8schain.Options{})
```
negz, line 151: Will `NewInCluster` always fail if in-cluster auth isn't supported?

Or more generally, what causes `NewInCluster` to fail and what causes `remote.WithAuthFromKeychain` to fail? It may be worth adding comments to clarify.

internal/oci/pull.go
```diff
@@ -141,11 +142,19 @@ type RemoteClient struct{}
...
+	var image ociv1.Image
+	if err == nil {
+		image, err = remote.Image(ref, remote.WithContext(ctx), remote.WithAuthFromKeychain(keychain))
+	}
+	if err == nil {
+		return image, nil
 	}
 	return remote.Image(ref, remote.WithContext(ctx))
```
negz, line 159:
```suggestion
	keychain, err := k8schain.NewInCluster(ctx, k8schain.Options{})
	if err != nil {
		// In-cluster auth not supported. Fall back to no auth.
		return remote.Image(ref, remote.WithContext(ctx))
	}
	img, err := remote.Image(ref, remote.WithContext(ctx), remote.WithAuthFromKeychain(keychain))
	if err != nil {
		// In-cluster auth not supported. Fall back to no auth.	
		return remote.Image(ref, remote.WithContext(ctx))
	}
	return img, nil
```

Would this logic be equivalent? If so I'd prefer it, as I find it easier to follow.

negz (commented):
Thanks @AndrewChubatiuk. A few comments, but this looks good overall.

Please squash/cleanup your git commits before we merge this (but please do keep the `sysfs` fix in a separate commit).

apis/apiextensions/fn/proto/v1alpha1/run_function.proto
```diff
@@ -39,22 +39,11 @@ enum ImagePullPolicy {
...
-  string registry_token = 5;
-}
-
 // ImagePullConfig configures how a Composition Function container should be
 // pulled from a remote OCI registry.
 message ImagePullConfig {
   ImagePullPolicy pull_policy = 1;
-  ImagePullAuth auth = 2;
+  repeated string pull_secrets = 2;
 }
```
negz, line 47: I don't think we can make this change. We need the Crossplane core pod, not `xfn` to resolve pull secrets into image pull auth credentials. This is because we expect other Composition Function runners will exist, and those runners won't always be running in the same Kubernetes cluster as the Crossplane pod (they might not be running in Kubernetes at all).

apis/apiextensions/v1/composition_types.go
```diff
@@ -308,6 +312,14 @@ type ContainerFunction struct {
 	Runner *ContainerFunctionRunner `json:"runner,omitempty"`
 }
 
+// ContainerFunctionImagePullSecret specifies secrets which contain image credentials
+type ContainerFunctionImagePullSecret struct {
+	// Kubernetes secrets name for Composition Function images.
+	// Uses crossplane namespace and service account by default.
+	// +optional
+	Name string `json:"name,omitempty"`
```
negz, line 320: Is this actually optional? It's the only field in the type. If it's not, please remove the `+optional` marker and the `omitempty` tag. (Same for the revision type.)

internal/xpkg/fetch.go
```diff
@@ -135,7 +135,7 @@ func WithServiceAccount(sa string) FetcherOpt {
 func NewK8sFetcher(client kubernetes.Interface, opts ...FetcherOpt) (*K8sFetcher, error) {
 	k := &K8sFetcher{
 		client:    client,
-		transport: remote.DefaultTransport.Clone(),
+		transport: remote.DefaultTransport.(*http.Transport).Clone(),
```
negz, line 138: Why was this change needed?

negz (changes requested):

apis/apiextensions/v1/composition_types.go
```diff
@@ -295,6 +295,10 @@ type ContainerFunction struct {
 	// +kubebuilder:default="20s"
 	Timeout *metav1.Duration `json:"timeout,omitempty"`
 
+	// Secrets for pulling function images.
+	// +optional
+	ImagePullSecrets []ContainerFunctionImagePullSecret `json:"imagePullSecrets,omitempty"`
```
turkenh, line 300:
```suggestion
	ImagePullSecrets []corev1.LocalObjectReference `json:"imagePullSecrets,omitempty"`
```

Any reason not to use the same type as [packagePullSecrets](https://github.com/crossplane/crossplane/blob/580d15ac8ac79e72acec7f95a70c8ddd9026d6b8/apis/pkg/v1/package_types.go#L43)?

internal/controller/apiextensions/composite/composition_ptf.go
```diff
@@ -724,10 +730,34 @@ func RunFunction(ctx context.Context, fnio *iov1alpha1.FunctionIO, fn *v1.Contai
 		return nil, errors.Wrap(err, errDialRunner)
 	}
 
+	k8schainOpts := k8schain.Options{}
+	if sa := os.Getenv("POD_SERVICE_ACCOUNT"); sa != "" {
+		k8schainOpts.ServiceAccountName = sa
+	}
+	if ns := os.Getenv("POD_NAMESPACE"); ns != "" {
+		k8schainOpts.Namespace = ns
+	}
```
turkenh, line 739: We usually avoid reading environment values in arbitrary places in the codebase and, instead, parse them in all in a single location during startup and flow them down with options. 

The two variables here are already [parsed](https://github.com/crossplane/crossplane/blob/580d15ac8ac79e72acec7f95a70c8ddd9026d6b8/cmd/crossplane/core/core.go#L70-L71), and we should better pass them down [similar to how the package manager does](https://github.com/crossplane/crossplane/blob/580d15ac8ac79e72acec7f95a70c8ddd9026d6b8/cmd/crossplane/core/core.go#L171-L172).

internal/controller/apiextensions/composite/composition_ptf.go
```diff
@@ -724,10 +730,34 @@ func RunFunction(ctx context.Context, fnio *iov1alpha1.FunctionIO, fn *v1.Contai
...
+	if fn.ImagePullSecrets != nil {
```
turkenh, line 740: We can drop this `if` check, the `for` loop below would skip if `nil`.

turkenh (commented):
Thanks @AndrewChubatiuk, left a couple of comments!

I don't think this PR will be fixing https://github.com/crossplane/crossplane/issues/3718 though which is more about passing credentials for the function itself rather than pulling it image.

internal/controller/apiextensions/composite/composition_ptf.go
```diff
@@ -724,10 +741,29 @@ func RunFunction(ctx context.Context, fnio *iov1alpha1.FunctionIO, fn *v1.Contai
...
+		k8schainOpts.ImagePullSecrets = append(k8schainOpts.ImagePullSecrets, ips.Name)
+	}
+	keychain, err := k8schain.NewInCluster(ctx, k8schainOpts)
+	if err != nil {
+		if errors.Is(err, rest.ErrNotInCluster) {
+			keychain = nil
+		} else {
+			return nil, err
+		}
+	}
```
negz, line 758:
```suggestion
	// If we're not in a cluster keychain will be nil.
	if err != nil && !errors.Is(err, rest.ErrNotInCluster) {
		return nil, errors.Wrap(err, errNewKeychain)
	}
```

I'd prefer this simpler implementation. I believe `keychain` should always be `nil` if it returns an error.

https://github.com/google/go-containerregistry/blob/v0.14.0/pkg/authn/k8schain/k8schain.go#L61

negz (changes requested):
Sorry this review took so long! I had to switch focus for a while but I expect to be working on Composition Functions again for at least a few weeks. I do think this needs a few more changes - please ping me on Crossplane Slack when you're ready for me to take another look.

negz (approved):
Thanks for persisting with this @AndrewChubatiuk! It looks great. I have some ideas for small improvements, but I can open a follow-up PR for those. I'll copy you on that.

PR #3879, opened 2023-03-17
chore: update created property when building packages with crank

AbrohamLincoln:

When publishing packages to our private Gitlab, I noticed the timestamp associated with the image was always `0001-01-01T00:00:00Z` causing Gitlab to display it improperly as seen below.  I presume this would also cause the automatic registry cleanup to not operate as expected since it uses this property.

image

The changes in this MR add the current timestamp when publishing images using the crank bin.  
Be gentle...I really don't know Go.

I have:

I tested this by running a local registry, making a version of the crank bin with my changes, then pushing a package.
The manifest went from the old timestamp:
```
   "history": [
      {
         "v1Compatibility": "{\"architecture\":\"\",\"config\":{},\"created\":\"0001-01-01T00:00:00Z\",\"id\":\"bf1b087d164de1d5c0b6119fbd60801f2aa0cdebac4d14775f4fd330529f39e5\",\"os\":\"\"}"
      }
   ],
```
to the current timestamp:
```
   "history": [
      {
         "v1Compatibility": "{\"architecture\":\"\",\"config\":{},\"created\":\"2023-03-17T12:07:32.280151-04:00\",\"id\":\"75084279a52165c7384a2808c20e83bc7f3a75e9b739f790095111d35c4ddbb9\",\"os\":\"\"}"
      }
   ],
```

I then pushed the package to our private Gitlab and verified the timestamp was correct:
image

After adding the SOURCE_DATE_EPOCH environment variable support I performed the following tests:
Tested SOURCE_DATE_EPOCH unset:
`"created\":\"0001-01-01T00:00:00Z\"`

Tested SOURCE_DATE_EPOCH set to an invalid string:
kubectl crossplane: error: main.pushCmd.Run(): strconv.ParseInt: parsing "derp": invalid syntax

Tested SOURCE_DATE_EPOCH set to `date +%s`:
`"created\":\"2023-03-17T15:46:43-04:00\"`

### DIFF
--- a/cmd/crank/push.go
+++ b/cmd/crank/push.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"os"
+	"strconv"
 	"time"
 
 	"github.com/google/go-containerregistry/pkg/authn"
@@ -78,11 +79,20 @@ func (c *pushCmd) Run(child *pushChild, logger logging.Logger) error {
 		logger.Debug("Failed to create image from package tarball", "error", err)
 		return err
 	}
-	createTime := v1.Time{Time: time.Now()}
-	img, err = mutate.CreatedAt(img, createTime)
-	if err != nil {
-		logger.Debug("Failed to mutate image with current timestamp", "error", err)
-		return err
+
+	sourceDateEpoch := os.Getenv("SOURCE_DATE_EPOCH")
+	if sourceDateEpoch != "" {
+		sourceDateEpoch, err := strconv.ParseInt(sourceDateEpoch, 10, 64)
+		if err != nil {
+			logger.Debug("Invalid SOURCE_DATE_EPOCH value", "error", err)
+			return err
+		}
+		buildTime := v1.Time{Time: time.Unix(sourceDateEpoch, 0)}
+		img, err = mutate.CreatedAt(img, buildTime)
+		if err != nil {
+			logger.Debug("Failed to mutate image with current timestamp", "error", err)
+			return err
+		}
 	}
 
 	if err := remote.Write(tag, img, remote.WithAuthFromKeychain(authn.DefaultKeychain)); err != nil {
--- a/cmd/crank/push.go
+++ b/cmd/crank/push.go
@@ -18,9 +18,12 @@ package main
 
 import (
 	"os"
+	"time"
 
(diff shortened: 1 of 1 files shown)


hasheddan (commented):
@AbrohamLincoln this is intentional as setting a timestamp results in non-reproducible builds. While tools like Docker follow this pattern we do not recommend it.

AbrohamLincoln:
Would it be acceptable to toggle this with the `SOURCE_DATE_EPOCH` envar?  The default behavior would be the original behavior, setting the envar to a value such as `date +%s` would result in the timestamp being set.

cmd/crank/push.go
```diff
@@ -75,6 +79,22 @@ func (c *pushCmd) Run(child *pushChild, logger logging.Logger) error {
 		logger.Debug("Failed to create image from package tarball", "error", err)
 		return err
 	}
+
+	sourceDateEpoch := os.Getenv("SOURCE_DATE_EPOCH")
```
negz, line 83: Can we make this a flag instead? I believe Kong (our CLI library) allows you to specify `env:"SOURCE_DATE_EPOCH"` as a struct tag. This way you'd still be able to set this as an environment variable, but it would be a bit more discoverable and better documented.

negz (commented):
@hasheddan What do you think about putting this behind a flag/env var?

I suppose one downside is that some Crossplane packages would have a created property set while others wouldn't. Not sure if that's an issue though.

hasheddan (commented):
> @hasheddan What do you think about putting this behind a flag/env var?

> I suppose one downside is that some Crossplane packages would have a created property set while others wouldn't. Not sure if that's an issue though.

I feel comfortable with this and don't see it as an issue for some packages to have timestamps while others do not given that Crossplane packages may be constructed via any OCI tooling, and tools may choose to implement different behavior. The decision not to use timestamps is a feature of `crank` rather than a requirement of the xpkg specification 👍🏻

negz (changes requested):
I think we just need to change the env var to a flag and this is good to go.

negz:
Breadcrumbs to https://www.chainguard.dev/unchained/designing-build-date-epoch-in-chainguard-images, which seems like a clever way to solve this. Might be more involved than we'd like to commit to though.

PR #3884, opened 2023-03-21
Generate TLS certificates for ESS and use them 

ezgidemirel:

In this change we started to generate TLS certificates and mount them to XP and provider pods when **the External Secret Stores alpha feature is enabled**. These certificates will be used for pluggable secret store authentications. Design details can be found in this [doc](https://github.com/crossplane/crossplane/blob/master/design/one-pager-pluggable-secret-stores.md). 

Fixes https://github.com/upbound/team-control-planes/issues/966

I have:

E2E test with following steps:

1) Create a local kind cluster
```
/cluster/local/kind.sh up
```
2) Prepare Vault as described [here](https://github.com/crossplane/docs/blob/master/content/knowledge-base/integrations/vault-as-secret-store.md#prepare-vault)
3) Build and install Crossplane on this branch
```
make
HELM3_FLAGS="--set args={--debug,--enable-external-secret-stores} " ./cluster/local/kind.sh helm-install crossplane-system
```
4) Install a custom provider-gcp and create provider config
```
kubectl apply -f provider-gcp-ess.yaml
```
5) Install storeconfigs
```
kubectl apply -f storeconfig-xp.yaml -f storeconfig-gcp.yaml
```
6) Install [ESS plugin Vault](https://github.com/crossplane-contrib/ess-plugin-vault)
```
helm upgrade --install ess-plugin-vault oci://xpkg.upbound.io/crossplane-contrib/ess-plugin-vault --namespace crossplane-system -f values.yaml
```
7) Install vaultconfig
```
kubectl apply -f vaultconfig.yaml
```
8) Install XRD and composition:
```
kubectl apply -f xrd.yaml -f composition.yaml
```
9) Install claim
```
kubectl apply -f claim.yaml
```
10) Portforward Vault service and verify that secrets are written
```
kubectl port-forward svc/vault 8200:8200 -n vault-system
```
11) Delete claim and verify that secrets are deleted from Vault
```
kubectl delete -f claim.yaml
```

All the files mentioned above can be found [here](https://gist.github.com/ezgidemirel/190721cbc6e3b3449697888dbc001911).

Also tested two upgrade scenarios where the feature is enabled/disabled in the previous XP version and verified that k8s secrets with certificates are created, and mounted to XP and provider pods after the upgrade as expected. 


cluster/charts/crossplane/templates/deployment.yaml
```diff
@@ -92,6 +92,10 @@ spec:
           - name: "WEBHOOK_SERVICE_PORT"
             value: "9443"
           {{- end }}
+          {{- if has "--enable-external-secret-stores" .Values.args }}
```
turkenh, line 95: We may want to define a new variable in `_helpers.tpl` as a boolean like `essEnabled` and use it instead if checking the args array everywhere.

cluster/charts/crossplane/templates/deployment.yaml
```diff
@@ -92,6 +92,10 @@ spec:
...
+          - name: "ESS_GENERATE_CERTS"
```
turkenh, line 96: I remember we discussed this but now I feel like it would be cleaner if we pass the name of the secret names and use them as a feature flag to decide whether the initializer should generate certs or not.

I mean:

```yaml
          - name: "ESS_CLIENT_CERTS_SECRET_NAME"
            value: ess-client-certs
          - name: "ESS_SERVER_CERTS_SECRET_NAME"
            value: ess-server-certs
```

cluster/charts/crossplane/templates/secret.yaml
```diff
@@ -10,4 +11,36 @@ metadata:
...
+apiVersion: v1
+kind: Secret
+metadata:
+  name: ess-server-certs
+  namespace: {{ .Release.Namespace }}
+type: Opaque
+---
+# The reason this is created empty and filled by the init container is that it's
+# mounted by the actual container, so if it wasn't created by Helm, then the
+# deployment wouldn't be deployed at all with secret to mount not found error.
```
turkenh, line 39: This comment seems to be valid only for `ess-client-certs`. So, we should either remove other secrets from here or still have them to be consistent but update the comment accordingly.

cluster/crds/secrets.crossplane.io_storeconfigs.yaml
```diff
@@ -118,11 +118,39 @@ spec:
...
+                    type: string
+                type: object
               type:
                 default: Kubernetes
                 description: Type configures which secret store to be used. Only the
                   configuration block for this store will be used and others will
                   be ignored if provided. Default is Kubernetes.
+                enum:
+                - Kubernetes
+                - Vault
```
turkenh, line 152: Can we mark `Vault` as deprecated? Probably would require a runtime change though.

cmd/crossplane/core/core.go
```diff
@@ -76,6 +79,8 @@ type startCommand struct {
 	SyncInterval     time.Duration `short:"s" help:"How often all resources will be double-checked for drift from the desired state." default:"1h"`
 	PollInterval     time.Duration `help:"How often individual resources will be checked for drift from the desired state." default:"1m"`
 	MaxReconcileRate int           `help:"The global maximum rate per second at which resources may checked for drift from the desired state." default:"10"`
+	ESSTLSSecretName string        `help:"The name of the TLS Secret that will be used by the client of External Secret Store plugins and providers." env:"ESS_TLS_SECRET_NAME"`
```
turkenh, line 82:
```suggestion
	ESSTLSSecretName string        `help:"The name of the TLS Secret that will be used by Crossplane and providers as clients of External Secret Store plugins." env:"ESS_TLS_SECRET_NAME"`
```

cmd/crossplane/core/core.go
```diff
@@ -76,6 +79,8 @@ type startCommand struct {
...
+	ESSTLSCertsDir   string        `help:"The path of the folder which will store TLS certificates to be used by core Crossplane, External Secret Store plugins and providers." env:"ESS_TLS_CERTS_DIR"`
```
turkenh, line 83:
```suggestion
	ESSTLSCertsDir   string        `help:"The path of the folder which will store TLS certificates to be used by Crossplane and providers for communicating with External Secret Store plugins." env:"ESS_TLS_CERTS_DIR"`
```

cmd/crossplane/core/core.go
```diff
@@ -138,6 +139,22 @@ func (c *startCommand) Run(s *runtime.Scheme, log logging.Logger) error { //noli
...
 	}
 
+	if c.EnableExternalSecretStores {
+		feats.Enable(features.EnableAlphaExternalSecretStores)
+		log.Info("Alpha feature enabled", "flag", features.EnableAlphaExternalSecretStores)
+
+		tlsConfig, err := certificates.LoadMTLSConfig(filepath.Join(c.ESSTLSCertsDir, initializer.SecretKeyCACert),
+			filepath.Join(c.ESSTLSCertsDir, initializer.SecretKeyTLSCert), filepath.Join(c.ESSTLSCertsDir, initializer.SecretKeyTLSKey), false)
+		if err != nil {
+			return errors.Wrap(err, "Cannot load TLS certificates")
```
turkenh, line 149:
```suggestion
			return errors.Wrap(err, "cannot load TLS certificates for ESS")
```

internal/controller/pkg/manager/reconciler.go
```diff
@@ -179,6 +188,9 @@ func SetupProvider(mgr ctrl.Manager, o controller.Options) error {
 	if o.WebhookTLSSecretName != "" {
 		opts = append(opts, WithWebhookTLSSecretName(o.WebhookTLSSecretName))
 	}
+	if o.ESSOptions != nil && o.ESSOptions.TLSConfig != nil {
```
turkenh, line 191:
```suggestion
	if o.ESSOptions != nil && o.ESSOptions.TLSSecretName != "" {
```

go.mod
```diff
@@ -36,8 +36,10 @@ require (
 	sigs.k8s.io/yaml v1.3.0
 )
 
+require cloud.google.com/go/compute/metadata v0.2.3 // indirect
```
turkenh, line 39: nit: should this be grouped below?

turkenh (approved):

lsviben (approved):

PR #3892, opened 2023-03-22
fix(helm): #3891 - add hostNetwork functionality to crossplane pod

### LINKED ISSUE #3891: Crossplane does not function on EKS with a custom CNI
### What happened?

Crossplane, installed with the Helm chart, fails to create compositions with:
```
Warning  EstablishComposite  86s (x13 over 7m29s)   defined/compositeresourcedefinition.apiextensions.crossplane.io
cannot add composite resource finalizer: cannot update object: Internal error occurred: failed calling webhook
"compositeresourcedefinitions.apiextensions.crossplane.io": failed to call webhook:
Post "https://crossplane-webhooks.crossplane-system.svc:9443/validate-apiextensions-crossplane-io-v1-compositeresourcedefinition?timeout=10s":
Address is not allowed
```

### Why is it happening?
When using a custom CNI like Cilium or Calico on EKS the control plane can't route to pods for webhooks.

### How can this be fixed? 
The simplest, and most common way of dealing with this is to run pods with `hostNetwork: true`. Add this to the Helm chart to allow users to run crossplane in this way.

**This is a placeholder and will be accompanied by a PR shortly**

portswigger-tim:

Adds the ability to set `hostNetwork` on the crossplane pod for running in an environment where the Kubernetes control-plane can not access pods directly. For example, when running an EKS cluster with an alternative CNI such as Cilium or Calico. 

Fixes #3891

### DIFF
--- a/cluster/charts/crossplane/README.md
+++ b/cluster/charts/crossplane/README.md
@@ -80,6 +80,7 @@ and their default values.
 | `customAnnotations` | Custom annotations to add to the Crossplane deployment and pod | `{}` |
 | `serviceAccount.customAnnotations` | Custom annotations to add to the serviceaccount of Crossplane | `{}` |
 | `priorityClassName` | Priority class name for Crossplane and RBAC Manager (if enabled) pods | `""` |
+| `hostNetwork` | Enable hostNetwork for Crossplane | `false` |
 | `resourcesCrossplane.limits.cpu` | CPU resource limits for Crossplane | `100m` |
 | `resourcesCrossplane.limits.memory` | Memory resource limits for Crossplane | `512Mi` |
 | `resourcesCrossplane.requests.cpu` | CPU resource requests for Crossplane | `100m` |
--- a/cluster/charts/crossplane/README.md
+++ b/cluster/charts/crossplane/README.md
@@ -80,7 +80,7 @@ and their default values.
 | `customAnnotations` | Custom annotations to add to the Crossplane deployment and pod | `{}` |
 | `serviceAccount.customAnnotations` | Custom annotations to add to the serviceaccount of Crossplane | `{}` |
 | `priorityClassName` | Priority class name for Crossplane and RBAC Manager (if enabled) pods | `""` |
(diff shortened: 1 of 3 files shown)


cluster/charts/crossplane/README.md
```diff
@@ -80,6 +80,7 @@ and their default values.
 | `customAnnotations` | Custom annotations to add to the Crossplane deployment and pod | `{}` |
 | `serviceAccount.customAnnotations` | Custom annotations to add to the serviceaccount of Crossplane | `{}` |
 | `priorityClassName` | Priority class name for Crossplane and RBAC Manager (if enabled) pods | `""` |
+| `hostNetwork` | Enable hostNetwork for Crossplane | `false` |
```
hasheddan, line 83: @portswigger-tim if we add the ability to set this, it would be great to add a warning that this is giving Crossplane lots of privilege in a cluster and should be handled with caution.

AhmedGrati (approved):

cluster/charts/crossplane/README.md
phisco:
```suggestion
| `hostNetwork` | Enable hostNetwork for Crossplane. Caution: setting it to true means Crossplane's Pod will have high privileges. | `false` |
```

cluster/charts/crossplane/values.yaml.tmpl
```diff
@@ -11,6 +11,8 @@ nodeSelector: {}
 tolerations: []
 affinity: {}
 
+hostNetwork: false
```
phisco, line 14:
```suggestion
# -- Enable hostNetwork for Crossplane. Caution: setting it to true means Crossplane's Pod will have high privileges.
hostNetwork: false
```

phisco (changes requested):

phisco:
@portswigger-tim any feedback on the above suggestions?

portswigger-tim:
Apologies, I've been largely ignoring public Github lately 😬 . Committed your suggestions @phisco, thanks :)

phisco (approved):

turkenh (approved):

PR #3917, opened 2023-03-27
feat: ClampMin and ClampMax math transforms

julienduchesne:
Usecase:

For building cloud-agnostic bucket lifecycle policies, I have a parameter called `expireObjectsDays`. I use the value `0` of that parameter to disable object expiry. In AWS, I use a `match` transform to set the `status` of the lifecycle policy to `Disabled` when the days parameter is zero: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket_lifecycle_configuration#status

However, I also have a patch on the actual days attribute. Right now, I haven't found an easy way of setting a minimum of 1 while also keeping the original value if it was already higher. It would be easily done with a clamp:

```yaml
                      fromFieldPath: spec.parameters.expireObjectsDays
                      toFieldPath: spec.forProvider.rule[0].expiration[0].days
                      transforms:
                        - type: math
                          math:
                            type: ClampMin
                            clampMin: 1
```

I called the transforms `clampMin` and `clampMax` because it's clearer what it does rather than `min` and `max` (does it clamp to that value OR find the max between the input and given param?)

I have:

## How has this code been tested

Unit tests + deployed in my cluster and tested with the case shown above ⏫


jbw976 (commented):
Great to see more additions to the transform capabilites @julienduchesne!! 💪 

Similar to #3919, would it be possible for you to update the issue with a bit more context? e.g. an example scenario of how to use this functionality, how you tested the deployed code, etc.  That will help reviewers understand your full intent better, especially since this makes new additions to the composition API/UX.  Thank you! 🙇‍♂️

turkenh (approved):
Thanks @julienduchesne!

Similar to https://github.com/crossplane/crossplane/pull/3919, could you make sure the documentation is updated?

julienduchesne:
> Thanks @julienduchesne!
> 
> Similar to #3919, could you make sure the documentation is updated?

I added a doc PR here: https://github.com/crossplane/docs/pull/391

PR #3919, opened 2023-03-27
feat: In `match` transform, allow fallback to input

julienduchesne:
This would be useful to make conditional transformations to an input but, if nothing matches, keep the input unmodified As it is, anything that goes through this transform is squashed

Usecase:

One case this is useful is for the `newer_noncurrent_versions` attribute: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket_lifecycle_configuration#newer_noncurrent_versions. It actually takes in a string rather than a number. When the attribute should not be set, it has to be an empty string, not `"0"`, which fails because the value has to be higher than 0

So with this new match feature, I can transform my composition parameter to be replaced by an empty string when it's zero, but keep all other values:
```
                   -  type: FromCompositeFieldPath
                      fromFieldPath: spec.parameters.keepNonCurrentVersions
                      toFieldPath: spec.forProvider.rule[0].noncurrentVersionExpiration[0].newerNoncurrentVersions
                      transforms:
                        - parameters:
                            toType: string
                          type: convert
                        - fallbackTo: Input
                          patterns:
                            - literal: "0"
                              result: ""
                              type: literal
                          type: match
                     
```

I have:

## How has this code been tested

Unit tests + deployed in my cluster and tested with the case shown above ⏫

### DIFF
--- a/apis/apiextensions/v1/composition_transforms.go
+++ b/apis/apiextensions/v1/composition_transforms.go
@@ -290,6 +290,8 @@ type MatchTransform struct {
 	// The fallback value that should be returned by the transform if now pattern
 	// matches.
 	FallbackValue extv1.JSON `json:"fallbackValue,omitempty"`
+	// If set to true, the input value will be returned if no pattern matches.
+	FallbackToInput bool `json:"fallbackToInput,omitempty"`
 }
 
 // Validate checks this MatchTransform is valid.
--- a/apis/apiextensions/v1/composition_transforms.go
+++ b/apis/apiextensions/v1/composition_transforms.go
@@ -300,6 +300,9 @@ type MatchTransform struct {
 	// matches.
 	FallbackValue extv1.JSON `json:"fallbackValue,omitempty"`
 	// Determines to what value the transform should fallback if no pattern matches.
+	// +optional
+	// +kubebuilder:validation:Enum=Value;Input
+	// +kubebuilder:default=Value
 	FallbackTo MatchFallbackTo `json:"fallbackTo,omitempty"`
 }
 
--- a/apis/apiextensions/v1/composition_transforms.go
+++ b/apis/apiextensions/v1/composition_transforms.go
@@ -279,6 +279,15 @@ func (m *MapTransform) MarshalJSON() ([]byte, error) {
 	return json.Marshal(m.Pairs)
 }
 
+// MatchFallbackTo defines how a match operation will fallback.
+type MatchFallbackTo string
+
+// Valid MatchFallbackTo.
+const (
+	MatchFallbackToTypeValue MatchFallbackTo = "Value"
+	MatchFallbackToTypeInput MatchFallbackTo = "Input"
+)
+
 // MatchTransform is a more complex version of a map transform that matches a
 // list of patterns.
 type MatchTransform struct {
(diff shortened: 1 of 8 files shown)


julienduchesne:
> Thanks for making this contribution to refine the match transform's behavior - could you please provide a bit more details in the PR body that describes this behavior, an example scenario for when it would be useful, and how you've tested the functionality? it always helps the review process to understand this context and details 🙇

👋 I described my exact use case above and also tested the real life scenario in my cluster!

apis/apiextensions/v1/composition_transforms.go
```diff
@@ -110,6 +110,8 @@ type MatchTransform struct {
 	// The fallback value that should be returned by the transform if now pattern
 	// matches.
 	FallbackValue extv1.JSON `json:"fallbackValue,omitempty"`
+	// If set to true, the input value will be returned if no pattern matches.
```
turkenh, line 113: nit: should we mention that this is mutually exclusive with FallbackValue?

turkenh (approved):
Thanks for your contribution @julienduchesne! 

Could you make sure that this is included in the [documentation](https://github.com/crossplane/docs/blob/master/content/master/concepts/composition.md#transform-types)?

apis/apiextensions/v1/composition_transforms.go
negz: An alternative API here would be to use an enum. For example:

```yaml
fallbackTo: Value
fallbackValue: whatever
```

```yaml
fallbackTo: Input
```

This fits with [Kubernetes API conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types), which say:

> Think twice about bool fields. Many ideas start as boolean but eventually trend towards a small set of mutually exclusive options. Plan for future expansions by describing the policy options explicitly as a string type alias (e.g. TerminationMessagePolicy).

julienduchesne: Good idea. I've changed the PR to use that API

julienduchesne:
> Thanks for your contribution @julienduchesne!
> 
> Could you make sure that this is included in the [documentation](https://github.com/crossplane/docs/blob/master/content/master/concepts/composition.md#transform-types)?

Done. PR Here: https://github.com/crossplane/docs/pull/392

apis/apiextensions/v1alpha1/zz_generated.revision_types.go
```diff
@@ -581,6 +590,8 @@ type MatchTransform struct {
 	// The fallback value that should be returned by the transform if now pattern
 	// matches.
 	FallbackValue extv1.JSON `json:"fallbackValue,omitempty"`
+	// Determines to what value the transform should fallback if no pattern matches.
```
turkenh, line 593:
```suggestion
	// Determines to what value the transform should fallback if no pattern matches.
	// +optional
	// +kubebuilder:validation:Enum=Value;Input
	// +kubebuilder:default=Value
```

Could you also add enum and default marker comments?

turkenh:
@julienduchesne just pushed a commit after regenerating with the addition of the enum validation and default value.

PR #3929, opened 2023-03-31
Keep existing finalizers when updating package revision's objects

branden:

Finalizers I set on a CRD belonging to a Crossplane provider are removed whenever the provider's `ProviderRevision` is reconciled. I've also seen that a controller re-applying finalizers as they're removed can provoke "object has been modified" errors during `ProviderRevision` reconciliation, causing the `ProviderRevision` to become unhealthy. Such failed reconciliations can become more likely as a provider installs more CRDs, since the entire set of CRDs is created/updated with each reconciliation. provider-aws is particularly sensitive to this.

This PR modifies the `ProviderRevision` reconciler to keep existing finalizers when updating objects.

I have:

I didn't see a clear opportunity for a regression test, but I'm happy to add one with guidance on where it would belong.

I tested this manually according to the following steps:

1. Deploy the latest Crossplane release and install a provider, such as [provider-terraform](https://github.com/upbound/provider-terraform). Wait for its `ProviderRevision` to become healthy.
2. Add a finalizer to a CRD belonging to the provider, such as `workspaces.tf.upbound.io`.
3. Add any label to the `ProviderRevision` to force a reconciliation.
4. Observe that the finalizer added in step 2 has been removed.
5. Override the Crossplane image with an image built from this PR.
6. Repeat steps 2 and 3, then observe that the finalizer is not removed and the `PackageRevision` is healthy.

### DIFF
--- a/internal/controller/pkg/revision/establisher.go
+++ b/internal/controller/pkg/revision/establisher.go
@@ -364,6 +364,7 @@ func (e *APIEstablisher) update(ctx context.Context, current, desired resource.O
 		return err
 	}
 	desired.SetResourceVersion(current.GetResourceVersion())
+	desired.SetFinalizers(current.GetFinalizers())
 	return e.client.Update(ctx, desired, opts...)
 }
 


turkenh:
@branden, as you have found out, Provider CRDs are owned by `ProviderRevision` and managed by the package manager, and they are periodically reconciled for the desired state. AFAIU, this PR proposes to make an exception for finalizers so that they can be modified/controlled by a third party.

Could you give some more details on the use case (i.e. why we need this?) so that we can discuss if this is the right solution?

turkenh:
Thanks @branden.

I hesitate to make exceptions like this as it opens a door for further modifications on the resources owned by XP controllers, e.g. I want this label/annotation on CRDs. Also, for consistency, it is usual that an XP controller manages a resource and wholly owns its desired state reverting any external modification, e.g. you can not modify a provider controller deployment. If we really need this functionality, we should rather consider opening an API for changing desired CRDs.  

For your specific use case, did you consider running periodic garbage collections listing all CRDs and regenerating the config if needed? Another idea is since MR CRDs are always coming from Providers/ProviderRevisions, watching them trigger a config regeneration might also work.

hasheddan (commented):
I want to echo @turkenh's concerns here 👍🏻 Also, I think Hasan makes a good recommendation about watching `ProviderRevision`, which has a reference to all objects it owns in [`status.objectRefs`](https://doc.crds.dev/github.com/crossplane/crossplane/pkg.crossplane.io/ProviderRevision/v1@v1.11.3#status-objectRefs). One other thing worth noting is that Provider Runtime Interface could allow for custom behavior in this pipeline, which sounds like it could be useful in this case.

branden:
Thanks @turkenh @hasheddan. I'll take a shot at watching `ProviderRevision` instead 👍

Issue #3770, opened 2023-02-14
Future of `Composition Environment`


MisterMX:
### What problem are you facing?

Support for `EnvironmentConfigs` was added in Crossplane v1.11.0.

### How could Crossplane help solve your problem?

We need a plan to promote it to v1beta1, which may include accepting it as is, changing it, or dropping the feature. Community feedback would help here.
```[tasklist]
### Tracked issues
- [ ] https://github.com/crossplane/crossplane/pull/6016
- [ ] https://github.com/crossplane/docs/pull/824
- [x] https://github.com/crossplane/crossplane/pull/5938
- [x] https://github.com/crossplane-contrib/function-environment-configs/pull/44
- [x] #3965
- [x] https://github.com/crossplane/docs/issues/305
- [ ] #4583
- [ ] #4275
- [ ] https://github.com/crossplane/crossplane/issues/3454: not strictly related, but causing confusion with this feature
- [ ] patch data back to EnvironmentConfigs and have it persistently stored?
- [x] https://github.com/crossplane/crossplane/pull/3981
- [ ] #3723
- [x] #3941
- [x] #3865
- [ ] #3967
- [x] #4080
- [x] #4070
- [ ] #3926
- [ ] #4274
- [ ] #4393
- [ ] https://github.com/crossplane/crossplane/issues/4591
- [ ] https://github.com/crossplane-contrib/function-environment-configs/issues/43
```

blakebarnett:
We are using it successfully as-is.

Some clarity around the patch ordering in documentation and validation that a patch is just not going to work rather than having it fail silently would be great beta criteria. For example if attempting to patch a composed resource and then trying to use that same field elsewhere as if it were a composite field a maybe validatingAdmissionWebhook could deny it as invalid?

ytsarev:
I agree, https://github.com/crossplane/crossplane/issues/3454 behavior should be documented.

Otherwise it is totally unclear why construction like https://github.com/ytsarev/composition-environment-example/blob/main/xvpc/composition.yaml#L36-L38 does not work.

It's not the best UX in general so if there is chance to change the order for ToEnvironment patches, we should try to do it.

@plumbis where are old doc snippets are living now?

E.g. I'm failing to find https://github.com/crossplane/crossplane/pull/3007/files#diff-57c8cd7167236fac0edefa6b72c183bb0fdbd21bdc8ec0a521407eaf436ebc0e in the current state of repos.

[...]

PR #3931, opened 2023-03-31
Select multiple environment configs

P0t4T0o:

related to: https://github.com/crossplane/crossplane/issues/3770, see https://github.com/crossplane/crossplane/issues/3770#issuecomment-1491628182

with introduced changes composite will aggregate all environment configs which were selected by labels

Fixes #

I have:

via unit tests & in local environment


MisterMX (approved):

internal/controller/apiextensions/composite/environment/selector.go
```diff
@@ -106,28 +112,33 @@ func (s *APIEnvironmentSelector) buildEnvironmentConfigRefFromRef(ref *v1.Enviro
...
 		matchLabels[m.Key] = val
 	}
 	res := &v1alpha1.EnvironmentConfigList{}
 	if err := s.kube.List(ctx, res, matchLabels); err != nil {
-		return corev1.ObjectReference{}, errors.Wrap(err, errListEnvironmentConfigs)
+		return []corev1.ObjectReference{}, errors.Wrap(err, errListEnvironmentConfigs)
 	}
 	if len(res.Items) == 0 {
-		return corev1.ObjectReference{}, errors.New(errListEnvironmentConfigsNoResult)
+		return []corev1.ObjectReference{}, errors.New(errListEnvironmentConfigsNoResult)
```
bobh66, line 129: I know it's technically not part of this change, but it seems like it should be valid for a Selector to return no results, just as it is valid for a CompositionSelector to return no results and fallback to the defaultComposition.  In this case the environment may be constructed of a variety of named and selected EnvironmentConfigs, as well as patches from the Composite, and the absence of one or more selectors may not be a problem, and may even be expected.

P0t4T0o:
> Does this also happen to address #3941 too? 🙏

can I include changes mentioned here: https://github.com/crossplane/crossplane/issues/3941#issuecomment-1499203688 in this PR or do you want to introduce policy? that could be done in separate PR

P0t4T0o:
@MisterMX @jbw976 
I created new PR: https://github.com/crossplane/crossplane/pull/3981 which addresses aggregation of all env configs with matching labels and introduces resolve policy. Could you please review changes? Will then close this PR in favour of the new one

phisco:
This was actually superseded by https://github.com/crossplane/crossplane/pull/3981, which was merged.

Issue #2869, opened 2022-01-28
Add ability to disable certain CRDs on installation.


mvaal:
Edit from @negz:

```
I0328 13:34:48.091330 25599 request.go:665] Waited for 1.033772408s due to client-side throttling, not priority and fairness, request: GET:https://xyz.com/apis/pkg.crossplane.io/v1alpha1?timeout=32s
```

**If you're finding this issue because you're seeing messages like the above when you run kubectl please make sure you're using the latest version of kubectl - i.e. 1.25.0 or above.** Similarly, other tools built against client-go 0.25.0 or above should not experience these issues that are due to overly aggressive client-side rate limits.

Original issue follows below.

---

### What problem are you facing?
We want to limit which resources are installed on our cluster so that we can restrict which resources can be used in AWS by our cluster's users.
In addition to this, we are having throttling issues due to the large amount of CRDs, being able to turn of 90% of the the unused CRDs would make the throttling issues go away.

### Related issues

- [x] #2895 
- [x] https://github.com/crossplane/docs/issues/336

muvaf:
@mvaal I have transferred the issue to core crossplane.

@liubog2008 opened https://github.com/crossplane/provider-aws/pull/1076 to do this for controllers, I think consolidating the whole discussion in this issue would be more helpful. I've [answered](https://github.com/crossplane/provider-aws/pull/1076#issuecomment-1017162860) why we didn't do it for Jet-based providers which have way more CRDs.

It seems like the main concerns are as following:
* Ability to limit capabilities of users in the cluster.
  * This can be addressed with RBAC for most cases but in cases where you don't want `Composition` authors be able to use certain managed resource types, there isn't much you can do other than using OpenPolicyAgent because the creator of managed resources resulting from `Composition` is Crossplane pod, which has permissions over all CRDs packages bring.
* Throttling of client requests.
  * Even if throttling wasn't there, the number of requests discovery client makes increases linearly with number of CRDs which makes non-cached queries run longer.

@mvaal @liubog2008 could you add other concerns if you have any? I'd like to have all the feedback in one place so that when we reconsider, we have more data to depend on.

cc @negz @hasheddan

mvaal:
You pretty much nailed it. We are using OPA, but we would rather just not worry about the unused resources at all and add them when required.
As for the throttling issue, its mostly in reference to this: https://github.com/kubernetes/kubectl/issues/1126
They are already talking about crossplane in that thread as well.

[...]

PR #3939, opened 2023-04-03
Proposal: Break Up Large Providers by Service

### LINKED ISSUE #3754: Proposal: Break up large providers by service

Crossplane installs a lot of CRDs. More specifically, some of the most widely used Crossplane providers install a lot of CRDs. Upbound's [Official AWS Provider](https://marketplace.upbound.io/providers/upbound/provider-aws/v0.29.0) for example currently installs almost 800 CRDs.

Per the [CRD scaling one pager](https://github.com/crossplane/crossplane/blob/2536c8/design/one-pager-crd-scaling.md), Kubernetes and the tools in its ecosystem struggle when so many CRDs are installed. The API server uses a lot of memory (around 3MB per CRD). Many tools like `kubectl` and libraries like `client-go` are designed under the assumption that only 10s of CRDs will exist, and thus rely on inefficient queries. Crossplane maintainers and others (notably @jonnylangefeld and @apelisse) have made [a lot of headway](https://blog.upbound.io/scaling-kubernetes-to-thousands-of-crds/) into improving this situation. Unfortunately it has become evident that we can't wait for the ecosystem to catch up to our scale. Despite there being a handful of performance improvements with each new Kubernetes release it may take years for everyone to be running versions of Kubernetes and Kubernetes tooling that handles hundreds or thousands of CRDs well.

In addition to the performance issues that come with loading a lot of CRDs, per https://github.com/crossplane/crossplane/issues/2869, some folks have expressed security concerns about having to deploy APIs and controllers that they don't plan to use. Others have stated they just plain don't like the "bloat" of knowing that there's a bunch of superfluous things installed.

* https://github.com/crossplane/crossplane/issues/2869
* https://github.com/crossplane/docs/issues/336

Per the preceding issues, some folks have started rolling their own Crossplane providers that include only the types they need. These are typically large existing providers (like `provider-aws`) forked and with the superfluous controllers and types removed.

I propose that we break up large providers by service. That is, instead of `provider-aws` we would have `provider-aws-rds`, `provider-aws-eks`, etc. The goal is to lower the ratio of installed-to-used CRDs, and intuitively doing that by service seems like about the right ratio to me. Thanks to the Crossplane package manager I believe it should be possible (perhaps with a few package manager tweaks) to make for example `provider-aws` a "meta-package" that pulled in all of the smaller providers for backward compatibility.

I like this approach because:

* It's opt-in, rather than opt-out. You install the things you do need, as opposed to turning off the things you don't need.
* No changes would be required to the many providers that already have a low installed-to-used CRD ratio. Providers like `provider-terraform` and `provider-helm` that only have 2-3 CRDs can stay as they are.
* It increases the granularity at which providers can be updated. For example if you had to stick with a certain version of `provider-aws-rds` you could still update `provider-aws-eks` to the latest version.
* It will likely increase the scalability of providers, by sharding work across multiple provider processes (i.e. Pods).

Some examples for context. Presuming that each provider mapped to an existing API group (e.g. `provider-aws-rds` mapped to `rds.aws.upbound.io`):

* `provider-aws-rds` would have [~22 CRDs](https://marketplace.upbound.io/providers/upbound/provider-aws/v0.29.0/crds?query=rds.aws.upbound.io).
* `provider-aws-eks` would have [~6 CRDs](https://marketplace.upbound.io/providers/upbound/provider-aws/v0.29.0/crds?query=eks.aws.upbound.io).
* `provider-aws-s3` would have [~23 CRDs](https://marketplace.upbound.io/providers/upbound/provider-aws/v0.29.0/crds?query=s3.aws.upbound.io).
* `provider-aws-iam` would have [~22 CRDs](https://marketplace.upbound.io/providers/upbound/provider-aws/v0.29.0/crds?query=iam.aws.upbound.io).

negz:

Fixes #3754

This design document proposes that the 6-7 largest Crossplane providers be broken down into smaller, service-scoped ones. This would help folks install fewer CRDs, thus improving the ratio of installed-to-used Crossplane CRDs. Installing fewer CRDs is necessary to workaround performance issues in the Kubernetes API server and Kubernetes clients.

I have:

I proof-read it. 😄

### DIFF
--- a/design/design-doc-smaller-providers.md
+++ b/design/design-doc-smaller-providers.md
@@ -53,9 +53,7 @@ present at:
 * https://blog.upbound.io/scaling-kubernetes-to-thousands-of-crds/
 
 Apart from performance issues, some folks have expressed concerns about the
-security and general “bloat” of having many unused CRDs installed. I believe the
-security concerns to be an educational issue - I haven't yet seen a convincing
-argument for removing CRDs as opposed to using RBAC.
+security and general “bloat” of having many unused CRDs installed.
 
 Crossplane maintainers have invested in improving the performance of the
 Kubernetes control plane and clients in the face of many CRDs. These
@@ -72,6 +70,30 @@ We have explored the following options:
 * Make it possible to filter unnecessary CRDs by configuring a provider.
 * Break the biggest providers up into smaller, service-scoped providers.
 * Load CRDs into the API server lazily, on-demand.
+
+## Goals
+
+The goal of this proposal is to eliminate _or significantly defer_ the
+performance issues that arise when installing too many CRDs. By "deferring"
+performance issues, I mean buying time for the Kubernetes ecosystem to improve
+its performance under CRD load.
+
+## Non-goals
+
+The goal is not that no Crossplane deployment _ever_ crosses the ~500 CRD mark.
+Instead, that should be a rare edge case until the vast majority of Kubernetes
+deployments can handle it without issue.
+
+This proposal _does not_ intend to address any security concerns related to
+unused CRDs being present in a Crossplane deployment. At this time the
+significance of these security concerns is not well understood. It's also
+unclear how many community members are concerned, relative to the overwhelming
+concern about performance.
+
+At the present time I do not believe installing a CRD (and controller) that you
+don't intend to use poses a significant risk. At least two layers of security
+prevent anyone using the Crossplane Managed Resources (MRs) these CRDs define -
+Kubernetes RBAC in front of the API call, and cloud IAM permissions behind it.
 
 ## Proposal
 
--- a/design/design-doc-smaller-providers.md
+++ b/design/design-doc-smaller-providers.md
@@ -424,6 +424,50 @@ could consider:
   importing the Terraform provider code directly. Terraform makes this hard to
   do, but it’s possible. It appears to be what Config Connector does.
 
+### Rollout and Migration
+
+As with any major Crossplane change, it's important to allow the community to
+experiment and provide feedback before we commit. Given that this change is
+largely decoupled from Crossplane core and its release cycle, I propose that we:
+
+* Continue to publish the existing, too-large, providers for now.
+* Begin publishing smaller, service scoped providers in parallel.
+
+These smaller providers would be marked as 'release candidates' to begin with. I
+prefer this to 'alpha' in this context as it better communicates that the
+_packaging_ is what's changing. The code itself does not become any less mature
+(it's today's existing providers, repackaged).
+
+Once we have confidence in the service scoped provider pattern we would mark the
+too-large providers as deprecated and set a date at which they would no longer
+be built and published.
+
+Migration from (for example) `upbound/provider-aws` to `upbound/provider-aws-*`
+should not be significantly more complicated or risky than migrating from one
+version of `upbound/provider-aws` to the next. This is because there would be no
+schema changes to the MRs within the provider.
+
+To upgrade a control plane running (for example) `upbound/provider-aws` with
+several existing MRs in-place, the process would be:
+
+1. Install the relevant new `upbound/provider-aws-*` provider packages, with
+   `revisionActivationPolicy: Manual`. This policy tells the providers not to
+   start their controllers until they are activated.
+2. Uninstall the `upbound/provider-aws` provider package.
+3. Set the `revisionActivationPolicy` of the new `upbound/provider-aws-*`
+   provider packages to `Automatic`, letting them start their controllers.
+
+Behind the scenes the Crossplane package manager:
+
+1. Adds owner references to the CRDs the new provider packages intend to own.
+2. Removes owner references from the CRDs the old provider package owned.
+
+As the old provider is uninstalled, the CRDs it owned will be garbage collected
+(i.e. deleted) if no new provider added an owner reference. If a new provider
+_did_ add an owner reference it will become the only owner of the CRD and mark
+itself as the controller reference. Each CRD can have only one controller
+reference, which represents the provider that is responsible for reconciling it.
+
 ## Alternatives Considered
 
 I considered the following alternatives before making this proposal.
--- a/design/design-doc-smaller-providers.md
+++ b/design/design-doc-smaller-providers.md
@@ -186,14 +186,46 @@ Based on a survey of ~40 community members, the average deployment would:
 * Install ~100 provider CRDs, down from over 900 today.
 * Run ~9 provider pods, up from ~2 today.
 
-A “multi-cloud Kubernetes” scenario in which a hypothetical Crossplane user
-installed support for EKS, AKS, and GKE along with the Helm and Kubernetes
-providers would install 10 providers and 348 CRDs. This scenario is interesting
-because it triggers the pathological case of needing to install the biggest API
-group from each of the three major providers - `ec2.aws.upbound.io` (94 CRDs),
-`compute.gcp.upbound.io` (88 CRDs), and `networking.azure.upbound.io` (100
-CRDs). Note that this scenario is still well below the approximately 500 CRD
-mark at which folks may start seeing issues.
+### Large Services
+
+One service API group from each of the three major providers is quite large:
+
+* `ec2.aws.upbound.io` (94 CRDs)
+* `compute.gcp.upbound.io` (88 CRDs)
+* `networking.azure.upbound.io` (100 CRDs)
+
+`datafactory.azure.upbound.io` is also quite large at 45 CRDs. All other
+services currently have 25 or less CRDs.
+
+These large services are problematic because they contain types that are very
+commonly used. For example `ec2.aws.upbound.io` contains the `VPC` and
+`SecurityGroup` types, which are dependencies of many other services. For
+example to deploy an EKS cluster you may want to create and reference a `VPC`
+and a `SecurityGroup` (or three). To do so, you would need to bring in
+`provider-aws-ec2` and thus install ~90 superfluous CRDs.
+
+I propose that in these cases where a service:
+
+* Is large, containing 25 or more CRDs
+* Is commonly used as a dependency of other services
+* Contains a few 'core' types (e.g. `VPC`) alongside many more obscure types
+
+We break the service up into two providers; "core" and "everything else". For
+example in the case of `ec2.aws.upbound.io` we would create:
+
+* `upbound/provider-aws-ec2-core` - Contains the ~10-20 most commonly used types
+  (e.g. `VPC`, `SecurityGroup`, `Subnet`, etc).
+* `upbound/provider-aws-ec2` - Contains the remaining types.
+
+These providers would "share" the `ec2.aws.upbound.io` API group. That is, the
+core provider would not be `ec2-core.aws.upbound.io`. The core provider would be
+a dependency of the full provider, such that:
+
+* Installing `upbound/provider-aws-ec2` installed support for all of EC2,
+  including the things in core (as a dependency).
+* Installing `upbound/provider-aws-ec2-core` installed support for just the
+  core types.
+
 
 ### Cross-Resource References
 
--- a/design/design-doc-smaller-providers.md
+++ b/design/design-doc-smaller-providers.md
@@ -382,8 +382,7 @@ Of these options the aggregation API server seems to be the most feasible. We
 found that an API request to create a non-existent type was rejected before a
 webhook gets a chance to load the CRD for the type on-demand. The proxy option
 would require all Kubernetes clients to talk to the proxy rather than talking to
-the API server directly. This may be feasible for API servers we control (e.g.
-MCPs), but is probably not feasible in the wild.
+the API server directly. This is probably not feasible in the wild.
 
 I believe all forms of lazy-loading to be ultimately infeasible due to the
 discovery API (see the links in [Background](#background) if you’re unfamiliar
--- a/design/design-doc-smaller-providers.md
+++ b/design/design-doc-smaller-providers.md
@@ -1,0 +1,423 @@
+# Break Up Large Providers by Service
+
+## Background
+
+The most popular and largest Crossplane providers install hundreds of CRDs.
+No-one actually wants to use all of them. Installing them brings a severe
+performance penalty - enough to make some hosted Kubernetes services such as GKE
+unresponsive.
+
+Here are the large providers we have today:
+
+https://marketplace.upbound.io/providers/upbound/provider-aws/v0.30.0 - 872
+https://marketplace.upbound.io/providers/upbound/provider-azure/v0.28.0 - 692
+https://marketplace.upbound.io/providers/upbound/provider-gcp/v0.28.0 - 325
+https://marketplace.upbound.io/providers/crossplane-contrib/provider-tencentcloud/v0.6.0 - 268
+https://marketplace.upbound.io/providers/crossplane-contrib/provider-aws/v0.37.1 - 178
+https://marketplace.upbound.io/providers/frangipaneteam/provider-flexibleengine/v0.4.0 - 169
+https://marketplace.upbound.io/providers/scaleway/provider-scaleway/v0.1.0 - 73
+
+Based on a brief survey of the community and Marketplace, no other provider
+installs more than ~30 CRDs.
+
+Quantifying the performance penalty of installing too many CRDs is hard, because
+it depends on many factors. These include the size of the Kubernetes control
+plane nodes and the version of Kubernetes being run. As recently as March 2023
+we’ve seen that installing just provider-aws on a new GKE cluster can bring the
+control plane offline for an hour while it scales to handle the load.
+
+Problems tend to start happening once you’re over - very approximately - 500
+CRDs. Keep in mind that many folks are installing Kubernetes alongside other
+tools that use CRDs. Here’s a brief survey of the number of CRDs common tools
+install:
+
+* Crossplane (without any providers) - 12
+* Google Config Connector - ~200
+* Azure Service Operator - ~190
+* Kyverno - 12
+* Istio - 14
+* Argo Workflows - 8
+
+Installing too many CRDs also affects the performance of Kubernetes clients like
+kubectl, Helm, and ArgoCD. These clients are often built under the assumption
+that there won’t be more than 50-100 CRDs installed on a cluster.
+
+You can read more about the issues Kubernetes faces when too many CRDs are
+present at:
+
+* https://github.com/crossplane/crossplane/blob/v1.11.2/design/one-pager-crd-scaling.md
+* https://blog.upbound.io/scaling-kubernetes-to-thousands-of-crds/
+
+Apart from performance issues, some folks have expressed concerns about the
+security and general “bloat” of having many unused CRDs installed. I believe the
+security concerns to be an educational issue - I haven't yet seen a convincing
+argument for removing CRDs as opposed to using RBAC.
+
+Crossplane maintainers have invested in improving the performance of the
+Kubernetes control plane and clients in the face of many CRDs. These
+improvements have been insufficient to alleviate community pain. The assumption
+that there won’t be “too many” CRDs is pervasive in the Kubernetes codebase.
+There’s no one easy fix, and the conservative speed at which cloud providers and
+their customers pick up new Kubernetes releases means it can take years for
+fixes to be deployed.
+
+Allowing folks to lower the ratio of installed-to-used CRDs has been proposed
+for some time. The Crossplane maintainers are now aligned that it’s necessary.
+We have explored the following options:
+
+* Make it possible to filter unnecessary CRDs by configuring a provider.
+* Break the biggest providers up into smaller, service-scoped providers.
+* Load CRDs into the API server lazily, on-demand.
+
+## Proposal
+
+I propose that provider maintainers break the largest providers into smaller,
+service-scoped providers. Each service-scoped provider would correspond to what
+is currently an API group within a provider. For example,
+`upbound/provider-aws-rds` would extend Crossplane with support for the 22
+managed resources (and thus 22 CRDs) in the `rds.aws.upbound.io` API group.
+
+### Overview
+
+I proposed we only break up the largest providers. For example:
+
+* `upbound/provider-aws` - Becomes ~150 smaller providers.
+* `upbound/provider-azure` - Becomes ~100 smaller providers.
+* `upbound/provider-gcp` - Becomes ~75 smaller providers.
+* `crossplane-contrib/provider-tencentcloud` - Becomes ~50 smaller providers.
+
+It's important to note that we don't expect anyone to be installing and
+operating all of these providers. Per [Community Impact][#community-impact], we
+expect folks to install ~10 providers on average.
+
+These providers would:
+
+* Be built from “monorepos” - the same repositories they are today, like
+  upbound/provider-aws.
+* Share a `ProviderConfig` by taking a dependency on a (e.g.)
+  `provider-aws-config` package.
+* Continue to use contemporary “compiled-in” cross-resource-references, short to
+  medium term.
+
+Provider generation tooling like [`upjet`][upjet] would be updated to support
+generating service-scoped providers.
+
+
+No changes would be required to Crossplane core - i.e. to the package manager.
+No changes would be required to the long tail of existing, smaller Crossplane
+providers such as `provider-terraform` and `provider-kubernetes`.
+
+The following are the high level pros and cons of the proposed approach.
+
+Pros
+
+* Fewer CRDs is the default user experience - no new concepts or configuration
+  required.
+* Requires no changes to Crossplane core - no waiting months for core changes to reach
+  beta and be on by default.
+* Provider work spread over many provider processes (i.e. pods) should scale
+  better.
+* Folks can create their own “meta-providers” (e.g. `acmeco/provider-aws`) that
+  pull in just the services they want. This is just packaging - no compiling
+  required.
+* Granular deployments. You could upgrade the EKS provider without having to
+  upgrade RDS. 
+
+Cons
+
+* The average deployment may require more compute resources, short term.
+* Three widely used service groups (including EC2) are still quite large (~100
+  CRDs).
+* Some folks are going to feel it's not enough granularity.
+
+
+### Community Impact
+
+Based on a survey of ~40 community members, the average deployment would:
+
+* Install ~100 provider CRDs, down from over 900 today.
+* Run ~9 provider pods, up from ~2 today.
+
+A “multi-cloud Kubernetes” scenario in which a hypothetical Crossplane user
+installed support for EKS, AKS, and GKE along with the Helm and Kubernetes
+providers would install 10 providers and 348 CRDs. This scenario is interesting
+because it triggers the pathological case of needing to install the biggest API
+group from each of the three major providers - `ec2.aws.upbound.io` (94 CRDs),
+`compute.gcp.upbound.io` (88 CRDs), and `networking.azure.upbound.io` (100
+CRDs). Note that this scenario is still well below the approximately 500 CRD
+mark at which folks may start seeing issues.
+
+### Cross-Resource References
+
+Some MRs can reference other MRs; for example an EKS cluster MR can reference a
+VPC MR to indicate that it should be connected to that VPC. We call these
+cross-resource references.
+
+Today cross-resource references are implemented at the code level. That is, the
+EKS cluster MR controller needs to be compiled with knowledge of the VPC MR type
+and how to reference it.
+
+Even outside of this proposal this is limiting; it’s not possible today to make
+a reference from one provider to another. Doing so would require all providers
+to be compiled against all other providers. This means for example that a
+provider-helm Release MR cannot today reference a provider-aws EKS cluster MR.
+
+In the medium-to-long term I believe contemporary cross-resource references
+should be deprecated in favor of a generic alternative that supports references
+across providers. This is tracked in Crossplane issue [#1770][issue-1770].
+
+In the short term, breaking up the biggest providers would cause many existing
+cross-resource references to be from one provider to another. For example today
+an EKS-cluster-to-VPC reference is “internal” to a single provider
+(provider-aws). If providers were broken up by API group this reference would
+cross provider boundaries; from provider-aws-eks to provider-aws-ec2.
+
+I’m not concerned about this as a short term problem. As long as all AWS
+providers (for example) are built from a monorepo it should be possible to
+compile them with support for the types they reference. This does in theory
+create a coupling between AWS providers, but I don’t expect this to be an issue.
+
+Consider for example an EKS cluster that references a VPC by its external name
+annotation. Then assume we discarded external names as a concept. Both the EKS
+and EC2 providers would need to be upgraded to support this change - there’s a
+coupling. In practice cross-resource reference implementations rarely change
+once implemented, so I think this problem is largely theoretical.
+
+### Provider Configs
+
+All Crossplane providers install CRDs for a few special custom resources (CRs):
+
+* ProviderConfig - Configures the provider. Typically with authentication
+  credentials.
+* ProviderConfigUsage - Tracks usage of a ProviderConfig. Prevents it being
+  deleted.
+* StoreConfig - Configures how to access a Secret Store.
+
+In the [upstream Crossplane issue][issue-3754] that tracks this proposal a key
+theme was that community members don’t want to manage more ProviderConfigs than
+they do today. A community member running 10 smaller AWS providers, each with
+MRs spanning 10 AWS accounts would go from 10 ProviderConfigs to 10 x 10 = 100
+ProviderConfigs.
+
+ProviderConfigs are typically created by humans, or as part of a Composition. MR
+controllers create a ProviderConfigUsage for each MR. A ProviderConfig
+controller places a finalizer on each ProviderConfig to prevent it from being
+deleted while ProviderConfigUsages reference it.
+
+I propose that as part of breaking up the biggest providers, each provider
+"family" include a special “config” provider - e.g. `provider-aws-config`, or
+`provider-azure-config`. This config provider would:
+
+* Install the ProviderConfig, ProviderConfigUsage, and StoreConfig CRDs.
+* Start a controller to prevent the ProviderConfig being deleted while
+  ProviderConfigUsages referenced it.
+
+All other providers in the provider family would take a package manager
+dependency on the relevant config provider. This:
+
+* Ensures the config provider is installed automatically.
+* Reduces coupling - dependent providers can depend on a semantic version range.
+
+Consider the following scenarios:
+
+* A ProviderConfig is updated with a new field.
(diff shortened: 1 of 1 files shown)


design/design-doc-smaller-providers.md
```diff
@@ -0,0 +1,425 @@
...
+
+I proposed we only break up the largest providers. For example:
+
+* `upbound/provider-aws` - Becomes ~150 smaller providers.
+* `upbound/provider-azure` - Becomes ~100 smaller providers.
+* `upbound/provider-gcp` - Becomes ~75 smaller providers.
+* `crossplane-contrib/provider-tencentcloud` - Becomes ~50 smaller providers.
+
+It's important to note that we don't expect anyone to be installing and
+operating all of these providers. Per [Community Impact][#community-impact], we
```
jeanduplessis-alt, line 90:
```suggestion
operating all of these providers. Per [Community Impact](#community-impact), we
```

design/design-doc-smaller-providers.md
```diff
@@ -0,0 +1,425 @@
...
+For upjet-based providers it's a different story. Currently upjet-based
+providers use a lot of CPU and memory. There’s an ongoing effort to address
+this, but we suspect we can get providers down to _at least_:
+
+* ~600MB memory for the Crossplane provider process.
+* ~50MB per Terraform CLI invocation (we run 2-3 of these per reconcile).
+* ~300MB per Terraform provider process.
+
+The current thinking is that we need to run one Terraform provider process for
+each ProviderConfig` that is used by at least one MR. So for example one upjet
```
jeanduplessis-alt, line 260:
```suggestion
each ProviderConfig that is used by at least one MR. So for example one upjet
```

jeanduplessis-alt: Unless you mean it to be "each `ProviderConfig` that is used by at least "

design/design-doc-smaller-providers.md
```diff
@@ -0,0 +1,425 @@
...
+
+Of these options the aggregation API server seems to be the most feasible. We
+found that an API request to create a non-existent type was rejected before a
+webhook gets a chance to load the CRD for the type on-demand. The proxy option
+would require all Kubernetes clients to talk to the proxy rather than talking to
+the API server directly. This may be feasible for API servers we control (e.g.
+MCPs), but is probably not feasible in the wild.
+
+I believe all forms of lazy-loading to be ultimately infeasible due to the
+discovery API (see the links in [Background][#background] if you’re unfamiliar
```
jeanduplessis-alt, line 385:
```suggestion
discovery API (see the links in [Background](#background) if you’re unfamiliar
```

design/design-doc-smaller-providers.md
```diff
@@ -0,0 +1,429 @@
+# Break Up Large Providers by Service
+
+* Owner: Nic Cope (@negz)
+* Reviewers: Hasan Türken (@turkenh), Alper Ulucinar (@ulucinar)
```
negz, line 4: My rationale here is:

* Hasan is a Crossplane maintainer (who has a lot of experience working on providers)
* Alper is an `upjet` maintainer, and also maintains the three biggest Crossplane providers.

design/design-doc-smaller-providers.md
```diff
@@ -0,0 +1,429 @@
...
+`upbound/provider-aws-rds` would extend Crossplane with support for the 22
+managed resources (and thus 22 CRDs) in the `rds.aws.upbound.io` API group.
+
+### Overview
+
+I proposed we only break up the largest providers. For example:
+
+* `upbound/provider-aws` - Becomes ~150 smaller providers.
+* `upbound/provider-azure` - Becomes ~100 smaller providers.
+* `upbound/provider-gcp` - Becomes ~75 smaller providers.
```
negz, line 90: It's outside the scope of this design, but for the record we have commitment from folks at Upbound to update:

* Any large official providers we maintain (i.e. AWS, GCP, and Azure).
* The `upjet` tooling.
* The Upbound marketplace

blakebarnett:
I'm happy to see a solution for the short-term. It seems like filtering is the ideal solution to shoot for, but one for the long-term (2.0?). It'd be great if it could be targeted for later though. For a multi-cloud or operator-heavy cluster I could see them hitting 500 CRDs pretty frequently, and I assume the number of supported resources per-provider is only going to continue growing in a lot of cases.

negz:
> I'm happy to see a solution for the short-term. It seems like filtering is the ideal solution to shoot for, but one for the long-term (2.0?)

Just to be explicit, filtering is not something I propose we support long term.

> For a multi-cloud or operator-heavy cluster I could see them hitting 500 CRDs pretty frequently, and I assume the number of supported resources per-provider is only going to continue growing in a lot of cases.

Based on my projections and surveys of the community it doesn't seem that likely folks would hit 500 CRDs. FWIW the 500 number is also quite conservative (i.e. it's not like you hit 500 CRDs and everything starts to degrade immediately).

blakebarnett:
> Just to be explicit, filtering is not something I propose we support long term. Or rather, while I don't want to rule _anything_ out long term I'm not proposing we break providers up as a stop-gap proposal. I'm proposing it because I think it's the best option.

Understood. And I agree it's the most pragmatic solution, but we'll have to see how painful the break-up turns out to be later on. Perhaps a way to make it more user-friendly would be to include a "recommended" meta-providerconfig with the most commonly used set of providers per cloud-provider? Or maybe some other way for a user to understand which of the provider configs are going to be required for a given configuration? I'm imagining it being pretty confusing for someone new to crossplane.

negz:
> Perhaps a way to make it more user-friendly would be to include a "recommended" meta-providerconfig with the most commonly used set of providers per cloud-provider?

Yeah, we considered something like that. I think the tricky part would be landing on what the most commonly used bits are (given a broad audience) and evolving that over time without it getting too bloated.

One similar thing I am really excited about though is the fact that each org or team could put together their own meta-provider that has just the services _they_ use. I could imagine starting small with 1-2 services and adding more over time as you learn what services you need. I'm pretty sure that if you're just doing this for your own org you'll almost certainly end up with more tightly scoped providers (i.e. fewer CRDs) than if we as a community tried to curate "one-size-fits-most" providers.

haarchri:
if we take the `platform-ref-aws` from https://github.com/upbound/platform-ref-aws and apply this proposal to it we have the following crds in the cluster:

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
</head>

<body link="#0563C1" vlink="#954F72">



  | provider-split
-- | --
crossplane | 13
eks-composition | 2
network-compositon | 2
service-composition | 2
app-composition | 2
provider-helm | 3
provider-aws-ec2 | 98
provider-aws-iam | 22
provider-aws-eks | 6
  | 150



</body>

</html>


so we see that **126** crds are directly installed from the 3 splitted providers (ec2, iam, eks) - but we only use **15** of those crds -

if we would use a crd filter mechanism like @blakebarnett mentioned we would end up with:

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">

</head>

<body link="#0563C1" vlink="#954F72">



  | crd-filter
-- | --
crossplane | 13
eks-composition | 2
network-compositon | 2
service-composition | 2
app-composition | 2
provider-helm | 3
provider-aws | 15
  | 39



</body>

</html>

so you can see by splitting the providers we would install **111** CRDs that are not needed - so clearly this simple `platform-ref-aws` hits the following assumption that there won't be more than 50-100 CRDs in a cluster:

> Installing too many CRDs also affects the performance of Kubernetes clients like
> kubectl, Helm, and ArgoCD. These clients are often built under the assumption
> that there won’t be more than 50-100 CRDs installed on a cluster.


so lets have a look in our companies crossplane installation  we would have the following aws-providers installed after the split:

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">

</head>

<body link="#0563C1" vlink="#954F72">



providers | crds
-- | --
provider-aws-acm | 2
provider-aws-acmpca | 5
provider-aws-appmesh | 7
provider-aws-cloudwatchevents | 8
provider-aws-cloudwatchlogs | 8
provider-aws-directconnect | 16
provider-aws-docdb | 7
provider-aws-dynamodb | 7
provider-aws-ec2 | 98
provider-aws-efs | 6
provider-aws-eks | 6
provider-aws-opensearch | 3
provider-aws-elb | 9
provider-aws-elbv2 | 5
provider-aws-grafana | 5
provider-aws-iam | 22
provider-aws-kafka | 2
provider-aws-kms | 7
provider-aws-networkfirewall | 4
provider-aws-rds | 22
provider-aws-route53 | 9
provider-aws-route53resolver | 3
provider-aws-s3 | 23
proider-aws-ses | 13
provider-aws-sesv2 | 6
sum | 303



</body>

</html>

so we see that **303** crds are directly installed from the 25 splitted providers (...) - but we only use **97** of those crds - so you can see by splitting the providers we would install **206** CRDs that are not needed 

so at the moment i don't really see that the effort to split the providers will help to decrease the number of installed CRDs in a real world scenario this much ?! - also the effort needed by the operation teams to manage more (in our case 25) providers only to manage aws resources and the ability of crossplane to run smootly with this number of providers is not even part of this discussion - is someone from our community folks running crossplane installations with more then 10 or 20 providers ?

negz:
@haarchri In the numbers you've crunched it seems to me that we're seeing quite a big improvement in the ratio of installed-to-used CRDs. This is also what I saw when I ran through similar scenarios. Today for example the official provider-aws has 900 CRDs, so:

* For `platform-ref-aws` we go from 900:15 to 126:15.
* For your company we go from 900:97 to 303:97.

These numbers seem well within the capabilities of Kubernetes. I would not expect to be seeing the API server struggle with excess compute resource consumption, and I would not expect `kubectl` to experience discovery or CRD category query issues.

Is the only acceptable result as far as you're concerned to completely optimize the ratio of installed-to-used CRDs? That is, to make sure Crossplane never installs a CRD that you don't intend to use? If so, I would ask why? Also, do you use other CRD-based tools? Do those tools allow you to leave out the CRDs you don't need? What about built-in Kubernetes types?

I think based on https://github.com/crossplane/crossplane/issues/2869 you've previously answered my first question:

> we have the same requirement from security department to not deploy all CRDs - only approved CRDs - which matches our IAM Policies

I'd love to hear more about it. Perhaps we should setup a call? My guess is you have fairly extreme defense-in-depth requirements? My thinking is:

* Even when CRDs are installed Crossplane must still be granted IAM permissions at the cloud level, and API server users must still be granted RBAC access to work with the relevant CRs. This seems like quite a lot of defense alone.
* Even when you don't install the CRDs, the underlying code is still there in the provider. Sure you can't trigger it, so the attack surface is smaller, but doesn't RBAC also achieve this?

At the end of the day if it's a hard requirement that Crossplane _never_ installs a CRD that you don't want then yes the only tenable option is down-to-the-type level filtering (or compiling your own providers as you currently do). I wonder how many folks have this requirement. I believe "just adding filtering" is not as simple as it first sounds to most folks and in fact increases cognitive complexity for folks learning Crossplane quite a lot, so I think there should be a pretty high bar for doing it.

> also the effort needed by the operation teams to manage more (in our case 25) providers only to manage aws resources and the ability of crossplane to run smootly with this number of providers is not even part of this discussion

I had intended to cover this in the design - see for example discussion of ProviderConfigs, compute resources, etc. Are there other areas you'd like me to cover? If so, let me know.

negz:
> Only this approach would allow a decrease of the number installed crds by one order.

@akesser Can you help me understand why you need the number of installed CRDs to be so dramatically reduced?

> In short term the filtering could happen in crossplane-core by not applying crd as configured by the user when installing a provider.

It is unfortunately not that simple. I could expand on this in the existing "Alternatives Considered" section if that would help.

> The proposed splitting sound like well designed for clusters one shows in live demos of talks with perhaps three resources and therefore threes (sub-) providers installed. In a real world cluster where the goal is to really move to IaC, this only adds the burden to manage more providers and to keep them in sync.

The design doc includes a survey of ~40 community members and their real usage - I would not be proposing something if I didn't think it would work in the real-world. It would be helpful if you could elaborate on what concerns you have around management burden.

akesser:
@negz I talked about decreasing the numbers of installed CRDs because you explicitly mentioned that most tools are designed to work well with clusters containing 50 to 100 CRDs and the suggested approach to split the providers does not decrease the numbers of CRDs in a way to reach this. 

> stalling too many CRDs also affects the performance of Kubernetes clients like
 kubectl, Helm, and ArgoCD. These clients are often built under the assumption
 that there won’t be more than 50-100 CRDs installed on a cluster.

By having ten to hundred times more providers you have to keep them updated, you have to keep them running and you need more effort until you have the code on the cluster, e.g. by four eye principle for merging into your codebase

negz:
> By having ten to hundred times more providers you have to keep them updated, you have to keep them running and you need more effort until you have the code on the cluster, e.g. by four eye principle for merging into your codebase

FWIW by my estimates a hundred times more providers would be quite a rare edge case.

Isn't keeping pods running Kubernetes's bread and butter? I'm not arguing it's _zero_ additional operational burden, but I'm not sure its meaningful.

nabuskey:
Putting technical details aside, how does the community feel about the risk? What if this approach does not ultimately solve the problem for one reason or another? If Kubernetes improves its CRD handling drastically and can scale CRDs to 100000, do these providers need to be kept broken down? 

Breaking up providers or going back to the single provider model in isolation is fine. However, taking recent changes to providers into consideration, I can't say with confidence it won't be a problem. For context, recent changes I am referring to are: 
1. Introduction of jet providers.
2. Deprecation of jet providers.
3. Introduction of upjet. 
4. Introduction of providers with same names. 

In my opinion these changes caused fragmentation and confusion in reference docs, blog posts, and examples. I understand providers themselves do not define what Crossplane is, but I worry what kind of impression people get about Crossplane. I'd love to hear others' thoughts here.

negz:
> What if this approach does not ultimately solve the problem for one reason or another?

As far as I'm aware there's only two feasible solutions to the problem at hand - breaking up providers or adding support for filtering the types providers install. In either of those paths we should consider a way to ship the feature/fix and get feedback before we fully commit.

With filtering it would be the typical alpha feature lifecycle we use for Crossplane (complicated by the fact that the functionality needs to be added at both the core and provider level). So the feature would be behind a feature flag and off by default for some time to allow folks to try it out before we commit. If we found it didn't work well anyone using it would be disrupted when we removed it or changed in a breaking way.

If we instead break up providers I could imagine publishing some 'alpha' service-scoped providers to let folks try them out 
and give us feedback before we fully committed to the approach. If we found it didn't work well anyone using it would be disrupted when we stopped building the service-scoped providers and they needed to switch back to the monoliths.

I suspect given how many folks are impacted by the problem this proposal solves that a lot of folks will opt-in to the alpha (whether it be filtering or smaller providers) and thus a lot of folks will be disrupted if we found we didn't get it right up-front. It doesn't seem to me like either option is significantly less disruptive to "roll back".

This topic is probably worth a section in the proposal - I'll add one.

> If Kubernetes improves its CRD handling drastically and can scale CRDs to 100000, do these providers need to be kept broken down?

This is an interesting thought experiment. I will say that given the multi-faceted nature of the issues and the slow, incremental pace of improvements I don't expect to see this be the case for _at least_ 24 months. Keep in mind that we have pursued the "fix Kubernetes" approach for well over a year with limited success. I don't see any reason to believe things will change in the foreseeable future.

akesser:
Our Team created two PRs to show filtering of CRDs in crossplane-core and crossplane-contrib/provider-aws:

https://github.com/crossplane/crossplane/pull/3987
https://github.com/crossplane-contrib/provider-aws/pull/1727

Feedback is wellcome

design/design-doc-smaller-providers.md
```diff
@@ -0,0 +1,429 @@
...
+
+* Fewer CRDs is the default user experience - no new concepts or configuration
+  required.
+* Requires no changes to Crossplane core - no waiting months for core changes to reach
+  beta and be on by default.
+* Provider work spread over many provider processes (i.e. pods) should scale
+  better.
+* Folks can create their own “meta-providers” (e.g. `acmeco/provider-aws`) that
+  pull in just the services they want. This is just packaging - no compiling
+  required.
```
jaypipes, line 126: +1 In ACK land we are working on something similar [here](https://github.com/aws-controllers-k8s/community/issues/1360). Of course, we're coming to this from the opposite angle :) We started life with lots of individual service-specific controllers and got user feedback that installing lots of controllers was annoying!

negz: > got user feedback that installing lots of controllers was annoying

Could you link some of this feedback?

I'm wondering whether Crossplane's package manager might make it a little less onerous in our case.

negz: I went looking - here's some feedback https://github.com/aws-controllers-k8s/community/issues/1534.

design/design-doc-smaller-providers.md
```diff
@@ -0,0 +1,429 @@
...
+expect folks to install ~10 providers on average.
+
+These providers would:
+
+* Be built from “monorepos” - the same repositories they are today, like
+  upbound/provider-aws.
+* Share a `ProviderConfig` by taking a dependency on a (e.g.)
+  `provider-aws-config` package.
+* Continue to use contemporary “compiled-in” cross-resource-references, short to
+  medium term.
```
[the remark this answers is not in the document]
negz: Kind of.

The thinking is that we'd build them all together from the same monorepo, and release them all at once with the same version. So `provider-aws-eks:v0.40.0` and `provider-aws-ec2:v0.40.0` would be released at the same time.

In practice you could almost certainly run `provider-aws-eks:v0.42.0` with `provider-ec2:v0.40.0` without any issue. This will only become more true as we get rid of compiled-in cross-resource-references (see the Cross Resource References heading for more).

That said, running your set of providers at the same version would be the most conservative approach. We probably wouldn't integration test all combinations of provider versions for example - we'd just run our integration tests (that often test across service boundaries) with a set of providers at the same version.

design/design-doc-smaller-providers.md
```diff
@@ -0,0 +1,429 @@
...
+  pull in just the services they want. This is just packaging - no compiling
+  required.
+* Granular deployments. You could upgrade the EKS provider without having to
+  upgrade RDS. 
+
+Cons
+
+* The average deployment may require more compute resources, short term.
+* At least three widely used service groups (including EC2) are still quite
+  large (~100 CRDs).
```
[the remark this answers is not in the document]
negz: Agreed. I'd be happy to break these down a bit more if folks thought there was value doing so. It's not clear whether they all break down super cleanly (e.g. the Azure group is just "networking") so maybe (e.g. `provider-aws-ec2-core` and `provider-aws-ec2` or something like you propose would make sense.

I think the key think we need to consider in making this decision is how impactful we think the size of these frequently used, large services are. If you're using a single large cloud provider with Crossplane, installing EC2 support would consume ~100 CRDs, giving you ~400 CRDs worth of other providers to install before you risked any performance issues. If you're using all three large cloud providers your buffer drops down to about ~200 CRDs before you risk performance issues. I do wonder what scenarios would push folks up to around 500 CRDs installed. I can't personally think of any that I think we'd see commonly.

Another thing to consider is that I am assuming the ~500 CRD ceiling will grow slowly over time as more folks update to newer versions of Kubernetes, and newer versions are released with improved performance.

design/design-doc-smaller-providers.md
```diff
@@ -0,0 +1,429 @@
...
+  required.
+* Requires no changes to Crossplane core - no waiting months for core changes to reach
+  beta and be on by default.
+* Provider work spread over many provider processes (i.e. pods) should scale
+  better.
+* Folks can create their own “meta-providers” (e.g. `acmeco/provider-aws`) that
+  pull in just the services they want. This is just packaging - no compiling
+  required.
+* Granular deployments. You could upgrade the EKS provider without having to
+  upgrade RDS.
```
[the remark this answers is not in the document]
negz: Can you elaborate on the maintenance version? I agree that it's non-zero, but wouldn't it be mostly bumping a few more semantic versions in your configurations? Is there more that I'm missing?

design/design-doc-smaller-providers.md
```diff
@@ -0,0 +1,429 @@
...
+The gist of this proposal is to update the existing Provider.pkg.crossplane.io
+API to include a configurable allow list of enabled CRDs. As far as I can tell
+this is the only feasible alternative to breaking up the largest Providers.
+
+If we took this approach I would propose that:
+
+* We teach the Package Manager that only some Providers support filtering. Such
+  Providers would have something like `pkg.crossplane.io/filtering: enabled` in
+  their `crossplane.yaml`. This is necessary to avoid filtering being silently
+  ineffectual when not supported by a Provider.
```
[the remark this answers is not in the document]
negz: I don't love it. I could agree that it's "good enough", but it doesn't seem like an ideal UX that if you configure your `Provider.pkg` in a certain way (that has nothing obvious to do with CLI flags) you risk your provider failing to install with an "unknown CLI flag" error.

design/design-doc-smaller-providers.md
```diff
@@ -0,0 +1,429 @@
...
+  replicated to the ProviderRevision.pkg type (that actually installs/owns a
+  Provider’s CRDs).
+* If you installed a filtering-enabled Provider into an empty control plane and
+  didn’t specify what types to enable, no types would be enabled.
+* If you installed a filtering-enabled Provider into a control plane the Package
+  Manager would automatically detect what types were in use and enable them
+  automatically.
+* If you installed a filtering-enabled Provider as a dependency of a
+  Configuration the Package Manager would scan its Compositions for types and
+  enable them automatically.
```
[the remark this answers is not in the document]
negz: We know today that installing a single, sufficiently large provider will degrade some control planes, right? For example we know installing a recent version of `upbound/provider-aws` on a new GKE control plane causes it to become unresponsive. In my mind this makes installing all CRDs by default a non-starter.

design/design-doc-smaller-providers.md
```diff
@@ -0,0 +1,429 @@
...
+
+This allows us to have an “opt-in” to the types you want approach, without
+uninstalling the types you’re already using when you upgrade from a Provider
+that doesn’t support filtering to one that does. This is preferable to an
+"opt-out" approach, which I believe to be untenable. The default UX for a large
+provider cannot continue to be installing all of its CRDs: this would result in
+a bad user experience (i.e. broken clusters, slow tools) for anyone who did not
+know they needed to filter.
```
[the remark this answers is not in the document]
negz: "It breaks unless you've read the documentation and know you need to enable an optional feature" seems like a really bad user experience to me. I know there is a lot of demand for filtering, but I really struggle with this.

design/design-doc-smaller-providers.md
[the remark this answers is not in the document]
phisco: I assume someone would not install a provider in a prod environment at first, so I agree it's not the best UX "breaking" the cluster at first, but having it clearly documented and maybe explicitly allowing all groups in the Provider's installation manifests to make it even more clear that a lot of CRDs are going to be installed, could make it acceptable IMO.

design/design-doc-smaller-providers.md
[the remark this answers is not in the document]
turkenh: I think it is more like a cognitive load, previously there was 1 provider that you need to keep up to date, but now there are ~10 providers. In theory, yes, you can always bump all together just like you bump 1 but in practice, especially due to the granularity you mentioned as _you could upgrade the EKS provider without having to RDS_, users may end up ~10 providers running at different versions and needs to be bumped one by one.
It is typical that people avoid updating versions of software, especially in production, unless they need a fix or feature which could contribute to such a version skew over time.

design/design-doc-smaller-providers.md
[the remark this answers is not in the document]
turkenh: > you risk your provider failing to install with an "unknown CLI flag" error.

Given this is a recoverable error with no harm by removing `enabledAPIs` from `Provider.pkg`, I don't see this as a big risk.

I don't love it too, but given that it is some transitive behavior until providers support filtering I still believe we could live with it.

design/design-doc-smaller-providers.md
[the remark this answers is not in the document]
turkenh: Agree that it is not an ideal solution, but given that there is no perfect solution, I would like us to reevaluate different alternatives here.

What about something in between? Instead of installing all or none by default, can we install a default set of CRDs defined by the providers themselves? I imagine each provider defines a set of APIs enabled by default in their `crossplane.yaml`'s, and it would be used to _initialize_ the `enabledAPIs` in `provider.pkg` by package manager during installation. For upgrades, the package manager will also consider existing MRs and extend the list of enabled APIs. 

I mean; 

- Installing a provider on an empty control plane as today: Package manager checks if `enabledAPIs` configured, if not, it will initialize with ~100 CRDs defined by the provider package. 
- Upgrading provider aws: Package manager checks the existing CRs and merges them with default APIs to initialize `enabledAPIs`.

This will be more like a defaulting mechanism, and the package manager will not interfere with `enabledAPIs`, if configured already.

I believe solutions like this were considered as hairy in the past. Still, given there is no ideal solution, I would like us to investigate "how hairy it is" and whether we can find a relatively complex but working solution.

design/design-doc-smaller-providers.md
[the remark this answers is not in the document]
negz: I continue to like this idea (the simpler ec2/advanced option specifically) so I've added a heading to the document including it.

tobru:
After studying the proposal in detail, also checking the potential other options we have, I came to the conclusion that the proposed way of splitting big providers into smaller ones is the best way forward to solve the issue at hand. Thank you so much for the well-thought proposal :partying_face:

chobostar:
Thanks for you effort in splitting large providers!

I'm sure with internals of providers, but is there any risk of hitting rate limit or getting throttled by Cloud providers API if multiple "small" providers will poll API instead of 1 large?

negz:
> I'm sure with internals of providers, but is there any risk of hitting rate limit or getting throttled by Cloud providers API if multiple "small" providers will poll API instead of 1 large?

Good question. I don't think there's a large risk. At a high level it's the same set of control loops, just running spread over a few pods instead of all running in one.

That said, one thing that will change will be the "scope" of the `--max-reconcile-rate` and `--poll-interval` flags. These flags are the main things that influence how often a provider interacts with the external system (e.g. AWS). Currently these flags are set at the scope of the entire cloud provider, whereas under this proposal they would be set per service. So it may make sense to lower their values (especially for `--max-reconcile-rate`) relative to what they are today.

This increased granularity does have an upside though - for example you could configure some providers (i.e. cloud services) to poll more frequently and others to poll less frequently depending on your needs.

design/design-doc-smaller-providers.md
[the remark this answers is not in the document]
negz: > What about something in between?

I think this shakes out as about the same as defaulting to nothing, right? The key point is that whatever set the Provider defaults to might not be what you're running when you upgrade, so we have to build the "magic" that figures out what you _are_ using and implicitly enables it.

Doable, but it's more magic added to the system.

design/design-doc-smaller-providers.md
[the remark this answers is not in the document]
negz: I really think this is a non-starter. Going to close this thread.

design/design-doc-smaller-providers.md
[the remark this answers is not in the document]
negz: I touch on this now in the design. If we did go the filtering route I think we should do this - it shouldn't be that hard to do and it would provide a better UX.

design/design-doc-smaller-providers.md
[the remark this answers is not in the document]
negz: I've updated the doc to point out that granular deployments is both a pro and a con.

jeanduplessis (approved):
@negz Thanks for the well-thought-out proposal and the effort you put into understanding and clarifying the various associated challenges. Your proposal is the most pragmatic for the problem at hand. After receiving lots of feedback about this challenge from attendees at KubeCon Amsterdam while serving at the Crossplane booth, it's clear that a solution is desperately needed in the community.

shanproofpoint:
is crossplane able to intercept the point at which the api server reads the new yaml and checks its understanding of crd's ?  is there such a thing as install on 404? like make the api server talk to the provider if the api group is handled by the provider?

negz:
> Then we can install crds when it is first used vs at provider install.

@shanproofpoint We investigated that option, and unfortunately it doesn't seem feasible. It's discussed in the design - See "Lazy-loading CRDs" under "Alternatives Considered", at the end.

shanproofpoint:
> > Then we can install crds when it is first used vs at provider install.
> 
> @shanproofpoint We investigated that option, and unfortunately it doesn't seem feasible. It's discussed in the design - See "Lazy-loading CRDs" under "Alternatives Considered", at the end.

thank you @negz

design/design-doc-smaller-providers.md
```diff
@@ -0,0 +1,704 @@
...
+  `upbound/provider-aws`.
+* Share a `ProviderConfig` by taking a dependency on a (e.g.)
+  `provider-aws-config` package.
+* Continue to use contemporary “compiled-in” cross-resource-references, short to
+  medium term.
+
+Provider generation tooling like [`upjet`][upjet] would be updated to support
+generating service-scoped providers.
+
+No changes would be required to Crossplane core - i.e. to the package manager.
```
negz, line 131: I need to update this. We did find in testing that we needed a small change to the RBAC manager, which you can see in https://github.com/crossplane/crossplane/pull/4010.

design/design-doc-smaller-providers.md
```diff
@@ -0,0 +1,704 @@
...
+installs more than ~30 CRDs.
+
+Quantifying the performance penalty of installing too many CRDs is hard, because
+it depends on many factors. These include the size of the Kubernetes control
+plane nodes and the version of Kubernetes being run. As recently as March 2023
+we’ve seen that installing just `provider-aws` on a new GKE cluster can bring
+the control plane offline for an hour while it scales to handle the load.
+
+Problems tend to start happening once you’re over - very approximately - 500
+CRDs. Keep in mind that many folks are installing Kubernetes alongside other
```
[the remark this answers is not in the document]
negz: Good catch, thanks.

design/design-doc-smaller-providers.md
```diff
@@ -0,0 +1,704 @@
...
+I would argue that most folks already think this way. Folks don't think "I want
+AWS support" without having some idea which AWS _services_ they want to use.
+Confirming "Does it support AWS?" is a tiny speed bump before you confirm "Does
+it support RDS?" (or whatever services you need).
+
+### Community Impact
+
+Based on a survey of ~40 community members, the average deployment would:
+
+* Install ~100 provider CRDs, down from over 900 today.
```
turkenh, line 187: Is this number updated after the recently added [Large Services](https://github.com/crossplane/crossplane/pull/3939/files?diff=unified&w=0#diff-c35d366039b2ed54517b98b83bc01e4e421bae2ea9ed3d0ee53a9231afcaf8d3R193) section? If not, this would go further down.

design/design-doc-smaller-providers.md
```diff
@@ -0,0 +1,704 @@
...
+I propose that in these cases where a service:
+
+* Is large, containing 25 or more CRDs
+* Is commonly used as a dependency of other services
+* Contains a few 'core' types alongside many less commonly used types
+
+We break the service up into two providers; "core" and "everything else". For
+example in the case of `ec2.aws.upbound.io` we would create:
+
+* `upbound/provider-aws-ec2-core` - Contains the ~10-20 most commonly used types
```
turkenh, line 219: @ulucinar fyi. 
We should keep this in mind while implementing the tooling for code generation, as this would be an exception to the **group by service** logic.

design/design-doc-smaller-providers.md
```diff
@@ -0,0 +1,704 @@
...
+### Inter-Provider Dependencies
+
+As established in the [Provider Configs](#provider-configs) section, all
+providers within a provider family will depend on a single config provider. For
+example all `provider-aws-*` providers will have a package dependency on
+`provider-aws-config`.
+
+A common misconception is that providers will need to depend on the providers
+they may (cross-resource) reference. This is not the case. I expect that the
+_only_ dependency providers within a family will have is on their config
```
turkenh, line 333: Another exception seems to be recently added advanced to  `core`, e.g. `provider-aws-ec2` depends on `provider-aws-ec2-core`

turkenh (approved):
Thanks for all the details and well thought proposal @negz 🙌 

After all the recent details and discussions, I agree that breaking up large providers is the best way to go!

design/design-doc-smaller-providers.md
negz: I think it was actually miscalculated before. 😞  I went to update it and found that _with_ the new Large Services plan the average was ~100. I will double-check my calculations today.

design/design-doc-smaller-providers.md
negz: Just FYI, @bassam has some thoughts about this. He might suggest we hold off for now - I'm chatting to him today to learn more.

design/design-doc-smaller-providers.md
negz: Yes good catch. I'll update after I chat with @bassam to get his perspective on whether we should break up larger providers later today.

design/design-doc-smaller-providers.md
bassam: I think we should hold off on `Large Services` for now. Service scoped providers are natural grouping and I think any further breakdown starts to impact the user experience and is something we would only be doing for short term performance optimization.

design/design-doc-smaller-providers.md
bassam: this is another reason to hold of on Large Services for now. There is a natural isolation between two independent services, but if we break the large services further we end up with arbitrary packaging of core vs. non-core and the dependencies might not line up as cleanly.

design/design-doc-smaller-providers.md
negz: > I think we should hold off

I agree after our discussion yesterday. I'll move it to a "potential future work" section. Note that this will change the projected average number of CRDs based on our survey of community members - I think it will go from ~100 to ~180 or so.

design/design-doc-smaller-providers.md
negz: Done.

PR #3948, opened 2023-04-06
Update XR and MR if claim map/array fields are removed

### LINKED ISSUE #3421: Tag deleted from claim spec is added back
### What happened?

We have observed an issue with @stevendborrelli where a tag is deleted from a map in a claim spec parameter, and it gets added back to the claim. Please see below for an example XRD, claim and composition.

After you remove the tag key from the claim’s `spec.parameters.tags` map (and expect first the XR’s corresponding map to be updated, and then the MR’s `spec.forProvider.tags` map to be replaced because of the composition’s relevant patch), the removed key gets reinserted by the [`claim.APIClaimConfigurator`](https://github.com/crossplane/crossplane/blob/d5c522e9b29b2b1d050c9181ddfa4608a3725ebb/internal/controller/apiextensions/claim/reconciler.go#L490), which configures a claim using the associated XR. At this point, the tag key is missing (from `spec.parameters.tags`) in the claim but it still exists in the XR. The `APIClaimConfigurator` merges the spec of the XR into the claim’s spec using `mergo.Merge`. And  `mergo.Merge` puts the missing key in the claim’s `spec.parameters.tags` from the XR’s map (which still contains it) while merging these maps.

There is no race condition involved. The claim reconciler uses a `resource.APIPatchingApplicator` to patch the XR via a JSON-merge patch but this can not remove the key (map not replaced, there is now a missing key in a map, which has no effect). However, `resource.APIPatchingApplicator` [fetches the current spec of the XR](https://github.com/crossplane/crossplane-runtime/blob/fa5545f26b0458d4680502cfc0415048318baa85/pkg/resource/api.go#L136) and puts it into the same object that will later be used to [configure the claim](https://github.com/crossplane/crossplane/blob/d5c522e9b29b2b1d050c9181ddfa4608a3725ebb/internal/controller/apiextensions/claim/reconciler.go#L490) (via the `APIClaimConfigurator` as explained above).
Thus the claim acquires back the deleted key.

### How can we reproduce it?

Using the following XRD, claim and composition, please delete the `test` key from claim's (`AWSGlueTrigger`) `spec.parameters.tags`:

```yaml

apiVersion: dataservices.lmig.com/v1alpha1
kind: AWSGlueTrigger
metadata:
  namespace: upbound-system
  name: ateam-glue-test-trigger-instance1
  annotations:
    crossplane.io/external-name: ateam-glue-trigger-tags-test
spec:
  parameters:
    description: "test trigger for A-team test 2"
    workflowName: "a-team-workflow1"
    type: ON_DEMAND
    actions:
      - jobName: "ateam-glue-name1"
    tags:
      deployment_guid: "0000-9-9-99999"
      lm-troux_uid: "0000-9-9-99999"
      test: new_tag
  compositionSelector:
    matchLabels:
      purpose: glue-trigger
      provider: aws

---
kind: Composition
metadata:
  name: gds-glue-trigger-pattern
  labels:
    crossplane.io/xrd: xawsgluetriggers.dataservices.lmig.com
    purpose: glue-trigger
    provider: aws
spec:
  compositeTypeRef:
    apiVersion: dataservices.lmig.com/v1alpha1
    kind: XAWSGlueTrigger
  writeConnectionSecretsToNamespace: default
  resources:
  - name: glue-trigger
    base:
      apiVersion: glue.aws.upbound.io/v1beta1
      kind: Trigger
      spec:
        forProvider:
          region: us-east-1
          tags: {}
    patches:
      - fromFieldPath: "metadata.annotations[crossplane.io/external-name]"
        toFieldPath: "metadata.annotations[crossplane.io/external-name]"
        fromFieldPath: spec.parameters.description
        toFieldPath: spec.forProvider.description
        fromFieldPath: spec.parameters.workflowName
        toFieldPath: spec.forProvider.workflowName
        fromFieldPath: spec.parameters.type
        toFieldPath: spec.forProvider.type
        fromFieldPath: spec.parameters.actions
        toFieldPath: spec.forProvider.actions
        fromFieldPath: spec.parameters.predicate
        toFieldPath: spec.forProvider.predicate
        fromFieldPath: spec.parameters.startOnCreation
        toFieldPath: spec.forProvider.startOnCreation
        fromFieldPath: spec.parameters.tags
        toFieldPath: spec.forProvider.tags


---

metadata:
  name: xawsgluetriggers.dataservices.lmig.com
  labels:
    purpose: glue-trigger
    provider: aws
spec:
  defaultCompositionRef:
    name: gds-glue-trigger-pattern
  group: dataservices.lmig.com
  names:
    kind: XAWSGlueTrigger
    plural: xawsgluetriggers
  claimNames:
    kind: AWSGlueTrigger
    plural: awsgluetriggers
  versions:
    - name: v1alpha1
      served: true
      referenceable: true
      schema:
        openAPIV3Schema:
          type: object
          properties:
            spec:
              type: object
              properties:
                parameters:
                  type: object
                  properties:
                    description:
                      type: string
                    workflowName:
                      type: string
                    startOnCreation:
                      type: boolean
                    type:
                      type: string
                      enum: [CONDITIONAL, ON_DEMAND, SCHEDULED]
                    actions:
                      type: array
                      items:
                        type: object
                        properties:
                          jobName:
                            type: string
                          arguments:
                            type: object
                            additionalProperties:
                              type: string
                    predicate:
                      type: array
                      items:
                        type: object
                        properties:
                          logical:
                            type: string
                            enum: [AND, ANY]
                          conditions:
                            type: array
                            items:
                              type: object
                              properties:
                                operator:
                                  type: string
                                jobName:
                                  type: string
                                state:
                                  type: string
                                  enum: [SUCCEEDED, STOPPED, TIMEOUT, FAILED]
                        required:
                          - conditions
                    tags:
                      description: Tags map required for comliance
                      type: object
                      additionalProperties:
                          type: string
                  required:
                    - actions
                    - tags
                    - type
              required:
                - parameters
```

Crossplane version: `v1.10.0`

ezgidemirel:

Currently, we're applying XR and MR changes with the patch call. Since removing fields with patch is not possible, we're switching to update calls. 

Even though the PR seems small, there might be unwanted side effects. I tried to cover several scenarios with manual testing and there were no issues. 

Fixes #3421 

I have:

- Created a composition/xrd
- Applied a claim with a map
- Removed one of the map fields 
- Verified that claim, XR and MR no longer have that field. 

You can find relevant files [here](https://gist.github.com/ezgidemirel/0c14f53d8729eca2eb5fe1509bf3ddca). 

### DIFF
--- a/internal/controller/apiextensions/claim/configurator_test.go
+++ b/internal/controller/apiextensions/claim/configurator_test.go
@@ -372,6 +372,154 @@ func TestCompositeConfigure(t *testing.T) {
 				},
 			},
 		},
+		"RemoveMapField": {
+			reason: "A map field should be deleted if it is removed from the claim",
+			c: &test.MockClient{
+				MockCreate: test.NewMockCreateFn(nil),
+			},
+			args: args{
+				ctx: context.Background(),
+				cm: &claim.Unstructured{
+					Unstructured: unstructured.Unstructured{
+						Object: map[string]any{
+							"apiVersion": apiVersion,
+							"kind":       kind,
+							"metadata": map[string]any{
+								"namespace": ns,
+								"name":      name,
+							},
+							"spec": map[string]any{
+								"coolness": 23,
+								"stuff": map[string]any{
+									"foo": "bar",
+								},
+							},
+						},
+					},
+				},
+				cp: &composite.Unstructured{
+					Unstructured: unstructured.Unstructured{
+						Object: map[string]any{
+							"metadata": map[string]any{
+								"generateName": name + "-",
+								"labels": map[string]any{
+									xcrd.LabelKeyClaimNamespace: ns,
+									xcrd.LabelKeyClaimName:      name,
+								},
+							},
+							"spec": map[string]any{
+								"coolness": 42,
+								"stuff": map[string]any{
+									"foo": "bar",
+									"baz": "qux",
+								},
+							},
+						},
+					},
+				},
+			},
+			want: want{
+				cp: &composite.Unstructured{
+					Unstructured: unstructured.Unstructured{
+						Object: map[string]any{
+							"metadata": map[string]any{
+								"generateName": name + "-",
+								"labels": map[string]any{
+									xcrd.LabelKeyClaimNamespace: ns,
+									xcrd.LabelKeyClaimName:      name,
+								},
+							},
+							"spec": map[string]any{
+								"coolness": 23,
+								"stuff": map[string]any{
+									"foo": "bar",
+								},
+								"claimRef": map[string]any{
+									"apiVersion": apiVersion,
+									"kind":       kind,
+									"namespace":  ns,
+									"name":       name,
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+		"RemoveArrayField": {
+			reason: "An array entry field should be deleted if it is removed from the claim",
+			c: &test.MockClient{
+				MockCreate: test.NewMockCreateFn(nil),
+			},
+			args: args{
+				ctx: context.Background(),
+				cm: &claim.Unstructured{
+					Unstructured: unstructured.Unstructured{
+						Object: map[string]any{
+							"apiVersion": apiVersion,
+							"kind":       kind,
+							"metadata": map[string]any{
+								"namespace": ns,
+								"name":      name,
+							},
+							"spec": map[string]any{
+								"coolness": 23,
+								"stuff": []any{
+									"foo",
+								},
+							},
+						},
+					},
+				},
+				cp: &composite.Unstructured{
+					Unstructured: unstructured.Unstructured{
+						Object: map[string]any{
+							"metadata": map[string]any{
+								"generateName": name + "-",
+								"labels": map[string]any{
+									xcrd.LabelKeyClaimNamespace: ns,
+									xcrd.LabelKeyClaimName:      name,
+								},
+							},
+							"spec": map[string]any{
+								"coolness": 42,
+								"stuff": []any{
+									"foo",
+									"bar",
+								},
+							},
+						},
+					},
+				},
+			},
+			want: want{
+				cp: &composite.Unstructured{
+					Unstructured: unstructured.Unstructured{
+						Object: map[string]any{
+							"metadata": map[string]any{
+								"generateName": name + "-",
+								"labels": map[string]any{
+									xcrd.LabelKeyClaimNamespace: ns,
+									xcrd.LabelKeyClaimName:      name,
+								},
+							},
+							"spec": map[string]any{
+								"coolness": 23,
+								"stuff": []any{
+									"foo",
+								},
+								"claimRef": map[string]any{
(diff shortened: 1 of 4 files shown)


internal/controller/apiextensions/claim/configurator_test.go
```diff
@@ -372,6 +372,154 @@ func TestCompositeConfigure(t *testing.T) {
...
+									"name":       name,
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+		"RemoveArrayField": {
+			reason: "A map field should be deleted if it is removed from the claim",
```
jeanduplessis, line 450:
```suggestion
			reason: "An array entry should be deleted if it is removed from the claim",
```

ytsarev:
/fresh I believe this issue is still valid and affecting Crossplane users

stevendborrelli:
Given that Server Side Apply seems to be a big lift, should we look at this PR again to fix the immediate problem?

internal/controller/apiextensions/composite/composition_pt.go
```diff
@@ -143,7 +143,7 @@ func NewPTComposer(kube client.Client, o ...PTComposerOption) *PTComposer {
 	kube = unstructured.NewClient(kube)
 
 	c := &PTComposer{
-		client: resource.ClientApplicator{Client: kube, Applicator: resource.NewAPIPatchingApplicator(kube)},
+		client: resource.ClientApplicator{Client: kube, Applicator: resource.NewAPIUpdatingApplicator(kube)},
```
turkenh, line 146: I don't think we can simply switch to the updating applicator from the patching applicator without breaking some existing cases. I know, at least in the Usage implementation, I [rely on this fact](https://github.com/crossplane/crossplane/pull/4444/files#diff-539ca26eb860a34f63a58b7896e1287da46840b0a6f144b20c14cdf79ac71c38R330-R332) not to get labels put by the Usage controller deleted by composition controller. 

There could be other cases that rely on this fact that need to be handled properly, including the ones we know, e.g. what would happen to external-name annotation put on composed resources by providers, and we don't know, e.g. some external/custom controller adding some custom metadata on composed resources. In that sense, this could be considered as a behavioral breaking change.

turkenh (changes requested):
The approach for fixing **Claim to XR** cases in this PR looks good. However, we need to do more investigation/work on **XR to MR** case since I believe we cannot simply switch from the **patching applicator** to **updating applicator** and be ok as I commented in detail above.

turkenh:
> Given that Server Side Apply seems to be a big lift, should we look at this PR again to fix the immediate problem?

@stevendborrelli, this sounds reasonable. 
Just left a comment above on why I am concerned about this PR more specifically. I would love to hear @negz's thoughts on this as he is [refactoring the PTF composer](https://github.com/crossplane/crossplane/issues/4339).

jeanduplessis:
Closing this PR as it will be succeeded by the server-side apply work @pedjak is focusing on.

PR #3951, opened 2023-04-06
feat(chart): enable webhooks by default

### LINKED ISSUE #3748: Enable XRD validating webhook by default

Back in https://github.com/crossplane/crossplane/pull/2919 we added a bunch of infrastructure for webhooks. We also added our first one - a small validation hook for the XRD type. It's off by default - you enable it by passing `--set webhooks.enabled=true` to Helm.

Should we turn it on by default? I suspect most folks don't know its there (I certainly forgot it was) and won't benefit from it. I wonder if enough folks have tried it that we can consider it "out of alpha" and now a beta feature?

phisco:

Sets `webhooks.enabled` to `true` by default in the Crossplane Helm chart, this will mean that by default both CompositeResourceDefinitions and Compositions Webhooks are going to be validated.

Fixes #3748.

I have:

Locally deploying the chart with the new value.

### DIFF
--- a/cluster/charts/crossplane/values.yaml.tmpl
+++ b/cluster/charts/crossplane/values.yaml.tmpl
@@ -35,7 +35,7 @@ imagePullSecrets: {}
 registryCaBundleConfig: {}
 
 webhooks:
-  enabled: false
+  enabled: true
 
 rbacManager:
   deploy: true


jbw976 (approved):
Looks good to me now that https://github.com/crossplane/crossplane/pull/3921 has been merged 🙌

turkenh (approved):

PR #3955, opened 2023-04-08
Add EnvironmentConfig support to packages

### LINKED ISSUE #3865: Support EnvironmentConfig in crossplane packages

The recent addition of `EnvironmentConfig` custom resources allows passing parameters back and forth to one or more `Compositions`.  But this also creates a runtime dependency between Compositions and EnvironmentConfigs which needs to be resolved by some mechanism.

Crossplane packages is the standard mechanism to deliver Crossplane providers and configurations. It also supports dependency management. Therefore it would make sense to extend the crossplane [package format](https://github.com/crossplane/crossplane/blob/1aa83092172bdf0d2ed64754d33517c612ff7368/design/one-pager-package-format-v2.md) to support the packaging of one or more EnvironmentConfig:
- either a configuration package containing only EnvironmentConfigs (probably preferred)
- or a configuration package containing a XRD, Compositions and EnvironmentConfigs (assuming the environment is not shared)

And then we could rely on crossplane to manage dependencies between Compositions and EnvironmentConfigs.

That would make the deployment of `EnvironmentConfig` significantly simpler and more consistent.

PS: My understanding is that the current package format does not support/deploy additional custom resources of Kind = EnvironmentConfig (error below)
`kubectl crossplane: error: main.buildCmd.Run(): failed to build package: failed to parse package: {path:/home/user/project/src/cluster/premium-cluster.yaml position:0}: no kind "EnvironmentConfig" is registered for version "apiextensions.crossplane.io/v1alpha1" in scheme "pkg/runtime/scheme.go:100"`

bobh66:
Added support for EnvironmentConfig files in the `kubectl crossplane build` command for Provider and Configuration packages.

Fixes #3865 

I have:

- [C] Added `backport release-x.y` labels to auto-backport this PR if necessary.

Added two EnvironmentConfig yaml files to a Configuration package and verified that the package is generated successfully and contains the two EnvironmentConfig objects.

### DIFF
--- a/internal/xpkg/lint.go
+++ b/internal/xpkg/lint.go
@@ -27,6 +27,7 @@ import (
 	"github.com/crossplane/crossplane-runtime/pkg/parser"
 
 	v1 "github.com/crossplane/crossplane/apis/apiextensions/v1"
+	v1alpha1 "github.com/crossplane/crossplane/apis/apiextensions/v1alpha1"
 	pkgmetav1 "github.com/crossplane/crossplane/apis/pkg/meta/v1"
 	"github.com/crossplane/crossplane/internal/version"
 )
@@ -41,6 +42,7 @@ const (
 	errNotMutatingWebhookConfiguration   = "object is not a MutatingWebhookConfiguration"
 	errNotValidatingWebhookConfiguration = "object is not an ValidatingWebhookConfiguration"
 	errNotComposition                    = "object is not a Composition"
+	errNotEnvironmentConfig              = "object is not an EnvironmentConfig"
 	errBadConstraints                    = "package version constraints are poorly formatted"
 	errCrossplaneIncompatibleFmt         = "package is not compatible with Crossplane version (%s)"
 )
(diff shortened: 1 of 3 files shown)


hasheddan (commented):
Thanks for working on this @bobh66! A few notes / questions:
1. I read your comments [here](https://github.com/crossplane/crossplane/issues/3770#issuecomment-1500464348) and mostly agree with them, particularly:
> I view it as creating a runtime context from a number of sources, where different sources may exist in different scenarios and the absence of a given source is not necessarily an error condition.

With that being the case, I think it could potentially make more sense to be able to supply a set of defaults when no `EnvironmentConfig` is selected rather than shipping one in the package. As a general rule, we have tried to have an opinion about what gets created and owned via the package manager and what gets created by the user (e.g. via kubectl, gitops, etc.). In my mind `EnvironmentConfig` belongs in the latter category and coupling them to a package creates the possibility of collisions and / or reduced portability of packages.
2. If we were to allow for adding `EnvironmentConfig` to packages, I would prefer to wait until they reach `v1beta1` status.
3. You had mentioned a potential use-case for `EnvironmentConfig` in `Provider` packages -- could you outline some of those cases?

Thanks again for the thoughts and insight here!

bobh66:
Thanks for the comments @hasheddan  -  I think I agree, with the possible exception of EnvironmentConfigs in Configuration packages.  I can imagine a Configuration package with a set of Compositions that all want to share a common set of default configuration data and it would be useful to be able to maintain/ship that data in an EnvironmentConfig rather than duplicated in each Composition.

This can definitely wait for beta and in the interim I'll start work on the environment default configuration.

The Provider scenario I had in mind relates to creating a provider for an internal proprietary system that may have several different flavors, but I think that would be better handled in ProviderConfigs.  It might be nice to be able to ship default ProviderConfigs in a Provider package, but that can wait for a more concrete use case scenario.

I'll close this PR for now and open an issue for the environment defaults.  We can revisit this after EnvironmentConfigs reach beta if necessary.

Issue #3745, opened 2023-02-06
Support compsitionUpdatePolicy default value


bobh66:
### What problem are you facing?
The default value for compositionUpdatePolicy is Automatic which preserves the existing behavior. In order to fully take advantage of CompositionRevisions the compositionUpdatePolicy must be explicitly set to Manual on every Claim.

### How could Crossplane help solve your problem?
It would be very useful to be able to set a defaultCompositionUpdatePolicy on the XRD and have it copied to the Claim/Composite if it is not overridden in the Claim.  That would allow for composites to have a Manual composition update policy by default.

negz:
This sounds reasonable to me. Is this something you'd be interested in working on? If so, should we add it to the v1.12 milestone?

bobh66:
Yes - I think it's close enough to #3744 that I can handle both, though the implementation will likely be different.

phisco:
was addressed by #3963

PR #3963, opened 2023-04-10
Add defaults for CompositeDelete and CompositionUpdate policies

### LINKED ISSUE #3744: Support for compositeDeletionPolicy Default Value
We want to keep deleted claims by default if the underlying MRs haven't been removed yet.

One use-case for this is for our own provider where we want to provide some form of deletion policy. For instance, only delete a bucket if it's empty. If a user deletes a claim with such a deletion policy and the bucket isn't empty, then the claim will vanish but everything else will stay and still generate costs. For the user though it looks like the deletion was successful.

Currently a user has to set the `compositeDeletionPolicy` manually for this behavior to work, which is not really useful.

Ability to set the `compositeDeletionPolicy` to `Foreground` for by default for a given composite/claim.

bobh66:
Added defaultCompositeDeletePolicy and defaultCompositionUpdatePolicy attributes to the XRD to enable the author to specify defaults for how the Composite should be deleted and how Compositions should be updated.

Fixes #3744 #3745 

I have:

Created a Composite resource with:

```
spec:
  claimNames:
    kind: TEST
    plural: tests
  defaultCompositeDeletePolicy: Foreground
  defaultCompositionRef:
    name: test
  defaultCompositionUpdatePolicy: Manual
```

and verified that the Claim is modified to include:

```
  spec:
    compositeDeletePolicy: Foreground
    compositionRef:
      name: test
    compositionUpdatePolicy: Manual
```

and the Composite is modified to have:

```
  spec:
    compositionRef:
      name: test
    compositionUpdatePolicy: Manual
```

### DIFF
--- a/apis/apiextensions/v1/xrd_types.go
+++ b/apis/apiextensions/v1/xrd_types.go
@@ -56,6 +56,12 @@ type CompositeResourceDefinitionSpec struct {
 	// +optional
 	ConnectionSecretKeys []string `json:"connectionSecretKeys,omitempty"`
 
+	// DefaultCompositeDeletePolicy is the policy used when deleting the Composite
+	// that is associated with the Claim if no policy has been specified.
+	// +optional
+	// +kubebuilder:default=Background
+	DefaultCompositeDeletePolicy *xpv1.CompositeDeletePolicy `json:"defaultCompositeDeletePolicy,omitempty"`
+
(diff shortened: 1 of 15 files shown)


jbw976:
@turkenh would you be able to give this a final approval/merge now that @phisco took a pass through already? 🙏

turkenh (approved):

Issue #3715, opened 2023-02-03
Promote `CompositionRevision` to v1


negz:
### What problem are you facing?

Support for `CompositionRevision` was added at v1alpha1 in v1.4.0 and promoted to `v1beta1` in v1.11.0 per https://github.com/crossplane/crossplane/issues/2656.

### How could Crossplane help solve your problem?

We should look for any feedback around revisions for at least one release, then promote it to v1. When we do so we should:

* Remove the feature flag and make the CLI flag a no-op (i.e. the feature will be enabled regardless of the flag).
* Mark the CLI flag hidden so no-one sees it in the help output but anyone who is setting the flag won't be broken.
* Refactor the Composition -> CompositionRevision flow. Currently when revisions are enabled the XR reconciler converts the relevant CompositionRevision back to a Composition (internally) and uses that for its composition logic. I did this to keep the XR code identical whether revisions were enabled or not. Once revisions can't be disabled we can update the XR reconciler to be only concerned with CompositionRevisions and get rid of the code that converts from a CompositionRevision to a Composition.

negz:
* https://github.com/crossplane/crossplane/issues/3274

The above issue is the only one I'm aware of that may block promotion to v1 at the moment. I personally don't think it should though. In fact I think a lot of the problem that issue is trying to solve would be resolved because we would move `CompositionRevision` into the same Go package as `Composition` and allow them to share types.

negz:
Something else we'd need to do is merge the revisions documentation into the main Composition docs (related: https://github.com/crossplane/docs/issues/323).

PR #3964, opened 2023-04-11
Promote `CompositionRevision` to v1

negz:

Fixes https://github.com/crossplane/crossplane/issues/3715
Fixes https://github.com/crossplane/crossplane/issues/3274

This PR adds a v1 `CompositionRevision` type. __It also removes the v1alpha1 type__. The v1beta1 type will stick around for at least a few releases to give folks a chance to migrate to the v1 type. Technically we don't consider dropping a v1alpha API as a breaking change, but I'm adding the `breaking-api-change` label just to draw folks attention to this.

I have:

I've applied upbound/platform-ref-gcp and confirmed it works as expected. @ezgidemirel also tested updating from `v1beta1` to the new `v1` API types.


internal/controller/apiextensions/composite/reconciler.go
```diff
@@ -347,11 +354,18 @@ func NewReconciler(mgr manager.Manager, of resource.CompositeKind, opts ...Recon
...
-				return in.Validate().ToAggregate()
-			},
+		revision: revision{
+			CompositionRevisionFetcher: NewAPIRevisionFetcher(resource.ClientApplicator{Client: kube, Applicator: resource.NewAPIPatchingApplicator(kube)}),
+			CompositionRevisionValidator: CompositionRevisionValidatorFn(func(rev *v1.CompositionRevision) error {
+				// TODO(negz): Presumably this validation will eventually be
+				// removed in favor of the new Composition validation
+				// webhook.
+				// This is the last remaining use of conv.FromRevisionSpec -
+				// we can stop generating that once this is removed.
```
negz, line 364: @phisco Does this XR reconciler need to keep validating Compositions?

I'm guessing we do for now, assuming that the new validation webhook will be alpha and off by default.

negz:
This PR will need a corresponding docs change. Namely we can merge all the content pertaining to CompositionRevisions out of "guides" and into the mainline docs.

internal/controller/apiextensions/composite/reconciler.go
phisco: I would leave it, users can always delete the validating webhook, and validating webhooks do not run on existing resources upon upgrades, obviously it should never happen that a valid composition becomes invalid after an upgrade, nonetheless we should catch it as soon as possible.

As per your suggestion [here](https://github.com/crossplane/crossplane/issues/3748) webhooks are now enabled by default, given that the validation is also done in the reconciliation loop, it didn't make sense to split the `webhooks.enabled` value, so both XRD and Compositions will be enabled. The logical validation is not going to be behind feature flag, while future iterations using schemas will be, see https://github.com/crossplane/crossplane/pull/3937.

phisco:
https://github.com/crossplane/crossplane/pull/3964/commits/dffcf4c12cffef7304395a9a825b2783ce06f896 can be dropped, that code was removed by crossplane/crossplane#3937

internal/controller/apiextensions/composite/reconciler.go
negz: I'll leave it in place for this PR.

In a future PR I'd like to move it into the existing Composition controller, and have that controller set some status condition for invalid Compositions. I would prefer this because it:

* Could be used to ensure we never create a `CompositionRevision` for an invalid `Composition`, thus ensuring no-one can actually use an invalid `Composition`.
* Simplifies the XR controller.
* Avoids us having to convert from `Composition` -> `CompositionRevision` -> `Composition` to run this validation logic.

Does that sound reasonable? If so I may try do that before the freeze on Tuesday.

phisco: Sounds good to me! In the long run, once the schema-aware validation is promoted to GA, we can maybe think of running that too to compute the validity of a Composition.

ezgidemirel (approved):
Thanks for the PR @negz! It looks good to me. I also tested the upgrade scenario from v1beta1 to v1 and everything worked as expected.

negz:
Thank you @ezgidemirel!

Do you feel you did enough testing to feel comfortable merging this, or would there still be benefit in me testing it out manually before we do?

ezgidemirel:
> Thank you @ezgidemirel!
> 
> 
> 
> Do you feel you did enough testing to feel comfortable merging this, or would there still be benefit in me testing it out manually before we do?

I have compared the revision resources created before and after the upgrade. And only the apiversions were different as expected. After the upgrade I verified that automatic upgrade policy is working. I think it should be good to merge from regression point of view.

Issue #3941, opened 2023-04-04
Additionally added EnvironmentConfigs in Composition are not propagated to XR


P0t4T0o:
### What happened?

Given composition which references at least one env config:
if `environmentConfigs` list in Composition resource is extended by adding new references, these references are not propagated to XR, hence patching from newly added environment config fails.

### How can we reproduce it?

1. create composition which references env config `env-1` and patches some value from it into managed resource
   - eg. claim Demo and composite resource XDemo
2. update composition by extending `environmentConfigs` list with new env config `env-2` and apply
3. managed resource is not patched

describe XR (new env config is missing):
```yaml
Spec:
    API Version:  composite.demo.io/v1alpha1
    Kind:         Demo
    Name:         claim-1
    Namespace:    customer
  Composition Ref:
    Name:  xdemos.composite.demo.io
  Composition Revision Ref:
    Name:                     xdemos.composite.demo.io-8228a38
  Composition Update Policy:  Automatic
  Environment Config Refs:
    API Version:  apiextensions.crossplane.io/v1alpha1
    Kind:         EnvironmentConfig
    Name:         env-1
```

describe current composition revision:

```yaml
Spec:
  Composite Type Ref:
    API Version:  composite.demo.io/v1alpha1
    Kind:         XDemo
  Environment:
    Environment Configs:
      Ref:
        Name:  env-1
      Type:    Reference
      Ref:
        Name:  env-2
      Type:    Reference

```

---
note: if whole `Environment Config Refs:` section in XR is removed. XR immediately selfheals and successfully patches values from new env configs 

---

### What environment did it happen in?
Crossplane version: 

kubernetes: rancher desktop k8s 1.25.6 & EKS 1.22
kubectl: 1.26.1
kernel: Darwin Kernel Version 21.6.0

P0t4T0o:
@MisterMX what is that intention? Im just thinking about if it wouldnt be enough to remove
[if len(cr.GetEnvironmentConfigReferences())](https://github.com/crossplane/crossplane/blob/master/internal/controller/apiextensions/composite/environment/selector.go#L75) in `selector.go` without adding a policy? from XR point of view there will be no breaking change since all "old" env configs are fetched every reconciliation cycle anyway - so they are already treated like if they had `ResolvePolicy=Always`. Removal of the if-statement will enable us to include new and exclude old env configs in updated compositions and newer compositionrevisions.
[...]

Issue #3723, opened 2023-02-03
Allow ignoring environment config if it does not exists


duizabojul:
### What problem are you facing?

I don't know if I am using environment config the bad way but I can't figure how to ignore selection of an `EnvironmentConfig` if it does not exists.

Let's say I have the following `EnvironmentConfig` resources:

```yaml
apiVersion: apiextensions.crossplane.io/v1alpha1
kind: EnvironmentConfig
metadata:
  labels:
    provider-name: default-provider
  name: defaut-provider-config
data:
  id: 123456
 ```
 
```yaml
apiVersion: apiextensions.crossplane.io/v1alpha1
kind: EnvironmentConfig
metadata:
  labels:
    provider-name: other-provider
  name: other-provider-config
data:
  id: 456789
 ```

Then I have a XRD with an optional field `providerConfigRef.name` in spec

In my composition I want to load the `default-provider-config EnvironmentConfig` if `providerConfigRef.name` is not referenced in my XR, and the correct provider `EnvironmentConfig` if it is referenced (let's say `other-provider-config EnvironmentConfig` if `providerConfigRef.name : other-provider`). 

So I tried to add the following environment in the associated composition:

```yaml
...
environment:
  environmentConfigs:
    - type: Selector
      selector:
        matchLabels:
        - key: provider-name
          type: Value
          value: default-provider
    - type: Selector
      selector:
        matchLabels:
        - key: provider-name
          type: FromCompositeFieldPath
          valueFromFieldPath: spec.providerConfigRef.name
```

The problem is if `spec.providerConfigRef.name` does not exists, crossplane will fail reconciliation with the following message: `cannot select environment: failed to build reference at index 1: failed to resolve value for label at index 0: spec.providerConfigRef: no such field` 

### How could Crossplane help solve your problem?

Maybe there is a workaround to this issue, but if not it would be neat to have a way to ignore missing environment config, the same way as patches are ignored if field path does not exists

negz:
CC @MisterMX

phisco:
This was addressed introducing `spec.environment.policy.resolution: Optional` (default: Required) by: https://github.com/crossplane/crossplane/pull/3981

phisco:
Actually not, I misread the issue, this is still valid

MisterMX:
A workaround here in conjuction with #3981 is to use `selector` instead of `ref` which can be defined to be optional.

[...]

PR #3981, opened 2023-04-13
Support selecting multiple environment configs

P0t4T0o:

related to: https://github.com/crossplane/crossplane/issues/3770, see https://github.com/crossplane/crossplane/issues/3770#issuecomment-1491628182 & https://github.com/crossplane/crossplane/issues/3941

This PR is successor of https://github.com/crossplane/crossplane/pull/3931. It introduces resolve policy to EnvironmentConfig and updates selection to aggregate all environment configs with matching labels. 

Fixes https://github.com/crossplane/crossplane/issues/3941 & https://github.com/crossplane/crossplane/issues/3770#issuecomment-1491628182 

I have:

In local environment & via unit tests.
E2Es will come in a separate PR.


apis/apiextensions/v1/composition_environment.go
```diff
@@ -41,11 +41,19 @@ type EnvironmentConfiguration struct {
...
 const (
 	// EnvironmentSourceTypeReference by name.
 	EnvironmentSourceTypeReference EnvironmentSourceType = "Reference"
 	// EnvironmentSourceTypeSelector by labels.
 	EnvironmentSourceTypeSelector EnvironmentSourceType = "Selector"
+
+	// ResolvePolicyAlways is a resolve option.
+	// When the ResolvePolicy is set to ResolvePolicyAlways the reference will
+	// be tried to resolve for every reconcile loop.
+	ResolvePolicyAlways ResolvePolicy = "Always"
```
bobh66, line 56: Should there also be a ResolvePolicyOnce value as the default?

apis/apiextensions/v1/composition_environment.go
```diff
@@ -71,12 +79,40 @@ type EnvironmentSource struct {
...
 }
 
 // An EnvironmentSourceSelector selects an EnvironmentConfig via labels.
 type EnvironmentSourceSelector struct {
 	// MatchLabels ensures an object with matching labels is selected.
 	MatchLabels []EnvironmentSourceSelectorLabelMatcher `json:"matchLabels,omitempty"`
+
+	// Policies for referencing.
+	// +optional
+	Policy *Policy `json:"policy,omitempty"`
```
bobh66, line 95: The Policy is attached to the Source/Selector?  So different sources/selectors can have different Policies? Does any one source/selector having the policy attached trigger the entire environment to be recreated?  If so why not have the Policy at the Composition.environment level?

bobh66 (commented):
A couple of comments inline - I'm not sure that this level of fine-grained control is needed?

apis/apiextensions/v1/composition_environment.go
P0t4T0o: makes sense, will adjust it

apis/apiextensions/v1/composition_environment.go
```diff
@@ -36,16 +36,28 @@ type EnvironmentConfiguration struct {
 	// Patches is a list of environment patches that are executed before a
 	// composition's resources are composed.
 	Patches []EnvironmentPatch `json:"patches,omitempty"`
+
+	// Policies for referencing.
```
negz, line 40: Please add more detail this comment. How does the `EnvironmentConfiguration` type use this policy? What for?

apis/apiextensions/v1/composition_environment.go
```diff
@@ -36,16 +36,28 @@ type EnvironmentConfiguration struct {
...
+	// +optional
+	Policy *Policy `json:"policy,omitempty"`
 }
 
 // EnvironmentSourceType specifies the way the EnvironmentConfig is selected.
 type EnvironmentSourceType string
 
+// ResolvePolicy is a type for resolve policy.
```
negz, line 48: Please add more detail to this comment. What is a resolve policy?

apis/apiextensions/v1/composition_environment.go
```diff
@@ -79,6 +91,26 @@ type EnvironmentSourceSelector struct {
 	MatchLabels []EnvironmentSourceSelectorLabelMatcher `json:"matchLabels,omitempty"`
 }
 
+// Policy represents policies of EnvironmentSourceReference instance.
+type Policy struct {
```
negz, line 95: https://github.com/crossplane/crossplane-runtime/blob/1316ae6695eec09cf47abdfd0bc6273aeaab1895/apis/common/v1/resource.go#L79

Could we use the existing `Policy` type from crossplane-runtime? Or are you trying to avoid introducing a `ResolutionPolicy` as well as a `ResolvePolicy`? (I regret those confusing names.)

apis/apiextensions/v1/composition_environment.go
negz: Please do. I think its a mistake that crossplane-runtime doesn't define it.

internal/controller/apiextensions/composite/environment/selector.go
```diff
@@ -68,28 +68,45 @@ type APIEnvironmentSelector struct {
...
+
 // SelectEnvironment for cr using the configuration defined in comp.
 // The computed list of EnvironmentConfig references will be stored in cr.
-func (s *APIEnvironmentSelector) SelectEnvironment(ctx context.Context, cr resource.Composite, rev *v1.CompositionRevision) error {
-	// noop if EnvironmentConfig references are already computed
-	if len(cr.GetEnvironmentConfigReferences()) > 0 {
-		return nil
-	}
-	if rev.Spec.Environment == nil || rev.Spec.Environment.EnvironmentConfigs == nil {
+func (s *APIEnvironmentSelector) SelectEnvironment(ctx context.Context, cr resource.Composite, comp *v1.CompositionRevision) error {
```
negz, line 87:
```suggestion
func (s *APIEnvironmentSelector) SelectEnvironment(ctx context.Context, cr resource.Composite, rev *v1.CompositionRevision) error {
```

Is there a reason to change this variable name? I believe we call revisions `rev` elsewhere.

apis/apiextensions/fn/proto/v1alpha1/run_function.pb.go
```diff
@@ -17,7 +17,7 @@
 
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.29.1
+// 	protoc-gen-go v1.28.1-devel
```
negz, line 20: This change probably doesn't need to be in this PR.

internal/controller/apiextensions/composite/environment/selector.go
```diff
@@ -106,28 +123,33 @@ func (s *APIEnvironmentSelector) buildEnvironmentConfigRefFromRef(ref *v1.Enviro
...
-	}, nil
+
+	envConfigs := make([]corev1.ObjectReference, len(res.Items))
+	for i, v := range res.Items {
+		envConfigs[i] = corev1.ObjectReference{
+			Name:       v.Name,
+			Kind:       v1alpha1.EnvironmentConfigKind,
+			APIVersion: v1alpha1.SchemeGroupVersion.String(),
+		}
+	}
```
phisco, line 150: Aren't both the original implementation and this one assuming that List is always going to return EnvironmentConfigs in the same order? Kind of hard to test in practice, but couldn't that potentially lead to flapping, if for example two EnvironmentConfigs are defining the same key with different values? While with the original implementation couldn't it have selected a different one at the next reconciliation loop. Should we enforce some kind of ordering? Not sure which one though.

Moreover, I feel like the [RefForFirstLabelSelectedObject](https://github.com/crossplane/crossplane/blob/master/internal/controller/apiextensions/composite/environment/selector_test.go#L247) test-case should fail with this PR, but it's not 🤔 CC @MisterMX

phisco: I did some tests and given that the selected environmentConfigsRefs are stored in the XR at `spec.environmentConfigRefs`, it won't flap on a single resource, but it's going to pick an order randomly when a new XR is created. 🤔 Do we want to take the change to enforce an order?

apis/apiextensions/v1/composition_environment.go
P0t4T0o: refactored to `Policy` type from crossplane runtime. Initially I avoided it due to ResolutionPolicy - considered that any provided reference should either be resolved or error emitted. With this change I also added check to not return error when fetching of env config fails AND resolution policy is `Optional`

apis/apiextensions/fn/proto/v1alpha1/run_function.pb.go
P0t4T0o: reverted

internal/controller/apiextensions/composite/environment/selector.go
P0t4T0o: renamed back, this was probably leftover from resolving merge conflicts

internal/controller/apiextensions/composite/environment/selector.go
P0t4T0o: `RefForFirstLabelSelectedObject` test-case might have misleading name. since composite has set env config reference:
```go
cr: composite(
	withEnvironmentRefs(environmentConfigRef("test")),
),
```
and resolvePolicy is not set,`isNoOp` fn signals that no further resolution is needed. hence only already present ref is kept. This applies to original version as well which has this check in `SelectEnvironment()`:
```go
// noop if EnvironmentConfig references are already computed
if len(cr.GetEnvironmentConfigReferences()) > 0 {
    return nil
}
```

phisco: Created #4070 to track that as I feel it's independent to this PR, but could still be addressed here as I feel like we should not select randomly ordered configs at each time, same as we should not just pick one at random as it's currently done.

For the test case, I think it's more likely that the title is correct and the content wrong, the following changes would make it consistent with the test case name, pass with the current implementation and fail with this patch, which means the test case should be updated according to this patch new behavior:
```diff
diff --git a/internal/controller/apiextensions/composite/environment/selector_test.go b/internal/controller/apiextensions/composite/environment/selector_test.go
index 207614ce..e8e94da8 100644
--- a/internal/controller/apiextensions/composite/environment/selector_test.go
+++ b/internal/controller/apiextensions/composite/environment/selector_test.go
@@ -265,9 +265,7 @@ func TestSelect(t *testing.T) {
 						return nil
 					}),
 				},
-				cr: composite(
-					withEnvironmentRefs(environmentConfigRef("test")),
-				),
+				cr: composite(),
 				rev: &v1.CompositionRevision{
 					Spec: v1.CompositionRevisionSpec{
 						Environment: &v1.EnvironmentConfiguration{
@@ -277,6 +275,7 @@ func TestSelect(t *testing.T) {
 									Selector: &v1.EnvironmentSourceSelector{
 										MatchLabels: []v1.EnvironmentSourceSelectorLabelMatcher{
 											{
+												Type:  v1.EnvironmentSourceSelectorLabelMatcherTypeValue,
 												Key:   "foo",
 												Value: pointer.String("bar"),
 											},
```

phisco:
@P0t4T0o, this is also addressing #3723, right?

P0t4T0o:
> @P0t4T0o, this is also addressing #3723, right?

yes, based on [bobh66's comment](https://github.com/crossplane/crossplane/pull/3931#discussion_r1160831223) selection was adjusted to return empty list rather than error if no env configs with matching labels are found

MisterMX:
As described in https://github.com/crossplane/crossplane/issues/4070#issuecomment-1554411112 I would recommend against selecting multiple configs at once as this might cause unexpected behaviour that highly depends on the number of existing environment configs that might overwrite each other without the user noticing.

This might also be a possible security issue as one could utilize a bug in another composition to generate an environment config that is selected by a second composition possibly allowing an attacker to gain access to resources he should have.

Therefore I would choose security and strictness over convenience and return an error if multiple configs are selected.

phisco:
Food for thought: is a label selector then the right tool for the task at hand? Couldn't we add some kind of templated name selector?

bobh66:
> Food for thought: is a label selector then the right tool for the task at hand? Couldn't we add some kind of templated name selector?

We have found that there are generally two types of EnvironmentConfigs - one type is "statically generated" or manually created and has specific names (dev, prod, etc) and can be referenced using name selectors. The second type is "dynamically created", for example as part of a deployment of nested compositions, and therefore uses generated names to avoid naming collisions, and these are referenced using label selectors.

According to #3941 the original intent of the PR was to allow adding additional EnvironmentConfig selectors to an existing composition/composte.  That is significantly different than allowing a single EnvironmentConfig selector to select multiple EnvironmentConfigs.

This PR is already complicated enough that if there is a scenario for allowing multiple EnvironmentConfigs to be returned for a single selector, that should probably be documented and addressed by a separate issue/PR.  My .02

phisco:
@P0t4T0o, with @MisterMX we had a chat and we came up with the following design.

## Recap

- The current implementation has an issue: label selectors are non-deterministically picking a single  at "resolve" (#4070)
- This PR is introducing the following user requested changes:
  - multiple env selection: at "resolve" time when label selectors are evaluated all envs matching are stored in the XR (https://github.com/crossplane/crossplane/issues/3770#issuecomment-1491628182)
  - Resolve policies: allowing to have opt-in "live updates" of `EnvironmentConfigs` from Compositions to XRs, but also for envs selected through label selectors (https://github.com/crossplane/crossplane/issues/3941)
  - Resolution policies: allowing to ignore missing envs (#3723)
- With the current implementation, introducing the `Always` resolve policy alone, would make the current issue much worse, as that would mean XR's would flap at each reconciliation loop at random in case of envs matching the label selectors.
- The same would happen with multiple envs support, only that instead of 1, we would write all the envs matching into the XR.

## Design

**!!! DISCLAIMER: This is just a design we came up with after a quick chat, a proposal we'd like to have some feedback on !!!**

We identified the following requirements:
R1) Users should be able to select a single env, knowing its exact name
R2) Users should be able to select a single env, without knowing its exact name
R3) Users should be notified in case multiple matches are found, if that's what they want.
R4) Users should be able to select multiple envs they don't know the name of and merge them
R5) Bonus: Users should be able to select a single env by name if it follows some pattern (e.g. `...-prod`, `...-dev`)

To do so we thought about the following design.

Currently one in either if the names of the env to be dynamically selected was known or not, would write a Composition as follows:
```yaml
spec:
  environment:
    environmentConfigs:
    - type: Reference                   # satisfies R1
      ref:
        name: "my-awesome-default-config"
    - type: Selector                      # satisfies R2, but suffers the issue described in #4070
      selector:
        matchLabels:
          key: stage
          valueFromFieldPath: metadata.labels[stage]
```
So there is currently no way to satisfy R3, R4 nor R5.

The same config above, but with patch as we said would satisfy R4, but not satisfy R2 anymore, as there would be no way to select a single env not knowing its name.

But what if we implemented the following API?

R1 is satisfied by the `Reference` type of selector, so no issue there.

If you didn’t know the envs’ names but you still wanted a single one to be read and error out you could write:
```yaml
spec:
  environment:
    environmentConfigs:
    - type: Reference
      ref:
        name: "my-awesome-default-config"
    - type: Selector
      selector:
        mode: "single"         #  could be the default to preserve the current behaviour, instead of just returning all the matches by default, which would be a breaking change.
        matchLabels:
          key: stage
          valueFromFieldPath: metadata.labels[stage]
```
This would satisfy R2, but also R3, as in single mode an event would be emitted in case multiple matches were found.

If then, instead you wanted any number of matches, of witch you obviously don't know the name, you could do as follows:
```yaml
spec:
  environment:
    environmentConfigs:
    - type: Reference
      ref:
        name: "my-awesome-default-config"
    - type: Selector
      selector:
        mode: "multiple"
        maxMatch: N               # *integer, nil being unbounded if in "multiple" mode, any other number >0 being the max number of elements that should be accepted, erroring otherwise
        sortByFieldPath: metadata.name          # you would need to provide a sorting key as a fieldpath so that we can sort the matches
        matchLabels:
          key: stage
          valueFromFieldPath: metadata.labels[stage]
```
This would allow to also satisfy R4.

The above design plus the resolve and resolution policies as you have already implemented,  @P0t4T0o, we think could satisfy all the core requirements, fixing also the non-determinism issue we currently already have. 🎉 

**Bonus**, which is what I was hinting at with my comment above.

If someone knew the names of the envs are all matching the same pattern (e.g. `...-prod`, `...-dev`) and you wanted only at most a single one to be picked we could think of adding a new field to the `Reference` type, specifying how to translate a value from the Composite Resource leveraging the usual transform pipeline into a the env `name`. It could look like this:
```yaml
spec:
  environment:
    environmentConfigs:
    - type: Reference
      ref:
        name: "my-awesome-default-config"
    - type: Reference
      refFromComposite:       #  < ==== HERE
        fromFieldPath: metadata.labels[stage]
        transforms:
        - type: string
          string:
            fmt: "my-awesome-patch-%s"
```

phisco:
Hi @P0t4T0o, the patch looks good, there is still a design question regarding erroring out on more matches found I'd like to discuss with the wider team (see [comment](https://github.com/crossplane/crossplane/pull/3981#discussion_r1235378900)), it would be awesome if you could join the #sig-composition-environments channel on the Crossplane Slack and also if you could join the first bi-weekly SIG meeting we are going to kick-off next week, you can find the "Crossplane SIG Environment Configs" event on the Crossplane community calendar. See the [README.md](https://github.com/crossplane/crossplane/tree/master#get-involved) for links to the calendar and slack. 🙏

P0t4T0o:
hi @phisco, 
sorry for my late response and many thanks for your quick review. Added changes should address review comments and functionality should follow discussed design. Thank you for the invitation, I would be more than happy to join SIG meeting. Unfortunately Im abroad till the end of this month and cannot promise that I will join for sure, but I can join the slack channel and reach out to you guys there for further discussion

phisco:
Hi @P0t4T0o,
so yesterday we had the first SIG meeting as I was saying, you can find a Doc with a few notes on the `#sig-composition-environments` channel on Slack, during which we discussed this PR, we came to the following conclusions about the two modes:
- `Single` mode (default): **should error out** if more than one is found => which is a breaking API change
- `Multiple` mode: **should not error out**, but take up to maxMatch environmentConfigs, nil meaning all matches found, sorted by sortByFieldPath
  - Setting Multiple + `maxMatch` = 1 would result in the current behaviour, so of picking just one without erroring out, but deterministically sorted
- We agreed we should improve the erroring out part though, as at the moment it's just a `Warning` and should at least be an `Error` if the `Resolve` policy is set to `Required`, but just a `warning` if set to `Optional`

WDYT?

Overall as I was saying the PR looks good, I still have to test it out thoroughly I would only add validation and maybe some E2Es, as we started introducing them in the meantime, if you want I can help you on these things or I can add them later on, just let me know 🙏

phisco (approved):
Thanks @P0t4T0o! LGTM now, we'll add E2Es in a separate PR, but I tested it manually and everything works as intended!

phisco:
Used the following files to test it out:yaml
```
# ==> test/e2e/manifests/apiextensions/environment/setup/composition.yaml <==
---
kind: Composition
metadata:
  name: nop.sqlinstances.example.org
spec:
  compositeTypeRef:
    kind: XSQLInstance
  environment:
    policy:
      resolve: Always
      resolution: Required
    environmentConfigs:
      - type: Reference
        ref:
          name: example-environment-1
      - type: Selector
        selector:
          mode: Multiple
          maxMatch: 1
          sortByFieldPath: data.priority
          matchLabels:
              key: stage
              valueFromFieldPath: metadata.labels[stage]
  resources:
    - name: nop
      base:
        kind: NopResource
        spec:
          forProvider:
            conditionAfter:
              - time: 10s
                conditionType: Ready
                conditionStatus: "True"
          providerConfigRef:
            name: default
      patches:
        - type: FromEnvironmentFieldPath    # <- setting the value of complex.c.f from the computed environment, will be 2 or 3 according to whatever env is selected or would be the last selected if multiple envs are selected.
          fromFieldPath: complex.c.f
          toFieldPath: metadata.annotations[nop.crossplane.io/complex.c.f]

# ==> test/e2e/manifests/apiextensions/environment/setup/definition.yaml <==
---
metadata:
  name: xsqlinstances.example.org
spec:
  group: example.org
  names:
    kind: XSQLInstance
    plural: xsqlinstances
  claimNames:
    kind: SQLInstance
    plural: sqlinstances
  defaultCompositionRef:
    name: nop.sqlinstances.example.org
  defaultCompositeDeletePolicy: Background
  defaultCompositionUpdatePolicy: Automatic
  versions:
    - name: v1alpha1
      served: true
      referenceable: true
      schema:
        openAPIV3Schema:
          type: object
          properties:
            spec:
              type: object
              properties:
                parameters:
                  type: object
                  properties:
                    storageGB:
                      type: integer
                  required:
                    - storageGB
              required:
                - parameters
            status:
              type: object
              properties:
                address:
                  description: Address of this MySQL server.
                  type: string

# ==> test/e2e/manifests/apiextensions/environment/setup/environmentConfigs.yaml <==
---
kind: EnvironmentConfig
metadata:
  name: example-environment-1
data:
  complex:
    a: b
    c:
      d: e
      f: "1"
---
kind: EnvironmentConfig
metadata:
  name: example-environment-2
  labels:
    stage: prod
data:
  priority: 2 # <- lower priority than example-environment-3 below
  complex:
    c:
      f: "2" # <- same path, different value w.r.t. example-environment-3 below
---
kind: EnvironmentConfig
metadata:
  name: example-environment-3
  labels:
    stage: prod
data:
  priority: 1 # <- higher priority than example-environment-2 above
  complex:
    c:
      f: "3" # <- same path, different value w.r.t. example-environment-2 above

# ==> test/e2e/manifests/apiextensions/environment/setup/provider.yaml <==
kind: Provider
metadata:
  name: provider-nop
spec:
  package: xpkg.upbound.io/crossplane-contrib/provider-nop:v0.2.0

# ==> test/e2e/manifests/apiextensions/environment/claim.yaml <==
---
kind: SQLInstance
metadata:
  name: example
  labels:
    stage: prod
spec:
  parameters:
    storageGB: 10
  writeConnectionSecretToRef:
    name: example-sqlinstance-connection

# ==> test/e2e/manifests/apiextensions/environment/modifiedEnvironmentConfig.yaml <==
---
kind: EnvironmentConfig
metadata:
  name: example-environment-3
  labels:
    stage: prod
data:
  priority: 1 # <- higher priority than example-environment-2 above
  complex:
    c:
      f: "4" # <- same path, different value w.r.t. example-environment-3 below
```
Playing around with the different settings introduced by this PR, resolve/resolution policies and selector's modes, Multiple/Single. More structured checks will be done via e2e in a separate PR as I was saying.

P0t4T0o:
> Thanks @P0t4T0o! LGTM now, we'll add E2Es in a separate PR, but I tested it manually and everything works as intended!

Many thanks to you too, @phisco . I really appreciate your advice, willingness and approach to help throughout the entire PR !

phisco:
As agreed with @P0t4T0o, I rebased and cleaned up the history a little bit. Took the chance to manually test it further while defining the scenarios and the manifests for the e2es, a PR will follow for that.

PR #3989, opened 2023-04-14
feat: math transform supporting floats

### LINKED ISSUE #3059: Enhance `multiply` of transform `math` type with support of `float64`

phisco:

Fixes #3059.

Added float64 support for all current types of math transforms:
- multiply will return float64 if the input is a float64, int64 otherwise
- clampMin/clampMax will return the threshold value as int64 if the input is below/above it, the original input otherwise.

I'd like to highlight that it's suggested not to use floats as outputs, see https://github.com/kubernetes-sigs/controller-tools/issues/245 and https://github.com/crossplane/crossplane/issues/3059#issuecomment-1112744496, but currently there is no transform from float64 to quantities, should we add that too in the context of this PR or should we handle it separately?
Something like the following could be added to https://github.com/crossplane/crossplane/blob/7a55c644a1b50c60d09c2695d56ee6ad61b37424/apis/apiextensions/v1/composition_transforms.go#L182:
```go
	{from: TransformIOTypeFloat64, to: TransformIOTypeString, format: ConvertTransformFormatQuantity}: func(i any) (any, error) { //nolint:unparam // See note above.
		q := resource.NewQuantity(int64(i.(float64)), resource.DecimalSI)
		return q.String(), nil
	},
```

I have:

Unit tests added.

### DIFF
--- a/internal/controller/apiextensions/composite/composition_patches_test.go
+++ b/internal/controller/apiextensions/composite/composition_patches_test.go
@@ -1382,3 +1382,94 @@ func TestComposedTemplates(t *testing.T) {
 		})
 	}
 }
+
+func TestResolveTransforms(t *testing.T) {
+	type args struct {
+		ts    []v1.Transform
+		input any
+	}
+	type want struct {
+		output any
+		err    error
+	}
+	tests := []struct {
+		name string
+		args args
+		want want
+	}{
+		{
+			name: "NoTransforms",
+			args: args{
+				ts: nil,
+				input: map[string]interface{}{
+					"spec": map[string]interface{}{
+						"parameters": map[string]interface{}{
+							"test": "test",
+						},
+					},
+				},
+			},
+			want: want{
+				output: map[string]interface{}{
+					"spec": map[string]interface{}{
+						"parameters": map[string]interface{}{
+							"test": "test",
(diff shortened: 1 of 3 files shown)


internal/controller/apiextensions/composite/composition_patches_test.go
```diff
@@ -1382,3 +1382,94 @@ func TestComposedTemplates(t *testing.T) {
...
+			},
+		},
+		{
+			name: "MathTransformWithConversionToFloat64",
+			args: args{
+				ts: []v1.Transform{{
+					Type: v1.TransformTypeConvert,
+					Convert: &v1.ConvertTransform{
+						ToType: v1.TransformIOTypeFloat64,
+					},
```
lsviben, line 1429: Do we need convert transforms before the Multiply transform in these tests?

lsviben (approved):
Added a minor comment, otherwise LGTM. 

If we are worried about supporting float outputs we could add a warning in the composition validation code if a transform chain ends in a float?

negz (approved):

PR #4002, opened 2023-04-19
build(helm): remove values template, generate readme

### LINKED ISSUE #3994: Drop templated helm chart values

Currently Crossplane's chart only has a templated values file committed, the actual version is only rendered at build time either to deploy it locally or to publish it. This makes it confusing for users, which are used to look for a `values.yaml` file, more difficult to use tools that assume values to be there, e.g. [helm-docs](https://github.com/norwoodj/helm-docs), and adds useless cognitive load.

We should switch to use the appVersion and an empty tag as already done by other widely adopted charts, e.g. https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/values.yaml#L2174 and https://github.com/argoproj/argo-helm/blob/main/charts/argo-workflows/values.yaml#L2. By doing so we could eliminate the `values.yaml.tmpl` file and we could remove the `helm.prepare` target from the build submodule.

Given that the appVersion bumping logic is already in place for publishing, we would only need to `--set image.tag="$(cat _output/version)"` in the [kind.sh](https://github.com/crossplane/crossplane/blob/master/cluster/local/kind.sh#L86) script used to deploy Crossplane locally.

phisco:

Fixes #3994.

I have:


cluster/local/integration_tests.sh
```diff
@@ -71,7 +71,7 @@ docker tag "${BUILD_IMAGE}" "${CROSSPLANE_IMAGE}"
 "${KIND}" load docker-image "${CROSSPLANE_IMAGE}" --name="${K8S_CLUSTER}"
 
 echo_step "installing helm package(s) into \"${CROSSPLANE_NAMESPACE}\" namespace"
-"${HELM3}" install --create-namespace -n "${CROSSPLANE_NAMESPACE}" "${PROJECT_NAME}" "${projectdir}/cluster/charts/${PROJECT_NAME}" --set replicas=2,args={'-d'},rbacManager.replicas=2,rbacManager.args={'-d'},image.pullPolicy=Never,imagePullSecrets=''
+"${HELM3}" install --create-namespace -n "${CROSSPLANE_NAMESPACE}" "${PROJECT_NAME}" "${projectdir}/cluster/charts/${PROJECT_NAME}" --set replicas=2,args={'-d'},rbacManager.replicas=2,rbacManager.args={'-d'},image.pullPolicy=Never,imagePullSecrets='',image.tag=${helm_tag}
```
hasheddan, line 74: Do we need to set the image tag explicitly here if we are falling back to chart version as the default?

phisco: I was hoping the same here, but turns out we are not populating the helm chart's `Chart.yaml' before running the e2es, so it won't be populated with a meaningful version and then I think we would want to set the same tag we just used to upload images nonetheless.

hasheddan: Ah, I didn't notice we were installing from the `cluster/` dir rather than the built `_output`. Since we have to build the image anyway, we could opt to install from `_output`, but the nice aspect of the current approach is that we only have to rebuild the image, not the helm chart as well 👍🏻

hasheddan (approved):
Thanks @phisco!

PR #4010, opened 2023-04-21
La familia

negz:

The design in https://github.com/crossplane/crossplane/pull/3939 proposes we break up larger providers like provider-aws by service. We'd group these providers into a 'family' for two reasons:

1. So they could all share one ProviderConfig (and StoreConfig etc).
2. So they could cross-resource-reference each other.

We believe 1 will always be true. Hopefully 2 will eventually go away with https://github.com/crossplane/crossplane/issues/1770.

During testing we realised providers in the same family would need RBAC access to read all types in their family, i.e. to reference an MR or read a ProviderConfig.

This PR introduces the `pkg.crossplane.io/provider-family` label. This optional label can be added to provider package metadata (i.e. `crossplane.yaml`). It will be propagated to the `ProviderRevision`, where the RBAC manager will use it to ensure that all Provider pod service accounts have RBAC access to read, create, and update all of the types (i.e. MRs, ProviderConfigs, etc) within its family.

I have:

Just unit tests at this point. I'll test it before it leaves draft.


apis/pkg/v1/interfaces.go
```diff
@@ -28,6 +28,19 @@ const (
...
+	// active revision's label changed for some reason. There's no technical
+	// reason to need it, but being able to list provider.pkg by family seems
+	// convenient.
+
+	// LabelProviderFamily is used as key for the provider family label. This
+	// label is added to any provider that rolls up to a larger 'family', such
+	// as 'family-aws'. It is propagated from provider metadata to provider
+	// revisions, and can be used to select all provider revisions that belong
+	// to a particular family. It is not added to providers, only revisions.
+	LabelProviderFamily = "pkg.crossplane.io/provider-family"
```
negz, line 43: CC @nullable-eth. I'm going with this as the label (not annotation) that will show up in `crossplane.yaml` for any provider that's part of a family. It will also show up on the revision, per the comment.

negz:
I want to note that there is an alternative to making this change.

Providers can already ask for additional permissions beyond access to reconcile the types they install. They do this by specifying a desired set of RBAC rules under `spec.permissionRequests` in their package metadata (i.e. their `crossplane.yaml`). This means a provider could request access to all the things it needs.

The benefit of this alternative approach is that it's more scoped. A provider that could only use a `ProviderConfig` and reference `VPC` resources could request only access to those types, not every type in its family.

The downside is that these permission requests must be approved at runtime. Crossplane administrators must choose which permissions a provider can request by allow-listing them. The allow list takes the form of a single `ClusterRole`. Providers may request any permission that is allowed by that `ClusterRole`. Any permission not allowed by the `ClusterRole` is not granted.

apis/pkg/v1/interfaces.go
negz: We're discussion some potential alternatives to family - specifically "cloud" and "vendor".

negz:
@turkenh Thanks for testing that! Do you know whether you see the same CPU consumption when you install ~70 packages _without_ this change? What I am wondering is whether we think this change specifically causes the increased CPU utilization, or whether it's (mostly) a pre-existing issue with the RBAC manager?

If we don't think the issue was specifically introduced by this PR I'd like to proceed with it. That way using provider families is possible (but not performant) and we can work on performance in a separate PR.

negz:
> I am happy to proceed with this PR

Thanks, I agree. It seems like the RBAC manager is [respecting its soft limits](https://github.com/crossplane/crossplane/blob/028a7634fe6dffa1b3d6c0b01418bc11f5c5e26a/cluster/charts/crossplane/templates/rbac-manager-deployment.yaml#L83). Given those limits are fairly low (100m CPU, 512Mi memory) I feel good about proceeding despite the increased usage. We don't in practice expect anyone to be installing 50+ providers at once anytime soon, and the behaviour should remain the same for anyone not using a provider family.

I'll continue to poke at this today, but if I find any improvements I will follow up with a separate PR.

PR #4024, opened 2023-04-27
set xfn image tag in kind.sh

lsviben:
Sets the xfn sidecar image tag that was missing in `kind.sh` install/upgrade scripts.

Noticed this with @pedjak while running some tests using the `kind.sh` script to set up the local environment with Composition functions enabled. Its missing a set for the xfn image [here](https://github.com/crossplane/crossplane/blob/9e41a7395bfafd84fab1b3223432a8fe4520fbc6/cluster/local/kind.sh). 

I have:

Ran the `kind.sh` helm-install/helm-upgrade and checked if the proper image is set.

### DIFF
--- a/cluster/local/kind.sh
+++ b/cluster/local/kind.sh
@@ -73,7 +73,7 @@ case "${1:-}" in
 
     [ "$2" ] && ns=$2 || ns="${DEFAULT_NAMESPACE}"
     echo "installing helm package into \"$ns\" namespace"
-    ${HELM3} install ${PROJECT_NAME} --namespace ${ns} --create-namespace ${projectdir}/cluster/charts/${PROJECT_NAME} --set image.pullPolicy=Never,imagePullSecrets='',image.tag="${helm_tag}" ${HELM3_FLAGS}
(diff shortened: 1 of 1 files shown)


phisco (approved):

negz:
> But instead of adjusting the script it seems better to just fix the image tag by default to the crossplane image.

Why is that?

phisco:
can you expand on why/how it was failing in the `kind.sh` script, @lsviben?

lsviben:
> can you expand on why/how it was failing in the `kind.sh` script, @lsviben?

It was failing as it was missing the `--set xfn.image.tag="${helm_tag}"` similar to the existing `--set image.tag="${helm_tag}"` that is present for the Crossplane image. So the xfn image was 0.0.1 which i guess is the `.Chart.AppVersion` that is generated in the script.


> Why is that?
My assumption is that we would suggest running Crosspland and XFN with the same version/image tag. 
Before this change they were also both being defaulted to the same `.Chart.AppVersion`. Taking another look at this now and I see that a possible issue could be if somebody sets Crossplane to some custom repository and image tag and leaves the XFN to default. 

BTW I dont have strong feelings about this, I can just change the script and leave the defaulting as it was. Looking at the above case, maybe its for the best

lsviben:
> Got it, yes, definitely forgot to set that too in `kind.sh`. I'd rather set that explicitly in the script too instead of defaulting to the crossplane image tag, also because with this change it wouldn't work by default, given that the tag is `""` now.

Right, I tottaly forgot that we are setting it explicitly in the script and thats why it works :facepalm: . Ill change it now

turkenh (approved):

PR #4031, opened 2023-04-28
Avoid listing CRDs in the RBAC manager

### LINKED ISSUE #4016: Stress Testing Crossplane with High Number of Packages

The design in #3939 proposes we break up larger providers like provider-aws by service. This would mean a typical Crossplane installation deploys more provider packages than before. See the proposal for more details.

We need to experiment with deploying a high number of packages with Crossplane and observe the behavior, including resource consumption.

We expect an average Crossplane deployment will run 10 providers on average see the [community impact](https://github.com/crossplane/crossplane/pull/3939/files#diff-c35d366039b2ed54517b98b83bc01e4e421bae2ea9ed3d0ee53a9231afcaf8d3R183).

As a means of stress testing, we can deploy 50-100 providers and observe the behavior of both `crossplane` pod (for package manager) and `crossplane-rbac-manager` including the changes [in this PR](https://github.com/crossplane/crossplane/pull/4010).

negz:

Fixes #4016

We found in https://github.com/crossplane/crossplane/issues/4016 that working with CRDs was quite expensive. It turns out we don't need to. In order to create RBAC roles we only need the plural and group of the types defined by a ProviderRevision (i.e. the types defined by CRDs installed by a ProviderRevision). We can derive this from the status of the ProviderRevision, which has references to all CRDs it creates.

My testing shows that this significantly reduces CPU and memory consumed by the RBAC manager when installing many (~70) providers from the same family at once.

I have:

I've confirmed that Crossplane creates RBAC cluster roles just like it did before this change, and that it does so while using much less CPU and memory. Previously installing ~70 providers from the same family all at once would cause the RBAC manager to hit its CPU and memory limits. Now I see it using ~50m CPU and 50Mi to 100Mi of memory even in the "worst case scenario" where provider package and controller images are pre-loaded, allowing ~70 providers to be created and ready in under 2 minutes.

Screenshot 2023-04-27 at 11 55 29 PM

Screenshot 2023-04-27 at 11 55 40 PM

Note that in the above screenshot the CPU limit is 100m (1/10th of a core) and the memory limit is 512Mi.

### DIFF
--- a/cmd/crossplane/core/core.go
+++ b/cmd/crossplane/core/core.go
@@ -18,6 +18,8 @@ limitations under the License.
 package core
 
 import (
+	"net/http"
+	"net/http/pprof"
 	"path/filepath"
 	"time"
 
@@ -46,6 +48,8 @@ import (
 	"github.com/crossplane/crossplane/internal/xpkg"
 )
 
+const pprofPath = "/debug/pprof/"
+
 // Command runs the core crossplane controllers
 type Command struct {
 	Start startCommand `cmd:"" help:"Start Crossplane controllers."`
@@ -67,6 +71,8 @@ func (c *Command) Run() error {
 }
 
 type startCommand struct {
+	Profile string `placeholder:"host:port" help:"Serve runtime profiling data via HTTP at /debug/pprof."`
+
 	Namespace            string `short:"n" help:"Namespace used to unpack and run packages." default:"crossplane-system" env:"POD_NAMESPACE"`
 	ServiceAccount       string `help:"Name of the Crossplane Service Account." default:"crossplane" env:"POD_SERVICE_ACCOUNT"`
 	CacheDir             string `short:"c" help:"Directory used for caching package images." default:"/cache" env:"CACHE_DIR"`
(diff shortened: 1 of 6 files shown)


cmd/crossplane/core/core.go
```diff
@@ -96,7 +102,34 @@ type startCommand struct {
...
+				ReadTimeout:  2 * time.Minute,
+				WriteTimeout: 2 * time.Minute,
+				Handler:      mux,
+			}
+			log.Debug("Starting server", "type", "pprof", "path", pprofPath, "addr", s.Addr)
+			err := s.ListenAndServe()
+			log.Debug("Profiling server has stopped listening", "error", err)
+		}()
+	}
+
```
negz, line 132: This is repeated here and in the RBAC manager. I considered adding it to crossplane-runtime as a helper, but so far it's pretty straightforward. I'd prefer to hold off for now, but we can add it there if we start propagating it elsewhere (e.g. to providers).

I also considered adding Helm chart plumbing to enable this, but I also would prefer not to. I see this as mostly a developer tool - not something that needs to be made particularly easy to turn on.

internal/controller/rbac/provider/roles/reconciler.go
```diff
@@ -296,27 +296,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 				continue
 			}
 
-			inFamily[member.GetUID()] = true
-		}
-	}
-
-	l := &extv1.CustomResourceDefinitionList{}
-	if err := r.client.List(ctx, l); err != nil {
```
negz, line 304: I thought removing this might mean the RBAC manager would not itself need RBAC access to get and list CRDs anymore. It turns out it does need them for its init container.

turkenh (approved):
I can also confirm that this improves resource consumption significantly, great work!

I also conducted some testing with provider packages both with/without the family label and observed no issues.

Issue #3903, opened 2023-03-23
`ControllerConfig` deprecation leads to log warning messages without which API is affected


haarchri:
### What happened?

since this change https://github.com/crossplane/crossplane/pull/3678  https://github.com/crossplane/crossplane/pull/3678/files#diff-982fd9d95958e72fca5b25e9e689ec838f0c532ac7a731989e206c91d99b0c12R175 and crossplane v1.11.x we see the following log messages in crossplane core - but it is not clear which API is affected 

we can change it to: `+kubebuilder:deprecatedversion:warning="ControllerConfig.pkg.crossplane.io/v1alpha1 is deprecated and is scheduled to be removed in a future release."`

```
W0323 14:35:26.462328       1 warnings.go:70] This API is deprecated and is scheduled to be removed in a future release.
W0323 14:41:11.463708       1 warnings.go:70] This API is deprecated and is scheduled to be removed in a future release.
W0323 14:51:09.465653       1 warnings.go:70] This API is deprecated and is scheduled to be removed in a future release.
W0323 14:58:16.466160       1 warnings.go:70] This API is deprecated and is scheduled to be removed in a future release.
W0323 15:05:02.469498       1 warnings.go:70] This API is deprecated and is scheduled to be removed in a future release.
W0323 15:13:12.471131       1 warnings.go:70] This API is deprecated and is scheduled to be removed in a future release.
W0323 15:19:42.474187       1 warnings.go:70] This API is deprecated and is scheduled to be removed in a future release.
W0323 15:25:03.476043       1 warnings.go:70] This API is deprecated and is scheduled to be removed in a future release.
W0323 15:30:27.478370       1 warnings.go:70] This API is deprecated and is scheduled to be removed in a future release.
W0323 15:38:43.480896       1 warnings.go:70] This API is deprecated and is scheduled to be removed in a future release.
W0323 15:48:05.482689       1 warnings.go:70] This API is deprecated and is scheduled to be removed in a future release.
W0323 15:56:37.484082       1 warnings.go:70] This API is deprecated and is scheduled to be removed in a future release.
W0323 16:01:46.486261       1 warnings.go:70] This API is deprecated and is scheduled to be removed in a future release.
W0323 16:08:54.487827       1 warnings.go:70] This API is deprecated and is scheduled to be removed in a future release.
W0323 16:15:42.489198       1 warnings.go:70] This API is deprecated and is scheduled to be removed in a future release.
```

### What environment did it happen in?
[...]

PR #4037, opened 2023-04-30
Add volume and volumeMount support to controllerConfig

### LINKED ISSUE #2535: Volume support in ControllerConfig

I need to use provider-aws and I need to override the AWS client's CA bundle (this can be done in `aws-sdk-go-v2` via `AWS_CA_BUNDLE=/path/to/pem` ... https://github.com/aws/aws-sdk-go-v2/blob/f5b68d198dbefe144696aa0051ed854571b5d614/config/env_config.go#L146-L162)
My issue is that there does not seem to be a way (without using mutating webhooks on the provider pod or something similar) to get a pem file mounted on the AWS provider (I looked in `ControllerConfig` but did not see anything for volume mounts).

If `ControllerConfig` supported adding volume mounts from `ConfigMap` or `Secret` I could configure the CA bundle reference and the `AWS_CA_BUNDLE` env all from the `ControllerConfig` definition.

Additional reference:
https://crossplane.slack.com/archives/C01718T2476/p1630066636007500

stevendborrelli:
Add `volume` and `volumeMount` to ControllerConfig. This allows providers to mount custom files into the provider pods, including x509 certificates. 

I understand this API is deprecated, but I am not sure when the PRI will be available. 

Fixes #2535

I have:

Testing process:

Create a configmap containing two keys:

```yaml
kubectl create configmap -n crossplane-system volume-test --from-literal=file.1=hello --from-literal=file.2=world
```

Create a  `controllerConfig` to mount the configMap: 

```yaml
kind: ControllerConfig
metadata:
  name: volume-test
spec:
  volumeMounts: 
      - name: volume-test
        mountPath: /etc/vols
  volumes:
    - name: volume-test
      configMap:
        name: volume-test
```

Install a test provider using the controllerconfig:

```yaml
cat <<EOF | kubectl apply -f -
kind: Provider
metadata:
  name: upbound-provider-aws
spec:
  package: xpkg.upbound.io/upbound/provider-aws:v0.27.0
  controllerConfigRef:
    name: volume-test
EOF
```

Verify volumes have been mounted in the provider pod: 

```console
✗ kubectl exec -it  -n crossplane-system upbound-provider-aws-b52ad4c74cbf-75b976c69f-f57q7 cat /etc/vols/file.1
hello                                                                                                 

✗ kubectl exec -it  -n crossplane-system upbound-provider-aws-b52ad4c74cbf-75b976c69f-f57q7 cat /etc/vols/file.2
world                                                  
```


hasheddan (commented):
I don't think we should be expanding this deprecated API, but I agree we need a direction forward on PRI. Perhaps we can dedicate a portion of this week's community meeting to discussing. cc @jbw976

marshmallory:
> I don't think we should be expanding this deprecated API, but I agree we need a direction forward on PRI. Perhaps we can dedicate a portion of this week's community meeting to discussing. cc @jbw976

We desperately need this for local development behind the corporate proxy.

turkenh:
This change looks reasonable to me given we don't have the alternative available. I don't see a reason to block this PR/functionality for a solution that wouldn't land in the short term.

ezgidemirel (approved):
LGTM, thanks @stevendborrelli

negz:
I'm sympathetic to the "death by 1,000 paper cuts" concern of continuing to expand the deprecated `ControllerConfig` API. On the other hand, we don't have any good alternative for folks who need this functionality at the moment.

* https://github.com/crossplane/crossplane/issues/2671

We believe the provider runtime interface (PRI) idea is likely to be the path forward. At this stage that's not on our roadmap, and hasn't moved from RFC to design phase so it _seems_ to me that it's a while off. I think the design has a number of benefits, but few of them seem very urgent (relative to everything else) except for allowing us to rethink ControllerConfig.

The other alternatives that might work (e.g. a Kyverno webhook) feel like they put us in an overly complicated situation for the time being - some controller fields are influenced by `ControllerConfig`, others are influenced by a webhook, etc.

All that said, I lean toward accepting this change.

negz:
If we did not accept this change, and _did_ find that a Kyverno webhook (or similar third party tool) would work in this case I'd like us to investigate using such a webhook for _all_ use cases that `ControllerConfig` currently satisfies. This would:

* Allow us to remove `ControllerConfig` without waiting for new functionality (i.e. PRI) to be designed and built.
* Ensure folks aren't stuck in a split tooling situation where some things are done using `ControllerConfig` and others done using webhooks.

ytsarev:
The current ControllerConfig deprecation caused the community more harm and confusion than good.
People are observing the ControllerConfig deprecation message and trying to find alternatives to configure the provider pod and can't find any.
IMHO, without the working alternative we should remove the deprecation message, add Volume support, and work on PRI implementation in an unblocking manner.

muvaf:
Seems like everyone is aligned, there is no blocker to merge this change and then follow up with PRI work. Merging.

> IMHO, without the working alternative we should remove the deprecation message, add Volume support, and work on PRI implementation in an unblocking manner.

Is there an issue for discussion of the removal of the message? I'd not like to increase the scope of this PR.

ytsarev:
Thanks a lot, @muvaf!

We had a chat with @jbw976, @negz, and other folks and agreed not to fully remove the deprecation message(to avoid creating even more long-term confusion) but to enhance the message (and potentially the messaging mechanism) in https://github.com/crossplane/crossplane/issues/3903

PR #4051, opened 2023-05-05
feat: Add `Replace` composition string transform

julienduchesne:
Allows compositions to search and replace any attribute before they are sent to the provider This is useful for attributes that represent large configs. You can replace deprecated fields, escape chars, etc

Let me know if the general idea works and I can document it at https://github.com/crossplane/docs!


phisco:
Thanks @julienduchesne, the code looks good! My only doubt is whether it would be worth having something like a `RegexpReplace` type instead, I feel like it would cover the same use-case, but be a bit more flexible. 
Currently the `Regexp` string transformer is only allowing extracting sub-strings, this would allow a `sed`-like usage instead, allowing to reuse captured groups in the replacement text and all. WDYT?

CC @negz

negz:
> My only doubt is whether it would be worth having something like a RegexpReplace type instead, I feel like it would cover the same use-case, but be a bit more flexible.

I'm torn. I generally prefer simpler APIs, which leads me to avoid regular expressions where possible. On the other hand, if we think folks are going to push to have a regular expression variant of this eventually it may make sense to just go straight to that rather than having `Replace` and `ReplaceRegex`.

Issue #4047, opened 2023-05-04
Use server-side apply


negz:
### What problem are you facing?

Most core Crossplane reconcilers use an [`Applicator`](https://pkg.go.dev/github.com/crossplane/crossplane-runtime@v0.19.2/pkg/resource#Applicator) to create and update resources in the API server. For example, we use an `Applicator` to keep claims synced with XRs, and to create or update an XR's composed resources.

An `Applicator` provides similar "upsert" semantics to `kubectl apply`. It allows you to simply "apply" your desired state - the `Applicator` then figures out whether the desired state affects a resource that already exists and if so how it should update it. When we began the Crossplane project the API server and Go Kubernetes client tooling did not support this - you had to either `Create` a new resource, or `Get`, mutate, and then `Update` an existing resource.

There are currently two implementations of the `Applicator` interface:

* https://pkg.go.dev/github.com/crossplane/crossplane-runtime@v0.19.2/pkg/resource#APIPatchingApplicator
* https://pkg.go.dev/github.com/crossplane/crossplane-runtime@v0.19.2/pkg/resource#APIUpdatingApplicator

Both implementations simply call the underlying Kubernetes client's `Create` method if the relevant resource doesn't exist. Where they differ is what they do to update existing resources.

The `APIPatchingApplicator` uses the underlying client's `Patch` method. It creates a [JSON merge patch](https://datatracker.ietf.org/doc/html/rfc7386) where the patch is simply the entire new, desired resource. This has a few downsides:

* Patches will never remove a field from an object; they're only additive.
* Arrays are always replaced, not merged.

The `APIUpdatingApplicator` uses the underlying client's `Update` method. It circumvents the typical read, mutate, update approach by simply passing the new desired state to `Update`. Typically the `Update` method guards against this by using the passed resource's `meta.resourceVersion` as a form of eventual consistency. It expects to be passed a resource version equal to the latest version passed in the API server, to ensure clients don't accidentally revert state by writing an update based on a stale read. This is the major issue with the `APIUpdatingApplicator` - it manually sets the resource version to the latest read from the API server even if the state its passing is from an older version. This leads to the `APIUpdatingApplicator` unintentionally reverting changes made by other controllers.

Currently neither `Applicator` works for all situations. This results in us picking the least-bad option for a particular use case. We suspect these applicators __may__ be the cause of several issues in Crossplane, including:

```[tasklist]
### Related issues
- [ ] https://github.com/crossplane/crossplane/issues/5498
[...]

PR #4054, opened 2023-05-06
Refactor in-cluster auth into a `ContainerFunctionRunnerOption`

negz:

This reduces the complexity of RunFunction a little. This PR also has a few small cleanups following on to https://github.com/crossplane/crossplane/pull/3869.

I have:

https://github.com/crossplane-contrib/contribfest/blob/63c50e/lab-composition-functions/02-xfn-noop.md

I used a variant of the above lab, modified to use a private function (`index.docker.io/negz/private-fn:v0.0.1`). I confirmed that:

* When I specify a valid `imagePullSecret` everything works as expected.
* When I specify a non-existent `imagePullSecret` I get an error (see below).
* When I don't specify an `imagePullSecret` I get another error (see below).

```
  Warning  ComposeResources         23s (x6 over 53s)      defined/compositeresourcedefinition.apiextensions.crossplane.io  cannot compose resources: cannot run Composition Function pipeline: cannot run function "my-noop-function": cannot run container: rpc error: code = Unknown desc = exit status 1: xfn: error: spark.Command.Run(): cannot pull OCI image: cannot pull image from remote: GET https://index.docker.io/v2/negz/private-fn/manifests/v0.0.1: UNAUTHORIZED: authentication required; [map[Action:pull Class: Name:negz/private-fn Type:repository]]
```

```
  Warning  ComposeResources         0s (x2 over 0s)     defined/compositeresourcedefinition.apiextensions.crossplane.io  cannot compose resources: cannot run Composition Function pipeline: cannot run function "my-noop-function": cannot apply run function option: cannot get image pull secret: Secret "does-not-exist" not found
```

### DIFF
--- a/apis/apiextensions/v1/composition_common.go
+++ b/apis/apiextensions/v1/composition_common.go
@@ -286,14 +286,14 @@ type ContainerFunction struct {
 	// +kubebuilder:validation:Enum="IfNotPresent";"Always";"Never"
 	ImagePullPolicy *corev1.PullPolicy `json:"imagePullPolicy,omitempty"`
 
+	// ImagePullSecrets are used to pull images from private OCI registries.
+	// +optional
+	ImagePullSecrets []corev1.LocalObjectReference `json:"imagePullSecrets,omitempty"`
+
 	// Timeout after which the Composition Function will be killed.
 	// +optional
 	// +kubebuilder:default="20s"
 	Timeout *metav1.Duration `json:"timeout,omitempty"`
-
-	// Secrets for pulling function images.
-	// +optional
-	ImagePullSecrets []corev1.LocalObjectReference `json:"imagePullSecrets,omitempty"`
 
 	// Network configuration for the Composition Function.
 	// +optional
--- a/apis/apiextensions/v1beta1/zz_generated.deepcopy.go
+++ b/apis/apiextensions/v1beta1/zz_generated.deepcopy.go
@@ -284,15 +284,15 @@ func (in *ContainerFunction) DeepCopyInto(out *ContainerFunction) {
 		*out = new(v1.PullPolicy)
 		**out = **in
 	}
-	if in.Timeout != nil {
-		in, out := &in.Timeout, &out.Timeout
-		*out = new(metav1.Duration)
-		**out = **in
-	}
 	if in.ImagePullSecrets != nil {
 		in, out := &in.ImagePullSecrets, &out.ImagePullSecrets
 		*out = make([]v1.LocalObjectReference, len(*in))
 		copy(*out, *in)
+	}
+	if in.Timeout != nil {
+		in, out := &in.Timeout, &out.Timeout
+		*out = new(metav1.Duration)
+		**out = **in
 	}
 	if in.Network != nil {
 		in, out := &in.Network, &out.Network
--- a/internal/controller/apiextensions/composite/composition_ptf.go
+++ b/internal/controller/apiextensions/composite/composition_ptf.go
@@ -31,7 +31,6 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/json"
-	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/yaml"
 
@@ -60,6 +59,8 @@ const (
 	errMarshalXR                = "cannot marshal composite resource"
 	errMarshalCD                = "cannot marshal composed resource"
 	errParseImage               = "cannot parse image reference"
+	errGetServiceAccount        = "cannot get image pull secrets from service account"
+	errGetImagePullSecret       = "cannot get image pull secret"
 	errNewKeychain              = "cannot create a new in-cluster registry authentication keychain"
 	errResolveKeychain          = "cannot resolve in-cluster registry authentication keychain"
 	errAuthCfg                  = "cannot get in-cluster registry authentication credentials"
@@ -636,30 +637,42 @@ func FunctionIODesired(s *PTFCompositionState) (iov1alpha1.Desired, error) {
 }
 
 // A ContainerFunctionRunnerOption modifies the behavior of a
-// ContainerFunctionRunner, typically by mutating the RunFunctionRequest it
-// produces.
+// ContainerFunctionRunner by mutating the supplied RunFunctionRequest.
 type ContainerFunctionRunnerOption func(ctx context.Context, fn *v1.ContainerFunction, r *fnv1alpha1.RunFunctionRequest) error
 
-// WithInClusterAuth configures a ContainerFunctionRunner to use in-cluster
-// OCI registry authentication. Image pull secrets will be loaded from the
-// supplied namespace and service account.
-func WithInClusterAuth(namespace, serviceAccount string) ContainerFunctionRunnerOption {
+// WithKubernetesAuthentication configures a ContainerFunctionRunner to use
+// "Kubernetes" authentication to pull images from a private OCI registry.
+// Kubernetes authentication emulates how the Kubelet pulls private images.
+// Specifically, it:
+//
+// 1. Loads credentials from the Docker config file, if any.
+// 2. Loads credentials from the supplied service account's image pull secrets.
+// 3. Loads credentials from the function's image pull secrets.
+// 4. Loads credentials using the GKE, EKS, or AKS credentials helper.
+func WithKubernetesAuthentication(c client.Reader, namespace, serviceAccount string) ContainerFunctionRunnerOption {
 	return func(ctx context.Context, fn *v1.ContainerFunction, r *fnv1alpha1.RunFunctionRequest) error {
-		ko := k8schain.Options{
-			Namespace:          namespace,
-			ServiceAccountName: serviceAccount,
-			ImagePullSecrets:   make([]string, 0, len(fn.ImagePullSecrets)),
+
+		sa := &corev1.ServiceAccount{}
+		if err := c.Get(ctx, types.NamespacedName{Namespace: namespace, Name: serviceAccount}, sa); err != nil {
+			return errors.Wrap(err, errGetServiceAccount)
 		}
 
-		for i := range fn.ImagePullSecrets {
-			ko.ImagePullSecrets[i] = fn.ImagePullSecrets[i].Name
+		refs := make([]corev1.LocalObjectReference, 0, len(fn.ImagePullSecrets)+len(sa.ImagePullSecrets))
+		refs = append(refs, sa.ImagePullSecrets...)
+		refs = append(refs, fn.ImagePullSecrets...)
+
+		ips := make([]corev1.Secret, len(refs))
+		for i := range refs {
+			if err := c.Get(ctx, types.NamespacedName{Namespace: namespace, Name: refs[i].Name}, &ips[i]); err != nil {
+				return errors.Wrap(err, errGetImagePullSecret)
+			}
 		}
 
-		keychain, err := k8schain.NewInCluster(ctx, ko)
-		if errors.Is(err, rest.ErrNotInCluster) {
-			// We're not running in a cluster. Nothing to do.
-			return nil
-		}
+		// We use NewFromPullSecrets rather than NewInCluster so that we can use
+		// our own client.Reader rather than the default in-cluster client. This
+		// makes the option easier to test, ensures we re-use any existing
+		// client caches, and allows us to run out-of-cluster.
+		keychain, err := k8schain.NewFromPullSecrets(ctx, ips)
 		if err != nil {
 			return errors.Wrap(err, errNewKeychain)
 		}
--- a/internal/controller/apiextensions/composite/composition_ptf.go
+++ b/internal/controller/apiextensions/composite/composition_ptf.go
@@ -19,7 +19,6 @@ import (
 	"context"
 	"sort"
 
-	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/google/go-containerregistry/pkg/authn/k8schain"
 	"github.com/google/go-containerregistry/pkg/name"
 	"google.golang.org/grpc"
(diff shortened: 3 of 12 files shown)


negz:
CC @AndrewChubatiuk.

internal/controller/apiextensions/composite/composition_ptf.go
```diff
@@ -634,30 +635,87 @@ func FunctionIODesired(s *PTFCompositionState) (iov1alpha1.Desired, error) {
...
+			Namespace:          namespace,
+			ServiceAccountName: serviceAccount,
+			ImagePullSecrets:   make([]string, 0, len(fn.ImagePullSecrets)),
+		}
+
+		for i := range fn.ImagePullSecrets {
+			ko.ImagePullSecrets[i] = fn.ImagePullSecrets[i].Name
+		}
+
+		keychain, err := k8schain.NewInCluster(ctx, ko)
```
negz, line 658: I started thinking about how to unit test this functionality, and now I'm questioning whether:

* We should use `k8schain.NewFromPullSecrets` instead of `k8schain.NewInCluster`.
* Taking an array of `imagePullSecrets` for each function is a good API in the first place.

For context, `k8schain.NewInCluster` attempts to emulate how the Kubelet authenticates to a private registry. It does this by:

1. Creating a new Kubernetes API client that supports _only_ in-cluster authentication.
2. Fetching all of the supplied [service account's image pull secrets](https://kubernetes.io/docs/concepts/containers/images/#specifying-imagepullsecrets-on-a-pod).
3. Fetching alll of the explicitly named secrets (from within the supplied namespace).

It chains this Kubernetes-secret based authentication with several other credential sources - namely the Docker config file, and credential helpers for GKE, EKS, and AKS.

I believe it would be ideal to fetch any secrets using the same Kubernetes client as the controller manager. We could do this by fetching the secrets in advance, then using `k8schain.NewFromPullSecrets`. This would bring some minor improvements - we'd share client caches, and would be able to load secrets even if Crossplane wasn't running "in-cluster" (i.e. wasn't running as a pod in the same cluster as its API server). I believe it would also make the functionality easier to test, because we'd be injecting secrets rather than having the keychain create and try to use its own Kubernetes client.

This got me thinking about whether `imagePullSecrets` is the best API for us. Specifically what I wonder is:

* How often do folks actually want to specify secrets on a per-function basis? Would just using the secrets associated with the Crossplane pod's service account be sufficient?
* If we do want to keep specifying secrets on a per-function basis, should we continue to assume those secrets will exist in the same namespace as the Crossplane pod? Elsewhere when we reference secrets from a cluster-scoped resource we typically require the namespace of each secret to be specified explicitly - e.g. in an MR or XR's `writeConnectionSecretToRef`.
* Could we eventually want to load image pull secrets from some other secret store (e.g. Vault)?

I think the key reason to stick with `imagePullSecrets` is that we've already committed to it in the package manager API (as `packagePullSecrets`), though all my thoughts above apply to that API too.

negz: > How often do folks actually want to specify secrets on a per-function basis? Would just using the secrets associated with the Crossplane pod's service account be sufficient?

I suppose one approach here would be to not add the new `imagePullSecrets` field (or rather, remove it from the master branch before it gets released for the first time with v1.13) and instead support only service-account secrets for pulling functions. We could always add support for specifying secrets at the per-function level if/when there was demand for it.

hasheddan: > I believe it would be ideal to fetch any secrets using the same Kubernetes client as the controller manager. We could do this by fetching the secrets in advance, then using k8schain.NewFromPullSecrets. This would bring some minor improvements - we'd share client caches, and would be able to load secrets even if Crossplane wasn't running "in-cluster" (i.e. wasn't running as a pod in the same cluster as its API server). I believe it would also make the functionality easier to test, because we'd be injecting secrets rather than having the keychain create and try to use its own Kubernetes client.

This part all sounds good to me 👍🏻 When the initial package pull secrets functionality was introduced the `go-containerregistry` API looked very different and it was somewhat more difficult to do this. I'd love to see these improvements made across both composition functions and the package manager.

> How often do folks actually want to specify secrets on a per-function basis? Would just using the secrets associated with the Crossplane pod's service account be sufficient?

Fairly indifferent to this one -- the rationale for supporting per-package secrets was less about isolation and more about UX around not having to edit the SA. Since composition functions haven't committed to this model, its worth evaluating whether it is necessary here.

> If we do want to keep specifying secrets on a per-function basis, should we continue to assume those secrets will exist in the same namespace as the Crossplane pod? Elsewhere when we reference secrets from a cluster-scoped resource we typically require the namespace of each secret to be specified explicitly - e.g. in an MR or XR's writeConnectionSecretToRef.

This is arguably an unnecessary restriction, but packages and functions are a little different than other resources that reference secrets in that they are more about configuring how the "crossplane system" works. I haven't heard anyone complain about not being able to put secrets elsewhere, but that doesn't mean folks wouldn't prefer it.

> Could we eventually want to load image pull secrets from some other secret store (e.g. Vault)?

This sounds awesome -- I would highly endorse it.

turkenh: > > If we do want to keep specifying secrets on a per-function basis, should we continue to assume those secrets will exist in the same namespace as the Crossplane pod? Elsewhere when we reference secrets from a cluster-scoped resource we typically require the namespace of each secret to be specified explicitly - e.g. in an MR or XR's writeConnectionSecretToRef.
> 
> This is arguably an unnecessary restriction, but packages and functions are a little different than other resources that reference secrets in that they are more about configuring how the "crossplane system" works. I haven't heard anyone complain about not being able to put secrets elsewhere, but that doesn't mean folks wouldn't prefer it.

I have some concerns around multi-tenancy where a Crossplane instance is shared with different teams and the need for editing a shared XP ServiceAccount. However, for composition functions, it definitely makes sense to at least start with this, and consider extending it if needed.

In general, I do like the clarification we would get from categorizing these as _"about configuring how the "crossplane system" works"_ and providing a consistent experience.

> Could we eventually want to load image pull secrets from some other secret store (e.g. Vault)?

This sounds great, and the proposed changes here would prepare the code for this. However, not sure how users would configure this if we expect the pull secrets to be provided via the ServiceAccount. Wouldn't this require us to have an API for this similar to what we would like to drop now?

negz: > However, not sure how users would configure this if we expect the pull secrets to be provided via the ServiceAccount.

Right. To be clear, I agree with your statement that we'd need to introduce some new API to support loading secrets from a secret store. That would just become a "then" problem rather than a "now" problem. 😄 

> Wouldn't this require us to have an API for this similar to what we would like to drop now?

Yes I think it would, but the key word is _similar_.

If we add `imagePullSecrets` today as an array of `LocalObjectReference` it's possible we might find the API limiting tomorrow when we want to support loading these secrets from a different secret store. That said:

* Right now this is an alpha API, so we're not making that big a commitment.
* The current `publishConnectionDetailsTo` object (w/ `configRef`, `metadata`, and `name`) is actually backward compatible with `LocalObjectReference`.

negz: > That said

I talked myself back around to just leaving `imagePullSecrets` in for now, for the reasons in my most recent comment. 😄

ezgidemirel (approved):
Thanks @negz, LGTM!

apis/apiextensions/v1/composition_common.go
```diff
@@ -286,15 +286,15 @@ type ContainerFunction struct {
 	// +kubebuilder:validation:Enum="IfNotPresent";"Always";"Never"
 	ImagePullPolicy *corev1.PullPolicy `json:"imagePullPolicy,omitempty"`
 
+	// ImagePullSecrets are used to pull images from private OCI registries.
+	// +optional
+	ImagePullSecrets []corev1.LocalObjectReference `json:"imagePullSecrets,omitempty"`
```
pedjak, line 291: do we want to specify here patchStrategy and patchMergeKey, similar to what is defined in `PodSpec`?

apis/apiextensions/v1beta1/zz_generated.deepcopy.go
```diff
@@ -284,16 +284,16 @@ func (in *ContainerFunction) DeepCopyInto(out *ContainerFunction) {
...
-		in, out := &in.Timeout, &out.Timeout
-		*out = new(metav1.Duration)
-		**out = **in
-	}
 	if in.ImagePullSecrets != nil {
 		in, out := &in.ImagePullSecrets, &out.ImagePullSecrets
 		*out = make([]v1.LocalObjectReference, len(*in))
 		copy(*out, *in)
 	}
+	if in.Timeout != nil {
```
pedjak, line 292: this change looks like no-op, the if block was removed from line 287 and added back to line 292. Can we revert it?

internal/controller/apiextensions/composite/composition_ptf.go
```diff
@@ -634,30 +636,99 @@ func FunctionIODesired(s *PTFCompositionState) (iov1alpha1.Desired, error) {
...
+// WithKubernetesAuthentication configures a ContainerFunctionRunner to use
+// "Kubernetes" authentication to pull images from a private OCI registry.
+// Kubernetes authentication emulates how the Kubelet pulls private images.
+// Specifically, it:
+//
+// 1. Loads credentials from the Docker config file, if any.
+// 2. Loads credentials from the supplied service account's image pull secrets.
+// 3. Loads credentials from the function's image pull secrets.
+// 4. Loads credentials using the GKE, EKS, or AKS credentials helper.
+func WithKubernetesAuthentication(c client.Reader, namespace, serviceAccount string) ContainerFunctionRunnerOption {
```
pedjak, line 652: do we need some unit tests for the function?

internal/controller/apiextensions/composite/composition_ptf_test.go
```diff
@@ -1111,6 +1111,114 @@ func TestFunctionIODesired(t *testing.T) {
...
+			want: want{
+				err: errors.Wrap(errBoom, errGetImagePullSecret),
+			},
+		},
+		"Success": {
+			reason: "We should successfully parse OCI registry authentication credentials from an image pull secret.",
+			params: params{
+				c: &test.MockClient{
+					MockGet: test.NewMockGetFn(nil, func(obj client.Object) error {
+						if s, ok := obj.(*corev1.Secret); ok {
```
pedjak, line 1176: we should check for secret name and return error if it is not equal to `cool-secret`. Otherwise, the test would pass even if `WithKubernetesAuthentication` has bug, i.e. it reads an arbitrary secret, independently what was passed argument.

internal/controller/apiextensions/composite/composition_ptf_test.go
```diff
@@ -1111,6 +1111,114 @@ func TestFunctionIODesired(t *testing.T) {
...
+		err error
+	}
+
+	cases := map[string]struct {
+		reason string
+		params params
+		args   args
+		want   want
+	}{
+		"GetServiceAccountError": {
```
pedjak, line 1138: `WithKubernetesAuthentication` can return 5 different error types, but we test only for two here. Could we add tests for the rest of them?

internal/controller/apiextensions/composite/composition_ptf_test.go
```diff
@@ -52,9 +52,9 @@ import (
 	"github.com/crossplane/crossplane-runtime/pkg/resource/unstructured/composite"
 	"github.com/crossplane/crossplane-runtime/pkg/test"
 
-	"github.com/crossplane/crossplane/apis/apiextensions/fn/io/v1alpha1"
 	iov1alpha1 "github.com/crossplane/crossplane/apis/apiextensions/fn/io/v1alpha1"
 	fnpbv1alpha1 "github.com/crossplane/crossplane/apis/apiextensions/fn/proto/v1alpha1"
+	fnv1alpha1 "github.com/crossplane/crossplane/apis/apiextensions/fn/proto/v1alpha1"
```
pedjak, line 57: do we need to import the same package twice, under two different aliases?

negz: No, this is an oversight. Will fix.

apis/apiextensions/v1/composition_common.go
negz: Can you provide more context about what this does and why we'd do it?

I believe you're asking whether we should add these as comment markers / struct tags, and I think the answer is no. My understanding is that these comment markers control strategic merge patching, which isn't supported by CRDs. See:
* https://github.com/kubernetes-sigs/controller-runtime/blob/116a1b8/pkg/client/patch.go#LL172C4-L172C123
* https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/#advanced-features-and-flexibility

Strategic merge patches are deprecated in favor of [server-side-apply](https://kubernetes.io/docs/reference/using-api/server-side-apply/), which does support CRDs but isn't something we support yet in Crossplane per https://github.com/crossplane/crossplane/issues/4047.

apis/apiextensions/v1beta1/zz_generated.deepcopy.go
negz: This is a generated file. No idea why the deepcopy code generator decided to make this re-ordering but I think we should let it. If we don't it's just going to try make this change again next time someone runs `go generate`.

internal/controller/apiextensions/composite/composition_ptf_test.go
negz: I considered it when I was writing this test. I'm inclined not to. Triggering the three errors we're not covering right now would require us to either:

* "Test through" to the underlying implementations, e.g. by providing a reference that causes `name.ParseReference` to fail.
* Refactor the code under test to make it possible to inject all the go-containerregistry dependencies (like `name.ParseReference`) with fake implementations.

Generally I do prefer to get full coverage, but in this case I feel the code complexity required to make all these dependencies injectable doesn't really warrant the increased test coverage.

internal/controller/apiextensions/composite/composition_ptf_test.go
negz: Good idea, will do.

internal/controller/apiextensions/composite/composition_ptf_test.go
pedjak: is it possible to catch up this with some linting rule in the future?

Issue #3742, opened 2023-02-06
Improved revision resource ownership transition


hasheddan:
### What problem are you facing?

When a `Provider` or `Configuration` is updated, a new `ProviderRevision` / `ConfigurationRevision` is created and ownership of the resources that are installed by both revisions is transitioned from the old revision to the new one. However, if the old revision is unable to go through a successful reconciliation with status `Inactive`, its controller references on the resources may not be transitioned to owner references. This prevents the new revision from being installed successfully as it cannot gain control of all resources. This commonly occurs if an old revision is deleted before successful transition.

### How could Crossplane help solve your problem?

A simple solution here would be to include the controller --> owner transition in the [deletion path](https://github.com/crossplane/crossplane/blob/0162f73e10cdb6fb9e24ba3713fcd7f02f4326b0/internal/controller/pkg/revision/reconciler.go#L354), rather than just the `Inactive` path. This would mean that we would refuse to remove the finalizer on a revision until it successfully transitioned controller --> owner. If it was unable to do so, an appropriate error would be surfaced to users, rather than just the somewhat opaque "cannot establish control" that users typically see today on the new revision today.

haarchri:
@hasheddan  thanks for the ref - was not aware of this issue ;)

hasheddan:
@haarchri sure thing! Also with the Crossplane / UXP issue, we should absolutely get that sorted out, but in the mean time you can always use https://doc.crds.dev/github.com/crossplane/crossplane/pkg.crossplane.io/Configuration/v1@v1.11.3#spec-ignoreCrossplaneConstraints if needed 👍🏻

turkenh:
> xref [#3423 (comment)](https://github.com/crossplane/crossplane/issues/3423#issuecomment-1495717622)

I could reproduce the above issue with a slightly simpler flow only with a provider (thanks to @haarchri's detailed one):

1. Install Crossplane v1.11.0

```bash
helm upgrade --install crossplane --namespace crossplane-system crossplane-stable/crossplane --version v1.11.0 --create-namespace
```

2. Deploy a modified version of provider-helm (`v0.3423.0`) that has a strict dependency to Crossplane `v1.11.0`.

```
spec:
  crossplane:
    version: "v1.11.0"
```

```bash
echo "apiVersion: pkg.crossplane.io/v1
kind: Provider
metadata:
  name: provider-helm
spec:
[...]

PR #4067, opened 2023-05-12
Improve package dependency conflict error message

### LINKED ISSUE #4022: Improve package dependency conflict error message

I recently tried to install a provider with a package dependency on another provider. The package manager automatically tries to install the dependency if it does not exist. It names the new provider `{org}-{name}`. In this case `upbound-provider-gcp-config`.

I had accidentally already installed that same provider with a slightly different name - just `provider-gcp-config`. This caused a conflict - the package manager wasn't able to install the provider I wanted to install, or its dependency. This wasn't obvious to me at all due to the obscure error message it returned:

```
11s         Warning   ResolveDependencies      providerrevision/provider-gcp-datastore-dc66af0637b3             cannot resolve package dependencies: node already exists
```

Presuming this situation can only occur when a dependency conflicts with an existing package, it would be helpful for the error to state that.

turkenh:

Under [certain circumstances](https://github.com/crossplane/crossplane/issues/3742#issuecomment-1543843655), the Crossplane package manager may end up in a situation where the lock file contains duplicate entries resulting in failures in dependency resolution. 

This PR improves the error message we are getting in that case by:

- Adding more context indicating that it is an issue related to initializing dependency graph using the lock.
- Including the identifier of the node to point the failing package.

For example, it changes from

```
cannot resolve package dependencies: node already exists
```

to

```
cannot resolve package dependencies: cannot initialize dependency graph from the packages in the lock: node xpkg.upbound.io/hasan/provider-helm already exists
```

Fixes #4022

I have:

By reproducing inconsistent state by following step here: https://github.com/crossplane/crossplane/issues/3742#issuecomment-1543800668

### DIFF
--- a/internal/controller/pkg/revision/dependency.go
+++ b/internal/controller/pkg/revision/dependency.go
@@ -40,6 +40,7 @@ const (
 
 	errNotMeta                   = "meta type is not a valid package"
 	errGetOrCreateLock           = "cannot get or create lock"
+	errInitDAG                   = "cannot initialize dependency graph from the packages in the lock"
 	errIncompatibleDependencyFmt = "incompatible dependencies: %+v"
 	errMissingDependenciesFmt    = "missing dependencies: %+v"
 	errDependencyNotInGraph      = "dependency is not present in graph"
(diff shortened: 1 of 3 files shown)


phisco (approved):

jbw976 (approved):

hasheddan (approved):
Thanks @turkenh!

PR #4071, opened 2023-05-12
Deactivate inactive revisions without pulling/parsing the package contents

### LINKED ISSUE #3742: Improved revision resource ownership transition

### LINKED ISSUE #3942: Failed to fetch packages for Provider/Configuration old revisions
### What happened?
Crossplane tries to pull **old** `ProviderRevision` and `ConfigurationRevision` even if they are in state `Inactive`.
Is it intentional that Crossplane keeps trying to pull old revisions from the remote registry? 

#### Background: 
We see in the events many `Parser backend cannot be initialized: Package could not be fetched from remote` messages because the package version no longer exists in the remote registry due to a container image registry (Azure Container registry) cleanup. 

### How can we reproduce it?

Deploy at least two different Configuration (or Provider) package versions one after another. Try to delete the first version from the remote registry and monitor the events. You will find something like:

```
# events
default                   1s          Warning   ParsePackage                      configurationrevision/iac-platform-configuration-32e7d4450782         cannot initialize parser backend: failed to fetch package from remote: GET https://xx.azurecr.io/v2/configurations/iac-platform-configuration/manifests/1.3.21: MANIFEST_UNKNOWN: manifest tagged by "1.3.21" is not found; map[Tag:1.3.21]
default                   0s          Warning   ParsePackage                      configurationrevision/iac-platform-configuration-5447be2ca321         cannot initialize parser backend: failed to fetch package from remote: GET https://xxx.azurecr.io/v2/configurations/iac-platform-configuration/manifests/1.3.11: MANIFEST_UNKNOWN: manifest tagged by "1.3.11" is not found; map[Tag:1.3.11]
...
default             2m5s        Warning   ParsePackage                      providerrevision/provider-jet-github-ddd1c3ffa7fe                     cannot initialize parser backend: failed to fetch package from remote: GET https://xxx.azurecr.io/v2/acp/provider-jet-github/manifests/0.1.5: MANIFEST_UNKNOWN: manifest tagged by "0.1.5" is not found; map[Tag:0.1.5]

```

Crossplane version: 1.11.2
Kubernetes Server version: `Server Version: version.Info{Major:"1", Minor:"22+", GitVersion:"v1.22.17-eks-48e63af", GitCommit:"47b89ea2caa1f7958bc6539d6865820c86b4bf60", GitTreeState:"clean", BuildDate:"2023-01-24T09:34:06Z", GoVersion:"go1.16.15", Compiler:"gc", Platform:"linux/amd64"}`

turkenh:

This PR updates the package revision reconciler not to rely on package content (and all other steps including dependency resolution) during deactivation of revisions to improve the reliability against edge cases where it is no longer possible to successfully reconcile a previously installed revision. 

We do this by relying the fact that package revisions persist a list of owned resources in their status and we don't actually need the contents of the package once we leverage that. We still fallback to parsing the package contents as it is not ideal to rely on existence of the status sub-resource (e.g. status may got lost after a backup/restore operation), even though we expect this to be fairly infrequent.

Fixes #3742
Fixes #3942

This PR changing the Package Revision Reconciler as below:

Package Revision Reconciler (2)

Also [link to the recording](https://www.youtube.com/watch?v=EsJRj5ZRnGY&list=PL510POnNVaaYYYDSICFSNWFqNbx1EMr-M&t=828s) from the community meeting where I talked about this PR.

I have:

### DIFF
--- a/internal/controller/pkg/revision/establisher.go
+++ b/internal/controller/pkg/revision/establisher.go
@@ -27,6 +27,7 @@ import (
 	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/pointer"
@@ -59,6 +60,7 @@ const (
 // resources and then establishing it.
 type Establisher interface {
 	Establish(ctx context.Context, objects []runtime.Object, parent v1.PackageRevision, control bool) ([]xpv1.TypedReference, error)
+	Relinquish(ctx context.Context, parent v1.PackageRevision) error
 }
 
 // NewNopEstablisher returns a new NopEstablisher.
@@ -72,6 +74,11 @@ type NopEstablisher struct{}
 // Establish does nothing.
 func (*NopEstablisher) Establish(_ context.Context, _ []runtime.Object, _ v1.PackageRevision, _ bool) ([]xpv1.TypedReference, error) {
 	return nil, nil
+}
+
+// Relinquish does nothing.
+func (*NopEstablisher) Relinquish(_ context.Context, _ v1.PackageRevision) error {
+	return nil
 }
 
 // APIEstablisher establishes control or ownership of resources in the API
@@ -115,6 +122,47 @@ func (e *APIEstablisher) Establish(ctx context.Context, objs []runtime.Object, p
 		return nil, err
 	}
 	return resourceRefs, nil
+}
+
+// Relinquish removes control or ownership of resources in the API server for a
+// package revision.
+func (e *APIEstablisher) Relinquish(ctx context.Context, parent v1.PackageRevision) error {
+	// Note(turkenh): We rely on status.objectRefs to get the list of objects
+	// that are controlled by the package revision. Relying on the status is
+	// not ideal as it might get lost (e.g. if the status subresource is
+	// not properly restored after a backup/restore operation). However, we will
+	// handle this by conditionally fetching/parsing package if there is no
+	// referenced resources available and rebuilding the status.
+	// In the next reconciliation loop, and we will be able to remove the
+	// control/ownership of the objects using the new status.
+	objRefs := parent.GetObjects()
+	// Stop controlling objects.
+	for _, ref := range objRefs {
+		u := unstructured.Unstructured{}
+		u.SetAPIVersion(ref.APIVersion)
+		u.SetKind(ref.Kind)
+		u.SetName(ref.Name)
+
+		if err := e.client.Get(ctx, types.NamespacedName{Name: u.GetName()}, &u); err != nil {
+			if kerrors.IsNotFound(err) {
+				// This is not expected, but still not an error for relinquishing.
+				continue
+			}
+			return errors.Wrap(err, "cannot get controlled object")
+		}
+		ors := u.GetOwnerReferences()
+		for i := range ors {
+			if ors[i].UID == parent.GetUID() {
+				ors[i].Controller = pointer.Bool(false)
+				break
+			}
+		}
+		u.SetOwnerReferences(ors)
+		if err := e.client.Update(ctx, &u); err != nil {
+			return errors.Wrap(err, "cannot update controlled object")
+		}
+	}
+	return nil
 }
 
 func (e *APIEstablisher) addLabels(objs []runtime.Object, parent v1.PackageRevision) error {
--- a/internal/controller/pkg/revision/establisher.go
+++ b/internal/controller/pkg/revision/establisher.go
@@ -62,7 +62,7 @@ const (
 // resources and then establishing it.
 type Establisher interface {
 	Establish(ctx context.Context, objects []runtime.Object, parent v1.PackageRevision, control bool) ([]xpv1.TypedReference, error)
-	Relinquish(ctx context.Context, parent v1.PackageRevision) error
+	ReleaseObjects(ctx context.Context, parent v1.PackageRevision) error
 }
 
 // NewNopEstablisher returns a new NopEstablisher.
@@ -78,8 +78,8 @@ func (*NopEstablisher) Establish(_ context.Context, _ []runtime.Object, _ v1.Pac
 	return nil, nil
 }
 
-// Relinquish does nothing.
-func (*NopEstablisher) Relinquish(_ context.Context, _ v1.PackageRevision) error {
+// ReleaseObjects does nothing.
+func (*NopEstablisher) ReleaseObjects(_ context.Context, _ v1.PackageRevision) error {
 	return nil
 }
 
@@ -126,9 +126,9 @@ func (e *APIEstablisher) Establish(ctx context.Context, objs []runtime.Object, p
 	return resourceRefs, nil
 }
 
-// Relinquish removes control of owned resources in the API server for a package
-// revision.
-func (e *APIEstablisher) Relinquish(ctx context.Context, parent v1.PackageRevision) error {
+// ReleaseObjects removes control of owned resources in the API server for a
+// package revision.
+func (e *APIEstablisher) ReleaseObjects(ctx context.Context, parent v1.PackageRevision) error {
 	// Note(turkenh): We rely on status.objectRefs to get the list of objects
 	// that are controlled by the package revision. Relying on the status is
 	// not ideal as it might get lost (e.g. if the status subresource is
--- a/internal/controller/pkg/revision/establisher.go
+++ b/internal/controller/pkg/revision/establisher.go
@@ -53,6 +53,8 @@ const (
 	errConversionWithNoWebhookCA    = "cannot deploy a CRD with webhook conversion strategy without having a TLS bundle"
 	errGetWebhookTLSSecret          = "cannot get webhook tls secret"
 	errWebhookSecretWithoutCABundle = "the value for the key tls.crt cannot be empty"
+	errFmtGetOwnedObject            = "cannot get owned object: %s/%s"
+	errFmtUpdateOwnedObject         = "cannot update owned object: %s/%s"
 )
 
 // An Establisher establishes control or ownership of a set of resources in the
@@ -124,8 +126,8 @@ func (e *APIEstablisher) Establish(ctx context.Context, objs []runtime.Object, p
 	return resourceRefs, nil
 }
 
-// Relinquish removes control or ownership of resources in the API server for a
-// package revision.
+// Relinquish removes control of owned resources in the API server for a package
+// revision.
 func (e *APIEstablisher) Relinquish(ctx context.Context, parent v1.PackageRevision) error {
 	// Note(turkenh): We rely on status.objectRefs to get the list of objects
 	// that are controlled by the package revision. Relying on the status is
@@ -148,7 +150,7 @@ func (e *APIEstablisher) Relinquish(ctx context.Context, parent v1.PackageRevisi
 				// This is not expected, but still not an error for relinquishing.
 				continue
 			}
-			return errors.Wrap(err, "cannot get controlled object")
+			return errors.Wrapf(err, errFmtGetOwnedObject, u.GetKind(), u.GetName())
 		}
 		ors := u.GetOwnerReferences()
 		for i := range ors {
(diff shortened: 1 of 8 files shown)


internal/controller/pkg/revision/reconciler.go
```diff
@@ -623,3 +634,21 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+		return errors.Wrap(err, errRemoveLock)
+	}
+
+	if err := r.objects.Relinquish(ctx, pr); err != nil {
+		return errors.Wrap(err, "cannot relinquish objects")
+	}
+
+	if err := r.hook.Deactivate(ctx, pr); err != nil {
+		return errors.Wrap(err, "deactivation hook failed")
+	}
```
hasheddan, line 651: @turkenh these steps are already carried out by the existing establisher and hook logic when a revision is inactive. Can we get the same outcome here by just adding a check [here](https://github.com/crossplane/crossplane/blob/e40d04c66f0783352d6970809f30d75e3d396d8d/internal/controller/pkg/revision/reconciler.go#L559) and skip dependency resolution if the package is inactive?

turkenh: @hasheddan just to be clear, I am not trying to solve the specific problem in #3742; instead trying to fix the general class of issues where an inactive provider revision reconciliation might fail early in the reconciliation and not be deactivated successfully, including ensuring deactivation is completed before deletion, i.e. finalizer removed.

When I checked the `ProviderRevision` reconciler, there is [a bunch of code](https://github.com/crossplane/crossplane/blob/e40d04c66f0783352d6970809f30d75e3d396d8d/internal/controller/pkg/revision/reconciler.go#L407-L584) that needs to be executed successfully to reach until the establish step (as `inactive`). Apart from the dependency-related steps, e.g. [xp constraint check](https://github.com/crossplane/crossplane/blob/e40d04c66f0783352d6970809f30d75e3d396d8d/internal/controller/pkg/revision/reconciler.go#L542) and [resolution](https://github.com/crossplane/crossplane/blob/e40d04c66f0783352d6970809f30d75e3d396d8d/internal/controller/pkg/revision/reconciler.go#L559), an inactive revision might also fail in other steps, e.g. while [fetching](https://github.com/crossplane/crossplane/blob/e40d04c66f0783352d6970809f30d75e3d396d8d/internal/controller/pkg/revision/reconciler.go#L419) or [parsing the package](https://github.com/crossplane/crossplane/blob/e40d04c66f0783352d6970809f30d75e3d396d8d/internal/controller/pkg/revision/reconciler.go#LL478C26-L478C26), imagine an old revision package is removed from the registry and no longer in the cache. Currently, we still have to successfully pull an inactive revision just to leave control of resources. 

Imagine the following hypothetical scenario:

1. User installs provider-foo v1.
2. v1 version got deleted from the registry
3. User updates provider-foo to an existing version, say v2.

IIUC, if package v1 is not in the cache somehow, e.g. XP restarted, the user will not be able to migrate to v2 since the reconciliation of v1 inactive will fail early and they will get the "is already controlled by ProviderRevision" error. 
This is just one example of how the early steps might fail for an inactive revision.

In this PR, I am trying to leverage the fact that our controller has reference to all objects it owns/controls. So, we don't need to fetch and parse the whole package just to remove the `controller: false` from those objects.

hasheddan: > When I checked the ProviderRevision reconciler, there is [a bunch of code](https://github.com/crossplane/crossplane/blob/e40d04c66f0783352d6970809f30d75e3d396d8d/internal/controller/pkg/revision/reconciler.go#L407-L584) that needs to be executed successfully to reach until the establish step (as inactive). Apart from the dependency-related steps, e.g. [xp constraint check](https://github.com/crossplane/crossplane/blob/e40d04c66f0783352d6970809f30d75e3d396d8d/internal/controller/pkg/revision/reconciler.go#L542) and [resolution](https://github.com/crossplane/crossplane/blob/e40d04c66f0783352d6970809f30d75e3d396d8d/internal/controller/pkg/revision/reconciler.go#L559), an inactive revision might also fail in other steps, e.g. while [fetching](https://github.com/crossplane/crossplane/blob/e40d04c66f0783352d6970809f30d75e3d396d8d/internal/controller/pkg/revision/reconciler.go#L419) or [parsing the package](https://github.com/crossplane/crossplane/blob/e40d04c66f0783352d6970809f30d75e3d396d8d/internal/controller/pkg/revision/reconciler.go#LL478C26-L478C26), imagine an old revision package is removed from the registry and no longer in the cache. Currently, we still have to successfully pull an inactive revision just to leave control of resources.

It's worth noting that a large portion of this code is not executed on every reconciliation -- the vast majority of cases all of this translates to just reading a file from disk and parsing, which has a very low failure rate (note: I still would like to optimize this in the future. If we parse successfully and save to disk then we really don't need to be parsing every time. I have thought about just serializing Go structs for the cache, but I think that it outside the scope of this specific PR). If you also skip dependency resolution for an inactive revision as I recommended above, then reading and parsing is really the only operation in this entire block.

If a package no longer exists in the cache or in the registry, which I would also view as a fairly infrequent case, then we have a bigger problem because the actual content is unknown. We cannot count on the status to always reflect the proper state as it could have been mutated or lost. In this event, I think we have two options (not mutually exclusive):
- Add tooling to help users clean up dangling controller references and manually remove finalizers.
- Allow for deletion even if the ownership of objects is unable to be transitioned (which I suggest adding in https://github.com/crossplane/crossplane/issues/3742), then update the dependency manager (or add another component) that watches for failing revisions and triggers a manual cleanup of controller references if the controlling revision no longer exists.

turkenh: Just prepared the following to visualize what I am proposing here:

Package Revision Reconciler

turkenh: Marking this thread as resolved, feel free to unresolve it or start another one.

turkenh:
A couple of scenarios on how this PR would help. 

### Scenario 1
The provider version you're running was cleaned up from the registry and you realized this after you upgrade your Crossplane (hence lost the package cache). You are suggested to use a newer version available in the registry.

**Steps to Reproduce:**

1. Install provider `v0.1.0`, wait until it is installed/healthy.
2. Delete the provider `v0.1.0` from the registry (e.g. dockerhub)
3. Restart Crossplane pod, i.e. package in the cache lost. (simulating an XP upgrade or a pod restart could happen anytime as well)
4. You notice the following in the provider revision events indicating the provider version is not available:
```
  Warning  ParsePackage       4s (x3 over 6s)        packages/providerrevision.pkg.crossplane.io  cannot initialize parser backend: failed to fetch package from remote: GET https://index.docker.io/v2/turkenh/provider-nop/manifests/v0.1.0: MANIFEST_UNKNOWN: manifest unknown; unknown tag=v0.1.0
```
6. You upgrade the provider to the available/newer version, `v0.2.0`.

**Currently (with `v1.13.2`):**

You end up with the following situation:

```shell
❯ kubectl get providerrevisions.pkg.crossplane.io
NAME                        HEALTHY   REVISION   IMAGE                         STATE      DEP-FOUND   DEP-INSTALLED   AGE
provider-nop-44347a1cc6f8   False     1          turkenh/provider-nop:v0.1.0   Inactive                               2m43s
provider-nop-e9ee856db6c8   Unknown   2          turkenh/provider-nop:v0.2.0   Active                                 19s

❯ kubectl describe providerrevisions.pkg.crossplane.io provider-nop-44347a1cc6f8
  ...
  Warning  ParsePackage       18s (x8 over 79s)     packages/providerrevision.pkg.crossplane.io  cannot initialize parser backend: failed to fetch package from remote: GET https://index.docker.io/v2/turkenh/provider-nop/manifests/v0.1.0: MANIFEST_UNKNOWN: manifest unknown; unknown tag=v0.1.0

❯ kubectl describe providerrevisions.pkg.crossplane.io provider-nop-e9ee856db6c8
  ...
  Warning  ResolveDependencies  25s (x5 over 74s)  packages/providerrevision.pkg.crossplane.io  cannot resolve package dependencies: cannot initialize dependency graph from the packages in the lock: node turkenh/provider-nop already exists
```

So, basically you have a `failed to fetch package from remote` on old revision and a `cannot resolve package dependencies, node already exist error` in the new version.

**After this PR:**

```shell
❯ kubectl get providerrevisions.pkg.crossplane.io
NAME                        HEALTHY   REVISION   IMAGE                         STATE      DEP-FOUND   DEP-INSTALLED   AGE
provider-nop-44347a1cc6f8   True      1          turkenh/provider-nop:v0.1.0   Inactive                               4m54s
provider-nop-e9ee856db6c8   True      2          turkenh/provider-nop:v0.2.0   Active                                 41s
```

Both revisions are healthy, hence your provider as well.
Inactive one is happy, since it doesn't need to pull/parse the package again for deactivation.
Active one is happy, since the inactive one was successfully deactivated (hence removed itself from the lock).

turkenh:
### Scenario 2
You have a provider/configuration package depending on a specific Crossplane version. You upgraded Crossplane to a new version and the Crossplane constraint on the package is no longer satisfied, and your package transition to `Unhealthy`. You decide to update your package to a recent one which has the dependency satisfying the current XP version to remedy the situation.

**Steps to Reproduce:**
1. On a Crossplane v1.12.x, install the provider `turkenh/provider-nop:v0.2.0-xp.12` depending on Crossplane `v1.12.x`. Wait until installed/healthy.
2. Upgrade Crossplane to a new version (e.g. >v1.13).
3. Notice that the package revision is failing with incompatible Crossplane version.
```shell
  Warning  LintPackage        2m45s (x2 over 2m45s)  packages/providerrevision.pkg.crossplane.io  incompatible Crossplane version: package is not compatible with Crossplane version (v1.13.2)
```
4. Upgrade the provider to version `v0.3.0-xp.gt.13` which depends on Crossplane `>=v1.13.0` 


**Currently (with `v1.13.2`):**

```
❯ kubectl get providerrevisions.pkg.crossplane.io
NAME                        HEALTHY   REVISION   IMAGE                                  STATE      DEP-FOUND   DEP-INSTALLED   AGE
provider-nop-98dc571ed410   False     1          turkenh/provider-nop:v0.2.0-xp.12      Inactive                               6m20s
provider-nop-b1e50c713eba   Unknown   2          turkenh/provider-nop:v0.3.0-xp.gt.13   Active                                 51s

❯ kubectl describe providerrevisions.pkg.crossplane.io provider-nop-98dc571ed410 # inactive/old revision
  ...
  Warning  LintPackage        78s (x3 over 4m42s)    packages/providerrevision.pkg.crossplane.io  incompatible Crossplane version: package is not compatible with Crossplane version (v1.13.2)

❯ kubectl describe providerrevisions.pkg.crossplane.io provider-nop-b1e50c713eba # active/new revision
  ...
  Warning  ResolveDependencies  2s (x7 over 2m)    packages/providerrevision.pkg.crossplane.io  cannot resolve package dependencies: cannot initialize dependency graph from the packages in the lock: node turkenh/provider-nop already exists
```

**After this PR:**

```
❯ kubectl get providerrevisions.pkg.crossplane.io
NAME                        HEALTHY   REVISION   IMAGE                                  STATE      DEP-FOUND   DEP-INSTALLED   AGE
provider-nop-98dc571ed410   True      1          turkenh/provider-nop:v0.2.0-xp.12      Inactive                               3m55s
provider-nop-b1e50c713eba   True      2          turkenh/provider-nop:v0.3.0-xp.gt.13   Active                                 29s
```

Both revisions are healthy, hence your package as well.
Inactive one is happy, since it doesn't need to visit all the boxes (see the image in the PR description) including XP version check to deactivate itself.
Active one is happy, since the inactive one was successfully deactivated (hence removed itself from the lock).

internal/controller/pkg/revision/establisher.go
```diff
@@ -117,6 +126,53 @@ func (e *APIEstablisher) Establish(ctx context.Context, objs []runtime.Object, p
 	return resourceRefs, nil
 }
 
+// Relinquish removes control of owned resources in the API server for a package
+// revision.
+func (e *APIEstablisher) Relinquish(ctx context.Context, parent v1.PackageRevision) error {
```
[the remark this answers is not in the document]
turkenh: Makes sense, I also didn't know that word before 😅 

Changed it to `ReleaseObjects` as the implementation starts with `objRefs := parent.GetObjects()`.

phisco (approved):
Awesome, thanks @turkenh!

PR #4076, opened 2023-05-16
feat: composition's environment logical validation

phisco:

Add `spec.environment` logical validation, as part of #3756 and I feel should be considered part of #3770.

Minor refactor of the logic converting an EnvironmentPatch to a Patch to a separate EnvironmentPatch.ToPatch function, to allow reusing Patch validation logic instead of duplicating it.

Also added a missing sorting of field.ErrorList that made the tests a little bit too brittle, as we don't care about the errors ordering, we sort them before comparing, already used by the Validator's tests: https://github.com/crossplane/crossplane/blob/master/pkg/validation/apiextensions/v1/composition/validator_test.go#L571.

I have:

Unit tests added and manually deploying Compositions:
```sh
kubectl apply -f - <<EOF
---
kind: Composition
metadata:
  name: nop.sqlinstances.example.org
spec:
  compositeTypeRef:
    kind: XSQLInstance
  environment:
    patches:
    - # fromFieldPath: spec.parameters.engineVersion    <- required for patches of type FromCompositeFieldPath
      toFieldPath: complex.c.f
    environmentConfigs:
    - type: Reference
      ref:
        #name: example-environment                                    <- required for type Reference
    - type: Selector
      selector:
        matchLabels:
          key: stage
          #valueFromFieldPath: metadata.labels[stage]        <- required for FromCompositeFieldPath selector
EOF

The Composition "nop.sqlinstances.example.org" is invalid:
* spec.environment.patches[0].fromFieldPath: Required value: fromFieldPath must be set for patch type FromCompositeFieldPath
* spec.environment.environmentConfigs[0].type: Required value: ref is required
* spec.environment.environmentConfigs[1].selector.matchLabels[0].valueFromFieldPath: Required value: valueFromFieldPath is required
```
While it would have been accepted and resulted in an error at runtime before.

### DIFF
--- a/apis/apiextensions/v1/composition_environment.go
+++ b/apis/apiextensions/v1/composition_environment.go
@@ -16,6 +16,12 @@ limitations under the License.
 
 package v1
 
+import (
+	"k8s.io/apimachinery/pkg/util/validation/field"
+
+	"github.com/crossplane/crossplane/internal/validation/errors"
+)
+
 // An EnvironmentConfiguration specifies the environment for rendering composed
 // resources.
 type EnvironmentConfiguration struct {
@@ -36,6 +42,25 @@ type EnvironmentConfiguration struct {
 	// Patches is a list of environment patches that are executed before a
 	// composition's resources are composed.
 	Patches []EnvironmentPatch `json:"patches,omitempty"`
+}
+
+// Validate the EnvironmentConfiguration.
+func (e *EnvironmentConfiguration) Validate() field.ErrorList {
+	errs := field.ErrorList{}
+
+	for i, p := range e.Patches {
+		if err := errors.WrapFieldError(p.Validate(), field.NewPath("patches").Index(i)); err != nil {
+			errs = append(errs, err)
+		}
+	}
+
+	for i, ec := range e.EnvironmentConfigs {
+		if err := errors.WrapFieldError(ec.Validate(), field.NewPath("environmentConfigs").Index(i)); err != nil {
+			errs = append(errs, err)
+		}
+	}
+
+	return errs
 }
 
 // EnvironmentSourceType specifies the way the EnvironmentConfig is selected.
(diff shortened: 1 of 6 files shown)


apis/apiextensions/v1/composition_validation_test.go
```diff
@@ -587,9 +595,226 @@ func TestCompositionValidateResources(t *testing.T) {
...
+				},
+			},
+			args: args{
+				comp: &Composition{
+					Spec: CompositionSpec{
+						Environment: &EnvironmentConfiguration{
+							Patches: []EnvironmentPatch{
+								{
+									Type: PatchTypeFromCompositeFieldPath,
+									//FromFieldPath: pointer.String("spec.foo"), // missing
```
lsviben, line 681: love this, makes it really readable :1st_place_medal:

lsviben (approved):
Nice work, thanks @phisco !

apis/apiextensions/v1/composition_environment.go
```diff
@@ -67,12 +92,49 @@ type EnvironmentSource struct {
...
 // An EnvironmentSourceReference references an EnvironmentConfig by it's name.
 type EnvironmentSourceReference struct {
 	// The name of the object.
 	Name string `json:"name"`
 }
 
+// Validate the EnvironmentSourceReference.
+func (e *EnvironmentSourceReference) Validate() *field.Error {
+	if e.Name == "" {
+		return field.Required(field.NewPath("name"), "name is required")
```
pedjak, line 133: IMHO, we should define this constraint on CRD schema, so that k8s api server can verify it without a need to write any code on our side?

pedjak:
it looks that the majority of validations could be performed by k8s API server, if we set constraints CRD schemas. In that way we do not need to any code on our side?

apis/apiextensions/v1/composition_environment.go
phisco: Yes, this one in particular could be enforced by setting min length to 1, which I did in the beginning, but then it would only return that and any other error according to the schema first, and only after return any other error from the validation logic, so I preferred having them checked at a single level. Most of the other ones would require CEL rules given that are mostly depending on the "type", which we could add, but I prefer having them in code until we don't put in place E2Es able to validate them.

phisco:
> it looks that the majority of validations could be performed by k8s API server, if we set constraints CRD schemas. In that way we do not need to any code on our side?

Most of them would require CEL rules, which until we have E2Es in place to test those would be harder to test. I already have a mental todo to move all the checks to CEL rules later. Right now I feel the most important thing is to have the requirements clearly specified, well tested and enforced at deploy time.

jbw976:
> I already have a mental todo to move all the checks to CEL rules later.

Would it be possible to open a tracking issue for that so it's more clear in the project backlog? 😇

phisco:
> > I already have a mental todo to move all the checks to CEL rules later.
> 
> Would it be possible to open a tracking issue for that so it's more clear in the project backlog? 😇

https://github.com/crossplane/crossplane/issues/4084 😉

apis/apiextensions/v1/composition_environment.go
turkenh: This is something we discussed [early](https://github.com/crossplane/crossplane/pull/3921#discussion_r1150817325). I remember we removed them first but added them back for some reason, probably to be consistent with the validations that the API server cannot do.

turkenh (approved):

PR #4082, opened 2023-05-17
fix: remove ignored FromEnvironmentFieldPath for environment patches

### LINKED ISSUE #4080: Ignored `FromEnvironmentFieldPath` patch type in `spec.environment.patches`
### What happened?

The Composition's CRD is configured to accept patches of type `FromEnvironmentFieldPath` to be defined in `spec.environment.patches`, see [here](https://github.com/crossplane/crossplane/blob/c4031aab55c729df6af6a9e4d7b39e3cdd2dcce9/apis/apiextensions/v1/composition_environment.go#L119), but then the [ApplyEnvironmentPatches](https://github.com/crossplane/crossplane/blob/c4031aab55c729df6af6a9e4d7b39e3cdd2dcce9/internal/controller/apiextensions/composite/composition_patches.go#L64) function is not including them.

Given that patches in that context are only meant to be applied between the Composite Resource and the `in-memory` environment, the `FromEnvironmentFieldPath` patch would be semantically equivalent to the supported `ToCompositeFieldPath` type, so either we remove it from the APIs or we just remap it to a `ToCompositeFieldPath` in code. I'd lean toward the removal option given that that API is in alpha and it's actually confusing, but I'm open to discuss it.

### How can we reproduce it?

N.A. just looking at the code is enough.

Crossplane version: v1.12.1

phisco:

Fixes #4080.

I have:

N.A. just removing a previously ignored but supported parameter.

### DIFF
--- a/apis/apiextensions/v1/composition_environment.go
+++ b/apis/apiextensions/v1/composition_environment.go
@@ -116,7 +116,7 @@ type EnvironmentPatch struct {
 	// Type sets the patching behaviour to be used. Each patch type may require
 	// its own fields to be set on the Patch object.
 	// +optional
-	// +kubebuilder:validation:Enum=FromCompositeFieldPath;FromEnvironmentFieldPath;ToCompositeFieldPath;CombineFromComposite;CombineToComposite
+	// +kubebuilder:validation:Enum=FromCompositeFieldPath;ToCompositeFieldPath;CombineFromComposite;CombineToComposite
 	// +kubebuilder:default=FromCompositeFieldPath
 	Type PatchType `json:"type,omitempty"`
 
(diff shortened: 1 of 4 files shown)


phisco:
Could be worth backporting to 1.12, WDYT?

hasheddan (commented):
> Could be worth backporting to 1.12, WDYT?

@phisco I think I would lean towards not backporting. Though this is alpha functionality and we reserve the right to change / break it from one release to another, I generally think of that policy being applicable to minor versions rather than patches. Would defer to @negz here though.

negz (approved):
Yeah I wouldn't worry about back porting this.

MisterMX (approved):
Makes sense to me. Thanks a lot @phisco!

Issue #4013, opened 2023-04-24
Improved e2e integration testing


jbw976:
We've been discussing the need for more investment in automated end to end (e2e) integration tests for Crossplane, this issue will track the ideas and progress on this high level initiative.

Some background information:

* e2e testing will help a lot with giving us confidence during releases of Crossplane
* testing within core crossplane has a very cheap and low coverage integration test in https://github.com/crossplane/crossplane/blob/master/cluster/local/integration_tests.sh that essentially verifies a fresh install of Crossplane comes up and runs OK.
* There is also some high level scenario and upgrade related testing in https://github.com/crossplane/test/, but those are gotten a bit stale.  We can fix and improve this early starting point, or invest in something more robust for the long term. We aren't married to this repo and approach.
* @negz did some interesting test automation in the Conformance repo that is worth considering for patterns/resuse/etc.: https://github.com/crossplane/conformance.
* Upgrade testing of Crossplane from a real world perspective when various features are enabled and/or providers are installed in a previous version, then Crossplane is upgraded to the latest version and everything is verified to be healthy, would be really helpful.
* Testing at the provider level for specific resources and real world scenarios is out of scope of this core Crossplane issue.  It already has some investment with the https://github.com/upbound/uptest/ tooling and we expect this is going to be driven further soon, e.g. upgrade testing with live resources, to avoid issues like we saw with https://github.com/upbound/provider-aws/issues/646#issuecomment-1495071479, but once again this is out of scope for core Crossplane
* Further investments in integration/e2e testing of **core crossplane** would be a good area to focus on, given that the providers are going to get more investment already. @muvaf recently made https://github.com/upbound/provider-dummy that could be used to test interesting core Crossplane functionality and scenarios without having to actually spin up cloud provider resources.

negz:
I recently came across https://github.com/kubernetes-sigs/e2e-framework - something to consider for Crossplane e2e tests. 

I'm generally not a fan of "heavyweight" testing frameworks. Put otherwise, I prefer tests to look as much like regular Go code as possible. While https://github.com/golang/go/wiki/TestComments#assert-libraries is talking more specifically about assert libraries, it captures my rationale pretty well:

> It also forces the assert package to create a whole new sub-language instead of reusing the existing programming language (Go itself).

I feel that keeping tests as close to regular `*testing.T` style code as possible lowers the barrier for contributors.

[...]

PR #4101, opened 2023-05-24
e2e testing design proposal

lsviben:

As described in https://github.com/crossplane/crossplane/issues/4013, Crossplane lacks e2e tests. This PR adds a design proposal for Crossplane e2e tests. 

We also added the Cucumber e2e test which would replace the current e2e tests to the PR for better visibility.

Fixes #

I have:

N/A


design/design-doc-e2e-testing.md
```diff
@@ -0,0 +1,777 @@
...
+
+  Similar to kubernetes object management, each test action should specify what
+  is the system state we would like to reach/check against. Such approach allows 
+  us to test effectively business rules exposed to users
+
+* Excellent readability for everybody (developers, users, Crossplane community
+  in general)
+
+  This would enable all stakeholders to have better/clear understanding about
+  the system behaviors.
```
negz, line 62: I suspect we're over-indexing on this goal. I might go so far as to question whether it - _excellent_ readability for non-developers - should be a goal at all. It feels to me like we're optimizing for the minority case at the expense of the majority case.

It certainly seems like a nice property, but at what cost? Is readability for non-developers so useful that we think contributors should need to learn not only a framework but a new DSL (i.e. Gherkin) just for E2E testing? This is essentially a decision to introduce a new technology into the project - not something we should take lightly.

I would expect that the _vast_ majority of the time the folks writing an E2E test will be the same folks designing, implementing, and unit testing the features that they'd like to E2E test. This isn't a project where we have a bunch of less-technical business stakeholders that need to validate that the software engineers built the thing to spec.

It's not _inconceivable_ to me that a community member who is not comfortable reading or writing Go would want to add an E2E test to demonstrate that something was broken, or want to take a look at an E2E test (instead of documentation) to try to understand how something works. I'm quite skeptical that it will happen often enough to warrant us introducing a new DSL though. I would expect the vast majority of folks reading and writing these E2E tests will be the same folks reading and writing the rest of the codebase - Go developers.

design/design-doc-e2e-testing.md
```diff
@@ -0,0 +1,777 @@
...
+can understand. As such, Cucumber allows the execution of feature documentation
+written in business-facing text”. _
+
+Cucumber is a mature project/approach used successfully for testing systems of
+various complexity and size, for example:
+
+* [Service Binding Operator]
+* [Conformance tests for servicebinding specification]
+* [Primaza]
+* [Elastic end-to-end testing]
```
negz, line 137: Are we able to get a sense of whether choosing Cucumber resulted in tests that were significantly more useful to non-developers for these projects? How many tests for these projects were contributed by non-developers?

negz: It would be useful to add a section like this for the alternatives considered, too. i.e. What projects are using the alternatives.

design/design-doc-e2e-testing.md
```diff
@@ -0,0 +1,777 @@
...
+cluster) before we can focus on validating a particular system behavior.
+Defining the fixture only with the help of Go standard library can be
+time-consuming, requiring to write a significant amount of boiler-plate code,
+hence reducing test readability and maintenance, and eventually not very
+motivating for creating new tests. In order to improve the readability, one
+would start looking for patterns and either eventually end up with a custom
+framework or adopt an existing one.
+
+**_NOTE_**: The document uses end-to-end testing term interchangeably
+with [system testing].
```
negz, line 44: I know we use the term "E2E" in https://github.com/crossplane/crossplane/issues/4013, but I don't think we need to be too strict about this. I get the sense that in most cloud native projects "E2E testing" often includes integration testing and isn't strictly (whole) system testing.

design/design-doc-e2e-testing.md
```diff
@@ -0,0 +1,777 @@
...
+upstream e2e tests. Rather, it's an effort to make projects using Kubernetes
+have a standardized and convenient way to test. From what we see, it is for now
+used by Cilium.
+
+#### Conclusion
+
+The e2e-framework might be a good candidate for writing integration tests, but
+it is not well-suited for declarative e2e testing. Given the nature of e2e
+tests, the setup might be quite complex, requiring a lot of discipline to keep
+code readable and maintainable.
```
negz, line 629: Is there a disconnect between what we consider an E2E test and what these projects consider an E2E test. I'm surprised that we feel a project called `e2e-framework` is not well suited to writing E2E tests.

As you know, this framework _or something like it_ is my preferred approach. The overwhelming majority of cloud native and Kubernetes ecosystem tools seem to take its approach of using (and typically extending) the stdlib Go `testing` package to write E2E tests:

* https://github.com/kubernetes/kubernetes/tree/master/test/e2e
* https://github.com/knative/serving/tree/main/test/e2e
* https://github.com/istio/istio/wiki/Istio-Test-Framework
* https://github.com/rook/rook/tree/master/tests
* https://github.com/coredns/coredns/tree/master/test
* https://github.com/etcd-io/etcd/tree/main/tests/e2e
* https://github.com/argoproj/argo-cd/tree/master/test/e2e
* https://github.com/projectcontour/contour/tree/main/test/e2e

design/design-doc-e2e-testing.md
```diff
@@ -0,0 +1,777 @@
...
+time-consuming, requiring to write a significant amount of boiler-plate code,
+hence reducing test readability and maintenance, and eventually not very
+motivating for creating new tests. In order to improve the readability, one
+would start looking for patterns and either eventually end up with a custom
+framework or adopt an existing one.
+
+**_NOTE_**: The document uses end-to-end testing term interchangeably
+with [system testing].
+
+## Goals
```
negz, line 46: I think what we have here are the list of properties we want from an E2E testing framework. Would it also be helpful to agree here on why we want an E2E framework? What isn't working well today? Presumably our overall goal (the reason we're adding E2E tests) is to solve those problems.

pedjak:
As a part of this PR (commit https://github.com/crossplane/crossplane/pull/4101/commits/1f999a3043b98f297feb52f2053aef70e7032adb), we have converted the existing test under `test/e2e` to godog Cucumber implementation, so that people can get better feeling how the future might look like. The test can be executed with

```
make build e2e
```

design/design-doc-e2e-testing.md
pedjak: Excellent readability is important for everybody. Tests code similar to production code is more often read than written. In order to enable developers to move fast and be productive, the readability plays a key role.

> It certainly seems like a nice property, but at what cost? 

If we want to measure costs, then we we need to come up with a set of measurable criteria, otherwise we can easily slip into subjective debate. Some possible criteria could be:

* Time to learn new concepts required for writing an end-to-end test
* Time to write an end-to-end test
* Time to modify single end-to-end test
* Time to modify a group of end-to-end tests

The proposed approach opens up possibility to produce useful impact beyond Crossplane validation. As mentioned in [Benefits Section](https://github.com/crossplane/crossplane/pull/4101/files#diff-e18ff1e97964f135a7e28848ac1355e73c999d953bd7563e0d39dadf8a07aebfR412), we could obtain testable documentation, where all snippets/example are tested as a part PR merge checks (cc @plumbis) 

> contributors should need to learn not only a framework but a new DSL (i.e. Gherkin) just for E2E testing? This is essentially a decision to introduce a new technology into the project - not something we should take lightly.

All evaluated frameworks are accompanied with a sort of DSL, in some cases it is delivered as a set of Go APIs/functions/conventions.

You can see Gherkin language that we control, as a sort of configuration file, but written in format that supports pure English sentences. IMHO, that is more readable than yaml, or json. As a part of this PR we have converted the exiting e2e test to Cucumber/godog approach. For that we needed to define just a handful of step/sentences, yet this is already enough for majority of tests that need to be written. 

A Crossplane contributor would need to learn following concepts:
* Test step name is just a much more readable name of Go function that gets invoked eventually
* Mapping between test step name and Go function is declared during test setup, in Go code
* Scenario description maps to test name, but again we can get more clarity due to free form text format
* Feature maps to test suite

Let's see how this works on example, [this test](https://github.com/crossplane/crossplane/pull/4101/files#diff-35a0aa68b3658cb9c9b0c3b7c17ba49d64b33709d8c4a0b88971157157ae1814R85-R193) is godog version of [existing `TestCompositeResourceGetsReady`](https://github.com/crossplane/crossplane/blob/master/test/e2e/apiextensions/composition_test.go#L48-L364) test:

* Pure Go tests has 3 times more lines of code
* Crafting/Applying CompositeResourceDefinition in Go looks like:

```go
&extv1.CompositeResourceDefinition{
	ObjectMeta: metav1.ObjectMeta{Name: "clusternopresources.nop.example.org"},
	Spec: extv1.CompositeResourceDefinitionSpec{
		Group: "nop.example.org",
		Names: kextv1.CustomResourceDefinitionNames{
			Kind:     "ClusterNopResource",
			ListKind: "ClusterNopResourceList",
			Plural:   "clusternopresources",
			Singular: "clusternopresource",
		},
		ClaimNames: &kextv1.CustomResourceDefinitionNames{
			Kind:     "NopResource",
			ListKind: "NopResourceList",
			Plural:   "nopresources",
			Singular: "nopresource",
		},
		ConnectionSecretKeys: []string{"test"},
		Versions: []extv1.CompositeResourceDefinitionVersion{{
			Name:          "v1alpha1",
			Served:        true,
			Referenceable: true,
			Schema: &extv1.CompositeResourceValidation{
				OpenAPIV3Schema: runtime.RawExtension{Raw: []byte(`{
					"type": "object",
					"properties": {
						"spec": {
							"type": "object",
							"properties": {
								"coolField": {
									"type": "string"
								}
							},
							"required": ["coolField"]
						}
					}
				}`)},
			},
		}},
	},
}
```

whereas `CompositeResourceDefinition is present` steps allows passing pure YAML content as the parameter:

```gherkin
CompositeResourceDefinition is present
      """
        metadata:
          name: clusternopresources.nop.example.org
        spec:
          group: nop.example.org
          names:
            kind: ClusterNopResource
            listKind: ClusterNopResourceList
            plural: clusternopresources
            singular: clusternopresource
          claimNames:
            kind: NopResource
            listKind: NopResourceList
            plural: nopresources
            singular: nopresource
          connectionSecretKeys:
            - test
          versions:
            - name: v1alpha1
              served: true
              referenceable: true
              schema:
                openAPIV3Schema:
                  type: object
                  properties:
                    spec:
                      type: object
                      properties:
                        coolField:
                          type: string
                      required:
                        - coolField
```

IMHO, the later reads a way better.

* Similar to the above, step `claim becomes synchronized and ready` allows a reader to immediately understand the expectations, compared to the equivalent Go code:

```go
if err := wait.PollImmediate(10*time.Second, 90*time.Second, func() (done bool, err error) {
	d, err := dc.Resource(nopRes).Namespace("default").Get(context.TODO(), "nop-example", metav1.GetOptions{})
	if err != nil {
		t.Fatalf("Get nop-res %q: %v", nopresource.GetName(), err)
	}

	nopObj := composed.Unstructured{Unstructured: *d}
	isReady := nopObj.GetCondition(xpv1.TypeReady)

	if isReady.Status != corev1.ConditionTrue {
		t.Logf("nop-example %q is not yet Ready", nopresource.GetName())
		return false, nil
	}

	return true, nil
}); err != nil {
	t.Errorf("nop-example %q never became Ready: %v", nopresource.GetName(), err)
}
```

We could argue that the above code can be wrapped in a function and used like that, but IMHO, reading

`claim becomes synchronized and ready`

is much clearer then

`claimBecomesSynchronizedAndReady(context, claimName, namespace)`

A very important thing to state is that steps grammar is not dictated by Gherkin/godog - we are in full control of it, and are gonna design it according to our needs.

design/design-doc-e2e-testing.md
pedjak: > Are we able to get a sense of whether choosing Cucumber resulted in tests that were significantly more useful to non-developers

I was directly involved in the first two listed project, and must say that choosing Cucumber brought all benefits listed in this proposal.

design/design-doc-e2e-testing.md
pedjak: What would be reason for relaxing the term? In Crossplane, integration testing is already covered in [this proposal](https://github.com/crossplane/crossplane/blob/master/design/design-doc-integration-tests.md).

design/design-doc-e2e-testing.md
pedjak: > I'm surprised that we feel a project called e2e-framework that is designed specifically for testing Kubernetes controllers is not well suited to writing E2E tests for Kubernetes controllers.

As a part of evaluation, we wrote the same tests using all four frameworks, and analyzed if produced tests are matching the set goals. We are not claiming that `e2e-framework` cannot be used for testing controllers - we found that it does not fulfill the goals set in this proposal (e.g. declarative testing).

design/design-doc-e2e-testing.md
pedjak: The proposal is about introducing end-to-end tests that should match the listed goals (or properties). The goals are orthogonal to evaluated frameworks - they are matching more or less with the set goals.

The motivation for the proposal is stated in Background section. Feel free to propose some rephrasing.

design/design-doc-e2e-testing.md
negz: > we found that it does not fulfill the goals set in this proposal (e.g. declarative testing).

That's why I'm challenging some of these goals. I'm surprised that Crossplane needs to take what appears to be a novel path within our space.

negz:
> If that it is the case, perhaps we should think about removing it from the repository?

@pedjak We typically move things into https://github.com/crossplane/crossplane/tree/master/design/defunct in cases like this. I think we should do that.

design/design-doc-e2e-testing.md
negz: That proposal never went anywhere. We don't have any meaningful integration tests.

What we need is something higher level than unit tests that tests code as more of a black box than unit tests do. I think we can be pragmatic as to whether those new tests are integration tests, system tests, or a mix of both.

design/design-doc-e2e-testing.md
```diff
@@ -0,0 +1,809 @@
...
+				if isReady.Status != corev1.ConditionTrue {
+					t.Logf("claim %q is not yet Ready", claim.GetName())
+					return false
+				}
+				return isReady.Status == corev1.ConditionTrue
+			}), wait.WithTimeout(time.Minute*1))
+			if err != nil {
+				t.Fatalf("failed to wait for claim to be Ready: %v", err)
+			}
+			return ctx
```
turkenh, line 598: As someone who had a good amount of exposure to the Crossplane codebase and Golang in general, I feel less comfortable with maintaining or extending a test case looking like this compared to what we see with Cucumber above. Yes, you can make it more readable with some refactoring and adding some comments etc. (which is hard to keep quality under control, especially with the test code), but it would never reach that level of readability.

While looking for some other real-life examples of Cucumber, I came across [this test definition](https://github.com/awslabs/aws-cloudformation-controller-for-flux/blob/main/internal/integtests/features/cfn_controller.feature), imagining I need to implement a feature there, I felt very comfortable with extending it.

design/design-doc-e2e-testing.md
```diff
@@ -0,0 +1,809 @@
...
+			},
+			Namespace: fmt.Sprintf("test-%s", scenario.Id),
+		}
+		if err := sc.Cluster.createNamespace(sc.Namespace); err != nil {
+			return ctx, err
+		}
+		ctx = context.WithValue(ctx, scenarioContextKey, sc)
+		return ctx, nil
+	})
+	ctx.Step(`^claim becomes synchronized and ready$`, claimBecomesSynchronizedAndReady)
```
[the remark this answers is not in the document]
lsviben: The steps we made are an example of what can be done, I don't think any the framework examples has very optimized code. I see no reason why we couldn't implement a step like `<resource> becomes <conditions>...` . 

The expectation is that during the initial few tests we create we create a collection of steps which would be enough to write any test just by making a .feature, without the need to go into the steps and their implementation

design/design-doc-e2e-testing.md
pedjak: > I think we can be pragmatic as to whether those new tests are integration tests, system tests, or a mix of both.

Pragmatic can mean many things, but it is orthogonal to  being ambiguous. IMHO there is no one-size-fits-all solution for testing - each sort of tests requires specialized abstractions, because they are executed on different levels and expectations.

For me pragmatic means choosing a right tool for a task. As we can see, Crossplane unit tests are alive and kicking, but [the integration testing proposal](https://github.com/crossplane/crossplane/blob/master/design/design-doc-integration-tests.md) did not get traction. That a strong signal that we would need to approach the problem differently.

design/design-doc-e2e-testing.md
ytsarev: First of all, I want to say that I'm impressed with the overall quality of the document and the e2e testing endeavor.

I agree with @pedjak that dealing with Golang structs instead of Yaml in the testing context is ugly. So let's please not do that, no matter the framework we pick in the end :D However, this comparison is orthogonal to the tooling, as we can always reference arbitrary Yaml files and include them in the tests, no matter the framework.

I have to also agree with @negz - the readability is a very subjective thing, so I can subjectively share that human-readable Cucumber/Gherkin is usually less **understandable** to me than structured Golang code or Rspec-like https://github.com/onsi/ginkgo code.

In my previous life, I was active in a Ruby community where Cucumber and Rspec were born, and the BDD movement was rising. Eventually, the same patterns were adopted by a wider community of languages and stacks. 

So I was doing Cucumber before it was cool(tm) in several projects and companies and can share some real-life stories. The original goal of Cucumber was/is to bring technical and non-technical stakeholders onto the same page. It is a very noble goal, but it happened in precisely 0 projects and companies where I witnessed(and in early cases was driving) the Cucumber adoption.

The end-state is usually the same: developers deal with the Gherkin abstraction they never need. Non-technical people are doing non-technical stuff and never contribute to the test suite. It makes the whole dance around human readability and step definitions an expensive redundant layer.

Back in the day, I preferred to stick with https://rspec.info/, which provided a middle ground: some form of BDD metastructure without much hiding the actual implementation. 

In the Golang world, there is https://github.com/onsi/ginkgo, a symmetric tool to Rspec. It was popular in the core Kubernetes testing effort.

As a concerning point, it is [mentioned](https://github.com/kubernetes-sigs/e2e-framework/blob/c3b933fe4c7313a31cbc2ea3cad2d6a2a3c8962f/docs/design/README.md?plain=1#L12) in e2e-framework documentation that moving away from Ginkgo is one of the primary motivation of the project creation.

Overall, to summarize my point:

* While Cucumber does look fancy and nice, Gherkin will be an almost useless and expensive layer long-term.
* The test creation process will contain serious overhead. Developers will have to think about the 'human readable' Gherkin representation string, then the Step definition regexp for it, and only then about actual test implementation. Again, too many layers. 
* Instead, I would love to have an original RSpec-like test layout where I can structure my test with BDD annotations and write implementation at the same place without many *mandatory* abstraction layers. Meanwhile, in this case, nothing stops me from abstracting things when it makes sense from a developer's standpoint.

This makes https://github.com/onsi/ginkgo another potential candidate to evaluate. It looks like people are successfully [using it ](https://itnext.io/testing-kubernetes-operators-with-ginkgo-gomega-and-the-operator-runtime-6ad4c2492379)to test standalone k8s operators.

The lightweight testing abstractions like [terratest](https://terratest.gruntwork.io/) ( evaluated in this doc) also [worked nicely](https://github.com/k8gb-io/k8gb/tree/master/terratest) for me in another project. 

I apologize if my points came across as strong. The overall doc and e2e testing goals are just excellent stuff!

design/design-doc-e2e-testing.md
```diff
@@ -0,0 +1,809 @@
...
+#### Anatomy of a test
+
+Tests are organized in test suites, each test suite being a Go package. The
+function TestMain is used to define package-wide testing steps and configure 
+behavior. The framework exposes the Environment interface, which is used to
+interact with the cluster.
+
+Furthermore, the environment is set up by defining steps that are executed in
+different phases of the test, like Setup, Finish, BeforeEach, kinda similar to
+Ginkgo.
```
ytsarev, line 529: As a follow-up to my longer comment above, is there any reason why https://github.com/onsi/ginkgo didn't get into the evaluation round?

design/design-doc-e2e-testing.md
pedjak: > The test creation process will contain serious overhead. Developers will have to think about the 'human readable' Gherkin representation string, then the Step definition regexp for it, and only then about actual test implementation. Again, too many layers.

Godog has a possibility to generate the up-to-date list of available steps with documentation as well, so developers can always see what are the steps they can use. Also, at beginning, while adding missing e2e tests, i.e. matching them with the available functionalities, there will be a high chance that new step needs to be introduced. However, after the test base stabilize, the experience shows that adding new step becomes very rare, and most of the time adding a new test requires just editing some of feature files.

design/design-doc-e2e-testing.md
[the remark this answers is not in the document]
pedjak: > How reusable would these bindings be?

As you can see already in this PR (features files), the steps are reused across multiple scenarios. Yes, the steps can be parametrized and formulated so that it matches our expectations...

design/design-doc-e2e-testing.md
pedjak: @negz 

>  Read and write Gherkin. (Yes it's English-language but it does appear to have some meaningful syntax).

The syntax is what we control and can make it as meaningful as we need.
> Read and write write Go code that parses Gherkin (e.g. ctx.Step('^claim becomes synchronized and ready$\', claimBecomesSynchronizedAndReady))

Nope, this is what Godog does for us - we only need to tell it what Go function to invoke when certain step name is matched. This is done once when a step gets introduced.

>    Read and write the Go code that seems to actually run the test, i.e. claimBecomesSynchronizedAndReady.

Not really needed, especially after the test code base stabilizes. At beginning, while adding missing e2e tests, i.e. matching them with the available functionalities, there will be a high chance that new step needs to be introduced. However, after the test base stabilize, the experience shows that adding new step becomes very rare, and most of the time adding a new test requires just editing some of feature files

design/design-doc-e2e-testing.md
```diff
@@ -0,0 +1,809 @@
...
+that they do not require any provider to be installed, i.e. it would be enough
+to install some carefully crafted CRDs, and create XRDs and claims against them.
+Such a strategy would make required setups even lighter and tests would
+potentially run faster.
+
+### Blackbox Testing
+
+Steps implementation communicates with the cluster only by using [kubectl CLI].
+In this way we can be sure that users are able to reproduce tests either
+manually or in an automated way.
```
hasheddan, line 403: Does the framework handle installation of dependencies (e.g. CLI tools)? Or do we anticipate handling that via `make` or similar?

hasheddan: Also, is being able to manually execute some version of these tests using `kubectl` and other tools a goal here? I don't think it is a negative necessarily, but I also don't think manually going through the tests instill higher confidence in functionality, and I don't think we should strive for our e2e tests to also act as examples for end-users.

design/design-doc-e2e-testing.md
```diff
@@ -0,0 +1,809 @@
...
+
+#### Blackbox testing
+
+The framework is using the controller-runtime client underneath, wrapped under
+its `klient`, which offers some abstractions to make testing easier. In order to
+deploy Crossplane resources, we would need to register Crossplane schemas within
+the client. Such a setup is more convenient for running integration tests. For
+end-to-end testing we should simulate users' interaction with the system and
+that happens very often (as advocated in Crossplane doc)typically using kubectl
+CLI.
```
hasheddan, line 620:
```suggestion
that happens very often (as advocated in Crossplane doc) typically using kubectl
CLI.
```

test/e2e/features/composition.feature
```diff
@@ -0,0 +1,193 @@
...
+                openAPIV3Schema:
+                  type: object
+                  properties:
+                    spec:
+                      type: object
+                      properties:
+                        color:
+                          type: string
+                      required:
+                        - color
```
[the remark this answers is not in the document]
pedjak: Sure, we could make a step that reads from filesystem, anything is possible because we author what step logic should be. I can tell you that in the projects where I used Cucumber (e.g. [Service Binding Operator](https://github.com/redhat-developer/service-binding-operator)), we never had issues with indentation. Furthermore, the issues could be easily detected either at the test execution or through [linting](https://github.com/vsiakka/gherkin-lint).

The benefit of having yaml content inline is that it is easy to follow the flow, all test details are in the single place.

design/design-doc-e2e-testing.md
hasheddan: I think there is a lot to gain from utilizing a project that comes from upstream, both in getting functionality "for free" and in enabling new Crossplane developers to write e2e tests in a familiar way.

design/design-doc-e2e-testing.md
```diff
@@ -0,0 +1,809 @@
...
+
+
+* The tests are executed using the standard Go approach → they could be easily
+  triggered/debugged from various IDEs as well.
+* The amount of Go code implementing steps is very low, and yet the implemented
+  steps are enough for declaring a large number of test cases
+
+### Declarative Testing
+
+Cucumber allows us to perform [both imperative and declarative testing].
```
hasheddan, line 183: It might be worth adding some examples of output from these tests. Even if a test passes, it is useful to know number of retries, execution time for each step, etc.

design/design-doc-e2e-testing.md
```diff
@@ -0,0 +1,809 @@
...
+  assertion logic. Similar to that, global changes in setup/assertion logic
+  should be kept in the single place ([DRY principle])
+
+As a part of this effort, already existing tests should be converted to the
+chosen approach/framework.
+
+## Non-Goals
+
+* Further followup actions should introduce more tests covering all Crossplane
+  core user-facing functionalities
```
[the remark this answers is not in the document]
pedjak: Yes, this PR already rewrite the existing tests under `test/e2e` using the proposed framework. Further PRs are going to add test that cover the rest of user-exposed functionalities.

design/design-doc-e2e-testing.md
```diff
@@ -0,0 +1,809 @@
...
+        * Patching → should have a patch of each type tested
+        * ConnectionDetails → verify the secret
+        * Late initialization → verify its propagated from mr->xr->claim
+        * Composition functions
+        * Composition validation
+        * ESS
+        * ObserveOnly scenarios
+        * Error scenarios
+    * Crossplane upgrades
+* Setting up CI to run e2e tests as a pre-merge checks
```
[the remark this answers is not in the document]
pedjak: In general, setting up CI is gonna follow in future PRs).

design/design-doc-e2e-testing.md
pedjak: > Does the framework handle installation of dependencies (e.g. CLI tools)? Or do we anticipate handling that via make or similar?

The only needed CLI is `kubectl` - if needed we could install it via `make`.

> is being able to manually execute some version of these tests using kubectl and other tools a goal here?

No, it is more just a statement that anybody reading the test could even manually reproduce it. A nice bonus point, but as you said, not a goal at all.

design/design-doc-e2e-testing.md
pedjak: Sure, you can check it easily. On this branch you can execute the tests simply with

```
make build e2e
```

design/design-doc-e2e-testing.md
```diff
@@ -0,0 +1,809 @@
...
+
+The tests are written in pure Go, requesting to keep the number of library
+dependencies at minimum.
+
+Contrary to unit tests, end-to-end testing requires very often complex test 
+fixtures (e.g. installing Crossplane, providers, reaching some state in the
+cluster) before we can focus on validating a particular system behavior.
+Defining the fixture only with the help of Go standard library can be
+time-consuming, requiring to write a significant amount of boiler-plate code,
+hence reducing test readability and maintenance, and eventually not very
```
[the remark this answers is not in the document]
pedjak: > primarily that it is also written in Go.

@hasheddan the testing code (step implementations) are written in Go. You can see Cucumber feature file as configuration that gets parsed at runtime.

design/design-doc-e2e-testing.md
ytsarev: > Ginkgo is explicitly mentioned in [the contribution guide](https://github.com/crossplane/crossplane/blob/9ab0598dd964d796574dc5a1a712f4ee6687087f/contributing/README.md?plain=1#L50)

IIUC ginkgo is mentioned there in the unit test context so it does not invalidate using it in e2e scenario

design/design-doc-e2e-testing.md
negz: > That is a strong signal that we would need to approach the problem differently.

I disagree. The problem we've had in the past was not choice of framework, but rather bandwidth to prioritize actually writing the tests. Picking any framework is pointless if we can't form and execute a plan to bootstrap the test coverage. If every new PR to the project sees there are no tests, no one is motivated to add tests for their new code.

design/design-doc-e2e-testing.md
negz: > you can check it easily

Let's add an example to the document per @hasheddan's suggestion please. That way reviewers can get an idea of what we're dealing with without needing to check out the branch and run the code.

test/e2e/features/composition.feature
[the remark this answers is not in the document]
ytsarev: Complex Composition yamls are a frequent source of obvious understandable human indentation errors. I am not convinced it will be somehow different in between Gherkin strings :) But if it fails fast with e.g. validation webhook enabled on the test cluster, then it is probably fine.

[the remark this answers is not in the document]
negz: Definitely agree with @ytsarev's concerns here.

My personal preference is writing Go structs rather than YAML manifests for the type safety. That said, I can live with manifests and see their advantages. Loading them as test fixtures from a regular `.yaml` file seems like it would allow editors, linters, etc to more easily operate on the file as standard YAML, rather than YAML-inside-Gherkin.

negz:
I want to explain how I think about technology choices in the Crossplane project. It may help explain my views on this proposal.

Here are two guiding principals:

1. The fewer technologies in use, the better. Fewer variants of languages, build tools, frameworks, libraries, patterns, etc means fewer things to learn for contributors. We should have a high bar for adopting new technologies.

2. When making technology choices, _especially_ subjective ones, I like to "appeal to a higher authority". What do the folks who designed our tech stack recommend? What choices to they make? They're likely to be well informed, and likely to shape the majority view within their ecosystem. Crossplane is built using Go, and built on Kubernetes so to me these are our obvious higher authorities.

I think we agree _some_ new technology is needed to solve the problem at hand. We do want _something_ beyond the `testing` stdlib package to make writing E2E tests more convenient. So the lens I'm looking at is how "disruptive" is the new technology. How different is the new technology from the ones we already use? On one end of the spectrum a highly disruptive choice would be introducing a new programming language. At the other end of the spectrum a minimally disruptive choice would be adopting a small Go library.

When I look to our higher authorities here's what I see.

The Go community appears to prefer to avoid _heavyweight_ frameworks that abstract away what is being done. This is contrary to other popular languages like Ruby (e.g. Rails) and Java (e.g. Spring). For example many Go web frameworks like [chi](https://github.com/go-chi/chi), [gorilla](https://www.gorillatoolkit.org), and [gokit](https://gokit.io) _supplement_ the stdlib. They don't _abstract away_ the stdlib. They're more like set of libraries that you pick and choose useful parts from.

The Go project itself is pretty explicit about this when it comes to testing frameworks, specifically. For example, the Go test review comments document captures frequent feedback that is given on Go changes. It [says](https://github.com/golang/go/wiki/TestComments#assert-libraries):

> Make use of Go itself instead of creating a mini-language inside Go.

The [Go FAQ](https://go.dev/doc/faq#testing_framework) also says:

> Testing frameworks tend to develop into mini-languages of their own, with conditionals and controls and printing mechanisms, but Go already has all those capabilities; why recreate them? We'd rather write tests in Go; it's one fewer language to learn and the approach keeps the tests straightforward and easy to understand.

When I look to the Kubernetes ecosystem, I see that they _now_ appear to be largely aligned with the Go community.

Their tests were originally built on [Ginkgo](https://github.com/onsi/ginkgo), another BDD-focused testing framework. I would argue that Ginkgo is a heavyweight framework. It abstracts away the Go stdlib `testing` user experience; writing Ginkgo tests does not feel like writing tests using the Go `testing` package. It describes itself as a DSL, and someone who knows Go but not Ginkgo [would need to spend some time reading their documentation](https://onsi.github.io/ginkgo/#writing-specs) to learn how to read and write tests.

Leaders within the Kubernetes test ecosystem identified use of Ginkgo as a problem around [three years ago](https://kubernetes.slack.com/archives/C09QZ4DQB/p1588863399476800) and are [intentionally moving away from it](https://github.com/kubernetes-sigs/e2e-framework/blob/main/docs/design/README.md#motivations) toward what appears to be a framework that seems quite aligned with the Go community practices I touched on above. Some relevant goals of this new framework include:

* Move away from reliance on Ginkgo or other non-standard Go based tests packages
* Rely and use Go testing package constructs to create test suites and test functions
* Rely on built-in Go test features to easily select/filter tests to run during execution
* Provide sensible programmatic API components to compose tests

When I look at what this design proposes - adopting Cucumber and Godog - I'm seeing what I would consider to be another _heavyweight framework_, like Ginkgo. If I understand correctly, in order to write tests using Godog I:

1. Write one `TestE2E(t *testing.T)` test that "hands control" from the `testing` stdlib package to Godog.
1. Write a `feature` file in the Gherkin "language" with my test inputs and outputs.
1. Write regular expressions to associate each Gherkin feature with a Godog-style test function.
1. Write a Godog-style function (in Go) to actually implement the test logic.

I consider this to be a heavyweight framework because contributors need to learn:

* A new language (Gherkin) to describe test inputs and outputs - even if that language is pretty simple.
* How to write Godog-style test functions, and plumb them up to a Godog test suite.

Notably a Godog-style test function does not seem to be a typical `TestFoo(t *testing.T)` function that anyone who has written Go will be familiar with. Instead the signature of the function appears to be derived from the regular expression in the `ctx.Step` call. So, it's not just new test function logic that folks need to learn. We're asking them to learn Godog's way of writing, structuring, and "plumbing" tests.

I also note that this PR also begins to introduce essentially our own library for interacting with Kubernetes. So we're proposing two frameworks in a sense; the general BDD testing Cucumber framework, as well as our own library for interacting with Kubernetes (`ApplyYamlToNamespace`, `WaitForResourceConditionMatch` etc).

So with this in mind, I start wondering _what's special about our project_? Why would we make a novel technology choice here - a choice that seems to go against the majority of our ecosystem?

* Do we have a unique requirement that others in our ecosystem do not? (If so, why?)
* Do we think [the majority of our ecosystem](https://github.com/crossplane/crossplane/pull/4101#discussion_r1204813037) just got it wrong?
* Do we have a strong consensus within the project that the novel path is better?

design/design-doc-e2e-testing.md
negz: @turkenh Isn't Cucumber pretty similar though if you actually have to add test logic?

Adding a feature seems fairly straightforward (though I worry that in reality you need to understand the regex that binds each to actual test logic). The feature appears to essentially be specifying text fixtures with a bit of flow control though - you still need to write a bunch of Go code to actually do something with the test fixture.

https://github.com/awslabs/aws-cloudformation-controller-for-flux/blob/0dd8444/internal/integtests/cfn_controller_test_scenario.go#L210

I would argue for example the above code is pretty similar to what we see here. It just seems a lot less like writing a Go `testing` test and more like writing arbitrary Go code.

design/design-doc-e2e-testing.md
negz, line 620: I would not over-index on needing to use `kubectl` to interact with the system as a black box.

I would argue that any client of the API server - even a programmatic client - is perfectly fine. You're still treating the system (the control plane) as a black box. I can see the value of testing that a really popular client (`kubectl`) works as we'd expect it. On the other hand we have plenty of control plane clients that are not `kubectl` (and are probably written with client-go) like ArgoCD, Helm, https://github.com/upbound/xgql, etc.

In my mind the fragility of shelling out to `kubectl`, `jq`, etc is not worth the benefit we get from testing that that client specifically.

lsviben:
There are a lot of threads here so maybe ill try to sum up my feelings in one place. 

My goal with driving the e2e test topic was simply to have e2e tests in Crossplane, so that the project matures and we get better confidence in our releases, code changes, bug fixes, etc. 
We could have just continued writing some tests to cover different functionalities based on the tests we currently have ([xp e2e](https://github.com/crossplane/crossplane/tree/9ab0598dd964d796574dc5a1a712f4ee6687087f/test/e2e)) but it seemed better to agree beforehand on what we exactly wanted from the tests, and to define some structure(choose a framework) istead of spending time and effort to evolve some custom framework of our own. From what I see, the projects mentioned [ here all have some own custom frameworks](https://github.com/crossplane/crossplane/pull/4101#discussion_r1204813037), which their contributors need to learn and maintain. 

As I said, my goal is to choose somehting so we can move forward with the higher goal, which is adding the e2e tests. For that, we need the project maintainers, reviewers and contributors all at least slightly positive/neutral, if not super happy, with the framework through which we will achieve that goal. So if Cucumber/Godog gets a lot of pushback, and some convincing does not help, for me that means that it doesnt pass the threshold of being at least slightly positive for our group. 

That said, I would like to shortly express how I feel about the solutions proposed and alternatives, and suggest to others to do the same so we can come to a common acceptable solution that we can move forward with.

**Cucumber**: Positive
Although it introduces a new language, and we need to create the steps with godog, I feel that after a few tests, we will have a step library wide enough that it would allow us to add new tests in a little amount of time, without the need to go into the step details for most tests. 

**Kuttl**: Slightly negative
Its neat, but I feel the lack of flexibility will make writing tests sometimes fustrating and that its not the right tool for our case.

**e2e-framework**: Slightly positive
I like the idea, but feel that the setups could become complicated. Furthermore, the project is not yet widely accepted in the community(from what I see its [only used by Cillium](https://github.com/kubernetes-sigs/e2e-framework/network/dependents?package_id=UGFja2FnZS0yNjQ4OTc1OTA3)), my worry is that its not mature enough yet. On the positive side it provides structure to tests that we wont need to implement ourselves if we go with this route. 

**Terratest**: Slightly positive
Great helpers, but thats about it. We would need to build our own framwork on top of it, but if we want to go that way, its a good project to use.

Ill mention Ginkgo as well as it was being discussed
**Ginkgo**: Slightly negative
Its a powerful tool but for me it becomes unreadable in more complicated cases. In this case I would prefer to use e2e-framework as it has most of the gingko setup functionalities but flatter, and closer to Go testing.

design/design-doc-e2e-testing.md
lsviben: Yes, e2e-framework comes from upstream, but its not widely in use yet. From what I see (but I may be wrong), its [only used in Cilium for now](https://github.com/kubernetes-sigs/e2e-framework/network/dependents). So I don't think that the argument of "enabling Crossplane developers to write e2e tests in a familiar way applies". Of course that can change in time if the project gets traction. In the meantime there is probably a bigger chance that a developer already knows Cucumber or with even a higher probability Ginkgo. 

> As you know, this framework _or something like it_ is my preferred approach. The overwhelming majority of cloud native and Kubernetes ecosystem tools seem to take its approach of using (and typically extending) the stdlib Go `testing` package to write E2E tests:
> 
> * https://github.com/kubernetes/kubernetes/tree/master/test/e2e
> * https://github.com/knative/serving/tree/main/test/e2e
> * https://github.com/istio/istio/wiki/Istio-Test-Framework
> * https://github.com/rook/rook/tree/master/tests
> * https://github.com/coredns/coredns/tree/master/test
> * https://github.com/etcd-io/etcd/tree/main/tests/e2e
> * https://github.com/argoproj/argo-cd/tree/master/test/e2e
> * https://github.com/projectcontour/contour/tree/main/test/e2e

From what I see, the projects mentioned above have created their own custom frameworks, which is something I think we should try to avoid in Crossplane, as that means folks will for sure need to get familiar with our framework, plus the added complexity of creating and maintaining our own solution.

phisco:
Thanks @lsviben for the recap, I felt too that the discussion was becoming hard to follow through all the different threads, so I do agree it's better to have more lengthy personal recaps.

So, here are my thoughts about the frameworks we are considering.

### Thoughts and feelings

- **Cucumber**: positive
  - I really like the "self-documenting" nature of the Gherkin feature files, as a user I think those could be really helpful, they can obviously be replaced by writing good documentation, but this would be "automatically tested" documentation that we would be motivated to write in order to have our change/feature properly tested.
  - I don't like so much the regex based matching as it feels like a not so easy to maintain level of indirection, but I don't expect regexes to become much more complex then the ones we see in the example provided.
  - I agree with the fact that it feels like a heavyweight framework, but I think we can say the same for all the frameworks considered here.
  - I agree it might be scaring to newcomers with Go development experience to have this additional level of indirection, but I also feel that once we'll implement a few main steps, a new contributor, even with no Go experience at all, could easily add new test cases by just copy-pasting an existing file and modifying the needed parts there. Obviously we'll be in charge of defining the abstraction and taking care that doesn't become unusable, but I don't see any difference with any other framework in this regard, we would need to maintain a coherent set of helper functions.
  - Being it "just" a testing framework, it's not going to provide any Kubernetes specific helper, we'll need to either use an existing library (e.g. `terratest`) or build our own. This could be seen as both a good and bad thing at the same time, as we won't be forced in any choice, but on the other hand we might end up reimplementing the wheel.
  - I agree embedding YAML files feels like it would be complex to handle, but I'd rather embed them than to have them hidden in some other folder, I've spent too much time due to variables pointing to the wrong file by name. I feel the whole purpose of the proposed format is to be self-contained, so if we decide to go with this I would stick with embedding.
  - I have a feeling the additional layer of indirection introduced by the steps definitions could force us to write better self-contained code, avoiding the mess that other frameworks allow (e.g. global state usage with Ginkgo).

- **Kuttl**: negative
  - Nice and simple, but I think we would rapidly outgrow it.

- **e2e-framework**: slightly positive
  - I agree the project feels early stage, but there could be some value in adopting it at such an early stage as we could contribute to its development.
  - Being it clearly targeting Kubernetes, it comes with a lot of nice helpers to deal with eventual consistency and Kubernetes resources, other than its own way to way to handle target environments.
  - The "Setup/Assert/Teardown" fluent API looks like pretty straightforward, but enough to cover our needs.
  - Less heavyweight as a framework w.r.t. other options, but still requires learning a DSL.

- **Ginkgo/Gomega**: neutral
  - If the Kubernetes community is moving away from it, I'd say that it doesn't make much sense to introduce it in our project now.
  - Definitely an heavyweight framework, tests can easily become hard to follow and requires a lot of discipline to keep clean.
  - Comes with some helpers to deal with eventual consistency, we would still need to roll our own on the Kubernetes side.

### General notes:
- I would prefer not having to interact with external tools that could make error handling harder and tests too brittle, so, for any choice this might apply to, I would rather avoid using `kubectl` as interface to the cluster and tools such as `jq` to manipulate responses.
- I feel like the difference between Cucumber, Ginkgo and e2e-framework (without considering helper libraries provided) is mainly syntactical, as the last two could definitely be converted to the former if we were to hide all the Go code and only leave the descriptions in plain english and the data itself. The main difference being how far from the actual implementation are the flow control and data part kept.

### Conclusions
Overall, at the moment, I feel like I would either try going with Cucumber, because of the nice human-readable narrative provided by feature files, keeping embedded YAML manifests, and spend some time defining the surrounding interfaces to be used (kubectl vs client-go), possible helper libraries (terratest vs e2e-framework/klient vs ...)  and targets' setup (kind or/and envtest). Otherwise, if we decide we don't care about the documentation aspect because we'll cover it through actual documentation, I would fully commit to `e2e-framework` and Go (no hidden YAML fixtures, only structs), and maybe try help shape its development.

design/design-doc-e2e-testing.md
phisco: +1

design/design-doc-e2e-testing.md
phisco: +1 to e2es serving also as documentation being a nice to have, but not a requirement

negz:
@lsviben @pedjak - Apologies for being a little all-over-the-place with my feedback on this PR. I like the summary pattern you've introduced a lot. Here's my attempt.

First, my ranked preferences, starting with my strongest preference:

1. e2e-framework
2. Terratest
3. kuttl
4. Rolling our own framework
5. Ginkgo and Gomega
6. Cucumber

Here's a summary of my thoughts on each. I actually had Cucumber above Ginkgo but after writing out the out the pros and cons of each I moved it to the bottom, sorry folks. 😞 

__e2e-framework__
* Pro: Designed for our specific use case - E2E/integration testing on Kubernetes
* Pro: Designed based on learnings from a lot of highly relevant prior experience (i.e. k/k E2E tests)
* Pro: "Official" Kubernetes SIG project - higher chance of becoming a de-facto standard (like controller-runtime)
* Pro: We don't need to roll our own utilities for interacting with Kubernetes
* Pro: Idiomatic approach to writing Go tests is likely to lower the learning curve for Go developers (i.e. most contributors)
* Pro: Our choice as to whether test inputs and outputs are YAML manifests or Go structs
* Con: Not widely adopted in our space
* Con: Young project

__Terratest__
* Pro: Idiomatic approach to writing Go tests is likely to lower the learning curve for most contributors
* Pro: _In part_ designed for our specific use case - i.e. testing cloud infrastructure
* Pro: We don't need to roll our own utilities for interacting with Kubernetes
* Pro: Our choice as to whether test inputs and outputs are YAML manifests or Go structs
* Pro: Seems quite active and mature 
* Con: Kubernetes is just one of many things it does - perhaps not as focused?
* Con: Seems heavily driven by a single vendor (Terragrunt)
* Con: Not widely adopted in our space

__kuttl__
* Pro: Designed for our specific use case - E2E/integration testing on Kubernetes
* Pro: Writing _only_ YAML manifests is likely to be familiar to _any_ Crossplane contributor (even non-coders)
* Pro: Already in use within the Crossplane ecosystem (i.e. https://github.com/upbound/uptest)
* Con: Writing _only_ YAML limits expressiveness
* Con: Not widely adopted in our space
* Con: Young project

__Rolling our own__
* Pro: Designed for our specific use case - E2E/integration testing on Kubernetes
* Pro: (Presumably) idiomatic approach to writing Go tests is likely to lower the learning curve for most contributors
* Pro: Our choice as to whether test inputs and outputs are YAML manifests or Go structs
* (Huge) Con: We end up with something like e2e-framework or Terragrunt, but we have to do it all ourselves

__Ginkgo__
* Pro: Very mature project - broad adoption within the Go ecosystem and the Kubernetes ecosystem
* Con: Considered an anti-pattern by the Go ecosystem and (some of?) the Kubernetes ecosystem
  * I spent a bunch of time years ago _removing_ Gomega from our codebase and would prefer not to see it return 😆 
* Con: [Unidiomatic Go-based DSL](https://onsi.github.io/ginkgo/#writing-specs) is likely to increase the learning curve for most contributors
  * Somewhat balanced out by the broad use within the Kube ecosystem - folks more likely to have come across it
* Con: Doesn't help _test Kubernetes controllers_
  * Instead it's a different way to write tests in general
  * We'd need to build or choose _another_ framework to make working with Kubernetes convenient 

__Cucumber__
* Pro: Very mature project - broad adoption within the software engineering space in general
* Pro: It's a nice idea in theory - English-language feature files appeal to some (many?) folks
* Pro: Tests act as a form of documentation (without needing to know how to read Go)
* Con: Doesn't help _test Kubernetes controllers_
  * Instead it's a different way to write tests in general
  * We'd need to build or choose _another_ framework to make working with Kubernetes convenient 
* Con: Syntax of English-language Gherkin features has meaning (and that's not obvious)
  * You need to [learn Gherkin](https://cucumber.io/docs/gherkin/reference/) to write it correctly
  * You probably also need to learn regular expressions, which are notoriously hard to read and write
  * For example [Kubedog's syntax docs](https://github.com/keikoproj/kubedog/blob/ac4684d/docs/syntax.md) concern me 
* Con: New language (Gherkin) means new tooling
  * New linters in our build pipeline, new syntax highlighting extension in editors etc
  * Is it even possible to lint and syntax highlight YAML-nested-in-Gherkin?
* Con: In reality most test updates seem to require writing Gherkin _and_ Go
  * Based on a quick sample of a handful of commits under the test directories of 2-3 Godog-based projects
* Con: Unidiomatic approach to writing Go tests (Godog) is likely to increase the learning curve for most contributors
* Con: Not widely adopted in our space

One thing I'm noticing is that I'm personally not completely aligned with the goals this proposal lays out, which may explain why my preferences differ so much. If I quickly run through the goals:

* Support for declarative tests.
  * I don't think this should be a goal.
  * I think kuttl (which comes closest) demonstrates that purely declarative tests are too limiting in at least some cases.
  * None of the other frameworks are actually declarative (even Cucumber - you still write Go).
* Excellent readability for everybody (developers, users, Crossplane community in general)
  * I don't think this should be a goal. 
  * That said, it is certainly _nice-to-have_ - a nice property.
  * I don't think our tests need to be any more or less readable than any other part of our codebase.
  * I think we should optimize for the folks who will read and write tests most - that will be code contributors.
* Black-box testing.
  * Agree this is important.
  * Our black-box is Crossplane+API server+etcd. We don't need to test `kubectl`.
* Parallel test execution.
  * Agree this is important.
* Run tests against an arbitrary system deployment.
  * Agree this is important.
* Test code reusability/maintainability/ease of development
  * Agree maintainability and ease of dependency is important.
  * I wouldn't over-index on reusability - not more than we do elsewhere. Often a little copying helps readability.

design/design-doc-e2e-testing.md
negz: Let me try to frame what I think is the problem we need to solve. Apologies if some of this is stating the obvious.

Our problem is that our unit tests are not sufficient. We rely on an honor system of folks running manual black-box tests to make sure code works in real-world conditions. These tests:
  * Are not very repeatable, or portable for others to run.
  * Are often only run on one environment - i.e. `kind` on the author's laptop.
  * Are not re-run on future changes to catch regressions.
  * Are likely to miss things like performance degradation.
  * Are usually skipped entirely for innocuous seeming things like Renovate dependency bumps.
  * Often don't cover upgrade scenarios.

So I believe our goal is to have black-box tests that solve the above problems. i.e. Are repeatable, portable, run frequently, and catch performance and upgrade issues. We need to both _introduce_ these tests and _ensure folks keep writing them_.

Where I think we're diverging is that I _do not_ believe solving the above problems requires introducing a new testing framework. That is, I don't think the way we write, structure, and run tests today needs to change. I feel the (stdlib) `testing` and `cmp`  packages that we already use are just as well suited to integration and E2E tests as they are to unit tests. You can do a lot with `t.Run()`, `t.Cleanup()`, etc.

I do think as part of implementing E2E tests we will need some convenience utilities - a library - for doing things like starting a control plane, deploying Crossplane, applying configuration (test inputs), waiting for readiness, etc. Ideally that library would be something that already exists rather than something we roll ourselves.

So what I personally would like to see from this proposal is:

* Align on the problems we're solving - is it what I wrote above? Less? More? Completely different?
* Identify a library that will make _black-box testing Crossplane_ (not writing tests in general) more convenient.
* Identify what tests add the most value - what should we start with?
* Figure out how we're going to ensure folks actually write tests going forward.

muvaf:
Just wanted to drop my 2 cents with the disclaimer that I'm not writing code for Crossplane every day nowadays so my thoughts should weigh less than the ones that do.

I'm mostly considering Gherkin and e2e-framework. There are great summaries in this thread about each option but the ones that stand out the most for me is the following:
* Gherkin introduces an inherent indirection which means slight friction for every time you read, write and run the tests and it increases with more and more integration you have in place like IDEs, CI/CD pipelines.
* Gherkin is its own DSL which means we'll need to learn and use it effectively and have to come up with hacks to make it work in cases where we don't want repetition, e.g. I see regex-es and immediately think that they won't stay as is. Probably the weakest part of the chain is those bindings.

The bar to bear these costs is quite high for me given that the alternative is a Go-based framework that we can all be comfortable using from day one and be as flexible as we'd like. Some of the justifications could have been that the audience isn't familiar with Go or the tests validate the public documentation that's written in English but neither seems to be the case with the proposal. If folks really like it for readability, we could make the tests look very similar to those Gherkin files with the right indentation instead of table-driven tests which I suspect is one of the reasons that folks find our tests hard to read. So I'd suggest going for the e2e-framework.

All that said, I think folks who make day-to-day code contributions should make the call given that they are the main audience of these files.

design/design-doc-e2e-testing.md
turkenh: I was under the impression that we would write those binding once and rarely need to extend them. This is why I was looking for [the possibility of defining](https://github.com/crossplane/crossplane/pull/4101#discussion_r1206341582) `claim becomes synchronized and ready` as `<resource> becomes <conditions>...` instead.

pedjak:
@negz 

> Con: Doesn't help test Kubernetes controllers

Interesting view, so what are doing the tests submitted as the part of this PR? They are testing Crossplane as whole, not taking care if something is implemented as controller or not, i.e. we perform black-box testing.

> We'd need to build or choose another framework to make working with Kubernetes convenient

Not really, we should and could use the same approach we advocate to our users (the documentation is full of `kubectl` usage).

> Syntax of English-language Gherkin features has meaning (and that's not obvious) 

In what sense is not obvious? For example, if a steps says `claim becomes synchronized and ready`, what is remained unclear? The syntax and the meanings are something that we control. If a step descriptions need to rephrased to improve clearness, we can do that.

> Is it even possible to lint and syntax highlight YAML-nested-in-Gherkin?

Sure, as mentioned already in one of the comments, there is [linter](https://github.com/vsiakka/gherkin-lint).

>  In reality most test updates seem to require writing Gherkin and Go

  Not really. After adding a few tests, adding new step implementation (i.e. Go code) should happen rarely. 

> You probably also need to learn regular expressions, which are notoriously hard to read and write

We do not need to learn it - we use a regular expression only to bind step to Go function, and that happens only once when we implement that step.

> For example [Kubedog's syntax docs](https://github.com/keikoproj/kubedog/blob/ac4684d/docs/syntax.md) concern me

The proposal is not based anyhow on Kubedog.

bobh66:
I'll add my .02 from the perspective of someone who is NOT a Go developer by trade.

I had a lot of experience with BDD from previous projects (in Python) and my views line up with @ytsarev - it sounded great on paper and lots of people worked really hard to make it work, but it never lived up to the hype and in the end was no more effective than if we had written everything in Python from the start.  We did not have anyone on the business side who was interested in contributing to or maintaining the Gherkin so it all fell on the developers to describe what was being implemented, which may or may not have matched up with what was desired.

My feeling is that a community project will usually not have anyone invested in maintaining the high level Gherkin from a "business" standpoint, so it will end up just validating what was implemented instead of what was actually desired.  Those two things may always/usually be the same in an open-source project, but it eliminates some of the value proposition and adds non-trivial abstraction and complexity for not a lot of value.

I had no experience with Go before I got involved with Crossplane, and while I think I understand most of the basics, I still rely on GoLand extensively to correct what I'm doing, and I really don't understand the reasoning behind a lot of the layers of abstraction and indirection that seem to be common patterns.  As a new contributor this is probably my #1 stumbling block - working my way through all of the indirection/abstraction to figure out what is really going on.  It's hard enough to understand the existing test structure, and how to write new tests (copy existing tests and tweak until they appear to be doing what you think they should be doing).  This gets harder if another layer is placed on top with another interface that has to be created/maintained.

I have no particular knowledge about any of the tools/frameworks under discussion, but I tend to agree with @negz especially relating to introducing new technologies and abstraction layers.

Another possibility might be that Upbound decides to invest in BDD/Gherkin testing for the `uxp` fork while still using e2e-framework or some other solution for the community version.  I don't think they are necessarily mutually exclusive, but I do think that Gherkin will be less useful/more work/less successful in a community project than in a project that has actual business ties that can justify the extra effort and abstraction required.

design/design-doc-e2e-testing.md
pohly: Disclaimer: I am one of the SIG Testing tech leads, primarily because of my work on the in-tree E2E framework.

> In the Golang world, there is https://github.com/onsi/ginkgo, a symmetric tool to Rspec. It was popular in the core Kubernetes testing effort.

I'd like to point out that we recently moved to Ginkgo v2, which addressed a lot of the deficiencies that were identified earlier by people using it for Kubernetes. In particular, we now have:
- Excellent support for defining timeouts and interrupting a running test suite, including running cleanup code after the tests have stopped.
- Intermediate progress reporting with details about what a test is waiting, if requested. This is useful for running tests interactively where waiting for a 5 minute timeout to get a failure message is extremely cumbersome.

See https://www.kubernetes.dev/blog/2023/04/12/e2e-testing-best-practices-reloaded/ for a summary of the recent changes.

Another enhancements that is partly implemented, partly coming include enabling linters to ensure that best practices are being followed.

> As a concerning point, it is [mentioned](https://github.com/kubernetes-sigs/e2e-framework/blob/c3b933fe4c7313a31cbc2ea3cad2d6a2a3c8962f/docs/design/README.md?plain=1#L12) in e2e-framework documentation that moving away from Ginkgo is one of the primary motivation of the project creation.

The direct consequences is that the e2e-framework now doesn't have any of the improvements above. They also don't support per-test log output from the underlying libraries (like client-go). This is annoying when running tests with `go test` (no `-v`) and the output for a failed test then doesn't include that klog output. It gets worse when running tests in parallel because the klog output from different tests gets mixed up.

The in-tree E2E framework is currently not recommended for usage outside of Kubernetes because depending on k/k is discouraged (too many indirect dependencies, not meant to be reused like that). It would be possible to change that by moving the E2E framework into a staging repo. However, this only makes sense for SIG Testing when it helps spread the work of maintaining that framework among more people. It doesn't make sense if we just get more issues filed against Kubernetes :sweat_smile:

design/design-doc-e2e-testing.md
```diff
@@ -0,0 +1,861 @@
...
+		}).Feature()
+
+	// test feature
+	testenv.Test(t, f)
+}
+...
+```
+
+The test steps are run in the order based on their level (Setup, Assess,
+Teardown).
```
pohly, line 658: I've mentioned above that Ginkgo handles cleanup also when a test times out or gets interrupted when running it interactively. I'd be happy to be proven wrong, but because e2e-framework is run through `go test`, I'm pretty sure that the cleanup code will not run when `go test` decides that it is time to stop (global timeout, CTRL-C).

When testing against a throwaway cluster that's not an issue (just recreate it), but otherwise leaving garbage behind can be annoying.

design/design-doc-e2e-testing.md
pohly: Perhaps because Ginkgo by itself doesn't help at all with testing things running in a Kubernetes cluster. That becomes easier when using some additional code that helps with common tasks like connecting to the cluster, handling test namespaces, creating objects and polling state.

The decision to use or not use Ginkgo can make a significant difference for the usability of an E2E framework - see some of my other comments.

negz:
@pedjak 

>> Con: Doesn't help test Kubernetes controllers
>
> Interesting view, so what are doing the tests submitted as the part of this PR? They are testing Crossplane as whole, not taking care if something is implemented as controller or not, i.e. we perform black-box testing.

What I mean is that Cucumber and Gingko are not purpose built to test "Kubernetes stuff". They don't include any utilities to make working with Kubernetes easier (think starting kind clusters, applying manifests, etc). This means that we need to find that functionality elsewhere - either find another library or build it ourselves. In your PR you've built it yourself by shelling out to `kubectl`.

>> We'd need to build or choose another framework to make working with Kubernetes convenient
> 
> Not really, we should and could use the same approach we advocate to our users (the documentation is full of kubectl usage).

By building code that shells out to `kubectl` we're essentially building our own library to work with Kubernetes. Some of the other options (e.g. Terragrunt and e2e-framework) get us this "for free".

>> Syntax of English-language Gherkin features has meaning (and that's not obvious)
>
> In what sense is not obvious? For example, if a steps says claim becomes synchronized and ready, what is remained unclear? The syntax and the meanings are something that we control. If a step descriptions need to rephrased to improve clearness, we can do that.

It's readable without much or any context. That's great. It's not writable without context though. To add a new feature I need to make sure I structure my sentences using syntax that is meaningful to Gherkin, and correctly matches the regular expressions defined in the Go code. So I probably need to be able to read those regular expressions to make sure I'm writing sentences that match them correctly.

>> Is it even possible to lint and syntax highlight YAML-nested-in-Gherkin?
>
> Sure, as mentioned already in one of the comments, there is [linter](https://github.com/vsiakka/gherkin-lint).

Does this linter lint the YAML inside the `""" """` tags, or only the surrounding Gherkin syntax? We need both.

>> In reality most test updates seem to require writing Gherkin and Go
>
> Not really. After adding a few tests, adding new step implementation (i.e. Go code) should happen rarely.

I would like to believe this but have seen little evidence that it's the case. Do you have some?

>> You probably also need to learn regular expressions, which are notoriously hard to read and write
>
> We do not need to learn it - we use a regular expression only to bind step to Go function, and that happens only once when we implement that step.

See my comment above WRT needing to be able to read these regular expressions in order to write Gherkin sentences that match them.

>> For example [Kubedog's syntax docs](https://github.com/keikoproj/kubedog/blob/ac4684d/docs/syntax.md) concern me
>
> The proposal is not based anyhow on Kubedog.

I know that. I'm using Kubedog as an (extreme) example of the kind of syntax one might need to learn to correctly write Gherkin that is meaningful to the underlying code.

design/design-doc-e2e-testing.md
negz: One more datapoint [from Kubernetes Slack](https://kubernetes.slack.com/archives/C09QZ4DQB/p1685476779340529?thread_ts=1685410850.250519&cid=C09QZ4DQB):

> Hello there, we (at Datadog) are undergoing an effort to replace our ginkgo k/k E2E suite with the e2e-framework.
Everyone which interacted with the new tests gave a lot of positive feedback in terms of user experience.
I think that for us the biggest problem at the moment is not being able to run tests in parallel - which I'm planning to help fixing :) I am also writing a new internal controller and writing its tests with the e2e-framework and it's really nice to be able to use the same framework for  different use cases.

design/design-doc-e2e-testing.md
negz: @turkenh @pedjak I spent some time experimenting with e2e-framework in https://github.com/crossplane/crossplane/pull/4187. After doing that, I'm confident we're not making a fair comparison here.

I believe regardless of which of the frameworks we choose it's possible to:

1. Have declarative test fixtures - i.e. YAML manifests as input.
2. Write reusable Go functions to validate that Crossplane does what it should when YAML manifests are applied.

I'm not entirely sure whether @turkenh is comparing the code he's seeing here to the Gherkin language, or to the Go code that powers the Gherkin language.

It does not make sense to compare this code to Gherkin, because Gherkin also need you to write something like this code to function. Gherkin is largely providing and organizing test fixtures that are applied and assessed by Go code (i.e. `godog`).

It does make sense to compare this code to the `godog` code. I believe to the extent that that `godog` code appears simpler, it's because it's largely shelling out to tools like `kubectl` and `jq` to implement the "test logic". I don't think we should do that, but we _could_ just as easily do it with any of the frameworks assessed here.

pedjak:
> I'm not prepared to approve using Cucumber and godog to write our E2E tests, sorry @lsviben and @pedjak.

Thanks a lot for the great feedback @negz 

Going forward, I could see two possible paths:

* close this PR as unmerged, leaving comments and discussion visible that led to such decision. #4187 gets polished and merged and we start to write more tests on top of that in followup PRs, or
* rewrite the proposal in this PR to select e2e-framework as the approach, remove Cucumber tests and potentially add changes from #4187 into it.

For the latter, we would definitely need your help on refining the proposal goals and input on e2e-framework section.

negz:
> Going forward, I could see two possible paths

@pedjak @lsviben Let's close this, and I'll add a one-pager to my POC capturing the decision. I'll link to this PR to cover the alternatives considered.

Thanks again for all your work on this - I know we didn't go with the route you preferred but the work was definitely not in vain. Our discussions have shifted some of my views (e.g. on declarative tests) and I hope to integrate those learnings into the solution we end up with.

PR #4107, opened 2023-05-25
feat: add matchCondition readinessCheck type and use as default

### LINKED ISSUE #4091: Add ReadinessCheck Type `MatchCondition` to Check Custom Conditions

The composition engine considers a resource as `Ready` by [checking the Ready Condition](https://github.com/crossplane/crossplane/blob/c4031aab55c729df6af6a9e4d7b39e3cdd2dcce9/internal/controller/apiextensions/composite/ready.go#L198) of the resource by default. This behavior can be changed by configuring the `readinessChecks` field for a given composed resource with the following types: `MatchString`, `MatchInteger`, `NonEmpty`, and `None`. 

However, with the available options, it is not possible to configure Crossplane to check another condition with a different type (e.g. `TypeAvailable`) or at another path (e.g. `status.atProvider.status.conditions`). See the [related comment](https://github.com/crossplane-contrib/provider-kubernetes/issues/45#issuecomment-1158876167). 

Introduce another ReadinessCheck type `MatchCondition` to check custom conditions at arbitrary paths.

phisco:

Fixes #4091

Currently, if no `readinessCheck` is provided, Crossplane's default behaviour is to consider ready a Composed Resource by checking whether its `Ready` condition has status set tot `True`.

This PR introduces a readiness check of type `MatchCondition`, allowing users to customise which condition Crossplane should use instead of the current hardcoded `Ready` and `True` tuple. But, being set as default for the `readinessChecks` list, also helps users by explicitly showing them the default behaviour.

The patch became bigger due to the need to add this new type in three different places, both the user-facing Composition and FunctionIO APIs and the internal representation of them.

I have:

Manually deploying a Composition with no `spec.resources[*].readinessChecks` set, correctly defaults to a single element in the list readinessCheck of type MatchCondition with type Ready and status `True`. While user-provided ones are preserved if defined.

e.g.
```sh
kubectl apply -f - <<EOF
kind: Composition
metadata:
  name: xpostgresqlinstances.azure.platformref.upbound.io
  labels:
    provider: azure
spec:
  compositeTypeRef:
    apiVersion: azure.platformref.upbound.io/v1alpha1
    kind: XPostgreSQLInstance
  resources:
  - name: postgresqlserver
    readinessChecks:
    - type: None                                   # <= setting only this to None
    base:
      apiVersion: dbforpostgresql.azure.upbound.io/v1beta1
      kind: Server
      spec:
        forProvider:
          version: "11"
  - name: vnetrule
    base:
      apiVersion: dbforpostgresql.azure.upbound.io/v1beta1
      kind: VirtualNetworkRule
      spec:
        forProvider:
          serverNameSelector:
            matchControllerRef: true
EOF
```

Now Running `kubectl get compositions.apiextensions.crossplane.io -o yaml xpostgresqlinstances.azure.platformref.upbound.io` will result in:
```yaml
kind: Composition
metadata:
  annotations:
      {"apiVersion":"apiextensions.crossplane.io/v1","kind":"Composition","metadata":{"annotations":{},"labels":{"provider":"azure"},"name":"xpostgresqlinstances.azure.platformref.upbound.io"},"spec":{"compositeTypeRef":{"apiVersion":"azure.platformref.upbound.io/v1alpha1","kind":"XPostgreSQLInstance"},"resources":[{"base":{"apiVersion":"dbforpostgresql.azure.upbound.io/v1beta1","kind":"Server","spec":{"forProvider":{"version":"11"}}},"name":"postgresqlserver","readinessChecks":[{"type":"None"}]},{"base":{"apiVersion":"dbforpostgresql.azure.upbound.io/v1beta1","kind":"VirtualNetworkRule","spec":{"forProvider":{"serverNameSelector":{"matchControllerRef":true}}}},"name":"vnetrule"}]}}
  creationTimestamp: "2023-05-25T15:06:49Z"
  generation: 1
  labels:
    provider: azure
  name: xpostgresqlinstances.azure.platformref.upbound.io
  resourceVersion: "2173"
  uid: 14d5cc2b-7eb2-4b46-95ad-76e78423e0e2
spec:
  compositeTypeRef:
    apiVersion: azure.platformref.upbound.io/v1alpha1
    kind: XPostgreSQLInstance
  publishConnectionDetailsWithStoreConfigRef:
    name: default
  resources:
  - base:
      apiVersion: dbforpostgresql.azure.upbound.io/v1beta1
      kind: Server
      spec:
        forProvider:
          version: "11"
    name: postgresqlserver
    readinessChecks:               # <= preserved user input
    - type: None
  - base:
      apiVersion: dbforpostgresql.azure.upbound.io/v1beta1
      kind: VirtualNetworkRule
      spec:
        forProvider:
          serverNameSelector:
            matchControllerRef: true
    name: vnetrule
    readinessChecks:
    - type: MatchCondition            # <= correctly defaulted
      matchCondition:
        status: "True"
        type: Ready
      
```


apis/apiextensions/fn/io/v1alpha1/functionio_types.go
```diff
@@ -227,6 +230,46 @@ type DesiredReadinessCheck struct {
...
+	if rc.MatchInteger != nil {
+		c.MatchInteger = *rc.MatchInteger
+	}
+	if rc.MatchCondition != nil {
+		c.MatchCondition = &v1.MatchConditionReadinessCheck{
+			Type:   rc.MatchCondition.Type,
+			Status: rc.MatchCondition.Status,
+		}
+	}
+	return c
```
negz, line 272: Is this change necessary? I generally prefer to avoid one package of user-facing API types importing another package of user-facing API types where possible. Largely to avoid a "slippery slope" that couples the two APIs more strongly - i.e. adding fields to this API whose types are imported from `apiextensions/v1`.

Would it be possible to keep the existing pattern from `ready.go` - i.e. translate both user-facing types to an internal type?

apis/apiextensions/v1/composition_common.go
```diff
@@ -142,9 +146,41 @@ type ReadinessCheck struct {
...
+	}
+	if m.Status == "" {
+		return field.Required(field.NewPath("status"), "cannot be empty for type MatchCondition")
+	}
+	return nil
 }
 
 // Validate checks if the readiness check is logically valid.
+//
+//nolint:gocyclo // This function is not that complex, just a switch
```
negz, line 183: Slight preference for putting this after the `{` on the next line. I'm pretty sure that will stop it from showing up as parted of the generated godoc for this function.

internal/controller/apiextensions/composite/ready.go
```diff
@@ -77,37 +119,20 @@ func ReadinessChecksFromTemplate(t *v1.ComposedTemplate) []ReadinessCheck {
...
+		out[i] = ReadinessCheckFromV1(&t.ReadinessChecks[i])
 	}
 	return out
 }
 
-// ReadinessChecksFromDesired derives readiness checks from the supplied desired
+// ReadinessChecksFromDesiredResource derives readiness checks from the supplied desired
 // resource.
-func ReadinessChecksFromDesired(dr *iov1alpha1.DesiredResource) []ReadinessCheck {
+func ReadinessChecksFromDesiredResource(dr *iov1alpha1.DesiredResource) []ReadinessCheck {
```
negz, line 129: Does this function need to be renamed?

I have a slight preference for the shorter name - `DesiredResource` already appears in the signature as the type of the argument. I don't feel super strongly, but if you do want to give this a longer name let's be consistent and rename `ReadinessChecksFromTemplate` to `ReadinessChecksFromComposedTemplate` too.

internal/controller/apiextensions/composite/ready.go
```diff
@@ -67,6 +69,72 @@ type ReadinessCheck struct {
...
+		out.MatchCondition = &MatchConditionReadinessCheck{
+			Type:   in.MatchCondition.Type,
+			Status: in.MatchCondition.Status,
+		}
+	}
+	return out
+}
+
+// readinessCheckFromDesired derives a ReadinessCheck from the supplied iov1alpha1.DesiredReadinessCheck.
+func readinessCheckFromDesired(in *iov1alpha1.DesiredReadinessCheck) ReadinessCheck {
```
phisco, line 121: not sure whether to keep these private, everything else is public but used only in the package, so these might as well be public

apis/apiextensions/fn/io/v1alpha1/functionio_types.go
phisco: I inverted the dependency moving the conversion code to be alongside the internal representation package, I would have probably preferred have some "ToInternal" methods, just to keep this logic nearer to the actual struct, so that if someone adds a new feature can see it needs to be propagated to the internal representation too, but that would cause a dependency cycle. That was the original reason to use v1 as intermediate representation

lsviben (approved):
Tested the defaulting part, looks good.

added my 2 cents to the naming discussion, but thats a non-blocker for me, so approved

internal/controller/apiextensions/composite/ready.go
negz: The public functions are a bad(?) habit of mine. 😬

I like to write everything as if it could be used as a public-facing library. I like that doing this:

* Encourages me to think about dependency injection, testability, etc.
* Requires me to write documentation for every function and method. (Linters complain if they're missing)

In practice we have everything under an `internal/` package, so only other packages inside of Crossplane _could_ use any of these public things.

I have a slight preference for keeping stuff public (inside) Crossplane, but I do admit it's a little bit of a "weird take". 😄 Perhaps here we could stick with the pattern of the rest of the file (i.e. public functions) and separately chat about whether we should be using more private functions and types?

lsviben: Ok, thanks for the explanation. Makes sense for me, lets keep them public.

apis/apiextensions/fn/io/v1alpha1/functionio_types.go
```diff
@@ -227,6 +228,22 @@ type DesiredReadinessCheck struct {
...
+// MatchConditionReadinessCheck is used to indicate how to tell whether a resource is ready
+// for consumption
+type MatchConditionReadinessCheck struct {
+	// Type indicates the type of condition you'd like to use.
+	// +kubebuilder:default="Ready"
+	Type string `json:"type,omitempty"`
+
+	// Status is the status of the condition you'd like to match.
+	// +kubebuilder:default="True"
+	Status string `json:"status,omitempty"`
```
negz, line 246: These fields aren't optional, right? If so I think we should remove the `omitempty` per https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required (same elsewhere).

negz (approved):
I think my comment about the function name still stands, but not a blocker.

apis/apiextensions/fn/io/v1alpha1/functionio_types.go
phisco: you are right, these are actually required fields with a default, I removed them everywhere, which correctly resulted in these fields being marked as required in the schema.

internal/controller/apiextensions/composite/ready.go
turkenh: > I wouldn't go with ReadinessChecksFromComposedTemplate instead of ReadinessChecksFromTemplate as that would be unnecessary

I would prefer that for consistency, but okay with either.  
Please let me know if you want to make a change or prefer to go as is so we can merge the PR.

PR #4110, opened 2023-05-25
Remove crun v1.5 distinction

hasheddan:

Unfortunately every build of the xfn image is using whatever the latest version of crun is available at that time, which results in non-reproducible build steps. This removes the comment specifying that we are using crun v1.5 to eliminate confusion until we have a better strategy.

I have:

n/a -- just cleaning up a comment

### DIFF
--- a/cluster/images/xfn/Dockerfile
+++ b/cluster/images/xfn/Dockerfile
@@ -1,4 +1,4 @@
-# This is debian:bookworm-slim (i.e. Debian 12, testing), which has crun v1.5.
+# This is debian:bookworm-slim (i.e. Debian 12, testing)
 FROM debian:bookworm-slim@sha256:2fd76b2b011a8b171f9ee0540fb74e9be0dc55fafc8f7f2858a482c0e7776ebc
 
 ARG TARGETOS


bobh66 (approved):
Is it possible to install a specific version of crun so it is always reproducible?

hasheddan (commented):
@bobh66 there are some suggestions for how we might do so in the comments below:

> TODO(negz): Find a better way to get an OCI runtime? Ideally we'd grab a
> static build of crun (or runc) that we could drop into a distroless image. We
> slightly prefer crun for its nascent WASM and KVM capabilities, but they only
> offer static builds for amd64 and arm64 and building our own takes a long
> time.

I think producing a base image that we build in a separate CI process (so that we aren't rebuilding `crun` every time as mentioned), would be a great idea in the future.

negz (approved):

negz:
FWIW I tried a whole bunch of different things to get a `crun` binary before settling on what we have here. I don't love it but it was the least-bad option I could come up with. I'm not sure we'll need new features from `crun` very often, so one low-effort option would be to just wait until Debian stable has a non-ancient version of `crun` and just switch to that. (I presume stable will churn much less than testing.)

negz:
> one low-effort option would be to just wait until Debian stable

That's apparently happening June 10 per https://lists.debian.org/debian-devel-announce/2023/04/msg00007.html

Issue #3601, opened 2023-01-19
Mark `ControllerConfig` as deprecated in Crossplane `v1.11.0`


hasheddan:
### What problem are you facing?

The `ControllerConfig` API has been marked as `v1alpha1` since it was added to Crossplane in `v0.14.0`. There is a [longstanding issue](https://github.com/crossplane/crossplane/issues/2468) tracking promoting it to `v1beta1`, but there has been push back for a variety of reasons:
- The `ControllerConfig` API has grown to essentially mirror the k8s `Deployment` API, without being fully compliant. This results in some community members needing fields that are not present. The quick solution here would be to make the `ControllerConfig` mirror `Deployment` 1:1.
- The current API, as well as an API that became identical to `Deployment`, forces all Providers to be run as `Deployments`. While this is the common mode of operation today, it is not necessarily the only one that should be supported. Provider controllers need to be scheduled to compute and be given proper access to the Kubernetes API server, but the type of compute is not fundamentally constrained to being provided by Kubernetes, and certainly not the same Kubernetes cluster where core Crossplane is running.
- `ControllerConfig` only considers one Provider during processing. You cannot author a `ControllerConfig` with logic that depends on the state of other Providers or other external information.

Ideally, users would be able to have full freedom in determining how Provider controllers are run. While many will likely find a solution similar to the built-in one we have today acceptable, there are others who will want to adopt other implementations, or even design their own. Decoupling that functionality from the core Crossplane logic means that iteration and experimentation can happen more freely. The current thinking is that functionality similar to what is described will be implemented via a mechanism that resembles the [Provider Runtime Interface RFC](https://github.com/crossplane/crossplane/issues/2671).

### How could Crossplane help solve your problem?

[Alpha features](https://docs.crossplane.io/v1.10/reference/feature-lifecycle/#alpha-features) in Crossplane are subject to removal from one version to the next without warning. However, `ControllerConfig` was introduced prior to this policy being established and so despite being an alpha API, it is **not** off by default. There is also strong signal (based on PRs / issues / slack activity) that many folks use the `ControllerConfig` API today to configure functionality such as [IAM Roles for Service Accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html).

[...]

PR #4124, opened 2023-05-30
Clarify ControllerConfig deprecation message to reduce confusion

### LINKED ISSUE #3903: `ControllerConfig` deprecation leads to log warning messages without which API is affected

jbw976:

This PR clarifies the ControllerConfig deprecation message to make it more clear that although the type has been marked as deprecated, it will not be removed until there is a suitable alternative solution.

I've also added a suggestion from this [blog post](https://kubernetes.io/blog/2020/09/03/warnings/#customize-client-handling) to use a new warning handler that deduplicates warnings from the k8s REST API.  Warnings should only be logged once, as opposed to multiple times.

Discussion about a future replacement for ControllerConfig can be found in #3601 and #2671.

Fixes #3903

I advocate backporting to v1.12 and v1.11 (the version where this type was deprecated) to reduce confusion in those releases as well.

I have:

I have tested locally on a kind cluster with kind.sh, and observed the log output as well as output when applying and getting `ControllerConfig` objects.

```console
❯ make generate
❯ make build
❯ KUBE_IMAGE=kindest/node:v1.27.1 ./cluster/local/kind.sh up
❯ ./cluster/local/kind.sh helm-install
```

The first log output before any `ControllerConfig` objects have been applied:

```console
❯ kubectl -n crossplane-system logs -lapp=crossplane -f

Defaulted container "crossplane" out of: crossplane, crossplane-init (init)
I0602 13:20:08.281355       1 leaderelection.go:248] attempting to acquire leader lease crossplane-system/crossplane-leader-election-core...
I0602 13:20:08.389007       1 leaderelection.go:258] successfully acquired lease crossplane-system/crossplane-leader-election-core
Warning: ControllerConfig.pkg.crossplane.io/v1alpha1 has been deprecated and will be removed in a future release, but only after a comparable alternative is available.
```

The content of a file with a `ControllerConfig` and a `Provider` referencing it:
```yaml
❯ cat provider-cc.yaml
kind: ControllerConfig
metadata:
  name: debug-config
spec:
  args:
    - --debug
---
kind: Provider
metadata:
  name: provider-aws
spec:
  package: xpkg.upbound.io/crossplane-contrib/provider-aws:v0.35.0
  controllerConfigRef:
    name: debug-config
```

Applying the `ControllerConfig` object:
```console
❯ k apply -f provider-cc.yaml
Warning: ControllerConfig.pkg.crossplane.io/v1alpha1 has been deprecated and will be removed in a future release, but only after a comparable alternative is available.
controllerconfig.pkg.crossplane.io/debug-config created
provider.pkg.crossplane.io/provider-aws created
```

Getting `ControllerConfig` objects:
```console
❯ k get controllerconfig
Warning: ControllerConfig.pkg.crossplane.io/v1alpha1 has been deprecated and will be removed in a future release, but only after a comparable alternative is available.
NAME           AGE
debug-config   39s
```

The logs after letting the pod run for awhile, note that no further warning statements seem to be logged.

```
❯ kubectl -n crossplane-system logs -lapp=crossplane -f

Defaulted container "crossplane" out of: crossplane, crossplane-init (init)
I0602 13:20:08.281355       1 leaderelection.go:248] attempting to acquire leader lease crossplane-system/crossplane-leader-election-core...
I0602 13:20:08.389007       1 leaderelection.go:258] successfully acquired lease crossplane-system/crossplane-leader-election-core
Warning: ControllerConfig.pkg.crossplane.io/v1alpha1 has been deprecated and will be removed in a future release, but only after a comparable alternative is available.
```

### DIFF
--- a/cmd/crossplane/core/core.go
+++ b/cmd/crossplane/core/core.go
@@ -20,6 +20,7 @@ package core
 import (
 	"net/http"
 	"net/http/pprof"
+	"os"
 	"path/filepath"
 	"time"
 
@@ -27,6 +28,7 @@ import (
 	"github.com/google/go-containerregistry/pkg/name"
 	"github.com/spf13/afero"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/leaderelection/resourcelock"
 	ctrl "sigs.k8s.io/controller-runtime"
 
@@ -135,6 +137,11 @@ func (c *startCommand) Run(s *runtime.Scheme, log logging.Logger) error { //noli
 	if err != nil {
 		return errors.Wrap(err, "Cannot get config")
 	}
+
+	cfg.WarningHandler = rest.NewWarningWriter(os.Stderr, rest.WarningWriterOptions{
+		// Warnings from API requests should be deduplicated so they are only logged once
+		Deduplicate: true,
+	})
 
 	mgr, err := ctrl.NewManager(ratelimiter.LimitRESTConfig(cfg, c.MaxReconcileRate), ctrl.Options{
 		Scheme:     s,


contributing/README.md
```diff
@@ -613,7 +613,7 @@ make
 ./cluster/local/kind.sh help
 
 # Start a new kind cluster. Specifying KUBE_IMAGE is optional.
-KUBE_IMAGE=kindest/node:v1.23.0 ./cluster/local/kind.sh up
+KUBE_IMAGE=kindest/node:v1.27.1 ./cluster/local/kind.sh up
```
phisco, line 616: Worth automating this too using renovate, I'll do that in a separate pr

phisco (approved):

jbw976:
@phisco looks like the [trivy-scan-fs](https://github.com/crossplane/crossplane/actions/runs/5120074615/jobs/9206132153?pr=4124) is failing with:

```
go.mod (gomod)
==============
Total: 1 (HIGH: 1, CRITICAL: 0)

┌────────────────────────────────┬───────────────┬──────────┬────────────────────┬───────────────┬───────────────────────────────────────────┐
│            Library             │ Vulnerability │ Severity │ Installed Version  │ Fixed Version │                   Title                   │
├────────────────────────────────┼───────────────┼──────────┼────────────────────┼───────────────┼───────────────────────────────────────────┤
│ github.com/docker/distribution │ CVE-2023-2253 │ HIGH     │ 2.8.1+incompatible │ 2.8.2-beta.1  │ DoS from malicious API request            │
│                                │               │          │                    │               │ https://avd.aquasec.com/nvd/cve-2023-2253 │
└────────────────────────────────┴───────────────┴──────────┴────────────────────┴───────────────┴───────────────────────────────────────────┘
```

Is this something Renovate already found for us? 🤔

phisco:
Yes, it's not detecting that because it's an indirect dependency which was not yet updated in our direct dependency (go-containerregistry). We could set it up to bump also indirect dependencies in case of security issues, but that's off by default as it's not guaranteed to build.

It's not a required step, so can Ben ignored while we fine tune trivy's config

hasheddan (commented):
Thanks @jbw976! Did you happen to see if there was an easy way to suppress the deprecation warning in core crossplane logs ([ref](https://github.com/crossplane/crossplane/issues/3903#issuecomment-1481610618))? The intention was for this to be shown when users are interacting via `kubectl`, but it doesn't provide much value in the context of controllers reconciling `ControllerConfig`.

jbw976:
I did not look too hard @hasheddan, but an outcome where the crossplane pod logs do not emit this at all would be nice! I will dig a bit deeper there to try to find a solution and update this PR with my findings 🙇

hasheddan (commented):
> I did not look too hard @hasheddan, but an outcome where the crossplane pod logs do not emit this at all would be nice! I will dig a bit deeper there to try to find a solution and update this PR with my findings 🙇

Thanks @jbw976! I don't think this necessarily needs to be a blocker as what you have added here is already an improvement, but if there is a quick way to solve the log issue I would be in favor 🙂

turkenh (approved):

cmd/crossplane/core/core.go
```diff
@@ -136,6 +138,11 @@ func (c *startCommand) Run(s *runtime.Scheme, log logging.Logger) error { //noli
 		return errors.Wrap(err, "Cannot get config")
 	}
 
+	cfg.WarningHandler = rest.NewWarningWriter(os.Stderr, rest.WarningWriterOptions{
```
jbw976, line 141: @hasheddan @turkenh what do you think of this approach I found from a Jordon Liggitt blog post? https://kubernetes.io/blog/2020/09/03/warnings/#customize-client-handling

It does indeed result in k8s REST API warnings only being printed once per warning, but I can't say I fully understand other potential impact of making this change, so it would be good to hear your thoughts! 🙇‍♂️

hasheddan: Thanks for finding this @jbw976! I was initially wondering if we could turn off warnings entirely:

```
  // set the warning handler for this client to ignore warnings
  config.WarningHandler = rest.NoWarnings{}
 ```

My rationale here is that warnings emitted by core Crossplane are not necessarily something I would consider “user actionable”. Perhaps the exception to that rule is if a user is composing resources and is using a deprecated API in their `Composition`. A core Crossplane log _could_ encourage them to update their `Composition` to a newer version.

It looks like the only downside of using the deduplication strategy is increased memory usage, but it looks like memory usage should [scale linearly](https://github.com/kubernetes/client-go/blob/7c4f8c88f6cb8817c04724e28bc783f3fd1f4b00/rest/warnings.go#L117) with the number of _type_ of warnings, not the total number of warnings, so I’m not too concerned about that (unless we start experiencing some warning that has a dynamic message that includes a time stamp, counter, etc.).

Overall, it seems like deduplication is at least a good first step, and we can make the decision whether to turn off entirely in the future if needed.

hasheddan (approved):

cmd/crossplane/core/core.go
jbw976: That all makes sense to me @hasheddan, thank you for diving into this a bit and then sharing your thoughts! 🙇‍♂️

PR #4125, opened 2023-05-30
Use a programmatic identifier as reason for Waiting condition.

### LINKED ISSUE #4094: Claim "Ready" condition has a free text message in reason field
### What happened?

The status.conditions.[].reason field for the "Ready" condition on Claims contains a free text message instead of a programmatic identifier as specified [here](https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/types.go#L1510-L1520).

For example:
```yaml
status:
  conditions:
  - lastTransitionTime: "2023-05-16T12:18:10Z"
    reason: ReconcileSuccess
    status: "True"
    type: Synced
  - lastTransitionTime: "2023-05-16T12:18:10Z"
    reason: Composite resource claim is waiting for composite resource to become Ready
    status: "False"
    type: Ready
```

### How can we reproduce it?

Create a claim and look at the status.

Crossplane version: 1.12.1

jeanduplessis:

Fixes #4094

I have:

Re-ran the unit tests.

### DIFF
--- a/internal/controller/apiextensions/claim/reconciler.go
+++ b/internal/controller/apiextensions/claim/reconciler.go
@@ -47,11 +47,6 @@ const (
 	reconcileTimeout = 1 * time.Minute
 )
 
-// Reasons a composite resource claim is or is not ready.
-const (
-	ReasonWaiting = "Composite resource claim is waiting for composite resource to become Ready"
-)
-
 // Error strings.
 const (
 	errGetClaim           = "cannot get composite resource claim"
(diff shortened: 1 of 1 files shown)


internal/controller/apiextensions/claim/reconciler.go
```diff
@@ -586,6 +586,7 @@ func Waiting() xpv1.Condition {
 		Type:               xpv1.TypeReady,
 		Status:             corev1.ConditionFalse,
 		LastTransitionTime: metav1.Now(),
-		Reason:             ReasonWaiting,
+		Reason:             xpv1.ConditionReason("Waiting"),
+		Message:            ReasonWaiting,
```
negz, line 590:
```suggestion
		Message:            MessageWaiting,
```

I'd probably either rename this constant or just add the message string here inline (instead of using a constant).

negz (approved):

internal/controller/apiextensions/claim/reconciler.go
jeanduplessis: I've moved the string inline instead of using a constant.

turkenh (approved):

Issue #4127, opened 2023-05-30
Fine tune indirect vulnerabilities management


phisco:
### What problem are you facing?

Right now, both `crossplane/crossplane` and `crossplane/crossplane-runtime`, are using `github.com/docker/distribution v2.8.1+incompatible` which is affected by [CVE-2023-2253](https://github.com/advisories/GHSA-hqxw-f8mx-cpmw), which is triggering our trivy scan in CI and being correctly reported via vulnerability alerts from Dependabot (which Renovate then uses):
```
# In crossplane/crossplane

$ go mod why -m github.com/docker/distribution
# github.com/docker/distribution
github.com/crossplane/crossplane/cmd/crank
github.com/google/go-containerregistry/pkg/v1/remote
github.com/google/go-containerregistry/pkg/v1/remote/transport
github.com/docker/distribution/registry/client/auth/challenge

# In crossplane/crossplane-runtime

go mod why -m github.com/docker/distribution
# github.com/docker/distribution
github.com/crossplane/crossplane-runtime/apis
github.com/bufbuild/buf/cmd/buf
github.com/bufbuild/buf/private/buf/cmd/buf
github.com/bufbuild/buf/private/buf/cmd/buf/command/alpha/plugin/pluginpush
github.com/bufbuild/buf/private/bufpkg/bufplugin/bufplugindocker
github.com/docker/docker/client
github.com/docker/distribution/reference
```

As you can see above both repos indirectly import `github.com/docker/distribution/reference`, because of two different outdated direct dependencies:
- `xp/xp-runtime`: requires directly `github.com/bufbuild/buf` `v1.10.0`, now at [v1.19.0](https://github.com/bufbuild/buf/tree/v1.19.0)
- `xp/xp`: requires directly `github.com/google/go-containerregistry` `v0.14.0`, now at [v0.15.2](https://github.com/google/go-containerregistry/releases/tag/v0.15.2)

Both direct dependencies have been updated in the meantime and have bumped these dependencies, so manually bumping the would solve the issue at hand.

However, I feel it's worth taking the chance to re-evaluate our config.

In fact, to avoid too much noise from Renovate, we disabled all non-security related updates of Go dependencies on both repos, but given that indirect dependencies updates are disabled by default, as those could break the build and should come through the direct dependencies first. Unfortunately it looks like Renovate is not able to transitively propagate the vulnerability alert to the direct dependency, leaving the trivy scan in CI failing.

[...]

PR #4133, opened 2023-05-31
chore(renovate): enable PRs for all non-kubernetes dependencies

phisco:

Simplified config to address #4127.

We initially decided to only have security-related updates to dependencies, see https://github.com/crossplane/crossplane/pull/3575#discussion_r1072854569, but that complicates a lot dealing with indirect dependencies (see #4127), I feel like we now have been using renovate enough to switch back to opening PRs for all updates only on master, we can always close the PRs if we don't think we need them, but at least we'll know about the updates and we can always reopen them if needed. We'd still keep the recently introduced rule to disable kubernetes related updates, as those should come in as part of `crossplane-runtime` version bumps first.

We shouldn't fear regressions being introduced, we have a strong suite of unit tests and work is being done to improve also on the E2Es, #4101.

I have:

Deployed in my fork, see https://github.com/phisco/crossplane/issues/137 for the PRs that would be created at the moment.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -21,7 +21,7 @@
   "labels": ["automated"],
   "regexManagers": [
     {
-      "description": "Bump Go version ued in workflows",
+      "description": "Bump Go version used in workflows",
       "fileMatch": ["^\\.github\\/workflows\\/[^/]+\\.ya?ml$"],
       "matchStrings": [
         "GO_VERSION: '(?<currentValue>.*?)'\\n"
@@ -69,12 +69,12 @@
       matchBaseBranches: [ "/^release-.*/"],
       enabled: false,
     }, {
-      "description": "Still update docker images on release branches though",
+      "description": "Still update Docker images on release branches though",
       "matchDatasources": ["docker"],
       matchBaseBranches: [ "/^release-.*/"],
       enabled: true,
     }, {
-      "description": "Only get docker image updates every 2 weeks to reduce noise",
+      "description": "Only get Docker image updates every 2 weeks to reduce noise",
       "matchDatasources": ["docker"],
       "schedule": ["every 2 week on monday"],
       enabled: true,
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -1,7 +1,8 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": [
-    "config:base"
+    "config:base",
+    "helpers:pinGitHubActionDigests"
   ],
 // We only want renovate to rebase PRs when they have conflicts,
 // default "auto" mode is not required.
@@ -20,8 +21,7 @@
   "labels": ["automated"],
   "regexManagers": [
     {
-// We want a PR to bump Go versions used through env variables in any Github
-// Actions, taking it from the official Github repository.
+      "description": "Bump Go version ued in workflows",
       "fileMatch": ["^\\.github\\/workflows\\/[^/]+\\.ya?ml$"],
       "matchStrings": [
         "GO_VERSION: '(?<currentValue>.*?)'\\n"
@@ -29,9 +29,7 @@
       "datasourceTemplate": "golang-version",
       "depNameTemplate": "golang"
     }, {
-// We want a PR to bump golangci-lint versions used through env variables in
-// any Github Actions or Makefile, taking it from the official Github
-// repository tags.
+      "description": "Bump golangci-lint version in workflows and the Makefile",
       "fileMatch": ["^\\.github\\/workflows\\/[^/]+\\.ya?ml$","^Makefile$"],
       "matchStrings": [
         "GOLANGCI_VERSION: 'v(?<currentValue>.*?)'\\n",
@@ -41,9 +39,7 @@
       "depNameTemplate": "golangci/golangci-lint",
       "extractVersionTemplate": "^v(?<version>.*)$"
     }, {
-// We want a PR to bump the helm version used through env variables in
-// any Github Actions or Makefile, taking it from the official Github
-// repository tags.
+      "description": "Bump helm version in the Makefile",
       "fileMatch": ["^Makefile$"],
       "matchStrings": [
         "HELM3_VERSION = (?<currentValue>.*?)\\n"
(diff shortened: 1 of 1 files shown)


.github/renovate.json5
```diff
@@ -20,18 +21,15 @@
   "labels": ["automated"],
   "regexManagers": [
     {
-// We want a PR to bump Go versions used through env variables in any Github
-// Actions, taking it from the official Github repository.
+      "description": "Bump Go version ued in workflows",
```
HonkingGoose, line 24:
```suggestion
      "description": "Bump Go version used in workflows",
```

.github/renovate.json5
```diff
@@ -66,23 +60,26 @@
...
+// be at the beginning, high priority at the end
   "packageRules": [
-// We don't want dependency updates which are not security related for
-// release branches, except for docker images
     {
+      "description": "Ignore non-security related updates to release branches",
       matchBaseBranches: [ "/^release-.*/"],
       enabled: false,
     }, {
+      "description": "Still update docker images on release branches though",
```
HonkingGoose, line 72:
```suggestion
      "description": "Still update Docker images on release branches though",
```

.github/renovate.json5
```diff
@@ -66,23 +60,26 @@
...
       "matchDatasources": ["docker"],
       matchBaseBranches: [ "/^release-.*/"],
       enabled: true,
     }, {
+      "description": "Only get docker image updates every 2 weeks to reduce noise",
```
HonkingGoose, line 77:
```suggestion
      "description": "Only get Docker image updates every 2 weeks to reduce noise",
```

HonkingGoose (commented):
I see you're using the `helpers:pinGitHubActionDigests` preset, and use `description` fields, like I suggested. 😄 

There's a typo, and the proper noun `Docker` should be capitalized.

negz:
@phisco If it were possible for Renovate to catch when we needed to update an indirect dependency due to a known vulnerability I would still prefer that. Given that doesn't seem possible, I mostly feel okay with this change.

I'm not really worried about regressions, or increased maintenance work due to extra Renovate PRs to review. My only concern is (or was) that semi-automatically updating all of our dependencies whenever a new version was available could negate some of the benefits of Go's minimal version selection. I re-read https://research.swtch.com/vgo-principles today to refresh my memory on what those benefits are, and to see whether my concern was founded.

I believe the short version is that given a hypothetical diamond dependency:

```mermaid
  graph TD;
      crossplane-->controller-runtime;
      crossplane-->client-go;
      controller-runtime-->yaml;
      client-go-->yaml;
```

Where:

* The latest available `yaml == 1.5`
* `controller-runtime` requires `yaml <= 1.0`
* `client-go` requires `yaml <= 1.1`

Most package managers would pick (and lock) `yaml == 1.5`. Go would instead pick `yaml == 1.1` - i.e. the oldest possible version that satisfies all constraints. It chooses to do this because:

* It enables reproducible builds without using a lock file (which is hard to make work for libraries).
* It lets you avoid dealing with incompatible diamond dependencies until you actually need to. (e.g. To address a vulnerability, or get access to new functionality.)

I don't think semi-automatically updating the minimum versions we depend on hurts reproducible builds in any way. The nature of minimal version selection ensures any particular commit of Crossplane (or something depending on a particular commit of  crossplane-runtime) will be reproducible, dependencies wise.

It does seem like semi-automatically updating the minimum versions we depend on will mean we'll need to address incompatible diamond dependencies before we might otherwise need to. Not really a big deal, given we can just close the PR if there's no obvious benefit to bumping and we don't feel like dealing with it. One consideration though is that for crossplane-runtime (i.e. a library) we'll effectively be signing up anyone using it to resolve any conflicts that result from a bunch of potentially unnecessary dependency updates any time they update to a new version of crossplane-runtime.

negz (approved):

phisco:
@negz I'm not sure I've got your point, but in you example, if `yaml` is an indirect dependency, you won't get a PR to bump it per se, it will only be bumped to the minimum version selected by the MVS algorithm.

The only difference with the current behaviour would be that now we bump only direct dependencies and only in case of vulnerabilities, and there is no way to get those updates on indirect ones and no way to transitively trigger bumps on direct dependencies importing affected indirect ones, while enabling all updates to direct dependencies we'll hopefully get less vulnerabilities on indirect dependencies. This should happen because the indirect dependency will be bumped as soon as the minimum version required by all our up-to-date direct dependencies will be higher than the affected one, which if our direct dependencies are healthy and maintained, should happen pretty fast.

HonkingGoose:
The Renovate docs mention indirect updates in the `gomod` manager section: [^1]

> Indirect updates are disabled by default. To enable them, add a package rule such as:
>
> 
> ```json
> {
>   "packageRules": [
>     {
>       "matchManagers": ["gomod"],
>       "matchDepTypes": ["indirect"],
>       "enabled": true
>     }
>   ]
> }
> ```

I also found a related issue:

- https://github.com/renovatebot/renovate/issues/12999

[^1]: [Renovate docs, `gomod` manager, Post-update options](https://docs.renovatebot.com/modules/manager/gomod/#post-update-options)

phisco:
@HonkingGoose yes, but as per the discussion [here](https://github.com/renovatebot/renovate/discussions/22505#discussioncomment-6040777), it's not possible to enable indirect update only on security vulnerabilities

negz:
> I'm not sure I've got your point

@phisco My understanding is that with this change in place, Renovate will bump all Go dependencies whenever a new version of a dependency is available. So assuming we merge all the Renovate PRs, the versions listed in `go.mod` will always be the latest possible versions. Is that right?

The big comment I wrote earlier was really just me thinking out loud to try understand how that relates to "idiomatic" Go version selection. My conclusion was they seem pretty compatible. Go obviously still uses MVS at build-time, it's just that we'll always be telling it the minimum allowed version is the latest available version, even if there were no _known_ security issues with an older version, and even if there's no features we or any other dependency needs in the latest version.

The only downside I could see is that proactively updating all our dependencies means we also have to proactively resolve new dependency conflicts if and when they appear. The good thing is we'd be doing this in order to get a Renovate PR to merge, not in order to unbreak our build.

So TL;DR assuming all this works like I think it works I feel good about merging. Should I go ahead and do that?

phisco:
> So assuming we merge all the Renovate PRs, the versions listed in go.mod will always be the latest possible versions. Is that right?

@negz this will be true only for direct dependencies, indirect ones will only be updated if the MVS algorithm decides it's time to do so because all direct dependencies require an higher minimum version.

And yes, please, you can merge this 🙏

PR #4174, opened 2023-06-13
fix(xfn): close grpc connection with errors on options

phisco:

Closing the gRPC connection to the function runner in case of errors while applying options, to avoid leakages.

I didn't move it to a `defer` as it looks there is only one case in which we care about it, and handling that properly would have been not so straightforward, but I'm open to discussing it.

Added comment to avoid forgetting about it in the future.

I have:

Locally deployed.


negz (approved):
Nice, thank you!

PR #4188, opened 2023-06-15
Return nil error when we successfully run a MatchCondition check

negz:

I discovered this bug (which only exists in master) while adding E2E tests in https://github.com/crossplane/crossplane/pull/4187. 🙂 

I have:

I found that the Composition E2E tests I implemented in https://github.com/crossplane/crossplane/pull/4187 failed without this change, but passed after it. I also added some unit tests to cover this case.

### DIFF
--- a/internal/controller/apiextensions/composite/ready.go
+++ b/internal/controller/apiextensions/composite/ready.go
@@ -227,7 +227,7 @@ func (c ReadinessCheck) IsReady(p *fieldpath.Paved, o ConditionedObject) (bool, 
 	case ReadinessCheckTypeMatchCondition:
 		// we should have checked this outside of this function
 		val := o.GetCondition(xpv1.ConditionType(c.MatchCondition.Type))
-		return string(val.Status) == c.MatchCondition.Status, errors.New(errInvalidCheck)
+		return string(val.Status) == c.MatchCondition.Status, nil
 	}
 
 	return false, nil
--- a/internal/controller/apiextensions/composite/ready.go
+++ b/internal/controller/apiextensions/composite/ready.go
@@ -19,6 +19,7 @@ package composite
 import (
 	"context"
 
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/utils/pointer"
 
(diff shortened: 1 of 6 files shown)


internal/controller/apiextensions/composite/ready_test.go
```diff
@@ -65,6 +66,40 @@ func TestIsReady(t *testing.T) {
...
 			},
 		},
+		"MatchConditionReady": {
+			reason: "If no custom check is given, Ready condition should be used",
+			args: args{
+				o: composed.New(composed.WithConditions(xpv1.Available())),
+				rc: []ReadinessCheck{{
+					Type: ReadinessCheckTypeMatchCondition,
+					MatchCondition: &MatchConditionReadinessCheck{
+						// TODO(negz): These should not be plain strings.
```
phisco, line 76: I guess this can now be removed

internal/controller/apiextensions/composite/ready_test.go
```diff
@@ -65,6 +66,40 @@ func TestIsReady(t *testing.T) {
...
+			},
+		},
+		"MatchConditionNotReady": {
+			reason: "If no custom check is given, Ready condition should be used",
+			args: args{
+				o: composed.New(composed.WithConditions(xpv1.Unavailable())),
+				rc: []ReadinessCheck{{
+					Type: ReadinessCheckTypeMatchCondition,
+					MatchCondition: &MatchConditionReadinessCheck{
+						// TODO(negz): These should not be plain strings.
```
phisco, line 93: Same here

phisco (approved):
Good catch, thanks!

internal/controller/apiextensions/composite/ready.go
```diff
@@ -226,8 +224,8 @@ func (c ReadinessCheck) IsReady(p *fieldpath.Paved, o ConditionedObject) (bool,
 		return val == *c.MatchInteger, nil
 	case ReadinessCheckTypeMatchCondition:
 		// we should have checked this outside of this function
```
phisco, line 226: This comment too can actually be removed, it feels like this was a default case that got lost during a refactoring

PR #4203, opened 2023-06-19
fix(xfn): set max layers number limit for images

phisco:

Enforce maximum number of layers for images, setting it to 256 by default as a symbolic upper limit which we should never  hit except if something shady is going on.

I have:

Manually deploying crossplane with xfn enabled.

### DIFF
--- a/internal/oci/store/overlay/store_overlay.go
+++ b/internal/oci/store/overlay/store_overlay.go
@@ -173,8 +173,11 @@ func (c *CachingBundler) Bundle(ctx context.Context, i ociv1.Image, id string, o
 	if err != nil {
 		return nil, errors.Wrap(err, errGetLayers)
 	}
-
-	lowerPaths := make([]string, len(layers))
+	nLayers := len(layers)
+	if nLayers > store.MaxLayers {
+		return nil, errors.Errorf(store.ErrFmtTooManyLayers, nLayers, store.MaxLayers)
+	}
+	lowerPaths := make([]string, nLayers)
 	for i := range layers {
 		p, err := c.layer.Resolve(ctx, layers[i], layers[:i]...)
 		if err != nil {
--- a/internal/oci/store/overlay/store_overlay.go
+++ b/internal/oci/store/overlay/store_overlay.go
@@ -169,15 +169,16 @@ func (c *CachingBundler) Bundle(ctx context.Context, i ociv1.Image, id string, o
 		return nil, errors.Wrap(err, errReadConfigFile)
 	}
 
+	if err := store.Validate(i); err != nil {
+		return nil, err
+	}
+
 	layers, err := i.Layers()
 	if err != nil {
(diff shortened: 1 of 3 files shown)


internal/oci/store/overlay/store_overlay.go
```diff
@@ -173,8 +173,11 @@ func (c *CachingBundler) Bundle(ctx context.Context, i ociv1.Image, id string, o
 	if err != nil {
 		return nil, errors.Wrap(err, errGetLayers)
 	}
-
-	lowerPaths := make([]string, len(layers))
+	nLayers := len(layers)
+	if nLayers > store.MaxLayers {
+		return nil, errors.Errorf(store.ErrFmtTooManyLayers, nLayers, store.MaxLayers)
+	}
```
negz, line 179: Is it worth making this a function in the `oci` package? We could then export and call that function rather than exporting a const error string. Something like `oci.Validate(img)`.

We could also contribute this to go-container registry, which already has a `validate.Image(img)` function. I could imagine something like `validate.Image(img, validate.MaxLayers(256))` being useful for other if this is indeed a general security concern.

phisco: See https://github.com/crossplane/crossplane/pull/4203/commits/7fd862047d6aa81d1811214db55fde3966bf26ed, I couldn't move it to `oci` because I would have needed to either export or duplicate `errGetLayers`, which we could definitely do, but doesn't feel right. The main drawback I can see is that we have to call `Layers` twice, which is necessary if we keep the `Image` as input.

internal/oci/store/store.go
```diff
@@ -346,3 +356,16 @@ func copyChunks(dst io.Writer, src io.Reader, chunkSize int64) (int64, error) {
 		}
 	}
 }
+
+// Validate returns an error if the supplied image is invalid,
+// e.g. the number of layers is above the maximum allowed.
```
negz, line 361: Maybe add a TODO mentioning that this _could_ be merged with go-containerregistry's `validate.Image` in future?

internal/oci/store/store.go
```diff
@@ -78,6 +78,12 @@ const (
 	errOpenLayer        = "cannot open layer"
 	errStatLayer        = "cannot stat layer"
 	errCheckExistence   = "cannot determine whether layer exists"
+	errFmtTooManyLayers = "image has too many layers: %d (max %d)"
+)
+
+var (
+	// MaxLayers is the maximum number of layers an image can have.
+	MaxLayers = 256
```
negz, line 86: I wonder if this is also a concern for the package manager?

Haven't checked the code, but iirc it will look for a layer within an image that is specifically annotated to contain the package contents before falling back to pulling the entire image.

negz (approved):

internal/oci/store/store.go
phisco: Yes it is, but I'll address that in a separate PR

Issue #4196, opened 2023-06-19
A `Provider.pkg` upgraded to a non-existent package appears healthy


negz:
### What happened?

While working on https://github.com/crossplane/crossplane/pull/4187 I updated a healthy `Provider.pkg` with an active revision to a nonsense, non-existent package image. I noticed that while the `Provider.pkg` does emit an event to tell me it can't unpack the new package, it otherwise appears healthy - i.e. it has `Healthy: True`, `Installed: True` when I describe it:

```
$ k describe provider.pkg
Name:         provider-nop
Namespace:
Labels:       <none>
Annotations:  <none>
API Version:  pkg.crossplane.io/v1
Kind:         Provider
Metadata:
  Creation Timestamp:  2023-06-19T01:10:09Z
  Generation:          2
  Resource Version:  845
  UID:               89b7c2a5-c22b-45be-b23d-3c528adc44cb
Spec:
  Ignore Crossplane Constraints:  true
  Package:                        crossplane/provider-wat:latest
  Package Pull Policy:            IfNotPresent
  Revision Activation Policy:     Automatic
  Revision History Limit:         1
  Skip Dependency Resolution:     false
Status:
  Conditions:
    Last Transition Time:  2023-06-19T01:10:14Z
    Reason:                HealthyPackageRevision
    Status:                True
    Type:                  Healthy
    Last Transition Time:  2023-06-19T01:10:10Z
    Reason:                ActivePackageRevision
    Status:                True
    Type:                  Installed
  Current Identifier:      crossplane/provider-nop:v0.1.1
  Current Revision:        provider-nop-15e9008df94e
Events:
  Type     Reason                  Age                    From                                 Message
  ----     ------                  ----                   ----                                 -------
  Warning  InstallPackageRevision  9m47s (x5 over 9m48s)  packages/provider.pkg.crossplane.io  current package revision health is unknown
  Warning  InstallPackageRevision  9m47s (x2 over 9m47s)  packages/provider.pkg.crossplane.io  current package revision is unhealthy
  Normal   InstallPackageRevision  9m44s (x4 over 9m47s)  packages/provider.pkg.crossplane.io  Successfully installed package revision
[...]

PR #4205, opened 2023-06-19
Set provider condition on package fetch error

sttts:

Fixes https://github.com/crossplane/crossplane/issues/4196

Before this, the only way to notice a 404 when fetching a revision was the following event:

```
37s   Warning   UnpackPackage   provider/provider-gcp   cannot unpack package: failed to fetch package digest from remote: failed to fetch package descriptor with a GET request after a previous HEAD request failure: HEAD https://xpkg.upbound.io/v2/upbound/provider-gcp/manifests/latest: unexpected status code 404 Not Found (HEAD responses have no body, use GET for details): GET https://xpkg.upbound.io/v2/upbound/provider-gcp/manifests/latest: MANIFEST_UNKNOWN: manifest unknown; map[Tag:latest]
```

The conditions of the provider object were never set, making diagnoses hard, especially as this happens to first-time users of Crossplane when installing the first provider.

I have:

Added a unit test.

### DIFF
--- a/internal/controller/pkg/manager/reconciler.go
+++ b/internal/controller/pkg/manager/reconciler.go
@@ -286,7 +286,13 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 	if err != nil {
 		log.Debug(errUnpack, "error", err)
 		err = errors.Wrap(err, errUnpack)
+		p.SetConditions(v1.Unpacking().WithMessage(err.Error()))
 		r.record.Event(p, event.Warning(reasonUnpack, err))
+
+		if updateErr := r.client.Status().Update(ctx, p); updateErr != nil {
+			return reconcile.Result{}, errors.Wrap(updateErr, errUpdateStatus)
+		}
+
 		return reconcile.Result{}, err
 	}
 
--- a/internal/controller/pkg/manager/reconciler_test.go
+++ b/internal/controller/pkg/manager/reconciler_test.go
@@ -129,6 +129,41 @@ func TestReconcile(t *testing.T) {
 				err: errors.Wrap(errBoom, errListRevisions),
 			},
 		},
+		"ErrFetchRevision": {
+			reason: "We should return an error if fetching the revision for a package fails.",
+			args: args{
+				req: reconcile.Request{NamespacedName: types.NamespacedName{Name: "test"}},
+				rec: &Reconciler{
+					newPackage:             func() v1.Package { return &v1.Configuration{} },
+					newPackageRevisionList: func() v1.PackageRevisionList { return &v1.ConfigurationRevisionList{} },
+					client: resource.ClientApplicator{
+						Client: &test.MockClient{
+							MockGet:  test.NewMockGetFn(nil),
+							MockList: test.NewMockListFn(kerrors.NewNotFound(schema.GroupResource{}, "")),
+							MockStatusUpdate: test.NewMockSubResourceUpdateFn(nil, func(o client.Object) error {
+								want := &v1.Configuration{}
+								want.SetConditions(v1.Unpacking().WithMessage(errors.Wrap(errBoom, errUnpack).Error()))
+								if diff := cmp.Diff(want, o); diff != "" {
+									t.Errorf("-want, +got:\n%s", diff)
+								}
+								return nil
+							}),
+						},
+						Applicator: resource.ApplyFn(func(_ context.Context, _ client.Object, _ ...resource.ApplyOption) error {
+							return nil
+						}),
+					},
+					log:    testLog,
+					record: event.NewNopRecorder(),
+					pkg: &MockRevisioner{
+						MockRevision: NewMockRevisionFn("", errBoom),
+					},
+				},
+			},
+			want: want{
+				err: errors.Wrap(errBoom, errUnpack),
+			},
+		},
 		"SuccessfulNoExistingRevisionsAutoActivate": {
 			reason: "We should be active and not requeue on successful creation of the first revision with auto activation.",
 			args: args{


internal/controller/pkg/manager/reconciler.go
```diff
@@ -284,10 +284,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 
 	revisionName, err := r.pkg.Revision(ctx, p)
 	if err != nil {
-		log.Debug(errUnpack, "error", err)
 		err = errors.Wrap(err, errUnpack)
+		p.SetConditions(v1.Unpacking())
 		r.record.Event(p, event.Warning(reasonUnpack, err))
-		return reconcile.Result{}, err
+		return reconcile.Result{RequeueAfter: time.Minute}, errors.Wrap(r.client.Status().Update(ctx, p), errUpdateStatus)
```
hasheddan, line 290: We previously opted to return the error to enqueue immediately and rely on exponential backoff here. I'm open to adjusting, but 1 minute feels like it could be a fairly long time to wait just to attempt fetching the digest again. Do you think updating to always enqueue after 1 minute makes more sense?

hasheddan (commented):
Thanks @sttts! Left one question below.

internal/controller/pkg/manager/reconciler.go
hasheddan: @sttts it depends -- I believe we have used it in places where the issue is likely to be transient or requiring some external intervention (in this case we likely either have a temporary network issue, or the digest does not exist at all). In those cases we want to retry quickly, then retry infrequently, as opposed to retrying on a fixed interval (e.g. 1 minute). Worth noting that we globally configure backoff today, so there is likely some room for improvement.

sttts: > Do you think updating to always enqueue after 1 minute makes more sense?

Not exactly sure what you mean here. Some agressive backoff with 1min max would IMO make most sense. Wdyt?

sttts: Check again. That's what you mean?

internal/controller/pkg/manager/reconciler.go
```diff
@@ -284,9 +284,19 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 
 	revisionName, err := r.pkg.Revision(ctx, p)
 	if err != nil {
-		log.Debug(errUnpack, "error", err)
 		err = errors.Wrap(err, errUnpack)
+		p.SetConditions(v1.Unpacking())
```
sttts, line 288: actually this is wrong. It always sets the lastTransitionTime.

sttts: nvmd. The setter applies the right equality.

internal/controller/pkg/manager/reconciler.go
sttts: Here is the logic in controller-runtime (fun-fact: one must go into the implementation to find the semantics): https://github.com/kubernetes-sigs/controller-runtime/blob/451530c09937350ff7f5ebf7a00d059b4334e8ee/pkg/internal/controller/controller.go#L325

sttts: > I think just returning the error is fine for now here.

:+1:

hasheddan: > Here is the logic in controller-runtime (fun-fact: one must go into the implementation to find the semantics):

Also linked in the post :)

internal/controller/pkg/manager/reconciler.go
```diff
@@ -284,9 +284,14 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
```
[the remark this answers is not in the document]
hasheddan: Which message are you referring to @sttts? In the condition? We use a standard set of conditions for each API type. I think it is worth considering whether this case should be exposed as some sort of error as @negz described in https://github.com/crossplane/crossplane/issues/4196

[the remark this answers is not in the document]
sttts: Yes, the condition message aka `p.SetCOnditions(v1.Unpacking().WithMessage(err.String())`.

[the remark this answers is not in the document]
hasheddan: We don't attach messages to events anywhere to my knowledge -- we opt to use events to communicate more detail.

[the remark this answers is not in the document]
sttts: So maybe we have to check first, and keep the condition true, but with some other reason "InstalledButUpgradeFailed".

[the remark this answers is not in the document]
hasheddan: > I think it is worth considering whether this case should be exposed as some sort of error as @negz described in https://github.com/crossplane/crossplane/issues/4196

@sttts I think my comment here was actually already addressed -- I didn't immediately realize that your addition of `v1.Unpacking()` was setting `Installed: false`, which I think is plenty to address the concerns in #4196 (at least for the time-being). Apologies for the extra back and forth here!

internal/controller/pkg/manager/reconciler.go
```diff
@@ -284,9 +284,14 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 
 	revisionName, err := r.pkg.Revision(ctx, p)
 	if err != nil {
-		log.Debug(errUnpack, "error", err)
```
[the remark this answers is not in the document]
sttts: done

internal/controller/pkg/manager/reconciler.go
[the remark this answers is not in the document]
sttts: done

negz (approved):

hasheddan (approved):

ezgidemirel (approved):

Issue #3393, opened 2022-10-17
Introduce a generic `Usage` type to enforce ordered deletion


negz:
### What problem are you facing?

Crossplane is built on Kubernetes, which leans into eventual consistency. When several things need to happen in a particular order the “Kubernetes way” is to specify the desired end state and rely on the relevant controllers to ensure the system eventually arrives at that state, largely through the use of constant reconciliation (aka retries).

Take for example the creation of a VPC Network and a Subnet. The VPC must be created and ready before a Subnet can be created within it. In Kubernetes it’s typical to request both resources be created at once (thereby declaring the desired state). The creation of the Subnet will simply fail until its dependency - the VPC - becomes available.

This loosely coupled, eventually consistent approach is simple and resilient but it can appear chaotic. Operations must be attempted more times than would be necessary in a “smarter” system, and often the temporarily failing operations will emit warnings that may be confusing or concerning to the uninitiated. Some other systems - notably Terraform - avoid this by computing a dependency graph before attempting to enact the desired state. This allows the system to avoid trying operations it knows will fail due to missing dependencies until those dependencies are expected to be satisfied.

A frequent negative side-effect of the eventually consistent approach that we see in Crossplane is resources becoming orphaned at delete time because their successful deletion depends on the existence of other resources which are often deleted before their dependents. It’s possible to author a set of interdependent resources that will eventually create successfully, but may not eventually delete successfully.

Two variants of this problem are:

* **Orphaned Managed Resources (MRs)**. Consider a Helm Release MR that is deployed to an EKS Cluster MR. The Helm Release has a dependency on the EKS Cluster’s connection details - it uses them as Provider credentials. If the EKS Cluster is deleted before the Helm Release the Helm Release controller becomes unable to connect to its Provider in order to delete the Release. In practice the Release External Resource (ER) was implicitly deleted along with the GKE Cluster - there’s actually no work left for the Helm release controller to do - but it cannot know that. This results in the Helm Release MR being ‘orphaned’ in a pending deletion that can never be satisfied.
[...]

Issue #2439, opened 2021-07-18
Way to block managed resource from being deleted till its dependency deleted


morningspace:
### What problem are you facing?

I have two managed resources where one depends on the other. When delete the two resources, one will fail if the other is deleted too early. Right now, it looks the general reconciler handles managed resources deletion independently, which is good. The thing is that wether it could allow provider owner to customize this behavior somehow using `finalizer`.

Couple of observations after investigation for a while:
- Some provider implementation are kind of taking over the reconciliation logic by themselves, so that they fully manage the finalizer adding/removing by themselves. e.g.: the [storage/container controller](https://github.com/crossplane/provider-azure/blob/master/pkg/controller/storage/container/container.go) in provider-azure. But this will lose all benefits provided by the general managed resource reconciler, e.g. its Connect/Observer/Create/Update/Delete lifecycle.
- Inside the provider, It looks I can handle the finalizer of the managed resource passed in, but it looks the current implementation for the general reconciler assumes there is only one finalizer, `finalizer.managedresource.crossplane.io`, which is added by itself. This may lead to the confusion even multi-finalizer approach is technically feasible. e.g.: after the reconciler remove its finalizer from the managed resource, it will report "Successfully deleted managed resource".

### How could Crossplane help solve your problem?

It looks the scenario itself is something common, but maybe I'm too opinionated:-). It would be great if there's a nice way to handle it in crossplane core or runtime, but if that's something not too generic, maybe it makes sense to keep that inside the particular provider at least for now, in which case, I think provider owners may need some guidance on this from the community owner.

morningspace:
Seeing a similar issue reported here: https://github.com/crossplane-contrib/provider-sql/issues/47, so I see more people having this similar issue. It could be fixed somehow at provider level, but given more providers potentially having this problem, it would be better to address it properly at crossplane core & runtime level.

negz:
> I have two managed resources where one depends on the other.

Can you give an example of the nature of the dependency?

> Some provider implementation are kind of taking over the reconciliation logic by themselves, so that they fully manage the finalizer adding/removing by themselves. e.g.: the storage/container controller in provider-azure.

This is an anti-pattern that will be removed. The provider-azure storage controllers in fact predate the `managed.Reconciler`, so the issue is more that they have not yet been updated rather than that they're doing something special on purpose.

> It could be fixed somehow at provider level, but given more providers potentially having this problem, it would be better to address it properly at crossplane core & runtime level.

[...]

PR #4215, opened 2023-06-20
Add One-Pager for Deletion Ordering

turkenh:

This PR adds a one-pager design document for the deletion ordering problem.

Related: 
- https://github.com/crossplane/crossplane/issues/3393
- https://github.com/crossplane/crossplane/issues/2439

I have:

N/A - but a POC is available at https://github.com/turkenh/crossplane/pull/2


design/one-pager-generic-usage-type.md
```diff
@@ -0,0 +1,397 @@
+# Generic `Usage` Type for Deletion Ordering
+
+* Owner: Hasan Türken (@turkenh)
+* Reviewers: @bobh66, @negz
+* Status: Draft
+
+## Background
```
turkenh, line 7: The background section is courtesy of Nic, I borrowed it from his great write-up in the description of the [proposal issue](https://github.com/crossplane/crossplane/issues/3393). 
Thank you @negz!

bobh66 (approved):
Overall I think this is excellent.  A couple of questions/comments inline but nothing blocking.  Thanks for the hard work!

design/one-pager-generic-usage-type.md
```diff
@@ -0,0 +1,397 @@
...
+apiVersion: apiextensions.crossplane.io/v1alpha1
+kind: Usage
+metadata:
+  name: release-uses-cluster
+spec:
+  of:
+  - apiVersion: eks.upbound.io/v1beta1
+    kind: Cluster
+    selector:
+      matchControllerRef: true
```
negz, line 127: This is asymmetrical with our current reference/selector implementations, right? It looks like a selector _inside_ a reference, rather than a selector _at the same level_ as a reference.

Would we ever have a use for different kinds of selectors - e.g. matchLabels? I think yes - e.g. if there are two `Cluster` in a Composition and we only want to select one (without knowing its name in advance) matching controller ref alone is insufficient.

negz: It might be useful to include an example of a Usage (or two) inside a Composition.

design/one-pager-generic-usage-type.md
```diff
@@ -0,0 +1,397 @@
...
+(i.e. resources that declare the parent as an owner reference) will later be
+deleted by the Kubernetes garbage collector. This makes it difficult to diagnose
+or even notice resources that become orphaned when their parent XR is deleted.
+
+## Goals
+
+- Support "cross-provider" dependencies, for example an MR from provider-helm
+depending on a ProviderConfig from provider-aws.
+- Don’t require Crossplane users to manage fine-grained RBAC access.
+- Don’t grant Crossplane providers broad RBAC access (e.g. to all other providers).
```
negz, line 72: Adding brief note below explaining how the proposal achieves this would probably be useful. i.e. The RBAC manager already grants Crossplane RBAC access to all types in order for it to do Composition.

design/one-pager-generic-usage-type.md
```diff
@@ -0,0 +1,397 @@
...
+
+Implementation of the core logic for the webhook handler:
+
+```go
+func (h *handler) validateNoUsages(ctx context.Context, u *unstructured.Unstructured) admission.Response {
+	usageList := &v1alpha1.UsageList{}
+	if err := h.reader.List(ctx, usageList, client.MatchingFields{inUseIndexKey: getIndexValueForObject(u)}); err != nil {
+		return admission.Errored(http.StatusInternalServerError, err)
+	}
+	if len(usageList.Items) > 0 {
```
negz, line 220: Not sure it's fair to nitpick code in a design, but this looks like a good place to [return early](https://github.com/crossplane/crossplane/tree/master/contributing#return-early). 😉

design/one-pager-generic-usage-type.md
```diff
@@ -0,0 +1,397 @@
...
+The admission webhook will intercept the `DELETE` requests for the used
+resource(s). To achieve this, the controller will add [webhook rules] for the
+used resource(s) when a `Usage` is created. It is also the controller's job to
+remove these rules when the `Usage` is deleted, and no other `Usage` resources
+reference the same used resource(s).
+
+#### Admission Webhook
+
+The admission webhook will be responsible for intercepting the `DELETE` requests
+for the used resource(s) and rejecting them with a ["409 Conflict" error code].
```
negz, line 203: Would we fail open or closed if the webhook failed to process the request in time?

https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#failure-policy

design/one-pager-generic-usage-type.md
```diff
@@ -0,0 +1,397 @@
...
+			},
+		}
+	}
+	return admission.Allowed("")
+}
+```
+
+### User Experience
+
+**Defining a Usage Resource as part of a Composition:**
```
negz, line 237: Nit: `####` style headings would make this section a little easier to read as markdown (which is typically how I end up reading these when reviewing).

design/one-pager-generic-usage-type.md
```diff
@@ -0,0 +1,397 @@
...
+simpler solution, namely the webhook-based solution_. We think it's better to
+start with the simple option, and then, if we find out that it's not enough, we
+can think about changing to a more complex solution. If we check the above table,
+it is clear that the webhook-based solution is the simpler one.
+
+If we decide to change the underlying implementation later, we don't expect
+an API change except getting rejected deletes vs. not. This also provides us a
+good opportunity to test the webhook-based solution in the field and gather
+feedback from the community before committing to a more complex solution.
+
```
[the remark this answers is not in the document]
negz: Agreed. I could see `Usage` being used for two things:

1. Guarantee order of deletion.
2. Prevent something important (e.g. a `VPC`) being deleted until someone manually removes a `Usage`.

Per https://github.com/kubernetes/enhancements/pull/2840#discussion_r1224844740 I feel like immediately rejecting a delete makes a lot more sense in the second case.

design/one-pager-generic-usage-type.md
```diff
@@ -0,0 +1,397 @@
...
+  name: release-uses-cluster
+spec:
+  of:
+  - apiVersion: eks.upbound.io/v1beta1
+    kind: Cluster
+    name: my-cluster
+  by:
+    apiVersion: helm.crossplane.io/v1beta1
+    kind: Release
+    name: my-prometheus-chart
```
negz, line 102: Should `by` be optional?

I'm thinking folks might want to manually create a `Usage` as a form of deletion protection. For example to add a layer of protection that a `VPC` not be accidentally deleted.

design/one-pager-generic-usage-type.md
```diff
@@ -0,0 +1,397 @@
...
+## Proposal
+
+We propose introducing a new `Usage` type that can be used to declare usage
+relationships between Crossplane resources. This type will be defined by the
+Core Crossplane repository, and be available for use as part of a composition.
+
+The relations defined by this type will be enforced by an admission webhook that
+will be running as part of the Crossplane core. Similar to the upstreams [liens
+proposal], the webhook will reject deletions of resources that are in use by
+other resources with a ["409 Conflict" error code].
```
sttts, line 83: We briefly talked about that in Istanbul: this is a misuse of the webhook API. A 409 is not considered as a non-admission, but as a failure of the webhook itself. The only valid response is through: https://github.com/kubernetes/kubernetes/blob/9e0569f2ed3934060fabe51be4e15232bbea3877/pkg/apis/admission/types.go#L124.

In systems like OpenShift and maybe also other Kube distributions and/or cloud providers, there will be monitoring alerts in place that will consider the 409 as a failure and warn the user, and at worst even disable the webhook or backoff in order to keep the cluster stable.

design/one-pager-generic-usage-type.md
sttts: We probably have to fail closed because otherwise the 409 is ignored.

sttts: xref https://github.com/crossplane/crossplane/pull/4215/files#r1236446471

design/one-pager-generic-usage-type.md
sttts: Nvmd. The code snipped below shows what you actually do: you set 409 in the admission response. Let's clarify that here:
```suggestion
other resources with a ["409 Conflict" error code] as `metav1.Status` as part of the admission response (not the actual http code of the admission request).
```

design/one-pager-generic-usage-type.md
```diff
@@ -0,0 +1,397 @@
...
+resource hence `Usage` is deleted.
+
+> One caveat with the garbage collector is the substantial delay it can cause
+during the deletion of composites. This delay caused by the garbage collector's
+use of an exponential backoff when the DELETE API call is rejected. The maximum
+backoff time is 1000 seconds, which means that after no usage of an object is
+left, it could take up to 1000 seconds before the garbage collector attempts to
+delete the resource that was previously blocked. To mitigate this issue, we
+could introduce a mechanism in the controller to re-trigger the deletion of the
+used resource as soon as the Usage resource is deleted.
```
sttts, line 290: The downside of this is that we have to reimplement basically the GC logic of "when not to delete an object". I.e. there is risk that Kube introduces something that stops the GC from deleting an object (think of another ownership mechanism, or even something like this `Usage` idea here in the proposal). Our code won't be aware and too quickly delete an object bypassing GC.

My gut feeling is that this risk is limited as these things don't show up out of nowhere, but api-machinery is slow moving and if it ever happens, we have to revisit our code.

design/one-pager-generic-usage-type.md
```diff
@@ -0,0 +1,397 @@
...
+We will introduce a new `Usage` type as follows:
+
+```yaml
+apiVersion: apiextensions.crossplane.io/v1alpha1
+kind: Usage
+metadata:
+  name: release-uses-cluster
+spec:
+  of:
+  - apiVersion: eks.upbound.io/v1beta1
```
sttts, line 96: Clarify that such a reference applies to every group version of that kind. E.g. a usage of `eks.upbound.io/v1beta1` clusters implies a usage of `eks.upbound.io/v1` clusters.

sttts: Another details, not sure I saw that specified anywhere: what do dangling references mean where the GVK is unknown to discovery (rest mapper)?

design/one-pager-generic-usage-type.md
bobh66: Wouldn't the DELETE request that we make be subject to the same processing/restrictions as any other DELETE request? I don't see that we are bypassing GC, we are just asking it to try again, and it may fail for other reasons, which will be outside the scope of our scenario, and we will never try again.

design/one-pager-generic-usage-type.md
```diff
@@ -0,0 +1,397 @@
...
+- Support "cross-provider" dependencies, for example an MR from provider-helm
+depending on a ProviderConfig from provider-aws.
+- Don’t require Crossplane users to manage fine-grained RBAC access.
+- Don’t grant Crossplane providers broad RBAC access (e.g. to all other providers).
+
+## Proposal
+
+We propose introducing a new `Usage` type that can be used to declare usage
+relationships between Crossplane resources. This type will be defined by the
+Core Crossplane repository, and be available for use as part of a composition.
```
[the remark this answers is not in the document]
turkenh: No. It is still a valid use case and is mentioned under user experience. 
Slightly updated the wording to indicate that.

design/one-pager-generic-usage-type.md
```diff
@@ -0,0 +1,397 @@
...
+
+#### Usage Controller
+
+The usage controller will be responsible for managing the lifecycle of the usage
+relationship by:
+
+- Resolving selectors, if any.
+- Adding owner reference from the using resource (i.e., `spec.by`) to the `Usage`.
+- Preventing deletion of `Usage` before the using resource is deleted.
+- Adding or removing admission webhook rules to intercept the `DELETE` requests
```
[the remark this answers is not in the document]
turkenh: Rules do not include the name of the resource so we would need one rule per GVR, so I don't think we would need to scale up to thousands of rules. If every installed type has at least one dependency (which is not likely), the number of rules would equal to the number of CRDs + the number of XR types. I don't think this would be a problem in practice.

However, using a selector sounds like a good idea if we observe this as a problem. I am wondering if we could instead add / remove a well defined label on the used resource by the controller and have a fixed webhook rule for all objects but matching that label. This way, the controller wouldn't need to interact with the webhook rules but would need to add/remove labels on the used resource. This could even be helpful to reset a client's backoff for deletion, e.g. garbage collector.

Something like:

```
webhooks:
- name: nousages.apiextensions.crossplane.io
  objectSelector:
    matchLabels:
      crossplane.io/in-use: true
  rules:
  - operations: ["DELETE"]
    apiGroups: ["*"]
    apiVersions: ["*"]
    resources: ["*"]
    scope: "*"
```

design/one-pager-generic-usage-type.md
```diff
@@ -0,0 +1,397 @@
...
+reference while defining a usage relationship does not make sense. Also, the
+[reasoning behind] supporting different policies for MRs does not apply to
+the `Usage`. However, depending on the feedback, we may consider supporting
+resolve policies to configure whether the resolution should be made once or
+continuously in a future iteration.
+
+**Adding owner reference:**
+
+We want the `Usage` to be deleted when the using resource is deleted. This will
+be achieved by adding an owner reference from the using resource to the `Usage`.
```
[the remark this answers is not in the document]
turkenh: Thanks, I will keep this in mind during implementation.

design/one-pager-generic-usage-type.md
turkenh: I believe we are already returning early; the return statement for the error case looks bigger than the success below.
Or, am I missing sth?

design/one-pager-generic-usage-type.md
```diff
@@ -0,0 +1,397 @@
...
+	if err := h.reader.List(ctx, usageList, client.MatchingFields{inUseIndexKey: getIndexValueForObject(u)}); err != nil {
+		return admission.Errored(http.StatusInternalServerError, err)
+	}
+	if len(usageList.Items) > 0 {
+		return admission.Response{
+			AdmissionResponse: admissionv1.AdmissionResponse{
+				Allowed: false,
+				Result: &metav1.Status{
+					Code:   int32(http.StatusConflict),
+					Reason: metav1.StatusReason(fmt.Sprintf("The resource is used by %s/%s", usageList.Items[0].Spec.By.Kind, usageList.Items[0].Spec.By.Name)),
```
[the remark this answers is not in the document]
turkenh: Good call, updated as

```go
Reason: metav1.StatusReason(fmt.Sprintf("The resource is used by %d resources, including %s/%s", len(usageList.Items), usageList.Items[0].Spec.By.Kind, usageList.Items[0].Spec.By.Name)),
```

design/one-pager-generic-usage-type.md
turkenh: > This is asymmetrical with our current reference/selector implementations, right? It looks like a selector inside a reference, rather than a selector at the same level as a reference.

Yes, the main difference is that for MRs, name/id refers to the external resource we don't have. For example:

```
apiVersion: ec2.aws.upbound.io/v1beta1
kind: Subnet
...
spec:
  forProvider:
    vpcId: <external-resource-id>
    vpcIdRef:
      name: <name-of-the-k8s-resource>
    vpcIdSelector:
      matchControllerRef: true
```

Here, the external name does not apply; hence we there is no reason to put the name under ref.

```
apiVersion: helm.crossplane.io/v1beta1
kind: Release
name: <name-of-the-k8s-resource
selector:
  matchControllerRef: true
  matchLabels:
    foo: bar
```

> Would we ever have a use for different kinds of selectors - e.g. matchLabels?

Yes. This would be part of the API while not included in the example. Updated.

design/one-pager-generic-usage-type.md
turkenh: Agree with @sttts. 

I don't see a reason to fail open here, so we will set `failurePolicy: Ignore`.

design/one-pager-generic-usage-type.md
turkenh: Thanks, updated accordingly.

design/one-pager-generic-usage-type.md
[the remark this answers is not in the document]
turkenh: TBH, I was under the impression that we could just reset the backoff interval of the GC by making some modifications on the resource, e.g. adding an annotation. This is how one could retrigger a typical k8s controller watching a resource bypassing the backoff time so that the reconciler logic could rerun. I thought this would work the same with the Garbage Collector, I need to check that 🤞 or need to see if there are other ways to do so.
If not, I would be hesitant to store and trigger a deletion when the usage disappears, which could cause headaches due to different edge cases and result in a not-so-pleasant user experience for other clients. For example, a user gets rejected by trying to delete a used resource and thinks deletion **didn't happen**, but we do that behind the scenes.

sttts: My fear is that we are deleting more eagerly than Kube's GC because we miss something in our logic.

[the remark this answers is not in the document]
turkenh: Thank you @sttts, that trick was something I also found accidentally :) Obviously, adding an invalid owner ref is hacky but that sparked another idea.

What we want to achieve here is ideally the Garbage Collector not to attempt deleting a resource that has at least one Usage and only to do that when there is no Usage. One obvious way to achieve this is adding an owner reference from `Usage` to the used resource **as long as the Usage exists** (This owner reference will be cleaned up during the finalization of the `Usage` resource). In the case of composed resources, this _additional_ owner reference will prevent the resource to be Garbage collected, and once the Usage is gone, the resource will become GCable. This would mean no backoff during Garbage Collection.

We are already planning to add owner references from Using resource to the `Usage`, and with the above, it would look like:

image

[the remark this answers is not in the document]
bobh66: `blockOwnerDeletion` will be `true` on the ownerReference from the `Usage` to the `Cluster`?

Does this also block the deletion of the `Cluster`'s other child objects?

design/one-pager-generic-usage-type.md
negz: I'm a little wary of this API shape, which is kind of a-selector-inside-a-reference. Is it possible to stick to the existing pattern of selectors living next to references? My main concern is it seems to go against Kubernetes API conventions. To be fair our existing reference and selector pattern is based on old conventions (new ones are [here](https://github.com/kubernetes/community/blob/9885966e5015c3675b63116c268d7da0708b137d/contributors/devel/sig-architecture/api-conventions.md#object-references-examples)), but I'd prefer to be consistent with what we do elsewhere if possible.

I think we had a similar issue with EnvironmentConfigs and landed on this API (which to be fair I don't love either): https://github.com/crossplane/crossplane/pull/3008/files#diff-1894e5714a8f815a975c89cc55edb79c1c216dd5c1408c82ab84522d3acc3e83R94

design/one-pager-generic-usage-type.md
negz: > Clarify that such a reference applies to every group version of that kind.

FWIW this is pretty much a given for all of our references. We can be explicit here, but it's implicit in 99% of our other designs and APIs. 🙂 I believe API conventions now recommend a versionless pattern for such references, but most of our APIs use this older convention.

design/one-pager-generic-usage-type.md
negz: Yeah I could be convinced the use case I describe should be out of scope.

If we did keep it in scope, maybe a `type` field would help:

```yaml
apiVersion: crossplane.io/v1
kind: Usage
spec:
  type: Protection  # Protection vs Dependency? Human vs Resource?
  of:
  - apiVersion: ec2.upbound.io/v1beta1
    kind: VPC
    name: very-critical-vpc
```

This could also be used for CEL validation - i.e. make `by` required if the `type` indicates it should be.

design/one-pager-generic-usage-type.md
negz: You have a point. I was thinking we could reduce nesting by doing:

```go
if len(usageList.Items) == 0 {
	return admission.Allowed("")
}

return admission.Response{}  // etc..
```

We can close this thread though and worry about my code nitpicks at implementation time. 😉

negz:
Just noting that I'm 99% ready to approve this. Only thing I'd like to wrap up before I do is the shape of the reference / selectors.

design/one-pager-generic-usage-type.md
```diff
@@ -0,0 +1,410 @@
...
+#### Defining a Usage Resource as part of a Composition
+
+We expect a typical usage relationship to be defined as part of the same
+composition as the using and used resources. In this case, the user will
+define the `Usage` resource as part of the composition template.
+
+Let's take [upbound/platform-ref-aws] as an example. In this composition, we
+have a [composition for `XCluster`] which composes `XEKS` composing an EKS cluster,
+and `XServices` composing a set of helm releases running on the cluster. In this
+case, we will add the following resource to the same composition to define a
```
jbw976, line 258: disclaimer: i haven't thought through this problem space really at all, so I'm not saying this is possible at all 😊 

i was naively hoping that these usage relationships could get automatically determined by the crossplane machinery - i.e. there would be no additional burden on the users to figure out, declare, and manage these `Usage` objects themselves.

Did we already write down why that simplification of the user experience is not possible? would be curious to read more about that 😇

bobh66: Crossplane has a very narrow view of the dependency, limited to Crossplane-native and kubernetes-native objects like `ProviderConfigs` and `Secrets`.  Dependencies on Composites are not really visible to Crossplane.  For example Crossplane "knows" that a `Release` object has a `ProviderConfig` which depends on a `Secret` that contains the cluster credentials, but it has no idea that the helm chart modeled by the `Release` also has implicit dependencies on the Cluster, Nodegroup and all of the other resources that make up the "Cluster Composite".  The logical dependency is actually on everything that comprises a running cluster, not just on the credentials in the Secret.  The author of the composition is really the only one who "knows" what the actual dependency is.

That being said, it would definitely be helpful to the composition author to have a shortcut notation for the `Usage` that does not involve defining the entire resource as an entry in the resource list.  Something along the lines of a terraform `dependsOn` attribute that could cause Crossplane to automatically generate the `Usage` object, but I understand the reluctance to extend the DSL to include more terraform-like syntax.  Another possibility is to use a Composition Function to generate the `Usage` objects based on one or more annotations defined in the resource or passed in using the Composition Functions `config` object.

I think this could be part of a library of common Composition Functions that provide specific functionality on top of the standard DSL.

design/one-pager-generic-usage-type.md
turkenh: Ok, if I attempt to apply that pattern here, we would have one of the following:

```yaml
kind: Usage
metadata:
  name: release-uses-clusters
spec:
  of:
    - type: Reference
      reference:
        apiVersion: eks.upbound.io/v1beta1
        kind: Cluster
        name: my-cluster
    - type: Selector
      selector:
        apiVersion: eks.upbound.io/v1beta1
        kind: Cluster
        matchControllerRef: true
        matchLabels:
          foo: bar
  by:
    type: Reference
    reference:
      apiVersion: helm.crossplane.io/v1beta1
      kind: Release
      name: my-prometheus-chart
```

or

```yaml
kind: Usage
metadata:
  name: release-uses-clusters
spec:
  of:
    - apiVersion: eks.upbound.io/v1beta1
      kind: Cluster
      type: Reference # can't we just drop this type?
      reference:
        name: my-cluster
    - apiVersion: eks.upbound.io/v1beta1
      kind: Cluster
      type: Selector # can't we just drop this type?
      selector:
        matchControllerRef: true
        matchLabels:
          foo: bar
  by:
    apiVersion: helm.crossplane.io/v1beta1
    kind: Release
    type: Reference
    reference:
      name: my-prometheus-chart
```

TBH, I am not happy with either but closer to the latter.

This is slightly different from then EnvironmentConfig case, where the kind is known not arbitrary. Is this the first time where we are defining a multi-kind reference/selector type relation? And I believe we will need the same for generic resource references.

negz: > Is this the first time where we are defining a multi-kind reference/selector type relation?

I think so, yes.

```yaml
  of:
    - type: Reference
      reference:
        apiVersion: eks.upbound.io/v1beta1
        kind: Cluster
        name: my-cluster
    - type: Selector
      selector:
        apiVersion: eks.upbound.io/v1beta1
        kind: Cluster
        matchControllerRef: true
        matchLabels:
          foo: bar
```

This is my preference of the three. All of the APIs feel a little awkward to me though, so I leave it up to you to pick which you think is best. 😉

negz (approved):
Approving to make sure this doesn't block while I'm on vacation. This looks great - thanks @turkenh!

design/one-pager-generic-usage-type.md
turkenh: Currently, I am closer to the second and `type` dropped but love to hear more thoughts:

```
  of:
    - apiVersion: eks.upbound.io/v1beta1
      kind: Cluster
      reference:
        name: my-cluster
    - apiVersion: eks.upbound.io/v1beta1
      kind: Cluster
      selector:
        matchControllerRef: true
        matchLabels:
          foo: bar
```

When the selector is configured, the resolution will be done and the controller change it from:

```
    - apiVersion: eks.upbound.io/v1beta1
      kind: Cluster
      selector:
        matchControllerRef: true
        matchLabels:
          foo: bar
```

to 

```
    - apiVersion: eks.upbound.io/v1beta1
      kind: Cluster
      reference:
        name: that-cluster
      selector:
        matchControllerRef: true
        matchLabels:
          foo: bar
```

By following the rules documented in the design:

> - Resolution will be made once when the `Usage` is created.
> - If multiple resources match the selector, a random one will be
>   selected.
> 
> The [existing policies] for resolving Managed Resource references does not
> necessarily apply to the `Usage`. For example, supporting an `optional`
> reference while defining a usage relationship does not make sense. Also, the
> [reasoning behind] supporting different policies for MRs does not apply to
> the `Usage`. However, depending on the feedback, we may consider supporting
> resolve policies to configure whether the resolution should be made once or
> continuously in a future iteration.

design/one-pager-generic-usage-type.md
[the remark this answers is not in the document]
turkenh: > `blockOwnerDeletion` will be `true` on the ownerReference from the `Usage` to the `Cluster`?

I don't see an immediate reason to set `blockOwnerDeletion` to `true` there.

> Does this also block the deletion of the Cluster's other child objects when using foreground cascading deletion?

I think so. Whether it is background or foreground, I would expect GC process not to start as long as there is at least one existing owner. But I will keep this in mind during implementation and make sure that it works as expected.

design/one-pager-generic-usage-type.md
turkenh: We discussed this offline with @negz and aligned on the following API.

```yaml
kind: Usage
metadata:
  name: release-uses-clusters
spec:
  of:
    apiVersion: eks.upbound.io/v1beta1
    kind: Cluster
    resourceSelector:
      matchControllerRef: true
      matchLabels:
         foo: bar
  by:
    apiVersion: helm.crossplane.io/v1beta1
    kind: Release
    resourceRef:
      name: my-prometheus-chart
```

Given there is no solid reason to have `spec.of` as an array and multiple `Usage` objects could provide the same functionality, we wanted to drop it being an array due to potential confusion when a selector is provided, i.e. when a selector is provided, would all matching resources be added to the array?

The controller would resolve the references based on selectors, if any, and persist them into `resourceRef.name`, similar to how MR selectors are resolved today. So, the above would turn to the following after the first reconcile:

```yaml
kind: Usage
metadata:
  name: release-uses-clusters
spec:
  of:
    apiVersion: eks.upbound.io/v1beta1
    kind: Cluster
    resourceRef: # added by the controller based on the selectors
      name: my-cluster # added by the controller based on the selectors
    resourceSelector:
      matchControllerRef: true
      matchLabels:
         foo: bar
  by:
    apiVersion: helm.crossplane.io/v1beta1
    kind: Release
    resourceRef:
      name: my-prometheus-chart
```

design/one-pager-generic-usage-type.md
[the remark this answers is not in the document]
turkenh: I've updated the design with the label/selector approach above.

design/one-pager-generic-usage-type.md
turkenh: Updated the doc accordingly.

design/one-pager-generic-usage-type.md
turkenh: Introducing `type` would help to figure out whether `by` should be required or not makes sense, however, does not still help with "enforcing documenting the source of the `Usage`". I am wondering if we can have a string `reason` field and require either `spec.by` or `spec.reason` with CEL rules? TBH, `spec.type` would be redundant in that case even I believe it is a convention but not sure applies here.

Protection:

```
apiVersion: crossplane.io/v1
kind: Usage
spec:
  reason: "Production Database - should never be deleted"
  of:
    apiVersion: rds.aws.upbound.io/v1beta1
    kind: Instance
    resourceRef:
      name: my-cluster
```

Dependency:

```
kind: Usage
metadata:
  name: release-uses-cluster
spec:
  of:
    apiVersion: eks.upbound.io/v1beta1
    kind: Cluster
    resourceRef:
      name: my-cluster
  by:
    apiVersion: helm.crossplane.io/v1beta1
    kind: Release
    resourceRef:
      name: my-prometheus-chart
```

design/one-pager-generic-usage-type.md
turkenh: I believe we are aligned on this in [the other thread](https://github.com/crossplane/crossplane/pull/4215#discussion_r1239579950).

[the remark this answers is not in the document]
turkenh: > One obvious way to achieve this is adding an owner reference from Usage to the used resource as long as the Usage exists (This owner reference will be cleaned up during the finalization of the Usage resource)

One caveat I noticed during implementation is that it could cause unexpected deletion of `used` resource if someone _manually_ removes the finalizer on the `Usage` resource since owner ref from `Usage` to `used` resource is not cleaned up. I would be ok to this by saying "one should never manually remove a finalizer; otherwise unexpected things could happen", however, there is at least one use case , which I believe, we would need manual removal of finalizer on the `Usage`, which is deleting the `Usage` while `using` resource is still there.

Imagine I created a `Release`, `Cluster`, and a `Usage` as `release-using-cluster` as in the above figure. Since both the `Release` and `Usage` could be part of the same composition, we propose using finalizers to block the deletion of `Usage` before `Release` disappears. So, a `Usage` would never be deleted unless `Release` is gone. However, there could be some cases where someone changed their mind and want to just clean up the `Usage` just to break the dependency. Removing the finalizer would be a straightforward solution in this case unless we implement additional knobs like an annotation to force delete etc.

So, I am on the fence regarding implementing this workaround to speed up garbage collector as it could be fatal (used resource could be deleted) if someone messes up with finalizers.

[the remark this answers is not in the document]
bobh66: > (used resource could be deleted) if someone messes up with finalizers.

I think (or I hope) most people understand that removing finalizers is at-your-own-risk.  I like using finalizers better than re-triggering the delete call, and the controller can maintain the finalizer until the `Release` (in this case) has been deleted.  I think it's worth a try if the only alternative is waiting for GC to come around again in 17m.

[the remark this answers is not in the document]
turkenh: Testing this with Compositions and unfortunately observed a behaviour that invalidates it as a solution :disappointed:

**TL;DR:** If the Garbage collector finds an owner to a nonexisting resource, it removes that owner from the owner list if it is not the only owner (which would otherwise delete the resource). Due to this, we are losing owner ref to composite before we clean up the one for Usage, hence we end up with a non GC-ed resource with no owner ref.

Here is what happened:

1. Created a composite composing `used` and `using` resources together with the `Usage` itself.
2. Composite controller created `used` resource with owner reference to itself.
3. The `Usage` controller added owner reference to the `used` resource for itself, as proposed above.

```yaml
metadata:
  labels:
    usage: used
  name: test-hwgrp-pj72k
  ownerReferences:
  - apiVersion: nop.example.org/v1alpha1
    blockOwnerDeletion: true
    controller: true
    kind: XNopResource
    name: test-hwgrp
    uid: 697b0b9e-0a41-48ea-b624-007bbf444175
  - apiVersion: apiextensions.crossplane.io/v1alpha1
    kind: Usage
    name: test-hwgrp-g55x8
    uid: 35975f94-d3dd-444e-8095-c4894f81a089
```

So far, so good, as planned.

4. Deleted the composite, and I ended up with the `used` resource not deleted without any ownerReferences. After some thinking, I think this is expected, and here is what happens:
    1. The composite is deleted.
    2. On the used resource, the garbage collector finds an owner ref to a non-existing resource, but it is **not** the only owner ref. So, removes the owner ref to the composite from the owner ref list but do not delete the resource since there is another owner, i.e. the Usage resource.
    3. The Usage controller acts after GC and removes its owner ref.
    4. We end up an orphaned composed resource without any owner ref.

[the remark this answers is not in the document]
turkenh: My plan for now:

I'll focus on implementing the first version of my PR without this and put more thought and look for alternative solutions (like [tricks to re-queue for GC](https://github.com/crossplane/crossplane/pull/4215#discussion_r1238738203)) later.

design/one-pager-generic-usage-type.md
turkenh: One thing that I don't like with the [current convention](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#multiple-resource-reference) is, users would need to provide group & resource which is less intuitive than apiVersion & kind. The latter could easily be copied/pasted from the object to be referenced but when we use group & resource, they would need to make a conversion, e.g. `Object` -> `objects` and `kubernetes.crossplane.io/v1alpha1` -> `kubernetes.crossplane.io`

bobh66: Can we do the conversion internally - remove version from group and convert kind to resource?

Issue #4095, opened 2023-05-23
disabling validating webhooks requires manual intervention


mateusz-lubanski-sinch:
### What happened?
We upgraded crossplane from `1.11.0` to `1.12.11` which by default enabled [validating webhook](https://github.com/crossplane/crossplane/blob/master/design/design-doc-composition-validating-webhook.md). After disabling it on HelmChart (via `webhooks.enabled: false`) I had to manually:

1. delete validating webhook
```
kubectl delete validatingwebhookconfiguration crossplane
```
2. delete `spec.webhookTLSSecretName` property from `providerrevisions.pkg.crossplane.io` objects
```
for providerrevision in $(kubectl get providerrevisions.pkg.crossplane.io --output=jsonpath={.items..metadata.name}); do kubectl patch providerrevisions.pkg.crossplane.io $providerrevision --type=json -p='[{"op": "remove", "path": "/spec/webhookTLSSecretName"}]'; done
```

### What environment did it happen in?

Crossplane version: `1.21.1`
Kubernetes version: `EKS 1.23.17`

phisco:
This is because we are not handling webhooks via helm, but directly from the operator, which is only creating them, but not cleaning them up if "disabled" later on.
Keeping this approach, would require either periodically, or at least at startup, to delete the any existing but not required ValidatingWebhookConfiguration. The only other option I can think of would be to move the creation of the ValidatingWebhookConfiguration to the helm chart, leaving only the ca injection to the operator.

phisco:
the issue

PR #4246, opened 2023-06-22
Update all the runtimes

negz:

Closes https://github.com/crossplane/crossplane/pull/4223

Namely crossplane-runtime, and controller-runtime. Also k8schain.

A few breaking changes in this new controller-runtime. We need to bump to be able to bump to k8s v0.27.x, which other dependencies like k8schain are starting to require.

I have:

E2E tests are passing. 😄


cmd/crossplane/core/core.go
```diff
@@ -238,20 +245,14 @@ func (c *startCommand) Run(s *runtime.Scheme, log logging.Logger) error { //noli
 		return errors.Wrap(err, "Cannot add packages controllers to manager")
 	}
 
-	if c.WebhookTLSCertDir != "" {
```
phisco, line 241: Does this mean we are going to deploy webhooks always now? (I have to check the code better, just wanted to drop a comment before forgetting and before this got merged)

negz: Yeah this is something I want to understand better too. I _think_ before this change:

* A webhook server was always started and listening (by controller runtime).
* If `c.WebhookTLSCertDir` was supplied, we'd tweak its config and register our webhooks.
* We'd only actually expose the webhook server if `webhooks.enabled` was true in the Helm chart

My thinking with this change was it might be fine to just register the webhooks whether they were going to be exposed/used or not.

negz: Hrm, no this now breaks. I get the following when I set `webhooks.enabled=false`:

```
crossplane: error: core.startCommand.Run(): Cannot start controller manager: open /tmp/k8s-webhook-server/serving-certs/tls.crt: no such file or directory
```

phisco: yes, that was what I was expecting, due to how webhooks are handled we also hit this issue: https://github.com/crossplane/crossplane/issues/4095
I don't think we should address it here, but could be worth keeping it in mind and if it becomes to complex we could think of alternatives, e.g. moving validatingwebhookconfiguration creation to the helm chart and leave in the controller only the ca injection

negz: I tweaked things a bit (see new commits) but it's still not passing the E2E suite in https://github.com/crossplane/crossplane/pull/4187.

negz: Looks like a lot of breaking changes related to webhooks in controller-runtime v0.15. I've added a commit that implements an `admission.CustomValidator`, but I'm still seeing failures when I run the E2E tests. Interestingly the failures are now to do with XRDs. 🤔 

```
2023-06-22T09:10:09Z    ERROR   crossplane      Reconciler error        {"controller": "defined/compositeresourcedefinition.apiextensions.crossplane.io", "controllerGroup": "apiextensions.crossplane.io", "controllerKind": "CompositeResourceDefinition", "CompositeResourceDefinition": {"name":"xnopresources.nop.example.org"}, "namespace": "", "name": "xnopresources.nop.example.org", "reconcileID": "3defa3c9-085d-4b2f-9f77-b88eb7322ec3", "error": "cannot add composite resource finalizer: cannot update object: Internal error occurred: failed calling webhook \"compositeresourcedefinitions.apiextensions.crossplane.io\": failed to call webhook: the server could not find the requested resource"}
```

internal/validation/apiextensions/v1/composition/handler.go
```diff
@@ -50,22 +44,21 @@ const (
...
-	options controller.Options
-}
+// Error strings.
+const (
+	errNotComposition = "supplied object was not a Composition"
+	errUnexpectedOp   = "unexpected operation"
+	errValidationMode = "cannot get validation mode"
 
-// InjectDecoder injects the decoder.
-func (h *handler) InjectDecoder(decoder *admission.Decoder) error {
```
negz, line 61: This no longer works with controller-runtime v0.15.

internal/validation/apiextensions/v1/composition/handler.go
```diff
@@ -74,103 +67,91 @@ func SetupWebhookWithManager(mgr ctrl.Manager, options controller.Options) error
 		}
 	}
 
-	// TODO(lsviben): switch to using admission.CustomValidator when https://github.com/kubernetes-sigs/controller-runtime/issues/1896 is resolved.
-	mgr.GetWebhookServer().Register(v1.CompositionValidatingWebhookPath,
-		&webhook.Admission{Handler: &handler{
-			reader:  unstructured.NewClient(mgr.GetClient()),
```
negz, line 80: I removed the `unstructured.NewClient` that wraps the manager's client here. I can't see any reason we'd need it. Let me know if I missed it.

internal/validation/apiextensions/v1/composition/handler.go
```diff
@@ -50,22 +44,21 @@ const (
...
-}
+	errFmtTooManyCRDs = "more than one CRD found for %s.%s: %v"
+	errFmtGetCRDs     = "cannot get the needed CRDs: %v"
+)
 
 // SetupWebhookWithManager sets up the webhook with the manager.
 func SetupWebhookWithManager(mgr ctrl.Manager, options controller.Options) error {
 	if options.Features.Enabled(features.EnableAlphaCompositionWebhookSchemaValidation) {
+		// TODO(negz): It's not obvious what this does and why. Someone who
+		// understands it should add a comment. :)
```
negz, line 61: @phisco Do you know. I can kinda guess but it was confusing enough that a comment would be useful.

internal/validation/apiextensions/v1/composition/handler.go
phisco: yes, this and the whole implementation of the Handler can be removed, it was only needed because we weren't able to use the CustomValidator until https://github.com/kubernetes-sigs/controller-runtime/issues/1896 was merged and released as said a few line below.

internal/validation/apiextensions/v1/composition/handler.go
phisco: Something like this would make sense?
```suggestion
		// set up indexes to be able to list CRDs by Group and Kind they define later on
```

internal/validation/apiextensions/v1/composition/handler.go
phisco: I think it was needed for rendered resource validation IIRC, fine removing it now, I'll have to manually test it though, would be nice to have an e2e for this.

cmd/crossplane/core/core.go
negz: Found it - XRD needed its methods updated to be a valid `admission.Validator`. Pushed a commit.

internal/validation/apiextensions/v1/composition/handler.go
negz: @phisco I added a rudimentary E2E test, which seems to pass. Can you take a look and confirm whether it's actually testing anything useful? Maybe I should enable strict validation on the relevant `Composition`.

test/e2e/apiextensions_test.go
```diff
@@ -137,3 +138,74 @@ func TestComposition(t *testing.T) {
...
+	setup := funcs.AllOf(
+		funcs.AsFeaturesFunc(funcs.HelmUpgrade(HelmOptions(helm.WithArgs("--set args={--debug,--enable-composition-webhook-schema-validation}"))...)),
+		funcs.ReadyToTestWithin(1*time.Minute, namespace),
+	)
+
+	// Disable our feature flag.
+	teardown := funcs.AllOf(
+		funcs.AsFeaturesFunc(funcs.HelmUpgrade(HelmOptions()...)),
+		funcs.ReadyToTestWithin(1*time.Minute, namespace),
+	)
```
negz, line 199: I wonder whether it would make sense to have every test "upgrade" the Crossplane deployment to have whatever flags they need. Kinda weird, but it could be more resilient than relying on tests that change how Crossplane is installed to remember to undo their changes.

internal/validation/apiextensions/v1/composition/handler.go
phisco: The test case looks good to me as an initial smoke test, I would maybe just add another step checking that an invalid composition is rejected, e.g.:
Change the XRD to have one more `evenCoolerField` of type integer:
```yaml
metadata:
  name: xnopresources.nop.example.org
spec:
  group: nop.example.org
  names:
    kind: XNopResource
    plural: xnopresources
  claimNames:
    kind: NopResource
    plural: nopresources
  connectionSecretKeys:
  - test
  versions:
  - name: v1alpha1
    served: true
    referenceable: true
    schema:
     openAPIV3Schema:
       type: object
       properties:
        spec:
          type: object
          properties:
            coolField:
              type: string
          required:
          - coolField
        status:
          type: object
          properties:
            coolerField:
              type: string
            evenCoolerField:               # <== added field
              type: integer
```
which then we are going to patch from a string without any transformation:
```yaml
---
kind: Composition
metadata:
  name: xnopresources.nop.example.org
spec:
  compositeTypeRef:
    apiVersion: nop.example.org/v1alpha1
    kind: XNopResource
  resources:
  - name: nop-resource-1
    base:
     kind: NopResource
     spec:
      forProvider:
        conditionAfter:
        - conditionType: Ready
          conditionStatus: "False"
          time: 0s
        - conditionType: Ready
          conditionStatus: "True"
          time: 10s
    patches:
      fromFieldPath: spec.coolField
      toFieldPath: metadata.annotations[cool-field]
      transforms:
      - type: string
        string:
         type: Convert
         convert: ToUpper
    - type: ToCompositeFieldPath
      fromFieldPath: metadata.annotations[cool-field]
      toFieldPath: status.coolerField
    - type: ToCompositeFieldPath                                # <== added patch
      fromFieldPath: metadata.annotations[cool-field]
      toFieldPath: status.evenCoolerField
```
And once applied we should get an error, I wouldn't check here the specific error, as that's checked by the unit tests, but manually applying it results in the following error:
```
The Composition "xnopresources.nop.example.org" is invalid: spec.resources[0].patches[2].transforms: Required value: the fromFieldPath does not have a type compatible with the toFieldPath according to their schemas and no transforms were provided: string != integer
```

apis/apiextensions/v1/xrd_webhooks.go
```diff
@@ -35,49 +36,55 @@ const (
 	errConversionWebhookConfigRequired = "spec.conversion.webhook is required when spec.conversion.strategy is 'Webhook'"
 )
 
+// NOTE(negz): We only validate updates because we're only using the validation
+// webhook to enforce a few immutable fields. We should look into using CEL per
+// https://github.com/crossplane/crossplane/issues/4128 instead.
```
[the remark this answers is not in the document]
negz: Let me know if there's an action for me to take here.

apis/apiextensions/v1/xrd_webhooks.go
```diff
@@ -35,49 +36,55 @@ const (
...
+
 // +kubebuilder:webhook:verbs=update,path=/validate-apiextensions-crossplane-io-v1-compositeresourcedefinition,mutating=false,failurePolicy=fail,groups=apiextensions.crossplane.io,resources=compositeresourcedefinitions,versions=v1,name=compositeresourcedefinitions.apiextensions.crossplane.io,sideEffects=None,admissionReviewVersions=v1
 
 // ValidateCreate is run for creation actions.
-func (in *CompositeResourceDefinition) ValidateCreate() error {
+func (in *CompositeResourceDefinition) ValidateCreate() (admission.Warnings, error) {
+	// TODO(negz): Does this code ever get exercised in reality? We don't
+	// register the update verb when we generate a configuration above.
```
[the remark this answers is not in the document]
negz: Let me know if there's an action for me to take here.

apis/apiextensions/v1/xrd_webhooks.go
```diff
@@ -35,49 +36,55 @@ const (
...
 // ValidateDelete is run for delete actions.
-func (in *CompositeResourceDefinition) ValidateDelete() error {
-	return nil
+func (in *CompositeResourceDefinition) ValidateDelete() (admission.Warnings, error) {
+	return nil, nil
 }
 
 // SetupWebhookWithManager sets up webhook with manager.
-func (in *CompositeResourceDefinition) SetupWebhookWithManager(mgr ctrl.Manager) error {
+func SetupWebhookWithManager(mgr ctrl.Manager) error {
```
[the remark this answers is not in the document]
negz: > Do you expect to add here the code to setup al the webhooks for all resources in this package in one go?

Eventually - that's would be more similar to what we do with controllers (per the TODO in `core.go`).

Largely my thinking here was that where it's called:

```go
apiextensionsv1.SetupWebhookWithManager(mgr)
```

Seems a lot simpler (and perhaps more idiomatic?) than the approach where this is a method on a type:

```go
(&apiextensionsv1.CompositeResourceDefinition{}).SetupWebhookWithManager(mgr)
```

To me, it's a little confusing/unintuitive that calling Setup on one (empty) instance of a type is registering the entire type with the controller manager. Admittedly that's still happening with my update - it's just more in line with how we do every other `For(T)` call in the codebase, and IMO reads more like "setup a webook for this type" rather than "setup a webhook for this instance of this type".

internal/validation/apiextensions/v1/composition/handler.go
```diff
@@ -50,22 +44,21 @@ const (
...
+	errUnexpectedOp   = "unexpected operation"
+	errValidationMode = "cannot get validation mode"
 
-// InjectDecoder injects the decoder.
-func (h *handler) InjectDecoder(decoder *admission.Decoder) error {
-	h.decoder = decoder
-	return nil
-}
+	errFmtTooManyCRDs = "more than one CRD found for %s.%s: %v"
+	errFmtGetCRDs     = "cannot get the needed CRDs: %v"
```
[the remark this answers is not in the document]
negz: That's good to know. In cases like this the best thing to do is to raise an issue to discuss changing the pattern. Otherwise we start to have some files that are look like @phisco wrote them according to his preferences, others like @negz wrote them etc. 🙂 Would you mind raising an issue to discuss?

internal/validation/apiextensions/v1/composition/handler.go
```diff
@@ -74,103 +67,91 @@ func SetupWebhookWithManager(mgr ctrl.Manager, options controller.Options) error
...
+	gkToCRD, errs := v.getNeededCRDs(ctx, comp)
+	// if we have errors, and we are in strict mode or any of the errors is not
+	// a , return them
 	if len(errs) != 0 {
 		if validationMode == v1.CompositionValidationModeStrict || containsOtherThanNotFound(errs) {
-			return warns, xperrors.Errorf("there were some errors while getting the needed CRDs: %v", errs)
+			// TODO(negz): Do we really need to return all the errors? Typically
+			// we just return the first one we hit. This works well with
+			// errors.Wrap. Is there extra value in returning them all here as a
+			// special case?
```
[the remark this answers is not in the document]
negz: That seems like a good idea. I think first step would be to surface that function in `pkg/errors` in crossplane-runtime. I'll update the TODO.

I can see in this case where we're validating an object that it's more worth it than usual to try process the whole thing and return all errors so that you don't have to iterate through each error case to discover them.

internal/validation/apiextensions/v1/composition/handler.go
```diff
@@ -74,103 +67,91 @@ func SetupWebhookWithManager(mgr ctrl.Manager, options controller.Options) error
...
+			return warns, errors.Errorf(errFmtGetCRDs, errs)
 		}
-		// if we have errors, but we are not in strict mode, and all of the errors are not found errors,
-		// just move them to warnings and skip any further validation
-		// TODO(phisco): we are playing it safe and skipping validation altogether, in the future we might want to also support partially available inputs
+		// if we have errors, but we are not in strict mode, and all of the
+		// errors are not found errors, just move them to warnings and skip any
+		// further validation TODO(phisco): we are playing it safe and skipping
+		// validation altogether, in the future we might want to also support
+		// partially available inputs
```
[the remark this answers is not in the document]
negz: Agreed - wrapped that mistakenly.

test/e2e/apiextensions_test.go
```diff
@@ -137,3 +138,74 @@ func TestComposition(t *testing.T) {
 			Setup(setup).Feature(),
 	)
 }
+
+func TestValidation(t *testing.T) {
+
+	manifests := "test/e2e/manifests/apiextensions/validation/composition-schema"
+	composition := features.Table{
```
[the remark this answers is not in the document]
negz: I don't think I understand what you mean. That said, I don't want to change the established pattern in this PR. Could you raise a new issue (and label it `e2e`) to discuss please.

test/e2e/apiextensions_test.go
negz: I'm going to keep this as is - we can change if/when we find a better pattern.

apis/apiextensions/v1/xrd_webhooks.go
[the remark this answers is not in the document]
phisco: If you want to enable it by adding the create verb to the annotation above, otherwise I can do it on a separate PR next week.

apis/apiextensions/v1/xrd_webhooks.go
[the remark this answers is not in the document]
phisco: I agree it's not too nice, but I got used to it probably, it's a pretty common pattern: https://grep.app/search?q=Setupwebhookwithmanager

internal/validation/apiextensions/v1/composition/handler.go
negz: Done!

apis/apiextensions/v1/xrd_webhooks.go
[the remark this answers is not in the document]
negz: That seems like it could be a functional change, so let's keep it in a separate PR.

apis/apiextensions/v1/xrd_webhooks.go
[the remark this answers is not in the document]
negz: Yeah it's not fundamentally wrong, and is what controller-runtime scaffolds I think. We diverged from that scaffolding with our `Setup` (for controller) pattern so I'd like to keep consistent.

negz:
Adding a note that we now get pprof for free with controller-runtime, so I'd like to remove the one I manually setup. Doesn't need to block this PR though - can follow up later. https://github.com/kubernetes-sigs/controller-runtime/pull/1943

internal/validation/apiextensions/v1/composition/handler_test.go
```diff
@@ -14,7 +14,8 @@ See the License for the specific language governing permissions and
...
 */
 
-// Code generated by client-gen. DO NOT EDIT.
+package composition
 
-// This package has the automatically generated clientset.
-package versioned
+import "sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+var _ admission.CustomValidator = &validator{}
```
phisco, line 21: Why the need to have this in a test file instead of directly in internal/validation/apiextensions/v1/composition/handler.go?

phisco (approved):

internal/validation/apiextensions/v1/composition/handler_test.go
negz: Because it's a test. 🙂 We're testing that our `*validator` type satisfies the interface we need it to satisfy. No need for this code to be part of our production binary.

PR #4250, opened 2023-06-22
Allow Scaling of Controllerconfig

Breee:

Allows scaling of  `ControllerConfig` kind  wrt https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#scale-subresource 

This change enables automatic scaling of providers, e.g. by using KEDA

I have:

Tested in a kubernetes cluster with crossplane installed.

### DIFF
--- a/cluster/crds/pkg.crossplane.io_controllerconfigs.yaml
+++ b/cluster/crds/pkg.crossplane.io_controllerconfigs.yaml
@@ -3163,4 +3163,14 @@ spec:
         type: object
     served: true
     storage: true
-    subresources: {}
+    subresources:
+      # status enables the status subresource.
+      status: {}
+      # scale enables the scale subresource.
+      scale:
+        # specReplicasPath defines the JSONPath inside of a custom resource that corresponds to Scale.Spec.Replicas.
+        specReplicasPath: .spec.replicas
+        # statusReplicasPath defines the JSONPath inside of a custom resource that corresponds to Scale.Status.Replicas.
+        statusReplicasPath: .status.replicas
+        # labelSelectorPath defines the JSONPath inside of a custom resource that corresponds to Scale.Status.Selector.
+        labelSelectorPath: .status.labelSelector


bobh66:
Thanks for the PR @Breee - can you provide some more context around these changes?

It's true that the providers are running as Deployments and technically can be scaled, but as kubernetes controllers they run in an active/standby configuration with an elected leader, and only one pod can process events/workload at a time.  Adding more pods to the deployment would not affect the throughput rate, and there is even debate about whether having a "standby" is useful, since Kubernetes can usually restart the failed active pod faster than the standby pod can realize that it should take over.

Thanks

Breee:
> Thanks for the PR @Breee - can you provide some more context around these changes?
> 
> It's true that the providers are running as Deployments and technically can be scaled, but as kubernetes controllers they run in an active/standby configuration with an elected leader, and only one pod can process events/workload at a time. Adding more pods to the deployment would not affect the throughput rate, and there is even debate about whether having a "standby" is useful, since Kubernetes can usually restart the failed active pod faster than the standby pod can realize that it should take over.
> 
> Thanks

Greetings,  yes, I also realized that the leader election avoids that replicas are useful in any way. We are currently testing what happens if we disable leader election - we observe that all replicas process events, however, i'm not sure yet if everything is processed correctly and without conflicts. 


More context: 

Crossplane is very slow when we, for example, create 200 objects. In our case we created a keycloak-provider  to manage roles, groups and memberships in keycloak. 
Crossplane will become really slow once it has to resolve references and objects  depend on the creation of other objects, e.g.  memberships require roles, which have to be created first. 
I would like to increase the processing speed of crossplane - my initial idea was to scale the amount of provider containers

bobh66:
Thanks for the context @Breee - I think the main concern with running multiple replicas is that there is no way to prevent the same resource being processed by multiple replicas at the same time, with different events handled by different replicas, and resulting in "last update wins" race conditions.

Crossplane core and providers will take advantage of all of the memory and CPU that they are provided with, so you can use the `--max-reconcile-rate` and `--pollInterval` flags to increase the number of parallel reconciles and increase or decrease the polling interval as you desire.  I would suggest adding more CPU/memory to the worker(s) and make sure the pods are configured with the reconcile rate and poll interval that you need.

bobh66:
>  We have to be able to shard and process events in parallel across multiple nodes, i.e. horizontal scaling instead of vertical

I don't know of a way to do this with Kubernetes resource controllers, but if you find a way to make it work that would be great to know.

jbw976 (commented):
Which particular providers are you using @Breee? If you're using Upbound official providers, have you already been able to take a look at this sizing guide?
https://github.com/upbound/upjet/blob/main/docs/sizing-guide.md

negz:
Yeah, I'd prefer not to proceed with this. We know it won't help unless we figure out how to share work across controllers better. We also know we want to replace ControllerConfig with something better.

I agree in spirit with wanting to be able to scale providers horizontally though. @Breee could you raise an issue tracking that (i.e. the need, not so much a potential solution) please.

Breee:
See https://github.com/crossplane/crossplane/issues/4269

bittupaul:
Is there any update on this horizontal scaling problem?
I am running a controller through which I want to reconcile 200k resources; however, that's not possible to do it quickly if we are not horizontally scaling.

The approach that's in my mind is - 
1. Run controller instances in a multi-pod setup.
2. Disable the leader pod (so that all pods can be master).
3. Use distributed locking to ensure each pod reconciles resources exclusively (there will be contention to acquire lock for a same resource, but once acquired that resource won't be reconciled by other pod)

But this exposes me to one more challenge - How do I ensure that when a new resource (CR) is deployed/created in the k8s cluster, all pods do not again run into the contention for consuming the k8s event and create corresponding resource in external system (using external provider API)?

bittupaul:
Also, is there a way to access the work queue which has the k8s events, based on which crossplane controller methods are invoked - CRUD/Observer/Reconcile?

Issue #4228, opened 2023-06-21
XFN functions fail to run on GKE due to default AppArmour denying mount inside pod


haraldsk:
### What happened?

All composition functions on GKE 1.24 cluster running Crossplane 1.11.3 fails with the following log message in the crossplane container:

`sources: cannot run Composition Function pipeline: cannot run function "template-redacted-xyz": cannot run container: rpc error: code = Unknown desc = exit status 1: xfn: error: spark.Command.Run(): OCI runtime error: exit status 1`

After instrumenting the code custom debug code the failure was found while xfn is running the crun run command:

`crun --root=/xfn/runtime run  --bundle=/xfn/c/101205c0-0711-47fe-b26b-b1a79452dc40101205c0-0711-47fe-b26b-b1a79452dc40`

returns the following strderr: `set propagation for /: Permission denied`

This is due to the fact that Containerd on [GKE runs AppArmour](https://cloud.google.com/container-optimized-os/docs/how-to/secure-apparmor)  whichuses the [Docker Default security profile which denies mount inside a pod.](https://github.com/moby/moby/blob/master/profiles/apparmor/template.go#L42)

As a workaround setting the following setting in values.yaml i the Crossplane Helm chart.

```yaml
customAnnotations:
  "container.apparmor.security.beta.kubernetes.io/crossplane-xfn": "unconfined"
```

### How can we reproduce it?

Install XP 1.11.3 with the following Helm values, on GKE 1.24 - 1.27

```yaml
# crossplane helm chart
args:
  - "--debug"
  - "--enable-composition-functions"
  - "--enable-environment-configs"

# compostion-functions alpha feature
xfn:
  enabled: true
  args:
    - "--debug"
```

Run ANY composition function.

### What environment did it happen in?
Crossplane version:  Tested, 1.11.0, 1.11.3 
GKE 1.24 and 1.27
AKS also seems affected.

negz:
Ah - good catch. This is working as intended. Unfortunately the way `xfn` is implemented requires some additional syscalls (e.g. `unshare`) that most default AppArmor and seccomp profiles won't allow. This is part of why `xfn` itself is swappable, and could be replaced with another function runner implementation (e.g. if this tradeoff isn't acceptable).

We already run seccomp unconfined in the Helm chart, but I had forgotten to add this annotation to run AppArmor unconfined. Opened a PR to address that.

PR #4255, opened 2023-06-23
Support `-create-kind-cluster=false` in E2E tests

negz:

This allows you to run E2E tests against an existing Kubernetes cluster. Note that we can't load our local build of Crossplane into an existing Kubernetes cluster, so if you pass this flag you'll also need to make sure Crossplane is installed on the cluster where you'll run the tests.

I have:

### DIFF
--- a/test/e2e/README.md
+++ b/test/e2e/README.md
@@ -42,6 +42,10 @@ E2E_TEST_FLAGS="-feature=Install" make e2e
 
 # Stop immediately on first test failure, and leave the kind cluster to debug.
 E2E_TEST_FLAGS="-test.v -test.failfast -destroy-kind-cluster=false"
+
+# Use an existing Kubernetes cluster. Note that the E2E tests can't deploy your
+# local build of Crossplane in this scenario, so you'll have to do it yourself.
+E2E_TEST_FLAGS="-create-kind-cluster=false -destroy-kind-cluster=false -kubeconfig=$HOME/.kube/config"
 ```
 
 ## Test Parallelism
--- a/test/e2e/main_test.go
+++ b/test/e2e/main_test.go
@@ -88,27 +88,33 @@ var helmOptions = []helm.Option{
 var environment env.Environment
 
 func TestMain(m *testing.M) {
+	create := flag.Bool("create-kind-cluster", true, "create a kind cluster (and deploy Crossplane) before running tests")
 	destroy := flag.Bool("destroy-kind-cluster", true, "destroy the kind cluster when tests complete")
 
 	clusterName := envconf.RandomName("crossplane-e2e", 32)
 	environment, _ = env.NewFromFlags()
-	environment.Setup(funcs.EnvFuncs(
-		// TODO(negz): If we want to support running tests on existing clusters
-		// (e.g. GKE, EKS, your own kind cluster) we could wrap most of this in
-		// an env.Func that skips everything but adding types to the scheme when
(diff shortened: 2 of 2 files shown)


negz:
Hrm, I just realized this won't work with at least some tests. For example `TestCrossplane` uninstalls, reinstalls, and upgrades Crossplane using Helm. 🤔 It won't be able to do that if it doesn't have a build to upgrade to.

I could still see this functionality being useful for test that _don't_ have this restriction. I do think we'll have more such tests in future though - e.g. any test that wants to toggle a feature flag. Some thoughts:

 * We could label any test that needs to be able to manage Helm releases to work.
 * We could better parameterize specifically what Helm chart and Crossplane image(s) were being used - take them as CLI flags.

phisco:
We should probably identify different kind of tests that should not overlap, most probably:
- upgrade test from the latest stable
- tests requiring the default installation
- a sequence of tests requiring a particular configuration of Crossplane itself, e.g. alpha feature enable
  - one could argue that given how we handle alpha features we could even think of enabling them all as part of the default installation above, given that if a feature is in alpha it's hopefully going be enabled by default soon once it gets to beta, and given that we are usually careful not overlapping with the default behaviour in breaking ways
  - if we decide to go with "enable all alpha features by default" we should definitely label these tests according to their requirements, be able to disable one or more features and avoid running tests requiring it, to be able to understand whether a failure is due to something an alpha feature is wrong.
  - This could also help us have some feedback on alpha features, which sometimes are just forgotten and we only discover bugs once we promote them to beta (probably just a problem for more niche features)

Regarding the ability to run the suite of tests against a live cluster, I totally agree it's a necessary thing, ideally i would like to be able to spin up my own cluster, deploy Crossplane with the configuration i want and run the suite of tests against it, whitelisting any test that makes sense for my configuration via labels

negz:
> We should probably identify different kind of tests that should not overlap

FWIW no tests overlap today. See https://github.com/crossplane/crossplane/tree/master/test/e2e#test-parallelism. What you propose though is pretty close to how we achieve quasi-parallelism in CI by starting a separate test run (with its own cluster) per feature area label - for example the test that installs and uninstalls Crossplane is run in its own `lifecycle` runner.

> ideally i would like to be able to spin up my own cluster, deploy Crossplane with the configuration i want and run the suite of tests against it

Can you elaborate a little more on the use cases for this? Intuitively I agree but I've had a hard time imagining scenarios where I'd want to. Given that the E2E tests only test the fixtures they create, I would have to imagine this would be mostly about:

1. Testing Crossplane on different "flavors" of Kubernetes - i.e. GKE, EKS, etc.
2. Testing different builds of Crossplane?

1 feels like a low priority to me - I can't really think of _many_ reasons Crossplane core would care what cloud provider it was running in. There are _some_ though - e.g. we might have caught https://github.com/crossplane/crossplane/issues/4228 in E2E tests.

2 I'm less sure about. If you have a build of Crossplane you want to test, can't you just checkout that commit and `make e2e`? 🤔

negz:
> enabling them all

I think the most important thing is to test Crossplane with none of these flags enabled - i.e. to make sure "GA" is rock solid.

I could imagine updating the matrix that runs tests to run:

* All tests with no flags enabled.
* All tests with all flags enabled.
* All tests with each individual flag enabled.

Probably not something we'd want to do on every PR though.

negz:
I'd still like to proceed with this change - it's useful as-is for at least some tests. We can always iterate to figure out the flags/Helm part of it.

phisco (approved):

Issue #4257, opened 2023-06-23
Push all E2E packages to xpkg.upbound.io


negz:
### What problem are you facing?

The E2E tests currently install two Configuration packages. I imagine we'll have more in future.

https://github.com/crossplane/crossplane/blob/297f760c7de146f2b87d013379ac881754b0a197/test/e2e/manifests/pkg/configuration/private/configuration.yaml#L9

https://github.com/crossplane/crossplane/blob/297f760c7de146f2b87d013379ac881754b0a197/test/e2e/manifests/pkg/configuration/dependency/configuration.yaml#L6

One of these images is pulled from a private Google Artifact Registry. The credentials required to pull it don't grant access to anything else so we have them checked in in plaintext.

### How could Crossplane help solve your problem?

I'd like to get these packages pushed to xpkg.upbound.io, probably in a new organization (`crossplane-e2e`?). This will get all our test dependencies:

* In one place, even if they need to be private. (Private Docker Hub listings cost us money.)
* In an org that makes it clear they're for E2E tests, not a "real" Crossplane thing.

Note that the package metadata is checked in under `test/e2e/manifests`, but there's no CI job that automatically builds and pushes these images at the moment. I don't think they change often enough that it's worth automating build/push.

negz:
/fresh

I think we should still do this.

negz:
I wonder whether we should create an `xpkg.upbound.io/crossplane-e2e` org to push these packages to?

Currently we're a little all over the place. We have:

 * us-west1-docker.pkg.dev/crossplane-playground/
 * Packages in xpkg.upbound.io/crossplane that start with e2e
 * Packages in xpkg.upbound.io personal orgs (e.g. [ezgiorg](https://github.com/crossplane/crossplane/blob/79c3d347fbb9646ccfd7cc5e81ee2c936c630413/test/e2e/manifests/pkg/dependency-upgrade/version/configuration-initial.yaml#L7))

phisco:
Totally agree, if we could automate pushing them too that would be awesome.

haarchri:
@negz @jbw976 think we can close this issue ? We using ghcr as location ?

jbw976:
I think the intent of the issue still exists (have e2e packages in a single consolidated place that is clearly intended for test usage), we could just change the location to `xpkg.crossplane.io` 😉

PR #4261, opened 2023-06-23
fix: pull xfn images from registry configured with custom TLS certificate

### LINKED ISSUE #4041: xfn should support custom CA bundles
### What happened?
Before enabling Composition functions, I could override the Crossplane Helm chart with my root internal CABundle using a configmap and it worked as expected. 

image

When I enabled Composition functions, the XFN container receives 509 (cert) error when it attempts to pull function containers from an open Docker registry that's expecting our internal CABundle. 

Is there a way to explicitly configure which CABundle the XFN container uses?  

Is the Helm chart override not setting the XFN container CABundle? 

Crossplane version: v1.11.2

pedjak:

Additional `--ca-bundle-path` param got introduced to `xfn spark` command,
enabling to specify path to custom root certificate, e.g. `/cert/domain.crt`
The value can be passed as well via `CA_BUNDLE_PATH` env variable.

If set, the certificate will be added to the pool of system-wide certificates
and will be used for verifying the authority of the registry by each pull.

Helm chart templates are adjusted so to that the custom certificate
gets mounted into `crossplane-xfn` container and `CA_BUNDLE_PATH` is
set properly.

Additional changes:
* added e2e tests under `test/e2e` verifying that
  `--ca-bundle-path` works as expected
* crafted a fn image under `test/e2e/testdata/images/labelizer` that help us to assert that the pulling happens for real
* CI job runs the e2e area: xfn

Fixes #4041

I have:


pedjak:
xfn integration tests passed in the [CI run](https://github.com/crossplane/crossplane/actions/runs/5355252629/jobs/9713326261?pr=4261#step:10:150).

cmd/xfn/spark/spark.go
```diff
@@ -154,51 +165,56 @@ func (c *Command) Run() error { //nolint:gocyclo // TODO(negz): Refactor some of
 	// recommended; 'run' is more for testing. In practice though run seems to
 	// work just fine for our use case.
 
-	//nolint:gosec // Executing with user-supplied input is intentional.
-	cmd := exec.CommandContext(ctx, c.Runtime, "--root="+root, "run", "--bundle="+b.Path(), runID)
-	cmd.Stdin = bytes.NewReader(req.GetInput())
+	var rsp *v1alpha1.RunFunctionResponse
+	if c.DryRun {
+		rsp = &v1alpha1.RunFunctionResponse{Output: req.GetInput()}
+	} else {
```
phisco, line 171: Can you return early instead of having the else branch?

test/e2e/xfn/xfn_test.go
```diff
@@ -0,0 +1,297 @@
...
+var xfnImageRef *string = flag.String("xfn-image-ref", "", "xfn runner image ref")
+
+func TestCustomCA(t *testing.T) {
+	cases := map[string]struct {
+		body func(xfnImageRef, certDir string) error
+	}{
+		"retrieve xfn image from private registry by setting CA cert via CLI arg": {
+			body: func(fnImageRef, certDir string) error {
+				// start xfn runner container
+				cmd := exec.Command("docker",
```
phisco, line 53: I would rather avoid relying so much on docker and running pods, isn't there any other way to test this? Otherwise I'd rather integrate it with the e2es as many test case are going to pull stuff from a registry, which is now a public registry with some credentials, so it might be worth solving it once and for all there, maybe spinning up a registry inside or outside the cluster and putting in place the necessary networking to be able to pull stuff from there.

cmd/xfn/spark/spark.go
```diff
@@ -74,6 +75,8 @@ type Command struct {
 	CacheDir      string `short:"c" help:"Directory used for caching function images and containers." default:"/xfn"`
 	Runtime       string `help:"OCI runtime binary to invoke." default:"crun"`
 	MaxStdioBytes int64  `help:"Maximum size of stdout and stderr for functions." default:"0"`
+	DryRun        bool   `help:"pull images, but skip running fn container"`
```
turkenh, line 78: AFAIU, this flag is introduced for use only during e2e tests, right?
Could we instead have a test image doing nothing (e.g. available at [crossplane-e2e](https://github.com/crossplane/crossplane/issues/4257)), or always succeeding and we invoke it during tests?

test/e2e/xfn/xfn_test.go
turkenh: Agree with @phisco here. Too many command invocations, likely to cause potential headaches.

pedjak: well, we need to spin `crossplane-xfn` container somewhere, and try to pull images from various registries. Ofcourse, we could also run a real e2e test, but then the fixture would be even larger - we would need to create composition, function and put them in some registry (with custom certificate) in order to verify that we are able to pull using custom CA bundle.

This is an integration test, does not require kubernetes to run, does not require crossplane to be installed, and we can in an easy way test some aspects of xfn runner. And docker is a dependency on any machine where we need to build and test.

Please let me know if we should drop these integration tests and go with full fledged e2e tests.

.github/workflows/ci.yml
```diff
@@ -228,6 +228,9 @@ jobs:
       - name: Run Unit Tests
         run: make -j2 test
 
+      - name: Run Integration Tests
+        run: make test-integration
```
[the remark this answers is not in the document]
pedjak: My understanding is that integration tests are already something that is part of `upbound/build` module. There is `go.test.integration` target there, and `test-integration` existed before `39f24122e64d879ff41affbffebe9b236f5b0d26` got merged.

Is the decision now to only allow unit and e2e tests in the code base? Please let me know if I need to drop these and replace them with e2e tests.

cmd/xfn/spark/spark.go
pedjak: > AFAIU, this flag is introduced for use only during e2e tests, right?

I introduced so that we can have simpler integration tests, not e2e test.

> Could we instead have a test image doing nothing (e.g. available at https://github.com/crossplane/crossplane/issues/4257), or always succeeding and we invoke it during tests?

You mean a test fn image that does nothing? Sure, we could have them if we wanna have real e2e tests. Where we would manage and build such image? From this repo?

turkenh: I think it would be in a similar category as [provider-nop](https://github.com/crossplane-contrib/provider-nop), so we can have it in a separate repo.

test/e2e/xfn/xfn_test.go
negz: > Please let me know if we should drop these integration tests and go with full fledged e2e tests.

Yes please - stick to the newly introduced E2E test pattern for now. If we feel there's a need to also have integration tests in addition to E2E and unit tests I'd like to discuss that first rather than organically introduce them alongside as part of feature PR.

internal/oci/pull.go
```diff
@@ -117,6 +121,14 @@ func WithPullAuth(a *ImagePullAuth) ImageClientOption {
 	}
 }
 
+// WithCustomCA adds given root certificates to tls client configuration
+func WithCustomCA(rootCAs *x509.CertPool) ImageClientOption {
+	return func(c *ImageClientOptions) {
+		c.transport = remote.DefaultTransport.(*http.Transport).Clone()
```
negz, line 127: It's unlikely, but if `remote.DefaultTransport` stops being a `*http.Transport` in a future go-containerregistry release we'd panic at runtime. Can we add or update a unit test to call this function? I'm not worried about testing its behaviour really, I just want to make sure we invoke it at build time to make sure it doesn't panic.

cmd/xfn/spark/spark.go
negz: Agreed - https://github.com/crossplane/crossplane/tree/master/contributing#return-early

cmd/xfn/spark/spark.go
```diff
@@ -126,12 +129,20 @@ func (c *Command) Run() error { //nolint:gocyclo // TODO(negz): Refactor some of
 		return errors.Wrap(err, errParseRef)
 	}
 
+	opts := []oci.ImageClientOption{FromImagePullConfig(req.GetImagePullConfig())}
+	if c.CABundlePath != "" {
+		rootCA, err := xpkg.ParseCertificatesFromPath(c.CABundlePath)
```
negz, line 134: I don't love `internal/oci` taking a dependency on `internal/xpkg` just for this function:

* Eventually I think we'll want `internal/xpkg` to depend on `internal/oci` - xpkgs are built on OCI images.
* Nothing about `ParseCertificatesFromPath` is specific to `xpkg` functionality.

Is there somewhere else `ParseCertificatesFromPath` could live? Given the function is pretty simple I'd be fine just duplicating it in this package if there's not an obvious place for it to live right now.

cmd/xfn/spark/spark.go
negz: I agree with @turkenh here.

.github/workflows/ci.yml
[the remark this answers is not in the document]
negz: > Is the decision now to only allow unit and e2e tests in the code base?

Yes, for now at least. We can talk about also adding integration tests if needed.

For context, the build submodule and our Makefiles use "E2E" and "integration" interchangeably. I also don't think anything actually uses the `go.test.integration` target today.

cmd/xfn/spark/spark.go
negz: As to where such an image would live - I think we can find a place. Today there's a few Configuration images used to test that I want to find a home for https://github.com/crossplane/crossplane/issues/4257. I think we could do the same for functions.

internal/oci/pull.go
```diff
@@ -117,6 +121,14 @@ func WithPullAuth(a *ImagePullAuth) ImageClientOption {
...
+		c.transport.TLSClientConfig = &tls.Config{RootCAs: rootCAs, MinVersion: tls.VersionTLS12}
```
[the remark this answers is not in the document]
pedjak: I have used the version that is used already in https://github.com/crossplane/crossplane/blob/master/internal/xpkg/fetch.go#L98 - should we also update that side?

cmd/xfn/spark/spark.go
pedjak: perhaps `internal/transport`?

internal/oci/pull.go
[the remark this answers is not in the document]
negz: Hrm, no, let's stick with 1.2 for now since requiring 1.3 would be a potentially breaking behaviour change. Thanks!

cmd/xfn/spark/spark.go
negz: Ah, you're right. I mistakenly thought this file was in `internal/oci` for some reason. I'd be okay with:

* Creating `internal/transport` as you suggest. Maybe premature for just one function though.
* Maybe just move the function into `internal/oci`? Rationale being OCI is more generic than xpkg
* Just proceed as you have it now and we figure it out later. 🙂

.github/workflows/ci.yml
```diff
@@ -254,6 +254,9 @@ jobs:
           version: ${{ env.DOCKER_BUILDX_VERSION }}
           install: true
 
+      - name: install skopeo
+        run: sudo apt-get -y install skopeo
+
```
phisco, line 259: We usually install dependencies as part of the make target, see https://github.com/upbound/build/blob/a81d4729219d410567f848d10ac6798d9862e398/makelib/golang.mk#L301 or https://github.com/upbound/build/blob/a81d4729219d410567f848d10ac6798d9862e398/makelib/k8s_tools.mk#L146, helm and kind are then just linked for the e2es in the `e2e-run-tests` target.
But if possible I'd rather avoid adding one more tool.

test/e2e/xfn_test.go
```diff
@@ -0,0 +1,318 @@
...
+func CopyImagesToRegistry() features.Func {
+	return func(ctx context.Context, t *testing.T, config *envconf.Config) context.Context {
+		nodes := &corev1.NodeList{}
+		if err := config.Client().Resources().List(ctx, nodes); err != nil {
+			t.Fatal("cannot list nodes", err)
+		}
+		if len(nodes.Items) == 0 {
+			t.Fatalf("no nodes in the cluster")
+		}
+		i := slices.IndexFunc(nodes.Items[0].Status.Addresses, func(a corev1.NodeAddress) bool {
```
phisco, line 211: Given that this is the only reason we are adding `golang.org/x/exp` as a dependency, which although nice, is still untagged and therefore would result in noisy version bumps every month (according to the schedule we've set for go dependencies referenced by digest), I'd say here we can just switch to a simple for loop.

test/e2e/funcs/feature.go
```diff
@@ -384,6 +384,17 @@ func DeleteResources(dir, pattern string) features.Func {
...
+func CreateNamespace(name string) features.Func {
+	return func(ctx context.Context, t *testing.T, config *envconf.Config) context.Context {
+		ns := &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: name}}
+		if err := config.Client().Resources().Create(ctx, ns); err != nil {
+			t.Fatalf(fmt.Sprintf("error creating namespace %s", name), err)
+		}
+		return ctx
+	}
+}
+
```
phisco, line 397: There is an almost identical function available in the e2e-framework, which we already use in a similar context [here](https://github.com/crossplane/crossplane/blob/444267e84783136daa93568b364a5f01228cacbe/test/e2e/install_test.go#L134).
If you feel there are reasons we should not use envfuncs.CreateNamespace, feel free to refactor also that usage and add here a comment with the reasons why this should be preferred over the other option.

test/e2e/xfn_test.go
```diff
@@ -0,0 +1,318 @@
...
+				funcs.ApplyResources(FieldManager, manifests, "composition.yaml"),
+				funcs.ResourcesCreatedWithin(30*time.Second, manifests, "composition.yaml"),
+			)).
+			Assess("ClaimIsCreated", funcs.AllOf(
+				funcs.ApplyResources(FieldManager, manifests, "claim.yaml"),
+				funcs.ResourcesCreatedWithin(30*time.Second, manifests, "claim.yaml"),
+			)).
+			Assess("ClaimBecomesAvailable", funcs.ResourcesHaveConditionWithin(5*time.Minute, manifests, "claim.yaml", xpv1.Available())).
+			Assess("ManagedResourcesProcessedByFunction", ManagedResourcedProcessedByFunction()).
+			// Teardown(funcs.AsFeaturesFunc(funcs.HelmUpgrade(HelmOptions()...))).
```
phisco, line 72: A few things to make this similar to the design [proposal](https://github.com/crossplane/crossplane/pull/4270/files#diff-7d9a55ec9712c1ccc931efe126dbea316dab465e6ab265abd089736be5511f47R88) we're going to merge soon:
- Use `WithSetup` instead of `Setup` (same for `Teardown`), as that allows to specify names for setup steps too.
- Avoid defining helper functions here, we'll refactor tests by adding helpers to the `funcs` package once we start having more tests and seeing patterns. So please, if required add an helper there, otherwise, for simple wrappers around a single `funcs.AllOf` such as `CrossplaneDeployedWithFunctionsEnabled` and `ProvideNopDeployed`, etc..., or around single functions such as `InstallDockerRegistry`.
- Avoid relying on another test's manifests, as you are doing in ProvideNopDeployed (which being hidden in an helper function, adding to the point above, was not so easy to spot), as that would make the tests brittle

test/e2e/xfn_test.go
```diff
@@ -0,0 +1,318 @@
...
+		helm.WithNamespace("reg"),
+		helm.WithWait(),
+		helm.WithChart("https://helm.twun.io/docker-registry-2.2.1.tgz"),
+		helm.WithArgs(
+			"--set service.type=NodePort",
+			"--set service.nodePort=32000",
+			"--set tlsSecretName=reg-cert",
+		),
+	))
+}
```
phisco, line 241: Can you add the repo and install it without referencing the tgz directly? see the Crossplane upgrade test for an example.

test/e2e/xfn_test.go
```diff
@@ -0,0 +1,318 @@
...
+					"--set registryCaBundleConfig.name=reg-ca",
+					"--set registryCaBundleConfig.key=domain.crt",
+				))...)),
+		funcs.ReadyToTestWithin(1*time.Minute, namespace),
+	)
+}
+
+// ProvideNopDeployed assets that provider-nop is deployed and healthy
+func ProvideNopDeployed() features.Func {
+	manifests := "test/e2e/manifests/apiextensions/composition/minimal/prerequisites"
```
phisco, line 261: see comment above, let's avoid relying on other tests' prerequisites as that could lead to hard to debug issues.

test/e2e/xfn_test.go
```diff
@@ -0,0 +1,318 @@
...
+
+	v1 "github.com/crossplane/crossplane/apis/apiextensions/v1"
+	"github.com/crossplane/crossplane/test/e2e/funcs"
+)
+
+func TestFunctions(t *testing.T) {
+
+	manifests := "test/e2e/manifests/apiextensions/composition/functions/private-registry"
+	environment.Test(t,
+		features.New("PullImageFromPrivateRegistryWithCustomCert").
```
phisco, line 54:
```suggestion
func TestFunctionsFromPrivateRegistryWithCustomCert(t *testing.T) {

	manifests := "test/e2e/manifests/apiextensions/composition/functions/private-registry"
	environment.Test(t,
		features.New("FunctionsFromPrivateRegistryWithCustomCert").
```

go.mod
```diff
@@ -154,7 +156,7 @@ require (
 	github.com/spf13/pflag v1.0.5 // indirect
 	github.com/tetratelabs/wazero v1.2.1 // indirect
 	github.com/vbatts/tar-split v0.11.3 // indirect
-	github.com/vladimirvivien/gexe v0.2.0 // indirect
+	github.com/vladimirvivien/gexe v0.2.0
```
phisco, line 159: this should be moved to the require stanza above with all the other direct dependencies

.github/workflows/ci.yml
pedjak: I tried to avoid using `skopeo`, was thinking to use `kubectl port-forwarding`, and it did work on my laptop, but not on github, see previous runs: https://github.com/crossplane/crossplane/actions/runs/5427726428/jobs/9871266837#step:12:476  Looking around I found https://github.com/kubernetes/kubectl/issues/1169 - the issue is probably connected to kubectl version, but it would be really hard to debug it.

Unfortunately, we need to use https://github.com/containers/skopeo but it cannot be installed like helm or kind, it is distributed only through packages. BTW, I discovered that `scopeo` is already available in that ubuntu image and we could actually skip this step, like we do not install `kubectl` at all.

The reason for using `skopeo` is the test setup we need to achieve for testing the new functionality end-to-end:
* we need to deploy a registry with custom TLS certificate, with dns name that could be later accessible by crossplane
* we need to copy fn image to such registry from host where it got built. Such copy operation needs to ignore TLS certificate and establish the connection anyway. `docker push` does not have such flag, it would require to reconfigure the docker engine daemon, and such operation is even less portable between operating systems and docker installations.
* Hence it was needed to take `skopeo` - actually is a great tool for talking to registries and more.

All above was the reason why I initially started with the integration tests - they were running against local docker engine and in that case docker can push images without checking certificate. Also, the test setup was much simpler.

test/e2e/xfn_test.go
```diff
@@ -0,0 +1,318 @@
...
+		t.Logf("skopeo stdout: %s", string(out))
+		if p.ExitCode() != 0 {
+			t.Fatalf("copying image to registry not successful, exit code %v std out %v std err %v", p.ExitCode(), string(out), p.Err())
+		}
+		return ctx
+	}
+}
+
+// InstallDockerRegistry with custom TLS
+func InstallDockerRegistry() features.Func {
```
phisco, line 229: Down the line we could also think to move this to the Setup in TestMain as I can see it useful for many tests, is this already exposing both http and https at the same time or is it forcefully redirecting to https?

.github/workflows/ci.yml
phisco: But we need to run it also locally. Maybe the pull/push could be done using go-containerregistry directly with code?

.github/workflows/ci.yml
```diff
@@ -297,7 +300,17 @@ jobs:
...
+      - name: show logs
+        run: |
+          kubectl logs -l app=crossplane --all-containers=true -n crossplane-system
+          kubectl get deployments -o yaml -n crossplane-system
+          echo PODS
+          kubectl get pods -o yaml -n crossplane-system
+          kubectl get configmap -n crossplane-system
+          helm get values crossplane  -n crossplane-system
+
```
phisco, line 313: There is a function to export the logs of the kind cluster here, maybe we could export them on failures before destroying the cluster: https://github.com/kubernetes-sigs/e2e-framework/blob/main/pkg/envfuncs/kind_funcs.go

Makefile
```diff
@@ -116,10 +116,16 @@ cobertura:
 		grep -v zz_generated.deepcopy | \
 		$(GOCOVER_COBERTURA) > $(GO_TEST_OUTPUT)/cobertura-coverage.xml
 
-e2e-tag-images:
+e2e.test.images:
+	@$(INFO) Building E2E test images
+	@docker build --load -t $(BUILD_REGISTRY)/fn-labelizer-$(TARGETARCH) test/e2e/testdata/images/labelizer
```
phisco, line 121: We can push this somewhere and just take it from there as we are doing for packages.

pedjak: These images are used only for running e2e tests, they do not have any other purpose, or extra value. IMHO, I do not see much value in keeping it in an external registry, please correct me if you see it differently. At least I would start like this, and if we see the value later, we can always then decide to push it somewhere.

.github/workflows/ci.yml
pedjak: The added tests is failing at the moment at Github, and I added this just to be able to debug/see what is going on. I am going to remove it once the test starts to pass, and we can focus on logs gathering in an separate issue.

Makefile
phisco: I was just suggesting to start the other way around as that looks easier to me and we are already doing that for other tests. In a separate issue we could agree on the best approach and handle them all the same way instead of building this one here, while having the packages built and pushed beforehand, see the image used here: https://github.com/crossplane/crossplane/blob/master/test/e2e/manifests/pkg/configuration/dependency/configuration.yaml

pedjak: True, however, provider-nop is used for other things like demo - this particular image was crafted sollely for purpose of asserting that pulling from registry works.. and the function just add a label to each MR. IMHO, I do not see a value of publishing such image to a registry, it should be more generic and configurable for that.

.github/workflows/ci.yml
pedjak: btw, found why the test is failing - crossplane cannot be upgraded because of resource limitations: https://github.com/crossplane/crossplane/actions/runs/5430979412/jobs/9877067067?pr=4261#step:14:729

```
0/1 nodes are available: 1 Insufficient cpu. preemption: 0/1 nodes
        are available: 1 No preemption victims found for incoming pod..'
```

Need to see how to overcome this.

Makefile
pedjak: > https://github.com/crossplane/crossplane/blob/master/test/e2e/manifests/pkg/configuration/dependency/configuration.yaml

@phisco  where does the source repo live for e2e-depends-on-provider-nop exist?

phisco: `crossplane/e2e-depends-on-provider-nop:v0.1.0` doesn't look like a provider we are expecting other people to use...

.github/workflows/ci.yml
phisco: We are upgrading crossplane successfully in the composition validation test, maybe have a look at that, couldn't spot a difference, but if that's the case you should be able to reproduce it locally

Makefile
pedjak: and for me it would be a reason more not to publish it to a public registry as Docker hub. Pulling it from there could be sometimes problematic, given the known Dockerhub rate limiting. I would propose that we avoid publishing the image prematurely now in this PR - let it bake for some time, and if we see the value of consuming it in a context different from e2e testing, we could move it in a separate repo, and introduce workflows and publishing logic. If we wanna iterate fast, keeping the image here is beneficial for now.

phisco: Yes, but I'd rather have a single approach for now, so given that we are already building and pushing the other image separately, I'd keep it simple and use the same approach here, postponing a better decision to another issue.

pedjak: > Yes, but I'd rather have a single approach for now, so given that we are already building and pushing the other image separately, I'd keep it simple and use the same approach here, postponing a better decision to another issue.

Moving test images in a separate repo is going to establish a separate lifecycle for them. Everytime we need to tweak them for some tests, we are going to have two PRs instead of one and it might happen that we need to merge first image PR, and then to realize that the change was not right for e2e tests, so we need to do another rounds of PR. Here, everything is kept within single PR. @negz would like to hear your 2 cents on this subject?

phisco: Just to be clear, I'm not suggesting moving them to a separate repo, but more like, stick to the status quo and open another issue to solve it organically for all images used by e2es. E.g. we could think of having a dedicated directory `test/e2e/images` with subfolders for all images uses and build and push them all through a separate pipeline or step in the ci pipeline

pedjak: I do understand your intention @phisco but given that we just started adding e2e tests, not sure that we have any status quo yet. It would be great if we could treat lifecycle of all used images in the same way, but I rather see it as a graduation process:

* an image gets introduced within this repo due to a necessity to assert some crossplane functionality. 
* IMHO, we should always try to use a public image as possible, but we are going to have often situation that none of the images is specific enough for the given usecase (e.g. if we need to test some error case, we should craft an image that produces such error)
* If we see that the image is actually useful outside of e2e testing context, then we can promote it and move it to its own repo and publish in a public registry.

> we could think of having a dedicated directory test/e2e/images with subfolders for all images uses and build and push them all through a separate pipeline or step in the ci pipeline

This PR introduces `test/e2e/testdata/images` dir.

test/e2e/xfn_test.go
```diff
@@ -0,0 +1,320 @@
...
+	return funcs.AllOf(
+		funcs.AsFeaturesFunc(funcs.HelmUpgrade(
+			HelmOptions(
+				helm.WithArgs(
+					"--set args={--debug,--enable-composition-functions}",
+					"--set xfn.enabled=true",
+					"--set xfn.args={--debug}",
+					"--set registryCaBundleConfig.name=reg-ca",
+					"--set registryCaBundleConfig.key=domain.crt",
+					"--set xfn.resources.requests.cpu=100m",
```
pedjak, line 253: The e2e test was passing on my machine, but once the PR got created, the test was continuously failing on Github runner. After some debugging, I could see that the root cause was the inability to deploy crossplane with xfn enabled,  saying that none of nodes can fulfill the requested cpu resources. Further digging led me to the default resource requests and limit for `crossplane-xfn` container:

```yaml
  resources:
    limits:
      # -- CPU resource limits for the Composite function runner container.
      cpu: 2000m
      # -- Memory resource limits for the Composite function runner container.
      memory: 2Gi
    requests:
      # -- CPU resource requests for the Composite function runner container.
      cpu: 1000m
      # -- Memory resource requests for the Composite function runner container.
      memory: 1Gi

```

These are too high for Github runner (2 cpus). After lowering cpu requests at Helm chart upgrade, everything started to work again.

The default values got introduced in https://github.com/crossplane/crossplane/commit/0e210698c72fdd1b5cef16718a145d7eb09eeb08 by @negz as a part of https://github.com/crossplane/crossplane/pull/3465 - I could not find any details on the set resource values in that PR, or [the design document](https://github.com/crossplane/crossplane/blob/b9e4773dc431d974a44232b95500e6f815a9b8ac/design/design-doc-composition-functions.md) but I assume that there was a good reason for choosing them @negz ?

test/e2e/xfn_test.go
pedjak: [Docker registry](https://docs.docker.com/registry/deploying/) can either talk http or https but not both. For the purpose of this tests we needed to configure to listen for https trafic, using a custom TLS certificate.

I would rather keep the main setup as light as possible - if a test does not pull anything from a private registry, then we can avoid its deploying, shortening the testing time and resource usage.

Ofcourse, if we notice that some other tests need registry, we could then define a number crossplane deployments (stable features only, with beta features, with alpha features) and run the tests accordingly.

test/e2e/funcs/feature.go
pedjak: well, I have noticed that with each invocation of `envfuncs.CreateNamespace`, default namespace is set to that newly created namespace. Not sure if this is going to cause any issues in the feature, but I am gonna remove the added function.

test/e2e/xfn_test.go
pedjak: Yeah, the test name function is to generic - I have another proposal - we are not testing here functions at all, rather xfn runner. Hence, I would like to rename `TestFunction` to `TestXfnRunner`.

test/e2e/funcs/feature.go
phisco: That's a valid reason indeed, I'd say to switch to a dedicated function

test/e2e/xfn_test.go
phisco: I would be more specific given that we are going for a 1:1 relationship between features and Test functions, we could say the Test function name should be the same as the feature name.

go.mod
phisco: Not sure, but it was doing the right thing for me when i removed them and just run `go mod vendor`

test/e2e/xfn_test.go
phisco: Totally agree

Makefile
phisco: Fine, as you want, but please create an issue about properly handling images used by E2Es and a TODO here referencing it.

test/e2e/xfn_test.go
```diff
@@ -0,0 +1,344 @@
...
+		}
+		if addr == "" {
+			t.Fatalf("no nodes with private address")
+		}
+
+		srcRef, err := name.ParseReference("crossplane-e2e/fn-labelizer:latest")
+		if err != nil {
+			t.Fatal(err)
+		}
+		src, err := daemon.Image(srcRef)
```
phisco, line 229: Can't you instead use crane.Copy?

phisco: Ah, I guess not because the image is only built locally. I was suggesting that not to have to rely on the docker daemon directly

pedjak: > Ah, I guess not because the image is only built locally. I was suggesting that not to have to rely on the docker daemon directly

Correct, `crane.Copy()` can copy only between two remote repos.

test/e2e/xfn_test.go
pedjak: how about that we then call the function as `TestXfnRunnerImagePull`?

test/e2e/xfn_test.go
negz: @pedjak I believe the rationale was that we'd want enough CPU/compute for folks to be able to run at least a small handful of containers at once. For E2E tests I think lower limits/requests should be fine for the time being. (As you know) I'm planning to open an updated Functions design that should make this much less of an issue.

pedjak:
needed to rebase after #4270 got merged.

test/e2e/xfn_test.go
```diff
@@ -0,0 +1,344 @@
...
+					"--set xfn.resources.limits.cpu=100m",
+				),
+				helm.WithWait())...)),
+		funcs.ReadyToTestWithin(1*time.Minute, namespace),
+	)
+}
+
+// ProvideNopDeployed assets that provider-nop is deployed and healthy
+func ProvideNopDeployed() features.Func {
+	manifests := "test/e2e/manifests/xfnrunner/private-registry/pull/prerequisites"
```
phisco, line 287: can you remove this function and unwrap it directly in the only place where it's called above? Please avoid hardcoding values such as manifests in functions as it could be misleading and lead to hard to debug errors.

test/e2e/xfn_test.go
```diff
@@ -0,0 +1,344 @@
...
+		}
+		if err := client.Create(ctx, configMap); err != nil {
+			t.Fatalf("Cannot create config %s: %v", configMap.Name, err)
+		}
+		return ctx
+	}
+}
+
+// CopyImagesToRegistry copies fn images to private registry
+func CopyImagesToRegistry() features.Func {
```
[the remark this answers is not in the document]
pedjak: This function has only sense within this particular test and there is no intention to make it generic until there is a usecase that would need it. I would prefer to keep it in this PR on this level, once we are going to have a need we could/should refactor it.

test/e2e/xfn_test.go
```diff
@@ -0,0 +1,344 @@
...
+			t.Fatal(err)
+		}
+		u.SetGroupVersionKind(gv.WithKind(kind))
+		if err := client.Get(ctx, cr.ObjectKey{Name: name, Namespace: namespace}, u); err != nil {
+			t.Fatal("cannot get claim", err)
+		}
+		return u
+	}
+}
+func resourceValue(t *testing.T, u *unstructured.Unstructured, path ...string) map[string]string {
```
phisco, line 95: Nit: but would be awesome if we could make all these private functions somewhat reusable and move to funcs if we feel it would be worth it, but not a blocker

test/e2e/xfn_test.go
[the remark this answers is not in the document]
phisco: I could see it being reused by other tests pretty soon, depending on how we decide to handle images in #4257 and https://github.com/crossplane/crossplane/issues/4294. Nonetheless, then it should be at least private and mention the fact that it's only specific to this test in the signature or at least in a comment.

test/e2e/xfn_test.go
```diff
@@ -0,0 +1,344 @@
...
+		if err != nil {
+			t.Fatal("copying image to registry not successful", err)
+		}
+		return ctx
+	}
+}
+
+// InstallDockerRegistry with custom TLS
+func InstallDockerRegistry() features.Func {
+	return funcs.AllOf(
```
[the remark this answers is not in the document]
pedjak: sure, please check out the latest changes.

phisco (approved):
I think we should iterate further on the e2e, but I don't think it should be a blocker to the feature in this PR, so LGTM.

test/e2e/xfn_test.go
```diff
@@ -0,0 +1,207 @@
...
+						}
+						return true, nil
+					}, wait.WithTimeout(5*time.Minute))
+					if err != nil {
+						t.Fatalf("Expected label %v value to be true", labelName)
+					}
+
+				}
+				return ctx
+			}).
```
phisco, line 204: @pedjak, just noticed it's actually missing the teardown at the end, can you handle that before getting it merged?

pedjak: added

pedjak:
> I think we should iterate further on the e2e, but I don't think it should be a blocker to the feature in this PR, so LGTM.

The e2e test code got heavily adjusted based on your review - thanks! Feel free to open any further issue if you see some space for improvements.

BTW, moving from the initial integration tests, we lost ability to test the error cases. For example, if xfn runner cannot pull images due to bad or missing certificates, we are not able to detect it with an e2e test.

phisco (approved):
the e2e doesn't work out of the box on docker for mac due to the service node port used to access to the registry. It runs in CI, so as I was saying we can rework the e2e after merging, no need to stop the feature for this.

PR #4267, opened 2023-06-26
Helm Chart: add locks to the RBAC for pkg.crossplane.io

darkmuggle:

In the Upbound managed environment, users are not able to upgrade from the monolithic providers to SSOPs following [1] due to missing permissions to the lock. This changes add `locks.pkg.crossplane.io` to the RBACs for admin, view and edit.

[1] https://docs.upbound.io/knowledge-base/migrate-to-provider-families

I have:

- [x] ~Added or updated unit **and** E2E tests for my change.~

### DIFF
--- a/cluster/charts/crossplane/templates/rbac-manager-managed-clusterroles.yaml
+++ b/cluster/charts/crossplane/templates/rbac-manager-managed-clusterroles.yaml
@@ -101,7 +101,7 @@ rules:
   verbs: ["*"]
 - apiGroups:
   - pkg.crossplane.io
-  resources: [providers, configurations, providerrevisions, configurationrevisions]
+  resources: [locks, providers, configurations, providerrevisions, configurationrevisions]
   verbs: ["*"]
 # Crossplane administrators have access to view CRDs in order to debug XRDs.
 - apiGroups: [apiextensions.k8s.io]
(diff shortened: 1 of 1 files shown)


phisco (approved):

cluster/charts/crossplane/templates/rbac-manager-managed-clusterroles.yaml
```diff
@@ -101,7 +101,7 @@ rules:
   verbs: ["*"]
 - apiGroups:
   - pkg.crossplane.io
-  resources: [providers, configurations, providerrevisions, configurationrevisions]
+  resources: [locks, providers, configurations, providerrevisions, configurationrevisions]
```
jbw976, line 104: any ideas on the historical context for why `locks` wasn't included? was that oversight or any evidence that it was intentional?

jbw976 (commented):
I approved the CI/workflow runs so they'll execute on this PR now :)

cluster/charts/crossplane/templates/rbac-manager-managed-clusterroles.yaml
darkmuggle: I can only speak to the git history: the definition comes b5f01da6aaf786c0f7a6161e6e8c38796107d7f2 which was November 2020. On Nov 25, 2022, cc9bad7537597547005541a713e0a80ebd6c533c introduced the Lock API. @negz may not know, but my git archeology attempt didn't yield an answer to this question. That said, those who might need to do lock surgery like have cluster admin (which is not the case for the managed control planes).

negz: My guess is probably oversight. It's not something we think about much. It's mostly Crossplane machinery - humans reading and writing it is an edge case.

PR #4278, opened 2023-06-28
Add registry option for xfn

### LINKED ISSUE #4116: XFN container needs a default registry to pull images from
Composition functions require the image registry to be specified in the Composition along with the repository and tag, unlike Provider and Configuration `package` specifications which only require the repository and tag.  This makes portability of Compositions a lot more difficult when different environments use different registries.

The crossplane container accepts a `-r registry` command line option to allow for a default registry to pull packages from.  It defaults to `index.docker.io` but can be overridden to pull from a specific registry by default.  That allows Provider and Configuration packages to specify only the repository and tag information for an image, and they are registry-independent.

Provide support for a configurable default image registry to pull Composition Function images from so that the registry does not need to be specified in the Composition.

bobh66:
Added the `--registry` option to the `xfn` commands to allow a default registry to be specified for pulling containers in Composition Function processing.
 
This capability is enabled by adding the `--registry=myregistry.mydomain.com` argument to the `xfn.args` array in the helm chart.  That causes the `crossplane-xfn` container to run `xfn start` with the `--registry` argument, which starts a container runner with the default registry.

The container runner then calls `xfn spark` with the `--registry` option.

The user's specified container is checked to see if it has a valid registry component in the name, and if it does not the default registry is used.  The default registry is `index.docker.io`.

Fixes #4116 

I have:

~- [ ] Added or updated unit **and** E2E tests for my change.~

### DIFF
--- a/cmd/xfn/main.go
+++ b/cmd/xfn/main.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 
 	"github.com/alecthomas/kong"
+	"github.com/google/go-containerregistry/pkg/name"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 
 	"github.com/crossplane/crossplane-runtime/pkg/logging"
@@ -34,10 +35,17 @@ import (
 type debugFlag bool
 type versionFlag bool
 
+// KongVars represent the kong variables associated with the CLI parser
+// required for the Registry default variable interpolation.
+var KongVars = kong.Vars{
+	"default_registry": name.DefaultRegistry,
+}
+
 var cli struct {
 	Debug debugFlag `short:"d" help:"Print verbose logging statements."`
 
-	Version versionFlag `short:"v" help:"Print version and quit."`
+	Version  versionFlag `short:"v" help:"Print version and quit."`
+	Registry string      `short:"r" help:"Default registry used to fetch containers when not specified in tag." default:"${default_registry}" env:"REGISTRY"`
 
 	Start start.Command `cmd:"" help:"Start listening for Composition Function runs over gRPC." default:"1"`
 	Run   run.Command   `cmd:"" help:"Run a Composition Function."`
@@ -69,6 +77,7 @@ func main() {
 		kong.Description("Crossplane Composition Functions."),
 		kong.BindTo(logging.NewLogrLogger(zl), (*logging.Logger)(nil)),
 		kong.UsageOnError(),
+		KongVars,
 	)
-	ctx.FatalIfErrorf(ctx.Run())
+	ctx.FatalIfErrorf(ctx.Run(&start.Args{Registry: cli.Registry}))
 }
--- a/internal/controller/apiextensions/composite/composition_ptf.go
+++ b/internal/controller/apiextensions/composite/composition_ptf.go
@@ -649,7 +649,7 @@ type ContainerFunctionRunnerOption func(ctx context.Context, fn *v1.ContainerFun
 // 2. Loads credentials from the supplied service account's image pull secrets.
 // 3. Loads credentials from the function's image pull secrets.
 // 4. Loads credentials using the GKE, EKS, or AKS credentials helper.
-func WithKubernetesAuthentication(c client.Reader, namespace, serviceAccount string, registry string) ContainerFunctionRunnerOption {
+func WithKubernetesAuthentication(c client.Reader, namespace, serviceAccount, registry string) ContainerFunctionRunnerOption {
 	return func(ctx context.Context, fn *v1.ContainerFunction, r *fnv1alpha1.RunFunctionRequest) error {
 
 		sa := &corev1.ServiceAccount{}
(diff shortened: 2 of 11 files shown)


cmd/xfn/main.go
```diff
@@ -69,6 +77,7 @@ func main() {
 		kong.Description("Crossplane Composition Functions."),
 		kong.BindTo(logging.NewLogrLogger(zl), (*logging.Logger)(nil)),
 		kong.UsageOnError(),
+		KongVars,
 	)
-	ctx.FatalIfErrorf(ctx.Run())
+	ctx.FatalIfErrorf(ctx.Run(&start.Args{Registry: cli.Registry}))
```
negz, line 82: This pattern of passing arguments down to a child command is a little awkward, but AFAICT it's the right way to do this in kong.

I wonder whether we should define the other common flags - e.g. `MapRootUID` - this way too.

internal/controller/apiextensions/composite/composition_ptf.go
```diff
@@ -649,7 +649,7 @@ type ContainerFunctionRunnerOption func(ctx context.Context, fn *v1.ContainerFun
 // 2. Loads credentials from the supplied service account's image pull secrets.
 // 3. Loads credentials from the function's image pull secrets.
 // 4. Loads credentials using the GKE, EKS, or AKS credentials helper.
-func WithKubernetesAuthentication(c client.Reader, namespace, serviceAccount string) ContainerFunctionRunnerOption {
+func WithKubernetesAuthentication(c client.Reader, namespace, serviceAccount string, registry string) ContainerFunctionRunnerOption {
```
negz, line 652:
```suggestion
func WithKubernetesAuthentication(c client.Reader, namespace, serviceAccount, registry string) ContainerFunctionRunnerOption {
```

Nit: In Go you can "dedupe" the type when you have several arguments of the same type in a row. 🙂

internal/xfn/container_linux.go
```diff
@@ -95,10 +95,10 @@ func (r *ContainerRunner) RunFunction(ctx context.Context, req *v1alpha1.RunFunc
 
 		Therefore we execute a shim - xfn spark - in a new user and mount
 		namespace. spark fetches and caches the image, creates an OCI runtime
-		bundle, then then executes an OCI runtime in order to actually execute
+		bundle, then executes an OCI runtime in order to actually execute
 		the function.
 	*/
-	cmd := exec.CommandContext(ctx, os.Args[0], spark, "--cache-dir="+r.cache, fmt.Sprintf("--max-stdio-bytes=%d", MaxStdioBytes)) //nolint:gosec // We're intentionally executing with variable input.
+	cmd := exec.CommandContext(ctx, os.Args[0], spark, "--cache-dir="+r.cache, "--registry="+r.registry, fmt.Sprintf("--max-stdio-bytes=%d", MaxStdioBytes)) //nolint:gosec // We're intentionally executing with variable input.
```
negz, line 101: This is getting long enough that we might want to wrap it for readability.

negz (commented):
This looks great. Do you feel up to attempting to write an E2E test to test it? 😉 Probably a tall ask at the moment since it would be the first test of Functions.

bobh66:
> This looks great. Do you feel up to attempting to write an E2E test to test it? 😉 Probably a tall ask at the moment since it would be the first test of Functions.

I can give it a shot - it looks like adding it to the existing test structure should be pretty straightforward.  The only real question is what to use for a container function?  I can have one set of tests that validates a filly qualified container image path and another that verifies the default works.

cmd/xfn/main.go
bobh66: I thought this was odd too but I couldn't find a better implementation.

negz:
> The only real question is what to use for a container function? I can have one set of tests that validates a filly qualified container image path and another that verifies the default works.


Maybe just https://github.com/negz/xfns/tree/main/xfn-nop since it doesn't really matter what the Function does, only that we pull it from the right place? I have a TODO to find a home for all the images we use in tests per https://github.com/crossplane/crossplane/issues/4257.

pedjak:
@bobh66 I am also writing e2e test for https://github.com/crossplane/crossplane/pull/4261, and I crafted a function that just adds an extra label. IMHO, the function needs to modify somehow MRs, so that we know that it got pulled and executed. Otherwise, if a function does nothing, just by observing MRs we cannot assert if pulling works as expected.

bobh66:
> @bobh66 I am also writing e2e test for #4261, and I crafted a function that just adds an extra label.
Thanks @pedjak - I think it makes sense to build this test on top of yours.  The main differences will be deploying the Crossplane containers with the `--registry` option and omitting the container registry in the Composition Function container specification.  The results of the test should be identical to the original tests.

@negz if it's OK with you I'll defer the e2e test until #4261 is merged and build on top of those tests.

internal/controller/apiextensions/composite/composition_ptf.go
negz: Can you fix this up and then I think I'm ready to merge.

negz (approved):

pedjak:
> Thanks @pedjak - I think it makes sense to build this test on top of yours. The main differences will be deploying the Crossplane containers with the --registry option and omitting the container registry in the Composition Function container specification. The results of the test should be identical to the original tests.

@bobh66 - I have added e2e tests in https://github.com/crossplane/crossplane/pull/4261 - but they are failing on github, due to some issue on copying images to private registry.. the tests are passing on my laptop, hence I need to check what is going on here.

pedjak:
@bobh66 tests are green now.

PR #4292, opened 2023-07-03
feat(xrd): Add custom composition selector

### LINKED ISSUE #4281: Custom Composition Selectors

Currently, Crossplane supports the user to choose a composition through a `spec.compositionSelector` in the XRC. This works fine as long as you know exactly which labels you need to specify which requires you to have deep platform insights to know what compositions exist and which one to choose.

However, in our scenario users don't have access to compositions or other kind of cluster-scoped resource. Naturally they don't know which composition to choose from. This requires the platform engineers to write a distinct section in the documentation to explain which labels to specify.

Since `spec.compositionSelector.matchLabels` is a schemaless `map[string]string` it allows users to enter any kind of value possibly bearing the risk of typos due to the lack of input validation (say, `stgae` instead of `stage`).

Crossplane should provide a way to make composition selection transparent to the user by letting the platform engineers choose the labels based on the claim.

XRDs could be extended with a `spec.customCompositionSelector` that accepts a set of field paths from which the label values are extracted:

```yaml
---
metadata:
  name: tests.example.com
spec:
  # Overwrites `spec.compositionSelector` in the claim
  customCompositionSelector:
    matchLabels:
      - key: stage
        value:
          fromFieldPath: spec.stage
          policy:
            resolution: Optional
  versions:
    schema:
      openAPIV3Schema:
        properties:
          spec:
            properties:
              stage:
                description: The environment in which the resource should be deployed.
                type: string
                enum:
                  - production
                  - testing
```

MisterMX:

Add a new field `spec.customCompositionSelector` that allows the XRD writer to define custom paths where to draw the label selectors for the composition from and overrite what is specified in the XR.

Fixes #4281

I have:


apis/apiextensions/v1/xrd_types.go
```diff
@@ -95,6 +95,85 @@ type CompositeResourceDefinitionSpec struct {
...
+// loading a label value.
+type CustomCompositionSelectorLabelValuePolicy struct {
+	// FromFieldPath specifies how to read from a field path.
+	// The default is "Required" which means the selection will fail if the
+	// path is not specified.
+	// Use 'Optional' if the selection should be skipped if the specified path
+	// does not exist.
+	// +kubebuilder:validation:Enum=Optional;Required
+	// +optional
+	FromFieldPath *FromFieldPathPolicy `json:"fromFieldPath,omitempty"`
```
turkenh, line 176: If I could follow correctly, this would look like:

```yaml
  customCompositionSelector:
    matchLabels:
      - key: stage
        value:
          fromFieldPath: spec.stage
          policy:
            fromFieldPath: Required # <-- should we use 
    policy:
      resolve: IfNotPresent
```

I would suggest naming it as `Resolution` also to be consistent with [runtime](https://github.com/crossplane/crossplane-runtime/blob/0d8cbcef6fac5299f2ce724f3bed0dec2f72f444/apis/common/v1/policies.go#L114) policies (also we could re-use `xpv1.ResolutionPolicy`), which would make it: 

```yaml
  customCompositionSelector:
    matchLabels:
      - key: stage
        value:
          fromFieldPath: spec.stage
          policy:
            resolution: Required # <-- should we use 
    policy:
      resolve: IfNotPresent
```

Please also add kubebuilder default marker.

internal/controller/apiextensions/composite/api.go
```diff
@@ -265,6 +289,46 @@ func (r *APILabelSelectorResolver) SelectComposition(ctx context.Context, cp res
...
+	cp.SetCompositionSelector(customSelector)
+	return nil
+}
+
+func (r *APILabelSelectorResolver) resolveCustomCompositionSelectorLabel(cp resource.Composite, label v1.CustomCompositionSelectorLabel) (string, string, error) {
+	cpMap, err := runtime.DefaultUnstructuredConverter.ToUnstructured(cp)
+	if err != nil {
+		return "", "", err
+	}
+	value, err := fieldpath.Pave(cpMap).GetString(label.Value.FromFieldPath)
```
turkenh, line 322:
```suggestion
	paved, err := fieldpath.PaveObject(cp)
	if err != nil {
		return "", "", errors.Wrap(err, errPaveComposite)
	}
	value, err := paved.GetString(label.Value.FromFieldPath)
```

internal/controller/apiextensions/composite/api.go
```diff
@@ -265,6 +289,46 @@ func (r *APILabelSelectorResolver) SelectComposition(ctx context.Context, cp res
...
+	if err != nil {
+		if fieldpath.IsNotFound(err) && label.Value.GetFromFieldPathPolicy() == v1.FromFieldPathPolicyOptional {
+			return "", "", nil
+		}
+		return "", "", err
+	}
+	return label.Key, value, nil
```
turkenh, line 329: nit: code style/return early.

```suggestion
	if fieldpath.IsNotFound(err) && label.Value.GetFromFieldPathPolicy() == v1.FromFieldPathPolicyOptional {
		return "", "", nil
	}
	if err != nil {
		return "", "", errors.Wrap(err, errGetRequiredValueFromFieldPath)
	}
	return label.Key, value, nil
```

apis/apiextensions/v1/xrd_types.go
```diff
@@ -95,6 +95,85 @@ type CompositeResourceDefinitionSpec struct {
...
 	// +optional
 	Conversion *extv1.CustomResourceConversion `json:"conversion,omitempty"`
+
+	// CustomCompositionSelector provides a way to override the composition
+	// selector defined in the XR / XRC by generating the label matcher
+	// from arbitrary fields in the claim.
+	// It allows making composition selection transparent to the user who might
+	// not know which composition to select.
+	// +optional
+	CustomCompositionSelector *CustomCompositionSelector `json:"customCompositionSelector,omitempty"`
```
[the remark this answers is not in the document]
turkenh: Hmm, actually, it depends on policy I guess:

if `resolve: Always`, it works like (a).
if `resolve: IfNotPresent`, it works like (b).

Wondering if there is a way to stay consistent by implementing both `EnforcedCompositionSelector` and `DefaultCompositionSelector` 🤔

Issue #3751, opened 2023-02-08
Promote Composition Functions to beta


negz:
### What problem are you facing?

We added alpha Composition Function support to Crossplane v1.11 in https://github.com/crossplane/crossplane/pull/3465. Eventually we need to either drop support for the feature, or promote it to beta. This issue tracks any work that would prevent us promoting the feature to beta.

Since alpha, we've tweaked the design per https://github.com/crossplane/crossplane/pull/4306. With that design in mind, I believe we'll need to do the following to enable the base Function functionality.

```[tasklist]
### Tasks
- [ ] https://github.com/crossplane/crossplane/issues/4418
- [ ] https://github.com/crossplane/crossplane/issues/4419
- [ ] https://github.com/crossplane/crossplane/issues/4432
- [ ] https://github.com/crossplane/crossplane/issues/4339
- [x] Implement encryption and authentication for gRPC communications with the function runner.
- [ ] https://github.com/crossplane/crossplane/issues/3896
- [ ] https://github.com/crossplane/crossplane/issues/4288
- [ ] https://github.com/crossplane/crossplane/issues/4191
- [ ] https://github.com/crossplane/crossplane/issues/3812
- [ ] https://github.com/crossplane/crossplane/issues/4558
```

We'll also need to do the following things, which we might want to move into a more developer-experience focused tracking issue:

```[tasklist]
- [ ] https://github.com/crossplane/docs/issues/579
- [x] Implement a dummy Function (e.g. function-nop) for testing.
- [ ] https://github.com/crossplane/crossplane/issues/4574
- [ ] https://github.com/crossplane/crossplane/issues/4571
- [ ] https://github.com/crossplane/crossplane/issues/4573
- [ ] https://github.com/crossplane/crossplane/issues/4576
```

### How could Crossplane help solve your problem?

If and when we feel ready, we should promote Composition Functions to beta. To do so we should:

* Move the feature flag to the "beta features" group, and have it be on by default.
* Update the "alpha feature" warnings on the `Composition` type to beta feature warnings.
* Bump the relevant protobuf messages (`RunFunctionRequest` etc) to v1beta1

negz:
I feel like promoting these to beta in 1.12 is within the realm of possibility, but I'm not ready to commit to that just yet. I'd like to take at least another month to gather feedback.

jpoley:
please add in https://github.com/crossplane/crossplane/issues/3899

negz:
[...]

Issue #3197, opened 2022-07-19
XRD spec.versions[0].schema.openAPIV3Schema is weakly typed as an object instead of full openapi schema


gberche-orange:
### What problem are you facing?

As a crossplane XRD author, in order to write provide quality openapi schemas for my XRDs, I need to leverage IDE completion based on the XRD CRD schema definition.

Currently the XRD CRD is defining spec.versions[0].schema.openAPIV3Schema  as a plain object without detailing supported fields

https://github.com/crossplane/crossplane/blob/6494953ca4d5a56350fbe696d351f9a416aae48e/cluster/crds/apiextensions.crossplane.io_compositeresourcedefinitions.yaml#L273-L277

### How could Crossplane help solve your problem?

It might be possible for XRD CRD to instead reference the openapi v3 schema fields, possibly with a URI (see related specs at https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-00#section-5) , in order to get completion just like CRD authors experience natively, see the following example of editing a sample CRD in intellij:

![image](https://user-images.githubusercontent.com/4748380/179801851-4fd5676d-44e7-4d5c-b8f1-cd4a046b600f.png)

gberche-orange:
Actually, it seem the K8S CRD openapi schema is also currently defining top level fields which is insufficient to properly author a CRD schema
![image](https://user-images.githubusercontent.com/4748380/179807053-f2e7f871-c926-4959-8667-6a0183f6ae62.png)

MisterMX:
Defining a schema for `openAPIV3Schema` requires recursive schema definition which is currently not supported by Kubernetes at the moment. See https://github.com/kubernetes/kubernetes/issues/62872.

It works for CRDs because they are a built-int type.

negz:
Yeah, sadly this is a painful known issue. We've discussed writing a validation webhook for it since CRD machinery can't handle it internally - that seems like the most promising path forward.

gberche-orange:
Thanks @MisterMX and @negz for the follow up

> We've discussed writing a validation webhook for it since CRD machinery can't handle it internally - that seems like the most promising path forward.

I guess this would cover the case of rejecting invalid schemas in XRDs, but not yet improving the authoring experience by supporting coding assistance (completion, syntax check, inline documentation).

gberche-orange:
[...]

PR #4306, opened 2023-07-10
Composition Functions Beta: Long-running Functions

negz:

This pull request introduces an updated design for Composition Functions. Under this design Functions are long-running processes, like Providers. Crossplane makes RunFunctionRequest RPCs to a Function using gRPC.

This updated design also sketches out an improved Function development experience, as well as a few hypothetical 'generic' Functions. A generic Function is a Function that works with any XR. Generic Functions extend Crossplane with new ways to express Composition logic without actually needing to develop a Function of your own.

My intention is that if this design is approved, we'll implement it as `v1beta1` support for Functions in Crossplane v1.14 (due in October).

I have:


negz:
* https://github.com/crossplane/crossplane/issues/3751

If this design is accepted I'll need to update the plan laid out in the beta tracking issue. A lot of the issues it tracks are specific to the `xfn` Function runner, which under this design would become just one way to build a Function.

I believe the very high-level implementation plan for this design would be:

- [ ] Move `xfn` out of tree - e.g. to `crossplane/function-runtime-oci` - and remove it from the Helm chart
- [ ] Add `Function` support to the package manager
- [ ] Remove the `FunctionIO` type and update the `RunFunctionRequest` and `RunFunctionResponse` RPC types
- [ ] Refactor the `PTFComposer` to accommodate the design (which I think would be a simplification)
- [ ] Add the illustrated commands to `kubectl crossplane` (or a new CLI?)
- [ ] Start with one SDK (probably Go) and one generic Function (probably Go Templates).

design/design-doc-composition-functions.md
```diff
@@ -1,8 +1,8 @@
- # Composition Functions
+ # Composition Functions: Beta
 
-* Owners: Nic Cope (@negz), Sergen Yalçın (@sergenyalcin)
-* Reviewers: Crossplane Maintainers
-* Status: Draft
+* Owners: Nic Cope (@negz)
+* Reviewers: Hasan Turken (@turkenh), Jared Watts (@jbw976)
```
negz, line 4: I chose these reviewers because:

* @turkenh has context from reviewing the original, alpha design
* @jbw976 will be leading a new developer experience effort, which I feel relates closely to Functions

design/design-doc-composition-functions.md
```diff
@@ -175,604 +183,828 @@ metadata:
...
+}
+
+// A Resource represents the state of a resource.
+message Resource {
+  // The JSON representation of the resource.
+  google.protobuf.Struct resource = 1;
+
+  // The resource's connection details.
+  map<string, bytes> connection_details = 2;
+}
```
negz, line 336: Under the alpha design a Function can return a composed resource that specifies P&T style readiness checks and ways to derive connection details (e.g. `connectionSecretFromField` etc). I've removed that under this design. I don't think I've seen anyone use it, and I don't think we actually need it. We can always add it back if there's a demand.

It's possible for a Function to just derive and write whatever XR connection details it needs, and to set the XR's as Ready whenever it feels it is. I'm pretty sure @turkenh identified we didn't need this in the original design and I think I agreed but must have forgotten and built it anyway? See https://github.com/crossplane/crossplane/pull/2886#discussion_r991710579.

design/design-doc-composition-functions.md
```diff
@@ -175,604 +183,831 @@ metadata:
...
+package manager. Deploying Functions as Kubernetes Deployments (with Services,
+Service Accounts, etc) will no doubt necessitate something like a Provider's
+ControllerConfig type.
 
-### Using KRM Function Spec Compliant Functions
+We've identified that we [don't want to proceed][controllerconfig-deprecation]
+with ControllerConfig, but don't yet have a suitable alternative. Rather than
+propagating a ControllerConfig-like pattern I propose we prioritize finding an
+alternative. I intend to open a separate, simultaneous design to address this
+since it will affect Provider packages as well as Functions.
```
negz, line 978: The cliff-notes version of my impending proposal, which I hope will be a "good enough" alternative to a PRI (https://github.com/crossplane/crossplane/issues/2671):

Add a core Crossplane CLI flag like `--package-runtime` with two values - `Kubernetes` (the default) and `External`. When running in "Kubernetes" mode the package manager does what it does today - assumes it's running on a Kubernetes cluster and makes Deployments, Services etc to run Providers and Functions. When running in "External" mode the package manager revision controllers are disabled. e.g. When you create a `Provider` the package manager creates a `ProviderRevision` but doesn't reconcile it. It expects some external system (such as a controller running alongside Crossplane) to figure out what to do with the package revision.

The rationale here is that AFAIK today 99% of folks are running Crossplane in a Kubernetes cluster (or somewhere that quacks like a Kubernetes cluster - e.g. vcluster). It's really only advanced folks (i.e. vendors like Upbound) who might want to deploy extensions in a completely different way, so letting them write a controller to do that seems reasonable.

To solve the `ControllerConfig`-is-just-a-template-for-Deployment problem I propose we lean in and actually make it a template for `Deployment` (and friends - e.g. `Service`). I imagine this would look similar to a Deployment's pod spec template. I'd remove most of the hardcoded templating from the package manager codebase and push it into a `default` config (like the `default` `ProviderConfig` pattern). Given Functions aren't a controller I'd call this a `RuntimeConfig`.

CC @hasheddan

design/design-doc-composition-functions.md
```diff
@@ -99,74 +99,78 @@ experiences designing Kubernetes, its precursor Borg, and many generations of
...
-this document could serve as a test bed for new features in the contemporary
-'resources array' based form of Composition.
-
-The user experience around authoring and maintaining Composition Functions is
-out of scope for this proposal, which focuses only on adding foundational
-support for the feature to Crossplane. 
+* Let folks specify composition logic in their DSL or GPL of choice.
+* Make it easy to extend Crossplane with new ways to 'do composition'.
+* Decouple adding new ways to 'do composition' from the core release cycle.
+* Be possible to keep behind a feature flag until it is generally available.
```
turkenh, line 164: Should we have some details on the motivation for the updated design? What was wrong with the previous one, and why do we want to change it? 
And/or a list of lessons learned from the alpha.

design/design-doc-composition-functions.md
```diff
@@ -175,32 +179,557 @@ metadata:
...
-    type: Container
-    container:
-      image: xkpg.io/my-cool-function:0.1.0
+    # Each entry in the functions array has a name. This uniquely identifies
+    # the Function 'call' within the Composition.
+  - name: compose-xr-using-go-templates
+    # The functionRef tells the Composition which Function to use. This Function
+    # uses Go Templates (like Helm) to render composed resources.
+    functionRef:
+    - name: go-templates
```
turkenh, line 193: I believe this should not be an array.

design/design-doc-composition-functions.md
```diff
@@ -175,32 +179,557 @@ metadata:
...
+  # This Composition specifies functions instead of (P&T) resources.
+  # This is an array - you can specify a pipeline of Functions.
   functions:
-  - name: my-cool-function
-    type: Container
-    container:
-      image: xkpg.io/my-cool-function:0.1.0
+    # Each entry in the functions array has a name. This uniquely identifies
+    # the Function 'call' within the Composition.
+  - name: compose-xr-using-go-templates
```
turkenh, line 189: I don't feel great about having `name` here and under `functionRef`, i.e. `functions[*].name` vs ``functions[*].functionRef.name`. It is not obvious that the first one is the name of the Function 'call' as commented above.

Wondering if we can improve this by using something else, like, `step`, `call`, `action`, etc., instead of `name`.

turkenh: Hmm, indeed we might even consider renaming `functions:` to something else, like `pipeline:`

```yaml
  pipeline:
  - step: compose-xr-using-go-templates
    functionRef:
      name: go-templates
    config:
      apiVersion: example.org/v1
      kind: GoTemplate
      source: Remote
      remote: git://github.com/example/my-xpostgresql-go-templates
```

design/design-doc-composition-functions.md
```diff
@@ -209,570 +738,353 @@ metadata:
...
-      value: admin
+```shell
+$ kubectl crossplane composition render composition.yaml xr.yaml
+---
+# YAML stream omitted for brevity.
 ```
 
-A function wishing to delete a composed resource may do so by setting its
-`resource` to null, for example:
+Assuming a 'pure' Composition consisting only of Functions with no P&T
```
turkenh, line 877: > Assuming a 'pure' Composition consisting only of Functions with no P&T resources, 

Could we end up with P&T transform logic becoming yet another Function at some point and `spec.resources` disappearing with v2 of the Compositions :)? In this case, we wouldn't need assumptions like this and provide a more wholistic experience.

design/design-doc-composition-functions.md
```diff
@@ -209,570 +738,353 @@ metadata:
...
+resources, this command would need only to iterate through the Composition's
+`functions` array, and for each Function:
+
+1. Pull and start it.
+2. Form a `RunFunctionRequest` from the XR and any prior `RunFunctionResponse`.
+3. Print the `desired` block of final `RunFunctionResponse` as a YAML stream.
+
+This isn't quite an end-to-end test of Composition. The composed resources are
+not actually created and reconciled with an external system, so its not possible
+for example to derive XR status from composed resource status. It's also not
```
turkenh, line 887: > The composed resources are not actually created and reconciled with an external system, so its not possible for example to derive XR status from composed resource status

Wondering if this could be mocked by providing an `observed.yaml` as yet another input to the render command.

design/design-doc-composition-functions.md
turkenh: Nvm, just saw the `Patch-and-Transform as a Function` section :)

turkenh (commented):
Looking great, I like the direction we are heading toward!
Especially the possibility of converging to the same machinery with the _Patch-and-Transform as a Function._ instead of treating them as two different ways to compose resources.

design/design-doc-composition-functions.md
```diff
@@ -175,32 +179,557 @@ metadata:
...
+Successfully built 202dc6e5df4c
+
+# Save your OCI image to a tarball.
+$ docker save | gzip > function.tgz
+
+# Add some package metadata for your Function.
+$ vim crossplane.yaml
+
+# Build a Function from the tarball.
+$ kubectl crossplane function build --from-oci-tarball=function.tgz
```
jbw976, line 693: is there a more streamlined way to do this, e.g. a single step instead of separate `docker build` and `crossplane function build` steps?

design/design-doc-composition-functions.md
```diff
@@ -175,32 +179,557 @@ metadata:
...
+cheap enough to implement this Function variant that I believe it's worth doing,
+even if it ends up covering a relatively niche case.
+
+### Using a Function
+
+Composition Functions are very flexible, so the user experience could vary a
+lot. However the general pattern is:
+
+1. Find (or build!) a Function you want to use.
+2. Install the Function using the package manager.
```
jbw976, line 717: Can functions be a dependency of a Configuration package? 🤔 - so users won't have to install the Function explicitly, it will be installed automatically through dependency fulfillment when their Configuration package is installed?

design/design-doc-composition-functions.md
```diff
@@ -209,570 +738,353 @@ metadata:
...
     config:
-      apiVersion: database.example.org/v1alpha1
-      kind: Config
-      metadata:
-        name: cloudsql
-      spec:
-        version: POSTGRES_9_6
+      apiVersion: example.org/v1
+      kind: GoTemplate
+      source: Inline  # Or Remote, if you want to pull templates from git.
```
jbw976, line 750: good call to include Remote as an option - when these templates can stand alone, as opposed to inline in the config, it will probably be easier to use standard tooling to lint/test/etc. the template.

design/design-doc-composition-functions.md
```diff
@@ -209,570 +738,353 @@ metadata:
...
+        {{- range $i := until ( . desired.composite.resource.spec.count ) }}
+        ---
+        apiVersion: iam.dummy.upbound.io/v1alpha1
+        kind: Robot
+        metadata:
+          annotations:
+            crossplane.io/external-name: {{ .desired.composite.resource | getExternalName }}-robot-{{ $i }}
+        spec:
+          forProvider:
+          {{- if $i | printf "robot-%d" | getComposedResource }}
```
jbw976, line 761: how does the template rendering runtime know where this `getComposedResource` function lives and how to call it? 🤔

design/design-doc-composition-functions.md
```diff
@@ -209,570 +738,353 @@ metadata:
...
+    apiVersion: database.example.org/v1
+    kind: XPostgreSQLInstance
+  functions:
+  - name: compose-xr-using-go-templates
+    functionRef:
+      name: starlark
+    config:
+      apiVersion: example.org/v1
+      kind: StarlarkScript
+      script: |
```
jbw976, line 794: do you expect that all types of scripts/templates would have Remote option instead of Inline? This example doesn't imply Remote is a possibility, so I'm curious.

design/design-doc-composition-functions.md
```diff
@@ -209,570 +738,353 @@ metadata:
...
+# Test a single Function by passing it a JSON RunFunctionRequest. This pulls and
+# starts the Function, makes the request, stops it, then returns the result.
+$ kubectl crossplane function test xpkg.upbound.io/negz/go-templates:v0.1.0 run-function-request.json
+{
+  # JSON encoded RunFunctionResponse omitted for brevity.
+}
 ```
 
-A function that wanted to set only an XR connection detail could return:
+You could similarly imagine a more holistic, `helm template`-like experience
```
jbw976, line 867: absolutely love this thinking here to drastically speed up the development process! 💪

design/design-doc-composition-functions.md
negz: Thanks - I made and fixed this mistake in a bunch of places but missed this one. Will update.

design/design-doc-composition-functions.md
negz: I like this API. Will update. Thanks!

design/design-doc-composition-functions.md
jbw976: i think this answers my questions on how can we speed up the development process and allow folks to run their composition "inner loop" more in isolation when they do have patch & transform logic also...

to confirm: it would not be possible to use something like `composition render` unless P&T is also supported/converted to a Function also? 🤔

design/design-doc-composition-functions.md
negz: Hrm, potentially. It would be neat to do at least optionally. I'll update the doc to mention it.

design/design-doc-composition-functions.md
```diff
@@ -209,570 +738,353 @@ metadata:
...
+These metrics pertain to Kubernetes controllers, and will be insufficient when a
+significant portion of Composition logic is 'outsourced' to Functions. It will
+be important to establish a pattern for instrumenting Functions (e.g. as part of
+the Function SDKs). Request tracing in particular is likely to be useful in
+order to debug slow or failing Functions in a pipeline.
 
-```protobuf
-syntax = "proto3";
+Metrics and tracing for Functions must be implemented before the feature becomes
+generally available (GA).
```
jbw976, line 951: worth starting a GA vs Beta checklist in this doc (or tracking issue)?

design/design-doc-composition-functions.md
```diff
@@ -209,570 +738,353 @@ metadata:
...
+              settings:
+                tier: db-custom-1-3840
+                dataDiskType: PD_SSD
+        patches:
+        - type: FromCompositeFieldPath
+          fromFieldPath: spec.parameters.storageGB
+          toFieldPath: spec.forProvider.settings.dataDiskSizeGb
 ```
 
+Removing P&T support from the Composition type would be a breaking API change -
```
jbw976, line 1051: could they exist simultaneously? e.g. P&T section still exists, but could be left totally empty, while the Functions based P&T approach is also supported within the same Composition?

jbw976 (commented):
Awesome work @negz!! I really like the goals of this document around making functions easier to write and to use 🙇 🤩 

A few questions for your consideration to help explain a few gaps in my understanding, nothing major blocking from my side!

design/design-doc-composition-functions.md
```diff
@@ -175,32 +179,557 @@ metadata:
...
+previous example Composition. A Composition specifies which Function to run by
+referencing the package's name. Like any other package, the `Function` type will
+have a corresponding `FunctionRevision` type.
+
+### Calling a Function
+
+Despite the name, a 'Function' is actually more like a 'function server'. Under
+this proposal, Functions are long-running processes. When you install one, the
+package manager deploys it using a Kubernetes Deployment - the same way it would
+deploy a Provider.
```
phisco, line 239: What about calling the "function server" `FunctionRunner` and `Function` an actual configuration to be run with a function runner? I mean a function to me is something that takes an input and produce some output, so something like the following:
```yaml
---
kind: FunctionRunner
metadata:
  name: go-templates
spec:
  package: xpkg.upbound.io/negz/go-templates:v0.1.0
status:
  # The gRPC endpoint where Crossplane will send RunFunctionRequests.
  endpoint: https://go-templates-9sdfn2
---
kind: Function
metadata:
  name: my-xpostgresql-go-templates
spec:
  functionRef:
    name: go-template
  config:
    apiVersion: example.org/v1
    kind: GoTemplate
    source: Remote
    remote: git://github.com/example/my-xpostgresql-go-templates
---
apiVersion: apiextensions.crossplane.io/v2alpha1
kind: Composition
metadata:
  name: example
spec:
  compositeTypeRef:
    apiVersion: database.example.org/v1
    kind: XPostgreSQLInstance
  functions:
  - functionRef:
      name: my-xpostgresql-go-templates
```
which could be inlined as follows:
```yaml
---
apiVersion: apiextensions.crossplane.io/v2alpha1
kind: Composition
metadata:
  name: example
spec:
  compositeTypeRef:
    apiVersion: database.example.org/v1
    kind: XPostgreSQLInstance
  functions:
  - functionRunnerRef:
      name: go-templates
      config:
        apiVersion: example.org/v1
        kind: GoTemplate
        source: Remote
        remote: git://github.com/example/my-xpostgresql-go-templates
```

design/design-doc-composition-functions.md
negz: We should. I struggled to enunciate them so I just skipped it. I think I have a better framing now - will update.

design/design-doc-composition-functions.md
negz: > there is a guarantee that each function in this list is called in the exact order they are listed?

Yes there is.

> I probably prefer keeping the name functions: actually

I mostly like the `step` over yet-another `name`. Right now I slightly prefer `pipeline` too but not strongly.

design/design-doc-composition-functions.md
negz: > to confirm: it would not be possible to use something like composition render unless P&T is also supported/converted to a Function also?

@jbw976 I think it would be possible, but messy.

If all of a Composition's logic is packaged as versioned artifacts (i.e. Functions) then a client (e.g. `composition render`) can easily pull down and test against the same set of artifacts that will be used in production.

If some of the logic is baked into Crossplane then presumably we need to bake that logic into the client (`composition render`) too. Then it becomes important that the client is running the same code as the server (i.e. Crossplane).

It's not an unsolvable problem - maybe you just make sure to use `kubectl crossplane` v1.14 if you're using Crossplane v1.14 in production. Maybe `kubectl crossplane composition render` spins up a `kind` cluster and actually installs a specific version of Crossplane to test your composition. It's just a much easier problem to solve when everything is a Function.

design/design-doc-composition-functions.md
```diff
@@ -99,74 +99,78 @@ experiences designing Kubernetes, its precursor Borg, and many generations of
...
+  careful about what we accept. Changes take a long time to reach consensus.
+  They're coupled to Crossplane's release cycle, which means changes can take a
+  long time to become available, and aren't backported to older versions.
+* How folks would ideally configure Composition varies. Some folks prefer a
+  particular configuration DSL, others look for webhooks written in a GPL, etc.
+
+To address these issues, we added support for 'Composition Functions' in
+Crossplane v1.11. Functions are an alpha feature and are off by default. This
+document is an iteration on (and supersedes) the [previous design
+document][alpha-design] based on what we've learned during since the feature was
```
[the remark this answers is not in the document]
negz: Thank you! I noticed this typo at one point but then could not find it again when proof reading heh.

design/design-doc-composition-functions.md
```diff
@@ -175,32 +179,557 @@ metadata:
...
+
+The package manager is responsible for creating a headless Kubernetes Service
+where each Function's Deployment can be reached. The address of the Service will
+be exposed as the `status.endpoint` of the Function resource. The Service must
+be headless in order for Crossplane's gRPC client to load-balance connections
+when there are multiple Function replicas.
+
+Note that the fact this endpoint is powered by a Service is an implementation
+detail; it may be possible for Functions to be reached (and indeed deployed) by
+other means in future (see [Runtime Configuration][#runtime-configuration]).
```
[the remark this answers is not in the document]
negz: Ah yes - I think it needs to be `(#runtime-configuration)`. Will fix.

design/design-doc-composition-functions.md
```diff
@@ -175,32 +179,557 @@ metadata:
...
+gRPC communication between Crossplane and a Function will be secured by mutual
+transport layer security (mTLS). Crossplane has an established pattern for
+setting up mTLS certificate bundles - we do this today to secure communication
+between the API server and our validation webhooks, and between Crossplane and
+External Secret Stores. We might want to [rethink this][cert-per-entity] in
+future, but to begin with we'll stick with the established pattern, i.e.:
+
+* The Helm chart creates an empty Secret for the client bundle.
+* The core init container bootstraps the client bundle.
+* The package manager creates a Secret for each Function's server bundle.
```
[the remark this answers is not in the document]
negz: Presumably we want only the Function's deployment to be able to read this (i.e. to load its server cert).

I'm not sure whether the implementation we have for webhooks and secret stores enforces this limitation though - I imagine these secrets are in `crossplane-system` and therefore anything with RBAC access to read secrets in that namespace can read them.

design/design-doc-composition-functions.md
```diff
@@ -175,32 +179,557 @@ metadata:
...
+* There are 100 XRs, all of the same type, and all using the same Composition.
+* The XRs have all reached a steady state, polling at the default 60s frequency.
+* The Composition includes 1 Function call.
+
+In this case the Function would be called __100 * 1 = 100 times per 60s__.
+
+A few approaches are possible to scale Functions:
+
+* Tune the `--poll-interval`, e.g. to every 5m or 10m.
+* Scale heavily used Functions horizontally, by increasing Deployment replicas.
```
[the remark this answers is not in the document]
negz: Yes - they're gRPC servers which take care of concurrently serving multiple callers. Will update to be explicit.

design/design-doc-composition-functions.md
```diff
@@ -175,32 +179,557 @@ metadata:
...
+
+  return rsp, nil
+}
+```
+
+The goals of a Composition Function SDK are to:
+
+* Expose an API that makes writing Functions intuitive and self-documenting.
+* Steer Function authors toward best practices, and away from anti-patterns.
+* Eliminate boilerplate, like setting up a gRPC server.
```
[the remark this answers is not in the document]
negz: For the server setup, I think we can get rid of basically everything. I imagine it would all be in a separate file (e.g. `main.go`) so you could choose to override if if you wanted to, but did not have to.

The goal is definitely to make folks feel like they're writing (literally) a function, not building a web service. I'll make that explicit.

design/design-doc-composition-functions.md
```diff
@@ -175,32 +179,557 @@ metadata:
...
+think of this as a special kind of Composition Function SDK that builds a
+Composition Function from an OCI container, rather than building a Function a
+literal _function_ (e.g. written in Go).
+
+![Crossplane calling two containerized Functions via gRPC](assets/design-doc-composition-functions/containerized-functions.png)
+
+Under this design each containerized Function is "wrapped" in an `xfn`-like
+Composition Function. This means each Function exists as its own Deployment.
+
+This design merges the Protobuf `RunFunctionRequest` with the YAML `FunctionIO`
```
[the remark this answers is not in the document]
negz: Maybe a link to the `FunctionIO` schema (i.e. Go type) would be sufficient?

design/design-doc-composition-functions.md
negz: I'm not sure. I imagine there could be, but it might involve our CLI embedding bits of Docker or shelling out to `docker`, which may not be desirable.

design/design-doc-composition-functions.md
negz: Yes absolutely. I imagine the dependency relationship might become Configuration -> Function -> Provider in a lot of cases (as opposed to Configuration -> Provider today). I'll try find somewhere to work that in.

design/design-doc-composition-functions.md
negz: Keep in mind this is just a sketch, not something I've _fully_ worked out. 😄 

That said, I think the way this works is that the author of the Go Templating Composition Function would provide a set of template functions that could be used in templates.

Helm has a built-in set for example - https://helm.sh/docs/chart_template_guide/function_list. Many of them come from https://github.com/Masterminds/sprig, which is kind of a little stdlib for go-templates. Others are specific to Helm. I imagine we'd do the same - include sprig, and also write a few things like this that made life easier for folks writing Compositions.

I don't think these are easily discoverable or self-documenting, so we'd have to do like Helm does and write documentation that explains how this Function works. I imagine that would likely end up at https://marketplace.upbound.io and/or in the Function's README.

design/design-doc-composition-functions.md
negz: These are just sketches of possible Functions, so it would really be up to the author of each Function.

In this case I figure the main utility is "I want to write a quick script inline of my Composition" so I left it out, but there's no reason you couldn't support Remote here too. As you pointed out on the Go Templates example the advantage of doing that is you get better syntax highlighting, linting, etc when your code isn't nested in an opaque string in a YAML document. 😄

design/design-doc-composition-functions.md
negz: I'm inclined to use an issue for this since they're easier to update as we remember more things. 😄 

I think to start with if/when this is merged we'll want to update https://github.com/crossplane/crossplane/issues/3751 to reflect what needs to be done to hit beta. Maybe once we hit beta we can open another issue to track things we need to hit GA.

design/design-doc-composition-functions.md
negz: Yeah, that's close to what I was suggesting in the following sentence.

design/design-doc-composition-functions.md
negz: > I mean, a function to me is something that takes an input and produce some output

I think that's the "trick" here. We're really trying to frame these things (that are in practice more like servers) as "functions". We want folks to feel like they're building, installing support for, and calling "just a function" - not a server or runner. The fact that we make it a server at build time is a bit of an "implementation detail".

So to that end I think I prefer sticking with just "function" rather than introducing a runner type.

I see the (optional) config block in a Composition as kind of like an argument to a function call. In pseudocode Crossplane is making a call like this:

```
run_function(xr, composedResources, config) (xr, composedResources)
```

I think there will be a desire to avoid repeating the same config across multiple function calls, but I'd rather do that by allowing Functions to deliver a Config CRD, which could be instantiated and referenced. This has the added benefit of allowing Functions to deliver fully (OpenAPI) schemafied config types that the API server could validate.

negz: @turkenh @jbw976 what do you think here?

phisco: I think my confusion comes from the fact that I would think to be writing more “runtime” based functions than full-fledged ones, hence my point of separating functions and function runners (a.k.a. runtimes).

Maybe it’s because I’m thinking of something similar to [this](https://docs.konghq.com/hub/kong-inc/serverless-functions/) which I used in the past, but as a user I would expect to be writing more quick “scripts” in Python, Starlark or some other scripting language, rather than really complex “functions” that would require to be wrapped in a grpc server on their own.

But I feel your focus is instead more on full-fledged functions, so it makes sense to consider “runtime” ones as an exception/hack w.r.t. the concept of a function, in which one of the arguments to the function being called is actually custom logic/code, and therefore don’t deserve being first-class citizens of the API.

design/design-doc-composition-functions.md
```diff
@@ -175,32 +200,606 @@ metadata:
...
+API server to create, update, or delete composed resources. Instead, they
+instruct Crossplane which resources should be created, updated, or deleted.
 
-Under the proposed design functions could also be used for purposes besides
+Under the proposed design Functions could also be used for purposes besides
 rendering composed resources, for example validating the results of the
-`resources` array or earlier functions in the `functions` array. Furthermore, a
-function could also be used to implement 'side effects' such as triggering a
+`resources` array or earlier Functions in the `pipeline`. Furthermore, a
+Function could also be used to implement 'side effects' such as triggering a
```
sttts, line 239: While we cannot stop this from happening, it feels odd to call this out as an expected (and encouraged?) practice.

design/design-doc-composition-functions.md
```diff
@@ -175,32 +200,606 @@ metadata:
...
+* The package manager creates a Secret for each Function's server bundle.
+* The package manager bootstraps each Function's server bundle.
+
+See [#3884] for an example of this pattern.
+
+The frequency with which Crossplane will call each Function is relative to how
+frequently each XR is reconciled. There are two main triggers to reconcile an XR:
+
+1. A watch-triggered reconcile, i.e. because the XR changed.
+2. A poll-triggered reconcile, by default once per 60 seconds.
```
sttts, line 444: no watch on the composition?

design/design-doc-composition-functions.md
```diff
@@ -175,32 +200,606 @@ metadata:
...
+Function performance. In practice the throughput of a particular Function will
+be determined by the language it's written in, and what it does.
+
+Inevitably some Functions will be called frequently enough to exhaust the
+processing capacity of a single pod. A few potential approaches are possible to
+scale Functions include:
+
+* Tune the `--poll-interval`, e.g. to every 5m or 10m, to reduce calls.
+* Scale heavily used Functions horizontally, by increasing Deployment replicas.
+* Cache the responses of idempotent Function calls (see [Caching](#caching)).
```
sttts, line 474: do we call out that function must be deterministic? (nit: idem-potency is something else).

design/design-doc-composition-functions.md
```diff
@@ -175,32 +200,606 @@ metadata:
...
+The frequency with which Crossplane will call each Function is relative to how
+frequently each XR is reconciled. There are two main triggers to reconcile an XR:
+
+1. A watch-triggered reconcile, i.e. because the XR changed.
+2. A poll-triggered reconcile, by default once per 60 seconds.
+
+XRs rely on poll-triggered reconciliation to promptly correct drift of their
+composed resources. The poll interval can be set using the `--poll-interval`
+flag. The XR controller does not know what kinds of resources it will compose
+when started, so it cannot start a watch for them.
```
sttts, line 449: The function invocation should not depend on the composed objects.

We know that the first function in the pipeline only depends on the XR, claim (we make that available too, don't we?) and the composition, and with that the whole pipeline.

What we need is a way to detect drift of the composed objects in order to trigger another run of the pipeline on demand. Technically, we don't even need a cached image for the trigger (we have to detect deletions and updates to the spec of composed objects).

----
With that said, it would be good to clearly call out the assumptions a function author can depend on. Being called every 60s should not be one of them IMO. Determinism for a defined set of inputs should be a requirement for every function. That keeps the freedom for us to improve the implementation later.

design/design-doc-composition-functions.md
phisco: Totally agree with @sttts that it's mainly cosmetics, so definitely not a blocker.

Both approaches have pros and cons, for example, going with FunctionRunners + Functions, how would you define a full-fledged single-purpose function built with the flow you are describing? 
With this model it would have to be defined as a FunctionRunner and should either be referenced directly or through a empty useless Function, but it's not going to run other functions. A solution to this could be to have a `kubernetes` `FunctionRunner` which would take the grpc server as an OCI image built with the flow you are describing through a Function and deploy it as an additional deployment.
Something like this:
```yaml
---
kind: FunctionRunner
metadata:
  name: kubernetes
spec:
  package: xpkg.upbound.io/crossplane/kubernetes-xfn-runner:v0.1.0
---
kind: Function
metadata:
  name: my-awesome-func
spec:
  functionRunnerRef:
	name: kubernetes
  config:
	apiVersion: example.org/v1
	kind: Kubernetes
	package: phisco/my-awesome-func:v0.1.0
---
apiVersion: apiextensions.crossplane.io/v2alpha1
kind: Composition
metadata:
  name: example
spec:
  compositeTypeRef:
    apiVersion: database.example.org/v1
    kind: XPostgreSQLInstance
  pipeline:
  - step: compose-xr-using-my-awesome-func
    functionRef:
      name: my-awesome-func
  - step: validate-composed-resources
    functionRef:
      name: cel-validation
```
This would allow decoupling the interface between the Crossplane and `FunctionRunners` from that between the `kubernetes FunctionRunner` and functions built with the workflow you described, if needed and if we are ok having the communication going through the `FunctionRunner`, otherwise we could let `FunctionRunners` provide a grpc endpoint back to crossplane for functions too so that Crossplane could directly communicate with the Function. Obviously inlining this kind of functions would be inefficient as we would have to spin up the deployment, but could still be doable in some way.

design/design-doc-composition-functions.md
sttts: I learned today that compositions are not well-founded in the sense that they can depend on siblings from the life-system through `ToCompositeFieldPath`. We are actually not really building just compositions, but at the same time they are transitions of a state machine. Mind blowing and frightening.

Our innnocent looking sentence in the docs:
> FromCompositeFieldPath and ToCompositeFieldPath patches can also take a wildcarded field path in the toFieldPath parameter and patch each array element in the toFieldPath with the singular value provided in the fromFieldPath.

design/design-doc-composition-functions.md
sttts: Actually, we need something like "eventually idem-potent", i.e. for every x some n with `fn^n(x) == x`, i.e. convergence.

design/design-doc-composition-functions.md
negz: This is copied from the original design. It's a side-effect of @bassam pushing us to consider whether Functions should be used for "day two" operations. I don't think anyone feels great about that - I can remove it.

design/design-doc-composition-functions.md
negz: There is not today iirc. I'm not sure we'd need one - you need to update the XR to use a new CompositionRevision iirc. So a change of Composition would require changing the XR. That said, I could be forgetting something. No concerns about watching Composition (or CompositionRevision) if it helps.

design/design-doc-composition-functions.md
negz: Good catch - will update. (I mix up these terms all the time - symptom of not having had a formal CS education.)

design/design-doc-composition-functions.md
negz: Couldn't both work?

There's a section on potentially watching composed resources under "Future Considerations". Also https://github.com/crossplane/crossplane/issues/4316.

design/design-doc-composition-functions.md
negz: Added a section making this explicit.

design/design-doc-composition-functions.md
jbw976: I'm not sure I follow what having separate `FunctionRunner` and `Function` types buys us that is more desirable than @negz's current proposal.  I like the simplicity of the `Function` abstraction @negz has proposed where the details about how its run is not something the platform operator has to worry about.  I feel it keeps the concepts simple all around 😊

jbw976 (approved):
Thanks for addressing all of my feedback @negz, this is looking good to me with the feedback you've incorporated! 💪

design/design-doc-composition-functions.md
phisco: @jbw976, it only feels less confusing to me, but as we said it’s mainly cosmetics and is not changing the end result. However given that we are proposing to  promote it to Beta, I feel it’s worth discussing about the API and trying to make the UX as smooth as possible.

I think the simplicity of just having Functions is not going to last long, we are almost surely going to have something like a FunctionConfig down the line, to make Functions+config reusable, at least for the “runtime” based functions, such as the starlark example below, where the config is actually used to provide custom logic and it’s going to be really painful to duplicate across multiple Compositions if needed. Which is probably going to overlap with the config to run the Function server, which we’ll most probably need too down the line in some form.

So, instead of stretching the concept of Function all over the places and having Functions that are actually servers and “runtime” Functions that run custom user-provided functions and then complex Functions that are actually single purpose servers built by users. I feel more natural and clear as an API to have FunctionRunners that run Functions, this way “runtime” functions are actually FunctionRunners and function+config becomes just a Function, and complex user-provided functions just become Functions run by a specific FunctionRunner spinning up deployments.

turkenh: I prefer to avoid introducing a new `FunctionRunner` concept to our users and keep it as implementation details as long as it makes sense.

I agree with @phisco's concerns around the need for reusable config's for functions, however, would like to see it as `Function` (as package) + `FunctionConfig` (as CRD), which is consistent with what we have today with `Provider` (as package) + `ProviderConfig` (as CRD). (This could also help with the passing credential problem in #3718 /cc @ezgidemirel).

turkenh (approved):

design/design-doc-composition-functions.md
negz: After some discussion with @turkenh I've renamed `config` to `input`. This better communicates what this type is for. It doesn't configure the `Function` type in any global way - it just specifies optional input passed with the Function call.

phisco: Wouldn’t this mean that a function would have two types of inputs? The config and the actual input coming from the P&T pipeline or TM the previous function?

negz: > Wouldn’t this mean that a function would have two types of inputs? The config and the actual input coming from the P&T pipeline or the previous function?

Yes - that's what I was trying to communicate before. The input to a "function call" is like `function(xr, composed_resources, input)`.

The mental model of a call being more like `runner(input)(XR, composed_resources)` does conceptually make sense to me, especially given that we think some "inputs" will be repeated a lot - i.e. N Compositions might all call the same function with the same input (fka config).

I still prefer to not expose "runner" as a distinct concept just because:

* There's fewer concepts to think about overall. You just install ("define") a Function, then call it (from a Composition).
* Similarly, there's fewer API types to worry about. In many cases you just install a `Function` (often as a dependency of your Configuration) and call it from your Composition. You don't need to install a `FunctionRunner`, then define a `Function`, then call that from your Composition.

phisco: I still think most people are going to use “runtime” based functions, at least that’s what I would do. So the two options are actually Function (+ PackageRuntimeConfig) + FunctionConfig or FunctionRunner (+ PackageRuntimeConfig) + Function. I still think the latter would be clearer to me, and it’s actually more similar to the original proposal, just with less focus on OCI images. However as we said it’s mostly cosmetic, so I’m ok marking this thread as resolved given that it looks like there is definitely consensus around the proposed solution.

design/design-doc-composition-functions.md
sttts: just thinking that this matters in the development workflow. In prod, compositions never change, in dev's inner loop all the time. So a watch would be nice.

design/design-doc-composition-functions.md
```diff
@@ -175,32 +200,609 @@ metadata:
...
+  google.protobuf.Struct resource = 1;
+
+  // The resource's connection details.
+  map<string, bytes> connection_details = 2;
+}
+
+// A Result of running a Function.
+message Result {
+  // Severity of this result.
+  Severity severity = 1;
```
ezgidemirel, line 356: What is the expected behavior of Crossplane when a gRPC call returns an error? For example, should it trigger an event, modify the Sync condition etc.?

negz: I think we probably want it to fail the whole pipeline and return an error. Doing so would cause us to surface the error in a status condition and an event.

There's an entry in https://github.com/crossplane/crossplane/issues/3751 that mentions we should return errors with meaningful gRPC status codes, which will still be relevant under this design. In future we could potentially use this to better understand what errors are fatal vs returnable.

negz:
I believe we have good consensus that this is a better technical direction for Composition Functions. I'm going to merge this design so we can start moving in this direction in time for Crossplane v1.14.

I do want to call out again that the user and developer experience parts of this design should be considered _examples_ of potential experiences. We need to test them with community members and will surely iterate on them between now, beta, and beyond.

What we do have confidence on is that it's better to build Functions as packages, and deploy them as long-running processes.

gberche-orange:
@negz late comment regarding generic functions and Starlak example:

Carvel ytt (https://carvel.dev/ytt/)  is doing a great job of leveraging Starlak for specifically supporting a wide range of yaml files manipulation use-cases: mixing plain yaml and starlak fragments.  I have successfully used ytt to craft existing compositions while leveraging full IDE support (see https://github.com/crossplane/crossplane/issues/3197#issuecomment-1194624402). I planned to test an alpha xfn using ytt cli (https://github.com/orange-cloudfoundry/paas-templates/issues/1813 ) before I learned about this new proposal. 

Depending on roadmap for this proposal to be implemented, a containerized Function leveraging ytt cli might provide good feedback on how useful a ytt generic function could be, and help draft possible workflows such as "git-ops function authoring". 

A ytt generic function might need to leverage
- inline ytt fragments
- ytt fragments loaded from configmap possibly mounted as volumes in the Deployment
- ytt fragments loaded from a url/git repo

gberche-orange:
Another late idea related to this proposal

> a new pipeline array of Function calls 

I was wondering whether there are other examples of such pattern in the kubernetes ecosystem.

The admission control system seems to share many goals with crossplane composition functions.


https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/

Was it considered to leverage the mutating webhook infrastructure to invoke  composition functions developped in GPL or existing contemporary "Patch and Transform" libraries such as kyberno ?

This approach may have the potential to reuse and leverage many of the existing k8s infrastructure for invocating webhooks that crossplane would have to implement when issuing direct grpc calls to composition functions.

Examples benefits:
- Reusable composition functions could be published as kyberno policies which can easily be discovered/shared
- k8s webhook SDKs can be used to author GPL crossplane composition functions in many languages

I have not found traces of such discussion in the design proposal or from memory in the alpha composition design. Did I miss it ?

Ps: I'm currently out-of-office with limited connectivity. My responses on this thread will be late.

PR #4307, opened 2023-07-10
controller/rbac/namespace: edge based events for role updates

sttts:

The `Applied RBAC roles` event is repeated again and again:
```
default          9m59s (x450 over 3d4h)   Normal    ApplyRoles               Namespace/upbound-system                                         Applied RBAC Roles
default          9m59s (x455 over 3d4h)   Normal    ApplyRoles               Namespace/default                                                Applied RBAC Roles
default          9m59s (x440 over 3d4h)   Normal    ApplyRoles               Namespace/kube-system                                            Applied RBAC Roles
default          9m59s (x437 over 3d4h)   Normal    ApplyRoles               Namespace/kube-public                                            Applied RBAC Roles
default          9m59s (x484 over 3d4h)   Normal    ApplyRoles               Namespace/kube-node-lease                                        Applied RBAC Roles
```

This PR supresses the event if nothing happens, and it prints up to the first 3 role names in case it's not an noop. With that the event is actually useful.

I have:

- [ ] Added or updated unit **and** E2E tests for my change.


internal/controller/rbac/namespace/reconciler.go
```diff
@@ -210,9 +213,15 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		}
 
 		log.Debug("Applied RBAC Role")
```
jbw976, line 215: would it be useful in this debug log statement to include the name of the RBAC role that was just applied?

internal/controller/rbac/namespace/reconciler.go
```diff
@@ -210,9 +213,15 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+		applied = append(applied, rl.GetName())
 	}
 
-	r.record.Event(ns, event.Normal(reasonApplyRoles, "Applied RBAC Roles"))
+	sort.Strings(applied)
+	if len(applied) > 3 {
```
jbw976, line 220: is this > 3 pattern pretty common in upstream k8s? i feel like i may have seen this before - not 💯 on it though, as I have hazy memories of thinking to myself "well why didn't you just tell me the rest of them too then?" 😂 

more curious than anything - not blocking from me 😁

jbw976 (commented):
Seems reasonable to only create an event when we do something - do you have an example from your testing of what events in this style look like now?

internal/controller/rbac/namespace/reconciler.go
sttts: I think we use something like that in OpenShift. No strong opinions. I thought 3 is a good number :) 10 might be too.

internal/controller/rbac/definition/reconciler.go
```diff
@@ -193,11 +196,18 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
 		}
 		log.Debug("Applied RBAC ClusterRole")
+		applied = append(applied, cr.GetName())
+	}
+
+	sort.Strings(applied)
+	if len(applied) > 3 {
+		r.record.Event(d, event.Normal(reasonApplyRoles, fmt.Sprintf("Applied RBAC ClusterRoles: %s, %s, %s, and %d more", applied[0], applied[1], applied[2], len(applied)-3)))
+	} else if len(applied) > 0 {
+		r.record.Event(d, event.Normal(reasonApplyRoles, fmt.Sprintf("Applied RBAC ClusterRoles: %s", strings.Join(applied, ", "))))
```
negz, line 206: Given this pattern is repeated a few times I wonder whether a little utility could help? I'm on the fence. Normally I prefer to avoid such little utilities for things like this, until they're almost everywhere, but I must admit I'm a little (unreasonably?) allergic to `if then else` statements.

I'm thinking something like:

```go
r.record.Event(d, event.Normal(reasonApplyRoles, maybeTruncate("Applied RBAC ClusterRoles", applied)))

// maybeTruncate needs a better name :)
func maybeTruncate(prefix string, names []string) string {
	if len(names) > 3 {
		return fmt.Sprintf("%s: %s, and %d more", prefix, strings.Join(names[:3], ", "), len(names)-3)
	}
	return fmt.Sprintf("%s: %s", prefix, strings.Join(names, ", "))
}
```

negz (approved):
This looks great, thank you. FWIW there's a similar event WRT selecting a Composition in the XR controller that I've been meaning to fix if you're feeling motivated to do so.

Few style nits, but nothing blocking.

negz:
PS the `PR / checklist-completed` check is failing because it wants you to explicitly `~strikethrough~` the list items you have not completed.

internal/controller/rbac/provider/binding/reconciler.go
```diff
@@ -204,15 +209,25 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		"subjects", subjects,
 	)
 
-	if err := r.client.Apply(ctx, rb, resource.MustBeControllableBy(pr.GetUID())); err != nil {
+	if err := r.client.Apply(ctx, rb, resource.MustBeControllableBy(pr.GetUID()), resource.AllowUpdateIf(ClusterRoleBindingsDiffer)); err != nil && !resource.IsNotAllowed(err) {
 		log.Debug(errApplyBinding, "error", err)
 		err = errors.Wrap(err, errApplyBinding)
 		r.record.Event(pr, event.Warning(reasonBind, err))
 		return reconcile.Result{}, err
+	} else if err == nil {
```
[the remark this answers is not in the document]
sttts: done

internal/controller/rbac/definition/reconciler.go
sttts: added. Any good place to put it? Have 3 copies now.

internal/controller/apiextensions/claim/reconciler.go
```diff
@@ -528,8 +529,12 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		return reconcile.Result{Requeue: true}, errors.Wrap(r.client.Status().Update(ctx, cm), errUpdateClaimStatus)
 	}
 
-	log.Debug("Successfully applied composite resource")
-	record.Event(cm, event.Normal(reasonCompositeConfigure, "Successfully applied composite resource"))
+	if oldRV != cp.GetResourceVersion() {
```
sttts, line 532: not sure I am a fan of this construct. Am open for better patterns.

See my comments about the available applicators in Slack.

internal/controller/rbac/definition/reconciler.go
negz: I could imagine this being added as a utility in crossplane-runtime somewhere (not sure which package). I'm fine leaving a few copies scattered around for now though - we can figure out a better home if and when it becomes more widely used.

internal/controller/apiextensions/claim/reconciler.go
negz: Perhaps something like:

```go
err := r.client.Apply(ctx, cp, resource.AllowUpdateIf(func(current, desired) bool { !cmp.Equal(current, desired) }))
switch {
case resource.IsNotAllowed(err):
	log.Debug("Skipped no-op composite resource apply")
case err != nil:
	// Current error handling block
default:
	log.Debug("Successfully applied composite resource")
	record.Event(cm, event.Normal(reasonCompositeConfigure, "Successfully applied composite resource"))
}
```

I believe this would disallow the apply (i.e. patch) entirely if this reconcile didn't actually change `cp`. Using `cmp.Equal` on the entire `*Unstructured` feels inelegant but this approach at least seems syntactically more pleasant to me.

https://pkg.go.dev/github.com/crossplane/crossplane-runtime@v0.19.2/pkg/resource#AllowUpdateIf

sttts: Was wondering whether this can still lead to multiple events when the informers are lagging behind. Just because an update is attempted does not mean an actual update happened.

negz: > Was wondering whether this can still lead to multiple events when the informers are lagging behind.

FWIW I _think_ the controller-runtime machinery automatically dedupes multiple events in the queue to result in one reconcile call.

sttts: and in any case this will eventually settle. If two or more, but few events are creating on rare races, that's fine. Just cosmetics (different to the RV topic in the applicator which is about consistency).

sttts: updated.

negz (approved):

PR #4310, opened 2023-07-11
feat(xrd): add labels and annotations

### LINKED ISSUE #4149: Inject caBundle into CompositeResourceDefinition for conversions
Hi crossplane team,

As discussed with @ytsarev we are currently facing the issue, that we want convert XRDs from one to another version using conversions webhooks like defined here https://github.com/crossplane/crossplane/pull/3940. To get it running, we need to define a caBundle in the `spec.conversion.webhook.clientConfig.caBundle` field of the `CompositeResourceDefinition` according to https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#service-reference. As we currently have a big number of XRDs (stored inside a git repository, managed by different people) we would need to add the `caBundle` to every XRD manually and if the certificate changes we would need to change it in every XRD. 

There are multiple options how we could inject this needed information on the fly:

1. As CA Injector from cert-manager actually is made exactly for this purpose, we could add an annotation `cert-manager.io/inject-ca-from: example1-ns/webhook1-certificate<or-secret>` to the CRD and CA Injector would care about the `caBundle`. However, currently due to https://github.com/crossplane/crossplane/pull/3124 annotations of a XRD are not propagated to the rendered Claim/Composite CRDs

2. Implement a secret selector property in XRDs like `spec.conversion.webhook.secretSelector` and Crossplane core component would inject the secret into the rendered Claim/Composite CRDs. But this seems not that easy to get it running, as [extv1.CustomResourceConversion](https://github.com/crossplane/crossplane/blob/master/apis/apiextensions/v1/xrd_types.go#LL97C20-L97C44) is reusing the struct from upstream k8s 

- Define which annotations are propagated from XRD to CRD
- SecretSelector
- ... ?

fhopfensperger:

Allows one to specify labels and annotations to be added to the rendered composite resource Composite Resource (and Claim if specified) `CustomResourceDefiniton`s.

For example: 
```yaml
metadata:
  name: xpostgresqlinstances.database.example.org
  annotations:
    "example.com/some-other-annotation": "not-propagated"
  labels:
    "example.com/some-label": "value1",
spec:
  metadata:
    annotations:
      "cert-manager.io/inject-ca-from": "example1-ns/webhook1-certificate"
    labels:
       "example.com/some-additional-label": "value2",
  group: database.example.org
  names:
    kind: XPostgreSQLInstance
    plural: xpostgresqlinstances
  claimNames:
    kind: PostgreSQLInstance
    plural: postgresqlinstances
  versions:
  ...
```
The rendered CompositeResource and Claim CRDs would have 
```yaml
labels:
  "example.com/some-label": "value1",
  "example.com/some-additional-label": "value2",
annotations:
  "cert-manager.io/inject-ca-from": "example1-ns/webhook1-certificate"
```

This PR doesn't break existing functionality as the `labels` were already automatically propagated to XR/Claim CRDs instead it let one define additional labels/annotations which are exclusively added to XR and Claim CRD.

Fixes #4149

I have:


internal/xcrd/crd.go
```diff
@@ -233,6 +233,25 @@ func getProps(field string, v *v1.CompositeResourceValidation) (map[string]extv1
...
+func setCrdMetadata(crd *extv1.CustomResourceDefinition, xrd *v1.CompositeResourceDefinition) *extv1.CustomResourceDefinition {
+	crd.SetLabels(xrd.GetLabels())
+	if xrd.Spec.Metadata != nil {
+		if xrd.Spec.Metadata.Labels != nil {
+			if crd.GetLabels() == nil {
+				crd.SetLabels(map[string]string{})
+			}
+			for k, v := range xrd.Spec.Metadata.Labels {
+				crd.GetLabels()[k] = v
+			}
```
phisco, line 246: Minor nit: What  about something like this instead?
```suggestion
			inheritedLabels := crd.GetLabels()
			if  inheritedLabels == nil {
				inheritedLabels = map[string]string{}
			}
			for k, v := range xrd.Spec.Metadata.Labels {
				inheritedLabels[k] = v
			}
			crd.SetLabels(inheritedLabels)
```

fhopfensperger: I don't have a strong opinion about it :) Sure why no, it maybe improves readability.

internal/xcrd/crd_test.go
```diff
@@ -818,3 +818,189 @@ func TestForCompositeResourceClaim(t *testing.T) {
 		t.Errorf("ForCompositeResourceClaim(...): -want, +got:\n%s", diff)
 	}
 }
+
+func Test_setCrdMetadata(t *testing.T) {
```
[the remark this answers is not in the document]
fhopfensperger: 😅

phisco (approved):
LGTM, thanks @fhopfensperger! 🚀  Waiting for a feedback by @turkenh and/or @negz  on the overall design, given that they had some concerns in the original issue

fhopfensperger:
@phisco thank you very much for the review and pushing it forward :)

apis/apiextensions/v1/xrd_types.go
```diff
@@ -95,6 +95,10 @@ type CompositeResourceDefinitionSpec struct {
 	// Conversion defines all conversion settings for the defined Composite resource.
 	// +optional
 	Conversion *extv1.CustomResourceConversion `json:"conversion,omitempty"`
+
+	// Metadata specifies the desired metadata for the defined composite resource.
```
turkenh, line 99: nit: 

```suggestion
	// Metadata specifies the desired metadata for the defined composite resource and claim.
```

apis/apiextensions/v1/xrd_types.go
```diff
@@ -103,6 +107,23 @@ type CompositionReference struct {
 	Name string `json:"name"`
 }
 
+// CompositeResourceDefinitionSpecMetadata specifies the desired metadata of the defined composite resource.
+type CompositeResourceDefinitionSpecMetadata struct {
+	// Map of string keys and values that can be used to organize and categorize
+	// (scope and select) objects. May match selectors of replication controllers
+	// and services.
```
turkenh, line 114: Should we also document the behavior with labels, maybe here? something like "additional labels" for CRDs...

turkenh (approved):
LGTM, thanks @fhopfensperger 🙌

apis/apiextensions/v1/xrd_types.go
fhopfensperger: I added a proposal.

fhopfensperger: here i'm not sure about the `'`
https://github.com/crossplane/crossplane/pull/4310/files#diff-23609283c12720d62d533c0151d9b3830bb4b255ae2ae77b7586e66f5cf2f8adR270

fhopfensperger: I just checked the applied CRDs in a cluster and it seems to look good for me
```
$ k explain CompositeResourceDefinition.spec.metadata       
GROUP:      apiextensions.crossplane.io
KIND:       CompositeResourceDefinition
VERSION:    v1

FIELD: metadata <Object>

DESCRIPTION:
    Metadata specifies the desired metadata for the defined composite resource
    and claim CRD's.
    
FIELDS:
  annotations   <map[string]string>
    Annotations is an unstructured key value map stored with a resource that may
    be set by external tools to store and retrieve arbitrary metadata. They are
    not queryable and should be preserved when modifying objects. More info:
    https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations

  labels        <map[string]string>
    Map of string keys and values that can be used to organize and categorize
    (scope and select) objects. May match selectors of replication controllers
    More info:
    https://kubernetes.io/docs/concepts/overview/working-with-objects/labels and
    services. These labels are added to the composite resource and claim CRD's
    in addition to any labels defined by `CompositionResourceDefinition`
    `metadata.labels`.

```

PR #4311, opened 2023-07-11
guide-observability.md: add error guidelines

sttts:

Add a section about errors to the observability contribution document.

Fixes # 

I have:

### DIFF
--- a/contributing/guide-observability.md
+++ b/contributing/guide-observability.md
@@ -120,6 +120,86 @@ logs don't show up in production normally.
 For the question of what constitutes an error, errors should be actionable by a
 human. See the [Dave Cheney article] on this topic for some more discussion.
 
+## Good Errors
+
+An error message is good if it communicates
+
+1. what went wrong
+2. where it went wrong
+3. for which inputs
+
+and if it is concise without redundancies and at best displayable in one line.
+
+The general goal of an error message is to be actionable for the user who is 
+trying to avoid the error in the future.
+
+Examples for bad error messages:
+- `cannot resolve package dependencies: Invalid Semantic Version`
+- `cannot parse file foo.yaml: failed to parse YAML of file foo.yaml`
+
+Examples for how to improve them:
+- `cannot resolve package dependencies in crossplane.yaml: invalid Semantic Version "main" on package "provider-kubernetes"`
+- `failed to load OCI image provider-gcp/v0.42.0: failed to unpack layer: cannot parse file: failed to parse YAML of file foo.yaml: unexpected indention in line 4`.
+
+General rules:
+1. follow the pattern `generic error cause: more details: even more details`.
+2. put values into quotes (`%q` in Golang), with the following exceptions 
+   where quotes don't add to readability:
+   - resource names with namespace or resource type (e.g. `default/foo` or 
+     `providers/foo`), resource names alone are quoted
+   - kinds (e.g. `cannot find provider "provider-kubernetes"`; here, the kind 
+     provider is without quotes, the provider name is with)
+   - filenames with extensions and/or path (e.g. `cannot parse foo.yaml`).
+3. add a context of e.g. filenames, object names, or values which are wrong
+   (and the values are known to be insensitive).
+
+   Which context is added by which functions depends on the level of 
+   abstraction of the code at hand: the package loader will know the OCI 
+   image name, the YAML parser might only see a byte slice. The former adds 
+   the image name as context, some layer in-between adds the filename and 
+   the YAML parser adds the line and column in the byte stream.
+
+   Filenames should be relative to the logical working directory of the 
+   context at hand (e.g. a package root, current working directory, 
+   workspace root). Ensure that the context is clear in the error message (e.g.
+   `failed to load package "provider-kubernetes": failed to parse apis/xrd.yaml`). 
+
+4. add a context just once.
+
+   Rule of thumb: if a function gets a parameter, it will also include it in the 
+   error messages, so the caller does not have to.
+   
+   E.g. a function getting a filename as input will include the filename in 
+   errors it returns.
+5. don't return multi-line errors.
+6. aggregate errors if there is clear use for the user, but not in general, i.e.
+   fail fast is the default.
+
+   Aggregate through `kerrors.NewAggregate`. When the number of aggregated
+   errors is unbounded, return a subset, e.g. the first 3.
+
+   E.g.: `cannot parse workspace: failed to parse foo.yaml, bar.yaml, abc.yaml,
+   and 17 more files`
+7. don't include sensitive information like tokens into error messages, not even
+   in debug mode.
+
+   These are usually considered insensitive: field names, file paths, resources, 
+   kinds, object names, namespaces, labels, enum values, numbers, booleans.
+8. use error message constants, e.g. `errSomethingWentWrong` for messages
+   without interpolation, and `errFmtSomethingWentWrong` for those with
+   interpolation.
+9. use [wrapping errors] and github.com/crossplane/crossplane-runtime/pkg/errors
+   in general.
+10. error messages must be deterministic. Use sorting in loops over maps or sort
+    returned errors to ensure determinism.
+
+    Background: one never knows how an error is displayed. Not being
+    deterministic can lead to hot-loops with constant API updates, e.g. of a
+    condition.
+
+11. error messages are good candidates to be included in condition or event
+    messages.
+
 ## In Practice
 
 Crossplane provides two observability libraries as part of crossplane-runtime:
@@ -187,5 +267,5 @@ implementations.
 [not]: https://dave.cheney.net/2017/01/23/the-package-level-logger-anti-pattern
 [`Reconciler`]: https://godoc.org/sigs.k8s.io/controller-runtime/pkg/reconcile#Reconciler
 [managed resource reconciler]: https://github.com/crossplane/crossplane-runtime/blob/a6bb0/pkg/reconciler/managed/reconciler.go#L436
-[wrapping errors]: https://godoc.org/github.com/pkg/errors#Wrap
+[wrapping errors]: https://godoc.org/github.com/crossplane/crossplane-runtime/pkg/errors#Wrap
 [API conventions]: https://github.com/kubernetes/community/blob/09f55c6/contributors/devel/sig-architecture/api-conventions.md#events


contributing/guide-observability.md
```diff
@@ -120,6 +120,56 @@ logs don't show up in production normally.
 For the question of what constitutes an error, errors should be actionable by a
 human. See the [Dave Cheney article] on this topic for some more discussion.
 
+## Good Errors
+
+An error message is good if it communicates
+
+1. what went wrong
+2. where it went wrong
+3. for which inputs
```
s-urbaniak, line 129: note: obfuscation should potentially be considered if sensitive data is involved. This is a very broad topic though admittedly.

sttts: see below (8)

contributing/guide-observability.md
```diff
@@ -120,6 +120,56 @@ logs don't show up in production normally.
...
+Examples for bad error messages:
+- `cannot resolve package dependencies: Invalid Semantic Version`
+- `cannot parse file foo.yaml: failed to parse YAML of file foo.yaml`
+
+Examples for how to improve them:
+- `cannot resolve package dependencies in crossplane.yaml: invalid Semantic Version "main" on package "provider-kubernetes"`
+- `cannot parse file: failed to parse YAML of file foo.yaml` or the inverse.
+
+General rules:
+1. follow the pattern `generic error cause: more details: even more details`.
```
s-urbaniak, line 145: maybe more specific: add error context information that is bound it its current layer of abstraction. Taking your example of above: `error parsing package depencies content: failed to parse YAML of file foo.yaml`

s-urbaniak: here:
- high level layer: `error parsing package depencies content` happens at the layer that is trying to deserialize into concrete objects (package dependency objects)
- low level layer: `failed to parse YAML of file foo.yaml`: happens at the layer that deals with opening files and parsing generic yaml

contributing/guide-observability.md
```diff
@@ -120,6 +120,56 @@ logs don't show up in production normally.
...
+2. put values into quotes (`%q` in Golang), with the exception of
+   - resource names with namespace or resource type (e.g. `default/foo` or `providers/foo`), resource names alone are quoted
+   - kinds (e.g. `cannot find provider "provider-kubernetes`; here, the kind provider is without quotes, the provider name is with)
+   - filenames with extensions and/or path (e.g. `cannot parse foo.yaml`).
+3. add a context of e.g. filenames, object names, or values which are wrong. Filenames should be relative to the logical working directory of the context at hand (e.g. a package root, current working directory, workspace root). Ensure that the context is clear in the error message (e.g. `failed to load package provider-kubernetes: failed to parse apis/xrd.yaml`).
+4. add a context just once.
+
+   Rule of thumb: if a function gets a parameter, it will also include it in the error messages, so the caller does not have to.
```
s-urbaniak, line 153: 👍

contributing/guide-observability.md
```diff
@@ -120,6 +120,56 @@ logs don't show up in production normally.
...
+   
+   E.g. a function getting a filename as input will include the filename in errors it returns.
+6. don't return multi-line errors.
+7. aggregate errors if there is clear use for the user, but not in general, i.e. fail fast is the default.
+
+   Aggregate through `kerrors.NewAggregate`. When the number of aggregated errors is unbounded,
+   return a subset, e.g. the first 3.
+
+   E.g.: `cannot parse workspace: failed to parse foo.yaml, bar.yaml, abc.yaml, and 17 more files`
```
s-urbaniak, line 162: 👍

contributing/guide-observability.md
s-urbaniak: indeed

contributing/guide-observability.md
sttts: added

contributing/guide-observability.md
```diff
@@ -120,6 +120,61 @@ logs don't show up in production normally.
...
+
+Examples for how to improve them:
+- `cannot resolve package dependencies in crossplane.yaml: invalid Semantic Version "main" on package "provider-kubernetes"`
+- `failed to load OCI image provider-gcp/v0.42.0: failed to unpack layer: cannot parse file: failed to parse YAML of file foo.yaml: unexpected indention in line 4`.
+
+General rules:
+1. follow the pattern `generic error cause: more details: even more details`.
+2. put values into quotes (`%q` in Golang), with the exception of
+   - resource names with namespace or resource type (e.g. `default/foo` or `providers/foo`), resource names alone are quoted
+   - kinds (e.g. `cannot find provider "provider-kubernetes`; here, the kind provider is without quotes, the provider name is with)
```
ezgidemirel, line 148:
```suggestion
   - kinds (e.g. `cannot find provider "provider-kubernetes"`; here, the kind provider is without quotes, the provider name is with)
```

contributing/guide-observability.md
```diff
@@ -120,6 +120,61 @@ logs don't show up in production normally.
...
+   E.g.: `cannot parse workspace: failed to parse foo.yaml, bar.yaml, abc.yaml, and 17 more files`
+9. don't include sensitive information like tokens into error messages, not even in debug mode.
+10. use error message constants, e.g. `errSomethingWentWrong` for messages without interpolation, and `errSomethingWentWrongFmt` for those with interpolation.
+11. use [wrapping errors] and https://godoc.org/github.com/pkg/errors in general.
+12. error messages must be deterministic. Use sorting in loops over maps to ensure determinism.
+
+    Background: one never knows how an error is displayed. Not being deterministic can lead to hotloops
+    with constant API updates, e.g. of a condition.
+
+13. error messages are good candidates to be included in condition or event messages.
```
ezgidemirel, line 176: It would be nice to have a guideline about events as well

ezgidemirel (approved):
Thanks @sttts for adding a guideline! Maybe we can extend it with some good error message examples in the codebase. Other than that, LGTM.

contributing/guide-observability.md
```diff
@@ -120,6 +120,61 @@ logs don't show up in production normally.
...
+7. don't return multi-line errors.
+8. aggregate errors if there is clear use for the user, but not in general, i.e. fail fast is the default.
+
+   Aggregate through `kerrors.NewAggregate`. When the number of aggregated errors is unbounded,
+   return a subset, e.g. the first 3.
+
+   E.g.: `cannot parse workspace: failed to parse foo.yaml, bar.yaml, abc.yaml, and 17 more files`
+9. don't include sensitive information like tokens into error messages, not even in debug mode.
+10. use error message constants, e.g. `errSomethingWentWrong` for messages without interpolation, and `errSomethingWentWrongFmt` for those with interpolation.
+11. use [wrapping errors] and https://godoc.org/github.com/pkg/errors in general.
```
phisco, line 170: We currently prefer `github.com/crossplane/crossplane-runtime/pkg/errors`, which is mainly a wrapper around `github.com/pkg/errors`. Not saying we should continue to do so, but definitely worth taking into consideration if we want to change this.

contributing/guide-observability.md
```diff
@@ -120,6 +120,61 @@ logs don't show up in production normally.
...
+
+6. add a context just once.
+
+   Rule of thumb: if a function gets a parameter, it will also include it in the error messages, so the caller does not have to.
+   
+   E.g. a function getting a filename as input will include the filename in errors it returns.
+7. don't return multi-line errors.
+8. aggregate errors if there is clear use for the user, but not in general, i.e. fail fast is the default.
+
+   Aggregate through `kerrors.NewAggregate`. When the number of aggregated errors is unbounded,
```
phisco, line 164: what about the recently introduced https://pkg.go.dev/errors#Join?

contributing/guide-observability.md
sttts: oh, wasn't aware of that. Will fix.

contributing/guide-observability.md
sttts: Multi-line errors 😱😄

contributing/guide-observability.md
sttts: done, also at the bottom

pedjak:
> Thanks @sttts for adding a guideline! Maybe we can extend it with some good error message examples in the codebase. Other than that, LGTM.

It would be also nice, if this PR would adjust a few errors in this code to match the guidelines.

contributing/guide-observability.md
phisco: I don't have super strong opinions on the topic tbh. I could be fine seeing some `\n` somewhere given that it's in the standard library. But I would also be fine using `kerrors.NewAggregate`, as long as we add it in `github.com/crossplane/crossplane-runtime/pkg/errors`, just to avoid having multiple error libraries.

contributing/guide-observability.md
```diff
@@ -120,6 +120,61 @@ logs don't show up in production normally.
...
+   return a subset, e.g. the first 3.
+
+   E.g.: `cannot parse workspace: failed to parse foo.yaml, bar.yaml, abc.yaml, and 17 more files`
+9. don't include sensitive information like tokens into error messages, not even in debug mode.
+10. use error message constants, e.g. `errSomethingWentWrong` for messages without interpolation, and `errSomethingWentWrongFmt` for those with interpolation.
```
phisco, line 169: The current pattern is to use `errFmt...`:
```suggestion
10. use error message constants, e.g. `errSomethingWentWrong` for messages without interpolation, and `errFmtSomethingWentWrong` for those with interpolation.
```

sttts: we have both. Which do we want? Or both?

sttts:
> It would be also nice, if this PR would adjust a few errors in this code to match the guidelines.

Clearly out of scope. It will likely be an iterative process, maybe different people.

contributing/guide-observability.md
sttts: @jbw976 @negz opinions?

contributing/guide-observability.md
phisco: Any thought in general on the usage of error constants?

My feeling is that it's encouraging some bad behaviours, such as:
- reusing errors from different sources, resulting in confusion on where an error came from, obviously the added context should make it unique, but it's painful to follow the whole chain, so ideally I would like errors to be kind of unique to be able to quickly find their source.
- brittle tests due to error's string representation comparison, I'd rather have structured errors if we actually want to check something specific.
- added cognitive load due to one more layer of indirection, in the end I always have to jump to the actual string to see what's the string, even more for `errFmt...` once, if I'm not the original author and I'm refactoring the code or even if I am, but I wrote it more than a week ago.

I know it's a well known pattern, so I'm not suggesting we change our approach, but given that we are discussing guidelines, I thought it could be worth asking it, worst case scenario in the end I'll have a better understanding of the benefits of this approach. 🙏

contributing/guide-observability.md
sttts: https://goplay.tools/snippet/47cniu4DqpJ surprises me. This is really bad output, even misleading.

sttts: vs. https://goplay.tools/snippet/6I8E-WSj4Ev 
```
something went wrong: [Lorem ipsum dolor sit amet, consectetuer adipiscing elit]
```

contributing/guide-observability.md
phisco: given that we have 3 occurrences of `err...Fmt` and 76 of `errFmt...`, I'd say the former are mistakes and IIRC in a review by either @turkenh or @negz I was pointed to the latter as the standard indeed.

sttts: +1 for getting rid of constants, for all the reasons you mention.

contributing/guide-observability.md
sttts: An attempt: https://github.com/crossplane/crossplane-runtime/pull/478

phisco: > ```
> something went wrong: Lorem ipsum dolor sit amet
> consectetuer adipiscing elit
> ```
I agree it doesn't look good, really unfortunate they didn't provide a way to customise the separator, similar to https://pkg.go.dev/strings#Join.

contributing/guide-observability.md
```diff
@@ -120,6 +120,61 @@ logs don't show up in production normally.
...
+   
+   E.g. a function getting a filename as input will include the filename in errors it returns.
+7. don't return multi-line errors.
+8. aggregate errors if there is clear use for the user, but not in general, i.e. fail fast is the default.
+
+   Aggregate through `kerrors.NewAggregate`. When the number of aggregated errors is unbounded,
+   return a subset, e.g. the first 3.
+
+   E.g.: `cannot parse workspace: failed to parse foo.yaml, bar.yaml, abc.yaml, and 17 more files`
+9. don't include sensitive information like tokens into error messages, not even in debug mode.
```
tnthornton, line 168: Given that the log line author may not necessarily know what is being placed in the field value and whether or not it's sensitive, I wonder if we can cover this by asserting that field/property names are fine to include, but there isn't a reason to log the value of the field.

For example:
```
something went wrong: invalid type for field spec.forProvider.tag.owner: expected bool, got string
```

versus:
```
something went wrong: invalid type for field spec.forProvider.tag.owner: expected bool, got string, value: user@example.com
```

The first example still gives the author/user enough information to track down the problem without potentially leaking sensitive information, whereas the second leaks sensitive information.

This also looks like it may be in direct opposition to part of (3) above:
>add a context of e.g. filenames, object names, or values which are wrong.

specifically, "or values which are wrong".

sttts: If you know the field is not sensitive (like the `type` example, enum style), you can log it or include it into an error. If you write generic code (against any kind of yaml), be more careful. Field names can always be logged.

I guess a username is somewhere in a grey area, maybe depending on legislation.

A common example are object names. They are never encrypted in kube, not even with encryption at REST aka encryption in etcd. Of course you can store your password in the name of an object. But 🤷‍♂️ 

So as rule, maybe "log it if you know it is not sensitive, e.g. for enum fields, object names, file names".

sttts: @tnthornton updated. Please double check.

contributing/guide-observability.md
```diff
@@ -120,6 +120,63 @@ logs don't show up in production normally.
...
+
+General rules:
+1. follow the pattern `generic error cause: more details: even more details`.
+2. put values into quotes (`%q` in Golang), with the exception of
+   - resource names with namespace or resource type (e.g. `default/foo` or `providers/foo`), resource names alone are quoted
+   - kinds (e.g. `cannot find provider "provider-kubernetes"`; here, the kind provider is without quotes, the provider name is with)
+   - filenames with extensions and/or path (e.g. `cannot parse foo.yaml`).
+3. add a context of e.g. filenames, object names, or values which are wrong (and the values are known to be insensitive).
+
+   Which context is added by which functions depends on the level of abstraction of the code at hand: the package loader will know the OCI image name, the YAML parser might only see a byte slice. The former adds the image name as context, some layer in-between adds the filename and the YAML parser adds the line and column in the byte stream.
```
negz, line 152: Nit: Please wrap markdown at 80 chars. I know this is subjective, but it's the prevalent style in our markdown files (e.g. this one). There's a bunch of editor extensions that will do this automagically.

contributing/guide-observability.md
negz: Usually I would prefer the tooling from the stdlib, but I share the concerns about `errors.Join` joining with `\n`. I like the approach of adding a util to our own errors package.

contributing/guide-observability.md
```diff
@@ -120,6 +120,63 @@ logs don't show up in production normally.
...
+   Aggregate through `kerrors.NewAggregate`. When the number of aggregated errors is unbounded,
+   return a subset, e.g. the first 3.
+
+   E.g.: `cannot parse workspace: failed to parse foo.yaml, bar.yaml, abc.yaml, and 17 more files`
+7. don't include sensitive information like tokens into error messages, not even in debug mode.
+
+   These are usually considered insensitive: field names, file paths, resources, kinds, object names, namespaces, labels, enum values, numbers, booleans.
+8. use error message constants, e.g. `errSomethingWentWrong` for messages without interpolation, and `errFmtSomethingWentWrong` for those with interpolation.
+9. use [wrapping errors] and github.com/crossplane/crossplane-runtime/pkg/errors in general.
+10. error messages must be deterministic. Use sorting in loops over maps or sort returned errors to ensure determinism.
```
negz, line 173: Do you know if `kerrors.NewAggregate` takes care of this for you?

contributing/guide-observability.md
```diff
@@ -120,6 +120,63 @@ logs don't show up in production normally.
...
+
+    Background: one never knows how an error is displayed. Not being deterministic can lead to hotloops
+    with constant API updates, e.g. of a condition.
```
negz, line 176: 👍 FWIW there are concrete examples of this having been an issue for us. For example any AWS errors have unique IDs that change every time the same error occurs, causing constant condition updates.

contributing/guide-observability.md
negz: Happy to unwind the error constant pattern, but let's tackle that in a separate PR. (I'm guessing that's already the intent since this thread is resolved.)

negz (commented):
This looks great. I agree with everything. Only blocker for approving is the comment about wrapping at 80 chars.

contributing/guide-observability.md
tnthornton: >or values which are wrong (and the values are known to be insensitive)

I think the clarification works for me 👍.

Small nit - Your summation above "log it if you know it is not sensitive, e.g. for enum fields, object names, file names, otherwise don't" seems more clear to me versus strictly just "values which are wrong (and the values are known to be insensitive)". However, this is probably more subjective and I don't think it should block on the current progress.

Thanks @sttts !

contributing/guide-observability.md
sttts: it doesn't.

negz: @sttts Could the wrapper you added to crossplane-runtime help? i.e. Perhaps sort the errors by their string value?

contributing/guide-observability.md
```diff
@@ -120,6 +120,63 @@ logs don't show up in production normally.
...
+2. put values into quotes (`%q` in Golang), with the exception of
+   - resource names with namespace or resource type (e.g. `default/foo` or `providers/foo`), resource names alone are quoted
+   - kinds (e.g. `cannot find provider "provider-kubernetes"`; here, the kind provider is without quotes, the provider name is with)
+   - filenames with extensions and/or path (e.g. `cannot parse foo.yaml`).
+3. add a context of e.g. filenames, object names, or values which are wrong (and the values are known to be insensitive).
+
+   Which context is added by which functions depends on the level of abstraction of the code at hand: the package loader will know the OCI image name, the YAML parser might only see a byte slice. The former adds the image name as context, some layer in-between adds the filename and the YAML parser adds the line and column in the byte stream.
+
+   Filenames should be relative to the logical working directory of the context at hand (e.g. a package root, current working directory, workspace root). Ensure that the context is clear in the error message (e.g. `failed to load package provider-kubernetes: failed to parse apis/xrd.yaml`).
```
turkenh, line 154: nit: 
```suggestion
   Filenames should be relative to the logical working directory of the context at hand (e.g. a package root, current working directory, workspace root). Ensure that the context is clear in the error message (e.g. `failed to load package "provider-kubernetes": failed to parse apis/xrd.yaml`).
```

turkenh (approved):

contributing/guide-observability.md
```diff
@@ -120,6 +120,63 @@ logs don't show up in production normally.
...
+- `cannot resolve package dependencies: Invalid Semantic Version`
+- `cannot parse file foo.yaml: failed to parse YAML of file foo.yaml`
+
+Examples for how to improve them:
+- `cannot resolve package dependencies in crossplane.yaml: invalid Semantic Version "main" on package "provider-kubernetes"`
+- `failed to load OCI image provider-gcp/v0.42.0: failed to unpack layer: cannot parse file: failed to parse YAML of file foo.yaml: unexpected indention in line 4`.
+
+General rules:
+1. follow the pattern `generic error cause: more details: even more details`.
+2. put values into quotes (`%q` in Golang), with the exception of
```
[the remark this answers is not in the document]
sttts: Pure cosmetics. Too many quotes make thing less readable. For those exceptions the quotation doesn't add anything.

contributing/guide-observability.md
sttts: The wrapper could help for the multi-error case. But it's equally fine to sort keys of a map before processing each, or use any order that already exists (like in an associative list). Am not sure we want to order by error string in those cases. It might be confusing.

sttts: Maybe the most important learning here is that everything that could return an error ending up somewhere in the API must be deterministic. I.e. every map iteration.

phisco (approved):
LGTM

negz (approved):

PR #4313, opened 2023-07-11
chore: rename error constants according to our standard

phisco:

Just a minor rename of error constants as discussed here to adhere to the `errFmt...` standard we use in the rest of the codebase: https://github.com/crossplane/crossplane/pull/4311#discussion_r1259691340

I have:

### DIFF
--- a/internal/controller/pkg/resolver/reconciler.go
+++ b/internal/controller/pkg/resolver/reconciler.go
@@ -58,12 +58,12 @@ const (
 	errRemoveFinalizer      = "cannot remove lock finalizer"
 	errBuildDAG             = "cannot build DAG"
 	errSortDAG              = "cannot sort DAG"
-	errMissingDependencyFmt = "missing package (%s) is not a dependency"
+	errFmtMissingDependency = "missing package (%s) is not a dependency"
 	errInvalidConstraint    = "version constraint on dependency is invalid"
 	errInvalidDependency    = "dependency package is not valid"
 	errFetchTags            = "cannot fetch dependency package tags"
 	errNoValidVersion       = "cannot find a valid version for package constraints"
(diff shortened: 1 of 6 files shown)


internal/initializer/installer_test.go
```diff
@@ -39,7 +39,7 @@ const (
 	errPatchProviderSourceFmt      = "unexpected source in provider update: %s"
 	errGetConfigurationFmt         = "unexpected name in configuration get: %s"
 	errPatchConfigurationFmt       = "unexpected name in configuration update: %s"
```
lsviben, line 41: There is a lot of errors above with `err...Fmt` should they also be changed to `errFmt...`?

phisco: Good call, regexes 😅

internal/controller/pkg/resolver/reconciler.go
```diff
@@ -58,12 +58,12 @@ const (
 	errRemoveFinalizer      = "cannot remove lock finalizer"
 	errBuildDAG             = "cannot build DAG"
 	errSortDAG              = "cannot sort DAG"
-	errMissingDependencyFmt = "missing package (%s) is not a dependency"
+	errFmtMissingDependency = "missing package (%s) is not a dependency"
```
negz, line 61: No need to tackle it in this PR, but I like to keep the `errFmt` constants in a different block from the plain `err` constants. We've been talking lately about how errors with more context is better, which implies more `errFmt` is likely to be better. Keeping them in two blocks lets you quickly get an idea of how many errors in a file are the "better" `errFmt` style. e.g.:

```go
// Error constants
const (
    errFoo = ""
    errBar = ""

    errFmtFoo = ""
    errFmtBar = ""
)
```

negz (approved):

lsviben (approved):

Issue #4308, opened 2023-07-10
Providers should support K8s resource limits


stevendborrelli:
### What problem are you facing?

Currently when installing a Provider package, Kubernetes [resource limits](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/) are not set. These can be set in the [resources](https://doc.crds.dev/github.com/crossplane/crossplane/pkg.crossplane.io/ControllerConfig/v1alpha1@v1.12.2#spec-resources) of the controller config, but are not set by default.

The crossplane [helm chart ](https://github.com/crossplane/crossplane/blob/4161b8e57432e3010a987ac52c7f7d0f19d88865/cluster/charts/crossplane/values.yaml#L90) sets both request and limits.

Without built-in limits, there are times when provider pod CPU/Memory affects operations on the cluster. For example: 

- https://github.com/upbound/provider-aws/issues/325
- https://github.com/upbound/upjet/issues/116 
-  (there is not an open issue for this yet), provider-kubernetes spikes to 100% when the resources it deploys cannot be deleted due to finalizer of another controller. 

### How could Crossplane help solve your problem?

At the Crossplane level guidance should enforce that providers should come with reasonable default limits, or optionally the package deployment engine should deploy packages pods with requests and limits.

negz:
I'm about to start work on a pertinent design, so I'll take this. 😄

turkenh:
Just a heads up, this could already be achieved using `LimitRange`'s for the namespace where providers are running:

- https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/memory-default-namespace/
- https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/cpu-default-namespace/

kingdonb:
Since this is the top result in search engines for "crossplane provider set requests and limits" I'd just like to mention for anyone else that arrives here and is confused about whether or not providers can have their resources set,

The `ControllerConfig` has a `spec.resources` which you can populate, to set the requests and limits.

The new `DeploymentRuntimeConfig` also has a `deploymentruntimeconfig.spec.deploymentTemplate.spec.template.spec.containers.resources` that you can use, assuming you've moved past the deprecated `ControllerConfig`. You do not need to depend on `LimitRange`, which has the decisive disadvantage that it cannot apply independent resource policies to different pods in the same namespace.

[...]

Issue #2295, opened 2021-05-07
Default Provider Service Account always created when overridden in ControllerConfig 


benagricola:
### What happened?
Configuring `provider-aws` to use an existing IRSA-configured service account using the following `ControllerConfig`:

```yaml
apiVersion: pkg.crossplane.io/v1alpha1
kind: ControllerConfig
metadata:
  name: provider-aws
spec:
  serviceAccountName: crossplane-aws-provider
```

We end up with our existing `crossplane-aws-provider` as well as an [automatically created](https://github.com/crossplane/crossplane/blob/d04a059fe341a941760a3a94ef07085ae7158365/internal/controller/pkg/revision/deployment.go#L40) service account, which then has the `ClusterRoleBinding` applied to allow it to manage resources.

This means we have 1 service account with the right permissions to manage AWS resources and 1 with the right permissions to manage crossplane resources, but zero that can handle both:

```
$ k get sa -n crossplane-system | grep aws
aws-4ebd5d8c7cb7                              1         7m22s  # Crossplane-created with Cluster role binding
crossplane-aws-provider                       1         72m    # eksctl-created with IRSA role binding
```

### How can we reproduce it?
Create a service account manually, use the above `ControllerConfig` to configure a provider and watch as you get 2 service accounts with mixed permissions.

### Suggested fix
Do not create the service account if name is overridden by `ControllerConfig` `serviceAccountName`.

Ideally use the existing service account name when requesting permissions from the RBAC manager. I'm not sure if this is possible as rbac manager uses the `owner` reference to [work out what SA to apply the binding to](https://github.com/crossplane/crossplane/blob/d04a059fe341a941760a3a94ef07085ae7158365/internal/controller/rbac/provider/binding/reconciler.go#L172). When the service account is not created by crossplane, it might be necessary to add an additional check to the binding provider (linked above) to make sure it doesn't try to create a `ClusterRoleBinding` with no subjects.

### What environment did it happen in?
Crossplane version: master on EKS

ONordander:
Hi, could we get any information from maintainers regarding this?
Having the ability to specify the ServiceAccount and get the correct RBAC settings would really help our usecase.

wwentland:
This issue is still present, please reopen.

wwentland:
/fresh

This can still be reproduced, please re-open.

macmiranda:
I'm still seeing the same behavior on 1.9.1

PR #4322, opened 2023-07-13
Package Runtime Config one-pager

negz:

Fixes https://github.com/crossplane/crossplane/issues/2671
Fixes https://github.com/crossplane/crossplane/issues/3601
Fixes https://github.com/crossplane/crossplane/issues/4308
Relates to https://github.com/crossplane/crossplane/pull/4306

This one-pager proposes a lightweight implementation of the PRI proposal, along with an alternative for ControllerConfig. I believe this will be relevant for the in-flight Composition Functions beta design, as well as for Providers.

I have:

### DIFF
--- a/design/one-pager-package-runtime-config.md
+++ b/design/one-pager-package-runtime-config.md
@@ -177,6 +177,34 @@ revision will be used. This pattern will also apply to Deployments and Services,
 simply to make the behaviour of a DeploymentRuntimeConfig more consistent and
 thus less surprising.
 
+## Migration from ControllerConfig
+
+ControllerConfig is an alpha feature. Typically we do not provide an automated
+migration when we drop support for alpha features. ControllerConfig is however a
+special case - it predates our use of feature flags so it's on by default. It's
+also known to be very widely used. Dropping it without a migration story would
+be particularly disruptive.
+
+To ease the migration, we will add a new feature flag, `enable-runtime-config`.
+When true, Crossplane will support `runtimeConfigRef` and __not__
+`controllerConfigRef`. The new `DeploymentRuntimeConfig` type will be introduced
+as an alpha feature, and go through the typical alpha -> beta -> GA lifecycle.
+This means:
+
+1. When first released as alpha, `DeploymentRuntimeConfig` support will be off
+  by default, with `ControllerConfig` support on by default.
+2. When the feature is promoted to beta, `DeploymentRuntimeConfig` support will
+  be on by default, with `ControllerConfig` support off by default. It will
+  still be possible to specify `--enable-runtime-config=false` to force support
+  for `ControllerConfig`.
+3. When the feature is promoted to GA it will no longer be possible to disable
+  support for `DeploymentRuntimeConfig`. Support for `ControllerConfig` will be
+  removed.
+
+To assist with migration, a tool will be provided that automatically creates a
+`DeploymentRuntimeConfig` manifest given a `ControllerConfig` manifest, and
+updates references from a `Provider` manifest accordingly.
+
 ## Future Improvements
 
 The `--package-runtime` flag and `runtimeConfig` API are intentionally designed
--- a/design/one-pager-package-runtime-config.md
+++ b/design/one-pager-package-runtime-config.md
@@ -1,7 +1,7 @@
 # Package Runtime Config
 
 * Owner: Nic Cope (@negz)
-* Reviewers: TODO
+* Reviewers: Hasan Turken (@turkenh), Dan Mangum (@hasheddan)
 * Status: Accepted
 
 ## Background
--- a/design/one-pager-package-runtime-config.md
+++ b/design/one-pager-package-runtime-config.md
@@ -1,0 +1,222 @@
+# Package Runtime Config
+
+* Owner: Nic Cope (@negz)
+* Reviewers: TODO
+* Status: Accepted
+
+## Background
+
+Crossplane Providers are Kubernetes controller managers. They're packaged in an
+OCI container and expect to connect to an API server. You install a Provider
+declaratively by creating a Provider resource. When you do, the Crossplane
+package manager installs your provider by creating a Kubernetes Deployment,
+along with some accoutrements like a ServiceAccount and a Service. The RBAC
+manager also creates some RBAC ClusterRoles and ClusterRoleBindings to allow the
+Provider's ServiceAccount (and thus Deployment) to do what it needs to do.
+
+There are two things we'd like to improve about the way this works today:
+
+1. Folks want more control over how their Providers are deployed - over the
+   configuration of the Deployment (etc) Crossplane creates. ([#3601])
+2. In some cases, folks don't want Crossplane to create a Kubernetes Deployment
+   at all. They want to run a Provider some other way. ([#2671])
+
+Soon [Composition Functions][functions-beta-design] will also be deployed by the
+package manager as Kubernetes Deployments. We expect they'll have similar
+requirements. Since Functions aren't Kubernetes controllers, in this document
+I'll refer to the long-running processes some packages need to deploy as
+'package runtimes'.
+
+Crossplane has a v1alpha1 ControllerConfig type that addresses the first issue
+for Providers. It has been marked deprecated, to be removed if and when we find
+a suitable replacement. We deprecated ControllerConfig because:
+
+* It was growing piecemeal to support templatizing an entire Deployment.
+* We think in some rare cases package runtimes won't use Deployments.
+
+It's worth noting that the desire to deploy a package runtime as anything other
+than a Kubernetes Deployment in the same cluster where Crossplane is running is
+_quite rare_. To my knowledge only Upbound currently has this requirement.
+
+## Goals
+
+The goals of this design are to:
+
+* Give folks full control over package runtime Deployments.
+* Make it possible to run package runtimes as something other than a Deployment.
+
+## Proposal
+
+I propose a new flag to Crossplane: `--package-runtime`. This flag would have
+two possible values (at least to begin with):
+
+* `--package-runtime=Deployment` (default) - Create a Kubernetes deployment.
+* `--package-runtime=Disabled` - Do nothing.
+
+When running in Deployment mode the package manager will function as it does
+today. It will create a Kubernetes Deployment for each package runtime, plus any
+additional supporting configuration such as a ServiceAccount, etc.
+
+When running in Disabled mode the package manager won't create a package runtime
+at all. It will create a revision (e.g. a ProviderRevision) and deliver the
+package's payload (e.g. a Provider's CustomResourceDefinitions), but do nothing
+else. In Disabled mode it's expected that an external controller will take care
+of reconciling the relevant package revision to deploy a package runtime however
+it sees fit.
+
+I also propose we replace ControllerConfig with a new DeploymentRuntimeConfig
+type in the pkg.crossplane.io API group. This type is used to configure the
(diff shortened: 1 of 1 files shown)


design/one-pager-package-runtime-config.md
```diff
@@ -0,0 +1,222 @@
...
+* Give folks full control over package runtime Deployments.
+* Make it possible to run package runtimes as something other than a Deployment.
+
+## Proposal
+
+I propose a new flag to Crossplane: `--package-runtime`. This flag would have
+two possible values (at least to begin with):
+
+* `--package-runtime=Deployment` (default) - Create a Kubernetes deployment.
+* `--package-runtime=Disabled` - Do nothing.
```
negz: Open to better names here, though I think these are fine. What I'm trying to communicate is:

* The default: use typical Kubernetes patterns to deploy the runtime.
* The alternative: let some external (to Crossplane) controller figure out how to deploy the runtime.

design/one-pager-package-runtime-config.md
```diff
@@ -0,0 +1,222 @@
...
+```
+
+The above DeploymentRuntimeConfig matches the values that are currently
+[hardcoded into the package manager][hardcoded-pkg-deployment]. The
+`deploymentTemplate`, etc fields are similar to a Deployment's `template` field.
+
+The package manager will always be opinionated about some things, and will
+overlay the following settings over the top of the provided template:
+
+* The name and namespace of the deployment, service, and service account.
```
negz, line 163: Actually, I think we might need to let folks configure these names. Or at least the service account name to avoid a regression of https://github.com/crossplane/crossplane/issues/2295.

design/one-pager-package-runtime-config.md
```diff
@@ -0,0 +1,222 @@
+# Package Runtime Config
+
+* Owner: Nic Cope (@negz)
+* Reviewers: TODO
```
negz, line 4: Will figure out some explicit reviewers ASAP.

negz: @hasheddan @turkenh do you have bandwidth to review this?

negz: @turkenh confirmed he does. @hasheddan I'm hoping you do too since you're passionate about this topic. LMK if you want me to remove you. 😃

design/one-pager-package-runtime-config.md
```diff
@@ -0,0 +1,222 @@
...
+```yaml
+apiVersion: pkg.crossplane.io/v1
+kind: Provider
+metadata:
+  name: provider-example
+spec:
+  package: xpkg.upbound.io/crossplane-contrib/provider-example:v1.0.0
+  runtimeConfigRef:
+    apiVersion: pkg.crossplane.io/v1
+    kind: DeploymentRuntimeConfig
```
turkenh, line 84:
```suggestion
    apiVersion: runtime.pkg.crossplane.io/v1
    kind: DeploymentConfig
```

just as another idea :)

turkenh (approved):
Looking good to me!

I don't have any major concerns and like the idea of covering 99% of cases in tree without much overhead but leaving the door open for the rest.

It would be nice to also clarify how migration could look like, considering ControllerConfig is widely used despite it being alpha.

design/one-pager-package-runtime-config.md
negz: Haha I almost proposed exactly this, but was thinking perhaps a "flatter" API group namespace would be better. I could go either way on this - curious to hear whether folks have a preference.

design/one-pager-package-runtime-config.md
```diff
@@ -0,0 +1,222 @@
...
+* Make it possible to run package runtimes as something other than a Deployment.
+
+## Proposal
+
+I propose a new flag to Crossplane: `--package-runtime`. This flag would have
+two possible values (at least to begin with):
+
+* `--package-runtime=Deployment` (default) - Create a Kubernetes deployment.
+* `--package-runtime=External` - Do nothing, defer to an external controller.
+
```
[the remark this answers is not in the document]
negz: Do you mean will we put this new package runtime feature behind a feature flag? Good question. I think I could go either way.

The new `--package-runtime` flag by itself doesn't seem like something that would make sense to put behind a feature flag. It essentially controls two things: work broadly how Crossplane already works (make a Deployment), or turn off that functionality.

I could see an argument that we should put the new DeploymentRuntimeConfig behind an alpha feature flag.

On the other hand, you could think of it as a beta iteration of ControllerConfig based on what we've learned from that type. That still implies a feature flag, but at least it would be on by default.

design/one-pager-package-runtime-config.md
```diff
@@ -0,0 +1,222 @@
...
+            runAsGroup: 2000
+            privileged: false
+            allowPrivilegeEscalation: false
+          ports:
+          - name: metrics
+            containerPort: 8080
+    # A DeploymentRuntimeConfig can also be used to configure the Service and
+    # ServiceAccount the package manager creates to support the Deployment.
+    serviceTemplate: {}
+    serviceAccountTemplate: {}
```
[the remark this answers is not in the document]
negz: Yeah, I almost left them out entirely. Including just metadata seems reasonable to me - its easier to add their specs if we find we need to than to remove them.

negz:
> It would be nice to also clarify how migration could look like, considering ControllerConfig is widely used despite it being alpha.

Yeah, good point. I don't think we can break folks using ControllerConfig as we would a typical alpha feature.

It does seem like we could pretty reliably generate a DeploymentRuntimeConfig from a ControllerConfi`. Assuming we land this change in v1.14 I wonder if it would be sufficient to announce ControllerConfig was going to be removed in v1.15. We could supply a tool that automatically converted all ControllerConfigs to DeploymentRuntimeConfigs.

muvaf:
> We could supply a tool that automatically converted all ControllerConfigs to DeploymentRuntimeConfigs.

A controller for `ControllerConfig` that would apply a corresponding `DeploymentRuntimeConfig` resource could be the least breaking way to do it. Then even in the case folks update to a version where `ControllerConfig` does not exist anymore, things would keep working with the remaining `DeploymentRuntimeConfig`s.

design/one-pager-package-runtime-config.md
```diff
@@ -0,0 +1,222 @@
...
+## Goals
+
+The goals of this design are to:
+
+* Give folks full control over package runtime Deployments.
+* Make it possible to run package runtimes as something other than a Deployment.
+
+## Proposal
+
+I propose a new flag to Crossplane: `--package-runtime`. This flag would have
```
jbw976, line 50: because this flag is passed to core crossplane, it would seem that the decision for all packages to use a particular runtime is being made at the time crossplane is deployed - not at the time each package is deployed. Is it possible to mix and match though, e.g. one package uses a deployment and the next package uses something external?

I **think** that may be possible reading further down because this new flag to crossplane more so sets the default runtime for packages that do not have a runtime set.  If that is correct (I'm not sure though, sorry 😇), would it make more sense to call this flag `--package-runtime-default`? 🤔

negz: I was thinking no - it would be a global setting affecting all packages.

design/one-pager-package-runtime-config.md
```diff
@@ -0,0 +1,222 @@
...
+wish to change this to give all runtimes 2 CPU cores by default. Individual
+packages can still be explicitly configured to use a specific runtime config.
+
+Given that we saw ControllerConfig growing into a template for a Deployment, I
+propose we lean into that and make DeploymentRuntimeConfig exactly that. For
+example:
+
+```yaml
+apiVersion: pkg.crossplane.io/v1alpha1
+kind: DeploymentRuntimeConfig
```
ytsarev, line 121: Can we document the migration process of `ControllerConfig` to `DeploymentRuntimeConfig`? Will it be a hard switch, will we have any migration tooling support for conversion?

design/one-pager-package-runtime-config.md
```diff
@@ -0,0 +1,222 @@
...
+When running in Deployment mode the package manager will function as it does
+today. It will create a Kubernetes Deployment for each package runtime, plus any
+additional supporting configuration such as a ServiceAccount, etc.
+
+When running in External mode the package manager won't create a package runtime
+at all. It will create a revision (e.g. a ProviderRevision) and deliver the
+package's payload (e.g. a Provider's CustomResourceDefinitions), but do nothing
+else. In External mode it's expected that an external controller will take care
+of reconciling the relevant package revision to deploy a package runtime however
+it sees fit.
```
hasheddan, line 65: Just double checking -- this is the current expectation, but I don't see anything in this proposal that would prohibit evolution of an interface that notifies an external runtime of the existence of the a new package revision. Is that correct?

design/one-pager-package-runtime-config.md
```diff
@@ -0,0 +1,222 @@
...
+ProviderConfig named default. 
+
+Automatically setting a default runtime config has two advantages:
+
+* The default configuration is no longer [hardcoded][hardcoded-pkg-deployment]
+  into the package manager.
+* Administrators can override the configuration all runtimes use by default.
+
+The Crossplane init container will create a default DeploymentRuntimeConfig at
+install time. A Crossplane administrator could then replace it with their own.
```
hasheddan, line 109: I think this is implied, but could be worth saying explicitly 👍🏻 
```suggestion
The Crossplane init container will create a default DeploymentRuntimeConfig at
install time if one does not exist. A Crossplane administrator could then replace it with their own.
```

design/one-pager-package-runtime-config.md
hasheddan: Actually, reading the alternatives considered section, it sounds like a separate component that reconciled, then notified an external system if needed would be more appropriate with this proposal 👍🏻

design/one-pager-package-runtime-config.md
hasheddan: I was thinking about this as well, primarily in the context of having Crossplane manage functions "close" to itself, while deferring providers to an external system. However, there is a nice escape hatch in this proposal in that you always have an option to just say `--package-runtime=External` and implement whatever you want, which could include replicating the core Crossplane behavior for some packages and not for others.

design/one-pager-package-runtime-config.md
```diff
@@ -0,0 +1,222 @@
...
+1. Folks want more control over how their Providers are deployed - over the
+   configuration of the Deployment (etc) Crossplane creates. ([#3601])
+2. In some cases, folks don't want Crossplane to create a Kubernetes Deployment
+   at all. They want to run a Provider some other way. ([#2671])
+
+Soon [Composition Functions][functions-beta-design] will also be deployed by the
+package manager as Kubernetes Deployments. We expect they'll have similar
+requirements. Since Functions aren't Kubernetes controllers, in this document
+I'll refer to the long-running processes some packages need to deploy as
+'package runtimes'.
```
hasheddan, line 28: nitpick: I would think of the component that deploys and manages the package processes (or configuration of them) as the package runtime. In the default case in this proposal, Kubernetes (Deployments) would be the package runtime rather than each individual Deployment that gets created (e.g. `containerd` is a runtime for containerized processes, the containerized processes themselves are not a runtime)

hasheddan (commented):
This seems like a pragmatic solution to the present desired functionality, without restricting future growth if needed. Thanks @negz! Left a few comments below 👍🏻

phisco (approved):
LGTM 🙏

turkenh (approved):

PR #4323, opened 2023-07-13
Add e2e test for XFN /tmp write access

### LINKED ISSUE #4312: Add e2e tests for xfn /tmp write access

As discussed [here](https://github.com/crossplane/crossplane/pull/4304#pullrequestreview-1524395667), we decided to have two separate PRs for the feature and the e2e test. 

By adding e2e tests for the feature

ezgidemirel:
This PR adds a new e2e test for /tmp folder write access within composition functions. 

Fixes #4312 

I have:


test/e2e/xfn_test.go
```diff
@@ -240,3 +240,155 @@ func TestXfnRunnerImagePull(t *testing.T) {
...
+							helm.WithWait(),
+							helm.WithChart("twuni/docker-registry"),
+							helm.WithVersion("2.2.2"),
+							helm.WithArgs(
+								"--set service.type=NodePort",
+								"--set service.nodePort=32000",
+							),
+						))),
+			).
+			WithSetup("CopyFnImageToRegistry", func(ctx context.Context, t *testing.T, config *envconf.Config) context.Context {
```
phisco, line 271: Given that this is mostly the same as the one used above, it could be worth factoring out this to a dedicated helper function in funcs, something like:
```
func CopyImageToRegistry(image, dstRegistry string, timeout time.Duration) features.Func {
	return func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
		...
	}
}
```
something like

ezgidemirel: There are lots of common steps in the two e2e tests. I'll refactor the code once I verified that the test is working :)

test/e2e/xfn_test.go
```diff
@@ -240,3 +240,155 @@ func TestXfnRunnerImagePull(t *testing.T) {
...
+			Assess("CompositionWithFunctionIsCreated", funcs.AllOf(
+				funcs.ApplyResources(FieldManager, manifests, "composition.yaml"),
+				funcs.ResourcesCreatedWithin(30*time.Second, manifests, "composition.yaml"),
+			)).
+			Assess("ClaimIsCreated", funcs.AllOf(
+				funcs.ApplyResources(FieldManager, manifests, "claim.yaml"),
+				funcs.ResourcesCreatedWithin(30*time.Second, manifests, "claim.yaml"),
+			)).
+			Assess("ClaimBecomesAvailable", funcs.ResourcesHaveConditionWithin(5*time.Minute, manifests, "claim.yaml", xpv1.Available())).
+			Assess("ManagedResourcesProcessedByFunction", func(ctx context.Context, t *testing.T, config *envconf.Config) context.Context {
```
phisco, line 345: Could this be rewritten to use `funcs.ResourceHasFieldValueWithin`? 🤔 Same goes for the `TestXfnRunnerImagePull` test case above probably

test/e2e/xfn_test.go
phisco: Awesome thanks 🙏

test/e2e/testdata/images/tmp-writer/writer.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env sh
+
+touch "/tmp/foo.txt"
```
pedjak, line 3: this test is going to pass even if creation of the file is not possible, because the script does not configure to exit immediatelly

test/e2e/testdata/images/tmp-writer/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM mikefarah/yq:4.34.1
+
+COPY writer.sh /bin
+USER root
+RUN chmod +x /bin/writer.sh
```
pedjak, line 5: you can avoid this line if you set the proper file permission (+x) on `write.sh` within git repo.

phisco: You can also set them from the COPY command directly via the chmod flag: https://docs.docker.com/engine/reference/builder/#copy

test/e2e/xfn_test.go
ezgidemirel: `funcs.ResourceHasFieldValueWithin` requires a `k8s.Object` which is not convenient to use with claims/composites. I'll refactor to create a new helper to use.

test/e2e/testdata/images/tmp-writer/writer.sh
ezgidemirel: thanks for the catch! updated the script

test/e2e/testdata/images/tmp-writer/Dockerfile
ezgidemirel: Thanks guys, I extended the `COPY` command with `chmod`. I like explicitly showing the file permission.

test/e2e/testdata/images/tmp-writer/writer.sh
pedjak: lines 3-6 could be shortened to:

```
touch "/tmp/foo.txt" || exit 1
```

ezgidemirel: done

test/e2e/funcs/feature.go
```diff
@@ -385,6 +390,72 @@ func DeleteResources(dir, pattern string) features.Func {
...
+		}
+
+		return ctx
+	}
+}
+
+// ClaimsManagedResourcesHaveLabel fails a test if the managed resources
+// created by the claim does not have the supplied value at the supplied label
+// path within the supplied duration.
+func ClaimsManagedResourcesHaveLabel(ns, apiVersion, kind, name, label, want string, timeout time.Duration) features.Func {
```
phisco, line 431: nit: do you think it would be possible to make this more generic similar to `ResourcesHaveFieldValueWithin` below?
I would imagine something like:
```golang
func ClaimsManagedResourcesHaveFieldValueWithin(d time.Duration, ... something to specify the parent Claim ..., ... something to filter the managed resources ..., path string, want any) features.Func {
...
```
The other method is plural, so I kept it plural, but here the plural could be in the managed resources, instead of the claim 🤔 does it make sense?

test/e2e/funcs/feature.go
```diff
@@ -385,6 +390,72 @@ func DeleteResources(dir, pattern string) features.Func {
 	}
 }
 
+// CopyImageToRegistry tries to copy the supplied image to the supplied registry within the timeout
+func CopyImageToRegistry(clusterName, ns, sName, image string, timeout time.Duration) features.Func {
+	return func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
+		reg, err := ServiceIngressEndPoint(ctx, c, clusterName, ns, sName)
```
pedjak, line 396: perhaps we should detect if registry is available and if not fail with a meaningful message?

test/e2e/xfn_test.go
```diff
@@ -227,3 +178,87 @@ func TestXfnRunnerImagePull(t *testing.T) {
 			Feature(),
 	)
 }
+
+func TestXfnRunnerWriteToTmp(t *testing.T) {
```
pedjak, line 182: @phisco could we have a common setup for these two tests, so that we can execute them without going through two setup and teardown rounds? e2e framework should support that through Table tests?

phisco: talking with @ezgidemirel we saw that the two setups are slightly different. Given that we agreed to try to stick to having "1 Test, 1 feature" as much as possible, for the moment I'd keep them distinct, but we should definitely revisit it in the near future as we add more and more tests to reduce times. WDYT?

phisco (approved):
Thanks @ezgidemirel! LGTM! Let me know if you are planning to address @pedjak's [comment](https://github.com/crossplane/crossplane/pull/4323/files#r1268094477) or you want me to merge this. 🙏

ezgidemirel:
> Thanks @ezgidemirel! LGTM! Let me know if you are planning to address @pedjak's [comment](https://github.com/crossplane/crossplane/pull/4323/files#r1268094477) or you want me to merge this. 🙏

Hi @phisco, we discussed with @pedjak offline. I decided to improve the error messages in the `ServiceIngressEndPoint` function and I believe this should be enough. 

Please go ahead and merge :)

PR #4358, opened 2023-07-20
fix: max size of package parsed limited to 200MB

phisco:

Limit the size of the package parsed to ~100~200MBs.

I have:

### DIFF
--- a/internal/controller/pkg/revision/reconciler.go
+++ b/internal/controller/pkg/revision/reconciler.go
@@ -54,6 +54,8 @@ import (
 
 const (
 	reconcileTimeout = 3 * time.Minute
+	// the max size of a package parsed by the parser
+	maxPackageSize = 200 << 20 // 100 MB
 )
 
 const (
@@ -474,7 +476,13 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 	}
 
 	// Parse package contents.
-	pkg, err := r.parser.Parse(ctx, rc)
+	pkg, err := r.parser.Parse(ctx, struct {
+		io.Reader
+		io.Closer
+	}{
+		Reader: io.LimitReader(rc, maxPackageSize),
+		Closer: rc,
+	})
 	// Wait until we finish writing to cache. Parser closes the reader.
 	if err := <-cacheWrite; err != nil {
 		// If we failed to cache we want to cleanup, but we don't abort unless


internal/controller/pkg/revision/reconciler.go
```diff
@@ -54,6 +54,8 @@ import (
 
 const (
 	reconcileTimeout = 3 * time.Minute
+	// the max size of a package parsed by the parser
+	maxPackageSize = 100 << 20 // 100 MB
```
jbw976, line 58: did you already take a look at the size of the biggest packages already in https://marketplace.upbound.io/? curious as to how close we already are to this limit.  with a full controller environment for providers, i can imagine we're not that far away from already bumping into this 🤔

internal/controller/pkg/revision/reconciler.go
```diff
@@ -474,7 +476,13 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 	}
 
 	// Parse package contents.
-	pkg, err := r.parser.Parse(ctx, rc)
+	pkg, err := r.parser.Parse(ctx, struct {
```
jbw976, line 479: do you have a quick pointer to this golang syntax here? we're replacing `rc` with an anonymous struct that is composed of two anonymous fields that fit the shape of 2 interfaces....it makes sense, but just curious to read about this specific idiom/syntax and what their official nomenclature is 😂

internal/controller/pkg/revision/reconciler.go
phisco: what do you mean by "full controller environment for providers"

internal/controller/pkg/revision/reconciler.go
phisco: not sure there is a name for it, it's just a shorthand to avoid having to declare a type "LimitReaderCloser" used only here. We could argue whether it would be more idiomatic to pass a Reader in and deal with the Close part from the caller directly, but I didn't see the reason to change the code so much just for this.

internal/controller/pkg/revision/reconciler.go
phisco: BTW, not sure it's clear, but it's a 100MB of YAMLs, not the size of the whole image

phisco: wrote a small test checking the size for a few images: 

- xpkg.upbound.io/upbound/platform-ref-aws:v0.6.0: 0.04052734375MB
- xpkg.upbound.io/upbound/platform-ref-aws:latest: 0.03369140625MB
- xpkg.upbound.io/upbound/platform-ref-azure:v0.7.0: 0.03173828125MB
- xpkg.upbound.io/upbound/platform-ref-azure:latest: 0.02197265625MB
- xpkg.upbound.io/upbound/platform-ref-gcp:v0.4.0: 0.02734375MB
- xpkg.upbound.io/upbound/platform-ref-gcp:latest: 0.0224609375MB

As you can see all of them are actually 30KB, so well below the limit of 100MB

turkenh (approved):

internal/controller/pkg/revision/reconciler.go
phisco: Monolithic providers are indeed quite big, still way below 100MB: 
- xpkg.upbound.io/upbound/provider-aws:v0.37.0: 25.2275390625MB, annotation: true
- xpkg.upbound.io/upbound/provider-gcp:v0.34.0: 11.8583984375MB, annotation: true
- xpkg.upbound.io/upbound/provider-azure:v0.34.0: 23.216796875MB, annotation: true

we could maybe make this configurable via a flag, WDYT @jbw976 @turkenh?

internal/controller/pkg/revision/reconciler.go
jbw976: was just [looking it up](https://golangbyexample.com/embedding-interfaces-go/) now to refresh myself on the terminology - i think the right terminology here is that we're using an anonymous struct with embedded interfaces as unnamed/anonymous fields 🤓

internal/controller/pkg/revision/reconciler.go
jbw976: Is that the size of just the yaml layer @phisco or is that the entire OCI image?

I'd prefer to keep the complexity and moving parts here very slim and not add a configurable flag/machinery 🤓 

to give us a bit more room though, maybe we can bump the max value to something a bit higher, e.g. 256MB? what is the value at which we'd actually be concerned about a denial of service attack happening?

You can make the final call here @phisco, and @turkenh has already approved ✅

jbw976 (approved):

internal/controller/pkg/revision/reconciler.go
phisco: YAML layer only, I'll increase this to 200MB just to play safe and merge it 👍

Issue #4351, opened 2023-07-19
Support using jq for highly-expressive composition transformations


jutley:
### What problem are you facing?

I frequently find that the transformations offered by Crossplane are too crude for use cases that seem rather simple. The example I am running into now is that I would like to pass labels with a specific prefix onto the AWS resources created by a transformation. Using the prefix `example.com`, we may want to turn the following Kubernetes resource labels into the following AWS tags:

```yaml
# Kubernetes labels
example.com/foo: foo
example.com/bar: bar
something: else  # just to illustrate filtering
```

```yaml
# AWS tags
- key: example.com/foo
  value: foo
- key: example.com/bar
  value: bar
```

### How could Crossplane help solve your problem?

I would love to see Crossplane adopt some mechanism for highly-expressive transformations. I am particularly fond of jq, which has an [implementation written in golang](https://github.com/itchyny/gojq) – perfect for Crossplane.

For this use case, my patch may look like this:

```yaml
patches:
  - type: FromCompositeFieldPath
    fromFieldPath: metadata.labels
    toFieldPath: spec.forProvider.tags
    transforms:
      - type: jq
        jq:
          jqExpression: 'to_entries | map(select(.key | startswith("example.com/")))'
```

This is a rather simple expression and very effectively handles our use case.

We could alternatively build out functionality like this using a Composition Function, but I think this proposal provides a way that users could get a lot of power out of the box without needing to learn how Composition Functions work.

turkenh:
This makes sense, however, I would rather prefer to revisit this after composition functions are promoted to beta with the updated design. I have a hunch that it will be a lot easier to achieve this with composition functions compared to the current state (i.e. alpha, running with xfn runner, etc).

bobh66:
[json-e](https://json-e.js.org/) is another useful tool - I think there may be a generic case where a user may want to write their compositions with formatting from tool X (jq, json-e, JMESPath, yq, etc) embedded in it, and then run that through a composition function to process all of the references.  I discussed the possibility of running Composition Functions at various points in the composition process with @negz and it's something we might consider after the current iteration goes through.

pedjak:
@jutley - [there is a similar idea](https://github.com/crossplane/crossplane/issues/2772#issuecomment-1589405740) - if we create a generic function - a wrapper around `yq`, we could do achieve the same functionalities, please have a look.

jutley:
@pedjak That proposal could work, but I do want to make the argument to _not_ use a Crossplane Function.

[...]

PR #4366, opened 2023-07-21
Generic cross-reference design doc

### LINKED ISSUE #1770: Generic Cross-Resource References

pedjak:
Generic cross-reference design doc

Fixes #1770

I have:

### DIFF
--- a/design/design-generic-references.md
+++ b/design/design-generic-references.md
@@ -7,9 +7,9 @@
 ## Background
 
 Crossplane providers provision and monitor external resources in third party
-cloud providers, following the declared specification in the corresponding managed
-resources. Very often, creation of these objects cannot happen in parallel, i.e. some
-fields can be set only after a given dependency is created.
+cloud providers, following the declared specification in the corresponding
+managed resources. Very often, creation of these objects cannot happen in
+parallel, i.e. some fields can be set only after a given dependency is created.
 
 [Cross resource referencing] solves this problem for majority of cases by
 following [Kubernetes API object reference] convention:
@@ -20,26 +20,48 @@ spec:
     name: bar
 ```
 
-Field `spec.fooRef` contains the name of the resource where the value for `spec.foo` is to
-be found. The kind of referred resource and the field containing the value used later for
-setting `spec.foo` is determined during compile time.
+Field `spec.fooRef` contains the name of the resource where the value
+for `spec.foo` is to be found. The kind of referred resource and the path to the
+field containing the value used later for setting `spec.foo` is fixed, i.e. it
+is embedded in the resolving mechanism used by the provider.
 
-Although the used convention covers the majority of use cases, it lacks the
-following properties:
+Although the used convention covers the majority of use cases, user experience
+shows that there is still a need to refer a value on an arbitrary path within an
+object of arbitrary kind. For example, 
+[`datasync.aws.upbound.io` `Task`](https://github.com/upbound/provider-aws/blob/main/apis/datasync/v1beta1/zz_task_types.go#L375)
+has [`spec.forProvider.sourceLocationArn`](https://github.com/upbound/provider-aws/blob/main/apis/datasync/v1beta1/zz_task_types.go#L359)
+field that can obtain value from an instance of the following object types:
 
-* The referred object kind and the source field cannot be changed at 
-  runtime for a given object
-* The referencer field may not always target a single referred kind. For
-  example, a URL field in Route53 could get populated by a URL of an S3 bucket
-  or access endpoint of an RDS Instance
-* The referencer field may not always target a single field in the referred
-  object
-* The value that is about to be assigned to the field require some sort of
-  transformation
+* LocationS3
+* LocationSmb
+* LocationObjectStorage
+* LocationNfs
+* LocationHdfs
+* LocationFsxWindowsFileSystem
+* LocationFsxOpenzfsFileSytem
+* LocationFsxLustreFileSystem
+* LocationEfs
 
-Some of the above issues can be solved by defining a composition. However, there
-are cases where one would like to use and refer to a resource not bound by a 
-composite instance.
+However, [the implemented provider resolving mechanism](https://github.com/upbound/provider-aws/blob/main/apis/datasync/v1beta1/zz_generated.resolvers.go#L104-L112)
+looks only for the value in `LocationS3` instance named
+by `spec.forProvider.sourceLocationArnRef` field.
+
+Currently, the issue can be solved by defining the following composition:
+
+* Declare under managed resources the needed source object (one of `Location*`)
+  types with [observe only management policy]
+* Patch composite object with the value from the source object
+* Patch Task's `spec.forProvider.sourceLocationArn` field with the composite
+  value
+
+The proposed solution has the following drawbacks:
+
+* For each claim, additional observe only resource gets created, just to be able
+  to work around the limitations of the current reference mechanism on certain
+  fields. In a system with a high number of claims, that would put additional
+  pressure on k8s API and etcd.
+* Composite object needs to expose the patched value, although fully unneeded
+  from API design perspective
 
 __**NOTE**__: [a prior design exists on this topic](https://github.com/crossplane/crossplane/pull/2385).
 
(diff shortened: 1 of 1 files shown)


design/design-generic-references.md
```diff
@@ -0,0 +1,314 @@
...
+
+## Background
+
+Crossplane providers provision and monitor external resources in third party
+cloud providers, following the declared specification in the corresponding managed
+resources. Very often, creation of these objects cannot happen in parallel, i.e. some
+fields can be set only after a given dependency is created.
+
+[Cross resource referencing] solves this problem for majority of cases by
+following [Kuberentes API object referemce] convention:
```
jeanduplessis, line 15:
```suggestion
following [Kuberentes API object reference] convention:
```

design/design-generic-references.md
```diff
@@ -0,0 +1,314 @@
...
+
+```yaml
+spec:
+  fooRef:
+    name: bar
+```
+
+Field `spec.fooRef` contains the name of resource where the value for `spec.foo` is to
```
jeanduplessis, line 23:
```suggestion
Field `spec.fooRef` contains the name of the resource where the value for `spec.foo` is to
```

design/design-generic-references.md
```diff
@@ -0,0 +1,314 @@
...
+be found. The kind of referred resource and the field containing the value used later for
+setting `spec.foo` is determined during compile time.
+
+Although the used convention covers the majority of use cases, it lacks the
+following properties:
+
+* The referred object kind and the source field cannot be tuned/changed at 
+  runtime for a given object
```
jeanduplessis, line 31: Question: Purely for my own understanding, what would the difference be between "tuning" and "changing" the kind/source field at runtime?

design/design-generic-references.md
```diff
@@ -0,0 +1,314 @@
...
+  example, a URL field in Route53 could get populated by a URL of an S3 bucket
+  or access endpoint of an RDS Instance
+* The referencer field may not always target a single field in the referred
+  object
+* The value that is about to be assigned to the field require some sort of
+  transformation
+
+Some of the above issues can be solved by defining a composition. However, there
+are cases where one would like to use and refer a resource not bound by a 
+composite instance.
```
jeanduplessis, line 42:
```suggestion
Some of the above issues can be solved by defining a composition. However, there
are cases where one would like to use and refer to a resource not bound by a 
composite instance.
```

design/design-generic-references.md
```diff
@@ -0,0 +1,314 @@
...
+
+__**NOTE**__: there is 
+[a previous work on this topic](https://github.com/crossplane/crossplane/pull/2385).
```
jeanduplessis, line 45:
```suggestion
__**NOTE**__: [a prior design exists on this topic](https://github.com/crossplane/crossplane/pull/2385).
```

design/design-generic-references.md
```diff
@@ -0,0 +1,314 @@
...
+
+## Goals
+
+The generic cross-reference should be able to:
+
+* be configurable at the runtime
```
jeanduplessis, line 51:
```suggestion
* be configurable at runtime
```

design/design-generic-references.md
```diff
@@ -0,0 +1,314 @@
...
+* use the existing [Kuberentes API object referemce] convention
```
jeanduplessis, line 52:
```suggestion
* use the existing [Kuberentes API object reference] convention 
```

design/design-generic-references.md
```diff
@@ -0,0 +1,314 @@
...
+[a previous work on this topic](https://github.com/crossplane/crossplane/pull/2385).
+
+## Goals
+
+The generic cross-reference should be able to:
+
+* be configurable at the runtime
+* use the existing [Kuberentes API object referemce] convention 
+  as much as possible
+* provide alternative convention in cases where [Kuberentes API object referemce]
```
jeanduplessis, line 54:
```suggestion
* provide alternative convention in cases where [Kuberentes API object reference]
```

design/design-generic-references.md
```diff
@@ -0,0 +1,314 @@
...
+  convention is not suitable
+* be implemented without or light changes in RBAC
```
jeanduplessis, line 56:
```suggestion
* be implemented without or requiring only light changes in RBAC
```

design/design-generic-references.md
```diff
@@ -0,0 +1,314 @@
...
+* support referring multiple objects (namespaced and cluster-scoped) and their
+  fields
+* support value transformation before assignment
+
+## Proposal
+
+We would like to introduce a new cluster-scoped `Referable` (alternative name
+could be `InjectableValue`) type whose instances declare values that can be
+assigned to object field:
```
jeanduplessis, line 65:
```suggestion
assigned to an object field:
```

design/design-generic-references.md
```diff
@@ -0,0 +1,314 @@
...
+* provide alternative convention in cases where [Kuberentes API object referemce]
+  convention is not suitable
+* be implemented without or light changes in RBAC
+* support referring multiple objects (namespaced and cluster-scoped) and their
+  fields
+* support value transformation before assignment
+
+## Proposal
+
+We would like to introduce a new cluster-scoped `Referable` (alternative name
```
jeanduplessis, line 63: Could you maybe touch on why this should be cluster-scope and not namespace-scoped?

design/design-generic-references.md
```diff
@@ -0,0 +1,314 @@
...
+```
+
+Now, the value can be referred in the usual way within a managed resource:
+
+```yaml
+spec:
+  vpcIdRef: referable-vpc-id
+```
+
+The reference resolver could first try to find `referable-vpc-id` `VPC` instance,
```
jeanduplessis, line 113:
```suggestion
The reference resolver could first try to find a `referable-vpc-id` `VPC` instance, 
```

design/design-generic-references.md
```diff
@@ -0,0 +1,314 @@
...
+
+```yaml
+spec:
+  vpcIdRef: referable-vpc-id
+```
+
+The reference resolver could first try to find `referable-vpc-id` `VPC` instance, 
+and fallback to `referable-vpc-id` `Referable` if the former does not exist. 
+If the instance is ready, the value is read from `status.value` and 
+set to `spec.vpcId` of the managed resource. Ofcourse, if needed, and if it
```
jeanduplessis, line 116:
```suggestion
set to `spec.vpcId` of the managed resource. Of course, if needed, and if it 
```

jeanduplessis: Under what circumstances might it be desired to look for the `Referable` object first?

design/design-generic-references.md
```diff
@@ -0,0 +1,314 @@
...
+```
+
+The reference resolver could first try to find `referable-vpc-id` `VPC` instance, 
+and fallback to `referable-vpc-id` `Referable` if the former does not exist. 
+If the instance is ready, the value is read from `status.value` and 
+set to `spec.vpcId` of the managed resource. Ofcourse, if needed, and if it 
+makes more sense, the resolver could first look for the existence of `Referable` 
+object first.
+
+Referring a namespaced object would be possible as well:
```
jeanduplessis, line 120:
```suggestion
Referring to a namespaced object would be possible as well:
```

design/design-generic-references.md
```diff
@@ -0,0 +1,314 @@
...
+spec:
+  objects:
+    - apiVersion: v1
+      kind: ConfigMap
+      name: common-settings
+      namespace: crossplane-system
+      fieldPath: data.region
+```
+
+or we could match an object with a label selector, if its name is unknown or not
```
jeanduplessis, line 136:
```suggestion
or we could match an object with a label selector if its name is unknown or not
```

design/design-generic-references.md
```diff
@@ -0,0 +1,314 @@
...
+    - apiVersion: ec2.aws.crossplane.io/v1alpha1
+      kind: VPC
+      matchingLabels:
+        class: main-vpc
+      fieldPath: status.atProvider.id
+```
+
+### Value Assigning Without Counterpart Ref Field
+
+If a value needs to be assigned to a field that does not have
```
jeanduplessis, line 155:
```suggestion
If a value needs to be assigned to a field that does not have a
```

design/design-generic-references.md
```diff
@@ -0,0 +1,314 @@
...
+corresponding `Ref` field, the reference can be expressed using `spec.refs` block:
+
+```yaml
+spec:
+  refs:
+    - name: referable-vpc-id
+      toFieldPath: spec.myVPCId
+```
+
+The syntax of `toFieldPath` field is similar to what we
```
jeanduplessis, line 165:
```suggestion
The syntax of the `toFieldPath` field is similar to what we
```

design/design-generic-references.md
```diff
@@ -0,0 +1,314 @@
...
+use in Composition, in line [Kubernetes API Conventions on field selection].
```
jeanduplessis, line 166:
```suggestion
use in Composition, in line with the [Kubernetes API Conventions on field selection].
```

design/design-generic-references.md
```diff
@@ -0,0 +1,314 @@
...
+
+Alternatively, if we would like to avoid changing managed resource schema, the
```
jeanduplessis, line 168:
```suggestion
Alternatively, if we would like to avoid changing the managed resource schema, the
```

design/design-generic-references.md
```diff
@@ -0,0 +1,314 @@
...
+above can be stated using annotations as well:
+
+```yaml
+metadata:
+  annotations:
+    "referable.upbound.io/referable-vpc-id": "spec.myVPCId"
+```
+
+### Referring Value Existing Potentially in Multiple Source
```
jeanduplessis, line 177: It might be useful to refer to a real-world example that might lead to this scenario.

design/design-generic-references.md
```diff
@@ -0,0 +1,314 @@
...
+      kind: Bar
+      name: bar
+      optional: true
+      fieldPath: spec.myId
+```
+
+The above syntax can be used as well when a value might appear in a several
+places within single object.
+
+### Value Transformation
```
jeanduplessis, line 204: It might be useful to refer to a real-world example that might lead to this scenario.

design/design-generic-references.md
```diff
@@ -0,0 +1,314 @@
...
+The used templating engine in `mapping.value` is [Go templates]. Each object 
+reference with assigned `id` becomes available under that name in the template context.
+A number of useful transformation functions could be made available to that context.
+
+### Advantages
+
+* All referred values are declared on managed resources using already familiar
+  mechanism, keeping their discovery simple
+* Value extraction is detached from its consumption, same `Referable` can be
+  used by many multiple manged resources
```
jeanduplessis, line 240:
```suggestion
* Value extraction is detached from its consumption. The same `Referable` can be
  used by multiple managed resources
```

jeanduplessis (changes requested):

design/design-generic-references.md
```diff
@@ -0,0 +1,314 @@
...
+  well
+* `Referable` instances can be properly garbage collected if owner
+  or [`Usage`](https://github.com/crossplane/crossplane/blob/master/design/one-pager-generic-usage-type.md)
+  instances are declared
+* Multiple objects/fields can be referred as value sources
+* Value can be transformed using a Go template
+
+### Disadvantages
+
+* Providers might need to get read access to `Referable` type
```
jeanduplessis, line 251: Could we expand on why this is the case?

design/design-generic-references.md
```diff
@@ -0,0 +1,314 @@
...
+
+### Implementation
+
+It consists of two parts:
+
+* Adding new Crossplane controller for `Referable` type
```
jeanduplessis, line 257:
```suggestion
* Adding new Crossplane controller for the `Referable` type
```

design/design-generic-references.md
```diff
@@ -0,0 +1,314 @@
...
+* Update [controller-runtime](https://github.com/crossplane/crossplane-runtime/)
+  to support usage of `Referable` instances
```
jeanduplessis, line 259:
```suggestion
* Update [crossplane-runtime](https://github.com/crossplane/crossplane-runtime/)
  to support usage of `Referable` instances
```

design/design-generic-references.md
```diff
@@ -0,0 +1,314 @@
...
+The controller is responsible for:
+
+* Defining proper `Usage` instance to guard against improper deletion
+  of `Referable` instances
+* Retrieving the declared objects and the field value, based on the provided
+  references
+* Transforming the value (if requested) and exposing it under `status.value`
+* Marking the instance as ready
+
+We assume currently that the value does not change after it gets exposed.
```
jeanduplessis, line 272: What would happen if the value changes?

design/design-generic-references.md
```diff
@@ -0,0 +1,314 @@
...
+
+No additional RBAC rules are needed, since Crossplane already poses very broad
+permissions within the cluster.
```
jeanduplessis, line 275:
```suggestion
No additional RBAC rules are needed since Crossplane already defines broad
permissions within the cluster.
```

design/design-generic-references.md
```diff
@@ -0,0 +1,314 @@
...
+
+#### Controller-runtime Changes
```
jeanduplessis, line 277:
```suggestion
#### Crossplane-runtime Changes
```

Is it intentionally meant to be controller-runtime here?

design/design-generic-references.md
```diff
@@ -0,0 +1,314 @@
...
+
+* [APIResolver](https://github.com/crossplane/crossplane-runtime/blob/master/pkg/reference/reference.go#L280)
+  should be enriched to support `Referable` instances
+* If additional references need to be declared via `spec.refs` block or
+  annotations, [ResolveReferences function generator](https://github.com/crossplane/crossplane-tools/blob/master/internal/method/resolver.go#L33)
+  should be enriched to support them
```
jeanduplessis, line 283:
```suggestion
* If additional references need to be declared via the `spec.refs` block or
  annotations, the [ResolveReferences function generator](https://github.com/crossplane/crossplane-tools/blob/master/internal/method/resolver.go#L33)
  should be enriched to support them
```

design/design-generic-references.md
```diff
@@ -0,0 +1,314 @@
...
+* After fetching values, proper `Usage` instance need to be defined to guard
+  against improper deletion
+
+Finally, providers need to be upgraded to use the new version of controller-runtime 
+and each type `ResolveReference` function needs to be regenerated.
```
jeanduplessis, line 288:
```suggestion
Finally, providers need to be upgraded to use the new version of crossplane-runtime 
and for each type, the `ResolveReference` function needs to be regenerated.
```

design/design-generic-references.md
```diff
@@ -0,0 +1,314 @@
...
+
+## Alternatives Considered
+
+* There is [a previous work on this topic](https://github.com/crossplane/crossplane/pull/2385)  
+  It embeds the schema similar `Referable` type into a number of `spec.patches.fromObject`
+  fields in managed resource. Such approach would require that each provider 
+  has access to broad set of resources (even from other providers), demanding
+  a broad set of RBAC rules. Furthermore, if a certain value transformation is
+  needed on multiple managed resources, it would be required to repeat its 
+  declaration on every MR.
```
jeanduplessis, line 298:
```suggestion
* [A previous design](https://github.com/crossplane/crossplane/pull/2385) was proposed for this topic.
  It embeds the schema similar `Referable` type into a number of `spec.patches.fromObject`
  fields in managed resources. Such an approach would require that each provider 
  has access to a broad set of resources (even from other providers), demanding
  a broad set of RBAC rules. Furthermore, if a certain value transformation is
  needed on multiple managed resources, it would be required to repeat its 
  declaration on every MR.
```

design/design-generic-references.md
```diff
@@ -0,0 +1,314 @@
...
+
+* `Referable` instances could also contain the reference 
+   to destination managed resource and its field.
```
jeanduplessis, line 301:
```suggestion
* `Referable` instances could also contain the reference 
   to the destination managed resource and its field.
```

design/design-generic-references.md
```diff
@@ -0,0 +1,314 @@
...
+    - Such approach would make the discovery of references for a given manage
+      resource harder
```
jeanduplessis, line 303:
```suggestion
    - Such an approach would make the discovery of references for a given managed
      resource harder
```

design/design-generic-references.md
```diff
@@ -0,0 +1,314 @@
...
+    - Referable controller logic becomes more complex, because the destination
+      managed resource might not exist
+    - One more controller can now update a managed resource, increasing
+      potential conflict rate
+    - `Referable` instance could not be consumed by multiple managed resources
+    - However, providers would not need the read access to `Referable` instances
+
+[Cross resource referencing]: https://github.com/crossplane/crossplane/blob/master/design/one-pager-cross-resource-referencing.md
+[Kuberentes API object referemce]: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#object-references
```
jeanduplessis, line 312:
```suggestion
[Kubernetes API object reference]: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#object-references
```

jeanduplessis (changes requested):

design/design-generic-references.md
pedjak: actually, they mean the same... thanks.

design/design-generic-references.md
pedjak: simply because all our types are cluster-scoped, iirc.

design/design-generic-references.md
pedjak: sure, @turkenh or @haarchri do you have some better example from the field?

design/design-generic-references.md
pedjak: sure, @turkenh or @haarchri do you have some better example from the field?

design/design-generic-references.md
pedjak: Allowing updates and propagating them to the managed resources is out
of the scope for this proposal version.

design/design-generic-references.md
```diff
@@ -0,0 +1,319 @@
...
+
+```yaml
+spec:
+  fooRef:
+    name: bar
+```
+
+Field `spec.fooRef` contains the name of the resource where the value for `spec.foo` is to
+be found. The kind of referred resource and the field containing the value used later for
+setting `spec.foo` is determined during compile time.
```
sttts, line 25: Had to read a lot of what "compile time" means here. From https://github.com/crossplane/crossplane/blob/master/design/one-pager-cross-resource-referencing.md I figure this rather means "specified in provider code".

design/design-generic-references.md
```diff
@@ -0,0 +1,319 @@
...
+* be implemented without or requiring only light changes in RBAC
+* support referring multiple objects (namespaced and cluster-scoped) and their
+  fields
+* support value transformation before assignment
+
+## Proposal
+
+We would like to introduce a new cluster-scoped `Referable` (alternative name
+could be `InjectableValue`) type whose instances declare values that can be
+assigned to an object field:
```
sttts, line 64: alternative: name it by what these things are namely `FieldReferences`, `ObjectReferences`. Naming something by how it can be used is a little strange.

design/design-generic-references.md
```diff
@@ -0,0 +1,319 @@
...
+* If additional references need to be declared via the `spec.refs` block or
+  annotations, the [ResolveReferences function generator](https://github.com/crossplane/crossplane-tools/blob/master/internal/method/resolver.go#L33)
+  should be enriched to support them
+* After fetching values, proper `Usage` instance need to be defined to guard
+  against improper deletion
+
+Finally, providers need to be upgraded to use the new version of crossplane-runtime 
+and for each type, the `ResolveReference` function needs to be regenerated.
+
+## Alternatives Considered
```
sttts, line 293: - this proposal seems to make the same mistakes that the composition patch and transformation (P&T) has done, and that eventually led to the invention of functions:
  - desirable transformations are open ended and are never enough
  - patching by field paths is a very low-level mechanism, operating on low-level data-type rather than being MR kind aware
  - references don't solve the order problem: (a) no way to order without the MR author having prepared for a reference field (b) dependencies might be more than just a field in an object (think of probes).
- in general, this proposal seems to clone the P&T idea into the object system. Do we really want that? Then the new approach competes with compositions. This is kind of the goal stated in the beginning of the doc. I miss a discussion why this is an acceptable goal, i.e. a motivation section.
- the core motivation behind this proposal seems to be that compositions are too slow (e.g. not watch-driven, but run in 60s interval). Maybe we should fix the latter.

design/design-generic-references.md
```diff
@@ -0,0 +1,319 @@
...
+
+Field `spec.fooRef` contains the name of the resource where the value for `spec.foo` is to
+be found. The kind of referred resource and the field containing the value used later for
+setting `spec.foo` is determined during compile time.
+
+Although the used convention covers the majority of use cases, it lacks the
+following properties:
+
+* The referred object kind and the source field cannot be changed at 
+  runtime for a given object
```
sttts, line 31: this could be solved by being explicit about the resource rather than resolving with some magic in the background

design/design-generic-references.md
```diff
@@ -0,0 +1,319 @@
...
+  example, a URL field in Route53 could get populated by a URL of an S3 bucket
+  or access endpoint of an RDS Instance
+* The referencer field may not always target a single field in the referred
+  object
+* The value that is about to be assigned to the field require some sort of
+  transformation
+
+Some of the above issues can be solved by defining a composition. However, there
+are cases where one would like to use and refer to a resource not bound by a 
+composite instance.
```
sttts, line 42: Compositions today mix specification, relationships and orchestration:
1. **specification**: specify how objects look like
2. **relationships**: Patch&Transformations (P&T) establish field level dependencies
3. **orchestration**: P&T is used to define order, both by (a) implicit field dependencies of required fields and (b) by hacks to intentionally break objects or pause objects before some condition is true.

In that light, I would consider the goal of this proposal to promote (2) to the object system, and primarily leave compositions for (1) and (3). This bigger "philosophical" context must be clearer in this doc, maybe in some motivation section.

There is also a connection between (2) and (3a). An asynchronous controller cannot solve the required fields problem, i.e. late initialization. This means that we encourage late initialization more and more. Late initialization has the problem of more states to reason about. Resources will not be completely or correctly specified fromt he beginning, i.e. the creation of the object. This obviously has consequences: less invariants/consistency, less type-safety. 

Is that ^^ a higher level goal? Isn't being more dynamic, less consistent, less type-safety a step into the wrong direction? Thinking about DevEx and "Crossplane is complex".

design/design-generic-references.md
phisco: +1

design/design-generic-references.md
phisco: Yes, a more complex example would be nice to see, not necessarily a real world example, but at least a bigger toy example would be helpful. I'm not sure how you are envisioning this new API to be used in the context of Compositions. Should users define this new resource as part of their Composition? How will that look like? So users will most probably have to define `Usage` resources as defined by https://github.com/crossplane/crossplane/pull/4215 to achieve deletion ordering and then this new object to define mappings between values? how's that going to look like?

design/design-generic-references.md
phisco: +1
This feels like it's adding one more layer of indirection to Compositions and feels overlapping with both the `Usage` resource introduced by https://github.com/crossplane/crossplane/pull/4215 and a little bit with observe only resources too.

turkenh: It would be good to define and agree on the problem before talking about the solution.

Let's take the `VPC` / `Subnet` example. If someone wants to create a `Subnet` in a `VPC` created/managed by XP, they need a mechanism to read the `id` from `VPC` and put it into the `Subnet` spec. Today, this is possible in one of the following two ways:

**a.** Using cross-resource references where they can either use MR name for `VPC` (e.g. `vpcIdRef.name`) or labels/controllerRef (e.g. `vpcIdRef.selector`).
**b.** Using composition patches, but with the following limitations:
  - `VPC` and `Subnet` must exist in the same composition.
  - `id` needs to be patched to XR first, (e.g. `status.vpcId`) and then patched to the `Subnet` (toFieldPath: `spec.forProvider.vpcId`)

Typically, in compositions we see that the option (a) is used with `matchControllerRef` where the `id` of the `VPC` with the same controller ref (owned by the same XR) is extracted and used in `Subnet`:

```
            vpcIdSelector:
              matchControllerRef: true
```

Then, what is wrong? There are two problems invalidating both of the available options as such:

**option (a):** Not every MR field directly maps to a single type, so existing XRR falls short. There are a lot of issue references in the [original ticket](https://github.com/crossplane/crossplane/issues/1770), but one example is the `destinationArn` on a `SubscriptionFilter` resource where destination could be one of the [multiple types](https://github.com/upbound/upjet/issues/95#issuecomment-1487739037).
**option (b):** Resources may not be part of the same composite.

So, we need a solution here. If someone creates a `SubscriptionFilter` and a `DeliveryStream` (using the [same example](https://github.com/upbound/upjet/issues/95#issuecomment-1487739037)) with Crossplane but not necessarily within the same composition, they need a way to fill `spec.forProvider.destinationArn` on `SubscriptionFilter` with the arn of the `DeliveryStream` they intend to. 

> this proposal seems to make the same mistakes that the composition patch and transformation (P&T) has done, and that eventually led to the invention of functions:

I agree that the problem we're trying to solve here is **similar** to what (P&T) has done, but there is that fundamental difference that makes it impossible to solve with compositions or functions, which is `resources may not be part of the same composite`. I can see two ways to overcome this, both with some implications:

- If you want to patch from a resource that is not part of the same composition, define an `ObserveOnly` resource in the same composition and patch from it.
- Functions could access to k8s API and read anything it wants without being limited to `observed` state.

> desirable transformations are open ended and are never enough

Yes. While we lack very basics today, like _take that ARN and put it here_, it would eventually need more and more, just like we see in composition patches. Leveraging and relying on some [external](https://jmespath.org/),[tool](https://github.com/crossplane/crossplane/issues/4351) might help here.

> references don't solve the order problem: (a) no way to order without the MR author having prepared for a reference field (b) dependencies might be more than just a field in an object (think of probes).

I am not aware of a real (creation) ordering problem today where we are good enough with eventual consistency.
We had the deletion ordering problem and tackling it with the `Usage` type. 
I am not claiming it will not complicate anything but we can still stay on the "declarative" world by saying, _this value needs to be constructed this way and eventually patched there._

> in general, this proposal seems to clone the P&T idea into the object system. Do we really want that?

Not sure if the solution should be the one proposed here, but we need a solution to the problem defined above.

> the core motivation behind this proposal seems to be that compositions are too slow (e.g. not watch-driven, but run in 60s interval). Maybe we should fix the latter.

Not sure how you came up with this assumption, but I am not aware of such a motivation for Generic XRR. IIRC, compositions not being watch-driven came up recently in the composition function (re)design; however, Generic XRR has been an [open issue](https://github.com/crossplane/crossplane/issues/1770) for more than 2 years. One can claim that, if we could live without it for 2 years, then maybe it is not that an important problem to solve :) I would love to hear from the community on that, and I believe it is not a blocker mostly because there are workaround/hacks there which could unblock use cases but ultimately complicate things further.

turkenh (commented):
Thanks @pedjak!

I like the idea of separating the building and the consumption part of referable values, but I am yet to be convinced that it is the right way to go.

In general, I would love to see:

- More details on motivation - why do we need this?
- A more detailed comparison between build/patch with a single CR vs build here / patch there approach.
- How it plays with compositions? Some real-world examples of different use cases where the feature would be helpful.

pedjak:
Thanks all for the very live discussion. I think it would be important to reach consensus on the following:

(1) Do we really need a way to define generic cross-resource references on managed resources without using compositions?

or

(2) we would like to use (and extend if needed) compositions for that purpose?

@haarchri and @ytsarev - your feedback would be really helpful, i.e. what are the usecases currently not implementable using compositions (or requires some not-so-nice workarounds/hacks)?

design/design-generic-references.md
```diff
@@ -0,0 +1,319 @@
...
+  name: referable-vpc-id
+spec:
+  objects:
+    - apiVersion: ec2.aws.crossplane.io/v1alpha1
+      kind: VPC
+      name: main-vpc
+      fieldPath: status.atProvider.id
+```
+
+The syntax of the `fieldPath` field is similar to what we
```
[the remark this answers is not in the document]
pedjak: actually, it is the same, like in compositions... gonna fix that.

design/design-generic-references.md
```diff
@@ -0,0 +1,319 @@
...
+
+The generic cross-reference should be able to:
+
+* be configurable at runtime
+* use the existing [Kubernetes API object reference] convention 
+  as much as possible
+* provide alternative convention in cases where [Kubernetes API object reference]
+  convention is not suitable
+* be implemented without or requiring only light changes in RBAC
+* support referring multiple objects (namespaced and cluster-scoped) and their
```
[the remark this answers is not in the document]
pedjak: My understanding was that it could be useful to set some MR fields from configmaps or secrets. [@muvaf 's](https://github.com/crossplane/crossplane/pull/2385/files) proposal was tackling that as an issue, i.e. good to have. However, if not needed anymore, we can drop that from the proposal.

haarchri:
we discussed the generic reference design poc here -in a offline conversation with @pedjak and @sttts 
As part of that discussion, I have included a few examples to illustrate various aspects of the proposal. 

https://github.com/haarchri/crossplane-generic-references/tree/main

i will add later comments inline

design/design-generic-references.md
bobh66: I'd like to add a slightly different perspective based on my experience using `EnvironmentConfigs`.  I think there is a fair amount of overlap in the usage scenarios, and for me personally the `EnvironemtConfig` model is simpler and easier to understand and use than any of the other proposals for separate resource references that I have seen.

One of the common use cases we have found for `EnvironmentConfig` is making information available from one Composite to another.  If Composite A creates a set of resources and wants to make some information regarding those resources available to any other Composite/Composition that may need them, it can create an `EnvironmentConfig` object with the appropriate labels and the information to be "published".  Other Composites can retrieve these `EnvironmentConfigs` as needed in their Compositions using the appropriate label selectors.

The process of finding the appropriate `EnvironmentConfigs` and loading them into the Composition `environment` creates what is essentially a runtime context for that Composite.  The user (composition author) can leverage this context to pass information between resources and can reference external data using `EnvironmentConfigs`.  The structure of the local context is defined by the author and information is placed in specific locations by name.  This model works extremely well and provides necessary functionality.

So why limit the `environment` external data references to EnvironmentConfigs?  Why not allow the `environment` specification to include references to any existing Kubernetes object, which can be retrieved by the same mechanism that resolves EnvironmentConfigs today?  If an author wants to retrieve data from an existing XR or MR or ConfigMap or Secret they could specify the G/V/K information required to find the object along with a name or label selector,  and a path within the object to the data they want, and that information can be populated in the `environment` in the specified location.

From the author's perspective this would let me put all of my external data references in one place, and they can be retrieved dynamically at runtime as needed if they exist, or they can eventually reconcile as they become available.

I think one of the overarching goals needs to be keeping the user experience as simple as possible, and requiring a specific resource definition for every reference feels like it is adding complexity for the user and making the composition more difficult to develop and maintain.

design/design-generic-references.md
```diff
@@ -0,0 +1,319 @@
...
+```yaml
+apiVersion: apiextensions.crossplane.io/v1alpha1
+kind: Referable
+metadata:
+  name: referable-vpc-id
+spec:
+  objects:
+    - apiVersion: ec2.aws.crossplane.io/v1alpha1
+      kind: VPC
+      matchingLabels:
```
ytsarev, line 148: what happens if multiple objects are selected with labels? Reminds me of https://github.com/crossplane/crossplane/pull/3981 when we thought about similar quite late in the implementation

ytsarev (commented):
Can we craft some Composition examples of using the proposed referencing mechanism? It would closer illustrate the projected DevEx, how it simplifies patterns like bi-directional patching etc

haarchri:
@ytsarev  i added a few examples ideas here: https://github.com/haarchri/crossplane-generic-references/tree/main but with raw managed resources - because i guess in compositions you can use the patches for references...

pedjak:
I have pushed the second draft. It is heavily reworked, including the received feedback.

design/design-generic-references.md
pedjak: done.

design/design-generic-references.md
```diff
@@ -0,0 +1,319 @@
...
+* `Referable` instances can be properly garbage collected if owner
+  or [`Usage`](https://github.com/crossplane/crossplane/blob/master/design/one-pager-generic-usage-type.md)
+  instances are declared
+* Multiple objects/fields can be referred as value sources
+* Value can be transformed using a Go template
+
+### Disadvantages
+
+* Providers might need to get read access to `Referable` type 
+  in order to resolve generic cross-resource references
```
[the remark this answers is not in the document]
pedjak: added to the new version.

pedjak:
> I don't understand the purpose of the Referable resource if the environment can retrieve all of the required references? What scenarios would use Referable instead of environment?

Only if managed resource needs to refer a value in resource owned by another provider and RBAC rules cannot be changed, but users do not want to use compositions. 

I consider now `Referable` type a very optional thing, and perhaps could be dropped fully out of the proposal if we can agree that changing RBAC is ok for users.

design/design-generic-references.md
```diff
@@ -0,0 +1,477 @@
+# Generic Resource References
+
+* Owner: Predrag Knezevic (@pedjak)
+* Reviewers: Crossplane Maintainers
```
negz, line 4: Let's nominate two specific Crossplane maintainers to make it clear who we expect to approve this design.

design/design-generic-references.md
```diff
@@ -0,0 +1,477 @@
...
+
+Currently, the issue can be solved by defining the following composition:
+
+* Declare under managed resources the needed source object (one of `Location*`)
+  types with [observe only management policy]
+* Patch composite object with the value from the source object
+* Patch Task's `spec.forProvider.sourceLocationArn` field with the composite
+  value
+
+The proposed solution has the following drawbacks:
```
negz, line 57: Making sure I understand: by "the proposed solution" do you mean the pattern you describe above that requires using a Composition?

design/design-generic-references.md
```diff
@@ -0,0 +1,319 @@
...
+* The referencer field may not always target a single field in the referred
+  object
+* The value that is about to be assigned to the field require some sort of
+  transformation
+
+Some of the above issues can be solved by defining a composition. However, there
+are cases where one would like to use and refer to a resource not bound by a 
+composite instance.
+
+__**NOTE**__: [a prior design exists on this topic](https://github.com/crossplane/crossplane/pull/2385).
```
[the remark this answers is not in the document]
negz: I'll recap my understanding of why we want generic cross-resource references here, since it's relevant to what @muvaf wrote above. I touched on some of this at https://github.com/crossplane/crossplane/issues/1770#issuecomment-1633270957.

The current `spec.forProvider.fooRef` implementation:

1. Requires all possible references to be manually identified and declared at build time.
2. Is inflexible - users can only reference types the provider developers declared, in the ways they declared.
3. Requires the referring provider to import any types they refer to.

3 is of particular interest for provider families, because it creates a potential version dependency between two providers in a family. For example if provider-aws-eks v0.42.0 is compiled against the types from provider-aws-ec2 v0.42.0, there's a chance that installing provider-aws-ec2 v0.50.0 will break references. It would be ideal if there was no version coupling between providers within a family.

design/design-generic-references.md
```diff
@@ -0,0 +1,477 @@
...
+type [`Reference`](https://github.com/pedjak/crossplane-runtime/blob/master/apis/common/v1/resource.go#L116-L123):
+
+```go
+type Reference struct {
+// Name of the referenced object.
+Name string `json:"name"`
+
+// Policies for referencing.
+// +optional
+Policy *Policy `json:"policy,omitempty"`
```
negz, line 93: Out of scope for this design, but FYI the presence of this field is a bit of a problem - see https://github.com/crossplane/crossplane-runtime/issues/440.

design/design-generic-references.md
```diff
@@ -0,0 +1,477 @@
...
+```yaml
+apiVersion: datasync.aws.upbound.io/v1beta1
+kind: Task
+metadata:
+  name: datasync-task-example
+spec:
+  forProvider:
+    .
+    .
+    sourceLocationArnRef:
```
negz, line 132: Does this mean that provider developers still need to explicitly identify and declare what fields can be set by a reference at build time? For example here the developer would need to identify that `sourceLocationArn` could reference another type, and thus make sure `sourceLocationArnRef` existed.

design/design-generic-references.md
negz: Per https://github.com/crossplane/crossplane/issues/1770#issuecomment-1627868262 I share the concern that generic references grows to be something like P&T Composition.

Also, per https://github.com/crossplane/crossplane/issues/1770#issuecomment-1633270957 I'm pretty confident that today's hardcoded style of references isn't going to scale. It's too manual to support, too inflexible for users, and creates a coupling between providers when one provider references a type defined by another.

I could be convinced that the answer is to just not have MR-level references at all if we can find a clean way to address this at the Composition level. I would expect such an approach to be controversial though. I suspect there's a not-insignificant amount of folks who don't use Composition today who would be forced to if Composition was the only way to derive one

design/design-generic-references.md
pedjak: > Does this mean that provider developers still need to explicitly identify and declare what fields can be set by a reference at build time? For example here the developer would need to identify that sourceLocationArn could reference another type, and thus make sure sourceLocationArnRef existed.

My understanding so far (and chatting with @muvaf) is that we have automated that with Upjet, but I could be ofcourse wrong here.

design/design-generic-references.md
pedjak: > I could be convinced that the answer is to just not have MR-level references at all if we can find a clean way to address this at the Composition level. I would expect such an approach to be controversial though. I suspect there's a not-insignificant amount of folks who don't use Composition today who would be forced to if Composition was the only way to derive one

This proposal talks about that path as well. No matter what strategy/convention we would like to apply to MR-level reference, at the end of the day, it needs to be implemented by providers. On the other hand, solving the problem via compositions requires no changes at provider side.

design/design-generic-references.md
pedjak: @crossplane/crossplane-maintainers please your hand :)

bobh66: Count me in

design/design-generic-references.md
negz: > My understanding so far (and chatting with @muvaf) is that we have automated that with Upjet, but I could be ofcourse wrong here.

I thought we'd automated _generating the code_ for references, but not automated the process of _identifying which fields can be set using a reference_. It's possible I'm wrong too though - CC @ulucinar and @muvaf.

Either way though, I suspect only being able to use references for certain fields will be too limiting. If a field can reference an arbitrary type, how would we know at development which fields should be able to be set by a reference and which can't? Do we think that we'll be able to predict all the fields users might want to derive using a reference at provider development time?

pedjak: > Either way though, I suspect only being able to use references for certain fields will be too limiting. If a field can reference an arbitrary type, how would we know at development which fields should be able to be set by a reference and which can't? Do we think that we'll be able to predict all the fields users might want to derive using a reference at provider development time?

My understanding is that providers generated with the help of Upjet have mapped all fields that can reference external values, correct @ulucinar? These fields are all visible in MR schemas, hence easy discoverable for consumers. Other, non-upjet based providers needs to references in their MR schema in some way as well.

@haarchri, how is your experience? Did you stumble across some fields that should be referable, but they are not according to MR schema?

turkenh: Alper can correct me if I am wrong, but AFAIK, upjet infers some of those references by parsing their documentation and the examples there but not all. For unidentified ones, it is still a manual process to configure them.

ulucinar: Yes, correct:
> My understanding is that providers generated with the help of Upjet have mapped all fields that can reference external values, correct @ulucinar?

We attempt to deduce the reference parameters from the examples provided in the Terraform docs and this does *not* result in the generation of all reference parameters. For instance, none of the provided example Terraform configurations could be utilizing a possible reference. We do not attempt to parse the argument documentation for this purpose, only the provided HCL configurations are used.

Thus, we have some automation for determining the reference parameters themselves on top of the reference implementation generator in angryjet. But, it's not 100% reliable, optional (an upjet-based provider's maintainer can opt out not to generate the reference parameters) and to the best of my knowledge, is only applicable for upjet-based providers.

@negz, are you suggesting a referencing scheme where, at runtime, we can declare a reference whose source & target are arbitrary (no a priori knowledge about who can reference what), so that at build time, we don't need to know anything about the references?

My understanding is that the proposed approach here makes the reference source (where the data will copied from) dynamic (GVK+name+fieldpath) but the reference target (where the data will be copied into) is still static, i.e., you need to know the target at build time.

design/design-generic-references.md
```diff
@@ -0,0 +1,477 @@
...
+      fieldPath: metadata.annotations[crossplane.io/external-name]
+```
+
+Although the proposed solution sounds like a natural evolution of the existing
+reference mechanism, its adoption would require that:
+
+* Providers implement the support for it.
+* RBAC permissions need to be relaxed, if provider `A` needs to refer values from
+  instances owned by provider `B`. In case of provider families, each provider has
+  enough rights to access all resources within the family, requiring no changes
```
ulucinar, line 192: I wonder if enterprise users may ask for fine-granular access policies in the future. Currently, access is mutually granted to the (service accounts of the) providers A & B, if A & B are in the same family, even if their MRs are not supposed to refer to each other. This's been a natural evolution from the monoliths.

design/design-generic-references.md
```diff
@@ -0,0 +1,477 @@
...
+  in RBAC rules. Adding new rules should be done at deployment, by tweaking RBAC
+  rules for the given provider service account.
+
+### Referable Type
+
+If adding a number of RBAC rules is not an option for users, cross-provider
```
ulucinar, line 198: If I understand this correctly:
- There exists a cross-provider resource reference A->B (a resource in provider A references a resource in provider B)
- Because it's cross-provider, the Crossplane RBAC manager cannot install the rules that will allow A's service account to read B's resources
- The user is not willing to add the RBAC rules herself
- Instead provisions a `Referable` object
- The `Referable` controller is running as a core Crossplane controller and has read access to B's resources: Do we assume here source GVK will be a managed resource?
- `Referable` controller exposes the extracted value in `status` or reports an error if the extraction fails for some reason
- All installed provider's service accounts have read access to all the `Referable` objects in the cluster
- Provider A copies the data from the `Referable` object via a `GenericReference`

So this will allow data transfer across all MRs in the cluster. Do I understand this correctly?

design/design-generic-references.md
```diff
@@ -0,0 +1,477 @@
...
+* Avoid system pollution with [observe only management policy] objects
+* Using compositions, reference could be set even if provider does not support
+  the proposed reference mechanism
+
+### Disadvantages
+
+* Support for `Referable` type requires new controller on Crossplane side
+* Provider RBAC rules should be extended to include read permission 
+  for `Referable` type
+* Providers needs to support the new reference mechanism on their side
```
ulucinar, line 348: Maybe I'm missing this: How can custom transformation/extraction logic on the data path from source to target be handled? We are preserving the `reference.ResolutionRequest.Extract`, right? This proposal is mainly for making the reference target dynamic and rest of the data path is (mostly) untouched?

bobh66:
It seems like the two implementations described by this proposal are different enough that they should probably be addressed separately.

pedjak:
> It seems like the two implementations described by this proposal are different enough that they should probably be addressed separately.

They both talk about the same problem, but propose two different paths for solving it: (1) introducing change in MR schema, so that a reference field can point to an arbitrary type and the path within it, (2) adding values from an arbitrary k8s resource to composition environments and using an appropriate p&t strategy to set the right field on a MR.

These two paths are not mutually exclusive, they could coexist if needed, but the question is (as you mentioned earlier) what makes more sense to implement, i.e. what users really want. If (1) is not needed, then we could drop it from the proposal. Or the opposite. If we go only with (2), then we could even think about removing all reference fields from MR schemas, given that we can set references through compositions?

design/design-generic-references.md
bobh66: I would rephrase it as "this solution" so it's clear what it is referring to.

bobh66 (commented):
As I read through this again it strikes me as being three different approaches to several related problems, and I'm not entirely sure that they should all be combined in one design.

The MR-specific generic reference/selector solution is coming at the problem from the perspective of the MR provider, and seems like it is specific to the scenarios faced by MR designers/implementers.

The `Referable` and `environment`solutions approach the same set of problems from the Composition author's perspective, which makes more sense to me intuitively, but I can see both perspectives.  These two implementations seem redundant and I prefer the `environment` change from a user-experience perspective.

Is the goal of this document just to lay out the different design possibilities?  Or is there an implication that some or all of them will be implemented?

It's not clear by the end of the document what the final design plan is, but maybe there doesn't need to be one?

pedjak:
> It's not clear by the end of the document what the final design plan is, but maybe there doesn't need to be one?

@bobh66 after taking to @phisco and @turkenh it looks like that the majority of us are in favor of taking `environment` path. Interestingly enough, it turned out we can refer a value from a resource not being part of the composition even without declaring observe only duplicate, if we follow this pattern:

### Field `foo` of `MRA` needs to be set to the value of field `bar` of `MRB`

* Create CompositionB declaring `MRB` and `EnvironmentConfig` containing `bar: value` entry

```yaml
kind: Composition
metadata:
  name: composition-b
spec:
  resources:
      - name: MRB
        base:
          apiVersion: acme.com/v1beta1
          kind: MRB
          spec:
            forProvider:
              region: "us-east-2"
        patches:
            fromFieldPath: metadata.annotations[crossplane.io/external-name]
            toFieldPath: bar
      - name: Env
        base:
           kind: EnvironmentConfig
           metadata:
             labels:
                env: b
           data:
        patches:
          - type: FromEnvironmentFieldPath
            fromFieldPath: bar
            toFieldPath: data[bar]
```
* Create CompositionA declaring `MRA` and using `EnvironmentConfig` `env-b` to patch `foo` field

```yaml
kind: Composition
metadata:
  name: composition-a
spec:
  environment:
    environmentConfigs:
      - type: Selector
        selector:
          matchLabels:
            env: b
  resources:
    - name: MRA
      base:
        apiVersion: acme.com/v1beta1
        kind: MRA
        spec:
          forProvider:
            region: "us-east-2"
      patches:
        - type: FromEnvironmentFieldPath
          fromFieldPath: bar
          toFieldPath: spec.foo
```

IMHO, this still looks a bit more complex than it should be:
* it is just my feeling, but given that we typically want to use values initialized by providers, env config cannot contain the right key within the first reconcile loop. Hence, consumption of that value in another composition would be postponed and successful reconcile would probably take longer than referring that value directly in `composition-a`, as proposed in this design document.
* `composition-a` can refer the created env config only via labels, because the names are generated by crossplane. Thus, each generated env config should posses a unique label

Further chatting with @haarchri revealed that very often users needs to create singleton resources representing network, firewalls and similar and doing that via composition is a simple overkill, because exposing that through claims (and creating API around it) is not interesting for developers. Developers do want to create database claims, but network is created by infra teams just once, and relevant values are injected into other resources when needed.

Therefore, if nobody objects, I would like to restructure the proposal once again, mark the changes to `environment` as the proposed implementation, and move the MR fields changes to the considered alternatives. Thoughts?

bobh66:
> My only concern is that this would let Compositions read from arbitrary MRs,

Isn't this possible already with Observe Only resources?

turkenh:
> I will restructure the proposal and mark the changes to environment as the proposed implementation, and move the MR fields changes to the considered alternatives. Happy to get feedback on this direction from others.

@pedjak, my biggest concern is we are losing track of the actual problem we’re trying to solve with this proposal.
If we check the [original issue](https://github.com/crossplane/crossplane/issues/1770) that the proposal aims to fix, it is mainly about MR level referencing without composition involved, and we agreed not to invest in that since we didn't see a strong enough signal from the community which could motivate us building a not-so-straightforward solution. 

For more context; in our discussions with @phisco and @pedjak, we have identified the following use cases and came up with the following comparison table:

| Use Case \ Feature                                           | Environment Config | Observe Only Resources | Generic Cross Resource References<br>(This Proposal) |
|--------------------------------------------------------------|:------------------:|:----------------------:|:----------------------------------------------------:|
| Share between Compositions                                   |          ✅         |            ✅           |                           ✅                          |
| Share between composed resources in a Composition            |          ✅         |            ❌           |                           ✅                          |
| Get info from existing external resources into a Composition |          ❌         |            ✅           |                           ❌                          |
| Share between standalone MRs (No Composition)                |          ❌         |            ❌           |                           ✅                          |
| Get info from an arbitrary k8s object into a Composition     |          ❌         |            ✅           |                           ✅                          |

The only use case covered by **only** this proposal is sharing between standalone MRs with no composition involved. As mentioned above, we haven't seen a strong enough signal so far that this is something that we should cover at the cost of additional complexity. There are also some foundational questions with this, like how much functionality we should have at the MR level, which should rather be just building blocks.

So, I suggest holding off iterating on this proposal. We should instead create separate issues/proposals for improving the experience around the existing functionality, e.g. Environment Configs.

pedjak:
> So, I suggest holding off iterating on this proposal. We should instead create separate issues/proposals for improving the experience around the existing functionality, e.g. Environment Configs.

Sure @turkenh - I have created https://github.com/crossplane/crossplane/issues/4583 as the followup, let's continue the discussion overthere.

Rand0mF:
> it is mainly about MR level referencing without composition involved, and we agreed not to invest in that since we didn't see a strong enough signal from the community which could motivate us building a not-so-straightforward solution.

With that, given a simple problem: "Create a MR, patch one of its values from an already existing (non MR) object" 
with current state of crossplane, we need:
- a XRD, take care of its versioning, make sure to include the correct fields to later not need to do breaking changes,
- a composition with a lot of boilerplate. All required fields which we would normally just directly specify in the MR needs to be manually patched and additionally we need observe-only resources with some patches to status fields etc.
- a claim 

All of that introduces way more complexity and unwanted abstractions. With generic patches, not only could this problem be solved easily with just one MR, also maintaining and monitoring is way more easy. Also, some resources already provide such functionality, like the kubernetes provider [Object references](https://marketplace.upbound.io/providers/upbound/provider-kubernetes/v0.16.0/resources/kubernetes.crossplane.io/Object/v1alpha2#doc:spec-references)
Additionally, tools like helm, which are (and should be in my opinion) commonly used for crossplane resources will have it way more easy. 

I would strongly vote for making the MRs more powerful on crossplane level.

PR #4371, opened 2023-07-24
fix: Include top-level description in CRDs generated from XRDs

### LINKED ISSUE #4129: XRDs have empty top-level description field with kubectl explain
### What happened?
When I specify a `description` field for my XRDs at the top level, the way I would normally do for an openapiv3 spec, it shows up as `<empty>` when I call `kubectl explain` on the custom resource type. 

### How can we reproduce it?
When I define a CompositeResourceDefinition using `kubectl apply -f` pointing to a file containing the following yaml, and then run `kubectl explain xmissingdescriptions` (or similar on `xmissingdescriptions.spec`) I see 
```
DESCRIPTION:
     <empty>
```
rather than the descriptions I included. 

```
metadata:
  name: xmissingdescriptions.example.com
spec:
  claimNames:
    kind: MissingDescription
    plural: missingdescriptions
  group: example.com
  names:
    kind: XMissingDescription
    plural: xmissingdescriptions
  versions:
    - name: v1alpha1
      served: true
      referenceable: true
      schema:
        openAPIV3Schema:
          type: object
          description: I would like to have this description show up in kubectl explain, but it does not.
          properties:
            spec:
              type: object
              description: I would also like this description to show up in kubectl explain.
              properties:
                id:
                  description: I can put a description here.
                  type: string
                parameters:
                  type: object
                  description: I can put a description here too.
                  properties:
                    stringParameter:
                      type: string
                      description: This shows up fine in kubectl explain.
                    objectParameter:
                      type: object
                      description: This shows up as well.
              required:
                - id
                - parameters
            status:
              type: object
```

Crossplane version: 1.11
Client Version: v1.26.2
Kustomize Version: v4.5.7
Server Version: v1.24.13-eks-0a21954

mbbush:

Fixes #4129

While looking at this I discovered `status` had the same problem as `spec`, so I fixed that. I slightly improved the deserialization of the openapi spec; we now do it once rather than once for `status` and again for `spec`. I also improved a few of the error paths. 

This is my first time ever coding in go. So please let me know if I've made some rookie mistakes. 

I have:

- [x] Added or updated unit ~and E2E~ tests for my change.

### DIFF
--- a/internal/xcrd/crd.go
+++ b/internal/xcrd/crd.go
@@ -42,7 +42,7 @@ const (
 )
 
 const (
-	errFmtGetProps             = "cannot get %q properties from validation schema"
+	errFmtGenCrd               = "cannot generate CRD for %q %q"
 	errParseValidation         = "cannot parse validation schema"
 	errInvalidClaimNames       = "invalid resource claim names"
 	errMissingClaimNames       = "missing names"
@@ -71,48 +71,15 @@ func ForCompositeResource(xrd *v1.CompositeResourceDefinition) (*extv1.CustomRes
 	crd.Spec.Names.Categories = append(crd.Spec.Names.Categories, CategoryComposite)
 
 	for i, vr := range xrd.Spec.Versions {
-		crd.Spec.Versions[i] = extv1.CustomResourceDefinitionVersion{
-			Name:                     vr.Name,
-			Served:                   vr.Served,
-			Storage:                  vr.Referenceable,
-			Deprecated:               pointer.BoolDeref(vr.Deprecated, false),
-			DeprecationWarning:       vr.DeprecationWarning,
-			AdditionalPrinterColumns: append(vr.AdditionalPrinterColumns, CompositeResourcePrinterColumns()...),
-			Schema: &extv1.CustomResourceValidation{
-				OpenAPIV3Schema: BaseProps(),
-			},
-			Subresources: &extv1.CustomResourceSubresources{
-				Status: &extv1.CustomResourceSubresourceStatus{},
-			},
-		}
-
-		p, required, err := getProps("spec", vr.Schema)
+		crdv, err := genCrdVersion(vr)
 		if err != nil {
-			return nil, errors.Wrapf(err, errFmtGetProps, "spec")
+			return nil, errors.Wrapf(err, errFmtGenCrd, "Composite Resource", xrd.Name)
 		}
-		specProps := crd.Spec.Versions[i].Schema.OpenAPIV3Schema.Properties["spec"]
-		specProps.Required = append(specProps.Required, required...)
-		for k, v := range p {
-			specProps.Properties[k] = v
-		}
+		crdv.AdditionalPrinterColumns = append(crdv.AdditionalPrinterColumns, CompositeResourcePrinterColumns()...)
 		for k, v := range CompositeResourceSpecProps() {
-			specProps.Properties[k] = v
+			crdv.Schema.OpenAPIV3Schema.Properties["spec"].Properties[k] = v
 		}
-		crd.Spec.Versions[i].Schema.OpenAPIV3Schema.Properties["spec"] = specProps
-
-		statusP, statusRequired, err := getProps("status", vr.Schema)
-		if err != nil {
-			return nil, errors.Wrapf(err, errFmtGetProps, "status")
-		}
-		statusProps := crd.Spec.Versions[i].Schema.OpenAPIV3Schema.Properties["status"]
-		statusProps.Required = statusRequired
-		for k, v := range statusP {
-			statusProps.Properties[k] = v
-		}
-		for k, v := range CompositeResourceStatusProps() {
-			statusProps.Properties[k] = v
-		}
-		crd.Spec.Versions[i].Schema.OpenAPIV3Schema.Properties["status"] = statusProps
+		crd.Spec.Versions[i] = *crdv
 	}
 
 	return crd, nil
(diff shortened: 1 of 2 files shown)


internal/xcrd/crd.go
```diff
@@ -85,34 +85,50 @@ func ForCompositeResource(xrd *v1.CompositeResourceDefinition) (*extv1.CustomRes
...
+		if !ok {
+			return nil, errors.Wrapf(nil, errFmtGetProps, "spec")
+		}
+		cSpec := crd.Spec.Versions[i].Schema.OpenAPIV3Schema.Properties["spec"]
+		cSpec.Required = append(cSpec.Required, xSpec.Required...)
+
+		if specDesc := xSpec.Description; specDesc != "" {
+			cSpec.Description = specDesc
+		}
+		// This ordering prevents the user from overwriting crossplane properties.
```
mbbush, line 106: I'm not sure if this is desirable or undesirable behavior.

internal/xcrd/crd.go
```diff
@@ -85,34 +85,50 @@ func ForCompositeResource(xrd *v1.CompositeResourceDefinition) (*extv1.CustomRes
...
-			specProps.Properties[k] = v
+
+		xSpec, ok := s.Properties["spec"]
+		if !ok {
+			return nil, errors.Wrapf(nil, errFmtGetProps, "spec")
+		}
+		cSpec := crd.Spec.Versions[i].Schema.OpenAPIV3Schema.Properties["spec"]
+		cSpec.Required = append(cSpec.Required, xSpec.Required...)
+
+		if specDesc := xSpec.Description; specDesc != "" {
```
mbbush, line 103: Maybe this conditional is unneeded?

internal/xcrd/crd.go
```diff
@@ -144,53 +111,71 @@ func ForCompositeResourceClaim(xrd *v1.CompositeResourceDefinition) (*extv1.Cust
...
-			return nil, errors.Wrapf(err, errFmtGetProps, "spec")
-		}
-		specProps := crd.Spec.Versions[i].Schema.OpenAPIV3Schema.Properties["spec"]
-		specProps.Required = append(specProps.Required, required...)
-		for k, v := range p {
-			specProps.Properties[k] = v
+			return nil, err
 		}
+		crdv.AdditionalPrinterColumns = append(crdv.AdditionalPrinterColumns, CompositeResourceClaimPrinterColumns()...)
 		for k, v := range CompositeResourceClaimSpecProps() {
```
mbbush, line 119: One consequence of setting the default properties after the user's custom ones is that the user cannot overwrite (or extend) anything defined in `CompositeResourceClaimSpecProps()` (or the similar sets of properties for claims and status). 

I'm not sure if this is desirable or undesirable behavior, but I'm leaving it unchanged.

Azanul: I think this is desirable since Crossplane would want its own set of validations on the fields for proper functioning.

internal/xcrd/crd_test.go
```diff
@@ -111,10 +113,12 @@ func TestForCompositeResource(t *testing.T) {
           "type": "string"
         }
       },
-      "type": "object"
+      "type": "object",
+	  "description": "Status of the resource."
```
Azanul, line 117: You might want to format it once

Azanul (commented):
Looks too good for someone writing Go for the first time

ezgidemirel (commented):
Hi @mbbush, thanks for catching and fixing the issue! I suggested a couple of small changes. Could you please have a look?

internal/xcrd/crd.go
```diff
@@ -144,53 +111,71 @@ func ForCompositeResourceClaim(xrd *v1.CompositeResourceDefinition) (*extv1.Cust
...
-		p, required, err := getProps("spec", vr.Schema)
+		crdv, err := genCrdVersion(vr)
 		if err != nil {
-			return nil, errors.Wrapf(err, errFmtGetProps, "spec")
-		}
-		specProps := crd.Spec.Versions[i].Schema.OpenAPIV3Schema.Properties["spec"]
-		specProps.Required = append(specProps.Required, required...)
-		for k, v := range p {
-			specProps.Properties[k] = v
+			return nil, err
```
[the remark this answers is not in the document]
mbbush: I'm still trying to figure out why we wrap errors. If the only reason is to provide useful context about the error, I don't think it's necessary here, since all the errors in `genCrdVersion` are already wrapped with context, either in that function, or in `parseSchema`. Is there a reason to add an additional layer of wrapping? 

And what would I wrap it with? Something like "error generating custom resource definition for xrd spec version index `i`" (defined in the constants like the other messages, of course)

internal/xcrd/crd.go
```diff
@@ -144,53 +111,71 @@ func ForCompositeResourceClaim(xrd *v1.CompositeResourceDefinition) (*extv1.Cust
...
+		Schema: &extv1.CustomResourceValidation{
+			OpenAPIV3Schema: BaseProps(),
+		},
+		Subresources: &extv1.CustomResourceSubresources{
+			Status: &extv1.CustomResourceSubresourceStatus{},
+		},
+	}
+	s, err := parseSchema(vr.Schema)
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
mbbush: The wrapper that seems most appropriate is `errParseValidation`, but `parseSchema` is already wrapping all its errors with that message. Is there value in wrapping it again?

internal/xcrd/crd_test.go
mbbush: Oh, what's going on here is that most of this json object is indented with spaces and I added on tab here, which shows up as a different width in github than in my editor. 

I'll reformat the whole object with tabs instead of spaces, since that seems to be the convention in this repo.

Developers may quarrel about whitespace preferences, but the one thing that everyone should be able to agree on is that a mixture of tabs and spaces on a single line should be forbidden!

internal/xcrd/crd.go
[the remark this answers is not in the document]
jeanduplessis: @mbbush not sure if you're aware but we recently updated the contribution guides around errors: https://github.com/crossplane/crossplane/tree/master/contributing#wrap-errors and https://github.com/crossplane/crossplane/blob/master/contributing/guide-observability.md#error-reporting-as-events

internal/xcrd/crd_test.go
Azanul: Agreed

internal/xcrd/crd.go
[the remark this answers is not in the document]
mbbush: Thanks @jeanduplessis. I'd seen the doc about wrapping errors before, but not the event one. 

Is the idea that we're building a sort of stacktrace by adding an additional layer of wrapping every time there's a function return? 

I'm happy to wrap each error an additional time, I'd just really like to understand why so that I can write a helpful message.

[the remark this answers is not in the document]
ezgidemirel: Hi @mbbush, `genCrdVersion` is called in two different functions and wrapping the error will help us to distinguish. Also, it is a good practice to wrap errors each time we return for future-proofing as well.

internal/xcrd/crd.go
```diff
@@ -144,53 +111,71 @@ func ForCompositeResourceClaim(xrd *v1.CompositeResourceDefinition) (*extv1.Cust
...
+	}
+	crdv.Schema.OpenAPIV3Schema.Description = s.Description
+
+	xSpec, ok := s.Properties["spec"]
+	if !ok {
+		return nil, errors.Errorf(errFmtGetProps, "spec")
+	}
```
ezgidemirel, line 152: We should not return an error if there is no `spec` field in the properties. This is a behavior change. If you check the previous `getProps` function, we're not returning any error if the field doesn't exist.

internal/xcrd/crd.go
```diff
@@ -144,53 +111,71 @@ func ForCompositeResourceClaim(xrd *v1.CompositeResourceDefinition) (*extv1.Cust
...
+	cSpec.Description = xSpec.Description
+	for k, v := range xSpec.Properties {
+		cSpec.Properties[k] = v
+	}
+	crdv.Schema.OpenAPIV3Schema.Properties["spec"] = cSpec
+
+	xStatus, ok := s.Properties["status"]
+	if !ok {
+		return nil, errors.Wrapf(nil, errFmtGetProps, "status")
+	}
```
ezgidemirel, line 165: same here

ezgidemirel (changes requested):

internal/xcrd/crd.go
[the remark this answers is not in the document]
mbbush: Thanks, that's helpful context, and helps me think of a good message to put in the wrapper, too. Will do.

internal/xcrd/crd.go
mbbush: Thanks for noticing that! I'll add a unit test for an empty XRD, ensure the test passes with the previous version of the code, then fix my code so it behaves the same way.

internal/xcrd/crd.go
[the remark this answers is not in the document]
mbbush: I wrapped it in another layer of `errParseValidation`. It seems like that will just add another layer of noise, but I can't think of anything more useful to say instead. 

I'm open to suggestions.

internal/xcrd/crd_test.go
```diff
@@ -402,155 +409,19 @@ func TestForCompositeResource(t *testing.T) {
 	}
 }
 
-func TestValidateClaimNames(t *testing.T) {
```
mbbush, line 405: The way git decided to parse this diff is rather unfortunate. I didn't change anything in `TestValidateClaimNames` except perhaps what the go linting tool did automatically. 

I validated that the only changes in d793208b were the addition of the two `Test...EmptyXrd` functions by deleting them locally and observing an empty git diff with its parent commit.

ezgidemirel (approved):
Thanks a lot @mbbush for the PR, LGTM!

mbbush:
Thanks @ezgidemirel. The style guides, linting and existing tests made this a smooth, pleasant, easy, and safe experience.

I squashed my commits into three that seem like they'd be good to have in the commit history long term. I'm not really sure what to do about the backport tags, and obviously I can't merge it myself. Can you just take it from here, or is there anything else you need from me?

turkenh (approved):

ezgidemirel:
Hi @mbbush, backport actions for `release-1.11` and `release-1.12` have failed and require manual PR openings. Would you like to cherrypick your change to those branches as described [here](https://github.com/crossplane/crossplane/blob/ad0ff7d6d0e4850168883905ed8e1509089cea15/contributing/release-process.md#bug-fixes-in-release-branch) or shall we do that?

PR #4386, opened 2023-07-26
Bugfix/propagate composition revision ref

### LINKED ISSUE #4385: Composition Revision Ref is not propagated from claim to composite resource with compositionUpdatePolicy Manual
### What happened?

When using `compositionUpdatePolicy: Manual` in a claim/composite pair, we expect that if the compositionRevisionRef field is set in the claim, it gets propagated to the Composite Resource. 
However, this does not happen.

### How can we reproduce it?

1. Create a Claim with `compositionUpdatePolicy: Manual`
2. Composition will be created with latest compositionrevision reference. 
3. Create a new Composition Revision
4. Update Claim and add:
```
compositionRevisionRef:
    name: latest-composition-ref
````
5. Composite resource should reflect the change. 
**6. However, compositionRevisionRef does not get overriden in composite.**

### More info
`func (c *APIDryRunCompositeConfigurator) Configure` in `internal/controller/apiextensions/claim/configurator.go` already does the logic of propagating the CompositionRevisionReference from claim to Composite. (line 125 approx)
However, previously to writing the data to the Composite, the object gets filtered in: 
```
	claimSpecFilter := xcrd.GetPropFields(wellKnownClaimFields)
	ucp.Object["spec"] = filter(spec, claimSpecFilter...)
```

This happens because only some properties are propagated to avoid issues: 
```
	wellKnownClaimFields := xcrd.CompositeResourceClaimSpecProps()
	for _, field := range xcrd.PropagateSpecProps {
		delete(wellKnownClaimFields, field)
	}
```

By adding "compositionRevisionRef" in `/internal/xcrd/schemas.go`:

```diff
-      var PropagateSpecProps = []string{"compositionRef", "compositionSelector", "compositionRevisionRef", "compositionUpdatePolicy", "compositionRevisionSelector"}
+      var PropagateSpecProps = []string{"compositionRef", "compositionSelector", "compositionUpdatePolicy", "compositionRevisionSelector"}
```
**This property does not get filtered out and gets correctly propagated to the composite.**

When doing so, a check for avoiding setting nil compositionRevisionRefs should be put in place as well in `internal/controller/apiextensions/claim/configurator.go`

```diff
        if cp.GetCompositionUpdatePolicy() != nil && *cp.GetCompositionUpdatePolicy() == xpv1.UpdateManual {
-               cp.SetCompositionRevisionReference(cm.GetCompositionRevisionReference())
+               claimRevisionRef := cm.GetCompositionRevisionReference()
+               if claimRevisionRef != nil && claimRevisionRef.Name != "" {
+                       cp.SetCompositionRevisionReference(cm.GetCompositionRevisionReference())
+               }
        }
```

**I will be happy to provide a PR with my solution approach if you think that what I'm saying makes sense :)**

Crossplane version: 1.12.2 (it also happens in latest version: 696f398 at the time of writing)

dukanto:
- Added failing tests
- Do not filter out `compositionRevisionRef` field from claim data to be set in composite's spec if `updatePolicy` is set to `Manual`

Fixes #4385

I have:

- [x] Added or updated unit **and** ~E2E tests~ for my change.

### DIFF
--- a/internal/controller/apiextensions/claim/configurator.go
+++ b/internal/controller/apiextensions/claim/configurator.go
@@ -121,9 +121,10 @@ func (c *APIDryRunCompositeConfigurator) Configure(ctx context.Context, cm resou
 
 	// CompositionRevision is a special field which needs to be propagated
 	// based on the Update policy. If the policy is `Manual`, we need to
-	// overwrite the composite's value with the claim's
+	// remove CompositionRevisionRef from wellKnownClaimFields, so it
+	// does not get filtered out and is set correctly in composite
 	if cp.GetCompositionUpdatePolicy() != nil && *cp.GetCompositionUpdatePolicy() == xpv1.UpdateManual {
-		cp.SetCompositionRevisionReference(cm.GetCompositionRevisionReference())
+		delete(wellKnownClaimFields, xcrd.CompositionRevisionRef)
 	}
 
 	claimSpecFilter := xcrd.GetPropFields(wellKnownClaimFields)
--- a/internal/controller/apiextensions/claim/configurator_test.go
+++ b/internal/controller/apiextensions/claim/configurator_test.go
@@ -372,6 +372,222 @@ func TestCompositeConfigure(t *testing.T) {
 				},
 			},
 		},
+		"UpdatePolicyAutomatic": {
+			reason: "CompositionRevision of composite should NOT be set by the claim",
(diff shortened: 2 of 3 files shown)


ezgidemirel (approved):
Thanks a lot @dukanto for fixing the issue!

turkenh (approved):

PR #4389, opened 2023-07-27
fix: avoid panics due to reflect.TypeOf usage

phisco:

`reflect.TypeOf` of an interface being nil, returns nil, so will result in a nil pointer dereference and a panic if then used to call `.String()` or .Elem() on it, see example: https://go.dev/play/p/J_lEJmXrBBH
`reflect.TypeOf(input).Elem()` instead will directly panic if the underlying type is not a pointer or one of the other supported ones: https://cs.opensource.google/go/go/+/refs/tags/go1.20.6:src/reflect/type.go;l=158

Switching to `fmt.Sprintf("%T", input)`, where applicable, or properly checking returned values elsewhere.

I have:

### DIFF
--- a/internal/controller/apiextensions/composite/composition_transforms.go
+++ b/internal/controller/apiextensions/composite/composition_transforms.go
@@ -24,7 +24,6 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
-	"reflect"
 	"regexp"
 	"strconv"
 	"strings"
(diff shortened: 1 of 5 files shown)


internal/controller/apiextensions/composite/composition_transforms.go
```diff
@@ -485,3 +478,12 @@ var conversions = map[conversionPair]func(any) (any, error){
 		return i.(float64) == float64(1), nil
 	},
 }
+
+// typeOf is just a wrapper around reflect.TypeOf, but will return a string
+// representation of nil if the input is nil instead of panicking.
+func typeOf(input any) string {
+	if input == nil {
+		return fmt.Sprintf("%T", input)
```
pedjak, line 486: I played a bit with `fmt.Sprintf("%T, v)` and it seems that it works fine for non-nil inputs as well... could we use it then for all values:

```
func typeOf(input any) string {
  return fmt.Sprintf("%T", input)
}
```

Also, could we place this function in some util package and use it everywhere needed?

phisco: If we just use fmt.Sprintf i wouldn't bother adding that to an utils package. Will do.

phisco: removed and refactored a bit, can you have a look @pedjak ?

pedjak (approved):
lgtm - it seems that PR title and description (and probably commit messages) should be rewritten to reflect the latest changes.

PR #4399, opened 2023-07-27
feat: add MatchTrue and MatchFalse readiness checks

### LINKED ISSUE #4398: Add new readiness check to match boolean fields

I'm using the Kubernetes Provider to compose regular Kubernetes resources as Crossplane Managed Resources.

As soon as the manifest is applied, the provider marks it as ready. This does not reflect at all the actual status of the resource being managed, it only means "we've successfully applied this manifest".

This leads to troubleshooting being kind of a pain, as the resource is marked as healthy, even though it might not be.

The resources I'm discussing contain a boolean `.status.ready`. I tried matching it with the available [readiness check](https://docs.crossplane.io/latest/concepts/compositions/#resource-readiness-checks), but there is no check for bools.

Trying `MatchString` runs into:
```
  Warning  ComposeResources         5s (x3 over 10s)  defined/compositeresourcedefinition.apiextensions.crossplane.io  cannot compose resources: cannot check whether composed resource is ready: cannot run readiness check at index 0: status.atProvider.manifest.status.ready: not a string
```

Adding a new readiness check `MatchBool` or even two new checks `IsTrue` and `IsFalse`.

LCaparelli:

$title, allowing users to measure MR readiness via boolean fields.

Fixes #4398

I have:


LCaparelli:
I did not add e2e tests, only unit tests. I couldn't find e2e tests regarding the other existing checks, so I thought it was not necessary. If I missed them, or if you'd like some included for these changes, please let me know :grimacing:

LCaparelli:
Also working on a companion PR with docs for this on https://github.com/crossplane/docs/tree/master; please let me know if that's not where I should be documenting this

apis/apiextensions/v1beta1/zz_generated.composition_common.go
```diff
@@ -115,14 +115,16 @@ const (
 	ReadinessCheckTypeNonEmpty       ReadinessCheckType = "NonEmpty"
 	ReadinessCheckTypeMatchString    ReadinessCheckType = "MatchString"
 	ReadinessCheckTypeMatchInteger   ReadinessCheckType = "MatchInteger"
+	ReadinessCheckTypeIsTrue         ReadinessCheckType = "IsTrue"
+	ReadinessCheckTypeIsFalse        ReadinessCheckType = "IsFalse"
```
turkenh, line 119: Would `MatchBool` be more consistent than `IsTrue` and `IsFalse` ?

phisco: Yes, as I said I'd prefer MatchBool

phisco:
@LCaparelli yes, a PR to the doc would be good, we switched recently from a branch based doc to a folder based one, so you can just add that at `master`, near here: https://github.com/crossplane/docs/blob/master/content/master/concepts/compositions.md#match-a-string

apis/apiextensions/v1beta1/zz_generated.composition_common.go
LCaparelli: Initially I thought either way was fine. The more I think of it, the more I'm leaning to keeping this as they are right now.

Just want to say that yes, consistency is important, and I agree with your suggestion.

But, changing this to `MatchBool` would introduce a new field `matchBool`, like so:
```yaml
  readinessChecks:
  - type: MatchBool
    fieldPath: status.atProvider.manifest.status.ready
    matchBool: true
```

If left uninformed, what should be the behavior? For starters, in order to detect that, we would need to make `matchBool` a pointer (or we can't tell apart uninformed from `false`). Which is ok, I guess, but inconsistent with the other checks we got so far (which seem to solve this by not allowing 0-value to be a match, and we can't do that for bools -- `false` must be valid).

If we don't make it a pointer, we can't tell apart 0-value from `false`, meaning we would need to handle `false` as an implicit default, which well... Seems easy to get wrong and miss, as a user.

A final remark: bool has only two possible values. So having specific checks for each possible value won't spiral out of control. These are the only two checks which will ever exist regarding bools.

Wdyt @turkenh @phisco? I think the current implementation avoids a lot of complexity for us and for users, at the cost of, IMO, a small inconsistency

phisco: You are right, `MatchBool` is kind of tricky, maybe we could think to have something like this:
```
  - type: MatchBool
    fieldPath: status.atProvider.manifest.status.ready
    matchBool:
        inverted: true    # default: false, means just take  the value as is.
```

WDYT @LCaparelli @turkenh?

LCaparelli: Thanks for clarifying, I'll work on changes as soon as possible. One final nitpick: I find `inverted` a little less clear than `matchFalse`. Wdyt of this?
```yaml
  - type: MatchBool
    fieldPath: status.atProvider.manifest.status.ready
    matchBool:
        matchFalse: true    # default: false, means just match true
```

phisco: Sounds good to me, no strong opinions on that! Thanks @LCaparelli 🙏

LCaparelli:
@turkenh @phisco thanks for the suggestion guys, I pushed a new commit unifying both checks, lmkwyt

apis/apiextensions/v1beta1/zz_generated.composition_common.go
turkenh: Without reading all discussions above, I checked the latest state and was confused with its latest shape.

I was expecting the following, which was already discussed above:

```yaml
  readinessChecks:
  - type: MatchBool
    fieldPath: status.atProvider.manifest.status.ready
    matchBool: true
```

I kind of understand that we wouldn't be able to differentiate `false` with `not specified` with this API. But does this really have a use case? Why would anybody define a readiness check as follows, and what would they expect in that case?

```yaml
  readinessChecks:
  - type: MatchBool
    fieldPath: status.atProvider.manifest.status.ready
```

If I am missing some strong reasons, I wonder if we could define an enum as the type of `matchBool`, e.g. `*string`

```yaml
  readinessChecks:
  - type: MatchBool
    fieldPath: status.atProvider.manifest.status.ready
    matchBool: True # where True as an enum value
```

phisco: @turkenh I'd rather not add our own enum for booleans, they are already confusing enough in YAML.

To answer your question, about the following example: 

```
readinessChecks:
  - type: MatchBool
    fieldPath: status.atProvider.manifest.status.ready
```

To me it's pretty clear that the example above would mean "just match it as is" similar to matchCo dition defaulting to matching the Ready condition as is.

turkenh: So you mean users might expect `status.atProvider.manifest.status.ready` to be `true` with assuming it is the default behaviour, is this what you mean with  "just match it as is" ?

```yaml
readinessChecks:
  - type: MatchBool
    fieldPath: status.atProvider.manifest.status.ready
```

Assuming yes, ok fair. But I would find the following API more confusing than expecting users to always include `matchBool: false` or `matchBool: true`.

```yaml
  - type: MatchBool
    fieldPath: status.atProvider.manifest.status.ready
    matchBool:
        matchFalse: true    # default: false, means just match true
```

Couldn't we make `matchBool` a required parameter, so that users would always provide one of the following:


```yaml
readinessChecks:
  - type: MatchBool
    fieldPath: status.atProvider.manifest.status.ready
    matchBool: false
```

```yaml
readinessChecks:
  - type: MatchBool
    fieldPath: status.atProvider.manifest.status.ready
    matchBool: true
```

turkenh: Or an optional parameter (*bool), defaulting to `true` if not provided?

Sorry for pushing back, but really finding the latest API with `matchBool.matchFalse: true` confusing.

phisco: @turkenh, yes, type MatchBool with nothing else specified, should mean "mark as ready if the value at the `fieldPath` is true", I expect 99% of people to use that and they should not have to think about anything else in my opinion:
```
readinessChecks:
  - type: MatchBool
    fieldPath: status.atProvider.manifest.status.ready
```
If someone wants to negate it though because that's how their api works, we should provide a way to flip the matched boolean, that's why I was suggesting to call it "inverted" or maybe "negated".
With your approach, the options would be:
- either to have `matchBool` be just a boolean, defaulting to false, which means it would show even for other types when reading the created Composition, which would make no sense:
```yaml
readinessChecks:
  - type: MatchInteger
    fieldPath: status.atProvider.manifest.status.instances
    matchInteger: 2
    matchBool: false
```
- Or to have it be a pointer to boolean, which is similar to what we do for all the other types (MatchInteger,MatchString...) which would behave properly. Enforcing to specify it would worsen the experience of 99% of the users, as I said above, so I don't think making it mandatory is really necessary. Either we accept to handle `nil` as false too or we introduce a more complex struct which might allow us in the future to have more elasticity if we need to add other conditions. No strong opinions on that.

So for me the only two viable options are:
- `matchBool` is a `*bool` and we handle nil as false, I'm fine with that as it abide to the standard of all the other fields.
- `matchBool` is a pointer to a struct, containing a bool field to say whether it should "invert/negate/matchFalse", defaulting to `false`. Which I would be fine with too, given that it follows the same pattern as `MatchCondition`.

I would discard the plain bool and making it mandatory for the reasons above.

turkenh: Thanks for the explanation. 

I was thinking may be we can enforce with CEL only when the type is `MatchBool`. 

> matchBool is a *bool and we handle nil as false.

Wouldn't handling nil as `true` be an option?

LCaparelli: It feels like we're trying to generalize something that does not need to be general. It's ok to be specific if something only has, and only ever will have, two possible cases.

phisco: The struct approach didn't look so complex to me and left space for future additions, but if that's just my feeling I'd rather have matchTrue and matchFalse as you are suggesting, than having matchbool be a pointer with a counterintuitive semantic.

bobh66: Another vote for MatchTrue and MatchFalse - I think it's the cleanest user interface

LCaparelli: > The struct approach didn't look so complex to me and left space for future additions,

About future additions, I'm thinking all of this could eventually be deprecated in favor of a `MatchBool` that is capable of more complex logic (eg, `spec.foo.bar && !spec.baz`), but I'm not even sure that's a use case right now. That door is still open though, if we ever need it.

> It would be helpful to add a comment with a link to this thread for posterity so that they don't have the same discussion from scratch 😊

+1, I'll add it, nice!

> Let's go with MatchTrue/MatchFalse then! Sorry @LCaparelli for the back and forth and thanks for your patience :pray: :grimacing: 

No problem at all, it was an interesting discussion. Thanks for the review, and for being open to discuss and change this around. Y'all rock! :metal:

LCaparelli:
> How about following the existing pattern in `matchString` and `matchInteger`?

Yeah, it's one of the alternatives we were discussing. It seems we've got consensus on moving forward with `MatchTrue/MatchFalse` at this point, but feel free to add to the discussion if there are points there you think are still unaddressed :-)

LCaparelli:
Ok, managed to validate everything is working as expected with a local cluster/build/chart based on my fork. Should be ready for review when you got some time @turkenh @phisco @bobh66, thanks!

LCaparelli:
Docs PR: https://github.com/crossplane/docs/pull/513

QQ: should I add anything for backporting? I'm not familiar with the policy, but we wanted to use this feature ASAP. Would it be possible to backport and release it on v0.13 still? Or do we need to wait until v0.14?

phisco:
@LCaparelli unfortunately our policy is only to backport fixes, not new features, so you'll have to wait for 1.14 😅

jeanduplessis:
@LCaparelli once this PR is merged, the CI pipeline will create a pre-release image that you could use to test out the functionality, but as @phisco mentioned, the feature will only formally land in the 1.14 release.

LCaparelli:
> @LCaparelli unfortunately our policy is only to backport fixes, not new features, so you'll have to wait for 1.14 😅

No problem! I thought so hahahah

> @LCaparelli once this PR is merged, the CI pipeline will create a pre-release image that you could use to test out the functionality, but as @phisco mentioned, the feature will only formally land in the 1.14 release.

Ah, great! We'll give that a spin then, thanks!

LCaparelli:
Can someone please re-run the checklist action? :pray:

pkg/validation/apiextensions/v1/composition/readinessChecks.go
```diff
@@ -83,7 +83,7 @@ func getReadinessCheckExpectedType(r v1.ReadinessCheck) xpschema.KnownJSONType {
 		matchType = xpschema.KnownJSONTypeString
 	case v1.ReadinessCheckTypeMatchInteger:
 		matchType = xpschema.KnownJSONTypeInteger
-	case v1.ReadinessCheckTypeNone, v1.ReadinessCheckTypeNonEmpty, v1.ReadinessCheckTypeMatchCondition:
+	case v1.ReadinessCheckTypeNone, v1.ReadinessCheckTypeNonEmpty, v1.ReadinessCheckTypeMatchCondition, v1.ReadinessCheckTypeMatchTrue, v1.ReadinessCheckTypeMatchFalse:
```
phisco, line 86: here it should actually return `xpschema.KnownJSONTypeBoolean` for `MatchTrue` and `MatchFalse`

LCaparelli: Right, thanks! Fixed. Guess it was working before because we only compare types if this func returns a non-empty `xpschema.KnownJSONType`

phisco (approved):

LCaparelli:
Huh, the e2e tests that failed via GH actions are working locally. `make reviewable` runs them, right?

phisco:
No, `make reviewable` doesn't run them as they take time. You'd have to run `make e2e`. But don't worry about the failure in ci, we are just getting started on e2es and there are still a few flaky test cases.

turkenh (approved):
Thanks @LCaparelli 🙌

LCaparelli:
@phisco @turkenh @bobh66 any further input or are we good to move forward?

PR #4402, opened 2023-07-28
fix(intializer): update storage version of CompositionRevision CRs in etcd

### LINKED ISSUE #4400: crossplane core fails to start up : "cannot apply crd"
### What happened?
After upgrading from 1.7.0 to 1.12.2  the crossplane core doesn't startup.   We see error messages like 

```
crossplane: error: cannot initialize core: cannot apply crd: 
cannot patch object: CustomResourceDefinition.apiextensions.k8s.io 
"compositionrevisions.apiextensions.crossplane.io" is invalid: 
status.storedVersions[1]: Invalid value: "v1": must appear in spec.versions

```

### How can we reproduce it?
TBD.   We have at least 1 deploy where the problem didn't happen and 1 where it did.
They both went through the same upgrade of Crossplane.   However the one that is showing the problem is actually at least a couple years old and has gone through upgrades of k8s while the other was just created within the last month.

NOTE:  We saw a similar situation with service-catalog CRD in May of last year.    What I am saying is, while the symptoms were very different back then the determining factor for which deployments did or did not show the symptoms was whether k8s itself was an 'old' deploy that had been upgraded or if it was a 'young' deploy which had not been upgraded.

Crossplane version:  1.12.2
k8s: 1.25.10 <!-- will try to get history of k8s upgrades 
k8s distro :  aws <!-- Teammate is checking how widespread this problem is, could be on others as well
os: Unavailable at this time
kernel: Unavailable at this time

sttts:

Fixes #4400

I have:

### DIFF
--- a/internal/initializer/crds_migrator.go
+++ b/internal/initializer/crds_migrator.go
@@ -1,0 +1,112 @@
+/*
+Copyright 2023 The Crossplane Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package initializer
+
+import (
+	"context"
+
+	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/crossplane/crossplane-runtime/pkg/errors"
+)
+
+// NewCoreCRDsMigrator returns a new *CoreCRDsMigrator.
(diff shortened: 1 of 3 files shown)


haarchri:
related to #4400

phisco (approved):

ezgidemirel (approved):

internal/initializer/crds_migrator.go
```diff
@@ -0,0 +1,112 @@
...
+	crd.Status.StoredVersions = []string{storageVersion}
+	if err := kube.Status().Patch(ctx, &crd, client.MergeFrom(origCrd)); err != nil {
+		return errors.Wrapf(err, "couldn't update %s crd", c.crdName)
+	}
+
+	// One more check just to be sure we actually updated the crd
+	if err := kube.Get(ctx, client.ObjectKey{Name: c.crdName}, &crd); err != nil {
+		return errors.Wrapf(err, "cannot get %s crd to check", c.crdName)
+	}
+	if !sets.NewString(crd.Status.StoredVersions...).Has(storageVersion) {
```
turkenh, line 107: Shouldn't this check `StoredVersions` "only" has `storageVersion` ?

turkenh: Because otherwise, this check would pass without migration here.

turkenh (approved):

internal/initializer/crds_migrator.go
phisco: done and tested again 👍

turkenh:
There was a transient error with e2e testing, which was resolved with a retry.

internal/initializer/crds_migrator.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+		continueToken = resources.GetContinue()
+		if continueToken == "" {
+			break
+		}
+	}
+
+	origCrd := crd.DeepCopy()
+	crd.Status.StoredVersions = []string{storageVersion}
+	if err := kube.Status().Patch(ctx, &crd, client.MergeFrom(origCrd)); err != nil {
```
sttts, line 99: @phisco this must be `MergeFromWithOptions(origCrd, client.MergeFromWithOptimisticLock)`. Otherwise, we risk data loss.

PR #4403, opened 2023-07-28
inject noop logger by default to ctrl-runtime to avoid warnings

lsviben:

controller-runtime v0.15.0 introduced a warning if the logger is not set for it through `crtl.SetLogger`. 

Except in debug mode, we dont set a logger for it as its too verbose. But as now we get a stack trace warning, adding a noop logger explicity to avoid it. 

Fixes # 

I have:

### DIFF
--- a/cmd/crossplane/main.go
+++ b/cmd/crossplane/main.go
@@ -19,6 +19,7 @@ package main
 
 import (
 	"fmt"
+	"io"
 
 	"github.com/alecthomas/kong"
 	admv1 "k8s.io/api/admissionregistration/v1"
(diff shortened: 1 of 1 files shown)


phisco:
Given that we couldn't reproduce it, no need to get it in alongside #4402, still I think it's worth getting it in main and release-1.13 👍

turkenh:
I could reproduce this with v1.13.0 by installing provider-nop, and seeing stack trace both in core and rbac manager.

I think it happens after the first log attempt which is triggered by a package installation.

phisco (approved):

turkenh (approved):

PR #4421, opened 2023-07-31
Update the status of observe-only and ignore-changes design docs

lsviben:

Adds status for design docs:
ObserveOnly -> `Accepted`, but with a note in ManagenentPolicy section that it was changed in IgnoreChanges
IgnoreChanges -> `Accepted`

I have:

- ~[] Added or updated unit **and** E2E tests for my change.~
- ~[] Added `backport release-x.y` labels to auto-backport this PR if necessary.~

### DIFF
--- a/design/design-doc-observe-only-resources.md
+++ b/design/design-doc-observe-only-resources.md
@@ -2,7 +2,7 @@
 
 * Owners: Hasan Turken (@turkenh)
 * Reviewers: Crossplane Maintainers
-* Status: Draft
+* Status: Accepted
 
 ## Background
 
@@ -82,6 +82,10 @@ the desired state and the observed state of the resource and when the
 updated with the latest observation of the resource.
 
 ### Management Policy
+> [!NOTE]
+> The management policy was significantly changed in a
+subsequent design for [ignore changes]. Keeping this section for historical
+purposes.
 
 To support observing resources without taking ownership, we will introduce a new
 spec named `managementPolicy` to the Managed Resources. We will also deprecate
@@ -713,3 +717,4 @@ it as a migration path to Crossplane.
 [desired tags]: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/vpc#tags
 [passing sensitive configuration]: https://github.com/crossplane/crossplane/pull/2886#discussion_r862615416
 [`type: Webhook` composition function]: https://github.com/crossplane/crossplane/blob/master/design/design-doc-composition-functions.md#using-webhooks-to-run-functions
+[ignore changes]: https://github.com/crossplane/crossplane/blob/ad0ff7d6d0e4850168883905ed8e1509089cea15/design/one-pager-ignore-changes.md
--- a/design/one-pager-ignore-changes.md
+++ b/design/one-pager-ignore-changes.md
@@ -2,7 +2,7 @@
 
 * Owners: Lovro Sviben (@lsviben)
 * Reviewers: @turkenh, @negz
-* Status: Draft
+* Status: Accepted
 
 ## Background
 


design/design-doc-observe-only-resources.md
```diff
@@ -2,7 +2,7 @@
 
 * Owners: Hasan Turken (@turkenh)
 * Reviewers: Crossplane Maintainers
-* Status: Draft
+* Status: Defunct
```
jeanduplessis, line 5: Is it worth adding a link to the ignore change one-pager to show that its been superseded by it?

lsviben: Good point

design/design-doc-observe-only-resources.md
```diff
@@ -2,7 +2,7 @@
 
 * Owners: Hasan Turken (@turkenh)
 * Reviewers: Crossplane Maintainers
-* Status: Draft
+* Status: Defunct (superseded by [ignore changes])
```
turkenh, line 5: We move a defunct design under https://github.com/crossplane/crossplane/tree/master/design/defunct

turkenh (commented):
I am a bit hesitant to move the whole design to defunct as some sections are still relevant (e.g. [full state under status](https://github.com/crossplane/crossplane/blob/master/design/design-doc-observe-only-resources.md#api-changes---full-state-under-statusatprovider) and [future work](https://github.com/crossplane/crossplane/blob/master/design/design-doc-observe-only-resources.md#future-work)) and specific to what an Observe Only resource would mean for Crossplane.

What about putting a notice on top of [Management Policy](https://github.com/crossplane/crossplane/blob/master/design/design-doc-observe-only-resources.md#management-policy) section stating that the shape of the API has changed with the other design and leave the design as draft?

lsviben:
> I am a bit hesitant to move the whole design to defunct as some sections are still relevant (e.g. [full state under status](https://github.com/crossplane/crossplane/blob/master/design/design-doc-observe-only-resources.md#api-changes---full-state-under-statusatprovider) and [future work](https://github.com/crossplane/crossplane/blob/master/design/design-doc-observe-only-resources.md#future-work)) and specific to what an Observe Only resource would mean for Crossplane.
> 
> What about putting a notice on top of [Management Policy](https://github.com/crossplane/crossplane/blob/master/design/design-doc-observe-only-resources.md#management-policy) section stating that the shape of the API has changed with the other design and leave the design as draft?

Right, some parts are still valid. I can just put the notice. 

But not sure about leaving it as Draft, I would then set it to Accepted plus a note based on the instructions [here](https://github.com/crossplane/crossplane/blob/a53d5f1ff9e0f50e44b421ce5a811251a314437c/design/README.md#L61-L72):
```
Note that in some cases designs are "semi defunct". These designs are deemed
relevant enough to keep in the "accepted" state rather than marking them
completely defunct, but may:

* Be only partially implemented.
* Reference defunct concepts, like resource classes or Crossplane workloads.
* Have iterated either conceptually, API wise, or code wise since initial
  implementation.

The current convention for such documents is to leave them in the "Accepted"
state, but add a prominent note at the top of the document explaining the way(s)
in which they are no longer accurate.
```

turkenh:
> But not sure about leaving it as Draft, I would then set it to Accepted plus a note based on the instructions here:

Ah yes, good finding!
_Accepted with notice_ sounds like the way to go 👍

turkenh (approved):

design/design-doc-observe-only-resources.md
```diff
@@ -83,6 +83,10 @@ updated with the latest observation of the resource.
 
 ### Management Policy
 
+** Important Note ** The management policy was significantly changed in a
```
turkenh, line 86: Any way to get this rendered more eye-catching?

Screenshot 2023-08-01 at 17 40 21

lsviben: Ill try to find a way

phisco:
The old composition function proposal was moved to the defunct folder, do we want to do the same here for the observe only one that is getting marked as defunct (or undo that move if that was not the right thing to do?)

design/design-doc-observe-only-resources.md
lsviben: Thanks @phisco, was looking for something like that!

lsviben: Screenshot from 2023-08-01 17-00-56

turkenh:
> Maybe the composition function one should undergo the same, and come back to the Accepted designs.

The composition function (re)design was a [complete rewrite](https://github.com/crossplane/crossplane/pull/4306/files#diff-de0680e374a7d46cada7da9cbbd8cd9512b5c530a4554555a80a3b230011e92d) of the same document with the same title.

PR #4424, opened 2023-08-01
fix: copy x-kubernetes-validations from spec to composite/claim CRDs

### LINKED ISSUE #3604: `x-kubernetes-validations` on XRD's spec gets dropped in CRD
### What happened?

I created an XRD which uses the `x-kubernetes-validations` field on the for the top-level `spec` property. I also am using it in deeper fields within `spec`.

When the CRD is generated from the XRD, the spec's `x-kubernetes-validations` is gone, but the nested field has its `x-kubernetes-validations` intact.

I suspect this is a result of 

### How can we reproduce it?

Here is a minimal XRD that demonstrates the issue.

```yaml
---
metadata:
  name: xtests.test.example.com
spec:
  group: test.example.com
  names:
    kind: XTest
    plural: xtests
  claimNames:
    kind: Test
    plural: tests
  versions:
    - name: v1alpha1
      served: true
      referenceable: true
      schema:
        openAPIV3Schema:
          type: object
          properties:
            spec:
              type: object
              properties:
                counter:
                  type: integer
                  x-kubernetes-validations:
                    - rule: self >= oldSelf
                      message: counter can only increase
              x-kubernetes-validations:
                - rule: '!has(oldSelf.counter) || has(self.counter)'
                  message: cannot be removed
```

Check the keys in the resulting CRD's spec definition. The `x-kubernetes-validations` field is missing.
```
→  kubectl get crd tests.test.example.com -o json | jq '.spec.versions[0].schema.openAPIV3Schema.properties.spec | keys'
[
  "properties",
  "type"
]
```

Check the nested field and see that the field _does_ exist.
```
→  kubectl get crd tests.test.example.com -o json | jq '.spec.versions[0].schema.openAPIV3Schema.properties.spec.properties.counter | keys'
[
  "type",
  "x-kubernetes-validations"
]
```

Crossplane version: `1.11.0-rc.0.195.g3ce766b3`
Kubernetes version: `1.25.0`

tenstad:

Fixes #3604

Copies the `x-kubernetes-validations` field from XRD's `spec` to CRD's `spec`s, to enable validation on the `spec` field.

- Should it be copied from/to `status` as well?
- Should e2e tests cover this change (and the `required` field from https://github.com/crossplane/crossplane/pull/2011)?
- Open to suggestions for better example in test, current validation does not really do anything as `engineVersion` is required. 

I have:

- [x] Added or updated unit ~~**and** E2E~~ tests for my change.

### DIFF
--- a/internal/xcrd/crd.go
+++ b/internal/xcrd/crd.go
@@ -149,6 +149,7 @@ func genCrdVersion(vr v1.CompositeResourceDefinitionVersion) (*extv1.CustomResou
 	xSpec := s.Properties["spec"]
 	cSpec := crdv.Schema.OpenAPIV3Schema.Properties["spec"]
 	cSpec.Required = append(cSpec.Required, xSpec.Required...)
+	cSpec.XValidations = append(cSpec.XValidations, xSpec.XValidations...)
 
 	cSpec.Description = xSpec.Description
 	for k, v := range xSpec.Properties {
--- a/internal/xcrd/crd.go
+++ b/internal/xcrd/crd.go
@@ -160,6 +160,7 @@ func genCrdVersion(vr v1.CompositeResourceDefinitionVersion) (*extv1.CustomResou
 	xStatus := s.Properties["status"]
 	cStatus := crdv.Schema.OpenAPIV3Schema.Properties["status"]
 	cStatus.Required = xStatus.Required
+	cStatus.XValidations = xStatus.XValidations
 	cStatus.Description = xStatus.Description
 	for k, v := range xStatus.Properties {
 		cStatus.Properties[k] = v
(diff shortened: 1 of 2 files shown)


turkenh:
> Should it be copied from/to status as well?

I don't see a reason not to do so.

> Should e2e tests cover this change (and the required field from Fix CRD Required Schema #2011)?

More tests are better in general but for this specific case, I believe unit tests are good enough.

> Open to suggestions for better example in test, current validation does not really do anything as engineVersion is required.

What about "cannot change engine version" instead? https://kubernetes.io/blog/2022/09/29/enforce-immutability-using-cel/#immutability-upon-object-creation

tenstad:
Thanks for the feedback! Copied the status validation now as well, and updated the validation case in tests. They could in theory be moved into the fields they validate, rather than being in `spec`, but we need something in `spec` so I guess it's fine :)

phisco (approved):

lsviben (approved):

Issue #4432, opened 2023-08-02
Move `xfn` out of tree - e.g. to `crossplane/function-runtime-oci` - and remove it from the Helm chart.


phisco:

phisco:
Moved to https://github.com/crossplane/function-runtime-oci, I'll create a dedicated ticket to switch it to `v1beta1`

korenyoni:
Is there a Helm Chart for `xfn` now?

As of the time of writing this comment I cannot find it.

PR #4431, opened 2023-08-02
feat: add xfn v1beta1 grpc interface

phisco:
### Description of your changes

Address part of https://github.com/crossplane/crossplane/issues/4419, introducing the new interface defined in #4306.

I have:

- [x] Read and followed Crossplane's [contribution process].
- [ ] ~Added or updated unit **and** E2E tests for my change.~
- [x] Run `make reviewable` to ensure this PR is ready for review.
- [ ] ~Added `backport release-x.y` labels to auto-backport this PR if necessary.~

[contribution process]: https://git.io/fj2m9

PR #4433, opened 2023-08-02
Move xfn out-of-tree

phisco:

Partially address https://github.com/crossplane/crossplane/issues/4432, XFN will be moved to a dedicated repository and updated to implement the new xfn interface defined in https://github.com/crossplane/crossplane/pull/4431.

I have:


pedjak (approved):

cmd/crossplane/core/certs.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package oci
+package core
```
negz, line 17: What still uses the functionality in this file? I couldn't immediately find anything.

go.mod
```diff
@@ -30,15 +26,19 @@ require (
...
 	sigs.k8s.io/yaml v1.3.0
 )
 
-require google.golang.org/genproto/googleapis/rpc v0.0.0-20230525234030-28d5490b6b19 // indirect
+require (
+	github.com/cyphar/filepath-securejoin v0.2.3 // indirect
+	github.com/google/uuid v1.3.0 // indirect
+	golang.org/x/sys v0.10.0 // indirect
+	google.golang.org/genproto/googleapis/rpc v0.0.0-20230525234030-28d5490b6b19 // indirect
+)
```
negz, line 41: Does it make sense for these to all be in their own require block? Should they be merged with the following group?

negz (approved):
This LGTM! I'd prefer to merge sooner rather than later, since I'm about to start working on the PTFComposer. As long as the build passes we can remove anything we missed later.

cmd/crossplane/core/certs.go
phisco: It's used here in the same package: https://github.com/crossplane/crossplane/blob/e312b6e92b85d781718f20a29547f69ad6e7558a/cmd/crossplane/core/core.go#L237

negz:
@phisco Looks like this now needs a rebase.

negz:
Note: Code has moved to https://github.com/crossplane/function-runtime-oci/pull/1

phisco:
Ops, I guess it's for the E2Es, I'm not at my laptop now, if you want feel free to do that, otherwise I'll do it tomorrow morning 🙏

PR #4445, opened 2023-08-03
fix(transforms): properly compute hashes for strings

### LINKED ISSUE #4412: Composition: Calculation of hashes in strings
### What happened?

I'd like to understand what the proper behavior of Crossplane Transform hashing behavior on string inputs. It looks like for basic strings quotes are added and then those quotes are included in the hashing (i.e., if you have an input of `string`, then the hash will be calculated against `"string"` with the additional quotes.

For example, the byte Array `Crossplane` generates a sha256 hash of `19c8a7c24ed0067f606815b59e5b82d92935ff69deed04171457a55018e31224 `
On the other hand the byte Array `"Crossplane"` (note quotes) generates a sha256 hash of `e84ae541a0725d73154ee76b7ac3fec4b007dd01ed701d506cd7e7a45bb48935 `

As can be seen in the test at https://github.com/crossplane/crossplane/blob/f591e4e41ac66d3e5645576c8796fde4f2011670/internal/controller/apiextensions/composite/composition_transforms_test.go#L818, our tests are against strings that have the added quotation marks. 

I believe the quotes are coming from our marshaling into JSON at https://github.com/crossplane/crossplane/blob/f591e4e41ac66d3e5645576c8796fde4f2011670/internal/controller/apiextensions/composite/composition_transforms.go#L347. Running an internal test, for the input string `Crossplane` the generated bytes array `inputJSON` includes the ASCII 34 quote character:

`[34 67 114 111 115 115 112 108 97 110 101 34]`

The Crossplane [docs](https://docs.crossplane.io/latest/concepts/patch-and-transform/#string-transforms) don't seem to indicate that the transforms are performed on the JSON representation of the input data. 

This may cause an issue as upstream systems may calculate a different hash without the quotes. 

### How can we reproduce it?

Use an online calculator or python to calculate the hash of strings. 

Crossplane version: 1.13

stevendborrelli:

Update the way strings are handled during hash calculations.

Fixes #4412 

### Testing process

I created hashes first via <https://emn178.github.io/online-tools/sha512.html>, then to confirm the values generated I created a python script:

```python
if __name__ == '__main__':
  text = b'Crossplane'
  i = str(1234)
  print('sha1:', hashlib.sha1(text).hexdigest())
  print('sha256:', hashlib.sha256(text).hexdigest())
  print('sha512:', hashlib.sha512(text).hexdigest())
  print('sha512int:', hashlib.sha512(bytes(i, 'utf-8')).hexdigest())
```

```console
sha1: 3b683dc8ff44122b331a5e4f253dd69d90726d75
sha256: 19c8a7c24ed0067f606815b59e5b82d92935ff69deed04171457a55018e31224
sha512: 0016037c62c92b5cc4a282fbe30cdd228fa001624b26fd31baa9fcb76a9c60d48e2e7a16cf8729a2d9cba3d23e1d846e7721a5381b9a92dd813178e9a6686205
sha512int: d404559f602eab6fd602ac7680dacbfaadd13630335e951f097af3900e9de176b6db28512f2e000b9d04fba5133e8b1c6e8df59db3a8ab9d60be4b97cc9e81db
```

I have:

- [x] Opened a PR updating the [docs](https://docs.crossplane.io/contribute/contribute/), if necessary.

### DIFF
--- a/internal/controller/apiextensions/composite/composition_transforms.go
+++ b/internal/controller/apiextensions/composite/composition_transforms.go
@@ -344,12 +344,19 @@ func stringConvertTransform(t *v1.StringConversionType, input any) (string, erro
 }
 
 func stringGenerateHash[THash any](input any, hashFunc func([]byte) THash) (THash, error) {
-	inputJSON, err := json.Marshal(input)
-	if err != nil {
-		var ret THash
-		return ret, errors.Wrap(err, errMarshalJSON)
+	var b []byte
+	var err error
+	switch v := input.(type) {
+	case string:
+		b = []byte(v)
+	default:
+		b, err = json.Marshal(input)
+		if err != nil {
+			var ret THash
+			return ret, errors.Wrap(err, errMarshalJSON)
+		}
 	}
-	return hashFunc(inputJSON), nil
+	return hashFunc(b), nil
 }
 
 func stringTrimTransform(input any, t v1.StringTransformType, trim string) string {
--- a/internal/controller/apiextensions/composite/composition_transforms_test.go
+++ b/internal/controller/apiextensions/composite/composition_transforms_test.go
@@ -794,7 +794,7 @@ func TestStringResolve(t *testing.T) {
 				i:       "Crossplane",
 			},
 			want: want{
-				o: "f9fd1da3c0cc298643ff098a0c59febf1d8b7b84",
+				o: "3b683dc8ff44122b331a5e4f253dd69d90726d75",
 			},
 		},
 		"ConvertToSha1Error": {
@@ -815,7 +815,7 @@ func TestStringResolve(t *testing.T) {
 				i:       "Crossplane",
 			},
 			want: want{
-				o: "e84ae541a0725d73154ee76b7ac3fec4b007dd01ed701d506cd7e7a45bb48935",
+				o: "19c8a7c24ed0067f606815b59e5b82d92935ff69deed04171457a55018e31224",
 			},
 		},
 		"ConvertToSha256Error": {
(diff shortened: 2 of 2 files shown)


phisco (approved):

phisco:
I'd say this is actually a fix and should be backported supported versions, wdyt?

stevendborrelli:
I believe it should be fixed in earlier versions.  However any existing compositions that calculate a hash will generate new values, so we probably need to add a note.

stevendborrelli:
Since I am paranoid, here is one more confirmation of the generated values. We should definitely backport the fix.

```console
echo -n 1234 | shasum -a 512
d404559f602eab6fd602ac7680dacbfaadd13630335e951f097af3900e9de176b6db28512f2e000b9d04fba5133e8b1c6e8df59db3a8ab9d60be4b97cc9e81db  -
echo -n Crossplane | shasum -a 512
0016037c62c92b5cc4a282fbe30cdd228fa001624b26fd31baa9fcb76a9c60d48e2e7a16cf8729a2d9cba3d23e1d846e7721a5381b9a92dd813178e9a6686205  -
echo -n Crossplane | shasum -a 256
19c8a7c24ed0067f606815b59e5b82d92935ff69deed04171457a55018e31224  -
echo -n Crossplane | shasum -a 1  
3b683dc8ff44122b331a5e4f253dd69d90726d75  -
```

```

pedjak:
> I believe it should be fixed in earlier versions. However any existing compositions that calculate a hash will generate new values, so we probably need to add a note.

Even if we do not backport it, users will be faced with new hashes after Crossplane upgrade. Thus, perhaps we should skip this change in bugfix versions, to let people plan better their upgrades?

phisco:
I don't think anyone will complain if we go from wrong hashes to correct ones 😅

turkenh (approved):

PR #4465, opened 2023-08-07
Readme Get Started Section

shanecmiller23:

Add section for Get Started to the Readme. Right now there is no direction for brand new users who stumble upon the GitHub Readme.

I have:

- [X] Added or updated unit **and** E2E tests for my change.
- [X] Opened a PR updating the [docs](https://docs.crossplane.io/contribute/contribute/), if necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,9 +11,9 @@ control of the schema of the declarative API it offers.
 
 Crossplane is a [Cloud Native Computing Foundation][cncf] project.
 
-## Getting Started
+## Get Started
 
-Crossplane's [Getting Started Documentation](https://docs.crossplane.io/v1.13/getting-started/) covers Crossplane installation and cloud provider quickstarts.
+Crossplane's [Getting Started Documentation](https://docs.crossplane.io/latest/getting-started/) covers Crossplane installation and cloud provider quickstarts.
 
 ## Releases
 
--- a/README.md
+++ b/README.md
@@ -11,6 +11,10 @@ control of the schema of the declarative API it offers.
 
 Crossplane is a [Cloud Native Computing Foundation][cncf] project.
 
+## Getting Started
+
+Crossplane's [Getting Started Documentation](https://docs.crossplane.io/v1.13/getting-started/) covers Crossplane installation and cloud provider quickstarts.
+
 ## Releases
 
 Currently maintained releases, as well as the next few upcoming releases are
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ Crossplane is a [Cloud Native Computing Foundation][cncf] project.
 
 ## Get Started
 
-Crossplane's [Getting Started Docs](https://docs.crossplane.io/latest/getting-started/) cover install and cloud provider quickstarts.
+Crossplane's [Get Started Docs] cover install and cloud provider quickstarts.
 
 ## Releases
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -11,6 +11,10 @@ control of the schema of the declarative API it offers.
 
 Crossplane is a [Cloud Native Computing Foundation][cncf] project.
 
+## Getting Started
+
+Crossplane's [Getting Started Documentation](https://docs.crossplane.io/v1.13/getting-started/) covers Crossplane installation and cloud provider quickstarts.
```
negz, line 16: Is there some kind of latest link we can use? Just to remove one more place we'll need to keep the URL fresh as new versions are released.

Other than that, LGTM. Could you wrap at 80 chars to match the rest of the doc though?

plumbis: Use `/latest` instead of the version for an evergreen link

shanecmiller23: updated

README.md
```diff
@@ -11,6 +11,10 @@ control of the schema of the declarative API it offers.
 
 Crossplane is a [Cloud Native Computing Foundation][cncf] project.
 
+## Get Started
+
+Crossplane's [Getting Started Documentation](https://docs.crossplane.io/latest/getting-started/) covers Crossplane installation and cloud provider quickstarts.
```
jbw976, line 16: Link looks good to me now @shanecmiller23, but can you still wrap this new content to 80 chars so it's consistent with the rest of the file? 😇 

Feel free to also squash down to a single commit to keep the history clean 🤓

shanecmiller23: @jbw976 updated. I don't know how to squash down to a single commit, would that just entail opening a new PR?

jbw976: no sir, you can squash using a git interactive rebase, but let's not worry about that 🤓 - we can also let github do it for us when we merge the PR.

But I'm still not seeing this line wrapped at 80 chars, did you push a commit for that? See the screenshot below where the new line here goes much wider than the other lines in the doc? do you see differently? 🤔 
Screenshot 2023-08-18 at 9 30 24 AM

shanecmiller23: @jbw976 it does go longer in that screenshot, but that's only because it includes a link which spell out a full URL. In the actual readme that won't show

jbw976: yep agreed that it does get wrapped when it's fully rendered @shanecmiller23! what i'm getting at here though is that we want to keep the raw markdown clean and having a consistent style.  So in the raw markdown we like to do the following:

* wrap all the content at 80 chars (or immediately after a link if the link goes past 80 chars). A tool like [Rewrap](https://marketplace.visualstudio.com/items?itemName=stkb.rewrap) helps a lot with this.
* we prefer "reference style links" over "inline style links", so we may as well update that while we're here 🤪 - relevant doc: https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#links

So putting that together, I'd expect your contribution here to look like this in the raw markdown:

```
## Get Started

Crossplane's [Getting Started Docs] cover install and cloud provider
quickstarts.
```

Then at the bottom of the doc, you'd add this new reference style link value:

```
[Getting Started Docs]: https://docs.crossplane.io/latest/getting-started/
```

Does that make sense? I'm sorry for the back and forth here, happy to DM or meet for higher bandwidth discussion if there's any lingering questions. Thank you @shanecmiller23! 🙇

shanecmiller23: @jbw976 updated per your comments! apologies on my misunderstanding

jbw976 (approved):
Thanks @shanecmiller23! appreciate this contribution and iterating on the style preferences ✅

PR #4485, opened 2023-08-11
Generate TLS certificates for XP and providers 

### LINKED ISSUE #4305: Create one (or two?) TLS certificates per entity?

Currently we use mTLS for two things:

1. https://github.com/crossplane/crossplane/pull/2919
2. https://github.com/crossplane/crossplane/pull/3884

In more detail, I believe:

* Crossplane has a server cert bundle, used to serve webhooks (to the API server).
* Providers have a server cert bundle, used to serve webhooks (to the API server).
* Crossplane has a client cert bundle, used to connect to an ESS plugin.
* Providers have a client cert bundle, used to connect to an ESS plugin.
* ESS plugins have a server bundle, used to serve requests from Crossplane and Providers.

I expect we'll have more needs for mTLS connections in future.

Currently the configuration for all of this requires setting up a webhook cert bundle and an ESS cert bundle. Right now if I wanted to add another cert bundle (e.g. to allow Crossplane to speak mTLS gRPC to a Composition Function runner) I would presumably make it possible to configure a Function cert bundle for Crossplane, even though Crossplane already _has_ a client cert bundle (to be a client of ESS).

Do we want to stick with this pattern? I suppose the benefit is that if (e.g.) ESS communications are compromised we can rotate that certificate without rotating other client certificates. On the other hand, it seems like each entity having one certificate bundle (or perhaps one client bundle and one server bundle) might be simpler and "good enough"?

ezgidemirel:
With this change, we start to generate TLS certificates per entity and store them in  `xxx-tls-client` and `xxx-tls-server` k8s secrets where `xxx` is the entity name. Once this change is merged, we will start generating a root CA certificate to sign the rest of the TLS certificates for XP, providers and functions. These certificates will be mounted to the corresponding pods.

* For XP, we need a server certificate for webhooks and a client certificate for authenticating to ESS plugins and long-running functions.
* For Providers, we need a server certificate for webhooks and client certificate for authenticating to ESS plugins.
* For Functions, we need a server certificate. That certificate/secret will be generated with the following PR. 

This PR does not update existing webhook and ESS certificate usage. We plan to have the integration in the upcoming PRs.

Fixes #4305  

I have:

- [ ] ~Opened a PR updating the [docs](https://docs.crossplane.io/contribute/contribute/), if necessary.~


cluster/crds/pkg.crossplane.io_configurationrevisions.yaml
```diff
@@ -126,6 +126,14 @@ spec:
...
                   value to true may have unintended consequences. Default is false.
                 type: boolean
+              tlsClientSecretName:
+                description: TLSClientSecretName is the name of the TLS Secret that
+                  stores client certificates of the Provider to call Functions.
+                type: string
+              tlsServerSecretName:
+                description: TLSServerSecretName is the name of the TLS Secret that
+                  stores server certificates of the Provider.
+                type: string
```
phisco, line 136: doesn't make much sense for configurations 🤔 would there be a way not to have them added here?

internal/controller/pkg/manager/reconciler.go
```diff
@@ -392,6 +423,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 	pr.SetControllerConfigRef(p.GetControllerConfigRef())
 	pr.SetWebhookTLSSecretName(r.webhookTLSSecretName)
 	pr.SetESSTLSSecretName(r.essTLSSecretName)
+	sn := fmt.Sprintf(fmtTLSServerSecretName, p.GetName())
+	cn := fmt.Sprintf(fmtTLSClientSecretName, p.GetName())
```
phisco, line 427: Nit: can you make sure the resulting string is not longer than 64 chars and cut the provider name in case it is?

phisco (commented):
Just a few minor comments, I still have to test it out, but overall it looks great! thanks @ezgidemirel! 🙏

internal/controller/pkg/manager/reconciler.go
ezgidemirel: Thanks for reminding that @phisco. It's 253 chars for [resource names](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#:~:text=This%20means%20the%20name%20must,alphanumeric%20characters%2C%20'%2D'%20or%20'.). Updated accordingly.

cmd/crossplane/core/init.go
```diff
@@ -90,7 +93,10 @@ func (c *initCommand) Run(s *runtime.Scheme, log logging.Logger) error {
 
 	steps = append(steps, initializer.NewLockObject(),
 		initializer.NewPackageInstaller(c.Providers, c.Configurations),
-		initializer.NewStoreConfigObject(c.Namespace))
+		initializer.NewStoreConfigObject(c.Namespace),
+		initializer.NewTLSCertificateGenerator(c.Namespace, c.TLSCASecretName, c.TLSServerSecretName, c.TLSClientSecretName, "crossplane", nil, initializer.TLSCertificateGeneratorWithLogger(log.WithValues("Step", "TLSCertificateGenerator"))),
```
phisco, line 97: What about moving the `nil` owners list here to a `TLSCertificateGeneratorOption`?

internal/controller/pkg/revision/deployment.go
```diff
@@ -54,10 +54,18 @@ const (
...
+	tlsServerCertsDir        = "/tls/server"
+
+	tlsClientCertDirEnvVar   = "TLS_CLIENT_CERTS_DIR"
+	tlsClientCertsVolumeName = "tls-client-certs"
+	tlsClientCertsDir        = "/tls/client"
 )
 
 //nolint:gocyclo // TODO(negz): Can this be refactored for less complexity (and fewer arguments?)
-func buildProviderDeployment(provider *pkgmetav1.Provider, revision v1.PackageRevision, cc *v1alpha1.ControllerConfig, namespace string, pullSecrets []corev1.LocalObjectReference) (*corev1.ServiceAccount, *appsv1.Deployment, *corev1.Service) {
+func buildProviderDeployment(provider *pkgmetav1.Provider, revision v1.PackageRevision, cc *v1alpha1.ControllerConfig, namespace string, pullSecrets []corev1.LocalObjectReference) (*corev1.ServiceAccount, *appsv1.Deployment, *corev1.Service, *corev1.Secret, *corev1.Secret) {
```
phisco, line 68: can you add a connebt about the returned secrets or at least use named outputs to make it clearer what's returned?

apis/pkg/v1/interfaces.go
```diff
@@ -396,6 +396,12 @@ type PackageRevision interface {
 
 	GetESSTLSSecretName() *string
 	SetESSTLSSecretName(s *string)
```
negz, line 398: Which (if any) of the existing `ESSTLSSecretName` and `WebhookTLSSecretName` methods will be removed in a future PR? I think we plan to remove the webhook ones, but keep the ESS ones, right? If so it would help to add a comment explaining why we have client, server, and ESS specifically.

apis/pkg/v1/revision_types.go
```diff
@@ -101,6 +101,16 @@ type PackageRevisionSpec struct {
...
 	// +optional
 	ESSTLSSecretName *string `json:"essTLSSecretName,omitempty"`
+
+	// TLSServerSecretName is the name of the TLS Secret that stores server
+	// certificates of the Provider.
+	// +optional
+	TLSServerSecretName *string `json:"tlsServerSecretName,omitempty"`
+
+	// TLSClientSecretName is the name of the TLS Secret that stores client
+	// certificates of the Provider to call Functions.
```
negz, line 111: This comment looks misleading - Providers don't call Functions.

cluster/charts/crossplane/templates/deployment.yaml
```diff
@@ -100,6 +100,12 @@ spec:
           - name: "ESS_TLS_SERVER_SECRET_NAME"
             value: ess-server-certs
           {{- end }}
+          - name: "TLS_CA_SECRET_NAME"
+            value: root-ca-certs
```
negz, line 104: Should this secret have "crossplane" in its name?

cluster/charts/crossplane/templates/secret.yaml
```diff
@@ -42,4 +42,31 @@ metadata:
   name: ess-client-certs
   namespace: {{ .Release.Namespace }}
 type: Opaque
-{{- end }}
\ No newline at end of file
+{{- end }}
+---
+# The reason this is created empty and filled by the init container is we want
+# to manage the lifecycle of the secret via Helm.
```
negz, line 48: Nit: I know this is copying an existing comment, but I don't really understand what it's saying. Why do we want to manage the lifecycle via Helm?

internal/controller/pkg/controller/options.go
```diff
@@ -49,6 +49,14 @@ type Options struct {
 	// injected to CRDs so that API server can make calls to the providers.
 	WebhookTLSSecretName string
 
+	// TLSServerSecretName is the Secret that will be mounted to provider Pods.
```
negz, line 52: Is it worth mentioning what this is for? (Webhooks, right?)

internal/controller/pkg/revision/deployment.go
negz: The amount of arguments and return values of this function concerns me. That said, I'm okay with it growing more temporarily since it will be refactored soon to implement https://github.com/crossplane/crossplane/blob/a888ea/design/one-pager-package-runtime-config.md

cmd/crossplane/core/init.go
negz: I agree with @phisco - six arguments plus variadic options does seem like a lot. If owner refs aren't always used it makes sense to me that they should be an option rather than a fixed argument.

internal/initializer/tls.go
```diff
@@ -0,0 +1,332 @@
...
+		log:                 logging.NewNopLogger(),
+	}
+
+	for _, f := range opts {
+		f(e)
+	}
+	return e
+}
+
+func (e *TLSCertificateGenerator) loadOrGenerateCA(ctx context.Context, kube client.Client, nn types.NamespacedName) (*CertificateSigner, error) {
```
negz, line 98: I see that several methods of `TLSCertificateGenerator` take a `client.Client` as an argument. Would it make sense for `TLSCertificateGenerator` to have a `client.Client` struct field instead?

internal/initializer/tls.go
```diff
@@ -0,0 +1,332 @@
...
+	caSecret := &corev1.Secret{}
+
+	err := kube.Get(ctx, nn, caSecret)
+	if resource.IgnoreNotFound(err) != nil {
+		return nil, errors.Wrapf(err, errFmtGetTLSSecret, nn.Name)
+	}
+
+	if err == nil {
+		kd := caSecret.Data[SecretKeyTLSKey]
+		cd := caSecret.Data[SecretKeyTLSCert]
```
negz, line 108: It seems these constants are currently defined in `ess_tls.go`. FWIW I believe the constants are also defined in the upstream Kubernetes corev1 package. I suggest using using those rather than redefining our own, since it makes it clearer that the key names are an upstream standard.

internal/initializer/tls.go
```diff
@@ -0,0 +1,332 @@
...
+		if len(kd) != 0 && len(cd) != 0 {
+			e.log.Info("TLS CA secret is complete.")
+			return parseCertificateSigner(kd, cd)
+		}
+	}
+	e.log.Info("TLS CA secret is empty or not complete, generating a new CA...")
```
negz, line 114: Does "empty" here mean "not found"? Or are they two different cases?

internal/initializer/tls.go
```diff
@@ -0,0 +1,332 @@
...
+
+	a := &x509.Certificate{
+		SerialNumber:          big.NewInt(2022),
+		Subject:               pkixName,
+		Issuer:                pkixName,
+		DNSNames:              []string{"crossplane-root-ca"},
+		NotBefore:             time.Now(),
+		NotAfter:              time.Now().AddDate(10, 0, 0),
```
negz, line 122: This means our certificates will expire in 10 years, right?

Out of scope for this PR, but have we given any thought to using shorter lived certificates?

negz:
Just to be explicit: I've done a review pass but I trust @phisco to review and merge this PR. I think that will be faster than waiting on me since you're both in the same timezone. 🙂

internal/initializer/tls.go
phisco: Agree with @negz. We should probably also take into consideration implementing certificate rotation in the same PR.

cluster/crds/pkg.crossplane.io_configurationrevisions.yaml
ezgidemirel: I agree, but I couldn't find a way to do that.

cmd/crossplane/core/init.go
ezgidemirel: made it an option

apis/pkg/v1/interfaces.go
ezgidemirel: I was planning to remove ESS as well. We can start using client certificates per entity and have a single server cert for ESS plugins. Our issue is that we're not deploying ESS plugins via package manager. Therefore, cannot generate different server certificates for them.

apis/pkg/v1/revision_types.go
ezgidemirel: fixed

cluster/charts/crossplane/templates/deployment.yaml
ezgidemirel: Updated to `xp-root-ca`. I can also make it `crossplane-root-ca` if you prefer.

cluster/charts/crossplane/templates/secret.yaml
ezgidemirel: We want to delete these certificates when we delete Crossplane. Therefore, they're created via helm.

internal/controller/pkg/controller/options.go
ezgidemirel: correct, updated the comment

internal/initializer/tls.go
ezgidemirel: The reason I kept it as an argument is, the Run function is implementing the Step [interface](https://github.com/crossplane/crossplane/blob/a888ea7657f12d2dd9effa5ef8da8b17b0bea404/internal/initializer/initializer.go#L36).

internal/initializer/tls.go
ezgidemirel: I updated `tls.crt` and `tls.key` keys. But kept the `ca.crt` because the variable name defined in corev1 is `ServiceAccountRootCAKey` which may cause confusions.

internal/initializer/tls.go
ezgidemirel: Since the secret is created without any data via helm, it means "empty" here. We're checking the secret content's completeness to not overwrite the certificates each time the crossplane pod restarts.

cluster/charts/crossplane/templates/deployment.yaml
phisco: +1 for crossplane-root-ca

internal/initializer/tls.go
ezgidemirel: > This means our certificates will expire in 10 years, right?

Yes and with the current implementation, certificates can easily be rotated in case of leakage or expiration by deleting their content and restarting Crossplane pod (if the certificate is created by helm). I plan to document it once this PR and following PRs are merged.

I prefer having an automated rotation in place before creating shorter lived certificates.

apis/pkg/v1/interfaces.go
negz: Got it, thanks. I'd suggest adding comments noting they were deprecated but I think you're planning to fast follow to remove them right?

cluster/charts/crossplane/templates/secret.yaml
negz: It might be worth updating the comment(s) to be more explicit around that.

internal/initializer/tls.go
negz: Got it. I noticed we don't seem to be handling the case where the whole secret is actually not found above. I assume that's intentional though, since we'll just create one if it's not there.

internal/initializer/tls.go
```diff
@@ -0,0 +1,341 @@
...
+	errDecodeKey               = "cannot decode key"
+	errDecodeCert              = "cannot decode cert"
+	errFmtGetTLSSecret         = "cannot get TLS secret: %s"
+	errFmtCannotCreateOrUpdate = "cannot create or update secret: %s"
+)
+
+const (
+	// RootCACertSecretName is the name of the secret that will store CA certificates and rest of the
+	// certificates created per entities will be signed by this CA
+	RootCACertSecretName = "xp-root-ca"
```
phisco, line 52: @ezgidemirel can you switch it to `crossplane-root-ca` as agreed?

apis/pkg/v1/revision_types.go
```diff
@@ -101,6 +101,16 @@ type PackageRevisionSpec struct {
 	// by the provider for External Secret Stores.
 	// +optional
 	ESSTLSSecretName *string `json:"essTLSSecretName,omitempty"`
+
+	// TLSServerSecretName is the name of the TLS Secret that stores server
+	// certificates of the Provider.
+	// +optional
+	TLSServerSecretName *string `json:"tlsServerSecretName,omitempty"`
```
phisco, line 108: Are you planning to use the same field for Functions too in the future?

apis/pkg/v1/interfaces.go
ezgidemirel: Yes, that's what I was thinking. I'm also okay with labeling them as deprecated. But I don't think it will be noticed. 

However, we need to communicate that the providers' webhook and ESS certificate paths and their content will change after we start using the new ones.

internal/initializer/tls.go
ezgidemirel: If the secret belongs to a provider or a function, we will create it. If it belongs to Crossplane, it should be created by Helm and if it's deleted afterward, the Crossplane pod will complain about not being able to mount it. In that case, it should be created again manually.

internal/initializer/tls.go
ezgidemirel: done

phisco (approved):
LGTM, thanks @ezgidemirel! I think we can merge this so that we can unlock the Function's package manager work and ptf composer refactoring @negz is working on.

internal/controller/pkg/revision/deployment.go
```diff
@@ -66,6 +76,21 @@ func buildProviderDeployment(provider *pkgmetav1.Provider, revision v1.PackageRe
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:            *revision.GetTLSServerSecretName(),
+			Namespace:       namespace,
+			OwnerReferences: []metav1.OwnerReference{meta.AsController(meta.TypedReferenceTo(revision, v1.ProviderRevisionGroupVersionKind))},
+		},
+	}
+
+	secCli := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:            *revision.GetTLSClientSecretName(),
```
sttts, line 89: this crashes if the secret name is not set.

sttts:
```
E0906 15:18:08.574869       1 runtime.go:79] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
goroutine 406 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic({0x1883e60?, 0x2e4e750})
	k8s.io/apimachinery@v0.28.1/pkg/util/runtime/runtime.go:75 +0x7c
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile.func1()
	sigs.k8s.io/controller-runtime@v0.16.1/pkg/internal/controller/controller.go:108 +0xa4
panic({0x1883e60, 0x2e4e750})
	runtime/panic.go:884 +0x1f4
github.com/crossplane/crossplane/internal/controller/pkg/revision.buildProviderDeployment(0x40005c5600, {0x1ec2d10, 0x4000a42200}, 0x4000363680, {0x400015dc70, 0xe}, {0x0, 0x0, 0x0})
	github.com/crossplane/crossplane/internal/controller/pkg/revision/deployment.go:83 +0x394
github.com/crossplane/crossplane/internal/controller/pkg/revision.(*ProviderHooks).Post(0x40007613c0, {0x1ea4688, 0x400122e240}, {0x1e8f720, 0x40005c5600}, {0x1ec2d10?, 0x4000a42200})
	github.com/crossplane/crossplane/internal/controller/pkg/revision/hook.go:155 +0x1c8
github.com/crossplane/crossplane/internal/controller/pkg/revision.(*Reconciler).Reconcile(0x40000a9e10, {0x1ea46c0, 0x400122e1b0}, {{{0x0, 0x0}, {0x40006b3bc0, 0x33}}})
	github.com/crossplane/crossplane/internal/controller/pkg/revision/reconciler.go:675 +0x302c
github.com/crossplane/crossplane-runtime/pkg/ratelimiter.(*Reconciler).Reconcile(0x400076f630, {0x1ea46c0, 0x400122e1b0}, {{{0x0?, 0x5?}, {0x40006b3bc0?, 0x4000db5d38?}}})
	github.com/crossplane/crossplane-runtime@v1.14.0-rc.0.0.20230906075713-a2674ee167ac/pkg/ratelimiter/reconciler.go:54 +0x124
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile(0x1ea7a68?, {0x1ea46c0?, 0x400122e1b0?}, {{{0x0?, 0xb?}, {0x40006b3bc0?, 0x0?}}})
	sigs.k8s.io/controller-runtime@v0.16.1/pkg/internal/controller/controller.go:119 +0x8c
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler(0x4000765ea0, {0x1ea4618, 0x4000227400}, {0x196c540?, 0x400075f440?})
	sigs.k8s.io/controller-runtime@v0.16.1/pkg/internal/controller/controller.go:316 +0x30c
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem(0x4000765ea0, {0x1ea4618, 0x4000227400})
	sigs.k8s.io/controller-runtime@v0.16.1/pkg/internal/controller/controller.go:266 +0x1a0
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2()
	sigs.k8s.io/controller-runtime@v0.16.1/pkg/internal/controller/controller.go:227 +0x74
created by sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2
	sigs.k8s.io/controller-runtime@v0.16.1/pkg/internal/controller/controller.go:223 +0x43c
```

ezgidemirel: After discussing with @sttts, we noticed that this happens when we only change the Crossplane image without a helm upgrade. We shouldn't see this error with proper installation/upgrades. But I opened a [PR](https://github.com/crossplane/crossplane/pull/4590) to add nil checks to be on the safe side.

Issue #4419, opened 2023-07-31
Remove the `FunctionIO` type and update the `RunFunctionRequest` and `RunFunctionResponse` RPC types


phisco:
Implement changes to the RPC types as defined by #4306 [here](https://github.com/crossplane/crossplane/pull/4306/files#diff-de0680e374a7d46cada7da9cbbd8cd9512b5c530a4554555a80a3b230011e92dR268).

Issue #4339, opened 2023-07-17
Refactor the `PTFComposer` to accommodate the design (which I think would be a simplification)


negz:

jeanduplessis:
@negz we recently reviewed an issue reported by @duizabojul which ended up relating to composition functions. This re-write might be an opportunity to consider how to address this: https://github.com/crossplane/crossplane/issues/4409

Issue #4288, opened 2023-07-01
`FunctionIO` desired resources array is not bootstrapped with all observed resources


negz:
### What happened?

https://docs.crossplane.io/knowledge-base/guides/composition-functions/ says:

> The desired state of the XR and composed resources is how your Function tells Crossplane what it should do. Crossplane ‘bootstraps’ the initial desired state passed to a Function pipeline with:
>
> * A copy of the observed state of the XR.
> * __A copy of the observed state of any existing composed resources.__
> * Any new composed resources or modifications to observed resources produced from the resources array.

This isn't actually true today - it only includes observed resources that have a corresponding entry in the `resources` array.

https://github.com/crossplane/crossplane/blob/444267e84783136daa93568b364a5f01228cacbe/internal/controller/apiextensions/composite/composition_ptf.go#L613

### What environment did it happen in?
Crossplane version: 444267e84783136daa93568b364a5f01228cacbe

negz:
I've added the `docs` label but raised this here rather than c/docs, because ideally we'd update the implementation to match the documentation rather than the other way around. Sadly I'm not sure we _can_ do that, though.

In the current `PTFComposer` implementation Crossplane knows that an observed (i.e. existing) resource is no longer desired (i.e. should be deleted) if that resource does not appear in the desired resources array produced by the Function pipeline.

The problem then is P&T resources. It's possible to use P&T together with Functions, but if we just bootstrap the desired resources array with all observed resources it will become impossible to delete a resource just by removing it from the P&T resources array. Unless a Function chooses to remove the observed resource from the array we won't notice that nothing actually desires it anymore and it will stick around forever.

FWIW I'm working on this as part of https://github.com/crossplane/crossplane/issues/3896, which has turned into a larger refactor/rethink of the `PTFComposer` implementation.

erikgb:
Relates to https://github.com/crossplane/docs/pull/506

negz:
Just explicitly capturing how I think this should work (and will with https://github.com/crossplane/crossplane/pull/4500):

* The `observed` state passed to the first Function should include all observed resources.
* The `desired` state passed to the first Function should include only rendered P&T templates.

The difference from the current implementation is that we'll render the templates onto an empty composed resource, not the existing observed resource. So unlike the current implementation, `desired` (when passed to the first Function) will never include any observed state - only accumulated desired state.

Issue #3896, opened 2023-03-22
Dry-run apply should happen after xfn pipeline run


duizabojul:
### What happened?

When using xfns, Crossplane is trying to apply composed resources with dry-run option after patches apply but before running the functions pipeline. 

I have custom logic in my xfns where i template some fields of my composed resources, so before running xfn these fields could have a bad type according to xrd/crd so dry-run apply will fail. My xfn pipeline would replace these fields with correctly typed values so running the dry-run apply after xfn pipeline would succeed. 

I suspect if some fields are required in crd, not defined after patches and set in xfn pipeline, the same thing will happen.

I think xfns should allow complex transformations of composed resources and crossplane should only validate custom resources against k8s after xfn pipeline.

negz:
I feel like there might have been a reason I kept the dry-run apply _before_ xfn runs, but I can't remember what it was. What you propose makes sense to me.

djeremiah:
Another use case that needs dry-run after the xfn pipeline is the use of `generatename` in resources that are added during the function run. These don't get names properly set in the current implementation.

negz:
I'm taking a look at this one. It looks like it requires a lot more surgery to the `PTFComposer` (i.e. the Functions-aware composition engine) than I expected, but I think it's worthwhile.

duizabojul:
Thx a lot nic, I can't wait for this one 🙏

negz:
I have a branch where I've done a bunch of refactoring toward this, but given we're close to v1.13 feature freeze and given I'm OOO from July 18 -> Aug 5th I don't think this is going to make v1.13, sorry.

I'm hoping this won't be too overly disruptive in the broader context, which is that we plan to make some big changes to how Functions work for 1.14 per https://github.com/crossplane/crossplane/pull/4306.

djeremiah:
During a conversation at the Crossplane meetup in Munich, we were discussing another use case for XFNs that would require the dry-run apply to be before the xfn pipeline. @mleahu Is interested in injecting an xfn to validate the composed resources during a ci/cd process. But it would need to happen at the very end of the pipeline, after all resources had been initialized. Does this mean we need to dry-run multiple times? Or define different points in the pipeline where functions can run?

negz:
> During a conversation at the Crossplane meetup in Munich, we were discussing another use case for XFNs that would require the dry-run apply to be before the xfn pipeline. @mleahu Is interested in injecting an xfn to validate the composed resources during a ci/cd process.

[...]

PR #4500, opened 2023-08-14
Support v1beta1 style Functions

negz:

Fixes https://github.com/crossplane/crossplane/issues/4339
Fixes https://github.com/crossplane/crossplane/issues/4288
Fixes https://github.com/crossplane/crossplane/issues/3896
Fixes https://github.com/crossplane/crossplane/issues/4558
Fixes https://github.com/crossplane/crossplane/issues/4419
Fixes https://github.com/crossplane/crossplane/issues/3812

This PR:

* Promotes the new `Function` package types to v1beta1
* Enables the `--enable-composition-functions` feature flag by default
* Replaces the `PTFComposer` with a `FunctionComposer` that only calls long-running Function packages
* Updates the `PTComposer` (that is used when the flag is disabled) to accommodate the the changes
 
Note that I've crossed out the docs PR checklist option. This definitely needs a docs update, but I don't think we'll block this mega-PR on getting a docs PR open first.

I have:

- [ ] ~Opened a PR updating the [docs](https://docs.crossplane.io/contribute/contribute/), if necessary.~


apis/apiextensions/v1/composition_revision_types.go
```diff
@@ -56,13 +56,16 @@ type CompositionRevisionSpec struct {
 	// +optional
 	Resources []ComposedTemplate `json:"resources"`
 
-	// Functions is list of Composition Functions that will be used when a
+	// Pipeline is list of Composition Pipeline that will be used when a
```
phisco, line 59:
```suggestion
	// Pipeline is list of Composition Functions that will be used when a
```

negz:
@phisco @ezgidemirel This is 95% ready. I think all that's missing implementation wise is the code that reads Crossplane's TLS secret in order to create gRPC transport credentials.

With something this complex I generally like to see something work end-to-end before I move a PR out of draft, so I'm going to leave it at draft for now. That said, I think it would be pretty safe to start reviewing if you want to take an early pass.

internal/controller/apiextensions/composite/composition_pt.go
```diff
@@ -240,37 +250,45 @@ func (c *PTComposer) Compose(ctx context.Context, xr resource.Composite, req Com
...
 			continue
 		}
 		o := []resource.ApplyOption{resource.MustBeControllableBy(xr.GetUID())}
 		o = append(o, mergeOptions(filterPatches(cd.Template.Patches, patchTypesFromXR()...))...)
 		if err := c.client.Apply(ctx, cd.Resource, o...); err != nil {
-			return CompositionResult{}, errors.Wrap(err, errApply)
+			// TODO(negz): Include the template name (if any) in this error.
+			// Including the rendered resource's kind may help too (e.g. if the
+			// template is anonymous).
+			return CompositionResult{}, errors.Wrap(err, errApplyComposed)
```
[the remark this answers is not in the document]
negz: That seems like a fairly significant behaviour change that's not strictly related to Functions, so I'd prefer to discuss/address it separately from this PR. (Despite the green diff, this is in practice the same logic we had here already.)

apis/apiextensions/v1/composition_common.go
```diff
@@ -265,154 +263,29 @@ type ConnectionDetail struct {
...
-	// Endpoint specifies how and where Crossplane should reach the runner it
-	// uses to invoke containerized Composition Functions.
-	// +optional
-	// +kubebuilder:default="unix-abstract:crossplane/fn/default.sock"
-	Endpoint *string `json:"endpoint,omitempty"`
+// A FunctionReference references a Composition Function that may be used in a
+// Composition pipeline.
+type FunctionReference struct {
+	// Name of the referenced Function.
+	Name string `json:"name"`
```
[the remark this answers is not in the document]
phisco: yes, you are right, my bad. see https://github.com/crossplane/crossplane/pull/4500#discussion_r1309849512

internal/controller/apiextensions/composite/fuzz_test.go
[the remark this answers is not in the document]
phisco: I was looking at it via GH UI and only saw a bunch of helper functions, missed the additional `Fuzz*` tests. My bad, sorry.

internal/controller/apiextensions/composite/composition_ptf.go
```diff
@@ -279,71 +251,201 @@ func NewPTFComposer(kube client.Client, o ...PTFComposerOption) *PTFComposer {
...
+	r := make([]*fnv1beta1.Result, 0)
+	for _, fn := range req.Revision.Spec.Pipeline {
+		req := &fnv1beta1.RunFunctionRequest{Observed: o, Desired: d}
+
+		if fn.Input != nil {
+			in := &structpb.Struct{}
+			if err := in.UnmarshalJSON(fn.Input.Raw); err != nil {
+				return CompositionResult{}, errors.Wrapf(err, errFmtUnmarshalPipelineStepInput, fn.Step)
+			}
+			req.Input = in
```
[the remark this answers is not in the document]
phisco: yeah, I couldn't find a better solution either, that's just unfortunate, but I guess we'll have to live with it.

phisco:
just merged https://github.com/crossplane/crossplane/pull/4530, you can rebase and start e2es @negz 🙏

internal/controller/apiextensions/composite/composition_ptf.go
```diff
@@ -271,71 +236,263 @@ func NewPTFComposer(kube client.Client, o ...PTFComposerOption) *PTFComposer {
...
+		// apply a composed resource if a Required patch failed.
+		if err := RenderFromEnvironmentPatches(cd, req.Environment, ct.Patches); err != nil {
+			// Most likely a required FromEnvironment patch failed.
+			events = append(events, event.Warning(reasonCompose, errors.Wrapf(err, errFmtRenderFromCompositePatches, name)))
+			failed[name] = ComposedResourceState{Resource: cd}
+			continue
+		}
+		if err := RenderFromCompositePatches(cd, xr, ct.Patches); err != nil {
+			// Most likely a required FromComposite patch failed.
+			events = append(events, event.Warning(reasonCompose, errors.Wrapf(err, errFmtRenderFromEnvironmentPatches, name)))
```
negz, line 342: This is the wrong error constant. 😄

negz: I'm about to add a commit that addresses https://github.com/crossplane/crossplane/issues/4558 which will make this irrelevant.

design/design-doc-composition-functions.md
```diff
@@ -347,7 +354,39 @@ message Resource {
...
+  // Ready indicates whether the resource should be considered ready.
+  // 
+  // * Crossplane will never set this field in a RunFunctionRequest.
+  //
+  // * A Function should set this field to READY_TRUE in a RunFunctionResponse
+  //   to indicate that a desired composed resource is ready.
+  //
+  // * A Function should not set this field in a RunFunctionResponse to indicate
+  //   that the desired composite resource is ready. This will be ignored.
+  Ready ready = 3;
```
negz, line 378: @phisco @ezgidemirel @turkenh I'm curious how you feel about this.

Specifically, do you think it's okay to reuse this one `Resource` type in a bunch of different contexts? We currently use the one type to represent:

* The observed XR
* The desired XR
* Observed composed resources
* Desired composed resources

This keeps the protobuf schema simple, but calls for large comments like this explaining when each field will be honored and when each field will be ignored.

The alternative would be to have explicit `ObservedCompositeResource`, `DesiredCompositeResource`, ObservedComposedResource`, etc types. This would make it really obvious in the schema which fields are allowed where, at the cost of a more complex schema (and more complex corresponding code).

turkenh: Would it be sufficient to differentiate `Composed` from `Composite`? I believe what is set when observed vs. desired is more obvious.

negz: @turkenh Yeah, possibly.

At this point though my bias is to merge this as-is. We can make the change before beta in a follow-up PR if we decide one is needed.

apis/apiextensions/v1/composition_revision_types.go
```diff
@@ -41,33 +41,64 @@ type CompositionRevisionSpec struct {
...
+	// +kubebuilder:validation:Enum=Resources;Pipeline
+	// +kubebuilder:default=Resources
+	Mode *CompositionMode `json:"mode,omitempty"`
+
+	// PatchSets define a named set of patches that may be included by any
+	// resource in this Composition. PatchSets cannot themselves refer to other
+	// PatchSets.
+	//
+	// PatchSets are only used by the "Resources" mode of Composition. They
+	// are ignored by other modes.
```
phisco, line 66: I guess it's too late, but it starts to feel like we could have implemented composition through functions in a dedicated resource altogether.

apis/pkg/meta/v1alpha1/conversion.go
```diff
@@ -118,32 +113,3 @@ func (p *Provider) ConvertFrom(hub conversion.Hub) error {
...
-	in, ok := hub.(*Function)
-	if !ok {
-		return errors.New(errWrongConvertFromFunction)
-	}
-
-	conv := &GeneratedFromHubConverter{}
-	*f = *conv.Function(in)
-
-	return nil
-}
```
ezgidemirel, line 149: Don't we also want to keep meta/v1alpha1 just like provider and configuration packages? As a side note, I think we need a change in `up` cli to support the new version.

phisco (commented):
I still have to finish looking at the juicier parts, but I left a few minor comments here and there already, overall looks good so far! 🎉

apis/apiextensions/v1/composition_revision_types.go
negz: We did consider that back in the day. Using `Composition` feels right to me though. It's doing the same task as P&T Composition, we already have all the machinery to reference and select Compositions, etc.

apis/pkg/meta/v1alpha1/conversion.go
negz: No, I don't think so. We have `v1alpha1` for Provider and Configuration packages because there may be some packages in the wild who use the `meta.pkg.crossplane.io/v1alpha1` versions of their `crossplane.yaml` metadata. We know that's not the case with Functions.

apis/pkg/meta/v1beta1/function_interfaces.go
```diff
@@ -1,11 +1,25 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1beta1
 
 import (
-	"github.com/crossplane/crossplane/apis/pkg/meta/v1"
+	v1 "github.com/crossplane/crossplane/apis/pkg/meta/v1"
```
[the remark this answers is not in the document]
negz: I agree, but something (`gofmt`?) keeps putting it back for reasons I'm not sure of.

apis/pkg/meta/v1beta1/function_interfaces_test.go
```diff
@@ -0,0 +1,23 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1beta1
+
+import (
+	v1 "github.com/crossplane/crossplane/apis/pkg/meta/v1"
```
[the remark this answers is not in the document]
negz: See above.

cmd/crossplane/core/core.go
```diff
@@ -80,31 +82,33 @@ func (c *Command) Run() error {
...
-	TLSServerSecretName string        `help:"The name of the TLS Secret that will store Crossplane's server certificate." env:"TLS_SERVER_SECRET_NAME"`
-	TLSServerCertsDir   string        `help:"The path of the folder which will store TLS server certificate of Crossplane." env:"TLS_SERVER_CERTS_DIR"`
-	TLSClientSecretName string        `help:"The name of the TLS Secret that will be store Crossplane's client certificate." env:"TLS_CLIENT_SECRET_NAME"`
-	TLSClientCertsDir   string        `help:"The path of the folder which will store TLS client certificate of Crossplane." env:"TLS_CLIENT_CERTS_DIR"`
+	TLSServerSecretName  string `help:"The name of the TLS Secret that will store Crossplane's server certificate." env:"TLS_SERVER_SECRET_NAME"`
+	TLSServerCertsDir    string `help:"The path of the folder which will store TLS server certificate of Crossplane." env:"TLS_SERVER_CERTS_DIR"`
+	TLSClientSecretName  string `help:"The name of the TLS Secret that will be store Crossplane's client certificate." env:"TLS_CLIENT_SECRET_NAME"`
+	TLSClientCertsDir    string `help:"The path of the folder which will store TLS client certificate of Crossplane." env:"TLS_CLIENT_CERTS_DIR"`
```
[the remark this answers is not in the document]
negz: Yeah, just a strange diff of a regrouping. I grouped all the certname/certdir fields together.

internal/xfn/function_runner.go
```diff
@@ -0,0 +1,282 @@
...
+
+	dialFunctionTimeout = 10 * time.Second
+	runFunctionTimeout  = 10 * time.Second
+)
+
+// A PackagedFunctionRunner runs a Function by making a gRPC call to a Function
+// package's runtime. It creates a gRPC client connection for each Function. The
+// Function's endpoint is determined by reading the status.endpoint of the
+// active FunctionRevision. You must call GarbageCollectClientConnections in
+// order to ensure
```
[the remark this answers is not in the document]
negz: Whoops, yes.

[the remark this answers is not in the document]
negz: Yes, good catch. I previously had this returning `ctx.Err()` but then changed the signature a bit.

internal/controller/apiextensions/composite/reconciler.go
```diff
@@ -627,7 +619,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 	for i, cd := range res.Composed {
 		// Specifying a name for P&T templates is optional but encouraged.
 		// If there was no name, fall back to using the index.
-		id := cd.ResourceName
+		id := string(cd.ResourceName)
```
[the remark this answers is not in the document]
negz: I like to do this because it makes it obvious what the string map key is. e.g.:

```go
m: = map[ResourceName]Resource{}

// vs

n := map[string]Resource{}
```

internal/controller/apiextensions/composite/composition_render.go
```diff
@@ -0,0 +1,194 @@
...
+	o.SetName(name)
+	o.SetNamespace(namespace)
+
+	// This resource already had a GVK (probably because it already exists), but
+	// when we rendered its template it changed. This shouldn't happen. Either
+	// someone changed the kind in the template or we're trying to use the wrong
+	// template (e.g. because the order of an array of anonymous templates
+	// changed).
+	empty := schema.GroupVersionKind{}
+	if gvk != empty && o.GetObjectKind().GroupVersionKind() != gvk {
```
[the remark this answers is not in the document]
negz: Ah, nice. TIL.

internal/controller/apiextensions/composite/composition_render.go
```diff
@@ -0,0 +1,194 @@
...
+		xcrd.LabelKeyNamePrefixForComposed: xr.GetLabels()[xcrd.LabelKeyNamePrefixForComposed],
+		xcrd.LabelKeyClaimName:             xr.GetLabels()[xcrd.LabelKeyClaimName],
+		xcrd.LabelKeyClaimNamespace:        xr.GetLabels()[xcrd.LabelKeyClaimNamespace],
+	})
+
+	or := meta.AsController(meta.TypedReferenceTo(xr, xr.GetObjectKind().GroupVersionKind()))
+	return errors.Wrap(meta.AddControllerReference(cd, or), errSetControllerRef)
+}
+
+// TODO(negz): Is this really a 'renderer'?
```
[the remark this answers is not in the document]
negz: Yeah, potentially. I do kind of prefer being able to inject interfaces rather than concrete function signatures though, and it's nice to be able to inject this.

internal/xfn/function_runner.go
```diff
@@ -0,0 +1,283 @@
...
+
+	l := &pkgv1beta1.FunctionRevisionList{}
+	if err := r.client.List(ctx, l, client.MatchingLabels{pkgv1.LabelParentPackage: name}); err != nil {
+		return nil, errors.Wrapf(err, errListFunctionRevisions)
+	}
+
+	var active *pkgv1beta1.FunctionRevision
+	for i := range l.Items {
+		if l.Items[i].GetDesiredState() == pkgv1.PackageRevisionActive {
+			active = &l.Items[i]
```
turkenh, line 161:
```suggestion
			active = &l.Items[i]
			break
```
Considering that there could only be one active revision.

internal/controller/apiextensions/composite/composition_functions.go
```diff
@@ -0,0 +1,585 @@
...
+}
+
+// ObserveComposedResources begins building composed resource state by
+// fetching any existing composed resources referenced by the supplied composite
+// resource, as well as their connection details.
+func (g *ExistingComposedResourceObserver) ObserveComposedResources(ctx context.Context, xr resource.Composite) (ComposedResourceStates, error) {
+	ors := ComposedResourceStates{}
+
+	for _, ref := range xr.GetResourceReferences() {
+		// The PTComposer writes references to resources that it didn't actually
```
turkenh, line 437: Why do we care `PTComposer` here? I assume we won't support switching modes either.

apis/apiextensions/v1/composition_types.go
```diff
@@ -27,38 +27,59 @@ type CompositionSpec struct {
...
+	//
+	// "Pipeline" indicates that a Composition specifies a pipeline
+	// of Composition Functions, each of which is responsible for producing
+	// composed resources that Crossplane should create or update. THE PIPELINE
+	// MODE IS A BETA FEATURE. It is not honored if the relevant Crossplane
+	// feature flag is disabled.
+	// +optional
+	// +kubebuilder:validation:Enum=Resources;Pipeline
+	// +kubebuilder:default=Resources
+	Mode *CompositionMode `json:"mode,omitempty"`
```
turkenh, line 45: I assume this field would be immutable, i.e. it won't be possible to switch between composition modes?

If yes, should we mark it as immutable with CEL: https://kubernetes.io/blog/2022/09/29/enforce-immutability-using-cel/

internal/controller/apiextensions/composite/composition_functions.go
```diff
@@ -0,0 +1,585 @@
...
+	}
+
+	// TODO(negz): Both here and in the PTComposer we're relying on the fact
+	// that the calling Reconciler will update xr's status in order to persist
+	// any changes we make to it. It would be more obvious if we owned making
+	// our own changes.
+
+	// TODO(negz): Because we create, modify, and apply a new XR here we'll
+	// never apply any changes to status. I think this is what is causing
+	// https://github.com/crossplane/crossplane/issues/3812.
```
turkenh, line 300: Should a function be able to modify XR spec? Or should they only change the status? 
In the PTComposer, we allow modification of XR spec, however, this is against what is proposed in this ticket: https://github.com/crossplane/crossplane/issues/4581 and causing various problems regarding what to do next with that modified XR spec, e.g. Should it be propagated up to Claim? What if the Claim has the same spec with another value? 

I think this PR is a good chance to build to correct experience here which I believe is a function that should only be able to modify the status of an XR and we should explicitly return an error if done otherwise.

apis/apiextensions/v1/composition_types.go
phisco: Wouldn't making it immutable get the migration harder from one mode to the other?

internal/controller/apiextensions/composite/composition_functions.go
```diff
@@ -0,0 +1,585 @@
...
+				// introduced a severity that we don't know about.
+				events = append(events, event.Warning(reasonCompose, errors.Errorf("Pipeline step %q returned a result of unknown severity (assuming warning): %s", fn.Step, rs.Message)))
+			}
+		}
+	}
+
+	// TODO(negz): Both here and in the PTComposer we're relying on the fact
+	// that the calling Reconciler will update xr's status in order to persist
+	// any changes we make to it. It would be more obvious if we owned making
+	// our own changes.
```
phisco, line 296: +1, it took me a while to understand it, and it's clearly the cause of https://github.com/crossplane/crossplane/issues/3812, as you are suggesting below, we create a new xr one, but the caller is relying on the fact that we modify the resource in place. We should probably either move in here the status patching or explicitly return the xr as part of the CompositionResult

apis/apiextensions/v1/composition_revision_types.go
```diff
@@ -41,33 +41,64 @@ type CompositionRevisionSpec struct {
...
-
-	// Functions is list of Composition Functions that will be used when a
-	// composite resource referring to this composition is created. At least one
-	// of resources and functions must be specified. If both are specified the
-	// resources will be rendered first, then passed to the functions for
-	// further processing.
+	Resources []ComposedTemplate `json:"resources,omitempty"`
+
+	// Pipeline is a list of composition function steps that will be used when a
+	// composite resource referring to this composition is created. One of of
```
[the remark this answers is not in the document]
negz: Haha good catch. I thought I already fixed this, but it was in `composition_types.go`. :/

internal/xfn/function_runner.go
```diff
@@ -0,0 +1,283 @@
...
+	r.conns[name] = conn
+	r.connsMx.Unlock()
+
+	log.Debug("Created new gRPC client connection", "target", active.Status.Endpoint)
+	return conn, nil
+}
+
+// GarbageCollectConnections runs every interval until the supplied context is
+// cancelled. It garbage collects gRPC client connections to Functions that are
+// no longer installed.
```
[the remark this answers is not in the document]
negz: Were you thinking that we could pass the FunctionRunner into the package reconciler and have it delete Function clients when it noticed a Function was deleted? To be honest I didn't think of that. I think one challenge of doing it that way is that the Package controller seems pretty hard to add specialized code to. It wants to reconcile all types of packages (Provider, Configuration, Function) the same way and AFAICT doesn't have any hooks like the package revision controller does. We could refactor it, but that seems likely to be a lot of work.

I did initially explore passing an informer to this runner so that it could watch Functions and create/delete clients on-demand. I moved away from that approach mostly because:

* I wanted to create a client the first time a Function was called, so we could more easily return any errors setting up the client to the caller. (As opposed to reporting something like "error: no client exists for Function foo" because client-setup is done out-of-band.)
* This felt simpler to implement
* This takes a write lock slightly less often (e.g. if there are multiple Function clients to tear down in a 10 min period)

I could see us potentially splitting the difference. We could create clients on-demand but use an informer to delete them as soon as the Function was uninstalled. I suspect what is here is probably good enough though.

internal/controller/apiextensions/composite/composition_functions.go
```diff
@@ -0,0 +1,585 @@
...
+	errGetExistingCDs           = "cannot get existing composed resources"
+	errBuildObserved            = "cannot build observed state for RunFunctionRequest"
+	errBuildDesired             = "cannot build desired state for RunFunctionRequest"
+	errGarbageCollectCDs        = "cannot garbage collect composed resources that are no longer desired"
+	errApplyXR                  = "cannot apply composite resource"
+	errAnonymousCD              = "encountered composed resource without required \"" + AnnotationKeyCompositionResourceName + "\" annotation"
+	errUnmarshalDesiredXR       = "cannot unmarshal desired composite resource from RunFunctionResponse"
+	errXRAsStruct               = "cannot encode composite resource to protocol buffer Struct well-known type"
+
+	errFmtResourceName               = "composed resource %q"
```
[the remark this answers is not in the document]
negz: Thanks! I've noticed a super weird thing with my editor/linter in that _sometimes_ it will notice these unused constants but other times it won't. I swear it sometimes notices only when I start messing with the `const` block.

internal/controller/apiextensions/composite/composition_functions.go
```diff
@@ -0,0 +1,585 @@
...
+		or, ok := observed[ResourceName(name)]
+		if ok {
+			cd.SetNamespace(or.Resource.GetNamespace())
+			cd.SetName(or.Resource.GetName())
+		}
+
+		// TODO(negz): Should we try to automatically derive readiness if the
+		// Function returns READY_UNSPECIFIED? Is it safe to assume that if the
+		// Function doesn't have an opinion about readiness then we should look
+		// for the Ready: True status condition?
```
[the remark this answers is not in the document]
negz: Just to clarify, do you mean "is ready" or "check the ready condition to see if it's ready"?

apis/apiextensions/v1/composition_validation_test.go
```diff
@@ -33,6 +33,112 @@ func sortFieldErrors() cmp.Option {
...
+				output: field.ErrorList{field.Required(field.NewPath("spec", "resources"), "an array of resources is required in Resources mode (the default if no mode is specified)")},
+			},
+		},
+		"InvalidImplicitResources": {
+			reason: "A Composition with no explicit mode is assumed to be a Resources mode Composition, which is invalid without an array of resources",
+			args: args{
+				spec: CompositionSpec{},
+			},
+			want: want{
+				output: field.ErrorList{field.Required(field.NewPath("spec", "resources"), "an array of resources is required in Resources mode (the default if no mode is specified)")},
```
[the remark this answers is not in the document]
negz: Ah, I thought you meant that we didn't need to include details when we emit an error list. I think you're actually saying the value is ignored in the unit tests.

apis/apiextensions/v1/composition_types.go
negz: I think we do want to support switching modes. We could make `mode` immutable within a Composition, but it's still possible to change what Composition your XR uses. So you could switch from a `mode: Resources` Composition to a `mode: Pipeline` one.

I do think we should make it possible to migrate existing XRs to use Functions instead of P&T. Requiring folks to (re)create new XRs seems like it would hurt adoption.

internal/controller/apiextensions/composite/composition_functions.go
negz: Per https://github.com/crossplane/crossplane/pull/4500#discussion_r1322116042 I do think we should support switching modes. Or at least, somehow migrating existing P&T based XRs over to Functions.

So this code is covering the case where:

* An XR was using the `PTComposer`.
* The XR had failed to render a resource.
* The XR was migrated to the `FunctionComposer` before it had a chance to render the resource.

This does seem like a pretty slim edge case. On the other hand it's not too hard to have this little guard here.

negz: I've updated the comment to mention this explicitly.

negz: This does make me realize that there's another little edge case that we now don't cover:

* An XR is using a `mode: Resources` Composition with anonymous templates.
* The XR migrates to a `mode: Pipeline` Composition (which requires named resources)

We'll cut over to using the `FunctionComposer`, which assumes all existing composed resources will have the `crossplane.io/composition-resource-name` annotation. However, the composed resources won't have that annotation and we'll return an error here.

I think the `FallbackForAnonymoustemplates` composer used to guard against this case. I could be convinced that this is enough of an edge case we should cover it with just documentation. i.e. Tell folks that if they want to migrate to Functions they must first make sure all their templates are named (the PTComposer already has code that handles "migrating" from anonymous templates to named templates).

internal/controller/apiextensions/composite/composition_functions.go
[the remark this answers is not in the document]
negz: Also, I agree but let's tackle this in a follow-on PR.

internal/controller/apiextensions/composite/composition_functions.go
```diff
@@ -0,0 +1,585 @@
...
+	desired := ComposedResourceStates{}
+	for name, dr := range d.GetResources() {
+		cd := composed.New()
+		if err := FromStruct(cd, dr.GetResource()); err != nil {
+			return CompositionResult{}, errors.Wrapf(err, errFmtUnmarshalDesiredCD, name)
+		}
+
+		// If this desired resource state pertains to an existing composed
+		// resource we want to maintain its name and namespace. We generate the
+		// name,
```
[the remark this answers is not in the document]
negz: True! I can't actually remember what I wanted to say here so I've just deleted the dangling sentence, heh.

internal/controller/apiextensions/composite/composition_functions.go
negz: I've added a commit that fixes this. Per https://github.com/crossplane/crossplane/issues/4581 we now _only_ support updating XR status. For now I've only made this change in the `FunctionComposer` to avoid extra scope creep in this PR. 😄

internal/controller/apiextensions/composite/composition_functions.go
negz: I ended up adding a commit that addresses #4581 - or at least sets us up to address it when Functions are in use.

phisco (commented):
Still testing it out, but looks good so far

internal/controller/apiextensions/composite/composition_functions.go
```diff
@@ -0,0 +1,587 @@
...
+		return CompositionResult{}, errors.Wrap(err, errApplyXRRefs)
+	}
+
+	// Our goal here is to patch our XR's status using server-side apply. We
+	// want the resulting, patched object loaded into uxr. We need to pass in
+	// only our "fully specified intent" - i.e. only the fields that we actually
+	// care about. FromStruct will replace uxr's backing map[string]any with the
+	// content of GetResource (i.e. the desired status). We then need to set its
+	// GVK and name so that our client knows what resource to patch.
+	before := uxr.DeepCopy()
```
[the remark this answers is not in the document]
negz: Haha I had that at first but this felt more aesthetically pleasing. 😉 Changed it back.

phisco (approved):
LGTM! We can iterate in follow up PRs on the remaining TODOs. Thanks @negz! 🎉

internal/xfn/function_runner.go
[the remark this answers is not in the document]
turkenh: > I think one challenge of doing it that way is that the Package controller seems pretty hard to add specialized code to. It wants to reconcile all types of packages (Provider, Configuration, Function) the same way and AFAICT doesn't have any hooks like the package revision controller does.

FWIW, this would indeed be a better fit for the deactivation of a Function Revision so hooks on revisions would be a good place.

PR #4510, opened 2023-08-22
Fix Crossplane crash after creating an XRD with an empty schema

### LINKED ISSUE #4504: Controller crashes with panic after creating XRD
### What happened?
While dealing with issue https://github.com/komodorio/komoplane/issues/25, I have created an XRD that causes controller pod crash. I would expect Crossplane either to not allow me to create problematic resource, or handle it correctly. The controller pod gets into CrashLoopBackoff now.

The log of container contains panic:
```
I0815 12:15:19.954028       1 leaderelection.go:255] successfully acquired lease crossplane-system/crossplane-leader-election-core
Warning: ControllerConfig.pkg.crossplane.io/v1alpha1 has been deprecated and will be removed in a future release, but only after a comparable alternative is available.
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
    panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x192f16a]

goroutine 367 [running]:
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile.func1()
    sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:115 +0x1fa
panic({0x1ecd4e0, 0x34b0880})
    runtime/panic.go:884 +0x213
github.com/crossplane/crossplane/internal/xcrd.genCrdVersion({{0xc0007eee50, 0x8}, 0x1, 0x1, 0x0, 0x0, 0x0, {0x0, 0x0, 0x0}})
    github.com/crossplane/crossplane/internal/xcrd/crd.go:147 +0x66a
github.com/crossplane/crossplane/internal/xcrd.ForCompositeResource(0xc00041bb80)
    github.com/crossplane/crossplane/internal/xcrd/crd.go:74 +0x50a
github.com/crossplane/crossplane/internal/controller/apiextensions/definition.CRDRenderFn.Render(0xc00074b548?, 0xc00078c0c0?)
    github.com/crossplane/crossplane/internal/controller/apiextensions/definition/reconciler.go:109 +0x1f
github.com/crossplane/crossplane/internal/controller/apiextensions/definition.(*Reconciler).Reconcile(0xc00055d860, {0x24edde0, 0xc000728e10}, {{{0x0, 0x0}, {0xc00074b548, 0x15}}})
    github.com/crossplane/crossplane/internal/controller/apiextensions/definition/reconciler.go:262 +0x5e4
github.com/crossplane/crossplane-runtime/pkg/ratelimiter.(*Reconciler).Reconcile(0xc00020eaa0, {0x24edde0, 0xc000728e10}, {{{0x0?, 0x0?}, {0xc00074b548?, 0x40e007?}}})
    github.com/crossplane/crossplane-runtime@v0.20.1/pkg/ratelimiter/reconciler.go:54 +0x16b
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile(0x24edde0?, {0x24edde0?, 0xc000728e10?}, {{{0x0?, 0x1dcac20?}, {0xc00074b548?, 0x24da200?}}})
    sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:118 +0xc8
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler(0xc000251860, {0x24edd38, 0xc00020e7d0}, {0x1fb3300?, 0xc00039ac60?})
    sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:314 +0x377
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem(0xc000251860, {0x24edd38, 0xc00020e7d0})
    sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:265 +0x1d9
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2()
    sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:226 +0x85
created by sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2
    sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:222 +0x587
Stream closed EOF for crossplane-system/crossplane-8697f8cff4-tldm5 (crossplane)
```

### How can we reproduce it?
I have created clean Kind cluster, installed Crossplane via Helm there, then applied following YAML:
```yaml
---
metadata:
  name: eks1.test.komodor.com
spec:
  group: test.komodor.com
  names:
    kind: EKS
    plural: ekz
  versions:
    - name: v1alpha1
      served: true
      referenceable: true

```

Crossplane version: 1.13.2
Latest version of Kind on my Linux desktop.

ezgidemirel:

This PR adds nil checks for CRD version schema to prevent panic. 

Fixes #4504 

I have:

### DIFF
--- a/internal/xcrd/crd.go
+++ b/internal/xcrd/crd.go
@@ -42,11 +42,12 @@ const (
 )
 
 const (
-	errFmtGenCrd               = "cannot generate CRD for %q %q"
-	errParseValidation         = "cannot parse validation schema"
-	errInvalidClaimNames       = "invalid resource claim names"
-	errMissingClaimNames       = "missing names"
-	errFmtConflictingClaimName = "%q conflicts with composite resource name"
+	errFmtGenCrd                   = "cannot generate CRD for %q %q"
+	errParseValidation             = "cannot parse validation schema"
+	errInvalidClaimNames           = "invalid resource claim names"
+	errMissingClaimNames           = "missing names"
+	errFmtConflictingClaimName     = "%q conflicts with composite resource name"
+	errCustomResourceValidationNil = "custom resource validation cannot be nil"
 )
 
 // ForCompositeResource derives the CustomResourceDefinition for a composite
@@ -144,6 +145,11 @@ func genCrdVersion(vr v1.CompositeResourceDefinitionVersion) (*extv1.CustomResou
 	if err != nil {
 		return nil, errors.Wrapf(err, errParseValidation)
 	}
+
+	if s == nil {
+		return nil, errors.New(errCustomResourceValidationNil)
+	}
+
 	crdv.Schema.OpenAPIV3Schema.Description = s.Description
 
 	xSpec := s.Properties["spec"]
--- a/internal/xcrd/crd_test.go
+++ b/internal/xcrd/crd_test.go
@@ -23,6 +23,7 @@ limitations under the License.
 package xcrd
 
 import (
+	"fmt"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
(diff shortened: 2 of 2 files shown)


negz:
Do we want to support a version with an empty schema? This would result in an XR that only had Crossplane machinery fields, right? Obviously we shouldn't crash - I'm wondering though whether not providing any schema should result in an error.

lsviben:
> Do we want to support a version with an empty schema? This would result in an XR that only had Crossplane machinery fields, right? Obviously we shouldn't crash - I'm wondering though whether not providing any schema should result in an error.

Should we catch that during the validation? cc: @phisco 

BTW I am OK if we want to error out on empty schema here as well

phisco:
+1 for erroring out if the spec schema is missing. we can definitely check that in the webhook, still we should avoid crashing and maybe we should also think of running "logical validation" as a sanity check at the start of the reconciliation loop, just in case webhooks were disabled, as we are doing for compositions already.

ezgidemirel:
> Do we want to support a version with an empty schema? This would result in an XR that only had Crossplane machinery fields, right? Obviously we shouldn't crash - I'm wondering though whether not providing any schema should result in an error.

Updated the PR to return an error instead.

ezgidemirel:
We discussed with @phisco adding the validation in a separate PR since we'll backport this fix and may not want to backport the validation logic.

PR #4513, opened 2023-08-23
fix(helm): add conditionals around workload securityContexts

### LINKED ISSUE #4512: Unable to unset default securityContext using Helm values
### What happened?

We are attempting to install Crossplane on Openshift, and while setting `securityContext` is considered a best practice in the general Kubernetes community, Openshift thinks it does it better at least for unprivileged workloads. 😉 What I am trying to do _just using Helm values_ is to either:

1. Remove all securityContexts from the Crossplane workload specs.
2. Set all securityContexts to `{}` in the Crossplane workload specs.

And this seems currently impossible. If I set a securityContext-related value to `null`, it ends up as `null`. And if I set it to `{}`, I still get the nested default Helm chart values.

This is a very common problem when attempting to install Helm charts on Openshift. But I think the fix is common and simple: add conditionals around `securityContext` fields in the Helm templates. This will allow Openshift users to not set any securityContexts in workloads.

Related issues: https://github.com/crossplane/crossplane/issues/4289, https://github.com/crossplane/crossplane/issues/3264 and probably more...

Crossplane version: 1.13.2

erikgb:

This PR adds conditionals around all `securityContext` fields in Helm chart workload templates. This will allow Helm users to override the default securityContexts as described in #4512.

Fixes #4512

I have:

- ~[ ] Added or updated unit **and** E2E tests for my change.~
- ~[ ] Opened a PR updating the [docs], if necessary.~

### DIFF
--- a/cluster/charts/crossplane/templates/deployment.yaml
+++ b/cluster/charts/crossplane/templates/deployment.yaml
@@ -37,8 +37,10 @@ spec:
         release: {{ .Release.Name }}
         {{- include "crossplane.labels" . | indent 8 }}
     spec:
+      {{- with .Values.podSecurityContextCrossplane }}
       securityContext:
-        {{- toYaml .Values.podSecurityContextCrossplane | nindent 8 }}
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
       {{- if .Values.priorityClassName }}
       priorityClassName: {{ .Values.priorityClassName  | quote }}
       {{- end }}
@@ -61,8 +63,10 @@ spec:
           name: {{ .Chart.Name }}-init
           resources:
             {{- toYaml .Values.resourcesCrossplane | nindent 12 }}
+          {{- with .Values.securityContextCrossplane }}
           securityContext:
-            {{- toYaml .Values.securityContextCrossplane | nindent 12 }}
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
           env:
           - name: GOMAXPROCS
             valueFrom:
@@ -123,8 +127,10 @@ spec:
         - name: webhooks
           containerPort: 9443
         {{- end }}
+        {{- with .Values.securityContextCrossplane }}
         securityContext:
-          {{- toYaml .Values.securityContextCrossplane | nindent 12 }}
+          {{- toYaml . | nindent 12 }}
+        {{- end }}
         env:
           - name: GOMAXPROCS
             valueFrom:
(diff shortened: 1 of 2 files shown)


erikgb:
`make reviewable` seems to crash on my workstation with the following error, which seems unrelated to this PR:
````
cp: cannot stat '/home/erikbo/projects/github/crossplane/cluster/charts/crossplane/values.yaml.tmpl': No such file or directory
make[1]: *** [build/makelib/helm.mk:109: helm.prepare.crossplane] Error 1
make: *** [build/makelib/common.mk:440: reviewable] Error 2
````

I would also prefer to backport this fix, if accepted, but it doesn't seem like I am able to add labels. I left that item open in the checklist for now, but that breaks CI. 😄

**Comment by phisco** [comment]
@erikgb thanks for your PR! About the `make reviewable` issue you are having, you probably have to update the build submodule: `git submodule update --init`.

However I think it's not properly handling it:
```shell
$ helm template crossplane --set securityContextCrossplane={} cluster/charts/crossplane
```
Results in:
```yaml
...
          securityContext:
            - ""
...

erikgb:
> About the `make reviewable` issue you are having, you probably have to update the build submodule: `git submodule update --init`.

Ahhhh, we use those pesky git submodules. 😉 Thanks! Works now.

> 
> However I think it's not properly handling it:
> 
> ```shell
> $ helm template crossplane --set securityContextCrossplane={} cluster/charts/crossplane
> ```
> 
> Results in:
> 
> ```yaml
>           securityContext:
>             - ""
> ```

True, but how do we fix that? 🤔 Helm is a strange tool. Running `helm template crossplane --set securityContextCrossplane=null cluster/charts/crossplane` gives the correct/expected result. And that is what we would use. I'll take another look to support `{}` also.

phisco:
Maybe using a condition with empty? https://helm.sh/docs/chart_template_guide/function_list/#empty

erikgb:
> Maybe using a condition with empty? https://helm.sh/docs/chart_template_guide/function_list/#empty

I think your test case is hitting the limits of the Helm `--set` flag, ref. https://helm.sh/docs/intro/using_helm/#the-format-and-limitations-of---set, and that this will work if the value is expressed in a values file.

````yaml
Δ ~/projects/github/crossplane (fix-helm-securitycontext-openshift ⎈|k3d-flux) $ cat ./values.yaml 
securityContextCrossplane: {}
Δ ~/projects/github/crossplane (fix-helm-securitycontext-openshift ⎈|k3d-flux) $ helm template crossplane --values=./values.yaml cluster/charts/crossplane
````

yields that same as now - the chart default `securityContext` - which does not work on Openshift. Fixing this will require a rework of the Helm chart. This allows us to unset the chart default.

phisco (approved):

pedjak (approved):

PR #4525, opened 2023-08-25
feat: add build function subcommand to crank and image to meta functions spec

phisco:

Added `kubectl crossplane build function` subcommand, mirroring what we already have for `configurations` and `providers`.
Added `spec.image` to the meta type for Functions, mirroring `spec.controller.image` we have for Providers, which was required as we don't provide a `--controller` flag as we have in `up xpkg build`, to allow embedding the xpkg information into the base image containing the actual binaries.

I have:


phisco:
part of #3751

negz (approved):
Thanks!

I think longer term we'll want to steal the `xpkg` functionality from https://github.com/upbound/up so that we can build Functions where the package metadata and controller are just different layers in the same OCI image. For now this makes sense though.

lsviben (approved):

PR #4530, opened 2023-08-26
Deploy Functions with package manager

### LINKED ISSUE #4418: Add support for installing a `Function` to the package manager.
Deploy long-running functions with the package manager using the newly introduced types and mount the secrets

ezgidemirel:
With this change, we can start deploying long-running composition functions with the package manager. While deploying them, we start generating a server certificate and mounting it to the deployment which will be used by the gRPC server running in the function.

For each function package we're creating: 
- a k8s secret with name <package>-tls-server which contains `ca.crt`, `tls.crt`, `tls.key` keys
- a k8s service with name <package>

For each function package revision we're creating:
- a k8s service account
- a k8s deployment

This code has been tested with the following package:
```
kind: Function
metadata:
  name: philippe-function
spec:
  package: docker.io/phisco/function-jq:v0.0.7 
```
After seeing the package ready, I port forwarded the service with the following command:
```
kubectl port-forward svc/philippe-function 9443:9443
```
And ran the following command to call the function: 
```
grpcurl -cert client.crt -key client.key -cacert ca.crt -d @ function-jq.crossplane-system:9443 apiextensions.fn.proto.v1beta1.FunctionRunnerService.RunFunction <<EOM
{
  "desired": {
    "composite": {
      "resource": {
        "something": "something"
      }
    }
  }
}
EOM
```
I used certificates stored in `crossplane-tls-client` k8s secret and updated my `/etc/hosts` file to point `function-jq.crossplane-system` to localhost. 

After running the command, I got the following output:
```
{
  "desired": {
    "composite": {
      "resource": {
          "labels": {
                "labelizer.xfn.crossplane.io/processed": "true"
              },
          "something": "something"
        }
    }
  }
}
```

I verified that having different function revisions does not cause recreation of the tls-server secret or the k8s service.
I verified that deleting the function package also removes the corresponding resources.

Fixes #4418 

I have:


internal/controller/pkg/revision/hook.go
```diff
@@ -158,6 +165,9 @@ func (h *ProviderHooks) Post(ctx context.Context, pkg runtime.Object, pr v1.Pack
 	if err := initializer.NewTLSCertificateGenerator(h.namespace, initializer.RootCACertSecretName, *pr.GetTLSServerSecretName(), *pr.GetTLSClientSecretName(), pkgProvider.Name, initializer.TLSCertificateGeneratorWithOwner(owner)).Run(ctx, h.client); err != nil {
 		return errors.Wrapf(err, "cannot generate TLS certificates for %s", pkgProvider.Name)
 	}
+	if err := h.client.Apply(ctx, d); err != nil {
+		return errors.Wrap(err, errApplyProviderDeployment)
+	}
```
phisco, line 170: I guess this was moved here because you want to be sure providers are created after the certificates are generated, right?

ezgidemirel: yes, that's the reason

apis/pkg/v1alpha1/register.go
```diff
@@ -49,6 +49,14 @@ var (
 )
 
 // Function type metadata.
+var (
+	FunctionKind             = reflect.TypeOf(Function{}).Name()
+	FunctionGroupKind        = schema.GroupKind{Group: Group, Kind: FunctionKind}.String()
```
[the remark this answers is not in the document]
negz: I think we probably will. We could always do that in a follow-up PR though.

apis/pkg/v1alpha1/interfaces.go
```diff
@@ -0,0 +1,311 @@
...
+	f.Spec.IgnoreCrossplaneConstraints = b
+}
+
+// GetControllerConfigRef of this Function.
+func (f *Function) GetControllerConfigRef() *v1.ControllerConfigReference {
+	return nil
+}
+
+// SetControllerConfigRef of this Function.
+func (f *Function) SetControllerConfigRef(*v1.ControllerConfigReference) {}
```
negz, line 87: I don't think Functions should support ControllerConfigs. We know we don't want to pursue them, and we have [a design for their replacement](https://github.com/crossplane/crossplane/blob/51016ddd75c7a03f56e5da27edf55bab7847a9ee/design/one-pager-package-runtime-config.md).

I'm hoping we can build the runtime config implementation as essentially part of the Functions work for v1beta1. If not though, I'd prefer v1beta1 Functions to be pretty static controller config wise over propagating use of the ControllerConfig type.

internal/controller/pkg/revision/hook.go
```diff
@@ -215,6 +225,124 @@ func (h *ConfigurationHooks) Post(context.Context, runtime.Object, v1.PackageRev
...
+	if err := h.client.Delete(ctx, s); resource.IgnoreNotFound(err) != nil {
+		return errors.Wrap(err, errDeleteFunctionSA)
+	}
+
+	return nil
+}
+
+// Post creates a packaged function deployment, service account, service and secrets if the revision is active.
+//
+//nolint:gocyclo // TODO(ezgidemirel): Can this be refactored for less complexity?
```
negz, line 276: Nit: I prefer to put this at the end of the `func` line. That way the linter directive doesn't show up in the function's godoc. If you're worried about line length this can work too:

```go
func Example() error { //nolint:gocyclo // See below
    // TODO(ezgidemirel): ...
```

internal/controller/pkg/revision/reconciler.go
```diff
@@ -313,6 +319,56 @@ func SetupConfigurationRevision(mgr ctrl.Manager, o controller.Options) error {
...
+		WithRecorder(event.NewAPIRecorder(mgr.GetEventRecorderFor(name))),
+	)
+
+	return ctrl.NewControllerManagedBy(mgr).
+		Named(name).
+		For(&v1alpha1.FunctionRevision{}).
+		Owns(&appsv1.Deployment{}).
+		Watches(&v1alpha1.ControllerConfig{}, &EnqueueRequestForReferencingProviderRevisions{
+			client: mgr.GetClient(),
+		}).
```
negz, line 367: Is `EnqueueRequestForReferencingProviderRevisions` correct here?

apis/pkg/v1alpha1/function_types.go
```diff
@@ -40,8 +40,8 @@ type Function struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
-	Spec   FunctionSpec   `json:"spec"`
-	Status FunctionStatus `json:"status"`
+	Spec   FunctionSpec   `json:"spec,omitempty"`
+	Status FunctionStatus `json:"status,omitempty"`
 }
```
negz, line 46: FYI, down at line 93 you have `Endpoint` as a top-level field of `FunctionRevision`. I think that should be moved under `FunctionRevisionStatus`. I've done that in https://github.com/crossplane/crossplane/pull/4500 if you want to copy it.

negz (commented):
I think we might want to create a Service per revision. Not 100% sure yet, but my thinking is:

* Today we have a ControllerConfig per revision - i.e. two revisions can use two different ControllerConfigs.
* I imagine [`DeploymentRuntimeConfig`](https://github.com/crossplane/crossplane/blob/51016ddd75c7a03f56e5da27edf55bab7847a9ee/design/one-pager-package-runtime-config.md) would be the same - two revisions could use two different `DeploymentRuntimeConfig`s.
* `DeploymentRuntimeConfig` allows you to configure things about the Service and ServiceAccount.

apis/pkg/v1alpha1/interfaces.go
negz: If we feel pretty confident that we can implement the runtime config design in time for v1.14 I'd feel okay merging this as-is and refactoring before v1.14.

apis/pkg/v1alpha1/function_types.go
ezgidemirel: Moved the field to FunctionRevisionStatus.

internal/controller/pkg/revision/reconciler.go
ezgidemirel: Nope, it wasn't. I created `EnqueueRequestForReferencingFunctionRevisions` struct and implemented relevant methods.

internal/controller/pkg/revision/hook.go
ezgidemirel: done

negz:
> Should we use the revision name or the function package name?

I'm going to say revision name. 😄 My code in https://github.com/crossplane/crossplane/pull/4500 currently looks at the `status.endpoint` of the Function, so I imagine we'll need something like:

1. Revision controller creates a Service that uses its name.
2. Revision controller sets `status.endpoint` to the Service's address.
3. Function controller copies the `status.endpoint` from the active revision to its own `status.endpoint`.

phisco:
What would be the benefit of using a service per revision? Similarly, what's the benefit of using a Deployment per revision? I get we are doing the latter also for Providers, but Providers are different as Crossplane don't call them directly, while it's going to do so for Functions.  

Having a service/deployment per revision would mean that depending on the order we do things and the time each step takes, some calls might fail while we tear down the old revision, spin up the new one and update the endpoint. Updating the same service/deployment would allow us to smoothly handle transitions through rollouts.

I know the time spent rolling out new versions could be considered negligible, but given that the two approaches are equivalent in the effort they require to be implemented, I think we can discuss pros and cons before committing to one.

negz:
> Having a service/deployment per revision would mean that depending on the order we do things and the time each step takes, some calls might fail while we tear down the old revision, spin up the new one and update the endpoint. Updating the same service/deployment would allow us to smoothly handle transitions through rollouts.

Good point, I can see how having each revision update the same deployment and service rather than replacing them would be better.

Can we be consistent and make ProviderRevisions work this way too? It would be nice if all revisions managed their "runtime" resources (i.e. Deployments, etc) the same way. Seems like it could be useful if the Deployment is being used to serve webhooks. Plus it's just easier to reason about / less surprising.

I don't recall any specific reason that we create a new deployment per ProviderRevision - I'm guessing it was probably just cleaner "ownership chain" wise to do so. CC @hasheddan in case you remember.

> If the issue is with immutable fields, we can think of updating by default, but deleting and recreating on validation errors.

That seems a little fancy - could we keep it simple and just fail if you try to change an immutable field? Presumably the only way to do so would be by updating a ControllerConfig / DeploymentRuntimeConfig. I just did a quick skim of the Deployment, Service, ServiceAccount, and Secret API docs and I can't immediately find any immutable fields so this may be a non-issue.

phisco:
@negz do we want to do the switch in this PR for both functions and providers or should we merge this as is first and then switch both separately? I think the latter could be fine.

internal/controller/pkg/revision/deployment.go
```diff
@@ -265,136 +210,250 @@ func buildProviderDeployment(provider *pkgmetav1.Provider, revision v1.PackageRe
...
+								RunAsGroup:               &runAsGroup,
+								AllowPrivilegeEscalation: &allowPrivilegeEscalation,
+								Privileged:               &privileged,
+								RunAsNonRoot:             &runAsNonRoot,
+							},
+							Ports: []corev1.ContainerPort{
+								{
+									Name:          promPortName,
+									ContainerPort: promPortNumber,
+								},
```
phisco, line 302: "metrics"? prometheus? 🤔

ezgidemirel: yes, it's for metrics

phisco: but that's the only port being defined, we are missing the 9443, svc is still working fine, but it's usually better to define ports on containers nonetheless (dropping a message here too just not to forget)

internal/controller/pkg/revision/hook.go
```diff
@@ -215,6 +228,128 @@ func (h *ConfigurationHooks) Post(context.Context, runtime.Object, v1.PackageRev
...
+	// TODO(ezgidemirel): update any status fields relevant to package revisions.
+
+	// Do not clean up SA and controller if revision is not inactive.
+	if pr.GetDesiredState() != v1.PackageRevisionInactive {
+		return nil
+	}
+
+	// NOTE(hasheddan): we avoid fetching pull secrets and controller config as
+	// they aren't needed to delete Deployment, ServiceAccount, and Service.
+	s, d, _, _ := buildFunctionDeployment(pkgFunction, pr, nil, h.namespace, []corev1.LocalObjectReference{})
```
phisco, line 266: nit: maybe update the comment above and explain why service and server secret are not deleted here (which one could expect)

internal/controller/pkg/revision/hook.go
```diff
@@ -215,6 +228,128 @@ func (h *ConfigurationHooks) Post(context.Context, runtime.Object, v1.PackageRev
...
+		return errors.Wrap(err, errApplyFunctionSecret)
+	}
+	if err := h.client.Apply(ctx, d); err != nil {
+		return errors.Wrap(err, errApplyFunctionDeployment)
+	}
+	if pr.GetTLSServerSecretName() != nil && pr.GetTLSClientSecretName() != nil {
+		if err := initializer.NewTLSCertificateGenerator(h.namespace, initializer.RootCACertSecretName, *pr.GetTLSServerSecretName(), *pr.GetTLSClientSecretName(), pkgFunction.Name, initializer.TLSCertificateGeneratorWithOwner(owner)).GenerateServerCertificate(ctx, h.client); err != nil {
+			return errors.Wrapf(err, "cannot generate TLS certificates for %s", pkgFunction.Name)
+		}
+	}
```
phisco, line 311: nit: maybe switch the apply and the secret initialisation, just to make it similar to the hook for providers

internal/controller/pkg/revision/hook.go
```diff
@@ -215,6 +228,128 @@ func (h *ConfigurationHooks) Post(context.Context, runtime.Object, v1.PackageRev
...
+		return errors.Wrap(err, errApplyFunctionSA)
+	}
+	owner := pr.GetOwnerReferences()
+	if err := h.client.Apply(ctx, secSer); err != nil {
+		return errors.Wrap(err, errApplyFunctionSecret)
+	}
+	if err := h.client.Apply(ctx, d); err != nil {
+		return errors.Wrap(err, errApplyFunctionDeployment)
+	}
+	if pr.GetTLSServerSecretName() != nil && pr.GetTLSClientSecretName() != nil {
```
phisco, line 307: Why is the tls client secret populated for function revisions? it's not actually created.

ezgidemirel: We're calling `GenerateServerCertificate` for functions and not generation client certificates. Since `NewTLSCertificateGenerator` gets client secret name, I added a nil check there.

internal/controller/pkg/revision/hook.go
ezgidemirel: I can do it in the following PR where I'll start creating resources per packages, not revisions.

phisco (approved):
Gone once more through everything, the only relevant comment I have is https://github.com/crossplane/crossplane/pull/4530#discussion_r1311735035, plus as we discussed it would be nice to define a test suite even if no specific test cases at the moment, just to run all the base e2es against Crossplane with composition functions enabled. Otherwise lgtm! awesome work @ezgidemirel!

internal/controller/pkg/revision/hook.go
phisco: `NewTLSCertificateGenerator` should probably not require the name of both, but rather get some options, so that we don't have to fake it. I'd rather also not have `spec.webhookTLSSecretName`, nor `spec.tlsClientSecretName`, set if these are not actually used here, as that's misleading.

internal/controller/pkg/revision/hook.go
phisco: why not in this one? it's just a comment 😄

negz:
> I think the latter could be fine.

Yeah that was my thinking too - would be my preference as it unblocks me to start on E2E tests. 😄

internal/controller/pkg/revision/hook.go
ezgidemirel: done

internal/controller/pkg/revision/hook.go
ezgidemirel: made them option and removed the nil checks

PR #4538, opened 2023-08-29
fix(chart): explicitly set resourceFieldRef.divisor to avoid flapping

### LINKED ISSUE #4509: Deploying crossplane helm chart with GitOps shows out of sync
### What happened?

When deploying crossplane helm chart with argocd it always shows out of sync for 2 Deployment (crossplane, crossplane-rbac-manager)

image

The out of sync is on `env.valueFrom.resourceFieldRef.divisor`
In the helm chart the divisor is not set, and kubernetes mutates the Deployment to "0" which is the same as 1 or "1"

I proposed we set the `divisor` in the helm char to 1

```yaml
- env:
  - name: GOMAXPROCS
    valueFrom:
      resourceFieldRef:
        containerName: nginx
        divisor: "0"
        resource: limits.cpu
- name: GOMEMLIMIT
  valueFrom:
    resourceFieldRef:
      containerName: nginx
      divisor: "0"
      resource: limits.memory
```

### How can we reproduce it?
To reproduce deploy ArgoCD Application using crossplane chart
```yaml
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  name: addon-ex-crossplane-crossplane
  namespace: argocd
spec:
  destination:
    name: ex-crossplane
    namespace: crossplane-system
  project: default
  sources:
  - chart: crossplane
    helm:
      releaseName: crossplane
    repoURL: https://charts.crossplane.io/stable
    targetRevision: 1.13.2
  syncPolicy:
    automated: {}
    syncOptions:
    - CreateNamespace=true
    - ServerSideApply=true
```

Crossplane version: 1.13.2

### References
- https://github.com/kubernetes/kubernetes/blob/master/pkg/api/v1/resource/helpers.go#L343-L380
- [resource.Quantity](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/types.go#L2028-L2030)

phisco:

Fixes #4509. 

I have:

### DIFF
--- a/cluster/charts/crossplane/templates/deployment.yaml
+++ b/cluster/charts/crossplane/templates/deployment.yaml
@@ -69,11 +69,13 @@ spec:
               resourceFieldRef:
                 containerName: {{ .Chart.Name }}-init
                 resource: limits.cpu
+                divisor: "1"
           - name: GOMEMLIMIT
             valueFrom:
               resourceFieldRef:
                 containerName: {{ .Chart.Name }}-init
                 resource: limits.memory
+                divisor: "1"
           - name: POD_NAMESPACE
             valueFrom:
               fieldRef:


cluster/charts/crossplane/templates/deployment.yaml
```diff
@@ -69,11 +69,13 @@ spec:
               resourceFieldRef:
                 containerName: {{ .Chart.Name }}-init
                 resource: limits.cpu
+                divisor: "0"
```
turkenh, line 72: Shouldn't we set to its default, which is documented as "1" [here](https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/#store-container-fields)? 

> The divisor field is optional and has the default value of 1. A divisor of 1 means cores for cpu resources, or bytes for memory resources.

phisco: Yeah, that was my understanding too, I found a lot of examples setting it to "0", but couldn't really get why's that. I'll dig into it a bit more.

phisco: The current chart results in a default of "0", because the field is a quantity and is not a pointer, but it's not defaulted per se, so they end up handling it as "1" in code: https://github.com/kubernetes/kubernetes/blob/34b85c593d13168edee426c63e97dbfbc760395b/pkg/api/v1/resource/helpers.go#L345C1-L347C10

I'd say that although we are currently implicitly setting it to 0, we can set it to 1 to be more explicit.

phisco: Updated accordingly

turkenh (approved):
Thanks @phisco 🙌

PR #4547, opened 2023-08-30
feat(environmentConfig): allow setting sourceFieldPath label selector optional

### LINKED ISSUE #3723: Allow ignoring environment config if it does not exists

phisco:

Fixes #3723 .

I have:

### DIFF
--- a/apis/apiextensions/v1/composition_environment.go
+++ b/apis/apiextensions/v1/composition_environment.go
@@ -239,8 +239,21 @@ type EnvironmentSourceSelectorLabelMatcher struct {
 	// ValueFromFieldPath specifies the field path to look for the label value.
 	ValueFromFieldPath *string `json:"valueFromFieldPath,omitempty"`
 
+	// FromFieldPathPolicy specifies how to patch from a field path. The default is
+	// 'Required', which means the patch should fail if the path specified via valueFromFieldPath does not exist.
+	// Use 'Ignore' if instead you want it to result in a no-op.
+	// +kubebuilder:validation:Enum=Optional;Required
+	// +optional
+	FromFieldPathPolicy *FromFieldPathPolicy `json:"policy,omitempty"`
+
 	// Value specifies a literal label value.
 	Value *string `json:"value,omitempty"`
+}
+
+// FromFieldPathIsOptional returns true if the FromFieldPathPolicy is set to
+// Optional.
+func (e *EnvironmentSourceSelectorLabelMatcher) FromFieldPathIsOptional() bool {
+	return e.FromFieldPathPolicy != nil && *e.FromFieldPathPolicy == FromFieldPathPolicyOptional
 }
 
 // GetType returns the type of the label matcher, returning the default if not set.
--- a/cluster/crds/apiextensions.crossplane.io_compositionrevisions.yaml
+++ b/cluster/crds/apiextensions.crossplane.io_compositionrevisions.yaml
@@ -108,6 +108,17 @@ spec:
                                   key:
                                     description: Key of the label to match.
                                     type: string
+                                  policy:
+                                    description: FromFieldPathPolicy specifies how
+                                      to patch from a field path. The default is 'Required',
+                                      which means the patch should fail if the path
+                                      specified via valueFromFieldPath does not exist.
+                                      Use 'Ignore' if instead you want it to result
+                                      in a no-op.
+                                    enum:
+                                    - Optional
+                                    - Required
+                                    type: string
                                   type:
                                     default: FromCompositeFieldPath
                                     description: Type specifies where the value for
@@ -1602,6 +1613,17 @@ spec:
                                 properties:
                                   key:
                                     description: Key of the label to match.
+                                    type: string
+                                  policy:
+                                    description: FromFieldPathPolicy specifies how
+                                      to patch from a field path. The default is 'Required',
+                                      which means the patch should fail if the path
+                                      specified via valueFromFieldPath does not exist.
+                                      Use 'Ignore' if instead you want it to result
+                                      in a no-op.
+                                    enum:
+                                    - Optional
+                                    - Required
                                     type: string
                                   type:
                                     default: FromCompositeFieldPath
--- a/internal/controller/apiextensions/composite/environment/selector.go
+++ b/internal/controller/apiextensions/composite/environment/selector.go
@@ -114,15 +114,18 @@ func (s *APIEnvironmentSelector) buildEnvironmentConfigRefFromRef(ref *v1.Enviro
 }
 
 func (s *APIEnvironmentSelector) lookUpConfigs(ctx context.Context, cr resource.Composite, ml []v1.EnvironmentSourceSelectorLabelMatcher) (*v1alpha1.EnvironmentConfigList, error) {
+	res := &v1alpha1.EnvironmentConfigList{}
 	matchLabels := make(client.MatchingLabels, len(ml))
 	for i, m := range ml {
 		val, err := ResolveLabelValue(m, cr)
 		if err != nil {
+			if fieldpath.IsNotFound(err) && m.FromFieldPathIsOptional() {
+				return res, nil
+			}
 			return nil, errors.Wrapf(err, errFmtResolveLabelValue, i)
 		}
 		matchLabels[m.Key] = val
 	}
-	res := &v1alpha1.EnvironmentConfigList{}
 	if err := s.kube.List(ctx, res, matchLabels); err != nil {
 		return nil, errors.Wrap(err, errListEnvironmentConfigs)
 	}
@@ -132,7 +135,7 @@ func (s *APIEnvironmentSelector) lookUpConfigs(ctx context.Context, cr resource.
 func (s *APIEnvironmentSelector) buildEnvironmentConfigRefFromSelector(cl *v1alpha1.EnvironmentConfigList, selector *v1.EnvironmentSourceSelector) ([]corev1.ObjectReference, error) {
 	ec := make([]v1alpha1.EnvironmentConfig, 0)
 
-	if len(cl.Items) == 0 {
+	if cl == nil || len(cl.Items) == 0 {
 		return []corev1.ObjectReference{}, nil
 	}
 
--- a/internal/controller/apiextensions/composite/environment/selector_test.go
+++ b/internal/controller/apiextensions/composite/environment/selector_test.go
@@ -514,10 +514,8 @@ func TestSelect(t *testing.T) {
 		"NoErrorOnInvalidOptionalLabelValueFieldPath": {
 			reason: "It should not return an error if the path to a label value is invalid, but was set as optional.",
 			args: args{
-				kube: &test.MockClient{
-					MockList: test.NewMockListFn(nil),
-				},
-				cr: composite(),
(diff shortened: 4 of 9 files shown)


MisterMX (approved):

internal/controller/apiextensions/composite/environment/selector.go
```diff
@@ -132,7 +135,7 @@ func (s *APIEnvironmentSelector) lookUpConfigs(ctx context.Context, cr resource.
 func (s *APIEnvironmentSelector) buildEnvironmentConfigRefFromSelector(cl *v1alpha1.EnvironmentConfigList, selector *v1.EnvironmentSourceSelector) ([]corev1.ObjectReference, error) {
 	ec := make([]v1alpha1.EnvironmentConfig, 0)
 
-	if len(cl.Items) == 0 {
+	if cl == nil || len(cl.Items) == 0 {
```
pedjak, line 138: should be a regular case to pass `nil` for `cl` in this function? `lookUpConfig` never returns `nil`, `nil` so perhaps we should return error instead? Otherwise, could we add selector unit test for this change?

internal/controller/apiextensions/composite/environment/selector_test.go
```diff
@@ -510,6 +511,41 @@ func TestSelect(t *testing.T) {
 				err: errors.Wrapf(errors.Wrapf(errors.New("wrong: no such field"), errFmtResolveLabelValue, 0), errFmtReferenceEnvironmentConfig, 0),
 			},
 		},
+		"NoErrorOnInvalidOptionalLabelValueFieldPath": {
+			reason: "It should not return an error if the path to a label value is invalid, but was set as optional.",
+			args: args{
+				kube: &test.MockClient{
+					MockList: test.NewMockListFn(nil),
```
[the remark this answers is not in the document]
phisco: It's enough to not define it and it will panic.

internal/controller/apiextensions/composite/environment/selector.go
phisco: It should never happen but I feel it's ok to handle it as if it had 0 items.

internal/controller/apiextensions/composite/environment/selector_test.go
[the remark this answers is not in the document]
phisco: updated test accordingly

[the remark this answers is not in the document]
phisco: However I would argue that would not be a regression, I don't care if under the hood the function calls a list or not before returning, I only care this one is accepted, while it's erroring out if set to required. But that's mostly a matter of taste and mocking does push toward such an approach.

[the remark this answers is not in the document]
pedjak: IMHO, checking how the component under tests interacts with dependencies is important (on unit-test level). If a path does not exist and the label cannot be constructed, then interacting with cluster should never happen.

internal/controller/apiextensions/composite/environment/selector.go
pedjak: If it should never happen, then passing `nil` here might smell like a bug, and failing early would uncover it. 

https://en.wikipedia.org/wiki/Design_by_contract

phisco: Then let me rephrase: "it can happen and should be handled as if it was an empty list"

pedjak: ok, can you please add selector unit test for that situation?

pedjak (approved):

apis/apiextensions/v1/composition_environment.go
```diff
@@ -239,10 +239,23 @@ type EnvironmentSourceSelectorLabelMatcher struct {
 	// ValueFromFieldPath specifies the field path to look for the label value.
 	ValueFromFieldPath *string `json:"valueFromFieldPath,omitempty"`
 
+	// FromFieldPathPolicy specifies how to patch from a field path. The default is
+	// 'Required', which means the patch should fail if the path specified via valueFromFieldPath does not exist.
+	// Use 'Ignore' if instead you want it to result in a no-op.
+	// +kubebuilder:validation:Enum=Optional;Required
```
turkenh, line 245:
```suggestion
	// +kubebuilder:validation:Enum=Optional;Required
	// +kubebuilder:default=Required
```

apis/apiextensions/v1/composition_environment.go
```diff
@@ -239,10 +239,23 @@ type EnvironmentSourceSelectorLabelMatcher struct {
 	// ValueFromFieldPath specifies the field path to look for the label value.
 	ValueFromFieldPath *string `json:"valueFromFieldPath,omitempty"`
 
+	// FromFieldPathPolicy specifies how to patch from a field path. The default is
+	// 'Required', which means the patch should fail if the path specified via valueFromFieldPath does not exist.
+	// Use 'Ignore' if instead you want it to result in a no-op.
```
turkenh, line 244: Is it `Optional vs Required` or `Ignore vs Required`?

apis/apiextensions/v1/composition_environment.go
```diff
@@ -239,10 +239,23 @@ type EnvironmentSourceSelectorLabelMatcher struct {
...
+	// +kubebuilder:validation:Enum=Optional;Required
+	// +optional
+	FromFieldPathPolicy *FromFieldPathPolicy `json:"policy,omitempty"`
```
turkenh, line 247:
```suggestion
	FromFieldPathPolicy *FromFieldPathPolicy `json:"fromFieldPathPolicy,omitempty"`
```

Should we make it more obvious?

Otherwise, this looks like below in the API, right?

```yaml
          matchLabels:
              key: stage
              valueFromFieldPath: spec.someValue
              policy: Optional
```

This looks like matching this label (i.e. `matchLabels[0]`) is `Optional`, which I believe is kind of true if `spec.someValue` is not there, we ignore this label. However, it could be confusing in the sense that, even that value is there, this label is optional, which doesn't make sense though. For example, things get interesting if this is used in conjunction with `value` instead:

```yaml
        matchLabels:
          - type: Value
            key: stage
            value: prod
            policy: Optional
```

cluster/crds/apiextensions.crossplane.io_compositionrevisions.yaml
```diff
@@ -1603,6 +1614,17 @@ spec:
                                   key:
                                     description: Key of the label to match.
                                     type: string
+                                  policy:
+                                    description: FromFieldPathPolicy specifies how
```
turkenh, line 1618: `policy` vs `FromFieldPathPolicy` in the description of the API

internal/controller/apiextensions/composite/environment/selector.go
```diff
@@ -114,15 +114,18 @@ func (s *APIEnvironmentSelector) buildEnvironmentConfigRefFromRef(ref *v1.Enviro
 }
 
 func (s *APIEnvironmentSelector) lookUpConfigs(ctx context.Context, cr resource.Composite, ml []v1.EnvironmentSourceSelectorLabelMatcher) (*v1alpha1.EnvironmentConfigList, error) {
+	res := &v1alpha1.EnvironmentConfigList{}
 	matchLabels := make(client.MatchingLabels, len(ml))
 	for i, m := range ml {
 		val, err := ResolveLabelValue(m, cr)
 		if err != nil {
+			if fieldpath.IsNotFound(err) && m.FromFieldPathIsOptional() {
+				return res, nil
```
turkenh, line 123: Shouldn't we `continue` here instead of returning res which effectively has an empty environment config list?

phisco: I had the same doubt, I decided to make it "no-op" the whole environment, as I thought that omitting a single value would have been more confusing. This was my thought:
- if a single label selector is specified and marked as optional, we should return no environment at all.
- if `n>1` selectors are specified, one of them is marked as optional, if not found I'd rather have it result in a "no-op" than selecting an environment selected only through the `n-1` remaining selectors.
But I agree it's debatable, what would you suggest?

turkenh: > if n>1 selectors are specified, one of them is marked as optional, if not found I'd rather have it result in a "no-op" than selecting an environment selected only through the n-1 remaining selectors.

Since the policy is defined inside an element of `matchLabels` array, I would expect it only to configure the behavior for that single element. Take the following example:

```yaml
          matchLabels:
            - type: Value
              key: cloud-provider
              value: aws
              key: stage
              valueFromFieldPath: spec.someValue
              fromFieldPathPolicy: Optional
```

I would expect this still to filter `cloud-provider=aws` if `spec.someValue` is absent.

> if a single label selector is specified and marked as optional, we should return no environment at all

This definitely makes sense, instead of selecting all by considering this as "no selector provided".

However, I see the inconsistency between the two cases, if a single element - we filter all out, if multiple elements, we filter all in from the entry with policy POV. 
Also it is not only 1 vs n selectors being provided, e.g. what happens if multiple - all optional / none resolved:

```yaml
          matchLabels:
              key: cloud-provider
              valueFromFieldPath: spec.anotherValue
              fromFieldPathPolicy: Optional
              key: stage
              valueFromFieldPath: spec.someValue
              fromFieldPathPolicy: Optional
```

So, I agree that there is no perfect answer, but I still lean on policy impacting that single entry instead of invalidating all others. Otherwise, we may consider moving the policy up instead to make it more explicit:

```yaml
    environmentConfigs:
      - ref:
          name: example-environment-1
      - type: Selector
        selector:
          compositeValueResolutionPolicy: Optional # This environment config selector will be ignored if composite value resolution fails for any label
          matchLabels:
              key: stage
              valueFromFieldPath: metadata.labels[stage]
```

PR #4555, opened 2023-09-01
feat(environment): Environment init data

### LINKED ISSUE #4274: Add support for default environment in Composition
When using the Composition environment it would be very helpful to have a method for defining a default set of values that the environment should be initialized with before selection of EnvironmentConfigs.  This could be used to simplify patching of constant values throughout the composition, or to provide default values that may or may not be overridden by the selected EnvironmentConfig(s)

Add support for a "default" or "initial" environment context that would serve as the starting point for all subsequent EnvironmentConfig selection and patching.  The selected EnvironmentConfig(s) should be merged into the default environment, overwriting duplicate keys and leaving existing keys untouched.

A mergeOptions policy is probably necessary to specify the behavior for `appendSlice` and `keepMapValues`

There has also been discussion of adding support for a schema definition for the Composition environment, which would make it easier to understand and validate the expected environment structure, and could also provide a mechanism for specifying default values.  These two features can probably exist in parallel, with the schema validating the initial configuration as well as specifying default values.

MisterMX:

Add a new field `spec.environment.initData` to composition that allows prefilling the environment with default values before any environment config is loaded.

In YAML it would look like this:

```yaml
spec:
  environment:
    initData:
      region: us-west-1
      accountId: 12345665
```

Fixes #4274 

This also contains a bit refactoring, most prominently moving the environment out of its own package into the regular controller package so all code can be shared easily without having to worry about circular dependencies.

I have:


internal/controller/apiextensions/composite/environment_fetcher.go
```diff
@@ -77,19 +77,24 @@ type APIEnvironmentFetcher struct {
 //
 // Note: The `.Data` path is trimmed from the result so its necessary to include
 // it in patches.
-func (f *APIEnvironmentFetcher) Fetch(ctx context.Context, cr resource.Composite, required bool) (*Environment, error) {
+func (f *APIEnvironmentFetcher) Fetch(ctx context.Context, req EnvironmentFetcherRequest) (*Environment, error) {
 	var env *Environment
 
 	// Return an empty environment if the XR references no EnvironmentConfigs.
```
turkenh, line 83: nit:

```suggestion
	// Return an empty or default environment if the XR references no EnvironmentConfigs.
```

internal/controller/apiextensions/composite/environment_fetcher.go
```diff
@@ -77,19 +77,24 @@ type APIEnvironmentFetcher struct {
...
-	if len(cr.GetEnvironmentConfigReferences()) == 0 {
+	if len(req.Composite.GetEnvironmentConfigReferences()) == 0 {
```
turkenh, line 84: nitpick: 

I believe this piece of code could be refactored/simplified by dropping if/else here and having a single flow, this way we could also avoid handling initData in two different places. 
I am imagining a flow like below:

- loadedConfigs := []*v1alpha1.EnvironmentConfig{}
- if initData defined => loadedConfigs = append(loadedConfigs,initData)
- loadedConfigs = append(loadedConfigs, composite.EnvironmentConfigRefs...)
- if len(loadedConfigs) > 1 => merge
- return

turkenh (commented):
Thanks for tackling this @MisterMX, looking good 🙌  Just left some nitpicks.

Another alternative API would be to have`spec.environment.defaults`:

```yaml
spec:
  environment:
    defaults:
      region: us-west-1
      accountId: 12345665
```

This would indicate that they are default environment values but I believe you have chosen `initData` to better signal that it initializes the `data` of environment configs which makes sense as well. No strong opinion, just throwing it out as another possibility in case someone comes up with some good arguments on one or another :)

internal/controller/apiextensions/composite/environment_fetcher.go
```diff
@@ -27,14 +27,14 @@ import (
...
 
-	"github.com/crossplane/crossplane-runtime/pkg/resource"
-
 	v1alpha1 "github.com/crossplane/crossplane/apis/apiextensions/v1alpha1"
 )
 
 const (
-	errGetEnvironmentConfig = "failed to get config set from reference"
-	errMergeData            = "failed to merge data"
+	errGetEnvironmentInitData  = "cannot get environment init data"
```
phisco, line 34: Looks like this is unused.
```suggestion
```

internal/controller/apiextensions/composite/environment_fetcher.go
```diff
@@ -104,37 +109,61 @@ func (f *APIEnvironmentFetcher) Fetch(ctx context.Context, cr resource.Composite
...
-		},
-	}, nil
+	return loadedConfigs, nil
 }
 
-func mergeEnvironmentData(configs []v1alpha1.EnvironmentConfig) (map[string]interface{}, error) {
+func mergeEnvironmentData(configs []*v1alpha1.EnvironmentConfig) (map[string]interface{}, error) {
 	merged := map[string]interface{}{}
 	for _, e := range configs {
 		if e.Data == nil {
```
phisco, line 169: This could panic if passed a nil element, can you add a check on `e` too? or maybe just switch back everything not to use pointers altogether if there was no strong reason to do so.

MisterMX: This is just a little micro-optimization: I used a pointer here to have less data copied between function calls or iterations. It is only used within the same file and there is no chance where nil pointers are added to the list.

phisco:
> I am also not very happy with `initData` and I am open to change it for something else. Maybe `defaultData`?

+1 for `defaultData`

MisterMX:
Renamed `InitData` to `DefaultData`

```yaml
spec:
  environment:
    defaultData:
      region: us-west-1
      accountId: 12345665
```

phisco (approved):
Thanks @MisterMX! 🚀

turkenh (approved):

PR #4556, opened 2023-09-01
feat(environment): Validate environment schema at runtime

### LINKED ISSUE #4275: Add support for Composition environment schema definition/validation
The Composition environment has no defined schema and can be used as a generic object/scratchpad.  While this flexibility is extremely useful it does have drawbacks:
- it is difficult to find typographical errors when referencing environment variables
- it is difficult to understand the environment structure of the composition without reading the entire composition

Allow users to define an environment schema within the Composition.  This would be useful for:
- documenting the expected environment structure
- validating that the generated environment is "correct" as defined by the schema
- validating environment patches are referencing correct attributes
- providing default values to the environment

MisterMX:

Add a new field `spec.environment.schema` that can be used to validate the generated environment. It relies on the already included [`k8s.io/kube-openapi`](https://pkg.go.dev/k8s.io/kube-openapi) package so it uses [JSON Schema](http://json-schema.org/).

In YAML it would look like this:

```yaml
spec:
  environment:
    schema:
      jsonSchema:
        hello:
          type: string
```

Example environment would be something like:

```yaml
hello: world
```

Fixes #4275 

Depends on the refactoring changes in #4555

I have:


pedjak:
should this PR be closed in favor of https://github.com/crossplane/crossplane/pull/4555 ?

MisterMX:
> should this PR be closed in favor of #4555 ?

No, its only depends on the other PR because there was some refactoring done.

apis/apiextensions/v1/composition_environment.go
```diff
@@ -360,3 +365,9 @@ func (e *EnvironmentPatch) Validate() *field.Error {
 	}
 	return p.Validate()
 }
+
+// EnvironmentConfigSchema defines schema information for an environment.
+type EnvironmentConfigSchema struct {
+	// JSONSchema to validate the environment against.
+	JSONSchema extv1.JSON `json:"jsonSchema,omitempty"`
```
pedjak, line 372: https://github.com/crossplane/crossplane/issues/4275 speaks about `openAPIV3Schema`, can we rename the field to match the proposal?

internal/controller/apiextensions/composite/environment_fetcher.go
```diff
@@ -185,3 +194,14 @@ func mergeMaps(a, b map[string]interface{}) map[string]interface{} {
...
+
+func validateEnvironment(req EnvironmentFetcherRequest, env *Environment) error {
+	if req.Revision == nil || req.Revision.Spec.Environment == nil || req.Revision.Spec.Environment.Schema == nil {
+		return nil
+	}
+	schema := &openapispec.Schema{}
+	if err := json.Unmarshal(req.Revision.Spec.Environment.Schema.JSONSchema.Raw, schema); err != nil {
+		return errors.Wrap(err, errParseEnvironmentSchema)
+	}
+	return openapivalidate.AgainstSchema(schema, env.Object, openapistrfmt.Default)
```
pedjak, line 206: how do we handle the situation when `env` is `nil`?

pedjak (changes requested):
we need unit tests asserting the added functionality.

internal/controller/apiextensions/composite/environment_fetcher.go
MisterMX: `env` is never nil because `fetchEnvironment` always returns a value (if not an error).

apis/apiextensions/v1/composition_environment.go
MisterMX: The `openAPIV3Schema` described in https://github.com/crossplane/crossplane/issues/4275#issuecomment-1700590562 is only a proposal by me. I changed it to `jsonSchema` because the parser expects an actual JSONSchema and not an OpenAPI schema.

internal/controller/apiextensions/composite/environment_fetcher.go
pedjak: in that case, we could change the function signature to

```go
func validateEnvironment(req EnvironmentFetcherRequest, env Environment) error
```

apis/apiextensions/v1/composition_environment.go
pedjak: since we use openAPI schema in other places in crossplane and kubernetes, could we avoid a drift here?

phisco:
@MisterMX so this is validating the merged result of the selected EnvironmentConfig, but not the in-memory environment after applying patches, right?

MisterMX:
> @MisterMX so this is validating the merged result of the selected EnvironmentConfigs, but not the in-memory environment after applying patches, right?

Yes, because patches have their own `policy` control - although it is theoretically possible to implement runtime validation as well. It could also be validated statically within https://github.com/crossplane-contrib/crossplane-lint.

phisco:
Indeed I would leave patch validation to the webhook/crossplane-lint, I was more referring to an after the fact validation, if your patching results in an invalid environment in the end according to the schema you defined we should let users know, I guess. Or is this just a schema for the "merged" environment? If that's the case we should be more explicit.

apis/apiextensions/v1/composition_environment.go
MisterMX: I would keep it since it is very straightforward to use and implement. OpenAPI v3 is a also just a subset of JSONSchema and that is what is used by `k8s.io/kube-openapi/pkg/validation`.

internal/controller/apiextensions/composite/environment_fetcher.go
MisterMX: I would prefer to pass a pointer and save some performance by avoiding copying to much data around. In the standard usage `env` is never nil so it makes no difference. However, I added an additional check to set `env` to an empty object to prevent potential panics in the future.

pedjak (commented):
it would be really good add unit test and open a doc PR.

MisterMX:
I renamed the field `schema` to `validation` to give a better implication of what this property is about. So the structure now looks like this:

```yaml
spec:
  environment:
    validation:
      jsonSchema:
        type: object
        properties:
          myProp:
            type: string
```

To better explain to the user I added a note to the field description stating that (for now) it is only about validating the data of the merged environment.

I also added unit tests for the implementation.

apis/apiextensions/v1/composition_environment.go
MisterMX: As mentioned, the accepted schema is a JSON schema and not an OpenAPI schema. Although they are very similar, it might be misleading to some people if we change the name.

Also, unlike CRDs and XRDs, this structure does not define an API but only what content in a JSON blob is to be expected. So it makes more sense to me to stick with the current naming.

phisco:
@MisterMX, would you be interested in moving this to function-environment-configs?

Issue #3441, opened 2022-11-07
Report which resources are blocking the transition to Ready state


muvaf:
### What problem are you facing?

I was debugging a composition and got stuck at figuring out why it wouldn't transition to `Ready: True` state even though all resources returned with `kubectl get managed` were ready. Turns out I forgot to add the following to a `ProviderConfig` in my composition:
```
readinessChecks:
- type: None
```

### How could Crossplane help solve your problem?

We can have a status field with `omitempty` that has the list of non-ready/non-existing resources. Once all is good and the list is empty, it'd be omitted.

fcastarede:
`readinessChecks: [{"type": "None"}]` has to be specified for `Provider` also

PR #4565, opened 2023-09-04
apiextensions/composite: list unready resources in condition message

sttts:

Set the `Ready` condition message for reason `Creating` to be the first 3 unready resources. They are sorted in the composition and hence give a stable message.

Bildschirmfoto 2023-09-04 um 18 49 12

Fixes https://github.com/crossplane/crossplane/issues/3441.

I have:

### DIFF
--- a/internal/controller/apiextensions/composite/reconciler.go
+++ b/internal/controller/apiextensions/composite/reconciler.go
@@ -535,6 +535,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 
 	// Note that this 'Composition' will be derived from a
 	// CompositionRevision if the relevant feature flag is enabled.
+	origRev := xr.GetCompositionRevisionReference()
 	rev, err := r.revision.Fetch(ctx, xr)
 	if err != nil {
 		log.Debug(errFetchComp, "error", err)
@@ -542,6 +543,9 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		r.record.Event(xr, event.Warning(reasonCompose, err))
 		xr.SetConditions(xpv1.ReconcileError(err))
 		return reconcile.Result{Requeue: true}, errors.Wrap(r.client.Status().Update(ctx, xr), errUpdateStatus)
+	}
+	if rev := xr.GetCompositionRevisionReference(); rev != nil && (origRev == nil || *rev != *origRev) {
+		r.record.Event(xr, event.Normal(reasonResolve, "Selected composition revision: %s", rev.Name))
 	}
 
 	// TODO(negz): Update this to validate the revision? In practice that's what
--- a/internal/controller/apiextensions/composite/reconciler.go
+++ b/internal/controller/apiextensions/composite/reconciler.go
@@ -531,8 +531,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		r.record.Event(xr, event.Normal(reasonResolve, fmt.Sprintf("Successfully selected composition: %s", compRef.Name)))
 	}
 
-	// Note that this 'Composition' will be derived from a
-	// CompositionRevision if the relevant feature flag is enabled.
+	// Select (if there is a new one) and fetch the composition revision.
 	origRev := xr.GetCompositionRevisionReference()
 	rev, err := r.revision.Fetch(ctx, xr)
 	if err != nil {
--- a/internal/controller/apiextensions/composite/reconciler.go
+++ b/internal/controller/apiextensions/composite/reconciler.go
@@ -20,7 +20,9 @@ package composite
 import (
 	"context"
 	"fmt"
+	"sort"
 	"strconv"
+	"strings"
 	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -623,7 +625,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		r.record.Event(xr, event.Normal(reasonCompose, "Successfully composed resources"))
 	}
 
-	ready := 0
+	var unready []ComposedResource
 	for i, cd := range res.Composed {
 		// Specifying a name for P&T templates is optional but encouraged.
 		// If there was no name, fall back to using the index.
@@ -634,21 +636,27 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 
 		if !cd.Ready {
 			log.Debug("Composed resource is not yet ready", "id", id)
+			unready = append(unready, cd)
 			r.record.Event(xr, event.Normal(reasonCompose, fmt.Sprintf("Composed resource %q is not yet ready", id)))
 			continue
 		}
-
-		ready++
 	}
 
 	xr.SetConditions(xpv1.ReconcileSuccess())
 
 	// TODO(muvaf): If a resource becomes Unavailable at some point, should we
 	// still report it as Creating?
-	if ready != len(res.Composed) {
+	if len(unready) > 0 {
 		// We want to requeue to wait for our composed resources to
 		// become ready, since we can't watch them.
-		xr.SetConditions(xpv1.Creating())
+		names := make([]string, len(unready))
+		for i, cd := range unready {
+			names[i] = cd.ResourceName
+		}
+		// sort for stable condition messages. With functions, we don't have a
+		// stable order otherwise.
+		sort.Strings(names)
+		xr.SetConditions(xpv1.Creating().WithMessage(fmt.Sprintf("Unready resources: %s", firstNAndSomeMore(names))))
 		return reconcile.Result{Requeue: true}, errors.Wrap(r.client.Status().Update(ctx, xr), errUpdateStatus)
 	}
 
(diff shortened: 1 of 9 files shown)


internal/controller/apiextensions/composite/reconciler.go
```diff
@@ -644,3 +648,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 	xr.SetConditions(xpv1.Available())
 	return reconcile.Result{RequeueAfter: r.pollInterval}, errors.Wrap(r.client.Status().Update(ctx, xr), errUpdateStatus)
 }
+
+func firstNAndSomeMore(names []string) string {
```
phisco, line 652: would it make sense to sort names first? just to avoid flapping, which should not happen here atm, but could happen with future changes.

sttts: they are sorted by the order in the compisition.

sttts: We could order them alphabetically. That's merely cosmetics. No strong opinion.

phisco: yes, I meant alphabetically, and yes, that's why I was saying it wouldn't change much here at the moment, but it could change in the future if we changed the logic for any reason. However, no strong opinion either.

bobh66: Personally I would prefer the order from the Composition.  I tend to write my Compositions with the resources in the order that I would expect them to be created, which makes internal dependencies easier to follow.  So having the not-ready resources listed in the order they show up in the Composition would also serve to indicate "how far along" the Composite is in the creation process, and give an indication (to the author) of what might be hanging up.

Granted that relies on the resources being specified in a certain order, but I don't think there is any inherent value in alphabetizing them that would make the debugging process any easier.

Either way having the information is a huge help.

negz: We won't be able to guarantee a stable order with Composition Functions, so we'll need to defensively sort somehow.

phisco:
looks like this is addressing https://github.com/crossplane/crossplane/issues/3441, am I wrong?

sttts:
> looks like this is addressing https://github.com/crossplane/crossplane/issues/3441, am I wrong?

Yes. Added to the description.

internal/controller/apiextensions/composite/reconciler.go
sttts: Added the sorting.

phisco (approved):

internal/controller/apiextensions/composite/reconciler_test.go
```diff
@@ -517,7 +517,7 @@ func TestReconcile(t *testing.T) {
 						MockGet: test.NewMockGetFn(nil),
 						MockStatusUpdate: WantComposite(t, NewComposite(func(cr resource.Composite) {
 							cr.SetCompositionReference(&corev1.ObjectReference{})
-							cr.SetConditions(xpv1.ReconcileSuccess(), xpv1.Creating())
+							cr.SetConditions(xpv1.ReconcileSuccess(), xpv1.Creating().WithMessage("Unready resources: elephant"))
```
pedjak, line 520: we also need additional tests that:
* validate that names are really sorted (here we have just single resource) 
* we display a proper message when the number of unready resources are greater than 3

pedjak (changes requested):

internal/controller/apiextensions/composite/reconciler_test.go
sttts: done. Also compare https://github.com/crossplane/crossplane-runtime/pull/538 for a follow-up to replace this helper everywhere (we have like 5 or more of those now).

internal/controller/apiextensions/composite/reconciler.go
```diff
@@ -658,3 +670,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
```
pedjak, line 674: given that https://github.com/crossplane/crossplane-runtime/pull/538 got merged, would you like to use the new helper function?

sttts: done

internal/controller/apiextensions/claim/reconciler.go
```diff
@@ -546,13 +547,18 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 
 	cm.SetConditions(xpv1.ReconcileSuccess())
 
-	if !resource.IsConditionTrue(cp.GetCondition(xpv1.TypeReady)) {
-		log.Debug("Composite resource is not yet ready")
-		record.Event(cm, event.Normal(reasonBind, "Composite resource is not yet ready"))
+	if cpReady := cp.GetCondition(xpv1.TypeReady); !resource.IsConditionTrue(cpReady) {
+		log.Debug("Composite resource is not yet ready", "reason", cpReady.Reason, "message", cpReady.Message)
```
pedjak, line 551: would it be helpful to add to the log the composite name as well?

internal/controller/apiextensions/claim/reconciler.go
```diff
@@ -546,13 +547,18 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+		if msg := cpReady.Message; msg != "" {
+			record.Event(cm, event.Normal(reasonBind, fmt.Sprintf("%s: %s", Waiting().Message, msg)))
```
pedjak, line 553: same question as above

pedjak (approved):
LGTM, - I would add also composite name to the log/events, but not blocking on it

internal/controller/apiextensions/claim/reconciler.go
sttts: Could be. Not caring about the debug logs though. They are useless anyway. If we cared, we would think about info level ones.

sttts: Note that the claim should be part of the log. The claim includes the composition reference.

internal/controller/apiextensions/composite/reconciler.go
```diff
@@ -634,21 +638,26 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
 	// TODO(muvaf): If a resource becomes Unavailable at some point, should we
 	// still report it as Creating?
-	if ready != len(res.Composed) {
+	if len(unready) > 0 {
 		// We want to requeue to wait for our composed resources to
 		// become ready, since we can't watch them.
-		xr.SetConditions(xpv1.Creating())
+		names := make([]string, len(unready))
+		for i, cd := range unready {
+			names[i] = cd.ResourceName
```
jbw976, line 656: is `ResourceName` guaranteed to be a non empty string? https://github.com/crossplane/crossplane/blob/master/internal/controller/apiextensions/composite/reconciler.go#L628-L633 is making think that it doesn't have that guarantee - would we want to do something different with the name here, e.g. fall back to its index number? 🤔

jbw976 (commented):
Your continued efforts to improve the dev experience are really appreciated @sttts!! Just one question about how we handle resources that potentially don't have a name 🤔

internal/controller/apiextensions/composite/reconciler.go
sttts: we should print the group kind and name. I was there before actually, then thought that the names are what ppl refer to. Now you are (rightfully) saying they are optional 🙂

sttts: actually, kind and name are not guaranteed either. Especially, with the Kubernetes provider these are all objects. The index exists in p&f, not in functions afaik. Looks like we need some heuristic decided by the compositor. What matters here is that we find something pragmatic rather than perfect. Would rather do something imperfect than postponing this topic another 2 years. So +1 putting there the best we have to identify the object. This might also encourage ppl to put proper resource names there to avoid "resource 7" like messages.

sttts: Double checked: I think we are good. For p&f, anonymous resources get `%d` as name. I have made that a little prettier via `resource %d` now. In functions, anonymous resources are not allowed:
> The PTFComposer does not support anonymous templates

@jbw976 ptal

internal/controller/apiextensions/composite/reconciler.go
```diff
@@ -533,6 +533,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 
 	// Note that this 'Composition' will be derived from a
 	// CompositionRevision if the relevant feature flag is enabled.
```
negz, line 535: This comment is stale - revisions are GA and can't be disabled. Not related to this PR but it would be nice to update if doing so is easy.

negz (approved):
Thanks @sttts.

internal/controller/apiextensions/composite/reconciler.go
sttts: done

internal/controller/apiextensions/composite/reconciler.go
jbw976: sounds good @sttts, thanks for looking deeper into that! ✅

phisco (approved):

PR #4570, opened 2023-09-05
apiextensions/composite: tame "Successfully selected composition" event

sttts:

1. Only emit `Successfully selected composition` event on a change.
2. Add what matters: what was selected: `Successfully selected composition: <name>`.

I have:

### DIFF
--- a/internal/controller/apiextensions/composite/reconciler.go
+++ b/internal/controller/apiextensions/composite/reconciler.go
@@ -512,6 +512,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		return reconcile.Result{Requeue: true}, errors.Wrap(r.client.Status().Update(ctx, xr), errUpdateStatus)
 	}
 
+	orig := xr.GetCompositionReference()
 	if err := r.composite.SelectComposition(ctx, xr); err != nil {
 		log.Debug(errSelectComp, "error", err)
 		err = errors.Wrap(err, errSelectComp)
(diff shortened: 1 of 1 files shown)


internal/controller/apiextensions/composite/reconciler.go
```diff
@@ -512,14 +512,17 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+	orig := xr.GetCompositionReference()
 	if err := r.composite.SelectComposition(ctx, xr); err != nil {
 		log.Debug(errSelectComp, "error", err)
 		err = errors.Wrap(err, errSelectComp)
 		r.record.Event(xr, event.Warning(reasonResolve, err))
 		xr.SetConditions(xpv1.ReconcileError(err))
 		return reconcile.Result{Requeue: true}, errors.Wrap(r.client.Status().Update(ctx, xr), errUpdateStatus)
 	}
-	r.record.Event(xr, event.Normal(reasonResolve, "Successfully selected composition"))
+	if compRef := xr.GetCompositionReference(); (orig == nil) != (compRef == nil) || (orig != nil && *compRef != *orig) {
```
turkenh, line 523: Wouldn't this crash if `compRef` is `nil` but the `orig` is not? (which may not be possible depending on the implementation of SelectComposition though.

How about the following?

```suggestion
	if compRef := xr.GetCompositionReference(); compRef != nil && (orig == nil || *compRef != *orig) {
```

sttts: It won't crash: if orig is also nil, `(orig == nil) != (compRef == nil)` is false, and the second only evaluated `orig != nil`. If orig is not nil, the first clause is true.

turkenh: What am I missing? https://go.dev/play/p/A-qNj9pnEs6

sttts: Not the if clause is crashing, but the command afterwards.

sttts: We don't need the case of unsetting of the reference. Have switched to your variant. It's easier. Thanks!

turkenh (approved):
Thanks!

phisco (approved):

PR #4582, opened 2023-09-06
feat[compositions]: realtime compositor – part 1: changes to compositions

sttts:

This PR makes the compositor (= the controller for the XR) to watch `CompositionRevisions` too, and run the compositor on `CompositionRevision` creation for the relevant XRs, in order to select the new revision in automatic mode, and to reapply the new manifests.

This speeds up the inner loop for composition authors to be realtime.

Demo (left changing the composition, top right the composition revision being created, middle right the revision is selected for the XR, bottom right the effect on the MR):

I have:


sttts:
@jbw976 @negz I would like to see some initial review from you. Anything blocking?

Clearly, some (unit) tests are necessary.

negz:
@sttts +1 for adding this functionality. Good idea. As to the implementation, it mostly looks great. My only thought is whether it makes sense for `NewReconciler` to return the `EventHandler`. It seems like it's plumbing up types that are passed into `NewReconciler`. Could you export `EnqueueForCompositionRevisionFunc` and call that from `definition/reconciler.go`?

sttts:
> My only thought is whether it makes sense for NewReconciler to return the EventHandler. It seems like it's plumbing up types that are passed into NewReconciler. Could you export EnqueueForCompositionRevisionFunc and call that from definition/reconciler.go?

Done. You were right, looks better now.

internal/controller/apiextensions/composite/reconciler.go
```diff
@@ -659,3 +664,51 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 	xr.SetConditions(xpv1.Available())
 	return reconcile.Result{RequeueAfter: r.pollInterval}, errors.Wrap(r.client.Status().Update(ctx, xr), errUpdateStatus)
 }
+
+// EnqueueForCompositionRevisionFunc returns a function that enqueues (the
+// related) XRs when a new CompositionRevision is created. This speeds up
+// reconciliation of XRs on changes to the Composition by not having to wait for
+// the 60s sync period, but be instant.
+func EnqueueForCompositionRevisionFunc(of resource.CompositeKind, list func(ctx context.Context, list client.ObjectList, opts ...client.ListOption) error, log logging.Logger) func(ctx context.Context, createEvent runtimeevent.CreateEvent, q workqueue.RateLimitingInterface) {
```
pedjak, line 672: parameter `of` is used only to obtain `schema.GroupVersionKind` of it - so perhaps we can pass that as the function argument instead, `ofGvk schema.GroupVersionKind`?

internal/controller/apiextensions/composite/reconciler_test.go
```diff
@@ -863,3 +869,190 @@ func TestFilterToXRPatches(t *testing.T) {
...
+		name string
+		args args
+		want want
+	}{
+		{
+			name: "empty",
+			args: args{
+				of: resource.CompositeKind{Group: "example.com", Version: "v1", Kind: "Dog"},
+				list: func(ctx context.Context, list client.ObjectList, opts ...client.ListOption) error {
+					return nil
```
pedjak, line 892: we should assert that `list` arg is of proper GVK. and that `opts` are `nil`.

sttts: done

internal/controller/apiextensions/composite/reconciler_test.go
```diff
@@ -863,3 +869,204 @@ func TestFilterToXRPatches(t *testing.T) {
...
+			name: "empty",
+			args: args{
+				of: dog,
+				list: func(ctx context.Context, list client.ObjectList, opts ...client.ListOption) error {
+					// test parameters only here, not in the later tests for brevity.
+					u, ok := list.(*kunstructured.UnstructuredList)
+					if !ok {
+						t.Errorf("list was not an UnstructuredList")
+					} else if u.GroupVersionKind() != dogList {
+						t.Errorf("list was not an UnstructuredList")
```
pedjak, line 901: the error here is not about wron GVK.

sttts: fixed in next push.

sttts:
`Usage` flake

internal/controller/apiextensions/composite/reconciler_test.go
sttts: done

sttts:
Have added an e2e test.

sttts:
@pedjak @negz ptal

test/e2e/apiextensions_test.go
```diff
@@ -105,12 +103,43 @@ func TestCompositionPatchAndTransform(t *testing.T) {
...
+				funcs.ResourcesCreatedWithin(30*time.Second, manifests, "claim.yaml"),
+			)).
+			Assess("ClaimIsReady",
+				funcs.ResourcesHaveConditionWithin(5*time.Minute, manifests, "claim.yaml", xpv1.Available()),
+			).
+			Assess("UpdateComposition", funcs.AllOf(
+				funcs.ApplyResources(FieldManager, manifests, "composition-update.yaml"),
+			)).
+			Assess("ClaimHasPatchedField",
+				funcs.ResourcesHaveFieldValueWithin(10*time.Second, manifests, "claim.yaml", "status.coolerField", "I'M COOL!"),
```
pedjak, line 132: to be on the safe side, perhaps we should increase the time out to something that is slightly lower than the current composite controller polling interval (1minute)? In that way, the tests are gonna become more robust, i.e. they are not going to fail if CI is slow for some other reason.

pedjak (approved):
lgtm, a few non-blocking questions only.

internal/controller/apiextensions/composite/reconciler.go
```diff
@@ -659,3 +664,51 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+			// should not happen
+			return
+		}
+
+		// get all XRs
+		xrs := kunstructured.UnstructuredList{}
+		xrs.SetGroupVersionKind(schema.GroupVersionKind(of))
+		xrs.SetKind(schema.GroupVersionKind(of).Kind + "List")
+		if err := list(ctx, &xrs); err != nil {
+			log.Debug("cannot list in CompositionRevision handler", "type", schema.GroupVersionKind(of).String(), "error", err)
```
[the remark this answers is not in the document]
sttts: that's a more general question (out of scope here): should we log serious problems above debug if they most likely show a logic problem in code, but the end user cannot do anything with them. IMO we should. Got told elsewhere that debug is the right level. I disagree.

internal/controller/apiextensions/composite/reconciler_test.go
```diff
@@ -863,3 +869,204 @@ func TestFilterToXRPatches(t *testing.T) {
...
+								v1.LabelCompositionName: "dachshund",
+							},
+						},
+					},
+				},
+			},
+			want: want{},
+		},
+		{
+			name: "multiple",
```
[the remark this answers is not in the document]
sttts: it does, but I prefer to have simple tests too as it is easier to track down a problem.

test/e2e/apiextensions_test.go
sttts: 10s was intentional. In practice, even under load, we will be <<1s.

internal/controller/apiextensions/composite/reconciler.go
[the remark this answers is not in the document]
pedjak: Since this is a new piece of code, I think we could go with Info here.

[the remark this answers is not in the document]
sttts: done

internal/controller/apiextensions/composite/reconciler.go
```diff
@@ -659,3 +664,52 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+			xr := composite.Unstructured{Unstructured: u}
+
+			// only automatic
+			if pol := xr.GetCompositionUpdatePolicy(); pol != nil && *pol == xpv1.UpdateManual {
+				continue
+			}
+
+			// only those that reference the right Composition
+			compRef := meta.NamespacedNameOf(xr.GetCompositionReference())
+			if compRef.Name != compName {
```
phisco, line 705: nit: I don't think it's needed to make it a NamespacedName
```suggestion
			if ref := xr.GetCompositionReference(); ref == nil || ref.Name != compName {
```

phisco (approved):
LGTM! just a minor nit comment

sttts:
@phisco merged your suggestion.

PR #4584, opened 2023-09-06
apiextensions/composite: report environment errors in Ready condition

sttts:

Don't leave the user in the dark guessing where to look (e.g. events) if the environment selection and/or fetch fails. Set a proper condition that reconcile failed.

@negz @phisco it would be good if we understood from an error whether it is a permanent (e.g. a validation problem) or transient (e.g. some client error or some object does not exist, but might show up eventually). Now we requeue unconditionally. Requeue=true means rate-limited requeue which will eventually backoff, but create quite a number of retries before that.

I have:

### DIFF
--- a/test/e2e/funcs/feature.go
+++ b/test/e2e/funcs/feature.go
@@ -191,9 +191,9 @@ func ResourceDeletedWithin(d time.Duration, o k8s.Object) features.Func {
 
 // ResourcesHaveConditionWithin fails a test if the supplied resources do not
 // have (i.e. become) the supplied conditions within the supplied duration.
+// Comparison of conditions is modulo messages.
 func ResourcesHaveConditionWithin(d time.Duration, dir, pattern string, cds ...xpv1.Condition) features.Func {
 	return func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
-
 		rs, err := decoder.DecodeAllFiles(ctx, os.DirFS(dir), pattern)
 		if err != nil {
 			t.Error(err)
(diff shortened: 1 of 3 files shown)


negz:
> @negz @phisco it would be good if we understood from an error whether it is a permanent

Agreed. I swear we have an issue for this already, but I can't find it either here or on c/crossplane-runtime.

test/e2e/funcs/feature.go
```diff
@@ -219,7 +219,11 @@ func ResourcesHaveConditionWithin(d time.Duration, dir, pattern string, cds ...x
 			_ = fieldpath.Pave(u.Object).GetValueInto("status", &s)
 
 			for _, want := range cds {
-				if !s.GetCondition(want.Type).Equal(want) {
+				got := s.GetCondition(want.Type)
+				if want.Message == "" {
+					got.Message = ""
```
turkenh, line 224: Is this intentional? 
I am reading this as "ignore what you get as a message as long as other fields of the conditions match", right?

jbw976 (commented):
This looks like a reasonable enough pattern to extend to these two extra steps in the same workflow, thanks @sttts!

test/e2e/funcs/feature.go
sttts: Yes. Messages are for humans, not tests.

turkenh: But we could still test/validate that we emit the correct messages for humans, right? 

I can understand that it could be unfeasible in some cases, so don't have a strong opinion to keep them.

sttts: Not in e2e. In unit tests yes.

sttts:
> > This looks like a reasonable enough pattern to extend to these two extra steps in the same workflow
> 
> Done.

Actually I mixed up the two open PRs.

Which extra steps do you mean?

sttts:
The claim controller is intentionally left out for now as we have to think deeper there. Maybe it's even an API discussion as compisition authors want more control what to expose and when ("inertia" discussion).

sttts:
Flake with webhook. Repushed.

turkenh (approved):

PR #4590, opened 2023-09-07
Add nil checks for TLS secrets

ezgidemirel:
Even though the secret fields are filled by the controller, and it's not possible to have the values nil with proper installation/upgrades, it's safer to add nil checks to prevent controller crashes like [here](https://github.com/crossplane/crossplane/pull/4485/files#r1317467180).

Fixes # 

I have:


phisco (approved):

internal/controller/pkg/revision/deployment.go
```diff
@@ -78,20 +79,24 @@ func buildProviderDeployment(provider *pkgmetav1.Provider, revision v1.PackageRe
...
 	}
-
-	secCli := &corev1.Secret{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:            *revision.GetTLSClientSecretName(),
-			Namespace:       namespace,
-			OwnerReferences: []metav1.OwnerReference{meta.AsController(meta.TypedReferenceTo(revision, v1.ProviderRevisionGroupVersionKind))},
-		},
+	if revision.GetTLSClientSecretName() != nil {
+		secCli = &corev1.Secret{
```
sttts, line 93: I am seeing these being used where `buildProviderDeployment` is called. So we run into different issue there.

phisco:
@ezgidemirel linter is complaining about complexity, can you refactor `Providerhook.Pre` a bit or add an ignore directive?

internal/controller/pkg/revision/hook.go
```diff
@@ -116,6 +118,9 @@ func (h *ProviderHooks) Pre(ctx context.Context, pkg runtime.Object, pr v1.Packa
 	// NOTE(hasheddan): we avoid fetching pull secrets and controller config as
 	// they aren't needed to delete Deployment, ServiceAccount, and Service.
 	s, d, svc, secSer, secCli := buildProviderDeployment(pkgProvider, pr, nil, h.namespace, []corev1.LocalObjectReference{})
+	if secSer == nil || secCli == nil {
```
phisco, line 121: do we really want to error out here? or would it be sufficient to skip the deletion of the missing secrets?

internal/controller/pkg/revision/hook.go
```diff
@@ -154,6 +159,9 @@ func (h *ProviderHooks) Post(ctx context.Context, pkg runtime.Object, pr v1.Pack
 		return err
 	}
 	s, d, svc, secSer, secCli := buildProviderDeployment(pkgProvider, pr, cc, h.namespace, append(pr.GetPackagePullSecrets(), ps...))
+	if secSer == nil || secCli == nil {
+		return errors.New(errNoTLSSecretNameAssigned)
+	}
```
phisco, line 164: here it makes sense to error out, as it means something went wrong and we would otherwise end up with a broken installation.

internal/controller/pkg/revision/hook.go
phisco: btw if we finish up #4305 we won't have to delete them anymore, right? similar to what we do for functions.

phisco:
As discussed with @ezgidemirel, I'm addressing the final part of #4305, switching providers to actually use the new certificates, so given that this PR fell behind w.r.t. master and could actually have more complex implications than just a few nil checks, I'll just address this as part of that, so this PR can be closed.

PR #4594, opened 2023-09-09
test/e2e: print related objects

sttts:

When a condition cannot be reached, print related objects (those owned by the object in question) in e2e. The object with the wrong condition alone printed in e2e does not help to understand what's going on.

I have:

### DIFF
--- a/test/e2e/funcs/collect.go
+++ b/test/e2e/funcs/collect.go
@@ -21,17 +21,18 @@ import (
 	"fmt"
 	"strings"
 
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/dynamic"
-	"k8s.io/client-go/metadata"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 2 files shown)


lsviben (approved):
Looks very helpful, thanks @sttts !

Some thoughts, I was wondering if we should go deeper, as in getting the owned resources of the owned resources we are collecting. So waiting for a Claim to get READY you would get Claim->Composite->MRs.

Other thing is about the verbositiy. I wonder does it make sense to print out related resources which are Healthy, as they are not the culprints (probably) of the resource not getting in a Condition we desire? Dont feel strong about this, just wondering aloud.

sttts:
> Some thoughts, I was wondering if we should go deeper, as in getting the owned resources of the owned resources we are collecting. So waiting for a Claim to get READY you would get Claim->Composite->MRs.

The PR does that. It is recursive.

> Other thing is about the verbositiy. I wonder does it make sense to print out related resources which are Healthy, as they are not the culprints (probably) of the resource not getting in a Condition we desire? Dont feel strong about this, just wondering aloud.

I would rather not optimize on that side, but in a follow-up find a way to produce these yamls as Github CI artifact, and give an identifying path into some zip. Then we can store even more, e.g. events.

lsviben:
> > Some thoughts, I was wondering if we should go deeper, as in getting the owned resources of the owned resources we are collecting. So waiting for a Claim to get READY you would get Claim->Composite->MRs.
> 
> The PR does that. It is recursive.
I seem to have missed that, awesome!

turkenh (approved):

PR #4610, opened 2023-09-12
package-manager: always set Unhealthy() message

sttts:

Add condition messages to the package manager resources.

I have:

### DIFF
--- a/internal/controller/pkg/revision/reconciler.go
+++ b/internal/controller/pkg/revision/reconciler.go
@@ -496,8 +496,13 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 	// an error.
 	if rc == nil && pullPolicyNever {
 		log.Debug(errPullPolicyNever)
+
 		err := errors.New(errPullPolicyNever)
+		pr.SetConditions(v1.Unhealthy().WithMessage(err.Error()))
+		_ = r.client.Status().Update(ctx, pr)
+
 		r.record.Event(pr, event.Warning(reasonParse, err))
+
 		return reconcile.Result{}, err
 	}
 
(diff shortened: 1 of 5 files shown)


test/e2e/funcs/feature.go
```diff
@@ -219,7 +219,11 @@ func ResourcesHaveConditionWithin(d time.Duration, dir, pattern string, cds ...x
 			_ = fieldpath.Pave(u.Object).GetValueInto("status", &s)
 
 			for _, want := range cds {
-				if !s.GetCondition(want.Type).Equal(want) {
+				got := s.GetCondition(want.Type)
```
turkenh, line 222: Same as https://github.com/crossplane/crossplane/pull/4584#discussion_r1325368954

internal/controller/pkg/revision/reconciler.go
```diff
@@ -549,40 +556,45 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
 	// Lint package using package-specific linter.
 	if err := r.linter.Lint(pkg); err != nil {
-		pr.SetConditions(v1.Unhealthy())
+		log.Debug(errLintPackage, "error", err)
+
+		err = errors.Wrap(err, errLintPackage)
+		pr.SetConditions(v1.Unhealthy().WithMessage(err.Error()))
 		_ = r.client.Status().Update(ctx, pr)
 
+		r.record.Event(pr, event.Warning(reasonLint, err))
```
turkenh, line 577: Do you think if we have too much of similar lines that could be refactored into a dedicated function?

turkenh (approved):
Thanks @sttts 🙌

sttts:
> I would suggest dropping the change https://github.com/crossplane/crossplane/pull/4610#discussion_r1325377039 before merging this PR, though.

Don't think it makes sense to compare messages in e2e. They might differ by some details, and adding a predicate to ignore the message is overkill. Am happy to make that clearer in the godocs.

sttts:
Actually, I suspect some message comparison somewhere in the code as this PR consistently fails e2e.

test/e2e/funcs/feature.go
sttts: Updated. Ptal. Now it's unconditional, documented and the dev gets an error if a message is set in the comparison conditions.

internal/controller/pkg/revision/reconciler.go
sttts: They are not really equal but differ in details. Also this is preexisting in the code. I have only moved a few lines consistently to make sense.

test/e2e/funcs/feature.go
turkenh: This might break some existing e2e tests, did you check that?

sttts: Actually, you are right. It could break something :) Checking.

sttts: Double checked. We should be fine.

lsviben:
Is the last [commit](https://github.com/crossplane/crossplane/pull/4610/commits/ce2c28d1ebb9beaded1a9cb80569f7f74deb6a75) with making the ownership graph in the e2e tests supposed to be here? Not sure if on purpose or a testing leftover, feels out of place, although it seems useful

sttts:
It's kind of intentional, see https://github.com/crossplane/crossplane/pull/4594. I needed it to understand flakes here.

lsviben:
> It's kind of intentional, see #4594. I needed it to understand flakes here.

I see, looks super useful. Ill review that one after this to push it along

lsviben (approved):
LGTM, thanks @sttts !

PR #4611, opened 2023-09-13
feat(helm-chart): add option for installCRDs and extraObjects

### LINKED ISSUE #4551: Opt-in to install CRDs with Helm value

The poor handling of CRDs in Helm is well-known, but new tools and processes are able to work around the eschewed CRD support in Helm. FluxCD is one example, and I assume ArgoCD also has some support.

Crossplane seems to use the runtime/controller to install (even the core) Crossplane CRD. This means that even if you follow the detailed uninstall process described in [the docs](https://docs.crossplane.io/latest/software/uninstall/), the CRDs will remain orphaned in your cluster after the uninstall. This is unfortunate, and no way to avoid AFAIK. This was also causing issues for us when trying to reinstall Crossplane in a cluster we have because the version we came from was a very old version.

Also, the [Helm chart README](https://github.com/crossplane/crossplane/tree/master/cluster/charts/crossplane#readme) might give you the impression that CRDs are installed as part of the chart:

image

I posted a [question on Slack](https://crossplane.slack.com/archives/CEF5N8X08/p1692994523317239), and got some feedback from Crossplane maintainers.

I would suggest introducing a new opt-in value in the Helm chart to install CRDs. To overcome the issues with CRD handling in Helm, I propose to do it in a similar way as cert-manager does: https://cert-manager.io/docs/installation/helm/#helm-installation. The main idea is to provide the CRDs to Helm as standard templated resources, and not as a "Helm CRD".

I think the potential issue here is to verify that the Crossplane runtime fully supports patching a CRD created (and managed) by some GitOps tool. To not break the Helm chart for existing users, the Crossplane runtime should still ensure the CRDs are present and have the desired state.

haarchri:
- add option for crdsInstall
- add option for extraObjects

#### install and remove with  `installCRDs: true`
```
cat values.yaml
installCRDs: true
```

```
helm install crossplane . --create-namespace --namespace crossplane-system --wait

helm list -aA
NAME            NAMESPACE               REVISION        UPDATED                                 STATUS  CHART            APP VERSION
crossplane      crossplane-system       1               2023-09-13 14:09:06.736166 +0200 CEST   deployedcrossplane-0.0.1 1.13.1

 kubectl get crds | wc -l
      15

helm uninstall crossplane -n crossplane-system
release "crossplane" uninstalled

kubectl get crds | wc -l
No resources found
       0     
```

####  install with `installCRDs: true` and `extraObjects`
```
cat values.yaml
installCRDs: true

extraObjects:
    kind: ControllerConfig
    metadata:
      name: aws-config
      annotations:
        eks.amazonaws.com/role-arn: arn:aws:iam::123456789101:role/example
        helm.sh/hook: post-install
    spec:
      podSecurityContext:
        fsGroup: 2000
  - apiVersion: pkg.crossplane.io/v1
    kind: Provider
    metadata:
      name: provider-aws
      annotations:
        helm.sh/hook: post-install
    spec:
      package: xpkg.upbound.io/crossplane-contrib/provider-aws:v0.33.0
      controllerConfigRef:
        name: aws-config

helm install crossplane . --create-namespace --namespace crossplane-system --wait

kubectl get controllerconfig
NAME         AGE
aws-config   8s

kubectl get providers
NAME           INSTALLED   HEALTHY   PACKAGE                                                   AGE
provider-aws                         xpkg.upbound.io/crossplane-contrib/provider-aws:v0.33.0   10s

```

Fixes #4551

I have:

- [ ] Added `backport release-x.y` labels to auto-backport this PR, if necessary.


phisco:
@haarchri, is there a strong reason to have both fields introduced in the same PR? I think it would be better to discuss the two additions separately.

pedjak:
@haarchri what is the user story for this PR? 

* If `installCRDs` is not set, should we deploy CRDs using the existing init container? If yes, then `installCRDs` variable is unclear, and we should come up with some better name to reflect the intention
* If `installCRDs` is set to `true`, should we disable CRDs installation via the init container.? In this PR, CRDs are packed within the chart, but also embedded into the crossplane image and deployed at the startup time. If we skip the later, would be able to catch and fix issues similar to https://github.com/crossplane/crossplane/pull/4408 ?
* Furthermore, `crossplane core init` command deploys webhooks/rbac rules as well. After uninstalling crossplane, I would expect that these resources are removed as well

IMHO, if we add `installCRDs`  option, when we set to `false`, we should not install it via `crossplane core init` - we should provide to users a yaml file containing all CRDs, so that it can install it separately, before the chart. However, with that we are loosing the hook into CRD installation process and do some tweaks as needed.

pedjak:
> The CRDs remain consistent between the initialization of Crossplane Core and the Helm chart because we ensure that the correct CRDs are copied into the Helm chart. This has been verified locally, and when the optional installCRDs flag is used, Crossplane is fully removed.

It would be interesting to see how this behave during upgrades, i.e. what happens if upgrading CRD fails, or we drop say alpha version of API in the update.

BTW, just tried the chart from this PR and crossplane is not fully removed:

```sh
$ helm install crossplane cluster/charts/crossplane --create-namespace --namespace crossplane-system --set image.tag=v1.14.0-rc.0.278.g6d25fd66.dirty --set installCRDs=true --wait
NAME: crossplane
LAST DEPLOYED: Fri Sep 15 15:16:47 2023
NAMESPACE: crossplane-system
STATUS: deployed
REVISION: 1
TEST SUITE: None
NOTES:
Release: crossplane

Chart Name: crossplane
Chart Description: Crossplane is an open source Kubernetes add-on that enables platform teams to assemble infrastructure from multiple vendors, and expose higher level self-service APIs for application teams to consume.
Chart Version: 0.0.1
Chart Application Version: 0.0.1

Kube Version: v1.26.3
$ k get crds
NAME                                                       CREATED AT
compositeresourcedefinitions.apiextensions.crossplane.io   2023-09-15T13:16:47Z
compositionrevisions.apiextensions.crossplane.io           2023-09-15T13:16:47Z
compositions.apiextensions.crossplane.io                   2023-09-15T13:16:47Z
configurationrevisions.pkg.crossplane.io                   2023-09-15T13:16:47Z
configurations.pkg.crossplane.io                           2023-09-15T13:16:47Z
controllerconfigs.pkg.crossplane.io                        2023-09-15T13:16:47Z
environmentconfigs.apiextensions.crossplane.io             2023-09-15T13:16:47Z
functionrevisions.pkg.crossplane.io                        2023-09-15T13:16:47Z
functions.pkg.crossplane.io                                2023-09-15T13:16:47Z
locks.pkg.crossplane.io                                    2023-09-15T13:16:47Z
providerrevisions.pkg.crossplane.io                        2023-09-15T13:16:47Z
providers.pkg.crossplane.io                                2023-09-15T13:16:47Z
storeconfigs.secrets.crossplane.io                         2023-09-15T13:16:47Z
usages.apiextensions.crossplane.io                         2023-09-15T13:16:47Z
$k get validatingwebhookconfiguration
NAME                   WEBHOOKS   AGE
crossplane             2          67s
crossplane-no-usages   1          67s
$ helm uninstall crossplane --namespace crossplane-system
release "crossplane" uninstalled
$ k get validatingwebhookconfiguration
NAME                   WEBHOOKS   AGE
crossplane             2          117s
crossplane-no-usages   1          117s
$ k get crds
No resources found
```

the webhooks are still there.

Furthermore, let's create some claims, before trying to unstall crossplane:

```sh
helm install crossplane cluster/charts/crossplane --create-namespace --namespace crossplane-system --set image.tag=v1.14.0-rc.0.278.g6d25fd66.dirty --set installCRDs=true --wait
NAME: crossplane
LAST DEPLOYED: Fri Sep 15 15:30:10 2023
NAMESPACE: crossplane-system
STATUS: deployed
REVISION: 1
TEST SUITE: None
NOTES:
Release: crossplane

Chart Name: crossplane
Chart Description: Crossplane is an open source Kubernetes add-on that enables platform teams to assemble infrastructure from multiple vendors, and expose higher level self-service APIs for application teams to consume.
Chart Version: 0.0.1
Chart Application Version: 0.0.1

Kube Version: v1.26.3
$ k apply -f test/e2e/manifests/apiextensions/composition/minimal/setup
composition.apiextensions.crossplane.io/xnopresources.nop.example.org created
compositeresourcedefinition.apiextensions.crossplane.io/xnopresources.nop.example.org created
provider.pkg.crossplane.io/provider-nop created
$ k get provider
NAME           INSTALLED   HEALTHY   PACKAGE                                                  AGE
provider-nop   True        True      xpkg.upbound.io/crossplane-contrib/provider-nop:v0.2.0   26s
$ k apply -f test/e2e/manifests/apiextensions/composition/minimal/claim.yaml 
nopresource.nop.example.org/apiextensions-composition-minimal created
$ k get -A nopresource.nop.example.org
NAMESPACE   NAME                                SYNCED   READY   CONNECTION-SECRET   AGE
default     apiextensions-composition-minimal   True     True                        32s
 $ helm uninstall crossplane --namespace crossplane-system
release "crossplane" uninstalled
$ k get -A nopresource.nop.example.org 
NAMESPACE   NAME                                SYNCED   READY   CONNECTION-SECRET   AGE
default     apiextensions-composition-minimal   True     True                        86s
 $ k get crds
NAME                                                       CREATED AT
compositeresourcedefinitions.apiextensions.crossplane.io   2023-09-15T13:30:11Z
locks.pkg.crossplane.io                                    2023-09-15T13:30:11Z
nopresources.nop.crossplane.io                             2023-09-15T13:31:33Z
nopresources.nop.example.org                               2023-09-15T13:31:30Z
providerrevisions.pkg.crossplane.io                        2023-09-15T13:30:11Z
xnopresources.nop.example.org                              2023-09-15T13:31:31Z
```

Claim `apiextensions-composition-minimal` is orphaned now, because its composition got removed and if we wanna remove the claim we should manually remove finalizer before.

haarchri:
If you configure your claims using GitOps systems such as Flux or Argo CD with a "depends-on" configuration, it will first remove all resources, including Crossplane resources. The advantage is that Crossplane Custom Resource Definitions (CRDs) will also be removed in this process at last. The init-container will handle the application of CRDs and manage the migration process. In what scenario do you believe Helm might struggle to handle the CRD application?

turkenh:
I haven't look in depth yet, but I would suggest being cautious about this kind of change (speaking about CRDs). 

I remember a lot of pain in the early days while trying to manage CRDs with helm (both managing under `crds` and `templates` directory had different issues) which might have been solved/improved already. This is [the PR](https://github.com/crossplane/crossplane/pull/2160) introducing managing CRDs in init container could help to load up some context.

So, my suggestion would be:
- Make sure we tested different scenarios (install, upgrade, downgrade, uninstall) extensively. 
- Manage CRDs from a single place. (i.e. if we decide to manage with helm, remove them from the init container as @pedjak suggested.)

phisco:
@haarchri, my [comment](https://github.com/crossplane/crossplane/pull/4611#issuecomment-1717530958) about splitting this in two PRs is still valid, could you split this? 😅

haarchri:
I can split this PR shortly

pedjak:
Related: https://github.com/crossplane/crossplane/issues/4649#issuecomment-1730760525

Helm charts should also take care about webhooks.

haarchri:
@phisco  remove extraObjects in this PR

haarchri:
@turkenh  @pedjak today we not generating all webhooks in repo before, we have an tracking issue for this comment ?

https://github.com/crossplane/crossplane/blob/master/cluster/webhookconfigurations/usage.yaml#L2-L4

pedjak:
> today we not generating all webhooks in repo before, we have an tracking issue for this comment ?

Yes, most of them are generated, but this one not, because our project structure drifted over the time from kubebuilder recommended one.

Perhaps we should remove the drift, i.e. move back to the kubebuilder project structure @turkenh ?

Issue #4583, opened 2023-09-06
Add Data from Arbitrary K8s Object to Composition Environment


pedjak:
Followup of https://github.com/crossplane/crossplane/pull/4366

### What problem are you facing?

There are cases when field `fa` in managed resource `MRA` belonging to composition `CA` needs to be set to the value of field `fb` in managed resource `MRB` that does not belong to the same composition. At the same time, the schema of `MRA` does not have the reference field for `fb` field, or the reference field exists but allows referring only single source.

For example,  [`datasync.aws.upbound.io` `Task`](https://github.com/upbound/provider-aws/blob/main/apis/datasync/v1beta1/zz_task_types.go#L375) has [`spec.forProvider.sourceLocationArn`](https://github.com/upbound/provider-aws/blob/main/apis/datasync/v1beta1/zz_task_types.go#L359) field that can obtain value from an instance of the following object types:

* LocationS3
* LocationSmb
* LocationObjectStorage
* LocationNfs
* LocationHdfs
* LocationFsxWindowsFileSystem
* LocationFsxOpenzfsFileSytem
* LocationFsxLustreFileSystem
* LocationEfs

However, [the implemented provider resolving mechanism](https://github.com/upbound/provider-aws/blob/main/apis/datasync/v1beta1/zz_generated.resolvers.go#L104-L112) looks only for the value in `LocationS3` instance named by `spec.forProvider.sourceLocationArnRef` field.

At the moment, there are two possible workarounds:

#### Declare `MRB` within the composition `CA` as observe only and use the available patch strategies to set the value of `fa` within `MRA`.

Drawbacks:
* For each claim, additional observe only resource gets created, just to be able  to work around the limitations of the current reference mechanism on certain  fields. In a system with a high number of claims, that would put additional  pressure on k8s API and etcd.
* Composite object needs to expose the patched value, although fully unneeded  from API design perspective

#### Create a number of environment configs

* Create CompositionB declaring `MRB` and `EnvironmentConfig` containing `bar: value` entry

```yaml
apiVersion: apiextensions.crossplane.io/v1
kind: Composition
metadata:
  name: composition-b
spec:
  resources:
    - name: mr-b
      base:
        apiVersion: acme.com/v1beta1
        kind: MRB
        spec:
          forProvider:
            region: "us-east-2"
      patches:
        - type: ToEnvironmentFieldPath
          fromFieldPath: metadata.annotations[crossplane.io/external-name]
          toFieldPath: bar
[...]

PR #4632, opened 2023-09-15
feat(functions): support environment if present

### LINKED ISSUE #4600: Support `EnvironmentConfig` when using Composition Functions

Today it's not possible to use EnvironmentConfigs and Composition Functions together. The environment is not passed to the Function pipeline. This is largely because the two features are both alpha, and figuring out the intersection of two in-flux designs is tough. 😄 With Functions going to beta, we will need to figure out supporting EnvironmentConfigs. In particular I think convincing folks to move to the P&T Function (https://github.com/crossplane/crossplane/issues/4571) will be a hard sell if we don't support them.

I think it needs to be possible to:

* Pass "the environment" to the Function pipeline.
* Allow Functions to mutate the environment (i.e. to affect subsequent Functions in the pipeline).
* Avoid using EnvironmentConfigs feeling similar to P&T (so you can avoid P&T style completely when using Functions, if you want to).

This relates to https://github.com/crossplane/crossplane/issues/4583, which is tracking the future shape of the environment API.

phisco:

Fixes #4600

Even with the whole conversation going on about environment and EnvironmentConfigs, I think this simple change will allow us to accomodate whatever we decide to do with that

Environment being added to both the observed and desired state allows to first pass whatever environment a Composition built before calling the function pipeline, but also using it as a variable space where functions can write stuff to be used by following functions.

I have:


phisco:
not sure whether we want to also apply patches at `spec.environment.patches` here, I guess that would be a P&T leakage into function based compositions, the same thing can be done by a function using the provided environment and composite resource.

phisco:
The other option we could have considered was injecting into the observed state a fake EnvironmnentConfig, but I think such implementation would have the following issues with that:
- we would bind functions to EnvironmnentConfigs, while I think there is value into the separation between EnvironmentConfig and the in-memory environment.
- we would miss on the in my opinion useful "variable-like" capabilities I mentioned above as we don't allow functions to write to the observed state and we discard any change across different function calls.

internal/controller/apiextensions/composite/composition_functions.go
```diff
@@ -471,7 +472,16 @@ func AsState(xr resource.Composite, xc managed.ConnectionDetails, rs ComposedRes
...
 	}
 
-	return &v1beta1.State{Composite: oxr, Resources: ocds}, nil
+	var e *structpb.Struct
+	if env != nil {
+		e, err = AsStruct(env)
+		if err != nil {
+			return nil, errors.Wrap(err, errEnvAsStruct)
+		}
+
```
negz, line 481: Nit: This whitespace looks accidental/superfluous.

apis/apiextensions/fn/proto/v1beta1/run_function.proto
```diff
@@ -90,6 +90,9 @@ message State {
 
   // The state of any composed resources.
   map<string, Resource> resources = 2;
+
+  // The environment passed to the Function pipeline.
```
negz, line 94: Should we add some kind of `THIS IS AN ALPHA FIELD` warning like we do on our CRDs? Something that communicates that this field is more at-risk of changing than the rest of the API.

negz (commented):
Thanks @phisco! I agree this is probably good enough for now.

apis/apiextensions/fn/proto/v1beta1/run_function.proto
negz: As to how it might change - if we did make `environment` an array or map of resources I could imagine reflecting that here. So something more like:

```proto
map<string, google.protobuf.Struct> environment = 3;
```

phisco: Yes, you are right. I will update it 👍

phisco: Added the alpha warning as suggested 👍 

About the type of the environment field, wouldn't a `google.protobuf.Struct` already accomodate such a change? currently the environment is fully unstructured:
```go
// Environment defines unstructured data.
type Environment struct {
	unstructured.Unstructured
}
```
But if we decided to have two separate sections, "data" and "resources" as I'm [suggesting](https://github.com/crossplane/crossplane/issues/4583#issuecomment-1719193018), we would just have to populate the Environment in a different way, but we shouldn't need to change the code here.

negz: I agree we wouldn't _have to_, in that a `google.protobuf.Struct` can pretty much model anything. It does however seem to me that if we know the environment will always have a certain shape (i.e. data and array of resources) it would be a better experience for folks building functions for us to reflect that shape in our 'schema' as much as possible.

phisco:
> Would the "desired environment" work the same way? i.e. Functions just set the keys they want to set? What would we do with the desired environment at the end of the pipeline run? Discard it?

Yes, for consistency I would expect it to be exactly the same as the observed/desired states. But then a downstream function would have to "merge" the observed and desired environment to access any info. The only other option would be to have it at the same level as the states in both the request and response, with its own semantic, more similar to a context, rather than the observed/desired state.

Currently the state is discarded at the end of the reconciliation loop for P&T, so yes, I'd say for the moment we should stick to that.

phisco:
Yeah, I agree it feels odd. I will update the PR tomorrow with a dedicated structure for both request and response, using a map as you suggested. I'd arbitrarily already put it under a "data" top level key, just to accommodate future changes such as the ones we are discussing ("resources"), wdyt?

phisco:
@negz, pushed some changes, let me know wdyt 🙏

apis/apiextensions/fn/proto/v1beta1/run_function.proto
```diff
@@ -47,6 +47,13 @@ message RunFunctionRequest {
   // Optional input specific to this Function invocation. A JSON representation
   // of the 'input' block of the relevant entry in a Composition's pipeline.
   optional google.protobuf.Struct input = 4;
+
+  // The environment to run the Function with.
+  //
+  // THIS IS AN ALPHA FIELD. Do not use it in production. It is not honored
+  // unless the relevant Crossplane feature flag is enabled, and may be
+  // changed or removed without notice.
```
negz, line 55: I know I asked for this warning, but maybe we don't need it?

It sounds like we're both pretty confident that we'll need some kind of object that is passed through the Function pipeline as a way for Functions to pass information (that isn't desired state) to each other. So maybe it's fine to commit to this being a beta part of our API without a scary warning. What's "alpha" is more how we might choose to use it (e.g. loading it up with EnvConfigs etc).

internal/controller/apiextensions/composite/composition_functions.go
```diff
@@ -215,11 +216,21 @@ func (c *FunctionComposer) Compose(ctx context.Context, xr *composite.Unstructur
...
+	var env *v1beta1.Environment
+	if req.Environment != nil {
+		d, err := AsStruct(req.Environment)
+		if err != nil {
+			return CompositionResult{}, errors.Wrap(err, errEnvAsStruct)
+		}
+		env = &v1beta1.Environment{
+			Data: d,
+		}
+	}
```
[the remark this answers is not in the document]
negz: I feel pretty genuinely ambivalent about that, haha.

I do really like the idea of moving the "bootstrapping" of the environment into a Function. This gives us more room to experiment there, and allows folks to opt out of the complexity if they don't need it.

What I'm not so sure about is the general pattern of Functions reading stuff from the API server and passing them to the Function pipeline. (I presume a Function would select and read EnvConfigs). Is that something we should allow? Or do we want Crossplane to be the sole broker of everything read from and written to the API server. I'm not concretely sure of what risk Functions reading/writing to the API server poses, more of an intuition at this point.

If we did allow this we'd probably need to update the spec: https://github.com/crossplane/crossplane/blob/5797021/contributing/specifications/functions.md#runtime-environment

apis/apiextensions/fn/proto/v1beta1/run_function.proto
```diff
@@ -173,4 +187,10 @@ enum Severity {
   // Normal results are emitted as normal events and debug logs associated
   // with the composite resource.
   SEVERITY_NORMAL = 3;
+}
+
+// The environment to be passed down the Function pipeline.
+message Environment {
+  // The data section of the environment.
+  google.protobuf.Struct data = 1;
```
negz, line 195: Will Functions use keys in this `data` struct to pass information to each other?

My understanding from https://github.com/crossplane/crossplane/pull/4731#issuecomment-1748183984 is "no". If that's correct, I think it would be useful to bring that use case into scope for this PR. Would we add a new field to the `Environment` message for that?

negz: I could imagine foregoing the `Environment` message and just having something like:

```proto
message RunFunctionRequest {

  // Fields omitted for brevity

  google.protobuf.Struct environment = 5;
}
```

You could then put the merged `EnvironmentConfigs` under a well-known key like `apiextensions.crossplane.io/config` and let Functions use other fields within the `environment` struct to pass information without worrying about conflicting with what is loaded from the `EnvironmentConfigs`.

apis/apiextensions/fn/proto/v1beta1/run_function.proto
phisco: Yes, I think the in-memory environment is not alpha at all and a pretty useful concept, I agree what's alpha is actually EnvironmentConfig and the way we retrieve that via spec.environment . I'll remove

internal/controller/apiextensions/composite/composition_functions.go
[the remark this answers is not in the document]
phisco: Yes, adding side effects could make everything more complex, e.g. how do we handle that from xrender? However I feel it's going to be hard to prevent, people can always add a service account and use that to do whatever they want, so I guess we should just embrace it and make it a thing we take into account. For example, we could add an annotation to let xrender know a function should be skipped or add conditionals as we were discussing in some other thread, so that one can render compositions skipping the retrieval steps and passing stuff into the environment instead of letting it be retrieved from the outside world.

apis/apiextensions/fn/proto/v1beta1/run_function.proto
phisco: > Will Functions use keys in this data struct to pass information to each other?
What do you mean? "data" is what we get today merging EnvironmentConfig and patching to and from the environment, if we were to add other fields to the environment these would be on the same level as the "data".

`environment.data` is exactly what you were putting at `environment[apiextensions.crossplane.io/environment]`. I would let people use FromEnvironment patches adding the "data" prefix to fields, so that if we ever added something in the future it could still be reachable by FromEnvironmentPatches in the patch and transform function.

apis/apiextensions/fn/proto/v1beta1/run_function.proto
phisco: Removed them

internal/controller/apiextensions/composite/composition_functions.go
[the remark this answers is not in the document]
turkenh: > Yes, adding side effects could make everything more complex, e.g. how do we handle that from xrender? However I feel it's going to be hard to prevent, people can always add a service account and use that to do whatever they want, so I guess we should just embrace it and make it a thing we take into account. 

I am pretty skeptical to the idea of natively supporting direct access to Kube API from a function. This is mostly because, this would open a door invalidating most of the opinionated framework that we're trying to provide with `RunFunctionRequest`/`RunFunctionResponse`. If a function could access to the Kube API native, why should it bother getting resources from Observations or specifying the desired state instead of just getting/patching the resources directly. I am imagining that there will be a tendency to open that door with the first limitation in the framework that somebody hits and never get back to track even we resolve that afterwards. There is also a whole new complexity that would be introduced due to the RBAC. How and who will define the RBAC for such a service account, if it is users, then this will still require manual work and passing a service account wouldn't solve the problem. I don't want to think about Crossplane also taking care of RBAC for such a case 😅 

However, considering the Kube API is just yet another API and we are planning to enable passing credentials to functions, power users could still communicate with the Kube API if/once they really have to. 

This feels quite similar to what one can do with **provider-kubernetes** and **provider-helm**. They _can_ be used to create resources (other than the providers own CRDs and a limited set of other types) on the same Kubernetes cluster if a provider config provided accordingly. However, we didn't enable this as a native functionality for all Crossplane Providers, and working fine so far.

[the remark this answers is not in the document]
turkenh: Let me invalidate a bit what I wrote above 😄 

I just remembered there was a functionality in the Provider Meta definition that no provider uses currently (to the best of my knowledge), which is [PermissionRequests](https://github.com/crossplane/crossplane/blob/master/apis/pkg/meta/v1/provider_types.go#L38-L42). I am not sure if it got broken at some point but if we reflect that functionality to Function meta package, an environment-merger function should be able to give a list RBAC policy rules in its package definition to access EnvironmentConfig resources. And this doesn't feel that bad as it will be something defined at the package level and have a very limited scope.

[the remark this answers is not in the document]
phisco: Yes, having it already implemented makes it feel less bad, isn't it? 😂

I think it makes sense to embrace it, nothing is going to stop a power user from doing that already, a function is a pod (atm), and it gets a serviceAccount, so we should just accept it and help making it a good and controlled experience.

I don't really agree with the argument that if you are talking directly to the API server, you should just do that instead of going through Compositions. That's not really how I see it. The one who is willing to get into the complexity necessary to talk with the api server is the function author, which is probably an expert and wants to expose some simplified interface to Compositions users. If we merge the two personas, Compositions themselves become useless, as people could "just build their own Operator".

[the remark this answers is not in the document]
turkenh: > Yes, having it already implemented makes it feel less bad, isn't it? 😂

Not really 🙂
For me, the key here is RBAC and how/where it is defined.
Initially, I was thinking it as if it would enable/promote doing anything on the Kube API and functions evolving to k8s operators.
However, a permission model where only an explicit/limited set of resources defined by function author is accessible  makes it more controlled as you are stating and not harmful to our opinionated framework

apis/apiextensions/fn/proto/v1beta1/run_function.proto
negz: > I would let people use FromEnvironment patches adding the "data" prefix to fields

We're on the same page here pattern wise, i.e. nesting what we would call "the environment" today under a key in... the environment. I think we need a more descriptive, specific names than "environment" and "data" though. If my EnvironmentConfigs and defaults become the "data" key in the environment, what else would there be? Would the other "keys" be something other than data?

Perhaps if we really want to stick with "environment" as the top-level name `environment.data` could become `environment.config`? or `environment[apiextensions.crossplane.io/config]`?

> If we ever added something in the future it could still be reachable by FromEnvironmentPatches in the patch and transform function.

I feel that v1.14 we should have a story for Functions to write to the "environment" for v1.14. If we merge this PR as-is it seems like their only option will be to write to `environment.data`, which means they could potentially collide with the object that Crossplane loads into that struct from EnvConfigs.

I lean toward the top level of the "environment" being a struct or map because then a Function can "claim" an arbitrary field/key within that top-level struct/map without needing an update to the protobuf schema to add more fields.

phisco: I still don't feel great about having a well-known key we use "officially" to inject the current environment into an unstructured object.

I'd be more comfortable saying that the environment (or context) is just an unstructured object and letting functions define their convention, e.g. an EnvironmentConfig aggregating function defines a key it will put the aggregated result in some key, which will be the same key expected by the patch and transform function, ignoring the whole spec.environment section in Pipeline mode. This was what I was hinting at here: https://github.com/crossplane/crossplane/pull/4632#discussion_r1346177527.

My point is, if stuff is official it should be explicitly represented in the schema, otherwise let's fully go the other way and let functions define their own conventions, just providing them a way to pass data through the pipeline outside the desired state.

negz: Maybe a summary of my thoughts will help me communicate clearly. 😄 

* I get why we want to stick with "environment" as the top-level name. I'm on board with that.
* I'm on board with what we load from EnvConfigs ("data") being a field or key in the top-level environment.
* I think we need to find a more descriptive name than "data", though. [^1]
* I think Functions should be able to create arbitrary fields/keys in the top-level environment.
* I'm on board with `PatchFromEnvironment` fields being "rooted" at the top-level object. So it can access both "data" and whatever Functions write to it.

[^1]: Naming is important to me here because it will help bridge how folks think about "the environment". Today I think it means "stuff loaded from EnvConfigs" (+defaults). Here it sounds like "the environment" becomes one layer higher than that.

negz: > I think we need to find a more descriptive name than "data", though. 

If we do go with a map / arbitrary keys, I think it makes sense to reserve and use certain key namespaces for Crossplane (core, not Functions) to write to. Similar to how `kubernetes.io/` and `k8s.io/` annotation key prefixes are [reserved for use by Kubernetes components](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set), and `GITHUB_` env vars are [reserved for use by GitHub in GitHub Workflows](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable).

internal/controller/apiextensions/composite/composition_functions.go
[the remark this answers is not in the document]
bobh66: If I understand this correctly it means that while Functions can be used to modify the current environment/context, and could populate the environment from non-K8S API sources, they should not populate it from K8S API resources and that functionality needs to stay where it is (for now) in the Composition, and the discussion in #4583 should stay out of the realm of Functions.

apis/apiextensions/fn/proto/v1beta1/run_function.proto
negz: > I still don't feel great about having a well-known key we use "officially" to inject the current environment into an unstructured object.

Sorry, GitHub didn't show me this comment until I refreshed the page.

I think I get your reservations - explicit schemas are usually best. That said, your analogy of the environment being like environment variables, and looking at how other systems use environment variables is making me feel better about the pattern.

If we say that Functions can effectively communicate by reading from and writing to shared "environment variables", i.e. top-level keys in the environment, is it so bad if Crossplane "bootstraps" some of that environment, like GitHub workflows does with their env vars?

phisco: Yeah, I don't feel super strongly against that and I can see you point, so I would be fine with annotations-style reserved keyspaces. I'll revert the whole environment to be just a struct tomorrow. Now the only point is whether we want to wire the current EnvironmentConfig retrieval from configurations or just accept that people are going to access the API server (or any other external system) from functions and just deal with it, which is the topic of the other thread above, so let's continue there if you agree.

internal/controller/apiextensions/composite/composition_functions.go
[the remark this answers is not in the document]
phisco: @negz
> As a concrete example, a Function that read EnvConfigs from the API server would not work with xrender. Or rather, you'd need to somehow plumb it up to an API server to allow it to get access to the types it needs.

Same issue would apply to object retrieval defined at composition level, even more with what we are discussing in #4583.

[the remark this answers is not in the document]
negz: @phisco I don't think it would be quite the same. `xrender` would need to be aware that a Composition could reference EnvConfigs and have a flag like `--env-configs` or something to pass in the EnvConfigs, similar to the `--observed-resources` flag.

That all said, maybe there's a middle ground here where a Function could ask Crossplane to provide the pipeline with "extra" observed resources. I wrote up a sketch of how that might work in https://github.com/crossplane/crossplane/issues/4739.

apis/apiextensions/fn/proto/v1beta1/run_function.proto
phisco: done: https://github.com/crossplane/crossplane/pull/4632/commits/2b28d9a136c6bc33352039596250c9c5923b5779

phisco: called it `context` as you were doing in #4731, as it didn't make much sense to put the current environment in the environment but under a key `.../environment`...

phisco:
I think this can be merged as is, because if we ever decided to embrace functions accessing the API server as we are discussing in the [thread](https://github.com/crossplane/crossplane/pull/4632#discussion_r1348206997) above, we will just have to remove the environment fetching wiring in here and move the logic to a dedicated function without any major disruption for users.

negz:
@phisco Thanks, I agree we can merge this.

I do like your idea of letting `function-patch-and-transform` patch to and from (what we'd now call) the whole context, rather than just the environment key. I could see that being solvable in `function-patch-and-transform` by introducing a new `PatchFromContext` set of patches and leaving the `PatchFromEnvironment` ones for backward compatibility.

negz (approved):

Issue #4316, opened 2023-07-12
Only watch, don't poll, composite resources


negz:
### What problem are you facing?

The XR controller currently watches the XR type, but is also poll-triggered, by default polling desired state every 60 seconds. This interval can be changed by the `--poll-interval` flag. The XR reconciler is poll-triggered because it wants to know when composed resources change, in order to correct drift.

An XR controller doesn't know what kinds of resources it will compose at start time, which is typically when a controller's watches are configured. Furthermore, two different XRs of the same kind might compose completely different types of resources due to using different Compositions.

### How could Crossplane help solve your problem?

Ideally XR reconciliation would be purely watch-triggered. This would result in fewer invocations of the XR reconciler and thus presumably the ability for a particular Crossplane deployment to handle more XRs concurrently. This is also particularly important for Composition Functions, where every XR reconcile may call one or more Functions.

negz:
Thinking out loud, it  may be possible to:

1. Make one controller responsible (only) for selecting a Composition for XRs that don't have one.
2. Start another controller for every unique (XR, Composition) tuple.
3. Restart the (XR, Composition) controller whenever the XR's `spec.resourceRefs` changes

The idea being that each time the (XR, Composition) controller was restarted it would take a watch on all of the (types of?) composed resources that appeared in its `spec.resourceRefs` array.

negz:
FWIW while this relates to Functions I'm not ready to say this is a blocker for Functions going to beta or even GA at this point. I'd like to see some indication that it's a real constraint before we spend time optimizing.

bobh66:
@negz How would this account for the deletion of a composed resource?  I don't think that deletion of a composed resource would trigger an event on the owning composite?  The poll interval will catch this scenario and recreate the composed resource.

negz:
> I don't think that deletion of a composed resource would trigger an event on the owning composite?

I believe it would as long as the controller of the owning composite was watching whatever type of resource was deleted.

negz:
/fresh

@sttts is actively working on this in #4637

negz:
@sttts PS do you want to add this to the v1.14 milestone? I know its behind a feature flag so it would land as alpha.

sttts:
yes, 1.14 is the plan.

negz:
[...]

PR #4637, opened 2023-09-18
feat[compositions]: realtime compositor – part 2: changes to MRs

sttts:

Follow-up of https://github.com/crossplane/crossplane/pull/4582: also reconcile XRs when MRs change. This PR starts informers dynamically for GVRs used in XRs, including function XRs.

This feature is under the `EnableRealtimeCompositions` feature gate (`--enable-realtime-compositions`).

Design:
```
                    ┌─────────────────┐
                    │                 │
                    │     Manager     │
                    │                 │
                    │    ┌───────┐    │
                    │    │       │    │
                    └────┤ Cache ├────┘
                         │       │
                         └───▲───┘◄─────┬──────────┬───────────────────┐
                             │          │          │                   │
                    ┌────────┴────────┐ │ ┌────────┴────────┐ ┌────────┴────────┐
       ┌────────────┤ APIExtensions   │ │ │ APIExtensions   │ │ APIExtensions   │
       │ Composed   │ Definition      │ │ │ Compositions    │ │ Offered         │
┌──────┤ Resource   │ Controller      │ │ │ Controller      │ │ Controller      │ ...
│      │ Informers  │                 │ │ │                 │ │                 │
│      └─────┬──────┤  ┌───────────┐  │ │ │                 │ │                 │
│    creates │      └──┤           ├──┘ │ └─────────────────┘ └─────────────────┘
│  & deletes │         │ GVK Routed│    │
│     ┌──────▼────┐    │ Cache     ├────┘
│     │ Composed  │◄───┤           │fallback
│     │ Resource  │ N:1└─────▲─────┘
│     │ Cache     │          │
│     └───────────┘          │
│                            │ N:1 (dynamically started when an XRD changes)
│                   ┌────────┴────────┐
│                   │ APIExctensions  │
│ informs about gvks│ Composite       │
└───────────────────┤ Controller      │
                    │                 │
                    │                 │
                    └─────────────────┘
```

tl/dr: 
1. the APIExtensions definitions controller (which spawns the composite controllers dynamically through the controller engine), maintains a dynamic set of cached (= informers) per gvk. 
2. The composite controllers report (via the `KindObserver` interface) to the definition controller which GVKs the referenced objects have. For those, the "composedResourceInformers" (part of the definition controller) will start informers as part of a cache. 
3. When those caches become synced, they are registered in the GVK-routed cache as delegates. From that moment on that cache serves those objects. The composite controllers use that GVK-routed-cache for their lookups.
4. Last but not least the "composedResourceInformers" distributes events of changed composed resources to the right composite controllers. For that, an index is added to the XR informer by GVK of the referenced object of each XR instance. Through that it is efficient to find the right XRs for every event.

I have:

### DIFF
--- a/cmd/crossplane/core/core.go
+++ b/cmd/crossplane/core/core.go
@@ -112,6 +112,7 @@ type startCommand struct {
 	EnableExternalSecretStores               bool `group:"Alpha Features:" help:"Enable support for External Secret Stores."`
 	EnableCompositionWebhookSchemaValidation bool `group:"Alpha Features:" help:"Enable support for Composition validation using schemas."`
 	EnableUsages                             bool `group:"Alpha Features:" help:"Enable support for deletion ordering and resource protection with Usages."`
+	EnableRealtimeCompositions               bool `group:"Alpha Features:" help:"Enable support for realtime compositions, i.e. watching composed resources and reconciling compositions immediately when any of the composed resources is updated."`
 
 	EnableCompositionFunctions bool `group:"Beta Features:" default:"true" help:"Enable support for Composition Functions."`
 
@@ -270,6 +271,10 @@ func (c *startCommand) Run(s *runtime.Scheme, log logging.Logger) error { //noli
 		o.ESSOptions = &controller.ESSOptions{
 			TLSConfig: tcfg,
 		}
+	}
+	if c.EnableRealtimeCompositions {
+		o.Features.Enable(features.EnableRealtimeCompositions)
+		log.Info("Alpha feature enabled", "flag", features.EnableRealtimeCompositions)
 	}
 
 	ao := apiextensionscontroller.Options{
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/Masterminds/semver v1.5.0
 	github.com/alecthomas/kong v0.8.1
 	github.com/bufbuild/buf v1.27.0
-	github.com/crossplane/crossplane-runtime v1.14.0-rc.0.0.20231010042232-a8f755740977
+	github.com/crossplane/crossplane-runtime v1.14.0-rc.0.0.20231014010837-674082aae6e6
 	github.com/docker/docker v24.0.6+incompatible
 	github.com/docker/go-connections v0.4.0
 	github.com/go-git/go-billy/v5 v5.5.0
@@ -125,7 +125,7 @@ require (
 	github.com/felixge/fgprof v0.9.3 // indirect
 	github.com/fsnotify/fsnotify v1.6.0 // indirect
 	github.com/go-chi/chi/v5 v5.0.10 // indirect
-	github.com/go-logr/logr v1.2.4 // indirect
+	github.com/go-logr/logr v1.2.4
 	github.com/go-logr/stdr v1.2.2 // indirect
 	github.com/go-logr/zapr v1.2.4 // indirect
 	github.com/go-openapi/jsonpointer v0.19.6 // indirect
@@ -140,7 +140,7 @@ require (
 	github.com/google/go-containerregistry/pkg/authn/kubernetes v0.0.0-20230516205744-dbecb1de8cfa // indirect
 	github.com/google/gofuzz v1.2.0 // indirect
 	github.com/google/pprof v0.0.0-20230926050212-f7f687d19a98 // indirect
-	github.com/google/uuid v1.3.0 // indirect
+	github.com/google/uuid v1.3.0
 	github.com/imdario/mergo v0.3.16 // indirect
 	github.com/inconshreveable/mousetrap v1.1.0 // indirect
 	github.com/jmespath/go-jmespath v0.4.0 // indirect
--- a/internal/controller/apiextensions/composite/reconciler.go
+++ b/internal/controller/apiextensions/composite/reconciler.go
@@ -338,6 +338,14 @@ func WithComposer(c Composer) ReconcilerOption {
 	}
 }
 
+// WithKindObserver specifies how the Reconciler should observe kinds for
+// realtime events.
+func WithKindObserver(o KindObserver) ReconcilerOption {
+	return func(r *Reconciler) {
+		r.kindObserver = o
+	}
+}
+
 type revision struct {
 	CompositionRevisionFetcher
 	CompositionRevisionValidator
@@ -368,6 +376,23 @@ type compositeResource struct {
 	EnvironmentSelector
 	Configurator
 	managed.ConnectionPublisher
+}
+
+// KindObserver tracks kinds of referenced composed resources in composite
+// resources in order to start watches for them for realtime events.
+type KindObserver interface {
+	// WatchComposedResources starts a watch of the given kinds to trigger reconciles when
+	// a referenced object of those kinds changes.
+	WatchComposedResources(kind ...schema.GroupVersionKind)
+}
+
+// KindObserverFunc implements KindObserver as a function.
+type KindObserverFunc func(kind ...schema.GroupVersionKind)
+
+// WatchComposedResources starts a watch of the given kinds to trigger reconciles when
+// a referenced object of those kinds changes.
+func (fn KindObserverFunc) WatchComposedResources(kind ...schema.GroupVersionKind) {
+	fn(kind...)
 }
 
 // NewReconciler returns a new Reconciler of composite resources.
@@ -436,7 +461,8 @@ type Reconciler struct {
 	revision  revision
 	composite compositeResource
 
-	resource Composer
+	resource     Composer
+	kindObserver KindObserver
 
 	log    logging.Logger
 	record event.Recorder
@@ -598,6 +624,14 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		r.record.Event(xr, event.Warning(reasonCompose, err))
 		xr.SetConditions(xpv1.ReconcileError(err))
 		return reconcile.Result{Requeue: true}, errors.Wrap(r.client.Status().Update(ctx, xr), errUpdateStatus)
+	}
+
+	if r.kindObserver != nil {
+		var gvks []schema.GroupVersionKind
+		for _, ref := range xr.GetResourceReferences() {
+			gvks = append(gvks, ref.GroupVersionKind())
+		}
+		r.kindObserver.WatchComposedResources(gvks...)
 	}
 
 	published, err := r.composite.PublishConnection(ctx, xr, res.ConnectionDetails)
--- a/internal/controller/apiextensions/definition/composed.go
+++ b/internal/controller/apiextensions/definition/composed.go
@@ -1,0 +1,283 @@
+/*
+Copyright 2023 The Crossplane Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package definition
+
+import (
+	"context"
+	"strings"
+	"sync"
+
+	"github.com/google/uuid"
+	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	kunstructured "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	kcache "k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
+	cache "sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/cluster"
+	runtimeevent "sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	"github.com/crossplane/crossplane-runtime/pkg/controller"
+	"github.com/crossplane/crossplane-runtime/pkg/logging"
+
+	"github.com/crossplane/crossplane/internal/xcrd"
+)
+
+// composedResourceInformers manages composed resource informers referenced by
+// composite resources. It serves as an event source for realtime notifications
+// of changed composed resources, with the composite reconcilers as sinks.
+// It keeps composed resource informers alive as long as there are composites
+// referencing them. In parallel, the composite reconcilers keep track of
+// references to composed resources, and inform composedResourceInformers about
+// them via the WatchComposedResources method.
+type composedResourceInformers struct {
+	log     logging.Logger
+	cluster cluster.Cluster
+
+	gvkRoutedCache *controller.GVKRoutedCache
+
+	lock sync.RWMutex // everything below is protected by this lock
+
+	// cdCaches holds the composed resource informers. These are dynamically
+	// started and stopped based on the composites that reference them.
+	cdCaches map[schema.GroupVersionKind]cdCache
+	// xrCaches holds the composite resource informers. We use them to find
+	// composites referencing a certain composed resource GVK. If no composite
+	// is left doing so, a composed resource informer is stopped.
+	xrCaches map[schema.GroupVersionKind]cache.Cache
+	sinks    map[string]func(ev runtimeevent.UpdateEvent) // by some uid
+}
+
+type cdCache struct {
+	cache    cache.Cache
+	cancelFn context.CancelFunc
+}
+
+var _ source.Source = &composedResourceInformers{}
+
+// Start implements source.Source, i.e. starting composedResourceInformers as
+// source with h as the sink of update events. It keeps sending events until
+// ctx is done.
+// Note that Start can be called multiple times to deliver events to multiple
+// (composite resource) controllers.
+func (i *composedResourceInformers) Start(ctx context.Context, h handler.EventHandler, q workqueue.RateLimitingInterface, ps ...predicate.Predicate) error {
+	id := uuid.New().String()
+
+	i.lock.Lock()
+	defer i.lock.Unlock()
+	i.sinks[id] = func(ev runtimeevent.UpdateEvent) {
+		for _, p := range ps {
+			if !p.Update(ev) {
+				return
+			}
+		}
+		h.Update(ctx, ev, q)
+	}
+
+	go func() {
+		<-ctx.Done()
+
+		i.lock.Lock()
+		defer i.lock.Unlock()
+		delete(i.sinks, id)
+	}()
+
+	return nil
+}
+
+// RegisterComposite registers a composite resource cache with its GVK.
+// Instances of this GVK will be considered to keep composed resource informers
+// alive.
+func (i *composedResourceInformers) RegisterComposite(gvk schema.GroupVersionKind, ca cache.Cache) {
+	i.lock.Lock()
+	defer i.lock.Unlock()
+
+	if i.xrCaches == nil {
+		i.xrCaches = make(map[schema.GroupVersionKind]cache.Cache)
+	}
+	i.xrCaches[gvk] = ca
+}
+
+// UnregisterComposite removes a composite resource cache from being considered
+// to keep composed resource informers alive.
+func (i *composedResourceInformers) UnregisterComposite(gvk schema.GroupVersionKind) {
+	i.lock.Lock()
+	defer i.lock.Unlock()
+	delete(i.xrCaches, gvk)
+}
+
+// WatchComposedResources starts informers for the given composed resource GVKs.
+// The is wired into the composite reconciler, which will call this method on
+// every reconcile to make composedResourceInformers aware of the composed
+// resources the given composite resource references.
+//
+// Note that this complements cleanupComposedResourceInformers which regularly
+// garbage collects composed resource informers that are no longer referenced by
+// any composite.
+func (i *composedResourceInformers) WatchComposedResources(gvks ...schema.GroupVersionKind) {
+	i.lock.RLock()
+	defer i.lock.RUnlock()
+
+	// start new informers
+	for _, gvk := range gvks {
+		if _, found := i.cdCaches[gvk]; found {
+			continue
+		}
+
+		log := i.log.WithValues("gvk", gvk.String())
+
+		ca, err := cache.New(i.cluster.GetConfig(), cache.Options{})
+		if err != nil {
+			log.Debug("failed creating a cache", "error", err)
+			continue
+		}
+
+		// don't forget to call cancelFn in error cases to avoid leaks. In the
+		// happy case it's called from the go routine starting the cache below.
+		ctx, cancelFn := context.WithCancel(context.Background())
+
+		u := kunstructured.Unstructured{}
+		u.SetGroupVersionKind(gvk)
+		inf, err := ca.GetInformer(ctx, &u)
+		if err != nil {
+			cancelFn()
+			log.Debug("failed getting informer", "error", err)
+			continue
+		}
+
+		if _, err := inf.AddEventHandler(kcache.ResourceEventHandlerFuncs{
+			UpdateFunc: func(oldObj, newObj interface{}) {
+				old := oldObj.(client.Object)
+				obj := newObj.(client.Object)
+				if old.GetResourceVersion() == obj.GetResourceVersion() {
+					return
+				}
+
+				i.lock.RLock()
+				defer i.lock.RUnlock()
+
+				ev := runtimeevent.UpdateEvent{
+					ObjectOld: old,
+					ObjectNew: obj,
+				}
+				for _, handleFn := range i.sinks {
+					handleFn(ev)
+				}
+			},
+		}); err != nil {
+			cancelFn()
+			log.Debug("failed adding event handler", "error", err)
+			continue
+		}
+
+		go func() {
+			defer cancelFn()
+
+			log.Info("Starting composed resource watch")
+			_ = ca.Start(ctx)
+		}()
+
+		i.cdCaches[gvk] = cdCache{
+			cache:    ca,
+			cancelFn: cancelFn,
+		}
+
+		// wait for in the background, and only when synced add to the routed cache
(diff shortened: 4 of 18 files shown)


negz:
Adding a reference to https://github.com/crossplane/crossplane/issues/4316, which I think is related.

cmd/crossplane/core/core.go
```diff
@@ -108,6 +108,7 @@ type startCommand struct {
 	EnableExternalSecretStores               bool `group:"Alpha Features:" help:"Enable support for External Secret Stores."`
 	EnableCompositionWebhookSchemaValidation bool `group:"Alpha Features:" help:"Enable support for Composition validation using schemas."`
 	EnableUsages                             bool `group:"Alpha Features:" help:"Enable support for deletion ordering and resource protection with Usages."`
+	EnableRealtimeCompositions               bool `group:"Alpha Features:" help:"Enable support for realtime compositions, i.e. watching MRs and reconciling compositions immediately when any MR is updated."`
```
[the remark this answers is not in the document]
negz: While I'm nitpicking, "composed resources" would be more accurate than MRs. An XR can compose things like ProviderConfigs and other XRs, as well as MRs. I don't feel strongly about this one though - MRs seems like reasonable shorthand.

internal/controller/apiextensions/composite/reconciler.go
```diff
@@ -593,6 +618,14 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		return reconcile.Result{Requeue: true}, errors.Wrap(r.client.Status().Update(ctx, xr), errUpdateStatus)
 	}
 
+	if r.kindObserver != nil {
+		var gvks []schema.GroupVersionKind
+		for _, ref := range xr.GetResourceReferences() {
+			gvks = append(gvks, ref.GroupVersionKind())
+		}
+		r.kindObserver.Watch(gvks...)
```
negz, line 626: What if a GVK appears in the resource references but then is later removed? This could happen if the XR updates to a CompositionRevision that no longer produces a resource with that GVK. Do we need to stop the watch?

go.mod
```diff
@@ -170,3 +170,5 @@ require (
 	sigs.k8s.io/json v0.0.0-20221116044647-bc3834ca7abd // indirect
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 )
+
+replace github.com/crossplane/crossplane-runtime => github.com/sttts/crossplane-runtime v0.0.0-20231006112446-53296455f75d
```
negz, line 174: We can't merge this with this `replace` statement in place. 😄 Did the changes you need in c/crossplane-runtime get merged?

internal/controller/apiextensions/definition/indexes.go
```diff
@@ -0,0 +1,102 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/workqueue"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	runtimeevent "sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	"github.com/crossplane/crossplane-runtime/pkg/logging"
+	compositeresource "github.com/crossplane/crossplane-runtime/pkg/resource/unstructured/composite"
```
negz, line 33: Would just `composite` work okay here? I don't think we alias this import anywhere else.

internal/controller/apiextensions/definition/indexes.go
```diff
@@ -0,0 +1,102 @@
...
+)
+
+// IndexCompositeResourceRefGVKs assumes the passed object is a composite. It
+// returns gvk keys for every resource referenced in the composite.
+func IndexCompositeResourceRefGVKs(o client.Object) []string {
+	u, ok := o.(*kunstructured.Unstructured)
+	if !ok {
+		return nil // should never happen
+	}
+	x := compositeresource.Unstructured{Unstructured: *u}
```
negz, line 53:
```suggestion
	xr := compositeresource.Unstructured{Unstructured: *u}
```

I've been trying to slowly converge on `xr` as the variable name for composite resources (i.e. XRs) in our codebase, since it's pretty short and hopefully very obvious to folks.

internal/controller/apiextensions/definition/indexes.go
```diff
@@ -0,0 +1,102 @@
...
+	// from spec.resources. Hence, it will also work with functions.
+	compositeResourceRefGVKsIndex = "compositeResourceRefsGVKs"
+	// compositeResourcesRefsIndex is an index of resourceRefs that are owned
+	// by a composite.
+	compositeResourcesRefsIndex = "compositeResourcesRefs"
+)
+
+// IndexCompositeResourceRefGVKs assumes the passed object is a composite. It
+// returns gvk keys for every resource referenced in the composite.
+func IndexCompositeResourceRefGVKs(o client.Object) []string {
```
negz, line 48: Is this and `IndexCompositeResourcesRefs` satisfying some interface? If so, is it worth testing that it does (mostly for documentation purposes)?

internal/controller/apiextensions/definition/managed.go
```diff
@@ -0,0 +1,257 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	"github.com/crossplane/crossplane-runtime/pkg/controller"
+	"github.com/crossplane/crossplane-runtime/pkg/logging"
+
+	"github.com/crossplane/crossplane/internal/xcrd"
+)
+
+type managedResourceInformers struct {
```
negz, line 44: I'd suggest `s/managed/composed/g` for this whole thing. These informers are for composed resources. Managed resources are the most common set of composed resources, but not the only kind. These informers could potentially be handling XRs, ProviderConfigs, Usages, maybe other things in future.

internal/controller/apiextensions/definition/managed.go
```diff
@@ -0,0 +1,257 @@
...
+	log logging.Logger
+	mgr manager.Manager
+
+	mrInformersByCRDName map[schema.GroupVersionKind]mrInformer
```
negz, line 48: I assumed based on the whitespace around variables here that the two below maps were protected by the lock, but this one wasn't. That doesn't seem to be the case - maybe add a comment to `lock` explaining what it protects? Or just moving this down with the other maps would be enough of a hint for me personally.

negz: Also, is this actually `ByCRDName`? Personally I think `informers` would be fine - I can see in the type what the key is pretty easily.

internal/controller/apiextensions/definition/managed.go
```diff
@@ -0,0 +1,257 @@
...
+
+	"github.com/crossplane/crossplane-runtime/pkg/controller"
+	"github.com/crossplane/crossplane-runtime/pkg/logging"
+
+	"github.com/crossplane/crossplane/internal/xcrd"
+)
+
+type managedResourceInformers struct {
+	log logging.Logger
+	mgr manager.Manager
```
negz, line 46: Do we really need the whole manager in here? That's a pretty big interface if I recall correctly. Looks like you're only using `GetClient` and `GetConfig`.

internal/controller/apiextensions/definition/managed.go
```diff
@@ -0,0 +1,257 @@
...
+		if err != nil {
+			cancelFn()
+			log.Debug("failed getting informer", "error", err)
+			continue
+		}
+
+		if _, err := inf.AddEventHandler(kcache.ResourceEventHandlerFuncs{
+			UpdateFunc: func(oldObj, newObj interface{}) {
+				old := oldObj.(client.Object)
+				new := newObj.(client.Object) //nolint:revive // this is fine
```
negz, line 148: Was the linter complaining about redefining `new`? If so, worth mentioning in the comment. (i.e. "Redefining new here is fine"). "This is fine" alone doesn't tell me much as a reader. 🙂

internal/controller/apiextensions/definition/managed.go
```diff
@@ -0,0 +1,257 @@
...
+
+		// wait for in the background, and only when synced add to the routed cache
+		go func(gvk schema.GroupVersionKind) {
+			if synced := ca.WaitForCacheSync(ctx); synced {
+				log.Debug("Managed resource cache synced")
+				i.gvkRoutedCache.AddDelegate(gvk, ca)
+			}
+		}(gvk)
+	}
+} //nolint:govet // the context is stored and will be claused when the outer context is closed.
```
negz, line 190:
```suggestion
} //nolint:govet // the context is stored and will be closed when the outer context is closed.
```

🎅

negz (commented):
I haven't fully reviewed yet, but I need to head out for the afternoon so here's what I have. Mostly nitpicks so far.

internal/controller/apiextensions/composite/reconciler.go
sttts: There is another reconcile loop running every 10s removing old GVKs. Originally, I also started informer that, but changed that. So maybe we can reduce 10s to 30s or a minute or so.

go.mod
sttts: See https://github.com/crossplane/crossplane-runtime/pull/554. Not merged yet.

cmd/crossplane/core/core.go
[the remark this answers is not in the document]
sttts: Changed to "composed resources".

internal/controller/apiextensions/composite/reconciler.go
sttts: Changed to 1m.

internal/controller/apiextensions/definition/indexes.go
sttts: done

internal/controller/apiextensions/definition/indexes.go
sttts: done

internal/controller/apiextensions/definition/indexes.go
sttts: Not an interface, just a function type.

sttts: Added a type assertion.

internal/controller/apiextensions/definition/managed.go
sttts: done

internal/controller/apiextensions/definition/managed.go
sttts: Agree, especially that they are indexed by GVK, not string. Done.

internal/controller/apiextensions/definition/managed.go
sttts: right, `Cluster` is enough.

internal/controller/apiextensions/definition/managed.go
sttts: renamed to `obj`.

internal/controller/apiextensions/definition/composed.go
```diff
@@ -0,0 +1,256 @@
...
+
+	gvkRoutedCache *controller.GVKRoutedCache
+
+	lock      sync.RWMutex // everything below is protected by this lock
+	informers map[schema.GroupVersionKind]mrInformer
+	sinks     map[string]func(ev runtimeevent.UpdateEvent) // by some uid
+	xrs       map[string]xr                                // by controller name
+}
+
+type mrInformer struct {
```
negz, line 56:
```suggestion
type cdInformer struct {
```

I don't know that it's a very intuitive abbreviation, but FWIW we usually abbreviate `composed` to `cd` when we want shorter variable names.

internal/controller/apiextensions/definition/composed.go
```diff
@@ -0,0 +1,256 @@
...
+
+type xr struct {
+	gvk   schema.GroupVersionKind
+	cache cache.Cache
+}
+
+var _ source.Source = &composedResourceInformers{}
+
+// Start implements source.Source, i.e. starting composedResourceInformers as
+// source with h as the sink of update events.
```
negz, line 69: Maybe mention that `Start` can be called more than once. This might be obvious to folks familiar with `source.Source`, but I was confused why `i.sinks` was a map that would seemingly only ever have a single entry until I realized `Start` could be called repeatedly to add new event handlers.

internal/controller/apiextensions/definition/composed.go
```diff
@@ -0,0 +1,256 @@
...
+
+		i.lock.Lock()
+		defer i.lock.Unlock()
+		delete(i.sinks, id)
+	}()
+
+	return nil
+}
+
+// RegisterXR maps a controller name to a GVK.
```
negz, line 95:
```suggestion
// RegisterXRController maps a controller name to a GVK.
```

Perhaps `RegisterXRController` and `UnregisterXRController`? This would make it extra obvious to me that `name` is the name of the controller, not the `metadata.name` of an XR or something.

negz: Or actually, I guess it's really just an XR controller's cache and we happen to be using the controller name as a key. 🤔

internal/controller/apiextensions/definition/composed.go
```diff
@@ -0,0 +1,256 @@
...
+
+	// start new informers
+	for _, gvk := range gvks {
+		if _, found := i.informers[gvk]; found {
+			continue
+		}
+
+		log := i.log.WithValues("gvk", gvk.String())
+
+		ctx, cancelFn := context.WithCancel(context.Background()) //nolint:govet // the outer context is closed by the caller
```
negz, line 128: Perhaps add a comment to remind future contributors to make sure they call `cancelFn()` in error cases? It could also be handy to explicitly point out that we `defer cancelFn` in the goroutine, which is why it's not being done here.

internal/controller/apiextensions/definition/composed.go
```diff
@@ -0,0 +1,256 @@
...
+
+		// wait for in the background, and only when synced add to the routed cache
+		go func(gvk schema.GroupVersionKind) {
+			if synced := ca.WaitForCacheSync(ctx); synced {
+				log.Debug("Managed resource cache synced")
+				i.gvkRoutedCache.AddDelegate(gvk, ca)
+			}
+		}(gvk)
+	}
+} //nolint:govet // the context is stored and will be claused when the outer context is closed.
```
negz, line 189:
```suggestion
} //nolint:govet // the context is stored and will be closed when the outer context is closed.
```

internal/controller/apiextensions/composite/reconciler.go
negz: Is it worth noting on the `Watch` interface that the implementation is responsible for garbage collecting watches that are no longer needed?

internal/controller/apiextensions/definition/composed.go
```diff
@@ -0,0 +1,256 @@
...
+}
+
+// UnregisterXR removes a controller name to GVK mapping.
+func (i *composedResourceInformers) UnregisterXR(name string) {
+	i.lock.Lock()
+	defer i.lock.Unlock()
+	delete(i.xrs, name)
+}
+
+func (i *composedResourceInformers) Watch(gvks ...schema.GroupVersionKind) {
```
negz, line 116: A comment explaining what this method does would help. To replay my own understanding, I think:

* We create a new cache for each GVK we're told we should watch (e.g. each kind of composed resource)
* In practice that cache will be backed by a single informer, for said GVK.
* We get the informer for said GVK.
* We add an event handler to the informer that sends events to all registered sinks (i.e. event handlers)
* In practice each event handler will trigger a reconcile of any XR controller that composes this GVK.
* We tell the routed cache to delegate operations on this GVK to this cache.

internal/controller/apiextensions/definition/reconciler.go
```diff
@@ -115,10 +124,18 @@ func (fn CRDRenderFn) Render(d *v1.CompositeResourceDefinition) (*extv1.CustomRe
...
-	r := NewReconciler(mgr,
+	r := NewReconciler(mgr, o,
 		WithLogger(o.Logger.WithValues("controller", name)),
-		WithRecorder(event.NewAPIRecorder(mgr.GetEventRecorderFor(name))),
-		WithOptions(o))
+		WithRecorder(event.NewAPIRecorder(mgr.GetEventRecorderFor(name))))
+
+	if o.Features.Enabled(features.EnableRealtimeCompositions) {
+		if err := mgr.Add(manager.RunnableFunc(func(ctx context.Context) error {
+			wait.UntilWithContext(ctx, r.xrInformers.reconcileManagedResourcesInformers, time.Minute)
```
negz, line 133: Maybe add a comment explaining that `reconcileManagedResourcesInformers` will run every minute until the context is cancelled. As someone who doesn't use `util/wait` very often I suspected that was the case but had to go look at its documentation to confirm. It's not super obvious from the function name/signature - "wait" doesn't sound like "run this repeatedly".

internal/controller/apiextensions/definition/managed.go
sttts: done

internal/controller/apiextensions/definition/indexes_test.go
```diff
@@ -0,0 +1,120 @@
...
+							"name":       "xr",
+						},
+					},
+				},
+			},
+		}}, want: []string{"nop.crossplane.io/v1alpha1, Kind=NopResource", "nop.example.org/v1alpha1, Kind=NopResource"}},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := IndexCompositeResourceRefGVKs(tt.args.object); !reflect.DeepEqual(got, tt.want) {
```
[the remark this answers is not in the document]
sttts: done

internal/controller/apiextensions/definition/composed.go
sttts: done

internal/controller/apiextensions/definition/composed.go
sttts: done

internal/controller/apiextensions/definition/reconciler.go
sttts: done

internal/controller/apiextensions/definition/composed.go
sttts: done

sttts:
@negz addressed all your comments, added tons of go docs, and rename things to be clearer. I like it.

internal/controller/apiextensions/definition/composed.go
sttts: done

go.mod
negz: FYI, just merged that.

negz:
@sttts LGTM once the crossplane-runtime dependency is updated. Other than that, can you do a pass and `:s/cp/xr/g` for variables pertaining to composite resources. We're trying to move to the latter - I think folks find `xr == composite` to be more self documenting than `cp == composite`.

sttts:
@negz did the cp->xr rename and removed the replace from go.mod.

test/e2e/realtimecompositions_test.go
```diff
@@ -0,0 +1,111 @@
...
+				// 10 seconds is a long time for a realtime composition, but
+				// considerably below the normal reconcile time for a XR.
+				funcs.ResourcesHaveFieldValueWithin(10*time.Second, manifests, "claim.yaml", "status.coolerField", "I'M COOL!"),
+			).
+			WithTeardown("DeletePrerequisites", funcs.AllOf(
+				funcs.DeleteResources(manifests, "setup/*.yaml"),
+				funcs.ResourcesDeletedWithin(3*time.Minute, manifests, "setup/*.yaml"),
+			)).
+			// Disable our feature flag.
+			WithTeardown("DisableAlphaUsages", funcs.AllOf(
```
pedjak, line 105: you are not disabling usages feature here, right?

internal/controller/apiextensions/composite/reconciler.go
```diff
@@ -606,6 +632,14 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		return reconcile.Result{Requeue: true}, errors.Wrap(r.client.Status().Update(ctx, xr), errUpdateStatus)
 	}
 
+	if r.kindObserver != nil {
+		var gvks []schema.GroupVersionKind
+		for _, ref := range xr.GetResourceReferences() {
+			gvks = append(gvks, ref.GroupVersionKind())
+		}
+		r.kindObserver.WatchComposedResources(gvks...)
```
pedjak, line 640: if we now watch composed resources and trigger the reconcile in realtime, should we stop ending the loop with:

```golang
```

and do instead:

```golang
return reconcile.Result{}, errors.Wrap(r.client.Status().Update(ctx, xr), errUpdateStatus)

```

sttts: We can't I guess. 1. environments 2. functions relying on 60s poll.

Should we do everything towards removing poll? Yes, please. @negz

test/e2e/realtimecompositions_test.go
sttts: oops, great spot.

sttts: No, actually this was just cosmetics (feature name). Had no impact.

internal/controller/apiextensions/composite/reconciler.go
pedjak: if environment feature is not enabled and composition is not based on functions, then we could? Also, we could watch environment for changes using the same `WatchComposeResource` approach, but keep in memory relationships between environments and compositions, to know later what composition to reconcile.

test/e2e/apiextensions_test.go
```diff
@@ -53,6 +69,9 @@ func TestCompositionMinimal(t *testing.T) {
 			)).
 			Assess("CreateClaim", funcs.AllOf(
 				funcs.ApplyResources(FieldManager, manifests, "claim.yaml"),
+				funcs.InBackground(funcs.LogResources(nopList)),
```
pedjak, line 72: how is this change related to the work done in this PR? I looks like a leftover of some debugging?

test/e2e/realtimecompositions_test.go
```diff
@@ -0,0 +1,111 @@
...
+			)).
+			Assess("CreateClaim", funcs.AllOf(
+				funcs.ApplyResources(FieldManager, manifests, "claim.yaml"),
+				funcs.InBackground(funcs.LogResources(nopList, withTestLabels)),
+				funcs.InBackground(funcs.LogResources(xnopList, withTestLabels)),
+				funcs.InBackground(funcs.LogResources(nopCrossplaneList, withTestLabels)),
+				funcs.ResourcesCreatedWithin(30*time.Second, manifests, "claim.yaml"),
+				funcs.ResourcesHaveConditionWithin(5*time.Minute, manifests, "claim.yaml", xpv1.Available()),
+			)).
+			Assess("UpdateMR", funcs.AllOf(
```
pedjak, line 85: I think we can even avoid this step, because nop provider updates MRs even above, so line 83 would work even under a way shorter timeout.

internal/controller/apiextensions/composite/reconciler.go
sttts: Am not against that direction, quite the opposite. Worth an attempt in a follow-up to see how far we get (in e2e).

For functions, I would like to see us being explicit that there is no 60s poll guarantee. Maybe we need an API to ask for one if necessary.

test/e2e/apiextensions_test.go
sttts: These tests run under realtime composition as well. And in general, I would like us to have this kind of output to understand the core mechanisms at play in such a test. Hence, I added it also here. There are very likely more places where we need visibility of the main flow of a test.

test/e2e/realtimecompositions_test.go
```diff
@@ -0,0 +1,111 @@
...
+			)).
+			WithSetup("PrerequisitesAreCreated", funcs.AllOf(
+				funcs.ApplyResources(FieldManager, manifests, "setup/*.yaml"),
+				funcs.ResourcesCreatedWithin(30*time.Second, manifests, "setup/*.yaml"),
+				funcs.ResourcesHaveConditionWithin(1*time.Minute, manifests, "setup/definition.yaml", apiextensionsv1.WatchingComposite()),
+				funcs.ResourcesHaveConditionWithin(2*time.Minute, manifests, "setup/provider.yaml", pkgv1.Healthy(), pkgv1.Active()),
+			)).
+			Assess("CreateClaim", funcs.AllOf(
+				funcs.ApplyResources(FieldManager, manifests, "claim.yaml"),
+				funcs.InBackground(funcs.LogResources(nopList, withTestLabels)),
```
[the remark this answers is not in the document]
sttts: Not (only) for debugging. For visibility of the core mechanism at play in a test.

We cannot do these automatically as e.g. `ApplyResources` has no knowledge about what happens with those resources, i.e. the related resources they spawn.

test/e2e/realtimecompositions_test.go
sttts: Don't think. I had to switch the nop manifest to immediate readiness, because it is unexpectedly slow to do so if you e.g. set it to `1s`. So now it is `0s`. And as a composition does reconciles frequently in the beginning for different reasons, I don't want to step into that noise. Hence, this step is explicitly changing something *after* readiness, when the noise is hopefully over.

test/e2e/realtimecompositions_test.go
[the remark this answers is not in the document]
pedjak: ApplyResource could become a bit smarter and understand that we are applying claim, and then via references follow to composite and MRs, or?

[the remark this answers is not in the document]
sttts: PR welcome for an automatic related object tracker :-)

test/e2e/realtimecompositions_test.go
sttts: done

internal/controller/apiextensions/composite/reconciler.go
sttts: That will be step 3: https://github.com/crossplane/crossplane/pull/4822

cmd/crossplane/core/core.go
```diff
@@ -260,6 +261,10 @@ func (c *startCommand) Run(s *runtime.Scheme, log logging.Logger) error { //noli
 			TLSConfig: tcfg,
 		}
 	}
+	if c.EnableRealtimeCompositions {
+		o.Features.Enable(features.EnableRealtimeCompositions)
+		log.Info("Alpha feature enabled", "flag", features.EnableRealtimeCompositions)
+	}
```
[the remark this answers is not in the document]
sttts: https://github.com/crossplane/crossplane/issues/4828

phisco (commented):
This should also delete the claim first and then check all nop resources have been deleted, no?

PR #4656, opened 2023-09-21
feat: switch to new tls certs

### LINKED ISSUE #4305: Create one (or two?) TLS certificates per entity?

### LINKED ISSUE #4707: functions with configurations: authentication handshake failed: tls: failed to verify certificate: x509: certificate is valid for
### What happened?

As a follow-up to the discussion in https://github.com/crossplane/crossplane/pull/4706, it appears that configurations installing functions include the organization from registry

```
kubectl get functions
NAME                                              INSTALLED   HEALTHY   PACKAGE                                                                  AGE
crossplane-contrib-function-patch-and-transform   True        True      xpkg.upbound.io/crossplane-contrib/function-patch-and-transform:v0.1.3   11m
```

but this leads to the following error:
```
  status:
    conditions:
    - lastTransitionTime: "2023-09-29T17:22:16Z"
      message: 'cannot compose resources: cannot run Composition pipeline step "patch-and-transform":
        cannot run Function "crossplane-contrib-function-patch-and-transform": rpc
        error: code = Unavailable desc = last connection error: connection error:
        desc = "transport: authentication handshake failed: tls: failed to verify
        certificate: x509: certificate is valid for function-patch-and-transform.crossplane-system,
        not crossplane-contrib-function-patch-and-transform.crossplane-system"'
      reason: ReconcileError
      status: "False"
      type: Synced
```

it looks like that the tls certificates only created for `function-patch-and-transform.crossplane-system` but not for `crossplane-contrib-function-patch-and-transform.crossplane-system`

```
NAME                                                   HEALTHY   REVISION   IMAGE     STATE    DEP-FOUND   DEP-INSTALLED   AGE
configurationrevision.pkg.crossplane.io/test-test-gz   True      1          test.gz   Active   7           7               14m

NAME                                                                                     HEALTHY   REVISION   IMAGE                                                           STATE    DEP-FOUND   DEP-INSTALLED   AGE
providerrevision.pkg.crossplane.io/crossplane-contrib-provider-helm-503c3591121b         True      1          xpkg.upbound.io/crossplane-contrib/provider-helm:v0.15.0        Active                               14m
providerrevision.pkg.crossplane.io/crossplane-contrib-provider-kubernetes-fd7ab5be249e   True      1          xpkg.upbound.io/crossplane-contrib/provider-kubernetes:v0.9.0   Active                               14m
providerrevision.pkg.crossplane.io/upbound-provider-aws-ec2-89e3839355d3                 True      1          xpkg.upbound.io/upbound/provider-aws-ec2:v0.41.0                Active   1           1               14m
providerrevision.pkg.crossplane.io/upbound-provider-aws-eks-f9ea62d02802                 True      1          xpkg.upbound.io/upbound/provider-aws-eks:v0.41.0                Active   1           1               14m
providerrevision.pkg.crossplane.io/upbound-provider-aws-iam-ebb7af6959e9                 True      1          xpkg.upbound.io/upbound/provider-aws-iam:v0.41.0                Active   1           1               14m
providerrevision.pkg.crossplane.io/upbound-provider-family-aws-a21eb1597cf1              True      1          xpkg.upbound.io/upbound/provider-family-aws:v0.41.0             Active                               13m

NAME                                                                                              HEALTHY   REVISION   IMAGE                                                                    STATE    DEP-FOUND   DEP-INSTALLED   AGE
functionrevision.pkg.crossplane.io/crossplane-contrib-function-patch-and-transform-2a8093bacea0   True      1          xpkg.upbound.io/crossplane-contrib/function-patch-and-transform:v0.1.3   Active                               14m
```

### How can we reproduce it?

```
# need to use crank from https://github.com/crossplane/crossplane/pull/4694
./crank xpkg build --output=test.xpkg
up xpkg xp-extract --from-xpkg test.xpkg -o ./test.gz
kubectl -n crossplane-system cp ./test.gz -c dev $(kubectl -n crossplane-system get pod -l app=crossplane,patched=true -o jsonpath="{.items[0].metadata.name}"):/tmp/cache
kubectl apply -f init/xpkg.yaml
```

```
---
kind: Configuration
metadata:
  name: test
spec:
  package: test.gz
  packagePullPolicy: Never
```

Crossplane version:

phisco:

> [!IMPORTANT]
> We need to let people know that if they had External Secret Stores enabled before, they need to manually delete the `ess-server-certs` before upgrading, see [below](https://github.com/crossplane/crossplane/pull/4656#issuecomment-1746882547).

Fixes #4305.
Fixes https://github.com/crossplane/crossplane/issues/4707.
Supersedes https://github.com/crossplane/crossplane/pull/4590 and https://github.com/crossplane/crossplane/pull/4709.

Follow up to https://github.com/crossplane/crossplane/pull/4485 to actually use the secrets created there and already used by Functions in #4500.

To recap a bit of context, before #4485 we were creating the following secrets:
- `webhook-tls-secret`, used for:
  - Crossplane's validatingWebhookConfigurations and CRDs if needed
  - Provider's validatingWebhookConfigurations and CRDs if needed
- ExternalSecretStores related secrets, if enabled:
  - `ess-ca-certs`: the CA used to generated the following 2 secrets and used by components to verify the tls certificate served by any ESS.
  - `ess-server-certs`: the single server certificate any ESS plugin would use to serve tls traffic
  - `ess-client-certs`: the single client certificate any component wanting to call any ESS should use.

#4485 added to the above the following certificates, with the goal to then start using the proper ones with Functions and switch the ones above for the right ones later on (this PR):
- `crossplane-root-ca`: the root CA used by Crossplane to generate any other certificate below.
- `crossplane-tls-client`: the client certificate Crossplane should use to call any other component (ESS and Functions).
- `crossplane-tls-server`: the server certificate Crossplane should use to serve any tls traffic (webhook)
- `<provider>-tls-client`: one for each Provider, to be used to call any other component.
- `<provider>-tls-server`: one for each provider, to be used serve any tls traffic (webhook)
- `<function>-tls-server`: one for each function, to be used to serve any tls traffic (grpc)

In order to achieve the scenario in the diagram below:
image

This PR removes the following secrets:
- `webhook-tls-secret`, replacing its usages with:
  - `crossplane-root-ca` to generate any other certificate needed
  - `crossplane-tls-server` for Crossplane's webhook and CRDs
  - `<provider>-tls-server` for each providers' webhook and CRDs
- ESS related ones:
  -  `ess-ca-certs`, replaced by `crossplane-root-ca`
  - `ess-client-certs`, replaced by:
    - `<provider>-tls-client` for the providers to communicate with ESS
    - `crossplane-tls-client` for Crossplane to communicate with ESS

So, the only remaining certificates are:
- `crossplane-root-ca`
- `crossplane-tls-server`
- `crossplane-tls-client`
- `<provider>-tls-server`
- `<provider>-tls-client`
- `<function>-tls-server`
- `ess-server-certs`: which we still need to create as part of the helm-chart and not per ESS, because ESS are not deployed as packages.

In order to test it we added a validtingWebhook to [provider-nop](https://github.com/crossplane-contrib/provider-nop/pull/10), so that we could actually test a provider using webhooks as it looks like no other provider is using it at the moment (documentation about how to do so is actually missing, but we'll address that later). A release still has to be cut with the change, so in the meantime I used `docker.io/phisco/provider-nop-phisco:v0.0.2-webhook`.

Notes:
- providers will be restarted as their deployment is updated with the new environment variables and secrets mounted. 
- we update the current providerRevision's `spec` to set the new `tlsServerSecretName` and `tlsClientSecretName`, preserving the old value of `webhookTLSSecretName`.
- Preserved the current undocumented `WEBHOOK_TLS_CERT_DIR` mirroring the content of the new `TLS_SERVER_CERTS_DIR` to allow transitioning providers using webhooks (if any). Did the same for `ESS_TLS_CERTS_DIR`, mirroring `TLS_CLIENT_CERTS_DIR` to ease providers transition.
- switched provider's service to be per provider instead of per provider revision, similar to what we already did for functions, in order to have stabler names to set in the server certificates for the provider's webhooks; cleaning up the old per revision service if found.
- Switched all resources to be owned by revisions as that ensures we behave correctly on updates and that, if required, deleting the active revision results in all things being properly deleted.

I have:


internal/controller/pkg/revision/hook.go
phisco, line 1: feels like we could benefit from some refactoring around the way we run pre hook, establisher and post hook. 
Turns out the establisher needs the tls server certificate of the component to be able to inject the necessary ca into the webhooks and crds, so those need to be created by the pre hook, while the deployment can still be created by the post hook.
I'll spend a bit more time to understand how to refactor it.

phisco:
there is a lot of overlap with https://github.com/crossplane/crossplane/pull/4071, review that first and I'll then rebase this one to leverage new `Deactivate` function introduced there, or at least keep that in mind while reviewing this PR.

internal/controller/pkg/revision/deployment.go
```diff
@@ -71,31 +68,47 @@ const (
 	tlsClientCertsDir        = "/tls/client"
 )
 
+func buildProviderSecrets(revision v1.PackageRevision, namespace string) (serverSec *corev1.Secret, clientSec *corev1.Secret) {
```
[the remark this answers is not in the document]
turkenh: Thanks, I was already planning to do something similar in the context of https://github.com/crossplane/crossplane/issues/4699

internal/controller/pkg/revision/deployment.go
```diff
@@ -71,31 +68,47 @@ const (
...
+				Name:            *tlsClientSecretName,
+				Namespace:       namespace,
+				OwnerReferences: []metav1.OwnerReference{meta.AsController(meta.TypedReferenceTo(revision, revision.GetObjectKind().GroupVersionKind()))},
+			},
+		}
+	}
+	return serverSec, clientSec
+}
+
+func buildProviderService(provider *pkgmetav1.Provider, revision v1.PackageRevision, namespace string) *corev1.Service {
```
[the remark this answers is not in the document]
turkenh: Maybe we can pass selectors as input to avoid that hidden dependency 🤔

internal/controller/pkg/revision/deployment.go
```diff
@@ -171,52 +184,31 @@ func buildProviderDeployment(provider *pkgmetav1.Provider, revision v1.PackageRe
...
 	if revision.GetTLSServerSecretName() != nil {
 		mountTLSSecret(*revision.GetTLSServerSecretName(), tlsServerCertsVolumeName, tlsServerCertsDir, tlsServerCertDirEnvVar, d)
+		d.Spec.Template.Spec.Containers[0].Ports = append(d.Spec.Template.Spec.Containers[0].Ports,
+			corev1.ContainerPort{
+				Name:          webhookPortName,
+				ContainerPort: servicePort,
+			})
+		// for backward compatibility with existing providers, we set the
+		// environment variable WEBHOOK_TLS_CERT_DIR to the same value as
+		// TLS_SERVER_CERTS_DIR to ease the transition to the new certificates.
```
turkenh, line 194: What changes are required for providers to switch to the new structure? I assume it is just using `TLS_SERVER_CERTS_DIR` instead of `WEBHOOK_TLS_CERT_DIR` and `TLS_CLIENT_CERTS_DIR` instead of `ESS_TLS_CERTS_DIR`, anything else?
It could be helpful to open an example PR in one of the providers to demonstrate that.

internal/controller/pkg/revision/establisher.go
```diff
@@ -230,31 +230,35 @@ func (e *APIEstablisher) validate(ctx context.Context, objs []runtime.Object, pa
 				if len(webhookTLSCert) == 0 {
 					return nil
 				}
+				svcName := parent.GetName()
 				if pkgRef, ok := GetPackageOwnerReference(parent); ok {
 					conf.SetName(fmt.Sprintf("crossplane-%s-%s", strings.ToLower(pkgRef.Kind), pkgRef.Name))
+					svcName = pkgRef.Name
```
turkenh, line 236: Could we always use `rev.GetLabels()[v1.LabelParentPackage]` as `svcName` ?

internal/controller/pkg/revision/hook.go
```diff
@@ -112,12 +111,46 @@ func (h *ProviderHooks) Pre(_ context.Context, pkg runtime.Object, pr v1.Package
 
 	// TODO(hasheddan): update any status fields relevant to package revisions.
```
turkenh, line 113: Nit: consider returning early

```go
	if pr.GetDesiredState() != v1.PackageRevisionActive {
		return nil
	}
```

internal/controller/pkg/revision/hook.go
```diff
@@ -112,12 +111,45 @@ func (h *ProviderHooks) Pre(_ context.Context, pkg runtime.Object, pr v1.Package
...
 
+	if pr.GetDesiredState() == v1.PackageRevisionActive {
+		return h.ensurePrerequisites(ctx, pkgProvider, pr)
```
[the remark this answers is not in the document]
turkenh: > we need to setup the tls server secret before the APIEstablisher try to use it to inject the certificate into CRDs/Webhooks.

How was it working before? Or was it working at all? 🤔

internal/controller/pkg/revision/hook.go
```diff
@@ -182,22 +199,12 @@ func (h *ProviderHooks) Deactivate(ctx context.Context, pr v1.PackageRevision) e
...
 		return errors.Wrap(err, errDeleteProviderSA)
 	}
-	// Delete the service if it exists.
+
+	// TODO(phisco): only added to cleanup the service we were previously
+	// 	deploying for each provider revision, remove in a future release.
 	if err := h.client.Delete(ctx, &corev1.Service{ObjectMeta: metav1.ObjectMeta{Name: pr.GetName(), Namespace: h.namespace}}); resource.IgnoreNotFound(err) != nil {
 		return errors.Wrap(err, errDeleteProviderService)
 	}
-	// Delete the TLS Server Secret if it exists.
```
turkenh, line 189: Would you mind adding a note as a code comment on why we don't need to delete TLS secrets (and service?) during deactivation?

internal/controller/pkg/revision/hook.go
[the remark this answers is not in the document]
turkenh: A code comment with the above information would be helpful for posterity.

internal/controller/pkg/revision/hook.go
```diff
@@ -289,29 +331,35 @@ func (h *FunctionHooks) Post(ctx context.Context, pkg runtime.Object, pr v1.Pack
...
+	secSer := buildFunctionSecret(pr, h.namespace)
 	if err := h.client.Apply(ctx, secSer); err != nil {
 		return errors.Wrap(err, errApplyFunctionSecret)
 	}
-	if err := initializer.NewTLSCertificateGenerator(h.namespace, initializer.RootCACertSecretName, pkgFunction.Name,
-		initializer.TLSCertificateGeneratorWithServerSecretName(pr.GetTLSServerSecretName()),
-		initializer.TLSCertificateGeneratorWithOwner(owner)).GenerateServerCertificate(ctx, h.client); err != nil {
+	if err := initializer.NewTLSCertificateGenerator(h.namespace, initializer.RootCACertSecretName,
+		initializer.TLSCertificateGeneratorWithServerSecretName(secSer.GetName(), initializer.DNSNamesForService(svc.Name, svc.Namespace)),
+		initializer.TLSCertificateGeneratorWithOwner([]metav1.OwnerReference{meta.AsController(meta.TypedReferenceTo(pr, pr.GetObjectKind().GroupVersionKind()))})).Run(ctx, h.client); err != nil {
```
turkenh, line 351: Should these be removed from here, given we have them in `(h *FunctionHooks) ensurePrerequisites` ?

turkenh (commented):
Thanks a lot for your efforts on refactoring and cleaning up the codebase @phisco 🙌 

It would be great if we could do a quick testing with [ESS](https://docs.crossplane.io/knowledge-base/integrations/vault-as-secret-store) to make sure we don't break anything there.

internal/controller/pkg/revision/hook.go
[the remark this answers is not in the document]
phisco: It was working before because we were using the only webhook we had, the same crossplane was using, which was using a pretty wide *.<namespace>. I'll add a comment 👍

cluster/charts/crossplane/templates/deployment.yaml
```diff
@@ -164,17 +163,9 @@ spec:
...
-          - name: "WEBHOOK_TLS_CERT_DIR"
-            value: /webhook/tls
-          {{- end }}
-          {{- if $externalSecretStoresEnabled }}
-          - name: "ESS_TLS_SECRET_NAME"
-            value: ess-client-certs
-          - name: "ESS_TLS_CERTS_DIR"
-            value: /ess/tls
+          {{- if not .Values.webhooks.enabled }}
+          - name: "WEBHOOK_DISABLED"
```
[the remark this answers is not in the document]
phisco: you know how I feel about booleans defaulting to true 😆

cluster/charts/crossplane/templates/secret.yaml
```diff
@@ -55,6 +21,7 @@ metadata:
   name: crossplane-root-ca
   namespace: {{ .Release.Namespace }}
 type: Opaque
+{{- if .Values.webhooks.enabled }}
```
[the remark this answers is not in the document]
phisco: good catch, that was a last minute cleanup I thought I tested manually but I clearly didn't 😅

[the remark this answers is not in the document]
phisco: done: https://github.com/crossplane/crossplane/pull/4656/commits/de0efff96490507f683fa3e0f69deef69497a035

internal/controller/pkg/revision/hook.go
phisco: done: https://github.com/crossplane/crossplane/pull/4656/commits/8faf3845cacd95fb1a656b787c8ac0a1416d7688

internal/controller/pkg/revision/hook.go
phisco: done: https://github.com/crossplane/crossplane/pull/4656/commits/3e8db9d7e82b5f81faa03661126a37092f15e7cb

internal/controller/pkg/revision/hook.go
phisco: done: https://github.com/crossplane/crossplane/pull/4656/commits/96c9e226ee4ad7aa55ff640de121edc1cdf10714

internal/controller/pkg/revision/deployment.go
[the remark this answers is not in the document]
phisco: refactored in https://github.com/crossplane/crossplane/pull/4656/commits/c5d8ae5aee73b86a43fceba37ae1708b7fbb0554

cluster/charts/crossplane/templates/deployment.yaml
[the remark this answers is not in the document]
phisco: done: https://github.com/crossplane/crossplane/pull/4656/commits/deaec580af85804f4c939c667f2d022750ab5bef

turkenh (approved):

phisco:
manually tested ESS following https://docs.crossplane.io/knowledge-base/integrations/vault-as-secret-store/ and deploying the configuration from https://github.com/crossplane/crossplane/issues/4707.

apis/pkg/v1/interfaces.go
```diff
@@ -542,21 +534,6 @@ func (p *ProviderRevision) SetSkipDependencyResolution(b *bool) {
 	p.Spec.SkipDependencyResolution = b
 }
 
-// GetWebhookTLSSecretName of this ProviderRevision.
-func (p *ProviderRevision) GetWebhookTLSSecretName() *string {
-	return p.Spec.WebhookTLSSecretName
```
turkenh, line 547: Any reason not to clean those out from the spec? https://github.com/crossplane/crossplane/blob/2e6a7277e315c4150332e12f9cd2e6c8cd421ede/apis/pkg/v1/revision_types.go#L91-L103

phisco: Not really, I didn't want to remove fields from the API, but if we are fine with it we can definitely do that!

turkenh: I can take care of that in https://github.com/crossplane/crossplane/issues/4699
Just wanted to know if there was a reason.

phisco:
ESS certificates were previously signed by a dedicated CA (ess-ca-certs), while now we sign and expect them to be signed by the single root CA we use for everything (crossplane-root-ca). However, if we find secrets to be already populated, we don’t overwrite them as we assume these could have been provided by the user instead, which is never going to be the case, given that the Chart doesn’t allow it atm. 

This means that on an upgrade from a previous version with ess enabled, the old certificate will still be there. Therefore, we are going to try to connect to it with client certificates (crossplane’s or providers’ new ones) that are not going to be recognised properly. 

The manual solution is to delete the ess-server-certs secret manually before upgrading, I feel like it could be ok for an alpha feature, but WDYT @turkenh @ezgidemirel?

The optimal solution would be to introduce proper support for user provided certificates and certificate rotation if certs were to be handled by crossplane itself, but that will be more complex to implement and is not a priority atm.

phisco:
We agreed to merge the PR as is and potentially address the above issue in a follow up PR if we have time in this release cycle.

PR #4700, opened 2023-09-28
Deploy to kind cluster with debugging on

pedjak:
While development/testing locally, it is very useful to have
debug logs turned on.

I have:


phisco (approved):
I agree, I always run it as `HELM3_FLAGS="--set args={--debug} " ./cluster/local/kind.sh helm-upgrade`. The main issue of setting this as default is that helm is not going to merge args if you add custom ones, so you'll have to remember to add `--debug` it back again later, but I agree it's a better default.

bobh66 (approved):

Issue #4660, opened 2023-09-22
Enable user to get insights on composite reconciliation


pedjak:
### What problem are you facing?

If a composition does not produce an expected outcome, users can look at resources status to understand more what happened. However, if the resources are generated without errors, but do not much the expectations, it is quite common to end up in a trial and error loop until the composition is finally fixed.

### How could Crossplane help solve your problem?

User should be able to activate developer mode for a given composite by placing `crossplane.io/dev-mode` annotation on it. The annotation value could be simply `true`, or more fine grained, e.g. it could state explicitly what composition functionalities/aspects should be set in the developer mode.

With the dev mode on, at the end of composite reconciliation, the participating resources are enriched with following data:
* each resource gets `dev.crossplane.io/prev-version` annotation containing the version of the resource at the reconciliation start
* each resource gets `dev.crossplane.io/env` annotation containing the used environment
* each resource gets `dev.crossplane.io/diff` annotation containing the resource diff generated by the composition
* composite resource contains  `dev.crossplane.io/env-bootstrap` annotation containing details about how the initial environment is constructed, i.e. resource fetch order, their content, etc.

sttts:
I like the direction of making relevant information more visible. I have some questions and concerns. Not everything must be an annotation. We can log, we can create configmap, we can create events.

In general, writing edge data (like a diff or prev-version) is a bad idea. Convergence is not trivial. Risk of update loops. And with realtime compositions, these don't make much sense. Edges will be often and last only short time. An event or logging are better here.

dee0sap:
Hey @pedjak 

Maybe it would good to start by listing where, from a Crossplane consumer's point of view, problems may arise.      Off the top of my head, I can think of 
- Deploy tech ( helm, kustomize... ) did not produce expected results
- A patch was malformed or its transform didn't work as expected
- Selector that selects a k8s object produced didn't work as expected 
I am sure I am missing somethings.

Anyhoo, if you had such a list then we could consider what would useful for debugging each.  Hopefully a common theme emerges.  

About logging... some years ago I worked on a commercially sold on-prem product, a collection of services, and we ended up with 
- 'normal' log files that were mostly for our dev team and our support.   For the customer these were basically useless
- an audit log - Designed specifically for customer consumption.  Used terms/language they would always be able to understand
- event log - Also designed specifically for customer consumption.   Terms/language always could track back to the admin UI and or documentation.
[...]

PR #4702, opened 2023-09-28
WIP enable insight into constructed environment

pedjak:

WIP demonstration how we could enable users to get more insights into the composition reconciliation process.

The main purpose of this PR is to trigger the conversation on #4660

How to play with this:
* build and deploy Crossplane from this PR with environment configs enabled
* add `alpha.crossplane.io/dev-mode: env` to a claim
* eventually the counterpart composite is equipped with `alpha.crossplane.io/env-bootstrap` annotation containing the environment data before the composition was processed

Signed-off-by: Predrag Knezevic <predrag.knezevic@upbound.io>

### DIFF
--- a/internal/controller/apiextensions/composite/reconciler.go
+++ b/internal/controller/apiextensions/composite/reconciler.go
@@ -19,8 +19,11 @@ package composite
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"k8s.io/utils/set"
 	"strconv"
+	"strings"
 	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -581,7 +584,19 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		xr.SetConditions(xpv1.ReconcileError(err))
 		return reconcile.Result{Requeue: true}, errors.Wrap(r.client.Status().Update(ctx, xr), errUpdateStatus)
 	}
-
+	if devMode, ok := xr.GetAnnotations()["alpha.crossplane.io/dev-mode"]; ok {
+		do := set.New(strings.Split(devMode, ",")...)
+		if do.Has("env") && env != nil {
+			jsonEnv, _ := json.MarshalIndent(env.Object, "", "  ")
+			meta.AddAnnotations(xr, map[string]string{
+				"alpha.crossplane.io/env-bootstrap": string(jsonEnv),
+			})
+		}
+		if err := r.client.Update(ctx, xr); err != nil {
+			xr.SetConditions(xpv1.ReconcileError(err))
+			return reconcile.Result{Requeue: true}, errors.Wrap(r.client.Status().Update(ctx, xr), errUpdateStatus)
+		}
+	}
 	// TODO(negz): Pass this method a copy of xr, to make very clear that
 	// anything it does won't be reflected in the state of xr?
 	res, err := r.resource.Compose(ctx, xr, CompositionRequest{Revision: rev, Environment: env})


pedjak:
@ytsarev @bobh66 you can use e2e manifests to quickly see the outcome:

* `kubectl apply -f apply -f test/e2e/manifests/apiextensions/environment/setup`
* `kubectl apply -f apply -f test/e2e/manifests/apiextensions/environment/default/setup`
* create claim with the new annotation:

```yaml
kind: SQLInstance
metadata:
  name: example
  namespace: default
  annotations:
    alpha.crossplane.io/dev-mode: env
  labels:
    stage: prod
spec:
  parameters:
    storageGB: 10
```

Eventually, the environment content is published as `alpha.crossplane.io/env-bootstrap` annotation on the composite:

```yaml
kind: XSQLInstance
metadata:
  annotations:
    alpha.crossplane.io/dev-mode: env
    alpha.crossplane.io/env-bootstrap: |-
      {
        "apiVersion": "internal.crossplane.io/v1alpha1",
        "complex": {
          "a": "b",
          "c": {
            "d": "e",
            "f": "2"
          }
        },
        "kind": "Environment"
      }
```

An alternative place for publishing would be somewhere under `status.devmode`. In parallel to that we could also generate events/logs.

negz:
If we write to XR status, and once we resolve https://github.com/crossplane/crossplane/issues/4600, I believe this could be implemented as a Function. That could give us some extra flexibility - i.e. I could imagine being able to choose whether you wanted to log the environment (in the Function logs), emit it as an event, write it to status, etc.

bobh66:
If the annotation reconciles downward from the claim it will enable debugging on all nested composites, which could be several dozen or more.  That may or may not be desirable?

pedjak:
> If the annotation reconciles downward from the claim it will enable debugging on all nested composites, which could be several dozen or more. That may or may not be desirable?

From what I could see in [the code](
https://github.com/crossplane/crossplane/blob/master/internal/controller/apiextensions/composite/composition_render.go#L121), we do not pass any annotations from claim to MRs, only a few labels. Ofcourse, if needed, we could do that as well, so that we could debug the whole chain.

jankaacc:
Very happy to see this feature, we spent a lot of time debugging what is in the runtime environment. Although the provided [example](https://github.com/crossplane/crossplane/blob/master/test/e2e/manifests/apiextensions/environment/default/setup/composition.yaml) could be a bit extended.

The most difficult part with debugging environment configs is when we are dynamically patching values to the environment in one resource and then we are relying on those values in another resource.

This is a short example of what I am talking about:

Lets say we have existing EnvironmentConfig like

```yaml
kind: EnvironmentConfig
metadata:
  name: example-environment-1
data:
  devEnv: staging
```

Then in Composition we are selecting this Env Config and we do some bidirectional patching

Selection:

```yaml
...
spec:
  environment:
    environmentConfigs:
      - ref:
          name: example-environment-1
...
```

Patching

```yaml
resources:
    - name: resource1
      base:
        ...
      patches:
          fromFieldPath: status.someValueAfterMRCreation
          toFieldPath: myMRProp
    
    - name: resource2
      base:
        ...
      patches:
        - type: CombineFromEnvironment
           combine:
             variables:
               - fromFieldPath: devEnv
               - fromFieldPath: myMRProp
             strategy: string
             string:
               fmt: '%s-%s'
           toFieldPath: spec.forProvider.name
           fromFieldPath:
             policy: Required
```

The problem we were often facing was that `resource2` was not created because `status.someValueAfterMRCreation` was missing. 
It was time-consuming to debug the problem as we needed to check dependent resources and their statuses, If we could see the run time environment it would have saved us a lot of time. 
So very happy to see this functionality coming, but I would test this with an example which is patching the environment dynamically.

phisco:
We still have to merge and take a decision for https://github.com/crossplane/crossplane/pull/5061, but the overall agreement is to lean more and more on functions. In the meantime support for `environment` was added to Functions, and therefore this could either become a topic for `function-patch-and-transform`, if we still want to have more visibility inside the single function, or as suggested  by @negz [above](https://github.com/crossplane/crossplane/pull/4702#issuecomment-1739914337), the goal of a dedicated `debug` Function.

So, as agreed with @pedjak, this PR can be closed.

PR #4705, opened 2023-09-28
Support loading per-pipeline-step credentials from a Secret

negz:

Fixes https://github.com/crossplane/crossplane/issues/3718 

For example:

```yaml
kind: Composition
metadata:
  name: example
spec:
  mode: Pipeline
  pipeline:
  - step: run-a-function-that-needs-credentials
    functionRef:
      name: function-awscli
    credentials:
      source: Secret
      secretRef:
        name: aws-cli-creds
```

This API attempts to leave space for other secret sources in future, but for now we just support secrets. I could imagine supporting things like IRSA similar to how a ProviderConfig does. I considered adding a new FunctionConfig type, but that felt a bit like overkill.

I have:

- [ ] Added or updated unit **and** E2E tests for my change.
- [ ] Added `backport release-x.y` labels to auto-backport this PR, if necessary.


negz:
Draft for now, because I haven't touched E2E tests and because I'm not sure whether this is the best API. I think I mostly like it, but it irks me a little that we have both `input` and `credentials`. The latter feels like an optional subset of the former.

negz:
I'm in part leaving this in draft for now because I'm not sure whether this is the API we want, as opposed to https://github.com/crossplane/crossplane/issues/3718#issuecomment-1739996309.

negz:
Closing for now - I don't think it's clear that we'll take this approach. We can reopen once we get to tackling it in the next release.

PR #4714, opened 2023-10-02
fix(message): fix event message to print composition revision

### LINKED ISSUE #4708: Selected composition revision: %!s(MISSING)
### What happened?
got the following event in composite resource: `Selected composition revision: %!s(MISSING)`
is related to: https://github.com/crossplane/crossplane/blob/master/internal/controller/apiextensions/composite/reconciler.go#L541

Crossplane version:

haarchri:
before this PR:
`Selected composition revision: %!s(MISSING)`

after this PR:
`Selected composition revision: xclusters.aws.caas.upbound.io-5ebaa5e`

Fixes #4708

I have:

### DIFF
--- a/internal/controller/apiextensions/composite/reconciler.go
+++ b/internal/controller/apiextensions/composite/reconciler.go
@@ -538,7 +538,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		return reconcile.Result{Requeue: true}, errors.Wrap(r.client.Status().Update(ctx, xr), errUpdateStatus)
 	}
 	if rev := xr.GetCompositionRevisionReference(); rev != nil && (origRev == nil || *rev != *origRev) {
-		r.record.Event(xr, event.Normal(reasonResolve, "Selected composition revision: %s", rev.Name))
+		r.record.Event(xr, event.Normal(reasonResolve, fmt.Sprintf("Selected composition revision: %s", rev.Name)))
 	}
 
 	// TODO(negz): Update this to validate the revision? In practice that's what


negz (approved):
Thanks!

Perhaps we should have an `event.Normalf` for this, as a future improvement. `event.Normal` takes KV pairs as variadic args and includes them with its events, but those aren't super easy to find (I think you have do `kubectl describe events` to see them.

internal/controller/apiextensions/composite/reconciler.go
```diff
@@ -538,7 +538,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		return reconcile.Result{Requeue: true}, errors.Wrap(r.client.Status().Update(ctx, xr), errUpdateStatus)
 	}
 	if rev := xr.GetCompositionRevisionReference(); rev != nil && (origRev == nil || *rev != *origRev) {
-		r.record.Event(xr, event.Normal(reasonResolve, "Selected composition revision: %s", rev.Name))
+		r.record.Event(xr, event.Normal(reasonResolve, fmt.Sprintf("Selected composition revision: %s", rev.Name)))
```
phisco, line 541: Normal takes key/value pairs from the 3rd argument on, so the other option here is to make it
```suggestion
		r.record.Event(xr, event.Normal(reasonResolve, "Selected composition revision", "name", rev.Name)))
```

phisco: Looking better at it, we don't use annotations much, just in a few places, these are not shown when viewing events, I'd say we can continue using `fmt.Sprintf`, wdyt? CC @negz @sttts

phisco (approved):

Issue #4671, opened 2023-09-23
Improve discoverability (add docs and pointers) for installing Crossplane CLI


jbw976:
As we invest further into the Crossplane CLI for v1.14 as part of #4648, we should definitely improve the discoverability of the tool.  Just now I was forgetting how to install it, so I searched the docs and looked at the c/c main readme, and couldn't find the command to install the CLI.

This should be much easier to find so other folks don't get lost when trying to install the Crossplane CLI. At a minimum, this page should explain how to install the CLI: https://docs.crossplane.io/latest/software/install/

For posterity, I was able to install latest `stable` version with:
```
curl https://raw.githubusercontent.com/crossplane/crossplane/master/install.sh | sh
```

And install latest `master` version with:
```
curl https://raw.githubusercontent.com/crossplane/crossplane/master/install.sh | XP_CHANNEL=master sh
```

jbw976:
Fixed by https://github.com/crossplane/docs/pull/584

PR #4721, opened 2023-10-03
xpkg push

### LINKED ISSUE #4669: `build/compile` and `push` Functions to a registry
We need to invest further in the `build` and `push` commands for Composition Functions so they are able to create complete and compliant Crossplane packages that contain the Function implementation and all the proper metadata.

There is early support for building a Function package in the `master` branch that was added in #4525, but this functionality needs to be expanded as described in https://github.com/crossplane/crossplane/pull/4525#pullrequestreview-1596578473.

> I think longer term we'll want to steal the xpkg functionality from https://github.com/upbound/up so that we can build Functions where the package metadata and controller are just different layers in the same OCI image. For now this makes sense though.

Further details about this improvement can be found in the template repo for a golang based XFN: https://github.com/crossplane/function-template-go/blob/main/Dockerfile#L15-L19

This issue depends on #4657 to use the supporting `xpkg` functionality from `up`.

lsviben:

Moved a simplified version of `crossplane push provider/configuration` to `crossplane xpkg push`. 

Not moving the whole `up xpkg push`, at least for [now](https://github.com/crossplane/crossplane/issues/4669#issuecomment-1745272000).
The new version also supports functions, while the old one should be deprecated in time. 

Fixes #4669

I have:

Docs PR for CLI will be done separately in https://github.com/crossplane/crossplane/issues/4671

### How has this been tested

Ran `kubectl crossplane push provider <tag>` and the new `kubectl crossplane xpkg push <tag>` and got the same result.

### DIFF
--- a/cmd/crank/xpkg/push.go
+++ b/cmd/crank/xpkg/push.go
@@ -1,0 +1,89 @@
+/*
+Copyright 2023 The Crossplane Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package xpkg
+
+import (
+	"os"
+
+	"github.com/google/go-containerregistry/pkg/authn"
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/google/go-containerregistry/pkg/v1/tarball"
+	"github.com/spf13/afero"
+
+	"github.com/crossplane/crossplane-runtime/pkg/errors"
+	"github.com/crossplane/crossplane-runtime/pkg/logging"
+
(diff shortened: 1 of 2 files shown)


phisco:
Trying to diff this w.r.t. `up xpkg push` as defined [here](https://github.com/upbound/up/blob/main/cmd/up/xpkg/push.go#L210):
- `up` supports all the [upbound.Flags](https://github.com/upbound/up/blob/9654d12b9966135c21f93cf1f901364fe9b6dfc6/internal/upbound/flags.go#L25), which allow to specify Upbound specific settings to access the marketplace, we won't
- `up` creates a repository in the marketplace if it doesn't exist (`--create true`), we won't
- `up` supported passing multiple [packages](https://github.com/upbound/up/blob/9654d12b9966135c21f93cf1f901364fe9b6dfc6/cmd/up/xpkg/push.go#L69), we won't.
- `up` inferred desired annotations and applied them for flattened packages as introduced [here](https://upboundio.slack.com/archives/C8P85J3BJ/p1696466305377369), we won't

Diffing with `crank push <configuration|provider>` as defined [here](https://github.com/crossplane/crossplane/blob/91de251a2b9a5047a0043fa1baec43772e9292c2/cmd/crank/push.go) the only difference is that we are removing the `resource` specific subcommand.

phisco (approved):

phisco:
Do we want to deprecate `crank build|push`?

turkenh (approved):

lsviben:
> Do we want to deprecate `crank build|push`?

Makes sense. We can do in in a separate pr?

PR #4731, opened 2023-10-05
Support Composition environment via context

negz:

Fixes https://github.com/crossplane/crossplane/issues/4600
Relates to https://github.com/crossplane/crossplane/issues/3718

This commit adds a new field - context - to both `RunFunctionRequest` and `RunFunctionResponse`. Context is an arbitary set of key/value pairs that broadly represents the "context" in which the Function is running.

Context isn't desired state - Crossplane discards whatever context the last Function in the pipeline returns. It's also not strictly observed state, but it _could_ be used to supply ancillary observed state. Consider the Composition environment, or observed resources that are useful to the Composite Resource but not _part of it_. You could also imagine context being used to supply credentials to a Function.

Context is "bootstrapped" by Crossplane, then passed through the Function pipeline, like desired state, from `RunFunctionRequest` to `RunFunctionResponse`. This means context can originate either from Crossplane, or from a Function in the pipline.

I see context as having two key uses:

* Allowing Crossplane to inject things into the Function pipeline when we're not quite sure of the API (e.g `EnvironmentConfig`). In this way it's a bit like `metadata.annotations`. We can use it to pass data to the Function pipeline without having to actually update the protobuf types to accommodate alpha fields we may later make breaking changes to.

* Allowing Functions to share context with each other. Currently the only thing one Function can pass to another is desired state.

In this case, I've used context to pass the environment into the Function pipeline. The environment is still computed by Crossplane (for now). Currently only `function-patch-and-transform` has been updated to consume the environment. You can see how that works in the following PRs:

* https://github.com/crossplane-contrib/function-patch-and-transform/pull/17
* https://github.com/crossplane/function-sdk-go/pull/39

I have:


negz:
Here's an abridged version of [this Composition](https://github.com/upbound/platform-ref-aws/blob/c972e0cd2fe91ffdb0b181250832fae74e473b10/package/cluster/irsa/composition.yaml) from upbound/platform-ref-aws to illustrate how this looks with function-patch-and-transform:

```yaml
kind: Composition
metadata:
  name: xirsas.aws.platformref.upbound.io
spec:
  compositeTypeRef:
    apiVersion: aws.platformref.upbound.io/v1alpha1
    kind: XIRSA
  environment:
    environmentConfigs:
    - type: Selector
      selector:
        matchLabels:
          key: clusterRef
          valueFromFieldPath: spec.parameters.clusterRef.id
  resources:
  - name: irsa-role
    base:
      apiVersion: iam.aws.upbound.io/v1beta1
      kind: Role
      metadata:
        labels:
          resource: "Role"
    patches:
    - type: PatchSet
      patchSetName: Name
    - type: PatchSet
      patchSetName: providerConfigRef
    - type: PatchSet
      patchSetName: deletionPolicy
    - type: ToCompositeFieldPath
      toFieldPath: status.roleArn
      policy:
        fromFieldPath: Optional
    - type: ToCompositeFieldPath
      fromFieldPath: status.conditions
      toFieldPath: status.observed.role.conditions
      policy:
        fromFieldPath: Optional
    - fromFieldPath: spec.parameters.condition
      toFieldPath: metadata.annotations[condition]
    - fromFieldPath: spec.parameters.serviceAccount.namespace
      toFieldPath: metadata.annotations[serviceAccount/namespace]
    - fromFieldPath: spec.parameters.serviceAccount.name
      toFieldPath: metadata.annotations[serviceAccount/name]
      fromFieldPath: metadata.annotations[condition]
      toFieldPath: condition
      fromFieldPath: metadata.annotations[serviceAccount/namespace]
      toFieldPath: serviceAccount.namespace
      fromFieldPath: metadata.annotations[serviceAccount/name]
      toFieldPath: serviceAccount.name
    - type: CombineFromEnvironment
      toFieldPath: spec.forProvider.assumeRolePolicy
      combine:
        strategy: string
        variables:
        - fromFieldPath: oidcArn
        - fromFieldPath: condition
        - fromFieldPath: oidcHost
        - fromFieldPath: serviceAccount.namespace
        - fromFieldPath: serviceAccount.name
        string:
          fmt: |
            {
              "Version": "2012-10-17",
              "Statement": [
                {
                  "Effect": "Allow",
                  "Principal": {
                    "Federated": "%s"
                  },
                  "Action": "sts:AssumeRoleWithWebIdentity",
                  "Condition": {
                    "%s": {
                      "%s:sub": "system:serviceaccount:%s:%s"
                    }
                  }
                }
              ]
            }
```

phisco:
To be honest, I preferred having a known field in which to inject the current environment and potentially being able to explicitly add other stuff to it down the line as in #4632, instead of mimicking the Go context with arbitrary keys used to inject stuff into it. I guess it's not to commit to the environment, but I preferred keeping calling it environment for continuity and because it feels like the context is more a Go concept, while environment feels like a more broadly adopted, e.g. bash, LISP.

apis/apiextensions/fn/proto/v1beta1/run_function.proto
```diff
@@ -54,6 +54,12 @@ message RunFunctionRequest {
   // Optional input specific to this Function invocation. A JSON representation
   // of the 'input' block of the relevant entry in a Composition's pipeline.
   optional google.protobuf.Struct input = 4;
+
+  // Optional context. Crossplane may pass arbitary contextual information to a
+  // Function. A Function may also return context in its RunFunctionResponse,
+  // and that context will be passed to subsequent Functions. Crossplane
+  // discards all context returned by the last Function in the pipeline.
+  map<string, google.protobuf.Value> context = 5;
```
negz, line 62:
```suggestion
  optional google.protobuf.Struct context = 5;
```

It occurs to me in retrospect that these are pretty much equivalent, given that a `google.protobuf.Struct` _is_ a map of string key fields to `google.protobuf.Value`. So this is semantically equivalent to what @phisco did in https://github.com/crossplane/crossplane/pull/4632.

phisco: Which is what we were discussing here https://github.com/crossplane/crossplane/pull/4632#discussion_r1327686709 😅 however I ended up not making it just a map, but adding an explicit data field to the environment for the reasons I was mentioning above.

negz:
> You could also imagine context being used to supply credentials to a Function.

Just kidding, that's probably not a good idea. We want to pass credentials to _a_ Function, not all Functions in the pipeline.

negz:
> To be honest, I preferred having a known field in which to inject the current environment and potentially being able to explicitly add other stuff to it down the line as in https://github.com/crossplane/crossplane/pull/4632, instead of mimicking the Go context with arbitrary keys used to inject stuff into it. I guess it's not to commit to the environment, but I preferred keeping calling it environment for continuity and because it feels like the context is more a Go concept, while environment feels like a more broadly adopted, e.g. bash, LISP.

@phisco In retrospect, I think I've proposed effectively the same thing you did in https://github.com/crossplane/crossplane/pull/4632, but with a slightly different API. I suspect this is because we have slightly different mental models for the same thing.

When I think what you might pass to a Function pipeline, I think of:

* "The environment", as it exists today - an arbitrary JSON object made by merging EnvConfigs (+ defaults).
* "Friend resources" - observed resources that are useful to "do composition", but not actually composed by the XR.
* ~Function credentials~ (in retrospect, not a good idea per comment above).
* Other stuff we haven't thought of yet - in particular stuff Function N might want to pass to Function N+1.

I don't feel strongly about the name of the top-level field in the request/response (context vs environment). I do wonder whether it makes sense for what today we consider to be "the environment" (i.e. merged EnvConfigs) to be a key under the top-level field rather than being the top-level object itself, though. This is mostly because I worry about collisions. If Function N wants to use "the environment" to pass context to Functions N+1..., should it use the same "namespace" (top-level keys in the environment `google.protobuf.Struct`) as the object Crossplane has constructed by merging together EnvConfigs?

apis/apiextensions/fn/proto/v1beta1/run_function.proto
negz: Haha I somehow completely forgot about that conversation. My bad.

I think we're converging on something pretty close to what you already have in #4632 so I'll close this PR. Should have re-read it before I started working on this one. 🙃

negz:
Per https://github.com/crossplane/crossplane/pull/4731#discussion_r1346866400 it turns out we already covered a bunch of this ground in #4632 and I just forgot about it. 😬 I'll close this and we can discuss there.

phisco:
>  do wonder whether it makes sense for what today we consider to be "the environment" (i.e. merged EnvConfigs) to be a key under the top-level field rather than being the top-level object itself, though. This is mostly because I worry about collisions. If Function N wants to use "the environment" to pass context to Functions N+1..., should it use the same "namespace" (top-level keys in the environment google.protobuf.Struct) as the object Crossplane has constructed by merging together EnvConfigs?

That's why I was passing that in the Environment under a data field, leaving room for future additions

Issue #4251, opened 2023-06-22
Deleting a Provider before its MRs results in stuck MRs


negz:
### What happened?

While working on https://github.com/crossplane/crossplane/pull/4187 I was reminded that you must ensure all of a provider's MRs are completely deleted before uninstalling the provider. If you don't:

* Any MRs will be left in an orphaned, unreconciled, un-deletable state.
* The orphaned MRs will prevent the CRD that defines them being removed.
* The CRD that defines them will prevent the provider being reinstalled.

I believe this happens because:

* The `Provider.pkg` and its revision are deleted immediately, relying on Kubernetes garbage collection to delete the MR CRDs.
* An MR CRD can't be deleted while an MR (CR) exists. The CRD has `metadata.deletionTimestamp` set, but deletion is blocked by the `customresourcecleanup.apiextensions.k8s.io` finalizer.
* The `Provider.pkg` and its revisions being gone means the provider deployment no longer exists. This means you now can't delete the MR - deletion will be blocked by the `finalizer.managedresource.crossplane.io` finalizer.

I believe if you get into this state your only option is to do finalizer surgery on the MR(s) that are blocking the deletion (and to manually delete any corresponding external resources).

### How can we reproduce it?

1. Create a provider:

```yaml
apiVersion: pkg.crossplane.io/v1
kind: Provider
metadata:
  name: provider-nop
spec:
  package: xpkg.upbound.io/crossplane-contrib/provider-nop:v0.2.0
  ignoreCrossplaneConstraints: true
```

2. Create an MR:

```yaml
apiVersion: nop.crossplane.io/v1alpha1
kind: NopResource
metadata:
  name: example-resource
spec:
 forProvider:
   conditionAfter:
   - conditionType: Ready
     conditionStatus: "False"
     conditionReason: "Creating"
     time: 0s
   - conditionType: Ready
     conditionStatus: "True"
     conditionReason: "Available"
     time: 10s
```

3. Delete the provider
4. Try to delete the MR

### What environment did it happen in?
Crossplane version: b65fa47e5f270b3054840cce1e0c4fcc9e96b7b7

negz:
Presumably the only way to fix this is to prevent the provider being deleted until all of its MRs are deleted. Given that MR deletion blocks MR CRD deletion, potentially we could have MR CRD deletion in turn block Provider deletion.

As to how to block deletion, I think the options are:

1. Wait for https://github.com/kubernetes/enhancements/pull/2840
[...]

PR #4733, opened 2023-10-05
Speed up e2e

sttts:

No useless wait time, faster polling and output of durations.

Note: this is still not good. We must parallelize. Serial e2e tests won't scale.

I have:

### DIFF
--- a/test/e2e/funcs/feature.go
+++ b/test/e2e/funcs/feature.go
@@ -63,6 +63,9 @@ import (
 	"github.com/crossplane/crossplane-runtime/pkg/resource/unstructured/composite"
 )
 
+// DefaultPollInterval is the suggested poll interval for wait.For.
+const DefaultPollInterval = time.Millisecond * 500
+
 // AllOf runs the supplied functions in order.
 func AllOf(fns ...features.Func) features.Func {
 	return func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
@@ -100,7 +103,7 @@ func DeploymentBecomesAvailableWithin(d time.Duration, namespace, name string) f
 	return func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
 		dp := &appsv1.Deployment{ObjectMeta: metav1.ObjectMeta{Namespace: namespace, Name: name}}
 		t.Logf("Waiting %s for deployment %s/%s to become Available...", d, dp.GetNamespace(), dp.GetName())
-		if err := wait.For(conditions.New(c.Client().Resources()).DeploymentConditionMatch(dp, appsv1.DeploymentAvailable, corev1.ConditionTrue), wait.WithTimeout(d)); err != nil {
+		if err := wait.For(conditions.New(c.Client().Resources()).DeploymentConditionMatch(dp, appsv1.DeploymentAvailable, corev1.ConditionTrue), wait.WithTimeout(d), wait.WithInterval(DefaultPollInterval)); err != nil {
 			t.Fatal(err)
 			return ctx
 		}
@@ -127,7 +130,7 @@ func ResourcesCreatedWithin(d time.Duration, dir, pattern string) features.Func 
 			t.Logf("Waiting %s for %s to exist...", d, identifier(u))
 		}
 
-		if err := wait.For(conditions.New(c.Client().Resources()).ResourcesFound(list), wait.WithTimeout(d)); err != nil {
+		if err := wait.For(conditions.New(c.Client().Resources()).ResourcesFound(list), wait.WithTimeout(d), wait.WithInterval(DefaultPollInterval)); err != nil {
 			t.Errorf("resources did not exist: %v", err)
 			return ctx
 		}
@@ -143,7 +146,7 @@ func ResourceCreatedWithin(d time.Duration, o k8s.Object) features.Func {
 	return func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
 		t.Logf("Waiting %s for %s to be created...", d, identifier(o))
 
-		if err := wait.For(conditions.New(c.Client().Resources()).ResourceMatch(o, func(object k8s.Object) bool { return true }), wait.WithTimeout(d)); err != nil {
+		if err := wait.For(conditions.New(c.Client().Resources()).ResourceMatch(o, func(object k8s.Object) bool { return true }), wait.WithTimeout(d), wait.WithInterval(DefaultPollInterval)); err != nil {
 			t.Errorf("resource %s did not exist: %v", identifier(o), err)
 			return ctx
 		}
@@ -171,7 +174,7 @@ func ResourcesDeletedWithin(d time.Duration, dir, pattern string) features.Func 
 			t.Logf("Waiting %s for %s to be deleted...", d, identifier(u))
 		}
 
-		if err := wait.For(conditions.New(c.Client().Resources()).ResourcesDeleted(list), wait.WithTimeout(d)); err != nil {
+		if err := wait.For(conditions.New(c.Client().Resources()).ResourcesDeleted(list), wait.WithTimeout(d), wait.WithInterval(DefaultPollInterval)); err != nil {
 			objs := itemsToObjects(list.Items)
 			related, _ := RelatedObjects(ctx, c.Client().RESTConfig(), objs...)
 			events := valueOrError(eventString(ctx, c.Client().RESTConfig(), append(objs, related...)...))
@@ -191,7 +194,7 @@ func ResourceDeletedWithin(d time.Duration, o k8s.Object) features.Func {
 	return func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
 		t.Logf("Waiting %s for %s to be deleted...", d, identifier(o))
 
-		if err := wait.For(conditions.New(c.Client().Resources()).ResourceDeleted(o), wait.WithTimeout(d)); err != nil {
+		if err := wait.For(conditions.New(c.Client().Resources()).ResourceDeleted(o), wait.WithTimeout(d), wait.WithInterval(DefaultPollInterval)); err != nil {
 			t.Errorf("resource %s not deleted: %v", identifier(o), err)
 			return ctx
 		}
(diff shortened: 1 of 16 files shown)


pedjak (approved):

test/e2e/funcs/feature.go
```diff
@@ -57,6 +57,9 @@ import (
 	"github.com/crossplane/crossplane-runtime/pkg/resource/unstructured/composite"
 )
 
+// DefaultPollInterval is the suggested poll interval for wait.For.
+const DefaultPollInterval = time.Millisecond * 500
```
pedjak, line 61: not sure if this is going to speed up anything, because the default poll interval in e2e framework is 5s. For sure, it is going to produce more requests against api server.

sttts: 5s is far too much. The apiserver can serve thousands of requests per second. Why waste (developer) time?

sttts:
> All e2e cases failed on the first run, which is not something I saw often, so I'd say it's definitely not helping with e2e flakiness

I would call that success ;-) Flakes happening most of the time are good flakes, easier to fix.

phisco:
> I would call that success ;-) Flakes happening most of the time are good flakes, easier to fix.

Fair point!

test/e2e/funcs/feature.go
bobh66: 500ms == 0.5s

phisco (approved):

phisco:
I'll merge this as soon as I see the E2Es passing, ok @sttts?

sttts:
> I'll merge this as soon as I see the E2Es passing, ok @sttts?

@phisco to be honest, I am a little out of ideas what to do here. With the first commit, e2e tests fail. How can twice-per-second polling cause this instability?

phisco:
@sttts once we merge https://github.com/crossplane/crossplane/pull/4750 we should have enough information to understand what's going on, and faster E2Es already on failure as I configured them to fail immediately on failure.

phisco:
The only failure I can see is related to an issue we are already aware of: https://github.com/crossplane/crossplane/actions/runs/6479074672/job/17591938889?pr=4733#step:12:221

But it's sensibly reducing the time needed for the e2es to pass, from:
image
to
image

So I'll run the tests one more time to see if there is any issue and then merge it, sounds good @sttts?

phisco:
second run 3/4 hit a known issue, the 4th one resulted in a panic for unrelated issues, probably caused by the shorter times, running a third time.

here the stacktrace for future reference:
```
--- FAIL: TestEnvironmentConfigDefault (11.76s)
    --- FAIL: TestEnvironmentConfigDefault/TestEnvironmentConfigDefault (11.76s)
        --- PASS: TestEnvironmentConfigDefault/TestEnvironmentConfigDefault/CreateClaim (0.51s)
        --- FAIL: TestEnvironmentConfigDefault/TestEnvironmentConfigDefault/MRHasAnnotation (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0xba3729]

goroutine 99 [running]:
testing.tRunner.func1.2({0x164bc20, 0x2615190})
	testing/testing.go:1545 +0x238
testing.tRunner.func1()
	testing/testing.go:1548 +0x397
panic({0x164bc20?, 0x2615190?})
	runtime/panic.go:914 +0x21f
k8s.io/api/core/v1.(*ObjectReference).GroupVersionKind(0xc0004c6010?)
	k8s.io/api@v0.28.2/core/v1/objectreference.go:30 +0x29
github.com/crossplane/crossplane/test/e2e.TestEnvironmentConfigDefault.ComposedResourcesOfClaimHaveFieldValueWithin.func15({0x1a92810, 0xc0002e2fc0}, 0xc0001244e0, 0x0?)
	github.com/crossplane/crossplane/test/e2e/funcs/feature.go:509 +0x2ee
sigs.k8s.io/e2e-framework/pkg/env.(*testEnv).executeSteps(0xc0002e2930, {0x1a92810?, 0xc0002e2fc0?}, 0x21cd8c8?, {0xc0005a5f50?, 0x1, 0x52f01c?})
	sigs.k8s.io/e2e-framework@v0.3.0/pkg/env/env.go:428 +0x96
sigs.k8s.io/e2e-framework/pkg/env.(*testEnv).processTestFeature.(*testEnv).execFeature.func1.1(0xc0001244e0)
	sigs.k8s.io/e2e-framework@v0.3.0/pkg/env/env.go:466 +0xaa
testing.tRunner(0xc0001244e0, 0xc000502ab0)
	testing/testing.go:1595 +0xff
created by testing.(*T).Run in goroutine 38
	testing/testing.go:1648 +0x3ad
09:42:44 [FAIL]
```

phisco:
third try 4/4 hitting the known issue I was referring to above, e.g. from https://github.com/crossplane/crossplane/actions/runs/6479074672/job/17597537026?pr=4733#step:12:214 : 
```
    feature.go:241: - Provider provider-nop: Healthy=False Reason=UnhealthyPackageRevision: cannot establish control of object: nopresources.nop.crossplane.io is already controlled by ProviderRevision provider-nop-552a394a8acc (UID 63b31ac3-5eba-47ed-9b3c-c19ee1445885) (2023-10-11 10:05:07 +0000 UTC)
```

The issue is already present on `master` and after some digging it looks like it's due to a leftover from `provider-nop`'s deletion, as when we delete the Provider we leave the ProviderRevision and CRDs to be garbage collected, but before the CRDs are deleted, we move on to the next test, which most of the times creates a new `provider-nop` Provider and ProviderRevision, which then tries to take ownership of the CRDs, leading to the above error.
I think the solution to this would be to delete everything with a `Foreground` propagation policy, so that we are sure that once a test is finished the cleanup is complete, but I have to test it further.

turkenh:
> The issue is already present on `master` and after some digging it looks like it's due to a leftover from `provider-nop`'s deletion, as when we delete the Provider we leave the ProviderRevision and CRDs to be garbage collected, but before the CRDs are deleted, we move on to the next test, which most of the times creates a new `provider-nop` Provider and ProviderRevision, which then tries to take ownership of the CRDs, leading to the above error.

I am also having the same problem with [my PR](https://github.com/crossplane/crossplane/pull/4767) even reproducible locally. What I am kind of surprised is, for the specific test case, we are giving 2mins which should be enough for the Garbage collector to do its job and the package manager to install the new one. It could be a result of an accumulation of back-offs though. 
What I am wondering is whether we have a test case that could uninstall the provider before its MR is cleaned up. Imagine we triggered the deletion of MR but without waiting for it, we uninstall the provider. In this case, it would be expected to fail sporadically since sometimes the provider controller could finalize deletion of the resource but sometimes not. This might be a more reasonable explanation for such an issue.

pedjak:
> What I am wondering is whether we have a test case that could be uninstalling the provider before its MR is cleaned up. 

I would avoid uninstalling provider after each test. A provider should be installed if a test setup requires, but then we should not perform any cleanup, if not explicitly in a focus of some test.

turkenh:
> What I am wondering is whether we have a test case that could be uninstalling the provider before its MR is cleaned up. Imagine we triggered the deletion of MR, but without waiting for it to complete, we uninstall the provider. In this case, it would be expected to fail sporadically since sometimes the provider controller could finalize deletion of the resource but sometimes not. This might be a more reasonable explanation for such an issue.

I think I can confirm this behavior. Did run `make e2e E2E_TEST_FLAGS="-test.v --test-suite base -destroy-kind-cluster=false -test.failfast"` and when it failed what I am finding is:

```
   feature.go:220: Waiting 2m0s for Provider provider-nop to become HealthyPackageRevision, ActivePackageRevision...
    feature.go:233: - Provider provider-nop: Healthy=Unknown Reason=UnknownPackageRevisionHealth: "" (2023-10-11 13:55:29 +0300 +03)
    feature.go:233: - Provider provider-nop: Healthy=False Reason=UnhealthyPackageRevision: cannot establish control of object: nopresources.nop.crossplane.io is already controlled by ProviderRevision provider-nop-552a394a8acc (UID c6496a34-7e94-46d8-a516-5a5c5abea660) (2023-10-11 13:55:45 +0300 +03)
W1011 13:57:38.736997   75069 warnings.go:70] v1 ComponentStatus is deprecated in v1.19+
W1011 13:57:51.082502   75069 warnings.go:70] ControllerConfig.pkg.crossplane.io/v1alpha1 is deprecated. It is safe to use, but no new features will be added. Future replacement design: https://github.com/crossplane/crossplane/blob/master/design/one-pager-package-runtime-config.md
    feature.go:250: resources did not have desired conditions: HealthyPackageRevision, ActivePackageRevision: client rate limiter Wait returned an error: context deadline exceeded:
``` 

and 

```
❯ kubectl get providers
No resources found
❯ kubectl get providerrevisions.pkg.crossplane.io
No resources found
❯ kubectl get crd | grep nop
nopresources.nop.crossplane.io                             2023-10-11T10:55:01Z
❯ kubectl get nopresources.nop.crossplane.io
NAME                                          READY   SYNCED   AGE
apiextensions-composition-pandt-q5tz4-5k556   True    True     18m
❯ kubectl -n crossplane-system get pods | grep nop
```

So, seems that the provider is uninstalled before its MR is successfully deleted, and this blocks the CRD owned by previous provider to be garbage collected.

phisco:
@turkenh, strange you have no provider/providerrevision though, might be because you run the test with just `-test.failfas`t instead of  also `-fail-fast`, which prevents even the teardown from happening. (I know, it's confusing, I was surprised too when I understood it the other day when I added both flags to ci...)

turkenh:
> @turkenh, strange you have no provider/providerrevision though, might be because you run the test with just `-test.failfas`t instead of also `-fail-fast`, which prevents even the teardown from happening. (I know, it's confusing, I was surprised too when I understood it the other day when I added both flags to ci...)

Good call out. 

I did run it again and this is the state it ended up:

```
❯ kubectl get providers
NAME           INSTALLED   HEALTHY   PACKAGE                                                  AGE
provider-nop   True        False     xpkg.upbound.io/crossplane-contrib/provider-nop:v0.2.0   3m

❯ kubectl get providerrevisions.pkg.crossplane.io
NAME                        HEALTHY   REVISION   IMAGE                                                    STATE    DEP-FOUND   DEP-INSTALLED   AGE
provider-nop-552a394a8acc   False     1          xpkg.upbound.io/crossplane-contrib/provider-nop:v0.2.0   Active                               3m10s

❯ kubectl get nopresources.nop.
NAME                                          READY   SYNCED   AGE
apiextensions-composition-pandt-7fmlv-5llqb   True    True     3m40s

❯ kubectl get nopresources.nop apiextensions-composition-pandt-7fmlv-5llqb -o yaml

kind: NopResource
metadata:
  ...
  deletionGracePeriodSeconds: 0
  deletionTimestamp: "2023-10-11T12:19:38Z"
  finalizers:
  ...
```

So, yes, not having the provider was due to the teardown, but the reason for failing installation is that the provider was [uninstalled](https://github.com/crossplane/crossplane/issues/4251) before its MRs were cleaned up.

> The CRD that defines them will prevent the provider being reinstalled.

phisco:
Looking good, from:
image
to:
image

phisco:
3 runs all green, going to merge this 🎉 thanks @sttts

Issue #4645, opened 2023-09-19
Claim name can be max 57 chars long


pedjak:
### What happened?

If a claim name is longer than 63 characters, the corresponding XR cannot be created during the claim reconcile loop.

### How can we reproduce it?

* install provider nop

```yaml
apiVersion: pkg.crossplane.io/v1
kind: Provider
metadata:
  name: provider-nop
spec:
  package: xpkg.upbound.io/crossplane-contrib/provider-nop:v0.2.0
  ignoreCrossplaneConstraints: true
```

* add XRD

```yaml
apiVersion: apiextensions.crossplane.io/v1
kind: CompositeResourceDefinition
metadata:
  name: xnopresources.nop.example.org
spec:
  group: nop.example.org
  names:
    kind: XNopResource
    plural: xnopresources
  claimNames:
    kind: NopResource
    plural: nopresources
  connectionSecretKeys:
  - test
  versions:
  - name: v1alpha1
    served: true
    referenceable: true
    schema:
     openAPIV3Schema:
       type: object
       properties:
        spec:
          type: object
          properties:
            coolField:
              type: string
          required:
          - coolField
```

* add composition

```yaml
apiVersion: apiextensions.crossplane.io/v1
kind: Composition
metadata:
  name: xnopresources.nop.example.org
spec:
  compositeTypeRef:
    apiVersion: nop.example.org/v1alpha1
    kind: XNopResource
  resources:
  - name: nop-resource-1
    base:
     apiVersion: nop.crossplane.io/v1alpha1
     kind: NopResource
     spec:
      forProvider:
        conditionAfter:
        - conditionType: Ready
          conditionStatus: "False"
          time: 0s
        - conditionType: Ready
          conditionStatus: "True"
          time: 10s
```

* try to create the following claim

```yaml
apiVersion: nop.example.org/v1alpha1
kind: NopResource
metadata:
  namespace: default
  name: res-1234567890-1234567890-1234567890-1234567890-1234567890-12345
spec:
  coolField: "I'm cool!"
```

The claim gets created, but remains unsynced due to:

```yaml
status:
  conditions:
  - lastTransitionTime: "2023-09-19T18:31:45Z"
    message: 'cannot configure composite resource: cannot use dry-run create to name
      composite resource: XNopResource.nop.example.org "res-1234567890-1234567890-1234567890-1234567890-1234567890d5df8"
[...]

PR #4742, opened 2023-10-06
Derive XR name from claim name and its UUID

### LINKED ISSUE #4651: Take API server out of XR name generation path

After a claim is deployed to the cluster, its counterpart composite needs to be created. Its name is derived from the claim's name, by adding a suffix. The final name is obtained sending a dry-run create request to API server. 

[The Kubernetes docs](https://kubernetes.io/docs/reference/using-api/api-concepts/#generated-values) is strongly suggesting not to rely on such generated values - the name could be taken at the moment when the real request is submitted.

Claim controller can generate the composite name without contacting API server, and hence speeding up the reconciliation. Claim's UID poses enough entropy, and it would be enough to take some part of it (last `N` chars, e.g. `N > 6`) and append it to the original claim name.

pedjak:

There is no need to make a round-trip to k8s api server
to get XR name, we have enough data on the controller side
to construct an unique XR name, by using the convention:

<CLAIM_NAME>-<CLAIM_UUID_LAST_64_BITS>

Change details:
* `APIDryRunCompositeConfigurator` renamed to `APICompositeConfigurator`
to reflect that k8s api server is not contacted anymore
* removed `NewAPIDryRunCompositeConfigurator` because
`APIDryRunCompositeConfigurator` has no fields anymore
* updated `configurator_test.go`

Fixes #4651 

I have:


go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/crossplane/crossplane
 
 go 1.20
 
+replace github.com/crossplane/crossplane-runtime => github.com/pedjak/crossplane-runtime v0.0.0-20231004110250-0b3151613115
```
pedjak, line 5: this PR depends on https://github.com/crossplane/crossplane-runtime/pull/552 and it needs to merged first

internal/controller/apiextensions/claim/configurator.go
```diff
@@ -136,13 +128,8 @@ func (c *APIDryRunCompositeConfigurator) Configure(ctx context.Context, cm resou
...
 	if !meta.WasCreated(cp) {
-		// The API server returns an available name derived from
-		// generateName when we perform a dry-run create. This name is
-		// likely (but not guaranteed) to be available when we create
-		// the composite resource. If the API server generates a name
-		// that is unavailable it will return a 500 ServerTimeout error.
-		cp.SetGenerateName(fmt.Sprintf("%s-", cm.GetName()))
-		return errors.Wrap(c.client.Create(ctx, cp, client.DryRunAll), errName)
+		// XR name follows the convention <claim_name>-<claim_uid_suffix>
+		cp.SetName(fmt.Sprintf("%s-%s", cm.GetName(), cm.GetUID()[19:]))
```
sttts, line 132: curious, why `19:`? Why not `:8` or so?

sttts: what happens if the partial UID conflicts? Can it realistically? What's the propability?

sttts: wondering whether the format of the uid is specified in kube. Probably not. Could be different tomorrow such that `19:` breaks.

go.mod
sttts: why does it?

internal/controller/apiextensions/claim/configurator.go
pedjak: > curious, why 19:? Why not :8 or so?

UUIDs are 128 bits long, so I wanted to take half them (64) to have enough entropy, i.e. to avoid collisions. Ofcourse, we can take more bits (or less), but that would then have impact on collision and on max length of claim name.

> wondering whether the format of the uid is specified in kube. Probably not. Could be different tomorrow such that 19: breaks.

Kubernetes doc: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#uids and more https://en.wikipedia.org/wiki/Universally_unique_identifier#Encoding

It is a well know standard. Not sure how realistic is that kubernetes switches to something else. They could switch to different UUID encoding, but if we want to guard against that, we could [parse the string representation](https://pkg.go.dev/github.com/google/uuid#Parse) and then encode again the suffix we would like to take according to some schema.

> what happens if the partial UID conflicts? Can it realistically? What's the propability?

Kuberenets uses Version 4 UUID, which are randomly generated, and the probability for collision is very low:

https://en.wikipedia.org/wiki/Universally_unique_identifier#Collisions

Even if we take 64 bits, it is reasonably low that I think we should not add here any logic that should try to handle it.

sttts: just put that into the comment. It also matters here that conflict only happens for equal claim names (cross namespaces). I.e. with 10000ish namespaces and equal claim names:

Bildschirmfoto 2023-10-06 um 16 58 03

2^-12 should be safe.

go.mod
sttts: ack

internal/controller/apiextensions/claim/configurator.go
pedjak: @sttts I have made the code agnostic to the encoding format used by the k8s.

internal/controller/apiextensions/claim/configurator.go
```diff
@@ -48,26 +50,18 @@ const (
...
-type APIDryRunCompositeConfigurator struct {
-	client client.Client
-}
-
-// NewAPIDryRunCompositeConfigurator returns a Configurator of composite
-// resources that may perform a dry-run create against an API server in order to
-// name and validate the configured resource.
-func NewAPIDryRunCompositeConfigurator(c client.Client) *APIDryRunCompositeConfigurator {
-	return &APIDryRunCompositeConfigurator{client: c}
+type APICompositeConfigurator struct {
```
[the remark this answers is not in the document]
pedjak: Thanks for the suggestion, removed.

go.mod
sttts: Runtime PR merged.

pedjak: updated to latest crossplane

internal/controller/apiextensions/claim/configurator.go
```diff
@@ -134,13 +122,18 @@ func (c *APIDryRunCompositeConfigurator) Configure(ctx context.Context, cm resou
...
+		// For robustness, try to parse it first
+		id, err := uuid.Parse(string(cm.GetUID()))
+		if err != nil {
+			return err
+		}
+		// Construct XR name following the convention <claim_name>-<claim_uid_last_64_bits>
+		// risk for collision is super low (2^-12 for 10k claims with the same name)
+		// https://en.wikipedia.org/wiki/Universally_unique_identifier#Collisions
+		// therefore we do not add any logic that would handle it.
+		cp.SetName(fmt.Sprintf("%s-%s", cm.GetName(), hex.EncodeToString(id[8:])))
```
sttts, line 136: Hex encoding means 4 bits per character. Hence, we are appending a 64/4=16 long random string, plus the `-`. I.e. we are losing 17 characters out of 63 allowed. Before this change, with Kube doing the name generation we only lost 6 characters. This breaks claims with long names (> 46 chars; before 57 were fine).

In other words: this change is a breaking API change for Crossplane.

sttts: There are libraries for base62 encoding (2*26+ 10). With that we nearly squeeze 6 bits into one character, leading to 11ish characters for the suffix.

Assuming we use a 5 char suffix and base62, what would that mean for the math?

sttts: This made my day: https://x.com/the_sttts/status/1711702505590063478?s=20

sttts: Actually, names must be lower-case I guess. So we have only base36 at best.

jbw976 (commented):
@pedjak does the format of the names change majorly at all here, or is it mostly doing the same thing but avoiding the API server? Can you add examples of what the names used to look like vs. what they look like now?

Just to be sure, this doesn't affect existing XRs out in the wild, right? 🙇‍♂️

internal/controller/apiextensions/claim/configurator.go
pedjak: > Actually, names must be lower-case I guess. So we have only base36 at best.

Yes, and with base36 we save just 3 chars.

If we wanna take just 48bits instead of 64, then with 1000 claims, the probability for collision is 10<sup>-7</sup> and then using base36 encoding can get us to 10 char suffix.

> In other words: this change is a breaking API change for Crossplane.

It is already reported in https://github.com/crossplane/crossplane/issues/4645, not sure if we advertise, nor check name size limitations right now. Moving to annotations is going to solve this problem.

@sttts let me know if we should go with base36 encoding here.

sttts: How hard is it to optimistically choose a suffix in kube style (maybe even using their helper), go with it, and in case the XR exists but belongs to some other claim we retry by choosing another suffix and store it in the claim ?

sttts: Still a breaking change.

pedjak: we do check not report back to use what the max claim name length.

PR #4744, opened 2023-10-06
Beta support for Package Runtime Config

### LINKED ISSUE #4690: Consider using specific package types

The package manager really has two core reconcilers - a package reconciler, and a package revision reconciler. With Crossplane v1.14 these reconcilers are used to reconcile three package types and their revisions: Provider, Configuration, and Function.

In order for two reconcilers to support these three different package types the code makes heavy use of interfaces, with methods like `GetPackage()`, `GetControllerConfigRef()`, etc. Much of the reconciler code uses these interfaces to be able to reconcile generic packages, with anything specific to a particular package or revision type pushed into optional hooks. These interfaces are primarily getters and setters, backed by shared structs like `PackageSpec` and `PackageStatus`.

When the package manager was first designed, I believe these shared structs and interfaces only exposed data that truly was relevant to all package types. In the time since we've added a few things that are only relevant to certain package types. For example:

* `CommonLabels` in `PackageSpec` - Only relevant to Configurations.
* `ControllerConfigReference` in `PackageRevisionSpec` - Only relevant to Providers. This one is only exposed in the provider-specific `ProviderSpec`, not the shared `PackageSpec`, despite existing in the shared `PackageRevisionSpec`.
* Various TLS certificate settings, e.g. `TLSServerSecretName` in `PackageSpec`. Not relevant to Configurations.

The result of this is potential confusion for users. These fields show up in the schema of types, indicating that they could be configured when in fact they cannot. I believe they were most likely added to the shared types by mistake, though it could have been due to a limitation of the reconciler design.

Ideally we would not surface fields on types that will be ignored. So for example `CommonLabels` would appear in the `ConfigurationSpec` struct, not the `PackageSpec` that it embeds. Technically removing a field would be a breaking API change, but I'm not too concerned about this given the fields don't actually work today in the contexts where they'd be removed.

If removing the fields is too hard for now, we could add least update the GoDoc strings for the fields (and thus the generated API docs) to explicitly mention that they only work for certain types and not others (e.g. only for Configurations, only for Providers, etc).

### LINKED ISSUE #4699: Implement beta support for `RuntimeConfig`

We marked `ControllerConfig` deprecated in v1.11 but we haven't yet implemented a replacement. We have a design for a replacement that was merged in https://github.com/crossplane/crossplane/pull/4322. The idea is that we'd implement a new type called `RuntimeConfig`. This would be a successor to `ControllerConfig` - it's pretty similar but a bit more flexible.

I think we should implement this and ship `RuntimeConfig` at `v1beta1`. It's essentially the beta iteration of `ControllerConfig` - we just changed the name because it will be used for Functions too, and a Function is not a (Kubernetes) controller.

turkenh:

This PR implements Package Runtime Config (Fixes #4699) as proposed in the [design doc](https://github.com/crossplane/crossplane/blob/2c5e7f07ba9e3d83d1c85169bbde685de8514ab8/design/one-pager-package-runtime-config.md) with the following exception:

Instead of initializing and maintaining our defaults like `replicas: 1` or pod and runtime container `securityContext` in the `DeploymentRuntimeConfig` named `default`, it only initializes an empty `DeploymentRuntimeConfig` named as `default` _only if it does not exist_. It was discussed in [the issue](https://github.com/crossplane/crossplane/issues/4699#issuecomment-1748369504) in details but this is primarily to avoid the following:

- Creating a new `DeploymentRuntimeConfig` would require duplicating the values from the `default` one, assuming the user won't want to lose them. If you only intend to include additional labels, you would need to duplicate the entire default configuration and append labels to the newly created one.
- From now on, the user would need to maintain the defaults in case Crossplane wants to change or extend them in the upcoming releases.
- Since the user could change the `default` config in time, in the new releases, Crossplane would either overwrite the `default` `DeploymentRuntimeConfig` to reflect changes in defaulting or leave the `default` as is, which would mean Crossplane losing control over defaults.

Considering the above 3 being a degradation over User Experience and Crossplane controlling the defaulting behavior with new releases, I ended up changing it as follows:

- Crossplane will create an empty `default` `DeploymentRuntimeConfig` if it does not exist.
- Crossplane will set defaults only if `DeploymentRuntimeConfig` hasn't set them (i.e. unopinionated). For example, if the deployment `replicas` is not provided in the config, Crossplane will set the value as `1`.
- Crossplane is opinionated about some configurations and will override even if they were provided in the config. 

This way, the `default` `DeploymentRuntimeConfig` could also be owned by the user without any maintenance burden.

This PR also:
- Refactors existing hooks and manifest generation with a ManifestBuilder.
- Introduces `PackageWithRuntime` and `PackageRevisionWithRuntime` interfaces.
- Uses specific package types (Fixes #4690).
- Does some clean up including removing the unused `controllerRef` from PackageRevisionStatus.

See [the manifest](https://github.com/crossplane/crossplane/pull/4744/files#diff-461c203fbc0e0103ad1b20519e525fd491602181aad8bb0779d94cdc0a49282d) in the e2e test to get a sense on how `DeploymentRuntimeConfig` could be customized but still operate properly.

I have:

- [x] Opened a PR updating the [docs], if necessary => https://github.com/crossplane/docs/pull/570


internal/controller/pkg/revision/runtime.go
```diff
@@ -0,0 +1,336 @@
...
+	}
+
+	rcRef := pwr.GetRuntimeConfigRef()
+	if rcRef == nil {
+		return nil, errors.New(errNoRuntimeConfig)
+	}
+
+	rc := &v1beta1.DeploymentRuntimeConfig{}
+	if err := client.Get(ctx, types.NamespacedName{Name: rcRef.Name}, rc); err != nil {
+		return nil, errors.Wrap(err, errGetControllerConfig)
```
negz, line 129:
```suggestion
		return nil, errors.Wrap(err, errGetRuntimeConfig)
```

cmd/crossplane/core/core.go
```diff
@@ -273,13 +275,25 @@ func (c *startCommand) Run(s *runtime.Scheme, log logging.Logger) error { //noli
...
+	}
+
 	po := pkgcontroller.Options{
 		Options:         o,
 		Cache:           xpkg.NewFsPackageCache(c.CacheDir, afero.NewOsFs()),
 		Namespace:       c.Namespace,
 		ServiceAccount:  c.ServiceAccount,
 		DefaultRegistry: c.Registry,
 		FetcherOptions:  []xpkg.FetcherOpt{xpkg.WithUserAgent(c.UserAgent)},
+		PackageRuntime:  pr,
```
negz, line 296: Should (can?) we put package runtime support behind a beta feature flag?

internal/controller/pkg/revision/runtime_function.go
```diff
@@ -0,0 +1,178 @@
...
+// Pre performs operations meant to happen before establishing objects.
+func (h *FunctionHooks) Pre(ctx context.Context, _ runtime.Object, pr v1.PackageRevisionWithRuntime, manifests ManifestBuilder) error {
+	// TODO(ezgidemirel): update any status fields relevant to package revisions.
+
+	if pr.GetDesiredState() != v1.PackageRevisionActive {
+		return nil
+	}
+
+	// Ensure Prerequisites
+	svc := manifests.Service(functionServiceOverrides()...)
```
negz, line 76:
```suggestion
	svc := build.Service(functionServiceOverrides()...)
```

Just a naming suggestion. Or we could keep `manifests` and change the method to be `manifests.BuildService`.

internal/controller/pkg/revision/runtime_function.go
```diff
@@ -0,0 +1,178 @@
...
+type FunctionHooks struct {
+	client resource.ClientApplicator
+}
+
+// NewFunctionHooks returns a new FunctionHooks.
+func NewFunctionHooks(client client.Client) *FunctionHooks {
+	return &FunctionHooks{
+		client: resource.ClientApplicator{
+			Client:     client,
+			Applicator: resource.NewAPIPatchingApplicator(client),
```
negz, line 62: Should we use SSA for this new code, or would that be too big a change? 😬

internal/controller/pkg/revision/runtime_function.go
negz, line 76: Perhaps add a comment about why we created a service in the pre hook while everything else happens in post. I forget the reason - CC @ezgidemirel.

internal/controller/pkg/revision/runtime_function.go
```diff
@@ -0,0 +1,178 @@
...
+
+	return do
+}
+
+func functionServiceOverrides() []ServiceOverride {
+	return []ServiceOverride{
+		// We want a headless service so that our gRPC client (i.e. the Crossplane
+		// FunctionComposer) can load balance across the endpoints.
+		// https://kubernetes.io/docs/concepts/services-networking/service/#headless-services
+		ServiceWithClusterIP(corev1.ClusterIPNone),
```
negz, line 176: FWIW I had noticed recently this didn't seem to be working in master (our ClusterIP was not None) but didn't get around to investigating. Not a big deal/blocker for this PR but I'm hoping it magically fixes it.

internal/controller/pkg/revision/runtime_override_options.go
```diff
@@ -0,0 +1,423 @@
...
+	return func(d *appsv1.Deployment) {
+		d.Spec.Template.Spec.ImagePullSecrets = secrets
+	}
+}
+
+// DeploymentRuntimeWithOptionalImage set the image for the runtime container
+// if it is unset, e.g. not specified in the DeploymentRuntimeConfig.
+func DeploymentRuntimeWithOptionalImage(image string) DeploymentOverride {
+	return func(d *appsv1.Deployment) {
+		if d.Spec.Template.Spec.Containers[0].Image == "" {
```
negz, line 149: Do we need to guard against the case where there are no containers when this is called? (Here and elsewhere.)

internal/controller/pkg/revision/runtime_provider.go
```diff
@@ -0,0 +1,250 @@
...
+		// installed with a different name, e.g. "upbound-provider-aws-family"
+		// whereas meta name is "provider-aws-family".
+		// Below is a migration since deployment selectors are immutable and
+		// we cannot change it in-place.
+		// TODO(turkenh): Remove this migration in a future release, i.e. > v1.17 (v1.14  + 3 supported release ).
+		if kerrors.IsInvalid(err) && strings.Contains(strings.ToLower(err.Error()), "field is immutable") {
+			if err = h.client.Delete(ctx, d); err != nil {
+				return errors.Wrap(err, errDeleteProviderDeployment)
+			}
+			return errors.New("provider deployment was immutable, deleted and needs to be recreated")
```
negz, line 152: Do you think it's pretty safe to delete and recreate the provider deployment? An alternative would be to surface an error asking the operator to do it manually. That way they'd get more control over the timing.

test/e2e/manifests/pkg/deployment-runtime-config/setup/deployment-runtime-config.yaml
```diff
@@ -0,0 +1,45 @@
...
+  deploymentTemplate:
+    metadata:
+      name: iamfreetochoose
+    spec:
+      replicas: 2
+      selector: {}
+      template:
+        #metadata: # This is rejected with unknown field "spec.deploymentTemplate.spec.template.metadata.labels"
+        #  labels:
+        #    somepodlabels: cool
```
negz, line 23: Is this meant to be here?

negz (commented):
Thanks @turkenh!

WRT the design changes, I personally would still vote for what I wrote in https://github.com/crossplane/crossplane/issues/4699#issuecomment-1758631777. I prefer moving the default values out of code and into a configuration file. I think this makes it more explicit and predictable, and I think that's worth folks potentially having to copy ~30 lines of YAML to make a new config.

That said, I can accept this as a compromise. It satisfies the primary goal of the design, which was to offer a full template for a Deployment (and potentially in future a Service, ServiceAccount, etc) and I think its cleaner than the other alternatives we discussed.

cluster/crds/pkg.crossplane.io_functionrevisions.yaml
```diff
@@ -120,6 +117,26 @@ spec:
                   garbage collected based on the parent's RevisionHistoryLimit.
                 format: int64
                 type: integer
+              runtimeConfigRef:
+                default:
+                  name: default
```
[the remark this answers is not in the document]
turkenh: Yes, it is working fine.

I considered moving all defaulting to the same place (e.g. `+kubebuilder:default={"name": "default", "apiVersion": "pkg.crossplane.io/v1beta1", "kind": "DeploymentRuntimeConfig"`} but we would still need defaulting of apiVersion and kind when the name provided.

cluster/crds/pkg.crossplane.io_configurationrevisions.yaml
```diff
@@ -65,24 +65,10 @@ spec:
                   and categorize (scope and select) objects. May match selectors of
                   replication controllers and services. More info: http://kubernetes.io/docs/user-guide/labels'
                 type: object
-              controllerConfigRef:
```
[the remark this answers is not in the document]
turkenh: Added to the draft release notes for v1.14: https://github.com/crossplane/crossplane/releases/tag/untagged-b4401400814b03c01a34

internal/controller/pkg/revision/reconciler.go
```diff
@@ -281,8 +301,11 @@ func SetupProviderRevision(mgr ctrl.Manager, o controller.Options) error {
 		Watches(&v1alpha1.ControllerConfig{}, &EnqueueRequestForReferencingProviderRevisions{
 			client: mgr.GetClient(),
 		}).
+		Watches(&v1beta1.DeploymentRuntimeConfig{}, &EnqueueRequestForReferencingProviderRevisions{
+			client: mgr.GetClient(),
+		}).
```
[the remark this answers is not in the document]
turkenh: Yes, it should be safe since we always introduce the type.

Despite being safe, it was unnecessary, but I didn't feel motivated to change the code there as it would look different than how we used to. Now, you're noticing it, I did it properly: https://github.com/crossplane/crossplane/pull/4744/commits/ecd2b0bca2d4a56e7dd14e3acf8670cf81c896eb

cmd/crossplane/core/core.go
turkenh: This will not have much benefit at this moment, I cannot think of a reason for a user to want to disable the feature.
I prefer to keep this out of feature flagging as the successor of controllerConfig.

internal/controller/pkg/revision/runtime_function.go
turkenh: Yes, we should. 

I indeed made that change but reverted because I worried about the size of the PR and getting things mixed up after the first (unrelated) issue I had while testing that PR 😅 I'll create a follow-up issue instead.

turkenh: https://github.com/crossplane/crossplane/issues/4817

internal/controller/pkg/revision/runtime_function.go
turkenh: Just tried and can confirm that this is working fine with the PR:

```bash
❯ kubectl -n crossplane-system get svc -w
NAME                  TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)    AGE
crossplane-webhooks   ClusterIP   10.96.223.5     <none>        9443/TCP   4h19m
function-dummy        ClusterIP   None            <none>        9443/TCP   7s
```

internal/controller/pkg/revision/runtime_function.go
turkenh: Added a comment clarifying the point with some help from @phisco.

internal/controller/pkg/revision/runtime_override_options.go
turkenh: We have a special modifier that ensures this, `DeploymentWithRuntimeContainer()` and it is called as the first override:

```go
	allOverrides = append(allOverrides,
		// This will ensure that the runtime container exists and always the
		// first one.
		DeploymentWithRuntimeContainer(),
```

If the above changes unintentionally, it will be caught by unit tests. So, I am feeling good about not being defensive here. 

Alternatively, I can inject that one into all modifiers that are assuming the 0th container is a runtime container as follows, but don't like it being called multiple times unnecessarily.

```diff
func DeploymentRuntimeWithOptionalImage(image string) DeploymentOverride {
	return func(d *appsv1.Deployment) {
+               DeploymentWithRuntimeContainer()(d)
		if d.Spec.Template.Spec.Containers[0].Image == "" {
```

internal/controller/pkg/revision/runtime_provider.go
turkenh: Good point. 

I believe it is pretty safe (at least it should be safe) to delete any pod in a k8s cluster at any time, as it could also happen for other reasons (e.g. a node going down). However, it may surface some unrelated issues (e.g. a provider not being resistant to deletes during some operation) or cluster being loaded due to many providers (i.e. plenty of family providers) deleted altogether and need to be recreated simultaneously. 

If we leave this to users, the experience would be:

1. User upgrades to v1.14.
2. All providers and active provider revision will report healthy `false` with an error message asking the user to delete the provider deployment manually. 
3. Providers will still be functional and operating fine until the user does the delete.
4. The user must delete provider deployments to get them back to reporting healthy.

turkenh: A reminder on why I want to change the selectors (even though it is breaking): 

Previously, we were using the package name from `crossplane.yaml` in the package, which requires the package to be parsed, which we no longer [want to do](https://github.com/crossplane/crossplane/pull/4071) unless we have to. However, we could also use `pkg.crossplane.io/package` label on the revision, which is set to the name of the Package resource (e.g. Provider). This way, creating a manifest builder wouldn't require pulling/parsing package. 

This is migration I would like to do eventually, but maybe we can postpone it to the next release (so that we can do it early in the release cycle and get the behavior tested on master for a while) 🤔 For now, I can override those selectors for providers with the package name from provider meta for backward compatibility.

test/e2e/manifests/pkg/deployment-runtime-config/setup/deployment-runtime-config.yaml
turkenh: I'll eventually clean it up, but I would like to ask for some help here. 

I couldn't understand why controller tools didn't generate `metadata.labels` in the schema of `DeploymentRuntimeConfig` and would appreciate some help here. I don't want to ignore this, as this is how one could add labels and annotations on the pod instead of the deployment. Any ideas?

Screenshot 2023-10-16 at 15 04 10

internal/controller/pkg/revision/runtime_provider.go
turkenh: Just pushed a commit for this: https://github.com/crossplane/crossplane/pull/4744/commits/a85fd86f50a2e384e1d3f41f44e922a09ef62b85

turkenh: While testing, I noticed that we will anyway cause restarting provider pods since we made some changes in the TLS secret names, etc. So, the outcome wouldn't be different. 

Still, this could be done separately later and doesn't have to be part of this (already big) PR.

internal/controller/pkg/revision/runtime_override_options.go
negz: SGTM, let's leave it as is.

test/e2e/manifests/pkg/deployment-runtime-config/setup/deployment-runtime-config.yaml
negz: Is it possible controller-tools has special handling for types named ObjectMeta?

I believe you don't need to specify CRD OpenAPI schema for ObjectMeta when it's in the "usual place" (i.e. top-level `metadata`) because the API server handles that automatically. Perhaps controller-tools has some logic to handle that that is affecting us.

negz: Oh right, I just realized this is the deployment template's pod template's metadata. It looks like you can enable generation of this using a controller-runtime argument. See https://github.com/crossplane/crossplane/commit/f7e1ddf4f0782b1e5594ebacb077d3416a55db84

negz (approved):

test/e2e/manifests/pkg/deployment-runtime-config/setup/deployment-runtime-config.yaml
turkenh: Great, thanks a lot!

cmd/crossplane/core/core.go
turkenh: This PR combined two things I attempted to separate but later [gave up](https://github.com/crossplane/crossplane/pull/4767#issuecomment-1761561210): refactoring of existing manifest building and implementing the DeploymentRuntimeConfig. 

Currently, if we want to keep refactoring but bind `DeploymentRuntimeConfig` to a beta flag, what I would do when it is disabled would be to start with an empty deployment template instead of [reading](https://github.com/crossplane/crossplane/pull/4744/files#diff-84987f336046c313143ebf6fb99ef8883be00dd70ad1645ed3b8235cd24f1b1aR186) it from the referenced `DeploymentRuntimeConfig` which wouldn't cause much difference/value.

If I had taken this into account since the beginning, I could have kept the existing code and done the refactoring and implementation as another alternative hook implementation and plugged it when the feature is enabled. This way, I would see more value in providing a way to keep using the previous implementation in case I introduce some bugs here. However, with the current state, I believe that would require some effort that wouldn't pay off and TBH I am happy that we are getting rid of that previous code sooner than later :)

turkenh: > what I would do when it is disabled would be to start with an empty deployment template instead of reading it from the referenced DeploymentRuntimeConfig which wouldn't cause much difference/value.

This could still be helpful when someone is uncomfortable enabling the flexible/powerful configuration interface we're introducing here. 

https://github.com/crossplane/crossplane/pull/4857

PR #4748, opened 2023-10-09
feat: add readiness and liveness probes to crossplane

phisco:

Fixes part of E2E flakiness we are observing, avoiding Crossplane to be marked as Ready before the Webhook server is actually ready. If webhook is not enabled (`webhooks.enabled=false`) both probes are just set to the default no-op.

I have:

### DIFF
--- a/cmd/crossplane/core/core.go
+++ b/cmd/crossplane/core/core.go
@@ -149,7 +149,7 @@ func (c *startCommand) Run(s *runtime.Scheme, log logging.Logger) error { //noli
 
 	cfg, err := ctrl.GetConfig()
 	if err != nil {
-		return errors.Wrap(err, "Cannot get config")
+		return errors.Wrap(err, "cannot get config")
 	}
 
 	cfg.WarningHandler = rest.NewWarningWriter(os.Stderr, rest.WarningWriterOptions{
@@ -193,7 +193,7 @@ func (c *startCommand) Run(s *runtime.Scheme, log logging.Logger) error { //noli
 		HealthProbeBindAddress: ":5000",
 	})
 	if err != nil {
-		return errors.Wrap(err, "Cannot create manager")
+		return errors.Wrap(err, "cannot create manager")
 	}
 
 	o := controller.Options{
@@ -253,7 +253,7 @@ func (c *startCommand) Run(s *runtime.Scheme, log logging.Logger) error { //noli
 			filepath.Join(c.TLSClientCertsDir, corev1.TLSPrivateKeyKey),
 			false)
 		if err != nil {
-			return errors.Wrap(err, "Cannot load TLS certificates for external secret stores")
+			return errors.Wrap(err, "cannot load TLS certificates for external secret stores")
 		}
 
 		o.ESSOptions = &controller.ESSOptions{
@@ -270,7 +270,7 @@ func (c *startCommand) Run(s *runtime.Scheme, log logging.Logger) error { //noli
 	}
 
 	if err := apiextensions.Setup(mgr, ao); err != nil {
-		return errors.Wrap(err, "Cannot setup API extension controllers")
+		return errors.Wrap(err, "cannot setup API extension controllers")
 	}
 
 	po := pkgcontroller.Options{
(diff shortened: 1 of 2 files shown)


sttts (approved):

cmd/crossplane/core/core.go
```diff
@@ -298,17 +301,46 @@ func (c *startCommand) Run(s *runtime.Scheme, log logging.Logger) error { //noli
 		// fleshed out, implement a registration pattern similar to scheme
 		// registrations.
 		if err := xrd.SetupWebhookWithManager(mgr, o); err != nil {
-			return errors.Wrap(err, "cannot setup webhook for compositeresourcedefinitions")
+			return errors.Wrap(err, "Cannot setup webhook for compositeresourcedefinitions")
```
turkenh, line 304: Is this intentional? Shouldn't error messages start with lower case?

cmd/crossplane/core/core.go
```diff
@@ -298,17 +301,46 @@ func (c *startCommand) Run(s *runtime.Scheme, log logging.Logger) error { //noli
...
+	if err := mgr.AddReadyzCheck("ping", healthz.Ping); err != nil {
+		return errors.Wrap(err, "cannot create ping ready check")
+	}
+
+	// Add default health probe
+	if err := mgr.AddHealthzCheck("ping", healthz.Ping); err != nil {
+		return errors.Wrap(err, "cannot create ping health check")
+	}
+
+	// Add probes waiting for the webhook server if webhooks are enabled
```
turkenh, line 335: Is there a way to add something like a started checker for the controllers? 
Could it be helpful to report as ready only after controllers are operational, e.g. after acquiring the lease successfully.

turkenh (approved):
Thanks @phisco 🙌

cmd/crossplane/core/core.go
phisco: I saw other errors above starting with capital letters I thought that made sense give that these are root level errors so I just adapted, but I now noticed there are a few with lowercase, no strong opinion here I can switch them all to lowercase.

turkenh: No strong opinion either, but all lowercase would be my preference.

cmd/crossplane/core/core.go
turkenh: We could also check/do that only after we're aware of a use case that it would help. 
Feel free to ignore for now.

cmd/crossplane/core/core.go
phisco: Done 👌

cmd/crossplane/core/core.go
phisco: Yeah, we could definitely do something, but I couldn't find many examples of other projects doing much more than this. Acquiring the lease could not be the right condition though as we want to support multiple replicas serving webhooks, @sttts was talking about caches being populated, any additional hint/pointer to prior art?

sttts: I don't think readiness is connected to leader election. It's a blackbox property whether the pod is ready to serve traffic (via the webhook). Controllers are always async. No need to wait for them. What does waiting for them even mean?

phisco: Agree, @sttts. What did you have in mind by "waiting for informers" [here](https://crossplane.slack.com/archives/CEF5N8X08/p1696500988710769?thread_ts=1696496917.722149&channel=CEF5N8X08&message_ts=1696500988.710769)? I was referring to that when I asked if you had any prior reference implementation to share.

sttts: If functionality of the process depends on informers to sync, then they should be part of readiness. E.g. admission in kube cannot work correctly without synced informers. Hence, the kube-apiserver might return inconsistent results prior to that.

diegoavitorino:
Hi, I still didn't understood how fix the error Internal error occurred: failed calling webhook "compositions.apiextensions.crossplane.io": failed to call webhook: Post "https://crossplane-webhooks.crossplane-system.svc:9443/validate-apiextensions-crossplane-io-v1-composition?timeout=10s": context deadline exceeded

What I need to do to fix it?

Thanks

crd-n:
Hi @phisco

I read through some issues and feature requests of you regarding readiness, liveness and startup probes.
First, thank you very much for your work!
We usually configure readiness and liveness probes on all of our workload, as we consider it being best practice. Therefore I wanted to ask you, if it would make sense to introduce the readiness and liveness probes back into the helm chart? Maybe with configurable timeous / thresholds, to prevent the unwanted restarts you obversved when first introducing them?

Thanks for your answer!

Issue #4672, opened 2023-09-23
`diagnose` command - a kube-lineage like view for debugging


jbw976:

jbasement:
First of all thanks for picking up the crossplane kubectl cli again! I think that it is a great addition and makes the life of developers and SREs etc. much easier.

Secondly I wanted to ask where the CLI will be located in the future? I saw another issue with a few ideas of moving it to crossplane-contrib or maybe you unarchive the crossplane-cli repo? And crank only has limited capabilities compared to what the crossplane-cli can do. 
Asking so I can follow along on the progress and maybe contribute at some point.

Sorry if that is the wrong issue/ place to ask but I didn't know where else to ask.

jeanduplessis:
@jbasement the best single source for information on all the devex work we are tackling is in the Crossplane Slack [#sig-devex](https://crossplane.slack.com/archives/C05U1LLM3B2) channel.

jbw976:
Very good call out @jbasement, thanks for mentioning the old trace functionality! I updated the description of this issue to explicitly call that out as an areas to assess - it could accelerate our efforts 😉 

I think it's likely in the future that the crossplane CLI moves into its own repo in the crossplane org, but for now it's staying in the main crossplane/crossplane repo within the [`crank`](https://github.com/crossplane/crossplane/tree/master/cmd/crank) binary so we can leverage all the existing build/publish/release support.  Also, we'll make sure to assess any other functionality from the old crossplane-cli that can be salvaged.

jbasement:
For transparency: I'm currently creating a PR to implement[ this CLI functions](https://github.com/jbasement/cp-cli) (`describe` and `diagnose`) I created into crank as per [this slack thread](https://crossplane.slack.com/archives/C05U1LLM3B2/p1696328025719879).

PR #4751, opened 2023-10-09
feat(crank): add describe command

jbasement:
# Description of your changes

This PR adds the `describe` command to the crank/ crossplane CLI. The command is powered by the `graph` package.

## Commands 
`describe` - Describes a given resource (usually a claim or composite) by getting the manifest of the resource and all its children. The resource is then printed out by either a CLI table or a graphviz png. The user has the option to set the printed fields, output type and a few more parameters.

## Packages
`graph` - The graph package is able to get a resource (usually a claim or composite resource) from k8s, including all its children. The graph package implements a printer interface and currently comes with 3 printers to return the resource. A tree (default), table, or dot (must be rendered) printer.
This builds the foundation to implement more advanced commands like e.g. `diagnose`. Additional steps can be implemented between getting a resource and printing it out, like a diagnosis function.

Partially addresses #4672, with the `diagnose` command to come in a later PR

I have:

## Note for reviewer
- I didnt implement unit tests for `resource.go` yet as I'm not sure yet if and how much rework it still needs
- I didnt implement the logger to the packages yet as I wasnt sure what is the best way to implement it there. It would be happy to get a suggestion there how I could implement it

# Screenshots
## `describe` command with different output formats

CLI tree
image

CLI table
image

GraphViz png
`crank describe objectstorage test-bucket-00002 -o graph | dot -Tpng  > test.png && open test.png`
image


phisco:
Hi @jbasement, thanks for your contribution!
Could you add to the description screenshots of the describe and diagnose commands outputs, that would help people having a better grasp of what you are proposing without having to actually run it first. 🙏 
Then, would it make sense to address the two subcommands one at a time in two dedicated PRs, just to make the process smoother?

jbasement:
Good idea, I added some screenshots to the PR description.
Sure I could split up the PR for the two commands, but both commands depend on the packages added in this PR. So if we split up the PRs first this PR has to be merged before we can implement the second command. If we still want to go ahead and split the PR in two, I'd suggest first implementing the `describe` command as it gives a better overview of the k8s resource and then in a second PR implement the `diagnose` command. wdyt?

phisco:
Yes, @jbasement, I think splitting, even if they have to be serialized, would make the review process easier. Let's first get a good describe command and then on top of that add the diagnose functionalities 😉

jbasement:
@phisco Agreed that makes sense! :) Updated the branch and the PR description accordingly. Keeping the branch with the diagnose command local for now until this PR is merged.

internal/printer/graphviz.go
```diff
@@ -0,0 +1,98 @@
...
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"strings"
+
+	"github.com/crossplane/crossplane/internal/k8s"
+	"github.com/emicklei/dot"
+	"github.com/goccy/go-graphviz"
```
jbw976, line 11: i wanted to take this PR for a test spin to get a better feel for the functionality, but I'm having trouble building on my Mac M1 Max:

```
❯ make
<omitted...>
22:33:30 [ .. ] go build darwin_arm64
github.com/crossplane/crossplane/cmd/crossplane
package github.com/crossplane/crossplane/cmd/crank
	imports github.com/crossplane/crossplane/internal/printer
	imports github.com/goccy/go-graphviz
	imports github.com/goccy/go-graphviz/cgraph
	imports github.com/goccy/go-graphviz/internal/ccall: build constraints exclude all Go files in /Users/jared/go/pkg/mod/github.com/goccy/go-graphviz@v0.1.1/internal/ccall
22:33:33 [FAIL]
```

Is there additional build or environment config needed to build this? I'm not sure if we're already using cgo, but I'm  hesitant to add that complexity to our build and dependencies if we don't absolutely have to. e.g. https://relistan.com/cgo-when-and-when-not-to-use-it

internal/printer/cli.go
```diff
@@ -0,0 +1,76 @@
...
+import (
+	"fmt"
+	"os"
+
+	"github.com/crossplane/crossplane/internal/k8s"
+	"github.com/olekukonko/tablewriter"
+)
+
+// Prints out a CLI table of the passed Resource. The fields variable determines the header and values of the table.
+func CliTable(rootResource k8s.Resource, fields []string) error {
```
jbw976, line 12: I'm having trouble personally parsing the table view as that format doesn't effectively visually capture the relationships between the resources in the table.

Is it possible to instead pursue a tree view? That would very clearly demonstrate the hierarchy and relationships to the user and overall be more effective. That may be much harder to implement though...what do you think?

jbw976 (commented):
very very **very** cool to see your initiative @jbasement and your willingness to start contributing to improving the developer experience in Crossplane! Big kudos for jumping in here and sharing your efforts with the community! 🙇‍♂️ 

I love the idea here but I do have a couple high level questions about the experience that we'll want to figure out first, see comments for details.

* table view vs. tree view
* usage of cgo

Really looking forward to your contributions going forward too 😉

internal/printer/graphviz.go
jbasement: Hm no I didn't set any specific build or environment config for go-graphviz to work. 
But coming back to your point about not using cgo, we can make it work without the go-graphviz library. The graph file is built using the dot library. This library doesn't support saving the graph as a PNG that's why I started using go-graphviz in the first place. But the graph can be rendered like this by the user.`<my-command> | dot -Tpng  > test.png && open test.png` 
https://github.com/emicklei/dot#display-your-graph

internal/printer/cli.go
phisco: +1

jbasement: Sure we can try to implement a tree view instead of/ additionally to the table. What do you have in mind for the tree view? Something like kube-lineage, which is a mix of table and tree? Or really only a plain [tree](https://github.com/xlab/treeprint) that instead of file and dir names has the resource information?

internal/printer/graphviz.go
jbw976: +1 for offloading the rendering / image generation to an external tool. @phisco mentioned that this is symmetrical with what the golang tools do as well and it would remove complexity from our build/dependencies.

internal/printer/cli.go
jbw976: As discussed in the SIG-DevEx meeting today, a [kube-lineage](https://github.com/tohjustin/kube-lineage) hybrid tree/table view would be ideal, but is probably too difficult to implement before feature freeze.  Let's start with a very simple tree view that shows kind/name with hierarchical indentation and we can always make it more fancy later.

Let's also keep the table view too, no need to throw that work away, but let's make the tree view as the default experience.

internal/printer/graphviz.go
phisco: Referring to https://go.dev/ref/mod#go-mod-graph

jbasement: Also thinking about it now, it is easier to test if the function is working by comparing the dot output instead of checking a PNG file. Reworked the function now as discussed and removed the graphviz library

internal/printer/cli.go
jbasement: Added the tree output to the printer package and set it as default. I added a screenshot of how it looks atm. The user also has the option to set the fields as in the other output methods. Please let me know how you like the current design of the output. @phisco @jbw976 

I'll give the kube-lineage-like tree/table a try this weekend but will focus primarily on any feedback for the existing implementation in this PR.

jbw976 (commented):
This is really good @jbasement, I am loving the tree view approach that we default to now! Thanks for taking the directional feedback and incorporating that into the experience and thanks again for your efforts on this. A true member of SIG-DevEx! 🤩 😂  

There are multiple failing checks for the PR status, let's see if we can get those all fixed.  A great start to test those locally will be running `make reviewable`.

I have a few comments to consider, but I would actually love to get this merged pretty quickly and not spend a ton of time on a longer tail of work in this PR. This is because I think the `crossplane diagnose` command is even more important than this, and getting that in before feature freeze on Wednesday would be the best possible scenario in my book 🤓 

@phisco if you could also help get this PR to an approved state so we can merge soon and focus on `crossplane diagnose`, that would be awesome! 🙇‍♂️

internal/printer/cliTable_test.go
```diff
@@ -0,0 +1,117 @@
+package printer
+
+// import (
```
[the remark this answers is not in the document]
jbasement: I commented them out as I was implementing the Printer interface and didn't adjust the tests accordingly yet. But going to fix the tests today

internal/printer/graph_test.go
```diff
@@ -0,0 +1,84 @@
+package printer
+
+// import (
```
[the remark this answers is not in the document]
jbasement: Yes going to fix them today

internal/printer/cliTable.go
```diff
@@ -0,0 +1,92 @@
+package printer
```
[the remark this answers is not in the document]
jbasement: Agreed, I moved the printer to `cmd/crank/printer`

internal/k8s/resource.go
```diff
@@ -0,0 +1,347 @@
...
+
+type Client struct {
+	dclient   *dynamic.DynamicClient
+	clientset *kubernetes.Clientset
+	rmapper   meta.RESTMapper
+	dc        *discovery.DiscoveryClient
+}
+
+type Resource struct {
+	Manifest           *unstructured.Unstructured
```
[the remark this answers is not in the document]
jbasement: Yes you're right same for `LatestEventMessage`, I also set it to private. I left `Children` exported for now as they have to be accessed in the printer functions, but we could implement helper functions for the Children in the future as well

internal/k8s/resource.go
```diff
@@ -0,0 +1,347 @@
...
+	return r.Manifest.GetAPIVersion()
+}
+
+// This function takes a certain conditionType as input e.g. "Ready" or "Synced"
+// Returns the Status of the map with the conditionType as string
+func (r *Resource) GetConditionStatus(conditionKey string) string {
+	conditions, _, _ := unstructured.NestedSlice(r.Manifest.Object, "status", "conditions")
+	for _, condition := range conditions {
+		conditionMap, _ := condition.(map[string]interface{})
+		conditionType, _ := conditionMap["type"].(string)
```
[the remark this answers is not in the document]
jbasement: You're right I added a check to all assertions.
`type` should be fine ([source](https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/types.go#L1507)), still added the check in case the `type` field might change in the future

internal/k8s/resource.go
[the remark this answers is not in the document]
jbasement: I export the vars of the struct so I can init a complete struct for the tests in the `printer` package. In the future, I could maybe add a `CreateDummyResource` function that returns Resources for testing.
But let me know if there is a better or simpler way to solve this :)

internal/k8s/resource.go
```diff
@@ -0,0 +1,347 @@
...
+
+// Returns true if the Resource has children set.
+func (r *Resource) HasChildren() bool {
+	return len(r.Children) > 0
+}
+
+// The main function of resource. Returns a Resource and all its child resources.
+func GetResource(resourceKind string, resourceName string, namespace string, kubeconfig *rest.Config) (*Resource, error) {
+	// Init new client
+	client, err := newClient(kubeconfig)
```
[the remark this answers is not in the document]
jbasement: Good point. Moved the client init away from the k8s package

cmd/crank/describe.go
```diff
@@ -0,0 +1,96 @@
...
+	case "tree":
+		p = &printer.TreePrinter{
+			Indent: "",
+			IsLast: true,
+		}
+	case "table":
+		p = &printer.TablePrinter{}
+	case "graph":
+		p = &printer.GraphPrinter{}
+	}
```
phisco, line 87: I'd probably make "tree", "table" and "graph" constants with a dedicated type and move this switch to the printer package, doing also there the validation of the known types, something like:
```golang
func New(type string) (Printer, error) {
  switch Type(type) {
    case TypeTree:
      return &printer.TreePrinter{
			Indent: "",
			IsLast: true,
		}, nil
... etc ...
  default:
    return nil, errors.Errorf("unknown printer output type: %s", type)
  }
  ...
}
```

cmd/crank/describe.go
```diff
@@ -0,0 +1,96 @@
...
+	Namespace  string   `short:"n" name:"namespace" help:"Namespace of resource to describe." default:"default"`
+	Output     string   `short:"o" name:"output" help:"Output type of graph. Possible output types: tree, table, graph." enum:"tree,table,graph" default:"tree"`
+	Fields     []string `short:"f" name:"fields" help:"Fields that are printed out in the header." default:"kind,name"`
+	OutputPath string   `short:"p" name:"path" help:"Output path for graph PNG. Only valid when set with output=graph." default:"graph.png"`
+}
+
+func (c *describeCmd) Run(logger logging.Logger) error {
+	logger = logger.WithValues("Kind", c.Kind, "Name", c.Name)
+
+	allowedFields := []string{"parent", "name", "kind", "namespace", "apiversion", "synced", "ready", "message", "event"}
```
phisco, line 36: I would suggest moving this to be a global variable

cmd/crank/describe.go
```diff
@@ -0,0 +1,96 @@
...
+	allowedOutput := []string{"tree", "table", "graph"}
```
phisco, line 37: see comment below about the allowed values and validation

phisco (commented):
left a few comments 👍

internal/k8s/resource.go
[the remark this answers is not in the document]
jbasement: Yes, I agree that's a really good idea. Implemented the changes as suggested

internal/k8s/resource.go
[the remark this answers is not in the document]
jbasement: Merged the packages and made manifest a local now

internal/k8s/resource.go
```diff
@@ -0,0 +1,362 @@
...
+
+// GetNamespace returns resource namespace as string
+func (r *Resource) GetNamespace() string {
+	return r.Manifest.GetNamespace()
+}
+
+// GetAPIVersion returns resource apiversion as string
+func (r *Resource) GetAPIVersion() string {
+	return r.Manifest.GetAPIVersion()
+}
```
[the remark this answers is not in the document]
jbasement: Good point I should've checked that 👍

cmd/crank/describe.go
jbasement: Done

cmd/crank/describe.go
jbasement: Commented above

cmd/crank/describe.go
jbasement: I implemented your suggestion in this commit https://github.com/crossplane/crossplane/pull/4751/commits/30b979234e2b9b2dcacab017f82784ec4d25d952. Let me know if it is implemented as expected.

jbasement: And small fix https://github.com/crossplane/crossplane/pull/4751/commits/bf575090521f5d82f7829d355fb6f08394fbb547

jbasement:
@phisco Thanks for the thorough review! Regarding the 3 open comments, I don't think I'll find the time before the code freeze to implement these. Not sure if they are showstoppers or if someone else might be able to implement those in time?

cmd/crank/describe.go
phisco: Yes, that's exactly what I meant! Looks good 🙏

jbw976:
@phisco is there anything remaining that must be fixed before we can approve/merge this PR? It would be great to get it in and then iterate to add the `diagnose` functionality, which IMO is the most important feature in this particular area 🤩

jbasement:
> I think something broke as it's not able to lookup a claim: image
> 
> To reproduce:
> 
> * just install the latest helm chart (v1.13.2): `helm upgrade --install crossplane crossplane-stable/crossplane --namespace crossplane-system --create-namespace --set args='{--debug}'`
> * apply the manifests in the manifests from the e2es:`kubectl apply -f test/e2e/manifests/apiextensions/composition/minimal/setup`
> * create the claim in the default namespace: `kubectl apply -f test/e2e/manifests/apiextensions/composition/minimal/claim.yaml`

I think the issue is a typo in your command when executing. You are referencing `....example.or` instead of `example.org`. But would be a nice feature to have a suggestion like kubectl has

phisco:
good call 😅 however I think that was just the latest attempt and it was already failing given that it's still failing now:
image

Works fine for composite resources, but not for claims, I guess some issue with the namespace? 🤔

jbasement:
@phisco I was able to fix the issues.
The issue was related to the `isResourceNamespaced` function and the other issue related to the restmapper.
It mapped the passed group `nopresources.nop.example.org` to the similar group `nopresources.nop.crossplane.io`. This surely can be solved in a better way than I did but it works for now. My change added a bit of (probably unnecessary) complexity to the function now

phisco:
Awesome @jbasement! works like a charm now, thank you!

PR #4752, opened 2023-10-09
tests(e2e): add events to related objects and report also on deletion

phisco:
### Description of your changes

Add also events as related objects to show on E2E failures.

I have:

- [x] Read and followed Crossplane's [contribution process].
- [ ] ~Added or updated unit **and** E2E tests for my change.~
- [x] Run `make reviewable` to ensure this PR is ready for review.
- [ ] ~Added `backport release-x.y` labels to auto-backport this PR, if necessary.~
- [ ] ~Opened a PR updating the [docs], if necessary.~

[contribution process]: https://git.io/fj2m9
[docs]: https://docs.crossplane.io/contribute/contribute

PR #4754, opened 2023-10-09
Log events automatically

sttts:

To corrolate events and other log output easily, it's very helpful to log events too:

```
2023-10-09T10:54:16Z	INFO	crossplane	Event(v1.ObjectReference{Kind:"CompositeResourceDefinition", Namespace:"", Name:"xnopresources.nop.example.org", UID:"1a0324d0-f55e-4741-ba45-8384ddd3b971", APIVersion:"apiextensions.crossplane.io/v1", ResourceVersion:"911", FieldPath:""}): type: 'Normal' reason: 'RedactClaim' Stopped composite resource claim controller
```

With this we can also get rid of many logs that we manually create when an event is created in parallel.

I have:

### DIFF
--- a/cmd/crossplane/core/core.go
+++ b/cmd/crossplane/core/core.go
@@ -20,6 +20,7 @@ package core
 import (
 	"context"
 	"crypto/tls"
+	"fmt"
 	"net/http"
 	"net/http/pprof"
 	"os"
@@ -33,6 +34,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/leaderelection/resourcelock"
+	"k8s.io/client-go/tools/record"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -157,6 +159,7 @@ func (c *startCommand) Run(s *runtime.Scheme, log logging.Logger) error { //noli
 		Deduplicate: true,
 	})
 
+	eb := record.NewBroadcaster()
 	mgr, err := ctrl.NewManager(ratelimiter.LimitRESTConfig(cfg, c.MaxReconcileRate), ctrl.Options{
 		Scheme: s,
 		Cache: cache.Options{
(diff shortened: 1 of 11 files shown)


cmd/crossplane/core/core.go
```diff
@@ -196,6 +200,11 @@ func (c *startCommand) Run(s *runtime.Scheme, log logging.Logger) error { //noli
 		return errors.Wrap(err, "cannot create manager")
 	}
 
+	eb.StartLogging(func(format string, args ...interface{}) {
+		log.Info(fmt.Sprintf(format, args...))
```
phisco, line 204: Info or debug? 🤔

phisco (approved):
Makes sense to me, also useful for the E2Es, in the same direction as https://github.com/crossplane/crossplane/pull/4752

cmd/crossplane/core/core.go
sttts: I want to see when we are hotlooping. That we only get from `Info`.

sttts: Or another argument: considering logs lower-level than events, if we see too many event logs, we very clearly have too many kube events too 😄

turkenh: I am also concerned that it could be too verbose for info. This is also consistent with current manual logging behavior.

sttts: why too verbose?

> considering logs are lower-level than events

If this is too verbose for the logs, we should not create that many events in the first place. Events are end-user-facing.

negz: Please leave these as debug in this PR. We're due for a discussion on our approach to logging in this project, but I don't want to conflate that with this otherwise small and valuable change.

negz:
Does controller-runtime already log events automatically (at debug level)? I haven't checked, but I half-recall it might. If that is indeed the case we could probably just delete the redundant explicit log statements.

sttts:
> Does controller-runtime already log events automatically (at debug level)? I haven't checked, but I half-recall it might. If that is indeed the case we could probably just delete the redundant explicit log statements.

It's what I enable here. By default it's off.

sttts:
Updated. Ptal.

negz (approved):

Issue #4670, opened 2023-09-23
`render` to test composition functions locally


jbw976:
To greatly improve the experience and efficiency of testing Composition Functions, we should build tooling support in the Crossplane CLI for locally executing and testing a XFN. This could be exposed as a `render` command that takes as arguments your composition, composite resource, and functions to test.  The command would ensure all Functions are present (downloading if needed) locally, then render out the given composition with the input values from the given composite resource.

This functionality is described in the XFN design doc: https://github.com/crossplane/crossplane/blob/master/design/design-doc-composition-functions.md#iterating-on-compositions

This will allow 3 key personas to make rapid progress while iterating on what they are building:

1. XFN developers that are coding the logic for a new Function 
1. Composition authors that are configuring a pipeline of existing (e.g. general) functions
2. A hybrid persona that is both coding a new Function and configuring a Composition pipeline to consume the new Function

This experience would "shift left" a lot of work that's currently done on live clusters today, providing a much more efficient and safe experience.

@negz has started a great POC of this functionality that's available at https://github.com/crossplane-contrib/xrender.

lsviben:
I would assume so. Put it under an alpha flag and continue iterating in our new CLI.

negz:
Maybe `beta` for this command? 😬 I could be convinced to do alpha, but we're calling Functions in general "beta".

PR #4736, opened 2023-10-05
feat(crank): add init function subcommand

phisco:
<!--
Thank you for helping to improve Crossplane!

Please read through https://git.io/fj2m9 if this is your first time opening a
Crossplane pull request. Find us in https://slack.crossplane.io/messages/dev if
you need any help contributing.
-->

### Description of your changes

<!--
Briefly describe what this pull request does, and how it is covered by tests.
Be proactive - direct your reviewers' attention to anything that needs special
consideration.

You MUST either [x] check or ~strikethrough~ every item in the checklist below.

We love pull requests that fix an open issue. If yours does, use the below line
to indicate which issue it fixes, for example "Fixes #500".
-->

Fixes #4668
```shell
$ crank --help
Usage: crossplane <command>

A command line tool for interacting with Crossplane.

Commands:
  build              Build Crossplane packages.
    configuration    Build a Configuration package.
    provider         Build a Provider package.
    function         Build a Function package.

  install            Install Crossplane packages.
    configuration    Install a Configuration package.
      <package>      Image containing Configuration package.
      [<name>]       Name of Configuration.
    provider         Install a Provider package.
      <package>      Image containing Provider package.
      [<name>]       Name of Provider.

  update             Update Crossplane packages.
    configuration    Update a Configuration package.
      <name>         Name of Configuration.
      <tag>          Updated tag for Configuration package.
    provider         Update a Provider package.
      <name>         Name of Provider.
      <tag>          Updated tag for Provider package.

  push               Push Crossplane packages.
    configuration    Push a Configuration package.
      <tag>          Tag of the package to be pushed. Must be a valid OCI image tag.
    provider         Push a Provider package.
      <tag>          Tag of the package to be pushed. Must be a valid OCI image tag.

  xpkg               Crossplane package management.
    build            Build a package, by default from the current directory.
    push             Push Crossplane packages.
      <tag>          Tag of the package to be pushed. Must be a valid OCI image tag.

  alpha              Alpha features. WARN: May be changed or removed without notice
    init             Initialize directories for Crossplane packages from templates.
      function       Initialize a Function from the repo template for a selected language.
        <name>       Name of the Function to initialize.

  beta               Beta features. WARN: May be changed or deprecated in a future release

Flags:
  -h, --help                    Show context-sensitive help.
  -v, --version=VERSION-FLAG    Print version and quit.
      --verbose                 Print verbose logging statements.

Run "crossplane <command> --help" for more information on a command.
```

```shell
$ crank alpha init --help
[...]

PR #4755, opened 2023-10-09
feat(crank): alpha/beta subcommands and tree view

### LINKED ISSUE #4720: Add alpha command/gate that we can put new/experimental CLI commands behind
As we are adding new CLI commands, we need a mechanism to put the new/experimental ones behind an alpha flag, so that we have flexibility in iterating on them.

We could use the [up](https://github.com/upbound/up/blob/main/cmd/up/main.go#L84-L92) maturity feature for that.

phisco:

Fixes https://github.com/crossplane/crossplane/issues/4720 by setting up two placeholder subcommands ready to have subcommands added in later on, e.g. https://github.com/crossplane/crossplane/issues/4670.

From:
image

To:
image

I would have preferred not having the arguments in the "top level" help view, but looks like it's not so straightforward to implement, so this should be enough for now.

I have:

### DIFF
--- a/cmd/crank/main.go
+++ b/cmd/crank/main.go
@@ -65,6 +65,14 @@ var cli struct {
 	Push    pushCmd    `cmd:"" help:"Push Crossplane packages."`
 
 	XPKG xpkg.Cmd `cmd:"" help:"Crossplane package management."`
+
+	Alpha struct {
+		// Add here alpha subcommands
+	} `cmd:"" help:"Alpha features. WARN: May be changed or removed in a future release"`
+
+	Beta struct {
+		// Add here beta subcommands
+	} `cmd:"" help:"Beta features. WARN: May be changed or removed in a future release"`
 }
 
 func main() {
@@ -82,6 +90,10 @@ func main() {
 		// at runtime.
 		kong.Bind(buildChild, pushChild),
 		kong.BindTo(logger, (*logging.Logger)(nil)),
+		kong.ConfigureHelp(kong.HelpOptions{
+			Tree:      true,
+			FlagsLast: true,
+		}),
 		kong.UsageOnError())
 	err := ctx.Run()
 	ctx.FatalIfErrorf(err)
--- a/cmd/crank/main.go
+++ b/cmd/crank/main.go
@@ -68,11 +68,11 @@ var cli struct {
 
 	Alpha struct {
 		// Add here alpha subcommands
-	} `cmd:"" help:"Alpha features. WARN: May be changed or removed in a future release"`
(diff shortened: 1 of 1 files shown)


cmd/crank/main.go
```diff
@@ -65,6 +65,14 @@ var cli struct {
 	Push    pushCmd    `cmd:"" help:"Push Crossplane packages."`
 
 	XPKG xpkg.Cmd `cmd:"" help:"Crossplane package management."`
+
+	Alpha struct {
+		// Add here alpha subcommands
+	} `cmd:"" help:"Alpha features. WARN: May be changed or removed in a future release"`
```
jeanduplessis, line 71:
```suggestion
	} `cmd:"" help:"Alpha features. WARN: May be changed or removed without notice"`
```

cmd/crank/main.go
```diff
@@ -65,6 +65,14 @@ var cli struct {
...
+
+	Beta struct {
+		// Add here beta subcommands
+	} `cmd:"" help:"Beta features. WARN: May be changed or removed in a future release"`
```
jeanduplessis, line 75:
```suggestion
	} `cmd:"" help:"Beta features. WARN: May be changed or deprecated in a future release"`
```

jeanduplessis (commented):
Some non-blocking proposals for aligning the alpha/beta warning to our Crossplane docs definition.

turkenh (approved):

negz (approved):

cmd/crank/main.go
```diff
@@ -65,6 +65,14 @@ var cli struct {
...
 
 	XPKG xpkg.Cmd `cmd:"" help:"Crossplane package management."`
+
+	Alpha struct {
+		// Add here alpha subcommands
+	} `cmd:"" help:"Alpha features. WARN: May be changed or removed without notice"`
+
+	Beta struct {
+		// Add here beta subcommands
+	} `cmd:"" help:"Beta features. WARN: May be changed or deprecated in a future release"`
```
negz, line 75: Should we put the function related commands under here? I don't feel strongly, but it was where I intuitively expected them to be since Functions overall is a beta feature.

phisco: Yes, I was thinking about that too https://github.com/crossplane/crossplane/pull/4736#issuecomment-1752881163, I'll update the other PRs when this is merged

jbw976 (commented):
> I would have preferred not having the arguments in the "top level" help view, but looks like it's not so straightforward to implement, so this should be enough for now.

+1 on this, I'd prefer to have just the commands in the top level tree view for help, especially as the list of commands and their args/flags grows. Thanks for already looking into if that was easy/possible.

The tree view is an overall improvement either way! 💪

PR #4758, opened 2023-10-09
fix: just requeue all conflict errors

### LINKED ISSUE #2114: Suppress "object has been modified" warning events?

Crossplane relies heavily on Kubernetes events to help folks debug what it is doing. It's quite common for Crossplane to emit the following warning event:

```console
Warning  ConfigureCompositeResource  19s (x2 over 20s)  offered/compositeresourcedefinition.apiextensions.crossplane.io  cannot patch object: Operation cannot be fulfilled on compositenetworks.gcp.platformref.crossplane.io "test-network-d6xjh": the object has been modified; please apply your changes to the latest version and try again
```

We emit this event when two controllers race to update a resource. Kubernetes uses a form of optimistic concurrency in which each resource is versioned (via `metadata.resourceVersion`), and clients must have the latest version of the resource when attempting to update it. This means that if two clients get the same version of a resource then race to update it, the one that loses will encounter this error from the API server.

This is almost never a cause for concern; Crossplane will just try again. The event does make it appear like something is awry though, and is frequently a red herring for Crossplane users trying to debug unrelated issues.

Crossplane could detect this I think it manifests as a 409 Conflict, so we could detect it using a `resource.IsConfict(err)` helper and either completely suppress it or emit it as a `Normal` event.

In the meantime, this would be a good thing to add to an FAQ.

phisco:

Fixes https://github.com/crossplane/crossplane/issues/2114.

I went through all the Reconcilers looking at all the places we were emitting warning events on errors and checked whether the error could come from an update of a resource, if so added a check similar to what we already did for https://github.com/crossplane/crossplane-runtime/pull/540.

Took the chance to rename all aliased imports from "apierrors" to "kerrors" as the latter was the prevalent name on master already.

I have:

### DIFF
--- a/cmd/crank/install.go
+++ b/cmd/crank/install.go
@@ -26,7 +26,7 @@ import (
 	"github.com/alecthomas/kong"
 	"github.com/google/go-containerregistry/pkg/name"
 	corev1 "k8s.io/api/core/v1"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/client-go/tools/cache"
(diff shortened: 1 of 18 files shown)


sttts (approved):

negz (approved):
Thanks for handling this not very fun to do but very impactful change @phisco!

PR #4764, opened 2023-10-10
feat(crank): add render command from xrender

### LINKED ISSUE #4670: `render` to test composition functions locally

phisco:

Fixes https://github.com/crossplane/crossplane/issues/4670.
Based on #4755, so keeping it as draft waiting for that to be merged.

Add github.com/crossplane-contrib/xrender, originally authored by @negz, as  beta command to the `crossplane cli`.

Open Points:
- `xrender` defined a few annotations with `xrender.crossplane.io` as a prefix, do we want to rename the `xrender` part? just render? keep them as they are? This might depend on what we want to do with github.com/crossplane-contrib/xrender, archive it in favor of this?
- `xrender` defined a bunch of [examples](https://github.com/crossplane-contrib/xrender/tree/main/examples), do we want to port them over? I'd say so, but I wasn't sure where to put them in this repo.

Can be tested by building it first (`make build`) then running `_output/bin/darwin_arm64/crank beta render examples/xr.yaml examples/composition.yaml examples/functions.yaml` with the [examples](https://github.com/crossplane-contrib/xrender/tree/main/examples) from `xrender`.

I have:


haarchri:
@phisco can we add the live file option from @stevendborrelli  ? https://github.com/crossplane-contrib/xrender/pull/19

phisco:
@haarchri, totally missed that PR, I'd prefer get this in first, then we can start thinking moving PRs over from xrender and add new functionalities, makes sense?

stevendborrelli:
Yeah, I'd like to have that PR be in a little better shape. I agree with moving forward on the main functionality.

phisco:
> No strong preference from me here - render.crossplane.io or xrender.crossplane.io both seem good.

Went with render, as `xrender` didn't make much sense if we are going to archive `crossplane-contrib/xrender`, which I think we should do.

> Maybe we could go without these if examples were added to docs.crossplane.io instead?

Sounds good to me!

cmd/crank/render/cmd.go
```diff
@@ -41,6 +43,14 @@ func (c *Cmd) Run() error { //nolint:gocyclo // Only a touch over.
 		return errors.Wrapf(err, "cannot load Composition from %q", c.Composition)
 	}
 
+	warns, errs := comp.Validate()
+	for _, warn := range warns {
+		fmt.Fprintf(k.Stderr, "WARN(composition): %s\n", warn)
```
negz, line 48: Should we use `k.Errorf` here instead? I think it would be equivalent.

https://pkg.go.dev/github.com/alecthomas/kong#Kong.Errorf

cmd/crank/render/cmd.go
```diff
@@ -81,13 +91,13 @@ func (c *Cmd) Run() error { //nolint:gocyclo // Only a touch over.
 
 	s := json.NewSerializerWithOptions(json.DefaultMetaFactory, nil, nil, json.SerializerOptions{Yaml: true})
 
-	fmt.Println("---")
+	fmt.Fprintln(k.Stdout, "---")
```
negz, line 94: Perhaps https://pkg.go.dev/github.com/alecthomas/kong#Kong.Printf?

negz: Oh never mind. I see `k.Printf` includes the application name as a prefix, which we don't want.

negz (approved):
One nit, but otherwise this looks ready to go. Thanks!

cmd/crank/render/cmd.go
phisco: It will prefix the application name, I think we should review outputs across all commands to make it a more pleasant and cohesive experience

jbw976 (commented):
Awesome to see this functionality coming quick and super mega thank you to @negz for getting this feature in place with the `xrender` effort!  🙇‍♂️ 

@phisco looks like the code coverage check is failing, probably because of missing coverage in the cmd layer (the load functionality is well covered, render is 50%).  Did you already have ideas about how to handle unit test coverage for the cmd layer? 🤔 

* https://github.com/crossplane/crossplane/pull/4764/checks?check_run_id=17592983367
* https://app.codecov.io/gh/crossplane/crossplane/pull/4764?src=pr&el=h1

phisco:
I could try to improve the situation a little bit, but it will require some refactoring and therefore a bit of time, or we can just get the functionality in straight away and come back at it later on. As you prefer.

phisco:
I think codecov got a bit confused there for some reason, but with a bit of deduplication and some minor testing I think it should not be below the wanted testing %. Could be good to go 👍

PR #4767, opened 2023-10-10
Refactor Package Manager Runtime Manifest Building 

### LINKED ISSUE #4690: Consider using specific package types

turkenh:

This PR: 

- Refactors existing hooks and manifest generation with a ManifestBuilder.
- Introduces `PackageWithRuntime` and `PackageRevisionWithRuntime` interfaces.
- Uses specific package types (Fixes #4690).
- Cleans up the unused `controllerRef` from PackageRevisionStatus.

P.S: I wanted to have a dedicated PR for the work here instead of combining it with Package Runtime Config [Implementation](https://github.com/crossplane/crossplane/pull/4744). I'll rebase that PR on top of this.

I have:


internal/controller/pkg/revision/runtime.go
```diff
@@ -0,0 +1,253 @@
...
+func (b *RuntimeManifestBuilder) Deployment(serviceAccount string, overrides ...DeploymentOverrides) *appsv1.Deployment {
+	d := defaultDeployment(b.revision.GetName())
+
+	overrides = append(overrides,
+		DeploymentWithNamespace(b.namespace),
+		DeploymentWithOwnerReferences([]metav1.OwnerReference{meta.AsController(meta.TypedReferenceTo(b.revision, b.revision.GetObjectKind().GroupVersionKind()))}),
+		DeploymentWithSelectors(b.podSelectors()),
+		DeploymentWithServiceAccount(serviceAccount),
+		DeploymentWithImagePullSecrets(b.revision.GetPackagePullSecrets()),
+		DeploymentRuntimeWithImage(b.revision.GetSource()),
```
phisco, line 151: we used to default to the revision's source, but overwrite it with the meta if specified, which is the cause of the usage e2e failure

phisco: see https://github.com/crossplane/crossplane/pull/4767/files#diff-53e9c93d08b0ecb30843baebd9c68749a3bcc0640aa836972621e9a9243bb15fL125

turkenh: Good catch! Just pushed https://github.com/crossplane/crossplane/pull/4767/commits/bb11ae81672bf566f421b74de1cdd80e7f555104 to fix.

internal/controller/pkg/revision/runtime_defaults.go
```diff
@@ -0,0 +1,95 @@
...
+	return &corev1.ServiceAccount{
+		ObjectMeta: metav1.ObjectMeta{
+			// It is possible to override the name of the service account in the
+			// RuntimeConfig. So, we define it as a default here.
+			Name: name,
+		},
+	}
+}
+
+func defaultDeployment(name string) *appsv1.Deployment {
```
phisco, line 44: wouldn't it make sense to have this return something like a `DeploymentBuilder` which could be the receiver for all the functions in internal/controller/pkg/revision/runtime_override_options.go, such as `DeploymentWithName`, doing the same for Service and ServiceAccount?

phisco (commented):
left a few comments, still have to go through it more thoroughly

turkenh:
@phisco thanks for your review. 

Unfortunately, my goal of getting this merged until https://github.com/crossplane/crossplane/pull/4744 is ready hasn't worked out, and maintaining this PR and another on top of this would cause unnecessary rebase/resolve conflict work for me. So, I would like to merge these back. 

Closing this one, but I'll respond your comments here and fix them there if needed.

cluster/crds/pkg.crossplane.io_functionrevisions.yaml
```diff
@@ -66,8 +66,9 @@ spec:
                   replication controllers and services. More info: http://kubernetes.io/docs/user-guide/labels'
                 type: object
               controllerConfigRef:
-                description: ControllerConfigRef references a ControllerConfig resource
+                description: 'ControllerConfigRef references a ControllerConfig resource
```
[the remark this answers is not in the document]
turkenh: Looks like it is being added when there are some special characters in the description. In this case, it was added because of `:` after the word `Deprecated`. Given we have other descriptions starting with `'`, that is fine.

cluster/crds/pkg.crossplane.io_providerrevisions.yaml
```diff
@@ -79,10 +80,6 @@ spec:
                 description: DesiredState of the PackageRevision. Can be either Active
                   or Inactive.
                 type: string
-              essTLSSecretName:
```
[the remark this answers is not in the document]
turkenh: Do you have any concrete scenarios on how this could fail? 

I am planning to do some upgrade/rollback tests, please let me know if there is a specific case that I should consider.

internal/controller/pkg/revision/reconciler.go
```diff
@@ -163,11 +164,11 @@ func WithDependencyManager(m DependencyManager) ReconcilerOption {
 	}
 }
 
-// WithHooks specifies how the Reconciler should perform pre and post object
-// establishment operations.
-func WithHooks(h Hooks) ReconcilerOption {
+// WithRuntimeHooks specifies how the Reconciler should perform pre and post
+// object establishment operations.
+func WithRuntimeHooks(h RuntimeHooks) ReconcilerOption {
```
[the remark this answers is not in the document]
turkenh: Updated the comment as follows, hope this clarifies:

```
// WithRuntimeHooks specifies how the Reconciler should perform preparations
// (pre- and post-establishment) and cleanup (deactivate) for package runtime.
// The hooks are only used when the package has a runtime and the runtime is
// configured as Deployment.
``` 

Also we have the following in the revision reconciler setup:

```
	if o.PackageRuntime == controller.PackageRuntimeDeployment {
		ro = append(ro, WithRuntimeHooks(NewProviderHooks(mgr.GetClient())))
	}
```

internal/controller/pkg/revision/runtime_defaults.go
turkenh: Yes, it could be doable, but I wanted to be consistent with the pattern we're following in the project: functional options instead of builders.

Also, please note, that this function no longer exists in the latest version in https://github.com/crossplane/crossplane/pull/4744

PR #4777, opened 2023-10-11
internal/xcrd: limit claim and composite name length in OpenAPI to actual values

### LINKED ISSUE #4645: Claim name can be max 57 chars long

sttts:

This PR adds a `maxLength` constraint on the `metadata.name` of claims and XRs through OpenAPI (kube-apiserver allows that for the name).

Fixes #4645

I have:

### DIFF
--- a/internal/xcrd/crd_test.go
+++ b/internal/xcrd/crd_test.go
@@ -32,7 +32,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/utils/pointer"
+	"k8s.io/utils/ptr"
 
 	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
 	"github.com/crossplane/crossplane-runtime/pkg/errors"
@@ -352,7 +352,7 @@ func TestForCompositeResource(t *testing.T) {
 															Required: []string{"apiVersion", "kind"},
 														},
 													},
-													XListType: pointer.String("atomic"),
+													XListType: ptr.To("atomic"),
 												},
 												"publishConnectionDetailsTo": {
 													Type:     "object",
@@ -418,7 +418,7 @@ func TestForCompositeResource(t *testing.T) {
 												"conditions": {
 													Description:  "Conditions of the resource.",
 													Type:         "array",
-													XListType:    pointer.String("map"),
+													XListType:    ptr.To("map"),
 													XListMapKeys: []string{"type"},
 													Items: &extv1.JSONSchemaPropsOrArray{
 														Schema: &extv1.JSONSchemaProps{
@@ -623,7 +623,7 @@ func TestForCompositeResource(t *testing.T) {
 															Required: []string{"apiVersion", "kind"},
 														},
 													},
-													XListType: pointer.String("atomic"),
+													XListType: ptr.To("atomic"),
 												},
 												"publishConnectionDetailsTo": {
 													Type:     "object",
@@ -682,7 +682,7 @@ func TestForCompositeResource(t *testing.T) {
 												"conditions": {
 													Description:  "Conditions of the resource.",
 													Type:         "array",
-													XListType:    pointer.String("map"),
+													XListType:    ptr.To("map"),
 													XListMapKeys: []string{"type"},
 													Items: &extv1.JSONSchemaPropsOrArray{
 														Schema: &extv1.JSONSchemaProps{
@@ -1721,7 +1721,7 @@ func TestForCompositeResourceClaim(t *testing.T) {
 												"conditions": {
 													Description:  "Conditions of the resource.",
 													Type:         "array",
-													XListType:    pointer.String("map"),
+													XListType:    ptr.To("map"),
 													XListMapKeys: []string{"type"},
 													Items: &extv1.JSONSchemaPropsOrArray{
 														Schema: &extv1.JSONSchemaProps{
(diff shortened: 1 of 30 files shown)


negz (approved):

pedjak (commented):
we need e2e tests

internal/xcrd/crd_test.go
```diff
@@ -1164,16 +1152,62 @@ func TestForCompositeResourceClaim(t *testing.T) {
...
-		t.Fatalf("ForCompositeResourceClaim(...): %s", err)
+	type test struct {
+		name   string
+		schema string
+		want   *extv1.CustomResourceDefinition
+	}
+	tests := []test{
+		{name: "standard case", schema: fmt.Sprintf(schemaFmt, ""), want: want(63)},
+		{name: "small name restriction", schema: fmt.Sprintf(schemaFmt, `"metadata":{"properties":{"name":{"type":"string","maxLength":10}}},`), want: want(10)},
+		{name: "too large name restrictions", schema: fmt.Sprintf(schemaFmt, `"metadata":{"properties":{"name":{"type":"string","maxLength":100}}},`), want: want(63)},
```
phisco, line 1166: nit: took me a bit to understand the `want: want(63)` part, maybe a more explicit name for the `want` function could be helpful, also because we usually have a `want` struct defined to wrap the wanted outcome.

internal/xcrd/crd_test.go
```diff
@@ -1164,16 +1152,62 @@ func TestForCompositeResourceClaim(t *testing.T) {
...
+		}
 	}
 
-	got, err := ForCompositeResourceClaim(d)
-	if err != nil {
-		t.Fatalf("ForCompositeResourceClaim(...): %s", err)
+	type test struct {
+		name   string
+		schema string
+		want   *extv1.CustomResourceDefinition
```
phisco, line 1161: We usually make test cases a map with CamelCase names as keys and call `reason` what you are here setting as `name`. We also define `want` and `args` struct to wrap the desired outcome and provided inputs. Would it be possible to adhere to the standard?

internal/xcrd/crd_test.go
sttts: done

internal/xcrd/crd_test.go
sttts: I like that pattern.

sttts:
> we need e2e tests

No. This depends on standard kube-apiserver functionality. We don't test kube here.

sttts:
@phisco ptal

phisco (approved):

PR #4780, opened 2023-10-11
Generate XR name locally by appending random suffix to the claim name

### LINKED ISSUE #4651: Take API server out of XR name generation path

pedjak:

There is no need to make a round-trip to k8s api server to get XR name, we
have enough data on the controller side to construct an unique XR name, by
using the convention:

   `<CLAIM_NAME>-<5_CHAR_RANDOM_SUFFIX>`

Change details:
   * removed `APIDryRunCompositeConfigurator`
   * updated `configurator_test.go`
   * updated and added more tests in `reconciler_test.go`
   

Fixes #4651

I have:

### DIFF
--- a/internal/controller/apiextensions/claim/configurator.go
+++ b/internal/controller/apiextensions/claim/configurator.go
@@ -23,6 +23,7 @@ import (
 
 	"dario.cat/mergo"
 	"github.com/google/go-cmp/cmp"
+	"k8s.io/apiserver/pkg/storage/names"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
@@ -40,33 +41,21 @@ const (
 	errUnsupportedDstObject = "destination object was not valid object"
 	errUnsupportedSrcObject = "source object was not valid object"
 
-	errName                  = "cannot use dry-run create to name composite resource"
-	errBindCompositeConflict = "cannot bind composite resource that references a different claim"
-
 	errMergeClaimSpec   = "unable to merge claim spec"
 	errMergeClaimStatus = "unable to merge claim status"
 )
 
-// An APIDryRunCompositeConfigurator configures composite resources. It may
-// perform a dry-run create against an API server in order to name and validate
-// the configured resource.
-type APIDryRunCompositeConfigurator struct {
-	client client.Client
-}
+var (
+	// ErrBindCompositeConflict can occur if the composite refers a different claim
+	ErrBindCompositeConflict = errors.New("cannot bind composite resource that references a different claim")
+)
 
-// NewAPIDryRunCompositeConfigurator returns a Configurator of composite
-// resources that may perform a dry-run create against an API server in order to
-// name and validate the configured resource.
-func NewAPIDryRunCompositeConfigurator(c client.Client) *APIDryRunCompositeConfigurator {
-	return &APIDryRunCompositeConfigurator{client: c}
-}
-
-// Configure the supplied composite resource by propagating configuration from
-// the supplied claim. Both create and update scenarios are supported; i.e. the
-// composite may or may not have been created in the API server when passed to
-// this method. The configured composite may be submitted to an API server via a
-// dry run create in order to name and validate it.
-func (c *APIDryRunCompositeConfigurator) Configure(ctx context.Context, cm resource.CompositeClaim, cp resource.Composite) error { //nolint:gocyclo // Only slightly over (12).
+// ConfigureComposite configures the supplied composite resource
+// by propagating configuration from the supplied claim.
+// Both create and update scenarios are supported; i.e. the
+// composite may or may not have been created in the API server
+// when passed to this method.
+func ConfigureComposite(_ context.Context, cm resource.CompositeClaim, cp resource.Composite) error { //nolint:gocyclo // Only slightly over (12).
 	ucm, ok := cm.(*claim.Unstructured)
 	if !ok {
 		return nil
@@ -85,7 +74,7 @@ func (c *APIDryRunCompositeConfigurator) Configure(ctx context.Context, cm resou
 	existing := ucp.GetClaimReference()
 	proposed := ucm.GetReference()
 	if existing != nil && !cmp.Equal(existing, proposed) {
-		return errors.New(errBindCompositeConflict)
+		return ErrBindCompositeConflict
 	}
 
 	// It's possible we're being asked to configure a statically provisioned
@@ -144,13 +133,26 @@ func (c *APIDryRunCompositeConfigurator) Configure(ctx context.Context, cm resou
 	ucp.SetClaimReference(proposed)
 
 	if !meta.WasCreated(cp) {
-		// The API server returns an available name derived from
-		// generateName when we perform a dry-run create. This name is
-		// likely (but not guaranteed) to be available when we create
-		// the composite resource. If the API server generates a name
-		// that is unavailable it will return a 500 ServerTimeout error.
-		cp.SetGenerateName(fmt.Sprintf("%s-", cm.GetName()))
-		return errors.Wrap(c.client.Create(ctx, cp, client.DryRunAll), errName)
+		// The composite was not found in the informer cache,
+		// or in the apiserver watch cache,
+		// or really does not exist.
+		// If the claim contains the composite reference,
+		// try to use it to set the composite name.
+		// This protects us against stale caches:
+		// 1. If the composite exists, but the cache was not up-to-date,
+		//    then its creation is going to fail, and after requeue,
+		//    the cache eventually gets up-to-date and everything is good.
+		// 2. If the composite really does not exist, it means that
+		//    the claim got bound in one of previous loop,
+		//    but something went wrong at composite creation and we requeued.
+		//    It is alright to try to use the very same name again.
+		if ref := cm.GetResourceReference(); ref != nil &&
+			ref.APIVersion == ucp.GetAPIVersion() && ref.Kind == ucp.GetKind() {
+			cp.SetName(ref.Name)
+			return nil
+		}
+		// Otherwise, generate name with a random suffix, hoping it is not already taken
+		cp.SetName(names.SimpleNameGenerator.GenerateName(fmt.Sprintf("%s-", cm.GetName())))
 	}
 
 	return nil
--- a/internal/controller/apiextensions/claim/reconciler.go
+++ b/internal/controller/apiextensions/claim/reconciler.go
@@ -61,6 +61,7 @@ const (
 	errConfigureComposite = "cannot configure composite resource"
 	errBindComposite      = "cannot bind composite resource"
 	errApplyComposite     = "cannot apply composite resource"
+	errCreateComposite    = "cannot create composite resource"
 	errConfigureClaim     = "cannot configure composite resource claim"
 	errPropagateCDs       = "cannot propagate connection details from composite"
 
@@ -88,6 +89,8 @@ func ControllerName(name string) string {
 
 // A Configurator configures the supplied resource, typically either populating the
 // composite with fields from the claim, or claim with fields from composite.
+// It returns ErrBindCompositeConflict if the composite is already bound to
+// a different claim and cannot be rebound.
 type Configurator interface {
 	Configure(ctx context.Context, cm resource.CompositeClaim, cp resource.Composite) error
 }
@@ -210,7 +213,7 @@ type crComposite struct {
 
 func defaultCRComposite(c client.Client) crComposite {
 	return crComposite{
-		Configurator:         NewAPIDryRunCompositeConfigurator(c),
+		Configurator:         ConfiguratorFn(ConfigureComposite),
 		ConnectionPropagator: NewAPIConnectionPropagator(c),
 	}
 }
@@ -477,6 +480,19 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		if kerrors.IsConflict(err) {
 			return reconcile.Result{Requeue: true}, nil
 		}
+		if errors.Is(err, ErrBindCompositeConflict) {
+			// the claim refers to a composite belonging to a different claim
+			// this case can occur if:
+			// 1. composite name gets generated
+			// 2. claim sets and persists the reference to composite with the generated name
+			// 3. composite creation fails because the generated name is already taken
+			// 4. in the next reconcile loop we get the above conflict
+			// to unblock us, we need to remove the composite reference at the claim
+			// otherwise, we can move forward even if we requeue
+			cm.SetResourceReference(nil)
+			_ = r.client.Update(ctx, cm)
+			return reconcile.Result{Requeue: true}, nil
+		}
 		err = errors.Wrap(err, errConfigureComposite)
 		record.Event(cm, event.Warning(reasonCompositeConfigure, err))
 		cm.SetConditions(xpv1.ReconcileError(err))
@@ -505,8 +521,20 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		return reconcile.Result{Requeue: true}, errors.Wrap(r.client.Status().Update(ctx, cm), errUpdateClaimStatus)
 	}
 
-	err := r.client.Apply(ctx, cp, resource.AllowUpdateIf(func(old, obj runtime.Object) bool { return !cmp.Equal(old, obj) }))
+	var err error
+	if meta.WasCreated(cp) {
+		err = r.client.Apply(ctx, cp, resource.AllowUpdateIf(func(old, obj runtime.Object) bool { return !cmp.Equal(old, obj) }))
+	} else {
+		// if composite did not exist at the beginning of the loop, we want to create it
+		// so that we can check if the composite name is not already taken
+		err = r.client.Create(ctx, cp)
+	}
 	switch {
+	case kerrors.IsAlreadyExists(err):
+		// generated name is already taken
+		// let's requeue and try to recover in the next round
+		log.Debug("Cannot create composite, another already exists with the generated name. Requeuing to try again with a new name.")
+		return reconcile.Result{Requeue: true}, nil
 	case resource.IsNotAllowed(err):
 		log.Debug("Skipped no-op composite resource apply")
 	case err != nil:
--- a/internal/controller/apiextensions/claim/configurator_test.go
+++ b/internal/controller/apiextensions/claim/configurator_test.go
@@ -18,6 +18,7 @@ package claim
 
 import (
 	"context"
+	"fmt"
 	"strings"
 	"testing"
 
@@ -43,7 +44,6 @@ func TestCompositeConfigure(t *testing.T) {
 	apiVersion := "v"
 	kind := "k"
 	now := metav1.Now()
-	errBoom := errors.New("boom")
 
 	type args struct {
 		ctx context.Context
@@ -57,10 +57,11 @@ func TestCompositeConfigure(t *testing.T) {
 	}
 
 	cases := map[string]struct {
-		reason string
-		c      client.Client
-		args   args
-		want   want
+		reason      string
+		diffOptions []cmp.Option
+		extraAssert func(args args, t *testing.T)
+		args        args
+		want        want
 	}{
 		"ClaimNotUnstructured": {
 			reason: "We should return early if the claim is not unstructured",
@@ -161,73 +162,11 @@ func TestCompositeConfigure(t *testing.T) {
 						},
 					},
 				},
-				err: errors.New(errBindCompositeConflict),
-			},
-		},
-		"DryRunError": {
-			reason: "We should return any error we encounter while dry-run creating a dynamically provisioned composite",
-			c: &test.MockClient{
-				MockCreate: test.NewMockCreateFn(errBoom),
-			},
-			args: args{
-				cm: &claim.Unstructured{
-					Unstructured: unstructured.Unstructured{
-						Object: map[string]any{
-							"apiVersion": apiVersion,
-							"kind":       kind,
-							"metadata": map[string]any{
-								"namespace": ns,
-								"name":      name,
-							},
-							"spec": map[string]any{
-								"coolness": 23,
-
-								// These should be preserved.
-								"compositionRef":      "ref",
-								"compositionSelector": "ref",
-
-								// These should be filtered out.
-								"resourceRef":                "ref",
-								"writeConnectionSecretToRef": "ref",
-							},
-						},
-					},
-				},
-				cp: &composite.Unstructured{},
-			},
-			want: want{
-				cp: &composite.Unstructured{
-					Unstructured: unstructured.Unstructured{
-						Object: map[string]any{
-							"metadata": map[string]any{
-								"generateName": name + "-",
-								"labels": map[string]any{
-									xcrd.LabelKeyClaimNamespace: ns,
-									xcrd.LabelKeyClaimName:      name,
-								},
-							},
-							"spec": map[string]any{
-								"coolness":            23,
-								"compositionRef":      "ref",
-								"compositionSelector": "ref",
-								"claimRef": map[string]any{
-									"apiVersion": apiVersion,
-									"kind":       kind,
-									"namespace":  ns,
-									"name":       name,
-								},
-							},
-						},
-					},
-				},
-				err: errors.Wrap(errBoom, errName),
+				err: ErrBindCompositeConflict,
 			},
 		},
 		"ConfiguredNewXR": {
 			reason: "A dynamically provisioned composite resource should be configured according to the claim",
-			c: &test.MockClient{
-				MockCreate: test.NewMockCreateFn(nil),
-			},
 			args: args{
 				cm: &claim.Unstructured{
 					Unstructured: unstructured.Unstructured{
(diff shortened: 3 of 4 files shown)


internal/controller/apiextensions/claim/configurator.go
```diff
@@ -136,13 +121,7 @@ func (c *APIDryRunCompositeConfigurator) Configure(ctx context.Context, cm resou
...
 
 	if !meta.WasCreated(cp) {
-		// The API server returns an available name derived from
-		// generateName when we perform a dry-run create. This name is
-		// likely (but not guaranteed) to be available when we create
-		// the composite resource. If the API server generates a name
-		// that is unavailable it will return a 500 ServerTimeout error.
-		cp.SetGenerateName(fmt.Sprintf("%s-", cm.GetName()))
-		return errors.Wrap(c.client.Create(ctx, cp, client.DryRunAll), errName)
+		cp.SetName(names.SimpleNameGenerator.GenerateName(fmt.Sprintf("%s-", cm.GetName())))
```
sttts, line 124: does a dry-run request verify that the name is not taken? I.e. do we lose anything with this?

sttts: If the answer is yes, i.e. dry-run verifies, then this change means that
- before we risked only conflict when an equally named ob
ject was created between claim update and XP creation.
- now we risk conflict when any of the names conflict: 23 bits in 5 characters => more than 98% risk for 10k objects.

Bildschirmfoto 2023-10-11 um 15 45 26

sttts: Ways out: check whether the name is taken, either with a client (roundtrip again, but 🤷) or checking the cache.

pedjak: > does a dry-run request verify that the name is not taken? I.e. do we lose anything with this?

The [docs](https://kubernetes.io/docs/reference/using-api/api-concepts/#generated-values) advises not to use that, i.e. we should not rely on the generated names.

Looking at the code, I would say that in the dry-run case, we are not going to get an error if the generated name already exists: 

* apiserver uses the same name generator to generate `metadata.name`: https://github.com/kubernetes/apiserver/blob/bf71493219c2bffe59c0b8c8b497ca0aa6e91a5a/pkg/registry/generic/registry/store.go#L408
* later on, [`Storage.Create`](https://github.com/kubernetes/apiserver/blob/bf71493219c2bffe59c0b8c8b497ca0aa6e91a5a/pkg/registry/generic/registry/store.go#L448) is called, passing also potential dry-run option.
* in case of dry-run, `DryRunnableStorage` handles it like this: https://github.com/kubernetes/apiserver/blob/d2d2fc882803682a397b74a9a0635fb6c1ffeff1/pkg/registry/generic/registry/dryrun.go#L40-L45
  * retrieve the value under the given key from the storage
  * merge the stored value with the submitted one

> does a dry-run request verify that the name is not taken? I.e. do we lose anything with this?

Given the above reasoning, dry-run request does not verify it, and the change proposed in this PR have the equal semantic.

pedjak: > Ways out: check whether the name is taken, either with a client (roundtrip again, but 🤷) or checking the cache.

I think this is not needed: when a composite cannot be created, because the name is taken, we requeue - in the next iteration a new name will be generated that is hopefully not taken.

sttts: > https://github.com/kubernetes/apiserver/blob/d2d2fc882803682a397b74a9a0635fb6c1ffeff1/pkg/registry/generic/registry/dryrun.go#L40-L45

Means that it does check, no?

sttts: > Means that it does check, no?

Looking at https://github.com/kubernetes/apiserver/blob/bf71493219c2bffe59c0b8c8b497ca0aa6e91a5a/pkg/registry/generic/registry/store.go#L450C3-L450C70, after calling `Create` of the dry-runnable storage, which then returns AlreadyExists, this `err = rest.CheckGeneratedNameError(ctx, e.CreateStrategy, err, obj)` call turns the AlreadyExists into [a conflict error with 1s retry](https://github.com/kubernetes/apiserver/blob/master/pkg/registry/rest/create.go#L168). The client will do this retry and cause another name being selected, until there is no conflict.

Hence, I think we cannot move forward here without checking the names in the cache to at least not regress.

sttts: > I think this is not needed: when a composite cannot be created, because the name is taken, we requeue - in the next iteration a new name will be generated that is hopefully not taken.

Missed that comment.

If it succeeds to create the XR, but then naturally requeues, the informer might be lagging and your get request at the top will return NotFound, leading to another creation aka a leak.

sttts: This is in the binder:
```
	// We refuse to 're-bind' a claim that is already bound to a different
	// composite resource.
	if existing != nil && !equal {
		return errors.New(errBindClaimConflict)
	}
```
Doesn't it mean that this rebinding on requeue is actually not working, not even in the old code?

pedjak: > Hence, I think we cannot move forward here without checking the names in the cache to at least not regress.

Oh, I missed that code on client retry when the name is already taken, thanks for the clarification :)

> If it succeeds to create the XR, but then naturally requeues, the informer might be lagging and your get request at the top will return NotFound, leading to another creation aka a leak.

Hm, what is the scenario? I thought about this:

1. Initially, there is no XR, just the claim
2. we generate composite name and the reference to it into the claim
3. the claim gets updated
4. composite gets created
5. if everything successful, we do not requeue - the reconcile will be triggered by claim/composite change

Are you saying that  if step 4 is successful, the next reconcile will not see it and `client.Get` could return NotFound?

Error scenarios:
* step 3 fails, then no XR gets created and we can requeue and retry
* step 4 failed,no XR gets created, and by the next reconcile  `existing` would be `nil`, so the new name will be generated and the new ref set in the claim

What this PR really needs is just to call `client.Create` in cases when composite does not exist. Right now we use `client.Apply` both for creation and updates.

sttts: > Are you saying that if step 4 is successful, the next reconcile will not see it and client.Get could return NotFound?

yes. Reads are either backed by the cache (because controller-runtime does that automatically), or they are gets against the apiserver. But for the later, you have to add `resourceVersion=0` in order to get a quorum read from etcd. Otherwise, it is also backed by a cache (watch cache). So a 404 can happen, yes.

> if everything successful, we do not requeue

We do: we update both the claim and the xr. Both trigger a requeue (in the case of the xr of course without 404).

sttts: > What this PR really needs is just to call client.Create

Yes, would very much prefer to be very explicit in what we are doing, not hiding it behind some apply wrapper.

pedjak:
@sttts updated PR to handle situations when the generated name is already taken, ptal

internal/controller/apiextensions/claim/configurator.go
```diff
@@ -46,26 +47,12 @@ const (
...
-func NewAPIDryRunCompositeConfigurator(c client.Client) *APIDryRunCompositeConfigurator {
-	return &APIDryRunCompositeConfigurator{client: c}
-}
-
-// Configure the supplied composite resource by propagating configuration from
+// ConfigureComposite configures the supplied composite resource by propagating configuration from
 // the supplied claim. Both create and update scenarios are supported; i.e. the
 // composite may or may not have been created in the API server when passed to
 // this method. The configured composite may be submitted to an API server via a
 // dry run create in order to name and validate it.
```
sttts, line 54: the dry-run create is gone I guess

internal/controller/apiextensions/claim/reconciler.go
```diff
@@ -494,6 +496,22 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+			// this case can occur if:
+			// 1. composite name gets generated
+			// 2. claim sets and persists the reference to composite with generated name
+			// 3. composite creation fails because the generated name is already taken
+			// 4. in the next reconcile loop we get the above conflict
+			// to unblock us, we need to remove composite reference at claim
+			// otherwise, we can move forward even if we requeue
+			cm.SetResourceReference(nil)
+			if err := r.client.Update(ctx, cm); err != nil {
+				return reconcile.Result{}, errors.Wrap(err, errUpdateClaim)
```
sttts, line 510: see @phisco's PR handling conflict errors

internal/controller/apiextensions/claim/reconciler.go
```diff
@@ -494,6 +496,22 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+			}
+			cm.SetConditions(xpv1.ReconcileError(err))
```
sttts, line 512: would do this silently, not even set a condition. It's not an error, but an to be expected conflict.

internal/controller/apiextensions/claim/reconciler.go
```diff
@@ -520,8 +538,23 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+		err = r.client.Create(ctx, cp)
+	}
 	switch {
+	case kerrors.IsAlreadyExists(err):
+		// generated name is already taken
+		// let's requeue and try again with the new name
+		log.Debug("Cannot create composite, another already exists with that name. Requeue and try another name")
+		err = errors.Wrap(err, errCreateComposite)
+		record.Event(cm, event.Warning(reasonCompositeConfigure, err))
+		cm.SetConditions(xpv1.ReconcileError(err))
```
sttts, line 556: would stay silent. No event, no conditions. This is part of an to be expected conflict. Nothing to report to the user.

sttts:
Logic looks good :+1: All comments are about cosmetics, less events and conditions and go docs to be fixed.

pedjak:
@sttts I have addressed your comments, ptal

internal/controller/apiextensions/claim/reconciler.go
```diff
@@ -530,8 +544,20 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+		err = r.client.Apply(ctx, cp, resource.AllowUpdateIf(func(old, obj runtime.Object) bool { return !cmp.Equal(old, obj) }))
+	} else {
+		// if composite did not exist at the beginning of the loop, we want to create it
+		// so that we can check if the composite name is not already taken
+		err = r.client.Create(ctx, cp)
+	}
 	switch {
+	case kerrors.IsAlreadyExists(err):
+		// generated name is already taken
+		// let's requeue and try again with the new name
```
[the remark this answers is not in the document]
pedjak: oh, yes, I hope it is ok now.

sttts (approved):
For the reconciler logic.

internal/controller/apiextensions/claim/configurator.go
```diff
@@ -46,26 +47,11 @@ const (
...
-
-// NewAPIDryRunCompositeConfigurator returns a Configurator of composite
-// resources that may perform a dry-run create against an API server in order to
-// name and validate the configured resource.
-func NewAPIDryRunCompositeConfigurator(c client.Client) *APIDryRunCompositeConfigurator {
-	return &APIDryRunCompositeConfigurator{client: c}
-}
-
-// Configure the supplied composite resource by propagating configuration from
+// ConfigureComposite configures the supplied composite resource by propagating configuration from
```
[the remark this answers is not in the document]
pedjak: Done. Do we wrap comments at 80 chars in general, while the code lines could be longer? If yes, can we add some lint rules for this?

internal/controller/apiextensions/claim/configurator.go
```diff
@@ -134,13 +120,26 @@ func (c *APIDryRunCompositeConfigurator) Configure(ctx context.Context, cm resou
...
 
 	if !meta.WasCreated(cp) {
-		// The API server returns an available name derived from
-		// generateName when we perform a dry-run create. This name is
-		// likely (but not guaranteed) to be available when we create
-		// the composite resource. If the API server generates a name
-		// that is unavailable it will return a 500 ServerTimeout error.
-		cp.SetGenerateName(fmt.Sprintf("%s-", cm.GetName()))
-		return errors.Wrap(c.client.Create(ctx, cp, client.DryRunAll), errName)
+		// the composite was not found in the informer cache,
```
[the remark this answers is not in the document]
pedjak: Done, let me know if further improvements are needed.

internal/controller/apiextensions/claim/reconciler.go
```diff
@@ -501,6 +502,19 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		if kerrors.IsConflict(err) {
 			return reconcile.Result{Requeue: true}, nil
 		}
+		if err.Error() == errBindCompositeConflict {
```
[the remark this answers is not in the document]
pedjak: Sure, instead of creating a dedicated error type, I used k8s conflict error, IMHO it matches good here, but let me know if you disagree.

internal/controller/apiextensions/claim/reconciler.go
```diff
@@ -501,6 +502,19 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+			// the claim refers to a composite belonging to a different claim
+			// this case can occur if:
+			// 1. composite name gets generated
+			// 2. claim sets and persists the reference to composite with the generated name
+			// 3. composite creation fails because the generated name is already taken
+			// 4. in the next reconcile loop we get the above conflict
+			// to unblock us, we need to remove the composite reference at the claim
+			// otherwise, we can move forward even if we requeue
+			cm.SetResourceReference(nil)
+			r.client.Update(ctx, cm) //nolint:errcheck // we requeue in any case, no need to check for the error
```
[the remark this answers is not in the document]
pedjak: done.

internal/controller/apiextensions/claim/reconciler.go
```diff
@@ -530,8 +544,20 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+	} else {
+		// if composite did not exist at the beginning of the loop, we want to create it
+		// so that we can check if the composite name is not already taken
+		err = r.client.Create(ctx, cp)
+	}
 	switch {
+	case kerrors.IsAlreadyExists(err):
+		// generated name is already taken
+		// let's requeue and try to recover in the next round
+		log.Debug("Cannot create composite, another already exists with that name. Requeue and try to recover")
```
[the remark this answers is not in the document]
pedjak: done.

internal/controller/apiextensions/claim/reconciler.go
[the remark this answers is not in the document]
sttts: Don't think that's a good idea. It's brittle. If somebody adds some client call in that func it might have unexpected consequences. Better be super precise. Conflicts are defined errors from an apiserver, borrowing them for client code is not good.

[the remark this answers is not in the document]
pedjak: thanks @sttts for the feedback. Added global err variable.

internal/controller/apiextensions/claim/reconciler.go
```diff
@@ -501,6 +502,19 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		if kerrors.IsConflict(err) {
 			return reconcile.Result{Requeue: true}, nil
 		}
+		if errors.Is(err, errBindCompositeConflict) {
```
sttts, line 505: As long as `composite.Configurator` is a public interface and we handle this error here, the error must be public too and well documented as part of the interface.

Sidenote: am not convinced that `composite.Configurator` is of value as a public interface. Seems to be a very leaky abstraction.

pedjak: > As long as composite.Configurator is a public interface and we handle this error here, the error must be public too and well documented as part of the interface.

Done.

sttts: :+1:

internal/controller/apiextensions/claim/reconciler.go
```diff
@@ -475,6 +478,19 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		if kerrors.IsConflict(err) {
```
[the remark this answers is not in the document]
pedjak: Probably not - because some other implementation can return it, as @sttts pointed out that earlier.

internal/controller/apiextensions/claim/reconciler.go
```diff
@@ -503,8 +519,20 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+	} else {
+		// if composite did not exist at the beginning of the loop, we want to create it
+		// so that we can check if the composite name is not already taken
+		err = r.client.Create(ctx, cp)
+	}
 	switch {
+	case kerrors.IsAlreadyExists(err):
+		// generated name is already taken
+		// let's requeue and try to recover in the next round
+		log.Debug("Cannot create composite, another already exists with the generated name. Requeuing to try again with a new name.")
```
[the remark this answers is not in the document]
pedjak: composite name is already part of log context. And the error is always the same - object already exist.

internal/controller/apiextensions/claim/configurator.go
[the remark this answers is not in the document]
negz: Yes we do. Happy to add linter rules if there's a linter that will do this.

negz:
Would it make sense for this to use the NameGenerator added in https://github.com/crossplane/crossplane/pull/4858?

pedjak:
> Would it make sense for this to use the NameGenerator added in https://github.com/crossplane/crossplane/pull/4858?

  I do not think so, it does not add any additional value here, it is just a wrapper around `SimpleNameGenerator` so that @sttts can invoke `client.Get` method to check if the generated name is already taken.

internal/controller/apiextensions/claim/reconciler.go
```diff
@@ -477,6 +480,19 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		if kerrors.IsConflict(err) {
 			return reconcile.Result{Requeue: true}, nil
 		}
+		if errors.Is(err, ErrBindCompositeConflict) {
+			// the claim refers to a composite belonging to a different claim
+			// this case can occur if:
```
turkenh, line 485: There could other cases where this could occur and I don't think we should handle all like this, i.e. reset the reference. 
One example use case is, the user may intentionally get into this situation while moving a claim between namespaces without impacting the composite, see https://github.com/crossplane/crossplane/discussions/4081

Considering the probability of hitting this due to a name conflict is low, I would propose just erroring out as before after making sure we did all we can do to decrease that probability to its minimum. Given that we are not hearing this as a pain from users with the current implementation, this feels like a good tradeoff to me.

internal/controller/apiextensions/claim/reconciler.go
```diff
@@ -505,8 +521,20 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+	var err error
+	if meta.WasCreated(cp) {
+		err = r.client.Apply(ctx, cp, resource.AllowUpdateIf(func(old, obj runtime.Object) bool { return !cmp.Equal(old, obj) }))
+	} else {
+		// if composite did not exist at the beginning of the loop, we want to create it
+		// so that we can check if the composite name is not already taken
+		err = r.client.Create(ctx, cp)
+	}
 	switch {
+	case kerrors.IsAlreadyExists(err):
```
turkenh, line 533: Just nits: 

- this could be thrown only from the Create above, so bugging me to have here for both.
- Would it be possible to convert the  above `else` to an [early return](https://github.com/crossplane/crossplane/tree/master/contributing#return-early) somehow 🤔

turkenh (commented):
AFAIU, there are two different problems that this PR aims to solve:

1. Do not rely on Kube API (via dry-run) to generate a random name for the composite of a given claim => https://github.com/crossplane/crossplane/issues/4651
2. Handle the edge case where the name is taken between the generation and the actual create call.

These problems existed with both `claim -> composite` and `composite -> composed`. https://github.com/crossplane/crossplane/pull/4858 solved the first problem for `composite -> composed`.

The reason why we found ourselves dealing with the 2nd problem instead of just fixing the 1st one (which is what https://github.com/crossplane/crossplane/issues/4651 asks for) is that, we realized that we increased the probability of the 2nd with the solution we put in this PR, whereas, https://github.com/crossplane/crossplane/pull/4858 claims that the approach used there didn't increase the probability of 2nd compared to the existing solution.

Is my understanding correct?

If yes, would it be ok to apply the approach in https://github.com/crossplane/crossplane/pull/4858 here and leave 2nd problem out of scope for this specific PR? We can discuss it separately with a dedicated issue and decide on whether we can live without it or should fix it. 

I can also understand if we want to discuss here as well given there is already an implementation.

internal/controller/apiextensions/claim/reconciler.go
sttts: We must be careful here. The name generation as implemented increases the probability of conflict considerably. Dry-run (and my implementation) does an ad-hoc freeness check of the generated name. That reduced probability of conflicts to claims in-flight. The code here though does not check freeness and hence you get conflicts between all existing claims and the new claim. Over the runtime of a controlplane with many claim creations this will lead to a broken claim eventually (somewhere I did the math, 50% probability for a thousand claim or so).

Hence, either we keep full conflict resolution here or we switch to a freeness check. Otherwise, we regress.

negz:
> If yes, would it be ok to apply the approach in https://github.com/crossplane/crossplane/pull/4858 here and leave 2nd problem out of scope for this specific PR? We can discuss it separately with a dedicated issue and decide on whether we can live without it or should fix it.

I agree with @turkenh. If I understand correctly:

* In #4858 @sttts introduces a `NameGenerator` pattern. This generator uses the API server's generateName logic to randomly generate a name, then makes a best-effort attempt to ensure its available before creation.
* In this PR @pedjak also uses the API server's generateName logic to randomly generate a name. Instead of making a best-effort attempt to ensure the generated name is available before creation, in this PR @pedjak relies on requeues to retry on name collisions. A side-effect of this approach is that sometimes we must revert state (e.g. revert a persisted reference to the XR).

Is there a reason to use a different approach in this controller from the approach we take in the XR controller? If not, I would prefer us to solve the problem consistently (no pun intended 😄).

pedjak:
@turkenh @negz 

> Is my understanding correct?

This PR primary solves https://github.com/crossplane/crossplane/issues/4651 - it generates composite names using the same logic available on the API server side, but without making any requests to API server. In this way, we avoid unnecessary communication with the API server, lowering the number of request that API server needs to process.

In the case of composite creation we save one request per claim, but in the case of managed resource creation, we could avoid sending a significant number of requests, depending how many managed resources are defined within a composition ([real examples](https://github.com/upbound/configuration-caas/blob/main/apis/aws/eks/composition.yaml) easily declare 10-20 managed resources).

https://github.com/crossplane/crossplane/pull/4858 does generate the names locally, but checks their availability by issuing `Get` request to the API server, because `Get` requests for unstructured objects are not cached. Thus, we have replaced a dry-run `patch` request with a `get` one, but we did not reduce the number of calls going to API server.

The logic here checks for name availability at the composite creation - if the name is taken, the create request fails, and we are going to requeue, generate a new name and try to create the composite again.

`NameGenerator` introduced in https://github.com/crossplane/crossplane/pull/4858 does not give us any guaranties that the name is not going to be taken before we try to create the composite. Given that we use `patch` request for that under the hood, there is a chance that we are not going to create a new composite - instead we are going to update an existing one and even worse we are going to hijack it from another claim. Probability of such event is not high, but if it happens, it can have even security implications, because suddenly a claim gets attached to composite that could belong to a different team/organization/tenant.

This PR solves the described problem as well.

If unnecessary communication with the API server is not something we are concerned with (both claim and composite controller), this PR can be closed. For the consistency, `NameGenerator` can be applied on the claim controller side, but the possibility to hijack another composite still remains.

turkenh:
> If yes, would it be ok to apply the approach in #4858 here and leave 2nd problem out of scope for this specific PR? We can discuss it separately with a dedicated issue and decide on whether we can live without it or should fix it.

As proposed above, we decided not to proceed with this PR due to the complications it is introducing, especially the necessity to always do a [migration](https://github.com/crossplane/crossplane/pull/4896#discussion_r1406233166) in the first call after creation. So, currently we decided to:

- Split out and focus on the actual work for using SSA in Claim -> Composite (which is now [merged](https://github.com/crossplane/crossplane/pull/4896) 🎉 ). For example, an updated version of this PR would make it more clear that what are the exact changes we need to fix https://github.com/crossplane/crossplane/issues/5104 instead of getting things mixed up with SSA.
- Hold off with the approach here until we see clear signals on https://github.com/crossplane/crossplane/issues/5104 and https://github.com/crossplane/crossplane/issues/5122 which existed since the inceptions of Compositions and reevaluate if we can find a better solution (e.g. does not require continuous migration). 

@pedjak feel free to close this PR or keep it open by updating against latest master. I don't think it is a priority right now, but feel free to chose whatever makes more sense for you.

Issue #4783, opened 2023-10-11
Deduplicate xpkg subcommands in the Crossplane CLI


negz:
### What happened?

With things starting to fall into place in the new Crossplane CLI we've got a few redundant commands:

```
$ crossplane --help
Usage: crossplane <command>

A command line tool for interacting with Crossplane.

Commands:
  build              Build Crossplane packages.
    configuration    Build a Configuration package.
    provider         Build a Provider package.
    function         Build a Function package.

  install            Install Crossplane packages.
    configuration    Install a Configuration package.
      <package>      Image containing Configuration package.
      [<name>]       Name of Configuration.
    provider         Install a Provider package.
      <package>      Image containing Provider package.
      [<name>]       Name of Provider.

  update             Update Crossplane packages.
    configuration    Update a Configuration package.
      <name>         Name of Configuration.
      <tag>          Updated tag for Configuration package.
    provider         Update a Provider package.
      <name>         Name of Provider.
      <tag>          Updated tag for Provider package.

  push               Push Crossplane packages.
    configuration    Push a Configuration package.
      <tag>          Tag of the package to be pushed. Must be a valid OCI image tag.
    provider         Push a Provider package.
      <tag>          Tag of the package to be pushed. Must be a valid OCI image tag.

  xpkg               Crossplane package management.
    build            Build a package, by default from the current directory.
    push             Push Crossplane packages.
      <tag>          Tag of the package to be pushed. Must be a valid OCI image tag.

  alpha              Alpha features. WARN: May be changed or removed without notice

  beta               Beta features. WARN: May be changed or deprecated in a future release

Flags:
  -h, --help                    Show context-sensitive help.
  -v, --version=VERSION-FLAG    Print version and quit.
      --verbose                 Print verbose logging statements.

Run "crossplane <command> --help" for more information on a command.
```

Notably I think the `crossplane xpkg ...` commands are newer, better implementations of `crossplane build ...` and `crossplane push ...`.

### How can we reproduce it?

Build the CLI from 11c8aecace8a12a5f71dda368bb5eb7f16194a2d and run it. :)

### What environment did it happen in?
Crossplane version: 11c8aecace8a12a5f71dda368bb5eb7f16194a2d

negz:
I would suggest we:

* Remove the `build`, `install`, `update`, and `push` subcommands.
* Add new `crossplane xpkg install` and `crossplane xpkg update` subcommands.

phisco:
[...]

PR #4799, opened 2023-10-13
feat(crank): rename build controller flags to base image

### LINKED ISSUE #4784: Rename `crossplane xpkg build --controller ...` flags

```
$ crossplane xpkg build --help
Usage: crossplane xpkg build

Build a package, by default from the current directory.

The build command creates a xpkg compatible OCI image for a Crossplane package from the local file system. It packages the found YAML files containing Kubernetes-like object manifests into the meta data
layer of the OCI image. The package manager will use this information to install the package into a Crossplane instance.

Only configuration, provider and function packages are supported at this time.

Example claims can be specified in the examples directory.

For more generic information, see the xpkg parent command help. Also see the Crossplane documentation for more information on building packages:

    https://docs.crossplane.io/latest/concepts/packages/#building-a-package

Even more details can be found in the xpkg reference document.

Flags:
  -h, --help                          Show context-sensitive help.
  -v, --version=VERSION-FLAG          Print version and quit.
      --verbose                       Print verbose logging statements.

  -o, --output=STRING                 Path for package output.
      --controller=STRING             Controller image used as base for package.
      --controller-tar=STRING         Path to tar file, an alternative to Controller.
  -f, --package-root="."              Path to package directory.
  -e, --examples-root="./examples"    Path to package examples directory.
      --ignore=IGNORE,...             Paths, specified relative to --package-root, to exclude from the package.
```

The above command talks about a "controller" image to bake into a package. This makes sense in the context of Providers, where a package can be part of an OCI image that also contains a Kubernetes controller binary. For Functions, this makes less sense. A package can be part of an OCI image that contains a Function binary, but that binary is not a (Kubernetes) controller.

We could find a more generic name for this flag. Perhaps `--runtime` per https://github.com/crossplane/crossplane/issues/4699. `--image` was also suggested in https://github.com/crossplane-contrib/function-patch-and-transform/pull/18. (Or maybe both - `--runtime-image`.)

phisco:

Fixes https://github.com/crossplane/crossplane/issues/4784.

Removed `xor` tag as it didn't actually work, just made the tar file be used instead of the image if both are specified. Looks like `xor` file actually needs to be specified on all flags that have to be in a mutually exclusive group, but even switching to the correct usage, it didn't enforce the two flags from being used for some reason I couldn't understand.

I have:

### DIFF
--- a/cmd/crank/xpkg/build.go
+++ b/cmd/crank/xpkg/build.go
@@ -39,14 +39,14 @@ import (
 )
 
 const (
-	errGetNameFromMeta            = "failed to get package name from crossplane.yaml"
-	errBuildPackage               = "failed to build package"
-	errImageDigest                = "failed to get package digest"
-	errCreatePackage              = "failed to create package file"
-	errParseControllerImage       = "failed to parse controller image"
-	errPullControllerImage        = "failed to pull controller image"
-	errLoadControllerTar          = "failed to load controller tar"
-	errGetControllerBaseImageOpts = "failed to get controller base image options"
+	errGetNameFromMeta         = "failed to get package name from crossplane.yaml"
+	errBuildPackage            = "failed to build package"
+	errImageDigest             = "failed to get package digest"
+	errCreatePackage           = "failed to create package file"
+	errParseRuntimeBaseImage   = "failed to parse runtime base image"
+	errPullRuntimeBaseImage    = "failed to pull runtime base image"
+	errLoadRuntimeBaseImageTar = "failed to load runtime base image tarball"
+	errGetRuntimeBaseImageOpts = "failed to get runtime base image options"
 )
 
 // AfterApply constructs and binds context to any subcommands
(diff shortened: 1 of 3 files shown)


cmd/crank/xpkg/build.go
```diff
@@ -98,12 +98,12 @@ type buildCmd struct {
...
 	root    string
 
-	Output        string   `optional:"" short:"o" help:"Path for package output."`
-	Controller    string   `help:"Controller image used as base for package."`
-	ControllerTar string   `help:"Path to tar file, an alternative to Controller." xor:"controller" type:"existingfile"`
-	PackageRoot   string   `short:"f" help:"Path to package directory." default:"."`
-	ExamplesRoot  string   `short:"e" help:"Path to package examples directory." default:"./examples"`
-	Ignore        []string `help:"Paths, specified relative to --package-root, to exclude from the package."`
+	Output              string   `optional:"" short:"o" help:"Path for package output."`
+	RuntimeBaseImage    string   `help:"Name of the Docker image to be used as base for the package, e.g. 'xpkg.upbound.io/user/repo:v0.0.1'. Overwritten by '--runtime-base-image-tar'." `
```
turkenh, line 102:
```suggestion
	RuntimeBaseImage    string   `help:"Name of the Docker image to be used as base for the package, typically runtime image, e.g. 'xpkg.upbound.io/user/repo:v0.0.1'. Overwritten by '--runtime-base-image-tar'." `
```

turkenh: FWIW, I am not feeling great about `--runtime-base-image`. I believe it should be either `--base-image` if we name it per implementation or `--runtime-image` if we name it per expected usage. `--runtime-base-image` looks like an odd mix of both views.

So, either:

```go
	BaseImage    string   `help:"Name of the Docker image to be used as the base layer for the package, typically runtime image'. Overwritten by '--base-image-tar'." `
```

or 

```go
	RuntimeImage    string   `help:"Name of the Runtime Docker image to be included into package image as the base layer, applicable for packages with a runtime (e.g. Providers and Functions)'. Overwritten by '--runtime-image-tar'." `
```

I have a slight preference over the former.

phisco: I agree, I would vote for --base-image, as I feel users mostly care and is important for them to understand that a single image is going to be produced and used if they did not specify differently in the meta definition

negz: My vote is for `--runtime-image`. ;) It's more specific - we're building a Package around/containing a runtime, not just any old base image.

phisco: I don't really get what you mean by "contains a runtime". Runtime used like that sounds to me as the Java or Python runtime, but I guess it's not what you mean. I would get saying it's "containing an image you can actually run", but your phrasing makes me doubt my interpretation.

However, we could postpone this discussion to when we get an agreement on https://github.com/crossplane/crossplane/issues/4783, as if we end up with `crossplane <noun> build`, we could decide to use specific flags meaningful for each package type.

negz: A runtime is what we're running with to describe "the code that powers a package". i.e.:

* A Provider controller is a Provider's runtime.
* A Function server/runner is a Function's runtime.

I know runtime is an overloaded term, but I think it's too late in the game to debate whether it makes sense. We're already committed via the "runtime config" concept, which configures how these things are deployed.

phisco: Runtime config still made sense to me without this explanation, I read that as "the configuration to run a package" (for the kind of packages it make sense to do so).

phisco:
I will close this one and do the changes after the refactoring from https://github.com/crossplane/crossplane/issues/4783

PR #4804, opened 2023-10-13
Add ability to pause reconciliation of instances of all core types

### LINKED ISSUE #4619: "paused" annotation not being respected for configurations
### What happened?
It appears that the "crossplane.io/paused: "true"" annotation is not being respected for configurations. 

### How can we reproduce it?
To reproduce this issue, follow these steps: create a configuration, add the "paused" annotation, and manually update a composition. After a reconciliation cycle of the configuration, the composition reverts back to its previous state.

Crossplane version: v1.13.1

pedjak:

The reconciliation can be paused by adding `crossplane.io/paused: "true"`
to the requested object. Removing the annotation resumes its reconciliation.

Fixes #4619

I have:


pedjak:
@plumbis once this gets merged, we should also update the docs.

phisco:
Looks like the e2es are failing on the upgrade test in some interesting way 🤔

internal/controller/apiextensions/composition/reconciler.go
```diff
@@ -148,6 +151,22 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		"spec-hash", currentHash,
 	)
 
+	// Check the pause annotation and return if it has the value "true"
+	// after logging, publishing an event and updating the SYNC status condition
+	if meta.IsPaused(comp) {
```
negz, line 156: The idea of pausing a Composition seems unintuitive/surprising to me. I believe this means we just won't create new CompositionRevisions, right? I don't think the Composition controller does anything else.

pedjak: > The idea of pausing a Composition seems unintuitive/surprising to me. I believe this means we just won't create new CompositionRevisions, right? I don't think the Composition controller does anything else.

I was just following #4619 requirement. If pause reconciliation does not make sense for some of core types, I would be more than happy to revert the change. @haarchri @ytsarev your input here would be very valuable.

pedjak: thanks @ytsarev , is there any other core type that we should not pause? This PR introduce pausing ability on all core types, as requested in #4619 .

turkenh: I agree, there is no known need for pausing annotations on compositions, and could even be confusing as @ytsarev pointed out. 

I am wondering if we should keep this PR focused on Package Manager types only, which satisfies the [actual requirement](https://github.com/crossplane/crossplane/issues/4619#issuecomment-1721383822) and consider adding it to the rest in the future if needed? Because I don't think pausing annotations would be helpful for the Usage type or in RBAC manager controllers as well.

turkenh:
With the current implementation, pausing the `Package` (e.g. `Provider`, `Function`, or `Configuration`) wouldn't automatically pause the `Revision` (e.g. `ProviderRevision`, `FunctionRevision, or `ConfigurationRevision`). 

So, for the [main use case ](https://github.com/crossplane/crossplane/issues/4619#issuecomment-1721383822), users would need to pause the `Configuration` and the active `ConfigurationRevision` which might not be intuitive. I still believe this should be the way to go (as we discussed @pedjak) to be consistent and predictable, but it still will not provide the best UX. 

So it is worth explicitly calling this out in the PR description and eventually in the documentation so that folks can be aware and chime in if they disagree. /cc @ytsarev

internal/controller/apiextensions/composition/reconciler.go
pedjak: >  in RBAC manager controllers as well.

There is RBAC controller for ProviderRevision, hence I assume that that reconciliation should be paused as well, if we allow pausing `ProviderRevisions` because we would not like to change RBAC rules, correct?

pedjak:
@ytsarev and @haarchri is pausing Configuration the only thing that make sense? If yes, could you please update the title of #4619 to reflect that? Should we have some sort of automation around pausing `Revisions` or user should pause them explicitly? What should be user experience?

pedjak:
@turkenh if I understand you correct, we should remove from this PR pausing the following types:
* XRD
* Usage
* Composition

Anything else?

turkenh:
> @turkenh if I understand you correct, we should remove from this PR pausing the following types:
> 
> * XRD
> * Usage
> * Composition
> 
> Anything else?

@pedjak anything other than the package manager types (for now), e.g. anything not under `internal/controller/pkg/`

pedjak:
Closing it in favor of #4820 - please continue the review there.

PR #4812, opened 2023-10-14
feat: switch probes port to 8081

phisco:

We just introduced probes in #4748 on port 5000, unfortunately port 5000 on MacOS is [used](https://stackoverflow.com/questions/72369320/why-always-something-is-running-at-port-5000-on-my-mac) if not disabled explicitly, to avoid useless conflicts when running crossplane locally we should just switch to another port. Picked 8081 as suggested by @turkenh given that we expose metrics on port 8080.

I didn't notice it because I disabled the service listening on port 5000 a while back 😬 

I have:

### DIFF
--- a/cluster/charts/crossplane/templates/deployment.yaml
+++ b/cluster/charts/crossplane/templates/deployment.yaml
@@ -133,7 +133,7 @@ spec:
             port: healthz
         ports:
         - name: healthz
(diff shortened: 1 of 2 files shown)


turkenh (approved):
Thanks @phisco!

PR #4820, opened 2023-10-16
Add ability to pause reconciliation of Provider/Configuration/Function instances

### LINKED ISSUE #4619: "paused" annotation not being respected for configurations

pedjak:

The reconciliation can be paused by adding `crossplane.io/paused: "true"`
to the requested object. Removing the annotation resumes its reconciliation.

Fixes #4619 

I have:

### DIFF
--- a/internal/controller/apiextensions/claim/reconciler.go
+++ b/internal/controller/apiextensions/claim/reconciler.go
@@ -65,6 +65,8 @@ const (
 	errPropagateCDs       = "cannot propagate connection details from composite"
 
 	errUpdateClaimStatus = "cannot update composite resource claim status"
+
+	reconcilePausedMsg = "Reconciliation (including deletion) is paused via the pause annotation"
 )
 
 // Event reasons.
@@ -371,8 +373,8 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 	// Check the pause annotation and return if it has the value "true"
 	// after logging, publishing an event and updating the SYNC status condition
 	if meta.IsPaused(cm) {
-		r.record.Event(cm, event.Normal(reasonPaused, "Reconciliation is paused via the pause annotation"))
-		cm.SetConditions(xpv1.ReconcilePaused())
+		r.record.Event(cm, event.Normal(reasonPaused, reconcilePausedMsg))
+		cm.SetConditions(xpv1.ReconcilePaused().WithMessage(reconcilePausedMsg))
 		// If the pause annotation is removed, we will have a chance to reconcile again and resume
 		// and if status update fails, we will reconcile again to retry to update the status
 		return reconcile.Result{}, errors.Wrap(r.client.Status().Update(ctx, cm), errUpdateClaimStatus)
--- a/internal/controller/pkg/manager/reconciler.go
+++ b/internal/controller/pkg/manager/reconciler.go
@@ -32,6 +32,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
+	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
 	"github.com/crossplane/crossplane-runtime/pkg/errors"
 	"github.com/crossplane/crossplane-runtime/pkg/event"
 	"github.com/crossplane/crossplane-runtime/pkg/logging"
@@ -52,6 +53,8 @@ const (
 	// updated content for the given package reference. This behavior is only
 	// enabled when the packagePullPolicy is Always.
 	pullWait = 1 * time.Minute
+
+	reconcilePausedMsg = "Reconciliation (including deletion) is paused via the pause annotation"
 )
 
 func pullBasedRequeue(p *corev1.PullPolicy) reconcile.Result {
@@ -85,6 +88,7 @@ const (
 	reasonTransitionRevision event.Reason = "TransitionRevision"
 	reasonGarbageCollect     event.Reason = "GarbageCollect"
 	reasonInstall            event.Reason = "InstallPackageRevision"
+	reasonPaused             event.Reason = "ReconciliationPaused"
 )
 
 // ReconcilerOption is used to configure the Reconciler.
@@ -277,6 +281,19 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		// we'll be requeued implicitly because we return an error.
 		log.Debug(errGetPackage, "error", err)
 		return reconcile.Result{}, errors.Wrap(resource.IgnoreNotFound(err), errGetPackage)
+	}
+
+	// Check the pause annotation and return if it has the value "true"
+	// after logging, publishing an event and updating the SYNC status condition
+	if meta.IsPaused(p) {
+		r.record.Event(p, event.Normal(reasonPaused, reconcilePausedMsg))
+		p.SetConditions(xpv1.ReconcilePaused().WithMessage(reconcilePausedMsg))
+		// If the pause annotation is removed, we will have a chance to reconcile again and resume
+		// and if status update fails, we will reconcile again to retry to update the status
+		return reconcile.Result{}, errors.Wrap(r.client.Status().Update(ctx, p), errUpdateStatus)
+	}
+	if c := p.GetCondition(xpv1.ReconcilePaused().Type); c.Reason == xpv1.ReconcilePaused().Reason {
+		p.CleanConditions()
 	}
 
 	// Get existing package revisions.
--- a/internal/controller/pkg/revision/reconciler.go
+++ b/internal/controller/pkg/revision/reconciler.go
@@ -101,6 +101,8 @@ const (
 	errCannotBuildMetaSchema         = "cannot build meta scheme for package parser"
 	errCannotBuildObjectSchema       = "cannot build object scheme for package parser"
 	errCannotBuildFetcher            = "cannot build fetcher for package parser"
+
+	reconcilePausedMsg = "Reconciliation (including deletion) is paused via the pause annotation"
 )
 
 // Event reasons.
@@ -110,6 +112,7 @@ const (
 	reasonDependencies event.Reason = "ResolveDependencies"
 	reasonSync         event.Reason = "SyncPackage"
 	reasonDeactivate   event.Reason = "DeactivateRevision"
+	reasonPaused       event.Reason = "ReconciliationPaused"
 )
 
 // ReconcilerOption is used to configure the Reconciler.
@@ -458,6 +461,16 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		"name", pr.GetName(),
 	)
 
+	// Check the pause annotation and return if it has the value "true"
+	// after logging, publishing an event and updating the SYNC status condition
+	if meta.IsPaused(pr) {
+		r.record.Event(pr, event.Normal(reasonPaused, reconcilePausedMsg))
+		pr.SetConditions(xpv1.ReconcilePaused().WithMessage(reconcilePausedMsg))
+		// If the pause annotation is removed, we will have a chance to reconcile again and resume
+		// and if status update fails, we will reconcile again to retry to update the status
+		return reconcile.Result{}, errors.Wrap(r.client.Status().Update(ctx, pr), errUpdateStatus)
+	}
+
 	if meta.WasDeleted(pr) {
 		// NOTE(hasheddan): In the event that a pre-cached package was
 		// used for this revision, delete will not remove the pre-cached
@@ -495,6 +508,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 			return reconcile.Result{}, err
 		}
 		return reconcile.Result{Requeue: false}, nil
+	}
+
+	if c := pr.GetCondition(xpv1.ReconcilePaused().Type); c.Reason == xpv1.ReconcilePaused().Reason {
+		pr.CleanConditions()
 	}
 
 	if err := r.revision.AddFinalizer(ctx, pr); err != nil {
--- a/internal/controller/rbac/provider/binding/reconciler.go
+++ b/internal/controller/rbac/provider/binding/reconciler.go
@@ -134,7 +134,7 @@ type Reconciler struct {
 
 // Reconcile a ProviderRevision by creating a ClusterRoleBinding that binds a
 // provider's service account to its system ClusterRole.
-func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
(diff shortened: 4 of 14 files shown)


internal/controller/pkg/manager/reconciler.go
```diff
@@ -285,6 +287,20 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+	// after logging, publishing an event and updating the SYNC status condition
+	if meta.IsPaused(p) {
+		log.Debug("Reconciliation is paused via the pause annotation", "annotation", meta.AnnotationKeyReconciliationPaused, "value", "true")
+		r.record.Event(p, event.Normal(reasonPaused, "Reconciliation is paused via the pause annotation"))
+		p.SetConditions(xpv1.ReconcilePaused())
+		// If the pause annotation is removed, we will have a chance to reconcile again and resume
+		// and if status update fails, we will reconcile again to retry to update the status
+		return reconcile.Result{}, errors.Wrap(r.client.Status().Update(ctx, p), errUpdateStatus)
+	}
+	if c := p.GetCondition(xpv1.ReconcilePaused().Type); c.Reason == xpv1.ReconcilePaused().Reason {
```
turkenh, line 300: I think we don't need this. 

After noticing that this is different than [managed reconciler](https://github.com/crossplane/crossplane-runtime/blob/89e89660b5c6237f14be6563e7b1f104a9ac3589/pkg/reconciler/managed/reconciler.go#L758-L767), I found this discussion which explains why we don't need to clean the conditions up: https://github.com/crossplane/crossplane-runtime/pull/357#discussion_r984071756

internal/controller/pkg/revision/reconciler.go
```diff
@@ -468,6 +469,20 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		return reconcile.Result{Requeue: false}, nil
 	}
 
+	// Check the pause annotation and return if it has the value "true"
+	// after logging, publishing an event and updating the SYNC status condition
+	if meta.IsPaused(pr) {
```
turkenh, line 474: We should move this up before `meta.wasDeleted` check so that the pause will be effective for package revisins being deleted. This also aligns with how the feature was implemented in the managed reconciler and claim reconciler.

internal/controller/rbac/provider/binding/reconciler.go
```diff
@@ -163,6 +163,13 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		return reconcile.Result{Requeue: false}, nil
 	}
 
+	// Check the pause annotation and return if it has the value "true"
+	// after logging, publishing an event and updating the SYNC status condition
+	if meta.IsPaused(pr) {
```
turkenh, line 168: Same here, please move above `meta.WasDeleted`

internal/controller/rbac/provider/roles/reconciler.go
```diff
@@ -256,6 +256,13 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		return reconcile.Result{Requeue: false}, nil
 	}
 
+	// Check the pause annotation and return if it has the value "true"
+	// after logging, publishing an event and updating the SYNC status condition
+	if meta.IsPaused(pr) {
```
turkenh, line 261: Same here, please move above `meta.WasDeleted`

turkenh (commented):
Thanks @pedjak!

Do you already have plans for documenting this? I think we need something similar to [this](https://docs.crossplane.io/latest/concepts/managed-resources/#paused) but for packages. Please also note, there is an ongoing work with package documentation and may be we can include [there](https://github.com/crossplane/docs/pull/566) (or at least build on top of that)

internal/controller/pkg/revision/reconciler.go
pedjak: Meaning that we cannot remove paused objects?

pedjak:
> Do you already have plans for documenting [this](https://docs.crossplane.io/latest/concepts/managed-resources/#paused)? I think we need something similar to this but for packages. Please also note, https://github.com/crossplane/docs/pull/566 is an ongoing work with package documentation and may be we can include there (or at least build on top of that)

We should certainly document this after it gets merged. /cc @plumbis

internal/controller/pkg/revision/reconciler.go
turkenh: Yes, deleted objects would still need reconciliation to finalize their deletion. 
One can remove a paused object after removing the paused annotation.

internal/controller/pkg/manager/reconciler.go
turkenh: Ah ok, I missed that package types not having a `Synced` condition. I am wondering if introducing a `Synced` annotation could make sense, but something to consider later.

internal/controller/pkg/revision/reconciler.go
pedjak: Is that really what users expect to happen? In [the docs](https://docs.crossplane.io/latest/concepts/managed-resources/#paused) we do not say anything about being impossible to remove paused objects.

Imagine the following situation: a user attempts to remove a namespace containing a paused claim. This operation would hangs forever, right?

turkenh: > Is that really what users expect to happen?

As said, deletion is also part of reconciliation, the controller needs some operation / clean-up to finalize deletion. I would be surprised that part of reconciliation was not "paused". And that could be more desirable if I want to debug before the resource is deleted.

Another point is, one can go and find why their resource not deleted if that is the behavior they are looking for and remove the annotation to proceed. But if we do the opposite, there wouldn't be a way to revert for a user expecting the deletion to hang on with paused annotation, as the resource would already be gone.

> Imagine the following situation: a user attempts to remove a namespace containing a paused claim. This operation would hangs forever, right?

They can see why it hangs and remove the paused annotation to proceed. 
I don't think this is a feature that people would end up having paused resources around. It is more like a deliberate action where people would revert after they get what they are after.

phisco: I agree with @turkenh, pausing should pause everything, it would be surprising otherwise.

sttts: I tend to be on @pedjak's side here. Broken deletion seems to be unexpected and strange.

pedjak: Just tried the following:
* created namespace `example`
* created claim in it
* paused claim
* tried to remove namespace with `kubectl delete ns example`
* `kubectl` hangs with the message `namespace "example" deleted` but it does not return to the prompt

How a user can know that there is a claim in the namespace holding the delete?

Even if I do `kubectl describe` on that claim, there is no trace saying that deletion is blocked and can be resumed when pause annotation is removed.

Users do not need to know anything about how things are working under the hood, how many controllers are processing objects, in what order, etc. However, when users want to remove an object, we should either protect against that with a proper error message, or allow him to do that. If an object reconciliation is paused, we can still allow removal, cleaning up what is needed, etc.

turkenh: Let me link the discussion on slack: https://crossplane.slack.com/archives/CEF5N8X08/p1697618937115119?thread_ts=1697555426.091559&cid=CEF5N8X08

> Even if I do `kubectl describe` on that claim, there is no trace saying that deletion is blocked and can be resumed when pause annotation is removed.

This is something different. There should be. Shouldn't this output some events/logs: https://github.com/crossplane/crossplane/blob/af2496ebacd6227f19be3b1fb8af58100a4d8b2e/internal/controller/apiextensions/claim/reconciler.go#L371-L379

As I stated in the thread, it feels like we are trying to optimize user experience for a very specific scenario while ignoring others who would expect the reconciliation to be paused, including the deletion.

FWIW, it is also implemented in the same way in cluster API, which could be considered as an indication that this is what most people expect from such a functionality, see: [https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/b90810b70864664a00314f40a3f8a31d4215f1b5/controllers/awscluster_controller.go#L[…]4](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/b90810b70864664a00314f40a3f8a31d4215f1b5/controllers/awscluster_controller.go#L165-L194)

pedjak: > This is something different. There should be. Shouldn't this output some events/logs: 

The events are gone, because when you try to remove a namespace, k8s starts removing all of them in the given namespace, including events. Given that there is no problems to remove them (no finalizers, no blockers), the events are all gone.

> it feels like you are trying to optimize user experience for a very specific scenario while ignoring others who would expect the reconciliation to be paused, including the deletion.

Who are the others, i.e. what is the usecase we want to support?

We indicate that reconciliation is paused via condition of type `Synced` like this:

```yaml
reason: ReconcilePaused
status: "False"
type: Synced
```

when `False` it means that we are not syncing claim changes with the counterpart composite, i.e. we are not passing claim updates down the line. However, object deletion is not an update, at least not in CRUD/REST context, and therefore an user should be very surprised to discover that deletion of paused objects or higher order objects get stuck.

In order to [explore/learn](https://github.com/crossplane/crossplane/issues/4839) more about the use case mentioned by @turkenh I have updated the PR to match the behavior already available for claims/composites. Furthermore, I have added to the pause status condition a human readable message explain the consequences, and in the case of deletion we generate an additional k8s event and a log entry on info level.

apis/apiextensions/v1/conditions.go
```diff
@@ -86,3 +89,12 @@ func TerminatingClaim() xpv1.Condition {
 		Reason:             ReasonTerminatingClaim,
 	}
 }
+
+// ReconcilePaused returns paused condition with human readable message
+// TODO: ReconcilePaused from crossplane-runtime does not set the message
```
turkenh, line 94: I would suggest just opening a PR there, but given the feature freeze deadline, can you at least open an issue to follow up?

apis/apiextensions/v1/conditions.go
```diff
@@ -86,3 +89,12 @@ func TerminatingClaim() xpv1.Condition {
...
+// Remove this function after updating crossplane-runtime
+func ReconcilePaused() xpv1.Condition {
+	c := xpv1.ReconcilePaused()
+	c.Message = fmt.Sprintf("Reconciliation paused, object removal blocked. To unblock, remove %s annotation", meta.AnnotationKeyReconciliationPaused)
```
turkenh, line 98:
```suggestion
	c.Message = "Reconciliation (including deletion) is paused via the pause annotation"
```

internal/controller/apiextensions/claim/reconciler.go
```diff
@@ -372,7 +374,11 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 	// after logging, publishing an event and updating the SYNC status condition
 	if meta.IsPaused(cm) {
 		r.record.Event(cm, event.Normal(reasonPaused, "Reconciliation is paused via the pause annotation"))
-		cm.SetConditions(xpv1.ReconcilePaused())
+		cm.SetConditions(v1.ReconcilePaused())
+		if meta.WasDeleted(cm) {
+			log.Info("Deletion requested, but blocked by the pause annotation, remove it to unblock the request", "annotation", meta.AnnotationKeyReconciliationPaused, "value", "true")
+			r.record.Event(cm, event.Normal(reasonPaused, "Deletion requested, but blocked by the pause annotation, remove it to unblock the request"))
+		}
```
turkenh, line 381:
```suggestion
		r.record.Event(cm, event.Normal(reasonPaused, "Reconciliation (including deletion) is paused via the pause annotation"))
		cm.SetConditions(v1.ReconcilePaused())
```

It should be enough to update the message as above, no need for special handling for `meta.wasDeleted`.

After this message, it is clear that:

- Reconciliation is paused.
- This includes deletion.
- It is via pause annotation, so, you can remove it to resume. It is the same for both resuming deleted and non-deleted resources.

apis/apiextensions/v1/conditions.go
pedjak: Actually, I just realized that I do not need to modify crossplane-runtime: on condition there is `WithMessage()` method so the human readable message can be provided easily.

internal/controller/apiextensions/claim/reconciler.go
pedjak: Updated the PR with the suggested changes. I assumed I should apply it to all places in this PR. Please let me know if my understanding was correct.

> It should be enough to update the message as above, no need for special handling for meta.wasDeleted.

It should be enough, but this degrades the previously attempted user experience improvements. Namely:
* the logs are generated on DEBUG level, making `Reconciliation (including deletion) is paused via the pause annotation"` message invisible by default. The previous attempt wanted to bring them to INFO level, so that users can quickly spot them in the logs
* the current message is rather vague, refers to `pause annotation`. IMHO, we should be as much as explicit to users, so that they do not need to checks docs to understand what pause annotation is. Writing the exact annotation in the message is super helpful when one person pauses reconciliation, but another one performs some cluster maintenance/cleanup. In large orgs that is a common case. That person does not even need to know what Crossplane is, but the right message would unblock them quickly.
* Having an extra message that deletion is attempted on paused object would further help by debugging.

turkenh: Thanks!

> - the logs are generated on DEBUG level, making Reconciliation (including deletion) is paused via the pause annotation" message invisible by default. The previous attempt wanted to bring them to INFO level, so that users can quickly spot them in the logs

I agree that INFO would be better here 👍 

> - the current message is rather vague, refers to pause annotation. IMHO, we should be as much as explicit to users, so that they do not need to checks docs to understand what pause annotation is. Writing the exact annotation in the message is super helpful when one person pauses reconciliation, but another one performs some cluster maintenance/cleanup. In large orgs that is a common case. That person does not even need to know what Crossplane is, but the right message would unblock them quickly.

Users can just check the annotations on the resource and figure out which is the pause annotation without necessarily checking the documentation. However, being more explicit wouldn't hurt.

> - Having an extra message that deletion is attempted on paused object would further help by debugging.

Just like the above, I would agree with this in the context of being more explicit. However, I don't think this is worth introducing an additional branch (i.e. `if`) to the logic.

turkenh (approved):

internal/controller/pkg/manager/reconciler.go
```diff
@@ -279,6 +283,19 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		return reconcile.Result{}, errors.Wrap(resource.IgnoreNotFound(err), errGetPackage)
 	}
 
+	// Check the pause annotation and return if it has the value "true"
+	// after logging, publishing an event and updating the SYNC status condition
+	if meta.IsPaused(p) {
+		r.record.Event(p, event.Normal(reasonPaused, "Reconciliation is paused via the pause annotation"))
```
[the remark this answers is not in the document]
pedjak: done.

PR #4821, opened 2023-10-17
Stop propagation of claim's `*.kubernetes.io`/`*.k8s.io` annotations/labels down to XR

### LINKED ISSUE #4650: Stop propagating `kubectl.kubernetes.io/last-applied-configuration` annotation to XR

When a claim gets deployed using `kubectl`, its `kubectl.kubernetes.io/last-applied-configuration` annotation gets propagated to the corresponding composite resource, and this is misleading because:

* XR was not created using `kubectl`
* the content of the annotaton refers to the claim, not XR

Crossplane should stop propagating the annotation, perhaps even  all `*.kubernetes.io/*` [annotations](https://kubernetes.io/docs/reference/labels-annotations-taints/)

pedjak:

All `*.kubernetes.io`/`*.k8s.io` annotations/labels declared on a claim are not passed down to XR

Fixes #4650

I have:

### DIFF
--- a/internal/controller/apiextensions/claim/configurator.go
+++ b/internal/controller/apiextensions/claim/configurator.go
@@ -19,9 +19,11 @@ package claim
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"dario.cat/mergo"
 	"github.com/google/go-cmp/cmp"
+	"golang.org/x/exp/maps"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
@@ -92,7 +94,22 @@ func (c *APIDryRunCompositeConfigurator) Configure(ctx context.Context, cm resou
 	// external name.
 	en := meta.GetExternalName(ucp)
 
-	meta.AddAnnotations(ucp, ucm.GetAnnotations())
+	// Do not propagate *.kubernetes.io annotations down to the composite
+	// For example: when a claim gets deployed using kubectl,
+	// its kubectl.kubernetes.io/last-applied-configuration annotation
+	// should not be propagated to the corresponding composite resource,
+	// because:
+	// * XR was not created using kubectl
+	// * The content of the annotaton refers to the claim, not XR
+	// See https://kubernetes.io/docs/reference/labels-annotations-taints/
+	// for all annotations and their semantic
+	annot := ucm.GetAnnotations()
+	maps.DeleteFunc(annot, func(k string, _ string) bool {
+		s := strings.Split(k, "/")
+		return strings.HasSuffix(s[0], "kubernetes.io")
+	})
+	meta.AddAnnotations(ucp, annot)
+
 	meta.AddLabels(ucp, cm.GetLabels())
 	meta.AddLabels(ucp, map[string]string{
 		xcrd.LabelKeyClaimName:      ucm.GetName(),
--- a/internal/controller/apiextensions/claim/configurator.go
+++ b/internal/controller/apiextensions/claim/configurator.go
@@ -23,7 +23,6 @@ import (
 
 	"dario.cat/mergo"
 	"github.com/google/go-cmp/cmp"
-	"golang.org/x/exp/maps"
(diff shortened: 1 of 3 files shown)


internal/controller/apiextensions/claim/configurator.go
```diff
@@ -92,7 +94,22 @@ func (c *APIDryRunCompositeConfigurator) Configure(ctx context.Context, cm resou
...
+	// Do not propagate *.kubernetes.io annotations down to the composite
+	// For example: when a claim gets deployed using kubectl,
+	// its kubectl.kubernetes.io/last-applied-configuration annotation
+	// should not be propagated to the corresponding composite resource,
+	// because:
+	// * XR was not created using kubectl
+	// * The content of the annotaton refers to the claim, not XR
+	// See https://kubernetes.io/docs/reference/labels-annotations-taints/
+	// for all annotations and their semantic
+	annot := ucm.GetAnnotations()
```
turkenh, line 106: nit: we can use a shorter name here per https://github.com/crossplane/crossplane/tree/master/contributing#use-descriptive-variable-names-sparingly

```suggestion
	a := ucm.GetAnnotations()
```

internal/controller/apiextensions/claim/configurator.go
```diff
@@ -92,7 +94,22 @@ func (c *APIDryRunCompositeConfigurator) Configure(ctx context.Context, cm resou
...
 
-	meta.AddAnnotations(ucp, ucm.GetAnnotations())
+	// Do not propagate *.kubernetes.io annotations down to the composite
+	// For example: when a claim gets deployed using kubectl,
+	// its kubectl.kubernetes.io/last-applied-configuration annotation
+	// should not be propagated to the corresponding composite resource,
+	// because:
+	// * XR was not created using kubectl
+	// * The content of the annotaton refers to the claim, not XR
+	// See https://kubernetes.io/docs/reference/labels-annotations-taints/
```
turkenh, line 104: Do you think we should do the same for labels as well?

internal/controller/apiextensions/claim/configurator.go
```diff
@@ -92,7 +94,22 @@ func (c *APIDryRunCompositeConfigurator) Configure(ctx context.Context, cm resou
...
+	// for all annotations and their semantic
+	annot := ucm.GetAnnotations()
+	maps.DeleteFunc(annot, func(k string, _ string) bool {
```
turkenh, line 107: https://cs.opensource.google/go/x/exp

> This subrepository holds experimental and deprecated (in the old directory) packages.

This doesn't look ideal. Should we instead implement that simple functionality (delete keys from a map) instead of relying on an experimental package?

pedjak: That library was already project dependency, so I though it is ok to use it. Reverted.

internal/controller/apiextensions/claim/configurator.go
pedjak: Good idea, let's do it for both right away :)

internal/controller/apiextensions/claim/configurator.go
```diff
@@ -60,6 +61,16 @@ func NewAPIDryRunCompositeConfigurator(c client.Client) *APIDryRunCompositeConfi
 	return &APIDryRunCompositeConfigurator{client: c}
 }
 
+func withoutReservedK8sEntries(a map[string]string) map[string]string {
```
turkenh, line 64: nit: Should better be moved below `func (c *APIDryRunCompositeConfigurator) Configure` instead of appearing on top.

turkenh (approved):
Thanks @pedjak, looking good 🙌

internal/controller/apiextensions/claim/configurator.go
turkenh: Not sure how it ended up being there but couldn't find any previous usage in crossplane or crossplane-runtime.

internal/controller/apiextensions/claim/configurator.go
pedjak: done, take a look.

PR #4831, opened 2023-10-17
Update Helm chart to install Crossplane images from xpkg.upbound.io

### LINKED ISSUE #3560: Publish Crossplane image to xpkg.upbound.io

jbw976:

This PR simply updates Crossplane's Helm chart to install the Crossplane image from `xpkg.upbound.io`.

Fixes #3560

I have:

- [ ] ~Added or updated e2e tests, if necessary. (check with reviewers/maintainers if you're unsure whether E2E tests are necessary for the change).~
  - https://github.com/crossplane/docs/pull/572

### How has this PR been tested?

I tested these changes locally in a kind cluster by installing the local Helm chart with a `helm` client and setting the most recent version of the Crossplane image that has been published to `xpkg.upbound.io`: `v1.14.0-rc.0.611.g8121e6d6`:

```
❯ .cache/tools/darwin_arm64/helm-v3.13.1 install crossplane --namespace crossplane-system --create-namespace cluster/charts/crossplane --set image.tag=v1.14.0-rc.0.611.g8121e6d6 --set 'args={"--debug"}'
```

Note this doesn't set `image.repository`, so the new value of `xpkg.upbound.io/crossplane/crossplane` in the chart that is updated in this PR is used.

The Crossplane pods get to `Running` successfully:

```
❯ kcs get pod
NAME                                       READY   STATUS    RESTARTS   AGE
crossplane-7dd756b54d-jpxql                1/1     Running   0          29s
crossplane-rbac-manager-765d64bb64-wp4gw   1/1     Running   0          29s
```

And we can see that both pods are using the expected container image from `xpkg.upbound.io`:

```
❯ kcs get pod -o json | jq '.items[] | .metadata.name + " " + .spec.containers[].image'
"crossplane-7dd756b54d-jpxql xpkg.upbound.io/crossplane/crossplane:v1.14.0-rc.0.611.g8121e6d6"
"crossplane-rbac-manager-765d64bb64-wp4gw xpkg.upbound.io/crossplane/crossplane:v1.14.0-rc.0.611.g8121e6d6"
```

### DIFF
--- a/cluster/charts/crossplane/README.md
+++ b/cluster/charts/crossplane/README.md
@@ -78,7 +78,7 @@ and their default values.
 | `extraVolumesCrossplane` | Add custom `volumes` to the Crossplane pod. | `{}` |
 | `hostNetwork` | Enable `hostNetwork` for the Crossplane deployment. Caution: enabling `hostNetwork`` grants the Crossplane Pod access to the host network namespace. | `false` |
 | `image.pullPolicy` | The image pull policy used for Crossplane and RBAC Manager pods. | `"IfNotPresent"` |
-| `image.repository` | Repository for the Crossplane pod image. | `"crossplane/crossplane"` |
+| `image.repository` | Repository for the Crossplane pod image. | `"xpkg.upbound.io/crossplane/crossplane"` |
 | `image.tag` | The Crossplane image tag. Defaults to the value of `appVersion` in Chart.yaml. | `""` |
 | `imagePullSecrets` | The imagePullSecret names to add to the Crossplane ServiceAccount. | `{}` |
 | `leaderElection` | Enable [leader election](https://docs.crossplane.io/latest/concepts/pods/#leader-election) for the Crossplane pod. | `true` |
@@ -152,7 +152,7 @@ replicas: 1
 deploymentStrategy: RollingUpdate
 
 image:
-  repository: crossplane/crossplane
+  repository: xpkg.upbound.io/crossplane/crossplane
   tag: alpha
   pullPolicy: Always
 ```
--- a/cluster/charts/crossplane/README.md.gotmpl
+++ b/cluster/charts/crossplane/README.md.gotmpl
@@ -93,7 +93,7 @@ replicas: 1
 deploymentStrategy: RollingUpdate
 
 image:
-  repository: crossplane/crossplane
+  repository: xpkg.upbound.io/crossplane/crossplane
   tag: alpha
   pullPolicy: Always
 ```
--- a/cluster/charts/crossplane/values.yaml
+++ b/cluster/charts/crossplane/values.yaml
@@ -9,7 +9,7 @@ deploymentStrategy: RollingUpdate
 
 image:
   # -- Repository for the Crossplane pod image.
-  repository: crossplane/crossplane
+  repository: xpkg.upbound.io/crossplane/crossplane
   # -- The Crossplane image tag. Defaults to the value of `appVersion` in Chart.yaml.
   tag: ""
   # -- The image pull policy used for Crossplane and RBAC Manager pods.


cluster/charts/crossplane/README.md.gotmpl
```diff
@@ -93,7 +93,7 @@ replicas: 1
 deploymentStrategy: RollingUpdate
 
 image:
-  repository: crossplane/crossplane
+  repository: xpkg.upbound.io/crossplane/crossplane
```
jbw976, line 96: i'm not actually sure what this `README.md.gotmpl` file is used for or if there is any automation around it, any ideas?

cluster/charts/crossplane/README.md
```diff
@@ -78,7 +78,7 @@ and their default values.
 | `extraVolumesCrossplane` | Add custom `volumes` to the Crossplane pod. | `{}` |
 | `hostNetwork` | Enable `hostNetwork` for the Crossplane deployment. Caution: enabling `hostNetwork`` grants the Crossplane Pod access to the host network namespace. | `false` |
 | `image.pullPolicy` | The image pull policy used for Crossplane and RBAC Manager pods. | `"IfNotPresent"` |
-| `image.repository` | Repository for the Crossplane pod image. | `"crossplane/crossplane"` |
+| `image.repository` | Repository for the Crossplane pod image. | `"xpkg.upbound.io/crossplane/crossplane"` |
```
jbw976, line 81: @plumbis I think I'll need to open a docs PR to update https://github.com/crossplane/docs/blob/master/content/master/software/install.md - there's no automation keeping that file in sync with this README.md, is there?

cluster/charts/crossplane/README.md.gotmpl
phisco: Yes, it's automatically rendered to the actual README of the chart with all the values updated, this part feels a bit superfluous though, we could remove it

phisco (approved):

turkenh (approved):

cluster/charts/crossplane/README.md
jbw976: @plumbis confirmed that a separate manual PR is needed 🤓

Issue #4787, opened 2023-10-11
Finalize Crossplane CLI command format and structure


jbw976:
As part of https://github.com/crossplane/crossplane/issues/4648, we have a lot of new Crossplane CLI commands that will ship in v1.14.  We need to align and decide on the general format and structure of these commands.

Areas of consideration:

* **general command structure**
  * The current crossplane CLI was exclusively exposed as a `kubectl` plugin, so it consciously exposed its commands in the same format as `kubectl`, e.g. `kubectl crossplane <verb> <subject>`. From https://github.com/crossplane/crossplane/issues/4697, we are no longer tied to being just a `kubectl` plugin.
  * In general, what is the structure of all the commands we'll ship for v1.14?
* **subcommand per package type**
  * The current CLI exposes a specific subcommand for each type of package you can build (e.g. provider, configuration, and function), while `up` exposed just a general `build` command that works for all types of packages. We should decide which approach `crossplane xpkg build` should take.
* **alpha/beta commands**
  * Which commands that we are shipping in v1.14 should be included under the `alpha` or `beta` subcommands? Should function related commands be `beta` to match the overall XFN Beta maturity? Or is the CLI experiences/commands independently matured from the core feature? Should building a function be under `build` or under `function`? 

## Decisions

Decisions made in https://github.com/crossplane/crossplane/issues/4787#issuecomment-1766979905, resulting in the following command structure for v1.14:

* `crossplane`
  * `xpkg`
    * `login`
    * `build`
    * `push`
    * `install`
    * `update`
  * `beta`
    * `xpkg`
      * `init`
    * `render`
  * `alpha`
    * `describe`
    * `diagnose`

jbw976:
cross-post from @negz in https://github.com/crossplane/crossplane/pull/4736#issuecomment-1760531518

> WRT alpha vs beta - my slight preference is to keep everything Function related as "beta".

jbw976:
Below is a proposal for the general structure of all the commands we're hoping to ship in v1.14, in a hierarchical fashion to show command/subcommand relationships.

Not 💯 on this yet, so I may waffle more, but wanted to get this down to make progress together - feel free to provide feedback or suggest alternative proposals 😅 

### Conclusions

The following decisions naturally come out of this proposal:

* we are not tied to a strict `kubectl <verb> <subject>` format, other patterns can be used where/when they fit best
* prefer consistency in our `<verb> <subject>` ordering where possible
[...]

PR #4832, opened 2023-10-17
Rearrange the `crossplane` subcommands

negz:

Fixes https://github.com/crossplane/crossplane/issues/4787
Fixes https://github.com/crossplane/crossplane/issues/4784
Fixes https://github.com/crossplane/crossplane/issues/4783

Here's what it looks like:

```
$ go run cmd/crank/main.go --help
Usage: crossplane <command>

A command line tool for interacting with Crossplane.

Commands:
  xpkg login        Login to the default package registry (xpkg.upbound.io).
  xpkg logout       Logout of the default package registry (xpkg.upbound.io).
  xpkg build        Build a package, by default from the current directory.
  xpkg push         Push a package, by default to xpkg.upbound.io.
  xpkg install      Install a package.
  xpkg update       Update an installed package.
  beta xpkg init    Initialize a package from a template.
  beta render       Render a claim or XR locally.

Flags:
  -h, --help       Show context-sensitive help.
  -v, --version    Print version and quit.
      --verbose    Print verbose logging statements.

Run "crossplane <command> --help" for more information on a command.
```

Notable changes in this PR:

* Renamed `marketplace login` to `xpkg login`
* Renamed `xpkg build --controller` flag to `--embed-runtime-image`.
* Everything uses xpkg.upbound.io as the default registry
* Store xpkg.upbound.io auth config in `~/.crossplane` to avoid conflicts with `~/.up`.
* The kind of package to install (e.g. provider, function, etc) is now an argument to `xpkg install`, not a subcommand.
* You can use `xpkg init` to initialize a provider now too (or really anything - for now it's a tiny wrapper on `git clone`).

I have:

- [ ] ~Added or updated e2e tests, if necessary. (check with reviewers/maintainers if you're unsure whether E2E tests are necessary for the change).~


negz:
I personally found the tree view pretty ugly/hard to read with the new layout so I reverted that. Not sure if that will be controversial.

Some other options available to us are to skip expanding subcommands by default, which looks like:

```
Usage: crossplane <command>

A command line tool for interacting with Crossplane.

Commands:
  xpkg <command>
    Crossplane package management.

  beta <command>
    Beta features. WARN: May be changed or deprecated in a future release

Flags:
  -h, --help       Show context-sensitive help.
  -v, --version    Print version and quit.
      --verbose    Print verbose logging statements.

Run "crossplane <command> --help" for more information on a command.
```

Or to get compact help, which looks like:

```
$ go run cmd/crank/main.go --help
Usage: crossplane <command>

A command line tool for interacting with Crossplane.

Commands:
  xpkg login                    Login to the default package registry (xpkg.upbound.io).
  xpkg logout                   Logout of the default package registry (xpkg.upbound.io).
  xpkg build                    Build a package, by default from the current directory.
  xpkg push                     Push a package, by default to xpkg.upbound.io.
  xpkg install configuration    Install a Configuration package.
  xpkg install provider         Install a Provider package.
  xpkg update configuration     Update a Configuration package.
  xpkg update provider          Update a Provider package.
  beta xpkg init                Initialize a package from a template.
  beta xpkg install function    Install a Function package.
  beta xpkg update function     Update a Function package.
  beta render                   Render a claim or XR locally.

Flags:
  -h, --help       Show context-sensitive help.
  -v, --version    Print version and quit.
      --verbose    Print verbose logging statements.
```

jbw976:
I like the compact view the best! I agree that the tree view is now too big and unwieldy and it will only get worse with more commands that we add.  I do like having subcommands expanded so the user knows everything they can do up front without having to spelunk manually down through all the commands and get help for all of them.

Compact help seems to be a good balance of all that for now - we can always revisit if that gets too large in the future.

jbw976:
>  Make update and install take the kind of package as an argument, not a subcommand.

So this means that these commands would go away, right?

* `beta xpkg install function`
*  `beta xpkg update function`

cmd/crank/main.go
```diff
@@ -60,22 +59,9 @@ var cli struct {
...
 	Verbose verboseFlag `name:"verbose" help:"Print verbose logging statements."`
 
-	Install installCmd `cmd:"" help:"Install Crossplane packages."`
-	Update  updateCmd  `cmd:"" help:"Update Crossplane packages."`
-
 	XPKG xpkg.Cmd `cmd:"" help:"Crossplane package management."`
 
-	Marketplace marketplace.Cmd `cmd:"" help:"Interact with the Upbound Marketplace."`
-
-	Alpha struct {
```
jbw976, line 70: did you want to leave this alpha struct in for placeholder for the upcoming `describe`/`diagnose` commands?

cmd/crank/beta/xpkg/xpkg.go
```diff
@@ -0,0 +1,56 @@
...
+There are different kinds of Crossplane packages, each with a different set of
+meta information and files in the additional layer. The following kinds are 
+currently supported:
+
+- **Provider**: A Crossplane package that contains a Crossplane provider. The layer
+  contains a crossplane.yaml file with a "meta.pkg.crossplane.io/v1alpha1"
+  kind "Provider" manifest, and optionally CRD manifest.
+- **Configuration**: A Crossplane package that contains a Crossplane configuration,
+  with a "meta.pkg.crossplane.io/v1" kind "Configuration" manifest in crossplane.yaml.
+- in newer versions of Crossplane, more kinds will be supported.
```
jbw976, line 49: good place to add `**Function**` here too?

cmd/crank/beta/xpkg/xpkg.go
```diff
@@ -0,0 +1,56 @@
...
+how to represent the package in a user interface. This information is used by
+the Upbound marketplace to display packages and their contents. See the xpkg
+reference document for more information.
+
+There are different kinds of Crossplane packages, each with a different set of
+meta information and files in the additional layer. The following kinds are 
+currently supported:
+
+- **Provider**: A Crossplane package that contains a Crossplane provider. The layer
+  contains a crossplane.yaml file with a "meta.pkg.crossplane.io/v1alpha1"
```
jbw976, line 45: update this version to v1? and make sure the same updates are made to `cmd/crank/xpkg/xpkg.go` too?

internal/xpkg/upbound/context.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// Package upbound contains
+// Package upbound implements authentication for xpkg.upbound.io.
 package upbound
```
jbw976, line 18: out of curiousity, what's your thought process for putting logic into `internal/...` instead of under `cmd/crank...`? we had a similar discussion in https://github.com/crossplane/crossplane/pull/4751#discussion_r1359692506

jbw976 (commented):
Thanks for taking this clean-up effort on @negz!! took a quick pass through now with a few comments, but no major objections to anything here 🙇

cmd/crank/main.go
negz: Not really. 😄 I think the beta pattern (where the command lives in a subdir) is preferable, and should be easy to follow when we add an alpha command.

internal/xpkg/upbound/context.go
negz: Mostly intuition at this point, but my thinking is I'd like to keep `cmd/` focused on "top-level" command logic, with anything more complex living under `internal/`. I considered moving most of the `crossplane beta render` implementation to `internal/xrender` and just leaving the `Run` method under `cmd/`. This is similar to what we do with `cmd/crossplane`, which largely imports and plumbs together libraries from `internal/`.

cmd/crank/beta/xpkg/xpkg.go
negz: I might actually delete this help. I thought perhaps @plumbis had added it recently, but it looks like it was copied from `up` where it was added in https://github.com/upbound/up/pull/342. I do think we need something like this, but maybe not so lengthy and detailed?

negz: I ended up making this much shorter/less detailed than it was.

cmd/crank/beta/xpkg/init.go
```diff
@@ -16,38 +32,35 @@ import (
...
-		return "", errors.Errorf("unknown language %s", c.Language)
+func (c *initCmd) Help() string {
+	b := strings.Builder{}
+	b.WriteString("Crossplane initializes a package by using git to clone a template from a repository.\n")
+	b.WriteString("The template can be either a git repository URL, or a well-known template name.\n\n")
+	b.WriteString("Crossplane supports the following well-known-template names:\n\n")
+	for name, url := range WellKnownTemplates() {
+		b.WriteString(fmt.Sprintf(" - %s (%s)\n", name, url))
 	}
+	return b.String()
```
negz, line 60: FYI this renders like this:

```
remote:
$ go run cmd/crank/main.go beta xpkg init --help
Usage: crossplane beta xpkg init <name> <template>

Initialize a package from a template.

Crossplane initializes a package by using git to clone a template from a
repository. The template can be either a git repository URL, or a well-known
template name.

Crossplane supports the following well-known-template names:

  - provider-minimal (https://github.com/crossplane/provider-template)
  - provider-upjet (https://github.com/upbound/upjet-provider-template)
  - function-go (https://github.com/crossplane/function-template-go)

Arguments:
  <name>        Name of the package to initialize.
  <template>    Template to initialize the package from.

Flags:
  -h, --help             Show context-sensitive help.
  -v, --version          Print version and quit.
      --verbose          Print verbose logging statements.

  -d, --directory="."    Path of the directory to initialize.
```

phisco (commented):
A few feedbacks:
- I preferred keeping the beta/alpha commands where they would be without the beta/alpha prefix, as that would mean promoting them would be a 1 line change instead of moving all the related files.
- I like the direction of `--embed-runtime-image`, I think it conveys better how the image is being used by the build subcommands, however embed suggests it's a Boolean, I'd rather have it as `--embedded-runtime-image` probably.
- I guess you switched the tree view off because it wasn't playing nicely with having a single top subcommands, is this the case? I prefer the tree view, but I also found the default visualization not to work so well in this case

cmd/crank/xpkg/login.go
```diff
@@ -50,9 +50,9 @@ type loginCmd struct {
 	stdin  *os.File
 	client *http.Client
 
-	Username string `short:"u" env:"MARKETPLACE_USER" xor:"identifier" help:"Username used to execute command."`
-	Password string `short:"p" env:"MARKETPLACE_PASSWORD" help:"Password for specified user. '-' to read from stdin."`
-	Token    string `short:"t" env:"MARKETPLACE_TOKEN" xor:"identifier" help:"Token used to execute command. '-' to read from stdin."`
+	Username string `short:"u" env:"UP_USER" xor:"identifier" help:"Username used to execute command."`
+	Password string `short:"p" env:"UP_PASSWORD" help:"Password for specified user. '-' to read from stdin."`
+	Token    string `short:"t" env:"UP_TOKEN" xor:"identifier" help:"Token used to execute command. '-' to read from stdin."`
```
plumbis, line 55:
```suggestion
	Username string `short:"u" xor:"identifier" help:"Username for authentication."`
	Password string `short:"p" help:"Password for authentication. Use '-' to prompt for user input and hide the password."`
	Token    string `short:"t" xor:"identifier" help:"API token string for authentication. Use '-' to prompt for user input and hide the token."`
```

cmd/crank/xpkg/push.go
```diff
@@ -28,20 +28,23 @@ import (
...
+// DefaultRegistry for pushing Crossplane packages.
+const DefaultRegistry = "xpkg.upbound.io"
+
 // pushCmd pushes a package.
 type pushCmd struct {
 	fs afero.Fs
 
-	Tag     string `arg:"" help:"Tag of the package to be pushed. Must be a valid OCI image tag."`
+	Tag     string `arg:"" help:"Tag of the package to be pushed. Must be a valid OCI image tag. Unqualified tags will be pushed to xpkg.upbound.io."`
 	Package string `short:"f" help:"Path to package. If not specified and only one package exists in current directory it will be used."`
```
plumbis, line 48:
```suggestion
	Tag     string `arg:"" help:"Package filename and version tag to use in the container registry."`
	Package string `short:"f" help:"Path to the Crossplane xpkg file." default:"."`
```

What do you mean by "Unqualified tags"?

cmd/crank/xpkg/update.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/crossplane/crossplane/internal/xpkg"
+
+	_ "k8s.io/client-go/plugin/pkg/client/auth" // Load all the auth plugins for the cloud providers.
+)
+
+// UpdateCmd is exported so that it can be re-used by the beta xpkg subcommand.
+
+// UpdateCmd updates a package.
+type UpdateCmd struct {
+	Kind string `arg:"" help:"Kind of package to install. One of \"provider\", \"configuration\", or \"function\"." enum:"provider,configuration,function"`
```
plumbis, line 46: Same as install. Why do we have to specify the `Kind` instead of deriving it?

cmd/crank/xpkg/update.go
```diff
@@ -0,0 +1,116 @@
...
+	Ref  string `arg:"" help:"The package's OCI image reference (e.g. tag)."`
+	Name string `arg:""  optional:"" help:"Name of the package to update. Will be derived from the ref if omitted."`
```
plumbis, line 48:
```suggestion
	Ref  string `arg:"" help:"Package filename and version tag in the container registry."`
	Name string `arg:""  optional:"" help:"Name of the package inside Crossplane. Derived from the package metadata name by default."`
```

cmd/crank/beta/xpkg/init.go
```diff
@@ -16,38 +32,35 @@ import (
...
+	}
 }
 
-type initFunctionCmd struct {
-	Language functionLanguage `arg:"" help:"Language of the Function to initialize." enum:"go"`
-	Name     string           `arg:"" help:"Name of the Function to initialize."`
+// initCmd initializes a new package repository from a template repository.
+type initCmd struct {
+	Name     string `arg:"" help:"Name of the package to initialize."`
+	Template string `arg:"" help:"Template to initialize the package from."`
```
plumbis, line 47:
```suggestion
	Name     string `arg:"" help:"Name of the new package being initialized."`
	Template string `arg:"" help:"Template to use for initializing the package."`
```

cmd/crank/beta/xpkg/init.go
```diff
@@ -16,38 +32,35 @@ import (
...
 
-	CustomRepo string `short:"r" help:"URL of the custom template repository to use instead of the default one for the language."`
-	Directory  string `short:"d" help:"Path of the directory to initialize." default:"." type:"path"`
+	Directory string `short:"d" help:"Path of the directory to initialize." default:"." type:"path"`
```
plumbis, line 49:
```suggestion
	Directory string `short:"d" help:"The directory to create with initial package contents." default:"." type:"path"`
```

plumbis (changes requested):
The commands were modified in this PR but we should update cmd/crank/beta/render/cmd.go

```
type Cmd struct {
	CompositeResource string `arg:"" type:"existingfile" help:"A composite resource (XR) YAML file to render."`
	Composition       string `arg:"" type:"existingfile" help:"A Composition YAML file to use."`
	Functions         string `arg:"" help:"A stream or directory of YAML files containing the Composition Functions to use."`

	ContextValues map[string]string `placeholder:"KEY=JSON-VALUE;..." help:"Semicolon-separated JSON encoded context variables to pass to the Function pipeline. Takes precedence over --context-files."`
	ContextFiles  map[string]string `placeholder:"KEY=FILENAME;..." help:"Semicolon-separated JSON encoded context variables to pass to the Function pipeline."`

	IncludeResults    bool     `short:"r" default:"true" help:"Include the Function's Results output."`
	ObservedResources []string `short:"o" help:"A stream or directory of YAML files mocking the observed state of composed resources."`

	Timeout time.Duration `help:"Amount of time to wait for the function to complete." default:"1m"`
}
```

cmd/crank/xpkg/install.go
```diff
@@ -0,0 +1,198 @@
...
+	errPkgIdentifier = "invalid package image identifier"
+	errKubeConfig    = "failed to get kubeconfig"
+	errKubeClient    = "failed to create kube client"
+)
+
+// InstallCmd is exported so that it can be re-used by the beta xpkg subcommand.
+
+// InstallCmd installs a package.
+type InstallCmd struct {
+	Kind string `arg:"" help:"Kind of package to install. One of \"provider\", \"configuration\", or \"function\"." enum:"provider,configuration,function"`
```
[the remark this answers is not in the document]
negz: There is no YAML. This command essentially builds one (or rather, the internal struct representation) from its arguments.

cmd/crank/xpkg/update.go
negz: Answered in the other thread.

negz:
@phisco 

> I preferred keeping the beta/alpha commands where they would be without the beta/alpha prefix, as that would mean promoting them would be a 1 line change instead of moving all the related files.

I somewhat strongly prefer keeping the beta and alpha commands in a subdirectory:

* I feel it makes it more obvious whether code is beta or alpha when browsing the implementation. As opposed to where its called.
* It means we can easily have multiple implementations of the same subcommand (GA, alpha, beta) without potentially awkward naming issues (e.g. `xpkg.Cmd`, `xpkgBeta.Cmd`, etc).
* I'm not particularly worried about promoting commands requiring us to `git mv` some files.

> I like the direction of --embed-runtime-image, I think it conveys better how the image is being used by the build subcommands, however embed suggests it's a Boolean, I'd rather have it as --embedded-runtime-image probably.

To me `--embed-runtime-image` reads more like an action: "embed this runtime image". I think the fact it takes a string argument should hopefully make this clear. I've added placeholder examples to clarify further.

> I guess you switched the tree view off because it wasn't playing nicely with having a single top subcommands, is this the case? I prefer the tree view, but I also found the default visualization not to work so well in this case

I'll be honest, I really could not stand the tree view. It was especially bad when for a while this PR had added a few more commands. I don't think it will work very well when we have ~double the amount of commands we have now. My hope is that the compact view will be a good compromise. I think it gives you all the same information as the tree view, but:

* More compact
* Without arguments and flags showing up at top level (which I think we didn't want).

I also personally find it easier to see at a glance what the commands are when they're written as you'd copy and paste them to run - e.g. `crossplane xpkg build` rather than `crossplane / xpkg / build` all nested.

cmd/crank/xpkg/install.go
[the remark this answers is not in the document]
plumbis: I am probably overlooking a technical detail but what I'm getting as is what prevents us from reading the yaml stream and loading the `Kind`s inside instead of requiring the user to specify the package type?

negz:
@plumbis I'd prefer to go in depth on documentation in a separate PR. If we get this merged today I can open a follow on to get started on some of the changes you suggested.

cmd/crank/xpkg/install.go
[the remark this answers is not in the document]
negz: Oh gotcha. Yes that would technically be possible. It's an interesting idea, but I lean toward leaving this as is. The CLI would need to pull down the package and introspect it. It doesn't do that (anywhere) today. We'd presumably have to tackle proxies, caching, and all that good stuff to make it work.

Despite largely rewriting the implementation of this command it's a pretty small diff on the previous implementation functionally. We go from `crossplane install provider` (where provider is a subcommand) to `crossplane xpkg install provider` (where provider is an argument).

internal/xpkg/v2/build.go
```diff
@@ -1,261 +0,0 @@
-/*
```
jbw976, line 1: very cool to already clean up the `xpkg/v2` temp package also - you're getting a ton of things clean in this PR @negz!! 🧹 🗑️

jbw976 (commented):
I'm pretty good with this major refactor and major clean-up which puts us into good position for feature freeze to continue with polish and further CLI improvements.

Biggest question is on the removal of the generated clientset stuff, please help me understand that one! 🙇‍♂️

apis/generate.go
```diff
@@ -58,12 +58,6 @@ limitations under the License.
 //go:generate go run -tags generate github.com/jmattheis/goverter/cmd/goverter -output ./apiextensions/v1/zz_generated.conversion.go -packageName v1 -packagePath=github.com/crossplane/crossplane/apis/apiextensions/v1 ./apiextensions/v1
 //go:generate go run -tags generate github.com/jmattheis/goverter/cmd/goverter -output ./pkg/meta/v1alpha1/zz_generated.conversion.go -packageName v1alpha1 -packagePath=github.com/crossplane/crossplane/apis/pkg/meta/v1alpha1 ./pkg/meta/v1alpha1
 
-// Generate clientset for types.
```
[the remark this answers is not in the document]
negz: As far as I can tell they were only being used by the package install and update CLI commands, which now use the controller-runtime client instead. I don't think we need to announce anything. They were in the `internal` package, so no one would have been able to use them before.

cmd/crank/beta/xpkg/init.go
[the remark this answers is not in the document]
negz: Probably yes. I tried a few times but it was tough to come up with a succinct way to do this. Mind if we add this to the list for a follow-up PR more focused on CLI help output?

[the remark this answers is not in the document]
jbw976: yep! not a blocker for me :)

jbw976 (approved):

PR #4845, opened 2023-10-19
Use updated image for local dev

turkenh:

Follow up on https://github.com/crossplane/crossplane/pull/4831, use the updated image for local dev.

I have: 

### DIFF
--- a/cluster/local/kind.sh
+++ b/cluster/local/kind.sh
@@ -43,13 +43,14 @@ function check_context() {
 
 # configure kind
 KIND_NAME=${KIND_NAME:-"kind"}
+IMAGE_REPOSITORY="xpkg.upbound.io/${PROJECT_NAME}/${PROJECT_NAME}"
 case "${1:-}" in
   up)
     ${KIND} create cluster --name "${KIND_NAME}" --image "${KUBE_IMAGE}" --wait 5m
     ;;
   update)
     helm_tag="$(cat _output/version)"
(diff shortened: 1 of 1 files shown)


cluster/local/kind.sh
```diff
@@ -49,7 +49,7 @@ case "${1:-}" in
     ;;
   update)
     helm_tag="$(cat _output/version)"
-    copy_image_to_cluster ${BUILD_IMAGE} "${PROJECT_NAME}/${PROJECT_NAME}:${helm_tag}" "${KIND_NAME}"
+    copy_image_to_cluster ${BUILD_IMAGE} "xpkg.upbound.io/${PROJECT_NAME}/${PROJECT_NAME}:${helm_tag}" "${KIND_NAME}"
```
pedjak, line 52: instead of doing the same change 3 times, would it be simpler/less effort if we continue copying the image as is, using `${PROJECT_NAME}/${PROJECT_NAME}:${helm_tag}"` as the image ref, and then when interacting with to pass the arg:

```
--set image.repository=${PROJECT_NAME}
```

Orthogonal to that, we could extract `"${PROJECT_NAME}/${PROJECT_NAME}:${helm_tag}"` into a variable and use it everywhere, so that future changes would be one-liner?

Same goes for default helm cli args.

turkenh: Refactored to use a variable with updated value.

pedjak (approved):

PR #4849, opened 2023-10-19
feat(crank): add beta trace command

phisco:

Partially address #4672 
Rework based on the work done by @jbasement on https://github.com/crossplane/crossplane/pull/4751, unfortunately the original branch fell so much behind it was really hard to preserve a meaningful history, so as agreed I've squashed everything.

image
image
image

I have: 

- [ ] Linked a PR or a [docs tracking issue] to [document this change].


phisco:
Suggestions about the name of the command are welcome, describe reminds too much of the kubectl subcommand, but the resemblance stops there, so it could be misleading. The format is clearly

cmd/crank/beta/beta.go
```diff
@@ -20,12 +20,14 @@ limitations under the License.
...
 	"github.com/crossplane/crossplane/cmd/crank/beta/xpkg"
 )
 
 // Cmd contains beta commands.
 type Cmd struct {
-	XPKG   xpkg.Cmd   `cmd:"" help:"Manage packages."`
-	Render render.Cmd `cmd:"" help:"Render a claim or XR locally."`
+	XPKG     xpkg.Cmd     `cmd:"" help:"Manage packages."`
+	Render   render.Cmd   `cmd:"" help:"Render a claim or XR locally."`
+	Describe describe.Cmd `cmd:"" help:"Describe a Crossplane resource."`
```
plumbis, line 32:
```suggestion
	Describe describe.Cmd `cmd:"" help:"Describe a Crossplane resource."`
	Render   render.Cmd   `cmd:"" help:"Render a claim or XR locally."`
	XPKG     xpkg.Cmd     `cmd:"" help:"Manage packages."`
```
Sort commands alphabetically.

plumbis:
+1 to `trace`
-1 to `lineage`

cmd/crank/beta/describe/describe.go
```diff
@@ -0,0 +1,127 @@
...
+	errFmtInvalidResourceName = "invalid combined kind and name format, should be in the form of 'resource.group.example.org/name', got: %q"
+)
+
+// Cmd describes a Crossplane resource.
+type Cmd struct {
+	Resource string `arg:"" required:"" help:"'TYPE[.VERSION][.GROUP][/NAME]' identifying the Crossplane resource."`
+	Name     string `arg:"" optional:"" help:"Name of the Crossplane resource. Ignored if already passed as part of the RESOURCE argument."`
+
+	// TODO(phisco): add support for all the usual kubectl flags; configFlags := genericclioptions.NewConfigFlags(true).AddFlags(...)
+	Namespace string `short:"n" name:"namespace" help:"Namespace of resource to describe." default:"default"`
```
plumbis, line 53:
```suggestion
	Namespace string `short:"n" name:"namespace" help:"Namespace of the resource to describe." default:"default"`
```

bobh66:
> Suggestions about the name of the command are welcome. `describe` reminds too much of the `kubectl` subcommand, but the resemblance stops there, so it could be misleading. The original crossplane cli had `trace`, so that's an option. Lineage maybe? Given that it's similar to what [kube-lineage](https://github.com/tohjustin/kube-lineage) does? Any thought?

If the output is the same or close to `kubectl lineage` then I think `lineage` makes sense.  `trace` implies a path through a set of things (lines of code, resources, etc) so while that might be good for a list of `ownerReferences`, for example, I don't think it works for a `lineage`-like tree.

cmd/crank/beta/beta.go
phisco: Yes, I'll have to rebase it on top of the "sort everything" PR

phisco:
noticed the tree view requires a bit more refinement

phisco:
From the feedbacks here and I've collected irl, sounds like we have some consensus around `trace`, so if no one is strongly against it we'll go on with that. CC @bobh66

I added back the `dot` output too as discussed. I kept it simple and decided a set of meaningful labels for each node, here an example:
image
Down the line we could think of adding a template to let people customise it.

Right now this is only able to handle XRC/XR/MR, we can think of adding Packages and XRDs support down the line.

Piotr1215:
Could we extend help for the command to also include the usage of the `- o dot` output? It has a few parameters that users might not know how to handle. For example the fact that verbose outputs the graph source code as well:
`go run github.com/crossplane/crossplane/cmd/crank beta describe nopresources.nop.example.org apiextensions-composition-minimal -o dot --verbose | dot -Tpng -o output.png`

cmd/crank/beta/trace/trace.go
```diff
@@ -0,0 +1,143 @@
...
+	rootRef := &v1.ObjectReference{
+		Kind:       mapping.GroupVersionKind.Kind,
+		APIVersion: mapping.GroupVersionKind.GroupVersion().String(),
+		Name:       name,
+	}
+	if mapping.Scope.Name() == meta.RESTScopeNameNamespace && c.Namespace != "" {
+		rootRef.Namespace = c.Namespace
+	}
+	logger.Debug("Getting resource tree", "rootRef", rootRef.String())
+	root, err := resClient.GetResourceTree(context.Background(), rootRef)
```
jbw976, line 118: I'm think this `GetResourceTree()` call is failing for my scenario in a way that I would not expect it to:

```
❯ ./crank beta trace acmedatabases.v1.acme.com/acme-db-prod
crossplane: error: cannot get requested resource: couldn't get child resource: couldn't get resource: secrets "acme-db-prod-conn" not found
```

Basically my scenario here is creating a claim but forgetting to provide credentials to the cloud provider. So I'm expecting this claim to not become ready and I want to find out why with this `trace` command.

The connection secret for my claim hasn't been created yet, but it looks like this function doesn't handle that gracefully in its effort to discover the entire graph. It doesn't find the connection secret (as expected, it doesn't exist) and then returns an error for that instead of continuing to trace through the rest of the graph.

It would probably be a better experience to handle the missing connection secret and still print out for me the graph of resources so I can visualize them and see where the root failure is.

Some more debugging information:

General claim output: 

```
❯ k get acmedatabases.v1.acme.com/acme-db-prod
NAME           SYNCED   READY   CONNECTION-SECRET   AGE
acme-db-prod   True     False   acme-db-prod-conn   8m51s
```

`--verbose` output:

```
❯ ./crank beta trace acmedatabases.v1.acme.com/acme-db-prod --verbose
2023-10-21T07:51:38-07:00	DEBUG	Built printer	{"Resource": "acmedatabases.v1.acme.com/acme-db-prod", "output": "default"}
2023-10-21T07:51:38-07:00	DEBUG	Found kubeconfig	{"Resource": "acmedatabases.v1.acme.com/acme-db-prod"}
2023-10-21T07:51:38-07:00	DEBUG	Adding latest event option	{"Resource": "acmedatabases.v1.acme.com/acme-db-prod"}
2023-10-21T07:51:38-07:00	DEBUG	Built client	{"Resource": "acmedatabases.v1.acme.com/acme-db-prod"}
2023-10-21T07:51:38-07:00	DEBUG	Getting resource tree	{"Resource": "acmedatabases.v1.acme.com/acme-db-prod", "rootRef": "&ObjectReference{Kind:AcmeDatabase,Namespace:default,Name:acme-db-prod,UID:,APIVersion:acme.com/v1,ResourceVersion:,FieldPath:,}"}
2023-10-21T07:51:38-07:00	DEBUG	cannot get requested resource	{"Resource": "acmedatabases.v1.acme.com/acme-db-prod", "error": "couldn't get child resource: couldn't get resource: secrets \"acme-db-prod-conn\" not found"}
crossplane: error: cannot get requested resource: couldn't get child resource: couldn't get resource: secrets "acme-db-prod-conn" not found
```

The relationship of claim -> XR -> MR that I was hoping to see output, even though the connection secret doesn't exist yet in the claim:

```
❯ k get acmedatabase acme-db-prod -o json | jq '.spec.resourceRef'
{
  "apiVersion": "acme.com/v1",
  "kind": "XAcmeDatabase",
  "name": "acme-db-prod-7sc7m"
}

❯ k get xacmedatabase acme-db-prod-7sc7m -o json | jq '.spec.resourceRefs'
[
  {
    "apiVersion": "sql.gcp.upbound.io/v1beta1",
    "kind": "DatabaseInstance",
    "name": "acme-db-prod-7sc7m-2ggn9"
  }
]

❯ k get databaseinstance acme-db-prod-7sc7m-2ggn9 -o json | jq '.status'
{
  "atProvider": {},
  "conditions": [
    {
      "lastTransitionTime": "2023-10-21T14:38:18Z",
      "message": "connect failed: cannot get terraform setup: cannot extract JSON key credentials: cannot get credentials secret: Secret \"gcp-secret\" not found",
      "reason": "ReconcileError",
      "status": "False",
      "type": "Synced"
    }
  ]
}
```

jbw976 (commented):
Really great progress @phisco!! I took it for a test spin this morning and found a couple issues that aren't working for my test scenario I wanted to debug with this new command. I also had a bit of trouble with the resource name format.

What do you think about my experience and any potential improvements we can make to smooth it out for other folks new to the command? 😊

cmd/crank/beta/trace/trace.go
```diff
@@ -0,0 +1,143 @@
...
+	errKubeConfig             = "failed to get kubeconfig"
+	errCouldntInitKubeClient  = "cannot init kubeclient"
+	errCannotGetKindAndName   = "cannot get kind and name"
+	errCannotGetMapping       = "cannot get mapping for resource"
+	errCannotInitPrinter      = "cannot init new printer"
+	errFmtInvalidResourceName = "invalid combined kind and name format, should be in the form of 'resource.group.example.org/name', got: %q"
+)
+
+// Cmd builds the trace tree for a Crossplane resource.
+type Cmd struct {
```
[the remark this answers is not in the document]
phisco: Definitely! I will rebase on Monday and adhere to the new standard with sorted commands and helpful help messages!

cmd/crank/beta/trace/trace.go
```diff
@@ -0,0 +1,143 @@
...
+	Resource string `arg:"" required:"" help:"'TYPE[.VERSION][.GROUP][/NAME]' identifying the Crossplane resource."`
```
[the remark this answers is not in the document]
phisco: It should support the same format supported by `kubectl get`, but unfortunately using the upstream parsing code wasn't so straightforward. I'll double check it's working as expected and will add examples 👍

cmd/crank/beta/trace/trace.go
phisco: Good point! Properly handling expected resources not existing is definitely something we want to show and highlight as a possible source of issue. Any suggestion about how to graphically show that in the tree view? An additional boolean FOUND column? Or a generic ERROR column where we could show errors for each resource? I think the latter could be more flexible.

jbw976: Hmm actually I don't (yet) think it's important to display this particular scenario with any specific knowledge/logic.  The missing connection secret for the claim is a normal part of its lifecycle - it hasn't gotten far enough for the connection secret to be created. So I wouldn't elevate that symptom to a top level experience personally.

What I would do though is to not let that temporary condition stop the processing of the rest of the graph.  In this scenario, where we really want to get to is the event or condition at the MR - that's where the real root cause is here.

So just by displaying the entire tree/graph of claim --> XR --> MR and showing the last event or status or something on the MR would be enough for this scenario.  No need for extra columns or experiences here is my bet! 🙇

cmd/crank/beta/trace/trace.go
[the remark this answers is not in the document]
phisco: Tried giving it a quick shot, I'll iterate further on Monday.

cmd/crank/beta/trace/trace.go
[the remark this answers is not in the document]
phisco: Although supported by kubectl, the `kind.version.group/name` format wasn't really adding anything here. So I just reverted to the two arguments being required and added a few examples to explain the available syntaxes. LMK how it looks now.

cmd/crank/beta/trace/trace.go
jbw976: took the latest for another test spin, I was able to get the full output of the tree graph now! the root cause is printed within the output, so this is definitely helpful 💪 

```
❯ ./crank beta trace acmedatabases acme-db-prod
NAMESPACE           APIVERSION                   NAME                                                                 READY     SYNCED   STATUS
default             acme.com/v1                  AcmeDatabase/acme-db-prod                                            False     True     Waiting: "Composite resource claim is waiting for composite resource to become Ready"
                    acme.com/v1                  ├── XAcmeDatabase/acme-db-prod-7sc7m                                 False     True     Creating: ""
                    sql.gcp.upbound.io/v1beta1   │   ├── DatabaseInstance/acme-db-prod-7sc7m-2ggn9                    Unknown   False    ReconcileError: "connect failed: cannot get terraform setup: cannot extract JSON key credentials: cannot get credentials secret: Secret \"gcp-secret\" not found"
crossplane-system   v1                           │   │   └── Secret/5d2b9592-d3c7-4dc5-b27c-67a3c10c1b45-postgresql                      Error: secrets "5d2b9592-d3c7-4dc5-b27c-67a3c10c1b45-postgresql" not found
crossplane-system   v1                           │   └── Secret/5d2b9592-d3c7-4dc5-b27c-67a3c10c1b45                                     : ""
default             v1                           └── Secret/acme-db-prod-conn                                                            Error: secrets "acme-db-prod-conn" not found
```

Further thoughts on this output:

* **wrapping by default**: with the default font/zoom size of my terminal on my laptop screen, the output is wide enough to wrap some lines, which causes breaks in the tree lines and makes the valuable visualization of the tree much more difficult to parse. Thoughts on limiting the width of each line somehow?
* **namespace and apiversion columns** i'm wondering how valuable these columns are in the default view, esp. when the namespace column is often blank because of cluster scoped XRs and MRs. They are taking up almost 30% of the horizontal real estate - what do you think about not including them in the default view? I keep coming back to thinking the `NAME` column should be first so it sort of anchors the graph on the left hand side.
* **connection secrets** I was surprised to see connection secrets included in the graph and they don't seem to be helpful for finding root cause here either. I'm not sure how much valuable details they would be able to show in this graph since they don't have a READY/SYNCED value and they probably don't have a useful STATUS value either. I was imagining including only resources that are XRM compliant such as claims, XRs, and MRs.  Perhaps we can have a flag to `--show-secrets` for folks to optionally include them in the output, but have them excluded by default?

phisco: Totally agree with all your points, I've:
- cropped the condition message to 64 characters, the full output can be shown by using `-o wide` (we could think of additional columns later, as kubectl does)
- added a `--show-connection-secrets` flag, false by default
- dropped namespace and apiversion columns, namespace if present is added at the end of the name between brackets. The apiVersion could be useful in case of multiple colliding types, could potentially be a column we add to the wide output.

phisco: otherwise we could just drop the message part from the status similarly to what kubectl get does and leave getting that to further investigation 🤔

jbw976: Did you push those latest updates to your branch yet @phisco? i'd love to see that code and give those changes a test spin too! 🙇‍♂️ 

I'd be hesitant to remove the message part of the status because in my main testing scenario, and other unhealthy scenarios I can imagine, the message is going to have the most useful information about the root cause. Let's keep it until we arrive at a conclusion that it's causing more harm than good - my expectation is that it will cause lots of good 😊 

How'd you arrive at 64 chars by the way? Do you think we'd have room for more than that since the namespace/apiversion colums are now gone?

phisco: Just pushed, I forgot to do that earlier, let me know!

64 is just a random number that worked well in my scenario, we could probably compute that based on the terminal width, but I'm not sure it's worth it.

With wrapped errors the most meaningful part is at the end of the message so I just left the full condition status, e.g. Waiting, and then cut the message to just the last 64 characters, printing it in full with wide output.

E.g.

```
NAME                                                       READY     SYNCED   STATUS
ANopResource/apiextensions-composition-minimal (default)   False     True     Waiting: ...resource claim is waiting for composite resource to become Ready
└─ XANopResource/apiextensions-composition-minimal-2w2lr   Unknown   False    ReconcileError: cannot select Composition: no compatible Compositions found
```

jbw976: Awesome dude, just took it for a real quick test spin and it looks really good dude, I think we're landing on what we'd like to ship for v1.14!!

I'll take a deeper code review pass this afternoon too so hopefully we'll be able to merge first thing Monday 😅

phisco: Awesome, thanks! I could definitely add some more test coverage and the default printer's code could benefit a cleanup/comment pass for sure. I'll do that tomorrow morning, while integrating your feedback 🙏

cmd/crank/beta/trace/internal/resource/client_test.go
```diff
@@ -0,0 +1,292 @@
...
+	type want struct {
+		refs []v1.ObjectReference
+	}
+	cases := map[string]struct {
+		reason string
+		args   args
+		want   want
+	}{
+		"XRCWithChildrenMR": {
+			reason: "Should return the XR child for an XRC.",
```
jbw976, line 96: the test case name mentions `MR` but the test case content looks to mention XRs instead - is this test case testing what you expect it to?

cmd/crank/beta/trace/trace.go
```diff
@@ -0,0 +1,163 @@
...
+// Help returns help message for the trace command.
+func (c *Cmd) Help() string {
+	return `
+This command trace a Crossplane resource to get a detailed output of its
+relationships, helpful for troubleshooting.
+
+If needed the resource kind can be also specified further,
+'TYPE[.VERSION][.GROUP]', e.g. mykind.example.org.
+
+Examples:
```
jbw976, line 71: love all these examples, that will be really helpful for folks to use this command effectively!! 😎

cmd/crank/beta/trace/trace.go
```diff
@@ -0,0 +1,163 @@
...
+  crossplane beta trace mykind my-res -n my-ns -o json | jq
+
+  # Output debug logs to stderr while redirecting a dot formatted graph to dot
+  crossplane beta trace mykind my-res -n my-ns -o dot --verbose | dot -Tpng -o output.png
+`
+}
+
+// Run runs the trace command.
+func (c *Cmd) Run(k *kong.Context, logger logging.Logger) error { //nolint:gocyclo // TODO(phisco): refactor
+	logger = logger.WithValues("Resource", c.Resource)
```
jbw976, line 94: should the `Name` also be included in the `logger.WithValues`?

jbw976 (approved):
This is looking pretty good to me @phisco, I really like where this is landing to ship in v1.14 and then we can continue to iterate with more functionality.

I've added a few comments and questions, but I don't think there is anything blocking on my side, so I'm adding an approval and you can finish this PR off as you want 🤩

turkenh:
Just tested with the latest state with **platform-ref-aws:v0.6.0** and **XP v1.13.2**. 

In the beginning, the command failed with: 

```bash
./_output/bin/darwin_arm64/crank beta trace cluster platform-ref-aws-ht
crossplane: error: cannot get requested resource: couldn't get child resource: couldn't get resource: resource name may not be empty
```

While I was trying to debug it, it resolved itself, and started printing the tree. So, I don't have much clue on what was wrong before, but enhancing the error message above with some more details could help. 

Now, I am seeing a proper output as below:

```bash
❯ ./_output/bin/darwin_arm64/crank beta trace cluster platform-ref-aws-ht
NAME                                                                 READY     SYNCED    STATUS
Cluster/platform-ref-aws-ht (default)                                True      True      Available
└─ XCluster/platform-ref-aws-ht-ctk5x                                True      True      Available
   ├─ XNetwork/platform-ref-aws-ht-ctk5x-6lpkx                       True      True      Available
   │  ├─ VPC/platform-ref-aws-ht-ctk5x-fcx5v                         True      True      Available
   │  ├─ InternetGateway/platform-ref-aws-ht-ctk5x-5lt6p             True      True      Available
   │  ├─ Subnet/platform-ref-aws-ht-ctk5x-b98lz                      True      True      Available
   │  ├─ Subnet/platform-ref-aws-ht-ctk5x-6n5j8                      True      True      Available
   │  ├─ Subnet/platform-ref-aws-ht-ctk5x-fp6r9                      True      True      Available
   │  ├─ Subnet/platform-ref-aws-ht-ctk5x-fhwqv                      True      True      Available
   │  ├─ RouteTable/platform-ref-aws-ht-ctk5x-f9x5v                  True      True      Available
   │  ├─ Route/platform-ref-aws-ht-ctk5x-qr9db                       True      True      Available
   │  ├─ MainRouteTableAssociation/platform-ref-aws-ht-ctk5x-64mg4   True      True      Available
   │  ├─ RouteTableAssociation/platform-ref-aws-ht-ctk5x-cx9ns       True      True      Available
   │  ├─ RouteTableAssociation/platform-ref-aws-ht-ctk5x-vnvfl       True      True      Available
   │  ├─ RouteTableAssociation/platform-ref-aws-ht-ctk5x-sr6g7       True      True      Available
   │  ├─ RouteTableAssociation/platform-ref-aws-ht-ctk5x-b6zmz       True      True      Available
   │  ├─ SecurityGroup/platform-ref-aws-ht-ctk5x-q5nx7               True      True      Available
   │  ├─ SecurityGroupRule/platform-ref-aws-ht-ctk5x-lgpnx           True      True      Available
   │  └─ SecurityGroupRule/platform-ref-aws-ht-ctk5x-xsqlb           True      True      Available
   ├─ XEKS/platform-ref-aws-ht-ctk5x-svcm2                           True      True      Available
   │  ├─ Role/platform-ref-aws-ht-ctk5x-5ffp2                        True      True      Available
   │  ├─ RolePolicyAttachment/platform-ref-aws-ht-ctk5x-rwdmg        True      True      Available
   │  ├─ Cluster/platform-ref-aws-ht-ctk5x-59f5j                     True      True      Available
   │  ├─ ClusterAuth/platform-ref-aws-ht-ctk5x-pwkr9                 True      True      Available
   │  ├─ Role/platform-ref-aws-ht-ctk5x-hxtz2                        True      True      Available
   │  ├─ RolePolicyAttachment/platform-ref-aws-ht-ctk5x-phvxp        True      True      Available
   │  ├─ RolePolicyAttachment/platform-ref-aws-ht-ctk5x-pr272        True      True      Available
   │  ├─ RolePolicyAttachment/platform-ref-aws-ht-ctk5x-hglzh        True      True      Available
   │  ├─ NodeGroup/platform-ref-aws-ht-ctk5x-mkfrv                   True      True      Available
   │  ├─ OpenIDConnectProvider/platform-ref-aws-ht-ctk5x-mrt9x       True      True      Available
   │  └─ ProviderConfig/platform-ref-aws-ht                          Unknown   Unknown
   └─ XServices/platform-ref-aws-ht-ctk5x-lwgg2                      True      True      Available
      ├─ Release/platform-ref-aws-ht-ctk5x-5hh8r                     True      True      Available
      └─ Release/platform-ref-aws-ht-ctk5x-bk7d6                     True      True      Available
```

The only question is whether we could handle `ProviderConfig`'s better, but I don't have a great solution given we have readiness implemented in Composition and the cli doesn't have access there.

cmd/crank/beta/trace/internal/resource/client.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+
+	return refs
+}
+
+// MappingFor returns the RESTMapping for the given resource or kind argument.
+// Copied over from cli-runtime pkg/resource Builder,
+// https://github.com/kubernetes/cli-runtime/blob/9a91d944dd43186c52e0162e12b151b0e460354a/pkg/resource/builder.go#L768
+func (kc *Client) MappingFor(resourceOrKindArg string) (*meta.RESTMapping, error) {
+	// TODO(phisco): actually use the Builder.
```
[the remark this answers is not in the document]
phisco: it's more a wish, there is no real need. It would be awesome to be able to leverage the upstream code, but it shouldn't change the end result, so I wouldn't actively track it.

cmd/crank/beta/trace/internal/printer/dot.go
```diff
@@ -0,0 +1,101 @@
...
+		queue = queue[1:]
+
+		node := g.Node(fmt.Sprintf("%d", id))
+		id++
+
+		if item.parent != nil {
+			g.Edge(*item.parent, node)
+		}
+
+		label := &dotLabel{
```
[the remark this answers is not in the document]
phisco: yes, I think the dot view is more useful to give an overall view on the relationships between resources, not really for debugging, so I'd avoid showing the status message to keep it nicer.

phisco:
@turkenh `crossplane: error: cannot get requested resource: couldn't get child resource: couldn't get resource: resource name may not be empty` could mean the resource reference in a child resource didn't have a name set when we tried getting it, we know that's something that could happen, so I'll make sure to properly handle it, good call!

cmd/crank/beta/trace/internal/resource/client.go
```diff
@@ -119,43 +121,44 @@ func (kc *Client) getResource(ctx context.Context, ref *v1.ObjectReference) *Res
...
+		xrc := claim.Unstructured{Unstructured: obj}
+		if ref := xrc.GetResourceReference(); ref != nil {
+			refs = append(refs, v1.ObjectReference{
+				APIVersion: ref.APIVersion,
+				Kind:       ref.Kind,
+				Name:       ref.Name,
+				Namespace:  ref.Namespace,
+			})
+		}
+	case obj.GetName():
```
turkenh, line 142: This would hold if this is the top-level composite, but not if it is one of the nested ones. 

```yaml
kind: XNetwork
metadata:
  annotations:
    crossplane.io/composition-resource-name: compositeNetworkEKS
  generateName: platform-ref-aws-ht-ctk5x-
  labels:
    crossplane.io/claim-name: platform-ref-aws-ht
    crossplane.io/composite: platform-ref-aws-ht-ctk5x
  name: platform-ref-aws-ht-ctk5x-6lpkx
```

phisco: good point, rewritten as discussed, it's now working fine with nested Compositions:
```
go run github.com/crossplane/crossplane/cmd/crank beta trace nopresource.nop.example.org apiextensions-composition-functions
NAME                                                                   READY   SYNCED   STATUS
NopResource/apiextensions-composition-functions (default)              True    True     Available
└─ XNopResource/apiextensions-composition-functions-829f8              True    True     Available
   └─ XNopResource/apiextensions-composition-functions-829f8-j8wvd     True    True     Available
      └─ NopResource/apiextensions-composition-functions-829f8-mtn56   True    True
```

phisco:
I've added support for EnvironmentConfigs for XRs too, I think it could help with the environment debugging issues we have been discussing, here a really symptomatic example from our E2Es:
```
go run github.com/crossplane/crossplane/cmd/crank beta trace sqlinstance.example.org example
NAME                                            READY   SYNCED   STATUS
SQLInstance/example (default)                   True    True     Available
└─ XSQLInstance/example-gpwvk                   True    True     Available
   ├─ NopResource/example-gpwvk-ck9d9           True    True
   ├─ EnvironmentConfig/example-environment-1
   ├─ EnvironmentConfig/example-environment-2
   └─ EnvironmentConfig/example-environment-3
```
WDYT @jbw976?

phisco:
@turkenh, awesome!

About the `Unknown`s, we can definitely replace them with `-`, however when you see `Unknown` it means the resource actually has a condition status set to `Unknown`, not that the status is an empty string. I think there is value in knowing which case it is.

So, should we map empty condition status, `Unknown`s or both to `-`?

turkenh:
I think both resources I am interested in here, ProviderConfigs and Usages, don't have those conditions at all. 

So, seeing `Unknown` if the condition status is set to `Unknown` but seeing `-` when the resource doesn't have that condition at all feels like a good solution.

phisco:
> I think both resources I am interested in here, ProviderConfigs and Usages, don't have those conditions at all.

Then it shouldn't show them as `Unknown` here:
```
   │  ├─ ProviderConfig/platform-ref-aws-ht                          Unknown   Unknown
```

phisco:
I thought the `Unknown` was actually coming from the resource status, but turns out the defaulting was coming from the `ConditionedStatus.GetCondition`, so I mapped it to `-` as it makes the view more compact. 🎉 I think I can squash everything and we can get it merged, WDYT @turkenh @jbw976?

jbw976:
The `Unknown` discussion seems to have ended up at a good place! With my simple test scenario, I thought the `-` for `Ready` on the `DatabaseInstance` was incorrect, but looking further at the resource, it really doesn't have a `Ready` condition set.  So it's working as expected and makes sense for all scenarios, including `Usage` and `ProviderConfig` that don't have a `Ready` status condition.  Ship it!! :shipit: 

`DatabaseInstance` shows `-` for `Ready`:
```
❯ crank beta trace acmedatabases acme-db-prod -s
NAME                                                                    READY   SYNCED   STATUS
AcmeDatabase/acme-db-prod (default)                                     False   True     Waiting: ...resource claim is waiting for composite resource to become Ready
└─ XAcmeDatabase/acme-db-prod-7sc7m                                     False   True     Creating
   ├─ DatabaseInstance/acme-db-prod-7sc7m-2ggn9                         -       False    ReconcileError: ...ls: cannot get credentials secret: Secret "gcp-secret" not found
   └─ Secret/5d2b9592-d3c7-4dc5-b27c-67a3c10c1b45 (crossplane-system)
```

But that's OK, it doesn't have a `Ready` status condition 😊 
```
❯ k get DatabaseInstance/acme-db-prod-7sc7m-2ggn9 -o json | jq '.status'
{
  "atProvider": {},
  "conditions": [
    {
      "lastTransitionTime": "2023-10-21T14:38:18Z",
      "message": "connect failed: cannot get terraform setup: cannot extract JSON key credentials: cannot get credentials secret: Secret \"gcp-secret\" not found",
      "reason": "ReconcileError",
      "status": "False",
      "type": "Synced"
    }
  ]
}
```

cmd/crank/beta/trace/internal/printer/default_test.go
```diff
@@ -52,15 +52,15 @@ func TestDefaultPrinter(t *testing.T) {
...
-   └─ User/test-resource-user-hash                     True    True                                                           
+NAME                                                   READY   SYNCED   STATUS                                              
+ObjectStorage/test-resource (default)                  True    True                                                         
+└─ XObjectStorage/test-resource-hash                   True    True                                                         
+   ├─ Bucket/test-resource-bucket-hash                 True    True                                                         
+   │  ├─ User/test-resource-child-1-bucket-hash        False   True     SomethingWrongHappened: Error with bucket child 1   
+   │  ├─ User/test-resource-child-mid-bucket-hash      True    False    CantSync: Sync error with bucket child mid          
+   │  └─ User/test-resource-child-2-bucket-hash        False   True     SomethingWrongHappened: Error with bucket child 2   
+   │     └─ User/test-resource-child-2-1-bucket-hash   -       True
```
turkenh, line 62: nit: Should we still have a test for the real `Unknown`s :)

cmd/crank/beta/trace/internal/printer/default.go
```diff
@@ -0,0 +1,192 @@
...
+	case r.Error != nil:
+		// if there is an error we want to show it
+		status = "Error"
+		m = r.Error.Error()
+	case readyCond.Status == corev1.ConditionTrue && syncedCond.Status == corev1.ConditionTrue:
+		// if both are true we want to show the ready reason only
+		status = string(readyCond.Reason)
+
+	// The following cases are for when one of the conditions is false,
+	// prioritizing readiness over synced in case of issues.
```
turkenh, line 159: Sync comes before Ready, i.e. first the resource is synchronized, then becomes ready. So, prioritizing synced might make more sense.

cmd/crank/beta/trace/internal/printer/default.go
```diff
@@ -0,0 +1,192 @@
...
+	// Append the message to the status if it's not empty
+	if m != "" {
+		status = fmt.Sprintf("%s: %s", status, m)
+	}
+
+	return mapUnknownStatusToDash(readyCond.Status), mapUnknownStatusToDash(syncedCond.Status), status
+}
+
+func mapUnknownStatusToDash(s corev1.ConditionStatus) string {
+	if s == corev1.ConditionUnknown {
```
turkenh, line 188: I was under the impression that there are really cases where a resources `Synced` or `Ready` condition was set to `Unknown` and there are also cases where those condition not existing at all. We will still show the former as `Unknown` but the latter as `-`. However, it looks like we will be mapping both cases to `Unknown` with this implementation.

cmd/crank/beta/trace/trace.go
```diff
@@ -0,0 +1,164 @@
...
+	// TODO(phisco): move to namespace defaulting to "" and use the current context's namespace
+	Namespace             string `short:"n" name:"namespace" help:"Namespace of the resource." default:"default"`
+	Output                string `short:"o" name:"output" help:"Output format. One of: default, wide, json, dot." enum:"default,wide,json,dot" default:"default"`
+	ShowConnectionSecrets bool   `short:"s" name:"show-connection-secrets" help:"Show connection secrets in the output."`
+}
+
+// Help returns help message for the trace command.
+func (c *Cmd) Help() string {
+	return `
+This command trace a Crossplane resource to get a detailed output of its
```
turkenh, line 65:
```suggestion
This command trace a Crossplane resource (Claim, Composite, or Managed Resource) to get a detailed output of 
```

It would be helpful to be explicit about what we mean by "Crossplane resource". Not sure tracing an MR would make sense at all but AFAIU command works with them as well.

cmd/crank/beta/trace/trace.go
```diff
@@ -0,0 +1,164 @@
...
+	// Init new printer
+	p, err := printer.New(c.Output)
+	if err != nil {
+		return errors.Wrap(err, errInitPrinter)
+	}
+	logger.Debug("Built printer", "output", c.Output)
+
+	kubeconfig, err := ctrl.GetConfig()
+	if err != nil {
+		logger.Debug(errKubeConfig, "error", err)
```
turkenh, line 105: nit: Do we need to log given we're already returning. And also to be consistent with other error returns.

turkenh (approved):
Looking good, just a few nits!

cmd/crank/beta/trace/internal/printer/default.go
phisco: the defaulting was happening at a lower level, on the ConditionedStatus.GetCondition, so I thought it would be fine and more compact to just always show "-", but you are right it make more sense not to throw away this information.
I've duplicated the GetCondition code and now return a really empty condition, so now we have:
- True/False/Unknown => True/False/Unknown
- "" => "-"

cmd/crank/beta/trace/trace.go
phisco: done: https://github.com/crossplane/crossplane/pull/4849/commits/f8998ceee913e53c1c4585e448f5d0936c44fb38

cmd/crank/beta/trace/trace.go
phisco: done: https://github.com/crossplane/crossplane/pull/4849/commits/f8998ceee913e53c1c4585e448f5d0936c44fb38

cmd/crank/beta/trace/internal/printer/default.go
phisco: https://github.com/crossplane/crossplane/pull/4849/commits/f8998ceee913e53c1c4585e448f5d0936c44fb38

cmd/crank/beta/trace/internal/printer/default.go
phisco: done: https://github.com/crossplane/crossplane/pull/4849/commits/f8998ceee913e53c1c4585e448f5d0936c44fb38

cmd/crank/beta/trace/internal/printer/default_test.go
phisco: done: https://github.com/crossplane/crossplane/pull/4849/commits/f8998ceee913e53c1c4585e448f5d0936c44fb38

PR #4861, opened 2023-10-20
feat: switch probes to only startupProbe

phisco:

The probes introduced in #4748 could be too aggressive and lead to some undesired restarts under heavier operations, e.g. in my case installing `platform-ref-aws`, if the resources are constrained.

The original issue we were trying to solve was that we were seeing some failures in the e2es at startup due to the webhooks not being yet ready to receive requests. So, a [startupProbe](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-startup-probes) could be enough, although it could lead to the pod being restarted after the `failureThreshold * periodSeconds` time window, I've set it to be 60s (30*2s), but we could be even more permissive and increase the failureThreshold if needed. To compensate, I've configured the probe to be just a `tcpProbe`, which should be less affected even if for some reason the pod was under heavier usage.

I hit this while testing it locally on the latest [commit](https://github.com/crossplane/crossplane/commit/48f9ed536161cad46231a08ed8c593fe9d86e459), by just applying the following manifest:
```yaml
kind: Configuration
metadata:
  name: platform-ref-aws
spec:
  package: xpkg.upbound.io/upbound/platform-ref-aws:v0.6.0
```
Seeing a few restarts of the Crossplane pod after that.

With this change I couldn't see any restart, and even killing the crossplane pod while it was trying to deploy `platform-ref-aws` didn't result in any issue at restart. Has to be seen whether it's actually solving the flakiness we were seeing.

I have: 

### DIFF
--- a/cmd/crossplane/core/core.go
+++ b/cmd/crossplane/core/core.go
@@ -349,7 +349,7 @@ func (c *startCommand) Run(s *runtime.Scheme, log logging.Logger) error { //noli
 	return errors.Wrap(mgr.Start(ctrl.SetupSignalHandler()), "cannot start controller manager")
 }
 
-// SetupProbes sets up the health and ready probes.
+// SetupProbes sets up the health and readiness probes.
 func (c *startCommand) SetupProbes(mgr ctrl.Manager) error {
 	// Add default readiness probe
 	if err := mgr.AddReadyzCheck("ping", healthz.Ping); err != nil {
--- a/cluster/charts/crossplane/templates/deployment.yaml
+++ b/cluster/charts/crossplane/templates/deployment.yaml
@@ -123,16 +123,13 @@ spec:
         name: {{ .Chart.Name }}
         resources:
           {{- toYaml .Values.resourcesCrossplane | nindent 12 }}
-        livenessProbe:
-          httpGet:
-            path: /healthz
-            port: healthz
-        readinessProbe:
-          httpGet:
-            path: /readyz
-            port: healthz
+        startupProbe:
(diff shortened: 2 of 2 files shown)


phisco:
NB: I've tried disabling schema aware validation but nothing changed

phisco:
3 runs all green except a single failure for the realtime-composition job, which is flapping also on master. So looks like is enough to address the issue we were originally trying to resolve.

cmd/crossplane/core/core.go
```diff
@@ -348,27 +348,20 @@ func (c *startCommand) Run(s *runtime.Scheme, log logging.Logger) error { //noli
...
 func (c *startCommand) SetupProbes(mgr ctrl.Manager) error {
 	// Add default readiness probe
 	if err := mgr.AddReadyzCheck("ping", healthz.Ping); err != nil {
 		return errors.Wrap(err, "cannot create ping ready check")
 	}
 
-	// Add default health probe
-	if err := mgr.AddHealthzCheck("ping", healthz.Ping); err != nil {
-		return errors.Wrap(err, "cannot create ping health check")
-	}
```
negz, line 361: Should we leave these, even if we don't use them for now? I could see them being handy in future. Or even just handy so that folks can configure their Helm chart to use them if they want.

negz (approved):

cmd/crossplane/core/core.go
phisco: yes, make sense to me, reverted these changes

PR #4883, opened 2023-10-23
fix(crank): properly load yaml files from directories

### LINKED ISSUE #4854: `crossplane beta render` -o only accept specific file not folder
### What happened?

the `crossplane beta render -o` command currently only accepts "file.yaml" as input. However, in the help command, it mentions that it can handle an optional stream or directory of YAML manifests, simulating the observed state of composed resources. This appears to be inconsistent and may be a bug.

```
  -o, --observed-resources=OBSERVED-RESOURCES,...
                           An optional stream or directory of YAML manifests
                           mocking the observed state of composed resources.
```

```
crossplane beta render xr.yaml composition.yaml functions.yaml -o observe/                                  
crossplane: error: cannot load observed composed resources from ".up/examples/observe/": cannot load YAML stream from file: no YAML files found in ".up/examples/observe/" (.yaml or .yml)
```

this is working:

```
crossplane beta render xr.yaml composition.yaml functions.yaml -o observe/observe.yaml 
```

Crossplane version:

phisco:

Fixes #4854.

`filepath.Glob` was not properly handling the `*.{yaml,yml}` format, so I switched to manually iterating over the directory entries.

We should switch to an afero.FS so that we can write some proper tests for this, but that can be done in a separate PR.

I have: 

### DIFF
--- a/cmd/crank/beta/render/load.go
+++ b/cmd/crank/beta/render/load.go
@@ -19,10 +19,9 @@ package render
 import (
 	"bufio"
 	"io"
-	"os"
 	"path/filepath"
-	"strings"
 
+	"github.com/spf13/afero"
 	"k8s.io/apimachinery/pkg/util/yaml"
 
 	"github.com/crossplane/crossplane-runtime/pkg/errors"
@@ -34,8 +33,8 @@ import (
 )
 
 // LoadCompositeResource from a YAML manifest.
-func LoadCompositeResource(file string) (*composite.Unstructured, error) {
-	y, err := os.ReadFile(file) //nolint:gosec // Taking this input is intentional.
+func LoadCompositeResource(fs afero.Fs, file string) (*composite.Unstructured, error) {
+	y, err := afero.ReadFile(fs, file)
 	if err != nil {
 		return nil, errors.Wrap(err, "cannot read composite resource file")
 	}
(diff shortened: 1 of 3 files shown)


cmd/crank/beta/render/load.go
```diff
@@ -129,6 +131,24 @@ func LoadYAMLStream(fileOrDir string) ([][]byte, error) {
...
+	}
+	for _, entry := range entries {
+		if entry.IsDir() {
+			// We don't care about nested directories.
+			continue
+		}
+		if ext := strings.ToLower(filepath.Ext(entry.Name())); ext == ".yaml" || ext == ".yml" {
+			yamls = append(yamls, filepath.Join(dir, entry.Name()))
+		}
+	}
```
turkenh, line 148: Looks like `filepath.Glob` not working with multiple extensions, we could still leverage it here as follows:

```suggestion
	var yamls []string
	var ext := []string{"yaml", "yml"}

	for _, e := range ext {
		y, err := filepath.Glob(filepath.Join(dir, "*."+e))
		if err != nil {
			return nil, err
		}
		yamls = append(yamls, y...)
	}
	
	return yamls, nil
```

phisco: I also thought it would have been clearer, but then I realised this would result in all `.yml` files being sorted after all `.yaml` files in case of mixed usage, which depending on the caller's usage of the slice returned, could be undesirable, and didn't make much sense to just sort the files later. `ReadDir` [guarantees](https://golang.org/pkg/os#ReadDir) we iterate over entries in alphabetical order.

Both [ReadDir](https://pkg.go.dev/github.com/spf13/afero#ReadDir) and [Glob](https://pkg.go.dev/github.com/spf13/afero#Glob) have their counterpart in afero, so any of the two approaches would be ok w.r.t. that.

The other option I've considered was going with WalkDir, which would allow to support subdirectories easily, but I wanted to preserve the current "glob-like" behaviour for now.

If you feel strongly about the `filepath.Glob` approach I can switch, LMK.

turkenh (approved):

cmd/crank/beta/render/load.go
phisco: 👍  I'll just add a TODO to revisit it when we move to an `afero.FS` and some more context

phisco:
In the end I switched to `afero.FS` and added some tests, so please give it another round of review @turkenh 🙏

cmd/crank/beta/render/load.go
turkenh: Looks like this is still open - I couldn't see a related comment or Todo (which is no longer a todo though)

turkenh (approved):
Thanks @phisco 🙌

cmd/crank/beta/render/load.go
phisco: Yes, you are right, got lost in the refactoring, let me add a comment and then I'll merge this.

Issue #4796, opened 2023-10-12
Figure out CI for Composition Functions


negz:
### What problem are you facing?

We want Functions to be pretty simple, so we're explicitly not using https://github.com/upbound/build (or even Makefiles). We also expect Functions to use other languages in future that https://github.com/upbound/build does not support today, including Python.

At the same time, we probably do want CI for most Functions, to handle things like:

* Automatically running a linter and tests, if any.
* Building and pushing the Function to a registry (e.g. xpkg.upbound.io)

Usually https://github.com/upbound/build does this for Crossplane repos. I think we need a simpler pattern. 

### How could Crossplane help solve your problem?

I'm thinking we put together a very stripped down version of [`ci.yml`](https://github.com/crossplane/crossplane/blob/2c5e7f07ba9e3d83d1c85169bbde685de8514ab8/.github/workflows/ci.yml). How it invokes tests and linting would differ from language to language, but hopefully building and pushing Functions would just invoke the `crossplane` CLI.

negz:
I think we need to do this by the v1.14 release date, but not by feature freeze. The only interaction with feature freeze is making sure `crossplane` does what we need it to do in time - e.g. build and push Functions.

negz:
First step would be adding CI to https://github.com/crossplane/function-template-go.

turkenh:
Can we close this ticket, anything left here? 
Asking because this is one of the two issues left open in [v1.14 milestone](https://github.com/crossplane/crossplane/milestones/v1.14).

negz:
I'm waiting for #4885 to be merged before closing.

PR #4885, opened 2023-10-24
Add a one-pager guiding Function build tooling

negz:

Fixes https://github.com/crossplane/crossplane/issues/4796

I think as we build templates for languages like Go, Python, and TypeScript it will be useful to have a framework/philosophy for our patterns and tooling choices. The ultimate goal here is to make building a Function in a language you already know and work with regularly a familiar, unintimidating experience.

See https://github.com/crossplane/function-template-go/pull/23 for a reference implementation.

I have: 

### DIFF
--- a/design/one-pager-composition-function-build-tooling.md
+++ b/design/one-pager-composition-function-build-tooling.md
@@ -1,0 +1,147 @@
+# Building A Composition Function
+
+* Owner: Nic Cope (@negz)
+* Reviewers: Jared Watts (@jbw976)
+* Status: Draft
+
+## Background
+
+A Composition Function extends Crossplane to support a new way of configuring
+how to reconcile a composite resource (XR). Each Function is a gRPC server.
+Crossplane sends state to a Function via a `RunFunctionRequest` RPC. The
+Function is intended to return updated state via a `RunFunctionResponse` RPC.
+
+Part of the goal of Composition Functions is to allow Crossplane users to build
+Functions using their general purpose programming language (GPL) of choice. I
+think there will be broadly two types of Function:
+
+1. Generic, reusable Functions. These Functions support a common use case.
+   They're intended to work with any kind of XR, and often expose a domain
+   specific language (DSL) that their users must use to express Composition
+   logic.
+   [`function-patch-and-transform`][function-patch-and-transform] and
+   [`function-auto-ready`][function-auto-ready] are examples of generic
+   Functions.
+2. Purpose-build Functions. These Functions are strongly coupled to a the schema
+   of a specific kind of XR. Rather that exposing a configuration DSL, the
+   Function's GPL code _is_ the Composition logic.
+
+Take a look at the [Composition Functions design document][design-doc] for more
+context on Functions. Quoting from that document:
+
+> I believe it's critical that Composition Functions are easy to develop - much
+> easier than developing a Crossplane Provider.
+
+Another way to think about this is that the developer experience must scale.
+Some Functions will be software engineering projects. They'll be maintained by a
+team of contributors, have unit and end-to-end tests, release branches,
+continuous integration (CI), etc.
+
+At the other, simpler, end of the scale many Functions will be more like
+configuration that happens to be expressed using a GPL. For someone writing a
+Function like this, needing to learn and use a new set of build and CI tools
+is a potentially huge barrier to entry.
+
+The emergent process for developing a Composition Function is:
+
+1. Scaffold a new Function from a template using `crossplane beta xpkg init`.
+2. Add your logic to the scaffolded project (e.g. edit `fn.go`).
+3. Optionally, add and run unit tests for your Function logic.
+4. Optionally, test your Function end-to-end using `crossplane beta render`
+
+Once you're satisfied that your Function works end-to-end it's time to make it
+available to install and use in your Crossplane control plane. To do this you
+must:
+
+1. Build your Function's "runtime" - the OCI image used to deploy it.
+2. Build a package from the OCI image runtime using `crossplane xpkg build`.
+3. Push your package to an OCI registry using `crossplane xpkg push`
+
+This document proposes a set of guiding principles for the Composition Function
+developer experience, with a particular focus on build and CI. Ultimately
+Functions are open-ended enough that a developer could choose whatever path best
+suits them, so consider this a proposed "golden path". This golden path will
+inform what choices we make in Function template repositories like
+[`function-template-go`][function-template-go], and thus establish patterns for
+the broader Function ecosystem.
+
+## Proposal
+
+I propose that we strive to keep the set of tools and technologies a Crossplane
+user must learn to write a Function as small at possible. Learning Crossplane
+alone is hard enough - we don't want users to also need to learn a new build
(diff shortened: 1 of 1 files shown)


TerjeLafton:
Very nice! I really agree with the mindset of keeping functions as simple as possible.

But I would like to add that good scaffolding shouldn't be a replacement for good documentation. Not that I think that it is what is going to happen, I just feel like it has a tendency to happen in IT.

We have recently been developing both a provider and a function, and while it has been a lot of fun, it was usually hard to get started because of the scaffolding. The templates have you jump right into it, and it can be overwhelming for a new user trying to get started, especially if the documentation for the template is limited.

This might be an unpopular opinion, but I am not sure the template should even include CI, linting and things like that. It's great if the documentation says you should have linting, and it can be suggested both tools and configuration, but the scaffolding should be kept minimal.

JonasKs:
I completely agree with @TerjeLafton here (which might be obvious since we're working together on writing functions and providers), and therefor disagree with @stevendborrelli. 

I think a `makefile`, and the `build`-submodule is bad thing. It's very hard to understand, and as a newcommer to the ecosystem it feels like makefiles and templates have been created instead of documentation, instead of an addition to it. 

When we wanted to build using `podman` in our own pipelines, we couldn't find any documentation of what needed to be done, or how to do it. The only thing we could find was the [make -j2 build.all](https://github.com/crossplane/provider-template/blob/main/.github/workflows/ci.yml#L236) command. @TerjeLafton set out on a mission to de-cypher it, and you can see the result here:  
image

This is not a good DX.

In my opinion, docs should be written like [FastAPI does it](https://fastapi.tiangolo.com/tutorial/first-steps/), and then templates and makefiles can come after - but the target audience for templates/makefiles should not be first timers/beginners. We found it very hard, since a lot of code is generated, and the makefiles(and `build` submodule) took us days to understand. We wished there was a tutorial or documentation that took us through it all from start instead. 

We'll definitely contribute with docs when our project has come to a more mature state.

negz:
@TerjeLafton, @JonasKs - Thanks for the feedback! I definitely don't intend for this to be a stand-in for documentation. I agree we need both.

@stevendborrelli Would it help if the commands needed to lint, test, build etc a Function appeared in the `README`?

negz:
> This might be an unpopular opinion, but I am not sure the template should even include CI, linting and things like that. It's great if the documentation says you should have linting, and it can be suggested both tools and configuration, but the scaffolding should be kept minimal.

@TerjeLafton I could be convinced to drop linting. Personally though I do think our Go linter config at least helps catch real bugs, not just style things, so it is likely to result in folks writing less buggy Functions faster.

For CI, I think we need to set a direction. My expectation is that pushing a Function to xpkg.upbound.io to get it listed in https://marketplace.upbound.io will be a pretty common requirement, so I kinda want to have that be smooth and somewhat opinionated. FWIW the current CI setup is basically a no-op if you don't create an `XPKG_ACCESS_ID` secret in your GitHub repo. It will just lint, test, and build everything and attach the `xpkg` files to your GitHub workflow output without pushing anything anywhere.

jbw976 (commented):
This is pretty reasonable to me @negz! The feedback so far is fairly valid too and I'm seeing the high level themes fall into two camps:

* give developers granular commands that they can easily understand, enhance their productivity, and easily get to a high quality artifact
* automate as much as possible (e.g. with `make`) so developers don't have to think about the build details and can just focus on their code and getting that shipped ASAP

The spirit of both of these are valid and I'm wondering if there's a happy intermediate ground. Even though I like everything the crossplane `make` file automates for us, I agree that it's very challenging for just about anyone to understand, troubleshoot, or extract useful information out of it, e.g., fine grained commands to reuse in isolation.

My instincts so far is that the happy middle ground could be:

* provide all the useful (i.e. real world) commands for building/testing/publishing in the README of the templates - developers can use these locally with reasonable productivity and high understanding of the build process
  * perhaps these can be scripted a bit too, but not with `make`.
* codify the golden paths for CI in GitHub actions - this provides productive repeatable automation for shipping high quality artifacts

One potential gap I see right now is another type of "middle ground" - the e2e testing scenario in between:

* `go run`, `go test`
* CI publishing of production ready artifacts for real world usage

Is there a story for that scenario, where the developer can easily and build a test image locally on their laptop to test in their local cluster?

TerjeLafton:
>For CI, I think we need to set a direction. My expectation is that pushing a Function to xpkg.upbound.io to get it listed in https://marketplace.upbound.io/ will be a pretty common requirement, so I kinda want to have that be smooth and somewhat opinionated. FWIW the current CI setup is basically a no-op if you don't create an XPKG_ACCESS_ID secret in your GitHub repo. It will just lint, test, and build everything and attach the xpkg files to your GitHub workflow output without pushing anything anywhere.

I think all of this depends on what documentation is available. Like @JonasKs said, it would be really nice to have better documentation for both how to create a Provider and a Function; either from scratch or with very minimal scaffolding, kind of like the FastAPI docs he linked.
Could this be available on `docs.crossplane.io` as a `tutorial` section, maybe? This is even something we would love to contribute to!

If these docs are available like described above, I think it would be totally fine to have a full fledged template with opinionated CI and linting, which can be really nice for those who are creating their nth function. 

Neither of these should exist alone, but they usually complement each other very well. 
* Good documentation for new users to learn and see just how easy it is
* Good template for experienced users who just needs to get started with their project

pedjak:
> As someone building functions, the lack of a Makefile is frustrating. I create a pull request and it comes back with a bunch of linting and "gci" errors that I could have resolved locally. gci in particular will rewrite the import order of many .go files.

@stevendborrelli I have proposed something like what you are describing in https://github.com/crossplane/function-template-go/pull/16 please have a look. @JonasKs Makefile in that PR is just a tiny wrapper around multi-stage Dockerfile, just to make it simpler to invoke it. I agree with you that we should not include build submodule.

negz:
My feeling is that a Makefile should be opt-in. For the template everyone starts from, less is probably better.

It's totally reasonable for a function to adopt the whole build submodule if the authors want to. For example I could see us doing this for something like function-patch-and-transform, where the majority of maintainers are folks who use that build submodule everywhere else.

negz:
So we've had a few folks build functions, in Go, so far. I've also started experimenting with Python in https://github.com/negz/function-auto-python. Based on these experiences I'm feeling pretty good about what is proposed here. Unless we think it's way off the mark my preference would be to merge it and iterate.

negz:
> Is there a story for that scenario, where the developer can easily and build a test image locally on their laptop to test in their local cluster?

@jbw976 Is `crossplane beta render` covering that scenario well enough? I've also updated the Go template repo README to include instructions on how to locally build a package: https://github.com/crossplane/function-template-go

jbw976:
yes @negz I think the scenario of building a package to run on a local cluster is covered well enough now - I think when I made that comment initially we didn't really have the instructions in place for how to build the package, but now we do and it seems sufficient enough for when you wanna go past what `render` can currently and use a local test cluster via a package.

jbw976 (approved):

phisco (approved):
I agree with the approach, at least as a starting point, we can evolve it in the future if needed.

PR #4887, opened 2023-10-24
feat(crank/render): enforce function and composition type

### LINKED ISSUE #4852: Validate `crossplane beta render` inputs

Currently, the following order works as expected:
```
crank beta render xr.yaml composition.yaml functions.yaml
```

However, this order doesn't work:
```
crank beta render composition.yaml xr.yaml functions.yaml
```

The error message received in the latter case is:

```
error: crossplane: error: invalid Composition "dev-cluster-a": spec.resources: Required value: an array of resources is required in Resources mode (the default if no mode is specified)
```

We need to implement input validation for the "crank beta render" command or add specifics for the file 
```
crank beta render -xr xr.yaml -composition composition.yaml -function functions.yaml
```

phisco:

Fixes https://github.com/crossplane/crossplane/issues/4852.

I have: 

### DIFF
--- a/cmd/crank/beta/render/load.go
+++ b/cmd/crank/beta/render/load.go
@@ -53,7 +53,15 @@ func LoadComposition(fs afero.Fs, file string) (*apiextensionsv1.Composition, er
 		return nil, errors.Wrap(err, "cannot read composition file")
 	}
 	comp := &apiextensionsv1.Composition{}
-	return comp, errors.Wrap(yaml.Unmarshal(y, comp), "cannot unmarshal composition resource YAML")
+	if err := yaml.Unmarshal(y, comp); err != nil {
+		return nil, errors.Wrap(err, "cannot unmarshal composition resource YAML")
+	}
+	switch gvk := comp.GroupVersionKind(); gvk {
+	case apiextensionsv1.CompositionGroupVersionKind:
+		return comp, nil
+	default:
+		return nil, errors.Errorf("not a composition: %s/%s", gvk.Kind, comp.GetName())
+	}
 }
 
 // TODO(negz): Support optionally loading functions and observed resources from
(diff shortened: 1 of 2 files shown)


cmd/crank/beta/render/load.go
```diff
@@ -53,7 +53,15 @@ func LoadComposition(fs afero.Fs, file string) (*apiextensionsv1.Composition, er
...
 	comp := &apiextensionsv1.Composition{}
-	return comp, errors.Wrap(yaml.Unmarshal(y, comp), "cannot unmarshal composition resource YAML")
+	if err := yaml.Unmarshal(y, comp); err != nil {
+		return nil, errors.Wrap(err, "cannot unmarshal composition resource YAML")
+	}
+	switch gvk := comp.GroupVersionKind(); gvk {
+	case apiextensionsv1.CompositionGroupVersionKind:
+		return comp, nil
+	default:
+		return nil, errors.Errorf("not a composition: %s/%s", gvk.Kind, comp.GetName())
```
turkenh, line 63:
```suggestion
		return nil, errors.Errorf("expecting a \"Composition\" as the 2nd argument, but found: %s/%s", gvk.Kind, comp.GetName())
```

This could make the message more useful?

cmd/crank/beta/render/load.go
```diff
@@ -72,7 +80,12 @@ func LoadFunctions(filesys afero.Fs, file string) ([]pkgv1beta1.Function, error)
 		if err := yaml.Unmarshal(y, f); err != nil {
 			return nil, errors.Wrap(err, "cannot parse YAML Function manifest")
 		}
-		functions = append(functions, *f)
+		switch gvk := f.GroupVersionKind(); gvk {
+		case pkgv1beta1.FunctionGroupVersionKind:
+			functions = append(functions, *f)
+		default:
+			return nil, errors.Errorf("not a function: %s/%s", gvk.Kind, f.GetName())
```
turkenh, line 87:
```suggestion
			return nil, errors.Errorf("expecting a \"Function\" as the 3rd argument, but found: %s/%s", gvk.Kind, f.GetName())
```

turkenh (approved):

cmd/crank/beta/render/load.go
phisco: The caller is already wrapping the error adding some more information (`cannot load functions from <filename>`) and we print the usage text in case of errors too, so I think there is enough context to spot the issue, I would rather avoid leaking the argument position down to this level as it feels brittle. WDYT?

turkenh: Yeah, that sounds good then.

Issue #4581, opened 2023-09-06
Clear Ownership of Fields Between Claim and Composite


turkenh:
### What happened?

Claims are namespaced resources that (create and) bind to cluster-scoped Composite resources by propagating the spec provided by the user. Being created by the same schema defined in the CompositeResourceDefinition (XRD), they contain the same fields under spec and status in addition to the ones [dedicated](https://github.com/crossplane/crossplane/blob/226b81f575f2cf68986da827e87b5ba34aa5121f/internal/xcrd/schemas.go#L70) to each type. 

Today, in the claim controller, we merge the spec/status of the claim instead of clear ownership:

- Claim → composite: We [merge](https://github.com/crossplane/crossplane/blob/a888ea7657f12d2dd9effa5ef8da8b17b0bea404/internal/controller/apiextensions/claim/configurator.go#L234) the spec of the claim with the spec of the composite. (causing https://github.com/crossplane/crossplane/issues/3421)
- Claim → Composite: We [add](https://github.com/crossplane/crossplane/blob/a888ea7657f12d2dd9effa5ef8da8b17b0bea404/internal/controller/apiextensions/claim/configurator.go#L97-L98) (i.e. merge) to the existing labels/annotations on composites instead of overwriting with what is in the claim. (causing https://github.com/crossplane/crossplane/issues/4315)
- Composite → Claim: We [merge](https://github.com/crossplane/crossplane/blob/a888ea7657f12d2dd9effa5ef8da8b17b0bea404/internal/controller/apiextensions/claim/configurator.go#L192) the status of the claim with the status of the composite. (causing https://github.com/crossplane/crossplane/issues/3542)

### How can we reproduce it?

See reproduction steps under the corresponding directory at: https://github.com/turkenh/xp-ssa-experiments

### How could Crossplane help solve your problem?

I propose we define and stick to clear ownership of fields between claims and composites without no merging:

- Spec goes from claim to composite.
- Status goes from composite to claim.
- Annotation/Labels go from claim to composite.

The last one could be a bit controversial, and we may want to enable some ways to enable propagating metadata back (as requested in https://github.com/crossplane/crossplane/issues/3180) /cc @phisco 

This would also enable switching to [Server Side Apply](https://github.com/crossplane/crossplane/issues/4047) in the claim controller, instead of the Patching applicator. /cc @pedjak

negz:
[...]

Issue #4315, opened 2023-07-11
Removing annotation in Claim does not remove the corresponding annotation in XR


ZhiminXiang:
### What happened?
I added an annotation to a Claim, which populated the annotation to XR, which is expected.

However, after I deleted the annotation from the Claim, the annotation in XR was still preserved.

### How can we reproduce it?
I think this could easily reproduced by using any crossplane example.

### What environment did it happen in?
Crossplane version: 1.12.0

ZhiminXiang:
I am wondering if this is related to https://github.com/crossplane/crossplane/issues/4047.

ZhiminXiang:
Also we passed the annotation from XR to the managed resources. Looks like removing the annotation in the XR does not remove the corresponding annotation in the managed resources.

negz:
> I am wondering if this is related to https://github.com/crossplane/crossplane/issues/4047.

My guess is that it is.

sttts:
Yes, SSA should assign the XR controller as owner, and as soon as that controller stops applying that field (annotation) it should go away. cc @pedjak

pedjak:
Fixed via https://github.com/crossplane/crossplane/pull/4896.

Issue #4036, opened 2023-04-28
appendSlice does not remove items not in list


bradkwadsworth-mw:
### What happened?
If I create a patch that uses `appendSlice: true` to merge items in a list in a definition everything works fine. But, if I remove an item in that list in my XR, the item is not removed in the composed resource.Would it be possible to remove non-existent items in an XR that is merged/overwritten in the composed resource?

### How can we reproduce it?
definition.yaml
```yaml
...
cluster:
  type: object
  properties:
    authorizedNetworks:
    type: array
    items:
      type: object
      description: Authorized networks to connect to Kube API.
      required:
      - cidrBlock
      properties:
        cidrBlock:
          type: string
...
```

composition.yaml
```yaml
...
base:
  apiVersion: container.gcp.upbound.io/v1beta1
  kind: Cluster
  spec:
    forProvider:
      masterAuthorizedNetworksConfig:
      - cidrBlocks:
        - displayName: bastion
patches:
- type: FromCompositeFieldPath
   fromFieldPath: spec.parameters.bastion.subnet.cidr
   toFieldPath: spec.forProvider.masterAuthorizedNetworksConfig[0].cidrBlocks[0].cidrBlock
- type: FromCompositeFieldPath
   fromFieldPath: spec.parameters.cluster.authorizedNetworks
   toFieldPath: spec.forProvider.masterAuthorizedNetworksConfig[0].cidrBlocks
   policy:
     mergeOptions:
        appendSlice: true
...
```

XR
```yaml
...
spec:
  parameters:
    cluster:
      authorizedNetworks:
      - cidrBlock: 5.6.7.8/32
...
```

composed resource
```yaml
...
apiVersion: container.gcp.upbound.io/v1beta1
kind: Cluster 
spec:
  forProvider:
    masterAuthorizedNetworksConfig:
    - cidrBlocks:
       - cidrBlock: 10.2.16.0/28
          displayName: bastion
       - cidrBlock: 5.6.7.8/32
...
```
modified XR
```yaml
...
spec:
  parameters:
    cluster:
      authorizedNetworks: []
...
```

composed resource after edit
```yaml
...
apiVersion: container.gcp.upbound.io/v1beta1
kind: Cluster 
spec:
  forProvider:
    masterAuthorizedNetworksConfig:
    - cidrBlocks:
       - cidrBlock: 10.2.16.0/28
          displayName: bastion
       - cidrBlock: 5.6.7.8/32
...
```

### What environment did it happen in?
[...]

PR #4896, opened 2023-10-25
Fix synchronization between claim and the counterpart composite

pedjak:

Removing of a claim annotation, label, field and/or array member is properly propagated to the counterpart composite by using [k8s server-side apply](https://kubernetes.io/docs/reference/using-api/server-side-apply/) feature. The same approach is used by propagating composite changes back to the claim.

Implementing the server-side apply requested the following changes on the claim reconciler:

* Given that a submitted patch needs to contain all fields that a reconciler care about and wants to take the ownership, the existing configurators and binders cannot perform updates. Instead, they are only adding the requested changes to the patch, that gets applied in one shot.
* The claim configurator requires access to actual claim and composite, but  produces a claim patch, similar goes for the composite configurator. Hence, `Configurator` was not applicable anymore got removed.
* Given that `WithClaimConfigurator` and `WithCompositeConfigurator` are used only for mocking in the unit tests, their visibility is reduced.
* Field manager name follows the convention: `crossplane-claim-controller/<CLAIM_GROUP_KIND_LOWERCASE>`
* Claim status subresource is still modified by sending UPDATE requests, because the controller should own it completely. That is expressed explicitly by setting the status field ownership.
* Given that `Applicator` is not used anymore, `Reconciler.client` is redeclared to `client.Client`
* Unit tests reworked/updated to verify the new implementation

Following the proposal stated in #4581, the following composite fields are propagated back to the claim:
* `spec.compositionRef`
* `spec.compositionSelector`
* `spec.compositionUpdatePolicy`
* `spec.compositionRevisionSelector`
*  `status` minus `status.conditions`

Fixes:
* #3421
* #3542
* #3992
* #4026
* #4036
* #4315

An e2e test added for verifying the above fixes and catching future regressions.

Part of #4047 epic.

Implements: #4581

I have: 


pedjak:
Note to reviewers: this PR depends on https://github.com/crossplane/crossplane/pull/4780 and it still need to be merged separately. All changes relevant to this PR are in [this commit](https://github.com/crossplane/crossplane/pull/4896/commits/99b206feba84e3f0325ebbfbdc79c9676d06cf11).

internal/controller/apiextensions/claim/configurator.go
```diff
@@ -125,35 +126,37 @@ func ConfigureComposite(_ context.Context, cm resource.CompositeClaim, cp resour
 	}
 
 	claimSpecFilter := xcrd.GetPropFields(wellKnownClaimFields)
-	ucp.Object["spec"] = filter(spec, claimSpecFilter...)
+	cpPatch.(*composite.Unstructured).Object["spec"] = filter(spec, claimSpecFilter...)
```
sttts, line 129: this is ugly. Why not use that type for `cpPatch` directly? A type cast is always suspicious, unguarded even more so.

internal/controller/apiextensions/claim/configurator.go
```diff
@@ -185,21 +188,36 @@ func filter(in map[string]any, keys ...string) map[string]any {
 	return out
 }
 
-// APIClaimConfigurator configures the supplied claims with fields
-// from the composite. This includes late-initializing spec values
-// and updating status fields in claim.
-type APIClaimConfigurator struct {
-	client client.Client
-}
+func keep(in map[string]any, keys ...string) map[string]any {
```
sttts, line 191: now that I see `keep`, `filter` looks like a bad name for what it does. Something like `withoutKeys` and `onlyKeys` would be clearer.

internal/controller/apiextensions/claim/configurator.go
```diff
@@ -209,58 +227,69 @@ func (c *APIClaimConfigurator) Configure(ctx context.Context, cm resource.Compos
 		return nil
 	}
 
-	if err := merge(ucm.Object["status"], ucp.Object["status"],
+	udesiredCm, ok := cmPatch.(*claim.Unstructured)
+	if !ok {
+		return nil
```
sttts, line 232: what does this mean? Similar to above: don't cast. Use the proper types directly.

internal/controller/apiextensions/claim/reconciler.go
```diff
@@ -458,25 +420,38 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
 
-	if err := r.claim.AddFinalizer(ctx, cm); err != nil {
-		if kerrors.IsConflict(err) {
-			return reconcile.Result{Requeue: true}, nil
+	// If composite exists, i.e. got created in the previous iteration,
+	// it might be needed to fix field ownerships,
+	// so that the claim controller becomes an exclusive owner
+	// of the fields mirrored from the claim to the composite
+	if meta.WasCreated(cp) && r.maybeFixFieldOwnership(cp) {
+		if err := r.client.Update(ctx, cp); err != nil {
```
sttts, line 436: Does this mean that annotations, labels, finalizers are now all owned by the controller? That sounds wrong and dangerous.

internal/controller/apiextensions/claim/reconciler.go
```diff
@@ -489,45 +464,49 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+		record.Event(cm, event.Warning(reasonClaimConfigure, err))
 		cm.SetConditions(xpv1.ReconcileError(err))
-		return reconcile.Result{Requeue: true}, errors.Wrap(r.client.Status().Update(ctx, cm), errUpdateClaimStatus)
+		return reconcile.Result{Requeue: true}, errors.Wrap(r.client.Status().Update(ctx, cm, r.fieldOwner), errUpdateClaimStatus)
+	}
+
+	// The following patch operation is going to override the status part of the claim patch
+	// with the status of the actual claim version.
+	// given that the status update come later, we need to preserve it temporarily.
+	desiredClaimStatus := cmPatch.(*claim.Unstructured).Object["status"]
```
sttts, line 494: annother unnecessary and dangerous cast.

internal/controller/apiextensions/claim/reconciler.go
sttts: How can the update know which fields owned by other managers should be taken over?

sttts: Oh, got it. Please move `r.maybeFixFieldOwnership(cp)` into its own if clause to make it clear that it has side-effects.

internal/controller/apiextensions/claim/configurator.go
sttts: I see why you do that. Still ugly, but I don't see a better way.

internal/controller/apiextensions/claim/reconciler.go
pedjak: done.

internal/controller/apiextensions/claim/configurator.go
pedjak: Me neither, but I see this sort of casting often in this part of the codebase. This gives me a signal that we have just one implementation of `resource.Composite` or `resource.CompositeClaim` interface and perhaps it would be better to use it directly where needed. Using casts mean that we have single data model, and hence there is no much benefit of hiding it behind some interfaces.

pedjak: I have removed type casting in [this commit](https://github.com/crossplane/crossplane/pull/4896/commits/e44abc93a87207d1ffb69e299fa8a52925de6ec4). Let me know what do you think. I like it.

internal/controller/apiextensions/claim/configurator.go
```diff
@@ -183,82 +179,94 @@ func filter(in map[string]any, keys ...string) map[string]any {
...
 	}
 
 	// CompositionRevision is a special field which needs to be propagated
 	// based on the Update policy. If the policy is `Automatic`, we need to
 	// overwrite the claim's value with the composite's which should be the
 	// `currentRevision`
-	if cp.GetCompositionUpdatePolicy() != nil && *cp.GetCompositionUpdatePolicy() == xpv1.UpdateAutomatic {
-		cm.SetCompositionRevisionReference(cp.GetCompositionRevisionReference())
+	if cp.GetCompositionUpdatePolicy() != nil &&
+		*cp.GetCompositionUpdatePolicy() == xpv1.UpdateAutomatic && cp.GetCompositionRevisionReference() != nil {
```
turkenh, line 248: With this added condition `cp.GetCompositionRevisionReference() != nil` here, we will only set `compositionRevisionReference` once in the beginning instead of keeping it always up to date when the policy Automatic.

internal/controller/apiextensions/claim/reconciler.go
```diff
@@ -539,7 +555,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		// We should be watching the composite resource and will have a
 		// request queued if it changes, so no need to requeue.
 		cm.SetConditions(Waiting())
-		return reconcile.Result{}, errors.Wrap(r.client.Status().Update(ctx, cm), errUpdateClaimStatus)
+		return reconcile.Result{}, errors.Wrap(r.client.Status().Update(ctx, cm, r.fieldOwner), errUpdateClaimStatus)
 	}
 
 	record.Event(cm, event.Normal(reasonBind, "Successfully bound composite resource"))
```
turkenh, line 561: Not related to this PR, but is this event in correct place? Claim bound to composite before this line and this should rather be something like "Composite is ready" 🤔

internal/controller/apiextensions/claim/reconciler.go
```diff
@@ -323,26 +289,26 @@ func WithPollInterval(after time.Duration) ReconcilerOption {
...
 		},
-		composite: defaultCRComposite(c),
-		claim:     defaultCRClaim(c),
-		log:       logging.NewNopLogger(),
-		record:    event.NewNopRecorder(),
+		composite:  defaultCRComposite(c),
+		claim:      defaultCRClaim(c),
+		log:        logging.NewNopLogger(),
+		record:     event.NewNopRecorder(),
+		fieldOwner: client.FieldOwner(fmt.Sprintf("crossplane-claim-controller/%s", strings.ToLower(schema.GroupVersionKind(of).GroupKind().String()))),
```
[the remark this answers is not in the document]
pedjak: > I don't see a value with including group/kind into the field owner. 

Having the same owner name shared by all claim controllers could lead to implicit shared ownership between controllers and degrades the user experience. We have a controller per specific claim type, and browsing through the kubernetes code base shows that controllers set an unique field manager name, a few examples:

* https://github.com/kubernetes/kubernetes/blob/d61cbac69aae97db1839bd2e0e86d68f26b353a7/pkg/controller/validatingadmissionpolicystatus/controller.go#L38
* https://github.com/kubernetes/kubernetes/blob/d61cbac69aae97db1839bd2e0e86d68f26b353a7/pkg/controller/clusterroleaggregation/clusterroleaggregation_controller.go#L141
* https://github.com/kubernetes/kubernetes/blob/d61cbac69aae97db1839bd2e0e86d68f26b353a7/pkg/controlplane/controller/defaultservicecidr/default_servicecidr_controller.go#L45
* https://github.com/kubernetes/kubernetes/blob/d61cbac69aae97db1839bd2e0e86d68f26b353a7/pkg/controller/resourceclaim/controller.go#L66

Including group/kind as a part of the field owner name is an easy way to establish the name uniqueness and at the same time it gives an user a clear information about the field ownership, e.g. when user performs something like:

```sh
kubectl get foo --show-managed-fields
```

We use similar strategies when declaring controller name for the log context. Ofcourse, we can polish the proposed convention if needed, but we should keep an unique name per claim controller.

> Proposing just to use apiextensions.crossplane.io/claim also to be consistent with the Composite's: 

Based on the above explanation, I was about to propose the convention in followup PRs that are about fixing composite -> managed resources synchronization, i.e. we should introduce an unique name per composite controller.

internal/controller/apiextensions/claim/reconciler.go
```diff
@@ -559,7 +575,48 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+// way to figure out if the generated name is already taken.
+// If the operation is successful, managed fields operation will be set to "Update".
+// Later patches are going to produce another managed fields set for the same manager,
+// but with a different operation "Apply".
+// Such configuration would prevent removal of claim fields on the composite, i.e.
+// it would only remove the ownership.
+// In order to fix that, we need to manually change operation to "Apply",
+// before the first composite patch is sent to k8s api server.
+// Returns true if the ownership was fixed.
+func (r *Reconciler) maybeFixFieldOwnership(obj *unstructured2.Unstructured, fixSubresource bool) bool {
```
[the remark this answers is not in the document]
pedjak: > What exactly do we lose without this migration?

For claims created with the previous version of crossplane, we are going to establish shared field ownership between the new manager and the old one (even when server-side apply is not used, k8s sets the field manager on claim/composite). 

For example, if we create [claim](https://github.com/crossplane/crossplane/blob/44eb7ed980a56b8db8c8a1cb8ba7cf285882f9bf/test/e2e/manifests/apiextensions/composition/propagate-field-removals/claim.yaml) with the latest released Crossplane version and perform:

```sh
kubectl get xnopresources.nop.example.org --show-managed-fields -o yaml
```

K8s API server sers the managed fields:

```yaml
managedFields:
- apiVersion: nop.example.org/v1alpha1
  fieldsType: FieldsV1
  fieldsV1:
    f:metadata:
      f:annotations:
        .: {}
        f:test/bar: {}
        f:test/foo: {}
      f:finalizers:
        .: {}
        v:"composite.apiextensions.crossplane.io": {}
      f:generateName: {}
      f:labels:
        .: {}
        f:bar: {}
        f:crossplane.io/claim-name: {}
        f:crossplane.io/claim-namespace: {}
        f:crossplane.io/composite: {}
        f:foo: {}
    f:spec:
      .: {}
      f:claimRef:
        .: {}
        f:apiVersion: {}
        f:kind: {}
        f:name: {}
        f:namespace: {}
      f:compositionRef:
        .: {}
        f:name: {}
      f:compositionRevisionRef:
        .: {}
        f:name: {}
      f:compositionUpdatePolicy: {}
      f:coolField: {}
      f:numbers: {}
      f:parameters:
        .: {}
        f:tags:
          .: {}
          f:newtag: {}
          f:tag: {}
      f:resourceRefs: {}
  manager: crossplane
  operation: Update
  time: "2023-11-23T12:16:07Z"
- apiVersion: nop.example.org/v1alpha1
  fieldsType: FieldsV1
  fieldsV1:
    f:status:
      .: {}
      f:conditions:
        .: {}
        k:{"type":"Ready"}:
          .: {}
          f:lastTransitionTime: {}
          f:reason: {}
          f:status: {}
          f:type: {}
        k:{"type":"Synced"}:
          .: {}
          f:lastTransitionTime: {}
          f:reason: {}
          f:status: {}
          f:type: {}
      f:coolerField: {}
  manager: crossplane
  operation: Update
  subresource: status
  time: "2023-11-23T12:16:07Z"
```

So, even we use patching applicator, the operation set by API server is `Update`. With introducing server-side apply, we are going to get another field manager listed here, with the operation set to `Apply`. Two field managers are going to share the same set of fields.

As the consequence, if we remove field/label/annotation on the existing claim, that change is NOT going to propagate down to the composite, i.e. the reported bugs are still valid and not fixed.

IMHO, Crossplane behavior should be the same both on existing and the new claims. How is user supposed to understand that label removals are propagated only on claims created by Crossplane version using server-side apply?

> But we need this for new claims as well since we use CREATE

Even if we stop using create, the issue is still valid for the existing claims, and `maybeFixFieldOwnership` takes care about that.

internal/controller/apiextensions/claim/configurator.go
```diff
@@ -183,82 +179,94 @@ func filter(in map[string]any, keys ...string) map[string]any {
...
+	// If existing claim has the status set,
+	// copy the conditions to the patch
+	if s, ok := cm.Object["status"]; ok {
+		fs, ok := s.(map[string]any)
+		if !ok {
+			return errors.Wrap(errors.New(errUnsupportedSrcObject), errMergeClaimStatus)
+		}
+		cmPatch.Object["status"] = keep(fs, "conditions")
+	} else {
+		cmPatch.Object["status"] = map[string]any{}
```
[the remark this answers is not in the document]
pedjak: What would be the benefit? `Object["status"]` or `Object["spec"]` are used across configurator.

At this particular line, we need to create an empty status placeholder, so that we can perform merge from the composite later on. Given that we are producing here a patch, `cmPatch.Object` does not contains `status` at all.  Hence, `fieldpath.Pave(cmPatch.Object).GetValueInto("status", map[string]any())` is going to produce an error.

internal/controller/apiextensions/claim/configurator.go
```diff
@@ -183,82 +179,94 @@ func filter(in map[string]any, keys ...string) map[string]any {
...
 }
 
-// Configure the supplied claims with fields from the composite.
+// configure the supplied claim patch with fields from the composite.
 // This includes late-initializing spec values and updating status fields in claim.
-func (c *APIClaimConfigurator) Configure(ctx context.Context, cm resource.CompositeClaim, cp resource.Composite) error {
-	ucm, ok := cm.(*claim.Unstructured)
-	if !ok {
-		return nil
+func configureClaim(_ context.Context, cm *claim.Unstructured, cmPatch *claim.Unstructured, cp *composite.Unstructured, cpPatch *composite.Unstructured) error { //nolint:gocyclo // Only slightly over (10)
```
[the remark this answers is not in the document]
pedjak: Sure, we can rename `cm` to `cmObserved`, but `cmDesired` is not the right name, because we only produce here a patch, not the complete desired resource.

internal/controller/apiextensions/claim/configurator.go
```diff
@@ -183,82 +179,94 @@ func filter(in map[string]any, keys ...string) map[string]any {
...
+		}
+		cmPatch.Object["status"] = keep(fs, "conditions")
+	} else {
+		cmPatch.Object["status"] = map[string]any{}
 	}
 
-	if err := merge(ucm.Object["status"], ucp.Object["status"],
+	// merge from the composite status everything
+	// except conditions and connectionDetails
+	if err := merge(cmPatch.Object["status"], cp.Object["status"],
```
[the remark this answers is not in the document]
pedjak: Yes, the issue is fixed. I used [the example](https://github.com/turkenh/xp-ssa-experiments/tree/fa71035a49f655802c569481bc7b6038bb89e475/issue-3542) you crafted for the verification, many thanks for that.

For creating the claim patch, we use `status.conditions` from the observed claim and merge all but `status.conditions` from the observed composite. In that way, when a field does not exists in the composite status, it will be removed from claim status as well.

internal/controller/apiextensions/claim/configurator.go
pedjak: The semantic is not changed - it is going to keep mirroring the revision reference. 

`cp.GetCompositionRevisionReference() != nil` was needed to introduce here, in order to prevent creation of invalid claim patches. Namely, 2nd claim reconciliation can happen before the composite got reconciled and the revision reference can be still `nil`. Without the new condition, claim patch would be invalid. If a field is nil, then it should not be a part of patch at all.

internal/controller/apiextensions/claim/reconciler.go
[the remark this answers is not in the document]
turkenh: > Having the same owner name shared by all claim controllers could lead to implicit shared ownership between controllers and degrades the user experience

I don't think this is practically possible. Each controller [works](https://github.com/crossplane/crossplane/blob/6756027ce951286d0c07ea38b5c1db141e89136f/internal/controller/apiextensions/claim/reconciler.go#L331) with its own type and there is no way for one claim controller to operate on another gvk. So, my preference would be to keep it simple.

> an user a clear information about the field ownership, e.g. when user performs something like:

I think it would still be clear enough if/when we just say this is owned by claim controller. No one would be confused about which claim's controller is operating on that claim, obviously, the claim controller for that GVK.

internal/controller/apiextensions/claim/reconciler.go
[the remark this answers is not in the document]
turkenh: > IMHO, Crossplane behavior should be the same both on existing and the new claims. How is user supposed to understand that label removals are propagated only on claims created by Crossplane version using server-side apply?

Agreed. 
As we discussed offline, I am okay with one-time migrations (and this code eventually being dropped after the previous release dropped from the supported releases). I am more concerned about this being part of regular operation, i.e. for all new composites, we always do a migration because it was created with a Create.

internal/controller/apiextensions/claim/configurator.go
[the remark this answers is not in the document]
turkenh: I am not sure I understand the difference between a patch vs complete desired resource considering we use `composite.Unstructured` as type 🤔

internal/controller/apiextensions/claim/configurator.go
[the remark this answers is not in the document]
turkenh: I see. 
Couldn't we avoid a merge here? AFAIU, what we need is something like:

```
	cmPatch.Object["status"] = cp.Object["status"]
	cmPatch.Object["status"].(map[string]any)["conditions"] = cp.Object["status"].(map[string]any)["conditions"]
```

\* code is just to give an idea, e.g. leveraging the fieldpath library could help to have a more readable and reliable code.

internal/controller/apiextensions/claim/configurator.go
turkenh: ah ok, it seems like I read it as `if cp.GetCompositionRevisionReference() == nil` for some reason 🤦

internal/controller/apiextensions/claim/reconciler.go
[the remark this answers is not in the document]
pedjak: > I think it would still be clear enough if/when we just say this is owned by claim controller. No one would be confused about which claim's controller is operating on that claim, obviously, the claim controller for that GVK.

Claim controller creates composites as well and owns a good part of it. As we discussed offline, it might be useful for users when debugging to understand which claim controller owns certain fields.

internal/controller/apiextensions/claim/reconciler.go
```diff
@@ -323,26 +289,26 @@ func WithPollInterval(after time.Duration) ReconcilerOption {
...
+		log:        logging.NewNopLogger(),
+		record:     event.NewNopRecorder(),
+		fieldOwner: client.FieldOwner(fmt.Sprintf("crossplane-claim-controller/%s", strings.ToLower(schema.GroupVersionKind(of).GroupKind().String()))),
 	}
 
 	for _, ro := range o {
 		ro(r)
 	}
 
+	r.patchOptions = []client.PatchOption{client.ForceOwnership, r.fieldOwner}
```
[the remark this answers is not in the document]
pedjak: there are 3 patch calls during the reconciliation, hence I am not sure if repeating unwrapping patch options next `client.Patch` is good.

internal/controller/apiextensions/claim/reconciler.go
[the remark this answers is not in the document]
pedjak: As promised, I had a look at `client-go` [UpgradeManagedFields()](https://github.com/kubernetes/client-go/blob/master/util/csaupgrade/upgrade.go#L81) function, but unfortunately we cannot use it:

* it does not give a feedback if patching was performed, so cannot know if we need to send update to API server
* it does not support skipping subresources (important for patching composites)

Other than that, the patching logic in our function is the same as in `UpgradeManagedFields()`.

internal/controller/apiextensions/claim/configurator.go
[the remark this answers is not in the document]
pedjak: I spend some over this, but I could not come with a more elegant code. The current code does the following:

* keeps `status.conditions` from the observed claim
* merges from the composite everthing except `status.conditions` and `status.connectionDetails`

Therefore, the code looks like this:

```go
// merge from the composite status everything
// except conditions and connectionDetails
if err := merge(cmPatch.Object["status"], cpObserved.Object["status"],
   // Status fields from composite overwrite non-empty fields in claim  
  withMergeOptions(mergo.WithOverride),
  withSrcFilter(xcrd.GetPropFields(xcrd.CompositeResourceStatusProps())...)); err != nil {
  return errors.Wrap(err, errMergeClaimStatus)
}
```

`merge` functon excepts a filter that gets populated with composite resource status prop names.

internal/controller/apiextensions/claim/reconciler.go
pedjak: Agree. Should we open a separate PR to fix that?

pedjak:
@turkenh https://github.com/crossplane/crossplane/pull/5062 got merged and I have rebased this one on the top of the latest master, answered your comments as well. PTAL.

internal/controller/apiextensions/claim/reconciler.go
turkenh: Given it is a log line I would just do it here, but a separate PR sound good too.

internal/controller/apiextensions/claim/configurator.go
turkenh: I agree with this.

internal/controller/apiextensions/claim/reconciler.go
```diff
@@ -561,7 +573,48 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+			obj.SetManagedFields(mfs)
+			fixed = true
+			continue
+		}
+
+		// Even if the server-side apply is not used, api server derives manager name
+		// from the submitted user agent (see net/http/request.go).
+		// After crossplane update, we need to replace the ownership so that
+		// field removals can be propagated properly.
+		if mfs[i].Manager == "Go-http-client" && mfs[i].Operation == "Update" {
```
turkenh, line 609: Are we sure that this will migrate old versions? 

Just deployed a claim/composite with `v1.14.2` and below is what I am seeing:

```
      manager: crossplane
      operation: Update
```

<details>

<summary>composite</summary>

```yaml
managedFields:
    - apiVersion: nop.example.org/v1alpha1
      fieldsType: FieldsV1
      fieldsV1:
        f:metadata:
          f:finalizers:
            .: {}
            v:"composite.apiextensions.crossplane.io": {}
          f:generateName: {}
          f:labels:
            .: {}
            f:crossplane.io/claim-name: {}
            f:crossplane.io/claim-namespace: {}
            f:crossplane.io/composite: {}
        f:spec:
          .: {}
          f:claimRef:
            .: {}
            f:apiVersion: {}
            f:kind: {}
            f:name: {}
            f:namespace: {}
          f:compositionRef:
            .: {}
            f:name: {}
          f:compositionRevisionRef:
            .: {}
            f:name: {}
          f:compositionUpdatePolicy: {}
          f:coolField: {}
          f:resourceRefs: {}
      manager: crossplane
      operation: Update
      time: "2023-12-05T06:54:58Z"
    - apiVersion: nop.example.org/v1alpha1
      fieldsType: FieldsV1
      fieldsV1:
        f:status:
          .: {}
          f:conditions:
            .: {}
            k:{"type":"Ready"}:
              .: {}
              f:lastTransitionTime: {}
              f:reason: {}
              f:status: {}
              f:type: {}
            k:{"type":"Synced"}:
              .: {}
              f:lastTransitionTime: {}
              f:reason: {}
              f:status: {}
              f:type: {}
          f:nginx:
            .: {}
            f:conditions: {}
            f:observedGeneration: {}
            f:replicas: {}
            f:unavailableReplicas: {}
            f:updatedReplicas: {}
      manager: crossplane
      operation: Update
      subresource: status
      time: "2023-12-05T06:54:59Z"
```

</details>

<details>

<summary>claim</summary>

```yaml
managedFields:
    - apiVersion: nop.example.org/v1alpha1
      fieldsType: FieldsV1
      fieldsV1:
        f:metadata:
          f:finalizers:
            .: {}
            v:"finalizer.apiextensions.crossplane.io": {}
        f:spec:
          f:compositionRef:
            .: {}
            f:name: {}
          f:compositionRevisionRef:
            .: {}
            f:name: {}
          f:compositionUpdatePolicy: {}
          f:resourceRef:
            .: {}
            f:apiVersion: {}
            f:kind: {}
            f:name: {}
      manager: crossplane
      operation: Update
      time: "2023-12-05T06:54:58Z"
    - apiVersion: nop.example.org/v1alpha1
      fieldsType: FieldsV1
      fieldsV1:
        f:metadata:
          f:annotations:
            .: {}
            f:kubectl.kubernetes.io/last-applied-configuration: {}
        f:spec:
          .: {}
          f:compositeDeletePolicy: {}
          f:coolField: {}
      manager: kubectl-client-side-apply
      operation: Update
      time: "2023-12-05T06:54:58Z"
    - apiVersion: nop.example.org/v1alpha1
      fieldsType: FieldsV1
      fieldsV1:
        f:status:
          .: {}
          f:conditions:
            .: {}
            k:{"type":"Ready"}:
              .: {}
              f:lastTransitionTime: {}
              f:reason: {}
              f:status: {}
              f:type: {}
            k:{"type":"Synced"}:
              .: {}
              f:lastTransitionTime: {}
              f:reason: {}
              f:status: {}
              f:type: {}
          f:nginx:
            .: {}
            f:availableReplicas: {}
            f:conditions: {}
            f:observedGeneration: {}
            f:readyReplicas: {}
            f:replicas: {}
            f:unavailableReplicas: {}
            f:updatedReplicas: {}
      manager: crossplane
      operation: Update
      subresource: status
      time: "2023-12-05T06:55:59Z"
```

</details>

turkenh (commented):
I tried an upgrade test and noticed that the composite keep getting updates, see below:

https://github.com/crossplane/crossplane/assets/9900707/84c58894-1bc2-4e1a-8860-7ecf582f2a82

Reproduction steps:
1. Install Crossplane v1.14.2
2. Follow the steps [here](https://github.com/turkenh/xp-ssa-experiments/blob/fa71035a49f655802c569481bc7b6038bb89e475/issue-3542/README.md).
3. Upgrade Crossplane to a local build from this branch.
4. Run `kubectl get xnopresources.nop.example.org -w` 

Could you investigate that?

---

Well, actually this is true for just regular install as well without upgrading Crossplane:

Reproduction steps:
1. Install Crossplane with a local build from this branch.
2. Follow the steps [here](https://github.com/turkenh/xp-ssa-experiments/blob/fa71035a49f655802c569481bc7b6038bb89e475/issue-3542/README.md).
3. Run `kubectl get xnopresources.nop.example.org -w` 

https://github.com/crossplane/crossplane/assets/9900707/e984fc09-4810-4c82-b91c-d214b1fc515f

It become True/True only after around 5mins:

Screenshot 2023-12-05 at 10 29 40

pedjak:
@turkenh I have addressed your comments. PTAL.

turkenh:
> @turkenh I have addressed your comments. PTAL.

Did you have a chance to do some testing and see what was going on [here](https://github.com/crossplane/crossplane/pull/4896#pullrequestreview-1764012662)?

internal/controller/apiextensions/claim/configurator.go
pedjak: done, please reopen the conversation if does not match the expectations.

internal/controller/apiextensions/claim/reconciler.go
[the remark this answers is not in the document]
pedjak: done, reverted to the approach you are proposing.

internal/controller/apiextensions/claim/configurator.go
```diff
@@ -182,82 +189,94 @@ func filter(in map[string]any, keys ...string) map[string]any {
...
+	// copy the conditions to the patch
+	if s, ok := cmObserved.Object["status"]; ok {
+		fs, ok := s.(map[string]any)
+		if !ok {
+			return errors.Wrap(errors.New(errUnsupportedSrcObject), errMergeClaimStatus)
+		}
+		cmPatch.Object["status"] = keep(fs, "conditions")
+	} else {
+		cmPatch.Object["status"] = map[string]any{}
 	}
```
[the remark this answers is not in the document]
pedjak: done. Given that `cmObserved.Object["status"]` does not exists only at the initial creation of the claim, `cmPatch.Object["status"] ` got assinged two times the most of time. Not sure how much pressure make creation of empty maps to the garbage collector.

internal/controller/apiextensions/claim/reconciler.go
pedjak: Good catch! I have fixed the issue and updated the e2e tests for check explicitly for that.

pedjak:
> Did you have a chance to do some testing and see what was going on https://github.com/crossplane/crossplane/pull/4896#pullrequestreview-1764012662

Yes, fixed, but please double check.

internal/controller/apiextensions/claim/reconciler.go
```diff
@@ -457,80 +437,112 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+	// Restore the status calculated in this round
+	// and use cmPatch until the end of the reconciliation
+	// for setting claim conditions
+	cm.Object["status"] = desiredClaimStatus
+
+	// If composite did not exist at the beginning of the loop, we want to create it
+	// but there is a probability (very unlikely) that the generated name
+	// gets taken before we submit the request. In that case, we are going to
+	// update a composite, hijacking it from another claim.
+	// TODO: investigate if we need to prevent it.
```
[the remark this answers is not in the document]
turkenh: Thanks for https://github.com/crossplane/crossplane/issues/5104

turkenh (approved):
Great work @pedjak, see how many (long running issues) fixed by this single PR 💪 
Thank for bearing with me during reviews 🙌

PR #4870, opened 2023-10-22
Support pushing multi-platform packages

negz:
### Description of your changes

Fixes https://github.com/crossplane/crossplane/issues/4869

Tested manually:

```shell
$ docker buildx build --platform linux/amd64 --output=type=docker,dest=runtime-amd64.tar --quiet .
sha256:a6d3ce87e3d429fe77b5ff4faf623b4a7b8c7b57f9b0e79acb11ae7c86c24544
$ docker buildx build --platform linux/arm64 --output=type=docker,dest=runtime-arm64.tar --quiet .
sha256:a38acf1e6c1da2ee66b058de01f704784e8c6f03ddf11219c042a187f2259b45
$ crossplane xpkg build -f package --embed-runtime-image-tarball=runtime-amd64.tar -o function-amd64.xpkg
$ crossplane xpkg build -f package --embed-runtime-image-tarball=runtime-arm64.tar -o function-arm64.xpkg
$ crossplane xpkg --verbose push -f function-amd64.xpkg,function-arm64.xpkg crossplane-contrib/function-patch-and-transform:v0.1.4
2023-10-22T01:08:09-07:00       DEBUG   Pushed package  {"path": "/home/negz/control/crossplane-contrib/function-patch-and-transform/function-arm64.xpkg", "ref": "xpkg.upbound.io/crossplane-contrib/function-patch-and-transform@sha256:1132e7dc2f2acf46066dd0e7309634fa3e08b8b04848e05fc9447850a827fa4f"}
2023-10-22T01:08:09-07:00       DEBUG   Pushed package  {"path": "/home/negz/control/crossplane-contrib/function-patch-and-transform/function-amd64.xpkg", "ref": "xpkg.upbound.io/crossplane-contrib/function-patch-and-transform@sha256:f1498cfdb0365f107e933e6602c399065607dd6a4faeeca5b8a710e943606f9b"}
2023-10-22T01:08:09-07:00       DEBUG   Wrote OCI index {"ref": "crossplane-contrib/function-patch-and-transform:v0.1.4", "manifests": 2}
```

I have: 

- [x] Read and followed Crossplane's [contribution process].
- [x] Run `make reviewable` to ensure this PR is ready for review.
- [ ] ~Added or updated unit tests.~
- [ ] ~Added or updated e2e tests.~
- [ ] ~Linked a PR or a [docs tracking issue] to [document this change].~
- [ ] ~Added `backport release-x.y` labels to auto-backport this PR.~

[...]

PR #4901, opened 2023-10-26
Propagate package annotations from config file labels to layer annotations

negz:

This is necessary for both the Crossplane package manager and xpkg.upbound.io to be able to find the package layer when a package image also contains a runtime. Without this change:

* The package manager will needlessly pull and cache the whole image, including its runtime.
* It won't be possible for packages pushed to xpkg.upbound.io to be listed at https://marketplace.upbound.io because they'll be too big

It's the layer annotations that the package manager actually cares about, not the labels in the config file. I don't 100% follow, but it appears that we use the config file as temporary storage for these annotations because they're not persisted to the tarball format. Or possibly the tarball format just loses them in certain situations?

This is a follow on to https://github.com/crossplane/crossplane/pull/4870 where per https://github.com/crossplane/crossplane/pull/4870#pullrequestreview-1691409130 @phisco and I both mistook this for some Upbound special sauce that wasn't relevant to the Crossplane CLI. 😄 

I have: 

### DIFF
--- a/internal/xpkg/layers.go
+++ b/internal/xpkg/layers.go
@@ -24,9 +24,17 @@ import (
 	"os"
 
 	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/empty"
+	"github.com/google/go-containerregistry/pkg/v1/mutate"
 	"github.com/google/go-containerregistry/pkg/v1/tarball"
 
 	"github.com/crossplane/crossplane-runtime/pkg/errors"
+)
+
+// Error strings.
+const (
+	errLayer  = "cannot get image layers"
+	errDigest = "cannot get image digest"
 )
 
 // Layer creates a v1.Layer that represents the layer contents for the xpkg and
@@ -61,7 +69,10 @@ func Layer(r io.Reader, fileName, annotation string, fileSize int64, mode os.Fil
 		return nil, errors.Wrap(err, errDigestInvalid)
 	}
 
-	// add annotation label to config if a non-empty label is specified
+	// Add annotation label to config if a non-empty label is specified. This is
+	// an intermediary step. AnnotateLayers must be called on an image for it to
+	// have valid layer annotations. It propagates these labels to annotations
+	// on the layers.
 	if annotation != "" {
 		cfg.Labels[Label(d.String())] = annotation
 	}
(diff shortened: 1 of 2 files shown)


internal/xpkg/layers.go
```diff
@@ -87,3 +101,61 @@ func writeLayer(tw *tar.Writer, hdr *tar.Header, buf io.Reader) error {
...
 	return fmt.Sprintf("%s:%s", AnnotationKey, annotation)
 }
+
+// NOTE(negz): AnnotateLayers originated in upbound/up. I was confused why we
+// store layer annotations as labels in the OCI config file when we build a
+// package, then propagate them to OCI layer annotations when we push one. I
+// believe this is because an xpkg file is really an OCI image tarball, and the
+// tarball format doesn't support layer annotations (or may just lose them in
+// some circumstances?), so we're using the config file to store them.
+// See https://github.com/upbound/up/pull/177#discussion_r866776584.
```
negz, line 111: CC @hasheddan @tnthornton to check my math here.

negz:
Tested manually with a local build of `crossplane`:

```shell
$ docker buildx build --platform linux/arm64 --output=type=docker,dest=runtime-arm64.tar --quiet .
sha256:b8a2d930a4068ba385b83c1483751060d28236cb4e1fdea5c9ac8d30fa8f9647
$ docker buildx build --platform linux/amd64 --output=type=docker,dest=runtime-amd64.tar --quiet .
sha256:ff2c0d704bbeac2100e115c2fa83ee00358a04791c9aec7d176f85aaee75298c
$ crossplane xpkg build -f package --embed-runtime-image-tarball=runtime-arm64.tar -o function-arm64.xpkg
$ crossplane xpkg build -f package --embed-runtime-image-tarball=runtime-amd64.tar -o function-amd64.xpkg
$ crossplane --verbose xpkg push -f function-arm64.xpkg,function-amd64.xpkg crossplane-contrib/function-patch-and-transform:v0.1.4
2023-10-25T18:54:21-07:00       DEBUG   Pushed package  {"path": "/home/negz/control/crossplane-contrib/function-patch-and-transform/function-amd64.xpkg", "ref": "xpkg.upbound.io/crossplane-contrib/function-patch-and-transform@sha256:2a470750e6443539c8d598e3131b4b3f79ea10a00bf6e778c3b237912a9a51ba"}
2023-10-25T18:54:21-07:00       DEBUG   Pushed package  {"path": "/home/negz/control/crossplane-contrib/function-patch-and-transform/function-arm64.xpkg", "ref": "xpkg.upbound.io/crossplane-contrib/function-patch-and-transform@sha256:b608e624c5299d5cd8f4d68bb6d61170aa0fb114b7bd5ae2df229d47ea3734bc"}
2023-10-25T18:54:22-07:00       DEBUG   Wrote OCI index {"ref": "crossplane-contrib/function-patch-and-transform:v0.1.4", "manifests": 2}
```

I see that the resulting image is accepted by https://marketplace.upbound.io. Previously it was being rejected for being too big - a sign that the Marketplace wasn't able to find the annotated package layer. The Marketplace uses the same logic as the Crossplane Package manager, so without this change the package manager was most likely inefficiently pulling and caching the whole image rather than a single annotated package layer.

Screenshot 2023-10-25 at 6 54 52 PM

internal/xpkg/layers.go
negz: @phisco Looks like we needed this code after all. 😄

phisco (commented):
😨 totally misunderstood that bit... then why aren't we doing it at build time too? Otherwise it means that people building with crossplane/up and pushing with docker or any other tool won't be able to push proper images, isn't it?
Does this mean one could actually build a valid image from just a Dockerfile using the LABEL command??

phisco (approved):
LGTM, I still have some doubts about the whole thing, but this shouldn't block this functionality we care about

phisco:
> 😨 totally misunderstood that bit... then why aren't we doing it at build time too? Otherwise it means that people building with crossplane/up and pushing with docker or any other tool won't be able to push proper images, isn't it? 
Yes, I think we should do this and double check that on push.

> Does this mean one could actually build a valid image from just a Dockerfile using the LABEL command??
No, the LABEL directive is just setting a label in the root config, exactly as we are doing on `build`. Annotating single layers is not supported by Docker afaict.

jbw976:
Does this need to be backported to the [`release-1.14` branch](https://github.com/crossplane/crossplane/tree/release-1.14)?

phisco:
@jbw976 the release branch was created after merging this

PR #4904, opened 2023-10-26
controller/pkg: handle package.crossplane.io/breaking-change-version

sttts:

The v1.0 update of the official providers will be a breaking change in some APIs. We don't want that `> 0.40.0`-like dependencies in configurations pull in the v1 version. We want that a breaking change version is explicitly required, e.g. through `>= 1`.

This PR uses an (OCI image descriptor) annotation to on a provider (or any other) package specifying the last breaking change version.

To implement this logic, we have a small fork of the semver library that adds a `CheckWithBreakingVersion` method. It verifies that the lower bound specified in a dependency constraint is at least the breaking change version. If not, that version of the package is not considered a valid version for the constraint.

I have: 

- [ ] ~~Added or updated e2e tests.~~
- [ ] ~~Linked a PR or a [docs tracking issue] to [document this change].~~
- [ ] ~~Added `backport release-x.y` labels to auto-backport this PR.~~

### DIFF
--- a/internal/controller/pkg/resolver/reconciler.go
+++ b/internal/controller/pkg/resolver/reconciler.go
@@ -24,7 +24,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/Masterminds/semver"
 	"github.com/google/go-containerregistry/pkg/name"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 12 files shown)


negz:
Let's set to draft while we figure out when/whether it's needed.

negz:
I'm pretty sure we decided not to go this path, so closing. Let me know if I'm mistaken.

PR #4918, opened 2023-10-27
Skip reconciliation on no-op claim and composite changes

### LINKED ISSUE #4703: Avoid no-op reconcilations

Resources are reconciliated on any change, but very it does not generate any change in the system, thus it could be easily skipped. The affected controller could be less busy, using less system resources.

Fine tune when a given reconcile loop is triggered:
* Claim reconciliation should be triggered only when (1) claim's annotation, labels an/or spec is changed, or (2) when composite status is modified
* Composite reconciliation should be triggered when (1) labels, annotations or spec is changed, (2) composition changes and/or (3) managed resource is updated

pedjak:

Claim reconciliation needs to happen only on:
* a non-status claim change and/or
* a non-metadata composite change for the given claim

Composite reconciliation needs to happen only on:
* a non-status composite change

Changes:
* Added `IgnoreStatusChanges` and `IgnoreMetadataChanges` predicates for proper event filtering

Fixes #4703 

I have: 

### DIFF
--- a/internal/controller/apiextensions/definition/reconciler.go
+++ b/internal/controller/apiextensions/definition/reconciler.go
@@ -59,6 +59,7 @@ import (
 	apiextensionscontroller "github.com/crossplane/crossplane/internal/controller/apiextensions/controller"
 	"github.com/crossplane/crossplane/internal/features"
 	"github.com/crossplane/crossplane/internal/xcrd"
+	"github.com/crossplane/crossplane/pkg/controller/predicate"
 )
 
 const (
@@ -462,7 +463,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 	name := composite.ControllerName(d.GetName())
 	var ca cache.Cache
 	watches := []controller.Watch{
-		controller.For(u, &handler.EnqueueRequestForObject{}),
+		controller.For(u, &handler.EnqueueRequestForObject{}, predicate.IgnoreStatusChanges()),
 		// enqueue composites whenever a matching CompositionRevision is created
 		controller.TriggeredBy(source.Kind(r.mgr.GetCache(), &v1.CompositionRevision{}), handler.Funcs{
 			CreateFunc: composite.EnqueueForCompositionRevisionFunc(ck, r.mgr.GetCache().List, r.log),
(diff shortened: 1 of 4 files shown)


bobh66 (commented):
This looks OK to me but I'd prefer to have an approval from someone who has better golang skills than I do.  Added @turkenh and @phisco to get some better-trained eyes on it.

pkg/controller/interceptor/interceptors.go
```diff
@@ -0,0 +1,47 @@
...
+			d := time.Since(start).String()
+			if err == nil {
+				l.Info("Reconciliation done", "result", res, "duration", d)
+			} else {
+				l.Info("Reconciliation failed", "error", err, "duration", d)
+			}
+		}()
+		return r.Reconcile(ctx, req)
+	}
+}
```
phisco, line 47: We currently don't log much at `Info` level, so I think it would make more sense to at least lower the level here to `Debug` or if you feel this needs to be at `Info` level, please start a conversation to change the project's approach to logging (if there is no dedicated issue already) and move this to a dedicated PR, so that it doesn't hold up the rest of the PR 🙏 WDYT @pedjak?

sttts: and please split into two commits. This is not one change.

pedjak: > We currently don't log much at Info level, 

True,  we do it in a handful of places: https://github.com/search?q=repo%3Acrossplane%2Fcrossplane%20%22.Info(%22&type=code

Given that this PR is about lowering the number of reconcile rounds, I thought it would be good also to make that trackable even if debug logs are not turned on, hence this change here.

Having single info line for each reconciliation improves UX, i.e. user get a notion what is going on, i.e. what gets reconciled. Today this mostly happens silently and it is super hard to diagnose whether it happens and why.

phisco: We do expose prometheus metrics with similar, although less granular, [metrics](https://book.kubebuilder.io/reference/metrics-reference). But, I do agree we could benefit from being more verbose, I just feel that's a major policy change that should be discussed more broadly.

pedjak: `LoggingInterceptor` moved into the separate PR: #5143.

negz:
I see that @sttts had some concerns about doing this per https://github.com/crossplane/crossplane/issues/4703#issuecomment-1746809710. @sttts can you elaborate? My guess would be you're worried about false negatives - i.e. we skip some reconciles that were actually useful in some way we didn't expect?

If we did want to build confidence, we could use a feature flag.

sttts:
Yes, skipping reconciles that matter. It complicates the state machine.

Metrics are nice, but how often does anybody look at metrics? Especially us devs. We might eventually notice that Crossplane is heavy but have no clue why. Metrics don‘t tell a helpful story.

PR #4923, opened 2023-10-27
Push Protobufs to Buf Schema Registry

negz:

This registry generates us documentation. I believe it also allows us to avoid copying around protobuf files, which we currently need to do for each SDK (e.g. https://github.com/crossplane/function-sdk-go/tree/main/proto). The `buf` tool is OSS but the Schema Registry is a commercial product of https://buf.build. It does appear to be free for OSS use.

Example docs: https://buf.build/crossplane/crossplane/docs/main:apiextensions.fn.proto.v1beta1

I'd like to link to these docs from https://docs.crossplane.io.

I have: 


.github/workflows/ci.yml
```diff
@@ -451,3 +451,35 @@ jobs:
...
+        with:
+          input: apis
+          against: "https://github.com/${GITHUB_REPOSITORY}.git#branch=master,subdir=apis"
+        
+      - name: Push Protocol Buffers to Buf Schema Registry
+        if: github.ref == 'refs/heads/master'
+        uses: bufbuild/buf-push-action@v1
+        with:
+          input: apis
+          buf_token: ${{ secrets.BUF_TOKEN }}
```
negz, line 485: Unfortunately the Buf Schema Registry doesn't seem to support robots, so I'll need to create a token associated with my account and use that. We'll also need at least one other maintainer to create an account for redundancy.

apis/buf.lock
```diff
@@ -0,0 +1,7 @@
+# Generated by buf. DO NOT EDIT.
```
jbw976, line 1: what's the expected workflow for this type of file being updated? is it something a developer will ever need to do with the `buf` tooling and include in a PR?

.github/workflows/ci.yml
```diff
@@ -451,3 +451,35 @@ jobs:
...
+
+      - name: Setup Buf
+        uses: bufbuild/buf-setup-action@v1
+      
+      - name: Lint Protocol Buffers
+        uses: bufbuild/buf-lint-action@v1
+        with:
+          input: apis
+
+      - name: Detect Breaking Changes in Protocol Buffers
```
jbw976, line 474: should there be a local part of the workflow here? e.g. `make reviewable` does this same check so devs can have confidence in their commit before opening a PR?

jbw976 (commented):
Directionally, this seems reasonable, and we could always change direction in the future if we wanted more direct integration of some fashion with docs.crossplane.io.

A few question about the expected workflow here, but overall this is reasonable.

.github/workflows/ci.yml
negz: It would be nice to have, but I feel a sense of urgency to have something in place and I don't want to take on teaching the Makefile and/or build submodule about `buf` right now. I've raised https://github.com/crossplane/crossplane/issues/4924 to track this.

apis/buf.lock
negz: Exceedingly rarely. The command is `buf mod update`, which I've mentioned in https://github.com/crossplane/crossplane/issues/4924 for now.

jbw976 (approved):

negz:
@phisco @turkenh @bobh66 FYI I'm wanting to move fast on this to get Functions documented for the v1.14 release. I'm going to use @jbw976's steering committee approval as a stand-in for maintainer approval. Let me know if you have any concerns with this PR.

PR #4925, opened 2023-10-28
Handle externally managed Provider service accounts properly

### LINKED ISSUE #4552: Providers take exclusive Ownership of Existing K8s Serviceaccounts
### Context 

There are times Crossplane users need to customize the Kubernetes service accounts that providers run under.  Some security policies require all pods to run with annotated service accounts. Another example is [AWS IAM policies](https://docs.aws.amazon.com/eks/latest/userguide/associate-service-account-role.html) which can be matched to a combination Kubernetes Service Account and namespace.

In the case of labels and annotations, the ControllerConfig will add its labels and annotations to the ServiceAccount.

### What happened?

- When existing k8s service accounts are used in a `ControllerConfig` `serviceAccountName` setting, the Provider will take exclusive ownership of the resource. 
- If a service account is shared among multiple providers (as in the case of family providers sharing a single service account), only one ownerReference will be written.
- The single ownerReference causes the `ClusterRoleBindings` that are generated to only support one of the providers.
- When deleting a Provider, the serviceAccount will be deleted

### How can we reproduce it?

Create service accounts:

```yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: provider-aws-ec2
  namespace: upbound-system
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: provider-aws-iam
  namespace: upbound-system
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: provider-aws
  namespace: upbound-system
  annotations:
    a: b
    c: d
  labels:
    one: two
    three: four
```

Next create ControllerConfigs, note the `provider-aws` will be shared among family providers:

```yaml
kind: ControllerConfig
metadata:
  name: controllerconfig-aws-iam
spec:
  serviceAccountName: provider-aws-iam
  args:
    - --enable-management-policies 
---
kind: ControllerConfig
metadata:
  name: controllerconfig-aws-ec2
spec:
  serviceAccountName: provider-aws-ec2
  args:
    - --enable-management-policies 
---
kind: ControllerConfig
metadata:
  name: controllerconfig-aws
spec:
  serviceAccountName: provider-aws
  args:
    - --enable-management-policies
 ```

Finally install the providers. Note that the s3, eks and rds family providers are set to use the same `serviceAccountName`:

```yaml
kind: Provider
metadata:
  name: provider-aws-iam
spec:
  package: xpkg.upbound.io/upbound/provider-aws-iam:v0.39.0
  controllerConfigRef:
    name: controllerconfig-aws-iam
---
kind: Provider
metadata:
  name: provider-aws-ec2
spec:
  package: xpkg.upbound.io/upbound/provider-aws-ec2:v0.39.0
  controllerConfigRef:
    name: controllerconfig-aws-ec2
---
kind: Provider
metadata:
  name: provider-aws-s3
spec:
  package: xpkg.upbound.io/upbound/provider-aws-s3:v0.39.0
  controllerConfigRef:
    name: controllerconfig-aws
---
kind: Provider
metadata:
  name: provider-aws-eks
spec:
  package: xpkg.upbound.io/upbound/provider-aws-eks:v0.39.0
  controllerConfigRef:
    name: controllerconfig-aws
---
kind: Provider
metadata:
  name: provider-aws-rds
spec:
  package: xpkg.upbound.io/upbound/provider-aws-rds:v0.39.0
  controllerConfigRef:
```

Once the providers are healthy, we can look at the Service accounts:

 `provider-aws-iam` has a single owner  and `"controller": true,` even though the account was created outside of the Crossplane package manager.

```console
kubectl get sa -n upbound-system provider-aws-iam -o jsonpath='{.metadata.ownerReferences}' | jq
[
  {
    "apiVersion": "pkg.crossplane.io/v1",
    "blockOwnerDeletion": true,
    "controller": true,
    "kind": "ProviderRevision",
    "name": "provider-aws-iam-e4667f1b5f01",
    "uid": "3e04dfd7-d9f5-4e99-9b28-db539956ca6f"
  }
]
```

For the shared `provider-aws` only one of the three controllers is listed as an `ownerReference`. T

```console
kubectl get sa -n upbound-system provider-aws -o jsonpath='{.metadata.ownerReferences}' | jq 
[
  {
    "apiVersion": "pkg.crossplane.io/v1",
    "blockOwnerDeletion": true,
    "controller": true,
    "kind": "ProviderRevision",
    "name": "provider-aws-s3-e6123c010ecc",
    "uid": "61f520b6-a9df-4b1a-9a45-8191c1a3f7b0"
  }
]
```

This lack of additional owners causes the `ClusterRoleBinding` for the eks family provider to not have a subjet k8s service account:

```
kubectl get clusterrolebinding crossplane:provider:provider-aws-eks-b731c789c55c:system    -o yaml 
kind: ClusterRoleBinding
metadata:
  creationTimestamp: "2023-08-31T20:26:27Z"
  name: crossplane:provider:provider-aws-eks-b731c789c55c:system
  ownerReferences:
  - apiVersion: pkg.crossplane.io/v1
    blockOwnerDeletion: true
    controller: true
    kind: ProviderRevision
    name: provider-aws-eks-b731c789c55c
    uid: 1e6ec2bc-b24e-46ce-b85f-1cd0c411e6f2
  resourceVersion: "2697"
  uid: 85060c1c-4bdf-45e1-bfcc-3797a32feb3f
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: crossplane:provider:provider-aws-eks-b731c789c55c:system
```

If we delete the provider-aws-iam package, the linked serviceAccount will be deleted:

```console
kubectl delete provider.pkg provider-aws-iam
provider.pkg.crossplane.io "provider-aws-iam" deleted

kubectl get sa -n upbound-system provider-aws-iam 
Error from server (NotFound): serviceaccounts "provider-aws-iam" not found
```

Crossplane version: UXP v1.13.2

 hardware configuration
* Kubernetes version: v1.27.1
* Kubernetes distribution: Kind

### LINKED ISSUE #4908: DeploymentRuntimeConfig `deploymentTemplate.spec.template.spec.serviceAccountName` is ignored
### What happened?

When creating a DeploymentRuntimeConfig, the field `deploymentTemplate.spec.template.spec.serviceAccountName` is ignored by the controller in favour of the `serviceAccountTemplate` field.

If the `serviceAccountTemplate` is not set, or empty, the expected behaviour should be that `deploymentTemplate.spec.template.spec.serviceAccountName` is applied. However the behaviour appears to be that this field is ignored in all instances and a new, dynamic service account is created to be used by the runtime.

Where both `serviceAccountTemplate` and `deploymentTemplate.spec.template.spec.serviceAccountName` are given, my opinion would be that `deploymentTemplate.spec.template.spec.serviceAccountName` takes precedence as this is explicitly stating that "I want this account to be used with this deployment" however again in this instance `serviceAccountTemplate` takes precedence here.

If the name given to `serviceAccountTemplate` is the name of a service account that exists in the cluster, crossplane adopts this service account and applies it to the `Deployment`, creating the relevant `ClusterRoleBindings` for it.

Whilst I don't have a strong opinion on how this is resolved, the unused existence of `deploymentTemplate.spec.template.spec.serviceAccountName` leads to a level of confusion when the referenced service account is not adopted by the deployment and instead a new account is created.

Towards this I see at least 3 options

1. Always use `deploymentTemplate.spec.template.spec.serviceAccountName` if it is provided and fail the apply if the SA doesn't exist
2. Use `deploymentTemplate.spec.template.spec.serviceAccountName` irrespective of `serviceAccountTemplate` if it coheres to a valid service account in the `crossplane` install namespace. This implies ignoring `serviceAccountTemplate` if `deploymentTemplate...serviceAccountName` is set
3. Deprecate `deploymentTemplate.spec.template.spec.serviceAccountName` in favour of `serviceAccountTemplate` and make it clear through documentation that setting this field has no effect

Associated slack thread: https://crossplane.slack.com/archives/CEG3T90A1/p1698071278410109

### How can we reproduce it?

Apply the following resources:

```yaml
---
# Create the service account to be used for AWS providers
apiVersion: v1
kind: ServiceAccount
metadata:
  name: upbound-provider-aws
  namespace: crossplane

---
# DeploymentRuntimeConfig for AWS resources
kind: DeploymentRuntimeConfig
metadata:
  name: aws-family
spec:
  deploymentTemplate:
    spec:
      selector: {}
      template:
        spec:
          serviceAccountName: upbound-provider-aws
          containers:
            - name: package-runtime
              args:
                - --debug
                - --enable-management-policies
---
kind: Provider
metadata:
  name: upbound-provider-family-aws
  namespace: crossplane
spec:
  package: xpkg.upbound.io/upbound/provider-family-aws:v0.42.0
  runtimeConfigRef:
    name: aws-family
```

Check the SAs
```bash
$ k get sa -n crossplane
NAME                                       SECRETS   AGE
crossplane                                 0         5m2s
default                                    0         5m2s
rbac-manager                               0         5m2s
upbound-provider-aws                       0         4m43s
upbound-provider-family-aws-ab7dbb587523   0         4m35s
```

This last SA: `upbound-provider-family-aws-ab7dbb587523` should not exist.

Check the deployment:

```bash
$ k get deploy upbound-provider-family-aws-ab7dbb587523 -o yaml | yq '.spec.template.spec.serviceAccount*'
upbound-provider-family-aws-ab7dbb587523
upbound-provider-family-aws-ab7dbb587523
```

I would expect these to both read `upbound-provider-aws`

Cluster type: `kind v0.20.0 go1.21.1 linux/amd64`
Crossplane version: `crossplane:v1.15.0-rc.0.3.g35b34e85`

```nohighlight
$ helm install crossplane --namespace crossplane --create-namespace crossplane-master/crossplane --devel

NAME: crossplane
LAST DEPLOYED: Thu Oct 26 18:01:12 2023
NAMESPACE: crossplane
STATUS: deployed
REVISION: 1
TEST SUITE: None
NOTES:
Release: crossplane

Chart Name: crossplane
Chart Description: Crossplane is an open source Kubernetes add-on that enables platform teams to assemble infrastructure from multiple vendors, and expose higher level self-service APIs for application teams to consume.
Chart Version: 1.15.0-rc.0.3.g35b34e85
Chart Application Version: 1.15.0-rc.0.3.g35b34e85

Kube Version: v1.27.3
```

turkenh:

This PR:
- Changes the mechanism to find the Service Account used by a ProviderRevision from relying on owner references on the ServiceAccounts to checking the ServiceAccount used by the runtime Deployment. (Fixes #4552)
- Allows passing an externally managed Service Account via `DeploymentRuntimeConfig` and skipping the creation of the Service Account. (Fixes #4908)

After this PR, if the `deploymentTemplate.spec.template.spec.serviceAccountName` is set in the referenced `DeploymentRuntimeConfig`, Crossplane will:

- Just use it without trying to own it or creating a new one.
- Bind necessary RBAC permission to that Service Account.

I have: 

### DIFF
--- a/internal/controller/pkg/revision/runtime.go
+++ b/internal/controller/pkg/revision/runtime.go
@@ -207,12 +207,12 @@ func (b *RuntimeManifestBuilder) Deployment(serviceAccount string, overrides ...
 			Privileged:               &privileged,
 			RunAsNonRoot:             &runAsNonRoot,
 		}),
+		DeploymentWithOptionalServiceAccount(serviceAccount),
 
 		// Overrides that we are opinionated about.
 		DeploymentWithNamespace(b.namespace),
 		DeploymentWithOwnerReferences([]metav1.OwnerReference{meta.AsController(meta.TypedReferenceTo(b.revision, b.revision.GetObjectKind().GroupVersionKind()))}),
 		DeploymentWithSelectors(b.podSelectors()),
-		DeploymentWithServiceAccount(serviceAccount),
 		DeploymentWithImagePullSecrets(b.revision.GetPackagePullSecrets()),
 		DeploymentRuntimeWithAdditionalPorts([]corev1.ContainerPort{
 			{
--- a/internal/controller/pkg/revision/runtime_function.go
+++ b/internal/controller/pkg/revision/runtime_function.go
@@ -116,11 +116,17 @@ func (h *FunctionHooks) Post(ctx context.Context, pkg runtime.Object, pr v1.Pack
 	}
 
 	sa := build.ServiceAccount()
-	if err := h.client.Apply(ctx, sa); err != nil {
-		return errors.Wrap(err, errApplyFunctionSA)
-	}
-
 	d := build.Deployment(sa.Name, functionDeploymentOverrides(functionMeta, pr)...)
+	// Create/Apply the SA only if the deployment references it.
+	// This is to avoid creating a SA that is NOT used by the deployment when
+	// the SA is managed externally by the user and configured by setting
+	// `deploymentTemplate.spec.template.spec.serviceAccountName` in the
+	// DeploymentRuntimeConfig.
+	if sa.Name == d.Spec.Template.Spec.ServiceAccountName {
+		if err := h.client.Apply(ctx, sa); err != nil {
+			return errors.Wrap(err, errApplyFunctionSA)
+		}
+	}
 	if err := h.client.Apply(ctx, d); err != nil {
 		return errors.Wrap(err, errApplyFunctionDeployment)
 	}
--- a/internal/controller/rbac/provider/binding/reconciler.go
+++ b/internal/controller/rbac/provider/binding/reconciler.go
@@ -25,7 +25,6 @@ import (
 
 	"github.com/google/go-cmp/cmp"
 	appsv1 "k8s.io/api/apps/v1"
-	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -76,7 +75,7 @@ func Setup(mgr ctrl.Manager, o controller.Options) error {
 		Named(name).
 		For(&v1.ProviderRevision{}).
 		Owns(&rbacv1.ClusterRoleBinding{}).
-		Watches(&corev1.ServiceAccount{}, handler.EnqueueRequestForOwner(mgr.GetScheme(), mgr.GetRESTMapper(), &v1.ProviderRevision{})).
+		Watches(&appsv1.Deployment{}, handler.EnqueueRequestForOwner(mgr.GetScheme(), mgr.GetRESTMapper(), &v1.ProviderRevision{})).
 		WithOptions(o.ForControllerRuntime()).
 		Complete(ratelimiter.NewReconciler(name, errors.WithSilentRequeueOnConflict(r), o.GlobalRateLimiter))
 }
--- a/internal/controller/rbac/provider/binding/reconciler.go
+++ b/internal/controller/rbac/provider/binding/reconciler.go
@@ -24,6 +24,7 @@ import (
 	"time"
 
 	"github.com/google/go-cmp/cmp"
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
@@ -50,7 +51,7 @@ const (
 	timeout = 2 * time.Minute
 
 	errGetPR        = "cannot get ProviderRevision"
-	errListSAs      = "cannot list ServiceAccounts"
+	errDeployments  = "cannot list Deployments"
 	errApplyBinding = "cannot apply ClusterRoleBinding"
 
 	kindClusterRole = "ClusterRole"
@@ -169,27 +170,30 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		return reconcile.Result{Requeue: false}, nil
 	}
 
-	l := &corev1.ServiceAccountList{}
+	l := &appsv1.DeploymentList{}
 	if err := r.client.List(ctx, l); err != nil {
-		err = errors.Wrap(err, errListSAs)
+		err = errors.Wrap(err, errDeployments)
 		r.record.Event(pr, event.Warning(reasonBind, err))
 		return reconcile.Result{}, err
 	}
 
-	// Filter down to the ServiceAccounts that are owned by this
+	// Filter down to the Deployments that are owned by this
 	// ProviderRevision. Each revision should control at most one, but it's easy
 	// and relatively harmless for us to handle there being many.
 	subjects := make([]rbacv1.Subject, 0)
 	subjectStrings := make([]string, 0)
-	for _, sa := range l.Items {
-		for _, ref := range sa.GetOwnerReferences() {
+	for _, d := range l.Items {
+		for _, ref := range d.GetOwnerReferences() {
 			if ref.UID == pr.GetUID() {
+				sa := d.Spec.Template.Spec.ServiceAccountName
+				ns := d.Namespace
+
 				subjects = append(subjects, rbacv1.Subject{
 					Kind:      rbacv1.ServiceAccountKind,
-					Namespace: sa.GetNamespace(),
-					Name:      sa.GetName(),
+					Namespace: ns,
+					Name:      sa,
 				})
-				subjectStrings = append(subjectStrings, sa.GetNamespace()+"/"+sa.GetName())
+				subjectStrings = append(subjectStrings, ns+"/"+sa)
 			}
 		}
 	}
--- a/cluster/charts/crossplane/templates/rbac-manager-clusterrole.yaml
+++ b/cluster/charts/crossplane/templates/rbac-manager-clusterrole.yaml
@@ -20,11 +20,18 @@ rules:
   - ""
   resources:
   - namespaces
-  - serviceaccounts
   verbs:
   - get
   - list
   - watch
+- apiGroups:
+    - apps
+  resources:
+    - deployments
+  verbs:
+    - get
+    - list
+    - watch
 # The RBAC manager creates a series of RBAC roles for each namespace it sees.
 # These RBAC roles are controlled (in the owner reference sense) by the namespace.
 # The RBAC manager needs permission to set finalizers on Namespaces in order to
--- a/internal/controller/pkg/revision/runtime_function_test.go
+++ b/internal/controller/pkg/revision/runtime_function_test.go
@@ -24,6 +24,7 @@ import (
 	"github.com/google/go-cmp/cmp"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/utils/ptr"
@@ -188,6 +189,9 @@ func TestFunctionPostHook(t *testing.T) {
 				manifests: &MockManifestBuilder{
 					ServiceAccountFn: func(overrides ...ServiceAccountOverride) *corev1.ServiceAccount {
 						return &corev1.ServiceAccount{}
+					},
+					DeploymentFn: func(serviceAccount string, overrides ...DeploymentOverride) *appsv1.Deployment {
+						return &appsv1.Deployment{}
 					},
 				},
 				client: &test.MockClient{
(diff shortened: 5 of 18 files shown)


cluster/crds/pkg.crossplane.io_functionrevisions.yaml
```diff
@@ -137,6 +137,20 @@ spec:
                 required:
                 - name
                 type: object
+              serviceAccountRef:
+                description: ServiceAccountRef references the ServiceAccount resource
+                  that is used by the provider revision's runtime.
```
negz, line 142:
```suggestion
                  that is used by the function revision's runtime.
```

internal/controller/pkg/revision/runtime.go
```diff
@@ -68,6 +68,9 @@ var (
 
 // ManifestBuilder builds the runtime manifests for a package revision.
 type ManifestBuilder interface {
+	// ReferencedServiceAccountName returns the name of the referenced service
+	// account that already exist, typically managed externally.
+	ReferencedServiceAccountName() string
```
negz, line 73: Nit: I find interfaces like this a little easier to read when there is whitespace between the method and the following comment string. Like the one immediately below, for example.

negz: The method seems a little out of place in this interface, where all the other methods build and return manifests.

What does it method return if there isn't a referenced service account?

internal/controller/pkg/revision/runtime_function.go
```diff
@@ -115,16 +115,37 @@ func (h *FunctionHooks) Post(ctx context.Context, pkg runtime.Object, pr v1.Pack
...
+	// Persist it only if it's not set or if it has changed.
+	if saRef := pr.GetServiceAccountRef(); saRef == nil || (saRef.Name != saName || saRef.Namespace != d.Namespace) {
+		pr.SetServiceAccountRef(&v1.ServiceAccountReference{
+			Name:      saName,
+			Namespace: d.Namespace,
+		})
+		if err := h.client.Apply(ctx, pr); err != nil {
+			return errors.Wrap(err, "cannot apply function revision with service account reference")
+		}
+	}
```
negz, line 147: I find it a little surprising/unintuitive that this ref is set by the post hook as part of processing a runtimeconfig. I generally think of information as flowing from revision spec -> generated manifests. I suppose this manifests -> revision spec flow does match how we set TLS certificate names there today.

negz (commented):
Out of curiosity, would it be feasible for us to keep using owner references to track the relationship between a provider and a service account, but not take exclusive ownership? i.e. Be careful to add the provider an owner reference (without overwriting any existing references), and to not make the provider a controller of the service account unless it actually  creates it?

internal/controller/pkg/revision/runtime.go
negz: Answering my own question - I see it returns the empty string if there isn't a referenced service account.

internal/controller/pkg/revision/runtime_function.go
turkenh: Agreed. I am not happy with this either, but this seems to be better than the other alternative. 

When I try to move this to the same place where we set TLS certificate names, which is a package reconciler,
- I needed to duplicate the places where we're reading the DeploymentRuntimeConfig.
- I needed to make the package manager aware of the deployment namespace (to set SA namespace), which is only a runtime concept and currently only used by RuntimeHooks.
- I still needed to set `revision.Spec.ServiceAccountRef` here if no existing one is configured and we're building/creating a new one. 

I think the main difference here is, that we're not only using this field to transfer information only from "package -> revision" but also "revision -> rbac manager". So, when I try to move things up for the first part, I still need to persist in the service account (when we're creating it) for the second part. So, I would prefer having it in one place rather than duplicating logic code in two places.

turkenh:
> Out of curiosity, would it be feasible for us to keep using owner references to track the relationship between a provider and a service account, but not take exclusive ownership? i.e. Be careful to add the provider an owner reference (without overwriting any existing references), and to not make the provider a controller of the service account unless it actually creates it?

But wouldn't it end up owning and eventually deleting an SA that is externally managed and passed us just to use? We could implement a finalizer logic to remove ownerRef **IF** it is an SA that we haven't created but it gets ugly, e.g. how to decide that it is an SA that we didn't create and garbage collector dance if SA is changed later etc. Using owner references to decide which SA is used by a revision runtime feels like a bit stretch to me.

apis/pkg/v1/package_runtime_types.go
```diff
@@ -46,6 +46,19 @@ type PackageRevisionRuntimeSpec struct {
 	// certificates of the Provider.
 	// +optional
 	TLSClientSecretName *string `json:"tlsClientSecretName,omitempty"`
+
+	// ServiceAccountRef references the ServiceAccount resource that is used
+	// by the provider revision's runtime.
```
phisco, line 51: it this saying "provider" because we only expect to use and respect it for Providers?

phisco: I'd say that's not the case, given that you updated also functions to use that. However, didn't we decide Functions should not have to have access to the kubernetes API, so should we just make it a no-op for functions? 🤔

internal/controller/rbac/provider/binding/reconciler.go
```diff
@@ -169,29 +169,16 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		return reconcile.Result{Requeue: false}, nil
 	}
 
-	l := &corev1.ServiceAccountList{}
-	if err := r.client.List(ctx, l); err != nil {
-		err = errors.Wrap(err, errListSAs)
-		r.record.Event(pr, event.Warning(reasonBind, err))
-		return reconcile.Result{}, err
+	if pr.Spec.ServiceAccountRef == nil {
```
pedjak, line 172: do we have unit test for this change?

pedjak (commented):
do changes introduced in this PR could benefit from additional e2e tests?

internal/controller/pkg/revision/runtime.go
negz: > The method seems a little out of place in this interface, where all the other methods build and return manifests.

I'm wondering whether it's possible to avoid adding this method. Could you:

1. Update the `Deployment` implementation to not call `DeploymentWithServiceAccount` if `template.serviceAccountName` is already set. (Or update `DeploymentWithServiceAccount` to be a no-op in that case.)
2. Call `ServiceAccount` but not apply the resulting service account.
3. Call `Deployment` and check whether the `template.serviceAccountName` matches the service account's

The idea being that if the deployment has a `template.serviceAccountName` that isn't the service's we assume we don't need to create a service.

I could be pretty easily convinced that this is worse/more magic than what you're doing here - just thinking out loud to see whether there's a way to avoid this out-of-place seeming interface method.

internal/controller/pkg/revision/runtime_function.go
negz: Another idea: does the provider deployment's `spec.template.spec.serviceAccountName` always point to the SA that we should bind roles to?

If so, could the RBAC manager find the SA it wants to operate on by listing deployments owned by the provider, and  granting access to the `serviceAccountName` they reference? That would let us avoid having to add this new field to the revision type.

internal/controller/pkg/revision/runtime.go
turkenh: Thanks, I agree that your suggested flow is better. 
Updated accordingly.

internal/controller/pkg/revision/runtime_function.go
turkenh: Thanks, I like this as this enables fixing the issue with less diff and checking the service account from the Deployment makes  much sense (e.g. I would do that if I need to figure out the SA being used for a provider revision).

Updated accordingly.

apis/pkg/v1/package_runtime_types.go
turkenh: Thanks for the suggestion, I believe those changes are gone in the latest state.

internal/controller/rbac/provider/binding/reconciler.go
turkenh: Changed the implementation as per Nic's feedback and tried to cover the new changes with unit tests.

turkenh:
> do changes introduced in this PR could benefit from additional e2e tests?

Yes, I added a new test case named `TestExternallyManagedServiceAccount`.

internal/controller/rbac/provider/binding/reconciler.go
```diff
@@ -169,27 +169,30 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		return reconcile.Result{Requeue: false}, nil
 	}
 
-	l := &corev1.ServiceAccountList{}
+	l := &appsv1.DeploymentList{}
 	if err := r.client.List(ctx, l); err != nil {
```
negz, line 173: We could probably filter this down to the namespace where we expect runtime deployments to exist (i.e. `crossplane-system`). Not a blocker though - this would be true for the previous service account based implementation too.

negz (approved):

Issue #4573, opened 2023-09-05
Implement Go Templates as a Composition Function


negz:
### What problem are you facing?

We think there will be several generic/reusable Composition Functions. Functions that folks can grab (e.g. from https://marketplace.upbound.io) and use without needing to write code.

I think a great place to start with these would be a "Helm-like" Function. Helm's Go Templates are a familiar UX to a lot of folks in the Crossplane community. There's also an emergent pattern we call "abstracted Helm" where folks are effectively using provider-helm to do Composition - an XR renders a Helm release, which in turn "composes" several managed resources. It's important to make sure we can replace this pattern with Composition Functions.

### How could Crossplane help solve your problem?

We should build a Helm-like Composition Function. You can see [a sketch of a possible UX](https://github.com/crossplane/crossplane/blob/6869a5f62420fd1f914980894c482c2ba8ffa470/design/design-doc-composition-functions.md#calling-a-function-1) in the Functions design. @turkenh also put together a POC alpha-style Function that you can find at https://github.com/negz/xfns/tree/main/v1alpha1/xfn-gotemplate.

Note that we believe targeting Go Templates makes more sense than Helm because:

* We probably can't support any-old Helm chart (our "Values" would be the `RunFunctionRequest`)
* We probably don't need any of the packaging/release stuff - we want more like `helm template render`

We do want to give folks all the tools Helm has though, so the Function should support http://masterminds.github.io/sprig/ and potentially any special Go templates functions Helm supports.

turkenh:
I believe it would be helpful to come up with a one-pager (if not a design document) clarifying the API and how we would map things like connection details (e.g [this](https://github.com/negz/xfns/blob/main/v1alpha1/xfn-gotemplate/example/templates/service-account-key.yaml#L15-L16)) in a template dir to what Crossplane expects.

PR #4935, opened 2023-10-31
Add one-pager for go-templating function

ezgidemirel:

Fixes https://github.com/crossplane/crossplane/issues/4573

This PR adds a one-pager to discuss `go-templating` function user interface. Some of these concepts are already implemented the [PR](https://github.com/crossplane-contrib/function-go-templating/pull/1) with `v1alpha1` version
and can be changed based on the discussion.  

I have: 

### DIFF
--- a/design/one-pager-function-go-templating.md
+++ b/design/one-pager-function-go-templating.md
@@ -58,6 +58,8 @@ spec:
 apiVersion: iam.aws.upbound.io/v1beta1
 kind: AccessKey
 metadata:
+  annotations:
+    gotemplating.fn.crossplane.io/composition-resource-name: sample-access-key-{{ $i }}
   name: sample-access-key-{{ $i }}
 spec:
   forProvider:
@@ -130,7 +132,7 @@ spec:
 
 ## Helper Functions
 
-A set of helper functions like `getComposedResource` or `getEnvVar` will be 
+A set of helper functions like `getComposedResource` or `getCompositionEnvVar` will be 
 provided to make it easier to access resources or environment config variables 
 referenced in the composition. Without these functions, users can access 
 resources or environment config variables by using the `index` function like 
@@ -173,7 +175,7 @@ metadata:
     testing.upbound.io/example-name: test-user-{{ $i }}
     {{ $composed := getComposedResource $.observed ( print "test-user-" $i ) }}
     dummy: {{ default ( randomChoice "foo" "bar" "baz" ) $composed.resource.metadata.labels.dummy }}
-    env: {{ getEnvVar $. "key1" }}
+    env: {{ getCompositionEnvVar $. "key1" }}
   name: test-user-{{ $i }}
 spec:
   forProvider: {}
--- a/design/one-pager-function-go-templating.md
+++ b/design/one-pager-function-go-templating.md
@@ -1,0 +1,183 @@
+# Go Templating Function
+ 
+* Owner: Ezgi Demirel (@ezgidemirel)
+* Reviewers: Crossplane Maintainers
+* Status: Accepted
+
+## Background
+
+With Crossplane 1.14, Composition Functions will be promoted to `v1beta1`, and 
+several composition functions will be introduced as an extension point to be 
+used in pipelines. One of these extensions is the [Go Templating Function], 
+which allows users to render Crossplane resources using Go templating 
+capabilities like conditionals and loops. In addition to that, users can use 
+values in the environment configs or other resource fields to create/configure
+Crossplane resources.
+
+## Getting Template Files
+
+With the initial implementation of the Go Templating Function, users will be 
+able to provide templates in two different ways:
+- An `inline` source can be used to provide template files as a string in the
+  `input.inline.template` field of the pipeline step.
+- A `fileSystem` source can be used to provide template files with folder path
+  in the `input.fileSystem.dirPath` field of the pipeline step.
+
+Both of these options have their own advantages and disadvantages. While 
+`inline` source is faster and easier to use, it is not suitable for large 
+template files and hard to maintain. On the other hand, `fileSystem` source
+requires the files to be present in the local file system, which might be 
+possible by mounting config maps with `DeploymentRuntimeConfig` or building 
+another container image on top of the existing one with the template files. 
+
+In the next iterations, it is planned to support fetching template files from a
+remote location like GitHub repositories or S3 buckets.
+
+## Rendering Templates and Accessing Resources
+
+To make rendering templates and accessing resource fields easier, function 
+requests will be converted to `map[string]any` and passed to the 
+template engine. This will allow users to access resource fields in the 
+template files like:
+
+```yaml
+{{- range $i := until ( .observed.composite.resource.spec.count | int ) }}
+---
+apiVersion: iam.aws.upbound.io/v1beta1
+kind: User
+metadata:
+  annotations:
+    gotemplating.fn.crossplane.io/composition-resource-name: test-user-{{ $i }}
+  name: test-user-{{ $i }}
+  labels:
+    testing.upbound.io/example-name: test-user-{{ $i }}
+spec:
+  forProvider: {}
+
+---
+apiVersion: iam.aws.upbound.io/v1beta1
+kind: AccessKey
+metadata:
+  name: sample-access-key-{{ $i }}
+spec:
+  forProvider:
+    userSelector:
+      matchLabels:
+        testing.upbound.io/example-name: test-user-{{ $i }}
+  writeConnectionSecretToRef:
+    name: sample-access-key-secret-{{ $i }}
+    namespace: crossplane-system
+{{- end }}
+
+```
+
+To compare and prepare the desired composed resources map with consistent keys,
+`function-go-templating` requires a special annotation 
+`gotemplating.fn.crossplane.io/composition-resource-name` to be assigned to the
+input resources. This annotation will match the resource name in the template
+file with the resource name in the desired resources.
+
+## Configuring Composite Connection Details
+
+A new meta resource, `CompositeConnectionDetails` with API version 
+`meta.gotemplating.fn.crossplane.io/v1alpha1` will be used to set connection 
+details of the composite resource. This resource will contain a `data` field,
+which expects a map with string keys and base64 encoded string values. This 
+meta resource is not a real Crossplane resource, and it will not be applied to
+the cluster.
+
+```yaml
+apiVersion: meta.gotemplating.fn.crossplane.io/v1alpha1
+kind: CompositeConnectionDetails
+metadata:
+  annotations:
+    gotemplating.fn.crossplane.io/composition-resource-name: connection-details
+  name: connection-details
+{{ if eq $.observed.resources nil }}
+data: {}
+{{ else }}
+data:
+  username: {{ ( index $.observed.resources "sample-access-key-0" ).connectionDetails.username }}
+  password: {{ ( index $.observed.resources "sample-access-key-0" ).connectionDetails.password }}
+  url: {{ "http://www.example.com" | b64enc }}
+{{ end }}
+
+```
+
+## Configuring Resource Status
+
+In addition to that, a new custom annotation, 
+`gotemplating.fn.crossplane.io/ready` will be introduced to set the ready 
+status of the resource, which can be assigned to `Unspecified`, `True` or 
+`False`. This annotation will be removed after the resource is rendered and not
+applied to the cluster.
+
+```yaml
+apiVersion: kubernetes.crossplane.io/v1alpha1
+kind: ProviderConfig
+metadata:
+  name: composed-providerconfig
+  annotations:
+    gotemplating.fn.crossplane.io/ready: True
+spec:
+  credentials:
+    source: Secret
+    secretRef:
+      namespace: crossplane-system
+      name: cluster-config
+      key: kubeconfig
+```
+
+## Helper Functions
+
+A set of helper functions like `getComposedResource` or `getEnvVar` will be 
+provided to make it easier to access resources or environment config variables 
+referenced in the composition. Without these functions, users can access 
+resources or environment config variables by using the `index` function like 
+below:
+
+```yaml
+{{- range $i := until ( .observed.composite.resource.spec.count | int ) }}
+---
+apiVersion: iam.aws.upbound.io/v1beta1
+kind: User
+metadata:
+  annotations:
+    gotemplating.fn.crossplane.io/composition-resource-name: test-user-{{ $i }}
+  name: test-user-{{ $i }}
+  labels:
+    testing.upbound.io/example-name: test-user-{{ $i }}
+  {{ if eq $.observed.resources nil }}
+    dummy: {{ randomChoice "foo" "bar" "baz" }}
+  {{ else }}
+    dummy: {{ ( index $.observed.resources ( print "test-user-" $i ) ).resource.metadata.labels.dummy }}
+  {{ end }}
+    env: {{ (index $.context "apiextensions.crossplane.io/environment").key1 }}
+spec:
+  forProvider: {}
+{{- end }}
+```
+
+With the helper functions, users can access resources or environment config 
+variables like:
+
+```yaml
+{{- range $i := until ( .observed.composite.resource.spec.count | int ) }}
+---
+apiVersion: iam.aws.upbound.io/v1beta1
+kind: User
+metadata:
+  annotations:
+    gotemplating.fn.crossplane.io/composition-resource-name: test-user-{{ $i }}
+  labels:
+    testing.upbound.io/example-name: test-user-{{ $i }}
+    {{ $composed := getComposedResource $.observed ( print "test-user-" $i ) }}
+    dummy: {{ default ( randomChoice "foo" "bar" "baz" ) $composed.resource.metadata.labels.dummy }}
+    env: {{ getEnvVar $. "key1" }}
+  name: test-user-{{ $i }}
+spec:
+  forProvider: {}
+{{- end }}
+```
+
+[Go Templating Function]: https://github.com/crossplane-contrib/function-go-templating


negz:
Awesome, great design @ezgidemirel! Please wrap the text at 80 chars, but otherwise I approve.

I can't literally approve right now due to what appears to be a bug in some new GitHub UI I'm seeing. 🤔

negz (approved):

design/one-pager-function-go-templating.md
```diff
@@ -0,0 +1,171 @@
...
+  in the `input.fileSystem.dirPath` field of the pipeline step.
+
+Both of these options have their own advantages and disadvantages. While `inline` source
+is faster and easier to use, it is not suitable for large template files and hard to maintain.
+On the other hand, `fileSystem` source requires the files to be present in the local file system,
+which might be possible by mounting config maps with `DeploymentRuntimeConfig` or building another
+container image on top of the existing one with the template files. 
+
+In the next iterations, it is planned to support fetching template files from a remote location like
+GitHub repositories or S3 buckets.
```
Piotr1215, line 32: This is fascinating. Would pushing to git repository (approving PR) enable GitOps scenarios where the changed template would be fetched by the running composition and updated in the system by creating a new composition revision? There are some pros/cons to this.

design/one-pager-function-go-templating.md
```diff
@@ -0,0 +1,171 @@
...
+* Reviewers: Crossplane Maintainers
+* Status: Accepted
+
+## Background
+
+With Crossplane 1.14, Composition Functions will be promoted to `v1beta1`, and several 
+composition functions will be introduced as an extension point to be used in pipelines. 
+One of these extensions is the [Go Templating Function], which allows users to render Crossplane
+resources using Go templating capabilities like conditionals and loops. In addition to that, 
+users can use values in the environment configs or other resource fields to create/configure
```
Piotr1215, line 13: Is there an example somewhere how to do it? If I understand correctly, this looks like the _values_ file in helm. How is it different from the `environmentConfigs`?

design/one-pager-function-go-templating.md
```diff
@@ -0,0 +1,171 @@
...
+      matchLabels:
+        testing.upbound.io/example-name: test-user-{{ $i }}
+  writeConnectionSecretToRef:
+    name: sample-access-key-secret-{{ $i }}
+    namespace: crossplane-system
+{{- end }}
+
+```
+
+To compare and prepare the desired composed resources map with consistent keys, `function-go-templating` requires a special
```
Piotr1215, line 71: Is this similar to the functionality as the `- name` key in a standard composition resource?

design/one-pager-function-go-templating.md
```diff
@@ -0,0 +1,171 @@
...
+annotation `gotemplating.fn.crossplane.io/composition-resource-name` to be assigned to the input resources. This
```
Piotr1215, line 72: This could be a good idea for a _helper function_, something like `setName()` so composition authors don't have to remember the whole annotation.

design/one-pager-function-go-templating.md
```diff
@@ -0,0 +1,171 @@
...
+annotation will match the resource name in the template file with the resource name in the desired resources.
+
+## Configuring Composite Connection Details
```
Piotr1215, line 75: Can you help me understand the motivation behind this design? Is the providerConfigRef(Name) still being used to connect to the provider and this serves different purpose? Also what is `meta resource` :)?

Just to make sure the terminology is the same, in the platform team we would typically use the term `composite` to mean the actual composition and separately XR to mean the `composite resource`. In this case the `CompositeConnectionDetails` provides connection details for the XR or the composition?

design/one-pager-function-go-templating.md
```diff
@@ -0,0 +1,171 @@
...
+```
+
+## Helper Functions
+
+A set of helper functions like `getComposedResource` or `getEnvVar` will be provided to make it
+easier to access resources or environment config variables referenced in the composition. Without these functions,
+users can access resources or environment config variables by using the `index` function like below:
+
+```yaml
+{{- range $i := until ( .observed.composite.resource.spec.count | int ) }}
```
Piotr1215, line 129: Could we create a special variable for `.observed.composite.resource`? Maybe something like .xrd that would give access to all the fields in the XRD?

design/one-pager-function-go-templating.md
```diff
@@ -0,0 +1,171 @@
...
+apiVersion: iam.aws.upbound.io/v1beta1
+kind: User
+metadata:
+  annotations:
+    gotemplating.fn.crossplane.io/composition-resource-name: test-user-{{ $i }}
+  labels:
+    testing.upbound.io/example-name: test-user-{{ $i }}
+    {{ $composed := getComposedResource $.observed ( print "test-user-" $i ) }}
+    dummy: {{ default ( randomChoice "foo" "bar" "baz" ) $composed.resource.metadata.labels.dummy }}
+    env: {{ getEnvVar $. "key1" }}
```
Piotr1215, line 164: It would be great to also provide an option for default, something like:
`env: {{ getEnvVarOrDefault $. "key1" "default_value" }}

design/one-pager-function-go-templating.md
```diff
@@ -0,0 +1,171 @@
...
+spec:
+  credentials:
+    source: Secret
+    secretRef:
+      namespace: crossplane-system
+      name: cluster-config
+      key: kubeconfig
+```
+
+## Helper Functions
```
Piotr1215, line 122: Are we planning to support the template authoring process with and LSP? It would be great if there is autocompletion for variables and helper functions in an IDE.

Piotr1215 (approved):
Thank you for putting this together @ezgidemirel . I've left a few non-blocking comments/questions.

design/one-pager-function-go-templating.md
ezgidemirel: We can implement something like `templatePullPolicy`. But as you mentioned, providing a gitOps tooling has pros and cons.

design/one-pager-function-go-templating.md
ezgidemirel: Yes, definitely. Since we don't have any `name` field inside the templates, we're forcing composition writers to set this annotation.

design/one-pager-function-go-templating.md
Piotr1215: So those are actual environmental variables that the pod has access to as opposite to the `environmentConfig` file?

ezgidemirel: No, they're `environmentConfig` variables. Maybe naming the function as `getEnvConfigValue` makes more sense.

Piotr1215: Thank you for clarifying, I need to catch up on the `environmentConfig` design :)

design/one-pager-function-go-templating.md
ezgidemirel: > Can you help me understand the motivation behind this design? Is the providerConfigRef(Name) still being used to connect to the provider and this serves different purpose?

Sorry, I couldn't understand this question. Can you explain a bit more?

> Just to make sure the terminology is the same, in the platform team we would typically use the term composite to mean the actual composition and separately XR to mean the composite resource. In this case the CompositeConnectionDetails provides connection details for the XR or the composition?

`CompositeConnectionDetails` is a meta resource which configures composite resource's connection details. And by meta resource what I mean is, it's not a real k8s resource. But it contains metadata information for other resources.

design/one-pager-function-go-templating.md
ezgidemirel: We already have a predefined function called `default`. The example you gave can be written as:

```
env: {{ default "default_value" ( getEnvVar $. "key1" ) }}
```

But still, we can also add `getEnvVarOrDefautl` as a helper function.

design/one-pager-function-go-templating.md
ezgidemirel: It's not in the current plan. But that's a good idea :)

phisco (approved):
Awesome work @ezgidemirel! 👏

design/one-pager-function-go-templating.md
Piotr1215: Thank you for clarifying. What I meant by my first question is that I'm trying to understand what is the purpose of the  `CompositeConnectionDetails` and why do we need it. Is it somehow equivalent to `[writeConnectionSecretsToNamespace](https://docs.crossplane.io/v1.13/concepts/compositions/#hl-15-5)`? Most probably I'm just misunderstanding something.

design/one-pager-function-go-templating.md
Piotr1215: I think the `default` is better as it doesn't need introducing additional constructs and already covers the desired functionality.

design/one-pager-function-go-templating.md
ezgidemirel: It's equivalent to [here](https://docs.crossplane.io/v1.13/concepts/compositions/#hl-18-14). Assume that you define several resources in your composition and want to propagate some of their connection details like username/password from one resource and endpoint from another. In that case, you can use this meta resource to collect different information from different resources and write it to composite resource's connection secret (the one created with the UID of the composite resource).

design/one-pager-function-go-templating.md
ezgidemirel: We don't have access to the XRD in functions.

design/one-pager-function-go-templating.md
```diff
@@ -0,0 +1,183 @@
...
+  labels:
+    testing.upbound.io/example-name: test-user-{{ $i }}
+spec:
+  forProvider: {}
+
+---
+apiVersion: iam.aws.upbound.io/v1beta1
+kind: AccessKey
+metadata:
+  name: sample-access-key-{{ $i }}
```
bobh66, line 61: Is the `gotemplating` annotation required here too?

design/one-pager-function-go-templating.md
```diff
@@ -0,0 +1,183 @@
...
+    source: Secret
+    secretRef:
+      namespace: crossplane-system
+      name: cluster-config
+      key: kubeconfig
+```
+
+## Helper Functions
+
+A set of helper functions like `getComposedResource` or `getEnvVar` will be
```
bobh66, line 133: I think `getEnvVar` will be confusing - the suggestion to use `getEnvConfigVar` or something similar is more descriptive.

design/one-pager-function-go-templating.md
ezgidemirel: Yes, it is. Nice catch!

design/one-pager-function-go-templating.md
Piotr1215: Apologies, when I say XRD I mean the fields defined in the XRD and available in claim/XR.

design/one-pager-function-go-templating.md
ezgidemirel: Discussed with @phisco offline. Apparently, there may be more key/value pairs coming apart from `environmentConfig` resources with `"apiextensions.crossplane.io/environment"` annotation. Therefore using `getCompositionEnvVar` makes more sense here. However, we should definitely document `Composition Environment` and try to prevent the confusion.

phisco: Don't forget it's an alpha feature. Clarifying the terms I think will play a key role in the promotion to beta of the functionality.

design/one-pager-function-go-templating.md
```diff
@@ -0,0 +1,185 @@
...
+    gotemplating.fn.crossplane.io/composition-resource-name: connection-details
+  name: connection-details
+{{ if eq $.observed.resources nil }}
+data: {}
+{{ else }}
+data:
+  username: {{ ( index $.observed.resources "sample-access-key-0" ).connectionDetails.username }}
+  password: {{ ( index $.observed.resources "sample-access-key-0" ).connectionDetails.password }}
+  url: {{ "http://www.example.com" | b64enc }}
+{{ end }}
```
ytsarev, line 105: This is great, I was just recently solving complex custom secret structure creation together with one of our customers. The templating will simplify the solution a lot and will make it much more straightforward.

design/one-pager-function-go-templating.md
```diff
@@ -0,0 +1,185 @@
...
+{{- range $i := until ( .observed.composite.resource.spec.count | int ) }}
+---
+apiVersion: iam.aws.upbound.io/v1beta1
+kind: User
+metadata:
+  annotations:
+    gotemplating.fn.crossplane.io/composition-resource-name: test-user-{{ $i }}
+  labels:
+    testing.upbound.io/example-name: test-user-{{ $i }}
+    {{ $composed := getComposedResource $.observed ( print "test-user-" $i ) }}
```
ytsarev, line 176: just to clarify, the `getComposedResource ` can access any of the composed resources involved in Composition?

design/one-pager-function-go-templating.md
```diff
@@ -0,0 +1,185 @@
...
+
+Both of these options have their own advantages and disadvantages. While 
+`inline` source is faster and easier to use, it is not suitable for large 
+template files and hard to maintain. On the other hand, `fileSystem` source
+requires the files to be present in the local file system, which might be 
+possible by mounting config maps with `DeploymentRuntimeConfig` or building 
+another container image on top of the existing one with the template files. 
+
+In the next iterations, it is planned to support fetching template files from a
+remote location like GitHub repositories or S3 buckets.
```
ytsarev, line 34: Do we plan to package templates to OCI images?

In the abstracted helm approach( which is kind of the contextual predecessor to the Go templating function) I was packaging helm chart together with the main Configuration and used the same tag to create a relation between them and then uptest them together.

So we have two OCI repos there
* platform-ref-apigateway for the main Configuration
* platform-ref-apigateway-chart for the Templates/Chart

This approach enabled me to create reproducible builds of both the Configuration and the Templates(Chart) that are used within this Configuration.

The main concern with just free style pulling templates from arbitrary places is that Templates will have totally different lifecycle from the main Composition. This concern is probably valid for the cases when we are mixing in together with the standard patch & transform function(similar to https://github.com/upbound/platform-ref-apigateway/blob/main/package/composition.yaml) or some other one. 

 In any case, testing Configuration that is pulling the logic from the remote sources does impose the testing and eventual reliability challenges as we might pull in some code that might lead to undesired side effects.

design/one-pager-function-go-templating.md
```diff
@@ -0,0 +1,185 @@
...
+`False`. This annotation will be removed after the resource is rendered and not
+applied to the cluster.
+
+```yaml
+apiVersion: kubernetes.crossplane.io/v1alpha1
+kind: ProviderConfig
+metadata:
+  name: composed-providerconfig
+  annotations:
+    gotemplating.fn.crossplane.io/ready: True
```
ytsarev, line 123: Is it analog for ReadinessChecks in P&T?

ytsarev (commented):
That looks great, @ezgidemirel ! I've left some comments and discussion points

design/one-pager-function-go-templating.md
ezgidemirel: `getComposedResource` will take two parameters. The first one is the `observed` map which contains **all of the** composed resources. The second parameter is the key which is equal to the value of `gotemplating.fn.crossplane.io/composition-resource-name` annotation.

design/one-pager-function-go-templating.md
ezgidemirel: I agree with your concerns. We'll enable so many different GitOps combinations for compositions. Even with config map mounting + Argo can cause unexpected side effects.

design/one-pager-function-go-templating.md
ezgidemirel: The annotation can be used to mark a composed resource `Ready`.

turkenh (approved):

ytsarev (commented):
@ezgidemirel In addition to https://helm.sh/docs/chart_template_guide/function_list/ , are we considering future extension of the go templating function with the infra specific functions similar to https://developer.hashicorp.com/terraform/language/functions/cidrsubnet ? It might be super useful in general and especially for folks who are migrating from terraform

ezgidemirel:
> @ezgidemirel In addition to https://helm.sh/docs/chart_template_guide/function_list/ , are we considering future extension of the go templating function with the infra specific functions similar to https://developer.hashicorp.com/terraform/language/functions/cidrsubnet ? It might be super useful in general and especially for folks who are migrating from terraform

The sky is the limit with go-templating function and cloud specific functions will definitely be helpful. But we need to prioritize what we want to include. I'll open a bunch of issues in the repo with possible helper functions and enhancements. Feel free to add 👍 or create new issues.

design/one-pager-function-go-templating.md
turkenh: > Do we plan to package templates to OCI images?

I like this, I believe this is a nice interim step between all inline templates and read from git which could probably give us what we want with the minimal effort and without dealing with all sort of complexities. 

We can have an API like:

```
      input:
        apiVersion: gotemplating.fn.crossplane.io/v1beta1
        kind: GoTemplate
        source: image
        image:
          repo: turkenh/my-templates
          tag: v0.1.0
```

and read templates inside the image from a predefined path, e.g. `/templates`. Coupling/decoupling the lifecycle of templates with the configuration package could be left to the author depending on their use case.

@ezgidemirel please open an issue for this as well 🙏

bmutziu:
From an actual customer starting to translate EKS Blueprints for Terraform, I subscribe to @ytsarev initiative related to the enhancement of the go-templating function with particular IaC functions ! For example, to follow the Custom Networking best practice in order for the CNI to acquire enough secondary IP address, without using the CIDRSubnet function, we are forced to resort to a static assignment of like a dozen subnets, classified into various categories like private, public or intra ...

ezgidemirel:
I think we resolved all the comments in this PR. I created several issues in the function-go-templating repo to follow up. 
Can someone with the required permissions merge the PR, please?

PR #5044, opened 2023-11-20
Improve robustness of releasing objects

turkenh:

Deactivation of package revisions may fail with the following log/event when there is a high number of objects managed by the package and Crossplane/K8s is under load. 

```
2023-11-20T12:01:18Z    DEBUG   crossplane      Event(v1.ObjectReference{Kind:"ConfigurationRevision", Namespace:"", : type: 'Warning' reason: 'DeactivateRevision' cannot deactivate package revision: cannot release objects: cannot update owned object: Composition/redacted: Put "[https://redacted:443/apis/apiextensions.crossplane.io/v1/compositions/redacted](https://redacted/apis/apiextensions.crossplane.io/v1/compositions/redacted)": http2: client connection force closed via ClientConn.Close
```
or 
```
  Warning  DeactivateRevision  8m47s  packages/configurationrevision.pkg.crossplane.io  cannot deactivate package revision: cannot release objects: cannot update owned object: Composition/redacted: Put "https://redacted:443/apis/apiextensions.crossplane.io/v1/compositions/redacted": context canceled
```

The code is brittle as is since it goes over all managed objects sequentially and returns in case of failure at any point. It also makes an update call against kube API no matter there is a change in the owner references or not.

This PR aims to improve this situation with the following in the release object method:
- Run in parallel (similar to establishment)
- Only update if needed
 
I have: 

### DIFF
--- a/internal/controller/pkg/revision/establisher.go
+++ b/internal/controller/pkg/revision/establisher.go
@@ -128,7 +128,7 @@ func (e *APIEstablisher) Establish(ctx context.Context, objs []runtime.Object, p
 
 // ReleaseObjects removes control of owned resources in the API server for a
 // package revision.
-func (e *APIEstablisher) ReleaseObjects(ctx context.Context, parent v1.PackageRevision) error {
+func (e *APIEstablisher) ReleaseObjects(ctx context.Context, parent v1.PackageRevision) error { //nolint:gocyclo // complexity coming from parallelism.
 	// Note(turkenh): We rely on status.objectRefs to get the list of objects
 	// that are controlled by the package revision. Relying on the status is
 	// not ideal as it might get lost (e.g. if the status subresource is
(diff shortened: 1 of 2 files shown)


phisco (approved):

internal/controller/pkg/revision/establisher.go
```diff
@@ -137,39 +137,64 @@ func (e *APIEstablisher) ReleaseObjects(ctx context.Context, parent v1.PackageRe
...
-		if err := e.client.Get(ctx, types.NamespacedName{Name: u.GetName()}, &u); err != nil {
-			if kerrors.IsNotFound(err) {
-				// This is not expected, but still not an error for relinquishing.
-				continue
+	allObjs := parent.GetObjects()
+	if len(allObjs) == 0 {
+		return nil
+	}
+
+	g, ctx := errgroup.WithContext(ctx)
```
turkenh, line 145: Following the same approach as https://github.com/crossplane/crossplane/blob/256178492e736055caed433f75014b938addd511/internal/controller/pkg/revision/establisher.go#L212-L331

internal/controller/pkg/revision/establisher.go
```diff
@@ -137,39 +137,64 @@ func (e *APIEstablisher) ReleaseObjects(ctx context.Context, parent v1.PackageRe
...
-			return errors.Wrapf(err, errFmtUpdateOwnedObject, u.GetKind(), u.GetName())
-		}
+			if changed {
+				u.SetOwnerReferences(ors)
+				if err := e.client.Update(ctx, &u); err != nil {
+					return errors.Wrapf(err, errFmtUpdateOwnedObject, u.GetKind(), u.GetName())
+				}
+			}
+			select {
+			case <-ctx.Done():
```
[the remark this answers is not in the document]
turkenh: TBH, I try to follow the [existing](https://github.com/crossplane/crossplane/blob/256178492e736055caed433f75014b938addd511/internal/controller/pkg/revision/establisher.go#L212-L331) implementation in the establisher.

But I see your point and agree that it makes more sense.
Just pushed a [commit](https://github.com/crossplane/crossplane/pull/5044/commits/5469e1ff27a1ab2cbf3320149f7bd6d756d78ba9).

internal/controller/pkg/revision/establisher.go
```diff
@@ -137,39 +137,64 @@ func (e *APIEstablisher) ReleaseObjects(ctx context.Context, parent v1.PackageRe
...
+	if len(allObjs) == 0 {
+		return nil
+	}
+
+	g, ctx := errgroup.WithContext(ctx)
+	g.SetLimit(maxConcurrentEstablishers)
+	for _, ref := range allObjs {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
```
sttts, line 150: maybe just break to wait for the other go routines? `g.Wait` below will return the first error of some go routine.

sttts: but then we need another such context check below.

Also note: is `ctx.Err()` really the go routine error? There is a `context.Cause` function for that.

sttts: Yes, looking into the context code, there is `err` ("context cancelled") and `cause` (the go routine error).

internal/controller/pkg/revision/establisher.go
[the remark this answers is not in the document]
turkenh: As we discussed offline, I've moved the `select` to the beginning of the goroutine. 

With this, all remaining goroutines (we set limit as 10) will immediately return with context error instead of making calls to kube API to get the same error.

internal/controller/pkg/revision/establisher.go
turkenh: https://github.com/crossplane/crossplane/pull/5044#discussion_r1400388847

turkenh:
@phisco there is a small change after @sttts's feedback. 
Please check and let me know if your approval is no longer valid.

phisco (approved):
LGTM still

PR #5051, opened 2023-11-21
fix(environment): Revert patch order back to v1.13

### LINKED ISSUE #5050: Composition patch order changed with v1.14.0
### What happened?

There is an undocumented breaking change introduced with #4500 to the ordering of patches, specifically patches from environment patches:

* Before v1.14.0 environment patches have been executed in order with "regular" from-XR patches.
* With v1.14.0 environment patches are no always executed after all from-XR patches have been run.

This breaks with existing patching behaviour and can make compositions that rely on patch ordering fail or even unusable with v1.14.0. For example if the environment contains the default values which might be overriden by an optional user field provided with the XR.

The breaking change has been made in https://github.com/crossplane/crossplane/pull/4500/files#diff-574bed872a28203735722ddf7a98e52ad596702e555ed0dda6525e4e578481f2R225

### How can we reproduce it?

Create a composition with environment patches at index 0 and an XR patch that write to the same field at index 1. In v1.13 the field will contain the XR value. In v1.14 the field will contain the value of the environment.

Crossplane version: `v1.14.0`

MisterMX:

Reinline environment patches into from-XR patches as this would otherwise break with previous behaviours. This breaking changes was introduced with v1.14.

Fixes #5050

I have: 

### DIFF
--- a/internal/controller/apiextensions/composite/composition_render.go
+++ b/internal/controller/apiextensions/composite/composition_render.go
@@ -80,26 +80,19 @@ func RenderFromJSON(o resource.Object, data []byte) error {
 	return nil
 }
 
-// RenderFromCompositePatches renders the supplied composed resource by applying
-// all patches that are _from_ the supplied composite resource.
-func RenderFromCompositePatches(cd resource.Composed, xr resource.Composite, p []v1.Patch) error {
+// RenderFromCompositeAndEnvironmentPatches renders the supplied composed
+// resource by applying all patches that are _from_ the supplied composite
+// resource or are from or to the supplied environment.
+func RenderFromCompositeAndEnvironmentPatches(cd resource.Composed, xr resource.Composite, e *Environment, p []v1.Patch) error {
 	for i := range p {
 		if err := Apply(p[i], xr, cd, patchTypesFromXR()...); err != nil {
 			return errors.Wrapf(err, errFmtPatch, p[i].Type, i)
 		}
-	}
-	return nil
-}
 
-// RenderToAndFromEnvironmentPatches renders the supplied composed resource by
-// applying all patches that are from or to the supplied environment.
-func RenderToAndFromEnvironmentPatches(cd resource.Composed, e *Environment, p []v1.Patch) error {
-	if e == nil {
-		return nil
-	}
-	for i := range p {
-		if err := ApplyToObjects(p[i], e, cd, patchTypesFromToEnvironment()...); err != nil {
-			return errors.Wrapf(err, errFmtPatch, p[i].Type, i)
+		if e != nil {
+			if err := ApplyToObjects(p[i], e, cd, patchTypesFromToEnvironment()...); err != nil {
+				return errors.Wrapf(err, errFmtPatch, p[i].Type, i)
+			}
 		}
 	}
 	return nil
--- a/internal/controller/apiextensions/composite/composition_pt.go
+++ b/internal/controller/apiextensions/composite/composition_pt.go
@@ -48,15 +48,14 @@ const (
 	errFetchDetails  = "cannot fetch connection details"
 	errInline        = "cannot inline Composition patch sets"
 
-	errFmtPatchEnvironment             = "cannot apply environment patch at index %d"
-	errFmtParseBase                    = "cannot parse base template of composed resource %q"
-	errFmtRenderFromCompositePatches   = "cannot render FromComposite patches for composed resource %q"
-	errFmtRenderToCompositePatches     = "cannot render ToComposite patches for composed resource %q"
-	errFmtRenderFromEnvironmentPatches = "cannot render FromEnvironment patches for composed resource %q"
-	errFmtRenderMetadata               = "cannot render metadata for composed resource %q"
-	errFmtGenerateName                 = "cannot generate a name for composed resource %q"
-	errFmtExtractDetails               = "cannot extract composite resource connection details from composed resource %q"
-	errFmtCheckReadiness               = "cannot check whether composed resource %q is ready"
+	errFmtPatchEnvironment           = "cannot apply environment patch at index %d"
+	errFmtParseBase                  = "cannot parse base template of composed resource %q"
+	errFmtRenderFromCompositePatches = "cannot render FromComposite or environment patches for composed resource %q"
+	errFmtRenderToCompositePatches   = "cannot render ToComposite patches for composed resource %q"
+	errFmtRenderMetadata             = "cannot render metadata for composed resource %q"
+	errFmtGenerateName               = "cannot generate a name for composed resource %q"
+	errFmtExtractDetails             = "cannot extract composite resource connection details from composed resource %q"
+	errFmtCheckReadiness             = "cannot check whether composed resource %q is ready"
 )
 
 // TODO(negz): Move P&T Composition logic into its own package?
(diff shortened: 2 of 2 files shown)


internal/controller/apiextensions/composite/composition_render.go
```diff
@@ -80,26 +80,19 @@ func RenderFromJSON(o resource.Object, data []byte) error {
...
 }
 
-// RenderFromCompositePatches renders the supplied composed resource by applying
-// all patches that are _from_ the supplied composite resource.
-func RenderFromCompositePatches(cd resource.Composed, xr resource.Composite, p []v1.Patch) error {
+// RenderFromCompositeAndEnvironmentPatches renders the supplied composed
+// resource by applying all patches that are _from_ the supplied composite
+// resource or are from or to the supplied environment.
+func RenderFromCompositeAndEnvironmentPatches(cd resource.Composed, xr resource.Composite, e *Environment, p []v1.Patch) error {
 	for i := range p {
```
phisco, line 87: nit: wdyt of rewriting it as follows?
```golang
func RenderFromCompositeAndEnvironmentPatches(cd resource.Composed, xr resource.Composite, e *Environment, p []v1.Patch) error {
	for i := range p {
		var src runtime.Object
		switch {
		case filterPatch(p[i], patchTypesFromXR()...):
			src = xr
		case e != nil && filterPatch(p[i], patchTypesFromToEnvironment()...):
			src = e
		default:
			continue
		}
		if err := ApplyToObjects(p[i], src, cd); err != nil {
			return errors.Wrapf(err, errFmtPatch, p[i].Type, i)
		}
	}
	return nil
}
```

MisterMX: I think that is way to complicated and I would stick to `if`.

phisco: With the original implementation it looks like both the `Apply` and `ApplyToObjects` will be called every time unnecessarily (which is actually the case), moving the filtering outside of the Apply calls makes it much more explicit in my opinion, but as I said it's just a nit, so not a blocker.

phisco (approved):
Thanks @MisterMX, except for the nit above LGTM! 

We should port it over to `function-patch-and-transform`: https://github.com/crossplane-contrib/function-patch-and-transform/blob/f6c8462ad358fe7a1f41be7632964f3ae37f0343/fn.go#L200-L247, would you want to address this there too @MisterMX?

MisterMX:
> Thanks @MisterMX, except for the nit above LGTM!
> 
> We should port it over to `function-patch-and-transform`: [crossplane-contrib/function-patch-and-transform@`f6c8462`/fn.go#L200-L247](https://github.com/crossplane-contrib/function-patch-and-transform/blob/f6c8462ad358fe7a1f41be7632964f3ae37f0343/fn.go#L200-L247), would you want to address this there too @MisterMX?

Thanks. Though I am not very familiar with the function code. If possible it would be good if you guys could handle this.

phisco:
@MisterMX could you provide a manifest that would have not worked as expected with the new behaviour or add an e2e too directly?

internal/controller/apiextensions/composite/composition_render.go
```diff
@@ -80,26 +80,19 @@ func RenderFromJSON(o resource.Object, data []byte) error {
...
-// RenderToAndFromEnvironmentPatches renders the supplied composed resource by
-// applying all patches that are from or to the supplied environment.
-func RenderToAndFromEnvironmentPatches(cd resource.Composed, e *Environment, p []v1.Patch) error {
-	if e == nil {
-		return nil
-	}
-	for i := range p {
-		if err := ApplyToObjects(p[i], e, cd, patchTypesFromToEnvironment()...); err != nil {
-			return errors.Wrapf(err, errFmtPatch, p[i].Type, i)
+		if e != nil {
```
bobh66, line 92: I'm probably missing something, but if the from composite patches are applied first, won't the from environment patches override the composite values in cases where they overlap?  I thought we wanted the environment patches to serve as a default that can be overridden by the claim/composite?  Like I said I'm probably missing something.

MisterMX:
> @MisterMX could you provide a manifest that would have not worked as expected with the new behaviour or add an e2e too directly?

<details>
<summary>Composition</summary>

```yaml
---
kind: Composition
metadata:
  labels:
    version: v1
  name: test-composition
spec:
  compositeTypeRef:
    apiVersion: example.com/v1alpha1
    kind: XTest
  environment:
    environmentConfigs:
    - type: Reference
      ref:
        name: test
  resources:
    - name: test
      base:
        kind: EnvironmentConfig
        metadata:
          name: result
        data:
          value: # Should match value in XR
      patches:
        - type: FromEnvironmentFieldPath
          fromFieldPath: value
          toFieldPath: data.value
        - fromFieldPath: spec.testValue
          toFieldPath: data.value
```
</details>

<details>
<summary>XRD</summary>

```yaml
---
metadata:
  name: xtests.example.com
spec:
  claimNames:
    kind: Test
    plural: tests
  defaultCompositionRef:
    name: test-composition
  group: example.com
  names:
    kind: XTest
    listKind: XTestList
    plural: xtests
    singular: xtest
  versions:
  - name: v1alpha1
    referenceable: true
    schema:
      openAPIV3Schema:
        properties:
          spec:
            type: object
            properties:
              testValue:
                type: string
        required:
        - spec
        type: object
    served: true
```
</details>


<details>
<summary>Environment Config</summary>

```yaml
---
kind: EnvironmentConfig
metadata:
  name: test
data:
  value: "From environment"

```
</details>

<details>
<summary>Claim</summary>

```yaml
---
apiVersion: example.com/v1alpha1
kind: Test
metadata:
  name: test
  namespace: default
spec:
  testValue: "From Claim"
```
</details>

If you now get the generated environment config it will show the value from the environment despite being before the from-composite patch:

```bash
$ kubectl get environmentconfig -l crossplane.io/claim-name=test -ojsonpath='{.items[0].data.value}' 
From environment% 
```

internal/controller/apiextensions/composite/composition_render.go
MisterMX: That's how it works in v1.14. In earlier versions the environment patches were executed in order with the regular from-composite patches.

This PR would restore the previous behaviour.

phisco:
@MisterMX,sure, but that's a pathological example, the environment patch was useless in 1.13. I was more interested in a real world example impacted by the behavior change.

MisterMX:
> @MisterMX,sure, but that's a pathological example, the environment patch was useless in 1.13. I was more interested in a real world example impacted by the behavior change.

@phisco I don't understand what you mean. The patches were fully usable in pre v1.14 and we are using it successfully in production.

The example I gave above pretty much represents the way we are using it in our environment.

phisco:
@MisterMX, I mean in your example you have:
```yaml
        - type: FromEnvironmentFieldPath
          fromFieldPath: value
          toFieldPath: data.value
        - fromFieldPath: spec.testValue
          toFieldPath: data.value
```
did the first patch in 1.13 had any actual reason to be defined? it was always shadowed by the second one. That's what I meant by "pathological" example.

phisco:
Got it I missed the composite field was optional.

phisco (approved):

internal/controller/apiextensions/composite/composition_render.go
ravilr: Bob, i think the fix here is to process(apply) the patches `p` in a single pass (loop) rather than the 2 separate loops (1.14 change, which results in the overridden behaviour). the `only` filter last arg to Apply/ApplyToObjects methods ensures that a patch can be either a FromXR types patch or a Environment related patch types.

bobh66: @ravilr thanks - that's what I was missing.

PR #5076, opened 2023-11-30
fix(environment): erroring if no matches in Single mode, configurable minMatch for Multiple

### LINKED ISSUE #5075: Select Required EnvironmentConfig doesnt work properly
### What happened?

Expectation: given Composition consumes at least 2 EnvironmentConfigs with resolution policy: Required. Second EnvCfg uses label selectors and might not be present when XR of the given Composition is created. Therefore no managed resource of XR of given Composition should be created unless all configs are fetched.

Currently: Composite with label selector doesnt wait for 2nd EnvCfg and creates managed resources that need only values from the 1st EnvCfg.

Why is this happening? 

[SelectEnvironment()](https://github.com/crossplane/crossplane/blob/master/internal/controller/apiextensions/composite/reconciler.go#L567) constructs a list of references which are then fetched by [Fetch()](https://github.com/crossplane/crossplane/blob/master/internal/controller/apiextensions/composite/reconciler.go#L575). In case of Refs, user inputs are just added to the list. in case of Selector, an empty list is returned if config with matching labels is not found.

### How can we reproduce it?

- Composition with 2 EnvironmentConfig sources
- both EnvCfg sources or at least the last one uses label selector
- environment resolution policy is required
- make only the 1st EnvCfg present in the environment and create a Claim

Crossplane version: 1.14.2

P0t4T0o:
### Description of Fix

Fixes #5075 

I have: 

- [ ] Added or updated e2e tests.
- [ ] Added `backport release-x.y` labels to auto-backport this PR.

### DIFF
--- a/apis/apiextensions/v1/composition_environment.go
+++ b/apis/apiextensions/v1/composition_environment.go
@@ -194,6 +194,9 @@ type EnvironmentSourceSelector struct {
 	// MaxMatch specifies the number of extracted EnvironmentConfigs in Multiple mode, extracts all if nil.
 	MaxMatch *uint64 `json:"maxMatch,omitempty"`
 
+	// MinMatch specifies the required minimum of extracted EnvironmentConfigs in Multiple mode.
+	MinMatch *uint64 `json:"minMatch,omitempty"`
+
 	// SortByFieldPath is the path to the field based on which list of EnvironmentConfigs is alphabetically sorted.
 	// +kubebuilder:default="metadata.name"
 	SortByFieldPath string `json:"sortByFieldPath,omitempty"`
@@ -207,6 +210,10 @@ func (e *EnvironmentSourceSelector) Validate() *field.Error {
 
 	if e.Mode == EnvironmentSourceSelectorSingleMode && e.MaxMatch != nil {
 		return field.Forbidden(field.NewPath("maxMatch"), "maxMatch is not supported in Single mode")
+	}
+
+	if e.Mode == EnvironmentSourceSelectorSingleMode && e.MinMatch != nil {
+		return field.Forbidden(field.NewPath("minMatch"), "minMatch is not supported in Single mode")
 	}
 
 	for i, m := range e.MatchLabels {
--- a/internal/controller/apiextensions/composite/environment_selector.go
+++ b/internal/controller/apiextensions/composite/environment_selector.go
@@ -38,6 +38,7 @@ const (
 	errFmtReferenceEnvironmentConfig   = "failed to build reference at index %d"
 	errFmtResolveLabelValue            = "failed to resolve value for label at index %d"
 	errListEnvironmentConfigs          = "failed to list environments"
+	errFmtSelectorNotEnoughResults     = "expected at least %d EnvironmentConfig(s) with matching labels, found: %d"
 	errFmtInvalidEnvironmentSourceType = "invalid source type '%s'"
 	errFmtInvalidLabelMatcherType      = "invalid label matcher type '%s'"
 	errFmtUnknownSelectorMode          = "unknown mode '%s'"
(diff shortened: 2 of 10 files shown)


internal/controller/apiextensions/composite/environment_selector.go
```diff
@@ -87,7 +88,7 @@ func (s *APIEnvironmentSelector) SelectEnvironment(ctx context.Context, cr resou
 				s.buildEnvironmentConfigRefFromRef(src.Ref),
 			)
 		case v1.EnvironmentSourceTypeSelector:
-			ec, err := s.lookUpConfigs(ctx, cr, src.Selector.MatchLabels)
+			ec, err := s.lookUpConfigs(ctx, cr, src.Selector.MatchLabels, rev.Spec.Environment.IsRequired())
```
phisco, line 91: nit: what about doing the check outside of the lookupConfigs method? just to avoid the additional bool parameter

internal/controller/apiextensions/composite/environment_selector.go
```diff
@@ -38,6 +38,7 @@ const (
 	errFmtReferenceEnvironmentConfig   = "failed to build reference at index %d"
 	errFmtResolveLabelValue            = "failed to resolve value for label at index %d"
 	errListEnvironmentConfigs          = "failed to list environments"
+	errListEnvironmentConfigsNoResult  = "no EnvironmentConfig Found that matches labels"
```
phisco, line 41:
```suggestion
	errListEnvironmentConfigsNoResult  = "no EnvironmentConfig found that matches labels"
```

P0t4T0o:
> However, `spec.envirnment.policy.resolution/resolve` are at the root level, and should only influence how references are actually handled once defined. In case of label selection, it's tricky, as we need to actually list resources before "saving" them as references. So, wouldn't it make sense to change it according to the mode selected (`spec.environment.environmentConfigs[*].mode`), `Single` meaning exact 1, while `Multiple` meaning at most `maxMatch` (potentially adding a `minMatch` later on). WDYT @P0t4T0o?

Agree with you, as we discussed over slack, even though this would fix the issue and bring back the default behaviour, it is misusing resolution policy for the 2nd thing: construction of list of refs. What we really miss is a control over how many configs should be **at least** matched and put into the list of refs.
I like the idea of `minMatch` field as this would address the problem for `mode: Multiple` without much effort or changes. The single mode should already follow default behaviour:
```go
case v1.EnvironmentSourceSelectorSingleMode:
		switch len(cl.Items) {
		case 1:
			ec = append(ec, cl.Items[0])
		default:
			return nil, errors.Errorf(errFmtFoundMultipleInSingleMode, len(cl.Items))
		}
```
> I would still count it as a fix, as we didn't want to change the default behavior in the first place....

Since addition of `minMatch` shouldnt result in way too many changes than currently this PR brings. I can update the PR and add this field in following days. WDYT @phisco ?

phisco:
Unfortunately, no matches don't result in an error even in `Single` mode as before the switch case you shared we are also checking the length of the array and exiting if it's 0 😅
I'd be fine adding `minMatch` as part of the fix, @P0t4T0o, no issue on my side, maybe we could keep the default (`nil`) to 0 though.

P0t4T0o:
> Unfortunately, no matches don't result in an error even in `Single` mode as before the switch case you shared

ahh right, I checked the wrong function :D

apis/apiextensions/v1/composition_environment.go
```diff
@@ -194,6 +194,10 @@ type EnvironmentSourceSelector struct {
 	// MaxMatch specifies the number of extracted EnvironmentConfigs in Multiple mode, extracts all if nil.
 	MaxMatch *uint64 `json:"maxMatch,omitempty"`
 
+	// MinMatch specifies the required minimum of extracted EnvironmentConfigs in Multiple mode.
+	// +kubebuilder:default=1
+	MinMatch uint64 `json:"minMatch,omitempty"`
```
phisco, line 199:
```suggestion
	MinMatch *uint64 `json:"minMatch,omitempty"`
```
Given that `minMatch` is not meant to be set in Single mode, would it be possible to keep it a pointer too, remove the default and expand the check we do for MaxMatch in EnvironmentSourceSelector.Validate method to also enforce this not being set in Single mode?

We do already have a few places where `nil` is handled as `1`, but in this case I'd make it mean `0`, just to avoid introducing a breaking change also for `Multiple` mode.

I think it's fine to have `Single` mode meaning "exactly" one match, while `Multiple` defaults to any number of matches, but can be constrained further if needed.

phisco (commented):
The `resolve` policy `Required` is a little bit confusing, but technically as we discussed, is not about the behavior of these selectors, but rather what Crossplane should with the references only once they are set. So, what do you think about rewording the PR title to something like "fix(environment): erroring if no matches in Single mode, configurable minMatch for Multiple"?

PR #5098, opened 2023-12-07
crank: inject dotenv file into docker runtime env vars

jaylevin:

Fixes # 

I have: 

- [ ] Added or updated unit tests.
- [ ] Added or updated e2e tests.
- [ ] Linked a PR or a [docs tracking issue] to [document this change].
- [ ] Added `backport release-x.y` labels to auto-backport this PR.

### DIFF
--- a/cmd/crank/beta/render/runtime_docker.go
+++ b/cmd/crank/beta/render/runtime_docker.go
@@ -22,13 +22,13 @@ import (
 	"io"
 	"net"
 
+	"github.com/crossplane/crossplane-runtime/pkg/errors"
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/container"
 	"github.com/docker/docker/client"
 	"github.com/docker/docker/errdefs"
 	"github.com/docker/go-connections/nat"
-
-	"github.com/crossplane/crossplane-runtime/pkg/errors"
+	"github.com/joho/godotenv"
 
 	pkgv1beta1 "github.com/crossplane/crossplane/apis/pkg/v1beta1"
 )
@@ -43,6 +43,9 @@ const (
 	// used to run the Function. By default render assumes the Function package
 	// (i.e. spec.package) can be used to run the Function.
 	AnnotationKeyRuntimeDockerImage = "render.crossplane.io/runtime-docker-image"
+
+	// AnnotationKeyRuntimeDockerEnvFile configures the dotenv file to load into the containers environment.
+	AnnotationKeyRuntimeDockerEnvFile = "docker.render.crossplane.io/env-file"
 )
 
 // DockerCleanup specifies what Docker should do with a Function container after
@@ -94,6 +97,9 @@ type RuntimeDocker struct {
 
 	// PullPolicy controls how the runtime image is pulled.
 	PullPolicy DockerPullPolicy
+
+	// EnvFile specifies the path to a dotenv file used to inject environment variables into the container.
+	EnvFile string
 }
 
 // GetDockerPullPolicy extracts PullPolicy configuration from the supplied
@@ -143,6 +149,9 @@ func GetRuntimeDocker(fn pkgv1beta1.Function) (*RuntimeDocker, error) {
 	if i := fn.GetAnnotations()[AnnotationKeyRuntimeDockerImage]; i != "" {
 		r.Image = i
 	}
+	if f := fn.GetAnnotations()[AnnotationKeyRuntimeDockerEnvFile]; f != "" {
+		r.EnvFile = f
+	}
 	return r, nil
 }
 
@@ -184,6 +193,14 @@ func (r *RuntimeDocker) Start(ctx context.Context) (RuntimeContext, error) { //n
 		if err != nil {
 			return RuntimeContext{}, errors.Wrapf(err, "cannot pull Docker image %q", r.Image)
 		}
+	}
+
+	if r.EnvFile != "" {
+		envs, err := parseEnvFile(r.EnvFile)
+		if err != nil {
+			return RuntimeContext{}, errors.Wrapf(err, "cannot parse dotenv file %q", r.EnvFile)
+		}
+		cfg.Env = envs
 	}
 
 	// TODO(negz): Set a container name? Presumably unique across runs.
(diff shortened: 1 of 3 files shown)


phisco (commented):
What kind of information would you like to pass to the function through envs? Would it be possible to pass that through the input configuration instead? I'm not entirely sure we want to encourage using env vars to pass configuration to functions.

jaylevin:
> What kind of information would you like to pass to the function through envs? Would it be possible to pass that through the input configuration instead? I'm not entirely sure we want to encourage using env vars to pass configuration to functions.

It would be sensitive/secret values that cannot be comitted to the composition repo, so using the input is not an option… The other option I can think of is to use a DeploymentRuntimeConfig.. Does beta render currently support/have plans to support passing a DeploymentRuntimeConfig?

phisco:
No, that's not support at the moment and it would be hard to do that without running Functions in an actual Kubernetes cluster. My only reluctance is that if we start allowing env variable, we might end up reinventing docker compose 😅

Moreover, per the spec Functions should not assume to be deployed in any specific way, nor to have network connectivity, this is in fact to allow running them also in more constrained contexts such as the beta render command, so if environment variables are to inject secrets to connect to external systems, we should first discuss relaxing our definition.

jaylevin:
@phisco That makes sense and I agree we shouldn't assume them to be deployed in any specific way..

However, this issue states that functions are expected to make requests/queries to external systems: https://github.com/crossplane/crossplane/issues/3718

The purpose of this PR is to provide a solution for the below problem:
- There is currently no way to pass credentials to a function that makes authenticated requests to an external system via the `crossplane beta render` CLI.

This feature would be useful for the scenario where we want to "integration test" a composition that uses xfns that communicate with external systems. The only other solution I can think of is to **expect** functions that communicate with external systems that are being executed via `crossplane beta render` to use a mocked client.

Interested to hear your thoughts on this, mainly these two options:
1. Should `crossplane beta render` support executing/testing a composition in an integrated fashion? (ie provide the capability to pass sensitive values to a function running via `crossplane beta render` (either locally or via CI machine)
2. OR should `crossplane beta render` explicitly **not** support this feature. Instead, it would be documented/clarified that `crosslane beta render` is to only be used for unit testing compositions/composition functions. It would be up to the function to use a mocked client when it recognizes it is being executed via `crossplane beta render`.

phisco:
`render` already supports externally run Functions, so nothing is stopping users from running them however they want, e.g. a docker compose file, the only requirement being exposing the needed port, and setting up functions accordingly. So, I don't see it as should `render` supports integration testing for functions with external connectivity, but rather if we want `render` to be able to run functions however people wants, which I think we could consider out of scope.

jaylevin:
> `render` already supports externally run Functions, so nothing is stopping users from running them however they want, e.g. a docker compose file, the only requirement being exposing the needed port, and setting up functions accordingly. So, I don't see it as should `render` supports integration testing for functions with external connectivity, but rather if we want `render` to be able to run functions however people wants, which I think we could consider out of scope.

@phisco I might not have done a good job explaining the use case for this PR and caused more confusion than understanding. However, I still am fairly confident there is a missing feature that can offer some value to teams developing/testing custom functions. I'd like to try one more time at explaining the motivation behind this PR and the feature gap it intends to fill.


**User story:** As a function/composition developer, I want to inject environment variables and/or pass CLI arguments to a Function in `Docker` runtime mode when the function is being executed via `crossplane beta render`.

- As you mentioned, it is possible to run functions with `Development` runtime mode however we want and attach environment variables/pass arguments to the function with docker compose or plain `docker` CLI. However, this PR specifically focuses on functions running in `Docker` runtime mode. There is currently no way to attach environment variables or pass arguments to a Function running in `Docker` runtime mode when it is executed with `crossplane beta render`. 

This is the motivation behind this PR: to add the capability to inject environment variables into a Function running in `Docker` runtime mode. The ability to pass arguments to the function running in `Docker` runtime mode is not covered in this PR, but it's the same idea and can be refined in this PR or a follow-up.

Please let me know your thoughts as I am curious if this new explanation changes any of your previously expressed opinions.

negz:
How likely is it that this is the beginning of a "slippery slope"? Do you think env variables is really all we need to set, or would we end up wanting to expose other Docker configuration (of which there's quite a lot) - e.g. ports, volumes, etc etc?

jaylevin:
> How likely is it that this is the beginning of a "slippery slope"? Do you think env variables is really all we need to set, or would we end up wanting to expose other Docker configuration (of which there's quite a lot) - e.g. ports, volumes, etc etc?

@negz That is a valid question and concern and I see your point... I think env variables and args are going to be the most common since those are common inputs to a program. If the function is reading from the filesystem, the path for that volume/directory could be supplied by an env variable/argument, therefore eliminating the need for the volumes to be specified directly. Ports are automatically determined with `Docker` runtime mode so that is also not likely..

I do understand your reluctance on the slippery scope, and for now we have a workaround for this. But just for clarity I'll leave our team's use case that motivated this PR:

- When a developer is developing a composition that uses custom XFN(s), we want to use `Docker` runtime mode. This is for obvious reasons so that the developer does not need to worry about running the XFN locally in order to test the composition with `crossplane beta render`. Another reason is so that we can use the same `Function` manifest deployed in our live environments (which uses `Docker` runtime mode) in our composition unit tests -- ie what we are testing is what we intend to deploy to customers.  This is currently not possible because our custom XFN requires custom arguments/env variables that are not set when it runs in a live environment, forcing us to use `Development` runtime mode when unit testing compositions.
- When a developer is developing a composition function, they use `Development` runtime mode.

Our current workaround is a simple wrapper script around `crossplane beta render` that the composition developer uses when testing their changes locally:
```shell
	container_id=$(shell docker run --detach -p 9443:9443 custom-xfn-image --mock --insecure)
	crossplane beta render ...
	docker kill $$container_id
```

This works, but is not ideal because we now have to keep two copies of our custom `Function` manifest in our composition's repo: 
- one that uses `Docker` runtime mode (for kind+kuttl unit tests) -- these execute in an actual Kind cluster so we can use a `DeploymentRuntimeConfig` to set the custom `--mock` flag
-  one with `Development` runtime mode for `beta render` to work locally.

It's a minor inconvenience at the end of the day, and after your slippery slope comment I can understand the reluctancy to consider this in scope. I will leave this up to the maintainers to close or keep open for further discussion.

negz:
Hrm, I don't feel super strongly either way about this one.

I wonder if there's some third option that could be useful. I'm wondering if there's some general way to allow folks to extend the driver - e.g. shell script hooks or similar.

Issue #3765, opened 2023-02-13
Support Yaml filter expressions in composition patches


nakamume:
### What problem are you facing?

Consider the following composition def
```
resources:
- name: res-a
  base:
    apiVersion: pq.xyz.io/v1beta1
    kind: Workspace
    spec:
      forProvider:
        vars:
        - key: var_a
          value: ""
        - key: var_c
          value: ""
  patches:
  - type: FromCompositeFieldPath
    fromFieldPath: spec.var_a
    toFieldPath: spec.forProvider.vars[0].value
  - type: FromCompositeFieldPath
    fromFieldPath: spec.var_c
    toFieldPath: spec.forProvider.vars[1].value
```

In order to patch the `vars`, the array index numbers need to be mentioned (eg. `toFieldPath: spec.forProvider.vars[0].value`). It's very problematic and error prone when there are lots of fields in the array and/or when fields are added/removed (as this changes the indexes of other fields).

### How could Crossplane help solve your problem?

I tried using [JMESPath](https://jmespath.org/) style filter expressions to patch based on filters. Something like:
```
resources:
- name: res-a
  base:
    apiVersion: pq.xyz.io/v1beta1
    kind: Workspace
    spec:
      forProvider:
        vars:
        - key: var_a
          value: ""
        - key: var_c
          value: ""
  patches:
  - type: FromCompositeFieldPath
    fromFieldPath: spec.var_a
    toFieldPath: spec.forProvider.vars[?key == 'var_a'].value
  - type: FromCompositeFieldPath
    fromFieldPath: spec.var_c
    toFieldPath: spec.forProvider.vars[?key == 'var_c'].value
```

But it seems not supported. Getting following warning:
```
cannot apply the patch at index 0: spec.forProvider.vars is not an object
```

Can we add support for JMESPath style filter expressions. (It would be great if entire JMESPath spec)

bobh66:
Maybe related to #3335

negz:
I'm equal parts intrigued and concerned about potentially adding https://jmespath.org support. Last Kubecon I chatted to some of the Kyverno folks, who have had a good experience with it.

It's not off the table, but it is a lot of DSL-like logic stuffed into opaque strings that seem tough to validate. This is something we've wanted to avoid in Crossplane - you can read about some of our motivation in the background section of the [Composition Function design doc](https://github.com/crossplane/crossplane/blob/v1.11.0/design/design-doc-composition-functions.md). 

[...]

PR #5099, opened 2023-12-08
design: Functions requesting extra resources

### LINKED ISSUE #4739: Allow a Function to request "extra" observed resources

Sometimes when "doing Composition" you need to load data from a resource that isn't the claim, XR, or a composed resource. Examples include:

* Deriving a composed resource field from a property of some unrelated resource - for example setting the `spec.forProvider.vpcID` field of a composed resource to the ARN of an AWS VPC MR that isn't part of the XR.
* Loading `EnvironmentConfig` resources, to setup the Composition environment, per https://github.com/crossplane/crossplane/pull/4632#discussion_r1346177527.

We've sketched out some potential ways to extend the `Composition` type to include the ability to express a set of extra resources that the Composition needs (e.g. in https://github.com/crossplane/crossplane/issues/4583), but the APIs for doing so get convoluted pretty quickly. This is largely because we want to support:

* Arbitrary types of resources.
* Referencing and/or selecting resources.
* Deriving what exactly to select or reference from fields of the XR.

With https://github.com/crossplane/crossplane/pull/4632#discussion_r1346177527 we started discussing more seriously whether a Function should be able (or encouraged) to communicate directly with the API server in order to load extra resources. We're reluctant to do this because it provides a "backdoor" through which Functions can load and mutate Kubernetes resources without Crossplane knowing about it. Crossplane controlling all the inputs to, and being able to track all the outputs from a Function enables powerful tools like `xrender`, per https://github.com/crossplane/crossplane/issues/4723.

An interesting middle ground would be to allow a Function to ask Crossplane to provide it with extra observed resources. Observed resources that the Function doesn't _compose_ - that aren't children of the XR - but are useful to the Function pipeline for some reason. This would allow Functions to load extra resources in a fashion that was still tracked/brokered by Crossplane.

You could imagine an API like this:

```protobuf
message RunFunctionResponse {
  // Existing fields omitted for brevity.

  // Extra resources that should be supplied to all subsequent Functions in
  // the pipeline. The map key uniquely identifies the group of resources.
  map<string, ResourceSelector> extra_resources = 42;
}

message ResourceSelector {
  string api_version = 1;
  string kind = 2;
  map<string, string> match_labels = 3; // Optional. Matches all resources if omitted.
}

// We probably need distinct DesiredState and ObservedState messages,
// since Functions can't return desired state for extra resources.
message ObservedState {
  Resource composite = 1;

  // The 'resources' field becomes 'composed' to more clearly
  // distinguish it from 'extra' resources.
  map<string, Resource> composed = 2;

  // Note that extra resources is a map to Resources, plural.
  // The map key corresponds to the key in a RunFunctionResponse's
  // extra_resources field.
  map<string, Resources> extra = 3;
}

// Resources just exists because you can't have map<string, repeated Resource>.
message Resources {
  repeated Resource resources = 1;
}
```

Whenever a Function returned a `RunFunctionResponse` with `extra_resources`, Crossplane would list those resources and pass them in to any subsequent Functions in the pipeline as the `extra` field of their `ObservedState`.

I think this would allow us to move selecting and merging `EnvironmentConfig` resources entirely into the Composition Function pipeline. For example[^1]:

```yaml
kind: Composition
metadata:
  name: extra
spec:
  mode: Pipeline
  pipeline:
  # This helper Function just sets extra_resources in its output.
  - step: get-env-configs
    functionRef:
      name: function-select-extra-resources
    input:
      apiVersion: extra.fn.crossplane.io/v1beta1
      kind: ExtraResourceSelectors
      selectors:
      - key: env-configs
        kind: EnvironmentConfig
        matchLabels:
          region: au-west
  # This helper Function sets explicit (or "default") env vars.
  # It can also load env vars from an extra resources key, which
  # must contain zero or more EnvironmentConfig resources.
  - step: set-default-environment-vars
    functionRef:
      name: function-set-environment
    input:
      apiVersion: environment.fn.crossplane.io/v1beta1
      kind: Variables
      variables:
      - name: some-var
        value: some-value
      variablesFromEnvConfigs:
        extraResourcesKey: env-configs
```

I believe this would be compatible with `xrender`. You'd need to pass it in any "extra" resources similarly to how you pass in observed resources today using the `--observed-resources` flag. So something like:

```shell
$ xrender xr.yaml composition.yaml functions.yaml --extra-resources=env-configs=envconfigs.yaml
```

Note the `key=filename.yaml` style.

[^1]: This example assumes a "flat" environment as proposed in https://github.com/crossplane/crossplane/issues/4738, but I think it would work if the environment continued to be a complex JSON object.

phisco:

Fixes #4739

I have: 

### DIFF
--- a/design/design-doc-composition-functions-extra-resources.md
+++ b/design/design-doc-composition-functions-extra-resources.md
@@ -1,0 +1,1196 @@
+# Composition Functions - Extra Resources
+
+* Owner: Philippe Scorsolini (@phisco)
+* Reviewers: TBD
+* Status: Draft
+
+## Background
+
+Composition Functions extend Crossplane to support a new way of configuring
+how to reconcile a Composite Resource (XR). Each Function is a gRPC server.
+Crossplane sends state to a Function via a `RunFunctionRequest` RPC. The
+Function is intended to return the desired state via a `RunFunctionResponse`
+RPC.
+
+Currently, a Composition author who needs to access information from a resource
+external to their Composition has the following options:
+- [**ALPHA**] Create an `ObserveOnly` resource from their Composition and
+  reading the required information from its `status`.
+- [**ALPHA**] Create an `EnvironmentConfig`, either manually or from another
+  Composition, with the required information and reference it from their
+  Composition via `spec.environment.environmentConfigs`.
+- Use MR-level referencing, if supported for the source and target of the
+  information, being implemented by the provider.
+
+Each of these strategies has its drawbacks:
+- The creation of `ObserveOnly` resources could result in a significant number
+  of resources being created.
+- The use `EnvironmentConfigs` requires the Composition author to create and
+  maintain a separate resource, which can be cumbersome and error-prone.
+  However, has to be said that there is value in the clean separation of concerns
+  between the `EnvironmentConfig`'s producer and consumer(s).
+- Using MR-level referencing is not always possible, as it requires explicit
+  provider support, and hardly ever crosses a single provider (family) boundaries.
+
+As per the Composition Functions' [specification]:
+```text
+A Function MUST NOT assume it is deployed in any particular way, for example
+that it is running as a Kubernetes Pod in the same cluster as Crossplane.
+
+A Function MUST NOT assume it has network access. A Function SHOULD fail
+gracefully if it needs but does not have network access, for example by
+returning a Fatal result.
+```
+
+A clear consequence of the above assumptions is that a Function should not
+assume access to the Kubernetes API. Which means, that letting Functions fetch
+external resources on their own is not an option, Crossplane needs to provide
+them.
+
+With this proposal we aim to provide Functions a way to request additional
+existing Crossplane resources.
+
+## Options Considered
+
+We established the feature design space by addressing these questions:
+
+- **Q0**: Should Functions be able to request additional resources?
+- **Q1**: How should Functions request additional resources? 
+- **Q2**: Which steps in the pipeline should have access to the requested resources?
+- **Q3**: Should Functions be able to request additional resources incrementally?
+- **Q4**: How should Functions specify what resources they want?
+- **Q5**: How can Composition authors specify what resources a Function can
+  request?
+- **Q6**: What about Resolve and Resolution Policies?
+- **Q7**: How can we support "Realtime Compositions"?
+
+### Q0: Should Functions be able to request additional resources?
+
+Clearly this is a yes or no question, but we need to do a step back. What we
+actually care about is to have a way to allow **Composition authors** to specify
+additional resources Functions should have access to. So, the options we have
+are actually:
+- **Option 1**: Yes, Composition authors should rely on Functions to request
+  additional resources.
+- **Option 2**: No, Composition authors should not need Functions at all to
+  request additional resources, it should be part of the Composition API.
+
+**Option 2** would feel more "native", everything would be handled by
+Crossplane, and we could design an API up to our standards. Ignoring native
+patch-and-transform Compositions for now, which would present their own set of
+issues around handling multiple results, due to the limited expressiveness of
+the downstream language (P&T). Even if we decided to make it only available for
+Functions based compositions, the API would need to be pretty expressive to
+allow Composition authors to specify what resources they want, e.g. by name, by
+label selector, based on fields from the Composite Resource, or even
+another Composed resource someday... The API would need to be at least as
+expressive, and therefore complex, as what we have today for
+`EnvironmentConfigs` at `spec.environment.environmentConfigs`, or even as that
+of [Kyverno][kyverno-context].
+
+**Option 1** is clearly more flexible, as it would allow Function authors to
+expose whatever API to Composition authors through inputs, or just decide it
+according to whatever logic they want. However, this will obviously require at
+least one more round-trip to get the requested resources from Crossplane to the
+Function.
+
+So, it feels like a more sensible choice to only consider the more flexible
+**Option 1**, allowing Function authors to expose whatever API they want to let
+Composition authors specify what resources they want, without needing to commit
+to any specific overly complex API. At least not immediately. We can always
+revisit this decision at a later stage if a perfect or "good enough" API arose
+from Functions and make it a native part of the Composition API.
+ 
+### Q1: How should Functions request additional resources?
+
+We considered the following options:
+- **Option 1**: Functions request additional resources through the
+  `RunFunctionResponse`.
+- **Option 2**: Functions request additional by calling back Crossplane through
+  a new RPC.
+
+We discarded **Option 2** for the following reasons:
+- it would require Crossplane to be a gRPC server too, which would be a
+  significant change to the current architecture, requiring Crossplane to be
+  reachable and addressable from Functions. This would, for example, make
+  `crossplane beta render` much more complex.
+- to be able to handle it securely we would need to implement some way for
+  Crossplane to check the "origin" of the request to validate that the requests
+  are authorized according to the policy defined in the Composition. To do so, we
+  would need to implement some sort of complex short-lived token mechanism.
+
+Therefore, **Option 1** is the only option we took into consideration.
+
+### Q2: Which steps in the pipeline should have access to the requested resources?
+
+We considered the following options:
+- **Option 1**: Only the Function that requested the resources should have
+  access to them.
+- **Option 2**: Only the next step in the pipeline should have access to the
+  requested resources.
+- **Option 3**: All the steps in the pipeline should have access to the
+  requested resources.
+- **Option 4**: Both the Function that requested the resources and the next
+  step in the pipeline should have access to the requested resources.
+
+We discarded **Option 3** because of the following reasons:
+- sending the requested resources to all the steps in the pipeline would
+  significantly increase network usage.
+- sending the requested resources to all the steps in the pipeline would
+  significantly increase the security risk as potentially sensitive information
+  could be leaked to steps that don't need it.
+- how would we handle multiple steps in the pipeline requesting resources? Would
+  Crossplane need to keep accumulating all the requested resources and continue
+  sending them to all the steps in the pipeline? Would be complex to define the
+  expected behavior.
+
+**Option 1**, **Option 2** and **Option 4** are all valid, we'll evaluate them
+in conjunction with the answer to **Q3**.
+
+### Q3: Should Functions be able to request additional resources incrementally?
+
+We considered the following options:
+- **Option 1**: Functions can request additional resources incrementally.
+- **Option 2**: Functions can request additional resources only once.
+
+**Option 1** is more flexible, as it allows Functions to request additional
+resources based on the results of previous requests. However, it also makes the
+implementation more complex. We would want to limit the number of
+iterations to a reasonable number by default and allow the user to override this
+limit if needed. This would allow for example a Function to request a resource,
+get the reference to another resource from it, and then request that resource
+too, and so on. However, implementing it would require us to answer the
+following questions:
+- **Q3.1**: What should Crossplane send over at the `N`-th iteration?
+   - **Option 1 (Q3.1)**: All the resources requested at all the previous
+     iterations, `[0..N-1]`.
+   - **Option 2 (Q3.1)**: Only the resources requested at the previous iteration, `N-1`.
+
+**Option 2 (Q3.1)** is more efficient with respect to both network and memory
+usage. However, given that Functions should be considered stateless, it could
+make the logic of the Functions much more complex if they needed to infer the
+current state from the returned resources.
+
+**Option 1 (Q3.1)** would allow Functions to be completely stateless and much simpler
+in their logic, as they would always receive all the resources they requested at
+all the previous iterations. However, it would require Crossplane to keep track
+of all the resources requested at all the previous iterations, which would make
+the implementation a much more complex and more expensive both in terms of
+memory and network.
+
+However, **Option 2 (Q3.1)** would still allow Functions to implement the same
+behavior as **Option 1 (Q3.1)** by requesting all the resources requested at the
+previous iteration, plus the new ones, at each iteration.
+
+For this reason we decided to only consider **Option 2 (Q3.1)**.
+
+**Option 2 (Q3)** is definitely simpler to implement but also less flexible, and
+it would be equivalent to **Option 2 (Q3.1)** with a limit of `1` iteration,
+therefore we decided to discard it, and only consider **Option 1 (Q3)**.
+
+The cap on the number of iterations could be specified in the `PipelineStep`
+struct as follows:
+```golang
+// A PipelineStep in a Composition Function pipeline.
+type PipelineStep struct {
+	// Other fields omitted for brevity ...
+
+	// MaxIterations is the maximum number of iterations this step should
+    // execute. Set it to -1 to execute it indefinitely.
+	// +kubebuilder:default=1
+    // +optional
+	MaxIterations *int64 `json:"maxIterations,omitempty"`
+}
+```
+
+We can now go back to **Q2** and evaluate the options we left open, in
+conjunction with the decision we've now taken. 
+
+**Option 2 (Q3.1)** imply that the Function itself already has access to the
+requested resources, and therefore already implies at least **Option 1 (Q2)**.
+
+**Option 2 (Q2)** doesn't apply anymore, as we already decided the Function
+itself will have access to the resources it requested.
+
+We decided to go with **Option 1 (Q2)** and discard **Option 4 (Q2)** for the
+following reasons:
+- Crossplane will decide that the Function has finished requesting extra
+  resources at the `<N>`-th iteration once the Function returns an empty list of
+  extra resource requests, so what Crossplane would need to send to the next step
+  in the pipeline is not obvious. All the accumulated resources requested by the
+  Function before? the resources requested at the `<N-1>`-th iteration?
+- Sending the resources to the next step in the pipeline would increase the
+  security risk without any obvious benefit.
+- Functions can already pass additional information to the next step in the
+  pipeline through the `Context` in the `RunFunctionResponse`, and it would be
+  better not to introduce additional ways to do so, without a really good reason.
+
+### Q4: How should Functions specify what resources they want?
+
+This question is orthogonal to the previous ones, as the answer to it
+would be the same regardless of the answers to the previous questions.
+
+We considered the following options:
+- **Option 1**: by name.
+- **Option 2**: via label selector.
+- **Option 3**: both by name and via label selector.
+
+Given that one of the thought exercises we tried to do while designing this
+feature, was to allow reimplementing the same functionality today at
+`spec.environment.environmentConfigs` level, we decided to only consider
+**Option 3**.
+
+Taking into account this decision and the one for **Q3**, the
+`RunFunctionResponse` and `RunFunctionRequest` `protobuf` messages could look as
+follows:
+
+```protobuf
+syntax = "proto3";
+
+message RunFunctionResponse {
+  // Existing fields omitted for brevity
+
+  // Requests that must be satisfied for this Function to run successfully.
+  Requests requests = 5;
+}
+
+message Requests {
+  // Extra resources that this Function requires.
+  // The map key uniquely identifies the group of resources.
+  map<string, ResourceSelector> extra_resources = 1;
+}
+
+message ResourceSelector {
+  string api_version = 1;
+  string kind = 2;
+
+  // Here we would want to use a oneof, but due to the need of wrapping
+  // messages, the syntax would become a bit cumbersome, so we'll see how to handle
+  // that at implementation time.
+  optional string match_name = 3;
+  map<string, string> match_labels = 4;
+}
+
+message RunFunctionRequest {
+  // Existing fields omitted for brevity
+
+  // Note that extra resources is a map to Resources, plural.
+  // The map key corresponds to the key in a RunFunctionResponse's
+  // extra_resources field. If a Function requests extra resources that
+  // don't exist Crossplane sets the map key to an empty Resources
+  // message to indicate that it attempted to satisfy the request.
+  map<string, Resources> extra_resources = 3;
+}
+
+// Resources just exists because you can't have map<string, repeated Resource>.
+message Resources {
+  repeated Resource resources = 1;
+}
+```
+
+We considered moving the `extra_resources` field from the `RunFunctionResponse`
+to the `State` so that it could be set in the `observed` `State` of the
+`RunFunctionRequest`, however, we decided not to do so for the following
+reasons:
+- currently we don't have anything else we need to modify in the observed state
+  between Functions, keeping it so would make the implementation simpler to
+  reason about. 
+- that would be awkward as we would need to ignore that field in the `desired`
+  `State` for both `RunFunctionRequest` and `RunFunctionResponse`.
+Another option would have been to split the observed and desired state into two
+separate messages, however, this would be a breaking change for the existing
+API, we could reconsider this choice for `v1beta2`, or `v1`.
+
+### Q5: How can Composition authors specify what resources Functions can request?
+
+To answer this question, we need to answer the following questions first,
+**Q5.1**: **WHEN** should the policy be evaluated?
+- **Option 1 (Q5.1)**: Before gathering the requested extra resources.
+- **Option 2 (Q5.1)**: After having gathered the requested extra resources.
+
+**Option 1 (Q5.1)** would be more efficient, as we would not request any
+resource in case of an invalid request. However, it would be less flexible and
+more coarse-grained, as the policy would only have to take into account the
+information provided in the request, so as per the decision taken above for
+**Q4**, only names and labels. While with **Option 2 (Q5.1)** the policy would
+have the whole resources available and could therefore be much more expressive.
+
+We decided to only consider **Option 1 (Q5.1)**, as we deemed unnecessary to
+have the extra flexibility of **Option 2 (Q5.1)**, and we could always add that
+later, if needed, but there would still be value in having a simpler API to
+cover the majority of the use cases.
+
+So, decided that the policy should be evaluated before actually gathering the
+requested extra we can answer **Q5.2**: What should be the policy's input?
+- **Option 1 (Q5.2)**: The request only.
+- **Option 2 (Q5.2)**: Also the observed state.
+
+**Option 1 (Q5.2)** would require a less expressive policy language that could
+exactly mirror the extra resources requests language, e.g. only names and
+labels, and therefore would be extremely simple to validate against the request.
+
+Note that while Functions could still implement label selectors based on the
+current observed state, e.g. a label having a key or value based on a field in
+the Composite Resource, a policy language having as input only the request
+without any other input would have to be coarser-grained.
+
+**Option 2 (Q5.2)** would allow for a much more expressive policy language, as
+it would have access to the current observed state, and as long as it "rendered"
+to something easily comparable to the requests language, it would be easy to
+implement too, e.g. JSONPath expressions, Crossplane-flavored fieldpaths or
+regexes.
+
+Kubernetes users are already used to RBAC policies being evaluated only against
+the request. So, we deemed sufficient, at least as a first implementation, to
+choose **Option 1 (Q5.2)**. We can always revisit this decision at a later stage
+if needed, but a coarser-grained and simpler policy language would still be
+useful.
+
+Note: we could reconsider **Option 2 (Q5.1)** at a later stage and add a more
+expressive policy language to be evaluated after having gathered the requested
+extra resources, for example using a full-fledged policy language such as
+[rego](https://www.openpolicyagent.org/docs/latest/policy-language/), or a
+similar syntax as the one used by [Kyverno](https://kyverno.io/), or a list of
+[Common Expression Language](https://github.com/google/cel-go) (CEL)
+expressions, which would allow the Composition author to specify much more
+complex policies.
+
+Given the above choices, the answer to **Q5** is already decided. The policy
+language will have to mirror the request language. To make this a bit more
+flexible, we could allow specifying regexes for the names and labels.
+
+An example Composition could look like the following:
+```yaml
+apiVersion: apiextensions.crossplane.io/v1
+kind: Composition
+metadata:
+  name: example
+spec:
+  mode: Pipeline
+  pipeline:
+    - step: get-extra-resources
+      functionRef:
+        name: function-extra-resources
+      allowedExtraResources:
+          # all Something.example.test.com/v1 resources
+        - apiVersion: example.test.com/v1
+          kind: Something
+          # both name and matchLabels to nil
+          
+          # all SomethingElse.example.test.com/v1 resources named according to the provided regex
+        - apiVersion: example.test.com/v1
+          kind: SomethingElse
+          name: "something-else-.*"
+          
+          # all EnvironmentConfigs with label environment=dev, prod or qa
+        - apiVersion: apiextensions.crossplane.io/v1alpha1
+          kind: EnvironmentConfig
+          matchLabels:
+            environment: "(dev|prod|qa)" # "*" would match any label value
+        
+      input:
+        apiVersion: extra.fn.crossplane.io/v1beta1
+        kind: ExtraResourceSelectors
+        # a list of resources by name or label selector
+```
+
+By default, the policy should be to deny all requests, so the Composition author
+would have to explicitly allow the requests they want to allow.
+
+The PipelineStep struct would look like the following:
+```golang
+// A PipelineStep in a Composition Function pipeline.
+type PipelineStep struct {
+	// Step name. Must be unique within its Pipeline.
+	Step string `json:"step"`
+
+	// FunctionRef is a reference to the Composition Function this step should
+	// execute.
+	FunctionRef FunctionReference `json:"functionRef"`
+
+	// Input is an optional, arbitrary Kubernetes resource (i.e. a resource
+	// with an apiVersion and kind) that will be passed to the Composition
+	// Function as the 'input' of its RunFunctionRequest.
+	// +optional
+	// +kubebuilder:pruning:PreserveUnknownFields
+	// +kubebuilder:validation:EmbeddedResource
+	Input *runtime.RawExtension `json:"input,omitempty"`
+
+	// AllowedExtraResources is a list of resources by name or label selector
+	// that this Function is allowed to request.
+	// +optional
+    AllowedExtraResources []ResourceSelector `json:"allowedExtraResources,omitempty"`
+}
+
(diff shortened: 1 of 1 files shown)


design/design-doc-composition-functions-extra-resources.md
```diff
@@ -0,0 +1,1196 @@
...
+
+Composition Functions extend Crossplane to support a new way of configuring
+how to reconcile a Composite Resource (XR). Each Function is a gRPC server.
+Crossplane sends state to a Function via a `RunFunctionRequest` RPC. The
+Function is intended to return the desired state via a `RunFunctionResponse`
+RPC.
+
+Currently, a Composition author who needs to access information from a resource
+external to their Composition has the following options:
+- [**ALPHA**] Create an `ObserveOnly` resource from their Composition and
```
jeanduplessis, line 17: Isn't management policies now in BETA?

design/design-doc-composition-functions-extra-resources.md
```diff
@@ -0,0 +1,1196 @@
...
+- Use MR-level referencing, if supported for the source and target of the
+  information, being implemented by the provider.
+
+Each of these strategies has its drawbacks:
+- The creation of `ObserveOnly` resources could result in a significant number
+  of resources being created.
+- The use `EnvironmentConfigs` requires the Composition author to create and
+  maintain a separate resource, which can be cumbersome and error-prone.
+  However, has to be said that there is value in the clean separation of concerns
+  between the `EnvironmentConfig`'s producer and consumer(s).
```
jeanduplessis, line 31:
```suggestion
- The use of `EnvironmentConfigs` requires the Composition author to create and
  maintain a separate resource, which can be cumbersome and error-prone.
  However, it has to be said that there is value in the clean separation of concerns
  between the `EnvironmentConfig`'s producer and consumer(s).
```

design/design-doc-composition-functions-extra-resources.md
```diff
@@ -0,0 +1,1196 @@
...
+- Using MR-level referencing is not always possible, as it requires explicit
+  provider support, and hardly ever crosses a single provider (family) boundaries.
```
jeanduplessis, line 33:
```suggestion
- Using MR-level referencing is not always possible, as it requires explicit
  provider support, and hardly ever crosses a single provider's (family) boundaries.
```

design/design-doc-composition-functions-extra-resources.md
```diff
@@ -0,0 +1,1196 @@
...
+
+A Function MUST NOT assume it has network access. A Function SHOULD fail
+gracefully if it needs but does not have network access, for example by
+returning a Fatal result.
+```
+
+A clear consequence of the above assumptions is that a Function should not
+assume access to the Kubernetes API. Which means, that letting Functions fetch
+external resources on their own is not an option, Crossplane needs to provide
+them.
```
ytsarev, line 48: I am not against the proposed design in general but is it a correct interpretation of the specification statement?

> A Function SHOULD fail
gracefully if it needs but does not have network access

sounds to me that we can expect the scenarios when Function requires network access but should fail gracefully in case it does not have it.

It does not sound to me that the network access should be fully avoided.

design/design-doc-composition-functions-extra-resources.md
```diff
@@ -0,0 +1,1196 @@
...
+
+Each of these strategies has its drawbacks:
+- The creation of `ObserveOnly` resources could result in a significant number
+  of resources being created.
+- The use of `EnvironmentConfigs` requires the Composition author to create and
+  maintain a separate resource, which can be cumbersome and error-prone.
+  However, it has to be said that there is value in the clean separation of concerns
+  between the `EnvironmentConfig`'s producer and consumer(s).
+- Using MR-level referencing is not always possible, as it requires explicit
+  provider support, and hardly ever crosses a single provider's (family) boundaries.
```
ytsarev, line 33: In practice, all 3 strategies can be currently combined in the same Crossplane Configuration.

EnvironmentConfig keeps the environment-dependant value -> ObserveOnly MR is using the value from EnvConfig to construct external-name for the import -> MR level selectors are used to reference the ObserveOnly MR from another MRs in the same Composition

design/design-doc-composition-functions-extra-resources.md
```diff
@@ -0,0 +1,1196 @@
...
+    functionRef:
+      name: function-environment-configs
+    allowedExtraResources:
+    - apiVersion: apiextensions.crossplane.io/v1alpha1
+      kind: EnvironmentConfig
+    input:
+      apiVersion: extra.fn.crossplane.io/v1beta1
+      kind: ExtraResourceSelectors
+      extraResourceSelectors:
+        environmentConfigs:
```
ytsarev, line 370: I can imaging environmentConfig selection rendering just by looking at it:) It will be next level DX when working with EnvironmentConfigs 👍

design/design-doc-composition-functions-extra-resources.md
phisco: Sure, that's why I say Functions SHOULD NOT fetch external resources on their own. They obviously can, but if they do, running them in less standard ways, e.g. `crossplane beta render`, would be much harder. So, we should at least provide some kind of native mechanism to allow doing that without breaking our assumptions for an "easy-to-run" Function.

phisco:
This could potentially allow `function-patch-and-transform` to request and cache CRDs for all MRs and XRs needed to validate the `input` configuration, allowing to port over the schema-aware validation today performed from Crossplane's Validating Webhook for "native" Compositions.

design/design-doc-composition-functions-extra-resources.md
```diff
@@ -0,0 +1,1196 @@
...
+another Composed resource someday... The API would need to be at least as
+expressive, and therefore complex, as what we have today for
+`EnvironmentConfigs` at `spec.environment.environmentConfigs`, or even as that
+of [Kyverno][kyverno-context].
+
+**Option 1** is clearly more flexible, as it would allow Function authors to
+expose whatever API to Composition authors through inputs, or just decide it
+according to whatever logic they want. However, this will obviously require at
+least one more round-trip to get the requested resources from Crossplane to the
+Function.
```
turkenh, line 546: Nothing comes for free. Flexibility is the obvious benefit, but another drawback is distributing the logic to multiple places. Instead of being able to say, "this pipeline in my composition requires values from the EnvironmentConfig named foo" in a single place (e.g. composition.yaml), now I need to do that in two distinct places, "my composition can request EnvironmentConfigs" and in my function "I need EnvironmentConfig named foo". Former would work in most cases but not all unless we commit to a complex API and the latter complicates the overall system. However, reusable functions could alleviate this and I agree that starting with the flexible option makes more sense and better aligns with the direction that Crossplane is heading with the introduction of functions.

design/design-doc-composition-functions-extra-resources.md
```diff
@@ -0,0 +1,1196 @@
...
+useful.
+
+Note: we could reconsider **Option 2 (Q5.1)** at a later stage and add a more
+expressive policy language to be evaluated after having gathered the requested
+extra resources, for example using a full-fledged policy language such as
+[rego](https://www.openpolicyagent.org/docs/latest/policy-language/), or a
+similar syntax as the one used by [Kyverno](https://kyverno.io/), or a list of
+[Common Expression Language](https://github.com/google/cel-go) (CEL)
+expressions, which would allow the Composition author to specify much more
+complex policies.
```
turkenh, line 807: If we make such a decision, wouldn't it be unnecessary to build what is proposed in this design? If we can express exactly which resources could be requested as extra with an API in composition, why should we introduce another way to define what to request (i.e. function responding with extra_resources). 

I believe, this should be something we should consider sooner than later before committing to this design.

```yaml
- step: get-extra-resources
  functionRef:
    name: function-extra-resources
  extraResources:
      # all Something.example.test.com/v1 resources
    - apiVersion: piextensions.crossplane.io/v1alpha1
      kind: EnvironmentConfig
      # Some Kyverno like API ...
      name: "{{observed.composite.spec.environment}}"
```

FWIW, @negz already has some thoughts regarding JMESPath [here](https://github.com/crossplane/crossplane/issues/3765#issuecomment-1428878515).

design/design-doc-composition-functions-extra-resources.md
```diff
@@ -0,0 +1,1196 @@
...
+  // +optional
+  // +kubebuilder:default=Required
+  // +kubebuilder:validation:Enum=Required;Optional
+  Resolution *ResolutionPolicy `json:"resolution,omitempty"`
+}
+```
+
+Originally, references were only resolved once, which is equivalent to the
+current default, `IfNotPresent`, and resolution was `Required` only. Other
+options were added [later][resolve-resolution-policy] to allow more flexibility,
```
turkenh, line 927: nit: link not working.

turkenh (commented):
Great write up @phisco 🙌 

Loved the "**Options Considered**" section which is well structured and filling all the gaps on how the overall design was shaped, even though I got lost sometimes while reading **Qn.x** style notations 😅

phisco:
> even though I got lost sometimes while reading Qn.x style notations 😅

I got lost too 😅 I hope I didn't mess up any part when moving stuff around, if something feels strange feel free to point it out as it might well be a reference I forgot to update. I double checked it, but you never know 😅

design/design-doc-composition-functions-extra-resources.md
```diff
@@ -0,0 +1,1196 @@
...
+			// if there are still extra resources getting requested on the last iteration, error out
+			if len(extraResourcesRequests) > 0 && i == *fn.MaxIterations-1 {
+				return CompositionResult{}, errors.Errorf( /* ... */ )
+			}
+
+			// check the references are in the observed state we just sent to the Function
+			// if not, error out
+			// otherwise add them to resourcesToWatch
+            ws := rs.GetWatchResources()
+			if err := checkAllowedWatchedResources(requestedResources, ws); err != nil {
```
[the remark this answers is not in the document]
phisco: `maxIterations` set to 1 won't allow requesting any resource at all, as 1 means only a single request will be sent. Maybe we could make it default to 2 🤔 

`maxIterations` set to 2 will allow a Function to request some extra resource on the first iteration, getting them back at the second iteration and possibly requesting to watch them in the responses. This allows a Function to request a resource A and watch the same resource, but doesn't allow a Function to watch a resource B, e.g. a Secret, referenced by A, as the Function would need to request B before being able to watch it. I felt this was enough as if you care about just the reference to B, but not its content, you only actually want to watch A, while if you care about B's content, you will have to get it first and so you will be able to then watch it, and will need one more iteration.

design/design-doc-composition-functions-extra-resources.md
```diff
@@ -0,0 +1,1196 @@
...
+The `RunFunctionResponse` and `RunFunctionRequest` protobuf messages have to be
+updated as follows:
+```protobuf
+syntax = "proto3";
+
+message RunFunctionResponse {
+  // Existing fields omitted for brevity
+
+  // Requests that must be satisfied for this Function to run successfully.
+  Requests requests = 5;
```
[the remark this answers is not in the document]
phisco: Yes, it does feel a bit odd, I agree, although we are piggybacking a request in the response.

`Constraints` was another option suggested in the original issue, but it didn't sound right to me for some reason, however it allowed to talk about "constraints to be satisfied", which was a nice way to describe this.`Requirements` allows to do the same, but feels more appropriate to me, so I'll switch to that 🙏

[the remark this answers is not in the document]
phisco: Done, I hope I switched it everywhere 🤞

design/design-doc-composition-functions-extra-resources.md
phisco: With an API as that one, it would be really hard to define iterative requests for resources, e.g. resource X named A references resource Y named B and I want resource B's content. That would require a much more complex API. While a more complex policy language evaluated against the observed state AND previously requested resources, could allow to define a coarser grained policy such as "This step can request resources of type Y only if referenced by a resource of type X ...".

design/design-doc-composition-functions-extra-resources.md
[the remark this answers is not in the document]
negz: I wonder whether it makes sense for maxIterations to be configurable at all. Would it be enough to start with some sane default, just to avoid infinite iteration/retries?

design/design-doc-composition-functions-extra-resources.md
```diff
@@ -0,0 +1,1199 @@
...
+  sending them to all the steps in the pipeline? Would be complex to define the
+  expected behavior.
+
+**Option 1**, **Option 2** and **Option 4** are all valid, we'll evaluate them
+in conjunction with the answer to **Q3**.
+
+### Q3: Should Functions be able to request additional resources incrementally?
+
+We considered the following options:
+- **Option 1**: Functions can request additional resources incrementally.
```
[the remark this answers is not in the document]
phisco: It would be inefficient and Functions can always do that and we'll happily return all resources every time, I think it's an optimization we can implement from the beginning without too much additional effort.

design/design-doc-composition-functions-extra-resources.md
```diff
@@ -0,0 +1,1199 @@
...
+	// Function as the 'input' of its RunFunctionRequest.
+	// +optional
+	// +kubebuilder:pruning:PreserveUnknownFields
+	// +kubebuilder:validation:EmbeddedResource
+	Input *runtime.RawExtension `json:"input,omitempty"`
+
+	// AllowedExtraResources is a list of resources by name or label selector
+	// that this pipeline step is allowed to request.
+	// +optional
+	AllowedExtraResources []ResourceSelector `json:"allowedExtraResources,omitempty"`
```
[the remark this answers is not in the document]
phisco: We could postpone implementing this and have a separate discussion about it.

This said, allowing an external component such as Functions, to get arbitrary additional resources through Crossplane, means it's able to escalate it's (usually zero) privileges to those Crossplane has, which feels like a huge increase in scope w.r.t. the current situation where external components can ONLY access resources THEY previously created. 

I see how Observe-Only resources could feel like they already granted the same "escape" capabilities, however I think they are actually not, what external resources you can get are already constrained by the provider's permissions. `provider-kubernetes`'s "ante litteram" Observe-Only resources do allow to do the same IIRC, but they are constrained by the permissions granted to the provider.

I would be fine not implementing this part, but that would mean we should explicitly declare Functions to be inside our trust boundary and therefore users should be even more careful picking trusted Functions only.

design/design-doc-composition-functions-extra-resources.md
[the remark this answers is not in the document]
phisco: I agree that as a Composition Author picking a low value for `maxIterations` could be a hard task. So we could definitely pick a sane default and postpone providing such knob to a later iteration (😅)

design/design-doc-composition-functions-extra-resources.md
```diff
@@ -0,0 +1,1199 @@
...
+relies on two indexes mapping Composed Resources to the XR, they are referenced
+by. Implementing this would require something much more complex, assuming it's
+even possible, hence we discarded this option.
+
+**Option 2 (Q7.2)** would be the simplest to implement, as it would only require
+to extend the `RunFunctionResponse` with a list of references of resources to
+watch. Crossplane could then keep track of all the requests, until the Function
+doesn't return any more extra resources requests, collect the set of resources
+to watch throughout the whole pipeline, and only at last, properly set them in
+the XR.
```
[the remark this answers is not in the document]
phisco: Doing so would make constraining resources to be watched harder.

Assuming that functions can only request to watch extra resources they requested at the previous iteration allowed to easily check whether the former was a subset of the latter.

Functions can always request and watch all resources at each iteration, if they want to simplify their logic. But knowing that would be inefficient, I think we should allow optimizing it straight away.

design/design-doc-composition-functions-extra-resources.md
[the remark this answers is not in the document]
negz: > We could postpone implementing this and have a separate discussion about it.

I think we should. Even if this does turn out to be something folks need, I suspect the Composition is probably not the right place to configure it.

Consider a case where I write a function, write a Composition that uses that function, then package it as a Configuration and push it to https://marketplace.upbound.io. Then you, a control plane administrator, want to use that Configuration. It's you the control plane administrator, not me the Composition author, who is probably going to have opinions about what resources a function should be allowed to access on your cluster.

If we did want to place constraints around what a function can ask Crossplane to send it, it might make sense to use a control plane scoped allow list [like we do for providers](https://github.com/crossplane/crossplane/blob/0e47c907d53d50a93cd64e6d63523017029e1737/cmd/crossplane/rbac/rbac.go#L75).

negz:
What if we:

* Require functions to return their full set of requirements every time they're called.
* Watch the full set of requirements by default. [^1]
* Store the full set of requirements [^2] (not watches) in XR status.

This approach appeals to me because:

* Requirements become declarative, like desired state. Every time your function is called you return all of its requirements.
* Function authors don't need to think about watches - Crossplane handles that implementation detail for them.
* Storing the requirements means in the "happy path" Crossplane only needs to run a function with requirements once. Crossplane would only need to re-run a function if its requirements differed from those stored in the XR status.

[^1]: I'm assuming it will be uncommon for a function to want extra resources, but for watching those resources to be a problem. If my assumption is wrong, we could make it possible to opt out of watches by adding a flag to each requirement.
[^2]: Effectively the `Requirements` protobuf message serialized as JSON.

phisco:
> What if we:
> 
> * Require functions to return their full set of requirements [1](#user-content-fn-1-c9a3cfe42e60e9713d36d30918e7c2be) every time they're called.
> * Watch the full set of requirements by default. [2](#user-content-fn-2-c9a3cfe42e60e9713d36d30918e7c2be)
> * Store the full set of requirements [3](#user-content-fn-3-c9a3cfe42e60e9713d36d30918e7c2be) (not watches) in XR status.
> 
> This approach appeals to me because:
> 
> * Requirements become declarative, like desired state. Every time your function is called you return all of its requirements.
> * Function authors don't need to think about watches - Crossplane handles that implementation detail for them.
> * Storing the requirements means in the "happy path" Crossplane only needs to run a function with requirements once. Crossplane would only need to re-run a function if its requirements differed from those stored in the XR status.

Regarding the first point, given the same input (observed state + desired state + requested resources at the previous iteration), it should always result in the same resources being requested back. This assumes that the function is stateless and without side effects or access to the external world. Therefore, I would say that the requirements are already "declarative" with the proposed implementation.

> Every time your function is called, it means that you return all of its requirements. 

The above implies that when multiple iterations are necessary (for example, when a composite references resource A, which in turn references another resource B that we are interested in), we have two options. Either we rely on Functions to "cache" previously requested resources in previous iterations and only send new resources, or we send all resources during each iteration, which would be unnecessarily inefficient. With the proposed approach, simpler functions have the option to request all resources at each iteration if desired. However, more complex functions can optimize network usage immediately, albeit at the cost of potentially requiring a more intricate implementation on our end (not entirely sure that's actually the case).

Moreover, if we assume functions to request all resources at each iteration, what should we do if the function requested fewer resources than those requested in a previous iteration (but not none at all, as that would signal it's done)? Should we error out? Should we still send over the previously requested resources? Should we only send over the requested resources? The latter would be equivalent to the proposed approach, with the difference that we would consider the resources requested in the last iteration as "the set of all requested resources", which wouldn't actually be true since we allow requesting fewer resources across iterations. The other two options don't seem like good choices. We could also accumulate resources across iterations, but then how to deal with key conflicts?

Clearly, we are interested in defining the "set of all requested resources" for the sole purpose of caching through the XR status, as defined in your third point. Therefore, we should perhaps discuss the "set of resources we want cached" instead. I agree that this is definitely something we should address, but I think we can postpone it to a later iteration.

Regarding the second point, should we change "watched resources = requested resources"? The current proposal already allows this, but only if the Function explicitly requests it. We could switch to the behavior you suggest and watch all requested resources by default. However, since watching resources can be costly, I suggest keeping it as an "opt-in" feature.

negz:
@phisco I'm not sure I fully understand what problems incremental requirements are solving, so it's possible I'm missing something. A concrete example (or two) would help. What scenarios would become impossible if we re-ran a function at most once, to satisfy its requirements, on each reconcile? How common do we expect those scenarios to be?

> Therefore, I would say that the requirements are already "declarative" with the proposed implementation.

Declarative-ness was probably not the best way to frame my concern. Instead it's the asymmetry with how the rest of the function RPC works.

Consider desired composed resources. A function must always return the desired state for all of its resources. This is true even if Crossplane told the function its desired state was satisfied, by sending it the current observed state of those resources.

With requirements, I think you're proposing that a function that has requirements would communicate to Crossplane that its requirements were satisfied by returning an empty set of requirements. That would be more like a function communicating to Crossplane that its desired state was satisfied by returning empty desired state.

I feel it's worth being as consistent as possible in the RPC, to make writing functions easier to reason about. That's pretty much where my proposed tweaks above came from: trying to make returning requirements more similar to how folks already return desired state.

> [...] we send all resources during each iteration, which would be unnecessarily inefficient

Arguably returning all desired resources every time the function is called is unnecessarily inefficient too. Is there a reason to be more concerned about how much we're sending over the wire when it comes to extra resources? It's not something we've optimized for much in the rest of the function RPC.

Roughly what percent of functions do we think will make use of extra resources? How many of those functions will be asking for a lot of extra resources, relative to the observed and desired state they're already sent?

> but not none at all, as that would signal it's done

I'm not sold on the idea that a function should return an empty set of requirements to indicate that its requirements are satisfied (as opposed to having no requirements). This is why I suggested storing requirements in XR status. Doing so absolves the function of having to tell Crossplane whether its requirements were satisfied at all. Instead the function tells Crossplane what its requirements are, and Crossplane is able to figure out whether or not it has satisfied them to the best of its ability.

> However, since watching resources can be costly, I suggest keeping it as an "opt-in" feature.

This is another area that seems like a tradeoff between cost/efficiency, and cognitive burden for function authors.

Today function authors don't have to think about watches at all. It's a Kubernetes detail that we absolve them of. I'd happily pay some performance cost for that to remain true with extra resources. I lean toward opt-out because:

* We could defer exposing watches as a concept at all until we had signal that folks needed to opt-out.
* If and when we did expose them, only folks who needed to opt-out (a minority?) would need to understand them.

What are some scenarios where we're worried about the performance of watching extra resources? I'm guessing something like a function returning very broad requirements, e.g. all RDS instances?

phisco:
> @phisco I'm not sure I fully understand what problems incremental requirements are solving, so it's possible I'm missing something. A concrete example (or two) would help. What scenarios would become impossible if we re-ran a function at most once, to satisfy its requirements, on each reconcile? How common do we expect those scenarios to be?

@negz, let's assume we are on the first reconciliation loop. The function only receives the composite resource as part of the observed state, so it can only request resources that are directly specified by the composite resource, such as a resource of kind A named `foo`. The function will request that resource back, and Crossplane will fetch it and call the function again with that resource.

On successive reconciliation loops, instead of just on the XR, functions could request resources also based on observed resources, which is more than what we can do today, but less than what could be achieved with iterative requests.

If we were to decide to re-run functions at most once, the maximum number of hops in the resource graphs Functions could do would be one; this would cover the EnvironmentConfig scenario we have today (composite -> extra resource) and a bit more natively (composed resource -> extra resource), while today we have to resort to ObserveOnly resources for that; but not more complex ones (composite/observed resource -> extra resource 1 -> ... -> extra resource N), for no actual reason in my opinion.

If we were also to add caching as you propose, Functions would still be able to achieve multi hops traversal, but across multiple reconciliation loops instead of across multiple iterations in the same reconciliation loop. By requesting some resource specified by the composite resource at the first reconciliation loop, then some resource by that resource at the second reconciliation loop, and so on...

The real question therefore is: do we want functions to be able to request resources on the observed state alone or also on previously requested resources?

I think the former constraint would be difficult to enforce, as the information would be available to Functions and we could not prevent them from using it to request more resources later on at some point.

So, I don't think it would make sense to constraint them to at most one re-run.

> Consider desired composed resources. A function must always return the desired state for all of its resources. This is true even if Crossplane already told the function its desired state was satisfied, by sending it the current observed state of those resources.
>
> With requirements, I think you're proposing that a function that has requirements would communicate to Crossplane that its requirements were satisfied by returning an empty set of requirements. That would be more like a function communicating to Crossplane that its desired state was satisfied by returning empty desired state.
> 
> I feel it's worth being as consistent as possible in the RPC, to make writing functions easier to reason about. That's pretty much where my proposed tweaks above came from: trying to make returning requirements more similar to how folks already return desired state.

I totally get the point.

Assuming we agree we don't want to go down the path of "re-run at most once", should we error out if the set of requested resources shrinks or relax the constraint and consider it done once the requirements stabilise for 2 consecutive iterations, allowing shrinking the set of resources requested if needed? I'd say the latter sounds more reasonable, as I wouldn't be comfortable erroring out for such a trivial reason, but let me know what you think.

I would be fine switching from `empty means done` to `stable means done`, where by stable I mean they are the same as at the previous iteration. I don't think it makes much difference on the Function side, as the former logic would be something like "if resource A is not available request it", while the latter would be "always request resource A", but I agree it would be more "symmetrical", at the cost of a bit more logic on our side.

> If we can cache the last seen requirements for a function, the function doesn't need to tell Crossplane whether its requirements are satisfied. Instead it just tells Crossplane what its requirements are. Crossplane knows that it has satisfied a function's requirements (to the best of its ability) when it calls a function using the cached requirements, and the function returns unchanged requirements.

By doing as I suggest above, we would just cache "stable requirements" instead of "all requirements". However, I still think we should postpone implementing caching at a later stage, as I think it's not going to be so straightforward and feels like premature optimisation, but it's surely worth keeping in mind as a driving principle.

> Arguably returning all desired resources every time the function is called is unnecessarily inefficient too. Is there a reason to be more concerned about how much we're sending over the wire when it comes to extra resources? It's not something we've optimized for much in the rest of the function RPC.
> 
> Roughly what percent of functions do we think will make use of extra resources? How many of those functions will be asking for a lot of extra resources, relative to the observed and desired state they're already sent?

Roughly 77.62% 😬

Jokes aside, it is indeed inefficient, but not sending other the whole observed/desired state would be hard, it would mean we assume Functions to be able to cache them, which we don't want to, so it's not unnecessary; while only sending back exactly the resources requested feels natural, and any over behaviour would be surprising at best.

> This is another area that seems like a tradeoff between cost/efficiency, and cognitive burden for function authors.
> 
> Today function authors don't have to think about watches at all. It's a Kubernetes detail that we absolve them of. I'd happily pay some performance cost for that to remain true with extra resources. I lean toward opt-out because:
> 
> * We could defer exposing watches as a concept at all until we had signal that folks needed to opt-out.
> * If and when we did expose them, only folks who needed to opt-out (a minority?) would need to understand them.
> 
> What are some scenarios where we're worried about the performance of watching extra resources? I'm guessing something like a function returning very broad requirements, e.g. all RDS instances?

Watching resources is a nice-to-have, resources would still be reconciled if we didn't implement it at all, just at longer intervals, but it's something Kubernetes users learned to expect, that's why I wanted to make it opt-in, leaving to the Functions authors to define whether it make sense in their case or not.

Any reconciliation loop means 1 or more calls to 1 or more functions, which means CPU and network costs, I think we should be extra careful with the number of reconciliation loops we trigger.

I would be fine watching by default the stable set of requirements and postponing allowing opting-out more finely later on.

So to sum it up, WDYT of the following?

```golang
// Compose resources using the Functions pipeline.
func (c *FunctionComposer) Compose(ctx context.Context, xr *composite.Unstructured, req CompositionRequest) (CompositionResult, error) {
	// Existing code omitted for brevity...

	// the set of watched resources, will be stored in `status.watchedResourceRefs`,
	// will need to be converted in a sorted list before storing it
	var resourcesToWatch map[string]v1beta1.ResourceRef

	// Run any Composition Functions in the pipeline.
	for _, fn := range req.Revision.Spec.Pipeline {
		
		// used to store the resources fetched at the previous iteration 
		var extraResources map[string]v1beta1.Resources
		// used to store the requirements returned at the previous iteration 
		var requirements map[string]v1beta1.ResourceSelector
		
		for i := int64(0); i < *fn.MaxIterations; i++ {
			req := &v1beta1.RunFunctionRequest{Observed: o, Desired: d, Context: fctx, extraResources: extraResources}

			// Run the Composition Function and get the response.
			rsp, err := c.pipeline.RunFunction(ctx, fn.FunctionRef.Name, req)
			if err != nil {
				return CompositionResult{}, errors.Wrapf(err, /* ... */)
			}

			// Pass the desired state and context returned by this Function to the next iteration.
			d = rsp.GetDesired()
			fctx = rsp.GetContext()
			
			rs := rsp.GetRequirements()
			if rs.Equal(requirements) {
				// the requirements are the same as the previous iteration
				// we need to add the resources requested at the previous iteration
				// to the watched resources
				for _, resources := range extraResources {
					for _, resource := range resources {
						resourcesToWatch[resource.GetAPIVersion() + "." + resource.GetKind() + "." + resource.GetName()] = resource
					}
				}
				break
			}
			
			// store the requirements for the next iteration 
			requirements = rs

			extraResourcesRequirements := rs.GetExtraResources()

			// if there are still extra resources getting requested on the last iteration, error out
			if len(extraResourcesRequirements) > 0 && i == *fn.MaxIterations-1 {
				return CompositionResult{}, errors.Errorf( /* ... */ )
			}

			// Cleanup the extra resources from the previous iteration to store the new ones
			extraResources = make(map[string]v1beta1.Resources)
			
			// Fetch the requested resources and add them to the desired state.
			for name, selector := range extraResourcesRequirements {
				// Fetch the requested resources and add them to the desired state.
				resources, err := fetchResources(ctx, c.client, selector)
				if err != nil {
					return CompositionResult{}, errors.Wrapf(err, /* ... */)
				}
				
				// resources would be nil in case of not found resources
				extraResources[name] = resources
			}

		}
		
		// Existing code omitted for brevity...
	}
	// replace the XR's `status.watchedResourceRefs` with the set of resources to watch
	// making sure to sort them to avoid flapping
	for _, ref := range resourcesToWatch {
		// ...
	}
	// ...

	// Existing code omitted for brevity...
}

// fetchResources fetches resources that match the given selector.
func fetchResources(ctx context.Context, client client.Reader, selector *v1beta1.ResourceSelector) ([]*unstructured.Unstructured, error) {
	// Fetch resources by name or label selector...
	// return nil on not found error
}
```

```protobuf
syntax = "proto3";

message RunFunctionResponse {
  // Existing fields omitted for brevity

  // Requirements that must be satisfied for this Function to run successfully.
  Requirements requirements = 5;
}

message Requirements {
  // Extra resources that this Function requires.
  // The map key uniquely identifies the group of resources.
  map<string, ResourceSelector> extra_resources = 1;
}

message ResourceRef {
  string api_version = 1;
  string kind = 2;
  string name = 3;
}

message ResourceSelector {
  string api_version = 1;
  string kind = 2;

  // here we would actually want to use a oneof, but due to the need of wrapping
  // messages, the syntax would become a bit cumbersome, so we'll see how to handle
  // that at implementation time.
  optional string match_name = 3;
  map<string, string> match_labels = 4;
}

message RunFunctionRequest {
  // Existing fields omitted for brevity

  // Note that extra resources is a map to Resources, plural.
  // The map key corresponds to the key in a RunFunctionResponse's
  // extra_resources field. If a Function requests extra resources that
  // don't exist Crossplane sets the map key to an empty Resources
  // message to indicate that it attempted to satisfy the request.
  map<string, Resources> extra_resources = 3;
}

// Resources just exists because you can't have map<string, repeated Resource>.
message Resources {
  repeated Resource resources = 1;
}
```

negz:
@phisco and I just had a call to discuss. I think we're pretty much on the same page.

We discussed three broad areas:

1. Should we support incrementally loading extra resources?
2. How does a function signal that its requirements are met?
3. What should we do about watches?

I'll summarize what we discussed below.

### Incremental Requirements

I don't feel very strongly about this one. I suspect @phisco is probably right that it would be useful. I'd like to see one or two really specific examples of where it could be useful though, to back that up. Can we think of a concrete idea for a function that would need incremental resources?

Assume we do decided to support incremental requirements from the beginning I don't think we should make the re-run limit configurable. I'd fix it at something like 5 or 10 re-runs, and treat it mostly as a guard to make sure a broken function doesn't get stuck being re-run an infinite number of times.

### Signaling that Requirements are Met

We were mostly already on the same page here. We can consider a function's requirements to be met when they're stable. If Crossplane runs a function with a set of requirements satisfied and the function responds with the same set of requirements Crossplane already satisfied, we'd consider it stable.

I think the key benefit of this approach is that it makes the RunFunctionResponse API consistent, not that it allows us to cache function requirements. Being able to call a function that has requirements only once (with cached requirements) in the happy path would be a nice to have, but not critical.

I definitely don't love needing to store cached function requirements in the XR's status. That seems like it could clutter up the status with verbose Crossplane machinery. If we did cache, maybe something in-memory would be enough? I could imagine two in-memory options:

* We always run a function with requirements at least twice per reconcile. Once to detect its requirements, then once again to satisfy them (and confirm they're met). This way we only cache requirements in memory long enough to re-run a function.
* We cache function requirements across reconciles maybe using an LRU cache with limited slots. This would be more of an optimization.

Another option we discussed was to store a hash of each function's requirements in the XR status. In retrospect I don't think that would be any better than the first in-memory option above. We'd still have to re-run the function at least once per reconcile.

### Watching Extra Resources

I don't feel strongly about whether we watch or not. I mostly just want to avoid function authors needing to think about it. I'd prefer to avoid exposing a knob for them until we have signal from function authors that they want/need it. We discussed trying to use some good-enough heuristic to watch stuff automatically. For example perhaps if a function requests a specific set of extra resources by name we'll watch them, but we won't watch (potentially large) groups of resources matched by label.

negz:
> Assuming we agree we don't want to go down the path of "re-run at most once", should we error out if the set of requested resources shrinks or relax the constraint and consider it done once the requirements stabilise for 2 consecutive iterations, allowing shrinking the set of resources requested if needed? I'd say the latter sounds more reasonable, as I wouldn't be comfortable erroring out for such a trivial reason, but let me know what you think.

@phisco Yeah, I think we should allow requirements to shrink. Crossplane would treat that as the function no longer having whatever additional requirements it had in the past.

phisco:
@negz, I've [updated](https://github.com/crossplane/crossplane/pull/5099/commits/ee75f9e7de151688254f59283f4e63adfe3444d7) the doc accordingly.

In the meantime I tried unofficially surveying around whether being able to iteratively request extra resources would be useful or not and couldn't find a real world example, so I'd say we can start small and either pick a relatively safe low number, e.g. 5, or directly setting it to 2 and see if there is demand for more iterations later. WDYT?

negz:
> I'd say we can start small and either pick a relatively safe low number, e.g. 5, or directly setting it to 2 and see if there is demand for more iterations later. WDYT?

I don't feel strongly. Maybe split the difference and cap it at 3? 😄

negz (approved):
Looks good, thank you @phisco!

I'd find this design easier to read if you refactored the handful of things we decided not to do (yet) into headings under an Alternatives Considered and/or Potential Future Improvements heading. Right now the information is all there, but a little hard to follow (there's the summarized bullet points and then later notes throughout the "Questions" section). I'm fine merging as-is though, given we're hoping to get this done pretty quick.

PR #5143, opened 2023-12-18
Log at info level reconciliation progress

pedjak:

* In order to get better insights how often the reconciliation
  gets triggered for a given claim/composite, added
  `LoggingInterceptor` that logs info messages before
  and after.
* Consolidated the claim/composite logger context, i.e.
  removed unneeded noise,
  so that we have now the following fields available for filtering:
    * level
    * ts
    * msg
    * controller
    * request
* Composite controller name in the logs follows the convention `composite/<XR_NAME_PLURAL>.<XRD_GROUP>`
* Claim controller name follows in the logs the convention `claim/<CLAIM_NAME_PLURAL>.<XRD_GROUP>`

For example, running `TestCompositionMinimal` e2e tests produces the following log:

```sh
{"level":"info","ts":"2023-10-27T10:25:17Z","logger":"crossplane","msg":"Reconciling","controller":"composite/xnopresources.nop.example.org","request":{"name":"apiextensions-composition-minimal-pdh85"}}
{"level":"info","ts":"2023-10-27T10:25:17Z","logger":"crossplane","msg":"Reconciling","controller":"claim/nopresources.nop.example.org","request":{"name":"apiextensions-composition-minimal","namespace":"default"}}
{"level":"info","ts":"2023-10-27T10:25:17Z","logger":"crossplane","msg":"Reconciliation done","controller":"claim/nopresources.nop.example.org","request":{"name":"apiextensions-composition-minimal","namespace":"default"},"result":{"Requeue":false,"RequeueAfter":0},"duration":"324.429155ms"}
{"level":"info","ts":"2023-10-27T10:25:17Z","logger":"crossplane","msg":"Reconciling","controller":"claim/nopresources.nop.example.org","request":{"name":"apiextensions-composition-minimal","namespace":"default"}}
{"level":"info","ts":"2023-10-27T10:25:17Z","logger":"crossplane","msg":"Reconciliation done","controller":"composite/xnopresources.nop.example.org","request":{"name":"apiextensions-composition-minimal-pdh85"},"result":{"Requeue":false,"RequeueAfter":60000000000},"duration":"330.229859ms"}
{"level":"info","ts":"2023-10-27T10:25:17Z","logger":"crossplane","msg":"Reconciling","controller":"composite/xnopresources.nop.example.org","request":{"name":"apiextensions-composition-minimal-pdh85"}}
{"level":"info","ts":"2023-10-27T10:25:17Z","logger":"crossplane","msg":"Reconciliation done","controller":"claim/nopresources.nop.example.org","request":{"name":"apiextensions-composition-minimal","namespace":"default"},"result":{"Requeue":false,"RequeueAfter":0},"duration":"127.127239ms"}
{"level":"info","ts":"2023-10-27T10:25:17Z","logger":"crossplane","msg":"Reconciliation done","controller":"composite/xnopresources.nop.example.org","request":{"name":"apiextensions-composition-minimal-pdh85"},"result":{"Requeue":false,"RequeueAfter":60000000000},"duration":"168.077422ms"}
```

A json log viewer like [fblog](https://github.com/brocode/fblog) or [humanlog](https://github.com/humanlogio/humanlog) can be applied for further analysis.

I have: 

- ~[ ] Added or updated unit tests.~

### DIFF
--- a/pkg/controller/interceptor/interceptors.go
+++ b/pkg/controller/interceptor/interceptors.go
@@ -1,0 +1,47 @@
+/*
+Copyright 2020 The Crossplane Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package interceptor offers reconciler interceptors
+package interceptor
+
+import (
+	"context"
+	"time"
+
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	"github.com/crossplane/crossplane-runtime/pkg/logging"
+)
+
+// LoggingInterceptor wraps a Reconciler and logs messages
+// before and after the reconciliation,
+// reporting if the operation was successful together with its duration.
+func LoggingInterceptor(log logging.Logger, r reconcile.Reconciler) reconcile.Func {
+	return func(ctx context.Context, req reconcile.Request) (res reconcile.Result, err error) {
+		l := log.WithValues("request", req)
+		l.Info("Reconciling")
+		start := time.Now()
+		defer func() {
+			d := time.Since(start).String()
+			if err == nil {
+				l.Info("Reconciliation done", "result", res, "duration", d)
+			} else {
+				l.Info("Reconciliation failed", "error", err, "duration", d)
+			}
+		}()
+		return r.Reconcile(ctx, req)
+	}
+}
--- a/internal/controller/apiextensions/definition/reconciler.go
+++ b/internal/controller/apiextensions/definition/reconciler.go
@@ -59,6 +59,7 @@ import (
 	apiextensionscontroller "github.com/crossplane/crossplane/internal/controller/apiextensions/controller"
 	"github.com/crossplane/crossplane/internal/features"
 	"github.com/crossplane/crossplane/internal/xcrd"
+	"github.com/crossplane/crossplane/pkg/controller/interceptor"
 )
 
 const (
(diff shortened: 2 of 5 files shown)


pkg/controller/interceptor/interceptors.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/crossplane/crossplane-runtime/pkg/logging"
+)
+
+// LoggingInterceptor wraps a Reconciler and logs messages
+// before and after the reconciliation,
+// reporting if the operation was successful together with its duration.
+func LoggingInterceptor(log logging.Logger, r reconcile.Reconciler) reconcile.Func {
+	return func(ctx context.Context, req reconcile.Request) (res reconcile.Result, err error) {
+		l := log.WithValues("request", req)
+		l.Info("Reconciling")
```
negz, line 35: I don't think we should include this log line. It should be enough to log only when the reconcile completes. That log line contains all the same information, with the added benefit of knowing the duration.

The only reasons I can think of to _also_ log when a reconcile starts would be:

* You're particularly concerned about emitting a log line before a potential crash in the reconciler.
* You're actively watching logs and want to see when something is happening.

For the above cases debug logging should be enough, and I'm 99% sure controller-runtime already emits a debug log when each reconcile starts.

Remember that logs are expensive to emit, store, etc relative to other metrics. This is part of why I believe we should be very sparing with what we emit at info level.

pkg/controller/interceptor/interceptors.go
```diff
@@ -0,0 +1,47 @@
...
+		start := time.Now()
+		defer func() {
+			d := time.Since(start).String()
+			if err == nil {
+				l.Info("Reconciliation done", "result", res, "duration", d)
+			} else {
+				l.Info("Reconciliation failed", "error", err, "duration", d)
+			}
```
negz, line 43:
```suggestion
			l.Info("Reconciled", "result", res, "error", err, "duration", d)
```

I think this would work fine, right? By using a single case regardless of success/failure you could more easily query for `msg == "Reconciled"` in a log aggregator to see all results, successful or not. You could still find successful or error cases by filtering on the the `error` field (i.e. nil or not).

pkg/controller/interceptor/interceptors.go
```diff
@@ -0,0 +1,47 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package interceptor offers reconciler interceptors
+package interceptor
```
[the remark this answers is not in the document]
pedjak: > Would it make sense to add these to crossplane-runtime instead? That's typically where we put generic reconciler-related things, which this appears to be.

I am not biased here, thought that crossplane-runtime contains common code used by many projects. This one is used by two controllers (claim and composite) only. Would we like to broader the scope and add this to all providers and the rest of controllers in crossplane as well?

Issue #4162, opened 2023-06-08
Deletion from composition is not propagated to the composed objects


darkmuggle:
### What happened?

Composed objects where a new version of the composition removes a value did not remove it from the composed objects. Instead, the new composition merged with the existing. The effect is that if you want to delete the value, you have to manually patch it.

This was observed with both the Helm and Kubernetes providers. 

### How can we reproduce it?

Consider this composition: 
```
---
apiVersion: apiextensions.crossplane.io/v1
kind: Composition
metadata:
  name: kubeburnerservices.mcp.upbound.io 
spec:
  compositeTypeRef:
    apiVersion: mcp.upbound.io/v1alpha1
    kind: KubeBurnerService
  resources:
    - name: release-prometheus
      base:
        apiVersion: helm.crossplane.io/v1beta1
        kind: Release
        metadata:
          name: prometheus
        spec:
          providerConfigRef:
            name: host-cluster
          forProvider:
            chart:
              name: kube-prometheus-stack
              repository: "https://prometheus-community.github.io/helm-charts"
            namespace: kube-burner
            wait: true
            values:
               foo: bar
```

Then update the `spec.resources[0].spec.forProvider` with:
```
values:
  bar: foo
```

On reconciliation, the composed object will have:
```
values:
   foo: bar
   bar: foo
```
              
### What environment did it happen in?
Crossplane version: 1.12.0

negz:
This sounds like one of the problems described in https://github.com/crossplane/crossplane/issues/4047:

> Patches will never remove a field from an object; they're only additive.

I think the solution is as proposed by that issue - start using server-side apply instead of our bespoke client-side patching.

akesser:
This is also a problem on our side. We use claims to allow DevOps guys to create managed resources by them self, e.g. RDS dbinstances in aws. If they misconfigure something, eg. by using invalid filed combinations, we need to manually edit the managed resources to remove properties because an update of the claim or composite will not fix the problem.

This is a time-consuming manual process and people use this as an excuse to blame crossplane for all sorts of things.

Fixing this would allow a better user experience and would let us spend less time on manual interventions on the cluster

pedjak:
it is still relevant.

negz:
I think the way forward here is to use composition functions.

https://docs.crossplane.io/v1.14/concepts/composition-functions/

[...]

Issue #3993, opened 2023-04-17
MergeOption: appendSlice is still overwriting instead of appending


b-m-f:
Same problem as https://github.com/crossplane/crossplane/issues/2690.

Using 
```
        mergeOptions:
          appendSlice: true
          keepMapValues: true
```
as described in the [docs](https://docs.crossplane.io/v1.9/reference/composition/#compositions) leads to whole values being replaced instead of having arrays appended to each other.

Running `helm.sh/chart: universal-crossplane-1.11.1-up.1`.

I was able to trace the code https://github.com/crossplane/crossplane/blob/2015c64c48aba497ef87275b62d78f77e5e41498/internal/controller/apiextensions/composite/merge.go#L33 but could not find where it is actually being invoked.

Where previously https://github.com/crossplane/crossplane/blob/2015c64c48aba497ef87275b62d78f77e5e41498/apis/apiextensions/v1/composition_patches.go#L198 applied the MergeOptions this does not seem to be done on [master](https://github.com/crossplane/crossplane/blob/9a0f0acb31c6b94b62262ad83cbd3899f9505b4d/apis/apiextensions/v1/composition_patches.go) anymore.

Thats where I stopped digging and decided to open this Issue.

phisco:
This sounds like a behaviour we want to investigate further to check whether we introduced a regression at some point.
@b-m-f, do you happen to have an example manifest reproducing the issue?

b-m-f:
Will double check next week and come up with a manifest.

b-m-f:
Well, didn't recheck yet, but I'll post here when I get to it ;)

AaronME:
We are also seeing that append is failing for us.

This works. but when mapOne is absent, we get a "null" entry in the first position.
```    patches:
    - type: FromCompositeFieldPath
      fromFieldPath: status.mapOne
      toFieldPath: spec.forProvider.fields.maps[0]
      policy:
        mergeOptions:
          appendSlice: true
    - type: FromCompositeFieldPath
      fromFieldPath: status.mapTwo
      toFieldPath: spec.forProvider.fields.maps[1]
      policy:
        mergeOptions:
          appendSlice: true
```

This will overwrite mapOne instead of appending:
```
    patches:
    - type: FromCompositeFieldPath
      fromFieldPath: status.mapOne
      toFieldPath: spec.forProvider.fields.maps[0]
      policy:
        mergeOptions:
          appendSlice: true
    - type: FromCompositeFieldPath
[...]

PR #5144, opened 2023-12-18
Fix synchronization between composition/XR and composed resources

pedjak:

* Removing a field in P&T template is correctly propagated to the composed resource
* Changes to composed resources performed by providers and other non-crossplane actors
  are preserved

The changes are propagated using [k8s server-side apply](https://kubernetes.io/docs/reference/using-api/server-side-apply/).

Implementing the server-side apply requested the following changes on the composite reconciler:

* Given that the submitted patch needs to contain all fields that a reconciler
  care about and wants to take the ownership, the existing configurators and composers
  cannot perform partial updates.
* In order to minimize code changes, `overlayComposite` type got introduced to
  catch all changes on the composite during the reconciliation, applying them
  as a patch with the help of `overlayAwareClient`.
* Composite status subresource is still modified by sending UPDATE requests, because
  the controller should own it completely. That is expressed explicitly by setting
  the status field ownership.
* `composition_function.go` adjusted to use `overlayAwareClient` and to let
  reconciler update the composite status at the end.
* in order to adjust to the improved added by `composition_function.go`,
  changes on composite `.spec` and `.metadata` are ignored even for P&T compositions
* unit tests reworked/updated to verify the new implementation
* e2e test added
* resource migration from Crossplane 1.13+ supported

Fixes:

* #3335
* #4162 
* #3993 

Part of https://github.com/crossplane/crossplane/issues/4047 epic.

Implements: https://github.com/crossplane/crossplane/issues/4581

I have: 

### DIFF
--- a/internal/controller/apiextensions/composite/overlay.go
+++ b/internal/controller/apiextensions/composite/overlay.go
@@ -1,0 +1,141 @@
+/*
+Copyright 2023 The Crossplane Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package composite
+
+import (
+	"context"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
+	"github.com/crossplane/crossplane-runtime/pkg/meta"
+	"github.com/crossplane/crossplane-runtime/pkg/resource"
+	"github.com/crossplane/crossplane-runtime/pkg/resource/unstructured/composite"
+)
+
+var _ resource.Composite = &overlayComposite{}
+var _ runtime.Unstructured = &overlayComposite{}
+
+type overlayComposite struct {
+	*composite.Unstructured
+	patch *composite.Unstructured
+}
+
+func newOverlayComposite(base *composite.Unstructured) *overlayComposite {
+	p := &composite.Unstructured{}
+	p.SetKind(base.GetKind())
+	p.SetAPIVersion(base.GetAPIVersion())
+	p.SetName(base.GetName())
+
+	return &overlayComposite{Unstructured: base, patch: p}
+}
+
+func (r *overlayComposite) GetCompositionReference() *corev1.ObjectReference {
+	if ref := r.patch.GetCompositionReference(); ref != nil {
+		return ref
+	}
+	return r.Unstructured.GetCompositionReference()
+}
+
+func (r *overlayComposite) SetCompositionReference(ref *corev1.ObjectReference) {
+	r.patch.SetCompositionReference(ref)
+	r.Unstructured.SetCompositionReference(ref)
+}
+
+func (r *overlayComposite) GetCompositionRevisionReference() *corev1.ObjectReference {
+	if ref := r.patch.GetCompositionRevisionReference(); ref != nil {
+		return ref
+	}
+	return r.Unstructured.GetCompositionRevisionReference()
+}
+
+func (r *overlayComposite) SetCompositionRevisionReference(ref *corev1.ObjectReference) {
+	r.patch.SetCompositionRevisionReference(ref)
+	r.Unstructured.SetCompositionRevisionReference(ref)
+}
+
+func (r *overlayComposite) GetEnvironmentConfigReferences() []corev1.ObjectReference {
+	if refs := r.patch.GetEnvironmentConfigReferences(); len(refs) > 0 {
+		return refs
+	}
+	return r.Unstructured.GetEnvironmentConfigReferences()
+}
+
+func (r *overlayComposite) SetEnvironmentConfigReferences(refs []corev1.ObjectReference) {
+	r.patch.SetEnvironmentConfigReferences(refs)
+	r.Unstructured.SetEnvironmentConfigReferences(refs)
+}
+
+func (r *overlayComposite) SetResourceReferences(refs []corev1.ObjectReference) {
+	r.patch.SetResourceReferences(refs)
+	r.Unstructured.SetResourceReferences(refs)
+}
+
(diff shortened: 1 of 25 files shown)


internal/controller/apiextensions/composite/overlay.go
```diff
@@ -0,0 +1,141 @@
...
+	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
+	"github.com/crossplane/crossplane-runtime/pkg/meta"
+	"github.com/crossplane/crossplane-runtime/pkg/resource"
+	"github.com/crossplane/crossplane-runtime/pkg/resource/unstructured/composite"
+)
+
+var _ resource.Composite = &overlayComposite{}
+var _ runtime.Unstructured = &overlayComposite{}
+
+type overlayComposite struct {
```
turkenh, line 36: I am not sure about this. At a high level:

a. Do we really need Server Side Apply for Composite Controller to XR spec? Does it solve a real problem?
b.  (if ^ is "yes") What we're trying to achieve here looks similar to [UnstructuredExtractor](https://github.com/kubernetes/client-go/blob/5a0a4247921dd9e72d158aaa6c1ee124aba1da80/applyconfigurations/meta/v1/unstructured.go#L38) of client-go. Can we leverage it instead? 

Regarding (a), when I compare latest master with this PR, here is what I am seeing in the managedFields of the composite when I follow the steps [here](https://github.com/turkenh/xp-ssa-experiments/tree/main/issue-4162#reproduction-steps).

Screenshot 2023-12-26 at 14 30 00

Obviously, the values for **manager** and **operation** is looking better with this PR, but are there any advantage in terms of functionality that would justify the changes here? Can we replace the updates (e.g. [this](https://github.com/crossplane/crossplane/blob/a7a6049900e26b23fe37551a0f0971214bb7f64b/internal/controller/apiextensions/composite/api.go#L265) one) with a json patch given all we need is to patch a specific field (e.g. compositionRef.name) with the expected field owner, e.g. `apiextensions.crossplane.io/composite` ?

turkenh (commented):
> Fixes:
> 
> * [Deletion from composition is not propagated to the composed objects #4162](https://github.com/crossplane/crossplane/issues/4162)

Did you have a chance to test this PR with #4162? When I follow the steps [here](https://github.com/turkenh/xp-ssa-experiments/tree/main/issue-4162#reproduction-steps), I am still observing duplicate keys as before.

```yaml
kind: NopResource
metadata:
  ...
  managedFields:
  - apiVersion: nop.crossplane.io/v1alpha1
    fieldsType: FieldsV1
    fieldsV1:
      f:metadata:
        f:annotations:
          f:crossplane.io/composition-resource-name: {}
        f:generateName: {}
        f:labels:
          f:crossplane.io/claim-name: {}
          f:crossplane.io/claim-namespace: {}
          f:crossplane.io/composite: {}
        f:ownerReferences:
          k:{"uid":"545ea863-a4c9-4985-929d-03b5d11c9f9e"}: {}
      f:spec:
        f:forProvider:
          f:conditionAfter: {}
          f:fields:
            f:someField: {}
            f:values:
              .: {}
              f:bar: {}
              f:foo: {}
    manager: apiextensions.crossplane.io/composed
    operation: Apply
    time: "2023-12-26T11:22:08Z"
  - apiVersion: nop.crossplane.io/v1alpha1
    fieldsType: FieldsV1
    fieldsV1:
      f:metadata:
        f:annotations:
          f:crossplane.io/external-name: {}
        f:finalizers:
          .: {}
          v:"finalizer.managedresource.crossplane.io": {}
    manager: provider-nop
    operation: Update
    time: "2023-12-26T11:20:51Z"
  - apiVersion: nop.crossplane.io/v1alpha1
    fieldsType: FieldsV1
    fieldsV1:
      f:status:
        .: {}
        f:atProvider: {}
        f:conditions: {}
    manager: provider-nop
    operation: Update
    subresource: status
    time: "2023-12-26T11:21:01Z"
  name: apiextensions-composition-pandt-fxsp6-zl84x
spec:
    ...
    fields:
      someField: someValue
      values:
        bar: foo
        foo: bar
```

internal/controller/apiextensions/composite/overlay.go
pedjak: > Did you have a chance to test this PR with https://github.com/crossplane/crossplane/issues/4162? When I follow the steps [here](https://github.com/turkenh/xp-ssa-experiments/tree/main/issue-4162#reproduction-steps), I am still observing duplicate keys as before.

Yes, and just tried once again and it still works. Also, [the added e2e test](https://github.com/crossplane/crossplane/pull/5144/files#diff-d236b9f66a48e924036c2e8dc50733e0aa54076e023b8d2ebd690987a73f8867R258) checks exactly that issue for regressions.

> a. Do we really need Server Side Apply for Composite Controller to XR spec? Does it solve a real problem?

IMHO, it does - it clearly divides the field ownership between two controllers and hence also avoid update conflicts, because they update two distinct sets of fields.  On the screenshot the difference is not just the manager/operation pair - the right handside owns less fields.

Status subresource should be updated further with `Update()` calls, because the composite controller is the exclusive owner of it as the whole.

SSA is already used by Function composer, so it should be good to consolidate us here.

`overlayComposite` was introduced in order to minimize the changes in the rest of the system, since the composite `.spec` is modified inside composer, not the controller, and at that point we wanted to catch all changes that were performed by a number of configurators/env fetchers and the selected composer.

> What we're trying to achieve here looks similar to [UnstructuredExtractor](https://github.com/kubernetes/client-go/blob/5a0a4247921dd9e72d158aaa6c1ee124aba1da80/applyconfigurations/meta/v1/unstructured.go#L38) of client-go. Can we leverage it instead?

`UnstructuredExtractor` gives back a partial resource, populated with the fields owned by the specified field manager.  Here, during the first reconcile loop we need to add to the composite fields owned by the controller. Given that the fields are still not persisted at the api server, I do not see how we could use the extractor.

negz:
I'm wondering whether we should proceed with this PR.

I believe this updates the "native P&T" Composition code path to use SSA. In https://github.com/crossplane/crossplane/issues/4746 I've proposed we deprecate this codepath in Crossplane v1.17. That's two releases and several months away, but I wonder whether it's worth the churn to update these controllers.

I imagine we'd need to put this change behind a feature flag. We'd need a feature flag because this change would have behavior changes. For example merge semantics being affected by [SSA merge strategy markers](https://kubernetes.io/docs/reference/using-api/server-side-apply/#merge-strategy). 

If we introduce the change behind a flag in v1.16, by the time the flag hits beta with v1.17 we'll have marked native P&T deprecated.

Plus, I assume there would be a negligible difference in effort to migrate to this feature compared to migrating to composition functions (which already use SSA).

Assuming this feature is behind a flag the migration path would look something like:

1. Enable flag (or update to a version where flag is on by default)
2. Make sure your XRs still work as expected

Whereas to migrate to https://github.com/crossplane-contrib/function-patch-and-transform/ the path is something like:

1. Run `crossplane beta convert` to update Composition to use functions
2. Apply updated Composition to create a new CompositionRevision
3. Make sure your XRs still work as expected

negz:
I don't think we're planning to proceed here. Let me know if that's not the case and we can reopen.

PR #5148, opened 2023-12-19
feat(transform): Support string joining arrays

MisterMX:

Add a new string tranform `Join` that wraps around `strings.Join` and converts any given array into a single string separated by a user defined string.

Docs PR: https://github.com/crossplane/docs/pull/662

I have: 


pkg/validation/apiextensions/v1/composition/patches.go
```diff
@@ -500,6 +500,10 @@ func IsValidInputForTransform(t *v1.Transform, fromType v1.TransformIOType) erro
 			default:
 				return errors.Errorf("unknown string conversion type %s", *t.String.Convert)
 			}
+		case v1.StringTransformTypeJoin:
+			if t.String.Join == nil {
+				return errors.Errorf("string transform join type is required for join transform")
+			}
```
phisco, line 506: here the fromType should be checked, given that this transform will only work with arrays as input, you should change this to:
```suggestion
		case v1.StringTransformTypeJoin:
			if fromType != v1.TransformIOTypeArray {
				return errors.Errorf("string transform join type can only be used with arrays as input, got %s", fromType)
			}
```

MisterMX: Added.

phisco: It shouldn't be this function's job to validate the correctness of the API, just that the types match. We check that for `StringTransformTypeConvert` above only because we need to further switch on it to know the correct type.  That's why I was suggesting replacing it altogether instead of adding it. But that doesn't change much, it's just a nit. Let me know if you want to address that or we can proceed further.

MisterMX: I was assuming it would to that because it does so in https://github.com/crossplane/crossplane/pull/5148/files#diff-7728e09aa45c2c48325829b0553db452c72cbff4ff04874ba739e36e0f3d7175R490

MisterMX: I removed the changes

phisco (approved):
LGTM, thanks!

PR #5163, opened 2023-12-20
Do not reconcile Provider and Functions on serviceaccount and secret changes

### LINKED ISSUE #5142: Providers and Functions are constantly reconciled on OpenShift because the ServiceAccount imagePullSecret is modified
Slack discussion:  https://crossplane.slack.com/archives/CEG3T90A1/p1701165506306089?thread_ts=1701096709.193549&cid=CEG3T90A1

### What happened?
Crossplane is constantly reconciling the providers and functions when deployed on OpenShift because the ServiceAccount `imagePullSecrets` field is constantly being modified by a webhook:

https://github.com/openshift/openshift-controller-manager/blob/2a28a78279b127075b399a3442a2042f4a7d190d/pkg/serviceaccounts/controllers/create_dockercfg_secrets.go#L315-L400

### How can we reproduce it?
Deploy providers/functions on OpenShift

Run `kubectl get sa -n crossplane-system -w` and observe that the service accounts for the providers and functions are continuously modified.

Crossplane version: 1.14.4

* Kubernetes distribution (e.g. Tectonic, GKE, OpenShift) OpenShift

pedjak:

After provider/function deployment got created there is no need to watch
the created service account and secrets for changes:

* changes to secrets are propagated automatically to the pods:

https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-files-from-a-pod

* the service account is referred inside the deployment by its name,
and its content has no impact on the provider/function reconciliation

We should allow other controllers to decorate our service account with
extra annotations/labels/pull secrets and there is no need to react on
these changes.

Openshift is known to use a huge number of secrets for managing various
internal states. They are changed quite frequently.

A fresh OpenShift cluster with no user workload contains >1200 secrets.

Such number of secrets creates a pressure on our informer - it needs to
process all 1200 secrets at least initially, even if none of them is
owned by Crossplane.

Fixes #5142 

I have: 

- ~[ ] Added or updated unit tests.~

### DIFF
--- a/internal/controller/pkg/revision/reconciler.go
+++ b/internal/controller/pkg/revision/reconciler.go
@@ -299,8 +299,6 @@ func SetupProviderRevision(mgr ctrl.Manager, o controller.Options) error {
 		For(&v1.ProviderRevision{}).
 		Owns(&appsv1.Deployment{}).
 		Owns(&corev1.Service{}).
-		Owns(&corev1.Secret{}).
-		Owns(&corev1.ServiceAccount{}).
 		Watches(&v1alpha1.ControllerConfig{}, &EnqueueRequestForReferencingProviderRevisions{
 			client: mgr.GetClient(),
 		})
@@ -407,8 +405,6 @@ func SetupFunctionRevision(mgr ctrl.Manager, o controller.Options) error {
 		For(&v1beta1.FunctionRevision{}).
 		Owns(&appsv1.Deployment{}).
 		Owns(&corev1.Service{}).
-		Owns(&corev1.Secret{}).
-		Owns(&corev1.ServiceAccount{}).
 		Watches(&v1alpha1.ControllerConfig{}, &EnqueueRequestForReferencingFunctionRevisions{
 			client: mgr.GetClient(),
 		})


pedjak:
this is partial revert of d2e256ea0a1a9f3216ec3649cdf3d4d4b6f2d1df.

IMHO, we could also stop watching services.

phisco:
AFAICT, this doesn't fix #5142. It only reduces the rate at which the issue happens, both us and openshift will still reconcile the service account with different specs and hit the very same issue.

I would have a look at something like https://github.com/kubernetes-sigs/controller-runtime/blob/main/designs/use-selectors-at-cache.md maybe.

pedjak:
> this doesn't fix https://github.com/crossplane/crossplane/issues/5142.

It does stop the constant reconcile on Openshift, because it is not triggered on service account/secret changes.

>  It only reduces the rate at which the issue happens, both us and openshift will still reconcile the service account with different specs and hit the very same issue.

Not sure if I understood what you mean, but this PR turns off reconcile on service accounts/secret changes. Do we have some other controller watching for service accounts?

What was the motivation for watching service accounts/secrets? https://github.com/crossplane/crossplane/commit/d2e256ea0a1a9f3216ec3649cdf3d4d4b6f2d1df was added as a part of #4656 - the scope was switching to new TLS certificates, right?

phisco:
> It does stop the constant reconcile on Openshift, because it is not triggered on service account/secret changes.

The fact that resources are watched and trigger a reconciliation of their owner, only makes the issue worst because in case of another controller having opinions on the same resource, the two are going to continuously update the resource again and again. However, removing the watch only means we are going to hit that less often, not that it's not going to happen anymore. In the case of the service account on openshift, by doing so, next time we reconcile for any reason a package, we'll still update the SA, and the openshift's controller is going to reconcile it again and so.

It's pretty common to watch owned resources, not doing so means that if for any reason they get modified or deleted, you'll need to wait for the configured poll interval at most before reconciling them. Would that be acceptable? Sure! But maybe there is a way to resolve the original issue completely, optimizing memory consumption and still be reactive.

pedjak:
> It's pretty common to watch owned resources, not doing so means that if for any reason they get modified or deleted, you'll need to wait for the configured poll interval at most before reconciling them. 

We should take into account the semantic of the data being changed. In case of secrets, for this concrete case we manage there TLS data. If someone else modifies it, what should be our reaction? Regenerate certificates? Remove added labels/annotations? If a change on TLS secret should end of in a no-op reconcile, then we should not watch for secrets at all.

bobh66:
In the specific case of ServiceAccount I think the controller should be smart enough to determine that the SA contains the `imagePullSecrets` entry that the controller cares about and leave it alone.  It should not remove the additional `imagePullSecrets` entry that is being added by the OpenShift or other controller.

bobh66:
I applied this change to my OCP cluster and verified that it stops the constant reconciliation.  Is there any reason not to start with this and make additional changes later if needed?

phisco:
@bobh66 see my comment above: https://github.com/crossplane/crossplane/pull/5163#issuecomment-1865221418

This is not preventing that from happening, just reducing drastically its frequency, I would have preferred a better approach as the one you suggested, properly handling the update.

bobh66:
Agreed, but it's the controller-runtime that is doing the update, so how do we tell the controller that it should ignore entries in `imagePullSecrets` as long as the one we specified is there?

bobh66:
It seems to me that the root cause of this issue is that the `imagePullSecrets` attribute of the `ServiceAccount` does not have the patch merge type and patch merge key so the reconcile is not merging properly.

When I look at the latest `ServiceAccount` definition the merge type was recently updated to `atomic`:

https://github.com/kubernetes/api/blob/master/core/v1/types.go#L5513

https://github.com/kubernetes/api/commit/0c53f558a11de4224cc78d7967fc82f274a0540b

and based on this comment in Slack:

https://kubernetes.slack.com/archives/CJUQN3E4T/p1709663646257159

it seems that the expectation is that only one controller will manage `imagePullSecrets`

bobh66:
@phisco as much as I would like to fix this "correctly" I don't know that there is a really good solution beyond what this PR providers.

Since the controller owns the ServiceAccount it has a watch that notifies the controller of the update when the Openshift controllers adds an imagePullSecret, which causes the reconciler to rebuild and reapply the manifests:

https://github.com/crossplane/crossplane/blob/master/internal/controller/pkg/revision/reconciler.go#L848

which wipes out the added imagePullSecret, and around and around it goes.

A couple of other possibilities are:
- disable the Update predicate on the Owns/Watch so that the controller is not notified of updates - maybe this is better than nothing?
- add logic to the manifest builder to look for an existing SA and retrieve any existing imagePullSecrets that have been added and include them in the desire manifest

Thoughts?

bobh66:
This PR has been replaced by #5558 Thanks for your help @pedjak

PR #5174, opened 2023-12-25
Re-add ownership for existing CompositionRevisions after backup/restore

turkenh:

Owner references are stripped out when a resource is moved from one cluster to another (i.e. backup/restore) since the UID of the resources (i.e. owner) is not preserved. We need to make sure to re-add the owner reference to existing compositions revisions.

Currently, observing the following error after restoring the Crossplane state on another control plane using a backup. 

In Composite: 
```
  Warning  ComposeResources  11m (x15 over 19m)    defined/compositeresourcedefinition.apiextensions.crossplane.io  cannot fetch Composition: no compatible CompositionRevisions found
```

In Composition:
```
  Warning  CreateRevision  24m (x101 over 119m)  revisions/composition.apiextensions.crossplane.io  compositionrevisions.apiextensions.crossplane.io "xnopresources.nop.example.org-fc2bf72" already exists
```

I have: 

### DIFF
--- a/internal/controller/apiextensions/composition/reconciler.go
+++ b/internal/controller/apiextensions/composition/reconciler.go
@@ -51,6 +51,7 @@ const (
 	errGet             = "cannot get Composition"
 	errListRevs        = "cannot list CompositionRevisions"
 	errCreateRev       = "cannot create CompositionRevision"
+	errOwnRev          = "cannot own CompositionRevision"
 	errUpdateRevStatus = "cannot update CompositionRevision status"
 	errUpdateRevSpec   = "cannot update CompositionRevision spec"
 )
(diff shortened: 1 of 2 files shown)


internal/controller/apiextensions/composition/reconciler.go
```diff
@@ -180,6 +176,13 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 			continue
 		}
 
+		// Make sure this revision is owned/controlled by the Composition.
+		// Note(turkenh): Owner references are stripped out when a resource is
+		// moved from one cluster to another (i.e. backup/restore) since the UID
+		// of the owner is not preserved. We need to make sure to re-add the
+		// owner reference here.
+		meta.AddOwnerReference(rev, meta.AsController(meta.TypedReferenceTo(comp, v1.CompositionGroupVersionKind)))
+
```
ezgidemirel, line 185: I think we should move this logic before the following check 
```
if rev.GetLabels()[v1.LabelCompositionHash] != currentHash[:63] {
	continue
}
```
Otherwise, we cannot set the owner reference to inactive revisions.

negz: https://pkg.go.dev/github.com/crossplane/crossplane-runtime@v1.14.3/pkg/meta#AddControllerReference

Would it make sense to use AddControllerReference rather than AddOwnerReference?

I think AddOwnerReference will make this composition the controller of the revision, even if another composition (somehow) already controls it. AddControllerReference will make the composition the controller only if it has no controller reference. If the revision already has a controller it will return an error.

ezgidemirel (approved):
Thanks @turkenh, LGTM!

phisco (approved):

PR #5178, opened 2023-12-29
Ensure ownerRef on objects for inactive package revisions

turkenh:

This PR, extends the `ReleaseObjects` method to also make sure the owner ref for the revision exist instead of just setting controller to false only there is the owner ref. This could be considered as an extra precaution not to lose ownership (hence trigger a garbage collection) on object (i.e. CRDs).

After https://github.com/crossplane/crossplane/pull/4071, we do not execute full pull/parse flow for inactive revisions with existing object references (i.e. `status.objectRefs`), rather only release objects (i.e. controller=false) so that active revisions could take control of them. 

Test Scenario:

1. Install provider-helm v0.15.0
2. Update provider-helm to v0.16.0
3. Edit CRD `releases.helm.crossplane.io` and remove owner ref for inactive revision.
4. Wait for some time (or put a random label on inactive revision to trigger a reconciliation)
5. Make sure manually removed owner ref is put back but not as a controller.

P.S.: The closest scenario which could be relevant here is the [migration](https://docs.upbound.io/providers/migration/#manual-migration) to family providers. During migration, an inactive provider revision of family providers create to take ownership of CRDs created by the monolith provider. However, since the inactive revision comes without having object references (i.e. `status.objectRefs`) set, the full pull/parse flow executed and ownerRef for inactive revision added to the objects. So, this PR is not mandatory there.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -115,7 +115,7 @@ require (
 	github.com/aws/aws-sdk-go-v2/service/sso v1.12.10 // indirect
 	github.com/aws/aws-sdk-go-v2/service/ssooidc v1.14.10 // indirect
 	github.com/aws/aws-sdk-go-v2/service/sts v1.19.0 // indirect
-	github.com/aws/smithy-go v1.13.5 // indirect
+	github.com/aws/smithy-go v1.13.5
 	github.com/awslabs/amazon-ecr-credential-helper/ecr-login v0.0.0-20230510185313-f5e39e5f34c7 // indirect
 	github.com/beorn7/perks v1.0.1 // indirect
 	github.com/bufbuild/protocompile v0.6.1-0.20231108163138-146b831231f7 // indirect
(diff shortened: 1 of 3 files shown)


phisco (approved):
We can think a bit more about whether to backport it or not, but the code looks good 🙏

turkenh:
> We can think a bit more about whether to backport it or not, but the code looks good 🙏

I was on the fence as well, however, I'll go for it with the following considerations:

- Change is small, hopefully with no side effect. 
- On a regular/standard operation, changed code will not execute.
- I don't want us to have different behavior in this regard only in v1.14 different than versions before and after.
- While we don't have an immediate bug that this PR fixes (yet), it could be catastrophic (CRD garbage collected, hence MRs deleted). Therefore, I tend to favor being extra cautious.

